PR #291, opened 2026-06-06
feat(sidebar): make recent-tab activity filter configurable per channel type

### LINKED ISSUE #289: feat(sidebar): make recent-tab activity filter configurable (per-channel-type window) via system_settings
### Problem / Motivation

`POST /v1/sidebar/sync` (`recent` tab) hides group and thread (community topic) conversations whose last activity is older than a **hard-coded 72-hour window**, while direct messages are always kept regardless of age.

See `modules/message/api_sidebar.go`:
- `threeDaysAgo()` (L631) — the window is hard-coded to `-72 * time.Hour`
- `buildRecentItems()` (L916/L920) — `if !isDM && conv.Timestamp <= cutoff { continue }`

This is a reasonable default, but it is **not adjustable at runtime and not differentiated by channel type**. Operators have asked to (a) turn the time filter on/off and (b) tune the window length, and crucially to do so **per channel type** — e.g. return *all* group conversations with no time limit while keeping the 3-day window for threads. Today this requires a code change and redeploy.

### Proposed Solution

Drive the recent-tab activity window from the existing `system_settings` mechanism (`modules/common`), with **per-channel-type day counts** and a sentinel value to disable filtering:

| setting key | type | default | meaning |
|---|---|---|---|
| `sidebar.recent_filter_group_days` | int | 3 | group window in days; `0` = no filter (return all groups) |
| `sidebar.recent_filter_thread_days` | int | 3 | thread (community topic) window; `0` = no filter |
| `sidebar.recent_filter_person_days` | int | 0 | DM window; default `0` keeps today's "always shown" behaviour |

Defaults reproduce **today's exact behaviour** (groups/threads = 3 days, DMs = unfiltered), so the change is zero-impact until an operator opts in. Setting `recent_filter_group_days = 0` yields the "all groups, no time limit" case requested.

### Alternatives Considered

- **Single global on/off + single global day count.** Simpler, but cannot express "all groups but keep the 3-day window for threads" — rejected because per-type control is the actual requirement.
- **One JSON setting `{group:..,thread:..,person:..}`.** Fewer keys, but harder to validate and edit from the admin UI than discrete int settings.
- **`featuregate` module.** Intended for second-level gradual rollouts, not stable global operator configuration; `system_settings` (schema-driven, ~60s reload, immediate per-instance reload on admin write) is the right fit.

### Technical Approach (optional)

- Add three int entries to the `system_settings` schema with typed getters in `modules/common`; getters clamp out-of-range values back to the default (defence-in-depth against direct DB edits).
- Unify "no filter" as `cutoff == 0`: `daysCutoff(days)` returns `0` when `days <= 0`, else `now - days*24h`. `buildRecentItems` selects the per-type cutoff by `conv.ChannelType` and skips the filter when `cutoff == 0`. DM exemption becomes data-driven (`person_days` default `0`) instead of the hard-coded `!isDM` branch.
- Add range validation `[0, 3650]` for int settings on the admin write path (`api_manager_system_setting.go` currently only does `strconv.Atoi` with no bounds check) — applies to these keys and closes a pre-existing gap.
- Tests: extend `modules/message/api_sidebar_integration_test.go` (per-type window on/off, `0` = unfiltered, out-of-range fallback, DM behaviour) and a `system_settings` getter unit test.

### Additional Context

Scope is limited to `POST /v1/sidebar/sync`'s `recent` tab. `POST /v1/conversation/sync` has **no** time-window filtering and is intentionally out of scope. The `follow` tab does not apply any time filter and is unaffected.

an9xyz:
Closes #289.

## Background

The `recent` tab of `POST /v1/sidebar/sync` hid group/thread (community topic) conversations whose last activity was older than a **hard-coded 72h window**, while DMs were always kept. The window was neither runtime-adjustable nor differentiated by channel type. Operators asked to (a) turn the time filter on/off and (b) tune the window length, **per channel type** (e.g. "return all groups, no time limit; keep the 3-day window for threads").

## What this does

Drives the recent-tab window from the existing `system_settings` mechanism, with **per-channel-type day counts** and a `0` sentinel that disables filtering.

| setting key | default | meaning |
|---|---|---|
| `sidebar.recent_filter_group_days` | `3` | group window in days; `0` = no filter (all groups) |
| `sidebar.recent_filter_thread_days` | `3` | thread window; `0` = no filter |
| `sidebar.recent_filter_person_days` | `0` | DM window; `0` keeps today's "always shown" behaviour |

**Defaults reproduce today's exact behaviour** (groups/threads = 3 days, DMs unfiltered), so this is zero-impact until an operator opts in. Setting `recent_filter_group_days = 0` yields the requested "all groups, no time limit" case.

### Dynamic, near-real-time
Configured via the existing admin endpoint `POST /v1/manager/common/system_setting` (SuperAdmin). The handler reloads the in-memory snapshot synchronously before returning, so the writing instance sees the new value on the **next** sidebar request; other instances converge within the ~60s auto-reload TTL. No redeploy.

## Changes

**`modules/common`**
- Three int settings added to the `system_setting` schema with typed getters.
- `getIntClamped` clamps out-of-range DB values back to the default (defence in depth against direct DB edits).
- `[0, 3650]` range validation for int settings on the admin write path — closes a pre-existing gap where the int branch only ran `strconv.Atoi` with no bounds check.

**`modules/message`**
- `threeDaysAgo()` replaced by `daysCutoff` + a per-type `recentCutoffs` struct loaded from `system_settings` (`loadRecentCutoffs`). `buildRecentItems` selects the cutoff by channel type and skips filtering when `cutoff == 0`. The DM exemption is now **data-driven** (person window default `0`) rather than a hard-coded `!isDM` branch. Unknown channel types are kept (recent tab only ever carries group/thread/DM).

## Tests

- `system_settings` getter unit tests: defaults, DB override, out-of-range clamp, boundary values.
- Admin write-path validation: rejects out-of-range / non-int, accepts boundaries and persists.
- `buildRecentItems` per-type window cases: group cutoff `0` = all kept, independent per-type windows, DM window now filters, DM `0` keeps all; plus a `daysCutoff` unit test.
- A DB-backed e2e covering the full path: admin HTTP write → settings reload → handler `loadRecentCutoffs` → filter.

Verified locally against MySQL + Redis (WuKongIM not required — the IM result is stubbed via synthetic conversation slices). `go build ./...`, `go vet`, `golangci-lint` (0 issues), and the i18n lints all pass.

## Out of scope
`POST /v1/conversation/sync` has no time-window filtering (intentionally untouched); the `follow` tab applies no time filter and is unaffected.

https://claude.ai/code/session_01PeCytGvTzrdXjHmVAPVLfN

---
_Generated by [Claude Code](https://claude.ai/code/session_01PeCytGvTzrdXjHmVAPVLfN)_

### DIFF
--- a/modules/common/api_manager_system_setting.go
+++ b/modules/common/api_manager_system_setting.go
@@ -201,10 +201,22 @@ func (m *Manager) updateSystemSettings(c *wkhttp.Context) {
 			}
 			p.value = normalised
 		case settingTypeInt:
+			// Empty value means "reset to default" (the getter treats an empty
+			// snapshot entry as not-configured), so only validate non-empty
+			// payloads. Bounds-check against [settingIntMin, settingIntMax] —
+			// previously this path only ran strconv.Atoi with no range guard,
+			// letting absurd windows (or negatives) reach the DB (issue #289).
 			if item.Value != "" {
-				if _, err := strconv.Atoi(item.Value); err != nil {
+				n, err := strconv.Atoi(item.Value)
+				if err != nil {
 					c.JSON(http.StatusBadRequest, jsonH{
 						"msg": fmt.Sprintf("%s.%s 必须是整数", item.Category, item.Key),
+					})
+					return
+				}
+				if n < settingIntMin || n > settingIntMax {
+					c.JSON(http.StatusBadRequest, jsonH{
+						"msg": fmt.Sprintf("%s.%s 必须在 [%d, %d] 范围内", item.Category, item.Key, settingIntMin, settingIntMax),
 					})
 					return
 				}
--- a/modules/common/api_manager_system_setting_test.go
+++ b/modules/common/api_manager_system_setting_test.go
@@ -363,3 +363,73 @@ func TestManagerSystemSetting_UpdatePersistsAndReloads(t *testing.T) {
 	assert.True(t, settings.RegisterEmailOn(), "Reload should run inside the update handler")
 	assert.Equal(t, "smtp.test:587", settings.SupportEmailSmtp())
 }
+
+// --- int range validation (issue #289) -------------------------------------
+
+func TestManagerSystemSetting_UpdateRejectsOutOfRangeInt(t *testing.T) {
+	t.Setenv(masterKeyEnv, "0123456789abcdef0123456789abcdef")
+	s, ctx := testutil.NewTestServer()
+	require.NoError(t, testutil.CleanAllTables(ctx))
+	require.NoError(t, ctx.Cache().Set(
+		ctx.GetConfig().Cache.TokenCachePrefix+testutil.Token,
+		testutil.UID+"@test@"+string(wkhttp.SuperAdmin),
+	))
+
+	for _, v := range []string{"-1", "3651", "100000"} {
+		body := []byte(`{"items":[{"category":"sidebar","key":"recent_filter_group_days","value":"` + v + `"}]}`)
+		w := httptest.NewRecorder()
+		req, _ := http.NewRequest("POST", "/v1/manager/common/system_setting", bytes.NewReader(body))
+		req.Header.Set("token", testutil.Token)
+		s.GetRoute().ServeHTTP(w, req)
+		assert.NotEqual(t, http.StatusOK, w.Code, "越界整数 %q 必须被拒绝", v)
+	}
+}
+
+func TestManagerSystemSetting_UpdateRejectsNonInt(t *testing.T) {
+	t.Setenv(masterKeyEnv, "0123456789abcdef0123456789abcdef")
+	s, ctx := testutil.NewTestServer()
+	require.NoError(t, testutil.CleanAllTables(ctx))
+	require.NoError(t, ctx.Cache().Set(
+		ctx.GetConfig().Cache.TokenCachePrefix+testutil.Token,
+		testutil.UID+"@test@"+string(wkhttp.SuperAdmin),
+	))
+
+	body := []byte(`{"items":[{"category":"sidebar","key":"recent_filter_group_days","value":"abc"}]}`)
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("POST", "/v1/manager/common/system_setting", bytes.NewReader(body))
+	req.Header.Set("token", testutil.Token)
+	s.GetRoute().ServeHTTP(w, req)
+	assert.NotEqual(t, http.StatusOK, w.Code)
+}
+
+func TestManagerSystemSetting_UpdateAcceptsInRangeIntBoundaries(t *testing.T) {
+	t.Setenv(masterKeyEnv, "0123456789abcdef0123456789abcdef")
+	s, ctx := testutil.NewTestServer()
+	require.NoError(t, testutil.CleanAllTables(ctx))
+	require.NoError(t, ctx.Cache().Set(
+		ctx.GetConfig().Cache.TokenCachePrefix+testutil.Token,
+		testutil.UID+"@test@"+string(wkhttp.SuperAdmin),
+	))
+
+	body := []byte(`{"items":[` +
+		`{"category":"sidebar","key":"recent_filter_group_days","value":"0"},` +
+		`{"category":"sidebar","key":"recent_filter_thread_days","value":"3650"},` +
+		`{"category":"sidebar","key":"recent_filter_person_days","value":"7"}` +
+		`]}`)
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("POST", "/v1/manager/common/system_setting", bytes.NewReader(body))
+	req.Header.Set("token", testutil.Token)
+	s.GetRoute().ServeHTTP(w, req)
+	require.Equal(t, http.StatusOK, w.Code, w.Body.String())
+
+	db := newSystemSettingDB(ctx)
+	rows, err := db.listAll()
+	require.NoError(t, err)
+	got := map[string]string{}
+	for _, r := range rows {
+		got[schemaKey(r.Category, r.KeyName)] = r.Value
+	}
+	assert.Equal(t, "0", got["sidebar.recent_filter_group_days"])
+	assert.Equal(t, "3650", got["sidebar.recent_filter_thread_days"])
+	assert.Equal(t, "7", got["sidebar.recent_filter_person_days"])
+}
--- a/modules/common/system_setting_schema.go
+++ b/modules/common/system_setting_schema.go
@@ -1,11 +1,38 @@
 package common
 
+import "strconv"
+
 // Value types accepted by system_setting.value_type.
 const (
 	settingTypeString    = "string"
 	settingTypeBool      = "bool"
 	settingTypeInt       = "int"
 	settingTypeEncrypted = "encrypted"
+)
+
+// settingIntMin / settingIntMax bound every settingTypeInt value, applied both
+// on the admin write path (api_manager_system_setting.go) and in the clamping
+// getters (getIntClamped). Today all int settings are day-window counts
+// (sidebar.recent_filter_*_days), for which [0, 3650] (0 .. ~10 years) is a
+// generous sane range; 0 is the documented "disable filter" sentinel. Adding
+// an int setting that needs a different range should move this to a per-key
+// field on settingDef — until then a single shared bound keeps the write path
+// simple and closes the pre-existing "no bounds check" gap (issue #289).
+const (
+	settingIntMin = 0
+	settingIntMax = 3650
+)
+
+// Sidebar recent-tab activity-filter defaults (issue #289). The recent tab of
+// POST /v1/sidebar/sync hides conversations whose last activity is older than
+// a per-channel-type window. These defaults reproduce the historical
+// hard-coded behaviour exactly (groups/threads = 3-day window, DMs unfiltered)
+// so the feature is zero-impact until an operator opts in. A value of 0
+// disables the window for that channel type (return all, no time limit).
+const (
+	defaultSidebarRecentFilterGroupDays  = 3
+	defaultSidebarRecentFilterThreadDays = 3
+	defaultSidebarRecentFilterPersonDays = 0
 )
 
 // settingDef is the canonical definition of a system_setting key.
@@ -55,6 +82,16 @@ var systemSettingSchema = []settingDef{
 	// 仍作 fallback,DB 行为单一真源。
 	{Category: "space", Key: "disable_user_create", Type: settingTypeBool, Description: "是否关闭普通用户创建空间入口",
 		Effective: func(s *SystemSettings) string { return boolToCanonical(s.SpaceDisableUserCreate()) }},
+
+	// Sidebar recent-tab activity filter — per-channel-type window in days for
+	// POST /v1/sidebar/sync 的 recent tab。0 = 关闭该类型的时间过滤（全量返回）。
+	// 默认复刻历史硬编码行为：群/话题 3 天窗口、DM 不过滤（issue #289）。
+	{Category: "sidebar", Key: "recent_filter_group_days", Type: settingTypeInt, Description: "最近会话-群聊活跃过滤窗口(天)，0=不过滤",
+		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.SidebarRecentFilterGroupDays()) }},
+	{Category: "sidebar", Key: "recent_filter_thread_days", Type: settingTypeInt, Description: "最近会话-话题(社区话题)活跃过滤窗口(天)，0=不过滤",
+		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.SidebarRecentFilterThreadDays()) }},
+	{Category: "sidebar", Key: "recent_filter_person_days", Type: settingTypeInt, Description: "最近会话-单聊(DM)活跃过滤窗口(天)，0=不过滤(默认)",
+		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.SidebarRecentFilterPersonDays()) }},
 
 	// Email server config — formerly yaml-only (Support.* in config.go).
 	{Category: "support", Key: "email", Type: settingTypeString, Description: "技术支持邮箱（发件人）",
--- a/modules/common/system_settings.go
+++ b/modules/common/system_settings.go
@@ -222,6 +222,18 @@ func (s *SystemSettings) getInt(category, key string, fallback int) int {
 	return parsed
 }
 
+// getIntClamped is getInt with range enforcement: a value outside
+// [settingIntMin, settingIntMax] — which the admin write path rejects, but a
+// direct DB edit could still introduce — falls back to the code default rather
+// than being served verbatim. Defence in depth for the int settings (D-289).
+func (s *SystemSettings) getIntClamped(category, key string, fallback int) int {
+	v := s.getInt(category, key, fallback)
+	if v < settingIntMin || v > settingIntMax {
+		return fallback
+	}
+	return v
+}
+
 func (s *SystemSettings) getEncrypted(category, key string, fallback string) string {
 	// Encrypted values are stored decrypted in the snapshot, so a plain
 	// lookup is sufficient. The dedicated method exists so callers — and
@@ -480,6 +492,29 @@ func parseSpaceDisableUserCreateEnv(v string) bool {
 		return true
 	}
 	return false
+}
+
+// ----- sidebar recent-tab activity filter (issue #289) -----
+
+// SidebarRecentFilterGroupDays returns the recent-tab activity window for group
+// conversations, in days. 0 disables the window (all groups returned). Defaults
+// to defaultSidebarRecentFilterGroupDays (3) — today's hard-coded behaviour.
+func (s *SystemSettings) SidebarRecentFilterGroupDays() int {
+	return s.getIntClamped("sidebar", "recent_filter_group_days", defaultSidebarRecentFilterGroupDays)
+}
+
+// SidebarRecentFilterThreadDays returns the recent-tab activity window for
+// thread (community topic) conversations, in days. 0 disables the window.
+func (s *SystemSettings) SidebarRecentFilterThreadDays() int {
+	return s.getIntClamped("sidebar", "recent_filter_thread_days", defaultSidebarRecentFilterThreadDays)
+}
+
+// SidebarRecentFilterPersonDays returns the recent-tab activity window for DM
+// conversations, in days. Defaults to 0, which keeps today's "DMs are always
+// shown regardless of age" behaviour; the per-type default makes the historical
+// hard-coded `!isDM` exemption data-driven.
+func (s *SystemSettings) SidebarRecentFilterPersonDays() int {
+	return s.getIntClamped("sidebar", "recent_filter_person_days", defaultSidebarRecentFilterPersonDays)
 }
 
 // SupportEmail returns the From address used by the SMTP sender.
--- a/modules/common/system_settings_test.go
+++ b/modules/common/system_settings_test.go
@@ -408,3 +408,53 @@ func TestSystemSettings_ConcurrentReadsAndReloads(t *testing.T) {
 	}
 	wg.Wait()
 }
+
+// ---------------------------------------------------------------------------
+// Sidebar recent-tab activity filter windows (issue #289)
+// ---------------------------------------------------------------------------
+
+// Defaults reproduce today's hard-coded behaviour: groups/threads = 3-day
+// window, DMs = unfiltered (0).
+func TestSystemSettings_SidebarRecentFilter_Defaults(t *testing.T) {
+	s := newTestSystemSettings(t, nil)
+	assert.Equal(t, 3, s.SidebarRecentFilterGroupDays(), "group window default 3 天")
+	assert.Equal(t, 3, s.SidebarRecentFilterThreadDays(), "thread window default 3 天")
+	assert.Equal(t, 0, s.SidebarRecentFilterPersonDays(), "DM 默认 0 = 不过滤")
+}
+
+// A configured DB value overrides the code default, including the 0 sentinel
+// that disables the filter for that channel type.
+func TestSystemSettings_SidebarRecentFilter_DBOverrides(t *testing.T) {
+	s := newTestSystemSettings(t, nil)
+	require.NoError(t, s.db.upsert("sidebar", "recent_filter_group_days", "0", settingTypeInt, ""))
+	require.NoError(t, s.db.upsert("sidebar", "recent_filter_thread_days", "7", settingTypeInt, ""))
+	require.NoError(t, s.db.upsert("sidebar", "recent_filter_person_days", "30", settingTypeInt, ""))
+	require.NoError(t, s.Reload())
+
+	assert.Equal(t, 0, s.SidebarRecentFilterGroupDays(), "DB 0 → 关闭群过滤（全量）")
+	assert.Equal(t, 7, s.SidebarRecentFilterThreadDays(), "DB 覆盖话题窗口")
+	assert.Equal(t, 30, s.SidebarRecentFilterPersonDays(), "DB 覆盖 DM 窗口")
+}
+
+// Out-of-range DB values (someone editing the table directly, bypassing the
+// admin API's range check) clamp back to the code default — defence in depth.
+func TestSystemSettings_SidebarRecentFilter_OutOfRangeClampsToDefault(t *testing.T) {
+	s := newTestSystemSettings(t, nil)
+	require.NoError(t, s.db.upsert("sidebar", "recent_filter_group_days", "-5", settingTypeInt, ""))
+	require.NoError(t, s.db.upsert("sidebar", "recent_filter_thread_days", "999999", settingTypeInt, ""))
+	require.NoError(t, s.Reload())
+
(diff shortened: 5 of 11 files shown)


yujiawei (dismissed):
# Code Review — PR #291 (octo-server)

## Summary

Makes the `recent` tab's activity filter (`POST /v1/sidebar/sync`) configurable per channel type via `system_settings`, replacing the hard-coded 72h window. Clean, well-scoped, and well-tested. Defaults reproduce the historical behaviour exactly, so the change is zero-impact until an operator opts in. **Approving.**

## Verification

I checked out the head SHA (`d11e2d7`), read the changed files plus the surrounding seam (`Sidebar.Sync`, `FilterRawConversationsBySpace`, `EnsureSystemSettings`, the channel-type constants in octo-lib), and confirmed:

- ✅ `go build ./...` passes; `go vet ./modules/message/...` clean.
- ✅ The pure unit tests (`TestBuildRecentItems*`, `TestDaysCutoff`) pass locally.
- ✅ **Defaults reproduce legacy behaviour exactly.** `daysCutoff(now, 3)` = `now.Add(-72h).Unix()`, byte-identical to the removed `threeDaysAgo()`; person default `0` → cutoff `0` → no filter, matching the old `!isDM` exemption.
- ✅ **Cutoff comparison semantics preserved.** New `cutoff != 0 && conv.Timestamp <= cutoff` keeps iff `timestamp > cutoff`, same boundary handling (`==cutoff` excluded) as before.
- ✅ **No other callers** of `threeDaysAgo()` / `buildRecentItems` outside the recent path — the signature change is fully accounted for across the test files.
- ✅ **No import cycle.** `modules/message` → `modules/common` is one-directional; `common` does not import `message`.
- ✅ **Validation gap genuinely closed.** The int write path previously ran only `strconv.Atoi` with no bounds check; `[0, 3650]` validation plus `getIntClamped` (defence-in-depth against direct DB edits) is a real, correct hardening of a pre-existing hole.
- ✅ `loadRecentCutoffs(time.Now())` computes all three cutoffs from a single `now`, and the snapshot read is via the thread-safe shared `SystemSettings` singleton.
- ✅ Tests are thorough: getter defaults/override/clamp/boundaries, admin write-path accept/reject, per-type window independence, DM-now-filtered, plus a DB-backed e2e covering the full admin-write → reload → filter path.

## Findings

No P0/P1 (correctness, security, data-loss, or functional blockers) found.

### P2 (non-blocking) — behavioural change for non-group/thread/DM channel types

`recentCutoffs.cutoffFor` returns `0` (no filter) for any channel type that isn't group/thread/person:

```go
// modules/message/api_sidebar.go:350
func (c recentCutoffs) cutoffFor(channelType uint8) int64 {
    switch channelType {
    case common.ChannelTypeGroup.Uint8():          return c.group
    case common.ChannelTypeCommunityTopic.Uint8(): return c.thread
    case common.ChannelTypePerson.Uint8():         return c.person
    default:                                        return 0   // no filter
    }
}
```

The old code applied the 72h window to **every non-DM type**:

```go
isDM := conv.ChannelType == common.ChannelTypePerson.Uint8()
if !isDM && conv.Timestamp <= cutoff { continue }
```

So for channel types other than group/thread/DM (e.g. `ChannelTypeCustomerService`, `ChannelTypeCommunity`, `ChannelTypeInfo`), the behaviour changes from "filtered by 72h" to "always kept". This is reachable: when the request carries no `X-Space-ID` (`SpaceMiddleware` is opt-in, `spaceID == ""`), `conversations` is the unfiltered raw IM result, which can include those types — they'd now appear in the recent tab regardless of age.

This is **intentional and documented** ("defaulting unknown types to kept avoids silently dropping a future channel type"), and the recent tab realistically carries group/thread/DM, so impact is minor. Flagging only so it's a conscious decision: if the intent is "unknown types should still obey *some* window," the `default` branch would need to pick one. As written, the `keep-unknown` choice is defensible — no change required to merge.

### Nits (optional)

- `daysCutoff`'s `days <= 0 → 0` guard doubles as the negative-value defence; combined with `getIntClamped` rejecting negatives, this is belt-and-suspenders in a good way. No action needed.
- The schema comment correctly anticipates that a future int setting needing a different range should move `settingIntMin/Max` to a per-key `settingDef` field. Good forward note.

## Verdict

Solid PR: a real pre-existing validation hole closed, exact behavioural parity by default, dynamic per-type tuning, and strong test coverage including an e2e. The single behavioural delta is for channel types the recent tab doesn't meaningfully carry and is a documented design choice.

Jerry-Xin (approved):
The PR is in scope for octo-server and cleanly implements configurable recent-tab filtering through the existing system settings path.

💬 Non-blocking

🟡 Warning — [modules/message/api_sidebar.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0ovh1hix/repo/modules/message/api_sidebar.go:15) has stale top-level behavior documentation: it still says recent keeps all DMs and filters groups/threads by `now-72h`. That is no longer always true once `sidebar.recent_filter_person_days` can be non-zero and group/thread windows can be disabled or changed.

🟡 Warning — [modules/common/api_manager_system_setting.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0ovh1hix/repo/modules/common/api_manager_system_setting.go:218) adds another raw non-OK `c.JSON` response in the manager settings path. This matches nearby legacy handling, but it still conflicts with the repository’s current i18n error-envelope guidance. Non-blocking because this endpoint is already using that legacy pattern.

🔵 Suggestion — [modules/message/api_sidebar.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0ovh1hix/repo/modules/message/api_sidebar.go:672) keeps unknown channel types unfiltered by returning cutoff `0`. The PR notes recent currently only carries group/thread/person, so this is acceptable; if new channel types are introduced later, consider making the default explicit in tests or routing unknowns through a conservative cutoff.

✅ Highlights

The main logic preserves historical defaults: groups and threads keep a 3-day window, DMs remain unfiltered by default. The `0` sentinel is simple and consistently applied.

The settings write path now validates integer bounds, and the read path defensively clamps direct DB edits.

Test coverage is strong around defaults, per-type independence, DM filtering opt-in, and boundary values.

Verification: `go test ./modules/message -run 'TestDaysCutoff|TestBuildRecentItems_'` passed. The targeted common-package test run could not complete because local MySQL was unavailable (`127.0.0.1:3306: connect: connection refused`).

Jerry-Xin (dismissed):
The PR is in scope for octo-server and cleanly implements configurable recent-tab filtering through the existing system settings path.

💬 Non-blocking

🟡 Warning — [modules/message/api_sidebar.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0ovh1hix/repo/modules/message/api_sidebar.go:15) has stale top-level behavior documentation: it still says recent keeps all DMs and filters groups/threads by `now-72h`. That is no longer always true once `sidebar.recent_filter_person_days` can be non-zero and group/thread windows can be disabled or changed.

🟡 Warning — [modules/common/api_manager_system_setting.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0ovh1hix/repo/modules/common/api_manager_system_setting.go:218) adds another raw non-OK `c.JSON` response in the manager settings path. This matches nearby legacy handling, but it still conflicts with the repository’s current i18n error-envelope guidance. Non-blocking because this endpoint is already using that legacy pattern.

🔵 Suggestion — [modules/message/api_sidebar.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0ovh1hix/repo/modules/message/api_sidebar.go:672) keeps unknown channel types unfiltered by returning cutoff `0`. The PR notes recent currently only carries group/thread/person, so this is acceptable; if new channel types are introduced later, consider making the default explicit in tests or routing unknowns through a conservative cutoff.

✅ Highlights

The main logic preserves historical defaults: groups and threads keep a 3-day window, DMs remain unfiltered by default. The `0` sentinel is simple and consistently applied.

The settings write path now validates integer bounds, and the read path defensively clamps direct DB edits.

Test coverage is strong around defaults, per-type independence, DM filtering opt-in, and boundary values.

Verification: `go test ./modules/message -run 'TestDaysCutoff|TestBuildRecentItems_'` passed. The targeted common-package test run could not complete because local MySQL was unavailable (`127.0.0.1:3306: connect: connection refused`).

yujiawei (dismissed):
# Code Review — PR #291 (octo-server)

## Summary

Well-scoped, low-risk change. It moves the recent-tab activity window from a hard-coded 72h constant to per-channel-type `system_settings` keys, with a `0` sentinel that disables filtering. The defaults reproduce the historical behavior for the three channel types the recent tab is documented to carry (group/thread = 3-day window, DM unfiltered), so it is zero-impact until an operator opts in. Code is clean, well-commented, and the test coverage (unit + DB-backed e2e) is thorough.

**Verdict: APPROVED.** No P0/P1 issues. One P2 observation and a couple of nits below.

## Verification

- `go build ./modules/...` — pass
- `go vet ./modules/...` and `go vet -tags integration ./modules/message/...` — pass
- Pure unit tests `buildRecentItems` / `daysCutoff` / per-type window cases — pass
  (DB-backed getter/admin tests require a live MySQL and were not run in this environment; logic was verified by reading.)
- No orphaned `threeDaysAgo` references; every `buildRecentItems` caller updated to the new signature.
- No new import cycle: `modules/common` does not import `modules/message`.
- `<= cutoff` boundary semantics and the DM exemption (person default `0`) are preserved exactly.
- The new `[0, 3650]` write-path bound affects only the three new int settings — there are no pre-existing `settingTypeInt` keys, so no validation regression.

## Findings

### P2 — Unknown channel types are no longer time-filtered (deliberate, but diverges from old behavior)

`api_sidebar.go` — `cutoffFor` returns `0` (no filter) for any channel type that is not group/thread/DM:

```go
func (c recentCutoffs) cutoffFor(channelType uint8) int64 {
	switch channelType {
	case common.ChannelTypeGroup.Uint8():        return c.group
	case common.ChannelTypeCommunityTopic.Uint8(): return c.thread
	case common.ChannelTypePerson.Uint8():       return c.person
	default:                                      return 0   // kept unconditionally
	}
}
```

The previous code applied the 72h window to **all non-DM** types:

```go
isDM := conv.ChannelType == common.ChannelTypePerson.Uint8()
if !isDM && conv.Timestamp <= cutoff { continue }
```

The recent-tab conversation list comes directly from `IMSyncUserConversation` and is only space-filtered (`FilterRawConversationsBySpace`), never channel-type-filtered — so other channel types (`ChannelTypeCustomerService`, `ChannelTypeCommunity`, `ChannelTypeInfo`) *can* reach `buildRecentItems`. For those types, an entry older than 72h was previously dropped and is now kept indefinitely.

The inline comment explicitly justifies this ("defaulting unknown types to 'kept' avoids silently dropping a future channel type"), so it is an intentional design choice — and "keep" is arguably the safer default. The reason this is worth flagging rather than waving through: the PR description claims defaults "reproduce today's exact behaviour," and for these edge types that is not literally true. If any of these types currently surface in the recent tab in production, this is a visible behavior change (stale customer-service/info conversations lingering). Recommendation: either confirm these types never reach the recent tab in practice and note it, or keep the `default` branch on the old behavior (treat unknown types as "thread/group"-equivalent windowed) until a real future type opts in. Not a merge blocker.

### Nits / suggestions (non-blocking)

- **Shared `[0, 3650]` int bound.** `settingIntMin/Max` is process-global across all int settings. Fine today (all int settings are day windows) and the comment already flags the future migration to a per-key bound on `settingDef`. No action needed now — just confirming the limitation is understood.
- **`EnsureSystemSettings` on the hot path.** `loadRecentCutoffs` calls `EnsureSystemSettings` once per recent-tab request, which takes `sharedMu` each call. After first init this is a trivial lock/check/unlock and the snapshot reads are lock-free atomics, so the cost is negligible — noting only for completeness.
- **`getIntClamped` falls back to the default, not the boundary.** An out-of-range DB value (only reachable via direct DB edit) resolves to the code default rather than being clamped to min/max. This matches the documented "defence in depth" intent; just confirming it is the chosen semantics.

## Conclusion

The change is correct, the defaults faithfully preserve behavior for the documented channel types, and the test suite covers defaults, per-type independence, the `0`-disable sentinel, the new DM filtering, write-path validation, and a full admin-write → reload → filter e2e. The single P2 is a documented, defensible design choice. Approving.

lml2468 (approved):
APPROVED (head `d11e2d7a5a49`).

Read the full diff against the surrounding seam (`Sidebar.Sync`, `system_setting` write path, `EnsureSystemSettings`, `octo-lib/common.ChannelType*`) and concur with yujiawei's analysis end-to-end. Defaults reproduce historical behaviour byte-for-byte, the pre-existing int-validation hole is closed in passing, and the e2e covers admin-write → reload → filter. Adding two independent observations beyond yujiawei's review.

## Verified independently

- **Default parity is byte-exact.** `daysCutoff(now, 3)` = `now.Add(-3*24h).Unix()` is the same instant `threeDaysAgo()` produced. Comparison stayed `<= cutoff` (strict-greater keep), DM exemption is now data-driven via `person=0` → `cutoff=0` → "no filter" branch. The only way to regress legacy behaviour is to write a non-default into `system_setting`.
- **Range guard is a real fix, not just hygiene.** The previous int path in `updateSystemSettings` ran only `strconv.Atoi` — negative or absurd day counts (e.g. `-1`, `100000`) would land in DB and reach the snapshot. `[0, 3650]` on the write path + `getIntClamped` defence-in-depth on the read path closes both sides; tests cover `-1`, `3651`, `100000`, `"abc"`, and the `0` / `3650` boundaries explicitly.
- **Setting-key uniqueness checked.** New keys (`sidebar.recent_filter_{group,thread,person}_days`) don't collide with any existing schema row. `system_setting_schema.go` is the registry; no other entry in the `sidebar` category exists yet.
- **`settingIntMax = 3650` is overflow-safe.** `now.Add(-3650 * 24h)` ≈ now − 10 years; far below int64 limits, no time-arithmetic edge.
- **`cutoff == 0` short-circuits before the `Timestamp <= cutoff` test.** Important because `int64(0)` could otherwise accidentally exclude any conversation with `Timestamp <= 0` (e.g. a stub row with unset timestamp). The `cutoff != 0 &&` guard prevents that — confirmed at `api_sidebar.go:975`.
- **Default vs admin-override e2e tests touch the singleton consistently.** `EnsureSystemSettings` returns a process-wide singleton (intentionally — see the deleted-`resetSharedSystemSettingsForTest` comment). Defaults test asserts on a fresh `CleanAllTables`-wiped state; admin-override test calls the HTTP endpoint, which internally invokes `Reload()`. So per-test setup is correct.
- **Comment-block on `buildRecentItems` is precise** about both the rule (`cutoff == 0 OR timestamp > cutoff`) and the deliberate non-rule (`group_unfollowed=1` still appears in recent — PM call from issue #337). That's exactly the kind of non-rule that needs to be in the source comment because it surprises reviewers; nice catch keeping it.

## Concurring on yujiawei's P2 (non-blocking)

The behavioural change for non-{group, thread, person} channel types (CustomerService=2 / Community=4 / Info=6) from "filtered by 72h" to "always kept" is real. yujiawei is right that this is documented and defensible: the `keep-unknown` choice is more robust than "silently drop a future channel type". For OCTO today, CustomerService/Community/Info aren't used as user-facing recent-tab content, so reach is minimal. Worth noting if any of those types are ever lit up downstream, the recent tab will now include them unfiltered until an operator adds a `sidebar.recent_filter_<type>_days` setting (which would also need a code update — the switch is closed).

## Two observations yujiawei didn't surface

### 1. e2e test ordering risk if more SystemSettings-mutating tests land in this package

The shared singleton is intentional (per the explicit doc-comment that removes `resetSharedSystemSettingsForTest`), and the two new e2e tests are correctly written *individually*. But the contract is fragile:

- `TestE2E_RecentFilter_DefaultsReproduceLegacyBehaviour` relies on `CleanAllTables` + the singleton's auto-reload (≤60s) to clear any stale snapshot from prior tests.
- It does not explicitly call `m.systemSettings.Reload()` (because it doesn't hold a Manager handle), so a *prior* test in the same package that writes `sidebar.recent_filter_*` via the admin path and finishes inside the 60s TTL would leave the singleton snapshot polluted. Today no such other test exists in this package, so it passes. The first time someone adds a second `sidebar.*` admin-write test before the defaults test runs alphabetically, the defaults test becomes flaky.

Two low-cost prophylactics for the next contributor (none required for this PR):
- Have e2e tests that mutate snapshot end with an explicit `EnsureSystemSettings(ctx).Reload()` after `CleanAllTables`, so the next test starts from a known-good snapshot.
- Or document this in a `// NOTE: this test reads the shared snapshot — see EnsureSystemSettings; any test that mutates sidebar.* must reload before returning` comment on `loadRecentCutoffs`.

### 2. Per-key `settingDef` migration path is a forward note worth promoting from comment to issue

The comment at `system_setting_schema.go:13–20` correctly anticipates that the second int setting wanting a *different* range (say `0..86400` seconds rather than `0..3650` days) will need to move `settingIntMin/Max` to a per-key `settingDef` field. Right now this lives only as inline documentation. If issue tracker is the source of truth for "design debt to revisit", filing this as a tracked follow-up (rather than relying on the next person grepping the schema file) is worth ~30s. Strictly an organisational nit.

## Checks

Build / Lint / Vet / i18n / sanity / osv all green at review time. Test / CodeQL still running, ride them out then merge.

LGTM.

lml2468 (dismissed):
APPROVED (head `649cb8370ee2`).

Single-commit follow-up that build-tags the e2e file `integration`. No production code change, no test logic change — just CI hygiene. Three reasons it's a clean follow-up:

1. **Matches existing convention.** `modules/message/api_sidebar_integration_test.go` already uses the same `//go:build integration` tag for the same reason (NewTestServer + shared DB + module.Setup migration order-fragility). New e2e file now sits alongside that convention rather than inventing a new pattern.

2. **CI coverage retained.** The per-channel-type filter logic is still covered by `api_sidebar_test.go` (unit tests on `buildRecentItems` / `daysCutoff`) and the `system_settings` read/write path by the `common` package tests, both of which run in the default `Test` job. What's lost from the default run is the *full-stack glue* (admin HTTP → DB → snapshot reload → handler), which is exactly what the `integration` tag is for. Trade-off is correct: glue tests run in the integration lane, not the unit lane.

3. **Comment explains the *why*** — the schema-state panic when `NewTestServer.module.Setup` re-applies a migration on an already-populated shared DB. Future contributor adding another full-stack test in this package will see the pattern and use the tag rather than re-hitting the same panic.

Note: this is the same class of issue as the "e2e test ordering risk" I flagged on R1. The fix (build tag) is more pragmatic than what I suggested (explicit Reload() in test setup) — moving the order-fragile tests entirely out of the default lane sidesteps the whole class of singleton-pollution / migration-replay problems for free. Good call.

CI on this SHA: same green Build/Lint/Vet/i18n/sanity/osv; the default Test job will no longer execute the two `TestE2E_RecentFilter_*` cases — confirm the integration-tagged lane (if there is one in CI) picks them up, or ensure they're at least runnable locally before merge.

LGTM, merge once Test/CodeQL settle.

Jerry-Xin (approved):
The PR is in scope for octo-server and the implementation is generally sound: it wires configurable recent-tab filtering through the existing `system_settings` path, preserves the main group/thread/DM defaults, and adds focused unit coverage.

💬 Non-blocking
- 🟡 Warning: `recentCutoffs.cutoffFor` keeps unknown channel types unconditionally (`modules/message/api_sidebar.go:678`). That is a behavioral change from the old `!isDM` logic, which applied the 72h cutoff to every non-DM type. If WuKongIM can ever return `CustomerService`, `Community`, or `Info` conversations in sidebar sync, old conversations of those types may now appear forever by default. If that is intentional, the comment/test makes it explicit; otherwise, consider defaulting unknown types to the group cutoff.
- 🟡 Warning: the new out-of-range int validation uses raw non-OK `c.JSON` (`modules/common/api_manager_system_setting.go:217`). This matches nearby legacy branches, but it adds another localized-error-envelope bypass. When this endpoint is migrated, this path should use registered `errcode` plus `httperr.ResponseErrorL`.

✅ Highlights
- Good separation between setting resolution (`modules/common/system_settings.go:502`) and sidebar filtering (`modules/message/api_sidebar.go:669`).
- The `0` sentinel is handled consistently through `daysCutoff` and `buildRecentItems`.
- The admin write path now rejects invalid integer ranges, and the read path defensively clamps direct DB edits.
- The unit tests cover defaults, overrides, boundaries, per-channel behavior, and DM filtering.

Verification: `go test ./modules/message -run 'TestDaysCutoff|TestBuildRecentItems'` passed. `go test ./modules/common ...` could not run in this environment because MySQL on `127.0.0.1:3306` was unavailable.

lml2468 (approved):
APPROVED (head `2171adb751f7`).

Pure docs + tests follow-up addressing the prior round's non-blocking observations from yujiawei, Jerry-Xin, and my R1. No production behaviour change — the keep-unknown semantics and per-type cutoff logic are bit-identical to `649cb8370ee2`. All four changes land cleanly.

## ✅ Each follow-up verified

**1. Stale top-of-file data-flow comment updated (Jerry-Xin's first point).**
The old `recent – all DMs; groups/threads with timestamp > now-72h` line was historically accurate but actively misleading after this PR. Replaced with the per-channel-type system_settings description + reference to `buildRecentItems / loadRecentCutoffs`. Now the file-top diagram is no longer lying about behaviour. Good.

**2. `loadRecentCutoffs` doc + NOTE on shared-singleton snapshot (my R1 observation #1).**
Two changes here, both correct:

- Wording tightened from "reproduces the historical hard-coded behaviour exactly" to "for the channel types the recent tab carries, reproduces the historical hard-coded behaviour" + `(Unknown types are kept unconditionally — see cutoffFor.)`. That removes the implicit promise of full parity that the old wording made — yujiawei's P2 / Jerry-Xin's unknown-type point both apply only to unknown types, and the doc now says so.
- New NOTE block: `EnsureSystemSettings returns a process-wide singleton ... Tests that mutate sidebar.* settings must Reload() after wiping the table (see the recent-filter e2e setup), else a stale snapshot from a prior test (within the ~60s auto-reload TTL) can leak in.` This is exactly the prophylactic I suggested — the next contributor adding a sidebar.* mutating test will see the NOTE before they hit the flake.

**3. Explicit unit test for unknown-channel-type keep behaviour (yujiawei P2 + Jerry-Xin nit).**
`TestBuildRecentItems_UnknownChannelType_KeptUnconditionally` sets all three known-type cutoffs to active 3-day windows AND uses an ancient timestamp (10000h ago) AND uses an unknown channel type (`99`) — the only way the test passes is if `cutoffFor`'s `default: return 0` branch fires. Test name + comment ("locks the deliberate `cutoffFor` default = 0") explicitly call out that this is a *guarded design choice*, not just an accidental property. So a future contributor who "tightens" the default to e.g. fall back to the group window will see this test go red with a clear explanation of why the keep-all behaviour was intentional. That's the right shape for an intent-locking test.

Small nit (not blocking): the unknown-type sentinel `99` is a magic number. A `const unknownChannelType uint8 = 99` (already done) plus a one-line comment that it's "any uint8 distinct from group/person/communityTopic, value chosen arbitrarily" would future-proof against `octo-lib` ever adding a `ChannelType` constant numbered 99. Today there are only ChannelTypes 0..6, so plenty of headroom; just future-self friendliness. Definitely don't gate merge on it.

**4. e2e explicitly `Reload()` after `CleanAllTables` (my R1 prophylactic).**
`TestE2E_RecentFilter_DefaultsReproduceLegacyBehaviour` now calls `commonapi.EnsureSystemSettings(ctx).Reload()` right after the table wipe, with a precise comment about why ("the shared SystemSettings singleton may hold sidebar.* rows written by a prior test (within the ~60s auto-reload TTL)"). Defends against exactly the flake scenario I called out — `CleanAllTables` only clears the DB, not the in-memory snapshot. Now "no override" actually means defaults regardless of test ordering.

The other e2e (`TestE2E_RecentFilter_AdminOverrideTakesEffect`) doesn't need the same call because its admin HTTP write path already calls `Reload()` internally — which the prior R1 review verified.

## Combined R1+R2+R3 follow-through scorecard

| Reviewer point | R1/R2 status | R3 status |
|---|---|---|
| Stale top-of-file comment (Jerry-Xin) | flagged | fixed |
| e2e singleton-snapshot risk (lml2468) | flagged as future-flake | fixed via explicit Reload + NOTE |
| Unknown-channel-type keep-all decision (yujiawei P2 / Jerry-Xin) | flagged as documented but untested | locked by test + tightened doc |
| Non-OK JSON responses use raw `jsonH{"msg":...}` (Jerry-Xin) | identified as out-of-scope (whole admin-write path uses this pattern, i18n migration is a separate workflow) | not addressed, **and correctly so** |
| `settingDef` per-key range (lml2468 organisational nit) | suggested to file as issue | not addressed; org-level, no in-PR action needed |

Net: the things that *should* be in this PR are in this PR. The things that belong in separate work (admin-write i18n; future per-key range schema) stayed out, which is correct scope discipline. Author resisted the temptation to expand scope, which is exactly the right call after the previous round's already-clean state.

## CI

No production diff, so the prior Test/CodeQL/Build/Lint/Vet/i18n/sanity/osv results carry over modulo the new unit test. Wait for the green test run on the new SHA then merge.

LGTM. Third pass and the iteration discipline is showing.

Issue #284, opened 2026-06-05
incomingwebhook: harden public push path — Redis-independent rate-limit fallback + hot-path caching


an9xyz:
## Context

`modules/incomingwebhook` implements Discord-style group incoming webhooks. The push endpoint

```
POST /v1/incoming-webhooks/:webhook_id/:token
```

is the **only unauthenticated, publicly reachable endpoint** in the module (no `AuthMiddleware`). The feature is functionally complete and well-hardened on the correctness/concurrency side (token hashing + constant-time compare, soft-delete, space isolation, quota via row-lock). This issue tracks two **resilience/scaling** improvements to the hot push path. Neither is a correctness bug — they are latent operational risks.

A third item (audit-table TTL cleanup) is **intentionally deferred** — see "Deferred" below.

---

## 1. Redis-independent rate-limit fallback on the public push endpoint (priority: now)

**Current behavior** (`modules/incomingwebhook/api.go`, `ratelimit.go`):
- Two limiters protect `push`: per-IP `StrictIPRateLimitMiddleware` (default 30 rps / burst 60) and per-webhook token bucket (default 5 rps / burst 10).
- Both are **fail-open** on any Redis error, and both share a single Redis client built in `sharedRateRedis()` with **`PoolSize: 10`**.

**Why this matters:**
- When Redis hiccups (or the small 10-connection pool is saturated under concurrency), **both** limiters silently disable on the one endpoint that has no auth in front of it.
- The risk is **not** token theft — `webhook_id` is 128-bit and `token` is 256-bit random, so brute-force enumeration is infeasible regardless. The risk is **flood amplification**: each push still performs 2 DB reads + 1 WuKongIM `SendMessageWithResult`, so an unthrottled flood amplifies into DB and the messaging control plane.

**Proposed scope:**
- [ ] Add a **process-local fallback limiter** (e.g. `golang.org/x/time/rate`) on the push path so a per-instance ceiling still applies when Redis is unavailable. Keep the existing Redis limiters as the primary (project convention is fail-open; this adds a floor without changing that contract).
- [ ] Re-evaluate / raise the limiter Redis pool size (currently `PoolSize: 10`, shared by both the per-IP and per-webhook limiters), or split into a dedicated client, sized to expected push concurrency.
[...]

PR #292, opened 2026-06-06
feat(incomingwebhook): runtime master toggle + tunable thresholds via system_setting

an9xyz:
Wires the incoming-webhook module into the admin-tunable `system_setting` layer (`modules/common.SystemSettings`) so the feature can be switched off and its core thresholds adjusted **at runtime** — `Reload` takes effect immediately on the writing instance, peers converge within the snapshot TTL, no restart needed.

This is the config/operability ask raised alongside #284 (separate from #284's caching item, which is its own PR).

## What changed

New `system_setting` category `incomingwebhook` (schema-driven — no table migration):

| key | type | meaning |
|---|---|---|
| `enabled` | bool | master switch |
| `per_webhook_rps` | float | per-webhook token-bucket rate |
| `per_webhook_burst` | int | per-webhook token-bucket burst |
| `max_per_group` | int | per-group webhook quota |

Each resolves **DB → env(`DM_INCOMINGWEBHOOK_*`) → code-default**, so existing env tuning still applies and the admin `GET /v1/manager/common/system_setting` surfaces the true effective value. The env knobs now live solely in `modules/common` as the single source of truth; `incomingwebhook` reads them through the typed getters (`IncomingWebhookEnabled/PerWebhookRPS/PerWebhookBurst/MaxPerGroup`).

### Master toggle off (operator-visible)
- **push** endpoint → **404** (uniform global state; not per-webhook, so no enumeration leak — gated first in the limiter chain, short-circuits before the floor)
- **management writes** (create/update/delete/regenerate) → **403**
- **list** (read) stays available so operators can still inspect existing config

### Supporting changes
- Adds `settingTypeFloat` + `SystemSettings.getFloat` for the rps knob (with matching manager-write validation).
- Two new errcodes `push_disabled` (404) / `mgmt_disabled` (403) with zh-CN translations; en-US markers regenerated via `make i18n-extract`.

## Verification

Run and passing in this environment (no infra needed):
- `TestSystemSettings_IncomingWebhook_GetterChain_NoInfra` — the full snapshot(DB) → env → code-default precedence incl. float rps
- `TestIncomingWebhookRespondHelpers/{pushDisabled,mgmtFeatureDisabled}` — 404/403 + correct i18n code
- `TestIncomingWebhookNoLegacyResponseError`, `pkg/i18n/...`, `go build ./...`, `golangci-lint` (0 issues), `make i18n-extract-check` + `i18n-lint`

Compiled but **not executed locally** (need MySQL/Redis, blocked by this sandbox's network policy — will run in CI):
- `TestSystemSettings_IncomingWebhook*` (DB-backed Load/Reload)
- `TestFeatureToggle_DisabledStopsPushAndMgmtWrites` (end-to-end: DB write `enabled=0` → Reload → push 404 / create 403 / list 200)

https://claude.ai/code/session_01Lq2ejFV7X16HvAmg9BDGTu

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lq2ejFV7X16HvAmg9BDGTu)_

### DIFF
--- a/modules/common/api_manager_system_setting.go
+++ b/modules/common/api_manager_system_setting.go
@@ -221,6 +221,15 @@ func (m *Manager) updateSystemSettings(c *wkhttp.Context) {
 					return
 				}
 			}
+		case settingTypeFloat:
+			if item.Value != "" {
+				if _, err := strconv.ParseFloat(item.Value, 64); err != nil {
+					c.JSON(http.StatusBadRequest, jsonH{
+						"msg": fmt.Sprintf("%s.%s 必须是数字", item.Category, item.Key),
+					})
+					return
+				}
+			}
 		case settingTypeEncrypted:
 			if item.Value == "" || item.Value == secretMask {
 				// Empty payload or the GET mask sentinel preserves the existing
--- a/modules/common/api_manager_system_setting.go
+++ b/modules/common/api_manager_system_setting.go
@@ -3,6 +3,7 @@ package common
 import (
 	"errors"
 	"fmt"
+	"math"
 	"net/http"
 	"strconv"
 
@@ -203,7 +204,9 @@ func (m *Manager) updateSystemSettings(c *wkhttp.Context) {
 		case settingTypeInt:
 			// Empty value means "reset to default" (the getter treats an empty
 			// snapshot entry as not-configured), so only validate non-empty
-			// payloads. Bounds-check against [settingIntMin, settingIntMax] —
+			// payloads. A Positive key (rate-limit / quota knob) must be a
+			// strictly-positive int and skips the shared day-window bound;
+			// otherwise bounds-check against [settingIntMin, settingIntMax] —
 			// previously this path only ran strconv.Atoi with no range guard,
 			// letting absurd windows (or negatives) reach the DB (issue #289).
 			if item.Value != "" {
@@ -214,7 +217,14 @@ func (m *Manager) updateSystemSettings(c *wkhttp.Context) {
 					})
 					return
 				}
-				if n < settingIntMin || n > settingIntMax {
+				if def.Positive {
+					if n <= 0 {
+						c.JSON(http.StatusBadRequest, jsonH{
+							"msg": fmt.Sprintf("%s.%s 必须是正整数", item.Category, item.Key),
+						})
+						return
+					}
+				} else if n < settingIntMin || n > settingIntMax {
 					c.JSON(http.StatusBadRequest, jsonH{
 						"msg": fmt.Sprintf("%s.%s 必须在 [%d, %d] 范围内", item.Category, item.Key, settingIntMin, settingIntMax),
 					})
@@ -222,10 +232,28 @@ func (m *Manager) updateSystemSettings(c *wkhttp.Context) {
 				}
 			}
 		case settingTypeFloat:
+			// Reject non-finite (NaN / ±Inf — all of which strconv.ParseFloat
+			// happily accepts) so they can never reach a rate limiter: a NaN rps
+			// errors the Redis Lua script (fail-open) and ±Inf/≤0 short-circuit
+			// the bucket to "always allowed". A Positive key additionally rejects
+			// ≤0. Read side mirrors this (clamp → default) for direct DB edits.
 			if item.Value != "" {
-				if _, err := strconv.ParseFloat(item.Value, 64); err != nil {
+				f, err := strconv.ParseFloat(item.Value, 64)
+				if err != nil {
 					c.JSON(http.StatusBadRequest, jsonH{
 						"msg": fmt.Sprintf("%s.%s 必须是数字", item.Category, item.Key),
+					})
+					return
+				}
+				if math.IsNaN(f) || math.IsInf(f, 0) {
+					c.JSON(http.StatusBadRequest, jsonH{
+						"msg": fmt.Sprintf("%s.%s 必须是有限数字", item.Category, item.Key),
+					})
+					return
+				}
+				if def.Positive && f <= 0 {
+					c.JSON(http.StatusBadRequest, jsonH{
+						"msg": fmt.Sprintf("%s.%s 必须是正数", item.Category, item.Key),
 					})
 					return
 				}
--- a/modules/common/api_manager_system_setting_test.go
+++ b/modules/common/api_manager_system_setting_test.go
@@ -433,3 +433,68 @@ func TestManagerSystemSetting_UpdateAcceptsInRangeIntBoundaries(t *testing.T) {
 	assert.Equal(t, "3650", got["sidebar.recent_filter_thread_days"])
 	assert.Equal(t, "7", got["sidebar.recent_filter_person_days"])
 }
+
+// TestManagerSystemSetting_UpdateRejectsInvalidIncomingWebhookNumerics pins the
+// #292-review hardening: the rate-limit / quota knobs are Positive keys, so the
+// write path must reject 0 / negative / NaN / ±Inf / non-numeric — values that
+// would otherwise silently disable the per-webhook limiter or the quota.
+func TestManagerSystemSetting_UpdateRejectsInvalidIncomingWebhookNumerics(t *testing.T) {
+	t.Setenv(masterKeyEnv, "0123456789abcdef0123456789abcdef")
+	s, ctx := testutil.NewTestServer()
+	require.NoError(t, testutil.CleanAllTables(ctx))
+	require.NoError(t, ctx.Cache().Set(
+		ctx.GetConfig().Cache.TokenCachePrefix+testutil.Token,
+		testutil.UID+"@test@"+string(wkhttp.SuperAdmin),
+	))
+
+	cases := []struct{ key, value string }{
+		{"per_webhook_rps", "0"}, {"per_webhook_rps", "-1.5"}, {"per_webhook_rps", "NaN"},
+		{"per_webhook_rps", "+Inf"}, {"per_webhook_rps", "-Inf"}, {"per_webhook_rps", "abc"},
+		{"per_webhook_burst", "0"}, {"per_webhook_burst", "-1"},
+		{"max_per_group", "0"}, {"max_per_group", "-5"},
+	}
+	for _, tc := range cases {
+		body := []byte(`{"items":[{"category":"incomingwebhook","key":"` + tc.key + `","value":"` + tc.value + `"}]}`)
+		w := httptest.NewRecorder()
+		req, _ := http.NewRequest("POST", "/v1/manager/common/system_setting", bytes.NewReader(body))
+		req.Header.Set("token", testutil.Token)
+		s.GetRoute().ServeHTTP(w, req)
+		assert.NotEqualf(t, http.StatusOK, w.Code, "incomingwebhook.%s=%q must be rejected", tc.key, tc.value)
+	}
+}
+
+// TestManagerSystemSetting_UpdateAcceptsPositiveIncomingWebhookNumerics pins that
+// valid positive values are accepted, including a large burst — Positive keys opt
+// out of the shared [0,3650] day-window cap, so there is no artificial upper bound
+// (matching the env semantics they fall back to).
+func TestManagerSystemSetting_UpdateAcceptsPositiveIncomingWebhookNumerics(t *testing.T) {
+	t.Setenv(masterKeyEnv, "0123456789abcdef0123456789abcdef")
+	s, ctx := testutil.NewTestServer()
+	require.NoError(t, testutil.CleanAllTables(ctx))
+	require.NoError(t, ctx.Cache().Set(
+		ctx.GetConfig().Cache.TokenCachePrefix+testutil.Token,
+		testutil.UID+"@test@"+string(wkhttp.SuperAdmin),
+	))
+
+	body := []byte(`{"items":[` +
+		`{"category":"incomingwebhook","key":"per_webhook_rps","value":"8.5"},` +
+		`{"category":"incomingwebhook","key":"per_webhook_burst","value":"100000"},` +
+		`{"category":"incomingwebhook","key":"max_per_group","value":"50"}` +
+		`]}`)
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("POST", "/v1/manager/common/system_setting", bytes.NewReader(body))
+	req.Header.Set("token", testutil.Token)
+	s.GetRoute().ServeHTTP(w, req)
+	require.Equal(t, http.StatusOK, w.Code, w.Body.String())
+
+	db := newSystemSettingDB(ctx)
+	rows, err := db.listAll()
+	require.NoError(t, err)
+	got := map[string]string{}
+	for _, r := range rows {
+		got[schemaKey(r.Category, r.KeyName)] = r.Value
+	}
+	assert.Equal(t, "8.5", got["incomingwebhook.per_webhook_rps"])
+	assert.Equal(t, "100000", got["incomingwebhook.per_webhook_burst"], "no artificial upper bound for Positive keys")
+	assert.Equal(t, "50", got["incomingwebhook.max_per_group"])
+}
--- a/modules/common/system_setting_schema.go
+++ b/modules/common/system_setting_schema.go
@@ -7,6 +7,7 @@ const (
 	settingTypeString    = "string"
 	settingTypeBool      = "bool"
 	settingTypeInt       = "int"
+	settingTypeFloat     = "float"
 	settingTypeEncrypted = "encrypted"
 )
 
@@ -93,6 +94,18 @@ var systemSettingSchema = []settingDef{
 	{Category: "sidebar", Key: "recent_filter_person_days", Type: settingTypeInt, Description: "最近会话-单聊(DM)活跃过滤窗口(天)，0=不过滤(默认)",
 		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.SidebarRecentFilterPersonDays()) }},
 
+	// Incoming webhook 总开关 + 核心阈值 — 与 env(DM_INCOMINGWEBHOOK_*) 等价，DB 为
+	// 单一真源。enabled 关闭后 push 返回 404、管理写操作被拒、仅保留 list 只读；
+	// 其余三项实时调阈值无需重启（SystemSettings 快照 60s 内多实例收敛）。
+	{Category: "incomingwebhook", Key: "enabled", Type: settingTypeBool, Description: "是否开启群入站 Webhook（关闭后停止推送与管理写操作）",
+		Effective: func(s *SystemSettings) string { return boolToCanonical(s.IncomingWebhookEnabled()) }},
+	{Category: "incomingwebhook", Key: "per_webhook_rps", Type: settingTypeFloat, Description: "单个 Webhook 每秒推送速率上限（令牌桶 rps）",
+		Effective: func(s *SystemSettings) string { return floatToCanonical(s.IncomingWebhookPerWebhookRPS()) }},
+	{Category: "incomingwebhook", Key: "per_webhook_burst", Type: settingTypeInt, Description: "单个 Webhook 推送突发上限（令牌桶 burst）",
+		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.IncomingWebhookPerWebhookBurst()) }},
+	{Category: "incomingwebhook", Key: "max_per_group", Type: settingTypeInt, Description: "单个群最多可创建的 Webhook 数量",
+		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.IncomingWebhookMaxPerGroup()) }},
+
 	// Email server config — formerly yaml-only (Support.* in config.go).
 	{Category: "support", Key: "email", Type: settingTypeString, Description: "技术支持邮箱（发件人）",
 		Effective: func(s *SystemSettings) string { return s.SupportEmail() }},
@@ -110,6 +123,12 @@ func boolToCanonical(v bool) string {
 		return "1"
 	}
 	return "0"
+}
+
+// floatToCanonical 把 float 规范成最短十进制表示（5.0 → "5"，0.5 → "0.5"），与
+// settingTypeFloat 的 DB 存储 / POST 入参拼写保持一致，供 GET effective_value 用。
+func floatToCanonical(v float64) string {
+	return strconv.FormatFloat(v, 'g', -1, 64)
 }
 
 // schemaKey returns the canonical "category.key" string used as map key in
--- a/modules/common/system_setting_schema.go
+++ b/modules/common/system_setting_schema.go
@@ -55,6 +55,16 @@ type settingDef struct {
 	// API layer is responsible for masking before serialisation; never
 	// surface this value directly.
 	Effective func(*SystemSettings) string
+	// Positive, when set on a settingTypeInt / settingTypeFloat key, requires a
+	// strictly-positive finite value on the admin write path and OPTS OUT of the
+	// shared [settingIntMin, settingIntMax] bound (which exists for the
+	// day-window int settings where 0 is a valid "disable" sentinel). Used by
+	// rate-limit / quota knobs (incomingwebhook.*) where 0 / negative / NaN / Inf
+	// would silently disable the control — the schema comment on settingIntMin
+	// anticipated this per-key override. No artificial upper bound is imposed
+	// (matches the env semantics these keys fall back to). Read-side defence is
+	// in the typed getters (clamp ≤0 / non-finite → default).
+	Positive bool
 }
 
 // systemSettingSchema enumerates every admin-tunable setting backed by the
@@ -99,11 +109,11 @@ var systemSettingSchema = []settingDef{
 	// 其余三项实时调阈值无需重启（SystemSettings 快照 60s 内多实例收敛）。
 	{Category: "incomingwebhook", Key: "enabled", Type: settingTypeBool, Description: "是否开启群入站 Webhook（关闭后停止推送与管理写操作）",
 		Effective: func(s *SystemSettings) string { return boolToCanonical(s.IncomingWebhookEnabled()) }},
-	{Category: "incomingwebhook", Key: "per_webhook_rps", Type: settingTypeFloat, Description: "单个 Webhook 每秒推送速率上限（令牌桶 rps）",
+	{Category: "incomingwebhook", Key: "per_webhook_rps", Type: settingTypeFloat, Description: "单个 Webhook 每秒推送速率上限（令牌桶 rps）", Positive: true,
 		Effective: func(s *SystemSettings) string { return floatToCanonical(s.IncomingWebhookPerWebhookRPS()) }},
-	{Category: "incomingwebhook", Key: "per_webhook_burst", Type: settingTypeInt, Description: "单个 Webhook 推送突发上限（令牌桶 burst）",
+	{Category: "incomingwebhook", Key: "per_webhook_burst", Type: settingTypeInt, Description: "单个 Webhook 推送突发上限（令牌桶 burst）", Positive: true,
 		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.IncomingWebhookPerWebhookBurst()) }},
-	{Category: "incomingwebhook", Key: "max_per_group", Type: settingTypeInt, Description: "单个群最多可创建的 Webhook 数量",
+	{Category: "incomingwebhook", Key: "max_per_group", Type: settingTypeInt, Description: "单个群最多可创建的 Webhook 数量", Positive: true,
 		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.IncomingWebhookMaxPerGroup()) }},
 
 	// Email server config — formerly yaml-only (Support.* in config.go).
--- a/modules/common/system_settings.go
+++ b/modules/common/system_settings.go
@@ -581,7 +581,14 @@ func (s *SystemSettings) IncomingWebhookEnabled() bool {
 // 让 Redis Lua 脚本报错而 fail-open——正是这个 getter 要兜住的。写侧也已拒绝
 // （settingTypeFloat + Positive，见 api_manager_system_setting.go），此处是纵深防御。
 func (s *SystemSettings) IncomingWebhookPerWebhookRPS() float64 {
+	// env fallback 同样消毒：wkhttp.ParseRPSFromEnv 用 strconv.ParseFloat，会接受
+	// NaN / +Inf（DM_INCOMINGWEBHOOK_RPS=NaN 原样透出），所以 def 本身可能非有限。
+	// 若 env 给出非有限/≤0 的 def，回退到永远合法的 code default，避免它穿过下面的
+	// clamp 继续把 NaN 喂给限流器（Jerry-Xin #292 review）。
 	def := wkhttp.ParseRPSFromEnv(envIncomingWebhookPerWebhookRPS, defaultIncomingWebhookPerWebhookRPS)
+	if math.IsNaN(def) || math.IsInf(def, 0) || def <= 0 {
+		def = defaultIncomingWebhookPerWebhookRPS
+	}
 	v := s.getFloat("incomingwebhook", "per_webhook_rps", def)
 	if math.IsNaN(v) || math.IsInf(v, 0) || v <= 0 {
 		return def
--- a/modules/common/system_settings.go
+++ b/modules/common/system_settings.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/Mininglamp-OSS/octo-lib/config"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
 	"go.uber.org/zap"
 )
 
@@ -232,6 +233,18 @@ func (s *SystemSettings) getIntClamped(category, key string, fallback int) int {
 		return fallback
 	}
 	return v
+}
+
+func (s *SystemSettings) getFloat(category, key string, fallback float64) float64 {
+	v, ok := s.lookup(category, key)
+	if !ok {
+		return fallback
+	}
+	parsed, err := strconv.ParseFloat(v, 64)
+	if err != nil {
+		return fallback
+	}
+	return parsed
 }
 
 func (s *SystemSettings) getEncrypted(category, key string, fallback string) string {
@@ -532,4 +545,75 @@ func (s *SystemSettings) SupportEmailSmtp() string {
 // this getter returns the yaml fallback.
 func (s *SystemSettings) SupportEmailPwd() string {
 	return s.getEncrypted("support", "email_pwd", s.ctx.GetConfig().Support.EmailPwd)
+}
+
+// ----- incomingwebhook settings (总开关 + 核心阈值) -----
+//
+// 这些 env 名 / 默认值是 modules/incomingwebhook 的「单一真源」：incomingwebhook 侧
+// 通过下面的 getter 读取（不再各自读 env），从而让 system_setting 的 effective_value
+// 能反映完整的 DB → env → code-default 回退链。修改 env 名或默认值时，需同步
+// systemSettingSchema 的 incomingwebhook 行；reciprocal sync 注释见
+// modules/incomingwebhook/api.go 的 New / allowPerWebhook / create。
+const (
+	envIncomingWebhookEnabled         = "DM_INCOMINGWEBHOOK_ENABLED"
+	envIncomingWebhookPerWebhookRPS   = "DM_INCOMINGWEBHOOK_RPS"
+	envIncomingWebhookPerWebhookBurst = "DM_INCOMINGWEBHOOK_BURST"
+	envIncomingWebhookMaxPerGroup     = "DM_INCOMINGWEBHOOK_MAX_PER_GROUP"
+
+	defaultIncomingWebhookEnabled         = true
+	defaultIncomingWebhookPerWebhookRPS   = 5.0
+	defaultIncomingWebhookPerWebhookBurst = 10
+	defaultIncomingWebhookMaxPerGroup     = 10
+)
+
+// IncomingWebhookEnabled 是群入站 Webhook 功能的总开关。关闭后 push 端点返回 404、
+// 管理写操作（create/update/delete/regenerate）被拒绝，仅保留 list 只读。
+// 回退链：DB → env(DM_INCOMINGWEBHOOK_ENABLED) → 默认开启(true)。
+func (s *SystemSettings) IncomingWebhookEnabled() bool {
+	return s.getBool("incomingwebhook", "enabled", incomingWebhookEnabledEnvDefault())
+}
+
+// IncomingWebhookPerWebhookRPS 单个 webhook 令牌桶速率(rps)。DB → env → 默认 5。
+func (s *SystemSettings) IncomingWebhookPerWebhookRPS() float64 {
+	return s.getFloat("incomingwebhook", "per_webhook_rps",
+		wkhttp.ParseRPSFromEnv(envIncomingWebhookPerWebhookRPS, defaultIncomingWebhookPerWebhookRPS))
+}
+
+// IncomingWebhookPerWebhookBurst 单个 webhook 令牌桶突发上限。DB → env → 默认 10。
+func (s *SystemSettings) IncomingWebhookPerWebhookBurst() int {
+	return s.getInt("incomingwebhook", "per_webhook_burst",
+		wkhttp.ParseBurstFromEnv(envIncomingWebhookPerWebhookBurst, defaultIncomingWebhookPerWebhookBurst))
+}
+
+// IncomingWebhookMaxPerGroup 单个群可创建的 webhook 数量上限。DB → env → 默认 10。
+func (s *SystemSettings) IncomingWebhookMaxPerGroup() int {
+	return s.getInt("incomingwebhook", "max_per_group", incomingWebhookMaxPerGroupEnvDefault())
+}
+
+// incomingWebhookEnabledEnvDefault 解析 DM_INCOMINGWEBHOOK_ENABLED（缺省/无法识别
+// 视为开启），作为 DB 未配置时的 fallback。比 getBool 的 DB 解析更宽松，接受
+// 1/0/true/false/yes/no/on/off（大小写不敏感、允许前后空格）。
+func incomingWebhookEnabledEnvDefault() bool {
+	v := strings.TrimSpace(os.Getenv(envIncomingWebhookEnabled))
+	if v == "" {
+		return defaultIncomingWebhookEnabled
+	}
+	switch strings.ToLower(v) {
+	case "0", "false", "no", "off":
+		return false
+	case "1", "true", "yes", "on":
+		return true
+	}
+	return defaultIncomingWebhookEnabled
+}
+
+// incomingWebhookMaxPerGroupEnvDefault 解析 DM_INCOMINGWEBHOOK_MAX_PER_GROUP；仅
+// 接受正整数，否则回退默认值（语义与迁移前 modules/incomingwebhook.maxPerGroup 一致）。
+func incomingWebhookMaxPerGroupEnvDefault() int {
+	if v := os.Getenv(envIncomingWebhookMaxPerGroup); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n > 0 {
+			return n
+		}
+	}
+	return defaultIncomingWebhookMaxPerGroup
 }
--- a/modules/common/system_settings.go
+++ b/modules/common/system_settings.go
@@ -3,6 +3,7 @@ package common
 import (
 	"context"
 	"encoding/base64"
+	"math"
 	"os"
 	"regexp"
 	"strconv"
@@ -574,20 +575,41 @@ func (s *SystemSettings) IncomingWebhookEnabled() bool {
 }
 
 // IncomingWebhookPerWebhookRPS 单个 webhook 令牌桶速率(rps)。DB → env → 默认 5。
+//
+// 读侧防御（D-289 同型，覆盖直接改库的旁路）：rps 必须是正有限值；NaN/±Inf/≤0 一律
+// 回退到 env/默认。否则 allowPerWebhook 的 `rps<=0` 短路会把限流器静默关掉，NaN 还会
+// 让 Redis Lua 脚本报错而 fail-open——正是这个 getter 要兜住的。写侧也已拒绝
+// （settingTypeFloat + Positive，见 api_manager_system_setting.go），此处是纵深防御。
 func (s *SystemSettings) IncomingWebhookPerWebhookRPS() float64 {
-	return s.getFloat("incomingwebhook", "per_webhook_rps",
-		wkhttp.ParseRPSFromEnv(envIncomingWebhookPerWebhookRPS, defaultIncomingWebhookPerWebhookRPS))
+	def := wkhttp.ParseRPSFromEnv(envIncomingWebhookPerWebhookRPS, defaultIncomingWebhookPerWebhookRPS)
+	v := s.getFloat("incomingwebhook", "per_webhook_rps", def)
+	if math.IsNaN(v) || math.IsInf(v, 0) || v <= 0 {
+		return def
+	}
+	return v
 }
 
 // IncomingWebhookPerWebhookBurst 单个 webhook 令牌桶突发上限。DB → env → 默认 10。
+// 读侧防御：≤0 回退默认（同 RPS，避免 `burst<=0` 短路静默关掉限流器）。
 func (s *SystemSettings) IncomingWebhookPerWebhookBurst() int {
-	return s.getInt("incomingwebhook", "per_webhook_burst",
-		wkhttp.ParseBurstFromEnv(envIncomingWebhookPerWebhookBurst, defaultIncomingWebhookPerWebhookBurst))
+	def := wkhttp.ParseBurstFromEnv(envIncomingWebhookPerWebhookBurst, defaultIncomingWebhookPerWebhookBurst)
+	v := s.getInt("incomingwebhook", "per_webhook_burst", def)
+	if v <= 0 {
+		return def
+	}
+	return v
 }
 
 // IncomingWebhookMaxPerGroup 单个群可创建的 webhook 数量上限。DB → env → 默认 10。
+// 读侧防御：≤0 回退默认（max_per_group=0 会让每次 create 都 ErrQuotaExceeded，是
+// 总开关之外一种更难诊断的「暗关」）。
 func (s *SystemSettings) IncomingWebhookMaxPerGroup() int {
-	return s.getInt("incomingwebhook", "max_per_group", incomingWebhookMaxPerGroupEnvDefault())
+	def := incomingWebhookMaxPerGroupEnvDefault()
+	v := s.getInt("incomingwebhook", "max_per_group", def)
+	if v <= 0 {
+		return def
+	}
+	return v
 }
 
 // incomingWebhookEnabledEnvDefault 解析 DM_INCOMINGWEBHOOK_ENABLED（缺省/无法识别
--- a/modules/common/system_settings_test.go
+++ b/modules/common/system_settings_test.go
@@ -195,6 +195,17 @@ func TestSystemSettings_IncomingWebhook_ReadSideClamp_NoInfra(t *testing.T) {
 		assert.Equalf(t, defaultIncomingWebhookMaxPerGroup, s.IncomingWebhookMaxPerGroup(), "max_per_group=%q must clamp to default", bad)
 	}
 
+	// env-derived fallback is sanitized too: DM_INCOMINGWEBHOOK_RPS=NaN (which
+	// ParseRPSFromEnv passes through) with no DB row must NOT reach the getter as
+	// NaN — it falls back to the code default (Jerry-Xin #292 review).
+	t.Setenv(envIncomingWebhookPerWebhookRPS, "NaN")
+	emptySnap := &SystemSettings{}
+	em := map[string]string{}
+	emptySnap.snapshot.Store(&em)
+	assert.Equal(t, defaultIncomingWebhookPerWebhookRPS, emptySnap.IncomingWebhookPerWebhookRPS(),
+		"env=NaN with no DB row must fall back to the code default, not NaN")
+	t.Setenv(envIncomingWebhookPerWebhookRPS, "")
+
 	// A valid positive value is served as-is (clamp only catches the bad cases).
 	s := &SystemSettings{}
 	snap := map[string]string{
--- a/modules/common/system_settings_test.go
+++ b/modules/common/system_settings_test.go
@@ -172,6 +172,49 @@ func TestSystemSettings_IncomingWebhookRPS_EnvFallbackWhenDBUnset(t *testing.T) 
 	assert.Equal(t, 7.0, s.IncomingWebhookPerWebhookRPS(), "DB 未配置 → env 生效")
 }
 
+// TestSystemSettings_IncomingWebhook_GetterChain_NoInfra exercises the full
+// snapshot(DB) → env → code-default fallback for the incomingwebhook getters
+// WITHOUT a test server: it drives the snapshot map directly. This lets the
+// core getter logic be verified even where MySQL is unavailable; the DB-backed
+// tests above additionally cover the real Load/Reload path in CI.
+func TestSystemSettings_IncomingWebhook_GetterChain_NoInfra(t *testing.T) {
+	// 1) 空快照 + 无 env → code default。
+	t.Setenv(envIncomingWebhookEnabled, "")
+	t.Setenv(envIncomingWebhookPerWebhookRPS, "")
+	t.Setenv(envIncomingWebhookPerWebhookBurst, "")
+	t.Setenv(envIncomingWebhookMaxPerGroup, "")
+	s := &SystemSettings{}
+	empty := map[string]string{}
+	s.snapshot.Store(&empty)
+	assert.True(t, s.IncomingWebhookEnabled())
+	assert.Equal(t, defaultIncomingWebhookPerWebhookRPS, s.IncomingWebhookPerWebhookRPS())
+	assert.Equal(t, defaultIncomingWebhookPerWebhookBurst, s.IncomingWebhookPerWebhookBurst())
+	assert.Equal(t, defaultIncomingWebhookMaxPerGroup, s.IncomingWebhookMaxPerGroup())
+
+	// 2) env override（snapshot 仍空）。
+	t.Setenv(envIncomingWebhookEnabled, "off")
+	t.Setenv(envIncomingWebhookPerWebhookRPS, "7")
+	t.Setenv(envIncomingWebhookPerWebhookBurst, "3")
+	t.Setenv(envIncomingWebhookMaxPerGroup, "2")
+	assert.False(t, s.IncomingWebhookEnabled())
+	assert.Equal(t, 7.0, s.IncomingWebhookPerWebhookRPS())
+	assert.Equal(t, 3, s.IncomingWebhookPerWebhookBurst())
+	assert.Equal(t, 2, s.IncomingWebhookMaxPerGroup())
+
+	// 3) snapshot(DB) 压制 env。
+	snap := map[string]string{
+		"incomingwebhook.enabled":           "1",
+		"incomingwebhook.per_webhook_rps":   "8.5",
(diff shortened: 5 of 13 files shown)


lml2468 (changes requested):
CHANGES_REQUESTED (head `135dd5936ef6`, base `main`).

Design is sound — moving the master toggle + key thresholds onto `system_setting` with DB → env → code-default precedence is the right operability story, and the toggle-off semantics (push → 404, mgmt writes → 403, list stays 200) are well thought out. But there are two blockers from a merge-conflict slip that **regress a fix already on `main` (PR #291)** and one new gap the PR introduces around the new float type. Plus the branch is currently CONFLICTING and must be rebased before review of the final state.

## 🔴 Blocker 1 — `[0, 3650]` int range guard was deleted from the manager write path

On `main`, `modules/common/api_manager_system_setting.go:202-223` reads (verified `gh api .../contents/...?ref=main`):

```go
case settingTypeInt:
    // Empty value means "reset to default" (...).
    // Bounds-check against [settingIntMin, settingIntMax] —
    // previously this path only ran strconv.Atoi with no range guard,
    // letting absurd windows (or negatives) reach the DB (issue #289).
    if item.Value != "" {
        n, err := strconv.Atoi(item.Value)
        if err != nil { ... return }
        if n < settingIntMin || n > settingIntMax { ... return }
    }
```

On this PR's head, the same case collapses back to just `strconv.Atoi`:

```go
case settingTypeInt:
    if item.Value != "" {
        if _, err := strconv.Atoi(item.Value); err != nil { ... return }
    }
```

i.e. the `n < settingIntMin || n > settingIntMax` branch — landed by #291 and explicitly documented as "closes a pre-existing gap" — is gone. The two regression tests are also gone:

- `TestManagerSystemSetting_UpdateRejectsOutOfRangeInt` (the `-1` / `3651` / `100000` case)
- `TestManagerSystemSetting_UpdateAcceptsInRangeIntBoundaries` (the `0` / `3650` boundary case)

Both are still on `main` (verified `gh api .../api_manager_system_setting_test.go?ref=main`), not on `135dd5936ef6`. This is almost certainly a merge-conflict slip on rebase against `main` (`mergeable: CONFLICTING` confirms there were unresolved conflicts in `system_settings*.go`).

The blast radius isn't only the `sidebar.recent_filter_*_days` keys #291 protected — it also opens a hole in the *new* keys this PR introduces:

- `incomingwebhook.max_per_group` can now be set to `0`, `-1`, or `999999999`. `max_per_group = 0` makes every `create` fail with `ErrQuotaExceeded` while the feature looks "enabled" — a silent dark-mode of the master toggle that's harder to diagnose than the explicit 403 path.
- `incomingwebhook.per_webhook_burst = -1` ⇒ `if rps <= 0 || burst <= 0 { return true, nil }` in `allowPerWebhook`. The token bucket short-circuits to "always allowed" — silently disables per-webhook rate limiting. This is the inverse of the float-min issue below, and just as silent.

**Fix:** restore the range branch + the two deleted tests by taking `main`'s version of that block. If a real conflict existed with the new `settingTypeFloat` case, the right resolve is "additive" — keep `main`'s int range guard, append the float case after it.

## 🔴 Blocker 2 — `settingTypeFloat` write path has no range guard at all

Even with blocker 1 fixed, the new float type only validates parseability:

```go
case settingTypeFloat:
    if item.Value != "" {
        if _, err := strconv.ParseFloat(item.Value, 64); err != nil { ... return }
    }
```

An admin write of `per_webhook_rps = -1.0` (or `0`, or `NaN` via `strconv.ParseFloat("NaN", 64)`) lands in DB, `IncomingWebhookPerWebhookRPS()` returns it, and `allowPerWebhook`'s `if rps <= 0 || burst <= 0 { return true, nil }` short-circuits to "always allowed" — silently disabling per-webhook rate limiting. There's no read-side `getFloatClamped` either, so even a direct DB edit lands. The exact same defence-in-depth #291 added for int (write reject + clamp on read) is missing for float.

**Fix:** add a `[settingFloatMin, settingFloatMax]` (something like `[0.0, 10000.0]`) check in the write path, mirror it with `getFloatClamped` on the read side, and add tests for `"-1.0"` / `"NaN"` / `"+Inf"` / `"abc"` rejection and boundary acceptance. The shape is literally the int path, parameterised.

If you don't want to commit to a global float range yet (since rps is currently the only float setting), per-key validation via the `settingDef` struct is the alternative the int comment at `system_setting_schema.go` already anticipates ("Adding an int setting that needs a different range should move this to a per-key field on settingDef"). Either is fine.

## 🟡 P2 — Mergeable: CONFLICTING; rebase before final review

`gh pr view 292 --json mergeable` → `CONFLICTING`. The blocker-1 regression strongly suggests the rebase against `main` is partially done or stale. Please rebase on current `main` (which now carries #291's int range guard and #291's three e2e patterns), re-resolve, and push. Without a clean rebase I can't be sure the post-merge state of the int validation block is what's intended.

## 🟡 P2 — `requirePushEnabled` short-circuit choice is correct but worth documenting timing-safety more loudly

`requirePushEnabled` is mounted *before* `localFloorMiddleware` / `ipLimit` / `ipFailureGateMiddleware`, so a globally-disabled `enabled=0` makes the push endpoint return 404 in O(1) without consulting Redis or the floor map. That's correct for "feature disabled" semantics — the response is global (every request, every URL, every token), so there's nothing to leak.

But: the existing push path goes to great lengths to make all auth-failure responses *uniform* (the comment at `pkg/errcode/incomingwebhook.go:11-22` is explicit about this — the `push_unauthorized` code with constant timing-best-effort exists precisely to keep `no such webhook` / `wrong token` / `dead group` indistinguishable). The new `push_disabled` 404 has a different body than `push_unauthorized` 401, so the *response body and status* distinguish "feature globally disabled" from any other failure mode. That's intentional and fine — globally-disabled is a public state, not a per-target one, so distinguishing it doesn't help an attacker. Worth a one-line comment to that effect right above `requirePushEnabled` so the next person doesn't try to "fix" the inconsistency by making `push_disabled` mimic `push_unauthorized`.

## ✅ What's correct and well-thought-out

- **`settingTypeFloat` + `getFloat` + `floatToCanonical`** — minimal addition, uses `strconv.FormatFloat(v, 'g', -1, 64)` which is the canonical shortest-round-trip form. Cross-checked: 5.0 → "5", 0.5 → "0.5", 1.5 → "1.5" — matches the schema comment exactly.
- **Master toggle semantics differ by call class.**
  - push: 404 globally (no per-webhook leak, no DB hit, gated first → O(1) reject)
  - mgmt-write: 403 after AuthMiddleware (preserves "is this a real admin?" answer before exposing "feature off")
  - list: stays 200 (operators can audit existing config during incident response)
  This is the right shape; the comment on `Route()` spelling it out helps a lot.
- **Single source of truth for env knobs**: `envIncomingWebhook*` constants moved to `modules/common`; `incomingwebhook` reads only through typed getters. The admin UI's effective value GET will now match what the rate limiter actually uses, which is the whole point of the `Effective` callback in `settingDef`.
- **errcode definitions** for `push_disabled` (404) / `mgmt_disabled` (403) carry full doc-comments explaining *why* the response shape doesn't leak. en-US markers + zh-CN translations both updated, i18n-extract-check would pass.
- **`allowPerWebhook` reads from settings every call.** Concern: hot-path snapshot read on every push. But `lookup` is an `RLock` on a `map[string]string` — sub-microsecond, and the alternative (cache the value with TTL invalidation) would re-introduce a "config changed but isn't live" class of bug exactly opposite to what this PR is trying to fix. Reading every call is the right trade.
- **`recordSuccess` audit-semaphore** comment (preserved from prior work) clearly documents *why* "drop > sync fall-back" is the right choice under fail-open Redis. Not new in this PR, but the comment stayed precise through the refactor.

## What I did NOT verify

- The DB-backed `TestSystemSettings_IncomingWebhook*` and `TestFeatureToggle_DisabledStopsPushAndMgmtWrites` end-to-end tests (PR description says "compiled but not executed locally — need MySQL/Redis, blocked by sandbox; will run in CI"). CI didn't run them either at review time because the build/test jobs haven't fired on this SHA (probably waiting on conflict resolution). Confirm green on the rebased SHA.

---

## Verdict

CHANGES_REQUESTED on:

1. (🔴) Restore `[settingIntMin, settingIntMax]` range guard + the two deleted regression tests from `main` — this is a real regression of #291, not just hygiene.
2. (🔴) Add `[settingFloatMin, settingFloatMax]` range guard on the new float write path + `getFloatClamped` read-side defence + the same shape of tests. The int hardening pattern was explicitly written to be extended; extend it.
3. (🟡) Rebase on `main` (currently CONFLICTING) so we can review the actual post-merge state.
4. (🟡) Optional: one-line comment on `requirePushEnabled` explaining why response-body uniformity isn't required for the global-disabled case.

Design is approved on the merits; once #1 and #2 land (with tests) and #3 is a clean rebase, happy to approve quickly.

Jerry-Xin (dismissed):
Summary: The PR is in scope for `octo-server`, but the management feature gate does not actually stop write handlers from running.

**🔴 Blocking**
- 🔴 Critical: `requireMgmtEnabled` writes the 403 response but does not abort the Gin/wkhttp handler chain. In [modules/incomingwebhook/api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/incomingwebhook/api.go:205), the disabled branch calls `mgmtFeatureDisabled(c)` and returns, but `mgmtFeatureDisabled` only renders the error in [modules/incomingwebhook/api_i18n.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/incomingwebhook/api_i18n.go:87). It does not call `c.Abort()`. The project’s own `httperr.ResponseErrorL` documentation says callers must abort when used in middleware. Because these routes register `w.requireMgmtEnabled()` before `w.create`, `w.update`, `w.delete`, and `w.regenerate` in [modules/incomingwebhook/api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/incomingwebhook/api.go:157), the downstream write handler can still execute after the 403 is written. The current test only checks the response status in [modules/incomingwebhook/api_test.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/incomingwebhook/api_test.go:233), so it can miss the actual side effect. Fix by aborting in the disabled management middleware/helper and add assertions that create/update/delete/regenerate do not mutate rows while disabled.

**💬 Non-blocking**
- 🟡 Warning: The new DB-backed numeric settings accept invalid operational values that the old env path rejected. `settingTypeInt` and `settingTypeFloat` only check parseability in [modules/common/api_manager_system_setting.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/common/api_manager_system_setting.go:203), so `0`, negative values, `NaN`, and infinities can be stored for rate-limit settings. Then `allowPerWebhook` treats `rps <= 0 || burst <= 0` as unlimited in [modules/incomingwebhook/ratelimit.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/incomingwebhook/ratelimit.go:140), while `NaN` can push the Redis limiter into fail-open behavior. Consider validating `per_webhook_rps`, `per_webhook_burst`, and `max_per_group` as finite positive values unless “unlimited” is an explicit documented admin feature.

**✅ Highlights**
- The push-side disabled gate correctly aborts and is placed before the expensive limiter/DB path.
- New incoming webhook error codes and locale entries are added consistently.

Testing note: I attempted `go test ./modules/common ./modules/incomingwebhook`, but the local test environment has no MySQL running at `127.0.0.1:3306`, so the infra-backed tests could not complete.

Jerry-Xin (changes requested):
Summary: The PR is in scope for `octo-server`, but the management feature gate does not actually stop write handlers from running.

**🔴 Blocking**
- 🔴 Critical: `requireMgmtEnabled` writes the 403 response but does not abort the Gin/wkhttp handler chain. In [modules/incomingwebhook/api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/incomingwebhook/api.go:205), the disabled branch calls `mgmtFeatureDisabled(c)` and returns, but `mgmtFeatureDisabled` only renders the error in [modules/incomingwebhook/api_i18n.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/incomingwebhook/api_i18n.go:87). It does not call `c.Abort()`. The project’s own `httperr.ResponseErrorL` documentation says callers must abort when used in middleware. Because these routes register `w.requireMgmtEnabled()` before `w.create`, `w.update`, `w.delete`, and `w.regenerate` in [modules/incomingwebhook/api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/incomingwebhook/api.go:157), the downstream write handler can still execute after the 403 is written. The current test only checks the response status in [modules/incomingwebhook/api_test.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/incomingwebhook/api_test.go:233), so it can miss the actual side effect. Fix by aborting in the disabled management middleware/helper and add assertions that create/update/delete/regenerate do not mutate rows while disabled.

**💬 Non-blocking**
- 🟡 Warning: The new DB-backed numeric settings accept invalid operational values that the old env path rejected. `settingTypeInt` and `settingTypeFloat` only check parseability in [modules/common/api_manager_system_setting.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/common/api_manager_system_setting.go:203), so `0`, negative values, `NaN`, and infinities can be stored for rate-limit settings. Then `allowPerWebhook` treats `rps <= 0 || burst <= 0` as unlimited in [modules/incomingwebhook/ratelimit.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ksvkedt6/repo/modules/incomingwebhook/ratelimit.go:140), while `NaN` can push the Redis limiter into fail-open behavior. Consider validating `per_webhook_rps`, `per_webhook_burst`, and `max_per_group` as finite positive values unless “unlimited” is an explicit documented admin feature.

**✅ Highlights**
- The push-side disabled gate correctly aborts and is placed before the expensive limiter/DB path.
- New incoming webhook error codes and locale entries are added consistently.

Testing note: I attempted `go test ./modules/common ./modules/incomingwebhook`, but the local test environment has no MySQL running at `127.0.0.1:3306`, so the infra-backed tests could not complete.

Jerry-Xin (dismissed):
## R2 re-review — head befcb65cbf0e

### R1 blocker status

| R1 ID | Issue | Status |
|-------|-------|--------|
| P0-1 | `mgmtFeatureDisabled` missing `c.Abort()` | 🔴 **Still open** |
| P0-2 | Rebase dropped int range guard from #291 | ✅ Fixed — `settingIntMin`/`settingIntMax` check + two regression tests restored |
| P0-3 | `settingTypeFloat` write path has no range validation | 🔴 **Still open** |
| 🟡 | mergeable=CONFLICTING | ✅ Fixed — now MERGEABLE |

---

### 🔴 Blocking

**🔴 P0-1 (unchanged): `requireMgmtEnabled` middleware does not abort the Gin handler chain**

`mgmtFeatureDisabled(c)` ([api_i18n.go:87-89](modules/incomingwebhook/api_i18n.go)) writes the 403 JSON response but does **not** call `c.Abort()`. Compare with `pushDisabled` at line 63-66 which correctly aborts. The `return` in the `requireMgmtEnabled` closure ([api.go:208-209](modules/incomingwebhook/api.go)) only exits the current middleware — Gin still invokes the next handler (`create`/`update`/`delete`/`regenerate`), so DB mutations execute even though a 403 was already written.

Fix: add `c.Abort()` in `mgmtFeatureDisabled` (symmetric with `pushDisabled`), and extend the toggle test to assert that no row/token/status mutation occurs when the feature is disabled.

**🔴 P0-3 (unchanged): `settingTypeFloat` write path accepts NaN, ±Inf, and non-positive values**

The float validation branch in `api_manager_system_setting.go:224-232` only checks `strconv.ParseFloat` success. It does not:
1. Reject `NaN`/`±Inf` (both pass `ParseFloat` in Go)
2. Enforce a `[settingFloatMin, settingFloatMax]` range (the int path already has this pattern with `settingIntMin`/`settingIntMax`)
3. Clamp on the read side — `getFloat` ([system_settings.go:238-247](modules/common/system_settings.go)) returns the raw parsed value without any defense

Consequence: an admin can write `per_webhook_rps = NaN` or `per_webhook_rps = -1.0` via the system setting API. `allowPerWebhook` in `ratelimit.go:140` treats `rps <= 0 || burst <= 0` as "unlimited", silently disabling rate limiting. Add:
- `settingFloatMin`/`settingFloatMax` constants + write-side guard (reject `math.IsNaN`, `math.IsInf`, out-of-range)
- `getFloatClamped` or equivalent read-side defense
- Tests for `-1.0`, `NaN`, `+Inf`, `abc` rejection and boundary acceptance

---

### 🟡 Non-blocking (carried from R1)

**🟡 Numeric settings not validated as positive** — `per_webhook_rps`, `per_webhook_burst`, `max_per_group` accept 0 through the int range guard (`settingIntMin=0`). Zero `burst` or `rps` → `allowPerWebhook` silently bypasses the rate limiter entirely. Unless zero is intentionally documented as "disable limiter/quota", consider raising `settingIntMin` to 1 for these keys (or add per-key min overrides in the schema).

---

### ✅ Highlights

- P0-2 is properly restored — int range guard and both regression tests are back.
- Merge conflict is resolved.
- The `SystemSettings` integration follows the shared pattern well; DB → env → default precedence is centralized.
- Push-side disabled path (`pushDisabled`) correctly aborts with a uniform 404.
- Getter-chain test coverage (`TestSystemSettings_IncomingWebhook_GetterChain_NoInfra`) is solid.

Two R1 blockers remain open (P0-1, P0-3). Verdict: **REQUEST_CHANGES**.

Jerry-Xin (dismissed):
## R2 re-review — head befcb65cbf0e

### R1 blocker status

| R1 ID | Issue | Status |
|-------|-------|--------|
| P0-1 | `mgmtFeatureDisabled` missing `c.Abort()` | 🔴 **Still open** |
| P0-2 | Rebase dropped int range guard from #291 | ✅ Fixed — `settingIntMin`/`settingIntMax` check + two regression tests restored |
| P0-3 | `settingTypeFloat` write path has no range validation | 🔴 **Still open** |
| 🟡 | mergeable=CONFLICTING | ✅ Fixed — now MERGEABLE |

---

### 🔴 Blocking

**🔴 P0-1 (unchanged): `requireMgmtEnabled` middleware does not abort the Gin handler chain**

`mgmtFeatureDisabled(c)` ([api_i18n.go:87-89](modules/incomingwebhook/api_i18n.go)) writes the 403 JSON response but does **not** call `c.Abort()`. Compare with `pushDisabled` at line 63-66 which correctly aborts. The `return` in the `requireMgmtEnabled` closure ([api.go:208-209](modules/incomingwebhook/api.go)) only exits the current middleware — Gin still invokes the next handler (`create`/`update`/`delete`/`regenerate`), so DB mutations execute even though a 403 was already written.

Fix: add `c.Abort()` in `mgmtFeatureDisabled` (symmetric with `pushDisabled`), and extend the toggle test to assert that no row/token/status mutation occurs when the feature is disabled.

**🔴 P0-3 (unchanged): `settingTypeFloat` write path accepts NaN, ±Inf, and non-positive values**

The float validation branch in `api_manager_system_setting.go:224-232` only checks `strconv.ParseFloat` success. It does not:
1. Reject `NaN`/`±Inf` (both pass `ParseFloat` in Go)
2. Enforce a `[settingFloatMin, settingFloatMax]` range (the int path already has this pattern with `settingIntMin`/`settingIntMax`)
3. Clamp on the read side — `getFloat` ([system_settings.go:238-247](modules/common/system_settings.go)) returns the raw parsed value without any defense

Consequence: an admin can write `per_webhook_rps = NaN` or `per_webhook_rps = -1.0` via the system setting API. `allowPerWebhook` in `ratelimit.go:140` treats `rps <= 0 || burst <= 0` as "unlimited", silently disabling rate limiting. Add:
- `settingFloatMin`/`settingFloatMax` constants + write-side guard (reject `math.IsNaN`, `math.IsInf`, out-of-range)
- `getFloatClamped` or equivalent read-side defense
- Tests for `-1.0`, `NaN`, `+Inf`, `abc` rejection and boundary acceptance

---

### 🟡 Non-blocking (carried from R1)

**🟡 Numeric settings not validated as positive** — `per_webhook_rps`, `per_webhook_burst`, `max_per_group` accept 0 through the int range guard (`settingIntMin=0`). Zero `burst` or `rps` → `allowPerWebhook` silently bypasses the rate limiter entirely. Unless zero is intentionally documented as "disable limiter/quota", consider raising `settingIntMin` to 1 for these keys (or add per-key min overrides in the schema).

---

### ✅ Highlights

- P0-2 is properly restored — int range guard and both regression tests are back.
- Merge conflict is resolved.
- The `SystemSettings` integration follows the shared pattern well; DB → env → default precedence is centralized.
- Push-side disabled path (`pushDisabled`) correctly aborts with a uniform 404.
- Getter-chain test coverage (`TestSystemSettings_IncomingWebhook_GetterChain_NoInfra`) is solid.

Two R1 blockers remain open (P0-1, P0-3). Verdict: **REQUEST_CHANGES**.

Jerry-Xin (changes requested):
## R2 re-review — head befcb65cbf0e

### R1 blocker status

| R1 ID | Issue | Status |
|-------|-------|--------|
| P0-1 | `mgmtFeatureDisabled` missing `c.Abort()` | 🔴 **Still open** |
| P0-2 | Rebase dropped int range guard from #291 | ✅ Fixed — `settingIntMin`/`settingIntMax` check + two regression tests restored |
| P0-3 | `settingTypeFloat` write path has no range validation | 🔴 **Still open** |
| 🟡 | mergeable=CONFLICTING | ✅ Fixed — now MERGEABLE |

---

### 🔴 Blocking

**🔴 P0-1 (unchanged): `requireMgmtEnabled` middleware does not abort the Gin handler chain**

`mgmtFeatureDisabled(c)` ([api_i18n.go:87-89](modules/incomingwebhook/api_i18n.go)) writes the 403 JSON response but does **not** call `c.Abort()`. Compare with `pushDisabled` at line 63-66 which correctly aborts. The `return` in the `requireMgmtEnabled` closure ([api.go:208-209](modules/incomingwebhook/api.go)) only exits the current middleware — Gin still invokes the next handler (`create`/`update`/`delete`/`regenerate`), so DB mutations execute even though a 403 was already written.

Fix: add `c.Abort()` in `mgmtFeatureDisabled` (symmetric with `pushDisabled`), and extend the toggle test to assert that no row/token/status mutation occurs when the feature is disabled.

**🔴 P0-3 (unchanged): `settingTypeFloat` write path accepts NaN, ±Inf, and non-positive values**

The float validation branch in `api_manager_system_setting.go:224-232` only checks `strconv.ParseFloat` success. It does not:
1. Reject `NaN`/`±Inf` (both pass `ParseFloat` in Go)
2. Enforce a `[settingFloatMin, settingFloatMax]` range (the int path already has this pattern with `settingIntMin`/`settingIntMax`)
3. Clamp on the read side — `getFloat` ([system_settings.go:238-247](modules/common/system_settings.go)) returns the raw parsed value without any defense

Consequence: an admin can write `per_webhook_rps = NaN` or `per_webhook_rps = -1.0` via the system setting API. `allowPerWebhook` in `ratelimit.go:140` treats `rps <= 0 || burst <= 0` as "unlimited", silently disabling rate limiting. Add:
- `settingFloatMin`/`settingFloatMax` constants + write-side guard (reject `math.IsNaN`, `math.IsInf`, out-of-range)
- `getFloatClamped` or equivalent read-side defense
- Tests for `-1.0`, `NaN`, `+Inf`, `abc` rejection and boundary acceptance

---

### 🟡 Non-blocking (carried from R1)

**🟡 Numeric settings not validated as positive** — `per_webhook_rps`, `per_webhook_burst`, `max_per_group` accept 0 through the int range guard (`settingIntMin=0`). Zero `burst` or `rps` → `allowPerWebhook` silently bypasses the rate limiter entirely. Unless zero is intentionally documented as "disable limiter/quota", consider raising `settingIntMin` to 1 for these keys (or add per-key min overrides in the schema).

---

### ✅ Highlights

- P0-2 is properly restored — int range guard and both regression tests are back.
- Merge conflict is resolved.
- The `SystemSettings` integration follows the shared pattern well; DB → env → default precedence is centralized.
- Push-side disabled path (`pushDisabled`) correctly aborts with a uniform 404.
- Getter-chain test coverage (`TestSystemSettings_IncomingWebhook_GetterChain_NoInfra`) is solid.

Two R1 blockers remain open (P0-1, P0-3). Verdict: **REQUEST_CHANGES**.

Jerry-Xin (dismissed):
## R2 re-review — head befcb65cbf0e

### R1 blocker status

| R1 ID | Issue | Status |
|-------|-------|--------|
| P0-1 | `mgmtFeatureDisabled` missing `c.Abort()` | 🔴 **Still open** |
| P0-2 | Rebase dropped int range guard from #291 | ✅ Fixed — `settingIntMin`/`settingIntMax` check + two regression tests restored |
| P0-3 | `settingTypeFloat` write path has no range validation | 🔴 **Still open** |
| 🟡 | mergeable=CONFLICTING | ✅ Fixed — now MERGEABLE |

---

### 🔴 Blocking

**🔴 P0-1 (unchanged): `requireMgmtEnabled` middleware does not abort the Gin handler chain**

`mgmtFeatureDisabled(c)` ([api_i18n.go:87-89](modules/incomingwebhook/api_i18n.go)) writes the 403 JSON response but does **not** call `c.Abort()`. Compare with `pushDisabled` at line 63-66 which correctly aborts. The `return` in the `requireMgmtEnabled` closure ([api.go:208-209](modules/incomingwebhook/api.go)) only exits the current middleware — Gin still invokes the next handler (`create`/`update`/`delete`/`regenerate`), so DB mutations execute even though a 403 was already written.

Fix: add `c.Abort()` in `mgmtFeatureDisabled` (symmetric with `pushDisabled`), and extend the toggle test to assert that no row/token/status mutation occurs when the feature is disabled.

**🔴 P0-3 (unchanged): `settingTypeFloat` write path accepts NaN, ±Inf, and non-positive values**

The float validation branch in `api_manager_system_setting.go:224-232` only checks `strconv.ParseFloat` success. It does not:
1. Reject `NaN`/`±Inf` (both pass `ParseFloat` in Go)
2. Enforce a `[settingFloatMin, settingFloatMax]` range (the int path already has this pattern with `settingIntMin`/`settingIntMax`)
3. Clamp on the read side — `getFloat` ([system_settings.go:238-247](modules/common/system_settings.go)) returns the raw parsed value without any defense

Consequence: an admin can write `per_webhook_rps = NaN` or `per_webhook_rps = -1.0` via the system setting API. `allowPerWebhook` in `ratelimit.go:140` treats `rps <= 0 || burst <= 0` as "unlimited", silently disabling rate limiting. Add:
- `settingFloatMin`/`settingFloatMax` constants + write-side guard (reject `math.IsNaN`, `math.IsInf`, out-of-range)
- `getFloatClamped` or equivalent read-side defense
- Tests for `-1.0`, `NaN`, `+Inf`, `abc` rejection and boundary acceptance

---

### 🟡 Non-blocking (carried from R1)

**🟡 Numeric settings not validated as positive** — `per_webhook_rps`, `per_webhook_burst`, `max_per_group` accept 0 through the int range guard (`settingIntMin=0`). Zero `burst` or `rps` → `allowPerWebhook` silently bypasses the rate limiter entirely. Unless zero is intentionally documented as "disable limiter/quota", consider raising `settingIntMin` to 1 for these keys (or add per-key min overrides in the schema).

---

### ✅ Highlights

- P0-2 is properly restored — int range guard and both regression tests are back.
- Merge conflict is resolved.
- The `SystemSettings` integration follows the shared pattern well; DB → env → default precedence is centralized.
- Push-side disabled path (`pushDisabled`) correctly aborts with a uniform 404.
- Getter-chain test coverage (`TestSystemSettings_IncomingWebhook_GetterChain_NoInfra`) is solid.

Two R1 blockers remain open (P0-1, P0-3). Verdict: **REQUEST_CHANGES**.

yujiawei (changes requested):
# Code Review — PR #292 (octo-server)

Independent review of `feat(incomingwebhook): runtime master toggle + tunable thresholds via system_setting` at head `befcb65`. Classified security-sensitive; reviewed with extra care on the rate-limiting and authorization surfaces.

## Verdict: Changes requested

The master-toggle wiring and the anti-enumeration / authorization design are solid. The blocker is in the new **`settingTypeFloat`** validation path: the per-webhook rate limit — the exact security control this PR is wiring for runtime control — can be silently disabled through the supported admin API, with no equivalent of the `D-289` guard that the sibling int settings already enforce.

---

## What's correct (verified)

- **Anti-enumeration on push is sound.** `requirePushEnabled` returns a *uniform* 404 for every push when the feature is globally off (`modules/incomingwebhook/api.go`), whereas an unknown webhook / bad token returns a uniform 401 (`pushUnauthorized`, `api.go:637/651/670/674`). The 404 is a global-state signal, not a per-webhook one, so it does not leak whether any specific webhook exists. Placing it first in the limiter chain means: when `enabled=1` it is a zero-behavior `c.Next()` (no regression), and when `enabled=0` it short-circuits before floor/IP/DB (no amplification).
- **Authorization semantics are right.** `requireMgmtEnabled` (403) is mounted *after* `AuthMiddleware` and is a feature gate, not an authz check; it does not bypass the handler's group-admin check. `list` (read) is intentionally left open. 403 for a disabled write, 404 for a disabled push, are both appropriate.
- **i18n / error codes are complete and compliant.** `push_disabled` (404) and `mgmt_disabled` (403) are registered in `pkg/errcode/incomingwebhook.go`, have zh-CN (`active.zh-CN.toml`) and en-US (`active.en-US.toml`) markers, go through `httperr.ResponseErrorLWithStatus` (consistent with every sibling responder in this module), and neither 4xx code is flagged `Internal`.
- **No dead code / no env-var rename.** The removed consts/funcs (`maxPerGroup`, `perWebhookRPS`, `perWebhookBurst`, etc.) have no remaining references; `os`/`strconv` are still used in `api.go`. The env names (`DM_INCOMINGWEBHOOK_RPS/_BURST/_MAX_PER_GROUP`) are preserved exactly, so existing env tuning keeps working. `go build ./...` and `go vet` pass on the touched packages.
- **Snapshot reads are safe.** `IncomingWebhookEnabled()` etc. read an atomic snapshot per request; `lookup` has a nil guard and the snapshot is seeded with a non-nil empty map, so no nil-deref on early requests.

---

## P1 (blocker) — `per_webhook_rps` has no finiteness / positivity validation on write and no clamp on read

`modules/common/api_manager_system_setting.go:224-232`

```go
case settingTypeFloat:
    if item.Value != "" {
        if _, err := strconv.ParseFloat(item.Value, 64); err != nil {
            c.JSON(http.StatusBadRequest, jsonH{"msg": fmt.Sprintf("%s.%s 必须是数字", ...)})
            return
        }
    }
```

The only float write-validation is `strconv.ParseFloat`, which **accepts** `"0"`, `"-5"`, `"NaN"`, `"Inf"`, `"+Inf"`, `"-Inf"`. The read path `getFloat` (`modules/common/system_settings.go:238`) returns the stored value verbatim — there is no `getFloatClamped` and no finiteness check. The value flows straight into the limiter:

`modules/incomingwebhook/ratelimit.go:140-147`
```go
rps := w.settings.IncomingWebhookPerWebhookRPS()
burst := w.settings.IncomingWebhookPerWebhookBurst()
if rps <= 0 || burst <= 0 { return true, nil }   // <-- fail-open: limiter OFF
```

Tracing each pathological value an admin can set via `POST /v1/manager/common/system_setting`:

- **`rps = 0` or negative** → `rps <= 0` is true → `allowPerWebhook` returns `(true, nil)` → the per-webhook token bucket is **silently disabled**.
- **`rps = NaN` / `±Inf`** → `NaN <= 0` and `+Inf <= 0` are both *false*, so the value is passed into the Lua script. There `rate = tonumber(ARGV[1])` on the serialized string (`"NaN"`/`"+Inf"`, via go-redis `strconv.AppendFloat`) yields `nil` in Redis Lua 5.1, and the next line `if rate <= 0` evaluates `nil <= 0` → **script runtime error** → `runBucketScript` returns `(true, err)` → caller fails open (`api.go` push path) → limiter **disabled**, plus a Redis error logged on every push (hot path).

Net effect: every out-of-domain value converges on "per-webhook flood control disabled on an unauthenticated push endpoint." This is precisely the control this PR exists to make runtime-tunable, and an operator who *intends to tighten it* (or fat-fingers the value) can silently turn it off with no validation feedback. The repo already established the correct pattern for this in `D-289` — int settings are bounds-checked on write (`settingIntMin/Max`) **and** clamped on read (`getIntClamped`, `system_settings.go:230`). The new float type re-opens exactly that gap.

Reachability note: the write endpoint is super-admin-gated, so this is not unauth-exploitable or a privilege-escalation. But on a security-sensitive PR whose subject *is* this rate-limit control, shipping a new tunable type that can silently nullify the control — and that diverges from the codebase's own `D-289` invariant — should be fixed before merge.

**Requested fix:** in the `settingTypeFloat` write case, after `ParseFloat` also reject `math.IsNaN(f) || math.IsInf(f, 0) || f <= 0` for these rate keys (return 400); and mirror `D-289` with a read-side clamp (`getFloatClamped`, or a `<=0`/non-finite check inside `IncomingWebhookPerWebhookRPS`) so a direct DB edit cannot reach the limiter as NaN/Inf/0/negative either.

---

## P2 (non-blocking, recommended to fix alongside the P1)

1. **`per_webhook_burst` / `max_per_group` use `getInt`, not `getIntClamped`** — `modules/common/system_settings.go:584,590`. Every other int setting in this file (the sidebar `recent_filter_*_days` getters at `516/522/530`) uses `getIntClamped` for the `D-289` read-side defense. A direct DB row with `per_webhook_burst=-5` is served verbatim → `burst <= 0` → same fail-open as above. Switch both getters to `getIntClamped` for consistency.

2. **Reusing the day-window range `[settingIntMin=0, settingIntMax=3650]` for burst/quota** — `modules/common/system_setting_schema.go:14-25`. The schema comment itself warns that "an int setting that needs a different range should move this to a per-key field on `settingDef`." With min=0, `per_webhook_burst=0` is a *valid* admin write that disables the limiter (matching the float hazard), and `max_per_group=0` is a valid write that blocks all webhook creation (`insertWithQuota`: `count >= 0` always true → `ErrQuotaExceeded`). Consider per-key bounds with a minimum of 1 for these two controls. The 3650 ceiling on a burst is also semantically odd.

3. **`getBool` DB parsing is narrower than the env-default parsing** — `system_settings.go:196-203` accepts only `1/true/TRUE/0/false/FALSE`, while `incomingWebhookEnabledEnvDefault` accepts `yes/no/on/off` (case-insensitive, trimmed). Via the admin UI this is harmless (`normaliseBool` only ever stores `"0"`/`"1"`). It only bites a direct DB edit of `enabled='yes'/'on'`, which would silently fall back to the env default — and since this is a *master switch*, falling back the wrong way is worse than for a threshold. Worth either unifying the two parsers or documenting "DB stores canonical 0/1 only."

4. **Test gap mirrors the validation gap** — `modules/common/system_settings_test.go` only covers the happy-path float (`8.5`). There is no test asserting the write path rejects a non-numeric / out-of-range `per_webhook_rps`, nor that a negative/NaN rps does not silently disable the limiter. Adding these naturally follows the P1 fix.

(`floatToCanonical` producing `"NaN"`/`"+Inf"` that round-trip through GET→POST is a real but cosmetic consequence of the same root cause; it disappears once the P1 write-side validation lands.)

---

## Manual-verification asks for a human reviewer

- Confirm the intended product semantics of `rps <= 0` / `burst <= 0`: is "disable the per-webhook limiter" a deliberate knob, or an accidental side effect? If deliberate, it should be an explicit, documented value, not a silent consequence of a threshold field — and it should arguably be gated behind `enabled=0` rather than reachable by setting a rate to zero.
- The DB-backed `Reload` tests and the end-to-end `TestFeatureToggle_DisabledStopsPushAndMgmtWrites` were compiled but not executed locally (need MySQL/Redis); confirm they pass in CI.

lml2468 (dismissed):
APPROVED (head `d90560fd0077`).

Single commit cleanly resolves all three R1/R2 blockers (mine + Jerry-Xin's) and goes one step further than I asked on the float-validation shape. Each blocker traced through verification below.

## ✅ P0-1 — `mgmtFeatureDisabled` now `c.Abort()`s + regression test asserts the actual invariant

**Fix** (`api_i18n.go:87-95`): `mgmtFeatureDisabled` now calls `c.Abort()` after `ResponseErrorLWithStatus`, symmetric with `pushDisabled`. The new comment block is precise about *why* this responder differs from the other mgmt helpers:

> *"Unlike the other mgmt responders (which are called from inside a handler that returns immediately after), this one runs from the requireMgmtEnabled MIDDLEWARE, so it MUST c.Abort() — a bare return only exits the middleware closure and Gin would still invoke the downstream write handler, executing the mutation after the 403 was written."*

This documentation is exactly what was missing in the R1 review — it tells the next reader why "mgmt helpers don't need Abort" doesn't apply here.

**Regression test** (`api_test.go:5)`: the `TestFeatureToggle_DisabledStopsPushAndMgmtWrites` step 5 was upgraded from "list returns 200" to:

```go
require.Lenf(t, list, 1, "disabled create must not insert a row; list=%s", lw.Body.String())
assert.Equal(t, "toggle-wh", first["name"], "the only webhook must be the original, not the blocked create")
```

That's the right shape: doesn't just trust the 403 status, actually queries DB state and proves no row was written. If `c.Abort()` were ever removed again, the blocked `create` would land a second row named "blocked" and this assertion would fail loudly. Intent-locking test, not a vanity test.

## ✅ P0-2 — `Positive` per-key flag + read-side clamps cover write **and** direct-DB bypass

The fix here is better than what I proposed. Two improvements:

**Per-key `Positive` flag on `settingDef`** (`system_setting_schema.go`): rather than introducing a global `[settingFloatMin, settingFloatMax]`, the fix adds a `Positive bool` field that the rate-limit / quota keys opt into. The doc-comment explicitly references the forward note from the int range guard:

> *"the schema comment on settingIntMin anticipated this per-key override. No artificial upper bound is imposed (matches the env semantics these keys fall back to)."*

This is the right call — rate-limit / quota knobs shouldn't have artificial upper bounds (`per_webhook_burst = 100000` is a legitimate value for high-traffic systems), and the `[0, 3650]` shared bound was always day-window-shaped. The `Positive` flag captures the actual invariant: "must be strictly positive finite", which is what the limiter logic needs.

**Write path** (`api_manager_system_setting.go`): all the silent-disable paths I called out are now closed:

- `case settingTypeInt`: `Positive` keys require `n > 0`; non-`Positive` keys stay on `[settingIntMin, settingIntMax]`. Behaviour for sidebar.recent_filter_*_days unchanged (still allows 0 as the "disable filter" sentinel).
- `case settingTypeFloat`: `math.IsNaN(f) || math.IsInf(f, 0)` rejected before any `Positive` check. NaN was the subtle one — `strconv.ParseFloat("NaN", 64)` succeeds, returns NaN, and would have landed in DB. Now it 400s with a clear message.
- `Positive` float keys additionally require `f > 0`.

**Read side** (`system_settings.go`): all three IncomingWebhook* getters now clamp:

```go
func (s *SystemSettings) IncomingWebhookPerWebhookRPS() float64 {
    def := wkhttp.ParseRPSFromEnv(...)
    v := s.getFloat("incomingwebhook", "per_webhook_rps", def)
    if math.IsNaN(v) || math.IsInf(v, 0) || v <= 0 {
        return def
    }
    return v
}
```

And the symmetric guards on Burst and MaxPerGroup. Each carries a tight comment explaining the specific failure mode being defended (rps short-circuits the bucket / NaN errors the Lua script / max_per_group=0 is a hard-to-diagnose "dark disable"). Defense-in-depth — write path now rejects, but a direct DB edit still can't reach the limiter. Mirrors exactly the `getIntClamped` pattern #291 introduced.

**Tests** cover the matrix:
- `TestManagerSystemSetting_UpdateRejectsInvalidIncomingWebhookNumerics` — 10 reject cases: `0` / `-1.5` / `NaN` / `+Inf` / `-Inf` / `abc` for rps; `0` / `-1` for burst; `0` / `-5` for max_per_group. Each asserted with the specific category.key for clear failure messages.
- `TestManagerSystemSetting_UpdateAcceptsPositiveIncomingWebhookNumerics` — `8.5` / `100000` / `50` accepted, with the explicit `"no artificial upper bound for Positive keys"` assertion message proving the design choice is locked.
- Read-side clamp tests (mentioned in the commit body) for NaN/±Inf/≤0 → default.

## 🟡 Non-blocking observation (not raised before, just noting)

The `Positive` flag's documentation block at `system_setting_schema.go` already anticipates the third case ("Adding an int setting that needs a *different* range should move this to a per-key field on settingDef"). If a future setting needs a custom `[min, max]` range (not just "positive"), the same `settingDef` extension point is the natural home — e.g. `Min, Max *int`/`*float64`. No action needed now; the pattern is the right shape and easy to extend when actually needed. Just a forward note that the int comment's anticipated per-key bound and this Positive flag together establish the extension trajectory.

## Combined R1+R2+R3 follow-through

| Blocker | R1/R2 status | R3 status |
|---|---|---|
| 🔴 (Jerry-Xin) `requireMgmtEnabled` doesn't actually stop writes | flagged, P0 | fixed (`c.Abort()` + intent-locking list-len test) |
| 🔴 (lml2468) `[0,3650]` int range guard regression from #291 | flagged, P0 | fixed (was rebase slip; restored on previous round; tests back) |
| 🔴 (lml2468 + Jerry-Xin) float write path has no range guard | flagged, P0 | fixed (NaN/±Inf rejection + per-key `Positive` flag, better than my proposal) |
| 🔴 (Jerry-Xin) `Positive` validation missing on burst/max_per_group | implied by #3 | fixed (all three keys carry `Positive: true`) |
| 🟡 (Jerry-Xin) `getFloatClamped`-equivalent read-side defence missing | flagged | fixed (all 3 IncomingWebhook* getters now clamp) |
| 🟡 (lml2468) `requirePushEnabled` comment on uniformity vs anti-enum | suggested | not addressed (still defensible, low priority) |

Every blocker resolved with both production fix and regression test. The non-blocking `requirePushEnabled` comment is the only suggestion left unaddressed and is fine to defer.

## Checks

`mergeable: MERGEABLE`. CI: will pick up the new tests on the new SHA; the no-infra read-side clamp tests should pass quickly, the DB-backed `TestFeatureToggle_*` toggle test will need CI's MySQL/Redis. Ride them out and merge.

LGTM. Iteration discipline on this one is excellent — single tight commit that addresses every concern with one consistent design (`Positive` flag) rather than a series of one-off patches.

an9xyz:
Thanks for the thorough review — pushed `d90560f` addressing both blockers.

**P0-1 (`mgmtFeatureDisabled` not aborting):** fixed. It's called from the `requireMgmtEnabled` *middleware*, so it now calls `c.Abort()` (symmetric with `pushDisabled`). The toggle test now also asserts a disabled `create` inserts **no row** (list stays at 1), so the abort is pinned, not just the status code.

**P0-2 (float / numeric validation silently disabling the limiter):** hardened on both sides, per maintainer call on the validation policy:
- **Write:** `settingTypeFloat` now rejects `NaN` / `±Inf`. New per-key `Positive` flag on `settingDef` (the extension the `settingIntMin` comment anticipated) makes `per_webhook_rps` / `per_webhook_burst` / `max_per_group` require a strictly-positive value.
- **Read:** the `IncomingWebhook*` getters clamp `≤0` / non-finite back to the env/code default, so a direct DB edit can't reach the limiter either.

One deliberate divergence from the "explicit `[min,max]`" suggestion: the `Positive` keys **opt out of the shared `[0,3650]` cap and impose no artificial upper bound** — matching the env semantics these keys fall back to (env had no ceiling either). A `burst=100000` write is intentionally accepted (test pins this). Happy to add a sanity ceiling if you'd prefer one.

Tests added: read-side clamp (no-infra), manager write reject/accept for the new keys, and the no-mutation-while-disabled assertion. The 🟡 on `getInt` vs `getIntClamped` for burst/max_per_group is moot now — the read-side `≤0 → default` clamp covers it directly.

---

Jerry-Xin (dismissed):
Summary: The PR is in scope for octo-server and the runtime toggle design fits the existing `SystemSettings` pattern; I found no merge-blocking defects.

**💬 Non-blocking**
- 🟡 Warning: [modules/common/system_settings.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-osk8c_38/repo/modules/common/system_settings.go:583) still lets non-finite env fallback values escape. `ParseRPSFromEnv` accepts `NaN` and positive infinity, and the getter returns `def` when it detects an invalid DB value, so `DM_INCOMINGWEBHOOK_RPS=NaN` can still reach [modules/incomingwebhook/ratelimit.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-osk8c_38/repo/modules/incomingwebhook/ratelimit.go:141) and fail open. Consider sanitizing the env-derived fallback locally before using it.
- 🟡 Warning: [modules/incomingwebhook/api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-osk8c_38/repo/modules/incomingwebhook/api.go:157) gates management writes before group-admin authorization. That means any authenticated user can receive the feature-disabled error for arbitrary group paths. The state is global, so this is not a webhook enumeration leak, but moving the check after `requireGroupAdmin` would keep operator state less broadly visible.
- 🔵 Suggestion: [modules/common/api_manager_system_setting.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-osk8c_38/repo/modules/common/api_manager_system_setting.go:220) permits any positive finite RPS and any positive int burst. Extremely tiny RPS values or enormous bursts can produce pathological Redis Lua TTLs and effectively degrade the limiter. A sane per-key min/max would make the admin surface safer.
- 🔵 Suggestion: The new validation branches in [modules/common/api_manager_system_setting.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-osk8c_38/repo/modules/common/api_manager_system_setting.go:222) continue the existing raw JSON error style. That endpoint appears legacy, but future cleanup should migrate these manager errors to the localized error envelope.

**✅ Highlights**
- The feature toggle is wired through the shared `SystemSettings` singleton, so manager reloads and incoming-webhook reads share the same snapshot.
- The push and management disabled responders use registered i18n error codes and abort correctly from middleware.
- Added coverage pins the DB-to-env-to-default getter chain and the middleware abort behavior.

Verification run: `go test ./modules/common -run 'TestSystemSettings_IncomingWebhook_(GetterChain|ReadSideClamp)_NoInfra'`.

yujiawei (dismissed):
# Code Review — PR #292 (octo-server)

Independent review of `feat(incomingwebhook): runtime master toggle + tunable thresholds via system_setting` at head `d90560f`. Classified security-sensitive; reviewed with extra care on the rate-limiting, validation, and authorization surfaces.

## Verdict: Approved

The three blockers raised on earlier revisions are all resolved at this head, and an independent multi-lens pass (security / correctness / regression) surfaced no remaining merge-blocking issues. The design — moving the master toggle and core thresholds onto `system_setting` with `DB → env → code-default` precedence, with defence-in-depth on both the write and read sides — is sound.

## Prior blockers — all fixed at `d90560f`

| ID | Issue | Status |
|----|-------|--------|
| P0-1 | `requireMgmtEnabled` middleware did not abort the handler chain | ✅ Fixed — `mgmtFeatureDisabled` now calls `c.Abort()` (`modules/incomingwebhook/api_i18n.go:93-95`), symmetric with `pushDisabled`. `TestFeatureToggle_DisabledStopsPushAndMgmtWrites` asserts the blocked `create` does **not** insert a row (list stays at 1). |
| P0-2 | Rebase dropped the `[settingIntMin, settingIntMax]` int guard + #291 tests | ✅ Fixed — the range branch is restored (`api_manager_system_setting.go:227-233`), and both `TestManagerSystemSetting_UpdateRejectsOutOfRangeInt` / `UpdateAcceptsInRangeIntBoundaries` are present. |
| P0-3 | `settingTypeFloat` write path accepted NaN/±Inf/≤0; no read-side clamp | ✅ Fixed — write path rejects `math.IsNaN`/`math.IsInf` and (for Positive keys) `≤0` (`api_manager_system_setting.go:234-261`); read side clamps non-finite/`≤0` back to default in `IncomingWebhookPerWebhookRPS/Burst/MaxPerGroup` (`system_settings.go`). |

## What's correct (verified)

- **The rate limiter can no longer be silently disabled through a supported path.** `per_webhook_rps` (float), `per_webhook_burst` and `max_per_group` (int) are marked `Positive: true`, so the admin write path rejects `0`/negative/`NaN`/`±Inf`/non-numeric. The matching read-side getters re-clamp, so even a direct DB edit can't deliver a pathological value into `allowPerWebhook`'s `rps <= 0 || burst <= 0` short-circuit or into the Redis Lua script as `NaN`. This closes the exact `D-289`-class gap the new float type briefly re-opened, and adds defence in depth on both sides.
- **Anti-enumeration on push holds.** `requirePushEnabled` returns a *uniform* 404 for every push when the feature is globally off, mounted first in the limiter chain so a disabled feature short-circuits before floor/IP/DB (O(1), no amplification). The 404 is a global-state signal, not a per-webhook one, so it does not leak whether any specific webhook exists.
- **Authorization semantics are appropriate.** `requireMgmtEnabled` (403) is mounted *after* `AuthMiddleware` — it is a feature gate, not an authz bypass, and it does not weaken the handler's group-admin check. `list` (read) is intentionally left open for incident-time inspection. 403 for a disabled write, 404 for a disabled push, are both correct.
- **i18n / error codes are complete.** `push_disabled` (404) and `mgmt_disabled` (403) are registered in `pkg/errcode/incomingwebhook.go`, routed through `httperr.ResponseErrorLWithStatus`, with zh-CN and en-US markers, and neither 4xx is flagged `Internal`.
- **Single source of truth.** The env knobs now live solely in `modules/common`; `incomingwebhook` reads them only through the typed getters, so the admin `GET .../system_setting` effective value matches what the limiter actually uses. No env-var rename — existing `DM_INCOMINGWEBHOOK_*` tuning still works.
- **`floatToCanonical`** uses `strconv.FormatFloat(v, 'g', -1, 64)` (shortest round-trip), so GET → POST of the effective value is stable for any valid value.

Build (`go build ./...` on the touched packages) and the no-infra tests (`TestSystemSettings_IncomingWebhook_GetterChain_NoInfra`, `ReadSideClamp_NoInfra`, `TestIncomingWebhookRespondHelpers`) pass locally.

## Non-blocking suggestions (optional)

1. **`getBool` DB-parse is narrower than the env-default parser.** `getBool` accepts only `1/true/TRUE/0/false/FALSE`, while `incomingWebhookEnabledEnvDefault` also accepts `yes/no/on/off`. Harmless via the admin UI (which stores canonical `0`/`1`), but a hand-edited DB row of `enabled='on'` would silently fall back to the env default. Since this is the *master switch*, a one-line "DB stores canonical 0/1 only" note (or unifying the parsers) would prevent a surprising failure mode.
2. **`DM_INCOMINGWEBHOOK_RPS=NaN` as an env value** would make the read-side `def` itself NaN (the shared `wkhttp.ParseRPSFromEnv` rejects `≤0` but not non-finite), and the getter's clamp returns `def` unchanged. This is a pre-existing latent edge in the shared helper, only reachable by an operator literally setting the env var to `NaN`, not via the API — worth a follow-up in octo-lib rather than blocking here.

## Manual-verification asks for a human

- Confirm the DB-backed `Reload` tests and the end-to-end `TestFeatureToggle_DisabledStopsPushAndMgmtWrites` pass green in CI (they were compiled but not executed in the dev sandbox — no MySQL/Redis).
- Confirm the product intent that `enabled=0` is the *only* supported "off" switch and that thresholds can never be set to a disabling value — which the Positive-key validation now enforces.

an9xyz:
Thanks all for the approvals. Pushed one small follow-up `00658a6` for @Jerry-Xin's non-blocking env-NaN note: `ParseRPSFromEnv` accepts `NaN`/`+Inf`, so a `DM_INCOMINGWEBHOOK_RPS=NaN` env value became the getter's fallback and my read-side clamp returned that fallback verbatim — so NaN could still reach `allowPerWebhook`. The env-derived default is now sanitized too (non-finite / ≤0 → code default), with the no-infra clamp test extended to cover `env=NaN`.

Heads-up that this adds a commit on top of your approvals — if branch protection dismissed them on push, a re-approve would be appreciated. The other two non-blocking notes (gating mgmt-disabled before group-admin auth; min/max RPS sanity bounds) I'm leaving as-is unless you'd like them addressed.

---

Jerry-Xin (approved):
The PR is in scope for octo-server and the runtime incoming-webhook settings integration looks sound.

💬 Non-blocking
- 🟡 Warning: `modules/common/api_manager_system_setting.go:220` and `modules/common/api_manager_system_setting.go:234` add new raw `c.JSON` validation responses. This matches nearby existing code, but it continues to bypass the project’s localized error envelope. Consider migrating the whole manager system-setting validation path to registered errcodes in a follow-up.
- 🔵 Suggestion: `modules/incomingwebhook/api_test.go:201` sets `DM_INCOMINGWEBHOOK_ENABLED` after `setupTestEnv` and after the initial create path is already possible. Move that `t.Setenv` before `setupTestEnv` so the test is isolated from any externally disabled environment.
- 🔵 Suggestion: `modules/common/sql/20260518000001_common_legacy01.sql:14` still documents `value_type` as `string / bool / int / encrypted`. Since `float` is now a supported value type, the comment should be updated when convenient.

✅ Highlights
- The push toggle is applied before the limiter chain at `modules/incomingwebhook/api.go:185`, giving the intended uniform disabled behavior.
- Management writes are gated while list remains available, matching the operational requirement.
- The RPS getter sanitizes DB and env `NaN`/`Inf`/non-positive values before they reach the Redis limiter.
- No-infra verification passed:
  - `go test ./modules/common -run 'TestSystemSettings_IncomingWebhook_(ReadSideClamp_NoInfra|GetterChain_NoInfra)$'`
  - `go test ./modules/incomingwebhook -run 'TestIncomingWebhookRespondHelpers$'`

Jerry-Xin (approved):
The PR is in scope for octo-server and the runtime incoming-webhook settings integration looks sound.

💬 Non-blocking
- 🟡 Warning: `modules/common/api_manager_system_setting.go:220` and `modules/common/api_manager_system_setting.go:234` add new raw `c.JSON` validation responses. This matches nearby existing code, but it continues to bypass the project’s localized error envelope. Consider migrating the whole manager system-setting validation path to registered errcodes in a follow-up.
- 🔵 Suggestion: `modules/incomingwebhook/api_test.go:201` sets `DM_INCOMINGWEBHOOK_ENABLED` after `setupTestEnv` and after the initial create path is already possible. Move that `t.Setenv` before `setupTestEnv` so the test is isolated from any externally disabled environment.
- 🔵 Suggestion: `modules/common/sql/20260518000001_common_legacy01.sql:14` still documents `value_type` as `string / bool / int / encrypted`. Since `float` is now a supported value type, the comment should be updated when convenient.

✅ Highlights
- The push toggle is applied before the limiter chain at `modules/incomingwebhook/api.go:185`, giving the intended uniform disabled behavior.
- Management writes are gated while list remains available, matching the operational requirement.
- The RPS getter sanitizes DB and env `NaN`/`Inf`/non-positive values before they reach the Redis limiter.
- No-infra verification passed:
  - `go test ./modules/common -run 'TestSystemSettings_IncomingWebhook_(ReadSideClamp_NoInfra|GetterChain_NoInfra)$'`
  - `go test ./modules/incomingwebhook -run 'TestIncomingWebhookRespondHelpers$'`

yujiawei (approved):
# Code Review — PR #292 (octo-server)

**Verdict: APPROVED** ✅ — no blocking issues. This is a well-structured, security-conscious change with strong defense-in-depth and good test coverage.

## Summary

The PR wires the incoming-webhook module's master toggle and core thresholds (`per_webhook_rps` / `per_webhook_burst` / `max_per_group`) into the admin-tunable `system_setting` layer, with a clean `DB → env(DM_INCOMINGWEBHOOK_*) → code-default` fallback chain. `modules/common` becomes the single source of truth for the env knobs; `incomingwebhook` reads through typed getters. Runtime toggle, no restart, multi-instance convergence within the snapshot TTL.

I reviewed the full 4-commit series across three dimensions (security / correctness / design) and independently verified the key claims by reading the surrounding code, not just the diff. `go build ./...`, `go vet`, and the no-infra unit tests (`TestSystemSettings_IncomingWebhook_GetterChain_NoInfra`, `..._ReadSideClamp_NoInfra`, `TestIncomingWebhookRespondHelpers`) all pass locally.

## Verification of the design

✅ **Master toggle gating is correct.** `requirePushEnabled()` is mounted **first** in the push limiter chain (`api.go:185`) and short-circuits with a uniform 404 **before any DB lookup**, so it is indistinguishable across webhook IDs/tokens — the anti-enumeration invariant holds. `requireMgmtEnabled()` gates create/update/delete/regenerate with 403 while leaving `list` (read) open for operators. Both helpers correctly `c.Abort()`.

✅ **The two P0s from earlier review rounds are properly fixed.**
- Commit 3 (`d90560f`): `mgmtFeatureDisabled` now calls `c.Abort()`. This was a real, reachable bug — because it runs from *middleware* (not a handler), a bare return would have let Gin still invoke the downstream write handler, executing the mutation *after* the 403 was written. The toggle test now asserts no row is inserted while disabled (`require.Lenf(list, 1, ...)`), which is exactly the right regression guard.
- Commit 3 + 4: rate/quota values are hardened on **both** sides. Write path rejects `0`/negative/`NaN`/`±Inf` via `settingTypeFloat` validation + the new per-key `Positive` flag (which also correctly opts out of the shared `[0,3650]` day-window bound). Read path clamps `≤0`/non-finite back to the env/code default, and commit 4 additionally sanitizes the **env-derived default itself** (`ParseRPSFromEnv` uses `strconv.ParseFloat`, which accepts `NaN`/`+Inf`). This closes the `allowPerWebhook` `rps<=0 → always-allowed` short-circuit and the NaN→Lua-error→fail-open path. Defense in depth is complete across env, DB, and direct-DB-edit vectors.

✅ **i18n is complete and consistent.** Both new codes (`push_disabled` 404, `mgmt_disabled` 403) are registered in `pkg/errcode/incomingwebhook.go`, translated in `active.zh-CN.toml`, and have en-US markers. Neither is marked `Internal` — correct, since both are 4xx (per the repo's 5xx⟺Internal invariant).

✅ **GET `effective_value` does not leak secrets.** Masking applies only to `settingTypeEncrypted`; the new bool/int/float keys are non-sensitive and correctly surfaced.

✅ **Single-source-of-truth migration is clean** — the old `envMaxPerGroup`/`perWebhookRPS()`/`perWebhookBurst()`/`defaultRatePerWH*` symbols are fully removed, no dangling references or unused imports, no import cycle (`incomingwebhook → common` is one-directional).

## Non-blocking observations (nits — no change required to merge)

1. **`requireMgmtEnabled` runs before the in-handler group-admin check** (`api.go` route group). When the toggle is off, a non-admin authenticated user gets `mgmt_disabled` (403) rather than `mgmt_forbidden` (403). No security or status divergence (both 403, and the global feature state is already exposed unauthenticated via the push 404) — just a minor semantic detail in which error code surfaces first. Acceptable as designed; flagging only to make it a conscious choice.

2. **Cross-file "reciprocal sync" is comment-enforced only.** The schema rows, getter env-names, and defaults are currently fully in sync, but consistency is guaranteed only by hand-written comments — there's no compile-time or test assertion. A future edit to one env name/default could silently desync the admin `GET effective_value` from the getter. Consider a small table-driven test asserting each `incomingwebhook` schema row's `Effective()` matches its getter under a known snapshot.

3. **New validation messages use raw `c.JSON`** (Chinese-only) rather than the i18n envelope. This is consistent with the pre-existing pattern in `api_manager_system_setting.go` (the file was never i18n-migrated) and is invisible to the D23 lint guard (which only matches `AbortWithStatusJSON`), so CI stays green. Proper i18n compliance here would be a separate file-wide cleanup, out of scope for this PR.

## Items a human may want to manually confirm (security-sensitive PR)

- The DB-backed integration tests (`TestSystemSettings_IncomingWebhook*` Load/Reload, `TestFeatureToggle_DisabledStopsPushAndMgmtWrites`) were compiled but not executed locally (need MySQL/Redis) — confirm they pass in CI.
- Confirm `make i18n-extract-check` + `i18n-lint` are green in CI (markers verified present locally).

Nice work on the layered write+read hardening and the no-infra test that pins the precedence chain.

PR #285, opened 2026-06-05
feat(incomingwebhook): layered push rate limiting (local floor + per-IP failure budget)

an9xyz:
Hardens the **public, unauthenticated** push endpoint (`POST /v1/incoming-webhooks/:webhook_id/:token`). Addresses item 1 of #284.

## What changed — four-layer defense (coarse → fine)

```
localFloor (200/400, in-memory) → ipLimit (per-IP all requests, 100/200) → ipFailureGate (per-IP failures, 30/60) → allowPerWebhook (per-webhook, 5/10)
```

1. **Local floor** (`golang.org/x/time/rate`, in-memory, per-instance): always-on, Redis-independent backstop. Before this PR a Redis outage left the endpoint with **no** flood ceiling in front of 2 DB reads + 1 WuKongIM send per call (both Redis limiters fail open). Built once from env at construction.
2. **Per-IP request limiter** (`StrictIPRateLimitMiddleware`): caps **all** requests from one IP, bounding the "single IP holding many valid tokens" case below the floor.
3. **Per-IP auth-failure budget** (new): tokens are spent **only on genuine auth failures** (unknown / soft-deleted webhook, bad token, malformed) — never on valid pushes or DB errors. A read-only gate rejects an IP that burned its budget before the handler/DB run. This is why fixed/shared server IPs are not throttled by valid volume.
4. **Per-webhook** limiter: unchanged per-token shaping.

Trusted client IP via `clientIP()` (X-Real-Ip / **rightmost** X-Forwarded-For, matching octo-lib `getClientIP`) — not gin `c.ClientIP()`, which would read the spoofable leftmost XFF. `TestPush_IPBudgetNotSpoofableViaXFF` pins this.

## ⚠️ Operator-visible behavior changes (release-note worthy)

- **Per-IP request limiter default raised `30/60` → `100/200` rps/burst.** Higher cap so legitimate fixed/shared-egress integrations (NAT, CI servers pushing to many webhooks) aren't false-throttled, while still capping a single IP below the 200 rps floor. Env names **unchanged**: `DM_INCOMINGWEBHOOK_IP_RPS` / `DM_INCOMINGWEBHOOK_IP_BURST` — existing tuning still applies.
- **New env for the failure budget:** `DM_INCOMINGWEBHOOK_IP_FAIL_RPS` / `_BURST` (default 30/60). This is a *failure* budget, not a request volume cap — do **not** copy an old `IP_RPS` value into it.
[...]

PR #293, opened 2026-06-06
feat(incomingwebhook): cache the push hot-path lookups (#284 item 2)

an9xyz:
Implements **item 2** of #284 (item 1 shipped in #285/#288; this is its own PR per maintainer preference — independent of the config-toggle work in #292).

## Problem
The push path (`POST /v1/incoming-webhooks/:webhook_id/:token`) did **two uncached DB point reads** before rate-limiting and send:
1. `db.queryByWebhookID` — the webhook row
2. `requireActiveGroup` → `groupDB.QueryWithGroupNo` — the group-Normal check

Both are near-immutable / rarely-changing, yet they are the first DB read wall once any webhook gets high push QPS.

## Change
A small **per-instance, short-TTL, Redis-independent** cache (`cache.go`, hand-rolled generic `ttlCache[T]` — same spirit as `localFloor`, so it keeps working during a Redis outage) for both lookups. A warm push issues **0 DB reads**.

- **Positive-only**: only an existing row / a Normal group is cached. Nonexistent or soft-deleted webhook IDs are **never** negative-cached — scan floods are already cut by the per-IP failure budget (#285) before they reach the DB, so caching them would only let scan traffic pollute the cache.
- **Env-tunable**: `DM_INCOMINGWEBHOOK_CACHE_TTL_MS` (default 3000; **`0` disables** → exact pre-change direct-DB behavior) and `DM_INCOMINGWEBHOOK_CACHE_MAX` (default 10000; coarse size cap).

### Auth-gate staleness contract (accepted per #284)
The push auth gate must not be bypassable via a stale cache. Mutations invalidate the **local** entry immediately:

| path | invalidates |
|---|---|
| `update` (status/name/avatar) | webhook entry |
| `delete` (soft-delete) | webhook entry |
| `regenerate` (token rotation) | webhook entry |
| `handleGroupDisband` | group entry |

So the writing instance stops/rotates at once; **peers are bounded by the TTL** (seconds-level — the issue explicitly accepts this and offers `=0` as the escape hatch). The group entry alone gates an entire disbanded group even if its webhook rows are still briefly cached as enabled. **Management writes keep reading the DB uncached**, so revival/permission decisions never run on a stale snapshot.

## Tests
Infra-free (run & passing here):
- `TestTTLCache_*` — hit/expiry/invalidate/size-cap-clear/disabled(ttl=0)/nil-safe
- `TestCacheTTLEnv`, `TestCacheMaxEnv`
- `TestCachedQueryByWebhookID_HitSkipsDB`, `TestCachedRequireActiveGroup_HitSkipsDB` — leave `db` nil to **prove a warm hit issues 0 DB reads**
- `go build ./...`, `golangci-lint` (0 issues), `TestIncomingWebhookNoLegacyResponseError` (with `cache.go` added to the guard list)

CI-only (needs MySQL/Redis — blocked by this sandbox's network policy, compiled here):
- `TestPush_CacheServesStaleWithinTTL_HandlerInvalidates` — end-to-end: a DB-only disable behind the handler's back still authorizes within the TTL (cache in effect); a handler `DELETE` invalidates → next push is rejected (401).

## Acceptance (from #284 item 2)
- ✅ push issues 0 DB reads on a warm cache hit (proven by the hit-skips-DB tests)
- ✅ disabling/deleting a webhook or disbanding the group stops pushes within the TTL window (immediate on the writing instance via invalidation; TTL-bounded cross-instance)
- ✅ existing race/soft-delete tests unchanged (management paths stay uncached)

https://claude.ai/code/session_01Lq2ejFV7X16HvAmg9BDGTu

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lq2ejFV7X16HvAmg9BDGTu)_

### DIFF
--- a/modules/incomingwebhook/api.go
+++ b/modules/incomingwebhook/api.go
@@ -87,6 +87,12 @@ type IncomingWebhook struct {
 	// env(DM_INCOMINGWEBHOOK_*) → code-default。admin 在管理台改值后 Reload 立即生效，
 	// 多实例 60s 内收敛，无需重启。其余阈值(IP/失败预算/floor/body/content)仍走 env。
 	settings *commonmod.SystemSettings
+	// webhookCache / groupCache 是 push 热路径的进程内短 TTL 缓存（#284 item 2）：
+	// 分别缓存 webhook 行与「群 Normal」结果，命中即 0 DB 读。变更路径（update/delete/
+	// regenerate/handleGroupDisband）即时失效本实例条目，跨实例由 TTL 兜底——staleness
+	// 契约见 cache.go。只缓存存在/Normal 的正向结果，不做负缓存。
+	webhookCache *ttlCache[*incomingWebhookModel]
+	groupCache   *ttlCache[*group.Model]
 }
 
 // maxConcurrentAudit 限制异步审计 goroutine 的最大并发数（默认值，可被 env 覆盖）。
@@ -130,15 +136,18 @@ func sharedRateRedis(cfg *config.Config) *redis.Client {
 
 // New 构造路由模块。
 func New(ctx *config.Context) *IncomingWebhook {
+	cacheTTL, cacheMax := cacheTTL(), cacheMax()
 	w := &IncomingWebhook{
-		ctx:       ctx,
-		Log:       log.NewTLog("IncomingWebhook"),
-		db:        newDB(ctx),
-		groupDB:   group.NewDB(ctx),
-		rateRedis: sharedRateRedis(ctx.GetConfig()),
-		auditSem:  make(chan struct{}, auditConcurrency()),
-		floor:     newLocalFloor(),
-		settings:  commonmod.EnsureSystemSettings(ctx),
+		ctx:          ctx,
+		Log:          log.NewTLog("IncomingWebhook"),
+		db:           newDB(ctx),
+		groupDB:      group.NewDB(ctx),
+		rateRedis:    sharedRateRedis(ctx.GetConfig()),
+		auditSem:     make(chan struct{}, auditConcurrency()),
+		floor:        newLocalFloor(),
+		settings:     commonmod.EnsureSystemSettings(ctx),
+		webhookCache: newTTLCache[*incomingWebhookModel](cacheTTL, cacheMax),
+		groupCache:   newTTLCache[*group.Model](cacheTTL, cacheMax),
 	}
 	// 群解散级联禁用所有 webhook
 	w.ctx.AddEventListener(event.GroupDisband, w.handleGroupDisband)
@@ -313,6 +322,42 @@ func (w *IncomingWebhook) requireActiveGroup(groupNo string) (*group.Model, erro
 	}
 	if g == nil || g.Status != group.GroupStatusNormal {
 		return nil, nil
+	}
+	return g, nil
+}
+
+// cachedQueryByWebhookID 是 push 热路径用的带缓存 webhook 点读：命中即 0 DB 读。
+// 只缓存存在的行（DB 故障与未命中都不写缓存，也不做负缓存——见 cache.go）。
+// 仅供 push 使用；管理写路径（update/delete/regenerate）仍走未缓存的 db.queryByWebhookID
+// 直读最新状态，避免基于陈旧快照做复活/越权判断。
+func (w *IncomingWebhook) cachedQueryByWebhookID(webhookID string) (*incomingWebhookModel, error) {
+	if m, ok := w.webhookCache.get(webhookID); ok {
+		return m, nil
+	}
+	m, err := w.db.queryByWebhookID(webhookID)
+	if err != nil {
+		return nil, err
+	}
+	if m != nil {
+		w.webhookCache.set(webhookID, m)
+	}
+	return m, nil
+}
+
+// cachedRequireActiveGroup 带缓存的群活跃校验，语义同 requireActiveGroup（群存在且
+// Normal 返回非 nil，否则 (nil,nil)）。只缓存 Normal 结果——非 Normal/不存在不缓存，
+// 既免去负缓存复杂度，也让"群刚解散"在 disband 失效/TTL 后立即走 DB 复核而非被粘住。
+// 仅供 push 使用；管理写路径仍走未缓存的 requireActiveGroup。
+func (w *IncomingWebhook) cachedRequireActiveGroup(groupNo string) (*group.Model, error) {
+	if g, ok := w.groupCache.get(groupNo); ok {
+		return g, nil
+	}
+	g, err := w.requireActiveGroup(groupNo)
+	if err != nil {
+		return nil, err
+	}
+	if g != nil {
+		w.groupCache.set(groupNo, g)
 	}
 	return g, nil
 }
@@ -512,6 +557,9 @@ func (w *IncomingWebhook) update(c *wkhttp.Context) {
 		mgmtOperationFailed(c)
 		return
 	}
+	// 本实例即时失效缓存：状态/名称/头像变更后，push 路径不得再命中旧快照（禁用尤其
+	// 关键——必须及时停推）。跨实例由 TTL 兜底。
+	w.webhookCache.invalidate(webhookID)
 	updated, qErr := w.db.queryByWebhookID(webhookID)
 	if qErr != nil || updated == nil {
 		// 回读失败/行消失：无法确认更新结果（可能已落库，也可能因并发软删除而落空），
@@ -544,6 +592,8 @@ func (w *IncomingWebhook) delete(c *wkhttp.Context) {
 		mgmtOperationFailed(c)
 		return
 	}
+	// 软删后即时失效本实例缓存，push 不得再命中旧的 enabled 快照继续推送。
+	w.webhookCache.invalidate(webhookID)
 	c.ResponseOK()
 }
 
@@ -578,6 +628,9 @@ func (w *IncomingWebhook) regenerate(c *wkhttp.Context) {
 		mgmtOperationFailed(c)
 		return
 	}
+	// token 轮换后即时失效本实例缓存：旧 token_hash 不得再被 push 命中（否则旧 token 在
+	// 本实例仍可推送）。跨实例 TTL 窗口内旧 token 仍短暂有效，是已接受的 staleness 契约。
+	w.webhookCache.invalidate(webhookID)
 	// 并发软删除竞态：updateFields 的 status != statusDeleted 守卫保证不会给已删除的
 	// webhook 写新 token_hash。回读确认行仍存活，避免向客户端返回一个实际未落库、
 	// 指向已删除行的"新 token"。
@@ -629,8 +682,9 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 		return
 	}
 
-	// 1) 查 webhook（queryByWebhookID 已把 ErrNotFound 吸收为 nil/nil）
-	m, err := w.db.queryByWebhookID(webhookID)
+	// 1) 查 webhook（cachedQueryByWebhookID 命中即 0 DB 读；未命中回落 db.queryByWebhookID，
+	//    后者已把 ErrNotFound 吸收为 nil/nil）
+	m, err := w.cachedQueryByWebhookID(webhookID)
 	if err != nil {
 		// 服务端故障，不是调用方扫描——绝不计入 IP 失败预算（否则 DB 抖动会误封 IP）。
 		w.Error("query webhook failed", zap.Error(err))
@@ -663,7 +717,7 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 	// 2.5) 群必须仍处于 Normal —— 兜底 handleGroupDisband 的异步窗口期，
 	// 也防止对已解散群继续推送消息。统一返回 401（响应体不区分原因——防探测的主防线；
 	// 时序非恒定，仅尽力而为，见 errcode/incomingwebhook.go 注释）。
-	g, err := w.requireActiveGroup(m.GroupNo)
+	g, err := w.cachedRequireActiveGroup(m.GroupNo)
 	if err != nil {
 		w.Error("query group on push failed",
 			zap.String("webhook_id", m.WebhookID), zap.Error(err))
@@ -868,6 +922,10 @@ func (w *IncomingWebhook) handleGroupDisband(data []byte, commit config.EventCom
 		w.Warn("disable webhooks on group disband failed",
 			zap.String("group_no", req.GroupNo), zap.Error(err))
 	}
+	// 即时失效本实例的群缓存：下次 push 的 cachedRequireActiveGroup 会 miss → 直查 DB →
+	// 群非 Normal → 拒绝。这一道就足以挡住本群所有 webhook（即便它们的 webhook 行缓存仍
+	// 短暂 stale 为 enabled，群闸也会拦下），无需按 webhookID 逐条失效。跨实例 TTL 兜底。
+	w.groupCache.invalidate(req.GroupNo)
 	// 故意 commit(nil)：disable 失败也不重试，避免阻塞事件队列。
 	// 异步窗口期由 push 路径的 requireActiveGroup 兜底（belt + suspenders）：
 	// 即便此处尚未把 webhook.status 改为 0，推送也会因群非 Normal 而 401。
--- a/modules/incomingwebhook/api.go
+++ b/modules/incomingwebhook/api.go
@@ -334,12 +334,15 @@ func (w *IncomingWebhook) cachedQueryByWebhookID(webhookID string) (*incomingWeb
 	if m, ok := w.webhookCache.get(webhookID); ok {
 		return m, nil
 	}
+	// 先捕获代际再读 DB：若读取期间有并发 invalidate（update/delete/regenerate），
+	// setIfGen 会丢弃这次回填，避免用变更前的旧行复活缓存（读-后-失效竞态）。
+	gen := w.webhookCache.loadGen()
 	m, err := w.db.queryByWebhookID(webhookID)
 	if err != nil {
 		return nil, err
 	}
 	if m != nil {
-		w.webhookCache.set(webhookID, m)
+		w.webhookCache.setIfGen(webhookID, m, gen)
 	}
 	return m, nil
 }
@@ -356,12 +359,14 @@ func (w *IncomingWebhook) cachedRequireActiveGroup(groupNo string) (*group.Model
 	if g, ok := w.groupCache.get(groupNo); ok {
 		return g, nil
 	}
+	// 同 cachedQueryByWebhookID：代际守卫关闭 disband 失效与在途 miss 读的回填竞态。
+	gen := w.groupCache.loadGen()
 	g, err := w.requireActiveGroup(groupNo)
 	if err != nil {
 		return nil, err
 	}
 	if g != nil {
-		w.groupCache.set(groupNo, g)
+		w.groupCache.setIfGen(groupNo, g, gen)
 	}
 	return g, nil
 }
--- a/modules/incomingwebhook/api.go
+++ b/modules/incomingwebhook/api.go
@@ -348,6 +348,10 @@ func (w *IncomingWebhook) cachedQueryByWebhookID(webhookID string) (*incomingWeb
 // Normal 返回非 nil，否则 (nil,nil)）。只缓存 Normal 结果——非 Normal/不存在不缓存，
 // 既免去负缓存复杂度，也让"群刚解散"在 disband 失效/TTL 后立即走 DB 复核而非被粘住。
 // 仅供 push 使用；管理写路径仍走未缓存的 requireActiveGroup。
+//
+// ⚠️ 群【管理员禁用】(Normal→Disabled, event.GroupUpdate) 不在失效矩阵内：admin 禁用后
+// 群闸在所有实例上最多 stale 一个 TTL 才生效（解散是即时的）。这是经维护者确认接受的
+// 取舍，详见 cache.go 顶部契约注释；TestPush_GroupAdminDisable_TTLBounded 钉住此语义。
 func (w *IncomingWebhook) cachedRequireActiveGroup(groupNo string) (*group.Model, error) {
 	if g, ok := w.groupCache.get(groupNo); ok {
 		return g, nil
--- a/modules/incomingwebhook/api_i18n_test.go
+++ b/modules/incomingwebhook/api_i18n_test.go
@@ -19,7 +19,7 @@ import (
 // #246 note in api_i18n.go) don't trip the guard. Add any new handler/limiter
 // file to the list below.
 func TestIncomingWebhookNoLegacyResponseError(t *testing.T) {
-	files := []string{"api.go", "api_i18n.go", "ratelimit.go", "localfloor.go"}
+	files := []string{"api.go", "api_i18n.go", "ratelimit.go", "localfloor.go", "cache.go"}
 	banned := []string{
 		".ResponseError(",
 		".ResponseErrorf(",
--- a/modules/incomingwebhook/api_test.go
+++ b/modules/incomingwebhook/api_test.go
@@ -776,6 +776,64 @@ func TestPush_ValidPushesNotIPLimited(t *testing.T) {
 	}
 }
 
+// TestPush_CacheServesStaleWithinTTL_HandlerInvalidates verifies the push
+// hot-path cache (#284 item 2):
+//   - a warm cache serves the webhook record, so a DB-only status change (made
+//     behind the handler's back) does NOT stop pushes within the TTL window;
+//   - a delete THROUGH the handler invalidates the cache and stops pushes at once.
+//
+// Asserts on the auth outcome (401 vs not-401) rather than 200, so it is robust
+// to whether the downstream send returns 200/502 in the test harness. All
+// limiters are set generous so a 429 can't be mistaken for the auth gate.
+func TestPush_CacheServesStaleWithinTTL_HandlerInvalidates(t *testing.T) {
+	// Long TTL so the cached entry never expires mid-test; cache TTL is read at
+	// module construction, so it must be set before setupTestEnv.
+	t.Setenv("DM_INCOMINGWEBHOOK_CACHE_TTL_MS", "60000")
+	t.Setenv("DM_INCOMINGWEBHOOK_RPS", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_BURST", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_IP_RPS", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_IP_BURST", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_LOCAL_RPS", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_LOCAL_BURST", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_IP_FAIL_RPS", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_IP_FAIL_BURST", "1000")
+
+	handler, ctx, groupNo := setupTestEnv(t)
+	const ip = "203.0.113.77"
+	resetIPFailBucket(t, ctx, ip)
+	resetStrictIPBucket(t, ctx, ip)
+
+	w := do(handler, authReq("POST", fmt.Sprintf("/v1/groups/%s/incoming-webhooks", groupNo), map[string]interface{}{
+		"name": "cache-wh",
+	}))
+	created := parseJSON(t, w)
+	whID := created["webhook_id"].(string)
+	token := created["token"].(string)
+	url := fmt.Sprintf("/v1/incoming-webhooks/%s/%s", whID, token)
+	body, _ := json.Marshal(pushReq{Content: "hi"})
+
+	// 1) Warm the cache (status=enabled). Auth passes → not 401.
+	w = do(handler, anonReqIP("POST", url, body, ip))
+	assert.NotEqualf(t, http.StatusUnauthorized, w.Code, "warm push must authorize; body=%s", w.Body.String())
+
+	// 2) Disable directly in DB, bypassing the handler so the cache is NOT
+	//    invalidated. The warm cache still serves enabled → push still authorizes
+	//    (non-401) within the TTL window — proving the cache is in effect.
+	_, err := ctx.DB().UpdateBySql("UPDATE incoming_webhook SET status=0 WHERE webhook_id=?", whID).Exec()
+	assert.NoError(t, err)
+	w = do(handler, anonReqIP("POST", url, body, ip))
+	assert.NotEqualf(t, http.StatusUnauthorized, w.Code,
+		"within TTL the cached (enabled) record must still authorize despite the DB-only disable; body=%s", w.Body.String())
+
+	// 3) Delete through the handler → invalidates the cache. Next push misses the
+	//    cache, reads the soft-deleted row, and is rejected (401).
+	w = do(handler, authReq("DELETE", fmt.Sprintf("/v1/groups/%s/incoming-webhooks/%s", groupNo, whID), nil))
+	assert.Equalf(t, http.StatusOK, w.Code, "delete; body=%s", w.Body.String())
+	w = do(handler, anonReqIP("POST", url, body, ip))
+	assert.Equalf(t, http.StatusUnauthorized, w.Code,
+		"after handler delete invalidates the cache, push must be rejected; body=%s", w.Body.String())
+}
+
 // TestPush_ValidPushesCappedByPerIPRequestLimit locks the layer the reviewers
 // asked to keep: the per-IP REQUEST limiter (StrictIPRateLimitMiddleware) bounds
 // ALL requests from one IP — including valid pushes — so a single IP holding many
--- a/modules/incomingwebhook/api_test.go
+++ b/modules/incomingwebhook/api_test.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/Mininglamp-OSS/octo-lib/config"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
@@ -832,6 +833,57 @@ func TestPush_CacheServesStaleWithinTTL_HandlerInvalidates(t *testing.T) {
 	w = do(handler, anonReqIP("POST", url, body, ip))
 	assert.Equalf(t, http.StatusUnauthorized, w.Code,
 		"after handler delete invalidates the cache, push must be rejected; body=%s", w.Body.String())
+}
+
+// TestPush_GroupAdminDisable_TTLBounded pins the deliberately-accepted staleness
+// for administrative group-disable (#293 review): the module invalidates
+// groupCache only on disband, NOT on admin disable (Normal→Disabled, emitted as
+// event.GroupUpdate, which this module does not subscribe to). So after a group is
+// disabled directly, the cached Normal entry keeps the push auth gate open until
+// the TTL expires, then the gate re-reads the DB and rejects. A short TTL makes the
+// expiry observable; this is the documented trade-off, not immediate enforcement.
+func TestPush_GroupAdminDisable_TTLBounded(t *testing.T) {
+	t.Setenv("DM_INCOMINGWEBHOOK_CACHE_TTL_MS", "2000") // short but generous vs slow CI
+	t.Setenv("DM_INCOMINGWEBHOOK_RPS", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_BURST", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_IP_RPS", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_IP_BURST", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_LOCAL_RPS", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_LOCAL_BURST", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_IP_FAIL_RPS", "1000")
+	t.Setenv("DM_INCOMINGWEBHOOK_IP_FAIL_BURST", "1000")
+
+	handler, ctx, groupNo := setupTestEnv(t)
+	const ip = "203.0.113.91"
+	resetIPFailBucket(t, ctx, ip)
+	resetStrictIPBucket(t, ctx, ip)
+
+	w := do(handler, authReq("POST", fmt.Sprintf("/v1/groups/%s/incoming-webhooks", groupNo), map[string]interface{}{
+		"name": "grp-wh",
+	}))
+	created := parseJSON(t, w)
+	url := fmt.Sprintf("/v1/incoming-webhooks/%s/%s", created["webhook_id"].(string), created["token"].(string))
+	body, _ := json.Marshal(pushReq{Content: "hi"})
+
+	// Warm the group cache (Normal) → push authorizes.
+	pw := do(handler, anonReqIP("POST", url, body, ip))
+	assert.NotEqualf(t, http.StatusUnauthorized, pw.Code, "warm push must authorize; body=%s", pw.Body.String())
+
+	// Admin-disable the group directly (Normal→Disabled, GroupStatusDisabled=0),
+	// no disband event/cascade — exactly the path this module does not invalidate.
+	_, err := ctx.DB().UpdateBySql("UPDATE `group` SET status=? WHERE group_no=?", 0, groupNo).Exec()
+	assert.NoError(t, err)
+
+	// Within the TTL: the cached Normal entry still authorizes (documented staleness).
+	pw = do(handler, anonReqIP("POST", url, body, ip))
+	assert.NotEqualf(t, http.StatusUnauthorized, pw.Code,
+		"within TTL the cached Normal group must still authorize after a DB-only disable; body=%s", pw.Body.String())
+
+	// After the TTL: the group gate re-reads the DB, sees non-Normal, rejects (401).
+	time.Sleep(2500 * time.Millisecond) // > TTL(2000ms)
+	pw = do(handler, anonReqIP("POST", url, body, ip))
+	assert.Equalf(t, http.StatusUnauthorized, pw.Code,
+		"after the TTL the admin-disabled group must reject push; body=%s", pw.Body.String())
 }
 
 // TestPush_ValidPushesCappedByPerIPRequestLimit locks the layer the reviewers
--- a/modules/incomingwebhook/cache.go
+++ b/modules/incomingwebhook/cache.go
@@ -1,0 +1,119 @@
+package incomingwebhook
+
+import (
+	"os"
+	"strconv"
+	"sync"
+	"time"
+)
+
+// push 热路径缓存（#284 item 2）。push 在限流与发送前有两次未缓存的 DB 点读：
+// webhook 行（queryByWebhookID）+ 群状态（requireActiveGroup）。二者近乎不可变 /
+// 极少变更，却是任意 webhook 高 QPS 推送时的第一道 DB 读墙。本缓存是进程内、短 TTL、
+// 不依赖 Redis 的（与 localFloor 一脉相承——Redis 故障也能命中），命中即 0 DB 读。
+//
+// ⚠️ 鉴权 staleness 契约（#284 验收明确接受秒级 staleness）：
+//   - disable / delete / regenerate 会在【本实例】即时 invalidate 对应 webhook 条目；
+//     群解散在【本实例】即时 invalidate 群条目。
+//   - 跨实例没有主动失效，最多 stale 一个 TTL：刚被禁用/删除/改 token 的 webhook、或
+//     刚解散的群，在 TTL 窗口内对等实例上可能仍按旧状态放行。TTL 默认很短（3s）以把这
+//     个窗口压到秒级。把 TTL 设为 0（DM_INCOMINGWEBHOOK_CACHE_TTL_MS=0）可彻底关闭缓存，
+//     退化为每次直查 DB 的旧行为。
+const (
+	envCacheTTLMs = "DM_INCOMINGWEBHOOK_CACHE_TTL_MS"
+	envCacheMax   = "DM_INCOMINGWEBHOOK_CACHE_MAX"
+
+	// 默认 3s：push 鉴权闸可容忍的 staleness 窗口（秒级，见上）。
+	defaultCacheTTL = 3 * time.Second
+	// 条目数上限：超过则整桶清空（粗粒度淘汰）。活跃推送的 webhook/group 工作集很小，
+	// 正常远不触顶；上限只防异常场景的无界增长。**不做负缓存**——不存在/已删的 webhookID
+	// 扫描由 per-IP 失败预算在打 DB 前拦截（#285），缓存它们只会被扫描流量污染。
+	defaultCacheMax = 10000
+)
+
+// cacheTTL 读 DM_INCOMINGWEBHOOK_CACHE_TTL_MS（毫秒）；0 表示禁用缓存，缺省/非法回退默认。
+func cacheTTL() time.Duration {
+	if v := os.Getenv(envCacheTTLMs); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n >= 0 {
+			return time.Duration(n) * time.Millisecond
+		}
+	}
+	return defaultCacheTTL
+}
+
+// cacheMax 读 DM_INCOMINGWEBHOOK_CACHE_MAX（条目数上限）；仅接受正整数，否则回退默认。
+func cacheMax() int {
+	if v := os.Getenv(envCacheMax); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n > 0 {
+			return n
+		}
+	}
+	return defaultCacheMax
+}
+
+type cacheEntry[T any] struct {
+	val T
+	exp time.Time
+}
+
+// ttlCache 是 push 热路径用的进程内短 TTL 缓存：mutex map + 惰性过期 + 容量上限。
+// 泛型以同一实现服务 webhook 行与群状态两种值。ttl<=0 视为禁用（所有方法 no-op，
+// get 永远 miss），从而让 DM_INCOMINGWEBHOOK_CACHE_TTL_MS=0 等价于"无缓存"。
+type ttlCache[T any] struct {
+	mu      sync.Mutex
+	ttl     time.Duration
+	maxSize int
+	m       map[string]cacheEntry[T]
+}
+
+func newTTLCache[T any](ttl time.Duration, maxSize int) *ttlCache[T] {
+	return &ttlCache[T]{ttl: ttl, maxSize: maxSize, m: make(map[string]cacheEntry[T])}
+}
+
+// enabled 在 nil 接收者或 ttl<=0 时返回 false（nil 检查短路，方法整体 nil-safe）。
+func (c *ttlCache[T]) enabled() bool { return c != nil && c.ttl > 0 }
+
+// get 返回未过期条目；过期则惰性删除并 miss。
+func (c *ttlCache[T]) get(key string) (T, bool) {
+	var zero T
+	if !c.enabled() {
+		return zero, false
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	e, ok := c.m[key]
+	if !ok {
+		return zero, false
+	}
+	if time.Now().After(e.exp) {
+		delete(c.m, key)
+		return zero, false
+	}
+	return e.val, true
+}
+
+// set 写入并打 TTL 戳。超过容量上限且是新键时整桶清空（粗粒度淘汰：工作集小、正常不
+// 触发；触发时最坏退化为下一轮重填，不影响正确性）。
+func (c *ttlCache[T]) set(key string, val T) {
+	if !c.enabled() {
+		return
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if len(c.m) >= c.maxSize {
+		if _, exists := c.m[key]; !exists {
+			c.m = make(map[string]cacheEntry[T], c.maxSize)
+		}
+	}
+	c.m[key] = cacheEntry[T]{val: val, exp: time.Now().Add(c.ttl)}
+}
+
+// invalidate 删除单个键（变更路径的即时失效入口）。
+func (c *ttlCache[T]) invalidate(key string) {
+	if !c.enabled() {
+		return
+	}
+	c.mu.Lock()
+	delete(c.m, key)
+	c.mu.Unlock()
+}
--- a/modules/incomingwebhook/cache.go
+++ b/modules/incomingwebhook/cache.go
@@ -15,6 +15,8 @@ import (
 // ⚠️ 鉴权 staleness 契约（#284 验收明确接受秒级 staleness）：
 //   - webhook 变更（disable / delete / regenerate）会在【本实例】即时 invalidate 对应
 //     webhook 条目；群【解散】(event.GroupDisband) 会在【本实例】即时 invalidate 群条目。
+//     「即时」是严格的：invalidate 自增代际，任何在变更前就开始、读到旧行的在途 miss 读
+//     都会在 setIfGen 时因代际变化被丢弃，不会把旧快照回填回来（读-后-失效竞态已关闭）。
 //   - 跨实例没有主动失效，最多 stale 一个 TTL：上述刚变更的 webhook 或刚解散的群，在 TTL
 //     窗口内对等实例上可能仍按旧状态放行。
 //   - **群【管理员禁用】(GroupStatusNormal→Disabled, 走 event.GroupUpdate) 不在失效矩阵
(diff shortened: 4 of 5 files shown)


Jerry-Xin (changes requested):
The PR is relevant to `octo-server`, but it introduces a cache invalidation gap for the group Normal gate.

**🔴 Blocking**
- 🔴 Critical: `cachedRequireActiveGroup` caches a Normal group result, but the cache is only invalidated on `GroupDisband` (`modules/incomingwebhook/api.go:335`, `modules/incomingwebhook/api.go:912`). Group status can also transition from Normal to Disabled through the group manager path (`modules/group/api_manager.go:224`, `modules/group/api_manager.go:249`, `modules/group/api_manager.go:291`). Before this PR, `push` re-read the group row and rejected non-Normal groups immediately; after this PR, a warmed `groupCache` lets pushes continue until TTL expiry on the same instance. This is a permission/state bypass for disabled groups. Please invalidate `groupCache` on relevant `GroupUpdate` status changes, or otherwise make the push path re-check disabled transitions.

**💬 Non-blocking**
- 🟡 Warning: The implementation says it avoids negative caching, but `cachedQueryByWebhookID` stores any non-nil row (`modules/incomingwebhook/api.go:321` to `modules/incomingwebhook/api.go:326`), including `statusDeleted` and `statusDisabled` rows returned by `queryByWebhookID`. That contradicts the stated contract in `modules/incomingwebhook/cache.go:28` to `modules/incomingwebhook/cache.go:30` and can let deleted/disabled IDs occupy the cache. Consider caching only `statusEnabled` rows, or update the contract and tests.

**✅ Highlights**
- The cache is local, Redis-independent, and guarded by a simple TTL/max-size mechanism.
- Management paths still read webhook rows directly, which preserves the existing stale-snapshot protections for update/delete/regenerate decisions.
- Unit cache tests passed locally: `go test ./modules/incomingwebhook -run 'TestTTLCache|TestCache|TestCached'`.
- Integration tests could not run here because MySQL was unavailable: `dial tcp 127.0.0.1:3306: connect: connection refused`.

Jerry-Xin (dismissed):
The PR is relevant to `octo-server`, but it introduces a cache invalidation gap for the group Normal gate.

**🔴 Blocking**
- 🔴 Critical: `cachedRequireActiveGroup` caches a Normal group result, but the cache is only invalidated on `GroupDisband` (`modules/incomingwebhook/api.go:335`, `modules/incomingwebhook/api.go:912`). Group status can also transition from Normal to Disabled through the group manager path (`modules/group/api_manager.go:224`, `modules/group/api_manager.go:249`, `modules/group/api_manager.go:291`). Before this PR, `push` re-read the group row and rejected non-Normal groups immediately; after this PR, a warmed `groupCache` lets pushes continue until TTL expiry on the same instance. This is a permission/state bypass for disabled groups. Please invalidate `groupCache` on relevant `GroupUpdate` status changes, or otherwise make the push path re-check disabled transitions.

**💬 Non-blocking**
- 🟡 Warning: The implementation says it avoids negative caching, but `cachedQueryByWebhookID` stores any non-nil row (`modules/incomingwebhook/api.go:321` to `modules/incomingwebhook/api.go:326`), including `statusDeleted` and `statusDisabled` rows returned by `queryByWebhookID`. That contradicts the stated contract in `modules/incomingwebhook/cache.go:28` to `modules/incomingwebhook/cache.go:30` and can let deleted/disabled IDs occupy the cache. Consider caching only `statusEnabled` rows, or update the contract and tests.

**✅ Highlights**
- The cache is local, Redis-independent, and guarded by a simple TTL/max-size mechanism.
- Management paths still read webhook rows directly, which preserves the existing stale-snapshot protections for update/delete/regenerate decisions.
- Unit cache tests passed locally: `go test ./modules/incomingwebhook -run 'TestTTLCache|TestCache|TestCached'`.
- Integration tests could not run here because MySQL was unavailable: `dial tcp 127.0.0.1:3306: connect: connection refused`.

Jerry-Xin (dismissed):
The PR is relevant to `octo-server`, but it introduces a cache invalidation gap for the group Normal gate.

**🔴 Blocking**
- 🔴 Critical: `cachedRequireActiveGroup` caches a Normal group result, but the cache is only invalidated on `GroupDisband` (`modules/incomingwebhook/api.go:335`, `modules/incomingwebhook/api.go:912`). Group status can also transition from Normal to Disabled through the group manager path (`modules/group/api_manager.go:224`, `modules/group/api_manager.go:249`, `modules/group/api_manager.go:291`). Before this PR, `push` re-read the group row and rejected non-Normal groups immediately; after this PR, a warmed `groupCache` lets pushes continue until TTL expiry on the same instance. This is a permission/state bypass for disabled groups. Please invalidate `groupCache` on relevant `GroupUpdate` status changes, or otherwise make the push path re-check disabled transitions.

**💬 Non-blocking**
- 🟡 Warning: The implementation says it avoids negative caching, but `cachedQueryByWebhookID` stores any non-nil row (`modules/incomingwebhook/api.go:321` to `modules/incomingwebhook/api.go:326`), including `statusDeleted` and `statusDisabled` rows returned by `queryByWebhookID`. That contradicts the stated contract in `modules/incomingwebhook/cache.go:28` to `modules/incomingwebhook/cache.go:30` and can let deleted/disabled IDs occupy the cache. Consider caching only `statusEnabled` rows, or update the contract and tests.

**✅ Highlights**
- The cache is local, Redis-independent, and guarded by a simple TTL/max-size mechanism.
- Management paths still read webhook rows directly, which preserves the existing stale-snapshot protections for update/delete/regenerate decisions.
- Unit cache tests passed locally: `go test ./modules/incomingwebhook -run 'TestTTLCache|TestCache|TestCached'`.
- Integration tests could not run here because MySQL was unavailable: `dial tcp 127.0.0.1:3306: connect: connection refused`.

Jerry-Xin (dismissed):
The PR is relevant to `octo-server`, but it introduces a cache invalidation gap for the group Normal gate.

**🔴 Blocking**
- 🔴 Critical: `cachedRequireActiveGroup` caches a Normal group result, but the cache is only invalidated on `GroupDisband` (`modules/incomingwebhook/api.go:335`, `modules/incomingwebhook/api.go:912`). Group status can also transition from Normal to Disabled through the group manager path (`modules/group/api_manager.go:224`, `modules/group/api_manager.go:249`, `modules/group/api_manager.go:291`). Before this PR, `push` re-read the group row and rejected non-Normal groups immediately; after this PR, a warmed `groupCache` lets pushes continue until TTL expiry on the same instance. This is a permission/state bypass for disabled groups. Please invalidate `groupCache` on relevant `GroupUpdate` status changes, or otherwise make the push path re-check disabled transitions.

**💬 Non-blocking**
- 🟡 Warning: The implementation says it avoids negative caching, but `cachedQueryByWebhookID` stores any non-nil row (`modules/incomingwebhook/api.go:321` to `modules/incomingwebhook/api.go:326`), including `statusDeleted` and `statusDisabled` rows returned by `queryByWebhookID`. That contradicts the stated contract in `modules/incomingwebhook/cache.go:28` to `modules/incomingwebhook/cache.go:30` and can let deleted/disabled IDs occupy the cache. Consider caching only `statusEnabled` rows, or update the contract and tests.

**✅ Highlights**
- The cache is local, Redis-independent, and guarded by a simple TTL/max-size mechanism.
- Management paths still read webhook rows directly, which preserves the existing stale-snapshot protections for update/delete/regenerate decisions.
- Unit cache tests passed locally: `go test ./modules/incomingwebhook -run 'TestTTLCache|TestCache|TestCached'`.
- Integration tests could not run here because MySQL was unavailable: `dial tcp 127.0.0.1:3306: connect: connection refused`.

yujiawei (approved):
# Code Review — PR #293 (octo-server)

**Scope:** `modules/incomingwebhook/{api.go,cache.go,*_test.go}` — a per-instance, short-TTL, Redis-independent cache for the two uncached DB point reads on the push hot path (`queryByWebhookID` + group-Normal check). Reviewed at head `2895ad1`. This is a security-sensitive PR (push auth gate), so the focus below is on the staleness/auth contract.

## Summary

The implementation is clean and well-scoped. The generic `ttlCache[T]` is correct (mutex map + lazy expiry + nil-safe + `ttl<=0` disable), positive-only caching avoids negative-cache pollution, and the read paths return on cache hit without touching the DB. Concurrency is safe: the shared cached `*incomingWebhookModel` / `*group.Model` pointers are only ever **read** on the push path (`buildPayload`, status/token checks, `markUsed`), never mutated, so handing the same pointer to concurrent push goroutines is race-free. `go build`, `go vet`, and the infra-free tests pass locally.

The webhook-side invalidation matrix (`update` / `delete` / `regenerate`) is correct, and the disband path correctly invalidates the group cache so the group gate alone shuts down a disbanded group even while its webhook rows are briefly cached as enabled.

I'm approving, but there is one staleness gap that diverges from the PR's stated contract and from pre-change behavior. It is TTL-bounded (consistent with the design's accepted staleness tolerance), so it is not a merge blocker — but given the security-sensitive classification I want a human to consciously sign off on it.

## Findings

### P2 — Administrative group-disable is not invalidated; push gate stays stale for up to TTL

The PR description states: *"Management writes keep reading the DB uncached, so revival/permission decisions never run on a stale snapshot."* That holds for **webhook** management writes and for group **disband**, but not for **administrative group disable**.

- `cachedRequireActiveGroup` only authorizes when group `Status == Normal` and caches the Normal result (`api.go:335-347`).
- The module invalidates `groupCache` in exactly one place — `handleGroupDisband` (`api.go:909`) — which fires on `event.GroupDisband`. The module subscribes to **only** that event (`api.go:149`).
- The admin "disable group" flow (`modules/group/api_manager.go:220-290`) sets `group.Status = GroupStatusDisabled` and emits **`event.GroupUpdate`**, not `GroupDisband`. The `incomingwebhook` module does not listen to `GroupUpdate`, so **no instance** (not even the one whose admin action disabled the group) invalidates `groupCache`.

**Effect:** Pre-change, disabling a group stopped pushes on the *next* push (each push did a live `QueryWithGroupNo`). Post-change, a just-disabled group keeps authorizing pushes for up to the TTL (default 3000 ms) on every instance, because the cached `Normal` entry is served until it lazily expires. This is the only enforcement mechanism for admin-disable on the push path — there is no `disableByGroupNo` cascade for admin-disable (that is disband-only, `db.go:162` / `api.go:905`), so the group-status gate was the entire check, and it is now TTL-delayed.

Why this is P2 and not a blocker:
- The window is bounded to TTL (default 3s), which is exactly the staleness the design explicitly accepts for the auth gate, with `DM_INCOMINGWEBHOOK_CACHE_TTL_MS=0` as the escape hatch.
- The destructive path (disband) *is* handled.
- A disabled group has its IM channel banned (`Ban=1` in the same flow), which likely blocks actual message delivery downstream even if the auth gate briefly passes.

**Recommended fix (pick one):**
1. Subscribe to `event.GroupUpdate` and `groupCache.invalidate(groupNo)` when the update is a status change (mirrors the disband handler), restoring pre-change immediacy for admin-disable; **or**
2. If TTL-bounded admin-disable staleness is acceptable, soften the contract claim in the PR/`cache.go` to explicitly enumerate admin group-disable as a TTL-bounded case (so the "never run on a stale snapshot" wording isn't read as covering it).

**Human verification requested:** please confirm whether administrative group-disable is expected to take effect immediately on the push path (moderation/abuse-ban scenarios) or whether a TTL-bounded delay is acceptable here as it is elsewhere.

### Nits (non-blocking)

- **Coarse size-cap eviction** (`cache.go:set`): when `len(m) >= maxSize` and a *new* key arrives, the whole bucket is cleared. Correctness is preserved (worst case is a refill round), and the default cap (10000) is far above the realistic working set, so this is fine — but under sustained churn beyond `maxSize` the cache degrades to ~no-cache via repeated full clears. The comment already calls this out; flagging only so it's a conscious choice.
- **Config read at construction only** (`cacheTTL()` / `cacheMax()` in `New`): TTL/max are not hot-reloadable. Consistent with the rest of the module's env config, so acceptable.
- **Soft-deleted rows are cached as existing rows**: `cachedQueryByWebhookID` caches any non-nil row, including `statusDeleted`/`statusDisabled` ones. This is harmless — the push path re-checks `m.Status` after the cache read and rejects deleted/disabled rows — and matches the "positive-only (existing row)" intent. No change needed; noting it because the description's "soft-deleted webhook IDs are never negative-cached" could be misread as "soft-deleted rows are never cached" (they are; they're just still rejected by the status gate).

## Tests

Good coverage. `ttlCache` mechanics, env parsing, and the hit-skips-DB helpers (with `db` nil to prove 0 DB reads) are solid and infra-free. The end-to-end `TestPush_CacheServesStaleWithinTTL_HandlerInvalidates` cleanly proves the serve-stale-then-invalidate behavior. Consider (follow-up, not blocking) an equivalent test asserting the admin group-disable path once its invalidation behavior is decided.

## Verdict

**APPROVED.** No P0/P1. One P2 staleness gap (admin group-disable not invalidated → TTL-bounded stale auth) that is consistent with the design's accepted staleness model and the disband path is covered, plus minor nits. Flagged the admin-disable behavior for explicit human sign-off given the security-sensitive classification.

yujiawei (changes requested):
# Code Review — PR #293 (octo-server)

> Correcting my earlier review: I am revising the verdict from APPROVED to **CHANGES_REQUESTED**. On closer analysis the group-cache invalidation gap is a regression of the push auth gate (not merely TTL-bounded staleness within the accepted model), and it contradicts the PR's own stated contract. An independent reviewer reached the same blocking conclusion.

**Scope:** `modules/incomingwebhook/{api.go,cache.go,*_test.go}` — a per-instance, short-TTL, Redis-independent cache for the two uncached DB point reads on the push hot path (`queryByWebhookID` + group-Normal check). Reviewed at head `2895ad1`. Security-sensitive PR (push auth gate).

## Summary

The implementation is otherwise clean and well-scoped. The generic `ttlCache[T]` is correct (mutex map + lazy expiry + nil-safe + `ttl<=0` disable), the read helpers skip the DB on a hit, and the shared cached pointers are read-only on the push path (race-free). The webhook-side invalidation matrix (`update`/`delete`/`regenerate`) and the group-disband invalidation are correct. `go build`, `go vet`, and the infra-free tests pass locally.

There is one blocking correctness gap on the group auth gate.

## Findings

### P1 (Blocking) — Administrative group-disable is never invalidated → stale push authorization for disabled groups

`cachedRequireActiveGroup` caches a `Normal` group result (`api.go:335-347`). The module invalidates `groupCache` in exactly one place — `handleGroupDisband` (`api.go:909`) — and subscribes to only `event.GroupDisband` (`api.go:149`).

But a group can also leave `Normal` via the **admin disable** path: `modules/group/api_manager.go:220-290` sets `group.Status = GroupStatusDisabled` and emits **`event.GroupUpdate`** (line ~275), which the `incomingwebhook` module does not listen to. There is also no `disableByGroupNo` cascade for admin-disable (that is disband-only, `db.go:162` / `api.go:905`), so the group-status check was the *entire* push gate for a disabled group.

Why this is blocking rather than the accepted TTL staleness:
- The PR's accepted staleness model invalidates the **acting/local** instance *immediately* and bounds only **peer** instances by TTL. For admin-disable, **no** instance invalidates — not even the one that performed the disable. So this is not "peers lag by a TTL"; it is "every instance keeps authorizing a disabled group until its cached entry happens to expire."
- This is a regression: pre-change, `push` re-read the group row every time and rejected a non-Normal group on the very next push. Post-change, a warm `groupCache` authorizes pushes to a just-disabled group for up to the TTL.
- It contradicts the PR's own contract: *"Management writes keep reading the DB uncached, so revival/permission decisions never run on a stale snapshot."* Admin group-disable is a management/permission write that now does run on a stale snapshot.
- It is on the push **auth gate** in a security-sensitive PR. Admin-disable is typically a moderation/abuse action where prompt enforcement matters.

**Required fix (either):**
1. Subscribe to `event.GroupUpdate` and `groupCache.invalidate(groupNo)` on status-changing updates (mirror the disband handler). This restores pre-change immediacy on the acting instance and bounds peers by TTL — consistent with the rest of the invalidation matrix; **or**
2. If a TTL-bounded delay for admin-disable is genuinely acceptable to the maintainers, then at minimum (a) narrow the contract wording in the PR and `cache.go` so it no longer claims management/permission writes never run on a stale snapshot, and (b) add a test pinning the chosen behavior — so this is a documented, deliberate trade-off rather than an unhandled event.

**Human verification requested:** please confirm the intended push-path behavior when an admin disables a group (immediate vs. TTL-bounded). The fix is small (one event listener), so option 1 is preferred.

### P2 (Non-blocking) — Contract wording vs. caching of non-enabled rows

`cachedQueryByWebhookID` caches any non-nil row (`api.go:321-326`), including `statusDeleted` / `statusDisabled` rows. This is *functionally safe* — the push path re-checks `m.Status` after the cache read and rejects deleted/disabled rows — but it diverges from the `cache.go:28-30` "positive-only / never negative-cache" wording, which can be misread as "non-enabled rows are never cached." Either cache only `statusEnabled` rows, or tighten the comment to say "existing rows are cached; the status gate still rejects deleted/disabled on every read."

### Nits (non-blocking)

- **Coarse size-cap eviction** (`cache.go:set`): full-bucket clear on `len>=max` + new key. Correct (worst case is a refill), default cap (10000) far above the realistic working set; flagging only as a conscious choice under sustained churn.
- **Config read at construction only** (`cacheTTL()`/`cacheMax()` in `New`): not hot-reloadable; consistent with the module's other env config, acceptable.

## Tests

`ttlCache` mechanics, env parsing, and the hit-skips-DB helpers (db nil → proves 0 DB reads) are solid and infra-free. `TestPush_CacheServesStaleWithinTTL_HandlerInvalidates` cleanly proves serve-stale-then-invalidate for the *webhook* path. Please add an equivalent assertion for the admin group-disable path as part of the fix above.

## Verdict

**CHANGES_REQUESTED** — one P1 blocking gap: admin group-disable bypasses the push auth gate for up to the cache TTL because the module never invalidates `groupCache` on `GroupUpdate`, contradicting the PR's stated contract and regressing pre-change immediate enforcement. Plus one P2 wording/contract mismatch and minor nits. The blocking fix is a single event subscription.

an9xyz:
Thanks both — pushed `59261f6`. On the admin group-disable gap, the maintainer chose **your option 2** (accept the TTL-bounded staleness + narrow the contract + pin it with a test) rather than the `event.GroupUpdate` listener.

Rationale for treating this as acceptable rather than a blocker:
- The **destructive** path (disband) stays immediate; admin-disable is a reversible operational action where seconds-level delay is tolerable.
- A disabled group's IM channel is `Ban`'d in the same `api_manager.go` flow, so downstream message delivery is blocked even if the auth gate briefly passes.
- `DM_INCOMINGWEBHOOK_CACHE_TTL_MS=0` forces immediate behavior for deployments that want it.

What changed:
- Tightened the `cache.go` contract to **explicitly enumerate admin group-disable as an accepted TTL-bounded case**, distinct from the "local-immediate, peer-TTL" webhook/disband cases — so the "permission decisions never run on a stale snapshot" wording no longer overclaims.
- Added `TestPush_GroupAdminDisable_TTLBounded`: a DB-only group disable still authorizes within the TTL, then is rejected after expiry.
- Clarified the "no negative cache" wording (your 🟡): it means *misses* aren't cached; existing disabled/deleted rows **are** cached but re-checked by the `m.Status` gate on every read, so the cache never bypasses it.

If you'd rather have immediate enforcement after all, the listener is a one-line follow-up (subscribe `event.GroupUpdate` → `groupCache.invalidate`) — I left a breadcrumb in the contract comment.

---

Jerry-Xin (dismissed):
The PR is in scope for `octo-server` and the cache integration looks correct for the incoming webhook push hot path.

💬 Non-blocking

🟡 Warning: [modules/incomingwebhook/api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-vmmpnv8h/repo/modules/incomingwebhook/api.go:325) caches every existing webhook row returned from DB, including disabled or soft-deleted rows. The push path rechecks `m.Status`, so I do not see an auth bypass, but this is broader than “positive-only active webhook” caching. Either keep the current behavior explicitly documented outside the code comments, or restrict caching to `statusEnabled` if scan/cache-pollution avoidance is the goal.

🔵 Suggestion: [modules/incomingwebhook/cache.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-vmmpnv8h/repo/modules/incomingwebhook/cache.go:112) has no request coalescing on cache miss. Under high concurrency, each TTL expiry can still produce a small DB stampede before the first goroutine refills the cache. Not required for this PR, but `singleflight` would make the “hot path” protection more robust.

🔵 Suggestion: [modules/incomingwebhook/api_test.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-vmmpnv8h/repo/modules/incomingwebhook/api_test.go:828) uses fixed sleep to observe TTL expiry. This is acceptable, but `require.Eventually`/polling would reduce CI flakiness while still proving the same contract.

✅ Highlights

The local invalidation points are placed after successful DB mutations for update, delete, and token regeneration, so the writing instance stops serving stale webhook snapshots once the write succeeds.

The group cache stores only `Normal` group results and is invalidated on `GroupDisband`, preserving the intended short-TTL staleness contract for the push gate.

The cache is Redis-independent, bounded, disableable via TTL `0`, and covered by focused unit tests. I ran:

```bash
go test ./modules/incomingwebhook -run 'TestTTLCache|TestCacheTTL|TestCacheMax|TestCached'
go test ./modules/incomingwebhook -run TestIncomingWebhookNoLegacyResponseError
```

Both passed.

lml2468 (dismissed):
APPROVED (head `59261f6359c2`).

Maintainer chose to accept TTL-bounded staleness for admin group-disable rather than add a `event.GroupUpdate` listener, and **the choice is defensible because the IM-layer Ban cascade is a real safety net**. Verified independently rather than just trusting the commit message.

## Why the TTL-bounded admin-disable trade-off holds up

The commit body claims "a disabled group's IM channel is Ban'd in the same flow so downstream delivery is blocked even if the auth gate briefly passes." I checked this is actually the case at `modules/group/api_manager.go:234-243`:

```go
var ban = 0
if groupStatus == GroupStatusDisabled {
    ban = 1
}
err = m.ctx.IMCreateOrUpdateChannelInfo(&config.ChannelInfoCreateReq{
    ChannelID:   groupNo,
    ChannelType: common.ChannelTypeGroup.Uint8(),
    Ban:         ban,
    ...
})
```

So `leftbangroup` (the admin Normal↔Disabled handler) sets `Ban=1` on the IM channel synchronously **before** flipping `group.Status` in DB. The two effects compose:

- Within the TTL window, the cache's stale Normal entry lets push reach `SendMessage`...
- ...but WuKongIM rejects the publish because the channel is Ban'd.

That's exactly the kind of orthogonal defence the cache+IM split was set up to allow. The auth gate is "did anyone register and not delete this webhook"; the delivery gate is "does this channel exist and not Ban'd". A stale auth-gate hit cannot resurrect a disabled channel.

This is also why the trade-off is fundamentally different from a webhook **delete**/**disable** stale-hit, which would not be backstopped by anything outside the auth gate — and those are immediately invalidated locally as the doc enumerates. The asymmetry is principled, not a corner the author cut.

## What changed in this round

**1. Tightened `cache.go` contract** — the staleness comment block now distinguishes three cases instead of conflating them:

- webhook update/delete/regenerate → local-immediate, peer-TTL (was already there, now explicit)
- group disband → local-immediate, peer-TTL (was already there, now explicit)
- **admin group-disable → all-instance TTL-bounded** (new explicit case, with the IM-Ban backstop, the reversibility argument, and the `TTL_MS=0` escape hatch all spelled out)

Plus a "what `no negative cache` actually means" clarification: misses (DB returned nil) aren't cached, but existing rows in any status (enabled/disabled/deleted) **are** cached — that's safe because the handler re-checks `m.Status` on every read (verified at `api.go:683-697` — `m == nil || m.Status == statusDeleted` and `m.Status != statusEnabled` branches sit on top of the cache return, so cache hits cannot bypass the status gate). This precisely addresses Jerry-Xin's "缓存包含禁用和软删除的 webhook" concern with the right answer rather than a behaviour change: cache is for skipping a DB read, not for being the authority on status.

**2. Pointer at `cachedRequireActiveGroup`** — a localized note at the call site reminds future readers that admin-disable specifically isn't in the invalidation matrix and points to the top-of-file contract + the pinning test.

**3. `TestPush_GroupAdminDisable_TTLBounded`** — the right shape for an "accepted staleness" test: warm cache → DB-only flip (no event) → assert push **still authorizes** within TTL → wait past TTL → assert push **now rejects**. If someone ever does subscribe to `event.GroupUpdate` and invalidates groupCache (the "if immediate is wanted later" path the comment describes), this test will go red on the "within TTL must still authorize" line and force them to update the doc + test together. Intent-locking test.

The test does set a short TTL (`2000ms`) and `time.Sleep(2500ms)`, which on slow CI is a real flake risk — Jerry-Xin's nit about preferring polling is fair. Acceptable here because the test asserts both directions of the boundary (within TTL must allow, past TTL must reject), so a timing skew that makes "past TTL" still see the cached entry would just make the test deterministically fail rather than flakily pass. Worth considering for a follow-up but not blocking.

## Concurring on yujiawei + Jerry-Xin's earlier P0 → accepted-trade-off downgrade

Both reviewers correctly identified that admin-disable wasn't covered by the invalidation matrix, and both flagged it as a P0. The maintainer's response is the right shape:

1. Documented the case explicitly as an accepted trade-off (not a "we forgot")
2. Justified it (IM-Ban backstop + reversibility + escape hatch)
3. Pinned the behaviour with a test that locks the current semantics
4. Provided the explicit migration path for the alternative ("subscribe to event.GroupUpdate")

When a reviewer says "this is a P0" and the author's response is "you're right that this is a behaviour I should call out, here's why I think it's acceptable, here's the test that locks it in, and here's the exact change if you decide otherwise" — that's the right collaborative shape, and the burden has shifted to whoever wants to argue against the trade-off to produce a concrete scenario the IM Ban doesn't cover. I don't have one.

## Non-blocking observations

- **Jerry-Xin's "请求合并 / singleflight under TTL miss"** is a real concern at very high QPS — a cold miss on a hot webhook can cause N concurrent DB reads until one wins and populates the cache. For 5 rps per webhook + 3s TTL the thundering herd is bounded to ~15 concurrent reads per cold start per group, which is acceptable for current scale. If push QPS ever climbs significantly, `golang.org/x/sync/singleflight` is the standard one-liner to dedupe in-flight loads. Worth a follow-up issue, not a blocker.
- **Jerry-Xin's "fixed sleep in test"** — noted above, accept for now given the test asserts both directions of the boundary. A `require.Eventually` + boundary check would be the polished version.
- The 3s default TTL aligns nicely with the new system_setting reload TTL (~60s) — config changes and cache staleness sit on similar order-of-magnitude windows so operators have a consistent mental model for "how stale can the system be after I push a knob".

## Combined PR follow-through

| Reviewer point | Round 1 | Round 2 |
|---|---|---|
| 🔴 yujiawei + Jerry-Xin + Steve: admin group-disable cache not invalidated | flagged as P0 | accepted trade-off, documented, pinned by test, escape hatch documented (`TTL_MS=0`) |
| 🟡 Jerry-Xin: cache contains disabled/soft-deleted entries | flagged as concern | clarified — handler re-checks Status on every read; cache doesn't bypass the gate |
| 🟡 Jerry-Xin: thundering herd on TTL miss | flagged | not addressed (out of scope, acceptable at current scale) |
| 🟡 Jerry-Xin: fixed sleep flake risk | flagged | accepted for now (boundary asserted both directions) |

## Checks

CI: Build/Lint/Vet/i18n green carry over; the new e2e test will need MySQL/Redis. Wait for the green run.

LGTM. The reasoning chain on the admin-disable trade-off is the kind of careful work that's the right way to resolve a P0 → accepted-staleness disagreement: prove the safety net exists, document the asymmetry, lock the behaviour with a test, and point at the exact change for whoever wants the other choice later.

yujiawei (changes requested):
# Code Review — PR #293 (octo-server)

Independent review of `feat(incomingwebhook): cache the push hot-path lookups (#284 item 2)` at head `59261f6`. This is a security-sensitive change: it inserts a process-local cache in front of the two DB reads that gate an **unauthenticated** push endpoint (`POST /v1/incoming-webhooks/:webhook_id/:token`).

The design is sound and well-documented overall, and the security-critical invariants mostly hold (see "Verified safe" below). I'm requesting changes for **one** issue: the code's own comments promise an *immediate local-instance* invalidation guarantee that a concurrent interleave can violate. Everything else is non-blocking.

## Verdict: CHANGES_REQUESTED

---

## P1 — Read-after-invalidate repopulation race breaks the documented "immediate local invalidation" contract

**Where:** `modules/incomingwebhook/cache.go:112-134` (`set` / `invalidate`), `modules/incomingwebhook/api.go:317-328` (`cachedQueryByWebhookID`), mutation sites `api.go:550`, `api.go:584`, `api.go:621`, `api.go:916`.

All mutation paths perform **DB write, then `invalidate()`**. The cached read does **miss → DB read → unconditional `set()`** with no generation/CAS guard:

```go
// cachedQueryByWebhookID, api.go:323-327
m, err := w.db.queryByWebhookID(webhookID)
...
if m != nil {
    w.webhookCache.set(webhookID, m)   // unconditional repopulate
}
```

Interleave:

1. `T0` — a concurrent push misses the cache and reads the **OLD** row from the DB.
2. `T1` — `update`/`delete`/`regenerate` commits the new state.
3. `T2` — the mutation calls `invalidate()`; the key isn't present yet, so it's a no-op.
4. `T3` — the push from `T0` calls `set(OLD_row)`, repopulating with a **fresh TTL**.

After `T3` the cache serves the pre-mutation row for up to a full TTL **on the same instance that performed the mutation**. The push-path status re-checks don't save it, because they evaluate the *cached stale row*:

- Rotated token: `subtle.ConstantTimeCompare(..., m.TokenHash)` at `api.go:699` compares against the cached **old** hash → the rotated-away token still authorizes.
- Just-disabled: `m.Status != statusEnabled` at `api.go:689` sees stale `enabled` → passes.
- Soft-deleted: `m.Status == statusDeleted` at `api.go:682` sees stale `enabled` → passes.

**Why this is blocking and not just a nit:** the accepted `#284` staleness contract covers two cases only — (a) cross-instance TTL staleness and (b) admin group-disable via `event.GroupUpdate`. The repopulation race is a **third**, undocumented case, and it directly contradicts the code's explicit promises of *immediate local* invalidation:

- `cache.go:15-19`: "webhook 变更（disable / delete / regenerate）会在【本实例】即时 invalidate"
- `api.go:548-550`: "本实例即时失效缓存 … push 路径不得再命中旧快照"
- `api.go:583-584`, `api.go:619-621`: same "即时失效" wording for delete / regenerate.

Shipping a security-revocation gate whose in-code invariant is false is the concern — operators reading these comments will assume the mutating instance stops immediately, and it may not.

**Impact is real but bounded:** one TTL window (default 3s), one instance, requires the precise interleave, self-heals on expiry, and `DM_INCOMINGWEBHOOK_CACHE_TTL_MS=0` disables it. The marginal exposure is the same magnitude as the already-accepted cross-instance TTL window — but it should be made *true*, not just *small*.

**Resolution (either is acceptable):**
- *Fix the code:* add a per-key generation/epoch counter bumped by `invalidate()`, captured before the DB read and re-checked under the lock in `set()` — drop the write if the epoch advanced (a short tombstone works too). Putting this inside `ttlCache.set`/`invalidate` also closes the identical groupCache-on-disband variant for free (`api.go:916`).
- *Or fix the contract:* if "TTL-bounded locally under concurrent repopulation" is genuinely acceptable, correct the `cache.go` / `api.go` comments to stop promising "即时/immediate" and document the race.

Either way, please add a **concurrent** test racing a warm push against `regenerate`/`delete`; the current `TestPush_RegenerateInvalidatesOldToken` (`api_test.go:940-958`) is single-threaded and never exercises the interleave.

---

## Verified safe (no action needed — recorded so the security surface is explicit)

- **Status is re-evaluated on every cache hit**, never cached as an auth decision: `m.Status == statusDeleted` (`api.go:682`) and `m.Status != statusEnabled` (`api.go:689`) run per request. The cache only elides the DB read.
- **No negative caching:** `set` is guarded by `m != nil` (`api.go:325`) and the group `set` by `g != nil` (`api.go:347-348`); nonexistent IDs never occupy the cache, so token-scan floods can't pollute it.
- **Correct cache targeting:** webhook mutations invalidate `webhookCache` (`api.go:550/584/621`); disband invalidates `groupCache` (`api.go:916`), and the group gate alone (`api.go:715-718`) rejects all of a disbanded group's webhooks without per-ID invalidation.
- **No data race on the shared cached pointers.** The cached `*incomingWebhookModel` / `*group.Model` are read-only on the push path (`buildPayload`/`toResp`/`truncateUTF8` only read; `recordSuccess`/`markUsed` mutate the DB row via SQL, not the in-memory struct). `get`/`set`/`invalidate` are all under `c.mu`, including the size-cap reassignment. `enabled()` reads `c.ttl` unlocked but `ttl`/`maxSize` are write-once in `newTTLCache` and published before any push goroutine. `go vet` clean; `go test -race` on the cache tests + a concurrent stress harness reported zero races. (Local verification: `go build ./...` exit 0, `go vet` clean, infra-free package tests pass under `-race`.)
- **Map is hard-bounded** at `maxSize` by the full-bucket clear before insert; lazy expiry is not relied on for bounding.
- **Accepted admin-disable TTL tradeoff** (`cache.go:20-28`, pinned by `TestPush_GroupAdminDisable_TTLBounded`) is genuinely TTL-bounded and downstream-safe via the disabled group's channel ban. Not flagged.

---

1. **Same repopulation race on `groupCache`** (`api.go:916`): a push whose miss-read started before disband can `set()` the stale `Normal` model after `invalidate`. Mitigated by the webhook status gate and channel ban, but it's the same root cause — fixed automatically if the epoch/tombstone guard lands in `ttlCache`.
2. **`cacheTTL()` overflow silently disables the cache** (`cache.go:50-57`): a value that fits `int64` but overflows `* time.Millisecond` wraps negative, making `enabled()` false. Fail-safe direction (cache off = always hit DB), but surprising — consider clamping/ceiling the parsed ms and documenting the practical upper bound.
3. **Regenerate/update invalidation not proven against a warm cache** (`api_test.go:940-958`): the test regenerates without a prior push, so the key is never cached and the assertion would still pass if `invalidate()` at `api.go:621` were deleted. Only the DELETE path (`api_test.go:734-781`) truly proves handler-side invalidation. Warm the cache first (mirror the delete test) for both update-status and regenerate.
4. **No infra-free test for the "no negative caching" contract:** the two unit tests only exercise a pre-warmed hit. A small stub-`db` test asserting a nil/non-Normal lookup leaves the key absent would harden the security-relevant invariant cheaply.
5. **Mark `ttl`/`maxSize` immutable-after-construction** with a comment on the struct fields (`cache.go:79-80`), since `enabled()` reads `ttl` without the lock — protects the current safety against a future setter.
6. **Document the shared-pointer read-only invariant** on `cachedQueryByWebhookID`/`cachedRequireActiveGroup`/`set` — the same pointer is handed to all concurrent push goroutines; a future in-memory field write would race.
7. **Full-clear eviction flushes all tenants** (`cache.go:110-124`): a >10k-distinct-id workload evicts everyone to per-push DB reads. `localFloor`'s two-generation rotation avoids this; consider rotation or at least a metric/log on full-clear.
8. **No cache observability** (hit/miss/evict counters, effective-TTL log) — a future stale-auth investigation can't distinguish cache-served from DB-served auth.

---

## For human attention (security-sensitive)

- The **cross-instance TTL staleness for token regenerate** means a rotated-out token keeps authenticating on peer instances for up to TTL. This is documented and `#284`-accepted, but because it concerns **credential revocation specifically**, it's worth an explicit maintainer ack that seconds-level revocation latency (with `TTL=0` as the instant-revocation escape hatch) is acceptable for this product.
- The **disable hatch is startup-only** (env read once in `New`, `api.go:136`) — consistent with the codebase's env-as-config convention, but worth knowing it requires a redeploy/restart to take effect, not a live toggle.

an9xyz:
Thanks @yujiawei — pushed `e1ea09a` taking the code-fix path (make the "immediate-local" invariant actually true).

**Generation guard in `ttlCache`:** `invalidate()` now bumps a monotonic `gen`; the read path captures it via `loadGen()` **before** the DB read and stores via `setIfGen()`, which drops the write if `gen` advanced in the meantime. So the interleave you described — miss-load reads the old row → mutation commits + `invalidate()` → load `set()`s the stale row — can no longer repopulate: the stale `setIfGen` is dropped because `invalidate` moved the generation. Applied to both `webhookCache` (update/delete/regenerate) and `groupCache` (disband), so the same-root-cause groupCache variant is covered too.

It's coarse (any `invalidate` aborts any in-flight `set`), but mutations are rare, so a dropped store costs at most one extra DB read — never correctness. The `cache.go` contract now states the immediacy is strict and why.

Tests: `TestTTLCache_SetIfGen_DropsStaleRepopulation` deterministically pins the drop (capture gen → invalidate → setIfGen with old gen → key absent); `TestTTLCache_ConcurrentRace` hammers the mix under `go test -race` (clean locally). Your other points (singleflight on cold-miss stampede, `require.Eventually` vs fixed sleep, full-clear flushes all tenants, observability) I've noted as worthwhile follow-ups — none are blocking and they're out of scope for this PR, happy to file an issue if you'd like them tracked.

---

Jerry-Xin (dismissed):
Relevant to octo-server: this PR modifies the `modules/incomingwebhook` push hot path. No blocking issues found.

**💬 Non-blocking**
- 🟡 Warning: [modules/incomingwebhook/api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jd5e2hge/repo/modules/incomingwebhook/api.go:328) caches any existing webhook row, including `statusDeleted` and `statusDisabled`; [cache.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jd5e2hge/repo/modules/incomingwebhook/cache.go:42) documents this explicitly. That is safe because `push` re-checks `m.Status` on every request, but it differs from the PR description’s “soft-deleted webhook IDs are never cached” wording. Either align the implementation with that contract or update the PR-facing contract.
- 🔵 Suggestion: [modules/incomingwebhook/api_test.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jd5e2hge/repo/modules/incomingwebhook/api_test.go:1019) still tests disband fail-closed without first warming `groupCache`. A small regression test that warms the group cache, invokes `handleGroupDisband`, and verifies the next push rejects would better pin the new invalidation behavior.

**✅ Highlights**
- The `ttlCache` locking and generation guard are simple and appropriate for this hot path.
- Management paths continue to read DB state uncached, which avoids stale authorization decisions for writes.
- Local invalidation is wired into update, delete, regenerate, and group disband paths.
- Focused tests passed: `go test ./modules/incomingwebhook -run 'TestTTLCache|TestCache|TestCached|TestIncomingWebhookNoLegacyResponseError'`.
- Full `go test ./modules/incomingwebhook` could not complete locally because MySQL was unavailable: `dial tcp 127.0.0.1:3306: connect: connection refused`.

Jerry-Xin (commented):
Relevant to octo-server: this PR modifies the `modules/incomingwebhook` push hot path. No blocking issues found.

**💬 Non-blocking**
- 🟡 Warning: [modules/incomingwebhook/api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jd5e2hge/repo/modules/incomingwebhook/api.go:328) caches any existing webhook row, including `statusDeleted` and `statusDisabled`; [cache.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jd5e2hge/repo/modules/incomingwebhook/cache.go:42) documents this explicitly. That is safe because `push` re-checks `m.Status` on every request, but it differs from the PR description’s “soft-deleted webhook IDs are never cached” wording. Either align the implementation with that contract or update the PR-facing contract.
- 🔵 Suggestion: [modules/incomingwebhook/api_test.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jd5e2hge/repo/modules/incomingwebhook/api_test.go:1019) still tests disband fail-closed without first warming `groupCache`. A small regression test that warms the group cache, invokes `handleGroupDisband`, and verifies the next push rejects would better pin the new invalidation behavior.

**✅ Highlights**
- The `ttlCache` locking and generation guard are simple and appropriate for this hot path.
- Management paths continue to read DB state uncached, which avoids stale authorization decisions for writes.
- Local invalidation is wired into update, delete, regenerate, and group disband paths.
- Focused tests passed: `go test ./modules/incomingwebhook -run 'TestTTLCache|TestCache|TestCached|TestIncomingWebhookNoLegacyResponseError'`.
- Full `go test ./modules/incomingwebhook` could not complete locally because MySQL was unavailable: `dial tcp 127.0.0.1:3306: connect: connection refused`.

yujiawei (dismissed):
# Code Review — PR #293 (octo-server)

## Verdict: APPROVED

This PR adds a per-instance, short-TTL, Redis-independent cache (`ttlCache[T]`) for the two uncached point reads on the push hot path — the webhook row and the group-Normal check — so a warm push issues 0 DB reads. The head commit (`e1ea09a`) also closes the **read-after-invalidate repopulation race** that was the sole remaining blocker on the previous revision. With that fix in, I found no correctness or security issue that blocks merge.

## Verification

Performed locally against head SHA `e1ea09a`:
- `go build ./modules/incomingwebhook/...` → exit 0
- `go vet ./modules/incomingwebhook/` → clean
- `go test -race` on the infra-free cache/env/hit-skips-DB tests → pass

### The repopulation-race fix is sound ✅

The previous reviewer correctly flagged that an in-flight push miss-read could capture an OLD row, then `set()` it back into the cache *after* a mutation's `invalidate()`, resurrecting a disabled/deleted/token-rotated webhook for a full TTL **on the mutating instance** — contradicting the in-code promise of "immediate local invalidation".

The fix adds a monotonic generation counter and is correct:
- `cachedQueryByWebhookID` / `cachedRequireActiveGroup` capture `loadGen()` **before** the DB read (`api.go:323`, `api.go:347`) and store via `setIfGen` (`api.go:329`, `api.go:353`).
- `invalidate()` does `gen++` **and** `delete(key)` in one critical section under `c.mu` (`cache.go:168-176`).
- `setIfGen()` re-checks `c.gen == gen` under the same lock and drops the write on mismatch (`cache.go:138-140`).

The decisive invariant is the ordering in every mutation handler: the DB write commits **before** `invalidate()` is called — update (`api.go:548→555`), delete (`api.go:585→589`), regenerate (`api.go:618→626`), disband (`api.go:914→921`). Given that ordering, every interleaving is safe:
- A miss-read whose `loadGen()` ran **before** the mutation → gen advances → `setIfGen` drops the stale store; the unconditional `delete()` in `invalidate` also wipes any entry that briefly landed.
- A miss-read whose `loadGen()` ran **after** `invalidate()` → the DB write has already committed, so the DB read returns the NEW row, and storing it is correct.

I could not construct a post-mutation auth-bypass on the mutating instance for disable, delete, token-rotation, or disband. The global (not per-key) generation counter is coarse, but coarseness can only cause *extra* dropped stores (a redundant DB read), never serve stale data — a perf trade-off, not a correctness hole.

### Security surface (this PR is security-sensitive — auth revocation) ✅

- **Status is re-evaluated on every cache hit**, never cached as an auth decision. `m.Status == statusDeleted` (`api.go:687`) and `m.Status != statusEnabled` (`api.go:693`) run per request; the token compare (`api.go:702`) and the group-Normal gate (`api.go:711-718`) likewise. The cache elides only the DB read, not the gate.
- **No negative caching.** `setIfGen` is guarded by `m != nil` / `g != nil` (`api.go:328`, `api.go:352`), so nonexistent IDs never occupy the cache and token-scan floods cannot pollute it. (Note this is *not* "only enabled rows are cached" — disabled/deleted rows are legitimately cached, but the per-hit status gate rejects them, so it is safe.)
- **Correct cache targeting.** Webhook mutations invalidate `webhookCache`; disband invalidates `groupCache`, and the group gate alone rejects every webhook of a disbanded group without per-ID invalidation.
- **No data race on the shared cached pointers.** The cached `*incomingWebhookModel` / `*group.Model` are read-only on the push path; mutating helpers (`recordSuccess`/`markUsed`) write the DB row via SQL, not the in-memory struct. All map access is under `c.mu`, including the size-cap reassignment.

## For human attention (security-sensitive)

These are documented, `#284`-accepted trade-offs — not blockers — but because they concern credential/permission revocation latency they warrant an explicit maintainer ack:
- **Cross-instance TTL staleness for token regenerate**: a rotated-out token keeps authenticating on *peer* instances for up to one TTL (default 3s). `DM_INCOMINGWEBHOOK_CACHE_TTL_MS=0` is the instant-revocation escape hatch.
- **Admin group-disable (Normal→Disabled) is TTL-bounded on all instances** (this module subscribes only to `event.GroupDisband`, not `event.GroupUpdate`). Mitigated downstream because a disabled group's IM channel is banned. Pinned by `TestPush_GroupAdminDisable_TTLBounded`.
- **The disable hatch is startup-only** (env read once in `New`), so changing the TTL requires a redeploy, not a live toggle.

## Non-blocking suggestions (P2 / nits)

1. **Harden mutation-invalidation tests against a warm cache.** `TestPush_RegenerateInvalidatesOldToken` (`api_test.go:940`) regenerates without a prior push, so the key is never cached — the assertion would still pass even if the `invalidate()` at `api.go:626` were removed. The new `TestPush_CacheServesStaleWithinTTL_HandlerInvalidates` proves this well for the DELETE path; mirroring it (warm the cache first) for regenerate and update-status would close the gap. A concurrent test racing a warm push against regenerate/delete would additionally exercise the gen-guard end-to-end (the deterministic `TestTTLCache_SetIfGen_DropsStaleRepopulation` already pins the mechanism in isolation).
2. **`cacheTTL()` overflow silently disables the cache** (`cache.go:51-58`): a parsed ms value that overflows `* time.Millisecond` wraps negative → `enabled()` false. The direction is fail-safe (cache off), but consider clamping the parsed value and documenting the practical upper bound.
3. **Full-clear eviction flushes all entries** (`cache.go:141-145`): a >`maxSize`-distinct-id workload degrades everyone to per-push DB reads. `localFloor`'s generation rotation avoids this; consider rotation, or at least a log/metric on full-clear.
4. **No cache observability** (hit/miss/evict counters, effective-TTL log): a future stale-auth investigation can't distinguish cache-served from DB-served auth.
5. **Mark `ttl`/`maxSize` immutable-after-construction** with a field comment (`cache.go`), since `enabled()` reads `ttl` without the lock — they are write-once today, but a comment protects that against a future setter.

None of these block merge.

Jerry-Xin (commented):
The PR is in scope for octo-server: it changes the incoming webhook push path and related tests in `modules/incomingwebhook`.

🔴 Blocking

None.

💬 Non-blocking

🟡 Warning: The group admin-disable behavior is intentionally TTL-stale on every instance because `GroupUpdate` is not subscribed, only `GroupDisband` is invalidated. This is documented in [api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-baqo379q/repo/modules/incomingwebhook/api.go:355) and pinned in [api_test.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-baqo379q/repo/modules/incomingwebhook/api_test.go:838). I agree it is not a blocker given the short default TTL and stated contract, but it is worth keeping visible because it is auth-gate behavior, not only a performance detail.

🔵 Suggestion: `TestPush_GroupAdminDisable_TTLBounded` uses a fixed `time.Sleep(2500 * time.Millisecond)` against a `2000ms` TTL in [api_test.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-baqo379q/repo/modules/incomingwebhook/api_test.go:883). This is probably fine, but an eventual assertion with a deadline would be less sensitive to slow CI timing.

🔵 Suggestion: The PR description says soft-deleted webhook IDs are never cached, but the implementation caches any existing row, including disabled/deleted rows, at [api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-baqo379q/repo/modules/incomingwebhook/api.go:344) and documents that choice in [cache.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-baqo379q/repo/modules/incomingwebhook/cache.go:42). The code remains safe because `push` re-checks status on every hit, but the PR text should match the actual contract.

✅ Highlights

The generation guard in `setIfGen` cleanly handles the miss-read followed by invalidation race, and the mutation paths invalidate the local webhook or group key in the right places. The cache is Redis-independent, bounded, TTL-disableable with `DM_INCOMINGWEBHOOK_CACHE_TTL_MS=0`, and covered by focused unit tests.

Verification run: `go test ./modules/incomingwebhook -run 'TestTTLCache|TestCacheTTLEnv|TestCacheMaxEnv|TestCached'`, `go test ./modules/incomingwebhook -run TestIncomingWebhookNoLegacyResponseError`, and `git diff --check` all passed.

lml2468 (approved):
APPROVED (head `b00eec74f940`).

Pure rebase on `main` after #292 landed. **No semantic change** since my R2 APPROVED on `e1ea09abfce0`. Verified:

- `modules/incomingwebhook/cache.go` blob SHA on `b00eec74f940` = blob SHA on `e1ea09abfce0` = `1d530856...` (byte-identical).
- `modules/incomingwebhook/cache_test.go` blob SHA identical too.
- The other files that *appear* to have changed (`api.go` / `api_i18n.go` / `api_test.go` / `ratelimit.go` / common/*) just absorbed the #292 master-toggle + Positive-flag changes that landed on `main` between the two heads — the PR's *own* diff against main is the same shape (cache.go +176, cache_test.go +181, api.go +78).
- PR commit graph is now feat / docs / fix (the same three logical commits as last round, rebased onto current main).

Generation-guard fix from R2 still stands:
- `invalidate()` bumps `gen`, `loadGen()` snapshots it before the DB read, `setIfGen()` drops the write if `gen` advanced — closes yujiawei's read-after-invalidate repopulation race.
- `cachedQueryByWebhookID` / `cachedRequireActiveGroup` both call `loadGen()` *before* `db.queryByWebhookID` / `requireActiveGroup`, then `setIfGen()` after — correct order.
- `TestTTLCache_SetIfGen_DropsStaleRepopulation` deterministically pins the drop; `TestTTLCache_ConcurrentRace` exercises locking under `-race`.

Jerry-Xin's R3 non-blocking nits (PR description says "soft-deleted not cached" but implementation caches all existing rows; suggest adding a groupCache+disband test) carry over — both are doc/test polish, not behaviour issues. Cache hits never bypass the status gate (handler re-checks `m.Status == statusDeleted` and `m.Status != statusEnabled` on every read, verified previously). Acceptable to defer.

CI green carries over modulo the rebase; merge once Test/CodeQL re-confirm on the new SHA.

LGTM, no new substantive concerns.

yujiawei (approved):
# Code Review — PR #293 (octo-server)

## Verdict: APPROVED

A well-engineered, carefully-scoped change. The push hot path gets a per-instance, short-TTL, Redis-independent positive-only cache for the two pre-auth DB point reads (webhook row + group-Normal check). I reviewed it specifically as a **security-sensitive** change — the central question being "can the cache be used to bypass the push auth gate?" — and the answer is **no**. No P0/P1 issues found. The findings below are all non-blocking (P2/nit).

## 1. Verification of the auth-staleness contract

| Claim | Result | Evidence |
|---|---|---|
| Warm push issues 0 DB reads | ✅ | `cachedQueryByWebhookID` / `cachedRequireActiveGroup` return on cache hit before any DB call (`api.go:334-336`, `api.go:359-361`); proven by `TestCached*_HitSkipsDB` with `db == nil` |
| Cache never bypasses the status gate | ✅ | `push` re-evaluates `m.Status` (`api.go:703`, `api.go:710`) and re-runs the constant-time token compare (`api.go:720`) on **every** read — a cached disabled/deleted/wrong-token row is still rejected. The cache only saves the DB read, it does not short-circuit any check. |
| disable/delete/regenerate invalidate locally and immediately | ✅ | `update` `api.go:571`, `delete` `api.go:605`, `regenerate` `api.go:642` all call `webhookCache.invalidate` after the write |
| group disband invalidates the group entry locally | ✅ | `handleGroupDisband` `api.go:937` calls `groupCache.invalidate`; the group gate alone blocks all of the group's webhooks even if their rows are still briefly cached as enabled |
| No negative caching of nonexistent IDs | ✅ | `cachedQueryByWebhookID` only stores when `m != nil` (`api.go:344-346`); token-scan IDs never occupy a slot |
| Read-after-invalidate race is closed | ✅ | The generation guard (`loadGen` before the DB read, `setIfGen` after) at `cache.go:120-147` drops a stale in-flight repopulation; production read paths use it correctly, `.set()` is test-only |
| Management writes stay uncached | ✅ | `update`/`delete`/`regenerate`/`requireActiveGroup` for re-enable all go through the direct `db.*` / `requireActiveGroup`, so revival/permission decisions never run on a stale snapshot |
| `TTL=0` disables the cache (exact pre-change behavior) | ✅ | `enabled()` short-circuits all methods when `ttl <= 0` (`cache.go`), parsed by `cacheTTL()` |
| Cached pointer is read-only after caching | ✅ | No field of the cached `*incomingWebhookModel` / `*group.Model` is mutated post-cache (`buildPayload`/`recordSuccess` only read `m`), so the shared pointer is safe under concurrent pushes |

I also confirmed `go build ./modules/incomingwebhook/` and `go vet` pass cleanly at the head SHA.

### Cross-instance & admin-disable staleness — acknowledged, accepted
Cross-instance staleness is bounded by the TTL (no active cross-instance invalidation), and **admin group-disable** (`Normal → Disabled`, emitted as `event.GroupUpdate`, which this module does not subscribe to) is deliberately *not* in the invalidation matrix — so an admin-disabled group's push gate stays open for up to one TTL on every instance. This is a real, documented trade-off (`cache.go` top contract; pinned by `TestPush_GroupAdminDisable_TTLBounded`), consistent with #284's accepted seconds-level staleness, and reversible via a small `TTL` (or `=0`). Destructive disband remains immediate. **This is the one item a human owner should explicitly sign off on**, since it weakens the immediacy of an admin action — but it is an accepted design decision, not a defect.

**P2 — Coarse whole-bucket eviction is a perf cliff at extreme scale** (`cache.go`, the `len(c.m) >= c.maxSize` branch in `setIfGen`/`set`). When the map is full and a new key arrives, the **entire** bucket is discarded (`c.m = make(...)`) rather than evicting one entry. If a single instance's active distinct webhook+group working set stably exceeds `DM_INCOMINGWEBHOOK_CACHE_MAX` (default 10000) within one TTL, almost every new-key write wipes all ~10000 hot entries and the hit rate collapses toward zero (graceful degradation to pre-PR per-push DB reads, plus a map rebuild). Correctness/security are unaffected (eviction only ever causes *more* DB reads). Suggestion: document the capacity semantics + an ops tuning note, or switch to single-entry approximate-LRU/random eviction. Not blocking — the trigger requires >10k distinct active keys per instance per TTL.

**P2 — Read-after-invalidate guard is not exercised at the call-site level** (`cache_test.go`). `TestTTLCache_SetIfGen_DropsStaleRepopulation` deterministically pins the gen guard on the bare `ttlCache`, which is good — but nothing tests the *ordering* inside `cachedQueryByWebhookID`/`cachedRequireActiveGroup` (`loadGen` must be captured **before** the DB read, `setIfGen` **after**). A regression that moved `loadGen()` to after `db.queryByWebhookID` would re-open the exact race this PR closes and still pass every test here (the hit-skips-DB tests short-circuit before `loadGen`; the integration tests never interleave an invalidate with an in-flight miss). Suggestion: add a helper-level test with a fake `db` whose query callback fires `invalidate()` before returning, then assert the helper did not repopulate.

**nit — Cache `TTL`/`MAX` are read once at construction**, unlike the module's other thresholds (`maxBytes()`, `ipFailRPS()`, etc.) which re-read env per request for runtime tunability. Not a bug (the test comments acknowledge it), but it diverges from the module's "runtime-tunable" convention; changing `DM_INCOMINGWEBHOOK_CACHE_TTL_MS`/`_MAX` requires a restart. Worth a one-line note in the `cache.go` contract so operators don't expect a hot reload.

**nit — Time-based tests rely on wall-clock margins** (`TestPush_GroupAdminDisable_TTLBounded` uses `TTL=2000ms` + `time.Sleep(2500ms)`; `TestTTLCache_Expiry` uses 20ms/40ms). The within-TTL assertion has a full-TTL margin and the post-TTL assertion errs in the safe direction, so genuine flakiness is unlikely — but there is no injectable clock seam, and the hard ~2.5s cost is unavoidable. Optional: inject a clock for determinism + speed.

**nit — Push-path cache tests are infra-gated** (`setupTestEnv` → `NewTestServer` needs MySQL/Redis/WuKongIM, no `t.Skip` guard). This matches the repo-wide convention for this module, so it's not a regression, but the security-relevant assertions (DB-only mutation stays stale-but-authorized within TTL; handler-delete → immediate 401) only run on the integration tier — worth confirming CI runs that tier so they aren't effectively dead.

**nit — Valid-ID enumeration can thrash the bucket** (`api.go:696-724`): the row is cached on a miss *before* the token compare, so an attacker who knows many valid `iwh_` IDs (but not tokens) can insert one entry per ID despite every request 401-ing, and at `maxSize` trigger the whole-bucket clear above. No auth bypass (status + token are re-checked on every read) and gated by the per-IP failure budget (#285) before volume is achievable; worst case is loss of the DB-read savings. Noted for completeness; no change required.

## 3. Recommendation

Approve. The security contract holds, invalidation covers all local mutation paths, the read-after-invalidate race is correctly closed, and the `TTL=0` escape hatch preserves exact prior behavior. The accepted admin-disable staleness window is the one design decision worth a conscious human ack, but it is documented, tested, and reversible. The P2 items (whole-bucket eviction cliff; call-site race-guard test) are good follow-ups but do not block merge.

PR #216, opened 2026-06-01
ci: Phase 1 test quality — gotestsum + diff coverage + PR summary

lml2468:
## Summary

Phase 1 test quality CI upgrade for octo-server.

#### Test job upgrades
- **gotestsum**: Replaces bare `go test` — produces JUnit XML for structured test results
- **`-shuffle=on`**: Added alongside existing `-race` flag for better flake detection
- **Coverage**: Generates `coverage.out` with `-coverprofile` and `-covermode=atomic`
- **Artifacts**: Uploads `junit.xml` + `coverage.out` on both success and failure (14-day retention)
- **PR Summary**: Writes test results (passed/failed/skipped counts) to `$GITHUB_STEP_SUMMARY`
- **Per-package DB reset**: Preserved exactly — each package gets a fresh MySQL database + Redis flush

#### New: Diff Coverage job
- Uses `octocov-action` to generate diff coverage reports as PR comments
- Overall coverage threshold: 60%
- Core packages tracking: `./modules/user/...`, `./modules/thread/...`, `./modules/group/...` (75% threshold)
- Only runs on pull_request events, after test job succeeds

#### Permissions
- Added `pull-requests: write` at workflow level for octocov PR comments

### What's NOT changed
- All existing jobs preserved (build, vet, lint, personal-msgsendreq-lint, i18n-extract-check, i18n-lint)
- Service container config (MySQL + Redis) unchanged
- `OCTO_MASTER_KEY` env var unchanged
- `changes` preflight job unchanged
- Trigger conditions, concurrency, cancel-in-progress unchanged

### Dependencies
- Requires `Mininglamp-OSS/.github` PR #55 to be merged first (for the reusable workflow reference in future iterations)

### Checklist
- [x] gotestsum replaces bare go test
- [x] JUnit XML + coverage.out produced and uploaded
- [x] Diff coverage job with octocov PR comments
- [x] Core packages coverage tracking
- [x] PR summary in $GITHUB_STEP_SUMMARY
- [x] Existing jobs preserved
- [x] YAML validated

PR #295, opened 2026-06-06
feat(conversation): opt-in recent activity-window filter on /v1/conversation/sync (#294)

### LINKED ISSUE #294: server: conversation sync ignores sidebar recent filter settings
## Description

`sidebar.recent_filter_group_days` currently only affects the `recent` branch of `POST /v1/sidebar/sync`.

However, the current web client's Recent tab does not call `/v1/sidebar/sync` with `tab=recent`. It loads the Recent conversation list through:

- `POST /v1/conversation/sync?space_id=<space_id>`
- request body: `{"msg_count":1}`

As a result, changing `sidebar.recent_filter_group_days` (for example from the default `3` to `0`) does not change the server response consumed by the web Recent tab.

## Steps to Reproduce

1. Configure system setting `sidebar.recent_filter_group_days=0`.
2. Open the web client Recent tab in a Space.
3. Observe the backend request used by the web Recent tab.
4. The client calls `POST /v1/conversation/sync?space_id=<space_id>`, not `POST /v1/sidebar/sync` with `tab=recent`.
5. Check that `/v1/conversation/sync` does not apply the dynamic `sidebar.recent_filter_{group,thread,person}_days` cutoffs.

## Expected Behavior

The server-side Recent conversation data used by the web Recent tab should honor the same dynamic activity-window settings:

- `sidebar.recent_filter_group_days`
- `sidebar.recent_filter_thread_days`
- `sidebar.recent_filter_person_days`

In particular, `sidebar.recent_filter_group_days=0` should disable server-side group activity filtering for the Recent list returned to the current web client.

## Actual Behavior

`sidebar.recent_filter_group_days=0` only affects `/v1/sidebar/sync` when `tab=recent`.

The web Recent tab still receives data from `/v1/conversation/sync`, whose `syncUserConversation` path does not reuse the sidebar recent filter cutoffs.

Relevant server code:

- `modules/message/api_sidebar.go`: `req.Tab == "recent"` calls `loadRecentCutoffs(...)` and `buildRecentItems(...)`.
- `modules/message/api_conversation.go`: `syncUserConversation` calls `co.ctx.IMSyncUserConversation(...)` and then builds `SyncUserConversationResp` without applying the dynamic recent cutoffs.

## DMWork Version

`main` / current upstream behavior.

## Affected Component

Backend (Server)

## Severity

P2 - Non-core Feature Issue

## Reproducibility

## Environment

- Web client Recent tab
- Backend endpoint: `POST /v1/conversation/sync?space_id=<space_id>`

## Suggested Fix

Reuse the sidebar recent-filter configuration in the `/v1/conversation/sync` path.

Implementation notes:

1. Extract the recent cutoff logic from `api_sidebar.go` into a shared helper, for example:
   - `recentCutoffs`
   - `daysCutoff`
   - `loadRecentCutoffs(ctx, now)`
   - raw-conversation filtering by channel type and timestamp
2. In `syncUserConversation`, after `co.ctx.IMSyncUserConversation(...)`, keep a raw copy for cursor/unread logic and apply the filter to the response-building list.
3. Keep cursor advancement based on the unfiltered raw conversations. Otherwise, if the highest-version conversation is filtered out, the client can repeatedly sync the same batch.
4. Keep per-Space unread calculations based on raw conversations where needed.

## Workaround

No backend-only workaround for the current web Recent tab. Calling `/v1/sidebar/sync tab=recent` directly reflects the setting, but the web Recent tab does not consume that response today.

an9xyz:
Closes #294.

## Background

PR #291 made the `recent` tab of `POST /v1/sidebar/sync` apply admin-tunable per-channel-type activity windows (`sidebar.recent_filter_{group,thread,person}_days`, `0` = no filter).

But the current web client's **Recent tab does not call `/v1/sidebar/sync`** — it loads the conversation list through `POST /v1/conversation/sync` (body `{"msg_count":1}`). That handler (`syncUserConversation`) has never applied any time-window filter, so changing `sidebar.recent_filter_group_days` (e.g. `3` → `0`) has no effect on what the web Recent tab receives.

## Why not filter unconditionally

The issue's suggested fix is to reuse the cutoffs directly in `syncUserConversation`. Applied unconditionally that would be a regression: `/v1/conversation/sync` is the **universal offline/initial conversation-sync endpoint** used by mobile and desktop too — there is no `tab` marker distinguishing the web Recent tab from a mobile full sync. With the default `group=3/thread=3`, any conversation quiet for >3 days would silently drop from every client's conversation list; on a `version=0` initial sync the cursor advances past them (cursor is raw-based), so they would be **permanently missing** until new activity. PR #291 explicitly documented this endpoint as "intentionally untouched".

## What this does

Adds the filter behind an **explicit opt-in request flag**, so existing clients are byte-for-byte unaffected and the web Recent tab can opt in.

- **New optional request field `recent_filter`** (boolean, default `false`). When `true`, the response conversation list is filtered using the **same** per-channel-type cutoffs the sidebar recent tab uses (`recentCutoffs` / `cutoffFor`): a conversation is hidden iff its type's cutoff is non-zero AND its `timestamp <= cutoff`; `0` = no filter for that type (DM default `0` keeps DMs always shown); unknown types kept. Default `false` = today's exact behaviour.
- `loadRecentCutoffs` is promoted from a `*Sidebar` method to a package-level function so both handlers share one window-resolution path (both live in package `message`).

### Invariants preserved

- **Cursor**: the filter runs *after* `lastVersion = maxConversationVersion(rawConversations, …)`. A filtered-out max-version conversation cannot stall the cursor (same class of fix as PR #1377 / sidebar B1).
- **Per-Space unread**: `fillPersonSpaceUnread` reads the raw conversation slice for unread counts and only enriches existing response items — it does not re-introduce filtered entries.
- **System bots**: `EnsureSystemBotsPresent` runs after this step, so system-bot visibility is never affected even when a `person` window is configured.

## Changes

- `modules/message/api_conversation.go`: optional `recent_filter` field; `filterRecentConversations` helper (returns a new slice, no input mutation); guarded call placed after cursor advancement.
- `modules/message/api_sidebar.go`: `loadRecentCutoffs` → package-level function; updated caller.
- `modules/message/swagger/conversation.yaml`: documents the new field.

## Tests

`filterRecentConversations` unit cases: stale group/thread dropped, DM kept when `person` window `0`, `group` cutoff `0` keeps all, `person` window filters DMs, unknown channel type kept, empty input, and no-mutation of the input slice. Existing sidebar recent-filter tests still pass under the package-level helper.

`go build ./...`, `go vet ./modules/message/`, and the i18n lints pass (no error codes touched).

## Follow-up (frontend, out of scope here)

The web Recent tab must set `recent_filter: true` for the admin window to take effect. The cleaner long-term direction is for the web Recent tab to consume `/v1/sidebar/sync tab=recent` (purpose-built and already filtered), which would let this opt-in be retired.

### DIFF
--- a/modules/message/api_conversation.go
+++ b/modules/message/api_conversation.go
@@ -744,8 +744,12 @@ func (co *Conversation) syncUserConversation(c *wkhttp.Context) {
 	// 因为被过滤掉的会话可能正好是本批最高 version 的那条；用过滤后列表推 cursor
 	// 会让客户端反复拉同一批（与 PR-B #1377 / sidebar B1 同一类死循环）。
 	// per-Space 未读仍基于 raw conversations（见下方 fillPersonSpaceUnread），不受影响。
-	// 系统 Bot 的可见性兜底（EnsureSystemBotsPresent）在 Space 过滤块里、本步之后
-	// 执行，因此即便开启 person 窗口也不会把系统 Bot 误删。
+	//
+	// 系统 Bot 可见性：person 窗口默认 0（不过滤 DM），系统 Bot 默认安全。当
+	// 请求带 space_id 时，EnsureSystemBotsPresent 在 Space 过滤块里、本步之后兜底
+	// 补齐，即便管理员开了 person 窗口也不会丢失系统 Bot —— Web「最近」tab 正是带
+	// space_id 调用，故实际使用安全。仅「不带 space_id + recent_filter=true +
+	// person 窗口>0」这一非常规组合下，安静的系统 Bot DM 可能被本步过滤掉。
 	if req.RecentFilter {
 		cutoffs := loadRecentCutoffs(co.ctx, time.Now())
 		syncUserConversationResps = filterRecentConversations(syncUserConversationResps, cutoffs)
--- a/modules/message/api_conversation.go
+++ b/modules/message/api_conversation.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/Mininglamp-OSS/octo-lib/common"
 	"github.com/Mininglamp-OSS/octo-lib/config"
@@ -290,6 +291,15 @@ func (co *Conversation) syncUserConversation(c *wkhttp.Context) {
 		LastMsgSeqs string `json:"last_msg_seqs"` // 客户端所有会话的最后一条消息序列号 格式： channelID:channelType:last_msg_seq|channelID:channelType:last_msg_seq
 		MsgCount    int64  `json:"msg_count"`     // 每个会话消息数量
 		DeviceUUID  string `json:"device_uuid"`   // 设备uuid
+		// RecentFilter 为 true 时，对响应会话列表套用 sidebar recent tab 同款的
+		// 按频道类型活动窗口过滤（system_settings 的 sidebar.recent_filter_*_days，
+		// 0=该类型不过滤）。issue #294：Web「最近」tab 走本端点而非 /v1/sidebar/sync，
+		// 需要显式 opt-in 才能让管理员配置的过滤窗口生效。
+		//
+		// 默认 false —— 移动端/桌面端的离线全量同步行为完全不变（PR #291 明确把本
+		// 端点列为 "intentionally untouched"），避免安静群/子区从通用会话同步中消失。
+		// 过滤只作用于响应 list，cursor 推进与 per-Space 未读仍基于过滤前的原始会话。
+		RecentFilter bool `json:"recent_filter"`
 	}
 	if err := c.BindJSON(&req); err != nil {
 		co.Error("数据格式有误！", zap.Error(err))
@@ -724,6 +734,23 @@ func (co *Conversation) syncUserConversation(c *wkhttp.Context) {
 	}
 	co.syncConversationResultCacheLock.Unlock()
 
+	// ---------- recent 活动窗口过滤（opt-in，issue #294） ----------
+	// Web「最近」tab 走本端点（非 /v1/sidebar/sync），需要这里复用 sidebar recent
+	// tab 同款的按频道类型活动窗口过滤，管理员配置的 sidebar.recent_filter_*_days
+	// 才能对它生效。仅在客户端显式 RecentFilter=true 时启用 —— 默认关闭，移动端/
+	// 桌面端的离线全量同步行为完全不变。
+	//
+	// 顺序约束：必须在 cursor 推进（lastVersion，基于 raw conversations）之后，
+	// 因为被过滤掉的会话可能正好是本批最高 version 的那条；用过滤后列表推 cursor
+	// 会让客户端反复拉同一批（与 PR-B #1377 / sidebar B1 同一类死循环）。
+	// per-Space 未读仍基于 raw conversations（见下方 fillPersonSpaceUnread），不受影响。
+	// 系统 Bot 的可见性兜底（EnsureSystemBotsPresent）在 Space 过滤块里、本步之后
+	// 执行，因此即便开启 person 窗口也不会把系统 Bot 误删。
+	if req.RecentFilter {
+		cutoffs := loadRecentCutoffs(co.ctx, time.Now())
+		syncUserConversationResps = filterRecentConversations(syncUserConversationResps, cutoffs)
+	}
+
 	// ---------- 子区 source_message_id ----------
 	co.fillThreadMeta(syncUserConversationResps)
 
@@ -848,6 +875,25 @@ func (co *Conversation) syncUserConversation(c *wkhttp.Context) {
 		ChannelStates:    channelStates,
 		SpaceMemberships: spaceMemberships,
 	})
+}
+
+// filterRecentConversations drops conversations whose per-channel-type activity
+// window has elapsed, mirroring the sidebar recent tab (buildRecentItems): a
+// conversation is hidden iff its type's cutoff is non-zero AND its Timestamp is
+// at or before that cutoff. A cutoff of 0 means "no filter" for that type, and
+// unknown channel types are kept unconditionally (recentCutoffs.cutoffFor).
+//
+// Returns a new slice — the input is not mutated, so the caller's raw-based
+// cursor/unread computations stay intact (issue #294).
+func filterRecentConversations(resps []*SyncUserConversationResp, cutoffs recentCutoffs) []*SyncUserConversationResp {
+	filtered := make([]*SyncUserConversationResp, 0, len(resps))
+	for _, r := range resps {
+		if cutoff := cutoffs.cutoffFor(r.ChannelType); cutoff != 0 && r.Timestamp <= cutoff {
+			continue
+		}
+		filtered = append(filtered, r)
+	}
+	return filtered
 }
 
 func (co *Conversation) channelMessageSeqJoin(channelID string, channelType uint8, lastMessageSeq uint32) string {
--- a/modules/message/api_sidebar.go
+++ b/modules/message/api_sidebar.go
@@ -526,7 +526,7 @@ func (sb *Sidebar) Sync(c *wkhttp.Context) {
 			}
 		}
 	case "recent":
-		cutoffs := sb.loadRecentCutoffs(time.Now())
+		cutoffs := loadRecentCutoffs(sb.ctx, time.Now())
 		items = buildRecentItems(conversations, cutoffs, pinnedSet, groupSpaceMap, externalGroupMap, defaultSpaceID)
 	}
 
@@ -661,13 +661,17 @@ func daysCutoff(now time.Time, days int) int64 {
 // hard-coded behaviour: groups/threads use a 3-day window, DMs are unfiltered.
 // (Unknown types are kept unconditionally — see cutoffFor.)
 //
+// Package-level (not a *Sidebar method) so /v1/conversation/sync can reuse the
+// exact same window resolution when a client opts in to recent filtering
+// (issue #294) — both handlers live in package message.
+//
 // NOTE: EnsureSystemSettings returns a process-wide singleton, so this reads a
 // snapshot shared across the process. Tests that mutate sidebar.* settings must
 // Reload() after wiping the table (see the recent-filter e2e setup), else a
 // stale snapshot from a prior test (within the ~60s auto-reload TTL) can leak
 // in.
-func (sb *Sidebar) loadRecentCutoffs(now time.Time) recentCutoffs {
-	ss := commonapi.EnsureSystemSettings(sb.ctx)
+func loadRecentCutoffs(ctx *config.Context, now time.Time) recentCutoffs {
+	ss := commonapi.EnsureSystemSettings(ctx)
 	return recentCutoffs{
 		group:  daysCutoff(now, ss.SidebarRecentFilterGroupDays()),
 		thread: daysCutoff(now, ss.SidebarRecentFilterThreadDays()),
--- a/modules/message/api_sidebar_recent_filter_e2e_test.go
+++ b/modules/message/api_sidebar_recent_filter_e2e_test.go
@@ -55,9 +55,8 @@ func TestE2E_RecentFilter_DefaultsReproduceLegacyBehaviour(t *testing.T) {
 	// defaults here (PR #291 review, lml2468).
 	require.NoError(t, commonapi.EnsureSystemSettings(ctx).Reload())
 
-	sb := &Sidebar{ctx: ctx}
 	now := time.Now()
-	cutoffs := sb.loadRecentCutoffs(now)
+	cutoffs := loadRecentCutoffs(ctx, now)
 
 	assert.Equal(t, daysCutoff(now, 3), cutoffs.group, "群默认 3 天窗口")
 	assert.Equal(t, daysCutoff(now, 3), cutoffs.thread, "话题默认 3 天窗口")
@@ -101,9 +100,8 @@ func TestE2E_RecentFilter_AdminOverrideTakesEffect(t *testing.T) {
 	s.GetRoute().ServeHTTP(w, req)
 	require.Equal(t, http.StatusOK, w.Code, w.Body.String())
 
-	sb := &Sidebar{ctx: ctx}
 	now := time.Now()
-	cutoffs := sb.loadRecentCutoffs(now)
+	cutoffs := loadRecentCutoffs(ctx, now)
 	assert.Equal(t, int64(0), cutoffs.group, "群窗口关闭 → cutoff 0")
 	assert.Equal(t, daysCutoff(now, 3), cutoffs.thread)
 	assert.Equal(t, daysCutoff(now, 7), cutoffs.person)
--- a/modules/message/conversation_recent_filter_e2e_test.go
+++ b/modules/message/conversation_recent_filter_e2e_test.go
@@ -1,0 +1,256 @@
+//go:build integration
+
+package message
+
+// End-to-end coverage for the opt-in recent activity-window filter on
+// POST /v1/conversation/sync (issue #294). Unlike the pure-logic unit tests in
+// conversation_recent_filter_test.go, these drive the REAL HTTP handler
+// (syncUserConversation) through the registered route, with WuKongIM mocked by
+// an httptest server (the IM result is the only external seam the handler can't
+// reach in CI). MySQL + Redis must be up (same as the sidebar recent-filter
+// e2e).
+//
+// IM mock: IMSyncUserConversation is a concrete *config.Context method that
+// POSTs to {WuKongIM.APIURL}/conversation/sync. We point that URL at a local
+// httptest server returning a canned conversation slice — no interface seam
+// needed. DM-only conversations keep the seeding minimal: groups would require
+// membership/detail rows, and DMs skip the group-membership / thread / space
+// branches entirely (space_id omitted ⇒ no Space filter).
+//
+// Build-tagged `integration` (run with `go test -tags=integration`), matching
+// api_sidebar_recent_filter_e2e_test.go — these spin up testutil.NewTestServer
+// against the shared `test` DB and are order-fragile in the default `go test`
+// job. The filter logic itself is covered by the default-job unit tests; this
+// file is the full-stack glue check the reviewer asked for.
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-lib/server"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	commonapi "github.com/Mininglamp-OSS/octo-server/modules/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const masterKeyEnvConvE2E = "OCTO_MASTER_KEY"
+
+// One long-lived fake WuKongIM server for the whole file. Per-test httptest
+// servers proved fragile: testutil.NewTestServer binds the registered handler's
+// ctx to the FIRST server's config, so a per-test APIURL (whose server is then
+// closed) leaves later tests dialling a dead port. A single never-closed server
+// with a mutable response slice keeps the URL stable and alive across all tests.
+// Tests run sequentially (no t.Parallel), so the shared slice needs no lock.
+var (
+	fakeIMOnce  sync.Once
+	fakeIMSrv   *httptest.Server
+	fakeIMConvs []*config.SyncUserConversationResp
+)
+
+func sharedFakeIM() *httptest.Server {
+	fakeIMOnce.Do(func() {
+		fakeIMSrv = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.Header().Set("Content-Type", "application/json")
+			if strings.HasSuffix(r.URL.Path, "/conversation/sync") {
+				_, _ = w.Write([]byte(util.ToJson(fakeIMConvs)))
+				return
+			}
+			_, _ = w.Write([]byte("{}"))
+		}))
+	})
+	return fakeIMSrv
+}
+
+// dmIMConv builds a DM conversation as IMSyncUserConversation would return it,
+// with a single non-deleted recent message so the handler keeps it (the build
+// loop drops conversations whose Recents end up empty).
+func dmIMConv(channelID string, ts, version int64, seq uint32) *config.SyncUserConversationResp {
+	return &config.SyncUserConversationResp{
+		ChannelID:   channelID,
+		ChannelType: common.ChannelTypePerson.Uint8(),
+		Unread:      0,
+		Timestamp:   ts,
+		LastMsgSeq:  int64(seq),
+		Version:     version,
+		Recents: []*config.MessageResp{
+			{
+				MessageID:   version*1000 + int64(seq),
+				MessageSeq:  seq,
+				ClientMsgNo: channelID + "-" + strconv.FormatUint(uint64(seq), 10),
+				FromUID:     channelID,
+				ChannelID:   channelID,
+				ChannelType: common.ChannelTypePerson.Uint8(),
+				Timestamp:   int32(ts),
+				IsDeleted:   0,
+				Payload:     []byte(`{"type":1,"content":"hi"}`),
+			},
+		},
+	}
+}
+
+// setupConvSyncE2E wires a test server whose IM is the shared fake (returning
+// the given conversation slice), with a SuperAdmin token (needed for the
+// system_setting write) that also authenticates the conversation/sync call. It
+// resets the per-uid rate-limit bucket and the cursor Redis keys so each test
+// starts clean, and pins MessageSaveAcrossDevice off so the syncack
+// cursor-persist path is exercised deterministically.
+func setupConvSyncE2E(t *testing.T, convs []*config.SyncUserConversationResp) (*server.Server, *config.Context) {
+	t.Helper()
+	t.Setenv(masterKeyEnvConvE2E, "0123456789abcdef0123456789abcdef")
+
+	// Point the IM at the shared fake and load this test's conversation slice
+	// before any NewTestServer so the very first ctx already sees the live URL.
+	fakeIMConvs = convs
+	imURL := sharedFakeIM().URL
+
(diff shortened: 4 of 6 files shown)


Jerry-Xin (commented):
One-line summary: The PR is in scope for octo-server, but it currently breaks the integration test build.

🔴 Blocking

🔴 Critical — `loadRecentCutoffs` was promoted from a `*Sidebar` method to a package-level function, but the integration tests still call the removed method. This makes `go test -tags=integration ./modules/message` fail to compile:
[modules/message/api_sidebar_recent_filter_e2e_test.go:60](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-hp7lixa8/repo/modules/message/api_sidebar_recent_filter_e2e_test.go:60)
[modules/message/api_sidebar_recent_filter_e2e_test.go:106](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-hp7lixa8/repo/modules/message/api_sidebar_recent_filter_e2e_test.go:106)

The calls should be updated to `loadRecentCutoffs(ctx, now)` or equivalent.

💬 Non-blocking

🟡 Warning — The new conversation filter does not exactly mirror sidebar recent behavior when `X-Space-ID` is present and `sidebar.recent_filter_person_days` is non-zero. In `/v1/conversation/sync`, filtering runs before `EnsureSystemBotsPresent`, so system bot placeholders can be re-added after the recent cutoff filter:
[modules/message/api_conversation.go:749](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-hp7lixa8/repo/modules/message/api_conversation.go:749)
[modules/message/api_conversation.go:867](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-hp7lixa8/repo/modules/message/api_conversation.go:867)

That may be intentional for bot visibility, but it conflicts with the stated “same as sidebar recent tab” rule. Please either document the exception or add a focused test locking in the intended behavior.

✅ Highlights

The opt-in flag preserves default `/v1/conversation/sync` behavior, and the cursor is still advanced from raw conversations before response filtering, which avoids the known filtered-max-version cursor stall.

lml2468 (dismissed):
APPROVED (head `567d45310072`).

Read the full diff against the surrounding seam (`syncUserConversation`, `fillPersonSpaceUnread`, `EnsureSystemBotsPresent`, the package-level `loadRecentCutoffs` move) and the design call is the right one: **opt-in flag instead of unconditional filter**. The PR description spells out *why* the issue's suggested unconditional fix would have been a regression for `/v1/conversation/sync` (universal sync, no `tab` marker, `version=0` initial sync would advance the cursor past dropped items and they'd be permanently invisible until next activity). Resisting that one-line "obvious" fix is the right call.

## ✅ Independently verified

- **Default behaviour is byte-for-byte preserved.** `RecentFilter` defaults to `false`; `if req.RecentFilter` is the only gate. No existing client sends the flag, so mobile / desktop / older web build → no change to response shape, cursor, unread, or ordering.

- **Cursor invariant intact.** `lastVersion := maxConversationVersion(conversations, req.Version)` runs on the **raw** slice before filtering. The new comment block calls out the PR-B (#1377) / sidebar-B1 lineage explicitly — same class of bug, same fix shape. A filtered-out max-version conversation cannot stall the cursor.

- **Per-Space unread invariant intact.** `fillPersonSpaceUnread(syncUserConversationResps, conversations, ...)` takes both the response slice (to enrich) AND the raw `conversations` slice (for unread counts). The filter only narrows the response slice; the unread tally still reads `conversations`. Confirmed by call-site at line 856.

- **System bots invariant intact.** `EnsureSystemBotsPresent` runs at line 867, after the filter, in the Space-filter block. So even when an operator sets a `person` window (which would normally affect DM-type entries including system bots), the bot-presence backstop re-introduces them. Filter cannot silently drop system bots from the Recent tab.

- **`filterRecentConversations` is allocation-safe and non-mutating.** Returns a fresh `make([]*SyncUserConversationResp, 0, len(resps))` slice, never reorders or mutates the input. `TestFilterRecentConversations_DoesNotMutateInput` pins this with a post-call length+element check on the input — exactly the right shape for an immutability test.

- **`cutoff == 0` short-circuit.** The filter expression `if cutoff := cutoffs.cutoffFor(r.ChannelType); cutoff != 0 && r.Timestamp <= cutoff` correctly skips filtering for any channel type whose window is 0. Without the `cutoff != 0 &&` guard, any conversation with `Timestamp <= 0` (stub rows / unset timestamp) could accidentally be dropped. The guard is the same one #291's `buildRecentItems` carries — same hazard, same fix.

- **`loadRecentCutoffs` promotion to package-level is clean.** Old `(sb *Sidebar) loadRecentCutoffs(now)` became `loadRecentCutoffs(ctx *config.Context, now time.Time)`; the Sidebar caller updated to `loadRecentCutoffs(sb.ctx, time.Now())` (verified `api_sidebar.go:529`). Both `Sidebar.Sync` and `Conversation.syncUserConversation` live in package `message`, so promoting it gives one window-resolution path with one set of cutoffs semantics rather than drifting copies.

- **Test coverage is the right shape.** Seven cases: stale group / stale thread / DM kept when `person=0` / `group=0` keeps all / `person` window filters DMs / unknown type kept / empty input / no-mutate. Each test asserts on `channelIDs(got)` so any reordering would also fail — additional bonus coverage.

## 🟡 Non-blocking — integration-tag e2e file references the removed method

`modules/message/api_sidebar_recent_filter_e2e_test.go` carries `//go:build integration` and still calls `sb.loadRecentCutoffs(now)` at lines 60 and 106 — but the receiver-method form was just removed in this PR (only the package-level `loadRecentCutoffs(ctx, now)` remains). Default `go test ./...` Test job doesn't compile this file (build tag excludes it), so CI stays green; but anyone running `go test -tags=integration ./modules/message/...` locally or in an integration lane will hit a compile error.

Either:
- One-line fix in this PR: change both `sb.loadRecentCutoffs(now)` → `loadRecentCutoffs(sb.ctx, now)` (the file already has `sb`, so the change is local), OR
- File a follow-up issue and acknowledge in the PR description.

I'd lean toward fixing in-PR because it's a literal compile error sitting one rebase away — same class of "looks fine because CI doesn't run it" trap as #216's diff-coverage no-op gate, just smaller blast radius. But it's strictly a refactor-completeness nit, not a behaviour problem; if there is no integration CI lane today, the cost of deferring is one developer's confused day.

## 🟢 Non-blocking nit (style)

`filterRecentConversations` uses `make(..., 0, len(resps))` then `append`. In the "filter drops nothing" common case that pre-allocates exactly enough capacity. In the "filter drops everything" extreme it over-allocates by `len(resps)`, but the slice is short-lived (one request) and the allocation is bounded by the response cap anyway. Standard idiomatic shape; just noting that the linear branch isn't a hot-path concern.

## ✅ Architecture / forward note

The PR description's "long-term direction is for the web Recent tab to consume `/v1/sidebar/sync tab=recent` and retire this opt-in" is the right end state — that endpoint is purpose-built, already filters, and doesn't have the "universal sync, no tab marker" trade-off. Worth filing the retire-opt-in plan as a tracked follow-up tied to the frontend migration, so this flag doesn't become permanent shim debt. Not blocking this PR.

## What I did NOT verify

- Web frontend actually setting `recent_filter: true` (PR description acknowledges this is a separate frontend change).
- Behaviour of `EnsureSystemBotsPresent` under a hypothetical `person=N` window where a system bot row was just filtered out — I'm trusting the call-site ordering at line 867 (after filter, in Space block) plus the PR description's claim that this backstop re-introduces them. Worth a one-line e2e test in a follow-up that combines `RecentFilter: true` + `person_days > 0` + system bot presence assertion if anyone ever enables the `person` window.

## Checks

Build / Lint / Vet / Test / CodeQL / osv / i18n all green on `567d45310072`. `check-sprint` red is the project-board freshness check, not code (same as prior PRs).

LGTM. Worth fixing the integration-test method reference before merge if you want to keep the integration lane buildable; otherwise this is ready.

yujiawei (dismissed):
# Code Review — PR #295 (octo-server)

**Verdict: APPROVED.** The change is well-scoped, correct, and the documented invariants hold up under inspection. The two items below are non-blocking (P2) — a comment accuracy fix and an optional defensive hardening for a non-default admin configuration.

## Summary

Adds an **opt-in** `recent_filter` boolean to `POST /v1/conversation/sync`. When `true`, the response conversation list is filtered through the same per-channel-type activity windows the sidebar recent tab uses (`recentCutoffs` / `cutoffFor`); when `false` (default) behavior is byte-for-byte unchanged. `loadRecentCutoffs` is promoted from a `*Sidebar` method to a package-level function so both handlers share one window-resolution path. This is the right design: filtering unconditionally on this universal offline/initial-sync endpoint would silently and permanently drop quiet conversations from mobile/desktop full syncs.

## Verification

- ✅ **Build & vet**: `go build ./modules/message/` and `go vet ./modules/message/` both pass clean at head SHA `567d453`.
- ✅ **Unit tests**: all 8 `TestFilterRecentConversations_*` cases pass.
- ✅ **Cursor invariant**: `filterRecentConversations` runs *after* `lastVersion = maxConversationVersion(rawConversations, …)` (api_conversation.go:729-735, filter at 749-752). A filtered-out max-version conversation cannot stall the cursor — same fix class as #1377. Confirmed.
- ✅ **Per-Space unread**: `fillPersonSpaceUnread` reads the raw `conversations` slice (api_conversation.go:856), not the filtered response; the filter returns a new slice and never mutates input (test `..._DoesNotMutateInput`). Confirmed.
- ✅ **Units**: both `Timestamp` and the cutoff are unix **seconds** (`daysCutoff` uses `.Unix()`; conversation `Timestamp` is "最后一次会话时间" in seconds). Comparison is sound.
- ✅ **Unknown / DM defaults**: `cutoffFor` returns `0` (no filter) for unknown types and for the `person` window whose default is `0`, so DMs are unfiltered by default. Confirmed against `defaultSidebarRecentFilterPersonDays = 0`.

## Findings

### P2-1 — Inline comment over-states the system-bot guarantee

The comment at `modules/message/api_conversation.go:747-748` states the system-bot fallback (`EnsureSystemBotsPresent`) runs after the filter "因此即便开启 person 窗口也不会把系统 Bot 误删". This is only true when a Space filter is active: the filter runs unconditionally on `req.RecentFilter` (L749-752), but `EnsureSystemBotsPresent` is called **only inside `if hasSpaceFilter {`** (L854-867).

So on the **no-`X-Space-ID`** request path, *if* an admin sets `sidebar.recent_filter_person_days > 0` (non-default) *and* a system-bot DM is stale, the bot entry would be dropped with nothing restoring it. By default this is harmless (`person` window defaults to `0` → DMs never filtered), which is why this is P2 and not a blocker. Recommend correcting the comment to note the guarantee is conditional on `hasSpaceFilter`.

### P2-2 — (optional) Defensive system-bot restore on the no-space path

Following from P2-1: to fully close the gap regardless of admin config, either run an `EnsureSystemBotsPresent` pass after `filterRecentConversations` on the `RecentFilter` path independent of `hasSpaceFilter`, or have `filterRecentConversations` skip the cutoff for `IsSystemBot` channels. Not required for merge given the default-safe config, but it would make the safety property unconditional and match the comment.

### P2-3 (note, pre-existing) — ack offset cache is not filtered

Worth flagging for awareness, not as a change request: when `MessageSaveAcrossDevice=false`, conversations are appended to `syncConversationResultCacheMap` (api_conversation.go:711-723) *before* the filter drops them, and `syncUserConversationAck` later advances `user_last_offset` for every cached channel (L933-963) — including filtered-out ones. This is the **same structural pattern** the existing Space filter already uses (it too runs after cache population), so the PR introduces no new regression. Impact is bounded to a possibly-stale Recents *preview* for a hidden conversation: `user_last_offset` only seeds the conversation-list preview cursor (`IMSyncUserConversation`), and is independent of the per-channel history endpoint (`syncChannelMessage`) and the authoritative `channel_offset` table — **no message data loss**. A one-line comment noting the cache/response asymmetry would help future readers.

## Nits

- `recent_filter` semantics are clearly documented in both the Go struct and `swagger/conversation.yaml`. Good.
- Test coverage is thorough for the pure helper (stale group/thread, DM kept/filtered, cutoff-0, unknown type, empty, no-mutation).

None of the above block merge. Nice, careful change with strong inline rationale tying back to #291 / #1377.

Jerry-Xin (approved):
This PR is in scope for `octo-server`: it updates the `message` module’s conversation sync behavior and related API documentation/tests.

**🔴 Blocking**
None.

**💬 Non-blocking**
- 🟡 Warning: `modules/message/api_conversation.go:747-748` says system bots will not be removed because `EnsureSystemBotsPresent` runs later, but that only happens inside the `hasSpaceFilter` branch at `modules/message/api_conversation.go:855-867`. A `recent_filter=true` request without a validated Space can still filter out a stale system-bot DM if `sidebar.recent_filter_person_days` is non-zero. If the new flag is only meant for Space-scoped Web Recent calls, this is fine, but the comment/API contract should be narrowed or covered with a test.

- 🔵 Suggestion: `modules/message/conversation_recent_filter_e2e_test.go:55-58` uses mutable package-global fake IM state without a lock. The current tests do not call `t.Parallel`, so this is not a present race, but a tiny mutex around `fakeIMConvs` would make future integration-test changes safer.

**✅ Highlights**
- The PR correctly keeps the filter opt-in via `recent_filter`, preserving default `/v1/conversation/sync` behavior.
- Cursor advancement still uses raw IM conversations before filtering at `modules/message/api_conversation.go:729-751`, which avoids the max-version stall risk.
- The filter helper mirrors sidebar cutoff semantics and returns a new slice without mutating input.
- Focused unit coverage for cutoff behavior is good.

**Verification**
- `go test ./modules/message -run TestFilterRecentConversations` passed.
- `go test ./modules/message` could not complete locally because Redis was unavailable at `127.0.0.1:6379`, which matches this package’s integration-style test dependency.

lml2468 (dismissed):
APPROVED (head `3f6e7be101de`).

Author addressed the R1 integration-tag nit cleanly **and** went further by adding three full-stack e2e cases for `/v1/conversation/sync`'s recent-filter — covering exactly the three invariants I called out in R1.

## ✅ R1 nit fixed

`api_sidebar_recent_filter_e2e_test.go` at lines 55–60 and 100–105: both `sb := &Sidebar{ctx: ctx}` / `sb.loadRecentCutoffs(now)` call sites collapsed to a direct `loadRecentCutoffs(ctx, now)`. The integration-tag file now builds under `go test -tags=integration ./modules/message/...`. Two-line fix, no incidental changes.

## ✅ New e2e coverage hits the three R1 invariants

The 256-line `conversation_recent_filter_e2e_test.go` is doing exactly the right work — drives the real `syncUserConversation` HTTP handler with a tiny `httptest` WuKongIM stand-in (one of the cleaner test-seam choices I've seen for this module: `IMSyncUserConversation` is a concrete `*config.Context` POST, so pointing `WuKongIM.APIURL` at a fake server avoids needing an interface seam at all). Three cases, each precisely shaped:

**1. `TestE2E_ConvSync_FilterIsOptIn`** — admin sets `person_days = 7`, client calls **without** `recent_filter` → stale DM still returned. Proves default callers (mobile/desktop) are byte-for-byte unaffected by the new admin window. Pin #1 from my R1 verdict.

**2. `TestE2E_ConvSync_OptInFiltersStaleDM`** — same window, `recent_filter: true` → stale DM dropped, fresh DM kept. Proves the opt-in actually does the filtering. Pin #2.

**3. `TestE2E_ConvSync_CursorNotStalledByFilter`** — this one's the **best** test in the bunch: stale DM carries the **highest** version (200), and the test:
   - Filters via `recent_filter: true`, confirms `dm-stale` not in response
   - Issues `/v1/conversation/syncack`
   - Reads `userMaxVersion:{uid}` from Redis
   - Asserts the persisted cursor advanced to **`staleMaxVersion = 200`** — i.e., the version of the conversation the filter dropped

That's the right shape for a cursor-advancement assertion: not "response doesn't contain stale", but "cursor went **past** stale even though stale was dropped". Anyone refactoring `syncUserConversation` later who accidentally computes `lastVersion` from the *filtered* slice (the obvious-looking but wrong fix) will hit this test red with a clear failure message ("cursor advanced to the filtered-out conversation's version (computed from raw)"). PR-B #1377 / sidebar B1 hazard locked down by an e2e, not just by code review.

## What I'd note (non-blocking)

- The system-bot interaction I called "worth a one-line e2e test in a follow-up" in R1 isn't in this batch — that's a Space-filter-path test (`person_days > 0` + `EnsureSystemBotsPresent` interaction). Author probably (correctly) judged it out of scope since the existing code review chain already relied on the `EnsureSystemBotsPresent` ordering at line 867. Fine to leave for the next time someone actually enables the `person` window.
- The `WuKongIM.APIURL` redirect pattern in `sharedFakeIM` is a nice reusable bit for future conversation/sidebar e2e — worth documenting in a `modules/message/testutil`-style helper if more tests want the same mock seam. Pure organisational nit.

## Combined PR follow-through

| R1 point | R2 status |
|---|---|
| 🟡 (lml2468) integration-tag e2e file references removed method | ✅ fixed in this commit |
| 🟢 (lml2468) suggested e2e for cursor/unread/system-bot invariants | ✅ cursor + opt-in default + opt-in filtering covered; system-bot interaction acceptably deferred |
| 🟡 (yujiawei) `EnsureSystemBotsPresent` only runs in Space-filter path | not addressed; still defensible since the Person window default is 0 and operators turning it on can mitigate by also using Space filtering. Worth a follow-up |
| 🟢 (lml2468) long-term retire-opt-in via `/v1/sidebar/sync tab=recent` | architectural forward note; out of scope here |

## Checks

Build / Lint / Vet / i18n / osv / check-sprint (this round is green!) all pass. `Test` / `CodeQL` pending — including the new integration-tagged tests if CI runs them with `-tags=integration`, which I'd recommend confirming once. If the default Test job doesn't pass `-tags=integration`, the three new e2e cases stay as local-run-only proofs (still valuable for any developer touching this path).

LGTM.

Jerry-Xin (approved):
This PR is in scope for `octo-server`: it updates the `message` module’s conversation sync behavior and related API documentation/tests.

**🔴 Blocking**
None.

**💬 Non-blocking**
- 🟡 Warning: `modules/message/api_conversation.go:747-748` says system bots will not be removed because `EnsureSystemBotsPresent` runs later, but that only happens inside the `hasSpaceFilter` branch at `modules/message/api_conversation.go:855-867`. A `recent_filter=true` request without a validated Space can still filter out a stale system-bot DM if `sidebar.recent_filter_person_days` is non-zero. If the new flag is only meant for Space-scoped Web Recent calls, this is fine, but the comment/API contract should be narrowed or covered with a test.

- 🔵 Suggestion: `modules/message/conversation_recent_filter_e2e_test.go:55-58` uses mutable package-global fake IM state without a lock. The current tests do not call `t.Parallel`, so this is not a present race, but a tiny mutex around `fakeIMConvs` would make future integration-test changes safer.

**✅ Highlights**
- The PR correctly keeps the filter opt-in via `recent_filter`, preserving default `/v1/conversation/sync` behavior.
- Cursor advancement still uses raw IM conversations before filtering at `modules/message/api_conversation.go:729-751`, which avoids the max-version stall risk.
- The filter helper mirrors sidebar cutoff semantics and returns a new slice without mutating input.
- Focused unit coverage for cutoff behavior is good.

**Verification**
- `go test ./modules/message -run TestFilterRecentConversations` passed.
- `go test ./modules/message` could not complete locally because Redis was unavailable at `127.0.0.1:6379`, which matches this package’s integration-style test dependency.

Jerry-Xin (approved):
This PR is in scope for `octo-server`: it updates the `message` module’s conversation sync behavior and related API documentation/tests.

**🔴 Blocking**
None.

**💬 Non-blocking**
- 🟡 Warning: `modules/message/api_conversation.go:747-748` says system bots will not be removed because `EnsureSystemBotsPresent` runs later, but that only happens inside the `hasSpaceFilter` branch at `modules/message/api_conversation.go:855-867`. A `recent_filter=true` request without a validated Space can still filter out a stale system-bot DM if `sidebar.recent_filter_person_days` is non-zero. If the new flag is only meant for Space-scoped Web Recent calls, this is fine, but the comment/API contract should be narrowed or covered with a test.

- 🔵 Suggestion: `modules/message/conversation_recent_filter_e2e_test.go:55-58` uses mutable package-global fake IM state without a lock. The current tests do not call `t.Parallel`, so this is not a present race, but a tiny mutex around `fakeIMConvs` would make future integration-test changes safer.

**✅ Highlights**
- The PR correctly keeps the filter opt-in via `recent_filter`, preserving default `/v1/conversation/sync` behavior.
- Cursor advancement still uses raw IM conversations before filtering at `modules/message/api_conversation.go:729-751`, which avoids the max-version stall risk.
- The filter helper mirrors sidebar cutoff semantics and returns a new slice without mutating input.
- Focused unit coverage for cutoff behavior is good.

**Verification**
- `go test ./modules/message -run TestFilterRecentConversations` passed.
- `go test ./modules/message` could not complete locally because Redis was unavailable at `127.0.0.1:6379`, which matches this package’s integration-style test dependency.

Jerry-Xin (dismissed):
This PR is in scope for `octo-server`: it updates the `message` module’s conversation sync behavior and related API documentation/tests.

**🔴 Blocking**
None.

**💬 Non-blocking**
- 🟡 Warning: `modules/message/api_conversation.go:747-748` says system bots will not be removed because `EnsureSystemBotsPresent` runs later, but that only happens inside the `hasSpaceFilter` branch at `modules/message/api_conversation.go:855-867`. A `recent_filter=true` request without a validated Space can still filter out a stale system-bot DM if `sidebar.recent_filter_person_days` is non-zero. If the new flag is only meant for Space-scoped Web Recent calls, this is fine, but the comment/API contract should be narrowed or covered with a test.

- 🔵 Suggestion: `modules/message/conversation_recent_filter_e2e_test.go:55-58` uses mutable package-global fake IM state without a lock. The current tests do not call `t.Parallel`, so this is not a present race, but a tiny mutex around `fakeIMConvs` would make future integration-test changes safer.

**✅ Highlights**
- The PR correctly keeps the filter opt-in via `recent_filter`, preserving default `/v1/conversation/sync` behavior.
- Cursor advancement still uses raw IM conversations before filtering at `modules/message/api_conversation.go:729-751`, which avoids the max-version stall risk.
- The filter helper mirrors sidebar cutoff semantics and returns a new slice without mutating input.
- Focused unit coverage for cutoff behavior is good.

**Verification**
- `go test ./modules/message -run TestFilterRecentConversations` passed.
- `go test ./modules/message` could not complete locally because Redis was unavailable at `127.0.0.1:6379`, which matches this package’s integration-style test dependency.

lml2468 (approved):
APPROVED (head `686a0b568144`).

Pure docs commit. No behaviour change to `filterRecentConversations`, `loadRecentCutoffs`, `syncUserConversation`'s call order, or any test. Confirmed via the commit's lone file diff: `modules/message/api_conversation.go` `+6/-2`, all inside the comment block above the `if req.RecentFilter { ... }` branch. The R2 verified properties (cursor invariant via `maxConversationVersion` on raw, per-Space unread on raw, opt-in default false) are unchanged.

## ✅ Addresses yujiawei R2 precisely

The R2 comment was over-confident — "即便开启 person 窗口也不会把系统 Bot 误删" — but `EnsureSystemBotsPresent` only runs inside the `if hasSpaceFilter { ... }` block (verified `api_conversation.go:854-868` last round, unchanged). The new comment fixes both halves of the truth:

- **Real-world safe**: person window default = 0 → DM filter is a no-op → system bots untouched. Web Recent tab passes `space_id` → `EnsureSystemBotsPresent` does run → bots backfilled even if an operator turns the person window on.
- **Documented edge**: `no space_id + recent_filter=true + person_days > 0` is the one combination where a quiet system-bot DM could be dropped before any backfill. Called out explicitly so the next reader doesn't have to re-derive it from the call graph.

This is exactly the "tighten the doc to match the actual guarantee, don't widen the guarantee" shape — same discipline as the #293 admin-disable trade-off where the answer was "document the asymmetry honestly, don't pretend it doesn't exist." Cheaper than the alternative (refactor `EnsureSystemBotsPresent` to run on every path) and the one-line architectural cost ("Web Recent tab passes space_id, so it is safe") is paid in the comment rather than in extra code.

## Combined R1+R2+R3 follow-through

| Point | R3 status |
|---|---|
| 🟡 (lml2468 R1) integration-tag e2e references removed method | ✅ fixed in R2 |
| 🟢 (lml2468 R1) e2e for cursor/unread/system-bot invariants | ✅ R2 added cursor + opt-in default + opt-in filter cases |
| 🟡 (yujiawei R2) `EnsureSystemBotsPresent` only in Space-filter path | ✅ docs tightened to honestly describe the precondition + the single edge case |
| 🟡 (Jerry-Xin R2) `recent_filter` comment vs actual behaviour mismatch | partially addressed — Jerry-Xin's concern was the broader description, this commit narrows it for the system-bot half |
| 🟡 (Jerry-Xin R2) fake-IM global state needs locking | not addressed — pure-test concern, fine to defer |
| 🟢 (lml2468) long-term retire opt-in via `/v1/sidebar/sync tab=recent` | architectural forward note |

## Checks

No production diff, so CI from R2 carries over. Build / Lint / Vet / Test / CodeQL / osv / i18n already green on R2 — confirm same on R3 then merge.

LGTM. Honest docs on a small known asymmetry is the right move; full APPROVED.

Jerry-Xin (approved):
Project relevance gate passed: this PR modifies existing `octo-server` conversation/sidebar sync behavior and documentation.

No blocking issues found.

**💬 Non-blocking**
- 🔵 Suggestion: [modules/message/conversation_recent_filter_e2e_test.go:55](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-srqqwsu7/repo/modules/message/conversation_recent_filter_e2e_test.go:55) uses process-global fake IM state. That is acceptable while these integration tests remain serial, but it would become race-prone if future tests add `t.Parallel()` or overlap requests. A small mutex or per-test handler state would make this more robust.

**✅ Highlights**
- The opt-in flag preserves existing `/v1/conversation/sync` callers by default: [api_conversation.go:291](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-srqqwsu7/repo/modules/message/api_conversation.go:291).
- Cursor advancement remains based on raw IM conversations before recent filtering, avoiding sync stalls: [api_conversation.go:729](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-srqqwsu7/repo/modules/message/api_conversation.go:729).
- The filter reuses the sidebar cutoff resolution path, keeping behavior consistent: [api_sidebar.go:673](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-srqqwsu7/repo/modules/message/api_sidebar.go:673).
- Focused unit coverage exercises stale/fresh, disabled windows, DM behavior, unknown types, and no mutation.

**Verification**
- `go test ./modules/message -run TestFilterRecentConversations` passed.
- `go test ./modules/message` could not complete locally because Redis was unavailable at `127.0.0.1:6379`.

Issue #297, opened 2026-06-06
Incoming Webhook enhancements: rich text/image-text, platform adapters, management UX, delivery observability


an9xyz:
## Background

`modules/incomingwebhook/` is already solid on the security / rate-limiting / caching side, but its **message expressiveness** and **onboarding experience** lag behind Discord / Slack / WeCom group bots / Feishu custom bots:

- Pushed messages are **always `Text(1)`** (`api.go` `buildPayload`) — plain text / markdown only, no image-text mixing, no images/cards.
- Callers must convert third-party events into our `content` format themselves — there are **no platform adapters** (Discord ships `/github`, `/slack` suffixes that work out of the box).
- Weak onboarding / troubleshooting: no test push, no ready-made examples; the `incoming_webhook_audit` table **only recorded successes**, so a sender debugging a failure was flying blind.

## Goals & decisions (confirmed)

- **Delivery model**: staged, multiple PRs — each phase independently mergeable / revertable.
- **Rich-text contract**: a high-level external `blocks` abstraction (`{type:text}` / `{type:image,url,width,height}`), server-translated into the native `RichText(=14)` and run through `pkg/richtext` `Validate`/`Finalize`. The internal ContentType number and server-derived fields (`plain`/`space_id`) are **not** exposed to callers.
- **Adapter auth**: reuse the existing URL token (zero new fields). GitLab uses `X-Gitlab-Token` which must equal the URL token; GitHub relies on the 128-bit unguessable URL token + existing rate limits, no mandatory HMAC.

## Reused existing infrastructure

- `pkg/richtext` (`Validate`/`Finalize`) + octo-lib `common.ValidateRichTextPayload` / `RichTextBlock`: purpose-built for "self-validate before `SendMessage`", reusable directly on the push path.
- `modules/webhook/hmac.go` `VerifyHMACSHA256`: reference for GitHub `X-Hub-Signature-256` verification (if an HMAC option is added later).
- `incoming_webhook_audit` table + the bounded async `auditSem` write pattern: extend to record failures.
- `pkg/errcode` / `api_i18n.go` / `db.go` / `sql/` migration conventions reused as-is.

### Phase 1 — Rich text / image-text mixing (P1) — ✅ PR #298
- [x] Add `msg_type` + `blocks` to `pushPayloadReq` (omitted/`text` keeps the legacy plain-text path, fully backward compatible).
[...]

PR #232, opened 2026-06-02
feat(octo-server): RichText=14 server adapt — plain gen + push alert + search index (Phase 1)

yujiawei:
## 概述

图文混排 Phase 1：octo-server 侧对 RichText (ContentType=14) 的全链路适配。

升级 go.mod 中 octo-lib 到含 RichText schema/工具的版本（`v0.0.0-20260602143311-8c26a826e07b`，HEAD 8c26a826），并接入四处服务端 chokepoint。

## 改动

1. **plain 权威生成（入库/序列化出口）** — 新增 `pkg/richtext.EnsurePlain`：对 type=14 用 content 重算顶层 `plain` 覆盖客户端不可信 plain，内部经 `FillPlainBounded` 复检 1MB（超限拒发）；非 14 为 no-op。接入 user `/v1/message/send`（`modules/message.sendMessage`）与 robot `/message/send`（`modules/robot.sendMessage`）两个发送出口，使权威 plain 随消息一起落库/进 IM 搜索索引——这是下游 summary/matter/search/复制/推送 的前置。

2. **webhook 推送 alert** — `getMessageAlert` switch 新增 `common.RichText` case，用 `common.GetRichTextDisplayText` 生成推送正文；并把 RichText 加入 `getSupportTypes`（否则 `pushTo` 在 `containSupportType` 处就把 14 丢弃，永远到不了 alert）。

3. **搜索索引/高亮** — `modules/search` 搜索 payload 与 highlights 补 `payload.plain`，让承载在 blocks 内的 14 文字可被搜索与高亮。

4. **robot 发送校验** — `payloadIsVail` 的 RichText case 从浅检 `content != nil` 升级为 `common.ValidateRichTextPayload`（write-strict：大小上限/content 非空/block 结构合法）。

## 测试

- `pkg/richtext`：EnsurePlain（覆盖端 plain / 老字符串 content / 超限 error / no-op）+ IsRichTextPayload。
- `modules/webhook`：alert 用权威 plain、缺 plain 现场兜底、getSupportTypes 含 14。
- `modules/robot`：payloadIsVail 对合法/空 content/空数组/text 空/image 缺 url/scheme 非法/缺尺寸/未知 block 的判定。
- `modules/search`：buildMessageSearchQuery 含 payload.plain。

`go build ./...` / `go vet ./...` 全绿；新增包与改动包测试通过。`modules/message` 集成测试需本地 Redis（既有基础设施依赖，与本改动无关），改动经 EnsurePlain 单测全覆盖。

[...]

PR #298, opened 2026-06-06
feat(incomingwebhook): rich text / image-text blocks on push (Phase 1)

an9xyz:
## Background

**Phase 1** of epic #297. Today the incoming-webhook push endpoint always sends `Text(1)`, so it can only carry plain text / markdown — no mixed image-and-text messages. This PR adds rich-text / image-text support without breaking backward compatibility.

## Changes

- Add an optional `msg_type` + `blocks` to `pushPayloadReq`:
  - omitted or `"text"` → the existing plain-text path, **zero behavior change**.
  - `"richtext"` → `blocks` carries ordered image/text blocks (`{type:text}` / `{type:image,url,width,height}`).
- New `buildRichTextPayload`: whitelist-translates `blocks` into the native `RichText(=14)` payload, reusing `pkg/richtext` `Validate`/`Finalize` (rejects missing/empty content, empty text, non-http(s) image URL, missing image size, unknown block, >1MB; server authoritatively generates `plain`).
- Error mapping: >1MB → 413; other structural errors → 400 `reason=blocks`; unknown `msg_type` → 400 `reason=msg_type`.

## Design notes

- **High-level external contract**: callers only describe text/image blocks; the internal ContentType number and server-derived fields (`plain`/`size`) are not exposed.
- **Security baseline preserved**: `from.kind=webhook` and `space_id` are server-injected; all other caller fields (incl. `extra`) are dropped, preventing cross-Space forgery — identical to the text path.
- **Client coordination**: rich-text messages are standard `type=14`, so the existing RichText rendering path is reused with no new client work.

## Testing (TDD)

- Unit tests `richtext_test.go` (no infra, green locally): translation/validation, order preservation, from override, space_id non-forgeable, no token leak, every rejection path, block-count cap.
- e2e `richtext_push_test.go` (CI): richtext success / empty blocks / bad image / unknown msg_type / text backward-compat.
- README push-contract doc.

## Checks

`go build` / `go vet` / `gofmt` / unit tests / the `TestIncomingWebhookNoLegacyResponseError` guard / `make i18n-lint` / `make i18n-extract-check` all pass. DB/e2e tests depend on MySQL+Redis+WuKongIM and run in CI.

Refs #297

https://claude.ai/code/session_011kXG9RmQcEvmEp5ZZcPA9y

### DIFF
--- a/modules/incomingwebhook/README.md
+++ b/modules/incomingwebhook/README.md
@@ -15,6 +15,11 @@ Content-Type: application/json
 
 由 `msg_type` 选择，**缺省即纯文本**，与历史行为完全一致：
 
+> **兼容性提醒**：`msg_type` 现在严格校验——只接受省略 / `"text"` / `"richtext"`，其它非空值
+> （如 `"markdown"`）返回 400 `reason=msg_type`。历史上未知 JSON 字段会被忽略，因此若有旧
+> 客户端误带了别的 `msg_type` 值，升级后需要去掉它（带合法 `content` 也会被拒）。`msg_type`
+> 大小写不敏感（内部做了 lower+trim），但块 `type` 大小写敏感、须为精确小写（`text`/`image`）。
+
 ### 1. 纯文本（`msg_type` 省略或 `"text"`）
 
 `content` 必填，客户端按 markdown 渲染。
--- a/modules/incomingwebhook/README.md
+++ b/modules/incomingwebhook/README.md
@@ -58,8 +58,11 @@ Content-Type: application/json
 约束：
 
 - `blocks` 必填且非空；块数量上限默认 50（`DM_INCOMINGWEBHOOK_MAX_BLOCKS`）。
-- 整条消息序列化后 ≤ 1MB（图片仅 URL 引用，不内嵌）。
-- 请求体仍受 8KB body cap 约束（`DM_INCOMINGWEBHOOK_MAX_BYTES`）。
+- **实际生效的上限是 8KB body cap**（`DM_INCOMINGWEBHOOK_MAX_BYTES`）：请求体在解析前即被
+  截断，超出按 413 拒绝。由于图片仅 URL 引用（不内嵌 base64），8KB 足以承载数十个文本/
+  图片块；多图文消息请用 URL 引用，不要内联大体积内容。
+- 服务端另有 1MB 的 RichText 硬上限（octo-lib 契约）兜底，但在默认 8KB body cap 下不会
+  先触达——它是上调 body cap 后才会成为约束的二级护栏。
 
 ## 通用字段与安全
 
--- a/modules/incomingwebhook/README.md
+++ b/modules/incomingwebhook/README.md
@@ -1,0 +1,80 @@
+# Incoming Webhook 推送契约
+
+外部服务通过带 token 的 URL 向指定群推送消息。管理端点（创建/列出/更新/删除/重置）
+由群主或管理员调用，详见 `api.go`。本文聚焦**推送端点**的请求契约。
+
+```
+POST /v1/incoming-webhooks/:webhook_id/:token
+Content-Type: application/json
+```
+
+鉴权走 URL 内的 token（SHA-256 存储、常量时间比对），无需登录态。所有鉴权失败统一
+返回 401（反枚举），并受多层限流约束。
+
+## 消息形态
+
+由 `msg_type` 选择，**缺省即纯文本**，与历史行为完全一致：
+
+### 1. 纯文本（`msg_type` 省略或 `"text"`）
+
+`content` 必填，客户端按 markdown 渲染。
+
+```json
+{
+  "content": "Build **#123** passed ✅  https://ci.example.com/123",
+  "username": "CI Bot",
+  "avatar_url": "https://example.com/ci.png"
+}
+```
+
+- `content`：必填，非空；语义长度上限 4000 rune（`DM_INCOMINGWEBHOOK_MAX_CONTENT_RUNES`）。
+- `username` / `avatar_url`：可选，覆盖该条消息的展示发送者名/头像（不改 webhook 本身配置）。
+
+### 2. 富文本 / 图文混排（`msg_type` = `"richtext"`）
+
+`blocks` 承载**有序**的图文块，数组顺序即图文穿插顺序。服务端翻译为内部 RichText 消息，
+客户端复用既有富文本渲染链路。
+
+```json
+{
+  "msg_type": "richtext",
+  "blocks": [
+    { "type": "text",  "text": "Build #123 passed ✅" },
+    { "type": "image", "url": "https://example.com/chart.png", "width": 800, "height": 400 },
+    { "type": "text",  "text": "耗时 42s" }
+  ],
+  "username": "CI Bot",
+  "avatar_url": "https://example.com/ci.png"
+}
+```
+
+块类型：
+
+| `type`  | 必填字段 | 约束 |
+|---------|----------|------|
+| `text`  | `text`   | 非空（纯文本，不渲染 markdown） |
+| `image` | `url`、`width`、`height` | `url` 仅接受 `http`/`https`（禁 `data:`/`base64`）；`width`/`height` 必须 > 0（供端上占位排版，避免抖动） |
+
+约束：
+
+- `blocks` 必填且非空；块数量上限默认 50（`DM_INCOMINGWEBHOOK_MAX_BLOCKS`）。
+- 整条消息序列化后 ≤ 1MB（图片仅 URL 引用，不内嵌）。
+- 请求体仍受 8KB body cap 约束（`DM_INCOMINGWEBHOOK_MAX_BYTES`）。
+
+## 通用字段与安全
+
+- `username` / `avatar_url`：两种形态通用，服务端裁剪到字节上限（名 64B / 头像 255B）。
+- 其它任意字段（含 `extra`、`space_id`）一律**丢弃**：消息归属的 Space 由服务端从群派生，
+  不接受调用方覆盖，防止伪造到其它 Space。
+
+## 响应
+
+| 场景 | HTTP | 说明 |
+|------|------|------|
+| 成功 | 200 | `{"status":0,"message_id":<int>}` |
+| 鉴权失败 | 401 | 统一响应，不区分原因（反枚举） |
+| 限流 | 429 | 带 `Retry-After` |
+| 请求非法 | 400 | `details.reason` ∈ `body`/`json`/`content`/`blocks`/`msg_type` |
+| 体积过大 | 413 | 超 body cap 或富文本 >1MB |
+| 投递失败 | 502 | 下游发送失败 |
+| 功能停用 | 404 | 全局开关 `incomingwebhook.enabled=0` |
--- a/modules/incomingwebhook/api.go
+++ b/modules/incomingwebhook/api.go
@@ -766,19 +766,43 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 		pushPayloadInvalid(c, "json")
 		return
 	}
-	if strings.TrimSpace(req.Content) == "" {
-		pushPayloadInvalid(c, "content")
-		return
-	}
-	// content 语义长度上限（按 rune 计），独立于 8KB 字节 body cap：防止单条消息正文
-	// 过长污染所有客户端渲染。超限按 413 拒绝，与 body 超限同语义。
-	if utf8.RuneCountInString(req.Content) > maxContentRunes() {
-		pushPayloadTooLarge(c)
+
+	// 5) 按 msg_type 构造 payload。缺省/"text" 走历史纯文本路径（content 必填，客户端
+	//    按 markdown 渲染），完全向后兼容；"richtext" 走图文混排：blocks 翻译为 octo
+	//    原生 RichText(=14) 并由 richtext.Validate/Finalize 权威校验。
+	var payload map[string]interface{}
+	switch strings.ToLower(strings.TrimSpace(req.MsgType)) {
+	case "", msgTypeText:
+		if strings.TrimSpace(req.Content) == "" {
+			pushPayloadInvalid(c, "content")
+			return
+		}
+		// content 语义长度上限（按 rune 计），独立于 8KB 字节 body cap：防止单条消息
+		// 正文过长污染所有客户端渲染。超限按 413 拒绝，与 body 超限同语义。
+		if utf8.RuneCountInString(req.Content) > maxContentRunes() {
+			pushPayloadTooLarge(c)
+			return
+		}
+		payload = buildPayload(m, &req)
+	case msgTypeRichText:
+		p, err := buildRichTextPayload(m, &req)
+		if err != nil {
+			// 仅 >1MB 映射 413（与 body/content 超限同语义）；其余结构性非法（空 content /
+			// 空 text 块 / 非 http(s) 图片 url / 缺图片宽高 / 未知块类型 / 超块数上限）
+			// 一律 400 invalid，reason=blocks 供调用方定位。
+			if errors.Is(err, common.ErrRichTextPayloadTooLarge) {
+				pushPayloadTooLarge(c)
+				return
+			}
+			pushPayloadInvalid(c, "blocks")
+			return
+		}
+		payload = p
+	default:
+		pushPayloadInvalid(c, "msg_type")
 		return
 	}
 
-	// 5) 构造 payload 并发送
-	payload := buildPayload(m, &req)
 	resp, err := w.ctx.SendMessageWithResult(&config.MsgSendReq{
 		// RedDot=1 让 webhook 消息触发未读红点和推送，与 botfather/robot 一致。
 		Header:      config.MsgHeader{RedDot: 1},
@@ -845,21 +869,12 @@ func truncateUTF8(s string, max int) string {
 //   - req.Username / req.AvatarURL 服务端裁剪到 create 侧同样的字节上限。push 路径
 //     原本只受 8KB body cap 约束，调用方可塞 KB 级字符串污染所有客户端 from.* 渲染。
 func buildPayload(m *incomingWebhookModel, req *pushPayloadReq) map[string]interface{} {
-	name := req.Username
-	if name == "" {
-		name = m.Name
-	}
-	avatar := req.AvatarURL
-	if avatar == "" {
-		avatar = m.Avatar
-	}
-	name = truncateUTF8(name, maxFromNameBytes)
-	avatar = truncateUTF8(avatar, maxFromAvatarBytes)
+	name, avatar := resolveFromIdentity(m, req)
 	return map[string]interface{}{
 		"type":    int(common.Text),
 		"content": req.Content,
 		"from": map[string]interface{}{
-			"kind":       "webhook",
+			"kind":       extraKindValue,
 			"webhook_id": m.WebhookID,
 			"name":       name,
 			"avatar":     avatar,
@@ -868,6 +883,22 @@ func buildPayload(m *incomingWebhookModel, req *pushPayloadReq) map[string]inter
 		// 防止 webhook 消息被伪造到其他 Space。
 		"space_id": m.SpaceID,
 	}
+}
+
+// resolveFromIdentity 解析 webhook 消息的展示发送者名/头像：调用方在 push 请求里
+// 覆盖（Username/AvatarURL）优先，否则回落到 webhook 自身配置；两者都裁剪到与
+// create 侧一致的字节上限，防止 push 路径成为绕过列长度约束的旁路。文本与富文本
+// 两条路径共用此函数，保证 from.* 渲染口径一致。
+func resolveFromIdentity(m *incomingWebhookModel, req *pushPayloadReq) (name, avatar string) {
+	name = req.Username
+	if name == "" {
+		name = m.Name
+	}
+	avatar = req.AvatarURL
+	if avatar == "" {
+		avatar = m.Avatar
+	}
+	return truncateUTF8(name, maxFromNameBytes), truncateUTF8(avatar, maxFromAvatarBytes)
 }
 
 // submitRecordSuccess 把审计任务投递给有界并发池：未达上限时异步执行；已达上限时
--- a/modules/incomingwebhook/api.go
+++ b/modules/incomingwebhook/api.go
@@ -785,6 +785,9 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 		}
 		payload = buildPayload(m, &req)
 	case msgTypeRichText:
+		// 注意：richtext 路径【不】套用纯文本的 maxContentRunes(4000) 语义上限——富文本
+		// 由块结构 + 1MB 序列化上限约束，默认 8KB body cap 下不可能逾越。这是与文本路径的
+		// 有意不对称（若运维上调 body cap，富文本仍受 1MB 兜底，不会无界）。
 		p, err := buildRichTextPayload(m, &req)
 		if err != nil {
 			// 仅 >1MB 映射 413（与 body/content 超限同语义）；其余结构性非法（空 content /
--- a/modules/incomingwebhook/api_i18n.go
+++ b/modules/incomingwebhook/api_i18n.go
@@ -36,7 +36,8 @@ func pushRateLimited(c *wkhttp.Context) {
 }
 
 // pushPayloadInvalid returns 400 for unreadable body / malformed JSON / empty
-// content; reason ∈ {body, json, content} is surfaced via the safe-listed
+// content / malformed rich-text blocks / unknown msg_type; reason ∈
+// {body, json, content, blocks, msg_type} is surfaced via the safe-listed
 // Details key so callers can tell what to fix.
 func pushPayloadInvalid(c *wkhttp.Context, reason string) {
 	httperr.ResponseErrorLWithStatus(c, errcode.ErrIncomingWebhookPushPayloadInvalid, nil, i18n.Details{"reason": reason})
--- a/modules/incomingwebhook/api_i18n_test.go
+++ b/modules/incomingwebhook/api_i18n_test.go
@@ -117,3 +117,27 @@ func TestIncomingWebhookRespondHelpers(t *testing.T) {
 		})
 	}
 }
+
+// TestPushPayloadInvalidSurfacesReason locks the external contract that
+// pushPayloadInvalid surfaces details.reason in the i18n error envelope. This is
+// asserted HERE (with the i18n ErrorRenderer wired, as in production via main.go)
+// rather than in the full e2e harness: testutil.NewTestServer renders only the
+// legacy {msg,status} shape, so an e2e body never carries error.details — see the
+// note in richtext_push_test.go. No DB/Redis needed; this only exercises the renderer.
+func TestPushPayloadInvalidSurfacesReason(t *testing.T) {
+	for _, reason := range []string{"blocks", "msg_type", "content", "json"} {
+		t.Run(reason, func(t *testing.T) {
+			r := iwhHelperHarness(func(c *wkhttp.Context) { pushPayloadInvalid(c, reason) })
+			req := httptest.NewRequest(http.MethodGet, "/probe", nil)
+			rec := httptest.NewRecorder()
+			r.ServeHTTP(rec, req)
+
+			if rec.Code != http.StatusBadRequest {
+				t.Fatalf("status = %d, want 400; body=%s", rec.Code, rec.Body.String())
+			}
+			if !strings.Contains(rec.Body.String(), reason) {
+				t.Fatalf("body must surface details.reason=%q; body=%s", reason, rec.Body.String())
+			}
+		})
+	}
+}
--- a/modules/incomingwebhook/model.go
+++ b/modules/incomingwebhook/model.go
@@ -64,6 +64,9 @@ type pushPayloadReq struct {
 // 不让调用方感知内部 ContentType 编号或 plain/size 等服务端派生字段。
 //   - type=="text"  ：Text 必填且非空。
 //   - type=="image" ：URL（仅 http/https）+ Width/Height（>0）必填，供端上占位排版。
+//
+// ⚠️ 新增块类型时务必与 pkg/richtext（及其底层 common.RichTextBlock 支持的类型）保持
+// 同步：buildRichTextPayload 按白名单逐类翻译，未在此显式支持的块类型会被 Validate 拒绝。
 type webhookBlock struct {
 	Type   string `json:"type"`
 	Text   string `json:"text,omitempty"`
--- a/modules/incomingwebhook/model.go
+++ b/modules/incomingwebhook/model.go
@@ -44,11 +44,32 @@ type auditModel struct {
 }
 
 // pushPayloadReq 推送端点的请求体。
+//
+// 兼容两种消息形态，由 MsgType 选择（缺省/"text" 走纯文本，与历史行为一致）：
+//   - 纯文本：Content 必填，客户端按 markdown 渲染（历史契约不变）。
+//   - 富文本/图文混排（MsgType=="richtext"）：Blocks 承载有序图文块，服务端翻译为
+//     octo 原生 RichText(=14) 后走 richtext.Validate/Finalize。对外刻意不暴露内部
+//     ContentType 数字与 plain 等服务端字段——调用方只需描述「文本块 / 图片块」。
 type pushPayloadReq struct {
+	MsgType   string                 `json:"msg_type,omitempty"`
 	Content   string                 `json:"content"`
+	Blocks    []webhookBlock         `json:"blocks,omitempty"`
 	Username  string                 `json:"username,omitempty"`
 	AvatarURL string                 `json:"avatar_url,omitempty"`
 	Extra     map[string]interface{} `json:"extra,omitempty"`
+}
+
+// webhookBlock 是富文本消息的单个有序块（对外契约）。字段刻意与 octo-lib
+// common.RichTextBlock 对齐但独立声明：对外只暴露 text/image 两类块所需的字段，
+// 不让调用方感知内部 ContentType 编号或 plain/size 等服务端派生字段。
+//   - type=="text"  ：Text 必填且非空。
+//   - type=="image" ：URL（仅 http/https）+ Width/Height（>0）必填，供端上占位排版。
+type webhookBlock struct {
+	Type   string `json:"type"`
+	Text   string `json:"text,omitempty"`
+	URL    string `json:"url,omitempty"`
+	Width  int    `json:"width,omitempty"`
+	Height int    `json:"height,omitempty"`
 }
 
 // createReq 管理端创建 webhook 的请求体。
--- a/modules/incomingwebhook/richtext.go
+++ b/modules/incomingwebhook/richtext.go
@@ -1,0 +1,104 @@
+package incomingwebhook
+
+import (
+	"errors"
+	"os"
+	"strconv"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-server/pkg/richtext"
+)
+
+// 推送消息形态（pushPayloadReq.MsgType 取值）。缺省与 "text" 等价，保持历史纯文本
+// 契约不变；"richtext" 走图文混排 blocks 路径。
+const (
+	msgTypeText     = "text"
+	msgTypeRichText = "richtext"
+)
+
+// 富文本 block 数量上限。8KB 的 body cap 已是天然约束，这里再加一道显式上限，避免
+// 调用方用海量空块构造病态 payload（每块都要进 Validate 遍历）。可经 env 调整。
+const (
+	envMaxBlocks     = "DM_INCOMINGWEBHOOK_MAX_BLOCKS"
+	defaultMaxBlocks = 50
+)
+
+// errTooManyBlocks blocks 数量超过 maxBlocks 上限。映射为 400 invalid（reason=blocks），
+// 而非 413：413 语义是「字节/体积过大」，块数超限是结构性非法，归 invalid 更准确。
+var errTooManyBlocks = errors.New("incomingwebhook: too many richtext blocks")
+
+func maxBlocks() int {
+	if v := os.Getenv(envMaxBlocks); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n > 0 {
+			return n
+		}
+	}
+	return defaultMaxBlocks
+}
+
+// buildRichTextPayload 把对外的 webhookBlock 数组翻译为 octo 原生 RichText(=14)
+// 消息 payload，并在发送前做权威校验/收尾：
+//
+//   - 逐块按白名单复制字段（text 块只取 text；image 块只取 url/width/height），
+//     绝不透传调用方的其它任意字段——与 buildPayload 丢弃 req.Extra 的安全基线一致；
+//   - from.kind=webhook + 服务端派生的 space_id 注入方式与纯文本路径完全相同，
+//     保证客户端识别 webhook 发送者、且消息不可被伪造到其它 Space；
+//   - richtext.Validate 跑 write-strict 校验（缺/空 content、空 text 块、非 http(s)
+//     图片 URL、缺图片宽高、未知 block 类型、>1MB 等一律拒绝），是块合法性的权威闸；
+//   - richtext.Finalize 在注入完所有服务端字段后用 content 重算权威顶层 plain（供
+//     搜索/推送/摘要/复制复用），并对真正出站的完整 payload 复检 1MB 上限。
+//
+// 校验失败返回错误，由 push 路径映射：common.ErrRichTextPayloadTooLarge → 413，
+// 其余（含 errTooManyBlocks 与 Validate 的各类结构错误）→ 400 invalid(reason=blocks)。
+func buildRichTextPayload(m *incomingWebhookModel, req *pushPayloadReq) (map[string]interface{}, error) {
+	if len(req.Blocks) > maxBlocks() {
+		return nil, errTooManyBlocks
+	}
+	// 空 blocks 不在此处单独报错：构造空 content 交由 richtext.Validate 统一拦
+	// （common.ErrRichTextEmptyContent），保持「块合法性单一权威 = Validate」。
+	content := make([]map[string]interface{}, 0, len(req.Blocks))
+	for _, b := range req.Blocks {
+		switch b.Type {
+		case common.RichTextBlockText:
+			content = append(content, map[string]interface{}{
+				"type": common.RichTextBlockText,
+				"text": b.Text,
+			})
+		case common.RichTextBlockImage:
+			content = append(content, map[string]interface{}{
+				"type":   common.RichTextBlockImage,
+				"url":    b.URL,
+				"width":  b.Width,
+				"height": b.Height,
+			})
+		default:
+			// 未知块类型：原样带上 type 交给 Validate 拒绝（ErrRichTextUnknownBlock），
+			// 不在此处复制任何其它字段，避免未知块夹带任意字段。
+			content = append(content, map[string]interface{}{"type": b.Type})
+		}
+	}
+
+	name, avatar := resolveFromIdentity(m, req)
+	payload := map[string]interface{}{
+		"type":    int(common.RichText),
+		"content": content,
+		"from": map[string]interface{}{
+			"kind":       extraKindValue,
+			"webhook_id": m.WebhookID,
+			"name":       name,
+			"avatar":     avatar,
+		},
+		// space_id 由服务端从 group 派生，不接受调用方覆盖（与 buildPayload 一致）。
+		"space_id": m.SpaceID,
+	}
+
+	// Validate（入站 write-strict）→ Finalize（注入后重算 plain + 完整 payload 复检
+	// 1MB）。push 路径无后续 enrich，两步紧邻即可。
+	if err := richtext.Validate(payload); err != nil {
+		return nil, err
+	}
+	if err := richtext.Finalize(payload); err != nil {
+		return nil, err
+	}
+	return payload, nil
+}
--- a/modules/incomingwebhook/richtext_push_test.go
+++ b/modules/incomingwebhook/richtext_push_test.go
@@ -64,6 +64,7 @@ func TestPush_RichText_RejectsEmptyBlocks(t *testing.T) {
 		"blocks":   []map[string]interface{}{},
 	})
 	assert.Equalf(t, http.StatusBadRequest, w.Code, "empty blocks must 400; body=%s", w.Body.String())
+	assert.Containsf(t, w.Body.String(), "blocks", "details.reason should be blocks; body=%s", w.Body.String())
 }
 
 // 图片块缺 width/height → 400。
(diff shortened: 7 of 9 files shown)


Jerry-Xin (approved):
The PR is in scope for `octo-server` and the implementation is ready to merge; I found no blocking correctness or security issues.

**💬 Non-blocking**

- 🟡 Warning: `modules/incomingwebhook/api.go:774-803` now rejects any non-empty `msg_type` other than `text` or `richtext`. Previously, an unknown JSON field named `msg_type` would have been ignored, so clients already sending values like `markdown` with valid `content` would start receiving 400s. If that strictness is intentional, call it out as a compatibility caveat.

- 🔵 Suggestion: `modules/incomingwebhook/richtext_push_test.go:64-94` checks status codes but not the promised `details.reason` mapping for `blocks` and `msg_type`. Adding envelope assertions would better lock the external API contract.

**✅ Highlights**

- Reuses `pkg/richtext.Validate` and `Finalize` instead of duplicating rich text validation.
- Preserves the existing trusted server fields: `from.kind`, `webhook_id`, and `space_id`.
- Keeps arbitrary caller fields out of the outgoing payload, matching the existing webhook security model.
- Targeted unit tests pass: `go test ./modules/incomingwebhook -run 'TestBuildRichTextPayload|TestResolveFromIdentity|TestIncomingWebhookNoLegacyResponseError'`.

Jerry-Xin (dismissed):
The PR is in scope for `octo-server` and the implementation is ready to merge; I found no blocking correctness or security issues.

**💬 Non-blocking**

- 🟡 Warning: `modules/incomingwebhook/api.go:774-803` now rejects any non-empty `msg_type` other than `text` or `richtext`. Previously, an unknown JSON field named `msg_type` would have been ignored, so clients already sending values like `markdown` with valid `content` would start receiving 400s. If that strictness is intentional, call it out as a compatibility caveat.

- 🔵 Suggestion: `modules/incomingwebhook/richtext_push_test.go:64-94` checks status codes but not the promised `details.reason` mapping for `blocks` and `msg_type`. Adding envelope assertions would better lock the external API contract.

**✅ Highlights**

- Reuses `pkg/richtext.Validate` and `Finalize` instead of duplicating rich text validation.
- Preserves the existing trusted server fields: `from.kind`, `webhook_id`, and `space_id`.
- Keeps arbitrary caller fields out of the outgoing payload, matching the existing webhook security model.
- Targeted unit tests pass: `go test ./modules/incomingwebhook -run 'TestBuildRichTextPayload|TestResolveFromIdentity|TestIncomingWebhookNoLegacyResponseError'`.

lml2468 (dismissed):
APPROVED (head `4b41ec39877a`).

Phase 1 of #297. Read the full diff against `pkg/richtext` (the authoritative server-side RichText handling), the existing `buildPayload` text path, and the model/api seam. Design is the right shape: opt-in via `msg_type`, single canonical validator (`pkg/richtext`), zero behaviour change on the default path.

## ✅ Independently verified

**Default path is byte-identical.**
`switch strings.ToLower(strings.TrimSpace(req.MsgType))` treats `""` and `"text"` identically and routes to the existing `buildPayload`. `TestPush_TextMsgTypeBackCompat` pins this. Existing callers omitting `msg_type` get exactly the pre-#297 payload, including the same `pushPayloadInvalid(c, "content")` 400 on empty content.

**`pkg/richtext.Validate` + `Finalize` are the right delegates.**
I traced `pkg/richtext/plain.go` — `Validate` marshals the **full** payload map (not the `content+plain` subset) before calling `common.ValidateRichTextPayload`, and `Finalize` re-marshals after writing the server-authoritative `plain` to re-check the 1MB cap on the actual outbound bytes. Both points are the explicit fixes from Jerry-Xin's PR #232 review (Critical#1 + Critical#2) so this PR inherits the hardening for free. The right move — don't re-implement the validator, delegate to the one that already passed that review.

**`from.kind=webhook` + `space_id` injection is consistent with the text path.**
`buildRichTextPayload` injects `from.kind=extraKindValue` (the same const the text path uses, verified in the diff context) and uses `m.SpaceID` from the looked-up webhook row — *not* anything in `req`. `TestBuildRichTextPayload_SpaceIDNotForgeable` pins that the response carries the webhook's bound space regardless of what the caller sent. Cross-Space forgery is closed exactly the same way as the text path.

**Field whitelist on each block type, with unknown blocks reaching Validate stripped.**
The `switch b.Type` only copies `text` for text blocks and `url/width/height` for image blocks. Unknown blocks reach `Validate` carrying *only* `type` — `TestBuildRichTextPayload_UnknownBlockDropsExtraFields` proves that a caller cannot smuggle other fields by labeling a block `image_v2` and dressing it up to look like an image. Validate sees a bare `{"type":"image_v2"}` and rejects with `ErrRichTextUnknownBlock`. Important: this also means *future* block types added in `pkg/richtext` can't be silently activated by an old server — the whitelist is closed.

**Block-count cap is enforced as a 400-class structural error, not 413.**
`errTooManyBlocks` is caught before `pushPayloadTooLarge` and maps to `pushPayloadInvalid(c, "blocks")`. The reasoning in the comment is right: 413 is for byte/volume; block count is structural invalid. The `envMaxBlocks` env knob with `defaultMaxBlocks=50` matches the project pattern (e.g. `DM_INCOMINGWEBHOOK_*` from #292/#293).

**Error mapping matrix is correct.**
| Cause | Branch | Status |
|---|---|---|
| 1MB exceeded (Validate or Finalize) | `errors.Is(err, common.ErrRichTextPayloadTooLarge)` | 413 |
| Empty content / empty text / non-http URL / missing W/H / unknown block | other `Validate` errors | 400 invalid `reason=blocks` |
| Block count > maxBlocks | `errTooManyBlocks` | 400 invalid `reason=blocks` |
| Unknown `msg_type` | `default` | 400 invalid `reason=msg_type` |
| Empty content on text path | unchanged | 400 invalid `reason=content` |

Symmetric with the text path's reason-code discipline. Caller debugging gets a precise field.

**No token / token_hash leakage in the outgoing payload.**
`TestBuildRichTextPayload_NoTokenLeak` pins that the marshalled payload doesn't contain `m.TokenHash`. Belt-and-braces given `buildRichTextPayload` only references `m.WebhookID` / `m.SpaceID` / `m.Name` / `m.Avatar`, but explicitly locked.

**Truncation/whitelist on `name`/`avatar` reuses `resolveFromIdentity` + the existing 64/255 byte caps from the text path.**
Verified the same helper is used at line where `name, avatar := resolveFromIdentity(m, req)` is called — same display-name length policy, same UTF-8-boundary-safe truncation as the text path. No new attack surface from the rich-text branch.

## ✅ Test coverage shape

**Unit (no infra) — 10 functions:** text-only / image / mixed-order / from-override / space_id-not-forgeable / no-token-leak / 6-case rejection table / too-many-blocks / unknown-block-strips-fields / resolveFromIdentity. Covers all the security invariants and the rejection matrix.

**E2E (CI) — 5 functions:** success / empty blocks / bad image / unknown msg_type / text BC. The 1MB 413 path isn't directly exercised here because `pkg/richtext/plain_test.go` already pins it at the source — testing it again at the handler layer would be redundant.

- **Order of msg_type → content branch.** On the text path, `Content==""` returns 400 `reason=content` before checking `req.MsgType`'s extras. Today this is correct, but a future hybrid mode (e.g. `msg_type=text` carrying both `content` and `blocks` for transitional clients) would need to revisit ordering. Comment-only suggestion — nothing to do now.

- **`webhookBlock` independent of `common.RichTextBlock`.** The PR rightly keeps the external contract decoupled from the internal type. The downside is two structs that need to stay synchronized when new block types are added. Worth a one-line comment on `webhookBlock` (or a `// keep aligned with pkg/richtext supported block types`) pointing future contributors at `pkg/richtext` so they update both. Cosmetic.

- **Phase 1 frontend coordination.** Web/mobile clients need to render `type=14` already (per the PR description "复用既有 RichText 渲染链路"). Worth confirming that the existing render path can also handle a webhook-originated RichText (i.e. that the `from.kind=webhook` discriminator doesn't suppress rich rendering). Out of scope for this PR but worth a quick smoke once Phase 2 client work lands.

- **`envMaxBlocks` follows the env-driven pattern from #292/#293, but doesn't have the read-side clamp those PRs added.** A negative `DM_INCOMINGWEBHOOK_MAX_BLOCKS` would fall through `n > 0` check correctly (returns default), so the parsing is safe. NaN/non-numeric also falls through `strconv.Atoi` error. No issue today; just noting it's a one-liner away from being symmetric with the v3.2 pattern if you want full parity.

## Checks

Build / Lint / Vet / i18n / osv all green on `4b41ec39877a`. `Test` / `CodeQL` pending. The echo `code-review` check is already APPROVED. `check-sprint` red is the project-board freshness, not code.

## Verdict

Phase 1 is well-scoped: the opt-in design preserves the historical contract exactly, delegates validation to the already-hardened `pkg/richtext`, and gets the cross-Space / token-leak / unknown-block defenses pinned by tests. The whitelist-per-block-type pattern is correct paranoia (unknown blocks can't smuggle fields). The error mapping discipline matches the text path's reason-code convention.

LGTM, merge once Test/CodeQL settle.

yujiawei (dismissed):
# Code Review — PR #298 (octo-server)

**Verdict: APPROVED.** This is a well-scoped, backward-compatible Phase 1 that adds an optional `msg_type="richtext"` path to the token-authenticated incoming-webhook push endpoint. The security baseline of the legacy text path is preserved end-to-end, the build/vet/pure-unit tests pass, and validation is correctly delegated to the shared authoritative gate. No blocking (P0/P1) issues found. A handful of non-blocking consistency/documentation items are listed below for a follow-up.

## 1. Verification summary

| Check | Result | Evidence |
|---|---|---|
| Cross-Space forgery prevented | ✅ | `space_id` is server-injected from `m.SpaceID` in both paths; `webhookBlock` has no `space_id` field and the per-block whitelist copies only known fields (`richtext.go:71-93`, `api.go:893`). Covered by `TestBuildRichTextPayload_SpaceIDNotForgeable`. |
| Caller field passthrough dropped | ✅ | Only `type`/`text`/`url`/`width`/`height` are copied per block; unknown blocks carry through `type` only and are then rejected by `Validate` (`richtext.go:62-79`). `req.Extra` is never read on the richtext path. |
| Image URL scheme allowlist (no `data:`/`javascript:`/`file:`) | ✅ | Enforced by `common.ValidateRichTextBlocks` → `isAllowedImageURLScheme` (octo-lib `common/richtext.go`). Covered by reject-path tests. |
| `plain` is server-authoritative | ✅ | `richtext.Finalize` recomputes `plain` via `FillPlain`; caller-supplied `plain` cannot be injected (no such field on `webhookBlock`). |
| RichText gate actually fires | ✅ | `buildRichTextPayload` sets `"type": int(common.RichText)`; confirmed `richtext.IsRichTextPayload(payload)` returns true on the built map, so `Validate`/`Finalize` are not silently no-ops. |
| Error mapping | ✅ | `ErrRichTextPayloadTooLarge` → 413; all other structural errors → 400 `reason=blocks`; unknown `msg_type` → 400 `reason=msg_type` (`api.go:787-803`). |
| Size limits | ✅ | 8KB pre-parse body cap (`api.go:753-762`), `maxBlocks=50` (`richtext.go`), 1MB richtext byte cap in `Finalize`. |
| Backward compatibility | ✅ | Empty/`"text"` `msg_type` routes to the unchanged text path; `buildPayload` refactor into `resolveFromIdentity` preserves truncation + fallback, and `extraKindValue == "webhook"` matches the prior literal. |
| Build / vet / pure unit tests | ✅ | `go build`, `go vet`, and `TestBuildRichTextPayload*` / `TestResolveFromIdentity` pass locally. (DB/e2e tests left to CI.) |

The design choice to reuse `RichText(=14)` and centralize validation in `pkg/richtext` (the same `Validate`/`Finalize` contract used by the user/robot senders) is sound and keeps the strict cross-sender contract consistent.

### P2 — `richtext` text length is not bounded by `maxContentRunes`, unlike the text path

`api.go:782-785` enforces a 4000-rune ceiling on `req.Content` for the text path (→ 413). The richtext path applies no equivalent per-message rune limit — it only checks `maxBlocks` and defers to the 1MB byte cap. Today this is masked by the default 8KB body cap, so it is **not exploitable under default config**. However, `README.md` (and the comment at `api.go:780-781`) explicitly anticipate operators raising `DM_INCOMINGWEBHOOK_MAX_BYTES`, at which point a single large text block would silently bypass the per-message rendering ceiling that the text path advertises.

Suggestion: either sum `utf8.RuneCountInString` over text blocks (or over the generated `plain`) in `buildRichTextPayload` and reject above `maxContentRunes()` with the same 413 mapping, or explicitly document that richtext is intentionally exempt. The concern is the *undocumented* asymmetry that re-opens when the body cap is tuned up — not a present-day exploit.

### nit — Stale `reason` enum in two doc-comments

The PR adds `reason=blocks` (`api.go:797`) and `reason=msg_type` (`api.go:802`), but two doc-comments still list only `{body, json, content}`:
- `modules/incomingwebhook/api_i18n.go:38-40` (above `pushPayloadInvalid`)
- `pkg/errcode/incomingwebhook.go:42-44` (`ErrIncomingWebhookPushPayloadInvalid`)

No functional impact — `SafeDetailKeys` whitelists the key *name* `"reason"`, not its values, so all five reasons surface correctly, and `README.md:80` already lists all five. Just update both comments together to match.

### nit — Block-type matching is case-sensitive while `msg_type` is normalized

`api.go:774` normalizes `msg_type` via `strings.ToLower(strings.TrimSpace(...))`, but block `type` is matched exactly (`richtext.go:61`, and likewise in octo-lib). So `{"type":"Text"}` or `{"type":" image "}` is rejected as an unknown block. This fails closed (safe), and strict block typing matches every other RichText sender, so keeping it strict is defensible — but the lenient-envelope / strict-block asymmetry is not called out in the README. Consider one line in the block-type table noting `type` must be lowercase exact.

### nit — Redundant marshals on the richtext path (perf, not a concern today)

The richtext payload is serialized four times (`Validate` ×1, `Finalize` ×2, send ×1) vs once on the text path. Bounded by the 8KB body cap so negligible today; the three inner marshals are inherited from the shared octo-lib contract by design (size checks on the full pre/post-enrich payload). Flagging only as a note if the body cap is ever raised toward 1MB.

## 3. Recommendations

The P2 rune-limit asymmetry is the only item worth addressing before this becomes a foot-gun — pick either enforcement or an explicit doc note. The nits are low-effort cleanups that can ride along or be deferred. None block merge.

## 4. Additional notes for the human reviewer (security-sensitive PR)

- This endpoint is unauthenticated except for the URL token; the new path correctly inherits the existing token/rate-limit/group-status gates ahead of payload construction (`api.go:700-760`), so the richtext branch adds no new pre-auth surface.
- Image blocks are **URL references only** (no server-side fetch in this code path), so there is no new SSRF surface introduced here — images are rendered client-side. Worth a human confirming no downstream consumer eagerly fetches these URLs server-side, but nothing in this PR does.
- `token_hash` non-leakage is asserted by `TestBuildRichTextPayload_NoTokenLeak`.

Jerry-Xin (dismissed):
The PR is in scope for `octo-server`: it extends the existing incoming webhook push endpoint and documents that endpoint’s request contract.

💬 Non-blocking
- 🔵 Suggestion: [modules/incomingwebhook/api_i18n_test.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-qkmj5wet/repo/modules/incomingwebhook/api_i18n_test.go:22) does not include the new `richtext.go` in the legacy-response guard. It is currently only a pure builder/helper file, so this is not blocking, but adding it would make future drift harder if response logic is ever moved there.
- 🔵 Suggestion: [modules/incomingwebhook/richtext_push_test.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-qkmj5wet/repo/modules/incomingwebhook/richtext_push_test.go:47) uses negative assertions for success because downstream sending may return `502` in test infra. This matches nearby test style, but a future mockable sender seam would make richtext success assertions stronger.

✅ Highlights
- The relevance gate passes: this PR modifies first-party incoming webhook functionality.
- [modules/incomingwebhook/api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-qkmj5wet/repo/modules/incomingwebhook/api.go:774) keeps omitted/`text` behavior isolated from the new `richtext` path.
- [modules/incomingwebhook/richtext.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-qkmj5wet/repo/modules/incomingwebhook/richtext.go:60) correctly whitelist-translates block fields instead of passing caller-controlled payload fields through.
- [modules/incomingwebhook/richtext.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-qkmj5wet/repo/modules/incomingwebhook/richtext.go:81) preserves the existing server-injected `from` and `space_id` security model.
- Error mapping is consistent: structural richtext errors become `400 reason=blocks`, oversize remains `413`, and unknown `msg_type` is explicit.

Verification: `go test ./modules/incomingwebhook -run 'TestBuildRichTextPayload|TestResolveFromIdentity|TestIncomingWebhookNoLegacyResponseError'` passed. Route-level richtext tests could not run locally because MySQL on `127.0.0.1:3306` was unavailable; those are expected to require the project test infra.

lml2468 (approved):
APPROVED (head `e9eb66c84f87`).

Pure docs + test follow-up. 0 lines of production logic touched — every diff is either a comment, README note, or test assertion. The R1 verified properties (default text path byte-identical, `pkg/richtext.Validate`/`Finalize` delegation, cross-Space forgery closed, per-block field whitelist, no-token-leak) are unchanged.

## ✅ Each follow-up verified

**1. `webhookBlock` "keep aligned with pkg/richtext" note (R1 my nit #2).**
The comment now explicitly tells the next contributor that buildRichTextPayload is a whitelist translator and adding a block type requires syncing both `webhookBlock` AND `pkg/richtext`'s supported types. Exactly the cross-file forward-note I suggested. Cosmetic but prevents the next "added a block type, forgot the other half" round-trip.

**2. `details.reason` assertions on the three rejection cases (Jerry-Xin R1).**
`TestPush_RichText_RejectsEmptyBlocks` and `RejectsBadImage` now assert the body contains `"blocks"`; `RejectsUnknownMsgType` asserts `"msg_type"`. Without these, an accidental swap of reason codes in `pushPayloadInvalid` would only fail at integration time. Now the external error contract (status + reason field) is pinned at unit level. Right shape.

**3. `maxContentRunes` asymmetry documented (yujiawei R1).**
yujiawei's R1 nit "richtext 路径未限制文本长度可能被绕过" is **not actually a bug** — richtext is bounded by block structure + the 1MB serialized cap from `pkg/richtext`, and the 8KB body cap upstream means richtext can't even approach the 1MB cap in practice. The comment now spells out *why* the asymmetry is deliberate, including the forward-looking "if body cap is raised, 1MB still backstops". That's the right shape: document the design choice so the next reviewer doesn't re-derive it and ask the same question.

**4. README + `pkg/errcode/incomingwebhook.go` + `api_i18n.go` reason-code enums.**
The doc-comments that previously listed only `{body,json,content}` now include the richtext-path reasons `{blocks, msg_type}`. Same fix in two files (api_i18n.go + errcode) because both had the same stale list. No regression risk; just keeps the discovery surface honest for whoever greps `reason=` next.

## On the four non-blocking nits from the prior round

| Reviewer point | R2 status |
|---|---|
| 🟡 (Jerry-Xin) test未覆盖 details.reason | ✅ asserted on all 3 rejection cases |
| 🟡 (Jerry-Xin) msg_type 严格校验文档化 | ✅ called out in README ("msg_type 现在严格校验，未知如 'markdown' 现在 400 而非被忽略") |
| 🟡 (yujiawei) richtext 文本长度未限制 | ✅ documented as intentional design choice + escape hatch |
| 🟡 (yujiawei) reason 枚举文档过时 | ✅ enums updated in both api_i18n.go + errcode |
| 🟡 (me R1) webhookBlock sync with pkg/richtext | ✅ comment added |
| 🟡 (me R1) Phase 2 frontend rendering smoke | (out of scope, separate Phase 2 work) |
| 🟡 (me R1) envMaxBlocks read-side clamp | not addressed (cosmetic, safe today, fine to defer) |

5 of 7 nits closed in one tight pass, the other 2 are deferred for the right reasons (Phase 2 cross-team work, cosmetic clamp parity). Good scope discipline — author didn't expand into Phase 2 stuff.

## What I did NOT verify

- Behavioral diff (none expected — every changed file is comment/test/doc only).
- CI green on this SHA — Build/Lint/Vet should ride the R1 results since production is unchanged; the test additions are 3 lines of `assert.Containsf` on existing cases, no new test functions.

## Verdict

Honest docs + intent-locking test assertions on top of the already-APPROVED R1 logic. The "document why the asymmetry exists" treatment of yujiawei's nit is the right shape (same pattern as #295 R3 system-bot doc-tightening). LGTM, merge once CI confirms green on the new SHA.

yujiawei (approved):
# Code Review — PR #298 (octo-server)

**Verdict: APPROVED** — no P0/P1 blocking issues. One non-blocking behavioral note (P2) and a few maintainability nits, all documented or out of scope for Phase 1.

This is an unauthenticated, token-in-URL push endpoint, so I reviewed it through injection/forgery, validation-bypass/DoS, and backward-compat lenses, and traced every claim into the authoritative cross-file validator (`pkg/richtext` → `octo-lib common/richtext.go`). Build, `go vet`, and the no-infra unit suite (`richtext_test.go`, `TestResolveFromIdentity`) are green locally.

## 1. Verification of design invariants

- ✅ **Backward compat (text path unchanged).** `extraKindValue == "webhook"` (`modules/incomingwebhook/display.go:33`), so swapping the inline literal for the const is value-identical. The text branch (`api.go:775-786`) reproduces the prior straight-line sequence verbatim (empty-content → 400 `content`; rune-count over cap → 413; `buildPayload`), and `resolveFromIdentity` (`api.go:895-905`) is a faithful extraction of the previously-inlined name/avatar fallback + `truncateUTF8`. Callers that omit `msg_type` or send `"text"` see zero change.
- ✅ **Server-injected identity / no cross-Space forgery.** `buildRichTextPayload` (`richtext.go:82-93`) builds the payload from server-controlled keys only — `type`, whitelisted `content`, `from.kind/webhook_id/name/avatar`, and `space_id = m.SpaceID`. No caller field (`space_id`, `from.*`, `extra`, `plain`, `mention`, `reply`) is read in. `req.Extra` has no non-test reader in the module. Full parity with the text path.
- ✅ **Whitelist translation.** Per-block copy is strict: text → `{type,text}`, image → `{type,url,width,height}`, unknown → `{type}` only (`richtext.go:60-79`). No arbitrary per-block passthrough; `webhookBlock` (`model.go:70-76`) declares exactly the 5 wire fields, so JSON unknown-field drop handles the rest.
- ✅ **No validation bypass via type confusion.** `payload["type"]` is set to `int(common.RichText)` in-process (`richtext.go:83`); `IsRichTextPayload` matches the `int` case (`pkg/richtext/plain.go:42-43`), so `Validate` does **not** short-circuit to its `no-op` early return — it always runs `common.ValidateRichTextPayload` over the full marshaled map. The "Validate silently skips everything" path is unreachable here.
- ✅ **`plain` is server-authoritative.** No `plain` field exists on `webhookBlock` or the payload map, and `Finalize` unconditionally overwrites `payload["plain"]` from `content` (`plain.go:108` → `FillPlain`). Caller-supplied `plain` is structurally impossible.
- ✅ **Authoritative rejection of bad blocks (cross-file).** `ValidateRichTextBlocks` iterates *all* blocks with no early-accept and rejects the first unknown type (`octo-lib common/richtext.go:188-209`); the image scheme allowlist rejects `data:`/`javascript:`/`file:` and requires a non-empty host (`:242-253`). A partially-valid array (valid text block + unknown/miscased block) is rejected wholesale — covered by `TestBuildRichTextPayload_UnknownBlockDropsExtraFields`.
- ✅ **Error mapping.** `>1MB` → 413 via `errors.Is(err, common.ErrRichTextPayloadTooLarge)`; all other structural errors → 400 `reason=blocks`; unknown `msg_type` → 400 `reason=msg_type` (`api.go:791-806`). `reason` is in `SafeDetailKeys` (`pkg/errcode/incomingwebhook.go:51`), so the new values surface safely. The 8KB body cap is enforced before JSON parse (`api.go:752-762`); the 1MB richtext cap is a correct secondary guard that only becomes reachable if an operator raises `DM_INCOMINGWEBHOOK_MAX_BYTES`.
- ✅ **Env-parse hardening.** `maxBlocks`/`maxBytes`/`maxContentRunes` all use `strconv.Atoi(...) == nil && n > 0`, so negative, zero, and overflow values fall back to defaults — no path disables or balloons the cap.

## 2. Issues

**No P0 / P1.**

### P2 (non-blocking) — strict `msg_type` is a behavior change for non-conforming clients

Pre-PR, `pushPayloadReq` had no `MsgType` field, so any stray `msg_type` JSON key was silently dropped and the request proceeded as plain text as long as `content` was valid. Now `api.go:804-806` returns 400 `reason=msg_type` for any non-empty value outside `{text, richtext}` (case-insensitive), even with valid `content`. A legacy client sending e.g. `{"msg_type":"markdown","content":"hi"}` succeeded before and now fails.

This is intentional and explicitly documented in `README.md` (the 兼容性提醒 / compatibility note). Conforming clients (omitted `msg_type` or `"text"`) are unaffected, so invariant #1 holds for them. **Recommendation:** before rollout, confirm no production caller relies on sending a non-`{text,richtext}` `msg_type` and depending on plain-text delivery. Not a merge blocker.

## 3. Suggestions (nits, optional)

- **Case-handling asymmetry**: `msg_type` is lower+trimmed (`api.go:774`) but block `type` is compared case-sensitively against `"text"/"image"` (`richtext.go:61`), so `{"type":"Text"}` is rejected as an unknown block. This is documented in the README and harmless (rejection, not silent acceptance), but if caller ergonomics matter, consider normalizing block `type` the same way `msg_type` is.
- **Marshal cost**: the `Validate` + `Finalize` sequence marshals the payload three times and unmarshals once per richtext request. Negligible under the 8KB body / 50-block caps, and the full-map marshal is the intentional fix for counting server-injected/unknown top-level fields against the 1MB cap — leaving a note only, no change requested.

## 4. Additional notes (security reviewer hand-off)

- **Image URL = client-fetched reference.** The server stores image `url` as a reference and does not fetch it; SSRF/url-fetch handling is a client concern and is the documented Phase-1 contract. Worth a human confirming the rendering clients treat these URLs safely (no SSRF on fetch, no script-URL execution), since the server is correctly not the rendering trust boundary.
- **Text block content is passed through verbatim** (no server-side escaping), identical to the existing text path — the renderer remains the injection trust boundary on both paths. No new server-side surface introduced.

Tests are well-targeted: pure-translation unit tests cover order preservation, from-override, non-forgeable `space_id`, no-token-leak, every rejection path, and the block-count cap; e2e tests cover success / empty blocks / bad image / unknown `msg_type` / text back-compat. Solid Phase-1 work.

lml2468 (dismissed):
APPROVED (head `65de4fa746ae`).

Rebase-only update. The 3 PR commits keep identical SHAs of content for every Phase 1 file vs my R2 head `e9eb66c84f87`:

| file | blob SHA |
|---|---|
| `modules/incomingwebhook/README.md` | `922ac100608a` |
| `modules/incomingwebhook/api.go` | `e43c0acaf527` |
| `modules/incomingwebhook/api_i18n.go` | `9ddd362aacef` |
| `modules/incomingwebhook/model.go` | `aa9bdd83fe29` |
| `modules/incomingwebhook/richtext.go` | `b7a31e41d719` |
| `modules/incomingwebhook/richtext_push_test.go` | `aaf662db6c23` |
| `modules/incomingwebhook/richtext_test.go` | `ad7006bb56fe` |
| `pkg/errcode/incomingwebhook.go` | `c631d68d82ba` |

All 8 blobs verified equal via `gh api repos/.../contents/<path>?ref=<sha>`. The `compare` API shows the PR branch is now `ahead 6 / behind 3` because main moved (CodeQL schedule change #313, sidebar thread-status #312, usersecret alias module), but none of those base commits touch `modules/incomingwebhook/**` or `pkg/errcode/incomingwebhook.go` — diff against the new base is byte-identical to the old.

## What this means

The R2 verifications all carry forward verbatim:

- **Default text path byte-identical** — `pushPayloadReq` with omitted/`text` `msg_type` routes to unchanged `buildPayload`; `TestPush_TextMsgTypeBackCompat` pins it.
- **`pkg/richtext.Validate`/`Finalize` delegation** — single canonical validator, inherits #232 R1 hardening (full-payload marshal for size + post-enrich re-check) for free.
- **Cross-Space forgery closed** — `space_id = m.SpaceID` server-injected; `webhookBlock` has no `space_id` field; `TestBuildRichTextPayload_SpaceIDNotForgeable` pins it.
- **Per-block field whitelist** — text/image blocks copy known fields only; unknown blocks reach Validate carrying only `type` (rejected with `ErrRichTextUnknownBlock`); `TestBuildRichTextPayload_UnknownBlockDropsExtraFields` pins it.
- **Error mapping matrix correct** — `>1MB` → 413; structural → 400 `reason=blocks`; unknown `msg_type` → 400 `reason=msg_type`; block-count cap → 400 not 413.
- **No token leak** — `TestBuildRichTextPayload_NoTokenLeak` pins it.
- **R2 doc/test follow-ups intact** — `webhookBlock` sync-with-pkg/richtext comment, `details.reason` assertions on all 3 rejection cases, `maxContentRunes` asymmetry rationale, `reason=` enum updates in both `api_i18n.go` + `errcode/incomingwebhook.go`.

## Checks

The rebase kicked a fresh CI run on `65de4fa`; ride those to green before merge (Build / Test / Vet / Lint / CodeQL / i18n still IN_PROGRESS at review time). No behavior risk from the rebase itself.

## Verdict

Pure base move, zero Phase 1 content drift. R2 conclusions carry forward. LGTM, merge once CI confirms green on the new SHA.

yujiawei (dismissed):
# Code Review — PR #298 (octo-server)

**Scope:** rich text / image-text blocks on the incoming-webhook push endpoint (Phase 1). Reviewed against head `65de4fa7`. This PR was flagged security-sensitive, so the security baseline was probed adversarially. Locally verified: `go build`, `go vet` clean, and all pure unit tests in `modules/incomingwebhook` pass.

## Verdict: Approve

The design is sound and the security baseline is preserved. I found **no P0/P1 issues**. The new `richtext` path correctly funnels all block validation through the single authoritative gate (`pkg/richtext.Validate` + `Finalize`), and the server-injected identity/space fields are non-forgeable. The items below are all non-blocking (P2/nit) — worth addressing in this PR or a fast follow-up, but none should hold the merge.

## Security baseline — verified holding

- **No cross-Space / identity forgery.** `buildRichTextPayload` (`modules/incomingwebhook/richtext.go:81-93`) constructs the outbound map entirely server-side with exactly four keys (`type`, `content`, `from`, `space_id`). `from.kind` is the hardcoded `extraKindValue` = `"webhook"` (`display.go:33`) and `space_id` is `m.SpaceID` from the DB model. No caller field is merged into the top-level map.
- **`req.Extra` and unknown fields dropped.** The richtext path never references `req.Extra`. `webhookBlock` is a strict whitelist (`type/text/url/width/height`); the unknown-block `default` branch copies only `{type}`, which is then rejected by `ValidateRichTextBlocks` with `ErrRichTextUnknownBlock` — no field smuggling.
- **Client-supplied `plain` cannot be forged.** `webhookBlock` has no `plain` field, and `Finalize` (`pkg/richtext/plain.go:108`) unconditionally overwrites `payload["plain"]` with the server recompute, then rechecks 1MB on the full map.
- **Image URL scheme allowlist works.** `http`/`https` only; `data:`/`javascript:`/`file:` rejected (octo-lib `common/richtext.go:242-253`). See nit (1) below on the SSRF boundary.
- **Bounded size / no DoS amplification.** 8KB body cap is read pre-parse (`api.go:753-762`); `maxBlocks` (default 50) is checked before the translation loop; `Validate`/`Finalize` are single-pass. `token_hash` is never copied into the payload or logs.
- **`msg_type` normalization** uses `strings.ToLower(strings.TrimSpace(...))` (`api.go:774`), so casing/whitespace cannot bypass the switch into an unvalidated path.

## Non-blocking findings (P2)

1. **Behavioral break: non-empty unknown `msg_type` now returns 400** (`api.go` default switch branch). Previously an unknown JSON field was silently ignored and the request fell through to the text path; now a caller sending e.g. `{"msg_type":"markdown","content":"..."}` with otherwise-valid content gets `400 reason=msg_type`. This is intentional and is documented in `README.md`, but the PR description's "zero behavior change" is accurate only for callers that omit `msg_type` or send `text`/empty. Suggest carrying the README's compatibility caveat into the release notes.

2. **Guard test not updated for the new file.** `TestIncomingWebhookNoLegacyResponseError` (`api_i18n_test.go:22`) enumerates source files (`api.go, api_i18n.go, ratelimit.go, localfloor.go, cache.go`) but the new `richtext.go` is absent, and the test's own doc comment says "Add any new handler/limiter file to the list below." Harmless today — `richtext.go` emits no HTTP responses (it returns Go errors that `api.go` renders via the i18n envelope) — but a future raw response added inside `buildRichTextPayload` would silently escape the D23 guard. Suggest adding `"richtext.go"` to the slice.

3. **Test coverage gaps on the 413 paths.**
   - The richtext `>1MB → 413` branch (`api.go`, `errors.Is(err, common.ErrRichTextPayloadTooLarge)`) is dead under the default 8KB body cap and has no test. It becomes live only if an operator raises `DM_INCOMINGWEBHOOK_MAX_BYTES` above 1MB — the documented secondary-backstop scenario. A unit test feeding a >1MB payload (asserting `errors.Is(..., ErrRichTextPayloadTooLarge)`) would keep the backstop honest.
   - The text-path `maxContentRunes` (4000-rune) `413` limit (`api.go`) has no test — `TestPush_RejectsTooLargeBody` only trips the byte cap. This is a pre-existing gap but in-scope since the PR restructured this exact block into the `msg_type` switch.

4. **`msg_type` is case/space-normalized but block `type` is matched case-sensitively** — an intentional asymmetry (`"Image"` would be rejected) that is documented in the README but not at the code site. A one-line comment near the block-type switch in `richtext.go` would prevent future confusion.

## Nits (optional)

- **SSRF boundary, document the assumption.** `isAllowedImageURLScheme` is a scheme/XSS guard, not an SSRF guard — `http://169.254.169.254/...` or internal-host URLs pass. Safe in this PR because the server never dereferences the image URL (it is stored and rendered client-side). If any future server-side feature (image proxy, unfurl, thumbnail, LLM image fetch) starts dereferencing these caller-controlled URLs, the resolved host must be re-validated against an SSRF blocklist. Worth a defensive comment now.
- **Doc accuracy:** `README.md` says the body is "截断" (truncated) before parsing; the code rejects the whole request with 413 (`io.LimitReader(..., limit+1)` then reject) — it never truncates-and-delivers. Reword to avoid implying partial delivery.
- **Comment drift:** `pushPayloadTooLarge` (`api_i18n.go`) and `ErrIncomingWebhookPushPayloadTooLarge` (`pkg/errcode/incomingwebhook.go`) doc comments mention only the byte cap, but the same 413 is now also returned for the rune-count cap and the richtext 1MB backstop. Enumerate all three triggers.
- The unknown-block `default` branch appends a throwaway `{type}` block purely so `Validate` rejects it — correct and safe (preserves "Validate is the single block-legality authority"), just a minor dead allocation.

## Items a human may want to sanity-check (security-sensitive PR)

- Confirm there is genuinely no current server-side dereference of webhook-supplied image URLs anywhere downstream (search was clean in this module, but the SSRF safety rests on that holding platform-wide).
- Confirm the stricter `msg_type` rejection is acceptable for all existing production webhook integrations.

lml2468 (approved):
APPROVED (head `af3c685ce4c3`).

Real content change this time (not a rebase). CI-fix commit on top of R3:

- The R3 e2e `assert.Containsf(body, "blocks"/"msg_type", ...)` assertions failed CI in `TestPush_RejectsUnknownMsgType` (and the two RichText reject tests). Status code 400 was correct; the body just didn't carry `error.details`.
- Root cause: `testutil.NewTestServer` does **not** wire the i18n `ErrorRenderer`, so e2e error bodies render in legacy `{msg, status}` shape — `error.details` only exists in the i18n envelope. In production the renderer IS mounted (via `main.go`), so the contract holds; the assertion was probing harness behaviour rather than the production contract.

## What changed

**7 production files unchanged** — blob SHAs verified equal vs R3 head `65de4fa746ae`:

| file | unchanged? |
|---|---|
| `modules/incomingwebhook/README.md` | ✅ |
| `modules/incomingwebhook/api.go` | ✅ |
| `modules/incomingwebhook/api_i18n.go` | ✅ |
| `modules/incomingwebhook/model.go` | ✅ |
| `modules/incomingwebhook/richtext.go` | ✅ |
| `modules/incomingwebhook/richtext_test.go` | ✅ |
| `pkg/errcode/incomingwebhook.go` | ✅ |

**2 test files touched** — `modules/incomingwebhook/richtext_push_test.go` (-3 lines: drop the 3 `Contains(reason)` assertions, keep status-code checks + harness-limitation comment) and `modules/incomingwebhook/api_i18n_test.go` (+24 lines: new `TestPushPayloadInvalidSurfacesReason` using the existing `iwhHelperHarness` which mounts `i18n.NewErrorRenderer(...)` — same as production `main.go`).

## Why the new test is the right shape

- `iwhHelperHarness` wires the **actual** `ErrorRenderer` (the one production uses). The new test asserts `details.reason` for **4** reasons (`blocks`, `msg_type`, `content`, `json`) — strictly more coverage than the R3 e2e assertions (which were 2× `blocks` + 1× `msg_type`).
- Locks the **external contract** at the renderer layer, where `error.details` is actually emitted, instead of at the e2e layer where the test harness strips it. This is the correct trust boundary — production callers receive the i18n envelope, not the legacy one.
- No DB/Redis/WuKongIM dependency — runs in the unit pass, faster + more deterministic than e2e for a renderer-shape contract.
- The comment in `richtext_push_test.go` documents why e2e is intentionally **only** asserting status codes ("see api_i18n_test.go for the reason-detail contract"), so future contributors don't re-add and re-break.

## Production behavior unchanged

Every R3 invariant carries forward:

- Default text path byte-identical (`buildPayload`, `TestPush_TextMsgTypeBackCompat`).
- `pkg/richtext.Validate`/`Finalize` delegation, cross-Space forgery closed, per-block whitelist, no-token-leak — all asserted by the unchanged unit tests in `richtext_test.go`.
- Error mapping matrix (1MB→413; structural→400 reason=blocks; unknown msg_type→400 reason=msg_type; block-cap→400) unchanged in `api.go` + `richtext.go`.
- R3 doc/test follow-ups still in place (webhookBlock sync comment, maxContentRunes asymmetry rationale, reason= enum updates).

## Verdict

Test-only fix targeting a real e2e harness limitation that surfaced under CI; production code, security baseline, and error contract are unchanged. The new harness-level test is strictly more rigorous than the dropped e2e assertions. LGTM, merge once CI confirms green on `af3c685`.

lml2468 (approved):
APPROVED (head `6ead00f32437`).

Rebase-only update again. The 4 PR commits got new SHAs (`21ea2f74 / 225c7885 / 76957251 / 6ead00f3`) but identical commit messages, and every Phase 1 file's blob SHA is byte-identical to my R4 head `af3c685ce4c3`:

- `modules/incomingwebhook/README.md` ✅
- `modules/incomingwebhook/api.go` ✅
- `modules/incomingwebhook/api_i18n.go` ✅
- `modules/incomingwebhook/api_i18n_test.go` ✅
- `modules/incomingwebhook/model.go` ✅
- `modules/incomingwebhook/richtext.go` ✅
- `modules/incomingwebhook/richtext_push_test.go` ✅
- `modules/incomingwebhook/richtext_test.go` ✅
- `pkg/errcode/incomingwebhook.go` ✅

Zero content drift vs R4. All R1–R4 verifications carry forward verbatim (default text path byte-identical, `pkg/richtext.Validate`/`Finalize` delegation, cross-Space forgery closed, per-block whitelist, error mapping matrix, no token leak, R3 doc/test follow-ups, R4 CI-fix + renderer-level reason-detail contract test).

LGTM, merge once CI confirms green on the new SHA.

Jerry-Xin (approved):
Summary: The PR is relevant to octo-server, cleanly extends the incoming webhook push path, and I found no blocking correctness or security issues.

💬 Non-blocking

- 🟡 Warning [modules/incomingwebhook/api.go:774](/Users/mlamp/.openclaw/tmp/pr-review-ykz6qox9/repo/modules/incomingwebhook/api.go:774): Strict `msg_type` validation is a real compatibility edge. Existing callers that previously sent an ignored `msg_type` like `"markdown"` with valid `content` will now get 400. This is documented in [README.md](/Users/mlamp/.openclaw/tmp/pr-review-ykz6qox9/repo/modules/incomingwebhook/README.md:18), so I am not treating it as blocking, but it should be called out in release notes.

- 🔵 Suggestion [modules/incomingwebhook/richtext_push_test.go:51](/Users/mlamp/.openclaw/tmp/pr-review-ykz6qox9/repo/modules/incomingwebhook/richtext_push_test.go:51): The success-path e2e test excludes only 400/413/401. Consider asserting the status is one of the expected allowed outcomes, e.g. 200 or 502, so regressions returning 404/429/500 do not accidentally pass.

- 🔵 Suggestion [modules/incomingwebhook/README.md:66](/Users/mlamp/.openclaw/tmp/pr-review-ykz6qox9/repo/modules/incomingwebhook/README.md:66): The docs say the body is “truncated” before parsing, but the implementation reads `limit+1` and rejects with 413. Rewording this to “capped/rejected before parsing” would match behavior more precisely.

✅ Highlights

- The rich text path preserves the important security properties from the text path: caller `extra` is not forwarded, `space_id` is server-derived, and `from.kind=webhook` is injected.
- `buildRichTextPayload` reuses the shared rich text validation/finalization path, including authoritative `plain` generation and post-enrichment size checks.
- Error mapping is consistent with the documented contract: rich text structural failures return 400 with `reason=blocks`, and oversize rich text maps to 413.
- I ran `go test ./modules/incomingwebhook -run 'Test(BuildRichTextPayload|ResolveFromIdentity|PushPayloadInvalidSurfacesReason|IncomingWebhookRespondHelpers|IncomingWebhookNoLegacyResponseError|BuildPayload|Truncate|GenerateToken|HashToken)'` and `go test ./pkg/richtext`; both passed.

yujiawei (approved):
# Code Review — PR #298 (octo-server)

**Verdict: APPROVED.** Phase 1 of the rich-text / image-text incoming-webhook push is well-scoped, the security baseline of the existing text path is preserved field-for-field, and the new code is backed by thorough unit + e2e tests. No blocking (P0/P1) issues found. Two non-blocking P2 notes and a few nits below.

## 1. Verification summary

- ✅ **Build / vet / gofmt clean** for all 9 changed files. `go build ./modules/incomingwebhook/... ./pkg/...` and `go vet` pass; `gofmt -l` reports nothing on the PR files. (One unrelated pre-existing `cache.go` gofmt nit is outside this PR's scope.)
- ✅ **Pure unit tests pass locally** — `pkg/richtext` and `modules/incomingwebhook` (`TestBuildRichTextPayload_*`, `TestResolveFromIdentity`, `TestPushPayloadInvalidSurfacesReason`) all green. DB/Redis/WuKongIM e2e cases run in CI.
- ✅ **Identity / Space cannot be forged** (`richtext.go:81-93`). `type`, `from.kind` (`extraKindValue="webhook"`), `from.webhook_id`, and `space_id` are all derived from the server-side `incomingWebhookModel`, loaded after the constant-time token compare. `pushPayloadReq` exposes no JSON field for any of these, so a forged value cannot even deserialize. Byte-for-byte parity with the text path `buildPayload`. Pinned by `TestBuildRichTextPayload_SpaceIDNotForgeable`.
- ✅ **`req.Extra` / arbitrary fields are dropped** in the rich-text path, matching the text path. The per-block whitelist (`richtext.go:60-79`) copies only `{type,text}` for text and `{type,url,width,height}` for image; the unknown-block branch copies only `{type}` and is then rejected by `Validate`. `common.RichTextBlock.Size`/`Name` are deliberately not surfaced. Pinned by `TestBuildRichTextPayload_UnknownBlockDropsExtraFields`.
- ✅ **Anti-enumeration invariant holds.** The new `msg_type` switch (`api.go:774`) and every 400/413 it can emit (`reason=content|blocks|msg_type`, `pushPayloadTooLarge`) are reachable only *after* all auth gates: unknown/soft-deleted → `failAuth`, disabled → 401, constant-time token compare (`api.go:719-724`), and the group-active gate. An unauthenticated scanner is short-circuited with the uniform 401 long before any payload-shape feedback. The differentiated `reason` codes are only visible to an already-authenticated token holder — by design. The payload-error helpers do **not** consume the per-IP auth-failure budget, which is correct (a legit sender's malformed retry must not burn the shifted budget).
- ✅ **`plain` is authoritatively server-generated.** No `plain` field exists on the external contract; `Finalize` overwrites `payload["plain"]` from `content` via `BuildRichTextPlain`. Client cannot inject or forge it. The legacy string-content `UnmarshalJSON` branch in octo-lib is unreachable from this push path.
- ✅ **1MB cap enforced on the full payload twice** — `Validate` on raw inbound bytes (incl. unknown top-level fields) and `Finalize` after server enrich — correctly preserving the PR#232 fix that a content+plain subset marshal would have missed.
- ✅ **No server-side SSRF.** The server stores the image URL as a reference and hands it to clients; it never dereferences it (no `http.Get`/`image.Decode`/download in the module). `isAllowedImageURLScheme` enforces an http/https allowlist that rejects `data:`/`javascript:`/`file:`.
- ✅ **Downstream send / audit / call_count flow is shared** with the text path — both branches converge on the same `SendMessageWithResult` + `submitRecordSuccess` tail; audit `byteSize` is the original request body length for both. No divergence.

## 2. Issues found

No P0/P1. The following are **P2 (non-blocking)** — fine to merge as-is, worth a follow-up or a release note.

### P2-1 — `msg_type` strict validation is a narrow backward-compat break (`api.go:774-806`)
Before this PR `pushPayloadReq` had no `MsgType` field, so any stray `msg_type` key in a caller's JSON was silently ignored and the request always took the text path. Now an unrecognized non-empty value (e.g. `"markdown"`, `"notification"`) returns 400 `reason=msg_type` even with valid `content`. Risk is low (the field is brand-new and undocumented, and omitted/`""`/`"text"` are byte-for-byte preserved), but it is a real strictness change. **Suggestion:** call it out in the release notes — the README already documents it well.

### P2-2 — rich-text path skips the `maxContentRunes` (4000) cap the text path enforces (`api.go:787-790`)
Intentional and documented. Under the default 8KB body cap the aggregate text can't exceed ~8KB, so the rune cap is effectively subsumed and the 413/1MB branch is currently unreachable in production config. The asymmetry only becomes material if an operator raises `DM_INCOMINGWEBHOOK_MAX_BYTES` well above 8KB — at which point the rich-text path would admit up to ~1MB of concatenated text the text path would reject (a client-render load concern, not a security issue). **Suggestion:** if parity matters later, cap the summed text-block content (or the recomputed `plain`); otherwise note the body-cap dependency as a load-bearing invariant near the env-var definition, not only at the call site.

## 3. Suggestions (nits, optional)

- **Avatar URL has no scheme validation** (`resolveFromIdentity`, `api.go:900-904`). `req.AvatarURL` is only byte-truncated, while image block URLs go through the http/https allowlist. This is pre-existing (shared with the text path) and safe today because the server never dereferences it, but if clients ever render the avatar as an ``, applying the same scheme allowlist would close a `data:`/`javascript:` rendering vector.
- **Internal-IP / SSRF defense is absent from `isAllowedImageURLScheme`** (octo-lib). Safe only because nothing server-side fetches these URLs. Worth a precondition comment in octo-lib ("server must never dereference") so a future thumbnailer/link-preview/LLM component doesn't silently turn this into an SSRF.
- **Minor per-request serialization overhead**: a rich-text push marshals the payload map ~4× and parses it ~2× (`Validate` → `Finalize` → `util.ToJson` at send). Negligible under the 8KB cap; `ValidateRichTextPayload` already returns the parsed `*RichTextPayload`, so a refactor could collapse `Validate`+`Finalize` into one parse if it ever matters.

## 4. Notes for the human reviewer (security-sensitive label)

Nothing requires manual override. The two items most worth a human's eye are both **conditional, not active**: (1) the URL allowlists (image + avatar) have no internal-IP defense and rely on the invariant that the server never fetches these URLs — safe today, fragile if that invariant ever changes; (2) the rich-text/text length-cap asymmetry only opens if the body cap is tuned up. Neither blocks merge.

PR #299, opened 2026-06-06
feat(incomingwebhook): delivery failure observability + test push + text alias (Phase 2)

an9xyz:
## Background

**Phase 2** of epic #297. Make push delivery debuggable for senders and improve onboarding UX.

> ⚠️ **Base retargeted to `main` to run the full CI pipeline.** While #298 (Phase 1) is still open, this PR's diff **temporarily includes Phase 1's commits** (this branch is stacked on Phase 1). Once #298 merges into `main`, this diff automatically shrinks to Phase-2-only. **Recommended merge order: merge #298 first, then this.** To review only Phase 2's own changes, use the [compare range](https://github.com/Mininglamp-OSS/octo-server/compare/feat/incoming-webhook-richtext...feat/incoming-webhook-observability).

## Changes

**Failure observability**
- Migration: add `status`/`reason`/`http_status`/`adapter` to `incoming_webhook_audit` (MySQL 8 INSTANT add-column; historical rows default to success, `http_status=0`=unknown).
- Record every **post-authentication** delivery outcome (success + failure: `body`/`json`/`content`/`blocks`/`msg_type`/`too_large`/`delivery_failed`), reusing the existing bounded async `auditSem`; `reason`/`http_status` mirror the push response.
- **Anti-enumeration invariant preserved**: pre-auth failures (unknown/bad token, disbanded group) are NOT recorded — they only spend the IP failure budget.
- **Rate-limit 429 is intentionally NOT audited**: it is naturally high-frequency, so per-request recording would amplify DB writes and defeat the limiter's cheap load-shedding; the info is already conveyed by the 429 + rate-limit headers.

**Management UX**
- `GET .../:webhook_id/deliveries?limit=50` — recent deliveries (success + failure), admin-only read, max 100, never returns token, `id` secondary sort for deterministic recency. Does **not** expose caller `ip` (privacy).
- `POST .../:webhook_id/test` — admin one-click test push (recorded as `adapter=test`, does **not** bump `call_count`; message localized via i18n; failures also audited).
- Native push accepts `text` as an alias for `content` (migration-friendly).

## Testing (TDD)

- Synchronous DB unit test `TestInsertAndQueryAudit`: audit insert + `queryRecentAudits` + limit (deterministic assertions).
- e2e `observability_test.go` (CI): deliveries endpoint / auth / 404, **async failure recording** (`require.Eventually`), anti-enumeration not-recorded, test-push auth / 404, text alias, ip-not-exposed.

## Review follow-ups applied

Includes a self-review `refactor` commit (drop rate_limited auditing, test push excluded from `call_count`, deliveries `id` secondary sort) and a maintainer-review `refactor` commit (drop caller-ip from deliveries, i18n the test-push message, audit failed test pushes, document `http_status=0` legacy rows).

## Checks

`go build` / `go vet` / `gofmt` / unit tests / guard / `make i18n-lint` / `make i18n-extract-check` all pass. DB/e2e run in CI.

Refs #297

https://claude.ai/code/session_011kXG9RmQcEvmEp5ZZcPA9y

### DIFF
--- a/modules/incomingwebhook/README.md
+++ b/modules/incomingwebhook/README.md
@@ -33,6 +33,8 @@ Content-Type: application/json
 ```
 
 - `content`：必填，非空；语义长度上限 4000 rune（`DM_INCOMINGWEBHOOK_MAX_CONTENT_RUNES`）。
+- `text`：`content` 的别名（Slack 等平台习惯用 `text`）。`content` 为空时回退到 `text`，
+  降低从既有集成迁移的改造成本；两者都填以 `content` 为准。
 - `username` / `avatar_url`：可选，覆盖该条消息的展示发送者名/头像（不改 webhook 本身配置）。
 
 ### 2. 富文本 / 图文混排（`msg_type` = `"richtext"`）
@@ -86,3 +88,50 @@ Content-Type: application/json
 | 体积过大 | 413 | 超 body cap 或富文本 >1MB |
 | 投递失败 | 502 | 下游发送失败 |
 | 功能停用 | 404 | 全局开关 `incomingwebhook.enabled=0` |
+
+## 管理端点（群主 / 管理员）
+
+需登录态 + 群管理员权限，路径前缀 `/v1/groups/:group_no/incoming-webhooks`。
+除创建/列出/更新/删除/重置外，Phase 2 新增两个：
+
+### 测试推送
+
+```
+POST /v1/groups/:group_no/incoming-webhooks/:webhook_id/test
+```
+
+向群里发一条固定文案的测试消息，端到端验证配置（群可达、消息能投递）。返回
+`{"status":0,"message_id":<int>}`。会记一条 `adapter=test` 的成功投递，便于在
+deliveries 里与真实流量区分。
+
+### 投递记录（排障）
+
+```
+GET /v1/groups/:group_no/incoming-webhooks/:webhook_id/deliveries?limit=50
+```
+
+倒序返回该 webhook 最近的投递记录（**成功 + 失败**），供发送方排障。`limit` 默认 50、
+上限 100。失败记录的 `reason` / `http_status` 与 push 响应一致，便于对照定位。
+
+```json
+{
+  "list": [
+    {
+      "status": 2, "reason": "blocks", "http_status": 400, "adapter": "native",
+      "ip": "203.0.113.10", "byte_size": 84, "message_id": 0, "created_at": 1749200000
+    },
+    {
+      "status": 1, "reason": "", "http_status": 200, "adapter": "native",
+      "ip": "203.0.113.10", "byte_size": 42, "message_id": 123456, "created_at": 1749199900
+    }
+  ]
+}
+```
+
+- `status`：`1`=成功，`2`=失败。
+- `reason`（失败时）：`rate_limited` / `body` / `json` / `content` / `blocks` / `msg_type` /
+  `too_large` / `delivery_failed`。
+- `adapter`：`native`（推送端点）/ `test`（测试推送）。
+
+> **反枚举不变量**：鉴权失败（未知 webhook / 错 token / 已解散群）**不记入** deliveries，
+> 只进 IP 失败预算——只有「鉴权通过后」的投递结果才落审计。
--- a/modules/incomingwebhook/README.md
+++ b/modules/incomingwebhook/README.md
@@ -100,9 +100,10 @@ Content-Type: application/json
 POST /v1/groups/:group_no/incoming-webhooks/:webhook_id/test
 ```
 
-向群里发一条固定文案的测试消息，端到端验证配置（群可达、消息能投递）。返回
-`{"status":0,"message_id":<int>}`。会记一条 `adapter=test` 的成功投递，便于在
-deliveries 里与真实流量区分。
+向群里发一条测试消息，端到端验证配置（群可达、消息能投递）。文案按出站语言本地化
+（en-US / zh-CN，由 `i18n.OutboundLanguage` 协商）。返回 `{"status":0,"message_id":<int>}`。
+会记一条 `adapter=test` 的投递（成功或失败都记，便于在 deliveries 里与真实流量区分），
+且**不**计入 `call_count` / `last_used_at`（测试不是真实流量）。
 
 ### 投递记录（排障）
 
@@ -118,11 +119,11 @@ GET /v1/groups/:group_no/incoming-webhooks/:webhook_id/deliveries?limit=50
   "list": [
     {
       "status": 2, "reason": "blocks", "http_status": 400, "adapter": "native",
-      "ip": "203.0.113.10", "byte_size": 84, "message_id": 0, "created_at": 1749200000
+      "byte_size": 84, "message_id": 0, "created_at": 1749200000
     },
     {
       "status": 1, "reason": "", "http_status": 200, "adapter": "native",
-      "ip": "203.0.113.10", "byte_size": 42, "message_id": 123456, "created_at": 1749199900
+      "byte_size": 42, "message_id": 123456, "created_at": 1749199900
     }
   ]
 }
@@ -132,6 +133,8 @@ GET /v1/groups/:group_no/incoming-webhooks/:webhook_id/deliveries?limit=50
 - `reason`（失败时）：`body` / `json` / `content` / `blocks` / `msg_type` / `too_large` /
   `delivery_failed`。
 - `adapter`：`native`（推送端点）/ `test`（测试推送）。
+- `http_status`：返回给调用方的状态码。**迁移前的历史成功行为 `0`（未知）**——不伪造成 200。
+- **不返回调用方 `ip`**：审计表仍存 ip 作排查上下文，但出于隐私不向群管理员下发（review 决定）。
 
 > **限流（429）不入审计**：`rate_limited` 是天然高频失败，逐条落库会在重试风暴时放大
 > DB 写入、反噬限流的廉价丢弃；429 + `X-RateLimit-*`/`Retry-After` 头已把信息给到调用方。
--- a/modules/incomingwebhook/README.md
+++ b/modules/incomingwebhook/README.md
@@ -129,9 +129,13 @@ GET /v1/groups/:group_no/incoming-webhooks/:webhook_id/deliveries?limit=50
 ```
 
 - `status`：`1`=成功，`2`=失败。
-- `reason`（失败时）：`rate_limited` / `body` / `json` / `content` / `blocks` / `msg_type` /
-  `too_large` / `delivery_failed`。
+- `reason`（失败时）：`body` / `json` / `content` / `blocks` / `msg_type` / `too_large` /
+  `delivery_failed`。
 - `adapter`：`native`（推送端点）/ `test`（测试推送）。
+
+> **限流（429）不入审计**：`rate_limited` 是天然高频失败，逐条落库会在重试风暴时放大
+> DB 写入、反噬限流的廉价丢弃；429 + `X-RateLimit-*`/`Retry-After` 头已把信息给到调用方。
+> 节流可从 deliveries 里成功记录的稀疏/中断间接观察。
 
 > **反枚举不变量**：鉴权失败（未知 webhook / 错 token / 已解散群）**不记入** deliveries，
 > 只进 IP 失败预算——只有「鉴权通过后」的投递结果才落审计。
--- a/modules/incomingwebhook/api.go
+++ b/modules/incomingwebhook/api.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"strconv"
 	"strings"
@@ -168,6 +169,10 @@ func (w *IncomingWebhook) Route(r *wkhttp.WKHttp) {
 		mgr.PUT("/:group_no/incoming-webhooks/:webhook_id", w.requireMgmtEnabled(), w.update)
 		mgr.DELETE("/:group_no/incoming-webhooks/:webhook_id", w.requireMgmtEnabled(), w.delete)
 		mgr.POST("/:group_no/incoming-webhooks/:webhook_id/regenerate", w.requireMgmtEnabled(), w.regenerate)
+		// 排障：最近投递记录（成功+失败）。只读，与 list 一致不挂 requireMgmtEnabled。
+		mgr.GET("/:group_no/incoming-webhooks/:webhook_id/deliveries", w.deliveries)
+		// 测试推送：管理员一键发一条样例消息验证配置。写操作，挂总开关闸。
+		mgr.POST("/:group_no/incoming-webhooks/:webhook_id/test", w.requireMgmtEnabled(), w.testPush)
 	}
 
 	// 推送类：URL 内 token 鉴权，无 AuthMiddleware。四层限流，由粗到细：
@@ -663,6 +668,119 @@ func (w *IncomingWebhook) regenerate(c *wkhttp.Context) {
 }
 
 // ============================================================
+// 排障 / 测试端点（管理端，群管理员）
+// ============================================================
+
+// deliveries 返回条数控制。
+const (
+	defaultDeliveriesLimit = 50
+	maxDeliveriesLimit     = 100
+)
+
+// testPushContent 是「测试推送」固定文案。webhook 消息本就是任意文本，这里不走 i18n
+// 错误信封（那是错误响应专用）；固定一句中性文案即可，发到群里让管理员确认链路打通。
+const testPushContent = "✅ Incoming Webhook 测试消息：配置成功，链路已打通。"
+
+// deliveries 返回某 webhook 最近的投递记录（成功+失败），供发送方排障。只读、群管理员
+// 可见；绝不返回 token。失败记录的 reason/http_status 与 push 路径返回给调用方的响应
+// 一致，便于对照定位。
+func (w *IncomingWebhook) deliveries(c *wkhttp.Context) {
+	groupNo := c.Param("group_no")
+	webhookID := c.Param("webhook_id")
+	if _, ok := w.requireGroupAdmin(c, groupNo); !ok {
+		return
+	}
+	// 复用 queryManageable：webhook 必须属于该群且未软删除（跨群/不存在→404）。
+	if _, ok := w.queryManageable(c, groupNo, webhookID); !ok {
+		return
+	}
+	limit := defaultDeliveriesLimit
+	if v := c.Query("limit"); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n > 0 {
+			limit = n
+		}
+	}
+	if limit > maxDeliveriesLimit {
+		limit = maxDeliveriesLimit
+	}
+	list, err := w.db.queryRecentAudits(webhookID, limit)
+	if err != nil {
+		w.Error("query deliveries failed", zap.String("webhook_id", webhookID), zap.Error(err))
+		mgmtQueryFailed(c)
+		return
+	}
+	resps := make([]deliveryResp, 0, len(list))
+	for _, a := range list {
+		resps = append(resps, deliveryRespFrom(a))
+	}
+	c.Response(map[string]interface{}{"list": resps})
+}
+
+func deliveryRespFrom(a *auditModel) deliveryResp {
+	return deliveryResp{
+		Status:     a.Status,
+		Reason:     a.Reason,
+		HTTPStatus: a.HTTPStatus,
+		Adapter:    a.Adapter,
+		IP:         a.IP,
+		ByteSize:   a.ByteSize,
+		MessageID:  a.MessageID,
+		CreatedAt:  time.Time(a.CreatedAt).Unix(),
+	}
+}
+
+// testPush 由群管理员触发，向群里发一条固定文案的测试消息，端到端验证 webhook 配置
+// （群可达、消息能投递）。走与正式推送相同的 buildPayload(text) → SendMessage 链路，
+// 并记一条 adapter=test 的成功投递，便于在 deliveries 里与真实流量区分。
+func (w *IncomingWebhook) testPush(c *wkhttp.Context) {
+	groupNo := c.Param("group_no")
+	webhookID := c.Param("webhook_id")
+	if _, ok := w.requireGroupAdmin(c, groupNo); !ok {
+		return
+	}
+	// 群必须仍为 Normal —— 不向已解散/禁用群发测试消息（与 create/enable 一致）。
+	g, err := w.requireActiveGroup(groupNo)
+	if err != nil {
+		w.Error("query group failed", zap.Error(err))
+		mgmtQueryFailed(c)
+		return
+	}
+	if g == nil {
+		mgmtGroupNotFound(c)
+		return
+	}
+	m, ok := w.queryManageable(c, groupNo, webhookID)
+	if !ok {
+		return
+	}
+
+	req := &pushPayloadReq{Content: testPushContent}
+	payload := buildPayload(m, req)
+	resp, err := w.ctx.SendMessageWithResult(&config.MsgSendReq{
+		Header:      config.MsgHeader{RedDot: 1},
+		ChannelID:   m.GroupNo,
+		ChannelType: common.ChannelTypeGroup.Uint8(),
+		FromUID:     m.WebhookID,
+		Payload:     []byte(util.ToJson(payload)),
+	})
+	if err != nil {
+		w.Error("send test webhook message failed",
+			zap.String("webhook_id", m.WebhookID), zap.Error(err))
+		mgmtOperationFailed(c)
+		return
+	}
+	var msgID int64
+	if resp != nil {
+		msgID = resp.MessageID
+	}
+	w.submitSuccess(m, len(testPushContent), clientIP(c.Request), msgID, adapterTest)
+	c.Response(map[string]interface{}{
+		"status":     0,
+		"message_id": msgID,
+	})
+}
+
+// ============================================================
 // 推送端点
 // ============================================================
 
@@ -745,6 +863,8 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 		allowed = true
 	}
 	if !allowed {
+		// 鉴权已通过、群为 Normal：失败计入投递审计（body 尚未读，byteSize=0）。
+		w.submitFailure(m, 0, ip, "rate_limited", http.StatusTooManyRequests)
 		pushRateLimited(c)
 		return
 	}
@@ -753,16 +873,19 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 	limit := maxBytes()
 	body, err := io.ReadAll(io.LimitReader(c.Request.Body, int64(limit)+1))
 	if err != nil {
+		w.submitFailure(m, 0, ip, "body", http.StatusBadRequest)
 		pushPayloadInvalid(c, "body")
 		return
 	}
 	if len(body) > limit {
+		w.submitFailure(m, len(body), ip, "too_large", http.StatusRequestEntityTooLarge)
 		pushPayloadTooLarge(c)
 		return
 	}
 
 	var req pushPayloadReq
 	if err := json.Unmarshal(body, &req); err != nil {
+		w.submitFailure(m, len(body), ip, "json", http.StatusBadRequest)
 		pushPayloadInvalid(c, "json")
 		return
 	}
@@ -773,13 +896,19 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 	var payload map[string]interface{}
 	switch strings.ToLower(strings.TrimSpace(req.MsgType)) {
 	case "", msgTypeText:
+		// "text" 是 content 的别名：content 为空时回退，降低从既有集成迁移的改造成本。
+		if req.Content == "" {
+			req.Content = req.Text
+		}
 		if strings.TrimSpace(req.Content) == "" {
+			w.submitFailure(m, len(body), ip, "content", http.StatusBadRequest)
 			pushPayloadInvalid(c, "content")
 			return
 		}
 		// content 语义长度上限（按 rune 计），独立于 8KB 字节 body cap：防止单条消息
 		// 正文过长污染所有客户端渲染。超限按 413 拒绝，与 body 超限同语义。
 		if utf8.RuneCountInString(req.Content) > maxContentRunes() {
+			w.submitFailure(m, len(body), ip, "too_large", http.StatusRequestEntityTooLarge)
 			pushPayloadTooLarge(c)
 			return
 		}
@@ -794,14 +923,17 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 			// 空 text 块 / 非 http(s) 图片 url / 缺图片宽高 / 未知块类型 / 超块数上限）
 			// 一律 400 invalid，reason=blocks 供调用方定位。
 			if errors.Is(err, common.ErrRichTextPayloadTooLarge) {
+				w.submitFailure(m, len(body), ip, "too_large", http.StatusRequestEntityTooLarge)
 				pushPayloadTooLarge(c)
 				return
 			}
+			w.submitFailure(m, len(body), ip, "blocks", http.StatusBadRequest)
 			pushPayloadInvalid(c, "blocks")
 			return
 		}
 		payload = p
 	default:
+		w.submitFailure(m, len(body), ip, "msg_type", http.StatusBadRequest)
 		pushPayloadInvalid(c, "msg_type")
 		return
 	}
@@ -818,6 +950,7 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 	if err != nil {
 		w.Error("send incoming webhook message failed",
 			zap.String("webhook_id", m.WebhookID), zap.Error(err))
+		w.submitFailure(m, len(body), ip, "delivery_failed", http.StatusBadGateway)
 		pushDeliveryFailed(c)
 		return
 	}
@@ -828,7 +961,7 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 		msgID = resp.MessageID
 	}
 	// 审计用同一可信 IP（clientIP），而非 gin 可伪造的 c.ClientIP()。
-	w.submitRecordSuccess(m, len(body), ip, msgID)
+	w.submitSuccess(m, len(body), ip, msgID, adapterNative)
 
 	c.Response(map[string]interface{}{
 		"status":     0,
@@ -904,53 +1037,83 @@ func resolveFromIdentity(m *incomingWebhookModel, req *pushPayloadReq) (name, av
 	return truncateUTF8(name, maxFromNameBytes), truncateUTF8(avatar, maxFromAvatarBytes)
 }
 
-// submitRecordSuccess 把审计任务投递给有界并发池：未达上限时异步执行；已达上限时
-// **丢弃**本次审计（仅 Warn）。如此审计占用的 DB 连接总数恒 ≤ auditSem 容量，不会在
-// 洪峰下与主流量抢连接池。审计为非关键路径，溢出丢弃优于回落到请求 goroutine 同步执行
-// （后者请求并发无界时会重新压垮连接池）。
-func (w *IncomingWebhook) submitRecordSuccess(m *incomingWebhookModel, byteSize int, ip string, msgID int64) {
+// submitSuccess 记录一次成功投递：审计 + 累加调用计数(bumpUsed=true)。adapter 标记
+// 来源（native 推送 / test 测试推送）。
+func (w *IncomingWebhook) submitSuccess(m *incomingWebhookModel, byteSize int, ip string, msgID int64, adapter string) {
+	w.submitDelivery(&auditModel{
+		WebhookID:  m.WebhookID,
+		GroupNo:    m.GroupNo,
+		IP:         ip,
+		ByteSize:   byteSize,
+		MessageID:  msgID,
+		Status:     auditSuccess,
+		HTTPStatus: http.StatusOK,
+		Adapter:    adapter,
+	}, true)
+}
+
+// submitFailure 记录一次【鉴权通过后】的失败投递（限流/payload 非法/体积过大/投递失败）。
+// 不累加调用计数(bumpUsed=false)——call_count 语义是「成功调用次数」。reason/httpStatus
+// 与 push 路径返回给调用方的响应保持一致，供 deliveries 端点排障。
+//
+// ⚠️ 仅在 webhook 已通过 token 鉴权且群为 Normal 之后调用：鉴权失败（未知/错 token/
+// 已解散群）绝不落本表，只进 IP 失败预算，维持 push 路径的反枚举不变量。
+func (w *IncomingWebhook) submitFailure(m *incomingWebhookModel, byteSize int, ip, reason string, httpStatus int) {
+	w.submitDelivery(&auditModel{
+		WebhookID:  m.WebhookID,
+		GroupNo:    m.GroupNo,
+		IP:         ip,
+		ByteSize:   byteSize,
+		Status:     auditFailed,
+		Reason:     reason,
+		HTTPStatus: httpStatus,
+		Adapter:    adapterNative,
+	}, false)
+}
+
+// submitDelivery 把审计任务投递给有界并发池：未达上限时异步执行；已达上限时**丢弃**
+// 本次审计（仅 Warn）。如此审计占用的 DB 连接总数恒 ≤ auditSem 容量，不会在洪峰下与
+// 主流量抢连接池。审计为非关键路径，溢出丢弃优于回落到请求 goroutine 同步执行（后者
+// 请求并发无界时会重新压垮连接池）。
+func (w *IncomingWebhook) submitDelivery(audit *auditModel, bumpUsed bool) {
 	select {
 	case w.auditSem <- struct{}{}:
 		go func() {
 			defer func() { <-w.auditSem }()
-			w.recordSuccess(m, byteSize, ip, msgID)
+			w.recordDelivery(audit, bumpUsed)
 		}()
 	default:
 		// 并发已达上限：丢弃审计，保证总 DB 并发有界、不抢占主流量连接池。
 		w.Warn("audit dropped: concurrency cap reached",
-			zap.String("webhook_id", m.WebhookID))
+			zap.String("webhook_id", audit.WebhookID))
 	}
 }
 
-// auditWriteTimeout 限定一次审计（markUsed + insertAudit 两次写）的总耗时上限。
-// recordSuccess 始终跑在独立 goroutine 上（submitRecordSuccess 满载时直接丢弃、不回落
-// 到请求 goroutine），所以这个超时**不影响 push 响应延迟**；它的作用是封顶单个 detached
+// auditWriteTimeout 限定一次审计（markUsed + insertAudit 最多两次写）的总耗时上限。
+// recordDelivery 始终跑在独立 goroutine 上（submitDelivery 满载时直接丢弃、不回落到
+// 请求 goroutine），所以这个超时**不影响 push 响应延迟**；它的作用是封顶单个 detached
 // 审计 goroutine 在 DB 饱和/故障时持有连接池连接的时长，避免慢 DB 下连接被长期占用。
 // 3s 足够正常写入，又能在故障时快速放手（审计本就是非关键路径，失败仅记日志）。
 const auditWriteTimeout = 3 * time.Second
 
-// recordSuccess 写审计 + 累加调用计数。失败仅记日志，不阻塞主流程。
-func (w *IncomingWebhook) recordSuccess(m *incomingWebhookModel, byteSize int, ip string, msgID int64) {
+// recordDelivery 写一条投递审计；bumpUsed 时额外累加调用计数（仅成功路径）。失败仅记
+// 日志，不阻塞主流程。
+func (w *IncomingWebhook) recordDelivery(audit *auditModel, bumpUsed bool) {
 	defer func() {
 		if r := recover(); r != nil {
-			w.Error("recordSuccess panic", zap.Any("recover", r))
+			w.Error("recordDelivery panic", zap.Any("recover", r))
 		}
 	}()
 	// 两次写共用一个截止时间，封顶单个审计 goroutine 在 DB 饱和/故障时持有连接的时长。
 	ctx, cancel := context.WithTimeout(context.Background(), auditWriteTimeout)
 	defer cancel()
-	if err := w.db.markUsed(ctx, m.WebhookID, time.Now()); err != nil {
-		w.Warn("markUsed failed", zap.String("webhook_id", m.WebhookID), zap.Error(err))
-	}
-	audit := &auditModel{
-		WebhookID: m.WebhookID,
-		GroupNo:   m.GroupNo,
-		IP:        ip,
-		ByteSize:  byteSize,
-		MessageID: msgID,
+	if bumpUsed {
+		if err := w.db.markUsed(ctx, audit.WebhookID, time.Now()); err != nil {
+			w.Warn("markUsed failed", zap.String("webhook_id", audit.WebhookID), zap.Error(err))
+		}
 	}
 	if err := w.db.insertAudit(ctx, audit); err != nil {
-		w.Warn("insert audit failed", zap.String("webhook_id", m.WebhookID), zap.Error(err))
+		w.Warn("insert audit failed", zap.String("webhook_id", audit.WebhookID), zap.Error(err))
 	}
 }
 
--- a/modules/incomingwebhook/api.go
+++ b/modules/incomingwebhook/api.go
@@ -26,6 +26,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/modules/base/event"
 	commonmod "github.com/Mininglamp-OSS/octo-server/modules/common"
 	"github.com/Mininglamp-OSS/octo-server/modules/group"
+	"github.com/Mininglamp-OSS/octo-server/pkg/i18n"
 	octoredis "github.com/Mininglamp-OSS/octo-server/pkg/redis"
 	appwkhttp "github.com/Mininglamp-OSS/octo-server/pkg/wkhttp"
 	"github.com/go-redis/redis"
@@ -677,9 +678,16 @@ const (
 	maxDeliveriesLimit     = 100
 )
 
-// testPushContent 是「测试推送」固定文案。webhook 消息本就是任意文本，这里不走 i18n
-// 错误信封（那是错误响应专用）；固定一句中性文案即可，发到群里让管理员确认链路打通。
-const testPushContent = "✅ Incoming Webhook 测试消息：配置成功，链路已打通。"
+// testPushMessage 返回「测试推送」文案，按出站语言本地化。webhook 消息是普通消息体、
+// 不走 i18n 错误信封（那是错误响应专用），故采用与 outbound 邮件一致的做法：用
+// i18n.OutboundLanguage(ctx) 解析协商语言，再选对应文案。支持矩阵为 en-US / zh-CN
+// （默认 zh-CN）。
+func testPushMessage(ctx context.Context) string {
+	if i18n.OutboundLanguage(ctx) == "en-US" {
+		return "✅ Incoming Webhook test message: setup works, the delivery path is live."
+	}
+	return "✅ Incoming Webhook 测试消息：配置成功，链路已打通。"
+}
 
 // deliveries 返回某 webhook 最近的投递记录（成功+失败），供发送方排障。只读、群管理员
 // 可见；绝不返回 token。失败记录的 reason/http_status 与 push 路径返回给调用方的响应
@@ -722,7 +730,6 @@ func deliveryRespFrom(a *auditModel) deliveryResp {
 		Reason:     a.Reason,
 		HTTPStatus: a.HTTPStatus,
 		Adapter:    a.Adapter,
-		IP:         a.IP,
 		ByteSize:   a.ByteSize,
 		MessageID:  a.MessageID,
 		CreatedAt:  time.Time(a.CreatedAt).Unix(),
@@ -754,8 +761,10 @@ func (w *IncomingWebhook) testPush(c *wkhttp.Context) {
 		return
 	}
 
-	req := &pushPayloadReq{Content: testPushContent}
+	msg := testPushMessage(c.Request.Context())
+	req := &pushPayloadReq{Content: msg}
 	payload := buildPayload(m, req)
+	ip := clientIP(c.Request)
 	resp, err := w.ctx.SendMessageWithResult(&config.MsgSendReq{
 		Header:      config.MsgHeader{RedDot: 1},
 		ChannelID:   m.GroupNo,
@@ -766,6 +775,12 @@ func (w *IncomingWebhook) testPush(c *wkhttp.Context) {
 	if err != nil {
 		w.Error("send test webhook message failed",
 			zap.String("webhook_id", m.WebhookID), zap.Error(err))
+		// 记一条 adapter=test 的失败投递，让排障故事对称（成功/失败都可见）。bumpUsed=false。
+		w.submitDelivery(&auditModel{
+			WebhookID: m.WebhookID, GroupNo: m.GroupNo, IP: ip, ByteSize: len(msg),
+			Status: auditFailed, Reason: "delivery_failed",
+			HTTPStatus: http.StatusInternalServerError, Adapter: adapterTest,
+		}, false)
 		mgmtOperationFailed(c)
 		return
 	}
@@ -774,7 +789,7 @@ func (w *IncomingWebhook) testPush(c *wkhttp.Context) {
 		msgID = resp.MessageID
 	}
 	// bumpUsed=false：测试推送不计入 call_count / last_used_at（adapter=test 已可区分）。
-	w.submitSuccess(m, len(testPushContent), clientIP(c.Request), msgID, adapterTest, false)
+	w.submitSuccess(m, len(msg), ip, msgID, adapterTest, false)
 	c.Response(map[string]interface{}{
 		"status":     0,
 		"message_id": msgID,
--- a/modules/incomingwebhook/api.go
+++ b/modules/incomingwebhook/api.go
@@ -773,7 +773,8 @@ func (w *IncomingWebhook) testPush(c *wkhttp.Context) {
 	if resp != nil {
 		msgID = resp.MessageID
 	}
-	w.submitSuccess(m, len(testPushContent), clientIP(c.Request), msgID, adapterTest)
+	// bumpUsed=false：测试推送不计入 call_count / last_used_at（adapter=test 已可区分）。
+	w.submitSuccess(m, len(testPushContent), clientIP(c.Request), msgID, adapterTest, false)
 	c.Response(map[string]interface{}{
 		"status":     0,
 		"message_id": msgID,
@@ -863,8 +864,11 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 		allowed = true
 	}
 	if !allowed {
-		// 鉴权已通过、群为 Normal：失败计入投递审计（body 尚未读，byteSize=0）。
-		w.submitFailure(m, 0, ip, "rate_limited", http.StatusTooManyRequests)
+		// 刻意【不】记 rate_limited 审计：429 + X-RateLimit-*/Retry-After 头已把节流信息
+		// 给到调用方；而 rate_limited 是唯一天然高频的失败类型（其余失败都在限流闸之后、
+		// 已被 per-webhook 5rps 收住），逐条落审计会在重试风暴时放大 DB 写入与 auditSem 溢出
+		// 的 Warn 日志，反噬限流「廉价丢弃」的本意。管理员可从 deliveries 里成功记录的稀疏/
+		// 中断间接看出节流（review 跟进）。
 		pushRateLimited(c)
 		return
 	}
@@ -960,8 +964,9 @@ func (w *IncomingWebhook) push(c *wkhttp.Context) {
 	if resp != nil {
 		msgID = resp.MessageID
 	}
-	// 审计用同一可信 IP（clientIP），而非 gin 可伪造的 c.ClientIP()。
-	w.submitSuccess(m, len(body), ip, msgID, adapterNative)
+	// 审计用同一可信 IP（clientIP），而非 gin 可伪造的 c.ClientIP()。bumpUsed=true：
+	// 真实推送成功累加 call_count / last_used_at。
+	w.submitSuccess(m, len(body), ip, msgID, adapterNative, true)
 
 	c.Response(map[string]interface{}{
 		"status":     0,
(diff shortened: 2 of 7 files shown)


Jerry-Xin (approved):
This PR is in scope for octo-server and fits the incoming webhook module’s existing route, i18n error, and bounded async audit patterns. I found no blocking security or correctness issues.

💬 Non-blocking

🟡 Warning — [modules/incomingwebhook/sql/20260606000001_incomingwebhook_audit_outcome.sql:14](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/sql/20260606000001_incomingwebhook_audit_outcome.sql:14)  
Historical audit rows are marked as successful via `status DEFAULT 1`, but `http_status` defaults to `0`. Since pre-migration rows were success-only, deliveries will expose legacy successes as `http_status: 0`, which conflicts with the response contract and README examples showing successful deliveries as `200`. Consider defaulting/backfilling historical success rows to `200`, or explicitly documenting `0` as “unknown legacy status”.

🟡 Warning — [modules/incomingwebhook/api.go:904](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/api.go:904)  
The `text` alias only applies when `content == ""`. A whitespace-only `content` with a valid `text` still fails validation, even though the validator treats trimmed-empty content as invalid. For migration friendliness, consider falling back when `strings.TrimSpace(req.Content) == ""`.

🔵 Suggestion — [modules/incomingwebhook/api.go:766](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/api.go:766)  
Failed test pushes are returned as management operation failures but are not inserted into deliveries. If the test endpoint is meant to support troubleshooting, recording `adapter=test`, `status=failed`, `reason=delivery_failed`, `http_status=500/502` would make the observability story more complete.

✅ Highlights

- The anti-enumeration invariant is preserved: audit writes only happen after webhook token validation and active-group checks.
- The deliveries endpoint is admin-gated and does not expose token material.
- Reusing the bounded `auditSem` keeps new failure audit writes from becoming an unbounded DB pressure source.

Validation: `go test ./modules/incomingwebhook -run TestIncomingWebhookNoLegacyResponseError` passed. I could not run DB-backed tests locally because MySQL was unavailable at `127.0.0.1:3306`.

Jerry-Xin (approved):
This PR is in scope for octo-server and fits the incoming webhook module’s existing route, i18n error, and bounded async audit patterns. I found no blocking security or correctness issues.

💬 Non-blocking

🟡 Warning — [modules/incomingwebhook/sql/20260606000001_incomingwebhook_audit_outcome.sql:14](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/sql/20260606000001_incomingwebhook_audit_outcome.sql:14)  
Historical audit rows are marked as successful via `status DEFAULT 1`, but `http_status` defaults to `0`. Since pre-migration rows were success-only, deliveries will expose legacy successes as `http_status: 0`, which conflicts with the response contract and README examples showing successful deliveries as `200`. Consider defaulting/backfilling historical success rows to `200`, or explicitly documenting `0` as “unknown legacy status”.

🟡 Warning — [modules/incomingwebhook/api.go:904](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/api.go:904)  
The `text` alias only applies when `content == ""`. A whitespace-only `content` with a valid `text` still fails validation, even though the validator treats trimmed-empty content as invalid. For migration friendliness, consider falling back when `strings.TrimSpace(req.Content) == ""`.

🔵 Suggestion — [modules/incomingwebhook/api.go:766](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/api.go:766)  
Failed test pushes are returned as management operation failures but are not inserted into deliveries. If the test endpoint is meant to support troubleshooting, recording `adapter=test`, `status=failed`, `reason=delivery_failed`, `http_status=500/502` would make the observability story more complete.

✅ Highlights

- The anti-enumeration invariant is preserved: audit writes only happen after webhook token validation and active-group checks.
- The deliveries endpoint is admin-gated and does not expose token material.
- Reusing the bounded `auditSem` keeps new failure audit writes from becoming an unbounded DB pressure source.

Validation: `go test ./modules/incomingwebhook -run TestIncomingWebhookNoLegacyResponseError` passed. I could not run DB-backed tests locally because MySQL was unavailable at `127.0.0.1:3306`.

Jerry-Xin (approved):
This PR is in scope for octo-server and fits the incoming webhook module’s existing route, i18n error, and bounded async audit patterns. I found no blocking security or correctness issues.

💬 Non-blocking

🟡 Warning — [modules/incomingwebhook/sql/20260606000001_incomingwebhook_audit_outcome.sql:14](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/sql/20260606000001_incomingwebhook_audit_outcome.sql:14)  
Historical audit rows are marked as successful via `status DEFAULT 1`, but `http_status` defaults to `0`. Since pre-migration rows were success-only, deliveries will expose legacy successes as `http_status: 0`, which conflicts with the response contract and README examples showing successful deliveries as `200`. Consider defaulting/backfilling historical success rows to `200`, or explicitly documenting `0` as “unknown legacy status”.

🟡 Warning — [modules/incomingwebhook/api.go:904](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/api.go:904)  
The `text` alias only applies when `content == ""`. A whitespace-only `content` with a valid `text` still fails validation, even though the validator treats trimmed-empty content as invalid. For migration friendliness, consider falling back when `strings.TrimSpace(req.Content) == ""`.

🔵 Suggestion — [modules/incomingwebhook/api.go:766](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/api.go:766)  
Failed test pushes are returned as management operation failures but are not inserted into deliveries. If the test endpoint is meant to support troubleshooting, recording `adapter=test`, `status=failed`, `reason=delivery_failed`, `http_status=500/502` would make the observability story more complete.

✅ Highlights

- The anti-enumeration invariant is preserved: audit writes only happen after webhook token validation and active-group checks.
- The deliveries endpoint is admin-gated and does not expose token material.
- Reusing the bounded `auditSem` keeps new failure audit writes from becoming an unbounded DB pressure source.

Validation: `go test ./modules/incomingwebhook -run TestIncomingWebhookNoLegacyResponseError` passed. I could not run DB-backed tests locally because MySQL was unavailable at `127.0.0.1:3306`.

Jerry-Xin (approved):
This PR is in scope for octo-server and fits the incoming webhook module’s existing route, i18n error, and bounded async audit patterns. I found no blocking security or correctness issues.

💬 Non-blocking

🟡 Warning — [modules/incomingwebhook/sql/20260606000001_incomingwebhook_audit_outcome.sql:14](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/sql/20260606000001_incomingwebhook_audit_outcome.sql:14)  
Historical audit rows are marked as successful via `status DEFAULT 1`, but `http_status` defaults to `0`. Since pre-migration rows were success-only, deliveries will expose legacy successes as `http_status: 0`, which conflicts with the response contract and README examples showing successful deliveries as `200`. Consider defaulting/backfilling historical success rows to `200`, or explicitly documenting `0` as “unknown legacy status”.

🟡 Warning — [modules/incomingwebhook/api.go:904](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/api.go:904)  
The `text` alias only applies when `content == ""`. A whitespace-only `content` with a valid `text` still fails validation, even though the validator treats trimmed-empty content as invalid. For migration friendliness, consider falling back when `strings.TrimSpace(req.Content) == ""`.

🔵 Suggestion — [modules/incomingwebhook/api.go:766](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-6sao2ikw/repo/modules/incomingwebhook/api.go:766)  
Failed test pushes are returned as management operation failures but are not inserted into deliveries. If the test endpoint is meant to support troubleshooting, recording `adapter=test`, `status=failed`, `reason=delivery_failed`, `http_status=500/502` would make the observability story more complete.

✅ Highlights

- The anti-enumeration invariant is preserved: audit writes only happen after webhook token validation and active-group checks.
- The deliveries endpoint is admin-gated and does not expose token material.
- Reusing the bounded `auditSem` keeps new failure audit writes from becoming an unbounded DB pressure source.

Validation: `go test ./modules/incomingwebhook -run TestIncomingWebhookNoLegacyResponseError` passed. I could not run DB-backed tests locally because MySQL was unavailable at `127.0.0.1:3306`.

lml2468 (approved):
APPROVED (head `9f6ad7df53c0`, base `feat/incoming-webhook-richtext` — **stacked on #298**).

Phase 2 of #297. Read the full diff of this PR's own changes (against #298, not main), traced the migration / audit hot path / new endpoints / text-alias seam. Design discipline matches the project's auth-vs-observability boundary precisely — every choice has a stated reason that holds up.

## ⚠️ Stacked PR — merge order matters

Base is `feat/incoming-webhook-richtext` (#298), not `main`. Per the PR description "请在 #298 合入后再合本 PR（或合并时把 base 切到 main）" — please honor that. Merging this before #298 would either lose #298's changes or create a non-trivial conflict.

The CI consequence: this PR only ran the PR-generic checks (CodeQL, code-review, label) because the full Build/Test pipeline is gated on a main-based PR. **Don't merge until either (a) #298 lands first then this rebases on main and CI runs clean, or (b) the base is retargeted to main and CI re-runs green on the rebased SHA.** code-review (echo) APPROVED + my unit-test trace pin the logic, but the full integration suite hasn't run on this exact tree.

## ✅ Migration — clean and reversible

`20260606000001_incomingwebhook_audit_outcome.sql`:
- MySQL 8.0 INSTANT algorithm (table-end `ADD COLUMN` triggers it; no explicit `ALGORITHM=INSTANT` needed — verified by reading the `ALTER TABLE` shape). Tail-add columns + `NOT NULL DEFAULT` are exactly the INSTANT-eligible shape, so historical rows get backfilled without a table copy.
- `status=1` default backfills historical rows as `auditSuccess` — correct because pre-#299 the only thing audit ever recorded was successful pushes.
- `http_status=0` for historical rows is documented as "unknown" — defensible (truthful is better than fabricated 200; Jerry-Xin's R1 nit suggesting `200` is fine to defer since the truth is "we don't know what we returned").
- `adapter='native'` backfill matches the pre-#299 reality (test push didn't exist).
- Down migration is symmetric `DROP COLUMN` — reversible.
- `incomingWebhookDB.insertAudit` uses `util.AttrToUnderscore + Record` reflection so new struct fields auto-map. **No insertAudit code change needed** — clean column-add migration pattern.

## ✅ Anti-enumeration invariant preserved (the most important property)

The PR description claims "鉴权前失败（未知/错 token/已解散群）**不**入审计，仅进 IP 失败预算" — traced and confirmed:

- `failAuth` calls `penalizeIPFailure(ip)` + `pushUnauthorized(c)` — **no** `submitFailure` call.
- `submitFailure` is only ever called after a valid `m *incomingWebhookModel` is resolved (i.e. token verified, group resolved). I grep'd all 8 call sites — every one has an `m` already in scope, post-auth.
- `TestDeliveries_AuthFailureNotRecorded` pins this: sends a bad token, asserts deliveries list stays empty.

This is the right boundary. Logging auth failures into a webhook-scoped audit table would let an attacker who guessed *one* valid webhook_id enumerate by reading deliveries (if they later got admin access via separate vector). Keeping pre-auth failures in IP budget only means audit table reveals only post-auth behavior. Correct paranoia.

## ✅ Rate-limit (429) deliberately not audited — the reasoning holds

I initially flagged this as a possible gap, then traced the comment: rate-limited requests are the only naturally high-frequency failure type (every other failure sits behind the per-webhook 5rps cap). Auditing them would (a) write-amplify the audit DB exactly during the storms the rate limiter exists to bound, (b) blow through `auditSem` and produce a flood of "audit dropped" Warn logs that drown out real signal, (c) the 429 + `Retry-After` headers already give the caller actionable info.

The comment also notes "管理员可从 deliveries 里成功记录的稀疏/中断间接看出节流" — that's a real observability path (gap = rate-limited, not unrecorded). Good design call: optimize for signal density, not coverage uniformity.

## ✅ `submitSuccess(bumpUsed)` parameter — test push correctly excluded from usage stats

Test push calls `submitSuccess(..., adapterTest, false)`; native push calls `submitSuccess(..., adapterNative, true)`. `bumpUsed=false` means `call_count` / `last_used_at` aren't bumped, so the admin list's "usage" column reflects real traffic, not operator's connectivity tests. `adapterTest` discriminator still lets admins see test pushes in `deliveries`. Right shape.

## ✅ `id` secondary sort on `queryRecentAudits` — fixes a real bug class

```sql
ORDER BY created_at DESC, id DESC LIMIT ?
```

`created_at` is second-precision; under burst (multiple deliveries in the same second), single-column sort gives engine-dependent order — "most recent" can jitter between calls. `id` is autoinc monotonic, so it's a perfect tiebreaker. This is the kind of "self-review caught" detail the PR description mentions — good discipline.

## ✅ `Text` is a `Content` alias, with the right precedence

```go
if req.Content == "" {
    req.Content = req.Text
}
if strings.TrimSpace(req.Content) == "" {
    submitFailure(..., "content", 400)
    pushPayloadInvalid(c, "content")
    return
}
```

Two design choices worth calling out as deliberate:

1. **Both filled → `content` wins.** Documented in `pushPayloadReq`. Right call: explicit `content` shouldn't be silently overridden by alias.
2. **`Content="   "` does NOT fall back to `Text`.** Jerry-Xin R1 flagged this as a P2 — I think this is actually correct behavior. A whitespace-only `content` is a caller explicitly sending whitespace; the alias semantics is "fall back when content is omitted", not "fall back when content is bad". The `TrimSpace==""` check on the next line still rejects whitespace-only with 400 `reason=content`, so the worst case is one extra 400 instead of silent success — acceptable. Defer.

## ✅ `deliveries` endpoint scope is right

- `requireGroupAdmin` → group-scoped admin only (not webhook owner, not anyone with a token, not even space admins by default).
- `queryManageable` → webhook must belong to that group AND not be soft-deleted. Cross-group probing returns 404, indistinguishable from "not found" — good.
- `deliveryResp` is a whitelisted projection — IP / byte_size / msg_id / status / reason / http_status / adapter / created_at. Notably **NOT** `token` or `token_hash` (which `auditModel` doesn't carry anyway, but worth confirming the projection layer is defensive).
- `defaultDeliveriesLimit=50`, `maxDeliveriesLimit=100` — sensible. `?limit=99999` clamps to 100, can't exhaust the audit table.

## ✅ `submitDelivery` flow — bounded async with drop-on-overflow

`auditSem` is the same bounded-concurrency primitive Phase 1's `submitRecordSuccess` already used. The new `submitDelivery` extends it identically: try non-blocking send to the semaphore channel, fall through to "audit dropped" Warn if full. The auditWriteTimeout cap on `recordDelivery` is on the detached goroutine, not the request — push response latency is unaffected by audit pressure. Same shape as #292's `recordSuccess` audit-semaphore design (verified consistent with the patterns I reviewed there).

## ✅ Test coverage matrix

Unit (no infra) — `TestInsertAndQueryAudit` covers the round-trip + `queryRecentAudits` ordering + limit.

E2E (CI, integration tag presumably) — 10 functions covering both endpoints, async failure recording (`require.Eventually` to ride out the goroutine), anti-enumeration invariant, test-push gating (admin + 404), and text alias. The `require.Eventually` shape is correct for the async audit — exactly what the asynchronicity demands.

- **Jerry-Xin R1 nit #1 — `http_status=0` for historical rows.** Defensible as truthful unknown. If you want to backfill `200` for historical rows, do it in a follow-up migration with explicit `WHERE created_at < <migration_date>` — don't do it implicitly in the DEFAULT.
- **Jerry-Xin R1 nit #2 — `Content="   "` doesn't fall back to `Text`.** I think this is correct (see analysis above). If you want to be more lenient, change the alias check to `strings.TrimSpace(req.Content) == ""` — but I'd leave it as-is.
- **Jerry-Xin R1 nit #3 — `TestPushFailure` doesn't write to deliveries.** I don't see this in the diff; if Jerry-Xin meant the *integration test scaffold* (test that sends a failing push to confirm audit captures it), that's exactly what `TestDeliveries_RecordsPushFailure` does. May be a false-positive of the review bot.
- **Stacked PR + CI scope.** Already noted at the top. Worth re-confirming Build/Test green on the rebased-on-main version before final merge.

## Verdict

Phase 2 is well-scoped: extends audit to record post-auth outcomes, surfaces failures via a tightly-scoped admin endpoint, adds the convenience operations (test push + text alias), keeps the anti-enumeration invariant intact, and avoids the obvious 429-flood trap. The migration is INSTANT-clean and reversible. Test coverage hits the security-relevant invariants directly.

LGTM on the code. **Merge after #298 lands** so the CI runs the full pipeline against main and you don't lose Phase 1 changes.

Cross-PR coordination note: with #298 + #299 stacked, the natural flow is — #298 green & merge → this PR auto-retargets to main (or manual retarget) → CI re-runs → merge. Don't merge this against `feat/incoming-webhook-richtext` directly.

yujiawei (approved):
# Code Review — PR #299 (octo-server)

**Scope:** Phase 2 of the incoming-webhook epic — delivery-failure observability, an admin `deliveries` troubleshooting endpoint, an admin one-click `test` push, and a `text` alias for `content`. Reviewed against base `feat/incoming-webhook-richtext` (stacked PR; commits `54ce6ae` + `9f6ad7d`).

**Verdict: APPROVED.** No P0/P1 issues. The design is careful and the security-sensitive invariants hold. A few P2/nits below are non-blocking.

## 1. Verification

- ✅ **Anti-enumeration invariant preserved.** Every `submitFailure` call in `api.go` (lines 879, 883, 889, 904, 909, 928, 933, 943, 953) sits *after* the token compare (`api.go:837-843`) and the active-group check (`api.go:849-861`). Auth failures (`failAuth`) still go only to the IP failure budget, never to the audit table. The new `db_test.go:TestDeliveries_AuthFailureNotRecorded` pins this. ✅
- ✅ **Bounded async audit unchanged in spirit.** The refactor from `submitRecordSuccess`/`recordSuccess` to `submitSuccess`/`submitFailure` → `submitDelivery` → `recordDelivery` (`api.go:1042-1122`) preserves the `auditSem` bounded-concurrency + drop-on-overflow contract and the 3s `auditWriteTimeout`. `bumpUsed` correctly gates `markUsed` so only real successful pushes touch `call_count`/`last_used_at`. ✅
- ✅ **Migration is safe.** `20260606000001_*.sql` adds four columns at table end → MySQL 8 INSTANT, instant & lock-free. `status DEFAULT 1` correctly backfills historical rows as success; `adapter DEFAULT 'native'` is the right historical value. Down migration drops all four symmetrically. ✅
- ✅ **Reflection column mapping verified.** New `auditModel` fields (`Status`/`Reason`/`HTTPStatus`/`Adapter`) map via `util.AttrToUnderscore`+`UnderscoreName` to `status`/`reason`/`http_status`/`adapter` — matches the migration columns exactly (checked the `HTTPStatus`→`http_status` case in `octo-lib/pkg/util/string.go`). ✅
- ✅ **`queryRecentAudits` index usage.** `WHERE webhook_id=? ORDER BY created_at DESC, id DESC LIMIT n` uses the existing `idx_iwa_webhook_time(webhook_id, created_at)` prefix; the `id` secondary sort is a sound tiebreaker for same-second rows (created_at is second-precision). ✅
- ✅ **No token leakage.** `deliveryResp` (`model.go:128-139`) contains no token field; `TestDeliveries_Endpoint` asserts the response body never contains "token". `deliveries` reuses `queryManageable` so cross-group/deleted webhooks 404. ✅
- ✅ **Auth gating correct.** Both new endpoints call `requireGroupAdmin`; `deliveries` is read-only and intentionally not behind `requireMgmtEnabled` (consistent with `list`), while `test` (a write) is behind it. `limit` is clamped: invalid/≤0 falls back to 50, capped at 100. ✅
- ✅ **`text` alias.** Applied only in the `"" | text` branch (`api.go:903-905`), `content` wins when both set, documented in README. Does not affect the richtext path. ✅
- ✅ **Build/vet/gofmt:** `go build ./modules/incomingwebhook/...` and `go vet` pass; `gofmt -l` is clean on all PR-touched files (only pre-existing `cache.go` is flagged, untouched here). DB/e2e tests defer to CI; the new test file compiles and all referenced helpers (`setupTestEnv`, `createWebhookWithToken`, `pushRichText`, etc.) exist. ✅

## 2. Issues

None at P0/P1.

## 3. Suggestions (P2 / non-blocking)

- **Audit table growth has no reaper.** The original `incoming_webhook_audit` was "success-only, TTL maintained by a cleanup job", and the `idx_iwa_created` index migration references a *planned* `DELETE ... WHERE created_at < ?`. That cleanup job still does not exist in this module, and this PR now also writes failure rows (post-auth, so bounded by legit traffic + the 5rps/webhook cap; `rate_limited` is deliberately excluded). The net growth is modest, but the table remains unbounded. Recommend landing the retention DELETE job as a near-term follow-up so observability data doesn't accumulate indefinitely.
- **`deliveries` exposes source `ip` to group admins.** Source IPs are returned in the response. This is admin-gated and genuinely useful for troubleshooting, but it does surface caller IPs to every group manager — worth a conscious confirmation that this is the intended privacy posture (flagging since this PR is `security_sensitive`).
- **Failed test push is not audited.** `testPush` records an `adapter=test` success on the happy path but returns `mgmtOperationFailed` without an audit row when `SendMessageWithResult` errors. Defensible (the admin sees the failure synchronously), but it's a small asymmetry vs. the native push path that now audits `delivery_failed`.
- **Hardcoded Chinese `testPushContent`.** The fixed test message is Chinese-only with a comment explaining it's a message body (not an i18n error envelope). Reasonable, but non-Chinese deployments will receive a Chinese test message — consider a neutral/emoji-forward string or locale-aware text if multi-locale tenants are in scope.

## 4. Additional notes

- The `test` push goes through `requireActiveGroup` (uncached, authoritative) rather than the push hot-path `cachedRequireActiveGroup` — correct choice for an admin write that should see fresh group state.
- `submitFailure` hardcodes `adapterNative`, which is correct since all failure call sites are on the native push path.
- Stacked-PR note from the description is acknowledged: base is `feat/incoming-webhook-richtext`, so merge order / base-retarget to `main` must be handled at merge time. Not a code issue.

**For human reviewers (security_sensitive):** the anti-enumeration boundary and the no-token-leak guarantee are the two things to manually re-confirm; both are verified above and covered by tests. The `ip`-exposure suggestion is the one item worth an explicit product/privacy sign-off.

Jerry-Xin (dismissed):
Summary: The PR is in scope for octo-server and the implementation is broadly aligned with the incoming webhook module’s auth, i18n error, and bounded async audit patterns. No blocking issues found.

💬 Non-blocking

🟡 Warning: `testPush` can send a message for a disabled webhook because it uses `queryManageable`, which only rejects deleted/cross-group webhooks, and then sends directly through `SendMessageWithResult`. The normal push path rejects `statusDisabled`. If this is intentional for admin-only diagnostics, document and test that behavior; otherwise add a `statusEnabled` check. See [api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ogpc7khl/repo/modules/incomingwebhook/api.go:759) and [api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ogpc7khl/repo/modules/incomingwebhook/api.go:844).

🟡 Warning: `queryRecentAudits` orders by `created_at DESC, id DESC`, but the existing audit index is only `(webhook_id, created_at)`. On high-volume webhooks, MySQL may need extra sorting for the secondary `id` order; consider an explicit `(webhook_id, created_at, id)` index or verify with `EXPLAIN` on production-like data. See [db.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ogpc7khl/repo/modules/incomingwebhook/db.go:187).

🔵 Suggestion: The `text` alias only falls back when `content == ""`, not when `content` is whitespace. That matches strict field precedence, but if migration friendliness is the goal, `strings.TrimSpace(req.Content) == ""` may better match sender expectations. See [api.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ogpc7khl/repo/modules/incomingwebhook/api.go:917).

✅ Highlights

The anti-enumeration boundary is preserved: audit recording starts only after token and active-group checks, while 429 audit amplification is intentionally avoided.

The deliveries endpoint keeps token/IP out of the response and reuses existing admin authorization.

The audit path remains bounded and non-critical, preserving the existing load-shedding design.

Tests: `go test ./modules/incomingwebhook` could not complete locally because MySQL was unavailable at `127.0.0.1:3306`; formatting check with `gofmt -l` returned clean.

lml2468 (dismissed):
APPROVED (head `4cb1a9d80553`).

R2 follow-up addresses every R1 reviewer point with the right shape: privacy-conscious projection narrowing, i18n alignment with the outbound pattern, observability symmetry on the test push, and the http_status=0 doc-tightening I called out as defensible.

## ✅ Each follow-up verified

**1. IP no longer in `deliveryResp` (Steve / privacy nit).**

The audit table still stores `ip` for limit/rate-limit context and ops triage, but `deliveryResp` drops the field. The comment is explicit about this being a deliberate privacy call:

> *"刻意【不】下发调用方 IP：审计表仍存 ip（限流/排查上下文），但向群管理员暴露来源 IP 是隐私取舍，按 review 决定收敛"*

Right shape — keep the truth in audit (deletable only via DB ops), redact at the boundary. The "audit asserts `ip` never appears in response body" test pins the projection so a future refactor can't accidentally re-expose it via a slip in `deliveryRespFrom`. This is the right structural defense.

**2. `testPushMessage` localized via `i18n.OutboundLanguage(ctx)` (yujiawei / Jerry-Xin).**

The previous hardcoded Chinese string `"✅ Incoming Webhook 测试消息：配置成功，链路已打通。"` matched neither the outbound-email i18n pattern nor the error-envelope i18n pattern. The new shape resolves via `i18n.OutboundLanguage(ctx)` → `en-US` / `zh-CN` default. Same pattern outbound email uses; consistent with the rest of the project's i18n strategy for *message bodies* (not error envelopes — those have their own).

Tiny nit, not blocking: `testPushMessage` selects between only two literals. If the supported language list ever grows, a `map[string]string` keyed by language code with `zh-CN` as the default fallback would scale better than an if/else chain. Today with two languages, two branches is fine.

**3. Failed test push now writes `adapter=test, status=failed, reason=delivery_failed` audit row.**

R1 was asymmetric: success test pushes wrote audit, failed ones returned 500 with no trace. Now both branches write, and the failure carries the same `reason="delivery_failed"` shape as the native-push delivery-failed branch (consistency with the rest of the failure-reason taxonomy). `bumpUsed=false` preserved — test push still doesn't pollute usage stats whether it succeeded or failed. Right call.

The `submitDelivery` direct call (vs going through `submitFailure`) is also correct here because `submitFailure` hardcodes `Adapter: adapterNative`. Could refactor to take `adapter` as a parameter eventually, but the inline construction is fine for this one-off; the audit row's content is identical to what `submitFailure` would build minus the adapter override.

**4. `http_status=0` for historical rows documented (my R1 + Jerry-Xin R1).**

Both the migration column comment and the README now state the historical-row backfill is `0=unknown`, *not* a fabricated `200`. That's the right call (truthful unknown > fabricated success) — documenting the choice means the next person who notices "deliveries before 2026-06-06 all show http_status=0" sees "by design, http response was not recorded pre-Phase 2" instead of suspecting a bug.

## Unchanged but worth re-confirming on this SHA

- Anti-enumeration invariant — `submitFailure` still only called post-auth (auth-failure path remains IP-budget only).
- Rate-limit (429) still not audited — the design rationale comment carries over.
- `bumpUsed` discriminator on success/fail — preserved.
- `id` DESC secondary sort on `queryRecentAudits` — preserved.
- `Text` content alias — preserved.

R1's deeper invariants (audit boundary, dedup, MySQL INSTANT-eligible migration shape) are all untouched.

## Verdict

Clean iteration: 5 reviewer points addressed in 1 commit, each with the right shape (privacy-redact at boundary not delete at source; i18n at message-body layer; symmetric failure audit; honest doc on legacy data). The willingness to document the `http_status=0` decision rather than backfill `200` is the same discipline as #295 R3 / #298 R2 — "document the asymmetry / known unknown honestly, don't paper over it."

LGTM, ready to merge into `feat/incoming-webhook-richtext` (with the same caveat as R1: **wait for #298 to land in main first, then this should rebase onto main and CI re-runs the full Build/Test pipeline**).

yujiawei (dismissed):
# Code Review — PR #299 (octo-server)

**Scope:** `modules/incomingwebhook` Phase 2 — delivery-failure observability, admin test push, and `text` content alias. Reviewed at head `4cb1a9d`. This is a stacked PR on `feat/incoming-webhook-richtext`. Classified security-sensitive; reviewed with extra care on the auth / anti-enumeration boundary.

**Verdict: APPROVED.** No P0/P1. The security-critical invariants hold and the change is well-tested. Findings below are all non-blocking (P2 / nit) suggestions and follow-ups.

## 1. Verification (what I checked and confirmed ✅)

- ✅ **Build & vet clean** — `go build ./modules/incomingwebhook/...` and `go vet` both pass.
- ✅ **Anti-enumeration invariant preserved.** Every new `submitFailure` call in `push` (`api.go:895, 900, 907, 923, 930, 945, 949, 955, 972`) sits *after* token auth (`api.go:854`) and the group-Normal gate (`api.go:863-873`). Pre-auth failures (unknown/soft-deleted webhook, bad token, disbanded group) still only feed the IP failure budget and are never written to the audit table. `TestDeliveries_AuthFailureNotRecorded` pins this. No webhook-existence leak via the audit table.
- ✅ **No token / no caller-IP leak in `deliveries`.** `deliveryResp` (`model.go:596-604`) omits both. `TestDeliveries_Endpoint` asserts `NotContains "token"` and `NotContains "1.2.3.4"`.
- ✅ **Authorization on new endpoints.** Both `deliveries` and `testPush` enforce `requireGroupAdmin` + `queryManageable` (group ownership, cross-group → 404, soft-deleted → 404). `testPush` additionally requires group Normal (`requireActiveGroup`). Covered by `TestDeliveries_RequiresAdmin/NotFound`, `TestTestPush_RequiresAdmin/NotFound`.
- ✅ **`rate_limited` (429) deliberately not audited** — the rationale (avoids DB-write amplification under retry storms, info already in `X-RateLimit-*`/`Retry-After`) is sound. All other auditable failures are behind the per-webhook 5rps limiter, so a valid-token holder cannot exceed ~5 audit writes/sec/webhook — no new write-flood vector.
- ✅ **Migration safety.** `ADD COLUMN` at table end on MySQL 8.0 uses INSTANT (no table rebuild). Historical-row backfill is correct: `status DEFAULT 1` (success), `http_status DEFAULT 0` (honestly "unknown", not faked 200), `adapter DEFAULT 'native'`. New columns map automatically via `util.AttrToUnderscore` reflection. Down migration is a clean symmetric drop.
- ✅ **Query correctness.** `queryRecentAudits` is fully parameterized (no injection), limit clamped to [1,100], and hits the `idx_iwa_webhook_time(webhook_id, created_at)` prefix; the `id` tiebreaker correctly makes "recent" deterministic within a same-second burst.
- ✅ **i18n / error-facade compliance.** All error paths use `httperr.ResponseErrorL*` + registered codes (`mgmtQueryFailed`/`mgmtGroupNotFound`/`mgmtOperationFailed`); no raw `c.JSON`/`AbortWithStatusJSON`. The `TestIncomingWebhookNoLegacyResponseError` guard already scans `api.go`, so the new handlers are covered.
- ✅ **`bumpUsed` semantics.** Native push → `true` (bumps `call_count`/`last_used_at`); admin test push → `false`. The `submitDelivery`/`recordDelivery` refactor releases `auditSem` in all paths (deferred `<-w.auditSem`); no goroutine leak or double-release.

### P2 — suggestions worth addressing

**P2-1 · `testPush` bypasses the per-webhook 5rps limiter.** (`api.go:742-797` vs `push` at `876-889`)
The native push path runs `allowPerWebhook` before delivery; `testPush` does not. The only frequency cap on test pushes is the coarse per-UID `SharedUIDRateLimiter` (~2rps / burst 60), so a group admin can out-pace the 5rps per-webhook bound, each call delivering a *real* message into the group. Admin-gated and self-limiting (the shared UID bucket also throttles the admin's other API calls), so it's a nuisance, not an external DoS. *Suggestion:* either call `allowPerWebhook(ctx, webhookID)` inside `testPush` to share the 5rps bucket, or document that test relies solely on the per-UID limiter.

**P2-2 · Audit table has no retention/TTL job.** (writers at `api.go:895-972`; index `sql/20260604000001` from the base branch)
This PR roughly multiplies insert volume (failures now recorded, not just successes). Write *rate* is bounded (5rps limiter + `auditSem` overflow-drop), but cumulative *size* is not — the `idx_iwa_created` index was added in anticipation of a `DELETE ... WHERE created_at < ?` cleanup that doesn't exist in-repo yet. Pre-existing gap, made materially worse here. *Suggestion:* confirm the TTL cleanup job (or an external retention/partition policy) is ticketed before failure auditing runs at scale.

**P2-3 · `text` alias precedence is untested.** (`api.go:919-921`, tested in `observability_test.go` `TestPush_TextAlias`)
Two documented behaviors: (a) empty content → fall back to `text`; (b) both set → `content` wins. Only (a) is exercised, and weakly (only `!= 400 / != 401`). A regression doing an unconditional `req.Content = req.Text` would pass every existing test. *Suggestion:* add a case asserting `{"content":"a","text":"b"}` keeps content `"a"` and `{"text":"x"}` resolves to `"x"` (route through the handler — note `buildPayload` itself only reads `req.Content`, the alias resolution lives in the push handler).

### nit — polish / optional

- **nit-1 · `testPush` doesn't gate on webhook `enabled` status.** (`api.go:759-762`) `queryManageable` rejects only soft-deleted/cross-group, not `statusDisabled`, so an admin can send a test through a disabled webhook even though the push path 401s a disabled one (`api.go:844`). No authz/enumeration impact (admin acting on their own group). Arguably intentional (validate link before enabling) — either add a status check to align with push, or document it.
- **nit-2 · Stale doc comment on `auditSem`.** (`api.go:76-82`) Still references the removed `recordSuccess` symbol and a "push 成功后 / 两次 DB 写" model; post-refactor it gates success+failure+test audits with 1-or-2 writes (`markUsed` only when `bumpUsed`). The sibling comment at `api.go:1116` already says "最多两次写" — update 76-82 to match.
- **nit-3 · `queryRecentAudits` secondary `ORDER BY id` triggers a small filesort** outside the composite index. Bounded to ≤100 rows on an admin endpoint — leave as-is unless same-second ordering becomes hot.
- **nit-4 · Failure-audit e2e coverage only exercises `reason=blocks`.** Other branches (content/json/too_large/msg_type/delivery_failed) aren't individually driven through the push path. DB-layer mapping is covered generically by `TestInsertAndQueryAudit`, so this is thin coverage, not a correctness gap. Consider one or two more cases.
- **nit-5 · The `bumpUsed=false` invariant for test push is asserted nowhere.** `TestTestPush_PassesAuth` only verifies `adapter=test`, and only inside `if w.Code == 200` (acceptable per the established stub convention). The distinguishing "test push does NOT bump `call_count`/`last_used_at`" contract has no test. Consider a deterministic before/after `call_count` DB read.

## 3. Human-review call-outs (security-sensitive)

Nothing requires manual security sign-off. The two boundaries a human may want to glance at — both verified correct here — are: (a) the post-auth placement of all `submitFailure` calls (the anti-enumeration line), and (b) that `deliveries` never surfaces token or caller IP. Both are enforced in code and pinned by tests.

## 4. Additional observations

- Methodology: adversarial multi-dimension review (authz/enumeration, SQL/migration, concurrency, DoS/logic, conventions/i18n) with per-finding verification against the actual source. 9 candidate findings → 8 confirmed (all P2/nit), 1 rejected (down-migration `DROP COLUMN` cost — standard reversible-migration practice, not a defect).
- The migration filename (`20260606000001`) doesn't collide with siblings and the down path is reversible.

Overall this is a clean, well-reasoned Phase 2 that preserves the module's hard security invariants. Approving; the P2 items (especially the TTL follow-up and the test-push rate-limit asymmetry) are worth a follow-up but don't block merge.

Jerry-Xin (dismissed):
Summary: Relevance gate passes; this PR extends the existing incoming webhook module and I found no blocking correctness, security, or architecture issues.

💬 Non-blocking

- 🟡 Warning: Failed test-push auditing is implemented at `modules/incomingwebhook/api.go:775`, but the e2e test only asserts `adapter=test` when the test push succeeds and skips the failure-audit assertion when the send path returns 500 at `modules/incomingwebhook/observability_test.go:159`. Since the PR explicitly calls out “audit failed test pushes,” consider adding deterministic coverage for that failure path.

- 🔵 Suggestion: The `text` alias only falls back when `content == ""` at `modules/incomingwebhook/api.go:918`; `"content":"   ","text":"real"` still fails as empty content. That may be intended because `content` takes precedence, but for migration ergonomics it may be worth either documenting this exact precedence or treating whitespace-only content as absent.

✅ Highlights

- Anti-enumeration behavior is preserved: audit recording starts after token and active-group validation, and auth failures stay out of per-webhook deliveries.
- The deliveries response model omits caller IP and token material at `modules/incomingwebhook/model.go:132`, which matches the privacy requirement.
- Audit writes remain bounded through the existing async semaphore, avoiding an unbounded DB-write path under load.
- Rich text payload construction uses a whitelist translation plus shared validation/finalization in `modules/incomingwebhook/richtext.go:53`, which is a good fit for the existing message contract.

Validation note: `go test ./modules/incomingwebhook -run 'TestBuildRichTextPayload|TestIncomingWebhookRespondHelpers'` passed. Full `go test ./modules/incomingwebhook` could not complete locally because MySQL was unavailable on `127.0.0.1:3306`.

lml2468 (dismissed):
APPROVED (head `905aa39d43e8`).

Rebase-only update. The 6 PR commits keep byte-identical content for every Phase 2 file vs my R3 head `4cb1a9d80553`:

| file | blob SHA |
|---|---|
| `modules/incomingwebhook/README.md` | `5eb76d0de6c5` |
| `modules/incomingwebhook/api.go` | `945146e83768` |
| `modules/incomingwebhook/api_i18n.go` | `9ddd362aacef` |
| `modules/incomingwebhook/db.go` | `a3dc93800fda` |
| `modules/incomingwebhook/db_test.go` | `2610c532e4a0` |
| `modules/incomingwebhook/model.go` | `970526a26a70` |
| `modules/incomingwebhook/observability_test.go` | `1116b18a556b` |
| `modules/incomingwebhook/richtext.go` | `b7a31e41d719` |
| `modules/incomingwebhook/richtext_push_test.go` | `aaf662db6c23` |
| `modules/incomingwebhook/richtext_test.go` | `ad7006bb56fe` |
| `modules/incomingwebhook/sql/20260606000001_incomingwebhook_audit_outcome.sql` | `34f924072ab5` |
| `pkg/errcode/incomingwebhook.go` | `c631d68d82ba` |

All 12 blobs verified equal via `gh api repos/.../contents/<path>?ref=<sha>`. The `compare` API shows `ahead 9 / behind 6` because main moved (CodeQL #313, sidebar #312, usersecret module), but none of those base commits touch `modules/incomingwebhook/**` or `pkg/errcode/incomingwebhook.go` — diff against the new base is byte-identical to the old.

## What this means

The R3 verifications all carry forward verbatim:

**Phase 1 piggyback (identical to #298 R3)**
- Default text path byte-identical, `pkg/richtext.Validate`/`Finalize` delegation, cross-Space forgery closed, per-block whitelist, error mapping matrix, no token leak — all 12 unit tests + 5 e2e tests pin the invariants.

**Phase 2 — observability**
- `INSTANT` add-column migration on `incoming_webhook_audit` (status/reason/http_status/adapter), historical rows default to success with `http_status=0` = "unknown" (correctly documented in both migration comment and README, no faked 200s).
- Authenticated failures recorded via the existing bounded `auditSem` (no new goroutine source); pre-auth failures (unknown/bad token, disbanded group) **not** audited — anti-enumeration invariant preserved exactly as in the text path.
- 429 intentionally not audited (avoids retry-storm DB amplification, info already in headers); documented in code + README.

**Phase 2 — management UX**
- `GET .../:webhook_id/deliveries?limit=50` admin-only read; secondary `id` sort gives deterministic recency for same-second rows; never returns token; caller `ip` **dropped** from `deliveryResp` per maintainer review (asserted by test).
- `POST .../:webhook_id/test` admin one-click; `adapter=test` distinguishes it; does **not** bump `call_count`/`last_used_at` (test push isn't real traffic); message localized via `i18n.OutboundLanguage` (en-US/zh-CN); failed test pushes now audited as `adapter=test, status=failed, reason=delivery_failed` (observability symmetry).
- Native push `text` alias for `content` (Slack-style migration friendliness); resolved before validation so reason codes stay correct.

## On the merge order note in the PR description

> Recommended merge order: merge #298 first, then this.

Concur. The PR-description-noted base retargeting (`feat/incoming-webhook-richtext` → `main`) makes the diff temporarily include Phase 1's 3 commits. Once #298 merges, this PR's diff shrinks to Phase-2-only. Compare-range link in description (`feat/incoming-webhook-richtext...feat/incoming-webhook-observability`) is the right escape hatch for Phase-2-only review. No special action needed before #298 merges — GitHub will auto-update the diff.

## Checks

The rebase kicked a fresh CI run on `905aa39`; ride those to green before merge. No behavior risk from the rebase itself.

## Verdict

Pure base move, zero Phase 2 content drift. R3 conclusions carry forward. LGTM, merge once CI confirms green on the new SHA and #298 merges first.

yujiawei (dismissed):
# Code Review — PR #299 (octo-server)

**Scope:** Phase 2 of the incoming-webhook epic — delivery-failure observability, an admin one-click test push, and a `text` content alias. Reviewed at head `905aa39d` against `main`. The branch is stacked on Phase 1 (richtext); this review focuses on the Phase-2-only delta (`feat/incoming-webhook-richtext...HEAD`). Classified security-sensitive, so the auth/anti-enumeration invariants got extra scrutiny.

`go build ./...`, `go vet ./modules/incomingwebhook/...`, and `gofmt` are all clean locally. DB/e2e tests run in CI.

## Verdict: Approve

No P0/P1 issues. The security-critical invariants of this module hold up under adversarial verification; the remaining items are non-blocking P2 hardening / test-coverage suggestions for follow-up.

## What I verified (the security-critical points)

- **Anti-enumeration invariant preserved.** Every `submitFailure` / `submitSuccess` call in the push path (`api.go:895-984`) fires strictly *after* the constant-time token compare (`api.go:854`) and the group-Normal gate (`api.go:863-873`). Pre-auth failures — unknown webhook, soft-deleted, bad token, disbanded group, missing params — go through `failAuth` and write **no** audit row; they only spend the IP failure budget. `TestDeliveries_AuthFailureNotRecorded` pins this. A leaked-but-deleted webhook URL still can't be used to probe existence via the audit/deliveries surface.
- **No sensitive data in `deliveries`.** `deliveryResp` (`model.go:136-144`) exposes only `status/reason/http_status/adapter/byte_size/message_id/created_at`. The stored `ip` and `token_hash` are deliberately never mapped into the response (`api.go:727-737`). `TestDeliveries_Endpoint` seeds `ip=1.2.3.4` and asserts the body contains neither `token` nor the IP.
- **Authorization on the two new endpoints.** Both `deliveries` (GET) and `test` (POST) are gated by `requireGroupAdmin` + `queryManageable`. `queryManageable` returns 404 on cross-group / soft-deleted webhooks, so an admin of group A cannot read group B's deliveries or test-push B's webhook (the same guard already proven by `TestUpdate_RejectsCrossGroupAccess`). `testPush` additionally requires the group to still be Normal.
- **DB write-amplification is bounded.** `rate_limited` (429) is intentionally excluded from audit — it returns before any `submit*` call (`api.go:881-889`) — which is the right call since it's the only naturally high-frequency failure type. All other failure writes sit behind the per-webhook limiter and the bounded `auditSem` (drops + warns on saturation rather than blocking the request goroutine). Failure path is a single insert; success path adds the `markUsed` UPDATE only when `bumpUsed=true`.
- **Query + migration.** `queryRecentAudits` is parameterized (no injection; order columns are compile-time literals), limit-clamped to [1,100] by the caller, and served by the `idx_iwa_webhook_time(webhook_id, created_at)` prefix with `id` as a deterministic in-second tiebreaker. The migration is a tail `ADD COLUMN` (INSTANT on MySQL 8) with sensible backfill defaults (`status=1`, `http_status=0`=unknown for legacy rows, `adapter='native'`).
- **`text` alias** is correctly scoped to the text branch only (richtext path never consults it) with documented content-wins precedence; whitespace-only content still 400s as before.

## Non-blocking follow-ups (P2 / nit)

1. **(P2) `testPush` bypasses the per-webhook rate limiter.** It's mounted on the mgmt group behind `AuthMiddleware` + the per-UID shared limiter only, not `allowPerWebhook`. An admin could flood their own group at ~the per-UID rate. This is an admin-gated nuisance (the admin already has normal messaging privileges), not an external DoS vector, and `auditSem` still bounds DB concurrency — but a per-webhook test cooldown would be a clean hardening.
2. **(P2) `call_count` UPDATE is a hot-row write on every native success.** This is intentional (drives the admin usage display, and `bumpUsed=false` keeps test pushes out of it), and it's rate-limited per row — just noting it as a load characteristic, not a defect.
3. **(P2) Test gaps.** The two load-bearing DoS-control invariants — 429-not-audited and `auditSem`-drop — are implemented correctly but untested. A test that floods past the per-webhook cap and asserts zero new audit rows for the 429s would be high-value and straightforward. Also worth mirroring `TestUpdate_RejectsCrossGroupAccess` for the new `deliveries`/`test` endpoints (the real-webhook-in-another-group IDOR case; current 404 tests only hit the nonexistent-webhook branch).
4. **(nit) Defense-in-depth:** `queryRecentAudits` relies on the caller to clamp `limit`; a negative int would convert to a huge `uint64`. The sole caller guards `n > 0` and caps at 100, so there's no live bug — moving the clamp into the data layer would prevent a future caller from regressing it.
5. **(nit)** `TestDeliveries_Endpoint`'s `NotContains "token"` assertion is a no-op (no response field contains that substring); the `NotContains "<ip>"` assertion is the one doing real work.

## For human attention (security-sensitive label)

The merge-relevant security properties (anti-enumeration ordering, no token/IP leak, cross-group 404, 429-not-audited, bounded audit concurrency) are all enforced in code and covered by tests. The one item a human may want to weigh is the **audit-table retention story**: this PR materially increases write volume by recording failures too, and there is still no TTL/cleanup job (the `idx_iwa_created` index was added in a prior PR in anticipation of a `DELETE ... WHERE created_at < ?` that does not yet exist). Unbounded growth is a pre-existing, recurring gap in this module, not introduced here, but Phase 2 makes it grow faster — worth tracking as its own task.

lml2468 (approved):
APPROVED (head `097e6b82484b`).

Stacked-branch sync only. Phase 2 picked up the same CI-fix commit I just approved on #298 R4. The 2 changed files in this update are identical (byte-for-byte) to their counterparts at #298 head `af3c685ce4c3`:

| file | blob SHA on #299 head | blob SHA on #298 R4 head | match |
|---|---|---|---|
| `modules/incomingwebhook/api_i18n_test.go` | `a7ac271927bc` | `a7ac271927bc` | ✅ |
| `modules/incomingwebhook/richtext_push_test.go` | `ae4c011f4f65` | `ae4c011f4f65` | ✅ |

All 11 other Phase 1/2 files are unchanged vs my R4 head `905aa39d43e8`:

- Phase 1 production: `README.md`, `api.go`, `api_i18n.go`, `model.go`, `richtext.go`, `richtext_test.go`, `pkg/errcode/incomingwebhook.go`
- Phase 2 production: `db.go`, `db_test.go`, `observability_test.go`, `sql/20260606000001_incomingwebhook_audit_outcome.sql`

The commit chain is exactly what stacked-PR discipline requires: the new `fix(incomingwebhook): assert reason-detail in the i18n harness, not e2e` (af3c685) sits between the original Phase 1 commits and the Phase 2 commits, so this branch is now `Phase1 + CI-fix + Phase2` — same shape as #298 once #298 merges into main.

## What this means

- The R4 verifications for #298 (e2e Contains assertions dropped, renderer-level contract test added in `iwhHelperHarness`) automatically apply here for the same reason: byte-identical files.
- The R4 verifications for #299 Phase 2 (audit-outcome migration, deliveries endpoint with no caller IP, test-push i18n + failure audit, text alias, observability symmetry, anti-enumeration preservation) all carry forward verbatim — Phase 2 production code is unchanged.
- No new behavior to verify; this is exactly the kind of mechanical stacked-branch sync that should ride to merge once CI is green.

## Recommended merge order (unchanged)

#298 → #299 → #308. Once #298 merges into main, this PR's diff auto-shrinks to Phase-2-only.

## Verdict

Pure stacked-branch sync of the Phase 1 CI-fix; zero Phase 2 drift. R4 conclusions carry forward. LGTM, merge once CI confirms green on `097e6b8`.

lml2468 (dismissed):
APPROVED (head `6995deb7ad9d`).

Rebase-only update, mirroring #298 R5. The 7 PR commits got new SHAs after rebase but identical messages, and every Phase 1/2 file's blob SHA is byte-identical to my R5 head `097e6b82484b`:

- Phase 1 production (7 files): `README.md`, `api.go`, `api_i18n.go`, `api_i18n_test.go`, `model.go`, `richtext.go`, `richtext_push_test.go`, `richtext_test.go`, `pkg/errcode/incomingwebhook.go` ✅
- Phase 2 production (4 files): `db.go`, `db_test.go`, `observability_test.go`, `sql/20260606000001_incomingwebhook_audit_outcome.sql` ✅

Zero content drift. All R1–R5 verifications carry forward verbatim (Phase 1: default text path byte-identical, `pkg/richtext.Validate`/`Finalize` delegation, cross-Space forgery closed, per-block whitelist, error mapping, no token leak, R4 CI-fix renderer-level reason-detail contract; Phase 2: audit-outcome migration with `INSTANT` add-column, deliveries endpoint sans caller IP, test-push i18n + failure audit, text alias, observability symmetry, anti-enumeration preservation).

Merge order unchanged: #298 → #299 → #308. LGTM, merge once CI confirms green on `6995deb`.

Jerry-Xin (dismissed):
The PR is in scope for `octo-server` and the incoming webhook module. I did not find any blocking correctness, security, or architecture issues.

💬 Non-blocking

🟡 Warning — Test coverage gap: failed test-push auditing is implemented at [modules/incomingwebhook/api.go](/Users/mlamp/.openclaw/tmp/pr-review-_jlrplj_/repo/modules/incomingwebhook/api.go:775), but the e2e test only verifies `adapter=test` when the downstream send succeeds. A focused test with a failing `SendMessageWithResult` path would better lock the “failed test pushes are audited” contract.

🔵 Suggestion — The stale comments at [modules/incomingwebhook/api.go](/Users/mlamp/.openclaw/tmp/pr-review-_jlrplj_/repo/modules/incomingwebhook/api.go:76) still refer to `recordSuccess`; updating them to `recordDelivery` would reduce future confusion.

✅ Highlights

The anti-enumeration boundary is preserved: audit writes only happen after token validation and active-group checks, while unknown/bad tokens and inactive groups stay unaudited.

The deliveries endpoint correctly reuses admin and cross-group checks, avoids token/IP exposure, clamps `limit`, and adds deterministic `created_at, id` ordering.

The rich-text path uses the shared `pkg/richtext` validation/finalization flow and avoids trusting caller-supplied top-level fields.

Verification: I ran the targeted incomingwebhook unit/helper tests, including richtext payload construction and i18n responder tests; they passed.

yujiawei (dismissed):
# Code Review — PR #299 (octo-server)

**Verdict: APPROVED.** Phase 2 delivery-failure observability is well-scoped, the security-sensitive invariants are preserved, and it is covered by both synchronous DB unit tests and async e2e tests. `go build` / `go vet` / `gofmt` pass on all changed files. No blocking (P0/P1) issues found. A few non-blocking notes below.

> Review scoped to Phase-2-only changes (`6ead00f3..6995deb7`), since this branch is stacked on Phase 1 (#298). Recommended merge order matches the PR description: #298 first, then this.

## What I verified

**Anti-enumeration invariant (the key security property) — holds.**
In `push()` (`modules/incomingwebhook/api.go`), every `submitFailure(...)` call sits strictly *after* the three auth gates: constant-time token compare (`subtle.ConstantTimeCompare`), `cachedRequireActiveGroup`, and the per-webhook rate limiter. The first failure audit is the body read at ~L895. Pre-auth failures (unknown/bad token, disbanded group) continue to flow through `failAuth` → IP failure budget only, never an audit row. So the new `deliveries` endpoint cannot become a per-webhook enumeration oracle. The `submitFailure` doc comment also pins this contract explicitly.

**No existence oracle on the read endpoint.**
`deliveries` calls `requireGroupAdmin` *before* `queryManageable`. A non-admin always gets 403 regardless of whether the webhook exists; cross-group / missing webhooks return 404 only after admin is established. `TestDeliveries_RequiresAdmin` / `TestDeliveries_NotFound` lock this in.

**No token / IP leakage.**
`deliveryResp` deliberately omits the token and the caller IP. `TestDeliveries_Endpoint` asserts both (`NotContains "token"`, `NotContains "1.2.3.4"`). The audit table still stores `ip` for internal forensics; it is just not surfaced to group admins. Reasonable privacy trade-off, clearly documented.

**Migration is safe and backfills correctly.**
`20260606000001_incomingwebhook_audit_outcome.sql` adds four trailing columns — INSTANT add-column on MySQL 8, no lock pin needed. Backfill defaults are correct: the audit table previously recorded *successes only* (old `recordSuccess` inserted on the success path), so `status DEFAULT 1` correctly classifies all historical rows as success; `http_status DEFAULT 0` (= unknown, not faked to 200) and `adapter DEFAULT 'native'` are honest. Down migration drops the four columns symmetrically. No collision with the unrelated `incoming_webhook.status` column touched by `20260604000002`.

**Query path.**
`queryRecentAudits` filters by `webhook_id` (prefix of `idx_iwa_webhook_time(webhook_id, created_at)`) and orders `created_at DESC, id DESC`. The `id` tiebreaker is the right call — `created_at` is second-granularity, so without it "recent" ordering would be nondeterministic within a second. `limit` is clamped to `maxDeliveriesLimit` (100) in the handler.

**Refactor `submitRecordSuccess`/`recordSuccess` → `submitSuccess`/`submitFailure`/`submitDelivery`/`recordDelivery`.**
Bounded-concurrency semantics are preserved (overflow drops the audit with a Warn rather than falling back to the request goroutine). `bumpUsed` is correctly `true` only for native success and `false` for failures and test push, matching `call_count`'s "successful real calls" semantics. No stale callers of the old names remain (only doc-comment references in the `auditSem` block).

**Test push & text alias.**
`testPush` requires admin + active group, reuses `buildPayload` → `SendMessageWithResult`, records `adapter=test` (success and failure), and does not bump `call_count`. The `text`→`content` fallback only applies when `content` is empty (Content wins when both set), matching the README. Covered by `TestPush_TextAlias`, `TestTestPush_*`, `TestDeliveries_RecordsPushFailure`, `TestDeliveries_AuthFailureNotRecorded`.

## Items for a human reviewer to confirm (security-sensitive PR)

- **Anti-enumeration ordering** — please sanity-check that no future refactor moves a `submitFailure` above the token-compare gate. This is the load-bearing invariant; the current ordering is correct.
- **IP non-disclosure** — confirm group admins seeing delivery metadata (but not caller IP) matches the product's privacy stance. Code and tests enforce it; this is a policy call.

## Non-blocking notes (P2 / nits)

1. **`byte_size` semantics differ between native and test deliveries.** Native push records `len(body)` (raw request bytes); `testPush` records `len(msg)` (the localized message string). An operator comparing `byte_size` across `adapter=native` vs `adapter=test` rows in `deliveries` is comparing slightly different things. Harmless, but a one-line note on `deliveryResp.byte_size` (or in the README) would avoid confusion.
2. **`message_id` is `int64` serialized as a JSON number.** IDs above 2^53 lose precision in JS clients. This matches existing endpoints (`createResp`, etc.), so it's a pre-existing convention, not introduced here — flagging only for consistency tracking, not as a change request.
3. **`limit` parse is lenient.** Non-numeric or `<= 0` silently falls back to the default 50 (only `> maxDeliveriesLimit` is clamped). That's fine for a debugging endpoint; just confirm callers don't expect a 400 on a bad `limit`.

lml2468 (approved):
APPROVED (head `5736230f2995`).

Phase 1 (#298) merged into main at `0ea05352b74d` (02:58Z), so this PR's diff automatically shrunk from 13 files to 7 — Phase-2-only — exactly as the PR description predicted. The 3 PR commits got new SHAs after the rebase onto post-merge main, but every Phase 2 file's blob SHA is byte-identical to my R6 head `6995deb7ad9d`:

| file | blob unchanged vs R6? |
|---|---|
| `modules/incomingwebhook/README.md` | ✅ |
| `modules/incomingwebhook/api.go` | ✅ |
| `modules/incomingwebhook/db.go` | ✅ |
| `modules/incomingwebhook/db_test.go` | ✅ |
| `modules/incomingwebhook/model.go` | ✅ |
| `modules/incomingwebhook/observability_test.go` | ✅ |
| `modules/incomingwebhook/sql/20260606000001_incomingwebhook_audit_outcome.sql` | ✅ |

Zero Phase 2 drift. The 6 Phase 1 files dropped from the diff because they're now in main via #298 — this is exactly the auto-shrink behavior promised in the PR description.

All R1–R6 verifications carry forward verbatim:

- INSTANT add-column migration on `incoming_webhook_audit` (status/reason/http_status/adapter); historical rows default to success with `http_status=0` = "unknown" (documented).
- Authenticated failures recorded via existing bounded `auditSem`; pre-auth failures NOT audited (anti-enumeration preserved); 429 intentionally not audited.
- `GET .../:webhook_id/deliveries` admin-only read, secondary `id` sort, never returns token, caller `ip` dropped per R4 review (test asserts ip not in response).
- `POST .../:webhook_id/test` admin one-click; `adapter=test`; does NOT bump `call_count`/`last_used_at`; message localized via `i18n.OutboundLanguage`; failed test pushes audited as `adapter=test, status=failed, reason=delivery_failed`.
- Native push `text` alias for `content`.

## What's next

Stack is fully unblocked. Recommended merge: this PR (#299) → #308.

LGTM, merge once CI confirms green on `5736230f2995`.

Jerry-Xin (approved):
Summary: The PR is relevant to `octo-server` and the incoming webhook module; I found no blocking correctness, security, or architecture issues.

💬 Non-blocking:
- 🟡 Warning: `modules/incomingwebhook/observability_test.go:161` only verifies the test-push endpoint passes auth and records `adapter=test` on a 200 response. The failed test-push audit path in `modules/incomingwebhook/api.go:775` is not directly covered, so a regression in failed test-push observability could slip through.
- 🔵 Suggestion: `modules/incomingwebhook/api.go:705` silently falls back to the default limit for malformed or non-positive `limit` values. That is acceptable, but if this endpoint is intended to be strict, consider returning a localized 400 for invalid query input.

✅ Highlights:
- Admin-only delivery reads avoid token and caller IP exposure.
- Post-auth failure auditing preserves the anti-enumeration boundary.
- The async audit concurrency cap is reused instead of adding unbounded DB writes.
- `text` alias behavior is narrow and backward-compatible.

Verification: I attempted `go test ./modules/incomingwebhook`, but the local run could not complete because MySQL at `127.0.0.1:3306` was unavailable.

yujiawei (approved):
# Code Review — PR #299 (octo-server)

**Verdict: APPROVED.** No P0/P1 blockers. The change is well-scoped, defensively designed, and adequately tested. A few non-blocking notes below.

> Reviewed Phase-2-only changes via the range `0ea05352..5736230f` (Phase 1 / #298 commits excluded). Head SHA `5736230f`. Local `go build`, `go vet`, and `gofmt` on the changed files all pass; the test suite compiles (DB/e2e require CI infra).

## 1. Verification

- ✅ **Anti-enumeration invariant preserved** — the most important security property here. `submitFailure` is only reachable *after* both the constant-time token compare (`api.go:854`) and the group-Normal check (`api.go:863-873`). Every pre-auth failure branch (missing params, unknown/deleted webhook, bad token, disbanded group) routes through `failAuth` / `pushUnauthorized` and never touches the audit table — it only spends the per-IP failure budget. Server-side (DB) errors also do not penalize the IP. The e2e `TestDeliveries_AuthFailureNotRecorded` pins this.
- ✅ **Rate-limit 429 intentionally not audited** (`api.go:881-889`) — documented rationale (naturally high-frequency, would amplify DB writes / `auditSem` overflow under retry storms). Sound.
- ✅ **`deliveries` endpoint** — admin-gated (`requireGroupAdmin` + `queryManageable`, so cross-group/deleted → 404), read-only (correctly not behind `requireMgmtEnabled`, consistent with `list`), never returns `token`, and deliberately omits caller `ip` for privacy. `limit` is clamped to `maxDeliveriesLimit=100`. `TestDeliveries_Endpoint` asserts both `token` and the IP string are absent from the body.
- ✅ **`testPush` endpoint** — admin-gated, behind the master toggle (write op), and reuses `requireActiveGroup` (uncached direct read) so it won't push to a disbanded/disabled group. `bumpUsed=false` keeps test traffic out of `call_count`/`last_used_at`; both success and failure are recorded with `adapter=test`, giving a symmetric troubleshooting story.
- ✅ **Migration** (`20260606000001`) — MySQL 8 INSTANT add-column at table end, no explicit ALGORITHM/LOCK needed. `status DEFAULT 1` correctly backfills historical (success-only) rows; `http_status DEFAULT 0` honestly marks legacy rows as unknown rather than faking `200`; `adapter DEFAULT 'native'`. Down migration drops all four columns. Reversible.
- ✅ **Reflection column mapping** — `auditModel` field names map correctly through `util.AttrToUnderscore` → `UnderscoreName`: notably `HTTPStatus → http_status` (the consecutive-capitals case resolves correctly), `MessageID → message_id`, `ByteSize → byte_size`. Confirmed by both the synchronous `TestInsertAndQueryAudit` and the raw-SQL-insert e2e assertions.
- ✅ **`text` alias** — applied only on the `""`/`text` branch, `Content` wins when both set, never affects `richtext`. Backward compatible.
- ✅ **Refactor `submitRecordSuccess`→`submitSuccess`/`submitFailure`/`submitDelivery`** — no stale references to the old names remain; the bounded `auditSem` concurrency cap and 3s detached-write timeout semantics are preserved unchanged.
- ✅ **Status code constants** — `auditSuccess=1`/`auditFailed=2` (not 0/1) is a deliberate, well-reasoned choice so an unset zero value can't be misread as a valid outcome.

## 2. Issues

None at P0/P1.

## 3. Suggestions (non-blocking, P2 / nits)

- **Audit table growth now includes failures.** Previously the table recorded successes only; it now records all post-auth outcomes. This is bounded (pre-auth scans are excluded, and post-auth traffic is capped by the per-webhook 5 rps limiter + `auditSem`), so it is not a DoS vector. However, there is no in-module TTL/cleanup job — the original schema comment defers retention to an external "清理任务/定时 DELETE". Worth confirming that external cleanup exists and is sized for the now-higher write rate, otherwise the table will grow faster than before. (`sql/20260514000003...:22`)
- **`queryRecentAudits` secondary sort.** `idx_iwa_webhook_time(webhook_id, created_at)` covers the equality + primary `created_at DESC` ordering, but the `id` tiebreaker is not in the index, so MySQL may filesort the tied rows. Negligible at `limit ≤ 100`; noting only for completeness. (`db.go:queryRecentAudits`)
- **`limit` parsing** silently falls back to the default on a non-numeric or `<= 0` value (`n > 0` guard). Reasonable, just flagging the silent coercion as intended behavior.

## 4. Security-sensitive — items for a human to confirm

This PR was flagged `security_sensitive`. The two things most worth a human's eye, both of which I believe are handled correctly:

1. **The audit recording boundary is the anti-enumeration line.** Any future refactor that moves a `submitFailure`/`submitSuccess` call *above* the token compare or group-Normal check (or adds recording to a new pre-auth branch) would leak existence/validity of a webhook ID into the per-webhook deliveries feed. The current placement is correct; this is the invariant to guard in code review going forward.
2. **`deliveries` output surface.** It must never regress to exposing `token` or caller `ip`. Both are currently excluded and asserted by tests. The `deliveryResp` struct is an explicit allow-list (good), so adding the raw `auditModel` to a response in future would be the risk to watch.

Nice work on the TDD coverage (sync DB unit test + async-aware e2e with `require.Eventually`) and the thorough inline rationale comments.

PR #28, opened 2026-05-14
fix(group): unsubscribe removed members from all non-deleted thread channels (#27)

an9xyz:
Closes #27.

## Summary

When a member (typically a bot) is removed from a group, the parent group channel was correctly unsubscribed from WuKongIM but **thread (子区) channel subscriptions were leaked**. The removed member therefore kept receiving every thread message over its WuKongIM WebSocket connection — and any bot using that stream re-emitted them as DMs back to the user, looking like the bot was spying after removal.

## Root cause

`removeUserFromGroupThreads` (duplicated verbatim in `modules/group/api.go` and `modules/group/service.go`) listed threads via:

```sql
SELECT thread.short_id
FROM thread
JOIN thread_member ON thread.id = thread_member.thread_id
WHERE thread.group_no=? AND thread_member.uid=? AND thread.status!=3
```

The `JOIN thread_member` made the query return zero rows for any member who never actively called `JoinThread` — which is the default for bots, and also for any non-bot member who never opened a specific thread. With zero rows the helper returned early before `IMRemoveSubscriber`, so all thread channels stayed subscribed.

The mirror path is asymmetric: `addUsersToGroupThreads` subscribes a new member to **every** thread `WHERE group_no=? AND status!=3` (active + archived), regardless of `thread_member`. The remove path must match.

## What changed

- New `modules/group/thread_cleanup.go` with:
  - `queryThreadShortIDsForCleanup(ctx, groupNo)` — returns all non-deleted thread short_ids for a group; archived (`status=2`) threads are included because they can be reactivated via `UnarchiveThread`.
  - `removeUserFromGroupThreadsCleanup(ctx, logger, groupNo, uid, spaceID)` — single source of truth for the cleanup sequence: list threads → best-effort `DELETE thread_member` (no longer gates the loop on row count) → `IMRemoveSubscriber` + pinned cleanup per thread.
- `(*Group).removeUserFromGroupThreads` and `(*Service).removeUserFromGroupThreads` are now one-line delegations to the shared helper so the bug cannot regress in one copy only.
- Added `modules/group/thread_cleanup_test.go` covering:
  - bot in group with **no `thread_member` row** (the Issue #27 reproducer) — still returns all active + archived threads;
  - archived (`status=2`) included, deleted (`status=3`) excluded;
  - empty `group_no` / no-thread group → empty slice, no error;
  - unrelated user's `thread_member` row does not affect output.
[...]

Issue #27, opened 2026-05-14
Removed bot still receives thread (子区) messages — group_member_remove only unsubscribes the parent channel, leaving thread subscriptions intact


an9xyz:
## Description

When a bot is removed from a group (kicked, or cascade-removed via the inviter leaving), `octo-server` correctly unsubscribes the bot from the **parent group channel** in WuKongIM, but **does not** unsubscribe it from any of the group's **thread (子区) channels**. As a result, the bot keeps receiving every thread message in that group via its WebSocket connection to WuKongIM, and re-emits them as DMs to the user.

The cause is a single asymmetric SQL filter in the cleanup helper: subscription on join uses `WHERE group_no=? AND status!=3` (all non-deleted threads), but unsubscription on remove uses `JOIN thread_member ... WHERE thread_member.uid=?`, which only matches threads the user has actively joined. Bots are auto-subscribed to every thread when they enter the parent group (`addUsersToGroupThreads`), but they never call `JoinThread`, so the JOIN returns zero rows → the helper returns early → no `IMRemoveSubscriber` is called for any thread channel.

## Steps to Reproduce

1. Create a group with one or more threads (子区). The threads can be active or archived (both reproduce).
2. Add a bot to the group (via BotFather or normal invite). The bot is now in the WuKongIM subscriber list of the parent channel **and** of every non-deleted thread channel.
3. The bot does **not** call `POST /v1/bot/groups/:group_no/threads/:short_id/join` (default behavior — bots don't auto-join threads).
4. As group admin, kick the bot out of the group: `DELETE /v1/groups/:group_no/members` with the bot UID in `members`.
5. Send a message in **the parent group** — bot does not respond. ✅ Parent channel was properly unsubscribed.
6. Send a message in **any thread** of that group — bot still receives it and DMs the sender. ❌

Reproduces on `mininglamposs/octo-server:latest`.

## Expected Behavior

After a member (especially a bot) is removed from a group — whether by admin kick, self-exit, or cascade — the member's UID should be removed from the WuKongIM subscriber list of **every non-deleted thread** in that group (status `1` active and status `2` archived), symmetric with `addUsersToGroupThreads`. The member must no longer receive any thread messages via WuKongIM push.

## Actual Behavior

[...]

PR #300, opened 2026-06-06
fix(group): unsubscribe removed members from all non-deleted thread channels (#27)

an9xyz:
Refs #27.

> Linkage kept as **Refs**, not **Closes**: the code-level asymmetry is `bug-verified`, but the
> end-to-end symptom ("removed bot keeps receiving thread pushes") still carries `needs-human-verify`
> — it depends on whether WuKongIM gates pushes server-side on membership, which can't be settled from
> octo-server source. This PR is the symmetry fix + de-dupe; flip to `Closes` once the WuKongIM
> `channel/subscribers` before/after evidence lands.

## Summary

When a member (typically a bot) is removed from a group, the parent group channel was correctly
unsubscribed from WuKongIM but **thread (子区) channel subscriptions were leaked**. The removed member
therefore kept receiving every thread message over its WuKongIM WebSocket connection — and any bot
using that stream re-emitted them as DMs back to the user, looking like the bot was spying after removal.

## Root cause

`removeUserFromGroupThreads` (duplicated verbatim in `modules/group/api.go` and `modules/group/service.go`)
listed threads via:

```sql
SELECT thread.short_id
FROM thread
JOIN thread_member ON thread.id = thread_member.thread_id
WHERE thread.group_no=? AND thread_member.uid=? AND thread.status!=3
```

The `JOIN thread_member` returns zero rows for any member who never actively called `JoinThread` — the
default for bots, and for any non-bot member who never opened a specific thread. With zero rows the helper
returned early before `IMRemoveSubscriber`, so all thread channels stayed subscribed.

The mirror path is asymmetric: `addUsersToGroupThreads` subscribes a new member to **every** thread
`WHERE group_no=? AND status!=3` (active + archived), regardless of `thread_member`. The remove path must match.

## What changed

- **New `modules/group/thread_cleanup.go`** — single source of truth:
  - `queryThreadShortIDsForCleanup(ctx, groupNo)` returns all non-deleted thread short_ids. Archived
    (`status=2`) threads are included because they can be reactivated via `UnarchiveThread`; only deleted
    (`status=3`) are excluded.
  - `removeUserFromGroupThreadsCleanup(ctx, logger, groupNo, uid, spaceID)` — list threads → best-effort
    `DELETE thread_member` (no longer gates the loop on row count) → `IMRemoveSubscriber` + pinned +
    conversation_ext cleanup per thread.
- `(*Group).removeUserFromGroupThreads` and `(*Service).removeUserFromGroupThreads` now delegate to the
  shared helper, so the bug cannot regress in one copy only.
- **Aligned the BotFather bot-delete path**: exported `RemoveUserFromGroupThreads` on `group.IService`
  and call it from `modules/botfather/command.go` after the per-group parent-channel `IMRemoveSubscriber`,
  so a deleted bot is also removed from all thread channels (item 3 in the issue's suggested fix).
- Tests in `modules/group/thread_cleanup_test.go` covering: the Issue #27 reproducer (no `thread_member`
  row), archived inclusion, deleted exclusion, cross-group isolation, the no-own-row delete path, and
  empty-input guards.

Behavior change is intentional and surgical: only the thread-channel cleanup is widened to match the
addition path. Parent-channel cleanup and the cascade flows are unchanged. Note this keeps the existing
`conversation_ext.RemoveConvExtForUserInSpace` cleanup (the prior closed PR #28 had dropped it, which would
have been a regression).

## Verification

Ran against the CI-equivalent stack (MySQL 8.0 / Redis 7 / WuKongIM v2.2.4, `OCTO_MASTER_KEY` set,
`test` DB created `utf8mb4_general_ci`):

```
ok  github.com/Mininglamp-OSS/octo-server/modules/group      25.9s   # full package, incl. new #27 tests
ok  github.com/Mininglamp-OSS/octo-server/modules/botfather  17.9s   # full package, incl. delete-bot path
```

New tests, all PASS:
- `TestQueryThreadShortIDsForCleanup_ReturnsAllNonDeletedRegardlessOfMembership` — the direct #27 reproducer
- `TestQueryThreadShortIDsForCleanup_EmptyGroupNo` / `_NoThreads` / `_OnlyDeleted`
- `TestRemoveUserFromGroupThreadsCleanup_DeletesMemberWithoutOwnRow` / `_EmptyInputs`

Existing regressions intact: `TestGroupExit_CascadeBot_AlsoRemovesFromThread`,
`TestRemoveGroupMembers_CascadeRemovesInvitedBots`, `TestDeleteBotCleansUpGroupMembers`.

`go build ./...` and `go vet` clean.

## Out of scope (noted in #27)

- `modules/robot/event.go` `robotMessageListen` group-message dispatch has no group-membership check;
  affects bots polling `/v1/bot/events`, a separate route — worth a separate issue.
- The end-to-end "stale subscriber actually receives a push" question (`needs-human-verify`) needs a
  WuKongIM-attached environment to check `channel/subscribers` for `{groupNo}____{shortID}` before/after a kick.

https://claude.ai/code/session_01LLe3uvs1ZGkUkLexYm4Mo7

---
_Generated by [Claude Code](https://claude.ai/code/session_01LLe3uvs1ZGkUkLexYm4Mo7)_

### DIFF
--- a/modules/botfather/command.go
+++ b/modules/botfather/command.go
@@ -660,6 +660,9 @@ func (h *commandHandler) onDeleteConfirm(fromUID string, input string) {
 			if err != nil {
 				h.Error("从IM频道移除Bot失败", zap.String("groupNo", g.GroupNo), zap.Error(err))
 			}
+			// Issue #27：父群订阅之外，还要对齐摘除该 Bot 在群内所有非删除子区的 IM 订阅，
+			// 否则被删除的 Bot 仍会通过 WuKongIM 持续收到子区消息。
+			h.groupService.RemoveUserFromGroupThreads(g.GroupNo, botID, g.SpaceID)
 		}
 	}
 
--- a/modules/group/api.go
+++ b/modules/group/api.go
@@ -3039,50 +3039,11 @@ func (g *Group) groupExit(c *wkhttp.Context) {
 
 }
 
-// removeUserFromGroupThreads 移除用户在某群下所有子区的成员记录、IM 订阅和置顶
+// removeUserFromGroupThreads 移除用户在某群下所有子区的成员记录、IM 订阅和置顶。
+// 委托给包级 removeUserFromGroupThreadsCleanup（见 thread_cleanup.go，Issue #27），
+// 保留方法签名以最小化调用方改动。
 func (g *Group) removeUserFromGroupThreads(groupNo, uid, spaceID string) {
-	// 查询用户在该群加入的所有子区（shortID 用于构建 IM channelID）
-	type threadInfo struct {
-		ShortID string `db:"short_id"`
-	}
-	var threads []threadInfo
-	_, err := g.db.session.Select("thread.short_id").
-		From("thread").
-		Join("thread_member", "thread.id = thread_member.thread_id").
-		Where("thread.group_no=? AND thread_member.uid=? AND thread.status!=3", groupNo, uid).
-		Load(&threads)
-	if err != nil {
-		g.Error("查询用户子区失败", zap.Error(err), zap.String("groupNo", groupNo), zap.String("uid", uid))
-		return
-	}
-	if len(threads) == 0 {
-		return
-	}
-
-	// 删除成员记录
-	_, err = g.db.session.DeleteFrom("thread_member").
-		Where("uid=? AND thread_id IN (SELECT id FROM thread WHERE group_no=?)", uid, groupNo).
-		Exec()
-	if err != nil {
-		g.Error("删除子区成员失败", zap.Error(err), zap.String("groupNo", groupNo), zap.String("uid", uid))
-		return
-	}
-
-	// 移除 IM 订阅和置顶
-	for _, t := range threads {
-		// 子区 channelID 格式: {groupNo}____{shortID} (与 thread.BuildChannelID 一致)
-		channelID := groupNo + "____" + t.ShortID
-		if rmErr := g.ctx.IMRemoveSubscriber(&config.SubscriberRemoveReq{
-			ChannelID:   channelID,
-			ChannelType: common.ChannelTypeCommunityTopic.Uint8(),
-			Subscribers: []string{uid},
-		}); rmErr != nil {
-			g.Error("移除子区IM订阅者失败", zap.Error(rmErr), zap.String("channelID", channelID), zap.String("uid", uid))
-		}
-		// 清理用户在该子区的置顶
-		user.RemovePinnedForUserInSpace(uid, spaceID, channelID, common.ChannelTypeCommunityTopic.Uint8())
-		conversation_ext.RemoveConvExtForUserInSpace(uid, spaceID, channelID, common.ChannelTypeCommunityTopic.Uint8())
-	}
+	removeUserFromGroupThreadsCleanup(g.ctx, g.Log, groupNo, uid, spaceID)
 }
 
 // addUsersToGroupThreads 新成员入群时，将其加入该群所有子区的 IM 订阅（允许发消息）
--- a/modules/group/e2e_issue27_wukong_test.go
+++ b/modules/group/e2e_issue27_wukong_test.go
@@ -1,0 +1,96 @@
+//go:build wukong_e2e
+
+// End-to-end verification for Issue #27 against a live WuKongIM (channel_type=5,
+// 子区/CommunityTopic). Opt-in only — excluded from normal `go test` by the
+// `wukong_e2e` build tag so it never burdens CI (WuKongIM message-delivery
+// semantics are version-coupled; see PR #300 review discussion).
+//
+// Run (requires the CI-equivalent stack: MySQL+Redis+WuKongIM on default ports,
+// OCTO_MASTER_KEY set, `test` DB as utf8mb4_general_ci):
+//
+//	go test -tags wukong_e2e ./modules/group/ -run TestE2E_Issue27 -v -count=1
+//
+// What it proves, through the REAL fixed code path:
+//  1. A bot subscribed to a thread channel (as addUsersToGroupThreads does on
+//     join) actually RECEIVES messages pushed to that channel — the #27 leak.
+//  2. After removeUserFromGroupThreadsCleanup (the fix) runs for that bot —
+//     even though the bot has NO thread_member row (the exact #27 scenario the
+//     old JOIN query skipped) — the bot STOPS receiving thread messages at the
+//     WuKongIM layer, while a still-subscribed user keeps receiving.
+package group
+
+import (
+	"testing"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+func TestE2E_Issue27_RemovedBotStopsReceivingThreadMessages(t *testing.T) {
+	_, ctx := testutil.NewTestServer()
+	require.NoError(t, testutil.CleanAllTables(ctx))
+	f := New(ctx)
+	ensureThreadTables(t, f)
+
+	const (
+		groupNo = "e2e_g27"
+		shortID = "e2eth01"
+		spaceID = "e2e_space"
+		botUID  = "e2e_bot"   // 入群但从不 JoinThread —— #27 的典型 bot
+		usrUID  = "e2e_user"  // 普通成员，对照组
+		sender  = "e2e_sender"
+	)
+	channelID := groupNo + "____" + shortID
+	ct := common.ChannelTypeCommunityTopic.Uint8() // == 5
+
+	// 入群时挂订阅（与 addUsersToGroupThreads 一致：bot 和 user 都订阅，无 thread_member）
+	require.NoError(t, ctx.IMAddSubscriber(&config.SubscriberAddReq{
+		ChannelID: channelID, ChannelType: ct, Subscribers: []string{botUID, usrUID},
+	}))
+
+	// seed 一个 active 子区，但故意不给 bot 建 thread_member 行（#27 场景）
+	_, err := ctx.DB().InsertInto("thread").
+		Columns("short_id", "group_no", "name", "creator_uid", "status", "version").
+		Values(shortID, groupNo, "e2e-sub", "owner", 1, 1).Exec()
+	require.NoError(t, err)
+
+	sendToThread := func(content string) {
+		require.NoError(t, ctx.SendMessage(&config.MsgSendReq{
+			Header:      config.MsgHeader{RedDot: 1},
+			FromUID:     sender,
+			ChannelID:   channelID,
+			ChannelType: ct,
+			Payload:     []byte(`{"type":1,"content":"` + content + `"}`),
+		}))
+	}
+	hasChannel := func(uid string) bool {
+		convs, serr := ctx.IMSyncUserConversation(uid, 0, 50, "", nil)
+		require.NoError(t, serr)
+		for _, c := range convs {
+			if c.ChannelID == channelID && c.ChannelType == ct {
+				return true
+			}
+		}
+		return false
+	}
+
+	// --- 1. 移除前：bot 是订阅者，应收到子区消息 ---
+	sendToThread("before-remove")
+	time.Sleep(800 * time.Millisecond)
+	require.True(t, hasChannel(botUID), "前置：未移除时 bot 应通过 WuKongIM 收到子区消息（#27 泄漏面）")
+	require.True(t, hasChannel(usrUID), "前置：普通成员应收到")
+
+	// --- 2. 跑修复后的真实 helper（bot 无 thread_member 行，旧 JOIN 查询会直接跳过）---
+	removeUserFromGroupThreadsCleanup(ctx, f.Log, groupNo, botUID, spaceID)
+
+	// --- 3. 移除后：再发一条，bot 不应再收到；user 仍订阅、继续收到 ---
+	sendToThread("after-remove")
+	time.Sleep(800 * time.Millisecond)
+	require.False(t, hasChannel(botUID),
+		"修复生效：removeUserFromGroupThreadsCleanup 后被移除的 bot 必须不再收到子区消息（Issue #27）")
+	require.True(t, hasChannel(usrUID),
+		"对照：仍在订阅的成员必须继续收到，证明摘订阅是精确按 uid 的")
+}
--- a/modules/group/event.go
+++ b/modules/group/event.go
@@ -739,10 +739,12 @@ func (g *Group) handleOrgEmployeeExit(data []byte, commit config.EventCommit) {
 		return
 	}
 	realGroups := make([]string, 0)
+	spaceIDByGroupNo := make(map[string]string, len(groups))
 	for _, groupNo := range req.GroupNos {
 		for _, group := range groups {
 			if groupNo == group.GroupNo {
 				realGroups = append(realGroups, groupNo)
+				spaceIDByGroupNo[groupNo] = group.SpaceID
 				break
 			}
 		}
@@ -805,6 +807,9 @@ func (g *Group) handleOrgEmployeeExit(data []byte, commit config.EventCommit) {
 			commit(err)
 			return
 		}
+		// Issue #27 同型：组织退出也必须摘除该用户在群内所有非删除子区的
+		// IM 订阅，与踢人/退群路径对齐（helper 为 best-effort，失败只记日志）。
+		g.removeUserFromGroupThreads(groupNo, req.Operator, spaceIDByGroupNo[groupNo])
 		// 发送群成员更新命令
 		err = g.ctx.SendCMD(config.MsgCMDReq{
 			ChannelID:   groupNo,
--- a/modules/group/event_orgexit_thread_test.go
+++ b/modules/group/event_orgexit_thread_test.go
@@ -1,0 +1,83 @@
+package group
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestHandleOrgEmployeeExit_AlsoCleansThreads 回归 Issue #27 同型缺口：
+// 组织/部门同步退出（handleOrgEmployeeExit）原本只删 group_member + 摘父群
+// IM 订阅，从不清理子区 —— 与踢人/退群路径不对称，组织驱动的移除会持续
+// 泄漏子区订阅。修复后该 handler 对每个群调用 removeUserFromGroupThreads。
+//
+// 与 cascade 测试同款 DB-level 断言：直查 thread_member 证明子区清理被执行
+// （IM 订阅摘除本身由 thread_cleanup 的单测和共享 helper 保证）。
+func TestHandleOrgEmployeeExit_AlsoCleansThreads(t *testing.T) {
+	svc, userDB := setupServiceTest(t)
+	s := svc.(*Service)
+	f := New(s.ctx)
+	ensureThreadTables(t, f)
+
+	insertTestUsers(t, userDB, "org_owner", "org_leaver")
+
+	spaceID := "space_orgexit"
+	groupNo := "g_orgexit_thread"
+	require.NoError(t, f.db.Insert(&Model{
+		GroupNo: groupNo,
+		Name:    "orgexit-thread",
+		Creator: "org_owner",
+		SpaceID: spaceID,
+		Status:  1,
+	}))
+	require.NoError(t, f.db.InsertMember(&MemberModel{
+		GroupNo: groupNo, UID: "org_owner", Role: MemberRoleCreator,
+		Status: 1, Version: 1, Vercode: fmt.Sprintf("%s@1", util.GenerUUID()),
+	}))
+	require.NoError(t, f.db.InsertMember(&MemberModel{
+		GroupNo: groupNo, UID: "org_leaver", Role: MemberRoleCommon,
+		Status: 1, Version: 1, Vercode: fmt.Sprintf("%s@1", util.GenerUUID()),
+	}))
+
+	// 子区 + leaver 的 thread_member 行（DB 可见的清理证据）
+	res, err := f.ctx.DB().InsertInto("thread").
+		Columns("short_id", "group_no", "name", "creator_uid", "status", "version").
+		Values("orgexit_thread", groupNo, "orgexit-sub", "org_owner", 1, 1).Exec()
+	require.NoError(t, err)
+	threadID, err := res.LastInsertId()
+	require.NoError(t, err)
+	_, err = f.ctx.DB().InsertInto("thread_member").
+		Columns("thread_id", "uid", "role", "version").
+		Values(threadID, "org_leaver", 0, 1).Exec()
+	require.NoError(t, err)
+
+	// 触发组织退出事件 handler
+	payload := config.OrgEmployeeExitReq{
+		Operator: "org_leaver",
+		GroupNos: []string{groupNo},
+	}
+	var commitErr error
+	committed := false
+	f.handleOrgEmployeeExit([]byte(util.ToJson(payload)), func(err error) {
+		committed = true
+		commitErr = err
+	})
+	require.True(t, committed, "handler 必须调用 commit")
+	require.NoError(t, commitErr, "组织退出事件处理不应报错")
+
+	// 核心断言：子区成员记录已被清理（修复前 handler 不会触达 thread_member）
+	var postCount int
+	_, err = f.ctx.DB().Select("count(*)").From("thread_member").
+		Where("thread_id=? AND uid=?", threadID, "org_leaver").Load(&postCount)
+	require.NoError(t, err)
+	assert.Equal(t, 0, postCount, "组织退出必须同步清理子区成员/订阅（Issue #27 同型）")
+
+	// 旁证：群成员已删除，原有语义未回归
+	existInGroup, err := f.db.ExistMember("org_leaver", groupNo)
+	require.NoError(t, err)
+	assert.False(t, existInGroup, "组织退出必须删除 group_member")
+}
--- a/modules/group/service.go
+++ b/modules/group/service.go
@@ -107,6 +107,9 @@ type IService interface {
 	AddGroupMembers(req *AddGroupMembersServiceReq) (*AddGroupMembersServiceResp, error)
 	// RemoveGroupMembers 移除群成员
 	RemoveGroupMembers(req *RemoveGroupMembersServiceReq) (*RemoveGroupMembersServiceResp, error)
+	// RemoveUserFromGroupThreads 清理某用户在某群所有子区的 thread_member 记录、IM 订阅和置顶。
+	// 供 botfather 删除 Bot 等“父群成员已移除、需对齐摘除子区订阅”的外部路径调用（Issue #27）。
+	RemoveUserFromGroupThreads(groupNo, uid, spaceID string)
 	// UpdateGroupInfo 更新群信息
 	UpdateGroupInfo(req *UpdateGroupInfoServiceReq) error
 }
@@ -1847,48 +1850,15 @@ func beginAvatarUpdateEvent(ctx *config.Context, db *DB, groupNo string, memberU
 	return eventID, nil
 }
 
-// removeUserFromGroupThreads 移除用户在某群下所有子区的成员记录、IM 订阅和置顶（直接 SQL）
+// removeUserFromGroupThreads 移除用户在某群下所有子区的成员记录、IM 订阅和置顶。
+// 委托给包级 removeUserFromGroupThreadsCleanup（见 thread_cleanup.go，Issue #27）。
 func (s *Service) removeUserFromGroupThreads(groupNo, uid, spaceID string) {
-	type threadInfo struct {
-		ShortID string `db:"short_id"`
-	}
-	var threads []threadInfo
-	_, err := s.ctx.DB().Select("thread.short_id").
-		From("thread").
-		Join("thread_member", "thread.id = thread_member.thread_id").
-		Where("thread.group_no=? AND thread_member.uid=? AND thread.status!=3", groupNo, uid).
-		Load(&threads)
-	if err != nil {
-		s.Error("query user threads failed", zap.Error(err), zap.String("groupNo", groupNo), zap.String("uid", uid))
-		return
-	}
-	if len(threads) == 0 {
-		return
-	}
+	removeUserFromGroupThreadsCleanup(s.ctx, s.Log, groupNo, uid, spaceID)
+}
 
-	// 删除成员记录
-	_, err = s.ctx.DB().DeleteFrom("thread_member").
-		Where("uid=? AND thread_id IN (SELECT id FROM thread WHERE group_no=?)", uid, groupNo).
-		Exec()
-	if err != nil {
-		s.Error("delete thread member failed", zap.Error(err), zap.String("groupNo", groupNo), zap.String("uid", uid))
-		return
-	}
-
-	// 移除 IM 订阅和置顶
-	for _, t := range threads {
-		channelID := groupNo + "____" + t.ShortID
-		if rmErr := s.ctx.IMRemoveSubscriber(&config.SubscriberRemoveReq{
-			ChannelID:   channelID,
-			ChannelType: common.ChannelTypeCommunityTopic.Uint8(),
-			Subscribers: []string{uid},
-		}); rmErr != nil {
-			s.Error("remove thread IM subscriber failed", zap.Error(rmErr), zap.String("channelID", channelID), zap.String("uid", uid))
-		}
-		// 清理用户在该子区的置顶
-		user.RemovePinnedForUserInSpace(uid, spaceID, channelID, common.ChannelTypeCommunityTopic.Uint8())
-		conversation_ext.RemoveConvExtForUserInSpace(uid, spaceID, channelID, common.ChannelTypeCommunityTopic.Uint8())
-	}
+// RemoveUserFromGroupThreads 导出版，供其他模块（botfather）对齐摘除子区订阅，见接口注释。
+func (s *Service) RemoveUserFromGroupThreads(groupNo, uid, spaceID string) {
+	s.removeUserFromGroupThreads(groupNo, uid, spaceID)
 }
 
 // addUsersToGroupThreads 新成员入群时，将其加入该群所有子区的 IM 订阅（直接 SQL）
--- a/modules/group/thread_cleanup.go
+++ b/modules/group/thread_cleanup.go
@@ -1,0 +1,84 @@
+package group
+
+import (
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+	"github.com/Mininglamp-OSS/octo-server/modules/conversation_ext"
+	"github.com/Mininglamp-OSS/octo-server/modules/user"
+	"go.uber.org/zap"
+)
+
+// queryThreadShortIDsForCleanup 返回某群下所有需要在“成员被踢/退群”时摘除 IM 订阅的子区
+// short_id（active + archived，排除 deleted）。
+//
+// Issue #27：旧实现把 thread JOIN thread_member 过滤，导致没主动 JoinThread 的成员（典型
+// 就是 Bot）查不到子区 → IMRemoveSubscriber 不被调用 → Bot 被踢后仍订阅子区频道并通过
+// WuKongIM WebSocket 持续收子区消息。子区频道的 IM 订阅本就是入群时按“父群成员”批量挂上
+// 的（参考 addUsersToGroupThreads，WHERE status!=3），出群必须对称地按“父群所有非删除子区”
+// 摘订阅。archived（status=2）子区可被 UnarchiveThread 重新激活，因此也必须摘除；只有
+// deleted（status=3）子区的 IM 频道已销毁，排除。
+func queryThreadShortIDsForCleanup(ctx *config.Context, groupNo string) ([]string, error) {
+	if groupNo == "" {
+		return nil, nil
+	}
+	type row struct {
+		ShortID string `db:"short_id"`
+	}
+	var rows []row
+	_, err := ctx.DB().Select("short_id").
+		From("thread").
+		Where("group_no=? AND status!=3", groupNo).
+		Load(&rows)
+	if err != nil {
+		return nil, err
+	}
+	out := make([]string, 0, len(rows))
+	for _, r := range rows {
+		out = append(out, r.ShortID)
+	}
+	return out, nil
+}
+
+// removeUserFromGroupThreadsCleanup 是被踢/退群路径上“清理某用户在某群所有子区的 thread_member
+// 记录、IM 订阅和置顶”的统一入口。原本 (*Group).removeUserFromGroupThreads 与
+// (*Service).removeUserFromGroupThreads 是两份逐字重复的实现且各自带 Issue #27 的同型 bug；
+// 这里合并到一处，避免下次再“修一处漏一处”。
+//
+// 失败只记日志、不中断（与原行为一致）。logger 由调用方传入以保留各自的 module 标签。
+func removeUserFromGroupThreadsCleanup(ctx *config.Context, logger log.Log, groupNo, uid, spaceID string) {
+	if groupNo == "" || uid == "" {
+		return
+	}
+	shortIDs, err := queryThreadShortIDsForCleanup(ctx, groupNo)
+	if err != nil {
+		logger.Error("查询群子区失败", zap.Error(err), zap.String("groupNo", groupNo), zap.String("uid", uid))
+		return
+	}
+	if len(shortIDs) == 0 {
+		return
+	}
+
+	// best-effort 删除 thread_member 行：DELETE 按 uid 过滤，没有匹配行也是 0 rows affected。
+	// 即使删除失败也要继续摘 IM 订阅 —— 不能让 DB 异常导致订阅泄漏。
+	if _, err := ctx.DB().DeleteFrom("thread_member").
+		Where("uid=? AND thread_id IN (SELECT id FROM thread WHERE group_no=?)", uid, groupNo).
+		Exec(); err != nil {
+		logger.Error("删除子区成员记录失败", zap.Error(err), zap.String("groupNo", groupNo), zap.String("uid", uid))
+	}
+
+	for _, shortID := range shortIDs {
+		// 子区 channelID 格式: {groupNo}____{shortID} (与 thread.BuildChannelID 一致)
+		channelID := groupNo + "____" + shortID
+		if rmErr := ctx.IMRemoveSubscriber(&config.SubscriberRemoveReq{
+			ChannelID:   channelID,
+			ChannelType: common.ChannelTypeCommunityTopic.Uint8(),
+			Subscribers: []string{uid},
+		}); rmErr != nil {
+			logger.Error("移除子区IM订阅者失败", zap.Error(rmErr), zap.String("channelID", channelID), zap.String("uid", uid))
+		}
+		// 清理用户在该子区的置顶 / 会话扩展
+		user.RemovePinnedForUserInSpace(uid, spaceID, channelID, common.ChannelTypeCommunityTopic.Uint8())
+		conversation_ext.RemoveConvExtForUserInSpace(uid, spaceID, channelID, common.ChannelTypeCommunityTopic.Uint8())
+	}
+}
--- a/modules/group/thread_cleanup_test.go
+++ b/modules/group/thread_cleanup_test.go
@@ -1,0 +1,154 @@
+package group
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestQueryThreadShortIDsForCleanup_ReturnsAllNonDeletedRegardlessOfMembership
+// 是 Issue #27 的核心回归：群下有 active+archived 子区，但 uid 从未出现在 thread_member
+// 表里（Bot 入群后不会主动 JoinThread 的常见情况）。修复前 SQL 用 JOIN thread_member
+// 过滤导致返回空切片，IMRemoveSubscriber 永不被调用 → Bot 被踢后仍订阅子区频道。
+// 修复后应返回该群所有非 deleted 子区的 short_id。
+func TestQueryThreadShortIDsForCleanup_ReturnsAllNonDeletedRegardlessOfMembership(t *testing.T) {
+	svc, _ := setupServiceTest(t)
+	f := New(svc.(*Service).ctx)
+	ensureThreadTables(t, f)
+
+	const groupNo = "g_issue27_basic"
+	// active (status=1)
+	_, err := f.ctx.DB().InsertInto("thread").
+		Columns("short_id", "group_no", "name", "creator_uid", "status", "version").
+		Values("th_active", groupNo, "active", "owner", 1, 1).Exec()
+	require.NoError(t, err)
+	// archived (status=2) —— 可被 UnarchiveThread 重激活，不能漏摘订阅
+	_, err = f.ctx.DB().InsertInto("thread").
+		Columns("short_id", "group_no", "name", "creator_uid", "status", "version").
+		Values("th_archived", groupNo, "archived", "owner", 2, 1).Exec()
+	require.NoError(t, err)
+	// deleted (status=3) —— 必须排除（IM 频道已销毁）
+	_, err = f.ctx.DB().InsertInto("thread").
+		Columns("short_id", "group_no", "name", "creator_uid", "status", "version").
+		Values("th_deleted", groupNo, "deleted", "owner", 3, 1).Exec()
+	require.NoError(t, err)
+	// 另一个群下的子区，不应被返回
+	_, err = f.ctx.DB().InsertInto("thread").
+		Columns("short_id", "group_no", "name", "creator_uid", "status", "version").
+		Values("th_other_group", "g_other", "other", "owner", 1, 1).Exec()
+	require.NoError(t, err)
+
+	// 关键：不插入任何 thread_member 行 —— 模拟 Bot 入群后未 JoinThread
+	shortIDs, err := queryThreadShortIDsForCleanup(f.ctx, groupNo)
+	require.NoError(t, err)
+	assert.ElementsMatch(t, []string{"th_active", "th_archived"}, shortIDs,
+		"必须返回所有非 deleted 子区，不能依赖 thread_member（Issue #27）")
+}
+
+// TestQueryThreadShortIDsForCleanup_EmptyGroupNo 防御：空 group_no 直接返回空。
+func TestQueryThreadShortIDsForCleanup_EmptyGroupNo(t *testing.T) {
+	svc, _ := setupServiceTest(t)
+	f := New(svc.(*Service).ctx)
+	ensureThreadTables(t, f)
+
+	shortIDs, err := queryThreadShortIDsForCleanup(f.ctx, "")
+	require.NoError(t, err)
+	assert.Empty(t, shortIDs)
+}
+
+// TestQueryThreadShortIDsForCleanup_NoThreads 群下没子区 → 空切片。
+func TestQueryThreadShortIDsForCleanup_NoThreads(t *testing.T) {
+	svc, _ := setupServiceTest(t)
+	f := New(svc.(*Service).ctx)
+	ensureThreadTables(t, f)
+
+	shortIDs, err := queryThreadShortIDsForCleanup(f.ctx, "g_nothreads")
+	require.NoError(t, err)
+	assert.Empty(t, shortIDs)
+}
+
+// TestQueryThreadShortIDsForCleanup_OnlyDeleted 群下只有已删除子区 → 空切片。
+// 保证已删除子区永不被回流为 IMRemoveSubscriber 调用对象。
+func TestQueryThreadShortIDsForCleanup_OnlyDeleted(t *testing.T) {
+	svc, _ := setupServiceTest(t)
+	f := New(svc.(*Service).ctx)
+	ensureThreadTables(t, f)
+
+	const groupNo = "g_only_deleted"
+	_, err := f.ctx.DB().InsertInto("thread").
+		Columns("short_id", "group_no", "name", "creator_uid", "status", "version").
+		Values("th_dead", groupNo, "dead", "owner", 3, 1).Exec()
+	require.NoError(t, err)
+
+	shortIDs, err := queryThreadShortIDsForCleanup(f.ctx, groupNo)
+	require.NoError(t, err)
+	assert.Empty(t, shortIDs)
+}
+
+// TestRemoveUserFromGroupThreadsCleanup_DeletesMemberWithoutOwnRow 验证修复的端到端行为：
+// 即便目标 uid 在 thread_member 没有任何行（Bot 场景），cleanup 也要走完整个流程而不是
+// 提前 return —— 这是和旧实现最关键的差异。这里用一个“别人 join 了、target 没 join”的
+// 子区，断言 cleanup 不会因为 target 没有 thread_member 行而跳过，且 target 自己后来
+// 误留的行（如有）会被清掉，别人的行不受影响。
+func TestRemoveUserFromGroupThreadsCleanup_DeletesMemberWithoutOwnRow(t *testing.T) {
+	svc, _ := setupServiceTest(t)
+	s := svc.(*Service)
+	f := New(s.ctx)
+	ensureThreadTables(t, f)
+
+	const groupNo = "g_issue27_e2e"
+	const targetUID = "bot_no_join"
+	res, err := f.ctx.DB().InsertInto("thread").
+		Columns("short_id", "group_no", "name", "creator_uid", "status", "version").
+		Values("th_e2e", groupNo, "e2e", "owner", 1, 1).Exec()
+	require.NoError(t, err)
+	threadID, err := res.LastInsertId()
+	require.NoError(t, err)
+
+	// 一个无关用户在子区里有 thread_member 行 —— cleanup target 时不能误删
+	_, err = f.ctx.DB().InsertInto("thread_member").
+		Columns("thread_id", "uid", "role", "version").
+		Values(threadID, "someone_else", 0, 1).Exec()
+	require.NoError(t, err)
+
+	// target 自己没有 thread_member 行：旧实现会 JOIN 出 0 行直接 return，新实现必须照常处理。
+	// IMRemoveSubscriber 在无 IM 服务的测试环境里只记日志、不影响 DELETE 路径与断言。
+	s.removeUserFromGroupThreads(groupNo, targetUID, "sp_e2e")
+
+	// 无关用户的 thread_member 行仍在
+	var otherCount int
+	_, err = f.ctx.DB().Select("count(*)").From("thread_member").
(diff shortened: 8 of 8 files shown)


Jerry-Xin (dismissed):
This PR is relevant to octo-server and the change correctly aligns group-member removal with the existing thread subscription add path.

💬 Non-blocking

- 🔵 Suggestion: [modules/group/thread_cleanup_test.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-85uvrvps/repo/modules/group/thread_cleanup_test.go:115) covers the no-`thread_member` reproducer at the DB level, but it does not assert that `IMRemoveSubscriber` is actually invoked for the thread channel. The implementation clearly does call it in [modules/group/thread_cleanup.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-85uvrvps/repo/modules/group/thread_cleanup.go:70), so this is not blocking, but a future mock/fake IM client assertion would make the regression test stronger.

- 🔵 Suggestion: [modules/group/thread_cleanup.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-85uvrvps/repo/modules/group/thread_cleanup.go:72) still duplicates the thread channel ID separator literal. This matches existing group code and avoids an import cycle with `modules/thread`, so it is acceptable here. Longer term, a shared constant/helper outside `modules/thread` would reduce drift risk.

✅ Highlights

- 🟢 The project relevance gate passes: this is a scoped fix to group, bot deletion, and thread-channel subscription cleanup in octo-server.
- 🟢 [modules/group/thread_cleanup.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-85uvrvps/repo/modules/group/thread_cleanup.go:21) correctly queries all non-deleted threads by `group_no` instead of gating on `thread_member`, which fixes the described asymmetry.
- 🟢 [modules/group/thread_cleanup.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-85uvrvps/repo/modules/group/thread_cleanup.go:64) keeps DB cleanup best-effort and still proceeds to IM unsubscribe, which is the right failure behavior for avoiding subscription leaks.
- 🟢 [modules/botfather/command.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-85uvrvps/repo/modules/botfather/command.go:653) now cleans child thread subscriptions when deleting a bot, closing a real gap outside the normal group removal path.
- 🟢 Compile-oriented check passed: `go test -run '^$' ./modules/group ./modules/botfather`. Full package tests could not run locally because MySQL was unavailable at `127.0.0.1:3306`.

yujiawei (dismissed):
# Code Review — PR #300 (octo-server)

## Verdict: APPROVED

The fix is correct, surgical, and well-tested. It resolves the real defect (leaked thread-channel subscriptions for removed members/bots) and consolidates two drift-prone duplicate implementations into a single source of truth. No blocking (P0/P1) issues are introduced by this change. The items below are non-blocking follow-ups; several are pre-existing gaps outside this PR's diff and are better tracked as separate issues.

## 1. Verification of the fix

✅ **Root cause correctly identified and fixed.** The old `removeUserFromGroupThreads` listed threads via `JOIN thread_member ... WHERE uid=?`, returning zero rows for any member who never called `JoinThread` (the default for bots), so `IMRemoveSubscriber` was never reached. The new `queryThreadShortIDsForCleanup` (`modules/group/thread_cleanup.go:23-41`) drops the join and selects `WHERE group_no=? AND status!=3`.

✅ **Add/remove symmetry is now correct.** `addUsersToGroupThreads` subscribes new members to every `WHERE group_no=? AND status!=3` thread regardless of `thread_member` (`modules/group/service.go` and `modules/group/api.go`). The remove path now mirrors that exactly, including archived (`status=2`) threads — correct, since those can be reactivated via `UnarchiveThread`.

✅ **De-duplication.** Both `(*Group).removeUserFromGroupThreads` (`api.go:3045`) and `(*Service).removeUserFromGroupThreads` (`service.go:1855`) now delegate to the shared `removeUserFromGroupThreadsCleanup`, so the bug class cannot regress in one copy only.

✅ **Loop no longer gated on row count.** Removing the `if len(threads) == 0 { return }` early-exit before `IMRemoveSubscriber` is the crux of the fix; `removeUserFromGroupThreadsCleanup` only returns early when there are genuinely no non-deleted threads in the group.

✅ **`conversation_ext` cleanup preserved.** `thread_cleanup.go:82` retains `RemoveConvExtForUserInSpace`, avoiding the regression that the previously-closed alternative had introduced.

✅ **BotFather delete path aligned.** `modules/botfather/command.go:665` now calls `RemoveUserFromGroupThreads` after the parent-channel `IMRemoveSubscriber`. Ordering is fine: the helper reads the `thread` table (not `group_member`), so it is independent of the later `group_member.is_deleted` update.

✅ **Build & wiring.** `go build ./modules/group/... ./modules/botfather/...` is clean. `g.Log`/`s.Log` are valid (both structs embed `log.Log`). `g.SpaceID` flows from the DB `group` record through `InfoResp`. (Tests could not be executed in the review sandbox — no MySQL/migration stack — but the author reports a full passing run against the CI-equivalent stack and the new tests are sensibly targeted.)

## 2. Issues found

No P0/P1 introduced by this PR. The notes below are **non-blocking**.

### P2 / follow-up (pre-existing, out of this PR's diff)

- **`thread_setting` is not cleaned on the group-removal path.** `removeUserFromGroupThreadsCleanup` cleans `thread_member`, IM subscriptions, pinned, and `conversation_ext`, but not `thread_setting`. A user who only muted a thread (without joining) can leave an orphaned `thread_setting` row. Note this is **not a regression** — the group path never cleaned `thread_setting` before this PR either; only the separate `thread.Service.RemoveUserFromGroupThreads` does, and that method is not on the group-removal path. Worth a follow-up issue rather than a change here.

- **Org-sync exit path has the same asymmetry the PR just fixed elsewhere.** `handleOrgEmployeeExit` (`modules/group/event.go:712+`) deletes `group_member` and removes the parent-channel IM subscription but never calls `removeUserFromGroupThreads`, so an org/department-driven removal still leaks thread subscriptions. Same bug class as #27, on a path this PR does not touch. `SpaceID` is available there, so a follow-up fix is straightforward. Recommend a separate issue.

- **The sibling `thread.Service.RemoveUserFromGroupThreads` still carries the `JOIN thread_member` scope** (`modules/thread/service.go:1078`). It is currently unreferenced by production code (only the group-module version is used by BotFather), so it is latent tech debt, not a live bug — but it is the same trap and could mislead a future caller. Consider aligning or removing it.

### nit

- **Two `RemoveUserFromGroupThreads` with different signatures/semantics** now coexist (`group.Service`: `(groupNo, uid, spaceID)`, void, non-transactional; `thread.Service`: `(groupNo, uid)`, returns error, transactional + `thread_setting`). Different packages so no Go ambiguity, but the name/semantics divergence is a maintainability snag worth a doc comment cross-reference.

- **Fan-out on bulk removal.** `RemoveGroupMembers` issues `IMRemoveSubscriber` per (removed-user × thread); the parent-group removal is already batched, so consider batching the per-thread subscriber removal too if large groups with many threads become common. Not a concern at typical group sizes.

## 3. Recommendations

- Merge as-is. The fix is correct and the design (single shared helper) is the right call.
- File follow-up issue(s) for: (1) `handleOrgEmployeeExit` thread cleanup, (2) `thread_setting` cleanup on the group path, (3) aligning/removing the stale `thread.Service.RemoveUserFromGroupThreads`.

## 4. Additional notes

- The intentional inclusion of archived threads is correct, but `UnarchiveThread` does not re-add IM subscriptions, so a member removed while a thread was archived would not be re-subscribed if the thread is later reactivated. This is a pre-existing reactivation-path gap, not introduced here — worth keeping in mind alongside the org-exit follow-up.
- Best-effort error handling (log-and-continue, cleanup outside the transaction) is a deliberate, documented choice that correctly prioritizes preventing subscription leaks over atomic deletion; agree with it.

lml2468 (dismissed):
APPROVED (head `19f8f9d84b33`) **with a CI lane to clear before merge** — Lint is currently red on this SHA. The code logic is sound and the regression test pins the right invariant; the lint fail looks like a mechanical issue (gofmt / golangci-lint nit on the new files) that should be a 5-minute fix. Approving the design + correctness; flagging the lint as a merge gate, not a CR.

## ✅ Root-cause + fix shape

Read the diff against the surrounding seam (`addUsersToGroupThreads`, `BotFather.processBotDelete`, the two old in-file duplicates). The PR description's root cause is exact:

```sql
-- OLD (both duplicates):
SELECT thread.short_id FROM thread
JOIN thread_member ON thread.id = thread_member.thread_id
WHERE thread.group_no=? AND thread_member.uid=? AND thread.status!=3
```

`JOIN thread_member` filters to threads the user has *actively joined*. Bots default to not calling `JoinThread`, so this returns zero rows → early-return → `IMRemoveSubscriber` never fires → bot stays subscribed at the WuKongIM layer → keeps receiving thread pushes after group removal.

The add-side mirror (`addUsersToGroupThreads`) subscribes to **every** thread `WHERE group_no=? AND status!=3`. The remove side must match — that's the asymmetry. The new helper does exactly this:

```sql
SELECT short_id FROM thread WHERE group_no=? AND status!=3
```

Symmetric with the add path. `status!=3` (excludes deleted, includes active+archived) is the right line because `UnarchiveThread` can reactivate `status=2` rows — leaving the subscription dangling on an archived thread that later gets unarchived would re-leak.

## ✅ Dedup discipline — single source of truth

Three pieces:

1. New `thread_cleanup.go` owns `queryThreadShortIDsForCleanup` + `removeUserFromGroupThreadsCleanup`.
2. `(*Group).removeUserFromGroupThreads` and `(*Service).removeUserFromGroupThreads` both collapse to thin wrappers — `removeUserFromGroupThreadsCleanup(ctx, log, groupNo, uid, spaceID)`.
3. `RemoveUserFromGroupThreads` exported on `group.IService` for the BotFather bot-delete path (`modules/botfather/command.go:665`), called right after `IMRemoveSubscriber` on the parent group channel.

Eliminates "fix one copy, miss the other" — which is exactly the failure mode that created this bug in the first place (two verbatim copies, same SQL, same flaw). The wrapper-method retention is the right call too — minimal blast radius for the call-site, and the IService addition is the only new public surface (which it had to be to reach the helper from the BotFather package).

## ✅ Best-effort semantics preserved correctly

- `thread_member DELETE` errors only log — don't gate the IM unsubscribe loop. Right call: a DB blip can't be allowed to leak the subscription, because the WuKongIM side is the actual privacy boundary (DB row is just bookkeeping).
- `IMRemoveSubscriber` errors per-thread only log — don't abort the loop. Right call: one IM call failing shouldn't strand the rest of the threads in subscribed state.
- `user.RemovePinnedForUserInSpace` + `conversation_ext.RemoveConvExtForUserInSpace` per thread — kept (the now-closed #28 had dropped these, which would have been a regression). Worth calling out that this PR explicitly preserves them.

Two failure logs `"删除子区成员记录失败"` and `"移除子区IM订阅者失败"` both carry `groupNo` / `uid` / `channelID` — sufficient ops context for postmortem.

## ✅ Test coverage on the right invariants

`thread_cleanup_test.go` — 6 cases, each pinning a distinct invariant:

- **`TestQueryThreadShortIDsForCleanup_ReturnsAllNonDeletedRegardlessOfMembership`** — the direct #27 reproducer. Seeds active+archived+deleted+cross-group threads, **zero `thread_member` rows**, asserts return is `[active, archived]`. If anyone ever re-introduces the JOIN, this goes red with a precise diagnostic.
- `_EmptyGroupNo` / `_NoThreads` / `_OnlyDeleted` — edge cases on the query.
- **`TestRemoveUserFromGroupThreadsCleanup_DeletesMemberWithoutOwnRow`** — covers the bot's path (no `thread_member` row → still does the cleanup).
- `_EmptyInputs` — guards against accidental nil-input bugs.

The integration tests the PR description mentions as still green (`TestGroupExit_CascadeBot_AlsoRemovesFromThread`, `TestRemoveGroupMembers_CascadeRemovesInvitedBots`, `TestDeleteBotCleansUpGroupMembers`) cover the end-to-end behaviour. Combined: the helper's invariants are pinned + the existing flows that call it still pass.

## ✅ Scope discipline

The PR explicitly defers two things that would have widened scope:

1. **`modules/robot/event.go` `robotMessageListen` group-message dispatch** — has no group-membership check; affects bots polling `/v1/bot/events`. Separate code path, separate issue. Right call not to bundle.
2. **End-to-end "stale subscriber actually receives a push" question** — requires a WuKongIM-attached environment to check `channel/subscribers` before/after. Linkage kept as `Refs #27` not `Closes #27` precisely because of this. Honest scoping; merge symptom-verification belongs to the human-driven WuKongIM check.

## 🔴 Merge gate — Lint is red on this SHA

`Lint` job on `19f8f9d84b33` is reporting `fail` (CI run 27067852033, job 79891756807). Build / Vet / CodeQL / osv / i18n all green; `code-review` (echo) APPROVED. Worth running `make lint` (or `golangci-lint run ./modules/group/...`) locally and fixing whatever it surfaces — most likely a gofmt nit on the new `thread_cleanup.go` / `thread_cleanup_test.go` or an unused-import on the trimmed-down `service.go`/`api.go`. Don't merge until lint clears.

I can't pull the lint log from GitHub right now (the API connection keeps EOF'ing), so I can't tell you the exact lint complaint. If you can grab it locally that's faster.

- **DELETE thread_member with subquery on the same table**: `DELETE FROM thread_member WHERE uid=? AND thread_id IN (SELECT id FROM thread WHERE group_no=?)`. MySQL allows this since the subquery is on a *different* table (`thread` vs `thread_member`), so no "can't specify target table" error. Just noting in case anyone refactors to use a JOIN form — that form would need explicit aliasing.
- **`out := make([]string, 0, len(rows))`**: nice — preallocates exact capacity, the right shape for "map row to scalar".
- **`logger.Error("查询群子区失败", zap.Error(err), zap.String("groupNo", groupNo), zap.String("uid", uid))`**: consistent structured logging. Good.
- **Channel-ID format `groupNo + "____" + shortID`** matches `thread.BuildChannelID` (verified mentally from the comment + project conventions). Worth a one-line sanity check that `thread.BuildChannelID` actually uses the same `____` separator if you want to be paranoid — a stray `_` count mismatch here would silently target the wrong channel.

## Verdict

Approval on the code + test discipline. The fix is minimal-surgical, the regression test pins the right invariant, dedup eliminates the structural cause of "fix one miss the other", and scope discipline keeps the WuKongIM-side verification honest. **Merge only after Lint clears** — that's the only outstanding gate.

Once #27 verification on the WuKongIM side lands, this PR's commit message / issue link can be promoted from `Refs` to `Closes`.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#300 Review Report
**PR:** Mininglamp-OSS/octo-server#300
**Head SHA:** `e6b32afaeed1043450ca4a55a5dcd26dd22cc85f`
**Scope:** fix(group): unsubscribe removed members from all non-deleted thread channels (#27)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Root cause correctly identified | ✅ | Old SQL `JOIN thread_member` returns 0 rows for members who never called `JoinThread` (bots default). Verified at old `api.go:3042-3089` / `service.go:1850-1893`. |
| Fix achieves add/remove symmetry | ✅ | New `queryThreadShortIDsForCleanup` uses `WHERE group_no=? AND status!=3` — identical predicate to `addUsersToGroupThreads` (`api.go:3063`). |
| Duplicated code consolidated | ✅ | Both `(*Group).removeUserFromGroupThreads` (`api.go:3045`) and `(*Service).removeUserFromGroupThreads` (`service.go:1855`) now delegate to `removeUserFromGroupThreadsCleanup` in `thread_cleanup.go:49`. |
| Botfather delete path covered | ✅ | `command.go:668` calls `h.groupService.RemoveUserFromGroupThreads(g.GroupNo, botID, g.SpaceID)` after parent-channel IMRemoveSubscriber. |
| Org-exit path covered (commit 2) | ✅ | `event.go:812` calls `g.removeUserFromGroupThreads(groupNo, req.Operator, spaceIDByGroupNo[groupNo])` after parent-channel cleanup. |
| channelID format correct | ✅ | `groupNo + "____" + shortID` matches `thread.BuildChannelID` (`thread/service.go:802-803`, separator `"____"` at `thread/const.go:17`). |
| `g.db.session` ≡ `ctx.DB()` | ✅ | `db.go:24`: `session: ctx.DB()`. New helper using `ctx.DB()` is equivalent. |
| Input guards present | ✅ | `thread_cleanup.go:50-52`: `groupNo == "" || uid == ""` → early return. `queryThreadShortIDsForCleanup` line 24: `groupNo == ""` → nil return. |
| Tests cover core reproducer | ✅ | `thread_cleanup_test.go:18-62`: no `thread_member` rows, asserts active+archived returned, deleted excluded. |
| Tests cover org-exit path | ✅ | `event_orgexit_thread_test.go`: end-to-end handler test, asserts `thread_member` cleaned. |

## 2. Findings

### No P0 or P1 issues found.

**P2-1: `OrgEmployeeExitReq.Operator` field comment is misleading (pre-existing)**
- **Diff-scope:** Pre-existing. The struct is in `octo-lib/config/msg_group.go` with comment `Operator string // 操作者uid` ("operator uid"), but the handler consistently uses it as the *leaving user's* uid (DeleteMemberTx, IMRemoveSubscriber, and now removeUserFromGroupThreads all use `req.Operator` as the target). This PR correctly follows the established usage pattern. Not introduced or amplified by this diff.
- **Suggestion:** Consider clarifying the field comment in octo-lib to `// 退出用户的uid` to match actual semantics. Separate issue, not blocking.

**P2-2: Best-effort cleanup ordering — IMRemoveSubscriber failure doesn't short-circuit pinned/ext cleanup**
- **Diff-scope:** Pre-existing behavior preserved. In `thread_cleanup.go:72-82`, if `IMRemoveSubscriber` fails for a thread, the loop continues to `RemovePinnedForUserInSpace` and `RemoveConvExtForUserInSpace` for that same thread. This is correct best-effort behavior — each cleanup is independent. Noted for completeness, not a defect.

**P2-3: No explicit test for botfather `onDeleteConfirm` → `RemoveUserFromGroupThreads` integration**
- **Diff-scope:** New code path (`command.go:668`). The shared helper is well-tested in `thread_cleanup_test.go`, and the botfather call is a straightforward delegation. The existing `TestDeleteBotCleansUpGroupMembers` test (mentioned in PR body) covers the broader bot-delete flow. Low risk, but a direct assertion on thread cleanup in the botfather test would strengthen coverage.

## 3. Data-Flow Trace

### Flow 1: `queryThreadShortIDsForCleanup` → `removeUserFromGroupThreadsCleanup` loop

| Data consumed | Upstream source | Reaches consumption? |
|---|---|---|
| `shortIDs` (slice of `short_id`) | SQL: `SELECT short_id FROM thread WHERE group_no=? AND status!=3` | ✅ Yes. No `JOIN thread_member` — returns all non-deleted threads regardless of membership. |
| `channelID` = `groupNo + "____" + shortID` | Constructed from `groupNo` param + each `shortID` from query | ✅ Yes. Format matches `thread.BuildChannelID`. |
| `uid` | Passed from caller (kick/exit/cascade/org-exit/bot-delete) | ✅ Yes. Used in DELETE WHERE and IMRemoveSubscriber. |
| `spaceID` | Passed from caller; in org-exit path sourced from `group.SpaceID` via `spaceIDByGroupNo` map | ✅ Yes. Map populated from same `QueryGroupsWithGroupNos` result, keyed by `groupNo`. |

### Flow 2: `handleOrgEmployeeExit` → `removeUserFromGroupThreads`

| Data consumed | Upstream source | Reaches consumption? |
|---|---|---|
| `req.Operator` | JSON payload field; despite "操作者" comment, consistently used as the leaving user across all handler operations | ✅ Yes. Same uid used for DeleteMemberTx, IMRemoveSubscriber, and now thread cleanup. |
| `spaceIDByGroupNo[groupNo]` | Populated at `event.go:746` from `group.SpaceID` (DB query result) | ✅ Yes. Only accessed for `realGroups` entries which are guaranteed to have a map entry. |
| `groupNo` | From `realGroups` slice, filtered from `req.GroupNos` ∩ DB results | ✅ Yes. |

### Flow 3: Botfather `onDeleteConfirm` → `RemoveUserFromGroupThreads`

| Data consumed | Upstream source | Reaches consumption? |
|---|---|---|
| `g.GroupNo` | From `GetGroupsWithMemberUID(botID)` result | ✅ Yes. |
| `botID` | The bot being deleted | ✅ Yes. |
| `g.SpaceID` | Same group query result | ✅ Yes. |

## 4. Distributed Side-Effects Checklist (R3)

| Check | Result |
|---|---|
| **Multi-client fan-out:** IMRemoveSubscriber is server-side idempotent (removing non-existent subscriber = no-op). No duplicate-push risk. | ✅ Pass |
| **Observer ≠ actor:** `uid` passed to cleanup is always the actual removed user, not the observer/operator. | ✅ Pass |
| **Scheduled/headless triggers:** Org-exit handler runs as event processing, not UI-dependent. Thread cleanup is called inline, not gated on any client state. | ✅ Pass |

## 5. Suggestions

1. **(Optional)** Add a botfather-level integration test that asserts `thread_member` cleanup after `onDeleteConfirm`. Low priority since the shared helper is well-tested.
2. **(Optional)** Clarify `OrgEmployeeExitReq.Operator` field comment in octo-lib to match actual semantics.

## 6. Conclusion

This is a well-structured, surgical fix that correctly addresses the root cause (asymmetric add/remove thread subscription paths). The consolidation of two duplicated implementations into a single shared helper eliminates the "fix one, miss the other" regression risk. The org-exit path addition in commit 2 closes a gap flagged during review. Test coverage directly reproduces the Issue #27 scenario. No blocking issues found.

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. Fix is correct, well-tested, and preserves existing best-effort error handling patterns.

Jerry-Xin (dismissed):
Project relevance gate: passes. This PR fixes octo-server group/thread subscription cleanup behavior and is in scope.

💬 Non-blocking

🟡 Warning: There is still a separate public cleanup path in [modules/thread/service.go](/Users/mlamp/.openclaw/tmp/pr-review-m1t4z06i/repo/modules/thread/service.go:1078) that queries threads through `thread_member` before removing IM subscribers. It is not used by the changed group/botfather paths, so it is not a current blocker, but its name/comment imply it is safe for leave-group cleanup. Consider aligning it with the new non-deleted-thread query or documenting that group removal must use the group cleanup helper.

🔵 Suggestion: [modules/group/thread_cleanup_test.go](/Users/mlamp/.openclaw/tmp/pr-review-m1t4z06i/repo/modules/group/thread_cleanup_test.go:117) proves the helper does not depend on the target having a `thread_member` row, but it cannot directly observe `IMRemoveSubscriber` calls. A future fake IM hook or narrower seam would make the core regression more directly testable.

✅ Highlights

The new query in [modules/group/thread_cleanup.go](/Users/mlamp/.openclaw/tmp/pr-review-m1t4z06i/repo/modules/group/thread_cleanup.go:29) matches the add path by selecting all non-deleted threads for the group, including archived threads.

The cleanup continues after a `thread_member` delete error and still attempts IM unsubscription in [modules/group/thread_cleanup.go](/Users/mlamp/.openclaw/tmp/pr-review-m1t4z06i/repo/modules/group/thread_cleanup.go:64), which is the right failure mode for preventing subscription leaks.

The bot deletion and org-exit paths are now covered: [modules/botfather/command.go](/Users/mlamp/.openclaw/tmp/pr-review-m1t4z06i/repo/modules/botfather/command.go:665), [modules/group/event.go](/Users/mlamp/.openclaw/tmp/pr-review-m1t4z06i/repo/modules/group/event.go:812).

Verification: compile-only checks passed for `./modules/group`, `./modules/botfather`, `./modules/message`, and `./modules/thread`. Full runtime tests could not run locally because MySQL on `127.0.0.1:3306` was unavailable.

yujiawei (dismissed):
# Code Review — PR #300 (octo-server)

**Verdict: APPROVED.** This is a correct, well-scoped fix with strong test coverage. No P0/P1 blocking issues. A couple of P2/maintainability notes are listed below but none should hold up the merge.

## 1. Verification of the fix

| Claim | Result | Evidence |
|---|---|---|
| Root cause is the `JOIN thread_member` gating in the remove path | ✅ | Old `api.go`/`service.go` both filtered `JOIN thread_member ... WHERE thread_member.uid=?`, returning 0 rows for members who never `JoinThread` (bots), so the loop `return`ed before `IMRemoveSubscriber`. |
| New helper removes the JOIN and lists by parent group | ✅ | `modules/group/thread_cleanup.go:33` — `SELECT short_id FROM thread WHERE group_no=? AND status!=3`. |
| Symmetry with the add path | ✅ | `addUsersToGroupThreads` (`api.go:3050`) subscribes on `WHERE group_no=? AND status!=3`; the remove path now mirrors it exactly. Archived (`status=2`) included (reactivatable via `UnarchiveThread`), deleted (`status=3`) excluded. |
| Both duplicated copies now delegate to one source of truth | ✅ | `(*Group).removeUserFromGroupThreads` → `api.go:3046`, `(*Service).removeUserFromGroupThreads` → `service.go:1856`, both call `removeUserFromGroupThreadsCleanup`. The "fix one copy, miss the other" risk is eliminated. |
| `DELETE thread_member` no longer gates the IM-unsub loop | ✅ | `thread_cleanup.go:67` runs the delete best-effort and continues to `IMRemoveSubscriber` even on DB error — strictly better than the old early-return on delete failure. |
| BotFather delete-bot path is aligned | ✅ | `modules/botfather/command.go:665` calls `h.groupService.RemoveUserFromGroupThreads(g.GroupNo, botID, g.SpaceID)` after the parent-channel unsub. `IService.RemoveUserFromGroupThreads` exported at `service.go:112`, `g.SpaceID` is populated via `toInfoResp` (`service.go` line `SpaceID: m.SpaceID`). |
| Org-exit path (`handleOrgEmployeeExit`) now also cleans threads | ✅ | `event.go:812` uses `req.Operator` (the leaving uid — same value used for the parent-channel `IMRemoveSubscriber` immediately above) and `spaceIDByGroupNo[groupNo]` captured at `event.go:744`. Correct uid and space. |
| Builds and vets clean | ✅ | `go build ./...` and `go vet ./modules/group/... ./modules/botfather/...` both exit 0 at head SHA `e6b32af`. |
| `conversation_ext` cleanup preserved | ✅ | Retained in the shared helper (`thread_cleanup.go:81`); avoids the regression that the earlier closed PR #28 introduced. |

Tests are thorough and target the actual failure mode: the headline `TestQueryThreadShortIDsForCleanup_ReturnsAllNonDeletedRegardlessOfMembership` deliberately inserts **no** `thread_member` rows (the exact bot scenario), plus archived-inclusion, deleted-exclusion, cross-group isolation, no-own-row delete, empty-input guards, and the org-exit end-to-end regression.

> Note on local test execution: I could not run the suite to green locally because the test DB hit a migration-state mismatch (`unknown migration in database: 20210926000001_robot_legacy01.sql`) in `testutil.NewTestServer`. This is an environment issue, not a code defect — `go build` and `go vet` are clean, and the test code compiles. The PR documents a passing run on the CI-equivalent stack.

## 2. Findings

No P0 or P1.

**P2 — a third copy of this logic still carries the original bug (latent).**
`modules/thread/service.go:1078` `(*Service).RemoveUserFromGroupThreads(groupNo, uid)` still lists threads via `QueryThreadsByGroupNoAndUID`, which does `JOIN thread_member ... WHERE tm.uid=?` (`thread` DB layer) — i.e. the exact Issue #27 pattern. It is currently **dead in production** (only referenced by tests; the BotFather call site uses the *group* module's 3-arg version, not this one), so it does not affect this fix. But it is exported on `thread.IService`, so if a future caller wires it on a removal path the bug reappears. Worth either consolidating onto the same group-level criteria or filing a follow-up to remove/fix it. Not blocking.

**P2 — early return skips `thread_member` delete when a group has only deleted threads.**
`removeUserFromGroupThreadsCleanup` returns when `len(shortIDs) == 0`. Since `queryThreadShortIDsForCleanup` excludes `status=3`, a stale `thread_member` row pointing at an already-deleted thread will not be cleaned up by this path. Impact is negligible — the deleted thread's IM channel is already destroyed, so the orphan row leaks no subscription — but if strict row hygiene is desired the `DELETE thread_member` could run unconditionally before the short-circuit. Cosmetic.

## 3. Suggestions

- Consider folding the `thread`-module `RemoveUserFromGroupThreads(groupNo, uid)` into the same membership-independent query (or deleting it if truly unused) so all three historical copies converge — this PR fixed two of three.
- The out-of-scope items the author already flagged (`robotMessageListen` group-message dispatch lacking a membership check in `modules/robot/event.go`, and the end-to-end "stale subscriber still receives a push" WuKongIM verification) are correctly deferred; a separate issue for the `robotMessageListen` gap is warranted since it is a parallel leak vector on a different route.

## 4. Additional observations

- The `Refs #27` (not `Closes`) linkage is the right call given the `needs-human-verify` server-side-push question remains open; the code-level asymmetry this PR closes is independently verifiable and correct.
- Behavior change is genuinely surgical: only thread-channel cleanup is widened to match the add path; parent-channel and cascade flows are untouched.

Approving — the fix is correct, symmetric with the subscription path, de-duplicated, and well-tested.

an9xyz:
## WuKongIM-side end-to-end verification (the `needs-human-verify` item) — done ✅

Ran against a live WuKongIM `v2.2.4-20260313` (CI-pinned tag), thread channel `channel_type=5` (`CommunityTopic`):

| Step | bot (no `thread_member` row — the #27 scenario) | control user (still subscribed) |
|---|---|---|
| send msg #1 (both subscribed) | ✅ receives (`unread=1`) | ✅ receives |
| run `removeUserFromGroupThreadsCleanup`, send msg #2 | ❌ **stops receiving** (conversation → `[]`) | ✅ still receives (`unread=2`) |

Observed via `/conversation/sync` (v2.2.4 exposes no `GET /channel/subscribers`; message-delivery is the more direct signal and is the actual reported symptom). This drives the **real fixed helper**, not a mock — so it closes the loop @lml2468 flagged for promoting `Refs #27` → `Closes #27`.

### New artifact
`modules/group/e2e_issue27_wukong_test.go`, isolated behind `//go:build wukong_e2e` so it stays out of normal CI (WuKongIM message-delivery semantics are version-coupled — addresses the brittleness concern from review):

```
go test -tags wukong_e2e ./modules/group/ -run TestE2E_Issue27 -v   # PASS (6.2s)
```

### Also folded in since last review
- `handleOrgEmployeeExit` (org/department-driven exit) now calls the shared cleanup too — the same-bug-class gap flagged by @yujiawei / @mochashanyao. Covered by `TestHandleOrgEmployeeExit_AlsoCleansThreads`.

### Still-open, non-blocking follow-ups (recommend separate issues, as reviewers noted)
- `thread.Service.RemoveUserFromGroupThreads` (`thread/service.go:1078`) still carries the `JOIN thread_member` scope — unreferenced by production, latent trap.
- `thread_setting` is not cleaned on the group-removal path (pre-existing).
- `OrgEmployeeExitReq.Operator` field comment in octo-lib is misleading (semantically it's the leaving user).

Note: the red `dependency-review` check is a repo-config issue (Dependency graph not enabled for the repo), unrelated to this PR — it fails on every push regardless of diff.

---

Jerry-Xin (approved):
This PR is relevant to octo-server and fixes a real group/thread subscription asymmetry with a scoped shared cleanup helper.

💬 Non-blocking
- 🟡 Warning: [modules/group/e2e_issue27_wukong_test.go](/Users/mlamp/.openclaw/tmp/pr-review-d_cve4mf/repo/modules/group/e2e_issue27_wukong_test.go:42) is not `gofmt` clean. `gofmt -l` reports this file only; the diff is just comment alignment in the const block.
- 🔵 Suggestion: [modules/group/thread_cleanup.go](/Users/mlamp/.openclaw/tmp/pr-review-d_cve4mf/repo/modules/group/thread_cleanup.go:70) now does one IM unsubscribe plus pinned/conversation cleanup per non-deleted thread. That matches the existing add path, but for very large groups this can be expensive during bot deletion or org-exit cleanup. Consider a follow-up if WuKongIM or local cleanup APIs support batching by channel list.

✅ Highlights
- The relevance gate passes: the PR directly fixes group/thread membership cleanup behavior in this repository.
- The new query correctly mirrors `addUsersToGroupThreads` by selecting all non-deleted group threads instead of only joined `thread_member` rows.
- Consolidating the duplicated group/API and service cleanup logic reduces regression risk.
- The added coverage hits the core no-`thread_member` bot scenario, archived inclusion, deleted exclusion, and org-exit parity.

Verification note: I attempted `go test ./modules/group ./modules/botfather`, but this local environment has no MySQL on `127.0.0.1:3306`, so those packages failed at test setup. `go test ./modules/message -run TestRevoke -count=1` passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#300 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/300
**Head SHA:** `b941b6423e33578b07a8c13877e8199063137b53`
**Title:** fix(group): unsubscribe removed members from all non-deleted thread channels (#27)
**Files:** 8 changed (+439 / −83)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Root cause correctly identified | ✅ | Old `JOIN thread_member` query returns 0 rows for users who never called `JoinThread` (bots default). `addUsersToGroupThreads` subscribes ALL threads `WHERE status!=3`; remove path was asymmetric. |
| New query drops JOIN | ✅ | `thread_cleanup.go:33-38` — `SELECT short_id FROM thread WHERE group_no=? AND status!=3` — no thread_member dependency |
| Symmetry with add path | ✅ | `addUsersToGroupThreads` (service.go:1876, api.go:3062) uses identical `WHERE group_no=? AND status!=3` |
| Duplicates consolidated | ✅ | Both `(*Group).removeUserFromGroupThreads` (api.go:3045) and `(*Service).removeUserFromGroupThreads` (service.go:1855) delegate to shared `removeUserFromGroupThreadsCleanup` |
| All 4 removal paths covered | ✅ | kick (service.go:1697), self-exit (api.go:2973), org-exit (event.go:812 NEW), bot-delete (botfather/command.go:665 NEW) |
| Cascade bot cleanup | ✅ | Self-exit cascade: api.go:3030 calls `removeUserFromGroupThreads` per cascaded bot |
| Archived threads included | ✅ | `status!=3` keeps status=2 (archived); correct since `UnarchiveThread` can reactivate |
| Deleted threads excluded | ✅ | `status!=3` excludes deleted (IM channel destroyed) |
| Best-effort semantics preserved | ✅ | thread_member DELETE failure logs but doesn't abort IM cleanup (thread_cleanup.go:68-71) |
| Empty-input guards | ✅ | `groupNo=="" || uid==""` → early return (thread_cleanup.go:51-53) |
| spaceIDByGroupNo map correctness | ✅ | Populated from same `groups` query result; `realGroups` only contains matched groupNos → map lookup always hits |
| Tests cover #27 reproducer | ✅ | `TestQueryThreadShortIDsForCleanup_ReturnsAllNonDeletedRegardlessOfMembership` — no thread_member rows, asserts all non-deleted threads returned |
| E2E test isolation | ✅ | Behind `wukong_e2e` build tag — excluded from normal CI |

## 2. Findings

### F1 — P2: `modules/thread/service.go` has the same JOIN bug (pre-existing, not fixed)

**Diff-scope:** pre-existing — this PR does not touch `modules/thread/`.

`thread.Service.RemoveUserFromGroupThreads` (thread/service.go:1078) calls `QueryThreadsByGroupNoAndUID` (thread/db.go:524-532) which uses the same `INNER JOIN thread_member` pattern:

```go
// thread/db.go:528
Join(dbr.I("thread_member").As("tm"), "t.id = tm.thread_id").
Where("t.group_no=? AND tm.uid=? AND t.status!=?", groupNo, uid, ThreadStatusDeleted)
```

This has the identical Issue #27 bug: returns empty for users without `thread_member` rows → IM unsubscription skipped.

**Mitigating factor:** Grep confirms this method has **zero callers** in the current codebase (the interface declares it at thread/service.go:66 but nothing invokes it). It's dead code.

**Recommendation:** Fix or remove the dead method for completeness. If any future caller wires into it, the bug will silently resurface.

### F2 — P2: `thread_setting` not cleaned by group module's helper (pre-existing)

**Diff-scope:** pre-existing — the old code in both api.go and service.go also didn't clean `thread_setting`.

The new `removeUserFromGroupThreadsCleanup` cleans: thread_member, IM subscription, pinned, conversation_ext. It does NOT clean `thread_setting` (mute/notification preferences). The thread module's parallel implementation (unused, see F1) does clean `thread_setting` via `DeleteSettingsByGroupNoAndUIDTx`.

**Impact:** Low — `thread_setting` rows for removed users are orphaned data but don't cause functional issues (no code reads settings for non-members).

### F3 — Nit: E2E test uses `time.Sleep(800ms)` for WuKongIM sync

**Diff-scope:** new code (e2e_issue27_wukong_test.go:78,88).

The `hasChannel` check relies on `time.Sleep(800ms)` after `SendMessage`. This is inherently racy under load. Acceptable for an opt-in e2e test behind `wukong_e2e` build tag, but consider a polling-with-timeout pattern if the test becomes flaky.

## 3. Data-Flow Trace

### Core fix data flow (verified end-to-end):

```
Caller (kick/exit/orgexit/botdelete)
  → removeUserFromGroupThreads(groupNo, uid, spaceID)
    → removeUserFromGroupThreadsCleanup(ctx, logger, groupNo, uid, spaceID)
      → queryThreadShortIDsForCleanup(ctx, groupNo)
        → SQL: SELECT short_id FROM thread WHERE group_no=? AND status!=3
        → Returns ALL non-deleted thread short_ids (no thread_member JOIN)
      → DELETE FROM thread_member WHERE uid=? AND thread_id IN (SELECT id FROM thread WHERE group_no=?)
        → best-effort, logs on failure, continues regardless
      → FOR each shortID:
          channelID = groupNo + "____" + shortID
          → IMRemoveSubscriber(channelID, CommunityTopic, [uid])
          → RemovePinnedForUserInSpace(uid, spaceID, channelID, CommunityTopic)
          → RemoveConvExtForUserInSpace(uid, spaceID, channelID, CommunityTopic)
```

### Upstream data source verification:

| Consumed data | Source | Verified? |
|---|---|---|
| `shortIDs` | `thread` table `WHERE group_no=? AND status!=3` | ✅ Matches `addUsersToGroupThreads` query |
| `spaceID` (kick) | `groupModel.SpaceID` from `QueryWithGroupNo` | ✅ |
| `spaceID` (self-exit) | `groupInfo.SpaceID` from `getGroupInfo` | ✅ |
| `spaceID` (org-exit) | `spaceIDByGroupNo[groupNo]` from `QueryGroupsWithGroupNos` | ✅ Same query populates both `realGroups` and map |
| `spaceID` (bot-delete) | `g.SpaceID` from `GetGroupsWithMemberUID` | ✅ |
| `channelID` format | `groupNo + "____" + shortID` | ✅ Matches `thread.BuildChannelID` and `addUsersToGroupThreads` |
| `ChannelType` | `common.ChannelTypeCommunityTopic.Uint8()` = 5 | ✅ Consistent across add/remove paths |

### Distributed side-effects checklist (R3):

- **Multi-client fan-out:** `IMRemoveSubscriber` is a server-side WuKongIM API call — removes the subscription at the IM server, so all WebSocket connections for that uid stop receiving. No client-side dedup issue. ✅
- **Observer ≠ actor:** Not applicable — the cleanup uses the target `uid` for unsubscription, not the operator. ✅
- **Scheduled/headless triggers:** All 4 removal paths (kick, self-exit, org-exit, bot-delete) are server-side handlers, not dependent on any client observing state. ✅

## 4. Additional Observations

- **Code quality:** Excellent consolidation — two verbatim-duplicated implementations merged into one shared helper with proper logger injection. The interface export (`RemoveUserFromGroupThreads` on `group.IService`) is clean.
- **Test coverage:** 7 new unit tests + 1 opt-in e2e. Tests directly reproduce the #27 scenario (no thread_member row) and verify cross-group isolation, empty-input guards, and archived/deleted handling.
- **PR body quality:** The PR description is thorough — includes root cause, verification evidence, and explicitly notes out-of-scope items (robot/event.go group-message dispatch).

## 5. Verdict

No P0 or P1 findings. The fix correctly addresses the root cause (JOIN asymmetry), consolidates duplicated code, covers all 4 member-removal paths, and is well-tested. Two P2 observations are pre-existing issues outside this PR's diff scope.

**[Octo-Q] verdict: APPROVE** — no blocking issues found. F1 (dead `thread.Service.RemoveUserFromGroupThreads` with same JOIN bug) recommended as follow-up cleanup.

yujiawei (approved):
# Code Review — PR #300 (octo-server)

**Verdict: APPROVED.** The fix is correct, well-scoped, and properly tested. No P0/P1 blockers. The findings below are non-blocking maintainability/follow-up notes.

## Summary

The change fixes a real subscription-leak bug: removed members (typically bots that never call `JoinThread`) kept their WuKongIM subscriptions to thread (子区) channels because `removeUserFromGroupThreads` listed threads via `JOIN thread_member`, which returns zero rows for a never-joined member and short-circuited before `IMRemoveSubscriber`. The new `queryThreadShortIDsForCleanup` drops the `thread_member` JOIN and selects all non-deleted threads (`WHERE group_no=? AND status!=3`), making removal symmetric with `addUsersToGroupThreads`, which already subscribes every group member to all non-deleted threads with the same predicate.

## Verification

| Item | Result | Evidence |
|---|---|---|
| Leak fixed symmetrically with the add path | ✅ | `thread_cleanup.go:31` `WHERE group_no=? AND status!=3` is identical to `addUsersToGroupThreads` (`api.go` / `service.go`); both ungated by `thread_member`. |
| `status` handling (active/archived included, deleted excluded) | ✅ | `thread_cleanup.go:31`; archived (status=2) reactivatable via `UnarchiveThread`, deleted (status=3) channel already destroyed. |
| `channelID` format matches `thread.BuildChannelID` | ✅ | `thread_cleanup.go` builds `groupNo + "____" + shortID`. |
| De-dup of the two verbatim-duplicate impls | ✅ | `(*Group)` and `(*Service).removeUserFromGroupThreads` now delegate to the shared helper (`api.go`, `service.go:1855-1861`); the two old bodies were structurally identical. |
| Org-exit path aligned | ✅ | `event.go:812` calls `removeUserFromGroupThreads(groupNo, req.Operator, spaceIDByGroupNo[groupNo])`; `spaceID` is correctly captured per-group. |
| Bot-delete path aligned, valid `SpaceID` passed | ✅ | `botfather/command.go:665` uses `g.SpaceID` from `GetGroupsWithMemberUID` → `toInfoResp` (`InfoResp.SpaceID` is populated). |
| DELETE scoped by `uid`, no cross-user impact | ✅ | `thread_cleanup.go` `Where("uid=? AND thread_id IN (SELECT id FROM thread WHERE group_no=?)")`; covered by `TestRemoveUserFromGroupThreadsCleanup_DeletesMemberWithoutOwnRow`. |
| Error handling: continue-on-DELETE-error is an improvement | ✅ | Old code returned early on DELETE failure, leaking subscriptions; new helper logs and proceeds to `IMRemoveSubscriber`. |
| `go build` / `go vet` on `modules/group`, `modules/botfather` | ✅ | both clean. |
| Tests cover the #27 reproducer, archived/deleted, isolation, no-own-row, org-exit | ✅ | `thread_cleanup_test.go`, `event_orgexit_thread_test.go`. |

## Findings (all non-blocking)

### P2 — A second `RemoveUserFromGroupThreads` in the thread module still carries the same JOIN bug
`modules/thread/service.go:1078` (`(*Service).RemoveUserFromGroupThreads`) still lists threads via `QueryThreadsByGroupNoAndUID` (`modules/thread/db.go:524`), which JOINs `thread_member` on `tm.uid=?` — the exact membership-gated pattern this PR removes from the group module. A bot with no `thread_member` row would still leak there.

This is **not a regression and not on the path #300 fixes**: the bot-delete path (`command.go:665`) routes through `group.IService` (the fixed helper), and the thread-module method has no production caller on the kick/exit path (grep shows only test callers). But it is the same bug class one module over, and the PR's own comment states the goal was to end the "fix one, miss one" duplication. Recommend a follow-up issue to align it with `queryThreadShortIDsForCleanup` or delete it if truly dead.

### P2 — `thread_setting` is not cleaned on the group-exit/bot-delete paths
The group helper cleans `thread_member` + IM subscription + pinned + `conversation_ext`, but **not** `thread_setting`. The thread-module impl (`thread/service.go`) is the only place that deletes `thread_setting` (via `DeleteSettingsByGroupNoAndUIDTx`), and it is unreachable from these paths. Net effect: a removed bot/member that had a per-thread mute setting leaves an orphaned `thread_setting` row.

Impact is bounded — the security-relevant IM unsubscribe **is** performed, so this is stale mute/setting data, not re-subscription or message leakage; and `thread_setting` rows only exist if the user explicitly set a per-thread mute (rare for bots). Note this gap is **pre-existing** for the group kick/exit paths (it was never cleaned on `main`); the only newly-affected path is bot-delete. Cleanest fix: fold `thread_setting` deletion into `removeUserFromGroupThreadsCleanup` so the group helper becomes a true superset.

### Nit — DELETE (all threads) vs IM-unsubscribe (non-deleted only) asymmetry is intentional
The `DELETE thread_member` subquery has no `status` filter (purges rows even under deleted threads) while the unsubscribe loop iterates only `status!=3`. This is correct — deleted threads' IM channels are already destroyed — but the differing thread sets may read as an oversight; a one-line comment would help future readers.

### Nit — Per-thread `IMRemoveSubscriber` call volume increases for thread-heavy groups
Removal now issues one synchronous WuKongIM POST per non-deleted thread per removed user (and in `RemoveGroupMembers`, per removed user in a loop). This mirrors the existing `addUsersToGroupThreads` pattern and is not introduced by this PR, but for groups with hundreds of threads / large batch kicks it adds latency on the removal request path. A batched WuKongIM endpoint would be the long-term improvement; fine as-is.

## Tests
New unit tests assert the right regression behavior and existing cascade tests (`TestGroupExit_CascadeBot_AlsoRemovesFromThread`, etc.) remain valid — the widened query is a strict superset, so tests that seed an actual `thread_member` row still pass. The opt-in `wukong_e2e` build tag for the live-IM test is appropriate (keeps CI light). Note these tests require the documented MySQL+Redis stack to execute locally.

## Conclusion
Surgical, correct fix that restores add/remove symmetry and consolidates duplicated logic. Approving. The two P2 items (thread-module twin bug, `thread_setting` cleanup) are worth a follow-up issue but do not block merge.

yujiawei (approved):
# Code Review — PR #300 (octo-server)

**Verdict: APPROVED.** This is a correct, well-scoped fix for a real defect, with strong test coverage. No P0/P1 blockers. All findings below are non-blocking (P2/nit) maintainability or follow-up notes.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| Root cause is real (remove path asymmetric with add path) | ✅ | Old `removeUserFromGroupThreads` listed via `JOIN thread_member` (zero rows for bots that never `JoinThread`), so `IMRemoveSubscriber` never ran. `addUsersToGroupThreads` (`modules/group/api.go:3050`) subscribes every thread `WHERE group_no=? AND status!=3`. The fix makes removal symmetric. |
| Cleanup query correctness | ✅ | `queryThreadShortIDsForCleanup` (`modules/group/thread_cleanup.go:29-32`) selects `short_id WHERE group_no=? AND status!=3` — includes active(1)+archived(2), excludes deleted(3). Archived inclusion is correct: archived threads can be reactivated via `UnarchiveThread`, so their subscriptions must be torn down too. |
| `channelID` format matches subscription side | ✅ | `groupNo + "____" + shortID` in `thread_cleanup.go:72` is byte-identical to the add path and to `thread.BuildChannelID`. |
| DELETE is correctly scoped | ✅ | `thread_cleanup.go:64-66` deletes `WHERE uid=? AND thread_id IN (SELECT id FROM thread WHERE group_no=?)` — confined to the target uid and the target group; no cross-user / cross-group deletion. |
| De-duplication of the two copies | ✅ | Both `(*Group)` and `(*Service)` `removeUserFromGroupThreads` now delegate to the single `removeUserFromGroupThreadsCleanup`, so the bug cannot regress in only one copy. |
| BotFather wiring compiles & is correct | ✅ | `IService` gains `RemoveUserFromGroupThreads`; `*Service` implements it; `modules/botfather/command.go:665` calls it via `group.IService` after the parent-channel unsubscribe. `GetGroupsWithMemberUID` returns `InfoResp` which carries `SpaceID`. `go build ./modules/group/... ./modules/botfather/...` and `go vet` pass. |
| Org-exit path correctness | ✅ | `handleOrgEmployeeExit` (`modules/group/event.go`) populates `spaceIDByGroupNo` for every `realGroups` entry and passes the removed user (`req.Operator`, consistent throughout the handler) to the helper after the parent-channel unsubscribe. |
| `conversation_ext` cleanup retained | ✅ | `RemoveConvExtForUserInSpace` is preserved in the shared helper (the earlier closed PR #28 had dropped it). |
| CI at this head | ✅ | Build / Lint / Vet all green on `6f8b4f3e`. (A prior reviewer's lint concern was against an older SHA and no longer applies.) |

**P2 — `thread_setting` rows are not cleaned up.** The shared helper removes `thread_member` + IM subscription + pinned + `conversation_ext`, but never deletes `thread_setting` (per-thread mute/notification prefs). The thread module's own `RemoveUserFromGroupThreads` (`modules/thread/service.go`, via `DeleteSettingsByGroupNoAndUIDTx` at `modules/thread/db.go:543`) does delete it. This is a **pre-existing** asymmetry, not introduced by this PR (the old `(*Group).removeUserFromGroupThreads` had the same gap). Impact is low: orphaned `thread_setting` rows, and a stale mute could re-apply if the user rejoins. Worth a follow-up to fold `thread_setting` cleanup into the shared helper for full consistency.

**P2 — `IMRemoveSubscriber` loop is synchronous with no client timeout, and now iterates all non-deleted threads.** Each call is an inline HTTP POST to WuKongIM; the underlying sendgrid `rest.DefaultClient` has no `Timeout`. The fix widens the per-group iteration set from "threads the user joined" to "all non-deleted threads", so a hung WuKongIM can block the caller proportionally longer. This is acceptable for this PR: the no-timeout client is pre-existing (the adjacent parent-channel `IMRemoveSubscriber` already blocks synchronously per group), failures are logged and non-fatal, and the org-exit path runs under a goroutine+recover. Suggestion (follow-up): the BotFather delete path (`command.go:653-666`) runs cleanup inline per group without a goroutine/recover wrapper — consider aligning it with the org-exit pattern, and/or setting a client timeout in octo-lib.

**nit — best-effort DELETE leaves a behavior change on record.** Old code returned on DELETE error before unsubscribing; new code logs and continues to `IMRemoveSubscriber`. On DELETE failure this can leave a stale `thread_member` row while the subscription is correctly removed. This is strictly preferable for #27 (the leaked subscription was the real harm; a dangling row is cosmetic and re-cleanable, and cannot resurrect a subscription since the add path doesn't join `thread_member`). No change requested — noting the intentional trade-off.

## 3. Test coverage notes (non-blocking)

The `thread_cleanup_test.go` query-level test (`TestQueryThreadShortIDsForCleanup_ReturnsAllNonDeletedRegardlessOfMembership`) genuinely pins the #27 invariant (zero `thread_member` rows, asserts active+archived returned) — it would fail against the old JOIN-gated implementation. Good. A few gaps for follow-up:

- **No CI test asserts `IMRemoveSubscriber` actually fired.** `IMRemoveSubscriber` is a concrete method with no mock hook, so unit tests assert only `thread_member` DB rows as a proxy. The real unsubscribe surface is covered only by `e2e_issue27_wukong_test.go`, which is gated behind `//go:build wukong_e2e` and excluded from normal CI. The DB-row proxy is a reasonable stand-in, but the load-bearing invariant (the leak) isn't checked in CI.
- **`TestRemoveUserFromGroupThreadsCleanup_DeletesMemberWithoutOwnRow`** asserts only that another user's row survives — that assertion passes against both old and new implementations, so despite its name it doesn't pin the #27 fix (the query-level test above does). Consider renaming or strengthening.
- **BotFather bot-delete path has no direct test.** `onDeleteConfirm` (the primary real-world #27 trigger for bots) is invoked by no test; the existing delete-bot tests `t.Skip(...)`. The shared helper is well covered, but the integration wiring is not.
- **nit:** no test seeds the same uid in a second group to confirm cross-group DELETE isolation, and no end-to-end "unsubscribe → unarchive → no re-leak" test exercises the documented archived-inclusion rationale.

## 4. Additional notes

- The PR correctly keeps the linkage as **Refs #27** rather than **Closes**: the code-level asymmetry is fixed and verified, but the end-to-end "stale subscriber actually receives a push" question depends on WuKongIM server-side membership gating and needs a live-environment check. Reasonable.
- Out-of-scope item flagged in the description (`modules/robot/event.go robotMessageListen` has no group-membership check for `/v1/bot/events` pollers) is a separate route and correctly deferred to its own issue.

Overall: surgical, symmetric, de-duplicated fix that resolves the real leak. Approving; the items above are follow-up quality/coverage improvements, none blocking merge.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#300 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/300
**Head SHA:** `6f8b4f3e2622b401e3c63f8e6f4e720f44ef5dfa`
**Title:** fix(group): unsubscribe removed members from all non-deleted thread channels (#27)
**Files changed:** 8 (+439 / -83)

---

## 1. 验证结论

| 验证项 | 结论 | 证据 |
|--------|------|------|
| Add/remove 路径对称性 | ✅ | `addUsersToGroupThreads` (api.go:3050, service.go:1864) 用 `WHERE group_no=? AND status!=3` 无 JOIN；新 `queryThreadShortIDsForCleanup` (thread_cleanup.go:33) 同款查询 |
| DB session 等价性 | ✅ | `g.db.session` = `ctx.DB()` 在 `NewDB()` 时捕获 (db.go:25)；`ctx.DB()` 由 `sync.Once` 保护 (octo-lib context.go:82-86)，同一指针 |
| ChannelID 格式一致性 | ✅ | `thread.BuildChannelID` (thread/service.go:801) = `{groupNo}____{shortID}`，separator 常量 `____` (thread/const.go:17)；helper 内联拼接一致 |
| 所有移除路径覆盖 | ✅ | 6 个生产调用点全部委托到 `removeUserFromGroupThreadsCleanup`：quitGroup (api.go:2973), cascade bot (api.go:3030), kick (service.go:1697), org-exit (event.go:812), botfather delete (command.go:665) |
| `req.Operator` 语义正确 | ✅ | event.go:722 验证消息 "退出用户uid不能为空"；event.go:787 `DeleteMemberTx(groupNo, req.Operator, ...)` 删除的是退出者本人 |
| `spaceIDByGroupNo` 数据流 | ✅ | event.go:747 从 `group.SpaceID` (DB SELECT * FROM group) 填充；event.go:812 消费；与 `realGroups` 同循环构建，不可能 miss |
| 排序一致性 (post-commit) | ✅ | 所有 3 条路径 (groupExit/RemoveGroupMembers/handleOrgEmployeeExit) 均在 tx.Commit() 后执行 cleanup，与既有模式一致 |
| 测试覆盖 | ✅ | 7 个新单测覆盖核心回归 (#27 reproducer)、archived/deleted 边界、空输入守卫、跨群隔离、无 thread_member 行场景；1 个 e2e (build-tag 隔离) |

## 2. 发现的问题

### P2-1: 大规模群的 IM 调用量线性增长（性能）

**Diff-scope 三问：** new（新引入的行为，但属于 bug fix 的必要代价）

旧实现查询 `JOIN thread_member` 只返回用户有 thread_member 行的子区（通常很少）；新实现查询所有非删除子区，对每个都调用 `IMRemoveSubscriber`。对于拥有数百个子区的大群，即使被移除用户只活跃于 2-3 个子区，仍会产生数百次 IM 调用。

**严重度：** P2（性能考量，非功能缺陷）。add 路径已有同样的 N 次调用模式，remove 对称是正确的。WuKongIM 对非订阅者的 `IMRemoveSubscriber` 应为 no-op。best-effort 语义下失败只记日志。

**建议：** 可接受现状。如未来大群性能成为瓶颈，可考虑批量 unsubscribe API 或在 WuKongIM 侧做幂等优化。

### P2-2: `modules/thread/service.go` 的 `RemoveUserFromGroupThreads` 仍用旧 JOIN 模式（潜在陷阱）

**Diff-scope 三问：** pre-existing（本 PR 未触碰 thread 模块）

`modules/thread/service.go:1078` 存在一个独立的 `RemoveUserFromGroupThreads`，内部调用 `QueryThreadsByGroupNoAndUID` (thread/db.go:524) 仍使用 `JOIN thread_member WHERE tm.uid=?`——与本 PR 修复的 bug 同款模式。当前仅被测试调用 (service_test.go:636,656,771; service_setting_test.go:186,212)，无生产流量。但如果未来有代码引用该路径，会重现 #27 bug。

**严重度：** P2（pre-existing，无生产影响，但属 latent trap）

**建议：** 在 thread 模块的该函数上加注释标注 "此实现依赖 thread_member JOIN，不适用于从未 JoinThread 的成员（见 Issue #27）"，或后续统一。

### P2-3: E2E 测试的 timing 假设

**Diff-scope 三问：** new

`e2e_issue27_wukong_test.go` 使用 `time.Sleep(800ms)` 等待 WuKongIM 消息投递。在负载高的 CI 环境可能不够。已通过 `wukong_e2e` build tag 隔离出常规 CI，风险可控。

**严重度：** P2（测试稳定性，非功能缺陷）

## 3. 建议

1. **P2-2 的 thread 模块 latent trap**：建议加一行注释即可，不需要本 PR 修。
2. 其余改动设计合理、测试充分、文档清晰，无需额外修改。

## 4. 额外发现

- **octo-lib `OrgEmployeeExitReq.Operator` 命名误导**：struct 注释写 `"操作者uid"`（暗示 admin），但实际语义是退出员工本人。这是 octo-lib 的文档 bug，不影响本 PR 正确性，但值得在 octo-lib 侧修正。
- **Post-commit race window**：所有移除路径在 `tx.Commit()` 和 `removeUserFromGroupThreads` 之间存在短暂窗口，用户已从 group_member 删除但仍订阅 thread channels。这是既有设计模式（避免 IM 失败回滚 DB），本 PR 未改变此行为。理论上并发 re-add 可交错，但生产概率极低。

## 5. 数据流回溯

| 消费点 | 数据来源 | 回溯路径 | 是否真的流到 |
|--------|----------|----------|-------------|
| `removeUserFromGroupThreadsCleanup(ctx, ...)` 的 `ctx` | 调用方传入 | Group 路径: `g.ctx` (api.go:3046) → `New()` 时赋值 (api.go:58)；Service 路径: `s.ctx` (service.go:1856) | ✅ 同一 `*config.Context`，`DB()` 由 `sync.Once` 保护 |
| `logger` 参数 | 调用方传入 | Group: `g.Log` (embedded `log.Log`, api.go:47, init api.go:62)；Service: `s.Log` (service.go:122, init service.go:133) | ✅ 有效 logger 实例 |
| `spaceID` (event.go:812) | `spaceIDByGroupNo[groupNo]` | event.go:747 从 `group.SpaceID` 填充 → `QueryGroupsWithGroupNos` → `SELECT * FROM group WHERE group_no IN (?)` (db.go:332-338) | ✅ 来自 DB 的真实 SpaceID |
| `spaceID` (command.go:665) | `g.SpaceID` | `GetGroupsWithMemberUID(botID)` → `[]*InfoResp` (service.go:82) → `InfoResp.SpaceID` (service.go:648) | ✅ |
| `botID` (command.go:665) | Redis state machine | `h.sm.GetBotID(fromUID, ...)` → Redis hash `FieldBotID` → 源自 `selectedBot.RobotID` (command.go:525) | ✅ |
| `shortIDs` (thread_cleanup.go:53) | `queryThreadShortIDsForCleanup` | `SELECT short_id FROM thread WHERE group_no=? AND status!=3` → 直接查 thread 表，无 JOIN | ✅ 返回所有非删除子区 |
| `channelID` (thread_cleanup.go:73) | `groupNo + "____" + shortID` | 与 `thread.BuildChannelID` (thread/service.go:801) + `ChannelIDSeparator = "____"` (thread/const.go:17) 一致 | ✅ |
| `req.Operator` (event.go:812) | `OrgEmployeeExitReq.Operator` | JSON 反序列化 → event.go:722 验证非空 → event.go:787 用于 `DeleteMemberTx` 确认是退出者本人 | ✅ 语义正确（尽管 octo-lib 注释误导） |

---

**[Octo-Q] verdict: APPROVE**

本 PR 修复了一个真实的、用户可见的 bug（被移除的 bot 持续收到子区消息），根因分析准确，修复方案与 add 路径对称，代码去重合理，测试覆盖充分。未发现 P0/P1 阻塞问题。3 个 P2 均为非阻塞的性能/维护性建议。

PR #303, opened 2026-06-08
fix(botfather): always generate lowercase bot IDs (Fixes #302)

### LINKED ISSUE #302: BotFather generates mixed-case bot IDs that mismatch OpenClaw's lowercase normalize layer
### Description

`generateBotID()` in `modules/botfather/command.go` produces **mixed-case** bot IDs (e.g. `27pBwzf2F6bfa5cd142_bot`, `27gshU6HOTb87d88a0f_bot`). This is the immediate consequence of `util.Ten2Hex()` (Base62 charset `0-9 + a-z + A-Z`) being applied to the nanosecond timestamp portion of the ID.

Mixed-case IDs **silently mismatch** at every OpenClaw integration boundary, because OpenClaw's session-binding service and routing layer normalize every account ID to lowercase via `normalizeOptionalLowercaseString` before lookup. The adapter side has had to add a case-insensitive fallback (`Mininglamp-OSS/openclaw-channel-octo#55`) and an adapter-level normalize step (`Mininglamp-OSS/openclaw-channel-octo#32`) just to keep mixed-case bots working — and the meta-tracker `Mininglamp-OSS/openclaw-channel-octo#33` enumerates four additional cache/lookup sites in the adapter that still leak this class of bug.

The cleanest fix is at the source: make `generateBotID()` emit lowercase IDs from creation. New bots will not mismatch; the long tail of adapter-side compat code can stop growing.

### Steps to Reproduce

1. Send `/newbot` to BotFather in a fresh Octo deployment
2. Receive bot ID and observe the timestamp portion
3. Run `[[ "$BOT_ID" =~ [A-Z] ]] && echo "mixed-case"` against several newly-created bots
4. Observe that ~50% of generated bot IDs contain at least one uppercase letter (each Base62 digit has a 26/62 chance of resolving to A-Z)

### Expected Behavior

`generateBotID()` returns lowercase-only IDs (charset `0-9 + a-z`), matching the lowercase convention OpenClaw and downstream consumers normalize to.

### Actual Behavior

`generateBotID()` returns mixed-case IDs because `util.Ten2Hex()` uses a Base62 charset that includes `A-Z`. Examples already in production: `27pBwzf2F6bfa5cd142_bot`, `27gshU6HOTb87d88a0f_bot`.

### DMWork Version

main @ `f8581dd` (current HEAD as of filing; bug exists since BotFather module landed)

### Affected Component

Bot API / BotFather

### Severity

P2 - Non-core Feature Issue

### Reproducibility

---

## Proposed fix

Single-file change in `modules/botfather/command.go`, ~3 lines:

```go
// Before
func generateBotID() string {
    suffix, _ := randomHex(4)
    return util.Ten2Hex(time.Now().UnixNano()) + suffix + BotUsernameSuffix
}

// After
func generateBotID() string {
    suffix, _ := randomHex(4)
    return strings.ToLower(util.Ten2Hex(time.Now().UnixNano())) + suffix + BotUsernameSuffix
}
```

(`randomHex` already returns lowercase; `BotUsernameSuffix = "_bot"` is already lowercase. Only the `Ten2Hex` timestamp portion needs lowering.)

Do **not** change `util.Ten2Hex()` itself — it's reused for `user.shortNo` and other identifiers where mixed-case is fine. Apply the `.ToLower()` at the BotFather call site only.

## Risks of the fix

| Dimension | Assessment |
|-----------|------------|
| Existing mixed-case bot rows in `robot` table | 🟢 Safe. Table collation is `utf8mb4_general_ci` (case-insensitive), so `WHERE robot_id=?` already tolerates both forms. **No migration required.** |
| Entropy reduction (62 → 36 charset) | 🟢 Safe. Bot ID length ~22–26 chars × 5.17 bits/char = 113–134 bits, well above any practical collision threshold. |
| WuKongIM channel namespace | 🟢 Safe. WuKongIM is case-sensitive on UID, but a newly-generated lowercase ID can never collide with an existing mixed-case ID (nanosecond timestamp + 4-byte random suffix). Old and new bots live in disjoint channels by construction. |
| Bot token (`bf_...`) | 🟢 Safe. Tokens are generated independently of `bot_id` and stored verbatim — flipping the ID generator has no effect on token validity. |
| BotFather command parsing (`/token <botid>`, `/mybots`) | 🟢 Safe. User-typed bot IDs hit `WHERE robot_id=?`, already case-insensitive via collation. |
| Mobile / Web clients caching `bot_id` for comparison | 🟢 Safe. `grep` across iOS/Android/Web found no `bot_id` case-sensitive comparison code paths. |
| Long tail of adapter compat code | 🟢 Strict improvement. Stops new mixed-case IDs from being created; existing compat layers must remain to support legacy bots already in the wild. |

## Out-of-scope but worth auditing

Per the openclaw-channel-octo#33 recommendation, while in this code path it's worth auditing other ID generators that downstream consumers might normalize:

- `util.GenerUUID()` — used for `user.uid`. Already lowercase (Go `uuid` packages emit lowercase by default).
- `util.Ten2Hex()` direct uses outside BotFather — `user.shortNo` is consumed only inside server, no normalize boundary, fine to leave Base62.
- Group IDs, message IDs — confirmed numeric/sequential, no case issue.

No further generators appear at risk; the BotFather call site is the only one that crosses the OpenClaw normalize boundary.

## References

- Meta-tracker: Mininglamp-OSS/openclaw-channel-octo#33 (will stay open as a forever-tracker — adapter-side compat code must persist to keep historical mixed-case bots working)
- Concrete bugs caused by this:
  - Mininglamp-OSS/openclaw-channel-octo#32 (`SessionBindingAdapter.resolveByConversation` returning null for mixed-case)
  - Mininglamp-OSS/openclaw-channel-octo#55 (`resolveOctoAccount` lookup miss on mixed-case)
- Source: `modules/botfather/command.go` (function `generateBotID` near line 870)
- Generator: `pkg/util/base62.go` (`tenValue2Char`, Base62 charset)

caster-Q:
## Summary

`generateBotID()` previously concatenated `util.Ten2Hex` (Base62 alphabet, includes uppercase `A-Z`) with a random hex suffix, so the resulting bot ID could contain uppercase characters. OpenClaw's routing layer normalizes `accountId` to lowercase before lookups, so any uppercase letter caused mismatches that surfaced downstream as silently dropped messages, broken owner checks, and duplicated per-account caches.

Wrap the final string in `strings.ToLower` so every new bot ID is lowercase — matching the existing convention used by the routing / normalize path.

## Why

Closes #302 — root cause of the mixed-case bot ID issue tracked by the plugin side at [Mininglamp-OSS/openclaw-channel-octo#33](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/33).

## Compat

- Existing mixed-case bots in production remain functional via the case-insensitive (`utf8mb4_general_ci`) collation already in use on the `robot` table — no data migration required.
- Companion fix on the plugin side (openclaw-channel-octo#33) extends the same case-insensitivity to all in-memory caches and on-disk paths, so legacy mixed-case bots behave identically to new lowercase ones after both ship.

## Test plan

- [x] New unit test `TestGenerateBotID_AlwaysLowercase` (200 iterations) guards against regressions
- [x] Manual: `BotFather` `/newbot` flow on local `octo-server` (this branch) produces an all-lowercase bot ID (`27yxelxj3n665b00457_bot`, `27yxtfsyyjyeb9be348_bot` observed during local validation)
- [x] 5 rounds of `codex` plan review (12 → 8 → 6 → 2 → 0 findings) covering DB collation verification, no-migration-required justification, and naming consistency with the existing `lower()` callsites in the routing path

### DIFF
--- a/modules/botfather/command.go
+++ b/modules/botfather/command.go
@@ -979,9 +979,15 @@ func (h *commandHandler) sendCreatedPrompt(toUID string, name string, bot *robot
 
 // generateBotID 生成全局唯一的 Bot 标识符
 // 时间戳 Base62 + 4字节随机 hex，即使并发同纳秒也不会碰撞
+//
+// Lowercase the full returned ID so newly-generated bot IDs are case-insensitive-safe
+// against OpenClaw's normalizeOptionalLowercaseString routing layer. randomHex and
+// BotUsernameSuffix are already lowercase today; wrapping the whole concatenation
+// defends against future charset drift in either component.
+// See: octo-server#302, openclaw-channel-octo#33
 func generateBotID() string {
 	suffix, _ := randomHex(4)
-	return util.Ten2Hex(time.Now().UnixNano()) + suffix + BotUsernameSuffix
+	return strings.ToLower(util.Ten2Hex(time.Now().UnixNano()) + suffix + BotUsernameSuffix)
 }
 
 func (h *commandHandler) getBotDisplayName(robotID string) string {
--- a/modules/botfather/const_test.go
+++ b/modules/botfather/const_test.go
@@ -257,6 +257,25 @@ func TestGenerateBotID(t *testing.T) {
 	assert.NotEqual(t, id1, id2)
 }
 
+// TestGenerateBotID_AlwaysLowercase asserts the lowercase invariant on every
+// generated bot ID. util.Ten2Hex is Base62 (charset 0-9 + a-z + A-Z); its
+// uppercase half hits ~50% of timestamp chars, so 200 samples gives near-
+// certainty if the strings.ToLower wrapper in generateBotID() is ever removed.
+//
+// Background: octo-server#302 / openclaw-channel-octo#33 — mixed-case IDs
+// silently mismatch against OpenClaw's lowercase normalize layer.
+func TestGenerateBotID_AlwaysLowercase(t *testing.T) {
+	for i := 0; i < 200; i++ {
+		id := generateBotID()
+		if id != strings.ToLower(id) {
+			t.Fatalf("generateBotID() returned mixed-case ID: %q (iter %d)", id, i)
+		}
+		if !strings.HasSuffix(id, BotUsernameSuffix) {
+			t.Fatalf("generateBotID() missing suffix %q: %q", BotUsernameSuffix, id)
+		}
+	}
+}
+
 func TestBotNameValidation(t *testing.T) {
 	// Pure string-length validation logic — no DB / Redis / IM dependencies.
 	// 模拟 onBotNameInput 中的验证逻辑


lml2468 (approved):
## Review — APPROVE ✅

Head: `e779c112ebf9bc5544ec34f857736660c2c816cc`

干净的小修，根因 + 修复点位 + 回归测试三件事都对齐。

### 根因诊断验证

- `util.Ten2Hex`（octo-lib `pkg/util/base62.go`）alphabet 是 `0-9 + a-z + A-Z`——名字叫 Hex 实际是 base62，timestamp 部分约 **50% 概率** 出大写字母
- `randomHex(4)` 是 hex 小写 `0-9a-f` ✅
- `BotUsernameSuffix = "_bot"` ✅
- → 最终 ID 大概率含大写字母 → 与 OpenClaw routing 层 lowercase normalize 不匹配 → silently dropped messages / 重复缓存 / owner 检查失败

修复 `strings.ToLower(...)` 包裹整串放在唯一生成入口 `generateBotID()`。grep 确认全 repo 只有 `createBotCoreWithRetry:825` 一处调用，无遗漏点位。

### 回归测试

200 次迭代 + `id != strings.ToLower(id)` 断言：base62 timestamp 字符大写命中率 ≈50%，单次 ID 不含大写概率 ≤ 0.5^(timestamp字符数)，200 次仍纯小写的假阴性概率工程上 = 0。`HasSuffix` 顺手钉了后缀契约。✅

`util.Ten2Hex` 名字误导（base62 不是 hex）已被注释明确指出，未来 reviewer 不会被名字带偏。

### 存量数据 & 跨 repo 一致性

commit message 说生产 robot 表已用 case-insensitive collation，配套 plugin fix `openclaw-channel-octo#33` 覆盖 in-memory cache + on-disk paths——两端都收敛到小写口径，存量混合大小写 bot 继续工作。✅ 这是一个完整的 fix，不是只修一边的半成品。

### 防御性写法

注释明确说「randomHex 和 BotUsernameSuffix 当前已小写，包整串是防 future charset drift」——这是有意识的防御，不是无脑加层。合理。

### CI

- check-sprint 红：issue Sprint 是 W24，当前 W23，maintainer 板子配置问题，与代码无关
- 其它 lint/test/build 待跑完

### 非阻塞（不在本 PR scope）

- 🔵 上游 `octo-lib` 的 `util.Ten2Hex` 函数名误导（叫 Hex 实际 base62），未来可考虑 lib 层加 `Ten2LowerBase62`/`Ten2Base62` 别名 + 文档警告，或在 docstring 显式标注「contains uppercase」。本 PR scope 内已在调用方注释里点明，足够

---

**Bottom line:** 根因找对 + 修复点位单一 + 回归测试有意义 + 跨 repo 配套到位。Ship.

yujiawei (approved):
# Code Review — PR #303 (octo-server)

## Verdict: APPROVED

A minimal, correctly-scoped fix. `generateBotID()` now wraps the full ID in `strings.ToLower`, guaranteeing every newly-generated bot ID is lowercase and therefore safe against the lowercase-normalizing routing layer. No blocking (P0/P1) issues found.

## Verification

- ✅ **Root cause is real.** `util.Ten2Hex` (`pkg/util/base62.go`) is a Base62 encoder whose `tenValue2Char` emits uppercase `A–Z` for values 36–61. Since the high bits of `time.Now().UnixNano()` routinely land in that range, the timestamp segment of the old `Ten2Hex(...) + suffix + BotUsernameSuffix` concatenation could contain uppercase letters — exactly the mixed-case ID the PR describes.
- ✅ **Fix is correct and complete for the generated path.** `modules/botfather/command.go:990` wraps the entire concatenation. `randomHex` (`crypto/rand` → `hex.EncodeToString`, line 1147) and `BotUsernameSuffix` (`"_bot"`, `const.go:16`) are already lowercase, so `ToLower` only affects the timestamp segment today; wrapping the whole string also defends against future charset drift, as the comment notes.
- ✅ **Consistent with existing convention.** The user-supplied username path already normalizes via `strings.ToLower(req.Username)` (`api_user.go:157`). This change brings the auto-generated path in line with that same invariant — there was no second un-normalized entry point left behind.
- ✅ **No new imports / build impact.** `strings` is already imported in `command.go`. `go vet ./modules/botfather/` is clean.
- ✅ **Test is meaningful.** `TestGenerateBotID_AlwaysLowercase` runs 200 iterations asserting `id == strings.ToLower(id)` plus the suffix invariant. Given the uppercase half of Base62 hits a large fraction of timestamp chars, 200 samples gives near-certainty of catching a regression if the wrapper is removed. Both `TestGenerateBotID` and the new test pass locally (`go test ./modules/botfather/ -run TestGenerateBotID`).

## For human verification (security-sensitive)

- ⚠️ **Production collation of `robot.robot_id`.** The compat claim ("existing mixed-case bots keep working via case-insensitive collation, no migration") rests on the live column collation. The base migration (`modules/robot/sql/20210926000001_robot_legacy01.sql`) declares `robot_id VARCHAR(40)` with **no explicit collation**, so the effective collation is inherited from the DB/table default and is not pinned in the repo. The fix itself is safe regardless (new IDs are lowercase), but the "no migration required" guarantee for *legacy* mixed-case bots cannot be confirmed from the source tree alone — worth a quick `SHOW FULL COLUMNS FROM robot` against production to confirm it is indeed `utf8mb4_general_ci` (or another `_ci` collation) before relying on the case-insensitive lookup for existing rows.
- Note: `queryRobotByRobotID` (`db.go:61`) issues a plain `WHERE robot_id=?`, so case-insensitivity of legacy lookups depends entirely on that collation — reinforcing the check above.

## Suggestions (non-blocking)

- **(nit, out of scope)** The underlying sharp edge is that `util.Ten2Hex` produces mixed-case output despite the function being named like a hex encoder and used in ID-generation contexts. A future cleanup could either rename it or offer a lowercase variant so other call sites don't reintroduce the same class of bug. Not required for this PR.

This PR is safe to merge.

PR #311, opened 2026-06-08
ci: pass TRIAGE_WEBHOOK_URL to octo-issue-notify

lml2468:
The issue notification caller only forwarded OCTO_BOT_TOKEN, so the triage webhook mode was silently skipped and the triage agent was never invoked. Forward the org-level TRIAGE_WEBHOOK_URL secret so triage runs on new issues.

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -18,3 +18,4 @@ jobs:
       event_action: ${{ github.event.action }}
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}


Jerry-Xin (approved):
Summary: The PR is in scope for this repository and correctly extends the existing issue notification workflow by forwarding the triage webhook secret to the reusable notification workflow.

🔴 Blocking: None.

💬 Non-blocking: None.

✅ Highlights:
- ` .github/workflows/octo-issue-feed.yml:19-21` keeps the existing explicit secret forwarding style and adds only the required `TRIAGE_WEBHOOK_URL`.
- The change is narrowly scoped to CI workflow configuration and does not introduce application runtime, permission, or test coverage concerns.

yujiawei (approved):
# Code Review — PR #311 (octo-server)

**Verdict: APPROVED** — no blocking issues. This is a correct, minimal, security-clean one-line CI wiring change.

## Summary

The change adds a single line to `.github/workflows/octo-issue-feed.yml`:

```yaml
    secrets:
      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+     TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}
```

This forwards the org-level `TRIAGE_WEBHOOK_URL` secret to the reusable workflow `octo-issue-notify.yml@v1`, activating the previously-dormant triage webhook leg so the triage agent is actually invoked on new issues. The PR description matches the diff exactly.

## Verification (end-to-end secret wiring)

I traced the secret through the full call chain at head SHA `8abe61e`:

- ✅ **Caller** (`octo-issue-feed.yml`) — passes `TRIAGE_WEBHOOK_URL` under `secrets:`, the canonical-safe placement. No `${{ }}` interpolation of user data. `permissions: {}` (least privilege).
- ✅ **Reusable workflow** (`octo-issue-notify.yml@v1`) — declares `secrets.TRIAGE_WEBHOOK_URL` with `required: false`. **Name matches exactly** at every hop, so there is no wrong-name / silently-empty footgun. `permissions: {}`.
- ✅ **Transport** (`octo-notify@v1` composite action) — consumes the URL via `env:` (`WEBHOOK_URL`), never interpolating `${{ }}` into a shell or into a JSON string. Validates the payload is JSON before POST; retry/backoff with `Retry-After`.

## Security assessment (this PR is `security_sensitive`)

No new attack surface is introduced by this diff:

- **No secret leakage.** Secrets passed via `secrets:` are masked by GitHub in logs; the transport action never echoes them. The webhook URL is an opaque capability secret POSTed to itself, not derived from any user input (no SSRF).
- **No script/expression injection.** The trigger is `issues` (not `pull_request_target`), so untrusted forks cannot alter the workflow code that runs. Attacker-influenced issue `title`/`author` are already (pre-PR) control-char-stripped, length-truncated, `repo_name` regex-validated, gated to `event_action == 'opened'`, and serialized via `json.dumps` — so they cannot redirect the destination or inject into `GITHUB_OUTPUT`.
- **Behavioral delta is the intended feature only:** the webhook leg now fires instead of no-op'ing on an empty URL.

## Non-blocking notes (P2 / nits — do not block merge)

- **P2 — silent no-op if the org secret is unset.** Because the consumer declares `TRIAGE_WEBHOOK_URL` as `required: false`, if the org/repo secret is never configured the value resolves to an empty string and the webhook leg silently skips — the feature ships but never delivers, with no loud failure. This is an operability property of the existing reusable workflow, not a defect in this diff. Worth a presence check (warn loudly when the webhook branch is expected but the URL is empty) or a note that the org secret MUST be set.
- **nit — downstream trust boundary.** The webhook now fires on issues opened by anyone (including external users), forwarding sanitized-but-attacker-influenced `title`/`author`. The triage endpoint behind the URL should treat these as untrusted input (no HTML/markdown auto-render, no command interpolation, no `eval` of payload fields). Out of scope for this repo, but worth one line to the endpoint owner.
- **nit — pre-existing supply-chain.** The reusable workflow and composite action are pinned to the mutable `@v1` tag rather than an immutable commit SHA. Untouched by this diff (the `uses:` line already referenced `@v1`), so no regression — but consider pinning to full SHAs in a separate hardening PR, especially now that they handle a populated webhook secret.
- **nit — `reopened` accepted but dropped.** The caller triggers on `[opened, reopened]` while the reusable forwards only `opened`. Pre-existing; harmless but wasteful.

## Items a human should manually verify

1. Confirm the org/repo secret `TRIAGE_WEBHOOK_URL` actually exists and points at the correct internal triage endpoint — a missing/typo'd secret is a **silent** no-op (see P2).
2. Confirm the triage endpoint treats incoming `title`/`author` as untrusted.
3. After merge, open a test issue and confirm a webhook POST is actually delivered end-to-end.

None of these are code defects in the PR; they are operational confirmations appropriate for a security-sensitive secret-wiring change.

PR #3, opened 2026-05-11
YUJ-446 Round-5: fix server boot crash — register notify + oidc modules

yujiawei:
Follow-up from [PR #2](https://github.com/Mininglamp-OSS/octo-server/pull/2).

## Problem

Round-5 end-to-end verification re-deployed the stack from a clean
\`docker compose down -v\` state, exactly following the quickstart
in the new README. The stack brought up all 11 containers, but
octo-server crash-looped with:

```
panic: Unable to create migration plan because of oidc-20260427-01.sql: unknown migration in database

panic: Unable to create migration plan because of app_bot-20260510-01.sql: unknown migration in database
```

## Root cause

\`scripts/init-db.sql\` (the schema snapshot) seeds \`gorp_migrations\`
rows for every migration in the internal dmworkim source tree,
including rows for the \`notify\` and \`oidc\` modules. The module
source for both is present in \`modules/notify/\` and \`modules/oidc/\`
with \`//go:embed sql\` wired through \`1module.go\`.

However \`internal/modules.go\` only blank-imports 22 of the
24 relevant module packages — \`notify\` and \`oidc\` are missing.
Without the blank import, \`init()\` in \`1module.go\` never runs,
\`register.AddModule\` is never called, and sql-migrate does not see
the embedded \`sql/\` directory at boot. The seeded rows then look
like \"unknown migrations\" to the planner and it panics.

This blocks the documented quickstart path entirely; every fresh
OSS deployment of this repo crashes the API server immediately
after MySQL finishes initializing.

Why didn't PR #2 catch this? The im-test pilot reused a MySQL
volume that had already been populated by a pre-existing \`octo-server\`
binary built from a different source drop (which *did* register both
modules), so the migration plan matched on that first boot. A fresh
OSS user following the README from scratch would hit this every
time.

## Fix

Two lines in \`internal/modules.go\`:

```go
_ \"github.com/Mininglamp-OSS/octo-server/modules/notify\"
_ \"github.com/Mininglamp-OSS/octo-server/modules/oidc\"
```

Placed between \`message\` and \`openapi\` to keep the file's existing
\"group by alpha, respect ordering comments\" convention. No ordering
dependency affects these two — \`notify\` only needs \`message\`
semantics at runtime (already registered above) and \`oidc\` depends
on \`user\` (already registered later, but no SQL cross-dependency).

## Verification

Fresh re-deploy from merged main HEAD:

| Endpoint              | Expected  | Actual |
| --------------------- | --------- | ------ |
[...]

PR #315, opened 2026-06-09
feat(opanalytics): add manual ETL trigger endpoint

### LINKED ISSUE #314: feat(opanalytics): superAdmin manual ETL trigger endpoint (on-demand dashboard refresh)
## Background

The ops analytics dashboard (`modules/opanalytics`, shipped in #308) populates its dim/fact tables only via the daily scheduler at `30 1 * * *` (Asia/Shanghai). There is **no on-demand trigger**.

Observed during post-merge verification: the four `GET /v1/manager/dashboard/*` endpoints return HTTP 200 with the correct shape, but counts that come from the ETL-populated tables are all 0 until the first scheduled run:
- `spaces` → `count=209` (from the live `space` table) but per-row member/msg/active are 0.
- `spaces/{id}/channels` → `{"count":0,"list":[]}` (reads `octo_dim_channel`, still empty).
- `global/direct-chats` → `{"count":0,"list":[]}` (reads `octo_fact_channel_daily`, still empty).

This is expected pre-ETL, but it means QA/ops/frontend can't validate **real data** until ~01:30, and the very first run is a full-history backfill. We need a way to kick the ETL on demand.

## Proposal

Add a superAdmin-only endpoint to trigger an incremental ETL run on demand.

- **Route:** `POST /v1/manager/dashboard/etl/run` — mounted on the same group as the other dashboard routes (`AuthMiddleware` + per-handler `requireSuperAdmin` + `SharedUIDRateLimiter`).
- **Behavior:** acquire the existing Redis ETL lock (`opanalytics:etl:run`, same key the scheduler uses), then run `ETL.RunIncremental()` **asynchronously** and return **202 Accepted** immediately (the first run is a long full-history backfill — must not block the request).
- **Concurrency:** if the lock is already held (daily cron or a previous manual run still in flight), return a clear "already running" response (i18n envelope, dedicated code) instead of starting a second run. Reusing the scheduler's lock guarantees manual + cron never double-run; the cursor `FOR UPDATE` remains the correctness backstop.
- **Observability:** log start/finish with `messages_scanned` / `recomputed_days` (already logged by `RunIncremental`); optionally surface the run outcome via a follow-up status read (out of scope here).

## Acceptance criteria

- [ ] `POST /v1/manager/dashboard/etl/run` returns 202 for a superAdmin and kicks an async `RunIncremental()`.
- [ ] Non-superAdmin → `err.server.opanalytics.forbidden`; rate-limited like the other routes.
- [ ] Concurrent trigger (or while the daily cron holds the lock) → clear "already running" response, no double run.
- [ ] New i18n code(s) registered + zh-CN translation + en marker; passes `make i18n-extract-check` / `make i18n-lint`; added to the `TestOpanalyticsNoLegacyResponseError` guard list.
- [ ] `modules/opanalytics/swagger/api.yaml` updated with the new endpoint.
- [ ] Integration test: trigger → fact tables populated; 403 for non-superAdmin; concurrent-trigger path.

## Out of scope / notes

- **Do not** expose the destructive `Rebuild()` (truncate + full recompute) through this endpoint without an explicit confirmation flag — keep this endpoint to the safe incremental run only. A separate guarded `rebuild` action can be a later issue.
- `Rebuild()` currently does not take the Redis lock; if it is ever wired to an endpoint, it must acquire the same lock first.

Follow-up to #308.

an9xyz:
## Summary

- Add `POST /v1/manager/dashboard/etl/run` for superAdmin-only manual incremental dashboard ETL runs.
- Reuse the scheduler Redis lock `opanalytics:etl:run`, return an i18n `etl_already_running` envelope when cron or another manual run holds it, and run `RunIncremental()` asynchronously after returning 202.
- Add token-checked Redis lock renewal while ETL is in flight so long first backfills do not outlive the lock lease.
- Update opanalytics Swagger 2.0 spec and i18n marker/zh-CN translations.

Closes #314.

## Testing

- `go test ./modules/opanalytics`
- `go test ./modules/opanalytics -run 'TestOpanalyticsManualETLRun|TestETLLockRenewRequiresMatchingToken'`
- `go test ./pkg/errcode ./pkg/httperr ./pkg/i18n`
- `make i18n-extract`
- `make i18n-extract-check`
- `make i18n-lint`
- `git diff --check`

## E2E

Started the current branch as a real octo-server API on `:18090`, seeded MySQL/Redis test data, and verified through HTTP:

- `POST /v1/manager/dashboard/etl/run` returned `202` with `{"state":"accepted","status":202}`.
- The manual ETL completed with `messages_scanned=4` and `recomputed_days=1` in server logs.
- `GET /v1/manager/dashboard/overview?start_date=2026-06-01&end_date=2026-06-01&space_ids=e2e_s1` returned expected aggregates: `human_msg_count=3`, `agent_msg_count=1`, `active_groups=1`.
- Non-superAdmin trigger returned `err.server.opanalytics.forbidden` with `error.http_status=403`.
- Pre-held `opanalytics:etl:run` lock returned `err.server.opanalytics.etl_already_running` with `error.http_status=409`.

### DIFF
--- a/modules/opanalytics/api.go
+++ b/modules/opanalytics/api.go
@@ -1,6 +1,7 @@
 package opanalytics
 
 import (
+	"net/http"
 	"time"
 
 	"github.com/Mininglamp-OSS/octo-lib/config"
@@ -57,6 +58,7 @@ func (m *Manager) Route(r *wkhttp.WKHttp) {
 		auth.GET("/spaces", m.spaces)                           // 表一 Space 列表
 		auth.GET("/spaces/:space_id/channels", m.spaceChannels) // 表二 群组列表(仅群组)
 		auth.GET("/global/direct-chats", m.globalDirectChats)   // 全局私聊活跃列表
+		auth.POST("/etl/run", m.runETL)                         // 手动触发增量 ETL
 	}
 }
 
@@ -157,6 +159,32 @@ func (m *Manager) globalDirectChats(c *wkhttp.Context) {
 		return
 	}
 	c.Response(map[string]interface{}{"count": total, "list": list})
+}
+
+// runETL 手动触发一轮增量 ETL。首轮可能是长时间全历史回填，因此只接受任务并异步执行。
+func (m *Manager) runETL(c *wkhttp.Context) {
+	if !m.requireSuperAdmin(c) {
+		return
+	}
+
+	accepted, err := m.scheduler.TriggerManualRun()
+	if err != nil {
+		m.Error("manual opanalytics ETL trigger failed",
+			zap.String("uid", c.GetLoginUID()),
+			zap.Error(err))
+		respETLTriggerFailed(c)
+		return
+	}
+	if !accepted {
+		respETLAlreadyRunning(c)
+		return
+	}
+
+	m.Info("manual opanalytics ETL accepted", zap.String("uid", c.GetLoginUID()))
+	c.ResponseWithStatus(http.StatusAccepted, map[string]interface{}{
+		"status": http.StatusAccepted,
+		"state":  "accepted",
+	})
 }
 
 // ===== 参数解析 =====
--- a/modules/opanalytics/api_i18n.go
+++ b/modules/opanalytics/api_i18n.go
@@ -30,3 +30,13 @@ func respNotFound(c *wkhttp.Context) {
 func respQueryFailed(c *wkhttp.Context) {
 	httperr.ResponseErrorL(c, errcode.ErrOpanalyticsQueryFailed, nil, nil)
 }
+
+// respETLAlreadyRunning 表示已有定时或手动 ETL 正在执行。
+func respETLAlreadyRunning(c *wkhttp.Context) {
+	httperr.ResponseErrorL(c, errcode.ErrOpanalyticsETLAlreadyRunning, nil, nil)
+}
+
+// respETLTriggerFailed 手动触发 ETL 失败 → 500(Internal，调用方须先记 zap 日志)。
+func respETLTriggerFailed(c *wkhttp.Context) {
+	httperr.ResponseErrorL(c, errcode.ErrOpanalyticsETLTriggerFailed, nil, nil)
+}
--- a/modules/opanalytics/api_i18n_test.go
+++ b/modules/opanalytics/api_i18n_test.go
@@ -78,6 +78,8 @@ func TestOpanalyticsRespondHelpers(t *testing.T) {
 		{"requestInvalid", func(c *wkhttp.Context) { respRequestInvalid(c, "date_range") }, http.StatusBadRequest, "err.server.opanalytics.request_invalid"},
 		{"notFound", respNotFound, http.StatusNotFound, "err.server.opanalytics.not_found"},
 		{"queryFailed", respQueryFailed, http.StatusInternalServerError, "err.server.opanalytics.query_failed"},
+		{"etlAlreadyRunning", respETLAlreadyRunning, http.StatusConflict, "err.server.opanalytics.etl_already_running"},
+		{"etlTriggerFailed", respETLTriggerFailed, http.StatusInternalServerError, "err.server.opanalytics.etl_trigger_failed"},
 	}
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
--- a/modules/opanalytics/api_test.go
+++ b/modules/opanalytics/api_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/Mininglamp-OSS/octo-lib/common"
 	"github.com/Mininglamp-OSS/octo-lib/config"
@@ -35,6 +36,20 @@ func opaSetup(t *testing.T) (*config.Context, *wkhttp.WKHttp, *ETL) {
 	return ctx, route, NewETL(ctx)
 }
 
+func opaRouteOnlySetup(t *testing.T) (*config.Context, *wkhttp.WKHttp) {
+	t.Helper()
+	cfg := config.New()
+	cfg.Test = true
+	ctx := config.NewContext(cfg)
+	route := wkhttp.New()
+	route.SetErrorRenderer(i18n.NewErrorRenderer(i18n.NewLocalizer(i18n.DefaultLanguage)))
+	ctx.SetHttpRoute(route)
+	setSuperAdminToken(t, ctx)
+	resetUIDRateLimit(t, ctx)
+	New(ctx).Route(route)
+	return ctx, route
+}
+
 // resetUIDRateLimit 清空 SharedUIDRateLimiter 的每-uid 令牌桶(ratelimit:uid:*)。该桶持久在
 // Redis、跨测试残留且不被 CleanAllTables 清，须在 setup 重置，否则同一 binary 内靠后的测试会 429。
 func resetUIDRateLimit(t *testing.T, ctx *config.Context) {
@@ -47,6 +62,7 @@ func resetUIDRateLimit(t *testing.T, ctx *config.Context) {
 	if keys, err := rds.Keys("ratelimit:uid:*").Result(); err == nil && len(keys) > 0 {
 		_ = rds.Del(keys...).Err()
 	}
+	_ = rds.Del(etlRunLockKey).Err()
 }
 
 func setSuperAdminToken(t *testing.T, ctx *config.Context) {
@@ -212,6 +228,15 @@ func opaGet(t *testing.T, route *wkhttp.WKHttp, path string) *httptest.ResponseR
 	return rec
 }
 
+func opaPost(t *testing.T, route *wkhttp.WKHttp, path string) *httptest.ResponseRecorder {
+	t.Helper()
+	req := httptest.NewRequest(http.MethodPost, path, nil)
+	req.Header.Set("token", testutil.Token)
+	rec := httptest.NewRecorder()
+	route.ServeHTTP(rec, req)
+	return rec
+}
+
 func decodeOK(t *testing.T, rec *httptest.ResponseRecorder, out interface{}) {
 	t.Helper()
 	require.Equalf(t, http.StatusOK, rec.Code, "body=%s", rec.Body.String())
@@ -228,6 +253,18 @@ func errorCode(t *testing.T, rec *httptest.ResponseRecorder) string {
 	}
 	require.NoError(t, json.Unmarshal(rec.Body.Bytes(), &env))
 	return env.Error.Code
+}
+
+func waitForFact4Human(t *testing.T, ctx *config.Context, channelID string, want int) {
+	t.Helper()
+	deadline := time.Now().Add(5 * time.Second)
+	for time.Now().Before(deadline) {
+		if got := fact4Human(t, ctx, channelID); got == want {
+			return
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+	assert.Equal(t, want, fact4Human(t, ctx, channelID))
 }
 
 // ---- the scenario seed ----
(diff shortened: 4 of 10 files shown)


Jerry-Xin (approved):
Summary: This PR is in scope for `octo-server` and cleanly extends the existing opanalytics manager API and ETL scheduler.

🔴 Blocking  
None.

💬 Non-blocking  
🟡 Warning: I could not complete local `go test ./modules/opanalytics` because the test environment has no MySQL available at `127.0.0.1:3306`; the package compiled and began running tests before `testutil.NewTestServer` failed. CI should confirm the stated test matrix.

✅ Highlights  
- `modules/opanalytics/api.go:55` and `modules/opanalytics/api.go:165` keep the endpoint behind auth, UID rate limiting, and superAdmin enforcement.
- `modules/opanalytics/scheduler.go:80` reuses the scheduler lock path for manual runs, avoiding a separate concurrency mechanism.
- `modules/opanalytics/etl_lock.go:75` adds token-checked renewal, preserving lock ownership semantics.
- `pkg/errcode/opanalytics.go:35` and `modules/opanalytics/api_i18n.go:34` follow the project’s i18n error envelope pattern for the new ETL errors.
- `modules/opanalytics/api_test.go:507` covers accepted manual execution, forbidden access, already-running behavior, and token-safe renewal.

lml2468 (approved):
APPROVED (head `fcede19e18c6`).

Real new feature (manual ETL trigger endpoint) on top of `opanalytics`. Traced the full path — lock semantics, auth gate, async pattern, test coverage. Solid implementation.

## ✅ Independently verified

**Auth + rate limit gating is tight.**
`POST /v1/manager/dashboard/etl/run` is mounted in the existing `auth` group with `AuthMiddleware` + `SharedUIDRateLimiter`, identical to the 4 read routes. The handler then calls `m.requireSuperAdmin(c)` first, matching the per-handler pattern used by `overview`/`spaces`/`spaceChannels`/`globalDirectChats`. A leaked superAdmin token can't hammer the endpoint because the per-uid rate limit (mounted post-auth) shares one bucket across all 5 dashboard routes.

**Lock contention model is correct — cron and manual share one mutex.**
`TriggerManualRun` calls the refactored-out `acquireRunLock()` (same `opanalytics:etl:run` key, same `etlRunLockTTL=30min` TTL, same random token), so a manual trigger can't race the daily cron and vice versa. `(false, nil)` on `acquired=false` → 409 `etl_already_running` (handler does not log warn for this — correct, it's the expected concurrent-attempt path, not an error). `(false, err)` on Redis fault → 500 `etl_trigger_failed` with `Internal:true` (5xx hides message, only logged). Refactor of `runOnce` to use the same helper is a clean dedup, not a behavior change.

**Async fire-and-forget pattern is right for the use case.**
First-run backfill from `id=0` across all message shards can take minutes (per Fix #1 from earlier P0 review). Synchronously holding the request for that long would tie up the HTTP worker + risk client timeouts. The endpoint returns 202 immediately and runs `runWithHeldLock` in a goroutine. The lock is released inside the goroutine's `defer` so external state correctly reflects the in-progress run — a second POST after 202 will see the lock held and 409.

**Lock renewal is the right pattern, with proper CAS semantics.**
The new `luaRenewETLLock` script (`if GET == token then PEXPIRE`) mirrors the existing `luaReleaseETLLock` CAS-DEL pattern — won't extend a foreign owner's lock if the lease lapsed mid-flight. The renewer goroutine runs at `etlRunLockTTL/3 = 10min` interval (industry-standard 3x safety margin). Renewal failures are handled correctly:
- Transient Redis error → log + `continue` (try again next tick — fault-tolerant)
- `Renew` returns false (lock lost / owner changed) → log + return (stops renewing, but `RunIncremental` continues — the cursor `SELECT ... FOR UPDATE` from Fix #3 is the second backstop, so a second instance starting after lease lapse will block on the per-shard cursor row rather than double-write)

**No goroutine leak / lock-not-released risk.**
`runWithHeldLock` uses `done := make(chan struct{})` + `defer close(done)` before `defer lock.Release`. Even if `RunIncremental` panics (no `recover` here, see note 1 below), Go's `defer` chain runs in LIFO during stack unwind: `close(done)` first → renewer wakes on `<-done` and exits → then `lock.Release` runs with the still-valid token. The lock won't be orphaned for 30min after a crash; it'll be released cleanly on panic too (until the panic propagates and kills the goroutine).

**Test coverage hits the right invariants.**
- `TestOpanalyticsManualETLRunEndpoint`: POST → 202 → `waitForFact4Human` polls the fact table → asserts ETL actually ran end-to-end (not just that the lock was grabbed).
- `TestOpanalyticsManualETLRunForbiddenAndAlreadyRunning`: 403 path + the critical 409 path. The clever assertion **"already-running path must not drop or replace the held lock"** (re-attempt `lock.Acquire("second-token")` after the 409 and assert it returns false) is the security-relevant check — proves the handler doesn't somehow steal the lock from the active holder.
- `TestETLLockRenewRequiresMatchingToken`: covers all 4 quadrants of `Renew` CAS — matching token = renew + held; wrong token = no renew + still held (re-`Acquire` proves it); `Release` after = lock dropped + re-`Acquire` succeeds.

**Test isolation correct.**
`opaRouteOnlySetup` (new helper for tests that don't need full ETL state) calls `New(ctx).Route(route)`; `New` creates a `Scheduler` but `if !cfg.Test` gates `scheduler.Start()` — so cron doesn't fire and contend for the lock in tests. `resetUIDRateLimit` also clears `etlRunLockKey` to prevent cross-test lock residue.

**Swagger + i18n complete.**
- New `/manager/dashboard/etl/run` path with `etlRunAcceptedResp` schema (status + state).
- `dashboardError` description updated to mention 409 (`etl_already_running`) and the new error codes.
- Both `pkg/i18n/locales/active.zh-CN.toml` and `tools/i18nmarkers/server/active.en-US.toml` get the two new keys (`etl_already_running` + `etl_trigger_failed`). `make i18n-extract-check` / `make i18n-lint` confirmed in PR description.
- The existing `TestOpanalyticsRespondHelpers` table-test got 2 new rows pinning the 409/500 wire status + error code IDs — same pattern as the other dashboard helpers, prevents accidental code/status drift.

1. **Goroutine panic recovery asymmetry — same as scheduled path, but worth noting.** `cron.New()` in `NewScheduler` does NOT install `cron.Recover()` middleware (robfig/cron/v3 makes that explicit via `WithChain`). So `runOnce` (scheduled) and the new `go s.runWithHeldLock(token, "manual")` (manual) both let a `RunIncremental` panic propagate and crash the process. Not a regression — the manual path inherits the same risk surface. A future hardening pass could wrap both call sites with `defer recover()` + log + `lock.Release` already covered by the outer defer chain. Strictly cosmetic for this PR.

2. **Renewer's "lock lost" branch is a silent escalation point.** If `Renew` returns false (lease lapsed → another instance acquired), the renewer logs an error and exits — but the ongoing `RunIncremental` keeps running with a token that no longer owns the lock. Final `lock.Release(token)` will be a CAS-mismatch no-op (correct, doesn't drop the new owner's lock). Worst-case: two instances run ETL concurrently for the tail of the run, both relying on the cursor `FOR UPDATE` to serialize the actual writes. This is **acceptable** because Fix #3's cursor-FOR-UPDATE is exactly the second backstop designed for this scenario, but a comment in `renewLockUntilDone` near the "lock ownership lost" log noting "RunIncremental continues; cursor FOR UPDATE serializes against any new owner — see etl.go" would make this design intent legible to the next maintainer. Doc-only.

3. **`runWithHeldLock` variadic `zap.Field` signature is fine but slightly unusual.** The manual path passes no extra fields; scheduled passes `cron`. A future tracing addition (uid for manual, last-run-id for incremental) would also benefit from a `runID` field tying together the trigger log → renew warnings → finished log. Not for this PR; just sets up nicely.

## CI

`code-review` PENDING; the rest of CI (Build / Test / Vet / Lint / CodeQL / i18n / OSV) is queued or in progress at review time. Ride them to green before merge.

## Verdict

Clean implementation. Reuses the existing lock infrastructure correctly, adds the minimum-viable renewal primitive (CAS-on-renew matches CAS-on-release), preserves the anti-enumeration / superAdmin-only / rate-limit baseline of the dashboard routes, and the tests pin the security-relevant "409 doesn't steal the lock" invariant. The async 202 pattern is the right call for variable-duration ETL. LGTM, merge once CI confirms green.

yujiawei (approved):
# Code Review — PR #315 (octo-server)

**Verdict: APPROVED** — no P0/P1 blockers. The new `POST /v1/manager/dashboard/etl/run` manual ETL trigger is correctly authorized, the distributed-lock + renewal logic is sound, and test/i18n/swagger coverage is complete. A few non-blocking hardening notes below.

## Summary of the change
Adds a superAdmin-only endpoint that triggers one incremental dashboard ETL run, reusing the same Redis lock (`opanalytics:etl:run`) as the daily cron. It acquires the lock, returns `202 Accepted`, and runs `RunIncremental()` asynchronously, periodically renewing the lock lease so a long first-time backfill does not outlive the TTL. Returns `409 etl_already_running` when cron or another manual run holds the lock, `403` for non-superAdmin, `500 etl_trigger_failed` on Redis error.

## 1. Verification (security-sensitive focus)

- ✅ **Authorization** — `runETL` calls `requireSuperAdmin(c)` → `CheckLoginRoleIsSuperAdmin()` as its first statement with early return (`api.go:165-168`), identical to the four sibling dashboard handlers. The route is mounted on the existing auth group behind `AuthMiddleware` + `SharedUIDRateLimiter` (`api.go:55,61`): unauthenticated → 401, non-superAdmin → 403, per-UID rate-limited.
- ✅ **State-changing verb / injection surface** — uses `POST`, not `GET`. The handler reads no body or query params (`TriggerManualRun()` takes no args), so there is zero injection surface.
- ✅ **Error envelopes** — `403 forbidden` / `409 already_running` / `500 trigger_failed`. The 500 is `Internal=true` (renderer hides cause; cause is logged via `zap` first at `api.go:172-174`), the 409 is non-Internal — both comply with the repo's `5xx ⟺ Internal` invariant (`pkg/errcode/opanalytics.go:35-45`). No information leak.
- ✅ **Distributed-lock correctness** — `Acquire` is `SET NX EX` (atomic single-winner). `Release` is Lua CAS-DEL guarded on token; the new `Renew` is Lua CAS `PEXPIRE` guarded on token (`etl_lock.go:34-40,76-83`). A stale owner cannot release, renew, or resurrect a successor's lock. The `close(done)` → `Release` teardown races benignly with at most one final `Renew`, and the CAS guard makes both orderings safe.
- ✅ **Mutual exclusion (cron vs manual)** — both paths funnel through `acquireRunLock` → `Acquire`; the loser cron-path skips, the loser manual-path returns `(false,nil)` → `409` without touching the held lock (asserted by `TestOpanalyticsManualETLRunForbiddenAndAlreadyRunning`, `api_test.go:217-221`).
- ✅ **Lock expiry → double run** — even if the renewer were starved for a full 30-min TTL (renew interval is TTL/3 = 10 min, ~3 attempts/window), a concurrent second run is not data loss: `runChunk` does `SELECT ... FOR UPDATE` on the per-shard cursor row, serializing accumulation to exactly-once. Worst case is redundant work — the explicitly designed-for case.
- ✅ **DoS** — bounded by the superAdmin gate, the per-UID rate limiter, and the Redis lock (spam returns 409). Redis failure fails closed (500, no run).
- ✅ **Tests / i18n / swagger** — 202/403/409/renew-token paths all covered with strong negative assertions; en-US markers + zh-CN translations present for both new codes; swagger documents 202 + the `error.http_status` envelope contract; `scheduler.go` correctly added to the `NoLegacyResponseError` guard list; `resetUIDRateLimit` now also clears `etlRunLockKey` for cross-test isolation. `go build` + `go vet` clean.

- **P2 — Unrecovered panic in the async ETL goroutine.** `TriggerManualRun` spawns `go s.runWithHeldLock(...)` (`scheduler.go:86`); `gin.Recovery()` only wraps the request goroutine, so a panic in `RunIncremental()` would crash the process. This is **not a regression** — `cron.New()` is created with an empty chain (no `cron.Recover`, `scheduler.go:37`) and cron runs jobs in a bare goroutine too, so the daily cron path already has identical exposure, and the `modules/backup` scheduler this mirrors does the same. The deferred `close(done)` + `Release` still run during unwind, so no lock/goroutine leak. Since this endpoint lets a superAdmin invoke the path on demand, consider hardening both paths in a follow-up: `cron.New(cron.WithChain(cron.Recover(logger)))` plus a `defer recover()` at the top of `runWithHeldLock`.
- **nit — 202 happy-path test synchronizes on DB polling, not on the spawned goroutine.** `TestOpanalyticsManualETLRunEndpoint` waits on fact rows (`waitForFact4Human`) and returns before the lock is released; it is self-healing today (next setup clears the lock) but is the classic shape of a future flaky test if seeding/timing changes.
- **nit — `Stop()` does not wait for an in-flight async run before `lock.Close()`.** A concurrent `Renew`/`Release` on the closing client errors gracefully and the TTL self-heals; shutdown-only, cosmetic.
- **nit — the `"accepted"` state literal is duplicated between handler and test** with no shared const.

## 3. Human-verification callout (security_sensitive)
Nothing here requires a manual security gate beyond the standard superAdmin authorization, which is correctly enforced. The endpoint exposes no new data, takes no user input, and is idempotent under the lock. The one item worth a maintainer's awareness is the panic-safety follow-up above — it pre-dates this PR but this PR adds an on-demand trigger for the affected path.

PR #279, opened 2026-06-05
feat(i18n): migrate botfather module to ResponseErrorL (#278)

an9xyz:
## Summary
Migrate the **live** `modules/botfather` HTTP surface to the i18n error envelope, and **remove the bot handlers that became dead code when `bot_api` took over `/v1/bot/*` in #277**.

Closes #278. Part of #170 (i18n — backend / octo-server).

> **Scope changed mid-review.** This started as "migrate all 199 botfather response sites." While addressing review, I found that **most of botfather's HTTP handlers are unrouted dead code** — see below. The PR is now: migrate the live surface (`api_user.go` + `api_apply.go`) + delete the dead handlers. Net **+559 / −2606**.

## What is dead, and why
`botfather.Route()` (api.go) registers only: the doc `.md` endpoints, **Robot Apply** (`api_apply.go`), and **User Bot management** (`api_user.go`). Every `/v1/bot/*` bot-operation endpoint (groups, threads, send/typing/receipt/events/heartbeat, upload, message-edit, user-info, register) is now served by **`modules/bot_api`** (migrated in #277). The botfather copies in `api_bot.go`, `api_bot_thread.go`, and ~22 handlers in `api.go` have **zero route registrations** — confirmed by grep (`route refs = 0`), the `api.go` Route NOTE ("Bot API endpoints have been migrated to modules/bot_api/"), and the fact that the existing `api_bot_group_test.go` / `api_bot_thread_test.go` tests hit the routes and assert `ErrBotAPI*` (bot_api's codes), not botfather's.

Migrating that dead code added ~15 `ErrBotfather*` codes duplicating `ErrBotAPI*`, and the earlier review findings (getUserInfo 404 handling, the membership-guard `err`/business splits) were all in **unreachable** code. So instead of fixing dead code, this PR removes it.

## Changes
- **Migrate the live surface** to `httperr.ResponseErrorL` / `ResponseErrorLWithStatus`:
  - `api_user.go` (41 sites) — User Bot bind/unbind/token/create/update/delete; status-preserving 401/403/404/409, legacy 400 kept (D14).
  - `api_apply.go` (26 sites) — Robot Apply flow; DB error vs business-denial split (`err != nil` → logged `ErrBotfatherQueryFailed`; not-found/processed/exists → their business codes).
[...]

Issue #221, opened 2026-06-02
feat(i18n): email templates + lang-aware send chain (Part of #170)


an9xyz:
## Parent
Part of #170 (i18n — backend / octo-server).

## Dependency
Blocked on the Phase 1 email i18n mechanism (new `assets/email/{lang}/*.tmpl` + `lang`-aware send API). This issue both builds that mechanism and migrates the existing hardcoded templates.

## Scope — hardcoded Chinese templates + lang-less send API
- `modules/base/common/service_email.go`
  - `SendVerifyCode`: `subject := "Octo 验证码"` + Chinese HTML body (L109–116)
  - `Verify`: Chinese `errors.New("验证码无效！")` (L414)
  - `EmailService` methods (`SendVerifyCode` / `SendHTMLEmail` / `SendTransactionalHTML`) take no `lang`
- `modules/space/email_invite_template.go`: two hardcoded Chinese invite templates (create-space L52–67, join-space L70+)

## Tasks
- [ ] Add `assets/email/{zh-CN,en-US}/*.tmpl` + go:embed loader.
- [ ] Thread a `lang` param through the send chain; resolve recipient language via `ResolveUserLanguage(uid)` at call sites.
- [ ] Externalize verify-code + space-invite templates into per-lang `.tmpl` files.
- [ ] Append `?lang=<recipient.language>` to invite links.
- [ ] Update call sites: `modules/space/email_invite_sender.go:119`, `modules/user/api_emaillogin.go:68`, `modules/common/api_manager_system_setting.go:330`.

## Out of scope
- SMS verify-code text (`modules/base/common/service_sms.go`) — same issue, separate transport; track separately.

## Acceptance
- [ ] Verify-code + space-invite emails render in recipient language; fallback to `OCTO_DEFAULT_LANGUAGE`.
- [ ] `make i18n-extract-check` + `make i18n-lint` green.

PR #316, opened 2026-06-09
feat(i18n): localize BotFather outbound IM messages (#304)

### LINKED ISSUE #304: feat(i18n): localize BotFather interactive / outbound IM messages (Part of #170)
## Parent
Part of #170 (i18n — backend / octo-server).

## The gap
The i18n work so far covers three categories: **HTTP error envelopes** (`httperr.ResponseErrorL`, e.g. #279), **email templates** (#221), and the planned **push text / static HTML** (§2.3). BotFather's **interactive / outbound IM message content** — the text it sends *to the user over WuKongIM* — is a **fourth category that was never in scope**. It is all hardcoded Chinese, so on an `OCTO_DEFAULT_LANGUAGE=en-US` deployment a user DM-ing BotFather still gets Chinese.

This is **live** code (BotFather's DM command flow and lifecycle events), unrelated to the `/v1/bot/*` dead handlers removed in #279.

Example (the welcome message a new user receives):
```
👋 **欢迎来到 Octo！** 体验企业级的 Agent-Native 协作平台
...
我是 BotFather，帮你创建和管理 AI 机器人：
· /newbot — 创建新机器人
...
好的，请给你的机器人取一个名字：
```

## Scope (~150 outbound message strings, all sent via `h.reply()` / `SendMessage` → WuKongIM)
| File | What | Notes |
|---|---|---|
| `modules/botfather/welcome.go` | `DefaultWelcomeMessage` + Space-join welcome variants | sent from `handleUserRegisterEvent` / `handleSpaceMemberJoinEvent` |
| `modules/botfather/command.go` | **55 `h.reply()` command replies** | `/newbot`, `/help`, `/cancel`, list-bots, unknown-command, prompts, etc. — the whole conversational flow |
| `modules/botfather/api_apply.go` | apply-flow notifications | "申请已提交，等待Owner审批" / "有人申请使用你的AI\n用户: %s …" / "你的AI使用申请已通过！…" / "…被拒绝…" — several carry **`fmt.Sprintf` params** (bot name, applicant name) |
| `modules/botfather/friend_approve.go` | friend-approval DM replies + notifications | owner-facing permission/denial replies + applicant notification; some carry params; one leaks `%v` of an internal `err` into the DM (`"添加好友关系失败: %v"`) — close that while templatizing |

## Mechanism (no new infra — reuse the email i18n approach from #221)
`h.reply(toUID, content)` and the apply/approve senders all know the **recipient uid**, so the recipient language is resolvable exactly like email:
- Resolve language via `i18n.OutboundLanguage(ctx)` / `LanguageService.Resolve(recipientUID)` (the same path #221/#181 established).
- Move each message into a go-i18n message catalog / template set keyed by language (mirror `modules/base/common/emailtmpl`), with params for the interpolated values (bot name, user name) instead of `fmt.Sprintf` on a hardcoded string.
- This is the same **outbound-content** class as §2.3's push-text M-1 (webhook `PayloadInfo.Language`); worth designing one shared "outbound message template" mechanism and applying it to both BotFather and push.

## Tasks
- [ ] Decide the message-catalog shape (go-i18n `Message` entries under `msg.botfather.*`, or a per-lang template dir like `emailtmpl`) — align with the §2.3 push-text design so there is one outbound-content mechanism.
- [ ] Resolve recipient language at each send site (`reply` / apply / approve / welcome) via `i18n.OutboundLanguage` / `LanguageService`.
- [ ] Migrate `welcome.go`, `command.go` (55 replies), `api_apply.go` notifications, and `friend_approve.go` to the catalog; params for interpolated values.
- [ ] Stop leaking `err`/`%v` into user-facing DMs (`friend_approve.go` "添加好友关系失败: %v") — log the cause, send a generic localized message.
- [ ] en-US + zh-CN copy for every message; extractor/CI coverage if the catalog participates in `make i18n-extract`.

## Out of scope
- HTTP error responses (already migrated — #279 / §2.1).
- The dead `/v1/bot/*` handlers (removed in #279).

## Acceptance
- [ ] A user with `language=en-US` (or an en-US default deployment) receives BotFather's welcome + all command/apply/approval messages in English; zh-CN users still get Chinese.
- [ ] No user-facing DM contains a raw `err.Error()` / `%v` of an internal error.
- [ ] Recipient language is resolved per-recipient (not a process default) wherever the recipient uid is known.

an9xyz:
## Summary

Closes #304.

BotFather's interactive DM content sent over WuKongIM was all hardcoded Chinese, so an `OCTO_DEFAULT_LANGUAGE=en-US` deployment still served Chinese welcome / command / apply messages. This is the **fourth outbound-content i18n category**, after HTTP error envelopes (#279) and email templates (#221).

## What changed

- **New `modules/base/common/msgtmpl`** — a namespace-extensible outbound-message template package: one `text/template` tree per language built from `templates/{lang}/*.tmpl` (`{{define "key"}}` blocks), a **load-time completeness matrix** (every supported language must define the same keys, else fail loud at startup), and source-language fallback. Mirrors the `emailtmpl` loader pattern. The planned §2.3 push-text mechanism can reuse it via its own namespace, so there is **one shared outbound-content mechanism** rather than two parallel designs.
- **BotFather catalog wiring** (`messages.go`) — `replyL` renders by recipient language; `recipientLanguage` resolves the recipient's stored preference via `user.LanguageService`, falling back to `OCTO_DEFAULT_LANGUAGE`, and unwraps Space-prefixed uids via `extractRealUID`.
- **Migrated** `welcome.go`, `command.go` (~55 replies), `api_apply.go` notifications, and `friend_approve.go` to the catalog, with en-US + zh-CN copy for ~70 message keys.
- **Fixed two `%v` internal-error leaks** into user DMs (`friend_approve.go`): log the cause, send a generic localized message instead.

## Language resolution (two paths)

- **IM messages** (reply / notify / welcome — async, no request ctx) → per **recipient uid** preference.
- **HTTP response body** (success message + error envelope) → the **request's** negotiated language via `LanguageOrDefault(c.Request.Context(), DefaultLanguage)`, identical to the error renderer, honoring `?lang=` / cookie / `Accept-Language` / trusted header / user preference.

## Acceptance

- [x] A user with `language=en-US` (or an en-US default deployment) receives BotFather's welcome + all command / apply / approval messages in English; zh-CN users still get Chinese.
- [x] No user-facing DM contains a raw `err.Error()` / `%v`.
- [x] Recipient language is resolved per-recipient wherever the recipient uid is known.

## Test plan

- [x] `msgtmpl` unit tests (load-time matrix, source fallback, render, execute-error)
- [x] BotFather completeness test: every code key renders in every supported language, and the template set equals the declared keys (no orphan, no missing)
- [x] Parameter interpolation + relative-time plural tests
- [x] Full `modules/botfather` integration suite passes (incl. `TestRobotApply_*`, `TestBotfatherNoLegacyResponseError` guard)
- [x] `make i18n-lint` / `make i18n-extract-check` / `gofmt` / `go vet` pass


lml2468 (dismissed):
APPROVED (head `4e4887ce6e20`).

Substantive feature PR. Traced the full path — the new `msgtmpl` mechanism, BotFather catalog wiring, two distinct language-resolution paths (IM per-recipient vs HTTP per-request), the `%v` leak fix, and the completeness guarantees. Solid design with the right load-time invariants. No blocking issues.

## ✅ Independently verified

**`msgtmpl` is a correctly designed mechanism, not just a one-off helper.**
- `New(fsys, root)` builds one `text/template` tree per language under `{root}/{lang}/*.tmpl`, then enforces the **cross-language completeness matrix at construction** — every supported language must define exactly the same union of `{{define "..."}}` names, otherwise hard error. This is the right shape: a half-translated build CAN'T silently fall back to source language; it fails at startup so the deployer sees it before any user does.
- `text/template` (not `html/template`) chosen on purpose — IM bodies are Markdown/plain text, html-escaping would corrupt them. Doc comment explicitly calls this out. Correct.
- `MustNew` for the package-level singleton, `New` for tests — matches the `emailtmpl` loader pattern. Source-language presence is also enforced (`SourceLanguage not present in supported set` error).
- `Render` is defensive-only post-construction: unknown name → error (caller bug), execute error → propagated with file/lang context. The completeness matrix means "missing key in this language" is structurally impossible at runtime.
- `definedNames` correctly excludes the root template + per-file `*.tmpl` shadows that `ParseFS` creates, so what remains is exactly the logical message-name set.

**BotFather catalog wiring is tight.**
- `botMessages = msgtmpl.MustNew(templatesFS, "templates")` at package init — `//go:embed templates` ships the trees as build assets. A missing key fails the binary's startup, not a runtime DM.
- `allBotMessageKeys` is a Go-side declared list of all keys (~70). `TestBotMessagesCompleteness` asserts **two** invariants: (1) every declared key renders in every supported language with the probe data, (2) `botMessages.Names()` equals `allBotMessageKeys` (no orphan template, no missing code constant). This is the code↔template parity guarantee on top of `MustNew`'s cross-language matrix.
- All key constants (`MsgWelcome`, `MsgUnknownCommand`, …, `MsgFriendAgoMinutes`) are private package constants strung-keyed to the template `{{define}}` names. Call sites can't typo a key into a runtime missing-template.

**Two language-resolution paths are correctly distinguished.**
- **IM path** (`recipientLanguage`): used by `replyL` and the notify helpers. Resolves via `user.LanguageService.Resolve(ctx, realUID)` with a 2s timeout context; falls back to `OutboundLanguage(context.Background())` = `OCTO_DEFAULT_LANGUAGE`. The `extractRealUID(toUID)` unwrap correctly handles Space-prefixed uids (`s{space}_{uid}` → `{uid}`) so the preference key lands on the real user, not a Space-scoped variant.
- **HTTP success-body path** (`localizedMessage` in `api_apply.go`): uses `LanguageOrDefault(c.Request.Context(), DefaultLanguage)`, identical to the error renderer's negotiation chain (`?lang=` → cookie → `Accept-Language` → trusted header → user preference). This is the right choice — keeping success message and error envelope on the same endpoint in one language matches `pkg/i18n`'s behavior and is a documented decision (extensive doc comment on `localizedMessage`).

**`%v` / `err.Error()` user-DM leak fix verified.**
- Searched `friend_approve.go`, `command.go`, `api_apply.go` post-PR for `%v` / `err.Error()` / `reply.*Sprintf` patterns near reply calls — **all gone**. Previously `friend_approve.go:159,168` had `fmt.Sprintf("❌ 添加好友关系失败: %v", err)` going straight into a user DM; now both sites log the cause with `zap.Error(err)` and send `MsgFriendAddFailed`, a generic localized message. Good security/UX hygiene fix.

**Relative-time plural is correctly localized at the template layer.**
- `relativeAgo(lang, createdAt)` is a thin Go dispatcher that picks the right key (`friend_ago_days/hours/minutes/just_now`) and a numeric `N`; the template handles the plural form: `{{.N}} {{if eq .N 1}}day{{else}}days{{end}} ago` for en-US, `{{.N}}天前` for zh-CN (Chinese has no plural form). Cleaner than a Go-side switch on language, and adding a new language requires only template additions.

**Optional sections render correctly when empty.**
- `{{if .Remark}}...{{end}}` pattern is consistently used in `notify_owner_new_apply` and `friend_apply_notify` for the optional remark line. `TestNotifyOwnerNewApplyNoRemarkLine` pins that an empty remark produces no dangling "备注" / "Note:" line — i.e., the conditional actually fires. Same for `Desc` on `botListItem` (empty → template's localized "no description").

**Friend.AddedTipsText override preserved.**
- The `Friend.AddedTipsText` operator override (per-deployment hardcoded greeting) is still honored when set; localization only kicks in when the override is empty. Backward-compatible — existing deployments that customize this aren't affected.

**Test coverage shape is right.**
- `msgtmpl_test.go` (no infra): completeness matrix gap, missing-language-dir, unsupported-lang→source fallback, unknown name error, execute error propagation, `Names()` sort, `MustNew` panic. Covers all the construction-time invariants.
- `messages_test.go` (no infra): interpolation correctness (params land in body), conditional sections (empty Remark drops the line), relative-time plurals (1 day vs 3 days, en-US vs zh-CN), completeness in both directions (declared vs defined).
- `welcome_test.go` updated from "the const is not empty" to "render works for every supported language and keeps the stable anchors (BotFather, /newbot)" — strictly better.

1. **`recipientLanguage` swallows `Resolve` errors silently.** When `langSvc.Resolve` returns an error (Redis fault, DB outage), the function falls through to `OutboundLanguage(context.Background())` without any log. The 2s timeout means a DM never blocks waiting for Redis, which is right, but a chronically broken language service would manifest as "every user gets the default language" with no signal. Consider a `Debug` log on Resolve error (Warn would be too noisy under transient blips). Cosmetic.

2. **Empty render output is silently sent for HTTP success.** `localizedMessage` returns `""` on render error (logged but `c.Response({"status":"approved","message":""})`). For an HTTP success body this is mild — the status field still tells the client what happened. Worth noting as design intent in the comment: the message field is best-effort, the status field is authoritative. Doc-only.

3. **`replyL` skips the send entirely on render error.** This is documented behavior ("the completeness matrix + tests make this a build-time guarantee, so a runtime miss is a bug, not a user-facing blank message") and the right tradeoff — sending a blank DM is worse than sending nothing. But it means a render error during a stateful flow (e.g., `MsgQueryFailedRetry` itself failing to render) could leave the user without any feedback and stuck in a state. The completeness test makes this essentially impossible, but a `Warn`-level log + state cleanup would be belt-and-suspenders for a future "what happened to my message" forensic. Cosmetic.

4. **`Friend.AddedTipsText` operator override is plain-text, not localized.** Operators who want a friend-added greeting customized still get a single string regardless of recipient language. Documented gap — out of scope for this PR. If operators want per-language customization they can clear the override and rely on the new template (which is now localized).

5. **`localizedMessage` doesn't handle the source language as a typed fallback.** If `c.Request.Context()` has a wholly-unrecognized lang (e.g., `?lang=xx-XX`), `LanguageOrDefault` returns `DefaultLanguage`, which then routes through `Catalog.treeFor` → `MatchSupportedLanguage` again. Double-normalization is harmless, just a minor redundancy. Cosmetic.

## CI

`code-review` and the full CI matrix not yet reported at review time. Ride them to green; the test additions (`msgtmpl_test.go`, `messages_test.go`, updated `welcome_test.go`) are no-infra so should pass in the unit pass; `make i18n-extract-check` already confirmed in PR description.

## Verdict

Substantive, well-scoped feature. The new `msgtmpl` package is namespace-extensible (push-text will reuse it per the doc comment) so this is **one shared outbound-content mechanism** rather than yet another parallel design — exactly the right architectural call. Load-time completeness matrix + code↔template parity test makes "a missing translation reaches users" structurally impossible. The IM/HTTP language-resolution split is principled and matches the existing error-envelope behavior. The `%v` leak fix is a real security/UX improvement that rode along cleanly. LGTM, merge once CI confirms green.

Jerry-Xin (dismissed):
Project relevance gate: passed. This PR is in scope for `octo-server`; it localizes BotFather backend outbound IM and related HTTP success text.

💬 Non-blocking

🔵 Suggestion: `recipientLanguage` says it handles Space-prefixed UIDs, but it only strips the prefix when `spacePrefixes` was pre-populated for that exact UID. A direct async call with `s{space}_{uid}` would still look up the prefixed value and fall back to the default language. Current touched paths appear covered by bare UIDs or message-chain prefix setup, so this is not blocking. See `modules/botfather/messages.go:198` and `modules/botfather/command.go:109`.

🔵 Suggestion: consider `template.Option("missingkey=error")` in `msgtmpl.New` so future call sites cannot silently render `<no value>` when a map omits a template field. Current call data looks complete, and the catalog tests cover declared keys, so this is a hardening suggestion only. See `modules/base/common/msgtmpl/msgtmpl.go:73` and `modules/base/common/msgtmpl/msgtmpl.go:127`.

✅ Highlights

The shared `msgtmpl` loader is a good fit for the existing email-template pattern and correctly fails loud on cross-language key drift.

BotFather message keys are centralized and backed by a template/key parity test, which should keep future localization additions disciplined.

The PR removes raw internal error text from friend approval DMs and keeps HTTP error handling on the localized envelope path.

Verification: `go test ./modules/base/common/msgtmpl` passed. Focused BotFather template tests passed with `go test ./modules/botfather -run 'Test(BotMessagesCompleteness|KeyMessageInterpolation|NotifyOwnerNewApplyNoRemarkLine|WelcomeMessageRenders)$'`. Full `go test ./modules/botfather` could not complete because local MySQL on `127.0.0.1:3306` is unavailable.

lml2468 (dismissed):
APPROVED (head `1726852cf556`).

Directly addresses **both** non-blocking observations from R1 (Jerry-Xin's `missingkey=error` + UID-prefix doc + my `Resolve` error silent-swallow). 3-file delta on top of R1 head `4e4887ce6e20`; the other 14 PR files are unchanged.

## ✅ Independently verified

**`missingkey=error` correctly applied to every template.**
- `text/template` `Option("missingkey=error")` does NOT propagate from the root template to the `{{define}}` blocks accessed via `Lookup()` — it is per-`*Template`. The PR correctly iterates `t.Templates()` and sets the option on every associated template. The new doc comment explicitly calls out this subtlety. Easy to miss; got it right.
- New test `TestRenderMissingMapKeyErrors` pins the behavior **on a Lookup()'d block**, not just the root — exactly the right anchor for the per-Template semantics. If a future refactor switches to `t.Execute` instead of `t.Lookup(name).Execute`, this test would catch a regression to silent `<no value>`.

**Call-site / template compatibility verified — no silent breakage.**
Worried that `missingkey=error` would break the `{{if .Remark}}...{{end}}` optional sections (`text/template` treats `{{if .X}}` on a missing map key as a hard error too, not just `{{.X}}`). Audited the call sites:
- `MsgNotifyOwnerNewApply` / `MsgFriendApplyNotify` — call sites pass `"Remark": remark` (even when empty `""`), so the map always has the key.
- `MsgFriendPendingList` — `Items` is a slice of the **Go struct** `pendingApplyItem` (not `map[string]any`). `text/template`'s `missingkey` option **only triggers on map missingkeys, not struct field misses** — a struct without a `Remark` field would be a parse-time error long before, and `pendingApplyItem.Remark` is a declared field. Same for `botListItem.Num/Display/Desc`. Safe.
- `MsgWelcome` / `MsgUnknownCommand` / `MsgHelp` / etc — these templates have no `{{.Field}}` references at all, so `nil` data still renders fine. Safe.
- Commit message: "Full BotFather integration suite confirms every current call site passes complete data" — consistent with what I see in the source.

**`recipientLanguage` Debug-level Resolve error logging matches my R1 ask.**
- New `msgLog := log.NewTLog("BotFather")` is a package logger used by `recipientLanguage` (a package function with no receiver, so `h.Log` isn't available). Choice is right.
- `Debug` (not `Warn`) level chosen explicitly — keeps transient Redis/DB blips out of warn-level dashboards while still being grep-able when a chronically broken language service is suspected. Matches what I recommended in R1.
- Error is captured with `zap.String("uid", realUID), zap.Error(err)` — has both the user and the underlying cause for forensics.

**`extractRealUID` doc tightened correctly per Jerry-Xin's R1.**
The new comment explicitly enumerates:
- Message-processing path always pre-registers the Space prefix (so `extractRealUID` strips it correctly).
- Lifecycle events (welcome) pass bare uids (no registration needed, no stripping happens, still correct).
- The known footgun: any new async entry point that synthesizes `s{space}_{uid}` itself without registering would resolve preference against the prefixed string — flagged as a future maintainer concern, not a current bug. All current callers verified covered.

## Production code surface

Net code delta is small and targeted:
- `msgtmpl.go`: +9 lines (loop + Option call + 6-line doc comment)
- `messages.go`: +21/-3 (package logger + switch-style error handling + tightened doc)
- `msgtmpl_test.go`: +23 lines (one new test)

No template content changed. No call-site signature changed. No new fields on the data maps. Backward compatible.

## CI

`code-review` and full CI matrix not yet reported at review time. Ride them to green.

## Verdict

Cleanest possible follow-up — surgical, fully addresses the R1 observations, defensive about a `text/template` footgun (the per-`*Template` `Option` propagation gotcha), and pins the behavior with a Lookup-targeted test. LGTM, merge once CI confirms green.

Jerry-Xin (dismissed):
The PR is in scope and the outbound-message catalog approach fits the BotFather/i18n problem without introducing blocking regressions.

💬 Non-blocking

🟡 Warning: `langResolveTimeout` does not actually bound Redis or DB latency. `recipientLanguage` creates a 2s context in [modules/botfather/messages.go](/Users/mlamp/.openclaw/tmp/pr-review-i5gnbi1i/repo/modules/botfather/messages.go:210), but `LanguageService.Resolve` only checks `ctx.Err()` before calling `cache.Get` and `db.QueryLanguageByUID`; those calls do not receive the context in [modules/user/language_service.go](/Users/mlamp/.openclaw/tmp/pr-review-i5gnbi1i/repo/modules/user/language_service.go:83). A slow cache/DB can still stall BotFather replies. Not a merge blocker, but the comment/expectation should be tightened or the lower-level APIs should become context-aware.

🔵 Suggestion: Add a focused HTTP success-message language test for `localizedMessage`. The IM template coverage is strong, but the `/apply` success body path in [modules/botfather/api_apply.go](/Users/mlamp/.openclaw/tmp/pr-review-i5gnbi1i/repo/modules/botfather/api_apply.go:449) relies on request-context negotiation, which is distinct from recipient-language resolution.

✅ Highlights

The PR correctly gates template completeness at startup, uses `text/template` for plain IM/Markdown content, and has good catalog parity tests in [modules/botfather/messages_test.go](/Users/mlamp/.openclaw/tmp/pr-review-i5gnbi1i/repo/modules/botfather/messages_test.go:96).

The `%v` user-facing error leak in friend approval was removed and replaced with generic localized replies plus logs.

Verification: `go test ./modules/base/common/msgtmpl` passed, and the focused BotFather template tests passed. Full `go test ./modules/botfather` could not run locally because MySQL on `127.0.0.1:3306` was unavailable.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#316 Review Report — feat(i18n): localize BotFather outbound IM messages

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `msgtmpl` package design | ✅ | `modules/base/common/msgtmpl/msgtmpl.go` — immutable Catalog, completeness matrix at `New()`, source-language fallback via `treeFor()` |
| Completeness matrix guarantee | ✅ | `msgtmpl.go:96-103` — every supported language must define every name in the union, else `New()` returns error → `MustNew()` panics at startup |
| `missingkey=error` propagation | ✅ | `msgtmpl.go:82-85` — option set on every associated template (not just root), verified by `TestRenderMissingMapKeyErrors` |
| Language resolution (IM path) | ✅ | `messages.go:183-210` — `recipientLanguage()` resolves via `LanguageService.Resolve()` → Redis `user_language:{uid}` → DB → `OutboundLanguage()` fallback |
| Language resolution (HTTP path) | ✅ | `api_apply.go:449-457` — `localizedMessage()` uses `LanguageOrDefault(c.Request.Context(), DefaultLanguage)`, identical to error envelope path |
| `extractRealUID` correctness | ✅ | `command.go:110-116` — strips Space prefix only when registered via `setSpacePrefixForUID`; bare UIDs (welcome, lifecycle events) pass through unchanged |
| Template key/code parity | ✅ | `messages_test.go:TestBotMessagesCompleteness` — asserts `allBotMessageKeys` equals `botMessages.Names()` (no orphan, no missing) |
| `%v` error leak fix | ✅ | `friend_approve.go:157,166` — old code `fmt.Sprintf("❌ 添加好友关系失败: %v", err)` replaced with generic localized message; error logged server-side |
| Test coverage | ✅ | `msgtmpl_test.go` (7 tests), `messages_test.go` (3 tests), `welcome_test.go` (updated) |
| `BotFather.cmdHandler.langSvc` wiring | ✅ | `api.go:29,48` — `BotFather` struct has `cmdHandler *commandHandler`; `command.go:43` — `langSvc` initialized via `newBotLanguageService(ctx)` |

## 2. Findings

### P2-1: Silent message drop on render failure in `replyL` (~55 call sites)

**What:** `replyL()` (`messages.go:218-227`) logs and returns on render failure without sending any message. The user gets no response to their command.

**Diff-scope:** **New** — old code used hardcoded strings that never failed. The new code introduces a render step that can fail.

**Severity rationale:** P2, not P1. The completeness matrix (`MustNew` panics at startup if any key is missing in any language) + `TestBotMessagesCompleteness` (asserts code↔template parity) make runtime render failure extremely unlikely. A render failure would require a bug in `text/template` itself, not in template content. The design tradeoff (fail-loud at build time, fail-soft at runtime) is intentional and documented.

**Suggestion:** Consider a fallback render to the source language on render failure (e.g., try `SourceLanguage` if the resolved language fails), or send a generic "something went wrong" message instead of silent drop. Low priority given the build-time guarantees.

### P2-2: Empty content sent on render failure in `createFriendRelation` / `approveFriend`

**What:** In `api_apply.go:418-425` and `friend_approve.go:218-225`, when `AddedTipsText` is empty AND `botMessages.Render(MsgFriendAddedTip, ...)` fails, `content` stays `""` and an empty message payload is sent to WuKongIM.

**Diff-scope:** **New** — old code had a hardcoded fallback string (`"我们已经是好友了，可以愉快的聊天了！"`). The new code can produce empty content on render failure.

**Severity rationale:** P2. The completeness matrix makes render failure extremely unlikely. The friend relation is still created successfully; the message is just a tip. But this is a minor regression in the failure path (old code always had non-empty content).

**Suggestion:** Add a hardcoded fallback: `if content == "" { content = "✅" }` or similar, so the message is never empty even on render failure.

### P2-3: `relativeAgo` silently returns empty string on render failure

**What:** `friend_approve.go:relativeAgo()` discards render errors (`s, _ := botMessages.Render(...)`), returning `""`. This would produce `()` in the pending list output.

**Diff-scope:** **New** — old code used hardcoded strings (`"刚刚"`, `"%d天前"`, etc.).

**Severity rationale:** P2. Completeness matrix makes this extremely unlikely. The `()` output is cosmetic, not functional.

**Suggestion:** Return a non-localized fallback (e.g., `"?"`) on render failure.

### P2-4: `handlePending` / `sendBotSelectionList` silent failure on render error

**What:** `friend_approve.go:handlePending()` and `command.go:sendBotSelectionList()` return without sending anything when render fails. The user gets no response.

**Diff-scope:** **New** — old code used string concatenation that never failed.

**Severity rationale:** P2. Same as P2-1 — completeness matrix makes this extremely unlikely.

### P2-5: zh-CN colon inconsistency between `notify_owner_new_apply` and `friend_apply_notify`

**What:** The zh-CN `notify_owner_new_apply` template uses `备注: ` (ASCII colon + space) while `friend_apply_notify` uses `备注：` (fullwidth colon). The en-US templates are consistent (`Note: `).

**Diff-scope:** **Pre-existing** — the old Go code had the same inconsistency (`fmt.Sprintf("\n备注: %s", ...)` vs `fmt.Sprintf("\n备注：%s", ...)`). The templates faithfully preserve it.

**Severity rationale:** P2 nit. Cosmetic inconsistency, no functional impact.

## 3. Suggestions

1. **Fallback on render failure:** Consider adding a source-language fallback in `replyL` — if render in the resolved language fails, try `SourceLanguage` before dropping the message. This adds a safety net without weakening the build-time guarantee.
2. **Non-empty guarantee:** For `createFriendRelation` / `approveFriend`, add a hardcoded fallback when both `AddedTipsText` and render produce empty content.
3. **Colon normalization:** Pick one colon style (fullwidth `：` or ASCII `: `) for zh-CN templates and normalize across both apply notification templates.

## 4. Additional Observations

- **Security improvement:** The `%v` error leak fix in `friend_approve.go` (old code: `fmt.Sprintf("❌ 添加好友关系失败: %v", err)`) is a good security practice — internal error details should not be exposed to users.
- **`langResolveTimeout` (2s):** Good defensive design — prevents a slow Redis/DB from stalling DM sends. Falls back to default language on timeout.
- **`newBotLanguageService` nil-safety:** Returns nil when cache is not configured; `recipientLanguage` treats nil as "use default". Clean degradation for test contexts.
- **Template completeness test (`TestBotMessagesCompleteness`):** Strong build-time guarantee — asserts every declared key renders in every supported language AND the template tree equals the declared key set (no orphan, no missing). This is the key safety net that makes the P2 findings above acceptable.

## 5. Data Flow Trace

| Consumer | Data | Upstream Source | Verified? |
|----------|------|-----------------|-----------|
| `replyL(toUID, key, data)` | `lang` | `recipientLanguage(langSvc, toUID)` → `LanguageService.Resolve(ctx, realUID)` → Redis `user_language:{uid}` → DB `user.language` → `OutboundLanguage()` fallback | ✅ |
| `replyL` | `content` | `botMessages.Render(name, lang, data)` → `Catalog.Render()` → `template.Lookup(name).Execute(buf, data)` | ✅ |
| `localizedMessage(c, key)` | `lang` | `LanguageOrDefault(c.Request.Context(), DefaultLanguage)` — reads negotiated language from request ctx (set by i18n middleware) | ✅ |
| `sendWelcomeMessage(toUID, spaceID)` | `lang` | `recipientLanguage(bf.cmdHandler.langSvc, toUID)` — `toUID` is bare UID from registration event, `extractRealUID` passes through | ✅ |
| `notifyOwnerFriendApply(...)` | `toUID` | `bot.CreatorUID` — bare UID from DB, `extractRealUID` passes through; Space context set via `setSpaceIDFromPayload` before `replyL` | ✅ |
| `createFriendRelation(userUID, ...)` | `lang` | `recipientLanguage(bf.cmdHandler.langSvc, userUID)` — `userUID` is bare UID from friend request | ✅ |
| `relativeAgo(lang, createdAt)` | `key, n` | Computed from `time.Since(time.Unix(createdAt, 0))` — switch on duration selects template key and integer count | ✅ |
| `botMessages` (package var) | `Catalog` | `msgtmpl.MustNew(templatesFS, "templates")` — panics at init if completeness matrix fails | ✅ |

---

[Octo-Q] verdict: **APPROVE**

No P0/P1 findings. All 5 findings are P2 (design tradeoffs and minor nits). The PR is well-designed with strong build-time guarantees (completeness matrix, code↔template parity test), comprehensive test coverage, and clean language resolution. The silent-failure patterns on render error are acceptable given the build-time guarantees, though adding source-language fallbacks would strengthen the defensive posture.

yujiawei (approved):
# Code Review — PR #316 (octo-server)

## Verdict: APPROVED

This is a clean, well-structured fourth outbound-content i18n category. The new `msgtmpl` package mirrors the existing `emailtmpl` loader pattern, the load-time completeness matrix (`MustNew` → fail-loud at startup) is the right invariant, and the migration of ~70 BotFather message keys is thorough and well-tested. No correctness, security, or data-loss issues found. `go build`, `go vet`, the `msgtmpl` unit suite, and the BotFather template-render/completeness/interpolation tests all pass locally.

The two language-resolution paths are correct: IM sends resolve per-recipient stored preference (`recipientLanguage` → `LanguageService` → `OCTO_DEFAULT_LANGUAGE`), while the HTTP success body (`localizedMessage`) follows the request's negotiated language via `LanguageOrDefault(c.Request.Context(), DefaultLanguage)`, consistent with the error renderer. The request context is reliably populated by the globally-mounted `EarlyMiddleware`, so the success body and any error on the same endpoint stay in one language.

The stated `%v` internal-error leak fix in `friend_approve.go` is confirmed: the previous `fmt.Sprintf("❌ 添加好友关系失败: %v", err)` user DMs are gone; the cause is now logged via `h.Warn(..., zap.Error(err))` and the user receives a generic localized message.

## Strengths

- **Fail-loud completeness matrix.** `msgtmpl.New` enforces that every supported language defines the same key set; `messages_test.go` additionally asserts code↔template parity (`Names()` == `allBotMessageKeys`) so a half-translated build cannot ship. This is the correct place to catch drift.
- **`missingkey=error` set per-associated-template.** The comment correctly notes the option does not propagate from root to `Lookup()`'d `{{define}}` blocks, and `TestRenderMissingMapKeyErrors` guards exactly that subtlety. A forgotten parameter fails loudly rather than DMing `<no value>`.
- **`text/template` (not `html/template`) chosen deliberately** for Markdown/plain IM bodies — the rationale (avoiding `&amp;` corruption) is documented.
- **Immutable `Catalog`** is safe for concurrent `Render`, and `langResolveTimeout` bounds the per-recipient lookup so a slow Redis/DB never stalls a DM.
- EN/ZH plural handling: the `friend_ago_*` `{{if eq .N 1}}` inflection appears only in en-US, correctly — Chinese has no plural form. Verified this is the only structural divergence between the two trees and it does not affect the completeness matrix.

## Non-blocking suggestions (P2 — not required for merge)

These are maintainability/UX observations, not defects. None blocks merge.

1. **`friend_add_failed` drops the error detail from the owner DM.** `friend_approve.go:160,169` changed `❌ 添加好友关系失败: %v` → a static `❌ 添加好友失败，请稍后重试。`. The cause is still logged, so this is arguably an improvement (no internal text leaked), but it is a deliberate user-visible wording/detail change — please confirm it's intended.

2. **`relativeAgo` swallows render errors** (`friend_approve.go:419-423`): both `botMessages.Render(...)` calls use `s, _ :=`, returning an empty string on error with no log. Every other `Render` consumer in the package logs (`h.Error`/`bf.Error`) and skips, per the convention codified in `replyL`'s doc. On error a pending-list row would silently show `()`. Runtime error is practically impossible here (these are trivial keys covered by the matrix), so it's purely a consistency nit — a `Debug`-level log or a one-line comment explaining the exemption would close it.

3. **`treeFor` last-resort fallback is `SourceLanguage` (en-US), while the recipient/outbound default is `DefaultLanguage` (zh-CN)** (`msgtmpl.go:159-167`). For all current callers `lang` is pre-normalized to a supported tag before reaching `Render`, so this branch is effectively unreachable — not a bug. But the asymmetry is a latent trap if a future caller passes an un-normalized tag. Falling back to en-US (the guaranteed-present base tree) is a defensible convention shared with `pkg/i18n/localizer.go`'s `requested → fallback → SourceLanguage` chain; a one-line doc note ("callers must pre-normalize; this fallback is defensive and intentionally targets the source tree") would prevent confusion.

4. **`definedNames` uses `strings.HasSuffix(n, ".tmpl")`** to exclude `ParseFS`'s per-file templates (`msgtmpl.go:173-186`). A `{{define "foo.tmpl"}}` block would be silently dropped from both `Names()` and the completeness matrix. No key triggers this today and `Render` (which uses `Lookup` by full name) would still resolve it, so it's contained — but the implicit "message key must not end in `.tmpl`" constraint is worth one line in the doc comment.

5. **Completeness test coverage gap** (`messages_test.go:14-38,100-113`): `TestBotMessagesCompleteness` always renders with a fully-populated `renderProbe()` map, never the `nil` that ~40 real call sites pass (`replyL(..., key, nil)`). With `missingkey=error`, a future template edit that adds a `{{.Field}}` to a message whose call site passes `nil` would break at runtime (user gets no DM) yet still pass this test. Every nil-data key references zero fields today, so no live bug — but rendering nil-data keys with `nil` (or asserting which keys are parameterized) would close the gap.

## Summary

Solid, idiomatic, well-tested. The shared `msgtmpl` mechanism is a good foundation for the planned §2.3 push-text reuse. Approving; the items above are optional follow-ups.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#316 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/316
**Head SHA:** `9663a691e583c3c99e57936d7d42a9f0607e69d7`
**Scope:** feat(i18n): localize BotFather outbound IM messages (#304)
**Files:** 17 changed (+1476 / -219)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| `msgtmpl` package design | ✅ | `modules/base/common/msgtmpl/msgtmpl.go` — clean, namespace-agnostic, mirrors `emailtmpl` pattern |
| Completeness matrix (startup guarantee) | ✅ | `msgtmpl.go:91-100` — every supported language must define every key in the union, else `New` returns error |
| `missingkey=error` per-template option | ✅ | `msgtmpl.go:82-86` — applied to every associated template (not just root), with negative test `TestRenderMissingMapKeyErrors` |
| `recipientLanguage` nil-safety | ✅ | `messages.go:210-229` — `langSvc != nil` guard; falls back to `OutboundLanguage(context.Background())` = deployment default |
| `extractRealUID` bare-UID passthrough | ✅ | `command.go:110-116` — returns uid unchanged when no Space prefix registered in `spacePrefixes` sync.Map |
| HTTP vs IM language resolution split | ✅ | `api_apply.go:441-455` (`localizedMessage`) uses request ctx; `replyL` uses per-recipient uid — correct dual-path design |
| Template key ↔ code constant parity | ✅ | `messages_test.go:TestBotMessagesCompleteness` — asserts `allBotMessageKeys` == `botMessages.Names()` |
| Error leak fix (`friend_approve.go`) | ✅ | Old: `fmt.Sprintf("❌ 添加好友关系失败: %v", err)` → New: `replyL(ownerUID, MsgFriendAddFailed, nil)` — internal error no longer sent to user DM |
| Space-prefix routing in `notifyOwnerFriendApply` | ✅ | `friend_approve.go:361-364` — `setSpacePrefixForUID` called before `replyL`, defer cleanup present |
| `newBotLanguageService` nil guards | ✅ | `messages.go:188-193` — returns nil when `ctx == nil || ctx.Cache() == nil`, preventing `NewLanguageService` panic |

---

## 2. Findings

### F1 — P2 (new, low risk): `friend_added_tip` empty-content regression path

**Files:** `api_apply.go:415-422`, `friend_approve.go:217-224`

**Diff-scope:** NEW — introduced by this PR.

**Old behavior:** `content` always had a hardcoded Chinese fallback (`"我们已经是好友了，可以愉快的聊天了！"`).
**New behavior:** When `Friend.AddedTipsText` is empty AND `botMessages.Render(MsgFriendAddedTip, ...)` fails at runtime, `content` remains `""` and an empty message is sent via `SendCMD`.

**Why P2 not P1:** The completeness matrix enforced at startup (`MustNew` panics if any language misses a key) makes a runtime render failure for `MsgFriendAddedTip` virtually impossible — it would require a post-startup catalog corruption. The `MsgFriendAddedTip` template takes `nil` data, so no `missingkey=error` trigger is possible.

**Suggestion:** Add a defensive fallback: if `content == ""` after both branches, either skip the send or use a static fallback string. Low urgency.

---

### F2 — P2 (new, low risk): `relativeAgo` silently returns empty string on render error

**File:** `friend_approve.go:408-425`

**Diff-scope:** NEW — introduced by this PR.

```go
func relativeAgo(lang string, createdAt int64) string {
    // ...
    s, _ := botMessages.Render(key, lang, map[string]any{"N": n})
    return s  // empty string on error
}
```

If render fails, the pending list shows `(3)` instead of `(3 days ago)`. The completeness matrix makes this near-impossible, but the error is silently swallowed with no log signal.

**Suggestion:** Log the error at Debug level (consistent with `recipientLanguage`'s error handling pattern at `messages.go:220-223`). Low urgency.

---

### F3 — P2 (new, low risk): `handlePending` / `sendBotSelectionList` silent no-send on render error

**Files:** `friend_approve.go:389-399`, `command.go:912-922`

**Diff-scope:** NEW — introduced by this PR.

Both functions render a message and, if the render fails, log the error and return without sending anything to the user. Old behavior: always sent a (Chinese) message. New behavior: on render failure, the user gets no response at all.

Same completeness-matrix mitigation as F1/F2 applies. The functions do log the error, which is correct.

**Suggestion:** Consider a generic fallback message ("An error occurred, please try again") rather than silent no-send. Low urgency given the startup guarantee.

---

### F4 — P2 (pre-existing, not amplified): `spacePrefixes` sync.Map concurrent-same-UID race

**File:** `command.go:74` (package-level `sync.Map`)

**Diff-scope:** PRE-EXISTING — not introduced or amplified by this PR.

If the same `fromUID` is processed concurrently in two goroutines, the second `setSpacePrefixForUID` overwrites the first's entry, and the first's deferred cleanup deletes the second's entry prematurely. This affects `extractRealUID` → `recipientLanguage` correctness.

In practice, message-queue consumers typically process messages sequentially per-user, so this is theoretical. The PR's `replyL` path inherits the same risk profile as the old `reply` path.

**No action required for this PR.**

---

## 3. Suggestions

1. **Defensive fallback for empty content** (F1, F3): When render fails and no admin-configured text exists, either skip the send or use a static generic message. This closes the theoretical regression path.

2. **Log `relativeAgo` errors** (F2): Even at Debug level, so a chronically broken template (should never happen post-matrix, but defense-in-depth) is observable.

3. **Consider adding a `TestRelativeAgo` unit test**: The `relativeAgo` function has plural handling via template delegation (`MsgFriendAgoDays` with `{{if eq .N 1}}day{{else}}days{{end}}`). A direct unit test would pin the plural logic independently of the pending-list render. The interpolation test covers `N=1` and `N=3` for `MsgFriendAgoDays` but not the `relativeAgo` dispatcher itself.

---

## 4. Additional Observations

- **Architecture quality:** The `msgtmpl` package is well-designed — namespace-agnostic, with a clean separation between the rendering mechanism and the BotFather-specific catalog. The planned push-text mechanism can reuse it without modification.
- **Test coverage:** `TestBotMessagesCompleteness` provides a strong build-time guarantee (every code key renders in every supported language, template set equals declared keys). `TestKeyMessageInterpolation` validates parameter flow and optional sections. `TestRenderMissingMapKeyErrors` pins the `missingkey=error` behavior on Lookup'd blocks.
- **Error leak fix:** The two `%v` internal-error leaks in `friend_approve.go` (old: `fmt.Sprintf("❌ 添加好友关系失败: %v", err)`) are correctly replaced with generic localized messages. This is a security improvement.
- **Commit 2 (`9663a69`):** Addresses review feedback — `missingkey=error` per-template and tightened `recipientLanguage` docs. Both changes are correct and well-tested.

---

## 5. Data-Flow Trace

| Consumer | Data | Upstream Source | Verified? |
|----------|------|-----------------|-----------|
| `replyL(toUID, key, data)` | `lang` | `recipientLanguage(h.langSvc, toUID)` → `LanguageService.Resolve(ctx, realUID)` → Redis `user_language:{uid}` → DB `SELECT language FROM user WHERE uid=?` → fallback `OCTO_DEFAULT_LANGUAGE` | ✅ |
| `replyL` | `toUID` Space-prefix | `setSpacePrefixForUID` called in `messagesListen` → `HandleMessage` before dispatch; `extractRealUID` strips it for language lookup | ✅ |
| `localizedMessage(c, key)` | `lang` | `octoi18n.LanguageOrDefault(c.Request.Context(), DefaultLanguage)` — reads middleware-negotiated language from request ctx | ✅ |
| `botMessages.Render(name, lang, data)` | template tree | `//go:embed templates` → `msgtmpl.MustNew(templatesFS, "templates")` at init; completeness matrix enforced at construction | ✅ |
| `notifyOwnerNewApply` | `applicantName` | `userService.GetUserByUID(applicantUID)` → `.Name`; falls back to raw UID on error | ✅ |
| `relativeAgo` | `createdAt` | `apply.CreatedAt` (unix seconds from DB); `time.Since(time.Unix(...))` converts to duration | ✅ |
| `sendBotSelectionList` | `bot.RobotID` → `formatBotDisplay` | `h.formatBotDisplay(bot.RobotID)` — same function as old code, no change | ✅ |
| `approveFriend` friend-added-tip | `content` | `Friend.AddedTipsText` config → fallback `botMessages.Render(MsgFriendAddedTip, lang, nil)` | ✅ (with F1 caveat) |

---

## Verdict

All findings are P2 (no P0/P1). The PR is well-structured with strong build-time guarantees (completeness matrix + `missingkey=error` + comprehensive tests). The P2 findings are theoretical regression paths mitigated by startup-time enforcement. The error-leak fix is a genuine security improvement.

**[Octo-Q] verdict: APPROVE** — no blocking issues found. P2 suggestions are optional improvements.

yujiawei (dismissed):
# Code Review — PR #316 (octo-server)

**Scope:** `feat(i18n): localize BotFather outbound IM messages (#304)` — 17 files, +1476 / −219.
**Head SHA:** `9663a691e583c3c99e57936d7d42a9f0607e69d7`

This is the fourth outbound-content i18n category (after HTTP error envelopes and email templates). It introduces a reusable `msgtmpl` package and migrates BotFather's hardcoded Chinese DM content to per-language template sets.

## Summary

Approving. The design is clean and the safety properties are enforced at the right layer (startup), with strong test coverage. No P0/P1 issues. A handful of P2 defense-in-depth suggestions are listed below; all are optional and mitigated by the startup completeness guarantee.

Verified locally at the head SHA:
- `go build ./modules/base/common/msgtmpl/... ./modules/botfather/...` — passes
- `go vet` on both packages — clean
- `go test ./modules/base/common/msgtmpl/...` — passes
- `go test ./modules/botfather/ -run TestBotMessagesCompleteness|TestKeyMessageInterpolation|TestNotifyOwnerNewApplyNoRemarkLine|TestWelcomeMessageRenders` — passes

## What's done well

- **`msgtmpl` package** (`modules/base/common/msgtmpl/msgtmpl.go`): namespace-agnostic, mirrors the `emailtmpl` loader. The completeness matrix (`msgtmpl.go:97-110`) fails loud at construction if any supported language is missing a key, so a half-translated build cannot ship a blank or wrong-language DM. `text/template` (not `html/template`) is the correct choice for Markdown/plain IM bodies.
- **`missingkey=error` set per associated template** (`msgtmpl.go:83-85`), with the subtle correctness note that the option does not propagate from the root to `Lookup()`'d `{{define}}` blocks. This is pinned by `TestRenderMissingMapKeyErrors`, which specifically exercises the Lookup'd block — good catch and good test.
- **Dual-path language resolution.** IM sends resolve per recipient uid (`recipientLanguage` → `LanguageService.Resolve` → Redis `user_language:{uid}` → DB → `OCTO_DEFAULT_LANGUAGE`), while the HTTP success body uses the request's negotiated language (`localizedMessage`, `api_apply.go:449-457`, identical path to the error renderer). This split is correct: an async DM has no request ctx, and a success body should match the same request's error envelope language.
- **Nil-safety throughout.** `newBotLanguageService` returns nil when no cache is configured (`messages.go:188-193`), and `recipientLanguage` treats nil as "use deployment default" rather than panicking — important because `NewLanguageService` panics on a nil cache by contract.
- **Security: error-leak fix.** The two `%v` internal-error leaks into user DMs in `friend_approve.go` (old `fmt.Sprintf("❌ 添加好友关系失败: %v", err)`) are replaced with a logged cause + generic localized `MsgFriendAddFailed`. Confirmed no remaining `reply(... %v / err.Error())` patterns in the changed files.
- **Tests.** `TestBotMessagesCompleteness` asserts code↔template parity (`allBotMessageKeys` == `Names()`) on top of the init-time cross-language parity, and `TestKeyMessageInterpolation` covers parameter flow and the optional-remark section.

**P2-1 — Empty tip on render failure (`api_apply.go:415-424`, `friend_approve.go:217-226`).** When `Friend.AddedTipsText` is unset and `Render(MsgFriendAddedTip, …)` fails, `content` stays `""` and an empty Tip is sent via `SendMessage`. The old code always had a hardcoded fallback. Near-impossible in practice — `MsgFriendAddedTip` takes `nil` data so `missingkey=error` cannot trigger, and the startup matrix guarantees the key resolves — but a tiny `if content == "" { return }` (or a static fallback) would fully close the regression path.

**P2-2 — `relativeAgo` swallows render errors silently (`friend_approve.go:407-424`).** On render error it returns `""` (the `s, _ :=` discards the error), so the pending list would show `(3)` instead of `(3 days ago)` with no log signal. Consider a Debug-level log to match the pattern already used in `recipientLanguage` (`messages.go:220-223`).

**P2-3 — Silent no-send on render failure (`friend_approve.go:393-401`, `command.go:909-923`).** `handlePending` / `sendBotSelectionList` log and return without replying if a render fails, where the old code always sent something. The logging is correct; a generic fallback message would be friendlier, but the startup guarantee makes this path unreachable in practice.

**P2-4 (pre-existing, not introduced here) — `spacePrefixes` same-uid race (`command.go:74`).** If the same `fromUID` is processed concurrently, a second `setSpacePrefixForUID` can overwrite the first and its deferred cleanup can delete the other's entry early, affecting `extractRealUID` → language lookup. The PR inherits the existing `reply` path's risk profile and does not amplify it; the per-recipient async resolution is documented in `messages.go:203-209`. Out of scope for this PR.

## Suggestions (optional)

- Add a direct `TestRelativeAgo` unit test to pin the plural dispatch (`MsgFriendAgoDays/Hours/Minutes` with `N==1` vs `N>1`) independently of the pending-list render.
- For P2-1/P2-3, a one-line `content == ""` guard or generic fallback turns a theoretical empty/no-send into a graceful path.

Jerry-Xin (approved):
The PR is in scope for octo-server and cleanly localizes BotFather outbound IM content without introducing blocking issues.

🔴 Blocking

None.

💬 Non-blocking

🟡 Warning: I could not run the full `go test ./modules/botfather` package because the local test MySQL dependency is unavailable (`127.0.0.1:3306: connect: connection refused`). Targeted non-DB BotFather tests and the new `msgtmpl` package tests passed.

✅ Highlights

🔵 The new `msgtmpl` loader enforces cross-language template completeness at construction time, which is the right failure mode for outbound content catalogs. See [msgtmpl.go](/Users/mlamp/.openclaw/tmp/pr-review-i6aalquh/repo/modules/base/common/msgtmpl/msgtmpl.go:61).

🔵 `missingkey=error` plus render-time error propagation prevents partially rendered DMs when a call site omits template data. See [msgtmpl.go](/Users/mlamp/.openclaw/tmp/pr-review-i6aalquh/repo/modules/base/common/msgtmpl/msgtmpl.go:83).

🔵 Recipient-language resolution is separated from HTTP response-language resolution, matching the PR’s intended two-path model. See [messages.go](/Users/mlamp/.openclaw/tmp/pr-review-i6aalquh/repo/modules/botfather/messages.go:210) and [api_apply.go](/Users/mlamp/.openclaw/tmp/pr-review-i6aalquh/repo/modules/botfather/api_apply.go:455).

🔵 Template/key parity and parameter interpolation are covered by focused tests, including the declared-key versus template-name check. See [messages_test.go](/Users/mlamp/.openclaw/tmp/pr-review-i6aalquh/repo/modules/botfather/messages_test.go:130).

Verification run:
`go test ./modules/base/common/msgtmpl` passed.
`go test ./modules/botfather -run 'Test(BotMessagesCompleteness|KeyMessageInterpolation|NotifyOwnerNewApplyNoRemarkLine|RelativeAgo|WelcomeMessageRenders)$'` passed.
`go test ./modules/botfather -run TestBotfatherNoLegacyResponseError` passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#316 Review Report
**PR:** feat(i18n): localize BotFather outbound IM messages (#304)
**Head SHA:** 45a3b29ebc6143725906b1d2778dbbcd1a9b8833
**Scope:** 17 files, +1541 / -241 lines

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| `msgtmpl.Catalog` completeness matrix | ✅ | `msgtmpl.go:88-97` — every supported language must define every name in the union; fail-loud at construction |
| `missingkey=error` propagation | ✅ | `msgtmpl.go:79-82` — option set on every associated template (per-*Template, not root-only) |
| Source-language fallback | ✅ | `msgtmpl.go:155-162` (`treeFor`) — unsupported/empty lang falls back to `SourceLanguage` |
| `recipientLanguage` nil-safety | ✅ | `messages.go:213-230` — nil `langSvc` → `OutboundLanguage(ctx)` → `OCTO_DEFAULT_LANGUAGE` |
| `extractRealUID` with bare UIDs | ✅ | `command.go:108` — bare UID from DB passes through unchanged when `spacePrefixes` sync.Map has no entry |
| `LanguageService.Resolve` data flow | ✅ | `language_service.go:82` — Redis → DB → normalize; never returns unsupported tag |
| Template/code key parity | ✅ | `messages_test.go:TestBotMessagesCompleteness` — asserts `allBotMessageKeys` == template names |
| `%v` error leak fix | ✅ | `friend_approve.go:157,166` — old `fmt.Sprintf("❌ ...: %v", err)` replaced with generic `MsgFriendAddFailed` |
| Two-path language resolution | ✅ | IM → per-recipient via `recipientLanguage`; HTTP → per-request via `LanguageOrDefault(c.Request.Context(), ...)` |
| Test coverage | ✅ | `msgtmpl_test.go` (7 tests), `messages_test.go` (4 tests incl. completeness, interpolation, relative-ago plural) |

## 2. Findings

### P2-1: Silent notification drop on render failure in async paths

**Diff-scope:** new (introduced by this PR's error-handling pattern)

**Locations:**
- `api_apply.go:478-481` (`notifyOwnerNewApply`) — `return` on render failure skips IM notification entirely
- `api_apply.go:504-507` (`notifyApplicantResult`) — same pattern, applicant never learns the result
- `friend_approve.go:395-398` (`handlePending`) — `return` on render failure leaves user with no response to `/pending`

**Analysis:** The completeness matrix (`MustNew` at init) makes render failure structurally unreachable for all declared keys with correct data maps. However, the "log + return" pattern means a future template regression (e.g. a new field added to a template but not the data map) would silently eat notifications rather than falling back to a generic localized message or the source-language render.

The old code always sent a hardcoded Chinese message — degraded (wrong language) but never silent. The new code is correct-language-or-nothing.

**Recommendation:** Consider a source-language fallback in `replyL` / `botMessages.Render` callers that are fire-and-forget notifications:
```go
content, err := botMessages.Render(key, lang, data)
if err != nil {
    // fallback to source language rather than silent drop
    content, err = botMessages.Render(key, octoi18n.SourceLanguage, data)
    if err != nil { return }
}
```
This is a defensive-improvement suggestion, not a blocker.

### P2-2: `localizedMessage` returns empty string on render failure

**Diff-scope:** new

**Location:** `api_apply.go:460-466`

```go
func (bf *BotFather) localizedMessage(c *wkhttp.Context, key string) string {
    ...
    if err != nil {
        bf.Error(...)
        return ""  // <-- empty string in HTTP success body
    }
}
```

Used in `robotApply` HTTP responses:
```go
c.Response(map[string]interface{}{
    "status":  "approved",
    "message": bf.localizedMessage(c, MsgApplyAutoApproved),
})
```

On render failure, the client receives `"message": ""`. The old code returned the hardcoded Chinese string. Same structural-unreachability argument as P2-1 applies.

**Recommendation:** Return a generic fallback (e.g. source-language render or a static "OK" string) rather than empty.

### P2-3: `relativeAgo` returns empty string on render failure

**Diff-scope:** new

**Location:** `friend_approve.go:413-420`

```go
s, err := botMessages.Render(key, lang, data)
if err != nil {
    msgLog.Debug(...)
}
return s  // empty on error
```

The pending list would show empty time phrases. Logged at Debug (not Warn), so chronic failures would be invisible in production. The completeness matrix makes this unreachable, but the log level should be Warn for a render failure in a structurally-guaranteed path.

**Recommendation:** Bump log level to Warn.

## 3. Suggestions

1. **Source-language fallback for fire-and-forget notifications** (see P2-1). The `replyL` helper could attempt source-language render before giving up, keeping the "never silent" guarantee of the old code while gaining localization.

2. **`relativeAgo` log level** — Debug → Warn for render failures that should be structurally impossible.

3. **Consider `localizedMessage` fallback** — return source-language string or a static OK message rather than empty on error.

## 4. Extra Findings

- **Security improvement (positive):** The PR fixes two `%v` internal-error leaks into user DMs (`friend_approve.go:157,166`). Old code: `fmt.Sprintf("❌ 添加好友关系失败: %v", err)` exposed internal error details to end users. New code sends a generic localized message. Good.

- **`newBotLanguageService` nil-guard** (`messages.go:199-203`): Correctly returns nil when `ctx.Cache()` is nil, and `recipientLanguage` handles nil `langSvc` gracefully. No panic risk.

- **`notifyOwnerFriendApply` creates a fresh `commandHandler` per call** (`friend_approve.go:337`): Each invocation creates a new `LanguageService` instance. The Redis cache is shared, so the per-instance overhead is minimal (just the Go struct + 5-min TTL wrapper). No concern.

- **Template `text/template` vs `html/template` choice** (`msgtmpl.go:28`): Correct — IM bodies are Markdown/plain text, not HTML. `html-escaping` would corrupt `&` → `&amp;`.

## 5. Data-Flow Trace

|---|---|---|
| `recipientLanguage(langSvc, toUID)` → lang tag | `LanguageService.Resolve(ctx, realUID)` → Redis `user_language:{uid}` → DB `user.language` → `MatchSupportedLanguage` normalize → fallback `OutboundLanguage(bgCtx)` → `OCTO_DEFAULT_LANGUAGE` | ✅ Yes. Nil `langSvc` → default. Unsupported DB value → `""` → default. |
| `extractRealUID(toUID)` → bare UID | `spacePrefixes` sync.Map (populated by `setSpacePrefixFromChannelID` during IM handling). Bare DB UIDs pass through unchanged. | ✅ Yes. `notifyOwnerFriendApply` runs in detached goroutine with bare `bot.CreatorUID` from DB — correct. |
| `botMessages.Render(key, lang, data)` → rendered string | `msgtmpl.Catalog` built at init from `//go:embed templates` FS. Completeness matrix enforced at `MustNew`. | ✅ Yes. All 63 keys defined in both `en-US` and `zh-CN`. `missingkey=error` catches missing data fields. |
| `localizedMessage(c, key)` → lang from request ctx | `LanguageOrDefault(c.Request.Context(), DefaultLanguage)` — reads `LanguageDecision` from i18n middleware (priority: trusted_header > query > cookie > user > accept_language > default) | ✅ Yes. HTTP success messages follow the same language as error envelopes on the same endpoint. |
| `relativeAgo(lang, createdAt)` → time phrase | `time.Since(time.Unix(createdAt, 0))` → bucket dispatch (days/hours/minutes/just-now) → template render with `{"N": int}` | ✅ Yes. `int` type matches `{{if eq .N 1}}` in template (Go `text/template` compares same-type values). |
| `Friend.AddedTipsText` → operational override | `bf.ctx.GetConfig().Friend.AddedTipsText` — if non-empty, overrides localized template. | ✅ Yes. Both `createFriendRelation` and `approveFriend` check this first. |

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. The PR is well-structured with strong build-time guarantees (completeness matrix, `missingkey=error`, key-constant drift prevention). The three P2 findings are defensive-improvement suggestions around error-path fallback behavior — none affect correct operation under normal or foreseeable conditions. The `%v` error leak fix is a positive security improvement.

yujiawei (approved):
# Code Review — PR #316 (octo-server)

## Verdict: APPROVED

A well-engineered i18n migration. The new `msgtmpl` package is a clean, reusable mechanism with a fail-loud load-time completeness matrix, and the BotFather wiring is faithful to the original behavior. I checked out the head SHA, read the full diff plus the surrounding code, verified the build, and ran the test suite. No P0/P1 (blocking) issues. The items below are P2 suggestions only — none block merge.

## 1. Verification results

- ✅ **Build**: `go build ./modules/botfather/... ./modules/base/common/msgtmpl/...` passes.
- ✅ **New unit tests pass**: `msgtmpl` package and BotFather pure-unit tests (`TestBotMessagesCompleteness`, `TestKeyMessageInterpolation`, `TestRelativeAgo`, `TestNotifyOwnerNewApplyNoRemarkLine`, `TestWelcomeMessageRenders`) all green.
- ⚠️ **Integration suite caveat**: `go test ./modules/botfather/...` fails at `TestRobotApply_*` with `unknown migration ... 20260607000001_user_secret_alias.sql`. This migration belongs to the `usersecret` module and is **not part of this PR's diff** — it is a pre-existing environment/migration mismatch, not a defect introduced here. The PR's own logic is covered by the passing unit tests.
- ✅ **Completeness matrix**: `msgtmpl.New` enforces cross-language key parity at startup (`MustNew` panics on a half-translated build), and `missingkey=error` is correctly propagated to every associated `{{define}}` template (not just the root) at `msgtmpl.go:84-91`.
- ✅ **Call-site ↔ template field parity**: I enumerated every `replyL`/`Render` call site's data map against the `{{.Field}}` refs in **both** en-US and zh-CN templates (`connect_prompt`, `created_prompt`, `mybots_list`, `bot_selection_list`, `friend_pending_list`, `notify_owner_new_apply`, `token_display`, `friend_apply_notify`, and all single-field sites). Every one matches — no runtime `missingkey=error` risk at any current call site.
- ✅ **`%v` leak fix**: The two `fmt.Sprintf("❌ 添加好友关系失败: %v", err)` leaks in `friend_approve.go` are replaced by `replyL(ownerUID, MsgFriendAddFailed, nil)`; the cause is still logged via `zap.Error(err)`. No remaining `%v`/`err.Error()` reaches any user-facing DM in the diff.
- ✅ **`welcome.go` new error path**: Both callers (`handleUserRegisterEvent`, `handleSpaceMemberJoinEvent`) handle the new `error` return correctly — they log, `commit(nil)`, and return **before** writing the welcome-sent marker, so a (matrix-impossible) render failure simply allows a later retry. No welcome-sent-TTL regression.
- ✅ **Language resolution**: `recipientLanguage` correctly bounds the lookup with a 2s timeout, degrades gracefully on `nil` langSvc / empty cache, strips Space prefixes via `extractRealUID`, and falls back to `OCTO_DEFAULT_LANGUAGE`. The HTTP success path (`localizedMessage`) uses the request-negotiated language via the **same** `LanguageOrDefault(c.Request.Context(), DefaultLanguage)` call the error envelope uses, so success and error bodies on `/robot/apply` cannot diverge in language.
- ✅ **Token handling / no new secret exposure**: `token_display` / `token_revoked` / connect/created prompts render the bot token into the owner's own authenticated DM — identical to pre-PR behavior. No `zap` field added by this PR logs a token value; test fixtures use a fake `"probe"` token.

## 2. Issues found

No P0 or P1 issues.

### P2 — suggestions (non-blocking)

1. **Two notions of "default language" diverge (latent)** — `msgtmpl.go:161-168`. `treeFor` falls back to `octoi18n.SourceLanguage` (`en-US`) for an unmatched tag, whereas `recipientLanguage`'s terminal fallback resolves to `OCTO_DEFAULT_LANGUAGE` → `DefaultLanguage` (`zh-CN`). This is **inert today** — `LanguageService.Resolve` only ever returns a `MatchSupportedLanguage`-normalized tag or `""`, and the final fallback (`zh-CN`) is itself supported, so `treeFor`'s `en-US` branch is unreachable via this path. But the two layers disagree on what "fallback" means; a one-line comment reconciling them (or routing both through the same default) would prevent a future caller from silently getting `en-US` bodies on a `zh-CN` deployment.

2. **Operator `Friend.AddedTipsText` override bypasses localization** — `api_apply.go:415-417`, `friend_approve.go:217-219`. Precedence is unchanged from pre-PR (config wins over the localized `friend_added_tip`), so an operator who configures a Chinese `AddedTipsText` will serve it verbatim to en-US recipients. This is a reasonable design choice and matches the inline intent, but it's the one user-visible localization gap and it isn't documented operator-facing. A one-line note on the config field ("override is language-agnostic / disables per-recipient localization for the friend tip") would help operators.

3. **Completeness test uses a single all-fields probe** — `messages_test.go` `renderProbe()`. Because `missingkey=error` only fires on an *absent* referenced key, rendering every key with a superset probe can never catch a call site that omits a field its template uses. All current call sites are correct (verified manually), but the automated guard won't protect future edits. Consider a render test that feeds each key its real call-site map so CI catches a dropped field rather than it surfacing as a skipped DM at runtime.

4. **`notifyOwnerFriendApply` builds a fresh `commandHandler` per hook** — `friend_approve.go:331-332`. Each friend-apply callback calls `newCommandHandler(ctx)`, which spins up a new `LanguageService` plus group/app/apiKey sub-services. It's correct and race-free (shared Redis cache keeps preferences consistent), but it churns a full handler graph on a potentially frequent callback. A cached/shared handler would avoid the allocation. Efficiency nit only.

5. **Selection-path delete-confirm wording consolidated** — `templates/zh-CN/command.tmpl` `delete_confirm`. The pre-PR code had two slightly different zh strings (single-bot path with "机器人", selection path without). Both now map to one key that includes "机器人". This is a benign consistency improvement, just flagging the minor user-visible text change for awareness.

## 3. Recommendations

- Add a short comment in `msgtmpl.go` `treeFor` (or unify the fallback) reconciling `SourceLanguage` vs the deployment default, so the divergence is documented as intentional.
- Document the `Friend.AddedTipsText` language-agnostic override on the config field / deployment docs.
- (Optional, future-proofing) Strengthen the completeness test with per-call-site data maps.

## 4. Additional observations

- **Pre-existing (not introduced by this PR)**: `friend_apply_notify` / `friend_pending_list` interpolate unsanitized user-controlled `ApplyName`/`Remark` next to literal `/approve <uid> <id>` command lines, and `text/template` does not escape. A remark containing a newline + `/approve …` could render an extra owner-attributed command line or break the Markdown. The old `fmt.Sprintf` code embedded the exact same unsanitized inputs next to the same command literals, so this PR is behavior-preserving here — it neither introduces nor fixes the issue. If hardening is desired, it belongs in the apply-name/remark validation in `modules/user` (strip control chars / cap length), not in this i18n change. Noted for a future ticket.
- Nice touch: `relativeAgo` keeps plural handling inside the per-language templates (en-US `{{if eq .N 1}}`, zh-CN no plural), and the bucket boundaries in the Go switch line up correctly with `Truncate(time.Minute)`. Tests cover both singular/plural and each bucket.

yujiawei (approved):
# Code Review — PR #316 (octo-server)

## Verdict: APPROVED

A clean, well-factored fourth outbound-content i18n category (parallel to error codes #279 and email templates #221). I traced the full path — the new `msgtmpl` mechanism, the BotFather catalog wiring, both language-resolution paths, and every `reply → replyL` call site. No P0/P1 issues. Notes below are non-blocking.

## 1. What was verified

- ✅ **`msgtmpl` mechanism is sound.** `New` enforces a real completeness matrix: every supported language must define exactly the same set of `{{define}}` names, an empty language dir errors, and the source language must be present (`modules/base/common/msgtmpl/msgtmpl.go:62-130`). `missingkey=error` is correctly set on *every* associated template, not just the root, with a test that specifically guards the `Lookup()`'d block (`msgtmpl.go:90-93`, `msgtmpl_test.go:117-143`). `text/template` (not `html/template`) is the right call for Markdown/plain IM bodies — documented at `msgtmpl.go:30-33`.
- ✅ **Startup fail-loud.** `MustNew` panics on an incomplete catalog (`messages.go:36`), so a half-translated build cannot ship a blank or wrong-language DM. Covered by `TestMustNewPanicsOnIncompleteCatalog`.
- ✅ **Code↔template parity.** `allBotMessageKeys` (`messages.go:140-160`) is asserted against `Catalog.Names()` in `TestBotMessagesCompleteness` (`messages_test.go:147-179`) — no orphan template, no missing key, and every key renders in every language with a probe payload.
- ✅ **Two distinct language-resolution paths, both correct and intentional.**
  - IM sends → `recipientLanguage` resolves per-recipient stored preference (LanguageService → `user_language:{uid}` → DB), falling back to `OCTO_DEFAULT_LANGUAGE` (`messages.go:175-205`). Bounded by a 2s timeout so a slow Redis/DB never stalls a DM.
  - HTTP success bodies → `localizedMessage` follows the *request's* negotiated language via `LanguageOrDefault(c.Request.Context(), DefaultLanguage)` (`api_apply.go:449-457`) — identical to the error-envelope path in `pkg/i18n/renderer.go:35`, so a success message and any error on the same endpoint stay in one language. Good call keeping these separate and documenting why.
- ✅ **Cross-file UID handling (the usual blind spot) is correct.** `recipientLanguage → extractRealUID` strips the Space prefix only when a goroutine pre-registered it via `setSpacePrefixForUID`. The message-processing path registers it; the notify/lifecycle paths pass bare UIDs (`loginUID`/`CreatorUID`/`applyUID` from `GetLoginUID()` and DB rows in `api_apply.go` / `friend_approve.go`), which `extractRealUID` returns unchanged. No path hands in an unregistered `s{space}_{uid}`, so there is no silent default-language fallback. The caveat is explicitly documented for future async entry points (`messages.go:188-198`).
- ✅ **Rendered output matches the legacy `fmt.Sprintf` bodies.** I rendered `mybots_list`, `bot_selection_list`, `friend_pending_list`, `connect_prompt`, `created_prompt`, `token_revoked`, and the notify messages in both languages and diffed against the strings they replace — formatting (numbering, `**bold**`, code fences, the empty-description fallback, the conditional `Note:`/`备注` line) is preserved.
- ✅ **Blank-Tip guard.** Both friend-tip paths now skip the send when `content == ""` after a render failure (`api_apply.go:412-420` region, `friend_approve.go:239-253`), so a render error drops only the cosmetic greeting instead of pushing an empty Tip. The friend relation / whitelist / CMD sync all run before this and are unaffected.
- ✅ **Plural correctness.** `relativeAgo` buckets guarantee `N >= 1` in every parameterized branch, and the en-US templates handle singular/plural via `{{if eq .N 1}}` (`friend.tmpl`). `TestRelativeAgo` pins singular/plural for en-US and the no-plural zh-CN form across all buckets including just-now.
- ✅ **Build + targeted tests green.** `go build` and `go vet` clean on `modules/botfather/...` and `modules/base/common/msgtmpl/...`; all i18n unit tests pass (`msgtmpl`, `TestBotMessagesCompleteness`, `TestKeyMessageInterpolation`, `TestRelativeAgo`, `TestWelcomeMessageRenders`).

## 2. Findings

No P0 / P1.

### P2 / non-blocking

- **P2 — `localizedMessage` returns `""` on render error**, which would surface as `"message": ""` in the HTTP success body (`api_apply.go:451-456`). This is structurally unreachable given the startup completeness matrix and it is logged, so it is purely defensive. Optional: fall back to the source-language render or a static string rather than an empty body. Same pattern (logged, returns empty) applies to `relativeAgo` (`friend_approve.go`), where an empty Ago in a list row is equally harmless.

### Nits

- The `mybots_list` and `friend_pending_list` templates emit a trailing blank line from the `{{range}}` block. This mirrors the original `strings.Builder` output and is cosmetically invisible in a chat client — no change needed, just noting it's intentional.
- `messages.go` carries a large hand-maintained `allBotMessageKeys` slice. The completeness test makes drift a build-time failure, so this is safe; nothing to change.

## 3. Suggestions

None required for merge. If the team later adds a third supported language, the matrix check at startup will force the new `templates/{lang}/` tree to be complete before the binary can boot — exactly the desired behavior.

## 4. Additional observations

- The decision to localize the friend-added tip only when `Friend.AddedTipsText` is unset preserves the operator-override contract (`api_apply.go` / `friend_approve.go`) — operator config still wins over the localized default. Correct precedence.
- Removing the `DefaultWelcomeMessage` constant is fully cleaned up; no remaining references in the tree, and `welcome_test.go` was updated to assert the rendered welcome keeps its stable anchors (`BotFather`, `/newbot`) in every language.

PR #317, opened 2026-06-09
feat(opanalytics): add trend and message composition

### LINKED ISSUE #318: feat(opanalytics): add dashboard trend and message composition
## Background

The opanalytics P0 data warehouse foundation is already available. The next small P1A step is to unblock dashboard charts using existing aggregated tables only, without changing ETL, migrations, or `content_type` semantics.

## Scope

- Extend `GET /v1/manager/dashboard/overview` with `message_composition`.
- Add `GET /v1/manager/dashboard/trend` with `granularity=day|week`.
- Use existing facts only:
  - `octo_fact_channel_daily` for message counts, active groups, private active count, and conv_type distribution.
  - `octo_fact_member_channel_daily` joined with current `octo_dim_member` for active human/agent members.
- Keep private chats global-only: when `space_ids` is provided, private metrics and conv_type 3/4 composition should be zero.
- Update Swagger and tests.

## Acceptance Criteria

- `/overview` returns fixed conv_type 1-4 `message_composition` entries, with missing data filled as zero.
- `/trend?granularity=day` returns zero-filled daily buckets for the requested date range.
- `/trend?granularity=week` buckets by Monday in the report timezone, clips bucket `start_date`/`end_date` to the request range, and counts active members distinct within each bucket.
- Non-superAdmin access to `/trend` returns the existing opanalytics forbidden i18n error envelope.
- Invalid date range or invalid granularity uses the existing i18n error envelope.
- `go test ./modules/opanalytics/...` passes.

## Out of Scope

- `content_type` classification and message type distribution.
- `/channels/{id}/members` detail table.
- co-active, member-pair facts, bitmap work, and exports.

an9xyz:
## Summary
- add message_composition to the manager dashboard overview response with fixed conv_type 1..4 buckets
- add GET /v1/manager/dashboard/trend with day/week buckets and zero-filled gaps
- document the new trend API and response shapes in Swagger

## Issue
Closes #318

## Notes
- uses existing opanalytics dim/fact tables only; no ETL/content_type/pair changes
- keeps private chats global-only: space-filtered overview/trend returns private metrics as 0

## Tests
- go test ./modules/opanalytics/...
- go test ./modules/opanalytics -run ^$ -count=1
- go test ./modules/opanalytics -run 'TestOpanalytics(NoLegacyResponseError|RespondHelpers)$'
- go vet ./modules/opanalytics/...
- make i18n-extract-check
- make i18n-lint
- git diff --check

### DIFF
--- a/modules/opanalytics/api.go
+++ b/modules/opanalytics/api.go
@@ -55,6 +55,7 @@ func (m *Manager) Route(r *wkhttp.WKHttp) {
 	auth := r.Group("/v1/manager/dashboard", m.ctx.AuthMiddleware(r), appwkhttp.SharedUIDRateLimiter(r, m.ctx))
 	{
 		auth.GET("/overview", m.overview)                       // 模块A 概览卡片
+		auth.GET("/trend", m.trend)                             // 模块C 趋势
 		auth.GET("/spaces", m.spaces)                           // 表一 Space 列表
 		auth.GET("/spaces/:space_id/channels", m.spaceChannels) // 表二 群组列表(仅群组)
 		auth.GET("/global/direct-chats", m.globalDirectChats)   // 全局私聊活跃列表
@@ -75,6 +76,30 @@ func (m *Manager) overview(c *wkhttp.Context) {
 	resp, err := m.service.overview(start, end, parseSpaceIDs(c))
 	if err != nil {
 		m.Error("overview query failed", zap.Error(err))
+		respQueryFailed(c)
+		return
+	}
+	c.Response(resp)
+}
+
+// trend 模块C 趋势(day/week，缺桶补 0)。
+func (m *Manager) trend(c *wkhttp.Context) {
+	if !m.requireSuperAdmin(c) {
+		return
+	}
+	start, end, ok := parseDateRange(c)
+	if !ok {
+		respRequestInvalid(c, "date_range")
+		return
+	}
+	granularity, ok := normalizeGranularity(c.Query("granularity"))
+	if !ok {
+		respRequestInvalid(c, "granularity")
+		return
+	}
+	resp, err := m.service.trend(start, end, granularity, parseSpaceIDs(c))
+	if err != nil {
+		m.Error("trend query failed", zap.Error(err))
 		respQueryFailed(c)
 		return
 	}
@@ -257,6 +282,18 @@ func normalizeActiveStatus(v string) string {
 		return v
 	default:
 		return "all"
+	}
+}
+
+// normalizeGranularity 收敛为 day/week，缺省 day。
+func normalizeGranularity(v string) (string, bool) {
+	switch v {
+	case "", "day":
+		return "day", true
+	case "week":
+		return "week", true
+	default:
+		return "", false
 	}
 }
 
--- a/modules/opanalytics/api_i18n.go
+++ b/modules/opanalytics/api_i18n.go
@@ -12,7 +12,7 @@ func respForbidden(c *wkhttp.Context) {
 	httperr.ResponseErrorL(c, errcode.ErrOpanalyticsForbidden, nil, nil)
 }
 
-// respRequestInvalid 请求参数无效(reason ∈ {date_range, space_id, sort})。
+// respRequestInvalid 请求参数无效(reason 如 date_range / granularity / space_id / sort)。
 func respRequestInvalid(c *wkhttp.Context, reason string) {
 	details := i18n.Details{}
 	if reason != "" {
--- a/modules/opanalytics/api_test.go
+++ b/modules/opanalytics/api_test.go
@@ -406,6 +406,30 @@ func countFacts(t *testing.T, ctx *config.Context) (int64, int64) {
 	return r3, r4
 }
 
+func compositionByType(items []*messageCompositionItem) map[uint8]*messageCompositionItem {
+	out := make(map[uint8]*messageCompositionItem, len(items))
+	for _, item := range items {
+		out[item.ConvType] = item
+	}
+	return out
+}
+
+func trendByBucket(items []*trendItem) map[string]*trendItem {
+	out := make(map[string]*trendItem, len(items))
+	for _, item := range items {
+		out[item.Bucket] = item
+	}
+	return out
+}
+
+func trendConvByType(items []*trendConvTypeMsgItem) map[uint8]*trendConvTypeMsgItem {
+	out := make(map[uint8]*trendConvTypeMsgItem, len(items))
+	for _, item := range items {
+		out[item.ConvType] = item
+	}
+	return out
+}
+
 func TestOpanalyticsEndpoints(t *testing.T) {
 	ctx, route, etl := opaSetup(t)
 	seedScenario(t, ctx)
@@ -426,6 +450,21 @@ func TestOpanalyticsEndpoints(t *testing.T) {
 	assert.Equal(t, int64(12), ov.HumanMsgCount) // g1:5 + g2:2 + private:5
 	assert.Equal(t, int64(5), ov.AgentMsgCount)
 	assert.Equal(t, int64(1), ov.PrivateActiveCount)
+	ovComp := compositionByType(ov.MessageComposition)
+	require.Len(t, ovComp, 4, "overview composition must always return conv_type 1..4")
+	assert.Equal(t, int64(2), ovComp[convTypeHHGroup].HumanMsgCount)
+	assert.Equal(t, int64(0), ovComp[convTypeHHGroup].AgentMsgCount)
+	assert.Equal(t, int64(2), ovComp[convTypeHHGroup].TotalMsgCount)
+	assert.Equal(t, int64(1), ovComp[convTypeHHGroup].ActiveChannelCount)
+	assert.Equal(t, int64(5), ovComp[convTypeHAGroup].HumanMsgCount)
+	assert.Equal(t, int64(5), ovComp[convTypeHAGroup].AgentMsgCount)
+	assert.Equal(t, int64(10), ovComp[convTypeHAGroup].TotalMsgCount)
+	assert.Equal(t, int64(1), ovComp[convTypeHAGroup].ActiveChannelCount)
+	assert.Equal(t, int64(5), ovComp[convTypeHHPrivate].HumanMsgCount)
+	assert.Equal(t, int64(0), ovComp[convTypeHHPrivate].AgentMsgCount)
+	assert.Equal(t, int64(5), ovComp[convTypeHHPrivate].TotalMsgCount)
+	assert.Equal(t, int64(1), ovComp[convTypeHHPrivate].ActiveChannelCount)
+	assert.Zero(t, ovComp[convTypeHAPrivate].TotalMsgCount)
 
 	// ---- overview 限定 Space=s1：总数也随筛选收敛(否则活跃比例失真) ----
 	var ovS1 overviewResp
@@ -440,6 +479,23 @@ func TestOpanalyticsEndpoints(t *testing.T) {
 	assert.Equal(t, int64(5), ovS1.HumanMsgCount, "私聊/g2 不计入 s1")
 	assert.Equal(t, int64(5), ovS1.AgentMsgCount)
 	assert.Equal(t, int64(0), ovS1.PrivateActiveCount, "选中 Space 时私聊数置 0(私聊无 space 归属)")
+	ovS1Comp := compositionByType(ovS1.MessageComposition)
+	require.Len(t, ovS1Comp, 4)
+	assert.Equal(t, int64(10), ovS1Comp[convTypeHAGroup].TotalMsgCount)
+	assert.Equal(t, int64(1), ovS1Comp[convTypeHAGroup].ActiveChannelCount)
+	assert.Zero(t, ovS1Comp[convTypeHHGroup].TotalMsgCount)
+	assert.Zero(t, ovS1Comp[convTypeHHPrivate].TotalMsgCount, "选中 Space 时私聊 composition 为 0")
+	assert.Zero(t, ovS1Comp[convTypeHAPrivate].TotalMsgCount, "选中 Space 时私聊 composition 为 0")
+
+	// ---- overview 空范围：composition 仍固定 4 类且全部补 0 ----
+	var empty overviewResp
+	decodeOK(t, opaGet(t, route, "/v1/manager/dashboard/overview?start_date=2026-05-31&end_date=2026-05-31"), &empty)
+	emptyComp := compositionByType(empty.MessageComposition)
+	require.Len(t, emptyComp, 4)
+	for _, item := range emptyComp {
+		assert.Zero(t, item.TotalMsgCount)
+		assert.Zero(t, item.ActiveChannelCount)
+	}
 
 	// ---- spaces (表一) ----
 	var spaces struct {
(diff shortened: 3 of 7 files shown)


Jerry-Xin (approved):
This PR is in scope for `octo-server` and adds relevant opanalytics dashboard functionality with no blocking issues found.

**💬 Non-blocking**
- 🟡 Warning: I could not run the full `go test ./modules/opanalytics/...` suite because this workspace has no MySQL listening on `127.0.0.1:3306`. The narrower source-guard test passed.
- 🔵 Suggestion: Consider adding one test for a partial-week request that starts midweek, since `buildWeekBuckets` intentionally returns the Monday bucket while clipping `start_date`/`end_date` to the request range in [modules/opanalytics/service.go:231](/Users/mlamp/.openclaw/tmp/pr-review-msnr5neh/repo/modules/opanalytics/service.go:231).

**✅ Highlights**
- Route registration keeps the existing authenticated, UID-rate-limited manager dashboard pattern in [modules/opanalytics/api.go:55](/Users/mlamp/.openclaw/tmp/pr-review-msnr5neh/repo/modules/opanalytics/api.go:55).
- The trend endpoint validates date range and granularity before reaching SQL construction in [modules/opanalytics/api.go:86](/Users/mlamp/.openclaw/tmp/pr-review-msnr5neh/repo/modules/opanalytics/api.go:86).
- Space-filtered private metrics naturally zero out through `space_id` filtering, matching the PR description, in [modules/opanalytics/db.go:187](/Users/mlamp/.openclaw/tmp/pr-review-msnr5neh/repo/modules/opanalytics/db.go:187) and [modules/opanalytics/db.go:223](/Users/mlamp/.openclaw/tmp/pr-review-msnr5neh/repo/modules/opanalytics/db.go:223).
- Fixed `conv_type` 1..4 response buckets are normalized in service code, which gives stable client shapes even for empty ranges in [modules/opanalytics/service.go:171](/Users/mlamp/.openclaw/tmp/pr-review-msnr5neh/repo/modules/opanalytics/service.go:171) and [modules/opanalytics/service.go:189](/Users/mlamp/.openclaw/tmp/pr-review-msnr5neh/repo/modules/opanalytics/service.go:189).

Verification performed: `git diff --check` passed, and `go test ./modules/opanalytics -run 'TestOpanalytics(NoLegacyResponseError|RespondHelpers)$' -count=1` passed.

lml2468 (approved):
APPROVED (head `9ce6844e2a1d`).

Real new feature: overview gains module B (`message_composition`), and a new `GET /v1/manager/dashboard/trend` endpoint adds module C (day/week trend with zero-filled gaps). Reuses the existing dim/fact tables — no ETL changes. Traced the SQL, the zero-fill logic, the bucket alignment between Go and MySQL, and the test invariants.

## ✅ Independently verified

**Module B (`message_composition`) — correct fixed-bucket invariant.**
- `queryMessageComposition` aggregates `octo_fact_channel_daily` by `conv_type` with `WHERE conv_type BETWEEN 1 AND 4` — this correctly excludes orphan group rows (`conv_type=0`, written when a group disappears from the `group` table per `etl.go`). Good — matches the existing dashboard's "don't surface unknown conv_type" posture.
- `normalizeMessageComposition` fills the missing buckets server-side so the response **always** has exactly 4 items in `convType ∈ {1=HH群, 2=HA群, 3=HH私聊, 4=HA私聊}` order. Frontend / clients can rely on positional access; the empty-range test (`TestOpanalyticsEndpoints` new section `---- overview 空范围 ----`) pins this.
- `applySpaceFilter` reuses the existing helper, so the established "private chats are global-only" rule (private rows have `space_id=''`, get filtered out under any non-empty space filter) carries forward automatically. `TestOpanalyticsEndpoints` `ovS1Comp` block asserts: when `space_ids=s1`, both private conv_types report 0, exactly matching the documented "选中 Space 时私聊置 0" invariant.
- `TotalMsgCount = HumanMsgCount + AgentMsgCount` consistently computed in Go (not in SQL), so an additional message_type added later only needs the Go side and templates updated, not the SQL.

**Module C (`trend`) — bucket alignment Go ↔ MySQL is correct.**
The week-bucket logic must agree between Go and MySQL or the join in the service layer would lose data. Both sides snap to Monday in the report timezone:
- Go: `mondayOf` uses `(weekday+6) % 7` (Go's `time.Weekday`: Sun=0..Sat=6 → offset Sun=6, Mon=0, Tue=1, ..., Sat=5). `d.AddDate(0,0,-offset)` snaps to the Monday on/before `d`.
- MySQL: `DATE_SUB(stat_date, INTERVAL WEEKDAY(stat_date) DAY)` snaps to the Monday — MySQL's `WEEKDAY()` returns Mon=0..Sun=6, identical semantics.
- Both produce the same `YYYY-MM-DD` Monday string as bucket key; the service-layer map join `channelAgg[b.Bucket]` and `memberAgg[b.Bucket]` aligns by construction. The week test (`TestOpanalyticsTrendEndpoint` `weekTrend`) asserts the bucket key is `2026-06-01` (a Monday) for a `2026-06-01..2026-06-07` range — confirms both halves of the calculation match.

**Trend zero-fill correctness.**
- `buildTrendBuckets` materializes every expected bucket Go-side; `service.trend` then looks up SQL results in maps. Missing days/weeks naturally fall through to zero-valued aggregates because Go map lookup returns the zero value for an unset key. `TestOpanalyticsTrendEndpoint` `may31 := byDay["2026-05-31"]` block pins this: a day with no messages still appears in the response with `TotalMsgCount=0`. Solid pattern.
- `start_date`/`end_date` on a bucket are clipped to the requested range: a week bucket whose Monday is before `start` reports `start` as `start_date`, and a week bucket whose Sunday is after `end` reports `end` as `end_date`. This is the right surface — clients don't need to do their own clipping when displaying partial weeks at the edges of the range.

**Trend active-members correctness (distinct, not sum).**
This was the trickiest invariant. For a week bucket, `active_human` must be `COUNT(DISTINCT sender_uid)` over the whole week — summing daily counts would double-count anyone who messaged multiple days. `queryTrendActiveMembers` does this correctly: `COUNT(DISTINCT CASE WHEN m.member_type=1 THEN f.sender_uid END)` grouped **by bucket** (where bucket is the week's Monday), so distinct is applied per-week. The test pins this explicitly: `assert.Equal(t, int64(3), week.ActiveHumanMembers, "week active members must be bucket-level distinct, not daily sums")`. Same for `active_groups` on the channel side (`COUNT(DISTINCT CASE WHEN channel_type=2 THEN channel_id END)`). Worth the assertion text — this is exactly the kind of metric that quietly inflates without it.

**SQL injection surface is closed.**
- `granularity` is normalized to a literal `"day"` or `"week"` (or rejected with 400) **before** `trendBucketSQL` builds the column expression. The 400 path uses `respRequestInvalid(c, "granularity")` — same pattern as `date_range` — keeping the error code surface consistent and adding `granularity` to the documented `reason` enum in `api_i18n.go`.
- `start`/`end` and `spaceIDs` go through bound parameters (`?`); the only literal SQL concat is the bucket expression with the enum-validated `granularity`. `inClause` is reused from existing code for the `IN (?, ?, ?)` expansion. Good.

**Auth & rate-limit gating consistent with sibling endpoints.**
- `/v1/manager/dashboard/trend` is mounted in the same `auth` group with `AuthMiddleware` + `SharedUIDRateLimiter`, and `requireSuperAdmin(c)` is the first thing the handler does — identical to the 5 existing dashboard routes. Test pins the 403 path (`setPlainUserToken` + GET returns `err.server.opanalytics.forbidden`).

**Test data covers the right edges.**
- Cross-day data (`nextDay := "2026-06-02"`) is inserted after the initial seed, then `etl.RunIncremental()` is called twice — proves the trend reflects the **post-ETL** state, not just the seed.
- Three space-filter scenarios on trend: global vs `space_ids=s1` for both endpoints, and a `granularity=month` 400 path. Composition's 3 scenarios (global, `s1`-filtered, empty-range) cover the fixed-bucket invariant in both directions.
- Conv_type distribution on the week bucket cross-validates against the same SQL aggregation: `2 + 10 + 5 + 0 = 17`, and `week.HumanMsgCount + week.AgentMsgCount = 13 + 6 = 19` — wait, those should match... let me re-check. `human_msg + agent_msg = 19` and conv_type sum = `2+10+5+0 = 17`. There's a 2-message gap, which is exactly the `nextDay` `1+1=2` HA-Group messages from `g1`, bringing conv_type HA-Group from 10 to 12. So `2+12+5+0 = 19`. The test asserts `weekConv[convTypeHAGroup].TotalMsgCount` is 12, and the channel totals are 19. ✓ Numbers are consistent end-to-end.

1. **`buildTrendBuckets` re-parses dates that `parseDateRange` already validated.** `parseDateRange` returns `start`/`end` as strings (`YYYY-MM-DD`), so `buildTrendBuckets` calls `time.ParseInLocation` again to walk the buckets. Tiny redundancy; not worth refactoring `parseDateRange` to return `time.Time` for this PR.
2. **`trendBucketSQL` is a string concat, not a prepared statement template.** This is fine because `granularity` is whitelist-normalized first, but a future contributor copy-pasting this pattern with a non-validated input would have a SQL injection. Consider a `// SECURITY: granularity must already be normalized via normalizeGranularity before this is called.` comment on `trendBucketSQL`. Cosmetic doc gap.
3. **Week bucket key is the calendar Monday even when clipped.** A query for `2026-06-03..2026-06-07` (Wed–Sun) returns one bucket with `Bucket: "2026-06-01"`, `StartDate: "2026-06-03"`, `EndDate: "2026-06-07"`. This is the documented behavior (and tested), but if a client renders the `bucket` field as the headline date they'd surface a Monday outside their requested range. The `start_date`/`end_date` clipped values are the right thing to display; worth a Swagger description nudge ("clients should display `start_date`/`end_date`, not `bucket`, for partial weeks"). Cosmetic.
4. **`message_composition` ordering relies on the Go slice append order matching the `[1,2,3,4]` iteration.** It does — `normalizeMessageComposition` appends in that fixed order — but a unit test asserting `out[0].ConvType==1 && out[1].ConvType==2 && ...` would lock that contract independently of the map iteration which is unordered. The current test only spot-checks individual `convTypeXxx` keys via the helper map; either is fine, but the positional contract isn't currently pinned. Cosmetic.
5. **Two `space_ids` parsing paths.** `trend` uses `parseSpaceIDs(c)` (existing helper), so it inherits the same "multi-value" parsing the other endpoints use. Good consistency.

## CI

No prior reviews; CI/code-review not yet reported at review time. PR description confirms `go test ./modules/opanalytics/...` + `go vet` + `make i18n-extract-check` + `make i18n-lint` + `git diff --check` all pass locally. Ride CI to green before merge.

## Verdict

Clean module-C addition built entirely on the existing dim/fact tables. The trickiest invariants — week-bucket alignment between Go and MySQL, distinct-per-bucket active counts (not daily sums), and the "private chats are global-only" rule under space filters — are all correct AND independently pinned by tests with descriptive assertion messages. Module B's fixed-4-row composition rule is cleanly normalized server-side so clients can rely on positional access. LGTM, merge once CI confirms green.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#317 Review Report — feat(opanalytics): add trend and message composition

**PR**: https://github.com/Mininglamp-OSS/octo-server/pull/317
**Head SHA**: `9ce6844e2a1d895ba379cdc8280125c6e5de9c44`
**Scope**: 7 files, +669/−9 — new `/trend` endpoint, `message_composition` in overview, Swagger docs, tests

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Route registration & auth | ✅ | `api.go:58` — `/trend` under same `auth` group with `AuthMiddleware` + `SharedUIDRateLimiter`; `requireSuperAdmin` check at handler top |
| Input validation (date_range, granularity) | ✅ | `api.go:89-98` — `parseDateRange` (existing, 366-day cap) + `normalizeGranularity` (whitelist: day/week, default day); invalid → `respRequestInvalid` |
| SQL correctness — channel totals | ✅ | `db.go:queryTrendChannelTotals` — `SUM(IFNULL(...))`, `COUNT(DISTINCT CASE...)`, parameterized `BETWEEN ? AND ?`, optional `inClause` for space filter |
| SQL correctness — conv_type breakdown | ✅ | `db.go:queryTrendConvTypeMsg` — `conv_type BETWEEN 1 AND 4` guard, `GROUP BY bucket, conv_type` |
| SQL correctness — active members | ✅ | `db.go:queryTrendActiveMembers` — mirrors `overviewActiveMembers` pattern: `JOIN octo_dim_member`, optional `JOIN space_member`, `m.is_excluded=0`, `COUNT(DISTINCT CASE WHEN member_type=...)` |
| Week bucket alignment (SQL↔Go) | ✅ | SQL `DATE_SUB(col, INTERVAL WEEKDAY(col) DAY)` (MySQL WEEKDAY: Mon=0) ≡ Go `mondayOf` via `(int(Weekday)+6)%7`; both produce Monday in report timezone |
| Zero-fill for missing buckets | ✅ | `service.go:trend` — iterates `buildTrendBuckets` (complete day/week range), map lookup returns zero-value struct for gaps |
| `normalizeTrendConvTypes` nil-safety | ✅ | `service.go:191` — nil map read returns zero-value `trendConvTypeAgg`, always emits 4 conv_types |
| `normalizeMessageComposition` gap-fill | ✅ | `service.go:179` — iterates fixed `[1,2,3,4]`, zero-fills missing types |
| Overview backward compatibility | ✅ | `model.go:overviewResp` — additive field `message_composition`; existing JSON fields unchanged |
| Test coverage | ✅ | `api_test.go` — `TestOpanalyticsEndpoints` (composition: global, space-filtered, empty range) + `TestOpanalyticsTrendEndpoint` (day/week/zero-fill/space-filter/auth/error cases) |
| Swagger documentation | ✅ | `swagger/api.yaml` — `/manager/dashboard/trend` path + `trendResp`/`trendItem`/`trendConvTypeMsgItem`/`messageCompositionItem` definitions match Go structs |

## 2. Findings

### No P0 or P1 issues found.

### P2 — Trend endpoint: private metrics not defensively zeroed when space-filtered

**Diff-scope**: new (this PR introduces the trend endpoint)

**Description**: The overview explicitly guards private data under space filter:
```go
// service.go:49-52 (overview)
var privateActive int64
if len(spaceIDs) == 0 {
    if privateActive, err = s.db.privateActiveCount(start, end); err != nil { ... }
}
```

The trend endpoint does NOT have an equivalent guard. `queryTrendChannelTotals` includes `COUNT(DISTINCT CASE WHEN channel_type=1 THEN channel_id END) AS private_active` inside the space-filtered query, relying on private channels having `space_id=""` in the ETL to be naturally excluded by `space_id IN (...)`.

Similarly, `queryTrendConvTypeMsg` includes conv_type 3/4 (HH-Private, HA-Private) inside the space-filtered query without explicit zeroing.

**Current behavior**: Correct — tests verify `PrivateActiveCount=0` and `convTypeHHPrivate=0` when space-filtered, because ETL assigns `space_id=""` to private channels.

**Risk**: If ETL ever assigns private channels to actual space_ids (e.g., for a feature where private chats belong to a workspace), the trend endpoint would silently leak cross-space private data while the overview would still show 0. The two endpoints would diverge.

**Severity**: P2 (maintainability / defense-in-depth). Not blocking — behavior is correct today and tested.

**Suggestion**: Add an explicit post-query zeroing in `service.trend()` when `len(spaceIDs) > 0`, mirroring the overview pattern:
```go
if len(spaceIDs) > 0 {
    item.PrivateActiveCount = 0
    // zero out conv_type 3,4 in ConvTypeMsgCounts
}
```

### P2 (nit) — `trendBucketSQL` uses string concatenation for column names

**Diff-scope**: new

**Description**: `db.go:trendBucketSQL(col, granularity)` builds SQL via `"DATE_FORMAT(" + col + ...)"`. All 4 call sites use hardcoded values (`"stat_date"`, `"f.stat_date"`), and `granularity` is validated by `normalizeGranularity`. No injection vector exists today.

**Suggestion**: Consider a whitelist map or adding a comment noting that `col` must be a trusted/hardcoded value, to prevent future misuse.

## 3. Data Flow Tracing

|---|---|---|
| `trendItem.HumanMsgCount` / `AgentMsgCount` | `queryTrendChannelTotals` → `SUM(human_msg_count)` / `SUM(agent_msg_count)` from `octo_fact_channel_daily` | ✅ Parameterized BETWEEN, IFNULL guards |
| `trendItem.ActiveHumanMembers` / `ActiveAgentMembers` | `queryTrendActiveMembers` → `COUNT(DISTINCT CASE WHEN m.member_type=N THEN f.sender_uid END)` from `octo_fact_member_channel_daily ⋈ octo_dim_member` | ✅ Mirrors overview pattern; is_excluded=0; optional space_member join |
| `trendItem.ActiveGroups` | `COUNT(DISTINCT CASE WHEN channel_type=2 THEN channel_id END)` | ✅ |
| `trendItem.PrivateActiveCount` | `COUNT(DISTINCT CASE WHEN channel_type=1 THEN channel_id END)` | ✅ (with P2 note above) |
| `trendItem.ConvTypeMsgCounts` | `queryTrendConvTypeMsg` → `GROUP BY bucket, conv_type` with `conv_type BETWEEN 1 AND 4` | ✅ `normalizeTrendConvTypes` always emits 4 types, nil-safe |
| `overviewResp.MessageComposition` | `queryMessageComposition` → `GROUP BY conv_type` with `conv_type BETWEEN 1 AND 4` | ✅ `normalizeMessageComposition` always emits 4 types with zero-fill |
| `trendItem.Bucket` / `StartDate` / `EndDate` | `buildTrendBuckets` (Go) using `reportLocation()` timezone | ✅ Aligned with SQL `trendBucketSQL` week computation |
| Zero-filled gaps | `buildTrendBuckets` generates complete day/week range; map miss → zero-value struct | ✅ |

## 4. Additional Observations

- **Backward compatibility**: `overviewResp` gains `message_composition` (additive). Existing API consumers unaffected.
- **Performance**: Trend queries run 3 SQL queries (channel totals + conv-type + active members) per request. All use `GROUP BY` on indexed `stat_date` with bounded range (≤366 days). Consistent with existing overview pattern.
- **Week bucket clamping**: `buildWeekBuckets` correctly clamps `StartDate`/`EndDate` to user's requested range while SQL `BETWEEN start AND end` handles data-side clamping. No double-counting risk.
- **Error propagation**: All DB errors propagate up to the handler, which logs and returns `respQueryFailed`. Consistent with existing handlers.

## 5. Verdict

No P0/P1 issues. Two P2 notes (defense-in-depth for private data under space filter; string concatenation nit in SQL builder). Code is well-structured, follows existing patterns, has thorough test coverage, and correctly implements the described feature.

[Octo-Q] verdict: **APPROVE** — no blocking issues found. P2 notes are advisory.

yujiawei (approved):
# Code Review — PR #317 (octo-server)

## Verdict: APPROVED

Solid, well-scoped addition. The new `GET /v1/manager/dashboard/trend` endpoint and the `message_composition` block on the overview response are correct, parameterized, super-admin gated, and read-only against the existing pre-aggregated fact tables — no ETL/schema changes. Build and `go vet` pass; the i18n legacy-response guard and respond-helper unit tests pass. No P0/P1 issues found. The findings below are all P2 (maintainability, test coverage, documentation) and do not block merge.

## What was verified ✅

- **Auth & rate limiting**: `trend` calls `requireSuperAdmin` and inherits the `SharedUIDRateLimiter` mounted on the `/v1/manager/dashboard` group — consistent with sibling endpoints.
- **SQL injection**: all new raw SQL in `db.go` (`queryTrendChannelTotals`, `queryTrendConvTypeMsg`, `queryTrendActiveMembers`) uses parameterized placeholders via `inClause`; `granularity` is normalized to `day`/`week` before it reaches `trendBucketSQL`, so the only interpolated token is a fixed internal string. No user value is concatenated unescaped.
- **Week-bucket alignment (day/week)**: the MySQL bucket `DATE_SUB(col, INTERVAL WEEKDAY(col) DAY)` (Monday=0…Sunday=6) and the Go `mondayOf` (`offset=(weekday+6)%7`, Go Sunday=0) both resolve to Monday in the report timezone. Both operate on whole calendar dates, so there is no DST/offset drift between the SQL grouping and the Go zero-fill bucket generation. They agree.
- **Distinct active members per week**: `queryTrendActiveMembers` groups by the week bucket and uses `COUNT(DISTINCT … sender_uid)`, so week-level active counts are bucket-distinct rather than a sum of daily counts — the test at `api_test.go` (`week.ActiveHumanMembers`/`ActiveGroups`) pins this correctly.
- **Selected-space privacy semantics**: private chats (`channel_type=1`, `space_id=""`) correctly return 0 for private metrics when a space filter is applied, matching the overview contract and the swagger note. Covered by the `space_ids=s1` trend assertions.
- **Zero-fill**: missing day/week buckets are emitted with zeroed aggregates (`buildTrendBuckets` drives the bucket list, DB maps only fill present buckets). Covered by the `2026-05-31` empty-day assertion.
- **Swagger ↔ structs ↔ tests** are consistent for the new shapes (`trendResp`, `trendItem`, `trendConvTypeMsgItem`, `messageCompositionItem`). Note: `human_msg_count`/`agent_msg_count`/`private_active_count` on `overviewResp` are **pre-existing**; the diff only re-aligns them via gofmt. The single net-new overview field is `message_composition`.

### 1. Orphan-group (`conv_type=0`) messages make the composition / conv-type breakdown under-sum vs the headline totals
`db.go` `queryMessageComposition` and `queryTrendConvTypeMsg` filter `conv_type BETWEEN 1 AND 4`, but `overviewMsgAndGroups` and `queryTrendChannelTotals` do **not** filter `conv_type`. Per `etl.go` `resolveChannelMeta` (the orphan-group branch returns `convType: 0` for a group hard-deleted from the `group` table while its traffic is retained), such rows land in `octo_fact_channel_daily` as `channel_type=2, conv_type=0` with real message counts. Consequence: when an orphan group has traffic in range, `overview.human_msg_count + agent_msg_count` (and `trendItem.total_msg_count`) are strictly greater than the sum over the four `message_composition` / `conv_type_msg_counts` buckets, and `active_groups` counts the orphan while the breakdown does not — a dashboard rendering "total" next to a composition pie/stacked bar will show numbers that don't reconcile, with no "other/unknown" bucket to absorb the remainder.

This is a defensible design choice (an unknown-type row genuinely can't be placed in a typed bucket, and the ETL deliberately keeps orphan traffic in totals), but it is **undocumented in the read path** and **untested**.

*Suggestion*: pick one reconciliation contract and pin it — either (a) also filter `conv_type BETWEEN 1 AND 4` in the totals queries so headline numbers reconcile with the breakdown, or (b) keep totals inclusive but add a code comment at the two breakdown queries explaining the intentional `conv_type=0` drop, and seed an orphan-group fact row in a test asserting the chosen behavior.

### 2. Test coverage gaps on the new endpoint
- **Partial-week clamping is never exercised.** The only week-granularity test uses `start_date=2026-06-01` (a Monday) → `2026-06-07` (Sunday), so both clamp branches in `buildWeekBuckets` (`if bs.Before(start)` / `if be.After(end)`) are no-ops. A mid-week range, e.g. `start_date=2026-06-03` (Wed) `end_date=2026-06-16` (Tue), would exercise both branches and the multi-week loop (expect first bucket `bucket=2026-06-01, start_date=2026-06-03, end_date=2026-06-07`; last `bucket=2026-06-15, start_date=2026-06-15, end_date=2026-06-16`).
- **Trend `date_range` failure modes are untested.** `parseDateRange` rejects `start>end` and malformed dates, surfaced by `trend` as `request_invalid`, but no test hits these for `/trend` (only `granularity=month` and forbidden are covered). Add two cheap assertions: `start_date>end_date` and a malformed date (e.g. `2026-13-40`), both expecting `err.server.opanalytics.request_invalid`.
- **Populated day buckets don't assert date bounds.** Only the zero-filled `may31` bucket asserts `start_date`/`end_date`; `jun1`/`jun2` assert counts only. Adding `start_date`/`end_date` assertions on a populated row guards `buildTrendBuckets` for day granularity.

### 3. Maintainability nits
- **Duplicated active-member SQL.** `queryTrendActiveMembers` and `overviewActiveMembers` share a byte-for-byte FROM/JOIN/WHERE body (`octo_fact_member_channel_daily f JOIN octo_dim_member m`, optional `space_member status=1` join, `is_excluded=0`, member-type `CASE` split); the trend variant only adds the bucket SELECT + `GROUP BY bucket`. The `is_excluded`/member-type/`status=1` semantics are now defined in two places and must be kept in sync by hand. Consider extracting a shared builder.
- **Repeated raw-SQL space-filter fragment.** The `if len(spaceIDs) > 0 { … inClause(…) ; sql += " AND " + clause ; args = append(…) }` block is copy-pasted across the three new trend builders (each with a different column prefix: `space_id` / `f.space_id`). The dbr-builder paths centralize this via `applySpaceFilter`; a raw-SQL counterpart (`appendSpaceFilter(sql, args, col, spaceIDs)`) would remove the repetition and the risk of a future prefix typo on a joined query.
- **`queryMessageComposition` loads into the response VO** (`messageCompositionItem`, which has only json tags) rather than a `db`-tagged row struct like every sibling query. It works because dbr falls back to camelCase→snake_case field-name mapping that happens to match the SELECT aliases, but it couples the wire shape and the DB-row shape into one struct. Loading into a local `db`-tagged row and mapping in the service layer (as `queryTrendChannelTotals` already does) would be more consistent and robust to a future field rename.

## Summary
Approve. The endpoint is correct, safe, and consistent with the existing dashboard module. The orphan-group reconciliation behavior (finding 1) is the most consequential item and is worth a follow-up to document/test, but it is an edge case (group hard-deleted while traffic retained) that does not produce incorrect per-query results and does not block merge. The remaining items are test-coverage and maintainability polish.

PR #321, opened 2026-06-09
test(opanalytics): harden trend edge cases

an9xyz:
## Summary
- add fixed conv_type order assertions for overview composition and trend conv_type buckets
- add partial-week trend coverage for clipped start_date/end_date while preserving Monday bucket
- defensively strip private conv_type 3/4 metrics from space-filtered overview/trend results in service code
- document trusted SQL fragment inputs and clarify partial-week Swagger semantics

## Tests
- go test ./modules/opanalytics -run TestOpanalyticsTrendEndpoint -count=1
- go test ./modules/opanalytics/...
- go vet ./modules/opanalytics/...
- git diff --check

Closes #319

PR #322, opened 2026-06-09
fix: add stable bot default avatars

an9xyz:
## Summary
- add 13 embedded Bot default avatar assets under the user module
- return a stable Bot default avatar from /v1/users/:uid/avatar when a non-system Bot has not uploaded an avatar
- keep official system Bot avatars and ordinary user avatar fallback behavior unchanged

## Testing
- go test ./modules/user

Refs Mininglamp-OSS/octo-web#219


lml2468 (dismissed):
APPROVED (head `0cb42daa0101`).

Small, focused PR that adds stable per-bot default avatars. Traced the handler precedence chain, the embed mechanism, the determinism guarantee, and the test coverage. No issues.

## ✅ Independently verified

**Handler precedence chain preserves manually-curated avatars.**
`UserAvatar` walks an explicit early-return ladder before reaching the new bot-default branch:
1. `IsVisitor(uid)` → `visitor.png`
2. `uid == Account.SystemUID` → `u_10000.png`
3. `uid == Account.FileHelperUID` → `fileHelper.jpeg`
4. `strings.HasPrefix(uid, webhookUIDPrefix)` → `writeWebhookAvatar` (existing PR #298 path)
5. DB lookup, then `userInfo.IsUploadAvatar == 1` → user-uploaded asset
6. **NEW**: `shouldUseBotDefaultAvatar(uid, userInfo)` → embedded colorful PNG by crc32(uid)%13
7. Existing local `Avatar.Default` / `Avatar.DefaultBaseURL` / `generateDefaultAvatar` chain

This precedence is the right call:
- `u_10000` (the typical `SystemUID`) hits branch 2 and **never** reaches the new bot-default branch, so it keeps its handcrafted `u_10000.png` (not a colored 1-of-13).
- `fileHelper` hits branch 3, keeps `fileHelper.jpeg`.
- `botfather` and `notification` fall through to the new branch (they're in `SystemBots` but NOT in `Account.SystemUID`/`FileHelperUID`), so they get stable colorful defaults — good UX win for the two most-visible bots without a curated PNG.
- Ordinary bot users (`Robot == 1`) get colorful defaults.
- Ordinary users (`Robot == 0`, not a `SystemBot`) get the existing default path unchanged. ✓

**`shouldUseBotDefaultAvatar` is correctly bounded.**
- `userInfo == nil` → false (defensive nil-guard before reading `.Robot`; `TestShouldUseBotDefaultAvatar/nil user` pins this).
- `userInfo.Robot == 1 || spacepkg.IsSystemBot(uid)` → captures both user-created bots and the 4 well-known system bots (`botfather`, `u_10000`, `fileHelper`, `notification`). The OR is right — a system bot's `userInfo.Robot` may be 0 (if it's not registered via the bot pipeline), and the explicit `IsSystemBot` covers that gap.
- Verified `spacepkg.IsSystemBot` is the existing source of truth in `pkg/space/query.go` (same package referenced by `#321`'s `SystemBots` exclusion). Reusing it is the right call — single point of truth.

**Determinism is real and tested.**
`crc32.ChecksumIEEE([]byte(uid)) % uint32(len(botDefaultAvatarFiles))` is byte-exact deterministic. `TestReadBotDefaultAvatarDeterministic` reads the same UID twice and asserts byte-equal output — the right invariant to pin (asserting "index = 7" would be brittle if assets are reordered; asserting "same uid → same bytes" survives any internal reshuffle). The mod-13 distribution will skew slightly for sequential UIDs (typical generation pattern), but for color assignment that's aesthetic, not correctness.

**Embed FS is build-time safe.**
- `//go:embed assets/bot_default_avatar/*.png` embeds at compile, so `botDefaultAvatarFS.ReadFile(path)` is a memory-map lookup at runtime — no IO failure mode in production.
- `TestBotDefaultAvatarAssets` enforces 3 invariants at unit-test time: `len == 13` (path slice matches expectations), no duplicate paths, every file is a decodable PNG at 512×512. So a build that accidentally drops or corrupts an asset fails the test pass, not production.
- If `ReadFile` ever errors (which shouldn't happen for embedded files), the handler logs `读取 Bot 默认头像失败` and **falls through** to the existing local-default chain. Graceful degradation, not user-visible 500.

**Cache-Control choice matches the existing local-default branch.**
- `Cache-Control: public, max-age=86400` (24h) matches the existing `Avatar.Default` branch right below it. Consistent.
- `Content-Disposition: inline; filename=avatar.png` and `Content-Type: image/png` also mirror the existing branch.

1. **Cache invalidation when a bot uploads a real avatar.** A bot that gets a default colorful avatar gets 24h `max-age`. If the bot then uploads its own avatar via `IsUploadAvatar=1`, the URL stays `/v1/users/:uid/avatar` so clients with the previous default cached won't see the new one until expiry. This is the SAME issue as the existing local-default branch (unchanged from before this PR), and the `?v=X` cache-busting pattern at upload-call sites is the right place to fix it — out of scope here. Worth noting for a future follow-up: have `shouldUseBotDefaultAvatar` return an asset version and propagate it into `ETag` so clients revalidate cheaply.

2. **The crc32 mod-13 distribution skew on sequential UIDs.** For monotonic-counter UIDs (e.g., `27baor9nu_bot` style), crc32 has reasonable avalanche but mod-13 may still cluster mildly. Purely aesthetic concern — if you want strict uniform-13 distribution for Aesthetics, use FNV-1a or full hash → modulo. Not worth a change.

3. **`v := c.Query("v")` is unused in the new bot-default branch** (just like the existing local-default branch below). Consistent, but a future cache-busting story would want to plumb it through.

4. **`Content-Disposition: inline; filename=avatar.png`** is the same for every bot. If browsers honor the filename for a "Save image as..." default, all 13 colors save as `avatar.png` (overwriting each other). Trivial UX nit, matches existing branch.

5. **No `webhookUIDPrefix` cross-check needed** — the early-return at step 4 already short-circuits webhook synthetic UIDs (`iwh_*` from #298) before they hit the user DB lookup, so they never reach `shouldUseBotDefaultAvatar`. ✓

## Verdict

Surgical, well-tested, reuses existing patterns (`spacepkg.IsSystemBot`, embed FS, Cache-Control timing). The precedence chain preserves manually-curated avatars for `u_10000`/`fileHelper` while giving `botfather`/`notification`/user-created bots stable colorful defaults. The deterministic test pins the right invariant (same input → same output), not a brittle index-specific output. LGTM, merge once CI confirms green (`check-sprint` aside — known sprint-board check, not a code signal).

yujiawei (dismissed):
# Code Review — PR #322 (octo-server)

## Verdict: APPROVED

This is a focused, well-tested change that adds stable embedded default avatars for bots. The implementation is correct, low-risk, and consistent with the existing avatar-serving code. No blocking issues found.

## What the change does

- Adds 13 embedded 512x512 PNG bot avatars under `modules/user/assets/bot_default_avatar/`, served via `go:embed`.
- In `UserAvatar` (`modules/user/api.go`), when a bot user has not uploaded a custom avatar, returns a deterministic per-bot default avatar.
- Ordinary user avatar fallback behavior is untouched.

## Correctness review

✅ **Branch placement is correct.** The new logic sits inside the `userInfo.IsUploadAvatar != 1` branch (`api.go:519`), so it only triggers when the bot has no custom avatar. Bots that uploaded an avatar still get their uploaded one.

✅ **Deterministic and stable selection.** `botDefaultAvatarIndex` uses `crc32.ChecksumIEEE([]byte(uid)) % len(files)` (`bot_default_avatar.go:36`), so the same bot UID always maps to the same avatar. This mirrors the existing default-avatar selection style already used in the same function (`crc32(uid) % DefaultCount`), so the approach is consistent with the codebase.

✅ **Graceful degradation.** On read error, the code logs (`u.Error("读取 Bot 默认头像失败", ...)`) and falls through to the existing default-avatar logic (config file → DefaultBaseURL → generated avatar) instead of returning an error. Good defensive behavior — a missing/corrupt asset can never break the endpoint.

✅ **Bot detection covers both cases.** `shouldUseBotDefaultAvatar` returns true for `userInfo.Robot == 1` OR `spacepkg.IsSystemBot(uid)` (`bot_default_avatar.go:33`). The system-bot check matters because system bots (`botfather`, `u_10000`, `fileHelper`, `notification`) may carry `Robot == 0`, and the `nil`-guard prevents a panic. The test suite covers all four of these cases.

✅ **Response headers are consistent.** `Content-Type: image/png`, `Content-Disposition: inline`, and `Cache-Control: public, max-age=86400` match the existing config-default and generated-avatar branches exactly.

## Verification performed

- `go build ./modules/user/...` — passes.
- `go test -run 'TestBotDefaultAvatar|TestReadBotDefaultAvatar|TestShouldUseBotDefaultAvatar' ./modules/user/` — passes. Tests verify: 13 unique assets, all decode as 512x512 PNG, deterministic byte-stable resolution, and the bot-detection truth table.
- `go vet` and `gofmt -l` on the new files — clean.

(Note: the full `go test ./modules/user` run fails on unrelated `TestRedisExistingTokenSetter_SetXX_RealRedis` due to a DB migration/Redis infra precondition in this environment — not related to this PR.)

## Non-blocking suggestions (nits)

- **N1 — Cache header keyed only on `max-age`.** Unlike uploaded avatars (which append `?v=<version>`), the bot default avatar response has no version/ETag query param, and clients will cache it for 24h. If the embedded asset set is ever changed, clients won't see the new image until the cache expires. This is acceptable for a default avatar but worth noting; an `ETag`/content hash could make cache invalidation cleaner. Not required for merge.
- **N2 — `botDefaultAvatarFiles` duplicates the embed glob.** The slice manually lists all 13 files that `//go:embed assets/bot_default_avatar/*.png` already captures. A future file added to the directory but forgotten in the slice would be silently unused. The existing `TestBotDefaultAvatarAssets` (len == 13 + per-file read) guards against drift well enough, so this is a minor maintainability note only.

Overall this is a clean, self-contained fix. Approving.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#322 Review Report — `fix: add stable bot default avatars`
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/322
**Head SHA:** `0cb42daa0101d9a596b2193c1556c9fa0aa55800`
**Changed files:** 16 (3 Go source + 13 binary PNG assets)
**Scope:** +150 / -0 lines

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Bot detection logic | ✅ | `shouldUseBotDefaultAvatar` at `bot_default_avatar.go:30-34` — nil-safe, checks `Robot==1 \|\| IsSystemBot(uid)` |
| Deterministic avatar selection | ✅ | `botDefaultAvatarIndex` at `bot_default_avatar.go:36-38` — CRC32 mod 13, pure function |
| embed.FS correctness | ✅ | `//go:embed assets/bot_default_avatar/*.png` matches all 13 files on disk (468K total) |
| Error handling & fallback | ✅ | `api.go:520-529` — on error, logs and falls through to existing default avatar paths |
| Existing user path unchanged | ✅ | New code only executes when `IsUploadAvatar != 1` AND `shouldUseBotDefaultAvatar` returns true |
| System bot early returns | ✅ | `botfather`, `u_10000`, `fileHelper` still served by hardcoded paths at `api.go:463-494`; only `notification` system bot reaches new path |
| Tests pass | ✅ | All 3 test functions (7 sub-tests) pass: `go test ./modules/user/ -run TestBotDefaultAvatar\|TestReadBotDefaultAvatar\|TestShouldUseBotDefaultAvatar` |
| Concurrency safety | ✅ | `embed.FS.ReadFile` is read-only/thread-safe; CRC32 is pure computation |
| Security (path traversal) | ✅ | UID is never used as a file path — only CRC32'd to index into a hardcoded slice |

---

## 2. Findings

### No P0 or P1 issues found.

### P2 — Minor observations (non-blocking)

**P2-1: CRC32 modulo 13 distribution bias**
- `botDefaultAvatarIndex` (`bot_default_avatar.go:37`): `crc32.ChecksumIEEE([]byte(uid)) % 13`
- Since 2^32 is not evenly divisible by 13, there is a negligible bias (~0.000008% overrepresentation for indices 0–8).
- **Impact:** Practically zero. Not blocking.
- **Diff-scope:** New code, but follows the same CRC32-modulo-N pattern already used at `api.go:558` for `DefaultBaseURL` avatars. Pre-existing pattern.

**P2-2: Binary size impact**
- 13 embedded PNGs at 512×512 total 468K. Modest and acceptable for a server binary.
- **Diff-scope:** New. Non-blocking note for awareness.

**P2-3: `notification` is the only system bot reaching the new path**
- `botfather`, `u_10000`, `fileHelper` all have early-return hardcoded avatar paths at `api.go:463-494` and never reach `shouldUseBotDefaultAvatar`.
- `notification` (in `SystemBots` map at `pkg/space/query.go:25`) has no early return, so it IS served by the new bot default avatar — which is an improvement over falling through to the generic colored-circle.
- **Diff-scope:** Pre-existing routing; new code correctly handles this case. Non-blocking note.

---

## 3. Suggestions

- **Optional future enhancement:** Consider adding an `ETag` header (e.g., based on the avatar file's content hash) alongside `Cache-Control`. This would allow conditional requests and more efficient cache invalidation if a bot later uploads a custom avatar. Not blocking — the current `max-age=86400` matches existing patterns.

---

## 4. Extra Findings

None beyond the P2 observations above.

---

## 5. Data-Flow Trace

|---|---|---|
| `userInfo.Robot` | DB `user.robot` column via `u.db.QueryByUID(uid)` → `Model.Robot` (`db.go:511`) | ✅ Yes — queried at `api.go:507`, checked at `bot_default_avatar.go:33` |
| `spacepkg.IsSystemBot(uid)` | `SystemBots` map (`pkg/space/query.go:20-25`) — 4 entries: botfather, u_10000, fileHelper, notification | ✅ Yes — pure map lookup |
| `userInfo.IsUploadAvatar` | DB `user.is_upload_avatar` column via `QueryByUID` → `Model.IsUploadAvatar` (`db.go:506`) | ✅ Yes — gate at `api.go:517` |
| `botDefaultAvatarFS` (embed.FS) | `//go:embed assets/bot_default_avatar/*.png` — 13 files verified on disk | ✅ Yes — all 13 readable, 512×512 PNG, tests confirm |
| `botDefaultAvatarFiles[idx]` | Hardcoded slice of 13 paths (`bot_default_avatar.go:13-27`) | ✅ Yes — CRC32 index always in [0,13) |
| `imageData` from `readBotDefaultAvatar` | `botDefaultAvatarFS.ReadFile(...)` returns `[]byte` | ✅ Yes — served via `c.Data(200, "image/png", imageData)` at `api.go:527` |
| Error fallback | If `readBotDefaultAvatar` returns error, logs and falls through to existing default avatar paths | ✅ Yes — graceful degradation at `api.go:521-522` |

---

## Verdict

**[Octo-Q] verdict: APPROVE**

No P0/P1 issues. The implementation is clean, well-tested, and correctly integrated into the existing avatar serving pipeline. Data flow is sound — all consumed values trace back to valid upstream sources with proper nil/error guards. The three P2 observations are non-blocking informational notes.

lml2468 (dismissed):
APPROVED (head `e74d607d8b37`).

Single-character semantic tightening of the bot-default-avatar predicate. R2 narrows it from "all bots OR all system bots" to **"user-created bots that are NOT system bots"**, which preserves any deployment-curated avatar path for `botfather` / `notification` / future system bots.

## ✅ Independently verified — exact delta

Compared file-by-file vs R1 head `0cb42daa0101`:

| file | change |
|---|---|
| `modules/user/api.go` | byte-identical |
| All 13 `assets/bot_default_avatar/*.png` | byte-identical (blob SHAs match) |
| `modules/user/bot_default_avatar.go` | 1-line predicate change |
| `modules/user/bot_default_avatar_test.go` | 2 test cases updated + 1 new |

The functional delta is one operator:

```go
// R1:
return userInfo.Robot == 1 || spacepkg.IsSystemBot(uid)
// R2:
return userInfo.Robot == 1 && !spacepkg.IsSystemBot(uid)
```

## ✅ Why the new semantic is right (revising my R1 read)

My R1 framed routing `botfather` / `notification` through the new colored defaults as a "good UX win". R2 makes the opposite call, and I think it's actually the more correct one. Trace the precedence chain under R2 semantics:

1. `IsVisitor` → `visitor.png`
2. `Account.SystemUID` (typically `u_10000`) → handcrafted PNG
3. `Account.FileHelperUID` (typically `fileHelper`) → handcrafted JPEG
4. webhook prefix → `writeWebhookAvatar`
5. DB hit + `IsUploadAvatar==1` → user-uploaded asset
6. **NEW (R2)**: `Robot==1 && !IsSystemBot(uid)` → 1-of-13 colored default
7. Existing chain: `Avatar.Default` config → `Avatar.DefaultBaseURL` → `generateDefaultAvatar(uid)`

Under R2:
- **`u_10000` / `fileHelper`** still hit branches 2/3 — keep their handcrafted PNG/JPEG (unchanged in both R1 and R2).
- **`botfather` / `notification`** (system bots NOT in branches 2/3) now skip branch 6 and fall through to branch 7's existing `Avatar.Default` / `generateDefaultAvatar` chain. **If a deployment has uploaded a real avatar for `botfather` via the normal upload path**, branch 5 already hit and they get that. If not, they get the existing generic per-uid colorful circle — same as before this PR.
- **User-created bots (`Robot==1, !IsSystemBot`)** get the new stable 1-of-13 colored asset.
- **Ordinary users** unchanged.

This is the **more defensive** design: system bots are a brand-controlled surface (they're the "infrastructure" bots, users see them as the platform speaking), so leaving them on the existing deployment-tunable path is better than auto-rotating them into one of 13 colors. User-created bots are where the colorful identity helps (lots of them, no central branding, visual disambiguation matters).

My R1 read missed that distinction. R2 author got it right.

## ✅ Test coverage matches the new contract

`TestShouldUseBotDefaultAvatar` updated:
- `system botfather keeps official avatar path` — `Robot:1` (also updated from R1's `Robot:0` to be realistic) — asserts `want:false`. Pins the carve-out.
- `notification keeps official avatar path` — new case — asserts `want:false`. Explicitly covers the other system bot that's not in branches 2/3.
- `bot user` — `Robot:1`, non-system UID — still asserts `want:true`. Pins the positive path.
- `ordinary user` (`Robot:0`) — `want:false`. Pins user exclusion.
- `nil user` guard — preserved.

Each test name now reads as the invariant it pins ("keeps official avatar path"), so future readers see the intent without diving into the helper.

## ✅ All R1 verifications carry forward

- Handler precedence chain unchanged (api.go byte-identical).
- Embed FS + 13 byte-identical assets (`TestBotDefaultAvatarAssets` still pins 512×512 / no-duplicate / decodable).
- `TestReadBotDefaultAvatarDeterministic` still pins "same uid → same bytes".
- Cache-Control `public, max-age=86400` and `image/png` Content-Type unchanged — still matches the existing local-default branch.
- Graceful degradation on `ReadFile` error unchanged.

## 🟡 Non-blocking observations (R1 carry-overs still valid)

1. **24h cache invalidation** — yujiawei + mochashanyao + me all flagged the lack of ETag / version. Still unaddressed; same scope-out as R1 (existing local-default branch has the same issue, fix belongs in a cache-busting story).
2. **`botDefaultAvatarFiles` slice + `//go:embed` glob duplicates the file list** (yujiawei's R1 observation, still valid) — adding a 14th PNG without updating the slice would leave it unreachable. Could compute the slice at init from `fs.ReadDir(botDefaultAvatarFS, "assets/bot_default_avatar")` + sort, removing the dual-source-of-truth. Cosmetic.
3. **~468K binary size** for 13 PNGs (mochashanyao's R1 observation) — still a one-time cost, acceptable for the UX gain.

## Verdict

Clean semantic correction. Surgical 1-character predicate change preserves deployment-curated avatar paths for system bots while giving user-created bots their stable colored identity. Test coverage updated to pin the new contract. No regressions to any other invariant pinned in R1. LGTM, merge once CI confirms green.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#322 Review Report — `fix: add stable bot default avatars`
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/322
**Head SHA:** `e74d607d8b3712dd557c1786917292ac8adc9b3d`
**Files:** 16 changed (+156 / −0) — 2 Go source, 1 test, 13 binary PNG assets

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Bot gating logic (Robot==1 && !system bot) | ✅ | `bot_default_avatar.go:30-34`, `pkg/space/query.go:44-45` |
| Deterministic avatar selection (CRC32 mod 13) | ✅ | `bot_default_avatar.go:36-38` |
| embed.FS paths match file list | ✅ | `//go:embed assets/bot_default_avatar/*.png` covers all 13 listed paths |
| Error fallback to existing avatar chain | ✅ | `api.go:522-524` — on error, logs and falls through |
| System bots excluded (botfather, notification, u_10000, fileHelper) | ✅ | `pkg/space/query.go:20-25`, test cases cover botfather + notification |
| Test coverage | ✅ | 3 tests: asset integrity (512×512 PNG), determinism, shouldUse gating |
| No regression to existing avatar paths | ✅ | New block is additive; only entered when `IsUploadAvatar != 1` and bot qualifies |

## 2. Findings

### No P0 or P1 issues found.

### P2 — Minor observations (non-blocking)

**P2-1: `botDefaultAvatarFiles` is a mutable package-level `var`**
- **Path:** `bot_default_avatar.go:14`
- **Risk:** If any code path accidentally mutates this slice, `readBotDefaultAvatar` could panic on out-of-range index or serve wrong data. Low probability in this codebase (no mutation observed).
- **Suggestion:** Consider making it a `const`-like pattern (e.g., unexported + no setter) or add a `sync.Once` guard. Not blocking.

**P2-2: Binary size increase from 13 embedded 512×512 PNGs**
- **Diff-scope:** new
- **Risk:** Each 512×512 PNG could be 10–100KB. Total ~130KB–1.3MB added to binary. Acceptable for the feature, but worth noting for deployment pipelines with binary size budgets.
- **Suggestion:** Consider noting the approximate size addition in the PR description.

**P2-3: Test determinism check uses a single UID**
- **Path:** `bot_default_avatar_test.go:38` — `const uid = "27ba6or9nu_bot"`
- **Diff-scope:** new
- **Risk:** Only one UID is tested for deterministic behavior. While CRC32 is inherently deterministic, testing multiple UIDs would increase confidence in the index distribution.
- **Suggestion:** Add a loop with several UIDs asserting `idx ∈ [0, 13)`.

## 3. Data Flow Trace

|---|---|---|
| `userInfo.Robot` | `u.db.QueryByUID(uid)` → `Model.Robot` (db.go:509) | ✅ DB field populated from `user` table `robot` column |
| `spacepkg.IsSystemBot(uid)` | `SystemBots` map (query.go:20-25) — hardcoded `{botfather, u_10000, fileHelper, notification}` | ✅ Map lookup, always available |
| `botDefaultAvatarFS` | `//go:embed assets/bot_default_avatar/*.png` — compile-time embed | ✅ All 13 files exist in PR, paths match slice entries |
| `botDefaultAvatarIndex(uid)` | `crc32.ChecksumIEEE([]byte(uid)) % 13` | ✅ Always returns [0,12]; same UID → same index |
| `readBotDefaultAvatar` return | `embed.FS.ReadFile(path)` | ✅ Returns `([]byte, error)`; error path falls through to existing avatar chain |

## 4. Architecture Assessment

The integration point in `api.go:519-530` is well-placed:
- **After** the `IsUploadAvatar == 1` check (so bots with custom avatars are unaffected)
- **Before** the generic default avatar chain (so bots get their distinctive avatar first)
- **Error fallback** correctly falls through to existing logic rather than returning 500

The approach mirrors the existing `DefaultBaseURL` avatar selection pattern (`crc32 % count`), maintaining codebase consistency.

## 5. Additional Observations

- `Cache-Control: public, max-age=86400` is appropriate — the avatar for a given UID is deterministic and won't change unless the bot uploads a custom avatar (which flips `IsUploadAvatar` to 1, bypassing this path entirely).
- The `Content-Disposition: inline` header is correct for avatar rendering in browsers.
- `embed.FS` is concurrency-safe for reads — no mutex needed.

---

**[Octo-Q] verdict: APPROVE**

Clean, well-scoped PR. The bot default avatar feature is correctly gated (Robot==1, non-system-bot), deterministic (CRC32 mod 13), and gracefully falls back on error. Tests cover the key paths. No P0/P1 issues found. Minor P2 observations above are non-blocking suggestions.

Jerry-Xin (dismissed):
The PR is in scope for `octo-server`, and I did not find blocking issues in the stable bot default avatar implementation.

💬 Non-blocking
- 🔵 Suggestion: Add a handler-level regression test for `/v1/users/:uid/avatar` covering a non-system bot with `is_upload_avatar=0`, a system bot, and a normal user. The helper tests in `modules/user/bot_default_avatar_test.go` are good, but they do not directly pin the routing/fallback behavior added at `modules/user/api.go:520`.

✅ Highlights
- The system-bot exclusion uses the shared `pkg/space.IsSystemBot` source of truth at `modules/user/bot_default_avatar.go:33`.
- Uploaded avatars still take precedence because the new branch only runs when `IsUploadAvatar != 1` at `modules/user/api.go:517`.
- Asset tests validate count, uniqueness, embed availability, PNG decoding, and 512x512 dimensions.
- Focused tests passed: `go test ./modules/user -run 'TestBotDefaultAvatarAssets|TestReadBotDefaultAvatarDeterministic|TestShouldUseBotDefaultAvatar'`.
- Full `go test ./modules/user` could not run locally because MySQL was unavailable at `127.0.0.1:3306`.

yujiawei (dismissed):
# Code Review — PR #322 (octo-server)

## Verdict: APPROVED

A small, focused, well-tested change that gives non-system Bots a stable, deterministic default avatar from `GET /v1/users/:uid/avatar`. The logic is correct, the fallback is graceful, and existing behavior for ordinary users and system Bots is preserved. No blocking issues found.

## What the PR does

- Embeds 13 Bot default avatar PNGs under `modules/user/assets/bot_default_avatar/` via `//go:embed`.
- Adds `modules/user/bot_default_avatar.go` with `shouldUseBotDefaultAvatar` / `botDefaultAvatarIndex` / `readBotDefaultAvatar`.
- Inserts a new branch in `UserAvatar` (`modules/user/api.go:520-531`) that serves the stable Bot avatar for non-system Bots that have not uploaded an avatar.

## Verification (✅ confirmed against the code)

- **Gating is correct.** `shouldUseBotDefaultAvatar` returns true only when `userInfo.Robot == 1 && !spacepkg.IsSystemBot(uid)` (`bot_default_avatar.go:29-34`). `SystemBots = {botfather, u_10000, fileHelper, notification}` (`pkg/space/query.go:20-25`), so the four system Bots keep their official avatars. `u_10000` and `fileHelper` are additionally short-circuited earlier in `UserAvatar` (`api.go:474-495`), so there is no overlap or regression.
- **Ordering is safe.** The new branch sits inside the `IsUploadAvatar != 1` else-block (`api.go:519-531`), so a Bot that has uploaded a real avatar still takes the upload path. The `iwh_` incoming-webhook synthetic senders are handled earlier (`api.go:499-502`) and never reach this code.
- **Determinism holds.** `botDefaultAvatarIndex` = `crc32.ChecksumIEEE(uid) % 13` — stable across requests for the same uid; the same hashing approach is already used for `DefaultBaseURL` avatars (`api.go:550`). Covered by `TestReadBotDefaultAvatarDeterministic`.
- **Graceful degradation.** If `readBotDefaultAvatar` errors, the code logs and falls through to the existing default-avatar logic (configured file → `DefaultBaseURL` → `generateDefaultAvatar`) rather than returning a broken response (`api.go:521-531`). Good.
- **Assets are valid.** All 13 PNGs are present (512×512, ~33–37 KB each); `TestBotDefaultAvatarAssets` asserts count, uniqueness, decodability, and 512×512 dimensions.
- **Headers are consistent.** `Content-Type: image/png`, `Content-Disposition: inline`, `Cache-Control: public, max-age=86400` match the sibling default-avatar branches (`api.go:541-544`, `561-564`).
- **Build & tests pass.** `go vet ./modules/user/` clean; `go test ./modules/user/ -run 'BotDefaultAvatar|ReadBotDefaultAvatar|ShouldUseBotDefaultAvatar'` → `ok`. `gofmt` clean.

## Non-blocking suggestions (P2 / nit — not required for merge)

1. **Embed glob vs. explicit slice can silently drift.** `//go:embed assets/bot_default_avatar/*.png` embeds every PNG, but only the 13 paths hard-coded in `botDefaultAvatarFiles` are ever served. Adding a 14th PNG would embed it but leave it unused, and the `len == 13` test would not catch that (it only fails if the slice itself changes). Consider deriving the list from `fs.ReadDir(botDefaultAvatarFS, "assets/bot_default_avatar")` (sorted) so the slice and the asset directory can never disagree — or add a test asserting the embedded file count equals `len(botDefaultAvatarFiles)`.

2. **Avatars are served inline on every cache-miss request (~35 KB from memory) rather than redirected.** This is consistent with the existing `generateDefaultAvatar` path, so it is not a regression; the `max-age=86400` header mitigates load. No ETag/Last-Modified is set, but again this matches the surrounding code. Worth keeping in mind if Bot avatar traffic grows.

3. **Cosmetic:** `botDefaultAvatarIndex` uses `% uint32(len(...))`; a zero-length slice would panic on the modulo. The `len == 13` test makes this impossible in practice, so this is purely defensive — no action needed.

None of the above block merge.

an9xyz:
Addressed the non-blocking review suggestions in the latest push:

- Added handler-level regression coverage for `GET /v1/users/:uid/avatar` covering a non-system Bot, a normal user, and the `notification` system Bot.
- Strengthened asset coverage so the embedded `assets/bot_default_avatar` directory and `botDefaultAvatarFiles` list cannot silently drift.
- Expanded deterministic avatar selection coverage across multiple Bot UIDs.

Testing:
- `go test ./modules/user`

lml2468 (approved):
APPROVED (head `5988d8254ebd`).

Test-only follow-up. Production code (`api.go`, `bot_default_avatar.go`) byte-identical to R2 head `e74d607d8b37`; the +80-line delta is entirely in `bot_default_avatar_test.go`. Directly addresses three R1/R2 reviewer observations.

## ✅ Independently verified — exact delta

| file | change |
|---|---|
| `modules/user/api.go` | byte-identical to R2 |
| `modules/user/bot_default_avatar.go` | byte-identical to R2 |
| All 13 `assets/bot_default_avatar/*.png` | byte-identical (verified in R2) |
| `modules/user/bot_default_avatar_test.go` | +80 lines: 3 test improvements |

## ✅ Three R2-reviewer observations closed

**1. yujiawei's "embed glob + hard-coded slice can drift" (R1/R2) → fixed.**

`TestBotDefaultAvatarAssets` now reads the embedded directory and cross-validates:
```go
entries, err := botDefaultAvatarFS.ReadDir("assets/bot_default_avatar")
// ...for each entry, assert seen[path] (i.e., listed in botDefaultAvatarFiles)
if embeddedCount != len(botDefaultAvatarFiles) {
    t.Fatalf("embedded bot default avatar count = %d, want %d", ...)
}
```

Now if anyone adds a 14th PNG to `assets/bot_default_avatar/` without updating the `botDefaultAvatarFiles` slice (or removes a PNG but leaves the slice entry), this test fails. The two sources of truth are now structurally checked against each other. yujiawei's original suggestion of "compute the slice at init from ReadDir" is one way; this is the equivalent invariant pinned as a test, which keeps the explicit slice for IDE jump-to-definition / reviewability while catching drift. Both shapes work; this one is fine.

**2. mochashanyao's "single UID determinism test" (R2) → fixed.**

`TestReadBotDefaultAvatarDeterministic` parameterized over 4 UIDs (`27ba6or9nu_bot`, `sales_bot_001`, `ops_bot_002`, `support_bot_003`) as subtests via `t.Run(uid, ...)`. Each subtest still pins the "same input → same bytes" invariant + index-in-range. The widened input set makes a future regression that breaks one UID's resolution (e.g., an off-by-one in `botDefaultAvatarIndex`) more likely to surface and easier to debug.

**3. NEW: `TestUserAvatarHandlerBotDefaultRouting` — HTTP-layer integration pinning R2 semantics.**

This is the test I would've asked for in R2 if I'd anticipated it. Seeds 3 user models via `u.db.Insert(m)`:

- `avatar_bot_001` (`Robot:1`, NOT a system bot) → handler returns **exact bytes** of `readBotDefaultAvatar("avatar_bot_001")` + `Cache-Control: public, max-age=86400`. Proves the colored-asset branch fires for the intended population.
- `avatar_user_001` (`Robot:0`) → handler returns a 200×200 generated avatar (the `generateDefaultAvatar(uid)` fallback at branch 7), and explicitly `NotEqual` to `readBotDefaultAvatar("avatar_user_001")`. Pins that ordinary users never get the colored asset.
- `notification` (`Robot:1`, IS a system bot per `spacepkg.IsSystemBot`) → handler returns the same 200×200 generated avatar AND explicitly `NotEqual` to `readBotDefaultAvatar("notification")`. **Pins the R2 semantic carve-out at the HTTP boundary**, not just the predicate level.

The negative assertions (`assert.NotEqual(t, wantBotAvatar, normalResp.Body.Bytes())`) are precisely what's needed — they prove the colored asset is **not** being served, regardless of which fallback branch fires. If a future refactor accidentally widened the bot-default predicate (R1's `||` instead of R2's `&& !`), the `notification` case would fail with `want != got` and you'd see exactly which path leaked.

The test uses `testutil.NewTestServer` + `CleanAllTables(ctx)` + explicit `Avatar.Default = ""` / `Avatar.DefaultBaseURL = ""` to deterministically route to `generateDefaultAvatar` for the non-bot paths. Test setup is clean — no reliance on default config that could drift.

## ✅ All R1/R2 verifications carry forward

- Handler precedence chain unchanged.
- `shouldUseBotDefaultAvatar` predicate unchanged (R2 semantic: `Robot==1 && !IsSystemBot(uid)`).
- 13 PNG assets unchanged (still 512×512, deterministic crc32 mod-13 routing).
- Cache-Control / Content-Type / Content-Disposition headers unchanged.
- Graceful degradation on `ReadFile` error unchanged.

1. **R1/R2 caching observation (ETag / cache-busting on upload)** still unaddressed — same scope-out as before, fix belongs in a cache-busting story.
2. **mochashanyao R2 "binary size" observation** still applies — 13 PNGs ~468K in the binary, acceptable cost.
3. **mochashanyao R2 "mutable package-level slice"** still applies — `botDefaultAvatarFiles` is a `var` not a `const` (Go has no slice consts). Could be wrapped behind a function for true immutability, but as a package-local var with no external mutation point, the practical risk is zero. Cosmetic.
4. **The new HTTP integration test relies on real DB inserts** — `testutil.NewTestServer` + `CleanAllTables` means this test won't run without the MySQL test fixture, same dependency every other module test in this repo has. Worth noting that local-only Go test passes (`go test ./modules/user`) will skip or fail this test if the test infra isn't up — but CI does have the infra, so green-on-CI confirms the contract.

## Verdict

Pure test-tier hardening, zero production drift. Closes 3 distinct R1/R2 reviewer observations (yujiawei's embed-list drift, mochashanyao's single-UID test, and an HTTP-layer integration test that nobody explicitly asked for but is the right shape to lock in the R2 semantic carve-out). The negative assertions on `notification` and `avatar_user_001` are the strongest possible pin of "this colored asset is NOT served here". LGTM, merge once CI confirms green.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#322 Review Report — `fix: add stable bot default avatars`
**Head SHA:** `5988d8254ebdec6d4f782411e6ca344bc9779057`
**Files changed:** 16 (3 code + 13 binary PNG assets)
**Scope:** +236 / −0

---

## 1. 验证结论

| 检查项 | 状态 | 证据 |
|--------|------|------|
| Bot gate 逻辑正确性 | ✅ | `shouldUseBotDefaultAvatar` — `bot_default_avatar.go:30-34`: nil guard → `Robot==1 && !IsSystemBot(uid)` |
| System bot 排除 | ✅ | `pkg/space/query.go:20-25`: SystemBots = {botfather, u_10000, fileHelper, notification} — all excluded from new path |
| 确定性映射 (CRC32 mod 13) | ✅ | `bot_default_avatar.go:36-38`: `crc32.ChecksumIEEE([]byte(uid)) % 13` — always in [0,13), no overflow risk |
| embed.FS 资产完整性 | ✅ | Test `TestBotDefaultAvatarAssets` verifies all 13 files are 512×512 PNG, no orphans in embed dir |
| 错误回退路径 | ✅ | `api.go:521-523`: on `readBotDefaultAvatar` error → logs + falls through to existing default avatar chain |
| 已上传头像不受影响 | ✅ | `api.go:516`: `IsUploadAvatar == 1` branch is before the new code; bots with custom avatars bypass new logic entirely |
| 测试覆盖 | ✅ | 4 test functions: asset integrity, determinism, gate logic, handler integration with real DB |
| Cache-Control 一致性 | ✅ | `public, max-age=86400` — appropriate for deterministic per-UID content |

## 2. 发现的问题

**无 P0/P1 发现。**

### P2 — 无阻塞性问题

无。代码逻辑清晰，边界处理到位。

### Nit / 观察

1. **CRC32 分布均匀性** — `crc32.ChecksumIEEE` 对短字符串（典型 bot UID）的分布在 mod 13 下会有极轻微偏差（2^32 mod 13 ≠ 0），但对 13 个桶来说偏差可忽略（< 0.00001%），不影响视觉效果。非阻塞。

2. **`readBotDefaultAvatar` 失败时的回退链** — 如果 embed 读取失败（理论上不应发生），handler 会 fall through 到 `generateDefaultAvatar`（200×200 彩色圆形），对 bot 用户来说会看到一个不同于预期的头像。但这是 graceful degradation，不是 bug。

## 3. 数据流回溯

| 消费点 | 数据来源 | 回溯路径 | 是否真的流到 |
|--------|----------|----------|-------------|
| `shouldUseBotDefaultAvatar(uid, userInfo)` | `userInfo` from `u.db.QueryByUID(uid)` | `api.go:505` → DB query → nil-checked at L509-513 | ✅ 确认非 nil 时才到达 L519 |
| `userInfo.Robot == 1` | DB `user` table `robot` column | `db.go:511` field definition → populated by `QueryByUID` | ✅ |
| `spacepkg.IsSystemBot(uid)` | `pkg/space/query.go:20-25` static map | Hardcoded 4-entry map, O(1) lookup | ✅ |
| `readBotDefaultAvatar(uid)` → `botDefaultAvatarFS.ReadFile(...)` | `embed.FS` from `//go:embed assets/bot_default_avatar/*.png` | Compile-time embed → runtime `ReadFile` with deterministic index | ✅ |
| `c.Data(200, "image/png", imageData)` | `imageData` from `readBotDefaultAvatar` | Only reached when `avatarErr == nil` (L521 guard) | ✅ |

## 4. Diff-scope 三问

本 PR 所有改动均为 **new**（新增文件 + 新增 handler 分支）。未修改任何既有路径的行为契约：
- 既有 `IsUploadAvatar == 1` 路径：未触碰
- 既有 `Avatar.Default` / `Avatar.DefaultBaseURL` / `generateDefaultAvatar` 回退链：未触碰（仅在前面插入了 bot-specific 短路）
- System bot 头像路径：未触碰（被 `!IsSystemBot(uid)` gate 排除）

## 5. 额外发现

无。PR 范围小且聚焦，没有发现任务未覆盖的风险。

---

**[Octo-Q] verdict: APPROVE** — 无 P0/P1 发现。代码逻辑正确，数据流回溯全部通过，错误处理有 graceful fallback，测试覆盖全面（资产完整性 + 确定性 + gate 逻辑 + handler 集成）。建议直接合入。

yujiawei (approved):
# Code Review — PR #322 (octo-server)

## Verdict: APPROVED

A focused, well-scoped change that gives non-system bots a stable, deterministic default avatar. The logic is correct, the fallback behavior is preserved, and the unit tests cover the meaningful cases. No blocking (P0/P1) issues found. A few non-blocking observations below.

## What the change does

- Embeds 13 PNG bot default avatars under `modules/user/assets/bot_default_avatar/` via `//go:embed`.
- Adds `modules/user/bot_default_avatar.go` with three small helpers:
  - `shouldUseBotDefaultAvatar(uid, userInfo)` → `Robot == 1 && !space.IsSystemBot(uid)`
  - `botDefaultAvatarIndex(uid)` → `crc32.ChecksumIEEE(uid) % 13` (stable per uid)
  - `readBotDefaultAvatar(uid)` → reads the embedded file for that index
- Wires it into `UserAvatar` (`modules/user/api.go:520`) inside the `IsUploadAvatar == 0` branch, before the configured/generated fallbacks.

## Strengths

- **Deterministic & stable**: `crc32 % len` keys the avatar to the uid, so the same bot always resolves to the same image (`TestReadBotDefaultAvatarDeterministic`). Same idiom already used for `Avatar.DefaultCount` in this file, so it's consistent with existing code.
- **System bots untouched**: `IsSystemBot` correctly excludes `botfather / u_10000 / fileHelper / notification` (`pkg/space/query.go:20`), preserving official avatars. Verified against `TestShouldUseBotDefaultAvatar`.
- **Graceful degradation**: if `readBotDefaultAvatar` errors, it logs and falls through to the existing default-avatar path rather than failing the request (`api.go:522-525`). Good defensive behavior.
- **Asset/list invariant guarded**: `TestBotDefaultAvatarAssets` asserts list length (13), no duplicate paths, every file decodes as 512×512, and the embedded directory and the `botDefaultAvatarFiles` slice stay in sync — so adding/removing a PNG without updating the list fails CI. Nice.
- Build, `go vet`, and `gofmt` are clean; the pure-unit tests pass locally.

## Non-blocking observations

1. **Precedence over operator-configured defaults (by design, worth confirming).** The bot branch runs *before* the `Avatar.Default` / `Avatar.DefaultBaseURL` checks. So for a non-system bot with no uploaded avatar, the embedded avatar now wins even when an operator has configured a custom default avatar file/URL. This matches the PR intent (bots should look distinct), but it is a behavior change for deployments that set `Avatar.Default*` and previously saw bots use it. Flagging only so it's a conscious choice, not a regression.

2. **`shouldUseBotDefaultAvatar` takes both `uid` and `userInfo` but `userInfo.UID` is ignored.** At the only call site `uid == userInfo.UID`, so this is fine today. Passing `uid` separately keeps the function testable without a populated `UID`, which the tests rely on — acceptable. Just note the implicit assumption that the two refer to the same user.

3. **Cache-Control without an explicit version.** The response sets `Cache-Control: public, max-age=86400`. Because the avatar is deterministic per uid and clients request `/v1/users/:uid/avatar?v=<version>` (cache keyed by full URL incl. `v`), a later real upload changes `v` and busts the cache, and the upload path returns a redirect instead of this branch. So staleness is not a practical issue — noting for completeness.

## Tests

- `TestBotDefaultAvatarAssets`, `TestReadBotDefaultAvatarDeterministic`, `TestShouldUseBotDefaultAvatar` pass locally.
- `TestUserAvatarHandlerBotDefaultRouting` is DB-backed (`testutil.NewTestServer`) and could not run in this review environment due to a local migration-state mismatch unrelated to this PR; the handler-routing assertions it makes (bot → embedded bytes + `max-age=86400`; ordinary user and system bot → generated default) match the code path and should pass in CI.

## Recommendation

Approve. The non-blocking items above are observations/confirmations, not changes required before merge.

Jerry-Xin (approved):
The PR is in scope for `octo-server` and implements the requested bot avatar fallback without changing uploaded-avatar behavior or known system bot handling.

💬 Non-blocking

🟡 Warning: [modules/user/bot_default_avatar.go](/Users/mlamp/.openclaw/tmp/pr-review-j962pqn2/repo/modules/user/bot_default_avatar.go:36) assumes `botDefaultAvatarFiles` is never empty. The current test pins the count at 13, so this is covered, but a tiny guard would make the helper more robust if the list is ever refactored.

🔵 Suggestion: [modules/user/api.go](/Users/mlamp/.openclaw/tmp/pr-review-j962pqn2/repo/modules/user/api.go:520) reads the embedded PNG on every bot avatar request. The files are small, so this is acceptable, but caching the embedded bytes at init would avoid repeated allocations on hot avatar endpoints.

✅ Highlights

The routing order in [modules/user/api.go](/Users/mlamp/.openclaw/tmp/pr-review-j962pqn2/repo/modules/user/api.go:517) preserves uploaded avatars first, then applies the bot-only fallback, then keeps the existing ordinary-user fallback chain.

The system bot exclusion via `spacepkg.IsSystemBot` in [modules/user/bot_default_avatar.go](/Users/mlamp/.openclaw/tmp/pr-review-j962pqn2/repo/modules/user/bot_default_avatar.go:33) aligns with the repository’s existing single source of truth for system bot UIDs.

The tests cover asset presence, PNG validity, deterministic UID mapping, system-bot exclusion, and handler behavior.

Test note: I attempted `go test ./modules/user`, but the local environment does not have MySQL running: `dial tcp 127.0.0.1:3306: connect: connection refused`.

PR #327, opened 2026-06-09
docs(opanalytics): plan P1B message type classification

### LINKED ISSUE #320: feat(opanalytics): spike message type classification for dashboard P1B
## Background

P1A intentionally used only existing aggregated facts for trend and conversation-type composition. The next dashboard capability needs message type distribution, but the current `content_type` field is effectively reserved and not yet populated with stable semantics.

Before implementing charts, we need a P1B spike to confirm where message type can be derived reliably and what ETL/backfill changes are required.

## Goals

- Identify the authoritative source for message type classification:
  - persisted message payload / signal fields,
  - webhook payloads,
  - WuKongIM message metadata,
  - or another stable server-side source.
- Define production-safe category mapping, for example:
  - text,
  - image,
  - voice,
  - video,
  - file,
  - call,
  - rich/card/custom,
  - system/notification,
  - unknown.
- Decide whether system messages, notification messages, and call events should be excluded or reported as explicit categories.
- Design ETL changes for populating `content_type` or an equivalent dimension in opanalytics facts.
- Define historical backfill strategy, idempotency, rollout, and rollback approach.
- Define API shapes for:
  - message type distribution,
  - conversation type x message type matrix.

## Acceptance Criteria

- A concrete classification table is documented with source fields, examples, and fallback behavior for unknown/custom payloads.
- ETL impact is specified, including which fact table(s) change and how exactly-once / dirty-day recompute behavior is preserved.
- Historical backfill plan is documented with operational guardrails for large datasets.
- Dashboard API contract proposal is documented before implementation.
- Risk points are called out explicitly, especially unstable payload schemas, system message treatment, and backward compatibility.

## Out of Scope For The Spike

- Implementing production ETL changes before the classification source is confirmed.
- Adding co-active/member-pair facts or bitmap aggregation.
- Export endpoints.
- `/channels/{id}/members` detail table.

an9xyz:
Refs #320

## Summary
- document the P1B message type classification plan before changing production ETL
- explain why the existing `content_type` reserve column cannot safely model mixed message types under the current fact-table primary key
- define the no-call initial scope, stable dashboard categories, `unknown` handling, new content fact tables, ETL/backfill shape, runtime separation, and API contract draft

## Scope Notes
- call analytics are explicitly out of scope for this P1B pass per current product direction
- interaction analysis remains separate and should use pair/bitmap-oriented facts later
- this PR is a spike/plan PR only; it does not change runtime behavior, migrations, ETL, or API responses
- this PR should not close #320 by itself; #320 remains open until the P0/P1 implementation gap is handled

## Verification
- `git diff --cached --check` before commit
- Go tests not run: documentation-only change

## CI Notes
- docs-only path detection skipped Build/Test/Vet/Lint/i18n jobs as expected
- `check-sprint` passed after linking #320 to Sprint W24
- `code-review` approved
- current red `dependency-review` is a repository configuration failure: dependency graph is not enabled
- current red `secret-scan` reports pre-existing test fixture values in unrelated files, not content added by this PR

### DIFF
--- a/docs/opanalytics-message-type-classification.md
+++ b/docs/opanalytics-message-type-classification.md
@@ -1,0 +1,291 @@
+# Opanalytics P1B Message Type Classification Plan
+
+Status: proposal for #320.
+
+## Goal
+
+P1B should add a production-safe message type dimension for the operation
+analytics dashboard without changing the existing P0/P1A message totals.
+
+The first implementation scope excludes call analytics. Voice/video calls should
+not be surfaced as a dashboard category in P1B. If call-like payloads are seen
+while classifying messages, they should be treated only as non-chat/system-like
+traffic for governance decisions, or as `unknown` until a separate call analytics
+scope is defined.
+
+## Current State
+
+The message shard tables persist the fields needed for an offline classifier:
+`id`, `signal`, `from_uid`, `channel_id`, `channel_type`, `timestamp`,
+`payload`, and deletion metadata.
+
+`octo_fact_member_channel_daily.content_type` exists, but it is a reserved
+column and the ETL currently writes `0` for every row. More importantly, the
+current primary key is `(channel_id, stat_date, sender_uid)`, so the table
+cannot represent one sender producing multiple content types in the same
+channel on the same day. Reusing this column directly would collapse mixed
+message types into one row and lose information.
+
+Therefore P1B should add dedicated content facts rather than mutating the
+existing member-channel daily fact shape.
+
+## Product Scope
+
+In scope:
+
+- Message type distribution.
+- Conversation type x message type matrix.
+- Human/agent x message type counts.
+- System/notification classification for future "effective chat message" count
+  governance.
+- Unknown coverage reporting for encrypted, malformed, and unsupported payloads.
+
+Out of scope:
+
+- Call analytics, call duration, call success rate, and call trend charts.
+- Co-active member interaction analysis.
+- Pair/bitmap facts.
+- Export endpoints.
+- Online queries against message shard tables.
+
+## Classification Categories
+
+Use stable dashboard categories instead of exposing raw `common.ContentType`
+values directly.
+
+| Category | Raw content types | Notes |
+| --- | --- | --- |
+| `text` | `Text=1` | Plain text messages. |
+| `image` | `Image=2`, `GIF=3`, `VectorSticker=12`, `EmojiSticker=13` | Sticker/GIF are image-like for dashboard use. |
+| `voice` | `Voice=4` | Voice messages, not voice call analytics. |
+| `video` | `Video=5` | Video file/message, not video call analytics. |
+| `file` | `File=8` | File messages. |
+| `rich_custom` | `Location=6`, `Card=7`, `MultipleForward=11`, `RichText=14`, `InviteJoinOrganization=16`, unknown app custom types that still parse cleanly | Product-facing "rich/custom" bucket. |
+| `system_notification` | `CMD=99`, `FriendApply=1000`, `GroupCreate=1001`, `GroupMemberAdd=1002`, `GroupMemberRemove=1003`, `FriendSure=1004`, `GroupUpdate=1005`, `RevokeMessage=1006`, `GroupMemberScanJoin=1007`, `GroupTransferGrouper=1008`, `GroupMemberInvite=1009`, `GroupMemberBeRemove=1020`, `GroupMemberQuit=1021`, `GroupUpgrade=1022`, `HotlineAssignTo=1200`, `HotlineSolved=1201`, `HotlineReopen=1202`, `Tip=2000` | Countable as system/notification, and optionally excluded from effective chat totals. |
+| `unknown` | `signal=1`, missing `payload.type`, malformed JSON, non-numeric `type`, `ContentError=97`, `SignalError=98`, unsupported raw types | Never guess. Track coverage so operators can evaluate dashboard completeness. |
+
+`VideoCallResult=9989` and payload commands such as `room.invoke` /
+`rtc.p2p.invoke` are intentionally not productized in P1B. The classifier can
+record them as `system_notification` or `unknown` internally, but P1B APIs must
+not expose a dedicated `call` bucket.
+
+## Classifier Rules
+
+Input:
+
+- `signal`
+- `payload`
+- message metadata used by the existing opanalytics ETL
+
+Rules:
+
+1. If `signal=1`, return `unknown` with reason `signal_encrypted`.
+2. Decode `payload` as JSON with `UseNumber`.
+3. If `payload.type` is absent, non-numeric, or outside the supported mapping,
+   return `unknown` with a reason such as `type_missing`, `type_invalid`, or
+   `type_unsupported`.
+4. Map known raw content types to the stable categories above.
+5. Do not log raw payload content. Logs may include message shard, message id,
+   raw type, category, and reason.
+
+The classifier should be a pure Go helper with table-driven tests. It should not
+depend on HTTP/webhook request context.
+
+## Storage Design
+
+Add new facts instead of changing the existing P0/P1A fact tables.
+
+### `octo_fact_member_channel_content_daily`
+
+One row per day, channel, sender, and dashboard category.
+
+```sql
+CREATE TABLE `octo_fact_member_channel_content_daily` (
+  `stat_date`    DATE         NOT NULL COMMENT '统计日(报告时区自然日)',
+  `channel_id`   VARCHAR(100) NOT NULL COMMENT '会话ID',
+  `channel_type` TINYINT      NOT NULL COMMENT '1=私聊 2=群',
+  `space_id`     VARCHAR(40)  NOT NULL DEFAULT '' COMMENT '群=group.space_id; 私聊=""',
+  `conv_type`    TINYINT      NOT NULL DEFAULT 0 COMMENT '1.HH群 2.HA群 3.HH私聊 4.HA私聊',
+  `msg_category` TINYINT      NOT NULL DEFAULT 0 COMMENT '0=unknown 1=text 2=image 3=voice 4=video 5=file 6=rich_custom 7=system_notification',
+  `sender_uid`   VARCHAR(40)  NOT NULL COMMENT '发送者uid',
+  `sender_type`  TINYINT      NOT NULL DEFAULT 1 COMMENT '1=human 2=agent',
+  `msg_count`    INT          NOT NULL DEFAULT 0 COMMENT '当日该成员该类型消息数',
+  `last_msg_at`  BIGINT       NOT NULL DEFAULT 0 COMMENT '当日该类型最后一条消息时间戳',
+  `created_at`   TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY (`channel_id`,`stat_date`,`sender_uid`,`msg_category`),
+  KEY `idx_date_category` (`stat_date`,`msg_category`,`sender_type`),
+  KEY `idx_space_date_category` (`space_id`,`stat_date`,`msg_category`,`sender_type`),
+  KEY `idx_sender_date_category` (`sender_uid`,`stat_date`,`msg_category`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci
+  COMMENT='看板事实表-成员×会话×日×消息类型';
+```
+
+### `octo_fact_channel_content_daily`
+
+One row per day, channel, and dashboard category. It is recomputed from the
+member-channel content fact for dirty days.
+
+```sql
+CREATE TABLE `octo_fact_channel_content_daily` (
+  `stat_date`       DATE         NOT NULL COMMENT '统计日(报告时区自然日)',
+  `channel_id`      VARCHAR(100) NOT NULL COMMENT '会话ID',
+  `channel_type`    TINYINT      NOT NULL COMMENT '1=私聊 2=群',
+  `space_id`        VARCHAR(40)  NOT NULL DEFAULT '',
+  `conv_type`       TINYINT      NOT NULL DEFAULT 0,
+  `msg_category`    TINYINT      NOT NULL DEFAULT 0,
+  `human_msg_count` INT          NOT NULL DEFAULT 0,
+  `agent_msg_count` INT          NOT NULL DEFAULT 0,
+  `last_msg_at`     BIGINT       NOT NULL DEFAULT 0,
+  `created_at`      TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY (`space_id`,`stat_date`,`channel_id`,`msg_category`),
+  KEY `idx_date_category` (`stat_date`,`msg_category`),
+  KEY `idx_conv_category_date` (`conv_type`,`msg_category`,`stat_date`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci
+  COMMENT='看板事实表-会话×日×消息类型';
+```
+
+## ETL And Backfill
+
+Incremental ETL:
+
+- Keep the current total-message facts unchanged.
+- In the same chunk pass, classify each accepted source message and accumulate
+  content facts by `(day, channel, sender, category)`.
+- Use the existing channel metadata resolution so `space_id`, `conv_type`, and
+  sender type are consistent with current opanalytics totals.
+- Mark touched days dirty for both channel total recompute and content-channel
+  recompute.
+
+Backfill:
+
+- Add a separate content backfill cursor, not the production incremental cursor.
+- Scan each message shard by primary-key keyset pagination.
+- Run from a read replica when available; do not backfill from the online primary
+  for large deployments.
+- Support pause/resume and rate limits.
+- Report progress by shard, last id, rows scanned, rows classified, unknown
+  count, and error count.
+- API rollout must either wait for backfill completion or return a clear
+  `data_complete=false` signal.
+
+Rollback:
+
+- API rollback is safe because existing P0/P1A fields are unchanged.
+- ETL rollback can stop writing content facts while leaving old content rows
+  unused.
+- A down migration may drop the new content fact tables only if operators accept
+  losing backfilled analytics data.
+
+## Runtime Shape
+
+This work is analytics-heavy and should not keep growing inside the online IM
+request path.
+
+Short term:
+
+- Keep the code in this repo for shared models, migrations, auth, and release
+  simplicity.
+- Run ETL/backfill as a separate worker process or deployment.
+- The HTTP API reads only analytics tables.
+- The worker reads message shards from a replica and writes analytics facts.
+
+Medium term:
+
+- Split opanalytics worker/API into an independent service if query volume,
+  backfill cost, or pair/bitmap interaction facts become operationally heavy.
+- Reuse octo-server auth or an internal auth boundary for superAdmin-only
+  dashboard access.
+
+## API Contract Draft
+
+All endpoints stay under `/v1/manager/dashboard`, use superAdmin checks,
+`SharedUIDRateLimiter`, and i18n error envelopes.
+
+### `GET /v1/manager/dashboard/message-types`
+
+Query parameters:
+
+- `start_date`
+- `end_date`
+- `space_ids`
+
+Response:
+
+```json
+{
+  "data_complete": true,
+  "unknown_msg_count": 12,
+  "unknown_rate": 0.0034,
+  "items": [
+    {
+      "msg_category": 1,
+      "category_key": "text",
+      "human_msg_count": 100,
+      "agent_msg_count": 20,
+      "total_msg_count": 120
+    }
+  ]
+}
+```
+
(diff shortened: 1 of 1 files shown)


lml2468 (approved):
APPROVED (head `2991b73f3cde`).

Documentation-only spike/plan PR. Single 291-line markdown file at `docs/opanalytics-message-type-classification.md`; no code, no migrations, no API changes. Author explicitly flags this as "spike/plan PR only; it does not change runtime behavior, migrations, ETL, or API responses".

## ✅ What's right about the plan

**Correct root-cause analysis of why `content_type` can't be reused.**
The doc opens with the strongest single point: `octo_fact_member_channel_daily.content_type` is a reserved column the ETL writes `0` to, BUT the PK is `(channel_id, stat_date, sender_uid)` — so the same sender producing multiple content types in one channel on one day cannot be represented without losing rows. Therefore P1B needs **new** fact tables with `msg_category` in the PK, not a mutation of the existing shape. This is the right call and avoids the most likely footgun ("just start populating the reserved column"). 

**Stable dashboard categories vs raw `common.ContentType`.**
The 8-category mapping (`text` / `image` / `voice` / `video` / `file` / `rich_custom` / `system_notification` / `unknown`) abstracts the dashboard surface from the underlying `common.ContentType` enum. Adding a new raw type later (e.g., `Sticker3D=15`) only needs the classifier table updated, not the API contract. Right separation of concerns.

**Call analytics explicitly out of scope.**
First section flags it three times: "first implementation scope excludes call analytics", "Voice/video calls should not be surfaced as a dashboard category in P1B", "`VideoCallResult=9989` and payload commands such as `room.invoke` / `rtc.p2p.invoke` are intentionally not productized in P1B". This is the right discipline — calls are a different analytical surface (pair-oriented, duration-weighted, success-rate metrics), and conflating them with message-type would lock in a bad model.

**`unknown` as a first-class category with coverage tracking.**
Rather than silently bucketing malformed/encrypted/unsupported into the closest match, the plan keeps `unknown` as its own category and surfaces `unknown_msg_count` + `unknown_rate` on the API. Operators see dashboard completeness; engineering can use the rate as a signal that the classifier is drifting from production traffic. Same defensive shape as `data_complete=false` during backfill rollout.

**Classifier rules order is right.**
1. `signal=1` → `unknown` with `signal_encrypted` reason (never try to read encrypted payload)
2. JSON decode with `UseNumber` (preserves int precision — necessary for type values approaching int32 limits)
3. Missing/non-numeric/out-of-mapping type → `unknown` with explicit reason taxonomy (`type_missing` / `type_invalid` / `type_unsupported`)
4. Map known types to categories
5. **Never log raw payload content** — only shard/id/raw-type/category/reason

The "never log raw payload" rule is important — encrypted/private chat content should never end up in stdout/log aggregators. Explicitly documented is better than implicit.

**Backfill separation from incremental ETL.**
- Separate cursor (not the production incremental cursor — backfill must not move the live water-mark)
- Read replica preferred for large deployments
- Pause/resume + rate limits
- Progress reported per-shard with last-id, scanned, classified, unknown, error counts

This separates the "catch up history" load from the "keep up with live writes" load — the right operational shape, matches the dual-cursor pattern other repos in the org use for analytics backfill.

**Rollout discipline.**
Step 5 explicitly orders: (a) deploy schema + inert code, (b) run backfill, (c) enable APIs after data completeness acceptable. The data_complete signal is the gate. This avoids "API shipped before data exists" embarrassment.

**API contract zero-fill matches existing P0/P1A pattern.**
The response always includes all fixed categories (zero-filled if no data), and `space_ids` filter zeros private conv_types — matching the exact pattern PR #317/#321 established for the existing dashboard. Consistency means clients use the same positional/defensive patterns across all dashboard endpoints.

## 🟡 Open decisions correctly flagged (not gaps in the plan)

The four "Open Decisions" are real product/policy calls that belong to product/娃爹, not to the engineer drafting the plan:

1. **`system_notification` in headline totals or only as a category?** — both choices are defensible; depends on whether "messages" in the dashboard means user-perceptible messages or wire-protocol messages. Worth asking product before P1B-step-4 (read APIs).
2. **`unknown_rate` thresholds → dashboard warnings?** — operational call; could default to "no warning, just expose the rate" and add thresholds later if the field shows persistent skew.
3. **`rich_custom` one bucket vs split?** — start unified, split when product has a visualization need. Right default.
4. **Call-like payloads → silent `system_notification` vs `unknown`?** — I'd prefer `unknown` with a specific reason (`call_like_pending_scope`) so operators see the volume and can prioritize the separate call analytics scope. `system_notification` would silently inflate that bucket.

These are flagged as Open Decisions, not committed designs — exactly right for a plan doc.

## 🟡 Non-blocking observations on the doc itself

1. **No "test plan" section for the runtime separation discussion.** The doc says "Run ETL/backfill as a separate worker process or deployment" but doesn't sketch how the worker boundary will be tested in CI (e.g., process isolation, separate config, separate health endpoints). Worth a future addition. Cosmetic.

2. **The `content_type` reserved column on the existing fact table will keep being written as `0`.** The doc doesn't explicitly state whether to leave it alone, deprecate it via a column comment update, or drop it via a future migration. Worth a one-line "reserved column stays at 0; new content facts supersede" sentence to prevent a future contributor from "fixing" it.

3. **`msg_category` enum mapping is hard-coded in the table column comment** (`0=unknown 1=text 2=image 3=voice 4=video 5=file 6=rich_custom 7=system_notification`). If a future P1C adds a 9th category, the column comment drifts. Consider keeping the source-of-truth in a Go constant + a lint that asserts the column comment matches it. Future hardening; not for the plan PR.

4. **Schema includes both `KEY idx_date_category` and `KEY idx_space_date_category`** on the channel-content fact. The first is a strict prefix of the second's column set (`stat_date, msg_category` vs `space_id, stat_date, msg_category, sender_type`); depending on how queries shape, the first might be redundant. Worth EXPLAIN-ing the planned read queries before adding both. Future tuning; not for the plan PR.

5. **Markdown style minor**: code block at line 102 uses 3 backticks with `sql` highlight; consistent throughout. README-style. ✓

## CI

Single doc file, no code paths exercised, no migrations applied. `code-review` PENDING; the rest of CI should be no-op for this file set. PR description confirms `git diff --cached --check` (whitespace) passes; no Go tests run because there's no Go.

## Verdict

Solid plan doc. Identifies the right root cause (PK shape rejects mixed content types), proposes the right structural fix (new content facts, not column reuse), keeps `unknown` as first-class, separates backfill from incremental ETL, gates rollout on data completeness, flags real product decisions as Open Decisions instead of guessing. The "what's intentionally out of scope" is as valuable as the "what's in scope" — call analytics being explicitly carved out (three times!) means the next contributor won't quietly conflate them. LGTM for merging as a planning artifact; the actual implementation will land as separate PRs in the order specified in §"Implementation Split".

Jerry-Xin (approved):
One-line summary: This documentation-only PR is relevant to octo-server’s existing opanalytics dashboard and is safe to merge, with a couple of non-blocking clarity issues to tighten before implementation starts.

💬 Non-blocking
- 🟡 Warning: [docs/opanalytics-message-type-classification.md](/Users/mlamp/.openclaw/tmp/pr-review-v30xxfpk/repo/docs/opanalytics-message-type-classification.md:63) conflicts with the classifier rules at [line 84](/Users/mlamp/.openclaw/tmp/pr-review-v30xxfpk/repo/docs/opanalytics-message-type-classification.md:84). The category table says cleanly parsed unknown app custom types go to `rich_custom`, while the rules say any type outside the supported mapping returns `unknown`. Pick one rule now, otherwise the classifier tests and dashboard totals can drift.
- 🔵 Suggestion: [docs/opanalytics-message-type-classification.md](/Users/mlamp/.openclaw/tmp/pr-review-v30xxfpk/repo/docs/opanalytics-message-type-classification.md:102) uses non-English SQL column comments inside an otherwise English OSS design document. Existing migrations do use similar comments, so this is not blocking, but this new public-facing plan would be easier to review and implement if the embedded DDL comments were English.
- 🔵 Suggestion: The plan asks for unknown coverage reporting, but the proposed facts only persist `msg_category`, not an `unknown_reason` dimension or separate reason aggregate. If operators need to distinguish encrypted, malformed, missing type, and unsupported payload coverage, add that to the storage/API plan before coding.

✅ Highlights
- The relevance gate passes: this documents a planned extension to `modules/opanalytics`, its existing fact tables, ETL, and `/v1/manager/dashboard` API surface.
- The plan correctly avoids reusing `octo_fact_member_channel_daily.content_type` under the current `(channel_id, stat_date, sender_uid)` primary key.
- The proposed runtime separation, no online shard queries, superAdmin access, UID rate limiting, and i18n envelope requirements align with current repository patterns.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#327 Review Report
**PR:** [docs(opanalytics): plan P1B message type classification](https://github.com/Mininglamp-OSS/octo-server/pull/327)
**Head SHA:** `2991b73f3cdeb68f1a988968d852165d2c8dd7ef`
**Scope:** Documentation-only — single file `docs/opanalytics-message-type-classification.md` (+291/−0)

---

## 1. Verification Summary

| Item | Status | Evidence |
|---|---|---|
| `octo_fact_member_channel_daily` PK = `(channel_id, stat_date, sender_uid)` | ✅ | `sql/20260608000001_opanalytics_init.sql:60` |
| `content_type` reserved column, ETL writes `0` | ✅ | `sql/...init.sql:53`, `etl.go:263` (`ContentType: 0`) |
| All 33 ContentType constant values in classification table | ✅ | Verified 1:1 against `octo-lib/common/msg.go` lines 17–88 |
| Message shard fields (`id`, `signal`, `from_uid`, `channel_id`, `channel_type`, `timestamp`, `payload`) | ✅ | `webhook/sql/20210226000001_webhook_legacy01.sql:4–22` |
| Existing dashboard API pattern (superAdmin, SharedUIDRateLimiter) | ✅ | `opanalytics/api.go:55–63` |
| New table PK design consistent with existing `octo_fact_channel_daily` PK `(space_id, stat_date, channel_id)` | ✅ | New `octo_fact_channel_content_daily` PK = `(space_id, stat_date, channel_id, msg_category)` — additive |

**All factual claims in the plan verified against the codebase. No discrepancies found.**

## 2. Findings

### P2-1: ETL `srcMessageRow` does not read `signal` or `payload` — implementation dependency not explicitly called out

**Diff-scope:** new (this plan introduces a classifier that needs these fields)

The current `srcMessageRow` struct (`model.go:67–74`) reads only `ID, FromUID, ChannelID, ChannelType, Timestamp, CreatedUnix`. The classifier rules in §"Classifier Rules" require `signal` and `payload` as inputs. The plan implies this will be handled but does not explicitly call out that:
- The ETL SELECT query in `etl_db.go` `runChunk()` must be extended to include `signal` and `payload`.
- The `srcMessageRow` struct needs two new fields.

**Recommendation:** Add an explicit note in §"Implementation Split" step 2 that the ETL SELECT and `srcMessageRow` struct must be extended.

### P2-2: `payload` is `mediumblob` — ETL throughput impact unspecified

**Diff-scope:** new (classifier introduces payload reading into the ETL hot path)

The message shard `payload` column is `mediumblob` (up to 16 MB per row). The plan proposes classifying every message during the incremental ETL chunk pass. Reading `payload` for every message could significantly increase I/O and memory pressure per chunk, especially for messages with large attachments or rich content.

**Recommendation:** The plan should address:
- Whether `payload` should be read lazily (only for messages that pass initial signal/type filters).
- Chunk size tuning if payload reading is added.
- Whether the classifier should run in a separate pass from the existing count aggregation to isolate throughput impact.

### P2-3: `RevokeMessage=1006` classified as `system_notification` — potential count inflation

**Diff-scope:** new (classification decision in this plan)

`RevokeMessage` represents a message recall/unsend event, not an actual message. Counting revocations alongside regular system notifications could inflate dashboard message totals and mislead operators about actual message volume.

**Recommendation:** Consider either:
- Moving `RevokeMessage` to a separate `revocation` sub-category.
- Explicitly excluding it from `msg_count` aggregation while still tracking it for audit.
- Calling this out in §"Open Decisions" as a specific item.

### P2-4: Backfill cursor table DDL not specified

**Diff-scope:** new (plan proposes a separate cursor but provides no schema)

The plan states "Add a separate content backfill cursor, not the production incremental cursor" but does not provide the CREATE TABLE DDL. The existing `octo_etl_message_cursor` schema (`sql/...init.sql:89–96`) is a simple `(shard_table, last_id)` pair. The backfill cursor likely needs additional fields for pause/resume state, progress tracking, and rate limiting.

**Recommendation:** Add a DDL sketch for the backfill cursor table, even if provisional.

### P2-5: `signal` values beyond `0` and `1` not addressed

**Diff-scope:** new (classifier rules only handle `signal=1`)

The classifier rules state "If `signal=1`, return `unknown` with reason `signal_encrypted`." The shard schema defines `signal` as `smallint` (range −32768 to 32767). The plan does not specify behavior for other non-zero signal values that may exist or be introduced.

**Recommendation:** Add a rule for `signal != 0 && signal != 1` → `unknown` with reason `signal_unrecognized`, or explicitly state that only `signal ∈ {0, 1}` is expected and other values should be treated as `unknown` with reason `signal_invalid`.

## 3. Suggestions

1. **Add `updated_at` to proposed fact tables.** The existing tables have only `created_at`. Since the recompute pattern is DELETE+INSERT, `created_at` gets reset on every recompute. An `updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP` would help operators distinguish fresh rows from recomputed ones.

2. **Consider index-only query paths for the API.** The proposed `idx_date_category (stat_date, msg_category, sender_type)` on `octo_fact_member_channel_content_daily` is good for the message-types endpoint. For the matrix endpoint, consider whether `idx_conv_category_date` on the channel-level table covers the `conv_type × msg_category` cross-tab without a full scan.

3. **Zero-fill strategy.** The API contract says "always include all fixed categories with zero-filled counts." Consider specifying whether this is done in SQL (LEFT JOIN against a category enum table) or in Go (post-query fill). A Go-side fill is simpler and avoids an extra table.

## 4. Additional Observations

- **No code changes, no risk.** This is a documentation-only PR. All findings are P2 design-review items for the implementation phase, not blockers for landing the plan document.
- **Strong plan quality.** The classification table is a faithful 1:1 mapping of `octo-lib/common/msg.go` constants. The PK limitation analysis of the existing fact table is correct. The rollback strategy is sound. The scope boundaries (excluding call analytics, interaction analysis) are well-justified.
- **Open Decisions section is well-scoped.** The four open questions are concrete and actionable, not vague.

## 5. Data-Flow Trace

Since this is a documentation-only PR with no code changes, data-flow tracing applies to the *proposed* data flows described in the plan:

| Consumer | Upstream Source | Verified? |
|---|---|---|
| Classifier input: `signal` | Message shard `signal` column | ✅ exists in shard DDL; ⚠️ not yet in ETL SELECT (P2-1) |
| Classifier input: `payload` | Message shard `payload` column (mediumblob) | ✅ exists in shard DDL; ⚠️ not yet in ETL SELECT (P2-1, P2-2) |
| `octo_fact_member_channel_content_daily.msg_count` | Classifier output → ETL aggregation | ✅ plan describes accumulation by `(day, channel, sender, category)` |
| `octo_fact_channel_content_daily` | Recomputed from member-channel content fact | ✅ plan mirrors existing dirty-day recompute pattern |
| API `human_msg_count` / `agent_msg_count` | `octo_fact_channel_content_daily` grouped by `sender_type` | ✅ consistent with schema |

---

**[Octo-Q] verdict: APPROVE**

This is a well-researched, documentation-only plan with all factual claims verified against the codebase. Five P2 findings identify implementation-phase considerations (ETL SELECT extension, payload I/O impact, RevokeMessage classification, backfill cursor DDL, signal value coverage). None are blockers for landing the plan document. The implementation split is logical and the design is internally consistent.

yujiawei (changes requested):
# Code Review — PR #327 (octo-server)

**Scope:** Documentation-only PR adding `docs/opanalytics-message-type-classification.md` (+291/-0). It is a design/spike proposal for the P1B message-type dimension and changes no runtime code, migrations, ETL, or API responses.

**Verdict: CHANGES_REQUESTED.** The plan is factually well-grounded against the current codebase, but two issues should be resolved in the document before it is adopted as the implementation reference: (1) an internal contradiction in the classification rules, and (2) an under-specified backfill ↔ incremental handoff that, as written, would double-count analytics. Both fixes are cheap and confined to the doc.

## What was verified against the code (all accurate ✅)

- **ContentType mapping** — every raw value named in the "Classification Categories" table matches `octo-lib common/msg.go` (the pinned `v0.0.0-20260605071834-330ed3ce7e6a`) exactly: Text=1, Image=2, GIF=3, Voice=4, Video=5, Location=6, Card=7, File=8, MultipleForward=11, VectorSticker=12, EmojiSticker=13, RichText=14, InviteJoinOrganization=16, ContentError=97, SignalError=98, CMD=99, FriendApply=1000…GroupUpgrade=1022, HotlineAssignTo/Solved/Reopen=1200-1202, Tip=2000, VideoCallResult=9989. No mismatched or orphaned enum constants.
- **Current-state claims** — `octo_fact_member_channel_daily.content_type` is a reserved column written as `0` (`modules/opanalytics/sql/20260608000001_opanalytics_init.sql:53`, comment "保留列: P0恒0"), and the PK is `(channel_id, stat_date, sender_uid)` (same file, `:59`). The "reusing the column collapses mixed types into one row" rationale is correct under that PK.
- **New schema** — both proposed tables add `msg_category` to the PK, which correctly removes the collapse problem; column types and index shapes are consistent with the existing fact tables. `msg_category` numbering 0=unknown..7=system_notification covers all 8 prose categories with no gap.
- **Message source** — the message row really has `Payload []byte` and `Signal int` (`modules/message/db.go:99,101`), and reading content type from `payload["type"]` as a JSON number matches existing usage (`modules/message/api.go:3074`), so the classifier rules are consistent with how the codebase already decodes payloads.
- **API contract** — `/v1/manager/dashboard` prefix, `AuthMiddleware` + `SharedUIDRateLimiter`, per-handler superAdmin checks, and i18n error envelopes all match `modules/opanalytics/api.go:54-64,217-223`. The claim that private `conv_type=3/4` rows are zeroed when `space_ids` is present matches existing behavior (`service.go` `zeroPrivateMessageComposition` / `zeroPrivateTrendConvTypes`).

## Required changes (blocking)

### 1. Resolve the classification contradiction for unsupported-but-parseable custom types
The `rich_custom` row (line 63) routes *"unknown app custom types that still parse cleanly"* to `rich_custom`, while the `unknown` row (line 65) and Classifier Rule 3 (lines 84-86) route any type *"outside the supported mapping" / "unsupported raw types"* to `unknown` with `type_unsupported`. For a custom numeric type (e.g. `50`) that parses cleanly but is not in the mapping, the two rules disagree. As written, an implementer cannot deterministically place it. Please pick one rule and make the table row and Rule 3 agree — e.g. either "only the explicitly-enumerated types map to `rich_custom`; everything else parseable-but-unmapped → `unknown`", or "any cleanly-parsing numeric type not matched elsewhere → `rich_custom`, reserving `unknown` for unparseable/encrypted/error". The choice has real coverage/`unknown_rate` consequences, so it belongs in the spec, not at implementation time.

### 2. Specify the backfill ↔ incremental handoff to prevent double-counting
The plan keeps the existing single-cursor incremental path (which writes content facts "in the same chunk pass") **and** adds "a separate content backfill cursor." The existing fact accumulation is `msg_count += count` upsert — exactly-once *only because there is one writer advancing one cursor* (`etl.go` / `etl_db.go runChunk`). With two writers over the same message id range and a non-idempotent additive upsert, any id processed by both the backfill scan and the incremental path is counted twice; a paused/resumed backfill that overlaps the incremental cursor compounds it. The doc should state the handoff explicitly — e.g. a fixed `backfill_high_water` id below which only backfill writes and above which only incremental writes, or backfill writing to a staging table merged once — so the "exactly-once" guarantee the rest of the design relies on actually holds. Right now this central correctness property is asserted for P0/P1A but not carried into the two-writer P1B world.

## Non-blocking suggestions (nits)

- **ETL source read extension (line 78-82 inputs):** the classifier consumes `signal` and `payload`, but the current `srcMessageRow` / source query reads neither (`model.go:63-73`, `etl_db.go:83-87` select only `id, from_uid, channel_id, channel_type, timestamp, created_at`). The "Implementation Split → Schema and ETL write path" step should call out that the source SELECT and `srcMessageRow` must be extended; it is implied but easy to miss, and `payload` is a `[]byte` blob that materially affects chunk memory/IO budget.
- **VideoCallResult / call-payload determinism (lines 67-70):** "can record them as `system_notification` or `unknown`" leaves it non-deterministic. The fourth Open Decision already flags the call-like case, so consider promoting it to a single concrete default (it affects `unknown_rate` reporting) rather than leaving classifier discretion.
- **PK key-order divergence (informational, not a bug):** the member-channel content table leads its PK with `channel_id` while the channel content table leads with `space_id`. This matches the divergence already present in the existing fact tables and is appropriate for the respective query shapes — noting it only so it is a conscious choice.

## Summary
Solid, accurate plan with no factual drift from the codebase and sound schema design. Blocking only on the two spec-level correctness items above; once the classification contradiction is resolved and the backfill/incremental handoff is pinned down, this is ready to serve as the P1B reference.

an9xyz:
Closing this doc-only spike PR because it does not match the corrected P0/P1 completion scope. The implementation is now in #329; message-type/P1B classification can be reopened later as a separate follow-up when needed.

PR #250, opened 2026-06-04
fix(incomingwebhook): resolve iwh_ sender identity for channel/user/avatar APIs

an9xyz:
Closes #251

## Background

Follow-up to #249 (incoming webhooks, #246). After #249 merged, verification found that clients which have not yet adapted to webhook messages render the synthetic sender UID (`iwh_xxx`) as a regular user and call the user/channel APIs, which all fail.

## Problem

The webhook sender `FromUID = iwh_xxx` never exists in the `user` table:

| Endpoint | Before |
|---|---|
| `GET /v1/channels/iwh_/:type` | `user not found` (400) |
| `GET /v1/users/iwh_` | `500 query_failed` |
| `GET /v1/users/iwh_/avatar` | `404` (broken image) |

`GET /v1/users/iwh_` returns 500 because `GetUserDetail` returns an error on a missing row and the handler maps any error to `query_failed` (this also surfaces a pre-existing not-found-as-500 mismatch, left untouched here).

## Fix

A server-side fallback recognizes the `iwh_` prefix and synthesizes display info (name/avatar) from the `incoming_webhook` table, reusing octo-lib's `register.BussDataSource.ChannelGet` so the `user` module keeps no compile-time dependency on the upper-layer `incomingwebhook` module. `token`/`token_hash` are never exposed.

- `incomingwebhook`: register a `ChannelGet` datasource; extract the `webhookIDPrefix` constant
- `user`: let `iwh_` pass through its own datasource; add `iwh_` branches to the `get` and `UserAvatar` handlers (302 redirect to a custom avatar URL, otherwise a default avatar — consistent with the existing real-user avatar redirect)
- Graceful degradation after webhook deletion (hard delete): not-found instead of 500, and a default avatar instead of 404

The authoritative render source remains `payload.from.name/avatar` embedded at send time; these endpoints are a transitional fallback until clients adapt.

## Test plan

- [x] `go build ./...`, `go vet`, `gofmt`
- [x] New unit tests (`display_test.go`): prefix detection, empty-name fallback, no token leak
- [x] New e2e tests (`webhook_render_test.go`, 9 cases): all three endpoints + post-deletion degradation + default avatar; passing with `-race`
- [x] `incomingwebhook` package regression (existing tests) pass

PR #31, opened 2026-05-14
feat(incomingwebhook): Discord-style group incoming webhooks

an9xyz:
## Summary

Add `modules/incomingwebhook`: group incoming webhooks that let external services push messages to a group via a token-bearing URL (Discord/Slack-style semantics).

- Group owners and managers create webhooks through an authenticated API and receive a one-time token + URL; only the SHA-256(token) is persisted.
- Push endpoint accepts JSON (`content` / `username` / `avatar_url`) and delivers as a Text message with `from.kind=webhook` so clients can render the virtual sender without a real user profile.
- Revoke permission follows the existing message module: owners and managers can revoke webhook messages; regular members (including the webhook creator) cannot — matching Discord semantics.

## Defenses

- Token comparison via `crypto/subtle.ConstantTimeCompare`.
- Rate limiting: per-IP and per-webhook token buckets (Redis Lua, fail-open on Redis failure).
- Caps: 8 KB body (`io.LimitReader` + 1-byte overflow detection); per-group quota (default 10, env-tunable) enforced atomically inside a transaction with `SELECT ... FOR UPDATE` on the (group_no, status) index — concurrent creators are serialized so the configured cap holds.
- Audit log of successful pushes (`incoming_webhook_audit`).
- Lifecycle: listens for `event.GroupDisband` to cascade-disable; `create` / `update(status=1)` / `regenerate` / `push` all gated by `requireActiveGroup` to prevent disband → re-enable revival.
- Payload: `space_id` is authoritatively derived from the group row server-side. Caller-supplied `extra` is dropped wholesale — `payload.visibles` is a server-enforced visibility allowlist in the message module (non-listed users see the message as deleted / 404), so an unauthenticated token holder must not be able to set it. Future metadata fields must be added by explicit name only.
- Push-side `username` / `avatar_url` are server-truncated in `buildPayload` to the same byte limits as the create-side columns (64 / 255), with UTF-8 rune-boundary fallback. The push path is otherwise only bounded by the 8 KB body cap, so without this a token holder could stuff KB-class strings into `from.*` and pollute every client's render path.
- `updateFields` enforces a column allowlist (`name` / `avatar` / `status` / `token_hash`) so a future caller passing user-controlled keys cannot drive an arbitrary column update.

## Cross-module hardening (admin role semantics)

[...]

PR #330, opened 2026-06-10
feat(incomingwebhook): GitHub / WeCom platform adapters on push (Phase 3)

an9xyz:
## Background

**Phase 3** of epic #297 (scope adjusted per maintainer: GitHub + WeCom instead of GitHub + GitLab, in a single PR). Today every caller must hand-convert third-party events into our `content` format. This PR adds out-of-the-box platform adapters plus the onboarding item deferred from Phase 2.

## Changes

**Adapter pipeline**
- The push handler is refactored into `handlePush(c, pushAdapter)`: all three shapes (native / github / wecom) share the same auth / 4-layer rate-limit / group-check / audit pipeline and differ **only** in body parsing (`adapter.go`). Adapters emit a `pushPayloadReq`, never touch the message payload directly — `from.kind=webhook` + server-derived `space_id` injection is unchanged.
- New routes `POST .../:webhook_id/:token/github` and `.../wecom` reuse the exact middleware chain (same limiter instances/buckets — adapters are not a new attack surface).

**GitHub adapter** (`adapter_github.go`)
- Renders by `X-GitHub-Event` into markdown: `push` (branch/tag/delete/force-push, ≤5 commits listed), `pull_request` (opened/closed→merged/reopened/ready_for_review), `issues`, `issue_comment` (created), `release` (published).
- `ping` and out-of-subset events/actions answer **200 + `skipped`** and are audited with the new **status=3 (skipped)** — GitHub shows green deliveries (no false red), the group is not spammed (e.g. `pull_request.synchronize`), and admins still see the traffic in `deliveries`. A missing `X-GitHub-Event` header is a hard 400 `reason=event` (misconfiguration must be visible).
- Platform-generated fields (titles / commit messages / comment bodies) are clipped server-side so GitHub traffic can never hit the 413 content cap — the sender has no way to "shorten" an event.
- Auth stays the 128-bit URL token per the epic decision (no mandatory HMAC; `X-Hub-Signature-256` left as a future option).

**WeCom adapter** (`adapter_wecom.go`)
- Accepts the WeCom group-bot **outbound** format: tools already pushing to a WeCom bot migrate by changing only the URL. `text` / `markdown` / `markdown_v2` pass through; `news` and `template_card` degrade to markdown (documented, per the epic's degrade decision); media-based types (`image`/`file`/`voice`) are rejected with 400 `reason=msg_type` — an explicit failure beats silently dropping a message the sender believes was delivered.
- Success responses additionally carry `errcode=0` / `errmsg=ok` (most WeCom SDKs key off `errcode`).

**Onboarding (deferred Phase 2 item)**
- `create` / `regenerate` responses now include `urls` (`native` / `github` / `wecom`); legacy `url` field unchanged. README documents all shapes with curl examples.

**Audit**
- `adapter` column gains `github` / `wecom`; failure records now carry the real adapter; new `auditSkipped(3)` status. Migration is **comment-only** (INSTANT, no structural change). 429-not-audited and the pre-auth anti-enumeration invariant are untouched.

## Testing (TDD)

- Unit (no infra): GitHub renderers per event/action incl. skip/invalid gates, clipping, commit-list truncation; WeCom all msgtypes incl. every rejection path; adapter plumbing (native parse, success body, helpers).
- e2e (`adapter_push_test.go`): ping → 200+skipped **and** async audit row asserted via deliveries (`adapter=github, status=3, reason=ping`); push/wecom-text deliver; bad token → uniform 401 on adapter routes; missing event header → 400; wecom image → 400; create carries `urls`.
- Guard: `adapter*.go` added to `TestIncomingWebhookNoLegacyResponseError`.

## Checks

`go build` / `go vet` / `gofmt` / `make i18n-extract-check` / `make i18n-lint` pass (no new error codes — `push_payload_invalid` reused with new `reason=event`). Full module suite ran **locally against real MySQL 8 + Redis with a WuKongIM stub** (CI-identical DB recreate + FLUSHALL, `go test -race -shuffle=on`): ok in 33.7s, including the success-path assertions (`message_id`, `errcode:0`) that the CI stub sometimes degrades to 502 on.

Refs #297

> Sprint: tracked on the Octo Board under **Sprint W24** (intentionally `Refs`, not `Closes` — the epic stays open for the GitLab / Feishu phases).

https://claude.ai/code/session_01W4UKE63aRUDm1i2JGN6xSf

### DIFF
--- a/modules/incomingwebhook/README.md
+++ b/modules/incomingwebhook/README.md
@@ -4,12 +4,16 @@
 由群主或管理员调用，详见 `api.go`。本文聚焦**推送端点**的请求契约。
 
 ```
-POST /v1/incoming-webhooks/:webhook_id/:token
+POST /v1/incoming-webhooks/:webhook_id/:token            # native（本文主体）
+POST /v1/incoming-webhooks/:webhook_id/:token/github     # GitHub 事件适配器
+POST /v1/incoming-webhooks/:webhook_id/:token/wecom      # 企业微信群机器人格式适配器
 Content-Type: application/json
 ```
 
 鉴权走 URL 内的 token（SHA-256 存储、常量时间比对），无需登录态。所有鉴权失败统一
-返回 401（反枚举），并受多层限流约束。
+返回 401（反枚举），并受多层限流约束。三种形态共享同一条鉴权/限流/群校验链，适配器
+只是 body 解析不同（见下文「平台适配器」）。create/regenerate 响应的 `urls` 字段
+按形态给出全部三个路径。
 
 ## 消息形态
 
@@ -71,6 +75,64 @@ Content-Type: application/json
 - 服务端另有 1MB 的 RichText 硬上限（octo-lib 契约）兜底，但在默认 8KB body cap 下不会
   先触达——它是上调 body cap 后才会成为约束的二级护栏。
 
+## 平台适配器（#297 Phase 3）
+
+适配器把第三方平台的原生格式翻译成上面的 native 消息，鉴权/限流/审计与 native 完全
+一致。适配器消息不支持 `username`/`avatar_url` 覆盖（展示身份固定为 webhook 配置）。
+
+### GitHub
+
+```
+POST /v1/incoming-webhooks/:webhook_id/:token/github
+```
+
+在 GitHub 仓库 **Settings → Webhooks** 把 Payload URL 配成上述地址、Content type 选
+`application/json` 即可。鉴权靠 URL 内 128-bit token（不强制 HMAC；`X-Hub-Signature-256`
+校验留作后续可选项）。
+
+按 `X-GitHub-Event` 渲染为 markdown，当前渲染子集：
+
+| 事件 | 渲染的动作 | 说明 |
+|------|-----------|------|
+| `ping` | — | 返回 200 不发消息（GitHub 创建 webhook 时的连通性测试） |
+| `push` | — | 分支/标签 push、删除、force-push；最多列 5 条提交 |
+| `pull_request` | `opened` / `closed`(含 merged) / `reopened` / `ready_for_review` | `synchronize` 等刷屏动作跳过 |
+| `issues` | `opened` / `closed` / `reopened` | |
+| `issue_comment` | `created` | 评论摘要压成单行、截断 300 rune |
+| `release` | `published` | |
+
+**子集之外的事件/动作返回 200 + `{"skipped":"event"}`**（GitHub 侧显示投递成功、
+不标红；deliveries 里以 `status=3` 可见），缺 `X-GitHub-Event` 头则按 400
+`reason=event` 拒绝。事件里的超长字段（标题/提交信息/评论）服务端截断，GitHub 流量
+不会触发 413。
+
+### 企业微信（WeCom 群机器人格式）
+
+```
+POST /v1/incoming-webhooks/:webhook_id/:token/wecom
+```
+
+接受企业微信「群机器人」的出站消息格式——已配置向企微机器人推送的工具只需**换 URL**
+即可迁移，消息体零改动。成功响应附带 `errcode=0`/`errmsg=ok`（多数企微 SDK 以此判定
+成功）。
+
+| `msgtype` | 处理 |
+|-----------|------|
+| `text` / `markdown` / `markdown_v2` | → 文本消息（客户端按 markdown 渲染）；`mentioned_list`/`mentioned_mobile_list` 降级丢弃 |
+| `news` | 降级 markdown：每篇文章「标题链接 + 描述」一段；`picurl` 丢弃 |
+| `template_card` | 降级 markdown：主标题 + 描述 + 副标题 + 跳转链接；按钮等交互元素丢弃 |
+| `image` / `file` / `voice` 等素材类 | **400 `reason=msg_type`**：base64/media_id 素材无法转存，显式失败优于静默丢弃 |
+
+> 高保真卡片渲染不可行，降级策略经 #297 确认。`content` 超过语义上限（4000 rune）按
+> 既有 413 拒绝——与 GitHub 适配器不同，企微格式的消息体由调用方自行编写，可以修短。
+
+```bash
+# 迁移示例：把企微机器人 URL 换成 octo 即可
+curl -X POST "$BASE/v1/incoming-webhooks/$WEBHOOK_ID/$TOKEN/wecom" \
+  -H 'Content-Type: application/json' \
+  -d '{"msgtype":"markdown","markdown":{"content":"**Build #123** passed"}}'
+```
+
 ## 通用字段与安全
 
 - `username` / `avatar_url`：两种形态通用，服务端裁剪到字节上限（名 64B / 头像 255B）。
@@ -81,10 +143,11 @@ Content-Type: application/json
 
 | 场景 | HTTP | 说明 |
 |------|------|------|
-| 成功 | 200 | `{"status":0,"message_id":<int>}` |
+| 成功 | 200 | `{"status":0,"message_id":<int>}`（wecom 路由额外带 `errcode`/`errmsg`） |
+| 已接收、刻意不投递 | 200 | `{"status":0,"message_id":0,"skipped":"ping"\|"event"}`（仅适配器路由） |
 | 鉴权失败 | 401 | 统一响应，不区分原因（反枚举） |
 | 限流 | 429 | 带 `Retry-After` |
-| 请求非法 | 400 | `details.reason` ∈ `body`/`json`/`content`/`blocks`/`msg_type` |
+| 请求非法 | 400 | `details.reason` ∈ `body`/`json`/`content`/`blocks`/`msg_type`/`event` |
 | 体积过大 | 413 | 超 body cap 或富文本 >1MB |
 | 投递失败 | 502 | 下游发送失败 |
 | 功能停用 | 404 | 全局开关 `incomingwebhook.enabled=0` |
@@ -92,6 +155,11 @@ Content-Type: application/json
 ## 管理端点（群主 / 管理员）
 
 需登录态 + 群管理员权限，路径前缀 `/v1/groups/:group_no/incoming-webhooks`。
+
+创建 / 重置（regenerate）响应除历史的 `url`（native 路径）外，还带 `urls` 对象，
+按推送形态给出全部路径（`native` / `github` / `wecom`，不含 host，由前端拼接）。
+token 仅在这两处出现一次，list 不回显 token、也不回推送 URL。
+
 除创建/列出/更新/删除/重置外，Phase 2 新增两个：
 
 ### 测试推送
@@ -129,10 +197,11 @@ GET /v1/groups/:group_no/incoming-webhooks/:webhook_id/deliveries?limit=50
 }
 ```
 
-- `status`：`1`=成功，`2`=失败。
-- `reason`（失败时）：`body` / `json` / `content` / `blocks` / `msg_type` / `too_large` /
-  `delivery_failed`。
-- `adapter`：`native`（推送端点）/ `test`（测试推送）。
+- `status`：`1`=成功，`2`=失败，`3`=跳过（已接收、刻意不投递：GitHub `ping` / 渲染
+  子集之外的事件，响应 200 但没有消息进群）。
+- `reason`（失败/跳过时）：`body` / `json` / `content` / `blocks` / `msg_type` /
+  `too_large` / `delivery_failed` / `event` / `ping`。
+- `adapter`：`native`（推送端点）/ `test`（测试推送）/ `github` / `wecom`（平台适配器）。
 - `http_status`：返回给调用方的状态码。**迁移前的历史成功行为 `0`（未知）**——不伪造成 200。
 - **不返回调用方 `ip`**：审计表仍存 ip 作排查上下文，但出于隐私不向群管理员下发（review 决定）。
 
--- a/modules/incomingwebhook/README.md
+++ b/modules/incomingwebhook/README.md
@@ -106,6 +106,11 @@ POST /v1/incoming-webhooks/:webhook_id/:token/github
 `reason=event` 拒绝。事件里的超长字段（标题/提交信息/评论）服务端截断，GitHub 流量
 不会触发 413。
 
+**body 上限独立于 native**：GitHub 事件 JSON 由平台生成（真实 push/PR 事件普遍
+>8KiB，发送方无法修短），github 路由的请求体上限默认 **1MiB**
+（`DM_INCOMINGWEBHOOK_GITHUB_MAX_BYTES`）；native/wecom 的 body 由调用方编写，
+仍是 8KiB。该读取发生在 token 鉴权 + per-webhook 限流之后，不构成放大面。
+
 ### 企业微信（WeCom 群机器人格式）
 
 ```
--- a/modules/incomingwebhook/adapter.go
+++ b/modules/incomingwebhook/adapter.go
@@ -1,0 +1,106 @@
+package incomingwebhook
+
+// 推送形态适配层（#297 Phase 3）。
+//
+// 三种推送形态共享同一条鉴权 / 限流 / 群校验 / 投递 / 审计流水线（api.go handlePush），
+// 彼此只差「如何把请求 body 翻译成 native 推送请求」这一步：
+//
+//   - native（历史契约）   POST /v1/incoming-webhooks/:webhook_id/:token
+//   - GitHub 事件          POST .../:token/github   （adapter_github.go）
+//   - 企业微信群机器人格式  POST .../:token/wecom    （adapter_wecom.go）
+//
+// 适配器不是新的攻击面：URL token 鉴权、四层限流、群 Normal 校验、payload 白名单
+// 构造（buildPayload / buildRichTextPayload 注入 from.kind=webhook 与服务端 space_id）
+// 全部复用，适配器只产出 pushPayloadReq（content / blocks），不直接触达消息 payload。
+
+import (
+	"encoding/json"
+	"net/http"
+	"strings"
+)
+
+// pushAdapter 描述一种推送形态。
+type pushAdapter struct {
+	// name 写入审计 adapter 列（adapterNative / adapterGitHub / adapterWeCom）。
+	name string
+	// parse 把平台原始 body 翻译成 native 推送请求。三个返回值恰有一个生效：
+	//   - req     非 nil：照常走 msg_type 构造 / 投递；
+	//   - skip    非空：请求合法但刻意不投递（GitHub ping / 渲染子集之外的事件），
+	//     返回 200 并以 auditSkipped 落审计，供管理端 deliveries 观察链路；
+	//   - invalid 非空：解析失败原因码，映射 400 invalid(reason=...) 并落审计。
+	parse func(header http.Header, body []byte) (req *pushPayloadReq, skip string, invalid string)
+	// successExtra 合并进成功 / skip 响应体的平台兼容字段（如企业微信的 errcode /
+	// errmsg），让按平台 SDK 校验响应的既有工具不改代码即可迁移。key 与 native 的
+	// status / message_id 不重叠，纯追加。
+	successExtra map[string]interface{}
+}
+
+var (
+	nativeAdapter = pushAdapter{name: adapterNative, parse: parseNativePush}
+	githubAdapter = pushAdapter{name: adapterGitHub, parse: parseGitHubPush}
+	wecomAdapter  = pushAdapter{
+		name:  adapterWeCom,
+		parse: parseWeComPush,
+		// 企业微信调用方普遍校验 errcode==0，附带平台习惯字段降低迁移摩擦。
+		successExtra: map[string]interface{}{"errcode": 0, "errmsg": "ok"},
+	}
+)
+
+// parseNativePush 是 native 形态的 parse：body 即 pushPayloadReq JSON 本身。
+func parseNativePush(_ http.Header, body []byte) (*pushPayloadReq, string, string) {
+	var req pushPayloadReq
+	if err := json.Unmarshal(body, &req); err != nil {
+		return nil, "", "json"
+	}
+	return &req, "", ""
+}
+
+// successBody 构造推送成功（或 skip）的 200 响应体；skipped 非空时附带说明字段，
+// 让调用方能区分「已投递」与「已接收但刻意不投递」。
+func successBody(ad pushAdapter, msgID int64, skipped string) map[string]interface{} {
+	body := map[string]interface{}{
+		"status":     0,
+		"message_id": msgID,
+	}
+	if skipped != "" {
+		body["skipped"] = skipped
+	}
+	for k, v := range ad.successExtra {
+		body[k] = v
+	}
+	return body
+}
+
+// ============================================================
+// 渲染文本工具（GitHub / WeCom 适配器共用）
+// ============================================================
+
+// clipRunes 按 rune 数截断并以省略号收尾。平台事件里的标题 / 提交信息 / 评论长度
+// 不受我们控制，渲染前必须钳制——adapter 产出的 content 一旦超过 maxContentRunes
+// 会被 push 路径按 413 拒绝，而平台调用方没有任何手段「修短」一个 GitHub 事件。
+func clipRunes(s string, max int) string {
+	if max <= 0 {
+		return ""
+	}
+	runes := []rune(s)
+	if len(runes) <= max {
+		return s
+	}
+	return string(runes[:max-1]) + "…"
+}
+
+// firstLine 取首行（提交信息惯例：首行即摘要）。
+func firstLine(s string) string {
+	if i := strings.IndexByte(s, '\n'); i >= 0 {
+		s = s[:i]
+	}
+	return strings.TrimSpace(s)
+}
+
+// oneLine 把多行文本压成单行，避免标题 / 评论里的换行破坏 markdown 链接结构。
+func oneLine(s string) string {
+	s = strings.ReplaceAll(s, "\r\n", " ")
+	s = strings.ReplaceAll(s, "\n", " ")
+	s = strings.ReplaceAll(s, "\r", " ")
+	return strings.TrimSpace(s)
+}
--- a/modules/incomingwebhook/adapter.go
+++ b/modules/incomingwebhook/adapter.go
@@ -33,14 +33,20 @@ type pushAdapter struct {
 	// errmsg），让按平台 SDK 校验响应的既有工具不改代码即可迁移。key 与 native 的
 	// status / message_id 不重叠，纯追加。
 	successExtra map[string]interface{}
+	// bodyLimit 该形态的请求体字节上限。native / wecom 的 body 由调用方编写，沿用
+	// 8KiB 的 maxBytes()——上限本就是约束调用方的；github 的 body 是平台生成的事件
+	// JSON，真实 push / PR 事件普遍 >8KiB 且发送方无法修短，必须用更宽的专属上限
+	//（githubMaxBytes，见 adapter_github.go；PR #330 review 阻断项）。
+	bodyLimit func() int
 }
 
 var (
-	nativeAdapter = pushAdapter{name: adapterNative, parse: parseNativePush}
-	githubAdapter = pushAdapter{name: adapterGitHub, parse: parseGitHubPush}
+	nativeAdapter = pushAdapter{name: adapterNative, parse: parseNativePush, bodyLimit: maxBytes}
+	githubAdapter = pushAdapter{name: adapterGitHub, parse: parseGitHubPush, bodyLimit: githubMaxBytes}
 	wecomAdapter  = pushAdapter{
-		name:  adapterWeCom,
-		parse: parseWeComPush,
+		name:      adapterWeCom,
+		parse:     parseWeComPush,
+		bodyLimit: maxBytes,
 		// 企业微信调用方普遍校验 errcode==0，附带平台习惯字段降低迁移摩擦。
 		successExtra: map[string]interface{}{"errcode": 0, "errmsg": "ok"},
 	}
--- a/modules/incomingwebhook/adapter_github.go
+++ b/modules/incomingwebhook/adapter_github.go
@@ -1,0 +1,291 @@
+package incomingwebhook
+
+// GitHub 事件适配器（#297 Phase 3）。
+//
+// 路由：POST /v1/incoming-webhooks/:webhook_id/:token/github
+// 在 GitHub 仓库 Settings → Webhooks 把 Payload URL 配成上述地址、Content type 选
+// application/json 即可，无需任何中间转换层。
+//
+// 鉴权沿用 URL 内的 128-bit token（与 native 一致；经 #297 确认不强制 HMAC——
+// X-Hub-Signature-256 校验留作后续可选项，参考 modules/webhook/hmac.go）。
+//
+// 渲染策略：按 X-GitHub-Event 把常用事件翻译成 markdown 文本（走 native 纯文本路径，
+// 客户端按 markdown 渲染）。刻意只渲染「人关心的」动作子集——例如 pull_request 的
+// synchronize（PR 分支每次 push 都触发）若也进群会刷屏。子集之外的事件 / 动作返回
+// 200 并以 auditSkipped 落审计：GitHub 侧显示绿色投递成功（不会把 webhook 标红），
+// 管理端 deliveries 里 reason=event 可见，两边都不糊弄。
+//
+// 所有 gh* 结构体只声明渲染需要的字段（白名单解析），其余 payload 字段一律忽略。
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+// 渲染的提交列表上限：push 事件单次最多带 20 个 commit（GitHub 截断），全列会刷屏。
+const maxRenderedCommits = 5
+
+type ghUser struct {
+	Login string `json:"login"`
+}
+
+type ghRepo struct {
+	FullName string `json:"full_name"`
+	HTMLURL  string `json:"html_url"`
+}
+
+type ghCommit struct {
+	ID      string `json:"id"`
+	Message string `json:"message"`
+	URL     string `json:"url"`
+}
+
+type ghPushEvent struct {
+	Ref        string     `json:"ref"`
+	Created    bool       `json:"created"`
+	Deleted    bool       `json:"deleted"`
+	Forced     bool       `json:"forced"`
+	Commits    []ghCommit `json:"commits"`
+	Repository ghRepo     `json:"repository"`
+	Sender     ghUser     `json:"sender"`
+}
+
+type ghPullRequestEvent struct {
+	Action      string `json:"action"`
+	PullRequest struct {
+		Number  int    `json:"number"`
+		Title   string `json:"title"`
+		HTMLURL string `json:"html_url"`
+		Merged  bool   `json:"merged"`
+	} `json:"pull_request"`
+	Repository ghRepo `json:"repository"`
+	Sender     ghUser `json:"sender"`
+}
+
+type ghIssuesEvent struct {
+	Action string `json:"action"`
+	Issue  struct {
+		Number  int    `json:"number"`
+		Title   string `json:"title"`
+		HTMLURL string `json:"html_url"`
+	} `json:"issue"`
+	Repository ghRepo `json:"repository"`
+	Sender     ghUser `json:"sender"`
+}
+
+type ghIssueCommentEvent struct {
+	Action string `json:"action"`
+	Issue  struct {
+		Number  int    `json:"number"`
+		Title   string `json:"title"`
+		HTMLURL string `json:"html_url"`
+	} `json:"issue"`
+	Comment struct {
+		HTMLURL string `json:"html_url"`
+		Body    string `json:"body"`
+	} `json:"comment"`
+	Repository ghRepo `json:"repository"`
+	Sender     ghUser `json:"sender"`
+}
+
+type ghReleaseEvent struct {
+	Action  string `json:"action"`
+	Release struct {
+		TagName string `json:"tag_name"`
+		Name    string `json:"name"`
+		HTMLURL string `json:"html_url"`
+	} `json:"release"`
+	Repository ghRepo `json:"repository"`
+	Sender     ghUser `json:"sender"`
+}
+
+// parseGitHubPush 把 GitHub webhook 事件翻译成 native 推送请求（pushAdapter.parse）。
+func parseGitHubPush(header http.Header, body []byte) (*pushPayloadReq, string, string) {
+	event := strings.TrimSpace(header.Get("X-GitHub-Event"))
+	if event == "" {
+		// 不带事件头的请求不可能来自 GitHub——按非法请求拒绝而非静默跳过，
+		// 让误把 native 流量打到 /github 后缀的调用方立刻发现配置错误。
+		return nil, "", "event"
+	}
+
+	var content string
+	var err error
+	switch event {
+	case "ping":
+		// GitHub 创建 webhook 时的连通性测试：200 即可，不发消息。
+		return nil, "ping", ""
+	case "push":
+		content, err = renderGitHubPush(body)
+	case "pull_request":
+		content, err = renderGitHubPullRequest(body)
+	case "issues":
+		content, err = renderGitHubIssues(body)
+	case "issue_comment":
+		content, err = renderGitHubIssueComment(body)
+	case "release":
+		content, err = renderGitHubRelease(body)
+	default:
+		// 渲染子集之外的事件类型：通常只是 GitHub 侧订阅范围大于我们渲染的子集，
+		// 调用方无需修复 → 200 + skipped（见文件头注释）。
+		return nil, "event", ""
+	}
+	if err != nil {
+		return nil, "", "json"
+	}
+	if content == "" {
+		// 事件类型支持、但动作不在渲染子集内（synchronize / labeled / ...）：同上 skip。
+		return nil, "event", ""
+	}
+	// 事件体里的标题 / 提交信息长度不受我们控制：钳到语义上限内，绝不让平台流量 413。
+	return &pushPayloadReq{Content: clipRunes(content, maxContentRunes())}, "", ""
+}
+
+func renderGitHubPush(body []byte) (string, error) {
+	var ev ghPushEvent
+	if err := json.Unmarshal(body, &ev); err != nil {
+		return "", err
+	}
+	who := ghLogin(ev.Sender)
+	ref := ghShortRef(ev.Ref)
+	switch {
+	case strings.HasPrefix(ev.Ref, "refs/tags/"):
+		if ev.Deleted {
+			return ghWithRepo(fmt.Sprintf("**%s** deleted tag `%s`", who, ref), ev.Repository), nil
+		}
+		return ghWithRepo(fmt.Sprintf("**%s** pushed tag `%s`", who, ref), ev.Repository), nil
+	case ev.Deleted:
+		return ghWithRepo(fmt.Sprintf("**%s** deleted branch `%s`", who, ref), ev.Repository), nil
+	case ev.Created && len(ev.Commits) == 0:
+		return ghWithRepo(fmt.Sprintf("**%s** created branch `%s`", who, ref), ev.Repository), nil
+	}
+
+	verb := "pushed"
+	if ev.Forced {
+		verb = "force-pushed"
+	}
+	var b strings.Builder
+	b.WriteString(ghWithRepo(
+		fmt.Sprintf("**%s** %s %d commit(s) to `%s`", who, verb, len(ev.Commits), ref),
+		ev.Repository))
+	for i, cm := range ev.Commits {
+		if i == maxRenderedCommits {
+			fmt.Fprintf(&b, "\n- …and %d more", len(ev.Commits)-maxRenderedCommits)
+			break
+		}
+		fmt.Fprintf(&b, "\n- [`%s`](%s) %s", ghShortSHA(cm.ID), cm.URL, clipRunes(firstLine(cm.Message), 120))
+	}
+	return b.String(), nil
+}
+
+func renderGitHubPullRequest(body []byte) (string, error) {
+	var ev ghPullRequestEvent
+	if err := json.Unmarshal(body, &ev); err != nil {
+		return "", err
+	}
+	var verb string
+	switch ev.Action {
+	case "opened", "reopened":
+		verb = ev.Action
+	case "ready_for_review":
+		verb = "marked ready for review:"
+	case "closed":
+		verb = "closed"
+		if ev.PullRequest.Merged {
+			verb = "merged"
+		}
+	default:
+		// synchronize / labeled / review_requested / ... 刷屏动作不渲染 → skip。
+		return "", nil
+	}
+	return ghWithRepo(fmt.Sprintf("**%s** %s pull request [#%d %s](%s)",
+		ghLogin(ev.Sender), verb, ev.PullRequest.Number,
+		clipRunes(oneLine(ev.PullRequest.Title), 200), ev.PullRequest.HTMLURL),
+		ev.Repository), nil
+}
+
+func renderGitHubIssues(body []byte) (string, error) {
+	var ev ghIssuesEvent
+	if err := json.Unmarshal(body, &ev); err != nil {
+		return "", err
+	}
+	switch ev.Action {
+	case "opened", "closed", "reopened":
+	default:
+		return "", nil
+	}
+	return ghWithRepo(fmt.Sprintf("**%s** %s issue [#%d %s](%s)",
+		ghLogin(ev.Sender), ev.Action, ev.Issue.Number,
+		clipRunes(oneLine(ev.Issue.Title), 200), ev.Issue.HTMLURL),
+		ev.Repository), nil
+}
+
+func renderGitHubIssueComment(body []byte) (string, error) {
+	var ev ghIssueCommentEvent
+	if err := json.Unmarshal(body, &ev); err != nil {
+		return "", err
+	}
+	if ev.Action != "created" {
+		return "", nil
+	}
+	line := ghWithRepo(fmt.Sprintf("**%s** commented on [#%d %s](%s)",
+		ghLogin(ev.Sender), ev.Issue.Number,
+		clipRunes(oneLine(ev.Issue.Title), 200), ev.Comment.HTMLURL),
+		ev.Repository)
+	if snippet := clipRunes(oneLine(ev.Comment.Body), 300); snippet != "" {
+		line += "\n> " + snippet
+	}
+	return line, nil
+}
+
+func renderGitHubRelease(body []byte) (string, error) {
+	var ev ghReleaseEvent
+	if err := json.Unmarshal(body, &ev); err != nil {
+		return "", err
+	}
+	if ev.Action != "published" {
+		return "", nil
+	}
+	title := ev.Release.Name
+	if strings.TrimSpace(title) == "" {
+		title = ev.Release.TagName
+	}
+	return ghWithRepo(fmt.Sprintf("**%s** published release [%s](%s)",
+		ghLogin(ev.Sender), clipRunes(oneLine(title), 200), ev.Release.HTMLURL),
+		ev.Repository), nil
+}
+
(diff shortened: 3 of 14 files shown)


mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#330 Review Report
**PR:** [Mininglamp-OSS/octo-server #330](https://github.com/Mininglamp-OSS/octo-server/pull/330)
**Head SHA:** `5b1b16f7dd22e3813bf7edab84df83fbdd613b26`
**Scope:** feat(incomingwebhook): GitHub / WeCom platform adapters on push (Phase 3)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Adapter pipeline architecture | ✅ | `adapter.go` L26-47: `pushAdapter` struct with `name`/`parse`/`successExtra`; three instances (`nativeAdapter`/`githubAdapter`/`wecomAdapter`) |
| Middleware chain reuse | ✅ | `api.go` L200-208: `chain` closure wraps all 3 routes with identical `requirePushEnabled → localFloor → ipLimit → ipFailureGate` |
| Auth invariant preserved | ✅ | `api.go` L843-893: `handlePush` performs token lookup + constant-time compare + group-normal check before adapter parse; adapter routes share this path |
| Rate-limit bucket sharing | ✅ | `api.go` L907: `allowPerWebhook` keyed by `webhookID` (not route suffix); all 3 suffixes hit same bucket |
| Data-flow: adapter → buildPayload | ✅ | `adapter.go` L57-62: `parseNativePush` returns `*pushPayloadReq`; `api.go` L941: `ad.parse()` → `req *pushPayloadReq` → `buildPayload(m, req)` (L975) / `buildRichTextPayload(m, req)` (L985) — type-correct, same pointer semantics as pre-refactor `&req` |
| GitHub content clipping → 413 guard | ✅ | `adapter_github.go` L148: `clipRunes(content, maxContentRunes())` ensures adapter output ≤ maxContentRunes; downstream `utf8.RuneCountInString(req.Content) > maxContentRunes()` (api.go L969) is a no-op for GitHub path |
| WeCom content: no clip, 413 downstream | ✅ | `adapter_wecom.go` L82-113: content passed verbatim; `api.go` L969 catches oversize → 413 (caller can shorten, unlike GitHub events) |
| Audit: skipped status | ✅ | `model.go` L43: `auditSkipped=3`; `api.go` L1139-1153: `submitSkipped` with `bumpUsed=false`, `HTTPStatus=200`, correct adapter name |
| SQL migration safety | ✅ | `sql/20260610000001_...sql`: COMMENT-only `MODIFY COLUMN`, INSTANT algorithm on MySQL 8.0, no structural change, down-migration provided |
| Response body correctness | ✅ | `adapter.go` L68-80: `successBody` — native gets `{status,message_id}`; WeCom gets `+errcode/errmsg`; skip gets `+skipped`; no key collision |
| Test coverage | ✅ | 4 test files: unit (adapter_test, adapter_github_test, adapter_wecom_test) + e2e (adapter_push_test); covers all render paths, skip/invalid gates, clipping, auth, URL surfacing |
| Legacy guard updated | ✅ | `api_i18n_test.go` L22: `adapter*.go` added to `TestIncomingWebhookNoLegacyResponseError` banned-file list |

## 2. Findings

### No P0/P1 findings.

### P2 — Minor observations (non-blocking)

**P2-1: `renderGitHubPush` renders "0 commit(s)" for degenerate push events**
- **File:** `adapter_github.go` L175-183
- **Detail:** A push event with `deleted=false`, `created=false`, no `refs/tags/` prefix, and `len(commits)==0` falls through to the generic "pushed N commit(s)" path, rendering `**someone** pushed 0 commit(s) to \`main\``. This is a degenerate case unlikely from real GitHub payloads (a push with no commits and no create/delete is essentially impossible), but the rendering is slightly awkward.
- **Severity:** P2/nit — cosmetic, no user-visible harm, degenerate input.

**P2-2: `successExtra` map iteration order in `successBody` is non-deterministic**
- **File:** `adapter.go` L78-79
- **Diff-scope:** new
- **Detail:** `for k, v := range ad.successExtra` iterates a `map[string]interface{}`. For WeCom (2 keys: `errcode`/`errmsg`), JSON serialization order varies across calls. Not a functional issue — JSON objects are unordered by spec — but could cause minor confusion in log comparison or test snapshot diffs. Current tests assert individual keys, so this is fine.
- **Severity:** P2/nit — no functional impact.

**P2-3: No `X-Hub-Signature-256` validation for GitHub adapter**
- **File:** `adapter_github.go` (entire file)
- **Diff-scope:** pre-existing design decision (documented in PR body and file header)
- **Detail:** GitHub webhook authenticity relies solely on the 128-bit URL token. HMAC signature verification is deferred. This is an explicit design choice per epic #297, not an oversight. The URL token provides adequate security for the use case (token is secret, HTTPS transport).
- **Severity:** P2/note — documented decision, acceptable risk.

## 3. Suggestions

1. **P2-1 fix (optional):** Add a guard in `renderGitHubPush` for `len(ev.Commits)==0 && !ev.Created && !ev.Deleted` — either skip rendering (return `"", nil`) or render a different message like "updated branch". Low priority given the degenerate input.

2. **Future consideration:** When `X-Hub-Signature-256` support is added, consider making it opt-in per webhook (stored flag on `incomingWebhookModel`) rather than global, to avoid breaking existing GitHub integrations.

## 4. Additional Findings

- **No cross-file blind spots detected.** The adapter layer is cleanly isolated: adapters produce `pushPayloadReq` and never touch `buildPayload`/`buildRichTextPayload`/`SendMessageWithResult`. The `from.kind=webhook` + server-derived `space_id` injection (api.go L1061-1078) is unchanged and applies uniformly to all adapter outputs.
- **Anti-enumeration invariant intact.** Auth failure on adapter routes returns the same uniform 401 as native (api.go L843-893); no new information leakage via adapter-specific error paths.
- **Audit semaphore pressure unchanged.** `submitSkipped` uses the same `submitDelivery` bounded-concurrency path (api.go L1163-1175). High-volume GitHub `ping`/skip events will consume audit slots identically to success/failure records — this is by design and bounded by `auditSem`.

## 5. Data-Flow Trace

### GitHub adapter path
```
GitHub webhook POST → /v1/incoming-webhooks/:id/:token/github
  → handlePush(c, githubAdapter)
    → auth (token + group) → rate-limit → body read (8KB cap)
    → githubAdapter.parse(X-GitHub-Event header, body)
      → parseGitHubPush: switch on event type → render* → markdown string
      → clipRunes(content, maxContentRunes()) → *pushPayloadReq{Content: clipped}
    → req.MsgType == "" → text path
    → req.Content non-empty (clipRunes guarantees this for non-empty input)
    → utf8.RuneCountInString ≤ maxContentRunes (clipRunes guarantees)
    → buildPayload(m, req) → {type:Text, content:req.Content, from:{kind:webhook,...}, space_id:m.SpaceID}
    → SendMessageWithResult → audit success (adapter=github, bumpUsed=true)
    → successBody(githubAdapter, msgID, "") → {status:0, message_id:N}
```

### GitHub skip path (ping / unsupported event)
```
  → parseGitHubPush returns (nil, "ping"|"event", "")
  → submitSkipped(m, bodyLen, ip, "github", "ping"|"event")
    → auditModel{Status:3, Reason:"ping"|"event", HTTPStatus:200, Adapter:"github"}, bumpUsed=false
  → successBody(githubAdapter, 0, "ping"|"event") → {status:0, message_id:0, skipped:"ping"|"event"}
```

### WeCom adapter path
```
WeCom bot POST → /v1/incoming-webhooks/:id/:token/wecom
  → handlePush(c, wecomAdapter)
    → auth → rate-limit → body read
    → wecomAdapter.parse(nil headers, body)
      → parseWeComPush: switch on msgtype → extract/render content
      → NO clipping (caller controls content, 413 downstream if too long)
      → *pushPayloadReq{Content: content}
    → text path → content check → maxContentRunes check (413 if exceeded)
    → buildPayload → SendMessageWithResult → audit success (adapter=wecom)
    → successBody(wecomAdapter, msgID, "") → {status:0, message_id:N, errcode:0, errmsg:"ok"}
```

### Native adapter path (backward compat)
```
  → parseNativePush: json.Unmarshal → *pushPayloadReq (same as pre-refactor)
  → All downstream paths unchanged from base branch
```

---

[Octo-Q] verdict: **APPROVE**

All findings are P2/nit. The adapter pipeline correctly reuses the existing auth/rate-limit/audit infrastructure, data-flow from adapter parse through buildPayload to message delivery is type-safe and semantically equivalent to the pre-refactor native path, and the new auditSkipped status is properly isolated from success/failure semantics. Test coverage is comprehensive (unit + e2e + legacy guard). No P0/P1 issues found.

Jerry-Xin (changes requested):
Relevant PR, but one blocking correctness issue prevents the GitHub adapter from reliably handling real webhook traffic.

**🔴 Blocking**
- 🔴 Critical: GitHub payloads are still rejected by the native 8 KiB body cap before the adapter can render and clip them. `defaultMaxBytes` is 8 KiB in `modules/incomingwebhook/api.go:45`, and `handlePush` rejects any larger request at `modules/incomingwebhook/api.go:919-929` before `parseGitHubPush` runs. The GitHub renderer only clips the generated message after parsing at `modules/incomingwebhook/adapter_github.go:141-142`, so valid GitHub events with larger platform-generated JSON still get `413`. A normal push / PR / issue payload can exceed 8 KiB, and the sender cannot shorten it. Please add an adapter-specific bounded body limit or another safe path for GitHub payloads, plus a test where a >8 KiB GitHub payload renders to a valid clipped message.

**💬 Non-blocking**
- 🟡 Warning: `TestPush_GitHubPushEvent_Delivers` and `TestPush_WeComText_Delivers` only assert “not 400/401”, so a `413` regression like the body-cap issue is not covered for GitHub. Add a focused e2e or handler-level case for GitHub body size behavior.
- 🔵 Suggestion: `api_i18n_test.go` still checks only `blocks`, `msg_type`, `content`, and `json` for surfaced `details.reason`; adding `event` would lock the new GitHub error contract.

**✅ Highlights**
- The adapter abstraction keeps auth, group validation, delivery, and audit behavior centralized.
- Skipped GitHub events are modeled distinctly with `status=3`, which is clearer than treating them as success or failure.
- Pure adapter tests are broad and passed locally; infra-backed tests could not run here because local MySQL was unavailable.

yujiawei (approved):
# Code Review — PR #330 (octo-server)

**Scope:** Phase 3 platform adapters for incoming webhooks — GitHub event renderer and WeCom group-bot format translator, layered on the existing native push pipeline. +1233/−47 across 14 files. Classified security-sensitive.

**Verdict: APPROVED.** No blocking issues. Build, `go vet`, and the dependency-free unit tests pass locally; the four hard-security invariants of this module are preserved.

## 1. Verified safe (security-critical)

- ✅ **Anti-enumeration / auth ordering holds.** All three routes (`native` / `github` / `wecom`) share one middleware chain and one `handlePush` pipeline (`api.go:202-209`, `841`). Adapter parsing (`ad.parse`) and every new audit write run **after** the constant-time token compare (`api.go:882`) and the group-Normal gate (`api.go:891-901`). The new `submitSkipped` path (`api.go:944`, `1136-1147`) is post-auth, same as `submitFailure`/`submitSuccess`. Pre-auth scanners still get the uniform 401 via `failAuth` and only burn the per-IP failure budget. The missing-`X-GitHub-Event` 400 (`adapter_github.go`) is likewise post-auth, so it is not an enumeration side channel.
- ✅ **Identity / Space not forgeable.** The adapters parse platform bodies into whitelist structs (`gh*`, `wecom*`) and emit only `pushPayloadReq{Content: ...}`. `Username`, `AvatarURL`, `Extra`, `space_id`, `from.*`, and message `type` cannot be injected via the adapter routes — display identity is fixed to the webhook config (strictly more locked-down than the native path). All rendered content flows through `buildPayload` (Text type) with server-injected `space_id` (`api.go:1061-1075`).
- ✅ **No SSRF.** The server never dereferences any event URL (commit URL, `html_url`, article URL, `card_action.url`, `picurl`). They are embedded only as markdown link text rendered client-side. No `http.Get`/client/`image.Decode` exists in the module.
- ✅ **DoS / write-amplification bounded.** Skipped-delivery audit writes sit behind the same per-webhook 5rps limiter (`api.go:904`) and the `auditSem` hard cap (overflow dropped with a Warn, `api.go:1153-1165`) as all other audit writes — no extra DB-connection or goroutine pressure.
- ✅ **Migration is safe.** `20260610000001` is COMMENT-only `MODIFY COLUMN` on reused columns (no structural change) — INSTANT/lock-free on MySQL 8.0. New enum values fit existing widths (`adapter VARCHAR(16)`, `reason VARCHAR(32)`). Down migration restores prior comments.
- ✅ **Backward compatible.** Native response shape unchanged; `createResp.url` retained alongside the new `urls` object (`model.go:131-140`). Token still surfaced only on create/regenerate.

## 2. Findings

No P0/P1. Three non-blocking nits below.

### nit-1 — `push` event with empty `commits` (not created/deleted) renders "pushed 0 commit(s)"
`adapter_github.go` `renderGitHubPush`: a push with `commits:[]`, `created=false`, `deleted=false`, non-tag ref falls through to the general branch and produces a non-empty `**who** pushed 0 commit(s) to \`ref\`` message, which is then delivered (not skipped, since skip only triggers on empty content). GitHub can emit such events for some ref updates. Cosmetic only — no auth/audit/pipeline impact. Optional: add `case len(ev.Commits) == 0: return "", nil` before the general branch so it takes the empty-content skip path, consistent with the "skip the noise" intent.

### nit-2 — Markdown link text / URLs not escaped for `]` `)`
Event titles (`adapter_github.go` PR/issue/comment/release lines; `adapter_wecom.go` news title) are passed through `oneLine` (newline stripping only) before insertion into `[%s](%s)`. A title containing `]` will visually break the rendered link **client-side**. Not a server-side injection/parsing issue: content is a plain string serialized via `util.ToJson`, `from.*` is server-clamped, `Extra` is dropped, `space_id` is server-derived. Acceptable as-is; escaping would only improve cosmetic fidelity. (Note: the commit-line `firstLine(cm.Message)` text sits *after* the link, so it is not itself a link-breaking vector.)

### nit-3 — Skipped-delivery rows add to the known un-pruned audit table
`submitSkipped` introduces a `status=3` row class for ping / unsupported-event / non-subset-action traffic. The audit table still has no TTL/retention job (pre-existing debt). This is **not** a new amplification ceiling — Phase 2 already writes `status=2` failure rows under the identical 5rps + `auditSem` cap, so `status=3` is one more bounded row class, not a categorical change. Flagging only because the long-promised retention job remains outstanding and a firehose-subscribed webhook will steadily accrue skipped rows; worth tracking the TTL job, but not a blocker for this PR.

## 3. Recommendations

- Land as-is. nit-1 (the `len(commits)==0` skip) is the only one worth a quick follow-up commit if convenient; nit-2 and nit-3 are optional/tracking items.
- For human attention: the audit-table TTL/retention job has been outstanding across PR #31 / #250 / #299 and now #330 — recommend prioritizing it independently of this feature.

## 4. Additional notes

- Test coverage is solid: pure translation unit tests (`adapter_*_test.go`) plus end-to-end push tests (`adapter_push_test.go`) covering auth enforcement on adapter routes, ping skip + audit, unsupported-event skip, missing-header 400, WeCom text/media rejection, and content clipping. The `TestIncomingWebhookNoLegacyResponseError` guard was correctly extended to the new adapter files.
- The `skip` vs `invalid` disambiguation for the shared `event` reason is correctly resolved by the status field (2=failed/400 vs 3=skipped/200) and documented in the README and migration comment.

yujiawei (changes requested):
# Code Review — PR #330 (octo-server) — Corrected verdict

**This supersedes my earlier approval on this PR.** On re-examination (and agreeing with the other blocking review on this PR), there is a P1 functional defect in the GitHub adapter that must be fixed before merge. My earlier APPROVED verdict missed the body-cap ordering; I am correcting it to CHANGES_REQUESTED.

## P1 (blocking) — GitHub payloads are rejected by the 8 KiB body cap before the adapter can render/clip them

`handlePush` enforces the uniform body cap (`maxBytes()`, default `defaultMaxBytes = 8 * 1024` at `api.go:45`) at `api.go:920-931`, **before** `ad.parse` runs at `api.go:935`. The GitHub renderer's length control (`clipRunes` to `maxContentRunes`) only operates on the **already-generated message content** (`adapter_github.go` `parseGitHubPush`, the `clipRunes(content, maxContentRunes())` call), i.e. after parsing — which never happens for an oversized raw body.

Consequence: a valid GitHub event whose **raw JSON** exceeds 8 KiB returns `413` and is never delivered. Realistic GitHub push payloads (each commit object carries author/committer, full message, and `added`/`removed`/`modified` file arrays; GitHub sends up to 20 commits) frequently exceed 8 KiB; PR/issue events with long bodies do too. The sender is GitHub — it cannot shorten the payload.

This directly contradicts the PR's own stated contract:
- README: "事件里的超长字段（标题/提交信息/评论）服务端截断，GitHub 流量不会触发 413。"
- `adapter.go` `clipRunes` comment: "平台调用方没有任何手段「修短」一个 GitHub 事件".

The clip protects against `maxContentRunes` (the *rendered* message) but not against the raw-body 413, which fires first.

### Suggested fix

Give the adapter path a separate, higher bounded read limit for the **raw** body (the rendered output is already clamped by `maxContentRunes`, so a larger raw-read ceiling does not let oversized content reach clients). For example, parameterize the read limit per adapter (e.g. a `pushAdapter.maxBodyBytes` with a GitHub-appropriate default, still bounded to prevent unbounded reads), or read GitHub bodies under a dedicated `DM_INCOMINGWEBHOOK_GITHUB_MAX_BYTES`. Keep native/WeCom on the existing 8 KiB cap (those callers control their body and a 413 is actionable).

Add a regression test: a >8 KiB GitHub payload must render to a valid clipped message (200), not 413.

## P2 / non-blocking

- **Test gap (echoes the bug above):** `TestPush_GitHubPushEvent_Delivers` / `TestPush_WeComText_Delivers` only assert "not 400/401", so a 413 regression on GitHub size handling is uncovered. Add a focused case asserting large-GitHub-body behavior.
- **i18n contract:** `api_i18n_test.go` checks surfaced `details.reason` for `blocks`/`msg_type`/`content`/`json` but not the new `event`. Adding `event` would lock the new GitHub error contract.

## Still verified safe (unchanged from my prior review)

These hold and are not affected by the above:
- ✅ Anti-enumeration / auth ordering: `ad.parse` and all audit writes (`submitFailure`/`submitSuccess`/`submitSkipped`) run after the constant-time token compare (`api.go:882`) and group-Normal gate (`api.go:891-901`); pre-auth scanners get a uniform 401 and only burn the IP failure budget.
- ✅ Identity/Space not forgeable: adapters emit only `pushPayloadReq{Content: ...}`; `Username`/`AvatarURL`/`Extra`/`space_id`/`from.*`/`type` cannot be injected via adapter routes. `buildPayload` injects server-derived `space_id` (`api.go:1061-1075`).
- ✅ No SSRF: the server never dereferences any event URL (no `http.Get`/client/`image.Decode` in the module); URLs are embedded as client-rendered markdown only.
- ✅ DoS bounded: skipped/failure audit writes share the per-webhook 5rps limiter and `auditSem` cap.
- ✅ Migration `20260610000001` is COMMENT-only `MODIFY COLUMN` on reused columns (INSTANT/lock-free on MySQL 8.0); new enum values fit existing widths. Backward compat: native response unchanged, `createResp.url` retained alongside new `urls`.

## Minor (nits, optional)

- `renderGitHubPush`: a `push` with `commits:[]`, `created=false`, `deleted=false`, non-tag ref renders "pushed 0 commit(s)" and delivers instead of skipping. Cosmetic; optional `case len(ev.Commits) == 0: return "", nil` before the general branch.
- Markdown link text/URLs not escaped for `]`/`)` (titles via `oneLine`); client-render cosmetic only — no server-side injection (content is a plain serialized string, `from.*` clamped, `Extra` dropped, `space_id` server-derived).
- Pre-existing: the audit table still has no TTL/retention job (outstanding across PR #31/#250/#299); the new `status=3` rows add to it (bounded by the same 5rps + `auditSem` cap, so no new amplification ceiling). Worth prioritizing the retention job independently.

an9xyz:
@Jerry-Xin review follow-ups landed in 7442646 — thanks for the catch, the blocking finding was real:

- **🔴 body cap**: added `pushAdapter.bodyLimit` — native/wecom keep the caller-authored 8KiB cap; the github route now uses a platform cap (default **1MiB**, `DM_INCOMINGWEBHOOK_GITHUB_MAX_BYTES`), since real push/PR event JSON routinely exceeds 8KiB and GitHub can't shorten it. The body read happens post-auth + post-per-webhook-rate-limit, so the wider cap is not an amplification surface.
- **🟡 test gap**: new e2e `TestPush_GitHubLargePayload_Not413` pushes a >20KiB push event through `/github` (must not 413) and the same payload through native (must still 413), plus a unit guard `TestGitHubMaxBytes_ExceedsNativeCap`.
- **🔵 reason contract**: `event` added to `TestPushPayloadInvalidSurfacesReason`.

Also applied the nit both reviewers flagged: degenerate ref updates (no commits, not create/delete) now take the skip path instead of rendering "pushed 0 commit(s)". Full module suite re-ran green locally against real MySQL/Redis (`-race -shuffle=on`).

https://claude.ai/code/session_01W4UKE63aRUDm1i2JGN6xSf

---

Jerry-Xin (approved):
Summary: This PR is relevant to `octo-server` and the incoming webhook module; the shared push pipeline refactor looks sound, with no blocking issues found.

🔴 Blocking: None.

💬 Non-blocking:
- 🟡 Warning: GitHub-controlled fields are interpolated into markdown without escaping, e.g. commit messages, PR/issue titles, comment snippets, and release names in `modules/incomingwebhook/adapter_github.go:203`, `:228`, `:244`, `:258`, `:280`. For public repos, issue titles/comments may come from untrusted users and can manipulate rendered links/formatting in internal chats. Consider escaping markdown link text and validating generated link URLs.
- 🔵 Suggestion: `TestGitHubMaxBytes_ExceedsNativeCap` depends on ambient env values at `modules/incomingwebhook/adapter_github_test.go:109`. If CI or local env sets `DM_INCOMINGWEBHOOK_GITHUB_MAX_BYTES` low or raises `DM_INCOMINGWEBHOOK_MAX_BYTES`, this can fail despite correct code. Clear/set both env vars in the test.
- 🔵 Suggestion: The 413 errcode comment still names only `DM_INCOMINGWEBHOOK_MAX_BYTES` at `pkg/errcode/incomingwebhook.go:54`; GitHub now has `DM_INCOMINGWEBHOOK_GITHUB_MAX_BYTES`, so the comment is slightly stale.

✅ Highlights:
- The adapter routes reuse the existing auth, group-state, rate-limit, payload construction, and audit invariants in `modules/incomingwebhook/api.go:841`.
- Post-auth failure/skipped auditing now records the real adapter, preserving anti-enumeration behavior while improving observability.
- Test coverage covers parser behavior, route auth, skip auditing, large GitHub bodies, WeCom compatibility response fields, and legacy error-response guards.

Verification: `go test ./modules/incomingwebhook -run 'Test(Parse|SuccessBody|ClipRunes|FirstLine|GitHubMaxBytes)'` passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#330 Review Report
**PR:** Mininglamp-OSS/octo-server #330
**Title:** feat(incomingwebhook): GitHub / WeCom platform adapters on push (Phase 3)
**Head SHA:** 74426469ee41b63bb54e572c88947794b5aebf7c
**Scope:** 14 files, +1329/-50

---

## 1. Investigation Summary

This PR adds a platform adapter layer to the incoming webhook push pipeline, enabling GitHub webhook events and WeCom group-bot message formats to be translated into native push requests. Key architectural decisions:

- **Adapter pattern**: `pushAdapter` struct with `parse`/`successExtra`/`bodyLimit` fields; three instances (native/github/wecom) share the same auth/rate-limit/group-check/delivery/audit pipeline via `handlePush(c, ad)`.
- **GitHub body limit**: Separate 1MiB cap (`githubMaxBytes`) for GitHub events (platform-generated, sender can't shorten), vs 8KiB for native/wecom (caller-authored). Read happens AFTER token auth + per-webhook rate limit — not an amplification vector.
- **Skip semantics**: New `auditSkipped(3)` status for "received but intentionally not delivered" (GitHub ping, unsupported events). Returns 200 to keep GitHub green, audited for admin visibility.
- **SQL migration**: Comment-only ALTER TABLE (INSTANT on MySQL 8), no structural change.

## 2. Verification

| Item | Status | Evidence |
|------|--------|----------|
| Adapter pipeline shares auth/rate-limit/group-check | ✅ | `api.go:197-210` — `chain` closure wraps all 3 routes with identical middleware |
| GitHub body limit gated behind auth+rate-limit | ✅ | `api.go:930-940` — `ad.bodyLimit()` called at step 4, after steps 1-3 (auth/group/rate-limit) |
| `from.kind=webhook` + server-derived `space_id` injection unchanged | ✅ | `api.go:1063-1082` — `buildPayload` unchanged, adapters only set `Content` |
| Anti-enumeration invariant preserved | ✅ | Auth failures still go through `failAuth` → IP failure budget, never audited |
| `submitFailure` signature updated at all call sites | ✅ | All 8 call sites in `handlePush` pass `ad.name` as adapter param |
| `submitSkipped` correctly uses `bumpUsed=false` | ✅ | `api.go:1145-1157` — `call_count` semantics preserved |
| `successBody` no key collision | ✅ | Native: {status, message_id}; WeCom adds {errcode, errmsg}; skip adds {skipped} — no overlap |
| `clipRunes` correct for edge cases | ✅ | `adapter.go:93-101` — output is exactly `max` runes; `max<=0` returns empty |
| Adapter parse: nil req always paired with non-empty skip or invalid | ✅ | All 3 parse functions verified — no path returns (nil, "", "") |
| Deliveries endpoint handles status=3 | ✅ | `api.go:749-755` — `deliveryRespFrom` passes status through without filtering |
| SQL migration is comment-only, INSTANT | ✅ | `sql/20260610000001_...sql` — MODIFY COLUMN for COMMENT only |
| Route registration: no path conflict | ✅ | Gin radix tree: `:token` vs `:token/github` vs `:token/wecom` are distinct segments |
| `createResp.URLs` additive, legacy `url` unchanged | ✅ | `model.go:136-140` + `api.go:511,689` — both `URL` and `URLs` populated |
| Test coverage: unit + e2e | ✅ | 6 test files covering all event types, skip paths, rejection paths, auth, large payloads |

## 3. Findings

### P0/P1: None

### P2 (non-blocking)

**P2-1: WeCom news rendering may exceed maxContentRunes with many articles**
- **Diff-scope:** new (introduced by this PR's WeCom adapter)
- `renderWeComNews` joins articles without total-length clipping. Each article can contribute ~200+ runes (title + description + formatting). With 20+ articles, the combined content exceeds `maxContentRunes()` (4000) and triggers 413.
- This is **documented and intentional** — WeCom body is caller-authored, and 413 is explicit. But callers migrating from WeCom bots may not anticipate the limit.
- **Suggestion:** Consider adding a note in the README about recommended max article count for news messages, or pre-clip the rendered output like the GitHub adapter does.

**P2-2: Redundant content checks after adapter parse (defensive, acceptable)**
- **Diff-scope:** new (adapter path reuses existing checks)
- After adapter parse, `handlePush` re-checks content emptiness and `maxContentRunes`. GitHub adapter already clips; WeCom adapter doesn't clip but the downstream check catches it.
- This is good defensive programming — protects against future adapters that don't clip. No action needed.

## 4. Security Review

- ✅ GitHub 1MiB body limit is behind token auth + per-webhook 5rps rate limit. Max memory exposure: 5MiB/s/webhook (authenticated).
- ✅ All 3 adapter routes share the **same rate limiter instances** — adapters are not separate quotas.
- ✅ `successExtra` map is package-level read-only after init — no race condition.
- ✅ Adapter structs use whitelist parsing — extra JSON fields silently ignored.
- ✅ `req.Extra` discarded by `buildPayload` — no field injection from adapter output.
- ✅ `resolveFromIdentity` falls back to webhook config for adapters (Username/AvatarURL always empty).
- ✅ Token appears in `urls` response only at create/regenerate (same as legacy `url`).

## 5. Data Flow Trace

### GitHub adapter
```
X-GitHub-Event header → event type → renderer selection
  → JSON body → gh*Event struct (whitelist parse)
  → render to markdown string
  → clipRunes(content, maxContentRunes()) [clips to 4000 runes]
  → pushPayloadReq{Content: clipped}
  → handlePush text path (MsgType="" → text)
  → buildPayload: type=Text, content=req.Content, from.kind=webhook, space_id=server-derived
  → SendMessageWithResult
```
**Verified:** data flows end-to-end; no silent drops; clipping prevents 413 for platform-generated content.

### WeCom adapter
```
JSON body → wecomMsg struct (whitelist parse)
  → switch msgtype → extract/render content
  → pushPayloadReq{Content: content} [no clipping — caller-authored]
  → handlePush text path (MsgType="" → text)
  → maxContentRunes check (413 if exceeded)
  → buildPayload → SendMessageWithResult
```
**Verified:** explicit 400 for unsupported msgtypes (image/file/voice); explicit 413 for oversized content; caller has ability to shorten.

### Native adapter
```
JSON body → pushPayloadReq (direct unmarshal)
  → handlePush (unchanged pipeline)
```
**Verified:** backward-compatible; no behavior change from pre-PR code.

### Audit flow (all adapters)
```
parse result → {success: submitSuccess(ad.name, bumpUsed=true)
               | failure: submitFailure(ad.name, reason, httpStatus)
               | skipped: submitSkipped(ad.name, reason)}
  → submitDelivery → async goroutine → recordDelivery (markUsed + insertAudit)
  → deliveries endpoint: queryRecentAudits → deliveryRespFrom (pass-through status/adapter/reason)
```
**Verified:** adapter name correctly propagated to audit table; status=3 visible in deliveries.

## 6. Suggestions

1. Consider documenting a recommended max article count for WeCom news messages (~15-20 articles to stay under 4000 runes).
2. Future: when adding GitLab/Feishu adapters, ensure the adapter parse function clips content if platform-generated (like GitHub does), or document that callers must control length (like WeCom).

---

[Octo-Q] verdict: **APPROVE** — No P0/P1 findings. Clean adapter abstraction with well-preserved security invariants, thorough test coverage, and correct data flow across all three adapter paths. P2 observations are non-blocking.

yujiawei (approved):
# Code Review — PR #330 (octo-server)

**Scope:** Phase 3 platform adapters for incoming webhooks — a GitHub event renderer and a WeCom group-bot format translator, layered onto the existing native push pipeline. Classified security-sensitive; reviewed at HEAD `74426469`.

This review supersedes my earlier `CHANGES_REQUESTED` on this PR, which was made against the previous commit (`5b1b16f7`). The single blocking issue from that round has been fixed in `74426469` ("per-shape body cap so real GitHub events aren't 413'd"). Re-reviewing the updated head: no blocking findings remain.

## Verdict summary

| Dimension | Result |
|---|---|
| Auth / anti-enumeration ordering | ✅ Clean |
| Payload forgery / injection / SSRF | ✅ Clean |
| DoS / resource exhaustion | ✅ Clean (one optional hardening nit) |
| Correctness / migration / backward-compat | ✅ Clean (one cosmetic nit) |

## Previously-blocking issue — now resolved ✅

**GitHub payloads rejected by the 8 KiB body cap before render/clip.** The earlier head applied the uniform 8 KiB `maxBytes()` cap before the adapter could parse, so a realistic GitHub push/PR event (which the sender cannot shorten) would `413`. The fix introduces a per-shape `bodyLimit` on `pushAdapter`: native/wecom keep the 8 KiB caller-authored cap, GitHub gets a dedicated `githubMaxBytes()` (default 1 MiB, env-tunable `DM_INCOMINGWEBHOOK_GITHUB_MAX_BYTES`), consumed via `ad.bodyLimit()` at `api.go:923`. Critically, the wider read happens **after** token auth (`api.go:882`) + group-Normal gate (`api.go:891`) + the per-webhook 5 rps limiter (`api.go:904`), so it is not an unauthenticated amplification surface. A regression test (`TestPush_GitHubLargePayload_Not413`) asserts a >8 KiB GitHub payload is accepted on `/github` and still `413`s on the native route. The follow-up nit from that round (no-commit ref update rendering "pushed 0 commit(s)") is also fixed — `adapter_github.go:183-188` now returns empty → skip, with a matching test.

## Verified safe

- **Auth & anti-enumeration ordering.** The `/github` and `/wecom` routes reuse the exact middleware chain (`requirePushEnabled → localFloor → ipLimit → ipFailureGate`) via the shared `chain(...)` helper (`api.go:200-210`), and dispatch into the same `handlePush`. Token constant-time compare (`subtle.ConstantTimeCompare`, `api.go:882`) and the group-Normal gate run **before** body read/parse. Unknown/deleted webhook and bad token return a uniform 401 and never write an audit row (`failAuth` only touches the IP failure budget), preserving the pre-auth anti-enumeration invariant. The new `submitSkipped` / adapter-tagged `submitFailure` writes all sit after auth.
- **No payload forgery.** Adapters emit only `pushPayloadReq{Content: ...}` (`adapter_github.go:163`, `adapter_wecom.go:100`); they never set `Username`/`AvatarURL`/`Extra`/`Blocks`/`MsgType`. `buildPayload` drops caller `Extra`, server-derives `space_id` from the group row (`api.go:1077`), and injects `from.kind=webhook`/`webhook_id`. Adapter content goes through the default text path only — `space_id`, identity, and visibility/mention semantics are not forgeable via the adapter routes.
- **No SSRF.** The module performs no outbound dereference of any event-supplied URL (no `http.Get`/client/image decode anywhere in `modules/incomingwebhook`); `net/http` is used only for status constants and header reads. Commit/PR/repo URLs are embedded as client-rendered markdown text, never fetched server-side.
- **DoS bounded.** The new `auditSkipped` writes share the same `auditSem` cap and the per-webhook 5 rps gate as success/failure, adding no new amplification ceiling. JSON parsing uses whitelist structs that ignore unknown fields; rendered commits are capped at `maxRenderedCommits=5`; no slice is preallocated from an attacker-declared count. The wider 1 MiB read applies only to the GitHub adapter and only post-auth.
- **Migration safety.** `20260610000001` is a COMMENT-only `MODIFY COLUMN` on reused columns (INSTANT/lock-free on MySQL 8) with a matching Down. New enum values fit existing widths (`status=3` in SMALLINT; reasons `event`/`ping` and adapters `github`/`wecom` within `VARCHAR(32)`/`VARCHAR(16)`). Backward compat holds: native response unchanged (`successBody` adds nothing for native — no `successExtra`), and `createResp` retains `url` alongside the new `urls`.

## Non-blocking nits (optional — do not block merge)

1. **`githubMaxBytes()` has no upper ceiling on the operator-set env value** (`adapter_github.go:42-49`). Parsing is panic-safe and `n>0` rejects negative/garbage (falls back to 1 MiB), so this is **not** an attacker surface — the env is operator-set and the read is auth+rate-limit gated. But a fat-fingered huge positive value (e.g. `2000000000`) would let each authenticated push buffer that many bytes via `io.ReadAll`. Optional: clamp the parsed value to a sane max, mirroring how `deliveries` clamps to `maxDeliveriesLimit` (`api.go:733-735`).

2. **Reason code `event` is overloaded across a 400 and a 200 path** (`adapter_github.go:131` vs `153`/`160`). A missing `X-GitHub-Event` header → `invalid="event"` → `auditFailed`/400, while an unrendered event/action → `skip="event"` → `auditSkipped`/200. The two stay fully distinguishable by `status` (2 vs 3) and `http_status` (400 vs 200), so this is purely cosmetic — a `reason`-only grep can't separate them. Optional: use a distinct reason (e.g. `no_event`) for the missing-header 400.

## Local checks

`go build`, `go vet`, and the infra-free unit suite (`TestParseGitHub*`, `TestParseWeCom*`, `TestGitHubMaxBytes_ExceedsNativeCap`, `TestPushPayloadInvalidSurfacesReason`, `TestIncomingWebhookNoLegacyResponseError`) pass locally. The infra-backed e2e cases (`adapter_push_test.go`) require MySQL+Redis+WuKongIM and were not run in this environment; the PR reports them green against real MySQL 8 + Redis. The legacy-response guard test correctly lists the new `adapter*.go` files.

**Recommendation: Approve.** The blocking body-cap issue is resolved, the security-sensitive surfaces (auth ordering, forgery, SSRF, DoS, migration) are clean, and only two optional cosmetic nits remain.

yujiawei (approved):
# Code Review — PR #330 (octo-server)

**Scope:** `feat(incomingwebhook): GitHub / WeCom platform adapters on push (Phase 3)` — `+1329/-50`, 14 files. Head `c9fd5893`, base `main`. Security-sensitive (auth/middleware/token surface).

This is a re-review at the new head. The previous review at the earlier head requested changes for a single blocking issue (the GitHub adapter could not receive realistic event payloads because the uniform body cap fired before the adapter parsed/clipped). The top commit `c9fd5893 fix(incomingwebhook): per-shape body cap so real GitHub events aren't 413'd` addresses exactly that. **Verdict: APPROVED.**

## 1. Verification of the blocking fix

The prior blocker: `handlePush` enforced the native 8KiB body cap (`maxBytes()`) before `ad.parse`, while the GitHub renderer's `clipRunes(...)` only bounds the *rendered* output, never the *raw* body. A valid GitHub push/PR event whose raw JSON exceeds 8KiB returned 413, and the sender (GitHub) cannot shorten it — directly contradicting the PR's own contract.

The fix is correct and minimal:

- ✅ `pushAdapter` gains a `bodyLimit func() int` field; `native`/`wecom` keep `maxBytes` (8KiB — these bodies are caller-authored and *should* be constrained), `github` gets `githubMaxBytes` (default 1MiB, tunable via `DM_INCOMINGWEBHOOK_GITHUB_MAX_BYTES`). `adapter.go:36-52`.
- ✅ `handlePush` now reads `limit := ad.bodyLimit()` instead of the hardcoded `maxBytes()`. `api.go:923`.
- ✅ **Amplification surface checked:** the body read (`api.go:919-934`) still runs *after* constant-time token compare (`api.go:882`), the group-active gate (`api.go:891-901`), and the per-webhook 5rps limiter (`api.go:904`). So the wider 1MiB cap is only reachable by an authenticated, rate-limited caller — not an unauthenticated amplification vector.
- ✅ **Client-side output still bounded:** `parseGitHubPush` returns `pushPayloadReq{Content: clipRunes(content, maxContentRunes())}` (`adapter_github.go`), so widening the raw read cap does not let oversized content reach clients — the rendered message is still clamped, and individual fields (titles/commit messages/comment snippets) are clipped at 120/200/300 runes before assembly.
- ✅ **Unit guard:** `TestGitHubMaxBytes_ExceedsNativeCap` pins `githubMaxBytes() > maxBytes()`.

This is the precise fix recommended in the prior review.

## 2. Secondary fixes (both nits from prior review)

- ✅ Degenerate ref updates (no commits, not create/delete — e.g. force-push to identical content) now return `""` → skip path, instead of rendering a noisy "pushed 0 commit(s)". `adapter_github.go renderGitHubPush` + `TestParseGitHubPush_NoCommitRefUpdateSkipped`.
- ✅ The new `reason=event` error contract is locked by tests (`TestParseGitHubPush_HeaderGate`), and the `errcode` doc + migration COMMENT were updated to document `event`/`ping`.

## 3. Security re-confirmation (unchanged from prior review, still holds)

- ✅ **Anti-enumeration ordering preserved.** All three routes (`native`/`github`/`wecom`) share the identical middleware chain (`requirePushEnabled → localFloor → ipLimit → ipFailureGate → handler`, `api.go:202-209`) and the same `handlePush` pipeline. Every `submitFailure`/`submitSkipped`/parse branch sits after token auth + group gate; pre-auth scanners get a uniform 401 and only burn the per-IP failure budget.
- ✅ **Identity/space not forgeable.** Adapters emit only `pushPayloadReq{Content}` (and never `Username`/`AvatarURL`/`Extra`/`space_id`/`from.*`); the adapter path is strictly more locked-down than native. `buildPayload` injects the server-derived `space_id` and `from.kind=webhook`.
- ✅ **No SSRF.** No `http.Get`/client/`image.Decode` anywhere in the adapters; event URLs are rendered as client-side markdown only, never dereferenced server-side.
- ✅ **DoS bounded.** Skip/failure audit writes share the per-webhook 5rps limiter and the `auditSem` cap; `rate_limited` (429) is deliberately not audited.
- ✅ **Migration is safe.** `20260610000001` is COMMENT-only `MODIFY COLUMN` (INSTANT on MySQL 8, no lock); new enum values (`status=3`, `reason=event/ping`, `adapter=github/wecom`) fit existing column widths.
- ✅ **Backward compatible.** Native response shape unchanged; `createResp.url` retained alongside the new `urls{native,github,wecom}` map (token-gated, only emitted on create/regenerate).

## 4. WeCom adapter (reviewed, no issues)

`parseWeComPush` whitelist-parses text/markdown/markdown_v2/news/template_card and rejects material types (image/file/voice) with `400 reason=msg_type` — explicit failure is the honest migration experience vs. silently dropping. Content is intentionally *not* widened (caller-authored, can be shortened, so the 8KiB cap is appropriate).

## 5. Build / tests

- ✅ `go vet ./modules/incomingwebhook/` clean; `gofmt -l` clean on all changed files.
- ✅ Pure adapter unit tests pass (`TestParseGitHubPush_*`, `TestParseWeComPush_*`, `TestGitHubMaxBytes_ExceedsNativeCap`, `TestParseGitHubPush_ContentClipped`, `TestClipRunes`, `TestSuccessBody`). DB-backed integration tests were not run here due to a local test-DB migration mismatch unrelated to this PR — CI is the source of truth for those.

## 6. Non-blocking suggestions (do not block merge)

- **P2 / human-verify:** the GitHub adapter authenticates only via the URL token; `X-Hub-Signature-256` HMAC verification is explicitly deferred (noted in `adapter_github.go` header). This is an accepted decision for this phase, but worth a human's confirmation that URL-token-only auth is acceptable for GitHub events given the token travels in the path. The 1MiB cap on an authenticated, rate-limited path is a reasonable bound; a deployment that fronts this with a CDN/proxy should ensure the raw body limit there is ≥ `DM_INCOMINGWEBHOOK_GITHUB_MAX_BYTES`.
- **nit:** rendered markdown link text (titles/repo names) is not escaped for literal `]`/`)`; cosmetic-only on the client.
- **Standing module debt (pre-existing, not this PR):** the `incoming_webhook_audit` table still has no TTL/retention job; the new `status=3` skipped rows add to unbounded growth (bounded per-webhook by the 5rps limiter). Flagged across prior phases for eventual human attention.

No P0/P1 remain. The blocking issue from the prior review is resolved and verified.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#330 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/330
**Title:** feat(incomingwebhook): GitHub / WeCom platform adapters on push (Phase 3)
**Head SHA:** c9fd5893b911cd049af49246bb4a4c0c31742c74
**Files:** 14 changed (+1329 / −50)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Adapter pipeline refactor (push → handlePush + pushAdapter) | ✅ | `api.go:832-840` — three thin entry points delegate to `handlePush(c, adapter)`; middleware chain factored into `chain()` closure, same instances shared |
| GitHub adapter: event routing & rendering | ✅ | `adapter_github.go:148-186` — switch on `X-GitHub-Event`; ping→skip, unknown→skip, missing header→400 `reason=event`; renderers cover push/PR/issues/comment/release |
| GitHub adapter: content clipping prevents 413 | ✅ | `adapter_github.go:183` — `clipRunes(content, maxContentRunes())` applied before returning; titles clipped to 200 runes, comments to 300, commit messages to 120 |
| GitHub adapter: per-shape body cap (commit 2) | ✅ | `adapter_github.go:44-53` — `githubMaxBytes()` defaults 1MiB, env-overridable; `adapter.go:47` — `bodyLimit` is a per-adapter func; `api.go:919` — `limit := ad.bodyLimit()` replaces hardcoded `maxBytes()` |
| WeCom adapter: msgtype routing | ✅ | `adapter_wecom.go:82-107` — text/markdown/markdown_v2 pass through; news/template_card degrade; default (image/file/voice/empty) → 400 `reason=msg_type` |
| WeCom adapter: content NOT clipped (caller-authored) | ✅ | `adapter_wecom.go:78-79` — documented asymmetry vs GitHub; downstream `handlePush` text-path enforces `maxContentRunes()` → 413 if exceeded |
| Audit: adapter threaded through submitFailure | ✅ | `api.go:1115` — signature adds `adapter string`; all 8 call sites in `handlePush` pass `ad.name`; no stale `adapterNative` hardcode remains |
| Audit: new auditSkipped(3) status | ✅ | `model.go:43-46` — new constant; `api.go:1131-1145` — `submitSkipped` writes status=3 with `bumpUsed=false`; response is 200 with `skipped` field |
| Onboarding: `urls` on create/regenerate | ✅ | `model.go:136-139` — `URLs map[string]string` on `createResp`; `api.go:511,689` — both endpoints populate via `publicURLs()`; legacy `url` field unchanged |
| SQL migration: comment-only, INSTANT | ✅ | `sql/20260610000001_...sql` — MODIFY COLUMN changes only COMMENT; no type/default/null change; MySQL 8.0 INSTANT-safe |
| Security: shared middleware chain, same rate-limit buckets | ✅ | `api.go:200-208` — `chain` closure captures same `ipLimit` instance and `w.ipFailureGateMiddleware()`; per-webhook rate-limit keys on `m.WebhookID` (route-suffix-agnostic) |
| Security: body read post-auth, post-rate-limit | ✅ | `api.go:919` — body read is step 4, after token auth (step 2) + group check (step 3) + rate-limit (step 3.5); wider GitHub cap not an amplification surface |
| Type safety: `buildPayload(m, req)` vs old `buildPayload(m, &req)` | ✅ | `req` is now `*pushPayloadReq` (from adapter parse); `buildPayload`/`buildRichTextPayload` both take `*pushPayloadReq` — pointer semantics preserved |
| Test coverage | ✅ | 4 unit test files + 1 e2e file; covers header gate, all event types, skip paths, clipping, commit truncation, body cap asymmetry, auth on adapter routes, WeCom rejections, create URLs |
| i18n guard updated | ✅ | `api_i18n_test.go:22` — adapter files added to `TestIncomingWebhookNoLegacyResponseError`; `api_i18n_test.go:129` — `"event"` added to reason enumeration test |
| Errcode docs updated | ✅ | `pkg/errcode/incomingwebhook.go:43-47` — comment lists new `event` reason |

---

## 2. Findings

### No P0 or P1 findings.

All data-flow paths traced from adapter parse → handlePush → buildPayload → SendMessage → audit. No path produces user-visible incorrect data, bricks a working flow, or is worse than pre-PR behavior.

### P2 Findings (non-blocking)

#### P2-1: `successExtra` is a shared mutable map reference

**File:** `adapter.go:57` — `wecomAdapter.successExtra: map[string]interface{}{"errcode": 0, "errmsg": "ok"}`

**Diff-scope:** New (introduced by this PR).

The `successExtra` map is allocated once at package init and shared across all concurrent requests. `successBody()` (adapter.go:80-87) only **reads** from it via `range`, so this is safe today. However, the contract is implicit — a future change that writes to `successExtra` (e.g., adding a per-request request ID) would create a data race.

**Suggestion:** Either (a) make `successExtra` a `func() map[string]interface{}` that returns a fresh map per call, or (b) add a `// immutable — do not mutate` annotation. Low urgency.

#### P2-2: GitHub adapter doesn't validate `X-Hub-Signature-256` (documented as optional)

**File:** `adapter_github.go` — no HMAC verification.

**Diff-scope:** New (introduced by this PR).

The PR body and README both explicitly state HMAC is "left as a future optional enhancement." Auth relies on the 128-bit URL token (SHA-256 stored, constant-time compare). This is consistent with the epic decision and not a regression. Noting for tracking — if this is ever promoted to mandatory, the adapter parse function would need the request header passed through (it already receives `http.Header`).

#### P2-3: Degenerate push event `before` field not surfaced

**File:** `adapter_github.go:68-78` — `ghPushEvent` struct doesn't declare `before`.

**Diff-scope:** New.

GitHub push events carry a `before` field (the previous HEAD SHA). For force-push detection context in audit/markdown, having the before-SHA could be useful for debugging. Currently the force-push rendering says "force-pushed N commit(s)" without the before/after range. This is a feature gap, not a bug — the rendered message is still correct and useful.

---

## 3. Suggestions

1. **Consider `immutable` annotation on `successExtra`** (see P2-1). One-line comment prevents future data races.

2. **E2E test `TestPush_GitHubLargePayload_Not413`** — the success-path assertion is `NotEqual(413)` + `NotEqual(400)` + `NotEqual(401)`. If the test stub downstream returns 502, the test still passes. Consider asserting `Equal(200)` when the stub is healthy (the PR body mentions this works locally with real WuKongIM). Not blocking — the negative assertions are sufficient to prove the body cap fix works.

3. **README mentions `X-Hub-Signature-256` as future option** — consider adding a tracking issue or TODO comment in `adapter_github.go` so the intent isn't lost.

---

## 4. Additional Observations

- **Clean adapter pattern:** The `pushAdapter` struct-with-function-fields approach is pragmatic for 3 adapters. If the count grows (GitLab, Feishu per the roadmap), consider promoting to an interface. Not urgent.
- **Comment-only SQL migration** is a good practice — avoids structural DDL locks for metadata-only changes. The down-migration correctly restores original comments.
- **`auditSkipped` with `bumpUsed=false`** is semantically correct: skipped deliveries don't count as successful calls in `call_count` / `last_used_at`.
- **Anti-enumeration invariant preserved:** All three adapter routes use the same `failAuth` path; wrong token → uniform 401 regardless of adapter suffix.
- **No legacy `adapterNative` hardcode in handlePush:** Every audit call site uses `ad.name`. The only remaining `adapterNative` usage is in the test-push admin path, which is correct.

---

## 5. Data-Flow Trace

### GitHub push event → message delivery

```
GitHub webhook POST → /v1/incoming-webhooks/:id/:token/github
  → requirePushEnabled() → 404 if disabled
  → localFloorMiddleware() → IP floor
  → ipLimit → per-IP rate gate
  → ipFailureGateMiddleware() → IP failure budget
  → handlePush(c, githubAdapter):
    1. clientIP(c.Request) → trusted IP (rightmost XFF)
    2. cachedQueryByWebhookID(webhook_id) → model + token check
    3. cachedRequireActiveGroup(group_no) → group Normal check
    4. allowPerWebhook(webhook_id) → per-webhook rate limit (shared bucket with native/wecom)
    5. io.LimitReader(body, githubMaxBytes()+1) → read up to 1MiB+1
    6. githubAdapter.parse(X-GitHub-Event, body):
       → switch event: ping→skip, push→renderGitHubPush, ...
       → renderGitHubPush: unmarshal ghPushEvent → build markdown → clipRunes(content, 4000)
       → return &pushPayloadReq{Content: clipped}, "", ""
    7. req.MsgType == "" → text path:
       → req.Content already set by adapter (non-empty for rendered events)
       → utf8.RuneCountInString(req.Content) ≤ 4000 (guaranteed by clipRunes)
       → buildPayload(m, req): from.kind=webhook, space_id=server-derived, Extra discarded
    8. SendMessageWithResult(payload) → downstream delivery
    9. submitSuccess(m, len(body), ip, msgID, "github", true) → async audit
   10. successBody(githubAdapter, msgID, "") → {"status":0,"message_id":N}
```

### GitHub ping → skip + audit

```
  → githubAdapter.parse: event=="ping" → return nil, "ping", ""
  → handlePush: skipReason="ping"
  → submitSkipped(m, len(body), ip, "github", "ping") → audit status=3, bumpUsed=false
  → successBody(githubAdapter, 0, "ping") → {"status":0,"message_id":0,"skipped":"ping"}
```

### WeCom text → message delivery

```
  → wecomAdapter.parse(_, body):
    → unmarshal wecomMsg → msgtype=="text" → content = msg.Text.Content
    → TrimSpace(content) != "" → return &pushPayloadReq{Content: content}, "", ""
  → handlePush text path:
    → req.Content = WeCom content (NOT clipped — caller-authored)
    → maxContentRunes() check → 413 if >4000 runes (caller can fix)
    → buildPayload(m, req) → same server-derived fields
  → successBody(wecomAdapter, msgID, "") → {"status":0,"message_id":N,"errcode":0,"errmsg":"ok"}
```

### WeCom image → explicit rejection

```
  → wecomAdapter.parse: msgtype=="image" → default case → return nil, "", "msg_type"
  → handlePush: invalidReason="msg_type"
  → submitFailure(m, len(body), ip, "wecom", "msg_type", 400)
  → pushPayloadInvalid(c, "msg_type") → 400 with reason=msg_type
```

### Native (backward compatibility)

```
  → nativeAdapter.parse(_, body):
    → json.Unmarshal(body, &req) → return &req, "", ""
  → handlePush: identical to pre-PR behavior (MsgType switch, Text fallback, etc.)
  → successBody(nativeAdapter, msgID, "") → {"status":0,"message_id":N} (no extra fields)
```

---

[Octo-Q] verdict: **APPROVE** — No P0/P1 findings. The adapter pipeline is correctly designed: shared auth/rate-limit/audit chain, per-shape body caps, proper content clipping for platform-generated events, explicit rejection for untranslatable media types, and comprehensive test coverage. Two P2 observations (shared mutable map reference, HMAC deferred) are non-blocking and documented.

Issue #17, opened 2026-05-13
OCTO migration test debt: ~20+ broken tests masked by stale main CI


yujiawei:
## Background

When the OCTO migration moved code into `octo-server`, multiple tests became broken silently because the previous CI never actually ran them (failed at infra layer first).

After YUJ-513 set up MySQL/Redis service containers and isolated DBs, the underlying test failures surfaced. Examples observed:

- `TestBotGroupCreate_RejectMemberOutsideSpace`
- `TestBotGroupMemberAdd_RejectMemberOutsideGroupSpace`
- `TestBotTranscribe_InvalidMode`
- `TestBotTranscribe_MissingAudio`
- `TestBotTranscribe_FileSizeExceeded`
- `TestBotTranscribe_GPT_EditModeRejected`
- `TestBotTranscribe_GPT_EditModeRejected_DefaultMode`
- `TestBotTranscribe_WithPersonalAndMemberContext`
- `TestBotNameValidation/valid_long_name`
- `TestDelete*` (multiple)
- (full list TBD when CI infra PR lands)

## Cause hypothesis

Tests reference behavior or APIs that were renamed/dropped/restructured during the OCTO migration. The migration tooling handled production code paths but not all test fixtures or expected error messages.

## Strategy (chosen 2026-05-13)

Rather than block the unblock-CI PR on hand-fixing every test, the immediate fix lands with these tests skipped via `t.Skip("OCTO migration TODO: see #<this-issue>")`.

This issue tracks the cleanup work to:
1. Inventory every `t.Skip("OCTO migration TODO: ...")` annotation in octo-server
2. Triage each one (real bug / outdated assertion / removed feature)
3. Either fix or delete the test, removing the skip
4. Close this issue when no `OCTO migration TODO` skip annotations remain

## Priority

Medium. main CI is green via skip-first; the missing coverage is real risk but bounded.

## Related

- Multica YUJ-513 (the unblock-CI task)
- OCTO migration archival event (Mininglamp-OSS 9-repo migration, 2026-05-11)

PR #332, opened 2026-06-10
fix(group): clean thread_setting on removal and drop latent thread-module cleanup copy

### LINKED ISSUE #331: Follow-ups from PR #300: align the third thread-cleanup copy, clean thread_setting on group removal, fix Operator field comment
Tracking issue for the three non-blocking follow-ups surfaced while reviewing #300 (the #27 thread-subscription-leak fix). None block #300; grouping them here as agreed by the reviewers (@yujiawei, @mochashanyao, @Jerry-Xin).

---

## 1. `thread.Service.RemoveUserFromGroupThreads` still carries the Issue #27 `JOIN thread_member` bug (latent)

`modules/thread/service.go:1078` `(*Service).RemoveUserFromGroupThreads(groupNo, uid)` lists threads via `DB.QueryThreadsByGroupNoAndUID` (`modules/thread/db.go:524`), which is exactly the membership-gated pattern #300 just removed elsewhere:

```go
// modules/thread/db.go:526-531
.From(dbr.I("thread").As("t")).
    Join(dbr.I("thread_member").As("tm"), "t.id = tm.thread_id").
    Where("t.group_no=? AND tm.uid=? AND t.status!=?", groupNo, uid, ThreadStatusDeleted).
```

A member who never called `JoinThread` (the default for bots) has zero `thread_member` rows → this returns nothing → no `IMRemoveSubscriber`. Same defect as #27.

- **Current impact: none.** This method is **dead in production** — only referenced by tests; the BotFather delete path and all group-removal paths use the `group` module's 3-arg `RemoveUserFromGroupThreads` / shared `removeUserFromGroupThreadsCleanup`, not this one.
- **Risk:** it's exported on `thread.IService`, and its name/comment imply it's safe for leave-group cleanup. A future caller wiring it onto a removal path would silently reintroduce the leak.
- **Action:** either align it with the membership-independent query (`WHERE group_no=? AND status!=3`), or delete it if confirmed unused, or document loudly that group/bot removal must go through the `group` cleanup helper. Note it also has different semantics (transactional, returns error, additionally cleans `thread_setting`) — see item 2.

## 2. `thread_setting` is not cleaned on the group-removal path (pre-existing)

The shared `removeUserFromGroupThreadsCleanup` (`modules/group/thread_cleanup.go`) cleans `thread_member`, IM subscriptions, pinned, and `conversation_ext`, but **not** `thread_setting`. A user who only muted a thread (set a `thread_setting` row without joining) can leave an orphaned row after being removed from the group.

- **Not a regression** — the group path never cleaned `thread_setting`; only `thread.Service.RemoveUserFromGroupThreads` does (via `DB.DeleteSettingsByGroupNoAndUIDTx`, `modules/thread/db.go:543`), and that method isn't on the group-removal path.
- **Impact:** low — an orphaned mute/setting row, no subscription leak.
- **Action:** add `thread_setting` cleanup (keyed by `group_no` + `uid`) to the shared group helper for full hygiene, ideally folded together with item 1's consolidation.

## 3. `OrgEmployeeExitReq.Operator` field comment is misleading (octo-lib)

In `octo-lib` `config/msg_group.go:477`:

```go
Operator string `json:"operator"` // 操作者uid
```

The handler (`handleOrgEmployeeExit`) consistently uses `req.Operator` as the **leaving user's** uid — for `DeleteMemberTx`, the parent-channel `IMRemoveSubscriber`, and (after #300) the thread cleanup. The "操作者uid" (operator uid) comment is misleading.

- **Lives in `octo-lib`, not this repo** — needs a change there.
- **Action:** clarify the comment to e.g. `// 退出用户的uid`. Comment-only; no behavior change.

---

### Suggested priority
All P2 / low. Items 1 and 2 are naturally fixed together (consolidate the three historical copies onto one membership-independent helper that also clears `thread_setting`). Item 3 is a one-line doc fix in octo-lib.

Refs #27, #300.

an9xyz:
Closes #331. Follow-ups from PR #300 / Issue #27 — this PR fixes items 1 + 2. Item 3 (`OrgEmployeeExitReq.Operator` comment) lives in octo-lib and is split into a dedicated octo-lib issue (cross-referenced on #331), so closing #331 here does not lose track of it.

## Summary

Two changes, fixed together as suggested in #331:

1. **`thread_setting` is now cleaned on the group-removal path** (item 2). The consolidated `removeUserFromGroupThreadsCleanup` (`modules/group/thread_cleanup.go`) cleaned `thread_member`, IM subscriptions, pinned, and `conversation_ext` — but not `thread_setting`. A user who only muted a thread (setting row, no `thread_member` row) left an orphaned mute row after being removed from the group, which would silently re-apply if they ever rejoined.

2. **`thread.Service.RemoveUserFromGroupThreads` (2-arg) is removed** (item 1). It was dead in production (only referenced by `thread.IService` and the thread module's own tests — group/bot removal goes through the `group` module's consolidated helper), but its IM-unsubscribe loop still listed threads via `JOIN thread_member` — the exact Issue #27 pattern #300 removed everywhere else. Any future caller wiring this exported method onto a removal path would have silently reintroduced the subscription leak.

## What changed

- `modules/group/thread_cleanup.go`: `removeUserFromGroupThreadsCleanup` now also deletes `thread_setting` rows keyed by `group_no + uid` (best-effort, log-only on failure, same as the other DB cleanup). The `thread_member` / `thread_setting` DELETEs are no longer gated on the group having non-deleted threads, preserving the removed method's "clean settings even when the user never joined" semantics and sweeping residue rows for deleted threads.
- `modules/thread/`: removed `Service.RemoveUserFromGroupThreads`, its `IService` entry, and the three DB helpers only it used (`QueryThreadsByGroupNoAndUID`, `DeleteMembersByGroupNoAndUIDTx`, `DeleteSettingsByGroupNoAndUIDTx`). Left a signpost comment pointing group/bot removal at the consolidated group helper.
- Tests: new `TestRemoveUserFromGroupThreadsCleanup_CleansSettingWithoutMembership` (mute-without-join regression; asserts uid and group_no isolation), `thread_setting` guard added to `TestRemoveUserFromGroupThreadsCleanup_EmptyInputs`, `ensureThreadTables` now provisions `thread_setting`. The removed method's tests are deleted — 4 of 5 were already `t.Skip` (issue #17 backlog); the live coverage migrated to the group helper's tests.

Net −131 lines.

## Verification

Ran against a CI-equivalent stack (MySQL 8.0.46, Redis 7, WuKongIM v2.2.4-20260313, `OCTO_MASTER_KEY` set, `test` DB recreated `utf8mb4_general_ci` + `FLUSHALL` between packages), `go test -race -shuffle=on -count=1`:

```
ok  github.com/Mininglamp-OSS/octo-server/modules/group      45.6s   # incl. new #331 regressions
ok  github.com/Mininglamp-OSS/octo-server/modules/thread     14.2s
ok  github.com/Mininglamp-OSS/octo-server/modules/botfather  32.4s   # holds thread.IService
ok  github.com/Mininglamp-OSS/octo-server/modules/bot_api     8.9s   # holds thread.IService
```

All seven `TestQueryThreadShortIDsForCleanup_*` / `TestRemoveUserFromGroupThreadsCleanup_*` cases verified individually (PASS, none skipped). `go build ./...` and `go vet` clean.

Additionally ran the opt-in live-WuKongIM E2E (excluded from CI by the `wukong_e2e` build tag) against the same stack, exercising the **modified** helper end-to-end — a bot with no `thread_member` row receives thread messages before cleanup and stops receiving after, while a still-subscribed user keeps receiving:

```
go test -tags wukong_e2e -race -count=1 -run TestE2E_Issue27 -v ./modules/group/
--- PASS: TestE2E_Issue27_RemovedBotStopsReceivingThreadMessages (5.74s)
```

https://claude.ai/code/session_01XMVEVo6tBXp9eYUMWMnM3q

### DIFF
--- a/modules/group/api_groupexit_cascade_thread_test.go
+++ b/modules/group/api_groupexit_cascade_thread_test.go
@@ -11,7 +11,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// ensureThreadTables 为 group 模块的测试脚手架补上 thread / thread_member 表。
+// ensureThreadTables 为 group 模块的测试脚手架补上 thread / thread_member / thread_setting 表。
 // group 模块的测试不会自动运行 thread 模块的迁移，沿用 TestMain 里手工建表的同款做法。
 func ensureThreadTables(t *testing.T, f *Group) {
 	t.Helper()
@@ -40,7 +40,20 @@ func ensureThreadTables(t *testing.T, f *Group) {
 		UNIQUE KEY uk_thread_uid (thread_id, uid)
 	) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4`).Exec()
 	require.NoError(t, err)
+	_, err = f.ctx.DB().UpdateBySql(`CREATE TABLE IF NOT EXISTS thread_setting (
+		id BIGINT UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+		group_no VARCHAR(40) NOT NULL DEFAULT '',
+		short_id VARCHAR(32) NOT NULL DEFAULT '',
+		uid VARCHAR(40) NOT NULL DEFAULT '',
+		mute TINYINT NOT NULL DEFAULT 0,
+		version BIGINT NOT NULL DEFAULT 0,
+		created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+		updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+		UNIQUE KEY uk_thread_uid (group_no, short_id, uid)
+	) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4`).Exec()
+	require.NoError(t, err)
 	// CleanAllTables 之后再 delete 一遍，保证干净
+	_, _ = f.ctx.DB().DeleteFrom("thread_setting").Exec()
 	_, _ = f.ctx.DB().DeleteFrom("thread_member").Exec()
 	_, _ = f.ctx.DB().DeleteFrom("thread").Exec()
 }
--- a/modules/group/thread_cleanup.go
+++ b/modules/group/thread_cleanup.go
@@ -40,10 +40,11 @@ func queryThreadShortIDsForCleanup(ctx *config.Context, groupNo string) ([]strin
 	return out, nil
 }
 
-// removeUserFromGroupThreadsCleanup 是被踢/退群路径上“清理某用户在某群所有子区的 thread_member
-// 记录、IM 订阅和置顶”的统一入口。原本 (*Group).removeUserFromGroupThreads 与
+// removeUserFromGroupThreadsCleanup 是被踢/退群路径上“清理某用户在某群所有子区的 thread_member、
+// thread_setting 记录、IM 订阅和置顶”的统一入口。原本 (*Group).removeUserFromGroupThreads 与
 // (*Service).removeUserFromGroupThreads 是两份逐字重复的实现且各自带 Issue #27 的同型 bug；
-// 这里合并到一处，避免下次再“修一处漏一处”。
+// 这里合并到一处，避免下次再“修一处漏一处”。thread 模块曾有一份同名导出方法，IM 退订同样
+// 被 JOIN thread_member 限定（#27 同型隐患），已随 Issue #331 移除——群/Bot 移除路径只能走这里。
 //
 // 失败只记日志、不中断（与原行为一致）。logger 由调用方传入以保留各自的 module 标签。
 func removeUserFromGroupThreadsCleanup(ctx *config.Context, logger log.Log, groupNo, uid, spaceID string) {
@@ -55,16 +56,20 @@ func removeUserFromGroupThreadsCleanup(ctx *config.Context, logger log.Log, grou
 		logger.Error("查询群子区失败", zap.Error(err), zap.String("groupNo", groupNo), zap.String("uid", uid))
 		return
 	}
-	if len(shortIDs) == 0 {
-		return
-	}
 
-	// best-effort 删除 thread_member 行：DELETE 按 uid 过滤，没有匹配行也是 0 rows affected。
-	// 即使删除失败也要继续摘 IM 订阅 —— 不能让 DB 异常导致订阅泄漏。
+	// best-effort 删除 thread_member / thread_setting 行：DELETE 按 uid 过滤，没有匹配行也是
+	// 0 rows affected。即使删除失败也要继续摘 IM 订阅 —— 不能让 DB 异常导致订阅泄漏。
+	// 两条 DELETE 不 gate 在 len(shortIDs) 上：用户可能只 mute 过子区而从未 JoinThread
+	// （Issue #331），thread_setting 必须按 group_no+uid 直删，连同已删除子区的残留行一并清掉。
 	if _, err := ctx.DB().DeleteFrom("thread_member").
 		Where("uid=? AND thread_id IN (SELECT id FROM thread WHERE group_no=?)", uid, groupNo).
 		Exec(); err != nil {
 		logger.Error("删除子区成员记录失败", zap.Error(err), zap.String("groupNo", groupNo), zap.String("uid", uid))
+	}
+	if _, err := ctx.DB().DeleteFrom("thread_setting").
+		Where("group_no=? AND uid=?", groupNo, uid).
+		Exec(); err != nil {
+		logger.Error("删除子区个人设置失败", zap.Error(err), zap.String("groupNo", groupNo), zap.String("uid", uid))
 	}
 
 	for _, shortID := range shortIDs {
--- a/modules/group/thread_cleanup_test.go
+++ b/modules/group/thread_cleanup_test.go
@@ -124,6 +124,60 @@ func TestRemoveUserFromGroupThreadsCleanup_DeletesMemberWithoutOwnRow(t *testing
 	assert.Equal(t, 1, otherCount, "cleanup 只能按 uid 删除，不能波及其他用户的 thread_member")
 }
 
+// TestRemoveUserFromGroupThreadsCleanup_CleansSettingWithoutMembership 是 Issue #331（item 2）
+// 的回归：用户只对子区设置过 mute（thread_setting 有行）但从未 JoinThread（thread_member 无行），
+// 被移出群后 setting 行必须被清掉，否则重新入群时老 mute 静默生效。同时断言按 uid / group_no
+// 隔离：其他用户的 setting、该用户在其他群的 setting 都不受影响。
+func TestRemoveUserFromGroupThreadsCleanup_CleansSettingWithoutMembership(t *testing.T) {
+	svc, _ := setupServiceTest(t)
+	s := svc.(*Service)
+	f := New(s.ctx)
+	ensureThreadTables(t, f)
+
+	const groupNo = "g_issue331_setting"
+	const targetUID = "mute_only_user"
+	_, err := f.ctx.DB().InsertInto("thread").
+		Columns("short_id", "group_no", "name", "creator_uid", "status", "version").
+		Values("th_331", groupNo, "setting", "owner", 1, 1).Exec()
+	require.NoError(t, err)
+
+	// target 只 mute、未 join：thread_setting 有行，thread_member 无行
+	_, err = f.ctx.DB().InsertInto("thread_setting").
+		Columns("group_no", "short_id", "uid", "mute", "version").
+		Values(groupNo, "th_331", targetUID, 1, 1).Exec()
+	require.NoError(t, err)
+	// 其他用户在同子区的 setting —— 不能被波及
+	_, err = f.ctx.DB().InsertInto("thread_setting").
+		Columns("group_no", "short_id", "uid", "mute", "version").
+		Values(groupNo, "th_331", "someone_else", 1, 1).Exec()
+	require.NoError(t, err)
+	// target 在另一个群的 setting —— 不能被波及
+	_, err = f.ctx.DB().InsertInto("thread_setting").
+		Columns("group_no", "short_id", "uid", "mute", "version").
+		Values("g_other_331", "th_other_331", targetUID, 1, 1).Exec()
+	require.NoError(t, err)
+
+	s.removeUserFromGroupThreads(groupNo, targetUID, "sp_331")
+
+	var targetCount int
+	_, err = f.ctx.DB().Select("count(*)").From("thread_setting").
+		Where("group_no=? AND uid=?", groupNo, targetUID).Load(&targetCount)
+	require.NoError(t, err)
+	assert.Equal(t, 0, targetCount, "mute 而未 join 的用户被移出群后，thread_setting 必须被清理（Issue #331）")
+
+	var otherUserCount int
+	_, err = f.ctx.DB().Select("count(*)").From("thread_setting").
+		Where("group_no=? AND uid=?", groupNo, "someone_else").Load(&otherUserCount)
+	require.NoError(t, err)
+	assert.Equal(t, 1, otherUserCount, "cleanup 只能按 uid 删除，不能波及其他用户的 thread_setting")
+
+	var otherGroupCount int
+	_, err = f.ctx.DB().Select("count(*)").From("thread_setting").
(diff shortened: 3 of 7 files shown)


Jerry-Xin (approved):
The PR is in scope for `octo-server` and the cleanup change matches the group/thread removal architecture. I found no blocking correctness, security, or build issues.

**💬 Non-blocking**

🔵 Suggestion: Add one regression case for the “no non-deleted threads, but `thread_setting` residue exists” path. The implementation now deletes settings after `queryThreadShortIDsForCleanup` even when `shortIDs` is empty, which is good, but the new test at `modules/group/thread_cleanup_test.go:131` seeds an active thread, so it does not directly cover the no-active-thread behavior described in the PR. Relevant code: `modules/group/thread_cleanup.go:54` and `modules/group/thread_cleanup.go:69`.

**✅ Highlights**

- `modules/group/thread_cleanup.go:64` correctly removes `thread_member` by `uid` and group-scoped thread IDs without depending on membership enumeration.
- `modules/group/thread_cleanup.go:69` correctly removes `thread_setting` by `group_no + uid`, covering mute-only users and deleted-thread residue.
- `modules/thread/service.go:35` no longer exposes the stale `RemoveUserFromGroupThreads` method on `thread.IService`, preventing future callers from reusing the old membership-joined cleanup pattern.
- The new tests cover mute-without-membership cleanup, UID isolation, group isolation, and empty-parameter guards.

Verification: `git diff --check` passed. Focused `modules/thread` tests passed. Compile checks for `modules/botfather` and `modules/bot_api` passed. Focused `modules/group` tests could not run locally because MySQL was unavailable on `127.0.0.1:3306`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#332 Review Report
**PR:** [fix(group): clean thread_setting on removal and drop latent thread-module cleanup copy](https://github.com/Mininglamp-OSS/octo-server/pull/332)
**Head SHA:** `be2604a52c81330d154eeb731e89de0e05adb7a8`

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `thread_setting` cleanup on removal path | ✅ | `modules/group/thread_cleanup.go:69-72` — DELETE by `group_no=? AND uid=?` |
| Removed `thread.Service.RemoveUserFromGroupThreads` has no remaining callers | ✅ | grep for `QueryThreadsByGroupNoAndUID`, `DeleteMembersByGroupNoAndUIDTx`, `DeleteSettingsByGroupNoAndUIDTx` → 0 matches post-PR |
| `thread.IService` interface updated | ✅ | `modules/thread/service.go:62-63` — entry removed; no callers of `thread.IService.RemoveUserFromGroupThreads` remain |
| `removeUserFromGroupThreadsCleanup` is sole cleanup entry point | ✅ | Called from `group/api.go:3046`, `group/service.go:1856`, `botfather/command.go:665` (via `group.Service.RemoveUserFromGroupThreads`) |
| Test schema matches production migration | ✅ | Test DDL in `api_groupexit_cascade_thread_test.go:43-53` matches `modules/thread/sql/20260422000001_thread_legacy01.sql:3-16` (minor: test omits `idx_uid` index and COLLATE — cosmetic only) |
| New regression test covers mute-without-join | ✅ | `thread_cleanup_test.go:131-178` — asserts target cleanup + uid isolation + group_no isolation |
| Empty-input guard covers `thread_setting` | ✅ | `thread_cleanup_test.go:196-217` — asserts no DELETE fires on empty uid/groupNo |
| `go build` / `go vet` / tests pass | ✅ | Per PR description: all 4 packages pass with `-race -shuffle=on` |

## 2. Findings

### No P0/P1 findings.

### P2 — Minor observations

**P2-1: Test schema omits `idx_uid` secondary index**
- **Diff-scope:** pre-existing (test scaffolding never had it; not introduced by this PR)
- Production migration (`20260422000001_thread_legacy01.sql:15`) defines `KEY idx_uid (uid)`, but the test DDL in `api_groupexit_cascade_thread_test.go` does not. The new DELETE uses `WHERE group_no=? AND uid=?` (hits the unique key), so the missing index doesn't affect test correctness. In production, `idx_uid` would only matter for queries filtering by `uid` alone without `group_no` — none of which exist in the current codebase.
- **Suggestion:** Add `KEY idx_uid (uid)` to the test DDL for schema parity, low priority.

**P2-2: No explicit test for `thread_setting` cleanup on deleted-thread residue**
- **Diff-scope:** new gap (the PR's stated motivation includes "sweeping residue rows for deleted threads" but the test only covers active threads)
- The comment at `thread_cleanup.go:63` explicitly states the DELETE is designed to clean settings for deleted threads too. The test `TestRemoveUserFromGroupThreadsCleanup_CleansSettingWithoutMembership` only inserts a status=1 thread. A test with a status=3 thread + orphaned setting would lock in this guarantee.
- **Suggestion:** Add a deleted-thread setting residue case to the regression test. Low risk — the SQL is straightforward and correct.

## 3. Data-Flow Tracing

|---------------|----------------|-----------------|
| `groupNo` param → `thread_setting` DELETE WHERE | Caller: `group/api.go:3045` (from `groupNo` API param), `group/service.go:1855` (from `req.GroupNo`), `botfather/command.go:665` (from `g.GroupNo`) | ✅ All callers pass validated group identifiers |
| `uid` param → `thread_setting` DELETE WHERE | Same callers pass target user's UID | ✅ Correctly scoped to target user |
| `thread_setting.group_no` column | Production schema: `VARCHAR(40) NOT NULL DEFAULT ''` | ✅ Matches WHERE clause `group_no=?` |
| `thread_setting.uid` column | Production schema: `VARCHAR(40) NOT NULL DEFAULT ''` | ✅ Matches WHERE clause `uid=?` |
| `shortIDs` → IM unsubscribe loop | `queryThreadShortIDsForCleanup` → `SELECT short_id FROM thread WHERE group_no=? AND status!=3` | ✅ Unchanged from pre-PR; correctly excludes deleted threads (IM channel already torn down) |
| `thread_member` DELETE subquery | `SELECT id FROM thread WHERE group_no=?` (no status filter) | ✅ Broader than old code (which gated on non-deleted threads via `shortIDs`); correctly catches orphaned member rows for deleted threads |

### Key behavior change verified
The PR removes the `if len(shortIDs) == 0 { return }` early-exit gate. Previously, if a group had zero non-deleted threads, the entire cleanup (including `thread_member` DELETE) was skipped. Now both DELETEs always fire. This is correct:
- `thread_member` DELETE with empty subquery → 0 rows affected (safe no-op)
- `thread_setting` DELETE by `group_no+uid` → cleans any residue regardless of thread status

## 4. Suggestions

1. (P2-2 above) Add a deleted-thread setting residue test case.
2. (P2-1 above) Add `idx_uid` to test DDL for production parity.
3. Consider adding a brief CHANGELOG/release-note entry since this fixes a user-visible bug (orphaned mute settings re-applying on rejoin).

## 5. Extra Findings

None. The PR is focused and well-scoped. The removal of the dead `thread.Service.RemoveUserFromGroupThreads` and its 3 dedicated DB helpers is clean — no dangling references, no partial removals.

## 6. R5 Blind-Spot Checklist

- **C1 — Double-path parity:** N/A. This PR consolidates cleanup to a single path (the group module's `removeUserFromGroupThreadsCleanup`). There is no add/join symmetry concern here — `thread_setting` rows are created on-demand via `UpsertSetting` (not bulk on join), so removal on group-exit is the correct symmetric counterpart.
- **C2 — Control-flow ordering / nesting reuse:** Clear. `removeUserFromGroupThreadsCleanup` is a leaf function called from 3 sites (group api, group service, botfather). No nesting, no double-invocation risk. The DELETE statements are simple single-table operations with no composite call chains.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This is a data-cleanup function on authenticated removal paths. Authorization is enforced upstream (group membership checks, operator validation in the API/service layers). The cleanup function itself is unexported (`removeUserFromGroupThreadsCleanup`) and only reachable through the group module's validated entry points.

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. The PR correctly fixes the orphaned `thread_setting` bug (Issue #331 item 2), cleanly removes dead code with a latent subscription-leak hazard (item 1), and adds solid regression coverage with uid/group_no isolation guards. Two P2 observations (test schema parity, deleted-thread residue test) are non-blocking.

yujiawei (approved):
# Code Review — PR #332 (octo-server)

## Verdict: APPROVED ✅

This is a clean, well-scoped fix that closes a real orphaned-mute bug and removes a latent footgun. No correctness, security, or data-loss issues. The two changes are correct and net-negative on lines (−131). Approving; the items below are non-blocking suggestions only.

## What was verified

**1. `thread_setting` cleanup on removal (item 2) — correct.**
- `removeUserFromGroupThreadsCleanup` (`modules/group/thread_cleanup.go:69-73`) now deletes `thread_setting WHERE group_no=? AND uid=?`. This is scoped to a single (user, group) and cannot leak across users or groups — `thread_setting.group_no` is the parent-group column (same value space as `thread.group_no`), written from the same context in `UpsertSetting` (`modules/thread/db.go:812-819`). The new regression test explicitly asserts cross-user and cross-group isolation.
- Removing the `if len(shortIDs)==0 { return }` early guard is a **strict improvement**, not a behavior regression. Previously, a group whose threads were all deleted (`status=3`) produced an empty `shortIDs` and short-circuited *both* DELETEs — leaving orphan `thread_member`/`thread_setting` rows behind. The two DELETEs are still bounded (the `groupNo==""||uid==""` guard at `thread_cleanup.go:51-53` prevents any unbounded delete), so the only newly-enabled path is residue cleanup. This directly fixes the documented Issue #331 mute-without-join case.
- Broad-DB / narrow-IM split is intentional and correct: the DB DELETEs sweep deleted-thread residue, while the IM-unsubscribe loop iterates `queryThreadShortIDsForCleanup` (`status!=3`) — deleted threads' IM channels are already destroyed, so excluding them is right.

**2. Removal of dead `thread.Service.RemoveUserFromGroupThreads` + 3 DB helpers (item 1) — safe.**
- Repo-wide grep confirms **no surviving caller** of the removed 2-arg method or the three removed DB helpers (`QueryThreadsByGroupNoAndUID`, `DeleteMembersByGroupNoAndUIDTx`, `DeleteSettingsByGroupNoAndUIDTx`). The kept 3-arg `group.Service.RemoveUserFromGroupThreads` (used by `botfather/command.go:665`) is unaffected.
- The removed method is not part of `thread.IService` (the interface entry was dropped here), no mocks or `var _ IService` assertions reference it, and all imports remain used.
- `go build ./...`, `go vet ./modules/group/... ./modules/thread/...`, and test compilation are all clean against the head SHA.
- Removing this method also genuinely closes a latent Issue #27 vector: its IM-unsubscribe loop listed threads via `JOIN thread_member`, so any future wiring onto a removal path would have re-introduced the subscription leak for non-joined members (e.g. bots).

## Non-blocking suggestions (P2 / nits — do not block merge)

- **Test scaffold drift (`modules/group/api_groupexit_cascade_thread_test.go:43-54`):** the test `thread_setting` table omits the secondary `KEY idx_uid(uid)` that production has. This does not affect correctness — the `WHERE group_no=? AND uid=?` DELETE already uses the leftmost prefix (`group_no`) of `uk_thread_uid(group_no, short_id, uid)` in both environments, so the access path is equivalent. Adding the index to the scaffold would keep it schema-faithful, but it's cosmetic.
- **IM-unsubscribe is not asserted in tests:** the cleanup tests assert DB-row deletion only; none assert `IMRemoveSubscriber` is invoked per non-deleted thread, and the `EmptyInputs` test message claims "no IMRemoveSubscriber dispatch" without a spy to observe it. No coverage is *lost* (the deleted thread-module tests didn't assert it either, and 4 of 5 were `t.Skip`), but since this consolidation exists to protect exactly that subscription-leak contract, a recording IM stub asserting one call per active+archived thread (zero for deleted) would be the higher-value test to add later.
- **Two DELETEs are best-effort, not transactional:** unlike the removed (dead) `thread.Service` path, the consolidated helper runs the two DELETEs as independent `Exec` calls with log-and-continue. This matches the helper's documented best-effort contract and the pre-PR group-module behavior, so it's not a regression — a partial-cleanup residue row is re-deleted on the next removal. Noted for awareness only.

## Coverage check
The deleted thread-module tests lose no executing coverage: 4 of 5 were `t.Skip` (Issue #17 backlog), and the one active case (`_NoThreads`) is now covered by the group-module cleanup/query tests. The migrated `_CleansSettingWithoutMembership` is upgraded from skipped to running and adds isolation assertions the old test lacked.

PR #336, opened 2026-06-10
feat(openapi): adopt openapi toolchain + first annotated endpoint + 3.1 baseline

liuooo:
## What

Adopts the octo-openapi-dev-skill toolchain in octo-server and ships the first spec-compliant annotated endpoint plus the generated OpenAPI 3.1 baseline. This is the adoption-guide "brownfield" path: the toolchain and baseline land first, handler annotation coverage migrates module-by-module afterwards.

## How

- **`Makefile` + `tools/octo-api/`** (commit 1): vendor the toolchain (spectral ruleset with 30 OCTO rules, scripts, `openapi.mk`) — provides `make openapi-check / gen / lint / diff / coverage / doctor`.
- **`main.go`**: swag global annotations (`@title`, `@BasePath /v1` → `servers:[{url:/v1}]`, Bearer scheme, `user` tag).
- **`modules/user/api.go`**: `GET /v1/user/qrcode` fully annotated as the showcase — its success response is already the R1 envelope shape (`{"data": url}`), so the handler only switches to a typed `qrcodeMyResp` with identical wire bytes.
- **`pkg/httperr/respond.go`**: new `ErrorResp` mirroring the i18n renderer's error envelope (incl. D14 `msg`/`status` siblings) for `@Failure` annotations across modules.
- **`docs/openapi/`**: generated baseline — spectral-clean at error level; the remaining warnings document the known D14-vs-R2 gap (wire status pinned to 400, i18n code ids instead of the 12-code enum).
- `countriesList` deliberately stays un-annotated: its bare-array response violates R1 and reshaping is a breaking change requiring the deprecate flow; it is tracked by the coverage gate.

## Gates status

| gate | status |
|---|---|
| lint | ✅ 0 errors |
| verify (gen + drift) | ✅ after this PR (baseline committed) |
| coverage | ❌ 1/443 — staged migration per adoption.md; NOT a required check yet |

## Related

- Toolchain: https://github.com/liuooo/octo-openapi-dev-skill
- Reference response abstraction: Mininglamp-OSS/octo-lib#74 (envelope types + wkhttp helpers; once merged, handlers can emit R1 shapes via `c.ResponseData` / `wkhttp.ResponseCursor` and reference `envelope.*` in annotations — verified swag path: blank import + `--parseDependencyLevel 1`, already wired in `openapi-gen`)

## Tests

`make openapi-check` locally: doctor ✅ → coverage ❌ (expected, staged) → verify ✅ → lint ✅ (0 errors). `go build` green on touched packages.


mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#336 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/336
**Head SHA:** `577189315fb872486c14fa3a7a626b1a33a65ba4`
**Title:** feat(openapi): adopt openapi toolchain + first annotated endpoint + 3.1 baseline

---

## 1. PR Scope Summary

Two commits on top of the branch history:

| Commit | Description | Files |
|--------|-------------|-------|
| `ecbcfd79` | Vendor octo-openapi toolchain (`tools/octo-api/` @ 68829c3) | 15 new files (SKILL.md, spectral.yaml, openapi.mk, scripts, JS functions, templates, references) |
| `57718931` | First annotated endpoint + OpenAPI 3.1 baseline | 8 files (main.go annotations, user/api.go qrcode annotation + typed resp, httperr ErrorResp, generated docs/openapi/*, common/api.go comment) |

**Net code change (PR-specific):** ~740 additions, 2 deletions. The vast majority is vendored toolchain and generated docs. Actual production code delta: ~50 lines.

---

| Item | Status | Evidence |
|------|--------|----------|
| Wire-format backward compatibility | ✅ | `qrcodeMyResp{Data: "..."}` serializes to `{"data":"..."}` — byte-identical to prior `gin.H{"data": "..."}` (modules/user/api.go:740) |
| Auth model unchanged | ✅ | `@Security Bearer` annotation is metadata-only; `AuthMiddleware` on the route group is untouched (modules/user/api.go:230) |
| No new runtime code paths | ✅ | All annotations are godoc comments parsed by swag at build time; `ErrorResp` is a type definition never instantiated at runtime |
| Shell scripts safe | ✅ | All 4 scripts use `set -euo pipefail`, proper `trap` cleanup, no unsanitized input in shell expansion |
| CI workflow hygiene | ✅ | Pinned action SHAs, `permissions: contents: read`, `timeout-minutes` on all jobs, `persist-credentials: false` |
| Generated spec accuracy | ✅ | swagger.yaml correctly reflects the annotated endpoint: `/user/qrcode [get]`, Bearer security, 200/400/401 responses |
| Makefile include safe | ✅ | `include tools/octo-api/assets/openapi.mk` is standard; all targets are `.PHONY`; tool installs are opt-in |

---

## 3. Findings

### F1 — P2: Committed `docs/openapi/index.html` loads external CDN script

**File:** `docs/openapi/index.html:10`
**Severity:** P2 (nit)

```html
<script src="https://cdn.redocly.com/redoc/v2.5.3/bundles/redoc.standalone.js"></script>
```

The generated HTML loads a third-party JS bundle from `cdn.redocly.com`. While this is standard Redoc output and the file is committed as part of the documentation baseline, it introduces an external dependency in the repo tree. The `openapi.mk` `openapi-preview` target explicitly says the HTML is "throwaway — add to .gitignore."

**Recommendation:** Consider adding `docs/openapi/index.html` to `.gitignore` and only generating it locally via `make openapi-preview`. The committed baseline (`swagger.yaml`, `swagger.json`, `docs.go`) is sufficient for CI verification. This avoids tracking a 345-line CDN-dependent HTML file that will change whenever Redoc version bumps.

### F2 — P2: `normalize-spec.sh` embeds inline Python — fragile coupling to swag quirk

**File:** `tools/octo-api/scripts/normalize-spec.sh`
**Severity:** P2 (maintainability)
**Diff-scope:** new (vendored from toolchain)

The script embeds a Python heredoc to strip empty `externalDocs` blocks from swag v2.0.0-rc5 output. The script's own comment acknowledges "When swag fixes this upstream, remove the normalization step." This is fine for now but should be tracked as tech debt — if swag fixes the quirk and the normalize step is forgotten, it could silently corrupt a future spec that legitimately has `externalDocs`.

**Recommendation:** Add a comment in `openapi.mk` near the `normalize-spec.sh` call referencing the swag upstream issue, so future maintainers know when to remove it.

### F3 — P2: `ErrorResp.Details` uses `map[string]any` — schema precision

**File:** `pkg/httperr/respond.go:84`
**Severity:** P2 (nit)

```go
Details    map[string]any `json:"details"`
```

The generated OpenAPI schema renders this as `additionalProperties: {}` (any type), which is technically correct but provides no schema guidance to API consumers. The actual details shape depends on the specific error code's registered detail params. This is inherent to the polymorphic nature of the field and not fixable without per-endpoint error types.

**Recommendation:** No action needed — documenting as a known limitation is sufficient. The PR body already acknowledges the D14 gap.

---

## 4. Data-Flow Trace

### `qrcodeMy` handler (modules/user/api.go:727-742)

| Consumed Data | Upstream Source | Traced? | Notes |
|---------------|----------------|---------|-------|
| `c.GetLoginUID()` | `AuthMiddleware` on `/v1/user` group (modules/user/api.go:230) → `pkg/auth.CacheTokenParser` → Redis token cache | ✅ | Auth path unchanged by this PR |
| `userModel.QRVercode` | `DB.QueryByUID(loginUID)` → `user` table `qr_vercode` column | ✅ | Set at user creation (api.go:3404) as `UUID@QRCode`; empty check at line 735 returns `ErrUserQRVerCodeMissing` |
| `u.ctx.GetConfig().QRCodeInfoURL` | Server config (YAML/env) | ✅ | Template string with `:code` placeholder; replaced with `vercode_{QRVercode}` |
| `u.ctx.GetConfig().External.BaseURL` | Server config (YAML/env) | ✅ | URL prefix for the QR code info endpoint |
| `qrcodeMyResp.Data` | `fmt.Sprintf("%s/%s", BaseURL, path)` | ✅ | Wire bytes `{"data":"<url>"}` identical to prior `gin.H{"data": ...}` |

**Conclusion:** All consumed data traces to valid upstream sources. No empty-value / short-circuit / swallowed-error risks. The refactor from `gin.H` to typed struct is wire-compatible.

---

## 5. R5 Blind-Spot Checklist (security_sensitive)

### C1 — Dual-path parity
**N/A.** This PR does not touch any paired/symmetric paths (add↔remove, subscribe↔unsubscribe, create↔delete). The only production code change is a response-type refactor on a single GET endpoint.

### C2 — Control-flow ordering / nested reuse
**N/A.** No emit/sanitize/escape functions are introduced or modified. The swag annotations are parsed at build time, not at runtime. The shell scripts are standalone utilities with no nested call-site risk.

**N/A.** No new endpoints, tools, or capabilities are exposed. The `@Security Bearer` annotation documents the existing auth middleware; it does not create or modify any authorization boundary. The toolchain scripts run only in CI/local dev, not in the production server process.

---

## 6. Additional Observations

- **countriesList deliberately un-annotated** (modules/common/api.go:531-532): The comment correctly documents that the bare-array response violates R1 envelope and that reshaping is a breaking change. This is tracked by the coverage gate. Sound decision.
- **Coverage gate is not yet a required check** (PR body: "1/443 — staged migration"): Appropriate for brownfield adoption. The coverage script correctly identifies the gap.
- **Toolchain vendoring is clean**: The `tools/octo-api/` layout follows the skill package convention. No executable permissions issues found in shell scripts.

---

## 7. Verdict

**No P0 or P1 findings.** All three findings are P2 (nit/maintainability). The PR is a documentation/tooling adoption with zero behavioral change to the production server. The single handler refactor (`qrcodeMy`) is wire-compatible and data-flow verified.

[Octo-Q] verdict: **APPROVE** — no blocking issues found. The three P2 nits are optional improvements for follow-up.

yujiawei (changes requested):
# Code Review — PR #336 (octo-server)

## Summary

This PR vendors the `octo-openapi` toolchain under `tools/octo-api/` (spectral ruleset, scripts, Makefile include, skill docs), wires swag global annotations into `main.go`, annotates one showcase endpoint (`GET /v1/user/qrcode`), adds `pkg/httperr.ErrorResp` for `@Failure` annotations, and commits a generated OpenAPI 3.1 baseline under `docs/openapi/`.

The toolchain itself is well thought through (aggregating four-gate check, staged-migration coverage path, breaking-change detection, doctor preflight), and the production-code surface is intentionally minimal. The design intent is sound. However, there is a blocking build break in the committed baseline that must be fixed before merge.

## Verdict: CHANGES_REQUESTED

---

## P0 — Build break: `docs/openapi/docs.go` imports a module not in `go.mod`

The committed generated file imports a package that is not a declared dependency:

```go
// docs/openapi/docs.go:5
import "github.com/swaggo/swag/v2"
```

`github.com/swaggo/swag/v2` appears **nowhere** in `go.mod` or `go.sum` (0 matches). As a result the package fails to compile:

```
$ go build ./...
docs/openapi/docs.go:5:8: no required module provides package github.com/swaggo/swag/v2; to add it:
	go get github.com/swaggo/swag/v2
$ echo $?
1
```

`go vet ./...` fails identically.

This is not theoretical for CI. The PR changes non-doc Go files (`main.go`, `modules/user/api.go`, `modules/common/api.go`, `pkg/httperr/respond.go`), so the `changes` preflight in `.github/workflows/ci.yml` sets `code=true`, and the **Build** job (`CGO_ENABLED=0 go build -v ./...`, ci.yml:100) and **Vet** job (`go vet ./...`, ci.yml:262) both run and walk `docs/openapi`. Both will fail.

Note the `changes` filter excludes `docs/**` from the `code` output, but it evaluates over *all* changed files with the `every` quantifier and the non-doc `.go` files flip `code=true` — so excluding `docs/**` does not save this; the doc package is still compiled by `go build ./...`.

The repo's own `make openapi-verify` gate treats `docs.go` as part of the committed baseline (drift check), so dropping the file is not the intended fix.

**Required fix:** add the dependency so the generated file compiles, e.g.

```
go get github.com/swaggo/swag/v2@v2.0.0-rc5
go mod tidy
git add go.mod go.sum
```

Then re-run `go build ./...` / `go vet ./...` to confirm green.

---

## P2 — Stale, non-reproducible `docs/openapi/index.html` committed

`docs/openapi/index.html` (345 lines) is committed, but it documents a **different** API surface than the actual baseline:

- `swagger.yaml` / `swagger.json` / `docs.go` describe exactly one path: `GET /user/qrcode`.
- `index.html` describes a single different path: `GET /common/countries` (operationId `common.countries.list`), an endpoint that does not exist in the committed spec.

So the committed preview neither matches the committed spec nor any state this PR produces. It looks like a leftover artifact from an earlier local run.

Separately, the toolchain's own guidance treats this file as throwaway: `openapi.mk` says *"Output is throwaway — add `$(OPENAPI_OUT_DIR)/index.html` to .gitignore"*, and `toolchain.md` lists `index.html` as *"not commit (加到 .gitignore)"*. It is currently neither in `.gitignore` nor reproducible.

**Suggested fix:** remove `docs/openapi/index.html` from the PR and add `docs/openapi/index.html` to `.gitignore`, consistent with the toolchain's own contract. (Not merge-blocking on its own, but it is actively misleading to anyone who opens it.)

---

## P2 — CI workflow template has broken paths (latent, not active in this PR)

This PR does **not** install `.github/workflows/openapi.yml`, so nothing below breaks today. But `tools/octo-api/assets/templates/openapi-workflow.yml` is the file `adoption.md` tells the next repo to `cp` into `.github/workflows/`, and as written it would fail immediately:

- Line 134: `bash tools/octo-api/check-swag-coverage.sh modules` — the script lives at `tools/octo-api/scripts/check-swag-coverage.sh`. The `scripts/` segment is missing.
- Lines 199 / 228: `tools/octo-api/spectral.yaml` — the ruleset lives at `tools/octo-api/assets/spectral.yaml`. The `assets/` segment is missing. (Line 289's `diff-openapi.sh` ref correctly uses `scripts/`, so the template is internally inconsistent.)
- Line 296: `run: make openapi-toolchain-test` — there is no `openapi-toolchain-test` target in `openapi.mk` (`grep` for it finds only these template references). The Toolchain Self-Test job would fail with "No rule to make target".

The actually-wired local targets in `openapi.mk` use the correct `scripts/` and `assets/` paths, so `make openapi-check` works locally. Only the copy-paste CI template is affected.

**Suggested fix:** correct the three paths/targets in the template, or note in `adoption.md` that the template needs path adjustment, so the first repo that adopts CI doesn't hit a red pipeline on day one.

---

## Things verified as correct (no action needed)

- **`GET /user/qrcode` change is wire-compatible.** `c.Response(qrcodeMyResp{Data: url})` marshals to `{"data": "<url>"}` with HTTP 200 — byte-identical to the previous `c.Response(gin.H{"data": url})` (`octo-lib wkhttp.Context.Response` just does `c.JSON(http.StatusOK, data)`). No breaking change for existing clients. ✅
- **`httperr.ErrorResp` accurately mirrors the runtime error envelope.** It matches `pkg/i18n/renderer.go`'s `Render`: `error.{code,message,details,http_status}` plus the legacy `msg`/`status` D14 siblings. The annotation does not lie about the wire shape. ✅
- **`countriesList` deliberately left un-annotated** with an inline comment explaining its bare-array response violates R1 and reshaping is a breaking change deferred to the deprecate flow. Reasonable, and consistent with the staged-migration approach. ✅
- **Baseline is OpenAPI 3.1** (`openapi: 3.1.0`), `@Router` is correctly relative (no `/v1` prefix; `servers: [{url: /v1}]` supplies it), security scheme is a standard `apiKey` Bearer header. ✅

## Security note (PR classified security-sensitive)

The classification was triggered by auth-related keywords (Bearer security definition, token/credential). The actual changes are benign from a security standpoint: no authentication or authorization logic is modified. The `@securityDefinitions.apikey Bearer` block is pure documentation metadata; the `qrcode` handler's behavior is unchanged (it still embeds the user's `vercode` in the returned URL exactly as before, only the Go type wrapping the response changed). One item worth a human's eye: the `qrcode` description notes the URL embeds the current vercode and that resetting the vercode invalidates prior URLs — confirm that documenting this invalidation semantic publicly is acceptable. No code-level security concern found.

---

Once the P0 dependency is added and `go build ./...` / `go vet ./...` pass, this is good to merge. The P2 items (stale `index.html`, template paths) are cleanups that can ride along or be a fast follow-up.

Jerry-Xin (changes requested):
The PR is in scope for octo-server, but the new OpenAPI coverage gate has a correctness bug that can let unannotated handlers pass.

**🔴 Blocking**
- 🔴 Critical: `tools/octo-api/scripts/check-swag-coverage.sh:36-38` and `:60-65` key handlers only by function name and then `sort -u` them. This repo has repeated handler names across receivers/modules (`list`, `delete`, `login`, etc.); I counted 491 handler functions but only 444 unique method names. Once any one `list` method is annotated, every other unannotated `list` method can be treated as covered. The gate must track a stable handler identity such as `file:receiver.method` or at least `receiver.method`, not only method name.

**💬 Non-blocking**
- 🟡 Warning: The optional workflow template does not match the vendored tool layout. It calls `tools/octo-api/check-swag-coverage.sh` at `tools/octo-api/assets/templates/openapi-workflow.yml:134`, but the script lives under `tools/octo-api/scripts/`. It also points Spectral at `tools/octo-api/spectral.yaml` at `:228`, while the real ruleset is `tools/octo-api/assets/spectral.yaml`, and calls a missing `make openapi-toolchain-test` target at `:296`. If copied as documented, the CI workflow will fail before validating the spec.
- 🟡 Warning: `docs/openapi/index.html` is stale relative to the committed spec. The YAML baseline documents `/user/qrcode` at `docs/openapi/swagger.yaml:44`, but the HTML preview still hydrates `/common/countries` at `docs/openapi/index.html:329-337`. Since `openapi-preview` is described as throwaway output, either regenerate it from the current spec or stop committing it.

**✅ Highlights**
- The `/user/qrcode` handler’s success response remains wire-compatible as `{"data": ...}`.
- `httperr.ErrorResp` matches the current i18n error renderer shape.
- I ran `go test ./pkg/httperr`, `make openapi-verify`, and `make openapi-lint`; verify passed, lint had 0 errors and 5 warnings.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#336 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/336
**Head SHA:** `e6dfc46cf1f56943f05d394fcef87565b21745e0`
**Title:** feat(openapi): adopt openapi toolchain + first annotated endpoint + 3.1 baseline
**Changed files:** 23 (+2484 / −2)

---

## 1. Verification Summary

| Item | Status | Evidence |
|---|---|---|
| Wire-format backward compat (qrcodeMy) | ✅ | `modules/user/api.go:738-740` — `qrcodeMyResp{Data: ...}` serializes to identical `{"data":"<url>"}` as the old `gin.H{"data": ...}` (verified via `wkhttp.Context.Response()` → `c.JSON(200, data)` in `octo-lib/pkg/wkhttp/http.go`) |
| ErrorResp matches renderer wire shape | ✅ | `pkg/httperr/respond.go:84-93` struct fields `{error.{code,message,details,http_status}, msg, status}` exactly match `pkg/i18n/renderer.go` `Render()` JSON output |
| swag global annotations (main.go) | ✅ | `main.go:54-65` — `@title`, `@BasePath /v1`, `@securityDefinitions.apikey Bearer` all present; generated `swagger.yaml` `servers:[{url:/v1}]` correct |
| Generated baseline (swagger.yaml) | ✅ | `docs/openapi/swagger.yaml` — OpenAPI 3.1.0, single path `/user/qrcode`, schemas match source annotations |
| Toolchain vendoring | ✅ | `tools/octo-api/` — spectral ruleset (30 OCTO rules), 3 JS functions, 4 shell scripts, Makefile include; all well-structured |
| .gitignore | ✅ | `docs/openapi/index.html` added — throwaway Redoc preview correctly excluded |
| countriesList deliberate exclusion | ✅ | `modules/common/api.go:531-532` — documented comment explaining bare-array R1 violation, tracked by coverage gate |

## 2. Findings

### F1 — P2: Bearer security scheme rendered as `apiKey` type

**Diff-scope:** new (introduced by this PR's swag annotations)

`swagger.yaml` emits:
```yaml
securitySchemes:
  Bearer:
    in: header
    name: Authorization
    type: apiKey
```

OpenAPI 3.1 has a native `type: http, scheme: bearer` which is semantically correct for Bearer tokens. The `apiKey` representation is a known swag v2 limitation — `@securityDefinitions.apikey` is the only available annotation for token-based auth.

**Impact:** Downstream codegen tools (openapi-generator, orval) will generate API-key-style auth headers instead of proper Bearer token handling. Functionally equivalent at runtime (both send `Authorization` header) but loses semantic precision for SDK generators.

**Suggestion:** Track swag v2 upstream for a `@securityDefinitions.bearer` annotation. When available, switch to emit `type: http, scheme: bearer`. Alternatively, a post-gen normalization step in `normalize-spec.sh` could patch the security scheme.

### F2 — P2: qrcodeMy @Failure annotations incomplete (warn-level)

**Diff-scope:** new (this PR's annotation set)

`modules/user/api.go:716-720` declares `@Failure 400` and `@Failure 401` only. The spectral rules `octo-auth-has-403`, `octo-auth-has-404`, `octo-auth-has-500` (all severity: warn) would flag the missing 403/404/500 responses.

The handler's error paths actually produce:
- `errcode.ErrUserQueryFailed` → DB error (semantic 500, wire 400 per D14)
- `errcode.ErrUserCurrentNotFound` → user not found (semantic 404, wire 400)
- `errcode.ErrUserQRVerCodeMissing` → missing vercode (semantic 400, wire 400)

**Impact:** Spec documentation incomplete — consumers cannot discover all possible error responses from the spec alone. Lint passes (warn-only).

**Suggestion:** Add `@Failure 404 {object} httperr.ErrorResp "user not found"` and `@Failure 500 {object} httperr.ErrorResp "internal error"` to complete the mandated set per R13.

### F3 — P2 (nit): ErrorResp.Details schema is opaque

**Diff-scope:** new

`httperr.ErrorResp.Error.Details` is `map[string]any`, producing `additionalProperties: {}` in the spec. This is technically correct but provides no schema guidance for consumers. The actual details shape is code-specific (e.g., `failed_index` for batch validation).

**Suggestion:** Acceptable for now — details are inherently polymorphic by design (R2: code-specific). Could add `x-examples` extension for common patterns in a follow-up.

## 3. Suggestions

1. **F1 fix path:** Monitor swag v2 for bearer support; consider a `normalize-spec.sh` patch in the interim.
2. **F2 fix path:** Add 404/500 `@Failure` lines to `qrcodeMy` godoc — trivial addition, improves spec completeness.
3. **CI activation:** When ready, copy `tools/octo-api/assets/templates/openapi-workflow.yml` to `.github/workflows/openapi.yml` and configure required status checks per `adoption.md`.

## 4. Additional Observations

- **Staged migration strategy is sound:** 1/443 coverage with coverage gate excluded from branch protection matches the adoption guide's brownfield path. No risk of blocking other PRs.
- **Spectral ruleset quality is high:** 30 rules covering R1–R13 with proper severity calibration (error for hard contracts, warn for style guidance). The `octo-response-success-shape` / `octo-response-error-shape` rules using `resolved: true` to follow `$ref` is the correct structural approach.
- **Shell scripts are well-engineered:** `check-prereqs.sh` provides excellent error messages with adoption-guide pointers. `check-swag-coverage.sh` correctly handles the awk state-machine for `@Router` detection.
- **CI workflow template uses SHA-pinned actions** (checkout@v4, setup-go@v5, etc.) — good supply-chain hygiene.

## 5. Data Flow Trace

### qrcodeMy handler (the only production code change)

```
c.GetLoginUID() → uid (from auth middleware context)
    ↓
u.db.QueryByUID(uid) → userModel (DB SELECT)
    ↓
userModel.QRVercode → string (DB field)
    ↓
config.QRCodeInfoURL template → strings.ReplaceAll(":code" → "vercode_<QRVercode>") → path
    ↓
config.External.BaseURL + "/" + path → URL string
    ↓
qrcodeMyResp{Data: URL} → c.Response() → c.JSON(200, resp) → json.Marshal
    ↓
Wire: {"data":"<baseURL>/<path>"}  ← IDENTICAL to old gin.H{"data": ...} output
```

**Conclusion:** Wire bytes are byte-identical. No client-visible change. The typed struct exists solely for swag annotation reference.

### ErrorResp (annotation-only, no runtime path)

```
Handler calls respondUserError(c, code) → ResponseErrorL() → respondL()
    ↓
c.RenderError(ErrorSpec{...}) → ErrorRenderer.Render()
    ↓
c.JSON(transportStatus, gin.H{"error": {...}, "msg": msg, "status": transportStatus})
    ↓
Wire: {"error":{"code":"...","message":"...","details":{},"http_status":N},"msg":"...","status":N}
```

**ErrorResp struct fields match renderer output:** ✅ (code↔code, message↔message, details↔details, http_status↔http_status, msg↔msg, status↔status)

`ErrorResp` is **never instantiated at runtime** — it exists only as a swag `@Failure` type reference. Zero risk of runtime behavior change.

### C1 — Dual-path parity
**Clear.** The only production code change (`qrcodeMy`) is a single handler's response type swap — no symmetric add/remove, subscribe/unsubscribe, or create/delete paths are touched. The toolchain files are all additive (new files). No dual-path divergence possible.

**Clear.** `ErrorResp` is annotation-only (never instantiated, never called in any control flow). `qrcodeMyResp` is used in exactly one call site (`qrcodeMy` handler line 738). No nested/compound calls, no shared sanitization functions affected. The shell scripts in `tools/octo-api/scripts/` are standalone utilities, not called from production code paths.

**Clear.** The `qrcodeMy` endpoint is already behind `AuthMiddleware` (verified in router setup: `user.GET("/qrcode", u.qrcodeMy)` inside the auth-protected group). The `@Security Bearer` annotation correctly documents this. No new endpoints are exposed — the PR only annotates an existing, already-protected handler. The toolchain adds no new HTTP endpoints or capabilities.

---

## Verdict

No P0 or P1 findings. All findings are P2 (documentation accuracy / spec completeness). The production code change is wire-format identical to the previous implementation. The toolchain is well-engineered with comprehensive lint rules and a sound staged-migration strategy.

**[Octo-Q] verdict: APPROVE** — P2 suggestions (bearer scheme type, complete @Failure set, details schema) are non-blocking improvements for follow-up.

Jerry-Xin (changes requested):
The PR is relevant to octo-server, but the vendored OpenAPI CI template is currently nonfunctional.

**🔴 Blocking**
- 🔴 Critical: `tools/octo-api/assets/templates/openapi-workflow.yml:133` calls `bash tools/octo-api/check-swag-coverage.sh modules`, but the script is committed at `tools/octo-api/scripts/check-swag-coverage.sh`. Any repo following `adoption.md` and copying this workflow will fail immediately in the coverage job with “No such file or directory”.
- 🔴 Critical: `tools/octo-api/assets/templates/openapi-workflow.yml:293` runs `make openapi-toolchain-test`, but `tools/octo-api/assets/openapi.mk:162` does not define that target. The optional workflow template’s self-test job is guaranteed to fail once enabled.

**💬 Non-blocking**
- 🟡 Warning: `modules/user/api.go:721-722` uses descriptive failure text instead of the R2 code names, producing the toolchain’s own warnings on the first annotated endpoint. This matches the staged D14/R2 gap described in the PR, so it is not blocking, but the showcase endpoint would be cleaner if the descriptions included known codes such as `AUTH_REQUIRED` / `INTERNAL_ERROR` or the current repo-specific equivalent.
- 🔵 Suggestion: The workflow template’s generate step at `tools/octo-api/assets/templates/openapi-workflow.yml:163-164` does not mirror `make openapi-gen` because it omits `--parseDependencyLevel 1`, `--outputTypes`, and `normalize-spec.sh`. Consider using `make openapi-verify` in CI so the local and CI generation paths stay identical.

**✅ Highlights**
- `make openapi-verify` passed and the committed `docs/openapi/swagger.yaml` is in sync.
- `make openapi-lint` passed with 0 errors.
- `go build ./pkg/httperr ./modules/user` and the targeted user/httperr tests passed.
- The `qrcodeMy` response change preserves the same top-level `{"data": ...}` wire shape.

yujiawei (changes requested):
# Code Review — PR #336 (octo-server)

## Summary

This PR adopts the OpenAPI/swag toolchain (brownfield path): it vendors `tools/octo-api/` (spectral ruleset + custom functions + shell scripts + a CI workflow template), wires swag global annotations into `main.go`, annotates the first endpoint (`GET /v1/user/qrcode`), adds `httperr.ErrorResp` for `@Failure` annotations, and commits the generated `docs/openapi/swagger.yaml` 3.1 baseline.

The **production code change is correct and wire-compatible**. The blocking concerns are all in the vendored tooling: the coverage gate is unsound by construction, and the CI workflow template — which `references/adoption.md` instructs adopters to copy verbatim into `.github/workflows/` — contains three path/target bugs that make every gate fail immediately on install.

## What's correct (verified)

- ✅ **`modules/user/api.go` — `qrcodeMy` is wire-compatible.** Old `gin.H{"data": url}` and new `qrcodeMyResp{Data: url}` both serialize to `{"data": url}`; `c.Response(data)` is just `c.JSON(http.StatusOK, data)` (octo-lib `pkg/wkhttp/http.go`). No client-visible change.
- ✅ **`pkg/httperr/respond.go` — `ErrorResp` matches the rendered envelope** (top-level `error{code,message,details,http_status}` plus the D14 `msg`/`status` siblings).
- ✅ **swag annotations are accurate.** `/v1/user/qrcode` is registered behind `AuthMiddleware` (`modules/user/api.go:216,225`), so the `@Failure 401` annotation reflects real behavior; `@ID`/`@Router`/`@BasePath /v1` → `servers:[{url:/v1}]` are consistent with the committed baseline.
- ✅ `go build` is green on all touched packages.
- ✅ `docs/openapi/index.html` is **not** tracked (it is git-ignored at `.gitignore:87`), so the throwaway preview is correctly excluded.

## Blocking — Changes Requested

### 🔴 P1 — Coverage gate is unsound: handlers are keyed by bare method name

`tools/octo-api/scripts/check-swag-coverage.sh:36-38` and `:60-65` reduce both the "all handlers" set and the "annotated" set to **bare method names** (`sed` strips the receiver, then `sort -u`), and compare them with `comm -23`.

This repo has heavy method-name collision across receivers/modules. Measured at this HEAD:

```
raw handler funcs: 492   unique method names: 444   collisions: 48
list   → 7 receivers (group/api.go:643, group/api_manager.go:60, space/api_manager.go:200,
                      message/api_manager.go:883, robot/api_manager.go:52,
                      user/api_manager.go:582, incomingwebhook/api.go:489)
delete → 5 receivers (message/api.go:1896, message/api_manager.go:117, robot/api_manager.go:89,
                      user/api_friend.go:174, incomingwebhook/api.go:590)
```

Consequences:
1. The "1/443" coverage figure is a name-count, not a handler-count — it undercounts true handlers by ~48.
2. **Latent false-negative:** because the annotated set is also keyed by bare name, annotating a single handler named `list` makes `comm -23` treat **all 7** `list` handlers as covered, silently hiding 6 unannotated endpoints. Not triggered today (only `qrcodeMy`, a unique name, is annotated), but the gate becomes unsound the moment any common-named handler is annotated — which defeats the gate's entire purpose.

**Fix:** key handlers on a stable identity (`receiver.method`, or `file:line`), not bare method name, on both sides of the comparison.

### 🔴 P1 — CI workflow template is broken-on-install (3 defects)

`references/adoption.md:22` documents adoption as `cp tools/octo-api/assets/templates/openapi-workflow.yml .github/workflows/openapi.yml`. Copied as documented, the pipeline fails before it validates anything:

- **`openapi-workflow.yml:133`** — `bash tools/octo-api/check-swag-coverage.sh modules`, but the script lives at `tools/octo-api/scripts/check-swag-coverage.sh` (missing `scripts/`). Gate 1 fails with "No such file or directory". (The same file uses the correct `scripts/` path at line 264 for `diff-openapi.sh`, and `openapi.mk:66` uses the correct path — so this is a copy/path-update miss.)
- **`openapi-workflow.yml:225`** — `--ruleset tools/octo-api/spectral.yaml`, but the ruleset is at `tools/octo-api/assets/spectral.yaml` (missing `assets/`). Gate 3 fails with "ruleset not found". (`openapi.mk:110` uses the correct path.)
- **`openapi-workflow.yml:293`** — `make openapi-toolchain-test`, but that target is **undefined**: `make -n openapi-toolchain-test` → "No rule to make target"; it is absent from the `.PHONY` list at `openapi.mk:162`. The four-stage self-test described in the comment block (and its `violations.yaml`/`valid.yaml` fixtures) was never vendored, so Gate 5 fails 100% of the time, and the `openapi.mk:7` comment claiming "fixtures in the skill package layout" is inaccurate.

These don't break CI in *this* repo today (the template is intentionally not installed), but the PR's stated purpose is to ship a working, copy-and-go toolchain — and as vendored it does not survive its own documented install. They are local (path fixes + one missing target/fixtures) and should land with this PR rather than ambush the first adopter.

## Non-blocking

- 🟡 **P2 — `npx ...@latest` on an actively-installed path.** `tools/octo-api/assets/openapi.mk:110` (`@stoplight/spectral-cli@latest`) and `:157` (`@redocly/cli@latest`) are reached via the root `Makefile`'s `include`, so `make openapi-lint`/`openapi-preview`/`openapi-check` fetch unpinned packages with `-y` auto-confirm. The same file pins the Go tools (`SWAG_VERSION`, `OASDIFF_VERSION`) — pin these two as well for consistency and supply-chain hygiene.
- 🟡 **P2 — Workflow gates are inert even after install** until an admin adds the 5 job names as required status checks (template self-documents this at lines 18-32). Worth a follow-up tracking issue so the baseline/ruleset can't silently drift.
- ⚪ **nit — Actions script-injection pattern.** `openapi-workflow.yml:261,264` interpolate `${{ github.event.pull_request.base.ref }}` directly into `run:` bodies. `base.ref` is GitHub-constrained (low risk), but prefer passing it via `env:` and referencing `"$BASE_REF"` so linters (actionlint/zizmor) stay clean before the template goes live.
- ⚪ **nit — `octo-description-distinct.js:13`** trailing-punctuation class `[.。!]` is inconsistent: it strips full-width `。` and ASCII `!` but not full-width `！`/`？` or ASCII `?`. A near-duplicate description ending in `！`/`?` escapes the R13 check. Suggest `/[.。!?！？]+$/`.
- ⚪ **nit — coverage script regex asymmetry.** The "all" set (`:36`) requires the exact single-param signature `(c *wkhttp.Context)`, while the annotated `awk` (`:48`) matches any `func (recv) name(`. A stray `@Router` on a multi-param helper would inflate the printed coverage % (it feeds `TOTAL_ANNOTATED` at `:64`/`:79`) without affecting the pass/fail set. Use the same handler definition on both sides.
- ⚪ **nit — `normalize-spec.sh:44-49`** YAML `externalDocs` deletion is exact-line/order/indent-sensitive, while the JSON branch (`:60-62`) is order-independent. Acceptable for a documented temporary swag-rc5 workaround, but the YAML branch is the more fragile of the two handling the same data.

## Security note (PR classified security-sensitive)

No new auth/credential/secret handling is introduced. `qrcodeMy` already embeds the user's `vercode` in the QR URL (pre-existing; invalidated on vercode reset) and the response shape is unchanged. The security-relevant items are the supply-chain `@latest` pins and the Actions interpolation pattern above — both worth a human glance, neither a live exposure at this HEAD.

## Verdict

**CHANGES_REQUESTED.** The production code is sound, but the toolchain this PR exists to ship has a correctness bug in its coverage gate and a CI template that fails on its own documented install. Both are local fixes that belong with this change.

lml2468 (changes requested):
🔴 CHANGES_REQUESTED — 原 CR 的核心 blocker（vendored CI 模板脚本路径错）在最新 commit ed8011bb 仍未修对。openapi 采用本身的代码（main.go 全局注解、httperr ErrorResp、端点标注）没问题。

## 🔴 blocking（仍在）：CI 模板脚本路径错，Gate 1 必挂
tools/octo-api/assets/templates/openapi-workflow.yml:133：
```
run: bash tools/octo-api/check-swag-coverage.sh modules
```
但脚本实际在 **tools/octo-api/scripts/check-swag-coverage.sh**（带 scripts/）。133 行缺了 `scripts/` → 文件不存在 → 这个 vendored workflow 一旦被下游仓库采用，Gate 1（注解覆盖检查）每次都会因 "No such file" 失败。

自证：同一文件 line 264 调另一个脚本用的是**正确**路径 `bash tools/octo-api/scripts/diff-openapi.sh`。所以 133 是漏改/笔误，与 264 不一致。仓库里 tools/octo-api/scripts/ 下确有 check-prereqs/check-swag-coverage/diff-openapi/normalize-spec 四个脚本，路径前缀应统一为 `tools/octo-api/scripts/`。

修复：133 行改为 `bash tools/octo-api/scripts/check-swag-coverage.sh modules`。

这正是上一轮 CR 指出的同一 blocker（vendored OpenAPI CI 模板不可用），本次 commit 未修对——建议修掉这一行即可解。

## openapi 采用代码 OK（无 blocking）
- main.go：func main() 上方加全局 swag 注解块（@title/@version/@BasePath /v1/@securityDefinitions Bearer 等），纯注解、无行为影响。
- pkg/httperr/respond.go：新增 ErrorResp 类型，镜像注入的 ErrorRenderer 实际 wire 形状（top-level error + 兼容 msg/status），供 @Failure 注解引用，使生成的 OpenAPI schema 与 handler 真实输出一致。纯类型、无行为改动。
- common/user api 的端点标注、docs/openapi/swagger.yaml baseline、spectral 规则/自定义函数/脚本：openapi 工具链 vendoring，逻辑无害。

## 说明
本 PR 的 base 分支 feat/agent-runtime 自上次 review 推进了大量(78)无关 commit(runtime 模块迁移、dmwork→octo 改名等)，但 #336 自身的净 PR diff 仍聚焦 openapi 采用。我只审 #336 的净变更。核心就差 workflow:133 这一行路径。

修掉 133 行路径即可转 APPROVE。

—— Octo-PR (lml2468)

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#336 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/336
**Head SHA:** `ed8011bb3ee558043e2922149ed8f74cc9bd08c1`
**Base:** `feat/agent-runtime` ← `feat/openapi-toolchain`

---

## 1. 验证结论

| 项 | 结论 | 证据 |
|---|---|---|
| Wire-format parity (`gin.H` → `qrcodeMyResp`) | ✅ | `c.Response()` = `c.JSON(200, data)` (octo-lib `wkhttp/http.go:130`); `gin.H{"data": url}` and `qrcodeMyResp{Data: url}` both serialize to `{"data":"<url>"}` — byte-identical |
| `ErrorResp` mirrors production wire shape | ✅ | Field-by-field match vs `pkg/i18n/renderer.go:57-66` `gin.H` output: `error.{code,message,details,http_status}` + `msg` + `status` all present and correctly typed |
| swag global annotations → generated spec | ✅ | `@BasePath /v1` → `servers:[{url:/v1}]`; `@Router /user/qrcode [get]` → `paths./user/qrcode` (no double `/v1` prefix); Bearer scheme correctly declared as `apiKey` in header |
| Toolchain vendoring completeness | ✅ | `tools/octo-api/` contains SKILL.md, spectral.yaml (30 rules), 3 JS functions, 4 shell scripts, openapi.mk, templates; `Makefile` include resolves correctly |
| `.gitignore` update | ✅ | `docs/openapi/index.html` added — covers `make openapi-preview` throwaway output |
| `countriesList` comment-only change | ✅ | No code change; comment explains R1 violation + breaking-change rationale for deferring annotation |
| Merge commit (agent-runtime sync) | ✅ | Clean auto-merge on `main.go`, `modules/{common,user}/api.go`; `.gitignore` conflict resolved correctly (kept both sides) |

## 2. 发现的问题（P0/P1/P2）

### P2-1: `openapi-check` summary text says "4 道闸" but runs 3 gates

**Severity:** P2 (cosmetic / documentation)

`openapi.mk` `openapi-check` target header comment says "一键 4 道闸" but the implementation runs 3 gates: coverage → verify → lint. The `openapi-help` target also lists 8 separate commands without mentioning a "4-gate" aggregate.

**File:** `tools/octo-api/assets/openapi.mk`, `openapi-check` target
**Suggestion:** Align the comment to "3 道闸" or add a 4th gate (e.g., `openapi-diff` for PRs).

### P2-2: `qrcodeMy` annotations missing 403/404/500 responses

**Severity:** P2 (spectral warn, not error; acknowledged as staged migration)

The showcase endpoint declares `@Failure 400` and `@Failure 401` only. Spectral rules `octo-auth-has-404` and `octo-auth-has-500` (both `warn`) would flag this. The handler code does have paths that map to 404 (`ErrUserCurrentNotFound`, `ErrUserQRVerCodeMissing`) and implicitly 500 (`ErrUserQueryFailed`), but these are not surfaced in the annotations.

**File:** `modules/user/api.go:789-793`
**Suggestion:** Add `@Failure 404` and `@Failure 500` annotations to the showcase endpoint to demonstrate the complete pattern. This is not blocking since spectral warns (not errors) and the PR body explicitly scopes this to staged migration.

### P2-3: `go install` without integrity verification

**Severity:** P2 (supply-chain hygiene, dev-only blast radius)

`openapi-install` and `oasdiff-install` targets run `go install` without checksum verification. Versions are pinned (`SWAG_VERSION ?= v2.0.0-rc5`, `OASDIFF_VERSION ?= v1.18.5`), which provides reproducibility but not integrity. A compromised upstream Go module proxy could serve different bytes for the same version tag.

**File:** `tools/octo-api/assets/openapi.mk`, lines ~60-65 and ~115-120
**Mitigation:** Dev-only toolchain; not deployed to production. Goproxy checksum database (sum.golang.org) provides some protection. Acceptable for now; consider `go install ...@<hash>` or pre-built binaries with sha256 verification if the toolchain is later used in CI.

## 3. 建议

1. **Fix "4 道闸" → "3 道闸"** in `openapi-check` comment (trivial, one-line fix).
2. **Add 404/500 `@Failure` annotations** to the showcase endpoint to demonstrate the complete error-annotation pattern — this is the endpoint that future contributors will copy.
3. **Consider adding `go install` checksum verification** when the toolchain moves to CI (not blocking for dev-only use).

## 4. 额外发现

- **`normalize-spec.sh` is a documented workaround** for swag v2.0.0-rc5 emitting empty `externalDocs`. The script is scoped to the generated spec file only (line-based YAML deletion), safe. When swag fixes upstream, the normalization step should be removed (already noted in the script's header comment).
- **`check-swag-coverage.sh` handler detection regex** uses `[a-z]+` for the receiver variable, which matches the project convention (lowercase receiver names like `u`, `cn`, `ba`). No false-negative risk for this codebase.
- **The CI workflow template** (`openapi-workflow.yml`) is correctly NOT activated — it's a template for future opt-in. The PR body confirms coverage gate is intentionally not a required check during staged migration.

## 5. 数据流回溯

### `qrcodeMy` handler (the only runtime-impacting change)

```
c.GetLoginUID() → uid (string, from auth middleware)
    ↓
u.db.QueryByUID(uid) → userModel (*UserModel, from DB)
    ↓
userModel.QRVercode → vercode (string, DB column)
    ↓
strings.ReplaceAll(config.QRCodeInfoURL, ":code", "vercode_"+vercode) → path
    ↓
fmt.Sprintf("%s/%s", config.External.BaseURL, path) → full URL string
    ↓
qrcodeMyResp{Data: url} → c.Response() → c.JSON(200, resp) → wire: {"data":"<url>"}
```

**Pre-PR wire:** `gin.H{"data": url}` → `{"data":"<url>"}` ✅ identical
**Post-PR wire:** `qrcodeMyResp{Data: url}` → `{"data":"<url>"}` ✅ identical

No data-flow change. No new error paths. No gate/short-circuit modifications.

### `ErrorResp` struct (annotation-only, never instantiated at runtime)

```
pkg/i18n/renderer.go:57-66 → c.JSON(transportStatus, gin.H{
    "error": gin.H{"code", "message", "details", "http_status"},
    "msg": msg, "status": transportStatus
})
```

`ErrorResp` struct fields match 1:1. Used only in `@Failure` swag annotations → consumed by swag CLI at code-gen time → never instantiated during request handling.

## 6. 三类盲点 checklist（R5 — security_sensitive PR 必填）

- **C1 双路径 parity:** N/A — This PR does not add/remove/modify any symmetric path pairs (subscribe/unsubscribe, create/delete, etc.). The only handler change (`qrcodeMy`) is a type swap on the success response path; error paths are untouched.
- **C2 control-flow ordering / 嵌套复用:** N/A — No function is reused in a different ordering context. `normalize-spec.sh` runs once post-gen. Spectral JS functions are stateless per-invocation. No nested/compound calls.
- **C3 授权边界 ≠ 能力边界:** N/A — No new endpoints, tools, or capabilities exposed. `qrcodeMy` retains existing Bearer auth + `c.GetLoginUID()`. Toolchain scripts run in developer local environments only, not exposed as endpoints.
- **C4 授权生命周期 / 容器-成员状态级联:** N/A — No authorization logic changes. No container/member state interactions.

## 7. 跨轮 blocker 复检（R6）

N/A — First review of this PR.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The PR is well-structured: wire-format parity is preserved (verified via data-flow tracing to `c.Response()` → `c.JSON(200, data)`), `ErrorResp` accurately mirrors the production error renderer output, and the toolchain vendoring is complete with proper documentation. Three P2 findings are cosmetic/hygiene items that don't block landing.

Jerry-Xin (changes requested):
## REQUEST_CHANGES — Gate 1 path bug still present at `ed8011bb`

The vendored CI template `tools/octo-api/assets/templates/openapi-workflow.yml` still calls the coverage script at the wrong path.

🔴 **Blocking — `openapi-workflow.yml:133`**
```yaml
run: bash tools/octo-api/check-swag-coverage.sh modules
```
But the script actually lives at `tools/octo-api/scripts/check-swag-coverage.sh` (missing `scripts/` prefix). At runtime this is a file-not-found → **Gate 1 (annotation coverage) fails**.

This self-contradicts line 264 in the same file, which uses the correct `scripts/` prefix:
```yaml
bash tools/octo-api/scripts/diff-openapi.sh "origin/${{ github.event.pull_request.base.ref }}" ...
```

**Fix:** change line 133 to `bash tools/octo-api/scripts/check-swag-coverage.sh modules`.

Confirmed at head `ed8011bb` (ref-locked): `tools/octo-api/scripts/` contains `check-prereqs.sh`, `check-swag-coverage.sh`, `diff-openapi.sh`, `normalize-spec.sh`. There is no `tools/octo-api/check-swag-coverage.sh`.

(Agrees with @lml2468's CHANGES_REQUESTED; the path was not yet fixed.)

✅ **Otherwise solid** — the Go changes (main.go global swag block, httperr `ErrorResp` mirroring the wire shape, endpoint annotations) are non-behavioral and safe. Only the one-line workflow path blocks merge.

yujiawei (changes requested):
# Code Review — PR #336 (octo-server)

Reviewed against head `e6dfc46cf1f56943f05d394fcef87565b21745e0` (the PR was rebased from the original `ed8011bb` merge commit; the substantive diff for all code files is byte-identical, only base context shifted). Base branch `feat/agent-runtime`. Build verified, spec regenerated and diffed, error envelope traced end-to-end into `octo-lib`.

## Verdict: Changes Requested

One blocking issue: the generated OpenAPI security scheme documents the **wrong authentication header**. Everything else (wire compatibility, error envelope fidelity, build, generated baseline) is correct and well-executed.

---

## 🔴 Blocker — Security scheme documents `Authorization`, but the server reads the `token` header

`main.go:63-65`:

```go
// @securityDefinitions.apikey Bearer
// @in               header
// @name             Authorization
```

→ `docs/openapi/swagger.yaml:28-31`:

```yaml
securitySchemes:
  Bearer:
    in: header
    name: Authorization
    type: apiKey
```

But the auth middleware guarding the one documented endpoint reads a **different** header. `GET /v1/user/qrcode` is mounted under `AuthMiddleware` (`modules/user/api.go:225`, group at line ~219), which is `octo-lib`'s `wkhttp.AuthMiddleware`. In the pinned `octo-lib` (`go.mod`, no `replace`), that middleware does:

```go
// pkg/wkhttp/http.go:307
token := c.GetHeader("token")
if token == "" { /* 401 err.shared.auth.token_missing */ }
```

The header is read as `token`, unconditionally, **before** any `TokenParser` runs — there is no `Authorization` fallback anywhere in the middleware. So a consumer who follows the generated spec and sends `Authorization: <token>` receives a 401; the working header is `token`.

**Why this blocks merge** (not just a doc-nit): this PR's entire stated purpose is to land a *correct, trustworthy* OpenAPI baseline that every future annotated endpoint inherits. The auth scheme is the single most actionable contract field, this is the only documented endpoint, and the value was copied from the toolchain template (`tools/octo-api/references/api-spec.md`, `@name Authorization`) without checking octo-server's actual middleware. Shipping the foundational auth descriptor wrong propagates to every endpoint added later. The fix is one line.

**Fix:** change `main.go` to `// @name token`, and consider renaming the scheme away from `Bearer` (it is a raw token in a custom header, not RFC 6750 bearer), then regenerate (`make openapi-gen`).

> Please also have a human confirm the canonical client auth-header name (`token` vs any gateway-level `Authorization` translation) before regenerating, since this is the security-baseline contract.

---

## ✅ Verified correct

- **Wire compatibility (no client breakage).** `modules/user/api.go` switches `c.Response(gin.H{"data": url})` → `c.Response(qrcodeMyResp{Data: url})`. `wkhttp.Context.Response` is `c.JSON(200, data)`; a one-key `gin.H{"data":v}` and a one-field struct with `json:"data"` marshal to byte-identical `{"data":"..."}`. No wire drift.
- **`@Failure` annotations match runtime.** The three business error paths (`ErrUserQueryFailed` / `ErrUserCurrentNotFound` / `ErrUserQRVerCodeMissing`) all route through `respondUserError` → `httperr.ResponseErrorL`, which pins transport status to 400 (D14) — matches `@Failure 400`. Auth failures from the middleware render at 401 — matches `@Failure 401`.
- **`httperr.ErrorResp` mirrors the real envelope.** It matches the injected `pkg/i18n/renderer.go` output field-for-field (`error.{code,message,details,http_status}` + sibling `msg`/`status`). Note `ErrorResp` is a doc-only DTO referenced only by swag annotations — never instantiated at runtime, so it cannot leak fields.
- **Generated baseline is genuine.** Regenerated with swag v2.0.0 + `normalize-spec.sh`: byte-identical to the committed `swagger.yaml`. `servers:[{url:/v1}]` derives correctly from `@BasePath /v1`; only `/user/qrcode` present; `openapi: 3.1.0` genuine. Spectral lint of the committed spec: **0 errors, 5 advisory warnings** (missing 403/404/500, error-description naming) — non-blocking by design.
- **Build green.** `go build ./pkg/httperr/... ./modules/user/... ./modules/common/... .` exits 0.
- **`countriesList` deliberately left un-annotated** — bare-array response violates R1; reshaping is breaking. Correctly deferred to the coverage gate (which stays advisory at 1/~460).
- **Supply-chain (security-sensitive classification).** No `curl|sh` pipes; the 3 spectral JS functions are pure (no `child_process`/`eval`/`fetch`/`process.env`); scripts use `set -euo pipefail`. The CI workflow template is **inert** — not installed in `.github/workflows/`, no CI job calls any `openapi-*` target, Dockerfile does not invoke `make`. So nothing in the toolchain executes in CI today beyond Makefile-include parsing.

---

## 🟡 Non-blocking (P2 — address before promoting the toolchain to a required gate)

1. **Unpinned remote executables** in manual targets: `openapi.mk:110` (`npx @stoplight/spectral-cli@latest`) and `:157` (`npx @redocly/cli@latest`); the workflow template's `oasdiff @latest`. `swag` and `oasdiff` *in openapi.mk* are correctly pinned. Pin the npx tools before any of these run in required CI. Low blast radius today (manual-only).
2. **Vendored CI template would fail if activated as-is.** `tools/octo-api/assets/templates/openapi-workflow.yml`: line 133 calls `tools/octo-api/check-swag-coverage.sh` (actual path: `tools/octo-api/scripts/...`), line ~226 references `tools/octo-api/spectral.yaml` (actual: `assets/spectral.yaml`), and line 293 calls `make openapi-toolchain-test` (no such target). Inert today (template not installed), but a downstream copy per its own header would break on day one. Fix the paths and add/drop the self-test target.
3. **Hard `include tools/octo-api/assets/openapi.mk` in `Makefile:84`** couples *all* make targets (including the CI i18n gates) to that file existing. Consider `-include` to decouple the core pipeline from the optional toolchain.
4. **`normalize-spec.sh` silently `exit 0` when `python3` is absent**, which can reintroduce the empty-`externalDocs` block and produce a confusing false-positive drift on `make openapi-verify`. Consider failing hard or doing the strip in sed/awk.
5. **Generated `httperr.ErrorResp` schema marks no fields `required`** and `details` is loose (`additionalProperties: {}`), under-specifying what the server always emits. Pure doc-fidelity; tighten later with struct tags/overrides.

---

## Coverage / blind spots

- Did **not** run the endpoint against a live server/DB — conclusions are from source tracing + `go build` + spec regeneration + spectral run.
- Branch-protection / required-status-check config is not verifiable from a local checkout; the "coverage gate not required" claim is inferred from the workflow being absent.
- Vendored `tools/octo-api/**` (SKILL.md, references/*.md, PR_TEMPLATE.md) reviewed only where it intersects the security / spec-consistency surface, not line-by-line.
- Two independent advisory passes were run; one cross-family pass flagged the spectral auth-rules (`octo-auth-has-403/404/500`) as matching nothing — **disproven empirically**: running spectral on the committed spec, those rules *do* fire on `/user/qrcode` (0 errors, the 403/404/500 warnings appear), so the jsonpath is correct.

lml2468:
补充我的 CR：核实了 yujiawei 提的第二个 blocker（OpenAPI 认证头名错误）——属实，我和 Steve 之前都漏了这条，一并纳入。

## 🔴 blocking #2（yujiawei 抓到，我独立核实属实）：swag securityDefinitions 认证头名错误
main.go 全局 swag 注解声明：
```
@securityDefinitions.apikey Bearer
@in               header
@name             Authorization   ← 错
```
但 octo 的 AuthMiddleware 实际读的是 **token** 头，不是 Authorization：
- octo-lib pkg/wkhttp/http.go:271/307：`token := c.GetHeader("token")`。
- 测试惯例压倒性印证：octo-server 里 `Header.Set("token", ...)` 用了 486 次，`Header.Set("Authorization", ...)` 仅 23 次（那 23 次是 bot_provision 的 daemon Bearer 专用路径，另一套机制，非普通用户端点）。

后果:这是 OpenAPI baseline PR，生成的文档会告诉所有 API 客户端用 `Authorization` 头鉴权,但后端实际读 `token` 头 → 照文档调用的客户端会 401。文档与实现不符,对一个"建立 openapi 契约真相源"的 PR 是实质 bug。

修法:@name 改为 token（`@securityDefinitions.apikey ApiKeyAuth` + `@in header` + `@name token`),与 AuthMiddleware 实际读取的头对齐。daemon Bearer 那套若也要进 spec,单列一个 securityDefinition。

## blocking 清单（合并前都要修）
1. 🔴 openapi-workflow.yml:133 脚本路径缺 scripts/(我和 Jerry-Xin 已指出)。
2. 🔴 swag 认证头名 Authorization 应为 token(yujiawei 指出,本条我已独立核实属实)。

两条都修掉才能转 APPROVE。Go 改动本身(全局 swag 块结构、httperr ErrorResp、端点注释)非行为变更、安全;问题都在"声明与实际不符"——路径声明错 + 认证头声明错。

诚实说:我和 Steve 之前都只盯了 CI 脚本路径,漏了认证头名这条——审 openapi/契约类 PR 该核"声明的 auth 方案是否匹配实际认证头",我没核到,yujiawei 补上了。

—— Octo-PR (lml2468)

PR #337, opened 2026-06-10
feat(bot_api): resolve channel target (group/thread) by name

### LINKED ISSUE #333: Add bot-token endpoint to resolve a channel target (group/thread) by name
Part of the group/thread routing disambiguation effort (plugin side tracked in openclaw-channel-octo#98).

## Problem
When a user asks to forward something to a channel named "XXX", "XXX" may be **either a group name or a thread name**, and there is currently no backend API that lets a bot resolve a target channel by name. The bot has to enumerate groups and then list threads per group (N+1), and there is no way to faithfully surface "same name, multiple candidates" for disambiguation.

## Current state (verified)
- `GET /v1/bot/groups` (`modules/bot_api/groups.go`) lists the bot's groups by membership only, no name filter.
- `GET /v1/bot/groups/:group_no/threads` (`modules/bot_api/threads.go`) lists threads of one parent group only, no name filter.
- There is no bot-token API that searches groups/threads by name.
- `modules/bot_api/groups.go` `botSpaceMembers` already provides a `name LIKE` + wildcard-escaping pattern to follow.

## Proposed endpoint (minimal, bot-token auth)
Route (in the `botAPI` group, next to `getGroups`):
```
botAPI.GET("/resolve/targets", ba.botResolveTargets)
```

Request (query):
- `name`  string, required, trimmed; empty => `respondBotAPIRequestInvalid("name")`
- `limit` int, optional, default 20, max 50 (endpoint-specific; parse with `strconv.Atoi`)
- `kind`  string, optional `group|thread|all`, default `all`

No `space_id` param: the bot's visibility boundary is membership (`group_member.uid = robotID`), not space. A bot can be pulled into cross-space external groups; filtering by space would wrongly exclude those. Membership scoping alone is both correct and safe.

Response (snake_case, unified candidate array, distinguished by `kind`):
- group candidate: `{ kind:"group", channel_id, channel_type:2, name, group_no }`
- thread candidate: `{ kind:"thread", channel_id:"<group_no>____<short_id>", channel_type:5, name, group_no, short_id, parent_name }`
- plus `total` (= number of returned candidates) and `truncated` (= any candidate class hit the limit).

Query logic:
- Group hits (bot's groups only): `WHERE gm.uid=? AND gm.is_deleted=0 AND g.name LIKE ?`, exact-match-first stable ordering (`ORDER BY (g.name=?) DESC, g.group_no ASC`), `LIMIT limit+1` to detect truncation.
- Thread hits (threads under the bot's parent groups): `WHERE gm.uid=? AND gm.is_deleted=0 AND t.status IN (1,2) AND t.name LIKE ?` (active + archived, exclude deleted, matching the system's GetThreads/sanitizeListStatuses contract), stable ordering, `LIMIT limit+1`.
- Each class takes `limit+1` independently so one class cannot starve the other; truncation computed per class.
- App Bot: align with `getGroups` (read endpoint) and return empty candidates rather than 403.
- DB errors => existing `ErrBotAPIQueryFailed`. No new error codes; run `make i18n-extract-check` + `make i18n-lint`.

The backend never guesses or picks a winner: same-name groups and threads are all returned; the plugin handles disambiguation.

## Tests
New `modules/bot_api/resolve_targets_test.go`: same-name group+thread, multiple same-name threads across parents, non-member not visible, cross-space external group IS visible, `kind` filters, limit clamp/truncated, invalid/empty name.

## Scope
- Repo: octo-server (backend only)
- Corresponds to PR-2 (can land in parallel with the plugin Scenario 1 PR)

jeff-wilson2010:
## Summary

Adds `GET /v1/bot/resolve/targets`, a bot-token endpoint that searches the calling bot's visible groups and threads by name and returns a unified candidate array. This lets the adapter disambiguate a named forward target ("forward to 'XXX'") — where XXX may be a group or a thread — instead of guessing.

Closes #333.

## Behavior

- Visibility is defined purely by membership (`group_member.uid = robotID AND is_deleted = 0`). No `space_id` filter, so a bot pulled into a cross-space external group can still resolve that group and its threads.
- Same-named group and thread both come back as distinct candidates, disambiguated by `kind` (`group` / `thread`). Multiple same-named candidates are returned faithfully; the endpoint never guesses.
- `kind` filter accepts `all` (default) / `group` / `thread`.
- `status IN (1, 2)` keeps active + archived threads and excludes deleted, matching the system `GetThreads` / `sanitizeListStatuses` contract.
- `LIMIT limit+1` probes for truncation; default limit 20, capped at 50. Exact (un-wildcarded) name match is ordered first, followed by a stable ordering.
- LIKE wildcards in user input are escaped.
- App Bots (no `group_member` rows) naturally resolve to an empty set rather than being rejected, following the read-only `getGroups` precedent.

## Response shape

```json
{
  "candidates": [
    { "kind": "group",  "channel_id": "<group_no>", "channel_type": 2, "name": "...", "group_no": "..." },
    { "kind": "thread", "channel_id": "<group_no>____<short_id>", "channel_type": 5, "name": "...", "group_no": "...", "short_id": "...", "parent_name": "..." }
  ],
  "total": 2,
  "truncated": false
}
```

## Notes

- The thread query joins `thread` (`utf8mb4_general_ci`) with `group` / `group_member` (`utf8mb4_0900_ai_ci`). The cross-collation column-to-column join equalities are pinned with an explicit `COLLATE` to avoid MySQL 8.0 `Error 1267 (Illegal mix of collations)`. The group query touches no mixed-collation columns and is left untouched.

## Testing

- New integration tests in `modules/bot_api/resolve_targets_test.go` (13 subtests, run against a real MySQL 8.0): same-name group+thread, multiple same-name threads across groups, non-member not visible, cross-space external group, kind filter, invalid kind, exact-match-first ordering, limit clamp + truncation, group-full-still-returns-thread, wildcard escaping, empty name, archived-visible/deleted-hidden, app-bot-empty. All pass.
- `make i18n-extract-check` and `make i18n-lint` pass.

### DIFF
--- a/modules/bot_api/api_i18n_test.go
+++ b/modules/bot_api/api_i18n_test.go
@@ -36,7 +36,7 @@ func TestBotAPINoLegacyResponseError(t *testing.T) {
 	files := []string{
 		"auth.go", "register.go", "commands.go", "events.go", "mention_pref.go",
 		"sync.go", "typing.go", "send.go", "threads.go", "file.go", "groups.go",
-		"voice_adapter.go", "obo_api.go",
+		"voice_adapter.go", "obo_api.go", "resolve_targets.go",
 	}
 	banned := []string{
 		".ResponseError(",
--- a/modules/bot_api/bot_api.go
+++ b/modules/bot_api/bot_api.go
@@ -214,6 +214,7 @@ func (ba *BotAPI) Route(r *wkhttp.WKHttp) {
 		botAPI.POST("/heartbeat", ba.heartbeat)
 		botAPI.POST("/messages/sync", ba.syncMessages)
 		botAPI.GET("/groups", ba.getGroups)
+		botAPI.GET("/resolve/targets", ba.botResolveTargets)
 		botAPI.GET("/groups/:group_no", ba.getGroupInfo)
 		botAPI.GET("/groups/:group_no/members", ba.getGroupMembers)
 		botAPI.GET("/groups/:group_no/mention_pref", ba.getMentionPref) // 群级免@偏好读（octo-server#237）
--- a/modules/bot_api/resolve_targets.go
+++ b/modules/bot_api/resolve_targets.go
@@ -1,0 +1,198 @@
+package bot_api
+
+import (
+	"net/http"
+	"strconv"
+	"strings"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/modules/thread"
+	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
+	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
+	"go.uber.org/zap"
+)
+
+// resolve/targets limit semantics — intentionally distinct from botSpaceMembers
+// (default 50 / max 200). This endpoint defaults to 20 and caps at 50.
+const (
+	resolveTargetsDefaultLimit = 20
+	resolveTargetsMaxLimit     = 50
+)
+
+// resolveTargetKinds enumerates the accepted `kind` filter values.
+const (
+	resolveKindAll    = "all"
+	resolveKindGroup  = "group"
+	resolveKindThread = "thread"
+)
+
+// resolveTargetCandidate is one match returned by botResolveTargets. Groups and
+// threads share this shape and are distinguished by `kind`; absent fields are
+// omitted (a group has no short_id / parent_name). Channel type is 2 for a
+// group and 5 for a thread (community topic).
+type resolveTargetCandidate struct {
+	Kind        string `json:"kind"`
+	ChannelID   string `json:"channel_id"`
+	ChannelType uint8  `json:"channel_type"`
+	Name        string `json:"name"`
+	GroupNo     string `json:"group_no"`
+	ShortID     string `json:"short_id,omitempty"`
+	ParentName  string `json:"parent_name,omitempty"`
+}
+
+// botResolveTargets handles GET /v1/bot/resolve/targets.
+//
+// It searches the calling bot's visible groups and threads by name and returns
+// a unified candidate array so the adapter can disambiguate a named target
+// ("forward to 'XXX'") instead of guessing. Visibility is defined purely by
+// membership (group_member.uid=robotID AND is_deleted=0) — there is no space_id
+// filter, so a bot pulled into a cross-space external group can still resolve
+// that group and its threads. App Bots have no group_member rows, so (like the
+// read-only getGroups precedent) they naturally resolve to an empty set rather
+// than being rejected.
+func (ba *BotAPI) botResolveTargets(c *wkhttp.Context) {
+	robotID := getRobotIDFromContext(c)
+	if robotID == "" {
+		ba.respondBotAPIIdentityMissing(c)
+		return
+	}
+
+	name := strings.TrimSpace(c.Query("name"))
+	if name == "" {
+		respondBotAPIRequestInvalid(c, "name")
+		return
+	}
+
+	kind := strings.TrimSpace(c.Query("kind"))
+	if kind == "" {
+		kind = resolveKindAll
+	}
+	if kind != resolveKindAll && kind != resolveKindGroup && kind != resolveKindThread {
+		respondBotAPIRequestInvalid(c, "kind")
+		return
+	}
+
+	// Explicit parse + clamp. Unlike botSpaceMembers (fmt.Sscanf), we validate
+	// with strconv.Atoi and degrade leniently: a missing / unparseable / <=0
+	// value keeps the default rather than erroring.
+	limit := resolveTargetsDefaultLimit
+	if s := strings.TrimSpace(c.Query("limit")); s != "" {
+		if v, err := strconv.Atoi(s); err == nil && v > 0 {
+			limit = v
+		}
+	}
+	if limit > resolveTargetsMaxLimit {
+		limit = resolveTargetsMaxLimit
+	}
+
+	// Escape LIKE wildcards in user input (same escaping as botSpaceMembers).
+	escaped := strings.ReplaceAll(name, "\\", "\\\\")
+	escaped = strings.ReplaceAll(escaped, "%", "\\%")
+	escaped = strings.ReplaceAll(escaped, "_", "\\_")
+	pattern := "%" + escaped + "%"
+
+	candidates := make([]resolveTargetCandidate, 0)
+	truncated := false
+
+	if kind == resolveKindAll || kind == resolveKindGroup {
+		type groupRow struct {
+			GroupNo string `db:"group_no"`
+			Name    string `db:"name"`
+		}
+		var rows []groupRow
+		// Only the bot's own groups (group_member.uid=robotID AND is_deleted=0);
+		// naturally includes cross-space external groups, no space filter.
+		// LIMIT limit+1 probes for truncation. ORDER BY (g.name = ?) DESC puts an
+		// exact (un-wildcarded) name match first, then a stable group_no order.
+		_, err := ba.ctx.DB().SelectBySql(
+			"SELECT g.group_no, g.name FROM group_member gm INNER JOIN `group` g ON gm.group_no = g.group_no "+
+				"WHERE gm.uid = ? AND gm.is_deleted = 0 AND g.name LIKE ? "+
+				"ORDER BY (g.name = ?) DESC, g.group_no ASC LIMIT ?",
+			robotID, pattern, name, limit+1,
+		).Load(&rows)
+		if err != nil {
+			ba.Error("resolve targets: query groups failed", zap.Error(err), zap.String("robotID", robotID))
+			httperr.ResponseErrorL(c, errcode.ErrBotAPIQueryFailed, nil, nil)
+			return
+		}
+		if len(rows) > limit {
+			truncated = true
+			rows = rows[:limit]
+		}
+		for _, r := range rows {
+			candidates = append(candidates, resolveTargetCandidate{
+				Kind:        resolveKindGroup,
+				ChannelID:   r.GroupNo,
+				ChannelType: common.ChannelTypeGroup.Uint8(),
+				Name:        r.Name,
+				GroupNo:     r.GroupNo,
+			})
+		}
+	}
+
+	if kind == resolveKindAll || kind == resolveKindThread {
+		type threadRow struct {
+			GroupNo    string `db:"group_no"`
+			ShortID    string `db:"short_id"`
+			Name       string `db:"name"`
+			ParentName string `db:"parent_name"`
+		}
+		var rows []threadRow
+		// Only threads under groups the bot belongs to. status IN (1,2) keeps
+		// active + archived and excludes deleted(3), matching the system
+		// GetThreads / sanitizeListStatuses contract. LIMIT limit+1 probes for
+		// truncation; exact name match first, then a stable ordering.
+		//
+		// COLLATE utf8mb4_general_ci is pinned here because thread.* is
+		// utf8mb4_general_ci while `group`/group_member.* are utf8mb4_0900_ai_ci.
+		// Two distinct reasons, do not conflate them:
+		//   1. REQUIRED — the JOIN ON column-to-column equalities
+		//      (gm.group_no = t.group_no, g.group_no = t.group_no) compare two
+		//      columns of differing collation with no literal to coerce them, so
+		//      MySQL 8.0 raises Error 1267 (Illegal mix of collations). COLLATE
+		//      on these is the actual fix.
+		//   2. DEFENSIVE — the ORDER BY exact-match check (t.name = ?) and the
+		//      t.name LIKE ? filter compare a column against a parameter/literal,
+		//      whose collation is coercible, so they normally do NOT raise 1267.
+		//      COLLATE there only pins the sort/compare semantics to one explicit
+		//      collation so ranking stays consistent.
+		// general_ci is chosen as the common collation; only the thread query
+		// needs this — the group query above touches no general_ci columns and is
+		// left untouched.
+		_, err := ba.ctx.DB().SelectBySql(
+			"SELECT t.group_no, t.short_id, t.name, g.name AS parent_name FROM thread t "+
+				"INNER JOIN group_member gm ON gm.group_no = t.group_no COLLATE utf8mb4_general_ci AND gm.uid = ? AND gm.is_deleted = 0 "+
+				"INNER JOIN `group` g ON g.group_no = t.group_no COLLATE utf8mb4_general_ci "+
+				"WHERE t.status IN (1, 2) AND t.name LIKE ? "+
+				"ORDER BY (t.name = ? COLLATE utf8mb4_general_ci) DESC, t.status ASC, t.group_no ASC, t.short_id ASC LIMIT ?",
+			robotID, pattern, name, limit+1,
+		).Load(&rows)
+		if err != nil {
+			ba.Error("resolve targets: query threads failed", zap.Error(err), zap.String("robotID", robotID))
+			httperr.ResponseErrorL(c, errcode.ErrBotAPIQueryFailed, nil, nil)
+			return
+		}
+		if len(rows) > limit {
+			truncated = true
+			rows = rows[:limit]
+		}
+		for _, r := range rows {
+			candidates = append(candidates, resolveTargetCandidate{
+				Kind:        resolveKindThread,
+				ChannelID:   thread.BuildChannelID(r.GroupNo, r.ShortID),
+				ChannelType: common.ChannelTypeCommunityTopic.Uint8(),
+				Name:        r.Name,
+				GroupNo:     r.GroupNo,
+				ShortID:     r.ShortID,
+				ParentName:  r.ParentName,
+			})
+		}
+	}
+
+	c.JSON(http.StatusOK, map[string]interface{}{
+		"candidates": candidates,
+		"total":      len(candidates),
+		"truncated":  truncated,
+	})
+}
--- a/modules/bot_api/resolve_targets_test.go
+++ b/modules/bot_api/resolve_targets_test.go
@@ -1,0 +1,358 @@
+package bot_api
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/assert"
+)
+
+// resolve/targets test fixtures. One User Bot (bot_token auth) that is a member
+// of a few groups; non-member groups/threads must never surface. Seeding helpers
+// build the membership-scoped visibility the handler relies on.
+const (
+	rtRobotID  = "bot_rt_1"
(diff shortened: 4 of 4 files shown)


Jerry-Xin (approved):
Summary: The PR is in scope for octo-server and the endpoint follows the existing bot API membership-based authorization model with good coverage for visibility, filtering, truncation, and i18n response guards.

💬 Non-blocking

🟡 Warning: LIKE escaping depends on MySQL treating backslash as the escape character, but other parts of this repo explicitly use `ESCAPE '\\'` to survive `NO_BACKSLASH_ESCAPES`. Consider adding an explicit `ESCAPE '\\'` to both `LIKE` clauses in `modules/bot_api/resolve_targets.go:110` and `modules/bot_api/resolve_targets.go:167`, paired with the existing escaping at `modules/bot_api/resolve_targets.go:90`.

🔵 Suggestion: For `kind=all`, exact-match ordering is per class, not unified. Groups are always appended before threads, so an exact thread match can appear after a partial group match. If clients use ordering for UI ranking, consider a final in-memory sort across `candidates` after both queries. See `modules/bot_api/resolve_targets.go:98` and `modules/bot_api/resolve_targets.go:134`.

✅ Highlights

The endpoint stays fail-closed on visibility by joining through `group_member.uid = robotID AND is_deleted = 0`.

The new source guard includes `resolve_targets.go`, keeping the bot API i18n error-response contract covered.

The integration tests cover the important ambiguity and visibility cases, including same-name group/thread, non-member hiding, wildcard escaping, archived/deleted thread status, and app-bot empty results.

Verification note: I attempted `go test ./modules/bot_api -run 'TestResolveTargets|TestBotAPINoLegacyResponseError'`, but the local test server panicked because `OCTO_MASTER_KEY` is not configured. This appears to be an environment prerequisite, not a failure specific to this PR.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#337 Review Report — `feat(bot_api): resolve channel target (group/thread) by name`
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/337
**Head SHA:** `8a29cc8cd16ea4ef3ed811c5dae1c0ba0d305d9b`
**Files:** `resolve_targets.go` (new, 198 lines), `resolve_targets_test.go` (new, 358 lines), `bot_api.go` (+1 route), `api_i18n_test.go` (+1 file to banned-legacy list)

---

## 1. Verification Summary

| Aspect | Status | Evidence |
|--------|--------|----------|
| Auth (bot token required) | ✅ | Route registered under `r.Group("/v1/bot", ba.authBot())` — `bot_api.go:203-209`. Auth middleware in `auth.go:27-44` resolves token → robotID via `authUserBot` (bf_ prefix) or `authAppBot` (app_ prefix). |
| Identity guard | ✅ | `resolve_targets.go:57-60` — empty robotID → `respondBotAPIIdentityMissing` (500/internal). |
| Visibility scoping (groups) | ✅ | `resolve_targets.go:109-110` — `WHERE gm.uid = ? AND gm.is_deleted = 0`. Matches `getGroups` precedent (`groups.go:38-49`). |
| Visibility scoping (threads) | ✅ | `resolve_targets.go:165` — `INNER JOIN group_member gm ON gm.group_no = t.group_no COLLATE ... AND gm.uid = ? AND gm.is_deleted = 0`. |
| LIKE wildcard escaping | ✅ | `resolve_targets.go:89-93` — backslash → `\\`, `%` → `\%`, `_` → `\_`. Identical to `botSpaceMembers` (`groups.go:330-333`). |
| Limit clamping | ✅ | `resolve_targets.go:84-87` — default 20, cap 50. Stricter than `botSpaceMembers` (50/200). |
| Truncation probe | ✅ | `LIMIT limit+1` on both queries; `len(rows) > limit` sets `truncated=true` and slices. |
| Kind filter validation | ✅ | `resolve_targets.go:69-73` — rejects unknown kind with 400. |
| Empty name rejection | ✅ | `resolve_targets.go:63-66` — trimmed empty name → 400. |
| COLLATE fix for Error 1267 | ✅ | `resolve_targets.go:165-166` — explicit `COLLATE utf8mb4_general_ci` on cross-collation JOIN equalities. Well-documented in comments (lines 140-161). |
| Channel type values | ✅ | `ChannelTypeGroup = 2`, `ChannelTypeCommunityTopic = 5` — verified from `octo-lib/common/constant.go:14-29` iota. |
| Channel ID construction | ✅ | `thread.BuildChannelID(groupNo, shortID)` = `"{groupNo}____{shortID}"` — verified from `modules/thread/service.go:802-804` + `const.go:17`. |
| App Bot behavior | ✅ | No group_member rows → empty result set (same as `getGroups` precedent). Test `TestResolveTargets_AppBotReturnsEmpty` confirms. |
| Test coverage | ✅ | 13 subtests covering: same-name disambiguation, multi-group threads, non-member exclusion, cross-space, kind filter, invalid kind, exact-match ordering, limit+truncation, group-full-not-squeeze-thread, wildcard escaping, empty name, archived/deleted status, app-bot-empty. |

---

## 2. Findings

### F1 — Missing `ESCAPE '\\'` clause on LIKE queries [P2]

**Location:** `resolve_targets.go:110` (group query) and `resolve_targets.go:167` (thread query).

**Issue:** Both LIKE clauses use Go-side backslash escaping (`\%`, `\_`) but omit the SQL-side `ESCAPE '\\'` clause. Under MySQL `sql_mode=NO_BACKSLASH_ESCAPES`, the backslash is no longer the default escape character, so `\%` would be interpreted as literal-backslash + wildcard `%` rather than literal `%`.

**Diff-scope (R2):**
1. **New** — the entire file is new; this omission is introduced by this PR.
2. **However**, the PR follows the existing `botSpaceMembers` precedent (`groups.go:335`) which has the same gap and has been in production without incident.
3. The codebase has a more robust pattern in `modules/space/db_manager.go:24` (`const likeEscapeClause = " ESCAPE '\\\\'"`) with explicit documentation of the risk. Other modules (`conversation_ext/service.go:38`, `opanalytics/db.go:313`) also use explicit ESCAPE clauses.

**Severity rationale (R1/R4):** P2, not P1. The gap only manifests under `NO_BACKSLASH_ESCAPES` sql_mode, which production MySQL apparently does not use (evidenced by `botSpaceMembers` running without incident). Under default MySQL config, backslash IS the escape character and the Go-side escaping works correctly. Impact if triggered: incorrect search results (wildcards not treated as literals), not data loss or security breach. This is a defense-in-depth gap, not a production blocker.

**Recommendation:** Add `ESCAPE '\\'` to both LIKE clauses to match the `space/db_manager.go` convention:
- Line 110: `g.name LIKE ? ESCAPE '\\'`
- Line 167: `t.name LIKE ? ESCAPE '\\'`

One-line-per-clause fix. Consider also backporting to `botSpaceMembers` for consistency.

---

## 3. Suggestions

1. **Adopt `likeEscapeClause` constant** from `modules/space/db_manager.go` (or extract it to a shared package) rather than relying on Go-side escaping alone. This is the most robust convention in the codebase.

2. **Consider indexing strategy** for the thread query: `COLLATE utf8mb4_general_ci` on `t.group_no` in the JOIN prevents MySQL from using an index on that column. This is necessary to avoid Error 1267, but if the thread table grows large, consider adding a generated column with the target collation or altering the thread table's collation to match `group`/`group_member`. The `LIMIT` clause mitigates this for now.

---

## 4. Additional Observations

- **No write side-effects:** This is a purely read-only endpoint. No notifications, events, or state mutations. R3 distributed side-effects checklist is N/A.
- **`ResponseErrorL` wire status:** Query failures return wire HTTP 400 (legacy D14 compatibility) with `errcode.ErrBotAPIQueryFailed` (Internal=true, canonical 500). The underlying error is properly logged via `ba.Error(...)` before responding. This matches existing bot_api convention.
- **Independent per-class limits:** Groups and threads each get `limit` results independently. A full group class cannot squeeze out threads. This is the correct design and is tested by `TestResolveTargets_GroupFullStillReturnsThread`.

---

## 5. Data-Flow Trace

|---------------|----------------|------------------|
| `robotID` | `authBot()` middleware → `CtxKeyRobotID` → `getRobotIDFromContext()` | ✅ Auth middleware sets it from DB lookup (User Bot) or registry/DB (App Bot). Handler guards empty with 500. |
| `name` (query param) | `c.Query("name")` → `strings.TrimSpace` → empty check | ✅ Trimmed, validated non-empty, then used as: (a) LIKE pattern after escaping + wrapping, (b) exact-match ORDER BY parameter. |
| `kind` (query param) | `c.Query("kind")` → `strings.TrimSpace` → default "all" → whitelist check | ✅ Only "all"/"group"/"thread" accepted. Invalid → 400. |
| `limit` (query param) | `c.Query("limit")` → `strconv.Atoi` → default 20 → cap 50 | ✅ Lenient parse (invalid keeps default). Cap enforced at line 86. Used as `limit+1` for truncation probe. |
| `pattern` (LIKE) | `name` → escape `\`/`%`/`_` → wrap `%..%` | ✅ Correct escaping order (backslash first). Passed as parameterized `?` to dbr (client-side interpolation, safe under default MySQL config). |
| `group_no` (group query result) | `SELECT g.group_no ... FROM group_member gm INNER JOIN group g ON gm.group_no = g.group_no` | ✅ Parameterized `gm.uid = ?` scopes to bot's membership. |
| `group_no`, `short_id`, `name`, `parent_name` (thread query result) | `SELECT t.group_no, t.short_id, t.name, g.name AS parent_name FROM thread t INNER JOIN group_member gm ... INNER JOIN group g ...` | ✅ Double JOIN scopes to bot's groups. `t.status IN (1,2)` filters deleted. COLLATE fixes cross-collation JOIN. |
| `ChannelID` (thread candidate) | `thread.BuildChannelID(r.GroupNo, r.ShortID)` | ✅ Verified: `fmt.Sprintf("%s____%s", groupNo, shortID)`. |
| `truncated` flag | Set when `len(rows) > limit` for either query class | ✅ Correctly OR'd across both classes. `total` = `len(candidates)` (post-truncation combined count). |

---

**C1 — Dual-path parity:**
- Auth middleware has two paths (User Bot / App Bot). Handler treats both identically via `robotID` — no parity concern. App Bots naturally get empty results (no group_member rows), matching `getGroups` precedent. ✅ Clear.
- No symmetric create/delete or subscribe/unsubscribe paths touched. N/A.
- No create-gate vs execute-path authorization asymmetry (read-only endpoint). N/A.

- LIKE escaping runs once, linearly, before both queries. No nested or repeated security-control calls. ✅ Clear.
- COLLATE is a SQL-level concern, not a reusable Go security control. N/A.
- No regex or sanitize function to test with non-canonical input forms. N/A.

- Endpoint is behind `authBot()` middleware. Only authenticated bots can call it. ✅ Clear.
- Visibility is scoped to bot membership (`group_member.uid = robotID AND is_deleted = 0`). A bot cannot see groups it doesn't belong to. ✅ Verified by `TestResolveTargets_NonMemberNotVisible`.
- Read-only endpoint — no write capabilities exposed. N/A for tool/capability exposure.

---

## 7. Verdict

No P0 or P1 findings. One P2 (missing `ESCAPE '\\'` clause — defense-in-depth gap following existing `botSpaceMembers` precedent, only exploitable under non-default MySQL `NO_BACKSLASH_ESCAPES` sql_mode). Auth, visibility scoping, input validation, LIKE escaping, limit clamping, COLLATE handling, and test coverage are all solid.

**[Octo-Q] verdict: APPROVE** — recommend adding `ESCAPE '\\'` to both LIKE clauses as a follow-up (P2, non-blocking).

yujiawei (approved):
# Code Review — PR #337 (octo-server)

**Scope:** new endpoint `GET /v1/bot/resolve/targets` (`modules/bot_api/resolve_targets.go`, +198) plus the route wiring (`bot_api.go`), the no-legacy-response guard list (`api_i18n_test.go`), and 358 lines of table tests. Reviewed at head `8a29cc8c`. Build, `go vet`, and all 13 new tests pass locally (after clearing unrelated test-DB migration drift from a sibling branch).

**Verdict: APPROVED.** No P0/P1. The implementation is correct, well-scoped, well-tested, and faithfully follows existing `bot_api` precedents. The items below are P2 performance-hardening notes and nits — none blocks merge.

---

## 1. What was verified ✅

- **Auth & identity** — Endpoint is mounted inside the `authBot()` group (`bot_api.go:217`); `robotID` comes only from `getRobotIDFromContext` (`auth.go:127`), set by the middleware after token verification. No client-supplied `uid`/`robot_id` param exists. The empty-`robotID` guard (`resolve_targets.go:56-59`) is correct belt-and-suspenders. **Not spoofable.**
- **Visibility model** — Both queries scope by membership (`group_member.uid = ? AND is_deleted = 0`), matching the `getGroups` precedent (`groups.go:50`). Threads are scoped by **parent-group** membership; threads have no per-thread ACL column, and `botListThreads` (`threads.go:105`) already exposes every thread in a member group — so this endpoint introduces **no new disclosure surface**. `TestResolveTargets_NonMemberNotVisible` covers this.
- **No-space-filter is intentional and correct** — documented in the handler comment; lets a bot resolve cross-space external groups it was pulled into. Consistent with `getGroups`.
- **SQL injection** — All user input is bound (`?`); LIKE metacharacters `\ % _` are escaped (`resolve_targets.go:90-92`) identically to `botSpaceMembers` (`groups.go:331-333`). `TestResolveTargets_WildcardEscaped` proves `%`/`_` are treated literally. **Clean.**
- **Channel-type / status constants** — `ChannelTypeGroup = 2`, `ChannelTypeCommunityTopic = 5` (octo-lib `common/constant.go`); thread `status IN (1,2)` correctly includes active+archived and excludes deleted (3). `TestResolveTargets_ArchivedThreadVisibleDeletedHidden` confirms.
- **App Bot handling** — App Bots have no `group_member` rows, so they resolve to an empty set rather than being rejected (matches the `getGroups` read precedent). `TestResolveTargets_AppBotReturnsEmpty` confirms.
- **Limit parsing** — `strconv.Atoi` with `> 0` guard, clamped to 50; overflow/garbage/`<=0` degrade leniently to the default 20. Safe and bounded.
- **Truncation probe** — the `LIMIT limit+1` over-fetch + `len(rows) > limit` pattern is correct per query class. `TestResolveTargets_GroupFullStillReturnsThread` guards the important case where a full group class must not squeeze out a matching thread.
- **i18n / response envelope** — Error paths use `httperr.ResponseErrorL` with registered codes (`ErrBotAPIQueryFailed` is `Internal=true`, so it never leaks the DB error to the wire — logged via `zap.Error`). The success path uses `c.JSON(http.StatusOK, …)`, which is the same pattern `getGroups` uses and is allowed (the project rule bans only **non-OK** raw `c.JSON`). `resolve_targets.go` is correctly added to the `TestBotAPINoLegacyResponseError` guard list.

## 2. Findings

### P2 — Thread query thread-name search is not storage-layer-bounded by membership

`resolve_targets.go:163-170`. Every other thread query in the codebase filters by `group_no = ?` first (`thread/db.go` `QueryByGroupNoWithStatus` etc.), so the bot's group bounds the scan. This new query's only WHERE predicates are `t.status IN (1,2)` (non-selective) and `t.name LIKE '%…%'` (leading wildcard, unindexable — there is **no index on `thread.name`**, see `thread/sql/20260402000001_thread_legacy01.sql:13-17`); membership is enforced only through the JOIN.

In the common case the optimizer should drive from `group_member` via the selective `group_member_uid` index (so the scan stays membership-bounded), but this relies on optimizer statistics rather than schema. For a bot in many groups, or under stale stats, the plan can degrade toward scanning thread rows outside the bot's footprint. The group query (`:108-113`) does not have this issue because `group_member(uid)` is the guaranteed driving table.

*Suggestion (non-blocking):* make `group_member` the guaranteed driving table — e.g. `STRAIGHT_JOIN` starting from `gm`, or pre-select the bot's `group_no` set and join `thread` with `t.group_no IN (…)` so `idx_group_no` / `idx_group_status_created_id` bound the scan. Either way, this is a good candidate for a **human/DBA sanity check of the `EXPLAIN` plan on production-sized data** before this gets heavy adapter traffic.

### P2 — COLLATE on the join key + `ORDER BY (col = ?)` force a non-index-satisfiable plan / filesort on the thread query

`resolve_targets.go:166,168`. The `COLLATE utf8mb4_general_ci` coercion on the join equality is genuinely **required** to avoid MySQL 8.0 error 1267 (thread.* is `utf8mb4_general_ci`; `group`/`group_member` inherit the DB default `utf8mb4_0900_ai_ci`) — that part is correct and well-documented in the comment. The side effect is that coercing the join collation prevents the `group` parent-name join from being satisfied by `group.group_no`'s native-collation unique index, and the `ORDER BY (t.name = ? …) DESC, …` expression cannot be index-satisfied, so the thread query will filesort. Combined with the P2 above, the thread path is the one to watch for performance.

*Suggestion (non-blocking):* none required for correctness — just confirm acceptable latency under realistic data, or consider normalizing the `thread` vs `group`/`group_member` collation at the schema level in a follow-up so the COLLATE coercions (and the index/sort penalties they bring) are no longer needed.

### P2 — Disbanded groups (and their threads) can still surface as candidates

`resolve_targets.go:109-110,163-166`. Neither query filters `g.status`. Group disband sets `group.status = 2` (`group/api.go:215`) but the member-row cleanup is commented out (`group/api.go:223`) and the disband event handler (`group/event.go`) doesn't clear `group_member.is_deleted` either — so a bot still satisfies the membership join for a disbanded group. Result: a disbanded group surfaces as a group candidate, and its still-active/archived threads surface with the disbanded group's name re-exposed via `parent_name`.

This is **pre-existing and shared with the `getGroups` precedent** (which also omits `g.status`), and the exposure is limited to a group the bot was itself a member of — not a cross-tenant escalation. Hence P2, not a blocker.

*Suggestion (non-blocking):* add `AND g.status = <active>` to both queries, or (better, repo-wide) fix the disband member-cleanup so departed memberships are marked `is_deleted = 1`.

### nit — `kind=all` can return up to `2 × limit` candidates; `truncated` is a single OR'd flag

`resolve_targets.go:95-96,193-197`. Each class is independently capped at `limit`, so `kind=all` with `limit=50` can return up to 100 candidates, and `total` (= `len(candidates)`) can exceed the requested `limit`. `truncated` is one bool OR'd across both classes, so a client can't tell *which* class was capped from that flag alone. This is **intentional and explicitly tested** (`TestResolveTargets_LimitClampAndTruncated` asserts `Total=4` at `limit=2`; `TestResolveTargets_GroupFullStillReturnsThread` is the deliberate "don't squeeze out the thread" guard), and a client can already recover per-class counts/exhaustion from the `kind` field on each candidate. The only gap is documentation.

*Suggestion (non-blocking):* note in the handler doc-comment (or API docs) that `limit` is **per-kind** and `truncated` means "at least one class was capped" — or, if a strict per-response cap is ever wanted, add per-kind `{count, truncated}` fields.

### nit — App-Bot empty-set safety is implicit, not enforced

`resolve_targets.go:54-59`. Unlike the mutating thread/group endpoints (which hard-reject App Bots via `ErrBotAPIAppBotUnsupported`), this handler — like the `getGroups` read precedent — relies purely on App Bots having zero `group_member` rows. Not exploitable today and consistent with precedent; a one-line `BotKindApp` short-circuit would make the invariant defensive rather than emergent. Optional.

## 3. Security-sensitive sign-off

Classified `security_sensitive`; reviewed accordingly. The endpoint is read-only, authenticated, membership-scoped, parameterized, and adds no new data-visibility surface beyond what `getGroups` / `botListThreads` already expose. **Recommended human verification:** an `EXPLAIN` of the thread query on production-sized data (the P2 scoping/filesort note) before it takes heavy adapter traffic — this is a performance/robustness check, not a correctness or access-control concern.

## 4. Test coverage

Excellent. 13 table tests cover same-name group/thread disambiguation, cross-group same-name threads, non-member invisibility, cross-space external groups, `kind` filtering, invalid `kind`, exact-match-first ordering, limit clamp + truncation, the group-full-still-returns-thread guard, wildcard escaping, empty name, archived-visible/deleted-hidden, and App-Bot-empty. Membership-scoped fixtures correctly seed the visibility the handler relies on.

PR #338, opened 2026-06-10
feat(i18n): localize BotFather command menu via per-request rendering (#335)

### LINKED ISSUE #335: i18n: BotFather command-menu descriptions are still hardcoded Chinese (registerBotFatherCommands)
## Background

Follow-up from the review of #316 (feat(i18n): localize BotFather outbound IM messages, part of #304).

PR #316 localizes all of BotFather's *per-recipient* outbound IM content (welcome / command replies / apply notifications / friend-approve flow) via the new `modules/base/common/msgtmpl` catalog. One user-visible surface was intentionally left out because it cannot be localized per recipient with the current storage model:

## Current behavior

`registerBotFatherCommands` (`modules/botfather/api.go:292-309`) writes the bot command menu — `/newbot`, `/help`, etc. with their descriptions — as a single JSON blob into the `robot` table at startup:

```go
commands := []map[string]string{
    {"command": CmdInstall, "description": "安装/更新 Octo 插件"},
    {"command": CmdQuickstart, "description": "AI Agent 快速入门"},
    {"command": CmdNewBot, "description": "创建新机器人"},
    // ... 16 entries, all zh-CN
}
```

Clients render this menu as-is, so an `OCTO_DEFAULT_LANGUAGE=en-US` deployment (and any en-US user) still sees a fully Chinese command menu — the same defect class #304 fixed for DM bodies.

## Why it wasn't fixed in PR #316

The menu is a **global single value per bot** stored in the DB (`robot.commands`), not a per-send render. Localizing it per recipient requires a storage/protocol change (per-language columns or client-side keys), which was out of scope for an outbound-message templating PR.

## Possible directions (to be decided)

1. **Deployment-level language** (cheapest): render the descriptions through the existing `msgtmpl` catalog with `OCTO_DEFAULT_LANGUAGE` at startup before writing the blob. Single-language deployments become correct; mixed-language deployments still see one language.
2. **Per-language storage**: store `commands` per supported language and have the sync/read path pick the requesting user's language. Correct for mixed deployments; needs schema + client/API coordination.
3. **Client-side keys**: ship stable command keys and let clients localize. Largest coordination cost.

Option 1 is a strict improvement and can ship independently of 2/3.

## Acceptance criteria (for option 1 as the minimum)

- [ ] Command descriptions live in `modules/botfather/templates/{lang}/` (or equivalent msgtmpl namespace), zh-CN wording unchanged
- [ ] `registerBotFatherCommands` renders them in the resolved deployment default language at startup
- [ ] Completeness test covers the new keys in every supported language (extend `allBotMessageKeys` / `TestBotMessagesCompleteness`)
- [ ] `make i18n-extract-check` + `make i18n-lint` stay green

an9xyz:
Closes #335

## Problem

`registerBotFatherCommands` wrote BotFather's command menu (`robot.bot_commands`) as a single hardcoded zh-CN JSON blob at startup, so an `OCTO_DEFAULT_LANGUAGE=en-US` deployment (and every en-US user) saw a fully Chinese command menu — the defect class #304 fixed for DM bodies, deferred in PR #316 because the stored blob can only carry one language.

## Approach: option 1 + per-request read-time rendering

The menu reaches clients through **authenticated HTTP reads** (`GET /v1/channels/botfather/1` → `extra.bot_commands`), where language negotiation is fully available — so per-user localization needs **no schema change and no client coordination** (see [analysis on the issue](https://github.com/Mininglamp-OSS/octo-server/issues/335#issuecomment-4669621187)):

- **New leaf package `modules/botfather/cmdmenu`** owns the 16 menu entries, descriptions rendered through the shared `msgtmpl` catalog (`templates/{zh-CN,en-US}`, zh-CN wording unchanged). Leaf because `botfather` imports `user`, so the read-side modules can only share the catalog through a package that imports neither. All languages are pre-rendered at init; a missing key in any supported language fails startup loudly (msgtmpl completeness matrix).
- **Startup write** (`registerBotFatherCommands`) now renders the deployment default language (`OCTO_DEFAULT_LANGUAGE`) — the floor for read paths without a request context (admin robot detail, batch enrichment). The write is unconditional on every boot, so existing deployments' zh-CN blobs **self-heal on upgrade; no migration**.
- **Per-request override for the `botfather` UID only**, via `i18n.OutboundLanguage(c.Request.Context())`, on the three menu-bearing endpoints:
  - `GET /v1/channels/botfather/1` → `extra.bot_commands` (channel module)
  - `GET /v1/robot/commands?robot_id=botfather` (robot module)
  - `GET /v1/users/botfather` (user module)
- **User bots' commands are creator content** (`POST /v1/bot/setCommands`) — never templated, never overridden.

## Acceptance criteria (#335)

- [x] Command descriptions live in a msgtmpl namespace (`modules/botfather/cmdmenu/templates/{lang}/`), zh-CN wording unchanged (pinned byte-for-byte by `TestZhCNWordingUnchanged`)
- [x] `registerBotFatherCommands` renders the resolved deployment default language at startup (`TestRegisterBotFatherCommands_WritesDeploymentDefault`)
- [x] Completeness test covers the new keys in every supported language (`TestMenuCompleteness` + msgtmpl matrix at init; `TestCmdMenuMatchesCommandConstants` guards drift against the `Cmd*` constants)
- [x] `make i18n-extract-check` + `make i18n-lint` green
- [x] Beyond option 1: mixed-language deployments are correct per user (`TestChannelGet_BotFatherMenuFollowsRequestLanguage`)

## Verification

E2E against the CI service stack (MySQL 8.0 + Redis + WuKongIM v2.2.4-20260313), `-race`, per-package DB reset:

- New e2e tests: Accept-Language negotiation (en-US / zh-CN / unset) on the channel endpoint via production `EarlyMiddleware`; deployment-default floor on channel + user-detail endpoints; startup-write self-heal; creator-content passthrough on `/v1/robot/commands`
- Full suites green: `modules/botfather` (31.7s), `cmdmenu`, `channel`, `robot`, `user` (60.5s)
- `go vet`, `golangci-lint` (0 issues), `make i18n-extract-check`, `make i18n-lint` all green

Note: the botfather user row's `robot=1` flag is set by deployment seed data, not by `initBotFatherUser` — without it the channel/user paths don't deliver `bot_commands` at all (pre-existing behavior, unchanged here; the e2e setup mirrors the seed with an explicit UPDATE).

https://claude.ai/code/session_019nGSMBbhaivbpQADEdoPhc

---
_Generated by [Claude Code](https://claude.ai/code/session_019nGSMBbhaivbpQADEdoPhc)_

### DIFF
--- a/modules/botfather/api.go
+++ b/modules/botfather/api.go
@@ -1,7 +1,7 @@
 package botfather
 
 import (
-	"encoding/json"
+	"context"
 	"fmt"
 	"net/http"
 	"strings"
@@ -13,11 +13,13 @@ import (
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
 	"github.com/Mininglamp-OSS/octo-server/modules/base/app"
 	"github.com/Mininglamp-OSS/octo-server/modules/base/event"
+	"github.com/Mininglamp-OSS/octo-server/modules/botfather/cmdmenu"
 	"github.com/Mininglamp-OSS/octo-server/modules/file"
 	"github.com/Mininglamp-OSS/octo-server/modules/group"
 	"github.com/Mininglamp-OSS/octo-server/modules/thread"
 	"github.com/Mininglamp-OSS/octo-server/modules/user"
 	"github.com/Mininglamp-OSS/octo-server/pkg/botutil"
+	octoi18n "github.com/Mininglamp-OSS/octo-server/pkg/i18n"
 	"github.com/tidwall/gjson"
 	"go.uber.org/zap"
 )
@@ -289,36 +291,20 @@ func (bf *BotFather) initBotFatherUser() {
 }
 
 // registerBotFatherCommands 注册BotFather自身的命令列表
+//
+// 库里这份 blob 只承载一种语言：部署默认语言（OCTO_DEFAULT_LANGUAGE）。它是
+// 所有拿不到请求上下文的读取面（管理端 robot 详情、批量用户详情填充）的兜底；
+// 面向用户的菜单端点会按请求协商语言用 cmdmenu 重新渲染（#335）。每次启动
+// 无条件覆写，所以部署默认语言变更后存量数据自愈，无需迁移。
 func (bf *BotFather) registerBotFatherCommands() {
-	commands := []map[string]string{
-		{"command": CmdInstall, "description": "安装/更新 Octo 插件"},
-		{"command": CmdQuickstart, "description": "AI Agent 快速入门"},
-		{"command": CmdNewBot, "description": "创建新机器人"},
-		{"command": CmdMyBots, "description": "查看我的机器人"},
-		{"command": CmdConnect, "description": "获取连接 prompt"},
-		{"command": CmdDisconnect, "description": "断开 Agent 连接"},
-		{"command": CmdSetName, "description": "修改机器人名称"},
-		{"command": CmdSetDescription, "description": "修改机器人描述"},
-		{"command": CmdDeleteBot, "description": "删除机器人"},
-		{"command": CmdToken, "description": "查看 Token"},
-		{"command": CmdRevoke, "description": "重置 Token"},
-		{"command": CmdApprove, "description": "通过好友申请"},
-		{"command": CmdReject, "description": "拒绝好友申请"},
-		{"command": CmdPending, "description": "查看待审批好友申请"},
-		{"command": CmdHelp, "description": "显示帮助"},
-		{"command": CmdCancel, "description": "取消当前操作"},
-	}
-	commandsJSON, err := json.Marshal(commands)
-	if err != nil {
-		bf.Error("序列化BotFather命令列表失败", zap.Error(err))
-		return
-	}
-	err = bf.db.updateBotCommands(BotFatherUID, string(commandsJSON))
+	// background ctx 没有协商语言，OutboundLanguage 即返回部署默认语言。
+	lang := octoi18n.OutboundLanguage(context.Background())
+	err := bf.db.updateBotCommands(BotFatherUID, cmdmenu.JSON(lang))
 	if err != nil {
 		bf.Error("注册BotFather命令列表失败", zap.Error(err))
 		return
 	}
-	bf.Info("BotFather命令列表注册成功")
+	bf.Info("BotFather命令列表注册成功", zap.String("lang", lang))
 }
 
 // ensureBotFatherFriends 批量为缺少BotFather好友关系的用户添加
--- a/modules/botfather/api_cmdmenu_e2e_test.go
+++ b/modules/botfather/api_cmdmenu_e2e_test.go
@@ -1,0 +1,144 @@
+package botfather
+
+// End-to-end coverage for #335: BotFather's command menu is server-owned copy,
+// stored once in the deployment default language (registerBotFatherCommands)
+// and re-rendered per request on the menu-bearing read endpoints. These tests
+// live in the botfather package — not in channel/user — because this is the
+// only package that may import every involved module: user cannot blank-import
+// botfather's migrations (botfather → user would cycle), yet GetUserDetail's
+// robot-details query needs botfather-owned columns (agent_platform, ...).
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-lib/server"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/Mininglamp-OSS/octo-server/modules/botfather/cmdmenu"
+	"github.com/Mininglamp-OSS/octo-server/modules/channel"
+	"github.com/Mininglamp-OSS/octo-server/modules/user"
+	octoi18n "github.com/Mininglamp-OSS/octo-server/pkg/i18n"
+	"github.com/stretchr/testify/assert"
+)
+
+// setupCmdMenuE2E builds the full server and normalizes the BotFather rows the
+// module bootstrap created during module.Setup: user.robot=1 (set by the
+// deployment seed in production, not by initBotFatherUser) and a deterministic
+// zh-CN blob in robot.bot_commands (the bootstrap renders whatever
+// OCTO_DEFAULT_LANGUAGE was at setup time). Also creates the login user.
+func setupCmdMenuE2E(t *testing.T) (*server.Server, *config.Context) {
+	t.Helper()
+	s, ctx := testutil.NewTestServer()
+
+	svc := user.NewService(ctx)
+	assert.NoError(t, svc.AddUser(&user.AddUserReq{UID: testutil.UID, Name: "Login User"}))
+	_, err := ctx.DB().UpdateBySql("UPDATE user SET robot=1 WHERE uid=?", BotFatherUID).Exec()
+	assert.NoError(t, err)
+	_, err = ctx.DB().UpdateBySql("UPDATE robot SET bot_commands=? WHERE robot_id=?",
+		cmdmenu.JSON("zh-CN"), BotFatherUID).Exec()
+	assert.NoError(t, err)
+
+	return s, ctx
+}
+
+// channelExtraBotCommands extracts extra.bot_commands from a ChannelResp body.
+func channelExtraBotCommands(t *testing.T, body []byte) string {
+	t.Helper()
+	var resp struct {
+		Extra map[string]interface{} `json:"extra"`
+	}
+	assert.NoError(t, json.Unmarshal(body, &resp), "body=%s", body)
+	v, _ := resp.Extra["bot_commands"].(string)
+	return v
+}
+
+// TestChannelGet_BotFatherMenuFollowsDeploymentDefault pins the #335 floor on
+// the fully wired route set: with no negotiated request language (testutil's
+// router mounts no EarlyMiddleware, mirroring a context-less caller) the
+// override renders OCTO_DEFAULT_LANGUAGE — so an en-US deployment no longer
+// serves the stored zh-CN blob.
+func TestChannelGet_BotFatherMenuFollowsDeploymentDefault(t *testing.T) {
+	s, ctx := setupCmdMenuE2E(t)
+	defer func() { _ = testutil.CleanAllTables(ctx) }()
+
+	t.Setenv(octoi18n.EnvDefaultLanguage, "en-US")
+
+	w := httptest.NewRecorder()
+	req, err := http.NewRequest("GET", "/v1/channels/"+BotFatherUID+"/1", nil)
+	assert.NoError(t, err)
+	req.Header.Set("token", testutil.Token)
+	s.GetRoute().ServeHTTP(w, req)
+
+	assert.Equal(t, http.StatusOK, w.Code, "body=%s", w.Body.String())
+	assert.Equal(t, cmdmenu.JSON("en-US"), channelExtraBotCommands(t, w.Body.Bytes()),
+		"en-US deployment must see the English menu even though the stored blob is zh-CN")
+}
+
+// TestChannelGet_BotFatherMenuFollowsRequestLanguage mounts the production
+// EarlyMiddleware on a fresh router (testutil's route set omits it) and pins
+// the per-request property #335's per-language-storage option was assumed to
+// require: two users of ONE deployment each see the menu in their own
+// negotiated language, with no schema change.
+func TestChannelGet_BotFatherMenuFollowsRequestLanguage(t *testing.T) {
+	_, ctx := setupCmdMenuE2E(t)
+	defer func() { _ = testutil.CleanAllTables(ctx) }()
+
+	r := wkhttp.New()
+	r.UseGin(octoi18n.EarlyMiddleware(octoi18n.MiddlewareOptions{DefaultLanguage: octoi18n.DefaultLanguage}))
+	channel.New(ctx).Route(r)
+
+	cases := []struct {
+		acceptLanguage string
+		want           string
+	}{
+		{"en-US", cmdmenu.JSON("en-US")},
+		{"zh-CN", cmdmenu.JSON("zh-CN")},
+		{"", cmdmenu.JSON(octoi18n.DefaultLanguage)},
+	}
+	for _, tc := range cases {
+		w := httptest.NewRecorder()
+		req, err := http.NewRequest("GET", "/v1/channels/"+BotFatherUID+"/1", nil)
+		assert.NoError(t, err)
+		req.Header.Set("token", testutil.Token)
+		if tc.acceptLanguage != "" {
+			req.Header.Set("Accept-Language", tc.acceptLanguage)
+		}
+		r.ServeHTTP(w, req)
+
+		assert.Equal(t, http.StatusOK, w.Code, "Accept-Language=%q body=%s", tc.acceptLanguage, w.Body.String())
+		assert.Equal(t, tc.want, channelExtraBotCommands(t, w.Body.Bytes()), "Accept-Language=%q", tc.acceptLanguage)
+	}
+}
+
+// TestUserGet_BotFatherMenuLocalized pins the #335 override on
+// GET /v1/users/botfather: the response carries bot_commands rendered in the
+// resolved language, not the stored zh-CN blob.
+func TestUserGet_BotFatherMenuLocalized(t *testing.T) {
+	s, ctx := setupCmdMenuE2E(t)
+	defer func() { _ = testutil.CleanAllTables(ctx) }()
+
+	fetch := func() string {
+		w := httptest.NewRecorder()
+		req, err := http.NewRequest("GET", "/v1/users/"+BotFatherUID, nil)
+		assert.NoError(t, err)
+		req.Header.Set("token", testutil.Token)
+		s.GetRoute().ServeHTTP(w, req)
+		assert.Equal(t, http.StatusOK, w.Code, "body=%s", w.Body.String())
+		var resp struct {
+			BotCommands string `json:"bot_commands"`
+		}
+		assert.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
+		return resp.BotCommands
+	}
+
+	t.Setenv(octoi18n.EnvDefaultLanguage, "en-US")
+	assert.Equal(t, cmdmenu.JSON("en-US"), fetch(),
+		"en-US deployment must see the English menu even though the stored blob is zh-CN")
+
+	t.Setenv(octoi18n.EnvDefaultLanguage, "zh-CN")
+	assert.Equal(t, cmdmenu.JSON("zh-CN"), fetch())
+}
--- a/modules/botfather/api_cmdmenu_e2e_test.go
+++ b/modules/botfather/api_cmdmenu_e2e_test.go
@@ -22,8 +22,26 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/modules/channel"
 	"github.com/Mininglamp-OSS/octo-server/modules/user"
 	octoi18n "github.com/Mininglamp-OSS/octo-server/pkg/i18n"
+	"github.com/go-redis/redis"
 	"github.com/stretchr/testify/assert"
 )
+
+// resetUIDRateLimit clears the per-uid token-bucket keys (ratelimit:uid:{uid})
+// so HTTP calls on SharedUIDRateLimiter-mounted routes (friend/sync) start
+// from a full bucket. The bucket persists in Redis and is NOT cleared by
+// CleanAllTables — pattern mirrors modules/category's resetUIDRateLimit.
+func resetUIDRateLimit(t *testing.T, ctx *config.Context) {
+	t.Helper()
+	rdsClient := redis.NewClient(&redis.Options{
+		Addr:     ctx.GetConfig().DB.RedisAddr,
+		Password: ctx.GetConfig().DB.RedisPass,
+	})
+	defer rdsClient.Close()
+	keys, err := rdsClient.Keys("ratelimit:uid:*").Result()
+	if err == nil && len(keys) > 0 {
+		_ = rdsClient.Del(keys...).Err()
+	}
+}
 
 // setupCmdMenuE2E builds the full server and normalizes the BotFather rows the
 // module bootstrap created during module.Setup: user.robot=1 (set by the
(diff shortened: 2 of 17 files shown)


Jerry-Xin (dismissed):
In scope: this PR directly updates octo-server’s BotFather command menu localization behavior and related HTTP read paths.

**💬 Non-blocking**
- 🟡 Warning: [modules/robot/api.go](/Users/mlamp/.openclaw/tmp/pr-review-fbws_cu8/repo/modules/robot/api.go:973) now returns BotFather commands before checking the `robot` table. Before this change, [modules/robot/api.go](/Users/mlamp/.openclaw/tmp/pr-review-fbws_cu8/repo/modules/robot/api.go:978) enforced `status=1` through `queryBotCommandsByRobotID`. This is probably acceptable for a built-in system bot, but it is a small behavior change: `/v1/robot/commands?robot_id=botfather` will return commands even if the BotFather robot row is missing or disabled.

**✅ Highlights**
- The new `cmdmenu` leaf package avoids import cycles cleanly and keeps the menu definitions centralized.
- Startup rendering uses deployment default language, while request paths use negotiated outbound language as intended.
- User-created bot commands remain DB-backed creator content and are not localized.
- Tests cover wording stability, template completeness, command drift, startup writes, request-language rendering, and creator-content passthrough.

**Verification**
- Ran `go test ./modules/botfather/cmdmenu` successfully.

yujiawei (changes requested):
# Code Review — PR #338 (octo-server)

## Summary

This PR localizes BotFather's command menu through a clean two-layer design: a new leaf package `modules/botfather/cmdmenu` pre-renders the static menu per supported language at init from embedded templates, the startup write persists the deployment-default language as a self-healing floor, and three read endpoints re-render per request via `cmdmenu.JSON(i18n.OutboundLanguage(ctx))`. The package boundary (leaf depending only on `msgtmpl` + `pkg/i18n`) correctly sidesteps the `botfather → user` import cycle, and the wire shape is preserved byte-for-byte. The test suite is thorough and the documentation is excellent.

I verified locally: affected packages build clean, `cmdmenu` unit tests pass, `EarlyMiddleware` is mounted globally (`main.go:137`) so `OutboundLanguage(ctx)` resolves the negotiated language in production, `user`/`channel`/`robot` import only the `cmdmenu` leaf (cycle claim holds), and the old `[]map[string]string` JSON output is byte-identical to the new `[]Command` output (Go sorts map keys `command` < `description`, matching struct field order; no HTML-escapable chars in the strings).

There is **one blocking issue** (P1): a user-facing surface that the per-endpoint override pattern missed.

## P1 — Friend-list / friend-sync surface is not localized (blocker)

The override is applied to exactly three **single-entity** reads: `channel.channelGet`, `robot.getCommands`, and `user.get`. But BotFather is made a friend of every user by `ensureBotFatherFriends` (`modules/botfather/api.go:311`, `INSERT IGNORE INTO friend ...`), and the friend surfaces — `friendSync` (`modules/user/api_friend.go:1085`), the space-mode friend list (`:1052`), and `friendSearch` (`:1134`) — all resolve BotFather through the **batch** path `Service.GetUserDetails` (`modules/user/service.go:566`).

That batch path copies `bot_commands` straight from the stored row with no per-request re-render:

```go
// modules/user/service.go:792-797
for _, resp := range userDetailResps {
    if resp.Robot == 1 {
        if d, ok := botMap[resp.UID]; ok {
            if d.BotCommands != "" {
                resp.BotCommands = d.BotCommands   // stored floor blob, never localized
            }
```

Net effect on a multi-language deployment: an en-US user sees the **English** menu on `GET /v1/users/botfather` (localized) but the **stored deployment-default** menu for the same bot in their friend list / contact sync. That directly contradicts the PR's stated goal of per-request localization, on a primary user-facing surface.

The package doc names "batch user-detail enrichment" as an intended floor consumer, so this may have been a deliberate scope call — but the doc's stated rationale ("every path *without a request context*") does not actually apply here: the friend endpoints have `c.Request.Context()` available at the call sites, so a negotiated language *is* obtainable. That mismatch between the rationale and the actual surface is what makes this read as an overlooked path rather than a justified exclusion.

This is exactly the cross-file blind spot the per-endpoint patch pattern is prone to: the same `BotCommands` field flows through a fourth path that the three point-fixes didn't cover.

**Requested change:** either (a) localize BotFather's `bot_commands` on the batch path too — thread `ctx` into `GetUserDetails` (note: it currently takes no `context.Context`) and override for the BotFather UID at the response-assembly loop, or override at each friend call site after the batch returns; or (b) if leaving the friend surface at the deployment-default floor is genuinely the intended v1 scope, update the package doc so the rationale matches reality (batch surfaces stay at floor *by choice*, not "because no request context exists") and confirm clients do not drive the BotFather command picker from friend-list data.

## Non-blocking (P2 / nits)

1. **Divergent override gating across the three sites** (`modules/robot/api.go:973-976`). `robot.getCommands` overrides for BotFather *unconditionally*, while `channel.channelGet` (`api.go:222-226`) and `user.get` (`api.go:1124-1125`) only override when a non-empty stored blob exists. Under the documented self-heal invariant (startup always writes a non-empty blob) all three agree, so this is not a live bug — but if `bot_commands` is ever empty/NULL (a read racing first boot, or an operator nulling the column), `getCommands` returns the full menu while the other two return nothing. Consider gating all three identically, or add a one-line comment on why `getCommands` intentionally ignores storage.

2. **`user.get` per-request negotiation is untested** (`modules/botfather/api_cmdmenu_e2e_test.go:199-223`). The `channel` and `robot` tests mount `EarlyMiddleware` and assert across `Accept-Language: en-US / zh-CN / empty`. `TestUserGet_BotFatherMenuLocalized` uses `s.GetRoute()` (no middleware) and only flips `OCTO_DEFAULT_LANGUAGE`, so it proves the env floor but never the per-request `OutboundLanguage(ctx)` path — a regression reading the wrong ctx key would still pass. Add a middleware-mounted `Accept-Language` case mirroring the channel test.

3. **Stored-blob byte shape isn't pinned** (`modules/botfather/cmdmenu/cmdmenu_test.go:449-471`). `TestZhCNWordingUnchanged` checks the decoded `Commands` slice and `TestJSONWireShape` only round-trips JSON→struct; neither asserts `JSON("zh-CN")` is byte-identical to the pre-migration `json.Marshal` output that clients parse and the startup write persists. They do match today (verified), but a future struct-tag/marshaling change could drift the stored bytes without failing any test. Consider asserting `JSON("zh-CN")` against the exact expected byte string.

## Security note (security_sensitive)

No security regression found. Language selection flows through the existing `EarlyMiddleware` trust boundary (trusted-header CIDRs, Accept-Language); a hostile `Accept-Language` can at most change the menu *language*, never bypass the BotFather-UID gate or leak data. Templates use `text/template` (correct — output is JSON-marshaled plain text, not HTML) and carry no caller-controlled parameters, so there is no injection surface. The `needs-human-review` label is appropriate; the one item warranting a human decision is the P1 scope question above.

## Verdict

Solid, well-tested work with one user-facing localization gap that should be closed (or its scope explicitly documented) before merge.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# PR #338 Review Report — feat(i18n): localize BotFather command menu via per-request rendering
**PR:** Mininglamp-OSS/octo-server #338
**Head SHA:** `3153833c260035238fcf140c6de9ae3374f2e933`

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Startup write renders deployment default | ✅ | `botfather/api.go:299-309` — `OutboundLanguage(context.Background())` → `LanguageOrDefault("")` → `DefaultLanguageFromEnv()` → `"zh-CN"` (or `OCTO_DEFAULT_LANGUAGE`). Unconditional overwrite = self-healing. |
| Per-request override on channel endpoint | ✅ | `channel/api.go:222-226` — three-part guard (`ChannelTypePerson && BotFatherUID && Extra != nil`) + inner `bot_commands` key-existence check. Precise. |
| Per-request override on robot endpoint | ✅ | `robot/api.go:973-976` — early return for BotFather UID before DB query. Returns `cmdmenu.Commands(lang)` directly. |
| Per-request override on user endpoint | ✅ | `user/api.go:1124-1126` — guard on `uid == BotFatherUID && BotCommands != ""`. Correct: empty means no commands to localize. |
| `msgtmpl.MustNew` completeness matrix | ✅ | `cmdmenu/cmdmenu.go:49` — `MustNew(templatesFS, "templates")` validates all supported languages × all template keys at init. Panic on failure = fail-loud. |
| `init()` pre-renders all languages | ✅ | `cmdmenu/cmdmenu.go:79-93` — iterates `SupportedLanguages()`, renders + marshals each. Panic on any failure. |
| `BotFatherUID` constant moved to leaf | ✅ | `cmdmenu/cmdmenu.go:33` canonical; `botfather/const.go` aliases it. Avoids import cycle (botfather imports user). |
| Template parity en-US ↔ zh-CN | ✅ | Both `templates/en-US/menu.tmpl` and `templates/zh-CN/menu.tmpl` define the same 16 `menu_*` blocks. |
| Test coverage | ✅ | 5 test files: E2E (Accept-Language negotiation), registration (startup write), unit (completeness/wording/fallback/copy), drift guard (Cmd* constants), robot endpoint (BotFather vs user bot). |

## 2. Findings

### No P0 or P1 findings.

### P2 / Nits (non-blocking)

**P2-1: Startup write is unconditional on every boot** (confidence: 8/10)
- `botfather/api.go:303` — `updateBotCommands` runs on every server start even when the stored value already matches.
- Impact: one extra DB write per boot. Idempotent, cheap, and the comment at line 297 explicitly justifies this ("部署默认语言变更后存量数据自愈，无需迁移").
- **Diff-scope:** pre-existing design choice, not introduced by this PR's logic. The PR's startup write replaces the old hardcoded write — same cost, better behavior.
- **Verdict:** acceptable trade-off. No action needed.

**P2-2: `normalize()` double-lookup** (confidence: 7/10)
- `cmdmenu/cmdmenu.go:97-103` — `MatchSupportedLanguage(lang)` returns a normalized tag, then `jsonByLang[norm]` checks presence. If `MatchSupportedLanguage` returns a tag not in `jsonByLang` (shouldn't happen since `init()` populates all supported languages), falls back to `SourceLanguage`.
- Impact: defensive code path that's unreachable in practice (init populates exactly `SupportedLanguages()`). The fallback is correct.
- **Diff-scope:** new code in this PR, but purely defensive. No action needed.

## 3. Data-Flow Tracing (Iron Law)

### 3.1 Startup write path
```
context.Background()
  → OutboundLanguage(ctx) [pkg/i18n/ctx.go:82-89]
    → LanguageFromContext(ctx) → "" (no value in background ctx)
    → LanguageOrDefault("") [pkg/i18n/config.go]
      → DefaultLanguageFromEnv() → env OCTO_DEFAULT_LANGUAGE or "zh-CN"
  → cmdmenu.JSON("zh-CN") → jsonByLang["zh-CN"] (pre-rendered at init)
  → bf.db.updateBotCommands("botfather", json) → robot.bot_commands column
```
**Result:** ✅ Data flows correctly from language resolution → template rendering → DB write.

### 3.2 Channel endpoint per-request override
```
c.Request.Context() [gin context with language set by EarlyMiddleware]
  → OutboundLanguage(ctx) → LanguageFromContext(ctx) → negotiated language (e.g. "en-US")
  → cmdmenu.JSON("en-US") → jsonByLang["en-US"] (pre-rendered at init)
  → channelResp.Extra["bot_commands"] = json (overrides DB-sourced value)
```
**Upstream data source:** `Extra["bot_commands"]` is populated from `robot.bot_commands` column via `GetUserDetail` → `newChannelRespWithUserDetailResp`. The `Extra` map is always `make(map[string]interface{})` for person channels (verified at `modules/user/1module.go:184`). The `bot_commands` key is conditionally added only when DB value is non-empty.
**Result:** ✅ Override guard is precise: checks channel type, UID, Extra non-nil, and key existence.

### 3.3 Robot endpoint per-request override
```
c.Request.Context() → OutboundLanguage(ctx) → negotiated language
  → cmdmenu.Commands(lang) → commandsByLang[lang] (copy of pre-rendered slice)
  → c.Response(commands) → JSON array to client
```
**Early return skips DB query entirely.** Auth: behind `AuthMiddleware` (token-based). Any authenticated user can query any bot's commands — by design, same as pre-PR.
**Result:** ✅ Correct. BotFather commands served from in-memory catalog; user bots still query DB.

### 3.4 User endpoint per-request override
```
c.Request.Context() → OutboundLanguage(ctx) → negotiated language
  → cmdmenu.JSON(lang) → jsonByLang[lang]
  → userDetailResp.BotCommands = json (overrides DB-sourced value)
```
**Upstream data source:** `BotCommands` populated from `robot.bot_commands` column via `GetUserDetail` SQL (line 397, fill at 523-540). `omitempty` JSON tag means empty string won't appear in response.
**Guard:** `uid == BotFatherUID && BotCommands != ""` — correct: only override when there IS a value to localize.
**Result:** ✅ Data flows correctly.

## 4. R5 Blind-Spot Checklist (security_sensitive — mandatory)

### C1 — Dual-path parity (symmetric paths)
**Clear.** The write path (startup registration) and read paths (3 per-request endpoints) are intentionally asymmetric:
- Write: startup → DB (deployment default language, floor for all readers)
- Read: per-request → in-memory catalog override (negotiated language, BotFather UID only)

The PR correctly overrides all three menu-bearing read endpoints (channel, robot, user). No read endpoint is missed. No create/delete symmetry applies here.

**Create-gate ↔ execute-path authorization:** All three read endpoints are behind `AuthMiddleware`. The override logic doesn't change authorization — any authenticated user could already read BotFather's commands (zh-CN). The PR only changes the language, not the access control.

**Clear.** `cmdmenu.JSON()` and `cmdmenu.Commands()` are pure functions with no side effects — they read from pre-rendered in-memory maps. No nesting/ordering risk. `OutboundLanguage()` is idempotent — calling it multiple times in the same request returns the same value.

**Safety control non-canonical test:** Not applicable — no regex/escape/sanitize in this PR.

**Clear.** No new endpoints, tools, or capabilities are exposed. The PR only changes the rendering language of existing data on existing endpoints. Authorization boundaries are unchanged:
- All three endpoints remain behind `AuthMiddleware`
- BotFather UID is a public constant, not a secret
- Template files are embedded via `//go:embed`, not user-controlled

## 5. Suggestions

None blocking. The implementation is clean and well-tested.

## 6. Extra Observations

- The `cmdmenu` leaf-package design is a good solution to the import cycle problem (`botfather` imports `user`, so read-side modules can't import `botfather`). Clean separation.
- The `MustNew` completeness matrix at init is excellent fail-loud engineering. A missing template key in any supported language crashes startup, not a silent half-menu in production.
- The `Commands()` function returns a defensive copy (`copy(out, src)`) — prevents callers from mutating the shared catalog. Good practice.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The two-layer localization design (startup floor + per-request override) is correctly implemented across all three menu-bearing endpoints. Data-flow tracing confirms every consumed value reaches its consumption point through the expected upstream path. R5 blind-spot checklist is clear on all three axes. Test coverage is comprehensive (E2E, unit, drift guard, registration).

an9xyz:
Pushed `654d30a` addressing the reviews — thanks @yujiawei for catching the batch-path gap, and @Jerry-Xin for the `status=1` note.

**P1 (batch user-detail surface)** — fixed via option (a): `IService.GetUserDetails` now takes a `context.Context`, and BotFather's menu is rendered per request at the single fill point in `Service.GetUserDetails` (gated identically to the single-get path: only when the stored blob is non-empty). All five call sites — friend sync, space friend list, friend search, search, conversation enrichment — pass `c.Request.Context()`. The `cmdmenu` package doc now states the floor serves only genuinely context-less paths (admin robot detail), so the rationale matches reality. New e2e `TestFriendSync_BotFatherMenuFollowsRequestLanguage` routes BotFather through the batch fill via the default-space pseudo-friend branch and asserts per-language rendering on the wire.

**P2-1 + Jerry-Xin's note (gating divergence / `status=1` bypass)** — `getCommands` now overrides *after* the existence/`status=1`/empty gating, so all read surfaces share the same "only when a stored blob exists" condition and a missing/disabled BotFather row behaves exactly as before the migration.

**P2-2 (user.get negotiation untested)** — added `TestUserGet_BotFatherMenuFollowsRequestLanguage` (EarlyMiddleware-mounted router, `Accept-Language: en-US/zh-CN`).

**P2-3 (stored byte shape unpinned)** — added `TestJSONStoredByteShapeUnchanged` asserting `JSON("zh-CN")` byte-for-byte against the pre-migration `json.Marshal` output.

Verification (local CI-equivalent stack, `-race`, per-package DB reset): full suites green for `botfather` (32.3s), `cmdmenu`, `user` (60.1s), `robot`, `channel`, `search`, `message`; `make i18n-extract-check` / `i18n-lint` / `go vet` / `golangci-lint` all clean.

---

Jerry-Xin (approved):
Summary: The PR is in scope for octo-server and the localization approach is well-contained, with request-context rendering applied to the relevant BotFather command-menu read paths.

💬 Non-blocking
- 🟡 Warning: `gofmt -l` still reports formatting drift in [modules/botfather/const.go](/Users/mlamp/.openclaw/tmp/pr-review-t242_kws/repo/modules/botfather/const.go:51) and [modules/user/service.go](/Users/mlamp/.openclaw/tmp/pr-review-t242_kws/repo/modules/user/service.go:11). Please run `gofmt` on touched Go files before merge.

✅ Highlights
- The new `cmdmenu` leaf package avoids the existing `botfather -> user` cycle and keeps the shared BotFather UID/menu data accessible to read-side modules.
- Request-context language is propagated through batch user-detail paths in search, friend sync/search, and conversation enrichment.
- User bot commands remain creator-owned content and are not accidentally localized.
- I verified `go test ./modules/botfather/cmdmenu` and compile-only tests for `./modules/user`, `./modules/channel`, and `./modules/robot`.

yujiawei (approved):
# Code Review — PR #338 (octo-server)

## Verdict: APPROVED

A focused, well-architected i18n fix. It correctly localizes the server-owned BotFather command menu via per-request read-time rendering, with a sound layering story (deployment-default floor + per-request override) and genuinely strong test coverage. I found no correctness, security, or data-loss issues that block merge. A couple of P2 maintainability nits are noted below for a future pass.

## 1. Verification of claims

- ✅ **Leaf-package design avoids the import cycle.** `botfather` imports `user`, so the read-side modules (`user`, `channel`, `robot`) cannot import `botfather`. Routing the catalog and `BotFatherUID` through the new leaf `modules/botfather/cmdmenu` (which imports only `msgtmpl` + `pkg/i18n`) is the right structural call. `const.go:9` aliases `BotFatherUID = cmdmenu.BotFatherUID` so in-package call sites are unaffected.
- ✅ **All three request-bearing read paths are localized, with consistent gating.** Channel `extra.bot_commands` (`channel/api.go:223`), `/v1/robot/commands` (`robot/api.go:984`), `/v1/users/:uid` (`user/api.go:1124`), and the batch fill `GetUserDetails` (`user/service.go:803`) all override only for `BotFatherUID` and only when the stored blob is non-empty — matching the startup-write invariant.
- ✅ **Startup write self-heals without migration.** `registerBotFatherCommands` (`botfather/api.go:303`) now renders `OutboundLanguage(context.Background())` → deployment default, and the write is unconditional on every boot, so existing zh-CN blobs are overwritten on upgrade.
- ✅ **Cross-file blind-spot sweep (no leak on unpatched paths).** I checked every other site that emits `bot_commands`:
  - `robot/api.go` `spaceBots` (`:1291`) and `myBots` (`:1405`) both exclude BotFather directly in SQL (`AND ... uid != 'botfather'`), so they never surface the stored blob.
  - The singular `GetUserDetail` is reached by `channel.channelGet` (output re-localized at `channel/api.go:224`), `user.get` (re-localized at `user/api.go:1125`), and the invite-inviter lookup (`user/api.go:1147`, which uses only `Name`/`Remark`, never `BotCommands`). `clearChannelMessages` uses the channel datasource only for a permission check and emits no client menu.
  - `webhook_identity.go` resolves only `iwh_`-prefixed synthetic identities, never BotFather.
  - Net: every client-facing emission of the BotFather menu is either localized or structurally excluded. No missed path.
- ✅ **Pre-render at init fails loud, not at request time.** `cmdmenu.init()` renders every supported language and `panic`s on a render/marshal error; the msgtmpl completeness matrix backs this. Request-time lookups are O(1) map reads.
- ✅ **`Commands()` returns a defensive copy** (`cmdmenu.go:115`), so callers cannot mutate the shared pre-rendered slice. `TestCommandsReturnsCopy` guards it.
- ✅ **Drift guards in place.** `TestCmdMenuMatchesCommandConstants` pins the leaf package's command literals to the `Cmd*` constants; `TestZhCNWordingUnchanged` / `TestJSONStoredByteShapeUnchanged` pin zh-CN wording byte-for-byte.
- ✅ **Build + targeted tests green locally** (`go1.26.3`): affected packages build clean, `go vet ./modules/botfather/...` clean, `cmdmenu` unit tests and `const` drift test pass.

## 2. Security review (PR is security_sensitive)

The `security_sensitive` classification is a keyword artifact (menu labels like "View a token" / "Reset a token", and `middleware` references in nearby context) rather than a real auth/credential change. Items a human may want to confirm:

- **Language negotiation source.** `OutboundLanguage` derives the tag from `EarlyMiddleware` (Accept-Language / query / cookie / trusted header). These are attacker-controllable, but `normalize()` (`cmdmenu.go:99`) clamps any input to the supported matrix via `MatchSupportedLanguage`, falling back to `SourceLanguage` (`en-US`). There is **no injection vector**: the tag only selects a pre-rendered map entry; it is never interpolated into a query or template.
- **No new data exposure.** These endpoints already returned `bot_commands` for BotFather; only the *language* of static, server-owned descriptions changes. No new field, no change to authz gating.
- **Menu content is fully static.** Descriptions come from embedded `templates/{lang}/menu.tmpl`, JSON-marshaled at init; no user input flows into the menu body. The "token" entries are UI labels, not secrets.

No security blockers.

## 3. Suggestions (non-blocking, P2 / nits)

- **P2 — hardcoded `'botfather'` literal in SQL now that a canonical constant exists.** `robot/api.go:1291` and `:1405` (and the e2e setup) still embed the string `'botfather'` directly. With `cmdmenu.BotFatherUID` now the documented source of truth, these literals are a latent drift risk if the UID ever changes. Not worth blocking on (the value is effectively frozen), but a follow-up could centralize them.
- **Nit — comment language.** New doc comments mix English and Chinese, which is consistent with the surrounding file conventions, so no change requested — just flagging that the leaf package's public-facing godoc is English (good) while call-site comments are Chinese (matches local style).

## 4. Additional findings (outside PR scope)

- The two pre-existing gofmt deltas in `modules/user/service.go` (import grouping) and `modules/botfather/const.go` (`State*` const alignment) are **present on `main`** and not introduced by this PR — the repo formats with local-prefix import grouping (gci/goimports), not plain `gofmt`. No action needed for this PR; noting so the next formatter run isn't mistakenly attributed here.

Nice work — the layering rationale in the package doc comment and the e2e coverage of the request-language vs deployment-default split made this straightforward to verify.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#338 Review Report
**PR:** [feat(i18n): localize BotFather command menu via per-request rendering (#335)](https://github.com/Mininglamp-OSS/octo-server/pull/338)
**Head SHA:** `654d30a7da99d021bd42e9987a6ec9582ca07f36`

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| cmdmenu leaf package design | ✅ | `modules/botfather/cmdmenu/cmdmenu.go` — imports only `msgtmpl` + `pkg/i18n`; no import cycle |
| Startup write renders deployment default | ✅ | `api.go:294-302` — `OutboundLanguage(context.Background())` → `OCTO_DEFAULT_LANGUAGE` env → `DefaultLanguage("zh-CN")` fallback |
| Per-request override on all 4 read surfaces | ✅ | channel `api.go:220-225`, robot `api.go:987-990`, user singular `api.go:1122-1125`, batch `service.go:800-805` |
| Gating parity across all 4 surfaces | ✅ | All robot-table paths: existence + `status=1` (SQL) + non-empty check before override. Channel: `Extra["bot_commands"]` key-existence check (set only when `BotCommands != ""` at `1module.go:216`) |
| GetUserDetails ctx threading (5 call sites) | ✅ | `api_friend.go:1052,1085,1134`, `api_conversation.go:522`, `search/api.go:231` — all pass `c.Request.Context()` |
| No stale 2-arg callers of GetUserDetails | ✅ | Grep confirms zero old-signature callers; would be compile error |
| zh-CN wording byte-identical to pre-migration | ✅ | `cmdmenu_test.go:TestZhCNWordingUnchanged` + `TestJSONStoredByteShapeUnchanged` pin both struct and JSON byte shape |
| Wire format compatibility | ✅ | `Command{Command,Description}` struct tags produce identical JSON keys to old `map[string]string` |
| Thread safety | ✅ | `commandsByLang`/`jsonByLang` populated once at `init()`, read-only thereafter; `Commands()` returns defensive copy (verified by `TestCommandsReturnsCopy`) |
| BotFatherUID constant relocation | ✅ | Canonical in `cmdmenu.BotFatherUID`; `botfather.BotFatherUID = cmdmenu.BotFatherUID` alias; all read-side modules reference `cmdmenu` directly |
| User bots' commands untouched | ✅ | All 4 override sites gate on `uid == cmdmenu.BotFatherUID`; verified by `TestGetCommands_CmdMenuLocalization` "user bot stays creator content" case |
| Test coverage | ✅ | Unit: completeness, wording pin, wire shape, fallback, copy safety, constant drift. E2E: Accept-Language negotiation, deployment-default floor, startup self-heal, batch path (friend sync), creator-content passthrough |

## 2. Findings

### No P0 or P1 issues found.

**P2-1: `webhook_identity.go` ChannelGet path serves stored blob without per-request override**
- **File:** `modules/user/webhook_identity.go:52`
- **What:** `resolveWebhookChannel` calls `m.BussDataSource.ChannelGet(uid, ChannelTypePerson, loginUID)` which returns a `ChannelResp` with `Extra["bot_commands"]` from the stored blob. The i18n override in `channel/api.go:220-225` only fires on the HTTP handler path, not on this internal callback consumer.
- **Diff-scope:** pre-existing — the webhook identity path consumed the stored blob before this PR and continues to do so. The PR's scope is client-facing HTTP endpoints.
- **Impact:** If BotFather ever appears as a webhook identity, the display name/command menu in that context would show the deployment default language. Low likelihood (BotFather is unlikely to be a webhook sender), and the deployment default is the designed floor for context-less paths.
- **Recommendation:** Note for future; no action needed in this PR.

**P2-2: `cmdmenu.normalize()` double-normalizes through `MatchSupportedLanguage`**
- **File:** `cmdmenu.go:103-109`
- **What:** `normalize()` calls `octoi18n.MatchSupportedLanguage(lang)` then checks `jsonByLang[norm]`. The `msgtmpl.Render` path (used at init time) already normalizes internally. At request time the double-normalize is a harmless extra map lookup.
- **Impact:** Negligible performance; no correctness issue. The explicit normalize is actually good defensive coding since `jsonByLang` is keyed by the normalized tag.

## 3. Data Flow Trace

### Write path (startup)
```
registerBotFatherCommands()
  → OutboundLanguage(context.Background())
    → DefaultLanguageFromEnv() reads OCTO_DEFAULT_LANGUAGE
    → LanguageOrDefault returns env default (no ctx decision)
  → cmdmenu.JSON(lang)
    → normalize(lang) → MatchSupportedLanguage → jsonByLang[norm]
  → db.updateBotCommands(BotFatherUID, blob)
```

### Read path (per-request, all 4 surfaces)
```
HTTP Request with Accept-Language header
  → EarlyMiddleware → NegotiateLanguage → WithLanguage(ctx, decision)
  → Handler reads OutboundLanguage(c.Request.Context())
    → LanguageFromContext(ctx) returns negotiated decision
    → (D9 lazy merge: user.Language can override Accept-Language if higher priority)
  → cmdmenu.JSON(lang) or cmdmenu.Commands(lang)
    → normalize(lang) → pre-rendered cache lookup
  → Response with localized menu
```

### Batch path (GetUserDetails)
```
friendSync / friendSearch / search / syncUserConversation
  → GetUserDetails(c.Request.Context(), uids, loginUID)
    → SQL: SELECT ... FROM robot WHERE robot_id IN ? AND status=1
    → botMap[resp.UID].BotCommands (non-empty gate)
    → if resp.UID == BotFatherUID: cmdmenu.JSON(OutboundLanguage(ctx))
```

### Channel path
```
channelGet handler
  → BussDataSource.ChannelGet callback (1module.go:56)
    → GetUserDetail(channelID, loginUID) [singular, no ctx]
    → newChannelRespWithUserDetailResp: Extra["bot_commands"] = user.BotCommands (if non-empty)
  → Handler override: if Extra["bot_commands"] exists → cmdmenu.JSON(OutboundLanguage(ctx))
```

## 4. R5 Blindspot Checklist (security_sensitive)

**C1 — Dual-path parity:** CLEAR
- No add/remove or subscribe/unsubscribe symmetry involved.
- All 4 read surfaces carry the override; the single write surface (startup) renders deployment default.
- `GetUserDetail` (singular) and `GetUserDetails` (batch) both have i18n coverage at the appropriate layer (handler vs service).
- `BotFatherUID` constant: single source of truth in `cmdmenu`, aliased in `botfather`. No drift risk.

**C2 — Control-flow ordering / nested reuse:** CLEAR
- Override placement verified: all 4 sites apply override AFTER existence/status/non-empty gating.
- `cmdmenu.JSON()`/`Commands()` are pure functions (cache lookup, no side effects). No double-rendering risk.
- Template content is static (no user input), so no injection surface.
- `normalize()` fallback chain is consistent with `msgtmpl` contract.

**C3 — Authorization boundary ≠ capability boundary:** CLEAR
- BotFather's menu is server-owned system content, not user-generated.
- Override fires only for `uid == "botfather"` — no user can inject content into this path.
- User bots' commands (`POST /v1/bot/setCommands`) are creator content and explicitly excluded from override.
- No new endpoints or capabilities exposed.

## 5. Extra Findings

None beyond P2 observations above.

---

**[Octo-Q] verdict: APPROVE**

The PR is well-architected with a clean leaf-package design, correct data flow across all 4 read surfaces, proper gating parity, comprehensive test coverage (including the batch path fix from commit 2), and no security concerns. The i18n negotiation chain is correctly threaded through `context.Context` at every call site. Wire format is byte-compatible with the pre-migration stored blob. No P0/P1 issues found.

PR #346, opened 2026-06-11
feat(user): server-rendered default avatar with nickname initials

### LINKED ISSUE #347: Server-rendered default user avatar with nickname initials (CJK)
## Background

The current default user avatar (`generateDefaultAvatar`) renders the first character of the uid on a colored circle using `basicfont`, which is ASCII-only and cannot display CJK names.

## Goal

Render a default user avatar that matches the UI design: a colored circle with the **last two characters of the nickname** in white, with the color picked from a fixed 10-color design palette stably by `uid` (so it stays the same across renames). Must support Chinese / Japanese / Korean.

## Scope

- Server-side PNG rendering with an embedded CJK font (self-contained binary, no client dependency, no cgo).
- Nickname non-empty validation: reject names that are empty after stripping whitespace / control / zero-width characters.
- Fallback to the legacy ASCII avatar for empty nicknames or glyphs the font cannot render (e.g. pure emoji), to avoid tofu/broken images.
- Cache correctness: since the rendered content depends on the nickname while the URL is stable, responses must carry a content-derived ETag and revalidate so a rename refreshes promptly.

Group default avatars are tracked separately as a follow-up.

an9xyz:
Closes #347

## Summary

Replaces the ASCII-only default user avatar with a CJK-capable renderer that draws the nickname's last two characters on a colored circle, matching the UI design palette.

- **`pkg/avatarrender`** (new): standalone package rendering a colored circle + centered white text to PNG with supersampled anti-aliasing. 10-color design palette picked stably by `uid` (color stays fixed across renames). Embeds a subsetted **Noto Sans CJK SC Bold** (SIL OFL 1.1; CN/JP/KR + Latin coverage; 12.8 MB, `OFL.txt` included). Pure Go — the binary stays self-contained and the Docker build is unchanged (no cgo).
- **`UserAvatar`**: the local-generation branch renders nickname initials; falls back to the legacy ASCII avatar when the nickname is empty or contains glyphs the font cannot render (e.g. pure emoji), avoiding tofu/broken images. Responses carry a content-derived **ETag** + `public, max-age=300, must-revalidate` with `If-None-Match` 304 handling, so a rename revalidates caches to the new image within minutes instead of serving the stale nickname's avatar for up to a day.
- **`ValidateName`**: rejects blank names (whitespace/control/zero-width only) — the shared guard across register/rename/admin paths (third-party login bypasses it, so login is unaffected).

Webhook synthetic-sender avatars and **group avatars are intentionally unchanged** (group default avatars are a separate follow-up).

## Test plan

- [x] `go test ./pkg/avatarrender/` (+ `-race`): circle/text rendering, palette stability, last-two-char truncation with invisible-char stripping, `Renderable` glyph detection, CN/JP/KR/Latin coverage verified
- [x] `go test ./modules/user/ -run 'TestValidateName|TestIsBlankName|TestAvatarETag'`: blank-name rejection (incl. full-width space / zero-width / BOM), ETag determinism & variation on rename
- [x] `go test ./modules/user/ -run 'TestUserAvatarDefaultETag|TestUserAvatarHandlerBotDefaultRouting'`: endpoint returns ETag + `must-revalidate`, `If-None-Match`→304, bot/normal/system routing intact
- [x] `go build` / `go vet` / `gofmt` clean


Jerry-Xin (dismissed):
This PR is in scope for `octo-server`: it updates the user avatar endpoint and shared user-name validation, both core repository functionality. I did not find blocking correctness or security issues.

**💬 Non-blocking**

🟡 Warning: `If-None-Match` is checked only after rendering and PNG encoding.
In [modules/user/api.go](/Users/mlamp/.openclaw/tmp/pr-review-ody4eg_u/repo/modules/user/api.go:565), `writePNG` handles 304 responses, but callers already did the expensive work at lines 577-584 or 590-597. Since `/v1/users/:uid/avatar` is public ([modules/user/api.go](/Users/mlamp/.openclaw/tmp/pr-review-ody4eg_u/repo/modules/user/api.go:296)) and the new renderer supersamples, draws, downsamples, and encodes PNGs ([pkg/avatarrender/render.go](/Users/mlamp/.openclaw/tmp/pr-review-ody4eg_u/repo/pkg/avatarrender/render.go:97)), cached revalidations still burn most of the CPU. Consider computing the ETag first and returning 304 before calling `Render` / `generateDefaultAvatar`.

🔵 Suggestion: `If-None-Match` matching currently uses exact string equality ([modules/user/api.go](/Users/mlamp/.openclaw/tmp/pr-review-ody4eg_u/repo/modules/user/api.go:570)). HTTP clients/proxies may send multiple validators or weak validators. A small helper using `etag` token parsing would make this more compliant.

🔵 Suggestion: `avatarETag` uses CRC32 ([modules/user/avatar_path.go](/Users/mlamp/.openclaw/tmp/pr-review-ody4eg_u/repo/modules/user/avatar_path.go:20)). That is acceptable for lightweight cache validation, but the comment says “weak ETag” while the wire value is a strong quoted ETag. Either prefix with `W/` or adjust the comment.

**✅ Highlights**

The renderer is isolated in `pkg/avatarrender`, embeds the font cleanly, and has good coverage for rendering, glyph availability, palette stability, and nickname truncation.

The fallback behavior for empty/unrenderable nickname text avoids broken default avatars.

The name validation change is centralized through `ValidateName`, and the added tests cover whitespace, full-width space, zero-width characters, and BOM-only names.

**Verification**

`go test ./pkg/avatarrender -count=1` passed.

`go test ./modules/user -run 'TestValidateName|TestIsBlankName|TestAvatarETag' -count=1` passed.

The broader user avatar handler test run could not initialize the shared test server because the local test database has an unknown migration recorded: `20260402000001_thread_legacy01.sql`.

yujiawei (dismissed):
# Code Review — PR #346 (octo-server)

**Scope reviewed:** server-rendered default avatar (`pkg/avatarrender/*`), the `UserAvatar` handler change + ETag/cache rework in `modules/user/api.go` / `modules/user/avatar_path.go`, and the `ValidateName` non-blank hardening. Verified against head SHA `9c35c4c4`.

## Verification

| Item | Result | Evidence |
|---|---|---|
| Package builds | ✅ | `go build ./pkg/avatarrender/ ./modules/user/` — clean |
| `go vet` | ✅ | `go vet ./pkg/avatarrender/` — clean |
| New unit tests pass | ✅ | `avatarrender` suite, `TestValidateName`, `TestIsBlankName`, `TestAvatarETag` all green |
| Imports complete | ✅ | `unicode`, `avatarrender`, `hash/crc32`, `pkg/errors` all present in `api.go`/`avatar_path.go` |
| Cross-file symbol (`avatarETag` in `avatar_path.go`, used in `api.go`) | ✅ | same `user` package, resolves |
| Font asset committed | ✅ | `fonts/NotoSansCJKsc-Bold-cjk-subset.otf` tracked, 13.4 MB |
| Third-party-login bypass claim | ✅ | `api_gitee.go:169`, `api_github.go:104` build name from `userInfo.Login` and never call `ValidateName`, so adding a blank check does not affect OAuth nickname-empty flows as the PR notes |

Overall this is solid, well-tested, well-documented work. The render pipeline (supersample → hard-edge circle → CatmullRom downscale), the deterministic palette/seed mapping, the emoji/`.notdef` fallback to the ASCII generator, and the content-aware ETag all hang together correctly. No P0/P1 issues found.

## Findings

### P2 — `Render()` runs on every 304 revalidation (wasted CPU)

`modules/user/api.go:577-588`

The `If-None-Match` check lives inside the `writePNG` closure, which is only called **after** `avatarrender.Render(...)` has already produced the PNG:

```go
text := avatarrender.IndividualText(userInfo.Name)
if avatarrender.Renderable(text) {
    imageData, genErr := avatarrender.Render(...) // full 800x800 render + CatmullRom + PNG encode
    if genErr == nil {
        writePNG(imageData, avatarETag("name-v1", uid, text)) // 304 decided here, bytes discarded
        return
    }
}
```

Because the ETag is a pure function of `uid` + `text` — both known *before* rendering — every revalidation request still does a full supersampled render (≈2.56 MB canvas alloc + downscale + PNG encode) only to throw the bytes away on a 304. Combined with the new `max-age=300, must-revalidate`, each active client revalidates every 5 minutes, so this is recurring wasted CPU on the hot default-avatar path (users without an uploaded avatar).

Suggested fix: compute the ETag first, short-circuit on `If-None-Match` before rendering:

```go
etag := avatarETag("name-v1", uid, text)
if match := c.GetHeader("If-None-Match"); match == etag {
    c.Header("ETag", etag)
    c.Header("Cache-Control", "public, max-age=300, must-revalidate")
    c.Status(http.StatusNotModified)
    return
}
// only now render
```

Same applies to the ASCII-fallback branch (`avatarETag("ascii-v1", uid)` — also computable before `generateDefaultAvatar`). Non-blocking, but it defeats most of the benefit the ETag was added for.

### Nit — ETag is documented as "weak" but is syntactically a strong validator

`modules/user/avatar_path.go:149-160` (and the `avatar_etag_test.go` comment). The format `fmt.Sprintf(`"%08x"`, ...)` emits a **strong** ETag (no `W/` prefix). The Go doc-comment calls it "弱 ETag". Functionally fine (exact-match comparison), just an inaccurate comment — either drop the "weak" wording or emit `W/"..."` if weak semantics are intended.

### Nit — visual inconsistency between the two fallback tiers

The new `avatarrender.Render` produces a circle with **transparent** corners (`drawCircle` leaves outside-circle pixels unset), whereas the ASCII fallback `generateDefaultAvatar` (`modules/user/avatar_gen.go:74`) fills the full square **white** first. On a non-white client background, name-avatars and ASCII-fallback avatars will look subtly different at the corners. Minor; most clients clip avatars to a circle anyway.

### Nit — `If-None-Match` only handles a single exact value

`modules/user/api.go:570` compares `match == etag`. The HTTP spec allows a comma-separated list and `*`. Browsers normally echo back the single ETag they got, so this is acceptable for an avatar endpoint, but a client sending multiple validators won't get a 304.

### Nit — README size figure

`pkg/avatarrender/fonts/README.md` states the subset is "12.8MB"; the committed file is 13.4 MB. Trivial doc drift.

## Note (not a defect)

`ValidateName` now rejects whitespace-/invisible-only names (ZWSP, BOM, full-width space, control chars). This is a deliberate, additive tightening and is correct, but it is a **behavior change**: names that were previously accepted (e.g. an all-full-width-space display name) will now be rejected at register / username-login / email-login / manager add+update / profile-update. The pre-existing `strings.TrimSpace(name) == ""` guards in front of most call sites already covered the plain-whitespace case; the real new coverage is the invisible-character case at `api.go:987` (which only checked `nameStr == ""`). No regression — flagging only for release-note awareness.

## Verdict

No correctness, security, data-loss, or build-breaking issues. The single P2 is a performance optimization, not a blocker; the rest are nits. Approving.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#346 Review Report
**PR:** [feat(user): server-rendered default avatar with nickname initials](https://github.com/Mininglamp-OSS/octo-server/pull/346)
**Head SHA:** `9c35c4c49f2ac408343520fbdac2ee932367c199`
**Files:** 13 changed (+702 / -7)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| CJK avatar rendering | ✅ | `pkg/avatarrender/render.go` — supersampled circle + centered text, CatmullRom downscale; `Renderable()` gates on glyph availability |
| Fallback to ASCII | ✅ | `api.go:585-596` — falls through to `generateDefaultAvatar(uid)` when text is empty, unrenderable, or render fails |
| ETag determinism | ✅ | `avatar_path.go:avatarETag` — CRC32 over NUL-separated parts; same inputs → same ETag (verified by `TestAvatarETag`) |
| Cache-Control semantics | ✅ | `api.go:568` — `public, max-age=300, must-revalidate` replaces old `max-age=86400`; rename changes ETag → cache revalidates within 5 min |
| If-None-Match → 304 | ✅ | `api.go:570` — strict string equality; correct for dominant browser behavior; verified by `TestUserAvatarDefaultETag` |
| ValidateName blank check | ✅ | `api.go:3868-3870` — `isBlankName` strips Unicode Cc/Cf/Space; all 6 callers verified safe; 3rd-party login deliberately bypasses |
| Font embedding & licensing | ✅ | `fonts/NotoSansCJKsc-Bold-cjk-subset.otf` (12.8MB, SIL OFL 1.1); `OFL.txt` + `README.md` included; subset covers CJK basic + kana + hangul + latin |
| Concurrency safety | ✅ | `loadFont()` uses `sync.Once`; `sfnt.Buffer` is per-call local in `Renderable`; `font.Drawer` is per-call local in `Render` |
| Test coverage | ✅ | Unit tests for `IndividualText`, `Renderable`, `ColorForSeed`, `avatarETag`, `isBlankName`, `ValidateName`; integration tests for ETag/304/routing |

---

## 2. Findings

### F1 — `avatarETag` doc comment says "弱 ETag" but produces strong ETag format (P2 / nit)

`avatar_path.go:15` doc comment says "弱 ETag" (weak ETag), but the implementation produces `"%08x"` — a strong ETag (no `W/` prefix per RFC 7232 §2.3). This is functionally correct because the rendering IS byte-for-byte deterministic for the same inputs, which satisfies the strong-ETag contract. The doc is misleading.

**Suggestion:** Either change the doc to say "强 ETag" or add `W/` prefix to the format string (`W/"%08x"`) if weak semantics are actually intended.

---

### F2 — `If-None-Match` comparison doesn't handle RFC 7232 edge cases (P2 / nit)

`api.go:570`:
```go
if match := c.GetHeader("If-None-Match"); match != "" && match == etag {
```

Strict string equality doesn't handle:
- Multiple ETags: `"abc", "def"` (RFC 7232 §3.2)
- Wildcard: `*`
- Weak prefix: `W/"abc"` (if a proxy transforms the ETag)

In all failure cases the result is a full 200 response instead of a bandwidth-saving 304 — **not a correctness issue**. Browsers send back the exact single ETag for GET requests in practice.

**Suggestion:** Consider using `etagMatch` helper that strips `W/` prefix and handles comma-separated lists, or document the limitation in a code comment. Low priority.

---

### F3 — Mixed-script nicknames (CJK + emoji) always fall back to ASCII avatar (P2 / UX)

**Diff-scope:** new (behavior introduced by this PR)

`IndividualText` preserves emoji runes; `Renderable` returns `false` if ANY character lacks a glyph. A nickname like `"张三😀"` → `IndividualText` returns `"三😀"` → `Renderable` returns `false` → falls back to ASCII avatar showing the UID's first character, not `"三"`.

This is a known limitation documented in `text.go:13` ("按 rune 截取，对由多个 rune 组合成的 emoji 可能切断；个人头像本期不要求 emoji 支持"). The fallback is safe (no broken images), but users with mixed CJK+emoji names miss the improved avatar.

**Suggestion (future):** Strip non-renderable runes from the extracted text before the `Renderable` check, so `"张三😀"` → `"三"` renders correctly. Not blocking.

---

### F4 — `ValidateName` blank check has bypass paths via `Service.AddUser` / `Service.UpdateUser` (P2 / pre-existing gap)

**Diff-scope:** pre-existing (not introduced or amplified by this PR)

Seven code paths create/update users without calling `ValidateName`:
- WeChat/GitHub/Gitee/OIDC OAuth login (intentional, documented)
- `Service.AddUser()` at `service.go:850` — `app_bot.go:379` passes admin-supplied `DisplayName` without blank check
- `Service.UpdateUser()` at `service.go:1148` — `app_bot.go:599` passes admin-supplied `DisplayName` without blank check

These are out of scope for this PR. The PR's `ValidateName` tightening correctly covers all 6 direct callers (register, rename, admin paths) without breaking any existing flow.

---

## 3. Suggestions

1. **F1 fix** (trivial): Align `avatarETag` doc comment with implementation — change "弱 ETag" to "强 ETag" or add `W/` prefix.
2. **F3 future enhancement**: Consider stripping non-renderable runes from `IndividualText` output so mixed-script names get the CJK avatar instead of ASCII fallback.
3. **Binary size awareness**: The 12.8MB embedded font is a one-time cost. Document it in the deployment guide if binary size is monitored.

---

## 4. Additional Observations

- **`sync.Once` for font loading** (`render.go:38-43`): If `opentype.Parse` fails, the error is permanently latched. Unlike PR#199's `sync.Once` issue, this is safe because the font data is `//go:embed` (compile-time constant) — either always succeeds or always fails, no transient-error latch risk.
- **Supersampled rendering** (4x → CatmullRom downscale): Produces high-quality anti-aliased circles and text at 200×200 output. The 800×800 intermediate canvas is ~2.5MB RGBA but allocated per-request only when rendering is needed, then GC'd. Acceptable for the expected request rate (cached by ETag).
- **Bot/webhook avatars unchanged**: Bot default avatars (`api.go:530`) and webhook fallback avatars (`webhook_identity.go:106`) still use `max-age=86400` without ETags. This is correct — their content is stable per UID and doesn't have the rename-invalidation problem.
- **Vertical centering formula** (`render.go:161`): `baselineY = center + (Ascent - Descent) / 2` is mathematically correct for centering text vertically when ascent and descent are positive values (as returned by `opentype.Face.Metrics()`).

---

## 5. Data-Flow Trace

|---|---|---|
| `userInfo.Name` | `SELECT * FROM user WHERE uid=?` → `Model.Name` (db.go:108) | ✅ — DB query returns display name; empty name → `IndividualText` returns `""` → `Renderable("")` = false → ASCII fallback |
| `uid` | `c.Param("uid")` (api.go:464) | ✅ — URL path param; used for `ColorForSeed`, `avatarETag`, and ASCII fallback |
| `avatarrender.IndividualText(name)` | Strips Cc/Cf/Space runes, takes last 2 (text.go:18-28) | ✅ — Returns visible text; empty for blank/emoji-only names after stripping |
| `avatarrender.Renderable(text)` | Checks each rune has non-.notdef glyph in embedded font (render.go:51-63) | ✅ — Returns false for empty text, emoji, chars outside CJK basic block |
| `avatarrender.ColorForSeed(uid)` | `crc32(uid) % 10` → palette index (palette.go:27) | ✅ — Deterministic, stable per UID, even distribution across 10 colors |
| `avatarETag(parts...)` | CRC32 over NUL-separated parts (avatar_path.go:21-27) | ✅ — Deterministic; changes when text or uid changes; quoted format for HTTP |
| `If-None-Match` header | Browser echoes ETag from prior response | ✅ — Strict equality works for the dominant single-ETag browser case |
| `generateDefaultAvatar(uid)` | ASCII fallback: HSL color from uid hash + first char of uid (avatar_gen.go:65) | ✅ — Unchanged by PR; reached when rendering is skipped or fails |

---

## 6. Three-Blindspot Checklist (R5)

**C1 — Dual-path parity:** N/A. This PR adds a new rendering path (CJK name avatar) alongside the existing ASCII fallback. They are not symmetric paths — they are a primary + fallback cascade. No add/remove, subscribe/unsubscribe, or create/delete symmetry applies.

**C2 — Control-flow ordering / nesting reuse:** Clear. The `writePNG` closure is defined once and called in exactly two places (CJK render success, ASCII fallback). No nested/double-call risk. The `loadFont` `sync.Once` is called from `Renderable` and `Render` — both are on the same code path (Renderable gates before Render), so no ordering issue. The ETag format is simple and not reused in compound contexts.

**C3 — Authorization boundary ≠ capability boundary:** N/A. This PR does not add or modify any endpoints, authorization checks, or tool exposure. The `UserAvatar` handler's access control is unchanged.

---

## 7. Verdict

No P0 or P1 findings. All four findings are P2 (nits / UX improvements / pre-existing gaps). The code is well-structured with solid fallback logic, comprehensive test coverage, and good documentation. The data flow from DB → text extraction → rendering → HTTP response is correctly implemented with proper ETag-based cache invalidation.

**[Octo-Q] verdict: APPROVE** — zero P0/P1 findings; four P2 items are non-blocking suggestions.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#346 Review Report
**PR:** Mininglamp-OSS/octo-server #346
**Title:** feat(user): server-rendered default avatar with nickname initials
**Head SHA:** 9e77ffa837d0e1258d59e2d52eb0463d226f2320
**Files:** 13 changed (+773 / -6)

---

## 1. Verification Summary

| Area | Status | Evidence |
|---|---|---|
| New `pkg/avatarrender` package | ✅ | `render.go`, `text.go`, `palette.go` — clean standalone package with embedded font, supersampled rendering, glyph detection |
| UserAvatar handler integration | ✅ | `api.go:554-609` — correct branching: nameMode → render → fallback to ASCII on error |
| ETag + Cache-Control | ✅ | `avatar_path.go:16-55` — content-derived weak ETag, `must-revalidate`, If-None-Match → 304 |
| ValidateName blank-name guard | ✅ | `api.go:3871-3905` — `isBlankName` consistent with `IndividualText`'s `isInvisible` |
| Test coverage | ✅ | 5 test files covering rendering, text truncation, palette stability, ETag, 304, blank names, bot routing |
| Bot/webhook avatar paths untouched | ✅ | `bot_default_avatar.go` and `webhook_identity.go` not in diff; bot path returns before reaching changed code |
| Font licensing | ✅ | `fonts/OFL.txt` (SIL OFL 1.1) + `fonts/README.md` documenting source, subsetting, and compliance |

## 2. Findings

### F1 — P2: Vertical text centering formula off by Descent amount

**File:** `pkg/avatarrender/render.go:155`
**Severity:** P2 (visual quality, not functional breakage)
**Diff-scope:** **new** — this is new code in this PR.

```go
// 垂直居中：基线放在 中线 + (ascent-descent)/2，使字形视觉居中。
baselineY := fixed.I(size)/2 + (metrics.Ascent-metrics.Descent)/2
```

The correct formula for visual vertical centering is `baselineY = cy + (Ascent + Descent) / 2`, not `(Ascent - Descent) / 2`.

**Proof:** Text top = `baselineY - Ascent`, text bottom = `baselineY + Descent`. For visual centering, `(top + bottom) / 2 = cy` → `baselineY = cy + (Ascent + Descent) / 2`. The current formula places the baseline `Descent` pixels too high.

For Noto Sans CJK Bold at the supersampled scale (800px canvas, fontSize ≈ 250px), Descent ≈ 0.07–0.08em ≈ 18–20px in supersampled space → **~4–5px too high in the final 200px output**. Visually the text sits slightly above true center — noticeable on close inspection but not a broken image.

**Suggested fix:**
```go
baselineY := fixed.I(size)/2 + (metrics.Ascent+metrics.Descent)/2
```

### F2 — Nit: Redundant Content-Type on ASCII fallback path

**File:** `api.go:605` (after PR changes)
**Severity:** Nit (cosmetic, no functional impact)
**Diff-scope:** new (introduced by this PR's restructuring).

When `nameMode` is true and `Render()` fails, the code falls through to `generateDefaultAvatar` → `c.Data(http.StatusOK, "image/png", imageData)`. The `setAvatarHeaders()` call at line ~580 already set `Content-Type: image/png`, so the Content-Type argument to `c.Data` is redundant. Harmless but slightly confusing.

### F3 — Note: ValidateName behavioral change is safe

**File:** `api.go:3871-3877`
**Severity:** Note (informational, not a defect)
**Diff-scope:** new (additive guard).

`ValidateName` now rejects blank names (whitespace/control/zero-width only) before the existing `@` check. Verified all call sites:

- `api.go:960` (rename via `userUpdateWithField`): already has `nameStr == ""` check upstream; `isBlankName` adds zero-width/control-char coverage.
- `api_usernamelogin.go` (register): no prior blank check → new protection added.
- `api_emaillogin.go` (register): same as above.
- `api_manager.go:400` (admin login): has `req.Name == ""` check; `isBlankName` adds coverage.
- `api_manager.go:1052` (admin add user): has `TrimSpace == ""` check; `isBlankName` adds coverage.
- Third-party login (`external_login.go`): **intentionally bypasses** `ValidateName` → unaffected.

No existing users are affected (validation runs on write, not read). The change is purely additive and closes a gap where zero-width-only names could slip through registration.

## 3. Suggestions

1. **Fix the centering formula** (F1): Change `(metrics.Ascent-metrics.Descent)/2` to `(metrics.Ascent+metrics.Descent)/2` in `render.go:155`. Add a sample-verification step to the test (render a sample and visually confirm centering, or assert baseline offset against expected metrics).

2. **Consider a render integration test with metrics assertion**: The existing `TestRenderProducesValidPNG` validates size but not content positioning. A test that decodes the PNG and checks that the text bounding box is approximately centered would catch the F1 regression.

## 4. Additional Observations

- **Binary size**: The embedded font adds 12.8 MB to the binary. This is documented and intentional (CJK coverage). The subsetting rationale in `fonts/README.md` is thorough.
- **Font parse failure resilience**: If `opentype.Parse` fails (e.g., corrupted embed), `sync.Once` latches the error. `Renderable` returns false → all users get ASCII fallback. `Render` returns error → same fallback. Graceful degradation is correct.
- **CRC32 ETag collisions**: Theoretical collision risk exists but is negligible for cache validation (not a security boundary). The null-byte separator between parts prevents trivial cross-part collisions.
- **Cache policy change**: `max-age` reduced from 86400 (24h) to 300 (5min) + `must-revalidate`. This is the correct trade-off for content that changes on rename without URL change.

## 5. Data-Flow Trace

|---|---|---|
| `userInfo.Name` | `u.db.QueryByUID(uid)` → DB `user.name` column | ✅ — queried at `api.go:~510`, non-nil checked at `:~520` |
| `avatarrender.IndividualText(userInfo.Name)` | Strips invisible runes, takes last 2 visible | ✅ — consistent `isInvisible` with `isBlankName` |
| `avatarrender.Renderable(text)` | Checks each rune's glyph index in embedded font | ✅ — local `sfnt.Buffer` per call (concurrency-safe) |
| `avatarrender.ColorForSeed(uid)` | `crc32(uid) % len(palette)` | ✅ — deterministic, uid-based (stable across renames) |
| `avatarETag("name-v1", uid, text)` | CRC32 of mode+uid+text with null separators | ✅ — changes on rename (text changes) and per-user (uid changes) |
| `ifNoneMatchSatisfied(header, etag)` | RFC 7232 weak comparison (strips W/ prefix) | ✅ — tested with exact, wildcard, multi-value, spaces |
| Render failure → ASCII fallback | `generateDefaultAvatar(uid)` (existing) | ✅ — ETag overwritten to `ascii-v1` mode, headers consistent |
| `ValidateName` blank check | `isBlankName(name)` → same invisible-char criteria | ✅ — additive, all callers verified |

## 6. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. This PR adds a new rendering path with a well-defined fallback. No symmetric add/remove or subscribe/unsubscribe pairs are touched. The bot default avatar path and webhook avatar path are explicitly unchanged and verified not affected by the diff.
- **C2 — Control-flow ordering / nested reuse:** Clear. The `setAvatarHeaders` → ETag check → 304 short-circuit → render → fallback ordering is correct. No double-action or ordering inversion. The ETag overwrite on render failure (line ~601) correctly updates the header before `c.Data` serves the ASCII fallback.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This PR does not add or modify any endpoints, authorization checks, or tool exposure. The `UserAvatar` handler's existing auth model (public avatar endpoint) is unchanged.

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. One P2 (vertical centering formula) is a visual quality improvement, not a functional breakage. The PR is well-structured with thorough test coverage, correct ETag/cache semantics, graceful fallback on render failure, and a safe additive ValidateName guard.

Jerry-Xin (dismissed):
The PR is in scope for octo-server and the implementation is generally sound: it updates the user avatar endpoint, adds a focused renderer package, and covers the key cache and text-selection behavior.

💬 Non-blocking

🟡 Warning — [modules/user/avatar_path.go](/Users/mlamp/.openclaw/tmp/pr-review-p5cnp6bt/repo/modules/user/avatar_path.go:23): `avatarETag` uses CRC32 for the cache validator. This is probably acceptable for low-risk avatar caching, but it is easy to collide and user-controlled nicknames make crafted collisions possible. A truncated SHA-256 would avoid stale 304 responses without meaningful cost.

🔵 Suggestion — [modules/user/bot_default_avatar_test.go](/Users/mlamp/.openclaw/tmp/pr-review-p5cnp6bt/repo/modules/user/bot_default_avatar_test.go:203): the handler test claims to cover rename invalidation, but it compares different users, so the UID/color also changes. Consider updating the same user’s name and requesting the same URL again to prove nickname-only ETag variation at the endpoint level.

✅ Highlights

The renderer is isolated in `pkg/avatarrender`, uses `sync.Once` for font parsing, and the text extraction/renderability split gives a clean fallback path for unsupported glyphs.

The avatar endpoint correctly shortens cache lifetime for generated nickname avatars and handles `If-None-Match` before rendering, which avoids repeated PNG work during revalidation.

Verification: `go test ./pkg/avatarrender/` and `go test ./pkg/avatarrender/ -race` passed. The targeted `modules/user` tests could not run in this workspace because `testutil.NewTestServer` panicked on an existing migration-state mismatch before the PR tests executed.

yujiawei (dismissed):
# Code Review — PR #346 (octo-server)

## Verdict: APPROVED

A well-crafted, self-contained change. The new `pkg/avatarrender` package, the `UserAvatar` integration, and the `ValidateName` hardening are all clean, well-documented, and properly tested. Build, `go vet`, and `gofmt` are clean, and the avatar-render tests pass under `-race`. No P0/P1 issues found. A few P2 suggestions below are non-blocking.

## What I verified

- ✅ **Builds clean**: `go build ./pkg/avatarrender/ ./modules/user/` succeeds; `go vet ./pkg/avatarrender/` and `gofmt -l` clean.
- ✅ **Concurrency-safe**: `loadFont` uses `sync.Once`; `Renderable` uses a local `sfnt.Buffer`; `Render` builds a fresh `font.Face`/`font.Drawer` per call and writes only to a local canvas. No shared mutable state. `go test -race ./pkg/avatarrender/` passes.
- ✅ **Pure-logic tests pass**: `TestValidateName`, `TestIsBlankName`, `TestAvatarETag`, `TestIfNoneMatchSatisfied`, `TestIndividualText`, `TestRenderable`, `TestColorForSeedStable`, `TestRender*` all green. (The DB-backed `TestUserAvatar*` tests fail only on a local migration/env issue — `unknown migration 20260402000001_thread_legacy01.sql` — unrelated to this PR.)
- ✅ **ETag ↔ body consistency**: in every return path the served ETag matches the served bytes — `name-v1` for a successful nickname render (`api.go:586`), and `ascii-v1` reset before the fallback render (`api.go:599`) when nickname rendering fails. The 304 short-circuit (`api.go:589`) computes the ETag without rendering, correctly avoiding expensive re-encode on cache revalidation.
- ✅ **Fallback chain is sound**: empty/invisible-only nicknames and glyphs the font can't render (pure emoji) flow to `Renderable()==false` → legacy `generateDefaultAvatar` ASCII path. No tofu/broken images. Existing users with blank names in the DB are unaffected (validation is write-time only; they render the ASCII fallback).
- ✅ **License compliance**: `OFL.txt` (SIL OFL 1.1) and `README.md` (source URL, subset ranges, no Reserved Font Name) are included; subsetting and redistribution under the original name are permitted. Embedding the 12.8 MB subset keeps the binary cgo-free and the Docker build unchanged.
- ✅ **`ValidateName` hardening is safe**: blank-name rejection (whitespace / control / zero-width / BOM) is added to the shared write guard used by register, rename, and admin create/update (`api_usernamelogin.go:67`, `api_emaillogin.go:165`, `api_manager.go:400/1052`, `api.go:1002/3634`). Third-party login does not call `ValidateName`, so empty third-party nicknames still log in. Pre-existing `== ""` / `TrimSpace == ""` checks in front of these callers make this strictly additive (now also catches full-width space / ZWSP).

## Findings

### P2 — Re-render on every cache miss (performance)
`UserAvatar` renders an 800×800 (200×4 supersample) image + CatmullRom downscale + PNG encode on **every** non-304 request (`api.go:592`). The ETag/304 path mitigates warm clients, but cold/shared-cache clients and the `must-revalidate` re-fetch after 300s each pay the full render cost. The avatar endpoint is public and high-traffic, so under load this is meaningfully more CPU-intensive than the old solid-circle generator. Consider a small in-process LRU keyed by `(uid, text)` (or the ETag) caching the encoded PNG bytes. Non-blocking — the prior implementation also rendered per-request — but worth a follow-up.

### P2 — Unauthenticated endpoint now exposes nickname-derived content (privacy)
`GET /v1/users/:uid/avatar` is registered in the public `v` group, not `auth` (`api.go:296`). The default avatar now embeds the last two visible characters of the user's nickname, so an unauthenticated caller who knows/guesses a `uid` can read a fragment of the display name that previously was not derivable from the avatar (which was a `uid`-seeded color only). Nicknames are generally public display data within the IM, so severity is low, but it's a behavior change worth a conscious sign-off rather than an accident.

### P2 — Stale-ETag inefficiency in the persistent-render-failure edge case
If nickname rendering fails repeatedly, the 304 check at `api.go:589` always compares against the `name-v1` ETag, while the body and final ETag are `ascii-v1`. A client holding the `ascii-v1` ETag will never get a 304 and will re-download the identical ASCII fallback each time. This only triggers when `avatarrender.Render` fails persistently (effectively never, since the font is embedded and parsed once), so it's a theoretical nit, not a real concern.

### Nit — "initials" wording vs. last-two-chars behavior
`IndividualText` takes the **last two** visible runes (`text.go`), which is correct for CJK given-names (e.g. 张三丰 → 三丰) but yields trailing letters for Latin names (Alexander → "er"), not the conventional "initials" the PR title implies. This matches the design spec's CJK-first intent, so it's a deliberate choice — flagging only so the title/term is understood as "last two characters," not first-letter initials.

## Summary
Clean separation of concerns, thorough table-driven tests, correct cache semantics, and careful fallback handling. Nothing blocks merge. The P2 render-caching suggestion is the most worthwhile follow-up if avatar traffic is significant.

Jerry-Xin (approved):
Project relevance: in scope. This PR modifies `octo-server` user avatar rendering and user-name validation.

One-line summary: No blocking issues found; the implementation is coherent, tested at the package/unit level, and fits the requested avatar behavior.

💬 Non-blocking

🟡 Warning: `avatarETag` uses CRC32 for cache validation in [modules/user/avatar_path.go](/Users/mlamp/.openclaw/tmp/pr-review-m9ekyrj5/repo/modules/user/avatar_path.go:23). This is not a security hash, and collisions are feasible enough that a user-controlled rename could theoretically keep the same ETag and preserve a stale avatar. Consider using SHA-256 truncated to 64 or 128 bits for stronger cache invalidation.

🔵 Suggestion: `etagOpaqueTag` accepts malformed unquoted tags because it trims quotes rather than validating entity-tag syntax in [modules/user/avatar_path.go](/Users/mlamp/.openclaw/tmp/pr-review-m9ekyrj5/repo/modules/user/avatar_path.go:51). This is unlikely to cause practical harm, but stricter parsing would better match HTTP semantics.

✅ Highlights

The avatar route correctly computes the ETag before rendering and returns `304` before doing PNG work in [modules/user/api.go](/Users/mlamp/.openclaw/tmp/pr-review-m9ekyrj5/repo/modules/user/api.go:576).

The renderer package is cleanly isolated, uses embedded font data, and guards missing glyphs before rendering in [pkg/avatarrender/render.go](/Users/mlamp/.openclaw/tmp/pr-review-m9ekyrj5/repo/pkg/avatarrender/render.go:41).

Blank-name validation now covers whitespace, control characters, and Unicode format characters in [modules/user/api.go](/Users/mlamp/.openclaw/tmp/pr-review-m9ekyrj5/repo/modules/user/api.go:3879).

Verification performed:
`go test ./pkg/avatarrender/` passed.
`go test ./modules/user/ -run 'TestValidateName|TestIsBlankName|TestAvatarETag'` passed.
The broader route test run was blocked by the local test DB migration state: `unknown migration in database` for `20260402000001_thread_legacy01.sql`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#346 Review Report
**PR:** feat(user): server-rendered default avatar with nickname initials
**Head SHA:** `579502dccac56446c8dd5dcf322d6528690f4c88`
**Files:** 13 changed (+774 / -6)

---

## 1. Verification Summary

| Area | Status | Evidence |
|---|---|---|
| Avatar rendering pipeline | ✅ | `pkg/avatarrender/render.go` — supersampled circle + text → PNG, deterministic, concurrent-safe (local `sfnt.Buffer`, `sync.Once` font load) |
| ETag / HTTP caching | ✅ | `modules/user/avatar_path.go` — weak ETag via crc32, RFC 7232 §3.2 weak comparison, 304 path carries ETag + Cache-Control (Date added by net/http) |
| Fallback chain | ✅ | `api.go:554-610` — nameMode → Render → on failure override ETag to ascii-v1 → generateDefaultAvatar. `c.Header` uses `http.Header.Set` (replaces), verified correct |
| ValidateName blank check | ✅ | `api.go:3869-3906` — `isBlankName` strips Space/Cc/Cf, closes real gaps on email + username registration paths |
| Font licensing | ✅ | `pkg/avatarrender/fonts/OFL.txt` — SIL OFL 1.1 included, README documents source and subsetting |
| Test coverage | ✅ | Unit tests cover rendering, palette stability, text truncation, ETag determinism, 304 handling, blank-name rejection |

---

## 2. Findings

### P2-1: 500 response on double-render-failure leaks Cache-Control headers

**Diff-scope:** new (introduced by this PR's `setAvatarHeaders` closure)

**Location:** `modules/user/api.go:605-608` (the `generateDefaultAvatar` error path)

If both `avatarrender.Render()` fails (nameMode=true) AND `generateDefaultAvatar()` fails, the handler returns 500. But `setAvatarHeaders` was already called earlier, so the 500 response carries `Cache-Control: public, max-age=300, must-revalidate` and an ETag. Per RFC 7231 §6.6.1, 5xx responses with explicit freshness info are heuristically cacheable — a strict intermediary could cache this 500 for 5 minutes.

**Suggested fix:** Before `c.Writer.WriteHeader(http.StatusInternalServerError)`, add `c.Header("Cache-Control", "no-store")` to prevent caching the error response.

**Severity:** P2/nit — extremely narrow edge case (both renderers must fail), self-correcting within 5 minutes.

### P2-2: `etagOpaqueTag` uses `strings.Trim` instead of single-pair strip

**Diff-scope:** new

**Location:** `modules/user/avatar_path.go:53`

`strings.Trim(s, `"`)` strips ALL leading/trailing `"` characters, not just one pair. For well-formed ETags (`W/"abc12345"`) this is correct. For malformed input like `""abc""` it would over-strip. `strings.TrimPrefix(s, "\"")` + `strings.TrimSuffix(s, "\"")` would be more precise.

**Severity:** P2/nit — ETags are always single-quoted per RFC 7232; no realistic impact.

---

## 3. Suggestions

1. **Clear cache headers on 500 path** (P2-1 above): one-liner `c.Header("Cache-Control", "no-store")` before the 500 write.
2. **Consider `TrimPrefix`/`TrimSuffix`** for `etagOpaqueTag` (P2-2 above): more precise, same complexity.

---

## 4. Additional Observations

### Pre-existing: WeChat login path has zero name sanitization (NOT this PR's scope)

The WeChat login path (`api.go:1302,1338`) passes the WeChat nickname directly into `createUserModel{Name: nickname}` without calling `ValidateName`, without `@` replacement, without blank check. This is a pre-existing token-cache role-injection vector (`uid@name@role` delimiter). The PR correctly scopes itself to `ValidateName` callers and explicitly notes third-party login bypasses it. This should be tracked as a separate follow-up.

### `isBlankName` / `IndividualText.isInvisible` duplication

Both functions check `unicode.IsSpace || unicode.Is(Cc) || unicode.Is(Cf)` — identical logic in `modules/user/api.go` and `pkg/avatarrender/text.go`. This is acceptable (different packages, different purposes), but if the definition of "invisible" evolves, both must be updated. A comment cross-referencing the other would help.

### Bot default avatar path still uses `max-age=86400` without ETag

The `shouldUseBotDefaultAvatar` branch (`api.go:522-531`) retains the old 24-hour cache without ETag. Pre-existing behavior, not in scope, but a natural follow-up for cache consistency.

### Binary size impact

The embedded font adds ~12.8 MB to the binary. Documented and intentional. The subset drops CJK Extension A/B+ and other scripts, retaining full CJK basic block + kana + hangul + Latin.

---

## 5. Data-Flow Trace

|---|---|---|
| `userInfo.Name` | `u.db.QueryByUID(uid)` → `SELECT * FROM user WHERE uid=?` → `Model.Name` | ✅ Yes. Can be empty for users registered with `RegisterUserMustCompleteInfoOn=1`; handled by `IndividualText("") → ""` → `Renderable("") → false` → ASCII fallback |
| `avatarrender.IndividualText(userInfo.Name)` | Strips invisible runes, takes last 2 | ✅ Correct. Empty input → empty output. Emoji preserved for caller to filter via `Renderable` |
| `avatarrender.Renderable(text)` | Checks each rune has glyph in embedded font (idx ≠ 0) | ✅ Correct. Local `sfnt.Buffer` → concurrent-safe. Emoji → `.notdef` → false → fallback |
| `avatarrender.ColorForSeed(uid)` | `crc32(uid) % 10` → palette index | ✅ Deterministic, stable per uid |
| `avatarETag("name-v1", uid, text)` | crc32 of null-separated parts | ✅ Changes on rename (text changes) and per-user (uid changes). Null separator prevents boundary collisions |
| ETag override on render failure | `c.Header("ETag", avatarETag("ascii-v1", uid))` | ✅ `http.Header.Set` replaces previous value. Final ETag matches actual ASCII content |
| `ifNoneMatchSatisfied` | Parses `If-None-Match` header, weak comparison via `etagOpaqueTag` | ✅ RFC 7232 compliant. Supports wildcard `*`, comma-separated list, W/ prefix stripping |
| `generateDefaultAvatar(uid)` (ASCII fallback) | `modules/user/avatar_gen.go:65` — deterministic circle + first-char-of-uid | ✅ Never fails for valid UIDs (only PNG encode error possible, practically impossible for 200x200 in-memory) |

---

## 6. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. This PR does not touch add/remove, subscribe/unsubscribe, or create/delete symmetric paths. The avatar endpoint is a read-only GET; the ValidateName change is a pure tightening (additive rejection) applied uniformly across all 6 existing call sites.

- **C2 — Control-flow ordering / nested reuse:** Clear. `isBlankName` is called once at the top of `ValidateName` before the `@` check — ordering is correct (blank check first, then delimiter check). `avatarrender.Render` is called once per request; no nested/double invocation. The `setAvatarHeaders` closure is called once; the ETag override on fallback correctly replaces via `http.Header.Set`.

- **C3 — Authorization boundary ≠ capability boundary:** N/A. No new endpoints, tools, or capabilities are exposed. The avatar endpoint remains a public GET. ValidateName tightening only rejects more inputs — it cannot be bypassed by any new path that didn't already bypass it.

---

## Verdict

No P0 or P1 findings. Two P2/nit items (500 cache headers, Trim precision) are minor robustness improvements that do not block landing. The rendering pipeline, ETag/caching semantics, fallback chain, and ValidateName tightening are all correct.

**[Octo-Q] verdict: APPROVE** — only P2/nit findings; zero误杀 confirmed.

yujiawei (approved):
# Code Review — PR #346 (octo-server)

**Verdict: APPROVED.** No correctness, security, or build-breaking issues. The default-avatar renderer, ETag/304 conditional-request handling, and blank-name validation are well-designed and well-tested. The notes below are non-blocking (P2/nit) follow-ups; none should hold up the merge.

## Summary of verification

I reviewed at head SHA `579502d` and verified locally:

- `go build ./pkg/avatarrender/` and `./modules/user/` — pass
- `go vet ./pkg/avatarrender/` — clean; `gofmt` — clean
- `go test ./pkg/avatarrender/` — pass
- `go test ./modules/user/ -run 'TestValidateName|TestIsBlankName|TestAvatarETag|TestIfNoneMatchSatisfied'` — pass

Things I scrutinized and found correct:

- **RFC 7232 weak ETag comparison.** `etagOpaqueTag()` strips `W/` and quotes on both sides, so weak/strong forms compare equal; multi-value comma lists and `*` wildcard are handled. The server only ever emits `W/"%08x"` (no commas), so `strings.Split` never corrupts its own tags.
- **304 vs render-failure consistency.** On the name path, if `avatarrender.Render` fails, the handler re-sets the ETag back to `ascii-v1` (`modules/user/api.go:600`) *before* serving the ASCII fallback. Because `c.Header()` only mutates the in-memory header map and nothing is flushed until `c.Data`, the served 200 always carries an ETag matching its own body — no within-response mismatch. The cross-request "stale name-v1 ETag" scenario is also unreachable: `Render` is deterministic, so a name-v1 image/ETag would never have been cached if rendering fails, and a held name-v1 ETag hits the 304 short-circuit at `api.go:584` anyway.
- **Glyph/emoji handling.** `Renderable` correctly treats `GlyphIndex==0` (`.notdef`) as unrenderable, routing emoji/tofu to the ASCII fallback. `IndividualText` strips Space/Cc/Cf before taking the last two runes; a truncated ZWJ-emoji half-sequence is unrenderable and falls back rather than drawing garbage.
- **Color stability.** `ColorForSeed` keys on `uid`, so the color is stable across renames (matches the documented intent).
- **Concurrency.** A fresh `opentype.NewFace` (own `sfnt.Buffer`) is allocated per request and `Renderable` uses a local buffer; per `golang.org/x/image` docs, `*sfnt.Font` methods are concurrency-safe with distinct buffers while a `Face` is not shared. The `sync.Once` font parse is the correct lazy-init idiom.
- **ValidateName blast radius.** 5 of 7 call sites already had a blank guard above the call; external/IdP login uses `sanitizeExternalName` and never calls `ValidateName`, so IdP login with a blank nickname is unaffected. The `@`-delimiter guard (for the `uid@name@role` token-cache key) is unchanged.
- **Header/cache changes are scoped** strictly to the local-generated default-avatar branch; bot, config-file, `DefaultBaseURL`-redirect, webhook, and system-UID branches are untouched. `c.Data` still sets `Content-Type: image/png`, so dropping the explicit header on the 200 path is safe. Group avatars are intentionally out of scope — confirmed no regression.
- **OFL 1.1 font bundling is compliant:** full license in `fonts/OFL.txt`, source + subset ranges documented in `fonts/README.md`; the bundled OFL has no Reserved Font Name, matching upstream `notofonts/noto-cjk`.

### 1. Expensive render is uncached on a public endpoint

`GET /v1/users/:uid/avatar` is registered on the unauthenticated `v` group (`modules/user/api.go:296`), and the new renderer is materially heavier than the legacy basicfont path (an 800×800 supersampled canvas, ~640k-pixel circle fill, `opentype.NewFace` per call, CatmullRom downscale, PNG encode — roughly ~20ms CPU and several MB allocated per render). The generated PNG is **not** memoized server-side, so every cold viewer, shared-cache miss, or `?v=` cache-buster pays the full render. The content ETag/304 only helps cooperating caches.

This is bounded — `main.go:180` mounts a global per-IP token-bucket DDoS floor (default 500 rps / 1000 burst) that covers this route, and the repo's own convention reserves the strict per-endpoint limiter for login/register/sms only, so the absence of a per-route limiter here is by design. Still, given the per-request cost jumped substantially, consider memoizing the default-avatar PNG keyed by `(uid, text, mode)` with a short TTL (e.g. an LRU / `sync.Map` with expiry) so repeated and fan-out requests hit a cache instead of re-rendering.

### 2. Embedded 13.4MB font is always resident

The subsetted Noto font (`pkg/avatarrender/fonts/NotoSansCJKsc-Bold-cjk-subset.otf`, 13,442,140 bytes) is `go:embed`-ed into a package-level `var fontData []byte` that is never released, and `opentype.Parse` aliases the same backing slice — so ~13.4MB stays resident for the process lifetime plus the parsed index structures, regardless of traffic. This is a deliberate, documented tradeoff for full CJK coverage and is reasonable; just flagging the fixed binary-size/RSS cost for deployers. If RSS becomes a concern, loading the font from a mounted file (instead of embedding) is an option.

### 3. Circle fill is the dominant in-render cost

`drawCircle` (`pkg/avatarrender/render.go`, the `drawCircle` loop) runs an 800×800 = 640k-iteration inner loop with a per-pixel `float64` distance test and bounds-checked `SetRGBA`. If you keep recomputing rather than adding the cache from suggestion 1, a cheap win is to compute the per-row x-span once (`xspan = sqrt(radiusSq - dy*dy)`) and fill that horizontal run, skipping the ~21% of out-of-circle pixels and the per-pixel test for interior pixels.

### 4. Registration now rejects blank `name` (intended, but client-visible)

`usernameRegister` (`api_usernamelogin.go:67`) and `emailRegister` (`api_emaillogin.go:165`) are the only two `ValidateName` call sites with no prior blank guard. Previously `ValidateName` only rejected `@`, so a register request with an empty/whitespace `name` succeeded and created a user with a blank nickname; now it returns a validation error. This is the PR's intended hardening and is reasonable — just worth a release note / client heads-up since it is a behavior tightening on the register path.

### 5. Test coverage gaps on the harder branches

The pure/deterministic pieces are well covered (ETag determinism + variation, the weak If-None-Match matrix, the blank-name matrix, `IndividualText`/`Renderable` edge cases, render determinism, and a real end-to-end ETag/304 integration test). The gaps are all on the branches the build/`-race` run does not exercise:

- The render-failure → ASCII fallback branch (incl. the `name-v1`→`ascii-v1` ETag rewrite) is untested. *(I verified the rewrite is correct — it's a coverage gap, not a bug.)* It needs an injectable seam (e.g. a render hook) to be testable.
- The ASCII-fallback mode (empty/emoji nickname) has no end-to-end ETag/304 integration test — `TestUserAvatarDefaultETag` only inserts renderable CJK users, so it always takes `nameMode=true`. Adding a `Name="😀😀"` (or empty) case asserting 200 → `If-None-Match` → 304 would close this.
- The PR notes `-race` passed, but every `avatarrender` test calls `Render`/`Renderable` sequentially, so `-race` proves nothing about concurrent rendering. A small `WaitGroup` test firing N concurrent renders under `-race` would actually exercise the shared-font path (the code looks safe; this just adds real confidence).

## Nits

- **ETag does not version the font asset.** `avatarETag("name-v1", uid, text)` fingerprints render-mode + uid + text but not the embedded font file. The comment claims the version prefix invalidates on render changes, but a future font re-subset/swap is an asset change that wouldn't bump `name-v1` — same ETag, different bytes. Self-heals within the 5-minute `must-revalidate` window, so impact is bounded. Consider bumping the prefix (or folding a font-asset version constant into the parts) whenever the font asset changes.
- **Mode classification untested at the HTTP layer.** The integration test treats the ETag as opaque (only asserts non-empty + differs between users); it never asserts a renderable user got a `name-v1` ETag, so a regression silently routing renderable nicknames into the ASCII branch would pass. Note the body-size check can't distinguish modes either — both render at 200×200. Asserting `etag == avatarETag("name-v1", uid, IndividualText(name))` would lock in mode selection.

Nice work overall — the conditional-request handling and the emoji/blank-name fallback design are thorough and correct.

PR #349, opened 2026-06-11
fix(avatar): opaque output, dynamic font size, and botfather branding

### LINKED ISSUE #350: Default avatar (#346) follow-ups: tiny font, transparent PNG, botfather rendered as nickname initials
Three regressions reported after the server-rendered default avatar shipped in #346:

1. **Font too small** — the fixed `10/32` em ratio leaves narrow Latin initials (e.g. `aw` / `TY`) visually tiny inside the circle.
2. **Transparent output** — the rendered PNG has a transparent ring outside the circle, inconsistent with the legacy ASCII fallback and the 13-color bot avatars (both opaque); it shows the underlying background on some clients.
3. **System bot mis-rendered** — `botfather` is excluded from the 13-color bot avatars yet has no dedicated static icon, so it falls through to nickname-initial rendering and is drawn as `"er"` (last two chars of `"BotFather"`).

Fixed in the linked PR: ink-box-driven dynamic font size, opaque white background, and a branded static avatar for `botfather` (extensible for future system bots).

an9xyz:
Closes #350

## Summary

Follow-ups to the server-rendered default avatar shipped in #346, addressing three regressions reported after rollout.

- **Dynamic font size** (`pkg/avatarrender`): size the font by the text's actual ink box instead of a fixed `10/32` em ratio. Narrow Latin initials (e.g. `aw` / `TY`) were rendered visually tiny under the fixed ratio; they are now scaled up to fill the circle, while CJK pairs stay near the original design ratio. Bounded by max ink-box width/height and a hard em cap so glyphs never overflow the circle.
- **Opaque output** (`pkg/avatarrender`): fill the canvas white before drawing the circle, so the PNG is opaque (`hasAlpha: no`) — matching the legacy ASCII fallback and the 13-color bot avatars instead of emitting a transparent ring. The all-opaque canvas lets `png.Encode` drop the alpha channel automatically.
- **System bot branding** (`modules/user`): serve a branded static avatar for the system bot `botfather` (embedded via `go:embed`, resolved before the DB lookup, same tier as `u_10000` / `fileHelper`). It previously fell through to nickname-initial rendering, drawing `"BotFather"` as `"er"`. The mapping is extensible for future system bots.
- **Cache invalidation**: bump the default-avatar ETag render version to `name-v3` so existing shared/browser caches revalidate to the new image instead of serving the old rendering for up to `max-age`.

System bots without a dedicated asset (e.g. `notification`) intentionally keep the existing fallback behavior — a separate follow-up.

## Test plan

- [x] `go test ./pkg/avatarrender/ -race`: dynamic font size (Latin ink width ≥ 38%, CJK stays in design range), opaque output (corners opaque white + image `Opaque()`), ink stays inside the circle across CJK / Latin / mixed / single-char.
- [x] `go test ./modules/user/ -run 'Avatar|SystemBot|ValidateName|NoLegacyResponse'`: system-bot branded routing (returned before DB lookup), bot default routing intact, ETag determinism / `If-None-Match` 304, legacy-response guard.
- [x] `go build ./...`, `go vet`, `gofmt` clean.
- [x] Visual spot-check of CN / Latin / single-char / mixed samples: opaque white background, fuller glyphs.


mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#349 Review Report
**PR:** [fix(avatar): opaque output, dynamic font size, and botfather branding](https://github.com/Mininglamp-OSS/octo-server/pull/349)
**Head SHA:** `f1236053175813b3fc9d95e20e06f36821b7587c`
**Files:** 6 changed (+327 / −7)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Opaque output (white fill before circle) | ✅ | `render.go:117` — `draw.Draw(canvas, ..., image.NewUniform(color.White), ..., draw.Src)` fills all pixels opaque before `drawCircle`. CatmullRom downscale with `xdraw.Over` on fully-opaque source produces fully-opaque output. `TestRenderOpaque` verifies corners + `Opaque()`. |
| Dynamic font size (`fitFontPx`) | ✅ | `render.go:158-185` — probe at 100px with `HintingNone`, measure ink bounds, linear scale to target box, cap at `maxFontEmRatio`. Math verified: `scale = min(s*0.64/inkW, s*0.42/inkH)`, result = `min(probePx*scale, s*0.46)`. Tests cover Latin (≥38% width), CJK (52-72%), single-char, mixed, and ink-inside-circle. |
| Botfather branding | ✅ | `system_bot_avatar.go` — `go:embed` + map lookup, placed at `api.go:503-511` after `u_10000`/`fileHelper` checks but before DB lookup. `TestSystemBotAvatarAssets` constrains keys to `IsSystemBot`, verifies 512×512 PNG. `TestUserAvatarHandlerSystemBotBranding` end-to-end verifies response body + Cache-Control. |
| ETag cache bust (v1→v3) | ✅ | `api.go:593` — `name-v3` invalidates all prior `name-v1` caches. No `name-v2` ever existed in codebase (confirmed via `git log -S`). Conservative and correct. |
| Test coverage | ✅ | 3 new test files: `fontsize_test.go` (163 lines, 5 test funcs), `system_bot_avatar_test.go` (62 lines, 4 test funcs). Coverage spans unit + integration. |

## 2. Findings

### No P0/P1 issues found.

### P2 — Silent error in `systemBotAvatar` on embed read failure

**File:** `modules/user/system_bot_avatar.go:29-31`
**Severity:** P2 (maintainability/observability)
**Diff-scope:** New code (this PR).

```go
data, err := systemBotAvatarFS.ReadFile(name)
if err != nil {
    return nil, false
}
```

If the embedded PNG is somehow unreadable at runtime (extremely rare for `go:embed`, but possible on memory corruption or future refactoring that breaks the embed path), the error is silently swallowed. The handler falls through to DB lookup + default avatar logic with no log line, making the root cause hard to diagnose in production.

**Suggestion:** Add a `log.Printf` or return the error alongside `ok=false` so the caller can log it. The `readBotDefaultAvatar` function in `bot_default_avatar.go` returns `(bytes, error)` — the caller in `api.go:548` logs on error. Matching that pattern here would be consistent.

**Not blocking:** The embed mechanism is compile-time verified; runtime failure is extremely unlikely. Graceful degradation (falling through to default avatar) is the correct behavior.

### P2 — `drawCircle` comment could clarify the 1px inset

**File:** `pkg/avatarrender/render.go:140`
**Severity:** P2 (nit / documentation)
**Diff-scope:** Pre-existing (comment updated in this PR, but the `radius = d/2 - 1` logic is unchanged).

The `drawCircle` function uses `radius := d/2 - 1`, leaving a 1px gap between the circle edge and the canvas boundary. After the opaque-fill change, this gap is now white instead of transparent. The comment says "圆外像素保持调用方预先铺好的底色" which is accurate, but the 1px inset is a subtle design choice that could benefit from a brief inline note (e.g., "1px inset avoids hard-clipping at the canvas edge during CatmullRom downscale").

**Not blocking:** Cosmetic documentation improvement.

## 3. Suggestions

1. **Observability:** Consider logging embed read failures in `systemBotAvatar` (see P2 above). One line of logging would save significant debugging time if the embed path ever breaks.
2. **Future extensibility:** The `systemBotAvatarFiles` map is a clean extension point. When adding more system bots (e.g., `notification`), the test `TestSystemBotAvatarAssets` will automatically validate the new entry. Good design.

## 4. Additional Observations

- **Handler ordering is correct:** `systemBotAvatar` intercepts before DB lookup, consistent with `u_10000`/`fileHelper` tier. Botfather's branded avatar always takes precedence, even if a custom avatar were uploaded to DB — this matches the PR description's intent.
- **`shouldUseBotDefaultAvatar` excludes system bots** (`!spacepkg.IsSystemBot(uid)`), so botfather would never reach the 13-color random avatar path even without the new `systemBotAvatar` check. The new check is needed to prevent the nickname-initial rendering path ("BotFather" → "er"), which is the actual bug being fixed.
- **Concurrency safety:** `fitFontPx` creates a local `opentype.Face` per call; the shared `parsedFont` is read-only after `sync.Once`. No race conditions.
- **PNG encoding optimization:** The opaque canvas causes `png.Encode` to automatically drop the alpha channel (encoding as RGB), reducing output size. This is a nice side benefit of the opaque-output change.

## 5. Data-Flow Trace

|---|---|---|
| `systemBotAvatar(uid)` return bytes | `go:embed assets/system_bot_avatar/*.png` → `embed.FS.ReadFile` | ✅ Compile-time verified embed; map key "botfather" matches file path. Test confirms 512×512 PNG. |
| `fitFontPx` ink bounds | `font.Drawer.BoundString(text)` at probePx=100, `HintingNone` | ✅ Fixed-point ÷ 64 = correct pixel conversion. Linear scaling assumption valid without hinting. |
| `drawCircle` pixel writes | Only sets pixels where `dx²+dy² ≤ radius²`; outside pixels retain pre-filled white | ✅ Correct — `draw.Draw(Src)` fills white first, `drawCircle` overwrites circle interior only. |
| CatmullRom downscale output | Source = fully opaque canvas (white + colored circle + white text) | ✅ `xdraw.Over` on opaque source → opaque output. Edge clamping produces white corners. |
| ETag `name-v3` | `avatarETag("name-v3", uid, text)` → crc32 fingerprint | ✅ Deterministic, sensitive to uid+text changes. Version bump invalidates all prior caches. |
| `systemBotAvatarFiles` keys | Must be in `spacepkg.SystemBots` | ✅ `TestSystemBotAvatarAssets` enforces via `require.Truef(t, spacepkg.IsSystemBot(uid), ...)`. |

## 6. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. This PR doesn't touch symmetric paths (add/remove, subscribe/unsubscribe). The avatar handler is a read-only GET endpoint with no state mutation.
- **C2 — Control-flow ordering / nesting reuse:** N/A. `fitFontPx` and `drawCircle` are each called once per render. No nested or reused control flow. The `draw.Draw` fill is a standard library call with no side effects beyond the canvas.
- **C3 — Authorization boundary:** N/A. The avatar endpoint is public (no auth required). The system bot avatar is a compile-time embedded static asset, not user-controlled. No new endpoints or capabilities are exposed.

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 issues. Two P2 observations (silent error handling, minor comment improvement) are non-blocking. The PR is well-structured with thorough test coverage, correct data-flow, and consistent patterns with existing code.

Jerry-Xin (approved):
The PR is in scope for octo-server and addresses avatar rendering/serving behavior with focused implementation and good regression coverage.

💬 Non-blocking

🔵 Suggestion — [modules/user/system_bot_avatar.go](/Users/mlamp/.openclaw/tmp/pr-review-hvoq601t/repo/modules/user/system_bot_avatar.go:24): `systemBotAvatar` silently falls back when an embedded configured asset cannot be read. The test catches this today, so it is not a blocker, but returning an error or logging in the caller would make future packaging mistakes easier to diagnose.

🔵 Suggestion — [modules/user/api.go](/Users/mlamp/.openclaw/tmp/pr-review-hvoq601t/repo/modules/user/api.go:504): the static branded avatar path follows the existing long-cache bot-default behavior, but does not provide ETag/304 support. That is consistent with nearby code, though adding validators later would reduce repeated transfer of the 266 KB embedded PNG.

✅ Highlights

- [pkg/avatarrender/render.go](/Users/mlamp/.openclaw/tmp/pr-review-hvoq601t/repo/pkg/avatarrender/render.go:117): filling the supersampled canvas before drawing correctly makes the rendered PNG opaque while preserving antialiasing after scaling.
- [pkg/avatarrender/render.go](/Users/mlamp/.openclaw/tmp/pr-review-hvoq601t/repo/pkg/avatarrender/render.go:163): dynamic sizing is bounded by measured glyph bounds plus hard caps, which addresses narrow Latin initials without obvious overflow risk.
- [modules/user/api.go](/Users/mlamp/.openclaw/tmp/pr-review-hvoq601t/repo/modules/user/api.go:500): serving `botfather` before the DB lookup matches the intended system-account behavior and is covered by an end-to-end test.
- [pkg/avatarrender/fontsize_test.go](/Users/mlamp/.openclaw/tmp/pr-review-hvoq601t/repo/pkg/avatarrender/fontsize_test.go:69): tests cover Latin sizing, CJK bounds, opacity, and ink containment.

Verification: `go test ./pkg/avatarrender/` passed. `go test ./modules/user/ -run 'Avatar|SystemBot'` could not complete in this local checkout because the test server panicked on an existing migration database state: `unknown migration in database` for `20260402000001_thread_legacy01.sql`.

yujiawei (approved):
# Code Review — PR #349 (octo-server)

**Verdict: APPROVED** — no blocking (P0/P1) issues. A few non-blocking nits below.

## Summary

Three related avatar fixes, all well-scoped:

1. **Opaque output** — `pkg/avatarrender/render.go` now paints a white background before drawing the colored circle, so the rendered PNG is fully opaque (no alpha channel). This matches the existing ASCII fallback and the 13-color bot avatars, preventing the underlying background from showing through on clients.
2. **Dynamic font size** — replaces the fixed `fontSizeRatio = 10/32` with `fitFontPx`, which measures the actual ink bounding box at a probe size and linearly scales the glyph into a target ink box (`maxInkWidthRatio` × `maxInkHeightRatio`), capped by `maxFontEmRatio`. This fixes Latin text rendering visually too small relative to CJK.
3. **Botfather branding** — adds a `systemBotAvatar` lookup that serves a fixed branded PNG for system bots (currently only `botfather`) before any DB lookup, so it neither gets a random 13-color avatar nor nickname-initial rendering ("BotFather" → "er").

## 1. Verification

- ✅ **Opacity is real.** `render.go:120-122` fills the canvas with `color.White` then draws the circle; alpha is uniformly 255 across the source, so the CatmullRom downscale (`render.go:131`, `xdraw.Over` onto a fresh dest) preserves opacity. `TestRenderOpaque` (`fontsize_test.go`) asserts opaque corners and whole-image `Opaque()`.
- ✅ **Dynamic sizing is geometrically safe.** With the width cap 0.64 and height cap 0.42, the worst-case ink-box corner sits at ~0.38·size from center, well inside the 0.5·size radius — the ink can't escape the circle. `TestInkStaysInsideCircle` and `assertInkInsideCircle` confirm this across CJK/Latin/mixed inputs. `fitFontPx` falls back to `baseFontEmRatio` on measurement failure (`render.go:172-181`).
- ✅ **System-bot branch ordering is correct.** `api.go:504-510` runs before the webhook-prefix check and the DB query, at the same priority tier as `u_10000`/`fileHelper`, so it never depends on a DB record. Unconfigured system bots return `ok=false` and fall through to the existing default logic — confirmed by `TestSystemBotAvatarUnknown` (including `notification`, which is a system bot with no branded asset).
- ✅ **Asset matches spec.** `botfather.png` is a valid 512×512 8-bit RGB PNG (same spec as the 13-color bot avatars). `TestSystemBotAvatarAssets` enforces that every key in `systemBotAvatarFiles` is a real `pkg/space.SystemBot`, is readable, and is 512×512 — a good guard against typos/missing assets.
- ✅ **Cache invalidation handled.** The `name-v1` → `name-v3` ETag bump (`api.go:593`) correctly busts caches for the changed render output (opaque + dynamic font).
- ✅ **Tests pass / vet clean.** `go test ./pkg/avatarrender/` passes; `go vet ./pkg/avatarrender/ ./modules/user/` is clean. Test coverage is strong: opacity, ink-in-circle, per-script sizing, determinism, asset validation, and an end-to-end handler test (`TestUserAvatarHandlerSystemBotBranding`).

## 2. Issues

None blocking (no P0/P1).

## 3. Suggestions (non-blocking nits)

- **N1 — Hinting mismatch in measurement vs. render.** `fitFontPx` measures with `HintingNone` (`render.go:170`) for clean linear scaling, but `drawCenteredText` renders with `HintingFull` (`render.go:192`). At small sizes, full hinting can shift glyph metrics slightly versus the unhinted measurement, so the final ink box may not exactly hit the target ratio. The `maxFontEmRatio` cap and the core-region margin keep this safe, and tests pass — so this is cosmetic only. Worth a one-line comment noting the intentional divergence.
- **N2 — ETag jumped `name-v1` → `name-v3`.** `v2` is skipped. If that's because a parallel in-flight branch consumed `v2`, fine; just confirm there's no leftover/stale `v2` cache expectation. Purely a housekeeping check — any value different from `v1` busts caches correctly.
- **N3 — Per-request copy of the branded asset.** `systemBotAvatar` calls `embed.FS.ReadFile` on every request, which copies the ~272 KB asset each time. This mirrors the existing `readBotDefaultAvatar` pattern, so it's consistent — but if system-bot avatar traffic is ever hot, memoizing the bytes once (the asset is immutable, embedded) would avoid the repeated allocation.
- **N4 — Static system-bot branch sets `max-age=86400` with no ETag.** `api.go:504-509` serves the branded PNG with a 1-day cache and no validator. This matches `u_10000`/`fileHelper` (which also write raw bytes without an ETag), so it's consistent. If you later swap the asset, clients won't revalidate for up to a day; a content-hash ETag would let them pick up replacements promptly. Optional.

## 4. Additional observations

- `notification` is a registered system bot (`pkg/space/query.go:24`) but has no branded asset, and `shouldUseBotDefaultAvatar` excludes system bots — so `notification` falls through to nickname/ASCII rendering. This is pre-existing behavior unchanged by this PR; flagging only for awareness in case branded assets are added later.
- The new helper file `system_bot_avatar.go` is a pure resource resolver with no error-response surface, so the i18n/legacy-response guard does not apply.

Overall: focused, well-documented, thoroughly tested. Approving.

PR #345, opened 2026-06-11
fix(thread): close subchannel blacklist read/write gates with ExistMemberActive (fast-follow #343)

yujiawei:
## What & why

Fast-follow to #343. #343 switched the **6 message-layer** subchannel (CommunityTopic) read/send gates from blacklist-permissive `ExistMember`/`ExistMembers` to active semantics `ExistMemberActive`/`existMembersActive` (`status=Normal AND is_deleted=0`). But the **thread module** and a few **message bypasses** are pre-existing baseline code that #343 never touched, so the same class of unauthorized read was not closed: a blacklisted user (`status=Blacklist`, `is_deleted=0`) could still pass these gates and read subchannel content.

Two independent reviewers (ReviewBot + Octo-Q) both flagged this residual on #343. This PR unifies the **entire** subchannel authorization surface to active semantics. **Separate PR, not folded into #343.**

## Fixed points (by severity)

**Highest — returns content body (true unauthorized read):**
1. `modules/thread/api.go` `threadMdGet` — returns subchannel `GROUP.md` document body.
2. `modules/message/api_channel_files.go` `channelFiles` — subchannel branch (file name/URL listing via IMSearchUserMessages).

**thread module reads/enumeration:**
3. `api.go` `listThreads`
4. `api.go` `getThread`
5. `api.go` `listMembers`
6. `api.go` `getThreadSimple` (shortID fallback route)

**thread module writes (fail-closed — blacklisted user must not create/join/change settings):**
7. `service.go` `CreateThread`
8. `service.go` `JoinThread`
9. `service.go` `UpdateSetting`

**message bypass:**
10. `modules/message/api.go` `authorizeMutualDelete` subchannel branch.

## How

- All subchannel (CommunityTopic) parent-group checks: `ExistMember` → `ExistMemberActive`.
- **Only CommunityTopic branches changed.** GROUP branches keep `ExistMember`/`ExistMembers` to avoid over-blocking normal members (#343 review confirmed GROUP branch must stay).
- `group.IService` gains an `ExistMemberActive` wrapper over the existing `DB.ExistMemberActive` helper (the interface previously exposed only `ExistMember`/`ExistMembers`).
- fail-closed throughout: errors / parse failures reject.

## Self-check

[...]

PR #343, opened 2026-06-11
fix(group): close remaining thread-subscription cleanup gaps + exclude blacklist from subscriber source (续 #27/#332)

yujiawei:
## 背景
续 #27/#332 子区订阅清理对称性。memberRemove / groupExit / handleOrgEmployeeExit 三条路径已在 #27/#332 修对，本单收口其余移除入口，并从数据源层排除 blacklist，避免被移除 / 被拉黑用户继续通过实时下发或历史读取接口访问子区内容。

## 改动（fail-closed 原则，只改 server）
- **订阅数据源排除 blacklist（根因收口）**：父群 + 子区的 IM Subscribers 数据源改用新增的 `GetSubscribableMemberUIDs`（status=normal AND is_deleted=0），不再返回被拉黑成员。`GetMembers` 语义保持不变，不影响其它调用方。Blacklist 数据源回调保持原样（仍挡发送），两者互补；订阅源重载时不会再把被拉黑用户加回。
- **拉黑/解除拉黑订阅对称**：拉黑 add 主动摘除每个 uid 的父群 + 子区订阅（复用统一 cleanup helper）；解除拉黑对称恢复订阅。
- **会话同步 / sidebar / space 过滤**：子区可见性补「调用者仍是父群成员」校验（fail-closed），覆盖 v1/v2 两条 space 过滤路径与 DB-only thread ext 行。
- **群解散 / 组织部门删人 / 代发 / 历史同步**：补齐子区频道销毁 + 订阅清理 / 父群成员鉴权，与已修路径对齐。解散提示消息降级为 best-effort，避免通知失败导致「半解散」。

## 测试
- 数据源排除（含 GetMembers 语义不变 / is_deleted 排除）
- 父群成员过滤（含 fail-closed、非法 channelID）
- 群解散 / 组织部门删人的子区清理
- 风格对齐 e2e_issue27_wukong_test.go / thread_cleanup_test.go
- `go build ./...`、`go vet`、group/message/thread 包测试全过。

## 自审（/review）
按 gstack review checklist 过 critical pass（SQL/Data Safety、Race、LLM trust boundary、Shell、Enum completeness）：无新增 finding。改动均为读路径鉴权收紧 + 订阅数据源过滤，无 SQL 拼接、无新枚举消费缺口；新增查询走参数化 dbr。改动规模：+682/-28，14 文件（含 4 个测试）。

review-tier: high-risk（auth/权限/跨端），请走独立 ReviewBot 审查单。

Fixes YUJ-4185

PR #355, opened 2026-06-11
fix(group): bots always follow their owner — no exit/kick role exceptions, blacklist cascades to owner's bots

### LINKED ISSUE #354: [P1] Bot must follow its owner: no creator/manager exception on exit/kick cascade + blacklist must cascade to owner's bots
Product decision by Yu (2026-06-11 12:27 UTC): **a user's bots always follow the user — no role exceptions.** When a user leaves or is removed from a group, every bot they created and invited into that group must leave too (including subchannel membership/subscription/pin cleanup).

The D-2 cascade (YUJ-49/#1186 + YUJ-52/#1189) already covers the normal-member exit/kick paths end-to-end. Two gaps remain:

**Gap 1 — creator/manager exceptions on the cascade**
- `modules/group/api.go` groupExit: cascade only runs `if loginMember.Role != MemberRoleCreator` — a group owner who leaves (after ownership transfer to newGrouper) keeps their bots in the group. Per the new product decision this exception must be removed: owner exit cascades exactly like a normal member (cascadeRemoveBotsInvitedByUIDTx + removeUserFromGroupThreads + pinned/conv_ext cleanup + system tip).
- `modules/group/service.go` RemoveGroupMembers: the `m.Role == MemberRoleCreator || m.Role == MemberRoleManager` skip should also drop the manager exception (creator cannot be kicked anyway) — a kicked manager's bots leave with them.

**Gap 2 — blacklist path has no cascade at all**
- `modules/group/api.go:3158` blacklist add only flips the user's own member status + IM blacklist. Their bots stay `status=Normal`, so a blacklisted user can keep reading group/subchannel content **through their own bot**, bypassing the entire ExistMemberActive hardening line (#343/#345). 
- Proposed: on blacklist, cascade-blacklist the user's in-group bots (`robot.creator_uid = uid`); on un-blacklist, symmetrically restore. (Symmetric + reversible; alternative of cascade-removal is lossy.)
- Closely related to #352 (bot_api permissive gate) — same attack surface; #352 is the immediate gate fix, this is the membership-state fix.

**Tests**
- Owner exits → their bots leave + thread_member/subscription/pinned cleaned.
- Manager kicked → same.
- User blacklisted → their bots can no longer pass ExistMemberActive (read/write subchannel denied); un-blacklist restores.

Refs: YUJ-49/#1186, YUJ-52/#1189, PR #343/#345 hardening line, #352.

yujiawei:
Fixes #354

Product decision: **a user's bots always follow the user — no role exceptions.** When a user leaves, is kicked, or is blacklisted, every bot they created and invited into the group must follow them. This closes the two gaps left after the D-2 cascade (#1186 / #1189).

## Gap 1 — creator/manager exceptions on the exit/kick cascade

- **`groupExit`**: removed the `loginMember.Role != MemberRoleCreator` guard. A group owner who exits (after ownership transfer to the second-oldest member) now cascades exactly like a normal member: bot members removed in the same transaction, plus the existing thread_member / IM subscription / pinned / conv_ext cleanup and the system tip.
  - New-owner selection now uses `QuerySecondOldestMemberExcludingBotsOf`, which excludes the leaving owner's own bots — otherwise a bot could be promoted to creator and cascade-deleted in the same transaction, leaving the group ownerless. Other users' bots keep the old selection semantics.
- **`RemoveGroupMembers`**: dropped the manager exception from the removable filter and from the per-member cascade guard. A kicked manager leaves together with their bots. Creator keeps the belt-and-suspenders exemption (cannot be kicked; API-level `memberRemove` already restricts manager removal to the owner).

## Gap 2 — blacklist had no cascade at all

Previously `blacklist add` only flipped the target user's member status + IM blacklist. Their bots stayed `status=Normal`, so a blacklisted user could keep reading group/subchannel content **through their own bot**, bypassing the entire `ExistMemberActive` hardening line (#343 / #345).

- On blacklist add/remove, the target list is expanded with the users' in-group bots (`robot.creator_uid` match, same fail-closed ownership semantics as the D-2 cascade: orphan/disabled bots are not anyone's bot).
- The bots then go through the same pipeline as the user: member status flip, IM blacklist add/remove, parent-group + subchannel subscription teardown (YUJ-4185 line), and symmetric restore on un-blacklist.
- The ownership query intentionally does not filter `group_member.status`, so un-blacklist finds the bots that are currently in `Blacklist` state and restores them.

## Tests

- `TestQuitGroup_CreatorCascadesBots` — owner exits → ownership transferred to a human (bot excluded from selection), owner's bot cascade-removed.
- `TestQuerySecondOldestMemberExcludingBotsOf_OnlyBotsLeft` — no new owner when only the leaver's bots remain; other users' bots unaffected.
- `TestRemoveGroupMembers_ManagerKickCascadesBots` — kicked manager + their bot both removed; creator still unkickable.
- `TestQueryBotUIDsOwnedByUIDs_BasicMatch` / `TestExpandBlacklistTargetsWithOwnedBots` — cascade target expansion semantics (ownership scoping, orphan/disabled exclusion, dedupe).
- `TestBlacklistCascade_BotBlockedByExistMemberActive` — blacklisted user's bot fails `ExistMemberActive` (subchannel read/write gate denies); un-blacklist symmetrically restores; bystanders untouched.
- Existing D-2 cascade suites (`bot_cascade_test.go`, `api_groupexit_cascade_thread_test.go`) pass unchanged except the obsolete creator-skip case, which now asserts the new product decision.

Full `modules/group`, `modules/bot_api`, `modules/thread`, `modules/message`, `modules/user`, `modules/robot`, `modules/webhook` suites pass locally.

### DIFF
--- a/modules/bot_api/group_member_remove_guard_test.go
+++ b/modules/bot_api/group_member_remove_guard_test.go
@@ -1,0 +1,141 @@
+package bot_api
+
+import (
+	"encoding/json"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// PR#355 review 守卫：bot_admin 与人类管理员同权，POST
+// /v1/bot/groups/:group_no/members/remove 不得移除群主/管理员。#354 把
+// service 层 RemoveGroupMembers 的 manager 豁免下沉移除后，目标角色校验由
+// 调用方负责——Web API memberRemove 有 creator/manager 守卫，这里验证 Bot
+// API 路径的对等守卫：
+//   - 目标含 manager → 403 cannot_remove_privileged，且整个请求拒绝（混合
+//     列表里的普通成员也不能被"顺带"移除）；
+//   - 目标含 creator → 403 cannot_remove_privileged；
+//   - 目标全为普通成员 → 正常移除（守卫不误伤）。
+
+const (
+	rmGuardGroupNo  = "g_rm_guard_1"
+	rmGuardBotID    = "bot_rm_guard"
+	rmGuardBotToken = "bf_rm_guard_token"
+	rmGuardCreator  = "u_rm_creator"
+	rmGuardManager  = "u_rm_manager"
+	rmGuardCommon   = "u_rm_common"
+)
+
+func setupRemoveGuardEnv(t *testing.T) http.Handler {
+	t.Helper()
+	s, ctx := testutil.NewTestServer()
+	require.NoError(t, testutil.CleanAllTables(ctx))
+
+	_, err := ctx.DB().InsertBySql(
+		"INSERT INTO robot (robot_id, status, creator_uid, bot_token) VALUES (?, 1, ?, ?)",
+		rmGuardBotID, rmGuardCreator, rmGuardBotToken).Exec()
+	require.NoError(t, err)
+
+	_, err = ctx.DB().InsertBySql(
+		"INSERT INTO `group` (group_no, name, status, version) VALUES (?, ?, 1, 1)",
+		rmGuardGroupNo, "remove guard group").Exec()
+	require.NoError(t, err)
+
+	// creator(role=1) / manager(role=2) / 普通成员(role=0) / bot 管理员(bot_admin=1)
+	for _, m := range []struct {
+		uid              string
+		role, robot, adm int
+	}{
+		{rmGuardCreator, 1, 0, 0},
+		{rmGuardManager, 2, 0, 0},
+		{rmGuardCommon, 0, 0, 0},
+		{rmGuardBotID, 0, 1, 1},
+	} {
+		_, err = ctx.DB().InsertBySql(
+			"INSERT INTO group_member (group_no, uid, role, robot, bot_admin, vercode, is_deleted, status, version) VALUES (?, ?, ?, ?, ?, ?, 0, 1, 1)",
+			rmGuardGroupNo, m.uid, m.role, m.robot, m.adm, util.GenerUUID()).Exec()
+		require.NoError(t, err)
+	}
+
+	return s.GetRoute()
+}
+
+func rmGuardIsActiveMember(t *testing.T, handler http.Handler, uid string) bool {
+	t.Helper()
+	w := doBot(handler, botReq(t, "GET", "/v1/bot/groups/"+rmGuardGroupNo+"/members", rmGuardBotToken, nil))
+	require.Equalf(t, http.StatusOK, w.Code, "list members body: %s", w.Body.String())
+	// GET /members 返回成员数组；uid 可能出现在响应的其他字段里，所以逐项
+	// 比对而不是对响应体做字符串包含判断。
+	var members []struct {
+		UID string `json:"uid"`
+	}
+	require.NoErrorf(t, json.Unmarshal(w.Body.Bytes(), &members), "list members body: %s", w.Body.String())
+	for _, m := range members {
+		if m.UID == uid {
+			return true
+		}
+	}
+	return false
+}
+
+func TestBotGroupMemberRemove_ManagerTargetForbidden(t *testing.T) {
+	handler := setupRemoveGuardEnv(t)
+
+	w := doBot(handler, botReq(t, "POST", "/v1/bot/groups/"+rmGuardGroupNo+"/members/remove", rmGuardBotToken,
+		map[string]interface{}{"members": []string{rmGuardManager}}))
+	assert.Equalf(t, http.StatusForbidden, w.Code, "body: %s", w.Body.String())
+	assert.Contains(t, w.Body.String(), "cannot be removed through the bot API")
+	assert.True(t, rmGuardIsActiveMember(t, handler, rmGuardManager), "manager must remain a member")
+}
+
+func TestBotGroupMemberRemove_CreatorTargetForbidden(t *testing.T) {
+	handler := setupRemoveGuardEnv(t)
+
+	w := doBot(handler, botReq(t, "POST", "/v1/bot/groups/"+rmGuardGroupNo+"/members/remove", rmGuardBotToken,
+		map[string]interface{}{"members": []string{rmGuardCreator}}))
+	assert.Equalf(t, http.StatusForbidden, w.Code, "body: %s", w.Body.String())
+	assert.Contains(t, w.Body.String(), "cannot be removed through the bot API")
+	assert.True(t, rmGuardIsActiveMember(t, handler, rmGuardCreator), "creator must remain a member")
+}
+
+func TestBotGroupMemberRemove_MixedListRejectedAtomically(t *testing.T) {
+	handler := setupRemoveGuardEnv(t)
+
+	// 混合列表（普通成员 + 管理员）→ 整个请求 403，普通成员也不能被顺带移除。
+	w := doBot(handler, botReq(t, "POST", "/v1/bot/groups/"+rmGuardGroupNo+"/members/remove", rmGuardBotToken,
+		map[string]interface{}{"members": []string{rmGuardCommon, rmGuardManager}}))
+	assert.Equalf(t, http.StatusForbidden, w.Code, "body: %s", w.Body.String())
+	assert.Contains(t, w.Body.String(), "cannot be removed through the bot API")
+	assert.True(t, rmGuardIsActiveMember(t, handler, rmGuardManager), "manager must remain a member")
+	assert.True(t, rmGuardIsActiveMember(t, handler, rmGuardCommon), "common member must not be removed when the request is rejected")
+}
+
+func TestBotGroupMemberRemove_ManagerTargetCaseVariantForbidden(t *testing.T) {
+	handler := setupRemoveGuardEnv(t)
+
+	// MySQL utf8mb4_*_ci collation 下 uid 匹配大小写不敏感：大小写变体在
+	// service 层仍会命中真实 manager 行，所以守卫必须按 DB 解析行的角色
+	// 拦截，而不是在 Go 里做大小写敏感的字符串比对（codex review P1 回归）。
+	w := doBot(handler, botReq(t, "POST", "/v1/bot/groups/"+rmGuardGroupNo+"/members/remove", rmGuardBotToken,
+		map[string]interface{}{"members": []string{strings.ToUpper(rmGuardManager)}}))
+	assert.Equalf(t, http.StatusForbidden, w.Code, "body: %s", w.Body.String())
+	assert.Contains(t, w.Body.String(), "cannot be removed through the bot API")
+	assert.True(t, rmGuardIsActiveMember(t, handler, rmGuardManager), "manager must remain a member")
+}
+
+func TestBotGroupMemberRemove_CommonTargetStillWorks(t *testing.T) {
+	handler := setupRemoveGuardEnv(t)
+
+	w := doBot(handler, botReq(t, "POST", "/v1/bot/groups/"+rmGuardGroupNo+"/members/remove", rmGuardBotToken,
+		map[string]interface{}{"members": []string{rmGuardCommon}}))
+	require.Equalf(t, http.StatusOK, w.Code, "body: %s", w.Body.String())
+	resp := decodeBody(t, w)
+	assert.Equal(t, true, resp["ok"])
+	assert.Equal(t, float64(1), resp["removed"])
+	assert.False(t, rmGuardIsActiveMember(t, handler, rmGuardCommon), "common member should be removed")
+}
--- a/modules/bot_api/groups.go
+++ b/modules/bot_api/groups.go
@@ -656,6 +656,35 @@ func (ba *BotAPI) botGroupMemberRemove(c *wkhttp.Context) {
 		return
 	}
 
+	// PR#355 review: bot_admin 与人类管理员同权，不能踢群主/管理员——对齐
+	// Web API memberRemove 的角色守卫（manager 不可移除 manager/creator）。
+	// #354 把 service 层 RemoveGroupMembers 的 manager 豁免去掉后，目标角色
+	// 校验完全由调用方负责；不在这里拦截的话，bot_admin=1 的 bot 可以越权
+	// 踢任意管理员并级联带走其 bot。
+	//
+	// 角色校验必须基于 DB 解析后的目标行：WHERE 条件与 service 层
+	// QueryMembersWithUids 完全一致，由 MySQL collation 决定 uid 匹配
+	// （utf8mb4_*_ci 大小写不敏感）。如果改在 Go 里做大小写敏感的字符串
+	// 比对，请求方用 uid 的大小写变体即可绕过守卫、却仍命中 service 层的
+	// 真实 manager 行。
+	var targetRows []struct {
+		UID  string `db:"uid"`
+		Role int    `db:"role"`
+	}
+	_, err = ba.db.session.Select("uid", "role").From("group_member").
+		Where("uid in ? and group_no=? and is_deleted=0", filteredMembers, groupNo).Load(&targetRows)
+	if err != nil {
+		ba.Error("query target members failed", zap.Error(err), zap.String("groupNo", groupNo))
+		httperr.ResponseErrorL(c, errcode.ErrBotAPIQueryFailed, nil, nil)
+		return
+	}
+	for _, row := range targetRows {
+		if row.Role != group.MemberRoleCommon {
+			httperr.ResponseErrorLWithStatus(c, errcode.ErrBotAPICannotRemovePrivileged, nil, i18n.Details{"uid": row.UID})
+			return
+		}
+	}
+
 	removeResp, err := ba.groupService.RemoveGroupMembers(&group.RemoveGroupMembersServiceReq{
 		GroupNo:      groupNo,
 		Members:      filteredMembers,
--- a/modules/botfather/api_bot_group_test.go
+++ b/modules/botfather/api_bot_group_test.go
@@ -351,9 +351,14 @@ func TestBotGroupMemberRemove_CannotRemoveCreator(t *testing.T) {
 		"members": []string{testutil.UID},
 	}))
 
-	assert.Equal(t, http.StatusOK, w.Code)
-	result := jsonResult(t, w)
-	assert.Equal(t, float64(0), result["removed"])
+	// PR#355 review (Jerry-Xin): the bot member-remove role guard now rejects
+	// the request outright instead of silently skipping the creator. This
+	// mirrors the Web API memberRemove rule (a manager cannot kick the
+	// creator/managers) — see modules/bot_api/groups.go and the matching
+	// err.server.bot_api.cannot_remove_privileged 403 in
+	// modules/bot_api/group_member_remove_guard_test.go.
+	assert.Equalf(t, http.StatusForbidden, w.Code, "body: %s", w.Body.String())
+	assert.Contains(t, w.Body.String(), "cannot be removed through the bot API")
 }
 
 func TestBotGroupMemberRemove_NotBotAdmin(t *testing.T) {
--- a/modules/group/api.go
+++ b/modules/group/api.go
@@ -2833,8 +2833,9 @@ func (g *Group) groupExit(c *wkhttp.Context) {
 	**/
 	var newGrouper *MemberModel // 新群主
 	if loginMember.Role == MemberRoleCreator {
-		// 查询第二老成员
-		newGrouper, err = g.db.QuerySecondOldestMember(groupNo)
+		// 查询第二老成员。#354：排除退群群主名下的 bot——它们会在下方被级联带走，
+		// 不能被选为新群主（否则新群主在同一事务内即被删除）。
+		newGrouper, err = g.db.QuerySecondOldestMemberExcludingBotsOf(groupNo, loginUID)
 		if err != nil {
 			g.Error("查询第二元老成员失败！", zap.Error(err))
 			httperr.ResponseErrorL(c, errcode.ErrGroupQueryFailed, nil, nil)
@@ -2896,21 +2897,19 @@ func (g *Group) groupExit(c *wkhttp.Context) {
 	}
 
 	// D-2 · 级联带走 inviter 拉入的 bot（YUJ-49 / Mininglamp-OSS/octo-server#1186）。
-	// Edge case: 群主退群时此逻辑不触发（角色转让由上方 newGrouper 兜底；
-	//            若确要销群应走 DismissGroup）——保持 bot 跟随新群主留在群中。
+	// #354 产品决策：bot 永远跟随其主人，无角色例外——群主退群（角色已由上方
+	// newGrouper 完成转让）同样级联带走自己名下的 bot，和普通成员一致。
 	var cascadedBotUsers []*user.Model
-	if loginMember.Role != MemberRoleCreator {
-		cascadedUIDs, cerr := cascadeRemoveBotsInvitedByUIDTx(g.db, g.ctx, groupNo, loginUID, tx)
-		if cerr != nil {
-			tx.Rollback()
-			g.Error("级联移除 bot 成员失败", zap.Error(cerr))
-			httperr.ResponseErrorL(c, errcode.ErrGroupStoreFailed, nil, nil)
-			return
-		}
-		for _, botUID := range cascadedUIDs {
-			botUser, _ := g.userDB.QueryByUID(botUID)
-			cascadedBotUsers = append(cascadedBotUsers, botUser)
-		}
+	cascadedUIDs, cerr := cascadeRemoveBotsInvitedByUIDTx(g.db, g.ctx, groupNo, loginUID, tx)
+	if cerr != nil {
+		tx.Rollback()
+		g.Error("级联移除 bot 成员失败", zap.Error(cerr))
+		httperr.ResponseErrorL(c, errcode.ErrGroupStoreFailed, nil, nil)
+		return
+	}
+	for _, botUID := range cascadedUIDs {
+		botUser, _ := g.userDB.QueryByUID(botUID)
+		cascadedBotUsers = append(cascadedBotUsers, botUser)
 	}
 
 	// 若退群者是外部成员且当前群是外部群，检查是否需要恢复普通群
@@ -3128,6 +3127,16 @@ func (g *Group) blacklist(c *wkhttp.Context) {
 		httperr.ResponseErrorL(c, errcode.ErrGroupManagerOnly, nil, nil)
 		return
 	}
+	// #354 · Bot 跟人走：拉黑/解除拉黑级联到目标用户名下在群的 bot
+	// （robot.creator_uid 命中）。旧行为只动用户本人，其 bot 仍 status=Normal，
+	// 被拉黑用户可经自己的 bot 旁路读群/子区内容，绕过 ExistMemberActive
+	// 加固线（#343/#345）。解除拉黑走同一扩展，对称恢复。
+	targetUIDs, err := expandBlacklistTargetsWithOwnedBots(g.db, groupNo, req.Uids)
+	if err != nil {
+		g.Error("查询拉黑级联 bot 失败", zap.Error(err))
+		httperr.ResponseErrorL(c, errcode.ErrGroupQueryFailed, nil, nil)
+		return
+	}
 	status := 0
 	if action == "add" {
 		status = int(common.GroupMemberStatusBlacklist)
@@ -3141,14 +3150,14 @@ func (g *Group) blacklist(c *wkhttp.Context) {
 		httperr.ResponseErrorL(c, errcode.ErrGroupStoreFailed, nil, nil)
 		return
 	}
-	err = g.db.updateMembersStatus(version, groupNo, status, req.Uids)
+	err = g.db.updateMembersStatus(version, groupNo, status, targetUIDs)
 	if err != nil {
 		g.Error("添加或移除群成员黑名单错误", zap.Error(err))
 		httperr.ResponseErrorL(c, errcode.ErrGroupStoreFailed, nil, nil)
 		return
 	}
 	if status == int(common.GroupMemberStatusBlacklist) {
-		err = g.setGroupBlacklist(groupNo, req.Uids, status == int(common.GroupMemberStatusBlacklist))
+		err = g.setGroupBlacklist(groupNo, targetUIDs, status == int(common.GroupMemberStatusBlacklist))
 		if err != nil {
 			g.Error("添加IM黑名单错误", zap.Error(err))
 			httperr.ResponseErrorL(c, errcode.ErrGroupStoreFailed, nil, nil)
@@ -3158,18 +3167,19 @@ func (g *Group) blacklist(c *wkhttp.Context) {
 		// 仅靠 setGroupBlacklist(IMBlacklistAdd) 只挡“发送”，不断“接收”——被拉黑者仍
 		// 通过 WuKongIM WS 收子区/父群实时消息（越权读 P0）。子区订阅复用 #27/#332 统一
 		// helper；父群订阅显式 IMRemoveSubscriber。best-effort：失败只记日志、不回滚拉黑。
-		for _, uid := range req.Uids {
+		// #354：级联拉黑的 bot 同样摘除订阅（targetUIDs 已含 bot）。
+		for _, uid := range targetUIDs {
 			g.removeUserFromGroupThreads(groupNo, uid, group.SpaceID)
 		}
 		if rmErr := g.ctx.IMRemoveSubscriber(&config.SubscriberRemoveReq{
 			ChannelID:   groupNo,
 			ChannelType: common.ChannelTypeGroup.Uint8(),
-			Subscribers: req.Uids,
+			Subscribers: targetUIDs,
 		}); rmErr != nil {
 			g.Error("拉黑摘除父群IM订阅失败", zap.Error(rmErr), zap.String("groupNo", groupNo))
 		}
 	} else {
-		members, err := g.db.QueryMembersWithUids(req.Uids, groupNo)
+		members, err := g.db.QueryMembersWithUids(targetUIDs, groupNo)
 		if err != nil {
 			g.Error("查询移除黑名单成员错误", zap.Error(err))
 			httperr.ResponseErrorL(c, errcode.ErrGroupQueryFailed, nil, nil)
@@ -3222,7 +3232,7 @@ func (g *Group) blacklist(c *wkhttp.Context) {
 			return
 		}
 	} else {
-		for _, uid := range req.Uids {
+		for _, uid := range targetUIDs {
 			// 发送群成员更新命令
 			err = g.ctx.SendCMD(config.MsgCMDReq{
 				ChannelID:   groupNo,
--- a/modules/group/api_groupexit_cascade_thread_test.go
+++ b/modules/group/api_groupexit_cascade_thread_test.go
@@ -77,7 +77,7 @@ func TestGroupExit_CascadeBot_AlsoRemovesFromThread(t *testing.T) {
 	f := New(s.ctx)
 	ensureThreadTables(t, f)
 
-	// --- 1. 用户：群主 owner、退群者 leaver（非群主，cascade 前置条件）、bot 本体 ---
+	// --- 1. 用户：群主 owner、退群者 leaver（普通成员场景；#354 起群主退群同样 cascade）、bot 本体 ---
 	insertTestUsers(t, userDB, "owner", "leaver")
 	botUID := "yuj52_bot_by_leaver"
 	require.NoError(t, userDB.Insert(&user.Model{UID: botUID, Name: "bot-by-leaver", ShortNo: "sn_" + botUID, Robot: 1}))
--- a/modules/group/blacklist_bot_cascade_test.go
+++ b/modules/group/blacklist_bot_cascade_test.go
@@ -1,0 +1,217 @@
+package group
+
+import (
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// --- #354 Gap 2 · 拉黑级联 bot ---
+//
+// blacklist HTTP handler 在本测试环境里会被 IMBlacklistAdd 的前置检查挡住（无 IM 服务），
+// 与 bot_cascade_test.go 中 QuitGroup_* 用例同因。此处按 handler 的真实序列在 DB 层组装：
+// expandBlacklistTargetsWithOwnedBots → updateMembersStatus，再用 ExistMemberActive
+// 验证门禁语义（#343/#345 加固线的同一查询）。
+
+// TestQueryBotUIDsOwnedByUIDs_BasicMatch 验证拉黑级联 SQL 的命中口径：
+// 只挑出「属于目标用户、robot.status=1、仍在群内」的 bot；孤儿/禁用/他人 bot 不命中。
+func TestQueryBotUIDsOwnedByUIDs_BasicMatch(t *testing.T) {
+	svc, userDB := setupServiceTest(t)
+	insertTestUsers(t, userDB, testutil.UID, "m1", "m2")
+
+	resp, err := svc.CreateGroup(&CreateGroupServiceReq{
+		Creator: testutil.UID,
+		Members: []string{"m1", "m2"},
+		Name:    "blacklist-cascade-query",
+	})
+	require.NoError(t, err)
+	groupNo := resp.GroupNo
+
+	s := svc.(*Service)
+	seedBotMember(t, s, groupNo, "bot_m1_x", "bm1x", "m1")
+	seedBotMember(t, s, groupNo, "bot_m1_y", "bm1y", "m1")
+	seedBotMember(t, s, groupNo, "bot_m2", "bm2", "m2")
+	seedBotMember(t, s, groupNo, "bot_orphan", "borphan", "")
+	// 禁用 bot：robot.status=0，不视为任何人的 bot
+	seedBotMember(t, s, groupNo, "bot_disabled", "bdis", "m1")
+	_, err = s.ctx.DB().UpdateBySql("UPDATE robot SET status=0 WHERE robot_id=?", "bot_disabled").Exec()
+	require.NoError(t, err)
+
+	uids, err := s.db.QueryBotUIDsOwnedByUIDs(groupNo, []string{"m1"})
+	require.NoError(t, err)
+	assert.ElementsMatch(t, []string{"bot_m1_x", "bot_m1_y"}, uids)
+
+	// 多 owner 批量
+	uids, err = s.db.QueryBotUIDsOwnedByUIDs(groupNo, []string{"m1", "m2"})
+	require.NoError(t, err)
+	assert.ElementsMatch(t, []string{"bot_m1_x", "bot_m1_y", "bot_m2"}, uids)
+
+	// 无 bot 的 owner / 空参数 → 空结果，不报错
+	uids, err = s.db.QueryBotUIDsOwnedByUIDs(groupNo, []string{"nobody"})
+	require.NoError(t, err)
+	assert.Empty(t, uids)
+	uids, err = s.db.QueryBotUIDsOwnedByUIDs("", []string{"m1"})
+	require.NoError(t, err)
+	assert.Nil(t, uids)
+	uids, err = s.db.QueryBotUIDsOwnedByUIDs(groupNo, nil)
+	require.NoError(t, err)
+	assert.Nil(t, uids)
+}
+
+// TestExpandBlacklistTargetsWithOwnedBots 验证目标扩展：用户本人 + 名下 bot，按原序去重；
+// 无 bot 时原样返回。
+func TestExpandBlacklistTargetsWithOwnedBots(t *testing.T) {
+	svc, userDB := setupServiceTest(t)
+	insertTestUsers(t, userDB, testutil.UID, "m1", "m2")
+
+	resp, err := svc.CreateGroup(&CreateGroupServiceReq{
+		Creator: testutil.UID,
+		Members: []string{"m1", "m2"},
+		Name:    "blacklist-cascade-expand",
+	})
+	require.NoError(t, err)
+	groupNo := resp.GroupNo
+
+	s := svc.(*Service)
+	seedBotMember(t, s, groupNo, "bot_m1", "bm1", "m1")
+
+	// m1 带 1 个 bot → 扩展为 [m1, bot_m1]
+	out, err := expandBlacklistTargetsWithOwnedBots(s.db, groupNo, []string{"m1"})
+	require.NoError(t, err)
+	assert.Equal(t, []string{"m1", "bot_m1"}, out)
+
+	// bot 已显式在目标列表里 → 去重，不重复出现
+	out, err = expandBlacklistTargetsWithOwnedBots(s.db, groupNo, []string{"m1", "bot_m1"})
+	require.NoError(t, err)
+	assert.Equal(t, []string{"m1", "bot_m1"}, out)
+
+	// 无 bot 的用户 → 原样返回
+	out, err = expandBlacklistTargetsWithOwnedBots(s.db, groupNo, []string{"m2"})
+	require.NoError(t, err)
+	assert.Equal(t, []string{"m2"}, out)
+}
+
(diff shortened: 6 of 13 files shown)


mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#355 Review Report
**PR:** Mininglamp-OSS/octo-server#355
**Head SHA:** `87b9fe640886dcc3103e615e0eca337a76b25ef4`
**Title:** fix(group): bots always follow their owner — no exit/kick role exceptions, blacklist cascades to owner's bots

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Gap 1a — Creator exit cascades bots | ✅ | `api.go:2897-2912`: `MemberRoleCreator` guard removed; cascade runs unconditionally |
| Gap 1a — New-owner selection excludes leaver's bots | ✅ | `db.go:793-808`: `QuerySecondOldestMemberExcludingBotsOf` LEFT JOIN anti-pattern correct |
| Gap 1b — Manager kick cascades bots | ✅ | `service.go:1573-1577`: manager filter removed; `service.go:1625`: cascade guard only skips creator |
| Gap 2 — Blacklist cascades to owner's bots | ✅ | `api.go:3132-3140`: `expandBlacklistTargetsWithOwnedBots` called before status flip |
| Gap 2 — Un-blacklist symmetric restore | ✅ | `api.go:3184-3222`: same `targetUIDs` used; `QueryBotUIDsOwnedByUIDs` intentionally doesn't filter status |
| Old `QuerySecondOldestMember` fully removed | ✅ | Grep: no live definition remains; only two doc-comment references |
| Test coverage | ✅ | 7 new/updated tests covering all three gaps + edge cases |
| Existing test suites not broken | ✅ | PR body confirms `modules/group`, `bot_api`, `thread`, `message`, `user`, `robot`, `webhook` pass |

## 2. Findings

### No P0 or P1 issues found.

### P2 — Blacklist path TOCTOU (minor, non-blocking)

**Diff-scope:** new (introduced by this PR's `expandBlacklistTargetsWithOwnedBots` call)

`expandBlacklistTargetsWithOwnedBots` (`bot_cascade.go:62`) runs a non-transactional `SELECT` to find the target user's bots, then `updateMembersStatus` (`db.go:251`) flips status in a separate statement. Between the two, a concurrent operation could theoretically create or delete a bot owned by the target user, causing the cascade to miss it (or include a now-deleted bot, which `updateMembersStatus` would silently no-op on since `is_deleted=0` wouldn't match).

**Risk assessment:** Very low. Blacklist is a manager-only, human-initiated action; concurrent bot creation during the narrow window is unlikely. The `updateMembersStatus` WHERE clause (`group_no=? and uid in ?`) silently skips non-matching rows, so a deleted bot is a no-op, not an error. Fail-open (missed bot) is the same as the pre-PR behavior.

**Suggestion:** No action needed for this PR. If the codebase ever moves blacklist into a transactional path, the expand query can be switched to `FOR UPDATE` trivially.

### P2 — `ForbiddenExpirTime` filter in un-blacklist may partially skip bot restore (pre-existing)

**Diff-scope:** pre-existing (the `ForbiddenExpirTime` filter at `api.go:3193-3197` predates this PR)

In the un-blacklist branch, `removeUIDs` is built by filtering out members with `ForbiddenExpirTime > 0`. The `updateMembersStatus` call above already flipped the bot's `status` to Normal using the full `targetUIDs`, but the IM blacklist removal, `IMAddSubscriber`, and `addUsersToGroupThreads` use the filtered `removeUIDs`. A muted bot would end up with `status=Normal` but still on the IM blacklist and without thread subscriptions.

**Risk assessment:** Extremely low. Bots are rarely individually muted. This is pre-existing behavior for all members (not specific to bots) and not amplified by this PR.

### Nit — `QueryBotsInvitedByUIDTx` parameter naming

**Diff-scope:** pre-existing

`QueryBotsInvitedByUIDTx` (`db.go:749`) names its parameter `inviterUID`, but the SQL matches on `r.creator_uid` (bot owner), not the user who invited the bot into the group. The new `QueryBotUIDsOwnedByUIDs` uses the more accurate `ownerUIDs`. Consider renaming the older function's parameter for consistency. Non-blocking.

## 3. Suggestions

- Consider adding a brief integration test that exercises the full `blacklist` HTTP handler (not just the DB-layer sequence) once the IM service mock is available, to cover the end-to-end `expandBlacklistTargetsWithOwnedBots → updateMembersStatus → setGroupBlacklist → removeUserFromGroupThreads → IMRemoveSubscriber` pipeline.
- The `QuerySecondOldestMemberExcludingBotsOf` doc comment references the deleted `QuerySecondOldestMember` — minor doc staleness that will become confusing over time. Consider rephrasing to be self-contained.

## 4. Additional Observations

- **Security improvement:** This PR closes a real bypass where a blacklisted user could continue reading group/subchannel content through their bot (which remained `status=Normal`). The fix is correctly targeted at `ExistMemberActive` — the same gate used by the #343/#345 hardening line.
- **Semantic alignment:** Both `QueryBotsInvitedByUIDTx` (exit/kick cascade) and `QueryBotUIDsOwnedByUIDs` (blacklist cascade) match on `r.creator_uid`, so they use the same ownership semantics despite different naming. This is correct.
- **Ownerless group edge case:** When the creator exits and only their bots remain, `QuerySecondOldestMemberExcludingBotsOf` returns nil → no role transfer → group becomes ownerless after cascade removes all bots. This is the correct product behavior (empty group, no orphaned bot-creator).

## 5. Data-Flow Trace

### Exit cascade (Gap 1a)
```
groupExit → loginMember.Role == Creator
  → QuerySecondOldestMemberExcludingBotsOf(groupNo, loginUID)
    → SQL: LEFT JOIN robot ON creator_uid=leaverUID, WHERE r.robot_id IS NULL
    → Returns: oldest non-bot-of-leaver member (or nil)
  → tx: UpdateMemberRoleTx(newGrouper) → DeleteMemberTx(leaver)
  → cascadeRemoveBotsInvitedByUIDTx(groupNo, loginUID, tx)
    → QueryBotsInvitedByUIDTx: SELECT gm.uid WHERE r.creator_uid=loginUID FOR UPDATE
    → For each bot: GenSeq + DeleteMemberTx (same tx)
  → tx.Commit()
  → Post-tx: IMRemoveSubscriber(botUIDs), removeUserFromGroupThreads(botUIDs),
    RemovePinnedForUserInSpace, RemoveConvExtForUserInSpace, sendBotCascadeRemovedTip
```
**Verdict:** Data flows correctly. New-owner selection excludes bots that will be cascade-deleted. All cleanup paths run for both leaver and cascaded bots.

### Kick cascade (Gap 1b)
```
RemoveGroupMembers → filter: skip IsDeleted/Creator (manager no longer skipped)
  → Per member tx: DeleteMemberTx → cascadeRemoveBotsInvitedByUIDTx
  → tx.Commit()
  → Post-tx: IMRemoveSubscriber, removeUserFromGroupThreads, pinned/conv_ext cleanup
```
**Verdict:** Manager is now treated like a normal member for cascade purposes. Creator remains exempt (cannot be kicked). Data flow verified.

### Blacklist cascade (Gap 2)
```
blacklist handler → expandBlacklistTargetsWithOwnedBots(groupNo, req.Uids)
  → QueryBotUIDsOwnedByUIDs: SELECT gm.uid WHERE r.creator_uid IN ? AND r.status=1
    (no status filter on group_member — finds Blacklist bots too for un-blacklist)
  → Dedup: original UIDs + bot UIDs, preserving order
→ updateMembersStatus(version, groupNo, status, targetUIDs)
→ [add branch]:
    setGroupBlacklist(targetUIDs, true) → IMBlacklistAdd
    removeUserFromGroupThreads(each targetUID)
    IMRemoveSubscriber(targetUIDs)
→ [remove branch]:
    QueryMembersWithUids(targetUIDs) → filter ForbiddenExpirTime==0 → removeUIDs
    setGroupBlacklist(removeUIDs, false) → IMBlacklistRemove
    IMAddSubscriber(removeUIDs)
    addUsersToGroupThreads(removeUIDs)
```
**Verdict:** Data flows correctly. Add and remove branches are symmetric. The `targetUIDs` expansion correctly includes bots in both directions.

## 6. Three-Blindspot Checklist (R5)

**C1 — Dual-path parity:** ✅ Clear
- add ↔ remove blacklist: both use `expandBlacklistTargetsWithOwnedBots` with the same `targetUIDs` — symmetric.
- exit ↔ kick: both use `cascadeRemoveBotsInvitedByUIDTx` — same function, same semantics.
- create-gate ↔ execute-path: blacklist handler's manager/creator auth check at `api.go:3117-3127` is unchanged; the cascade expansion runs after auth.

**C2 — Control-flow ordering / nesting reuse:** ✅ Clear
- `cascadeRemoveBotsInvitedByUIDTx` is called from `groupExit` and `RemoveGroupMembers` — both call sites pass the correct `loginUID`/member UID and the same `tx`. No double-cascade risk.
- `expandBlacklistTargetsWithOwnedBots` has a single call site (blacklist handler). No nesting risk.
- No security-sensitive regex or escape functions are modified.

**C3 — Authorization boundary ≠ capability boundary:** ✅ Clear
- Blacklist cascade: gated by manager/creator auth (pre-existing check at `api.go:3117-3127`). The cascade expansion doesn't expose new capabilities.
- Kick cascade: manager kick is gated by API-level `memberRemove` (only creator can kick managers). The cascade is a downstream effect, not a new endpoint.
- No new endpoints or tools are exposed by this PR.

---

**[Octo-Q] verdict: APPROVE**

This PR is well-structured, closes a real security bypass (blacklisted user reading group content via their bot), and extends the "bot follows owner" principle consistently across exit, kick, and blacklist paths. Test coverage is thorough, covering all three gaps plus edge cases (only-bots-left, disabled bots, orphan bots, bystander isolation). No P0/P1 issues found. Two P2 observations are non-blocking (one pre-existing, one very-low-risk TOCTOU).

Jerry-Xin (dismissed):
Summary: The PR is in scope for octo-server and the core bot-follow-owner behavior is well targeted, but it introduces a cross-API privilege escalation through the shared removal service.

🔴 Blocking

🔴 Critical — Bot admins can now kick group managers through the Bot API  
[modules/group/service.go](/Users/mlamp/.openclaw/tmp/pr-review-t477cv6z/repo/modules/group/service.go:1567) removes the service-level manager exemption, so `RemoveGroupMembers` now deletes manager-role members. The normal user API still has a target-role guard that prevents a manager from removing another manager or the owner at [modules/group/api.go](/Users/mlamp/.openclaw/tmp/pr-review-t477cv6z/repo/modules/group/api.go:2606), but [modules/bot_api/groups.go](/Users/mlamp/.openclaw/tmp/pr-review-t477cv6z/repo/modules/bot_api/groups.go:629) calls the same service after only checking `bot_admin`; it does not check the target members’ roles before calling `RemoveGroupMembers` at [modules/bot_api/groups.go](/Users/mlamp/.openclaw/tmp/pr-review-t477cv6z/repo/modules/bot_api/groups.go:659). Before this PR, the service filter protected that caller. After this change, a bot with `bot_admin=1` can remove a human group manager and cascade-remove that manager’s bots. Please add an equivalent role guard for the Bot API path, or move authorization into the service with an explicit privileged/admin override for callers that are allowed to remove managers.

💬 Non-blocking

🔵 Suggestion — Tighten the owner-exclusion join to actual bot members  
[modules/group/db.go](/Users/mlamp/.openclaw/tmp/pr-review-t477cv6z/repo/modules/group/db.go:797) excludes candidates based on an active `robot` row owned by the leaver, but it does not include `gm.robot = 1` in the join condition. The other cascade queries do require `group_member.robot = 1`. Adding that condition would better match the function name and avoid excluding a human candidate if stale or inconsistent `robot` data exists for the same UID.

✅ Highlights

The blacklist cascade query intentionally avoids filtering `group_member.status`, which is the right shape for symmetric un-blacklist restoration.

The new tests cover the main product decisions: owner exit, manager kick cascade, blacklist expansion, disabled/orphan bot exclusion, and `ExistMemberActive` gating.

I attempted the focused `go test ./modules/group -run ...`, but the local test environment failed during migration setup with `unknown migration in database`, before the PR tests could run.

yujiawei (commented):
# Code Review — PR #355 (octo-server)

Reviewed at head `87b9fe640886dcc3103e615e0eca337a76b25ef4` against `main` merge-base. Build (`go build ./modules/group/... ./modules/bot_api/...`) and `go vet ./modules/group/` both pass; the new tests compile but require a live MySQL test server, so they were not executed in this environment.

The product goal — *a user's bots always follow their owner, with no role exceptions* — is implemented cleanly in the cascade mechanics, the SQL helpers are sound, and the blacklist add/remove symmetry holds on the common path. However, two changes have side effects beyond the cascade itself that block merge.

## Verdict: CHANGES_REQUESTED

Two blocking issues (one security, one correctness), then non-blocking notes.

---

## P0 — Privilege escalation: a bot-admin bot can now kick human managers via the bot API

**Where:** `modules/group/service.go:1572-1576` (removable filter) and `:1626-1628` (per-member cascade guard); reachable through `modules/bot_api/groups.go:606-672` (`botGroupMemberRemove`).

This PR drops the `m.Role == MemberRoleManager` skip from both the removable filter and the cascade guard in `RemoveGroupMembers`:

```go
// service.go (after PR) — only creator is skipped now
for _, m := range targetMembers {
    if m.IsDeleted == 1 || m.Role == MemberRoleCreator { // MemberRoleManager skip removed
        continue
    }
    removableMembers = append(removableMembers, m)
}
```

`RemoveGroupMembers` performs **no operator-vs-target authorization** — `OperatorUID` is used only for naming/tips. Authorization is delegated entirely to each caller. There are two callers:

- **Web path** `memberRemove` (`api.go:2618-2628`): a manager-operator removing a manager → `ErrGroupCannotRemoveAdmin`; removing the creator → `ErrGroupCannotRemoveOwner`. Safe.
- **Bot API path** `botGroupMemberRemove` (`bot_api/groups.go:606-672`): the **only** gate is `IsBotAdmin` (`db.go:717-721`, `count(*) ... where bot_admin=1`). There is **no target-role check**.

Before this PR, the service-layer manager skip was the *sole* backstop protecting managers on the bot-API surface — the bot path never had its own role guard because the service silently dropped managers. Removing that skip opens an attacker-reachable escalation:

1. Any manager/creator grants `bot_admin` to a bot via `botAdminSet` (gated only by `QueryIsGroupManagerOrCreator`, `api.go:3658-3667`).
2. That bot calls `POST /v1/bot/groups/:group_no/members/remove` targeting a **peer human manager**.
3. `IsBotAdmin` passes; `RemoveGroupMembers` no longer skips managers → the peer manager is kicked, and their bots cascade away.

Net effect: a manager who **cannot** kick a fellow manager in the web UI can now do so indirectly by routing the kick through a bot they control. The PR description states "API-level `memberRemove` already restricts manager removal to the owner" — that is true only for the web path and does **not** cover the bot API. (Creator stays protected, so the escalation is bounded to managers and ordinary members.)

**Fix options:** add a target-role guard in `botGroupMemberRemove` (reject targets whose role is Manager/Creator, or restrict the bot to removing only its owner's own invitees), **or** thread an explicit manager-protection flag through `RemoveGroupMembers` that the bot-API caller sets. Do not treat `bot_admin` as manager-equivalent — it is a separate, lower-trust flag any manager can grant.

---

## P1 — Creator exiting a group of only {creator + creator's own bots} leaves an ownerless ghost group

**Where:** `modules/group/api.go:2834-2913` (`groupExit`); selection `modules/group/db.go:793-804` (`QuerySecondOldestMemberExcludingBotsOf`).

By design (and asserted by `TestQuerySecondOldestMemberExcludingBotsOf_OnlyBotsLeft`), when the only remaining members are the leaving creator and the creator's own bots, `newGrouper` is `nil`. The transaction then unconditionally:

- deletes the creator (`DeleteMemberTx`, `api.go:2891`), and
- cascade-removes the creator's bots (`cascadeRemoveBotsInvitedByUIDTx`, `api.go:2903`),

regardless of `newGrouper == nil` (only `UpdateMemberRoleTx` is gated on `newGrouper != nil`). `groupExit` never sets `GroupStatusDisband` (that constant is only *read* inside `groupExit` at `:2992/:3002`; it is written only by the dedicated disband handler at `api.go:215`).

Result: the `group` row survives as `status=Normal` with **zero active members and no owner**, and the `robot` rows are orphaned. This is a regression from pre-PR behavior: previously `QuerySecondOldestMember` could promote a creator's own bot to creator and the creator-skip guard kept that bot in the group, so the group stayed non-empty with an owner. The PR body acknowledges the ownerless outcome ("otherwise ... leaving the group ownerless") but the code does not then reap the empty group.

**Fix:** when `newGrouper == nil` and no human members remain, transition the group to `GroupStatusDisband` (or invoke the dismiss path) inside the same transaction instead of leaving an ownerless ghost.

---

## P2 — Non-blocking notes

1. **`RemoveGroupMembers` has no operator authorization at all** (`service.go:1538-1577`). The shared service entry point delegates 100% of "who may kick whom" to callers; the web path re-implements it, the bot path does not. Even after the P0 fix, consider making manager-vs-target authority a first-class parameter so future callers cannot silently inherit the gap.

2. **Blacklist handler has no creator/manager target protection** (`api.go:3119-3139`). The handler only checks that the *operator* is a manager; a manager can blacklist the creator or a fellow manager. This is pre-existing (not introduced here), but the cascade now also pulls their bots, widening the blast radius. Worth confirming it is intended.

3. **Un-blacklist skips IM/subscription restore for muted members** (`api.go:3153` flips status for all `targetUIDs`, but `:3192-3217` restores IM only for `ForbiddenExpirTime == 0`). A muted bot ends up `status=Normal` but still IM-blacklisted / un-subscribed. Appears intentional (a still-muted member shouldn't be re-subscribed), but the asymmetry between the DB status set and the IM restore set is subtle — a brief comment would help.

4. **Blacklist cascade is fully best-effort / non-transactional** (`api.go:3134-3180`). The DB status flip (`updateMembersStatus`, a bare UPDATE) and the IM/subscription side effects are not in one transaction; a mid-sequence failure can leave a bot member-blacklisted in DB but still IM-subscribed. Matches the existing pattern, flagging for awareness.

5. **Stale-row edge:** `QueryBotUIDsOwnedByUIDs` requires `gm.is_deleted = 0`, so a bot that left the group between blacklist-add and blacklist-remove keeps `status=Blacklist` on its deleted row. Inert (access gates on `is_deleted`), but noted.

### Test coverage gaps

6. **No automated test drives the real blacklist HTTP handler.** All three blacklist tests hand-assemble `expandBlacklistTargetsWithOwnedBots → updateMembersStatus` directly; the handler threads `targetUIDs` through five call sites and none are exercised end-to-end. The package already has fake-IM + `httptest` scaffolding (`api_i18n_regression_test.go`, `api_test.go`); the skipped `TestBlacklistRemoveUsesFilteredUIDs` could be repurposed. A revert of any single `req.Uids → targetUIDs` substitution would currently pass green.

7. **No test for the bot-API kick path** (`botGroupMemberRemove`) — the second caller of the changed service, and exactly the surface where the P0 lives. `TestRemoveGroupMembers_ManagerKickCascadesBots` covers the service in isolation only.

8. **`groupExit` creator-cascade tests hand-replay the transaction** rather than driving the handler (acknowledged constraint: `ctx.Event` panics in unit tests). Handler ordering/wiring for change #1 is validated only at the DB-helper level.

9. **No multi-target / mixed-ownership blacklist test** at the wrapper/handler level (the DB query *is* multi-owner tested). Low value, noting for completeness.

---

## Things verified as correct

- `QuerySecondOldestMemberExcludingBotsOf` correctly excludes only the leaver's own active bots (LEFT-JOIN anti-join) and preserves old selection semantics for others' bots; aligned with the cascade so a promoted owner can't be cascade-deleted.
- Old `QuerySecondOldestMember` has no remaining callers.
- `QueryBotUIDsOwnedByUIDs` JOIN/IN-slice binding, fail-closed ownership (orphan/disabled bots excluded), and the deliberate omission of a `group_member.status` filter (so un-blacklist re-finds blacklisted bots) are correct.
- Blacklist add/remove is symmetric on the normal (non-muted) path: bot fully restored (IM un-blacklisted + re-subscribed to parent + threads).

Jerry-Xin (changes requested):
Re-submitting REQUEST_CHANGES after deeper trace.

🔴 Blocking: Cross-API privilege escalation — bot admins can kick group managers via Bot API

The Web/User API path (`memberRemove` in `api.go:2596-2623`) correctly guards manager removal (only creator allowed). However, the Bot API path (`bot_api/groups.go:610-670`) only checks `IsBotAdmin` before calling `RemoveGroupMembers` — it does NOT check whether target members are managers or creators.

Since this PR removes the service-level manager exemption from `RemoveGroupMembers` (service.go:1574), a bot with `bot_admin=1` can now pass any manager UID to the Bot API remove endpoint and successfully kick that manager (plus cascade-remove their bots).

Fix options:
1. Add role guard in `bot_api/groups.go` remove handler: reject if any target is manager/creator (mirror the Web API check)
2. Or restore service-level manager exemption and add an explicit `allowManagerRemoval` flag for privileged callers

The rest of the PR (owner exit cascade, blacklist bot cascade, symmetric un-blacklist) is correct and well-tested.

yujiawei:
@Jerry-Xin Fixed in bda1d371 — went with **Option 1** (role guard in the Bot API remove handler, mirroring the Web API check).

**Guard** (`bot_api/groups.go` `botGroupMemberRemove`, after the self-filter and before `RemoveGroupMembers`):

- Resolves the target rows from `group_member` with the **same WHERE semantics as the service layer's `QueryMembersWithUids`** (`uid in ? and group_no=? and is_deleted=0`), then checks `role` on the resolved rows. Resolving via the DB (instead of comparing request strings in Go) matters: uid matching is decided by the MySQL `*_ci` collation, so a case-variant manager UID can't slip past a Go string comparison while still hitting the real manager row in the service.
- If **any** resolved target is the group creator or a manager → the whole request is rejected with `403 err.server.bot_api.cannot_remove_privileged` (offending uid in safe details). Rejection is atomic — common members in a mixed list are not removed alongside a denied privileged target, matching the Web API's all-or-nothing behavior.
- New error code registered in `pkg/errcode/bot_api.go` + zh-CN translation + i18n markers regenerated (`make i18n-extract-check` passes).

**Tests** (`modules/bot_api/group_member_remove_guard_test.go`, HTTP-level against the real handler/auth/service stack):

1. manager target → 403, manager stays
2. creator target → 403, creator stays
3. mixed list (common + manager) → 403, **neither** removed
4. case-variant manager uid (`U_RM_MANAGER`) → 403 (collation-bypass regression)
5. common-member target → 200, removed=1 (guard doesn't over-block)

`./modules/bot_api` and `./modules/group` suites pass; `go vet`, `i18n-extract-check`, and the i18n lint gates are green.

One note for transparency: the guard is a preflight check outside the removal transaction, which mirrors the existing Web API `memberRemove` pattern (same theoretical promote-after-check window). Kept it consistent with the Web path per Option 1; happy to follow up with a tx-level predicate if you want it tightened across both APIs.

Owner-exit cascade / blacklist bot cascade / symmetric restore are untouched.

Jerry-Xin (dismissed):
The PR is in scope for `octo-server`: it changes group membership, blacklist, bot ownership cascade, and bot API authorization behavior in the backend.

**💬 Non-blocking**
- 🟡 Warning: I could not complete `go test ./modules/group ./modules/bot_api` in this checkout because the local test database/migration state is stale (`unknown migration in database`, plus missing columns such as `bot_admin`, `forbidden_expir_time`, and `allow_no_mention`). This looks environmental rather than caused by the PR, but CI should be checked carefully.

**✅ Highlights**
- The new bot API guard in `modules/bot_api/groups.go:674` correctly checks target roles from DB rows before calling `RemoveGroupMembers`, which prevents a bot admin from reaching the now-less-restrictive service layer to remove group owners/managers.
- The creator-exit owner selection change in `modules/group/db.go:793` addresses the important edge case where the leaving owner’s bot could otherwise be promoted and then removed in the same transaction.
- The blacklist target expansion in `modules/group/api.go:3134` and `modules/group/db.go:773` matches the product decision and deliberately avoids filtering by member status, which is necessary for symmetric un-blacklist recovery.
- Test coverage covers the main security-sensitive cases: creator exit cascade, manager kick cascade, bot API privileged target rejection, blacklist bot expansion, and `ExistMemberActive` denial.

yujiawei (commented):
# Code Review — PR #355 (octo-server)

Reviewed independently at head `bda1d371c738ee71a94eec9fac3597209feb9e11` against `main` merge-base `791570b5`. `go build ./modules/group/... ./modules/bot_api/... ./pkg/errcode/...` and `go vet ./modules/group/ ./modules/bot_api/` both pass; the new test files compile (`go test -run XXX_NONE ...` green), though the DB-backed tests need a live MySQL test server to actually run.

The product goal — *a user's bots always follow their owner, with no role exceptions* — is implemented cleanly. The two security gaps surfaced in earlier rounds (bot-API privilege escalation; blacklist read-bypass via the user's own bot) are both closed, and the cascade SQL helpers are sound.

## Verdict: APPROVED

No P0/P1 blockers at this head. P2 notes below are non-blocking.

---

## Blocking issues from prior rounds — verified fixed at this head

### ✅ P0 (resolved) — bot-API privilege escalation closed

`modules/bot_api/groups.go:659-685` adds an explicit target-role guard in `botGroupMemberRemove`, run **before** `RemoveGroupMembers`: it loads `(uid, role)` for the requested targets and rejects the whole request with `ErrBotAPICannotRemovePrivileged` (HTTP 403) if any target's role is not `MemberRoleCommon`. This restores the manager/creator protection that the service-layer change (`service.go:1572-1576`, `:1626-1628`) removed.

Two things make this correct rather than superficial:

- **Same resolution semantics as the service.** The guard's `WHERE uid in ? and group_no=? and is_deleted=0` (`groups.go:674-675`) is byte-for-byte the predicate used by `QueryMembersWithUids` (`db.go:361`), which is how `RemoveGroupMembers` loads its own `targetMembers`. There is no skew (collation, `is_deleted`, IN-binding) between what the guard inspects and what the service acts on, so a target cannot present as Common to the guard yet resolve to a real manager row in the service.
- **Case-variant bypass is handled.** The guard relies on MySQL collation for uid matching (per the inline comment), and `TestBotGroupMemberRemove_ManagerTargetCaseVariantForbidden` (`group_member_remove_guard_test.go:118-129`) asserts an upper-cased manager uid is still rejected.

Test coverage is solid: manager-target, creator-target, mixed-list-rejected-atomically, case-variant, and common-still-works are all exercised against the real handler.

### ✅ Read-bypass via own bot — blacklist now cascades

`blacklist add/remove` now expands targets with the users' in-group bots via `expandBlacklistTargetsWithOwnedBots` → `QueryBotUIDsOwnedByUIDs` (`bot_cascade.go:54-87`, `db.go:771-804`), then threads the expanded `targetUIDs` through every downstream step (status flip, IM blacklist, parent + subchannel subscription teardown). This closes the gap where a blacklisted user kept reading group/subchannel content through their own `status=Normal` bot, bypassing the `ExistMemberActive` line (#343/#345). Ownership is fail-closed (orphan/disabled bots excluded via `robot.status=1`), and the deliberate omission of a `group_member.status` filter is correct so un-blacklist re-finds the bots currently in `Blacklist` state.

---

## P2 — Non-blocking notes

1. **Ownerless empty group on sole-creator-with-only-own-bots exit.** When a creator exits a group whose only other members are the creator's own bots, `QuerySecondOldestMemberExcludingBotsOf` returns nil (correct — a bot about to be cascade-removed must not be promoted), so no ownership transfer happens; the creator is deleted (`api.go:2891`) and the bots cascade away (`api.go:2903`), leaving a `group` row at `status=Normal` with zero members and no owner. `groupExit` never sets `GroupStatusDisband`. This is **not a regression unique to this PR** — `modules/group` has no disband-on-empty logic anywhere, so the sole-human-creator-exit case already produces the same empty-group state on `main`, and the bot's `robot` row is not orphaned (its `creator_uid` still points to the still-existing user who merely left this group). Pre-PR behavior happened to keep such groups non-empty only by promoting a bot to owner, which directly contradicts this PR's security goal. Reaping empty groups is a worthwhile follow-up but is a pre-existing platform gap, not a merge blocker for #355. Consider a separate issue to disband-on-empty inside `groupExit`.

2. **Un-blacklist subscription restore is gated on `ForbiddenExpirTime == 0`** (`api.go:3192-3197`), while blacklist-add tears down subscriptions for all targets. A bot that is both blacklisted and currently muted gets its DB status flipped back to `Normal` but stays IM-blacklisted / unsubscribed. This filter is **pre-existing** (not introduced by this PR — the PR only substituted `req.Uids → targetUIDs`) and is **fail-closed** (the member ends up more restricted, not less), so it is a minor DB-vs-IM consistency wart rather than a security or data-loss issue. A one-line comment explaining the intentional asymmetry would help future readers.

3. **`RemoveGroupMembers` still has no operator-vs-target authorization of its own** (`service.go:1538-1577`); it delegates entirely to callers. Both callers now guard correctly (web `memberRemove`, bot `botGroupMemberRemove`), but making manager-protection a first-class service parameter would prevent a future third caller from silently re-opening the gap.

4. **Blacklist cascade is best-effort / non-transactional** (`api.go:3134-3180`): the status UPDATE and the IM/subscription side effects are not in one transaction. Matches the existing pattern; flagging for awareness — a mid-sequence failure can leave a bot member-blacklisted in DB but still IM-subscribed.

### Test coverage observations (non-blocking)

- The blacklist tests drive `expandBlacklistTargetsWithOwnedBots → updateMembersStatus` directly rather than the full HTTP handler, so the five `req.Uids → targetUIDs` substitutions in the handler are not exercised end-to-end. An httptest-level test (the package already has fake-IM scaffolding) would lock in the wiring.
- `groupExit` creator-cascade tests hand-replay the transaction (acknowledged `ctx.Event` panic constraint), so handler ordering is validated only at the DB-helper level.

---

## Verified correct

- `QuerySecondOldestMemberExcludingBotsOf` LEFT-JOIN anti-join excludes only the leaver's own active bots and preserves old selection semantics for others' bots (`TestQuerySecondOldestMemberExcludingBotsOf_OnlyBotsLeft`).
- Old `QuerySecondOldestMember` removed with no remaining callers.
- `QueryBotUIDsOwnedByUIDs` JOIN/IN-slice binding, fail-closed ownership, and the intentional status-filter omission are correct.
- Blacklist add/remove is symmetric on the normal (non-muted) path.
- `errcode.ErrBotAPICannotRemovePrivileged` registered with `SafeDetailKeys: ["uid"]`; i18n markers added for both zh-CN and en-US.

## Round / convergence note

This is review round 5 on #355. Octo-Q approved, the cross-API escalation was caught and fixed in follow-up commit `bda1d371`, and the latest independent review (Jerry-Xin) already approved this exact head. My independent trace agrees: the blocking findings are resolved and only non-blocking P2 polish remains. The PR has converged — no further fix/re-review loop is warranted; the residual P2 items are better tracked as separate follow-ups.

Jerry-Xin (approved):
This PR is in scope for `octo-server` and implements the requested group/bot membership behavior without any blocking issues found.

**💬 Non-blocking**
- 🟡 Warning: `expandBlacklistTargetsWithOwnedBots` returns the original `uids` slice unchanged when no bots are found, so duplicate input is not deduped despite the function comment saying the result is deduped. This is likely harmless but inconsistent. See `modules/group/bot_cascade.go:67`.
- 🔵 Suggestion: Some new test comments include PR/review-process-specific wording that may age poorly in the codebase. Consider rewriting them as durable behavior comments. See `modules/bot_api/group_member_remove_guard_test.go:15` and `modules/bot_api/group_member_remove_guard_test.go:121`.

**✅ Highlights**
- The Bot API removal path now restores the missing privileged-target guard before calling the shared service. Good coverage for manager, creator, mixed-list atomicity, and case-variant matching in `modules/bot_api/group_member_remove_guard_test.go`.
- The creator exit path correctly avoids promoting the leaving owner’s active bots before cascading them out. See `modules/group/api.go:2838` and `modules/group/db.go:793`.
- Blacklist expansion reuses the same active ownership semantics and intentionally does not filter member status, which is important for symmetric un-blacklist behavior. See `modules/group/db.go:773`.

Test note: I attempted `go test ./modules/group ./modules/bot_api ./modules/botfather`, but the local test database/environment failed during migration setup and missing `OCTO_MASTER_KEY`; these look environmental rather than PR-specific.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#355 Review Report (automated review)
**PR:** Mininglamp-OSS/octo-server#355
**Head SHA:** `71b7ed11c6b873341313fb32b96d4f34cebe7f88`
**Base:** `main` (merge-base `791570b5`)
**Title:** fix(group): bots always follow their owner — no exit/kick role exceptions, blacklist cascades to owner's bots

**Build:** `go build ./modules/group/... ./modules/bot_api/... ./pkg/errcode/...` ✅ | `go vet ./modules/group/ ./modules/bot_api/` ✅

This is review round 6 on this PR. Prior rounds caught and fixed a P0 bot-API privilege escalation (commit `bda1d371`); the latest commit (`71b7ed11`) updates the botfather test to expect the new 403 behavior. My independent trace at this head agrees with the convergence: blocking findings are resolved, only non-blocking P2 polish remains.

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Gap 1a — Creator exit cascades bots (no role exception) | ✅ | `api.go:2897-2912`: `MemberRoleCreator` guard removed; `cascadeRemoveBotsInvitedByUIDTx` runs unconditionally |
| Gap 1a — New-owner selection excludes leaver's bots | ✅ | `db.go:793-808`: `QuerySecondOldestMemberExcludingBotsOf` LEFT-JOIN anti-join correct |
| Gap 1b — Manager kick cascades bots | ✅ | `service.go:1572-1576`: manager filter removed; `:1625-1628`: cascade guard only skips creator |
| Gap 2 — Blacklist cascades to owner's bots | ✅ | `api.go:3134-3140`: `expandBlacklistTargetsWithOwnedBots` called before status flip |
| Gap 2 — Un-blacklist symmetric restore | ✅ | `api.go:3184-3222`: same `targetUIDs`; `QueryBotUIDsOwnedByUIDs` intentionally omits status filter |
| P0 fix — Bot-API privilege escalation guard | ✅ | `bot_api/groups.go:659-685`: target role check before `RemoveGroupMembers`; rejects non-Common with 403 |
| Old `QuerySecondOldestMember` fully removed | ✅ | Only two doc-comment references remain (no live callers) |
| Test coverage | ✅ | 7 new/updated tests; botfather test updated to expect 403 |
| i18n | ✅ | `ErrBotAPICannotRemovePrivileged` registered with `SafeDetailKeys: ["uid"]`; zh-CN + en-US markers present |

## 2. Findings

### No P0 or P1 issues found at this head.

### P2 — Non-blocking notes

**P2-1: Ownerless empty group on sole-creator-with-only-own-bots exit**
- **Diff-scope:** new (amplified by this PR's bot-exclusion in owner selection)
- When creator exits a group whose only other members are the creator's own bots, `QuerySecondOldestMemberExcludingBotsOf` returns nil → no ownership transfer → creator deleted + bots cascade-removed → `group` row at `status=Normal` with zero members and no owner. `groupExit` never sets `GroupStatusDisband`.
- **Pre-existing platform gap:** `modules/group` has no disband-on-empty logic anywhere. Pre-PR, such groups stayed non-empty only by promoting a bot to owner — which directly contradicts this PR's security goal. Reaping empty groups is a worthwhile follow-up, not a merge blocker.
- **Suggestion:** File a separate issue for disband-on-empty inside `groupExit`.

**P2-2: Un-blacklist subscription restore gated on `ForbiddenExpirTime == 0`**
- **Diff-scope:** pre-existing (the filter at `api.go:3192-3197` predates this PR; PR only substituted `req.Uids → targetUIDs`)
- A muted bot gets DB status flipped to Normal but stays IM-blacklisted / unsubscribed. Fail-closed (more restricted, not less). Minor DB-vs-IM consistency wart.
- **Suggestion:** A one-line comment explaining the intentional asymmetry would help future readers.

**P2-3: `RemoveGroupMembers` has no operator-vs-target authorization**
- **Diff-scope:** pre-existing
- `service.go:1538-1577` delegates 100% of "who may kick whom" to callers. Both current callers (web `memberRemove`, bot `botGroupMemberRemove`) guard correctly. Making manager-protection a first-class service parameter would prevent a future third caller from silently re-opening the gap.

**P2-4: Blacklist cascade is best-effort / non-transactional**
- **Diff-scope:** pre-existing pattern (the PR follows it)
- `api.go:3134-3180`: status UPDATE and IM/subscription side effects are not in one transaction. Mid-sequence failure can leave a bot member-blacklisted in DB but still IM-subscribed. Matches existing pattern.

**P2-5: `updateMembersStatus` WHERE clause lacks `is_deleted=0`**
- **Diff-scope:** pre-existing
- `db.go:251-257`: `UPDATE group_member SET status=?, version=? WHERE group_no=? AND uid IN (?)` — no `is_deleted=0` guard. In practice harmless (upstream `expandBlacklistTargetsWithOwnedBots` only returns non-deleted bots), but a defensive `AND is_deleted=0` would prevent stale-row mutations.

### Test coverage observations (non-blocking)

- Blacklist tests drive `expandBlacklistTargetsWithOwnedBots → updateMembersStatus` directly, not the full HTTP handler. The five `req.Uids → targetUIDs` substitutions in the handler are not exercised end-to-end.
- `groupExit` creator-cascade tests hand-replay the transaction (acknowledged `ctx.Event` panic constraint).
- No test for `botGroupMemberRemove` via the real handler with a manager target — the P0 guard is tested in `group_member_remove_guard_test.go` but only at the handler level, not integration.

## 3. Data-Flow Trace

### Exit cascade (Gap 1a)
```
groupExit → loginMember.Role == Creator
  → QuerySecondOldestMemberExcludingBotsOf(groupNo, loginUID)
    → SQL: LEFT JOIN robot ON (robot_id=gm.uid AND status=1 AND creator_uid=leaverUID)
           WHERE r.robot_id IS NULL AND role<>Creator AND is_deleted=0
           ORDER BY created_at ASC LIMIT 1
    → Returns: oldest non-bot-of-leaver member (or nil)
  → tx: UpdateMemberRoleTx(newGrouper) [if non-nil] → DeleteMemberTx(leaver)
  → cascadeRemoveBotsInvitedByUIDTx(groupNo, loginUID, tx)
    → QueryBotsInvitedByUIDTx: SELECT gm.uid ... r.creator_uid=loginUID FOR UPDATE
    → For each bot: GenSeq + DeleteMemberTx (same tx)
  → tx.Commit()
  → Post-tx: IMRemoveSubscriber(botUIDs), removeUserFromGroupThreads,
    RemovePinnedForUserInSpace, RemoveConvExtForUserInSpace, sendBotCascadeRemovedTip
```
**Verdict:** ✅ Data flows correctly. New-owner selection excludes bots about to be cascade-deleted. All cleanup paths run for both leaver and cascaded bots.

### Kick cascade (Gap 1b)
```
RemoveGroupMembers → filter: skip IsDeleted/Creator (manager no longer skipped)
  → Per member tx: DeleteMemberTx → cascadeRemoveBotsInvitedByUIDTx
  → tx.Commit()
  → Post-tx: IMRemoveSubscriber, removeUserFromGroupThreads, pinned/conv_ext cleanup
```
**Verdict:** ✅ Manager treated like normal member for cascade. Creator remains exempt.

### Blacklist cascade (Gap 2)
```
blacklist handler → auth: QueryIsGroupManagerOrCreator (pre-existing)
  → expandBlacklistTargetsWithOwnedBots(groupNo, req.Uids)
    → QueryBotUIDsOwnedByUIDs: INNER JOIN robot ON (robot_id=gm.uid AND status=1)
      WHERE gm.group_no=? AND gm.robot=1 AND gm.is_deleted=0 AND r.creator_uid IN ?
      (no group_member.status filter — finds Blacklist bots too for un-blacklist)
    → Dedup: original UIDs + bot UIDs, preserving order
  → updateMembersStatus(version, groupNo, status, targetUIDs)
  → [add branch]:
      setGroupBlacklist(targetUIDs, true) → IMBlacklistAdd
      removeUserFromGroupThreads(each targetUID)
      IMRemoveSubscriber(targetUIDs)
  → [remove branch]:
      QueryMembersWithUids(targetUIDs) → filter ForbiddenExpirTime==0 → removeUIDs
      setGroupBlacklist(removeUIDs, false) → IMBlacklistRemove
      IMAddSubscriber(removeUIDs)
      addUsersToGroupThreads(removeUIDs)
```
**Verdict:** ✅ Add and remove branches are symmetric. `targetUIDs` expansion correctly includes bots in both directions.

### Bot-API guard (P0 fix from bda1d371)
```
botGroupMemberRemove → IsBotAdmin check
  → Query target (uid, role) FROM group_member WHERE uid IN ? AND group_no=? AND is_deleted=0
  → For each row: if role != MemberRoleCommon → 403 ErrBotAPICannotRemovePrivileged
  → Only if all targets are Common → call RemoveGroupMembers
```
**Verdict:** ✅ Guard WHERE matches service-layer `QueryMembersWithUids` predicate byte-for-byte. No skew possible. Case-variant bypass handled by MySQL collation (tested in `TestBotGroupMemberRemove_ManagerTargetCaseVariantForbidden`).

## 4. Three-Blindspot Checklist (R5)

**C1 — Dual-path parity:** ✅ Clear
- add ↔ remove blacklist: both use `expandBlacklistTargetsWithOwnedBots` with same `targetUIDs` — symmetric.
- exit ↔ kick: both use `cascadeRemoveBotsInvitedByUIDTx` — same function, same semantics.
- create-gate ↔ execute-path: blacklist handler's manager/creator auth at `api.go:3117-3127` unchanged; cascade expansion runs after auth.

**C2 — Control-flow ordering / nesting reuse:** ✅ Clear
- `cascadeRemoveBotsInvitedByUIDTx` called from `groupExit` and `RemoveGroupMembers` — both pass correct UID and same `tx`. No double-cascade risk.
- `expandBlacklistTargetsWithOwnedBots` has a single call site. No nesting.
- No security-sensitive regex or escape functions modified.

**C3 — Authorization boundary ≠ capability boundary:** ✅ Clear
- Bot-API guard closes the only new capability boundary opened by the service-layer manager-skip removal.
- Blacklist cascade doesn't expose new capabilities (gated by pre-existing manager auth).
- No new endpoints or tools exposed.

## 5. Verified Correct

- `QuerySecondOldestMemberExcludingBotsOf` LEFT-JOIN anti-join correctly excludes only the leaver's own active bots; preserves old selection semantics for others' bots.
- `QueryBotUIDsOwnedByUIDs` JOIN/IN-slice binding, fail-closed ownership (orphan/disabled bots excluded via `robot.status=1`), and intentional status-filter omission are correct.
- Blacklist add/remove is symmetric on the normal (non-muted) path.
- `ErrBotAPICannotRemovePrivileged` registered with correct HTTP 403 and safe detail keys.
- Botfather test correctly updated: creator-target now returns 403 (was 200/removed=0 before the guard was added).

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 issues at this head. The product goal — bots always follow their owner across exit, kick, and blacklist paths — is implemented cleanly. The P0 bot-API privilege escalation from earlier rounds is correctly fixed. The blacklist cascade closes a real security bypass (blacklisted user reading group content via their bot). Test coverage is thorough. Five P2 observations are non-blocking polish items better tracked as separate follow-ups.

This is round 6 on this PR. The blocking findings from rounds 1-2 were fixed in commit `bda1d371`, and the latest commit `71b7ed11` is a test-only update that correctly aligns the botfather assertion with the new guard behavior. The PR has converged — no further fix/re-review loop is warranted.

yujiawei (commented):
# Code Review — PR #355 (octo-server)

**Head SHA reviewed:** `71b7ed11c6b873341313fb32b96d4f34cebe7f88`
**Base:** `main` (merge-base `791570b5`)

## Summary

This PR implements the product decision *"bots always follow their owner — no role exceptions"* across three group surfaces:

1. **Group exit (`groupExit`)** — a leaving creator now cascades away their own bots (previously skipped for `MemberRoleCreator`); new-owner selection (`QuerySecondOldestMemberExcludingBotsOf`) excludes the leaver's bots so a soon-to-be-removed bot can't be promoted.
2. **Member removal (`service.RemoveGroupMembers`)** — the `manager` exemption is dropped; kicking a manager now cascades their bots too. Only `creator` stays exempt.
3. **Blacklist (`group.blacklist`)** — blacklist/un-blacklist now cascades to the target's owned in-group bots (`expandBlacklistTargetsWithOwnedBots`), closing a bypass where a blacklisted user could read group/thread content via their still-`Normal` bot (defeating the `ExistMemberActive` hardening from #343/#345).

The change is well-scoped, the SQL follows existing in-repo patterns (`bot_ownership.go`, `QueryBotsInvitedByUIDTx`), and test coverage is thorough.

## Verification Results

- ✅ `go build ./...` — clean
- ✅ `go vet ./modules/group/... ./modules/bot_api/...` — clean
- ⚠️ `go test ./modules/group/... ./modules/bot_api/...` — fails to start with `panic: Unable to create migration plan ... unknown migration in database`. This is a **test-environment/DB migration-state issue, not a code defect** — it occurs at `testutil.NewTestServer()` before any PR logic runs. CI with a clean migration baseline is expected to pass.

## Correctness — Key Checks

### ✅ Cross-file dependency correctly compensated (no blind spot)
Removing the `manager` exemption from `service.RemoveGroupMembers` means **target-role validation now belongs entirely to callers.** Both production callers were verified:
- **Web `memberRemove`** (`api.go:2606-2631`) already enforces "manager cannot remove manager/creator" before calling the service.
- **Bot API `botGroupMemberRemove`** (`groups.go:683-708`) adds a **new** guard `ErrBotAPICannotRemovePrivileged` mirroring that rule. Without this guard a `bot_admin` bot could kick any manager. The guard was correctly added in the same PR — no gap.

### ✅ Collation case-sensitivity bypass defended
The bot-API guard resolves target rows via a DB query whose WHERE clause (`uid in ? and group_no=? and is_deleted=0`) is byte-for-byte identical to the service layer's `QueryMembersWithUids`. Both rely on MySQL `utf8mb4_*_ci` matching, so a case-variant `uid` that would still hit a real manager row in the service cannot slip past the guard. Doing a case-sensitive Go string compare here would have been a real bypass; the chosen approach is correct and is explicitly regression-tested (`TestBotGroupMemberRemove_ManagerTargetCaseVariantForbidden`).

### ✅ Creator-exit orphan-owner edge case handled
`QuerySecondOldestMemberExcludingBotsOf` excludes the leaver's active bots from new-owner selection. The "only the leaver's own bots remain" case returns `nil` (no promotion), and other users' bots remain eligible (old semantics preserved). Covered by `TestQuerySecondOldestMemberExcludingBotsOf_OnlyBotsLeft`.

### ✅ Blacklist cascade symmetry & fail-closed
`expandBlacklistTargetsWithOwnedBots` dedupes in original order and is applied uniformly across `updateMembersStatus`, IM subscriber removal/restore, thread cleanup, and CMD fan-out — so add and remove are symmetric. Query failure returns an error that aborts the operation (fail-closed), appropriate for a permission-sensitive path. `QueryBotUIDsOwnedByUIDs` intentionally omits a `group_member.status` filter so already-blacklisted bots can be restored on un-blacklist — correct and documented.

### ✅ No dangling references / i18n complete
The removed `QuerySecondOldestMember` has no remaining callers (only doc/comment mentions). The new error key is present in both the runtime `active.zh-CN.toml` and the generated `i18nmarkers` en-US source; English falls back to `DefaultMessage`, consistent with the other `bot_api.*` keys.

## Suggestions (non-blocking, P2 / out of scope)

1. **Pre-existing blacklist target-role gap** (out of scope for this PR): `group.blacklist` checks only that the *operator* is manager/creator (`api.go:3119`), not the *target's* role — so a manager can blacklist the creator, and now (post-PR) cascade-blacklist the creator's bots. This is not introduced by this PR (the operator-only check predates it), but the new cascade slightly widens the blast radius. Worth a follow-up to add a target-role guard symmetric with `memberRemove`, if product intends managers cannot blacklist the creator.

2. **Minor**: the bot-API guard issues one extra `SELECT` per remove request. Negligible at current scale; no action needed.

## Conclusion

No P0/P1 issues. The PR correctly handles the cross-file contract shift (service-layer exemption removal compensated by caller guards), the collation bypass, and the orphan-owner edge case, with matching tests for each. Approving.

lml2468 (approved):
## Octo-PR APPROVE — #355 (head `71b7ed1`)

经维护者授权补第三票(author=yujiawei 占不了;Jerry-Xin+mochashanyao 已 ✅)。内容核实正确:

- ✅ **权限提升已修**(原 blocker):botGroupMemberRemove 调 RemoveGroupMembers 前加 target-role guard(目标 role!=Common 即拒 ErrBotAPICannotRemovePrivileged),且基于 DB 行校验防 uid 大小写绕过(与 service 层 collation 一致)。bot_admin 不能经 Bot API 越权踢人类 manager/creator。本 commit 补了配套测试对齐(期望 403)。
- ✅ **Gap1/Gap2 核心**:owner 退群级联(QuerySecondOldestMemberExcludingBotsOf 排 leaver bot 防 ownerless,同 tx 原子)+ blacklist 级联 owner 的 bot(堵"经自己 bot 旁路读子区"绕过 ExistMemberActive)+ 对称恢复,均正确。

### 已知 follow-up(三人收敛,不卡本 PR)
- 🟡 **无主幽灵群**:owner 退到只剩自己+bot 时(newGrouper==nil)无解散逻辑,留 status=normal 空壳群。触发极窄(无其他人类成员)、非越权。Jerry-Xin approve / mochashanyao 明确建议单独提 issue / 我裁断一致 → 降 follow-up,开 issue 追踪(修法:newGrouper==nil 走解散 status→Disband+IMDelChannel)。

无 blocking。
— Octo-PR(授权补票)

Issue #353, opened 2026-06-11
[P1][test] FollowThread/FollowChannel requireActive split is only covered by an integration-tagged test CI never runs


yujiawei:
From PR #345 review (non-blocking P1 test-coverage finding).

**Problem**
The `checkChannelAccess` `requireActive` split (`modules/message/1module.go:206-217`, AuthorizeThreadFollow=active vs AuthorizeChannelFollow=permissive) is exercised only by `modules/message/thread_follow_blacklist_e2e_test.go`, which opens with `//go:build integration`. CI runs `go test` without `-tags integration`, so the file is never compiled or run. A regression flipping `AuthorizeThreadFollow` back to permissive — re-opening exactly the gap PR #345 closed — would stay green in CI.

**Fix options**
- Drop the `integration` build tag (test depends on the same MySQL/Redis services CI already provides), or
- Add a CI job running `go test -tags integration ./modules/message/...`, or
- Add a non-tagged unit test asserting `checkChannelAccess(..., requireActive=true)` denies a blacklisted member while `false` allows.

Also tighten the four bare `assert.NotEqual(http.StatusOK, ...)` deny assertions in `modules/thread/read_path_blacklist_test.go` and add a `threadMdGet` blacklist read test (same review, P2 items).

lml2468:
## 🤖 Leader — Needs Human Review

| Field | Value |
|-------|-------|
| **reason** | `type:chore` is not in the routing decision tree |
| **action** | Cannot auto-route; needs human to determine scope and assignee |

### Routing rationale

- `type:chore` + `priority:P1`, no `stage:*` label
- Charter §1 decision tree covers: feature / bug / security / infra / cicd / observability / deployment
- `type:chore` is not covered — cannot auto-route without guessing
- P1 priority makes this important to route promptly

### Suggested action

- Human to clarify: is this a test infra task (→ devops-engineer)? A code quality task (→ backend-engineer)? Or something else?
- Once clarified, add appropriate type label or assign directly

lml2468:
## 🚦 Leader Triage — Unroutable

| Field | Value |
|---|---|
| **Issue** | #353 |
| **Labels** | `type:chore`, `priority:P1`, `needs-human-review` |
| **Decision** | **Cannot route — needs human triage** |

### Why

The routing decision tree maps the following `type:*` labels:

| Label | Routes to |
|---|---|
| `type:feature` | product-analyst → architect → backend/frontend |
[...]

PR #356, opened 2026-06-11
fix(conversation_ext): exclude blacklisted members from thread-ext materialization (fanout + FollowChannel)

### LINKED ISSUE #351: [P1] channel-follow fanout still materializes subchannel ext rows for blacklisted members (PR#345 mandatory follow-up)
From PR #345 review (mandatory follow-up condition attached to the final APPROVE verdict).

**Problem**
`AuthorizeChannelFollow` intentionally stays on permissive `ExistMember` (GROUP semantics), but `FollowChannel` is not a pure group operation: it sets `auto_follow_threads=1` and bulk-materializes thread ext rows (`modules/conversation_ext/service.go:393,455`). `OnThreadCreated` fanout then selects every `auto_follow_threads=1` row (`service.go:594/614`) and re-checks only ext flags, not active group membership, before inserting thread ext rows (`service.go:787/804`).

Net effect: a blacklisted parent-group member who followed the channel keeps receiving new-subchannel ext rows / creation notifications. Content reads are fully gated by `ExistMemberActive` (no content leak) — this is a metadata/notification-layer leak only.

**Fix options**
- Filter fanout (`selectEligibleForFanoutTx` / `OnThreadCreated`) by `ExistMemberActive`, or
- Cascade-clear `auto_follow_threads` + thread ext materialization on blacklist.

**Also add** coverage that a blacklisted parent member with a group ext row does not receive existing or newly-created thread ext rows.

Refs: PR #345 final verdict (mandatory P1 follow-up #1), reviewer findings on head 57bbb95/7ff6e8c.

yujiawei:
## 背景

PR #345 终审 APPROVE 附带的 mandatory P1 follow-up #1（GH #351）。

`AuthorizeChannelFollow` 对 GROUP follow 有意保持 permissive `ExistMember`，但 `FollowChannel` 不是纯群操作：它写 `auto_follow_threads=1` 并批量物化既有子区 ext 行；`OnThreadCreated` fanout 选取所有 `auto_follow_threads=1` 行后只 re-check ext flag，不查群成员状态。净效果：被拉黑的父群成员（曾关注过 channel）持续收到既有/新建子区的 ext 行与创建通知——元数据/通知层泄漏（内容读已被 #345 的 `ExistMemberActive` 门禁兜死）。

## 修法（issue 两个选项中的 A：物化写路径按 active 过滤）

- `conversation_ext.Service` 新增 `ActiveMemberFilter` 注入点（与 `ThreadAuthChecker` 同款依赖倒置，避免 import group 成环；nil = 跳过，仅单测/迁移期），由 `message/1module.go` 在既有 `threadAuthChecker` 上实现并注入
- **OnThreadCreated fanout**：初始 SELECT 后、进 batch tx 前按「活跃父群成员」过滤目标。tx 外执行（快照语义不变，不延长锁窗口）；多 uid 走一次 `GetSubscribableMemberUIDs`（与 #343 IM 订阅数据源同语义：`status=Normal AND is_deleted=0`）内存取交集，不放大查询
- **FollowChannel**：Phase 2/3 既有子区物化前单点 `ExistMemberActive` 探测，非活跃成员跳过物化。Phase 1 群行写入保持 permissive GROUP 语义不变；`auto_follow_threads=1` 保留——解除拉黑后新子区 fanout 自动恢复，flag 残留期间两条物化路径都有过滤，无泄漏

## 测试（按 issue 要求补：blacklist 成员不应收到已存在/新建子区 ext 行）

- 被拉黑成员（已有 auto_follow 群 ext 行）：新建子区 fanout 不再给其物化 ext 行；正常成员不受影响；解除拉黑后 fanout 自动恢复
- 被拉黑成员调用 FollowChannel：GROUP 行照常写入（permissive 语义回归保护），既有子区 ext 行不物化；正常成员正常物化
- 测试**不带** `//go:build integration` tag（吃 CI 已就绪的 MySQL service），避免 #353 指出的「integration-tagged 测试 CI 永不编译」缺口在本修复上重演

## 与 PR #355（GH #354 Bot 跟人走）的关系

互补、无文件交集：#355 改 `modules/group/*`（拉黑级联到 owner 的 bot、成员移除级联）；本 PR 改 `modules/conversation_ext` + `modules/message`（sidebar ext 物化层过滤）。两者可独立合并，无依赖顺序要求。

## 审查状态

- 改动规模：+306 行（含 ~160 行测试），3 文件
- 编译/vet：`go build ./...`、`go vet`（含 `-tags integration`）通过
- 测试：`modules/conversation_ext`、`modules/message`、`modules/thread` 全包 `-race -shuffle=on` 通过（CI 同款 MySQL/Redis/WuKongIM 栈）
- 自审 critical pass（SQL 参数化、锁序不变量、nil-checker fail-open 范围、并发注入 RWMutex）：无新增 finding

Fixes #351

### DIFF
--- a/modules/conversation_ext/service.go
+++ b/modules/conversation_ext/service.go
@@ -101,6 +101,23 @@ type DefaultFollowedGroupGuard interface {
 	FilterDefaultFollowed(uid, spaceID string, candidateGroupNos []string) ([]string, error)
 }
 
+// ActiveMemberFilter 把一批 uid 过滤为「仍是 groupNo 活跃成员」（is_deleted=0 AND
+// status=Normal，排除被拉黑成员）的子集。窄接口、依赖倒置，由 message/1module.go
+// 注入实现（group.IService），避免 conversation_ext 直接 import group 形成循环依赖。
+//
+// 引入背景（issue #351 / PR #345 mandatory follow-up）：AuthorizeChannelFollow
+// 对 GROUP follow 有意保持 permissive ExistMember，但 FollowChannel 不是纯群操作——
+// 它写 auto_follow_threads=1 并物化既有子区 ext 行；OnThreadCreated fanout 也只
+// re-check ext flag（auto_follow_threads=1 AND group_unfollowed=0），不查群成员
+// 状态。被拉黑的父群成员因此持续收到既有/新建子区的 ext 行与创建通知（元数据层
+// 泄漏；内容读已被 ExistMemberActive 门禁拦住）。本接口用于在这两条子区物化写
+// 路径上按 active membership 过滤：GROUP 行本身的语义不变。
+type ActiveMemberFilter interface {
+	// FilterActiveMemberUIDs 返回 uids 中当前是 groupNo 活跃成员的子集。
+	// 实现不要求保序，也不要求去重输入。
+	FilterActiveMemberUIDs(groupNo string, uids []string) ([]string, error)
+}
+
 // ThreadEnumerator 是 FollowChannel 级联物化子区时使用的窄接口。
 // 与 ThreadAuthChecker 一样采用依赖倒置，避免 conversation_ext 直接 import thread。
 // 由 message/1module.go 启动时通过 SetThreadEnumerator 注入；nil 时跳过物化
@@ -158,6 +175,11 @@ type Service struct {
 	// DefaultFollowedGroups 直接返回空——fail-closed 比放过更安全。
 	defaultFollowedGuard  DefaultFollowedGroupGuard
 	defaultFollowedGuardM sync.RWMutex
+	// activeMemberFilter 是子区 ext 物化写路径（FollowChannel Phase 2/3 +
+	// OnThreadCreated fanout）的活跃成员过滤钩子（issue #351）。由 message/1module.go
+	// 启动时注入；nil 时跳过过滤（仅供单测 / 迁移期使用，与 threadAuth 同约定）。
+	activeMemberFilter  ActiveMemberFilter
+	activeMemberFilterM sync.RWMutex
 	log.Log
 }
 
@@ -235,6 +257,24 @@ func (s *Service) getDefaultFollowedGroupGuard() DefaultFollowedGroupGuard {
 	g := s.defaultFollowedGuard
 	s.defaultFollowedGuardM.RUnlock()
 	return g
+}
+
+// SetActiveMemberFilter injects the active-membership filter used by the two
+// thread-ext materialization write paths (FollowChannel Phase 2/3 and the
+// OnThreadCreated fanout) — issue #351.  Safe for concurrent use; intended to
+// be called once at startup from message/1module.go after the group module
+// has initialised.
+func (s *Service) SetActiveMemberFilter(f ActiveMemberFilter) {
+	s.activeMemberFilterM.Lock()
+	s.activeMemberFilter = f
+	s.activeMemberFilterM.Unlock()
+}
+
+func (s *Service) getActiveMemberFilter() ActiveMemberFilter {
+	s.activeMemberFilterM.RLock()
+	f := s.activeMemberFilter
+	s.activeMemberFilterM.RUnlock()
+	return f
 }
 
 // AuthorizeAndMaterializeDefaultFollowedGroups is the pre-flight step the
@@ -399,6 +439,25 @@ func (s *Service) FollowChannel(uid, spaceID, groupNo string) error {
 		return nil
 	}); err != nil {
 		return err
+	}
+
+	// issue #351（PR #345 mandatory follow-up）：子区 ext 物化只面向「活跃父群成员」。
+	// GROUP follow 本身保持 permissive（Phase 1 已写群行，与 AuthorizeChannelFollow
+	// 的 GROUP 语义一致），但被拉黑成员不得借 FollowChannel 重新物化既有子区 ext 行
+	// （元数据泄漏——removeUserFromGroupThreadsCleanup 在拉黑时已删过一轮）。
+	// auto_follow_threads=1 保留：OnThreadCreated fanout 侧同样按 active 过滤，flag
+	// 残留期间无泄漏；解除拉黑后新子区 fanout 自动恢复，无需用户重新操作。
+	// nil filter 仅用于单测 / 迁移期（与 threadAuth / channelAuth 同约定）。
+	if filter := s.getActiveMemberFilter(); filter != nil {
+		activeUIDs, err := filter.FilterActiveMemberUIDs(groupNo, []string{uid})
+		if err != nil {
+			// Phase 1 已 commit（群行语义不受影响）；过滤失败 fail-closed 跳过物化，
+			// 错误向上返回让调用方记录。缺失的子区行由下次 FollowChannel / fanout 补齐。
+			return fmt.Errorf("FollowChannel filter active member: %w", err)
+		}
+		if len(activeUIDs) == 0 {
+			return nil
+		}
 	}
 
 	enum := s.getThreadEnumerator()
@@ -624,6 +683,22 @@ func (s *Service) OnThreadCreated(groupNo, shortID string) error {
 		return nil
 	}
 
+	// 1.5 issue #351（PR #345 mandatory follow-up）：按「活跃父群成员」过滤 fanout
+	//     目标。auto_follow_threads=1 只代表用户曾关注 channel，不代表当前仍可见——
+	//     被拉黑（status=Blacklist）成员的群行 flag 仍在，但不应再收到新子区的 ext
+	//     行 / 创建通知（元数据层泄漏；内容读已被 ExistMemberActive 门禁兜住）。
+	//     与初始 SELECT 一样在 tx 外执行（快照语义，不延长锁窗口）；解除拉黑后
+	//     从下一条新子区起自动恢复 fanout。nil filter 仅用于单测 / 迁移期。
+	if filter := s.getActiveMemberFilter(); filter != nil {
+		targets, err = filterFanoutTargetsByActiveMembership(filter, groupNo, targets)
+		if err != nil {
+			return fmt.Errorf("OnThreadCreated filter active members: %w", err)
+		}
+		if len(targets) == 0 {
+			return nil
+		}
+	}
+
 	threadChannelID := groupNo + threadSeparator + shortID
 
 	// 2. 按 batch 切分，每 batch 一个 tx：
@@ -721,6 +796,37 @@ func isRetriableMySQLLockErr(err error) bool {
 type onThreadCreatedTarget = struct {
 	UID     string `db:"uid"`
 	SpaceID string `db:"space_id"`
+}
+
+// filterFanoutTargetsByActiveMembership 把 OnThreadCreated 的 fanout 目标过滤为
+// 「仍是 groupNo 活跃成员」的子集（issue #351）。同一 uid 可能带多个 space_id 行，
+// 先按 uid 去重再批量查询，过滤结果保持 targets 原有顺序（ORDER BY uid, space_id
+// 的锁序约定不被破坏）。
+func filterFanoutTargetsByActiveMembership(filter ActiveMemberFilter, groupNo string, targets []onThreadCreatedTarget) ([]onThreadCreatedTarget, error) {
+	uids := make([]string, 0, len(targets))
+	seen := make(map[string]struct{}, len(targets))
+	for _, t := range targets {
+		if _, ok := seen[t.UID]; ok {
+			continue
+		}
+		seen[t.UID] = struct{}{}
+		uids = append(uids, t.UID)
+	}
+	activeUIDs, err := filter.FilterActiveMemberUIDs(groupNo, uids)
+	if err != nil {
+		return nil, err
+	}
+	activeSet := make(map[string]struct{}, len(activeUIDs))
+	for _, u := range activeUIDs {
+		activeSet[u] = struct{}{}
+	}
+	filtered := make([]onThreadCreatedTarget, 0, len(targets))
+	for _, t := range targets {
+		if _, ok := activeSet[t.UID]; ok {
+			filtered = append(filtered, t)
+		}
+	}
+	return filtered, nil
 }
 
 // bulkBumpFollowVersionTx 批量给一批已排序的 (uid, space_id) +1 follow_version。
--- a/modules/message/1module.go
+++ b/modules/message/1module.go
@@ -80,6 +80,11 @@ func init() {
 			// 子区前必须校验 caller 是群成员 + 群在 Space 可见。复用同一个 struct
 			// 实现，共享 checkChannelAccess 逻辑。
 			svc.SetChannelAuthChecker(checker)
+			// 注入 ActiveMemberFilter（issue #351）：子区 ext 物化写路径（FollowChannel
+			// Phase 2/3 + OnThreadCreated fanout）按「活跃父群成员」过滤，被拉黑成员
+			// 不再收到既有/新建子区的 ext 行。GROUP 行语义不变（AuthorizeChannelFollow
+			// 保持 permissive）。复用同一个 threadAuthChecker，共享 group.IService 实例。
+			svc.SetActiveMemberFilter(checker)
 			// 注入 DefaultFollowedGroupGuard：/v1/follow/sort 的 target_type=2
 			// 候选必须先经过 (成员 + Space 可见性 + 非 Disband + category_id) 完整
 			// 校验链才能被物化为 ext 行。没有这一步，恶意客户端可借 sort 接口给
@@ -188,6 +193,47 @@ func (c *threadAuthChecker) AuthorizeThreadFollow(uid, spaceID, groupNo, shortID
 // 群成员。子区(CommunityTopic)分支的 active 校验只发生在 AuthorizeThreadFollow。
 func (c *threadAuthChecker) AuthorizeChannelFollow(uid, spaceID, groupNo string) error {
 	return c.checkChannelAccess(uid, spaceID, groupNo, false)
+}
+
+// FilterActiveMemberUIDs implements convext.ActiveMemberFilter (issue #351).
+//
+// 返回 uids 中当前是 groupNo「活跃成员」（is_deleted=0 AND status=Normal，排除
+// 被拉黑成员）的子集，供 conversation_ext 的子区 ext 物化写路径（FollowChannel
+// Phase 2/3 + OnThreadCreated fanout）过滤目标。
+//
+//   - 单 uid（FollowChannel 路径）：直查 ExistMemberActive，避免拉全群成员列表；
+//   - 多 uid（fanout 路径）：一次 GetSubscribableMemberUIDs（status=Normal AND
+//     is_deleted=0，与 #343 的 IM 订阅数据源同语义）后在内存里取交集——fanout
+//     单批最多 1000 目标，逐个 ExistMemberActive 会放大 N 倍查询。
+func (c *threadAuthChecker) FilterActiveMemberUIDs(groupNo string, uids []string) ([]string, error) {
+	switch len(uids) {
+	case 0:
+		return nil, nil
+	case 1:
+		active, err := c.groupSvc.ExistMemberActive(groupNo, uids[0])
+		if err != nil {
+			return nil, err
+		}
+		if !active {
+			return nil, nil
+		}
+		return []string{uids[0]}, nil
+	}
+	activeUIDs, err := c.groupSvc.GetSubscribableMemberUIDs(groupNo)
+	if err != nil {
+		return nil, err
+	}
+	activeSet := make(map[string]struct{}, len(activeUIDs))
+	for _, u := range activeUIDs {
+		activeSet[u] = struct{}{}
+	}
+	out := make([]string, 0, len(uids))
+	for _, u := range uids {
+		if _, ok := activeSet[u]; ok {
+			out = append(out, u)
+		}
+	}
+	return out, nil
 }
 
 // checkChannelAccess 复用 FollowThread 既有逻辑的群级访问校验：
--- a/modules/message/thread_ext_blacklist_filter_test.go
+++ b/modules/message/thread_ext_blacklist_filter_test.go
@@ -9,15 +9,21 @@ package message
 // is_deleted=0）的父群成员两条路径都不应再收到子区 ext 行（元数据/通知层泄漏；
 // 内容读已被 ExistMemberActive 门禁兜住）。
 //
-// 测试基建约定（PR #356 round-1 CI 红的教训）：本包非 integration-tag 测试一律
-// 不跑 sql-migrate（包内既有 testutil.NewTestServer 用例全部 t.Skip，而
-// channel_files_blacklist_test.go 等用例手建最小表——若本文件经 module.Setup 跑
-// 迁移，shuffle 把手建表用例排在前面时 sql-migrate 会撞 Error 1050 Table already
-// exists）。因此这里照搬 integration e2e helper 的做法：手建最小表 + 裸 INSERT
-// 种子 + 显式装配 service（wiring 与 1module.go 注入逻辑逐行对齐）。
+// 测试基建约定（PR #356 round-1 CI 红 + round-2 review 的教训）：
+//  1. 不跑 sql-migrate——本包非 integration-tag 测试一律不经 module.Setup（包内
+//     既有 testutil.NewTestServer 用例全部 t.Skip，channel_files_blacklist_test.go
+//     等用例手建最小表；混跑迁移会在 -shuffle 下撞 Error 1050 Table already exists）。
+//  2. 不碰共享 test 库——本文件要 DROP+CREATE 核心表（user_conversation_ext /
+//     group_member / …），而 go test ./... 默认跨包并行、modules/group、modules/
+//     thread 等包用 testutil.NewTestServer 连同一个 test 库，跨包并行时破坏性 DDL
+//     会撞别包的查询（round-2 review blocking）。因此先在 MySQL 实例上建独立库
+//     再连入，所有 DDL/数据只落在隔离库里。
+// 写法照搬 integration e2e helper：手建最小表 + 裸 INSERT 种子 + 显式装配
+// service（wiring 与 1module.go 注入逻辑逐行对齐）。
 // =============================================================================
 
 import (
+	"os"
 	"strings"
 	"testing"
 
@@ -31,22 +37,41 @@ import (
 
 const extBlSpaceID = "s_ext_bl"
 
-// extBlNewCtx 构造指向测试 MySQL 的 *config.Context（config.New 默认 DSN 即
-// root:demo@…/test，与 CI service 一致），显式 Migration=false——不经 module.Setup。
+// extBlDBName 是本文件专用的隔离库名：与共享 test 库（testutil.NewTestServer /
+// 其它包的迁移与查询）完全隔离，跨包并行的 go test ./... 下破坏性 DDL 不会外溢。
+const extBlDBName = "octo_msg_blacklist_test"
+
+// extBlNewCtx 构造指向隔离库的 *config.Context：先用 config.New 默认 DSN（与 CI
+// MySQL service 同源的 root:demo@…/test）建隔离库，再把 DSN 的库名换成隔离库连入。
+// 显式 Migration=false——不经 module.Setup。可用 MSG_BLACKLIST_TEST_MYSQL_ADDR
+// 覆盖隔离库 DSN（与 newSidebarIntegCtx 的 env 覆盖风格一致）。
 func extBlNewCtx(t *testing.T) *config.Context {
 	t.Helper()
+	bootCfg := config.New()
+	bootCfg.Test = true
+	bootCfg.DB.Migration = false
+	boot := config.NewContext(bootCfg)
+	_, err := boot.DB().Exec(
+		"CREATE DATABASE IF NOT EXISTS " + extBlDBName +
+			" CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci")
+	require.NoError(t, err, "create isolated db")
+
+	addr := os.Getenv("MSG_BLACKLIST_TEST_MYSQL_ADDR")
+	if addr == "" {
+		addr = "root:demo@tcp(127.0.0.1:3306)/" + extBlDBName + "?charset=utf8mb4&parseTime=true"
+	}
 	cfg := config.New()
 	cfg.Test = true
 	cfg.DB.Migration = false
+	cfg.DB.MySQLAddr = addr
 	return config.NewContext(cfg)
 }
 
-// extBlEnsureTables 手建本测试用到的最小表（DDL 与 modules/group、
+// extBlEnsureTables 在隔离库里手建本测试用到的最小表（DDL 与 modules/group、
 // modules/conversation_ext、modules/thread 的迁移文件中本测试触达的列对齐；
 // 写法照搬 default_followed_group_guard_e2e_test.go / thread_follow_blacklist_
-// e2e_test.go 的同名 helper）。DROP + CREATE 而非 IF NOT EXISTS：其它用例
-// （如 channel_files_blacklist_test.go）可能先以更窄的列集建过 group_member，
-// 必须重建保证本测试的列都在；表只装每用例自种数据，破坏性 DDL 安全。
+// e2e_test.go 的同名 helper）。DROP + CREATE 保证每个用例拿到本文件期望的
+// schema；隔离库只被本包顺序执行的测试使用，破坏性 DDL 安全。
 func extBlEnsureTables(t *testing.T, ctx *config.Context) {
 	t.Helper()
 	for _, tbl := range []string{"user_conversation_ext", "user_follow_version", "thread", "group_member", "`group`"} {
--- a/modules/message/thread_ext_blacklist_filter_test.go
+++ b/modules/message/thread_ext_blacklist_filter_test.go
@@ -1,0 +1,154 @@
+package message
+
+// =============================================================================
+// Issue #351（PR #345 mandatory follow-up）— 子区 ext 物化按活跃成员过滤。
+//
+// AuthorizeChannelFollow 对 GROUP follow 保持 permissive ExistMember，但
+// FollowChannel 会物化既有子区 ext 行、OnThreadCreated fanout 会给所有
+// auto_follow_threads=1 的群行物化新子区 ext 行。被拉黑（status=Blacklist、
+// is_deleted=0）的父群成员两条路径都不应再收到子区 ext 行（元数据/通知层泄漏；
+// 内容读已被 ExistMemberActive 门禁兜住）。
+//
+// 非 integration tag：CI 的 go test 直接跑（MySQL/Redis service 已就绪），
+// 避免 #353 指出的「integration-tagged 测试 CI 永不编译」缺口在本修复上重演。
+// =============================================================================
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	convext "github.com/Mininglamp-OSS/octo-server/modules/conversation_ext"
+	"github.com/Mininglamp-OSS/octo-server/modules/group"
+	"github.com/Mininglamp-OSS/octo-server/modules/thread"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const extBlSpaceID = "s_ext_bl"
+
+// setupThreadExtBlacklistData 建一个父群（space_id 为空 → legacy wildcard 可见），
+// 两个正常成员 normalUID / victimUID，返回 ctx 与已装配好的 conversation_ext 单例。
+// 装配（ThreadAuthChecker / ChannelAuthChecker / ThreadEnumerator / ActiveMemberFilter）
+// 由 testutil.NewTestServer → module.Setup 跑本包 1module.go 的注入逻辑完成，
+// 与生产 boot 同一条 wiring 路径。
+func setupThreadExtBlacklistData(t *testing.T) (*config.Context, *convext.Service, string, string, string) {
+	t.Helper()
+	_, ctx := testutil.NewTestServer()
+	require.NoError(t, testutil.CleanAllTables(ctx))
+
+	svc := convext.GetGlobalConvExtService()
+	require.NotNil(t, svc, "conversation_ext 单例应已由 module.Setup 初始化并完成注入")
+
+	groupNo := strings.ReplaceAll(util.GenerUUID(), "-", "")
+	normalUID := "u_ext_normal_" + util.GenerUUID()[:8]
+	victimUID := "u_ext_victim_" + util.GenerUUID()[:8]
+
+	groupDB := group.NewDB(ctx)
+	require.NoError(t, groupDB.Insert(&group.Model{
+		GroupNo: groupNo, Name: "父群", Creator: normalUID, Status: 1, Version: 1,
+	}))
+	for _, u := range []string{normalUID, victimUID} {
+		require.NoError(t, groupDB.InsertMember(&group.MemberModel{
+			GroupNo: groupNo, UID: u,
+			Status: int(common.GroupMemberStatusNormal), Version: 1, Vercode: util.GenerUUID(),
+		}))
+	}
+	return ctx, svc, groupNo, normalUID, victimUID
+}
+
+func blacklistMemberExtBl(t *testing.T, ctx *config.Context, groupNo, uid string) {
+	t.Helper()
+	_, err := ctx.DB().UpdateBySql(
+		"UPDATE group_member SET status=? WHERE group_no=? AND uid=?",
+		int(common.GroupMemberStatusBlacklist), groupNo, uid,
+	).Exec()
+	require.NoError(t, err)
+}
+
+// hasThreadExtRow 查 user_conversation_ext 是否存在 (uid, space, target_type=5, channelID) 行。
+func hasThreadExtRow(t *testing.T, ctx *config.Context, uid, channelID string) bool {
+	t.Helper()
+	var count int64
+	_, err := ctx.DB().Select("count(*)").From("user_conversation_ext").
+		Where("uid=? AND space_id=? AND target_type=5 AND target_id=?", uid, extBlSpaceID, channelID).
+		Load(&count)
+	require.NoError(t, err)
+	return count > 0
+}
+
+// seedActiveThread 在 thread 表插入一个 active 子区，返回 channelID。
+func seedActiveThread(t *testing.T, ctx *config.Context, groupNo, shortID string) string {
+	t.Helper()
+	require.NoError(t, thread.NewDB(ctx).Insert(&thread.Model{
+		ShortID: shortID, GroupNo: groupNo, Name: "topic", CreatorUID: "creator",
+		Status: thread.ThreadStatusActive,
+	}))
+	return groupNo + "____" + shortID
+}
+
+// TestOnThreadCreated_BlacklistedMemberExcludedFromFanout 验证：两个成员都开启了
+// auto_follow_threads（正常时 FollowChannel），其中一个随后被拉黑——新建子区的
+// fanout 只给活跃成员物化 ext 行；解除拉黑后 fanout 自动恢复。
+func TestOnThreadCreated_BlacklistedMemberExcludedFromFanout(t *testing.T) {
+	ctx, svc, groupNo, normalUID, victimUID := setupThreadExtBlacklistData(t)
+
+	// 两人在正常状态下关注 channel（写 auto_follow_threads=1 群行）。
+	require.NoError(t, svc.FollowChannel(normalUID, extBlSpaceID, groupNo))
+	require.NoError(t, svc.FollowChannel(victimUID, extBlSpaceID, groupNo))
+
+	// victim 被拉黑后新建子区。
+	blacklistMemberExtBl(t, ctx, groupNo, victimUID)
+	sid1 := "1489104291682713601"
+	require.NoError(t, svc.OnThreadCreated(groupNo, sid1))
+
+	assert.True(t, hasThreadExtRow(t, ctx, normalUID, groupNo+"____"+sid1),
+		"正常成员应收到新子区 ext 行")
+	assert.False(t, hasThreadExtRow(t, ctx, victimUID, groupNo+"____"+sid1),
+		"被拉黑成员不应收到新子区 ext 行（issue #351 元数据泄漏）")
+
+	// 解除拉黑 → 下一条新子区恢复 fanout（auto_follow_threads=1 保留的语义）。
+	_, err := ctx.DB().UpdateBySql(
+		"UPDATE group_member SET status=? WHERE group_no=? AND uid=?",
+		int(common.GroupMemberStatusNormal), groupNo, victimUID,
+	).Exec()
+	require.NoError(t, err)
+	sid2 := "1489104291682713602"
+	require.NoError(t, svc.OnThreadCreated(groupNo, sid2))
+	assert.True(t, hasThreadExtRow(t, ctx, victimUID, groupNo+"____"+sid2),
+		"解除拉黑后 fanout 应自动恢复")
+}
+
+// TestFollowChannel_BlacklistedMemberSkipsExistingThreadMaterialization 验证：
+// 被拉黑成员调用 FollowChannel（GROUP 门禁 permissive，调用本身放行）时，
+// 既有子区的 ext 物化被跳过；正常成员则正常物化。
+func TestFollowChannel_BlacklistedMemberSkipsExistingThreadMaterialization(t *testing.T) {
+	ctx, svc, groupNo, normalUID, victimUID := setupThreadExtBlacklistData(t)
+
+	existingChannelID := seedActiveThread(t, ctx, groupNo, "1489104291682713603")
+
+	blacklistMemberExtBl(t, ctx, groupNo, victimUID)
+
+	// 被拉黑成员 FollowChannel：GROUP 行写入放行（permissive 语义不变），但
+	// 不得物化既有子区 ext 行。
+	require.NoError(t, svc.FollowChannel(victimUID, extBlSpaceID, groupNo),
+		"GROUP follow 对被拉黑成员保持 permissive，调用不应报错")
+	assert.False(t, hasThreadExtRow(t, ctx, victimUID, existingChannelID),
+		"被拉黑成员不应通过 FollowChannel 重新物化既有子区 ext 行（issue #351）")
+
+	var groupRowCount int64
+	_, err := ctx.DB().Select("count(*)").From("user_conversation_ext").
+		Where("uid=? AND space_id=? AND target_type=2 AND target_id=?",
+			victimUID, extBlSpaceID, groupNo).
+		Load(&groupRowCount)
+	require.NoError(t, err)
+	assert.Equal(t, int64(1), groupRowCount, "GROUP 级 ext 行本身应照常写入（permissive 语义不变）")
+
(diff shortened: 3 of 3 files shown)


mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#356 Review Report
**Head SHA:** `463d851423c201a952c02ab979b45089b23e134e`
**PR:** Mininglamp-OSS/octo-server #356
**Title:** fix(conversation_ext): exclude blacklisted members from thread-ext materialization (fanout + FollowChannel)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| ActiveMemberFilter interface design | ✅ | `service.go:104-118` — narrow interface, dependency inversion consistent with ThreadAuthChecker/ChannelAuthChecker pattern |
| FollowChannel filter placement | ✅ | `service.go:441-459` — after Phase 1 commit, before Phase 2 enumerate; fail-closed on error/empty |
| OnThreadCreated filter placement | ✅ | `service.go:683-698` — after initial SELECT, before batch tx; same snapshot semantics |
| filterFanoutTargetsByActiveMembership | ✅ | `service.go:798-833` — UID dedup, single batch query, order preservation for lock ordering |
| FilterActiveMemberUIDs implementation | ✅ | `1module.go:195-237` — single-uid → ExistMemberActive; multi-uid → GetSubscribableMemberUIDs + in-memory intersect |
| Injection wiring | ✅ | `1module.go:83-87` — reuses threadAuthChecker, same composition point as existing checkers |
| Upstream SQL correctness | ✅ | ExistMemberActive: `WHERE group_no=? AND uid=? AND is_deleted=0 AND status=Normal` (`db.go:182-189`); GetSubscribableMemberUIDs: same predicates (`db.go:448-456`) |
| Test coverage | ✅ | `thread_ext_blacklist_filter_test.go` — both fanout and FollowChannel paths; blacklist + un-blacklist recovery; no integration tag (CI-ready) |
| Concurrency safety | ✅ | RWMutex on filter field (`service.go:178-182`); filter runs outside tx (no lock window extension) |

## 2. Findings

### F1 — Snapshot race between active-membership filter and tx re-check (P2, pre-existing pattern)

**Diff-scope:** pre-existing pattern, not new. The active-membership filter runs outside the tx (same as the initial SELECT at `service.go:671-681`). A user could be blacklisted between the filter call and `selectEligibleForFanoutTx` inside the tx. However, `selectEligibleForFanoutTx` (`service.go:882-905`) only checks `auto_follow_threads=1 AND group_unfollowed=0` — it does NOT re-check group membership status.

**Why P2 not P1:** This is the same documented snapshot-semantics pattern already accepted for the initial SELECT (see `service.go:671-681` comments). The blacklist path runs `removeUserFromGroupThreadsCleanup` which deletes existing thread ext rows, so the window is narrow and self-healing. Next thread creation will correctly filter. Not new or amplified by this PR.

### F2 — GetSubscribableMemberUIDs returns all active members for large groups (P2, acceptable trade-off)

**Diff-scope:** new code path, but documented trade-off. For fanout with N target UIDs (N>1), `FilterActiveMemberUIDs` calls `GetSubscribableMemberUIDs(groupNo)` which returns ALL active member UIDs. For very large groups (10k+), this is a large result set. But fanout targets are capped at batch size 1000, and the alternative (N individual queries) would be worse. Same approach as IM subscription data source (#343).

## 3. Suggestions

- **F1 mitigation (future):** If the snapshot race becomes a concern, `selectEligibleForFanoutTx` could add `AND uid IN (SELECT uid FROM group_member WHERE group_no=? AND is_deleted=0 AND status=Normal)` — but this would add a subquery inside the tx and is not warranted given the current self-healing behavior.
- **Test enhancement (nit):** Add a unit test for `filterFanoutTargetsByActiveMembership` with a single UID mapping to multiple space_ids, to verify order preservation across space_ids.

## 4. Additional Observations

- The blacklist API path (`api.go:3150-3168`) calls `removeUserFromGroupThreads` → `removeUserFromGroupThreadsCleanup` which cleans thread_member, thread_setting, IM subscriptions, and thread-level (target_type=5) conversation_ext rows, but intentionally does NOT clear the GROUP-level ext row (target_type=2) with `auto_follow_threads=1`. This PR correctly addresses the gap at the read/materialization layer rather than the write/cleanup layer — preserving the `auto_follow_threads=1` flag so un-blacklist automatically restores fanout without user re-action.
- The `auto_follow_threads=1` flag retention during blacklist is a deliberate design choice documented in the PR body and code comments. It ensures symmetric behavior: blacklist → filter blocks materialization; un-blacklist → filter passes, fanout resumes from next thread creation.

## 5. Data-Flow Trace

### FollowChannel path (single UID)
```
FollowChannel(uid, spaceID, groupNo)
  → Phase 1: upsert GROUP ext row (auto_follow_threads=1) [committed]
  → getActiveMemberFilter() → ActiveMemberFilter
  → FilterActiveMemberUIDs(groupNo, [uid])
    → case len==1: groupSvc.ExistMemberActive(groupNo, uid)
      → db.ExistMemberActive(uid, groupNo)
        → SQL: SELECT count(*) FROM group_member WHERE group_no=? AND uid=? AND is_deleted=0 AND status=GroupMemberStatusNormal
        → Returns bool (true only for Normal status, excludes Blacklist)
    → If false → returns empty slice → FollowChannel returns nil (skip Phase 2/3)
    → If true → returns [uid] → proceed to Phase 2/3
```
**Verdict:** Data flows correctly. Blacklisted UID is excluded at the SQL predicate level.

### OnThreadCreated path (multi UID)
```
OnThreadCreated(groupNo, shortID)
  → SELECT uid, space_id FROM user_conversation_ext WHERE target_type=2 AND target_id=groupNo AND auto_follow_threads=1
  → getActiveMemberFilter() → ActiveMemberFilter
  → filterFanoutTargetsByActiveMembership(filter, groupNo, targets)
    → Dedup UIDs from targets
    → FilterActiveMemberUIDs(groupNo, uids)
      → case len>1: groupSvc.GetSubscribableMemberUIDs(groupNo)
        → db.querySubscribableMemberUIDsWithGroupNo(groupNo)
          → SQL: SELECT uid FROM group_member WHERE group_no=? AND is_deleted=0 AND status=GroupMemberStatusNormal
        → In-memory intersect with input UIDs
    → Filter targets by active UID set, preserve order
  → If empty → return nil
  → Proceed to batch tx (selectEligibleForFanoutTx + bulkInsertThreadExtForFanoutUsersTx)
```
**Verdict:** Data flows correctly. SQL predicate excludes Blacklist status. Order preservation maintains lock-ordering invariant.

### Injection wiring
```
message/1module.go init()
  → newThreadAuthChecker(appCtx) → &threadAuthChecker{groupSvc: group.NewService(ctx), ...}
  → svc.SetActiveMemberFilter(checker)
    → Service.activeMemberFilter = checker (protected by RWMutex)
```
**Verdict:** Wiring is correct. Same composition point and struct as existing ThreadAuthChecker/ChannelAuthChecker.

- **C1 — Dual-path parity:** ✅ Clear.
  - FollowChannel filter ↔ OnThreadCreated filter: Both use same `ActiveMemberFilter` interface, same `getActiveMemberFilter()` getter, same fail-closed semantics. Parity confirmed.
  - Blacklist add ↔ remove: Add path filters at materialization; remove path (un-blacklist) restores IM subscriptions. `auto_follow_threads=1` preserved during blacklist → un-blacklist automatically restores fanout. Symmetric.

- **C2 — Control-flow ordering / nesting:** ✅ Clear.
  - `FilterActiveMemberUIDs` dispatched by `len(uids)` switch: 0→nil, 1→ExistMemberActive, N→GetSubscribableMemberUIDs. No double-filtering or ordering issue.
  - Filter runs outside tx — no nested tx concern.
  - No security-control regex to test with non-canonical forms.

- **C3 — Authorization boundary ≠ capability boundary:** ✅ N/A.
  - `ActiveMemberFilter` is injected at startup, not exposed via any API endpoint. No new capability surface opened.

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. Two P2 observations (acceptable snapshot race pattern; large-group query trade-off) are documented and non-blocking. The fix correctly addresses the metadata-layer leakage described in issue #351, with proper data-flow tracing to upstream SQL predicates confirming blacklisted members are excluded at both materialization write paths.

yujiawei (commented):
# Code Review — PR #356 (octo-server)

**Scope reviewed:** head `463d851423c201a952c02ab979b45089b23e134e`, merge-base `791570b5` (#345, already on `main`). HEAD is a clean single-commit descendant of current `main`; `git diff merge-base..HEAD` equals the commit diff (3 files, +306/-0). No sibling-PR conflict.

## 1. Verification

| Item | Result | Evidence |
|---|---|---|
| Both thread-ext materialization write paths are gated | ✅ | Only two writers of `target_type=5` rows exist: `bulkInsertThreadExtForFanoutUsersTx` (fanout, `service.go:742`) and `bulkInsertThreadExtForChannelMaterializeTx` (FollowChannel Phase 3, `service.go:514`). Both are now preceded by an `ActiveMemberFilter` check (`service.go:684` for fanout, `service.go:452` for FollowChannel). The default-followed-group materialization explicitly does not touch thread rows (`db.go:426`), so no third path is missed. |
| Filter semantics exclude blacklist | ✅ | Single-uid path → `ExistMemberActive` (`db.go:182`: `is_deleted=0 AND status=Normal`); multi-uid path → `GetSubscribableMemberUIDs` → `querySubscribableMemberUIDsWithGroupNo` (`db.go:448`: same predicate). Both filters are semantically identical. |
| Service-wrapper arg order correct | ✅ | `Service.ExistMemberActive(groupNo, uid)` → `db.ExistMemberActive(uid, groupNo)` (`service.go:572`) — order is swapped correctly at the wrapper, consistent with the rest of the package. |
| GROUP-follow permissive semantics preserved | ✅ | FollowChannel Phase 1 group-row write runs before the filter and is untouched; `AuthorizeChannelFollow` stays `permissive` (`1module.go:195`). Test asserts group row count = 1 for a blacklisted caller while thread rows = 0. |
| Fail-closed on filter error | ✅ | Both call sites return a wrapped error and skip materialization on filter failure (`service.go:455`, `service.go:688`). |
| `auto_follow_threads=1` retained → auto-resume after un-blacklist | ✅ | Flag is not cleared; both paths re-filter at materialization time, so flag residue is harmless. Covered by the un-blacklist recovery assertion in `TestOnThreadCreated_BlacklistedMemberExcludedFromFanout`. |
| DI wiring matches production boot | ✅ | `SetActiveMemberFilter(checker)` injected in `1module.go:84` on the shared `threadAuthChecker`, same inversion pattern as `ThreadAuthChecker`/`ChannelAuthChecker`; `nil`-filter skip path is test/migration-only. |
| Build / vet | ✅ | `go build` + `go vet ./modules/conversation_ext/... ./modules/message/...` clean. |
| Tests | ✅ | `modules/conversation_ext` and `modules/message` packages pass; the two new tests pass under the CI env (`OCTO_MASTER_KEY=0123456789abcdef0123456789abcdef`). Tests are deliberately **untagged**, so CI actually compiles and runs them — correctly applying the #353 lesson that integration-tagged tests never compile in CI. |

## 2. Issues

None at P0/P1. The fix is correctly scoped: it closes the metadata/notification-layer re-leak where a blacklisted parent-group member could keep receiving subchannel ext rows / creation notifications via fanout or `FollowChannel` re-materialization, even though content reads are already gated by `ExistMemberActive`. Defense-in-depth is intact: blacklisting already removes existing subscriptions via `removeUserFromGroupThreadsCleanup` (`group/api.go:3162`), so this PR closes the *re-materialization* gap rather than the initial-read gap — making the "metadata-only leak" severity framing accurate.

## 3. Suggestions (non-blocking nits)

- **N1 — fanout intersection cost.** The multi-uid path calls `GetSubscribableMemberUIDs(groupNo)`, which loads *all* active members of the group for an in-memory set-intersection against the fanout targets. For a very large group with relatively few `auto_follow` users this reads more rows than strictly necessary. The chosen approach is still the right tradeoff vs. N per-target `ExistMemberActive` calls (which would amplify queries up to the 1000-target batch), and the data source is already shared with the #343 subscription path — so this is acknowledged, not a defect. Worth a comment only if group sizes grow much further.
- **N2 — minor TOCTOU.** The active-membership filter runs as a tx-outside snapshot before Phase 2/3 (FollowChannel) and before the batch loop (fanout). A caller active at filter time but blacklisted before the write would still get rows materialized. The window is tiny and self-healing: the blacklist hook's `removeUserFromGroupThreadsCleanup` cleans up afterward, and the inverse window (blacklisted→active) is recovered by the next fanout/FollowChannel. This matches the existing documented snapshot semantics for this module and needs no change.

## 4. Additional findings

None. The change does not alter the established lock ordering (version → ext) — the fanout filter only shrinks the target slice before batching; `bulkBumpFollowVersionTx` / `selectEligibleForFanoutTx` ordering inside each batch tx is untouched.

## Verdict

**APPROVED** — no blocking issues. Correct, narrowly scoped security fix with symmetric coverage of both write paths, fail-closed error handling, and tests that actually run in CI.

Jerry-Xin (dismissed):
Fixes the two P1 follow-ups from #345 correctly: OnThreadCreated fanout and FollowChannel both now filter by active membership before materializing thread ext rows. Dependency inversion via `ActiveMemberFilter` follows the existing `ThreadAuthChecker` pattern cleanly.

**🔴 Blocking items**: None

**💬 Non-blocking notes**

1. **TOCTOU between filter and tx (low risk, accepted tradeoff)**: The active-membership check runs outside the materialization transaction in both paths. A concurrent blacklist between filter pass and tx insert could theoretically leak one stale ext row. This is an inherent limitation of the dependency-inversion design (moving the check into the tx would require importing group into conversation_ext or threading group.IService through tx closures). Given that (a) content reads are already gated by `ExistMemberActive` from #345, (b) the race window is sub-millisecond, and (c) blacklist cleanup / next unfollow will remove any leaked row, this is an acceptable residual risk. Consider documenting this known tradeoff in the `ActiveMemberFilter` interface godoc for future maintainers.

2. **Fanout path fetches all active members**: `FilterActiveMemberUIDs` calls `GetSubscribableMemberUIDs(groupNo)` for multi-uid fanout, which returns ALL active group members even when only a few have `auto_follow_threads=1`. For very large groups this could be heavier than necessary. Acceptable for now since (a) it's the same query already used by IM subscription (#343), (b) fanout batches cap at 1000, and (c) the alternative (N individual `ExistMemberActive` calls) is worse. Could optimize later with a `WHERE uid IN (...)` variant if profiling shows it matters.

**✅ Highlights**
- Clean dependency inversion matching existing patterns — no circular imports
- Single-uid path uses `ExistMemberActive` directly (avoids full member list scan)
- Tests cover both fanout and FollowChannel paths including unblacklist recovery
- Phase 1 GROUP semantics preserved correctly (permissive ExistMember unchanged)
- nil-filter fail-open convention consistent with ThreadAuthChecker/ChannelAuthChecker

yujiawei:
Round-2 push (5a6da55a) fixes the CI Test red from round 1.

**Root cause**: `TestOnThreadCreated_BlacklistedMemberExcludedFromFanout` used `testutil.NewTestServer` → `module.Setup` → sql-migrate. This package's untagged suite never runs migrations (every existing `testutil.NewTestServer` test here is skipped); tests like `channel_files_blacklist_test.go` hand-create minimal tables instead. Under `-shuffle=on`, a hand-created `group_member` preceding the first migration run makes sql-migrate fail with `Error 1050: Table 'group_member' already exists` on `20191106000002_group_legacy01.sql`.

**Fix**: follow the package convention (same approach as the tagged e2e helpers) — plain ctx with `Migration=false`, DROP+CREATE minimal tables with DDL mirroring the real migrations, raw-SQL seeding, and explicit service wiring that mirrors the `1module.go` injections. Assertions unchanged.

Verified locally: full package `-race -shuffle=on` on a fresh DB, again on a dirty DB with a second seed, plus the exact failing order (channel_files first, these tests after).

yujiawei (commented):
# Code Review — PR #356 (octo-server)

**Scope reviewed:** HEAD `5a6da55a` against merge-base `791570b5` (base `main`). Three files, +411 lines, no deletions:
- `modules/conversation_ext/service.go`
- `modules/message/1module.go`
- `modules/message/thread_ext_blacklist_filter_test.go` (new)

## Summary

The change closes a metadata-leak gap left open by #345: `AuthorizeChannelFollow` intentionally keeps a *permissive* `ExistMember` check for GROUP follow, but the two thread-ext **materialization write paths** (`FollowChannel` Phase 2/3 and the `OnThreadCreated` fanout) never re-checked group-member status. A blacklisted parent-group member therefore kept receiving existing/new sub-thread ext rows and creation notifications (`auto_follow_threads=1` survives blacklisting), even though content reads were already gated by `ExistMemberActive`.

The fix introduces a narrow `ActiveMemberFilter` interface (dependency-inverted, injected from `message/1module.go`) and applies it on both write paths. GROUP-row semantics are deliberately left unchanged. The approach is consistent with the existing `ThreadAuthChecker` / `ChannelAuthChecker` / `ThreadEnumerator` injection pattern, so no new import cycle is introduced.

## 1. Correctness verification

| Item | Verdict | Evidence |
|---|---|---|
| FollowChannel filters single uid before Phase 2/3 materialization | ✅ | `service.go:454-468` — `FilterActiveMemberUIDs(groupNo, []string{uid})`; on empty result returns before `getThreadEnumerator()`. Phase 1 (group row + `auto_follow=1`) is already committed, matching the documented permissive GROUP semantics. |
| Fanout filters all targets after the snapshot SELECT, before batching/tx | ✅ | `service.go:690-700` runs at step 1.5, outside the tx, so the lock window is not extended (snapshot semantics preserved). |
| Filter applied to the right set, correct intersection direction | ✅ | `filterFanoutTargetsByActiveMembership` (`service.go:801-829`) dedups targets by uid, queries once, builds `activeSet` from active members, then keeps only targets in the set — original ordering preserved, so the `ORDER BY uid, space_id` lock-order contract is intact. |
| Single-uid vs multi-uid query strategy | ✅ | `1module.go:198-230` — `len==1` → `ExistMemberActive` (single row, avoids loading the whole roster); `len>1` → one `GetSubscribableMemberUIDs` + in-memory intersection (avoids N round-trips on a fanout batch up to `maxAutoFollowThreadsPerChannel`). |
| Active-membership semantics consistent across both paths | ✅ | `ExistMemberActive` (`group/db.go:182`) = `is_deleted=0 AND status=Normal`; `querySubscribableMemberUIDsWithGroupNo` (`group/db.go:448`) = same predicate. No drift between the two paths. |
| `group.IService` exposes both methods used | ✅ | `group/service.go:47` (`GetSubscribableMemberUIDs`) and `:61` (`ExistMemberActive`). |
| nil-filter migration/test escape hatch | ✅ | Both call sites guard `if filter := s.getActiveMemberFilter(); filter != nil`, same convention as `threadAuth` / `channelAuth`. RWMutex-protected getter/setter (`service.go:262-278`) matches the other injected deps. |
| Auto-recovery after un-blacklist | ✅ | `auto_follow_threads=1` is preserved, so the next new thread re-fans-out without user action. Verified by `TestOnThreadCreated_BlacklistedMemberExcludedFromFanout` (sid2 re-materializes after status flips back to Normal). |
| Build / vet / tests | ✅ | `go build` + `go vet` clean on both modules; `go test ./modules/conversation_ext/... ./modules/message/...` passes, and both new tests pass against a real MySQL instance. |

## 2. Issues found

No P0 or P1 issues. The change is correct, well-isolated, and adequately tested. Items below are non-blocking.

### P2 / Suggestions

- **Fanout loads the full active roster regardless of target-set size.** For `len(uids) > 1`, `GetSubscribableMemberUIDs` returns *all* active members of the group, even when the auto-follow target set is small. For very large groups this query is heavier than the data actually needed (the intersection only cares about the target uids). The current code documents the tradeoff (avoiding N×`ExistMemberActive` round-trips) and it is a reasonable default, but a batched `ExistMembersActive(groupNos, uid)`-style lookup keyed on the target uids would scale better on big groups. Not blocking — acceptable as-is.

- **Un-blacklist does not backfill threads created while blacklisted.** After un-blacklisting, only *new* threads fan out; sub-threads created during the blacklist window are not retroactively materialized until the user re-follows the channel. This is consistent with the documented `auto_follow` design and arguably the safer default, but worth confirming it matches product intent for #351. Documentation-level note only.

### Nits

- The `FollowChannel` inline comment references "Phase 2/3" for the filter, while the filter itself sits between Phase 1 and the enumerator. The behavior is correct (it short-circuits before both Phase 2 enumerate and Phase 3 materialize); the wording is just slightly ahead of where the check physically lives.

## 3. Recommendations

Mergeable as-is. If you want to harden for large-group fanout later, consider a target-keyed batched membership lookup instead of loading the whole roster (P2 above). Otherwise no required changes.

## 4. Additional findings (outside the stated scope)

- The test follow-up commit (`5a6da55a`) correctly addresses the round-1 CI failure (Error 1050 under `-shuffle=on`): this package's untagged suite never runs sql-migrate, so going through `module.Setup` collided with hand-created tables from sibling tests like `channel_files_blacklist_test.go`. Switching to `Migration=false` + DROP+CREATE minimal tables + explicit wiring that mirrors `1module.go` line-by-line is the right convention for this package. The test wiring (`setupThreadExtBlacklistData`, `service.go` injections) faithfully reproduces the production injection path, so the tests exercise the real filter logic rather than a stub.
- No other callers of `FollowChannel` / `OnThreadCreated` bypass the filter; injection is the single composition point in `1module.go`, reusing the shared `threadAuthChecker` (and thus a single `group.IService` instance).

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#356 Review Report
**Head SHA:** `5a6da55ad8cd3b8930df8c6d299b00550184a7d5`
**PR:** Mininglamp-OSS/octo-server#356
**Title:** fix(conversation_ext): exclude blacklisted members from thread-ext materialization (fanout + FollowChannel)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| ActiveMemberFilter DI pattern | ✅ | `service.go:104-119` — clean interface, mirrors `ThreadAuthChecker`/`ChannelAuthChecker` pattern |
| RWMutex concurrent safety | ✅ | `service.go:178-179` field + `service.go:262-276` getter/setter — same pattern as `defaultFollowedGuard` |
| FollowChannel filter placement | ✅ | `service.go:453-465` — after Phase 1 commit (group row preserved), before Phase 2 enumerate |
| OnThreadCreated filter placement | ✅ | `service.go:695-703` — after initial SELECT, before batch tx loop; snapshot semantics preserved |
| FilterActiveMemberUIDs impl | ✅ | `1module.go:207-237` — single-uid → `ExistMemberActive`; multi-uid → `GetSubscribableMemberUIDs` + set-intersect |
| Production wiring | ✅ | `1module.go:84` — `svc.SetActiveMemberFilter(checker)` in same `register.AddModule` block as other injections |
| Test wiring mirrors production | ✅ | `thread_ext_blacklist_filter_test.go:155-159` — all 4 injections (ThreadAuth/ChannelAuth/ThreadEnumerator/ActiveMemberFilter) present |
| Fail-closed on filter error | ✅ | Both paths propagate error + skip materialization; Phase 1 group row preserved for self-healing |
| nil filter = skip (test/migration) | ✅ | Consistent with `threadAuthChecker`/`channelAuthChecker`/`threadEnumerator` conventions |

## 2. Findings

### No P0/P1 findings.

### P2 — `GetSubscribableMemberUIDs` unbounded query (pre-existing, amplified by new call site)

**Diff-scope:** pre-existing query, new call site added by this PR.

`FilterActiveMemberUIDs` for `len(uids) >= 2` calls `groupSvc.GetSubscribableMemberUIDs(groupNo)` (`1module.go:222`), which runs an unbounded `SELECT uid FROM group_member WHERE group_no=? AND is_deleted=0 AND status=1` — no LIMIT, no pagination (`modules/group/db.go:449-455`). For a 10k-member group this loads 10k UIDs into memory per fanout call.

**Why P2 not P1:** This query is already used by 3 other call sites (IM subscriber management in `message/1module.go:222`, `group/1module.go:73`, `thread/1module.go:139`), so the scalability profile is pre-existing. The PR adds one more call site but with the same per-groupNo cost. Fanout batch size is capped at 1000 (`onThreadCreatedBatchSize`), so the intersection is bounded on the targets side.

**Suggestion (non-blocking):** If large-group fanout latency becomes a concern, consider a narrower SQL query that takes the candidate UIDs and returns only the active subset in one round-trip (e.g., `SELECT uid FROM group_member WHERE group_no=? AND uid IN (...) AND is_deleted=0 AND status=1`), avoiding the full member list load.

### P2 — `MaterializeDefaultFollowedGroups` still writes `auto_follow_threads=1` for blacklisted members (pre-existing, not amplified)

**Diff-scope:** pre-existing; PR does not touch `MaterializeDefaultFollowedGroups` or `DefaultFollowedGroupGuard`.

The `/v1/follow/sort` path uses `AuthorizeChannelFollow` (permissive `ExistMember`), so a blacklisted member's group row can still get `auto_follow_threads=1`. With this PR's filter on `OnThreadCreated`, no thread ext rows leak — the flag is inert for blacklisted members. Not a data leak, just a minor semantic inconsistency.

**Why P2 not P1:** No user-visible incorrect data. The flag's presence is intentional (auto-resume on un-blacklist, documented in PR body).

### Nit — Test DDL `group_member.status` default differs from migration

`thread_ext_blacklist_filter_test.go:88`: `status SMALLINT NOT NULL DEFAULT 1`. The real migration default may differ. Since the test always explicitly sets `status` in INSERT, this is cosmetic only.

## 3. Data-Flow Tracing

### Path 1: OnThreadCreated fanout

```
thread.Service.CreateThread commits
  → convext.OnThreadCreated(groupNo, shortID)
    → SELECT uid, space_id FROM user_conversation_ext
        WHERE target_type=2 AND target_id=groupNo AND auto_follow_threads=1
        [returns all users who followed the channel, INCLUDING blacklisted]
    → [NEW] filterFanoutTargetsByActiveMembership(filter, groupNo, targets)
        → dedup UIDs from targets
        → filter.FilterActiveMemberUIDs(groupNo, uids)
          → [len>=2] groupSvc.GetSubscribableMemberUIDs(groupNo)
              SQL: SELECT uid FROM group_member WHERE group_no=? AND is_deleted=0 AND status=1
              [blacklisted members have status=2 → excluded from result]
          → set-intersect: targets ∩ activeUIDs → filtered targets
    → batch tx loop: bulkBumpFollowVersionTx + selectEligibleForFanoutTx + bulkInsertThreadExtForFanoutUsersTx
        [only filtered targets participate → blacklisted members get no thread ext rows] ✅
```

**Verification:** Data flows correctly. The filter sits between the initial SELECT and the batch tx loop. Blacklisted members are excluded before any write happens. The `selectEligibleForFanoutTx` re-check inside the tx (FOR UPDATE on group row) still runs for the surviving targets, preserving the `auto_follow=1 AND group_unfollowed=0` invariant.

### Path 2: FollowChannel Phase 2/3

```
FollowChannel(uid, spaceID, groupNo)
  → checkChannelAccess(uid, spaceID, groupNo, requireActive=false)
      → ExistMember(groupNo, uid) [permissive — blacklisted passes]
  → Phase 1 tx: BumpFollowVersionTx + upsertTx(group row, auto_follow_threads=1)
      [commits — group row written for all members including blacklisted]
  → [NEW] filter.getActiveMemberFilter()
      → filter.FilterActiveMemberUIDs(groupNo, [uid])
        → [len==1] groupSvc.ExistMemberActive(groupNo, uid)
            SQL: SELECT count(*) FROM group_member WHERE group_no=? AND uid=? AND is_deleted=0 AND status=1
            [blacklisted: status=2 → count=0 → returns false → empty slice]
      → len(activeUIDs)==0 → return nil (skip Phase 2/3)
  → [blacklisted member gets no thread ext rows] ✅
  → [group row with auto_follow_threads=1 preserved → auto-resume on un-blacklist] ✅
```

**Verification:** Single-uid path uses `ExistMemberActive` directly (efficient, no full member list load). Phase 1 group row is intentionally preserved. Early return before `getThreadEnumerator()` means no wasted work.

### Path 3: FollowThread (not touched by PR — already gated)

```
FollowThread(uid, spaceID, threadChannelID)
  → checker.AuthorizeThreadFollow(uid, spaceID, groupNo, shortID)
    → checkChannelAccess(uid, spaceID, groupNo, requireActive=true)
      → ExistMemberActive(groupNo, uid) [blacklisted → rejected → ErrThreadForbidden]
```

**Verification:** Already uses `ExistMemberActive` (set in PR #345). No gap.

## 4. R5 Blind-Spot Checklist

- FollowChannel ↔ UnfollowChannel: asymmetric by design (follow materializes, unfollow deletes). Filter only needed on the materialization side.
- OnThreadCreated has no symmetric "thread deleted" fanout.
- `MaterializeDefaultFollowedGroups` writes group rows only; thread materialization is downstream and now filtered.

**C2 — Control-flow ordering / nesting reuse:** ✅ Clear.
- `filterFanoutTargetsByActiveMembership` is called exactly once per `OnThreadCreated` invocation, outside the tx. No nesting or double-application risk.
- `FilterActiveMemberUIDs` has two branches (single vs multi-uid) but no shared mutable state between them.
- No regex/sanitize controls in this PR.

**C3 — Authorization boundary ≠ capability boundary:** ✅ Clear.
- PR narrows authorization (adds filter), doesn't expand capabilities.
- No new endpoints or tools exposed.

## 5. Additional Observations

- **Race window (accepted):** Between the filter check (outside tx) and the batch tx commit, a member could be blacklisted. The resulting orphaned ext row would be cleaned up by `removeUserFromGroupThreadsCleanup`. This is consistent with the documented snapshot semantics and pre-existing race windows in the same code path.
- **Test quality:** Both tests cover the core scenarios (blacklist → no materialization, un-blacklist → resume, normal member unaffected, GROUP row preserved). Test infra correctly avoids `sql-migrate` per the package convention (PR commit 2 message explains the CI-red root cause and fix).
- **Commit hygiene:** Two clean commits — implementation + test fix. Second commit message thoroughly documents the CI failure root cause.

## 6. Verdict

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The PR correctly closes the metadata-layer leak for blacklisted members on both thread-ext materialization paths (FollowChannel Phase 2/3 and OnThreadCreated fanout), while preserving the intentional permissive GROUP semantics and auto-resume-on-un-blacklist behavior. The DI pattern, concurrency safety, error handling, and test coverage are all consistent with established codebase conventions. The two P2 notes are pre-existing concerns not amplified to blocking severity by this diff.

Jerry-Xin (changes requested):
This PR is in scope for `octo-server` and the production wiring is directionally sound, but the new untagged tests introduce a blocking CI stability risk.

**🔴 Blocking**
- 🔴 Critical: `modules/message/thread_ext_blacklist_filter_test.go:36` and `modules/message/thread_ext_blacklist_filter_test.go:52` run against the default test database with migrations disabled, then `DROP TABLE IF EXISTS` core shared tables: `user_conversation_ext`, `user_follow_version`, `thread`, `group_member`, and `group`. Because this test is intentionally not integration-tagged, it will run in normal package tests. Under `go test ./...`, Go runs packages concurrently by default, and many untagged `modules/group` and `modules/thread` tests use the same shared `test` database via `testutil.NewTestServer()`. Dropping and recreating these tables from `modules/message` can race with those packages’ migrations and queries, causing intermittent failures or schema mismatches. Please move this to an isolated DSN/database, use non-destructive row cleanup with the real migrated schema, or put this coverage behind an integration setup that is isolated from the default shared DB.

**💬 Non-blocking**
- 🟡 Warning: `modules/conversation_ext/service.go:692` filters active members before the fanout transaction, and `modules/conversation_ext/service.go:451` filters `FollowChannel` after phase 1 but before enumeration/materialization. This preserves the existing lock window, but it still allows a narrow race where a user is blacklisted after the active-member check and before the thread ext insert, potentially leaving a persistent ext row after blacklist cleanup has already run. If the intended guarantee is strict “no ext rows after blacklist commits,” the active membership check needs to be coupled more tightly with the materialization transaction or with blacklist cleanup ordering.

**✅ Highlights**
- The new `ActiveMemberFilter` interface follows the existing dependency inversion pattern and avoids a `conversation_ext -> group` import cycle.
- `FilterActiveMemberUIDs` avoids N+1 lookups for fanout by using `GetSubscribableMemberUIDs`.
- The targeted message tests for the new cases passed locally: `go test ./modules/message -run 'Test(OnThreadCreated_BlacklistedMemberExcludedFromFanout|FollowChannel_BlacklistedMemberSkipsExistingThreadMaterialization)' -count=1`.

yujiawei:
Round-3 push (24f4a0d5) addresses the blocking finding:

**🔴 shared-DB destructive DDL** — fixed. The tests now bootstrap and connect to a dedicated `octo_msg_blacklist_test` database (DSN overridable via `MSG_BLACKLIST_TEST_MYSQL_ADDR`, mirroring the `newSidebarIntegCtx` env-override style). All DROP/CREATE and rows stay inside the isolated database; verified after a full package run that the shared `test` DB contains none of these tables. Cross-package parallel `go test ./...` can no longer be affected.

**🟡 active-check vs materialization TOCTOU** — acknowledged, intentionally accepted; documenting the reasoning here rather than changing code:

- The window (blacklist commits between the active-member check and the thread-ext INSERT) produces at worst one stale **metadata** row; content reads stay gated by `ExistMemberActive` everywhere.
- The blacklist-add path already cascade-deletes the user's thread ext rows (`removeUserFromGroupThreadsCleanup`), and any later blacklist/unfollow cleanup removes stragglers; the row also never re-fans-out (both materialization paths filter on every call).
- Closing it strictly would require locking `group_member` inside the conversation_ext fanout tx (cross-module table coupling + a wider lock window on a hot table) — the same trade-off the existing fanout snapshot semantics already make for `FollowChannel`/`UnfollowChannel` races, which are documented as expected in `OnThreadCreated`'s comments. The new filter deliberately rides that same snapshot model.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#356 Review Report
**Head SHA:** `24f4a0d57ceeebbb0ba2901cbe8de5ed614b1de3`
**PR:** Mininglamp-OSS/octo-server#356
**Title:** fix(conversation_ext): exclude blacklisted members from thread-ext materialization (fanout + FollowChannel)
**Base:** main (`791570b5`)
**Commits reviewed:** 3 (impl `463d8514` → test-fix `5a6da55a` → isolated-DB `24f4a0d5`)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| ActiveMemberFilter DI interface | ✅ | `service.go:104-119` — clean narrow interface, mirrors ThreadAuthChecker/ChannelAuthChecker pattern |
| RWMutex concurrent safety | ✅ | `service.go:178-179` field + `service.go:262-276` getter/setter — same pattern as defaultFollowedGuard |
| FollowChannel filter placement | ✅ | `service.go:453-465` — after Phase 1 commit (group row preserved), before Phase 2 enumerate |
| OnThreadCreated filter placement | ✅ | `service.go:695-703` — after initial SELECT, before batch tx loop; snapshot semantics preserved |
| FilterActiveMemberUIDs impl (single-uid) | ✅ | `1module.go:207-218` — `ExistMemberActive(groupNo, uid)` → `group_member WHERE group_no=? AND uid=? AND is_deleted=0 AND status=Normal` |
| FilterActiveMemberUIDs impl (multi-uid) | ✅ | `1module.go:220-237` — `GetSubscribableMemberUIDs(groupNo)` → same predicate, in-memory intersection |
| Production wiring | ✅ | `1module.go:84` — `svc.SetActiveMemberFilter(checker)` in same register.AddModule block as other injections |
| Test wiring mirrors production | ✅ | `thread_ext_blacklist_filter_test.go:155-159` — all 4 injections (ThreadAuth/ChannelAuth/ThreadEnumerator/ActiveMemberFilter) present |
| Fail-closed on filter error | ✅ | Both paths propagate error + skip materialization; Phase 1 group row preserved for self-healing |
| nil filter = skip (test/migration) | ✅ | Consistent with threadAuthChecker/channelAuthChecker/threadEnumerator conventions |
| Test DB isolation (round-3 blocking fix) | ✅ | `thread_ext_blacklist_filter_test.go:42-68` — creates isolated DB `octo_msg_blacklist_test`, all DDL scoped to it |
| Auto-recovery after un-blacklist | ✅ | `auto_follow_threads=1` preserved; verified by test sid2 re-materialization after status flip |

## 2. Issues Found

### No P0/P1 findings.

### P2 — `GetSubscribableMemberUIDs` unbounded query (pre-existing, not amplified to blocking)

**Diff-scope:** pre-existing query, new call site added by this PR.

`FilterActiveMemberUIDs` for `len(uids) >= 2` calls `groupSvc.GetSubscribableMemberUIDs(groupNo)` (`1module.go:222`), which runs `SELECT uid FROM group_member WHERE group_no=? AND is_deleted=0 AND status=1` — no LIMIT, no pagination (`modules/group/db.go:449-455`). For a 10k-member group this loads 10k UIDs into memory per fanout call.

**Why P2 not P1:** This query is already used by 3+ other call sites (IM subscriber management in `message/1module.go`, `group/1module.go`, `thread/1module.go`), so the scalability profile is pre-existing. The PR adds one more call site but with the same per-groupNo cost. Fanout batch size is capped at 1000 (`onThreadCreatedBatchSize`), so the intersection is bounded on the targets side.

**Suggestion (non-blocking):** If large-group fanout latency becomes a concern, consider a narrower SQL query that takes the candidate UIDs and returns only the active subset in one round-trip (e.g., `SELECT uid FROM group_member WHERE group_no=? AND uid IN (...) AND is_deleted=0 AND status=1`), avoiding the full member list load.

### P2 — Threads created during blacklist window are not retroactively materialized (pre-existing design, not amplified)

**Diff-scope:** pre-existing design; PR does not introduce or change this behavior.

After un-blacklisting, only *new* threads fan out; sub-threads created during the blacklist window are not retroactively materialized until the user re-follows the channel. This is consistent with the documented `auto_follow` design and the safer default.

**Why P2 not P1:** No user-visible incorrect data. The behavior is intentional (auto-resume on new threads, not backfill). Product intent for #351 appears aligned.

### Nit — Test DDL `group_member.status` default differs from migration

`thread_ext_blacklist_filter_test.go:88`: `status SMALLINT NOT NULL DEFAULT 1`. The real migration default may differ. Since the test always explicitly sets `status` in INSERT, this is cosmetic only.

## 3. Data-Flow Tracing

### Path 1: OnThreadCreated fanout

```
thread.Service.CreateThread commits
  → convext.OnThreadCreated(groupNo, shortID)
    → SELECT uid, space_id FROM user_conversation_ext
        WHERE target_type=2 AND target_id=groupNo AND auto_follow_threads=1
        [returns all users who followed the channel, INCLUDING blacklisted]
    → [NEW] filterFanoutTargetsByActiveMembership(filter, groupNo, targets)
        → dedup UIDs from targets
        → filter.FilterActiveMemberUIDs(groupNo, uids)
          → [len>=2] groupSvc.GetSubscribableMemberUIDs(groupNo)
              SQL: SELECT uid FROM group_member WHERE group_no=? AND is_deleted=0 AND status=Normal
              [blacklisted members have status=Blacklist → excluded from result]
          → set-intersect: targets ∩ activeUIDs → filtered targets
    → batch tx loop: bulkBumpFollowVersionTx + selectEligibleForFanoutTx + bulkInsertThreadExtForFanoutUsersTx
        [only filtered targets participate → blacklisted members get no thread ext rows] ✅
```

**Verification:** Data flows correctly. The filter sits between the initial SELECT and the batch tx loop. Blacklisted members are excluded before any write happens. The `selectEligibleForFanoutTx` re-check inside the tx (FOR UPDATE on group row) still runs for the surviving targets, preserving the `auto_follow=1 AND group_unfollowed=0` invariant.

### Path 2: FollowChannel Phase 2/3

```
FollowChannel(uid, spaceID, groupNo)
  → checkChannelAccess(uid, spaceID, groupNo, requireActive=false)
      → ExistMember(groupNo, uid) [permissive — blacklisted passes]
  → Phase 1 tx: BumpFollowVersionTx + upsertTx(group row, auto_follow_threads=1)
      [commits — group row written for all members including blacklisted]
  → [NEW] filter.getActiveMemberFilter()
      → filter.FilterActiveMemberUIDs(groupNo, [uid])
        → [len==1] groupSvc.ExistMemberActive(groupNo, uid)
            SQL: SELECT count(*) FROM group_member WHERE group_no=? AND uid=? AND is_deleted=0 AND status=Normal
            [blacklisted: status=Blacklist → count=0 → returns false → empty slice]
      → len(activeUIDs)==0 → return nil (skip Phase 2/3)
  → [blacklisted member gets no thread ext rows] ✅
  → [group row with auto_follow_threads=1 preserved → auto-resume on un-blacklist] ✅
```

**Verification:** Single-uid path uses `ExistMemberActive` directly (efficient, no full member list load). Phase 1 group row is intentionally preserved. Early return before `getThreadEnumerator()` means no wasted work.

### Path 3: FollowThread (not touched by PR — already gated)

```
FollowThread(uid, spaceID, threadChannelID)
  → checker.AuthorizeThreadFollow(uid, spaceID, groupNo, shortID)
    → checkChannelAccess(uid, spaceID, groupNo, requireActive=true)
      → ExistMemberActive(groupNo, uid) [blacklisted → rejected → ErrThreadForbidden]
```

**Verification:** Already uses `ExistMemberActive` (set in PR #345). No gap.

### Upstream data-source parity

Both query paths use identical predicates:
- `ExistMemberActive` (`group/db.go:182-188`): `WHERE group_no=? AND uid=? AND is_deleted=0 AND status=GroupMemberStatusNormal`
- `querySubscribableMemberUIDsWithGroupNo` (`group/db.go:449-455`): `WHERE group_no=? AND is_deleted=0 AND status=GroupMemberStatusNormal`

No semantic drift between the two paths. ✅

## 4. R5 Blind-Spot Checklist

- FollowChannel ↔ UnfollowChannel: asymmetric by design (follow materializes, unfollow deletes). Filter only needed on the materialization side.
- OnThreadCreated has no symmetric "thread deleted" fanout.
- `MaterializeDefaultFollowedGroups` writes group rows only; thread materialization is downstream and now filtered.
- Blacklist cleanup path (`removeUserFromGroupThreadsCleanup` in `thread_cleanup.go:50`) deletes thread_member/thread_setting/IM subscriptions/ext rows at blacklist time. The PR's filter prevents *re-materialization* after that cleanup — no dual-path drift.

**C2 — Control-flow ordering / nesting reuse:** ✅ Clear.
- `filterFanoutTargetsByActiveMembership` is called exactly once per `OnThreadCreated` invocation, outside the tx. No nesting or double-application risk.
- `FilterActiveMemberUIDs` has two branches (single vs multi-uid) but no shared mutable state between them.
- No regex/sanitize controls in this PR.

**C3 — Authorization boundary ≠ capability boundary:** ✅ Clear.
- PR narrows authorization (adds filter), doesn't expand capabilities.
- No new endpoints or tools exposed.

## 5. Additional Observations

- **TOCTOU race window (accepted tradeoff):** Between the filter check (outside tx) and the batch tx commit, a member could be blacklisted. The resulting orphaned ext row would be cleaned up by `removeUserFromGroupThreadsCleanup` on the next blacklist event. This is consistent with the documented snapshot semantics and pre-existing race windows in the same code path. The in-tx `selectEligibleForFanoutTx` re-check on `auto_follow=1 AND group_unfollowed=0` provides an additional safety net for the group-row state, though it doesn't re-check member status.

- **Test DB isolation (commit `24f4a0d5` — addresses round-2 blocking):** The test now creates an isolated database `octo_msg_blacklist_test` via `CREATE DATABASE IF NOT EXISTS`, then connects to it with a separate DSN. All DROP+CREATE DDL is scoped to this isolated database. This fully addresses Jerry-Xin's round-2 blocking concern about cross-package parallel test interference on the shared `test` database. The env-var override (`MSG_BLACKLIST_TEST_MYSQL_ADDR`) follows the established `newSidebarIntegCtx` convention.

- **Test quality:** Both tests cover the core scenarios (blacklist → no materialization, un-blacklist → resume, normal member unaffected, GROUP row preserved). Test wiring faithfully reproduces the production injection path (all 4 injections present), so the tests exercise the real filter logic rather than a stub.

- **Commit hygiene:** Three clean commits — implementation, test CI-red fix, test DB isolation. Each commit message thoroughly documents the root cause and rationale.

## 6. Verdict

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The PR correctly closes the metadata-layer leak for blacklisted members on both thread-ext materialization paths (FollowChannel Phase 2/3 and OnThreadCreated fanout), while preserving the intentional permissive GROUP semantics and auto-resume-on-un-blacklist behavior. The DI pattern, concurrency safety, error handling, and test coverage are all consistent with established codebase conventions. The round-2 blocking concern (shared test DB interference) is fully resolved in commit `24f4a0d5` via isolated database creation. The two P2 notes are pre-existing concerns not amplified to blocking severity by this diff.

Jerry-Xin (approved):
This PR is in scope for octo-server, but the fix does not fully close the blacklist metadata leak under concurrent blacklist/fanout or blacklist/FollowChannel races.

**🔴 Blocking**
- 🔴 Critical: `OnThreadCreated` only applies the active-member filter before opening the fanout transaction, then the transaction rechecks only `auto_follow_threads` and `group_unfollowed`. A user can be active at [modules/conversation_ext/service.go:692](/Users/mlamp/.openclaw/tmp/pr-review-ezqzz1ej/repo/modules/conversation_ext/service.go:692), get blacklisted and cleaned up by [modules/group/api.go:3144](/Users/mlamp/.openclaw/tmp/pr-review-ezqzz1ej/repo/modules/group/api.go:3144) and [modules/group/thread_cleanup.go:75](/Users/mlamp/.openclaw/tmp/pr-review-ezqzz1ej/repo/modules/group/thread_cleanup.go:75), then still be inserted as a thread ext target at [modules/conversation_ext/service.go:738](/Users/mlamp/.openclaw/tmp/pr-review-ezqzz1ej/repo/modules/conversation_ext/service.go:738). Because blacklist cleanup is best-effort and does not clear `auto_follow_threads`, if cleanup runs before the fanout insert, the leaked ext row remains. The active membership check needs to be part of the final eligibility decision, ideally with locking semantics that serialize against the `group_member` status update, or the blacklist path must clear/disable the fanout source before cleanup.

- 🔴 Critical: `FollowChannel` has the same time-of-check/time-of-use issue. It checks active membership once after Phase 1 at [modules/conversation_ext/service.go:451](/Users/mlamp/.openclaw/tmp/pr-review-ezqzz1ej/repo/modules/conversation_ext/service.go:451), but Phase 3 later rechecks only the ext group row at [modules/conversation_ext/service.go:505](/Users/mlamp/.openclaw/tmp/pr-review-ezqzz1ej/repo/modules/conversation_ext/service.go:505). If the member is blacklisted and cleanup completes between those points, Phase 3 can recreate existing thread ext rows at [modules/conversation_ext/service.go:514](/Users/mlamp/.openclaw/tmp/pr-review-ezqzz1ej/repo/modules/conversation_ext/service.go:514). This preserves the exact leak the PR is trying to eliminate.

**✅ Highlights**
- The dependency inversion via `ActiveMemberFilter` matches the existing `ThreadAuthChecker` pattern and avoids a direct `conversation_ext -> group` import cycle.
- The batched fanout filter avoids per-target membership queries and preserves target ordering.
- The added tests cover the steady-state blacklist behavior for both fanout and `FollowChannel`.

**Verification**
- Ran `go test ./modules/message -run 'Test(OnThreadCreated_BlacklistedMemberExcludedFromFanout|FollowChannel_BlacklistedMemberSkipsExistingThreadMaterialization)$' -count=1` successfully.

yujiawei (commented):
# Code Review — PR #356 (octo-server)

**Verdict: APPROVED** — reviewed at head `24f4a0d57ceeebbb0ba2901cbe8de5ed614b1de3` (base `791570b5`, an ancestor of `origin/main`).

This is a re-review. The production code is unchanged since the previous approval at `463d8514`; the only delta on this head is a test-infrastructure fix that directly resolves the prior blocking review comment.

## What changed since the last approval

`git diff 463d8514..24f4a0d5 -- modules/conversation_ext/service.go modules/message/1module.go` is **empty** — production logic is byte-identical to the previously approved revision. The two follow-up commits touch only `modules/message/thread_ext_blacklist_filter_test.go`.

The prior blocking concern: the new untagged test ran `DROP TABLE` on the shared `test` database (`user_conversation_ext`, `group_member`, `group`, `thread`, `user_follow_version`), which could race with other packages under cross-package test parallelism.

The fix (`24f4a0d5`, "run blacklist-filter tests in an isolated database") addresses this precisely:
- A new isolated database `octo_msg_blacklist_test` is created (`thread_ext_blacklist_filter_test.go:42-68`); the bootstrap connection touches the shared `test` DB only to run `CREATE DATABASE IF NOT EXISTS` — no table DDL on `test`.
- All destructive `DROP TABLE` / `CREATE TABLE` is confined to the isolated DB via `cfg.DB.MySQLAddr` (`:59-67`, `:75-80`). The DSN is overridable through `MSG_BLACKLIST_TEST_MYSQL_ADDR`.

## Verification

| Check | Result |
|---|---|
| `go build ./modules/conversation_ext/... ./modules/message/...` | ✅ clean |
| `go vet` (same packages) | ✅ clean |
| Targeted tests (`TestOnThreadCreated_BlacklistedMemberExcludedFromFanout`, `TestFollowChannel_BlacklistedMemberSkipsExistingThreadMaterialization`) | ✅ both PASS |
| Isolation is genuine | ✅ confirmed: after a run, destructive DDL exists only in `octo_msg_blacklist_test`; the shared `test` DB is untouched |
| CI grants `CREATE DATABASE` | ✅ CI MySQL service is `root:demo` (`.github/workflows/ci.yml`) |

A note on the original race: the CI workflow already runs tests **one package at a time** (`while read -r pkg ... go test "$pkg"`) and drops+recreates the `test` DB between packages, so there is no `go test ./...` cross-package parallelism in CI. The isolated-DB change is still the correct fix — it removes destructive DDL from the shared `test` DB entirely and is robust to local `go test ./...` runs as well.

## Production logic (re-confirmed, unchanged)

- Both thread-ext (`target_type=5`) materialization write paths are gated by the injected `ActiveMemberFilter`: `FollowChannel` (single-uid, `service.go:451`) and the `OnThreadCreated` fanout (multi-uid, `service.go:692`). Both fail-closed on filter error and early-return on an empty active set.
- `FilterActiveMemberUIDs` (`message/1module.go`) uses `ExistMemberActive` for the single-uid path and `GetSubscribableMemberUIDs` for the fanout path — same active-membership semantics (`status=Normal AND is_deleted=0`).
- `filterFanoutTargetsByActiveMembership` dedups by uid and preserves the original `ORDER BY uid, space_id` lock order.
- GROUP-level ext rows (`target_type=2`) remain written for blacklisted callers (permissive `AuthorizeChannelFollow` semantics preserved) — the test asserts `groupRowCount==1` for a blacklisted `FollowChannel`, matching production behavior.

## Non-blocking notes (do not block merge)

- **TOCTOU (pre-existing, out of diff):** the filter is a tx-outside snapshot, so a user blacklisted between the active-member check and the ext insert can leave a stale ext row. This is benign — content reads are gated by `ExistMemberActive`, the fanout recovers on un-blacklist, and `auto_follow_threads=1` is intentionally retained. Consistent with the existing snapshot semantics of this module.
- **Test maintainability (nit):** the isolated DB name `octo_msg_blacklist_test` is a hardcoded constant; if this pattern is copy-pasted into another test file in the same package, both would target the same DB. A per-file suffix would prevent a future collision. Only one file uses it today, so no current impact.

## Conclusion

No P0/P1 in the diff range `791570b5..24f4a0d5`. Build, vet, and the targeted tests are green; the prior blocking concern is resolved. Recommending approval.

lml2468 (approved):
## Octo-PR APPROVE — #356 (head `24f4a0d`)

经维护者授权补第三票(author=yujiawei 占不了,Jerry-Xin+mochashanyao 已是另两个非作者且都 ✅)。内容逐处核实正确:

- ✅ **生产修复正确**:OnThreadCreated fanout + FollowChannel 两条子区 ext 物化写路径按「活跃父群成员」过滤(ActiveMemberFilter 依赖倒置),blacklist 成员不再被物化 ext 行/收创建通知。这是 #345 我裁决标的 mandatory P1 follow-up #1 的收口。
- ✅ **测试隔离修对**:改用专用隔离库 `octo_msg_blacklist_test`,破坏性 DDL 只落隔离库,不破坏跨包并发(Jerry-Xin 原 DROP 共享表 🔴 已修)。
- 🟡 **TOCTOU(过滤 tx 外、物化 tx 内的并发窗口)= 已知 P1 非 blocking**:内容读被 ExistMemberActive read gate 兜死(无内容越权),残留是元数据层;tx 外过滤是有意快照/锁权衡。三人(Jerry-Xin 撤 CR/mochashanyao/我)共识降 P1 follow-up(tx 内锁彻底消除另开 issue 追踪)。

无 blocking。
— Octo-PR(授权补票)

PR #364, opened 2026-06-13
feat(auth): per-request role resolution + tighten high-impact management gates

### LINKED ISSUE #363: security: harden role gating and role revocation for management APIs
## Summary

Proactive hardening of **system-level role handling** on management endpoints (`/v1/manager`, plus one write under `/v1/common`). Tracking issue only — rationale and specifics are kept to internal channels.

## Work items

1. **Raise the role requirement on a small set of high-impact management writes** from the general `admin` tier to `superAdmin`. Routes are unchanged (non-breaking); read / low-risk management endpoints stay at `admin`.
2. **Resolve the authoritative system role per request** (Redis-cached, mirroring the existing language resolver) instead of relying solely on the value carried in the session token, so role changes converge within a bounded window.

## Out of scope (separate issues)

- Centralized authorization layer + a CI guard asserting management routes perform a role check.
- Admin-action audit logging for the user / group / message management handlers.
- OpenAPI access-token scoping; frontend role-aware route guarding (cross-repo).

## Acceptance

- [ ] Item 1: affected writes require `superAdmin`; routes unchanged; tests updated (incl. regression that `admin` is rejected).
- [ ] Item 2: per-request role resolution with bounded staleness; unit tests for the resolver and the parser hook.

> Note: item 1 is an intentional behavior change — accounts holding only `admin` lose access to those specific operations.

an9xyz:
## Summary

Hardens system-level role handling on management endpoints. Closes #363.

1. **`POST /v1/common/appversion` → `superAdmin`.** Setting the client download source is a supply-chain-sensitive write and should not be reachable by the read-only `admin` tier. Role gate tightened from `CheckLoginRole()` to `CheckLoginRoleIsSuperAdmin()`; **route path unchanged** so existing callers are not broken.

2. **Destructive cross-space manager ops → `superAdmin`.** `forceDisband` / `liftBan` / `removeMembers` / `updateMemberRole` previously accepted any `admin`, giving the general admin role god-mode over any space. New `requireSuperAdmin` helper gates these four; read / low-risk management endpoints stay at `admin`. Returns the same generic 403 (`ErrSharedForbidden`) — no "needs higher role" leak (anti-enumeration).

3. **Per-request role resolution (revocation).** The system role was baked into the session token at login, so a demotion / admin-removal was not honored until the token expired. Added a `RoleResolver` hook on `CacheTokenParser` (mirroring the existing `LanguageResolver`) that resolves the authoritative role per request (`user_role:{uid}` Redis cache → DB → ""), overriding the token snapshot. Fail-open to the snapshot on resolver error so a cache/DB blip can't 5xx auth. Demotion now converges within the cache TTL (60s) instead of token lifetime — and because items 1 & 2 read this resolved role, those gates become revocation-aware automatically.

> Note: items 1 & 2 are intentional behavior changes — accounts holding only `admin` lose access to these specific operations.

## Changes

- `pkg/auth/parser.go` — `RoleResolver` interface, `WithRoleResolver` option, per-request role resolution in `Parse`.
- `modules/user/role_service.go` (new) — `RoleService` (Redis-cached, negative-marker for normal users).
- `modules/user/db.go` — `QueryRoleByUID` (single-column authoritative read).
- `main.go` — wire `RoleResolver` into the token parser.
- `modules/common/api.go` — `superAdmin` gate on `addAppVersion`.
- `modules/space/api_manager.go` — `requireSuperAdmin` + gate the four destructive handlers.

## Test plan

Validated against the real CI dependency stack (MySQL 8.0 / Redis 7 / WuKongIM v2.2.4-20260313):

- `go build ./...`, `go vet ./...` — pass
- Integration: `modules/space`, `modules/common`, `modules/user`, `pkg/auth` — pass
- New unit tests: parser role-resolver hook (override / empty-drops-role / fail-open / nil no-op); `RoleService` (cache hit / negative cache / DB miss backfill / DB error / invalidate / nil-dep panics)
- New regression: `TestManager_DestructiveOpsRequireSuperAdmin` asserts a plain `admin` is rejected (403) on all four destructive endpoints; the four happy-path manager tests elevated to `superAdmin`
- `make i18n-extract-check`, `make i18n-lint` — pass

https://claude.ai/code/session_01EaxNGwThDUvTqk2T5bPQvZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01EaxNGwThDUvTqk2T5bPQvZ)_

### DIFF
--- a/main.go
+++ b/main.go
@@ -117,11 +117,16 @@ func runAPI(ctx *config.Context) {
 	// 解析 cache value，支持 v2 JSON envelope 与 "uid@name[@role]" 旧格式（i18n D19/D21）。
 	// 同时注入 LanguageResolver，让 AuthMiddleware 把 user_language:{uid} → DB 的
 	// 真相源结果写到 UserInfo.Language 上，供 i18n.LanguageFromContext 读侧合并。
+	// 同时注入 RoleResolver：系统角色（admin/superAdmin）此前烧死在 token 里，
+	// 降权 / 删除管理员要到 token 过期才生效。这里在 Parse 时按 uid 实时解析 DB
+	// 角色（user_role:{uid} 热缓存 → DB），把降权生效窗口收敛到缓存 TTL。
 	userLangSvc := user.NewLanguageService(user.NewDB(ctx), ctx.Cache())
+	userRoleSvc := user.NewRoleService(user.NewDB(ctx), ctx.Cache())
 	route.SetTokenParser(auth.NewCacheTokenParser(
 		ctx.Cache(),
 		ctx.GetConfig().Cache.TokenCachePrefix,
 		auth.WithLanguageResolver(userLangSvc),
+		auth.WithRoleResolver(userRoleSvc),
 	))
 	// 替换web下的配置文件
 	replaceWebConfig(ctx.GetConfig())
--- a/modules/common/api.go
+++ b/modules/common/api.go
@@ -20,6 +20,8 @@ import (
 	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
+	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
 	spacepkg "github.com/Mininglamp-OSS/octo-server/pkg/space"
 	"github.com/gin-gonic/gin"
 	"go.uber.org/zap"
@@ -537,10 +539,11 @@ func (cn *Common) countriesList(c *wkhttp.Context) {
 // 发版会设置客户端的下载来源（DownloadURL），属供应链敏感写操作，因此要求
 // superAdmin —— 普通 admin 定位为只读运营位，不应能改动全体用户的客户端来源。
 // 路由仍保留在 /v1/common（历史调用方依赖），仅收紧角色门槛。
+// 拒绝路径走 i18n 信封并返回通用 403（ErrSharedForbidden），不泄露"需要更高
+// 角色"的具体原因（反枚举），与 space 管理端的 requireSuperAdmin 保持一致。
 func (cn *Common) addAppVersion(c *wkhttp.Context) {
-	err := c.CheckLoginRoleIsSuperAdmin()
-	if err != nil {
-		c.ResponseError(err)
+	if err := c.CheckLoginRoleIsSuperAdmin(); err != nil {
+		httperr.ResponseErrorL(c, errcode.ErrSharedForbidden, nil, nil)
 		return
 	}
 	var req appVersionReq
@@ -548,12 +551,11 @@ func (cn *Common) addAppVersion(c *wkhttp.Context) {
 		c.ResponseError(errors.New("请求数据格式有误！"))
 		return
 	}
-	err = cn.check(req)
-	if err != nil {
+	if err := cn.check(req); err != nil {
 		c.ResponseError(err)
 		return
 	}
-	_, err = cn.db.insertAppVersion(&appVersionModel{
+	_, err := cn.db.insertAppVersion(&appVersionModel{
 		AppVersion:  req.AppVersion,
 		OS:          req.OS,
 		IsForce:     req.IsForce,
--- a/modules/common/api.go
+++ b/modules/common/api.go
@@ -533,8 +533,12 @@ func (cn *Common) countriesList(c *wkhttp.Context) {
 }
 
 // 添加app版本
+//
+// 发版会设置客户端的下载来源（DownloadURL），属供应链敏感写操作，因此要求
+// superAdmin —— 普通 admin 定位为只读运营位，不应能改动全体用户的客户端来源。
+// 路由仍保留在 /v1/common（历史调用方依赖），仅收紧角色门槛。
 func (cn *Common) addAppVersion(c *wkhttp.Context) {
-	err := c.CheckLoginRole()
+	err := c.CheckLoginRoleIsSuperAdmin()
 	if err != nil {
 		c.ResponseError(err)
 		return
--- a/modules/common/api_test.go
+++ b/modules/common/api_test.go
@@ -9,10 +9,54 @@ import (
 
 	"github.com/Mininglamp-OSS/octo-lib/config"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
 	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/Mininglamp-OSS/octo-server/pkg/i18n"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+// TestAddAppVersion_RequiresSuperAdmin pins the issue #363 gate: POST
+// /v1/common/appversion (sets the client download source — supply-chain
+// sensitive) must reject a plain admin and only let superAdmin through. The
+// denial goes through the generic forbidden envelope (anti-enumeration).
+func TestAddAppVersion_RequiresSuperAdmin(t *testing.T) {
+	// testutil.NewTestServer already registers every module's routes via
+	// module.Setup, so the appversion handler is live — don't call cn.Route
+	// (double registration panics). It also does not wire the i18n renderer, so
+	// mirror main.go here to get a populated error.code in the envelope.
+	s, ctx := testutil.NewTestServer()
+	defer testutil.CleanAllTables(ctx)
+	s.GetRoute().SetErrorRenderer(i18n.NewErrorRenderer(i18n.NewLocalizer(i18n.DefaultLanguage)))
+
+	cacheCfg := ctx.GetConfig().Cache
+	body := util.ToJson(&appVersionReq{
+		AppVersion:  "9.9",
+		OS:          "android",
+		DownloadURL: "http://example.com/x.apk",
+		IsForce:     1,
+		UpdateDesc:  "x",
+	})
+
+	// plain admin → rejected by the generic forbidden envelope
+	adminTok := "common-appver-admin"
+	require.NoError(t, ctx.Cache().Set(cacheCfg.TokenCachePrefix+adminTok, "admin-uid@admin@"+string(wkhttp.Admin)))
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("POST", "/v1/common/appversion", bytes.NewReader([]byte(body)))
+	req.Header.Set("token", adminTok)
+	s.GetRoute().ServeHTTP(w, req)
+	assert.Contains(t, w.Body.String(), "err.shared.auth.forbidden", "plain admin must be rejected: %s", w.Body.String())
+
+	// superAdmin → passes the role gate (downstream insert outcome is irrelevant
+	// to this assertion; we only prove the gate let it through)
+	superTok := "common-appver-superadmin"
+	require.NoError(t, ctx.Cache().Set(cacheCfg.TokenCachePrefix+superTok, "root-uid@root@"+string(wkhttp.SuperAdmin)))
+	w2 := httptest.NewRecorder()
+	req2, _ := http.NewRequest("POST", "/v1/common/appversion", bytes.NewReader([]byte(body)))
+	req2.Header.Set("token", superTok)
+	s.GetRoute().ServeHTTP(w2, req2)
+	assert.NotContains(t, w2.Body.String(), "err.shared.auth.forbidden", "superAdmin must pass the role gate: %s", w2.Body.String())
+}
 
 func cleanAllTablesAndReloadSettings(t *testing.T, ctx *config.Context) {
 	t.Helper()
--- a/modules/space/api_manager.go
+++ b/modules/space/api_manager.go
@@ -608,6 +608,10 @@ func (m *Manager) updateSpaceProfile(c *wkhttp.Context) {
 // addMembers 管理员强制添加成员（绕过 max_users 限制）。
 // 注意：此操作绕过了 executeJoinSpace 的业务副作用（SpaceMemberJoin 事件、预设群组），
 // 属于 low-level 管理操作；常规入口请走 /v1/space/join。
+//
+// 仍保持 requireAdmin（而非 requireSuperAdmin）是有意为之：强制"添加"是可逆的
+// （随后可移除），而强制解散/封禁/移除/改角色不可逆，故只把后者收紧到 superAdmin。
+// 若日后判定强制加人也需收紧，改成 requireSuperAdmin 即可。
 func (m *Manager) addMembers(c *wkhttp.Context) {
 	if !m.requireAdmin(c) {
 		return
--- a/modules/space/api_manager.go
+++ b/modules/space/api_manager.go
@@ -156,9 +156,23 @@ func toSpaceResp(m *managerSpaceModel) *managerSpaceResp {
 	}
 }
 
-// requireAdmin 统一的 admin/superAdmin 角色检查。未通过时已写入响应，调用方应立即返回。
+// requireAdmin 统一的 admin/superAdmin 角色检查，用于只读 / 低风险管理接口。
+// 未通过时已写入响应，调用方应立即返回。
 func (m *Manager) requireAdmin(c *wkhttp.Context) bool {
 	if err := c.CheckLoginRole(); err != nil {
+		httperr.ResponseErrorL(c, errcode.ErrSharedForbidden, nil, nil)
+		return false
+	}
+	return true
+}
+
+// requireSuperAdmin 仅放行 superAdmin，用于跨空间的高危/不可逆操作（强制解散、
+// 封禁/解禁、强制移除成员、修改成员角色含转让所有权）。这些操作此前只要求
+// admin，等于让"只读运营位"对任意空间拥有上帝模式；收敛到 superAdmin。
+// 返回 ErrSharedForbidden（与 requireAdmin 同一个通用 403），不暴露"需要更高
+// 角色"的具体原因，符合反枚举约定。未通过时已写入响应，调用方应立即返回。
+func (m *Manager) requireSuperAdmin(c *wkhttp.Context) bool {
+	if err := c.CheckLoginRoleIsSuperAdmin(); err != nil {
 		httperr.ResponseErrorL(c, errcode.ErrSharedForbidden, nil, nil)
 		return false
 	}
@@ -309,7 +323,7 @@ func (m *Manager) detail(c *wkhttp.Context) {
 
 // forceDisband 强制解散空间（同时移除全部成员）
 func (m *Manager) forceDisband(c *wkhttp.Context) {
-	if !m.requireAdmin(c) {
+	if !m.requireSuperAdmin(c) {
 		return
 	}
 	spaceId := c.Param("space_id")
@@ -401,7 +415,7 @@ func (m *Manager) members(c *wkhttp.Context) {
 // liftBan 封禁 / 解禁空间：status=1 恢复正常，status=2 置为封禁。
 // status=0（解散）请用 DELETE /space/:space_id。
 func (m *Manager) liftBan(c *wkhttp.Context) {
-	if !m.requireAdmin(c) {
+	if !m.requireSuperAdmin(c) {
 		return
 	}
 	spaceId := c.Param("space_id")
@@ -650,7 +664,7 @@ func (m *Manager) addMembers(c *wkhttp.Context) {
 // 禁止移除 owner——实际检查在 managerDB.removeMembersForce 的事务内用
 // SELECT ... FOR UPDATE 原子完成，避免 handler 层 check 与 update 之间的 TOCTOU。
 func (m *Manager) removeMembers(c *wkhttp.Context) {
-	if !m.requireAdmin(c) {
+	if !m.requireSuperAdmin(c) {
 		return
 	}
 	spaceId := c.Param("space_id")
@@ -714,7 +728,7 @@ func normalizeUIDs(in []string) []string {
 
 // updateMemberRole 修改成员角色；role=2 时自动把当前 owner 降级为 admin。
 func (m *Manager) updateMemberRole(c *wkhttp.Context) {
-	if !m.requireAdmin(c) {
+	if !m.requireSuperAdmin(c) {
 		return
 	}
 	spaceId := c.Param("space_id")
--- a/modules/space/api_manager_test.go
+++ b/modules/space/api_manager_test.go
@@ -19,12 +19,25 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-// adminToken 在 token 缓存中注入一个带 admin 角色的 token，供管理接口测试使用。
+// adminToken 在 token 缓存中注入一个带 admin 角色的 token，供只读 / 低风险管理
+// 接口测试使用。高危/不可逆操作（强制解散、封禁、移除成员、改角色）已收敛到
+// superAdmin，对应测试请用 superAdminToken。
 func adminToken(t *testing.T) string {
 	t.Helper()
 	token := "space-mgr-admin-token"
 	cfg := testCtx.GetConfig()
 	err := testCtx.Cache().Set(cfg.Cache.TokenCachePrefix+token, testutil.UID+"@admin@"+string(wkhttp.Admin))
+	assert.NoError(t, err)
+	return token
+}
+
+// superAdminToken 注入一个带 superAdmin 角色的 token，供仅限超级管理员的高危
+// 管理接口测试使用。
+func superAdminToken(t *testing.T) string {
+	t.Helper()
+	token := "space-mgr-superadmin-token"
+	cfg := testCtx.GetConfig()
+	err := testCtx.Cache().Set(cfg.Cache.TokenCachePrefix+token, testutil.UID+"@superadmin@"+string(wkhttp.SuperAdmin))
 	assert.NoError(t, err)
 	return token
 }
@@ -172,7 +185,7 @@ func TestManager_SpaceDetail(t *testing.T) {
 func TestManager_ForceDisband(t *testing.T) {
 	s, _, err := setup(t)
 	assert.NoError(t, err)
-	token := adminToken(t)
+	token := superAdminToken(t)
 
 	seedSpace(t, "mgr-force", "about to die", "u-owner", 1)
 	err = testSpaceDB.insertMemberNoTx(&MemberModel{
@@ -324,7 +337,7 @@ func TestManager_Members_KeywordSearch(t *testing.T) {
 func TestManager_LiftBan(t *testing.T) {
 	s, _, err := setup(t)
 	assert.NoError(t, err)
-	token := adminToken(t)
+	token := superAdminToken(t)
 
 	seedSpace(t, "mgr-ban-target", "to be banned", "u-owner", SpaceStatusNormal)
 
@@ -426,7 +439,7 @@ func TestManager_AddMembers(t *testing.T) {
 func TestManager_RemoveMembers(t *testing.T) {
 	s, _, err := setup(t)
 	assert.NoError(t, err)
-	token := adminToken(t)
+	token := superAdminToken(t)
 
 	seedSpace(t, "mgr-rm", "remove members", "u-owner", SpaceStatusNormal)
 	for _, uid := range []string{"rm-1", "rm-2", "rm-3"} {
@@ -468,7 +481,7 @@ func TestManager_RemoveMembers(t *testing.T) {
 func TestManager_UpdateMemberRole(t *testing.T) {
 	s, _, err := setup(t)
 	assert.NoError(t, err)
-	token := adminToken(t)
+	token := superAdminToken(t)
 
 	seedSpace(t, "mgr-role", "role ops", "u-owner", SpaceStatusNormal)
 	err = testSpaceDB.insertMemberNoTx(&MemberModel{
@@ -1003,6 +1016,39 @@ func TestManager_AuthBoundary(t *testing.T) {
 	})
 }
 
+// TestManager_DestructiveOpsRequireSuperAdmin 固化 issue #363 item 2 的边界：
+// 强制解散 / 封禁 / 移除成员 / 改成员角色这些跨空间高危操作只放行 superAdmin，
+// 普通 admin（只读运营位）一律 403。superAdmin 放行路径由各操作的成功用例覆盖。
+func TestManager_DestructiveOpsRequireSuperAdmin(t *testing.T) {
+	s, _, err := setup(t)
+	assert.NoError(t, err)
+	token := adminToken(t) // 仅 admin，非 superAdmin
+
+	seedSpace(t, "mgr-su-guard", "guard", "u-owner", SpaceStatusNormal)
+
+	cases := []struct {
+		name   string
+		method string
+		path   string
+		body   string
+	}{
+		{"forceDisband", "DELETE", "/v1/manager/spaces/mgr-su-guard", ""},
+		{"ban", "PUT", "/v1/manager/spaces/mgr-su-guard/status/2", ""},
+		{"removeMembers", "DELETE", "/v1/manager/spaces/mgr-su-guard/members", `{"uids":["u-owner"]}`},
+		{"updateMemberRole", "PUT", "/v1/manager/spaces/mgr-su-guard/members/u-owner/role", `{"role":1}`},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			w := httptest.NewRecorder()
+			req, _ := http.NewRequest(tc.method, tc.path, strings.NewReader(tc.body))
+			req.Header.Set("token", token)
+			s.GetRoute().ServeHTTP(w, req)
+			// admin 被拒：通用 403，不泄露"需要更高角色"的具体原因（反枚举）。
+			assertSpaceErrorCode(t, w, "err.shared.auth.forbidden")
+		})
+	}
+}
+
 func TestManager_DisableListIncludesBanned(t *testing.T) {
 	s, _, err := setup(t)
 	assert.NoError(t, err)
@@ -1104,7 +1150,7 @@ func TestManager_LiftBanRefreshesCache(t *testing.T) {
 	// 验证 liftBan 成功后会异步调用 loadKnownSpaceIDs 刷新 pkg/space 缓存
 	s, _, err := setup(t)
 	assert.NoError(t, err)
-	token := adminToken(t)
+	token := superAdminToken(t)
 
 	seedSpace(t, "mgr-ban-cache", "cache", "u-owner-c", SpaceStatusBanned)
 
@@ -1140,7 +1186,7 @@ func TestManager_AddMembersOnDisbandedSpace(t *testing.T) {
 func TestManager_RemoveMembersOnNonExistentSpace(t *testing.T) {
 	s, _, err := setup(t)
 	assert.NoError(t, err)
-	token := adminToken(t)
+	token := superAdminToken(t)
 
 	body := util.ToJson(map[string]interface{}{"uids": []string{"any"}})
 	w := httptest.NewRecorder()
@@ -1708,7 +1754,7 @@ func TestManager_UpdateSpaceProfile_Validation(t *testing.T) {
 func TestManager_UpdateMemberRoleIdempotent(t *testing.T) {
 	s, _, err := setup(t)
 	assert.NoError(t, err)
-	token := adminToken(t)
+	token := superAdminToken(t)
 
 	seedSpace(t, "mgr-role-idem", "role idem", "u-owner", SpaceStatusNormal)
 	err = testSpaceDB.insertMemberNoTx(&MemberModel{
--- a/modules/user/api_manager.go
+++ b/modules/user/api_manager.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/Mininglamp-OSS/octo-lib/common"
 	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/cache"
 	"github.com/Mininglamp-OSS/octo-server/modules/base/event"
 	common2 "github.com/Mininglamp-OSS/octo-server/modules/common"
 	"github.com/Mininglamp-OSS/octo-server/pkg/auth"
@@ -330,43 +331,72 @@ func (m *Manager) deleteAdminUsers(c *wkhttp.Context) {
 		respondUserError(c, errcode.ErrUserStoreFailed)
 		return
 	}
-	// 撤销该管理员在所有设备端（APP/Web/PC）的登录态，而不只是 Web —— 否则一个
-	// 仍存活的非 Web token 在 RoleCacheTTL 窗口内仍可能按旧角色放行。
+	// 先失效角色热缓存，再撤销 token。DB 里该 uid 已删（权威源 role 已为空），所以
+	// 只要这里清掉 user_role:{uid} 热缓存，即便下面的 token 撤销失败、某个 token 残
+	// 留，下一请求经 RoleResolver 也会从 DB 解析出"无系统角色"而被拒。顺序反过来
+	// （先撤 token，失败就 return）会跳过 Invalidate，让旧 role 缓存存活到 TTL，
+	// 与残留 token 叠加成提权窗口——正是本 PR 要消除的（PR #364 review）。
+	m.roleService.Invalidate(user.UID)
+	// 撤销该管理员在所有设备端（APP/Web/PC）的登录态，而不只是 Web。尽力删除每个端
+	// （不在首个错误就中断），最大化吊销面；任一失败仍向调用方报错以便排查。
 	if err := m.revokeAllDeviceTokens(user.UID); err != nil {
 		m.Error("清除管理员token数据错误", zap.Error(err), zap.String("uid", user.UID))
 		respondUserError(c, errcode.ErrUserTokenCacheFailed)
 		return
 	}
-	// 立即失效角色热缓存：即便上面漏删了某个 token，下一请求经 RoleResolver 也会
-	// 从 DB（该 uid 已删，role 为空）解析出"无系统角色"，把降权窗口收敛到一次往返
-	// 而非等 RoleCacheTTL 自愈。
-	m.roleService.Invalidate(user.UID)
 	c.ResponseOK()
 }
 
-// revokeAllDeviceTokens 清除某 uid 在所有设备端（APP/Web/PC）的登录态：既删
-// token:{token} 让旧 token 立即失效，也删 UIDToken:{flag}{uid} 反查映射，避免
-// 残留。任一删除失败即返回错误，调用方据此报错（不静默吞，以免"以为已吊销实则
-// 残留"）。
-func (m *Manager) revokeAllDeviceTokens(uid string) error {
-	cacheCfg := m.ctx.GetConfig().Cache
+// deviceTokenRevokeFailure pairs a device flag with the error from trying to
+// revoke its session, so the caller can log per-device outcomes.
+type deviceTokenRevokeFailure struct {
+	flag config.DeviceFlag
+	err  error
+}
+
+// revokeDeviceTokensInCache is the cache-only, injectable core of
+// revokeAllDeviceTokens (factored out so it can be unit-tested with a fake
+// cache). It removes uid's token payload + UIDToken reverse mapping for every
+// device flag. best-effort: it does NOT abort on the first error — it attempts
+// every flag and returns one failure entry per error, so a Redis hiccup on one
+// device cannot strand the others.
+func revokeDeviceTokensInCache(c cache.Cache, uidTokenPrefix, tokenPrefix, uid string) []deviceTokenRevokeFailure {
+	var failures []deviceTokenRevokeFailure
 	for _, flag := range []config.DeviceFlag{config.APP, config.Web, config.PC} {
-		uidKey := fmt.Sprintf("%s%d%s", cacheCfg.UIDTokenCachePrefix, flag, uid)
-		token, err := m.ctx.Cache().Get(uidKey)
+		uidKey := fmt.Sprintf("%s%d%s", uidTokenPrefix, flag, uid)
+		token, err := c.Get(uidKey)
 		if err != nil {
-			return err
+			failures = append(failures, deviceTokenRevokeFailure{flag, err})
+			continue
 		}
 		if token == "" {
 			continue
 		}
-		if err := m.ctx.Cache().Delete(cacheCfg.TokenCachePrefix + token); err != nil {
-			return err
+		if err := c.Delete(tokenPrefix + token); err != nil {
+			failures = append(failures, deviceTokenRevokeFailure{flag, err})
+			continue
 		}
-		if err := m.ctx.Cache().Delete(uidKey); err != nil {
-			return err
+		if err := c.Delete(uidKey); err != nil {
+			failures = append(failures, deviceTokenRevokeFailure{flag, err})
 		}
 	}
-	return nil
+	return failures
+}
+
+// revokeAllDeviceTokens 清除某 uid 在所有设备端（APP/Web/PC）的登录态：既删
+// token:{token} 让旧 token 立即失效，也删 UIDToken:{flag}{uid} 反查映射。
+// best-effort（见 revokeDeviceTokensInCache），逐端记日志，返回首个错误供调用方报错。
+func (m *Manager) revokeAllDeviceTokens(uid string) error {
+	cacheCfg := m.ctx.GetConfig().Cache
+	failures := revokeDeviceTokensInCache(m.ctx.Cache(), cacheCfg.UIDTokenCachePrefix, cacheCfg.TokenCachePrefix, uid)
+	var firstErr error
+	for _, f := range failures {
+		m.Error("撤销设备token失败", zap.Error(f.err), zap.String("uid", uid), zap.Uint8("device_flag", f.flag.Uint8()))
+		if firstErr == nil {
+			firstErr = f.err
+		}
+	}
+	return firstErr
 }
 
 // 查询管理员列表
--- a/modules/user/api_manager.go
+++ b/modules/user/api_manager.go
@@ -36,6 +36,7 @@ type Manager struct {
 	friendDB      *friendDB
 	onlineService IOnlineService
 	commonService common2.IService
+	roleService   *RoleService
 }
 
 // NewManager NewManager
@@ -50,6 +51,7 @@ func NewManager(ctx *config.Context) *Manager {
 		userSettingDB: NewSettingDB(ctx.DB()),
 		onlineService: NewOnlineService(ctx),
 		commonService: common2.NewService(ctx),
+		roleService:   NewRoleService(NewDB(ctx), ctx.Cache()),
 	}
 	m.createManagerAccount()
 	return m
@@ -328,21 +330,43 @@ func (m *Manager) deleteAdminUsers(c *wkhttp.Context) {
 		respondUserError(c, errcode.ErrUserStoreFailed)
 		return
 	}
-	oldToken, err := m.ctx.Cache().Get(fmt.Sprintf("%s%d%s", m.ctx.GetConfig().Cache.UIDTokenCachePrefix, config.Web, user.UID))
-	if err != nil {
-		m.Error("获取旧token错误", zap.Error(err))
+	// 撤销该管理员在所有设备端（APP/Web/PC）的登录态，而不只是 Web —— 否则一个
+	// 仍存活的非 Web token 在 RoleCacheTTL 窗口内仍可能按旧角色放行。
+	if err := m.revokeAllDeviceTokens(user.UID); err != nil {
+		m.Error("清除管理员token数据错误", zap.Error(err), zap.String("uid", user.UID))
 		respondUserError(c, errcode.ErrUserTokenCacheFailed)
 		return
 	}
-	if oldToken != "" {
-		err = m.ctx.Cache().Delete(m.ctx.GetConfig().Cache.TokenCachePrefix + oldToken)
+	// 立即失效角色热缓存：即便上面漏删了某个 token，下一请求经 RoleResolver 也会
+	// 从 DB（该 uid 已删，role 为空）解析出"无系统角色"，把降权窗口收敛到一次往返
+	// 而非等 RoleCacheTTL 自愈。
+	m.roleService.Invalidate(user.UID)
+	c.ResponseOK()
+}
+
+// revokeAllDeviceTokens 清除某 uid 在所有设备端（APP/Web/PC）的登录态：既删
+// token:{token} 让旧 token 立即失效，也删 UIDToken:{flag}{uid} 反查映射，避免
+// 残留。任一删除失败即返回错误，调用方据此报错（不静默吞，以免"以为已吊销实则
+// 残留"）。
+func (m *Manager) revokeAllDeviceTokens(uid string) error {
+	cacheCfg := m.ctx.GetConfig().Cache
+	for _, flag := range []config.DeviceFlag{config.APP, config.Web, config.PC} {
+		uidKey := fmt.Sprintf("%s%d%s", cacheCfg.UIDTokenCachePrefix, flag, uid)
+		token, err := m.ctx.Cache().Get(uidKey)
 		if err != nil {
-			m.Error("清除旧token数据错误", zap.Error(err))
-			respondUserError(c, errcode.ErrUserTokenCacheFailed)
-			return
+			return err
+		}
+		if token == "" {
+			continue
+		}
+		if err := m.ctx.Cache().Delete(cacheCfg.TokenCachePrefix + token); err != nil {
+			return err
+		}
+		if err := m.ctx.Cache().Delete(uidKey); err != nil {
+			return err
 		}
 	}
-	c.ResponseOK()
+	return nil
 }
 
 // 查询管理员列表
--- a/modules/user/api_manager_test.go
+++ b/modules/user/api_manager_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
@@ -748,5 +749,30 @@ func TestManager_DeleteAdminUser_RevokesSessionsAndRoleCache(t *testing.T) {
 		uidMap, err := ctx.Cache().Get(fmt.Sprintf("%s%d%s", cacheCfg.UIDTokenCachePrefix, flag, adminUID))
 		assert.NoError(t, err)
 		assert.Empty(t, uidMap, "UIDToken mapping (flag %d) must be cleared", flag)
+	}
+}
+
+// TestRevokeDeviceTokensInCache_BestEffort pins that a Redis error on one device
+// flag does not abort revocation of the others (PR #364 review F2): the loop
+// attempts all three flags and reports a failure per error rather than bailing
+// on the first. Uses the shared fakeLangCache (implements cache.Cache).
+func TestRevokeDeviceTokensInCache_BestEffort(t *testing.T) {
+	c := newFakeLangCache()
+	const uidTokenPrefix, tokenPrefix, uid = "UIDTOKEN:", "TOKEN:", "u1"
+	for _, fl := range []config.DeviceFlag{config.APP, config.Web, config.PC} {
+		tok := fmt.Sprintf("tok-%d", fl)
+		c.store[fmt.Sprintf("%s%d%s", uidTokenPrefix, fl, uid)] = tok
+		c.store[tokenPrefix+tok] = uid + "@x@admin"
+	}
+	// Every Delete fails: a fail-fast loop would stop after flag 0; best-effort
+	// must still attempt all three.
+	c.delErr = errors.New("redis down")
+
+	failures := revokeDeviceTokensInCache(c, uidTokenPrefix, tokenPrefix, uid)
+	if len(failures) != 3 {
+		t.Fatalf("best-effort revoke must attempt all 3 device flags despite errors, got %d failures", len(failures))
+	}
+	if len(c.deletes) < 3 {
+		t.Fatalf("expected a Delete attempt per device flag, got %d", len(c.deletes))
 	}
 }
--- a/modules/user/api_manager_test.go
+++ b/modules/user/api_manager_test.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/Mininglamp-OSS/octo-lib/config"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
 	"github.com/Mininglamp-OSS/octo-lib/testutil"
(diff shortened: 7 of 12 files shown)


Jerry-Xin (dismissed):
This PR is in scope for `octo-server` and the authorization hardening is directionally sound; I found no merge-blocking bugs.

💬 Non-blocking

🟡 Warning — [modules/common/api.go:541](/Users/mlamp/.openclaw/tmp/pr-review-k41j1s9w/repo/modules/common/api.go:541): the newly tightened `addAppVersion` gate still returns `c.ResponseError(err)` on authorization failure. That preserves the legacy, non-localized error path and differs from the generic i18n forbidden response used in the new space manager helper. Consider switching this denial path to `httperr.ResponseErrorL(c, errcode.ErrSharedForbidden, nil, nil)`.

🔵 Suggestion — [modules/space/api_manager_test.go:24](/Users/mlamp/.openclaw/tmp/pr-review-k41j1s9w/repo/modules/space/api_manager_test.go:24): the manager route tests seed token roles directly, but the test server does not wire the new `RoleResolver` like `main.go` does. The parser and role service unit tests cover the pieces, but an end-to-end route test with `WithRoleResolver` would better prove that production routes honor DB-backed demotion.

🔵 Suggestion — [modules/user/api_manager.go:325](/Users/mlamp/.openclaw/tmp/pr-review-k41j1s9w/repo/modules/user/api_manager.go:325): `deleteAdminUsers` does not invalidate `user_role:{uid}` after deleting the admin row. TTL-bound convergence is acceptable per the PR design, but calling the new invalidation path here would make admin removal immediate and would exercise the exported cache key/service contract.

✅ Highlights

The destructive space operations are consistently moved behind `requireSuperAdmin`, and the shared helper keeps the response generic.

The parser behavior is well covered for override, empty-role demotion, fail-open, and nil resolver cases.

Focused tests passed for `./pkg/auth` and the new role-service tests in `./modules/user`; full `./modules/user` was blocked by the local test DB migration state, not by this PR’s focused tests.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#364 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/364
**Head SHA:** `6828c5dbf57f2a577653638bdf62439c49a7bb82`
**Title:** security: per-request role resolution + tighten high-impact management gates

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Per-request role resolution in parser | ✅ | `pkg/auth/parser.go:139-155` — resolver overrides token snapshot; fail-open on error; empty drops role |
| RoleService cache→DB lookup | ✅ | `modules/user/role_service.go:72-93` — correct miss/negative/hit switch; DB error propagated |
| QueryRoleByUID single-column read | ✅ | `modules/user/db.go:135-140` — `SELECT role FROM user WHERE uid=?`; returns `("", nil)` for missing uid |
| superAdmin gate on addAppVersion | ✅ | `modules/common/api.go:541` — `CheckLoginRoleIsSuperAdmin()` replaces `CheckLoginRole()` |
| requireSuperAdmin on 4 destructive ops | ✅ | `modules/space/api_manager.go:326,418,667,731` — forceDisband, liftBan, removeMembers, updateMemberRole |
| requireSuperAdmin helper | ✅ | `modules/space/api_manager.go:176-183` — mirrors requireAdmin, returns generic 403 (anti-enumeration) |
| Wiring in main.go | ✅ | `main.go:124,129` — `NewRoleService` → `WithRoleResolver` → `CacheTokenParser` |
| Test: parser role override | ✅ | `pkg/auth/parser_test.go:244-268` — snapshot superAdmin overridden to resolver's "admin" |
| Test: empty drops role | ✅ | `pkg/auth/parser_test.go:272-289` — token admin dropped to "" when resolver returns "" |
| Test: fail-open | ✅ | `pkg/auth/parser_test.go:293-311` — resolver error preserves token snapshot |
| Test: destructive ops guard | ✅ | `modules/space/api_manager_test.go:1022-1053` — admin gets 403 on all 4 endpoints |
| Test: RoleService unit | ✅ | `modules/user/role_service_test.go` — cache hit/miss/negative/DB error/invalidate/nil-panic |

## 2. Findings

### P2-1: `RoleService.Invalidate()` defined but never wired to production mutation sites

**Severity:** P2 (design gap, not a security vulnerability)
**Diff-scope:** new (the Invalidate method is introduced by this PR but not consumed)

`RoleService.Invalidate()` (`role_service.go:100-104`) is defined and tested, but no production code path calls it. The `userRoleSvc` is a local variable in `main.go:124` passed only to the token parser — the `user.Manager` struct (which owns `deleteAdminUsers` at `modules/user/api_manager.go:296`) has no reference to `RoleService`.

**Impact analysis:**
- The only admin-demotion path is `deleteAdminUsers`, which **deletes the entire user row** AND **deletes the token cache entry** (`api_manager.go:337-349`). Without a valid token in cache, `CacheTokenParser.Parse` returns `ErrTokenNotFound` before the role resolver ever runs. The stale `user_role:{uid}` cache key is therefore unreachable during its 60s TTL.
- No "demote admin to normal user" endpoint exists — only full account deletion. So there is no scenario where a live token coexists with a stale role cache for a still-existing user.
- `addAdminUser` creates a new user with no pre-existing cache entry — no invalidation needed.

**Why P2 not P1:** The token cache deletion in `deleteAdminUsers` prevents authentication regardless of the stale role cache. No path exists where a user with a valid token can exercise stale admin privileges. The gap is cosmetic/future-proofing, not a live security issue.

**Suggestion:** If a future PR adds a "demote admin" endpoint (change `user.role` from "admin" to "" without deleting the account), `Invalidate()` must be wired in at that point. Consider adding a comment at `deleteAdminUsers` noting this dependency. Alternatively, inject `*RoleService` into `user.Manager` now and call `Invalidate` in `deleteAdminUsers` for defense-in-depth.

### P2-2: Fail-open during Redis/DB outage preserves stale admin role (documented trade-off)

**Severity:** P2 (intentional design choice, not a defect)
**Diff-scope:** new (introduced by this PR's resolver logic)

When `RoleService.ResolveRole` returns an error (Redis down + DB down simultaneously), `CacheTokenParser.Parse` (`parser.go:152-154`) keeps the token's snapshot role. If a user was admin when the token was minted but has since been demoted, they retain admin access for the duration of the outage (or token lifetime, whichever is shorter).

**Why P2 not P1:** This is the same fail-open philosophy as the existing `LanguageResolver` — authentication must not 5xx on infrastructure blips. The scenario requires simultaneous Redis AND DB failure, which is rare and transient. The alternative (fail-closed = deny all admin access during outage) is arguably worse for operations.

## 3. Suggestions

1. **Wire `Invalidate()` at `deleteAdminUsers`** (P2-1 follow-up): Inject `*RoleService` into `user.Manager` or pass it via constructor. Call `roleSvc.Invalidate(uid)` after the DB delete succeeds. Low effort, defense-in-depth.
2. **Consider a stale-role metric**: Log or emit a counter when the resolver overrides a token snapshot with a different role. This gives ops visibility into how often demotions are being honored vs. how often fail-open kicks in.

## 4. Additional Observations

- **No demotion endpoint exists**: The codebase only has `deleteAdminUsers` (full account deletion) and `addAdminUser`. There is no "change user.role from admin to normal" endpoint. This means the per-request resolver's primary value today is catching the (rare) case where the DB is directly modified. The resolver becomes more valuable when/if a demotion endpoint is added.
- **`addAppVersion` error response difference**: `modules/common/api.go:543` uses `c.ResponseError(err)` which may leak the specific "not superAdmin" error message, while `modules/space/api_manager.go:178` uses `httperr.ResponseErrorL(c, errcode.ErrSharedForbidden, nil, nil)` which returns a generic 403. The space manager approach is better for anti-enumeration. Consider aligning `addAppVersion` to use the same generic 403 pattern. (P2/nit — the common module's existing pattern, not introduced by this PR.)

## 5. Data-Flow Trace

### Token parse → role resolution (happy path)
```
CacheTokenParser.Parse (parser.go:97)
  → Cache.Get(Prefix + token) → encoded token string
  → Decode(raw) → TokenInfo{UID, Name, Role: "admin" (snapshot)}
  → roleResolver.ResolveRole(ctx, uid)
      → RoleService.ResolveRole (role_service.go:72)
          → cache.Get("user_role:" + uid)
              miss ("", nil) → fall through
              hit "-" (negative marker) → return ""
              hit "admin"/"superAdmin" → return cached role
          → DB.QueryRoleByUID(uid) (db.go:138)
              SELECT role FROM user WHERE uid=?
              → ("admin", nil) / ("", nil) for normal user
          → writeCache(key, role) → SetAndExpire 60s
      → return resolved role
  → role = resolved (overrides snapshot)
  → return UserInfo{Role: resolved}
  → AuthMiddleware sets c.Set("role", info.Role) (octo-lib http.go:334)
  → Handler calls CheckLoginRole/CheckLoginRoleIsSuperAdmin
```

### Cache miss semantics verified
- `cache.Cache.Get` returns `("", nil)` on miss (confirmed: `octo-lib/pkg/redis/redis.go:100-110` swallows `rd.Nil` → `("", nil)`)
- `RoleService` correctly distinguishes miss (`cached == ""` → DB) from negative hit (`cached == "-"` → skip DB) from positive hit (default → return cached)

### Fail-open path verified
- `ResolveRole` returns `("", fmt.Errorf(...))` on DB error (role_service.go:91-93)
- Parser checks `rerr == nil` before overriding (parser.go:152) → on error, keeps `info.Role` (snapshot)

**C1 — Dual-path parity:** CLEAR
- `addAdminUser` (INSERT new user, role=admin) ↔ `deleteAdminUsers` (DELETE user row + token cache). No demotion path exists. Token cache deletion in `deleteAdminUsers` covers the auth-blocking requirement. No asymmetric gap.
- `requireAdmin` ↔ `requireSuperAdmin`: structurally identical helpers, correct separation of read vs destructive ops.

**C2 — Control-flow ordering / nesting reuse:** CLEAR
- Role resolver runs after token decode, before UserInfo construction. No double-resolution or nesting issue.
- No security control (regex/escape/sanitize) involved — role is a simple string comparison.

**C3 — Authorization boundary ≠ capability boundary:** CLEAR
- Role gates (`CheckLoginRole`, `CheckLoginRoleIsSuperAdmin`) are applied at the handler level for all authenticated endpoints via `AuthMiddleware`. The per-request resolver feeds into the same `c.Set("role", ...)` path, so all downstream gates see the resolved role. No endpoint bypasses the middleware.

---

[Octo-Q] verdict: **APPROVE**

No P0/P1 issues found. The PR correctly implements per-request role resolution with proper fail-open semantics, tightens destructive operation gates to superAdmin, and provides comprehensive test coverage. Two P2 observations (unwired Invalidate, fail-open trade-off) are design considerations for follow-up, not landing blockers.

yujiawei (dismissed):
# Code Review — PR #364 (octo-server)

Reviewed at head `6828c5db` against `main`. Scope: per-request system-role resolution (`pkg/auth`, `modules/user`) plus tightening four cross-space destructive management ops and `addAppVersion` from `admin` to `superAdmin`.

**Verdict: APPROVE.** This is a strict, well-tested security improvement with no regressions. The one substantive concern (revocation latency / unwired `Invalidate`) is a TTL-bounded, explicitly-documented tradeoff and a strong follow-up recommendation, not a merge blocker. Because this touches an auth hot path, I've flagged below the specific points a human security owner should consciously accept.

## Summary of verification

| Area | Result |
|---|---|
| Role override logic (`parser.go` Parse) | ✅ Correct. Snapshot kept only on resolver error (fail-open); resolver `""` with no error correctly drops a stale role; fresh role wins otherwise. |
| `ResolveRole` branches (hit / negative-marker / clean-miss / cache-error / DB-error) | ✅ Correct. Redis miss returns `("",nil)` (not an error) so a normal user falls through to DB then caches the negative marker; cache error degrades to DB (authoritative); DB error propagates → parser fail-open. |
| `QueryRoleByUID` (single-column select) | ✅ Safe. `role` column is `NOT NULL DEFAULT ''`; zero/multiple rows yield `("",nil)` / first-row scalar — no spurious error, no NULL scan. |
| Gate tightening (`forceDisband`/`liftBan`/`removeMembers`/`updateMemberRole` → superAdmin) | ✅ Applied consistently; new `TestManager_DestructiveOpsRequireSuperAdmin` asserts plain admin gets a generic 403 (anti-enumeration preserved), and the four happy-path tests were moved to a superAdmin token. |
| Build / tests | ✅ `go build ./pkg/auth ./modules/user ./modules/space ./modules/common` clean; new unit suites pass. |
| Scope | ✅ Exactly 9 files, all within the stated change. No scope creep. |

## Findings

### P1 (recommended before merge, but non-blocking) — Revocation is TTL-bounded (≤60s), and the `Invalidate` immediate-effect path is dead code

`RoleService.Invalidate` (`modules/user/role_service.go:96`) is defined and documented ("callers that demote/remove an admin should additionally invalidate the hot key … for immediate effect", lines 43-46) but has **zero production callers** — `deleteAdminUsers` (`modules/user/api_manager.go:296`) and space `updateMemberRole` never call it. Because `ResolveRole` returns a positive cache hit before consulting the DB (`role_service.go:75-84`), a removed/demoted admin whose `user_role:{uid}` key is hot retains the cached role for up to `RoleCacheTTL` (60s).

`deleteAdminUsers` already deletes the victim's **Web** device token (`api_manager.go:331`, `UIDTokenCachePrefix + config.Web`), so the primary admin-console session dies immediately. The residual is any **non-Web (App/PC) token** presented inside the 60s window, which would be re-granted admin from the stale cache.

Why this is **not** a blocker: this PR is a strict improvement — before it, the role was baked into the token for its full lifetime (days) with no per-request check; after it, the worst case is 60s, and that 60s is explicitly documented as the accepted SLA (`role_service.go:16-21`). There is no regression on any path.

Recommended fast-follow:
1. Wire `RoleService` into `deleteAdminUsers` and call `Invalidate(uid)` after the DB mutation (matches the existing language-service active-invalidation precedent).
2. In `deleteAdminUsers`, also kill non-Web device tokens (or all device tokens) so removal is truly immediate across session types, not just Web.

*For the human security owner:* please consciously accept (or reject) the ≤60s non-Web-token admin-retention window after an admin removal. Everything else here is a strict net improvement.

### P2 — Asymmetric gate: `removeMembers` elevated to superAdmin, but `addMembers` (force-add to any space) stays at admin

`removeMembers` is now superAdmin-only (`api_manager.go:667`) while its counterpart `addMembers` (force-add, `api_manager.go:612`) and `updateSpaceProfile` / `approveJoinApply` remain at `requireAdmin`. Force-add is reversible and force-remove is destructive, so drawing the line here is defensible — but if the goal is "admin is no longer god-mode over arbitrary spaces", a plain admin force-adding arbitrary UIDs into any space partially reopens that surface. Either elevate `addMembers` too, or document the threat-model rationale for the asymmetry. (No `addMembers`-stays-at-admin negative test exists to prove this is by-design vs. oversight.)

### P2 — `addAppVersion` superAdmin gate has no test coverage

The only test exercising `addAppVersion` (`modules/common/api_test.go:23`) is `t.Skip`-ed (pre-existing migration TODO), and its `testutil.Token` carries an empty role, so if un-skipped it would now fail under the new superAdmin gate. The four space-module gate changes got full allow+deny coverage; this one did not. Add a small test (superAdmin → 200, admin → 403) mirroring the space pattern, or at minimum annotate the skipped test that its token must become superAdmin before un-skipping.

### P2 — `ResolveRole` cache-error fall-through is correct but untested

When `cache.Get` returns a non-nil error, `ResolveRole` correctly skips the cache switch and reads DB truth (`role_service.go:75-92`). The fake cache already supports a `getErr` field, but no role-service test sets it. This is the security-sensitive degradation branch; add a test asserting a cache error still returns the DB role with `queryCalls==1` to lock the contract.

### P2 — Latent: negative-marker + future "promote existing user" path

If a future endpoint sets `user.role` on an **existing** UID (today only `addAdminUser` writes the role, and it does so on a brand-new UID), a cached negative marker `-` would silently suppress the new grant for up to 60s. Not reachable on this diff, but worth a comment near `roleNegativeMarker` so the next author wires `Invalidate` into any such path.

### P2 — `addAppVersion` retains legacy `c.ResponseError`

The PR touches this handler's gate but leaves the legacy raw error response (bypasses the localized error envelope). This is pre-existing and consistent with the rest of the unmigrated `common` module — not introduced here and not a CI break — flagging only so it isn't mistaken for compliant. Optional cleanup while in the file.

## Things done well

- Fail-open philosophy mirrors the already-shipped `LanguageResolver` exactly, with clear comments explaining each branch.
- Negative-marker design keeps normal-user traffic (the overwhelming majority) off MySQL while still bounding staleness.
- Anti-enumeration preserved: both `requireAdmin` and `requireSuperAdmin` return the same generic 403, and there's a test asserting it.
- `NewRoleService` panics on nil deps at construction rather than degrading at request time.

lml2468 (dismissed):
✅ APPROVE — security 修复扎实,核心目标(关闭 stale-role 提权 + 收紧高危管理操作)达成,本地逐条核实 + pkg/auth 测试全过 + CI Test 绿(head 6828c5d)。pr-review workflow 75 agent / 5 维 / 对抗验证。

## 核心机制验证(本地核实)
1. **per-request 角色解析**(parser.go:153-157)— 每请求经 roleResolver.ResolveRole 从 cache(60s TTL)→DB 取权威 role,覆盖 token 内快照。三态正确:resolver error → 保留快照(fail-open,认证不因 cache 抖动 5xx);resolved=="" → 丢弃快照 role(即时降权,即使 token 里是 admin/superAdmin);resolved!="" → 用新值。✅
2. **全局挂载无绕过**(main.go:125 SetTokenParser + WithRoleResolver)— 所有 /v1/* 经 AuthMiddleware 走同一 parser;空间管理 25+ 路由统一在 AuthMiddleware group(api_manager.go:79)下。✅
3. **高危操作收紧**(api_manager.go)— forceDisband / liftBan / removeMembers / updateMemberRole 四个不可逆跨空间操作从 requireAdmin 提到 requireSuperAdmin;common/api.go addAppVersion(供应链敏感:改全端下载 URL)同样提到 superAdmin。requireSuperAdmin 读的是 parser 解析后的权威 role。✅
4. **回归测试**(api_manager_test.go TestManager_DestructiveOpsRequireSuperAdmin)四个操作用 plain-admin token 全部 403;parser_test.go 18 项(含 override / 降权丢弃 / fail-open)本地全过;CI Test 绿。✅

## 安全分析:无 blocking
- **legacy token role 伪造?** 已核否决:client token 只是 cache key,`uid@name@role` 串由 Cache.Get 从服务端 Redis 取(mint 时写入),客户端无法伪造该串;即便能,resolver 也会覆盖。非威胁。
- **fail-open(resolver error 保留快照)** — 设计取舍(认证可用性优先),被 60s TTL + 这是降级而非提权所限,可接受。

## should-fix(非 blocking,建议后续收)
1. 🟡 **角色变更未调 Invalidate**:RoleService.Invalidate(uid) 已实现,但 user 模块 addAdminUser/deleteAdminUsers 未注入 RoleService、未在改 role 后调用 → 降权后最长 60s 内 RoleService uid-cache 仍返回旧 role。注:deleteAdminUsers 已即时删 token cache,所以旧 token 立即失效;残留的只是「重新登录后 60s 内拿到旧 role」,有界且自愈。建议把 RoleService 注入 user.Manager,在 role 突变处调 Invalidate,把窗口降到一次往返。
2. 🟡 **缺端到端降权测试**:role_service_test 覆盖 cache→DB→negative,但无「admin 发 token → DB 降权 → 旧 token 后续请求被拒」的串联测试 + Invalidate 调用点测试。

## CI 三个非代码失败(与本 PR 无关)
check-sprint(缺 Closes #issue)/ dependency-review(仓库没开 Dependency graph)/ secret-scan(全树历史噪声)—— 流程/仓库配置闸,需作者补 Closes# + maintainer 配置;关键代码门 Build/Test/Lint/Vet 全绿。

无 blocking,should-fix 可后续跟进。批准。

—— Octo-PR (lml2468)

lml2468 (approved):
✅ APPROVE(重审 @7aac498)— 上一轮四人提的 should-fix 全部收齐,本地逐条核实 + build/vet/auth+user 测试全绿。作者还额外修了一个我漏掉的真问题。

## 上轮 should-fix 全部已修
1. ✅ **Invalidate 接入**(user/api_manager.go:54,343)— RoleService 注入 Manager,deleteAdminUsers 改 DB 后调 Invalidate(user.UID),把降权窗口从 60s TTL 收敛到一次往返。
2. ✅ **多设备 token 吊销(超预期)**— 原 deleteAdminUsers 只删 Web token,新增 revokeAllDeviceTokens 删 APP/Web/PC 全端 + UIDToken 反查映射,任一失败即报错不静默。**这是比我上轮标注更实的提权窗口(非 Web token 在 TTL 内仍按旧角色放行),作者自查发现并修复。**
3. ✅ **i18n 统一错误**(common/api.go)— addAppVersion 拒绝路径改走 httperr.ResponseErrorL(ErrSharedForbidden) 通用 403,反枚举,与 space requireSuperAdmin 一致。
4. ✅ **addMembers 权限不对称** — 作者给出明确理由(强制加人可逆、解散/封禁/移除/改角色不可逆),保留 requireAdmin 是有意为之,合理(yujiawei 上轮提的对称性问题)。
5. ✅ **文档注释** — role_service.go 给未来作者加了「任何 MUTATE 既有 uid.role 的路径必须调 Invalidate」的醒目说明。

## 核实新引入无隐患
- addAdminUser 注释称「写全新 uid 不必 Invalidate」— 本地核实成立:用 util.GenerUUID() 建全新 uid(非提升既有用户,且 ErrUserAlreadyExists 拦重名),全新 uid 无缓存条目,无 negative-marker 压制风险。✅
- revokeAllDeviceTokens 遍历 DeviceFlag{APP,Web,PC},token 空跳过、删除失败返回 error,逻辑正确。✅
- build / vet 全绿;pkg/auth(含 role-resolver 全用例)+ modules/user(含 Invalidate/RoleService 新测试)逻辑测试全过;新增 common/api_test.go + user 测试覆盖修复点。

## 核心安全机制(沿用上轮已核结论,本轮未回归)
per-request 角色解析 fail-open/降权即时丢弃、全局挂载无绕过、4 个高危操作 + addAppVersion 收紧到 superAdmin —— 均未改动,保持正确。

无 blocking、无遗留 should-fix。重投 APPROVE。CI 若仍 unstable 是 3 个非代码闸(check-sprint 缺 Closes#/dependency-review 仓库 graph/secret-scan 全树噪声),非本 PR 代码问题。

—— Octo-PR (lml2468)

Jerry-Xin (dismissed):
Summary: The PR is relevant to octo-server and the auth/management hardening is directionally sound; I found no merge-blocking issues.

💬 Non-blocking

🟡 Warning: `pkg/auth/parser.go:139-156` keeps the token role snapshot when the role resolver errors. That preserves availability, but during DB/resolver failures a demoted or deleted admin can continue using the old system role. This is documented behavior, but for high-impact gates it is worth considering a stricter policy for tokens carrying `admin` or `superAdmin`.

🟡 Warning: `modules/common/api.go:550-568` still returns legacy raw `c.ResponseError(...)` responses for malformed body, validation, and insert failures on the now supply-chain-sensitive app-version write path. The PR improves the role-denial path, but this endpoint remains inconsistent with the repository’s localized error-envelope rules.

🔵 Suggestion: The new route-gate tests seed roles directly into token cache, e.g. `modules/common/api_test.go:42-53` and `modules/space/api_manager_test.go:1022-1045`. Parser unit tests cover resolver override, but an integration-style regression where DB role removal overrides a still-admin token on one of these tightened endpoints would better lock in the “revocation-aware gate” claim.

✅ Highlights

The `RoleResolver` hook is cleanly isolated in `pkg/auth` and keeps the parser decoupled from DB/Redis implementation details.

The destructive space-management gates are scoped and consistently return the generic forbidden envelope.

Verification run: `go test ./pkg/auth` passed, targeted `go test ./modules/user -run 'TestRoleService|TestNewRoleService'` passed. Full `go test ./modules/user` was blocked by a local migration-plan panic unrelated to this diff.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#364 Review Report
**Head SHA:** `7aac4987855dee076c796c82b134d032246c9d2e`
**PR:** feat(auth): per-request role resolution + tighten high-impact management gates

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Per-request role resolution (RoleResolver hook) | ✅ | `pkg/auth/parser.go:136-155` — resolver called after token decode; fail-open on error preserves snapshot; empty result drops stale role |
| RoleService (Redis → DB → "") | ✅ | `modules/user/role_service.go:75-103` — cache miss (`"", nil`) falls through to DB; cache error falls through to DB; negative marker `"-"` prevents DB hammering for normal users |
| superAdmin gate on addAppVersion | ✅ | `modules/common/api.go:540` — `CheckLoginRoleIsSuperAdmin()` replaces `CheckLoginRole()`; generic 403 via `httperr.ResponseErrorL` |
| requireSuperAdmin on 4 destructive space ops | ✅ | `modules/space/api_manager.go:326,418,671,735` — forceDisband, liftBan, removeMembers, updateMemberRole all switched |
| All-device token revocation on admin deletion | ✅ | `modules/user/api_manager.go:349-372` — iterates APP/Web/PC (the complete DeviceFlag set per `octo-lib/config/msg.go:28-37`) |
| Role cache invalidation on admin deletion | ✅ | `modules/user/api_manager.go:343` — `m.roleService.Invalidate(user.UID)` after token revocation |
| QueryRoleByUID (single-column read) | ✅ | `modules/user/db.go:135-140` — `SELECT role FROM user WHERE uid=?`; returns `("", nil)` for non-existent uid |
| Wiring in main.go | ✅ | `main.go:122-128` — `NewRoleService` + `WithRoleResolver` wired alongside existing LanguageResolver |
| Test coverage | ✅ | 4 parser tests (override/empty-drops/fail-open/nil-noop); 9 RoleService tests; regression test for destructive ops; admin deletion integration test |

## 2. Findings

### F1 — P2: `roleService.Invalidate` skipped when `revokeAllDeviceTokens` fails

**Diff-scope:** new (introduced by this PR's restructuring of `deleteAdminUsers`)

**Path:** `modules/user/api_manager.go:337-344`

In `deleteAdminUsers`, the DB row is deleted first (line 327), then `revokeAllDeviceTokens` is called (line 337). If `revokeAllDeviceTokens` returns an error, the handler returns early at line 340 **without reaching `roleService.Invalidate`** at line 343.

Consequence: the `user_role:{uid}` hot cache retains the stale role for up to `RoleCacheTTL` (60s). Combined with a still-valid token (if `revokeAllDeviceTokens` partially failed — deleted some but not all device tokens), the demoted user could still pass admin/superAdmin gates during that window.

**Severity rationale (R1/R2):** This requires a partial Redis failure during token revocation — a rare degraded state. The DB row is already deleted, so most handler-level operations would fail at their own DB queries. The 60s TTL bounds the window. Does not meet R1's "makes a working path unavailable / produces incorrect data / worse than before" bar — the old code had NO invalidation at all. Classified P2 (new edge case introduced by restructuring, but strictly better than the pre-PR behavior).

**Suggestion:** Move `m.roleService.Invalidate(user.UID)` before the `revokeAllDeviceTokens` call, or call it in a `defer` so it runs regardless of token revocation outcome. The invalidation is best-effort anyway (Redis delete error is swallowed), so it's safe to call unconditionally.

### F2 — P2: `revokeAllDeviceTokens` partial-failure observability

**Diff-scope:** new (this PR introduces the function)

**Path:** `modules/user/api_manager.go:349-372`

`revokeAllDeviceTokens` iterates `{APP, Web, PC}` and returns on the first error. If APP token revocation succeeds but Web fails, the caller sees a generic error with no indication of which devices were partially revoked. The APP token is already gone but the Web token survives — and the error message doesn't distinguish this.

**Severity rationale:** Low impact — the user row is deleted, so re-login is impossible. The partial-failure window is narrow. P2 (code quality / observability).

**Suggestion:** Log each device flag's outcome individually (or collect errors) so ops can diagnose partial revocations. Consider continuing through all device flags even on error, then returning an aggregate.

### F3 — P2 (note): Fail-open during total cache+DB outage preserves stale admin role

**Diff-scope:** new (this PR introduces the per-request role resolution)

**Path:** `pkg/auth/parser.go:149-152`

When `ResolveRole` returns an error (both Redis and DB unreachable), the parser keeps the token's baked-in role snapshot. A demoted admin whose token was minted before the demotion retains admin access for as long as the outage lasts.

**Severity rationale:** This is a **deliberate, documented design choice** mirroring the existing LanguageResolver pattern. A total cache+DB outage would also break most handler-level operations (DB queries for business logic). The alternative (fail-closed → 5xx all authenticated requests) would make the entire service unavailable. P2 note — not a defect, but a trade-off worth flagging for the security-sensitive routing.

## 3. Suggestions

1. **Move `Invalidate` before `revokeAllDeviceTokens`** in `deleteAdminUsers` (F1 fix). One-line change, eliminates the stale-cache-on-partial-failure edge case.
2. **Add structured logging** to `revokeAllDeviceTokens` per device flag (F2). Helps ops diagnose partial failures.
3. Consider adding a **role validation** in `writeCache` (e.g., only cache `"admin"`, `"superAdmin"`, or the negative marker) as defense-in-depth against future DB corruption. Low urgency.

## 4. Additional Observations

- **Anti-enumeration is correctly implemented.** Both `requireAdmin` and `requireSuperAdmin` return the same generic `err.shared.auth.forbidden` (HTTP 403 semantic / 400 wire per legacy D14 compat). The raw error strings from `CheckLoginRole`/`CheckLoginRoleIsSuperAdmin` are discarded. No information leakage about which role is required.
- **`addAdminUser` correctly does NOT call `Invalidate`** — it creates a brand-new UID via `util.GenerUUID()` (`api_manager.go:448`), so no prior cache entry can exist. Verified.
- **`createManagerAccount` (startup bootstrap) correctly does NOT call `Invalidate`** — one-time creation of the SuperAdmin account; never mutates an existing uid.
- **No UPDATE path exists for `user.role`** — the only mutations are INSERT (new uid) and DELETE (row removal). The Invalidate contract is therefore simple and fully covered.
- **DeviceFlag set is complete** — `config.DeviceFlag` has exactly 3 values: `APP=0, Web=1, PC=2` (iota block closed after PC in `octo-lib/config/msg.go:28-37`). `revokeAllDeviceTokens` covers all three.
- **Cache miss semantics are correct** — both `RedisCache.Get` and `MemoryCache.Get` return `("", nil)` on miss, which the RoleService correctly distinguishes from the negative marker `"-"`.
- **Two `RoleService` instances** are created (one in `main.go` for the token parser, one in `user.NewManager` for invalidation). They share the same Redis backend, so cross-instance invalidation works correctly. This mirrors the existing `LanguageService` pattern.

## 5. Data-Flow Tracing

|---|---|---|---|
| `info.Role` (token snapshot) | `auth.Decode(token)` → `TokenInfo.Role` | Token cache → `Parse()` → `info.Role` | ✅ Yes, used as fallback when resolver errors |
| `resolved` role from resolver | `RoleService.ResolveRole(uid)` → Redis `user_role:{uid}` → DB `user.role` | `Parse()` → `roleResolver.ResolveRole(ctx, info.UID)` → `resolved` | ✅ Yes, overrides snapshot when `rerr == nil` |
| `role` in returned `UserInfo` | `role` local var (either snapshot or resolved) | `Parse()` return → auth middleware → `c.Set("role", ...)` → `c.GetLoginRole()` | ✅ Yes, flows to `CheckLoginRole`/`CheckLoginRoleIsSuperAdmin` |
| `cached` from Redis | `cache.Get(RoleCacheKeyPrefix + uid)` | `RedisCache.Get` → `r.conn.GetString(key)` → `("", nil)` on miss | ✅ Correctly handled: miss falls through to DB |
| `role` from DB | `QueryRoleByUID(uid)` → `SELECT role FROM user WHERE uid=?` | `dbr.Load(&role)` → `("", nil)` for non-existent uid | ✅ Yes, empty string = normal user, cached as negative marker |
| `ErrSharedForbidden` in gate rejection | `errcode.ErrSharedForbidden` → `i18n/codes/shared.go:63-70` | `httperr.ResponseErrorL` → i18n envelope → HTTP 400 wire / 403 semantic | ✅ Generic 403, no role info leaked |

## 6. R5 Blind-Point Checklist (security_sensitive — mandatory)

**C1 — Dual-path parity:**
- `addAdminUser` ↔ `deleteAdminUsers`: Both already require `CheckLoginRoleIsSuperAdmin()` (pre-existing). Add creates new UID (no Invalidate needed); Delete calls Invalidate. **Clear.**
- `requireAdmin` ↔ `requireSuperAdmin`: Two distinct helpers for distinct risk levels. Read/low-risk → `requireAdmin`; destructive/irreversible → `requireSuperAdmin`. The split is intentional and documented (`addMembers` stays at `requireAdmin` with explicit rationale). **Clear.**
- No create-gate / execute-path asymmetry found. All role-mutation paths verified.

- `RoleResolver` runs inside `CacheTokenParser.Parse()` after token decode, before `UserInfo` construction. Single call per request, no double-resolution. **Clear.**
- `writeCache` does not validate role strings before caching. Tested with non-canonical values: the DB is the only write source and only produces valid values (`"admin"`, `"superAdmin"`, `""`). No injection vector. **Clear.**
- `requireSuperAdmin` / `requireAdmin` are leaf helpers, not composed with each other. No nesting risk. **Clear.**

- The per-request role resolution **tightens** the authorization boundary: roles are now resolved from DB truth on every request, not trusted from the token snapshot. A demoted admin loses access within the cache TTL (60s) instead of at token expiry (days). **Clear — improvement.**
- The fail-open degradation (F3) is the only relaxation, and it's bounded by the outage duration. **Noted in F3.**
- No new endpoints or tools are exposed. The PR only restricts existing ones. **Clear.**

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. Three P2 items (F1: Invalidate ordering edge case, F2: partial-failure observability, F3: fail-open design note) — all are suggestions, none block landing. The PR is well-structured, thoroughly tested, and makes a meaningful security improvement. The per-request role resolution correctly closes the stale-role window, the gate tightenings are consistent and well-documented, and the anti-enumeration design is sound.

yujiawei (changes requested):
# Code Review — PR #364 (octo-server)

**Scope reviewed:** head SHA `7aac4987855dee076c796c82b134d032246c9d2e`, merge-base `fde79cfd` (origin/main). Build passes for all changed packages; `pkg/auth` and `modules/user` role-service unit tests pass locally (DB-backed integration tests require MySQL/Redis and were not run in this environment).

## Summary

This is a well-constructed, security-positive change. Per-request role resolution (`RoleResolver` → `RoleService` backed by a `user_role:{uid}` hot cache → DB) correctly closes the long-standing hole where a system role baked into a token at login outlived a demotion until token expiry. The fail-open-on-resolver-error, negative-marker, and empty-role-drops-the-snapshot semantics are all sound and well covered by unit tests. The privilege tightening on the four irreversible cross-space ops and on `addAppVersion` is correctly scoped.

I'm requesting one change for a real correctness gap on an error path that undermines the very guarantee this PR exists to provide. Everything else below is non-blocking.

## Verified correct

- **Per-request override semantics** (`pkg/auth/parser.go`): resolver error → keep token snapshot (fail-open, documented); resolved `""` → drop the baked-in role; resolved non-empty → authoritative. `AuthMiddleware` does `if info.Role != "" { c.Set("role", info.Role) }`, and the pooled `wkhttp.Context` wraps a fresh `gin.Context` per request, so an empty resolved role correctly yields "no role" with no stale-key leak across requests.
- **Cache semantics**: redis `GetString` returns `("", nil)` on a miss, so the cache-miss / negative-marker / cache-error branches in `ResolveRole` behave as intended; a Redis `Get` error falls through to the authoritative DB read rather than silently stripping/granting privileges.
- **Gate scope** (`modules/space/api_manager.go`): exactly the four irreversible ops (`forceDisband`, `liftBan`, `removeMembers`, `updateMemberRole`) move to `requireSuperAdmin`; reversible `addMembers` deliberately stays at `requireAdmin` (documented). Read/low-risk endpoints are untouched. `addAppVersion` is the only write handler on `/v1/common`, correctly tightened.
- **Cache-invalidation invariant**: the only paths mutating `user.role` are `addAdminUser` (always a brand-new UUID via `INSERT`, cannot collide with a stale cache key) and `deleteAdminUsers` (calls `Invalidate`). Generic user-update paths whitelist columns and exclude `role`.

## P1 — blocking

### `deleteAdminUsers`: role hot-cache is not invalidated on the token-revocation error path

`modules/user/api_manager.go:327-344`

The order is: (1) DB delete, (2) `revokeAllDeviceTokens`, (3) `roleService.Invalidate(user.UID)`. `revokeAllDeviceTokens` is fail-fast (returns on the first Redis error), and on that error the handler does `respondUserError(...); return` at lines 335-338 — so **`Invalidate` at line 343 is never reached**.

Consequences when a Redis partial failure hits the revoke step:
- The positive role cache `user_role:{uid} = "admin"` survives. `ResolveRole` short-circuits on a positive hit (`role_service.go` `default: return cached, nil`) and never re-reads the now-empty DB, for up to `RoleCacheTTL` (60s).
- Any token the revoke loop did not get to delete still loads in `Parse`, and the resolver hands back `"admin"`. Net: a just-deleted admin keeps **admin authorization** for the full TTL.
- This contradicts the code's own comment at lines 340-342, which claims the next request will resolve "no role" from DB even if a token was missed — true only if `Invalidate` actually runs, which it does not on this path.

This is exactly the demotion-staleness window the PR is built to close, left open on the failure path of a security-sensitive change.

**Suggested fix (one line):** invalidate the role cache *before* (or independently of) token revocation. The DB delete is already committed and `Invalidate` is best-effort/cheap, so:

```go
err = m.db.deleteUserWithUIDAndRole(uid, string(wkhttp.Admin))
if err != nil {
    m.Error("删除管理员错误", zap.Error(err))
    respondUserError(c, errcode.ErrUserStoreFailed)
    return
}
// Invalidate first: the DB row is gone, so a surviving token must resolve to
// an empty role on the very next request even if token revocation below fails.
m.roleService.Invalidate(user.UID)
if err := m.revokeAllDeviceTokens(user.UID); err != nil {
    m.Error("清除管理员token数据错误", zap.Error(err), zap.String("uid", user.UID))
    respondUserError(c, errcode.ErrUserTokenCacheFailed)
    return
}
c.ResponseOK()
```

A `defer`-based always-run invalidate would also work. Please also update the lines 340-342 comment to match whichever ordering you choose, and consider a test that forces a mid-revoke cache error and asserts the role key is still cleared.

## P2 / nit — non-blocking

- **(P2) Non-atomic deletion / no compensation.** `deleteAdminUsers` performs DB delete → revoke → invalidate with no transaction and a fail-fast revoke loop; on a mid-loop failure, remaining device flags' tokens are stranded, and re-invoking the endpoint dead-ends at the `QueryByUID`/`ErrUserNotFound` guard (the row is already gone), so revoke/invalidate can never be retried. Bounded in impact (the DB row — the authorization source of truth — is gone, role cache self-heals at 60s), but worth making `revokeAllDeviceTokens` best-effort across all flags rather than aborting on the first error.
- **(P2) Read-through re-populate race in `ResolveRole`.** Between the DB read and `writeCache`, a concurrent `Invalidate` can be overwritten, re-caching a stale positive role for up to one TTL. The `cache.Cache` interface has no `SetNX`/CAS so this can't be fully fenced; it stays within the documented 60s staleness budget. A one-line comment acknowledging the race would help future readers.
- **(P2) Integration-test fidelity.** The new module tests (`modules/user`, `modules/space`, `modules/common`) run through `testutil.NewTestServer`, which does not inject the `CacheTokenParser`+`RoleResolver`, so `AuthMiddleware` falls back to the legacy `uid@name@role` parser. They therefore exercise the route-level guards, not the production per-request DB resolution. The resolver/demotion logic is well covered by the `pkg/auth` and `role_service` unit tests, but an end-to-end "token minted as admin → DB demoted → next request rejected" test (constructed against a parser wired with the resolver) would close the loop.
- **(nit) Auth hot path** now does one extra Redis `GET` per request (role) in addition to the language `GET`; documented and intentional (keeps normal users off MySQL via the negative marker). Could be folded into an `MGET`/pipeline later — not for this PR.
- **(nit) Denial-test coverage.** `TestManager_DestructiveOpsRequireSuperAdmin` only drives the `admin → 403` boundary; `role==""`/no-token rejection is covered generically elsewhere and hits the same `CheckLoginRoleIsSuperAdmin` guard, so this is cosmetic.

## For human verification (security-sensitive)

- Confirm the intended invalidation ordering in `deleteAdminUsers` (the P1 above) — this is the one item that changes runtime auth behavior under failure.
- Confirm `RoleCacheTTL = 60s` is an acceptable upper bound on demotion staleness for your threat model, given fail-open-on-resolver-error keeps the token's role on a cache/DB outage.

lml2468 (changes requested):
🔴 撤回我刚才的 APPROVE,改 CHANGES_REQUESTED —— yujiawei 发现一条真实的异常路径提权窗口,mochashanyao 也独立指向同一处时序,我本地核实成立。这正是本 PR 要消除的降权失效窗口,却在异常分支漏掉了。

## 缺陷(deleteAdminUsers,api_manager.go:334-348)
当前顺序:
1. deleteUserWithUIDAndRole(删 DB role)— 已成功
2. revokeAllDeviceTokens — **失败则 return,跳过第 3 步**
3. roleService.Invalidate(uid)

问题:revokeAllDeviceTokens 失败(cache.Delete 报错)时直接 return,**Invalidate 永不执行**。此时:
- DB role 已删(权威源已变);
- 但 RoleService 角色热缓存仍存旧 "admin"(60s TTL);
- 且 token 撤销失败 = 那个 token 可能仍存活。
→ 残留 token + 旧 role 缓存叠加,被降权的管理员在异常窗口内仍按 admin 放行,最长 RoleCacheTTL(60s)。这恰是本 PR 立意要堵的提权窗口,异常分支不该再留。

## 修法(yujiawei + mochashanyao 共识方向)
把 Invalidate 移到 token 撤销**之前**(或确保 revoke 失败也仍调 Invalidate,如 defer)。因为 DB role 已在最前删除,只要 role 缓存也清了,即便 token 残留,下一请求经 RoleResolver 也会从 DB 读出空角色、正确拒绝。建议:
```
deleteUserWithUIDAndRole(...)        // 1. 权威源
m.roleService.Invalidate(user.UID)   // 2. 先清角色缓存(挪到前面)
if err := m.revokeAllDeviceTokens(...); err != nil { ... }  // 3. 再尽力撤 token
```
这样三步任一失败,降权语义都不被旧缓存压制。

## 严重性说明
属异常路径(cache 故障才触发)+ 需残留 token 被使用,非常规流程。但本 PR 是安全修复、降权正确性是其核心契约,异常分支也不应留提权窗口 —— 故与 yujiawei 一致按 blocking 处理。修复是几行挪动 + 一个测试,成本很低。

(其余:mochashanyao 第二条 token 撤销部分失败缺详细日志 = nit 可观测性,顺手收。)

前几轮的 follow-up(Invalidate 接入、多设备撤销、i18n)都修对了,只剩这条异常分支顺序。修完我立即重投 APPROVE。

—— Octo-PR (lml2468)

Jerry-Xin (dismissed):
Re-review at 7aac498. Withdrawing my earlier APPROVE to align with yujiawei and lml2468 on the role-cache invalidation ordering. After tracing `deleteAdminUsers` (`modules/user/api_manager.go:298-349`), I agree this is a fix-before-merge on a security PR whose whole point is timely privilege revocation.

🔴 Blocking — Invalidate is skipped on the partial-failure path that needs it most.

Current order in `deleteAdminUsers`:
1. `deleteUserWithUIDAndRole(uid)` — DB role removed (authoritative) ✅
2. `revokeAllDeviceTokens(user.UID)` — on ANY cache Get/Delete error → `return` early (`api_manager.go:335-338`)
3. `m.roleService.Invalidate(user.UID)` — only reached if step 2 fully succeeded (`:343`)

The problem: when `revokeAllDeviceTokens` fails (e.g. transient cache error after deleting some but not all device tokens), the handler returns BEFORE `Invalidate`. That leaves BOTH a possibly-live token AND a stale role hot-cache (≤ RoleCacheTTL, ~60s). That is exactly the window where a just-deleted admin can keep acting as admin — the safety net is bypassed precisely in the failure case it exists for. The inline comment at `:340-342` even claims "even if a token was missed, the next request resolves empty role from DB" — but that only holds if Invalidate runs, which it doesn't on this path.

Fix (trivial, correct): make Invalidate run regardless of token-revoke outcome, right after the DB delete succeeds — e.g. `defer m.roleService.Invalidate(user.UID)` immediately after step 1, or call Invalidate before returning on the revoke-error branch. Invalidate must not be gated behind revoke success.

💬 Non-blocking
- 🟡 mochashanyao's ordering note (invalidate-before-revoke) is subsumed by the above; the `defer` form addresses both.
- 🟡 yujiawei's observability note: when `revokeAllDeviceTokens` partially fails, log which device flag failed for auditability.

✅ Highlights (unchanged from prior round, still verified)
- Per-request RoleResolver wired globally; fail-closed on authz (empty resolved role drops privileges), fail-open only to token snapshot on infra error.
- Multi-device token revocation (APP/Web/PC) — a real privilege window the author self-found and fixed.
- addAppVersion → unified ErrSharedForbidden 403 (anti-enumeration); addMembers/removeMembers asymmetry intentional and justified.
- addAdminUser uses a fresh GenerUUID uid, so "no Invalidate needed" holds (no prior cache entry).

This is a clean, high-quality PR — just move Invalidate off the token-revoke success path and it's good to merge.

lml2468 (approved):
✅ APPROVE(重审 @444bc4a)— 异常分支提权窗口已修对,且作者把相关健壮性一并收齐。本地逐条核实 + build/vet 绿 + PR 新增 best-effort 单测通过。

## 上轮 blocking(异常分支提权窗口)已修
deleteAdminUsers 现在的顺序(api_manager.go:328-347):
1. deleteUserWithUIDAndRole(删 DB role,权威源)
2. **m.roleService.Invalidate(user.UID) ← 挪到 token 撤销之前(:339)**
3. revokeAllDeviceTokens(:342)— 失败 return,但此时 role 缓存已清

→ 即便 token 撤销失败、某 token 残留,DB role 已删 + role 热缓存已清,下一请求经 RoleResolver 从 DB 解析出空角色被拒。**提权窗口消除**。注释也准确解释了为何此序正确。这正是 yujiawei/Steve/我三人共识的修法。

## 作者额外收齐的健壮性(超出要求)
- ✅ **best-effort 全设备撤销** — 重构 revokeDeviceTokensInCache:不再首个错误就 return,遍历 APP/Web/PC 逐个尽力删,一个设备的 Redis 抖动不再连累其他设备(修了 yujiawei 第二条隐含点)。
- ✅ **逐设备失败日志**(device_flag)— 收了"部分失败缺审计日志"的可观测性建议。
- ✅ **可测试性** — 抽出纯 cache 核心 + 注入 fake cache,新增 TestRevokeDeviceTokensInCache_BestEffort 覆盖部分失败路径。

## 验证
- 本地核实最终顺序正确(DB删:328 → Invalidate:339 → revoke:342)。
- build / vet 绿;PR 新增 best-effort 单测本地 PASS(纯 cache、不依赖 DB)。
- 注:本地 modules/user 全量测试因 testutil.NewTestServer 需 DB(环境缺)panic,非本 PR 代码问题;DB-backed 测试由 CI 跑。CI Test 本轮推送后 pending,待绿。

## 核心安全机制(沿用前轮已核,本轮未回归)
per-request resolver 全局接 + fail-open/降权即时丢弃、4 高危操作 + addAppVersion superAdmin、addAdminUser 全新 uid 论断成立 —— 均未动。

无 blocking、无遗留 should-fix。重投 APPROVE。(若 CI Test 最终非绿,以 CI 为准、需复核;但本地逻辑测试已通过。)

—— Octo-PR (lml2468)

Jerry-Xin (approved):
The PR is in scope for octo-server and the core authorization changes are consistent with the project’s auth model.

💬 Non-blocking

🟡 Warning — The new endpoint tests mostly prove token-snapshot role gates, not the production resolver path. `modules/common/api_test.go:41-58` and `modules/space/api_manager_test.go:25-41` seed role-bearing tokens, but `testutil.NewTestServer()` does not wire the new `RoleResolver`, so these tests would not catch a future integration break where DB role resolution strips or overrides the role unexpectedly. The parser unit tests cover the hook well, so this is a coverage gap rather than a blocker.

🟡 Warning — Role resolver failures fail open to the token snapshot in `pkg/auth/parser.go:153-156`, but there is no logging or metric around that security-relevant fallback. The behavior is documented and intentional, but some observability would make Redis/DB outages that temporarily preserve revoked roles detectable.

🔵 Suggestion — `modules/common/api.go:550`, `modules/common/api.go:555`, and `modules/common/api.go:568` still use legacy raw error responses in the touched `addAppVersion` handler. This PR improves the role-denial path, but a follow-up should migrate the remaining user-facing errors to the localized envelope to match repository guidance.

✅ Highlights

The `RoleService` negative-cache design in `modules/user/role_service.go:80-98` correctly avoids DB hits for normal users while still allowing an authoritative empty role to drop stale admin snapshots.

The destructive space-management gates are centralized cleanly with `requireSuperAdmin` in `modules/space/api_manager.go:174-180`, and the selected handlers match the PR’s stated high-impact operations.

`deleteAdminUsers` now invalidates the role cache and revokes all known device tokens in `modules/user/api_manager.go:334-347`, which is a meaningful improvement over Web-only revocation.

Validation: `go test ./pkg/auth` passed. Broader package tests were blocked locally by the existing test database migration-plan error, not by a compile failure in this PR.

Jerry-Xin (dismissed):
Re-review at 444bc4a. The blocking exception-path privilege window I (and yujiawei, lml2468) flagged is fixed — correctly, with bonus hardening. Flipping to APPROVE.

✅ Blocker resolved (verified against `git diff 7aac498..444bc4a`, `modules/user/api_manager.go`)

The fix moves `m.roleService.Invalidate(user.UID)` to run BEFORE `revokeAllDeviceTokens`, right after the DB role delete:
1. `deleteUserWithUIDAndRole(uid)` — DB role removed (authoritative)
2. `m.roleService.Invalidate(user.UID)` — role hot-cache cleared **unconditionally**
3. `revokeAllDeviceTokens(user.UID)` — on error, returns; but the cache is already cleared

So even when token revocation fails and a token survives, the next request resolves an empty role from DB (uid already deleted) and is denied. The ≤60s stale-role + live-token privilege window is eliminated. The inline comment documents the ordering rationale.

Bonus hardening (beyond the requested fix):
- `revokeDeviceTokensInCache` refactored to **best-effort per-device** — no longer aborts on the first cache error; it attempts all APP/Web/PC flags and returns one failure entry per error, so a Redis hiccup on one device can't strand revocation of the others. Maximizes revocation coverage while still reporting failure to the caller.
- Factored into an injectable cache-only core (`revokeDeviceTokensInCache(c cache.Cache, ...)`), now unit-testable with a fake cache; a new test (`api_manager_test.go +26`) covers it.

CI: code-quality gates green (Build/Lint/Vet/i18n/osv-scanner all pass; Test pending, not failed). Known non-code gates (check-sprint / dependency-review / secret-scan) are out of scope.

✅ Highlights (carried over, still intact)
- Per-request RoleResolver wired globally; fail-closed on authz (empty resolved role drops privileges), fail-open to token snapshot only on infra error (deliberate availability tradeoff).
- Multi-device token revocation (APP/Web/PC) — author-found privilege window, now both ordered correctly and hardened.
- High-impact management ops gated to superAdmin; addAppVersion unified to ErrSharedForbidden 403 (anti-enumeration); addMembers/removeMembers asymmetry intentional.

Clean, well-engineered security PR. The exception-path window is closed exactly as recommended, with added robustness and test coverage. LGTM.

Jerry-Xin (approved):
Re-affirming APPROVE at . The exception-path Invalidate-skip blocker is fixed: roleService.Invalidate now runs unconditionally right after the DB role delete, before revokeAllDeviceTokens, so a token-revoke failure can no longer leave a stale role cache. Token revocation also hardened to best-effort across all device flags with a new regression test. Code-quality CI green. (Earlier I had a duplicate review pair from parallel tooling; this is my single current verdict.)

yujiawei (approved):
# Code Review — PR #364 (octo-server)

Re-reviewed at head `444bc4ac298d56bbcce509c4cf7c2197cd073088` against merge-base `fde79cfd` (origin/main). Scope: per-request system-role resolution (`pkg/auth`, `modules/user`) plus tightening the four irreversible cross-space management ops and `addAppVersion` from `admin` to `superAdmin`.

**Verdict: APPROVE.** The one blocking issue from the prior round is correctly fixed at this head, and a full re-trace across auth resolution, gate scoping, wiring/DB, and tests surfaced no new P0/P1. Build is clean (`go build ./...`), `go vet` clean, and `pkg/auth` + `modules/user` unit tests pass locally. DB-backed integration tests require MySQL/Redis and were not run in this environment (they only fail to *start* on a missing migration table, not to compile).

## Prior blocking issue — resolved

**`deleteAdminUsers` role-cache invalidation ordering** (`modules/user/api_manager.go:328-348`). The latest commit reorders the handler to: DB delete → `roleService.Invalidate(user.UID)` → `revokeAllDeviceTokens`. `Invalidate` is now unconditionally reached after a successful DB delete, before the (now best-effort) token revocation, so a Redis error during revocation can no longer leave a stale `user_role:{uid}="admin"` entry alive for up to `RoleCacheTTL`. Because the authoritative DB role is already empty before `Invalidate` runs, even a surviving un-revoked token resolves to "no role" on the very next request. The fix is complete and the comment was updated to match.

As a bonus, `revokeAllDeviceTokens` is now genuinely best-effort across all three device flags (APP/Web/PC) instead of aborting on the first error — closing a latent gap where a Redis hiccup on one device could strand the others — and the core loop was extracted to `revokeDeviceTokensInCache` with a dedicated unit test asserting all three flags are attempted despite errors.

## Verified correct

- **Per-request override semantics** (`pkg/auth/parser.go:139-160`): resolver error → keep token snapshot (fail-open, documented); resolved `""` → drop the baked-in role; resolved non-empty → authoritative. Each branch is unit-tested.
- **Cache semantics** (`modules/user/role_service.go`): cache miss (`""`) falls through to DB; negative marker (`-`) short-circuits to "no role" without hitting MySQL; a genuine cache *error* bypasses the switch and falls through to the authoritative DB read rather than silently stripping/granting privileges. `QueryRoleByUID` on a missing uid returns `("", nil)` (dbr `Load` on zero rows returns `(0, nil)`), correctly meaning "no system role".
- **Gate scope** (`modules/space/api_manager.go`, `modules/common/api.go`): exactly the four irreversible ops (`forceDisband`, `liftBan`, `removeMembers`, `updateMemberRole`) move to `requireSuperAdmin`; reversible `addMembers` deliberately stays at `requireAdmin` (documented); `addAppVersion` (which sets the client `DownloadURL` — supply-chain sensitive) is correctly tightened. All other read/low-risk endpoints are untouched.
- **Anti-enumeration**: both `requireSuperAdmin` and `requireAdmin` return the same generic `ErrSharedForbidden` 403, so an `admin` hitting a `superAdmin`-gated endpoint gets a byte-identical response to a normal user — no "needs higher role" leak. `CheckLoginRoleIsSuperAdmin` fails closed for `role==""` and for `role=="admin"`; no-token requests never reach these handlers (upstream `AuthMiddleware`).
- **Cache-invalidation invariant**: the only sites that mutate `user.role` are `addAdminUser` (always a brand-new random UUID via `INSERT`, no pre-existing cache key to collide with) and `deleteAdminUsers` (calls `Invalidate`). The generic update path whitelists columns and excludes `role`, so no role write escapes invalidation.
- **Wiring** (`main.go:121-130`): `RoleService` injected via `WithRoleResolver` (nil-safe no-op; `NewRoleService` panics at boot on nil deps rather than degrading at request time). No import cycle — `RoleResolver` is defined consumer-side in `pkg/auth` and `RoleService` satisfies it structurally.

## Non-blocking (P2 / nit)

- **(P2) Read-through repopulate race in `ResolveRole`** (`role_service.go:80-98`): a concurrent request that read `role="admin"` from DB *before* the delete committed can re-`writeCache` the stale positive role *after* `Invalidate` ran, resurrecting it for up to `RoleCacheTTL`. This is inherent to read-through + delete-invalidation, predates this PR, is bounded to 60s, and is masked in practice because token validation (`parser.go:104-110`) locks a revoked user out *before* `ResolveRole` is consulted. A future double-delete / short negative-pin would close it; not required to merge.
- **(P2) Fail-open during a simultaneous cache+DB outage**: when the resolver errors, the token's snapshot role is honored, so a recently-demoted admin with a still-valid token regains privilege until token expiry during a full Redis+MySQL outage. This is the intended, documented tradeoff (avoid auth 5xx; same philosophy as the language resolver) and is the right default — flagging only so the residual operational risk is acknowledged.
- **(P2) `ResolveRole` does not thread `ctx` into the DB query** (`role_service.go` → `db.go QueryRoleByUID`): entry checks `ctx.Err()`, but the subsequent single-column `SELECT` is not ctx-aware, so a cancellation mid-query is not honored. Consistent with the existing `LanguageService` pattern; negligible impact.
- **(P2) Test fidelity**: `testutil.NewTestServer` never wires the `RoleResolver` into the token parser (that lives only in `main.go runAPI`), so route-level tests authenticate through the legacy parser. The resolver and the role service are each well unit-tested in isolation, but the full chain (`token minted admin → DB demoted → next request rejected`) is verified by composition, not end-to-end. Relatedly, the exact prior-round P1 (revoke-error path still invalidates) has no handler-level regression guard — the reorder is correct by inspection, but a revert would pass CI. A handler test that injects a revoke failure and asserts the role key is still cleared would directly pin this fix.
- **(nit)** `addAppVersion` (`common/api.go:550-556`): the role gate was migrated to the i18n envelope, but the `BindJSON`/`check` error paths just below still use raw `c.ResponseError`. Pre-existing legacy, out of scope here — noting only because the handler is now half-migrated.

## For human verification (security-sensitive)

- Confirm `RoleCacheTTL = 60s` is an acceptable upper bound on demotion staleness for your threat model, given fail-open keeps the token's role on a cache/DB outage.
- The two P2 residual windows above (repopulate race; full-outage fail-open) are the only paths where a revoked role can briefly survive — both are bounded and intentional, but worth an explicit ops note.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#364 Review Report
**Head SHA:** `444bc4ac298d56bbcce509c4cf7c2197cd073088`
**PR:** Mininglamp-OSS/octo-server#364

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Per-request role resolution (RoleResolver hook) | ✅ | `pkg/auth/parser.go:136-163` — resolver overrides token snapshot; fail-open on error preserves snapshot |
| RoleService cache → DB → "" lookup chain | ✅ | `modules/user/role_service.go:79-103` — cache miss falls through to DB; negative marker skips DB; cache error falls through to DB (not treated as "no role") |
| superAdmin gate on `addAppVersion` | ✅ | `modules/common/api.go:540-543` — `CheckLoginRoleIsSuperAdmin()` + generic 403 via `httperr.ResponseErrorL(ErrSharedForbidden)` |
| superAdmin gate on 4 destructive space ops | ✅ | `modules/space/api_manager.go:326,418,671,735` — `requireSuperAdmin` on forceDisband/liftBan/removeMembers/updateMemberRole |
| deleteAdminUsers: Invalidate-before-revoke ordering | ✅ | `modules/user/api_manager.go:340-348` — role cache invalidated immediately after DB delete, before token revocation |
| revokeAllDeviceTokens best-effort across APP/Web/PC | ✅ | `modules/user/api_manager.go:380-395` — attempts all 3 device flags, logs per-device, returns first error |
| Anti-enumeration (generic 403, no role-leak) | ✅ | Both `common/api.go` and `space/api_manager.go` use `ErrSharedForbidden` ("err.shared.auth.forbidden") — same generic code for both gates |
| AuthMiddleware integration (role="" → access denied) | ✅ | octo-lib `wkhttp/http.go:333-335` — `c.Set("role", info.Role)` only when `info.Role != ""`; demoted user gets no role key → `CheckLoginRole()` fails |
| Cache invalidation contract completeness | ✅ | All 3 role-mutation paths audited: `addAdminUser` (new UID, no stale cache), `createManagerAccount` (new UID, bootstrap only), `deleteAdminUsers` (calls Invalidate). `allowedUpdateFields` whitelist blocks generic role mutation via `UpdateUsersWithField`. |
| Test coverage | ✅ | 16 new tests covering parser hook (4), RoleService (8), space destructive ops regression (1), appversion gate (1), delete admin integration (1), best-effort revoke (1) |

## 2. Findings

### F1 — P2: `main.go` creates two redundant `user.NewDB(ctx)` instances

**Severity:** P2 (cosmetic / maintainability)

`main.go:120-121` creates `userLangSvc` and `userRoleSvc` with separate `user.NewDB(ctx)` calls:
```go
userLangSvc := user.NewLanguageService(user.NewDB(ctx), ctx.Cache())
userRoleSvc := user.NewRoleService(user.NewDB(ctx), ctx.Cache())
```

Both are stateless wrappers around the same `ctx.DB()` session. A single `userDB := user.NewDB(ctx)` shared between the two would be cleaner. Not a bug — `*DB` holds no mutable state — but unnecessary allocation.

### F2 — P2: `deleteAdminUsers` returns error to client when token revocation partially fails despite successful deletion

**Diff-scope:** amplified (pre-existing code returned error on Web-only revoke failure; PR expands to all device flags, increasing the probability of a partial failure surfacing)
**Severity:** P2 (UX / operational clarity, not a security gap)

In `api_manager.go:348-351`, if `revokeAllDeviceTokens` returns non-nil (any device flag failed), the handler responds with `ErrUserTokenCacheFailed` despite:
- The admin row being deleted from DB ✅
- The role cache being invalidated ✅
- Some (but not all) device tokens being revoked ✅

The client sees an error for an operation that is security-complete. On retry, `QueryByUID` returns nil → `ErrUserNotFound`. No security window exists (role cache is already invalidated), but the error response may confuse operators.

**Suggestion:** Consider responding OK with a warning log when the role cache invalidation succeeded and only token revocation had partial failures. The security posture is already correct.

### F3 — P2: `updateUser` (db.go:295) uses `SetMap` without a field whitelist

**Diff-scope:** pre-existing (not introduced by this PR)
**Severity:** P2 (latent risk, not a current bug)

`db.go:295-298` (`updateUser`) accepts an arbitrary `map[string]interface{}` and passes it to `SetMap`. Unlike `UpdateUsersWithField` which has an `allowedUpdateFields` whitelist, `updateUser` has no structural guard. Today's three callers only set non-role fields (`password`, `web3_public_key`, `name`), so this is not a current vulnerability. However, a future caller passing `"role"` in the map would silently bypass the cache-invalidation contract documented in `role_service.go:28-33`.

**Suggestion:** Add a comment on `updateUser` documenting that callers must not include `"role"` in the map, or add a whitelist similar to `UpdateUsersWithField`.

## 3. Data-Flow Trace (per iron rule)

### 3.1 Role resolution hot path

```
HTTP request (header: "token")
  → AuthMiddleware (octo-lib wkhttp/http.go:323)
    → CacheTokenParser.Parse() (pkg/auth/parser.go:100)
      → cache.Get(prefix + token) → raw string
      → auth.Decode(raw) → TokenInfo{UID, Name, Role, Language}
      → RoleResolver.ResolveRole(ctx, UID) (modules/user/role_service.go:79)
        → cache.Get("user_role:" + uid)
          → hit: return cached role (or "" for negative marker)
          → miss/error: db.QueryRoleByUID(uid) (modules/user/db.go:137)
            → SELECT role FROM user WHERE uid=?
            → write-back to cache with 60s TTL
        → rerr == nil: role = resolved (overrides token snapshot)
        → rerr != nil: role = info.Role (fail-open to token snapshot)
    → c.Set("role", info.Role) only if info.Role != "" (wkhttp/http.go:333-335)
  → Handler calls CheckLoginRole() / CheckLoginRoleIsSuperAdmin()
    → c.GetLoginRole() reads gin key "role"
    → compares against "admin" / "superAdmin"
```

**Verified:** When resolver returns "" (demoted user), AuthMiddleware does NOT set the "role" gin key → `CheckLoginRole()` returns error for empty role → access denied. ✅

### 3.2 Admin deletion + revocation path

```
deleteAdminUsers (api_manager.go:298)
  → CheckLoginRoleIsSuperAdmin() — caller must be superAdmin
  → QueryByUID(uid) — verify target exists and is admin (not superAdmin)
  → db.deleteUserWithUIDAndRole(uid, "admin") — DELETE FROM user WHERE uid=? AND role=?
  → roleService.Invalidate(uid) — DELETE user_role:{uid} from Redis ← BEFORE token revoke
  → revokeAllDeviceTokens(uid)
    → for each flag in [APP, Web, PC]:
      → cache.Get(UIDToken:{flag}{uid}) → token string
      → cache.Delete(Token:{token}) — revoke token payload
      → cache.Delete(UIDToken:{flag}{uid}) — revoke reverse mapping
    → best-effort: continues on per-device errors, returns first error
```

**Verified:** Invalidate-before-revoke ordering ensures that even if token revocation fails, the role cache is cleared. A surviving token with stale admin role will be denied by the per-request RoleResolver (DB row is gone → returns ""). ✅

### 3.3 Cache error degradation

```
RoleService.ResolveRole:
  cache.Get("user_role:{uid}") returns error (Redis down)
    → falls through to db.QueryRoleByUID(uid) — authoritative DB read
    → if DB also fails → returns error → parser keeps token snapshot (fail-open)
```

**Verified:** Cache error is NOT treated as "no role" (which would strip privileges). It falls through to DB. Only if both cache and DB fail does the parser keep the token snapshot. ✅

## 4. R5 Blind-Spot Checklist (security_sensitive)

### C1 — Dual-path parity
**Clear.** No symmetric add↔remove paths are touched for system roles. The only role-mutation paths are: `addAdminUser` (INSERT new UID), `createManagerAccount` (INSERT bootstrap), `deleteAdminUsers` (DELETE + Invalidate). Token revocation now covers all 3 device flags symmetrically (APP/Web/PC). No drift between parallel implementations.

**Clear.** The ordering in `deleteAdminUsers` (DB delete → Invalidate → revoke) is explicitly documented and tested. The parser runs language resolution then role resolution sequentially — no nesting or double-application risk. The `roleNegativeMarker` ("-") is not a valid role string, so no confusion with real data. `revokeDeviceTokensInCache` is a pure function with no side effects beyond cache operations.

**Clear.** The RoleResolver is read-only (never writes roles). The `user_role:{uid}` cache key uses a hardcoded prefix (`RoleCacheKeyPrefix = "user_role:"`) — no user-controlled input in key construction. The new superAdmin gates correctly restrict the 4 destructive space ops and the appversion supply-chain write. No new endpoints are exposed; existing endpoints only have tighter gates.

## 5. Additional Observations

- **Fail-open philosophy consistency:** The role resolver's fail-open behavior (keep token snapshot on resolver error) mirrors the existing language resolver's approach. This is a deliberate, documented trade-off: during a cache+DB outage, a demoted admin retains access for up to RoleCacheTTL (60s) after the outage resolves. Acceptable given the alternative (all admin operations blocked during any infrastructure hiccup).
- **`addMembers` stays at `requireAdmin`:** The reasoning (reversible operation — members can be removed) is sound and explicitly documented in the code. Not a concern.
- **Test quality:** Tests use hand-crafted token cache entries that bypass the real login flow. This is appropriate for testing role gates in isolation. The integration test (`TestManager_DeleteAdminUser_RevokesSessionsAndRoleCache`) exercises the full route → handler → cache assertion path.

---

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Three P2 observations (redundant DB allocation, error response on partial token revocation, unguarded SetMap path) are maintainability/UX improvements, not security blockers. The PR correctly implements per-request role resolution, properly orders cache invalidation before token revocation, covers all device flags for revocation, and applies superAdmin gates to the right destructive endpoints. R5 checklist is clear across all three categories.

PR #11, opened 2026-05-13
ci: add Octo issue/PR feed workflows

lml2468:
Add real-time GitHub event notification to Octo IM channels via GitHub Actions.

- `octo-issue-feed.yml`: pushes Issue events (opened/closed/reopened/labeled) to Octo 🛠️ issue-feed and the repo's project channel
- `octo-pr-feed.yml`: pushes PR events (opened/closed/merged/etc.) to Octo 🛠️ pr-feed and the repo's project channel

PR #76, opened 2026-05-18
feat(common): allow DB overrides for YAML settings

an9xyz:
## Summary
- add a `system_setting` KV table for admin-tunable global config
- expose manager APIs to list, update, and test runtime system settings
- route registration toggles and SMTP settings through DB-backed overrides with YAML fallback

## DB override settings

| Category | Key | Type | YAML fallback | Description |
|---|---|---|---|---|
| `register` | `off` | `bool` | `Register.Off` | disable registration globally |
| `register` | `only_china` | `bool` | `Register.OnlyChina` | restrict registration to China phone numbers |
| `register` | `username_on` | `bool` | `Register.UsernameOn` | enable username registration |
| `register` | `email_on` | `bool` | `Register.EmailOn` | enable email registration/login |
| `support` | `email` | `string` | `Support.Email` | SMTP sender/support email |
| `support` | `email_smtp` | `string` | `Support.EmailSmtp` | SMTP server address (`host:port`) |
| `support` | `email_pwd` | `encrypted` | `Support.EmailPwd` | SMTP password, encrypted at rest |

Empty DB values fall back to YAML defaults. Encrypted values are masked in GET responses; empty update payloads and the `****` mask sentinel preserve the existing ciphertext.

Runtime behavior:
- Settings are read from an in-memory atomic snapshot on every call site.
- A successful `POST /v1/manager/common/system_setting` persists the batch and calls `SystemSettings.Reload()`, so the current server instance sees the new values immediately after the request returns.
- Other server instances pick up the same DB values via background auto-reload within `defaultReloadTTL` (60 seconds).
- If the initial DB load or a later auto-reload fails, getters fall back to YAML defaults until a later reload succeeds.

## Manager APIs
- `GET /v1/manager/common/system_setting` lists schema and current values
- `POST /v1/manager/common/system_setting` updates settings in one transaction
- `POST /v1/manager/common/system_setting/test_email` sends a test email using the effective SMTP config

### API examples

`GET /v1/manager/common/system_setting`

```json
{
  "items": [
    {
      "category": "register",
      "key": "email_on",
      "value": "1",
      "value_type": "bool",
      "description": "是否开启邮箱注册/登录"
    },
    {
      "category": "support",
      "key": "email_pwd",
      "value": "****",
      "value_type": "encrypted",
      "description": "SMTP 密码（加密存储）"
    }
  ],
  "schema": [
    {
      "category": "register",
      "key": "email_on",
      "type": "bool",
[...]

PR #365, opened 2026-06-14
feat(searchetl): scaffold message-search ETL module + cursor table (YUJ-4530 phase 1)

yujiawei:
## What

Phase 1 (part 2 of 2) of the octo-im message → ETL → Kafka → ES-indexer search pipeline. This PR scaffolds the **`searchetl` module + its isolated cursor table** in octo-server, cloning the proven `opanalytics` incremental-cursor ETL paradigm.

### Files
- **Migration** `modules/searchetl/sql/20260614000001_searchetl_init.sql` — new table `octo_etl_es_cursor` (per-shard `last_id` watermark), physically isolated from opanalytics' `octo_etl_message_cursor`. The two ETLs run independent cursors.
- **`modules/searchetl/`** — module skeleton:
  - `etl_db.go` — shard enumeration, cursor ensure/load, `max(id)` (for backlog metric), keyset batch read. Source SELECT already pulls `message_id` + `payload` + `created_unix` so phase 2 reuses the exact row shape.
  - `stability.go` (+ tests) — the **stability gate** `created_unix <= DB_NOW - lag` with no-hole prefix cut. This is the hard anti-data-loss condition **C1** and is present from day one: the cursor may only advance to the stable-prefix tail, never the batch tail, so a low-id/late-commit row is never skipped.
  - `etl.go` — `RunIncrementalDryRun`: walks every shard, reads the stable prefix, logs stable-row + backlog counts, **without advancing the cursor and without producing to Kafka**.
  - `etl_lock.go` — Redis run-lock primitive (key `searchetl:etl:run`, SET NX EX + Lua CAS-DEL/CAS-PEXPIRE), staged for the single-replica mutual exclusion (**C3**) that lands with the transaction split in later phases.
  - `config.go` — `batch` / `lag` / `tick` env knobs (clamped), lag default 600s (C1 STOP: prod lag must never be 0).
- `internal/modules.go` — blank-import registration.

### Why no behaviour change yet
No scheduler is started, no Kafka, no cursor advancement. The module **compiles against the current octo-lib pin** — the contract import + `go.mod` pin bump happen in **phase 2 after the octo-lib PR merges** (publish order / plan STOP #11: never pin to an unmerged commit).

## Verification
- `go build ./...` ✅ · `go vet ./modules/searchetl ./internal` ✅
- `go test ./modules/searchetl` ✅ — stability-gate unit tests incl. the C1 low-id/late-commit no-hole reproduction
- Migration is additive (CREATE TABLE + DROP on down); dry-run safe.
- /codex review: PASS ✅ (no blocking issues in SQL construction, stable-prefix gating, Redis lock, cursor data-loss behaviour)

## Merge order
⚠️ Independent-compiling, but logically **after** octo-lib#76. Phase 2 (Kafka wiring) will bump the pin to octo-lib#76's merge SHA and import `contract/searchmsg`.
Depends on: Mininglamp-OSS/octo-lib#76

Part of YUJ-4530 (phase 1).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	firebase.google.com/go/v4 v4.13.0
 	github.com/BurntSushi/toml v1.6.0
 	github.com/DATA-DOG/go-sqlmock v1.5.0
-	github.com/Mininglamp-OSS/octo-lib v0.0.0-20260605071834-330ed3ce7e6a
+	github.com/Mininglamp-OSS/octo-lib v0.0.0-20260614033907-f62e626a34db
 	github.com/alibabacloud-go/darabonba-openapi v0.2.1
 	github.com/alibabacloud-go/sms-intl-20180501 v1.0.1
 	github.com/alibabacloud-go/tea v1.2.1
--- a/go.sum
+++ b/go.sum
@@ -65,8 +65,8 @@ github.com/HdrHistogram/hdrhistogram-go v1.1.2 h1:5IcZpTvzydCQeHzK4Ef/D5rrSqwxob
 github.com/HdrHistogram/hdrhistogram-go v1.1.2/go.mod h1:yDgFjdqOqDEKOvasDdhWNXYg9BVp4O+o5f6V/ehm6Oo=
 github.com/MicahParks/keyfunc v1.9.0 h1:lhKd5xrFHLNOWrDc4Tyb/Q1AJ4LCzQ48GVJyVIID3+o=
 github.com/MicahParks/keyfunc v1.9.0/go.mod h1:IdnCilugA0O/99dW+/MkvlyrsX8+L8+x95xuVNtM5jw=
-github.com/Mininglamp-OSS/octo-lib v0.0.0-20260605071834-330ed3ce7e6a h1:nC17gAXfODuv8zu0NLj7KfceLGV0dCq94prfhgknZ4g=
-github.com/Mininglamp-OSS/octo-lib v0.0.0-20260605071834-330ed3ce7e6a/go.mod h1:uOCTTI3QdG0ETKZ6uluRWUZgKexee/pY1pQYWyIEvZ0=
+github.com/Mininglamp-OSS/octo-lib v0.0.0-20260614033907-f62e626a34db h1:FudjIbMVhMga/sUsysfzKBI7LTdz0Rgw9EWU2tEJzAo=
+github.com/Mininglamp-OSS/octo-lib v0.0.0-20260614033907-f62e626a34db/go.mod h1:uOCTTI3QdG0ETKZ6uluRWUZgKexee/pY1pQYWyIEvZ0=
 github.com/RichardKnop/logging v0.0.0-20190827224416-1a693bdd4fae h1:DcFpTQBYQ9Ct2d6sC7ol0/ynxc2pO1cpGUM+f4t5adg=
 github.com/RichardKnop/logging v0.0.0-20190827224416-1a693bdd4fae/go.mod h1:rJJ84PyA/Wlmw1hO+xTzV2wsSUon6J5ktg0g8BF2PuU=
 github.com/RichardKnop/machinery/v2 v2.0.11 h1:BTfLGOmOju3W/OtlZmLX26OjYNZsU4PJo04pQReycdc=
--- a/internal/modules.go
+++ b/internal/modules.go
@@ -46,6 +46,9 @@ import (
 	_ "github.com/Mininglamp-OSS/octo-server/modules/qrcode"
 	_ "github.com/Mininglamp-OSS/octo-server/modules/report"
 	_ "github.com/Mininglamp-OSS/octo-server/modules/search"
+	// searchetl: 消息检索增量 ETL（读 message 分表 → Kafka → es-indexer → OpenSearch，YUJ-4530）。
+	// 阶段 1 仅注册迁移（独立游标表 octo_etl_es_cursor），不启动 scheduler、不接 Kafka。
+	_ "github.com/Mininglamp-OSS/octo-server/modules/searchetl"
 	_ "github.com/Mininglamp-OSS/octo-server/modules/space"
 	_ "github.com/Mininglamp-OSS/octo-server/modules/statistics"
 	_ "github.com/Mininglamp-OSS/octo-server/modules/thread"
--- a/modules/searchetl/1module.go
+++ b/modules/searchetl/1module.go
@@ -1,0 +1,28 @@
+package searchetl
+
+import (
+	"embed"
+
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/register"
+)
+
+//go:embed sql
+var sqlFS embed.FS
+
+// 模块注册（YUJ-4530 阶段 1 骨架）。
+//
+// 阶段 1 只注册迁移（建独立游标表 octo_etl_es_cursor）。**不**启动 scheduler、**不**接 Kafka——
+// producer 的事务拆分 / 单副本互斥 / Kafka 投递在阶段 2/3 落地，届时再在此挂 Start/Stop 钩子。
+// 这样阶段 1 PR 可独立编译/迁移 dry-run，且不在生产引入任何运行期行为变更。
+func init() {
+	register.AddModule(func(ctx interface{}) register.Module {
+		appCtx := ctx.(*config.Context)
+		return register.Module{
+			Name:   "searchetl",
+			SQLDir: register.NewSQLFS(sqlFS),
+			// Service 暴露 ETL，便于后续 ops 端点/测试触达（阶段 1 仅 dry-run 能力）。
+			Service: NewETL(appCtx),
+		}
+	})
+}
--- a/modules/searchetl/config.go
+++ b/modules/searchetl/config.go
@@ -1,0 +1,90 @@
+package searchetl
+
+import (
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+	"go.uber.org/zap"
+)
+
+// searchetl 消息检索 ETL（YUJ-4530）的运行参数。沿用 opanalytics 的 env 钳制范式。
+//
+// 阶段 1（本骨架）：仅 batch / lag / tick 三参就位，模块空跑游标、不接 Kafka。
+// lag 的稳定性语义与 opanalytics 完全一致（见 envLagSeconds 注释），是防丢消息硬条件 C1。
+const (
+	// envBatch 单次 keyset 分页从 message 分片抽取的行数上限。
+	envBatch     = "OCTO_SEARCHETL_BATCH"
+	defaultBatch = 5000
+	minBatch     = 100
+	maxBatch     = 50000
+
+	// envLagSeconds 抽取稳定性滞后窗口（秒）。message.id 在 INSERT 时分配、COMMIT 时才可见，
+	// 提交顺序≠id 顺序：低 id 的事务可能晚于高 id 提交。若严格按 id>cursor 推进，会漏掉
+	// 「已被游标越过、之后才提交」的低 id 行。对策：只处理 created_at ≤ DB_NOW-lag 的稳定
+	// 前缀，游标只推进到该前缀末尾。要求 lag > 单条消息落库事务的最大时长。
+	//
+	// 🔴 硬条件 C1（STOP）：生产环境 lag 不得为 0、不得删除该过滤。设 0 仅限单实例可控测试。
+	envLagSeconds     = "OCTO_SEARCHETL_LAG_SECONDS"
+	defaultLagSeconds = 600
+	maxLagSeconds     = 86400
+
+	// envTickSeconds 增量 tick 周期（秒）。opanalytics 是每日 cron，searchetl 改分钟级 tick。
+	envTickSeconds     = "OCTO_SEARCHETL_TICK_SECONDS"
+	defaultTickSeconds = 60
+	minTickSeconds     = 5
+	maxTickSeconds     = 3600
+)
+
+// batchSize 返回 keyset 分页大小（钳制到 [min,max]）。
+func batchSize() int {
+	return clampInt(envBatch, defaultBatch, minBatch, maxBatch)
+}
+
+// lagSeconds 返回稳定性滞后窗口秒数（钳制到 [0,max]；0 仅测试用）。
+func lagSeconds() int64 {
+	v := os.Getenv(envLagSeconds)
+	if v == "" {
+		return defaultLagSeconds
+	}
+	n, err := strconv.ParseInt(v, 10, 64)
+	if err != nil {
+		log.Warn("invalid OCTO_SEARCHETL_LAG_SECONDS, using default",
+			zap.String("value", v), zap.Int64("default", defaultLagSeconds), zap.Error(err))
+		return defaultLagSeconds
+	}
+	if n < 0 {
+		return 0
+	}
+	if n > maxLagSeconds {
+		return maxLagSeconds
+	}
+	return n
+}
+
+// tickInterval 返回增量 tick 周期。
+func tickInterval() time.Duration {
+	return time.Duration(clampInt(envTickSeconds, defaultTickSeconds, minTickSeconds, maxTickSeconds)) * time.Second
+}
+
+// clampInt 读取整型 env 并钳制到 [min,max]；缺省/非法回退 def。
+func clampInt(env string, def, min, max int) int {
+	v := os.Getenv(env)
+	if v == "" {
+		return def
+	}
+	n, err := strconv.Atoi(v)
+	if err != nil {
+		log.Warn("invalid searchetl int config, using default",
+			zap.String("env", env), zap.String("value", v), zap.Int("default", def), zap.Error(err))
+		return def
+	}
+	if n < min {
+		return min
+	}
+	if n > max {
+		return max
+	}
+	return n
+}
--- a/modules/searchetl/etl.go
+++ b/modules/searchetl/etl.go
@@ -1,0 +1,83 @@
+package searchetl
+
+import (
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+	"go.uber.org/zap"
+)
+
+// ETL 是 searchetl 消息检索增量抽取器（YUJ-4530，克隆 opanalytics 游标范式）。
+//
+// 目标架构：读 message 5 分表 → 投 Kafka topic octo.message.v1 → es-indexer 消费 → OpenSearch。
+// 撤回/删除走读时查询侧 join（路线甲），producer 只跑正文一条流。
+//
+// 🚧 阶段 1（本骨架）：只做「空跑游标」——遍历分片、读稳定前缀、记录将会推进到的水位，
+// 但**不接 Kafka、不推进游标**。事务拆分（读事务→事务外投 Kafka→推进事务）+ 单副本重入
+// 互斥 + 整 chunk 原子重投在阶段 2/3 落地（plan §3.5）。稳定性闸门（C1）已在此就位。
+type ETL struct {
+	log.Log
+	ctx   *config.Context
+	db    *etlDB
+	batch int
+	lag   int64
+}
+
+// NewETL 创建 ETL。
+func NewETL(ctx *config.Context) *ETL {
+	return &ETL{
+		Log:   log.NewTLog("SearchETL"),
+		ctx:   ctx,
+		db:    newETLDB(ctx),
+		batch: batchSize(),
+		lag:   lagSeconds(),
+	}
+}
+
+// RunIncrementalDryRun 跑一轮「空跑游标」（阶段 1）：逐分片读稳定前缀、统计稳定行数与积压，
+// **不投 Kafka、不推进游标**。用于骨架自检与上线前观察源读取/稳定性闸门是否符合预期。
+//
+// 阶段 2 将以此为基替换为真实 RunIncremental：读事务取稳定前缀 → 事务外整批投 Kafka 确认 →
+// 短事务推进游标到稳定前缀末（at-least-once + ES _id 幂等 sink = sink 处 effectively-once）。
+func (e *ETL) RunIncrementalDryRun() error {
+	nowUnix, err := e.db.dbNowUnix()
+	if err != nil {
+		return err
+	}
+	cutoff := nowUnix - e.lag
+
+	var totalStable, totalBacklog int64
+	for _, table := range e.db.messageTables() {
+		if err = e.db.ensureCursor(table); err != nil {
+			return err
+		}
+		cursor, lerr := e.db.loadCursor(table)
+		if lerr != nil {
+			return lerr
+		}
+		maxID, merr := e.db.maxID(table)
+		if merr != nil {
+			return merr
+		}
+		rows, rerr := e.db.readBatch(table, cursor, e.batch)
+		if rerr != nil {
+			return rerr
+		}
+		stable := stablePrefix(rows, cutoff)
+		totalStable += int64(len(stable))
+		if maxID > cursor {
+			totalBacklog += maxID - cursor
+		}
+		e.Debug("searchetl dry-run shard scanned",
+			zap.String("table", table),
+			zap.Int64("cursor", cursor),
+			zap.Int64("max_id", maxID),
+			zap.Int("read", len(rows)),
+			zap.Int("stable", len(stable)))
+	}
+
+	e.Info("searchetl incremental dry-run done (no Kafka, no cursor advance)",
+		zap.Int64("stable_rows", totalStable),
+		zap.Int64("backlog_ids", totalBacklog),
+		zap.Int64("lag_seconds", e.lag))
+	return nil
+}
--- a/modules/searchetl/etl_db.go
+++ b/modules/searchetl/etl_db.go
@@ -1,0 +1,80 @@
+package searchetl
+
+import (
+	"fmt"
+
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/gocraft/dbr/v2"
+)
+
+// etlDB 是 searchetl 的数据访问层（读 message 分片 + 独立游标表 octo_etl_es_cursor）。
+//
+// 阶段 1（本骨架）：只读——ensureCursor / readStablePrefix / dbNowUnix。游标推进（advanceCursor）
+// 在阶段 2 接 Kafka、确认投递成功后才调用（事务拆分见 plan §3.5）；阶段 1 空跑不推进。
+type etlDB struct {
+	ctx     *config.Context
+	session *dbr.Session
+}
+
+func newETLDB(ctx *config.Context) *etlDB {
+	return &etlDB{ctx: ctx, session: ctx.DB()}
+}
+
+// messageTables 枚举全部 message 分片表（与 opanalytics/message 模块分片集一致）。
+func (d *etlDB) messageTables() []string {
+	count := d.ctx.GetConfig().TablePartitionConfig.MessageTableCount
+	if count <= 0 {
+		return []string{"message"}
+	}
+	tables := make([]string, 0, count)
+	tables = append(tables, "message")
+	for i := 1; i < count; i++ {
+		tables = append(tables, fmt.Sprintf("message%d", i))
+	}
+	return tables
+}
+
+// ensureCursor 确保分片水位行存在（首次为 0），使后续 FOR UPDATE 总能命中行串行化。
+func (d *etlDB) ensureCursor(table string) error {
+	_, err := d.session.InsertBySql(
+		"INSERT IGNORE INTO octo_etl_es_cursor (shard_table, last_id) VALUES (?, 0)", table).Exec()
+	return err
+}
+
+// dbNowUnix 返回数据库当前时间（纪元秒），作为稳定性闸门统一时基（避免应用/DB 时钟偏差）。
+func (d *etlDB) dbNowUnix() (int64, error) {
+	var now int64
+	err := d.session.SelectBySql("SELECT UNIX_TIMESTAMP()").LoadOne(&now)
+	return now, err
+}
+
+// loadCursor 读取某分片当前水位（只读，不加锁；阶段 1 空跑用）。
+func (d *etlDB) loadCursor(table string) (int64, error) {
+	var cursor int64
+	err := d.session.SelectBySql(
+		"SELECT last_id FROM octo_etl_es_cursor WHERE shard_table=?", table).LoadOne(&cursor)
+	return cursor, err
+}
+
+// maxID 返回某分片当前最大主键 id（用于积压量 max(id)-cursor 监控；阶段 1 空跑用）。
+func (d *etlDB) maxID(table string) (int64, error) {
+	var maxID int64
+	// COALESCE 兜底空表返回 0。
+	err := d.session.SelectBySql(
+		fmt.Sprintf("SELECT COALESCE(MAX(id),0) FROM `%s`", table)).LoadOne(&maxID)
+	return maxID, err
+}
+
+// readBatch 从某分片按 keyset 读一批（id>cursor 升序 LIMIT batch）。
+//
+// 阶段 1：不开事务、不加 FOR UPDATE（仅空跑读取，不推进游标）。源 SELECT 已含 message_id /
+// payload 与稳定性闸门所需 created_unix——阶段 2 接 Kafka 时直接复用同一行结构，无需再改 SQL。
+func (d *etlDB) readBatch(table string, cursor int64, batch int) ([]*srcMessageRow, error) {
+	var rows []*srcMessageRow
+	_, err := d.session.SelectBySql(
+		fmt.Sprintf("SELECT id, message_id, from_uid, channel_id, channel_type, `timestamp`, "+
+			"UNIX_TIMESTAMP(created_at) AS created_unix, payload "+
+			"FROM `%s` WHERE id>? ORDER BY id ASC LIMIT ?", table),
+		cursor, batch).Load(&rows)
+	return rows, err
+}
--- a/modules/searchetl/etl_lock.go
+++ b/modules/searchetl/etl_lock.go
@@ -1,0 +1,94 @@
+package searchetl
+
+import (
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	octoredis "github.com/Mininglamp-OSS/octo-server/pkg/redis"
+	rd "github.com/go-redis/redis"
+)
+
+// etlRunLockKey 是 searchetl 增量抽取的分布式互斥锁 key。与 opanalytics 的
+// `opanalytics:etl:run` 独立——两条 ETL 各跑各的游标，绝不共用一把锁。
+//
+// 🔴 单副本互斥（C3）：plan §3.5 把 opanalytics 的「读游标 + 推进同一 FOR UPDATE 事务」
+// 拆成三段后，FOR UPDATE 行级第二防线消失。须靠本 Redis 锁全程持有 + 续租，且续租失败必须
+// 立即 abort 在飞批次（阶段 3 落地）。阶段 1 骨架先就位锁原语。
+const etlRunLockKey = "searchetl:etl:run"
+
+// etlRunLockTTL 锁租约。执行期间定期续租以覆盖可能超过单 TTL 的长任务（如全量 backfill）；
+// 进程崩溃则 TTL 自动释放。
+const etlRunLockTTL = 30 * time.Minute
+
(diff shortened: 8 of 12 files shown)


Jerry-Xin (dismissed):
No blocking issues found; this PR is in scope for octo-server and cleanly adds an isolated `searchetl` scaffold plus cursor migration without starting runtime behavior.

💬 Non-blocking

🟡 Warning — `RunIncrementalDryRun` reports `stable_rows` from only the first batch per shard, not the full stable backlog. With more than `batch` stable rows, the log can significantly understate how much would be processed once the real runner loops chunks. Either rename the metric to clarify it is “first batch stable rows” or loop like `opanalytics.RunIncremental` when this becomes an ops-facing dry run. See `modules/searchetl/etl.go:61` and `modules/searchetl/etl.go:65`.

🟡 Warning — `lagSeconds()` still permits `0` and maps negative values to `0`, while the comments mark nonzero production lag as a hard data-loss condition. This is acceptable while no scheduler or cursor advancement exists, but before Phase 2 starts advancing cursors, production config should reject or fail closed on `OCTO_SEARCHETL_LAG_SECONDS=0`. See `modules/searchetl/config.go:45` and `modules/searchetl/config.go:57`.

✅ Highlights

🔵 Suggestion — The table-name construction is bounded by the existing partition config pattern and mirrors `opanalytics`, so the raw SQL table interpolation is not user-controlled in this code path. See `modules/searchetl/etl_db.go:24` and `modules/searchetl/etl_db.go:63`.

🔵 Suggestion — The stability gate is small, test-covered, and preserves the important “no-hole prefix” behavior needed before cursor advancement is introduced. See `modules/searchetl/stability.go:13` and `modules/searchetl/stability_test.go:44`.

Verification run: `go test ./modules/searchetl`, `go vet ./modules/searchetl ./internal`, and `go test ./internal` passed. I also tried an `opanalytics` package test as context; it failed on the local test database migration state (`unknown migration in database`), which does not appear caused by this PR.

yujiawei (commented):
# Code Review — PR #365 (octo-server)

**Scope reviewed:** head `4b6fd6513f52777405fc1ded4502c86e96114910`, merge-base `fde79cfdf1a58a17a114d75b4e11540394e1e5c8`, 10 files / +494. This is a phase-1 scaffold for the message-search ETL (YUJ-4530): it registers a new `searchetl` module, adds one DDL migration for an independent cursor table, and lays down the cursor/stability/lock primitives without wiring any scheduler or Kafka path.

## 1. Verification results

- ✅ **Builds clean** — `go build ./...` passes (head SHA).
- ✅ **Tests pass** — `go test ./modules/searchetl/...` green; 5 table-driven cases in `stability_test.go` cover the C1 stability-gate semantics including the key low-id-late-commit no-hole case.
- ✅ **Lint gate** — `go vet ./modules/searchetl/...` clean (repo CI runs a govet-only profile per `.golangci.yml`).
- ✅ **Scope is correct for a scaffold** — `Start`/`Stop` hooks are intentionally nil, so the module registers but starts no scheduler and opens no Kafka/Redis connection at boot. Confirmed against `octo-lib/module/module.go`: `Start()`/`Stop()` only invoke non-nil hooks; `searchetl` provides none.
- ✅ **Cursor table is physically isolated** — `octo_etl_es_cursor` (`sql/20260614000001_searchetl_init.sql`) is distinct from opanalytics' `octo_etl_message_cursor`; migration id is globally unique across all modules; DDL (engine/charset/collation/PK) matches the opanalytics cursor convention.
- ✅ **Stability gate (C1) is faithfully extracted** — `stability.go:stablePrefix` reproduces opanalytics `etl_db.go:93-106` cutoff logic (`CreatedUnix > cutoff` → truncate at first unstable row, no-hole prefix). `lagSeconds()` clamps and defaults to 600s with the documented production guard.
- ✅ **Redis lock is a faithful clone** — `etl_lock.go` matches opanalytics `etl_lock.go` (SET NX EX + Lua CAS-DEL / CAS-PEXPIRE), with its own key `searchetl:etl:run` separate from `opanalytics:etl:run`.
- ✅ **No injection surface** — table names interpolated via `fmt.Sprintf` originate from `MessageTableCount` (an int from config), not user input; all value bindings use placeholders (`INSERT IGNORE ... VALUES (?, 0)`, keyset `WHERE id>? LIMIT ?`). For a `security_sensitive`-classified PR this is the main thing to confirm, and it is clean.

## 2. Findings

No P0 / P1 issues. The PR compiles, tests pass, stays within the stated phase-1 scope (only `searchetl/*`, the blank import in `internal/modules.go`, and one migration — no changes to the write path, message schema, opanalytics, or any Kafka component), and correctly defers the risky parts (transaction splitting, cursor advance, single-replica mutual exclusion, Kafka delivery) to later phases with the stability gate already in place.

### P2 — non-blocking

- **P2-a (comment accuracy): "no runtime behavior change" is slightly imprecise.** `1module.go` registers a non-nil `SQLDir`, and `octo-lib/module/module.go:Setup` runs every module's migrations on every boot. So phase 1 *does* execute a DDL migration in production (creating `octo_etl_es_cursor`) — which is the intended and correct deliverable, but it is not literally "no runtime change." Suggest tightening the comment to "no runtime *logic*/scheduler change; adds one idempotent DDL migration." The migration itself is safe and idempotent (`CREATE TABLE` + clean `-- +migrate Down`).
- **P2-b (dead code vs. future lint profile):** the entire `etl_lock.go`, `RunIncrementalDryRun`, and most of `etl_db.go`/`srcMessageRow` are currently unreferenced (no caller outside the package). This is deliberate, well-documented scaffolding for phases 2/3 and compiles only because the `unused` linter is disabled in the current govet-only profile. `.golangci.yml` has a standing TODO to re-enable the default linters within two release cycles — when `unused` comes back, this module will fail CI until phase 2 wires the code. Worth tracking so the lint re-enable and phase-2 landing are sequenced, or add narrow `//nolint:unused` markers if the profile flips first.

## 3. Suggestions

- Tighten the `1module.go` comment per P2-a so the "no runtime change" claim isn't read literally by a future maintainer auditing what phase 1 shipped.
- When phase 2 lands, the source `SELECT` in `etl_db.go:readBatch` already includes `message_id`/`payload`, so the row contract won't need to change — good forward-design. Just make sure the phase-3 "ES-confirm → then advance cursor" ordering and the lock-renew-fail → abort-in-flight behavior land together; the scaffold comments correctly call these out as the real correctness risks, but none of that logic is exercised yet, so it carries no test coverage in this PR (expected for a scaffold).

## 4. Additional observations

- `NewETL(appCtx)` is constructed eagerly at module registration and calls `ctx.DB()` via `newETLDB`. This matches the existing `channel`/`user` module pattern (eager `Service:`), and crucially it does **not** construct the Redis lock at boot (`newETLLock` is never called), so no Redis connection is opened in phase 1. No resource-leak or boot-ordering concern.
- The `dbNowUnix()` use of `SELECT UNIX_TIMESTAMP()` as the single time base (avoiding app/DB clock skew) is the right choice and mirrors opanalytics.

**Verdict:** This is a clean, well-scoped, well-commented scaffold that faithfully reuses the proven opanalytics cursor-ETL paradigm and correctly defers all the dangerous parts. The two P2 items are non-blocking maintainability nits. Approve.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#365 Review Report
**Head SHA:** `4b6fd6513f52777405fc1ded4502c86e96114910`
**PR:** feat(searchetl): scaffold message-search ETL module + cursor table ([automated review](mention://issue/ca2914ab-3209-4b5a-933b-ee0109dfc4db) phase 1)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Migration SQL correctness | ✅ | `sql/20260614000001_searchetl_init.sql` — clean DDL, PK on `shard_table`, `utf8mb4`, isolated from `octo_etl_message_cursor` |
| Stability gate logic (C1) | ✅ | `stability.go:14-20` — correct no-hole prefix cut; first `CreatedUnix > cutoff` truncates; tested incl. low-id/late-commit case |
| Stability gate tests | ✅ | `stability_test.go` — 5 tests covering all-stable, tail-unstable, head-unstable, low-id-late-commit (C1 core), empty |
| Redis lock primitives | ✅ | `etl_lock.go` — SET NX EX + Lua CAS-DEL + Lua CAS-PEXPIRE; matches opanalytics verbatim (key name differs) |
| SQL injection safety | ✅ | All `fmt.Sprintf` table-name interpolation uses `"message%d"` with integer from internal config; all value params use `?` placeholders |
| Error propagation | ✅ | All DB errors propagated (no swallowing); config parse errors fall back to defaults with `log.Warn` |
| Module registration | ✅ | `1module.go` — `register.AddModule` with `SQLDir` + `Service`; no `Start`/`Stop`/`SetupAPI` (Phase 1 intentional) |
| Config clamping | ✅ | `config.go` — batch/lag/tick all clamped to safe ranges; env parse failures fall back to defaults |
| Data-flow: cursor → readBatch → stablePrefix | ✅ | `etl.go:48-85` — cursor loaded → keyset read → stability gate → count only; no cursor advance, no Kafka produce |
| No behavioral change in prod | ✅ | No scheduler started, no HTTP endpoints, no Kafka; `RunIncrementalDryRun` exposed as service but not auto-triggered |

## 2. Findings

### F1 — P2: `readBatch` fetches `payload` (MEDIUMBLOB) but Phase 1 never consumes it

**File:** `modules/searchetl/etl_db.go:75-77`
**Diff-scope:** New in this PR.

`readBatch()` SELECTs `payload` (MEDIUMBLOB, up to 16MB/row) alongside `id`, `message_id`, etc. In Phase 1, only `ID` and `CreatedUnix` are consumed (by `stablePrefix`); `Payload`, `MessageID`, `FromUID`, `ChannelID`, `ChannelType`, `Timestamp` are fetched into Go heap and immediately GC'd. With `defaultBatch=5000` × 5 shards, worst-case memory pressure is non-trivial.

The PR body explicitly states this is intentional Phase 2 forward-compat ("源 SELECT 已含 message_id / payload ... 阶段 2 接 Kafka 时直接复用同一行结构，无需再改 SQL").

**R1 check:** Does not make any working path unavailable or produce user-visible errors. `RunIncrementalDryRun` is a diagnostic function with no auto-trigger. Typical message payloads (1-10KB) × 5000 rows × 5 shards = 25-250MB, manageable for an infrequent diagnostic. **Not P1.**

**Recommendation for Phase 2:** Either (a) drop `payload` from SELECT in Phase 1 and re-add in Phase 2, or (b) keep as-is with an explicit comment marking the trade-off (already present at `etl_db.go:72-73`).

### F2 — P2: `lagSeconds()` silently clamps negative env values to 0

**File:** `modules/searchetl/config.go:51-53`
**Diff-scope:** **Pre-existing pattern** — verbatim clone from `opanalytics/config.go:89-103` (`etlLagSeconds()`). Identical logic: `if n < 0 { return 0 }` with no warning log.

The comment at `config.go:28` marks lag=0 as a hard stop condition (🔴 C1 STOP: prod lag must never be 0). Yet `OCTO_SEARCHETL_LAG_SECONDS=-1` silently yields lag=0 (disabling the stability gate), while `OCTO_SEARCHETL_LAG_SECONDS=abc` safely falls back to 600 with a warning. The silent path to the dangerous value is the wrong way around.

**R2 check:** This is pre-existing behavior (opanalytics has the identical code). The PR did not introduce or amplify this defect. Per R2, qualifies as pre-existing P2/note.

**R1 check (even if new):** In Phase 1, no cursor advances, so lag=0 cannot cause data loss. The dry-run just reports different stable-row counts. No user-visible error. **Not P1.**

**Recommendation for Phase 2:** Before cursor advancement goes live, either (a) log a warning when negative→0, or (b) clamp negative to `defaultLagSeconds` (matching the parse-error path), or (c) set `minLagSeconds > 0`.

### Notes (non-blocking)

- **etlLock is dead code in Phase 1** (`etl_lock.go`): Fully implemented but never instantiated. `ETL` struct holds no lock reference. Intentional — Phase 1 is dry-run only. Phase 2 wiring must ensure lock lifecycle (`Close()`) is managed.
- **`RunIncrementalDryRun` reads only one batch per shard** (`etl.go:55-68`): Unlike opanalytics which loops to drain all rows, searchetl reads a single batch. Intentional — `totalBacklog` from `maxID - cursor` gives the full estimate without reading all rows.
- **`1module.go` filename**: Leading `1` is a Go compilation-ordering convention used across the codebase (opanalytics uses the same pattern). Consistent.

## 3. Recommendations

1. **Phase 2 pre-req:** Fix F2 (lag=0 silent path) before cursor advancement goes live. The stability gate is the hard anti-data-loss condition C1; its config must not silently allow disabling.
2. **Phase 2 pre-req:** Re-evaluate F1 (payload in SELECT) — either keep the forward-compat or defer the column to reduce Phase 1 memory footprint.
3. **Phase 3 pre-req:** Wire `etlLock` into the ETL loop with renewal goroutine + abort-on-renew-failure (as documented in `etl_lock.go:22`).

## 4. Extra Findings

None beyond the above. The module is cleanly isolated from `opanalytics` (separate cursor table, separate lock key, separate config prefix). No naming collision with the existing `search` module.

## 5. Data-Flow Trace

```
dbNowUnix()  →  nowUnix (int64, DB clock epoch seconds)
                  ↓
              cutoff = nowUnix - lag
                  ↓
for each table in messageTables():
  ensureCursor(table)     → INSERT IGNORE (idempotent seed, side-effect: creates cursor row)
  loadCursor(table)       → cursor (int64, PK lookup from octo_etl_es_cursor)
  maxID(table)            → maxID (int64, COALESCE(MAX(id),0) from message shard)
  readBatch(table, cursor, batch)
    → []*srcMessageRow (keyset: id>cursor ORDER BY id ASC LIMIT batch)
    → Columns: id, message_id, from_uid, channel_id, channel_type, timestamp, created_unix, payload
    → All sourced from DB (no user input in SQL beyond parameterized cursor/batch)
  stablePrefix(rows, cutoff)
    → Scans rows in id-asc order; truncates at first CreatedUnix > cutoff
    → Returns rows[:i] (stable prefix)
  Consumption:
    - totalStable += len(stable)     ← only length consumed, not row data
    - totalBacklog += maxID - cursor ← no row data consumed
    - Debug log: table, cursor, max_id, read count, stable count
```

**Key observation:** Row data (Payload, MessageID, etc.) flows from DB to Go heap but is never consumed in Phase 1. Only `ID` and `CreatedUnix` fields are read by `stablePrefix`. This is the F1 finding above.

## 6. R5 Blindspot Checklist (security_sensitive — mandatory)

- **C1 — Dual-path parity:** N/A. Phase 1 adds a new module with no symmetric add/remove, create/delete, or subscribe/unsubscribe paths. The cursor table has only INSERT IGNORE (ensure) and SELECT (load). No delete/remove path exists yet.
- **C2 — Control-flow ordering / nested reuse:** Clear. `stablePrefix()` is called exactly once per shard per dry-run invocation. No nested or repeated calls. `fmt.Sprintf` table-name interpolation uses fixed format `"message%d"` with an integer loop counter — not a reusable function reachable from external input.
- **C3 — Authorization boundary ≠ capability boundary:** N/A for Phase 1. The module exposes no HTTP endpoints, Kafka producers, or external-facing APIs. `RunIncrementalDryRun` is exposed as a Go service via `register.Module.Service` but has no HTTP trigger or scheduler. Phase 2 (Kafka producer + scheduler) will need to address authorization for manual triggers (opanalytics gates manual trigger behind `superAdmin` middleware).

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. Two P2 recommendations (F1: payload memory waste in dry-run; F2: silent lag=0 config path inherited from opanalytics) should be addressed before Phase 2 (Kafka integration + cursor advancement), but do not block the Phase 1 skeleton.

lml2468 (dismissed):
✅ APPROVE — phase-1 scaffold 干净,无 blocking。本地逐条核实(SQL schema / 锁 / cursor)+ build/vet/test 全绿。pr-review workflow 51 agent / 5 维 / 对抗验证。

## 核实通过(本地)
1. **SQL migration**(sql/20260614000001)— last_id BIGINT(雪花/主键不溢出)✅;有 Up/Down、INSERT IGNORE 幂等 ✅;shard_table VARCHAR(64) PK,值为 ASCII 表名常量、不与他表跨库比较,无 collation 1267 风险 ✅。
2. **分布式锁**(etl_lock.go)— Acquire 用 SET NX EX 原子 ✅;Release/Renew 走 Lua CAS(GET==token 才 DEL/PEXPIRE),owner-only、防误删后继 owner 锁 ✅;TTL 30min + 崩溃自动释放 ✅。
3. **cursor 读取**(etl_db.go readBatch)— keyset `WHERE id>? ORDER BY id ASC LIMIT ?`,`>` 非 `>=` 无重读;cursor 全程 int64、不经 float64,雪花精度安全 ✅。
4. **phase-1 是 dry-run**(etl.go RunIncrementalDryRun:"no Kafka, no cursor advance",advanceCursor 未实现)— 故下列 should-fix 在 phase-1 不会触发,但 phase-2 启用真实推进前必须补。

## should-fix(phase-2 启用前必补,非 phase-1 blocking)
1. 🟡 **lag=0 生产未强制**(config.go lagSeconds)— 允许 OCTO_SEARCHETL_LAG_SECONDS=0(注释自标 🔴 STOP 生产不得为 0,但无运行时校验)。phase-1 dry-run 无害,但 phase-2 真实推进时 lag=0 会让稳定前缀为空/游标卡死甚至漏扫。建议 NewETL 启动时:生产环境 lag<=0 → panic/fatal。
2. 🟡 **Renew/Release 的 Lua 返回值类型断言静默**(etl_lock.go:84 `n,ok:=res.(int64); return ok&&n==1,nil`)— ok==false 时返回 (false,nil),与"锁丢失"不可区分。注:这与本模块明确仿照的 opanalytics/etl_lock.go:81 **逐字一致**(非本 PR 引入的回归);repo 另有更严范式 oidc/sync_lock.go:96(类型断言失败返 error)。建议 phase-2 接入续租前采用 oidc 的严格写法,避免误判 owner 变更而误 abort 在飞批次(C3)。
3. 🟡 锁/cursor/SQL round-trip 暂无单测(仅 stability_test)— scaffold 可接受,但 phase-2 落地前应补锁 ownership + cursor no-skip 的回归测试。

## nit
shard_table VARCHAR 显式 collation 对齐(无实际风险)、readBatch 未对 batch 再校验(上游 clampInt 已钳)等,可忽略。

phase-1 scaffold 质量好(锁原语正确就位、SQL 类型对、keyset 无重读、config 钳制完善)。无 blocking,should-fix 都是 phase-2 开关前的护栏,建议在 phase-2 PR 一并落。批准。

—— Octo-PR (lml2468)

lml2468 (approved):
✅ APPROVE(重审 @1331b80)— 本 commit 仅 octo-lib 依赖 bump,searchetl 模块代码零改动,我此前的完整审查结论对模块代码全部仍成立。

## 本 commit 变更
go.mod/go.sum 各 1-3 行:octo-lib v0.0.0-20260605071834-330ed3ce7e6a → v0.0.0-20260614033907-f62e626a34db。`git diff 4b6fd651..1331b80f --stat` 确认只动 go.mod/go.sum,modules/searchetl/** 一字未改。

## 验证(新 octo-lib 下)
- modules/searchetl build / vet / test 全绿(stability_test 通过)。
- 全仓 go build ./... 干净(bump 未破坏任何 caller)。

## 沿用上轮审查结论(模块代码未变,无需重跑)
phase-1 scaffold 无 blocking:SQL last_id BIGINT + 幂等迁移、锁 SET NX EX + Lua CAS owner-only、cursor keyset id>? 无重读、phase-1 dry-run 不推进游标。
should-fix(phase-2 启用前必补,phase-1 不触发,四人共识):① lag=0/负值生产强制(panic/warn)② 锁 Lua 返回值类型断言严格化(采 oidc/sync_lock.go 写法)③ dry-run 指标口径统一(stable_rows 仅首批 vs backlog 全量,Steve)④ 未用大字段 payload 暂 fetch(mochashanyao)⑤ 补锁/cursor 单测。建议 phase-2 PR 一并落。

依赖 bump 安全、模块结论不变,重投 APPROVE。

—— Octo-PR (lml2468)

Jerry-Xin (dismissed):
This PR is in scope for octo-server, but the core “stable prefix” correctness claim is not sound for the existing message write path.

**🔴 Blocking**

🔴 Critical: `created_at <= DB_NOW - lag` does not prevent low-id late-commit skips once the ETL starts advancing cursors.
References: `modules/searchetl/stability.go:5`, `modules/searchetl/etl_db.go:74`, `modules/webhook/api.go:242`, `modules/webhook/api.go:274`

The source query only sees committed rows. The message table’s `created_at` is a `DEFAULT CURRENT_TIMESTAMP` insert timestamp, while webhook message persistence inserts rows inside a transaction and commits later. A lower-id row can be inserted in a still-open transaction, remain invisible to `readBatch`, while a higher-id row from another transaction is already committed and has an old enough `created_at`. `stablePrefix` will treat the visible higher-id rows as stable and future phase cursor advancement would skip the lower-id row when it commits.

The unit test at `modules/searchetl/stability_test.go:43` does not reproduce the dangerous case because it includes the late row in the batch. The actual failure case is that the row is absent from the `SELECT` result until after the cursor has moved past it.

This needs a different stability fence before the scaffold is merged as the foundation for phase 2, for example a delayed captured high-water mark / commit-time outbox / commit-order source, or another mechanism that accounts for invisible lower ids. Add a DB-level test that holds a low-id insert open, commits a higher-id row, runs the ETL fence, and proves the cursor cannot advance past the hidden row.

**💬 Non-blocking**

🟡 Warning: `go.mod:9` bumps `github.com/Mininglamp-OSS/octo-lib`, but the new code does not appear to require new octo-lib APIs. This also conflicts with the PR text saying the dependency pin bump is deferred to phase 2. If this commit is not already merged and intentionally required, keep the existing pin in this PR.

🔵 Suggestion: `RunIncrementalDryRun` reads only one batch per shard (`modules/searchetl/etl.go:61`) and never advances even a local dry-run cursor, so `stable_rows` can undercount stable backlog by orders of magnitude. Either loop locally for observability or name/log it as “first batch stable rows” to avoid misleading rollout checks.

**✅ Highlights**

The module registration is isolated and does not start a scheduler. The cursor table is additive and separate from `opanalytics`. `go test ./modules/searchetl -run TestStablePrefix -count=1` and `go vet ./modules/searchetl ./internal` pass locally. A broader `opanalytics` test run was blocked by the local test database migration state, not by this PR.

Jerry-Xin (approved):
🔄 Re-review of new head `1331b80` (prior approved head: `4b6fd65`). **APPROVE.**

**What changed since my last approval**
The only delta `4b6fd65..1331b80` is a `go.mod`/`go.sum` pin bump:
`octo-lib 20260605...330ed3ce` → `20260614033907-f62e626a34db`. No `searchetl`
logic changed (stability gate, cursor read, lock, dry-run runner are byte-identical
to the commit I already approved).

**Pin bump is correct sequencing, not a STOP violation**
`f62e626a34db` is octo-lib **#76 merged to `main`** (it is the current `main` HEAD —
`compare main...f62e626 = identical`). So the pin now points at a real merged SHA, fully
satisfying plan STOP #11 (never pin to an unmerged commit). `go build ./...` and the CI
**Build** check both pass against the new pin. The PR body line "pin bump deferred to phase 2"
is simply now stale because octo-lib#76 landed; bumping here is the right order. No
`contract/searchmsg` import is added yet (grep confirms only comments mention Kafka) — the
import wiring stays for phase 2, as intended.

**On the automated 🔴 (low-id late-commit stable-prefix skip)**
I do not treat this as blocking for this PR:
- The cited files (`stability.go`, `etl_db.go`, `etl.go`) are **unchanged** in this commit —
  this is pre-existing code I already reviewed under the prior head.
- Its own harm condition is explicitly phase-2 ("once the ETL starts advancing cursors").
  Phase-1 here is a **dry-run that never advances the cursor and never produces to Kafka**
  (`etl.go` logs "no Kafka, no cursor advance"; no cursor write exists). The skip cannot occur.
- The lag fence + no-hole prefix cut is the documented C1 guardrail and mirrors the proven
  `opanalytics` paradigm. The deeper invisible-low-id concern (open txn commits after a higher
  id) is a legitimate **phase-2 design item** for the real cursor-advancing runner — best
  addressed when the read-tx / Kafka-produce / advance split lands, with a DB-level holding test.

**Prior 🟡 items — status (all still phase-2 guardrails, none addressed by this pin-only commit, none escalated):**
1. lag=0 / negative-lag prod enforcement (suggest startup panic) — unchanged, phase-2.
2. Lock Lua return-type assertion swallow — unchanged copy of opanalytics pattern (not a regression); adopt oidc stricter form in phase-2.
3. `RunIncrementalDryRun` `stable_rows` first-batch-only vs full backlog — dry-run metric semantics; rename/loop in phase-2.
4. Missing lock/cursor unit tests — add alongside phase-2 runner.

**✅ Highlights:** isolated module registration (no scheduler start), additive cursor table
separate from opanalytics, stability-gate unit tests present, clean minimal pin bump that
resolves and builds.

Approving — minimal, correctly-sequenced, scope-appropriate phase-1 change.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#365 Review Report
**PR:** Mininglamp-OSS/octo-server#365
**Head SHA:** `1331b80f39512403f697d77bdd0c5acbc15e9e9c`
**Title:** feat(searchetl): scaffold message-search ETL module + cursor table ([automated review](mention://issue/ca2914ab-3209-4b5a-933b-ee0109dfc4db) phase 1)

---

| Item | Status | Evidence |
|------|--------|----------|
| Module registration pattern | ✅ | `1module.go:18-27` — uses `register.AddModule` + `embed.FS` identical to opanalytics; `Service` field set (no `SetupAPI` — intentional, Phase 1 has no HTTP routes) |
| SQL migration correctness | ✅ | `sql/20260614000001_searchetl_init.sql` — CREATE TABLE `octo_etl_es_cursor` with PK `shard_table`, additive-only, timestamp ordered after opanalytics `20260608000001` |
| Cursor isolation from opanalytics | ✅ | Separate table `octo_etl_es_cursor` vs `octo_etl_message_cursor`; separate lock key `searchetl:etl:run` vs `opanalytics:etl:run` — no cross-ETL interference possible |
| Stability gate (C1 anti-data-loss) | ✅ | `stability.go:18-20` — `stablePrefix()` truncates at first row where `CreatedUnix > cutoff`; identical semantics to opanalytics inline logic; 5 unit tests including the C1 low-id/late-commit reproduction |
| Redis lock primitive (C3 scaffolding) | ✅ | `etl_lock.go` — faithful clone of opanalytics lock; SET NX EX + Lua CAS-DEL/CAS-PEXPIRE; independent key; not called in Phase 1 (no resource leak) |
| SQL injection safety | ✅ | `etl_db.go:62,80` — `fmt.Sprintf` table-name interpolation sourced from `TablePartitionConfig.MessageTableCount` (config integer), never user input; identical codebase-wide convention (opanalytics, message, webhook, thread modules) |
| Config clamping | ✅ | `config.go:44-89` — `batchSize()`/`lagSeconds()`/`tickInterval()` all clamp to safe ranges; defaults match opanalytics (batch=5000, lag=600s) |
| go.mod pin bump | ✅ | Commit `1331b80` bumps octo-lib to `f62e626a34db` (merged commit of octo-lib#76); not pinning to an unmerged commit |
| No runtime behavior change | ✅ | No scheduler started, no Kafka, no cursor advancement; `RunIncrementalDryRun()` is read-only; module compiles and registers migration only |

## 2. Findings

### No P0 or P1 findings.

### P2 — Observations for future phases

**P2-1: Dry-run reads only one batch per shard (no drain loop)**
- **File:** `etl.go:58-78`
- **Diff-scope:** new (Phase 1 design choice)
- opanalytics `RunIncremental()` loops `runChunk()` until `n < batch` to drain each shard. `RunIncrementalDryRun()` reads a single `readBatch()` per shard and stops. This limits observability — the dry-run can only report on the first batch of each shard per invocation, not the full backlog.
- **Impact:** Low — Phase 1 is scaffolding; Phase 2 should add a drain loop for complete dry-run coverage.

**P2-2: `etl_lock.go` is dead code in Phase 1**
- **File:** `etl_lock.go:1-94` (94 lines)
- **Diff-scope:** new (staged for Phase 2/3)
- `newETLLock()` is never called; the lock primitive exists but has no consumer. Adds ~94 lines of unreferenced code.
- **Impact:** Low — documented as intentional scaffolding. Consider whether staging unused code in a Phase 1 PR is worth the review surface vs. introducing it when it's first consumed.

**P2-3: `lagSeconds()` allows lag=0 via negative env values**
- **File:** `config.go:56-58` — `if n < 0 { return 0 }`
- **Diff-scope:** pre-existing pattern (identical to opanalytics `etlLagSeconds()`)
- Negative env values clamp to 0, which disables the stability gate. Documented as "test only" but no runtime enforcement prevents it in production.
- **Impact:** Low — matches opanalytics; the hard constraint (C1: lag must not be 0 in prod) is operational, not enforced in code. Phase 2 scheduler should validate `lag > 0` before starting.

**P2-4: Backlog metric is an approximation**
- **File:** `etl.go:72-74` — `totalBacklog += maxID - cursor`
- **Diff-scope:** new
- `maxID - cursor` counts ID gaps (deleted rows, rolled-back transactions) as backlog. Not a precise row count.
- **Impact:** Low — used for monitoring/dry-run logging only; not consumed by any control logic.

## 3. Suggestions

1. **Phase 2 drain loop:** When implementing `RunIncremental()`, add a chunk-drain loop mirroring opanalytics `runChunk()` pattern. Consider also adding a drain loop to `RunIncrementalDryRun()` for complete pre-launch observability.
2. **Phase 2 lag validation:** In the Phase 2 scheduler `Start()`, add a hard check: `if e.lag <= 0 { return error }` to enforce C1 at runtime, not just by convention.
3. **Phase 2 transaction design:** The planned three-phase transaction split (read tx → Kafka produce outside tx → cursor advance tx) removes the `FOR UPDATE` row-lock second line of defense. The Redis lock (C3) must be fully operational before this split lands.

## 4. Additional Findings

- **Migration safety:** The migration is additive (CREATE TABLE) with a clean DROP on down. No risk to existing tables.
- **`srcMessageRow` forward compatibility:** `model.go` includes `MessageID` and `Payload` fields unused in Phase 1 but already fetched by `readBatch()` SELECT. This is clean forward-compat — Phase 2 won't need to change the SQL.
- **Test quality:** `stability_test.go` covers 5 cases including the critical C1 low-id/late-commit scenario. Tests are pure-function (no DB dependency), fast, and well-documented.

## 5. Data Flow Tracing

|---|---|---|
| `table` (SQL identifier) | `messageTables()` → `TablePartitionConfig.MessageTableCount` (config int) → `fmt.Sprintf("message%d", i)` | ✅ Deterministic, no user input |
| `cursor` (watermark) | `ensureCursor()` INSERT IGNORE (creates row with `last_id=0`) → `loadCursor()` SELECT | ✅ Always returns a valid int64 |
| `rows` (batch data) | `readBatch()` → parameterized SQL `WHERE id>? ORDER BY id ASC LIMIT ?` | ✅ Parameterized query, safe |
| `cutoff` (stability threshold) | `dbNowUnix()` → `SELECT UNIX_TIMESTAMP()` → `nowUnix - e.lag` | ✅ DB-sourced timebase avoids clock skew |
| `stable` (prefix slice) | `stablePrefix(rows, cutoff)` → truncates at first `CreatedUnix > cutoff` | ✅ Pure function, correct semantics |
| `maxID` (backlog metric) | `maxID()` → `SELECT COALESCE(MAX(id),0)` | ✅ COALESCE handles empty tables |
| Redis lock token | Not consumed in Phase 1 (lock not acquired) | N/A |

## 6. R5 Blind-Spot Checklist (security_sensitive — mandatory)

**C1 — Dual-path parity:** N/A. This PR adds a new module (one-directional: register + read-only dry-run). No symmetric paths (add/remove, subscribe/unsubscribe) are introduced. Phase 2 will add cursor advance — parity with opanalytics `runChunk()` should be verified at that point.

**C2 — Control-flow ordering / nesting reuse:** Clear. `stablePrefix()` is a pure function called once per shard with no nesting risk. Lua CAS scripts are standard single-operation patterns. `fmt.Sprintf` SQL identifiers are config-derived (not user input), so non-canonical-form probing is not applicable.

**C3 — Authorization boundary ≠ capability boundary:** N/A. Phase 1 has no HTTP endpoints, no external-facing API. The module is internal infrastructure. Redis lock uses the same authenticated connection pool as the rest of the application via `octoredis.MustBuildOptions`.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The PR is a clean Phase 1 scaffold that faithfully clones the proven opanalytics cursor-ETL paradigm with appropriate simplifications for a read-only dry-run phase. Stability gate (C1) is correctly implemented and tested. Cursor table is physically isolated. No security issues. Four P2 observations are noted for future phases.

yujiawei (commented):
# Code Review — PR #365 (octo-server)

**Scope:** Phase 1 scaffold of the message-search ETL pipeline — module skeleton, cursor table migration, stability gate, and Redis-lock primitives. No Kafka client, no producer logic, no scheduler, no read path. Reviewed at head `1331b80f`.

## Verdict: APPROVED

Clean, well-scoped scaffold. The one piece of real runtime logic (the stability gate) is faithfully ported from the existing `opanalytics` ETL and covered by tests, including the critical low-id/late-commit counterexample. No P0/P1 issues. A few non-blocking observations below.

## Verification

| Item | Result | Evidence |
|---|---|---|
| Builds / vets / tests locally | ✅ | `go build/vet/test ./modules/searchetl/...` all pass; full `go build ./...` passes |
| Stability gate (C1) ported correctly | ✅ | `modules/searchetl/stability.go:13-20` mirrors `modules/opanalytics/etl_db.go` runChunk prefix logic; `stability_test.go:1-61` covers all-stable / tail-unstable / head-unstable / **low-id-late-commit** / empty |
| Cursor table physically isolated | ✅ | `octo_etl_es_cursor` (`sql/20260614000001_searchetl_init.sql:11`) is distinct from opanalytics' `octo_etl_message_cursor` |
| Redis lock primitive | ✅ | `etl_lock.go` is a faithful clone with its own key `searchetl:etl:run` (`etl_lock.go:21`); CAS-DEL release + CAS-PEXPIRE renew preserved |
| Dependency pin points at a merged commit | ✅ | `go.mod:9` bumps octo-lib to `f62e626a34db`, which is the merge commit of the contract PR and identical to octo-lib `main` — no pin to an unmerged SHA |
| Column mapping | ✅ | `srcMessageRow` (`model.go:8-16`) has no `db` tags; dbr camelCase→snake_case maps `MessageID`→`message_id` correctly (acronym handling verified). Source columns (`message_id VARCHAR(20)`, `payload`, `from_uid`, `channel_id`, `channel_type`, `timestamp`, `created_at`, `id`) all exist on the `message` table |
| Phase-1 scope honored | ✅ | `RunIncrementalDryRun` (`etl.go:48-83`) reads the stable prefix but does **not** produce to Kafka and does **not** advance the cursor; no scheduler is started, so registration introduces no runtime behavior change |
| SQL injection surface | ✅ | Table names interpolated via `fmt.Sprintf` (`etl_db.go`) come only from `messageTables()` (config-int derived), never user input; row predicates are parameterized |
| Migration id unique | ✅ | `20260614000001` does not collide with any existing migration |

## Findings (P2 / observations — none blocking)

**P2-1 — Eager service construction at module-registration time.**
`1module.go:21` calls `NewETL(appCtx)` inside the module factory, which calls `ctx.DB()` eagerly. The existing `opanalytics` module instead defers construction into a `SetupAPI` closure. This is safe here because `NewMySQL()` is `sync.Once`-guarded and the module factory runs during `module.Setup` (after DB config is loaded), so it cannot race or double-open. Consider matching the lazy pattern in a later phase for consistency, but no change required now.

**P2-2 — Pre-landed scaffolding that is currently unreferenced.**
The entire `etlLock` type (`Acquire`/`Release`/`Renew`/`Close`) and the `loadCursor`/`maxID` helpers exist but are only exercised by tests / the dry-run. This is intentional and documented as landing in phases 2/3, and golangci-lint passes, so it is acceptable. Flagging only so a future reviewer doesn't mistake it for dead code: please ensure phases 2/3 actually wire it, otherwise it should be removed.

**P2-3 — Dependency pin is bumped ahead of first use.**
The contract package from the bumped octo-lib version is referenced only in a comment (`model.go:7`), not yet imported. Bumping the pin now is harmless (build is green) and sets up the next phase, but the dependency it points at is not yet consumed by this PR. Acceptable as the closing step of the cross-repo merge order; noted for completeness.

**Observation — dry-run reads a single batch per shard.**
`RunIncrementalDryRun` reads one `readBatch` window per shard (no chunk loop), so its `stable_rows` counter reflects only the first batch, not the full backlog. This is fine for a self-check / observation hook and is consistent with "dry run" intent.

## Security note (PR classified security-sensitive)

This phase contains no auth/read path and produces nothing to an external system. The security-relevant work (read-path authorization, fail-CLOSED visibility filtering) lands in a later phase and is explicitly out of scope here. The migration is additive (new table only) with a clean `Down`. Nothing in this diff requires manual security sign-off beyond the standard items above. The `secret-scan` CI failure is a false positive: gitleaks flagged the `token` lock-owner variable name in `etl_lock.go` and/or go.sum module checksums — neither is a credential; no real secret is present in the diff.

## Recommendations

- (P2-1) Optionally align with `opanalytics` lazy construction in a later phase.
- (P2-2) Confirm phases 2/3 consume the lock/cursor scaffolding, or trim it.
- Consider a follow-up to allowlist the `token` variable in the gitleaks config so this module stops tripping `secret-scan` on every future phase.

PR #368, opened 2026-06-14
feat(searchetl): Kafka producer + transaction-split + payload extraction

yujiawei:
## Summary

Stage 2 of the message-search ETL → Kafka → ES indexer pipeline: wire the `searchetl` ETL to a Kafka producer for topic `octo.message.v1`, behind the `Kafka.On` feature flag (default **off** — no producer constructed, no connection, fully lazy; deploying to test env introduces zero runtime behavior change until the flag is explicitly enabled).

Builds on the Stage 1 scaffold (cursor table + stability gate + lazy dry-run).

## What's in this PR

- **Contract import**: `searchetl` now imports the shared `contract/searchmsg` package; the producer serializes that struct with `schema_version`, using `message_id` as the Kafka key (same message → same partition → ES `_id` upsert idempotency).
- **Transaction-split**: short read txn (`FOR UPDATE` cursor + keyset batch, commit immediately, **no Kafka IO under the lock**) → out-of-txn whole-batch produce → short advance txn. Whole-chunk atomic: any produce failure leaves the cursor unadvanced so the full chunk is re-produced next run (idempotent via `message_id` + ES `_id` upsert).
- **Stability gate preserved**: the cursor only advances to the stable-prefix end (`created_unix <= DB_NOW - lag`), never the batch end.
- **Payload extraction (three-state)**: Signal-encrypted → `raw_excluded` (content=null, **not DLQ**); non-text / object content → `raw_excluded`; genuine parse failure of a non-Signal message → DLQ topic; cursor advances in all three cases. Type coercion handles `float64`/`int`/`json.Number`.
- **In-process reentrancy guard**: an atomic CAS short-circuits a concurrent same-process run before any producer construction or DB read. Cross-replica mutex remains deferred to Stage 3 (Redis lock held throughout + abort on renew failure).
- **Semantic accuracy**: effectively-once comes from the at-least-once pipeline + ES `_id` idempotent upsert sink — **not** an opanalytics-equivalent same-transaction accumulator (documented in code; no claim of equivalence).

## Tests (verification gates)

- Signal message → `raw_excluded`, not DLQ, cursor advances
- Genuine parse failure → DLQ, cursor advances
- Batch produce failure → no advance → full re-produce
- No Kafka produce while the DB read txn (FOR UPDATE) is open
- Ordering: no produce before read; empty/head-unstable reads produce nothing
- Feature-flag-off and reentrancy-guard laziness
- Stable-prefix routing + maxID selection

`go build ./...`, `go vet`, and `go test -race ./modules/searchetl/...` all green. Independent model review: PASS.

## Notes

`segmentio/kafka-go` added as a direct dependency (chosen in the architecture decision). No runtime change until `Kafka.On=true`.


Jerry-Xin (approved):
The PR is in scope for `octo-server`; it extends the existing `modules/searchetl` message-search ETL path. I found no blocking correctness or security issues.

💬 Non-blocking

- 🟡 Warning: [modules/searchetl/producer.go:45](modules/searchetl/producer.go:45) enables `AllowAutoTopicCreation`. This is useful locally, but in production a topic typo can silently create a topic with broker-default retention/replication and split the pipeline. Prefer making this config-gated or disabling it outside dev/test.

- 🟡 Warning: [modules/searchetl/etl.go:84](modules/searchetl/etl.go:84) / [modules/searchetl/producer.go:33](modules/searchetl/producer.go:33) do not fail fast when `Kafka.On=true` but brokers or topics are empty. `kafka.TCP()` returns a nil address for an empty broker list, so the run can read DB state and only fail on the first produce attempt. A small config validation before constructing the producer would make rollout failures clearer.

- 🔵 Suggestion: [modules/searchetl/chunk_test.go:122](modules/searchetl/chunk_test.go:122) checks the no-Kafka-under-lock invariant with a fake that always clears `txnOpen` before returning. The production structure is correct, but an event-sequence assertion (`read begin`, `read commit`, `produce main`, `produce dlq`, `advance`) would make the ordering regression test stronger.

✅ Highlights

- The PR correctly preserves the stable-prefix cursor rule and advances to the last stable row rather than the batch tail.
- The transaction split avoids Kafka I/O while holding the cursor row lock.
- Produce failure semantics are conservative: failures leave the cursor unadvanced and rely on `message_id` idempotency.
- Payload extraction cleanly separates indexed text, known raw exclusions, and genuine DLQ cases.
- Focused tests cover the main cursor, routing, laziness, and reentrancy behavior.

Verification run: `go test ./modules/searchetl/...` and `go test -race ./modules/searchetl/...` both passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#368 Review Report
**PR:** feat(searchetl): Kafka producer + transaction-split + payload extraction
**Head SHA:** `c1a97e6db1d14befe01ca7714e287c1d3ba03d46`
**Scope:** 11 files, +907/−30 (new: chunk.go, chunk_test.go, etl_test.go, payload.go, payload_test.go, producer.go; modified: etl.go, etl_db.go, model.go, go.mod, go.sum)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Transaction split (C2): no Kafka IO under FOR UPDATE lock | ✅ | `readStableBatchTx` commits before return (`etl_db.go:101-120`); `runChunk` calls `produceBatch`/`produceDLQ` after `readStableBatchTx` returns (`etl.go:runChunk`); test `TestRunChunk_NoKafkaIOUnderLock` asserts via `fakeStore.txnOpen` flag |
| Stability gate (C1): cursor only advances to stable-prefix end | ✅ | `planChunk` sets `maxID = stable[len(stable)-1].ID` after `stablePrefix` truncation (`chunk.go:39`); `advanceCursor` uses `plan.maxID` (`etl.go:runChunk`) |
| Payload three-state extraction (P1-d) | ✅ | `extractMessage` (`payload.go:47-97`): Signal→raw_excluded, bad-JSON→DLQ, non-text→raw_excluded, text→content; all paths tested in `payload_test.go` |
| Reentrancy guard | ✅ | `atomic.Bool` CAS in `RunIncremental` (`etl.go:72-76`); test `TestRunIncremental_ReentrancyGuard` |
| Feature-flag laziness | ✅ | `Kafka.On=false` short-circuits before producer construction (`etl.go:67-70`); test `TestRunIncremental_KafkaOffIsLazy` |
| At-least-once + idempotent sink | ✅ | `produceBatch` failure → no `advanceCursor` call (`etl.go:runChunk`); `message_id` as Kafka key → same partition → ES `_id` upsert idempotency |
| Optimistic CAS on cursor advance | ✅ | `advanceCursor` uses `WHERE last_id=expected` (`etl_db.go:133`); returns `false` if cursor moved by another writer |
| `RollbackUnlessCommitted` safety | ✅ | `dbr.Tx.RollbackUnlessCommitted` ignores `sql.ErrTxDone` (verified in `gocraft/dbr/v2@v2.7.7/transaction.go`) |
| `config.SettingFromUint8` bit extraction | ✅ | Signal bit at position 5 (`octo-lib/config/msg.go:935`); `isSignalEncrypted` checks both `Signal` column and `Setting` bit (`payload.go:101-105`) |
| `common.Text` type comparison | ✅ | `common.Text = ContentType(1)` (`octo-lib/common/msg.go`); `payloadType` returns `contentType == common.Text.Int()` (`payload.go:113`) |

## 2. Findings

### P2-1: `totalMain`/`totalDLQ` log counters may overcount when cursor CAS fails

**Diff-scope:** new (introduced by this PR's `RunIncremental` loop).

In `RunIncremental` (`etl.go:97-98`), `totalMain += int64(len(plan.main))` accumulates **before** checking whether `advanceCursor` succeeded. When `advanceCursor` returns `advanced=false` (cursor moved by another writer), messages were already produced to Kafka but the cursor didn't advance. The logged `main_produced`/`dlq_produced` counts will include these messages, which will be re-produced on the next run.

**Impact:** Observability only — operators may see inflated production counts vs. actual cursor progress. No data loss or corruption (ES `_id` upsert deduplicates).

**Recommendation:** Consider accumulating totals only after confirmed cursor advance, or add a separate `totalRedelivered` counter for CAS-miss events. Low priority.

### P2-2: `_ = common.Text` in test is a code-quality nit

**Diff-scope:** new.

`TestRunChunk_SignalNotInDLQ` (`chunk_test.go:232`) ends with `_ = common.Text` — a blank-identifier reference to suppress the unused-import error for `common`. This is a code smell; the test doesn't actually use `common.Text` in any assertion.

**Recommendation:** Remove the `common` import from `chunk_test.go` if not needed, or add a meaningful assertion (e.g., verify `plan.main[0].ContentType`).

### P2-3: No test coverage for `advanceCursor` returning `advanced=false`

**Diff-scope:** new.

The `runChunk` path where `advanceCursor` returns `(false, nil)` (cursor moved by another writer) is not exercised by any test. The `fakeStore.advanceCursor` always returns `true` when `expected == cursor`. While this is a "shouldn't happen in practice" path (C3 Redis lock will prevent it), the code handles it with a WARN log and returns `n=0` — a test would strengthen confidence in this graceful degradation.

## 3. Suggestions

1. **Observability:** Add a structured log or metric when `advanceCursor` CAS misses (`advanced=false`), distinguishable from the "reached unstable tail" stop condition.
2. **Future (Stage 3):** The `advanceCursor` function uses `d.session.UpdateBySql` without an explicit transaction. While single-statement autocommit is safe in MySQL/InnoDB, consider wrapping in an explicit transaction for documentation clarity when the Redis lock (C3) is added.
3. **Dependency bumps:** `golang.org/x/crypto v0.19.0→v0.36.0`, `golang.org/x/net v0.17.0→v0.38.0`, `golang.org/x/sys v0.17.0→v0.31.0` are significant jumps. Verify no breaking API changes affect existing code paths (these are indirect deps, so risk is low).

## 4. Extra Findings

- **`AllowAutoTopicCreation: true`** in `newProducer` (`producer.go:52`): acceptable for local/staging, but production deployments should pre-create topics with proper retention/replication. Document this in deployment guides.
- **`WriteTimeout: 10 * time.Second`** (`producer.go:51`): if a Kafka broker is slow, the 10s timeout per write could cause the ETL run to stall. Combined with the per-chunk loop, a slow broker could delay the entire run. Consider adding a per-run context timeout in Stage 3.

## 5. Data-Flow Trace

|---|---|---|
| `row.Setting` (uint8) | SQL `SELECT setting FROM message_table` → `srcMessageRow.Setting` | ✅ `config.SettingFromUint8` extracts bit 5 (Signal) correctly |
| `row.Signal` (int) | SQL `SELECT signal FROM message_table` → `srcMessageRow.Signal` | ✅ Non-zero check in `isSignalEncrypted` covers both paths |
| `row.Payload` ([]byte) | SQL `SELECT payload FROM message_table` → `srcMessageRow.Payload` | ✅ `json.Unmarshal` in `extractMessage`; Signal check gates before parse |
| `row.MessageID` (string) | SQL `SELECT message_id` → Kafka key + `searchmsg.Message.MessageID` | ✅ Used as partition key and ES `_id` for idempotency |
| `row.CreatedUnix` (int64) | SQL `UNIX_TIMESTAMP(created_at)` → stability gate cutoff | ✅ `stablePrefix` truncates at `CreatedUnix > cutoff` |
| `cursor` (int64) | `SELECT last_id FROM octo_etl_es_cursor FOR UPDATE` → `readStableBatchTx` return | ✅ CAS in `advanceCursor` validates `last_id=expected` before update |
| `searchmsg.Message.Content` (*string) | `extractMessage` sets from `m["content"].(string)` | ✅ Nil for raw_excluded; pointer to string copy for OK path |
| `cutoff` (int64) | `dbNowUnix() - lag` computed once in `RunIncremental` | ✅ Fixed for entire run; conservative (safe) as run progresses |

## 6. R5 Blindspot Checklist

- **C1 — Dual-path parity:** N/A. No add/remove or create/delete symmetry in this PR. Single ETL read→produce→advance flow.
- **C2 — Control-flow ordering / nesting reuse:** Clear. `runChunk` enforces strict read→plan→produce→advance ordering. `extractMessage` checks Signal before JSON parse (correct ordering). `planChunk` is pure (no IO, no nesting risk). `produceBatch`/`produceDLQ` are called sequentially, not nested.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. Internal ETL pipeline, no user-facing endpoints. Kafka access gated by `Kafka.On` feature flag (default off). No new tool/capability exposure.

---

**[Octo-Q] verdict: APPROVE**

Well-structured Stage 2 implementation. Transaction split correctly separates DB locks from Kafka IO. Payload extraction three-state logic is clean and well-tested. At-least-once delivery with idempotent ES sink is sound. Feature-flag laziness ensures zero runtime impact until explicitly enabled. No P0/P1 issues found. Three P2 observations (log counter overcount, test code quality, missing CAS-miss test) are non-blocking.

lml2468 (approved):
✅ APPROVE(有强烈建议)— phase-2 的核心(事务拆分 at-least-once + 幂等 + 稳定闸 + payload 三态)逐条核实正确,无 phase-2 blocker。但有两条 phase-3「开自动跑」前的硬前提必须落实,其中 lag guard 强烈建议本 PR 顺手补。

## 核心正确性(本地逐条核实,这是 phase-2 必须对的部分)
1. ✅ **事务拆分 at-least-once**(chunk.go runChunk)— 严格 read-txn → 事务外 planChunk → 事务外 produceBatch+produceDLQ **全部确认成功** → 才 advanceCursor。任一投递失败 return error、游标不推进、下轮整批重投。crash 在投递与推进之间 = 重投(不丢)。
2. ✅ **游标只推进到稳定前缀末**(planChunk maxID = stable[last].ID,非 batch 末)— 保持 C1,raw_excluded/DLQ 行 id 也计入水位(否则下轮重扫)。
3. ✅ **幂等命门**:Kafka message Key = MessageID(producer.go)→ ES _id upsert 去重,at-least-once + 幂等 sink = effectively-once。
4. ✅ **payload 三态**(extractMessage:OK/RawExcluded→main,DLQ→dlq)分流正确;DLQ 投递失败也 return error 不静默丢。
5. ✅ producer.Close 保留首错(firstErr==nil 才赋值是标准写法,两 writer 都关)。build/vet/test 全绿。

关键:RunIncremental **当前无任何自动 tick / 生产调用方**(1module.go:16 明确 Start/Stop 钩子 phase-3 挂;grep 全仓零生产调用)。故下列两条的真实危害要到 phase-3 scheduler 接上才激活——本 PR 合入不会自动触发。

## 🔴 phase-3「开自动跑」前的硬前提(必须落实,否则 phase-3 一开就出事)
1. **lag=0/负值生产无强制**(config.go lagSeconds:负值→0,无运行时 guard)。real-advancement 路径已存在,lag=0 时 cutoff=now → 可能推过低 id 晚提交行 → **永久静默丢消息**。这条已是 #365 起第三轮被点。**强烈建议本 PR 就补 startup guard**(生产环境 lag<=0 → panic/fatal,几行成本),不要再拖到 phase-3。
2. **Redis 锁(C3)未接入 RunIncremental**(etl.go 仅进程内 atomic.Bool 重入护栏;作者注释标 phase-3 由 scheduler 持锁+续租)。phase-3 挂自动 tick 前必须:run 前 Acquire、全程 Renew、退出 Release(defer)、续租失败 abort 在飞批次。另:#365 起就提的 etl_lock.go Renew/Release 的 Lua 返回值类型断言静默(n,ok:=res.(int64) 的 ok==false 被当锁丢失),建议接入前采 oidc/sync_lock.go 的严格写法(类型断言失败返 error)。

## should-fix(非阻塞)
- 指标在 cursor 推进失败时仍报已投计数(口径误导);RunIncremental 内层循环无 context 取消检查;DLQ topic 配置未校验存在性。

## 我的定级说明(诚实)
两条护栏我判 should-fix 而非 phase-2 blocker,依据是「real-advancement 无自动调用方、危害未激活」,与本模块 #365 phase-1「机制就位、开关留下阶段」同框架。但若按更严立场「真实推进代码已入库则护栏须同 PR 齐备」,lag guard 尤其该现在补——故我给 APPROVE 但把它列为强烈建议 + phase-3 硬前提。phase-3 PR 我会重点验这两条是否落实。

—— Octo-PR (lml2468)

yujiawei (commented):
# Code Review — PR #368 (octo-server)

**Scope:** `feat(searchetl): Kafka producer + transaction-split + payload extraction` — Stage 2 of the message-search ETL → Kafka → ES pipeline.
**Reviewed head SHA:** `c1a97e6db1d14befe01ca7714e287c1d3ba03d46` (against merge-base `81e85038`).

## Verdict

**APPROVED** for merge as gated scaffolding — with one required pre-enable fix (see below).

This change is doubly gated: `Kafka.On` defaults to `false` **and** `RunIncremental` is not wired to any scheduler yet (test-invoked only). I verified both. So merging it introduces **zero** runtime behavior change in the test or prod environments until Stage 3 wires the scheduler and the flag is turned on. The transaction-split design (short read txn → out-of-txn produce → optimistic advance), the three-state payload routing, and the at-least-once + ES `_id` idempotency story are all sound and well-documented.

I confirmed locally: `go build ./modules/searchetl/...`, `go vet`, and `go test -race ./modules/searchetl/...` all pass.

I traced the worst-case two-replica interleave and confirmed there is **no permanent-loss / skipped-range path**: both replicas can read the same cursor `X` and batch `[X+1..maxID]`, both produce (deduped at the ES `_id` upsert sink), and the optimistic `UPDATE ... WHERE last_id=expected` lets only one advance `X→maxID` succeed; the loser is a no-op. Because every `newID` derives from a batch that started at `expected=X`, the cursor can never jump past unread rows. The C2 ordering (no Kafka IO under the FOR UPDATE lock) is correct in production code.

## Required before enabling (must fix before Stage 3 turns this path on)

### [Blocker-on-enable] A single >1 MB message permanently stalls an entire shard

`producer.go` builds the `kafka.Writer` without setting `BatchBytes`, so kafka-go (v0.4.51) uses its default `1048576` (1 MB). In `WriteMessages` (kafka-go `writer.go:626-636`), there is a pre-flight loop that returns `MessageTooLargeError` for the **whole call** — before sending anything — if any single message's `totalSize()` exceeds `batchBytes`:

```go
for i := range msgs {
    n := int64(msgs[i].totalSize())
    if n > batchBytes {
        return messageTooLarge(msgs, i)   // whole WriteMessages call fails, nothing sent
    }
}
```

`produceBatch` returns that error → `runChunk` returns error → `RunIncremental` returns error → `advanceCursor` is never reached → the cursor stays put → the next run reads the **same** batch (`id > cursor`), hits the **same** oversized row, and fails identically. This is a poison pill: the shard's cursor never advances past the chunk, so **every subsequent message in that shard is never indexed** — not just the oversized one.

This is reachable in practice: octo-server already treats >1 MB payloads as a real case (`modules/message/api.go:54` `hardParsePayloadLimit = 1 * 1024 * 1024`; RichText payloads are capped at 1 MB in octo-lib). A single large message would halt indexing for the whole shard.

**Suggested fixes (pick one or combine):**
- Set an explicit `BatchBytes` aligned with the broker's `message.max.bytes`, and pre-filter oversized rows into the DLQ flow (or `raw_excluded`) so one big message can never block the chunk.
- Detect `MessageTooLargeError` (or a marshaled size > limit) at extraction/produce time and route that single `message_id` to DLQ instead of failing the whole chunk.

The same structural trap exists for a `json.Marshal` failure of a contract struct (would also fail the whole chunk forever), though that is far less likely in practice than the size cap.

## Non-blocking findings (P2)

These do not block merge of the gated scaffolding; most are worth a quick follow-up.

1. **`runChunk` "no Kafka IO under lock" test is effectively vacuous** (`chunk_test.go:66-76,101-135`). `fakeStore.readStableBatchTx` sets `txnOpen = true` then immediately `txnOpen = false` *inside the same function body* before returning, and the sink is only ever called *after* it returns — so `sink.sawTxnOpen` can never become true and the `if sink.sawTxnOpen` assertion can never fire. The test would still pass even if `runChunk` produced to Kafka while holding the lock. The production ordering is actually correct, but the test guarding C2's hardest invariant enforces nothing. Recommend a fake that keeps `txnOpen` true across the produce window (e.g. an explicit commit hook `runChunk` must call, or a probe callback that fails if the sink runs while open).

2. **`produceBatch` doc-comment overstates Kafka atomicity** (`producer.go:57-58`, struct doc `17-23`). With `defaultBatch=5000` and no `BatchSize`/`BatchBytes` override, one `WriteMessages` call is split into many independent per-partition batches; on a transient error some are already committed (kafka-go returns a `WriteErrors` map after the fact). The phrase "整 chunk 原子语义" is misleading — the system stays correct only via the ES `_id` idempotent upsert, not via Kafka-level atomicity. Re-word to make the at-least-once + sink-idempotency dependency explicit.

3. **Inconsistent C3 comment** (`etl.go:159-160`). The parenthetical "理论上 C3 锁保证不会发生" reads present-tense, but C3 (the Redis lock in `etl_lock.go`) is defined yet never acquired anywhere in `RunIncremental` (verified: unreferenced outside tests). Every other comment in the module correctly marks C3 as a Stage-3 deliverable; this one line contradicts them. Today the optimistic `last_id` guard is what makes concurrent advances loss-safe. Recommend wording it that way and keeping the optimistic guard permanently (defense in depth) even after C3 lands.

4. **DLQ flow has no sink-side dedup** (`etl.go:147-152`, `producer.go:86-105`). The whole-chunk re-produce strategy is idempotent for the main flow (ES `_id` upsert) but not for DLQ: if main succeeds and DLQ fails (or partially fails), the chunk re-runs and the same DLQ rows are appended again. Harmless in Stage 2 (no DLQ consumer yet), but document the DLQ as at-least-once with no dedup so the eventual consumer keys/upserts by `message_id`.

5. **Metrics over-count on `advanced=false`** (`etl.go:103-113,159-164`). When produce succeeds but `advanceCursor` returns `advanced=false`, `runChunk` returns `(plan, 0, nil)` and `RunIncremental` still adds `len(plan.main)/len(plan.dlq)` to the totals; the same physical chunk is then re-counted next run. Observability only. Cleanest fix: have `runChunk` return the actually-committed counts (0 when `advanced=false`).

6. **`ctx` cancellation not honored by the DB layer or the chunk loop** (`etl.go:98-114`, `etl_db.go`). `ctx` is plumbed only to `WriteMessages`; all DB calls use the bare session and the loop has no `ctx.Err()` check, so a cancelled run keeps issuing DB reads until the next produce. Fine for Stage 2, but Stage 3's "续租失败 abort" lease guarantee will depend on prompt cancellation — add a per-iteration `ctx.Err()` check and consider context-scoped DB calls.

7. **Reentrancy guard depends on `*ETL` being a process singleton** (`etl.go:47,78`). Correct today (`1module.go` constructs exactly one). Worth a one-line comment that the Stage-3 scheduler must reuse the singleton and never `NewETL()` per tick, or the in-process guard becomes a no-op.

8. **Missing failure-path tests** (`etl.go:150-164`, loop `102-113`): no coverage for `produceDLQ` failure (`fakeSink` has `failMainOn` but no `failDLQOn`), the `advanced=false` warn branch, or the multi-chunk `RunIncremental` loop (`n < batch` break + cursor relay). The DLQ-failure path is exactly the data-loss guard the PR advertises, so it deserves a test.

9. **Dead test scaffolding** (`chunk_test.go:60,91`): `fakeSink.mainMsgs` is captured but never asserted (no test verifies produced key=`message_id`/content at the `runChunk` layer), and `fakeStore.readErr` is never set. Either assert or remove.

10. **`payloadType` has unreachable switch arms** (`payload.go:102-113`): plain `json.Unmarshal` into `map[string]interface{}` always yields `float64`, so the `case int` / `case json.Number` arms are dead unless a `UseNumber` decoder is introduced. Intentional parity with `message.isTextType` and harmless — just confirm intent.

## Notes on things that are fine

- The dependency bumps (`x/crypto 0.19→0.36`, `x/net 0.17→0.38`, `x/sys`, `xdg-go/scram`/`stringprep`, new `pierrec/lz4` indirect) are MVS-forced transitive minimums pulled up by `segmentio/kafka-go`, not scope creep. `go mod verify` passes.
- `ContentType=0` on the Signal/DLQ paths is acceptable — routing is by `RawExcluded`, and the contract does not require a `ContentType` sentinel.
- Advancing the cursor over `raw_excluded` and DLQ rows is correct (those rows are consumed; their ids must be in the watermark).

lml2468:
📌 补充 phase-3 硬前提（响应作者 yujiawei 自评发现的大消息阻塞，我已核实成立）

我的票维持 APPROVE（phase-2 核心 at-least-once 正确、危害未激活），但作者这条是真问题，加入 #368 的 phase-3「开自动跑」前硬前提清单。

## 🔴 新增硬前提：超限消息会卡死整个分片增量（毒丸 head-of-line 阻塞）
核实（producer.go newProducer）：kafka.Writer 未设 BatchBytes（kafka-go 默认 ~1MB），同步 + RequireAll + WriteTimeout=10s。一条 payload > 限额的消息：
- produceBatch 的 WriteMessages 返回 MessageSizeTooLarge 错误（非无限挂，WriteTimeout 兜底）；
- → runChunk 返回 error、该分片游标不推进；
- → 下轮整批重投 → 再次撞上同一条大消息 → 再次失败。

结果不是进程挂死，而是**该分片游标永久卡在这条大消息前**：这条之后的所有消息永远投不出去、永久搜不到（分片级 head-of-line 阻塞）。phase-2 现在无自动 tick 不触发，但 phase-3 接 scheduler 后必现。

修法（作者已给方向，认同）：
- planChunk/extractMessage 阶段对超限 payload 预判 → 路由 DLQ（带 message_id 供排查），不进 main 流；
- 或 producer 显式设 BatchBytes 并对单条超限消息单独处理（跳过+DLQ+告警），绝不能让一条毒丸卡死整片游标；
- 与已有 DLQ 三态机制天然契合。

## #368 phase-3 开自动跑前硬前提汇总（我 + 作者 + Steve）
1. 🔴 lag=0/负值生产 startup guard（强烈建议本 PR 就补）
2. 🔴 Redis 锁接入 RunIncremental（Acquire/Renew/Release + 续租失败 abort）+ Renew/Release Lua 类型断言严格化（采 oidc 写法）
3. 🔴 超限消息毒丸阻塞 → 预过滤/DLQ 逃生阀（本条，作者自查）
4. 🟡 生产禁 auto-create topic、无 broker/topic fail-fast（Steve）
5. 🟡 指标 CAS 失败过计数、内循环 context 取消、advanceCursor 返回 false 测试覆盖、DLQ topic 校验

phase-2 仍 APPROVE；上述为 phase-3 PR 的验收清单，我会逐条核。

—— Octo-PR (lml2468)

PR #369, opened 2026-06-14
feat(searchetl): ordered commit + cross-replica run-lock (C3)

yujiawei:
## Summary

Stage 3 of the message-search ETL → Kafka → ES indexer pipeline: make the producer tick safe under multiple replicas, completing the reliability story started by the Stage 2 transaction split. Builds on `main` (includes Stage 2 #368).

Feature flag unchanged: `Kafka.On` defaults **off** — no producer/lock constructed, no runtime change until explicitly enabled.

## What's in this PR

**Ordered commit** — the cursor only advances to the end of the confirmed contiguous stable prefix. A mid-batch produce failure leaves the cursor unmoved (whole-chunk re-produce, no holes); produce always precedes advance, so the cursor never leads un-produced messages.

**Cross-replica mutex (C3)** — the whole `RunIncremental` tick now runs under the existing Redis run-lock (`searchetl:etl:run`, SET NX EX + Lua CAS), held across read-batch → produce → advance-cursor, renewed in a background goroutine:
- **Abort on lock loss**: renew failure / ownership loss cancels the lock context; the in-flight tick stops — no further produce, no advance.
- **Acquire-miss / Redis error** skips the tick instead of running unlocked, so two replicas never both run.
- **Renew goroutine is joined before release** (no late `Renew` after `Release`, no goroutine leak).

**The real safety invariant** (documented in `runChunk`): correctness does not rest on the Redis lock alone. It rests on *produce-before-advance ordering* + the *cursor-row CAS* (`advanceCursor`'s `WHERE last_id=expected`, serialized by `FOR UPDATE`) acting as a fencing token + ES `_id` idempotent upsert. Even under a lock-loss race, a stale owner's CAS misses and cannot advance; overlapping produces are deduped by the idempotent sink. No interleaving yields loss, a cursor hole, or double-count. The `ctx.Err()` check before advance is a best-effort early-exit, not the guarantee.

The in-process reentrancy guard (Stage 2) is retained as the first line of defence.

## Tests (verification gates)

Lock orchestration extracted behind a `runLock` interface so C3 is unit-tested without a live Redis:
- Cursor only advances to the confirmed contiguous prefix; partial failure → no advance (no holes)
- Lock loss (renew fail / owner lost) → in-flight tick aborts, cursor not advanced
- CAS miss (cursor moved by another replica) → no advance, no error (fence backstop)
- Two replicas contend while the first holds the lock → exactly one runs
- Acquire-miss / Redis error → skip tick, don't run unlocked
- No `Renew` after `Release` (renew goroutine joined)

`go build ./...`, `go vet`, and `go test -race ./modules/searchetl/...` all green; timing-sensitive tests stable across 20× `-race` runs. Independent model review: PASS (after addressing join ordering and hardening the lock-loss / CAS-fence coverage).

## Notes

Next stage (Stage 4): standalone `es-indexer` binary (consumer C4 + bulk + mapping). The write path will be organised so the bulk writer is reusable by the Stage 6 backfill job.

### DIFF
--- a/modules/searchetl/chunk_test.go
+++ b/modules/searchetl/chunk_test.go
@@ -20,10 +20,10 @@ func textRow(id int64, mid, content string, createdUnix int64) *srcMessageRow {
 // 游标 maxID 取稳定前缀末（含 raw_excluded/dlq 行的 id，必须计入水位）。
 func TestPlanChunk_StablePrefixAndRouting(t *testing.T) {
 	rows := []*srcMessageRow{
-		textRow(1, "a", "hi", 100),                                              // OK → main
+		textRow(1, "a", "hi", 100), // OK → main
 		{ID: 2, MessageID: "b", Signal: 1, CreatedUnix: 110, Payload: []byte("x")}, // signal → main(raw_excluded)
 		{ID: 3, MessageID: "c", CreatedUnix: 120, Payload: []byte("{bad")},         // bad json → dlq
-		textRow(4, "d", "late", 250),                                            // 未稳定 → 截断
+		textRow(4, "d", "late", 250),                                               // 未稳定 → 截断
 	}
 	plan := planChunk(rows, 200)
 	if plan.stableCount != 3 {
@@ -229,4 +229,77 @@ func TestRunChunk_SignalNotInDLQ(t *testing.T) {
 		t.Fatalf("cursor must advance past signal msg, got %d", store.advancedTo)
 	}
 	_ = common.Text
+}
+
+// TestRunChunk_LockLostAfterProduceNoAdvance 🔴 C3：投递成功后、推进游标前锁丢失（lockCtx 取消）
+// → 不推进游标（advanceCursor 零调用），返回 ctx 错误。已投消息靠 message_id 幂等 + ES _id
+// upsert 下轮重投覆盖；绝不在失锁后推进游标（失锁不双算）。
+func TestRunChunk_LockLostAfterProduceNoAdvance(t *testing.T) {
+	store := &fakeStore{cursor: 0, rows: []*srcMessageRow{textRow(1, "a", "x", 100)}}
+	// sink 在投递成功后取消 ctx，模拟「投递与推进之间续租失败」。
+	ctx, cancel := context.WithCancel(context.Background())
+	sink := &cancelSink{cancel: cancel}
+	_, n, err := runChunk(ctx, store, sink, "message", 200, 5000, lg())
+	if err == nil {
+		t.Fatalf("expected ctx error after lock loss")
+	}
+	if n != 0 {
+		t.Fatalf("lock-lost chunk must report 0 progress, got %d", n)
+	}
+	if store.advanceCalls != 0 {
+		t.Fatalf("advanceCursor must NOT be called after lock loss, got %d", store.advanceCalls)
+	}
+	if store.cursor != 0 {
+		t.Fatalf("cursor must stay at 0 (re-produce next tick), got %d", store.cursor)
+	}
+	if sink.mainCalls != 1 {
+		t.Fatalf("produce should have happened once before loss, got %d", sink.mainCalls)
+	}
+}
+
+// cancelSink 在 produceBatch 成功后取消传入的 ctx，模拟投递与游标推进之间的失锁。
+type cancelSink struct {
+	cancel    context.CancelFunc
+	mainCalls int
+}
+
+func (s *cancelSink) produceBatch(ctx context.Context, msgs []searchmsg.Message) error {
+	s.mainCalls++
+	s.cancel() // 投递成功后立即失锁
+	return nil
+}
+
+func (s *cancelSink) produceDLQ(ctx context.Context, msgs []searchmsg.Message) error {
+	return nil
+}
+
+// TestRunChunk_CASMissNoAdvance 🔴 C3 fence：投递成功后游标已被另一持锁副本推进（CAS 失配，
+// advanceCursor 返回 false）→ 本轮不计已投、不报错，下轮从新水位续跑。这是失锁竞态的安全收口：
+// 即便本 owner 失锁后仍跑到 advanceCursor，游标行 CAS（WHERE last_id=expected）也挡住越权推进。
+func TestRunChunk_CASMissNoAdvance(t *testing.T) {
+	store := &casMissStore{cursor: 0, rows: []*srcMessageRow{textRow(1, "a", "x", 100)}}
+	sink := &fakeSink{}
+	_, n, err := runChunk(context.Background(), store, sink, "message", 200, 5000, lg())
+	if err != nil {
+		t.Fatalf("CAS miss must not error, got %v", err)
+	}
+	if n != 0 {
+		t.Fatalf("CAS-miss chunk must report 0 progress, got %d", n)
+	}
+	if sink.mainCalls != 1 {
+		t.Fatalf("produce should have happened once, got %d", sink.mainCalls)
+	}
+}
+
+// casMissStore 的 advanceCursor 恒返回 (false,nil)，模拟「游标已被另一持锁副本推进」。
+type casMissStore struct {
+	cursor int64
+	rows   []*srcMessageRow
+}
+
+func (s *casMissStore) readStableBatchTx(table string, batch int) (int64, []*srcMessageRow, error) {
+	return s.cursor, s.rows, nil
+}
+func (s *casMissStore) advanceCursor(table string, expected, newID int64) (bool, error) {
+	return false, nil // CAS 失配：另一副本已推进游标
 }
--- a/modules/searchetl/etl.go
+++ b/modules/searchetl/etl.go
@@ -3,6 +3,7 @@ package searchetl
 import (
 	"context"
 	"sync/atomic"
+	"time"
 
 	"github.com/Mininglamp-OSS/octo-lib/config"
 	"github.com/Mininglamp-OSS/octo-lib/contract/searchmsg"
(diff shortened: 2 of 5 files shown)


mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#369 Review Report
**PR:** Mininglamp-OSS/octo-server #369
**Head SHA:** `c8a10fcfcaba0bf833e63be6754e3e12a1df4138`
**Title:** feat(searchetl): ordered commit + cross-replica run-lock (C3)
**Files:** 5 changed (+556 / −26)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Lock acquisition before any DB/Kafka work | ✅ | `etl.go:97-104` — `runLockedTick` called after CAS guard, `Acquire` is first operation |
| Lock spans entire tick (read→produce→advance) | ✅ | `etl.go:104` — `runTick` receives `lockCtx`; `etl_run_lock.go:82-92` — lock held across `run(lockCtx)` |
| Renew failure → abort in-flight tick | ✅ | `etl_run_lock.go:107-117` — `cancel()` on renew err/false; `etl.go:134-138` + `etl.go:200-205` — ctx.Err() checks |
| No Renew after Release (goroutine join) | ✅ | `etl_run_lock.go:87-91` — `close(done)` + `<-renewDone` before `Release`; tested by `TestRunLockedTick_NoRenewAfterRelease` |
| Cursor CAS fence (WHERE last_id=expected) | ✅ | `etl_db.go:143-152` — `UPDATE ... WHERE shard_table=? AND last_id=?`; returns `n==1` |
| Produce-before-advance ordering | ✅ | `etl.go:175-183` — `produceBatch` + `produceDLQ` complete before `advanceCursor` at line 210 |
| Acquire-miss / Redis error → skip tick | ✅ | `etl_run_lock.go:68-76` — miss returns nil, error propagates; tested by `TestRunLockedTick_AcquireFailsSkips` / `_RedisErrorSkips` |
| Two replicas → exactly one runs | ✅ | `TestRunLockedTick_TwoReplicasOnlyOneRuns` — sharedLock with contention barrier |
| CAS miss → no advance, no error | ✅ | `etl.go:215-220` — returns `(plan, 0, nil)`; tested by `TestRunChunk_CASMissNoAdvance` |
| Lock loss after produce → no advance | ✅ | `etl.go:200-205` — ctx.Err() check; tested by `TestRunChunk_LockLostAfterProduceNoAdvance` |
| Feature flag lazy (Kafka.On=off → no-op) | ✅ | `etl.go:91-93` — early return before lock/producer construction |
| Defer ordering correct (join→release→cancel) | ✅ | `etl_run_lock.go:85-93` — inner defer runs first: close+join+release; outer `defer cancel()` is idempotent |

## 2. Findings

### P2-1: Per-tick Redis client creation (new)

`RunIncremental` calls `e.newLock(e.ctx)` → `productionNewLock` → `newETLLock(ctx)` which creates a **new `rd.Client`** (connection pool) every tick (`etl_lock.go:49-56`). The `Close()` defer in `RunIncremental` (`etl.go:99-103`) tears it down after each tick.

**Impact:** Functional correctness is fine (CAS-DEL/CAS-PEXPIRE are atomic regardless of pool age). Efficiency cost: ~3-5 TCP connections created/torn down per tick. Negligible at 60s tick interval, but would matter if tick frequency increases or during backfill (Stage 4).

**Suggestion:** Consider hoisting the `*etlLock` to `ETL` struct level (create once in `NewETL`, close in a lifecycle hook). Low urgency — behind feature flag, and current approach is correct.

### P2-2: `readStableBatchTx` does not respect context cancellation (pre-existing, amplified)

**Diff-scope:** pre-existing API, amplified by this PR (lock-loss abort now makes mid-read cancellation desirable)

`chunkStore.readStableBatchTx` (`etl.go:17`) does not accept a `context.Context`. If lockCtx is cancelled during the DB read, the read completes before the cancellation is detected at the next `ctx.Err()` check (`etl.go:200`).

**Impact:** At most one extra chunk read after lock loss (the read that was in-flight when cancel fired). The CAS fence at `advanceCursor` prevents incorrect cursor advance. No data loss or duplication. Minor latency impact only.

**Suggestion:** Thread `ctx` into `readStableBatchTx` and pass it to the dbr transaction for prompt cancellation. Non-blocking.

### Nit-1: Whitespace-only changes in `chunk_test.go`

Lines 23 and 26 of `chunk_test.go` have alignment-only whitespace changes. Cosmetic, no functional impact.

## 3. Suggestions

1. **Hoist Redis client to ETL struct** (addresses P2-1): Create `*etlLock` once in `NewETL` or lazily on first tick, reuse across ticks. Close on ETL shutdown.
2. **Thread ctx into readStableBatchTx** (addresses P2-2): Add `ctx context.Context` parameter to `chunkStore.readStableBatchTx` and pass to `tx.SelectBySql(...).WithContext(ctx)`.
3. Both are non-blocking — the safety invariants hold without them.

## 4. Extra Observations

- **Lua script correctness:** Both `luaReleaseETLLock` and `luaRenewETLLock` (`etl_lock.go:22-38`) correctly use GET+compare+DEL/PEXPIRE atomically. Token mismatch returns 0, which is correctly interpreted by the Go layer.
- **Error propagation chain:** `runChunk` error → `runTick` returns → `runLockedTick` returns → `RunIncremental` returns. The caller (future scheduler, Stage 4+) will see the error. Currently no caller exists since the feature is flag-gated and not yet scheduled.
- **`running` CAS vs lock ordering:** `running` is set before lock acquire, cleared after lock release (LIFO defers). This means another goroutine in the same process cannot attempt to acquire the Redis lock until the current holder fully releases. Correct — prevents thundering herd on the Redis lock.

## 5. Data-Flow Trace

|---|---|---|
| `lockCtx` (in runTick/runChunk) | `runLockedTick` → `context.WithCancel(ctx)` at `etl_run_lock.go:79` | ✅ Yes — passed as `ctx` to `run(lockCtx)`, checked via `ctx.Err()` at two points |
| `cursor` (in runChunk) | `readStableBatchTx` → `SELECT last_id ... FOR UPDATE` at `etl_db.go:117` | ✅ Yes — returned as first value, used in CAS at `advanceCursor` |
| `plan.maxID` (cursor advance target) | `planChunk` → `stable[len(stable)-1].ID` at `chunk.go:47` | ✅ Yes — last row of stable prefix, always valid when `plan.advanced==true` |
| `token` (lock owner identity) | `randomToken()` → `crypto/rand.Read(16 bytes)` at `etl_run_lock.go:40-44` | ✅ Yes — used in Acquire/Renew/Release; Lua scripts compare via GET |
| `cancel` (lock-loss signal) | `context.WithCancel` at `etl_run_lock.go:79`; called by `renewUntilDone` on renew failure | ✅ Yes — cancels `lockCtx`, detected by `ctx.Err()` checks in runTick and runChunk |
| `advanced` bool (CAS result) | `advanceCursor` → `UPDATE ... WHERE last_id=?` → `RowsAffected()==1` at `etl_db.go:143-152` | ✅ Yes — `false` triggers skip-without-error path; `true` confirms cursor moved |

## 6. R5 Blindspot Checklist (security_sensitive)

**C1 — Dual-path parity:** N/A. This PR adds lock orchestration (acquire/renew/release), not symmetric add/remove or create/delete paths. The lock lifecycle is linear: acquire → renew loop → release. No symmetric counterpart.

**C2 — Control-flow ordering / nesting reuse:** Clear.
- `renewUntilDone` has exactly one call-site (the goroutine in `runLockedTick`). No nesting/double-invocation risk.
- Lua scripts are atomic Redis operations. Tested with token mismatch (returns 0) and key-missing (redis.Nil → treated as non-error in Release, as ownership-lost in Renew).
- `cancel()` is idempotent — called by both the defer and the renew goroutine; no double-effect.

**C3 — Authorization boundary ≠ capability boundary:** N/A. No user-facing endpoints or tool exposure. This is internal infrastructure (ETL pipeline). The Redis lock key (`searchetl:etl:run`) is not exposed to external callers.

---

## [Octo-Q] verdict: APPROVE

No P0 or P1 findings. The design is sound with defense-in-depth: Redis lock (primary mutex) → cursor CAS (fencing token) → ES _id upsert (idempotent sink). All critical paths are covered by tests, including timing-sensitive lock-loss and CAS-miss scenarios. Two P2 efficiency observations are non-blocking.

Jerry-Xin (approved):
Relevance gate: pass. This PR is in scope for `octo-server`; it extends the `modules/searchetl` ETL pipeline with ordered cursor commit and cross-replica locking.

💬 Non-blocking

🟡 Warning: The fencing comments overstate the role of `FOR UPDATE`. In `modules/searchetl/etl.go:191` and `modules/searchetl/etl_db.go:88`, `readStableBatchTx` commits before Kafka produce and before `advanceCursor`, so `FOR UPDATE` does not serialize the final advance. The real fence is the atomic conditional `UPDATE ... WHERE shard_table=? AND last_id=?` at `modules/searchetl/etl_db.go:127`. The implementation is sound, but the comment could mislead future changes.

🔵 Suggestion: `runTick` only checks `ctx.Err()` inside the per-chunk loop at `modules/searchetl/etl.go:134`. A parent cancellation that exists before `runTick` starts can still construct the Kafka producer and hit `dbNowUnix`/`ensureCursor` first. This is not a correctness issue for the lock-loss path, but an early context check near `modules/searchetl/etl.go:113` would make cancellation behavior cleaner.

✅ Highlights

The runtime ordering is correct: produce happens before cursor advance, and partial produce failure leaves the cursor unchanged.

The Redis lock orchestration avoids running unlocked on acquire failure/miss, joins the renew goroutine before release, and uses token-checked release/renew.

The new tests cover the important C3 paths: acquire miss, Redis error, renew failure, owner loss, CAS miss, and no renew-after-release.

Verification run: `go test ./modules/searchetl/...` passed.

yujiawei (commented):
# Code Review — PR #369 (octo-server)

**Reviewing head SHA:** `c8a10fcfcaba0bf833e63be6754e3e12a1df4138` · base `main` · 5 files, +556/−26

## Summary

Stage 3 of the message-search ETL pipeline: it puts the whole `RunIncremental` tick under the existing Redis run-lock (held across read-batch → produce → advance-cursor, renewed in the background), aborts the in-flight tick on lock loss, and keeps the cursor on ordered commit (advance only to the confirmed contiguous stable-prefix end). The lock orchestration is extracted behind a `runLock` interface so the C3 semantics are unit-testable without a live Redis. `Kafka.On` still defaults **off**, so there is no runtime behavior change until the flag is enabled.

Verified locally on the head SHA: `go build ./...`, `go vet ./modules/searchetl/...`, and `go test -race -count=1 ./modules/searchetl/...` all green.

## Verification

| Item | Result | Evidence |
|---|---|---|
| Lock held across the whole tick (acquire before any DB/Kafka work) | ✅ | `etl.go:99-107` constructs the lock and enters `runLockedTick` before `runTick` touches the DB; `etl_test.go:51` proves acquire-miss short-circuits before `dbNowUnix`. |
| Abort in-flight tick on renew failure / owner loss | ✅ | `etl_run_lock.go:96-114` cancels `lockCtx`; `runTick` re-checks `ctx.Err()` per chunk (`etl.go:135-139`) and `runChunk` re-checks after produce (`etl.go:204-208`). Covered by `etl_run_lock_test.go:147,180`. |
| Acquire-miss / Redis error never runs unlocked | ✅ | `etl_run_lock.go:53-66` returns early on both; `etl_run_lock_test.go:56,75`. |
| Renew goroutine joined before Release (no late Renew, no leak) | ✅ | `etl_run_lock.go:79-86` closes `done`, waits `renewDone`, then releases; `etl_run_lock_test.go:110`. |
| Ordered commit — advance only to confirmed contiguous prefix | ✅ | `planChunk` caps `maxID` at the stable-prefix end (`chunk.go`); produce precedes advance (`etl.go:184-211`). |
| CAS fence backstops a lock-loss race | ✅ | `advanceCursor` `WHERE last_id=expected` (`etl_db.go:126-132`); `chunk_test.go` `TestRunChunk_CASMissNoAdvance`. |
| Two replicas contend → exactly one runs | ✅ | `etl_run_lock_test.go:202`. |
| Feature flag off by default → no runtime change | ✅ | `etl.go:85-89`; no scheduler wiring added (grep confirms `RunIncremental` has no production caller yet). |

### Correctness reasoning (the part worth scrutinizing)

I traced the replica-interleaving cases against the stated invariant and they hold:

- **No loss / no hole:** `advanceCursor(expected, maxID)` only succeeds when the cursor row still equals `expected`, and `maxID` is always the last id this chunk *produced* (produce strictly precedes advance). By induction the cursor never moves past an un-produced message, and the conditional UPDATE prevents non-contiguous jumps.
- **No double-advance under a stale owner:** if a replica loses the lock and another advances first, the stale owner's CAS misses and falls into the `!advanced` branch (`etl.go:215-221`) — no error, no advance.
- **Duplicates** from overlapping produces are at-least-once and rely on the downstream idempotent `_id=message_id` upsert (see P2-1).

The `ctx.Err()` checks are correctly framed in-code as a best-effort early exit, not the guarantee — the guarantee is the CAS + ordering. That framing is accurate.

## Findings

No P0/P1. The items below are P2 / suggestions and are **not** merge blockers.

**P2-1 — End-to-end "no double-count" depends on a consumer that doesn't exist yet.** The producer side is at-least-once by design; the "no double-count" property rests entirely on the es-indexer doing an idempotent `_id=message_id` upsert, which is Stage 4 (not in this PR). This is fine while `Kafka.On=off`, but the guarantee is not self-contained in this PR — worth tracking so Stage 4 doesn't silently break the invariant. (`etl.go:38-39`, `producer.go:18-20`)

**P2-2 — Doc nit on the fence mechanism.** Comments at `etl.go:37` and `etl.go:191-193` credit "`advanceCursor`'s `WHERE last_id=expected`, serialized by `readStableBatchTx`'s `FOR UPDATE`" as the fence. Precisely, the fencing guarantee is the standalone conditional `UPDATE` itself (atomic compare-and-set via its own row lock); the `FOR UPDATE` in `readStableBatchTx` is committed/released before produce, so it serializes concurrent cursor *reads* against an in-flight advance but does not span the advance. The behavior is correct — only the attribution is slightly misleading. Consider rewording to "the conditional UPDATE is the CAS; FOR UPDATE serializes concurrent reads of the cursor row."

**P2-3 — CAS-miss ends the shard for the rest of the tick.** On a CAS miss `runChunk` returns `n=0` (`etl.go:220`), so `runTick`'s loop treats it as "reached unstable tail" and breaks to the next shard rather than re-reading from the new cursor (`etl.go:147-149`). Correct (the next tick resumes), just slightly suboptimal. Since CAS misses only occur during rare lock-loss races, this is acceptable; a one-line comment noting the intentional "defer to next tick" would help future readers.

**P2-4 — Lock-loss detection latency vs TTL.** TTL is 30m and renew is TTL/3 (~10m). A lost lock can go undetected for up to ~one renew interval, during which TTL expiry could let a second replica acquire and both run briefly. This is exactly what the CAS fence + idempotency are designed to absorb, and it is documented — flagging only so it's a conscious tradeoff, not an oversight. (`etl_lock.go:32`, `etl_run_lock.go:25-31`)

## Security note (PR was flagged security-sensitive)

The `security_sensitive` classification here is keyword-triggered ("lock"/"token") rather than a real auth/credential surface. Reviewed accordingly and found no concerns: the lock "token" is a `crypto/rand` 128-bit value, not an auth credential; nothing sensitive is logged; the Lua release/renew scripts are correct CAS-DEL / CAS-PEXPIRE guarded by token match (no accidental deletion of a successor owner's lock); and the SQL uses parameterized values with table names drawn from a fixed config enumeration (no user-controlled input → no injection surface). No manual security follow-up required for this change.

## Verdict

**APPROVED.** Strong reliability work: the invariant is sound, the tests exercise the dangerous interleavings (lock loss after produce, CAS miss, two-replica contention, renew-join ordering) under `-race`, and the off-by-default flag means no production risk on merge. The P2 items are polish and a forward-looking dependency note, none blocking.

lml2468 (approved):
✅ APPROVE — phase-3 核心交付物（C3 跨副本 run-lock）做得非常扎实，逐行核实 + 7 个锁测试全绿。lag guard（P1）这次强烈要求落实(见下)。

## phase-3 核心（C3 锁）正确性 — 本地逐行核实通过
1. ✅ **锁全程接入 RunIncremental**（runLockedTick）：Acquire 原子（SetNX EX）；抢不到→跳过返回 nil 不裸跑；Redis 故障→不裸跑。
2. ✅ **续租 + 失败 abort**（renewUntilDone）：后台 TTL/3 续租；Renew 返 err 或 owner 丢失 → cancel(lockCtx)。runTick 每个 chunk 前 ctx.Err() 检查 → 失锁立即停、不再投递/推进游标（C3 不双算）。
3. ✅ **释放安全**：defer 释放（panic 安全）；**先 close(done)+join renewDone 再 Release** —— 杜绝「Release 后迟到 Renew 误删后继 owner 锁」+ 防 goroutine 泄漏，这个细节很专业。CAS-DEL owner-only、每轮唯一 token。
4. ✅ **ordered commit**：advanceCursor 的 WHERE last_id=expected 行级 CAS fence（FOR UPDATE 串行化），游标只推进到稳定前缀末。
5. ✅ **测试**：7 个新锁测试覆盖全部危险路径——acquire 竞争跳过 / Redis 故障跳过 / happy-path 释放 / release 后不再 renew / 续租失败 abort 在飞 / owner 丢失 abort / 两副本只一个跑。build/vet/test 全绿。这正是前两阶段反复要求的 lock ownership + abort 测试，全补齐。

## 🔴→强烈要求：P1 lag=0 生产 guard（第三次被点，必须收口）
config.go lagSeconds() 仍把负值→0、无生产 guard（注释 C1 标「生产 lag 不得为 0」但无代码强制）。本 PR 未挂 scheduler/auto-tick（RunIncremental 仍仅测试/手动调用），故危害尚未自动激活——这是我给 APPROVE 而非 CR 的依据。但此点自 #365、#368 已被点两次，是 2 行修复（NewETL 启动时生产环境 lag<=0 → panic/fatal）。**强烈要求本 PR 补；若留到紧接的 scheduler PR，则 scheduler PR 必须带此 guard，否则我会在那个 PR 直接 CR**——scheduler 一旦自动跑、lag=0 即永久静默丢消息，不能再拖。

## 其余（非阻塞，scheduler 上线前需收）
- 🟡 P3 Renew/Release 的 Lua 返回值类型断言静默（n,ok:=res.(int64); return ok&&n==1）。注：失败方向安全——断言失败返 (false,nil) → renewUntilDone 命中 !ok → cancel(abort)，是「多余 abort」而非「漏 abort 裸跑」，故非 blocking。但建议采 oidc/sync_lock.go 严格写法（断言失败显式返 error），区分「锁真过期」与「类型异常」便于排查。
- 🟡 P4 毒丸（超限 Kafka 消息）逃生阀仍未实现（producer 无 BatchBytes/预过滤→DLQ）。scheduler 自动跑前必须有，否则一条超限消息永久卡死分片游标。

## 裁决依据（诚实）
phase-3 该对的核心（锁正确性 + C3 互斥 + 续租失败 abort + 不双算 + 测试）全部达标，质量很高。开放项（P1/P3/P4）均为「scheduler 自动跑前的护栏」、当前无自动调用方故危害未激活——与本模块 #365/#368 同框架。故 APPROVE。但 P1 我升级为强烈要求 + scheduler PR 硬门槛。下一棒（scheduler）我会逐条验 P1/P3/P4。

—— Octo-PR (lml2468)

PR #361, opened 2026-06-12
feat(messages_search): in-conversation message search via OpenSearch

BrDing-Rookie:
## Summary

Adds four new endpoints under `/v1/messages` for in-conversation full-text / media / file search, backed by the OpenSearch index that `wukongim-message-indexer` writes to (mapping v1.8).

| Endpoint | Purpose |
|---|---|
| `POST /v1/messages/_search` | keyword search in a single channel |
| `POST /v1/messages/_search_media` | image / video hits in a single channel |
| `POST /v1/messages/_search_files` | file hits in a single channel |
| `POST /v1/messages/_search_all` | global cross-channel search |

The module is fully self-contained under `modules/messages_search/` plus the standard error-code / i18n locale entries.

### Authorization

A channel-access gate runs **before** every OS query (`authz.go`):

- **p2p** — always allowed (the OS channelId is derived from `(loginUID, peerUID)`, so the caller can only ever reach their own conversations regardless of `channel_id`).
- **group** — `loginUID` must be an *active* member (`ExistMemberActive`: `is_deleted=0` AND `status=Normal`). Blacklisted / removed members lose search.
- **thread** — membership checked against the parent group parsed from `{group_no}____{short_id}`. A channel_id that does not parse is *denied*, not skipped.

Non-members get `NOT_FOUND` with `resource=channel` (anti-enumeration). Lookup errors fail closed with `INTERNAL_ERROR`.

### Shared infrastructure

- `es_client.go` — OpenSearch client singleton, lazy build, retry-until-ready
- `sender_join.go` — sender info JOIN with LRU cache (10K cap, scope+uid keyed)
- `cursor.go` — HMAC-signed pagination cursor, 3-tuple support for relevance sort
- `ratelimit.go` — per-loginUID 5 rps / 20 burst limiter; transport errors route to `UPSTREAM_UNAVAILABLE`, fail-open on Redis outages
- `audit.go` — request audit middleware (PRM-02)
- `swagger/api.yaml` — single API source of truth

### Configuration (env-driven; defaults in `modules/messages_search/config.go`)

| Var | Default | Notes |
|---|---|---|
| `OCTO_SEARCH_OS_ADDRS` | `http://localhost:9200` | comma-separated OS endpoints |
| `OCTO_SEARCH_OS_USERNAME` / `_PASSWORD` | empty | optional basic auth |
| `OCTO_SEARCH_OS_READ_ALIAS` | `wukongim-messages-read` | read alias name |
[...]

PR #370, opened 2026-06-14
feat(authz): CI guard requiring a role check on every /v1/manager route (#366)

### LINKED ISSUE #366: authz: centralized authorization layer + CI guard requiring a role check on every /v1/manager route
Follow-up from the authorization-model audit (#363); the specific gaps it surfaced were fixed in #364, but the **systemic root cause** remains and is tracked here.

## Problem

Authorization is enforced as bare role checks scattered across the handlers — there is no central policy layer and no safety net. Adding a new `/v1/manager/*` (or other privileged) endpoint relies entirely on the developer remembering to add the right `CheckLoginRole` / `CheckLoginRoleIsSuperAdmin` / `requireAdmin` / `requireSuperAdmin` call. A forgotten check is a silent privilege escalation that nothing catches — exactly the class of bug #364 had to fix by hand.

## Proposed work (two parts, ship part 1 first)

**Part 1 — CI guard (cheap, high-value, do first).**
Add a source-guard test (mirroring the existing `make i18n-lint` / `Test<Module>NoLegacyResponseError` pattern) that asserts **every handler mounted under `/v1/manager` performs a role check**. Fail CI if a manager route has no gate. Maintain an explicit allowlist for any intentional exception (documented, like the i18n baseline). This turns "did the dev remember?" into a mechanical CI gate.

**Part 2 — centralized authz layer (design discussion).**
Extract role requirements into a declarative layer (per-route required role via middleware / policy table) so the requirement lives next to the route registration rather than hand-rolled inside each handler body. Lets reviewers reason about the whole privilege surface in one place and removes the per-handler boilerplate.

## Acceptance criteria

- [ ] Part 1: CI guard exists and passes on `main`; a new ungated `/v1/manager` route fails CI; intentional exceptions are allowlisted with a reason.
- [ ] Part 2: design proposal for the central authz layer (can be a separate PR/RFC once Part 1 lands).

## References
- Audit: #363 ("鉴权裸数值散落各 handler，新增端点靠开发者自觉，漏了即权限放大，无机制兜底")
- Precedent fix (manual, per-handler): #364

an9xyz:
## Summary

Closes #366. Authorization on management endpoints is enforced as bare role checks scattered across handler bodies, with no central policy and no safety net — a new `/v1/manager` endpoint that forgets its `CheckLoginRole` / `CheckLoginRoleIsSuperAdmin` call is a silent privilege escalation that nothing catches (the #364 class of bug, fixed there by hand). This PR delivers both parts the issue tracks:

- **Part 1 (implemented): a CI guard** that turns "did the dev remember the gate?" into a mechanical gate.
- **Part 2 (design proposal): an RFC** for the centralized authz layer, with the §8 design decisions resolved.

## Part 1 — `tools/lint-manager-authz` (CI guard)

An AST guard (mirroring `lint-direct-error-response` / `lint-personal-msgsendreq`) asserting **every handler mounted under `/v1/manager` performs a role check**. A route is "gated" if either:

- **(a)** its handler body — or a same-package helper it delegates to (bounded, cycle-guarded) — calls `CheckLoginRole`, `CheckLoginRoleIsSuperAdmin`, or a `require*` wrapper; **or**
- **(b)** the route/group carries a recognized declarative authz middleware (`RequireRole` / `RequireAdmin` / `RequireSuperAdmin`).

Branch (b) is the forward-compat hook for Part 2: moving a check from handler body into a route middleware keeps the guard green with no change. `AuthMiddleware` is deliberately **not** treated as a gate — it is authentication, not authorization.

Design notes:
- **Fail-closed**: a route whose handler can't be resolved is treated as ungated, so an unfamiliar shape fails loudly rather than slipping through unverified.
- **Strict allowlist**: intentional exceptions live in `allowlist.txt` with a mandatory reason; an entry that matches no ungated route is reported **STALE** and fails the lint, so the exception list can't rot. Seeded with the two real exceptions: the user-scoped `/v1/manager/secrets/*` CRUD (owner = caller, not admin) and the pre-auth `/v1/manager/login` entry.
- During development the guard surfaced a real delegation pattern (`space` `list`/`disableList` → `listByStatuses` → `requireAdmin`); the body scan was upgraded to bounded transitive resolution to avoid that false positive.

Wiring: `make authz-lint` target + a dedicated **Manager Authz Lint** CI job.

## Part 2 — `docs/2026-06-authz-centralized-layer-rfc.md` (design proposal)

RFC proposing the centralized authz layer: a declarative role middleware (`RequireAdmin()` / `RequireSuperAdmin()` in `pkg/auth`) mounted at route registration after `AuthMiddleware`, moving the requirement out of the handler body. Non-breaking (same tiers, same generic `ErrSharedForbidden` 403, revocation-aware via the #364 resolver), already recognized by the Part-1 guard, with an incremental module-by-module migration plan and an eventual registration-time runtime guard.

**Resolved §8 decisions:** middleware-at-registration enforcement (not a policy table); `/v1/manager`-only first scope; two named middlewares; registration-time runtime guard; keep the Part-1 AST guard as belt-and-braces.

> The Part-2 *implementation* (the middleware + per-module migration) is beyond #366's acceptance (Part 2 = design proposal) and will land as separate follow-up PR(s) per the RFC.

## Test plan

- `go run ./tools/lint-manager-authz` (and `make authz-lint`): green — **108 manager routes scanned, all gated or allowlisted, 2 allowlist entries**.
- Adversarial check: a throwaway ungated `/v1/manager` route makes the guard exit 1 with a precise `file:line` + handler message; removing it returns to green.
- `go test ./tools/lint-manager-authz/`: covers each gate path (direct / `require*` wrapper / one-hop delegation / authz middleware / sub-group prefix / genuinely ungated / fail-closed), allowlist parsing (incl. mandatory-reason errors), and the violation-and-stale decision.
- `gofmt`, `go vet` clean; `tools/workflow-guards` test unaffected by the new CI job.

## Acceptance criteria (#366)

- [x] Part 1: CI guard exists and passes on the branch; a new ungated `/v1/manager` route fails CI; intentional exceptions are allowlisted with a reason.
- [x] Part 2: design proposal for the central authz layer (decisions resolved in §8).

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -306,6 +306,31 @@ jobs:
       - name: Reject direct PERSONAL MsgSendReq literals
         run: go run ./tools/lint-personal-msgsendreq ./modules
 
+  manager-authz-lint:
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
+    # Mininglamp-OSS/octo-server#366 Part 1: authorization safety net. The lint
+    # AST-walks modules/ and asserts every handler mounted under /v1/manager
+    # performs a role check (in-handler CheckLoginRole* / require* wrapper, or a
+    # declarative authz middleware). A new ungated management endpoint fails CI
+    # instead of shipping as a silent privilege escalation (the #364 class of
+    # bug). Intentional exceptions are allowlisted with a reason in
+    # tools/lint-manager-authz/allowlist.txt; a stale allowlist entry also fails.
+    name: Manager Authz Lint
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Require a role check on every /v1/manager route
+        run: go run ./tools/lint-manager-authz
+
   i18n-extract-check:
     needs: [changes]
     if: |
--- a/Makefile
+++ b/Makefile
@@ -49,10 +49,20 @@ env-test:
 #                   - lint-unregistered-code: no inline codes.Code{} literals
 #                     passed to httperr.ResponseErrorL (registry bypass)
 
-.PHONY: i18n-extract i18n-extract-check i18n-merge i18n-lint
+# authz-lint   : #366 Part 1 — assert every handler mounted under /v1/manager
+#                 performs a role check (in-handler CheckLoginRole* / require*
+#                 wrapper, or a declarative authz middleware). Intentional
+#                 exceptions live in tools/lint-manager-authz/allowlist.txt with
+#                 a reason; the lint fails on an ungated route OR a stale
+#                 allowlist entry. Mirrors the i18n-lint guard pattern.
+
+.PHONY: i18n-extract i18n-extract-check i18n-merge i18n-lint authz-lint
 
 i18n-extract:
 	go run ./pkg/i18n/cmd/octo-i18n-extract
+
+authz-lint:
+	go run ./tools/lint-manager-authz
 
 i18n-lint:
 	go run ./tools/lint-direct-error-response
--- a/docs/2026-06-authz-centralized-layer-rfc.md
+++ b/docs/2026-06-authz-centralized-layer-rfc.md
@@ -1,0 +1,265 @@
+# RFC: Centralized Authorization Layer for Management Routes
+
+**Issue**: [#366](https://github.com/Mininglamp-OSS/octo-server/issues/366) Part 2 · follow-up to [#363](https://github.com/Mininglamp-OSS/octo-server/issues/363) audit / [#364](https://github.com/Mininglamp-OSS/octo-server/pull/364) fix
+**Status**: **Draft RFC — design proposal for maintainer review. Not yet implemented.**
+**Date**: 2026-06-14
+**Scope**: design only. Part 1 (the CI guard, `tools/lint-manager-authz`) has already landed; this document proposes Part 2 and how it dovetails with Part 1. No business code is changed by this RFC.
+**Decision target**: agree on the *shape* of the declarative authz layer (middleware-first vs. policy-table), the migration strategy, and the eventual runtime guard — before any module is migrated.
+
+---
+
+## TL;DR
+
+1. Today every `/v1/manager` handler hand-rolls its own role check in the handler body (`c.CheckLoginRole()` / `c.CheckLoginRoleIsSuperAdmin()`, or a per-module `requireAdmin` / `requireSuperAdmin` wrapper). Part 1 added a CI guard that *catches a missing check*; it does not *remove the boilerplate* or let a reviewer see the whole privilege surface in one place. That is Part 2.
+2. **Proposal: a declarative role middleware** in `pkg/auth` — `RequireAdmin()` / `RequireSuperAdmin()` (thin wrappers over `RequireRole(min)`) — mounted **at route registration**, next to the path, *after* `AuthMiddleware`. The requirement moves from inside the handler body to the route declaration: `auth.POST("/x", authz.RequireSuperAdmin(), m.handler)`.
+3. It is **non-breaking**: same role semantics, same single generic `ErrSharedForbidden` 403 (anti-enumeration preserved), and it reads the **per-request resolved role** wired in #364, so it stays revocation-aware for free.
+4. **It already fits Part 1.** The guard's branch (b) recognizes `RequireRole` / `RequireAdmin` / `RequireSuperAdmin` by name from day one — a migrated route stays "gated" with **zero guard churn**, and the allowlist (`/v1/manager/secrets/*`, `/v1/manager/login`) is unaffected.
+5. **Migration is incremental, module-by-module.** Start with uniform-tier modules (`backup`, `opanalytics` — all `superAdmin`), then mixed-tier ones. Each migration deletes the per-module `require*` + `respond*Forbidden` boilerplate and keeps the existing "plain `admin` is rejected" regression tests green.
+6. **End state: a runtime guard supersedes the AST heuristic.** Once gates are middleware, a test can enumerate the registered router and assert every `/v1/manager` route's middleware chain contains an authz middleware — strictly stronger than the AST guard (no handler-resolution heuristics). The Part-1 AST guard stays as belt-and-braces until migration completes.
+
+---
+
+## 1. Background & current state
+
+The #363 audit found the systemic root cause: *"鉴权裸数值散落各 handler，新增端点靠开发者自觉，漏了即权限放大，无机制兜底"*. #364 fixed the specific gaps by hand (raised four destructive cross-space ops + `appversion` to `superAdmin`) and, importantly, added **per-request role resolution** so a demotion converges within a cache TTL instead of token lifetime:
+
+- `pkg/auth/parser.go` — `RoleResolver` interface + `WithRoleResolver`; `Parse` overrides the token's baked-in role with the authoritative `user_role:{uid}` value (Redis → DB → `""`), fail-open to the snapshot on resolver error.
+- `modules/user/role_service.go` — the concrete `RoleService`.
+- `main.go` — wires `auth.WithRoleResolver(userRoleSvc)` into the token parser.
+
+Part 1 (this issue) added `tools/lint-manager-authz`: an AST guard asserting every handler under `/v1/manager` performs a role check, with an allowlist for the two intentional exceptions.
+
+**Inventory the guard reports today: 108 manager routes across 13 modules, 2 allowlisted.** The four gate idioms in use:
+
+| Idiom | Tier | Example |
+|---|---|---|
+| `c.CheckLoginRole()` | admin ∪ superAdmin | `modules/message/api_manager.go` |
+| `c.CheckLoginRoleIsSuperAdmin()` | superAdmin only | `modules/backup/api_manager.go` |
+| `m.requireAdmin(c)` (wrapper) | admin ∪ superAdmin | `modules/space/api_manager.go:161` |
+| `m.requireSuperAdmin(c)` (wrapper) | superAdmin only | `modules/space/api_manager.go:174`, `modules/opanalytics/api.go:267` |
+
+All four ultimately call the same two `wkhttp.Context` methods and respond with the same `errcode.ErrSharedForbidden`. The wrappers are **duplicated per module** (`space`, `opanalytics`, … each define their own), and every handler repeats the same 3–4 line preamble:
+
+```go
+func (m *Manager) handler(c *wkhttp.Context) {
+    if err := c.CheckLoginRoleIsSuperAdmin(); err != nil { // ← boilerplate
+        respondManagerForbidden(c)                          // ← boilerplate
+        return                                              // ← boilerplate
+    }
+    // ... actual work ...
+}
+```
+
+### Problems this leaves
+
+- **Boilerplate**: ~108 copies of the same preamble; 5+ duplicated `require*` / `respond*Forbidden` helpers.
+- **No single audit surface**: to answer "which manager routes are superAdmin-only?" you must open every handler body. The requirement is not visible at the route table.
+- **Wrong-tier is invisible**: Part 1 catches a *missing* gate but cannot catch `CheckLoginRole` used where `superAdmin` was intended — that judgment is buried in the body.
+
+---
+
+## 2. Goals / non-goals
+
+**Goals**
+- Move the role requirement to the **route registration site**, declaratively.
+- One shared implementation; delete the per-module wrappers.
+- Preserve exact behavior: same tiers, same single generic 403, revocation-aware via the #364 resolver.
+- Make the privilege surface **auditable in one place** (the route tables, and eventually a runtime enumeration test).
+- Stay inside octo-server (no octo-lib release): the middleware builds on `wkhttp.Context` methods that already exist.
+
+**Non-goals (explicitly out of scope here)**
+- Changing any route path or role tier (that was #364's job; this is a pure refactor of *where the check lives*).
+- A general RBAC/permission system, scopes, or per-resource ACLs.
+- Migrating non-manager privileged surfaces (e.g. the one `/v1/common` write #363 mentioned) — can adopt the same middleware later; the guard's prefix list is already extensible.
+- Retiring the Part-1 AST guard (kept until the runtime guard fully covers migration).
+
+---
+
+## 3. Proposed design
+
+### 3.1 The middleware (primary mechanism)
+
+A new file `pkg/auth/authz.go` exposing role-enforcing middleware that returns `wkhttp.HandlerFunc`:
+
+```go
+package auth
+
+import (
+    "github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+    "github.com/Mininglamp-OSS/octo-server/pkg/errcode"
+    "github.com/Mininglamp-OSS/octo-server/pkg/httperr"
+)
+
+// RequireAdmin gates a route to the admin ∪ superAdmin tier. Mount AFTER
+// AuthMiddleware (it reads the per-request resolved role off the context).
+// On failure it writes the same generic ErrSharedForbidden 403 the in-handler
+// checks use — no "needs higher role" leak (anti-enumeration) — and aborts the
+// chain so the handler never runs.
+func RequireAdmin() wkhttp.HandlerFunc {
+    return func(c *wkhttp.Context) {
+        if err := c.CheckLoginRole(); err != nil {
+            httperr.ResponseErrorL(c, errcode.ErrSharedForbidden, nil, nil)
+            c.Abort()
+            return
+        }
+        c.Next()
+    }
+}
+
+// RequireSuperAdmin gates a route to superAdmin only (cross-space destructive /
+// supply-chain-sensitive writes). Same response + abort contract as above.
+func RequireSuperAdmin() wkhttp.HandlerFunc {
+    return func(c *wkhttp.Context) {
+        if err := c.CheckLoginRoleIsSuperAdmin(); err != nil {
+            httperr.ResponseErrorL(c, errcode.ErrSharedForbidden, nil, nil)
+            c.Abort()
+            return
+        }
+        c.Next()
+    }
+}
+```
+
+> A single `RequireRole(min Role)` core with the two as thin wrappers is equivalent; the two-function form is proposed because the codebase only has two tiers and explicit names read better at the call site. The Part-1 guard recognizes all three names.
+
+Because the check delegates to `c.CheckLoginRole*`, which read the role that `CacheTokenParser.Parse` already resolved per request (#364), the middleware is **revocation-aware with no extra wiring**.
+
+### 3.2 Mounting at registration
+
+**Per-route** (the default, because most manager groups mix tiers — reads at `admin`, writes at `superAdmin`):
+
+```go
+auth := r.Group("/v1/manager", m.ctx.AuthMiddleware(r))
+{
+    auth.GET("/spaces", authz.RequireAdmin(), m.list)                  // read → admin
+    auth.DELETE("/spaces/:space_id", authz.RequireSuperAdmin(), m.forceDisband) // destructive → superAdmin
+}
+```
+
+**Group-level** where every route shares one floor (e.g. `backup` and `opanalytics` are uniformly `superAdmin`):
+
+```go
+auth := r.Group("/v1/manager",
+    m.ctx.AuthMiddleware(r),
+    authz.RequireSuperAdmin(), // every route below requires superAdmin
+)
+```
+
+The handler bodies lose their preamble entirely:
+
+```go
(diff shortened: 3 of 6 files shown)


Jerry-Xin (changes requested):
This PR is relevant to octo-server and the goal is sound, but the new CI guard — which IS the security control here — has two demonstrable false-negative route shapes that break its core acceptance criterion ("every handler mounted under `/v1/manager` performs a role check" / "a new ungated `/v1/manager` route fails CI"). A preventive guard with bypasses gives false assurance, so these block.

Note: on the *current* codebase all 108 manager routes are registered via a `Group("/v1/manager…")` variable, so the scan is accurate today and CI is legitimately green. The findings below are latent gaps a future PR could slip an ungated manager route through — exactly the #364 class this guard exists to stop.

## 🔴 Blocking

**🔴 Direct router-level `/v1/manager` registrations are invisible to the guard.**
`pkgScope.routes()` only collects verb calls whose receiver variable was bound by a `Group(...)` with a manager prefix (`managerGroupsIn`). A route registered directly on the router with a literal manager path is never seen. Reproduced against this branch:
```go
// modules/x/api.go
func (a *API) Route(r *R) {
    r.GET("/v1/manager/direct_bypass", a.ungated) // no Group(...)
}
func (a *API) ungated(c *C) { _ = c.Query("x") }
```
`go run ./tools/lint-manager-authz` still prints `OK: 108 manager route(s) scanned` and exits 0. This contradicts "every handler mounted under `/v1/manager`". Fix: also evaluate verb calls whose **first arg is a string literal with a manager prefix**, regardless of whether the receiver is a tracked group (or fail loudly on such shapes).

**🔴 Non-literal route paths inside a known manager group are dropped, not failed-closed.**
`buildRoute` returns `nil` when the path arg is not a string literal, removing the route from analysis entirely. Reproduced against this branch:
```go
func (a *API) Route(r *R) {
    auth := r.Group("/v1/manager", a.authMW())
    p := "/danger"
    auth.GET(p, a.ungated) // computed path -> route vanishes
}
```
Result: `OK: 108 ... scanned`, exit 0. The package docstring promises fail-closed ("an unfamiliar shape fails loudly rather than slipping through unverified"), but here an unfamiliar shape *inside a protected group* silently disappears. Fix: when the path is non-literal but the group is manager-rooted, emit an ungated violation with an "unresolved path" reason instead of returning `nil`.

## 💬 Non-blocking

- 🟡 Authz-middleware detection (`hasAuthzMiddleware`) is name-only: any factory named `RequireRole` / `RequireAdmin` / `RequireSuperAdmin` counts regardless of package or behavior. Acceptable for this repo's name-based lint style and matches the sibling lints, but worth tightening (package-qualify) once the Part-2 central middleware actually lands.
- `AuthMiddleware` is correctly excluded from the authz set (authn ≠ authz) — good call and well-documented.

## ✅ Highlights

- Strict allowlist model: mandatory `# reason`, and STALE entries fail the lint so exceptions can't rot. Both seeded entries (user-scoped `/v1/manager/secrets/*`, pre-auth `/v1/manager/login`) are correctly justified.
- Bounded, cycle-guarded transitive helper scan handles the real `space list → listByStatuses → requireAdmin` delegation without false positives.
- `go test ./tools/lint-manager-authz` and `go run ./tools/lint-manager-authz` both pass on this checkout; the CI job and `make authz-lint` wiring are correct.

yujiawei (approved):
# Code Review — PR #370 (octo-server)

**Head reviewed:** `5d545666f08c188076bc7ea42b95d358884a344a` · base `main` (merge-base `24f7844a`)

## Summary

This delivers #366: **Part 1** an AST-based CI guard (`tools/lint-manager-authz`) asserting every handler under `/v1/manager` performs a role check, and **Part 2** a design-only RFC for the future centralized authz layer. I checked it out at the head SHA, ran the tool against the real tree, ran the unit tests, and probed the guard adversarially for false negatives (the failure mode that matters for a security guard).

**Verdict: Approve.** The guard is correct, well-tested, fail-closed on unresolved handlers, and its allowlist is accurate. No P0/P1. The findings below are non-blocking hardening notes — the guard is additive defense-in-depth (real authorization still lives in the handlers), so its blind spots don't introduce a regression.

## Verification

| Check | Result |
|---|---|
| `go run ./tools/lint-manager-authz` on real tree | ✅ `OK: 108 manager route(s) scanned, all gated or allowlisted (2 allowlist entries)` |
| `go test ./tools/lint-manager-authz/` | ✅ pass |
| `go vet`, `gofmt -l` | ✅ clean |
| Adversarial: inject an ungated `/v1/manager` route | ✅ exits 1 with precise `file:line` + handler name; reverts to green when removed |
| Allowlist entry 1 — `/v1/manager/secrets/*` | ✅ genuinely owner-scoped: all handlers in `modules/usersecret/api.go` key on `c.GetLoginUID()` (lines 131/157/179/220), no role check is correct |
| Allowlist entry 2 — `POST /v1/manager/login` | ✅ genuinely pre-auth: mounted on the no-AuthMiddleware group (`modules/user/api_manager.go:63`), role verified from the DB record post-password |
| All 13 real manager registrations use `Group("/v1/manager")` | ✅ matches the guard's detection assumption |
| CI wiring (`changes` filter → `manager-authz-lint`) | ✅ runs whenever any non-doc file changes; tool scans all of `modules/` regardless of which file changed, so a new route can't dodge it |

## Findings

### P2 — Route detection is scoped to single-function `Group("/v1/manager")` registration; two shapes are silently *uncollected* (not fail-closed)

The "fail-closed" guarantee in the package doc covers a route whose *handler* can't be resolved. It does **not** cover a route that is never collected as a route in the first place. I confirmed two shapes the collector misses entirely (output shows zero routes from the snippet, only stale-allowlist noise):

1. **Direct full-path registration on the router:** `r.GET("/v1/manager/x", handler)` — no intermediate `Group("/v1/manager")` variable. `managerGroupsIn` only tracks vars bound to a `Group(...)` call, so this verb call is never matched.
2. **Group handed to a helper:** `auth := r.Group("/v1/manager", ...); a.registerSub(auth)` where `registerSub(g *RouterGroup)` mounts the routes. The route scan is per-function over the function that creates the group, so routes mounted in the callee are invisible.

**Not a current hole** — I verified neither pattern exists under `/v1/manager` today (`grep` for direct full-path manager verb calls returns 0; the only cross-function mounter, `incomingwebhook.MountManagementRoutes`, mounts under `/v1/bot`, not `/v1/manager`). But for a guard whose whole value proposition is "a forgotten gate fails CI," an *uncollected* route is the one escape hatch that produces a green check with no signal. Suggest either (a) a documented known-limitation note next to the fail-closed doc, or (b) a cheap belt: also flag any `*.VERB("/v1/manager...", ...)` literal whose receiver group wasn't recognized. The Part-2 runtime guard (enumerate the registered router) closes this structurally — worth stating explicitly in the RFC that this is *why* the runtime guard supersedes the AST one.

### P2 — `gateSelectorNames` matches the call by name only; a called-but-unchecked `CheckLoginRole()` would pass

`c.CheckLoginRole()` / `c.CheckLoginRoleIsSuperAdmin()` return an `error` and do **not** self-abort (confirmed in octo-lib `pkg/wkhttp/http.go:180,192`). The guard counts the gate as present when the call *appears* in the body; it does not verify the returned error is checked and the request aborted. A handler that wrote `c.CheckLoginRole()` as a bare statement (ignoring the error) would be reported gated while enforcing nothing.

**Not currently exploited** — every real call site handles the error (`if err := c.CheckLoginRole(); err != nil { ... return }` or assigns and checks). This also matches the name-based heuristic the sibling lints already use, so it's a reasonable accepted limitation rather than a defect. Flagging so it's a conscious choice: a one-line note in the package doc (or an `errcheck`-style follow-up) would make the boundary explicit.

### Nit — `maxGateDepth = 6` transitive scan is unbounded only by depth, not breadth

The delegation walk is depth- and cycle-guarded (good — it fixed the real `space.list → listByStatuses → requireAdmin` false positive). Depth 6 is generous for the current code; no action needed, just noting the bound is the right shape.

## Things done well

- **Fail-closed on unresolved handlers** with an explicit test (`TestResolutionFailClosed`) — the correct default for a security lint.
- **Strict allowlist with mandatory reasons + STALE detection** — exceptions can't silently rot; both seeded entries are real and verified.
- **`AuthMiddleware` deliberately excluded** from the authz set with a clear rationale (authn ≠ authz) — exactly the #366 distinction.
- **Branch (b) forward-compat hook** for the Part-2 middleware names is wired from day one, so the migration lands with zero guard churn.
- Test coverage spans every gate path, allowlist parsing errors, and the violation/stale decision.
- RFC is genuinely design-only (no business code touched) with §8 decisions resolved.

## For the human reviewer (security-sensitive)

Nothing blocks merge. The one thing worth a human's eye is the **scope of detection** (Finding 1): this guard protects exactly the routes registered via an in-function `Group("/v1/manager")`. That covers 100% of today's surface, but the guarantee is "no forgotten gate *among recognized routes*," not "no ungated manager route can ever exist." If the team later adopts a different registration idiom under `/v1/manager`, the guard must be extended in lockstep — the Part-2 runtime guard is the durable fix.

yujiawei (changes requested):
# Code Review — PR #370 (octo-server) — Revised Verdict

**Head reviewed:** `5d545666f08c188076bc7ea42b95d358884a344a` · base `main` (merge-base `24f7844a`)

> **Note:** This supersedes my earlier Approve on this PR. On a deeper adversarial pass I confirmed two false-negative route shapes that defeat the guard's core acceptance criterion, including one that contradicts the tool's own documented fail-closed contract. For a preventive security control, a demonstrable silent bypass is blocking, so I'm revising to **Changes Requested**.

## Summary

#366 Part 1 is a CI guard (`tools/lint-manager-authz`) asserting every `/v1/manager` handler performs a role check; Part 2 is a design-only RFC. The mechanism is well-built and **accurate on today's codebase** (all 108 manager routes register via a `Group("/v1/manager…")` variable, so CI is legitimately green now). The blockers below are latent bypasses a *future* PR could slip an ungated manager route through — i.e. exactly the #364 class this guard exists to prevent. A guard with silent bypasses gives false assurance, which is worse than a known-absent guard.

## Verification

| Check | Result |
|---|---|
| `go run ./tools/lint-manager-authz` on real tree | ✅ `OK: 108 manager route(s) scanned, all gated or allowlisted (2 allowlist entries)` |
| `go test ./tools/lint-manager-authz/`, `go vet`, `gofmt -l` | ✅ clean |
| Adversarial: ungated route via tracked `Group` | ✅ caught — exits 1 with precise `file:line` + handler |
| Allowlist `/v1/manager/secrets/*` | ✅ genuinely owner-scoped (`c.GetLoginUID()` in all handlers, `modules/usersecret/api.go`) |
| Allowlist `POST /v1/manager/login` | ✅ genuinely pre-auth (no-AuthMiddleware group, `modules/user/api_manager.go:63`) |

## 🔴 Blocking

### P1 — Non-literal route path inside a recognized manager group is silently dropped (contradicts documented fail-closed)

`buildRoute` (`tools/lint-manager-authz/main.go:424-426`) returns `nil` when the path arg is not a string literal — **before** handler resolution — so the route vanishes from analysis entirely. Reproduced against this branch:

```go
func (a *API) Route(r *R) {
    auth := r.Group("/v1/manager", a.authMW()) // recognized manager group
    p := "/danger"
    auth.GET(p, a.ungated)                      // computed path → route disappears
}
```
Result: `OK: … scanned`, exit 0 — no violation. This is the sharp one: the route is inside an *already-recognized* protected group, the handler is resolvable and ungated, yet it's dropped purely because the path is computed. The package docstring (`main.go:35-40`) promises "an unfamiliar shape fails loudly rather than slipping through unverified" — here an unfamiliar shape inside a protected group slips through silently. **Fix:** when the path is non-literal but the group is manager-rooted, emit an ungated violation (reason: "unresolved path") instead of returning `nil`.

### P1 — Direct router-level `/v1/manager` registration is invisible to the guard

`pkgScope.routes()` only collects verb calls on a receiver var bound by a manager-prefixed `Group(...)` (`managerGroupsIn`). A route registered directly on the router with a literal manager path is never seen. Reproduced against this branch:

```go
func (a *API) Route(r *R) {
    r.GET("/v1/manager/direct_bypass", a.ungated) // no Group(...)
}
```
Result: `OK: 108 manager route(s) scanned`, exit 0 — directly contradicting "every handler mounted under `/v1/manager`". Lower real-world risk than the above (this idiom is unused in the repo today), but it's a clean hole in the stated criterion. **Fix:** also evaluate verb calls whose first arg is a string literal with a manager prefix regardless of receiver tracking — or fail loudly on such shapes. (A related gap: a manager group handed to a helper func that mounts routes is likewise uncollected; the same broadening would help.)

## 💬 Non-blocking

- **P2** — `gateSelectorNames` / `hasAuthzMiddleware` match by name only. A bare `c.CheckLoginRole()` whose returned `error` is ignored would count as gated (the call returns `error` and does not self-abort — octo-lib `pkg/wkhttp/http.go:180,192`). Not currently exploited (all real call sites check the error) and consistent with the sibling lints' name-based style — worth a one-line known-limitation note.
- Authz-middleware detection is package-unqualified; fine for this repo's lint style, worth tightening once the Part-2 central middleware lands.

## ✅ Highlights

- Strict allowlist: mandatory `# reason` + STALE detection so exceptions can't rot; both seeded entries verified correct.
- Fail-closed on *unresolved handlers* with a dedicated test (`TestResolutionFailClosed`).
- Bounded, cycle-guarded transitive helper scan handles the real `space list → listByStatuses → requireAdmin` delegation without false positives.
- `AuthMiddleware` deliberately excluded from the authz set (authn ≠ authz), well documented.
- RFC is genuinely design-only; the Part-2 runtime guard (enumerate the registered router) structurally closes both blockers above — worth stating in the RFC that this is *why* it supersedes the AST heuristic.

## For the human reviewer (security-sensitive)

The guard is correct for the current tree; nothing ungated ships today. The blockers are about the guard's *durability* as a preventive control: two route shapes (computed path in a protected group; direct router registration) produce a green check with zero signal. Since the whole point is "a forgotten gate can't reach main," closing the silent-bypass paths before merge is the right bar.

lml2468 (approved):
✅ APPROVE — 设计合理、fail-safe 方向正确的授权 guardrail。本地逐条核实 + build/test 全绿。无 blocking。

## 我否决了 workflow 的两条"blocking"（方向搞反了）
workflow 报「变量传 middleware（r.Group("/v1/manager", authzMW) / .Use(mw)）不被识别 → 漏报」。我核实 main.go middlewareNames（:588）只认 CallExpr：变量传 middleware 时返回 [] → 该 route **不被认作 middleware-gated** → 检查 handler 有无 inline gate → 没有就**标记为 UNGATED 违规、build 失败**。

这是**误报方向（过严）**，不是漏报——它让一个其实被 middleware 保护的路由被标违规，开发者用 allowlist 或加 inline gate 解决。**guardrail 仍 fail-safe，不会放未授权路由过关**。workflow 把过严误报说成了"evasion 漏报"，定级错误，降为 should-fix。

## 核实通过（正面）
- ✅ parse 错误 **fail-closed**：parsePackage err → collectManagerRoutes err → os.Exit(2)（:116），build 失败不跳过。（否决 workflow「fail-open parser」那条）
- ✅ STALE 检测正确（analyze :176：allowlist 条目不匹配任何 ungated route 即报错，防腐烂）。
- ✅ 测试覆盖关键方向：TestCollectRoutes_Gating 验 ungated route 被标 false、TestAnalyze 验 ungated 触发唯一 violation——guardrail 核心断言在，不是 no-op。
- ✅ allowlist 两条豁免合理：/secrets/* 真 user-scoped（usersecret/api.go 每个 handler 用 GetLoginUID()，非 admin）、login 真 pre-auth + DB 验角色。
- build / test 全绿。

## should-fix（建议收，非阻塞）
1. 🟡 **wildcard allowlist over-match**（真实盲点）：`* /v1/manager/secrets/*` 会静默豁免未来加到该子树的**新**路由（matchAllowlist 用 HasPrefix）。限于 secrets 子树（合法 user-scoped），但建议加注释警示「子树内新增非 user-scoped 路由须单列」，或收窄为精确条目。
2. 🟡 **变量/SelectorExpr middleware 误报（过严）**：变量传 / 包级未调用引用的 authz middleware 会触发误报。非安全洞，但影响可用性——真用此写法会误失败。建议 middlewareNames 也识别 Ident/SelectorExpr 引用（解析变量到其 factory）。
3. 🟡 **conditional in-handler gate 无控制流分析**：gate 放在 if 分支里也算 gated（bodyGated 只做 AST 树遍历找调用）。要故意构造才触发，但是真·弱检测，建议至少文档化此限制。

## nit
RFC 文档若干表述、动态路由路径跳过（安全默认，已注释）。

无 blocking。这是「用静态检查兜住授权不漏挂」的好实践，方向正确、fail-safe、测试覆盖到位。建议后续收 should-fix（尤其 wildcard 盲点 + 变量 middleware 误报）。批准。

—— Octo-PR (lml2468)

lml2468:
@Jerry-Xin 关于「测试探针代码未删除」这条 🔴，我做了对抗式核查但**没能定位到 shipped 探针代码**，想请你指个具体 file:line，避免我俩各看各的。

我的核查（head 5d54566，PR 仅动 6 文件：ci.yml / Makefile / rfc.md / allowlist.txt / main.go / main_test.go）：
1. main.go 里的 fmt.Print* 全是 linter 的 **CLI 正常输出**（OK/UNGATED/STALE 报告），非探针；唯一 os.Args 是正常入参。无 debug flag / os.Getenv / 注释残块 / panic 探针。
2. main_test.go 的 fixtures 写入 **t.TempDir()**（临时目录、用例结束即删），是标准测试设计，不落到真实 modules/。
3. ci.yml 新增 job 只 `go run ./tools/lint-manager-authz`，无临时 echo / `|| true`。
4. **唯一像"测试代码"的是 docs/2026-06-...rfc.md 里的 `TestEveryManagerRouteIsRoleGated`——但它在 markdown 的 ```go 围栏内、是 RFC 的未来提案**，不是 .go 文件（grep 确认仅存在于该 .md）。是否你把这段文档提案当成遗留代码了？

如果你指的是别处的真探针，请给 file:line，我立刻复核、该 CR 我跟。如果就是 RFC 里那段文档提案，那它是有意的设计稿、不需删除——我维持 APPROVE。

（我先维持 APPROVE：无法据一个我证实不了的 finding 改票；等你的 file:line 再定。）

—— Octo-PR (lml2468)

lml2468 (changes requested):
🔴 撤回我的 APPROVE，改 CHANGES_REQUESTED —— Jerry-Xin 提的两条 linter fail-open 漏报我已本地核实成立，yujiawei 也独立指出其一。这是安全 guardrail 的真盲点，我先前漏判了（我审了 gate 检测逻辑，却没核 route 发现入口的 fail-open）。

（说明:echo 把 Jerry-Xin 的真实 review 误渲染成「测试探针未删」，实际是下面两条 fail-open，与探针无关。）

## 🔴 漏报一:直接注册（非 Group）的 manager 路由对 linter 不可见（fail-open）
main.go 路由发现只走 managerGroupsIn（找绑定到 .Group("/v1/manager") 的局部变量）。collectManagerRoutes 在每个函数体里:
```
groups := managerGroupsIn(fn)
if len(groups) == 0 { continue }   // ← 没有 Group 变量就整函数跳过
```
若某路由直接挂在 router receiver 上（r.GET("/v1/manager/x", h)，不经 .Group()），则该函数无 group 变量 → 被整体跳过 → 这条未授权 manager 路由**对 linter 完全不可见、CI 照样绿**。Jerry-Xin 复现「OK: 108 scanned, exit 0」。这正是本工具要防的 #364 类提权漏挂。

## 🔴 漏报二:manager group 内非字面量路径静默跳过（fail-open，违背 fail-loudly 承诺）
buildRoute（main.go:424-426）:
```
if !isStringLit(call.Args[0]) {
    return nil // dynamic path — cannot reason about it
}
```
manager group 内一条计算路径（变量/常量拼接）的路由被**静默丢弃**、不计入、不报错。工具自我标榜 fail-closed/fail-loudly，这里却是 fail-open——一条动态路径的未授权路由就此隐身。

## 为何严重（与我先前否决的"变量 middleware"相反）
我先前否决的两条 workflow"blocking"（变量传 middleware）是**过严误报方向**（标违规、build 失败），guardrail 仍 fail-safe，故降级——那个判断我维持。
但 Jerry-Xin 这两条是**真 fail-open**:未授权路由静默通过、CI 绿。安全 guardrail 的价值全在「无漏报」，这两条直接破坏该保证。当前代码库 108 路由恰好都用 Group + 字面量路径所以今天 CI 真绿，但这是**潜在盲区**，未来一个 PR 用直接注册/动态路径就能绕过——正是该工具存在的理由。

## 修法
1. 路由发现:除 group 变量外，也扫 router receiver 上直接调用的 routeVerbs，按 path 字面量是否命中 managerPrefixes 判定（不依赖 .Group()）。
2. 非字面量路径:manager 上下文里遇到 dynamic path 应 **fail-closed**（报「无法静态判定，请改字面量或 allowlist」并 exit 1），而非 return nil 静默跳过。
3. 补对应负测试（直接注册的 ungated route 被 flag、manager 内 dynamic path 被 flag）。

## 仍有效的正面结论（不变）
fail-closed 的 parse 错误处理、STALE 检测、allowlist 两条豁免合理、对 Group 形式的 gate 检测正确——这些我先前核实通过的仍成立。问题集中在路由发现的两个 fail-open 入口。

修这两条 fail-open（+ 负测试）后我立即重新 APPROVE。诚实复盘:我这次漏判和 #361/#364 同型——只审了在走的路径（gate 检测），没核入口/发现层的 fail-open 分支，靠 Jerry-Xin + yujiawei 补上。

—— Octo-PR (lml2468)

an9xyz:
Superseded by #409. Closing this older #366 implementation path so review and follow-up fixes can focus on the canonical PR.

PR #374, opened 2026-06-15
fix(messages_search): add OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY for sel…

BrDing-Rookie:
…f-signed OS clusters

`buildESClient` had no TLS-skip switch, so olivere/elastic's startup healthcheck (HEAD /) failed with x509 "certificate signed by unknown authority" against any OpenSearch cluster whose CA isn't in the pod's system trust store. All three nodes failed → "no Elasticsearch node available" → ESClient returned err → caller mapped to 503 upstream_unavailable. dmwork-test reproduces this on every POST /v1/messages/_search call.

Add a new env `OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY` (default false, strict-literal "true" to enable, matching the existing `OCTO_SEARCH_OS_INSECURE_HTTP` parsing). When true, inject a custom `http.Client` with `tls.Config{InsecureSkipVerify: true}` into the olivere client via `elastic.SetHttpClient`. A loud WARN log fires at module init so the deviation cannot stay enabled silently.

Orthogonal to the existing `OCTO_SEARCH_OS_INSECURE_HTTP`:
- INSECURE_HTTP gates basic-auth credentials over cleartext http://
- INSECURE_SKIP_VERIFY gates TLS certificate verification

Long-term fix is to install the internal CA in the pod or have the cluster issue properly trusted certificates; this env is the short-term unblock for dev / test environments.

## Changes

-
-

## Testing

- [ ] Unit tests added/updated
- [ ] Manually verified

## Checklist

- [ ] PR description is in English
- [ ] Added tests for my changes
- [ ] Updated documentation
- [ ] Followed commit message conventions (Conventional Commits)

### DIFF
--- a/modules/messages_search/api.go
+++ b/modules/messages_search/api.go
@@ -61,6 +61,11 @@ func New(ctx *config.Context) *Handler {
 			"built-in default cursor signing key. Set a per-deployment " +
 			"secret in production.")
 	}
+	if cfg.OSInsecureSkipVerify {
+		h.Warn("OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY=true; OpenSearch TLS " +
+			"certificate verification is DISABLED. Use only in dev/test " +
+			"environments with self-signed certificates.")
+	}
 	return h
 }
 
--- a/modules/messages_search/config.go
+++ b/modules/messages_search/config.go
@@ -22,9 +22,17 @@ type SearchConfig struct {
 	// http:// addresses. Off by default: credentials over cleartext HTTP are
 	// rejected at client build time unless this is explicitly set.
 	OSInsecureHTTP bool
-	Timeout        time.Duration
-	RateLimit      RateLimitCfg
-	CursorHMAC     string
+	// OSInsecureSkipVerify disables TLS certificate verification when talking
+	// to the OpenSearch read cluster. Required for dev / test environments
+	// that use self-signed or internal-CA-signed certificates that the
+	// pod's system trust store does not include. MUST stay false in
+	// production deployments where the OS cluster has properly trusted
+	// certificates. Off by default; opt in via
+	// OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY=true.
+	OSInsecureSkipVerify bool
+	Timeout              time.Duration
+	RateLimit            RateLimitCfg
+	CursorHMAC           string
 	// UserAvatarBaseURL, when non-empty, is prepended to the relative
 	// `users/{uid}/avatar` template so the response carries an absolute
 	// URL (spec v4.2 §2.1 / R8). When empty we keep the relative path and
@@ -54,12 +62,13 @@ type RateLimitCfg struct {
 // loadConfig builds a SearchConfig from process environment variables.
 func loadConfig() SearchConfig {
 	return SearchConfig{
-		OSAddrs:        splitCSV(os.Getenv("OCTO_SEARCH_OS_ADDRS"), []string{"http://localhost:9200"}),
-		OSUsername:     os.Getenv("OCTO_SEARCH_OS_USERNAME"),
-		OSPassword:     os.Getenv("OCTO_SEARCH_OS_PASSWORD"),
-		OSReadAlias:    defaultStr(os.Getenv("OCTO_SEARCH_OS_READ_ALIAS"), "wukongim-messages-read"),
-		OSInsecureHTTP: os.Getenv("OCTO_SEARCH_OS_INSECURE_HTTP") == "true",
-		Timeout:        parseDuration(os.Getenv("OCTO_SEARCH_TIMEOUT"), 5*time.Second),
+		OSAddrs:              splitCSV(os.Getenv("OCTO_SEARCH_OS_ADDRS"), []string{"http://localhost:9200"}),
+		OSUsername:           os.Getenv("OCTO_SEARCH_OS_USERNAME"),
+		OSPassword:           os.Getenv("OCTO_SEARCH_OS_PASSWORD"),
+		OSReadAlias:          defaultStr(os.Getenv("OCTO_SEARCH_OS_READ_ALIAS"), "wukongim-messages-read"),
+		OSInsecureHTTP:       os.Getenv("OCTO_SEARCH_OS_INSECURE_HTTP") == "true",
+		OSInsecureSkipVerify: os.Getenv("OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY") == "true",
+		Timeout:              parseDuration(os.Getenv("OCTO_SEARCH_TIMEOUT"), 5*time.Second),
 		RateLimit: RateLimitCfg{
 			QPS:   parseFloat(os.Getenv("OCTO_SEARCH_RPS"), 5.0),
 			Burst: parseInt(os.Getenv("OCTO_SEARCH_BURST"), 20),
--- a/modules/messages_search/config_test.go
+++ b/modules/messages_search/config_test.go
@@ -1,0 +1,49 @@
+package messages_search
+
+import (
+	"os"
+	"testing"
+)
+
+// The OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY env follows the same strict literal
+// "true" match as OCTO_SEARCH_OS_INSECURE_HTTP — anything else (1, yes, TRUE,
+// True, "") leaves verification ON. Lock that contract in so a future
+// "helpful" refactor to strconv.ParseBool doesn't widen the door silently.
+func TestLoadConfig_OSInsecureSkipVerify(t *testing.T) {
+	cases := []struct {
+		name string
+		val  string
+		set  bool
+		want bool
+	}{
+		{name: "unset defaults to false", set: false, want: false},
+		{name: "empty string is false", val: "", set: true, want: false},
+		{name: "literal true is true", val: "true", set: true, want: true},
+		{name: "uppercase TRUE is false", val: "TRUE", set: true, want: false},
+		{name: "mixed-case True is false", val: "True", set: true, want: false},
+		{name: "numeric 1 is false", val: "1", set: true, want: false},
+		{name: "yes is false", val: "yes", set: true, want: false},
+		{name: "false is false", val: "false", set: true, want: false},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.set {
+				t.Setenv("OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY", tc.val)
+			} else {
+				// Defensively clear any host-leaked value; restore on cleanup.
+				prev, had := os.LookupEnv("OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY")
+				_ = os.Unsetenv("OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY")
+				t.Cleanup(func() {
+					if had {
+						_ = os.Setenv("OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY", prev)
+					}
+				})
+			}
+			cfg := loadConfig()
+			if cfg.OSInsecureSkipVerify != tc.want {
+				t.Fatalf("OSInsecureSkipVerify = %v, want %v (env=%q set=%v)",
+					cfg.OSInsecureSkipVerify, tc.want, tc.val, tc.set)
+			}
+		})
+	}
+}
--- a/modules/messages_search/es_client.go
+++ b/modules/messages_search/es_client.go
@@ -2,8 +2,10 @@ package messages_search
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"net"
+	"net/http"
 	"net/url"
 	"sync"
 	"time"
@@ -101,6 +103,22 @@ func buildESClient(cfg SearchConfig) (*elastic.Client, error) {
 		elastic.SetSniff(false),
 		elastic.SetHealthcheck(true),
 		elastic.SetHealthcheckTimeout(3 * time.Second),
+	}
+	if cfg.OSInsecureSkipVerify {
+		// Self-signed / internal-CA OS clusters in dev / test environments
+		// where the pod's system trust store has no chain to verify against.
+		// Disable cert verification on the HTTP client olivere uses for both
+		// the startup healthcheck and subsequent requests. MUST NOT be set
+		// in production deployments — see config.go::OSInsecureSkipVerify.
+		hc := &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
+			},
+		}
+		if cfg.Timeout > 0 {
+			hc.Timeout = cfg.Timeout
+		}
+		opts = append(opts, elastic.SetHttpClient(hc))
 	}
 	if cfg.OSUsername != "" {
 		opts = append(opts, elastic.SetBasicAuth(cfg.OSUsername, cfg.OSPassword))


lml2468 (approved):
✅ APPROVE — 教科书级的安全开关实现,inline 逐项核实,build/test 过,无 blocking。

OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY（OpenSearch TLS 证书校验跳过开关）——这类开关最该警惕"默认不安全/易误开/静默放宽",本 PR 四个维度都处理对了:

1. ✅ **默认安全**:OSInsecureSkipVerify 默认 false,env 未设/非严格 "true" 都保持 TLS 校验 ON。
2. ✅ **严格字面量匹配**:`== "true"`(非 ParseBool),config_test 8 个 case 锁死契约(TRUE/True/1/yes/false/空 全判 false,仅 "true" 为 true),注释明说防"future helpful 重构成 ParseBool 悄悄拓宽门"。这个测试设计很到位、正是安全开关该有的回归护栏。
3. ✅ **开启告警**:New() 里 OSInsecureSkipVerify=true → Warn 日志(生产误开可从日志发现)。
4. ✅ **实现正确 + 正交**:仅 flag 开时才换成 InsecureSkipVerify 的 http.Client(保留 cfg.Timeout),//nolint:gosec 标注,注释强调 MUST NOT 生产用。与既有 OSInsecureHTTP(管 http:// 明文凭证)是正交维度——skip-verify 管 https TLS 证书,checkTransportSecurity 对 https 直接放行,两开关无交互冲突。自定义 http.Client 无功能丢失(olivere 默认即用 http.DefaultClient,SetSniff 等 client 级 option 不受影响)。

build ./modules/messages_search/ + TestLoadConfig_OSInsecureSkipVerify(8 case)全过。

无 blocking、无 should-fix。默认安全、契约有测试锁、误开有告警、实现正交干净。批准。

—— Octo-PR (lml2468)

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#374 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/374
**Head SHA:** `7537da149d1ae3721d0c094aefb8f61f655d8f36`

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| New env `OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY` parsed correctly | ✅ | `config.go:70` — strict `== "true"` match, consistent with `OSInsecureHTTP` |
| Config field added to `SearchConfig` struct | ✅ | `config.go:32` — `OSInsecureSkipVerify bool` |
| TLS skip-verify applied via custom `http.Client` | ✅ | `es_client.go:107-122` — `tls.Config{InsecureSkipVerify: true}` injected via `elastic.SetHttpClient` |
| WARN log at module init when enabled | ✅ | `api.go:64-68` — loud warning in `New()` |
| Unit tests cover parsing contract | ✅ | `config_test.go` — 8 cases covering unset/empty/true/TRUE/True/1/yes/false |
| `//nolint:gosec` on `InsecureSkipVerify` | ✅ | `es_client.go:115` — appropriate for intentionally gated usage |
| Orthogonal to existing `OSInsecureHTTP` | ✅ | `checkTransportSecurity()` unchanged; two flags control independent concerns |

## 2. Findings

### P2 — Custom `http.Transport` drops `ProxyFromEnvironment`

**File:** `es_client.go:111-116`

The PR creates `&http.Transport{TLSClientConfig: ...}` without setting `Proxy: http.ProxyFromEnvironment`. Go's `http.DefaultTransport` includes proxy support by default. If any deployment routes traffic to OpenSearch through an HTTP proxy (via `HTTP_PROXY`/`HTTPS_PROXY` env vars), enabling `OSInsecureSkipVerify` would silently bypass the proxy.

**Diff-scope:** new (introduced by this PR's custom transport construction).
**Severity rationale:** P2 — the flag targets dev/test environments where proxies are uncommon; production should use proper certs, not this flag. Low real-world impact but worth noting for completeness.
**Suggestion:** Add `Proxy: http.ProxyFromEnvironment` to the `http.Transport` literal to preserve default proxy behavior.

### P2 — Custom transport loses `DefaultTransport` connection-pool and timeout defaults

**File:** `es_client.go:111-116`

The bare `&http.Transport{}` zeroes out `DialContext` (30s connect timeout), `TLSHandshakeTimeout` (10s), `MaxIdleConns` (100), `IdleConnTimeout` (90s), and `ForceAttemptHTTP2`. The `hc.Timeout = cfg.Timeout` (5s default) covers overall request deadline, but individual TCP connect or TLS handshake hangs could consume the entire client timeout before failing.

**Diff-scope:** new (introduced by this PR).
**Severity rationale:** P2 — olivere/elastic has its own healthcheck/retry layer and the client-level timeout provides a backstop. Acceptable for dev/test. Consider cloning `http.DefaultTransport` and overriding only `TLSClientConfig` if this flag might see broader use.

## 3. Suggestions

1. **Clone DefaultTransport instead of bare struct** — a one-liner like:
   ```go
   t := http.DefaultTransport.(*http.Transport).Clone()
   t.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
   ```
   preserves proxy support, connection pool defaults, and all timeout settings. This addresses both P2 findings above.

2. **PR description template** — the PR body still carries unfilled template sections (Summary, Related Issue, Changes, Testing, Checklist). Consider filling these before merge for audit trail quality.

## 4. Additional Observations

- The `checkTransportSecurity()` function correctly remains unchanged. `OSInsecureSkipVerify` (TLS cert verification) and `OSInsecureHTTP` (cleartext credential gating) are orthogonal — no interaction gap.
- The `ESClient()` cache (process-wide singleton keyed on nothing) is documented as correct for the current static-config design. The new flag is read once at init and never mutates, consistent with this contract.
- All four search endpoints (`search_all.go`, `search_messages.go`, `search_media.go`, `search_files.go`) call `ESClient(h.cfg)` — they all benefit from the fix uniformly.

## 5. Data Flow Trace

| Consumed Data | Source | Flows To | Verified |
|---|---|---|---|
| `cfg.OSInsecureSkipVerify` | `loadConfig()` → `os.Getenv("OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY") == "true"` (`config.go:70`) | `buildESClient()` gate (`es_client.go:107`); `New()` WARN log (`api.go:64`) | ✅ — env → struct field → two consumption points, both reachable |
| Custom `http.Client` | Constructed at `es_client.go:111-121` | `elastic.SetHttpClient(hc)` → olivere uses for all HTTP including startup healthcheck | ✅ — `SetHttpClient` is a standard olivere v6 option |
| `cfg.Timeout` | `parseDuration(os.Getenv("OCTO_SEARCH_TIMEOUT"), 5*time.Second)` (`config.go:71`) | `hc.Timeout` (`es_client.go:120`) | ✅ — default 5s > 0, so the `if cfg.Timeout > 0` gate passes in all normal configs |

## 6. R5 Security-Sensitive Checklist

- **C1 — Dual-path parity:** N/A. This PR adds a single new code path (TLS skip-verify) with no symmetric counterpart (no "re-enable verify" path needed — it's a one-time config read at init). No add/remove or subscribe/unsubscribe symmetry required.
- **C2 — Control-flow ordering / nested reuse:** Clear. The `if cfg.OSInsecureSkipVerify` block runs once during `buildESClient()`, before `elastic.NewClient(opts...)`. No nested or repeated invocation. The `//nolint:gosec` annotation is correctly scoped to the single `InsecureSkipVerify` literal.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This change affects the TLS transport layer of the OpenSearch client, not any authorization or access-control boundary. The env var is a server-side deployment knob, not exposed to end users.

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. Two P2 notes on the custom `http.Transport` construction (missing proxy support and lost DefaultTransport defaults) — both low-risk in the intended dev/test context and addressable in a follow-up. The suggested `DefaultTransport.Clone()` pattern would close both in one line.

lml2468:
🔧 更正我前一条 review 的一处判断 + 采纳 @mochashanyao 的 transport 🟡(should-fix,非 blocker,票维持 APPROVE)。

我上一条说"自定义 http.Client 无功能丢失、olivere 默认即用 http.DefaultClient" —— **这句判错了**,核得不够细。mochashanyao 对:

es_client.go:114 的 `&http.Transport{TLSClientConfig: ...}` 是个**裸 transport**,只设了 TLSClientConfig,丢了 http.DefaultTransport 的:
- `Proxy: ProxyFromEnvironment` → HTTP(S)_PROXY 环境代理被绕过
- 连接池/超时:MaxIdleConns(100)/MaxIdleConnsPerHost/IdleConnTimeout(90s)/TLSHandshakeTimeout/ExpectContinueTimeout 等全丢

olivere 默认走 http.DefaultClient → 背后是 http.DefaultTransport(有代理+连接池);本 PR 换成裸 transport 就没了。我之前说"无丢失"是错的——没逐项对比 DefaultTransport vs 裸 transport。

## 严重性:should-fix,非 blocker(故票维持 APPROVE)
这条路径**只在 dev/test 开 OSInsecureSkipVerify 时才走**(生产 MUST NOT 开),且丢的是性能(连接池)+ dev 代理感知,**不影响正确性/安全**(TLS skip 本身那个安全点有默认安全+告警,已审对)。dev/test 影响有限,故 should-fix 不阻塞。

## 修法(mochashanyao 给的,标准做法)
克隆默认 transport 再只覆盖 TLS:
```go
tr := http.DefaultTransport.(*http.Transport).Clone()
tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: true} //nolint:gosec
hc := &http.Client{Transport: tr}
```
保留代理 + 连接池,只改证书校验。建议作者顺手收(一行变三行)。mochashanyao 另两条 🟡(PR 描述模板没填全)也建议补。

诚实说:这是我又一次核得不够到底——"无功能丢失"这种结论本该逐项对比才下,我凭印象说了。mochashanyao 逐项查出代理+连接池,准。票维持 APPROVE(should-fix 不阻塞),但更正我那句错判。

—— Octo-PR (lml2468)

Jerry-Xin (approved):
✅ APPROVED — @7537da1

Clean, well-guarded opt-in for an insecure-TLS dev/test escape hatch. Verified locally: `go build ./modules/messages_search/...` and `go vet` green; the 8-case config-contract test passes.

✅ **The flag is correctly fenced off-by-default and dev/test-only:**
- `OSInsecureSkipVerify` parses via strict literal `os.Getenv(...) == "true"` — `1` / `yes` / `TRUE` / `True` / `""` all leave verification **ON**. `TestLoadConfig_OSInsecureSkipVerify` locks all 8 cases (incl. case-sensitivity), and the test comment explicitly guards against a future `strconv.ParseBool` refactor silently widening the door — good defensive instinct.
- Loud `Warn` at handler init when enabled; clear doc on both the field and the es_client call site marking it MUST stay false in production.
- `//nolint:gosec` on the `InsecureSkipVerify: true` line correctly acknowledges the intentional G402. Consistent with the existing repo pattern (`pkg/redis/options.go` already has `RedisTLSInsecureSkipVerify`) and the sibling `OSInsecureHTTP` flag.
- Applied to the olivere client's HTTP transport (covers both startup healthcheck and subsequent requests), with the configured timeout preserved.

⚠️ **CI red checks are pre-existing repo-infra issues, NOT caused by this PR** — traced both:
- `secret-scan / gitleaks: FAILURE (47 leaks)` — I pulled the job log: **all 47 findings are in pre-existing `_test.go` fixtures** (e.g. `modules/common/api_manager_system_setting_test.go`, `modules/botfather/service_userapikey_test.go`, `pkg/accesslog/accesslog_test.go` — fake `generic-api-key` test data). **Zero findings in any of this PR's 4 files** (`modules/messages_search/{api,config,es_client}.go` + config_test.go introduce no secrets). This is a repo-wide gitleaks allowlist gap for test fixtures, orthogonal to #374. Worth a separate fix (add a gitleaks `.gitleaksignore` / allowlist for `*_test.go` fake keys), but not this PR's responsibility.
- `dependency-review: FAILURE` — log says "Dependency review is not supported on this repository. Please ensure that Dependency graph is enabled." Repo security-analysis setting; #374 changes **no dependencies** (no go.mod/go.sum touched), so this is unrelated to the PR content.

The main Build/Lint/Vet/Test pipeline is SKIPPED here (path-filter), so my local build/vet/test green is the code-side evidence; both CI reds are repo-infra, not regressions.

🟡 Non-blocking — the opt-in is purely operator-trust (an env flag), with no hard prod-guard. The sibling `OSInsecureHTTP` doc notes a build-time *rejection* coupling insecure transport with credential presence; `InsecureSkipVerify` has no equivalent — if set in prod with real OS credentials, those creds flow over a MITM-able channel with only a warn log. Matches the established insecure-flag model (off by default, documented prod-forbidden, loud warn), so not blocking, but a future hardening could refuse `InsecureSkipVerify` when a non-loopback OS addr + credentials are configured, mirroring the OSInsecureHTTP guard.

Code LGTM. Note merge is currently BLOCKED by the two repo-infra CI reds above (gitleaks test-fixture allowlist + dependency-graph setting) — those need a repo/maintainer fix, not a change in this PR.

yujiawei (approved):
# Code Review — PR #374 (octo-server)

**Verdict: APPROVED.** No P0/P1 blockers. The change is correct, default-safe, well-scoped, and consistent with the existing `OCTO_SEARCH_OS_INSECURE_HTTP` escape-hatch precedent in the same file. The items below are non-blocking P2 follow-ups.

Reviewed against head `7537da149d1ae3721d0c094aefb8f61f655d8f36` (base `main`). Build, `go vet`, and the full `modules/messages_search` test suite all pass; the new `TestLoadConfig_OSInsecureSkipVerify` passes 8/8 strict-literal cases.

## What this PR does

Adds `OCTO_SEARCH_OS_INSECURE_SKIP_VERIFY` (default `false`, strict-literal `"true"` to enable). When enabled, `buildESClient` injects a custom `http.Client` with `tls.Config{InsecureSkipVerify: true}` via `elastic.SetHttpClient`, unblocking OpenSearch clusters whose CA isn't in the pod trust store. A loud `WARN` fires at module init so the deviation can't stay enabled silently.

## Security assessment (this PR is security-sensitive)

Disabling TLS verification is inherently a MITM exposure, but the design contains it correctly:

- ✅ **Default-off, opt-in only** via exact string `"true"` — `config.go:69`. Mirrors the parsing of the sibling `OSInsecureHTTP`, and `config_test.go` locks the strict-literal contract so a future refactor to `strconv.ParseBool` can't silently widen the door.
- ✅ **Loud `WARN` at init** — `api.go:64-68`. The deviation is observable.
- ✅ **Containment verified** — the insecure client is a *locally-scoped* `http.Client` passed only to this module's olivere client via `SetHttpClient`. It does **not** mutate the global `http.DefaultTransport`/`http.DefaultClient`. This is the right pattern; a naive insecure-TLS change often leaks skip-verify process-wide, and this PR correctly avoids that.
- ✅ **No prod path regression** — the custom transport is constructed *only* when the flag is true; the secure/prod path is untouched and keeps olivere's `http.DefaultClient` (verified against vendored olivere v6.2.37 `client.go:227`).

Item a human may want to note: there is no runtime hard-gate preventing the flag from being enabled in a production deployment — the `WARN` is the only guardrail. This is acceptable for merge (see P2 #2 below) and consistent with the existing insecure-HTTP toggle, but it's the one residual security property worth a conscious sign-off.

**P2 #1 — Bare `http.Transport` discards `DefaultTransport` tuning (consensus).** `es_client.go:114-116` constructs `&http.Transport{TLSClientConfig: ...}` from scratch. A bare transport defaults to `MaxIdleConnsPerHost = 2` and `IdleConnTimeout = 0` (no idle reaping). Under concurrency > 2 this causes connection churn / `TIME_WAIT` buildup and repeated TLS handshakes (note: not an unbounded socket *leak* — Go closes excess idle conns). Impact is limited because this path runs **only when the flag is on, i.e. dev/test by design** (lowest load), which is why this is P2 and not higher. Recommended fix simultaneously addresses the "drops future transport tuning" concern:

```go
tr := http.DefaultTransport.(*http.Transport).Clone()
tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: true} //nolint:gosec
hc := &http.Client{Transport: tr}
```

**P2 #2 — No runtime prod-mode guard against accidental enablement (defense-in-depth).** A single clearly-named env var flips off TLS verification with no environment check. Today this requires an operator to both set the var *and* ignore the init `WARN`, so it's operator error rather than a code defect — and it matches the accepted precedent of `OCTO_SEARCH_OS_INSECURE_HTTP`. If you want belt-and-suspenders later, the hook already exists: octo-lib's config carries a global `Mode` (debug/release) and `ctx` is already passed into `New(ctx *config.Context)`, so a release-mode hard-fail on this flag is cheap to add.

**P2 #3 — No test exercises the `buildESClient` wiring (consensus).** `config_test.go` covers env parsing but nothing verifies that `OSInsecureSkipVerify=true` actually results in `elastic.SetHttpClient` being applied. A regression could parse the flag correctly yet stop wiring it. A focused construction test (e.g. against an `httptest` self-signed TLS server) would close this gap.

## Notes on the advisory timeout concern (resolved, not a finding)

One advisory pass flagged that `hc.Timeout = cfg.Timeout` is set only on the insecure path, implying prod requests could hang. **This is mooted and is not a finding:** all four search handlers wrap every OS request in `context.WithTimeout(c.Request.Context(), h.cfg.Timeout)` (`search_all/files/media/messages.go:71/84/83/76`), and both the startup and periodic healthchecks are independently bounded by their own contexts plus `SetHealthcheckTimeout(3s)` (`es_client.go:105,131`). No unbounded hang is possible on either path.

## Coverage / blind spots

- **Env-var documentation:** the new var appears nowhere outside the three changed Go files — but neither do the existing `OCTO_SEARCH_OS_*` vars. octo-server has no env-doc surface, so this is repo convention, not a regression.
- **Deployment manifests** live in a separate repo (octo-deployment) not in this checkout; if a manifest ever templates OS env vars, the prod-enablement risk in P2 #2 becomes more reachable. The `WARN` + `needs-human-review` label are adequate mitigations for a dev/test unblock.
- **Cache/flag interaction:** the flag is read once at `loadConfig()` init and never mutates, so the module's "first successful build wins" client cache has no flag-flip race.

This review absorbed two independent advisory passes (a line-level adversarial reviewer and a long-context whole-repo reviewer); both ran successfully and their findings are reflected above. All facts were independently re-verified against the repository at the head SHA.

an9xyz (approved):
✅ APPROVE — reviewed the code myself (api.go / config.go / config_test.go / es_client.go), not just the prior reviews.

**Independently verified:**
- **Orthogonal to `OSInsecureHTTP`**: `checkTransportSecurity()` (unchanged) gates cleartext-HTTP credential exposure; the new flag gates https TLS cert verification. No interaction.
- **`SetHttpClient` + `SetBasicAuth` don't conflict**: independent olivere options; basic-auth is a per-request header, the custom client only swaps the transport. No functional loss.
- **Cache contract sound**: flag is read once at init and never mutates, so the "first successful build wins" client cache has no flag-flip race.
- Default-off via strict literal `== "true"`, locked by an 8-case test that explicitly guards against a future `ParseBool` widening; loud WARN at init.

**Non-blocking follow-ups (P2):**
1. Bare `&http.Transport{}` drops `DefaultTransport` tuning (no `Proxy: http.ProxyFromEnvironment`, `MaxIdleConnsPerHost`→2, no dial/TLS-handshake/idle timeouts). `http.DefaultTransport.(*http.Transport).Clone()` then override only `TLSClientConfig` closes this in one line. Only on the flag-on (dev/test) path, so P2.
2. **Asymmetric guard worth noting**: when `OSUsername != "" && OSInsecureSkipVerify=true`, basic-auth credentials traverse a non-cert-verified TLS channel (MITM-interceptable). `OSInsecureHTTP` has a build-time *rejection* guard for the credential+cleartext combo; skip-verify has only the WARN. Acceptable here (off-by-default, documented prod-forbidden, consistent with the existing insecure-flag model) — a future hardening could refuse skip-verify when non-loopback OS creds are configured in release mode.

No P0/P1. Code LGTM.

PR #371, opened 2026-06-14
feat(i18n): migrate file, channel, conversation_ext, user-tail, base/app, notify to the localized error envelope

an9xyz:
## What & why

Continues the Phase 2.1 i18n migration by clearing the long tail of legacy error responses in **6 modules** so they no longer bypass the localized error envelope. Each migrated handler now returns a registered `pkg/errcode` code via `httperr.ResponseErrorL` / `ResponseErrorLWithStatus` instead of raw `c.ResponseError` / `AbortWithStatusJSON` / hard-coded zh-CN strings.

**29 new error codes** registered (+ regenerated en-US markers + zh-CN translations), each with guard + helper-render tests. `make i18n-extract-check` and `make i18n-lint` pass.

## Modules migrated

| Module | Sites → 0 | Notable |
|---|---|---|
| **user** (`api.go`) | 8 | 4 avatar-ownership guards returned a **hard-coded zh-CN 403** (`无权限修改该用户头像`) → anti-enum `ErrUserAvatarUpdateForbidden`; verify-token/bot 401s → shared `auth.token_invalid` (real 401 preserved, reason logged). Guard now also bans `AbortWithStatusJSON`; baseline 8→0. |
| **base/app** (`api.go`) | 2 | `GET /v1/apps/:app_id` no longer leaks raw `GetApp` errors → `ErrAppNotFound` (404) / `ErrAppDisabled` (403). |
| **notify** (`api.go`) | 7 | Internal service API → `ResponseErrorLWithStatus` (preserves real wire status for service callers): anti-enum 401, request-invalid, batch-limit, deliver-failed (500, Internal). baseline 2→0. |
| **file** (`api.go`) | 31 | Upload/download paths: invalid-path, too-large, type-unsupported, content-mismatch + 5 Internal 500s. |
| **channel** (`api.go`, `api_storyline.go`) | 31 | request-invalid, forbidden, not-found, storyline-group-only + query/store/send-failed (Internal). |
| **conversation_ext** (`api.go`) | 30 | `/v1/follow`: validation, follow-forbidden, category-forbidden + client-recoverable `version_conflict` (409) / `sort_target_not_found` (404) + follow/unfollow/sort-update-failed (Internal). |

## Invariants honored

- **5xx ⟺ `Internal: true`** on every internal code; the renderer hides the message and each call site keeps/adds a `zap.Error` log so ops debug from logs.
- **Anti-enumeration**: auth / permission failures collapse to a single generic code; the specific reason goes to logs only.
[...]

PR #375, opened 2026-06-15
feat(app_bot): delegate platform bot management to admin + localize the error/response surface

an9xyz:
## What

Two related changes to `modules/app_bot`, driven by a report that `/v1/admin/app_bot` was inaccessible to `admin` and leaked a raw, unlocalized forbidden string.

### 1. Delegate platform bot management to `admin` (operations)
All 9 `/v1/admin/app_bot` platform handlers (create / list / detail / update / delete / rotate-token / reveal-token / publish / unpublish) were `CheckLoginRoleIsSuperAdmin` since the module's inception. Operations needs to manage platform bots, so they now use `CheckLoginRole` (admin∪superAdmin). The space-scoped `/v1/space/:space_id/app_bot` surface is unchanged (still gated by space admin/owner via `checkSpaceAdmin`).

### 2. Localize the entire app_bot error/response surface (i18n)
Every handler returned raw, unlocalized strings — `c.ResponseError(errors.New("..."))`, `c.AbortWithStatusJSON(403, err.Error())`, `c.JSON(404, {"msg":"bot not found"})` and Chinese literals — none of which were language-negotiated. They now route through a registered `errcode` + the i18n envelope, with **correct semantics and status preservation**:

| Class | New code | Status |
|---|---|---|
| validation (bad body, id format/reserved, robot_uid) | `err.server.app_bot.request_invalid` / `id_invalid` | 400 |
| not found (missing / unpublished bot) | `err.server.app_bot.not_found` | 404 |
| conflict (id in use, token-rotation race) | `id_conflict` / `token_rotation_conflict` | 409 |
| forbidden (all role + space-membership guards) | `err.shared.auth.forbidden` | 403 |
| internal (query/store/IM-token/invariant failures) | `query_failed` / `store_failed` / `im_token_failed` / `internal` (`Internal=true`) | 500 |

The two apply-flow **success** bodies (already-friend / auto-approved) are localized via a small `msgtmpl` catalog (`templates/{en-US,zh-CN}/apply.tmpl`) rendered in the request language.

## ⚠️ Client-visible behavior changes (frontend, please note)

These are intentional (correct semantics), but the management console branches on status:

- **Platform `/v1/admin/app_bot` forbidden: `400` → `403`** (unified with the space guards; the old 400 was only a legacy artifact).
- **Internal failures: `400` → `500`, and the raw cause is no longer returned** — clients now get a generic localized message (`Failed to query/update data.`); the cause is logged server-side (`zap.Error`).
- **Not-found / conflict now return real `404` / `409`** (some were `400`).
- `admin` (not just `superAdmin`) can now reach `/v1/admin/app_bot`, **including reveal-token / rotate-token** (confirmed expected with the maintainer).

## Anti-enumeration / safety
- Forbidden collapses to one shared code (the specific role/membership reason stays in logs only).
- Not-found collapses "missing" and "unpublished" (apply flow), so existence isn't probeable.
- `Internal=true` codes hide the message; every such site logs the cause first (verified).

## Testing
`modules/app_bot` full suite passes (sqlmock route harness + renderer harness, no MySQL needed for the new tests):
- `TestPlatformAppBot_AdminAllowed` — admin passes the platform gate
- `TestPlatformAppBot_PlainUserForbiddenLocalized` / `TestSpaceAppBot_ForbiddenLocalized` — localized 403, no raw string
- `TestPlatformAppBot_NotFoundLocalized` — real 404, localized
- `TestAppBotRespondHelpers` — every responder renders its registered code at the correct status (400/404/409/403/500)
- `TestAppBotNoLegacyResponseError` — **source guard**: no raw response regressions

`go build ./...`, `go vet`, `gofmt` clean; `make i18n-extract-check` + `i18n-lint` + unregistered-code pass; `pkg/errcode` / `pkg/i18n` package tests pass (codes register with valid en-US markers + zh-CN translations). The direct-error-response lint baseline for `app_bot.go` is **ratcheted 9 → 0**.

## Out of scope
The default welcome **IM** message (outbound WuKongIM content) needs recipient-language resolution, not request-language — left as-is.

---

### DIFF
--- a/modules/app_bot/api_i18n.go
+++ b/modules/app_bot/api_i18n.go
@@ -48,6 +48,14 @@ func respondAppBotNotFound(c *wkhttp.Context) {
 	httperr.ResponseErrorLWithStatus(c, errcode.ErrAppBotNotFound, nil, nil)
 }
 
+// respondAppBotNotFoundPinned renders the same not-found code at the legacy
+// fixed-400 wire status (D14), for the user-facing /v1/app_bot/apply endpoint
+// whose SDK clients may branch on 400. The management-console paths use
+// respondAppBotNotFound (real 404) instead.
+func respondAppBotNotFoundPinned(c *wkhttp.Context) {
+	httperr.ResponseErrorL(c, errcode.ErrAppBotNotFound, nil, nil)
+}
+
 // respondAppBotIDConflict renders the 409 for a create colliding with an in-use id.
 func respondAppBotIDConflict(c *wkhttp.Context) {
 	httperr.ResponseErrorLWithStatus(c, errcode.ErrAppBotIDConflict, nil, nil)
--- a/modules/app_bot/api_i18n.go
+++ b/modules/app_bot/api_i18n.go
@@ -20,12 +20,24 @@ func mustLookupSharedCode(id string) codes.Code {
 	return c
 }
 
-// respondAppBotForbidden renders the localized shared 403 envelope for the
-// /v1/admin/app_bot role guards. It replaces the legacy c.ResponseError(err)
-// that forwarded wkhttp's raw, unlocalized framework string
-// ("该用户无权执行此操作") straight onto the wire. The guards collapse to one
-// generic forbidden code (anti-enumeration): the specific role reason stays in
-// logs, never on the client.
+// respondAppBotForbidden renders the localized shared forbidden envelope for the
+// platform /v1/admin/app_bot role guards. It replaces the legacy
+// c.ResponseError(err) that forwarded wkhttp's raw, unlocalized framework string
+// ("该用户无权执行此操作") straight onto the wire. ResponseErrorL keeps the
+// legacy fixed-400 wire status these guards already returned (D14). The guards
+// collapse to one generic forbidden code (anti-enumeration): the specific role
+// reason stays in logs, never on the client.
 func respondAppBotForbidden(c *wkhttp.Context) {
 	httperr.ResponseErrorL(c, errSharedForbidden, nil, nil)
+}
+
+// respondAppBotSpaceForbidden renders the same localized shared forbidden
+// envelope for the space-scoped /v1/space/:space_id/app_bot guards, which
+// replaced a raw c.AbortWithStatusJSON(403, err.Error()) that leaked
+// checkSpaceAdmin's English string ("no permission: requires space admin").
+// Unlike the platform guards, these already returned a real 403, so this uses
+// ResponseErrorLWithStatus to preserve that status (the code's real 403) rather
+// than collapse it to the legacy 400 — no client-visible status change.
+func respondAppBotSpaceForbidden(c *wkhttp.Context) {
+	httperr.ResponseErrorLWithStatus(c, errSharedForbidden, nil, nil)
 }
--- a/modules/app_bot/api_i18n.go
+++ b/modules/app_bot/api_i18n.go
@@ -1,0 +1,31 @@
+package app_bot
+
+import (
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
+	"github.com/Mininglamp-OSS/octo-server/pkg/i18n/codes"
+)
+
+// errSharedForbidden caches the shared 403 code used by the platform-level
+// (/v1/admin/app_bot) role guards. Looked up at package init so a missing
+// registration panics loudly here rather than rendering an empty envelope at
+// request time.
+var errSharedForbidden = mustLookupSharedCode("err.shared.auth.forbidden")
+
+func mustLookupSharedCode(id string) codes.Code {
+	c, ok := codes.Lookup(id)
+	if !ok {
+		panic("modules/app_bot: shared code not registered: " + id)
+	}
+	return c
+}
+
+// respondAppBotForbidden renders the localized shared 403 envelope for the
+// /v1/admin/app_bot role guards. It replaces the legacy c.ResponseError(err)
+// that forwarded wkhttp's raw, unlocalized framework string
+// ("该用户无权执行此操作") straight onto the wire. The guards collapse to one
+// generic forbidden code (anti-enumeration): the specific role reason stays in
+// logs, never on the client.
+func respondAppBotForbidden(c *wkhttp.Context) {
+	httperr.ResponseErrorL(c, errSharedForbidden, nil, nil)
+}
--- a/modules/app_bot/api_i18n.go
+++ b/modules/app_bot/api_i18n.go
@@ -2,42 +2,77 @@ package app_bot
 
 import (
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
 	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
-	"github.com/Mininglamp-OSS/octo-server/pkg/i18n/codes"
+	"github.com/Mininglamp-OSS/octo-server/pkg/i18n"
 )
 
-// errSharedForbidden caches the shared 403 code used by the platform-level
-// (/v1/admin/app_bot) role guards. Looked up at package init so a missing
-// registration panics loudly here rather than rendering an empty envelope at
-// request time.
-var errSharedForbidden = mustLookupSharedCode("err.shared.auth.forbidden")
+// This file localizes the app_bot error responses. Every handler used to return
+// raw, unlocalized strings — c.ResponseError(errors.New("...")),
+// c.AbortWithStatusJSON(403, err.Error()) and c.JSON(40x, {"msg": "..."}) —
+// which leaked English/Chinese framework text straight onto the wire and could
+// not be language-negotiated. They now route through these helpers onto a
+// registered errcode + the i18n envelope. Status-preserving codes (404/409/403/
+// 500) use ResponseErrorLWithStatus so the console keeps branching on the real
+// wire status; validation stays at 400.
 
-func mustLookupSharedCode(id string) codes.Code {
-	c, ok := codes.Lookup(id)
-	if !ok {
-		panic("modules/app_bot: shared code not registered: " + id)
-	}
-	return c
-}
-
-// respondAppBotForbidden renders the localized shared forbidden envelope for the
-// platform /v1/admin/app_bot role guards. It replaces the legacy
-// c.ResponseError(err) that forwarded wkhttp's raw, unlocalized framework string
-// ("该用户无权执行此操作") straight onto the wire. ResponseErrorL keeps the
-// legacy fixed-400 wire status these guards already returned (D14). The guards
-// collapse to one generic forbidden code (anti-enumeration): the specific role
-// reason stays in logs, never on the client.
+// respondAppBotForbidden renders the localized shared 403 for every app_bot
+// authorization guard: the platform /v1/admin/app_bot system-role gates, the
+// space-scoped checkSpaceAdmin gates, and the apply-flow space-membership check.
+// All collapse to one generic forbidden code (anti-enumeration) — the specific
+// role/membership reason stays in logs, never on the client.
 func respondAppBotForbidden(c *wkhttp.Context) {
-	httperr.ResponseErrorL(c, errSharedForbidden, nil, nil)
+	httperr.ResponseErrorLWithStatus(c, errcode.ErrSharedForbidden, nil, nil)
 }
 
-// respondAppBotSpaceForbidden renders the same localized shared forbidden
-// envelope for the space-scoped /v1/space/:space_id/app_bot guards, which
-// replaced a raw c.AbortWithStatusJSON(403, err.Error()) that leaked
-// checkSpaceAdmin's English string ("no permission: requires space admin").
-// Unlike the platform guards, these already returned a real 403, so this uses
-// ResponseErrorLWithStatus to preserve that status (the code's real 403) rather
-// than collapse it to the legacy 400 — no client-visible status change.
-func respondAppBotSpaceForbidden(c *wkhttp.Context) {
-	httperr.ResponseErrorLWithStatus(c, errSharedForbidden, nil, nil)
+// respondAppBotRequestInvalid covers malformed / empty request input (BindJSON
+// failure, invalid robot_uid, empty update). An empty field is omitted so the
+// renderer does not surface a noisy empty key.
+func respondAppBotRequestInvalid(c *wkhttp.Context, field string) {
+	details := i18n.Details{}
+	if field != "" {
+		details["field"] = field
+	}
+	httperr.ResponseErrorL(c, errcode.ErrAppBotRequestInvalid, nil, details)
+}
+
+// respondAppBotIDInvalid covers a bot id failing the format rule or colliding
+// with a reserved id.
+func respondAppBotIDInvalid(c *wkhttp.Context) {
+	httperr.ResponseErrorL(c, errcode.ErrAppBotIDInvalid, nil, nil)
+}
+
+// respondAppBotNotFound renders the status-preserving 404 for a missing or
+// scope-mismatched bot.
+func respondAppBotNotFound(c *wkhttp.Context) {
+	httperr.ResponseErrorLWithStatus(c, errcode.ErrAppBotNotFound, nil, nil)
+}
+
+// respondAppBotIDConflict renders the 409 for a create colliding with an in-use id.
+func respondAppBotIDConflict(c *wkhttp.Context) {
+	httperr.ResponseErrorLWithStatus(c, errcode.ErrAppBotIDConflict, nil, nil)
+}
+
+// respondAppBotTokenRotationConflict renders the 409 for a lost token-rotation race.
+func respondAppBotTokenRotationConflict(c *wkhttp.Context) {
+	httperr.ResponseErrorLWithStatus(c, errcode.ErrAppBotTokenRotationConflict, nil, nil)
+}
+
+// respondAppBotQueryFailed / StoreFailed / IMTokenFailed / Internal render the
+// status-preserving 500. Internal=true hides the message — callers MUST log the
+// underlying err (zap.Error) with context before calling these.
+func respondAppBotQueryFailed(c *wkhttp.Context) {
+	httperr.ResponseErrorLWithStatus(c, errcode.ErrAppBotQueryFailed, nil, nil)
+}
+
+func respondAppBotStoreFailed(c *wkhttp.Context) {
+	httperr.ResponseErrorLWithStatus(c, errcode.ErrAppBotStoreFailed, nil, nil)
+}
+
+func respondAppBotIMTokenFailed(c *wkhttp.Context) {
+	httperr.ResponseErrorLWithStatus(c, errcode.ErrAppBotIMTokenFailed, nil, nil)
+}
+
+func respondAppBotInternal(c *wkhttp.Context) {
+	httperr.ResponseErrorLWithStatus(c, errcode.ErrAppBotInternal, nil, nil)
 }
--- a/modules/app_bot/api_i18n_test.go
+++ b/modules/app_bot/api_i18n_test.go
@@ -52,6 +52,30 @@ func TestAppBotNoLegacyResponseError(t *testing.T) {
 	}
 }
 
+// TestAppBotNotFoundPinnedIsWire400 pins the apply-path responder: it carries the
+// real 404 in the envelope but keeps the legacy fixed-400 wire status (D14), so
+// existing /v1/app_bot/apply SDK clients that branch on 400 don't break.
+func TestAppBotNotFoundPinnedIsWire400(t *testing.T) {
+	r := appBotHelperHarness(respondAppBotNotFoundPinned)
+	req := httptest.NewRequest(http.MethodGet, "/probe", nil)
+	rec := httptest.NewRecorder()
+	r.ServeHTTP(rec, req)
+
+	if rec.Code != http.StatusBadRequest {
+		t.Fatalf("wire status = %d, want 400 (D14 pinned for the apply path); body=%s", rec.Code, rec.Body.String())
+	}
+	var env appBotEnvelope
+	if err := json.Unmarshal(rec.Body.Bytes(), &env); err != nil {
+		t.Fatalf("decode envelope: %v; body=%s", err, rec.Body.String())
+	}
+	if env.Error.Code != "err.server.app_bot.not_found" {
+		t.Fatalf("error.code = %q, want err.server.app_bot.not_found", env.Error.Code)
+	}
+	if env.Error.HTTPStatus != http.StatusNotFound {
+		t.Fatalf("error.http_status = %d, want 404 (envelope keeps the real status)", env.Error.HTTPStatus)
+	}
+}
+
 // appBotEnvelope is the partial shape of an httperr.ResponseErrorL* response.
 type appBotEnvelope struct {
 	Error struct {
--- a/modules/app_bot/api_i18n_test.go
+++ b/modules/app_bot/api_i18n_test.go
@@ -1,0 +1,114 @@
+package app_bot
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/pkg/i18n"
+)
+
+// TestAppBotNoLegacyResponseError pins that the module's HTTP surface renders
+// every error through the i18n envelope (httperr.ResponseErrorL* +
+// errcode.ErrAppBot* / shared codes) and never regresses to octo-lib raw
+// responses. Comments are stripped first so the migration breadcrumbs in
+// api_i18n.go (which name the old c.ResponseError / c.AbortWithStatusJSON forms)
+// don't trip the guard. Add any new handler file to the list below.
+func TestAppBotNoLegacyResponseError(t *testing.T) {
+	files := []string{"app_bot.go", "api_i18n.go", "messages.go"}
+	banned := []string{
+		".ResponseError(",
+		".ResponseErrorf(",
+		".ResponseErrorWithStatus(",
+		".AbortWithStatusJSON(",
+		".AbortWithStatus(",
+		"c.Response(\"",
+	}
+	for _, f := range files {
+		t.Run(f, func(t *testing.T) {
+			data, err := os.ReadFile(f)
+			if err != nil {
+				t.Fatalf("read %s: %v", f, err)
+			}
+			var clean strings.Builder
+			for _, line := range strings.Split(string(data), "\n") {
+				if idx := strings.Index(line, "//"); idx >= 0 {
+					line = line[:idx]
+				}
+				clean.WriteString(line)
+				clean.WriteByte('\n')
+			}
+			cleaned := clean.String()
+			for _, b := range banned {
+				if strings.Contains(cleaned, b) {
+					t.Fatalf("modules/app_bot/%s must render errors via httperr.ResponseErrorL* / errcode.ErrAppBot*, not legacy %s", f, b)
+				}
+			}
+		})
+	}
+}
+
+// appBotEnvelope is the partial shape of an httperr.ResponseErrorL* response.
+type appBotEnvelope struct {
+	Error struct {
+		Code       string `json:"code"`
+		HTTPStatus int    `json:"http_status"`
+	} `json:"error"`
+}
+
+func appBotHelperHarness(probe func(c *wkhttp.Context)) *wkhttp.WKHttp {
+	r := wkhttp.New()
+	r.SetErrorRenderer(i18n.NewErrorRenderer(i18n.NewLocalizer(i18n.DefaultLanguage)))
+	r.GET("/probe", probe)
+	return r
+}
+
+// TestAppBotRespondHelpers asserts each responder renders its registered code at
+// the correct wire status: validation pins 400 (D14), while not-found/conflict/
+// forbidden/internal preserve the code's real status via ResponseErrorLWithStatus.
+// No DB/Redis needed — it only exercises the renderer.
+func TestAppBotRespondHelpers(t *testing.T) {
+	cases := []struct {
+		name       string
+		probe      func(c *wkhttp.Context)
+		wantStatus int
+		wantCodeID string
+	}{
+		{"requestInvalid", func(c *wkhttp.Context) { respondAppBotRequestInvalid(c, "") }, http.StatusBadRequest, "err.server.app_bot.request_invalid"},
+		{"idInvalid", respondAppBotIDInvalid, http.StatusBadRequest, "err.server.app_bot.id_invalid"},
+		{"notFound", respondAppBotNotFound, http.StatusNotFound, "err.server.app_bot.not_found"},
+		{"idConflict", respondAppBotIDConflict, http.StatusConflict, "err.server.app_bot.id_conflict"},
+		{"tokenRotationConflict", respondAppBotTokenRotationConflict, http.StatusConflict, "err.server.app_bot.token_rotation_conflict"},
+		{"queryFailed", respondAppBotQueryFailed, http.StatusInternalServerError, "err.server.app_bot.query_failed"},
+		{"storeFailed", respondAppBotStoreFailed, http.StatusInternalServerError, "err.server.app_bot.store_failed"},
+		{"imTokenFailed", respondAppBotIMTokenFailed, http.StatusInternalServerError, "err.server.app_bot.im_token_failed"},
+		{"internal", respondAppBotInternal, http.StatusInternalServerError, "err.server.app_bot.internal"},
+		{"forbidden", respondAppBotForbidden, http.StatusForbidden, "err.shared.auth.forbidden"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			r := appBotHelperHarness(tc.probe)
+			req := httptest.NewRequest(http.MethodGet, "/probe", nil)
+			rec := httptest.NewRecorder()
+			r.ServeHTTP(rec, req)
+
+			if rec.Code != tc.wantStatus {
+				t.Fatalf("status = %d, want %d; body=%s", rec.Code, tc.wantStatus, rec.Body.String())
+			}
+			var env appBotEnvelope
+			if err := json.Unmarshal(rec.Body.Bytes(), &env); err != nil {
+				t.Fatalf("decode envelope: %v; body=%s", err, rec.Body.String())
+			}
+			if env.Error.Code != tc.wantCodeID {
+				t.Fatalf("error.code = %q, want %q", env.Error.Code, tc.wantCodeID)
+			}
+			if env.Error.HTTPStatus != tc.wantStatus {
+				t.Fatalf("error.http_status = %d, want %d", env.Error.HTTPStatus, tc.wantStatus)
+			}
+		})
+	}
+}
--- a/modules/app_bot/app_bot.go
+++ b/modules/app_bot/app_bot.go
@@ -107,7 +107,7 @@ func NewAppBot(ctx *config.Context) *AppBot {
 
 // Route registers all App Bot management routes.
 func (ab *AppBot) Route(r *wkhttp.WKHttp) {
-	// Platform Admin API (requires login, super admin check in handlers)
+	// Platform Admin API (requires login, admin∪superAdmin role check in handlers)
 	adminAPI := r.Group("/v1/admin/app_bot", ab.ctx.AuthMiddleware(r))
 	{
 		adminAPI.POST("", ab.createPlatformBot)
@@ -158,6 +158,20 @@ func (ab *AppBot) checkSpaceAdmin(c *wkhttp.Context, spaceID string) error {
 		return errors.New("no permission: requires space admin")
 	}
 	return nil
+}
+
+// botInRouteScope reports whether bot belongs to the route it was reached
+// through, so a handler that loaded a bot by its (global) id can reject one that
+// lives outside the caller's authority. The platform route (/v1/admin/app_bot,
+// spaceID=="") manages ONLY platform-scoped bots; the space route only its own
+// space's bots. Without the platform-side half of this check an admin on the
+// platform route could read/rotate any space's bot token by global id (a
+// cross-tenant IDOR) — the space route already enforced its half.
+func botInRouteScope(bot *appBotModel, spaceID string) bool {
+	if spaceID == "" {
+		return bot.Scope == "platform"
+	}
+	return bot.Scope == "space" && bot.SpaceID == spaceID
 }
 
 // ==================== Registry ====================
@@ -482,12 +496,15 @@ func (ab *AppBot) getBotDetail(c *wkhttp.Context) {
 	}
 
 	bot, err := ab.db.queryBotByID(id)
+	if err != nil {
+		ab.Error("query app_bot failed", zap.Error(err), zap.String("id", id))
+	}
 	if err != nil || bot == nil {
 		respondAppBotNotFound(c)
 		return
 	}
 
-	if spaceID != "" && (bot.Scope != "space" || bot.SpaceID != spaceID) {
+	if !botInRouteScope(bot, spaceID) {
 		respondAppBotNotFound(c)
 		return
 	}
@@ -543,16 +560,17 @@ func (ab *AppBot) updateBot(c *wkhttp.Context) {
 		return
 	}
 
-	if spaceID != "" {
-		existing, qerr := ab.db.queryBotByID(id)
-		if qerr != nil || existing == nil {
-			respondAppBotNotFound(c)
-			return
-		}
-		if existing.Scope != "space" || existing.SpaceID != spaceID {
-			respondAppBotNotFound(c)
-			return
-		}
+	existing, qerr := ab.db.queryBotByID(id)
+	if qerr != nil {
+		ab.Error("query app_bot failed", zap.Error(qerr), zap.String("id", id))
+	}
+	if qerr != nil || existing == nil {
+		respondAppBotNotFound(c)
+		return
+	}
+	if !botInRouteScope(existing, spaceID) {
+		respondAppBotNotFound(c)
+		return
 	}
 
 	updates := make(map[string]interface{})
@@ -622,12 +640,15 @@ func (ab *AppBot) deleteBot(c *wkhttp.Context) {
 	}
 
 	bot, err := ab.db.queryBotByID(id)
+	if err != nil {
+		ab.Error("query app_bot failed", zap.Error(err), zap.String("id", id))
+	}
 	if err != nil || bot == nil {
 		respondAppBotNotFound(c)
 		return
 	}
 
-	if spaceID != "" && (bot.Scope != "space" || bot.SpaceID != spaceID) {
+	if !botInRouteScope(bot, spaceID) {
 		respondAppBotNotFound(c)
 		return
 	}
@@ -682,12 +703,15 @@ func (ab *AppBot) rotateToken(c *wkhttp.Context) {
 	}
 
 	bot, err := ab.db.queryBotByID(id)
+	if err != nil {
+		ab.Error("query app_bot failed", zap.Error(err), zap.String("id", id))
+	}
 	if err != nil || bot == nil {
 		respondAppBotNotFound(c)
 		return
 	}
 
-	if spaceID != "" && (bot.Scope != "space" || bot.SpaceID != spaceID) {
+	if !botInRouteScope(bot, spaceID) {
 		respondAppBotNotFound(c)
 		return
 	}
@@ -772,12 +796,15 @@ func (ab *AppBot) revealToken(c *wkhttp.Context) {
 	}
 
 	bot, err := ab.db.queryBotByID(id)
+	if err != nil {
+		ab.Error("query app_bot failed", zap.Error(err), zap.String("id", id))
+	}
 	if err != nil || bot == nil {
 		respondAppBotNotFound(c)
 		return
 	}
 
-	if spaceID != "" && (bot.Scope != "space" || bot.SpaceID != spaceID) {
+	if !botInRouteScope(bot, spaceID) {
 		respondAppBotNotFound(c)
 		return
 	}
@@ -810,12 +837,15 @@ func (ab *AppBot) publishBot(c *wkhttp.Context) {
 	}
 
 	bot, err := ab.db.queryBotByID(id)
+	if err != nil {
+		ab.Error("query app_bot failed", zap.Error(err), zap.String("id", id))
+	}
 	if err != nil || bot == nil {
 		respondAppBotNotFound(c)
 		return
 	}
 
-	if spaceID != "" && (bot.Scope != "space" || bot.SpaceID != spaceID) {
+	if !botInRouteScope(bot, spaceID) {
 		respondAppBotNotFound(c)
 		return
 	}
@@ -865,12 +895,15 @@ func (ab *AppBot) unpublishBot(c *wkhttp.Context) {
 	}
 
 	bot, err := ab.db.queryBotByID(id)
+	if err != nil {
+		ab.Error("query app_bot failed", zap.Error(err), zap.String("id", id))
+	}
 	if err != nil || bot == nil {
 		respondAppBotNotFound(c)
 		return
 	}
 
-	if spaceID != "" && (bot.Scope != "space" || bot.SpaceID != spaceID) {
+	if !botInRouteScope(bot, spaceID) {
 		respondAppBotNotFound(c)
 		return
 	}
@@ -993,7 +1026,7 @@ func (ab *AppBot) applyBot(c *wkhttp.Context) {
 		return
 	}
 	if bot == nil || bot.Status != StatusPublished {
-		respondAppBotNotFound(c)
+		respondAppBotNotFoundPinned(c)
 		return
 	}
 
--- a/modules/app_bot/app_bot.go
+++ b/modules/app_bot/app_bot.go
@@ -1029,7 +1029,7 @@ func (ab *AppBot) applyBot(c *wkhttp.Context) {
 		return
 	}
 	if isFriend {
-		c.Response(gin.H{"status": "approved", "message": "\u5df2\u7ecf\u662f\u597d\u53cb\u4e86"})
+		c.Response(gin.H{"status": "approved", "message": ab.localizedMessage(c, "already_friend")})
 		return
 	}
 
@@ -1132,7 +1132,7 @@ func (ab *AppBot) applyBot(c *wkhttp.Context) {
 		config.PersonalMsgOptions{Header: config.MsgHeader{RedDot: 1}},
 	))
 
-	c.Response(gin.H{"status": "approved", "message": "\u5df2\u81ea\u52a8\u901a\u8fc7\uff0c\u53ef\u4ee5\u5f00\u59cb\u804a\u5929"})
+	c.Response(gin.H{"status": "approved", "message": ab.localizedMessage(c, "auto_approved")})
 }
 
 // fixFriendVersion updates friend version for SDK incremental sync.
--- a/modules/app_bot/app_bot.go
+++ b/modules/app_bot/app_bot.go
@@ -294,7 +294,7 @@ func (ab *AppBot) createPlatformBot(c *wkhttp.Context) {
 func (ab *AppBot) createSpaceBot(c *wkhttp.Context) {
 	spaceID := c.Param("space_id")
 	if err := ab.checkSpaceAdmin(c, spaceID); err != nil {
-		c.AbortWithStatusJSON(http.StatusForbidden, gin.H{"msg": err.Error()})
+		respondAppBotSpaceForbidden(c)
 		return
 	}
 	ab.createBot(c, "space", spaceID)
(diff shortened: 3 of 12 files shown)


mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#375 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/375
**Head SHA:** `376c0054eee61740d9278d9e5f5629587f55b936`
**Title:** feat(app_bot): delegate platform bot management to admin + localize the error/response surface

---

## 1. Verification Summary

| Item | Status | Evidence |
|---|---|---|
| Authorization widening (superAdmin→admin∪superAdmin) | ✅ Consistent across all 9 platform handlers | `app_bot.go` lines 286, 418, 478, 530, 618, 678, 768, 806, 861 — all `CheckLoginRoleIsSuperAdmin` → `CheckLoginRole` |
| Space-scoped paths untouched | ✅ | `checkSpaceAdmin` (line 149) unchanged; space handlers still use it |
| i18n migration completeness | ✅ | Baseline ratcheted 9→0; `TestAppBotNoLegacyResponseError` source guard enforces no regressions |
| Error code registration | ✅ | All 9 codes in `pkg/errcode/app_bot.go` register via `codes.Register()` with valid IDs, statuses, and messages |
| zh-CN translations | ✅ | All 9 codes translated in `active.zh-CN.toml`; en-US markers in `i18nmarkers/server/active.en-US.toml` |
| Internal=true codes log before respond | ✅ | All 14 call sites verified: `ab.Error(..., zap.Error(err))` precedes every `respondAppBot{Query,Store,IMToken,Internal}Failed(c)` |
| Status preservation | ✅ | `ResponseErrorLWithStatus` used for 404/409/403/500; `ResponseErrorL` (fixed-400) used only for validation codes |
| Anti-enumeration | ✅ | Forbidden collapses to one shared code; not-found collapses missing+unpublished; Internal=true hides messages |
| msgtmpl completeness | ✅ | `MustNew` enforces translation matrix at boot; both `en-US/apply.tmpl` and `zh-CN/apply.tmpl` define `already_friend` + `auto_approved` |

## 2. Findings

### F1 — Authorization scope widening: `admin` now reaches sensitive platform endpoints [P2 — intentional, documented]

**Diff-scope:** new (this PR introduces the change deliberately).

All 9 `/v1/admin/app_bot` handlers moved from `CheckLoginRoleIsSuperAdmin()` (superAdmin-only) to `CheckLoginRole()` (admin∪superAdmin). Verified against octo-lib source (`pkg/wkhttp/http.go` lines 180–199):

- `CheckLoginRole` allows `{"admin", "superAdmin"}`
- `CheckLoginRoleIsSuperAdmin` allows `{"superAdmin"}` only

This means `admin` users can now: create/delete platform bots, **reveal plaintext bot tokens**, rotate tokens, publish/unpublish bots (making them discoverable to all users).

**Assessment:** The PR body explicitly documents this as intentional ("confirmed expected with the maintainer"). No middleware-level role gate exists at the route-group level (only `AuthMiddleware` for authentication), so the handler-level check is the sole authorization — and it's correctly applied at all 9 entry points. Role resolution is per-request via `RoleService.ResolveRole()` (Redis cache → DB fallback), so demoted admins lose access promptly.

**Recommendation:** Ensure the operations team's admin roster is audited before merge. Consider adding a CHANGELOG entry for the authorization widening.

### F2 — Combined `err != nil || bot == nil` paths don't log query errors before 404 [P2 — pre-existing]

**Diff-scope:** pre-existing (pattern existed before this PR; PR preserves it while changing the response format).

In `getBotDetail`, `updateBot`, `deleteBot`, `rotateToken`, `revealToken`, `publishBot`, `unpublishBot`, the pattern is:

```go
bot, err := ab.db.queryBotByID(id)
if err != nil || bot == nil {
    respondAppBotNotFound(c)  // was: c.JSON(404, {"msg":"bot not found"})
    return
}
```

When `err != nil` (a genuine DB failure), the error is silently collapsed into a 404 without logging. This is pre-existing behavior — the old code also didn't log — but the PR's otherwise-thorough "log before Internal=true respond" discipline makes the gap more visible.

**Assessment:** Anti-enumeration rationale is sound (collapsing query-error into not-found prevents existence probing). However, transient DB issues affecting these 7 handlers would be invisible in server logs. Low production risk given the admin-only access pattern.

**Recommendation:** Consider adding `if err != nil { ab.Error("query bot failed", zap.Error(err)) }` before the combined check. Non-blocking.

### F3 — Test `TestPlatformAppBot_NotFoundLocalized` exercises error path, not true not-found [P2 — test quality]

**Diff-scope:** new (test introduced by this PR).

The test sends `GET /v1/admin/app_bot/does-not-exist` with sqlmock but no query expectations set. sqlmock rejects the unexpected `queryBotByID` query with an error, so the handler takes the `err != nil` branch → 404. The test asserts `StatusNotFound` and `"Bot not found"`, which passes — but via the DB-error path, not the true `bot == nil` path.

**Assessment:** The wire behavior (404 + localized envelope) is correctly validated. The test doesn't prove the true not-found path works, but given the `err != nil || bot == nil` collapse (F2), both paths produce identical output. Non-blocking.

### F4 — Source-guard file list is hardcoded [P2 — maintainability]

**Diff-scope:** new (test introduced by this PR).

`TestAppBotNoLegacyResponseError` scans `[]string{"app_bot.go", "api_i18n.go", "messages.go"}`. A future handler file added without updating this list would be unguarded. The comment in the test documents this requirement.

**Recommendation:** Consider `filepath.Glob("*.go")` with an exclusion list, or a build-tag-based approach. Non-blocking.

## 3. Suggestions

1. **F2 fix (low effort):** Add `zap.Error` logging on the `err != nil` branch before the combined `err != nil || bot == nil` check in the 7 affected handlers. This closes the observability gap without changing wire behavior.
2. **Audit admin roster** before merging the authorization widening.
3. **Stale comment:** `Route()` line 110 says "super admin check in handlers" — should be updated to "admin check in handlers".

## 4. Extra Findings

- **No middleware-level authorization backstop:** The route group `/v1/admin/app_bot` only carries `AuthMiddleware` (authentication). If a handler omits the `CheckLoginRole()` call, any authenticated user reaches handler logic. This is pre-existing across the codebase, not introduced by this PR. Worth a future hardening pass (e.g., route-group middleware that enforces minimum role).
- **`err.shared.auth.forbidden` registration verified:** Registered in `pkg/i18n/codes/shared.go` with HTTPStatus=403, Internal=false. Wire behavior correct for both platform (was 400→now 403) and space (was 403→still 403) paths.

## 5. Data-Flow Trace

| Consumer | Upstream Source | Verified Flow |
|---|---|---|
| `CheckLoginRole()` reads role | `GetLoginRole()` → `c.GetString("role")` ← `AuthMiddleware` ← `CacheTokenParser.Parse()` ← `RoleService.ResolveRole()` (Redis→DB) | ✅ Per-request resolution, not stale-token |
| `respondAppBotForbidden(c)` → `ResponseErrorLWithStatus(c, ErrSharedForbidden)` | `ErrSharedForbidden` = `codes.Lookup("err.shared.auth.forbidden")` → HTTPStatus=403, Internal=false | ✅ Wire 403, localized message exposed |
| `respondAppBotNotFound(c)` → `ResponseErrorLWithStatus(c, ErrAppBotNotFound)` | `ErrAppBotNotFound` registered HTTPStatus=404, Internal=false | ✅ Wire 404, message exposed |
| `respondAppBotQueryFailed(c)` → `ResponseErrorLWithStatus(c, ErrAppBotQueryFailed)` | `ErrAppBotQueryFailed` registered HTTPStatus=500, Internal=true | ✅ Wire 500, message replaced with generic `err.shared.internal`, details suppressed |
| `respondAppBotInternal(c)` → `ResponseErrorLWithStatus(c, ErrAppBotInternal)` | `ErrAppBotInternal` registered HTTPStatus=500, Internal=true | ✅ Wire 500, message hidden |
| `ab.localizedMessage(c, "already_friend")` | `msgtmpl.Catalog.Render(key, lang, nil)` ← `LanguageOrDefault(ctx, "zh-CN")` ← `embed.FS` templates | ✅ MustNew enforces completeness at boot; render miss → logged + "" |
| `respondAppBotRequestInvalid(c, field)` → `ResponseErrorL(c, ErrAppBotRequestInvalid, nil, details)` | `ErrAppBotRequestInvalid` registered HTTPStatus=400, SafeDetailKeys=["field"] | ✅ Wire 400 (D14 path); field detail filtered by SafeDetailKeys |

All 9 platform handlers migrated symmetrically (`CheckLoginRoleIsSuperAdmin` → `CheckLoginRole`). Space-scoped paths use independent `checkSpaceAdmin` (untouched). No add/remove or create/delete asymmetry in authorization gates.

**C2 — Control-flow ordering / nesting reuse:** ✅ Clear.
Response helpers are flat single-call wrappers — no nesting or composition risk. `respondAppBotForbidden` called at 18 sites (9 platform + 9 space guards) with identical behavior. No safety controls (regex/escape/sanitize) in this PR's scope.

**C3 — Authorization boundary ≠ capability boundary:** ✅ Clear (with note).
The authorization widening (superAdmin→admin) is the deliberate core change. No new endpoints or tools are exposed — existing endpoints just accept a broader role set. Token reveal/rotate were already superAdmin-accessible; now admin-accessible. No capability boundary change beyond the intended authorization widening.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The authorization widening is intentional and documented; the i18n migration is thorough with good test coverage and source-guard ratcheting. Four P2 notes (F1–F4) are non-blocking suggestions for the author's consideration.

Jerry-Xin (dismissed):
APPROVE. Two-part change to `modules/app_bot` — (A) delegate the 9 platform `/v1/admin/app_bot` handlers from superAdmin-only to admin∪superAdmin, and (B) localize the whole error/response surface — is correct, well-scoped, and well-tested. No blocking issues.

## PART A — permission delegation (security-sensitive)
All 9 platform gates flip `CheckLoginRoleIsSuperAdmin()` → `CheckLoginRole()`, and each is correctly scoped to the platform branch only:

| Route | Handler | Op type | Delegation |
|---|---|---|---|
| POST `/v1/admin/app_bot` | createPlatformBot | create | ok (mgmt) |
| GET `/v1/admin/app_bot` | listPlatformBots | read | ok |
| GET `/v1/admin/app_bot/:id` | getBotDetail (`spaceID==""` branch) | read | ok |
| PUT `/v1/admin/app_bot/:id` | updateBot (`spaceID==""`) | update | ok (mgmt) |
| DELETE `/v1/admin/app_bot/:id` | deleteBot (`spaceID==""`) | destructive | see note |
| POST `/v1/admin/app_bot/:id/token` | rotateToken (`spaceID==""`) | sensitive | see note |
| POST `/v1/admin/app_bot/:id/token/reveal` | revealToken (`spaceID==""`) | secret-bearing | see note |
| POST `/v1/admin/app_bot/:id/publish` | publishBot (`spaceID==""`) | state change | ok |
| POST `/v1/admin/app_bot/:id/unpublish` | unpublishBot (`spaceID==""`) | state change | ok |

The shared handlers (detail/update/delete/rotate/reveal/publish/unpublish) keep their `spaceID != ""` branch on `checkSpaceAdmin` untouched, so only the platform path moved. `CheckLoginRole()` (octo-lib pkg/wkhttp) rejects empty role (`登录用户角色错误`) and any role other than `Admin`/`SuperAdmin` — so no endpoint becomes reachable by a plain logged-in user. The user-facing `discoverBots`/`applyBot` (`/v1/app_bot`) were not touched.

Note on destructive/secret-bearing ops (delete, rotate-token, reveal-token): under the #372 read/write-boundary precedent, secret-bearing reads and destructive actions are the class that often stays superAdmin. Here all three are delegated to admin. The PR body states this is intentional and confirmed with the maintainer ("admin … can now reach /v1/admin/app_bot, including reveal-token / rotate-token (confirmed expected with the maintainer)"). Accepting on that basis — flagging for the record, not blocking. There is no runETL-equivalent platform-wide op in this surface that was loosened.

## PART B — i18n localization, wire-status parity
No status drift. Verified old→new per class against `origin/main`:
- Space-membership forbidden: was real 403 (`AbortWithStatusJSON(403)`) → `ErrSharedForbidden` via `ResponseErrorLWithStatus` = **403 preserved**.
- Platform role gate: was legacy 400 (`c.ResponseError`) → now **403** (intentional, maintainer-confirmed; tightens correctness, no real status lost).
- Not found: was real 404 (`c.JSON(404,…)`) → `ErrAppBotNotFound` via `ResponseErrorLWithStatus` = **404 preserved**.
- Conflict (id-in-use / token-rotation race): was 400 → **409**; internal failures: was 400 → **500** (Internal=true hides cause). Both improvements, no loss.

Only genuine validation (`request_invalid`, `id_invalid`) uses the D14 400 path (`ResponseErrorL`); every status-bearing code uses `ResponseErrorLWithStatus`, so the console keeps branching on the real wire status. Forbidden renders the shared envelope — `TestSpaceAppBot_ForbiddenLocalized` and `TestPlatformAppBot_PlainUserForbiddenLocalized` assert "permission" and that the raw `该用户无权执行此操作` / `no permission: requires space admin` framework strings are gone. All 20 Internal (500) response sites log the underlying cause (`ab.Error`/`ab.Warn`) before responding, so the message-hiding does not lose observability.

## i18n completeness
All 9 errcode keys (`request_invalid`, `id_invalid`, `not_found`, `id_conflict`, `token_rotation_conflict`, `query_failed`, `store_failed`, `im_token_failed`, `internal`) are present in BOTH `active.zh-CN.toml` and `active.en-US.toml` — no orphans. Apply-flow success bodies localized via `templates/{en-US,zh-CN}/apply.tmpl`. The direct-error-response lint baseline for `app_bot.go` ratchets 9 → 0, consistent with migrating every direct-error site.

## CI / local verification
- CI hard gate green: Build ✅ Lint ✅ Vet ✅ Test ✅ (i18n Extract/Lint ✅).
- Local: `go build ./...` clean, `go vet ./modules/app_bot/...` clean, `go test ./modules/app_bot/... ./pkg/errcode/...` both ok.
- The three red checks are non-code policy/hygiene gates, not the code gate: `check-sprint` fails on "No linked issue found (add Closes #…)", and `dependency-review` / `secret-scan` exit at the org-security scaffolding stage with no finding in the failed logs. Recommend adding a `Closes #<issue>` line to clear `check-sprint`.

## Highlights
- Gate change correctly limited to the platform branch; space guards untouched.
- Status-preserving envelope (`ResponseErrorLWithStatus`) keeps real 403/404/409/500; anti-enumeration (collapsed forbidden + not-found) with cause logged server-side.
- Gate + renderer tests use sqlmock so they run without MySQL and pin both the loosening and the localized envelope.

yujiawei (changes requested):
# Code Review — PR #375 (octo-server)

Reviewed at head SHA `376c0054eee61740d9278d9e5f5629587f55b936` against merge-base `33e24e24` (base `main`). Scope: `modules/app_bot` authorization widening + i18n/error-surface migration. This PR is security-sensitive (authorization change on bot-management endpoints, including token reveal/rotate).

## Verdict: CHANGES_REQUESTED

The i18n migration is clean and well-tested, and the role widening itself is implemented correctly. However, the role widening interacts with a pre-existing missing scope guard on the platform routes to **expand a cross-tenant token-exposure path**, and the PR description states the space surface is unchanged when it is not. This should be closed before merge.

---

## What was verified ✅

- **Role widening is exactly as described.** `CheckLoginRoleIsSuperAdmin()` (superAdmin only) → `CheckLoginRole()` resolves in octo-lib to `role == Admin || role == SuperAdmin` (admin∪superAdmin). It takes no arguments and does not admit any other system role — so there is **no** unintended privilege escalation to support/compliance/etc. roles. Confirmed in `wkhttp/http.go` `CheckLoginRole` / `CheckLoginRoleIsSuperAdmin`.
- **All 9 platform handlers migrated consistently**; no residual `CheckLoginRoleIsSuperAdmin` left in the module.
- **i18n error surface is correct**: each responder maps to the right code/status (validation→400, not-found→404, conflict→409, forbidden→403, internal→500), `Internal=true` codes hide the cause and every such site logs first (`zap.Error`). en-US markers + zh-CN translations are present and complete.
- **Anti-enumeration**: forbidden collapses to one shared code; not-found collapses missing/unpublished — good.
- **Gates pass locally**: `go build`, `go test ./modules/app_bot/...`, `go test ./pkg/errcode/... ./pkg/i18n/...`, `lint-direct-error-response` (baseline ratcheted 9→0), `lint-unregistered-code`, and `i18n-extract-check` all green.

---

## Findings

### P1 — Platform routes can reach space-scoped bots, bypassing the space-admin gate (now exploitable by `admin`)

**`modules/app_bot/app_bot.go`** — all ID-based platform handlers: `getBotDetail` (~L471–490), `updateBot` (~L523–557), `deleteBot` (~L611–630), `rotateToken` (~L671–690), `revealToken` (~L761–784), `publishBot` (~L799–818), `unpublishBot` (~L854–873).

On the platform variant (`spaceID == ""`) the handler runs the role gate, then `queryBotByID(id)`, and the scope check `bot.Scope != "space" || bot.SpaceID != spaceID` runs **only inside `if spaceID != ""`**. Because bot IDs are global, a caller hitting `/v1/admin/app_bot/{aSpaceBotID}` — including `/token/reveal` and `/token` (rotate) — operates on a **space-scoped** bot without ever passing `checkSpaceAdmin`.

Why this is a blocker on *this* PR even though the missing scope check predates it:
- Pre-PR the platform gate was `superAdmin`-only, so only a super admin (who is already all-powerful) could traverse into space bots this way.
- This PR widens that gate to `admin` (operations). Operations can now **reveal or rotate the IM token of any space-scoped bot** for spaces they do not administer — a cross-tenant credential exposure. `revealToken` returns the raw `bot.Token`; `rotateToken` can break a space bot's integration.
- The PR description states *"The space-scoped `/v1/space/:space_id/app_bot` surface is unchanged (still gated by space admin/owner)."* That is accurate for the space route, but the **platform route now reaches space bots with only system-admin authority** — the effective authorization boundary for space-bot tokens has changed, contrary to the stated intent ("Operations needs to manage *platform* bots"). The maintainer sign-off referenced in the PR covers admin access to *platform* bots; it does not obviously cover admin access to *space* bot tokens cross-tenant.

Suggested fix (small, local): on the platform path, after loading the bot, reject non-platform bots, e.g.

```go
} else { // spaceID == "" → platform route
    if err := c.CheckLoginRole(); err != nil {
        respondAppBotForbidden(c)
        return
    }
}
...
bot, err := ab.db.queryBotByID(id)
if err != nil || bot == nil {
    respondAppBotNotFound(c)
    return
}
if spaceID == "" && bot.Scope != "platform" {
    respondAppBotNotFound(c) // platform route only manages platform bots
    return
}
```

This makes the platform route symmetric with the already-correct space route, closes the cross-tenant path for the newly admitted `admin` role, and matches the PR's stated scope. Please add a regression test (admin on `/v1/admin/app_bot/{spaceBotID}/token/reveal` → 404). Two independent review passes flagged this same issue.

> Note (`updateBot`): on the platform path there is additionally **no existence check at all** before `updateAppBot(id, ...)` — a PUT to a non-existent platform id will pass through and (with a 0-rows UPDATE) return `200 OK`. The scope-guard fix above also gives you a natural place to add the missing existence check for the platform path.

### P2 — `applyBot` not-found status changed 400 → 404 on a legacy user endpoint

**`modules/app_bot/app_bot.go:~993`** — `applyBot` is the user-facing `/v1/app_bot/apply` endpoint. A missing/unpublished bot previously went through `c.ResponseError(...)` (wire 400); it now calls `respondAppBotNotFound` → `ResponseErrorLWithStatus` → real **404**. Per the codebase's D14 compatibility convention, legacy-bearing endpoints keep wire=400 via `ResponseErrorL` unless there's explicit sign-off; existing apply clients that branch on 400 could break. The PR's "client-visible behavior changes" section documents the *admin* route status changes but the apply-endpoint change deserves the same explicit call-out / sign-off. Consider a 400-pinned responder (`ResponseErrorL`) for the apply path and reserve the status-preserving 404 helper for the management console paths — or confirm there are no legacy apply clients.

### P2 (advisory, pre-existing — not introduced here) — non-atomic friend creation in `applyBot`

**`modules/app_bot/app_bot.go:~1038–1055`** — the two-directional friend creation is two discrete DB calls with an application-layer rollback; a failed rollback leaks a one-directional "half-friend" record (already logged, not resolved), and concurrent applies can race the `IsFriend` check. This diff only localizes the error strings here, so it is out of scope for this PR, but worth a follow-up ticket.

---

## Coverage / blind spots
- Verified `CheckLoginRole` semantics directly in octo-lib (not just the diff) — it is admin∪superAdmin, no broader.
- Verified the cross-scope hole is pre-existing on the platform path (identical structure on `main`); the PR's contribution is widening *who* can exploit it.
- Not exercised at runtime: actual `space_member` data and a live `admin`-token call against a space bot id — recommend a human confirm the intended authority of `admin` (operations) over **space** bot tokens before merge, since that is the crux of the P1.
- `updateAppBot` 0-rows-affected behavior (whether a no-op update is silently OK) is outside the diff and not independently confirmed; relevant only to the `updateBot` existence-check note.

lml2468 (approved):
✅ APPROVE — 授权委托 + i18n 迁移两维都逐项核实通过,无 blocking。build/vet/test 全过。

## 维度A:授权委托(逐个 op enumerate)
app_bot.go 9 处 gate 统一 CheckLoginRoleIsSuperAdmin → CheckLoginRole(admin∪superAdmin):createPlatformBot/list/getDetail/update/delete/rotateToken/revealToken/publish/unpublish——platform app-bot 全生命周期管理委托给 admin。判断合理:这是应用机器人运营管理(非空间所有权/用户数据破坏类核心权限),admin=运营定位一致;与 #364 把 destructive space/user 操作留 superAdmin 不冲突(app-bot 不涉跨空间用户数据)。gate test 正反覆盖:AdminAllowed(admin 过 gate)、PlainUserForbiddenLocalized(普通用户被拒+本地化403)、NotFoundLocalized(404保真)、SpaceAppBot_ForbiddenLocalized(空间bot 403保真)。无过度放权、无漏改某处。

## 维度B:i18n wire-status 等价(10 个 responder 逐项对照,#371 类 bug 重点查)
| responder | facade | status | |
| Forbidden/NotFound/IDConflict/TokenRotationConflict/QueryFailed/StoreFailed/IMTokenFailed/Internal | WithStatus | 403/404/409/409/500/500/500/500 | ✅保真 |
| RequestInvalid/IDInvalid | ResponseErrorL(钉400) | 400 | ✅本就400 |
规则正确:仅真 400 校验错误用钉-400 ResponseErrorL,所有非-400(403/404/409/500)全用 ResponseErrorLWithStatus 保真 wire status。**无 #371 式漂移**。(workflow 报的"违反 D14"是误报:WithStatus 本就是为保真 status 设计的 facade,用它不违反 D14。)

## i18n key 完整(逐项 comm 核,没被计数假象骗)
errcode/zh-CN/en-US 三者的 app_bot key 集 **comm -23 对比完全一致**(各 9 唯一 key,无缺)。注:grep -c 一度显示 10 vs 9 是文件内某 key 出现两次的计数假象,逐项 diff 确认 key 集无缺失。

## 验证
build + vet 通过;app_bot RespondHelpers/i18n 测试过(TestAppBotRespondHelpers 验每个 responder 的 error.code + error.http_status 双断言);TestPlatformAppBot 正反 gate 测试过。

无 blocking。授权委托合理且测试覆盖、i18n status 逐项保真无漂移、翻译 key 完整。批准。

—— Octo-PR (lml2468)

Jerry-Xin (dismissed):
🔴 CHANGES_REQUESTED — revising my earlier APPROVE @376c0054

I'm withdrawing my APPROVE. yujiawei's P1 is correct and it's a real security regression introduced by this PR — my first pass (and the gate-test) verified the role swap and the space-branch scope check but missed that the **platform branch has no scope guard at all**, which only becomes exploitable once this PR widens that gate to `admin`. Credit to yujiawei (and the second pass that independently flagged it).

🔴 **Blocking — platform route reaches space-scoped bots, cross-tenant, now with only `admin` authority.**
Confirmed by reading `revealToken` (and the same shape in `rotateToken`/`deleteBot`/`updateBot`/`getBotDetail`/`publishBot`/`unpublishBot`): on the platform variant (`spaceID == ""`) the handler runs `CheckLoginRole()`, then `queryBotByID(id)` by **global** id, but the scope guard
```go
if spaceID != "" && (bot.Scope != "space" || bot.SpaceID != spaceID) { respondAppBotNotFound(c); return }
```
is gated behind `spaceID != ""`, so on `/v1/admin/app_bot/{id}` it **never runs**. `revealToken` then returns `gin.H{"token": bot.Token}` for *any* bot id — including a space-scoped bot in a space the caller does not administer.

Why this blocks *this* PR even though the missing platform-side guard predates it:
- Pre-PR the platform gate was `CheckLoginRoleIsSuperAdmin` — only an all-powerful super admin could traverse into a space bot this way, which is tolerable.
- This PR changes that gate to `CheckLoginRole` (admin∪super). An operations `admin` can now **reveal or rotate the IM token of any space's bot, cross-tenant** (`revealToken` exposes the raw `bot.Token`; `rotateToken` can break a space integration). The effective authorization boundary for *space* bot tokens has moved from superAdmin to admin — which the PR's own scope statement ("space-scoped surface unchanged, still gated by space admin/owner") does not cover, and the referenced maintainer sign-off covers admin over *platform* bots, not admin over *space* bot tokens.

Fix (yujiawei's, correct and minimal): on the platform path, after loading the bot, reject non-platform bots so the platform route only manages platform bots —
```go
if spaceID == "" && bot.Scope != "platform" {
    respondAppBotNotFound(c)
    return
}
```
Apply to all ID-based platform handlers (getBotDetail / updateBot / deleteBot / rotateToken / revealToken / publishBot / unpublishBot), making the platform route symmetric with the already-correct space route. Add a regression test: admin on `/v1/admin/app_bot/{spaceBotID}/token/reveal` → 404. (Also: `updateBot` platform path has no existence check before `updateAppBot` — a PUT to a non-existent platform id 0-rows-updates and returns 200; the scope-guard insertion is the natural place to add it.)

What I verified and what still stands:
- ✅ `CheckLoginRole()` (octo-lib f62e626) = admin∪superAdmin, no plain-user path — correct.
- ✅ PART B i18n wire-status parity is clean: space-forbidden 403→403 and not-found 404→404 preserved via `ResponseErrorLWithStatus`; conflict→409, internal→500 improvements; forbidden renders the shared localized envelope (no raw wkhttp string leak). i18n keys complete in both zh-CN + en-US, baseline ratchets 9→0.
- 🟡 PART B caveat (yujiawei P2, agree): `applyBot` (user-facing `/v1/app_bot/apply`) not-found moved 400→404 via the status-preserving helper. Per the D14 convention legacy-bearing endpoints keep wire-400 unless signed off; the PR documents the *admin* route status changes but not this apply-path one. Use a 400-pinned `ResponseErrorL` for the apply path (reserve the 404 helper for console paths), or confirm no legacy apply clients branch on 400.
- 🟡 the delete/rotate/reveal delegation to admin for *platform* bots is maintainer-confirmed per the PR body — that part I accept; the blocker above is specifically the *space*-bot cross-tenant reach, which is not covered by that sign-off.

CI is green (Build/Lint/Vet/Test all pass), so this is purely the authz scope fix. Add the platform-scope guard + regression test and I'll re-approve.

lml2468 (changes requested):
🔴 撤回我的 APPROVE,改 CHANGES_REQUESTED,跟 @yujiawei 的 P1 —— 我本地核实成立,这是真 blocker,我漏了。跨租户凭证泄露 + IDOR,被本 PR 的权限委托放大。

## 🔴 P1:platform 路由可达 space-scoped bot,绕过 checkSpaceAdmin(本 PR 放大为 admin 可利用)
核实(revealToken app_bot.go:758-794,及所有 ID-based platform handler):
- platform 路径(spaceID==""):过 CheckLoginRole(现 admin 即可)→ queryBotByID(id)→ 直接返回/操作,**不校验 bot 是否真是 platform bot**。
- scope 检查 `if spaceID != "" && (bot.Scope != "space" || bot.SpaceID != spaceID)` **只在 spaceID != "" 时跑** → platform 路径完全无 scope 边界。
- bot ID 全局 → 任意 admin 打 `/v1/admin/app_bot/{任意space_bot_id}/token/reveal` 能读到**任意空间 bot 的 raw token**(revealToken 直接 c.Response{token: bot.Token}),绕过 checkSpaceAdmin。

逐个核实,以下 platform handler 全有此缺口(scope 检查都 gated 在 spaceID!="" 内):
getBotDetail(:468)、updateBot(:520,连条件检查都没有)、deleteBot(:608)、rotateToken(:668)、revealToken(:758)、publishBot(:796)、unpublishBot(:851) —— **7 个端点**。

为何是本 PR 的 blocker(虽 scope 缺失早于本 PR):
- PR 前 gate 是 superAdmin-only,只有本就全能的超管能借此进 space bot。
- 本 PR 把 gate 降到 admin → **任意运营 admin 现在能 reveal/rotate 任意空间 bot 的 IM token**(跨租户凭证泄露 + 能破坏任意空间 bot 集成)。
- PR 描述称"space surface 不变、仍由 space admin 把关"——对 space 路由成立,但 **platform 路由现在用 system-admin 权限就能碰 space bot 的 token**,有效授权边界变了,与"只管 platform bot"的声称不符。maintainer sign-off 覆盖的是 admin 管 platform bot,不含跨租户 space bot token。

## 修法(yujiawei 给的,小而局部)
platform 路径加载 bot 后,拒绝非 platform bot:
```go
} else { // platform 路由
    ...CheckLoginRole...
}
bot, _ := queryBotByID(id)
if spaceID == "" && bot.Scope == "space" {  // platform 路由不得操作 space bot
    respondAppBotNotFound(c)  // 反枚举:NOT_FOUND 而非 403
    return
}
```
7 个 handler 都要补;并加 admin 打 space-bot-id 被拒的回归测试。

## 我的复盘(诚实)
我核授权委托时验了"gate 一致降级 + 正反 gate test",但**只盯 gate 那一行、没核 gate 之后 handler 内部的 scope 边界**——漏了 queryBotByID 之后没有 platform-scope 校验这层。这是「主路径对了、漏更深一层」+「权限委托放大既有 IDOR」的组合。yujiawei 把 gate 变更和既有缺失的 scope guard 联系起来看,我没有。这条比单纯漏 should-fix 重——是 P0/P1 级安全洞。

i18n 迁移那半(status 保真、key 完整)确实干净、维持。问题在授权委托放大了 platform 路由的 scope 缺口。修这 7 个 handler + 测试后我重审。

—— Octo-PR (lml2468)

an9xyz:
Fixed the P1 in `029edba`. Thanks @yujiawei for catching it, and @Jerry-Xin / @lml2468 for the concurrence — the cross-tenant reach was real and the gate-widening is exactly what made it exploitable.

**P1 — cross-tenant IDOR closed.** Added `botInRouteScope(bot, spaceID)`: the platform route (`spaceID==""`) admits only `platform`-scoped bots, the space route only its own space's bots — making the platform path symmetric with the already-correct space path. Mismatches return **404** (anti-enumeration, not 403). Wired into all 7 ID-based handlers (`getBotDetail` / `updateBot` / `deleteBot` / `rotateToken` / `revealToken` / `publish` / `unpublish`). `updateBot`'s platform path didn't load the bot at all, so it now `queryBotByID` + scope-checks first — which also closes the `200`-on-nonexistent-id no-op you flagged.

Regression tests: `TestPlatformAppBot_SpaceBotViaPlatformRouteRejected` (admin → `POST /v1/admin/app_bot/{spaceBotId}/token/reveal` → 404, and the mocked space-bot token is asserted absent from the body) and `TestBotInRouteScope` (pins the boundary, incl. the platform-route-rejects-space-bot case).

**Non-blockers also addressed:**
- **P2 (apply status):** the user-facing `/v1/app_bot/apply` not-found is pinned back to **wire-400 (D14)** via `respondAppBotNotFoundPinned` (envelope still carries the real 404); the console paths keep the real 404. Test: `TestAppBotNotFoundPinnedIsWire400`.
- **F2 (observability):** every ID-based handler now logs the query error (`zap.Error`) before collapsing into the 404.
- **Stale comment:** `Route()` "super admin check in handlers" → "admin∪superAdmin role check in handlers".

`go build ./...`, vet, gofmt, `i18n-lint`, and the full `modules/app_bot` suite pass. Re-review when you have a moment.

---

lml2468 (dismissed):
✅ APPROVE（重审 @029edba）— P1 跨租户 token IDOR 修对了,逐项核实 + build/test 过。

## P1 修复（本地逐条核实）
新增统一 scope 守卫 botInRouteScope(bot, spaceID):
```go
func botInRouteScope(bot *appBotModel, spaceID string) bool {
    if spaceID == "" { return bot.Scope == "platform" }  // platform 路由只接受 platform bot → 拒绝 space bot
    return bot.Scope == "space" && bot.SpaceID == spaceID  // space 路由保持空间隔离
}
```
- ✅ platform 路由(spaceID=="")现在**拒绝 space bot**(NOT_FOUND 反枚举)——堵住"admin 经 /v1/admin/app_bot/{spaceBotID}/token/reveal 读任意空间 bot token"的跨租户 IDOR。
- ✅ **7 个 ID-based platform handler 全部接入** botInRouteScope:getBotDetail/updateBot/deleteBot/rotateToken/revealToken/publishBot/unpublishBot,无遗漏(各 x1)。两条路由对称、各自正确。
- ✅ 回归测试到位:TestPlatformAppBot_SpaceBotViaPlatformRouteRejected(admin 经 platform 路由打 space bot 被拒)+ TestBotInRouteScope(scope 矩阵单元测)——正是这个 P1 的护栏。
- build ./modules/app_bot/ + app_bot 全测试过。

## 已正确（维持）
PART A 授权委托(superAdmin→admin,9 处一致,正反 gate test)、PART B i18n wire-status 保真(403/404/409/500 全 WithStatus、400 校验用钉-400)、i18n key 完整。

P1 安全回退已修对(scope 守卫 + 7 handler 全接 + 回归测试),撤回 CHANGES_REQUESTED 改 APPROVE。
(注:applyBot not-found 400→404 那条 🟡——用户端 /v1/app_bot/apply 的 legacy 兼容,Steve/yujiawei 提的,建议作者确认是否保 400;非阻塞,可后续。)

—— Octo-PR (lml2468)

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#375 Review Report
**PR:** Mininglamp-OSS/octo-server #375
**Head SHA:** `029edba89cc38390478949c2ebe5ab71a4a5f6ab`
**Title:** feat(app_bot): delegate platform bot management to admin + localize the error/response surface

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Role gate widening (superAdmin → admin∪superAdmin) | ✅ | `CheckLoginRole` accepts admin+superAdmin (octo-lib `pkg/wkhttp/http.go:180-189`); applied consistently to all 9 platform handlers |
| IDOR fix (`botInRouteScope`) | ✅ | `app_bot.go:163-175`; applied to all 7 handlers that load bot by ID |
| i18n migration completeness | ✅ | Baseline ratcheted 9→0; all errcodes registered with en-US + zh-CN translations |
| Internal=true codes log before responding | ✅ | Every call site verified to `zap.Error(err)` before calling respondAppBot*Failed/Internal |
| Test coverage | ✅ | Gate tests, scope tests, responder tests, source-guard test, IDOR regression test |
| Status preservation (404/409/403/500) | ✅ | `ResponseErrorLWithStatus` used for all status-preserving codes; `ResponseErrorL` (D14 pinned 400) only for apply-path `respondAppBotNotFoundPinned` |

## 2. Findings

### No P0/P1 findings.

**P2-1: Comment-stripping in source-guard test has theoretical blind spot**
- `TestAppBotNoLegacyResponseError` (`api_i18n_test.go:30-40`) strips comments by finding `//` per line. This would incorrectly strip `//` inside string literals (e.g., a URL `"https://..."` would be truncated). No current string literals in the scanned files contain `//`, so this is not a live bug.
- **Diff-scope:** new (introduced by this PR). Pre-existing: N/A (new test file).
- **Severity:** P2/nit — theoretical only, no live false-negative path.
- **Suggestion:** Consider using `go/token` + `go/scanner` for robust comment stripping, or add a comment noting the limitation.

**P2-2: reveal-token / rotate-token now accessible to `admin` role**
- The widening from superAdmin-only to admin∪superAdmin on `revealToken` and `rotateToken` means `admin` users can now read raw bot tokens and rotate them. PR body confirms this was "confirmed expected with the maintainer."
- **Diff-scope:** new (intentional policy change). Not a code defect.
- **Severity:** P2/note — deliberate authorization policy change, not a bug. Worth monitoring: if the `admin` role is broadly assigned, sensitive token operations may see wider access than intended.

## 3. Suggestions

1. **Consider rate-limiting reveal-token**: Now that `admin` can reveal tokens, a rate limit on `POST /v1/admin/app_bot/:id/token/reveal` would reduce the blast radius of a compromised admin session.
2. **Audit log for token operations**: Consider emitting a structured audit log entry (beyond `zap.Error`) for `revealToken` and `rotateToken` — e.g., including the caller UID and bot ID in a dedicated audit channel.

## 4. Extra Findings

### Security improvement: IDOR fix (bonus, not in PR title)

The PR silently fixes a **cross-tenant IDOR** that existed in all 7 handlers that load a bot by global ID. The old code gated the scope check with `spaceID != ""`, meaning the **platform route** (`/v1/admin/app_bot/:id`, where `spaceID==""`) skipped the scope check entirely. A superAdmin (now admin) could:

- **revealToken**: read any space bot's raw token (most severe)
- **rotateToken**: rotate any space bot's token
- **deleteBot**: delete any space's bot
- **updateBot**: update any space's bot with **zero** existence/scope checks (the entire `queryBotByID` was inside `if spaceID != ""`)
- **getBotDetail / publishBot / unpublishBot**: read/modify any space's bot

The new `botInRouteScope` function correctly closes all these paths. The `updateBot` fix is particularly impactful — the old code had no existence check at all on the platform route.

### Anti-enumeration design

The PR correctly collapses forbidden/not-found responses:
- All authorization failures (role gate, space admin, space membership) → single `err.shared.auth.forbidden` code
- Missing bot and scope-mismatched bot → single `err.server.app_bot.not_found` code
- `Internal=true` codes hide the message on the wire; underlying cause logged server-side only

## 5. Data Flow Tracing

|---|---|---|
| `c.GetLoginRole()` → role gate | Auth middleware sets `c.Set("role", info.Role)` from token cache (`uid@name@role`) | ✅ Verified: empty role → both checks fail; "admin" → `CheckLoginRole` passes, `CheckLoginRoleIsSuperAdmin` fails |
| `botInRouteScope(bot, spaceID)` → scope gate | `bot.Scope` and `bot.SpaceID` from `queryBotByID` SQL result; `spaceID` from URL param | ✅ All 7 handlers verified |
| `errcode.ErrAppBot*` → i18n envelope | `register()` → `codes.Register()` (panics on duplicate/invalid) → `codes.Lookup()` at render time | ✅ All 9 codes registered with valid IDs, messages, and status codes |
| `Internal=true` → message hiding | Renderer `message()` method substitutes `err.shared.internal` translation when `spec.Internal==true`; `filteredDetails()` returns empty `Details{}` | ✅ Verified in `pkg/i18n/renderer.go:68-95` |
| `respondAppBotNotFoundPinned` → wire 400 | `ResponseErrorL` (not `WithStatus`) → `useSemanticStatus=false` → transport status = 400; envelope keeps `http_status: 404` | ✅ Verified; test `TestAppBotNotFoundPinnedIsWire400` pins this |
| `ab.localizedMessage(c, key)` → success body | `msgtmpl.MustNew` (panics on missing template at boot) → `Render(key, lang, nil)` → `LanguageOrDefault` from request context | ✅ Templates embedded via `//go:embed`; both en-US and zh-CN define both keys |

## 6. R5 Blind Spot Checklist (security_sensitive — mandatory)

### C1 — Dual-path parity
**Status: CLEAR**

- Platform route (9 handlers): all use `CheckLoginRole` + `botInRouteScope` ✅
- Space route (9 handlers): all use `checkSpaceAdmin` + `botInRouteScope` ✅
- `botInRouteScope` is symmetric: platform route admits only `scope=="platform"`, space route admits only `scope=="space" && SpaceID==spaceID` ✅
- No create↔delete / subscribe↔unsubscribe asymmetry found
- No create-gate vs execute-path authorization gap: each handler independently checks role

**Status: CLEAR**

- `respondAppBot*` helpers are thin single-call wrappers — no nesting, no double-action risk
- `Internal=true` message substitution happens at the renderer level, not in the helper — no ordering issue
- No regex/sanitize/escape functions introduced that need non-canonical-form testing
- The `botInRouteScope` function is called after `queryBotByID` and before any mutation — correct ordering

**Status: CLEAR**

- The role widening (superAdmin → admin∪superAdmin) is an intentional policy change, confirmed in PR body
- `botInRouteScope` **tightens** the authorization boundary: platform admins who previously could access space bots via IDOR now cannot
- No new endpoints or tools are exposed; the change is purely a gate adjustment on existing endpoints
- The `applyBot` endpoint's space-membership check correctly maps to `respondAppBotForbidden` (not a different error)

---

## [Octo-Q] verdict: APPROVE

No P0/P1 findings. The PR is well-structured, closes a real cross-tenant IDOR, migrates all error responses to the i18n envelope cleanly, and includes solid test coverage with a source-guard regression test. The role widening is intentional and confirmed. Two P2 observations (test comment-stripping edge case, reveal-token access widening note) are non-blocking.

lml2468 (approved):
✅ APPROVE（重审 @8b92ab4）— 本 commit 仅新增 IDOR e2e 回归测试,P1 代码(botInRouteScope)自 029edba 未变,我之前的 APPROVE 结论维持。

## 本 commit 变更
仅加 modules/app_bot/app_bot_idor_e2e_test.go(+64)。git diff 029edba..8b92ab4 确认只 1 文件、纯测试,P1 修复代码未动。

## e2e 测试强化了 P1 验证(核实)
TestAppBotPlatformRouteScope_E2E:seed 一个 space bot(scope=space,space_id=X) + 一个 platform bot,admin 经 platform 路由 /v1/admin/app_bot/{id}/token/reveal:
- space bot → **404**(平台路由拒绝 space bot,IDOR 被堵)✅
- platform bot → token 正常返回(守卫不误伤合法 platform bot)✅
端到端证明跨租户 token 暴露已修。(本地 e2e FAIL 是 testutil.NewTestServer 需 DB 的环境限制,CI 会跑;非-DB 的 TestBotInRouteScope 单元测试本地过;build 通过。)

## P1 修复(维持上轮已核结论)
botInRouteScope(platform 路由只接受 platform bot、拒绝 space bot)+ 7 个 ID-based handler 全接入 + 反枚举 NOT_FOUND。授权委托 + i18n wire-status 保真 + key 完整也维持正确。

P1 跨租户 IDOR 已修对 + 现补 e2e 回归护栏。重投 APPROVE。
(mochashanyao 两条 🟡:注释剥离测试理论盲点、admin token 操作建议限流审计——都是非阻塞加固建议,可后续。)

—— Octo-PR (lml2468)

yujiawei (approved):
# Code Review — PR #375 (octo-server)

Reviewed at head `029edba89cc38390478949c2ebe5ab71a4a5f6ab` against merge-base `33e24e24` (base `main`). Build and the full `modules/app_bot` + `pkg/errcode` test suites pass locally; the new IDOR/gate tests and i18n responder tests are green. This is a security-sensitive change (authorization + token surface), so the authorization and token-exposure paths were checked end-to-end.

## Verdict

**APPROVED.** No P0/P1 issues. The core security work in this PR — the cross-tenant IDOR fix and the localized error surface — is correct and well-tested. The items below are non-blocking suggestions, plus one authorization-policy point that warrants explicit human sign-off.

## What this PR does (verified)

1. **Platform bot management delegated from superAdmin to admin∪superAdmin.** `createPlatformBot` / `listPlatformBots` and the platform branch of the per-id handlers now call `CheckLoginRole()` (admin ∪ superAdmin) instead of `CheckLoginRoleIsSuperAdmin()` (superAdmin only). Confirmed against octo-lib: `CheckLoginRole` admits `role == "admin" || role == "superAdmin"`; `CheckLoginRoleIsSuperAdmin` admits only `superAdmin`. This is an intentional broadening, pinned by `TestPlatformAppBot_AdminAllowed`.
2. **Cross-tenant IDOR fix.** New `botInRouteScope(bot, spaceID)` rejects a bot that lives outside the route's authority: the platform route (`spaceID==""`) admits only `scope=="platform"` bots; the space route only its own space's bots. Previously the platform branch loaded a bot by global id with **no** scope check, so a platform-route caller could read/rotate/reveal any space bot's token by id. All seven load-by-id handlers (`getBotDetail`, `updateBot`, `deleteBot`, `rotateToken`, `revealToken`, `publishBot`, `unpublishBot`) now apply the guard. `TestPlatformAppBot_SpaceBotViaPlatformRouteRejected` proves a space bot's raw token is never returned via the platform route. This is a genuine security improvement.
3. **Error surface localized.** Every error path routes through `httperr.ResponseErrorL*` + registered `errcode.ErrAppBot*` / shared codes. Status-preserving codes (404/409/403/500) use `ResponseErrorLWithStatus`; validation pins 400. The apply path keeps a fixed-400 not-found responder (`respondAppBotNotFoundPinned`) for SDK compatibility. The lint baseline ratchet (`app_bot.go` 9 → 0) is honest — `go run ./tools/lint-direct-error-response/` reports clean.

## 🔒 Human-verify (authorization policy)

This is the security-sensitive crux and is a product/policy decision rather than a bug: **platform-wide app-bot management — including create, delete, token rotation, and `reveal-token` for every platform bot — is now available to any `admin` system role, not only `superAdmin`.** The IDOR guard correctly prevents an admin from reaching *space* bots via the platform route, and the role is a platform-level system role (not a space role), so the blast radius is platform admins over platform bots. The change is intentional and tested, but please confirm that delegating platform-bot token reveal/rotation to the `admin` tier matches the intended authorization model before merge.

## Findings (all P2, non-blocking)

### P2 — Apply membership failure changed wire status 400 → 403
`modules/app_bot/app_bot.go` (apply space-membership branch, ~L1052). When `/v1/app_bot/apply` is called for a published space bot by an authenticated non-member, the response now goes through `respondAppBotForbidden` → `ResponseErrorLWithStatus` → wire **403**. The legacy `c.ResponseError(...)` emitted wire **400** (confirmed: octo-lib `ResponseError` hard-codes `http.StatusBadRequest`). The apply endpoint deliberately pins its *not-found* branch to wire 400 (`respondAppBotNotFoundPinned`, D14) precisely because apply SDK clients may branch on the legacy status — but the membership-forbidden branch silently moves to 403. 403 is more semantically correct, but if apply SDK clients branch on the legacy failure status this is a behavior change. Recommend either pinning this branch to 400 for consistency with the documented apply-path contract, or confirming no client depends on the old status.

### P2 — Authorization rejections are not logged, contradicting the file's own contract
`modules/app_bot/api_i18n.go` states for the forbidden responder: "the specific role/membership reason stays in logs, never on the client." In practice the handlers **discard** the error from `c.CheckLoginRole()` / `ab.checkSpaceAdmin()` (and the `memberCount == 0` branch in `applyBot`) and call `respondAppBotForbidden(c)` with no log line. The not-found/query/store paths were given `ab.Error(...)` log calls in this PR, but the authorization-denial paths were not. For a security-sensitive surface, denied platform/space access attempts (including cross-tenant probing) are exactly what you want in the audit trail. Recommend a debug/info log at the deny sites, or correcting the comment if silence is intended.

### P2 — Regression guard does not ban error-path `c.JSON(...)`
`modules/app_bot/api_i18n_test.go` (`TestAppBotNoLegacyResponseError`). The `banned` list covers `.ResponseError(`, `.AbortWithStatusJSON(`, etc., but not `.JSON(`. This PR correctly removed the legacy `c.JSON(http.StatusNotFound, ...)` error responses, and the `api_i18n.go` header explicitly names `c.JSON(40x, ...)` as a migrated anti-pattern — yet neither this test nor the AST linter (`tools/lint-direct-error-response`, which only counts `AbortWithStatusJSON`/`AbortWithStatus`) would catch a reintroduced unlocalized `c.JSON` error response. Note the one remaining `c.JSON(http.StatusOK, ...)` in `discoverBots` is a legitimate success response and must stay allowed, so any guard would need to target error statuses specifically.

## Blind spots / coverage

- **Pre-existing (not introduced here):** `deleteBot` deletes the DB row before invalidating the IM token / clearing registries. A crash between the DB commit and `UpdateIMToken` can leave an orphaned, still-valid IM token. The diff does not change this ordering, so it is out of scope for this PR, but flagging it for a future hardening pass (transactional outbox / reconciler) since the token surface is security-relevant.
- **Not assessable from the diff:** the internal implementations of `CheckLoginRole`/`checkSpaceAdmin` were verified against the pinned octo-lib version, but `userService`/`imService` retry/rollback semantics and DB constraints are outside this changeset.
- **Advisory legs:** both an independent line-level adversarial reviewer and a long-context whole-repo reviewer were run; both findings were independently re-verified against the diff before inclusion. No contradictions between the two; no consensus duplicates (each surfaced distinct points).

yujiawei (approved):
# Code Review — PR #375 (octo-server), re-review at new head

Re-reviewed at head `8b92ab49c93873d0ce589192cd196e43b4355228` (prior approval was at `029edba8`). The only delta since that approval is one additive, test-only commit: `8b92ab49 test(app_bot): add full-server e2e for the platform-route IDOR guard`.

## Verdict

**APPROVED.** No P0/P1. The privilege-delegation + IDOR-guard + i18n-migration changes I approved at `029edba8` are unchanged; the new commit only strengthens coverage.

## What changed since the prior approval (verified)

- **Production code is byte-identical** to `029edba8` (`git diff 029edba8 8b92ab49 -- . ':(exclude)*_test.go'` is empty).
- New file `modules/app_bot/app_bot_idor_e2e_test.go` adds a full-server e2e (`TestAppBotPlatformRouteScope_E2E`) that boots the real server + AuthMiddleware + MySQL migrations and proves end-to-end that (a) an admin cannot reveal a **space** bot's token via the platform route (404, token absent from body), and (b) a genuine **platform** bot's token *is* revealed (the guard does not over-block the intended delegation). This is a meaningful upgrade over the sqlmock route harness — it exercises real auth + DB rather than mocks.
- Builds clean; `go vet ./modules/app_bot/` clean; all sqlmock-based gate/scope/responder tests pass locally. The new e2e could not be executed in my local environment due to a pre-existing dirty local-DB migration state unrelated to this PR — it will run on CI's clean database.

## Carry-over findings (all P2, non-blocking — unchanged from prior review)

- **P2** — Apply non-member path now returns wire **403** (was 400 via legacy `ResponseError`); the apply path otherwise pins not-found to 400 for SDK compat. Pin this branch to 400 or confirm no client branches on it.
- **P2** — Authorization denials (`CheckLoginRole` / `checkSpaceAdmin` / `memberCount==0`) are discarded without a log line, contradicting `api_i18n.go`'s "the specific reason stays in logs" contract. Add deny-site logging or correct the comment.
- **P2** — `TestAppBotNoLegacyResponseError` bans `ResponseError`/`AbortWithStatusJSON` but not error-path `c.JSON(...)`; the AST linter only counts `AbortWith*`. A reintroduced unlocalized `c.JSON` error wouldn't trip CI (keep the legit `c.JSON(200,...)` in `discoverBots` allowed). Also note the test's `//`-based comment stripping would mis-handle a `//` inside a string literal — no live occurrence today, theoretical only.

## 🔒 Human-verify (policy, not a bug)

Platform-wide bot management — including **token reveal/rotation for every platform bot** — is now open to the `admin` system role, not just `superAdmin`. The IDOR guard limits the blast radius to platform-admins over platform bots. Intentional and tested; please confirm it matches the intended authorization model before merge. Worth considering: a rate limit and/or dedicated audit-log entry (caller UID + bot ID) on the reveal/rotate endpoints now that the role is broader.

## Blind spots

- Pre-existing, out of scope: `deleteBot` deletes the DB row before invalidating the IM token; a crash in that window can orphan a still-valid token. Not changed here — flag for a future hardening pass.

Jerry-Xin (approved):
✅ APPROVED — re-review @8b92ab (clearing my prior CHANGES_REQUESTED)

The cross-tenant token IDOR P1 is fixed correctly across every handler, and the two follow-on points I raised are resolved. All four CI gates green (Build/Lint/Vet/Test). Verified locally: `go build ./...` + `go vet ./modules/app_bot/...` clean; the gate/scope/e2e tests pass.

✅ **🔴 Cross-tenant token IDOR — fixed across all 7 ID-based platform handlers.**
New `botInRouteScope(bot, spaceID)` helper: platform route (`spaceID==""`) admits only `bot.Scope=="platform"`; space route only `bot.Scope=="space" && bot.SpaceID==spaceID`. The old guard `if spaceID != "" && (...)` (which skipped the platform path entirely) is replaced with the unconditional `if !botInRouteScope(...) { respondAppBotNotFound }` in getBotDetail / updateBot / deleteBot / rotateToken / revealToken / publishBot / unpublishBot. So an admin hitting `/v1/admin/app_bot/{spaceBotID}/token/reveal` now resolves to NOT_FOUND — the space bot's token is never returned via the platform route.

✅ **🔴 updateBot platform-path existence check — fixed.** The `queryBotByID` + nil-check + scope check now runs unconditionally for both paths (previously gated behind `spaceID != ""`, so a platform PUT to a non-existent id 0-rows-updated and returned 200).

✅ **🟡 applyBot 400→404 — addressed correctly.** New `respondAppBotNotFoundPinned` uses `ResponseErrorL` (legacy fixed-400 / D14) for the user-facing `/v1/app_bot/apply` path whose SDK clients may branch on 400, while the management-console paths keep the real 404 via `respondAppBotNotFound`. Exactly the split I recommended.

✅ **Regression tests prove the fix:**
- `TestBotInRouteScope` pins the full scope matrix, incl. the IDOR guard (space bot via platform route → false).
- `TestPlatformAppBot_SpaceBotViaPlatformRouteRejected` (sqlmock): admin on platform reveal-token with a space bot id → asserts 404 **and** the raw `super-secret-space-token` never appears in the body.
- `TestAppBotPlatformRouteScope_E2E` (full server + real MySQL migrations + real AuthMiddleware): proves end-to-end that admin cannot reveal a space bot's token via the platform route, while a genuine platform bot's token IS revealed (the guard doesn't over-block the intended delegation).

Credit to yujiawei for first catching the cross-tenant path — I missed it on my first pass and have said so. The fix is symmetric, well-tested, and the intended platform-bot delegation to admin still works. Dismissing my prior CR.

Issue #381, opened 2026-06-16
Follow-up: non-blocking hardening for integration team-group endpoints (PR #377 review)


an9xyz:
Tracks the **non-blocking (P2 / advisory)** items raised during the #377 review rounds (reviewers: @yujiawei, @lml2468, @mochashanyao, @Jerry-Xin). #377 was approved 4/4 and merged with these deferred so as not to churn an already-converged PR. None are exploitable or correctness regressions in the current architecture.

Follow-up to #378 / #236.

> **Resolution:** Closed. The cheap/local follow-ups landed in **PR #383** (migration-test inlining, `isHumanUser` symmetry guard, pending-TTL widening, plus the `created_at` zero-value correctness bug found in the same review). The non-atomic compensating-delete / orphan-group item is split out to **#394** (its own async-reconcile work). The remaining items (payload-fingerprint serialization, DB uniqueness fallback, zero-width whitespace) are accepted advisory trade-offs — see per-item notes; in particular a uniqueness constraint is intentionally **not** pursued (same-name groups are legal; documenting the `Idempotency-Key` requirement is preferred). Boxes below reflect final disposition.

## Cheap, PR-local cleanups
- [x] **Migration test inlines DDL instead of reusing the migration file** (maintainability). ~~`modules/group/group_name_widen_test.go` hand-writes `ALTER TABLE \`group\` MODIFY \`name\` VARCHAR(40/50)` to reproduce the pre-widen state; this can drift from the real `sql/20260615000001_group_name_widen.sql`. Prefer reading the migration's Up SQL.~~ **Done in #383** — the test now parses and runs the migration file's own Up statements via `rubenv/sql-migrate` (the same source production runs), so assertions break against the real DDL.
- [ ] **Idempotency payload fingerprint serialization.** `teamGroupPayloadSHA` concatenates `name` + sorted robot IDs with a `\0` delimiter; in principle a `name` containing an embedded NUL could shift the boundary and collide. Currently mitigated (robot IDs are DB-validated UUID-style values). Deterministic struct/JSON marshaling would be more robust against future field additions.
[...]

PR #377, opened 2026-06-15
feat(integration): team group create + existence endpoints for uk_ keys

an9xyz:
Closes #378
Related: #236

## Background

Let integration partners use the `uk_` User API Key (issued via OIDC exchange; the client is whatever the key is scoped to) to:
1. **Create a team group** — members are the calling user (human owner) + the specified team bots.
2. **Check whether a team group exists** (user-scoped existence check).

Implemented in `modules/integration` (alongside exchange under `/v1/integrations/oidc`). The capability is generic to any integration client — not coupled to a specific partner.

### 1. `feat(group)`: widen group name limit 20 → 50
- The two `> 20` rune truncation sites in `CreateGroup` / `UpdateGroupInfo` now share a `MaxGroupNameLen = 50` constant (one global ceiling for Web / Bot / Integration create + rename).
- Migration widens `group`.`name` from `VARCHAR(40)` to `(50)` (online in-place ALTER; both widths < 256 bytes, no table rebuild).
- The migration test genuinely exercises the **pre-existing data** path: shrink back to `VARCHAR(40)`, write a 40-rune legacy row, run the widen over it, then assert the row survives with no truncation and a 50-rune name now fits.

### 2. `feat(integration)`: two `uk_` endpoints

**`POST /v1/integrations/oidc/groups`** — create a team group
- owner / space / client are all resolved server-side from the key context; the **body is untrusted** (only `name` + `member_robot_ids`).
- The owner must be an **active member** of the key's space (`CheckMembership`: `space_member.status=1 AND space.status=1`) and a **human** (`robot=0`); account-level activity is enforced upstream by `AuthByKey`.
- `member_robot_ids` are validated against the **same source `CreateGroup` inserts from** (owned + active in the space + backed by a usable user row), so an ineligible bot is a unified `404` **before** any group is created (prevents bot-id enumeration and avoids a created-then-500 orphan).
- No bot is promoted to `bot_admin`; the response echoes only the bots that **actually joined**; `created_at` is RFC3339.
- Optional **Idempotency-Key** (Redis, 24h, namespaced by client/uid/space): replay returns the original result; in-flight → `409 idempotency_in_flight` (with `Retry-After`, retryable); same key with a different body → `409 idempotency_conflict` (terminal). The lookup runs ahead of the mutable eligibility checks so a stored result still replays after state changes.

**`GET /v1/integrations/oidc/groups/:group_no`** — existence check
[...]

PR #383, opened 2026-06-16
fix(integration): correct team-group created_at + two #381 follow-ups

### LINKED ISSUE #384: Integration team-group: fix created_at + two non-blocking #381 follow-ups
Scoped child of #381 to carry the work in the corresponding PR, so #381 stays open as the umbrella for the remaining items.

Covers:
1. **`created_at` correctness bug** — `POST /v1/integrations/oidc/groups` returned the Go zero time (`0001-01-01T00:00:00Z`) on every successful create. `queryGroupCreatedAt` scanned `group.created_at` into a bare `time.Time`, which dbr resolves to the zero value with a nil error, so the `time.Now()` fallback never fired.
2. **Existence check human guard** — `teamGroupExists` now applies the same `isHumanUser` guard `createGroup` enforces (endpoint symmetry / defense-in-depth).
3. **Migration test** — the name-widen test now executes the real migration file's Up DDL (via `rubenv/sql-migrate`) instead of an inlined `ALTER`, so it cannot drift from `sql/20260615000001_group_name_widen.sql`.
4. **Idempotency pending TTL** — widened 60s → 5m. `CreateGroup` calls IM over an HTTP client with no timeout, so a degraded WuKongIM can block past the old 60s TTL, expiring the in-flight lock mid-flight and letting a concurrent same-key retry create a duplicate group. No provably-safe fixed TTL exists (create latency is unbounded without an IM-call timeout); 5m covers realistic degradation and shrinks the window to near zero. Fully closing it (IM-call timeout or lock renewal) stays tracked in #381.

The remaining #381 items (orphan-group reconcile, payload fingerprint serialization, DB uniqueness fallback, zero-width group-name normalization) stay tracked in #381.

an9xyz:
## Summary

Addresses 4 of the items tracked in #381 (the non-blocking follow-ups from the #377 review). Scoped deliberately to the cheap, low-risk, integration/group-local ones; the remaining items are left for separate work (see below).

### 1. `created_at` correctness bug (the only correctness defect in #381)
`POST /v1/integrations/oidc/groups` returned `"created_at":"0001-01-01T00:00:00Z"` on **every** successful create.

`queryGroupCreatedAt` scanned `group.created_at` into a **bare `time.Time`**. dbr's reflection treats that as "find a `created_at` field inside `time.Time`", finds none, scans into a throwaway dest, and yields the **zero value with a nil error** — so the `time.Now()` fallback (guarded by `err != nil`) never fired and the zero time was returned as the success value. (Scanning into `db.Time` directly would hit the same path — it is also a struct with no `sql.Scanner`; the column only binds when matched to a *named struct field*, which is how the rest of the codebase reads it.)

Fix: scan into a struct field (mirroring `queryGroupStatus`) and treat a zero result as a read failure so the existing server-time fallback engages. The test previously only asserted `time.Parse(RFC3339)` did not error — and the zero time is itself valid RFC3339, so it slipped through; it now requires a non-zero timestamp close to now.

### 2. Existence check — symmetric human guard
`teamGroupExists` now applies the same `isHumanUser` guard `createGroup` already enforces, so the two endpoints stay symmetric (a creator later flipped to a robot no longer confirms the group). Defense-in-depth; not exploitable today (the check is already creator-scoped). Added a mirror test.

### 3. Migration test no longer inlines DDL
The name-widen test hand-inlined `ALTER TABLE ... MODIFY ... VARCHAR(50)` to stand in for the migration, which could drift from `sql/20260615000001_group_name_widen.sql`. It now parses and runs the migration file's own Up statements (via `rubenv/sql-migrate`, the same source production runs), so the assertions exercise — and break against — the real DDL.

### 4. Idempotency pending TTL (60s → 5m)
The in-flight pending lock expired after 60s, but `CreateGroup` calls `IMCreateOrUpdateChannel` over an HTTP client with **no timeout**, so a degraded WuKongIM can block the create well past 60s. The lock would then expire mid-flight and a concurrent same-key retry could acquire a fresh lock and run a parallel create → duplicate group. There is no provably-safe fixed TTL (create latency is unbounded without an IM-call timeout), so this widens the TTL to 5m — enough to cover realistic degradation and shrink the duplicate window to near zero. Trade-off: a same-key retry sees `409 in-flight` for longer after a between-commit-and-finalize crash, which is preferred over duplicate creates. Fully closing it needs an IM-call timeout or lock renewal — tracked in #381.

## Test plan
Run against MySQL + Redis + WuKongIM (all green locally and in CI):
- `go test ./modules/group/ -run TestGroupNameWidenMigration`
- `go test ./modules/integration/` (full package, no regression), incl. `TestIntegrationCreateTeamGroupSuccess`, `TestIntegrationCreateTeamGroupIdempotency`, and the new `TestIntegrationTeamGroupExistsRejectsRobotCreator`
- `gofmt` / `go vet` clean

No error codes or response envelopes changed, so no i18n regeneration needed.

## Not in this PR (remaining #381 items)
- **Compensating-delete non-atomicity (orphan group)** — the one cross-cutting data-consistency item; pre-existing shared-layer behavior (Web/Bot too), warrants its own issue + async reconcile.
- **Payload fingerprint serialization / no DB uniqueness fallback / zero-width whitespace in group name** — low priority; the uniqueness backstop in particular would change product semantics (same-name groups are legal), so documenting "send an Idempotency-Key for concurrent retries" is preferred over a DB constraint.

Closes #384

Refs #381 (umbrella issue; remaining items tracked there)

### DIFF
--- a/modules/group/group_name_widen_test.go
+++ b/modules/group/group_name_widen_test.go
@@ -1,15 +1,21 @@
 package group
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
 	"testing"
 
 	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
 	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	migrate "github.com/rubenv/sql-migrate"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+// groupNameWidenMigrationFile is the migration whose Up DDL this test exercises. Kept as a
+// named constant so the test fails loudly (file-not-found) if the migration is ever renamed.
+const groupNameWidenMigrationFile = "20260615000001_group_name_widen.sql"
 
 // TestGroupNameWidenMigration genuinely exercises the VARCHAR(40)->(50) in-place widen
 // against PRE-EXISTING data, rather than just asserting a 40-rune string fits a 50-wide
@@ -38,10 +44,28 @@ func TestGroupNameWidenMigration(t *testing.T) {
 	}
 	modifyNameWidth := func(width int) {
 		// DDL column length can't be a bound placeholder; width is a controlled int constant.
+		// Used only as test scaffolding (shrink to the pre-widen world / restore on cleanup) —
+		// the widen under test runs the migration file's OWN Up DDL via applyMigrationUp below.
 		_, err := ctx.DB().UpdateBySql(
 			fmt.Sprintf("ALTER TABLE `group` MODIFY `name` VARCHAR(%d) NOT NULL DEFAULT ''", width),
 		).Exec()
 		require.NoError(t, err)
+	}
+	// applyMigrationUp parses the real migration file from the embedded sql FS and executes its
+	// Up statements verbatim — the same source rubenv/sql-migrate runs in production. This is
+	// what keeps the test honest: if 20260615000001_group_name_widen.sql ever drifts (different
+	// width, added clause, etc.), the assertions below run against the changed DDL instead of a
+	// stale hand-inlined copy.
+	applyMigrationUp := func(filename string) {
+		data, err := sqlFS.ReadFile("sql/" + filename)
+		require.NoError(t, err, "read migration %s from embedded sql FS", filename)
+		m, err := migrate.ParseMigration(filename, bytes.NewReader(data))
+		require.NoError(t, err, "parse migration %s", filename)
+		require.NotEmpty(t, m.Up, "migration %s has no Up statements", filename)
+		for _, stmt := range m.Up {
+			_, err := ctx.DB().UpdateBySql(stmt).Exec()
+			require.NoError(t, err, "exec Up stmt of %s: %s", filename, stmt)
+		}
 	}
 
 	// Migration applied → column is already VARCHAR(50).
@@ -56,8 +80,8 @@ func TestGroupNameWidenMigration(t *testing.T) {
 	_, err := ctx.DB().InsertBySql("INSERT INTO `group` (group_no, name) VALUES (?, ?)", legacyNo, name40).Exec()
 	require.NoError(t, err)
 
-	// Run the widen (the migration's Up DDL) over the existing data.
-	modifyNameWidth(MaxGroupNameLen)
+	// Run the widen by executing the migration file's OWN Up DDL (not a hand-inlined copy).
+	applyMigrationUp(groupNameWidenMigrationFile)
 	require.Equal(t, MaxGroupNameLen, colLen(), "widen must take effect")
 
 	// Existing 40-rune data survives the in-place widen untouched.
--- a/modules/integration/api_groups.go
+++ b/modules/integration/api_groups.go
@@ -23,9 +23,13 @@ const (
 	maxTeamGroupMembers = 50
 	// idempotencyKeyHeader 客户端可选幂等 header（Stripe 式，缺省不保证幂等）。
 	idempotencyKeyHeader = "Idempotency-Key"
-	// idempotencyPendingTTL in-flight 占位 TTL；进程在「建群成功」与「写终值」之间崩溃
-	// 时最多保留这么久（之后过期，同 key 重试可能再建一个群——已知边界，可接受）。
-	idempotencyPendingTTL = 60 * time.Second
+	// idempotencyPendingTTL in-flight 占位锁的存活时间，需覆盖一次建群请求从占坑到写终值的最坏
+	// 处理时长。但 IMCreateOrUpdateChannel 走无客户端超时的 HTTP（IM 退化时服务端 goroutine 可长时间
+	// 阻塞），建群延迟没有硬上界，故不存在「绝对安全」的固定值。这里取一个工程上足以覆盖现实退化的
+	// 值，把「慢建群致 pending 锁提前过期 → 并发同 key 重试重复建群」的窗口收窄到极小。代价是进程在
+	// 「建群成功、未写终值」之间崩溃后，同 key 重试在该 TTL 内拿 409 in-flight 而非回放（宁可让客户端
+	// 稍后重试也不重复建群）。要彻底消除需给 IM 调用加超时或持锁续租——见 #381。
+	idempotencyPendingTTL = 5 * time.Minute
 	// idempotencyDoneTTL 终值（可回放）记录 TTL。
 	idempotencyDoneTTL = 24 * time.Hour
 )
--- a/modules/integration/api_groups.go
+++ b/modules/integration/api_groups.go
@@ -260,6 +260,16 @@ func (it *Integration) teamGroupExists(groupNo, spaceID, uid string) (bool, erro
 	if !found || status != group.GroupStatusNormal || creator != uid {
 		return false, nil
 	}
+	// 防御性对称（与 createGroup 的 owner 必须是人类一致）：建群时已校验 creator 是真人，正常情况
+	// 下 creator==uid 即意味着 uid 是人类；这里再复核 uid 当前仍是人类账号，让两个端点的人类约束
+	// 对称，兜住「账号被改成 robot 标记」这类极罕见的事后变更（非人类账号一律 exists=false）。
+	human, err := it.db.isHumanUser(uid)
+	if err != nil {
+		return false, err
+	}
+	if !human {
+		return false, nil
+	}
 	active, err := it.groupService.ExistMemberActive(groupNo, uid)
 	if err != nil {
 		return false, err
--- a/modules/integration/api_groups_test.go
+++ b/modules/integration/api_groups_test.go
@@ -80,8 +80,13 @@ func TestIntegrationCreateTeamGroupSuccess(t *testing.T) {
 	assert.Equal(t, uid, resp.OwnerUserID)
 	assert.Equal(t, "团队群", resp.Name)
 	assert.ElementsMatch(t, []string{botA, botB}, resp.MemberRobotIDs)
-	_, err := time.Parse(time.RFC3339, resp.CreatedAt)
+	createdAt, err := time.Parse(time.RFC3339, resp.CreatedAt)
 	assert.NoError(t, err, "created_at must be RFC3339, got %q", resp.CreatedAt)
+	// 0001-01-01T00:00:00Z (Go zero time) is itself valid RFC3339, so the parse check alone
+	// would pass on the regression. Require a real, recent timestamp instead.
+	assert.False(t, createdAt.IsZero(), "created_at must not be the zero time, got %q", resp.CreatedAt)
+	assert.WithinDuration(t, time.Now(), createdAt, 5*time.Minute,
+		"created_at must be close to now, got %q", resp.CreatedAt)
 
 	// 落库断言：owner=creator role，bot role=common + robot=1，无 bot_admin。
 	members := queryTeamGroupMembers(t, ctx, resp.GroupID)
@@ -361,6 +366,49 @@ func TestIntegrationTeamGroupExists(t *testing.T) {
 		Where("group_no=?", groupNo).Exec()
 	require.NoError(t, err)
 	assert.False(t, checkExists(groupNo).Exists)
+}
+
+// TestIntegrationTeamGroupExistsRejectsRobotCreator is the existence-side mirror of
+// TestIntegrationCreateTeamGroupRejectsRobotOwner: createGroup refuses a robot owner up front,
+// so groupExists must apply the same human guard to keep the two endpoints symmetric. We
+// create a group as a human, then flip the OIDC-linked account to a robot (AuthByKey still
+// authenticates — it ignores the robot flag) and assert the group no longer "exists".
+func TestIntegrationTeamGroupExistsRejectsRobotCreator(t *testing.T) {
+	route, ctx, mp := setupIntegrationAPITest(t)
+	subject := "sub-exists-robot"
+	uid := seedIntegrationUser(t, ctx, mp.Issuer, subject)
+	spaceID := "sp_" + util.GenerUUID()[:8]
+	seedSpaceMembership(t, ctx, uid, spaceID, "Team", 1, "2026-01-01 10:00:00")
+	bot := "bot_" + util.GenerUUID()[:8]
+	seedOwnBot(t, ctx, uid, spaceID, bot, "")
+	apiKey := exchangeTeamGroupKey(t, route, mp, subject, spaceID)
+
+	w := httptest.NewRecorder()
+	route.ServeHTTP(w, integrationRequest(t, http.MethodPost, "/v1/integrations/oidc/groups", apiKey, map[string]interface{}{
+		"name":             "团队群",
+		"member_robot_ids": []string{bot},
+	}))
+	require.Equal(t, http.StatusOK, w.Code, w.Body.String())
+	var created createGroupResp
+	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &created))
+
+	checkExists := func() bool {
+		ww := httptest.NewRecorder()
+		route.ServeHTTP(ww, integrationRequest(t, http.MethodGet, "/v1/integrations/oidc/groups/"+created.GroupID, apiKey, nil))
+		require.Equal(t, http.StatusOK, ww.Code, ww.Body.String())
+		var r groupExistsResp
+		require.NoError(t, json.Unmarshal(ww.Body.Bytes(), &r))
+		return r.Exists
+	}
+
+	// Human creator + active member → exists:true.
+	assert.True(t, checkExists())
+
+	// Flip the creator's account to a robot. The key still authenticates (AuthByKey ignores the
+	// robot flag), but the human guard in teamGroupExists must now report exists:false.
+	_, err := ctx.DB().Update("user").Set("robot", 1).Where("uid=?", uid).Exec()
+	require.NoError(t, err)
+	assert.False(t, checkExists(), "a robot creator must not confirm the group (symmetric with create)")
 }
 
 // TestIntegrationTeamGroupExistsIsSpaceScoped verifies the existence check stays inside the
--- a/modules/integration/db.go
+++ b/modules/integration/db.go
@@ -175,13 +175,26 @@ func (d *integrationDB) queryAvailableBotSpaces(uid string, spaceIDs []string) (
 // 格式（非 RFC3339），故这里直读原始列，由调用方 .UTC().Format(time.RFC3339) 输出。按 space_id
 // 一并限定，与同模块 queryGroupStatus 的 Space 隔离口径一致（防御性，避免跨 Space 直读）。
 func (d *integrationDB) queryGroupCreatedAt(groupNo, spaceID string) (time.Time, error) {
-	var createdAt time.Time
+	// Must scan into a struct field (mirroring queryGroupStatus), NOT a bare time.Time:
+	// time.Time is itself a struct and implements no sql.Scanner, so dbr's reflection treats
+	// LoadOne(&t) as "find a column named created_at inside time.Time", finds none, scans into
+	// a throwaway dummy dest, and yields the zero value with a nil error. dbr only binds the
+	// column when it can match it to a named field of a wrapper struct.
+	var row struct {
+		CreatedAt time.Time `db:"created_at"`
+	}
 	err := d.session.Select("created_at").From("`group`").
-		Where("group_no=? AND space_id=?", groupNo, spaceID).LoadOne(&createdAt)
+		Where("group_no=? AND space_id=?", groupNo, spaceID).LoadOne(&row)
 	if err != nil {
 		return time.Time{}, fmt.Errorf("integration: query group created_at %q: %w", groupNo, err)
 	}
-	return createdAt, nil
+	if row.CreatedAt.IsZero() {
+		// Defense-in-depth: a zero timestamp is never a real group creation time. Signal a read
+		// failure so the caller's server-time fallback fires instead of leaking the Go zero time
+		// (0001-01-01T00:00:00Z) into the response contract.
+		return time.Time{}, fmt.Errorf("integration: query group created_at %q: scanned zero time", groupNo)
+	}
+	return row.CreatedAt, nil
 }
 
 // queryGroupStatus returns the group's status and creator, scoped to the given space, and


lml2468 (dismissed):
✅ APPROVE — 精准的 follow-up fix:修了一个 created_at 零值真 bug + 收了 #377 review 的两条 should-fix。inline 逐项核实 + build/vet 过,无 blocking。

## created_at 真 bug 修对(PR 主项,根因诊断精准)
原 queryGroupCreatedAt 用 `LoadOne(&createdAt)`(bare time.Time):dbr 反射把 time.Time(本身是 struct、不实现 sql.Scanner)当成"在 time.Time 内找 created_at 列",找不到 → scan 进 throwaway dummy → **返回零值 + nil error**。结果 created_at 永远是 0001-01-01(还是合法 RFC3339,所以原 parse 测试没抓到)。
修法:scan 进带 `db:"created_at"` 的 wrapper struct(mirror queryGroupStatus)+ zero-time defense(返 error 让调用方 fallback)。调用方 api_groups.go:206-211 默认 time.Now().UTC()、query 失败/零值时 fallback server time——即便 query 出问题也返合理时间,不再泄露 Go 零时间进契约。注释把 dbr 这个坑解释得很清楚。

## 收了 #377 review 两条 should-fix
- ✅ **groupExists human guard**(mochashanyao #377 提的"与 createGroup 不对称"):加 isHumanUser(uid)——fail-closed(user 不存在/NotFound→false),兜"账号事后被改 robot 标记"。配对称回归测试 TestIntegrationTeamGroupExistsRejectsRobotCreator。
- ✅ **migration 测试改用真实 SQL**(mochashanyao #377 提的"测试未引用实际迁移、可能不同步"):applyMigrationUp 从 embedded sql FS 解析真实迁移文件的 Up DDL 执行(而非 hand-inline 一份),迁移文件漂移时测试会跟着变/失败。
- ✅ created_at 测试加强:断言非 zero time + WithinDuration(now, 5min),防"0001-01-01 也是合法 RFC3339"的回归漏检。

build ./modules/integration/ ./modules/group/ + vet 过。

精准修了 dbr 反射陷阱的 created_at 零值 bug,且把 #377 的两条 should-fix(human guard 对称、migration 测试真实性)一并收齐。高质量 follow-up。批准。

—— Octo-PR (lml2468)

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#383 Review Report
**Head SHA:** `b4abee3a1ac9bdbf42a7fd6a84b95965ec206be5`
**PR:** fix(integration): correct team-group created_at + two #381 follow-ups

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| created_at fix (db.go) | ✅ | `queryGroupCreatedAt` now scans into `struct{CreatedAt time.Time \`db:"created_at"\`}`, zero-time guard returns error → caller fallback fires |
| isHumanUser guard (api_groups.go) | ✅ | `teamGroupExists` adds `isHumanUser(uid)` check symmetric with `createGroup`; error propagation correct |
| Migration test (group_name_widen_test.go) | ✅ | `applyMigrationUp` reads from `sqlFS` (embed.FS), parses via `rubenv/sql-migrate`, executes real DDL |
| Test coverage | ✅ | `TestIntegrationCreateTeamGroupSuccess` now asserts non-zero + within-5min; new `TestIntegrationTeamGroupExistsRejectsRobotCreator` covers the guard |

## 2. Findings

### No P0/P1 findings.

### P2 Observations

**P2-1: `queryGroupCreatedAt` zero-time error message could be more specific**
- **Diff-scope:** new (introduced by this PR's fix)
- The zero-time path returns `fmt.Errorf("...scanned zero time")` — a non-wrapped sentinel. The caller in `doCreateTeamGroup` (api_groups.go:~195) treats any `cErr != nil` identically (warn + fallback to `time.Now()`), so this has no functional impact. But if a future caller needs to distinguish "row not found" from "row found but zero", there's no sentinel to match. Consider using `errors.Is`-compatible wrapping if the contract evolves.
- **Severity:** P2 (maintainability, no user-visible impact)

**P2-2: `isHumanUser` extra DB round-trip on every existence check**
- `teamGroupExists` now makes 3 sequential DB queries: `queryGroupStatus` → `isHumanUser` → `ExistMemberActive`. The `isHumanUser` call adds one `SELECT robot FROM user WHERE uid=?` per GET request. For the existence endpoint (likely polled), this is a minor latency/cost addition. The PR acknowledges this is defense-in-depth ("not exploitable today").
- **Severity:** P2 (performance, acceptable for the security posture gained)

## 3. Suggestions

- Consider caching the `isHumanUser` result for the duration of the request context if additional endpoints are added that need the same check. Not blocking.
- The `applyMigrationUp` helper in the test is well-structured. If more migration tests are added, consider extracting it to a shared test helper in `testutil`.

## 4. Extra Findings

None beyond the P2 observations above.

## 5. Data-Flow Tracing

### Fix 1: `queryGroupCreatedAt` → `doCreateTeamGroup` → response

```
SQL: SELECT created_at FROM `group` WHERE group_no=? AND space_id=?
  ↓ dbr LoadOne(&row) where row = struct{CreatedAt time.Time `db:"created_at"`}
  ↓ row.CreatedAt populated correctly (dbr matches column name to struct field tag)
  ↓ row.CreatedAt.IsZero() check:
  ↓   zero → return error → caller logs warn → createdAt = time.Now().UTC()
  ↓   non-zero → return row.CreatedAt → caller uses t.UTC()
  ↓ createdAt.Format(time.RFC3339) → response JSON
```
**Verdict:** Data flows correctly. The regression (zero time leaking) is fixed. ✅

### Fix 2: `isHumanUser` in `teamGroupExists`

```
teamGroupExists(groupNo, spaceID, uid):
  ↓ queryGroupStatus(groupNo, spaceID) → (status, creator, found)
  ↓ gate: !found || status!=normal || creator!=uid → false
  ↓ isHumanUser(uid) → SELECT robot FROM user WHERE uid=?
  ↓   robot==0 → true (human) → continue
  ↓   robot!=0 → false → return (false, nil)
  ↓   ErrNotFound → false → return (false, nil)
  ↓   other error → return (false, err) → 500
  ↓ ExistMemberActive(groupNo, uid) → final check
```
**Verdict:** Data flows correctly. Error paths properly handled. ✅

### Fix 3: `applyMigrationUp` in test

```
sqlFS (embed.FS, declared in 1module.go:17 with //go:embed sql)
  ↓ ReadFile("sql/20260615000001_group_name_widen.sql")
  ↓ migrate.ParseMigration(filename, reader) → parses Up/Down sections
  ↓ m.Up = ["ALTER TABLE `group` MODIFY `name` VARCHAR(50) NOT NULL DEFAULT ''"]
  ↓ for each stmt: ctx.DB().UpdateBySql(stmt).Exec()
  ↓ colLen() assertion verifies VARCHAR(50)
```
**Verdict:** Data flows correctly. Migration file exists in embedded FS. ✅

## 6. Blind-Spot Checklist (R5 — security_sensitive)

**C1 — Dual-path parity:**
- `createGroup` ↔ `teamGroupExists`: The PR explicitly adds `isHumanUser` to `teamGroupExists` to achieve symmetry with `createGroup`. Both paths now enforce: space membership → human user → (create: bot validation / exists: active member check). **Clear — parity achieved.**
- No other symmetric pairs touched by this diff.

**C2 — Control-flow ordering / nested reuse:**
- `isHumanUser` is called in exactly 2 places: `createGroup` (line ~113) and `teamGroupExists` (line ~270). No nesting, no double-call risk. **Clear.**
- `queryGroupCreatedAt` is called in exactly 1 place: `doCreateTeamGroup` (line ~195). No reuse/nesting concern. **Clear.**
- No security controls (regex/escape/sanitize) are modified. **N/A.**

- Both endpoints are behind `userAPIKeyAuth` middleware (verified in `TestIntegrationTeamGroupEndpointsRejectUnauthenticated`). The `isHumanUser` check is an additional authorization layer within the authenticated handler. No new endpoints or capabilities are exposed. **Clear.**

---

## [Octo-Q] verdict: APPROVE

All three changes are correct, well-tested, and properly scoped. The `created_at` regression fix addresses a real user-visible bug (zero time in response). The `isHumanUser` guard adds defense-in-depth symmetry. The migration test improvement eliminates DDL drift risk. No P0/P1 findings.

Jerry-Xin (dismissed):
**APPROVE** — the `created_at` correctness fix is right and both PR-local #381 follow-ups are properly addressed. Code CI (Build/Vet/Lint/Test) is green at this head; the remaining red checks are non-code noise (check-sprint, dependency-review, gitleaks).

### ✅ 1. `created_at` correctness fix — correct
The root-cause analysis in the PR body matches the code. `queryGroupCreatedAt` (`modules/integration/db.go:177`) previously scanned into a bare `time.Time`; dbr's reflection on a struct dest looks for a `created_at` *field inside* `time.Time`, finds none, scans into a throwaway dummy, and returns the zero value **with a nil error** — so the caller's `err != nil` fallback never fired and `0001-01-01T00:00:00Z` leaked into the response. The fix scans into `struct{ CreatedAt time.Time `db:"created_at"` }`, exactly mirroring the working `queryGroupStatus`, and adds an `IsZero()` defense-in-depth that converts a zero read into an error.

No drift reintroduced: the sole caller (`api_groups.go:206-209`) initializes `createdAt := time.Now().UTC()` and only overrides with `t.UTC()` on a clean read; both branches funnel through `.UTC().Format(time.RFC3339)`, so the IsZero guard at worst engages the same UTC-normalized server-time fallback that #377 established — never a timezone/RFC3339 mismatch. Verified `queryGroupCreatedAt` has exactly one caller repo-wide, and that caller treats any error as a benign fallback, so the new error path is safe.

### ✅ 2. `teamGroupExists` human-guard symmetry (#381 item)
`teamGroupExists` (`api_groups.go:260+`) now calls `isHumanUser(uid)` after the creator/status/found checks, mirroring `createGroup`'s up-front robot-owner rejection. Guard is correctly placed (still creator-scoped first), error is propagated, and a non-human returns `exists:false` (not an error) — consistent with the `!found` semantics, no info leak. Matches the #381 ask exactly; defense-in-depth, not exploitable today.

### ✅ 3. Migration test no longer inlines DDL (#381 item)
`group_name_widen_test.go` now parses the real `sql/20260615000001_group_name_widen.sql` via `rubenv/sql-migrate` (the same source production runs) instead of a hand-inlined `ALTER TABLE`. The hand-inlined `MODIFY` is now scoped purely to test scaffolding (shrink/restore), and a named-constant filename makes the test fail loudly on rename. Keeps the assertions honest against real DDL drift.

### ✅ Tests
- created_at: `TestIntegrationCreateTeamGroupSuccess` strengthened — the old `time.Parse(RFC3339)` check passed on the zero time (itself valid RFC3339); now asserts `!IsZero()` and `WithinDuration(now, 5m)`, which would have caught the regression.
- isHumanUser: new `TestIntegrationTeamGroupExistsRejectsRobotCreator` (create as human → flip account to robot → assert `exists:false`).
- migration: rewired to exercise the file's own Up statements.

### 🟡 Out of scope (correctly deferred, non-blocking)
The PR explicitly defers the remaining #381 items (compensating-delete non-atomicity, idempotency pending TTL, payload-fingerprint serialization, DB uniqueness fallback, zero-width whitespace). These are shared-layer / product-semantics changes that warrant separate work — agreed, not blockers here.

Scoped, well-reasoned, well-tested. LGTM.

yujiawei (dismissed):
# Code Review — PR #383 (octo-server)

Reviewed at head `b4abee3a1ac9bdbf42a7fd6a84b95965ec206be5` against merge-base `43ac0941` (4 files, +101/-6). This is a security-sensitive change (touches a human/robot guard on the integration OIDC `groups` endpoints), so the auth path was reviewed with extra care. `go vet ./modules/integration/... ./modules/group/...` is clean at this SHA.

## Verdict: Approve

All three changes are correct and well-scoped. No blocking (P0/P1) issues. A few advisory notes below; none should hold up the merge.

## What was verified ✅

**1. `queryGroupCreatedAt` zero-time fix (`modules/integration/db.go`)**
The root cause is correct: `dbr`'s `LoadOne(&bareTimeTime)` treats `time.Time` as a destination *struct* with no matching `created_at` field, scans into a throwaway dest, and returns the Go zero value with a `nil` error — so the old `err != nil` fallback never fired and `0001-01-01T00:00:00Z` leaked into the response. Scanning into `struct{ CreatedAt time.Time \`db:"created_at"\` }` matches the proven `queryGroupStatus` pattern and binds the column correctly. The added `IsZero()` guard cannot suppress a legitimate value — `group.created_at` is `NOT NULL DEFAULT CURRENT_TIMESTAMP`, so a real row is never zero — and on the zero/err path the sole caller (`doCreateTeamGroup`, api_groups.go:207) already logs a warning and falls back to server time. Fail-closed and correct.

**2. `teamGroupExists` human guard (`modules/integration/api_groups.go:260+`)**
Ordering is correct: the new `isHumanUser(uid)` check runs *after* `queryGroupStatus` gates `found && status==normal && creator==uid`, and *before* `ExistMemberActive`. `isHumanUser` returns `(false, nil)` on `ErrNotFound`/empty uid (→ `exists:false`, consistent with the always-200 GET contract) and propagates real DB errors (→ 500). This is legitimate symmetry with the owner-must-be-human check `createGroup` already enforces. The PR correctly characterizes it as defense-in-depth rather than an exploitable-today bypass (a `uk_` key's UID is fixed at exchange time and no `uk_`/OIDC-reachable path mutates `user.robot`).

**3. Migration test now runs the real Up DDL (`modules/group/group_name_widen_test.go`)**
`migrate.ParseMigration(id, io.ReadSeeker)` / `m.Up []string` usage is correct (verified against `rubenv/sql-migrate@v1.5.2`); `bytes.NewReader` satisfies `io.ReadSeeker`; `sqlFS` is reachable from package `group`. The test genuinely exercises the migration file's own `Up` statement instead of a hand-inlined copy, so future drift in the DDL is caught. The pre-existing `defer modifyNameWidth(MaxGroupNameLen)` (line 71) restores the shared schema on any mid-test failure, so there is no schema-pollution risk.

## Advisory notes (non-blocking, P2 — author's discretion)

- **`created_at` test does not isolate the fix path.** In `TestIntegrationCreateTeamGroupSuccess`, `assert.WithinDuration(time.Now(), createdAt, 5m)` passes whether the new struct-scan reads the DB value *or* `queryGroupCreatedAt` fails and `doCreateTeamGroup` falls back to `time.Now()`. It reliably kills the old zero-time regression (via the adjacent `assert.False(createdAt.IsZero())`), which is the important part — but it does not *prove* the struct-scan itself reads the column. A small unit test that inserts a fixed `created_at` and asserts `queryGroupCreatedAt` returns exactly that would close the gap. Optional.

- **dbr `LoadOne(&bareTime)` is a systemic footgun — but this was the only instance.** A codebase-wide audit of the ~90 `LoadOne(&x)` sites found all others scan into scalars or db-tagged wrapper structs; `queryGroupCreatedAt` was the lone bare-`time.Time` scan, and it's the one being fixed here. No follow-up needed, recorded for completeness.

- **Migration `Down` is lossy** (`UPDATE ... LEFT(name,40)` before the shrink). Out of scope for this PR (the test only runs `m.Up`), explicitly documented in the SQL comment, and the wide column is the lossless choice anyway — just a known operational fact for any future rollback with >40-char names persisted.

## Items to verify in CI (not reproducible in this review env)
The integration tests require live MySQL + Redis + WuKongIM, which were not available here; correctness of `dbr` time-column scanning and the MySQL online-DDL behavior are taken on static analysis + `go vet`, not a live run. Please confirm `go test ./modules/integration/` and `./modules/group/ -run TestGroupNameWidenMigration` are green in CI before merge.

lml2468 (approved):
✅ APPROVE（重审 @2097280）— 仅一处:把 idempotencyPendingTTL 60s→5min,收 yujiawei #377 提的"幂等锁 60s 可能不足"🟡。纯常量调整 + 诚实注释,无逻辑改动,build 过。

## 本 commit 变更(+7/-3,仅 api_groups.go)
idempotencyPendingTTL 60s → 5min,注释写得很到位:
- 说明无"绝对安全"固定值的根因:IMCreateOrUpdateChannel 走无客户端超时的 HTTP,IM 退化时建群延迟无硬上界。
- 5min 是工程上覆盖现实退化的值,把"慢建群致 pending 锁提前过期 → 并发同 key 重试重复建群"窗口收窄到极小。
- 代价讲清:进程在"建群成功、未写终值"间崩溃后,同 key 重试在 TTL 内拿 409 in-flight 而非回放——宁可客户端稍后重试也不重复建群(fail-safe,与上版幂等设计一致)。
- 彻底消除方向指向 #381(给 IM 调用加超时或持锁续租)。

合理的工程取舍,没引入新逻辑(只改 TTL 值)。

## 维持(上版已核)
created_at 零值真 bug 修复(dbr wrapper struct scan + zero defense)、groupExists human guard、migration 测试用真实 SQL——这些 2097280 未动,结论维持。

重投 APPROVE。(剩余非阻塞:isHumanUser 请求级缓存、zero-time 错误信息更具体——mochashanyao 🟡,可后续。)

—— Octo-PR (lml2468)

Jerry-Xin (approved):
✅ APPROVE — re-review of new commit `2097280e7085`. The delta is a single, well-reasoned constant change addressing the pending-TTL note.

**What changed (only `modules/integration/api_groups.go`, +7/-3):**
`idempotencyPendingTTL` raised from `60s` → `5 minutes`, addressing the earlier 🟡 note that 60s may be insufficient under IM-channel latency.

The rationale is correct: `IMCreateOrUpdateChannel` uses HTTP with no client timeout, so a degraded IM backend can block the create goroutine for an unbounded time → there is no provably-safe fixed TTL. 5 minutes is a sound engineering value that shrinks the "slow create → pending lock expires early → concurrent same-key retry creates a duplicate group" window to near-zero, at the acknowledged cost that a same-key retry within the TTL gets `409 in-flight` rather than a replay if the process crashed between commit and finalize. The comment correctly points to #381 for the complete fix (IM-call timeout or lock-lease renewal).

This is a strictly-conservative change — it only widens the in-flight lock window, no regression risk. Everything else from the prior review still stands:
- created_at zero-value bug fix (dbr bare-`time.Time` reflection trap → wrapper-struct scan + zero-time defense + server-time fallback)
- groupExists human-user guard (symmetry with createGroup)
- migration test now parses the real embedded SQL DDL
- strengthened created_at tests (non-zero + WithinDuration)

No blockers. Remaining #381 items (IM-call timeout / lock renewal, migration-rollback truncation) are correctly deferred as non-blocking follow-ups.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#383 Review Report
**Head SHA:** `2097280e708576977474eab34284223b34eab2ee`
**PR:** Mininglamp-OSS/octo-server#383
**Title:** fix(integration): correct team-group created_at + two #381 follow-ups

---

## 1. Verification Summary

| # | Change | Verdict | Evidence |
|---|--------|---------|----------|
| 1 | `queryGroupCreatedAt` struct-scan fix | ✅ Correct | `db.go:178-199` — scans into `struct{CreatedAt time.Time \`db:"created_at"\`}`, zero-check fallback |
| 2 | `teamGroupExists` isHumanUser guard | ✅ Correct | `api_groups.go:267-275` — symmetric with `createGroup:113-120` |
| 3 | `idempotencyPendingTTL` 60s→5m | ✅ Sound | `api_groups.go:28-34` — documented trade-off, narrows duplicate-create window |
| 4 | Migration test runs real DDL | ✅ Correct | `group_name_widen_test.go:57-68` — reads from `sqlFS` embed, parses via `rubenv/sql-migrate` |
| 5 | Test assertion tightening (created_at) | ✅ Correct | `api_groups_test.go:83-87` — non-zero + WithinDuration(5m) |
| 6 | New test: robot creator exists=false | ✅ Correct | `api_groups_test.go:371-413` — mirrors `TestIntegrationCreateTeamGroupRejectsRobotOwner` |

## 2. Findings

### F1 — `queryGroupCreatedAt` bare-time.Time scan regression (FIX)

**Diff-scope:** new fix (closes a pre-existing P1 bug introduced before this PR series).

**Root cause (confirmed):** dbr's `LoadOne(&createdAt)` where `createdAt` is `time.Time` (a struct with no `sql.Scanner`) causes dbr reflection to look for a `created_at` field *inside* `time.Time`, find none, scan into a throwaway destination, and return the zero value with nil error. The caller's `err != nil` fallback (`time.Now()`) never fires → every successful create returned `"0001-01-01T00:00:00Z"`.

**Fix verification:**
- Scan target changed to `struct{CreatedAt time.Time \`db:"created_at"\`}` — matches the pattern used by `queryGroupStatus` (`db.go:207-213`). ✅
- Zero-time defense: `row.CreatedAt.IsZero()` → returns error → caller's `time.Now().UTC()` fallback engages (`api_groups.go:213-218`). ✅
- Error wrapping: `fmt.Errorf("...: %w", err)` preserves `dbr.ErrNotFound` chain for callers that need it (though the current caller treats all errors uniformly — correct). ✅

**Data-flow trace:**
```
SQL: SELECT created_at FROM `group` WHERE group_no=? AND space_id=?
  → dbr LoadOne(&row) scans into row.CreatedAt via `db:"created_at"` tag
  → row.CreatedAt.IsZero() guard catches any residual zero-value
  → return row.CreatedAt (non-zero time.Time)
  → doCreateTeamGroup: createdAt = t.UTC()
  → response: createdAt.Format(time.RFC3339) → real timestamp
```

**Severity:** P1 fix (user-visible incorrect data on every create response). The fix is correct and complete.

### F2 — `teamGroupExists` isHumanUser symmetric guard

**Diff-scope:** new (defense-in-depth, not fixing an exploitable defect).

**Verification:**
- `isHumanUser(uid)` queries `SELECT robot FROM user WHERE uid=?` (`db.go:222-231`). Returns `robot == 0`. ✅
- Error propagation: DB error → `return false, err` → `groupExists` returns 500. Correct. ✅
- Empty UID: `isHumanUser("")` returns `(false, nil)` — safe default. ✅
- Placement: after `creator != uid` check (which already implies uid is the creator who was human at create time), before `ExistMemberActive`. Ordering is correct — avoids unnecessary membership queries for robot accounts. ✅

**Data-flow trace:**
```
teamGroupExists(groupNo, spaceID, uid)
  → queryGroupStatus: status, creator, found
  → guard: !found || status != Normal || creator != uid → false
  → isHumanUser(uid): SELECT robot FROM user WHERE uid=?
  → guard: !human → false
  → ExistMemberActive(groupNo, uid) → bool
```

**Severity:** P2 (defense-in-depth improvement). The PR correctly identifies this as non-exploitable today since `createGroup` already enforces the human guard at creation time. The symmetric guard catches the extremely rare case of a post-creation robot-flag flip.

### F3 — `idempotencyPendingTTL` 60s → 5m

**Diff-scope:** new (widening a pre-existing constant).

**Verification:**
- `idemReserve` uses `idempotencyPendingTTL` as the TTL for `SetNX` (`api_groups.go:302`). ✅
- Trade-off documented: longer 409-in-flight window after crash vs. narrower duplicate-create window during slow IM calls. ✅
- No new code paths introduced — only the constant value changes. ✅

**Risk assessment:** The PR correctly identifies that no provably-safe fixed TTL exists without an IM-call timeout. 5 minutes is a reasonable engineering choice. The remaining gap (IM-call timeout / lock renewal) is tracked in #381.

**Severity:** P2 (risk reduction, not a correctness fix).

### F4 — Migration test: real DDL execution

**Diff-scope:** new (test infrastructure improvement).

**Verification:**
- `sqlFS` is `//go:embed sql` (`1module.go:17`), embedding the `sql/` directory. ✅
- `sqlFS.ReadFile("sql/20260615000001_group_name_widen.sql")` resolves correctly. ✅
- `migrate.ParseMigration` parses the `-- +migrate Up` section → single statement: `ALTER TABLE \`group\` MODIFY \`name\` VARCHAR(50) NOT NULL DEFAULT ''`. ✅
- Migration file content verified: matches the expected widen DDL. ✅
- `modifyNameWidth` (hand-inlined DDL) is retained only for test scaffolding (shrink to pre-widen state / cleanup defer). The widen-under-test now goes through the real migration. ✅

**Severity:** P2 (test quality improvement).

### F5 — Test assertion tightening: created_at non-zero + WithinDuration

**Diff-scope:** new (tightening a pre-existing weak assertion).

**Verification:**
- Old: `time.Parse(time.RFC3339, resp.CreatedAt)` — zero time `0001-01-01T00:00:00Z` is valid RFC3339, so the bug slipped through. ✅ (correct diagnosis)
- New: `createdAt.IsZero()` check + `WithinDuration(time.Now(), createdAt, 5*time.Minute)`. ✅
- 5-minute window is generous enough for slow CI but tight enough to catch the zero-time regression. ✅

**Severity:** P2 (test quality improvement).

### F6 — New test: `TestIntegrationTeamGroupExistsRejectsRobotCreator`

**Diff-scope:** new.

**Verification:**
- Creates group as human → `exists:true`. ✅
- Flips creator to robot via `UPDATE user SET robot=1`. ✅
- Asserts `exists:false` — exercises the new `isHumanUser` guard through the full HTTP path. ✅
- AuthByKey still authenticates (ignores robot flag) — test correctly relies on this to reach the handler logic. ✅

**Severity:** P2 (test coverage for new guard).

## 3. Suggestions

None blocking. Minor observations:

- **S1 (nit):** The `queryGroupCreatedAt` zero-time defense comment says "a zero timestamp is never a real group creation time" — this is correct for the `group` table's `created_at` column (set by `NOW()` or application time at insert). No action needed.

## 4. Extra Findings

None. The PR is tightly scoped to the four stated changes with no unintended side effects.

## 5. Data-Flow Trace Summary

|---|---|---|
| `row.CreatedAt` (db.go:193) | `SELECT created_at FROM \`group\`` via dbr struct-scan | ✅ — struct field with `db:"created_at"` tag binds correctly |
| `queryGroupCreatedAt` return value | `doCreateTeamGroup` (api_groups.go:213-218) | ✅ — error → `time.Now().UTC()` fallback; success → `.UTC()` |
| `human` bool (api_groups.go:270) | `isHumanUser(uid)` → `SELECT robot FROM user WHERE uid=?` | ✅ — `robot==0` → true; DB error → propagated |
| `idempotencyPendingTTL` (api_groups.go:34) | `idemReserve` → `SetNX(..., idempotencyPendingTTL)` | ✅ — constant consumed directly |
| `sqlFS.ReadFile` (test:60) | `//go:embed sql` (1module.go:17) | ✅ — embed FS includes `sql/` directory |
| `m.Up` statements (test:65) | `migrate.ParseMigration` parses `-- +migrate Up` section | ✅ — single ALTER TABLE statement |

**C1 — Dual-path parity:**
- `createGroup` ↔ `teamGroupExists` human guard: **now symmetric** (this PR's F2). ✅
- `createGroup` has space-membership check (step 3) that `teamGroupExists` doesn't mirror — **correct asymmetry**: existence check is creator-scoped via `queryGroupStatus.creator==uid`, not membership-scoped. No parity gap.
- Idempotency: `idemReserve` (SETNX) ↔ `idemFinalize` (SET overwrite) — symmetric acquire/release. ✅
- No create↔delete parity changes in this PR.

- `isHumanUser` in `teamGroupExists` placed after `creator!=uid` guard, before `ExistMemberActive` — correct ordering (early rejection, avoids unnecessary queries). ✅
- `queryGroupCreatedAt` zero-check runs after successful scan — correct (defense-in-depth on the scan result, not a pre-filter). ✅
- No safety controls (regex/escape/sanitize) modified — N/A for non-canonical form testing.

- Both endpoints gated by `userAPIKeyAuth` middleware (verified by `TestIntegrationTeamGroupEndpointsRejectUnauthenticated`). ✅
- `createGroup`: space-membership + human-owner + bot-ownership checks. ✅
- `teamGroupExists`: space-scoped (via `queryGroupStatus` space_id filter) + creator-scoped + human guard (new). ✅
- No new endpoints or capabilities exposed in this PR.

---

## Verdict

All four changes are correct and well-scoped. The `created_at` fix (F1) addresses a real P1 correctness bug (every create returned zero-time); the fix is verified through data-flow tracing from SQL scan → struct field → caller fallback → response formatting. The remaining three changes (F2–F6) are P2 improvements: defense-in-depth, risk reduction, and test quality. No P0 or P1 defects found in the PR's own changes.

**[Octo-Q] verdict: APPROVE** — no blocking findings. All changes verified correct through data-flow tracing and cross-file analysis.

yujiawei (approved):
# Code Review — PR #383 (octo-server)

Reviewed at head SHA `2097280e708576977474eab34284223b34eab2ee` against merge-base `43ac0941`. Four files, +108/-9. This is a security-sensitive change (auth guard + migration), so it got extra scrutiny.

## Verdict: APPROVED

All three stated fixes are correct and well-reasoned. The findings below are advisory (P2) or pre-existing behavior outside this PR's scope — none of them block merge. One documentation/scope note is worth a maintainer's eye before merging.

## What the PR does, verified

### 1. `created_at` zero-time bug — correct fix (`modules/integration/db.go`)
The root-cause analysis is accurate. `queryGroupCreatedAt` previously did `LoadOne(&createdAt)` into a bare `time.Time`. dbr's reflection only binds a column when it matches a *named struct field*; a bare `time.Time` (no `sql.Scanner`) gets a throwaway dest and yields the Go zero value with a nil error, so the `err != nil` server-time fallback never fired and `0001-01-01T00:00:00Z` leaked into the response.

The fix scans into a wrapper `struct{ CreatedAt time.Time `db:"created_at"` }`, mirroring the existing `queryGroupStatus` pattern (`db.go`), and adds a defense-in-depth `IsZero()` → read-failure conversion so the caller's `time.Now()` fallback engages. Both branches normalize to UTC before formatting. Correct.

I confirmed the `group.created_at` column is `timeStamp NOT NULL DEFAULT CURRENT_TIMESTAMP` (`modules/group/sql/20191106000002_group_legacy01.sql`), so the new struct-field scan cannot hit a NULL-scan error — the column is never NULL.

### 2. Symmetric human guard in `teamGroupExists` — correct defense-in-depth (`modules/integration/api_groups.go`)
`createGroup` already requires a human owner; the existence endpoint did not re-check. The new `isHumanUser(uid)` guard makes the two endpoints symmetric, so an account flipped to a robot after group creation no longer confirms the group. I verified the premise: `AuthByKey` → `isActiveUser` checks only `status<>0 AND is_destroy=0` and ignores the `robot` flag, so a robot-flipped account would still authenticate — the guard is genuinely load-bearing, not dead code. The new mirror test exercises exactly this path. Errors propagate (→ 500), non-human → `exists=false`. Correct.

### 3. Migration test runs the real Up DDL — genuine improvement (`modules/group/group_name_widen_test.go`)
The test previously hand-inlined `ALTER TABLE ... MODIFY ... VARCHAR(50)`. It now reads the migration file from the embedded `sqlFS` and executes its parsed `Up` statements via `rubenv/sql-migrate`. I confirmed this is the same library production uses (`pkg/db/mysql.go`), so the test now breaks against real DDL drift instead of a stale copy. The named-constant filename makes a rename fail loudly. Good.

### 4. (Undeclared) idempotency pending TTL widened 60s → 5min (`api_groups.go:32`)
The head commit widens `idempotencyPendingTTL`. The rationale is sound: `IMCreateOrUpdateChannel` runs over HTTP with no client timeout, so create latency has no hard upper bound; a 60s pending lock could expire mid-create and let a concurrent same-key retry create a duplicate group. Widening to 5min shrinks that duplicate-create window. The tradeoff (a crashed worker holds the in-flight lock longer post-crash) is explicitly documented in the comment and deferred to #381. Acceptable.

## Findings (all advisory — non-blocking)

- **[P2] Test cannot distinguish DB-read success from server-time fallback** (`api_groups_test.go:88`). The new assertions (`!IsZero()`, `WithinDuration(now, 5min)`) prove the response is recent, but production falls back to `time.Now()` on read failure (`api_groups.go:210`). If `queryGroupCreatedAt` regressed again, the fallback would still produce a recent timestamp and the test would pass. To truly lock the DB-read path, assert against the stored DB value (or assert the warn-log/fallback did not fire). This is the regression the PR is fixing, so it's worth making the test detect it directly.

- **[P2] `Retry-After` is hard-coded to 2s while pending can now last 5min** (`api_groups.go:330`). In the crash-poison-pill case, clients honoring the header retry every 2s and get 409 `idempotency_in_flight` for up to 5 minutes. Consider deriving Retry-After from the Redis PTTL or returning a more conservative backoff. Pre-existing behavior, made slightly more visible by the longer TTL.

- **[P2] Pending idempotency branch does not compare payload SHA** (`api_groups.go:328`). A reused key with a *different* body during the pending window returns retryable `in_flight` rather than a terminal conflict. Pre-existing; the longer TTL extends the window in which this can occur. Not introduced by this PR.

- **[P2 / informational] `idemFinalize` is last-finisher-wins** (`api_groups.go:353`). A request that outlives the (now longer) pending TTL can overwrite a later retry's record, since the pending value carries no owner nonce. This is real but **pre-existing and untouched by this diff** — the fix (owner token + Redis CAS, or a hard IM timeout / lock renewal) is correctly scoped to #381 in the PR description. Flagging for the #381 tracking, not as a change request here.

## Scope / documentation note (please confirm before merge)
The PR description lists **"Idempotency pending TTL"** explicitly under **"Not in this PR (remaining #381 items)"**, but the head commit `2097280e` changes exactly that constant. The change itself is fine and reasonable; the description just contradicts the diff. Recommend updating the PR body so the "Not in this PR" list no longer claims the TTL was left out (it's a 4th change beyond the 3 summarized). This is a description fix, not a code change.

## Coverage — what I could not verify from static review
- I did not run the integration suite (`go test ./modules/integration/`, `./modules/group/ -run TestGroupNameWidenMigration`) against live MySQL + Redis + WuKongIM, so the green-locally claim is taken on trust. `go vet` and test-binary compilation pass clean on both packages.
- Runtime dbr scan behavior against the live MySQL driver is reasoned from the codebase's existing pattern, not executed.
- The concurrent duplicate-create race and the crash-during-finalize path are reasoned about, not reproduced under load.

For a human reviewer: the security-relevant surface here is small and conservative (a tightening guard plus a read-path correctness fix); the only item I'd want eyes on is the idempotency TTL/crash-window tradeoff, which is consciously deferred to #381.

PR #385, opened 2026-06-16
feat(search): harden message read path, fail-closed authz, explicit backend

yujiawei:
## Summary
Hardens the message read/search path: a fail-closed security collar on the
legacy search path, an explicit search-backend switch across all search
surfaces, a new context-window endpoint, and a pagination depth cap. Builds on
the existing OpenSearch-backed `modules/messages_search`.

## Changes
- **Legacy path fail-closed**: `modules/message` `filterResultsBySpace` now drops
  unknown `channel_type` rows instead of appending them, closing a cross-Space
  leak in the legacy `/v1/message/search` results.
- **Explicit search backend** (`pkg/searchbackend`, `OCTO_SEARCH_BACKEND=es|zinc|disabled`):
  no auto-detect/degradation. `es` is authoritative and never falls through to
  Zinc; `disabled` makes every surface (`/v1/messages/_search*`,
  `/v1/message/search`, `/v1/search/global` message portion) refuse uniformly
  with a new localized error; unset preserves current behaviour. `appconfig`
  now exposes `search_enabled` so clients can hide the search box.
- **Context window endpoint** `POST /v1/messages/_search_around`: the anchor is
  pre-validated through the same channel-access, Space-scope, visibility and
  command/revoked gates as a normal read before any surrounding context is
  fetched (a not-visible / cross-Space / revoked / command anchor returns 404).
  Bidirectional `search_after` (reversed sort for the older wing, re-reversed to
  chronological order) reusing the oversample fill loop.
- **Pagination depth cap** aligned with OpenSearch `max_result_window` (10000),
  encoded as a signed cumulative count in the cursor and checked as
  `priorDepth + page_size`, so changing `page_size` cannot walk past it.
- **Read-only drift gauges** (`search_recon_*`) for the reconciliation job to
  publish ES↔MySQL count/sample drift into.
- Extensive table-driven tests covering the above (cross-Space, authz,
  revoke/visibles, page-size/depth caps, backend states, window assembly,
  response-field non-leakage, join-load bounds). Two end-to-end gates that
  depend on the indexer writing `spaceId`/`visibles` are written but skipped
  with an explicit reason.

## Verification
- `go build ./...` ✅
- `go test ./modules/messages_search/... ./modules/message/... ./modules/search/... ./pkg/searchbackend/...` ✅
- `make i18n-extract-check` ✅ · `make i18n-lint` ✅ · `go vet` (touched pkgs) ✅
- Independent model review (Codex): 3 findings fixed (backend default
  resolution, legacy-global-search Zinc gating, depth-cap page-size overrun) +
  1 fixed (command-anchor exclusion); remaining advisory note on cursor-key
  forgeability is the pre-existing cursor posture (resource limit, not an auth
  boundary; gated per-page + loud startup WARN when the signing secret is unset).

### DIFF
--- a/modules/common/api.go
+++ b/modules/common/api.go
@@ -22,6 +22,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
 	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
 	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
+	"github.com/Mininglamp-OSS/octo-server/pkg/searchbackend"
 	spacepkg "github.com/Mininglamp-OSS/octo-server/pkg/space"
 	"github.com/gin-gonic/gin"
 	"go.uber.org/zap"
@@ -401,6 +402,7 @@ func (cn *Common) appConfig(c *wkhttp.Context) {
 			SystemBotUIDs:          spacepkg.SystemBotList(),
 			LocalLoginOff:          boolToFlag(cn.systemSettings.LocalLoginOff()),
 			DisableUserCreateSpace: boolToFlag(cn.systemSettings.SpaceDisableUserCreate()),
+			SearchEnabled:          searchbackend.Resolve(cn.ctx.GetConfig().ZincSearch.SearchOn).SearchEnabled(),
 		})
 		return
 	}
@@ -439,6 +441,7 @@ func (cn *Common) appConfig(c *wkhttp.Context) {
 		SystemBotUIDs:          spacepkg.SystemBotList(),
 		LocalLoginOff:          boolToFlag(cn.systemSettings.LocalLoginOff()),
 		DisableUserCreateSpace: boolToFlag(cn.systemSettings.SpaceDisableUserCreate()),
+		SearchEnabled:          searchbackend.Resolve(cn.ctx.GetConfig().ZincSearch.SearchOn).SearchEnabled(),
 	})
 }
 
@@ -753,6 +756,14 @@ type appConfigResp struct {
 	// 实时性。后端 POST /v1/space/create 也走同一个 getter 校验,客户端隐藏
 	// 与服务端拒绝由单一真源驱动,不存在前后端漂移。
 	DisableUserCreateSpace int `json:"disable_user_create_space"`
+
+	// SearchEnabled 告知客户端当前部署是否启用消息搜索（OCTO_SEARCH_BACKEND
+	// != disabled）。为 false 时前端应隐藏搜索入口，而不是每次调用搜索接口
+	// 再吃一个 SEARCH_DISABLED 错误（YUJ-4667 步骤 2 / YUJ-4662 §7-#10）。
+	//
+	// 与 app_config.version 解耦的原因同 LocalLoginOff / DisableUserCreateSpace：
+	// 运维切 backend 后老客户端命中 version 短路分支也必须拿到最新值。
+	SearchEnabled bool `json:"search_enabled"`
 }
 
 type oidcProviderResp struct {
--- a/modules/message/api.go
+++ b/modules/message/api.go
@@ -34,6 +34,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
 	"github.com/Mininglamp-OSS/octo-server/pkg/mentionrewrite"
 	"github.com/Mininglamp-OSS/octo-server/pkg/richtext"
+	"github.com/Mininglamp-OSS/octo-server/pkg/searchbackend"
 	spacepkg "github.com/Mininglamp-OSS/octo-server/pkg/space"
 	appwkhttp "github.com/Mininglamp-OSS/octo-server/pkg/wkhttp"
 	"github.com/gocraft/dbr/v2"
@@ -1322,6 +1323,17 @@ func (m *Message) typing(c *wkhttp.Context) {
 
 // 搜索消息
 func (m *Message) search(c *wkhttp.Context) {
+	// search.backend 三态收口（YUJ-4667 步骤 2）：旧 /v1/message/search 走
+	// WuKongIM→Zinc，只有显式声明 backend=zinc 时才服务。
+	//   - disabled：搜索全关，统一返回 SEARCH_DISABLED（与新 _search* 表面一致）。
+	//   - es：OpenSearch 路径为权威读路径，旧 Zinc 路径不再服务（绝不让 es
+	//     部署把流量悄悄落回 Zinc 读到陈旧索引）；客户端应改打 _search*。
+	//   - zinc：维持旧行为（迁移/回滚过渡态）。
+	backend := searchbackend.Resolve(m.ctx.GetConfig().ZincSearch.SearchOn)
+	if !backend.LegacyZinc {
+		httperr.ResponseErrorL(c, errcode.ErrMessagesSearchDisabled, nil, nil)
+		return
+	}
 	var req struct {
 		UID         string `json:"uid"` // 搜索的消息限定这某个用户内
 		ChannelID   string `json:"channel_id"`
@@ -1431,7 +1443,15 @@ func (m *Message) filterResultsBySpace(results []map[string]interface{}, spaceID
 				filtered = append(filtered, r)
 			}
 		default:
-			filtered = append(filtered, r)
+			// 安全收口（fail-CLOSED）：未知 channel_type 无法判定其 Space 归属，
+			// 一律剔除，绝不放行。此前 `default: append` 会把任何无法识别的
+			// channel_type 直接放进结果，构成跨 Space 越权暴露面（搜索 A Space
+			// 时可命中 B Space / 未加入频道的消息）。新链 modules/messages_search
+			// 已 fail-CLOSED，此处把旧 Zinc/WuKongIM 读路径对齐到同一安全语义。
+			m.Warn("搜索结果命中未知 channel_type，已按 fail-closed 剔除",
+				zap.Int("channel_type", channelType),
+				zap.String("channel_id", chID),
+				zap.String("space_id", spaceID))
 		}
 	}
 	return filtered, nil
--- a/modules/message/api_search_space_filter_test.go
+++ b/modules/message/api_search_space_filter_test.go
@@ -1,0 +1,102 @@
+package message
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+	"github.com/Mininglamp-OSS/octo-server/modules/group"
+	"github.com/stretchr/testify/assert"
+)
+
+// N0 安全收口验证门（YUJ-4667）：旧 /v1/message/search 读路径的
+// filterResultsBySpace 此前 `default: append` 把无法识别的 channel_type 直接
+// 放行，构成跨 Space 越权暴露面。本组用例锁定 fail-CLOSED 语义：
+//   - 群聊(2)：仅命中 space 匹配的群；越权（B-only 群）剔除。
+//   - DM(1)：仅命中 payload.space_id 匹配的；越权剔除。
+//   - 未知 channel_type：一律剔除（核心安全收口，防回归）。
+
+// fakeSpaceGroupService 只实现 filterResultsBySpace 调用到的 GetGroups。
+type fakeSpaceGroupService struct {
+	group.IService
+	bySpace map[string]string // group_no -> space_id
+}
+
+func (f *fakeSpaceGroupService) GetGroups(groupNos []string) ([]*group.InfoResp, error) {
+	out := make([]*group.InfoResp, 0, len(groupNos))
+	for _, no := range groupNos {
+		out = append(out, &group.InfoResp{GroupNo: no, SpaceID: f.bySpace[no]})
+	}
+	return out, nil
+}
+
+func newSpaceFilterTestMessage(gs group.IService) *Message {
+	return &Message{
+		Log:          log.NewTLog("space-filter-test"),
+		groupService: gs,
+	}
+}
+
+func dmResultWithSpace(channelID, spaceID string) map[string]interface{} {
+	payload, _ := json.Marshal(map[string]interface{}{"space_id": spaceID})
+	return map[string]interface{}{
+		"channel_id":   channelID,
+		"channel_type": float64(1),
+		"payload":      base64.StdEncoding.EncodeToString(payload),
+	}
+}
+
+func spaceFilterChannelIDs(results []map[string]interface{}) []string {
+	out := make([]string, 0, len(results))
+	for _, r := range results {
+		id, _ := r["channel_id"].(string)
+		out = append(out, id)
+	}
+	return out
+}
+
+// 场景：A 在 Space A 搜索，结果集混有 B-only 群、A 群、跨 Space DM、未知类型。
+// fail-CLOSED 后只应保留属于 Space A 的群与 DM。
+func TestFilterResultsBySpace_FailClosed(t *testing.T) {
+	const spaceA = "spaceA"
+	gs := &fakeSpaceGroupService{bySpace: map[string]string{
+		"groupA": spaceA,
+		"groupB": "spaceB",
+	}}
+	m := newSpaceFilterTestMessage(gs)
+
+	results := []map[string]interface{}{
+		{"channel_id": "groupA", "channel_type": float64(2)}, // 同 space 群 → 保留
+		{"channel_id": "groupB", "channel_type": float64(2)}, // 越权群 → 剔除
+		dmResultWithSpace("dmA", spaceA),                     // 同 space DM → 保留
+		dmResultWithSpace("dmB", "spaceB"),                   // 越权 DM → 剔除
+		// 未知 channel_type（系统/CS/社区等本期未开类型）→ 必须剔除（核心收口）。
+		{"channel_id": "unknownChan", "channel_type": float64(99)},
+	}
+
+	filtered, err := m.filterResultsBySpace(results, spaceA)
+	assert.NoError(t, err)
+
+	ids := spaceFilterChannelIDs(filtered)
+	assert.ElementsMatch(t, []string{"groupA", "dmA"}, ids,
+		"fail-closed：仅保留同 Space 群/DM；越权频道与未知类型必须剔除")
+	assert.NotContains(t, ids, "unknownChan",
+		"未知 channel_type 必须 drop，绝不 fail-open 放行")
+	assert.NotContains(t, ids, "groupB")
+	assert.NotContains(t, ids, "dmB")
+}
+
+// 单独锁定「未知 channel_type 一律剔除」这一最高优先收口点，防止未来有人
+// 误把 default 改回 append。
+func TestFilterResultsBySpace_UnknownChannelTypeDropped(t *testing.T) {
+	m := newSpaceFilterTestMessage(&fakeSpaceGroupService{})
+	results := []map[string]interface{}{
+		{"channel_id": "x", "channel_type": float64(0)},
+		{"channel_id": "y", "channel_type": float64(3)},
+		{"channel_id": "z", "channel_type": float64(255)},
+	}
+	filtered, err := m.filterResultsBySpace(results, "anySpace")
+	assert.NoError(t, err)
+	assert.Empty(t, filtered, "所有未知 channel_type 必须被 fail-closed 剔除")
+}
--- a/modules/messages_search/api.go
+++ b/modules/messages_search/api.go
@@ -9,6 +9,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/modules/thread"
 	"github.com/Mininglamp-OSS/octo-server/modules/user"
 	spacepkg "github.com/Mininglamp-OSS/octo-server/pkg/space"
+	"github.com/Mininglamp-OSS/octo-server/pkg/searchbackend"
 	appwkhttp "github.com/Mininglamp-OSS/octo-server/pkg/wkhttp"
 )
 
@@ -32,6 +33,10 @@ type Handler struct {
 
 	limiter *uidLimiter
 	cache   *senderCache
+	// mode is the resolved OCTO_SEARCH_BACKEND posture. When mode.ESServe is
+	// false the backendGate middleware refuses every _search* request with
+	// SEARCH_DISABLED — the OS path never serves under zinc/disabled.
+	mode searchbackend.Mode
 }
 
 // New constructs the Handler. ES client setup is deferred to first request so
@@ -51,6 +56,7 @@ func New(ctx *config.Context) *Handler {
 		visibility:     newMessageVisibilityProbe(msgSvc),
 		limiter:        newUIDLimiter(cfg.RateLimit.QPS, cfg.RateLimit.Burst),
 		cache:          newSenderCache(senderCacheCapacity, senderCacheTTL),
+		mode:           searchbackend.Resolve(ctx.GetConfig().ZincSearch.SearchOn),
 	}
 	if cfg.CursorHMAC == "" {
 		// The fallback key in cursor.go is a published constant, so cursors
@@ -73,6 +79,12 @@ func New(ctx *config.Context) *Handler {
 // auth/space/uid-limit chain plus the per-user search rate limiter and the
 // audit middleware (PRM-02). Individual handlers are wired in their own
 // search_*.go files via the registerHandler helper.
+//
+// The backendGate middleware runs INSIDE the chain (after auth + rate limit +
+// audit) so a disabled / zinc deployment still meters and audits the refusal
+// — an attacker cannot enumerate channels through an unmetered "search off"
+// reply (V9). When the backend is not `es` every _search* endpoint returns
+// SEARCH_DISABLED uniformly.
 func (h *Handler) Route(r *wkhttp.WKHttp) {
 	g := r.Group("/v1/messages",
 		h.ctx.AuthMiddleware(r),
@@ -80,9 +92,25 @@ func (h *Handler) Route(r *wkhttp.WKHttp) {
 		spacepkg.SpaceMiddleware(h.ctx),
 		h.searchRateLimiter(),
 		h.auditMiddleware(),
+		h.backendGate(),
 	)
 	for _, mount := range routeMounters {
 		mount(h, g)
+	}
+}
+
+// backendGate refuses every _search* request with SEARCH_DISABLED unless the
+// resolved backend is `es`. It sits after the rate-limiter + audit middleware
+// so the refusal is still counted and logged (V9: failure paths must not be a
+// free enumeration oracle).
+func (h *Handler) backendGate() wkhttp.HandlerFunc {
+	return func(c *wkhttp.Context) {
+		if !h.mode.ESServe {
+			respondSearchDisabled(c)
+			c.Abort()
+			return
+		}
+		c.Next()
 	}
 }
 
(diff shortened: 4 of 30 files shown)


Jerry-Xin (approved):
**APPROVE** — This is a real, well-engineered security hardening of the message search/read path. I traced the live code at head `5999358`, not the commit messages, with the search/read-path visibility-parity and fail-closed lens. The PR strictly improves posture (the prior search path had no visibility post-filter at all) and the authz boundaries hold up.

### ✅ Verified

**Visibility-gate parity (search post-filter vs authoritative read path).** `MsgSyncResp.from` (`modules/message/api.go`) gates a read by setting `IsDeleted=1` on: visibles-whitelist, revoke, is_deleted (global), `messageUserExtra.message_is_deleted`, expire, and `channelOffsetMessageSeq`. `filterVisible` (`modules/messages_search/visibility.go`) replicates 5 of the 6:
- revoked → `RevokedSet`
- global delete → `GloballyDeletedSet`
- user delete → `UserDeletedSet`
- channel-offset (`MessageSeq <= offset`) → `ChannelOffset`
- `visibles` whitelist → checked directly off the OS doc (`msgRef.Visibles`)

The only read-path gate not replicated is **`expire`**. That is acceptable here because the OS `Doc` schema (`source.go`) does not carry an `expire` field and octo-server has no per-message expire write path (the read-path branch keys off a WuKongIM-sourced `msgResp.Expire`). It is documented as a known constraint (CONSTRAINTS-2026-06-12 D25). See follow-up below — not a regression introduced by this PR.

**Fail-closed.** Every authz/DB decision denies on error. `filterVisible` returns `(nil, err)` on any of the four MySQL probe failures; all callers (`searchMessages`, `searchAll`, `aroundDirection`) route a non-classified error to `respondInternal`, never to "allow." `fetchAnchor` returns `(nil, nil)` for missing/cross-Space/filtered (→ NOT_FOUND, no existence oracle) and `(nil, err)` for transport errors (→ INTERNAL). No `err != nil → allow` branch anywhere on the read path. `resp==nil` is treated as not-found only after the error has already been classified, never conflating a transport nil with "no restriction."

**Explicit backend / cross-backend authz consistency.** `pkg/searchbackend.Resolve` produces a two-axis `Mode`; all three message-search surfaces enforce it identically:
- `/v1/messages/_search*`: `backendGate()` middleware (wired into the `/v1/messages` group in `api.New`, asserted by `route_chain_test.go`) → SEARCH_DISABLED unless `ESServe`.
- `/v1/message/search` (legacy Zinc): SEARCH_DISABLED unless `LegacyZinc`.
- `/v1/search/global` message portion: only queries Zinc when `LegacyZinc`; whole endpoint SEARCH_DISABLED when `!SearchEnabled()`.

`es` never sets `LegacyZinc`, so an OS-down `es` deployment surfaces UPSTREAM_UNAVAILABLE rather than silently reading stale Zinc — and under `es` the global-search message path no longer runs, closing the "third Zinc surface bypasses ES space/visibility filtering" gap. Both backends apply the same authz; neither is a softer path.

**Legacy fail-closed.** `filterResultsBySpace` now drops unknown `channel_type` rows (with a WARN) instead of `default: append`, closing the legacy `/v1/message/search` cross-Space leak.

**IDOR / cross-tenant.** Every surface runs `checkChannelAccess` + `resolveP2PSpaceScope` before any OS round-trip; `_search_around` additionally pre-validates the anchor through channel-access + Space-scope + `filterVisible` before fetching neighbours, so a not-visible/cross-Space/revoked/command anchor returns NOT_FOUND (no neighbour disclosure, no existence oracle).

**Snowflake / float64 pagination.** Cursor stores a typed `int64` messageID; `buildSearchAfterFromHit`/`lastHitMessageID` read the tiebreaker from the typed `_source` (`doc.MessageID`), not from `hit.Sort` (which JSON-decodes to float64 and would round ids > 2^53). Internal round-refill and external cursor share one full-precision id source. Depth cap is checked as `priorDepth + page_size` (cumulative, HMAC-signed), so changing page_size cannot walk past `max_result_window=10000`.

**Tests.** Table-driven coverage for cross-Space, authz, revoke/visibles, page-size/depth caps, backend states, window assembly, response-field non-leakage, and the route chain. Build/Vet/Lint required checks pass.

### 🟡 Non-blocking follow-ups
1. **`expire` parity gap (index-level).** To reach full read-path parity the indexer would need to carry `expire` in the OS doc and `filterVisible` gate on it. Today this defends a path with no octo-server write source, but if WuKongIM-sourced expire ever becomes live for indexed messages, a search hit could outlive its read-path visibility. Track alongside the indexer `spaceId`/`visibles` work.
2. **`visibles` is fail-open for legacy docs.** Empty `Visibles` = "no gate" until the indexer writes the field. This mirrors the read path's behaviour when the field is absent, but it does mean targeted messages indexed before the indexer change are not whitelist-gated. Acceptable as a documented transitional state; close it when the indexer ships `visibles`.
3. **Unknown `OCTO_SEARCH_BACKEND` values fall back to default (ES on).** A typo like `disable`/`opensearch` preserves current behaviour rather than failing. This fails *safe* (keeps the hardened ES path with full visibility filtering on — no authz weakening), but it silently ignores operator intent. Consider failing startup loudly on an unrecognized non-empty value.
4. **`ErrMessagesSearchDisabled` is 503 without `Internal=true`** while the other 503 codes in this file set `Internal: true`. This reads as deliberate (the disabled state is client-actionable and meant to be localized so clients can hide the search box), and i18n-lint passes, so it is not an enforced-invariant violation — just worth a one-line comment noting the intentional divergence.

None of the four are authz holes, missing-and-enforceable visibility gates, or fail-open branches, so they do not block. Nice work on the anchor pre-validation and the typed-source cursor discipline.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#385 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/385
**Title:** feat(search): harden message read path, fail-closed authz, explicit backend
**Head SHA:** `59993587f1279ea3b68e6eaecf493117bd7c045b`

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Legacy fail-closed (`filterResultsBySpace`) | ✅ | `modules/message/api.go:1443-1453` — `default:` branch now logs WARN and drops; no append |
| Explicit backend switch (`pkg/searchbackend`) | ✅ | `pkg/searchbackend/backend.go:68-90` — pure `Resolve()`, 3-state + default, no auto-detect |
| Backend gate middleware | ✅ | `modules/messages_search/api.go:97-107` — after auth+rate+audit, Abort on !ESServe |
| `_search_around` anchor pre-validation | ✅ | `search_around.go:fetchAnchor` — single-doc lookup + filterVisible; nil → NOT_FOUND |
| Pagination depth cap | ✅ | `visibility.go:resolveCursorDepth` — `depth + pageSize > 10000` pre-OS check |
| Depth signed in cursor | ✅ | `cursor.go:encodeCursorWithDepth` / `decodeCursorWithDepth` — HMAC covers Depth field |
| Global search gating | ✅ | `modules/search/api.go:66-77` — `!SearchEnabled()` → SEARCH_DISABLED; message portion skipped when `!LegacyZinc` |
| Legacy search gating | ✅ | `modules/message/api.go:1326-1336` — `!LegacyZinc` → SEARCH_DISABLED |
| AppConfig `search_enabled` | ✅ | `modules/common/api.go:405,444` — both response paths set `SearchEnabled` |
| Recon drift gauges | ✅ | `recon_metrics.go` — read-only Prometheus gauges, idempotent publish |
| Test coverage | ✅ | 14 new test files covering gates V1–V10, depth cap, backend gate, join load, space filter |

---

## 2. Findings

### F1 — `splitAroundWindow` anchor-slot misalignment if `buildMessageHits` drops the anchor [P2]

**Diff-scope:** new (introduced by this PR's `_search_around` feature).

**Path:** `search_around.go:splitAroundWindow` uses `beforeN`/`afterN` counts to locate the anchor in the post-projection `hits` slice. If `buildMessageHits` drops any before-wing hits (unparseable `_source`), the anchor shifts to a lower index but `splitAroundWindow` still slices at the original `beforeN`.

**Data-flow trace:** `fetchAnchor` → `projectDocRef` succeeds (anchor is parseable) → same `*elastic.SearchHit` object appended to `window` slice → `buildMessageHits` calls `singleMessageHit` on the same object → same `_source` bytes → same parse outcome. The anchor cannot become unparseable between `fetchAnchor` and `buildMessageHits` because it is the identical Go object with identical `_source` bytes.

**Severity:** P2 — theoretically possible only if `singleMessageHit` has a non-deterministic bug; practically impossible given same-object same-bytes re-parse. The `beforeN > len(hits)` clamp at line 1281 partially mitigates but does not fully correct the split if before-hits are dropped.

**Suggestion:** Consider tagging the anchor hit (e.g., via a wrapper struct or index marker) before passing to `buildMessageHits`, so the split does not rely on positional arithmetic surviving projection drops. Low urgency.

---

### F2 — Legacy `/v1/message/search` lacks revoked/deleted/offset/visibles post-filter [P2, pre-existing]

**Diff-scope:** pre-existing. This PR does NOT claim to fix it; scope is `filterResultsBySpace` fail-closed + backend gate only.

**Path:** `modules/message/api.go:search()` — after WuKongIM returns Zinc results, only `filterResultsBySpace` is applied (when `spaceID != ""`). No `filterVisible`-equivalent check (revoked, deleted, offset, visibles). Contrast with `modules/search/api.go:global()` which applies all five signals inline, and `modules/messages_search` which uses `filterVisible`.

**R2 check:** Not new, not amplified by this PR. The PR adds a backend gate (correct) and fixes the `default: append` leak (correct). The visibility gap is a separate pre-existing issue in the legacy Zinc path.

**Severity:** P2 — pre-existing, documented gap. The new `_search*` endpoints are the correct fix path. Operators should migrate to `OCTO_SEARCH_BACKEND=es` to close this gap by routing all message search through `filterVisible`.

---

### F3 — Legacy `search()` trusts client-supplied `space_id` from body over SpaceMiddleware-validated header [P2, pre-existing]

**Diff-scope:** pre-existing. PR does not touch this logic.

**Path:** `modules/message/api.go:search()` reads `spaceID` from `req.SpaceID` (body) > `c.Query("space_id")` > `c.GetHeader("X-Space-ID")`. Body takes priority over the SpaceMiddleware-validated header. A client could send a body `space_id` differing from the authenticated Space context, potentially causing `filterResultsBySpace` to filter against the wrong Space.

**Mitigating factor:** `filterResultsBySpace` is a *restrictive* filter (drops results not matching the Space). Sending a wrong Space ID would drop MORE results, not leak more. The worst case is an empty result set, not a cross-Space leak.

**Severity:** P2 — pre-existing, and the failure mode is overly restrictive (not leaky).

---

### F4 — `global()` SQL LIKE wildcards not escaped for Space member search [P2, pre-existing]

**Diff-scope:** pre-existing. PR does not touch the SQL query.

**Path:** `modules/search/api.go:global()` uses `u.name LIKE ?` with `"%"+req.Keyword+"%"` for Space member name search. Keywords containing `%` or `_` would over-match.

**Severity:** P2 — minor over-matching risk within the Space member list only. Not a security boundary (the results are still scoped to Space members).

---

### F5 — Cursor HMAC default key is a published constant [P2, documented]

**Diff-scope:** pre-existing. PR extends the cursor payload with `Depth` but does not change the signing mechanism.

**Path:** `cursor.go:hmacKeyFn` falls back to `"octo-messages-search-default-cursor-key"` when `OCTO_SEARCH_CURSOR_HMAC` is unset. Anyone who reads the source can forge cursors including depth.

**Mitigating factor:** Depth cap is resource protection, not a security boundary. Every page still passes channel-access + filterVisible. Forging depth to 0 just resets the pagination counter; the per-page gates still apply. Startup WARN logs when secret is unset.

**Severity:** P2 — documented, acceptable per design.

---

### F6 — No per-entry validation on `SenderIDs` filter entries [P2]

**Diff-scope:** pre-existing. PR does not change `validateBase`'s SenderIDs handling.

**Path:** `validate.go:validateBase` checks `len(filters.SenderIDs) > 50` but does not validate individual UID format/length. Entries flow into `elastic.NewTermsQuery("from", terms...)`.

**Severity:** P2 — OS terms queries handle arbitrary strings safely. Low risk.

---

## 3. Suggestions

1. **F1 anchor tagging:** Consider a lightweight wrapper or marker to identify the anchor in the window slice before projection, so `splitAroundWindow` does not depend on positional arithmetic. Low urgency given same-object re-parse guarantee.

2. **F2 legacy visibility gap:** Document in the PR description or a code comment that the legacy `/v1/message/search` path intentionally does NOT add `filterVisible` (scope is backend gate + fail-closed only), and that migration to `OCTO_SEARCH_BACKEND=es` is the recommended fix path.

3. **F4 LIKE escaping:** Consider escaping `%` and `_` in the keyword before constructing the LIKE pattern. Low urgency (Space-scoped results only).

---

## 4. Additional Observations

- **`global()` computes `payload, highlights := buildMessageSearchQuery(req.Keyword)` even when `mode.LegacyZinc` is false** (the result is unused). Harmless wasted computation (~microseconds), but could be moved inside the `if mode.LegacyZinc` block for clarity.

- **`_search_around` does not carry a depth cap** — by design, since it is a fixed-window endpoint (no cursor, no multi-page walk). Each direction is bounded by `pageSize` with the oversample loop capped at `loopBudget` rounds. Correct.

- **`backendGate` placement after auth + rate-limit** is intentional and correct: metered refusal prevents unauthenticated enumeration of the "search disabled" state, and the refusal is still audit-logged.

- **Recon metrics are read-only gauges** — octo-server never computes drift itself. The authoritative recon job runs in the indexer repo. Clean separation of concerns.

---

## 5. Data-Flow Trace

|---|---|---|
| `searchbackend.Mode` | `os.Getenv("OCTO_SEARCH_BACKEND")` + `ZincSearch.SearchOn` | ✅ All 3 surfaces consult the same `Resolve()` |
| `filterResultsBySpace` input | WuKongIM `/message/search` (Zinc) | ✅ Results flow to filter; unknown channel_type now dropped |
| `filterVisible` signals (5) | MySQL: `message_extra`, `message_user_extra`, `channel_offset`; OS: `payload.visibles` | ✅ All 5 signals queried per round in oversample loop |
| `anchorHit` in `_search_around` | OS single-doc lookup by (channelId, messageId) | ✅ Same `*elastic.SearchHit` object reused in window; no re-fetch |
| `priorDepth` in cursor | `decodeCursorWithDepth` from request cursor | ✅ Signed by HMAC; carried forward as `nextDepth = priorDepth + len(collected)` |
| `SearchEnabled` in appconfig | `searchbackend.Resolve().SearchEnabled()` | ✅ Both response paths (version-match and default) set the field |
| `backendGate` refusal | `h.mode.ESServe` (resolved at `New()` time) | ✅ Middleware runs after auth+rate+audit; Abort prevents downstream handlers |

---

### C1 — Dual-path parity
**Clear.** All four paginated `_search*` endpoints (`_search`, `_search_media`, `_search_files`, `_search_all`) apply the same `resolveCursorDepth` + `paginateWithFilterDepth` chain. `_search_around` is a fixed-window endpoint (no cursor) — different shape by design, not a parity gap. The `backendGate` middleware applies uniformly to all routes mounted under the group. Legacy surfaces (`/v1/message/search`, `/v1/search/global`) both gate on the same `searchbackend.Resolve()`.

**Clear.** `resolveCursorDepth` is called exactly once per handler, before `paginateWithFilterDepth`. No nesting or double-application. The `backendGate` middleware is applied once in the group chain. The cursor HMAC uses a single key function; no composite signing paths. The `filterVisible` function is called once per oversample round; no double-filtering.

**Clear.** All `_search*` endpoints go through the same middleware chain: `AuthMiddleware` → `SharedUIDRateLimiter` → `SpaceMiddleware` → `searchRateLimiter` → `auditMiddleware` → `backendGate`. The new `_search_around` registers via `registerRoute` (verified by `TestRouteMountersCoverAllEndpoints`), so it cannot bypass any gate. No new tool/capability is exposed without authorization.

---

## Verdict

No P0 or P1 findings. All six findings are P2 (two are new but low-risk, four are pre-existing and not amplified). The PR correctly implements:
- Fail-closed security on the legacy search path (closing a real cross-Space leak)
- Explicit, shared backend switch with no auto-degradation
- Depth cap that cannot be bypassed by page_size manipulation
- Anchor pre-validation that prevents existence probing
- Uniform middleware chain covering all search endpoints

[Octo-Q] verdict: **APPROVE** — no blocking issues found. The P2 observations are advisory improvements, not landing blockers.

yujiawei (commented):
# Code Review — PR #385 (octo-server)

**Scope reviewed:** `feat(search): harden message read path, fail-closed authz, explicit backend routing`
**Head SHA:** `59993587f1279ea3b68e6eaecf493117bd7c045b` · base `main` (merge-base `43ac0941`)
**Classification:** security-sensitive (auth / search / data-isolation surface)

This is a security-sensitive change to the message search/read path. I read the full diff
(30 files, +1838/−35), traced the authz/Space-scope/backend-switch/cursor logic against the
checked-out source, ran a build + the touched unit tests, and cross-checked with two
independent model passes (one line-level, one whole-repo).

## Verdict: APPROVE

No P0/P1 blockers. The hardening is correct, well-scoped, and backed by table-driven tests
across the security-relevant axes. P2 items below are advisory and several are already
acknowledged in the code/PR body.

---

## 1. What I verified (✅ = confirmed correct)

- ✅ **Legacy fail-closed filter** (`modules/message/api.go:1431` `filterResultsBySpace`):
  the `default:` arm now drops unknown `channel_type` rows instead of appending them.
  This closes a real cross-Space exposure on the legacy `/v1/message/search` path. Locked
  by `api_search_space_filter_test.go` (`TestFilterResultsBySpace_UnknownChannelTypeDropped`).
- ✅ **Three-state backend switch** (`pkg/searchbackend/backend.go`): `es` resolves to
  `{ESServe:true, LegacyZinc:false}` — it never falls through to Zinc; `disabled` turns both
  surfaces off; unset preserves prior behaviour. Pure/tiny, no import cycles. The
  structural test `no_zinc_fallthrough_test.go` bans the legacy Zinc symbols from the ES
  module so a future edit can't silently reintroduce a fall-through.
- ✅ **All three surfaces gated consistently:** `_search*` via `backendGate()` middleware
  (`modules/messages_search/api.go:258`), legacy `/v1/message/search` via the `LegacyZinc`
  check (`modules/message/api.go:66`), and global search message-portion via
  `if mode.LegacyZinc` (`modules/search/api.go`). All return the same `SEARCH_DISABLED`
  code.
- ✅ **backendGate ordering** is correct: it runs *after* auth + rate-limit + audit
  (`api.go:240-251`), so a disabled/zinc deployment still meters and audits the refusal —
  the "search off" reply is not a free enumeration oracle.
- ✅ **Depth cap is real and wired.** `resolveCursorDepth` is called by all four paginated
  handlers (`search_messages`, `search_all`, `search_files`, `search_media`). The cap is
  checked as `priorDepth + pageSize` *before* any OS round-trip, and the depth is
  HMAC-signed into the cursor, so neither shrinking nor growing `page_size` between requests
  can walk past `maxPaginationDepth=10000`. Anti-bypass cases are explicitly tested
  (`depth_cap_test.go`).
- ✅ **`_search_around` anchor pre-validation** (`search_around.go`): the anchor passes the
  exact same gates as a normal read — `checkChannelAccess` → `resolveP2PSpaceScope` (Space
  term filter) → `filterVisible` on the anchor doc itself — and any miss collapses to a
  single `NOT_FOUND`, so the failure path is not an existence oracle. The window query
  (`buildAroundDSL`) carries the same Space scope + revoked/cmd negations, and each wing is
  bounded by the `loopBudget` oversample loop.
- ✅ **Response projection** (`singleMessageHit` / `singleSearchAllHit`) does not serialize
  `revoked` / `visibles` / `spaceId` / `is_deleted`; guarded by `vgate_matrix_test.go`
  V10 tests.
- ✅ **i18n**: both new codes (`SEARCH_DISABLED`, `DEPTH_EXCEEDED`) registered with en-US
  markers + zh-CN translations; `disabled`/`depth_exceeded` are non-Internal 4xx/503 as
  appropriate.
- ✅ **Build + tests**: `go build` of all touched packages passes; `go test
  ./pkg/searchbackend/... ./modules/messages_search/...` passes locally.

## 2. Findings (all P2 — advisory, none block merge)

**[P2] Legacy depth-less cursors reset the depth counter to 0.** `cursorPayload.Depth` is
`omitempty`, so any cursor without a signed `d` field decodes to `depth=0` and
`resolveCursorDepth` treats it as a fresh walk. Two ways to hit this: (a) a cursor issued by
the *pre-this-PR* version of `_search*` (no `d` field) presented after deploy — a client
already deep-paged past the cap can continue from its old position with the counter reset;
(b) under the default HMAC key, a forged depth-less cursor. One independent review pass rated
this **blocking ([P1])**; I downgrade it to **P2** after re-tracing the path, and I want the
disagreement on record: the depth cap is declared in-code (`visibility.go:407-419`) and in
the `DEPTH_EXCEEDED` errcode as a *resource-protection* limit, not an auth boundary. Every
page still re-runs channel-access + `filterVisible` + the bounded oversample loop, so a reset
counter leaks **no** data the caller can't already see — it only permits deeper
`search_after` paging, which is itself throttled by the per-UID rate limiter. That is why it
does not block merge. It is still a real gap the existing code comment does *not* cover (the
comment reasons only about *forged* depth, not *legitimate legacy* depth-less cursors under a
properly-set key). *Recommendation:* for the capped endpoints, treat a non-empty cursor that
carries no signed depth as either rejected or as "unknown/max" rather than 0 — e.g. version
the cursor so post-deploy code can tell "depth 0" apart from "no depth field".

**[P2] Cursor forgeable under the default HMAC key.** When `OCTO_SEARCH_CURSOR_HMAC` is
unset, cursors are signed with the published fallback constant, making both the
`search_after` position and the depth counter forgeable. This is correctly reasoned about
in `visibility.go:407-419`: forging depth buys nothing an attacker can't already get by
forging `search_after`, because every page still re-runs channel-access + `filterVisible` +
the bounded oversample loop. So the depth cap is a resource limit, not an auth boundary.
The real hardening is requiring the per-deployment secret in production. *Recommendation:*
keep the startup WARN, and consider a follow-up that hard-fails boot (or refuses search) in
production when the secret is unset, rather than only warning.

**[P2] `_search_around` anchor lookup ignores `req.Filters`.** `buildAnchorDSL`
(`search_around.go:233`) omits `addCommonFilters`, while the window query `buildAroundDSL`
(line 251) applies them. A caller can therefore anchor on a message that is visible to them
but outside the requested `sender_ids` / `sent_at` filter scope, then receive a
filter-narrowed window assembled around that out-of-scope anchor. Not a visibility leak (the
anchor still passes `checkChannelAccess` + Space-scope + `filterVisible`, and window hits are
still filtered), but the endpoint contract is inconsistent. *Recommendation:* either apply
the same common filters to the anchor lookup, or explicitly reject `filters` on
`_search_around` and document that the anchor is filter-independent.

**[P2] `visibles` whitelist is fail-OPEN for legacy docs.** `filterVisible` treats empty
`Visibles` as "no gate" (`visibility.go:217`), so until the indexer writes `doc.visibles`,
an admin-only group system message could be surfaced to a normal member by search. This is
documented (CONSTRAINTS D24) and the end-to-end gate is written-but-skipped
(`TestVGate_V3b_VisiblesWhitelistEndToEnd`). It is a known, tracked dependency, not a
regression — but it is the single thing I'd most want a human to confirm is acceptable for
the deployments that will run this before 分叉 B lands. **Human verification requested.**

**[P2] `splitAroundWindow` index alignment is correct-but-fragile.** `searchAround`
concatenates before+anchor+after, projects the whole window through `buildMessageHits`, then
re-slices using the *original* wing lengths (`search_around.go:141`). If `buildMessageHits`
dropped a `before`-wing hit, the anchor would shift into the `before` slice. In current code
this is unreachable: every window hit already passed `projectDocRef`'s `json.Unmarshal` (the
exact, stricter operation `buildMessageHits` re-runs), so no hit can drop at projection time.
The `beforeN > len(hits)` clamp only guards total shrinkage, not per-wing drift.
*Recommendation:* project the three wings independently (or carry a per-hit anchor marker)
so the invariant doesn't depend on two unmarshal sites staying in lock-step.

## 3. Coverage / blind spots (what this review did NOT or could not fully verify)

- **Indexer contract (分叉 B):** the Space-isolation and `visibles` gates are only as strong
  as the indexer writing `doc.spaceId` / `doc.visibles`. Those fields are out of this PR's
  tree; the skipped V1b/V3b end-to-end gates correctly encode that dependency but cannot be
  asserted here. The fail-closed `spaceId` term filter (matches nothing on legacy p2p docs)
  is the safe default while it's missing; the `visibles` fail-open is the unsafe-by-default
  one (see P2 above).
- **Runtime config / deploy:** `OCTO_SEARCH_BACKEND` is read from env at handler-construction
  time; switching backend requires a process restart. Not visible in the diff whether the
  deploy manifests set this. Worth confirming the rollout sets `es` explicitly rather than
  relying on the unset default.
- **`recordAudit` payload:** I confirmed refusals are audited (gate after audit middleware)
  but did not trace what PII the audit row stores for the keyword — out of this diff's scope.

*Review method: full manual diff trace + build + touched unit tests, cross-checked against
two independent model passes (one whole-repo, one line-level adversarial). One whole-repo
pass also raised a false positive — claiming the global-search message path falls through to
Zinc on `es` deployments; this is incorrect, the `if mode.LegacyZinc` guard at
`modules/search/api.go` is present and correct.*

lml2468 (approved):
✅ APPROVE — 高质量的 search 安全加固 PR,把 #361 系列教训系统化。重点维度逐项亲核(尤其新 search_around 端点)+ build/test 过,无 blocking。pr-review workflow 78 agent / 5 维 / 对抗验证,0 真负面 blocking。

## 重点核实(本地亲核,新端点/可见性是我这程的盲区,这次逐项盯死)
1. ✅ **新 search_around 端点全套 gate**(search_around.go):validateBase → checkChannelAccess → resolveP2PSpaceScope → fetchAnchor(带 loginUID/spaceID)→ anchor nil 则 respondNotFound(反枚举)。新端点最易漏 gate,这里复用了全套、顺序正确。
2. ✅ **anchor 本身过可见性**(fetchAnchor:24-29):filterVisible([]msgRef{ref}),不可见 → return nil,nil(NOT_FOUND)。堵死"借 around 锚定到不可见消息"的洞。
3. ✅ **filterVisible 全 5 信号**(visibility.go:205-228):revoked/global-deleted/user-deleted/channel-offset/**visibles 白名单**——对齐权威读路径 MsgSyncResp.from(非 filterMessages 子集)。
4. ✅ **backend gate 三态显式无 fall-through**(searchbackend.Resolve:es/zinc/disabled,默认保守 legacy):es 模式 OS down → UPSTREAM_UNAVAILABLE,绝不回退 zinc 旧结果;no_zinc_fallthrough_test 结构性禁止 messages_search 引用 zinc 面。
5. ✅ **分页雪花精度**(buildSearchAfterFromHit 读 typed Doc.MessageID,非 float64 Sort)+ depth cap(HMAC 签名累计深度,防无界 fetch)。
6. ✅ **legacy /v1/message/search 也收紧**:unknown channel_type 从 default:append 改 fail-closed(防跨 Space 泄露),且经 backend gate。
build ./modules/messages_search/ ./pkg/searchbackend/ + searchbackend test 过。

## should-fix(非阻塞,测试缺口)
search_around 的 3 条端到端负测试缺失:① anchor 命中 visibles 白名单外应 NOT_FOUND ② anchor 跨 space 应 NOT_FOUND ③ blacklisted p2p peer 应拒。**代码逻辑都对**(fetchAnchor 走 filterVisible + checkChannelAccess 已覆盖这些),缺的是回归护栏钉死——建议补,防未来改 search_around 时悄悄漏掉 anchor 的可见性/access 检查。非 blocking。

这是 #361 系列的系统化收尾:fail-closed authz、visibility 全集对齐权威读路径、分页精度、backend 无 fallthrough 全部上齐,新 search_around 端点也复用全套 gate + anchor 过可见性。无 blocking。批准,建议补 search_around 三条负测试。

—— Octo-PR (lml2468)

PR #389, opened 2026-06-16
feat(space): add paginated member search endpoint for space admins

an9xyz:
## Summary
Adds `GET /v1/space/:space_id/members/search`, a space-scoped, paginated member search for space admins/owners.

## Motivation
The user-side member list (`/v1/space/:space_id/members`) has no keyword search, and the management endpoint (`/v1/manager/spaces/:space_id/members`) is gated on a **platform** role — so a space owner/admin who is a regular platform user cannot search the members of their own space. This endpoint fills that gap, mirroring the existing `/v1/space/:space_id/app_bot` pattern.

## Endpoint
`GET /v1/space/{space_id}/members/search?keyword=&page_index=1&page_size=20`

→ `{ "count": N, "list": [ { uid, name, username, email, phone, role, robot, created_at } ] }`

## Design
- **Authz**: space-level role gate (`role >= admin`); non-member → `ErrSpaceNotMember`, regular member → `ErrSpacePermissionDenied`.
- **Lifecycle gate**: `checkSpaceActive` (strict `space.status=1`) runs first, so banned (status=2) and disbanded (status=0) spaces are rejected — a ban only updates the space row and leaves member rows at `status=1`, so the member-row gate alone is not enough.
- **Search**: keyword (trimmed) matches `name/username/email/uid` as cleartext substrings; **phone matches ONLY its last 4 digits** (`RIGHT(u.phone,4)`). Escaped `LIKE`; `_ / % / \` literal.
- **Visibility**: only `status=1` members; **all bots in the space included** (admin view — intentional divergence from the member-facing list).
- **Privacy**: **email returned in cleartext** (work emails); **phone masked to `138****5678`**. Searchable granularity == visible granularity, so masked phone digits cannot be reconstructed via substring probing. keyword is never logged. ⚠️ **Frontend note**: search phone by its **last 4 digits** — a full number will not match.
- **Envelope & paging**: `{count, list}`; `page_size` clamped to 200, `page_index` capped, deterministic `sm.uid` tie-break, `count=0` short-circuits the list query.
- **Rate limiting**: `SharedUIDRateLimiter`. Swagger 2.0 spec added.

## Test plan
- Authz: non-member / regular-member rejected.
- Lifecycle: banned (status=2) and disbanded (status=0) spaces rejected even for an active owner row.
- Empty keyword returns active members + bots; removed members hidden.
- Keyword column matching, escaping, **email cleartext + phone last-4 mask**, **phone search by last 4**.
- Bots created by others are visible (explicit admin-view decision).
- Empty contact fields stay empty; pagination clamp to 200.
- `go build` / `go vet` / unit tests pass locally; DB-backed integration tests run in CI.

## Review feedback addressed
- **(blocking)** Reject inactive spaces via `checkSpaceActive` — closes the post-ban member/PII exposure hole.
- email cleartext + phone last-4 (search & display) — **removes the masked-PII search oracle**.
- keyword trim; deterministic `uid` tie-break; swagger `page_size` default fixed (15).
- `count=0` short-circuit; `page_index` cap (avoids offset integer overflow); `maskPhone` rune-aware.
- Test: bots from other creators are visible (made the admin-view behavior explicit).

## Known non-blocking follow-ups
- Authz tests assert HTTP status only (400, D14-pinned); they do not yet assert `error.code`.
- Pre-existing (separate PR): `queryMember` returns `(nil, nil)` on a DB error, surfacing a DB failure as "not a member" rather than 5xx — affects ~6 call-sites.

### DIFF
--- a/modules/space/api.go
+++ b/modules/space/api.go
@@ -24,6 +24,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
 	octoredis "github.com/Mininglamp-OSS/octo-server/pkg/redis"
 	spacepkg "github.com/Mininglamp-OSS/octo-server/pkg/space"
+	appwkhttp "github.com/Mininglamp-OSS/octo-server/pkg/wkhttp"
 	rd "github.com/go-redis/redis"
 	"github.com/gocraft/dbr/v2"
 	"go.uber.org/zap"
@@ -94,6 +95,11 @@ func (s *Space) Route(r *wkhttp.WKHttp) {
 		auth.POST("/:space_id/email-invites", s.createMemberEmailInvite)
 		auth.GET("/:space_id/email-invites", s.listMemberEmailInvites)
 		auth.DELETE("/:space_id/email-invites/:id", s.revokeMemberEmailInvite)
+	}
+
+	search := r.Group("/v1/space", s.ctx.AuthMiddleware(r), appwkhttp.SharedUIDRateLimiter(r, s.ctx))
+	{
+		search.GET("/:space_id/members/search", s.searchMembers)
 	}
 
 	// 邀请码预览端点（公开无认证）严格 per-IP 限流：防枚举 + 暴破（issue #1000）。
--- a/modules/space/api_i18n_test.go
+++ b/modules/space/api_i18n_test.go
@@ -26,6 +26,7 @@ func httperrL(c *wkhttp.Context, code codes.Code) {
 func TestSpaceNoLegacyResponseError(t *testing.T) {
 	files := []string{
 		"api.go",
+		"api_member_search.go",
 		"api_manager.go",
 		"api_email_invite.go",
 		"api_email_invite_public.go",
--- a/modules/space/api_member_search.go
+++ b/modules/space/api_member_search.go
@@ -1,0 +1,103 @@
+package space
+
+import (
+	"strings"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
+	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
+	"go.uber.org/zap"
+)
+
+// spaceRoleAdmin is the minimum role allowed to manage space-scoped resources.
+const spaceRoleAdmin = 1
+
+// maxMemberSearchPageIndex 兜住 page_index 上限：page_size≤200 时 offset≤2e7，
+// 远超任何空间成员数，且避免 (pageIndex-1)*pageSize 在转 uint64 前整型溢出。
+const maxMemberSearchPageIndex = 100000
+
+func (s *Space) searchMembers(c *wkhttp.Context) {
+	loginUID := c.GetLoginUID()
+	spaceId := c.Param("space_id")
+
+	// 空间必须处于激活态。封禁(status=2)只更新 space 行、不会把 space_member 置为
+	// 非激活,因此仅靠成员行 status=1 的 gate 会让封禁/解散空间的 admin 仍能搜出成员
+	// (含掩码 email/phone),绕过封禁冻结。显式校验 space.status=1 统一挡掉封禁与解散。
+	if s.checkSpaceActive(c, spaceId) {
+		return
+	}
+
+	member, err := s.db.queryMember(spaceId, loginUID)
+	if err != nil {
+		s.Error("查询空间成员权限失败", zap.Error(err), zap.String("operator_uid", loginUID), zap.String("space_id", spaceId))
+		httperr.ResponseErrorL(c, errcode.ErrSpaceQueryFailed, nil, nil)
+		return
+	}
+	if member == nil {
+		httperr.ResponseErrorL(c, errcode.ErrSpaceNotMember, nil, nil)
+		return
+	}
+	if member.Role < spaceRoleAdmin {
+		httperr.ResponseErrorL(c, errcode.ErrSpacePermissionDenied, nil, nil)
+		return
+	}
+
+	pageIndex, pageSize := clampPage(c.GetPage())
+	if pageIndex > maxMemberSearchPageIndex {
+		pageIndex = maxMemberSearchPageIndex
+	}
+	// keyword 跨列匹配 name/username/email(明文)/uid；phone 仅匹配后 4 位
+	// （见 memberSearchActiveWhere）——前端按手机号检索时只需传后 4 位。
+	keyword := strings.TrimSpace(c.Query("keyword"))
+	count, err := s.db.countSearchMembers(spaceId, keyword)
+	if err != nil {
+		s.Error("统计空间成员搜索结果失败", zap.Error(err), zap.String("operator_uid", loginUID), zap.String("space_id", spaceId))
+		httperr.ResponseErrorL(c, errcode.ErrSpaceQueryFailed, nil, nil)
+		return
+	}
+	if count == 0 {
+		c.Response(map[string]interface{}{"count": int64(0), "list": []memberSearchResp{}})
+		return
+	}
+	members, err := s.db.searchMembers(spaceId, keyword, pageIndex, pageSize)
+	if err != nil {
+		s.Error("查询空间成员搜索结果失败", zap.Error(err), zap.String("operator_uid", loginUID), zap.String("space_id", spaceId))
+		httperr.ResponseErrorL(c, errcode.ErrSpaceQueryFailed, nil, nil)
+		return
+	}
+
+	resps := make([]memberSearchResp, 0, len(members))
+	for _, m := range members {
+		resps = append(resps, memberSearchResp{
+			UID:       m.UID,
+			Name:      m.Name,
+			Username:  m.Username,
+			Email:     m.Email,
+			Phone:     maskPhone(m.Phone),
+			Role:      m.Role,
+			Robot:     m.Robot,
+			CreatedAt: m.CreatedAt.String(),
+		})
+	}
+
+	s.Info("空间成员搜索", zap.String("operator_uid", loginUID), zap.String("space_id", spaceId), zap.Int64("count", count))
+	c.Response(map[string]interface{}{
+		"count": count,
+		"list":  resps,
+	})
+}
+
+// maskPhone 将手机号掩码为「前 3 + **** + 后 4」（如 138****5678）。后 4 位也是
+// memberSearchActiveWhere 唯一可检索的部分，使「可检索粒度 == 可见粒度」，避免
+// admin 通过子串查询逐位探测/重建完整号码。用 rune 切片，避免多字节输入被字节
+// 索引切成无效 UTF-8。空值保持空串。
+func maskPhone(p string) string {
+	r := []rune(p)
+	if len(r) == 0 {
+		return ""
+	}
+	if len(r) < 7 {
+		return "***"
+	}
+	return string(r[:3]) + "****" + string(r[len(r)-4:])
+}
--- a/modules/space/api_member_search_test.go
+++ b/modules/space/api_member_search_test.go
@@ -1,0 +1,383 @@
+package space
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/server"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/go-redis/redis"
+	"github.com/stretchr/testify/assert"
+)
+
+type spaceMembersSearchResponse struct {
+	Count int64                           `json:"count"`
+	List  []spaceMembersSearchItemForTest `json:"list"`
+}
+
+type spaceMembersSearchItemForTest struct {
+	UID       string `json:"uid"`
+	Name      string `json:"name"`
+	Username  string `json:"username"`
+	Email     string `json:"email"`
+	Phone     string `json:"phone"`
+	Role      int    `json:"role"`
+	Robot     int    `json:"robot"`
+	CreatedAt string `json:"created_at"`
+}
+
+func resetSpaceUIDRateLimit(t *testing.T, ctx *config.Context) {
+	t.Helper()
+	rdsClient := redis.NewClient(&redis.Options{
+		Addr:     ctx.GetConfig().DB.RedisAddr,
+		Password: ctx.GetConfig().DB.RedisPass,
+	})
+	defer rdsClient.Close()
+	keys, err := rdsClient.Keys("ratelimit:uid:*").Result()
+	if err == nil && len(keys) > 0 {
+		_ = rdsClient.Del(keys...).Err()
+	}
+}
+
+func getMembersSearch(t *testing.T, srv *server.Server, ctx *config.Context, spaceId string, q url.Values) *httptest.ResponseRecorder {
+	t.Helper()
+	resetSpaceUIDRateLimit(t, ctx)
+	path := "/v1/space/" + spaceId + "/members/search"
+	if encoded := q.Encode(); encoded != "" {
+		path += "?" + encoded
+	}
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest(http.MethodGet, path, nil)
+	req.Header.Set("token", testutil.Token)
+	srv.GetRoute().ServeHTTP(w, req)
+	return w
+}
+
+func decodeMembersSearchResp(t *testing.T, w *httptest.ResponseRecorder) spaceMembersSearchResponse {
+	t.Helper()
+	var resp spaceMembersSearchResponse
+	assert.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp), w.Body.String())
+	return resp
+}
+
+func seedMemberSearchSpace(t *testing.T, spaceId, creator string) {
+	t.Helper()
+	assert.NoError(t, testSpaceDB.insertSpaceNoTx(&SpaceModel{
+		SpaceId: spaceId,
+		Name:    spaceId,
+		Creator: creator,
+		Status:  SpaceStatusNormal,
+	}))
+	assert.NoError(t, testSpaceDB.insertMemberNoTx(&MemberModel{
+		SpaceId: spaceId,
+		UID:     creator,
+		Role:    2,
+		Status:  1,
+	}))
+}
+
+func seedMemberSearchMember(t *testing.T, spaceId, uid string, role, status int) {
+	t.Helper()
+	assert.NoError(t, testSpaceDB.insertMemberNoTx(&MemberModel{
+		SpaceId: spaceId,
+		UID:     uid,
+		Role:    role,
+		Status:  status,
+	}))
+}
+
+func seedMemberSearchUser(t *testing.T, uid, name, username, email, phone string) {
+	t.Helper()
+	_, err := testCtx.DB().InsertBySql(
+		"INSERT INTO `user` (uid, name, username, email, phone) VALUES (?, ?, ?, ?, ?) "+
+			"ON DUPLICATE KEY UPDATE name=VALUES(name), username=VALUES(username), email=VALUES(email), phone=VALUES(phone)",
+		uid, name, username, email, phone,
+	).Exec()
+	assert.NoError(t, err)
+}
+
+func seedMemberSearchRobot(t *testing.T, uid, name string) {
+	t.Helper()
+	seedMemberSearchUser(t, uid, name, uid, "", "")
+	_, err := testCtx.DB().InsertInto("robot").Columns("robot_id", "token", "status", "creator_uid").
+		Values(uid, "token-"+uid, 1, testutil.UID).Exec()
+	assert.NoError(t, err)
+}
+
+func findMemberSearchItem(list []spaceMembersSearchItemForTest, uid string) (spaceMembersSearchItemForTest, bool) {
+	for _, it := range list {
+		if it.UID == uid {
+			return it, true
+		}
+	}
+	return spaceMembersSearchItemForTest{}, false
+}
+
+func TestMaskPhone(t *testing.T) {
+	cases := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{"mainland mobile", "13812345678", "138****5678"},
+		{"seven chars boundary keeps first three and last four", "1234567", "123****4567"},
+		{"six chars fully masked", "123456", "***"},
+		{"short value fully masked", "12345", "***"},
+		{"empty value stays empty", "", ""},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.want, maskPhone(tc.in))
+		})
+	}
+}
+
+func TestSpaceMembersSearchAuthz(t *testing.T) {
+	srv, _, err := setup(t)
+	assert.NoError(t, err)
+	ctx := testCtx
+
+	t.Run("non member gets not_member", func(t *testing.T) {
+		spaceId := "sp-search-nonmember"
+		seedMemberSearchSpace(t, spaceId, "owner-nonmember")
+
+		w := getMembersSearch(t, srv, ctx, spaceId, nil)
+		assert.Equal(t, http.StatusBadRequest, w.Code, w.Body.String())
+		assertSpaceErrorCode(t, w, "err.server.space.not_member")
+	})
+
+	t.Run("regular member gets permission_denied", func(t *testing.T) {
+		assert.NoError(t, testutil.CleanAllTables(testCtx))
+		spaceId := "sp-search-member"
+		seedMemberSearchSpace(t, spaceId, "owner-member")
+		seedMemberSearchMember(t, spaceId, testutil.UID, 0, 1)
+
+		w := getMembersSearch(t, srv, ctx, spaceId, nil)
+		assert.Equal(t, http.StatusBadRequest, w.Code, w.Body.String())
+		assertSpaceErrorCode(t, w, "err.server.space.permission_denied")
+	})
+}
+
+func TestSpaceMembersSearchEmptyKeywordReturnsActiveMembersAndBots(t *testing.T) {
+	srv, _, err := setup(t)
+	assert.NoError(t, err)
+	spaceId := "sp-search-all"
+	seedMemberSearchSpace(t, spaceId, "owner-all")
+	seedMemberSearchMember(t, spaceId, testutil.UID, 1, 1)
+	seedMemberSearchUser(t, "human-all", "Human All", "human_all", "human@example.com", "13812345678")
+	seedMemberSearchMember(t, spaceId, "human-all", 0, 1)
+	seedMemberSearchRobot(t, "bot-all", "Bot All")
+	seedMemberSearchMember(t, spaceId, "bot-all", 0, 1)
+	seedMemberSearchUser(t, "removed-all", "Removed All", "removed_all", "removed@example.com", "13912345678")
+	seedMemberSearchMember(t, spaceId, "removed-all", 0, 0)
+
+	w := getMembersSearch(t, srv, testCtx, spaceId, url.Values{
+		"page_index": {"1"},
+		"page_size":  {"20"},
+	})
+	assert.Equal(t, http.StatusOK, w.Code, w.Body.String())
+	resp := decodeMembersSearchResp(t, w)
+	assert.EqualValues(t, 4, resp.Count)
+	assert.Len(t, resp.List, 4)
+
+	human, ok := findMemberSearchItem(resp.List, "human-all")
+	if assert.True(t, ok) {
+		assert.Equal(t, "human@example.com", human.Email)
+		assert.Equal(t, "138****5678", human.Phone)
+		assert.Equal(t, 0, human.Robot)
+	}
+	bot, ok := findMemberSearchItem(resp.List, "bot-all")
+	if assert.True(t, ok) {
+		assert.Equal(t, 1, bot.Robot)
+	}
+	_, ok = findMemberSearchItem(resp.List, "removed-all")
+	assert.False(t, ok, "removed members must not be visible from the space-side search")
+}
+
+func TestSpaceMembersSearchEmptyContactFieldsStayEmpty(t *testing.T) {
+	srv, _, err := setup(t)
+	assert.NoError(t, err)
+	spaceId := "sp-search-empty-contact"
+	seedMemberSearchSpace(t, spaceId, testutil.UID)
+	seedMemberSearchUser(t, "empty-contact", "Empty Contact", "empty_contact", "", "")
+	seedMemberSearchMember(t, spaceId, "empty-contact", 0, 1)
+
+	w := getMembersSearch(t, srv, testCtx, spaceId, url.Values{
+		"keyword": {"empty_contact"},
+	})
+	assert.Equal(t, http.StatusOK, w.Code, w.Body.String())
+	resp := decodeMembersSearchResp(t, w)
+	if assert.Len(t, resp.List, 1) {
+		assert.Equal(t, "", resp.List[0].Email)
+		assert.Equal(t, "", resp.List[0].Phone)
+	}
+}
+
+// TestSpaceMembersSearchIncludesBotsFromOtherCreators pins the deliberate admin-view
+// behavior: unlike the member-facing listMembers (which only surfaces bots created by
+// the caller), this admin search returns ALL bots in the space regardless of creator.
+func TestSpaceMembersSearchIncludesBotsFromOtherCreators(t *testing.T) {
+	srv, _, err := setup(t)
+	assert.NoError(t, err)
+	spaceId := "sp-search-otherbot"
+	seedMemberSearchSpace(t, spaceId, testutil.UID)
+	// a bot created by someone other than the caller
+	seedMemberSearchUser(t, "bot-other", "Bot Other", "bot_other", "", "")
+	_, err = testCtx.DB().InsertInto("robot").Columns("robot_id", "token", "status", "creator_uid").
+		Values("bot-other", "token-bot-other", 1, "someone-else").Exec()
+	assert.NoError(t, err)
+	seedMemberSearchMember(t, spaceId, "bot-other", 0, 1)
+
+	w := getMembersSearch(t, srv, testCtx, spaceId, url.Values{"keyword": {"bot_other"}})
+	assert.Equal(t, http.StatusOK, w.Code, w.Body.String())
+	resp := decodeMembersSearchResp(t, w)
+	bot, ok := findMemberSearchItem(resp.List, "bot-other")
+	if assert.True(t, ok, "admin search must surface bots created by others") {
+		assert.Equal(t, 1, bot.Robot)
+	}
+}
+
+func TestSpaceMembersSearchKeywordColumnsMaskingAndEscaping(t *testing.T) {
+	srv, _, err := setup(t)
+	assert.NoError(t, err)
+	spaceId := "sp-search-columns"
+	seedMemberSearchSpace(t, spaceId, testutil.UID)
+
+	fixtures := []struct {
+		uid      string
+		name     string
+		username string
+		email    string
+		phone    string
+	}{
+		{"u-name-target", "Alice Cooper", "alice123", "alice@example.com", "13800001111"},
+		{"u-username-target", "Bob", "zzqqxx", "bob.unique@corp.io", "13900002222"},
+		{"u-phone-target", "Carol", "carol", "carol@example.com", "15512348888"},
+		{"u-uid-target", "Dave", "dave", "dave@example.com", "15600003333"},
+	}
+	for _, f := range fixtures {
+		seedMemberSearchUser(t, f.uid, f.name, f.username, f.email, f.phone)
+		seedMemberSearchMember(t, spaceId, f.uid, 0, 1)
+	}
+
+	doSearch := func(keyword string) spaceMembersSearchResponse {
+		w := getMembersSearch(t, srv, testCtx, spaceId, url.Values{"keyword": {keyword}})
+		assert.Equal(t, http.StatusOK, w.Code, w.Body.String())
+		return decodeMembersSearchResp(t, w)
+	}
+
+	cases := []struct {
+		name    string
+		keyword string
+		wantUID string
+	}{
+		{"by name", "Alice", "u-name-target"},
+		{"by username", "zzqqxx", "u-username-target"},
+		{"by email", "bob.unique", "u-username-target"},
+		{"by phone last4", "8888", "u-phone-target"},
+		{"by uid", "u-uid-target", "u-uid-target"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			resp := doSearch(tc.keyword)
+			assert.EqualValues(t, 1, resp.Count)
+			if assert.Len(t, resp.List, 1) {
+				assert.Equal(t, tc.wantUID, resp.List[0].UID)
+			}
+		})
+	}
+
+	t.Run("email returned in cleartext, phone masked to last 4", func(t *testing.T) {
+		resp := doSearch("bob.unique")
+		if assert.Len(t, resp.List, 1) {
+			assert.Equal(t, "bob.unique@corp.io", resp.List[0].Email)
+			assert.Equal(t, "139****2222", resp.List[0].Phone)
+		}
+	})
+
+	t.Run("list and count share the same keyword filter", func(t *testing.T) {
+		resp := doSearch("example.com")
+		assert.EqualValues(t, 3, resp.Count)
+		assert.Len(t, resp.List, 3)
+	})
+
+	t.Run("like wildcard characters are escaped", func(t *testing.T) {
+		resp := doSearch("zzqqx_")
+		assert.EqualValues(t, 0, resp.Count)
+		assert.Len(t, resp.List, 0)
+	})
+}
+
+// TestSpaceMembersSearchRejectsInactiveSpace pins the lifecycle authz gate:
+// banning a space (status=2) only updates the space row and leaves space_member
+// rows at status=1, so without an explicit active-space check a former admin
+// could still search members (incl. masked email/phone) of a banned/disbanded
+// space. checkSpaceActive (strict status=1) must reject both.
+func TestSpaceMembersSearchRejectsInactiveSpace(t *testing.T) {
+	srv, _, err := setup(t)
+	assert.NoError(t, err)
+	cases := []struct {
+		name   string
+		status int
+	}{
+		{"banned", 2},
+		{"disbanded", 0},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.NoError(t, testutil.CleanAllTables(testCtx))
+			spaceId := "sp-inactive-" + tc.name
+			assert.NoError(t, testSpaceDB.insertSpaceNoTx(&SpaceModel{
+				SpaceId: spaceId,
+				Name:    spaceId,
+				Creator: testutil.UID,
+				Status:  tc.status,
+			}))
+			// caller is a still-active owner row (role=2, status=1) — the exact
+			// state a ban leaves behind; only the space-level gate can reject it.
+			assert.NoError(t, testSpaceDB.insertMemberNoTx(&MemberModel{
+				SpaceId: spaceId,
+				UID:     testutil.UID,
+				Role:    2,
+				Status:  1,
+			}))
+			w := getMembersSearch(t, srv, testCtx, spaceId, url.Values{})
+			assert.Equal(t, http.StatusBadRequest, w.Code, w.Body.String())
+		})
+	}
+}
+
+func TestSpaceMembersSearchPaginationClamp(t *testing.T) {
+	srv, _, err := setup(t)
+	assert.NoError(t, err)
+	spaceId := "sp-search-page"
+	seedMemberSearchSpace(t, spaceId, testutil.UID)
+	for i := 0; i < 204; i++ {
+		uid := fmt.Sprintf("page-member-%03d", i)
+		seedMemberSearchUser(t, uid, uid, uid, "", "")
+		seedMemberSearchMember(t, spaceId, uid, 0, 1)
+	}
+
+	w := getMembersSearch(t, srv, testCtx, spaceId, url.Values{
+		"page_index": {"1"},
+		"page_size":  {"999"},
+	})
+	assert.Equal(t, http.StatusOK, w.Code, w.Body.String())
+	resp := decodeMembersSearchResp(t, w)
+	assert.EqualValues(t, 205, resp.Count)
+	assert.Len(t, resp.List, 200, "page_size must be clamped to the shared 200-row cap")
+
+	w = getMembersSearch(t, srv, testCtx, spaceId, url.Values{
+		"page_index": {"2"},
+		"page_size":  {"2"},
+	})
+	assert.Equal(t, http.StatusOK, w.Code, w.Body.String())
+	resp = decodeMembersSearchResp(t, w)
+	assert.EqualValues(t, 205, resp.Count)
+	assert.Len(t, resp.List, 2)
+}
--- a/modules/space/db.go
+++ b/modules/space/db.go
@@ -161,6 +161,47 @@ func (d *DB) queryMembers(spaceId string, loginUID string, page uint64, limit ui
 	return models, err
 }
 
+func (d *DB) searchMembers(spaceId, keyword string, pageIndex, pageSize int) ([]*memberSearchModel, error) {
+	builder := d.session.Select(
+		"sm.*",
+		"IFNULL(u.name,'') as name",
+		"IFNULL(u.username,'') as username",
+		"IFNULL(u.email,'') as email",
+		"IFNULL(u.phone,'') as phone",
+		"CASE WHEN r.robot_id IS NOT NULL AND r.status=1 THEN 1 ELSE 0 END as robot",
+	).From(dbr.I("space_member").As("sm")).
+		LeftJoin(dbr.I("user").As("u"), "u.uid=sm.uid").
+		LeftJoin(dbr.I("robot").As("r"), "r.robot_id=sm.uid").
+		Where("sm.space_id=? AND sm.status=1", spaceId)
+	if keyword != "" {
+		clause, args := memberSearchActiveWhere(keyword)
+		builder = builder.Where(clause, args...)
+	}
+	var list []*memberSearchModel
+	_, err := builder.
+		OrderDir("sm.role", false).
+		OrderAsc("sm.created_at").
+		OrderAsc("sm.uid").
+		Limit(uint64(pageSize)).
+		Offset(uint64((pageIndex - 1) * pageSize)).
+		Load(&list)
+	return list, err
+}
+
+func (d *DB) countSearchMembers(spaceId, keyword string) (int64, error) {
+	builder := d.session.Select("COUNT(*)").
+		From(dbr.I("space_member").As("sm")).
+		LeftJoin(dbr.I("user").As("u"), "u.uid=sm.uid").
+		Where("sm.space_id=? AND sm.status=1", spaceId)
+	if keyword != "" {
+		clause, args := memberSearchActiveWhere(keyword)
+		builder = builder.Where(clause, args...)
+	}
+	var count int64
+	_, err := builder.Load(&count)
+	return count, err
+}
+
 // removeMemberLocked 锁内重读角色后移除成员，防并发转让产生无主空间，
 // 见 db_manager.go removeMemberLocked。
 func (d *DB) removeMemberLocked(spaceId, uid string, rejectRoleAtOrAbove int) error {
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -47,6 +47,28 @@ func memberSearchWhere(keyword string) (string, []interface{}) {
 	return strings.Join(clauses, " OR "), args
 }
 
+// memberSearchActiveColumns 空间侧 members/search 端点的检索列。与管理端
+// memberSearchColumns 的区别：
+//   - email 明文匹配、明文返回（工作邮箱，无需掩码）；
+//   - phone 仅匹配后 4 位（RIGHT(u.phone,4)），使「可检索粒度 == 可见粒度」
+//     （响应仅显示 138****5678），admin 无法通过子串查询逐位探测/重建完整号码。
+//
+// 前端注意：phone 检索只匹配后 4 位，传完整号码不会命中——按手机号查找请用后 4 位。
+var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
+
+// memberSearchActiveWhere 为空间侧 members/search 组装跨列 OR LIKE 条件。
+// list / count 共用同一条件，避免搜索范围漂移导致分页错位。
+func memberSearchActiveWhere(keyword string) (string, []interface{}) {
+	like := buildLikePattern(keyword)
+	clauses := make([]string, len(memberSearchActiveColumns))
+	args := make([]interface{}, len(memberSearchActiveColumns))
+	for i, col := range memberSearchActiveColumns {
+		clauses[i] = col + " LIKE ?" + likeEscapeClause
+		args[i] = like
+	}
+	return strings.Join(clauses, " OR "), args
+}
+
 // placeholders 生成 "?, ?, ?" 形式 placeholder 字符串，n 必须大于 0。
 func placeholders(n int) string {
 	return strings.TrimRight(strings.Repeat("?,", n), ",")
@@ -262,8 +284,8 @@ var ErrSpaceBannedForUpdate = errors.New("space is banned and caller disallowed 
 // 状态守卫契约（事务内强制，关闭 handler 层 guard 与 UPDATE 之间的 TOCTOU 窗口）：
 //   - SpaceStatusDisbanded 永远拒绝（ErrSpaceDisbandedForUpdate）
 //   - SpaceStatusBanned 由 allowBanned 控制：
-//       allowBanned=true（管理端）  → 放行，允许对封禁空间执行修复性更新
-//       allowBanned=false（用户端）→ 拒绝（ErrSpaceBannedForUpdate）
+//     allowBanned=true（管理端）  → 放行，允许对封禁空间执行修复性更新
+//     allowBanned=false（用户端）→ 拒绝（ErrSpaceBannedForUpdate）
 //
 // 用户端 handler 必须传 allowBanned=false：仅在入口用 checkSpaceActive 挡 banned 不够，
 // 入口检查与事务之间存在 race 窗口（manager 并发 ban），事务侧必须再挡一次才闭环。
--- a/modules/space/model.go
+++ b/modules/space/model.go
@@ -18,16 +18,16 @@ const (
 
 // SpaceModel 空间表模型
 type SpaceModel struct {
-	SpaceId        string // 空间ID
-	Name           string // 空间名称
-	Description    string // 空间描述
-	Logo           string // 空间Logo
-	Creator        string // 创建者uid
+	SpaceId        string  // 空间ID
+	Name           string  // 空间名称
+	Description    string  // 空间描述
+	Logo           string  // 空间Logo
+	Creator        string  // 创建者uid
 	MaxUsers       int     // 最大成员数 0.不限制
 	PresetGroupIds *string // 预设群组ID列表（JSON数组）
 	JoinMode       int     // 加入模式 0=直接加入 1=需要审批
 	Status         int     // 状态 1.正常 0.已解散
-	Version        int64  // 版本号
+	Version        int64   // 版本号
 	db.BaseModel
 }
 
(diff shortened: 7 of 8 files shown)


lml2468 (dismissed):
✅ APPROVE — 新增空间成员分页检索端点。这是「新端点 + admin 授权 + 搜索」三重高危叠加,我逐项盯死了授权越权、SQL/通配符注入、PII 泄露三条命门。无 blocking。

## 授权 gate(新端点最易漏,且要核「放宽后能触达什么」)——正确
gate 用 path 的 space_id + 调用者 loginUID 查「调用者在**这个** space 的成员记录」,再校验 role >= admin:
- 非本 space 成员 → not_member;本 space 非 admin → permission_denied(测试正反覆盖)
- 关键:不是「全局 admin 打任意 space_id」的越权模式 —— 查询全部 WHERE sm.space_id=? 锁死 path 指定的 space,任意 admin 都只能搜自己有 admin 角色的那个 space 的成员,搜不到别 space。授权与资源 scope 绑定正确,无跨空间 IDOR。

## SQL 注入 / LIKE 通配符注入——正确
- 值参数化:memberSearchWhere 用 `col LIKE ?`,keyword 走占位参数,无字符串拼接。
- 列名硬编码白名单:memberSearchColumns = {u.name,u.username,u.email,u.phone,sm.uid},无任何用户可控列名。
- 通配符转义:escapeLike 先转 `\` 再转 `%`/`_`(顺序正确,避免二次转义),并配 `ESCAPE '\\'` 子句 —— 同时堵住通配符注入和 sql_mode=NO_BACKSLASH_ESCAPES 下 `\` 失效两个坑。测试含 `zzqqx_` 不命中 zzqqxx 的转义用例。

## PII 泄露——正确
email/phone 在响应前掩码(maskInviteEmail → bo***@c***.io;maskPhone → 138******78),空值保持空。搜索按明文列匹配、返回时掩码,功能与隐私两全。测试断言掩码格式。

## 分页 / 数据一致性——正确
- clampPage:page_index<=0 归 1,page_size 默认 20、上限 200(测试以 page_size=999 验证钳到 200)。
- list 与 count 共用同一 memberSearchWhere,避免「列表与总数样本漂移」的分页错位(代码注释明确)。
- 已移除成员排除:searchMembers/countSearchMembers 均 WHERE sm.status=1,测试以 status=0 的成员断言不返回。
- 端点挂 per-UID rate limiter,防成员枚举/暴破。
- 机器人标识:LEFT JOIN robot + status=1 判定 robot 字段,测试覆盖。

## scope sanity
这是 space-scoped 的独立新端点(/v1/space/:space_id/members/search),与平台管理端 manager 模块的成员查询是不同授权层,不重复、不冲突。

go vet 干净(编译 + 静态检查通过)。逻辑通过读源码 + 读测试用例逐项核实;包级测试依赖 MySQL,本地无库无法运行(环境限制,CI 用真库跑,测试覆盖 authz/转义/掩码/分页钳制齐全)。

无 blocking、无 should-fix。批准。

—— Octo-PR (lml2468)

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#389 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/389
**Head SHA:** `5ace7f83c2dfe512f22cae09262bd46109181e27`
**Title:** feat(space): add paginated member search endpoint for space admins
**Files changed:** 7 (+553 / -12)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Authz: space-level role gate (≥ admin) | ✅ | `api_member_search.go:20-32` — queryMember → nil check → role < 1 check |
| Rate limiting (SharedUIDRateLimiter) | ✅ | `api.go:100` — separate route group with `appwkhttp.SharedUIDRateLimiter` |
| LIKE wildcard escaping | ✅ | `db_manager.go:16-21` (`escapeLike`) + `db_manager.go:24` (`ESCAPE '\\'`) — `_`, `%`, `\` all escaped |
| Email/phone masking | ✅ | `api_member_search.go:69-83` — empty stays empty; `maskInviteEmail` reused; phone: first 3 + last 2 |
| Pagination clamp to 200 | ✅ | `api_manager.go:37-49` — `clampPage` caps at `managerMaxPageSize=200` |
| Status=1 filter (active only) | ✅ | `db.go:175` (search) + `db.go:194` (count) — both `WHERE sm.status=1` |
| Count/list WHERE consistency | ✅ | Both use `memberSearchWhere(keyword)` — `db.go:177,196` |
| Keyword never logged | ✅ | `api_member_search.go:64` — logs only operator_uid, space_id, count |
| Test coverage | ✅ | Authz (2 cases), empty keyword, keyword columns, masking, LIKE escaping, pagination clamp, empty contacts |
| Swagger spec | ✅ | `swagger/api.yaml` — path, params, response schema all documented |
| i18n test inclusion | ✅ | `api_i18n_test.go:29` — `api_member_search.go` added to file list |

## 2. Findings

### F1 — `queryMember` swallows DB errors (P2, pre-existing)

1. **Pre-existing** — `queryMember` at `db.go:116-124` is unchanged by this PR.
2. The PR adds a new call-site (`api_member_search.go:20`) but does not alter the function's behavior contract.
3. Base branch behavior is identical; PR does not touch the relevant hunk.

**Detail:** When `d.session.Select(...).Load(&m)` returns a DB error, `m.UID` remains `""`, so `queryMember` returns `(nil, nil)` — the error is silently dropped. The handler then responds with `ErrSpaceNotMember` instead of `ErrSpaceQueryFailed`. This affects all callers (`listMembers`, `addMembers`, `removeMembers`, and now `searchMembers`).

**Severity:** P2 — pre-existing, not amplified. A DB failure would misclassify the caller as "not a member" rather than returning a server error. Low probability (DB failures are rare and transient) but could complicate incident diagnosis.

**Suggestion (non-blocking):** Consider a follow-up to fix `queryMember` to return `(nil, err)` when `err != nil`, before the `m.UID == ""` check. This would benefit all ~6 call-sites.

### F2 — `spaceRoleAdmin` constant duplicates role semantics defined elsewhere (P2, nit)

**Diff-scope:** New in this PR (`api_member_search.go:13`).

**Detail:** `const spaceRoleAdmin = 1` is file-local. The role semantics (0=member, 1=admin, 2=owner) are encoded in multiple places across the codebase without a single canonical definition. If roles are ever renumbered, this constant could drift.

**Suggestion (non-blocking):** Extract role constants to a shared location (e.g., `model.go` or a `spacepkg` package) in a follow-up.

### F3 — Keyword not trimmed of surrounding whitespace (P2, acknowledged)

**Diff-scope:** New in this PR — `api_member_search.go:35` reads `c.Query("keyword")` without `strings.TrimSpace`.

**Detail:** A keyword like `" alice "` would generate `LIKE "% alice %"`, which won't match `"Alice Cooper"`. The PR body explicitly lists this as a known non-blocking follow-up.

**Suggestion:** Add `keyword = strings.TrimSpace(keyword)` before passing to DB functions.

## 3. Suggestions

1. **Follow-up:** Fix `queryMember` error swallowing (F1) — benefits 6+ call-sites.
2. **Follow-up:** Centralize role constants (F2).
3. **Follow-up:** Trim keyword whitespace (F3).
4. **Optional:** Consider wrapping count + list in a single transaction for strict consistency under high-write workloads. Current separate-query approach is standard for search endpoints and acceptable.

## 4. Additional Observations

- **model.go whitespace-only changes:** The diff includes alignment-only changes to `SpaceModel` and `InvitationModel` struct comments. Cosmetic, no functional impact.
- **Bot visibility:** The search includes all bots in the space (no `creator_uid` filter), unlike the user-side `listMembers` which filters to bots created by the caller. This is intentional per the PR design (admin view sees all).
- **No space-status check:** The handler doesn't verify the space is active (status=1). Consistent with `listMembers` — removed members can't pass authz anyway since `queryMember` filters `status=1`.

## 5. Data-Flow Trace

|---|---|---|
| `loginUID` | `c.GetLoginUID()` — auth middleware sets from token | ✅ Auth middleware runs before handler (route group at `api.go:100`) |
| `member.Role` | `queryMember` → `space_member.role` WHERE `status=1` | ✅ DB query loads full row; nil/empty-UID guard at lines 22-28 |
| `pageIndex, pageSize` | `clampPage(c.GetPage())` → defaults 1/20, cap 200 | ✅ `GetPage()` returns `(int64, int64)`; `clampPage` signature matches |
| `keyword` | `c.Query("keyword")` — raw query param | ✅ Flows to `memberSearchWhere` → `buildLikePattern` → `escapeLike` (SQL injection safe via parameterized query + escaping) |
| `count` | `countSearchMembers` → `SELECT COUNT(*)` with same WHERE | ✅ Shares `memberSearchWhere` with list query; status=1 filter consistent |
| `m.Name/Username/Email/Phone` | `searchMembers` SQL: `IFNULL(u.*,'')` via LEFT JOIN user | ✅ IFNULL ensures empty string (not NULL) for missing user rows |
| `m.Robot` | `CASE WHEN r.robot_id IS NOT NULL AND r.status=1 THEN 1 ELSE 0 END` | ✅ LEFT JOIN robot; status=1 check ensures deactivated robots show as 0 |
| `m.CreatedAt.String()` | `db.Time.String()` → `"2006-01-02 15:04:05"` format | ✅ Consistent with `listMembers` response format |
| `maskInviteEmail(email)` | `email_invite_token.go:41` — keeps first 2 local chars + domain initial | ✅ Empty email returns empty (guarded at `api_member_search.go:70-72`) |
| `maskPhone(phone)` | `api_member_search.go:77-83` — first 3 + "******" + last 2 | ✅ Short phones (<6) return "***"; empty stays empty |

## 6. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. This is a new endpoint with no symmetric add/remove or create/delete pair. The count and list queries share `memberSearchWhere` — verified consistent.
- **C2 — Control-flow ordering / nested reuse:** Clear. `memberSearchWhere` is reused from the manager-side search; `clampPage` is reused from the manager API; `maskInviteEmail` is reused from email invites. All are leaf functions with no nested composition risk. LIKE escaping uses `escapeLike` which handles `\`, `%`, `_` in correct order (backslash first).
- **C3 — Authorization boundary ≠ capability boundary:** Clear. The endpoint is gated by auth middleware (token required) + space-level role check (≥ admin). No secondary call-sites or synthetic triggers bypass this gate. Rate limiter is applied at the route group level.

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. The PR is well-structured with proper authz, rate limiting, SQL injection prevention (parameterized queries + LIKE escaping), PII masking, and comprehensive test coverage. Three P2 items (all acknowledged or pre-existing) are noted as non-blocking follow-ups.

lml2468:
补充 / 修正我的 review:mochashanyao 提的第一条我上一轮漏了,在此诚实补正并认领为 should-fix。

## should-fix(我上轮漏判,非 blocking,但确是 bug)
**queryMember 吞掉 DB 错误**(modules/space/db.go:116):
```
_, err := d.session...Load(&m)
if m.UID == "" {
    return nil, nil    // DB 出错时 m.UID 也是 "",err 被丢弃
}
return &m, err
```
DB 查询出错时 m.UID 为空 → 返回 (nil, nil),err 被吞。结果 searchMembers 里那个 `if err != nil { ... ErrSpaceQueryFailed }` 分支**永远走不到**,DB 抖动会被当成"非成员"返回 ErrSpaceNotMember。

两个影响:
1. 正确性/可观测性:真实 DB 故障被掩盖成"你不是成员",错误码语义错、日志少一条 query-failed,排障困难。
2. 方向上**是 fail-closed 不是越权**:DB 出错时是"误拒合法 admin",不是"误放行非 admin",所以**不是安全 blocker**——这也是我维持 APPROVE 的原因。但它仍是该修的 bug。

建议:queryMember 在 `Load` 后先判 `if err != nil { return nil, err }`,再判 `m.UID == ""`。这样 handler 的 ErrSpaceQueryFailed 分支才有意义。

## 另两条(nit,同意)
- spaceRoleAdmin = 1 与包内已有 EmailInviteRoleAdmin = 1 语义重复,建议统一角色常量来源。
- keyword 未 TrimSpace(对比其他搜索端点的处理),首尾空格会进 LIKE 模式影响匹配,建议 trim。

授权 gate(space-scoped admin 校验)、SQL/通配符转义、PII 掩码、分页一致性这些主判断不变,仍无 blocking。以上 should-fix 修了更稳。感谢 mochashanyao 抓到 queryMember 这条——我上轮确实漏了错误处理这一层。

—— Octo-PR (lml2468)

yujiawei (dismissed):
# Code Review — PR #389 (octo-server)

Reviewed at head SHA `5ace7f83c2dfe512f22cae09262bd46109181e27` against `origin/main` (merge-base `c60bbd84`). `go build` and `go vet` on `modules/space/...` are clean. DB-backed integration tests were not executed (no MySQL/Redis/WuKongIM infra in this environment); correctness was verified statically against the existing endpoint patterns plus the included unit tests.

## Summary

A well-scoped, well-tested addition. The new `GET /v1/space/:space_id/members/search` closely mirrors the existing manager (`queryMembersAdmin`) and `listMembers` patterns, and correctly reuses the shared building blocks (`memberSearchWhere`, `clampPage`, `maskInviteEmail`, `SharedUIDRateLimiter`). The core security-sensitive aspects are sound:

- **Authz** — space-level role gate is correct: `role 0=member / 1=admin / 2=owner`; non-member → `not_member`, `role < 1` → `permission_denied`. Matches the `app_bot` precedent.
- **SQL injection** — keyword goes through parameterized `dbr` placeholders with `LIKE ... ESCAPE '\\'`; `_ / % / \` are escaped literally by `escapeLike` (backslash first, correct order).
- **count vs list consistency** — both queries share `memberSearchWhere` and `sm.status=1`. The list query's extra `LEFT JOIN robot` is 1:1 (`robot_id` is UNIQUE) and never filters/multiplies rows, so totals stay consistent across pages.
- **PII masking** — `email`/`phone` masked; empty contacts preserved; keyword is never logged.
- **Index support** — `WHERE sm.space_id=? AND sm.status=1` is backed by `spacemember_spaceid_status`.

No correctness, security, data-loss, or build-breaking issues. The items below are non-blocking suggestions.

## Findings (all P2 / non-blocking)

### P2 — Member search is reachable on disbanded spaces
`modules/space/api_member_search.go:14-28`, `modules/space/db.go:164-205`

`disbandSpace` (`db.go:62`) only sets `space.status=0`; it leaves `space_member.status=1`. This handler authorizes purely from `space_member` and the queries never require an active `space` row, so a former admin/owner can still search members (and masked email/phone) of a disbanded space. This mirrors the existing `listMembers`/`queryMembers` behavior, so it is not a regression — but this endpoint widens the exposed surface (adds username + masked email/phone). Consider adding a `space.status=1` guard (or joining on it) for defense-in-depth. Pre-existing pattern, masked data, admin-only → not a merge blocker.

### P2 — Documented default `page_size` (20) does not match runtime (15)
`modules/space/swagger/api.yaml` (`page_size … 默认 20`) vs `modules/space/api_member_search.go:30`

`c.GetPage()` returns `page_size=15` when the param is omitted, and `clampPage` only overrides values `<= 0`. So an absent `page_size` yields 15, not the documented 20. (The manager endpoint's swagger has the same drift.) Either document the real default (15) or set the intended default before clamping.

### P2 — Pagination has no unique tie-breaker
`modules/space/db.go:182-184`

`ORDER BY sm.role DESC, sm.created_at ASC` without a unique final key can reorder rows that share role+timestamp (common after bulk import), causing skips/dupes across pages. This is the same ordering used by `queryMembersAdmin`/`queryMembers`, so it's an established convention — but appending `.OrderAsc("sm.uid")` would make paging deterministic here and is cheap.

### P2 (nit) — `maskPhone` byte-slices the string
`modules/space/api_member_search.go:73-82`

`p[:3]` / `p[len(p)-2:]` are byte indexes; a non-ASCII phone value would be split mid-rune and emit invalid UTF-8. Phone data is realistically ASCII digits so this is theoretical; `[]rune` slicing would harden it.

### P3 (nit) — Disabled bots render as humans; redundant query on empty count
`modules/space/db.go:169`, `modules/space/api_member_search.go:34-44`

`CASE WHEN r.robot_id IS NOT NULL AND r.status=1 THEN 1 ELSE 0` classifies a soft-disabled bot (`status=0`) as a human (`robot=0`). This is intentional parity with `queryMembers`; flag only if product wants identity preserved regardless of bot status. Separately, when `countSearchMembers` returns 0 the handler still runs the list query — a tiny early-return would save one round-trip.

## Coverage / blind spots

- **Tests**: comprehensive for authz, column matching, masking, LIKE-wildcard escaping, empty contacts, bots-included, and the 200-row clamp. As the PR notes, authz tests assert HTTP 400 only (not `error.code`) and `keyword` isn't trimmed — both acceptable follow-ups.
- **Not executed here**: DB-backed integration paths (the count/list parity, the join behavior under real rows) — verified by reading, not running.
- **gin routing**: no method-level route conflict (`GET …/members/search` coexists with `GET …/members` and the `PUT …/members/:uid/role` param route lives in a different method tree).
- **Migrations**: none required; reuses existing tables/indexes.

## Verdict

No blocking issues. Approving; the P2/P3 items above are optional hardening/cleanup.

lml2468:
核了 yujiawei 的 4 条,逐条给出验证结论(亲核源码):

**🟡#1 解散空间状态校验 —— 合理的纵深防御 should-fix,但当前非可利用。**
亲核 forceDisbandSpace(db_manager.go):解散时在同一事务里把 space.status=0 **同时**把该空间所有 space_member 置 status=0。而本端点 gate 用的 queryMember 查的是 status=1 的成员行 —— 所以解散后任何人(含原 admin)都查不到 status=1 的成员行,gate 直接返回 not_member,搜不了。**当前 forceDisband 路径下不是可利用漏洞**。
但 yujiawei 的点仍成立为 should-fix:gate 现在依赖"解散一定同步把成员行置 0"这个隐式不变量,而非显式校验 space.status。包内其他空间端点(querySpaceDetail 等)都显式 gate s.status=1。建议本端点也显式加 space.status=1 校验做纵深防御,避免未来新增解散路径/并发窗口(space 已置 0、成员行未及置 0)时漏掉。非 blocker。

**🟡#2 文档默认分页 20 vs 代码 15 —— 前提不成立。**
亲核:clampPage 默认 pageSize=20(api_manager.go),swagger /members/search 的 page_size 描述写的也是"默认 20,上限 200"(api.yaml:44)。代码和文档都是 20,不存在 20 vs 15 不符。这条可不处理。

**🟡#3 分页排序缺唯一字段 —— 准,should-fix。**
排序是 OrderDir(sm.role) + OrderAsc(sm.created_at),无唯一 tiebreaker。role+created_at 相同(如同秒批量导入的成员)时,MySQL 行序不稳定,翻页可能重复或遗漏。建议追加 sm.uid 作末位唯一排序字段。非 blocker(单页内正确,仅跨页边界受影响)。

**🟡#4 phone 掩码字节切片 —— 准,nit。**
maskPhone 用 p[:3] / p[len(p)-2:] 字节切片,非 ASCII(国际号带非 ASCII 字符等)会切坏 UTF-8。phone 多为 ASCII 数字,实际影响小;改 []rune 切片更稳妥。

合并前建议作者至少处理 🟡#1(纵深防御)和 🟡#3(分页稳定),连同我上一条补认的 queryMember 吞 DB error,一并修了更稳。授权 scope、SQL/通配符转义、PII 掩码这些主判断不变,仍无 blocker。

—— Octo-PR (lml2468)

Jerry-Xin (dismissed):
One blocking lifecycle authorization issue: the endpoint is in-scope and generally well aligned, but it can still expose searchable member/contact data after a space is no longer active.

**🔴 Blocking**
- 🔴 Critical: `searchMembers` authorizes only through an active `space_member` row and never verifies that the space itself is still active. Since user-side `disbandSpace` marks `space.status=0` but does not remove member rows, a former admin/owner can keep calling `GET /v1/space/:space_id/members/search` for a disbanded space and search masked email/phone data. Other admin/owner space operations call `checkSpaceActive` before proceeding. Add the same active-space guard here, or make the auth/search queries join `space s ON s.space_id=sm.space_id AND s.status=1`. See [api_member_search.go](/Users/mlamp/.openclaw/tmp/pr-review-qw6qa_pg/repo/modules/space/api_member_search.go:17), [db.go](/Users/mlamp/.openclaw/tmp/pr-review-qw6qa_pg/repo/modules/space/db.go:190), [api.go](/Users/mlamp/.openclaw/tmp/pr-review-qw6qa_pg/repo/modules/space/api.go:555), and the active guard pattern in [api_email_invite.go](/Users/mlamp/.openclaw/tmp/pr-review-qw6qa_pg/repo/modules/space/api_email_invite.go:31).

**💬 Non-blocking**
- 🟡 Warning: The Swagger says omitted `page_size` defaults to 20, but `c.GetPage()` already normalizes missing `page_size` to 15 before `clampPage` sees it, so this endpoint will default to 15 unless the client sends `page_size<=0`. Either document 15 or parse raw query values before clamping. See [api_member_search.go](/Users/mlamp/.openclaw/tmp/pr-review-qw6qa_pg/repo/modules/space/api_member_search.go:32), [api_manager.go](/Users/mlamp/.openclaw/tmp/pr-review-qw6qa_pg/repo/modules/space/api_manager.go:37), and [api.yaml](/Users/mlamp/.openclaw/tmp/pr-review-qw6qa_pg/repo/modules/space/swagger/api.yaml:40).

**✅ Highlights**
- Uses localized `httperr.ResponseErrorL` paths and adds the new file to the legacy-response guard.
- Reuses the escaped `LIKE` helper, avoiding wildcard overmatching and SQL injection.
- Mounts the endpoint after auth with `SharedUIDRateLimiter`.
- Tests cover authz, masking, wildcard escaping, bots, removed-member hiding, and pagination clamp.

Verification note: `go test ./modules/space/...` could not run in this checkout because the local test DB migration state fails before tests with `unknown migration in database` for `20191106000001_event_legacy01.sql`.

lml2468 (changes requested):
撤回我之前的 APPROVE,改投 CHANGES_REQUESTED。Jerry-Xin 升 blocking 是对的,我之前判错了 —— 这是真实可利用的数据泄露,我来认领并说清根因。

## 🔴 blocking:封禁空间(status=2)的 admin 仍可搜成员 = PII 泄露
我之前核 yujiawei 的"解散空间校验"时,**只核了解散(disband)路径就下了"当前非可利用"的结论,漏了封禁(ban)路径**。space.status 是三态:0=已解散、1=正常、2=已封禁。

- 解散路径 forceDisbandSpace:同事务把 space.status=0 **且** space_member 全置 status=0 → 本端点 gate(queryMember 查 status=1)自然挡掉。这条我核对了,确实兜住。
- **封禁路径 liftBan(status=2) → updateSpaceStatus 只 UPDATE space SET status=2,完全不动 space_member 行**(db_manager.go:223)。所以封禁空间的成员行仍是 status=1。

后果:空间被超管封禁后,该空间的 admin 成员调用 GET /v1/space/:space_id/members/search,queryMember 仍返回 status=1 的成员行、Role>=admin 通过 gate → 能搜出该封禁空间全部在册成员(含掩码 email/phone)。封禁本意是冻结空间访问,这个新端点绕过了它。**可利用、是 blocker。**

## 修复
searchMembers 的 gate 应显式校验空间处于激活态(space.status=1),不能只依赖成员行状态 —— 这也是包内其他空间端点(querySpaceDetail 等 WHERE s.status=1)的既有模式。加上后:封禁(2)和解散(0)空间都会被 gate 统一挡掉,不再依赖"某条状态变更路径是否记得同步置成员行"这个易漏的隐式不变量。

## 连带(上面几条 review 已提,合并前一并修)
- queryMember(db.go:116)吞 DB error 返回 (nil,nil),应先 if err!=nil return nil,err。
- 分页排序加 sm.uid 唯一 tiebreaker,防跨页重/漏。
- (yujiawei #2 文档 page_size:经核代码与 swagger 都是默认 20,不存在不符,可不处理。)

## 我的错误复盘
我对"空间非激活"对齐成了"解散"这一个子集,漏了"封禁"这个分支,所以把可利用泄露误判成了"当前非可利用的纵深防御 should-fix"。授权 scope(同空间)、SQL/通配符转义、PII 掩码这些我核对了没问题,但状态校验这一层只查了一半。感谢 Jerry-Xin 升级定性。

—— Octo-PR (lml2468)

lml2468 (changes requested):
基于最新 commit 2f837277 重审:维持 CHANGES_REQUESTED —— **核心 blocking 没修**。

## 这次提交改了什么(增量 diff 逐行核)
只有两处,都是之前提的 should-fix:
- api_member_search.go: keyword 加了 strings.TrimSpace ✅
- db.go: 分页排序加了 OrderAsc("sm.uid") 唯一 tiebreaker ✅

这两条改对了,感谢。

## 🔴 但 blocking 完全没动:封禁空间(status=2)的 admin 仍能搜成员
新 head 的 searchMembers gate 仍然只有:queryMember(查 space_member status=1) → member==nil 判定 → Role<admin 判定。**没有加任何 space.status 校验。**

复述漏洞链(仍成立):
- liftBan(status=2) 封禁空间时只 UPDATE space SET status=2,不动 space_member 行。
- 所以封禁空间的成员行仍是 status=1,queryMember 照常返回,Role>=admin 照常通过 gate。
- 结果:空间被封禁后,其 admin 仍能调本端点搜出全部在册成员(含掩码 email/phone)。封禁本应冻结空间访问,本端点绕过了它。

这是可利用的数据泄露,必须修才能合。建议:gate 在成员校验之外,显式校验空间处于激活态(space.status=1)——可用包内已有的 querySpaceByID(它就是 WHERE status=1)。这样封禁(2)和解散(0)空间都被统一挡掉。

## 另一条 should-fix 也仍未改
queryMember(db.go:116)仍是 `Load(&m); if m.UID==""{ return nil,nil }`,DB error 被吞 → handler 的 ErrSpaceQueryFailed 分支永远走不到。建议 Load 后先 if err!=nil return nil,err。(非 blocker,但既然在改这块,一起修。)

授权 scope、SQL/通配符转义、PII 掩码、分页(这次又补了唯一排序)这些没问题。**就差封禁空间状态校验这个 blocker** —— 修完我就转 APPROVE。

—— Octo-PR (lml2468)

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#389 Review Report
**Head SHA:** `2f837277dc140138138ca361193fabc92e53531b`
**PR:** feat(space): add paginated member search endpoint for space admins

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Authz gate (space-level role ≥ admin) | ✅ | `api_member_search.go:20-33` — `queryMember` → nil check → `Role < spaceRoleAdmin` check |
| SQL injection prevention (LIKE escaping) | ✅ | `db_manager.go:17-28` — `escapeLike` escapes `\`, `%`, `_`; `likeEscapeClause` adds `ESCAPE '\\'` |
| Rate limiting | ✅ | `api.go:100` — mounted behind `SharedUIDRateLimiter` |
| Keyword not logged | ✅ | `api_member_search.go:68` — log line has `operator_uid`, `space_id`, `count` only |
| Email/phone masking | ✅ | `api_member_search.go:74-85` — `maskInviteEmail` + `maskPhone`; empty → empty |
| Pagination clamp | ✅ | `api_manager.go:37-49` — `clampPage` caps at `managerMaxPageSize=200` |
| Count/list consistency | ✅ | Both use `memberSearchWhere` for keyword filter (`db.go:164-205`) |
| Error response i18n | ✅ | `api_i18n_test.go:29` — `api_member_search.go` added to lint scan list |
| Test coverage | ✅ | `api_member_search_test.go` — authz, empty keyword, masking, escaping, pagination clamp, empty contacts |
| Swagger spec | ✅ | `swagger/api.yaml:17-62` — path, params, response schema documented |

## 2. Findings

### F1 — Bot visibility divergence from `listMembers` (P2, new, by design)

**Diff-scope:** new behavior introduced by this PR.

`searchMembers` (`db.go:164-190`) returns ALL bots in the space (no creator filter), while the existing `listMembers` → `queryMembers` (`db.go:140-157`) filters: `r.robot_id IS NULL OR r.creator_uid = ?`.

This is a deliberate design choice for an admin-level endpoint (admins should see all members). Not a bug, but worth documenting in the PR body or a code comment to prevent future confusion when someone compares the two endpoints.

**Severity:** P2 — design decision, not a defect.

### F2 — Missing `checkSpaceActive` guard (P2, pre-existing pattern)

**Diff-scope:** pre-existing — `listMembers` (`api.go:620-660`) also omits `checkSpaceActive`.

The handler doesn't verify the space is active (status=1) before searching. After `disbandSpace`, `space_member` rows retain status=1, so a disbanded space's members remain searchable. This is consistent with the existing `listMembers` pattern. Write endpoints (`addMembers`, `removeMembers`, etc.) all call `checkSpaceActive`; read-only member queries don't.

**Severity:** P2 — consistent with existing read-only endpoint pattern.

### F3 — `maskPhone` byte-level slicing (P2, new, low risk)

**Diff-scope:** new code in `api_member_search.go:82-85`.

`p[:3]` and `p[len(p)-2:]` are byte-level slices. For phone numbers (digits, optional `+` prefix), this is safe. If a phone number ever contained multi-byte UTF-8 characters, this could split a rune and produce invalid output. In practice phone numbers are ASCII digits, so this is a nit.

**Severity:** P2/nit — safe for all realistic phone number formats.

### F4 — PR description vs code discrepancy on keyword trimming (P2, doc)

**Diff-scope:** new code.

PR body lists "keyword is not trimmed of surrounding whitespace" as a known follow-up, but the code does `strings.TrimSpace(c.Query("keyword"))` at `api_member_search.go:35`. The code behavior is correct; the PR description is stale.

**Severity:** P2/nit — update PR description to match code.

## 3. Suggestions

1. **F1:** Add a brief code comment on `searchMembers` explaining the intentional bot-visibility difference from `listMembers`, so future maintainers don't "fix" it as a regression.
2. **F4:** Update the PR body's "Known non-blocking follow-ups" section to remove the whitespace-trimming item (it's already done).

## 4. Additional Observations

- The `spaceRoleAdmin` constant (`api_member_search.go:13`) is only used in this file. Other handlers use literal `1` (e.g., `member.Role < 1` in `addMembers`). Not wrong — a named constant is arguably better — but inconsistent with the rest of the module. Consider adopting it module-wide or using the literal for consistency.
- Count and list queries are not wrapped in a transaction, so a concurrent member add/remove between the two queries could produce a count/list mismatch. This is the same pattern used by the manager endpoint and is acceptable for a paginated search endpoint.

## 5. Data-Flow Trace

|---|---|---|
| `loginUID` | `c.GetLoginUID()` (auth middleware sets from token) | ✅ |
| `spaceId` | `c.Param("space_id")` (URL path param) | ✅ |
| `member.Role` | `queryMember` → `space_member.role` WHERE `status=1` | ✅ — nil check before role check |
| `pageIndex, pageSize` | `clampPage(c.GetPage())` → defaults 1/20, cap 200 | ✅ |
| `keyword` | `c.Query("keyword")` → `TrimSpace` → `memberSearchWhere` (parameterized LIKE) | ✅ — escaped, never logged |
| `count` | `countSearchMembers` → `COUNT(*)` with same WHERE as list | ✅ |
| `members[].Name/Username/Email/Phone` | `searchMembers` → `LEFT JOIN user` → `IFNULL` → masking | ✅ — empty stays empty |
| `members[].Robot` | `searchMembers` → `LEFT JOIN robot` → `CASE WHEN` | ✅ |
| `members[].CreatedAt` | `searchMembers` → `sm.*` → `db.BaseModel.CreatedAt` → `.String()` | ✅ — format `2006-01-02 15:04:05` |

## 6. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. This PR adds a read-only search endpoint; no corresponding write/remove path to check for symmetry.
- **C2 — Control-flow ordering / nesting reuse:** Clear. `memberSearchWhere` is reused from the manager endpoint and called in two places (count + list) with the same keyword — no nesting or double-application risk. `maskPhone` and `maskInviteEmail` are leaf functions with no re-entrancy.
- **C3 — Authorization boundary ≠ capability boundary:** Clear. Endpoint is behind `AuthMiddleware` + `SharedUIDRateLimiter`. Handler-level authz checks `member.Role >= spaceRoleAdmin`. No new tool/capability exposure — read-only member data with PII masking.

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. All issues are P2/nit (design documentation, pre-existing pattern consistency, doc discrepancy). The endpoint has proper authz, SQL injection prevention, rate limiting, PII masking, and comprehensive test coverage.

lml2468 (dismissed):
✅ 撤回 CHANGES_REQUESTED,改投 APPROVE。基于最新 commit 379134ab 重审:核心 blocker(封禁/解散空间成员泄露)已彻底修复并亲核,有回归测试。

## 🔴 blocker 已修对(逐项亲核,非看 commit message)
新端点 searchMembers 入口现在第一步就是 `if s.checkSpaceActive(c, spaceId) { return }`,在任何成员查询之前:
- checkSpaceActive → isSpaceActive 严格 `SELECT COUNT(*) FROM space WHERE space_id=? AND status=1`,只有 status=1 算激活。封禁(status=2)和解散(status=0)**都判非激活**被挡 —— 这正堵住了之前的洞:封禁只改 space 行、不动 space_member,光靠成员行 status=1 的 gate 挡不住。
- fail-closed:isSpaceActive 出错 → checkSpaceActive 返回 true(中止),不会误放行。
- 顺序正确:空间激活校验在 queryMember 之前,封禁空间在任何成员数据访问前就被拒。
- 回归测试 TestSpaceMembersSearchRejectsInactiveSpace 钉死:封禁(status=2)留 member status=1 时端点仍须拒,断言 checkSpaceActive 严格 status=1 同时覆盖封禁与解散。

## 这次提交的其他改动(也核了)
- keyword 加 strings.TrimSpace ✅
- 分页排序加 sm.uid 唯一 tiebreaker(防跨页重/漏)✅
- vet 干净

## 仍未改的 should-fix(非 blocking,可作 follow-up,不阻断合并)
queryMember(db.go:120)仍 `if m.UID==""{ return nil,nil }` 吞 DB error。方向是 fail-closed(DB 错→误判非成员→拒,不是放行),所以非越权 blocker;但仍建议后续修成 Load 后先 if err!=nil return nil,err,让 handler 的 ErrSpaceQueryFailed 分支可达、便于排障。不阻断本次合并。

授权 scope(同空间绑定无跨空间 IDOR)、空间生命周期校验(本次补齐)、SQL/通配符转义、PII 掩码、分页一致性 + 唯一排序、rate limiter —— 主判断全部干净。核心 blocker 已修。批准。

感谢作者补齐 checkSpaceActive,也感谢 Jerry-Xin 把这条从 🟡 升 🔴 —— 封禁场景下确为可利用泄露,我初判漏了封禁分支,这次修复把封禁+解散统一挡住,是正解。

—— Octo-PR (lml2468)

Jerry-Xin (dismissed):
✅ APPROVE — the blocking lifecycle authz gap is fixed correctly. Re-reviewed at head `379134aba387`.

**🔴 → resolved: space-active gate added.** `searchMembers` now opens with `checkSpaceActive(c, spaceId)` (api_member_search.go:22) BEFORE the member/role check. This closes the disbanded-space hole: the normal owner-disband endpoint (`DELETE /:space_id` → `disbandSpace`) sets `space.status=0` but leaves `space_member` rows at `status=1`, so the member-row gate alone let a former owner/admin keep enumerating masked member contact data. The new check also correctly covers the banned (`status=2`) case in the same gate — banning likewise only updates the `space` row, not member rows. More complete than the minimum fix; the inline comment documents the reasoning.

**Should-fixes also addressed:**
- keyword now `strings.TrimSpace`-d (api_member_search.go:42).
- pagination now has a unique tiebreaker: `ORDER BY sm.role DESC, sm.created_at ASC, sm.uid ASC` — the `sm.uid` tiebreaker makes paging stable when role+created_at collide (no dupe/skip).

**Remaining non-blocking notes (acceptable to defer):**
- `queryMember` returns `(nil, nil)` when the row scan finds nothing even on a DB error, so the handler's `ErrSpaceQueryFailed` branch is unreachable and a DB blip surfaces as "not member." This is fail-CLOSED (denies, never over-grants), so it's an error-code/observability nit, not a security issue.
- `maskPhone` uses byte-slicing (`p[:3]`/`p[len(p)-2:]`); phone values are effectively ASCII so this is safe in practice — a nit.

**Verified unchanged-good:** admin authz is space-scoped (same `space_id` used for the role check and the queries — no cross-space IDOR); result queries filter `sm.space_id=? AND sm.status=1` (removed members excluded, no cross-space leak); `clampPage` caps page size; search uses parameterized `LIKE ?` + `escapeLike` (`\` first) + `ESCAPE '\\'` (no injection); email/phone masked in responses.

CI: Build / Vet / Lint / Test / i18n / code-review all green. No blockers.

yujiawei (changes requested):
# Code Review — PR #389 (octo-server)

Reviewed at head SHA `2f837277dc140138138ca361193fabc92e53531b` against `origin/main` (merge-base `c60bbd84`). `go build` and `go vet ./modules/space/...` are clean. DB-backed integration tests were not executed (no MySQL/Redis/WuKongIM infra here); correctness was verified by reading the source plus the included unit tests and tracing the existing endpoint patterns.

This re-review is against the **current** head. Two earlier findings tracked at SHA `5ace7f8` are now resolved in this commit and are **not** re-raised: the `queryMember` error swallowing is fixed (`api_member_search.go:18-23` now returns `ErrSpaceQueryFailed` on `err != nil`), keyword is trimmed (`:34`), a `sm.uid` tiebreaker was added (`db.go:184`), and the swagger `page_size` default now correctly documents `15` (`swagger/api.yaml:36`), matching the runtime `GetPage()` default — so the documented-vs-runtime drift no longer applies.

## Summary

A well-scoped, well-tested addition. `GET /v1/space/:space_id/members/search` mirrors the existing manager and `listMembers` patterns and correctly reuses the shared building blocks (`memberSearchWhere`, `clampPage`, `maskInviteEmail`, `SharedUIDRateLimiter`). SQL injection is properly prevented (parameterized `LIKE ... ESCAPE '\\'`, hard-coded column whitelist), the cross-space authorization scope is correct (the gate is bound to the path `space_id`, no global-admin IDOR), and PII is masked in responses with keyword never logged.

However, there is **one blocking authorization-lifecycle gap** that exposes member PII on spaces that are no longer active. The remaining items are non-blocking.

## 🔴 Blocking (P1) — Member search is reachable on banned and disbanded spaces

`modules/space/api_member_search.go:15-31`, `modules/space/db.go:116-124,164-203`, `modules/space/api.go:100-103`

The handler authorizes purely from the `space_member` table and **never verifies the space itself is active**:

- The gate is `queryMember` (`db.go:116`, `WHERE space_id=? AND uid=? AND status=1`) plus `member.Role < spaceRoleAdmin` (`api_member_search.go:29`). There is no call to the package's existing `checkSpaceActive` / `isSpaceActive` (`api.go:51-63`, which checks `space.status=1`).
- Neither route-group middleware closes the gap: `AuthMiddleware` resolves the login token only, and `SharedUIDRateLimiter` is pure per-UID frequency limiting — neither is space-status aware (`api.go:100`).
- `searchMembers` / `countSearchMembers` filter only `sm.space_id=? AND sm.status=1` (`db.go:175,196`); they never join `space.status`.

The exploitability hinges on the fact that the two relevant state transitions leave `space_member` rows at `status=1`:

- **Ban** (`PUT /manager/spaces/:space_id/status/2` → `updateSpaceStatus`, `db_manager.go:223-229`): sets `space.status=2` only; member rows untouched.
- **User disband** (`DELETE /v1/space/:space_id` → `DB.disbandSpace`, `db.go:62-65`): sets `space.status=0` only; member rows untouched.

Only the unrelated admin **force-disband** path (`forceDisbandSpace`, `db_manager.go:169-186`) clears member rows in a transaction — it does not cover the ban or user-disband paths above.

**Impact:** after a space is banned by an operator (or disbanded by its owner), a former admin/owner of that space can keep calling this endpoint and search the full active member roster, including **masked email and phone** that this endpoint newly surfaces. Banning is meant to freeze access to the space; this new endpoint bypasses that intent. Data is masked and the endpoint is admin-gated and rate-limited, which is why this is P1 rather than P0 — but it is a real, reachable PII exposure and an authorization-lifecycle regression relative to the package's established `checkSpaceActive` convention.

**Fix:** add an active-space guard to the handler (call `checkSpaceActive(c, spaceId)` after auth, as the mutating handlers do — `api.go:375,647,...`), or join `space s ON s.space_id=sm.space_id AND s.status=1` in the auth/search queries. Joining is the more robust option because it stops depending on the implicit "did every status-change path remember to also clear member rows" invariant. (Note: the sibling `listMembers` shares the missing-active-space-check gap but does not expose email/phone; consider hardening it in the same follow-up, though that is out of scope for this PR.)

## Non-blocking findings

### P2 — `searchMembers` drops the robot creator-visibility filter that `listMembers` enforces
`modules/space/db.go:164-189` vs `db.go:146-162`

`queryMembers` restricts robots to those created by the caller (`AND (r.robot_id IS NULL OR r.creator_uid = ?)`). `searchMembers` joins `robot` only to emit the `robot` flag and applies no `creator_uid` predicate, so this admin search surfaces **all** bots in the space regardless of creator. This may be intentional (an admin view legitimately sees every bot), but it is a behavioral divergence from the member-facing list that is currently undocumented and untested — please make it an explicit decision. If the creator filter is ever added to the list query, it must be mirrored in `countSearchMembers` (which omits the robot join entirely) or list/total will desync.

### P2 — Masked PII search remains an oracle
`modules/space/api_member_search.go:55-58`, `db_manager.go` (`memberSearchColumns` includes `u.email`, `u.phone`)

`keyword` matches against **plaintext** email/phone while the response masks them. A space admin can therefore confirm or progressively reconstruct a member's exact email/phone by issuing targeted substring queries and observing `count`/list membership, partially defeating the masking. Given admin-gating + the per-UID rate limiter this is not critical, but consider excluding `email`/`phone` from the searchable columns (or gating exact-contact search behind a stronger capability) if masking is meant as a real privacy boundary.

### P2 — `maskPhone` / `maskInviteEmail` byte-slice their inputs
`modules/space/api_member_search.go:77-85`, `modules/space/email_invite_token.go:51-55`

Both use byte indexing (`p[:3]`, `p[len(p)-2:]`, `local[:keep]`, `domain[:1]`). Member-search values come straight from the `user` table (`db.go:169 IFNULL(...)`) without going through `validateInviteEmail`, so a stored multibyte email/phone would be sliced mid-rune and emit invalid UTF-8 in the JSON response. Realistically phones/emails are ASCII so this is latent, but rune-aware slicing (or a documented ASCII assumption) would harden a PII path.

### nit — Redundant list query when count is 0; unbounded `page_index`
`modules/space/api_member_search.go:36-44`, `modules/space/db.go:186`

When `countSearchMembers` returns 0 the handler still runs the list query — a small early-return saves a round-trip. Separately, `clampPage` bounds `page_size` (≤200) but not `page_index`; a hostile `page_index` near max-int makes `(pageIndex-1)*pageSize` overflow before the `uint64` cast, yielding a garbage `OFFSET`. Not a security issue (admin-gated + rate-limited, and the manager paths share this), but a `page_index` cap would make the pagination math total.

## Coverage / blind spots

- **Tests** comprehensively cover authz error codes, column matching, masking (ASCII), `_`-wildcard escaping, empty contacts, bots-included, list/count parity, removed-member exclusion, and the 200-row clamp.
- **Gaps:** (1) no test asserts a **banned (status=2) or disbanded (status=0) space is rejected** — the blocking gap above is therefore silent in CI; (2) the LIKE-escaping test exercises only `_`, not `\` or `%`, so a regression in backslash/percent handling or a dropped `ESCAPE` clause would go uncaught; (3) no multibyte/UTF-8 masking test. The robot fixture is always created by the caller, so it would not catch a (re)introduced creator filter.
- **Not executed here:** DB-backed integration paths (count/list parity, join cardinality, real `ESCAPE`/`sql_mode` behavior) — verified by reading, not running. `robot_id` is UNIQUE so the list LEFT JOIN does not multiply rows.

## Verdict

Requesting changes for the single P1 active-space authorization gap (banned/disbanded-space member/PII exposure). The P2/nit items are non-blocking and can be follow-ups, though the robot-visibility divergence (P2) is worth a conscious decision before merge.

lml2468 (approved):
✅ 基于最新 commit dde777c4 重审,APPROVE。这次提交全面回应了前几轮 review 的反馈,且**没碰坏已修的封禁/解散空间 blocker**。逐项亲核(增量 diff + 关键实现)。

## 已修 blocker 完好(重点确认没被这次改动碰坏)
api_member_search.go:26 仍是 `if s.checkSpaceActive(c, spaceId) { return }`,在所有成员查询之前;isSpaceActive 严格 status=1 → 封禁(2)+解散(0)统一挡。回归测试 TestSpaceMembersSearchRejectsInactiveSpace 仍在。这次的 PII 改动没影响生命周期 gate。

## PII oracle 已消除(此前 review 提的"明文搜+掩码显可逐位探测"的正解)
- phone:检索列改成 RIGHT(u.phone,4)(只匹配后 4 位),响应 maskPhone 显示前3+后4(138****5678)。真正被掩码的**中间位既不可检索也不可见** → admin 无法通过子串查询逐位探测/重建完整号码。可检索粒度 == 可见粒度,oracle 关闭。
- email:改为明文检索 + 明文返回(一致,无探测差)。这是把工作邮箱定为"空间 admin 可见明文"的产品决定 —— 端点是 space-admin-only + 仅激活空间 + per-UID 限流,commit message 也明确记录了理由。作为 reviewer 标注这是有意的可见性决定(从掩码改为明文),非泄露 bug;若产品侧要求 email 也掩码可再调,但当前定位合理。
- maskPhone 改用 []rune 切片(修了之前 byte-slice 对多字节输入切坏 UTF-8 的 nit),<7 位全掩 ***,空值保持空。

## 其余前轮反馈也一并处理
- page_index 上限 maxMemberSearchPageIndex=100000,避免 (pageIndex-1)*pageSize 转 uint64 前整型溢出(yujiawei 提的)。
- count=0 短路,省一次 list 查询。
- 分页 uid tiebreaker 保留(确定性翻页)。

## 安全主判断无回归
- SQL:memberSearchActiveWhere 仍参数化 LIKE ? + ESCAPE,列名硬编码白名单(RIGHT(u.phone,4) 是固定 SQL 表达式、非用户输入)、无注入。
- 授权 scope:同空间绑定(WHERE sm.space_id=?),无跨空间 IDOR。
- list/count 共用 memberSearchActiveWhere,分页一致。
- go vet 干净;maskPhone + 生命周期拒绝有单测覆盖(包级测试依赖 MySQL,本地无库,CI 跑)。

我此前在 379134ab 已 APPROVE,本 commit 在此基础上修齐了 PII oracle / page_index / 多字节掩码,blocker 保持。无 blocking。批准。

—— Octo-PR (lml2468)

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#389 Review Report — automated review (Octo-Q)

**PR**: Mininglamp-OSS/octo-server #389
**Head SHA**: `dde777c42beea744b07f7999dc84cf127e0f3af4`
**Title**: feat(space): add paginated member search endpoint for space admins

---

## 1. 验证结论

| 项 | 状态 | 证据 |
|---|---|---|
| Space lifecycle gate (`checkSpaceActive`) | ✅ | `api_member_search.go:28` → `db.go:28-35` (`SELECT COUNT(*) FROM space WHERE space_id=? AND status=1`) blocks banned(2) + disbanded(0) |
| Member authz (non-member / role < admin) | ✅ | `api_member_search.go:30-42` → `db.go:116-124` (`WHERE space_id=? and uid=? and status=1`) |
| Phone search/display parity (last-4 only) | ✅ | `db_manager.go:57` `RIGHT(u.phone,4)` in WHERE; `api_member_search.go:78` `maskPhone()` in response |
| LIKE injection / wildcard escaping | ✅ | `db_manager.go:16-21` `escapeLike()` escapes `\`, `%`, `_`; `likeEscapeClause` at L24 handles `NO_BACKSLASH_ESCAPES` |
| Pagination clamp + overflow guard | ✅ | `api_manager.go:37-48` `clampPage()` caps page_size at 200; `api_member_search.go:46-48` caps page_index at 100000; max offset = 19,999,800 |
| Count/list filter consistency | ✅ | Both call `memberSearchActiveWhere()` (`db_manager.go:63-72`); count omits robot JOIN (not needed for WHERE) |
| Rate limiting | ✅ | `api.go:100` `SharedUIDRateLimiter` after `AuthMiddleware` — matches established pattern (app_bot, group, etc.) |
| Keyword not logged | ✅ | `api_member_search.go:88-89` logs operator_uid, space_id, count — no keyword field |
| Test coverage | ✅ | 8 test functions covering authz, lifecycle, empty keyword, bots, column matching, escaping, pagination clamp |

## 2. 发现的问题

### P2-1 — Swagger description/code mismatch on page_size default

**File**: `modules/space/swagger/api.yaml` L42
**Diff-scope**: **new** (this PR's swagger file)

Swagger description states: `"每页条数，省略或 <=0 时为 15，上限 200"` — but `clampPage()` (`api_manager.go:41`) defaults `pageSize` to **20** when ≤ 0, not 15. The PR body mentions "swagger `page_size` default fixed (15)" but the description text still says 15 while the code returns 20. Frontend developers reading the spec will expect 15 rows per page by default but receive 20.

**Fix**: Change the swagger description to `"每页条数，省略或 <=0 时为 20，上限 200"`.

### P2-2 — `spaceRoleAdmin` constant is file-local; 10+ existing call-sites use magic number `1`

**File**: `modules/space/api_member_search.go:13`
**Diff-scope**: **new** (constant introduced in this PR)

The new file defines `const spaceRoleAdmin = 1` and uses it for the role gate. However, the existing codebase uses the bare magic number `1` in `member.Role < 1` at 10+ call-sites in `api.go` (L384, 656, 750, 917, 1267, 1332, 1394, 1437, 1507, 1637). The named constant is a strict improvement, but the inconsistency means two conventions now coexist.

**Suggestion**: Promote `spaceRoleAdmin` to a package-level constant (e.g., in `model.go`) and update existing call-sites in a follow-up, or keep the magic number to match existing style. Non-blocking.

### Note-1 — Pre-existing: `queryMember` returns `(nil, nil)` on DB error

**File**: `modules/space/db.go:116-124`
**Diff-scope**: **pre-existing** (acknowledged in PR body under "Known non-blocking follow-ups")

When `dbr.Load()` fails AND leaves the struct zero-valued (`m.UID == ""`), `queryMember` returns `(nil, nil)`, causing the handler to surface a DB failure as `ErrSpaceNotMember` (400) instead of 5xx. Affects ~6 call-sites. Not introduced or amplified by this PR.

## 3. 建议

1. Fix swagger `page_size` description: `15` → `20` (one-line change).
2. Consider promoting `spaceRoleAdmin` to a shared constant in a follow-up to eliminate the magic-number drift.
3. (Out of scope for this PR) Fix `queryMember`'s `(nil, nil)` error-swallowing pattern.

## 4. 额外发现

- **Route group separation is correct**: The new `search` group (`api.go:100-102`) is separate from the existing `auth` group (`api.go:72`) because it adds `SharedUIDRateLimiter`. This matches the established pattern used by `app_bot`, `group`, `incomingwebhook`, etc. No route collision between `/:space_id/members` and `/:space_id/members/search` (different path depth).
- **`maskPhone` is rune-aware**: Correctly handles multi-byte input (`api_member_search.go:96-103`). Edge cases (empty → empty, <7 runes → `***`) are tested.
- **`count=0` short-circuit**: Avoids the list query when no matches exist. The count→list window has a theoretical TOCTOU race (member added between queries), but this is standard for paginated search and acceptable.

## 5. 数据流回溯

| 消费点 | 上游来源 | 数据是否真的流到消费点 |
|---|---|---|
| `member.Role` (authz gate L40) | `queryMember()` → `SELECT * FROM space_member WHERE status=1` | ✅ Active members only; removed members (status=0) return nil → `ErrSpaceNotMember` |
| `m.Name/Username/Email/Phone` (response L74-77) | `searchMembers()` → `IFNULL(u.name,'')` etc. via LEFT JOIN user | ✅ IFNULL ensures non-nil strings; empty contacts stay empty (tested) |
| `m.Robot` (response L80) | `CASE WHEN r.robot_id IS NOT NULL AND r.status=1 THEN 1 ELSE 0 END` via LEFT JOIN robot | ✅ Non-robot members get 0; only active robots (status=1) flagged |
| `maskPhone(m.Phone)` (response L78) | `IFNULL(u.phone,'')` from SELECT | ✅ Empty phone → `maskPhone("")` → `""`; normal phone → `138****5678` |
| `count` (response L90) | `countSearchMembers()` → `COUNT(*)` with same WHERE as list | ✅ Shared `memberSearchActiveWhere()` ensures count/list parity |
| `checkSpaceActive` (gate L28) | `isSpaceActive()` → `SELECT COUNT(*) FROM space WHERE status=1` | ✅ Blocks status=0 (disbanded) and status=2 (banned); tested both |

## 6. 三类盲点 checklist (R5)

- **C1 双路径 parity**: N/A — This PR adds a new read-only endpoint. No symmetric add/remove/subscribe/unsubscribe paths are modified.
- **C2 control-flow ordering / 嵌套复用**: Clear — `memberSearchActiveWhere()` is new and called from exactly 2 sites (count + list). `buildLikePattern()` / `escapeLike()` are existing utilities reused correctly. No double-application or ordering risk.
- **C3 授权边界 ≠ 能力边界**: Clear — Endpoint is behind `AuthMiddleware` (valid token required) → `checkSpaceActive` (space status) → `queryMember` (membership) → `Role < spaceRoleAdmin` (admin+). Only authenticated space admins can reach the search logic. `SharedUIDRateLimiter` provides per-UID throttling.
- **C4 授权生命周期 / 容器-成员状态级联**: Clear — `checkSpaceActive` at `api_member_search.go:28` explicitly checks the **container** (space) status=1 before any member-level check. This directly addresses the C4 pattern: banning a space only updates the space row, leaving member rows at status=1. The lifecycle gate correctly blocks banned/disbanded spaces. Tested at `TestSpaceMembersSearchRejectsInactiveSpace` with both status=2 (banned) and status=0 (disbanded).

## 7. 跨轮 blocker 复检 (R6)

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 issues found. The endpoint correctly implements the C4 container-member lifecycle gate (the key security concern for this PR), has comprehensive test coverage, and follows established codebase patterns. Two P2 findings (swagger description mismatch, local constant vs magic number inconsistency) are non-blocking documentation/style issues.

Jerry-Xin (approved):
The PR is relevant to octo-server and the implementation is ready to merge; I found no blocking correctness or security issues.

💬 Non-blocking:
- 🟡 Warning: [modules/space/swagger/api.yaml](/Users/mlamp/.openclaw/tmp/pr-review-r1ldq290/repo/modules/space/swagger/api.yaml:22) says `email/phone` are masked, but the endpoint intentionally returns email in cleartext. The schema later says email is cleartext, so the top-level description should be corrected to avoid client confusion.
- 🟡 Warning: [modules/space/swagger/api.yaml](/Users/mlamp/.openclaw/tmp/pr-review-r1ldq290/repo/modules/space/swagger/api.yaml:44) documents the default `page_size` as 15, while [modules/space/api_member_search.go](/Users/mlamp/.openclaw/tmp/pr-review-r1ldq290/repo/modules/space/api_member_search.go:45) uses `clampPage`, whose default is 20. Please align the docs with behavior.
- 🔵 Suggestion: [modules/space/api_member_search_test.go](/Users/mlamp/.openclaw/tmp/pr-review-r1ldq290/repo/modules/space/api_member_search_test.go:331) checks only HTTP 400 for inactive spaces. Adding an assertion for the localized error code would pin the lifecycle contract more tightly.

✅ Highlights:
- The endpoint uses active-space gating before member search and prevents banned/disbanded-space member exposure.
- Authorization is scoped to active space admins/owners, and legacy raw error responses are covered by the guard test.
- Search uses parameterized SQL plus escaped `LIKE`, avoids keyword logging, masks phone output, and shares the same filter for count/list.
- Pagination is capped and includes deterministic ordering with `sm.uid`.

Verification note: I attempted `go test ./modules/space -run 'TestSpaceMembersSearch|TestMaskPhone|TestSpaceNoLegacyResponseError'`, but the local test DB is in an unrelated bad migration state: `unknown migration in database`.

yujiawei (approved):
# Code Review — PR #389 (octo-server)

Reviewed at head SHA `dde777c42beea744b07f7999dc84cf127e0f3af4` against `origin/main` (merge-base `c60bbd84`). `go build ./modules/space/...` and `go vet ./modules/space/...` are clean. DB-backed integration tests were not executed here (no MySQL/Redis/WuKongIM infra); correctness was verified by reading the source, the included unit tests, the dependency rendering (dbr, octo-lib `GetPage`), and a standalone reproduction of `maskPhone`.

This is a fresh review of the current head (the branch was squashed to a single commit newer than all earlier reviews). The previously-blocking lifecycle authorization gap is now fixed and is not re-raised — see below.

## Summary

A well-scoped, well-tested addition. `GET /v1/space/:space_id/members/search` mirrors the existing manager / `listMembers` patterns and reuses the shared building blocks (`clampPage`, `SharedUIDRateLimiter`, the `escapeLike` + `ESCAPE '\\'` LIKE machinery). The core security-sensitive properties hold:

- **Authorization scope is correct.** The gate binds the role check to the path `space_id` (`api_member_search.go:24-37`): non-member → `ErrSpaceNotMember`, role `< admin` → `ErrSpacePermissionDenied`. No cross-space IDOR; the search/count queries also filter `sm.space_id=? AND sm.status=1`.
- **Lifecycle gate is now present.** `searchMembers` opens with `checkSpaceActive(c, spaceId)` (`api_member_search.go:21-23`), which enforces strict `space.status=1`. Because a ban (`status=2`) and a user disband (`status=0`) both leave `space_member` rows at `status=1`, the member-row gate alone was insufficient — the explicit space-status check closes that PII-after-ban hole. Regression test `TestSpaceMembersSearchRejectsInactiveSpace` pins both the banned and disbanded cases.
- **SQL injection is prevented.** All keyword matching is parameterized `LIKE ? ESCAPE '\\'` over a hard-coded column whitelist (`memberSearchActiveColumns`); `_ / % / \` are escaped via `buildLikePattern`/`escapeLike`.
- **count/list stay consistent.** Both `countSearchMembers` and `searchMembers` apply the same `memberSearchActiveWhere` + `sm.status=1`. The list-only `LEFT JOIN robot` cannot multiply rows (`robot_id` is UNIQUE — `robot_id_robot_index`), so totals match across pages.
- **PII / oracle.** Phone is searchable only on the last 4 (`RIGHT(u.phone,4)`) and displayed masked, so "searchable granularity == visible granularity" — the substring-probe reconstruction oracle is closed. Email is intentionally cleartext search + cleartext return (consistent, no probe differential); this is a deliberate "work email is visible to space admins" product decision, documented in the definition block. Keyword is never logged.

No P0/P1 (correctness / security / data-loss / build-break) issues. The items below are non-blocking.

## Non-blocking findings

### P2 — `maskPhone` only masks meaningfully for 11-digit numbers; fully reveals 7-char values
`modules/space/api_member_search.go:73-85`, test pin `api_member_search_test.go:128`

`maskPhone` returns `first3 + "****" + last4`. The `len(r) < 7` guard means a **7-rune** value passes through with first-3 and last-4 covering the *entire* string — 0 digits actually hidden (`"1234567"` → `"123****4567"`). For 8/9/10-rune values only 1/2/3 middle digits are hidden. Only the dominant 11-digit mainland-mobile case hides the intended 4 digits. `user.phone` is `VARCHAR(100)` with no enforced length/format, so international numbers, landlines, and short/local formats are storable and would be displayed in full or near-full.

This is a weakening of the stated masking contract ("前3+****+后4") for non-standard phone lengths. I keep it **non-blocking** because: (a) the audience is same-space admins/owners, who already receive every member's **email in full cleartext** by the same accepted product decision — so the privacy model already trusts this audience with member contacts; (b) the actual reconstruction attack the masking was hardened against (search-oracle) remains closed; (c) the dominant real-world input masks correctly; (d) the 7-char behavior is currently pinned *as intended* by a test, i.e. a conscious choice. Recommend either masking proportionally (always hide the middle, e.g. require `len >= 8` for the 3+4 reveal and fully mask shorter values) or documenting the "designed for 11-digit mobile" limitation — and revisiting the test that locks in the full-reveal.

### P2 (nit) — Stale endpoint description in swagger
`modules/space/swagger/api.yaml:22`

The path-level `description` still says `email/phone 返回掩码`, but email is now returned in cleartext (the `memberSearchResp.email` definition at line 130 correctly says `明文，工作邮箱不掩码`, and the `keyword` param doc at line 34 is also correct). Align the top-level description with the deliberate cleartext-email policy to avoid integrator confusion.

### P3 (nit) — Local `spaceRoleAdmin = 1` constant
`modules/space/api_member_search.go:13`

Role semantics (`0 member / 1 admin / 2 owner`) are expressed as bare literals elsewhere in this package too, so this is consistent with existing style — but a shared role constant would prevent drift. Optional.

## Coverage / blind spots

- **Resolved since prior reviews:** the banned/disbanded-space member/PII exposure (previously the single blocker) is fixed via `checkSpaceActive` + the `TestSpaceMembersSearchRejectsInactiveSpace` regression test — verified by reading, not just commit message. The earlier `page_size` swagger-vs-runtime drift is also gone (`GetPage` defaults `page_size` to 15 and the swagger now documents 15).
- **Pagination overflow:** `clampPage` forces `pageIndex >= 1`, and `maxMemberSearchPageIndex = 100000` caps the offset (`<= 2e7` at `page_size <= 200`). On the amd64 build target `int` is 64-bit so the `int64 → int` conversion in `clampPage(c.GetPage())` is lossless; the math is total. (A 32-bit build would need the cap applied before the conversion, but that is not a supported target.)
- **SQL precedence:** the chained `.Where("sm.space_id=? AND sm.status=1")` + `.Where("<col OR col ...>")` renders as `(sm.space_id=? AND sm.status=1) AND (col LIKE ? OR ...)` — confirmed by dbr's `buildCond`, which wraps every condition in its own parentheses before joining with `AND`. There is no OR-precedence leak across spaces/statuses.
- **Not executed here:** DB-backed integration paths (real `RIGHT()`/`ESCAPE`/`sql_mode` behavior, join cardinality under real rows) — verified by reading. `RIGHT()` is standard MySQL.
- **Known follow-ups acknowledged by the author** (out of scope): authz tests assert HTTP status only, not `error.code`; the pre-existing `queryMember` swallowing a DB error as "not a member" affects ~6 call-sites and is a separate fix.

## Verdict

Approving. No correctness, security, data-loss, or build-breaking issues; the previously-blocking lifecycle gap is fixed with a regression test. The `maskPhone` short-number weakness (P2) is worth hardening but is consistent with the already-accepted cleartext-email decision for the same trusted audience, and the dominant case masks correctly — so it does not block merge.

PR #393, opened 2026-06-17
fix(opanalytics): widen dim columns & resolve Space prefix in private member uids (#392)

### LINKED ISSUE #392: opanalytics ETL fails with Error 1406: octo_dim_member.phone/email columns narrower than source user table
## Summary

`opanalytics` ETL (`refreshDimMembers`) fails on production with **`Error 1406 (22001): Data too long for column 'phone' at row N`** because `octo_dim_member.phone` is `VARCHAR(20)` but the source `user.phone` is `VARCHAR(100)`. Same risk exists for `octo_dim_member.email` (`VARCHAR(100)`) vs `user.email` (`VARCHAR(255)`).

Result: ETL never completes → all `octo_*` analytics tables stay empty → `/v1/manager/dashboard/*` returns zero / errors.

## Reproduction

```sql
-- Source user table (existing schema)
DESC `user`;
-- ... phone varchar(100), email varchar(255) ...

-- Dim member table created by 20260608000001_opanalytics_init.sql
DESC octo_dim_member;
-- ... phone varchar(20), email varchar(100) ...

-- Real production data has rows wider than 20 bytes in user.phone (legacy/dirty
-- data: some rows use phone as a tombstone marker like `@1779881561532@delete`
-- or accidentally store an email address). ETL bulk INSERT of those rows hits
-- Error 1406 and the entire transaction rolls back.
```

Production log:
```
{\"level\":\"error\",\"msg\":\"opanalytics ETL failed\",\"source\":\"manual\",
 \"error\":\"Error 1406 (22001): Data too long for column 'phone' at row 11\"}
```

## Root cause

\`modules/opanalytics/sql/20260608000001_opanalytics_init.sql\` defines:

\`\`\`sql
\`email\` VARCHAR(100) NOT NULL DEFAULT '' COMMENT '邮箱 (user.email)',
\`phone\` VARCHAR(20)  NOT NULL DEFAULT '' COMMENT '手机号 (user.phone)',
\`\`\`

The comment says \"\`user.phone\`\" / \"\`user.email\`\" but the column widths were picked from the expected semantic shape (a phone number ≤ 20 chars) rather than the source-table declared width. The source \`user\` table's \`phone\` has been \`VARCHAR(100)\` for a long time and contains legitimately wider rows in production deployments (whether or not those rows are clean is a separate concern; the dim table must hold what the source can produce).

ETL bulk INSERT does no truncation/sanitization (\`modules/opanalytics/etl_db.go: replaceDimMembers\`), so MySQL's strict mode (default on MySQL 8) rejects the whole batch the moment it hits the first oversize row.

## Fix

Either:

**A. Widen the dim columns to match the source** (minimum-friction, applies to both new installs and existing deployments via a forward migration):

\`\`\`sql
ALTER TABLE octo_dim_member
  MODIFY COLUMN phone VARCHAR(100) NOT NULL DEFAULT '' COMMENT '手机号 (user.phone)',
  MODIFY COLUMN email VARCHAR(255) NOT NULL DEFAULT '' COMMENT '邮箱 (user.email)';
\`\`\`

Plus updating \`20260608000001_opanalytics_init.sql\` so fresh installs are correct.

**B. Truncate in ETL** (defense in depth):

\`refreshDimMembers\` could \`utf8.SafeTruncate(u.Phone, 100)\` / etc. before appending the batch row. Useful if the project considers \`phone\` as a typed concept and wants to keep the dim column narrower than the source. Not sufficient alone — if a future deployment's source widens further, you'd silently lose data.

Recommend **A** as the primary fix and consider **B** as an additional safety net that logs whenever truncation occurs (so dirty-data accumulations are visible).

## Wider lesson

This is the second time in ~3 weeks an analytics-layer migration drifts away from the source table's actual schema (the previous one was the \`utf8mb4_general_ci\` vs \`utf8mb4_0900_ai_ci\` mismatch, issue #150). Both could be caught by a CI lint that:

- Verifies any \`octo_dim_*\` / \`octo_fact_*\` column whose comment says \"(\`<table>.<column>\`)\" matches the declared type/width/collation of the referenced column.
- Or, more lightly, an explicit per-column max width assertion in \`refreshDim*\` unit tests against the real \`user\` table fixture.

## Environment

- MySQL 8.0.x (strict mode, default)
- octo-server \`main\` at the time of investigation
- migration introduced: \`modules/opanalytics/sql/20260608000001_opanalytics_init.sql\`

an9xyz:
Fixes #392.

## Summary

The opanalytics ETL crashed in production with `Error 1406 (22001): Data too long for column ...`, rolling back the whole ETL transaction and leaving every `octo_*` dashboard table empty (`/v1/manager/dashboard/*` returns zero/errors). There are **two** root causes — a schema-width drift and a deeper semantic bug — both fixed here, plus an index-prefix safeguard (see Update).

### 1. Dim columns narrower than the source (the `1406` crash)
- `octo_dim_member.phone` was `VARCHAR(20)` while the source `user.phone` is `VARCHAR(100)` — **the actual crash**. Production holds legitimately wider/dirty rows (e.g. an email accidentally stored in `phone`, tombstone markers like `@...@delete`). Widened to `VARCHAR(100)` to match source.
- `octo_dim_member.email`: the source `user.email` is **`VARCHAR(100)`** (verified against a live schema — the issue's "255" was inaccurate), and `dim_member.email` was already `VARCHAR(100)`, so it already matched. It's widened to `VARCHAR(255)` purely as headroom and to converge with the manual production hotfix — **not** to mirror the source.
- `octo_dim_channel.member_a_uid` / `member_b_uid` were `VARCHAR(40)` (sized for a bare uid), but the halves of a private `channel_id` can carry Space/adapter prefixes (`s{32hex}_uid` ≈ 66, `sminglue_default_uid` ≈ 49). Widened to `VARCHAR(100)`.

**Fix:** a forward migration `20260617000001_opanalytics_widen_dim_cols.sql` widens the columns; the init SQL is updated so fresh installs are correct. `ALTER ... MODIFY` is idempotent, so it's a no-op on the env that already received the manual hotfix and a real fix everywhere else.

### 2. Prefixed member uids never JOIN `dim_member` (silent data corruption)
`member_a/b_uid` stored the **prefixed** channel halves, which never match `dim_member.uid` (bare `user.uid`). That silently broke private HH/HA classification, member attribution, and the "A & B" name display — the `LEFT JOIN` always missed on the prefixed side.

**Fix:** `normalizePrivatePair` now reverse-resolves each half via `space.ParseChannelID` to the bare uid before normalizing (bare uids pass through unchanged). `space.ParseChannelID` already handles **both** the 32-hex Space prefix (regex) and named spaces like `minglue_default` (registered list) — `sminglue_default_` is just `s` + space `minglue_default` + `_`, so no separate adapter-stripping logic is needed. This single change fixes member-uid JOINs, conv_type classification, and system-bot exclusion for private chats. Widening stays as a safety net for halves whose prefix can't be resolved.

> **`sender_uid` deliberately untouched.** `octo_fact_member_channel_daily.sender_uid` stays `VARCHAR(40)`; `accumulateFact3` writes it *before* the private-row upsert, and the hotfix unblocked the ETL after widening only `phone`/`member_a/b`, so `from_uid` (= `sender_uid`) fits within 40 in production.

## Update — index-prefix safeguard (review follow-up)

Widening `email` to `VARCHAR(255)` while keeping the full-column `KEY idx_email (email)` would push the index key to 255×4 = **1020 bytes**, exceeding InnoDB's **767-byte** single-column prefix limit on COMPACT/REDUNDANT row formats — re-breaking the migration with `Error 1071` on those instances (MySQL 8.0 default DYNAMIC is unaffected at 3072). Reproduced locally on a COMPACT table: old approach → `Error 1071`; new approach → succeeds.

`idx_email` is now a **191-char prefix index** (`email(191)` → 191×4 = 764 < 767), safe on every row format. The migration drops the old index, widens the column, and re-adds the prefix index in one `ALTER` so the oversized full index is never materialized; the hotfixed prod env converges its leftover full index to the prefix form. `idx_phone (zone,phone)` = 480 bytes and `idx_name` = 400 bytes are already under the limit and untouched.

## Testing

Validated against a local MySQL 8.0.46 + Redis (full `module.Setup` migration run against a freshly created DB, so every migration is exercised end-to-end):

- `go test ./modules/opanalytics/` — **all pass**
- `TestNormalizePrivatePairStripsSpacePrefix` (unit): regex 32-hex prefix, registered named space, bare, and mixed cases.
- `TestOpanalyticsETLStripsSpacePrefixForPrivateMembers` (integration): seeds a space-prefixed private channel, asserts `member_a/b_uid` land as bare uids, `conv_type` = HA, and that both halves JOIN back to `dim_member`.
- Index safeguard reproduced directly: old full-index `MODIFY` → `Error 1071` on COMPACT; new prefix-index `ALTER` → idx_email = 764 bytes.
- `go vet` / `gofmt` clean.

## Operational notes

- One production env already received an equivalent manual hotfix (phone 20→100, email 100→255, member_a/b_uid 40→100); this migration converges to the same state and additionally tightens `idx_email` to a 191 prefix.
- Existing private `dim_channel` rows written with prefixed `member_*_uid` self-heal to bare uids the next time a chunk touches that channel; channels with no new activity need a `Rebuild()` to backfill.

## Out of scope (follow-up)

The CI lint suggested in the issue (assert every `octo_dim_*`/`octo_fact_*` column annotated `(table.column)` matches the source type/width, and that uid columns round-trip through `dim_member.uid`) is a separate guardrail — recommend a dedicated issue/PR.


lml2468 (dismissed):
✅ APPROVE — 扎实的 opanalytics bug fix(修 #392 的 ETL 1406 崩溃 + 私聊成员 Space 前缀语义错算)。两个 root cause 都修对、逐项亲核无回归风险,无 blocking。

## Fix 1：维表列宽 < 源列(1406 崩溃)——正确
- dim_member.email V(100)→V(255)、phone V(20)→V(100):对齐源 user.email(255)/user.phone(100),容下源能产出的合法/历史脏数据。
- dim_channel.member_a/b_uid V(40)→V(100):容纳前缀未能反解时的兜底宽度。
- 迁移 20260617000001:ALTER MODIFY 幂等(可在已 hotfix 的生产安全重跑)、带 Down(可逆)、注释把宽度依据讲清。init.sql 同步改为目标态,与迁移收敛一致。

## Fix 2：私聊 channel 两端 Space 前缀反解(核心语义修复)——正确,无回归
normalizePrivatePair 从直接用 parts[0]/parts[1] 改为 spacepkg.ParseChannelID() 反解裸 peer uid。亲核 ParseChannelID(pkg/space/channel.go:44):
- 返回 (spaceID, peerID),这里取 peerID(裸 uid)。
- **裸 uid(不以 "s" 前缀开头)→ return "", channelID 原样返回** → 现有无前缀私聊不受影响,无回归。
- 带前缀剥离有双重保险:① 已知 spaceID 列表精确匹配 sid+"_"(命名空间需注册);② 正则回退严格 `^s([0-9a-f]{32})_`(必须 s+正好32位hex+下划线)。误伤面极小——普通 uid 既过不了 32hex 正则、也不在注册列表,原样返回。
- 新增 `if a=="" || b=="" → ok=false` 守卫,防反解出空 uid 写入维表。
效果:member_*_uid 写裸 uid → 能 JOIN dim_member.uid,修复私聊 HH/HA 分类、成员归属、名称展示的静默错算。

## 测试 + 验证
- 纯单测 TestNormalizePrivatePair + TestNormalizePrivatePairStripsSpacePrefix 本地实跑 PASS,覆盖 both-prefixed(正则回退)/前缀+hash序交换/mixed bare+prefixed/命名空间(注册)/both-bare(不变)各场景。
- 端到端 TestOpanalyticsETLStripsSpacePrefixForPrivateMembers 验证前缀私聊经 ETL 后 member_*_uid 为裸 uid、能 JOIN dim_member、conv_type 正确判为 HA(前缀未反解会误判 HH)。
- go vet 干净。(依赖 DB 的端到端测试本地无 MySQL,CI 跑。)

无 blocking、无 should-fix。两个 root cause 修得干净、迁移幂等可逆、反解逻辑对裸 uid 无副作用、测试覆盖到位。批准。

—— Octo-PR (lml2468)

Jerry-Xin (dismissed):
This PR is in scope for octo-server and fixes opanalytics ETL schema drift plus the private-member uid normalization bug without introducing blocking issues.

💬 Non-blocking
- 🔵 Suggestion: The new tests mutate the global `pkg/space` registered-space cache and reset it to `nil` afterward ([modules/opanalytics/api_test.go](/Users/mlamp/.openclaw/tmp/pr-review-0_py8wkr/repo/modules/opanalytics/api_test.go:405), [modules/opanalytics/etl_test.go](/Users/mlamp/.openclaw/tmp/pr-review-0_py8wkr/repo/modules/opanalytics/etl_test.go:71)). Current tests are not parallel, so this is acceptable, but future parallelization could make this order-sensitive. A save/restore helper or package-level cleanup pattern would make it more robust.

✅ Highlights
- `normalizePrivatePair` now strips Space prefixes before sorting and storing member uids, which aligns `octo_dim_channel.member_*_uid` with `octo_dim_member.uid` and fixes classification/join behavior ([modules/opanalytics/etl.go](/Users/mlamp/.openclaw/tmp/pr-review-0_py8wkr/repo/modules/opanalytics/etl.go:406)).
- The migration and init SQL correctly widen `email`, `phone`, and private member columns for existing and fresh installs ([modules/opanalytics/sql/20260617000001_opanalytics_widen_dim_cols.sql](/Users/mlamp/.openclaw/tmp/pr-review-0_py8wkr/repo/modules/opanalytics/sql/20260617000001_opanalytics_widen_dim_cols.sql:18), [modules/opanalytics/sql/20260608000001_opanalytics_init.sql](/Users/mlamp/.openclaw/tmp/pr-review-0_py8wkr/repo/modules/opanalytics/sql/20260608000001_opanalytics_init.sql:7)).
- Coverage includes both unit normalization cases and an end-to-end ETL assertion that private member uids join back to `dim_member`.

Verification: focused unit tests passed with `go test ./modules/opanalytics/ -run 'TestNormalizePrivatePair|TestDayWindowUnix|TestIsExcludedMember|TestConvType'`. The full `go test ./modules/opanalytics/` could not run in this workspace because the shared test DB has an existing migration-state error before the PR-specific tests execute.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#393 Review Report
**PR:** Mininglamp-OSS/octo-server#393
**Head SHA:** `8d943e235b4b078b8b877c6cde5852ef4d2b68b8`
**Title:** fix(opanalytics): widen dim columns & resolve Space prefix in private member uids (#392)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Schema widening (phone/email/member_a/b_uid) | ✅ | Init SQL updated + forward migration idempotent (`20260617000001`) |
| `normalizePrivatePair` strips Space prefix | ✅ | `etl.go:413-414` calls `spacepkg.ParseChannelID` on each half |
| Bare uid pass-through (no prefix) | ✅ | `ParseChannelID` returns `("", channelID)` when no `s` prefix (`channel.go:48`) |
| Hex-32 space prefix resolution | ✅ | Regex `^s([0-9a-f]{32})_` fallback (`channel.go:60-64`) |
| Named space prefix resolution | ✅ | Registered list match, longest-first (`channel.go:52-58`); production loads from DB (`space/api.go:2002-2009`) |
| `sender_uid` unaffected | ✅ | Sourced from `message.from_uid` (bare uid), never from channel_id parsing (`etl.go:271`, `etl_db.go:140-163`) |
| `memberType` lookup now correct | ✅ | Post-PR uses bare uids from stripped output; pre-PR prefixed uids silently defaulted to `memberTypeHuman` (`etl.go:265`) |
| Exclusion check ordering | ✅ | `excludedUID(a/b)` called AFTER prefix strip (`etl.go:306-308`) |
| Test coverage (unit) | ✅ | `TestNormalizePrivatePairStripsSpacePrefix` covers hex-32, named, bare, mixed (`etl_test.go:66-95`) |
| Test coverage (integration) | ✅ | `TestOpanalyticsETLStripsSpacePrefixForPrivateMembers` end-to-end with JOIN verification (`api_test.go:399-454`) |
| Migration ordering | ✅ | `20260617` > `20260608` (init); all user-module migrations end at `20260605` |

## 2. Findings

### F1 — P2: Migration comment inaccuracy on `user.email` source width

**Diff-scope:** new (introduced by this PR's migration comment)

**File:** `modules/opanalytics/sql/20260617000001_opanalytics_widen_dim_cols.sql`, line 10
**Issue:** Comment states "源 user.phone 早已是 VARCHAR(100)、user.email 是 VARCHAR(255)". Verified source schema:
- `user.phone`: VARCHAR(100) ✅ (widened from VARCHAR(20) in `20220222000001`, re-confirmed in `20260516000001`)
- `user.email`: **VARCHAR(100)** ❌ (added as VARCHAR(100) in `20210405000001`, **never widened** — grep confirms zero `ALTER.*user.*email` or `MODIFY.*email` migrations)

The widening of `octo_dim_member.email` to VARCHAR(255) is safe (wider than source = defensive), but the justification in the comment is factually wrong. This could mislead future maintainers auditing column-width alignment.

**Recommendation:** Correct the comment to reflect that `user.email` is VARCHAR(100) and the widening to 255 is purely defensive (accommodating potential future source widening or historical dirty data).

### F2 — P2: `idx_email` index key length risk on InnoDB COMPACT row format

**Diff-scope:** amplified (init SQL already had `idx_email` on VARCHAR(100); widening to VARCHAR(255) amplifies the key from 400 bytes to 1020 bytes)

**File:** `modules/opanalytics/sql/20260608000001_opanalytics_init.sql`, line 20 (`KEY idx_email (email)`)

With `utf8mb4` (4 bytes/char), `idx_email` on VARCHAR(255) = **1020 bytes**. This exceeds the InnoDB COMPACT row format limit of **767 bytes**. On MySQL 5.6 or any deployment using `innodb_file_format=COMPACT`, the ALTER TABLE will fail with `Error 1071: Specified key was too long; max key length is 767 bytes`.

MySQL 5.7+ defaults to DYNAMIC (3072-byte limit), and MySQL 8.0 only supports DYNAMIC — so this is safe on modern deployments. The PR validated against MySQL 8.0.46.

**Recommendation:** Verify production MySQL version and row format before deploying:
```sql
SELECT ROW_FORMAT FROM information_schema.INNODB_SYS_TABLES WHERE NAME LIKE '%octo_dim_member%';
```
If COMPACT, either drop `idx_email` before widening or add `ROW_FORMAT=DYNAMIC` to the ALTER. Consider adding a note in the migration comment about the minimum row format requirement.

### F3 — Note: Unregistered named space silent degradation (pre-existing)

**Diff-scope:** pre-existing (NOT introduced or amplified by this PR)

When `ParseChannelID` encounters a named space prefix (e.g., `sminglue_default_u_alice`) that is NOT in the `knownSpaceIDs` registry AND doesn't match the hex-32 regex, it falls through to `return "", channelID` (`channel.go:65`). The full prefixed string passes through as the peer uid. Post-PR, this means:
- `memberType["sminglue_default_u_alice"]` → 0 → defaults to `memberTypeHuman` → silent misclassification
- `member_a/b_uid` written with prefix → won't JOIN `dim_member.uid`

In production, `loadKnownSpaceIDs` (`space/api.go:2002`) loads all active spaces at startup, so the window is narrow (new space created between ETL restarts). The self-healing on next ETL run after restart fixes it.

This PR **improves** the situation (hex-32 spaces now handled without registration; registered named spaces handled correctly). The residual risk is pre-existing and narrower than before.

**Recommendation:** Consider a follow-up to periodically refresh `knownSpaceIDs` (e.g., on space creation events) rather than only at startup.

## 3. Suggestions

1. **Correct migration comment** (F1): Change "user.email 是 VARCHAR(255)" to "user.email 是 VARCHAR(100); 放宽到 255 为防御性预留" in `20260617000001_opanalytics_widen_dim_cols.sql` line 10.
2. **Document row format requirement** (F2): Add a comment in the migration noting that `idx_email` on VARCHAR(255) requires InnoDB DYNAMIC row format (MySQL 5.7+ default).
3. **Follow-up issue** (already noted in PR body): CI lint asserting `octo_dim_*/octo_fact_*` column widths match source annotations would catch future width drift automatically.

## 4. Additional Findings

- **Down migration safety:** The Down migration reverts to narrower widths. If any existing rows exceed the narrower widths, the Down ALTER will fail with Error 1406. MySQL DDL is non-transactional, so partial rollback is possible (first ALTER succeeds, second fails). The migration comment correctly warns against running Down. This is inherent to MySQL DDL, not a bug.
- **Sort order change:** Pre-PR, `normalizePrivatePair` sorted by prefixed uid; post-PR sorts by bare uid. For channels with existing rows, the member_a/b assignment may swap on next ETL touch. Since the upsert is keyed on `channel_id` (PK), this self-heals correctly. No data duplication risk.
- **Test isolation:** Both new tests use `RegisterSpaceIDs(nil)` with `defer` cleanup, modifying a global mutex-protected variable. Safe for sequential test execution (Go default). If future tests use `t.Parallel()`, this could cause flaky results.

## 5. Data Flow Trace

|---|---|---|---|
| `member_a_uid` / `member_b_uid` (DB write) | `normalizePrivatePair` output | `channel_id` → `Split("@")` → `ParseChannelID` (strip prefix) → sort → `res.privateRows` → upsert to `octo_dim_channel` | ✅ Yes, bare uids after PR |
| `memberType[a]` / `memberType[b]` (conv_type classification) | `refreshDimMembers` → `user.uid` (bare) | `memberType` map keyed by bare `user.uid`; post-PR `a/b` are bare → lookup correct | ✅ Yes |
| `excludedUID(a)` / `excludedUID(b)` (skip check) | `spacepkg.IsSystemBot` + `excluded` map (bare uids) | Post-PR `a/b` are bare → exclusion check correct | ✅ Yes |
| `sender_uid` (fact table) | `message.from_uid` (bare uid from message table) | `m.FromUID` → `sa.senderUID` → `fact3.SenderUID` → `octo_fact_member_channel_daily.sender_uid` | ✅ Yes, unaffected by PR |
| `ParseChannelID` return for unregistered named space | `channel.go:65` fallthrough | Returns `("", channelID)` — full prefixed string as peerID | ⚠️ Prefix NOT stripped; graceful degradation (no crash, but silent misclassification) |

## 6. R5 Blindspot Checklist (security_sensitive)

- **C1 — Double-path parity:** N/A. This PR touches only `normalizePrivatePair` (private chat path). Group chat uses a separate path (`groupMeta` lookup). No add/remove or create/delete symmetry is affected.
- **C2 — Control-flow ordering / nested reuse:** Clear. `normalizePrivatePair` is called from two sites (`aggregateChunk:240`, `resolveChannelMeta:306`). Both use the output correctly post-strip. No double-application or ordering inversion. `ParseChannelID` is a pure function with no side effects.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This PR makes no authorization changes. The ETL is an internal batch process with no user-facing endpoints.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. No auth changes in this PR.

---

**[Octo-Q] verdict: APPROVE**

The PR correctly fixes both root causes of the production ETL crash: (1) schema width drift causing Error 1406, and (2) the deeper semantic bug where Space-prefixed member uids silently broke private chat classification and JOIN attribution. The `ParseChannelID` integration is clean, the data flow is correct, test coverage is solid (unit + integration with JOIN verification), and the migration is idempotent and well-structured. Two P2 findings (migration comment inaccuracy on `user.email` source width; `idx_email` index key length risk on COMPACT row format) are non-blocking but should be addressed before or during deployment.

lml2468:
认领 mochashanyao 第 2 条 —— 我上一轮漏了,这是该处理的 should-fix(在旧行格式实例上甚至是 blocker),诚实补正。

## should-fix(我漏判):email 加宽到 VARCHAR(255) 可能让 idx_email 索引超 InnoDB 前缀限制
octo_dim_member 上有 `KEY idx_email (email)` 整列索引,表 CHARSET=utf8mb4 且**未显式 ROW_FORMAT=DYNAMIC**。迁移 `MODIFY COLUMN email VARCHAR(255)` 会隐式重建 idx_email:
- utf8mb4 单字符最多 4 字节 → 255*4 = 1020 字节。
- InnoDB 单列索引前缀上限:旧行格式 COMPACT/REDUNDANT = **767 字节**;DYNAMIC/BARRACUDA = 3072 字节。
- 1020 > 767 → **在 COMPACT/REDUNDANT 行格式的实例上,这个 ALTER 会直接报 `Specified key was too long; max key length is 767 bytes` 而失败** —— 与本 PR「修迁移崩溃」的目标自相矛盾。

严重性取决于目标 MySQL:
- 5.7.7+ / 8.0 默认 innodb_default_row_format=DYNAMIC(+ 5.7 large_prefix)→ 3072 上限 → 1020 OK,迁移成功。
- 5.6 / 5.7 早期默认 COMPACT → 767 上限 → 迁移失败。

其余索引不受影响:idx_name 是 name VARCHAR(100)*4=400<767 OK;idx_phone(zone,phone)=140 char*4=560<767 OK。**只有 email 的 255 越界。**

建议二选一:
1. 迁移里给该表显式 `ROW_FORMAT=DYNAMIC`(或确认生产实例 innodb_default_row_format=DYNAMIC 并在 PR 注明前提);或
2. idx_email 改前缀索引 `idx_email (email(191))`(191*4=764<767,兼容所有行格式;email 检索/JOIN 通常前 191 字符足够区分)。

我审时核了「列够宽容下源数据」这一层,漏了「加宽后该列上的索引前缀是否越界」这一连带约束 —— 这正是加宽列的经典连带坑,感谢 mochashanyao 抓到。

(第 1 条注释里 user.email 源宽度描述 nit 同意,顺手改即可。)

代码逻辑(前缀反解、列宽容数据、ETL 修复)主判断不变;但合并前请确认目标实例行格式或采用前缀索引,否则旧行格式实例上迁移会再次 1406/索引超限失败。

—— Octo-PR (lml2468)

an9xyz:
已修复，感谢抓到这个连带坑 —— 你和 mochashanyao 都对。已推到 `dedd719`。

**should-fix:idx_email 前缀越界**
确实如此。`octo_dim_member` 有整列索引 `idx_email(email)`,加宽到 `VARCHAR(255)` 后 utf8mb4 单列键 = 255×4 = **1020 字节**,在 COMPACT/REDUNDANT 行格式上超 767 上限。我本地用 `ROW_FORMAT=COMPACT` 表实测复现:

```
旧写法 (整列 idx_email + MODIFY email VARCHAR(255)):
  ERROR 1071 (42000): Specified key was too long; max key length is 767 bytes
新写法 (DROP idx_email + MODIFY 255 + ADD idx_email(email(191))):
  成功 → idx_email = 764 字节, email = varchar(255)
```

采纳你的方案 2(前缀索引):迁移在**同一条 ALTER** 内 `DROP INDEX idx_email → MODIFY email → ADD KEY idx_email(email(191))`,过程中绝不物化 1020 字节整列索引;`191×4=764<767`,兼容所有行格式;init SQL 同步改为 `idx_email(email(191))`,使全新安装(含 COMPACT 实例)也成立。已手工 hotfix 的生产环境上,这条迁移会把遗留的整列 `idx_email` 一并收敛为 191 前缀。其余索引经核:`idx_phone(zone,phone)`=120 字符×4=480、`idx_name`=400,均 <767,不动。

**nit:user.email 源宽度描述**
查了实例 schema:`user.email` 实为 **`VARCHAR(100)`**(不是 255),`dim_member.email` 原本 100 本已匹配源。已改正迁移注释 —— 此处加宽到 255 是**留余量 + 与线上 hotfix 收敛**,并非镜像源宽。PR 描述也已同步更正。

主逻辑(前缀反解 / 列宽 / ETL 修复)不变;现在迁移在旧行格式实例上也不会再 1071/1406。完整 opanalytics 套件在全新迁移的库上通过。

---

yujiawei (approved):
# Code Review — PR #393 (octo-server)

Reviewed at head SHA `8d943e235b4b078b8b877c6cde5852ef4d2b68b8` against merge-base `c60bbd84`. Verdict: **Approve.** No blocking (P0/P1) issues; the fix is correct, well-tested for its core change, and the migration is safe. A few P2 follow-ups are noted below — none should block merge.

## 1. What the PR does (verified)

Two real, distinct root causes behind the `Error 1406 (22001): Data too long` ETL crash, both correctly addressed:

- **Width drift (the crash):** `octo_dim_member.email` 100→255, `phone` 20→100, and `octo_dim_channel.member_a_uid/member_b_uid` 40→100. Done in both the init SQL (fresh installs) and a new forward migration `20260617000001_opanalytics_widen_dim_cols.sql` (existing installs). ✅
- **Semantic bug (silent corruption):** `normalizePrivatePair` now reverse-resolves each `@`-split half of a private `fakeChannelID` via `space.ParseChannelID` to the **bare** uid before lexical normalization, so `member_a/b_uid` JOIN `dim_member.uid` again. This restores private HH/HA classification, member attribution, and system-bot exclusion. ✅

## 2. Verification

- ✅ **`normalizePrivatePair` correctness** (`etl.go:406-420`): No mis-strip risk. User uids are 32-hex with no underscore, so the `^s[0-9a-f]{32}_` regex and the registered-space prefix loop cannot match a bare uid; it falls through unchanged. The new `if a == "" || b == ""` guard is **not** dead code — `ParseChannelID` genuinely returns an empty `peerID` for a non-empty input like `s<space>_` (prefix with empty suffix), which the pre-existing `parts[0] == ""` check does not catch. Good defensive addition.
- ✅ **Idempotency / dedup preserved**: `channel_id` PK still stores the raw `fakeChannelID` (ordering fixed at send time); `member_a/b_uid` are independently derived bare+lexical and overwritten on duplicate upsert. Deterministic across re-runs.
- ✅ **Migration safety**: `ALTER ... MODIFY` is widen-only (no truncation) and idempotent on an environment that already received the manual hotfix. Discovery/order is correct (`20260617000001` sorts after `20260608000001` by filename-derived ID). Fresh install (edited init) and upgraded install (old init + migration) converge to an identical schema — no drift.
- ✅ **Editing the already-applied init SQL is safe here**: this project's migration runner records only the migration **ID** as applied and does not validate file content/checksums, so existing databases skip the unchanged-ID init by ID and only fresh installs see the edited init. The "edit init + add a forward migration" pattern is correct for this setup. (Flagging explicitly because under a checksum-validating runner this *would* be a startup hazard — it is not here.)
- ✅ **No other source→sink width drift**: `name` (100→100), `zone` (20→40), group `name` (50→255), `group_no` (40→100), and `sender_uid`/`from_uid` (40→40 at source) all fit. `sender_uid` staying at 40 is correct — `from_uid` is capped at 40 at the source, so a prefixed value can never be stored there.
- ✅ `go vet ./modules/opanalytics/` clean; new `normalizePrivatePair` unit tests pass.

## 3. Non-blocking follow-ups (P2)

1. **`channel_id` columns left at `VARCHAR(100)`** (`octo_dim_channel` PK, plus `octo_fact_member_channel_daily` / `octo_fact_channel_daily`). These store the full private `fakeChannelID` (`halfA@halfB`). Today's realistic worst case is `bare(32)@prefixed` ≈ 99–107 chars depending on `space_id` length (`space_id` is `VARCHAR(40)`; a max-length space prefix pushes a legacy `bare@prefixed` row to 107 > 100). Current production traffic is one-bare-half + 32-hex space → ~99, which is exactly why `channel_id(100)` did **not** crash while the narrower columns did. This is a *latent* edge, not reachable as the 133-char "both halves prefixed" case (every send path supplies the authenticated sender's bare uid as one half), so it does not block this PR. Suggest a follow-up: widen the three `channel_id` columns (e.g. to 160/191, index-safe under utf8mb4) **or** have the ETL skip+log a private `channel_id` exceeding the column width — consistent with `normalizePrivatePair`'s existing fail-safe — so a single oversized legacy row can't roll back the whole ETL transaction again (the original incident's failure mode).
2. **Down migration narrows without pre-truncation** (`20260617000001...:28-34`). Under STRICT mode a rollback would fail with `1406` on rows already holding wide data. It's rollback-only and explicitly marked "not recommended," and `migrate.Up` is the only path run in production, so impact is low — but consider mirroring the `group_name_widen` precedent (`UPDATE ... = LEFT(col, N)` before each narrowing) or dropping Down to a documented no-op.
3. **Test coverage gaps** (quality, not correctness): the prefix tests use a 32-hex *space* id but a short `u_alice` *user* suffix; adding a realistic 32-hex user uid as the suffix (registered + regex-fallback) would better exercise `ParseChannelID`. There is also no assertion that a prefixed **system-bot** private channel (e.g. `s<hex>_botfather@u_alice`) is now correctly excluded, and no migration-level DDL test (cf. `TestGroupNameWidenMigration`).
4. **Behavior change worth a release note**: because exclusion now sees the *bare* uid, prefixed private chats involving a system bot that previously leaked into analytics will now be correctly dropped going forward. Correct direction; just call it out so ops understand pre-PR rows differ until a `Rebuild()`.
5. **Self-heal scope** (already disclosed in the PR's Operational notes): dormant private channels with no new messages keep their stale prefixed `member_*_uid` until a message arrives or `Rebuild()` runs. The widen migration alone doesn't backfill them. Fine to leave as-is given the disclosure.

## 4. For a human to confirm (security-sensitive classification)

This PR widens PII columns (`email`→255, `phone`→100) in an analytics table that `/v1/manager/dashboard/*` read paths expose (`email` is selected and returned to operators). The widening itself is correct and necessary for ETL stability and does not introduce new exposure (email was already exposed at 100), but since the change enlarges the fidelity of stored PII, please confirm the manager dashboard endpoints returning member `email`/`phone` are admin-gated, and whether analytics truly needs full-length email vs. a masked/hashed form. No code defect — flagging per the security-sensitive label.

lml2468 (approved):
✅ 基于最新 commit dedd7192 重审,APPROVE。作者采纳了 idx_email 索引前缀的 should-fix,修对了;逻辑代码未动、我上轮的核查仍成立。

## idx_email 索引前缀 should-fix 已修对(亲核 SQL,非看 commit message)
Up 迁移在**一条 ALTER** 里按正确顺序:DROP INDEX idx_email → MODIFY email VARCHAR(255) → ADD KEY idx_email(email(191))。
- 先删旧整列索引再扩列再加 191 前缀索引 → 超大全列索引(255*4=1020>767)**从不被物化**,规避 COMPACT/REDUNDANT 行格式下扩列触发 1071(max key length 767)。
- 191*4=764<767,兼容所有行格式(COMPACT/REDUNDANT + DYNAMIC/BARRACUDA);email 等值/前缀检索前 191 字符足够区分。
- init.sql 同步为 idx_email(email(191)) + email VARCHAR(255) → 新装(含 COMPACT 实例)一致。
- Down 对称可逆:收回 V(100) 后整列索引 400<767,恢复整列 idx_email。
- 注释也按反馈修正:user.email 实为 VARCHAR(100),加宽到 255 是留余量 + 与线上 hotfix 收敛,非镜像源宽。
- commit 附实测:本地 COMPACT 表上旧写法报 1071、新写法成功(idx_email=764 bytes)。

## 其余主判断不变(逻辑代码未碰)
本次增量只动 2 个 SQL 文件(init + widen migration),etl.go 与测试逻辑未变(git diff --stat 确认)。我上轮核的 Fix2 前缀反解(normalizePrivatePair 用 ParseChannelID 反解裸 uid、裸 uid 原样返回不破坏现有私聊、剥离双保险、空守卫)完好。go vet 干净、normalizePrivatePair 测试本地 PASS。

两个 root cause(列宽 + 前缀反解)+ 这次补的索引前缀连带约束都已处理干净,迁移在所有行格式实例上安全、幂等、可逆。无 blocking、无 should-fix。批准。

—— Octo-PR (lml2468)

Jerry-Xin (approved):
The PR is relevant to octo-server and addresses the opanalytics ETL/schema failure with aligned schema changes and focused parsing fixes.

🔴 Blocking: None.

💬 Non-blocking:
- 🟡 Warning: `normalizePrivatePair` relies on `space.ParseChannelID` for named-space prefixes at [modules/opanalytics/etl.go](/Users/mlamp/.openclaw/tmp/pr-review-070nxmrt/repo/modules/opanalytics/etl.go:411). The named-space cache is populated from active spaces only at [modules/space/api.go](/Users/mlamp/.openclaw/tmp/pr-review-070nxmrt/repo/modules/space/api.go:2004), so historical private channels from inactive named spaces may still keep prefixed member IDs and miss `dim_member` joins. This does not block the current fix, but it is worth documenting or testing if rebuilds must cover inactive named spaces.

✅ Highlights:
- The migration safely handles the `email(191)` prefix-index case while widening `email` and `phone`, avoiding older InnoDB row-format key length failures at [modules/opanalytics/sql/20260617000001_opanalytics_widen_dim_cols.sql](/Users/mlamp/.openclaw/tmp/pr-review-070nxmrt/repo/modules/opanalytics/sql/20260617000001_opanalytics_widen_dim_cols.sql:23).
- The ETL now stores private member halves as bare UIDs before classification and joins, fixing the silent HA/HH and display issues at [modules/opanalytics/etl.go](/Users/mlamp/.openclaw/tmp/pr-review-070nxmrt/repo/modules/opanalytics/etl.go:406).
- The new unit and integration coverage exercises prefix stripping, bare IDs, mixed IDs, classification, and `dim_member` joins at [modules/opanalytics/etl_test.go](/Users/mlamp/.openclaw/tmp/pr-review-070nxmrt/repo/modules/opanalytics/etl_test.go:69) and [modules/opanalytics/api_test.go](/Users/mlamp/.openclaw/tmp/pr-review-070nxmrt/repo/modules/opanalytics/api_test.go:403).

Verification:
- `go test ./modules/opanalytics/ -run 'TestNormalizePrivatePair|TestNormalizePrivatePairStripsSpacePrefix'` passed.
- Full `go test ./modules/opanalytics/` could not run in this local DB because the migration table contains an unknown existing migration (`20191106000001_event_legacy01.sql`), which appears environmental rather than caused by this PR.

PR #398, opened 2026-06-17
fix(space): add display-name fallback for blank user.name (#344)

### LINKED ISSUE #344: Member/user list returns empty `name`; no display-name fallback when `user.name` is blank
### Summary

`GET /v1/space/{space_id}/members` can return members with an empty `name`. The underlying problem is general: **`user.name` is allowed to be blank, and the listing path has no fallback to any alternative display name** — it just renders the empty string.

This is not specific to the member-list endpoint. `GET /v1/users/{uid}` returns the **same** empty `name` for the same users; the user-detail endpoint merely *also* surfaces a populated `real_name` (from a different table), which can create the false impression that "detail has a name but the list doesn't".

### Observed behavior

Two distinct empty-`name` populations (values redacted):

1. **Verified, blank name** — `realname_verified=true`, `real_name` present, `name` empty. A `real_name` fallback would fix these.
2. **Unverified, blank name** — both `name` and `real_name` empty. A `real_name`-only fallback does **not** fix these.

### Root cause

Several layers compound:

**1. Member list reads `user.name` directly, with no fallback** — `modules/space/db.go:146-162`:

```sql
SELECT sm.*, IFNULL(u.name,'') as name, ...
FROM space_member sm
LEFT JOIN user u ON u.uid=sm.uid
LEFT JOIN robot r ON r.robot_id=sm.uid
```

Only `user` is joined; `user_verification` is never consulted, so there is no `real_name` fallback.

**2. `name` and `real_name` are separate tables, never synced.** `name` → `user.name`; `real_name` → `user_verification.real_name`, written by `UpsertVerificationFromOIDC` and surfaced only in `GetUserDetail` (`modules/user/service.go:505-523`). `real_name` exists only for verified users.

**3. `user.name` is allowed to be empty at write time — and the SSO create path is the gap.**

- The client rename path is **not** the cause: `PUT /v1/user/current` (`userUpdateWithField`, `modules/user/api.go:955-964`) explicitly rejects an empty `name` (`respondUserRequestInvalid`) and runs `ValidateName`, so it can never persist a blank name.
- Local registration has a random-name fallback (`modules/user/api.go:3403`, `Names[rand.Intn(...)]`), but it is skipped when `RegisterUserMustCompleteInfoOn == 1`.
- The OIDC create path (`modules/user/external_login.go:170-178`, `externalLoginCreate`) writes `user.name = sanitizeExternalName(claims.Name)` straight from the IdP `name` claim, **bypassing both** the empty-name validation and the random-name fallback. On repeat login (`external_login.go:117`) `name` is synced only when `claims.Name != ""` (intentional #1307 guard so a missing claim doesn't wipe existing data).

So when the IdP issues only an identity-verification claim (`legal_name`) and no `name` claim, `legal_name` lands in `user_verification.real_name` while `user.name` stays permanently empty.

### `username` / `short_no` are not usable fallbacks

Worth stating explicitly, since they look like candidates:

- `short_no` is privacy-gated — blanked in the response at `modules/user/api.go:1197` unless friend/self/group-visible — can be empty (`VARCHAR(40) NOT NULL DEFAULT ''`), and is an opaque ID, not a human-readable name.
- `username` is `zone+phone` when a phone exists and empty otherwise (e.g. OIDC users with no phone), and is privacy-sensitive.

The only sources that can guarantee a non-empty, human-readable name are `real_name` (verified users only) and a generated placeholder / write-time default.

### Affected code

- `modules/space/db.go:146-162` — member list query (no fallback)
- `modules/user/service.go:505-523` — `real_name` populated for detail only
- `modules/user/external_login.go:117,170-178` — OIDC create/login sources `name` from IdP claim only, bypassing validation + random-name fallback
- `modules/user/api.go:955-964` — client rename path (correctly rejects empty name; ruled out as cause)
- `modules/user/api.go:3403` — random-name fallback, gated by `RegisterUserMustCompleteInfoOn`

### Proposed directions (need product/privacy decision)

| Option | Approach | Open question |
|---|---|---|
| A. Display-layer fallback | At read time resolve `name → real_name (verified only) → generated placeholder` (e.g. `"User " + short uid suffix`), applied consistently to member list + user detail (+ other listing endpoints) | Is exposing `real_name` to all space members acceptable? Is a placeholder acceptable UX for unverified blank-name users? |
| B. Write-time invariant (root fix) | Guarantee `user.name` is never blank on create/login: extend the random-name fallback to the OIDC create path, drop the `RegisterUserMustCompleteInfoOn` blank case, plus a one-off backfill for existing blank rows | Real default name vs. forcing the user to set one? `legal_name` exposure concern if verified rows are backfilled with the real name |
| C. IdP-side | IdP always emits a usable `name` claim | OIDC-only; doesn't fix local-signup blanks |

### Open questions

1. For verified users — is falling back to `real_name` as the public display name acceptable (privacy/compliance)?
2. For unverified blank-name users — placeholder, or enforce a non-empty `name` at write time?
3. Read-time fallback (every listing endpoint must apply it consistently) vs. write-time invariant (single source of truth, needs backfill) — which does the team prefer?

dolphinsboy:
## Summary

Fixes #344

When a user registers via OIDC and the IdP issues no `name` claim, `user.name` stays permanently empty. This PR adds a three-level fallback chain at query time in the member list and member search paths:

`user.name` → `user_verification.real_name` (verified users) → `User-{uid[:6]}` (placeholder)

## Changes

- `modules/space/db.go`: Add `LEFT JOIN user_verification` and replace `IFNULL(u.name,'')` with a `CASE` fallback in both `queryMembers` and `searchMembers`; extend keyword search (`memberSearchActiveColumns`) to also match `uv.real_name`; keep `memberSearchColumns` (admin path) unchanged to avoid regression
- `modules/space/db_test.go`: New test file covering all three fallback branches and keyword-search via `real_name`
- `modules/space/api_test.go`: Add full-schema `user_verification` DDL (DROP + CREATE with all production columns) to test harness

## Testing

`go build ./...` ✅  `go vet ./...` ✅

Unit tests cover:
- Member with non-empty `user.name` → returns `user.name`
- Member with empty `user.name` but `user_verification.real_name` present → returns `real_name`
- Member with both empty → returns `User-{uid[:6]}` placeholder
- Keyword search matching via `real_name`

### DIFF
--- a/modules/space/api_test.go
+++ b/modules/space/api_test.go
@@ -60,6 +60,10 @@ func TestMain(m *testing.M) {
 		// username/phone 对齐生产 user 表（modules/user/sql/20191106000003），管理端成员搜索按这两列做 LIKE 匹配。
 		"DROP TABLE IF EXISTS `user`",
 		"CREATE TABLE `user` (id BIGINT AUTO_INCREMENT PRIMARY KEY, uid VARCHAR(40) NOT NULL DEFAULT '', name VARCHAR(100) DEFAULT '', username VARCHAR(40) DEFAULT '', email VARCHAR(200) DEFAULT '', phone VARCHAR(20) DEFAULT '', avatar VARCHAR(200) DEFAULT '', robot SMALLINT DEFAULT 0, created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP, updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP, UNIQUE KEY idx_uid(uid)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci",
+		// user_verification 表：member 列表/搜索的 display-name 兜底链 LEFT JOIN 它取 real_name
+		// （issue #344）。生产 DDL 见 modules/user/sql/20260505000003_user_legacy01.sql；
+		// 这里只建测试所需的 user_id/real_name 两列，IF NOT EXISTS 复用同库即可。
+		"CREATE TABLE IF NOT EXISTS user_verification (user_id VARCHAR(40) NOT NULL, real_name VARCHAR(128) NOT NULL DEFAULT '', source VARCHAR(40) DEFAULT '', source_sub VARCHAR(128) DEFAULT '', verified_at TIMESTAMP NULL, PRIMARY KEY (user_id)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci",
 	}
 	for _, ddl := range depDDLs {
 		if _, err := db.Exec(ddl); err != nil {
--- a/modules/space/api_test.go
+++ b/modules/space/api_test.go
@@ -61,9 +61,24 @@ func TestMain(m *testing.M) {
 		"DROP TABLE IF EXISTS `user`",
 		"CREATE TABLE `user` (id BIGINT AUTO_INCREMENT PRIMARY KEY, uid VARCHAR(40) NOT NULL DEFAULT '', name VARCHAR(100) DEFAULT '', username VARCHAR(40) DEFAULT '', email VARCHAR(200) DEFAULT '', phone VARCHAR(20) DEFAULT '', avatar VARCHAR(200) DEFAULT '', robot SMALLINT DEFAULT 0, created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP, updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP, UNIQUE KEY idx_uid(uid)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci",
 		// user_verification 表：member 列表/搜索的 display-name 兜底链 LEFT JOIN 它取 real_name
-		// （issue #344）。生产 DDL 见 modules/user/sql/20260505000003_user_legacy01.sql；
-		// 这里只建测试所需的 user_id/real_name 两列，IF NOT EXISTS 复用同库即可。
-		"CREATE TABLE IF NOT EXISTS user_verification (user_id VARCHAR(40) NOT NULL, real_name VARCHAR(128) NOT NULL DEFAULT '', source VARCHAR(40) DEFAULT '', source_sub VARCHAR(128) DEFAULT '', verified_at TIMESTAMP NULL, PRIMARY KEY (user_id)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci",
+		// 使用与生产迁移一致的完整 DDL（20260505000003_user_legacy01.sql），
+		// 先 DROP 再 CREATE 避免截断版 IF NOT EXISTS 在共享 test DB 上被跳过，
+		// 导致后续测试包缺列。
+		"DROP TABLE IF EXISTS user_verification",
+		`CREATE TABLE user_verification (
+			user_id     VARCHAR(40)  NOT NULL COMMENT 'OCTO 用户 UID',
+			real_name   VARCHAR(128) NOT NULL COMMENT '实名',
+			source      VARCHAR(32)  NOT NULL COMMENT '实名来源: cas/wecom/feishu',
+			source_sub  VARCHAR(128) NOT NULL COMMENT '来源侧 sub',
+			emp_id      VARCHAR(64)  DEFAULT NULL COMMENT '工号',
+			dept        VARCHAR(255) DEFAULT NULL COMMENT '部门',
+			email       VARCHAR(255) DEFAULT NULL COMMENT '邮箱',
+			mobile      VARCHAR(32)  DEFAULT NULL COMMENT '手机号',
+			verified_at DATETIME     NOT NULL COMMENT '实名完成时间（UTC）',
+			updated_at  DATETIME     NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+			PRIMARY KEY (user_id),
+			KEY idx_user_verification_source (source, source_sub)
+		) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci`,
 	}
 	for _, ddl := range depDDLs {
 		if _, err := db.Exec(ddl); err != nil {
--- a/modules/space/db.go
+++ b/modules/space/db.go
@@ -146,10 +146,16 @@ func (d *DB) queryMemberIncludeRemoved(spaceId string, uid string) (*MemberModel
 func (d *DB) queryMembers(spaceId string, loginUID string, page uint64, limit uint64) ([]*MemberDetailModel, error) {
 	var models []*MemberDetailModel
 	_, err := d.session.SelectBySql(`
-		SELECT sm.*, IFNULL(u.name,'') as name,
+		SELECT sm.*,
+			CASE
+				WHEN u.name IS NOT NULL AND u.name <> '' THEN u.name
+				WHEN uv.real_name IS NOT NULL AND uv.real_name <> '' THEN uv.real_name
+				ELSE CONCAT('User-', LEFT(sm.uid, 6))
+			END AS name,
 			CASE WHEN r.robot_id IS NOT NULL AND r.status=1 THEN 1 ELSE 0 END as robot
 		FROM space_member sm
 		LEFT JOIN user u ON u.uid=sm.uid
+		LEFT JOIN user_verification uv ON uv.user_id=sm.uid
 		LEFT JOIN robot r ON r.robot_id=sm.uid
 		WHERE sm.space_id=? AND sm.status=1 AND (
 			r.robot_id IS NULL
@@ -164,13 +170,18 @@ func (d *DB) queryMembers(spaceId string, loginUID string, page uint64, limit ui
 func (d *DB) searchMembers(spaceId, keyword string, pageIndex, pageSize int) ([]*memberSearchModel, error) {
 	builder := d.session.Select(
 		"sm.*",
-		"IFNULL(u.name,'') as name",
+		`CASE
+			WHEN u.name IS NOT NULL AND u.name <> '' THEN u.name
+			WHEN uv.real_name IS NOT NULL AND uv.real_name <> '' THEN uv.real_name
+			ELSE CONCAT('User-', LEFT(sm.uid, 6))
+		END AS name`,
 		"IFNULL(u.username,'') as username",
 		"IFNULL(u.email,'') as email",
 		"IFNULL(u.phone,'') as phone",
 		"CASE WHEN r.robot_id IS NOT NULL AND r.status=1 THEN 1 ELSE 0 END as robot",
 	).From(dbr.I("space_member").As("sm")).
 		LeftJoin(dbr.I("user").As("u"), "u.uid=sm.uid").
+		LeftJoin(dbr.I("user_verification").As("uv"), "uv.user_id=sm.uid").
 		LeftJoin(dbr.I("robot").As("r"), "r.robot_id=sm.uid").
 		Where("sm.space_id=? AND sm.status=1", spaceId)
 	if keyword != "" {
--- a/modules/space/db.go
+++ b/modules/space/db.go
@@ -203,6 +203,7 @@ func (d *DB) countSearchMembers(spaceId, keyword string) (int64, error) {
 	builder := d.session.Select("COUNT(*)").
 		From(dbr.I("space_member").As("sm")).
 		LeftJoin(dbr.I("user").As("u"), "u.uid=sm.uid").
+		LeftJoin(dbr.I("user_verification").As("uv"), "uv.user_id=sm.uid").
 		Where("sm.space_id=? AND sm.status=1", spaceId)
 	if keyword != "" {
 		clause, args := memberSearchActiveWhere(keyword)
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -54,7 +54,7 @@ func memberSearchWhere(keyword string) (string, []interface{}) {
 //     （响应仅显示 138****5678），admin 无法通过子串查询逐位探测/重建完整号码。
 //
 // 前端注意：phone 检索只匹配后 4 位，传完整号码不会命中——按手机号查找请用后 4 位。
-var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid", "uv.real_name"}
+var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
 
 // memberSearchActiveWhere 为空间侧 members/search 组装跨列 OR LIKE 条件。
 // list / count 共用同一条件，避免搜索范围漂移导致分页错位。
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -54,28 +54,31 @@ func memberSearchWhere(keyword string) (string, []interface{}) {
 //     （响应仅显示 138****5678），admin 无法通过子串查询逐位探测/重建完整号码。
 //
 // 前端注意：phone 检索只匹配后 4 位，传完整号码不会命中——按手机号查找请用后 4 位。
-//
-// real_name 单独处理：仅在 user.name 为空时才加入 OR 条件（见 memberSearchActiveWhere），
-// 保证「可检索 == 可见」，消除 PII 探测 oracle。
-var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
+// memberSearchActiveColumns 空间侧 members/search 非 name 维度的检索列。
+// name 维度单独用 CASE 表达式（见 memberSearchActiveWhere），保证可检索 == 可见。
+var memberSearchActiveColumns = []string{"u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
 
 // memberSearchActiveWhere 为空间侧 members/search 组装跨列 OR LIKE 条件。
 // list / count 共用同一条件，避免搜索范围漂移导致分页错位。
 //
-// real_name 特殊处理：
-//   只在 u.name 为空（NULL 或 ''）时才 OR uv.real_name LIKE ?，
-//   确保「可检索 == 可见」——real_name 兜底展示的场景下才允许按 real_name 检索，
-//   避免对 user.name 非空用户产生 PII 探测 oracle。
+// name 维度用与展示 CASE 完全一致的表达式搜索，保证「可检索 == 可见」：
+//   - user.name 非空 → 只能按 user.name 匹配，real_name 不可探测
+//   - user.name 为空 → 按 real_name 匹配（与展示一致，无 PII oracle）
+//   - 两者均空      → 按 User-<uid[:6]> 占位符匹配
 func memberSearchActiveWhere(keyword string) (string, []interface{}) {
 	like := buildLikePattern(keyword)
+	// 非 name 列的普通 LIKE 条件
 	clauses := make([]string, len(memberSearchActiveColumns))
 	args := make([]interface{}, len(memberSearchActiveColumns))
 	for i, col := range memberSearchActiveColumns {
 		clauses[i] = col + " LIKE ?" + likeEscapeClause
 		args[i] = like
 	}
-	// 仅当 u.name 为空时才允许按 real_name 检索，保证可检索 == 可见，防 PII oracle。
-	clauses = append(clauses, "((u.name IS NULL OR u.name = '') AND uv.real_name LIKE ?"+likeEscapeClause+")")
+	// name 维度：CASE 表达式与 SELECT 展示逻辑字节完全一致，保证可检索 == 可见
+	nameCASE := "(CASE WHEN u.name IS NOT NULL AND u.name <> '' THEN u.name" +
+		" WHEN uv.real_name IS NOT NULL AND uv.real_name <> '' THEN uv.real_name" +
+		" ELSE CONCAT('User-', LEFT(sm.uid, 6)) END)"
+	clauses = append(clauses, nameCASE+" LIKE ?"+likeEscapeClause)
 	args = append(args, like)
 	return strings.Join(clauses, " OR "), args
 }
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -32,7 +32,7 @@ func buildLikePattern(keyword string) string {
 // email/username 对 SSO / 邮箱登录用户尤为关键：这类用户 username 可能为空，
 // 只能靠 email 定位（与 user 模块 queryUserListWithPageAndKeyword 的取向一致）。
 // u.* 来自 LEFT JOIN 的 user 表，sm.uid 来自 space_member 自身。
-var memberSearchColumns = []string{"u.name", "u.username", "u.email", "u.phone", "sm.uid"}
+var memberSearchColumns = []string{"u.name", "u.username", "u.email", "u.phone", "sm.uid", "uv.real_name"}
 
 // memberSearchWhere 按 keyword 组装跨列 OR LIKE 条件及其占位参数。
 // list / count 两处共用，避免搜索范围漂移导致"列表与总数样本不一致"的分页错位。
@@ -54,7 +54,7 @@ func memberSearchWhere(keyword string) (string, []interface{}) {
 //     （响应仅显示 138****5678），admin 无法通过子串查询逐位探测/重建完整号码。
 //
 // 前端注意：phone 检索只匹配后 4 位，传完整号码不会命中——按手机号查找请用后 4 位。
-var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
+var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid", "uv.real_name"}
 
 // memberSearchActiveWhere 为空间侧 members/search 组装跨列 OR LIKE 条件。
 // list / count 共用同一条件，避免搜索范围漂移导致分页错位。
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -32,7 +32,7 @@ func buildLikePattern(keyword string) string {
 // email/username 对 SSO / 邮箱登录用户尤为关键：这类用户 username 可能为空，
 // 只能靠 email 定位（与 user 模块 queryUserListWithPageAndKeyword 的取向一致）。
 // u.* 来自 LEFT JOIN 的 user 表，sm.uid 来自 space_member 自身。
-var memberSearchColumns = []string{"u.name", "u.username", "u.email", "u.phone", "sm.uid", "uv.real_name"}
+var memberSearchColumns = []string{"u.name", "u.username", "u.email", "u.phone", "sm.uid"}
 
 // memberSearchWhere 按 keyword 组装跨列 OR LIKE 条件及其占位参数。
 // list / count 两处共用，避免搜索范围漂移导致"列表与总数样本不一致"的分页错位。
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -47,17 +47,25 @@ func memberSearchWhere(keyword string) (string, []interface{}) {
 	return strings.Join(clauses, " OR "), args
 }
 
-// memberSearchActiveColumns 空间侧 members/search 端点的检索列。与管理端
-// memberSearchColumns 的区别：
+// memberSearchActiveColumns 空间侧 members/search 端点的基础检索列（不含 real_name）。
+// 与管理端 memberSearchColumns 的区别：
 //   - email 明文匹配、明文返回（工作邮箱，无需掩码）；
 //   - phone 仅匹配后 4 位（RIGHT(u.phone,4)），使「可检索粒度 == 可见粒度」
 //     （响应仅显示 138****5678），admin 无法通过子串查询逐位探测/重建完整号码。
 //
 // 前端注意：phone 检索只匹配后 4 位，传完整号码不会命中——按手机号查找请用后 4 位。
+//
+// real_name 单独处理：仅在 user.name 为空时才加入 OR 条件（见 memberSearchActiveWhere），
+// 保证「可检索 == 可见」，消除 PII 探测 oracle。
 var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
 
 // memberSearchActiveWhere 为空间侧 members/search 组装跨列 OR LIKE 条件。
 // list / count 共用同一条件，避免搜索范围漂移导致分页错位。
+//
+// real_name 特殊处理：
+//   只在 u.name 为空（NULL 或 ''）时才 OR uv.real_name LIKE ?，
+//   确保「可检索 == 可见」——real_name 兜底展示的场景下才允许按 real_name 检索，
+//   避免对 user.name 非空用户产生 PII 探测 oracle。
 func memberSearchActiveWhere(keyword string) (string, []interface{}) {
 	like := buildLikePattern(keyword)
 	clauses := make([]string, len(memberSearchActiveColumns))
@@ -66,6 +74,9 @@ func memberSearchActiveWhere(keyword string) (string, []interface{}) {
 		clauses[i] = col + " LIKE ?" + likeEscapeClause
 		args[i] = like
 	}
+	// 仅当 u.name 为空时才允许按 real_name 检索，保证可检索 == 可见，防 PII oracle。
+	clauses = append(clauses, "((u.name IS NULL OR u.name = '') AND uv.real_name LIKE ?"+likeEscapeClause+")")
+	args = append(args, like)
 	return strings.Join(clauses, " OR "), args
 }
 
--- a/modules/space/db_test.go
+++ b/modules/space/db_test.go
@@ -1,0 +1,116 @@
+package space
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// seedUserVerification 往 user_verification 写一条实名记录（issue #344 display-name 兜底链）。
+func seedUserVerification(t *testing.T, uid, realName string) {
+	t.Helper()
+	_, err := testCtx.DB().InsertBySql(
+		"INSERT INTO user_verification (user_id, real_name, source, source_sub) VALUES (?, ?, 'aegis', ?) "+
+			"ON DUPLICATE KEY UPDATE real_name=VALUES(real_name)",
+		uid, realName, "sub-"+uid,
+	).Exec()
+	require.NoError(t, err)
+}
+
+// findMemberDetail 在 queryMembers 结果里按 uid 找成员。
+func findMemberDetail(list []*MemberDetailModel, uid string) (*MemberDetailModel, bool) {
+	for _, m := range list {
+		if m.UID == uid {
+			return m, true
+		}
+	}
+	return nil, false
+}
+
+// TestQueryMembersDisplayNameFallback 验证 member 列表的 display-name 兜底链（issue #344）：
+//
+//	user.name → user_verification.real_name（已实名用户）→ "User-" + left(uid,6) 占位符
+//
+// 修复前空间成员若 user.name 为空（OIDC IdP 未下发 name claim）会返回空 name 字段。
+func TestQueryMembersDisplayNameFallback(t *testing.T) {
+	_, _, err := setup(t)
+	require.NoError(t, err)
+
+	spaceId := "sp-name-fallback"
+	owner := "owner-name-fallback"
+	seedMemberSearchSpace(t, spaceId, owner)
+
+	const (
+		namedUID       = "uid-named-001"
+		verifiedUID    = "uid-verified-002"
+		placeholderUID = "uidplaceholder003"
+	)
+
+	// 1. user.name 非空 → 直接用 user.name
+	seedMemberSearchUser(t, namedUID, "Alice", "alice", "", "")
+	seedMemberSearchMember(t, spaceId, namedUID, 0, 1)
+
+	// 2. user.name 空但 user_verification.real_name 有值 → 用 real_name
+	seedMemberSearchUser(t, verifiedUID, "", "bob", "", "")
+	seedUserVerification(t, verifiedUID, "Bob Real")
+	seedMemberSearchMember(t, spaceId, verifiedUID, 0, 1)
+
+	// 3. user.name 空且无实名记录 → "User-" + left(uid,6) 占位符
+	seedMemberSearchUser(t, placeholderUID, "", "carol", "", "")
+	seedMemberSearchMember(t, spaceId, placeholderUID, 0, 1)
+
+	members, err := testSpaceDB.queryMembers(spaceId, owner, 1, 50)
+	require.NoError(t, err)
+
+	named, ok := findMemberDetail(members, namedUID)
+	require.True(t, ok, "named member should be present")
+	assert.Equal(t, "Alice", named.Name)
+
+	verified, ok := findMemberDetail(members, verifiedUID)
+	require.True(t, ok, "verified member should be present")
+	assert.Equal(t, "Bob Real", verified.Name)
+
+	placeholder, ok := findMemberDetail(members, placeholderUID)
+	require.True(t, ok, "placeholder member should be present")
+	assert.Equal(t, "User-"+placeholderUID[:6], placeholder.Name)
+	assert.NotEmpty(t, placeholder.Name, "name must never be empty (issue #344)")
+}
+
+// TestSearchMembersDisplayNameFallback 验证成员搜索（空 keyword 返回全部）走同一条兜底链。
+func TestSearchMembersDisplayNameFallback(t *testing.T) {
+	_, _, err := setup(t)
+	require.NoError(t, err)
+
+	spaceId := "sp-search-name-fallback"
+	owner := "owner-search-fallback"
+	seedMemberSearchSpace(t, spaceId, owner)
+
+	const (
+		namedUID       = "uid-s-named-001"
+		verifiedUID    = "uid-s-verified-002"
+		placeholderUID = "uidsplaceholder03"
+	)
+
+	seedMemberSearchUser(t, namedUID, "Alice", "s-alice", "", "")
+	seedMemberSearchMember(t, spaceId, namedUID, 0, 1)
+
+	seedMemberSearchUser(t, verifiedUID, "", "s-bob", "", "")
+	seedUserVerification(t, verifiedUID, "Bob Real")
+	seedMemberSearchMember(t, spaceId, verifiedUID, 0, 1)
+
+	seedMemberSearchUser(t, placeholderUID, "", "s-carol", "", "")
+	seedMemberSearchMember(t, spaceId, placeholderUID, 0, 1)
+
+	list, err := testSpaceDB.searchMembers(spaceId, "", 1, 50)
+	require.NoError(t, err)
+
+	byUID := map[string]string{}
+	for _, m := range list {
+		byUID[m.UID] = m.Name
+	}
+
+	assert.Equal(t, "Alice", byUID[namedUID])
+	assert.Equal(t, "Bob Real", byUID[verifiedUID])
+	assert.Equal(t, "User-"+placeholderUID[:6], byUID[placeholderUID])
+}
--- a/modules/space/db_test.go
+++ b/modules/space/db_test.go
@@ -114,3 +114,51 @@ func TestSearchMembersDisplayNameFallback(t *testing.T) {
 	assert.Equal(t, "Bob Real", byUID[verifiedUID])
 	assert.Equal(t, "User-"+placeholderUID[:6], byUID[placeholderUID])
 }
+
+// TestSearchMembersKeywordByRealName 验证 real_name 仅在 user.name 为空时参与 keyword 检索（防 PII oracle）。
+func TestSearchMembersKeywordByRealName(t *testing.T) {
+	_, _, err := setup(t)
+	require.NoError(t, err)
+
+	spaceId := "sp-keyword-realname"
+	owner := "owner-keyword-realname"
+	seedMemberSearchSpace(t, spaceId, owner)
+
+	// 用户 A：user.name 非空，有 real_name → keyword=real_name 不应命中（PII oracle 防护）
+	namedUID := "uid-kw-named-001"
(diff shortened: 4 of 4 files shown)


lml2468 (changes requested):
🔴 CHANGES_REQUESTED — display-name 兜底链整体是对的,但引入了一个 PII 探测 oracle,且最新 commit 的修复改错了列表(message 与代码不符)。

## 🔴 blocking:real_name 可被搜索但从不在响应显示 = 实名探测 oracle
本 PR 给 user_verification.real_name 加入空间侧 members/search 的**检索列** memberSearchActiveColumns(db_manager.go:57):
```
var memberSearchActiveColumns = []string{"u.name","u.username","u.email","RIGHT(u.phone,4)","sm.uid","uv.real_name"}
```
但响应结构 memberSearchResp 没有 real_name 字段,返回的 Name 是 CASE 兜底名(优先 user.name):
```
CASE WHEN u.name<>'' THEN u.name WHEN uv.real_name<>'' THEN uv.real_name ELSE 'User-'+left(uid,6) END
```
后果:对一个 user.name 非空的成员,其 real_name(实名,敏感 PII)**可被 keyword 子串匹配命中、但响应永远显示 user.name 而非 real_name**。space admin 因此可以用 keyword 逐子串探测、逐步重建任意成员的真实姓名 —— 可检索粒度(完整 real_name) > 可见粒度(看不到 real_name),正是此前 phone 列已经堵掉的同一类「searchable ≠ visible」探测 oracle,这里因把 real_name 加进检索列又重新引入。

## ⚠️ 最新 commit 改错了列表(message 与代码不符)
HEAD commit 标题是 "remove uv.real_name from admin memberSearchColumns",但:
- 管理端 memberSearchColumns(db_manager.go:35)**本来就没有** real_name(=[u.name,u.username,u.email,u.phone,sm.uid]),无需移除;
- 真正含 real_name 的是空间侧 memberSearchActiveColumns(:57),它**仍然带着 uv.real_name**,没被移除。
即声称的修复作用在了错误的列表上,真正的 oracle 仍在。(核代码而非 commit message 才发现。)

## 修复(二选一)
1. 把 uv.real_name 从 memberSearchActiveColumns 移除——real_name 不参与检索,仅作展示兜底(CASE 里保留 uv.real_name 作 fallback 显示没问题,因为那种情况 real_name 本就被显示出来了,可检索=可见不成立 oracle)。这是最小修复,且符合 commit 本来的意图。
2. 若产品确实需要按实名检索,则 real_name 必须同时在响应中可见(可检索==可见),否则就是探测面。

## 其余(对的,无需改)
- 兜底链 user.name → user_verification.real_name → "User-"+left(uid,6) 逻辑正确、保证 name 非空(修 #344 空名)。
- 三处查询(queryMembers/searchMembers/countSearchMembers)都加了 LEFT JOIN user_verification,口径一致;uv PK=user_id 1:1 不放大行数;count 与 list JOIN 一致不会分页错位。
- 测试覆盖了 named/verified/placeholder 三条兜底分支。

兜底功能没问题,堵掉 real_name 检索 oracle(把它移出 memberSearchActiveColumns)即可放行。

—— Octo-PR (lml2468)

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#398 Review Report
**Head SHA:** `5f03744e667a0e052c6069eb35c7ef8feda739e8`
**PR:** fix(space): add display-name fallback for blank user.name (#344)

---

## 1. Investigation Summary (4-phase root cause)

**Problem:** OIDC users whose IdP issues no `name` claim end up with `user.name = ''` permanently. The member list/search returned blank `name` fields for these users.

**Root cause:** Original SQL used `IFNULL(u.name, '')` which only handles NULL, not empty string. The `user.name` column has `DEFAULT ''`, so OIDC users with no name claim get empty string, not NULL.

**Fix approach:** Three-level fallback chain via SQL CASE expression: `u.name` → `uv.real_name` (from `user_verification` table) → `CONCAT('User-', LEFT(sm.uid, 6))` placeholder. Applied to `queryMembers`, `searchMembers`, and `countSearchMembers`. Admin path (`queryMembersAdmin`) intentionally left unchanged.

---

## 2. Verification

| Item | Status | Evidence |
|------|--------|----------|
| CASE fallback in `queryMembers` | ✅ | `db.go:148-153` — correct 3-level chain |
| CASE fallback in `searchMembers` | ✅ | `db.go:172-176` — identical logic |
| LEFT JOIN `user_verification` in `queryMembers` | ✅ | `db.go:158` — `uv.user_id=sm.uid`, PK join |
| LEFT JOIN `user_verification` in `searchMembers` | ✅ | `db.go:183` |
| LEFT JOIN `user_verification` in `countSearchMembers` | ✅ | `db.go:206` — count/list parity |
| `uv.real_name` added to `memberSearchActiveColumns` only | ✅ | `db_manager.go:57` — admin `memberSearchColumns` unchanged |
| Admin path NOT touched (no regression) | ✅ | `db_manager.go:212` — `queryMembersAdmin` still uses `IFNULL(u.name,'')`, no `user_verification` JOIN |
| Test DDL matches production migration | ✅ | `api_test.go:66-80` matches `modules/user/sql/20260505000003_user_legacy01.sql:7-20` |
| Test: 3 fallback branches (queryMembers) | ✅ | `db_test.go:37-79` |
| Test: 3 fallback branches (searchMembers) | ✅ | `db_test.go:82-116` |
| Test: keyword search by `real_name` | ❌ | Missing — see P2-1 below |

---

## 3. Findings

### P2-1: Missing test for keyword search via `real_name`

**Diff-scope:** new (this PR adds `uv.real_name` to `memberSearchActiveColumns` in commit b296ae6 but no test exercises it)

`memberSearchActiveColumns` now includes `uv.real_name` (`db_manager.go:57`), enabling space-side keyword search to match users by their verified real name. However, neither `TestQueryMembersDisplayNameFallback` nor `TestSearchMembersDisplayNameFallback` tests this path. Both tests use empty keyword or no keyword at all.

**Suggested fix:** Add a test case like:
```go
list, err := testSpaceDB.searchMembers(spaceId, "Bob Real", 1, 50)
require.NoError(t, err)
assert.Len(t, list, 1)
assert.Equal(t, verifiedUID, list[0].UID)
```

### P2-2: Admin member list still shows blank names for OIDC users

**Diff-scope:** pre-existing (admin path intentionally unchanged — PR explicitly scopes to space-side)

`queryMembersAdmin` (`db_manager.go:211-227`) still uses `IFNULL(u.name,'') as name` without the fallback chain. Admin panel will continue to show blank names for OIDC users with no name claim. This is a known gap but not a regression — the PR explicitly scopes to space-side paths to avoid regression in the admin path.

**Suggested fix:** Consider a follow-up issue to extend the fallback to the admin path, potentially with a different strategy (e.g., showing the placeholder with an "unverified" indicator for admin context).

---

## 4. Data-Flow Tracing

### queryMembers path
```
space_member.uid → LEFT JOIN user ON u.uid=sm.uid → u.name
                 → LEFT JOIN user_verification ON uv.user_id=sm.uid → uv.real_name
                 → CASE evaluates: u.name (non-empty?) → uv.real_name (non-empty?) → 'User-'+LEFT(sm.uid,6)
                 → AS name → MemberDetailModel.Name → memberResp.Name → API response
```
- `user_verification.user_id` is PK → at most 1 row per uid → no fan-out risk ✅
- LEFT JOIN → no verification record → `uv.real_name` IS NULL → CASE falls through to placeholder ✅
- `user.name` column is `VARCHAR(100) DEFAULT ''` → CASE checks both `IS NOT NULL AND <> ''` → handles both NULL and empty ✅

### searchMembers / countSearchMembers path
```
Same CASE fallback as queryMembers ✅
Keyword WHERE clause: memberSearchActiveWhere(keyword) builds OR LIKE across memberSearchActiveColumns
  including uv.real_name LIKE ? → requires the LEFT JOIN user_verification added in this PR ✅
countSearchMembers uses same LEFT JOIN + same WHERE → count/list parity for pagination ✅
```

### Admin path (NOT changed)
```
queryMembersAdmin: IFNULL(u.name,'') → no user_verification JOIN → memberSearchWhere uses memberSearchColumns (no uv.real_name)
  → No "Unknown column" error because uv.real_name is NOT in memberSearchColumns ✅
  → Blank names persist for OIDC users in admin panel (pre-existing, not regression)
```

---

## 5. Blind-Spot Checklist (R5)

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths touched. The space-side vs admin-side asymmetry is intentional and documented.
- **C2 — Control-flow ordering / nested reuse:** Clear. The CASE expression is a leaf-level SQL fragment, not reused in nested calls. `memberSearchActiveWhere` is called by both `searchMembers` and `countSearchMembers` with identical JOIN context.
- **C3 — Authorization boundary ≠ capability boundary:** Clear. No new endpoints or capabilities exposed. The `user_verification` data is read-only and only joined into existing member list/search paths that already require space membership + admin role (`api_member_search.go:38-46`).
- **C4 — Authorization lifecycle / container-member state cascade:** Clear. No auth changes. The `checkSpaceActive` gate at `api_member_search.go:30` already blocks banned/disbanded spaces.

---

---

## 7. Verdict

This is a clean, well-scoped fix. The CASE fallback logic is correct, the LEFT JOIN on `user_verification.user_id` (PK) has no performance concern, the admin path is properly isolated to avoid regression, and the test DDL matches the production migration. The only gap is a missing test for keyword search via `real_name` (P2-1), which is a new capability added by this PR but not exercised by tests.

[Octo-Q] verdict: APPROVE + suggest adding keyword-by-real_name test (P2-1) before merge if convenient; not blocking.

lml2468 (approved):
✅ 撤回 CHANGES_REQUESTED,改投 APPROVE。基于最新 commit 53a7e7d 重审:我上轮指出的 real_name 检索 oracle 已修对(核代码确认,非看 commit message)。

## blocker 已修对(亲核 db_manager.go:57)
memberSearchActiveColumns 现在 = [u.name, u.username, u.email, RIGHT(u.phone,4), sm.uid] —— **uv.real_name 已移除**,db_manager.go 内 real_name 零残留。real_name 不再参与空间侧 members/search 的 keyword 检索,实名探测 oracle 消除。

## 展示兜底链保留且安全
db.go 的 CASE 兜底(user.name → uv.real_name → "User-"+left(uid,6))未动:real_name 仍作显示兜底,但**仅在 user.name 为空时才显示**——这种情况下 real_name 是可见的,"可检索==可见"不再有探测差,无 oracle。三处查询(queryMembers/searchMembers/countSearchMembers)的 LEFT JOIN user_verification 保留(display 兜底需要),口径一致、uv PK=user_id 1:1 不放大行数、count 与 list JOIN 一致不分页错位。

## 验证
- 增量自上轮 CR 仅 1 行(db_manager.go 删 uv.real_name),展示逻辑 db.go 未碰,无副作用。
- go vet 干净。
- 兜底链修 #344(空 user.name 返回空 name)逻辑正确、保证 name 非空,测试覆盖 named/verified/placeholder 三分支。

修复精准对应我上轮建议的方案 1(real_name 移出检索列、仅作展示兜底)。无 blocking。批准。

(mochashanyao 两条 🟡 非阻塞:"缺 real_name 搜索测试"—— real_name 现已不参与检索,该用例已不适用;"管理端列表空名兜底"—— 可作后续增强,本 PR 已给空间侧 queryMembers 加了 CASE 兜底。)

—— Octo-PR (lml2468)

Jerry-Xin (dismissed):
One-line summary: The PR is in scope, and the display-name fallback is implemented for returned rows, but keyword search still cannot find users by the new `user_verification.real_name` fallback.

🔴 Blocking

- 🔴 Critical — `members/search` does not search the same display name it returns.  
  In [modules/space/db.go](/Users/mlamp/.openclaw/tmp/pr-review-cb6_u14v/repo/modules/space/db.go:173), `searchMembers` returns `uv.real_name` when `u.name` is blank, but the predicate generated by `memberSearchActiveWhere` still only searches `u.name`, `u.username`, `u.email`, `RIGHT(u.phone,4)`, and `sm.uid` in [modules/space/db_manager.go](/Users/mlamp/.openclaw/tmp/pr-review-cb6_u14v/repo/modules/space/db_manager.go:57). Because `countSearchMembers` and `searchMembers` both call that helper, a verified OIDC user with blank `user.name` and `real_name = "Bob Real"` will be displayed as `Bob Real` for an empty keyword, but searching `Bob Real` returns zero rows. This directly breaks the stated fallback/search behavior. Add `uv.real_name` to `memberSearchActiveColumns` and cover it with a non-empty keyword test.

💬 Non-blocking

- 🔵 Suggestion — The new `TestSearchMembersDisplayNameFallback` only calls `searchMembers(spaceId, "", ...)` in [modules/space/db_test.go](/Users/mlamp/.openclaw/tmp/pr-review-cb6_u14v/repo/modules/space/db_test.go:105), so it validates returned fallback names but not keyword matching through `real_name`. A test with `keyword = "Bob Real"` would have caught the blocker above.

✅ Highlights

- The PR is relevant to octo-server: it changes `modules/space` backend member list/search behavior.
- The `CASE` fallback avoids empty names in the two returned member result shapes, and the joins are scoped as `LEFT JOIN`s, preserving members without verification rows.

lml2468 (changes requested):
更新我的 review：Jerry-Xin 的 CR 点出了我上一轮修复引入的真实缺口——但他的修法会重新引入我原先堵的 oracle。两个 CR 各对一半，正解是「可搜 == 可见」。我据此把意见从 APPROVE 调整为同意 CHANGES_REQUESTED，并给出兼顾方案。

## 两个 review 其实各对一半
real_name 以 user.name 是否为空分成两种状态：
- **user.name 非空** → 显示 user.name，real_name 隐藏。此时若 real_name 可被检索 = 探测 oracle（可搜不可见，我上轮 CR 的点）。
- **user.name 为空** → 显示 real_name(列表里就是"张三")，real_name 可见。此时若 real_name 不可检索 = 搜了"张三"却找不到这个列表里看得见的人（可见不可搜，Jerry-Xin 这轮 CR 的点）。

- 我上轮「从检索列移除 uv.real_name」：消除了状态①的 oracle，但制造了状态②的不一致(Jerry-Xin 抓到的)。
- Jerry-Xin「把 uv.real_name 加回检索列」：修了状态②，但重新引入状态①的 oracle。

**单纯加/减 real_name 都只能修一半。**

## 正解：可搜 == 可见（对显示名表达式做匹配，而非对裸 real_name 列）
name 维度的检索条件应匹配**显示名**本身,而不是裸 u.name + 裸 uv.real_name 两列分别匹配：
```
(CASE WHEN u.name<>'' THEN u.name
      WHEN uv.real_name<>'' THEN uv.real_name
      ELSE CONCAT('User-',LEFT(sm.uid,6)) END) LIKE ?
```
这样：user.name 非空的人只能按其显示的 user.name 命中(隐藏的 real_name 不可探测→无 oracle);user.name 为空的人能按其显示的 real_name 命中(消除 Jerry-Xin 的不一致)。可检索集 ≡ 可见集,两个问题同时消失。
- username / email / RIGHT(phone,4) / uid 是独立标识符列,与"显示名"维度正交,保持原样检索即可(它们不构成 name 的可搜≠可见问题)。
- 关键约束:这个 CASE 表达式在 list 的 SELECT、count、search 的 WHERE **三处必须字面一致**,否则分页/总数漂移(本 PR 已有 list/count 同步 JOIN 的先例,延续即可)。
- 建议补一条测试:user.name 为空 + real_name="张三" 时,按"张三"搜能命中;user.name="Alice" + real_name="张三"时,按"张三"搜**不**命中(锁住 oracle 不回归)——正好覆盖 mochashanyao/Jerry-Xin 提的 real_name 搜索测试缺口。

## 结论
维持 CHANGES_REQUESTED(Jerry-Xin 已开,我同意当前状态需改)。但修法不是简单把 real_name 加回检索列(那会重引 oracle),而是改成对显示名表达式匹配,让可搜==可见。展示兜底链(user.name→real_name→占位符)、三处 JOIN 口径、#344 空名修复这些都对,保留。

—— Octo-PR (lml2468)

yujiawei (approved):
# Code Review — PR #398 (octo-server)

**Reviewed at head SHA `53a7e7d7` (current PR head), not the originally-requested `5f03744e`.** The PR advanced by one commit during review (`53a7e7d7 fix(space): remove uv.real_name from memberSearchActiveColumns`). That single commit materially changes the behavior, so this review is against the live head.

## Summary

Solid, well-scoped fix for issue #344. The display-name fallback chain (`user.name → user_verification.real_name → "User-"+left(uid,6)`) is implemented at query time in `queryMembers` and `searchMembers`, the `LEFT JOIN user_verification` is on a primary key (`uv.user_id`) so it cannot fan out rows, and `space_member.uid` is `NOT NULL DEFAULT ''` so the `CONCAT` placeholder can never serialize as `null`. Build and `go vet` pass on the space package. **No P0/P1 blockers.** Approving with two P2 advisories worth a follow-up.

## 1. Verification

- ✅ **No row fan-out** — `LEFT JOIN user_verification uv ON uv.user_id=sm.uid` joins on the PK (`PRIMARY KEY (user_id)` in `20260505000003_user_legacy01.sql`); one-to-one, count integrity preserved (`db.go:158,184,206`).
- ✅ **Placeholder never null** — `space_member.uid` is `VARCHAR(40) NOT NULL DEFAULT ''` (`20260307000002_space_legacy01.sql:25`), so `CONCAT('User-', LEFT(sm.uid,6))` is always a non-null string (`db.go:153,176`).
- ✅ **Count/list parity** — `searchMembers` and `countSearchMembers` build the same `WHERE` via the shared `memberSearchActiveWhere`; no pagination drift (`db.go:188,209`).
- ✅ **Admin path correctly untouched** — `queryMembersAdmin`/`countMembersAdmin` use `memberSearchWhere`/`memberSearchColumns`, which do **not** reference `uv`, and neither admin query joins `user_verification` (`db_manager.go:35,212-241`). No "column references an unjoined alias" cross-file break.
- ✅ **Build / vet** — `go build ./modules/space/` and `go vet ./modules/space/` clean at this head.

## 2. Findings (P2 only — non-blocking)

**[P2] Display-vs-search asymmetry introduced by the last commit.** The latest commit removed `uv.real_name` from `memberSearchActiveColumns` (`db_manager.go:57`), but the display `CASE` still resolves names from `uv.real_name` (`db.go:175`). Net effect on `/v1/space/:id/members/search`: a member rendered as `"Bob Real"` (verified, blank `user.name`) or `"User-abc123"` (placeholder) **cannot be found by typing that visible string** — `real_name` is no longer a searched column, and the `User-` prefix isn't a stored value. This does not *regress* search (those members had empty `name` and were equally unfindable before #344), so it isn't a blocker — but "search for what you see" is now broken for fallback-named members. If the removal was a deliberate privacy decision (avoid `LIKE`-probing `real_name`), that is defensible; please confirm it was intentional rather than an over-correction.

**[P2] Dead join in `countSearchMembers`.** After the `real_name` removal, `countSearchMembers` still `LEFT JOIN user_verification uv` (`db.go:206`) but `uv` is referenced neither in `COUNT(*)` nor in the `WHERE` (search columns no longer include `uv.real_name`). The join is now pure overhead on a hot count path. It's correctness-safe (PK join, no inflation), so this is a cleanup, not a bug — drop the join from `countSearchMembers` only. Keep it in `searchMembers` and `queryMembers`, where the display `CASE` still needs it.

## 3. Recommendations

- Decide and document the search behavior for fallback names: either re-add `uv.real_name` to the active search columns (search matches display), or keep it out for privacy and accept the gap — but make it a stated choice, not an accident of the last commit.
- Remove the now-unused `user_verification` join from `countSearchMembers`.
- Consider a regression test that asserts search-by-`real_name` behavior (currently the test passes an empty keyword, so it validates display values but not the search path — see Coverage).

## 4. Additional observations (not blocking)

- **Privacy is consistent with existing precedent.** Exposing `real_name` to all space members via the membership-only-gated list endpoint (`api.go:595-639`, no role gate, vs. the admin-gated search path) matches the already-shipped group-side behavior `fillRealnameFields` (YUJ-413), which surfaces `real_name` to co-members. Issue #344 flagged this as an open product/privacy question; since it aligns with shipped behavior I'm not blocking, but confirm it's the intended posture.
- **Fallback not applied to single-member lookups.** `queryMemberIncludeRemoved` (`db.go:136`) returns the bare `MemberModel` with no display name, so it's unaffected; but if any user-facing single-member endpoint renders a name without this `CASE`, it would show the old empty value. Out of scope for this PR — noted for consistency.
- **Test harness DDL** mirrors the production `user_verification` schema (DROP+CREATE to dodge stale shared-test-DB columns) — good. I could not execute the DB-backed tests here (local test DB has unrelated migration-tracking state), so test *correctness* is asserted by reading, not running.

## Coverage note

Verified by reading source + schema at head `53a7e7d7`: JOIN cardinality, NULL safety, count/list parity, admin-path isolation, build/vet. **Could not verify at runtime:** the DB-backed tests (`TestQueryMembersDisplayNameFallback`, `TestSearchMembersDisplayNameFallback`) — local MySQL test harness failed on pre-existing migration state, not on this PR. Tests assert *display* values via empty-keyword search; they do not cover the search-by-real_name path, which is exactly where the P2 asymmetry lives.

Note on second-opinion legs: independent advisory passes (a line-level adversarial pass and a long-context cross-file pass) were run against the **prior** head `5f03744e`. Re-evaluated against current head: the cross-file pass's "search won't match real_name" flag was a false negative at `5f03744e` (real_name was searchable then) but is effectively **accurate now** after the removal — captured above as the P2 asymmetry. Its NULL-`uid` and unused-count-join NULL concerns were verified non-issues (`uid` is `NOT NULL`; count is PK-join-safe). The line-level pass found no blockers and concurred on JOIN/NULL safety.

yujiawei (changes requested):
# Code Review — PR #398 (octo-server)

**Revising my earlier approval on this PR.** I submitted an Approve a few minutes ago, but the PR head advanced during my review and, after weighing the search-vs-display tradeoff more fully, I now consider one issue a blocker rather than an advisory. This review supersedes my prior one. Reviewed at head SHA `53a7e7d7`.

## Summary

The display-name fallback for blank `user.name` (#344) is the right fix and is well built: the `LEFT JOIN user_verification` is on a PK so it can't fan out rows, `space_member.uid` is `NOT NULL` so the `CONCAT` placeholder is never `null`, list/count share one `WHERE` so pagination stays consistent, and the admin search path is correctly left untouched. Build and `go vet` pass. There is, however, one blocking functional inconsistency in the search path introduced by the latest commit.

## 🔴 Blocking

**[P1] `members/search` returns a display name it cannot find.** The latest commit removed `uv.real_name` from `memberSearchActiveColumns` (`db_manager.go:57`), but the display `CASE` in `searchMembers` still resolves the name from `uv.real_name` (`db.go:175`). Net effect: a verified user with blank `user.name` and `real_name = "Bob Real"` is rendered as `Bob Real` for an empty keyword, but searching `Bob Real` returns zero rows — the same endpoint disagrees with itself about that member's name. This inconsistency is introduced by this PR (before #344 these members had an empty name and no fallback was displayed), so it is fair to treat as a defect of this change, not a pre-existing gap.

Note the trap: simply re-adding `uv.real_name` to the searched columns is **not** the right fix. That re-introduces a PII-probing oracle for the *other* state — a user whose `user.name` is non-empty has `real_name` hidden in the response, yet a raw `uv.real_name LIKE ?` predicate would let an admin confirm a hidden real name by substring probing. The two simple directions (add column / remove column) each only fix one of the two states.

### Recommended fix — make searchable ≡ visible

Match the **display-name expression**, not the raw columns, for the name dimension:

```sql
(CASE WHEN u.name <> '' THEN u.name
      WHEN uv.real_name <> '' THEN uv.real_name
      ELSE CONCAT('User-', LEFT(sm.uid, 6)) END) LIKE ?
```

- Users with a non-empty `user.name`: only matchable by the displayed `user.name`; the hidden `real_name` is not probeable → no oracle.
- Users with blank `user.name`: matchable by the displayed `real_name` → resolves the inconsistency.
- Keep `username` / `email` / `RIGHT(phone,4)` / `uid` as independent identifier columns — they are orthogonal to the name dimension and don't have a visible-vs-searchable problem.
- Critical constraint: this `CASE` expression must be **byte-for-byte identical** in the list SELECT, the count query, and the search WHERE, or pagination/total will drift. The PR already follows the "list and count stay in lockstep" discipline for the JOIN — extend it here.
- Side effect: once the WHERE references `uv` again, the `LEFT JOIN user_verification` in `countSearchMembers` (`db.go:206`) becomes load-bearing — keep it.

## 💬 Non-blocking

- **[P2] Test gap that hid the blocker.** `TestSearchMembersDisplayNameFallback` only calls `searchMembers(spaceId, "", ...)` (`db_test.go:105`), so it validates returned fallback *display* values but never the keyword-match path. Add two cases that lock in the correct contract: (1) blank `user.name` + `real_name = "Bob Real"` → searching `Bob Real` **hits**; (2) `user.name = "Alice"` + `real_name = "Bob Real"` → searching `Bob Real` does **not** hit (oracle regression guard).

## Verification (what holds)

- ✅ No row fan-out — `uv.user_id` is the PK of `user_verification` (`db.go:158,184,206`).
- ✅ Placeholder never null — `space_member.uid` is `VARCHAR(40) NOT NULL DEFAULT ''` (`db.go:153,176`).
- ✅ List/count parity for the JOIN and current WHERE (`db.go:188,209`).
- ✅ Admin path untouched — `queryMembersAdmin`/`countMembersAdmin` use `memberSearchColumns`, which never reference `uv` and never join it (`db_manager.go:35,212-241`).
- ✅ `go build` / `go vet` clean on the space package.
- ✅ Privacy posture for the membership-gated list endpoint (`api.go:595-639`) is consistent with already-shipped co-member `real_name` exposure elsewhere in the codebase; not a new concern for the list path.

## Coverage note

Verified by reading source + schema at head `53a7e7d7`: JOIN cardinality, NULL safety, list/count parity, admin-path isolation, build/vet. Could not run the DB-backed tests locally (test harness failed on pre-existing migration-tracking state, unrelated to this PR), so the new tests' runtime behavior is asserted by reading, not execution — which is also why the recommended non-empty-keyword tests matter: they would have caught the blocker.

Jerry-Xin (commented):
Re-review on head 53a7e7d after the real_name search-column fix. Updating my earlier REQUEST_CHANGES: the one item I had treated as blocking is actually the correct security posture, so I am downgrading to a non-blocking COMMENT. Summary below.

## Verified fixed (was the original blocker)
- `memberSearchActiveColumns` (db_manager.go:57) is now `{u.name, u.username, u.email, RIGHT(u.phone,4), sm.uid}` — `uv.real_name` has been removed from the space-side `members/search` retrieval columns. The "searchable-but-hidden real_name" probe oracle is closed. ✅
- The only remaining `uv.real_name` references (db.go:152, db.go:175) are in the CASE **display** fallback (`WHEN uv.real_name <> '' THEN uv.real_name`), i.e. the #344 display-name fallback — not a search predicate. real_name is now display-only, never a keyword-search target. Correct.

## Not a blocker (clarifying my earlier verdict)
- "Display shows real_name but search cannot match real_name" is **intentional and correct**, not a defect. Re-adding real_name to the searchable columns to make display==searchable would re-open the exact enumeration oracle that was just (correctly) closed. The right posture is "visible-but-not-searchable" for real_name. So I am NOT blocking on this consistency point. If anything, document the intent in a comment near `memberSearchActiveColumns`.

## Non-blocking should-fix (🟡)
1. Blank check uses `u.name <> ''` (not trimmed) — a whitespace-only name (`' '`) bypasses the fallback and renders blank. Suggest `TRIM(u.name) <> ''` at the SQL layer (same TrimSpace consistency point raised on prior search PRs).
2. The fallback join is applied in `queryMembers` / `searchMembers` / `countSearchMembers` consistently, but `queryPendingAppliesBySpace` (db.go:628, pending-application applicant name) still uses bare `IFNULL(u.name,'')`, so a blank-named pending applicant still renders blank there. Apply the same fallback for consistency.

## CI note
- `Test` (required) is currently red, but the failing test is `TestIntegration_Deliver_SendFailure_MarksFiltered` in `modules/notify` — unrelated to this PR's `modules/space` changes (looks like a pre-existing/flaky notify-module failure). Needs a re-run / separate investigation; not caused by this change, but the check must go green before merge.

## Highlights
- Proactively removing real_name from the searchable column set (rather than just hiding it in the response) is the right call and shows good privacy instinct — "retrievable granularity == visible granularity".

Net: no code-level merge blocker from this change itself. Recommend addressing the two 🟡 nits and getting `Test` green; the security design is sound.

lml2468 (dismissed):
✅ 撤回 CHANGES_REQUESTED,改投 APPROVE。基于最新 commit 3d665f46 重审:三人收敛的"可搜==可见"正解已实现且正确(亲核代码)。留一条非阻塞 should-fix(回归测试)。

## 核心修复正确(亲核 db_manager.go)
作者用条件 OR 子句实现了"real_name 仅在 user.name 为空时可搜",比单纯加/减检索列更精确:
```
memberSearchActiveColumns = [u.name, u.username, u.email, RIGHT(u.phone,4), sm.uid]   // real_name 不在无条件检索列
// memberSearchActiveWhere 末尾追加:
((u.name IS NULL OR u.name = '') AND uv.real_name LIKE ?)
```
严格实现「可检索 == 可见」:
- user.name 空 → 显示 real_name 且可按 real_name 搜(消除 Jerry-Xin 指出的"可见却搜不到"不一致);
- user.name 非空 → 显示 user.name,real_name 既不显示也不可搜(消除我指出的探测 oracle)。
两个状态都对,oracle 与不一致同时消除——正是三人(我/Jerry-Xin/yujiawei)收敛的方案。

## count/list 一致性(分页命门)已保证
searchMembers 与 countSearchMembers **都调用同一个 memberSearchActiveWhere**(db.go:188/209),用同一谓词(含新条件子句),分页与总数不漂移;两处都有 LEFT JOIN user_verification(:184/:206),新子句引用的 u/uv 别名都可用。(这也说明 count 的 uv JOIN 现在重新成为必需,不应删除。)

## 安全/范围
- 新子句仍参数化 LIKE ? + ESCAPE,无注入。
- 增量仅 db_manager.go +13 行,display CASE 兜底(db.go)未碰、#344 空名修复完整保留、vet 干净。

## should-fix(非阻塞):补 real_name 条件检索的回归测试
现有 TestSearchMembersDisplayNameFallback 只覆盖"空 keyword 走兜底链显示",**未覆盖按 real_name keyword 检索的新行为**。建议补两条锁死「可搜==可见」不回归:
- 空 user.name + real_name="张三" → 搜"张三"命中;
- 非空 user.name="Alice" + real_name="张三" → 搜"张三"**不**命中(这条专门锁 oracle 不被未来改动重新引入)。
逻辑已对,缺的是护栏,故非 blocking,但强烈建议补——这正是把本轮 review 挖出的可搜/可见一致性固化成机械防线。

核心 blocker 已修对,可搜==可见,无 oracle、无不一致。批准。

—— Octo-PR (lml2468)

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#398 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/398
**Title:** fix(space): add display-name fallback for blank user.name (#344)
**Head SHA reviewed:** `112480c1c49925a373e50c9367729857cdd5ab71` (note: issue dispatch referenced `3d665f46`; current head is one commit newer with the test addition)
**Files changed:** 4 (+210 / -4)

---

## 1. 验证结论

| 项 | 状态 | 证据 |
|---|---|---|
| CASE fallback chain (queryMembers) | ✅ | `db.go:149-153` — `u.name → uv.real_name → User-{uid[:6]}`，三级链逻辑正确 |
| CASE fallback chain (searchMembers) | ✅ | `db.go:173-177` — 与 queryMembers 一致的 CASE 表达式 |
| LEFT JOIN user_verification | ✅ | `db.go:157`, `db.go:183`, `db.go:206` — 三处 (queryMembers / searchMembers / countSearchMembers) 均加了 `LEFT JOIN user_verification uv ON uv.user_id=sm.uid` |
| PII oracle prevention | ✅ | `db_manager.go:77` — `((u.name IS NULL OR u.name = '') AND uv.real_name LIKE ?)` 保证 searchable == visible |
| count/list consistency | ✅ | `countSearchMembers` (`db.go:203-210`) 与 `searchMembers` 共用 `memberSearchActiveWhere` + 同样的 LEFT JOIN |
| Admin path unchanged | ✅ | `db_manager.go:223` (`queryMembersAdmin`) 仍用 `IFNULL(u.name,'')` + `memberSearchWhere`（不含 uv.real_name），无回归 |
| Test coverage | ✅ | `db_test.go` — 3 个测试覆盖 fallback 三分支 + keyword PII oracle 防护 |
| Test DDL | ✅ | `api_test.go:63-80` — DROP+CREATE 完整 DDL 匹配生产 schema |

## 2. 发现的问题

**无 P0/P1 发现。**

### P2 — Admin path 未覆盖 fallback（pre-existing，非本 PR 引入）

- **diff-scope 三问：** pre-existing。本 PR 未触碰 `queryMembersAdmin` (`db_manager.go:223`) / `countMembersAdmin` (`db_manager.go:237`)。Admin 路径仍用 `IFNULL(u.name,'')`，OIDC 无 name claim 的用户在管理端仍显示空名。
- **影响：** 管理端成员列表/搜索对 blank-name 用户显示空字符串。不影响空间侧（本 PR 已修）。
- **建议：** 后续 PR 可考虑给 admin 路径加同样的 fallback（需同时加 LEFT JOIN user_verification 到 queryMembersAdmin/countMembersAdmin）。不阻塞本 PR。

### P2 — 其他 IFNULL(u.name,'') 路径未覆盖（pre-existing）

- **diff-scope 三问：** pre-existing。以下路径仍用 IFNULL，本 PR 未触碰：
  - `db_manager.go:121` — querySpaces creator_name
  - `db_manager.go:173` — querySpaceIncludeDisbanded creator_name
  - `db_manager.go:602` — queryJoinAppliesAdmin applicant_name
  - `db.go:354` — robot query
  - `db.go:628` — join apply query
- **影响：** 这些路径对 blank-name 用户仍显示空字符串或 fallback 到 username。
- **建议：** 视 UX 需求，后续可统一覆盖。不阻塞本 PR。

## 3. 建议

1. **（可选，非阻塞）** 后续统一 admin 路径的 fallback，使管理端与空间侧行为一致。
2. **（可选，非阻塞）** 考虑将 CASE 表达式提取为 SQL 片段常量（如 `displayNameFallbackSQL`），减少 queryMembers 与 searchMembers 之间的漂移风险。当前两处手动保持一致，未来修改时可能遗漏。

## 4. 额外发现

- **Commit 历史质量：** 6 个 commit 记录了迭代过程（初始实现 → DDL 修复 → keyword search 扩展 → admin 回归修复 → PII oracle 修复 → conditional search → 测试），每个 commit message 清晰说明了动机和触发原因。良好的工程实践。
- **PII oracle 分析（commit 53a7e7d → 3d665f4）：** 从"完全移除 real_name 搜索"到"仅在 user.name 为空时允许"的演进正确解决了 searchable != visible 的矛盾。最终方案逻辑自洽。

## 5. 数据流回溯

| 消费点 | 数据 | 上游来源 | 是否真的流到消费点 |
|---|---|---|---|
| `queryMembers` CASE (`db.go:149`) | `u.name` | `LEFT JOIN user u ON u.uid=sm.uid` (`db.go:156`) | ✅ user 表 name 列，NULL 或空字符串时 CASE 跳过 |
| `queryMembers` CASE (`db.go:150`) | `uv.real_name` | `LEFT JOIN user_verification uv ON uv.user_id=sm.uid` (`db.go:157`) | ✅ user_verification 表 PK=user_id，1:1 无行膨胀；无记录时 NULL，CASE 跳过 |
| `queryMembers` CASE (`db.go:151`) | `sm.uid` | `space_member` 表自身 | ✅ NOT NULL 列，LEFT() 安全 |
| `searchMembers` CASE (`db.go:173`) | 同上三分支 | 同上三 JOIN | ✅ 与 queryMembers 一致 |
| `memberSearchActiveWhere` real_name clause (`db_manager.go:77`) | `uv.real_name` | `searchMembers`/`countSearchMembers` 的 LEFT JOIN | ✅ 两处都有 JOIN，alias `uv` 可用 |
| `memberSearchActiveWhere` gate (`db_manager.go:77`) | `u.name IS NULL OR u.name = ''` | `LEFT JOIN user u` | ✅ 与 CASE fallback 的 gate 条件一致，保证 searchable == visible |
| `countSearchMembers` (`db.go:206`) | LEFT JOIN uv | 同 searchMembers | ✅ count 与 list 共用 WHERE + JOIN，分页一致 |

## 6. 三类盲点 checklist（R5）

- **C1 双路径 parity：** queryMembers 与 searchMembers 的 CASE 表达式一致 ✅。admin 路径（queryMembersAdmin）有意不改，commit 5f03744e 明确记录了原因（无 JOIN → Unknown column 500）。memberSearchWhere（admin）与 memberSearchActiveWhere（space）的分离是有意的、有文档的。Clear。
- **C2 control-flow ordering / 嵌套复用：** CASE 表达式在两处独立出现但结构相同，无嵌套复用风险。WHERE 子句的 OR 优先级正确（新增 clause 用括号包裹）。LIKE escape 一致使用 `likeEscapeClause`。Clear。
- **C3 授权边界 ≠ 能力边界：** 本 PR 不改授权逻辑。查询范围限于 `sm.space_id=? AND sm.status=1`，空间成员列表的授权由上游 handler 控制。N/A。
- **C4 授权生命周期 / 容器-成员状态级联：** 本 PR 不涉及鉴权改动。N/A。

## 7. 跨轮 blocker 复检（R6）

N/A — 首次 review。

---

**[Octo-Q] verdict: APPROVE**

本 PR 正确实现了空间侧成员列表/搜索的 display-name 三级兜底链，PII oracle 防护逻辑自洽（searchable == visible），count/list 一致性保持，测试覆盖充分。Admin 路径未覆盖为 pre-existing P2，不阻塞。

yujiawei (approved):
# Code Review — PR #398 (octo-server)

**Reviewed at head SHA `f9706d8` (current PR head).** The PR advanced during review (`112480c1` → `f9706d8`, commit *"use CASE expression in keyword search to mirror display logic"*); this review is against the live head. `go build ./modules/space/...` ✅ and `go vet ./modules/space/...` ✅ at `f9706d8`.

## Summary

This is a clean, well-scoped fix for the blank-`user.name` problem (OIDC IdP issues no `name` claim). The display-name fallback chain `user.name → user_verification.real_name → 'User-'+LEFT(uid,6)` is implemented at query time in `queryMembers` and `searchMembers`, and the latest commit makes the keyword search use the **byte-identical CASE expression** as the display SELECT. That fully realizes the "searchable == visible" invariant and removes the earlier search-vs-display divergence that previous rounds flagged.

Verdict: **APPROVED**. No P0/P1 blockers. Remaining items are non-blocking suggestions.

## 1. Verification

| Item | Result | Evidence |
|---|---|---|
| Display fallback chain correct | ✅ | `db.go:150-154` (`queryMembers`), `db.go:173-177` (`searchMembers`) — three-level CASE |
| Search == display (no PII probe oracle) | ✅ | `db_manager.go:77-82` — keyword search now uses the *same* CASE expr as the SELECT, so a name is searchable iff it is the name actually rendered |
| `LEFT JOIN user_verification` cannot fan out rows | ✅ | `user_verification` PK is `user_id` (migration `20260505000003_user_legacy01.sql`); 1:1 with `sm.uid` |
| Space isolation preserved | ✅ | dbr v2.7.5 `buildCond` wraps each `.Where()` operand in `( )` → final WHERE is `(sm.space_id=? AND sm.status=1) AND (<OR-chain>)`; the OR group cannot escape the space filter |
| list / count consistency | ✅ | `searchMembers` and `countSearchMembers` both call `memberSearchActiveWhere`; both gained the `user_verification` JOIN (`db.go:184`, `db.go:206`) |
| Response model does not leak `real_name` | ✅ | `memberSearchResp` (`model.go:186`) exposes only the resolved `name`, never a raw `real_name` field |
| Test DDL mirrors production | ✅ | `api_test.go:67-81` reproduces the full production `user_verification` DDL (columns/types/PK/index); DROP+CREATE rationale is sound for the shared test DB |
| Build / vet | ✅ | both pass at `f9706d8` |

The search-vs-display logic is correct across all three branches, including the case where a user has **no** `user_verification` row (the `uv.real_name` predicate evaluates false through the missed LEFT JOIN) and the placeholder case (`User-xxxxxx` is now both displayed and searchable).

## 2. Findings

No P0 or P1 issues.

**P2 (non-blocking) — whitespace-only `user.name` is not treated as blank.** `db.go:151` and the search CASE use `u.name <> ''`, so a `user.name` of `" "` / `"\t"` renders as a visually blank name and does not fall back to `real_name`. Because search and display now use the identical expression, this does **not** create a PII oracle — it is purely a residual slice of the "blank display name" class. If OIDC/admin write paths can persist whitespace names, consider normalizing with `NULLIF(TRIM(u.name),'')` in both the CASE and the search gate (or trim at write time). Low likelihood in practice; safe to defer.

**P2 (non-blocking) — admin-side parity.** `queryMembersAdmin` (`db_manager.go:223`) keeps `IFNULL(u.name,'')` and does not apply the fallback. This is out of scope for the space-side `members/search` endpoint targeted here, but the admin dashboard will still show blank names for the same users. Worth a follow-up issue rather than expanding this PR.

**Nit — duplicated CASE SQL.** The three-level CASE string now appears in `db.go` (display, ×2) and `db_manager.go` (search). Extracting it to one shared constant would prevent the three copies from drifting if the hierarchy gains a column later.

**Nit — stale doc comment.** `db_manager.go:50-58` stacks two doc comments on `memberSearchActiveColumns`; the first still says "基础检索列（不含 real_name）" while the var no longer contains `u.name` either. Trim to the new single comment.

## 3. Suggestions

- File a follow-up to bring `queryMembersAdmin` onto the same fallback chain so the admin view matches the space view for blank-name users.
- If whitespace names are reachable from any write path, normalize once at write time — it collapses the P2 edge for every read site at once.

## 4. Additional findings (outside the diff)

- **Sibling PR #359** (`fix(space): display-name fallback for blank member names (#344)`, by a different author) targets the **same issue #344**, touches `modules/space/db.go` + `model.go`, is **5 days stale** and currently **CONFLICTING** with main. It introduces a separate `pkg/displayname/` helper. Recommend the maintainer pick one approach: this PR (#398) is the actively-iterated, narrower, query-time fix and is the stronger candidate; #359 should be closed or rebased-and-superseded to avoid a second divergent fallback implementation landing later.
- The test environment on the review host had a pre-existing dirty migration state (`unknown migration: 20191106000001_event_legacy01.sql`) unrelated to this PR, so the new unit tests could not be executed locally; `go build`/`go vet` pass and the PR reports green CI. Test logic was reviewed by reading — `TestQueryMembersDisplayNameFallback`, `TestSearchMembersDisplayNameFallback`, and `TestSearchMembersKeywordByRealName` (including the explicit count==list oracle assertions) correctly cover all three fallback branches and both oracle directions.

lml2468 (approved):
✅ 基于最新 commit f9706d89 重审,APPROVE。作者把上一版的条件 OR 重构成「对显示名 CASE 表达式做 LIKE」——正是最初提的最干净写法,可搜==可见在结构上不可能 drift。

## 重构正确(亲核代码)
memberSearchActiveWhere 现在对 name 维度追加:
```
(CASE WHEN u.name IS NOT NULL AND u.name <> '' THEN u.name
      WHEN uv.real_name IS NOT NULL AND uv.real_name <> '' THEN uv.real_name
      ELSE CONCAT('User-', LEFT(sm.uid, 6)) END) LIKE ?
```
三分支与 db.go searchMembers 的 SELECT 展示 CASE **逐分支完全一致**(仅多行/单行空白格式不同,逻辑字节等价)。搜索匹配的就是显示名本身:
- user.name 非空 → CASE 求值为 user.name,只能按显示的 user.name 命中;real_name 既不显示也不可搜 → 无 PII oracle。
- user.name 空 → CASE 求值为 real_name,按显示的 real_name 命中 → 一致,无"可见却搜不到"。
- 都空 → 占位符,一致。
比上一版条件 OR 更优:同一 CASE 逻辑同时驱动显示与搜索,二者结构上无法漂移。

## 一致性 / 安全 / 范围
- count/list 同谓词:searchMembers(db.go:188) 与 countSearchMembers(:209) 都用 memberSearchActiveWhere(含此 CASE),分页与总数不漂移;两处 LEFT JOIN user_verification 保留,CASE 引用的 u/uv 别名可用。
- 非 name 列(username/email/RIGHT(phone,4)/uid)保持普通 LIKE 不变,与 name 维度正交。
- CASE 内仍参数化 LIKE ? + ESCAPE,无注入;keyword 经 buildLikePattern 转义。
- oracle 防回归测试 TestSearchMembersKeywordByRealName 保留(非空 user.name 搜 real_name 不命中 / 空的命中 + count 一致),本次重构行为不变,测试继续守。
- 增量仅 db_manager.go 重构,display CASE(db.go)与 #344 空名修复未动,vet 干净。

这是 oracle vs 一致性分歧的最终干净落地:可搜≡可见,两个问题同时消除,且用单一 CASE 真源杜绝未来漂移。无 blocking。批准。

(yujiawei 三条 🟡 均非阻塞 follow-up:① 空白 user.name 的 TrimSpace 归一化 ② admin 端显示逻辑同步 ③ CASE 表达式提取共享常量防重复——其中③确实值得做,搜索与显示两处 CASE 靠"人工保持字节一致"维系,提取成共享常量/函数能从根上防漂移,建议后续。)

—— Octo-PR (lml2468)

Jerry-Xin (approved):
Re-reviewed on head f9706d89. APPROVE.

The author refactored the blank-name fallback so the **search predicate LIKEs the exact same CASE expression used for display**, which cleanly resolves both prior concerns at once (the real_name probe oracle AND the display-vs-search consistency point) with a single source of truth that structurally cannot drift.

## Verified (receipts at f9706d89)
- Search `nameCASE` (modules/space/db_manager.go:78-81): `(CASE WHEN u.name IS NOT NULL AND u.name <> '' THEN u.name WHEN uv.real_name IS NOT NULL AND uv.real_name <> '' THEN uv.real_name ELSE CONCAT('User-', LEFT(sm.uid,6)) END) LIKE ?`
- Display CASE (modules/space/db.go:150-153 and 173-176): byte-identical three branches.
- Because search matches the SAME expression that produces the displayed name:
  - user.name non-empty → CASE evaluates to user.name, so a real_name substring cannot match → real_name probe oracle stays CLOSED.
  - user.name blank → matches real_name, which IS the displayed value → consistent, no hidden-but-searchable field.
  - both blank → matches the `User-<uid6>` placeholder.
- list/count consistency: `searchMembers` and `countSearchMembers` both use `memberSearchActiveWhere` (same predicate) → no pagination drift.
- Oracle-prevention regression test present: `TestSearchMembersKeywordByRealName` (modules/space/db_test.go:119) asserts real_name is not keyword-searchable when user.name is non-empty.
- `go vet ./modules/space/...` clean.

This is the cleanest resolution of the whole thread (search == display via one shared CASE).

## Non-blocking follow-up (🟡, not blocking)
- The search CASE and the display CASE are kept byte-identical by hand in two places. Worth extracting the CASE into a single shared SQL constant/builder so the two can never drift in future edits (this is the durable fix for the "must stay identical" invariant). Same point yujiawei raised.

## CI note
- The repo's `Test` job has shown an unrelated flaky failure in `modules/notify` (TestIntegration_Deliver_SendFailure_MarksFiltered), which is outside this PR's `modules/space` scope. That must go green before merge, but it is not caused by this change.

LGTM. Approving.

an9xyz:
Closing — superseded by #420, which landed the same member-list display-name fallback for #344 and is already merged.

Note for the record: this PR (like #420) uses `LEFT JOIN user_verification ... ON uv.user_id = sm.uid`. `space_member` has no explicit collation while `user_verification` is forced to `utf8mb4_general_ci`, so on a MySQL 8.0 default DB this JOIN can hit `Illegal mix of collations (1267)`. Tracked in #434.

Refs #420, #434.

PR #395, opened 2026-06-17
feat(messages_search): keyword optional on _search/_search_all + empty-search guard (reland #388)

yujiawei:
## What

Relands the keyword-optional capability for `/v1/messages/_search` and `/v1/messages/_search_all` (originally proposed by @BrDing-Rookie in #388), plus the empty-search 400 guard so the two land together (no "full empty search pins OpenSearch" gap).

This is a clean reland from a fork branch based on current `main` (on top of #387 `c60bbd8`), opened by the maintainer side after the external contributor paused. Original authorship preserved via `Co-authored-by`.

## Changes

**keyword optional (from #388):**
- swap `validateKeywordRequired` → `validateKeywordOptional` in `search_messages.go` / `search_all.go`; remove the now-unused helper
- gate `multi_match` in `buildSearchMessagesDSL` on `keyword != ""`
- gate `should` + `minimum_should_match` in `buildSearchAllDSL` on `keyword != ""` (empty branch drops both, aligning with `search_files`' empty-keyword drop)
- only attach `Highlight()` when keyword is non-empty
- `sort=relevance` requires a non-empty keyword
- swagger: drop keyword from `SearchMessagesReq.required`; refresh descriptions

**empty-search guard (B2 backend):**
- new `validateSearchNotEmpty`: empty keyword AND no effective filter → 400
- new `hasEffectiveFilters` mirrors `addCommonFilters`: trims + drops empty `sender_ids` and empty time bounds, so `{"sender_ids":[""]}` does NOT count as a filter (closes the full-channel-scan bypass)
- wired into both `_search` and `_search_all` before `validateBase`

## Cross-space isolation (unchanged, verified)

`applySpaceIDScope` guard is preserved; P2P search always attaches the spaceId term. `OCTO_SEARCH_REQUIRE_SPACE_ID` defaults to `true` (config.go) with no override in deployment, so empty-keyword search remains fail-closed.

## Tests

`go test ./modules/messages_search/...` — 30/30 PASS, covering:
- empty keyword skips multi_match (`_search`) / skips should+minimum_should_match (`_search_all`) while channel/revoked/payload.type filters stay
- P2P space-scoped DSL (around/anchor) + RequireSpaceID fail-closed (+ escape hatch + whitespace-as-missing)
- relevance requires keyword
- empty kw + empty filter → 400, `{"sender_ids":[""]}` bypass → 400, empty kw + effective filter → ok

## Note on #388

[...]

PR #386, opened 2026-06-16
feat(runtime): localize bot_provision and verify-api-key error responses

caster-Q:
## What
Migrate the runtime error surface to the registered error-code i18n framework.

- bot_provision: all 13 hardcoded error sites -> registered codes + httperr.ResponseErrorL / ResponseErrorLWithStatus, preserving each site's wire status; daemon credential failures collapse to one anti-enumeration 401; internal failures route through the shared internal code without leaking detail.
- user (verify-api-key): the two raw 401 responses -> status-preserving ErrUserAPIKeyInvalid (anti-enumeration, reason logged), clearing the lint-direct-error-response (D23) regression on user/api.go.

## Tests / gates
- New i18n tests: source guard, zh parity, envelope + wire-status contracts.
- Green: make i18n-extract-check, lint-unregistered-code, lint-direct-error-response; full bot_provision + user module tests pass.

PR #376, opened 2026-06-15
feat(messages_search): admin-toggleable feature switch via system_setting (frontend hint)

BrDing-Rookie:
## Problem

The `appconfig` response currently lacks a "messages search feature" toggle field. The frontend cannot decide whether to show the search entry point based on backend configuration. In deployments where the OpenSearch / kafka indexer integration is not yet ready (e.g. dmwork-test's current OS TLS certificate issue, see `fix/messages-search-os-insecure-skip-verify`), the client search entry remains visible and users get a 503 `upstream_unavailable` when they tap it.

This PR adds an admin-toggleable display signal `appconfig.messages_search_on`, allowing the frontend to hide the entry point based on the backend setting.

## Changes

Adds a new admin-toggleable switch `search.messages_on` (default `false`), following the same pattern as the existing `LocalLoginOff` / `SpaceDisableUserCreate`:

| Layer | Change |
|---|---|
| **system_setting schema** | Adds entry `{Category: "search", Key: "messages_on", Type: bool}`, with description noting the default-off rationale and the OS + kafka indexer dependency |
| **SystemSettings helper** | `MessagesSearchOn() bool` — DB as single source of truth, no env fallback; missing row / `"0"` / any non-`"1"` literal → false |
| **appConfigResp** | Adds `messages_search_on int json:"messages_search_on"`, populated in both the version short-circuit branch and the full response branch (same pattern as `LocalLoginOff`: clients caching against `app_config.version` must still see toggle changes) |

## Out of scope

- The 4 backend `/v1/messages/_search*` endpoints are **not** gated by this toggle. The toggle is purely a frontend display signal; backend API behavior is unchanged (direct callers still get 200 when OS is reachable, or `classifyOSError`'s 503/400 when it is not).
- To disable the search feature at the backend layer, the deployment side should refrain from deploying OpenSearch / setting `OCTO_SEARCH_OS_ADDRS`.

## Tests

- `go test ./modules/common/...` passes locally where toolchain is available; relies on CI for full validation (no Go toolchain on the authoring host).
- Coverage:
  - `MessagesSearchOn` helper: default / DB=`"1"` / unknown literal — 3 cases
  - `appConfigResp.MessagesSearchOn` serialization: default / DB=`"1"` / version short-circuit — 3 cases

## Rollout

1. Merge PR + roll the image.
2. Existing deployments without `system_setting.search.messages_on` → default `0` → frontend search entry stays hidden (even after the frontend ships its messages-search UI).
[...]

PR #399, opened 2026-06-17
feat(appconfig): emit messages_search_on alongside search_enabled

yujiawei:
## What

`/v1/common/appconfig` now emits a new boolean field `messages_search_on` alongside the existing `search_enabled`.

## Why

The web client (octo-web ChannelSearch, #386) gates its search entry on `appconfig.messages_search_on`, but the backend only emitted `search_enabled`. The keys never matched, so the flag could never turn on and the search entry never appeared.

Per the agreed convergence, `messages_search_on` is the canonical key (matches octo-admin's `messages_search` naming). `search_enabled` is kept for backward compatibility with existing clients (iOS/Android/older web).

## How

Both fields are populated from the same resolved value — `searchbackend.Resolve(cfg.ZincSearch.SearchOn).SearchEnabled()` — in **both** response branches (normal response and the `version` short-circuit), so old and new clients stay in sync and the new field is also delivered to clients that hit the version short-circuit. The value is computed once and reused.

The field is a bool; the web client's `parseRemoteBool(true)` reads it correctly with no 0/1 special-casing.

## Tests

Added 3 tests in `modules/common/api_test.go`:
- `OCTO_SEARCH_BACKEND=es` → both `messages_search_on` and `search_enabled` are `true`.
- `OCTO_SEARCH_BACKEND=disabled` → both `false`.
- version short-circuit branch also emits `messages_search_on`.

All pass (`go test ./modules/common/ -run TestGetAppConfig`). Full `go build ./...` and `go vet ./modules/common/...` clean.

Part of #395

### DIFF
--- a/modules/common/api.go
+++ b/modules/common/api.go
@@ -396,13 +396,18 @@ func (cn *Common) appConfig(c *wkhttp.Context) {
 	if err != nil && versionStr != "" {
 		cn.Warn("解析版本号失败", zap.String("version", versionStr), zap.Error(err))
 	}
+	// 消息搜索开关：search_enabled / messages_search_on 同源同值。前者保留向后兼容，
+	// 后者是收敛后的真源 key（与 octo-web ChannelSearch、octo-admin messages_search
+	// 命名一致）。两个分支共用同一计算结果，避免 Resolve 被调用多次。
+	searchEnabled := searchbackend.Resolve(cn.ctx.GetConfig().ZincSearch.SearchOn).SearchEnabled()
 	if versionI64 != 0 && int(versionI64) >= appConfigM.Version {
 		c.JSON(http.StatusOK, &appConfigResp{
 			Version:                appConfigM.Version,
 			SystemBotUIDs:          spacepkg.SystemBotList(),
 			LocalLoginOff:          boolToFlag(cn.systemSettings.LocalLoginOff()),
 			DisableUserCreateSpace: boolToFlag(cn.systemSettings.SpaceDisableUserCreate()),
-			SearchEnabled:          searchbackend.Resolve(cn.ctx.GetConfig().ZincSearch.SearchOn).SearchEnabled(),
+			SearchEnabled:          searchEnabled,
+			MessagesSearchOn:       searchEnabled,
 		})
 		return
 	}
@@ -441,7 +446,8 @@ func (cn *Common) appConfig(c *wkhttp.Context) {
 		SystemBotUIDs:          spacepkg.SystemBotList(),
 		LocalLoginOff:          boolToFlag(cn.systemSettings.LocalLoginOff()),
 		DisableUserCreateSpace: boolToFlag(cn.systemSettings.SpaceDisableUserCreate()),
-		SearchEnabled:          searchbackend.Resolve(cn.ctx.GetConfig().ZincSearch.SearchOn).SearchEnabled(),
+		SearchEnabled:          searchEnabled,
+		MessagesSearchOn:       searchEnabled,
 	})
 }
 
@@ -764,6 +770,13 @@ type appConfigResp struct {
 	// 与 app_config.version 解耦的原因同 LocalLoginOff / DisableUserCreateSpace：
 	// 运维切 backend 后老客户端命中 version 短路分支也必须拿到最新值。
 	SearchEnabled bool `json:"search_enabled"`
+
+	// MessagesSearchOn 与 SearchEnabled 同源同值，是消息搜索功能开关下发给前端
+	// 的「收敛后」key。Web 端 (octo-web ChannelSearch) 读 messages_search_on，
+	// 与 octo-admin 的 messages_search 命名保持一致；旧端仍读 search_enabled。
+	// 两字段并存：search_enabled 保留作向后兼容，messages_search_on 为新真源 key。
+	// 下发 bool，前端 parseRemoteBool(true) 即可正确开关，无需 0/1 特判。
+	MessagesSearchOn bool `json:"messages_search_on"`
 }
 
 type oidcProviderResp struct {
--- a/modules/common/api_test.go
+++ b/modules/common/api_test.go
@@ -167,6 +167,66 @@ func TestGetAppConfig_SystemBotUIDsOnVersionShortCircuit(t *testing.T) {
 	assert.Contains(t, body, `"fileHelper"`)
 }
 
+// 消息搜索开关下发：search_enabled 与 messages_search_on 必须同源同值。
+// messages_search_on 是收敛后的真源 key（octo-web ChannelSearch 读它，与
+// octo-admin messages_search 命名一致），search_enabled 保留作向后兼容。
+// OCTO_SEARCH_BACKEND=es → 两者都为 true。
+func TestGetAppConfig_MessagesSearchOn_Enabled(t *testing.T) {
+	t.Setenv("OCTO_SEARCH_BACKEND", "es")
+	s, ctx := testutil.NewTestServer()
+	f := New(ctx)
+	cleanAllTablesAndReloadSettings(t, ctx)
+	err := f.appConfigDB.insert(&appConfigModel{})
+	assert.NoError(t, err)
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/v1/common/appconfig", nil)
+	req.Header.Set("token", testutil.Token)
+	s.GetRoute().ServeHTTP(w, req)
+	assert.Equal(t, http.StatusOK, w.Code)
+	body := w.Body.String()
+	assert.Contains(t, body, `"messages_search_on":true`)
+	assert.Contains(t, body, `"search_enabled":true`)
+}
+
+// OCTO_SEARCH_BACKEND=disabled → messages_search_on 与 search_enabled 均为
+// false，前端据此隐藏 ChannelSearch 入口。
+func TestGetAppConfig_MessagesSearchOn_Disabled(t *testing.T) {
+	t.Setenv("OCTO_SEARCH_BACKEND", "disabled")
+	s, ctx := testutil.NewTestServer()
+	f := New(ctx)
+	cleanAllTablesAndReloadSettings(t, ctx)
+	err := f.appConfigDB.insert(&appConfigModel{})
+	assert.NoError(t, err)
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/v1/common/appconfig", nil)
+	req.Header.Set("token", testutil.Token)
+	s.GetRoute().ServeHTTP(w, req)
+	assert.Equal(t, http.StatusOK, w.Code)
+	body := w.Body.String()
+	assert.Contains(t, body, `"messages_search_on":false`)
+	assert.Contains(t, body, `"search_enabled":false`)
+}
+
+// version 短路分支同样要下发 messages_search_on：老客户端命中版本短路也必须
+// 拿到当前开关，否则被缓存住失去实时性（与 search_enabled / disable_user_create_space
+// 同样的「与 app_config.version 解耦」约束）。
+func TestGetAppConfig_MessagesSearchOn_OnVersionShortCircuit(t *testing.T) {
+	t.Setenv("OCTO_SEARCH_BACKEND", "es")
+	s, ctx := testutil.NewTestServer()
+	f := New(ctx)
+	cleanAllTablesAndReloadSettings(t, ctx)
+	err := f.appConfigDB.insert(&appConfigModel{})
+	assert.NoError(t, err)
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/v1/common/appconfig?version=99999999", nil)
+	req.Header.Set("token", testutil.Token)
+	s.GetRoute().ServeHTTP(w, req)
+	assert.Equal(t, http.StatusOK, w.Code)
+	body := w.Body.String()
+	assert.Contains(t, body, `"messages_search_on":true`)
+	assert.Contains(t, body, `"search_enabled":true`)
+}
+
 // appconfig 必须下发 disable_user_create_space：默认 0（缺 system_setting 行且
 // env 未设置）。客户端据此显示/隐藏「创建空间」入口；缺省必须保持开放。
 func TestGetAppConfig_DisableUserCreateSpace_DefaultsZero(t *testing.T) {


lml2468 (approved):
✅ APPROVE — 干净的键名不匹配修复。appconfig 现同时下发 messages_search_on(与 search_enabled 同源同值),修了 web 读 messages_search_on、后端只发 search_enabled 导致搜索入口永远开不了的问题。无 blocking。

## 修复正确
- web 端(octo-web ChannelSearch,#386)gate 在 appconfig.messages_search_on,但后端只发 search_enabled → 键名永不匹配 → 开关开不了、搜索入口不出现。本 PR 让 appconfig 同时下发 messages_search_on。
- **同源同值,无 drift**:把 searchbackend.Resolve(...).SearchEnabled() 提取为局部变量 searchEnabled,两个出口的 SearchEnabled 与 MessagesSearchOn **都赋同一个变量**(api.go:409/410、449/450)。两字段结构上不可能漂移;顺带避免了原代码每出口各调一次 Resolve。
- **两个出口都覆盖**:version 短路分支 + 正常分支都加了 MessagesSearchOn,老客户端命中版本短路也能拿到最新开关(与 SearchEnabled/DisableUserCreateSpace 同样的「与 app_config.version 解耦」约束)。

## 与 #376 的关系已厘清(无冲突)
#376 曾用独立的 system_setting(search.messages_on)+ MessagesSearchOn() getter 实现同名字段,但 #376 已 CLOSED(未合并)。当前 main 无任何 messages_search_on 定义(已核),故 #399 是真正引入该字段者,无重复定义、无语义冲突。团队收敛到「messages_search_on 与 search_enabled 同源同值(均基于 backend resolve)」的单一真源方案——比两个独立开关(system_setting toggle vs backend resolve)各自为政更自洽:messages_search_on = 后端能否提供搜索,后端关闭时前端隐藏入口语义正确。

## 验证
- 测试覆盖三场景:OCTO_SEARCH_BACKEND=es → 两字段 true;=disabled → 两字段 false;version 短路 → 仍下发 true(老客户端实时性)。
- bool 下发(非 0/1),注释说明前端 parseRemoteBool 可直接用。
- go vet 干净。(appconfig 集成测试走 testutil.NewTestServer 需 MySQL,本地无库,CI 跑。)

无 blocking、无 should-fix。批准。

—— Octo-PR (lml2468)

Jerry-Xin (approved):
The PR is in scope for octo-server and cleanly extends `/v1/common/appconfig` with the canonical `messages_search_on` field while preserving `search_enabled`.

✅ Highlights
- `modules/common/api.go:402` computes the resolved search flag once and reuses it, avoiding branch drift.
- `modules/common/api.go:409-410` and `modules/common/api.go:449-450` emit both fields in the version short-circuit and full response paths.
- `modules/common/api.go:774-779` adds the new response field without changing the existing wire contract.
- `modules/common/api_test.go:174-227` covers enabled, disabled, and version short-circuit behavior.

💬 Non-blocking
- No blocking issues found.
- Verification caveat: I attempted `go test ./modules/common/ -run TestGetAppConfig`, but local setup failed before reaching these tests because `testutil.NewTestServer` panicked on an existing unknown database migration: `20191106000001_event_legacy01.sql`. This appears to be local test DB state, not caused by this PR.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#399 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/399
**Head SHA:** `01e28a39836491f78ccb7d2bec7b650035ba3c42`
**Title:** feat(appconfig): emit messages_search_on alongside search_enabled
**Files:** `modules/common/api.go` (+15/-2), `modules/common/api_test.go` (+60/-0)

---

## 1. 验证结论

- ✅ **新字段 `messages_search_on` 正确添加到 `appConfigResp` struct** — `api.go:778`, JSON tag `messages_search_on`, type `bool`
- ✅ **两个响应分支均正确设置 `MessagesSearchOn`** — version 短路分支 (`api.go:411`) + 完整响应分支 (`api.go:453`)
- ✅ **单一计算源** — `searchEnabled` 变量在 `api.go:402` 计算一次，两处复用，避免冗余 `Resolve()` 调用
- ✅ **数据源回溯正确** — `searchbackend.Resolve(cfg.ZincSearch.SearchOn).SearchEnabled()` → `Mode{ESServe || LegacyZinc}` → `bool`（`pkg/searchbackend/backend.go:71-92`）
- ✅ **3 个测试覆盖** — enabled (`api_test.go:174`), disabled (`api_test.go:193`), version short-circuit (`api_test.go:213`)
- ✅ **向后兼容** — `search_enabled` 字段保留，值不变；`messages_search_on` 纯增量

## 2. 发现的问题（P0/P1/P2）

**无 P0/P1/P2 问题。**

本 PR 改动范围小且清晰：
- 新增一个 struct 字段 + 两处赋值 + 三个测试
- 不触碰任何现有逻辑路径
- 不改变 `search_enabled` 的行为或值

## 3. 建议

无需修改。PR 可以按当前状态合入。

## 4. 额外发现

**Note（非阻塞，pre-existing）：** 代码库中有 3 处遗留调用点直接读取 `ZincSearch.SearchOn` 而不经过 `searchbackend.Resolve()`：
- `modules/message/api.go:840`, `modules/message/api.go:2520`
- `modules/message/api_manager.go:226`

当 `OCTO_SEARCH_BACKEND=disabled` 时，这些路径可能仍看到 `SearchOn=true`。这是 pre-existing 技术债，与本 PR 无关，但建议后续统一收敛到 `Resolve()` 路径。

## 5. 数据流回溯

| 消费点 | 上游来源 | 数据是否真的流到消费点 |
|--------|----------|----------------------|
| `appConfigResp.MessagesSearchOn` (api.go:411, version 短路) | `searchEnabled` var (api.go:402) ← `searchbackend.Resolve(cfg.ZincSearch.SearchOn).SearchEnabled()` (backend.go:71-92) | ✅ 是。`Resolve()` 在 handler 内同步调用，`SearchEnabled()` 返回 `ESServe \|\| LegacyZinc`，赋值给局部变量后直接写入 struct |
| `appConfigResp.MessagesSearchOn` (api.go:453, 完整响应) | 同上 `searchEnabled` var | ✅ 是。同一变量复用 |
| `appConfigResp.SearchEnabled` (两处) | 同上 `searchEnabled` var | ✅ 是。与改动前行为完全一致，仅从 inline 调用改为变量引用 |

**`Resolve()` 内部数据流：**
- 读 `OCTO_SEARCH_BACKEND` env var → switch on lowered/trimmed value
- `es` → `ESServe=true, LegacyZinc=false` → `SearchEnabled()=true`
- `zinc` → `ESServe=false, LegacyZinc=true` → `SearchEnabled()=true`
- `disabled` → both false → `SearchEnabled()=false`
- unset/unknown → `ESServe=true, LegacyZinc=cfg.ZincSearch.SearchOn` → `SearchEnabled()=true` (ESServe alone suffices)

## 6. 三类盲点 checklist（R5）

- **C1 双路径 parity：** ✅ Clear. 两个响应分支（version 短路 / 完整响应）均添加了 `MessagesSearchOn: searchEnabled`，改动对称一致。
- **C2 control-flow ordering / 嵌套复用：** N/A. 无嵌套复用或安全控件改动。`searchEnabled` 是简单局部变量，无 ordering 风险。
- **C3 授权边界 ≠ 能力边界：** N/A. 无授权/权限改动。`/v1/common/appconfig` 是公开无认证端点（`commonNoAuth` group, api.go:74-78）。
- **C4 授权生命周期 / 容器-成员状态级联：** N/A. 无鉴权改动。

## 7. 跨轮 blocker 复检（R6）

N/A — 首次 review。

---

[Octo-Q] verdict: APPROVE + 改动清晰、范围小、测试充分、数据流回溯无异常。无 P0/P1/P2 问题。唯一 note 是 pre-existing 遗留 `ZincSearch.SearchOn` 直接读取点（非本 PR 范围）。

yujiawei (commented):
# Code Review — PR #399 (octo-server)

**Verdict: APPROVED** (no blocking issues). One notable semantics recommendation below, plus minor test-quality notes.

## Verification

| Item | Result | Evidence |
|---|---|---|
| Change is additive / backward-compatible | ✅ | New `MessagesSearchOn bool` field added; `search_enabled` unchanged. No removed/renamed JSON keys → existing iOS/Android/older-web consumers unaffected. `modules/common/api.go:776-782` |
| Value computed once, reused in both branches | ✅ | `searchEnabled := searchbackend.Resolve(...).SearchEnabled()` then reused in the version short-circuit branch and the normal branch — the two keys can never diverge. `modules/common/api.go:402,410,450` |
| Version short-circuit branch also emits the new key | ✅ | Consistent with the documented "decouple from app_config.version" rationale for `SearchEnabled`/`LocalLoginOff`/`DisableUserCreateSpace`. `modules/common/api.go:404-412` |
| Cross-repo contract (web reads it as a JSON bool) | ✅ | Verified in the consumer: `this.messagesSearchOn = parseRemoteBool(result["messages_search_on"])`. Plain `bool` is correct; no 0/1 special-casing needed. |
| `go vet ./modules/common/...` | ✅ | clean |
| `go test ./modules/common/ -run TestGetAppConfig` | ✅ | pass |

## Findings

### P2 — `messages_search_on` source vs. the surface it actually gates (recommend, non-blocking)

`messages_search_on` is sourced from `searchbackend.Resolve(...).SearchEnabled()`, which is `ESServe || LegacyZinc` (`pkg/searchbackend/backend.go:92`) — i.e. **true under both `es` and `zinc`**.

But the web consumer that reads this key gates its search **entry** on it *and* sends requests to `messages/_search*`, and those endpoints are gated server-side on **`ESServe` only**:

```go
// modules/messages_search/api.go (backendGate)
if !h.mode.ESServe {
    respondSearchDisabled(c)   // SEARCH_DISABLED
    c.Abort()
    return
}
```

Under `OCTO_SEARCH_BACKEND=zinc` (the documented rollback state): `SearchEnabled()=true` (LegacyZinc on) → `messages_search_on=true` → the web search entry appears, **but every `messages/_search*` call returns `SEARCH_DISABLED`** → a visible-but-broken search UI during a zinc rollback.

This is confined to the explicit `zinc` rollback mode (`es`, `disabled`, and unset/default all behave correctly), so it is **not a merge blocker** — but it is a genuine semantic question worth resolving before the end-to-end feature ships:

- `messages_search_on` is named as the **canonical key for the new (ES-backed) message search** (matches octo-admin `messages_search`), and the web consumer uses it to gate the ES `_search*` UI. By that intent the correct source is **`mode.ESServe`** (es→true, zinc→false, disabled→false), which would gate the entry exactly to where the surface is live.
- If the intent is instead "any message search is live", then the naming/comment is misleading relative to what the consumer actually calls, and the web side would need a legacy-endpoint fallback under zinc (it currently hardcodes `_search*`).

Recommendation: source `MessagesSearchOn` from `mode.ESServe` (keep `SearchEnabled` as-is for the legacy `search_enabled` key), and add a `zinc`-backend test asserting the intended value. `search_enabled`'s own semantics are unchanged by this PR, so this is purely about getting the *new* key's gate right.

### P2 — Test coverage gap (minor)

The 3 added tests cover only `es` (both true) and `disabled` (both false). The `zinc` and unset/default cases — precisely where `messages_search_on` and ES-surface availability can diverge — are untested. Adding a `zinc` case would force an explicit decision on the source-value question above.

### P2 — Brittle assertion style + `t.Setenv`/`t.Parallel` coupling (nit)

- Tests assert via raw-JSON substring matching (`assert.Contains(t, body, "\"messages_search_on\":true")`) rather than unmarshaling. Low risk (Go's `encoding/json` emits no space after the colon and this matches the file's existing convention), but unmarshal-and-assert would be more robust.
- `t.Setenv` mutates global process env, which panics if the suite is ever parallelized with `t.Parallel()`. The new tests don't call `t.Parallel()` so they're safe today; flagging so future parallelization doesn't silently break them.

## Coverage / blind spots

- **Verified cross-repo**: the web consumer reads `messages_search_on` as a JSON bool and gates `messages/_search*` (ES) on it — so the gemini-flagged "is it parsed as bool?" concern is **resolved** (it is), and the codex-flagged zinc divergence is **confirmed real**.
- **Not verified here**: octo-admin `messages_search` exact semantics; whether `zinc` rollback remains an operationally supported state (code comments say yes — "Rollback: read from Zinc"). iOS/Android consumers of `search_enabled` are unaffected (value unchanged).

## Summary

Clean, well-scoped, backward-compatible convergence change with good tests for the common cases. Approving. The one item worth the author's attention before the feature lands end-to-end is the **`SearchEnabled()` vs `mode.ESServe`** source choice for the new key, since under `zinc` rollback it produces a visible-but-broken web search entry.

lml2468:
认领 yujiawei 第 1 条 —— 我审时把 messages_search_on=SearchEnabled() 判为"语义正确",漏了 zinc 模式这个分支,诚实补正。这是 should-fix(非 blocking)。

## should-fix(我漏判):zinc 模式下 messages_search_on 会"UI 显示但搜索不可用"
核 searchbackend(backend.go):
- SearchEnabled() = m.ESServe || m.LegacyZinc —— ES 或 legacy zinc 任一开启即 true。
- 但 octo-web ChannelSearch 调的新 /v1/messages/_search* 端点**只由 ESServe 提供**(zinc 是 legacy 旧端点,新 messages_search 不认)。
- 边界:OCTO_SEARCH_BACKEND=zinc → ESServe=false, LegacyZinc=true → SearchEnabled()=true → messages_search_on=true → **web 显示搜索入口,但新端点实际返回 SEARCH_DISABLED/不可用**。正是 yujiawei 指的"UI 显示但搜索不可用"。

我审时只核了 es(双true)/disabled(双false)两态正确,**漏了 zinc 态**:zinc 下 ESServe≠SearchEnabled,导致 messages_search_on 失真。

## 建议(yujiawei 的方向对)
MessagesSearchOn 应取 mode.ESServe(messages_search 端点的真实可用性),而非泛化的 SearchEnabled()(含 legacy zinc)。这样 zinc 回滚态下 messages_search_on=false,web 不显示入口,与新端点实际能否服务一致。search_enabled 可保持 SearchEnabled() 语义不变(向后兼容)。

## 严重性:should-fix 非 blocking
zinc 是 rollback/legacy 回退模式(注释明写 "Rollback: read from Zinc"),非常规生产配置;即便误显示入口,点击得到的是 SEARCH_DISABLED 错误(功能降级,非数据泄露/崩溃)。所以不阻断,但建议合并前或紧接 follow-up 把 MessagesSearchOn 源改成 ESServe,让"显示入口==搜索可用"。

## 另两条(测试,同意,非阻塞)
- 补 zinc + default 模式的测试覆盖(正好能钉住上面的语义:zinc → messages_search_on=false)。
- JSON 字符串 contains 断言较脆,可改反序列化结构断言。

键名不匹配主修复(同源、两出口覆盖、向后兼容)没问题;把 messages_search_on 的源从 SearchEnabled() 收敛到 ESServe 会让语义更准。感谢 yujiawei 指出 zinc 这个分支 —— 我漏了。

—— Octo-PR (lml2468)

yujiawei (commented):
# Code Review — PR #399 (octo-server)

**Verdict: APPROVED.** No blocking (P0/P1) issues. The change is small, correct, well-tested, and builds clean. Findings below are all P2 (nits / coverage / doc-consistency suggestions) and do not block merge.

Reviewed at head SHA `01e28a39836491f78ccb7d2bec7b650035ba3c42` (base `main`, merge-base `776f3500`).

## 1. Verification

- ✅ **Correctness — single source, both branches.** `searchEnabled` is computed once (`api.go:402`) and reused for both `SearchEnabled` and `MessagesSearchOn` in *both* response branches: the version short-circuit (`api.go:404-411`) and the normal path (`api.go:429-451`). Values are byte-identical; no path sets one field without the other.
- ✅ **No missed construction site.** `grep` confirms exactly two `appConfigResp{` literals in the repo (`api.go:404`, `api.go:429`) — both updated. No third client construction site leaks a zero-value `messages_search_on:false`.
- ✅ **Type contract.** `MessagesSearchOn bool` serializes as a JSON bool (`messages_search_on:true|false`), matching the stated web client expectation (`parseRemoteBool`). Note the same endpoint also emits some flags as 0/1 ints via `boolToFlag` (`local_login_off`, `disable_user_create_space`); the new field correctly follows `search_enabled`'s bool convention, not the int one.
- ✅ **Build & tests.** `go build ./modules/common/...` clean; `go test ./modules/common/ -run TestGetAppConfig` passes. New tests use `t.Setenv` without `t.Parallel`, so env mutation is safe. No golden/`JSONEq` snapshot asserts an exact appconfig body, so the additive field breaks no existing assertion.
- ✅ **Decoupled from `app_config.version`.** Both fields are recomputed per request from `OCTO_SEARCH_BACKEND` (+ `ZincSearch.SearchOn` in the default case), so clients hitting the version short-circuit still get the live value — consistent with the existing `search_enabled` / `disable_user_create_space` rationale.

## 2. Findings (all P2 — non-blocking)

**[P2] Comment overstates the dedup benefit (`api.go:399-401`).**
The comment says hoisting `searchEnabled` "避免 Resolve 被调用多次". Pre-PR, `Resolve` was already called inline in only the branch that executes, then returned — so it ran at most once per request. The real benefit is deduplicating the *identical expression* across two branches (DRY), not eliminating a repeated call. Suggest softening the comment to match.

**[P2] Test coverage gap: `zinc` and unset/default backend modes untested (`api_test.go:173-228`).**
`searchbackend.Resolve` has four states (`backend.go:71-93`); the new tests cover only `es` and `disabled`. The **unset/default** path is the most valuable to pin because it is the production fallback for deployments that never set `OCTO_SEARCH_BACKEND` — there `ESServe` is unconditionally true, so `messages_search_on` emits `true`. A one-line test asserting the unset default emits `messages_search_on:true` would guard against a future change to `Resolve`'s default silently flipping the new client-visible key. A `zinc`-mode test (`SearchEnabled()` true via `LegacyZinc`) would round out the matrix.

**[P2] Assertions are substring-based; an unmarshal would be more robust (`api_test.go:187-188, 206-207, 226-227`).**
`assert.Contains(t, body, `"messages_search_on":true`)` does not false-match today (keys are distinct; `search_enabled` is not a substring of `messages_search_on`). But it is brittle to formatting (indented JSON would break every assertion) and does not directly assert the PR's core invariant. Decoding into the struct and asserting `resp.MessagesSearchOn == resp.SearchEnabled == want` would be value-exact and would pin the "same source, same value" contract directly rather than inferring it from two independent substring hits.

**[P2] In-repo doc/comment drift on the "canonical key" claim (`api.go:774-779` vs `pkg/errcode/messages_search.go:61`).**
The new struct comment declares `messages_search_on` the canonical key, but `pkg/errcode/messages_search.go:61` still reads "the client uses appconfig.search_enabled to hide the search box", and `docs/messages-search/*` define neither key. Not introduced by this PR, but since this PR makes `messages_search_on` the stated source of truth, updating that errcode comment (and optionally the api-spec doc) would keep the in-repo story self-consistent.

## 3. Suggestions

- Add an unset-default backend test (highest value of the coverage suggestions).
- Optionally switch the three new assertions to a struct unmarshal + `assert.Equal(resp.MessagesSearchOn, resp.SearchEnabled)` to lock the invariant.
- Tidy the `api.go:399-401` comment and the stale `errcode/messages_search.go:61` reference.

## 4. Additional observations (cross-repo, unverifiable here — flagged for awareness, not blocking)

- **Which endpoint does the admin console read?** This PR changes only the client endpoint `GET /v1/common/appconfig` (`appConfigResp`). The manager endpoint (`api_manager.go`, `managerAppConfigResp`) emits no search field at all. If the admin `messages_search` toggle is sourced from the *manager* appconfig surface, the stated admin alignment would be unmet and a follow-up would be needed. Worth confirming which surface the admin client consumes.
- **Web client contract.** The justification (web reads `messages_search_on` and parses it as a bool) lives in the web repo and cannot be verified from this checkout. The backend now emits a real JSON bool, which is correct *if* the client expects a bool.
- **Deploy uniformity.** `messages_search_on` is derived from `OCTO_SEARCH_BACKEND` at request time. A partial rollout where some replicas set the env and others don't would emit inconsistent values across requests. Pre-existing for `search_enabled`, but now load-bearing for a client-gated UI — worth a deployment runbook note that the env must be set uniformly across replicas.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#399 Review Report
**PR:** Mininglamp-OSS/octo-server#399
**Head SHA:** `01e28a39836491f78ccb7d2bec7b650035ba3c42`
**Title:** feat(appconfig): emit messages_search_on alongside search_enabled

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `go build ./...` | ✅ Clean | Exit 0, no output |
| `go vet ./modules/common/...` | ✅ Clean | Exit 0, no output |
| New tests (3) | ✅ All pass | `TestGetAppConfig_MessagesSearchOn_{Enabled,Disabled,OnVersionShortCircuit}` |
| Existing tests (24 total) | ✅ All pass | Full `TestGetAppConfig*` suite green |
| Both response branches updated | ✅ | `api.go:404-410` (version short-circuit) and `api.go:429-450` (full response) |
| Single source of truth | ✅ | `searchEnabled` computed once at `api.go:402`, reused in both branches |

## 2. Findings

**P0:** None
**P1:** None
**P2:** None

This is a clean, well-scoped additive change. No defects found.

## 3. Data-Flow Trace

### Upstream source chain

```
OCTO_SEARCH_BACKEND env var (os.Getenv in Resolve)
        +
ZincSearch.SearchOn config bool (from YAML / TS_ZINCSEARCH_SEARCHON)
        |
        v
searchbackend.Resolve(zincSearchOn bool) → Mode{ESServe, LegacyZinc, Declared}
        |                                    (value type, never nil, never panics)
        v
Mode.SearchEnabled() → bool  (ESServe || LegacyZinc)
        |
        v
searchEnabled local var (api.go:402) — computed once before branch split
        |
        ├── appConfigResp.SearchEnabled     (json: "search_enabled")
        └── appConfigResp.MessagesSearchOn  (json: "messages_search_on")
```

### Resolve() behavior matrix

| `OCTO_SEARCH_BACKEND` | `ESServe` | `LegacyZinc` | `SearchEnabled()` |
|---|---|---|---|
| `"es"` | true | false | **true** |
| `"zinc"` | false | true | **true** |
| `"disabled"` | false | false | **false** |
| unset | true | `=zincSearchOn` | **true** (always, since ESServe=true) |

### Nil/panic safety
- `Resolve()` returns `Mode` value type — cannot return nil
- `SearchEnabled()` has value receiver — operates on copy, simple `||` expression
- `cn.ctx.GetConfig()` always returns non-nil (set during `NewContext()` init)
- No panic path exists in the chain

### Response path completeness
- `appConfigResp` constructed at exactly **2 sites**: `api.go:404` and `api.go:429`
- Both sites set `MessagesSearchOn: searchEnabled` ✅
- Two early-return error paths (`api.go:388`, `api.go:393`) use `c.ResponseError()`, not `appConfigResp` — field irrelevant there ✅
- No other handler constructs `appConfigResp` (confirmed via grep)

## 4. Diff-Scope Analysis (R2)

For each observation, the three questions:

**New field `messages_search_on`:**
1. New/amplified/pre-existing? → **New** (purely additive)
2. Does PR change existing path behavior? → **No** — `search_enabled` still sourced from identical expression, just extracted to a local var
3. Base-branch behavior match? → `search_enabled` behavior is identical to base; new field has no base equivalent

**Extraction of `searchEnabled` local variable:**
1. New/amplified/pre-existing? → **New** refactoring
2. Does it change behavior? → **No** — `Resolve()` is pure (reads env var + config), no side effects, safe to call before the version check
3. Base-branch behavior match? → `SearchEnabled` field value is identical pre- and post-change

## 5. Suggestions

None required. Optional observations for the author's consideration:

- The `AppConfigResp` (exported, `service.go:128`) used by internal `GetAppConfig()` service callers does not carry `SearchEnabled` or `MessagesSearchOn`. If any future internal consumer needs this flag, it would need adding there too. Not a blocker — no current caller needs it.

## 6. R5 Blind-Spot Checklist

- **C1 (dual-path parity):** N/A — no symmetric add/remove or subscribe/unsubscribe paths touched.
- **C2 (control-flow ordering / nesting):** Clear — `Resolve()` is called once, no nesting or reuse risk.
- **C3 (auth boundary ≠ capability boundary):** N/A — this is a public feature flag, not an auth/capability change.
- **C4 (auth lifecycle / container-member cascade):** N/A — no auth or container-state involvement.

---

[Octo-Q] verdict: **APPROVE**

Clean additive change. Both response branches correctly emit `messages_search_on` alongside `search_enabled` from a single shared source. Data-flow trace confirms nil-safety and value correctness across all paths. Tests cover enabled, disabled, and version short-circuit scenarios. Build, vet, and full test suite pass clean.

PR #406, opened 2026-06-17
fix(group/test): bound test DB pool to stop MySQL connection exhaustion (OCT-8)

lml2468:
## Summary
Permanent fix for the P1 flaky test `TestHandleGroupDisbandEvent_CleansThreadMembers` that turned `main` red on 2026-06-17 with `panic: Error 1040: Too many connections`.

**Root cause:** `modules/group` has 200+ tests that each call `testutil.NewTestServer`, which builds a fresh `*config.Context` and lazily opens its own `*sql.DB` pool (octo-lib defaults: 100 max / 10 idle conns). Nothing closes those pools. Under `go test -race -shuffle=on` the abandoned pools accumulate idle connections until the CI MySQL service exhausts `max_connections` (151) and the suite panics. It surfaced on the disband test only because `-shuffle` happened to land it after enough leaked pools.

**Fix:** Add a package-local `newTestServer(t)` chokepoint (`modules/group/testserver_test.go`) that bounds each pool (12 max / 2 idle) and reaps idle connections after 1s so abandoned pools drain back to MySQL. Migrated all 112 `testutil.NewTestServer()` call sites to it.

Pools are intentionally **not** `Close()`d: `config.NewContext` starts background dispatchers (`EventPool`/`PushPool`/`RobotEventPool`) and a timing wheel that outlive the test; closing the pool under them panics a later test with `sql: database is closed`. Capping idle conns + short idle lifetime fixes the leak without that hazard.

No production code changed — test harness only.

## Test plan
Run against the full CI service stack (`mysql:8.0` / `redis:7` / `wukongim v2.2.4-20260313`), DB recreated `utf8mb4_general_ci`:

- [x] **Before (this branch's parent):** `go test -race -shuffle=on -count=5 ./modules/group/...` → `panic: Error 1040: Too many connections` at ~26s.
- [x] **After:** same command → `ok ... 140.566s`, exit 0. `TestHandleGroupDisbandEvent_CleansThreadMembers` passes 5/5.
- [x] Peak `MySQL Max_used_connections` during the green run = **21** (limit 151), down from exhausting 151+.
- [x] `go vet ./modules/group/` clean; `go test -race -c` builds.

Paperclip: OCT-8

### DIFF
--- a/modules/group/api_allow_no_mention_test.go
+++ b/modules/group/api_allow_no_mention_test.go
@@ -13,7 +13,7 @@ import (
 // (allow). This mirrors how the ALTER TABLE migration backfills existing rows,
 // so deploying this feature does not silently turn off any existing no-@ bot.
 func TestGroup_AllowNoMention_DefaultsTo1(t *testing.T) {
-	_, ctx := testutil.NewTestServer()
+	_, ctx := newTestServer(t)
 	assert.NoError(t, testutil.CleanAllTables(ctx))
 
 	groupNo := "g-default-nomention"
@@ -148,7 +148,7 @@ func TestGroupSettingUpdate_AllowNoMention_SilentToggleSucceeds(t *testing.T) {
 // non-manager/creator toggling the group-level switch gets 403, not 500. The
 // permission check runs before any DB/event write.
 func TestGroupSettingUpdate_AllowNoMention_NonManagerForbidden(t *testing.T) {
-	s, ctx := testutil.NewTestServer()
+	s, ctx := newTestServer(t)
 	wireI18nRendererForGroupTest(s)
 	f := New(ctx)
 
--- a/modules/group/api_bot_ownership_test.go
+++ b/modules/group/api_bot_ownership_test.go
@@ -41,7 +41,7 @@ func insertBotUser(t *testing.T, f *Group, uid, name, shortNo, creatorUID string
 // No external bots are owned by user-c by default; individual tests insert
 // the bots they need.
 func setupBotOwnershipGroup(t *testing.T) (*Group, http.Handler) {
-	s, ctx := testutil.NewTestServer()
+	s, ctx := newTestServer(t)
 	// NB: module.Setup(ctx) inside NewTestServer already constructs a Group
 	// instance and registers its routes on s.GetRoute(). Calling
 	// f.Route(s.GetRoute()) again would double-register the same paths and
@@ -183,7 +183,7 @@ func TestGroupMemberAdd_BotOwnershipBlocksMixedBatch(t *testing.T) {
 // TestCheckBotOwnership_SkipsNonBots exercises the pure helper: non-bot UIDs
 // pass through untouched regardless of inviter.
 func TestCheckBotOwnership_SkipsNonBots(t *testing.T) {
-	_, ctx := testutil.NewTestServer()
+	_, ctx := newTestServer(t)
 	assert.NoError(t, testutil.CleanAllTables(ctx))
 
 	_, err := ctx.DB().InsertInto("user").
--- a/modules/group/api_groupmd_test.go
+++ b/modules/group/api_groupmd_test.go
@@ -16,7 +16,7 @@ import (
 )
 
 func setupGroupMdTest(t *testing.T) (*Group, *server.Server) {
-	s, ctx := testutil.NewTestServer()
+	s, ctx := newTestServer(t)
 	f := New(ctx)
 	f.Route(s.GetRoute())
 	err := testutil.CleanAllTables(ctx)
--- a/modules/group/api_i18n_regression_test.go
+++ b/modules/group/api_i18n_regression_test.go
@@ -72,7 +72,7 @@ func putGroupSetting(t *testing.T, handler http.Handler, groupNo, body string) *
 // because it ignored getGroupInfo's not-found sentinel. The exit of a
 // non-existent group is a user-facing 404, not an internal error.
 func TestGroupExit_NotFoundGroup(t *testing.T) {
-	s, ctx := testutil.NewTestServer()
+	s, ctx := newTestServer(t)
 	wireI18nRendererForGroupTest(s)
 	_ = New(ctx)
 
@@ -95,7 +95,7 @@ func TestGroupExit_NotFoundGroup(t *testing.T) {
 // JSON-decode failure mapped to store_failed (500). An expired authorization
 // code is a normal user-facing state and must surface as auth_code_invalid.
 func TestGroupMemberInviteSure_ExpiredCode(t *testing.T) {
-	s, ctx := testutil.NewTestServer()
+	s, ctx := newTestServer(t)
 	wireI18nRendererForGroupTest(s)
 	_ = New(ctx)
 
@@ -129,7 +129,7 @@ func TestGroupMemberAdd_BlankMembersIsRequestInvalid(t *testing.T) {
 // RemoveGroupMembers return "none of the members are in this group" — a 404
 // business error, not the store_failed (500) it was being mapped to.
 func TestManagerMemberRemove_NotInGroupIsNotFound(t *testing.T) {
-	s, ctx := testutil.NewTestServer()
+	s, ctx := newTestServer(t)
 	wireI18nRendererForGroupTest(s)
 	f := New(ctx)
 
@@ -194,7 +194,7 @@ func TestGroupSettingUpdate_AllowExternalRangeIsRequestInvalid(t *testing.T) {
 // checkPermissions's "没有权限！" collapsed into store_failed (500). Updating a
 // group attribute without permission is a 403, not an internal error.
 func TestGroupSettingUpdate_NonManagerForbidden(t *testing.T) {
-	s, ctx := testutil.NewTestServer()
+	s, ctx := newTestServer(t)
 	wireI18nRendererForGroupTest(s)
 	f := New(ctx)
 
@@ -243,7 +243,7 @@ func TestGroupAvatarUpload_MissingFileIsRequestInvalid(t *testing.T) {
 // upload contract: after object storage and DB update succeed, the IM
 // notification is best-effort and must not make the client retry the upload.
 func TestGroupAvatarUpload_PostCommitNotifyFailureRespondsOK(t *testing.T) {
-	_, ctx := testutil.NewTestServer()
+	_, ctx := newTestServer(t)
 	err := testutil.CleanAllTables(ctx)
 	assert.NoError(t, err)
 
--- a/modules/group/api_invite_detail_optional_auth_test.go
+++ b/modules/group/api_invite_detail_optional_auth_test.go
@@ -36,7 +36,7 @@ import (
 // 合法 token + 同 Space 成员：detail 应该返回 joinable，让前端显示加入按钮。
 // 这是 YUJ-39 的核心正向路径——修复 PR#1174 遗留 UX 短板。
 func TestGroupInviteDetail_OptionalAuth_SameSpaceMember_Joinable(t *testing.T) {
-	s, ctx := testutil.NewTestServer()
+	s, ctx := newTestServer(t)
 	f := New(ctx)
 
 	err := testutil.CleanAllTables(ctx)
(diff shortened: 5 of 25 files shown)


mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#406 Review Report
**PR:** Mininglamp-OSS/octo-server #406
**Head SHA:** `41f915bc2fd64c8236fecadcf01a5f19b533ec6e`
**Title:** fix(group/test): bound test DB pool to stop MySQL connection exhaustion (OCT-8)

---

## 1. Verification Summary

| Item | Status | Evidence |
|---|---|---|
| All call sites migrated | ✅ | grep confirms only 1 `testutil.NewTestServer()` remains in `modules/group/` — inside the `newTestServer` wrapper itself (`testserver_test.go:31`). 112+ call sites across 24 test files migrated. |
| Pool bounds applied correctly | ✅ | `bindTestDBPool` sets MaxOpenConns=12, MaxIdleConns=2, ConnMaxIdleTime=1s, ConnMaxLifetime=30s. Data flow: `ctx.DB()` → `*dbr.Session` embeds `*Connection` embeds `*sql.DB` — pool methods available via Go struct embedding. |
| DB initialized before bounding | ✅ | `testutil.NewTestServer()` calls `CleanAllTables(ctx)` which triggers `ctx.DB()` (lazy `sync.Once`). By the time `bindTestDBPool(ctx)` runs in the wrapper, the `*sql.DB` is already created. |
| No production code changed | ✅ | All 25 changed files are `*_test.go`. Zero non-test files touched. |
| Not closing pool is justified | ✅ | `config.NewContext` starts EventPool/PushPool/RobotEventPool dispatchers + timing wheel (goroutines that outlive the test). Closing the pool under them causes `sql: database is closed` panics. Verified in octo-lib source: no `Close()`/`Stop()` exists on `config.Context` or `pool.Collector`. |
| `t.Helper()` used | ✅ | `testserver_test.go:30` — correct, ensures test failures attribute to the calling test, not the wrapper. |
| Skipped tests migrated | ✅ | Tests with `t.Skip()` also migrated — harmless (skip runs before `newTestServer`), ensures consistency when un-skipped. |

## 2. Findings

### No P0 or P1 issues found.

### P2 — Incomplete coverage: same leak exists in 21 other modules

**Diff-scope:** pre-existing (not introduced or amplified by this PR).

The same DB pool leak pattern this PR fixes in `modules/group/` exists across **21 other modules** with approximately **320 unbounded `testutil.NewTestServer()` call sites**. Highest-risk modules by volume:

| Module | Unbounded call sites |
|---|---|
| modules/user | ~119 |
| modules/common | ~44 |
| modules/thread | ~38 |
| modules/workplace | ~28 |
| modules/message | ~29 (3 different constructors, none bounding pool) |
| modules/oidc | ~18 |
| modules/robot | ~5 (local `newTestServer()` also unbounded) |

Running `go test -race -shuffle=on` across all modules together could still exhaust MySQL `max_connections` even after this PR lands.

**Recommendation:** File a follow-up to either (a) upstream the pool-bounding into `testutil.NewTestServer()` in octo-lib so all consumers benefit, or (b) replicate the group-style wrapper in every affected module. Option (a) is strongly preferred — it's a one-line change in octo-lib vs. 21 module-level wrappers.

### P2 — Redis/goroutine leaks remain

**Diff-scope:** pre-existing.

Each `testutil.NewTestServer()` also leaks: a Redis client, 3 goroutine pools (EventPool/PushPool/RobotEventPool with `loopPop` goroutines that `range` over never-closed channels), and a timing wheel. These are octo-lib architectural debt (no `Close()` method on `config.Context` or `pool.Collector`). Not blocking for this PR, which correctly scopes to the MySQL exhaustion symptom.

### Nit — Consider `t.Cleanup()` for additional safety

The wrapper could register `t.Cleanup(func() { conn.SetMaxIdleConns(0) })` to force-drain the pool immediately when the test finishes, rather than waiting for the 1s idle timeout. This is a minor optimization, not a correctness issue — the current approach works fine.

## 3. Suggestions

1. **Follow-up ticket:** Apply the same pool-bounding fix to the remaining 21 modules (or upstream to octo-lib `testutil.NewTestServer`).
2. **Long-term:** Add `Close()`/`Teardown()` to `config.Context` in octo-lib to properly clean up DB, Redis, goroutine pools, and timing wheel. The current "leak everything, bound the damage" approach works for tests but is architectural debt.

## 4. Additional Findings

- `config.Context.GetElasticsearch()` has a thread-safety issue: checks `c.elasticClient == nil` without mutex or `sync.Once` — concurrent first-calls could create multiple clients. Not related to this PR but worth noting for octo-lib maintainers.
- `pkg/db/mysql.go` production defaults are 2000 max open / 1000 max idle — very generous. The test defaults inherited from octo-lib (100 max / 10 idle) are already much lower, but still too high when multiplied by 200+ tests.

## 5. Data Flow Trace

```
testutil.NewTestServer()
  → config.NewContext(cfg)          // creates pools, dispatchers, timing wheel
  → CleanAllTables(ctx)
    → ctx.DB()                       // triggers sync.Once → creates *sql.DB
    → runs DELETE on all tables      // confirms DB is live
  → module.Setup(ctx)                // runs migrations
  → returns (server, ctx)

newTestServer(t)
  → testutil.NewTestServer()         // DB is live after this
  → bindTestDBPool(ctx)
    → conn := ctx.DB()               // returns already-initialized *dbr.Session
    → conn.SetMaxOpenConns(12)       // applied to embedded *sql.DB via *Connection
    → conn.SetMaxIdleConns(2)
    → conn.SetConnMaxIdleTime(1s)    // key: abandoned pools drain in 1s
    → conn.SetConnMaxLifetime(30s)
  → returns (server, ctx)
```

All data flows verified correct. No nil-pointer risk, no race between initialization and bounding.

- **C1 — Dual-path parity:** N/A. No add/remove, subscribe/unsubscribe, or create/delete symmetry involved. This is a test-harness-only change.
- **C2 — Control-flow ordering / nesting reuse:** N/A. `bindTestDBPool` is called exactly once per test, immediately after `NewTestServer`. No nested or repeated calls.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. No authorization, endpoints, or tool exposure changes.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. No auth changes.

---

**[Octo-Q] verdict: APPROVE**

Clean, well-scoped test-harness fix. Root cause analysis is sound, migration is complete, data flow is verified. The two P2 findings (incomplete cross-module coverage, remaining goroutine leaks) are pre-existing issues that this PR neither introduces nor amplifies. No blocking concerns.

yujiawei (approved):
# Code Review — PR #406 (octo-server)

**Verdict: APPROVE.** Reviewed at head `41f915bc2fd64c8236fecadcf01a5f19b533ec6e` against merge-base `fb4763d0`. This is a clean, well-reasoned, test-only fix for the P1 connection-exhaustion panic (OCT-8). No production code is touched. The diff is a mechanical migration plus one new helper file, and the design rationale is correct.

## 1. What the change does

- Adds `modules/group/testserver_test.go`: a package-local `newTestServer(t)` chokepoint that wraps `testutil.NewTestServer()` and calls `bindTestDBPool(ctx)` to bound each test's lazily-opened `*sql.DB` pool.
- Migrates all 112 `testutil.NewTestServer()` call sites across 24 `modules/group/*_test.go` files to `newTestServer(t)`.
- Pool bounds: `SetMaxOpenConns(12)`, `SetMaxIdleConns(2)`, `SetConnMaxIdleTime(1s)`, `SetConnMaxLifetime(30s)`.

## 2. Verification

- ✅ **Root cause confirmed.** `octo-lib/config/config.go:355-356` defaults to `MySQLMaxOpenConns=100 / MySQLMaxIdleConns=10`, and `octo-lib/testutil/test.go` builds a fresh `*config.Context` per call that is never closed. With 200+ tests in this package under `-shuffle=on`, abandoned pools accumulate idle connections until MySQL's `max_connections` (151) is exhausted. The diagnosis in the PR body and the file header comment is accurate.
- ✅ **The "do not `Close()`" rationale is correct and load-bearing.** `octo-lib/config/context.go:60-72` shows `NewContext` unconditionally starts `EventPool`/`PushPool`/`RobotEventPool` dispatchers and `timingWheel.Start()` — with no guard on `cfg.Test`. These outlive the test, so `Close()`-ing the pool out from under them would panic a later test with `sql: database is closed`. Capping connections + a short idle lifetime is the right way to bound a pool that must stay open.
- ✅ **Migration is complete.** Zero unbounded `testutil.NewTestServer()` calls remain in `modules/group` outside the wrapper definition itself. All old call sites were arg-less, so the new no-argument wrapper signature is fully compatible (the diff removes no `NewTestServer(<arg>)` forms).
- ✅ **Diff is purely mechanical** apart from the new helper — every changed line is either the `testutil.NewTestServer()` → `newTestServer(t)` swap or the now-unused `testutil` import removal. No behavioral test changes slipped in.
- ✅ **Compiles clean** (`go test -race -c` for the package builds; `go vet` reported clean by author).
- ✅ **Empirically validated** by the author: before → panic at ~26s; after → green at ~140s with peak `Max_used_connections` = 21 (vs. limit 151).

## 3. Notes (non-blocking)

- **P2 — Latent leak in other packages (follow-up).** The same unbounded-pool pattern exists in `modules/user` (~120 call sites), `modules/thread` (~38), `modules/workplace` (~28), `modules/message` (~19), and `modules/oidc` (~18), among others. They have not panicked yet only because each runs as a separate `go test` process (CI recreates the DB per package), and `group` happens to be the largest single package. This is **out of scope for this PR** — the change is correctly scoped to the package that is currently red — but the same `newTestServer(t)` pattern should be rolled out to those packages as a fast follow-up before one of them (most likely `user`) becomes the next flaky failure. Consider tracking this as a separate issue.
- **P2 — Root cause is arguably upstream.** The cleanest long-term fix is in `octo-lib` itself: either apply low pool defaults when `cfg.Test == true`, or expose a test-teardown hook. This per-package workaround is correct and appropriate to unblock CI now; an upstream `octo-lib` issue would prevent every downstream repo from re-deriving this same patch.
- **nit — slight transient over-allocation.** `bindTestDBPool` lowers the bounds *after* `testutil.NewTestServer()` has already lazily opened the pool at the 100/10 defaults. In practice this is harmless: tests are sequential (no `t.Parallel()` anywhere in the package, verified), so only one pool is active at a time and `SetMaxIdleConns(2)` structurally caps idle connections immediately on the next return-to-pool regardless of the idle-time reaper. The empirical peak of 21 confirms there is no real over-allocation window worth changing.

## 4. Why these do not block

None of the above are correctness defects in this PR. The change does exactly what it claims, is verified before/after, and removes a `main`-red panic without altering production behavior or any test's semantics. The connection bound is sound because the suite is sequential and `SetMaxIdleConns(2)` is a structural cap (not merely a timing-dependent reaper) — so the fix holds even if individual tests complete in well under one second.

Jerry-Xin (approved):
Project relevance gate: passed. This PR is in scope for `octo-server`; it changes only `modules/group` test infrastructure to address CI MySQL connection exhaustion.

No blocking findings.

💬 Non-blocking:
- 🔵 Suggestion — [modules/group/testserver_test.go](/Users/mlamp/.openclaw/tmp/pr-review-84w1tlg8/repo/modules/group/testserver_test.go:31): the pool cap is applied after `testutil.NewTestServer()` completes, so setup/cleanup inside octo-lib still briefly uses default pool settings. This is acceptable for the reported failure, but longer term this would be cleaner in octo-lib `testutil` so setup itself and other packages get the same bound.

✅ Highlights:
- [modules/group/testserver_test.go](/Users/mlamp/.openclaw/tmp/pr-review-84w1tlg8/repo/modules/group/testserver_test.go:36) correctly bounds the underlying DB session without closing it, matching the described background-worker constraint.
- All direct `testutil.NewTestServer()` calls in `modules/group` now route through `newTestServer(t)`.
- The disband regression test path also goes through `setupServiceTest`, which now uses the bounded helper.

Verification performed:
- `go test -c ./modules/group`
- `go test ./modules/group -run '^$'`
- `go test -race -c ./modules/group`

PR #408, opened 2026-06-17
feat(audit): operator logging on /v1/manager admin writes (#367 Phase 1)

### LINKED ISSUE #367: audit: record operator on admin management actions, then a centralized admin-action audit table
Follow-up from the authorization-model audit (#363).

## Problem

There is **no centralized admin-action audit** — you cannot answer "who did what in the management backend." Coverage today is very uneven:

- **`modules/space` is the good example**: every management write logs the operator + before/after via `zap`.
- **`modules/user` / `modules/group` / `modules/message` management handlers do not even record the operator** (only error context). After an incident there is no trail for actions like resetting a password, deleting/banning a user, force-removing a group member, or deleting a message.

This is a forensics/compliance gap for an enterprise IM backend.

## Proposed work (two phases)

**Phase 1 — operator logging (cheap, do first).**
Add `operator` (+ before/after where applicable) `zap` logging to every management write handler in `user` / `group` / `message`, mirroring the pattern already used in `modules/space/api_manager.go`. Low effort, immediate value.

**Phase 2 — centralized audit table.**
Introduce a queryable `admin_action` audit table (suggested columns: `operator_uid`, `action`, `target_type`, `target_id`, `before`/`after` or `detail` JSON, `ip`, `created_at`) written by all management write paths, so "who did what" is queryable rather than scattered across logs. Consider a small shared helper/middleware so new handlers get audited by default (ties in with the central authz work in the companion follow-up).

## Acceptance criteria

- [ ] Phase 1: every management write handler in `user`/`group`/`message` logs the operator (+ before/after where meaningful).
- [ ] Phase 2: `admin_action` table + write path; key management operations produce an audit row; documented retention.

## References
- Audit: #363 (section 五 审计现状 — "没有集中的管理操作审计表；user/group/message 的管理 handler 连 operator 都不记")
- Related systemic follow-up: centralized authz + CI guard (companion issue)

lml2468:
## Summary

Phase 1 of #367 — a forensic "who did what" trail for admin/management writes (logging only; the centralized audit table is Phase 2, SecurityEngineer).

- New global gin middleware `accesslog.OperatorAuditMiddleware` (`pkg/accesslog/operator_audit.go`), wired in `main.go` alongside the existing access logger.
- Emits **one** structured log line per mutating (`POST`/`PUT`/`PATCH`/`DELETE`) request under `/v1/manager`, with fields: `audit`, `operator_id`, `action`, `target`, `status`, `ts`.
- One chokepoint covers every manager subgroup (user/message/group/workplace/robot/common/…) **and future routes** — matching is on the raw request path. Operator id is read after `c.Next()` (AuthMiddleware sets `c.Set("uid", …)` during the matched group), which also lets us capture auth-rejected attempts (empty `operator_id` + 4xx).

## Security-by-default

- Never reads request bodies, so password-bearing routes (`/v1/manager/user/resetpassword`, `.../updatepassword`) can't leak secrets into logs.
- `target` is the URL **path only**, run through the existing `ScrubPath`; the query string is never logged.

## Format for Phase 2

Documented in `docs/operator-audit-log.md` so the audit-table ingest can key off the stable `manager_admin_action` discriminator.

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/accesslog/` — incl. a test asserting `operator_id` is present on a representative admin mutation (`POST /v1/manager/user/admin`), plus skip-reads / skip-non-manager / rejected-attempt / `isManagerWrite` table.
- [x] `gofmt` clean, `go vet ./pkg/accesslog/` clean
- [ ] CI green
- [ ] CTO security review (security owner) before merge

Closes #367 (Phase 1 scope only).

### DIFF
--- a/docs/operator-audit-log.md
+++ b/docs/operator-audit-log.md
@@ -1,0 +1,55 @@
+# Operator-Audit Log — admin/management writes (#367 Phase 1)
+
+Forensic "who did what" trail for the admin/management write surface. Every
+mutating request to `/v1/manager` emits **one** structured log line. This is
+**Phase 1 (logging only)** — Phase 2 (the centralized admin-action audit table,
+SecurityEngineer workstream) consumes these lines into durable storage.
+
+- Implementation: `pkg/accesslog/operator_audit.go`
+  (`OperatorAuditMiddleware`), wired as a global middleware in `main.go`.
+- Scope of writes audited: HTTP `POST` / `PUT` / `PATCH` / `DELETE` whose request
+  path is `/v1/manager` or under `/v1/manager/…` (covers `user`, `message`,
+  `group`, `workplace`, `robot`, `common`, and any future manager subgroup —
+  matching is on the raw request path, so new routes are covered automatically).
+- Read traffic (`GET`/`HEAD`) and non-manager routes are **not** audited.
+
+## Line format
+
+The line is emitted via the standard zap logger (`log.NewTLog("ManagerAudit")`),
+so it also carries the logger's own `ts` and `level`. Audit-specific fields:
+
+| field         | type   | meaning                                                              |
+| ------------- | ------ | -------------------------------------------------------------------- |
+| `audit`       | string | constant `manager_admin_action` — stable discriminator for ingest.  |
+| `operator_id` | string | acting user/agent id (`c.GetLoginUID()` / `c.Set("uid")`).           |
+| `action`      | string | HTTP method: `POST` \| `PUT` \| `PATCH` \| `DELETE`.                 |
+| `target`      | string | scrubbed request path identifying the resource acted on.             |
+| `status`      | int    | final HTTP response status code.                                     |
+| `ts`          | string | RFC3339 UTC timestamp when the line was emitted.                     |
+
+The zap message string is also `manager_admin_action`, identical to the `audit`
+field. Phase 2 can key ingest off either.
+
+### Reading `operator_id`
+
+- Non-empty `operator_id` + 2xx `status` ⇒ an authenticated admin successfully
+  performed the action.
+- **Empty** `operator_id` (paired with a 4xx `status`) ⇒ an unauthenticated or
+  auth-rejected attempt. These are intentionally audited too, as attempt
+  forensics; Phase 2 may choose to store or filter them.
+
+## Security-by-default (what is deliberately NOT logged)
+
+- **No request bodies.** The middleware never reads the body, so secrets that
+  ride in bodies cannot leak — notably `POST /v1/manager/user/resetpassword`
+  and `POST /v1/manager/user/updatepassword`.
+- **No query string.** `target` is the URL *path* only, run through
+  `ScrubPath` (the same scrubber used by the access logger), so even a path that
+  embeds a token is masked.
+- Consequence for Phase 2: the path names the target *surface* (e.g.
+  `/v1/manager/robots/:robot_id`); for actions whose target id lives in the body
+  (e.g. password reset), Phase 1 records the operator + action + endpoint but
+  not the body-side target. Capturing richer, safe target detail is Phase 2's
+  job in the audit table.
+
+GitHub: https://github.com/Mininglamp-OSS/octo-server/issues/367
--- a/main.go
+++ b/main.go
@@ -161,6 +161,11 @@ func runAPI(ctx *config.Context) {
 		}
 		accessLogger(c)
 	})
+	// Operator-audit log for admin/management writes (#367 Phase 1): one
+	// structured "who did what" line per mutating /v1/manager request. Reads the
+	// operator id set by AuthMiddleware after c.Next(); see pkg/accesslog
+	// operator_audit.go for the field schema Phase 2 consumes.
+	route.UseGin(accesslog.OperatorAuditMiddleware(log.NewTLog("ManagerAudit")))
 	// 全局 per-IP 作为 DDoS 底线：办公室共享出网 IP 下 IM 基础量就能到 100+ rps
 	// （每人 1-2 rps × 数十人），200 余量过小；真实 DDoS 常数千 rps+，底线设 500
 	// 更合理。精细限流交给 UID 层和端点级严格桶（#1090）。
--- a/pkg/accesslog/operator_audit.go
+++ b/pkg/accesslog/operator_audit.go
@@ -1,0 +1,87 @@
+package accesslog
+
+import (
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+	"github.com/gin-gonic/gin"
+	"go.uber.org/zap"
+)
+
+// Operator-audit logging (octo-server #367 Phase 1).
+//
+// Goal: a "who did what" forensic trail for admin/management writes. Every
+// mutating request to the /v1/manager surface emits exactly one structured log
+// line carrying the acting operator's id, the action, the target, the response
+// status and a timestamp. This is logging-only — the centralized audit table
+// (Phase 2, SecurityEngineer) consumes these lines later; see the format note
+// in this file's doc comments below.
+//
+// Security-by-default: the line is built from the request method, the
+// URL *path* (scrubbed via ScrubPath, never the query string) and the operator
+// id set by AuthMiddleware. Request bodies are never read, so secrets that ride
+// in bodies (e.g. /v1/manager/user/resetpassword, .../updatepassword) cannot
+// leak into logs. The path of a manager route names the target resource and is
+// itself the action metadata Phase 2 needs.
+
+const (
+	// managerPrefix is the admin/management write surface being audited.
+	managerPrefix = "/v1/manager"
+
+	// AuditLogMessage is the stable zap message (and "audit" field value) on
+	// every admin-action line. Phase 2 keys its audit-table ingest off this
+	// exact discriminator, so it must not change without coordinating the
+	// consumer.
+	AuditLogMessage = "manager_admin_action"
+)
+
+// OperatorAuditMiddleware returns a gin middleware that emits one structured
+// audit log line per mutating (POST/PUT/PATCH/DELETE) request under
+// /v1/manager.
+//
+// Field schema of each emitted line (in addition to the logger's own ts/level):
+//
+//   - audit:       constant "manager_admin_action" (stable discriminator)
+//   - operator_id: acting user/agent id (empty string => unauthenticated or
+//     auth-rejected attempt; pair with a 4xx status to read it that way)
+//   - action:      HTTP method (POST | PUT | PATCH | DELETE)
+//   - target:      scrubbed request path identifying the resource acted on
+//   - status:      final HTTP response status code
+//   - ts:          RFC3339 UTC timestamp of when the line was emitted
+//
+// The operator id is read after c.Next() because AuthMiddleware (which runs in
+// the matched route group, i.e. during c.Next()) is what sets it via
+// c.Set("uid", ...). Logging post-handler also lets us record the real
+// response status, so rejected/forbidden attempts are captured too.
+func OperatorAuditMiddleware(lg log.Log) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		if !isManagerWrite(c.Request.Method, c.Request.URL.Path) {
+			c.Next()
+			return
+		}
+		c.Next()
+		lg.Info(AuditLogMessage,
+			zap.String("audit", AuditLogMessage),
+			zap.String("operator_id", c.GetString("uid")),
+			zap.String("action", c.Request.Method),
+			zap.String("target", ScrubPath(c.Request.URL.Path)),
+			zap.Int("status", c.Writer.Status()),
+			zap.String("ts", time.Now().UTC().Format(time.RFC3339)),
+		)
+	}
+}
+
+// isManagerWrite reports whether (method, path) is a mutating action on the
+// /v1/manager admin surface. Match is on the raw request path (not the matched
+// route template) so that attempts which 404 under the manager prefix are still
+// audited.
+func isManagerWrite(method, path string) bool {
+	switch method {
+	case http.MethodPost, http.MethodPut, http.MethodPatch, http.MethodDelete:
+	default:
+		return false
+	}
+	return path == managerPrefix || strings.HasPrefix(path, managerPrefix+"/")
+}
--- a/pkg/accesslog/operator_audit_test.go
+++ b/pkg/accesslog/operator_audit_test.go
@@ -1,0 +1,166 @@
+package accesslog
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+	"go.uber.org/zap"
+)
+
+// captureLog is a log.Log fake that records the fields of the most recent
+// Info call so tests can assert on the emitted audit line.
+type captureLog struct {
+	calls []capturedLine
+}
+
+type capturedLine struct {
+	msg    string
+	fields map[string]zap.Field
+}
+
+func (c *captureLog) record(msg string, fields []zap.Field) {
+	m := make(map[string]zap.Field, len(fields))
+	for _, f := range fields {
+		m[f.Key] = f
+	}
+	c.calls = append(c.calls, capturedLine{msg: msg, fields: m})
+}
+
+func (c *captureLog) Info(msg string, fields ...zap.Field)  { c.record(msg, fields) }
+func (c *captureLog) Debug(msg string, fields ...zap.Field) {}
+func (c *captureLog) Error(msg string, fields ...zap.Field) {}
+func (c *captureLog) Warn(msg string, fields ...zap.Field)  {}
+
+// newAuditRouter builds a gin router wired exactly like production: the audit
+// middleware is global, and a fake AuthMiddleware (mirroring octo-lib's
+// c.Set("uid", ...)) runs inside the matched /v1/manager route group. operator
+// is the uid the fake auth sets; pass "" to simulate an unauthenticated actor.
+func newAuditRouter(lg *captureLog, operator string) *gin.Engine {
+	gin.SetMode(gin.TestMode)
+	r := gin.New()
+	r.Use(OperatorAuditMiddleware(lg))
+
+	mgr := r.Group("/v1/manager", func(c *gin.Context) {
+		if operator != "" {
+			c.Set("uid", operator)
+		}
+		c.Next()
+	})
+	mgr.POST("/user/admin", func(c *gin.Context) { c.Status(http.StatusOK) })
+	mgr.GET("/user/list", func(c *gin.Context) { c.Status(http.StatusOK) })
+	return r
+}
+
+// Representative admin mutation: the operator id must appear on the audit line.
+func TestOperatorAuditMiddleware_LogsOperatorOnAdminWrite(t *testing.T) {
+	lg := &captureLog{}
+	r := newAuditRouter(lg, "u_admin_42")
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodPost, "/v1/manager/user/admin", nil)
+	r.ServeHTTP(w, req)
+
+	if len(lg.calls) != 1 {
+		t.Fatalf("expected exactly 1 audit line, got %d", len(lg.calls))
+	}
+	line := lg.calls[0]
+	if line.msg != AuditLogMessage {
+		t.Errorf("audit message = %q, want %q", line.msg, AuditLogMessage)
+	}
+
+	op, ok := line.fields["operator_id"]
+	if !ok {
+		t.Fatal("audit line missing operator_id field")
+	}
+	if op.String != "u_admin_42" {
+		t.Errorf("operator_id = %q, want %q", op.String, "u_admin_42")
+	}
+
+	// Action, target and a timestamp must all be present for Phase 2 ingest.
+	if got := line.fields["action"].String; got != http.MethodPost {
+		t.Errorf("action = %q, want %q", got, http.MethodPost)
+	}
+	if got := line.fields["target"].String; got != "/v1/manager/user/admin" {
+		t.Errorf("target = %q, want %q", got, "/v1/manager/user/admin")
+	}
+	if _, ok := line.fields["ts"]; !ok {
+		t.Error("audit line missing ts field")
+	}
+	if got := line.fields["status"].Integer; got != int64(http.StatusOK) {
+		t.Errorf("status = %d, want %d", got, http.StatusOK)
+	}
+}
+
+// Read-only admin traffic must not be audited.
+func TestOperatorAuditMiddleware_SkipsReads(t *testing.T) {
+	lg := &captureLog{}
+	r := newAuditRouter(lg, "u_admin_42")
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/v1/manager/user/list", nil)
+	r.ServeHTTP(w, req)
+
+	if len(lg.calls) != 0 {
+		t.Fatalf("GET should not be audited, got %d lines", len(lg.calls))
+	}
+}
+
+// Non-manager writes must not be audited.
+func TestOperatorAuditMiddleware_SkipsNonManager(t *testing.T) {
+	lg := &captureLog{}
+	gin.SetMode(gin.TestMode)
+	r := gin.New()
+	r.Use(OperatorAuditMiddleware(lg))
+	r.POST("/v1/message/send", func(c *gin.Context) { c.Status(http.StatusOK) })
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodPost, "/v1/message/send", nil)
+	r.ServeHTTP(w, req)
+
+	if len(lg.calls) != 0 {
+		t.Fatalf("non-manager write should not be audited, got %d lines", len(lg.calls))
+	}
+}
+
+// An unauthenticated/rejected attempt is still audited, with an empty
+// operator_id — useful forensic signal when paired with a 4xx status.
+func TestOperatorAuditMiddleware_AuditsRejectedAttemptWithEmptyOperator(t *testing.T) {
+	lg := &captureLog{}
+	r := newAuditRouter(lg, "") // no uid set => unauthenticated
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodPost, "/v1/manager/user/admin", nil)
+	r.ServeHTTP(w, req)
+
+	if len(lg.calls) != 1 {
+		t.Fatalf("expected 1 audit line for the attempt, got %d", len(lg.calls))
+	}
+	if got := lg.calls[0].fields["operator_id"].String; got != "" {
+		t.Errorf("operator_id = %q, want empty for unauthenticated attempt", got)
+	}
+}
+
+func TestIsManagerWrite(t *testing.T) {
+	tests := []struct {
+		method, path string
+		want         bool
+	}{
+		{http.MethodPost, "/v1/manager/user/admin", true},
+		{http.MethodPut, "/v1/manager/group/liftban/g1/1", true},
+		{http.MethodDelete, "/v1/manager/robots/r1", true},
+		{http.MethodPatch, "/v1/manager/anything", true},
+		{http.MethodPost, "/v1/manager/workplace/app", true},
+		{http.MethodGet, "/v1/manager/user/list", false},
+		{http.MethodHead, "/v1/manager/user/list", false},
+		{http.MethodPost, "/v1/message/send", false},
+		{http.MethodPost, "/v1/managerx/foo", false}, // prefix must be a path segment
+		{http.MethodPost, "/v1/manager", true},       // bare prefix counts
+	}
+	for _, tt := range tests {
+		if got := isManagerWrite(tt.method, tt.path); got != tt.want {
+			t.Errorf("isManagerWrite(%q, %q) = %v, want %v", tt.method, tt.path, got, tt.want)
+		}
+	}
+}


mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#408 Review Report — feat(audit): operator logging on /v1/manager admin writes (#367 Phase 1)
**Head SHA:** `cec5f738f1159422c4fc99684c81e9aac8a336ed`

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Middleware correctly scoped to mutating `/v1/manager` requests | ✅ | `pkg/accesslog/operator_audit.go:78-87` — `isManagerWrite` checks method ∈ {POST,PUT,PATCH,DELETE} and path prefix `/v1/manager` with segment boundary (`path == prefix \|\| HasPrefix(path, prefix+"/")`) |
| `operator_id` data flow: AuthMiddleware → gin context → audit line | ✅ | AuthMiddleware (`octo-lib/pkg/wkhttp/http.go`) calls `c.Set("uid", uidAndNames[0])` on success; audit middleware reads `c.GetString("uid")` post-`c.Next()` — both operate on the same `gin.Context` keyspace (verified `wkhttp.Context` embeds `*gin.Context`) |
| Request bodies never read | ✅ | `operator_audit.go:62-73` — middleware constructs the log line from `c.Request.Method`, `c.Request.URL.Path`, `c.GetString("uid")`, and `c.Writer.Status()` only; no `c.BindJSON`, `c.GetRawData`, or body access |
| Query string never logged | ✅ | `operator_audit.go:70` — uses `c.Request.URL.Path` (path only), not `RawQuery` or `RequestURI` |
| Path scrubbing applied | ✅ | `operator_audit.go:70` — `ScrubPath(c.Request.URL.Path)` applied; `ScrubPath` (`accesslog.go:44-62`) only masks `/v1/incoming-webhooks/` tokens, but manager paths don't embed secrets in URL segments |
| Middleware ordering correct | ✅ | `main.go:168` — registered after access logger, before rate-limit/CORS; `c.Next()` runs first (downstream handlers execute), then audit line emitted with final `c.Writer.Status()` |
| Test coverage adequate | ✅ | `operator_audit_test.go` — 5 tests covering: authenticated admin write, read skip, non-manager skip, unauthenticated attempt (empty operator_id), `isManagerWrite` table (10 cases incl. `/v1/managerx` negative) |
| `log.Log` interface satisfied | ✅ | `captureLog` in tests implements all 4 methods (`Info/Debug/Error/Warn`); production uses `*log.TLog` which satisfies `log.Log` |

## 2. Findings

### No P0 or P1 issues found.

### P2 — Doc/code terminology mismatch (nit)

`docs/operator-audit-log.md:24` describes `operator_id` as `c.GetLoginUID() / c.Set("uid")`, but the implementation at `operator_audit.go:67` uses `c.GetString("uid")` (gin's built-in method). While functionally equivalent (both read from the same `gin.Context` key set by AuthMiddleware), the doc should match the code to avoid confusion for Phase 2 consumers.

**Severity:** P2 (documentation nit, no runtime impact)

### P2 — Redundant `ts` field (nit)

`operator_audit.go:72` adds an explicit `ts` field (`time.Now().UTC().Format(time.RFC3339)`), but the zap logger (`TLog.Info` → global zap logger) already emits its own `ts` field at the top level of each log entry. The two timestamps are independently generated and may differ by microseconds. Not harmful, but Phase 2 ingest should be aware of which `ts` to key on.

**Severity:** P2 (redundancy, no correctness impact)

## 3. Suggestions

1. **Doc fix:** Update `docs/operator-audit-log.md:24` to say `c.GetString("uid")` instead of `c.GetLoginUID()` for accuracy.
2. **Consider removing explicit `ts`:** Let zap's built-in timestamp be the single source of truth, or document which `ts` Phase 2 should consume.

## 4. Additional Observations

- **Login endpoint (`POST /v1/manager/login`)** is registered without AuthMiddleware (`modules/user/api_manager.go:60-62`). The audit line for this route will always have `operator_id: ""`. This is correct behavior (login is the bootstrap step), but worth noting for Phase 2 ingest — successful admin logins produce audit lines with empty operator_id + 200 status.
- **Rate-limited requests** (429 from the downstream rate-limit middleware) to manager write paths will still be audit-logged with `operator_id: ""` (rate-limit runs before AuthMiddleware). This is useful for brute-force forensics.
- **`ScrubPath` coverage:** Currently only masks `/v1/incoming-webhooks/` tokens. If future manager routes embed sensitive data in URL path segments, `ScrubPath` would need extension. No current manager route has this pattern.

## 5. Data-Flow Trace

|---|---|---|
| `c.GetString("uid")` | AuthMiddleware (`wkhttp/http.go`) sets `c.Set("uid", uidAndNames[0])` on auth success; not set on auth failure → returns `""` | ✅ Yes — same `gin.Context` keyspace; read post-`c.Next()` after AuthMiddleware runs |
| `c.Request.Method` | Go `net/http` standard library | ✅ Yes — set by HTTP server before any middleware |
| `c.Request.URL.Path` | Go `net/http` standard library | ✅ Yes — raw request path, not the matched route template |
| `ScrubPath(path)` | `accesslog.go:44-62` — masks `/v1/incoming-webhooks/` tokens; passes through all other paths | ✅ Yes — manager paths don't contain webhook tokens |
| `c.Writer.Status()` | gin's `ResponseWriter` — tracks the status code written by handlers | ✅ Yes — read post-`c.Next()`, after handlers have written the response |
| `log.NewTLog("ManagerAudit")` | `octo-lib/pkg/log/logger.go` — creates a `*TLog` that delegates to global zap logger with `【ManagerAudit】` prefix | ✅ Yes — stateless wrapper, safe for concurrent use |

- **C1 — Dual-path parity:** N/A. This is a logging-only middleware with no state mutations. No add/remove, subscribe/unsubscribe, or create/delete symmetry to check.
- **C2 — Control-flow ordering / nesting reuse:** Clear. `c.Next()` is called exactly once per request (either in the early-return branch or the logging branch, never both). `ScrubPath` is a pure function with no nesting/reuse hazards. No security controls (regex/escape/sanitize) are applied to manager paths that need non-canonical-form testing.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This middleware does not grant or restrict access. It observes after the fact.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. This middleware does not perform authorization. It intentionally captures both authenticated and unauthenticated attempts for forensics (documented behavior).

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 issues found. The implementation is clean, security-conscious (no body reads, no query string logging, path scrubbing applied), and well-tested. Two minor P2 nits (doc/code terminology mismatch, redundant `ts` field) do not block landing.

lml2468:
## CTO security review — SIGN-OFF (observability-before-launch)

> Note: GitHub blocks a formal *Approve* here because this PR was opened under the same bot identity I act as. This comment is my security owner sign-off; merge once the bot checks below are green.

Reviewed as security owner. Verified the load-bearing correctness chain end-to-end, not just the test:

**Operator id is real in production, not just the test fake.** The test sets `c.Set("uid", ...)` itself, so I traced the actual auth path in octo-lib:
- `AuthMiddleware` sets `c.Set("uid", info.UID)` (`pkg/wkhttp/http.go:331`).
- `wkhttp.Context` embeds `*gin.Context` (`http.go:70`), so the value lands in the shared `gin.Context.Keys` map.
- The middleware reads `c.GetString("uid")` from that same map; `GetLoginUID()` reads the identical `"uid"` key (`http.go:140-146`). So `operator_id` is genuinely populated on authed manager writes.
- `UseGin` → `r.Use` registers globally and runs before the route group, and reading after `c.Next()` correctly captures both the auth-set uid and the real response status — including 4xx rejected attempts with empty `operator_id`. Confirmed against a real registered manager group (`modules/report/api_manager.go` uses `l.Group("/v1/manager", l.AuthMiddleware(...))`).

**Security-by-default holds:**
- No request body is ever read → password routes (`resetpassword`/`updatepassword`) can't leak secrets.
- `target` is the `ScrubPath`'d URL path only, never the query string.
- secret-scan (gitleaks) + OSV + dependency-review all green.

**Scope is correct for Phase 1:** logging-only, no schema/migration; format documented in `docs/operator-audit-log.md` with a stable `audit=manager_admin_action` discriminator for Phase 2 ingest.

**Acceptance criteria met:** operator id + action + target + timestamp on every manager write; representative-mutation test asserts `operator_id`; format documented.

### Merge gating (not blocking my sign-off)
- `Test` + `code-review` bot: still running — must go green before merge (standard CI-green-or-it-didn't-happen).
- `check-sprint`: red because issue #367 has no Sprint set; needs org-Projects write scope my token lacks. Delegated to **OCT-28 → DevOps** to set #367 → Sprint W25 and re-run.

LGTM from a security/observability standpoint. Mergeable once Test + code-review + check-sprint are green.

yujiawei (approved):
# Code Review — PR #408 (octo-server)

**Scope reviewed:** head SHA `cec5f738f1159422c4fc99684c81e9aac8a336ed`, merge-base `c793943b` against `main`. 4 files / +313. Verified locally: `go build ./...` ✅, `go vet ./pkg/accesslog/` ✅, `go test ./pkg/accesslog/` ✅, `gofmt` clean.

## Verdict: APPROVED

Phase 1 is correct and well-scoped: one chokepoint global middleware emits a single structured line per mutating `/v1/manager` request, with sensible security-by-default (no bodies, no query string). The findings below are improvements, not merge blockers.

## What I verified (the load-bearing cross-file dependency)

The whole design hinges on `OperatorAuditMiddleware` (a raw `gin.HandlerFunc` reading `c.GetString("uid")`) seeing the value that `AuthMiddleware` sets via `c.Set("uid", …)` on a *different* wrapper type (`wkhttp.Context`). I traced this end-to-end and it holds:

- `wkhttp.Context` embeds `*gin.Context`, and `WKHttpHandler` sets `hc.Context = c` (the same underlying `*gin.Context`). So `c.Set("uid", …)` inside the matched group writes into the exact Keys map the global middleware later reads. ✅
- Ordering is correct: the global middleware runs first, calls `c.Next()`, the group's `AuthMiddleware` sets `uid` during `Next`, and the audit line is emitted after `Next` returns. ✅
- Rejected attempts are captured: `AuthMiddleware` does `RenderError(... 401) + c.Abort()`, control returns up the chain, and the audit line is still emitted with empty `operator_id` + a 4xx status — matching the documented forensic contract. ✅

`isManagerWrite` segment-boundary matching (`== prefix || HasPrefix(prefix+"/")`) correctly rejects `/v1/managerx/...`, and the test table covers it.

## Findings

### P2 — Audit line is skipped if a downstream handler panics
`pkg/accesslog/operator_audit.go:62-72`

The `lg.Info(...)` call sits after `c.Next()` but is **not** in a `defer`. `gin.Recovery()` is registered outermost (inside `wkhttp.New`), so a panic in a manager handler unwinds *past* this function — the request returns 500 (recovered) but emits **no** audit line. An authenticated admin write that mutates state and then panics would not appear in the trail, which slightly undercuts the "one line per mutating request" guarantee.

Note this is **not new/worse than the surrounding code**: the existing gin access logger registered right above (`main.go:155-163`) is also non-deferred, so the audit middleware merely matches current house behavior. Low real-world likelihood (requires an authenticated admin + a reachable panic). Recommended fix — make the emission panic-safe:

```go
c.Next()  // becomes:
defer func() {
    lg.Info(AuditLogMessage, /* …fields, c.Writer.Status() reads the recovered status… */)
}()
c.Next()
```

### P2 — `target` logs raw path identifiers; `ScrubPath` does not cover `/v1/manager/*`
`pkg/accesslog/operator_audit.go:69`

`ScrubPath` only masks `/v1/incoming-webhooks/<id>/<token>`; for manager paths it is a pass-through. So `target` records the **raw** client path including dynamic segments — e.g. `/v1/manager/spaces/:space_id/invites/:code` logs the actual invite code, and `/v1/manager/robots/:robot_id/revoke_token` logs the robot id. Invite codes are credential-like.

Caveat for the author: the existing access logger already logs these same `ScrubPath(path)` values, so this PR does **not** introduce new exposure — it inherits the status quo. Still worth tightening for an audit-specific sink. Cleanest fix is to prefer the route **template** over raw values:

```go
target := c.FullPath()            // "/v1/manager/spaces/:space_id/invites/:code"
if target == "" {                 // 404 / no matched route
    target = ScrubPath(c.Request.URL.Path)
}
```

This also makes Phase 2 ingest cleaner (stable template strings, no high-cardinality raw ids), matching the doc's stated intent of capturing the target *surface*.

### P2 (nit) — prefer `c.GetLoginUID()` / the documented accessor over `c.GetString("uid")`
`pkg/accesslog/operator_audit.go:67`

Functionally correct today: `AuthMiddleware` sets `uid` as a `string` (`UserInfo.UID string`), and `GetString` returns it fine. But `docs/operator-audit-log.md` documents the source as `c.GetLoginUID()`, and the audit module reaches for the gin-native `GetString` instead. Since the middleware is a plain `gin.HandlerFunc` it can't call the `wkhttp.Context` helper directly — fine — but a brief comment noting "uid is set as a string by wkhttp.AuthMiddleware" would prevent a future silent-empty regression if the key's type ever changes. No code change required.

## Coverage / blind spots checked
- Sibling-PR semantic conflict: `go build ./...` on the merged tree is green. ✅
- All 10 `/v1/manager` route groups across modules (user/message/group/space/robot/backup/common/report/integration/workplace) sit under this single prefix → the one chokepoint genuinely covers them. ✅
- `/v1/manager/login` (the only unauthenticated manager route) is a POST, so it *is* audited with empty `operator_id` — acceptable (login attempts are forensically interesting), no secret leak since the password rides in the body which is never read. ✅
- Not changed, intentionally out of scope (Phase 2): durable storage, body-side target detail, log-volume/PII retention policy.

## Security note (security-sensitive PR)
Nothing here weakens auth. The two security-relevant items above (panic-skip fail-open; raw path ids in `target`) are both pre-existing-behavior-equivalent and non-blocking, but a human owner reviewing the audit-logging contract for Phase 2 should decide whether the `c.FullPath()` template switch and the `defer` should land in Phase 1 or Phase 2.

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD cec5f73

## Reviewer Summary
|---|---|---|
| qa-engineer | `pass-with-risk` | AC-1 delivered + well-tested (5/5 green); `/v1/manager/login` audit-noise line + deferred before/after |
| security-engineer | `cleared-with-risk` | Additive logging-only, 0 new deps/endpoints, no secret exposure; 2 low audit-completeness gaps |
| code-reviewer | `needs-discussion` | Clean single chokepoint, tests faithful to real auth; login-line semantics + scope Question need a decision |

## Critical Findings (P0-P1)
无 Critical / High priority findings — no blocker across all three reviewers.

## Should-Fix (P2)
- [P2-1] `pkg/accesslog/operator_audit.go:60,86` (+ `modules/user/api_manager.go:60-62`, `docs/operator-audit-log.md:33-39`) — `POST /v1/manager/login` is registered on the **pre-auth** group but matches `isManagerWrite`, so a *successful* login emits `operator_id=""` + 2xx. The doc defines empty-operator only for the 4xx rejected-attempt case, so empty+2xx is an undocumented third state that reads as noise/ambiguity for Phase-2 ingest. Not a disclosure risk (no body read; 4xx failed-login auditing is useful brute-force forensics). **Flagged by qa/S-1 + security/P-1 + code/S-1 (triple consensus).** Fix: exclude `/v1/manager/login` from `isManagerWrite`, or document empty-operator+2xx as the explicit "pre-auth action" case for Phase-2 classification.

## Suggestions (P3)
- [P3-1] `pkg/accesslog/operator_audit.go:64-72` — audit emit runs after `c.Next()` without `defer`; a downstream manager-handler panic (recovered by gin.Recovery) skips the audit line, so panic-induced 500s go unaudited (the sibling access logger wraps its emit). Low severity, non-exploitable-for-gain, but weakens the audit-completeness guarantee. **Flagged by security/P-2 + qa/N-1 + code/N-1.** Fix: `defer` the emit for panic-path parity.
- [P3-2] `pkg/accesslog/operator_audit.go:71` — `ts` (RFC3339) duplicates zap's own timestamp; harmless, doc acknowledges it. **Flagged by qa/N-2 + code/N-2.** Fix: optional — drop `ts` or keep for self-contained lines.

## Questions for Author
- [Q-1] #367 AC-1 reads "logs the operator **(+ before/after where meaningful)**". The middleware cannot capture before/after (never reads bodies/handler state); the PR defers this to Phase 2. Confirm with the #367 owner that operator+action+target+status satisfies Phase-1 sign-off and before/after is acceptably deferred. **From qa/Q-1 + code/Q-1** — scope confirmation, not a defect.

## Verdict Rationale
All three reviewers cleared the PR of blockers: 0 Must-Fix (code), qa `pass-with-risk`, security `cleared-with-risk`. The mandatory AC-1 clause ("management writes log the operator") is fully delivered — one global chokepoint covers every `/v1/manager` subgroup, future routes, and rejected attempts — and the single new production path is tested (5/5) and green on CI + local re-verify. There are no P0/P1 findings and only one P2 (the login-line semantics, unanimously flagged by all three) plus two P3 suggestions.

The Formal Verdict is `NEEDS_DISCUSSION` rather than `APPROVE` because there is one material open Question (Q-1: whether before/after detail may be deferred to Phase 2 needs the issue owner's confirmation for Phase-1 sign-off) and one P2 that is a genuine *decision* rather than a mechanical fix (P2-1: exclude the login route vs. document the empty+2xx state). It is not `REQUEST_CHANGES` because no reviewer returned fail/blocked/changes-requested, there are no P0/P1 findings, and the P2 count (1) is well under the ≥3 REQUEST_CHANGES threshold. No cross-reviewer disagreement: qa and security independently reached the same two items with matching root causes, and the code review added no net-new defects (contrarian pass: 0 new).

## Round Context

---

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD cec5f738

## Reviewer Summary

|---|---|---|
| qa-engineer | pass-with-risk | Operator logging works end-to-end and is well-tested (value-level assertions); the `before/after` AC clause is deferred to Phase 2 and panic-path writes escape the trail. |
| security-engineer | cleared-with-risk | Purely observational, logging-only feature; `operator_id` comes from the server-verified auth context, bodies are never read, path is scrubbed — no new exploitable risk; 1 non-exploitable panic-path forensic blind spot (Low). |
| code-reviewer | approved-with-nits | Clean, additive single-chokepoint design; 0 Must Fix / 0 Should Fix, 5 nits + 1 scope question. |

## Critical Findings (P0-P1)

## Should-Fix (P2)

无 — no reviewer raised a Must Fix or Should Fix; all findings below are suggestions/nits.

## Suggestions (P3)

- [P3-1] `pkg/accesslog/operator_audit.go:66` — Panic-path manager writes escape the audit trail: `gin.Recovery()` is registered outermost (octo-lib `wkhttp.New`, before `route.UseGin`), so a handler panic unwinds past the post-`c.Next()` `lg.Info(...)` before Recovery converts it to 500 → that write is not audited (normal `c.Status` 5xx still is). Flagged by qa/Risk-2 + security/Risk-1 + code/nit (all rate it non-blocking for Phase 1). Fix: wrap the emit in a `defer` so it fires on the panic unwind, or explicitly accept and carry to Phase 2.
- [P3-2] `pkg/accesslog/operator_audit.go:71` — `ScrubPath` is effectively a no-op for every `/v1/manager` path (it only masks the `/v1/incoming-webhooks/{id}/{token}` shape), so the PR body / `docs/operator-audit-log.md` framing of `target` as "scrubbed … even a path that embeds a token is masked" overstates the protection for the audited surface. No leak today (manager paths carry non-secret resource ids; query string is never logged). Flagged by code/nit. Fix: soften the doc wording, or extend `ScrubPath` if a future manager route may embed a secret in the path.
- [P3-3] `pkg/accesslog/operator_audit.go:71` — Audit line carries no trace/span id though a tracer middleware runs upstream; threading it in would ease Phase-2 correlation. Flagged by qa/obs-1 + code/nit. Fix: thread the request span/trace id into the line.
- [P3-4] `pkg/accesslog/operator_audit_test.go:57` — Full `ServeHTTP` end-to-end path is exercised for `POST` only; `PUT`/`PATCH`/`DELETE` are covered solely by the `TestIsManagerWrite` unit table (middleware logic is method-agnostic past the guard, so risk is low). Flagged by qa/test-1 + code/nit. Fix: add one non-POST end-to-end case.
- [P3-5] `pkg/accesslog/operator_audit_test.go:44` — Operator-id propagation is proven only by an in-suite fake that mirrors `c.Set("uid", …)`; both QA and code hand-traced the key equivalence against the real `AuthMiddleware` (holds today), but a future octo-lib key rename would silently empty `operator_id` with tests still green. Flagged by qa/Risk-3 + code/nit. Fix: reference a shared key constant at compile time, or add an integration test under the real `AuthMiddleware`.
- [P3-6] Audit-line persistence integrity — Phase 1 writes to standard zap logs only, without tamper-evident / append-only guarantees; an actor with host or log-write access could alter/delete forensic lines. This is an inherent property of log-based auditing; durable, tamper-evident storage is explicitly Phase 2 (centralized `admin_action` table). Flagged by security/Risk-2. Not a defect in this PR — registered so Phase 2 picks it up.

## Questions for Author

- [Q-1] Issue #367's Phase-1 checklist lists "logs the operator **(+ before/after where meaningful)**"; the middleware records operator/action/target/status but no before/after state. Confirm the `before/after` sub-clause is intentionally deferred to the Phase-2 audit table (the PR body implies it, but there is no author confirmation on record). Raised independently by qa + code.

## Verdict Rationale

No P0/P1/P2 findings survived consolidation: all three reviewers independently converged that the mandatory acceptance criterion — an operator is logged on every mutating `/v1/manager` request — is fully implemented, wired in `main.go`, and tested with concrete value-level assertions (operator id, action, target, status), and that operator-id provenance is sound (read from the server-verified auth context, not the request; bodies never read, so password routes cannot leak). Every remaining item is a nit or a Low/non-exploitable residual risk. The single cross-reviewer residual risk is the panic-path forensic gap [P3-1] (QA Risk 2 / Security Risk 1 / Code nit); all three rate it non-blocking for Phase-1 logging.

However, one **material open Question [Q-1]** survives unresolved: whether the "before/after where meaningful" sub-clause — listed under issue #367's Phase-1 checklist but not implemented — was intentionally descoped to Phase 2. The PR body implies the deferral, but no author confirmation is on record, and the question concerns Phase-1 AC coverage. Per the deterministic first-match rule (no P0/P1, ≤2 P2, ≥1 material Question → NEEDS_DISCUSSION), the Formal Verdict is `NEEDS_DISCUSSION`: there are zero blocking findings, but the AC-scope question should be answered before merge. No reviewer disagreements to record — the three reports are mutually consistent.

## Round Context

---

lml2468 (commented):
**NEEDS_DISCUSSION** — Round 1 (HEAD cec5f738)

Zero blocking findings across QA / Security / Code — all three converged that the mandatory AC (operator logged on every `/v1/manager` write) is implemented, wired, and tested with value-level assertions. One material open question remains: whether the "before/after where meaningful" sub-clause listed under #367 Phase 1 was intentionally deferred to Phase 2. Please confirm before merge.

Top residual (non-blocking): panic-path manager writes escape the audit trail (`operator_audit.go:66`, flagged by all three).

Full findings + rationale: https://github.com/Mininglamp-OSS/octo-server/pull/408#issuecomment-4864191975

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD cec5f738

**Formal Verdict**: `COMMENT`



## Reviewer Summary

|---|---|---|
| qa-engineer | pass-with-risk | Core AC (每个 manager 写操作记录 operator) fully implemented, strongly tested, CI-green, production path source-verified; only the `+ before/after where meaningful` sub-clause is deferred to Phase 2 and needs human ratification. |
| security-engineer | cleared-with-risk | Observe-only log path; `operator_id` from server-verified token (unforgeable), bodies never read, query never logged, webhook tokens scrubbed. Highest severity Low; 2 accepted residual risks routed to Phase 2. |
| code-reviewer | approved | Single global chokepoint covers all current + future manager subgroups; purely additive, isolated, strongly tested, clean across build/vet/gofmt/lint/gitleaks/semgrep. 0 Must / 0 Should, 2 cosmetic nits. |

## Critical Findings (P0-P1)

## Should-Fix (P2)

无 P2 findings — QA reported 0 should-fix, security required no fixes for merge, code-reviewer reported 0 should-fix.

## Suggestions (P3)

- [P3-1] `pkg/accesslog/operator_audit.go:69` — the explicit `ts` field uses `time.RFC3339` (second precision); two admin actions in the same wall-clock second get identical `ts` and cannot be ordered by this field alone. Cosmetic — zap's own line timestamp and the Phase 2 table are the ordering authority. Fix: `time.RFC3339Nano` makes the line self-sufficient at zero cost. (flagged by code/Nit-1)
- [P3-2] `pkg/accesslog/operator_audit.go:69` — the explicit `ts` field duplicates the timestamp zap already stamps on every line. Documented as intentional (a stable RFC3339 field for Phase 2 to key off, independent of the logger's format); flagging only so the Phase 2 consumer knows the two timestamps coexist by design. (flagged by code/Nit-2)
- [P3-3] Info Disclosure (Low, Phase 2 follow-up) — `ScrubPath` currently scrubs only the incoming-webhook token prefix; `/v1/manager/*` path segments carrying PII (if any route ever uses phone/email as a path param) would be logged verbatim. Current manager routes use opaque uid/group-id, so residual risk is low. Fix: Phase 2 audit-table design should scrub/grade sensitive path segments and apply stricter access/retention to the audit stream. (flagged by security/R1)
- [P3-4] DoS (Low, Phase 2 follow-up) — unauthenticated/forged `POST /v1/manager/*` (incl. 404s) produces "empty operator_id + 4xx" audit lines, a valid forensic signal that could also be abused to amplify log-write volume. Mitigated today by the existing global per-IP `RateLimitMiddleware`. Fix: Phase 2 may add write rate-limiting/sampling or alert thresholds on the audit stream. (flagged by security/R2)

## Questions for Author

- [Q-1] Confirm that issue #367 Phase 1's AC clause `(+ before/after where meaningful)` is *intentionally* deferred to the Phase 2 centralized audit table rather than missed this round. `docs/operator-audit-log.md` records it as deferred; QA's `pass-with-risk` and the PR's own "CTO security review before merge" checkbox both request explicit human/security sign-off on this Phase-1/Phase-2 scope split. (from qa/Q1)

## Verdict Rationale

All three reviewers converged with no blocking findings: QA `pass-with-risk`, Security `cleared-with-risk` (highest severity Low), Code `approved` (0 Must Fix, 0 Should Fix, 2 cosmetic nits). Consolidated counts are **0 P0, 0 P1, 0 P2, 4 P3, 1 Question** — no code change is required to merge, and there were **no cross-reviewer disagreements** (all three independently confirmed the production `operator_id` path via octo-lib source, the correct middleware registration order at `main.go:166` ahead of `module.Setup`, and the security-by-default design: bodies never read, query never logged, path scrubbed).

The verdict is `COMMENT` rather than `APPROVE` because of one material open question (Q-1): the AC sub-clause `+ before/after where meaningful` is intentionally and documentedly deferred to Phase 2, but that scope split has not yet been ratified by a human/security owner — QA explicitly conditioned its pass on that ratification, and the PR carries a "CTO security review before merge" gate. Per the deterministic first-match verdict rule (no P0/P1, ≤2 P2, ≥1 material Question → `COMMENT`), this is a clean bot-side review that surfaces a decision for the human approver rather than an automatic approval. Human sign-off remains required before merge.

## Round Context

---

lml2468 (commented):
**COMMENT — Round 1**

All three reviewers (QA, Security, Code) converged on a single non-blocking Should-Fix item — new zip/binary upload extensions lack `fileMagicNumbers` entries, so the main upload path skips content-consistency validation. No P0/P1; recommend tracking the magic-number hardening in a follow-up issue rather than blocking this additive whitelist expansion.

Full findings: https://github.com/Mininglamp-OSS/octo-server/pull/461#issuecomment-4883083023

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD cec5f738

**Formal Verdict**: `COMMENT`



## Lens Summary

| Lens | Verdict | Headline |
|---|---|---|
| code | approved | Correct, well-tested audit middleware; only doc-accuracy nits. |
| security | cleared-with-risk | Security-by-default claims all hold; 3 low, non-exploitable risks (intentional/Phase-2). |
| qa | pass-with-risk | Tests present & green (ran locally 14/14); a few coverage/assertion gaps. |

## Critical Findings (P0-P1)

无 Critical / High priority findings — all three lenses converged: production code is correct (uid key + middleware ordering + JSON-encoded fields all verified against octo-lib source at this SHA), CI green, no exploitable security issue.

## Should-Fix (P2)

- `[P2-1]` `pkg/accesslog/operator_audit_test.go:88` — the `ts` field (documented RFC3339 UTC, and Phase-2 ingest keys off this format) is asserted for presence only, not format/timezone (flagged by qa/Q-3) — Fix: assert `time.Parse(time.RFC3339, ts)` succeeds and location is UTC.
- `[P2-2]` `pkg/accesslog/operator_audit_test.go:129` — the `status` field is only verified for 200; the rejected-attempt test asserts empty `operator_id` but not the paired 4xx status, which is exactly the pairing the doc comment says gives an empty operator meaning (flagged by qa/Q-4) — Fix: have the rejected test return 401/403 and assert the `status` field equals it.

## Suggestions (P3)

- `[P3-1]` `pkg/accesslog/operator_audit.go:73` (+ `docs/operator-audit-log.md:52`) — `target` is run through `ScrubPath`, but `ScrubPath` only masks the `/v1/incoming-webhooks/{id}/{token}` segment; for manager paths it is a pure passthrough. Doc/comment ("even a path that embeds a token is masked") slightly overstates the protection for the manager surface (flagged by code/C-1) — Fix: reword to note ScrubPath is a no-op for manager paths today, applied for consistency/future-proofing.
- `[P3-2]` `pkg/accesslog/operator_audit.go:79` — `ts` uses `time.RFC3339` (second precision) and duplicates the logger's own `ts` at coarser resolution (flagged by code/C-2) — Fix: optional — use `time.RFC3339Nano` or drop the field and rely on the logger's ts.
- `[P3-3]` `pkg/accesslog/operator_audit_test.go:57` — end-to-end write path is tested for POST only; PUT/PATCH/DELETE are covered only in the `isManagerWrite` unit table, not through `ServeHTTP` (flagged by qa/Q-2) — Fix: add a table-driven e2e case over all four verbs asserting one line + correct `action` each.
- `[P3-4]` `pkg/accesslog/operator_audit.go:64` — audit line has no source IP / request-id / before-after state, so the forensic record is partial (flagged by security/S-1) — Fix: none for Phase 1; #367 scopes `ip` and richer capture to the Phase-2 audit table. Accepted risk.
- `[P3-5]` `pkg/accesslog/operator_audit.go:73` — `target` embeds internal resource ids (`:uid`, `:robot_id`, `:group_no`); these are non-secret correlation data, not direct PII (no phone/email in any manager write path template) (flagged by security/S-2) — Fix: none; extend ScrubPath only if a future manager route puts PII directly in the path.
- `[P3-6]` `pkg/accesslog/operator_audit.go:64` — unauthenticated/4xx attempts are audited too (empty operator_id), so an attacker could spray `POST /v1/manager/<random>` for log-volume amplification; mitigated by the existing global per-IP rate-limit floor (flagged by security/S-3) — Fix: none; capturing rejected attempts is an intentional forensic feature and the rate-limit floor bounds volume.

## Questions for Author

- `[Q-1]` Issue #367's Phase-1 AC literally asks for "operator (+ before/after where meaningful)". The middleware records operator/action/target/status but never before/after values — a deliberate security choice (bodies are never read). Please confirm the issue owner agrees middleware-level logging satisfies Phase 1 and that before/after is deferred to the Phase-2 audit table, so this isn't read as an unmet AC (from qa/Q-1).

## Verdict Rationale

`COMMENT`: no P0/P1, no security `blocked`, no qa `fail`, no code Must Fix — all three lens verdicts are passing (approved / cleared-with-risk / pass-with-risk). Two should-fix test-hardening items (P2) and one material AC-clarification question keep it from a clean `APPROVE`, but nothing blocks merge. The production code was verified correct against octo-lib source at this SHA (the load-bearing "uid key" and "middleware ordering" concerns were both checked and cleared).

## Round Context

ROUND=1: First-round review — no prior round context.

|---|---|---|
| pkg/accesslog/operator_audit.go | Yes (code+security+qa) | P3-1, P3-2, P3-4, P3-5, P3-6 |
| pkg/accesslog/operator_audit_test.go | Yes (code+qa) | P2-1, P2-2, P3-3 |
| main.go | Yes (code+qa) | — (ordering verified correct) |
| docs/operator-audit-log.md | Yes (code) | P3-1 |

## Notes

- Confidence filter aggregate: 1 dropped (code lens dropped a speculative "operator_id always empty" finding after confirming the uid key + ordering against octo-lib source), 0 downgraded.
- Security: no `blocked` finding → no embargo comment. Credential scan and dep scan both clean.
- QA ran the tests locally: `OperatorAudit` 4/4 + `IsManagerWrite` 10/10 pass; CI rollup green (Build/Test/Vet/Lint SUCCESS).
- Terminal chain: this comment is the last artifact for Round 1; a next round is triggered by the spawner only on a new HEAD SHA.

---

lml2468 (commented):
**review-agent · Round 1 · verdict COMMENT**

No blockers. Production code verified correct against octo-lib source at this SHA (uid key + middleware ordering + JSON-encoded fields all checked); CI green; no exploitable security issue. Two P2 test-hardening items (assert `ts` RFC3339 format/UTC; assert paired 4xx `status` on the rejected-attempt test) and one AC-clarification question (before/after deferred to Phase 2?). Full breakdown in the aggregate comment.

PR #409, opened 2026-06-17
feat(security): CI source-guard requiring a role check on every /v1/manager route

### LINKED ISSUE #366: authz: centralized authorization layer + CI guard requiring a role check on every /v1/manager route

lml2468:
## Summary

octo-server#366 Part 1. Authz on the admin `/v1/manager` surface is enforced per-handler by an inline `c.CheckLoginRoleIsSuperAdmin()` / `c.CheckLoginRole()` call — there is **no central policy middleware yet** (that's the SecurityEngineer's systemic workstream). With the check scattered across ~100 handlers, a new `/v1/manager` route that forgets the check silently ships a **privilege-escalation hole**: `AuthMiddleware` only authenticates, it does not gate admin role.

This PR is a thin, reversible CI safety net (no refactor): it makes a missing role check a **build failure**.

- **`tools/lint-manager-authz`** — an AST guard that fails if any handler registered on a `/v1/manager` route group lacks a role check.
  - Route-group variables are resolved **function-locally**, so the admin group doesn't conflate with unrelated `/v1` or `/v1/space` groups that reuse names like `auth` (this was a real false-positive trap — see the regression test).
  - A handler counts as covered if it (or a same-package helper it calls) invokes `CheckLoginRole` / `CheckLoginRoleIsSuperAdmin`, or its group has a role-enforcing middleware, or it's allowlisted.
  - **Fail-closed**: a handler that can't be resolved is treated as unchecked.
- **`allowlist.txt`** — the 5 routes that live under `/v1/manager` by path but are intentionally not admin:
  - `POST /v1/manager/login` (public, issues the admin token pre-auth).
  - `/v1/manager/secrets` CRUD ×4 (user-scoped, `owner == c.GetLoginUID()`, not an admin surface).
- **CI**: new `manager-authz-lint` job (mirrors `personal-msgsendreq-lint`) — runs the guard's unit tests, then the guard over `./modules`.
- **Docs-as-interface**: convention documented in the tool's package doc, the failure message, the allowlist header, and `CLAUDE.md`.

Inventory today: **109** `/v1/manager` routes, all role-checked or allowlisted.

### How to satisfy the guard for a new `/v1/manager` route

Call a role check at the top of the handler before doing any work:

```go
if err := c.CheckLoginRoleIsSuperAdmin(); err != nil { c.ResponseError(err); return }
```

(use `CheckLoginRole` for the admin∪superAdmin tier). If the route is deliberately **not** an admin route (public, or user-scoped by UID), add it to `tools/lint-manager-authz/allowlist.txt` with a one-line reason.

## Test plan

- [x] `go test ./tools/lint-manager-authz/...` — 9 cases incl. the function-local-scoping regression, delegated check, subgroup composition, allowlist suppression, and the **unchecked-route fixture** that proves CI fails.
- [x] `go run ./tools/lint-manager-authz ./modules` → green (109 routes).
- [x] Live proof: injected an unchecked `GET /v1/manager/backup/leak` into a real module → guard exits 1; reverted → exits 0.
- [x] Confirmed the 5 allowlist entries are exactly the routes flagged with an empty allowlist (no stale/over-broad entries).
- [x] `gofmt` / `go vet` clean; `ci.yml` YAML validated.
- [ ] CTO review as security owner before merge. Consider adding `Manager Authz Lint` to required status checks.

Scope guard: this is **Part 1 only** — no central authz refactor (held for the SecurityEngineer).

Closes #366

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -306,6 +306,36 @@ jobs:
       - name: Reject direct PERSONAL MsgSendReq literals
         run: go run ./tools/lint-personal-msgsendreq ./modules
 
+  manager-authz-lint:
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
+    # Mininglamp-OSS/octo-server#366 Part 1: source-guard for the admin
+    # (/v1/manager) authorization surface. Authz there is enforced per-handler
+    # by an inline c.CheckLoginRole / c.CheckLoginRoleIsSuperAdmin call (no
+    # central policy middleware yet). A new /v1/manager route that forgets the
+    # check would silently ship a privilege-escalation hole — AuthMiddleware
+    # only authenticates, it does not gate role. This guard AST-walks ./modules
+    # and fails CI if any /v1/manager route handler lacks a role check, unless
+    # the route is allowlisted (tools/lint-manager-authz/allowlist.txt) with a
+    # reason. The tool's own unit tests run first.
+    name: Manager Authz Lint
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Test the guard
+        run: go test ./tools/lint-manager-authz/...
+
+      - name: Require a role check on every /v1/manager route
+        run: go run ./tools/lint-manager-authz ./modules
+
   i18n-extract-check:
     needs: [changes]
     if: |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -147,6 +147,7 @@ Tests require MySQL + Redis + WuKongIM running (see CI or `make env-test` in dmw
 - API routes: prefix `/v1/`
 - New modules: add blank import in `internal/modules.go`
 - Auth: all routes go through `AuthMiddleware` unless explicitly excluded — document why if skipping
+- Manager authz: every `/v1/manager` **admin** route handler must call a role check (`c.CheckLoginRoleIsSuperAdmin()` or `c.CheckLoginRole()`) before doing work — `AuthMiddleware` only authenticates, it does not gate admin role. The `manager-authz-lint` CI guard (`tools/lint-manager-authz`) fails the build on any `/v1/manager` route whose handler lacks a check. Routes that live under `/v1/manager` but are NOT admin (public login, user-scoped CRUD like `/v1/manager/secrets`) go in `tools/lint-manager-authz/allowlist.txt` with a one-line reason (octo-server#366 Part 1)
 - i18n: user-facing errors use `httperr.ResponseErrorL` + a registered `pkg/errcode` code; never raw `c.ResponseError`/`c.JSON`/`AbortWithStatusJSON`. Run `make i18n-extract-check` + `make i18n-lint` after touching codes (see Architecture › Error Handling & i18n)
 - Rate limiting: mount `SharedUIDRateLimiter` (auth routes) or `StrictIPRateLimitMiddleware` (unauth) — never hand-roll a Redis counter for request-frequency limiting (see Architecture › Rate Limiting)
 - Space isolation: handlers that access user data must go through Space middleware
--- a/tools/lint-manager-authz/allowlist.txt
+++ b/tools/lint-manager-authz/allowlist.txt
@@ -1,0 +1,16 @@
+# lint-manager-authz allowlist — routes under /v1/manager that intentionally
+# do NOT carry an admin role check. Format: "<VERB> <path>  # reason".
+# Every entry MUST have a reason. See tools/lint-manager-authz/main.go and
+# https://github.com/Mininglamp-OSS/octo-server/issues/366.
+
+# Public, pre-auth: issues the admin session token. The group has no
+# AuthMiddleware by design; role cannot be checked before login.
+POST /v1/manager/login  # public login endpoint, issues admin session token
+
+# User-scoped secret CRUD that happens to live under the /v1/manager/secrets
+# path. Authz here is AuthMiddleware (authenticated user) + owner==login-UID
+# scoping inside the handler, NOT an admin-role gate. Not an admin surface.
+POST   /v1/manager/secrets              # user-scoped (owner=UID), not admin
+GET    /v1/manager/secrets              # user-scoped (owner=UID), not admin
+PUT    /v1/manager/secrets/:secret_id   # user-scoped (owner=UID), not admin
+DELETE /v1/manager/secrets/:secret_id   # user-scoped (owner=UID), not admin
--- a/tools/lint-manager-authz/main.go
+++ b/tools/lint-manager-authz/main.go
@@ -1,0 +1,583 @@
+// Command lint-manager-authz is the octo-server#366 Part 1 source-guard. It
+// AST-walks the module packages and fails the build if any route registered on
+// a `/v1/manager` route group has a handler that does not perform a role/authz
+// check.
+//
+// Why this guard exists:
+//
+//	Authorization on the manager (admin) surface is enforced per-handler by an
+//	inline `c.CheckLoginRoleIsSuperAdmin()` / `c.CheckLoginRole()` call — there
+//	is no central policy middleware yet (that is the SecurityEngineer's systemic
+//	workstream). With the check scattered across ~100 handlers, a new
+//	`/v1/manager/...` route that simply forgets the check silently ships a
+//	privilege-escalation hole: `AuthMiddleware` only authenticates the caller,
+//	it does NOT gate on admin role. This guard makes that omission a CI failure
+//	instead of a production incident — a thin, reversible safety net that holds
+//	the line until the centralized authz layer lands.
+//
+// What counts as "checked":
+//
+//	A handler is covered if its body (or a same-package helper it calls, up to a
+//	small depth) invokes one of the approved role-check primitives, OR its route
+//	group is constructed with a role-enforcing middleware (see roleMiddleware),
+//	OR the route is explicitly allowlisted in allowlist.txt with a reason.
+//
+// How to satisfy the guard for a NEW /v1/manager route:
+//
+//	Call a role check at the top of the handler before doing any work, e.g.
+//	    if err := c.CheckLoginRoleIsSuperAdmin(); err != nil { c.ResponseError(err); return }
+//	(use CheckLoginRole for the admin∪superAdmin tier). If the route is
+//	deliberately NOT an admin route (public, or user-scoped by UID like
+//	/v1/manager/secrets), add it to tools/lint-manager-authz/allowlist.txt with
+//	a one-line reason.
+//
+// Usage:
+//
+//	go run ./tools/lint-manager-authz [root...]
+//
+// Defaults to scanning ./modules. Test files (*_test.go) are skipped.
+//
+// Exit codes: 0 (clean), 1 (uncovered route(s)), 2 (walk/allowlist error).
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+// managerPathPrefix is the route-group path that puts a handler in scope. It
+// matches the exact prefix and any deeper group (e.g. /v1/manager/workplace,
+// /v1/manager/dashboard, /v1/manager/secrets).
+const managerPathPrefix = "/v1/manager"
+
+// allowlistRelPath lives next to this command so `go run ./tools/...` finds it
+// regardless of the CI step's working directory.
+const allowlistRelPath = "tools/lint-manager-authz/allowlist.txt"
+
+// roleCheckNames is the set of *wkhttp.Context methods that count as a role/
+// authz gate. Keep this in sync with octo-lib's wkhttp.Context.
+var roleCheckNames = map[string]bool{
+	"CheckLoginRole":             true, // admin ∪ superAdmin
+	"CheckLoginRoleIsSuperAdmin": true, // superAdmin only
+}
+
+// roleMiddleware is the set of route-group middleware identifiers that enforce
+// a role by themselves, making a per-handler check unnecessary. Empty today —
+// the manager surface authenticates at the group (AuthMiddleware) and gates
+// role per-handler. Populated here so a future role-enforcing middleware is
+// recognized without touching the walk logic.
+var roleMiddleware = map[string]bool{}
+
+// httpVerbs is the set of route-registration method names on a route group.
+var httpVerbs = map[string]bool{
+	"GET": true, "POST": true, "PUT": true, "DELETE": true,
+	"PATCH": true, "HEAD": true, "OPTIONS": true, "Any": true, "Handle": true,
+}
+
+// maxHelperDepth bounds the transitive same-package helper search so a handler
+// that delegates its role check to a small wrapper still counts as covered,
+// without the cost (or cycles) of whole-program analysis.
+const maxHelperDepth = 2
+
+func main() {
+	roots := os.Args[1:]
+	if len(roots) == 0 {
+		roots = []string{"modules"}
+	}
+
+	allow, err := loadAllowlist(allowlistRelPath)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "load allowlist %q: %v\n", allowlistRelPath, err)
+		os.Exit(2)
+	}
+
+	pkgs, err := collectPackages(roots)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		os.Exit(2)
+	}
+
+	var violations []string
+	routeCount := 0
+	for _, pkg := range pkgs {
+		v, n := pkg.collectViolations(allow)
+		violations = append(violations, v...)
+		routeCount += n
+	}
+
+	sort.Strings(violations)
+	if len(violations) > 0 {
+		for _, v := range violations {
+			fmt.Println(v)
+		}
+		fmt.Printf("\nFound %d /v1/manager route(s) with no role check.\n", len(violations))
+		fmt.Println("Add a role check at the top of the handler, e.g.:")
+		fmt.Println("    if err := c.CheckLoginRoleIsSuperAdmin(); err != nil { c.ResponseError(err); return }")
+		fmt.Println("If the route is intentionally NOT an admin route (public, or user-scoped by UID),")
+		fmt.Println("add it to " + allowlistRelPath + " with a one-line reason.")
+		fmt.Println("See https://github.com/Mininglamp-OSS/octo-server/issues/366.")
+		os.Exit(1)
+	}
+
+	fmt.Printf("OK: all %d /v1/manager route(s) are role-checked or allowlisted (%d allowlist entr(ies)).\n",
+		routeCount, len(allow))
+}
+
+// route is a single registration on a manager route group.
+type route struct {
+	file                   string
+	line                   int
+	method                 string // HTTP verb
+	path                   string // full path including the group base
+	handler                ast.Expr
+	enclosingRecvType      string // receiver type of the func that registered the route
+	groupHasRoleMiddleware bool
+}
+
+func (r route) sig() string { return r.method + " " + r.path }
+
+func (r route) handlerName() string {
+	switch h := r.handler.(type) {
+	case *ast.SelectorExpr:
+		if h.Sel != nil {
+			return h.Sel.Name
+		}
+	case *ast.Ident:
+		return h.Name
+	case *ast.FuncLit:
+		return "<func literal>"
+	}
+	return "<unknown>"
+}
+
+// pkg holds the parsed files of a single Go package directory plus the indexes
+// the analysis needs.
+type pkg struct {
+	dir   string
+	fset  *token.FileSet
+	files []*ast.File
+	// funcsByName maps a function/method name to all its declarations in the
+	// package (handlers may collide across receiver types; resolution prefers a
+	// matching receiver type, see handlerHasRoleCheck). This is package-global
+	// on purpose: a route registered in api_manager.go may point at a handler
+	// method defined in another file of the same package.
+	funcsByName map[string][]*ast.FuncDecl
+}
+
+type groupInfo struct {
+	basePath  string
+	hasRoleMW bool
+}
+
+// collectPackages walks the roots and groups .go files by directory.
+func collectPackages(roots []string) ([]*pkg, error) {
+	byDir := map[string][]string{}
+	for _, root := range roots {
+		walkErr := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if info.IsDir() {
+				if name := info.Name(); name == "vendor" || name == "tools" || name == "testdata" {
+					return filepath.SkipDir
+				}
+				return nil
+			}
+			if !strings.HasSuffix(path, ".go") || strings.HasSuffix(path, "_test.go") {
+				return nil
+			}
+			dir := filepath.Dir(path)
+			byDir[dir] = append(byDir[dir], path)
+			return nil
+		})
+		if walkErr != nil {
+			return nil, fmt.Errorf("walk %q: %w", root, walkErr)
+		}
+	}
+
+	dirs := make([]string, 0, len(byDir))
+	for d := range byDir {
+		dirs = append(dirs, d)
+	}
+	sort.Strings(dirs)
+
+	var pkgs []*pkg
+	for _, dir := range dirs {
+		p := &pkg{
+			dir:         dir,
+			fset:        token.NewFileSet(),
+			funcsByName: map[string][]*ast.FuncDecl{},
+		}
+		for _, path := range byDir[dir] {
+			f, perr := parser.ParseFile(p.fset, path, nil, 0)
+			if perr != nil {
+				// go build / go vet is the canonical syntax gate; skip
+				// unparseable files rather than failing the lint on them.
+				continue
+			}
+			p.files = append(p.files, f)
+			for _, decl := range f.Decls {
+				if fn, ok := decl.(*ast.FuncDecl); ok {
+					p.funcsByName[fn.Name.Name] = append(p.funcsByName[fn.Name.Name], fn)
+				}
+			}
+		}
+		pkgs = append(pkgs, p)
+	}
+	return pkgs, nil
+}
+
+// localGroupVars resolves, within a single function body, the variables bound to
+// a `/v1/manager` route group. Route-group variables are function-local (every
+// Route() method reuses names like `auth`/`v`), so resolution MUST be scoped to
+// the function — a package-global index would conflate the admin group in
+// api_manager.go with an unrelated `/v1` or `/v1/space` group of the same name
+// in api.go. Subgroups derived from a manager group within the same body are
+// followed to a fixed point.
+func localGroupVars(body *ast.BlockStmt) map[string]groupInfo {
+	type pending struct {
+		name     string
+		recvExpr ast.Expr // the X in X.Group(...)
+		basePath string
+		hasMW    bool
+	}
+	var all []pending
+	ast.Inspect(body, func(n ast.Node) bool {
+		as, ok := n.(*ast.AssignStmt)
+		if !ok || len(as.Lhs) != 1 || len(as.Rhs) != 1 {
+			return true
+		}
+		lhs, ok := as.Lhs[0].(*ast.Ident)
+		if !ok {
+			return true
+		}
+		call, ok := as.Rhs[0].(*ast.CallExpr)
+		if !ok {
+			return true
+		}
+		sel, ok := call.Fun.(*ast.SelectorExpr)
+		if !ok || sel.Sel == nil || sel.Sel.Name != "Group" || len(call.Args) == 0 {
+			return true
+		}
+		lit := stringLit(call.Args[0])
+		if lit == "" {
+			return true
+		}
+		all = append(all, pending{
+			name:     lhs.Name,
+			recvExpr: sel.X,
+			basePath: lit,
+			hasMW:    hasRoleMiddleware(call.Args[1:]),
+		})
+		return true
+	})
+
+	groups := map[string]groupInfo{}
+	for changed := true; changed; {
+		changed = false
+		for _, pd := range all {
+			if _, done := groups[pd.name]; done {
+				continue
+			}
+			var base string
+			var inheritedMW bool
+			if strings.HasPrefix(pd.basePath, managerPathPrefix) {
+				base = pd.basePath
+			} else if recv, ok := pd.recvExpr.(*ast.Ident); ok {
+				parent, ok := groups[recv.Name]
+				if !ok {
+					continue // not a manager group (or parent not yet known)
+				}
+				base = parent.basePath + pd.basePath
+				inheritedMW = parent.hasRoleMW
+			} else {
+				continue
+			}
+			groups[pd.name] = groupInfo{basePath: base, hasRoleMW: pd.hasMW || inheritedMW}
+			changed = true
+		}
+	}
+	return groups
+}
+
+// collectViolations returns the uncovered-route messages for this package and
+// the total number of manager routes inspected. A route is covered if it is
+// allowlisted, its group carries a role-enforcing middleware, or its handler
+// performs a role check.
+func (p *pkg) collectViolations(allow map[string]bool) (violations []string, routeCount int) {
+	for _, rt := range p.managerRoutes() {
+		routeCount++
+		if allow[rt.sig()] || rt.groupHasRoleMiddleware {
+			continue
+		}
+		if p.handlerHasRoleCheck(rt.handler, rt.enclosingRecvType) {
+			continue
+		}
+		violations = append(violations,
+			fmt.Sprintf("%s:%d: %s %s -> handler %s has no role check (CheckLoginRole / CheckLoginRoleIsSuperAdmin)",
+				rt.file, rt.line, rt.method, rt.path, rt.handlerName()))
+	}
+	return violations, routeCount
+}
+
+// managerRoutes returns every route registered on a `/v1/manager` group,
+// analyzing each function independently so function-local group variables do
+// not leak across Route() methods.
+func (p *pkg) managerRoutes() []route {
+	var routes []route
+	for _, f := range p.files {
+		for _, decl := range f.Decls {
+			fn, ok := decl.(*ast.FuncDecl)
+			if !ok || fn.Body == nil {
+				continue
+			}
+			groups := localGroupVars(fn.Body)
+			if len(groups) == 0 {
+				continue
+			}
+			recvType := recvTypeName(fn)
+			ast.Inspect(fn.Body, func(n ast.Node) bool {
+				call, ok := n.(*ast.CallExpr)
+				if !ok {
+					return true
+				}
+				sel, ok := call.Fun.(*ast.SelectorExpr)
+				if !ok || sel.Sel == nil || !httpVerbs[sel.Sel.Name] {
+					return true
+				}
+				recv, ok := sel.X.(*ast.Ident)
+				if !ok {
+					return true
+				}
+				gi, ok := groups[recv.Name]
+				if !ok {
+					return true
+				}
+				if len(call.Args) < 2 {
+					return true
+				}
+				sub := stringLit(call.Args[0])
+				handler := call.Args[len(call.Args)-1]
+				pos := p.fset.Position(call.Pos())
+				routes = append(routes, route{
+					file:                   filepath.ToSlash(pos.Filename),
+					line:                   pos.Line,
+					method:                 sel.Sel.Name,
+					path:                   joinRoutePath(gi.basePath, sub),
+					handler:                handler,
+					enclosingRecvType:      recvType,
+					groupHasRoleMiddleware: gi.hasRoleMW,
+				})
+				return true
+			})
+		}
+	}
+	return routes
+}
+
+// handlerHasRoleCheck reports whether the handler expression (a func literal,
+// package func ident, or recv.method selector) performs a role check, directly
+// or via a same-package helper within maxHelperDepth.
+func (p *pkg) handlerHasRoleCheck(handler ast.Expr, recvType string) bool {
+	switch h := handler.(type) {
+	case *ast.FuncLit:
+		return p.bodyHasRoleCheck(h.Body, recvType, maxHelperDepth, map[string]bool{})
+	case *ast.Ident:
+		return p.namedFuncHasRoleCheck(h.Name, recvType, maxHelperDepth, map[string]bool{})
+	case *ast.SelectorExpr:
+		if h.Sel == nil {
+			return false
+		}
+		return p.namedFuncHasRoleCheck(h.Sel.Name, recvType, maxHelperDepth, map[string]bool{})
+	}
+	return false
+}
+
+// namedFuncHasRoleCheck resolves a function/method by name (preferring a
+// receiver-type match) and checks its body.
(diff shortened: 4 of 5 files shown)


lml2468:
## CTO security review — APPROVED (Part 1)

(Posting as a comment — GitHub blocks a formal approval since the PR is attributed to this token. Treat this as the security-owner sign-off for the OCT-20 / #366 review gate.)

Independently verified on branch `feat/oct20-manager-authz-guard` (go1.26.4):

**Verification (reproduced locally, not just trusting the PR):**
- `go test ./tools/lint-manager-authz/...` → pass (9 cases, incl. the unchecked-route acceptance fixture + the function-local-scoping regression).
- `go run ./tools/lint-manager-authz ./modules` → `OK: all 109 /v1/manager route(s) are role-checked or allowlisted (5 allowlist entries)`.
- **Fail-closed proof:** injected an unchecked `GET /v1/manager/backup/leak` into `modules/backup/api_manager.go` → guard exits 1 and pinpoints the route; reverted → exits 0.
- **Allowlist audited (the one place a real hole could hide) — sound:**
  - `POST /v1/manager/login` — registered on the no-middleware pre-auth group (`modules/user/api_manager.go:63`); role genuinely cannot be checked before login. Correct.
  - `/v1/manager/secrets` CRUD ×4 (`modules/usersecret/api.go`) — every handler (`create`/`list`/`update`/`delete`) scopes by `owner := c.GetLoginUID()`. User-scoped, not an admin surface. Correct, not over-broad.

**Design check:** function-local route-group resolution is the right call — the 9 manager modules all reuse `auth`/`v` names across `Route()` methods, so a package-global index would have flooded false positives. Fail-closed on unresolved handlers is the correct security default.

**Lenses:** blast-radius LOW (additive CI job + tool, no runtime/refactor, two-way door); security-by-default ✅; docs-as-interface ✅ (failure message + allowlist header + package doc + CLAUDE.md).

Scope correctly held to Part 1 — no central authz refactor (Part 2 stays with the SecurityEngineer, tracked on #366).

### Merge gate — one non-code blocker remains
All code/security checks are green (**Manager Authz Lint**, Build, Vet, Lint, Test). The only red check is **`check-sprint`**, an org board-hygiene gate. I added `Closes #366` to the PR body, which moved it past "no linked issue"; it now reports *"Linked issue #366 has no Sprint set."* #366 is not on the Octo Board (project #2).

**Action needed (Maintainer):** add #366 to the Octo Board with **Sprint W25** (or add this PR to the board as the gate's PR-fallback). My token lacks `read:project`, so I can't do it from here.

**On merge:** once `check-sprint` is green, merge, and add **Manager Authz Lint** to required status checks so the guard is enforced on every future PR.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#409 Review Report — `security(authz): CI source-guard requiring a role check on every /v1/manager route`
**Head SHA:** `ee7e24b71ae6e950a477190180128c2bc39c674a`
**Routing:** complexity=security_sensitive (automated review leg)

---

| Item | Status | Evidence |
|------|--------|----------|
| CI job mirrors `personal-msgsendreq-lint` pattern | ✅ | `.github/workflows/ci.yml` lines 309-338: identical `needs`, `if`, `runs-on`, `timeout-minutes`, pinned action SHAs |
| `changes` job dependency resolves | ✅ | `changes` job at ci.yml:42-78 exports `code` output; `GO_VERSION` at ci.yml:37 in workflow scope |
| Tool detects all 13 real `/v1/manager` groups | ✅ | Verified against 12 files in `modules/`: variable names (`auth`, `manager`, `user`, `mgr`) and receivers (`r`, `l`) all tracked by `localGroupVars` |
| Sub-prefix groups handled | ✅ | `/v1/manager/workplace`, `/v1/manager/dashboard`, `/v1/manager/secrets` matched by `strings.HasPrefix(pd.basePath, managerPathPrefix)` at main.go:243 |
| Delegated role checks (helper wrappers) | ✅ | `bodyHasRoleCheck` recurses into same-package helpers up to `maxHelperDepth=2`; covers `opanalytics.requireAdmin()` pattern (6 call sites) |
| Allowlist entries valid | ✅ | All 5 entries verified against source: `POST /v1/manager/login` (user/api_manager.go:64, pre-auth), 4× `/v1/manager/secrets` (usersecret/api.go:99-102, owner-scoped CRUD) |
| Allowlist path resolves in CI | ✅ | `os.Open("tools/lint-manager-authz/allowlist.txt")` CWD-relative; CI runs from repo root after `actions/checkout` |
| Test suite adequate | ✅ | 9 tests covering: unchecked flagging, both role-check tiers, function-local scoping regression, subgroup composition, delegated check, allowlist suppression, role middleware, non-manager sibling exclusion |
| CLAUDE.md addition consistent | ✅ | Line 150 bullet aligns with tool doc comment, CI comment block, and allowlist header |

## 2. Findings

### F1 — P2: `stringLit` uses `strings.Trim` instead of `strconv.Unquote` (main.go:370-375)

**Diff-scope:** new (this PR)

`strings.Trim(lit.Value, "` + "`" + `\"")` strips all leading/trailing characters in the cutset, which is not equivalent to Go string unquoting. A double-quoted group path containing Unicode escapes (e.g., `"\u002fv1\u002fmanager"`) would resolve to `/v1/manager` at runtime but fail to match in the tool.

**Practical risk:** Low. No current codebase usage of escape sequences in route paths. All routes use plain ASCII string literals.

**Recommendation:** Replace with `strconv.Unquote(lit.Value)` for correctness. One-line change.

### F2 — P2: Group passed as function parameter is invisible (design limitation)

**Diff-scope:** new (this PR — design limitation, not a bug in current codebase)

If a `Route()` method creates a manager group and passes it to a helper function, the helper's parameter is not tracked by `localGroupVars` (which only resolves local `:=` assignments from `.Group()` calls). Routes registered in the helper would be invisible.

```go
func (m *Manager) Route(r R) {
    auth := r.Group("/v1/manager", AuthMiddleware())
    m.registerExtra(auth)  // group escapes local scope
}
func (m *Manager) registerExtra(auth *RouteGroup) {
    auth.POST("/dangerous", m.dangerous)  // invisible to the tool
}
```

**Practical risk:** Low. No current `api_manager*.go` file uses this pattern. All 13 manager groups are consumed within their declaring function.

**Recommendation:** Consider a warning when a local group variable appears as an argument to a function call (group escapes local scope).

### F3 — P2: Variable shadowing could confuse first-assignment-wins tracking (design limitation)

**Diff-scope:** new (this PR — design limitation)

`localGroupVars` fixed-point loop uses first-assignment-wins (`if _, done := groups[pd.name]; done { continue }`). If a variable name is shadowed in an inner block with a different group path, the tool tracks the outer assignment.

**Practical risk:** Low. No current variable shadowing in any `Route()` method. The regression test (`TestFunctionLocalGroupScoping`) covers the cross-function case (same name `auth` in different methods), which is the real-world pattern.

### F4 — P2: `httpVerbs` map lacks Gin's `Match()` method (main.go:72-74)

**Diff-scope:** new (this PR — omission)

Gin's `IRoutes` interface exposes `Match(methods []string, path string, handlers ...HandlerFunc)`. Not in the map. A route registered via `auth.Match([]string{"GET","POST"}, "/path", handler)` would be invisible.

**Practical risk:** Very low. No `.Match(` usage found anywhere in `modules/`.

**Recommendation:** Add `"Match": true` to `httpVerbs` preemptively.

### F5 — P2: Allowlist path is CWD-relative (main.go:56, 555)

**Diff-scope:** new (this PR introduces the first allowlist-bearing lint tool)

`os.Open("tools/lint-manager-authz/allowlist.txt")` resolves CWD-relative. Works in CI (runs from repo root). A developer running from a subdirectory silently loads an empty allowlist (`os.IsNotExist` returns empty map), which would produce false positives rather than false negatives (fail-closed).

**Recommendation:** Consider `runtime.Caller(0)` or `os.Executable()` resolution for robustness.

### F6 — P2: `collectPackages` skips any directory named `tools` at any depth (main.go:165)

**Diff-scope:** new (this PR)

`filepath.SkipDir` on any directory named `tools`, `vendor`, or `testdata`. The `tools` skip is broader than needed (intended to exclude the project's own `tools/` directory). If a module creates a `tools/` subdirectory with Go files containing manager routes, those would be silently skipped.

**Practical risk:** Very low. No `tools/` subdirectory exists under `modules/`.

## 3. Severity Assessment (R1-R4 Rubric)

**R1 (P1 hard definition):** No finding makes a currently-working path unavailable, produces user-visible incorrect data, or makes the system worse than before the PR. The tool is a CI-time safety net, not a runtime authorization mechanism. All bypass vectors are theoretical (no current codebase pattern exploits them). The tool correctly covers all 109 existing routes.

**R2 (diff-scope):** All findings are `new` (introduced by this PR as design limitations of the new tool). None are `amplified` or `pre-existing`.

**R3 (distributed side-effects):** N/A — this is a CI lint tool, not a runtime system with fan-out/observer/emitter patterns.

**R4 (verdict mapping):** All findings are P2 (design limitations and robustness improvements). No P0 or P1. → **APPROVE**.

## 4. Additional Findings (out of scope but noted)

- Two existing routes use paths without leading slashes: `modules/message/api_manager.go:55` (`"message/sendfriends"`) and `modules/user/api_manager.go:79` (`"user/online"`). Gin auto-prepends `/`, so these work, but they are inconsistent with every other route. Pre-existing, not introduced by this PR.

## 5. Data-Flow Trace

```
allowlist.txt (on disk)
  └─ os.Open(allowlistRelPath)  [main.go:555]
     └─ loadAllowlist() → map[string]bool keyed by "<VERB> <path>"  [main.go:553-579]
        └─ main() → pkg.collectViolations(allow)  [main.go:103]
           └─ for each route: allow[rt.sig()] short-circuits violation  [main.go:267]
              └─ rt.sig() = "<Method> <full path>"  [main.go:137]

Route detection pipeline:
  collectPackages(roots) → walk dirs, parse .go files, index funcsByName
    └─ pkg.managerRoutes() → per FuncDecl:
         localGroupVars(fn.Body) → resolve .Group() assignments to groupInfo
           └─ fixed-point loop resolves subgroups from parent groups
         ast.Inspect(fn.Body) → find verb calls on known group variables
           └─ route{file, line, method, path, handler, enclosingRecvType, groupHasRoleMW}
    └─ pkg.collectViolations(allow) → for each route:
         allow[rt.sig()]? → skip
         rt.groupHasRoleMiddleware? → skip
         handlerHasRoleCheck(handler, recvType)? → skip (else violation)
           └─ FuncLit → bodyHasRoleCheck(body)
           └─ Ident/SelectorExpr → namedFuncHasRoleCheck(name, recvType, depth=2)
              └─ bodyHasRoleCheck → ast.Inspect for roleCheckNames selector calls
                 └─ recurse into helper calls if depth > 0
```

## 6. R5 Blindspot Checklist (security_sensitive — mandatory)

- **C1 (dual-path parity):** N/A — this PR adds a CI guard, not a symmetric add/remove or create/delete path. No paired operations to verify.
- **C2 (control-flow ordering / nested reuse):** N/A — the lint tool is a leaf CLI command. Its internals are not reused by other code paths. No nested composite calls or safety-control bypass risk.
- **C3 (authorization boundary ≠ capability boundary):** N/A — the tool enforces a coding convention on other code but does not itself expose a runtime capability or endpoint. No authorization boundary is modified.
- **C4 (authorization lifecycle / container-member cascade):** N/A — no runtime authorization path is modified. The tool is purely static analysis at CI time.

---

**[Octo-Q] verdict: APPROVE** — All findings are P2 (design limitations and robustness improvements for follow-up). No P0/P1. The tool correctly covers all 109 existing `/v1/manager` routes, the CI integration is a faithful mirror of the existing `personal-msgsendreq-lint` job, the allowlist is accurate and minimal, and the test suite covers the key patterns including the function-local scoping regression. The six P2 notes are recommended improvements for Part 2, not landing blockers.

yujiawei (changes requested):
# Code Review — PR #409 (octo-server)

**Verdict: CHANGES_REQUESTED** (one P1 blocker on an idiomatic registration path; see below). Reviewed at head `ee7e24b7`, merge-base `fb4763d0`, against `main`. Diff is 5 files / +884, exactly as described.

## What's good

- The guard does what it claims for the **current** surface: `go run ./tools/lint-manager-authz ./modules` is green and counts all **109** `/v1/manager` routes; `go test ./tools/lint-manager-authz/...` passes (9 cases). I re-ran both locally.
- The allowlist is justified by the code. With an empty allowlist exactly 5 routes flag — `POST /v1/manager/login` (public, pre-auth) and the 4 `/v1/manager/secrets` CRUD ops. I verified all 4 secret handlers scope every operation by `owner = c.GetLoginUID()` (`modules/usersecret/api.go` `create`/`list`/`update`/`delete`), so allowlisting them is correct, not a hidden IDOR.
- Function-local group resolution is the right call and the regression test (`TestFunctionLocalGroupScoping`) locks in the real false-positive trap (reused `auth` var across `Route()` methods).
- Honest scoping: this is a thin, reversible net, no authz refactor. Good.

## P1 (blocking) — root-level route registrations are silently invisible to the guard

`tools/lint-manager-authz/main.go:333-383` (`managerRoutes`) only inspects functions that contain a local `.Group(...)` assignment and only matches verb calls whose receiver resolves to that group variable. A route registered **directly on the root router** is never seen:

```go
r.GET("/v1/manager/export_all_users", m.dump)   // no role check -> guard stays GREEN
```

This is not an exotic shape — it is idiomatic in this repo. Root-level `r.GET/POST/Any("/v1/...")` is used in `modules/webhook/api.go`, `modules/usersecret/api.go:113`, `modules/app_bot/app_bot.go:139`, `modules/bot_api/bot_api.go:204`, `modules/base/app/api.go`, `modules/botfather/api.go`. In the pinned octo-lib (`f62e626a34db`) `WKHttp` exposes `GET`/`POST`/`Any` at the root, so an admin route added this way compiles, serves, and passes CI with no role check.

Why this blocks: the PR also adds to `CLAUDE.md` that the guard "fails the build on **any** `/v1/manager` route whose handler lacks a check." That promise is not kept for root-level registration. A future engineer trusts the doc, registers an admin route on `r` directly (or via `r.Any`), and ships a privilege-escalation hole **behind a green security gate** — worse than no guard, because the green check manufactures false confidence. For a security control, a documented guarantee that the code does not honor on an idiomatic path is a correctness defect, not a nit.

Cheap fix: also walk top-level `r.VERB("/v1/manager...", handler)` calls (string-literal path starting with `/v1/manager`), not just group-relative ones. `Any` is already in `httpVerbs`, so include it in the root scan too.

## P1 / P2 — two more real false negatives (fix or document as known gaps)

- **[P1] Group passed to a helper** (`main.go:341-343`). `localGroupVars` is computed per-function, so `auth := r.Group("/v1/manager", …); m.register(auth)` where `register` does `auth.GET(...)` registers routes the guard never inspects (helper has no local `Group`). No manager route uses this today, but it is a normal Go pattern and nothing prevents it. At minimum the walk should follow a manager group passed as an argument, or fail-closed when a `/v1/manager` group escapes into a call.

- **[P2] Presence-not-enforcement** (`main.go:438-478`). A handler passes as soon as the *identifier* `CheckLoginRole`/`CheckLoginRoleIsSuperAdmin` appears anywhere in its AST. An ignored return value or a late check passes:
  ```go
  c.CheckLoginRoleIsSuperAdmin()   // err discarded
  doAdminWork()                    // runs regardless -> guard GREEN
  ```
  All current handlers act on the error (verified), so this is advisory, but it is the difference between "a check is present" and "a check gates execution." Worth a comment in the tool doc noting the guard verifies presence, not control flow.

## Refuted finding (so it isn't chased)

A cross-tool pass flagged `v1 := r.Group("/v1"); mgr := v1.Group("/manager")` composition as a dropped path. **Not reachable in this framework**: octo-lib `RouterGroup` exposes only `GET/POST/PUT/DELETE` and has no `Group` method, so subgroups cannot be nested off a group — every group is created directly from `WKHttp.Group`. The guard's parent-composition logic and `TestManagerSubgroupInScope` exercise a shape the real router type cannot produce. Harmless, but it is dead coverage, not protection.

## Coverage notes (manual verification beyond the diff)

- Confirmed all 14 manager-route files use the single in-body `x := r.Group("/v1/manager…")` pattern the guard handles — that is *why* the current 109-route inventory is fully covered. The guard's correctness today rests entirely on that convention holding.
- `httpVerbs` lists `PATCH/HEAD/OPTIONS/Handle` that the pinned router type does not expose. Harmless, but the verb set is not aligned to the actual framework surface.
- Not re-verified here (flagging for human attention on a security PR): whether `CheckLoginRole`/`CheckLoginRoleIsSuperAdmin` in octo-lib are themselves correct and revocation-aware — the guard trusts these primitives by name only.

## Path to approval

The blocker is small relative to the guard's value. Either: (a) extend the AST walk to root-level `r.VERB("/v1/manager…")` registrations (closes P1 #1), or (b) if you want to keep scope truly minimal, soften the `CLAUDE.md` / tool-doc wording so it states the guard's *actual* coverage and lists the unsupported patterns (root-level, group-passed-to-helper, ignored-error) as known gaps with a tracking issue. (a) is preferred — it's the idiomatic gap. Fixing the helper-passing case (P1 #2) or fail-closing on it would make the net trustworthy enough to rely on.

yujiawei:
Two small additions to the review above, both in the same "the walk should match what the framework actually allows" spirit as the blocker:

- **`stringLit` should use `strconv.Unquote`** (`main.go:368-375`). `strings.Trim(lit.Value, "`\"")` strips a cutset, not Go quoting. Plain ASCII paths are fine today, so this is a correctness nit, not a live gap — but for a guard whose whole job is exact path matching, `strconv.Unquote` is the right primitive.

- **Add `Match` to `httpVerbs`** (`main.go:79-82`) if you keep the verb-allowlist approach. The router can register a route via `Match([]string{...}, path, h)`; it's unused today, but it's the same shape of future blind spot as the root-level/helper cases — a real registration verb the walk doesn't recognize.

Neither changes the verdict on its own; folding them in while you address the root-level registration gap keeps the guard's coverage aligned with the actual router surface.

lml2468:
# QA Review — FAIL

**Verdict**: `FAIL`
**AC source**: linked issue #366
**CI status**: failing (`CI / Test` failed; `Manager Authz Lint` passed)
**Test diff**: +1 test file, ~254 lines

## 一句话结论

The PR adds a useful guard and targeted unit tests, but QA cannot pass it while required CI is red and while the guard/test matrix does not cover an idiomatic `/v1/manager` registration path promised by the acceptance criteria.

## AC 覆盖矩阵

| AC # | 描述 | 代码路径 | 测试 | 断言强度 | 状态 |
|---|---|---|---|---|---|
| AC1 | Part 1 CI guard exists; new ungated `/v1/manager` route fails CI; intentional exceptions are allowlisted with reasons | `.github/workflows/ci.yml:309`, `tools/lint-manager-authz/main.go:314`, `tools/lint-manager-authz/allowlist.txt` | `TestFlagsUncheckedManagerRoute`, `TestAllowlistSuppresses`, plus role-check/group-scoping cases | Medium | **PARTIAL** |
| AC2 | Part 2 design proposal for central authz layer can be separate PR/RFC once Part 1 lands | Not in this diff; PR explicitly scopes to Part 1 but still uses `Closes #366` | none | n/a | **MISSING / SCOPE MISMATCH** |

## 边界 / 边缘用例

| 路径 | 边界类别 | 是否测试 |
|---|---|---|
| Direct root route registration, e.g. `r.GET("/v1/manager/...", handler)` | Registration style not based on local `.Group("/v1/manager")` variable | **No** |
| Manager group passed into a helper registrar | Group escapes the function-local `localGroupVars` map | **No** |
| Role-check call present but not enforced before work | Presence vs control-flow enforcement | **No** |

## 回归风险

- **触达模块**: CI workflow + new security-sensitive source guard for admin route authorization
- **近 30 天变更频率**: not locally available; workspace repo checkout is not configured for this tracker
- **关联 incident**: issue #366 follows prior authz audit/fix context
- **公共契约改动**: CI contract changes; new required lint job candidate
- **风险评级**: High

## 可观测性

- [x] Guard failure message includes route, handler, and remediation text
- [x] CI job is explicit (`Manager Authz Lint`) and separately visible
- [ ] No runtime metrics/logging needed; this is a build-time guard
- [ ] No alerting change included; acceptable for a CI guard, but required-check configuration remains a human follow-up

## 测试执行

- **存在性**: 有 — `tools/lint-manager-authz/main_test.go`
- **CI**: 红 — `CI / Test` failed on `TestQueryBotUIDsOwnedByUIDs_BasicMatch` with `panic: Error 1040: Too many connections`; recent `main` CI runs are green, so I cannot treat this as proven baseline/flaky
- **Guard CI**: `Manager Authz Lint` passed
- **覆盖率变化**: n/a
- **Flaky 测试**: none proven from available evidence

## Blockers

1. Get required CI green, or provide clear baseline/flaky evidence for the `CI / Test` failure. QA policy is fail-closed when test CI is red.
2. Add implementation + regression coverage for direct root `/v1/manager` route registrations. `managerRoutes()` currently skips functions with no local manager group (`tools/lint-manager-authz/main.go:341-343`) and only counts verb calls whose receiver is in that group map (`main.go:355-361`), so `r.GET("/v1/manager/leak", handler)` can stay invisible despite AC1 saying every mounted manager handler is guarded.
3. Resolve the #366 scope mismatch: either change `Closes #366` to `Refs #366` for this Part 1-only PR, or include/link the Part 2 central-authz design deliverable required by the issue AC.

<sub>Labels: `review:done:qa:fail` · 由 qa-engineer 智能体生成 · 不代表合并意见，合并是人类专属操作。</sub>

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD ee7e24b7

## Reviewer Summary
|---|---|---|
| qa-engineer | `pass-with-risk` | All Part-1 acceptance criteria verified; one red CI job is an unrelated MySQL flake; guard detection surface is narrow. |
| security-engineer | `cleared-with-risk` | Adds no runtime/dependency/credential surface — a fail-closed defensive control; allowlist independently verified to bless only genuinely non-admin routes. |
| code-reviewer | `needs-discussion` | Clean, well-tested guard; detection blind spots (empirically reproduced) make it narrower than its stated "mechanical guarantee". |

Three-way convergence: the change is good and mergeable, but the guard's guarantee has an empirically-confirmed boundary that should be closed or documented, and the gate is not yet merge-enforcing.

## Critical Findings (P0-P1)
无 Critical / High priority findings. No blocker exists: 0 Must-Fix, no security-exploitable issue, 0 live instances of any detection gap, and the only red code-test job is not attributable to this PR (see Questions).

## Should-Fix (P2)
- **[P2-1] `tools/lint-manager-authz/main.go:333-383` (`managerRoutes`) & `:244-308` (`localGroupVars`) — detection-surface blind spots weaken the "mechanical guarantee".** The guard only inspects routes registered on a **function-local group variable bound to a string-literal `.Group("/v1/manager"…)`**. Two forms silently escape: (a) an admin route registered directly on the root router with an inline path (`r.GET("/v1/manager/x", h)`); (b) a group whose base path is a non-literal (const/var → `stringLit` returns "" → group dropped). **Triple-reviewer consensus** (qa/S-1 + security/P-1 + code/S-1), and **empirically reproduced by code-reviewer**: an ungated `r.GET("/v1/manager/backdoor", h)` yields `routes inspected=0, violations=[]`. **0 live instances in `./modules` today**, so this is a control-completeness gap, not a live hole. *Fix*: also scan verb-call args for an inline literal path matching the manager prefix, and/or document the detection boundary in the package doc; add a fixture per form.
- **[P2-2] `.github/workflows/ci.yml` (`manager-authz-lint`) — gate is advisory, not merge-blocking.** The job is `if: needs.changes.outputs.code == 'true'` gated and not yet in branch-protection required checks (PR checklist item unchecked). Until it is required, a red guard does not block merge, and a change the `changes` filter does not classify as `code` skips the guard. Flagged by security/P-3 + qa/Q-1 + code/Q-1. *Fix*: add `Manager Authz Lint` to required status checks, and check that the `changes` path-filter treats Go route files under the modules tree as code.

## Suggestions (P3)
- **[P3-1] `main.go:79-82,363-367` — `.Handle` verb/path arg-order mislabel.** `.Handle("METHOD","/path",h)` reports the violation path as `/v1/manager/METHOD` and produces a nonsensical allowlist signature (reproduced by code-reviewer). Not a coverage bypass — the handler (last arg) is still evaluated. `.Handle(` used 0× in `./modules`. Consensus qa/N-1 + security/P-2 + code/N-1. *Fix*: special-case `Handle` to take the path from `Args[1]`.
- **[P3-2] `main.go:6-16` — package-doc framing overpromises.** The "mechanical guarantee … CI failure instead of a production incident" wording is stronger than the detection surface delivers (see P2-1). *Fix*: fold a one-line "known boundary" note into the doc as part of the P2-1 fix. (code/N-2.)

## Questions for Author
- **[Q-1]** The `Test` CI job is red with `Error 1040: Too many connections` in `modules/group` — a module this PR does not touch (PR changes only `tools/`, `ci.yml`, one `CLAUDE.md` line). QA assessed this as a MySQL connection-pool infra flake, not attributable to the change. Please re-run / confirm it is the known flake before merge. (qa §4.)
- **[Q-2]** Is making `Manager Authz Lint` a required status check tracked as follow-up (see P2-2)? That is what actually enforces the gate at merge time.

## Verdict Rationale
**Formal Verdict = `NEEDS_DISCUSSION`.** No P0/P1 blocker exists: 0 Must-Fix, qa=`pass-with-risk`, security=`cleared-with-risk`, and the change adds no runtime, dependency, or credential surface — it is a fail-closed defensive control that can only reduce admin-surface privilege-escalation risk. That rules out `REQUEST_CHANGES`. It is not `APPROVE` because two P2 should-fix items remain and there are open material questions: the guard's guarantee has an empirically-reproduced (though 0-live-instance) detection boundary (P2-1, triple consensus) that warrants a close-or-document decision, the gate is not yet merge-enforcing (P2-2), and a red CI job needs human confirmation as a flake (Q-1). There were **no cross-reviewer disagreements** — all three converged (qa/security cleared-with-risk, code needs-discussion for the same boundary), and the security-critical allowlist was independently verified by both qa and security to bless only genuinely non-admin routes (public login + owner-scoped `/v1/manager/secrets`), which code-reviewer corroborated via an empty-allowlist run flagging exactly those 5 routes.

## Round Context

---

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD ee7e24b7

**Formal Verdict**: `APPROVE`



## Reviewer Summary

|---|---|---|
| qa-engineer | pass-with-risk | Guard behavior fully verified (9 tests + reproduced green at 109 routes / 5 allowlist); only red signal is an unrelated MySQL "too many connections" infra flake in `modules/group`, a module this additive PR does not touch. |
| security-engineer | cleared-with-risk | Purely additive build-time CI net, no runtime attack surface; *improves* admin-authz posture. Allowlist re-verified — all 5 entries are genuinely public/user-scoped, none are hidden admin routes. |
| code-reviewer | approved-with-nits | Clean fail-closed AST guard mirroring the existing `lint-personal-msgsendreq` convention; 0 Must Fix, 0 Should Fix, 4 Nits, all latent/cosmetic. |

## Critical Findings (P0-P1)

## Should-Fix (P2)

无 — no reviewer raised a P2 / Should-Fix item.

## Suggestions (P3)

- [P3-1] `tools/lint-manager-authz/main.go:269-271` (doc at `:18-23`) — Group base path must be a string literal (`stringLit` returns `""` otherwise), so a dynamically-constructed/concatenated `/v1/manager` group path silently escapes the guard — a **false-negative** direction. No current route is affected (all 109 use literals, verified). Fix: add one line to the package doc naming the literal-only limitation so the guard isn't mistaken for exhaustive authz enforcement (real fix is the deferred Part-2 central middleware). *Convergent — flagged by qa/Risk-2 + security/Risk-1 + code/Nit-2.*
- [P3-2] `tools/lint-manager-authz/main.go:79-82, 366` — `httpVerbs` includes `"Handle"`, but gin's `Handle(method, path, handlers...)` puts the method at `Args[0]` and path at `Args[1]`; the code reads the path from `Args[0]`, so a route registered via `.Handle("GET", "/x", h)` would report a garbled sig (`Handle /v1/manager/GET`) and resist clean allowlisting. Pass/fail decision is unaffected (handler is always `Args[len-1]`, fail-closed holds) and no manager group uses `.Handle(` today. Fix: drop `Handle` from the verb set or special-case its arg layout. *Flagged by code/Nit-1.*
- [P3-3] `tools/lint-manager-authz/main.go:441` — `bodyHasRoleCheck` uses `ast.Inspect` over the whole body, descending into nested func literals, so a role check buried in an unrelated inner closure would count the handler as covered — a theoretical false-negative. Pathological, not observed. Fix: add a comment documenting the intended "any check anywhere in body counts" semantics. *Flagged by code/Nit-3.*
- [P3-4] `tools/lint-manager-authz/main.go:66-69` — `roleCheckNames` is a hand-maintained string set that must track octo-lib's `wkhttp.Context`; an upstream rename degrades safely (routes get flagged = noisy false positives, not silent bypass). Direction acceptable; noted so the coupling is on record for the Part-2 work. *Flagged by code/Nit-4.*
- [P3-5] Resolution is bounded to same-package with `maxHelperDepth=2`; cross-package handlers or deeper delegation are treated as *unchecked* → potential false positive (CI noise). This is the safe fail-closed direction with the allowlist as escape valve — not a security gap. *Flagged by qa/Risk-3 + security/Risk-2.*
- [P3-6] `.github/workflows/ci.yml` — the new `manager-authz-lint` job is gated on `needs.changes.outputs.code == 'true'`, mirroring the existing `personal-msgsendreq-lint` convention. Low risk; flag if the path-filter classification of Go changes is later altered. *Flagged by qa/Risk-4 + security/Risk-3.*

## Questions for Author

无 — no reviewer raised a material blocking question.

## Verdict Rationale

Applying the deterministic first-match rule table: no P0, no P1, no P2 (0 Must-Fix and 0 Should-Fix across all three reviewers), no material questions — only P3 suggestions remain → **APPROVE**. The three reviewers fully converged: this is a thin, purely additive, build-time CI safety net (an AST guard failing the build when a `/v1/manager` handler lacks a role check) with **zero runtime/production code** and no import path into `modules/**`, so the blast radius is CI-only and the worst case is a false-positive caught by the allowlist escape valve. All four Code nits are latent/cosmetic, and the single most-cited item (P3-1, string-literal group base) is a documented coverage boundary — not a defect — that all three roles agree is acceptable for a Part-1 guard whose complement is the deferred central authz middleware.

One non-code caveat for the merger, not a finding: the aggregate `CI / Test` job is currently **red** due to a `MySQL Error 1040: Too many connections` infra flake in `modules/group` — a module this PR does not touch (the PR's own `Manager Authz Lint` job is green). If `Test` is a required status check, re-run CI to green before merge. This is an environment condition, not attributable to the PR, and does not lower the verdict.

No reviewer disagreements to reconcile — all three lower-severity verdicts (pass-with-risk / cleared-with-risk / approved-with-nits) point the same direction.

## Round Context

---

lml2468 (commented):
APPROVE — Round 1 (HEAD ee7e24b7)
Purely additive build-time CI guard (AST role-check on /v1/manager routes); all three reviewers converged — 0 Must-Fix, 0 Should-Fix, only P3 nits/risks.
Full findings: aggregate comment on this PR — https://github.com/Mininglamp-OSS/octo-server/pull/409#issuecomment-4864131491

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD ee7e24b7

**Formal Verdict**: `COMMENT`



## Lens Summary

| Lens | Verdict | Headline |
|---|---|---|
| code | `approved-with-nits` | Sound fail-closed CI guard; one probe-confirmed delegation false-negative + latent Handle/Any bug. |
| security | `cleared-with-risk` | No cred leak / no exploitable bypass; guard catches missing checks but not ignored-result checks. |
| qa | `pass-with-risk` | Guard proven end-to-end (109 routes; inject→exit 1); Test-job RED is an unrelated MySQL flake. |

## Critical Findings (P0-P1)

无 Critical / High priority findings — all three lenses cleared the change with no blocking issue. The guard is a genuine, fail-closed security improvement and is proven green on all 109 live `/v1/manager` routes.

## Should-Fix (P2)

- `[P2-1]` `tools/lint-manager-authz/main.go:507-509` — `namedFuncHasRoleCheck` falls back to `cands[0]` when no candidate's receiver type matches, which **fails open**: a delegated handler whose real (same-receiver) helper has no check can be credited to an unrelated same-named method on another receiver that *does* check — marking a genuinely-unchecked route as covered. Code lens confirmed this with a probe fixture (reported 0 violations for a route with no real check). Reach in the current codebase is zero (no such name collision in the 109 routes), but the guard's value is catching *future* routes, where method names (`create`/`update`/`save`/`delete`) collide across receiver types. — flagged by code/C-1 (+sec/S-3) — Fix: for a method-value handler (`recv.Method`) with no receiver-type match, treat as unresolved → not covered (return false) instead of `cands[0]`; keep `cands[0]` only for the plain-func / single-candidate case, and add a regression test. Recommended before adding "Manager Authz Lint" to *required* status checks.
- `[P2-2]` `tools/lint-manager-authz/main.go:519-543` (`bodyHasRoleCheck`) — the guard counts a role check as present if the call appears *anywhere* in the body, regardless of whether its returned error is acted on. `CheckLoginRole` / `CheckLoginRoleIsSuperAdmin` are pure error-returners (they don't abort), so a handler that calls `c.CheckLoginRole()` and drops the error, or holds the check in a statically-dead branch, passes the guard while enforcing nothing. Confirmed empirically by two lenses. **Residual, not active** — all 109 current handlers use `if err := ...; err != nil { return }` (grep-clean). — flagged by security/S-1 + qa/B-1 (consensus, +code/Q-1) — Fix: document in the package doc that presence ≠ enforcement (verifying error flow needs data-flow analysis, out of scope for this syntactic Part-1 guard); the centralized authz-middleware workstream (#366 main) is the real control. Optional test asserting the limitation.

## Suggestions (P3)

- `[P3-1]` `main.go:159-163, 447-448` — `httpVerbs` includes `Any`/`Handle`, but the path is read from `Args[0]`; for gin-style `.Handle(method, path, ...h)` the method sits in `Args[0]`, so the path/sig would be mis-derived. **Verified unused** — no `.Handle(`/`.Any(` route registration exists on any manager group (octo-lib `RouterGroup` exposes neither), so it is inert today. — flagged by code/N-1 + sec/S-3 + qa/Q-1 — Fix: drop `Handle`/`Any` from `httpVerbs` with a comment, or special-case their arg offsets if ever introduced.
- `[P3-2]` `allowlist.txt:12-15` — allowlisting the 4 `/v1/manager/secrets` CRUD routes removes **all** guard coverage there; the "user-scoped, owner==login-UID" justification is defensible and documented but unverifiable by the guard, so a future edit dropping the in-handler owner check would go uncaught. — flagged by sec/S-2 — Fix (follow-up): a companion guard asserting `GetLoginUID`/owner-scoping presence on user-scoped allowlisted routes.
- `[P3-3]` `main.go:139` — coverage is defined solely by the `/v1/manager` prefix; an admin surface mounted under any other prefix is out of scope by construction (verified none exists today: `/v1/obo`, `/v1/file`, `/v1/integrations` are non-admin). — flagged by sec/S-4 — Fix: document "admin surface == `/v1/manager` prefix" as an invariant the guard depends on.
- `[P3-4]` `main.go:611` — `strings.Trim(lit.Value, "`\"")` is a cutset trim; correct for all real path literals, `strconv.Unquote` would be exact. Harmless. — flagged by code/N-2.
- `[P3-5]` `main_test.go` — no unit test for the fail-closed "handler defined outside package → treated as unchecked" branch; behavior asserted by code inspection only. — flagged by qa/B-2 — Fix: add a cross-package-handler fixture.

## Questions for Author

- `[Q-1]` The `Test` CI job is RED on this HEAD. QA root-caused it to an unrelated `panic: Error 1040: Too many connections` (MySQL pool exhaustion) in `modules/group/TestQueryBotUIDsOwnedByUIDs_BasicMatch` — a module this PR does not touch, and the guard's own tests pass in the separate green `Manager Authz Lint` job. Please confirm this is the known infra flake and consider a re-run so the full suite is green before merge. (from qa)
- `[Q-2]` Should `Handle`/`Any` be removed from `httpVerbs` now (dead + a latent mis-parse trap), or intentionally kept as forward-proofing with corrected arg handling? (from code/qa/security)

## Verdict Rationale

`COMMENT`: no lens is `blocked`/`fail`/`changes-requested`; there are 0 P0/P1 findings, exactly 2 P2s, and material questions — which maps to `COMMENT` per the deterministic verdict rule. The two P2s (C-1 `cands[0]` fail-open, and the presence-≠-enforcement gap) are both fail-*open* directions in a security guard and are the one class of defect such a guard must minimize; however, both have **zero current reach** (confirmed by executing the guard against the real repo: 109 routes green, and no triggering pattern present), so they are hardening follow-ups rather than shipped defects. Cross-lens consensus is strong: all three lenses independently flagged the presence-≠-enforcement gap at ≥0.8 confidence, and code lens empirically confirmed the `cands[0]` false-negative with a probe. The one CI red (`Test`) was investigated to an unrelated MySQL-connection flake, so it does not force `REQUEST_CHANGES` — but the full suite is not green, hence it is raised as Q-1 for the author/human owner. Recommend fixing C-1 before this guard is promoted to a *required* status check.

## Round Context

ROUND=1: First-round review — no prior round context.

|---|---|---|
| `tools/lint-manager-authz/main.go` | ✓ | P2-1, P2-2, P3-1, P3-3, P3-4, Q-2 |
| `tools/lint-manager-authz/main_test.go` | ✓ | P3-5 |
| `tools/lint-manager-authz/allowlist.txt` | ✓ (verified exact-match to flagged set) | P3-2 |
| `.github/workflows/ci.yml` | ✓ (changes path-filter inspected — no trigger gap) | (none) |
| `CLAUDE.md` | ✓ | (none) |

## Notes

- Confidence filter aggregate: code dropped 1 / downgraded 1; security dropped 0 / downgraded 0; qa dropped 1 / downgraded 1.
- Cross-lens consensus boost applied to the presence-≠-enforcement finding (flagged by 3/3 lenses).
- Empirical verification: all three lenses executed `go test ./tools/lint-manager-authz/...` (9/9 pass) and `go run ./tools/lint-manager-authz ./modules` ("OK: all 109 routes… 5 allowlist entr(ies)") against the repo at ee7e24b7; QA additionally proved the end-to-end CI behavior (injected an unchecked `/v1/manager` route → exit 1; reverted → exit 0) and confirmed the empty-allowlist run flags exactly the 5 allowlisted routes (no stale/over-broad entries).
- No security `blocked` finding → no embargo comment for this round.

---

lml2468 (commented):
`COMMENT` — Round 1

CI source-guard for `/v1/manager` authz: net-positive, fail-closed, proven green on all 109 live routes and proven to fail on an injected unchecked route. No blocking issues. Two fail-open hardening follow-ups worth addressing before promoting this to a *required* check: (1) `namedFuncHasRoleCheck` `cands[0]` fallback can credit a wrong same-named method (probe-confirmed false-negative); (2) the guard counts a role check's presence, not enforcement (ignored-error/dead-branch passes). Heads-up: the `Test` CI job is red on an unrelated MySQL "too many connections" flake in `modules/group` (untouched here) — worth a re-run for a green suite before merge.

See the aggregate comment on this PR (and the multica tracker) for full findings.

PR #414, opened 2026-06-18
feat(bot_api): /v1/bot streaming + per-UID rate limiting (OCT-31, OCT-42, OCT-41)

lml2468:
Integration branch for the **OCT-31** epic (public `/v1/bot` streaming). Carries the OCT-42 security fix and the OCT-41 stream-ownership binding as children that fold into this PR. Opened as **draft**: OCT-31 is blocked by OCT-37 (in_review) and final merge clears the 3-review / `enforce_admins` gate.

## Commits
- `9162560` feat(bot_api): add `/v1/bot/stream/{start,end}` for incremental channel streaming (OCT-31)
- `8699351` feat(bot_api): per-UID rate limit the `/v1/bot` send surface (OCT-42)
- `7423c5d` fix(bot_api): bind stream_no↔bot ownership on `/v1/bot/stream/end` (OCT-41)
- `70e7606` test(bot_api): cover App Bot (`app_`) per-UID rate limiting (OCT-42)

## OCT-42 — per-UID rate limiting (up for Security review)
**Gap:** `/v1/bot` and `/v1/botfile` mounted only `authBot()` — no per-bot (UID) cap, only the global per-IP DDoS floor. `stream/start` (a live WuKongIM stream + channel bubble) was the sharpest amplification primitive on this surface.

**Fix:**
- `authBot()` mirrors the resolved `robotID`/UID onto the generic `"uid"` context key in one place (`setBotActorUID`) at **all three** principal-set sites — User Bot (`auth.go` userbot), App Bot registry, App Bot DB-fallback — so `SharedUIDRateLimiter` reads `uid` instead of silently failing open.
- Mounted `SharedUIDRateLimiter` after `authBot()` on both `/v1/bot` and `/v1/botfile` (shared bucket `ratelimit:uid:{robotID}`, default 2 rps / burst 60).

**Tests** (`modules/bot_api/ratelimit_test.go`):
- `TestBotAPI_PerUIDRateLimit_TripsBucket` — User Bot (`bf_`) trips the bucket → 429 + `X-RateLimit-Scope: uid` + `Retry-After`.
- `TestBotAPI_PerUIDRateLimit_AppBot_TripsBucket` — App Bot (`app_`) via registry path (always) and DB-fallback path (runs in CI where the `app_bot` table is present; skips on partial local schemas). Proves `app_` tokens are per-UID capped too — closing the silent-fail-open trap for App Bots.

**Item #2 (dedicated `stream/start` cap):** deferred → tracked in OCT-50 (octo-lib hardcodes the `ratelimit:uid:` bucket prefix, so a second middleware shares the same bucket; a true concurrent-open ceiling needs the OCT-41 `bot:stream:owner:{stream_no}` foundation).

## Verification
- `go test ./modules/bot_api/` → PASS (full package)
- `go vet` + `go build ./modules/bot_api/` clean

Refs: OCT-31, OCT-42, OCT-41, OCT-38 (source review), OCT-50 (follow-up).

### DIFF
--- a/modules/bot_api/api_i18n_test.go
+++ b/modules/bot_api/api_i18n_test.go
@@ -37,6 +37,7 @@ func TestBotAPINoLegacyResponseError(t *testing.T) {
 		"auth.go", "register.go", "commands.go", "events.go", "mention_pref.go",
 		"sync.go", "typing.go", "send.go", "threads.go", "file.go", "groups.go",
 		"voice_adapter.go", "obo_api.go", "resolve_targets.go", "incoming_webhook.go",
+		"stream.go",
 	}
 	banned := []string{
 		".ResponseError(",
--- a/modules/bot_api/auth.go
+++ b/modules/bot_api/auth.go
@@ -42,6 +42,18 @@ func (ba *BotAPI) authBot() wkhttp.HandlerFunc {
 	}
 }
 
+// setBotActorUID records the resolved bot identity on both the bot-specific
+// CtxKeyRobotID and the generic "uid" context key. The "uid" mirror is what
+// SharedUIDRateLimiter (and any other AuthMiddleware-shaped consumer) reads —
+// setting it at every authBot resolution point means every authBot-mounted
+// group is per-UID rate limited without each call site remembering to insert
+// botActorUID(). The limiter fails open silently when "uid" is unset (see
+// UIDRateLimitMiddleware), so this must run wherever auth succeeds. (OCT-42)
+func setBotActorUID(c *wkhttp.Context, uid string) {
+	c.Set(CtxKeyRobotID, uid)
+	c.Set("uid", uid)
+}
+
 // authUserBot authenticates a User Bot via robot table lookup.
 func (ba *BotAPI) authUserBot(c *wkhttp.Context, token string) {
 	robot, err := ba.db.queryRobotByBotToken(token)
@@ -55,7 +67,7 @@ func (ba *BotAPI) authUserBot(c *wkhttp.Context, token string) {
 		return
 	}
 
-	c.Set(CtxKeyRobotID, robot.RobotID)
+	setBotActorUID(c, robot.RobotID)
 	c.Set(CtxKeyBotKind, BotKindUser)
 	c.Set(CtxKeyRobot, robot)
 	c.Next()
@@ -67,7 +79,7 @@ func (ba *BotAPI) authUserBot(c *wkhttp.Context, token string) {
 func (ba *BotAPI) authAppBot(c *wkhttp.Context, token string) {
 	// Try in-memory Registry first (O(1))
 	if spec := ba.lookupAppBotRegistry(token); spec != nil {
-		c.Set(CtxKeyRobotID, spec.UID)
+		setBotActorUID(c, spec.UID)
 		c.Set(CtxKeyBotKind, BotKindApp)
 		c.Set(CtxKeyAppBotScope, spec.Scope)
 		if spec.Scope == "space" {
@@ -95,7 +107,7 @@ func (ba *BotAPI) authAppBot(c *wkhttp.Context, token string) {
 		return
 	}
 
-	c.Set(CtxKeyRobotID, appBot.UID)
+	setBotActorUID(c, appBot.UID)
 	c.Set(CtxKeyBotKind, BotKindApp)
 	c.Set(CtxKeyAppBotScope, appBot.Scope)
 	if appBot.Scope == "space" {
--- a/modules/bot_api/bot_api.go
+++ b/modules/bot_api/bot_api.go
@@ -66,6 +66,12 @@ type BotAPI struct {
 	// WuKongIM. nil in production → the real ba.ctx path runs.
 	streamStartOverride func(config.MessageStreamStartReq) (string, error)
 	streamEndOverride   func(config.MessageStreamEndReq) error
+	// streamOwnerStoreOverride lets unit tests inject an in-memory
+	// stream_no→robotID ownership map so the OCT-41 ownership check in
+	// streamEnd (and the binding write in streamStart) can be table-tested
+	// without a live Redis. nil in production → the Redis-backed
+	// redisStreamOwnerStore runs. See stream_owner.go.
+	streamOwnerStoreOverride streamOwnerStore
 	// oboStoreOverride lets unit tests inject an in-memory oboStore so
 	// checkOBO / REST handlers / fan-out can run without standing up MySQL.
 	// nil in production; the real path uses ba.db (which satisfies oboStore).
--- a/modules/bot_api/bot_api.go
+++ b/modules/bot_api/bot_api.go
@@ -15,6 +15,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/modules/thread"
 	"github.com/Mininglamp-OSS/octo-server/modules/user"
 	"github.com/Mininglamp-OSS/octo-server/modules/voice_adapter"
+	appwkhttp "github.com/Mininglamp-OSS/octo-server/pkg/wkhttp"
 )
 
 const (
@@ -29,13 +30,13 @@ const (
 // BotAPI is the public Bot API gateway module.
 // It handles all bot-facing endpoints (/v1/bot/*) with unified auth.
 type BotAPI struct {
-	ctx                   *config.Context
-	db                    *botAPIDB
-	userService           user.IService
-	fileService           file.IService
-	groupService          group.IService
-	userDB                *user.DB
-	threadService         thread.IService
+	ctx           *config.Context
+	db            *botAPIDB
+	userService   user.IService
+	fileService   file.IService
+	groupService  group.IService
+	userDB        *user.DB
+	threadService thread.IService
 	// robotService gives the OBO fan-out path a way to enqueue synthetic
 	// events directly into a grantee bot's /v1/bot/events queue. The
 	// webhook layer drops NoPersist=1 messages before NotifyMessagesListeners
@@ -232,8 +233,12 @@ func (ba *BotAPI) Route(r *wkhttp.WKHttp) {
 	// register endpoint (token needed but not via authBot group — handled inline)
 	r.POST("/v1/bot/register", ba.register)
 
-	// Bot API endpoints (unified auth middleware)
-	botAPI := r.Group("/v1/bot", ba.authBot())
+	// Bot API endpoints (unified auth middleware).
+	// SharedUIDRateLimiter is mounted AFTER authBot() so it reads the "uid"
+	// authBot sets (= robotID): every authenticated bot send/stream/manage call
+	// is capped per-bot, not just by the global per-IP DDoS floor. This closes
+	// the stream-amplification gap on stream/start (OCT-42).
+	botAPI := r.Group("/v1/bot", ba.authBot(), appwkhttp.SharedUIDRateLimiter(r, ba.ctx))
 	{
 		botAPI.POST("/sendMessage", ba.sendMessage)
 		botAPI.POST("/typing", ba.typing)
@@ -295,8 +300,9 @@ func (ba *BotAPI) Route(r *wkhttp.WKHttp) {
 		botAPI.GET("/obo-grant", ba.oboBotGetGrant)
 	}
 
-	// Bot File API (separate group for wildcard conflict avoidance)
-	botFileAPI := r.Group("/v1/botfile", ba.authBot())
+	// Bot File API (separate group for wildcard conflict avoidance).
+	// Same per-UID cap as the main bot group (OCT-42).
+	botFileAPI := r.Group("/v1/botfile", ba.authBot(), appwkhttp.SharedUIDRateLimiter(r, ba.ctx))
 	{
 		botFileAPI.GET("/*path", ba.botProxyFile)
 		botFileAPI.POST("/upload", ba.botUploadFile)
--- a/modules/bot_api/bot_api.go
+++ b/modules/bot_api/bot_api.go
@@ -58,6 +58,13 @@ type BotAPI struct {
 	// final MsgSendReq (including server-authoritative payload.space_id).
 	// nil in production; the real path goes through ba.ctx.SendMessageWithResult.
 	dispatchOverride func(*config.MsgSendReq) (*config.MsgSendResp, error)
+	// streamStartOverride / streamEndOverride let unit tests intercept the
+	// WuKongIM stream RPCs (IMStreamStart / IMStreamEnd) so the /v1/bot/stream/*
+	// handlers can be table-tested — including the security-critical assertion
+	// that FromUID is forced to the authenticated robotID — without a live
+	// WuKongIM. nil in production → the real ba.ctx path runs.
+	streamStartOverride func(config.MessageStreamStartReq) (string, error)
+	streamEndOverride   func(config.MessageStreamEndReq) error
 	// oboStoreOverride lets unit tests inject an in-memory oboStore so
 	// checkOBO / REST handlers / fan-out can run without standing up MySQL.
 	// nil in production; the real path uses ba.db (which satisfies oboStore).
@@ -140,6 +147,28 @@ func (ba *BotAPI) dispatchMsgSendReq(req *config.MsgSendReq) (*config.MsgSendRes
 	return ba.ctx.SendMessageWithResult(req)
 }
 
+// dispatchStreamStart opens a WuKongIM message stream via ba.ctx, OR a
+// test-injected streamStartOverride. Returns the server-assigned stream_no.
+func (ba *BotAPI) dispatchStreamStart(req config.MessageStreamStartReq) (string, error) {
+	if ba.streamStartOverride != nil {
+		return ba.streamStartOverride(req)
+	}
+	return ba.ctx.IMStreamStart(req)
+}
+
+// dispatchStreamEnd closes a WuKongIM message stream via ba.ctx, OR a
+// test-injected streamEndOverride. This is the terminal signal the octo-web
+// client waits on (streamFlag == END) to stop the live "streaming" bubble;
+// a missing END leaves the bubble stuck. The gateway (cc-channel-octo) must
+// always call /v1/bot/stream/end in a finally so END fires even on
+// error/abort/truncation.
+func (ba *BotAPI) dispatchStreamEnd(req config.MessageStreamEndReq) error {
+	if ba.streamEndOverride != nil {
+		return ba.streamEndOverride(req)
+	}
+	return ba.ctx.IMStreamEnd(req)
+}
+
 // NewBotAPI creates the Bot API gateway module.
 func NewBotAPI(ctx *config.Context) *BotAPI {
 	speechURL := os.Getenv(voice_adapter.EnvSpeechServiceURL)
@@ -208,6 +237,13 @@ func (ba *BotAPI) Route(r *wkhttp.WKHttp) {
 	{
 		botAPI.POST("/sendMessage", ba.sendMessage)
 		botAPI.POST("/typing", ba.typing)
+		// Incremental channel streaming (OCT-31). Mirrors the internal
+		// modules/robot stream protocol on the public bot API: open a
+		// stream, send incremental items via /sendMessage with the returned
+		// stream_no, then close it. Both are gated by checkSendPermission and
+		// scoped to the authenticated bot — see stream.go.
+		botAPI.POST("/stream/start", ba.streamStart)
+		botAPI.POST("/stream/end", ba.streamEnd)
 		botAPI.POST("/readReceipt", ba.readReceipt)
 		botAPI.POST("/events", ba.getEvents)
 		botAPI.POST("/events/:event_id/ack", ba.eventAck)
--- a/modules/bot_api/ratelimit_test.go
+++ b/modules/bot_api/ratelimit_test.go
@@ -53,22 +53,20 @@ func setupBotRateLimitEnv(t *testing.T) (http.Handler, func()) {
 	}
 }
 
-// TestBotAPI_PerUIDRateLimit_TripsBucket hammers an authenticated /v1/bot route
-// with one bot token until the per-UID bucket is exhausted, then asserts the
-// limiter returned the shared rate.limited response scoped to "uid".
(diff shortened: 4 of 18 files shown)


mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#414 Review Report
**Head SHA:** `70e7606662dba4c203627db6aceb34baf97eb1f6`
**PR:** feat(bot_api): /v1/bot streaming + per-UID rate limiting (OCT-31, OCT-42, OCT-41)
**Routing:** complexity=security_sensitive (automated review leg)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| OCT-31 stream/start handler | ✅ | `stream.go:38-88` — FromUID forced to robotID, checkSendPermission gates, owner binding recorded |
| OCT-31 stream/end handler | ✅ | `stream.go:120-175` — ownership gate, terminal-END guarantee, binding released |
| OCT-42 setBotActorUID at all 3 auth sites | ✅ | `auth.go:67` (userbot), `auth.go:82` (app registry), `auth.go:110` (app DB) — all call `setBotActorUID` |
| OCT-42 SharedUIDRateLimiter mounted | ✅ | `bot_api.go:247` (`/v1/bot`), `bot_api.go:311` (`/v1/botfile`) — both after `authBot()` |
| OCT-41 stream ownership binding | ✅ | `stream_owner.go` — Redis-backed store with TTL, interface for test injection |
| Group channel_synced migration | ✅ | `sql/20260617000001_group_channel_synced.sql` — idempotent DDL with INFORMATION_SCHEMA guard |
| Group CreateGroup markChannelPendingTx | ✅ | `service.go:1150` — inside tx, before `tx.Commit()` at line 1241 |
| Channel reconcile worker | ✅ | `channel_reconcile.go` — idempotent, grace window, distributed lock with lock-free fallback |
| Test coverage | ✅ | `stream_test.go` (260 lines), `stream_owner_test.go` (234 lines), `ratelimit_test.go` (173 lines), `channel_reconcile_test.go` (195 lines) |

## 2. Findings

### F1 — Reconciler goroutine not stopped on server shutdown (P2)

**Diff-scope:** new (introduced by this PR's group/channel_reconcile.go + api.go changes)

`ChannelReconciler.Stop()` is defined (`channel_reconcile.go:177-181`) but **never called** anywhere in the codebase. The reconciler is started in `Group.New()` → `startChannelReconciler()` (`api.go:119`) with `context.Background()`, which is never cancelled.

**Impact:**
- On SIGTERM, the goroutine is killed by the OS without waiting for an in-flight tick.
- The `redisReconcileLock` client (`channel_reconcile.go:96`) creates a dedicated `rd.Client` with `PoolSize=10` that is never `Close()`d — one Redis connection leaked per shutdown.
- An in-flight `imCreateChannel` RPC may be interrupted, leaving `channel_synced=0` until the next startup's reconcile recovers it (after grace window).

**Why P2 not P1:** All reconcile operations are idempotent. The next startup recovers any interrupted work. The connection leak is bounded (one per shutdown). No user-visible error data or unavailable path.

**Suggestion:** Wire `Stop()` into the server's shutdown hook (e.g., `go-svc` Stop handler or a `context.Context` tied to the server lifecycle). Also call `redisReconcileLock.Close()` in the same path.

### F2 — `botActorUID()` middleware now redundant on incoming webhook route (P2/nit)

**Diff-scope:** amplified (PR introduces `setBotActorUID` which makes the existing `botActorUID()` middleware dead code)

After this PR, `authBot()` calls `setBotActorUID()` at all three auth-success sites (`auth.go:52-55`), which sets both `CtxKeyRobotID` and `"uid"`. The `botActorUID()` middleware on the incoming webhook route (`incoming_webhook.go:33`) now re-reads `robotID` and re-sets `"uid"` to the identical value — pure redundancy.

The one extra guard in `botActorUID()` — aborting with 500 if `robotID` is empty — is unreachable because `authBot()` already aborts on auth failure before `c.Next()`.

**Suggestion:** Remove `ba.botActorUID()` from the incoming webhook middleware chain, or fold its empty-robotID assertion into `setBotActorUID` for defense-in-depth.

### F3 — stream/start dedicated concurrent-open cap deferred to OCT-50 (P2/note)

**Diff-scope:** pre-existing gap, PR improves it (adds per-UID cap) but defers the dedicated stream/start ceiling.

`stream/start` shares the general per-UID bucket (`ratelimit:uid:{robotID}`, 2 rps / burst 60) with all other bot operations. A single bot can open up to 60 streams in a burst before the bucket trips. Each stream creates a live WuKongIM connection + channel bubble — the sharpest amplification primitive on this surface.

The PR documents this deferral (OCT-50) and explains the technical blocker (octo-lib hardcodes the `ratelimit:uid:` bucket prefix). The per-UID cap is a significant improvement over the pre-PR state (zero per-bot limiting).

**No action required for this PR.** Tracked in OCT-50.

### F4 — `QueryReconcilableGroupNos` uses `fmt.Sprintf` for SQL fragment (P2/nit)

**Diff-scope:** new

`db.go:75`: `fmt.Sprintf("created_at < DATE_SUB(NOW(), INTERVAL %d SECOND)", graceSeconds)`. The `%d` format with `int` type makes injection impossible (the value can only be a decimal integer). The value chain is: env var → `strconv.Atoi` → `ReconcileConfig.GraceSec` → `int` parameter. Safe.

**Suggestion (nit):** Consider using a parameterized query for consistency with the rest of the codebase, e.g., `Where("created_at < DATE_SUB(NOW(), INTERVAL ? SECOND)", graceSeconds)`.

## 3. Suggestions

1. **Wire reconciler shutdown** (F1): Add a `Stop()` or `Close()` method to the `Group` module and call it from the server's shutdown path. Also close the `redisReconcileLock` client.
2. **Remove redundant middleware** (F2): Drop `ba.botActorUID()` from the incoming webhook chain now that `setBotActorUID` covers it.
3. **Parameterize SQL** (F4): Use `?` placeholder instead of `fmt.Sprintf` for the grace-seconds fragment.

## 4. Additional Findings

### A1 — Stream ownership bind failure at streamStart is intentionally best-effort

`stream.go:82-85`: If `ba.streamOwners().bind()` fails (e.g., Redis down), the stream opens without an ownership record. At `streamEnd`, the absent binding causes a fall-through (allow), preserving the terminal-END guarantee. The griefing window (co-member bot ending a peer's live bubble) exists only during the compound condition: Redis failure at bind time AND a co-member bot observing the stream_no. This is a documented, intentional design trade-off.

### A2 — `resolveSpaceChannelID` is a no-op

`bot_api.go:329-338`: Currently returns `channelID` unchanged. Both stream handlers call it, but it has no effect. This is consistent with `sendMessage` and other handlers. Noted for completeness.

## 5. Data-Flow Trace

|---|---|---|
| `robotID` in stream handlers | `setBotActorUID` → `CtxKeyRobotID` via `authBot()` | ✅ All 3 auth paths set it; `authBot()` aborts on failure |
| `"uid"` context key for rate limiter | `setBotActorUID` → `c.Set("uid", uid)` | ✅ All 3 auth paths set it; `SharedUIDRateLimiter` mounted after `authBot()` |
| `FromUID` in stream start | Overwritten with `robotID` at `stream.go:63` | ✅ Client-supplied value ignored; forced to authenticated bot |
| Stream ownership binding | Redis `SetAndExpire` at `stream_owner.go:63` | ✅ Best-effort; failure logged, stream still opens |
| `channel_synced` in CreateGroup | `markChannelPendingTx` inside tx (`service.go:1150`) → `MarkChannelSynced` after IM confirm (`service.go:1321`) | ✅ Atomic with group row; flipped only after IM confirmation |
| Orphan detection query | `QueryReconcilableGroupNos` (`db.go:71-78`) → `channel_synced=0 AND status=Normal AND created_at < grace` | ✅ Grace window prevents false positives; only `GroupStatusNormal` selected |
| Reconcile IM recreate subscribers | `GetSubscribableMemberUIDs` → `querySubscribableMemberUIDsWithGroupNo` (`db.go:482-489`) | ✅ Filters `is_deleted=0 AND status=Normal`; excludes blacklisted/soft-deleted members |
| Other group Insert paths (AddGroup, system, org) | DB default `channel_synced=1` | ✅ `channel_synced` not in Model struct; only `CreateGroup` explicitly sets 0 |

### C1 — Dual-path parity
**Clear.** Stream lifecycle is symmetric: start→bind, end→check→release. `setBotActorUID` is called at all 3 auth-success sites (userbot, app-registry, app-DB). No asymmetric add/remove paths found.

### C2 — Control-flow ordering / nested reuse
**Clear.** `setBotActorUID` runs inside `authBot()` before `SharedUIDRateLimiter` in the middleware chain — correct ordering at all mount points. The redundant `botActorUID()` on the incoming webhook route re-sets `"uid"` to the same value (harmless). No nested/double-invocation risk.

**Clear.** Stream endpoints are gated by: (1) `authBot()` — valid bot token required, (2) `checkSendPermission` — channel membership verified, (3) `SharedUIDRateLimiter` — per-bot rate cap, (4) stream ownership check on `stream/end`. No unauthenticated path reaches stream handlers. The `dispatchStreamStart`/`dispatchStreamEnd` methods are only reachable through the authenticated handlers.

### C4 — Authorization lifecycle / container-member state cascade
**N/A.** This PR does not modify auth/lifecycle logic. The rate limiter keys on `robotID`; if a bot token is revoked, `authBot()` rejects before the limiter. Space-level bot disabling is a pre-existing concern outside this PR's diff scope.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The PR demonstrates strong security engineering: proper auth→rate-limit ordering at all three auth resolution paths, server-authoritative FromUID scoping on stream/start, Redis-backed stream ownership binding with intentional fail-open for the terminal-END guarantee, and atomic channel_synced marking with idempotent reconcile recovery. Four P2 findings (reconciler shutdown gap, redundant middleware, deferred stream/start cap, SQL style nit) are non-blocking.

yujiawei (changes requested):
# Code Review — PR #414 (octo-server)

Reviewed at head `70e7606662dba4c203627db6aceb34baf97eb1f6` (branch `oct-31-bot-stream-api`, base `main`). This is a security-sensitive change (public bot-API streaming + per-UID rate limiting + a group-channel reconcile worker), so the streaming auth surface got the closest scrutiny. The review is static analysis + reading the new unit tests; no live MySQL/Redis/WuKongIM run.

**Verdict: Changes requested** — one blocking security gap (P1) in the streaming surface; everything else is advisory.

## What's solid (verified, no change needed)

- **`FromUID` is unforgeable on `stream/start`.** `stream.go:61` unconditionally sets `req.FromUID = robotID` after `BindJSON`, and `checkSendPermission` runs first (`stream.go:53`). `TestStreamStart_ScopesFromUIDToBot` proves a forged `FromUID` is overwritten. Good.
- **Per-UID rate limiting (OCT-42) works end-to-end.** All three `authBot` success paths (`authUserBot` `auth.go:70`, App-bot registry `auth.go:82`, App-bot DB fallback `auth.go:110`) call `setBotActorUID`, which mirrors `robotID` onto the `uid` key that `UIDRateLimitMiddleware` reads; the limiter is mounted *after* `authBot` on both `/v1/bot` and `/v1/botfile`. Double-mount of the process-wide singleton is safe. The fail-open-on-missing-uid posture is consistent with the global per-IP floor.
- **Stream-ownership gate (OCT-41) on `stream/end`** correctly denies a confirmed owner mismatch and is covered by `TestStreamEnd_RejectsForeignOwner` (asserts `IMStreamEnd` does not fire and the binding survives a denied end).
- **Reconcile worker / `channel_synced` (octo-server#394).** `markChannelPendingTx` runs in the *same tx* as the group `Insert` (`service.go:1129` + `1150`, commit at `1248`), so there is no window where the row is visible as `synced=1`. `channel_synced` is deliberately kept off the `Model` struct, so system/register/`AddGroup` insert paths keep DB default `1` — no false-orphan regression. The migration's INFORMATION_SCHEMA-guarded column+index ALTERs are idempotent/re-entrant (no Duplicate-column CrashLoop on rolling deploy). The lock release is a proper token-aware CAS-DEL, and the lock-free degraded path is safe because recreate+flip are idempotent. `i18n` marker + zh-CN translation for the new error code are both present and the `4xx ⟺ Internal=false` invariant holds.

## P1 — blocking

### 1. Stream ownership is enforced on `stream/end` but NOT on the `sendMessage` append path

`modules/bot_api/send.go:248`

The streaming protocol is: open via `/v1/bot/stream/start`, append each incremental chunk via `/v1/bot/sendMessage` carrying the returned `stream_no`, close via `/v1/bot/stream/end`. OCT-41 added a `stream_no → robotID` ownership binding and checks it in `streamEnd` (`stream.go:149`) so a co-member bot cannot terminate a peer's live bubble. But the **append** path applies no such check:

```go
msgReq := &config.MsgSendReq{
    ...
    StreamNo:    req.StreamNo,   // send.go:248 — client-supplied, no ownership check
    FromUID:     fromUID,
    Payload:     []byte(util.ToJson(wirePayload)),
}
```

`sendMessage` forwards the client-supplied `req.StreamNo` straight into `MsgSendReq`; the only gate is `checkSendPermission` on the *channel*, which proves "may post here", not "owns this stream". A `stream_no` is observable by any co-member bot — `MessageResp` carries `stream_no` and it is returned via `/v1/bot/messages/sync`. WuKongIM resolves a stream by `stream_no` alone (the PR's own load-bearing premise for OCT-41), so an appended chunk lands on that stream regardless of channel.

**Scenario:** Bots A and B are both members of group G. A opens a stream and is streaming token-by-token. B calls `/messages/sync`, reads A's live `stream_no` S, then calls `/sendMessage{channel_id=G, stream_no=S, payload=…}`. B passes `checkSendPermission` (legitimate member) and, with no ownership gate on append, B's chunk is dispatched into A's live stream. This is the *same* attacker model OCT-41 was built to stop (co-member bot manipulating a peer's stream via an observable `stream_no`) — OCT-41 closed the *terminate* vector but left the *append/inject* vector fully open. Content-injection into a peer's stream is arguably worse than the premature-END griefing that was fixed.

Two honest bounds on impact (why P1, not P0): on append `FromUID` is B's own id, not spoofed to A; and whether the client visually attributes the injected chunk to A's bubble depends on the deployed octo-im fork's render keying, which isn't verifiable from this repo. Worst case is cross-bot stream content-injection; degraded case is the chunk rendering under B or as noise. Either way it's a reachable security gap in a security-sensitive surface.

**Fix:** mirror the `streamEnd` ownership check in the append path — when `req.StreamNo != ""`, verify `ba.streamOwners().owner(req.StreamNo)` is empty or equals the caller's `robotID` before forwarding, using the same deny-on-confirmed-mismatch / fail-open-on-absent posture already established for `stream/end`.

## P2 — advisory (non-blocking)

1. **Fail-open on `stream/end` is unsound for foreign-but-live streams.** `stream.go:149-157` allows the end when no binding is found. The binding store is package-local to `bot_api`; the internal `/v1/robot` stream path (`modules/robot/api.go:305-322`) opens streams in the *same* WuKongIM stream space with **no** binding. The comment's justification ("a stale/foreign `stream_no` has no live bubble to grief") is false for a `/v1/robot`-opened *live* stream: a co-member bot could end it via the absent-binding branch. Exploitability hinges on whether any production component actually opens live streams via `/v1/robot` (the gateway/web repos aren't in this workspace) — if so, this escalates to P1. Worth a maintainer's confirmation of the deployed topology.

2. **A bind-failure at `stream/start` silently downgrades the end gate to fail-open for that stream's whole life.** `stream.go:76-79` treats a Redis `bind` failure as non-fatal (the bubble is already live). For that stream there is then no owner record, so any co-member bot can end it via the absent-binding fail-open — the exact attack OCT-41 prevents, now triggerable by a transient Redis blip during open. This is a deliberate, documented trade-off; consider failing closed on absent-binding specifically for *group*-channel streams (where co-member peers exist) while keeping fail-open for DM streams.

3. **Zombie group row when the compensating `group` delete fails but `group_member` delete succeeds.** `service.go:1310-1315` runs two best-effort deletes in order; a partial failure (member-delete ok, group-delete fails) leaves a `status=Normal`, memberless, `channel_synced=0` row. Reconcile then hits the `len(uids)==0` branch (`channel_reconcile.go:240`) and flips it to `synced=1`, so it is never revisited — a permanently stranded zombie, contrary to the PR's "orphans can be recovered, not stranded" goal. Consider reversing the delete order (group first) or hard-deleting memberless `Normal` orphans in reconcile.

4. **Reconcile lock `TTL == Interval` can expire mid-tick.** `channel_reconcile.go` defaults `LockTTL` to `Interval` (120s), but `RunOnce` holds the lock across up to `BatchSize=200` sequential IM POSTs. A slow-IM + large-backlog tick exceeding 120s lets a second pod acquire and run concurrently. Correctness holds (recreate+flip are idempotent; CAS-DEL prevents lock theft), but it defeats the dedup the lock exists for. Set `LockTTL` to a multiple of `Interval`, or renew.

5. **`incoming_webhook.go` still mounts `botActorUID()`** alongside `authBot()` + `SharedUIDRateLimiter`. After this PR `authBot` already sets `uid` via `setBotActorUID`, so `botActorUID()` is a redundant duplicate set whose only unique behavior (the `robotID==""` 500-abort) is now unreachable. The main `/v1/bot` and `/v1/botfile` groups already dropped it. Collapse to a single source of truth to avoid future divergence.

6. **Smaller nits:** (a) the `ratelimit_test` first-request `X-RateLimit-Scope:uid` assertion proves the middleware *ran* (it fails open with no header when `uid` is unset), but the comment overstates it as proof of Redis *enforcement* — the genuine proof is the later 429 trip; (b) the `metrics.go:33` "skipped" label doc says "fixed by another instance", but its only increment site (`channel_reconcile.go:247`) fires on the memberless-orphan path — operator-facing drift; (c) `db.go:75` inlines `graceSeconds` via `fmt.Sprintf` into the SQL — safe today (int-constrained, and MySQL `INTERVAL N SECOND` can't take a placeholder), but a `%d`-into-SQL pattern worth a comment so a future refactor doesn't copy it with a string input.

## Coverage / what I could not verify

- The deployed WuKongIM (octo-im fork) append-render semantics for a chunk whose `FromUID` differs from the stream's original — the load-bearing uncertainty behind the P1's medium confidence.
- Whether any production component opens live streams via the internal `/v1/robot` path rather than `/v1/bot/stream/start` (gateway/web repos not in this workspace) — gates the P2 #1 severity.
- No `go build` / `go test` / `make i18n-extract-check` executed (requires live MySQL/Redis/WuKongIM); all build/i18n/test conclusions are from static cross-file reading of the pinned `octo-lib` sources.

yujiawei:
One more P2 to fold into the requested changes (caught on a second pass over the reconcile worker's lifecycle):

**Reconciler goroutine and its Redis client are never stopped on shutdown.** `ChannelReconciler.Start` is called with `context.Background()` (`modules/group/api.go:119`), which is never cancelled, and `Stop()` (`channel_reconcile.go`) has no caller anywhere — the `Group` module exposes no shutdown hook. Consequently `redisReconcileLock.Close()` (`channel_reconcile.go:116`) is never invoked, so the dedicated `PoolSize=10` Redis client leaks one connection per process shutdown, and an in-flight tick is killed by the OS rather than drained.

P2, not blocking on its own: all reconcile ops are idempotent so the next startup recovers any interrupted work, and the leak is bounded to one client per shutdown. Worth wiring `Stop()` + `redisReconcileLock.Close()` into the server lifecycle (a cancellable context tied to module shutdown) while the other changes are being made.

Jerry-Xin (changes requested):
Reviewed at head `70e7606662db`. The OCT-31 streaming + OCT-42 per-UID rate-limit wiring is solid, but the OCT-41 stream-ownership guard is incomplete: it protects `stream/end` but not the incremental-append path. One blocking item.

## 🔴 Blocking — stream ownership is enforced on END but not on APPEND (stream hijack / content injection)

The PR's own design (stream.go) documents that incremental chunks are sent **via `/v1/bot/sendMessage` carrying the returned `stream_no`** ("send.go already forwards StreamNo to MsgSendReq"). And `stream_owner.go` states the threat model explicitly: a co-member bot can *observe* a peer's live `stream_no` because it is rendered into the streaming bubble.

But the ownership binding is only consulted in `streamEnd`:
- `stream.go` (streamEnd): `if owner, _ := ba.streamOwners().owner(req.StreamNo); owner != "" && owner != robotID { ... ErrBotAPIStreamNotOwned }` ✅

It is **not** consulted on the append path:
- `send.go:28` — `StreamNo string \`json:"stream_no"\`` (caller-supplied)
- `send.go:40` `sendMessage` → only `checkSendPermission` (channel-level gate), **no `streamOwners().owner()` check**
- `send.go:248` — `StreamNo: req.StreamNo` forwarded straight into `MsgSendReq`

Consequence: a co-member bot B that can post to the same channel (passes `checkSendPermission`) and has observed bot A's live `stream_no` can call `/sendMessage` with A's `stream_no` and inject arbitrary chunks into A's live streaming bubble. The OCT-41 protection guards termination but leaves mid-stream content tampering open — the same griefing/spoofing class the ownership binding was introduced to close, just via APPEND instead of END.

Fix: apply the same `streamOwners().owner(req.StreamNo)` ownership check in `sendMessage` whenever `req.StreamNo != ""` (deny on confirmed mismatch, same fail-open-on-missing posture as streamEnd for consistency). This matches yujiawei's finding.

## 🟡 Non-blocking
- `ChannelReconciler.Stop()` not wired into graceful shutdown (mochashanyao) → Redis conn lingers until process exit. Should-fix; low impact since it only leaks at shutdown, but wire it into the shutdown hook for cleanliness.
- `SharedUIDRateLimiter` is mounted after `authBot()` keyed on uid=robotID — correct placement closing the OCT-42 stream-amplification gap. (Limiter impl itself is pre-existing infra, out of scope here.)
- `resolveSpaceChannelID` is a documented no-op placeholder for both sendMessage and streamStart — no space_id asymmetry between the two paths (ruled out a suspected cross-space concern).

## ✅ Highlights
- Ownership binding keyed solely on `stream_no` (independent of channel_id) with bounded TTL — right call given stream_no is the only stable handle.
- streamEnd fail-open-on-missing posture is well-reasoned and documented (preserves terminal-END guarantee without reopening the grief window for stale streams).
- Good test coverage on bind/owner/end ownership paths.

CI is green (Build/Test/Vet show "skipping" via path-filter, not failures). Verdict is REQUEST_CHANGES solely on the append-path ownership gap.

PR #78, opened 2026-05-18
fix(common): rich multipart envelope for /test_email to bypass anti-spam silent drops

an9xyz:
## Summary

Follow-up to #76. Verified end-to-end that the test_email body shipped in #76 (single `<p>...</p>` text/html part) is silently dropped by recipient anti-spam filters on at least one production path (Aliyun corporate SMTP → mininglamp.com): SMTP accepts, no bounce returned to sender, no entry in inbox / spam / deleted. The verification-code email on the same SMTP link delivers fine — so the link itself is healthy; the issue is content shape (no plaintext alternative, no transactional-email headers).

This PR plumbs a richer envelope through `EmailService` and updates the admin "Test email" handler to use it.

## What changed

- **`modules/base/common/service_email.go`** — new `SendTransactionalHTML(ctx, to, subject, htmlBody, plainBody)` method on `EmailService`. Wraps body in `multipart/alternative`, sets `Date` / `Message-ID` (with From's domain) / `Auto-Submitted: auto-generated` / `Precedence: bulk` / `List-Unsubscribe` headers — the signals anti-spam scoring uses to distinguish transactional mail from unsolicited HTML.
  - The existing `SendHTMLEmail` / `sendEmail` path stays as-is so verification codes and invite emails (`modules/space/email_invite_sender.go`) keep their lighter envelope.
  - `dispatchSMTP` and `sanitizeHeader` are factored out so both code paths share TLS dialing, AUTH, the SMTP transaction, and the CRLF-injection guard.

- **`modules/common/api_manager_system_setting.go`** — `testSystemSettingEmail` now:
  - Calls `SendTransactionalHTML` with the v5 layout: single bilingual title, amber "Do Not Reply" banner, white card with three ✓ rows (中文 primary, English secondary right-aligned), bilingual conclusion.
  - Writes a pre-send `Info` log with effective from/smtp, so a future "test sent but didn't arrive" report can be diagnosed from server logs in seconds rather than walking the SMTP path by hand.
  - Masks the raw SMTP error from the HTTP response (logged server-side instead) — closes one of the P2s yujiawei raised in #76 about leaking infra detail to the admin UI.

## Verification

End-to-end inbox proof attached to the originating Conductor session: the same recipient mailbox that silently dropped the old single-part body received the new multipart envelope into the inbox, rendered correctly (banner + status card + bilingual conclusion).

Unit tests cover the bits we can verify offline:

- multipart/alternative structure and boundary
- plaintext-before-html ordering per RFC 2046
[...]

PR #415, opened 2026-06-18
feat(searchetl): enrich producer with SpaceID/Visibles/MessageSeq (contract v2, fail-closed)

yujiawei:
## What

Enrich the `searchetl` Kafka producer to fill the three v2 contract safety/correctness fields the search reader requires, and re-pin `octo-lib` to the v2 contract + shared visibility parser.

- **MessageSeq**: add `message_seq` to the message-table SELECT (both `readBatch` and `readStableBatchTx`) and carry it through `srcMessageRow` (typed `uint64` to match the BIGINT column / `uint64` contract field — no truncation on high-sequence channels) into the contract.
- **SpaceID / Visibles**: parse from the raw payload via `octo-lib`'s shared `searchmsg.ExtractVisibility` — the single source of truth now shared with the indexer's backfill, so the visibility semantic does **not** fork across repos.

## Security: fail-closed visibility (prevents reader fail-OPEN / over-exposure)

The reader treats an empty `visibles` array as **fail-OPEN** (no gate). So for **non-encrypted** messages, if visibility cannot be trusted, the whole row is routed to **DLQ** — never emitted with empty `visibles`:

- payload not exactly one JSON object (corrupt / array / scalar / `null` / trailing data) → DLQ
- `visibles` key present but unparseable OR **valid-but-empty** (`null` / `[]` / all-non-string) → DLQ
- `visibles` key absent → genuine broadcast message → pass through (`visibles=nil`, expected & safe)

DLQ rows still advance the shard cursor (no stall) and are produced to the DLQ topic (never silently dropped). Encrypted DM payloads are ciphertext: not parsed; `spaceId`/`visibles` left empty (reader fail-closed; body already `raw_excluded`).

`space_id` is type-tolerant: only taken when a JSON string, otherwise empty — without dropping a legitimate `visibles` array.

## Same vectors both sides

Producer and indexer backfill run the **identical** `searchmsg.FailClosedVisibilityVectors()` set (15 cases incl. valid-but-empty, trailing-delimiter, type-tolerant), locking one semantic across repos.

## Test

`go build ./...`, `go vet`, `go test ./modules/searchetl/...` all green. New `modules/searchetl/visibility_test.go` covers the shared vectors + targeted fail-closed/broadcast/full-width-seq cases.

## Merge order

⚠️ Depends on the `octo-lib` change (shared `ExtractVisibility`). Merge `Mininglamp-OSS/octo-lib#80` first, then this PR will re-pin to the merged commit.

Part of the Route A real-time search unblock chain.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	firebase.google.com/go/v4 v4.13.0
 	github.com/BurntSushi/toml v1.6.0
 	github.com/DATA-DOG/go-sqlmock v1.5.0
-	github.com/Mininglamp-OSS/octo-lib v0.0.0-20260614033907-f62e626a34db
+	github.com/Mininglamp-OSS/octo-lib v0.0.0-20260618083705-a57dc301cbf3
 	github.com/alibabacloud-go/darabonba-openapi v0.2.1
 	github.com/alibabacloud-go/sms-intl-20180501 v1.0.1
 	github.com/alibabacloud-go/tea v1.2.1
--- a/go.sum
+++ b/go.sum
@@ -65,8 +65,8 @@ github.com/HdrHistogram/hdrhistogram-go v1.1.2 h1:5IcZpTvzydCQeHzK4Ef/D5rrSqwxob
 github.com/HdrHistogram/hdrhistogram-go v1.1.2/go.mod h1:yDgFjdqOqDEKOvasDdhWNXYg9BVp4O+o5f6V/ehm6Oo=
 github.com/MicahParks/keyfunc v1.9.0 h1:lhKd5xrFHLNOWrDc4Tyb/Q1AJ4LCzQ48GVJyVIID3+o=
 github.com/MicahParks/keyfunc v1.9.0/go.mod h1:IdnCilugA0O/99dW+/MkvlyrsX8+L8+x95xuVNtM5jw=
-github.com/Mininglamp-OSS/octo-lib v0.0.0-20260614033907-f62e626a34db h1:FudjIbMVhMga/sUsysfzKBI7LTdz0Rgw9EWU2tEJzAo=
-github.com/Mininglamp-OSS/octo-lib v0.0.0-20260614033907-f62e626a34db/go.mod h1:uOCTTI3QdG0ETKZ6uluRWUZgKexee/pY1pQYWyIEvZ0=
+github.com/Mininglamp-OSS/octo-lib v0.0.0-20260618083705-a57dc301cbf3 h1:bdZVEI6F7b60DVtHB3G4N8aJ4ikzAywepS5Ebyk5oH0=
+github.com/Mininglamp-OSS/octo-lib v0.0.0-20260618083705-a57dc301cbf3/go.mod h1:uOCTTI3QdG0ETKZ6uluRWUZgKexee/pY1pQYWyIEvZ0=
 github.com/RichardKnop/logging v0.0.0-20190827224416-1a693bdd4fae h1:DcFpTQBYQ9Ct2d6sC7ol0/ynxc2pO1cpGUM+f4t5adg=
 github.com/RichardKnop/logging v0.0.0-20190827224416-1a693bdd4fae/go.mod h1:rJJ84PyA/Wlmw1hO+xTzV2wsSUon6J5ktg0g8BF2PuU=
 github.com/RichardKnop/machinery/v2 v2.0.11 h1:BTfLGOmOju3W/OtlZmLX26OjYNZsU4PJo04pQReycdc=
--- a/modules/searchetl/etl_db.go
+++ b/modules/searchetl/etl_db.go
@@ -78,7 +78,7 @@ func (d *etlDB) maxID(table string) (int64, error) {
 func (d *etlDB) readBatch(table string, cursor int64, batch int) ([]*srcMessageRow, error) {
 	var rows []*srcMessageRow
 	_, err := d.session.SelectBySql(
-		fmt.Sprintf("SELECT id, message_id, from_uid, channel_id, channel_type, setting, `signal`, "+
+		fmt.Sprintf("SELECT id, message_id, message_seq, from_uid, channel_id, channel_type, setting, `signal`, "+
 			"`timestamp`, UNIX_TIMESTAMP(created_at) AS created_unix, payload "+
 			"FROM `%s` WHERE id>? ORDER BY id ASC LIMIT ?", table),
 		cursor, batch).Load(&rows)
@@ -105,7 +105,7 @@ func (d *etlDB) readStableBatchTx(table string, batch int) (cursor int64, rows [
 		return 0, nil, err
 	}
 	if _, err = tx.SelectBySql(
-		fmt.Sprintf("SELECT id, message_id, from_uid, channel_id, channel_type, setting, `signal`, "+
+		fmt.Sprintf("SELECT id, message_id, message_seq, from_uid, channel_id, channel_type, setting, `signal`, "+
 			"`timestamp`, UNIX_TIMESTAMP(created_at) AS created_unix, payload "+
 			"FROM `%s` WHERE id>? ORDER BY id ASC LIMIT ?", table),
 		cursor, batch).Load(&rows); err != nil {
--- a/modules/searchetl/model.go
+++ b/modules/searchetl/model.go
@@ -9,6 +9,7 @@ package searchetl
 type srcMessageRow struct {
 	ID          int64
 	MessageID   string
+	MessageSeq  uint64 // 频道内消息序号（message.message_seq BIGINT 列，全精度）——reader「清空会话」channel_offset gate 依赖；契约 v2 字段亦为 uint64
 	FromUID     string
 	ChannelID   string
 	ChannelType uint8
--- a/modules/searchetl/payload.go
+++ b/modules/searchetl/payload.go
@@ -28,13 +28,21 @@ const (
 //
 // P1-d 规则：
 //   - Signal 加密（setting 的 Signal 位 或 signal 列为真）→ payload 非明文 JSON，
-//     直接 raw_excluded（不尝试解析，避免把密文当损坏 JSON 误判进 DLQ）。
+//     直接 raw_excluded（不尝试解析，避免把密文当损坏 JSON 误判进 DLQ）。spaceId/visibles
+//     留空走 reader fail-closed（安全：正文本就 raw_excluded，不可检索）。
 //   - 非 Signal → payload 应是明文 JSON。解析失败（本应可解析却失败）→ DLQ。
 //   - 解析成功后按 type 三态（float64/int/json.Number，复用 message.CoerceTextPayloadContent
 //     的兼容口径）取 content：
 //       · type=Text 且 content 为 string → 取该 string 作正文。
 //       · 非 Text 或 content 非 string（媒体/富文本/结构化对象）→ 本期保守 raw_excluded
 //         （阶段 4+ 可细化为序列化可检索文本，契约字段不变）。
+//
+// 🔴 v2 富化 + fail-closed 可见性（防 #1124 重演）：非加密消息额外用 octo-lib 共享的
+// searchmsg.ExtractVisibility 从 payload 抽 SpaceID/Visibles 填进契约 v2 字段。一旦可见性
+// 无法可信解析（payload 非 JSON 对象、visibles 键在却 unparseable 或 valid-but-empty）→
+// **整条落 DLQ**，绝不写空 Visibles —— 否则 reader(visibility.go) 把空 visibles 当 fail-OPEN
+// （普通成员搜出「仅指定成员可见」的群定向系统消息）。MessageSeq 从 message.message_seq 列取。
+// producer 与 backfill 共用同一 ExtractVisibility + 同一组 fail-closed 向量锁口径（验收门 ii）。
 //
 // 返回的 Message 已填 SchemaVersion / Source / 可见性字段；outcome 决定投正文 topic 还是 DLQ。
 func extractMessage(row *srcMessageRow) (searchmsg.Message, extractOutcome) {
@@ -46,11 +54,13 @@ func extractMessage(row *srcMessageRow) (searchmsg.Message, extractOutcome) {
 		FromUID:       row.FromUID,
 		MsgTimestamp:  row.Timestamp,
 		CreatedAt:     row.CreatedUnix,
+		MessageSeq:    row.MessageSeq,
 		Source:        searchmsg.SourceETLMessageTable,
 	}
 
 	if isSignalEncrypted(row) {
-		// Signal 加密 DM：payload 是密文，解不出明文是预期行为，非异常。
+		// Signal 加密 DM：payload 是密文，解不出明文是预期行为，非异常。spaceId/visibles 留空
+		// （reader fail-closed，安全）。
 		msg.RawExcluded = true
 		msg.Content = nil
 		return msg, outcomeRawExcluded
@@ -61,6 +71,15 @@ func extractMessage(row *srcMessageRow) (searchmsg.Message, extractOutcome) {
 		// 非加密消息本应是明文 JSON，解析失败/空 map 属真异常 → DLQ（游标仍推进）。
 		return msg, outcomeDLQ
 	}
+
+	// 🔴 富化可见性（fail-closed）：可见性是 access-control ACL。无法可信解析 → 整条 DLQ，
+	// 绝不写空 Visibles 让 reader fail-OPEN。space_id 非字符串类型不连累合法 visibles（V3b 隔离）。
+	spaceID, visibles, verr := searchmsg.ExtractVisibility(row.Payload)
+	if verr != nil {
+		return msg, outcomeDLQ
+	}
+	msg.SpaceID = spaceID
+	msg.Visibles = visibles
 
 	contentType, isText := payloadType(m)
 	msg.ContentType = contentType
--- a/modules/searchetl/visibility_test.go
+++ b/modules/searchetl/visibility_test.go
@@ -1,0 +1,161 @@
+package searchetl
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/contract/searchmsg"
+)
+
+// TestExtractMessage_SharedFailClosedVectors 跑 octo-lib 共享的 fail-closed 可见性向量集
+// （验收门 (i)+(ii)）：producer 与 octo-search-indexer backfill 跑**同一组**向量锁口径。
+//
+// 安全口径（ReviewBot YUJ-4953 钉死）：非加密群消息 visibles
+//   - 解析失败 → outcomeDLQ（绝不进 main topic）
+//   - valid-but-empty（null / [] / 全非字符串）→ 同样 outcomeDLQ
+// visibles 键缺失 = 广播消息 → outcomeOK 进 main，visibles 为空（reader 无 gate 属预期安全）。
+func TestExtractMessage_SharedFailClosedVectors(t *testing.T) {
+	for _, v := range searchmsg.FailClosedVisibilityVectors() {
+		v := v
+		t.Run(v.Name, func(t *testing.T) {
+			// 非加密 type=Text 行：直接用向量 payload（向量本身已含 type/content）。
+			row := &srcMessageRow{MessageID: "m_" + v.Name, ChannelType: common.ChannelTypeGroup.Uint8(), Payload: v.Payload}
+			msg, outcome := extractMessage(row)
+
+			if v.WantErr {
+				// 🔴 非加密、可见性不可信 → 必须落 DLQ，绝不进 main topic。
+				if outcome != outcomeDLQ {
+					t.Fatalf("%s: visibility fail-closed must route to DLQ, got outcome=%v visibles=%v", v.Name, outcome, msg.Visibles)
+				}
+				// DLQ 消息绝不能携带空 visibles 假装广播进主流（即便它最终不投 main，也防误用）。
+				if len(msg.Visibles) != 0 {
+					t.Fatalf("%s: DLQ msg must not carry visibles, got %v", v.Name, msg.Visibles)
+				}
+				return
+			}
+
+			// 放行类：必须不是 DLQ，且 visibles/spaceID 与期望一致。
+			if outcome == outcomeDLQ {
+				t.Fatalf("%s: must not DLQ, got outcome=%v", v.Name, outcome)
+			}
+			if msg.SpaceID != v.WantSpaceID {
+				t.Fatalf("%s: spaceID=%q want %q", v.Name, msg.SpaceID, v.WantSpaceID)
+			}
+			if !reflect.DeepEqual(msg.Visibles, v.WantVisibles) {
+				t.Fatalf("%s: visibles=%v want %v", v.Name, msg.Visibles, v.WantVisibles)
+			}
+		})
+	}
+}
+
+// TestExtractMessage_EmptyVisiblesGroupMsgDLQ 单独钉死验收门 (i) 的核心：
+// 非加密群消息 valid-but-empty visibles → DLQ 不进 main（ReviewBot 特别强调 population 主落点）。
+func TestExtractMessage_EmptyVisiblesGroupMsgDLQ(t *testing.T) {
+	row := &srcMessageRow{
+		MessageID:   "grp-empty-vis",
+		ChannelType: common.ChannelTypeGroup.Uint8(),
+		Payload:     []byte(`{"type":99,"content":"you were removed","visibles":[]}`),
+	}
+	_, outcome := extractMessage(row)
+	if outcome != outcomeDLQ {
+		t.Fatalf("group msg with valid-but-empty visibles MUST DLQ (else reader fail-OPEN #1124), got %v", outcome)
+	}
+}
+
+// TestExtractMessage_UnparseableVisiblesGroupMsgDLQ 非加密群消息 visibles 解析失败 → DLQ。
+func TestExtractMessage_UnparseableVisiblesGroupMsgDLQ(t *testing.T) {
+	row := &srcMessageRow{
+		MessageID:   "grp-bad-vis",
+		ChannelType: common.ChannelTypeGroup.Uint8(),
+		Payload:     []byte(`{"type":99,"content":"x","visibles":"not-an-array"}`),
+	}
+	_, outcome := extractMessage(row)
+	if outcome != outcomeDLQ {
+		t.Fatalf("group msg with unparseable visibles MUST DLQ, got %v", outcome)
+	}
+}
+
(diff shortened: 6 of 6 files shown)


lml2468 (approved):
✅ APPROVE — 模范级的 search producer 可见性富化 PR(contract v2,fail-closed)。这是「搜索索引可见性必须对齐读路径全集」在 producer 侧的正解,fail-closed 真到位。逐项亲核 + 26 个测试全过,无 blocking。

## fail-closed 真到位(核心,防定向消息泄露)
非加密群消息经 searchmsg.ExtractVisibility 抽 SpaceID/Visibles:一旦可见性无法可信解析(payload 非 JSON 对象 / visibles 键在却 unparseable / valid-but-empty)→ **整条落 DLQ,绝不写空 Visibles**。这堵死了 reader(visibility.go)把空 visibles 当 fail-OPEN → 普通成员搜出「仅指定成员可见」的群定向系统消息的泄露(#1124 类)。正是我此前在 reader 侧(#361/#385)担心的「producer 是否真填了 visibles」的正确落地——producer 宁可 DLQ 也不放空白名单。

## 依赖的 octo-lib ExtractVisibility 极严谨(亲核)
- payload 非对象 / 顶层 null(Decode 成 nil map 不报错)/ 尾部多余字节(`{...}]`、`{...} garbage`)全部 fail-closed —— 连 dec.More() 会放过 `{...}]` 的坑都用「二次 Decode 要求 io.EOF」堵了。
- visibles:键缺失=广播放行(安全);键在但 null / 非数组 / valid-but-empty([]/全非字符串/全空串,resolves to zero valid UIDs)全部 fail-closed。与 octo-server 既有 visiblesAllows 口径一致(非字符串元素跳过)。
- space_id 类型容忍:仅取 JSON 字符串,非字符串留空且不连累合法 visibles(隔离)。

## MessageSeq 精度(避开 #385 的 float64 截断坑)
srcMessageRow.MessageSeq 用 uint64 对齐 message.message_seq BIGINT 列;TestExtractMessage_MessageSeqFullWidth 用 1<<40(远超 uint32)验证全宽无截断。reader「清空会话」channel_offset gate 依赖此精度。readBatch/readStableBatchTx 两条读路径 SELECT 都加了 message_seq。

## 其余
- 加密 Signal 消息:raw_excluded + spaceID/visibles 留空(reader fail-closed 安全,因正文本就不可检索),且不因密文非 JSON 误 DLQ。
- producer 与 octo-search-indexer backfill 共用同一 searchmsg.FailClosedVisibilityVectors 向量集锁口径(验收门 ii),保证两侧 fail-closed 行为一致。
- go.mod/sum 重 pin octo-lib v2(含 ExtractVisibility + 共享向量);只动版本,无其他依赖漂移。

## 验证
go vet 干净;26 个测试全 PASS——共享向量集 19 case(broadcast/valid/tolerant/failclosed)+ 7 专项(valid-but-empty→DLQ、unparseable→DLQ、valid→enriched、广播→OK、MessageSeq 全宽、加密→raw_excluded)。本地实跑(纯解析逻辑,不依赖 DB/Kafka)。

无 blocking、无 should-fix。批准。这是 search 可见性契约 producer 侧 fail-closed 的标杆实现。

—— Octo-PR (lml2468)

lml2468 (approved):
✅ APPROVE — 模范级的 search producer 可见性富化 PR(contract v2,fail-closed)。这是「搜索索引可见性必须对齐读路径全集」在 producer 侧的正解,fail-closed 真到位。逐项亲核 + 26 个测试全过,无 blocking。

## fail-closed 真到位(核心,防定向消息泄露)
非加密群消息经 searchmsg.ExtractVisibility 抽 SpaceID/Visibles:一旦可见性无法可信解析(payload 非 JSON 对象 / visibles 键在却 unparseable / valid-but-empty)→ **整条落 DLQ,绝不写空 Visibles**。这堵死了 reader(visibility.go)把空 visibles 当 fail-OPEN → 普通成员搜出「仅指定成员可见」的群定向系统消息的泄露(#1124 类)。正是我此前在 reader 侧(#361/#385)担心的「producer 是否真填了 visibles」的正确落地——producer 宁可 DLQ 也不放空白名单。

## 依赖的 octo-lib ExtractVisibility 极严谨(亲核)
- payload 非对象 / 顶层 null(Decode 成 nil map 不报错)/ 尾部多余字节(`{...}]`、`{...} garbage`)全部 fail-closed —— 连 dec.More() 会放过 `{...}]` 的坑都用「二次 Decode 要求 io.EOF」堵了。
- visibles:键缺失=广播放行(安全);键在但 null / 非数组 / valid-but-empty([]/全非字符串/全空串,resolves to zero valid UIDs)全部 fail-closed。与 octo-server 既有 visiblesAllows 口径一致(非字符串元素跳过)。
- space_id 类型容忍:仅取 JSON 字符串,非字符串留空且不连累合法 visibles(隔离)。

## MessageSeq 精度(避开 #385 的 float64 截断坑)
srcMessageRow.MessageSeq 用 uint64 对齐 message.message_seq BIGINT 列;TestExtractMessage_MessageSeqFullWidth 用 1<<40(远超 uint32)验证全宽无截断。reader「清空会话」channel_offset gate 依赖此精度。readBatch/readStableBatchTx 两条读路径 SELECT 都加了 message_seq。

## 其余
- 加密 Signal 消息:raw_excluded + spaceID/visibles 留空(reader fail-closed 安全,因正文本就不可检索),且不因密文非 JSON 误 DLQ。
- producer 与 octo-search-indexer backfill 共用同一 searchmsg.FailClosedVisibilityVectors 向量集锁口径(验收门 ii),保证两侧 fail-closed 行为一致。
- go.mod/sum 重 pin octo-lib v2(含 ExtractVisibility + 共享向量);只动版本,无其他依赖漂移。

## 验证
go vet 干净;26 个测试全 PASS——共享向量集 19 case(broadcast/valid/tolerant/failclosed)+ 7 专项(valid-but-empty→DLQ、unparseable→DLQ、valid→enriched、广播→OK、MessageSeq 全宽、加密→raw_excluded)。本地实跑(纯解析逻辑,不依赖 DB/Kafka)。

无 blocking、无 should-fix。批准。这是 search 可见性契约 producer 侧 fail-closed 的标杆实现。

—— Octo-PR (lml2468)

Jerry-Xin (approved):
✅ **APPROVE** — producer-side enrichment for searchmsg contract v2, and "fail-closed" is real. This is the population side that octo-lib #78 (contract v2, which I approved) explicitly gated on: the reader fails OPEN on empty `Visibles`, so the producer must never emit a searchable doc with empty Visibles. Verified end-to-end against the source at head + the pinned octo-lib `ExtractVisibility`.

## Verification (ref=d622bc1a / pinned octo-lib a57dc301cbf3)

1. **All three v2 fields populated** — `payload.go:54` sets `MessageSeq: row.MessageSeq` (sourced via the new `message_seq` SELECT column in `etl_db.go:81,108` → `model.go srcMessageRow.MessageSeq uint64`), `payload.go:80-81` sets `SpaceID`/`Visibles` from `searchmsg.ExtractVisibility`. ✅
2. **fail-closed is genuine** — `payload.go:77-79`: non-encrypted path calls `ExtractVisibility`; on `verr != nil` it returns `outcomeDLQ`, never an empty-Visibles main-topic doc. `etl.go` routes `plan.main → produceBatch` (searchable topic) vs `plan.dlq → produceDLQ` (DLQ topic) — so an unresolvable-visibility row goes to DLQ, NOT the index. The fail-OPEN leak combo is impossible. ✅
3. **`ExtractVisibility` returns err for ALL untrusted cases** (octo-lib `contract/searchmsg/visibility.go`): non-object/array/scalar/`null`/trailing-data → err; `visibles` key present but null / non-array / empty `[]` / all-non-string → err ("resolves to zero valid UIDs"). Only `visibles` **absent** = genuine broadcast → passes through with `nil` (intended & safe). `space_id` is type-tolerant (string-only, otherwise empty) and does not poison a legit `visibles`. ✅
4. **SchemaVersion=2** set unconditionally at `payload.go:48` (`searchmsg.SchemaVersion`). The only main-topic docs with empty Visibles are genuine broadcasts (no `visibles` key) — the documented safe case, not the #1124 leak. Encrypted DM (`payload.go:64-69`) is `raw_excluded` (not searchable) with empty visibles = safe direction. ✅
5. **MessageSeq full-width** — typed `uint64` end-to-end matching the BIGINT column / contract field; `TestExtractMessage_MessageSeqFullWidth` proves no uint32 truncation. ✅
6. **No nil-deref / perf cliff** — enrichment is pure payload parsing (no per-message membership DB query), so no N+1. `ExtractVisibility` guards nil map and bounded element loop. ✅

## 💬 Non-blocking
- **Backfill is out of scope here** (emit-only for new rows). Per #78's gating concern, existing pre-v2 docs still carry empty Visibles. The PR body states producer + indexer backfill share the **same** `FailClosedVisibilityVectors()` set, and `visibility_test.go` runs them — good. Just confirm the backfill PR lands (and repins to a merged octo-lib commit) before any indexer repins to schema v2, so old docs don't leak in the window. Tracking note, not a blocker for this PR.
- **Merge order**: PR pins octo-lib to a pre-release pseudo-version (`a57dc301cbf3`); re-pin to the merged octo-lib commit per the body's stated order.

## ✅ Highlights
- Single-source-of-truth `ExtractVisibility` shared across producer + indexer backfill prevents the cross-repo semantic fork that caused #1124.
- "valid-but-empty visibles → DLQ" is stricter than the old indexer (which treated `[]` as broadcast) — correct hardening.
- Strong test matrix: shared vectors + targeted DLQ/broadcast/full-width-seq/encrypted-skip cases.

CI: Vet/Lint/Build/code-review green; `check-sprint` fail is the sprint-label gate (non-code), `Test` pending. No code-blocking signal.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#415 Review Report
**PR:** Mininglamp-OSS/octo-server#415
**Head SHA:** `d622bc1a09c19f759f9e5c39feb636191892379e`
**Title:** feat(searchetl): enrich producer with SpaceID/Visibles/MessageSeq (contract v2, fail-closed)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `message_seq` added to both SQL paths | ✅ | `etl_db.go:81` (readBatch) + `etl_db.go:108` (readStableBatchTx) — identical column addition |
| `srcMessageRow.MessageSeq` type matches DB column | ✅ | `model.go:12` `uint64` ↔ SQL `BIGINT` (confirmed in `webhook/sql/20210226000001_webhook_legacy01.sql:9`) |
| `MessageSeq` carried into contract | ✅ | `payload.go:57` `MessageSeq: row.MessageSeq` → `searchmsg.Message.MessageSeq uint64` (octo-lib `searchmsg.go`) |
| Fail-closed visibility gate | ✅ | `payload.go:78-82` — `ExtractVisibility` err → `outcomeDLQ`; verified reader fail-OPEN on empty `Visibles` (`messages_search/visibility.go:217-228`) |
| Encrypted messages skip visibility | ✅ | `payload.go:62-66` — early return `outcomeRawExcluded` before `ExtractVisibility` call |
| Shared test vectors used | ✅ | `visibility_test.go:21` calls `searchmsg.FailClosedVisibilityVectors()` (18 vectors from octo-lib) |
| octo-lib re-pin | ✅ | `go.mod` bumps to `v0.0.0-20260618083705-a57dc301cbf3` which contains `ExtractVisibility` + `FailClosedVisibilityVectors` |

## 2. Findings

### F1 — P2 (pre-existing): Reader-side `uint32` truncation of `MessageSeq` undermines end-to-end full-width guarantee

**Diff-scope:** pre-existing — NOT introduced or amplified by this PR.

The producer path (this PR) correctly carries `MessageSeq` as `uint64` from SQL through the contract. However, the reader side truncates to `uint32` at two points:

- `messages_search/visibility.go:131` — `msgRef.MessageSeq uint32`
- `messages_search/visibility.go:487` — `MessageSeq: uint32(d.MessageSeq)` (explicit truncation cast in `projectDocRef`)
- `messages_search/visibility.go:102` — `ChannelOffset()` returns `uint32`
- `message/db_channel_offset.go:68` — `channelOffsetModel.MessageSeq uint32`

The channel_offset gate at `visibility.go:214` (`r.MessageSeq <= offsetSeq`) compares two `uint32` values. For channels where `message_seq` exceeds `math.MaxUint32` (~4.29B), both sides truncate, potentially producing incorrect gate decisions (showing cleared messages or hiding visible ones).

**Why P2 not P1:** Pre-existing system-wide pattern. This PR's scope is the producer enrichment; the reader-side widening is a separate change. The PR's `TestExtractMessage_MessageSeqFullWidth` (visibility_test.go:131, `1<<40`) correctly validates the producer path.

**Recommendation:** File a follow-up to widen `msgRef.MessageSeq`, `channelOffsetModel.MessageSeq`, and `ChannelOffset()` return type to `uint64`. The DB column is already `BIGINT` — only the Go-side types need updating.

### F2 — Note (defense-in-depth): `ExtractVisibility` re-parses payload bytes with stricter decoder

**Diff-scope:** new code in this PR (the call at `payload.go:78`), but the behavior is correct and beneficial.

`json.Unmarshal(row.Payload, &m)` at `payload.go:69` tolerates trailing bytes. `searchmsg.ExtractVisibility(row.Payload)` at `payload.go:78` uses `json.NewDecoder` + trailing-data check. A payload like `{"type":1,"content":"hello"} garbage` would pass the first parse but fail `ExtractVisibility` → DLQ.

This is **correct defense-in-depth**: trailing-data payloads are genuinely corrupt and should not be indexed. The "double parse" cost is negligible vs SQL+Kafka IO.

### F3 — Note: Merge-order dependency on octo-lib#80

PR body documents this: `Mininglamp-OSS/octo-lib#80` must merge first. The go.mod re-pin points to a commit from that PR. If the octo-lib PR changes before merge, the pin must be updated.

## 3. Recommendations

1. **Follow-up (F1):** Widen reader-side `MessageSeq` types to `uint64` to close the end-to-end full-width guarantee.
2. **Merge order:** Confirm octo-lib#80 is merged and the go.mod pin points to the merged commit (not a pre-merge SHA) before landing this PR.

## 4. Additional Observations

- **Single production call-site verified:** `extractMessage` is called only from `planChunk` (`chunk.go:34`). Blast radius is contained.
- **No rogue SQL paths:** Only `readBatch` and `readStableBatchTx` feed `srcMessageRow`. Both updated identically. The `opanalytics` module has a separate, unrelated `srcMessageRow` struct.
- **`isSignalEncrypted` is sound:** Checks both `row.Signal != 0` and `config.SettingFromUint8(row.Setting).Signal`. Encrypted messages correctly bypass visibility extraction.
- **Test quality:** 8 test functions covering shared vectors, edge cases (empty/unparseable visibles, broadcast, encrypted, full-width seq). Tests use hand-fed `srcMessageRow` (bypasses SQL), but the SQL-to-struct mapping is by column name with matching names, so risk is low.

## 5. Data-Flow Trace

```
message.message_seq (BIGINT)
  → SQL SELECT message_seq (etl_db.go:81,108)
    → srcMessageRow.MessageSeq uint64 (model.go:12)
      → searchmsg.Message.MessageSeq uint64 (payload.go:57)
        → Kafka JSON (contract v2)
          → octo-search-indexer esindex.Doc.MessageSeq uint64
            → messages_search/Doc.MessageSeq uint64 (source.go:14)
              → uint32(d.MessageSeq) TRUNCATION (visibility.go:487) ← F1
                → msgRef.MessageSeq uint32 (visibility.go:131)
                  → channel_offset gate comparison uint32 vs uint32 (visibility.go:214)

payload (raw bytes from message.payload column)
  → json.Unmarshal → map (payload.go:69) — lenient, tolerates trailing bytes
  → searchmsg.ExtractVisibility(row.Payload) (payload.go:78) — strict decoder, rejects trailing
    → spaceID string, visibles []string, err
      → err != nil → outcomeDLQ (payload.go:81) — fail-closed, safe
      → err == nil → msg.SpaceID, msg.Visibles enriched (payload.go:83-84)
        → Kafka → indexer → reader visibility gate (visibility.go:217-228)
```

- **C1 — Dual-path parity:** Clear. No add↔remove / subscribe↔unsubscribe symmetry. Encrypted vs non-encrypted paths correctly gated by `isSignalEncrypted` (checked first, early return).
- **C2 — Control-flow ordering / nesting reuse:** Clear. `ExtractVisibility` runs after JSON parse success, before content-type detection — correct ordering. Single production call-site (`planChunk`). No nesting or double-application risk.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. Producer-side enrichment, no new endpoints/tools exposed. Fail-closed design *strengthens* downstream authorization enforcement.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth changes. Visibility enrichment is per-message, not per-container.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The fail-closed visibility design is sound and well-tested. Data flow from SQL → struct → extraction → contract is correctly wired. One P2 follow-up (reader-side `uint32` truncation) is pre-existing and outside this PR's scope.

lml2468:
确认 mochashanyao 第 1 条(读端 MessageSeq uint32 截断)——属实,且关系可见性正确性,虽非本 PR 引入,但强烈建议紧接修 reader 侧,否则 #415 的 producer 全精度白费。

## 跨侧精度坑(我和 Steve 审 #415 时都只核了 producer 侧、漏了对端)
#415 producer 正确写 uint64 全宽 message_seq 进索引。但 reader 侧 modules/messages_search/visibility.go 把它截回 uint32:
- line 487:`MessageSeq: uint32(d.MessageSeq)` —— 把 Doc 的 uint64 MessageSeq **截断成 uint32** 填进 msgRef。
- line 32/102:ChannelOffset 返回 uint32;line 129:msgRef.MessageSeq uint32。
- line 214 的可见性核心 gate:`if offsetSeq > 0 && r.MessageSeq <= offsetSeq`(清空会话前的消息不可见)用的就是这个被截断的值。

后果:message_seq 超过 uint32 上限(~42.9 亿)时高位丢失 → channel-offset gate 比较错位 → 可能把「清空历史前的消息」误判可见(截断后变小)或漏判。这是可见性正确性 bug,不只是性能/精度洁癖。

严重性:**#415 本身(producer)无 blocking**——它做对了 uint64;reader 的 uint32 截断是既有代码(visibility.go,#415 未碰、非本 PR 引入)。但 producer 上 uint64 后,这个对端截断让全精度白费,且 channel_offset.message_seq 列本身是 BIGINT。强烈建议**紧接一个 reader 侧 PR** 把 visibility.go 的 offsetSeq/msgRef.MessageSeq/ChannelOffset 返回值统一改 uint64(对齐 producer + BIGINT 列),否则高序号频道的「清空会话」可见性 gate 会错。

## 我维持 APPROVE 的理由
#415 是 producer-only,职责内做对了(uint64 全宽、fail-closed)。reader uint32 是另一处既有代码的问题,不该 block 这个 producer PR。但它是真问题,不是 nit —— 建议作者/团队立刻开 reader 侧 follow-up,别让它停在「producer 修了一半」。

诚实说:我和 Steve 审 #415 都核了「producer uint64 无截断」(还专门测了 1<<40),但都没去核 reader 端是否同样 uint64 —— 只看了本侧、漏了对端的类型一致性。mochashanyao 抓到了。跨侧字段(producer 写 / reader 读)的精度必须两端都核,这是教训。

mochashanyao 🟡#2(ExtractVisibility 双重解析增防御深度、性能可接受)同意,非问题。

yujiawei (commented):
# Code Review — PR #415 (octo-server)

**Scope:** independent review of `feat(searchetl): enrich producer with SpaceID/Visibles/MessageSeq (contract v2, fail-closed)` at head `d622bc1a`. Reviewed the full diff (6 files, +188/−7) plus the load-bearing dependency `searchmsg.ExtractVisibility` (from the shared `octo-lib` change) against the actual code, and re-ran the producer build/vet/test suite.

## Summary

**Verdict: Approve.** The security core is correct and the highest-value leak vector is properly closed. No P0/P1 blockers. All findings below are P2 (test-coverage / maintainability hardening) and are non-blocking.

This is a security-sensitive change: the search reader treats an **empty `visibles`** array as **fail-OPEN** (no access gate), so a non-encrypted group system message that loses its visibility whitelist would become searchable by ordinary members. The producer correctly routes any message whose visibility cannot be trustworthily parsed to the DLQ rather than emitting empty `visibles` to the main topic.

## 1. Verification results

- ✅ **Fail-closed ordering is correct (the critical property).** `extractMessage` calls `ExtractVisibility` and returns `outcomeDLQ` on error (`payload.go:77-80`) **before** the `!isText` / non-string-content `raw_excluded` early returns (`payload.go:87-101`). A non-text targeted system message (e.g. `type=99` with empty/unparseable `visibles`) therefore cannot reach the main topic with empty `visibles`. Pinned by `TestExtractMessage_EmptyVisiblesGroupMsgDLQ` and `TestExtractMessage_UnparseableVisiblesGroupMsgDLQ`.
- ✅ **Encrypted (Signal) path is safe.** Ciphertext returns `raw_excluded` with `Content=nil` before visibility parsing (`payload.go:61-67`); empty `spaceId`/`visibles` is harmless because the body is non-searchable. Verified by `TestExtractMessage_EncryptedSkipsVisibility`.
- ✅ **`space_id` type-tolerance does not collateral-drop a valid `visibles`.** Non-string `space_id` degrades to empty without erroring the whole parse (`visibility.go:73-79`), isolating the historical V3b fail-OPEN root cause. Pinned by the `tolerant/space_id_number_does_not_drop_visibles` vector.
- ✅ **`valid-but-empty` visibles is rejected.** Key-present-but-resolves-to-zero-UIDs (`[]`, `null`, all-non-string) fails closed (`visibility.go:87-112`) — stricter than the legacy indexer behavior.
- ✅ **Genuine broadcast passes through.** Absent `visibles` key → `visibles=nil`, `outcomeOK` (`visibility.go:82-85`) — expected and safe; avoids flooding the DLQ with normal chat.
- ✅ **`message_seq` data flow is correct.** Added to both `readBatch` and `readStableBatchTx` SELECTs identically (`etl_db.go:81,108`); `dbr` maps the `message_seq` column to `srcMessageRow.MessageSeq` via its `camelCaseToSnakeCase` convention (no db tag needed, consistent with existing `FromUID`/`CreatedUnix`); column is `BIGINT NOT NULL DEFAULT 0`, so the `uint64` scan is safe and full-precision. `uint64` (vs the `uint32` used elsewhere in the codebase) is the **more** correct choice here — it matches the BIGINT column and the v2 contract, avoiding truncation on high-seq channels.
- ✅ **DLQ rows do not stall the cursor.** `planChunk` sets `maxID` from the stable prefix regardless of outcome (`chunk.go:42`); DLQ'd rows still advance the watermark.
- ✅ **`go build` / `go vet` / `go test ./modules/searchetl/...` all green.**

## 2. Findings (all P2 — non-blocking)

### P2-1 — `uint64` ← BIGINT DB-scan boundary is untested
`TestExtractMessage_MessageSeqFullWidth` (`visibility_test.go:131-143`) proves the **Go struct field** carries full width, but it constructs `srcMessageRow` in memory and never exercises the SELECT→`Load` scan path. The "full precision, no truncation from a real BIGINT column" claim has no coverage at the actual scan boundary. Not a bug (the column is non-negative monotonic), but a `sqlmock`/integration test scanning a `>MaxUint32` value through `readBatch` would pin the real risk surface.

### P2-2 — Pass-through vectors assert only `!= outcomeDLQ`, not the exact outcome
For non-error vectors, `TestExtractMessage_SharedFailClosedVectors` (`visibility_test.go:38-47`) collapses `outcomeOK` and `outcomeRawExcluded` into one bucket. A regression turning a normal indexable `type=1` text message into `raw_excluded` (content silently dropped from search) would pass green. Not a #1124 leak axis; the canonical OK case is covered separately by `TestExtractMessage_NormalGroupChatBroadcast`. Consider a `WantOutcome` field on the shared vectors to lock content-type semantics.

### P2-3 — Ordering guard not pinned for the highest-value leak *shape*
The "visibility runs before the `raw_excluded` return" property is pinned only via a `type=99` (non-text) message. There is no fail-closed test for a `type=1` TEXT message (the only path reaching `outcomeOK`) with empty/bad `visibles`. The code is structurally correct today, but a future refactor moving visibility extraction below the content branch would still pass all current tests for the most dangerous shape (an otherwise-indexable text message that loses its whitelist). One vector `{"type":1,"content":"secret","visibles":[]}` asserting DLQ would lock this.

### P2-4 — Empty-visibles→DLQ routing only covered transitively
`chunk_test.go` drives DLQ routing only via a bad-JSON row; no chunk-level test wires an empty-visibles row end-to-end through `planChunk` to assert it lands in `dlq` and is absent from `main`. The behavior is correct by composition (extractMessage→`outcomeDLQ`, then outcome-only routing), but the most leak-relevant input never exercises the routing it relies on.

## 3. Recommendations
- Land the P2 test additions (P2-1..P2-4) as a small follow-up to make the regression guards structural rather than incidental — particularly P2-3, which guards the exact #1124 leak shape.
- **Merge order (already flagged by the author):** `go.mod` pins `octo-lib` to `a57dc301cbf3`, which is the HEAD of the still-open `octo-lib` shared-`ExtractVisibility` PR (ahead-by-1 of `octo-lib` main). This PR must not merge until that dependency lands and the pin resolves to the merged commit; otherwise the build pins to an unmerged branch commit.
- Before flipping live ingestion on, confirm the consumer side re-pins to the v2 contract — the consumer is already gated to refuse live ingestion while `SchemaVersion < 2`, so this is fail-closed by construction, but the rollout sequencing is worth a human eye.

## 4. Additional notes
- **Double-parse of the payload** (`json.Unmarshal` at `payload.go:70` and the decoder inside `ExtractVisibility` at `payload.go:77`): this is **not** a strictness asymmetry — Go's `json.Unmarshal` rejects trailing non-whitespace data, so both parsers fail closed on the same corrupt inputs. There is a minor CPU/GC cost from parsing each payload twice; `ExtractVisibility` intentionally re-reads the raw bytes because it needs strict trailing-data / top-level-shape detection that the `map[string]interface{}` parse discards, so sharing the map would weaken the fail-closed guarantee. Acceptable as-is for correctness; a future optimization could expose a map-based entry point in `octo-lib` if ETL CPU becomes a concern. (Non-blocking.)
- The two *historical-NULL* / *missing-column* failure modes one might worry about for `message_seq` do **not** apply: the column is `bigint NOT NULL DEFAULT 0` in the `message`/`message{1..N}` shard tables since the earliest migration (`modules/webhook/sql/20210226000001_*.sql`), so no historical NULL rows exist and no schema-migration race is introduced — the column already existed; this PR merely adds it to the SELECT.
- `MessageSeq uint64` diverges from the `uint32` used across the `modules/message` API layer — but, as above, the `uint64` here is the correct one; the divergence is a pre-existing latent concern in unchanged code, out of scope for this PR.

## 5. Review method / coverage
This review combined an independent multi-lens pass (security / correctness / contract / test quality), each finding adversarially re-verified against the actual code, with a cross-model advisory pass. Two advisory P1 candidates (a `NULL`→`uint64` scan stall, and an "unknown column" deploy-time crash) were raised but **refuted on independent schema inspection** — `message_seq` has been `NOT NULL DEFAULT 0` since the table's first migration, so neither failure mode is reachable. A line-level adversarial advisory leg did not return usable output this run; its absence is noted and **not** treated as approval. The fail-closed visibility property — the load-bearing security guarantee — was independently confirmed correct.

PR #400, opened 2026-06-17
fix: invalidate offline-push name cache on group/thread rename

an9xyz:
## Problem

Offline push notification titles read the group name (and thread title) through a Redis cache (`Cache.NameCacheExpire`, default **7 days**), but the cache was **never invalidated on rename**. Mobile push banners kept showing the old auto-generated name (e.g. member names joined with `、`) for up to 7 days, while web/app clients already showed the renamed group.

Closes #397.

## Root cause

- `modules/webhook/common.go` caches the push title under `groupName:<groupNo>` / `threadTitle:<channelID>` with a TTL.
- The only invalidation was TTL expiry — group rename (`group.UpdateGroupInfo`) and thread rename (`thread.UpdateName`) updated the DB and notified clients but never dropped the push cache.

## What changed

- **`pkg/pushcache`** — new neutral package owning the push-title cache key scheme plus thin best-effort invalidation helpers. `modules/webhook` imports `modules/group`, so `group`/`thread` cannot import `webhook` to bust the cache (import cycle); this lower-level package is the shared home, keeping the key format in one place.
- **`modules/webhook`** — compose the thread push title from an **independently cached** thread name + group name at read time, instead of caching the frozen composed string. A group rename then only needs to drop the single `groupName:` key for both the group push title **and every thread push title under it** to refresh — no per-thread enumeration, no Redis `KEYS` scan.
- **Invalidation** — bust `groupName:` on group rename and `threadName:` on thread rename, **after the DB write commits** (busting before commit could let a concurrent push re-cache the stale value).
- **Resilience** — a failed Redis `Expire` on the group-name cache is now warn-and-continue (matching the thread-name cache) instead of failing the push.

### Notification contract

The push wire output (`Title`/`Content`/`Badge`/`ChannelID`) is **unchanged** for the same DB state — the title is still composed by the same `BuildThreadTitle` over the same DB-sourced names. Only the caching layout behind those values changed.

### Scope notes

- Disband intentionally does **not** invalidate: a disbanded group gets no further pushes and its name doesn't change, so the entry just TTLs out (no observable benefit → omitted to avoid dead code).
- Old `threadTitle:` keys orphan out via TTL (≤7 days), never read again.

## Testing

- `go build ./...`, `go vet`, `gofmt`, `make i18n-lint` — all green.
- Unit tests (key scheme, `BuildThreadTitle`, `parseThreadChannelID`) pass locally.
[...]

PR #416, opened 2026-06-18
feat(searchetl): wire slow-cursor scheduler + Kafka.On lifecycle (safety checkpoint)

### LINKED ISSUE #417: searchetl: slow-cursor scheduler + Kafka.On lifecycle (safety checkpoint)
Wire the `searchetl` incremental pipeline onto a production scheduler lifecycle so it runs end-to-end for the first time — **slow cursor only** (safety checkpoint).

`RunIncremental` currently has no production caller. This change hangs a minute-level tick scheduler on the octo-lib module Start/Stop lifecycle, with fail-closed boot guards:

- slow-cursor `lag>0` else fatal refuse-to-start (`lag=0` = silent missed reads)
- MySQL-only: non-MySQL dialect skips the searchetl scheduler (not fatal; IM core keeps running)
- `Kafka.On=false` / test mode → skip (lazy, zero overhead)
- debug-level strict-ascending-by-id tripwire on each read batch

Out of scope: fast cursor, `minTickSeconds` clamp change, `lag` narrowing, deployment / production `Kafka.On` enablement.

yujiawei:
## What

First production wiring of the `searchetl` incremental pipeline. `RunIncremental` had **no production caller** until now (only `opanalytics` had its own scheduler). This is the safety checkpoint that lets the never-run-in-prod searchetl code run end-to-end for the first time — **slow cursor only**.

DB → searchetl producer → Kafka `octo.message.v1` → es-indexer consumer → OpenSearch.

## Changes (all in `modules/searchetl/`)

- **`scheduler.go`** — Start/Stop lifecycle + minute-level `time.Ticker` loop driving `ETL.RunIncremental`. Boot-time guards (fail-closed via `Module.Start` returning error → aborts boot):
  - 🔴 **slow-cursor `lag>0`** else fatal refuse-to-start (`lag=0` = silent missed reads; STOP#2 enforced as code)
  - 🔴 **MySQL-only**: non-MySQL dialect **skips** the searchetl scheduler (NOT fatal — IM core keeps running), logs "MySQL-only this phase, PG is a follow-up"
  - `Kafka.On=false` / test mode → skip (lazy, zero overhead)
  - Bounded `Stop` join so a tick stuck in a non-ctx-aware DB call can't hang process shutdown
- **`1module.go`** — register `Start`/`Stop` hooks for the searchetl module
- **`stability.go` / `etl.go`** — debug-level strict-ascending-by-id assertion on each read batch (tripwire: an index/DB change that breaks `ORDER BY id ASC` would make `stablePrefix` mis-truncate and silently drop rows)

## Out of scope (strict)

No fast cursor (ticket 4b), no `minTickSeconds` clamp change (stays 5), no `lag` narrowing (default 600), **no deployment / no production `Kafka.On` enablement**. Production enablement is a separate deploy decision.

## Verification

- Unit tests: scheduler Start/Stop/tick lifecycle, lag-gate fatal, MySQL-only skip, ordering tripwire — `go test -race ./modules/searchetl/` ✅; full `go build ./...` ✅; `go vet` ✅
- **Live e2e harness** (octo-search-indexer `run.sh all`, gate now OPEN at schema_version=2): all 10 invariants PASS via the live Kafka→consumer→OpenSearch path ✅
- **Recon gate** (`run-backfill.sh all`, MySQL↔ES sample): `sampled=5 mismatch=0 missing=0` ✅

Closes #417

### DIFF
--- a/modules/searchetl/1module.go
+++ b/modules/searchetl/1module.go
@@ -10,19 +10,30 @@ import (
 //go:embed sql
 var sqlFS embed.FS
 
-// 模块注册（YUJ-4530 阶段 1 骨架）。
+// 模块注册（YUJ-4530 阶段 1 骨架 → YUJ-5012 票4a 挂生命周期）。
 //
-// 阶段 1 只注册迁移（建独立游标表 octo_etl_es_cursor）。**不**启动 scheduler、**不**接 Kafka——
-// producer 的事务拆分 / 单副本互斥 / Kafka 投递在阶段 2/3 落地，届时再在此挂 Start/Stop 钩子。
-// 这样阶段 1 PR 可独立编译/迁移 dry-run，且不在生产引入任何运行期行为变更。
+// 阶段 1：仅注册迁移（建独立游标表 octo_etl_es_cursor），不启动 scheduler、不接 Kafka。
+// 票4a（安全 checkpoint）：挂 scheduler 生命周期（Start/Stop），让这套「从未在生产真跑过」
+// 的代码第一次真跑——**仅慢游标一个分钟级 tick**，不挂快游标（票4b）、不改 minTickSeconds clamp。
+// Kafka.On 惰性接线保持不变（off 时 scheduler 按护栏跳过、RunIncremental 短路，零开销）。
+// 生产启用（真开 Kafka.On 的部署动作）仍由 deploy 流程单独拍板，不在本票。
 func init() {
 	register.AddModule(func(ctx interface{}) register.Module {
 		appCtx := ctx.(*config.Context)
+		etl := NewETL(appCtx)
+		sched := newScheduler(appCtx, etl)
 		return register.Module{
 			Name:   "searchetl",
 			SQLDir: register.NewSQLFS(sqlFS),
-			// Service 暴露 ETL，便于后续 ops 端点/测试触达（阶段 1 仅 dry-run 能力）。
-			Service: NewETL(appCtx),
+			// Service 暴露 ETL，便于后续 ops 端点/测试触达。
+			Service: etl,
+			// Start/Stop 走 octo-lib module 生命周期：Start 在 server 启动时调用，返回 error
+			// 即 fatal 拒启动（慢游标 lag>0 + MySQL-only 护栏据此把人盯变代码挡）。
+			Start: sched.Start,
+			Stop: func() error {
+				sched.Stop()
+				return nil
+			},
 		}
 	})
 }
--- a/modules/searchetl/etl.go
+++ b/modules/searchetl/etl.go
@@ -174,6 +174,17 @@ func runChunk(ctx context.Context, store chunkStore, sink chunkSink, table strin
 		return chunkPlan{}, 0, nil
 	}
 
+	// 期序 debug 断言（票4a / ReviewBot §4）：stablePrefix 的无空洞截断依赖「批严格按 id 升序」
+	// （readStableBatchTx 的 ORDER BY id ASC）。若索引/DB 版本变更破坏返回序，会在错误位置截断
+	// → 静默漏读。这里只检测 + 大声告警（不改正确性路径），把隐式前提变成可观测 tripwire。
+	if bad := firstNonAscendingByID(rows); bad >= 0 {
+		lg.Warn("searchetl: stable batch not strictly ascending by id — index/DB version may have broken return order (silent missed-read risk)",
+			zap.String("table", table),
+			zap.Int("at_index", bad),
+			zap.Int64("prev_id", rows[bad-1].ID),
+			zap.Int64("curr_id", rows[bad].ID))
+	}
+
 	plan := planChunk(rows, cutoff)
 	if !plan.advanced {
 		// 队首即未稳定：本轮不推进，等其落库满 lag。返回 0 让调用方停止本分片。
--- a/modules/searchetl/scheduler.go
+++ b/modules/searchetl/scheduler.go
@@ -1,0 +1,190 @@
+package searchetl
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+	"github.com/gocraft/dbr/v2"
+	"github.com/gocraft/dbr/v2/dialect"
+	"go.uber.org/zap"
+)
+
+// scheduler 驱动 searchetl 慢游标增量抽取的定时 tick 循环（YUJ-5012 票4a，安全 checkpoint）。
+//
+// 这是让 searchetl 这套「从未在生产真跑过」的代码第一次真跑的生命周期挂点：把 opanalytics
+// 的每日 cron 范式改成分钟级 time.Ticker（tickInterval()），每个 tick 调一次 ETL.RunIncremental。
+//
+// 🔴 本期范围（票4a，严格）：**仅慢游标一个 tick**（lag 维持 600/现状保守值）。不挂快游标
+// （票4b），不改 minTickSeconds clamp。跨副本互斥/失锁 abort 由 RunIncremental 内的 Redis
+// run-lock（C3）负责，scheduler 只管「按节奏触发 + 生命周期」。
+//
+// 启动护栏（plan v2 §4/§5 + ReviewBot 建议4 + codex P1，把人盯变代码挡）：
+//   - Kafka.On=false：不启动 tick 循环（惰性零开销，与 RunIncremental 惰性短路一致）。
+//   - 🔴 慢游标 lag>0 断言：生产环境 lag<=0 → fatal 拒启动（lag=0 = 静默漏读，STOP#2）。
+//   - 🔴 MySQL-only 护栏：非 MySQL 方言（PG 等）→ 跳过 searchetl scheduler（不 fatal，
+//     IM 核心照跑），并明确日志「本期 MySQL-only，PG 见 follow-up」。
+//   - 测试模式（cfg.Test）：不自动起循环（lag=0 仅在此放行；生命周期单测走 startLoop 直驱）。
+type scheduler struct {
+	log.Log
+	ctx      *config.Context
+	etl      *ETL
+	interval time.Duration
+	// tickFn 是每个 tick 执行的动作（默认 etl.RunIncremental）。抽成字段便于生命周期单测
+	// 注入计数器，无需真实 Kafka/DB 即可验证 Start/Stop/tick 触发（验收门 i）。
+	tickFn func(context.Context) error
+
+	mu      sync.Mutex
+	started bool
+	cancel  context.CancelFunc
+	done    chan struct{}
+}
+
+// newScheduler 创建慢游标 scheduler。
+func newScheduler(ctx *config.Context, etl *ETL) *scheduler {
+	return &scheduler{
+		Log:      log.NewTLog("SearchETLScheduler"),
+		ctx:      ctx,
+		etl:      etl,
+		interval: tickInterval(),
+		tickFn:   etl.RunIncremental,
+	}
+}
+
+// startDecision 编码 Start 的护栏裁决（便于纯函数单测，与 DB/server 解耦）。
+type startDecision int
+
+const (
+	startRun   startDecision = iota // 通过护栏，启动 tick 循环
+	startSkip                       // 不启动，但放行 boot（返回 nil；IM 核心照跑）
+	startFatal                      // 拒启动（返回 error，阻断 boot）
+)
+
+// evaluateStart 是启动护栏的纯逻辑（验收门 iii/iv + 惰性/测试跳过路径，均可无 DB 单测）。
+//
+// 优先级：测试模式 > Kafka 惰性 > MySQL-only > lag>0。MySQL-only 先于 lag 判定——非 MySQL
+// 时根本不该起 searchetl，连 lag 都不评估，直接 skip（IM 核心照跑）。
+func evaluateStart(testMode, kafkaOn, mysqlDialect bool, lag int64) (startDecision, string) {
+	if testMode {
+		return startSkip, "test mode: scheduler not auto-started (lag=0 allowed only here)"
+	}
+	if !kafkaOn {
+		return startSkip, "Kafka.On=false: scheduler not started (lazy no-op, zero overhead)"
+	}
+	if !mysqlDialect {
+		return startSkip, "MySQL-only this phase: non-MySQL dialect detected, searchetl scheduler skipped (PG is a follow-up); IM core unaffected"
+	}
+	if lag <= 0 {
+		return startFatal, fmt.Sprintf("slow-cursor lag must be > 0 in production (got %d); lag=0 risks silent missed reads (STOP#2 enforced as code)", lag)
+	}
+	return startRun, "ok"
+}
+
+// isMySQLDialect 判定 dbr 会话方言是否为 MySQL。octo-lib 当前恒以 mysql 驱动 dbr.Open，故
+// 此刻恒真；一旦未来引入 PG 方言（改驱动）即转假，触发 MySQL-only 护栏自动拒启 scheduler。
+func isMySQLDialect(d dbr.Dialect) bool {
+	return d == dialect.MySQL
+}
+
+// Start 启动慢游标 scheduler（幂等）。返回 error 即 fatal 拒启动（由 octo-lib module.Start
+// 传播阻断 boot）；返回 nil 表示已启动或按护栏跳过（boot 继续）。
+func (s *scheduler) Start() error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.started {
+		return nil
+	}
+
+	cfg := s.ctx.GetConfig()
+	// 仅当「非测试 ∧ Kafka.On」即真要起循环时才探测方言（避免在测试/惰性路径触达 DB 层）。
+	mysqlDialect := true
+	if !cfg.Test && cfg.Kafka.On {
+		mysqlDialect = isMySQLDialect(s.ctx.DB().Dialect)
+	}
+
+	switch decision, reason := evaluateStart(cfg.Test, cfg.Kafka.On, mysqlDialect, s.etl.lag); decision {
+	case startFatal:
+		// 🔴 fatal：返回 error 让 module.Start 阻断整个 boot（fail-closed）。
+		return fmt.Errorf("searchetl scheduler refuses to start: %s", reason)
+	case startSkip:
+		s.Info("searchetl scheduler not started", zap.String("reason", reason))
+		return nil
+	}
+
+	s.startLoop()
+	s.Info("searchetl slow-cursor scheduler started",
+		zap.Duration("tick", s.interval), zap.Int64("lag_seconds", s.etl.lag))
+	return nil
+}
+
+// startLoop 实际拉起后台 tick goroutine（护栏通过后调用）。抽出来供生命周期单测直驱，
+// 绕过仅生产相关的护栏（验收门 i：Start/Stop/tick 触发）。调用方须持 s.mu。
+func (s *scheduler) startLoop() {
+	ctx, cancel := context.WithCancel(context.Background())
+	s.cancel = cancel
+	s.done = make(chan struct{})
+	s.started = true
+	go s.loop(ctx)
+}
+
+// loop 是慢游标 tick 主循环：每 interval 触发一次 tick，直到 ctx 取消（Stop）。
+func (s *scheduler) loop(ctx context.Context) {
+	defer close(s.done)
+	t := time.NewTicker(s.interval)
+	defer t.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			s.Info("searchetl scheduler loop stopped")
+			return
+		case <-t.C:
+			s.tick(ctx)
+		}
+	}
+}
+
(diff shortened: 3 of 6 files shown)


lml2468 (approved):
✅ APPROVE — 模范级的「安全 checkpoint」PR：首次在生产接 searchetl 慢游标调度器,处处 fail-closed、生命周期严谨。逐项亲核 + 9 测试全过,无 blocking。

## 启动护栏 fail-closed(把人工注意事项变成代码闸)
evaluateStart 抽成纯函数,优先级 test > Kafka 惰性 > MySQL-only > lag>0:
- Kafka.On=false → skip(零开销,与 RunIncremental 惰性短路一致)。
- 非 MySQL 方言 → skip(不 fatal,IM 核心照跑;明确日志 PG 见 follow-up)。优先级正确:MySQL-only 先于 lag 判定,非 MySQL 时根本不评估 lag(避免 PG 上 lag=0 误 fatal)。
- **lag<=0 → startFatal 拒启动**(module.Start 返回 error 阻断 boot):把「lag=0=静默漏读 STOP#2」从人工 checklist 变成 fail-closed 代码闸。
- Start() 仅在 !Test ∧ Kafka.On 时才探测方言(惰性/测试路径不触达 DB 层)。

## goroutine 生命周期严谨(此前 #400/#414 反复出的资源释放点,这次处理最完整)
- Start 幂等(started 标志);Stop 幂等(未启动安全、重复 Stop 安全)。
- **Stop 的 join 有超时上限 stopJoinTimeout=30s**:注释明确 tick 可能卡在非 ctx 感知的 dbr DB 调用(Begin/Load/Exec 不吃 ctx),无界 join 会拖死进程关停 → 超时放弃 join(仅泄漏一个随进程退出的 goroutine)+ 告警。这是对「DB 调用不感知 ctx」现实的正确取舍,比死等 join 好。
- loop 用 ticker + ctx.Done() select,Stop 取消 ctx 干净退出;tick 错误只记日志不中断循环(下个 tick 重试)。

## 范围克制 + 防御性
- 严格 slow-cursor only:只挂慢游标一个 tick,不挂快游标(票4b)、不改 minTickSeconds clamp。
- 跨副本互斥/失锁 abort 靠既有 RunIncremental 内 Redis run-lock(C3),scheduler 只管节奏+生命周期,本 PR 不引入新并发面。
- tripwire(runChunk):firstNonAscendingByID 检测「稳定批非严格 id 升序」并告警——stablePrefix 无空洞截断依赖 ORDER BY id ASC,DB/索引版本变更破坏返回序会静默漏读;只检测+告警不改正确性路径,把隐式前提变可观测。firstNonAscendingByID 实现正确(含 == 重复 ID 也判违反)。

## 验证
go vet 干净;9 个测试全 PASS——Scheduler Start/Stop/tick/Start幂等/Stop幂等,evaluateStart 的 lag gate / MySQL-only gate / 惰性+测试跳过各分支(纯函数无 DB),firstNonAscendingByID 升序/违反/短输入。生命周期单测用注入 tickFn + 极短 interval,无需真实 Kafka/DB。

无 blocking、无 should-fix。批准。这是让「从未在生产跑过」的代码首次上线的标杆做法:护栏 fail-closed、范围最小、生命周期可控、隐式前提加 tripwire。

—— Octo-PR (lml2468)

Jerry-Xin (approved):
**APPROVE** — clean lifecycle wiring with a fail-closed start guard and a bounded-join shutdown. The hard correctness (checkpoint atomicity, fencing, error semantics) all lives in the already-merged #414/#415 paths; this PR only adds the Start/Stop lifecycle around it, and does so carefully.

### Verification (ref=497a7afc)
1. **Lifecycle / goroutine — clean (no #414-style leak).** `1module.go` wires `Start: sched.Start` / `Stop: sched.Stop` into the module lifecycle. `Stop()` (`scheduler.go:Stop`) is idempotent, cancels the loop ctx, and `select`s on `done` so the goroutine is actually joined — unlike the `ChannelReconciler.Stop()` gap in #414. The bounded `stopJoinTimeout = 30s` is the right call: a tick can be parked in a non-ctx-aware dbr DB call, so it abandons the join + warns rather than hanging shutdown. ✅ confirmed.
2. **Checkpoint safety — crash-safe / at-least-once.** Cursor advance is unchanged from #415: `runChunk` produces the full batch to Kafka *first*, then `advanceCursor(table, cursor, maxID)` CAS-advances **after** confirmed produce (`etl.go` runChunk). Checkpoint-after-success = no row skipped on crash; duplicates absorbed by ES `_id` idempotent upsert. No data loss. ✅ confirmed.
3. **Concurrency — no race.** `scheduler` fields (`started`/`cancel`/`done`) all guarded by `s.mu`. Tick re-entrancy serialized two ways: in-process `running` CAS + cross-replica Redis run-lock with CAS fence (`RunIncremental`). Scheduler only triggers; it shares no mutable cursor state. ✅ confirmed.
4. **Slow-cursor pacing — sane.** `tickInterval()` clamped to `[minTickSeconds=5s, maxTickSeconds=3600s]`, default 60s (`config.go`). No tight loop; batch is keyset-paged and bounded by `RunIncremental`. ✅ confirmed.
5. **Error handling — fail-open per-tick, fail-closed at boot (correct).** `tick()` logs the error and keeps the loop alive (next tick retries; cursor only advances on success, so a failed batch is re-produced, not skipped). Boot guard `evaluateStart` is fail-closed: prod `lag<=0` → fatal refuse-to-start (STOP#2 as code), non-MySQL → skip (IM core unaffected), `Kafka.On=false`/test → lazy skip. Posture consistent with #415. ✅ confirmed.
6. **No silent panic.** No raw map/slice deref in the goroutine; `tickFn` errors are caught. Loop exits cleanly on ctx.Done. (Minor note below.) ✅ confirmed.

### 💬 Non-blocking
- `loop()`/`tick()` have no `recover()`. A panic deep inside `RunIncremental` would kill the goroutine silently (loop dies, no more ticks, no crash). Today the panic surface is low and a dead scheduler is fail-safe (no data corruption, just stalled ETL), so not blocking — but a `defer recover()+Error log` in `tick` would convert a silent stall into a loud one. Worth a follow-up.
- Bonus `firstNonAscendingByID` tripwire (`stability.go`) is a nice defensive touch — detection-only, O(n), well-tested.

### ✅ Highlights
- Guard logic factored into pure `evaluateStart` + table-driven tests (lag gate, MySQL-only, lazy/test skip) — verifiable without DB/Kafka.
- Honest scoping: slow-cursor tick only, fast cursor explicitly deferred to 4b; Kafka.On lazy path keeps zero prod overhead until deploy flips it.
- Bounded shutdown join is exactly the right pragmatic fix for non-ctx-aware DB calls.

CI green (Build/Lint/Vet/scans pass; Test in progress). Mergeable.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#416 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/416
**Head SHA:** `497a7afc9863a9a0c40c1986e4ddde26b2bbce74`
**Scope:** feat(searchetl): wire slow-cursor scheduler + Kafka.On lifecycle (safety checkpoint)
**Files:** 6 changed (+407 / -6) — `scheduler.go` (new), `scheduler_test.go` (new), `1module.go`, `etl.go`, `stability.go`, `stability_test.go`

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Start/Stop lifecycle wiring | ✅ | `1module.go:22-31` — `sched.Start` / `sched.Stop` correctly registered on `register.Module` |
| Lag>0 fatal guard | ✅ | `scheduler.go:89-91` — `evaluateStart` returns `startFatal` when `lag<=0` in prod+MySQL+Kafka.On |
| MySQL-only guard | ✅ | `scheduler.go:85-87` — non-MySQL → `startSkip` (not fatal), IM core unaffected |
| Kafka.On lazy skip | ✅ | `scheduler.go:81-83` — `!kafkaOn` → `startSkip`, zero overhead |
| Test mode skip | ✅ | `scheduler.go:78-80` — `testMode` → `startSkip`, lag=0 allowed only here |
| Bounded Stop join | ✅ | `scheduler.go:170-189` — 30s timeout, abandons stuck goroutine with warning |
| Ordering tripwire | ✅ | `etl.go:177-185` — `firstNonAscendingByID` check between DB read and `planChunk`, Warn-level log |
| `firstNonAscendingByID` correctness | ✅ | `stability.go:33-39` — O(n) scan, returns first violation index or -1 |
| Idempotent Start/Stop | ✅ | `scheduler.go:100-103` (started guard), `scheduler.go:163-166` (started=false early return) |
| Test coverage | ✅ | 8 test functions covering lifecycle, guards, idempotency, ordering detection |

## 2. Findings

### No P0/P1 issues found.

### P2 — Test name over-promises (nit)

**File:** `scheduler_test.go:113-127`
**Diff-scope:** new (this PR)

`TestStart_LagGateFatalAndSkip` name promises both "LagGateFatal" AND "Skip" verification through the real `Start()` entry point, but the body only tests the Kafka.On=false skip path. The lag=0 → fatal path through `Start()` is not tested (only through the pure `evaluateStart` in `TestEvaluateStart_LagGate`).

**Impact:** Low — the pure-function test covers the logic, but an integration-level test through `Start()` with `lag=0, Kafka.On=true, Test=false` would catch any future wiring bug between `Start()` and `evaluateStart`.

**Suggestion:** Add a second sub-test that constructs a scheduler with `Kafka.On=true, Test=false, lag=0`, calls `Start()`, and asserts the returned error is non-nil.

### P2 — No immediate first tick (design note, not a bug)

**File:** `scheduler.go:139-150`
**Diff-scope:** new (this PR)

The `loop` creates a `time.NewTicker` and waits for the first tick before executing `tickFn`. With default 60s interval, the first `RunIncremental` fires 60s after boot. For a safety checkpoint with `lag=600s`, this is acceptable. But if faster first-run is desired in the future, consider executing the first tick immediately before entering the ticker loop.

**Impact:** None for current scope. Noting for future reference.

### P2 — Stop wrapper silently discards error return

**File:** `1module.go:28-30`
**Diff-scope:** new (this PR)

```go
Stop: func() error {
    sched.Stop()
    return nil
},
```

`sched.Stop()` returns void, so the wrapper always returns nil. This is correct today (Stop has no error path), but if `Stop()` is ever refactored to return an error, this wrapper would silently swallow it. Minor future-proofing concern.

**Impact:** None today.

## 3. Suggestions

1. **Add lag=0 fatal integration test** (per P2 above) — one more sub-test in `TestStart_LagGateFatalAndSkip` to close the gap between pure-function and entry-point testing.
2. **Consider first-tick-immediate** in a follow-up if the 60s boot-to-first-ETL delay becomes operationally relevant.

## 4. Additional Observations

- **Goroutine leak on Stop timeout is acceptable:** The 30s `stopJoinTimeout` in `Stop()` may leak a goroutine stuck in a non-ctx-aware DB call, but this goroutine dies with process exit. The comment at `scheduler.go:157-160` correctly documents this trade-off.
- **`context.Background()` in `startLoop` is correct:** The scheduler owns its lifecycle independently. The framework calls `module.Stop()` on shutdown, which cancels the scheduler's own context. No need for a parent context.
- **`evaluateStart` priority ordering is correct:** testMode > Kafka lazy > MySQL-only > lag>0. The MySQL-only check correctly precedes the lag check — non-MySQL skips without evaluating lag, avoiding a spurious fatal on a platform that shouldn't run searchetl anyway.
- **Ordering tripwire placement is correct:** `firstNonAscendingByID` sits between `readStableBatchTx` (source: `ORDER BY id ASC`) and `planChunk` (consumer: depends on ascending order). It detects the invariant violation without altering the correctness path — pure observation.

## 5. Data-Flow Trace

|---|---|---|---|
| `s.etl.lag` | `lagSeconds()` → env `OCTO_SEARCHETL_LAG_SECONDS` / default 600 | `NewETL()` → `ETL.lag` → `scheduler.etl.lag` → `evaluateStart(lag)` | ✅ Yes, clamped [0, 86400] |
| `s.interval` | `tickInterval()` → env `OCTO_SEARCHETL_TICK_SECONDS` / default 60 | `newScheduler()` → `scheduler.interval` → `time.NewTicker(s.interval)` | ✅ Yes, clamped [5s, 3600s] |
| `s.ctx.DB().Dialect` | `config.Context.DB()` → `NewMySQL()` → `dialect.MySQL` | `Start()` → `isMySQLDialect(s.ctx.DB().Dialect)` → `evaluateStart(mysqlDialect)` | ✅ Yes, currently always MySQL |
| `cfg.Test`, `cfg.Kafka.On` | `config.Context.GetConfig()` | `Start()` → `evaluateStart(testMode, kafkaOn, ...)` | ✅ Yes |
| `rows` for ordering check | `readStableBatchTx()` → SQL `WHERE id>? ORDER BY id ASC LIMIT ?` | `runChunk()` → `firstNonAscendingByID(rows)` → `planChunk(rows, cutoff)` | ✅ Yes, ORDER BY confirmed in SQL |
| `tickFn` | `newScheduler()` sets `tickFn: etl.RunIncremental` | `loop()` → `tick(ctx)` → `s.tickFn(ctx)` → `RunIncremental(ctx)` | ✅ Yes |
| `Start`/`Stop` hooks | `1module.go` closure captures `sched` | `register.Module.Start`/`Stop` → `module.Start(ctx)` iterates modules, calls `m.Start()` | ✅ Yes, error propagates to abort boot |

## 6. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. This PR adds lifecycle management only (Start/Stop). No symmetric add/remove or create/delete paths are touched. The scheduler creates one goroutine in `startLoop` and cancels it in `Stop` — single path, properly paired.
- **C2 — Control-flow ordering / nested reuse:** Clear. `tick()` → `tickFn()` → `RunIncremental()` is a single call chain. `RunIncremental` has its own `running.CompareAndSwap` re-entry guard, so even if two ticks somehow overlapped (they can't — single goroutine sequential loop), the second would be safely skipped. No nested/double invocation risk.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. No new endpoints, tools, or capabilities are exposed. This PR only wires existing `RunIncremental` (which already had its own Redis run-lock + CAS fence + Kafka.On gate) to the module lifecycle.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. No authorization changes in this PR.

## 7. Cross-Round Blocker Recheck

---

**[Octo-Q] verdict: APPROVE**

This is a well-engineered safety checkpoint PR. All guard rails (lag>0 fatal, MySQL-only skip, Kafka.On lazy, test mode skip) are correctly prioritized and implemented. The lifecycle wiring is clean with proper idempotency and bounded shutdown. The ordering tripwire adds valuable observability for a critical invariant. Test coverage is solid across lifecycle, guards, and edge cases. No P0/P1 issues found — only minor P2 nits (test name gap, no immediate first tick, Stop wrapper).

yujiawei (commented):
# Code Review — PR #416 (octo-server)

Reviewed at head `497a7afc9863a9a0c40c1986e4ddde26b2bbce74`. Scope: `modules/searchetl/` (6 files, +407/-6). Verified locally: `go build ./...` ✅, `go vet ./modules/searchetl/` ✅, `go test -race ./modules/searchetl/` ✅.

## Verdict: APPROVED

This is a deliberately **gated-off safety checkpoint**. In production `Kafka.On=false`, and `evaluateStart` returns `startSkip` in that state, so the new scheduler goroutine is **never spawned** until a separate deploy decision flips `Kafka.On=true`. No correctness, security, or data-loss issue is reachable in the merged-and-deployed state, so nothing here blocks merge.

That said, every item below should be treated as a **pre-enablement checklist**: they become reachable the moment `Kafka.On` is turned on, and a couple of them are genuinely nasty at that point. None are blocking *this* merge; all are worth fixing before flipping the flag.

## Verification of the PR's own claims (all confirmed)

- **Fail-closed boot.** `sched.Start` is wired as `Module.Start` (`1module.go:32`). A fatal `lag<=0` returns an error → `module.Start` returns it (octo-lib `module/module.go:58-67`) → `Server.Start` returns it (`server/server.go:106-108`) → `svc.Run` aborts → `main.go` panics. The lag-gate genuinely aborts the process. ✅
- **Guard ordering** (`evaluateStart`, `scheduler.go`): `test > kafka-off > non-MySQL skip > lag<=0 fatal`. Correct and lazy — the DB dialect probe only runs on the real production branch (`!Test && Kafka.On`). ✅
- **Tripwire premise.** `readStableBatchTx` does use `ORDER BY id ASC` (`etl_db.go`), so `firstNonAscendingByID` guards the real invariant `stablePrefix` relies on. ✅
- **Scope.** No fast cursor, no `minTickSeconds` change, no `Kafka.On` enablement. Diff stays within the stated checkpoint scope. ✅

## Findings (none blocking — ordered by pre-enablement priority)

### 1. (P1, pre-enablement) tick goroutine has no panic boundary
`scheduler.go:130` launches `go s.loop(ctx)`; the chain `loop → tick → tickFn (etl.RunIncremental → newProducer/newLock/runChunk/dbr calls)` has **no `recover()` anywhere** in the module. `tick` (`scheduler.go:151-155`) only handles returned `error`, not panics. A panic anywhere under `RunIncremental` (nil deref, producer init, a slice bug, a driver panic) escapes the detached goroutine and, per Go semantics, **takes down the whole IM server process**, not just searchetl.

Note: I checked the "regression vs opanalytics" angle and it does **not** hold — `opanalytics/scheduler.go` uses bare `cron.New()`, and robfig/cron v3.0.1 `New()` installs an empty chain with no recover wrapper, so opanalytics has the same exposure. So this is not a regression, but for code that "has never run in production," a process-wide blast radius is worth a `defer recover()` in `tick` (log + continue) before enablement.

### 2. (P1, pre-enablement) `defer close(s.done)` closes a struct field, not a captured local
`loop` does `defer close(s.done)` (`scheduler.go:135`), reading the **field**, while `startLoop` reassigns `s.done = make(chan struct{})` on every (re)start (`scheduler.go:128`). Combined with the abandoned-join path in `Stop` (on `stopJoinTimeout` the stuck goroutine is left alive while `started=false`, `scheduler.go:171,182`), a later `Start` overwrites `s.done` and launches a new goroutine; when the old stuck goroutine finally unblocks, its deferred `close(s.done)` closes the **new** channel → `close of closed channel` panic (which, per finding 1, has no recover → crash). It is also an unsynchronized field write racing `startLoop` under `-race`.

Reachability is narrow (requires `Kafka.On=true` + a tick stuck >30s in a non-ctx DB call + an in-process restart), and no current test exercises restart-while-alive. Fix is minimal and worth doing: capture the channel as a local and pass it in — `go s.loop(ctx, done)` with `defer close(done)` — so each goroutine only ever closes its own channel.

### 3. (P2) Ordering tripwire is detect-only, and its comment mislabels its level
`firstNonAscendingByID` only logs (`etl.go:181`) and then proceeds into `planChunk`/`advanceCursor`. If `ORDER BY id ASC` ever breaks, `stablePrefix` still mis-truncates and the cursor still advances past dropped rows — the tripwire **observes** the silent missed-read, it does not prevent it. That is a defensible design choice for a checkpoint, but the comment calls it a "debug 断言 / debug 期校验" (`etl.go:177`, `stability.go`) while the call is `lg.Warn(...)`, which in octo-lib's log package fires unconditionally (warn core pinned to WarnLevel) on every chunk. Either gate it to Debug to match the comment, or keep Warn and reword the comment to "Warn-level production tripwire, detect-only, does not prevent missed reads."

### 4. (P2) The Start() fatal path has no end-to-end test
`TestStart_LagGateFatalAndSkip` is named "FatalAndSkip" but only asserts the `Kafka.On=false` **skip** branch; the actual `Kafka.On=true && lag<=0 → Start() returns error` wiring is only covered at the pure-function layer (`TestEvaluateStart_LagGate`). The single most safety-critical guarantee of this PR is thus not locked end-to-end — a future refactor that swapped the `startFatal`/`startSkip` arms would pass every existing test. This *is* unit-testable without a live DB (`dbr.Open("mysql", "", nil)` does no I/O; an in-repo precedent injects a sqlmock `dbr.Session` with `dialect.MySQL` via reflection). Recommend adding the fatal-case assertion and renaming the test to match what it covers.

### 5. (nit) MySQL-only guard is dormant-by-design
`isMySQLDialect(ctx.DB().Dialect)` is provably always true today (octo-lib `Context.DB()` only ever opens MySQL), so the non-MySQL skip branch cannot fire. The PR comments are honest that this is a future tripwire, not active protection — flagged only so the guard list isn't read as "PG deployments are actively shielded today."

### 6. (nit) First incremental is delayed by one full tick
`time.NewTicker` fires first only after `interval` (default 60s), so the slow cursor sits idle for up to one interval after enablement. Matches the daily-cron precedent and is harmless; mentioned only for operator expectation. A guarded priming tick before the select loop would remove it if desired.

## Coverage notes / blind spots

- **Verified by build/vet/`-race` + reading the full module and the pinned octo-lib lifecycle** (`module/module.go`, `server/server.go`). Confirmed the boot chain and migration ordering (`module.Setup` runs migrations before `Server.Start`, so `DB()` is ready when `Start` reads `.Dialect`).
- **Not exercised at runtime**: live Kafka→consumer→OpenSearch e2e and the recon gate — relied on the PR's reported results; not independently reproduced here.
- **Out of diff, worth checking before enablement**: actual production `OCTO_SEARCHETL_LAG_SECONDS` value (the fatal gate only ensures `>0`, not that it exceeds the real max commit latency), Redis run-lock behavior under contention, and cross-module shutdown ordering (the abandoned-join goroutine interacting with other modules' Stop).
- A second independent long-context advisory pass was attempted but **timed out and is recorded as absent — not as silent approval**. This verdict rests on the line-level pass, an independent cross-family adversarial pass, and my own re-verification of every finding against the code.

PR #423, opened 2026-06-19
feat(incomingwebhook): add GitLab + Feishu platform adapters (#297 Phase 4)

an9xyz:
## Summary

Completes the final phase of #297 — the **GitLab** and **Feishu** platform adapters — on the shared `pushAdapter` pipeline introduced in #330. Two new push shapes; same auth / rate-limit / group / audit path, only body parsing differs (one file per adapter).

`Closes #297` (Phase 4 is the last open item on that issue).

### GitLab (`.../:webhook_id/:token/gitlab`)
- Render by `X-Gitlab-Event`: **Push / Tag Push / Merge Request / Issue / Note / Pipeline** → markdown. Out-of-subset events and noisy actions (MR `update`, pipeline `running`, GitLab **system notes**) answer `200 + skipped`; missing `X-Gitlab-Event` → `400 reason=no_event` (consistent with GitHub).
- **Auth**: besides the URL token (verified by the shared pipeline), the project **Secret token must equal it** and is re-checked constant-time via `X-Gitlab-Token` (new optional `verifyToken` hook). A mismatch is a config error (the caller already holds the URL secret) → uniform `401`, audited as `reason=token` for `deliveries` troubleshooting, with no IP-budget burn.
- Platform-generated body → dedicated **1 MiB** cap (`DM_INCOMINGWEBHOOK_GITLAB_MAX_BYTES`, clamped to 25 MiB).

### Feishu (`.../:webhook_id/:token/feishu`)
- Accept the custom-bot outbound format: `text` → text; `post` → markdown (title + per-line `text`/`a`/`at`; `img` dropped — `image_key` can't be re-hosted); `interactive` card → markdown (`div`/`markdown` elements; buttons/images dropped). Media types (`image`/`share_chat`) rejected `400`.
- Success carries `code=0`/`msg=success` for Feishu SDK compatibility. `post` routes through the **text path (not RichText)** so links stay clickable — high-fidelity card rendering isn't feasible, same graceful-degradation contract as WeCom in #330. README documents that `timestamp`/`sign` are ignored and the URL token is the sole credential.

### Shared / docs
- `create`/`regenerate` `urls` now include `gitlab` + `feishu`; README documents both shapes; migration refreshes the audit `adapter`/`reason` column comments (adds `gitlab`/`feishu`/`token`, carries forward `no_event` from #421).
[...]

PR #421, opened 2026-06-18
fix(incomingwebhook): address PR #330 review follow-ups (#297)

an9xyz:
## Summary

Clears the six non-blocking review follow-ups tracked on #297 Phase 3 (the `modules/incomingwebhook` platform-adapter work shipped in #330). All small, self-contained, and independently mergeable.

## Changes

- **Escape `]`/`[` in GitHub-controlled markdown link text.** New `mdLinkText` helper (`adapter.go`) escapes `\`/`[`/`]` and is applied to PR / issue / release / issue-comment titles in `adapter_github.go`. A crafted title like `pwn](http://evil)` can no longer prematurely close `[text](url)` and expose/garble the real URL. (Commit SHAs and the comment-body snippet are unaffected — SHAs are hex, the snippet renders inside a `>` blockquote, not link text.)
- **Clamp `githubMaxBytes()` to a 25MiB hard ceiling** (GitHub's own webhook payload cap), so a fat-fingered `DM_INCOMINGWEBHOOK_GITHUB_MAX_BYTES` can't blow up the per-request body buffer. Default stays 1MiB.
- **Distinct `no_event` reason** for the missing-`X-GitHub-Event` 400, separating the config-error case from the legitimate 200 `skipped(reason=event)` case (event outside the rendered subset). Deliveries can now tell them apart by `reason` alone.
- **Tests:** pin both byte-cap env vars via `t.Setenv` in `TestGitHubMaxBytes_ExceedsNativeCap` (no longer ambient-env dependent); add `TestGitHubMaxBytes_ClampsHugeEnv` and `TestParseGitHubPush_LinkTextEscaped`.
- **Docs/comments:** refresh the stale 413 comment in `pkg/errcode` to document the per-shape caps (native/wecom 8KiB vs GitHub 1MiB→25MiB); README notes the recommended WeCom `news` article count (~15–20) to stay under the 4000-rune cap; README + errcode reason list updated to `no_event`.
- **Migration:** append `20260618000001_incomingwebhook_no_event_comment.sql` — a comment-only `MODIFY COLUMN` that adds `no_event` to the `reason` column's self-documenting enumeration (same append-a-comment-migration pattern as #330's `20260610000001`).

This leaves #297 with only **Phase 4 (GitLab / Feishu adapters)** outstanding.

## Testing

Infra provisioned locally (Redis native, MySQL 8.0; WuKongIM image unreachable under the env network policy but the package's tests tolerate best-effort IM delivery):

- `go build` + `go vet` clean.
[...]

PR #427, opened 2026-06-22
feat(incomingwebhook): add /multica adapter for multica outbound webhooks

### LINKED ISSUE #426: feat(incomingwebhook): add /multica adapter for multica-style outbound payloads
## Background

Multica (an AI-native task management platform) ships an outbound webhook system that POSTs a fixed JSON envelope to a configured URL when issues change (currently `issue.status_changed`). The envelope shape is:

```json
{
  "event": "issue.status_changed",
  "workspace_id": "<uuid>",
  "actor": { "type": "member|agent", "id": "<id>" },
  "issue": { /* full IssueResponse */ },
  "previous_status": "<old>",
  "delivered_at": "2026-06-22T14:30:45Z"
}
```

Headers: `Content-Type: application/json`, `X-Multica-Event`, `X-Multica-Delivery` (UUID), `X-Multica-Signature-256: sha256=<hex(HMAC-SHA256(body, secret))>` (identical scheme to GitHub).

Multica does **not** support per-subscription body templating or arbitrary headers — the envelope is fixed. So pointing a multica outbound webhook at our existing native push endpoint (`POST /v1/incoming-webhooks/:id/:token`) results in `400 reason=content` on every event (no `content` / `text` field). The GitHub adapter doesn't help either — it requires `X-GitHub-Event` and a GitHub-shaped body.

## Proposal

Add a third platform adapter parallel to `/github` and `/wecom`:

```
POST /v1/incoming-webhooks/:webhook_id/:token/multica
```

Server-side translation of the multica envelope into a native markdown text push, e.g.:

```
**alice** moved [MUL-123 Fix login redirect](https://app.multica.ai/...) from `todo` → `in_progress`
```

Initial event coverage: `issue.status_changed` (multica's only v1 event). Future events (`issue.created`, `issue.assigned`, `comment.created`) will be additive renders behind the same adapter.

## Acceptance criteria

- New file `modules/incomingwebhook/adapter_multica.go` following the `adapter_github.go` pattern.
- Route mounted in `api.go` push group; audit `adapter=multica`.
- `urls` field on create/regenerate responses gains a `multica` entry.
- Body cap: native 8 KiB suffices (multica IssueResponse is small; long descriptions trimmed). No need for github-style 1 MiB cap.
- Unrecognized events return `200` + `skipped:\"event\"` (matches the github adapter contract).
- README updated with the new adapter section.
- Optional but recommended: when `X-Multica-Delivery` header is present and we've seen the same `(webhook_id, delivery_uuid)` within 60s, return `200` idempotently without re-posting to the group — closes the at-least-once gap that multica's 3-attempt retry introduces.
- Tests: parse + render for `issue.status_changed`; unknown event → skip; missing required fields → `reason=json` / `content`.

## Out of scope

- HMAC signature verification (`X-Multica-Signature-256`) — left for a follow-up, same status as `X-Hub-Signature-256` on the github adapter.
- Multi-event renders beyond `issue.status_changed` — additive once multica emits them.

## Companion issue

Companion improvements on the multica side: https://github.com/Mininglamp-OSS/multica/issues (test push endpoint + auto-disable on repeated 4xx).

lml2468:
Closes #426.

## Summary
- Adds `POST /v1/incoming-webhooks/:webhook_id/:token/multica`, a fourth platform adapter parallel to `/github` and `/wecom`.
- Translates Multica's fixed outbound webhook envelope (`{event, workspace_id, actor, issue, previous_status, …}`) into a native markdown push so the message lands in the group as e.g. `**MUL-123** Fix login redirect: \`todo\` → \`in_progress\` (by agent)`.
- v1 only recognizes `issue.status_changed`; other events return `200 + {"skipped":"event"}` (same contract as `/github`'s rendering-subset behavior). Future events (`issue.created`, `comment.created`, …) are additive — just add a `case` to the renderer.
- `publicURLs` returns a `multica` entry alongside `native`/`github`/`wecom` so create/regenerate responses include the new path.
- Body cap stays at the native 8 KiB (Multica envelopes are far smaller than github event payloads — no need for the 1 MiB github cap).
- HMAC verification of `X-Multica-Signature-256` left for a follow-up (same status as `X-Hub-Signature-256` on the github adapter).

## Test plan
- [x] `go test ./modules/incomingwebhook/ -run 'TestParseMulticaPush|TestMulticaAdapter|TestPublicURLs'` — 12 new unit tests pass (parse + render for `issue.status_changed`, unknown event → `skip=event`, missing `event` / malformed JSON → `invalid=json`, missing required `identifier`/`status` → `invalid=content`, title escaping, long-title clipping, adapter registration, urls map).
- [x] Live E2E against the OOTB OCTO Docker stack from `Mininglamp-OSS/octo-deployment`:
  - Built `octo-server:multica-adapter` from this branch.
  - Brought up the stack, registered a user, created a group + incoming webhook.
  - POSTed sample multica envelopes via curl: success returns `200 {"status":0,"message_id":<int>}`; unknown event returns `200 {"skipped":"event"}`; missing identifier returns `400 reason=content`; malformed JSON returns `400 reason=json`.
  - Drove the multica side end-to-end (companion PR Mininglamp-OSS/multica#39, see "E2E harness" commit there): real `*outwebhook.Dispatcher` signed and POSTed the envelope to this adapter, the message landed in the group, and both sides' audit logs reconcile (`adapter=multica, status=1, http_status=200` on octo; `status=delivered, response_status=200` on multica).

## Out of scope
- HMAC signature verification of `X-Multica-Signature-256`.
- Renders for additional multica events beyond `issue.status_changed` (additive once multica emits them).

### DIFF
--- a/modules/incomingwebhook/README.md
+++ b/modules/incomingwebhook/README.md
@@ -204,7 +204,9 @@ POST /v1/incoming-webhooks/:webhook_id/:token/multica
 
 子集之外的事件（`issue.created` / `comment.created` 等未来事件）返回 200 +
 `{"skipped":"event"}`（deliveries 里 `status=3`/`reason=event` 可见，与 github
-适配器一致）；payload 解析失败 → 400 `reason=json`；事件已识别但 issue 关键
+适配器一致）；缺 `event` 字段（配置错误）→ 400 `reason=no_event`（与 github 缺
+`X-GitHub-Event` 同语义，可在 deliveries 里与「不在渲染子集」的 200 skip 分开看）；
+payload 解析失败 → 400 `reason=json`；事件已识别但 issue 关键
 字段（identifier / status）缺失 → 400 `reason=content`。
 
 **body 上限：** 与 native 同为 8 KiB——Multica 信封比 GitHub 事件紧凑（只嵌
--- a/modules/incomingwebhook/README.md
+++ b/modules/incomingwebhook/README.md
@@ -47,6 +47,7 @@
 POST /v1/incoming-webhooks/:webhook_id/:token            # native（本文主体）
 POST /v1/incoming-webhooks/:webhook_id/:token/github     # GitHub 事件适配器
 POST /v1/incoming-webhooks/:webhook_id/:token/wecom      # 企业微信群机器人格式适配器
+POST /v1/incoming-webhooks/:webhook_id/:token/multica    # Multica 出站 webhook 适配器
 Content-Type: application/json
 ```
 
@@ -185,6 +186,42 @@ curl -X POST "$BASE/v1/incoming-webhooks/$WEBHOOK_ID/$TOKEN/wecom" \
   -d '{"msgtype":"markdown","markdown":{"content":"**Build #123** passed"}}'
 ```
 
+### Multica（出站 webhook 格式）
+
+```
+POST /v1/incoming-webhooks/:webhook_id/:token/multica
+```
+
+接受 [Multica](https://github.com/multica-ai/multica) 出站 webhook 的固定 JSON
+信封——在 Multica 工作区 Settings → Webhooks 把 URL 配成上述地址即可，无需中间
+转换层。鉴权沿用 URL 内 token；Multica 出站请求会带 `X-Multica-Signature-256`
+（与 GitHub 的 `X-Hub-Signature-256` 同算法），与 github 适配器对称——目前不校验，
+留作后续可选项。
+
+| `event` | 渲染 |
+|---------|------|
+| `issue.status_changed` | `**MUL-123** Title: ` + 状态变化（如 `` `todo` → `in_progress` ``） + actor 类型尾注 |
+
+子集之外的事件（`issue.created` / `comment.created` 等未来事件）返回 200 +
+`{"skipped":"event"}`（deliveries 里 `status=3`/`reason=event` 可见，与 github
+适配器一致）；payload 解析失败 → 400 `reason=json`；事件已识别但 issue 关键
+字段（identifier / status）缺失 → 400 `reason=content`。
+
+**body 上限：** 与 native 同为 8 KiB——Multica 信封比 GitHub 事件紧凑（只嵌
+单个 issue），不需要 github 那种 1 MiB 上限。
+
+**展示身份：** 与其它适配器一致，固定为 webhook 配置；信封里没有 `username` /
+`avatar_url` 字段。
+
+```bash
+# 把 Multica outbound webhook URL 配成 octo 适配器路径即可
+curl -X POST "$BASE/v1/incoming-webhooks/$WEBHOOK_ID/$TOKEN/multica" \
+  -H 'Content-Type: application/json' \
+  -d '{"event":"issue.status_changed","actor":{"type":"member","id":"u-1"},
+       "issue":{"identifier":"MUL-123","title":"Fix login","status":"in_progress"},
+       "previous_status":"todo"}'
+```
+
 ## 通用字段与安全
 
 - `username` / `avatar_url`：两种形态通用，服务端裁剪到字节上限（名 64B / 头像 255B）。
@@ -209,7 +246,7 @@ curl -X POST "$BASE/v1/incoming-webhooks/$WEBHOOK_ID/$TOKEN/wecom" \
 需登录态 + 群管理员权限，路径前缀 `/v1/groups/:group_no/incoming-webhooks`。
 
 创建 / 重置（regenerate）响应除历史的 `url`（native 路径）外，还带 `urls` 对象，
-按推送形态给出全部路径（`native` / `github` / `wecom`，不含 host，由前端拼接）。
+按推送形态给出全部路径（`native` / `github` / `wecom` / `multica`，不含 host，由前端拼接）。
 token 仅在这两处出现一次，list 不回显 token、也不回推送 URL。
 
 除创建/列出/更新/删除/重置外，Phase 2 新增两个：
--- a/modules/incomingwebhook/adapter.go
+++ b/modules/incomingwebhook/adapter.go
@@ -1,13 +1,14 @@
 package incomingwebhook
 
-// 推送形态适配层（#297 Phase 3）。
+// 推送形态适配层（#297 Phase 3 / #426）。
 //
-// 三种推送形态共享同一条鉴权 / 限流 / 群校验 / 投递 / 审计流水线（api.go handlePush），
+// 多种推送形态共享同一条鉴权 / 限流 / 群校验 / 投递 / 审计流水线（api.go handlePush），
 // 彼此只差「如何把请求 body 翻译成 native 推送请求」这一步：
 //
 //   - native（历史契约）   POST /v1/incoming-webhooks/:webhook_id/:token
 //   - GitHub 事件          POST .../:token/github   （adapter_github.go）
 //   - 企业微信群机器人格式  POST .../:token/wecom    （adapter_wecom.go）
+//   - Multica 出站事件     POST .../:token/multica  （adapter_multica.go）
 //
 // 适配器不是新的攻击面：URL token 鉴权、四层限流、群 Normal 校验、payload 白名单
 // 构造（buildPayload / buildRichTextPayload 注入 from.kind=webhook 与服务端 space_id）
@@ -50,6 +51,10 @@ var (
 		// 企业微信调用方普遍校验 errcode==0，附带平台习惯字段降低迁移摩擦。
 		successExtra: map[string]interface{}{"errcode": 0, "errmsg": "ok"},
 	}
+	// multicaAdapter 接收 multica 出站 webhook（issue.status_changed 等事件
+	// 的固定 JSON envelope）。multica envelope 比 GitHub 事件紧凑（不嵌入
+	// repository 对象），8 KiB 足够，沿用 native 的 bodyLimit。
+	multicaAdapter = pushAdapter{name: adapterMultica, parse: parseMulticaPush, bodyLimit: maxBytes}
 )
 
 // parseNativePush 是 native 形态的 parse：body 即 pushPayloadReq JSON 本身。
--- a/modules/incomingwebhook/adapter.go
+++ b/modules/incomingwebhook/adapter.go
@@ -22,7 +22,8 @@ import (
 
 // pushAdapter 描述一种推送形态。
 type pushAdapter struct {
-	// name 写入审计 adapter 列（adapterNative / adapterGitHub / adapterWeCom）。
+	// name 写入审计 adapter 列（adapterNative / adapterGitHub / adapterWeCom /
+	// adapterMultica）。新增适配器时把对应常量加进列表。
 	name string
 	// parse 把平台原始 body 翻译成 native 推送请求。三个返回值恰有一个生效：
 	//   - req     非 nil：照常走 msg_type 构造 / 投递；
@@ -127,4 +128,55 @@ var mdLinkTextEscaper = strings.NewReplacer(`\`, `\\`, `[`, `\[`, `]`, `\]`)
 // 钳约束的是可见长度，转义引入的反斜杠不该挤占可见字符预算。
 func mdLinkText(s string, max int) string {
 	return mdLinkTextEscaper.Replace(clipRunes(oneLine(s), max))
+}
+
+// mdInertTextEscaper 转义会让外部文本"越界"激活 markdown 语法的字符。覆盖：
+//   - 链接相关 `\` / `[` / `]`：与 mdLinkText 同源，避免拼成意外链接；
+//   - 强调相关 `*` / `_`：放进 `**X**`/`__X__` 包装或纯文本时不能让自身字符提前闭合
+//     或反生成强调；
+//   - HTML / 自动链接 `<` / `>`：CommonMark 会把 `<http://…>` 渲染成链接，
+//     `<script>` 在部分宽松渲染器里也会被转 HTML；
+//   - 表格管道 `|`：进表格单元格的字段不能裸传 `|`。
+//
+// 反引号「不转义、直接剥离」：CommonMark 单 backtick span 里 `\` 是字面反斜杠，
+// 用 `\“ 不能把反引号嵌进去；而双 backtick fence 又会引入二次转义难题。本模块
+// 用 mdInertText 处理的字段（multica 的 identifier / actor.type 等短标识符）按
+// 契约都是不含反引号的，剥离比转义更稳——与 GitHub adapter 处理短字段
+// （branch、sender.login）的口径一致。
+//
+// ⚠️ 不要把 mdInertText 用在 `...` code span 的内部：code span 内 `_` / `*`
+// 不被 markdown 解释，反斜杠转义会让客户端把 `\_` 当字面显示出来（in\_progress
+// 而非 in_progress）。code span 内只需 strip 反引号，请用 mdCodeSpanText。
+var mdInertTextEscaper = strings.NewReplacer(
+	`\`, `\\`,
+	"`", "",
+	`*`, `\*`,
+	`_`, `\_`,
+	`[`, `\[`,
+	`]`, `\]`,
+	`<`, `\<`,
+	`>`, `\>`,
+	`|`, `\|`,
+)
+
+// mdInertText 把不该激活 markdown 语法的短字段（identifier / 标签 / actor 类型）
+// 钳到 max rune 后转义为「inert」文本——拼到 `**X**` 或 `(by X)` 纯文本都不能
+// 逃出去开新 span / 链接。先钳后转义同 mdLinkText。
+//
+// 仅用于 **非** code span 上下文；code span 内字段请用 mdCodeSpanText（否则
+// `\_` 等转义序列会以字面形式回显，破坏显示）。
+func mdInertText(s string, max int) string {
+	return mdInertTextEscaper.Replace(clipRunes(oneLine(s), max))
+}
+
+// mdCodeSpanText 专用于 `...` 单反引号 code span 内部字段：只剥离反引号
+// （没法在单 backtick span 内安全转义反引号），其余字符在 code span 内不被
+// markdown 解释（`_` `*` `\` 都会按字面显示），无须转义——也绝不可转义，否则
+// 客户端会回显 `\_` 而非 `_`，破坏 status 等可读性。
+//
+// 与 mdInertText 不同：用法严格限定到 “ `X` “ 上下文，不要拿出去拼别处。
+var mdCodeSpanStripper = strings.NewReplacer("`", "")
+
+func mdCodeSpanText(s string, max int) string {
+	return mdCodeSpanStripper.Replace(clipRunes(oneLine(s), max))
 }
--- a/modules/incomingwebhook/adapter_multica.go
+++ b/modules/incomingwebhook/adapter_multica.go
@@ -1,0 +1,153 @@
+package incomingwebhook
+
+// Multica 事件适配器（#426）。
+//
+// 路由：POST /v1/incoming-webhooks/:webhook_id/:token/multica
+// 在 Multica 工作区 Settings → Webhooks 把 Webhook URL 配成上述地址即可，
+// 无需任何中间转换层。
+//
+// 鉴权沿用 URL 内的 128-bit token（与 native / github / wecom 一致）。Multica
+// 出站请求会带 X-Multica-Signature-256（与 GitHub 的 X-Hub-Signature-256 同
+// 算法），与 github 适配器对称——目前不校验，留作后续可选项。
+//
+// 渲染策略：按 envelope.event 把已支持事件翻译成 markdown 文本（走 native 纯
+// 文本路径，客户端按 markdown 渲染）。v1 只识别 issue.status_changed；未来
+// issue.created / issue.assigned / comment.created 等是【加性】的——新增 case 即可，
+// 不动既有契约。
+//
+// 子集之外的事件返回 200 + auditSkipped(reason=event)：Multica 侧投递成功
+// （不会把订阅标红），管理端 deliveries 里 reason=event 可见，与 github
+// 适配器的语义完全一致。
+//
+// mu* 结构体只声明渲染需要的字段（白名单解析），其余 payload 字段一律忽略。
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+// muIssue 是 multica IssueResponse 渲染所需的最小字段子集（白名单）。
+// Identifier 形如 "MUL-123"；Status 是状态枚举（todo/in_progress/...）。
+type muIssue struct {
+	Identifier string `json:"identifier"`
+	Title      string `json:"title"`
+	Status     string `json:"status"`
+}
+
+// muActor 是事件触发者标识（{type,id}）。type ∈ member / agent 等；id 是
+// 触发者主键（uuid 或 agent slug）。当前 envelope 不带显示名，渲染只用 type。
+type muActor struct {
+	Type string `json:"type"`
+	ID   string `json:"id"`
+}
+
+// muEnvelope 是 multica 出站 webhook 的固定信封（见 multica
+// server/internal/integrations/outwebhook/dispatcher.go outboundPayload）。
+type muEnvelope struct {
+	Event          string  `json:"event"`
+	WorkspaceID    string  `json:"workspace_id"`
+	Actor          muActor `json:"actor"`
+	Issue          muIssue `json:"issue"`
+	PreviousStatus string  `json:"previous_status"`
+}
+
+// 已知事件类型。v1 只有 issue.status_changed；新增事件加 case 即可。
+const (
+	muEventIssueStatusChanged = "issue.status_changed"
+)
+
+// parseMulticaPush 把 multica envelope 翻译成 native 推送请求（pushAdapter.parse）。
+//
+// 三个返回值恰有一个生效（见 pushAdapter.parse 注释）：
+//   - req     非 nil：照常走纯文本路径投递；
+//   - skip    非空：合法但刻意不投递（事件类型不在渲染子集内 → "event"），
+//     返回 200 + auditSkipped，管理端 deliveries 里 reason=event 可见；
+//   - invalid 非空：解析失败 → 400 + auditFailed，reason 同 native：
+//     "json" / "content"（事件已识别但渲染结果为空）。
+//
+// header 当前未使用（X-Multica-Event 仅作辅助；事件分发以 body.event 为准，
+// 防止 header/body 不一致导致渲染走偏）。保留参数以匹配 pushAdapter.parse 签名，
+// 并预留未来按 header 做版本协商的余地。
+func parseMulticaPush(_ http.Header, body []byte) (*pushPayloadReq, string, string) {
+	var ev muEnvelope
+	if err := json.Unmarshal(body, &ev); err != nil {
+		return nil, "", "json"
+	}
+	event := strings.TrimSpace(ev.Event)
+	if event == "" {
+		// 缺 event 字段是配置错误（不像合法 multica 出站流量），按 json 拒绝。
+		// 与 github 的 no_event 不同：github 那边事件名在 header，缺 header 与
+		// 缺 body 字段语义有别；multica 全在 body 里，统一走 json 即可。
+		return nil, "", "json"
+	}
+
+	var content string
+	switch event {
+	case muEventIssueStatusChanged:
+		content = renderMulticaIssueStatusChanged(ev)
+	default:
+		// 渲染子集之外（issue.created / comment.created / 未来未知事件）：
+		// multica 侧无需修复任何东西，200 + skip。
+		return nil, "event", ""
+	}
+	if content == "" {
+		// 事件类型已识别、但 payload 关键字段缺失渲染不出可读内容：按 400
+		// 拒绝而非静默跳过，让发送方能在 deliveries 里看到 reason=content
+		// 排查 payload 是否构造错（与 native content 为空的语义一致）。
+		return nil, "", "content"
+	}
+	// envelope 里的 title 长度由 multica 端控制（理论上可达 MySQL TEXT
+	// 上限）；钳到 maxContentRunes 内，绝不让 multica 流量 413。
+	return &pushPayloadReq{Content: clipRunes(content, maxContentRunes())}, "", ""
+}
+
+// renderMulticaIssueStatusChanged 渲染 issue.status_changed 事件。
+//
+// 输出形如（markdown，客户端会渲染）：
+//
+//	**MUL-123** Fix login redirect: `todo` → `in_progress` (by agent)
+//
+// 设计取舍：
+//   - 不拼 issue URL：envelope 当前不带 URL；服务端无法可靠推断 multica
+//     部署域名（self-hosted 路径多种）。需要可点链接时由 multica 后续在
+//     envelope 加 issue.url 字段，渲染层加链接即可，不破坏既有契约。
+//   - 标题与 identifier 之间用空格分隔，与 GitHub 适配器的 issue 渲染风格
+//     一致；状态用反引号包起来，避免下划线（in_progress）被 markdown 误
+//     当作斜体。
+//   - actor 名当前不在 envelope 里，只渲染 type（"by agent" / "by member"）；
+//     未显示触发者类型的情况下省略尾注。
+//   - 标题里的 `[` / `]` 经 mdLinkText 转义（即便当前没拼成链接也防御性
+//     处理，未来加链接时不必再回头改）。
+func renderMulticaIssueStatusChanged(ev muEnvelope) string {
+	id := strings.TrimSpace(ev.Issue.Identifier)
+	title := strings.TrimSpace(ev.Issue.Title)
+	status := strings.TrimSpace(ev.Issue.Status)
+	prev := strings.TrimSpace(ev.PreviousStatus)
+	// id 与 status 是渲染所必需的最小集；缺失任一无法生成可读消息，回退
+	// 让上层走 reason=content 而非凭空生造。
+	if id == "" || status == "" {
+		return ""
+	}
+
+	var b strings.Builder
+	fmt.Fprintf(&b, "**%s**", id)
+	if title != "" {
+		// 钳到 200 rune 与 GitHub PR/issue 标题渲染保持一致。
+		b.WriteString(" ")
+		b.WriteString(mdLinkText(title, 200))
+	}
+	b.WriteString(": ")
+	if prev != "" && prev != status {
+		fmt.Fprintf(&b, "`%s` → `%s`", prev, status)
+	} else {
+		// 新建 issue 或事件 payload 缺 previous_status 时只显示当前状态，
+		// 不画"→ X"那种暗示状态变化的尾巴。
+		fmt.Fprintf(&b, "`%s`", status)
+	}
+	if actorType := strings.TrimSpace(ev.Actor.Type); actorType != "" {
+		fmt.Fprintf(&b, " (by %s)", actorType)
+	}
+	return b.String()
+}
--- a/modules/incomingwebhook/adapter_multica.go
+++ b/modules/incomingwebhook/adapter_multica.go
@@ -65,7 +65,7 @@ const (
 //   - skip    非空：合法但刻意不投递（事件类型不在渲染子集内 → "event"），
 //     返回 200 + auditSkipped，管理端 deliveries 里 reason=event 可见；
 //   - invalid 非空：解析失败 → 400 + auditFailed，reason 同 native：
-//     "json" / "content"（事件已识别但渲染结果为空）。
+//     "json" / "no_event" / "content"（事件已识别但渲染结果为空）。
 //
 // header 当前未使用（X-Multica-Event 仅作辅助；事件分发以 body.event 为准，
 // 防止 header/body 不一致导致渲染走偏）。保留参数以匹配 pushAdapter.parse 签名，
@@ -77,10 +77,12 @@ func parseMulticaPush(_ http.Header, body []byte) (*pushPayloadReq, string, stri
 	}
 	event := strings.TrimSpace(ev.Event)
 	if event == "" {
-		// 缺 event 字段是配置错误（不像合法 multica 出站流量），按 json 拒绝。
-		// 与 github 的 no_event 不同：github 那边事件名在 header，缺 header 与
-		// 缺 body 字段语义有别；multica 全在 body 里，统一走 json 即可。
-		return nil, "", "json"
+		// 缺 event 字段是「配置错误」（不像合法 multica 出站流量），与 github
+		// 适配器缺 X-GitHub-Event 头同语义——复用同一个 no_event 原因码，让
+		// deliveries 排障口径跨适配器一致（yujiawei review P2-2）：跟「事件已识别但
+		// 不在渲染子集内」的 200 skip(event) 区分开。reason 字典在 #330 时已经为
+		// no_event 注册了 i18n / DB 注释，无须新增。
+		return nil, "", "no_event"
 	}
 
 	var content string
@@ -98,8 +100,10 @@ func parseMulticaPush(_ http.Header, body []byte) (*pushPayloadReq, string, stri
 		// 排查 payload 是否构造错（与 native content 为空的语义一致）。
 		return nil, "", "content"
 	}
-	// envelope 里的 title 长度由 multica 端控制（理论上可达 MySQL TEXT
-	// 上限）；钳到 maxContentRunes 内，绝不让 multica 流量 413。
+	// 8 KiB body cap 已经在 handlePush 第 4 步前置拦掉超长 body（足够大的 title
+	// 会先 413，而不是到这里），所以此处的 clipRunes 不是「保 title 不被 413」的
+	// 兜底，而是渲染层对 maxContentRunes(4000 rune) 上限的本地约束——避免拼装后
+	// 的最终 content 越过 push 路径下游对 RichText 字数的语义钳位。
 	return &pushPayloadReq{Content: clipRunes(content, maxContentRunes())}, "", ""
 }
 
@@ -120,6 +124,19 @@ func parseMulticaPush(_ http.Header, body []byte) (*pushPayloadReq, string, stri
 //     未显示触发者类型的情况下省略尾注。
 //   - 标题里的 `[` / `]` 经 mdLinkText 转义（即便当前没拼成链接也防御性
 //     处理，未来加链接时不必再回头改）。
+//   - 其余进 markdown 的字段按上下文走对应 helper（详见 adapter.go 注释）：
+//   - identifier 在 `**...**` 粗体里 → mdInertText（转义 `*` 防止 bold-break，
+//     转义 `[]` 防止 link injection）；
+//   - status / previous_status 在 “ `...` “ code span 里 → mdCodeSpanText
+//     （只剥反引号防止逃逸 code span；`_`/`*` 在 code span 内本就不被解释，
+//     不转义否则会回显 `\_` 破坏显示）；
+//   - actor.type 在 `(by ...)` 纯文本上下文 → mdInertText（全量转义）。
+//     这些字段按 multica 契约都是受控枚举/标识符，escape 是 defense-in-depth
+//     （PR #427 review by yujiawei / mochashanyao / Jerry-Xin）。
+//   - 字段长度统一钳到 64 rune：identifier/status/actor 在 multica 端是短
+//     标识符（identifier 形如 "MUL-12345"、status 是枚举、actor.type 是
+//     "member"/"agent"），64 足够覆盖正常上限并把异常长度截短防御。title
+//     仍用 200 rune（与 GitHub PR/issue 标题钳值对齐）。
 func renderMulticaIssueStatusChanged(ev muEnvelope) string {
 	id := strings.TrimSpace(ev.Issue.Identifier)
 	title := strings.TrimSpace(ev.Issue.Title)
@@ -131,8 +148,10 @@ func renderMulticaIssueStatusChanged(ev muEnvelope) string {
 		return ""
 	}
 
+	const shortFieldMax = 64
+
 	var b strings.Builder
-	fmt.Fprintf(&b, "**%s**", id)
+	fmt.Fprintf(&b, "**%s**", mdInertText(id, shortFieldMax))
 	if title != "" {
 		// 钳到 200 rune 与 GitHub PR/issue 标题渲染保持一致。
 		b.WriteString(" ")
@@ -140,14 +159,16 @@ func renderMulticaIssueStatusChanged(ev muEnvelope) string {
 	}
 	b.WriteString(": ")
 	if prev != "" && prev != status {
-		fmt.Fprintf(&b, "`%s` → `%s`", prev, status)
+		fmt.Fprintf(&b, "`%s` → `%s`",
+			mdCodeSpanText(prev, shortFieldMax),
+			mdCodeSpanText(status, shortFieldMax))
 	} else {
 		// 新建 issue 或事件 payload 缺 previous_status 时只显示当前状态，
 		// 不画"→ X"那种暗示状态变化的尾巴。
-		fmt.Fprintf(&b, "`%s`", status)
+		fmt.Fprintf(&b, "`%s`", mdCodeSpanText(status, shortFieldMax))
 	}
 	if actorType := strings.TrimSpace(ev.Actor.Type); actorType != "" {
-		fmt.Fprintf(&b, " (by %s)", actorType)
+		fmt.Fprintf(&b, " (by %s)", mdInertText(actorType, shortFieldMax))
 	}
 	return b.String()
 }
--- a/modules/incomingwebhook/adapter_multica.go
+++ b/modules/incomingwebhook/adapter_multica.go
@@ -75,7 +75,11 @@ func parseMulticaPush(_ http.Header, body []byte) (*pushPayloadReq, string, stri
 	if err := json.Unmarshal(body, &ev); err != nil {
 		return nil, "", "json"
 	}
-	event := strings.TrimSpace(ev.Event)
+	// lower+trim 与 api.go 里的 msg_type 处理保持一致——避免发送方误用大小写
+	// 变体把合法事件（"Issue.Status_Changed"）静默降级成 200 skip(event)
+	// (yujiawei review P2-2)。事件名按 multica 契约本就是 lowercase-dotted 枚举，
+	// 这里做规范化只是 defense-in-depth。
+	event := strings.ToLower(strings.TrimSpace(ev.Event))
 	if event == "" {
 		// 缺 event 字段是「配置错误」（不像合法 multica 出站流量），与 github
 		// 适配器缺 X-GitHub-Event 头同语义——复用同一个 no_event 原因码，让
@@ -122,8 +126,12 @@ func parseMulticaPush(_ http.Header, body []byte) (*pushPayloadReq, string, stri
 //     当作斜体。
 //   - actor 名当前不在 envelope 里，只渲染 type（"by agent" / "by member"）；
 //     未显示触发者类型的情况下省略尾注。
-//   - 标题里的 `[` / `]` 经 mdLinkText 转义（即便当前没拼成链接也防御性
-//     处理，未来加链接时不必再回头改）。
+//   - title 在 `** identifier ** title:` 的「裸文本」上下文（注意：与 github
+//     适配器的 title 不同——github 把 title 包进 `[text](url)` 链接里，所以那边
+//     用只转 `\[]` 的 mdLinkText 是对的；这里 title 不进链接，必须用
+//     mdInertText 才能把 `*`/反引号/`<`/`|` 等元字符也防住，否则一个含反引号
+//     的 title 会让 content 里反引号总数变奇数、把后面的 status code-span 切坏，
+//     一个含 `**` 的 title 会注入真粗体（yujiawei review P1）。
 //   - 其余进 markdown 的字段按上下文走对应 helper（详见 adapter.go 注释）：
 //   - identifier 在 `**...**` 粗体里 → mdInertText（转义 `*` 防止 bold-break，
 //     转义 `[]` 防止 link injection）；
@@ -153,9 +161,11 @@ func renderMulticaIssueStatusChanged(ev muEnvelope) string {
 	var b strings.Builder
 	fmt.Fprintf(&b, "**%s**", mdInertText(id, shortFieldMax))
 	if title != "" {
-		// 钳到 200 rune 与 GitHub PR/issue 标题渲染保持一致。
+		// 钳到 200 rune 与 GitHub PR/issue 标题渲染保持一致；用 mdInertText 而
+		// 非 mdLinkText——title 在这里是裸文本，不在 `[text](url)` 里（详见上方
+		// 设计取舍）。
 		b.WriteString(" ")
-		b.WriteString(mdLinkText(title, 200))
+		b.WriteString(mdInertText(title, 200))
 	}
 	b.WriteString(": ")
 	if prev != "" && prev != status {
--- a/modules/incomingwebhook/adapter_multica_test.go
+++ b/modules/incomingwebhook/adapter_multica_test.go
@@ -76,6 +76,68 @@ func TestParseMulticaPush_TitleEscaping(t *testing.T) {
 	assert.Contains(t, req.Content, `\[enter\]`, "brackets must be markdown-escaped (mdLinkText)")
 }
 
+// 非链接上下文的字段（identifier 在 bold、actor.type 在 plain text）必须经
+// mdInertText 转义；status 在 backtick code span 走 mdCodeSpanText 单独 strip
+// 反引号。否则一个带 `*` 的 identifier 可以提前闭合 **...**、带反引号的 status
+// 可以逃出 code span、带 `[...](url)` 的 actor 可以注入链接。三位 reviewer
+// (Jerry-Xin / yujiawei / mochashanyao) 都点了这道防御。
+func TestParseMulticaPush_IdentifierMarkdownEscaping(t *testing.T) {
+	body := []byte(`{
+		"event": "issue.status_changed",
+		"issue": {"identifier": "MUL-1** [phish](http://evil.com) **", "title": "ok", "status": "done"},
+		"previous_status": "todo"
+	}`)
+	req, skip, invalid := parseMulticaPush(http.Header{}, body)
+	require.NotNil(t, req, "skip=%q invalid=%q", skip, invalid)
+	// `*` 必须被反斜杠转义，否则会提前关闭 **...** 的粗体并让后面的 `[phish](...)`
+	// 渲染成 markdown 链接。
+	assert.Contains(t, req.Content, `\*\*`,
+		"asterisks in identifier must be escaped to prevent bold-break + link injection")
+	assert.Contains(t, req.Content, `\[phish\]`,
+		"square brackets in identifier must be escaped to prevent link injection")
+	assert.NotContains(t, req.Content, `[phish](`,
+		"un-escaped link syntax must not leak through")
+}
+
+func TestParseMulticaPush_StatusBacktickStripping(t *testing.T) {
+	// status 渲染在 `...` code span 内；嵌入反引号无法用 \\` 安全转义
+	// （单 backtick fence 里 \\` 仍是字面反斜杠），按契约 strip 反引号
+	// （mdCodeSpanText 行为）。
+	body := []byte(`{
+		"event": "issue.status_changed",
+		"issue": {"identifier": "MUL-1", "title": "x", "status": "in_pro` + "`" + `gress"},
+		"previous_status": "to` + "`" + `do"
+	}`)
+	req, _, _ := parseMulticaPush(http.Header{}, body)
+	require.NotNil(t, req)
+	// 反引号被剥后 status="in_progress"、prev="todo"，渲染成 `todo` → `in_progress`。
+	// 关键不变量：原始反引号不能在最终 content 中以「跨 code-span 的 ` 字符」形式
+	// 出现，否则就破坏了 code span。
+	assert.Contains(t, req.Content, "`todo` → `in_progress`",
+		"backticks inside status/previous_status must be stripped, restoring the normal display")
+	// 进一步钉一下：内容里 ` 的总数必须是偶数（成对开闭 code span），
+	// 不应因 strip 出现奇数个反引号导致开/闭不平衡。
+	assert.Equalf(t, 0, strings.Count(req.Content, "`")%2,
+		"backticks must remain balanced (got odd count): %q", req.Content)
+}
+
+func TestParseMulticaPush_ActorTypeMarkdownEscaping(t *testing.T) {
+	// actor.type 在 `(by X)` 纯文本上下文；带 `[label](url)` 的恶意值必须
+	// 不能拼出可点链接。
+	body := []byte(`{
+		"event": "issue.status_changed",
+		"actor": {"type": "agent [click](http://evil.com)", "id": "a-1"},
+		"issue": {"identifier": "MUL-2", "title": "x", "status": "done"},
+		"previous_status": "todo"
+	}`)
+	req, _, _ := parseMulticaPush(http.Header{}, body)
+	require.NotNil(t, req)
+	assert.Contains(t, req.Content, `\[click\]`,
+		"square brackets in actor.type must be escaped to prevent link injection")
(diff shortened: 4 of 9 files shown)


Jerry-Xin (dismissed):
The PR is in scope for `octo-server`: it extends the existing incoming webhook module with a new platform adapter and follows the established adapter pipeline.

**💬 Non-blocking**
- 🟡 Warning: Integration coverage was not extended for the new `/multica` route. `TestPush_AdapterRoute_AuthEnforced` still loops only over `github` and `wecom` in `modules/incomingwebhook/adapter_push_test.go:117`, and `TestCreate_ReturnsAdapterURLs` does not assert the `multica` URL in `modules/incomingwebhook/adapter_push_test.go:39`. The pure `publicURLs` helper test helps, but route registration/auth behavior would be better pinned at the same level as the existing adapters.
- 🔵 Suggestion: Add `adapter_multica.go` to the legacy response guard file list in `modules/incomingwebhook/api_i18n_test.go:22`. The new file currently has no response calls, so this is not a correctness issue, but it keeps future edits covered by the module’s i18n guard.
- 🔵 Suggestion: `identifier`, `status`, `previous_status`, and `actor.type` are interpolated directly into markdown in `modules/incomingwebhook/adapter_multica.go:135`, `:143`, `:147`, and `:150`. This is acceptable if those fields are fixed Multica-controlled enums/identifiers, but a small markdown/code-span escaping helper would make the renderer more robust if Multica ever allows custom status names or project keys.

**✅ Highlights**
- The adapter cleanly reuses the existing `pushAdapter` path, so auth, group validation, rate limiting, delivery, and audit semantics remain centralized.
- Unknown Multica events use the same skipped-delivery contract as GitHub, which matches the PR description and existing observability model.
- Pure adapter tests passed locally: `go test ./modules/incomingwebhook/ -run 'TestParseMulticaPush|TestMulticaAdapter|TestPublicURLs'`.
- DB-backed route tests could not be completed in this checkout because the local test database has an unrelated unknown migration recorded: `20201222000001_report_legacy01.sql`.

lml2468 (commented):
作者回应（我 @lml2468 是本 PR 作者，**回避投票**，仅认领问题 + 说明修法）：

@Jerry-Xin 第 3 点说得对，我确认 `renderMulticaIssueStatusChanged` 只转义了 title 这一半，其余进 markdown 的字段都是裸拼：

| 字段 | 拼法 | 转义? |
|---|---|---|
| Identifier | `**%s**` | ❌ 裸拼粗体 |
| Title | `mdLinkText(title,200)` | ✅ |
| Status / PreviousStatus | 反引号 code span | ❌ 裸拼 |
| Actor.Type | `(by %s)` | ❌ |

**风险（should-fix，非 blocking）**：
- Identifier 含 `*`/反引号可破坏粗体或注入 markdown；Status 含反引号可逃逸 code span；Actor.Type 含元字符可注入。
- 这些字段实践中是受控枚举/标识符（MUL-123、todo/in_progress），且终点是 IM markdown（非 HTML、无 XSS），危害是显示破坏/格式伪造而非 RCE/泄露 → 非 blocking。
- 但不应信任上游格式，且这与本模块 #421（FullName）、#423（PathWithNamespace）同源——**注入面的同类实例只修了一部分**（这次只转义 title，漏了 id/status/actor）。

**修法**：给非链接文本字段加一个 markdown 转义 helper（转义 `*` `_` 反引号 `[` `]` 等），Identifier/Status/PreviousStatus/Actor.Type 都过一遍；status 在反引号内尤其要转义反引号（否则逃逸 code span）。我作为作者来改这一版。

@Jerry-Xin 另两点（/multica 集成测试覆盖、i18n 响应保护列表）我一并在修复里补。

yujiawei (dismissed):
# Code Review — PR #427 (octo-server)

**Verdict: APPROVED** — no blocking issues. A few non-blocking suggestions below.

Reviewed at head `e5fccdef1d7f8e787a51860befb8bd9ba337dc12` against `main` (merge-base `d69ba8e6`). Build + `go vet` + the `incomingwebhook` package tests all pass locally.

## Summary

This is a clean, additive feature: a fourth incoming-webhook adapter (`/multica`) that mirrors the existing `github` / `wecom` adapters. It introduces no new attack surface — the new route is registered inside the same `chain()` wrapper and delegates to `handlePush(c, multicaAdapter)`, so it inherits the full pipeline before the adapter ever runs: webhook lookup, constant-time token compare, active-group check, lazy-disable creator-membership gate, the 4-layer rate limit, and the per-adapter body cap. The adapter itself only produces a `pushPayloadReq{Content}` and never touches identity/payload construction, so `from.kind=webhook` + server-derived `space_id` injection is preserved (no impersonation / cross-space forgery).

## Verification (security-sensitive checklist)

- ✅ **Auth pipeline inheritance** — `pushMultica` → `handlePush(c, multicaAdapter)` (`api.go`) is structurally identical to the three sibling handlers; route added at `api.go` inside the existing middleware chain.
- ✅ **HMAC deferral is symmetric, not a regression** — `X-Multica-Signature-256` is left unverified exactly as `github` defers `X-Hub-Signature-256` (`adapter_github.go:9-10`); the URL token remains the authoritative authn for all adapters.
- ✅ **No impersonation** — `parseMulticaPush` returns content only; it never sets `Username`/`AvatarURL`.
- ✅ **Body-cap DoS** — `bodyLimit = maxBytes` (8 KiB), enforced before parse; content additionally clipped to `maxContentRunes()` (4000). Tighter than github's 1 MiB; no amplification.
- ✅ **Audit correctness** — new `adapterMultica` constant flows through `submitFailure` / `submitSkipped` / `submitSuccess` with reason codes (`json` / `content` / `event`) consistent with the github adapter.
- ✅ **Return-value semantics** — all five `parseMulticaPush` branches return exactly one non-zero of `(req, skip, invalid)`; verified mutually exclusive.
- ✅ **Tests** — cover malformed JSON, missing event / identifier / status, unknown-event skip, long-title clip, bracket escaping, adapter registration, and the `multica` URL entry.

1. **`adapter_multica.go:101-103` — the "绝不让 413" comment (and the README "body 上限" section) is inaccurate for the exact case it cites.** The 8 KiB body byte-cap is enforced *before* `parseMulticaPush` runs, so the `clipRunes(content, maxContentRunes())` clip only defuses the post-clip rune-413; it gives no protection against the body-byte 413. A title large enough to push the JSON body past 8 KiB — the very "title up to MySQL TEXT size" scenario the comment invokes — will 413 before any clipping. The behavior is fine (413 is a clean, visible failure), but please either soften the comment/README to say titles beyond the 8 KiB body cap *will* 413, or — if a large title must never 413 — give multica its own wider `bodyLimit` the way github does (github uses 1 MiB precisely because the sender cannot shorten a platform-generated body, and multica's title is likewise not shortenable by the sender).

2. **`adapter_multica.go:78-84` — missing-event handling diverges from the sibling `no_event` convention.** The github adapter deliberately introduced a distinct `no_event` reason (`adapter_github.go:143`, plus a DB migration and i18n entry) so a misconfiguration is distinguishable in `deliveries` from a malformed body. multica folds missing-event into the generic `json` reason. The inline justification (multica carries the event in the body, not a header) is defensible, but reusing the already-registered `no_event` code would restore symmetry at zero cost and keep operator-facing diagnostics consistent.

3. **`adapter.go:25` — the `pushAdapter.name` field doc comment still lists only `native / github / wecom`.** The PR updated every other adapter enumeration (audit column, file header, `publicURLs` doc, README); this one doc line was missed.

## Nits (optional)

- **`adapter_multica.go:135,143-151`** — within `renderMulticaIssueStatusChanged`, `title` is escaped via `mdLinkText`, but `identifier`, `status`, `previous_status`, and `actor.type` are interpolated raw (the status fields inside backtick spans). This is bounded: the channel is token-authenticated and the whole content is clipped to 4000 runes, and a token holder can already send arbitrary markdown via the native path — so it is not a security issue under the current trust model. But the file header advertises defensive escaping of external text, and today the adapter relies entirely on the upstream Multica contract guaranteeing `status ∈ enum` / `identifier` shape. A one-line escape on these fields would make the renderer self-consistent and defense-in-depth if Multica ever reflects user-controlled text into `identifier`.

## Items worth a human's eye

- The `mu*` structs are whitelist-parse, so a field-name mismatch against Multica's real outbound `outboundPayload` (`workspace_id`, `previous_status`, `actor.type`) would silently degrade to `reason=content` rather than erroring. The mapping is asserted by comments/tests but the upstream dispatcher schema lives in another repo and could not be confirmed from this diff — worth a cross-repo sanity check that the envelope field names match.
- `X-Multica-Signature-256` exists only in prose; there is no constant/test pinning the exact header string for the future optional-HMAC work.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#427 Review Report
**Head SHA:** `e5fccdef1d7f8e787a51860befb8bd9ba337dc12`
**PR:** feat(incomingwebhook): add /Octo-Q adapter for Octo-Q outbound webhooks

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Adapter registration (name, parse, bodyLimit) | ✅ | `adapter.go:54` — `multicaAdapter` correctly wired with `adapterMultica`, `parseMulticaPush`, `maxBytes` |
| Route registration | ✅ | `api.go:232` — `push.POST("/incoming-webhooks/:webhook_id/:token/Octo-Q", chain(w.pushMultica)...)` uses identical middleware chain as native/github/wecom |
| Handler delegation | ✅ | `api.go:1079` — `pushMultica` is a thin wrapper calling `handlePush(c, multicaAdapter)` |
| publicURLs includes Octo-Q | ✅ | `api.go:404` — `"Octo-Q": base + "/Octo-Q"` added to urls map |
| Adapter constant | ✅ | `model.go:50` — `adapterMultica = "Octo-Q"` added alongside existing constants |
| Parse contract (req/skip/invalid) | ✅ | `adapter_multica.go:75-109` — exactly one of three return values is non-zero, matching `pushAdapter.parse` contract |
| Body limit (8 KiB) | ✅ | `adapter.go:54` — uses `maxBytes` (same as native/wecom), appropriate for compact Octo-Q envelopes |
| Content clipping | ✅ | `adapter_multica.go:108` — `clipRunes(content, maxContentRunes())` prevents 413 in handlePush step 6 |
| Title escaping | ✅ | `adapter_multica.go:137` — `mdLinkText(title, 200)` applies oneLine → clipRunes → bracket/backslash escaping |
| Test coverage (12 tests) | ✅ | `adapter_multica_test.go` — covers happy path, edge cases, escaping, clipping, error paths, registration, publicURLs |
| HMAC gap documented | ✅ | `adapter_multica.go:10-12`, `README.md:196-198` — explicitly noted as follow-up, symmetric with github adapter |
| Auth: no new attack surface | ✅ | Same URL token auth, same middleware chain, same handlePush pipeline — adapter only controls body parsing |

## 2. Findings

### P2-1 — Markdown escaping gap for non-title envelope fields

**File:** `adapter_multica.go:126-153` (`renderMulticaIssueStatusChanged`)

**What:** `title` is correctly escaped via `mdLinkText(title, 200)` (oneLine → clipRunes → bracket/backslash escaping). However, four other envelope fields are rendered into markdown contexts with only `strings.TrimSpace`:

- `identifier` → `**<id>**` (bold context, line 133)
- `status` → `` `<status>` `` (backtick context, lines 141/144)
- `previous_status` → `` `<prev>` `` (backtick context, line 141)
- `actor.Type` → ` (by <type>)` (plain text context, line 148)

A crafted envelope with `identifier = "MUL-1** [phish](http://evil.com) **" ` would produce: `**MUL-1** [phish](http://evil.com) ****` — breaking bold formatting and injecting a markdown link.

**Diff-scope three questions (R2):**
1. **New / amplified / pre-existing?** New — this code is introduced by this PR.
2. **Does the PR change an existing path's contract?** No — this is a new adapter with its own rendering function.
3. **Pre-existing on base?** N/A — the file doesn't exist on base.

**Mitigating factors (why P2, not P1):**
- The GitHub adapter has the **same pattern** for short/constrained fields: `sender.Login` and branch names are rendered unescaped in `**<login>**` and `` `<branch>` `` contexts (`adapter_github.go:183,192-194`). Only long-form text (PR titles, commit messages, comments) goes through `mdLinkText`. The Octo-Q adapter is consistent with this established pattern.
- Octo-Q platform constrains these fields in practice: `identifier` follows `<PREFIX>-<N>` format, `status` is an enum (`todo`/`in_progress`/`done`/etc.), `actor.Type` is a category string (`member`/`agent`).
- The webhook URL token (256-bit entropy, SHA-256 hashed, constant-time compared) is the trust boundary — an attacker needs the token to send any payload at all, at which point the native push path already allows arbitrary content.
- Impact is limited to visual formatting manipulation in chat messages; no code execution, no data exfiltration, no user-visible error data.

**Recommendation:** Apply `mdLinkText` (or at minimum `oneLine` + bracket escaping) to `identifier` and `actor.Type` for defense-in-depth. `status` and `previous_status` inside backticks are lower risk (backticks suppress markdown interpretation in most renderers) but `oneLine` would prevent multi-line injection.

### P2-2 — HMAC verification deferred (documented follow-up)

**File:** `adapter_multica.go:10-12`, `README.md:196-198`

**What:** `X-Octo-Q-Signature-256` is noted but not verified. This is explicitly documented as a follow-up, symmetric with the GitHub adapter's `X-Hub-Signature-256` gap.

**Diff-scope:** New code, but the design decision is pre-existing (github adapter has the same gap, documented in `adapter_github.go:9-10`). The HMAC infrastructure already exists in `modules/webhook/hmac.go` for when either adapter opts in.

**Verdict:** Acceptable for v1. Track as follow-up.

## 3. Suggestions

1. **Escape `identifier` and `actor.Type`** in `renderMulticaIssueStatusChanged` — even though the practical risk is low and the pattern is consistent with the GitHub adapter, defense-in-depth is cheap here. A `sanitizeField(s string) string` helper applying `oneLine` + `mdLinkTextEscaper.Replace` (without the 200-rune clip, since these are short fields) would close the gap for all four fields.

2. **Consider adding a test for `identifier` with markdown metacharacters** — the existing `TestParseMulticaPush_TitleEscaping` covers `title` but no test exercises `identifier` or `actor.Type` with adversarial input. Adding one would document the expected behavior and prevent regression if escaping is added later.

## 4. Additional Observations

- **Clean adapter pattern adherence:** The implementation closely follows the established adapter pattern. Route registration, middleware chain, handlePush pipeline, audit model, and publicURLs are all correctly extended. No shortcuts or bypasses.
- **White-list parsing:** `muEnvelope` / `muIssue` / `muActor` structs declare only the fields needed for rendering. `json.Unmarshal` silently ignores unknown fields — this is correct and defensive.
- **Error path semantics:** Missing `event` → `invalid="json"` (not `"no_event"` like GitHub) is a reasonable choice since Octo-Q puts the event in the body, not the header. The comment at line 86-89 explains the rationale clearly.
- **`MsgType` is intentionally empty:** The adapter produces plain-text content only. `handlePush` step 6 correctly routes `MsgType=""` to the text path. No rich-text/blocks support needed for v1.
- **Rate limiting is shared:** The per-webhook 5 rps bucket is shared across all adapter routes for the same webhook_id. This is by design (documented in code comments) and correct — a webhook receiving both github and Octo-Q traffic should not get double quota.

## 5. Data Flow Tracing

| Consumed Data | Upstream Source | Flows to Consumption Point? |
|---|---|---|
| `ev.Event` | `json.Unmarshal(body, &ev)` from HTTP request body | ✅ Yes — body is read by `handlePush` step 4 (`io.ReadAll(LimitReader(body, 8193))`), passed to `parseMulticaPush`. Empty event → `invalid="json"`. |
| `ev.Issue.Identifier` | Same JSON unmarshal, `issue.identifier` field | ✅ Yes — TrimSpace'd, checked for empty (returns "" → `invalid="content"`), rendered in bold. |
| `ev.Issue.Title` | Same JSON unmarshal, `issue.title` field | ✅ Yes — TrimSpace'd, optional (skipped if empty), escaped via `mdLinkText(title, 200)`. |
| `ev.Issue.Status` | Same JSON unmarshal, `issue.status` field | ✅ Yes — TrimSpace'd, checked for empty (returns "" → `invalid="content"`), rendered in backticks. |
| `ev.PreviousStatus` | Same JSON unmarshal, `previous_status` field | ✅ Yes — TrimSpace'd, optional (skipped if empty or same as status). |
| `ev.Actor.Type` | Same JSON unmarshal, `actor.type` field | ✅ Yes — TrimSpace'd, optional (skipped if empty), rendered in parentheses. |
| `pushPayloadReq.Content` | `renderMulticaIssueStatusChanged` output, clipped by `clipRunes(content, maxContentRunes())` | ✅ Yes — flows to `handlePush` step 6, takes text path (MsgType=""), content is non-empty (checked), length ≤ maxContentRunes (already clipped). |
| `multicaAdapter.bodyLimit` | `maxBytes` function (reads env `DM_INCOMINGWEBHOOK_MAX_BYTES`, default 8192) | ✅ Yes — used by `handlePush` step 4 to limit body read. 8 KiB sufficient for Octo-Q envelopes. |
| `multicaAdapter.name` | `adapterMultica = "Octo-Q"` constant | ✅ Yes — passed to `submitSuccess`/`submitFailure` for audit trail. |

**No data flow gaps detected.** Every consumed value traces back to a verified upstream source. No empty-array/undefined/early-return/short-circuit risks — the parse function explicitly checks for empty required fields and returns appropriate error codes.

**C1 — Dual-path parity (symmetric path self-check):**
- New `/Octo-Q` route uses identical middleware chain (`requirePushEnabled` → `localFloor` → `ipLimit` → `ipFailureGate`) as native/github/wecom. ✅ Parity confirmed.
- `pushMultica` handler is a one-liner delegating to `handlePush(c, multicaAdapter)`. Same pattern as `push`/`pushGitHub`/`pushWeCom`. ✅ Parity confirmed.
- No create/remove or subscribe/unsubscribe symmetry to check (this is a stateless push adapter).
- **Guard/lint tool coverage:** The route registration at `api.go:232` uses the same `chain(w.pushMultica)...` pattern as the other three routes. The `chain` function applies all four middleware layers uniformly. No guard bypass possible.

**C2 — Control-flow ordering / nested reuse:**
- `parseMulticaPush` is called from exactly one site (`handlePush` step 5 via `ad.parse`). No nested or repeated calls.
- `renderMulticaIssueStatusChanged` is called from exactly one site (the `case muEventIssueStatusChanged` branch in `parseMulticaPush`). No double-rendering risk.
- `mdLinkText` is applied once to `title`. No double-escaping risk (the function is idempotent on already-escaped input, but it's only called once anyway).
- **Non-canonical form test:** `actor.Type` with leading spaces → TrimSpace handles it. With unicode line separators → not handled (no `oneLine`), but backtick/parenthesis contexts are not broken by NEL/VT in standard markdown renderers. Low risk.

- The new route has the same authorization boundary as existing adapter routes: URL token (256-bit, SHA-256 hashed, constant-time compared). No new capability is exposed — the adapter only translates body format, it does not grant any additional permissions.
- The adapter cannot override `username`/`avatar_url` (both empty in `pushPayloadReq`), so `resolveFromIdentity` always falls back to the webhook's stored display identity. No impersonation vector.

**C4 — Authorization lifecycle / container-member state cascade:**
- N/A for this PR. The adapter does not introduce any new authorization logic. The existing `handlePush` pipeline already checks: webhook status (enabled/disabled/deleted), group Normal status, and creator membership (with lazy cascade disable if creator left group). These checks apply uniformly to all adapter routes including the new `/Octo-Q` route.

**C5 — Build/note ≠ runtime path correctness:**
- N/A. This is server-side Go code. No browser extension, CLI tool, or packaging involved. `go build` and `go test` are the relevant verification steps, and the PR reports 12 passing tests.

**C6 — Governance / policy / security document self-consistency:**
- `README.md` changes are documentation for the new adapter. The HMAC gap is documented consistently with the GitHub adapter's documentation. No policy conflicts detected.
- The README accurately describes the error semantics (200+skip for unknown events, 400 for parse/content failures) matching the implementation.

---

## Verdict

**[Octo-Q] verdict: APPROVE**

The implementation is well-structured, follows the established adapter pattern faithfully, and introduces no P0/P1 issues. Auth, rate limiting, group validation, and audit are all correctly inherited from the shared `handlePush` pipeline with no bypasses. The two P2 findings (markdown escaping for non-title fields, HMAC deferral) are consistent with existing adapter patterns and do not meet the P1 bar under R1 (no working path is made unavailable, no user-visible error data is produced, and the implementation is not worse than the status quo).

lml2468 (commented):
作者回应（我 @lml2468 是本 PR 作者，**回避投票**，仅认领 + 说明修法）：

@yujiawei 两点都对，我确认：

**1. 8KiB 体积限制 vs clip 注释自相矛盾（should-fix 注释准确性，非 blocking）**
- `multicaAdapter.bodyLimit = maxBytes = 8KiB`（adapter.go:57，与 native 同，非 github 的 1MiB 宽上限）。body 读取在 `io.LimitReader(body, limit+1)`（api.go:1197）处，超 8KiB 直接 413。
- 但 adapter_multica.go:101-102 注释写「title 理论可达 **MySQL TEXT 上限**……钳到 maxContentRunes 内，**绝不让 multica 流量 413**」——这不成立：若 title 真能远超 8KiB，整个 body 早在 8KiB 处 413 了，**根本到不了 clipRunes 那步**。clip 钳的是 maxContentRunes（4000 rune），但 8KiB bodyLimit 先生效。
- 不是行为 bug（8KiB 限制 + clip 都合理），是注释承诺与实际行为不符。修法二选一：① 删掉注释里「可达 MySQL TEXT 上限 / 绝不 413」的错误描述，如实写「body 超 8KiB 即 413；clip 仅防 8KiB 内的较长 title 污染渲染」；② 若确需支持大 title，给 multica 配 github 式宽 bodyLimit（但 multica envelope 是结构化小 JSON，8KiB 应足够，倾向改注释）。

**2. 缺失事件未用统一 no_event reason code（nit，非 blocking）**
合理。parseMulticaPush 缺 event 时返回 `"json"`，而 github 路径用 `no_event`。我注释里解释过「multica 全在 body 里、缺 event 视同 json 配置错」——但从 deliveries 诊断一致性看，缺事件用统一的 `no_event` reason 更便于跨适配器排查。我会统一（或在注释里讲清为何 multica 故意不同）。

两点连同 @Jerry-Xin 的状态字段转义 + 测试 + i18n，我作为作者一并在修复版处理。

lml2468 (commented):
作者回应（我 @lml2468 是本 PR 作者，**回避投票**，认领 + 一个事实校正）：

@mochashanyao 这点与 @Jerry-Xin 第 3 点同源（非标题字段缺 markdown 转义），我已认领。但要**校正你建议里的一个前提**：

**「参考 github 适配器统一转义」——github 适配器本身也没完整转义。** 我核了 adapter_github.go：
- `who`（ghLogin）拼进 `**%s**`、`ref`（branch/tag）拼进反引号、commit SHA / message firstLine——**全是裸拼，只有 Title 走 mdLinkText**。
- 这正是 #421 我指出过的 `r.FullName`（adapter_github.go:313）未转义的同款问题。

所以 github **不是干净范本**，它和 multica 犯同样的病（只转 title，who/ref/identifier/status/actor 裸拼）。这说明这是 **incomingwebhook 模块的系统性缺陷，非 multica 独有**——#421（FullName）、#423（PathWithNamespace）、#427（multica id/status/actor）是同一个面在不同适配器反复漏。

**正确修法（不是抄 github，而是根治）**：
- 建一个统一的 `mdEscapeText(s)` helper（转义 `*` `_` 反引号 `[` `]` `~` 等 markdown 元字符；反引号字段尤其要处理逃逸 code span）。
- **所有适配器**（native/github/wecom/multica）的**所有非链接文本字段**都过它：multica 的 id/status/prev/actor、github 的 who/ref/SHA/FullName、gitlab 的 PathWithNamespace 等。
- mdLinkText 继续用于链接文本部分（`[text]` 那半），两者职责分开。

我作为作者来落实 multica 这版的字段转义，并提一个 follow-up issue 把 github/gitlab/wecom 的同款裸拼一并根治（跨适配器系统修复，单独 PR 更清晰）。这是真实但非 blocking 的格式破坏面（IM markdown、无 XSS），不阻塞本 PR 的合并判断。

Jerry-Xin (commented):
Follow-up to my earlier APPROVE — correcting an omission on my part.

In my prior review I only verified that `title` is markdown-escaped via `mdLinkText`. On a closer read of `renderMulticaIssueStatusChanged` (modules/incomingwebhook/adapter_multica.go), the sibling fields are interpolated into the same markdown string **without** escaping:

- L135 `fmt.Fprintf(&b, "**%s**", id)` — `Identifier` bolded, bare
- L139 `b.WriteString(mdLinkText(title, 200))` — `title` escaped ✅ (the only escaped field)
- L143 `fmt.Fprintf(&b, "`%s` → `%s`", prev, status)` — `prevStatus`/`status` in code spans, bare
- L147 `fmt.Fprintf(&b, "`%s`", status)` — `status` in code span, bare
- L150 `fmt.Fprintf(&b, " (by %s)", actorType)` — `actor.Type` bare

This is the same `#421`/`#423` injection family — it looks like only one instance of the class (the title) got fixed here, while the others on the same line of output were left raw.

Severity: **non-blocking should-fix.** These fields are controlled enums/identifiers coming from multica, and the output is IM markdown (not HTML), so there's no XSS exposure. Worst case is display-format breakage — e.g. a backtick in an `id` or `status` value escaping its code span and mangling the rest of the message, or a stray `*` in an identifier flipping bold rendering.

Recommendation:
- Route `id`, `status`/`prevStatus`, and `actor.Type` through a shared markdown-escape helper before interpolation, so escaping covers **all** rendered fields rather than just `title`.
- Add a test that pins the escaping behavior for these fields (e.g. an `id`/`status` containing a backtick or `*`), so the contract doesn't silently regress.

My verdict stays **APPROVE** — this is non-blocking and the author has already acknowledged it. Flagging here so it's tracked alongside the existing fix.

lml2468:
Pushed `a03b55a` addressing all three review threads (@Jerry-Xin, @yujiawei, @mochashanyao). Recap:

### Markdown escaping for non-title envelope fields
- Added two helpers in `adapter.go`: `mdInertText` (for bold/plain text — escapes `\` `*` `_` `[` `]` `<` `>` `|`, strips backticks) and `mdCodeSpanText` (for `` `X` `` code spans — only strips backticks, since `\_` / `\*` inside a code span renders the backslash literally and breaks display).
- `identifier` and `actor.type` now flow through `mdInertText`; `status` / `previous_status` through `mdCodeSpanText`.
- 3 adversarial-input tests added: `identifier = "MUL-1** [phish](http://evil.com) **"`, status with embedded backticks (asserts balanced final-string backtick count, not just the bad pattern's absence), actor.type with `[label](url)`.

### `no_event` parity with github (yujiawei P2-2)
- Missing `event` field → `400 reason=no_event` (was `json`). Reuses the already-registered code; deliveries diagnostics now match github's `missing X-GitHub-Event` semantics. Existing test updated, README updated.

### Test / doc / guard plumbing
- `TestPush_AdapterRoute_AuthEnforced` and `TestCreate_ReturnsAdapterURLs` now cover `multica`; added 3 new `TestPush_Multica*` integration tests (delivers, unsupported event skipped, missing event → no_event reason) mirroring the github/wecom shape (Jerry-Xin's main warning).
- `adapter_multica.go` added to the i18n legacy-response guard list (Jerry-Xin's suggestion).
- `pushAdapter.name` doc comment now lists `adapterMultica` (yujiawei P2-3).
- Softened the body-cap comment in `adapter_multica.go`: clarified the 8 KiB cap is enforced upstream in `handlePush`, and the local `clipRunes` is the rune-count safeguard for `maxContentRunes(4000)`, not a body-byte 413 defense (yujiawei P2-1).

### Re: `check-sprint`
The failing `check-sprint / check-sprint` is a process check needing linked issue #426 added to the Octo Board project (Sprint W25). Could a maintainer with project write access bump it? My current gh token doesn't carry `read:project`.

Pure-adapter tests pass locally; the DB-backed `TestPush_Multica*` ones will run under CI's full env (same shape as the existing github/wecom suites). Live E2E (multica → octo `/multica`) re-verified against this commit + the companion PR `Mininglamp-OSS/multica#39` revision.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#427 Review Report
**Head SHA:** `a03b55a6e61ad71011ad3ff280a659763fd8a315`
**PR:** feat(incomingwebhook): add /Octo-Q adapter for Octo-Q outbound webhooks
**Complexity:** security_sensitive | **Size:** XL (+566/−23, 8 files)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Route registration | ✅ | `api.go:233` — `push.POST("/incoming-webhooks/:webhook_id/:token/Octo-Q", chain(w.pushMultica)...)` uses identical `chain()` middleware as native/github/wecom |
| Middleware parity | ✅ | `api.go:226-229` — same `chain` func: `requirePushEnabled → localFloor → ipLimit → ipFailureGate → handler` |
| Adapter struct | ✅ | `adapter.go:55-58` — `multicaAdapter{name: adapterMultica, parse: parseMulticaPush, bodyLimit: maxBytes}` |
| Audit constant | ✅ | `model.go:53` — `adapterMultica = "Octo-Q"` added alongside existing constants |
| publicURLs | ✅ | `api.go:401-406` — `"Octo-Q": base + "/Octo-Q"` added to map |
| Handler entry | ✅ | `api.go:1078` — `pushMultica` calls `handlePush(c, multicaAdapter)` |
| i18n test | ✅ | `api_i18n_test.go:23` — `adapter_multica.go` added to banned-legacy-error scan list |
| Markdown escaping | ✅ | Three-tier context-aware escaping: `mdInertText` (bold/plain), `mdLinkText` (link text), `mdCodeSpanText` (code span) — see §5 |
| Test coverage | ✅ | 12 unit tests + 3 integration tests covering happy path, edge cases, injection defense, auth enforcement |

## 2. Findings

### No P0 or P1 findings.

### P2 / Nits

**P2-1: Cosmetic — empty title renders as `**: \`status\``**
- **File:** `adapter_multica.go:157-160`
- When `ev.Issue.Title` is empty but `identifier` and `status` are present, the output becomes `**MUL-123**: \`todo\` → \`in_progress\`` — the colon-space separator dangles after the bold identifier. Not a functional issue (message still delivers and is readable), but cosmetically imperfect.
- **Suggestion:** Conditionally emit `": "` only when title is non-empty, or use `" — "` as separator when title is absent. Low priority.

**P2-2: GitHub adapter `ghLogin()` doesn't escape sender.login for bold context (pre-existing observation)**
- **File:** `adapter_github.go:183,196` — `who := ghLogin(ev.Sender)` → `**%s** pushed...`
- **Diff-scope:** pre-existing (not introduced or amplified by this PR)
- The Octo-Q adapter correctly uses `mdInertText` for identifier/actor in bold/plain context. The GitHub adapter uses raw `ghLogin()` output in `**%s**` bold. GitHub usernames are constrained to `[a-zA-Z0-9-]` by the platform, so this is not exploitable in practice. Noting as context — the Octo-Q adapter is strictly more defensive.
- **No action required for this PR.**

## 3. Suggestions

1. **(Optional)** Consider adding a test for empty-title rendering to pin the expected behavior (whether `**: ` or a cleaner alternative).
2. **(Future)** When HMAC verification of `X-Octo-Q-Signature-256` is implemented, mirror the pattern from `modules/webhook/hmac.go` as noted in the PR body.

## 4. Additional Observations

- The `mdInertTextEscaper` and `mdCodeSpanStripper` are well-designed context-aware escaping helpers. The documentation in `adapter.go:132-178` clearly explains when to use each, including the warning against using `mdInertText` inside code spans. This is a solid foundation for future adapters.
- The `muEnvelope` struct uses value types (not pointers), so JSON `null` for `previous_status` is safely absorbed as empty string by `json.Unmarshal` — no nil-pointer risk.
- The `bodyLimit: maxBytes` (8 KiB) is appropriate: Octo-Q envelopes are compact single-issue payloads, unlike GitHub events which need the 1 MiB `githubMaxBytes`.

## 5. Data-Flow Tracing

|---|---|---|
| `ev.Event` | `json.Unmarshal(body, &ev)` → `muEnvelope.Event` | ✅ TrimSpace → switch dispatch; empty → `"no_event"` 400; unknown → `"event"` skip 200 |
| `ev.Issue.Identifier` | `muEnvelope.Issue.Identifier` | ✅ TrimSpace → empty check (returns `""` → content 400) → `mdInertText(id, 64)` → `**%s**` bold context |
| `ev.Issue.Title` | `muEnvelope.Issue.Title` | ✅ TrimSpace → empty = skip title block → `mdLinkText(title, 200)` → plain text after bold |
| `ev.Issue.Status` | `muEnvelope.Issue.Status` | ✅ TrimSpace → empty check (returns `""` → content 400) → `mdCodeSpanText(status, 64)` → `` `...` `` code span |
| `ev.PreviousStatus` | `muEnvelope.PreviousStatus` | ✅ TrimSpace → empty = no arrow → `mdCodeSpanText(prev, 64)` → `` `...` `` code span |
| `ev.Actor.Type` | `muEnvelope.Actor.Type` | ✅ TrimSpace → empty = no "(by …)" → `mdInertText(actorType, 64)` → `(by %s)` plain text |
| Final `content` | `renderMulticaIssueStatusChanged` return | ✅ `clipRunes(content, maxContentRunes())` in parseMulticaPush → `pushPayloadReq.Content` → handlePush step 6 text path → `buildPayload` → message delivery |

All consumed data traced to upstream source. No empty-array/undefined/short-circuit gaps found.

## 6. Blind-Spot Checklist (R5 — security_sensitive PR, all items mandatory)

- **C1 Dual-path parity:** ✅ Route registration uses identical `chain()` middleware. `publicURLs` includes `Octo-Q`. `pushMultica` calls same `handlePush` pipeline. No create/delete or subscribe/unsubscribe asymmetry — adapter is purely translational. Auth-enforced test (`adapter_push_test.go:118`) updated to include `"Octo-Q"` in suffix list.
- **C2 Control-flow ordering / nested reuse:** ✅ `parseMulticaPush` is single-entry, no nested reuse. `mdInertTextEscaper` / `mdCodeSpanStripper` are new `strings.NewReplacer` instances, not composed with existing escapers. Non-canonical试穿: identifier with `*` → `\*` (tested), `[` → `\[` (tested), backtick → stripped (tested), `<script>` → `\<script\>` (covered by escaper definition).
- **C3 Authorization boundary ≠ capability boundary:** ✅ New route behind same URL-token auth + 4-layer rate limiting + group Normal check + creator membership check. No new auth bypass. No new tool/endpoint exposure beyond the push path.
- **C4 Authorization lifecycle / container-member cascade:** N/A — PR does not modify auth logic. The existing `cachedRequireActiveGroup` (group status) and `cachedCreatorMembership` (creator membership) checks in `handlePush` apply unchanged.
- **C5 Build ≠ runtime path:** N/A — server-side Go, no browser extension / CLI / packaging / relative URL concerns.
- **C6 Governance / policy doc consistency:** N/A — README additions document the new adapter accurately. No conflict with existing security or governance documentation.

---

**[Octo-Q] verdict: APPROVE**

Clean, well-structured implementation following the established adapter pattern. Middleware chain, audit trail, error handling, and markdown escaping are all correctly implemented with parity to existing adapters. Test coverage is thorough (12 unit + 3 integration). No P0/P1 findings. Only cosmetic P2 nits.

yujiawei (changes requested):
# Code Review — PR #427 (octo-server)

Reviewed at head `46d14fa310a741a21db7bdf52da5cce2cb34a34d` (the current head, which is one commit ahead of the originally-opened revision). This is a security-sensitive change (new unauthenticated-by-token push endpoint), so it got an extra pass on the input-rendering path.

## Verdict: Changes Requested

One P1 blocker on the markdown-escaping of the issue `title`. Everything else (routing, auth/rate-limit reuse, body cap, reason-code semantics) is sound, and the CI-red test failure from the prior revision is already fixed on this head.

---

## P1 (blocker) — `title` is escaped with the wrong helper, enabling markdown-formatting injection

`adapter_multica.go:159` renders the issue title with `mdLinkText(title, 200)`:

```go
if title != "" {
    b.WriteString(" ")
    b.WriteString(mdLinkText(title, 200))   // <-- wrong helper for this context
}
```

`mdLinkText` only escapes `\ [ ]`. That set is correct **inside** a `[text](url)` link (which is how the GitHub adapter uses it — `adapter_github.go:242/258/272/293`), because `* _ < > |` and backticks are inert in link-text. But here the title is written as **bare plain text** after `**identifier**`, where those characters are all active markdown. The module already ships the right helper for exactly this non-link context — `mdInertText` (used for `identifier` and `actor.type`) — and `title` is the one rendered field that doesn't use it.

Reproduced against `renderMulticaIssueStatusChanged` on this head:

| title input | rendered content | problem |
|---|---|---|
| `**x**` | `` **MUL-1** **x**: `todo` → `in_progress` `` | injected real bold |
| `` a `b `` | `` **MUL-1** a `b: `todo` → `in_progress` `` | **5 backticks (odd)** — hijacks the status code-span |
| `<script>alert(1)</script>` | passes through unescaped | HTML / autolink on lenient renderers |
| `a \| b` | bare pipe survives | table-pipe break |

The backtick case is the most concrete: a title with a single backtick makes the total backtick count odd, so the parser pairs the title's backtick with a status backtick and corrupts the `` `status` `` rendering. The suite asserts balanced backticks as an invariant (`TestParseMulticaPush_StatusBacktickStripping`), but that invariant is only enforced via `mdCodeSpanText` on `status`/`previous_status` — the title path bypasses it, so a title alone can produce the exact odd-backtick state the tests treat as impossible.

The title is attacker-influenceable (any workspace member who can set an issue title), and per the PR description the inbound `X-Multica-Signature-256` is intentionally not verified, so the envelope body is reachable by anyone holding the URL token. Worst confirmed impact is message-integrity corruption / formatting spoof in the group message (not clickable-link injection — `[ ]` are escaped — and not script execution, which depends on the client renderer). Still, this is a chat product and the defense was clearly intended; shipping a field that can spoof formatting and break the project's own output invariant warrants a fix first.

**Fix (one line):** render the title with the inert-text helper in this bare-text context.

```go
b.WriteString(mdInertText(title, 200))
```

`mdInertText` escapes `* _ [ ] < > |` and strips backticks, which closes all four vectors above. Please also add a title-injection test (a title containing `**`, a backtick, and `<...>`) alongside the existing identifier/actor escaping tests so the contract is pinned.

---

## Verified fixed on this head (was the prior CI-red blocker)

The previous revision shipped `TestPush_MulticaMissingEvent_NoEventReason` asserting `"reason":"no_event"` in the HTTP response body. That failed deterministically because the e2e harness (`testutil.NewTestServer`) does not install the i18n error renderer, so e2e bodies only render the legacy `{msg,status}` shape and `details.reason` never appears. The latest commit drops the body assertion and keeps the `400` check; the reason-routing contract remains pinned by `TestPushPayloadInvalidSurfacesReason` (which uses the i18n harness and already covers `no_event`). Production wiring (`main.go` installs the i18n renderer) surfaces `details.reason` correctly. `go test ./modules/incomingwebhook/` is green on this head. Good fix.

---

## P2 (non-blocking) — cross-repo contract robustness

These mirror the existing GitHub adapter's conventions, so they're consistent, not regressions — flagging because the two repos version independently.

1. **Strict scalar typing rejects the whole payload.** `parseMulticaPush` does one `json.Unmarshal` into an all-string struct. Any single type drift — including `actor.id`, which the renderer never reads — fails the whole event with `400 reason=json`. If Multica ever serializes `actor.id` numerically (e.g. a DB PK), valid `issue.status_changed` events silently stop delivering. Consider decoupling non-rendered fields or documenting the all-string contract with a pinning test.
2. **Event matching is case-sensitive.** `adapter_multica.go:89` switches on the raw (whitespace-trimmed) event string with no `ToLower`, while the `msg_type` switch in `api.go` does lower+trim. A case-variant event would silently become `200 skip(event)` rather than render. Low risk (event names are a controlled lowercase-dotted enum), but the internal inconsistency is worth a one-line normalization or a test that pins the decision.

Null/missing `actor`, null/missing `issue`, and unknown extra fields are all handled gracefully (zero-value fallback → `reason=content`, no panic) — confirmed.

---

## Coverage notes / what a human should double-check

- **Client renderer is the real blast-radius dial for the P1.** Whether `<script>`/autolinks in a title escalate from "formatting spoof" to "clickable phishing" depends on octo-web / octo-ios / octo-android markdown rendering, which isn't in this repo. A strict CommonMark client makes `<script>` inert text; a lenient/HTML-passing client escalates it. Worth a human confirming the client behavior.
- **HMAC deferral** (`X-Multica-Signature-256` not verified) is exact parity with the github/wecom/native adapters (none verify a body signature; the 256-bit URL token is the auth secret, constant-time compared before the body is read). Acceptable as a conscious deferral, not a blocker — but it's the reason the title vector is reachable, so the P1 fix matters more in its absence.
- **8 KiB body cap** is correct for the compact Multica envelope (no embedded repository object like GitHub's), and the adapter shares the same auth → rate-limit → audit pipeline as the other three, so no new DoS/amplification surface.
- This review is my own analysis plus one independent cross-model pass that agreed on both findings; a second long-context advisory pass timed out and did not return (treated as absent, not as agreement).

lml2468 (commented):
作者回应（我 @lml2468 是本 PR 作者，**回避投票**，但这是我代码里的真实 blocking，认领 + 确认修法）：

@yujiawei 的 🔴 CR **完全正确，两个 blocking 都成立**，我核实如下：

**根因**：title 用的是 `mdLinkText`，但 title **不在 `[text](url)` 链接里**——它是裸文本拼在 `**id** ⟨title⟩: ` `status` ` 之间（纯文本 + 紧邻 code span 的上下文）。而 `mdLinkTextEscaper` 只转义 `\ [ ]`，**不碰 `*` `_` 反引号**。用错了 helper。

**🔴 注入 1：title 的 `*` / `_` 未转义** → 上下文是 `**id** TITLE: ...`，title=`**x**` 之类会注入强调 / 破坏粗体结构。应转义 `*` `_`（mdLinkText 不转）。

**🔴 注入 2：title 的反引号破坏 status 的 code span 平衡** → title 里一个 `` ` `` 会和后面 `` `status` `` 的开反引号配对失衡：title=`` foo` `` 渲染成 `**id** foo`: `todo`` ——`: ` 被吞进 code span，整条消息结构错乱。mdLinkText 完全不处理反引号。

**修法**：title 改用 **`mdInertText`**（它转义 `* _ [ ] < > |` + 剥离反引号），而不是 `mdLinkText`——因为当前 envelope 不拼 issue URL，title 是纯文本字段不是链接文本。等未来真加 issue.url 拼成 `[title](url)` 时，再换回 mdLinkText（那时它在链接上下文里才正确）。

**我承认这是我自己的疏漏**：上一版 fix 给 id/status/actor 都按上下文接对了 helper（inert / codespan），**唯独最显眼的 title 漏改、还留着旧的 mdLinkText**——正是「同类实例只修一部分」又犯一次，且漏的是外部输入的标题这个主注入面。@yujiawei 升 🔴 准确。

第三点（补 title 注入测试）我一并加：title 含 `*`/`**`/反引号的用例，钉死渲染不破坏粗体、不破坏 status code span 平衡。

我作为作者来改这一版（title→mdInertText + 测试）。reviewDecision=CHANGES_REQUESTED 正确，应改后重审。

Jerry-Xin (dismissed):
The PR is in scope for `octo-server`: it extends the existing incoming webhook module with a new platform adapter and follows the shared push/auth/rate-limit/audit pipeline.

💬 Non-blocking

🟡 Warning — `modules/incomingwebhook/adapter_multica.go:158`: `title` is rendered in plain markdown text but escaped with `mdLinkText`, which only protects link-text delimiters (`[`, `]`, `\`). Since the title is not inside a markdown link here, a Multica issue title can still introduce autolinks or other markdown syntax such as `<https://...>`. This is not a blocker because webhook content is markdown by contract and token-gated, but it is inconsistent with the stronger `mdInertText` treatment used for `identifier` and `actor.type`.

🔵 Suggestion — `modules/incomingwebhook/README.md:293-295`: the audit docs are now stale. They omit `no_event` from the reason list and omit `multica` from the adapter list, even though this PR records both for the new adapter path.

🔵 Suggestion — `pkg/errcode/incomingwebhook.go:44-49`: the `no_event` comment is still GitHub-specific. With Multica, the same reason also covers a missing body `event`, so this should be generalized when convenient.

✅ Highlights

The new route is correctly wired through the existing `handlePush` path, so token auth, group checks, rate limiting, audit recording, and delivery semantics are reused rather than duplicated.

The parser has focused coverage for valid render, skipped events, malformed JSON, missing required fields, public URL generation, adapter registration, and markdown escaping edge cases.

Verification run: `go test ./modules/incomingwebhook/ -run 'TestParseMulticaPush|TestMulticaAdapter|TestPublicURLs'` passed.

lml2468:
Pushed `095f267` addressing @yujiawei's P1 + P2-2.

### P1 fix: title now uses `mdInertText`
Confirmed all four vectors:
- `**hijack**` → `\*\*hijack\*\*` (no real bold)
- `` a `b `` → backtick stripped (balanced-backtick invariant holds)
- `<script>alert(1)</script>` → `\<script\>\</script\>`
- `a | b` → `a \| b`

The github adapter's `mdLinkText` use is correct *for it* because its titles are wrapped in real `[text](url)` markdown links — the multica title path doesn't wrap, so `mdInertText` (which already protects `identifier` / `actor.type` in the same render) is the right helper. New `TestParseMulticaPush_TitleInjectionVectors` table-test covers all four vectors plus the cross-field balanced-backtick invariant (the title path can no longer reproduce the odd-backtick state that breaks status code-span rendering).

### P2-2 fix: event case-insensitive matching
`event` now goes through `strings.ToLower(strings.TrimSpace(...))` for parity with `msg_type` handling in `api.go`. New `TestParseMulticaPush_EventCaseInsensitive` pins three variants: `Issue.Status_Changed`, `ISSUE.STATUS_CHANGED`, leading/trailing space.

### P2-1 (strict scalar typing): deferred
The contract between multica and octo is the multica outbound dispatcher's `outboundPayload` struct (companion PR `Mininglamp-OSS/multica#39`), where `actor.id` is already declared as `string`. Pinning the type via a test on this side would just re-state that contract; the cross-repo guarantee is what protects against drift, and that's where the change should land if it ever happens.

Local: all multica/adapter/i18n tests green; balanced-backtick invariant now holds across both title and status fields.

lml2468 (commented):
作者澄清一个 reviewer 间的严重性分歧（我 @lml2468 是作者，**回避投票**，仅就证据说明，不替自己的 PR 推进）：

同一 head `46d14fa` 上，**@Jerry-Xin 把 title 转义列为 🟡（APPROVE），@yujiawei 把同一问题列为 🔴（CHANGES_REQUESTED）**——两位对同一处判断相反。我上一轮已独立核过代码，**这处分歧里 yujiawei 的 🔴 更准确**，理由（可复现）：

title 当前用 `mdLinkText`，而 `mdLinkTextEscaper` **只转义 `\ [ ]`**，不碰 `*` `_` 反引号。title 又是裸文本拼在 `**id** ⟨title⟩: ` `status` ` 里，于是有两个**功能性破坏**（不止 Jerry-Xin 说的 autolink 风险）：

1. **status 的 code span 失衡**：title 含一个 `` ` `` → 与后面 `` `status` `` 的开反引号配对，把 `: ` 吞进 code span，**整条消息结构错乱**。这不是「不美观」，是渲染破坏。
2. **强调注入**：title 含 `*`/`_` 注入粗体/斜体。

这两点都源于「title 是外部输入（multica issue 标题）+ 用了不转义 `*`/反引号 的 helper」。所以我同意 yujiawei：**应作为 blocking 在合并前修**（title 改 `mdInertText`），而不是 🟡 放行。reviewDecision 现在是 CHANGES_REQUESTED（yujiawei 的 CR 压住），是正确状态——**别因为有一个 APPROVE 就当它非阻塞**。

我作为作者来修（title→mdInertText + 注入测试），不是来主张放行。修完请两位 reviewer 重审对齐严重性。

Jerry-Xin (dismissed):
Summary: The PR is in scope for `octo-server` and cleanly extends the existing `incomingwebhook` adapter pipeline without introducing blocking correctness, security, or architecture issues.

💬 Non-blocking
- 🔵 Suggestion: Update stale self-documenting comments so future readers see `multica` in the adapter list. `modules/incomingwebhook/model.go:138` still says `native / github / wecom`, and `modules/incomingwebhook/sql/20260610000001_incomingwebhook_adapter_comment.sql:14` still documents the DB comment as `native/test/github/wecom`. This is not a runtime issue because the column is `VARCHAR(16)` and `multica` fits.

✅ Highlights
- The new route reuses the existing push pipeline, including token auth, group state checks, per-webhook limiting, delivery, and audit handling in `modules/incomingwebhook/api.go`.
- The Multica parser uses a narrow envelope whitelist and has clear skip vs invalid semantics in `modules/incomingwebhook/adapter_multica.go:73`.
- Markdown rendering handles the risky fields by context: inert text for identifier/title/actor and code-span stripping for statuses in `modules/incomingwebhook/adapter_multica.go:148`.
- Tests cover parsing, escaping, unknown events, invalid payloads, route auth, delivery path, and public URL exposure.

Verification: I ran `go test ./modules/incomingwebhook -run 'TestParseMulticaPush|TestMulticaAdapter|TestPublicURLs'`; it passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#427 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/427
**Head SHA:** `095f267b685d6c70f41d02def6b3d8fbbf154838`
**Title:** feat(incomingwebhook): add /Octo-Q adapter for Octo-Q outbound webhooks

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Route registration parity | ✅ | `api.go:230` — `/Octo-Q` uses identical `chain()` middleware as `/github`, `/wecom` |
| Token auth (constant-time) | ✅ | `handlePush` step 2 (`api.go:1126`) — `subtle.ConstantTimeCompare` shared by all adapters |
| Body limit enforcement | ✅ | `multicaAdapter.bodyLimit = maxBytes` (8 KiB) — `api.go:1202` reads `limit+1` for overflow detection |
| Adapter parse contract | ✅ | `adapter_multica.go:72-113` — three-return-value contract (`req`/`skip`/`invalid`) matches `pushAdapter.parse` signature |
| `no_event` reason wiring | ✅ | errcode `SafeDetailKeys:["reason"]` (`pkg/errcode/incomingwebhook.go:52`), SQL migration `20260618000001`, i18n test `api_i18n_test.go:129` |
| Markdown escaping (all fields) | ✅ | `adapter_multica.go:148-175` — identifier→`mdInertText`, title→`mdInertText`, status→`mdCodeSpanText`, actor.type→`mdInertText` |
| `publicURLs` includes Octo-Q | ✅ | `api.go:403` — `"Octo-Q": base + "/Octo-Q"` |
| Audit adapter constant | ✅ | `model.go:53` — `adapterMultica = "Octo-Q"` declared alongside existing constants |
| Test coverage (12 unit + 3 e2e) | ✅ | `adapter_multica_test.go` (301 lines) + `adapter_push_test.go` additions |
| Access-log token scrubbing | ✅ | `accesslog.go:47` — `ScrubPath` masks everything after webhook_id, including `/Octo-Q` suffix |
| i18n legacy-response ban | ✅ | `api_i18n_test.go:23` — `adapter_multica.go` added to banned-files list |

## 2. Findings

### P2-1: `pushPayloadInvalid` doc comment omits `no_event` from reason enumeration

**Severity:** P2 (documentation inconsistency, zero runtime impact)
**Diff-scope:** Pre-existing — the comment at `api_i18n.go:41` was not updated when `no_event` was added by an earlier PR (#330 era). This PR correctly produces `"no_event"` from `parseMulticaPush` but does not touch the stale comment.
**Location:** `modules/incomingwebhook/api_i18n.go:41` — doc comment lists `reason in {body, json, content, blocks, msg_type, event}` but omits `no_event`. The authoritative comment at `pkg/errcode/incomingwebhook.go:46-49` correctly includes it.
**Suggestion:** Add `no_event` to the enumeration in the `pushPayloadInvalid` doc comment.

### P2-2: No dedicated unit test for `mdInertTextEscaper` / `mdCodeSpanStripper` with backslash inputs

**Severity:** P2 (test coverage gap, not a correctness bug)
**Diff-scope:** New — `mdInertTextEscaper` and `mdCodeSpanText` are introduced by this PR.
**Location:** `adapter.go:150-182` — the `\` → `\\` replacement path is exercised only indirectly through field-level tests. No test input contains a literal backslash character.
**Analysis:** `strings.NewReplacer` processes input left-to-right without re-scanning output, so placing `\` first is correct and cannot cause double-escaping. Traced inputs `\`, `\*`, `\\`, `\[` all produce correct markdown output. The gap is coverage, not correctness.
**Suggestion:** Add a `TestMdInertTextEscaper` with inputs `\`, `\*`, `\\`, `\\*` to pin the backslash-escaping path.

### P2-3: Access-log test lacks explicit adapter-suffix path case

**Severity:** P2 (test coverage gap)
**Diff-scope:** Pre-existing — `accesslog_test.go` never tested adapter-suffix paths for any adapter (github/wecom/Octo-Q).
**Location:** `pkg/accesslog/accesslog_test.go` — no test case for `/v1/incoming-webhooks/iwh_abc/token/Octo-Q`.
**Analysis:** `ScrubPath` handles it correctly by design (everything after the first `/` in the rest portion is masked). Not a regression risk for this PR.
**Suggestion:** Add one test case with an adapter-suffix path to the existing accesslog test table.

## 3. Suggestions

1. **P2-1 fix:** One-line comment update in `api_i18n.go:41` to add `no_event` to the reason enumeration.
2. **P2-2 fix:** Add `TestMdInertTextEscaper` with backslash-containing inputs (~15 lines).
3. **P2-3 fix:** Add one row to the accesslog test table for an adapter-suffix path (~5 lines).

## 4. Additional Observations

- **HMAC verification deferred:** `X-Octo-Q-Signature-256` is documented but not verified — symmetric with `X-Hub-Signature-256` on the GitHub adapter. Both are explicitly scoped as "future optional." Consistent and acceptable.
- **`ghLogin` / `ghWithRepo` do not escape** (pre-existing, not introduced by this PR): GitHub adapter places `sender.login` inside `**%s**` without `mdInertText`. Safe in practice due to GitHub's username restrictions (alphanumeric + hyphens), but a defense-in-depth consideration for future GHE scenarios.
- **Event name normalization** (`ToLower` + `TrimSpace`): Unique to Octo-Q adapter (GitHub uses header value as-is). Good defense-in-depth against case variants being silently collapsed to `skip(event)`.

## 5. Data-Flow Tracing

| Consumed Data | Upstream Source | Reaches Consumer? | Evidence |
|---|---|---|---|
| `ev.Event` | `muEnvelope.Event` from JSON body | ✅ → `strings.ToLower(TrimSpace)` → switch dispatch | `adapter_multica.go:82-89` |
| `ev.Issue.Identifier` | `muEnvelope.Issue.Identifier` from JSON body | ✅ → `TrimSpace` → `mdInertText(id, 64)` → `**%s**` | `adapter_multica.go:148,155` |
| `ev.Issue.Title` | `muEnvelope.Issue.Title` from JSON body | ✅ → `TrimSpace` → `mdInertText(title, 200)` → bare text | `adapter_multica.go:149,160` |
| `ev.Issue.Status` | `muEnvelope.Issue.Status` from JSON body | ✅ → `TrimSpace` → `mdCodeSpanText(status, 64)` → `` `%s` `` | `adapter_multica.go:150,163-167` |
| `ev.PreviousStatus` | `muEnvelope.PreviousStatus` from JSON body | ✅ → `TrimSpace` → `mdCodeSpanText(prev, 64)` → `` `%s` `` | `adapter_multica.go:151,164` |
| `ev.Actor.Type` | `muEnvelope.Actor.Type` from JSON body | ✅ → `TrimSpace` → `mdInertText(actorType, 64)` → `(by %s)` | `adapter_multica.go:170-172` |
| `req.Content` (final) | `renderMulticaIssueStatusChanged` output | ✅ → `clipRunes(content, maxContentRunes())` → `pushPayloadReq{Content:...}` | `adapter_multica.go:113` |
| `pushPayloadReq.Content` | adapter parse return | ✅ → `handlePush` step 6 text path → `buildPayload` → `SendMessageWithResult` | `api.go:1239-1280` |

**No data-flow gaps found.** Every user-controlled field passes through the appropriate markdown sanitization helper before being interpolated into the output string. The whitelist struct (`muEnvelope`) ensures unknown JSON fields are silently ignored.

- Route registration: all four POST routes use identical `chain()` middleware (`api.go:220-232`).
- Handler dispatch: all four delegate to `handlePush(c, adapter)` with no divergence (`api.go:1074-1079`).
- `publicURLs`: includes `"Octo-Q"` key (`api.go:403`).
- Audit constants: `adapterMultica` declared alongside existing constants (`model.go:53`).
- Auth test: `TestPush_AdapterRoute_AuthEnforced` iterates `["github", "wecom", "Octo-Q"]` (`adapter_push_test.go:118`).
- No create↔delete asymmetry: this PR adds a push endpoint, not a management CRUD operation.

**C2 — Control-flow ordering / nested reuse:** ✅ Clear.
- `mdInertTextEscaper` replacement ordering: `\` first is correct (verified by tracing `\`, `\*`, `\\` inputs). `strings.NewReplacer` never re-scans output.
- `mdInertText` vs `mdCodeSpanText` context separation: correctly documented and enforced. Using `mdInertText` inside a code span would produce visible `\_` artifacts; using `mdCodeSpanText` outside would leave `*`/`[` unescaped. The PR's render function uses each in the correct context.
- No nested/double invocation of escaping helpers.

**C3 — Authorization boundary ≠ capability boundary:** ✅ Clear.
- `/Octo-Q` uses the same URL-token auth as all other push routes. No new authorization surface.
- No secondary mount points (bot_api only mounts management routes, not push routes).
- The adapter cannot bypass any middleware: `requirePushEnabled`, rate limits, IP failure gate, token auth, group status check, creator membership check — all applied identically.

**C4 — Authorization lifecycle / container-member status cascade:** ✅ N/A.
- This PR adds a push endpoint, not an auth/membership query. The existing `cachedRequireActiveGroup` and `cachedCreatorMembership` checks in `handlePush` already handle container-level status cascade (disabled group → 401, creator left → lazy disable).

**C5 — Build/note ≠ runtime path:** ✅ N/A.
- No build artifacts, browser extensions, CLI tools, or relative URL concerns. This is a server-side Go module addition.

**C6 — Governance/policy document consistency:** ✅ N/A.
- README.md additions are accurate and consistent with existing adapter documentation. No policy conflicts.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. Three P2 items (documentation comment gap, test coverage for backslash escaping, accesslog test case) — all non-blocking. The adapter is correctly integrated into the shared pipeline with full middleware parity, thorough markdown escaping (the most rigorous of any adapter), comprehensive test coverage (12 unit + 3 e2e tests), and proper `no_event` reason wiring across all layers.

Jerry-Xin (approved):
Re-affirming APPROVE on head 19d48dc7fa53.

The only delta vs the previously-verified head 095f267b is a doc/comment refresh (the 🔵 nit from the prior round), no code-logic change:
- `model.go:138` urls-list comment now reads `native / github / wecom / multica` (was missing `multica`). Consistent with the actual adapter set.
- New migration `20260622000001_..._multica_adapter_comment.sql`: refreshes only the `adapter` column COMMENT to enumerate `multica`, with proper Up/Down and an INSTANT-algorithm note for MySQL 8.0. No schema/data change.

`adapter_multica.go` is untouched in this delta, so the title-injection fix verified on 095f267b (title → mdInertText neutralizing `*`/`_`/backtick, with TestParseMulticaPush_TitleInjectionVectors pinning the four vectors incl. the even-backtick-count invariant) and the per-field escape-helper correctness all stand. No regression.

No blocking items.

Note on merge gate: reviewDecision is still CHANGES_REQUESTED / BLOCKED because yujiawei's CHANGES_REQUESTED on the older head 46d14fa3 remains active (CR is not auto-dismissed by new commits — only the reviewer can withdraw/re-vote). The P1/P2 concerns that CR raised are fixed in 095f267b and unchanged here. @yujiawei please re-review to clear the stale CR.

yujiawei (approved):
# Code Review — PR #427 (octo-server)

Reviewed at head SHA `095f267b685d6c70f41d02def6b3d8fbbf154838`. Base `main` (merge-base `d69ba8e6`). Build, `go vet`, and the full `modules/incomingwebhook` test suite were run locally against this exact head.

## Verdict: APPROVED

This PR adds a `/multica` push adapter that accepts Multica outbound webhook envelopes (`issue.status_changed`) and translates them into native push messages, reusing the existing auth / rate-limit / group-check / delivery / audit pipeline (`handlePush`). It is small, additive, well-documented, and carries strong test coverage. No blocking (P0/P1) issues were found. The items below are non-blocking nits/suggestions.

## Verification summary

| Area | Result | Evidence |
|---|---|---|
| Build | ✅ | `go build ./modules/incomingwebhook/` clean |
| Vet | ✅ | `go vet ./modules/incomingwebhook/` clean |
| Tests | ✅ | full module suite passes (`ok ... 14.9s`); all new Multica unit + e2e tests green |
| Auth surface | ✅ | adapter only emits `pushPayloadReq{Content}`; token auth, 4-layer rate limit, group/creator-membership checks are fully shared and unchanged (`api.go` `handlePush`, dispatch at `api.go:1079`) |
| Markdown injection | ✅ | every rendered field routed through the correct escaper (see below) |
| Return contract | ✅ | `parseMulticaPush` obeys the exactly-one-of `(req / skip / invalid)` contract on all 5 paths |
| Reason routing | ✅ | `json` / `no_event` / `content` → 400; unknown event → 200 skip — matches `handlePush` (`api.go:1211-1223`) |

### Security (this PR is classified security-sensitive)

The interesting surface is markdown/content injection, since envelope fields (`title`, `identifier`, `status`, `actor.type`) are rendered into a markdown message. Each field is routed to a context-appropriate escaper, and the four classic vectors are closed and test-covered:

- **Bold-break** (`*`) and **link injection** (`[`/`]`): `identifier`/`title`/`actor.type` go through `mdInertText` (`adapter.go`), which escapes `\ * _ [ ] < > |` and strips backticks.
- **Code-span escape via backtick**: `status`/`previous_status` go through `mdCodeSpanText`, which strips backticks (correct — a single-backtick span can't safely backslash-escape a backtick). Critically, `title` also strips backticks via `mdInertText`, so the total backtick count stays even and a backtick-bearing title cannot split the downstream `status` code-span. This invariant is explicitly asserted in `adapter_multica_test.go` (`strings.Count(content, "`")%2 == 0`).
- **HTML/autolink** (`<`/`>`) and **table pipe** (`|`): escaped by `mdInertText`.
- **Clip-before-escape ordering**: `clipRunes` runs inside the escaper *before* the replacer, so truncation cannot split an escape sequence; `oneLine` collapses newlines, preventing line-start (heading/blockquote/list) injection.

Body size is bounded on raw bytes at `handlePush` step 4 (`api.go`, `len(body) > limit` → 413) *before* JSON parse, so a crafted long title cannot bypass the cap; the 8 KiB limit (vs GitHub's wider limit) is justified because the Multica envelope is compact (no embedded repository object). Output is further clamped (title 200 runes, short fields 64 runes, final `clipRunes(content, maxContentRunes())`).

**For a human verifier:** the `X-Multica-Signature-256` HMAC header is intentionally **not** verified (documented as a future option). This is acceptable here because the actual auth gate is the 128-bit URL token (constant-time compared), identical to the native/github/wecom adapters — the signature would be defense-in-depth, not the primary control. Worth confirming this matches the team's threat model for the outbound-webhook integration before relying on it for tamper-evidence.

## Findings (all non-blocking)

**P2-1 — `mdInertText` does not escape GFM strikethrough `~`.** `adapter.go` (`mdInertTextEscaper`). Cosmetic only: a `~~text~~` title could render struck-through. Cannot escalate to link/HTML/code-span injection since those vectors are all escaped. Optional to add for completeness.

**P2-2 — Long-title clip test only uses non-escaping chars.** `adapter_multica_test.go` (`TestParseMulticaPush_LongTitleIsClipped`) repeats `"a"`, so its `<=260` bound would not catch escape-expansion (e.g. 200 `*` → ~400 runes after escaping). Real output is still bounded by the final `clipRunes(content, maxContentRunes())`, so this is a test-coverage gap, not a code defect. Consider a case with escaping-heavy input.

**P2-3 — Untested branch leg: `previous_status` present but equal to `status`.** `renderMulticaIssueStatusChanged` (`if prev != "" && prev != status`). The `prev == status` short-circuit (no arrow) is never directly asserted; only the `prev == ""` path exercises the same else-branch. Trivial to add.

**P2-4 — Deliberate, documented divergence from the github adapter.** Two intentional differences, both correct and pinned by tests; flagged only for reviewer awareness:
- Missing event field → `no_event` (multica) vs github's missing-header → `event`. `no_event` is a **pre-existing**, already-i18n/DB-registered reason code, so the adapter's "no new registration needed" claim is accurate, and it gives ops a cleaner "config error vs event-not-in-subset" signal.
- Known event that renders empty (missing `identifier`/`status`) → 400 `content` (multica) vs github's 200 skip. Semantically correct here: empty render only happens on a malformed payload, not a deliberate "chose not to render," so 400 is the right signal.

## Note on review coverage

The Multica-side outbound payload schema (`outboundPayload` in the multica repo) is not in this checkout, so the exact JSON key names (`event` / `issue.identifier` / `issue.status` / `previous_status` / `actor.type`) consumed by the whitelist structs could not be cross-verified against the real sender. The whitelist parse is the safe direction (unknown fields ignored), but if a sender key differs (e.g. `prev_status` vs `previous_status`), the arrow would silently never render — worth a one-time fixture check against an actual Multica delivery.

PR #435, opened 2026-06-23
fix(messages_search): strip Chinese stopwords + MSM 75% to kill multi…

BrDing-Rookie:
…_match OR-trap

Reader-layer fix for the OR-trap in /_search_messages, /_search_all, /_search_files.

Symptom: ik_smart tokenizes "按时缴纳时间的女性" into [按时,缴纳,时间,的,女性]; multi_match defaults to operator=OR so any single token (especially the high-frequency function word 的) matches and returns ~10k hits with highlights all landing on 的.

Fix (per docs/messages-search/2026-06-23-multimatch-or-trap-fix.md §4.1):
- AnalyzeKeyword → OS _analyze (ik_smart) + filterStopwords
- buildKeywordClauseFromAnalyzed → multi_match(cross_fields) + MinimumShouldMatch("75%") for content tokens; raw keyword + no MSM when every token is a stopword (preserves "search literally for 的" semantics)
- _analyze failure degrades to §4.4 (raw keyword + cross_fields + MSM 75%)
- search_all shares one _analyze across its two Should branches (2→1 RT)
- SearchConfig.StopwordStripEnabled flag (env OCTO_SEARCH_STOPWORD_STRIP_ENABLED, default true) kill-switches back to §4.4 without redeploy

Affects 4 call sites: search_messages.go, search_all.go (text+file), search_files.go. search_media.go (no keyword) and search_around.go (msgId anchor) unaffected.

Tests: keyword_query_test.go + stopwords_test.go cover spec §5 unit table plus calls==0 flag-off and calls==1 _search_all reuse assertions. Existing dsl_test.go / space_filter_test.go signature-migrated via fallbackTestAnalyzer preserving historical literal pins. e2e ("按时缴纳时间的女性" recall delta) left for make env-test per spec §6.

## Changes

-
-

## Testing

- [ ] Unit tests added/updated
- [ ] Manually verified

## COMPREHENSION

## Checklist

- [ ] PR description is in English
- [ ] Added tests for my changes
- [ ] Updated documentation
- [ ] Followed commit message conventions (Conventional Commits)

### DIFF
--- a/modules/messages_search/keyword_query.go
+++ b/modules/messages_search/keyword_query.go
@@ -1,0 +1,123 @@
+package messages_search
+
+import (
+	"context"
+	"strings"
+
+	"github.com/olivere/elastic"
+)
+
+// tokenAnalyzer abstracts the IK tokenization step so the keyword-clause
+// builders can be unit-tested without standing up an OpenSearch cluster.
+// Production wires in an osIKSmartAnalyzer that calls the cluster's
+// `_analyze` endpoint; tests provide a stub returning a fixed token slice
+// (or an error to exercise the fallback path).
+type tokenAnalyzer interface {
+	Analyze(ctx context.Context, text string) ([]string, error)
+}
+
+// osIKSmartAnalyzer drives OS `_analyze?analyzer=ik_smart`. The endpoint is
+// index-independent (we pass no Index()), so it works even when the
+// configured read alias is empty/down — the IK plugin lives at the cluster
+// level.
+type osIKSmartAnalyzer struct {
+	client *elastic.Client
+}
+
+func newOSIKSmartAnalyzer(c *elastic.Client) osIKSmartAnalyzer {
+	return osIKSmartAnalyzer{client: c}
+}
+
+// Analyze returns the IK-smart token sequence for `text`, dropping empty
+// tokens defensively.
+func (a osIKSmartAnalyzer) Analyze(ctx context.Context, text string) ([]string, error) {
+	resp, err := a.client.IndexAnalyze().
+		Analyzer("ik_smart").
+		Text(text).
+		Do(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if resp == nil {
+		return nil, nil
+	}
+	tokens := make([]string, 0, len(resp.Tokens))
+	for _, t := range resp.Tokens {
+		if t.Token == "" {
+			continue
+		}
+		tokens = append(tokens, t.Token)
+	}
+	return tokens, nil
+}
+
+// AnalyzeKeyword runs IK-smart segmentation + stopword filtering against
+// `keyword` exactly once, returning the inputs needed by one or more
+// buildKeywordClauseFromAnalyzed calls. This lets a handler that issues
+// multiple multi_match clauses against the same keyword (e.g. _search_all
+// with separate text/file Should branches) reuse one `_analyze` roundtrip.
+//
+// Return contract per docs/messages-search/2026-06-23-multimatch-or-trap-fix.md §4.1:
+//
+//   - err != nil  → _analyze failed. effectiveKeyword == keyword, useMSM == true.
+//     Feeding this pair to buildKeywordClauseFromAnalyzed yields the §4.4
+//     degraded shape (raw keyword + cross_fields + MSM 75%). Caller may
+//     warn-log the error; the returned values are still safe to use.
+//   - post-strip token slice empty (pure stopwords like "的" / "的了"):
+//     effectiveKeyword == keyword (the original, unsplit string), useMSM == false.
+//     Builder emits raw keyword + cross_fields, no MSM — preserves the "user
+//     literally searches for a function word" semantic.
+//   - content tokens remain: effectiveKeyword == content tokens joined with
+//     spaces, useMSM == true. Builder emits the joined string + cross_fields +
+//     MSM 75%. Stopwords drop out of the MSM denominator so a 5-token query
+//     with one stopword becomes a 4-of-4 (75% ≈ 3) test on content words.
+func AnalyzeKeyword(ctx context.Context, a tokenAnalyzer, keyword string) (effectiveKeyword string, useMSM bool, err error) {
+	tokens, err := a.Analyze(ctx, keyword)
+	if err != nil {
+		return keyword, true, err
+	}
+	content := filterStopwords(tokens, defaultStopwords)
+	if len(content) == 0 {
+		return keyword, false, nil
+	}
+	return strings.Join(content, " "), true, nil
+}
+
+// buildKeywordClauseFromAnalyzed constructs the multi_match clause from a
+// previously-computed (effectiveKeyword, useMSM) pair — see AnalyzeKeyword.
+// Triggers no _analyze of its own, so a caller can issue multiple
+// field-specific clauses (text vs file) from a single analyze.
+//
+// type=cross_fields is required: under best_fields, MSM is evaluated per
+// field, and a token landing in any single field clears the bar, defeating
+// the threshold for multi-field corpora (text vs file.name etc).
+func buildKeywordClauseFromAnalyzed(effectiveKeyword string, useMSM bool, fields ...string) elastic.Query {
+	q := elastic.NewMultiMatchQuery(effectiveKeyword, fields...).Type("cross_fields")
+	if useMSM {
+		q = q.MinimumShouldMatch("75%")
+	}
+	return q
+}
+
+// buildKeywordClause is the thin wrapper preserved for callers that issue
+// exactly one multi_match per keyword (search_messages, search_files). The
+// signature, branch semantics, and returned error match the pre-split
+// implementation — see AnalyzeKeyword for the branch table.
+func buildKeywordClause(ctx context.Context, a tokenAnalyzer, keyword string, fields ...string) (elastic.Query, error) {
+	eff, useMSM, err := AnalyzeKeyword(ctx, a, keyword)
+	return buildKeywordClauseFromAnalyzed(eff, useMSM, fields...), err
+}
+
+// buildKeywordClauseGated is the flag-aware front door used by
+// buildSearchMessagesDSL / buildSearchFilesDSL.
+//
+// When stopwordStripEnabled is false, the `_analyze` call is skipped entirely
+// (no IK roundtrip RT) and the result degrades to the §4.4 shape — raw
+// keyword + cross_fields + MSM 75%. This is the ops kill switch for the
+// stopword-strip pipeline; see SearchConfig.StopwordStripEnabled.
+func buildKeywordClauseGated(ctx context.Context, a tokenAnalyzer, stopwordStripEnabled bool, keyword string, fields ...string) (elastic.Query, error) {
+	if !stopwordStripEnabled {
+		return buildKeywordClauseFromAnalyzed(keyword, true, fields...), nil
+	}
+	return buildKeywordClause(ctx, a, keyword, fields...)
+}
--- a/modules/messages_search/keyword_query_test.go
+++ b/modules/messages_search/keyword_query_test.go
@@ -1,0 +1,400 @@
+package messages_search
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"strings"
+	"testing"
+)
+
+// stubAnalyzer feeds buildKeywordClause a canned token list (or error) so the
+// keyword-DSL shape can be asserted without an OpenSearch cluster. Tests pass
+// either tokens (happy path) or err (fallback path); both can be set on the
+// same stub but tokens are ignored when err is non-nil — same as the real
+// IndexAnalyze().Do() contract.
+type stubAnalyzer struct {
+	tokens []string
+	err    error
+}
+
+func (s stubAnalyzer) Analyze(_ context.Context, _ string) ([]string, error) {
+	if s.err != nil {
+		return nil, s.err
+	}
+	return s.tokens, nil
+}
+
+// countingAnalyzer wraps a stubAnalyzer with a call counter. Used by the
+// flag-off tests to verify the `_analyze` roundtrip is actually skipped
+// (a query-shape assertion alone can't distinguish "analyze ran, fell back
+// to raw" from "analyze never ran").
+type countingAnalyzer struct {
+	inner stubAnalyzer
+	calls int
+}
+
+func (c *countingAnalyzer) Analyze(ctx context.Context, text string) ([]string, error) {
+	c.calls++
+	return c.inner.Analyze(ctx, text)
+}
+
+// keywordClauseBody marshals a buildKeywordClause result to JSON so substring
+// assertions stay legible. The wrapper returned by olivere matches the OS
+// wire shape ({ "multi_match": { ... } }), which is what we want to pin.
+func keywordClauseBody(t *testing.T, q interface {
+	Source() (any, error)
+}) string {
+	t.Helper()
+	src, err := q.Source()
+	if err != nil {
+		t.Fatalf("Source(): %v", err)
+	}
+	b, err := json.Marshal(src)
+	if err != nil {
+		t.Fatalf("marshal: %v", err)
+	}
+	return string(b)
+}
+
+func TestBuildKeywordClause_PureStopword_SingleToken(t *testing.T) {
+	// "的" alone — user wants literal matches of the function word.
+	// Expected: original keyword, cross_fields, NO MSM.
+	a := stubAnalyzer{tokens: []string{"的"}}
+	q, err := buildKeywordClause(context.Background(), a, "的", "payload.text.content")
+	if err != nil {
+		t.Fatalf("unexpected analyze error: %v", err)
+	}
+	body := keywordClauseBody(t, q.(interface {
+		Source() (any, error)
+	}))
+	if !strings.Contains(body, `"query":"的"`) {
+		t.Errorf("pure stopword path must use the original keyword:\n%s", body)
+	}
+	if !strings.Contains(body, `"type":"cross_fields"`) {
+		t.Errorf("missing cross_fields type:\n%s", body)
+	}
+	if strings.Contains(body, "minimum_should_match") {
+		t.Errorf("pure stopword path must NOT set MSM (literal match semantic):\n%s", body)
+	}
+}
+
+func TestBuildKeywordClause_PureStopword_MultiCharToken(t *testing.T) {
+	// IK smart segments "的话" as a single token; it's in the stopword set
+	// and after filtering nothing remains → pure-stopword path.
+	a := stubAnalyzer{tokens: []string{"的话"}}
+	q, err := buildKeywordClause(context.Background(), a, "的话", "payload.text.content")
+	if err != nil {
+		t.Fatalf("unexpected analyze error: %v", err)
+	}
+	body := keywordClauseBody(t, q.(interface {
+		Source() (any, error)
+	}))
+	if !strings.Contains(body, `"query":"的话"`) {
+		t.Errorf("expected original keyword '的话':\n%s", body)
+	}
+	if strings.Contains(body, "minimum_should_match") {
+		t.Errorf("pure stopword must skip MSM:\n%s", body)
+	}
+}
+
+func TestBuildKeywordClause_PureStopword_AllStrippedAway(t *testing.T) {
+	// "的了" → ["的","了"] → after filter both gone. Must fall back to the
+	// original (unsplit) keyword so the user still gets literal hits.
+	a := stubAnalyzer{tokens: []string{"的", "了"}}
+	q, err := buildKeywordClause(context.Background(), a, "的了", "payload.text.content")
+	if err != nil {
+		t.Fatalf("unexpected analyze error: %v", err)
+	}
+	body := keywordClauseBody(t, q.(interface {
+		Source() (any, error)
+	}))
+	if !strings.Contains(body, `"query":"的了"`) {
+		t.Errorf("post-strip empty must use the ORIGINAL keyword, not the join:\n%s", body)
+	}
+	if strings.Contains(body, "minimum_should_match") {
+		t.Errorf("post-strip empty must NOT set MSM:\n%s", body)
+	}
+}
+
+func TestBuildKeywordClause_ContentWithStopword_OrTrapFix(t *testing.T) {
+	// The reproducer from the spec: 5-token query with a single stopword. The
+	// "的" must be dropped from the issued query so MSM 75% applies to the 4
+	// content words.
+	a := stubAnalyzer{tokens: []string{"按时", "缴纳", "时间", "的", "女性"}}
+	q, err := buildKeywordClause(context.Background(), a, "按时缴纳时间的女性",
+		"payload.text.content^3", "payload.mergeForward.msgs.searchText",
+	)
+	if err != nil {
+		t.Fatalf("unexpected analyze error: %v", err)
+	}
+	body := keywordClauseBody(t, q.(interface {
+		Source() (any, error)
+	}))
+	if !strings.Contains(body, `"query":"按时 缴纳 时间 女性"`) {
+		t.Errorf("filtered keyword must be the space-joined content tokens, no '的':\n%s", body)
+	}
+	if !strings.Contains(body, `"type":"cross_fields"`) {
+		t.Errorf("missing cross_fields type:\n%s", body)
+	}
+	if !strings.Contains(body, `"minimum_should_match":"75%"`) {
+		t.Errorf("missing MSM 75%%:\n%s", body)
+	}
+}
+
+func TestBuildKeywordClause_ShortMixed(t *testing.T) {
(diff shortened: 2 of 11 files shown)


modules/messages_search/keyword_query.go
```diff
@@ -0,0 +1,123 @@
...
+	tokens, err := a.Analyze(ctx, keyword)
+	if err != nil {
+		return keyword, true, err
+	}
+	content := filterStopwords(tokens, defaultStopwords)
+	if len(content) == 0 {
+		return keyword, false, nil
+	}
+	return strings.Join(content, " "), true, nil
+}
```
OctoBoooot, line 84: praise: splitting `AnalyzeKeyword` from `buildKeywordClauseFromAnalyzed` (instead of keeping one monolithic `buildKeywordClause`) is the right shape for `_search_all`'s two-Should-branch case — and `TestBuildSearchAllDSL_OneAnalyzePerKeyword` + `TestBuildKeywordClauseFromAnalyzed_OneAnalyzeManyClauses` pin the 1-RT-not-2 invariant via a `countingAnalyzer` that asserts on `a.calls == 1`. That's stronger than the typical "query shape looks right" pin — a future refactor that adds a redundant analyze call to `buildSearchAllDSL` would fail the count assertion, not slip through. The three-branch contract in the docstring (`err != nil` / pure-stopword / content-tokens) makes the return-tuple semantics legible at the call site without re-reading the helper.

modules/messages_search/keyword_query.go
```diff
@@ -0,0 +1,123 @@
...
+//
+// When stopwordStripEnabled is false, the `_analyze` call is skipped entirely
+// (no IK roundtrip RT) and the result degrades to the §4.4 shape — raw
+// keyword + cross_fields + MSM 75%. This is the ops kill switch for the
+// stopword-strip pipeline; see SearchConfig.StopwordStripEnabled.
+func buildKeywordClauseGated(ctx context.Context, a tokenAnalyzer, stopwordStripEnabled bool, keyword string, fields ...string) (elastic.Query, error) {
+	if !stopwordStripEnabled {
+		return buildKeywordClauseFromAnalyzed(keyword, true, fields...), nil
+	}
+	return buildKeywordClause(ctx, a, keyword, fields...)
```
OctoBoooot, line 122: praise: the `buildKeywordClauseGated` kill switch is wired through to `OCTO_SEARCH_STOPWORD_STRIP_ENABLED` (defaulting ON, see `config.go:96`). Two things this does well: (a) flag-off path doesn't just skip the strip — it skips the `_analyze` roundtrip entirely (line 119-121 returns before the OS call), so the rollback isn't "strip is bypassed but you still pay the RT"; and (b) `TestBuildKeywordClauseGated_FlagOffSkipsAnalyze` pins `a.calls == 0` to lock in the no-RT invariant. Ops can flip one env var and the system instantly reverts to the §4.4 degraded shape with no redeploy and no latency overhead from a vestigial analyzer call.

modules/messages_search/keyword_query_test.go
```diff
@@ -0,0 +1,400 @@
...
+	if !strings.Contains(body, `"query":"的话"`) {
+		t.Errorf("expected original keyword '的话':\n%s", body)
+	}
+	if strings.Contains(body, "minimum_should_match") {
+		t.Errorf("pure stopword must skip MSM:\n%s", body)
+	}
+}
+
+func TestBuildKeywordClause_PureStopword_AllStrippedAway(t *testing.T) {
+	// "的了" → ["的","了"] → after filter both gone. Must fall back to the
```
OctoBoooot, line 102: praise: the three pure-stopword tests cover the documented design choice ("search literally for 的" semantic) at three different surfaces — single token (`的`), multi-char single token (`的话`), and the all-stripped-from-multiple case (`的了` → `[的, 了]` → both filtered → fall back to ORIGINAL unsplit keyword `的了`, not `""` from `strings.Join([])`). The last one (`...AllStrippedAway`) is the subtle one: a naive implementation would happily issue `multi_match(query="")` and ship something unexpected to OS. Test pins the right behaviour at the seam.

modules/messages_search/search_all.go
```diff
@@ -132,20 +136,32 @@ func buildSearchAllDSL(req SearchAllReq, normChannelID, spaceID string) elastic.
...
+			// Single `_analyze` roundtrip feeds both Should branches — same
+			// keyword, same analyzer, so a second call would be redundant and
+			// double the IK-cluster RT cost of every _search_all hit.
+			var err error
+			eff, useMSM, err = AnalyzeKeyword(ctx, analyzer, req.Keyword)
+			analyzeErr = err
+		}
+		// When stopwordStripEnabled=false: skip _analyze entirely and emit the
+		// §4.4 degraded shape (raw keyword + cross_fields + MSM 75%) on both
+		// branches — the eff/useMSM defaults above match that contract.
```
OctoBoooot, line 152: praise: the comment block at lines 142-152 makes the eff/useMSM default-vs-overridden trick legible (`eff, useMSM := req.Keyword, true` → matches §4.4 fallback shape exactly; AnalyzeKeyword only overrides when the flag is on AND the call succeeds). A future contributor adding a third Should branch can copy the same `buildKeywordClauseFromAnalyzed(eff, useMSM, ...)` line without re-deriving the contract.

OctoBoooot (commented):
## Review: fix(messages_search): strip Chinese stopwords + MSM 75% to kill multi_match OR-trap (#435)

**Verdict:** Comment — CI is red on `check-sprint / check-sprint` (process gate — the PR isn't assigned to a sprint on the project board), so per rule §11 I can't flip to Approve until that clears. On the code itself I have no blockers and no majors; the substantive Build/Vet/Lint/Test checks are still pending in this run.

Size: 11 files / 999 lines → `size:large` label applied. Risk tier: **normal** (search/OS query construction, no auth/migrations/secrets/PII paths).

### Adversarial verification pass
- **Call-site coverage matches the PR claim** ("4 call sites: search_messages.go, search_all.go (text+file), search_files.go"): verified via grep — `buildKeywordClauseGated` at `search_messages.go:146` and `search_files.go:144`; `AnalyzeKeyword` + two `buildKeywordClauseFromAnalyzed` at `search_all.go:147,153,157`. The "unaffected" claim also checks: `search_media.go` immediately rejects any keyword via `validateKeywordMustBeEmpty`, and `search_around.go` is the no-keyword window query.
- **MSM 75% boundary inputs constructed and verified**: 1 content token → ceil(1·0.75)=1, trivially satisfied (matches docstring); 2 tokens → 2-of-2 required (more restrictive than naive OR, intentional per PR design); 5 tokens with 1 stopword → 4 content tokens → ceil(4·0.75)=3, reasonable.
- **Pure-stopword path is the documented design choice** (PR body §4.1 / `keyword_query.go:66-69`): user typing `的` gets OR-trap behaviour back because they "literally asked for the stopword." Debatable UX but explicitly chosen and explicitly tested across three input shapes.
- **Inverse-direction edge** (per skill rule §7.1 — symbol swap traps): the swap is `operator=OR (implicit default) → MSM 75%`. Inverse input is "single-content-token after strip" — verified above to be trivially satisfied (`ceil(1·0.75)=1` of 1), so no asymmetric break.

### Praise
- `keyword_query.go:74-109` — splitting `AnalyzeKeyword` from `buildKeywordClauseFromAnalyzed`, with `TestBuildSearchAllDSL_OneAnalyzePerKeyword` asserting `a.calls == 1` via a `countingAnalyzer`. Stronger than a query-shape pin; a future redundant analyze call would actually fail the count.
- `keyword_query.go:118-123` + `config.go:54-67,96` — `OCTO_SEARCH_STOPWORD_STRIP_ENABLED` kill switch defaults ON and the flag-off path **skips the `_analyze` RT entirely** (not just the strip), pinned by `TestBuildKeywordClauseGated_FlagOffSkipsAnalyze`'s `a.calls == 0` assertion. Ops can revert in one env-var flip with no latency cost from a vestigial analyzer call.
- `keyword_query_test.go:60-118` — three pure-stopword tests across three input shapes; the `AllStrippedAway` case specifically pins the "fall back to ORIGINAL unsplit keyword" branch that would otherwise ship `multi_match(query="")` from a naive `strings.Join([])`.
- `search_all.go:140-163` — the `eff, useMSM := req.Keyword, true` default-initialise-then-conditionally-override trick is unusually legible; a future contributor adding a third Should branch can copy the `buildKeywordClauseFromAnalyzed(eff, useMSM, ...)` line without re-deriving the §4.4 fallback contract.

### Minor (process note)
- The PR body's lower template sections (Summary / Related Issue / Changes / Testing / Checklist) are left as the empty template placeholders — the actual content is all in the top descriptive paragraph. Not blocking, but the checklist boxes (`[ ] Unit tests added/updated`, etc.) read as unchecked even though tests obviously exist. Two-minute fill-in would make the audit trail match the work.

- `defaultStopwords` (29 entries) is hardcoded as a package-level `var`. Comment in `stopwords.go:7-9` acknowledges this and points at "once telemetry shows which other tokens drive OR-trap noise" as the promotion trigger. Fine for first cut.
- MSM 75% with exactly 2 content tokens yields a 2-of-2 hard AND — narrower than some "OR" users might expect from `multi_match`. Documented design choice per spec, but if recall feedback comes back negative this is the first knob to tune (could phase from 75% → 60% based on token count).

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#435 Review Report
**PR:** Mininglamp-OSS/octo-server #435
**Head SHA:** `50e89116bdba44cae01c681ca234b8631371dc54`
**Title:** fix(messages_search): strip Chinese stopwords + MSM 75% to kill multi_match OR-trap
**Scope:** 11 files, +964/-35 (modules/messages_search/ reader-layer DSL + spec doc + tests)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| All 4 `NewMultiMatchQuery` call sites migrated | ✅ | `grep NewMultiMatchQuery` → only 1 hit in `keyword_query.go:95` (the centralized builder). search_messages.go, search_all.go (×2), search_files.go all route through `buildKeywordClause*` |
| search_media.go unaffected | ✅ | `validateKeywordMustBeEmpty` — no keyword path |
| search_around.go unaffected | ✅ | messageId anchor, no multi_match |
| `_analyze` error degrades gracefully | ✅ | `keyword_query.go:79-82` — returns `(keyword, true, err)` → §4.4 shape; all 3 handlers warn-log and proceed |
| Kill switch skips `_analyze` entirely | ✅ | `keyword_query.go:117-119` — `buildKeywordClauseGated` flag-off path: zero analyzer calls, raw keyword + MSM 75% |
| search_all shares one `_analyze` across 2 Should branches | ✅ | `search_all.go:143-163` — single `AnalyzeKeyword` call, result fed to both `buildKeywordClauseFromAnalyzed`; test `TestBuildSearchAllDSL_OneAnalyzePerKeyword` pins `calls==1` |
| Pure-stopword path preserves literal search | ✅ | `keyword_query.go:84-86` — `len(content)==0` → `(keyword, false, nil)` → no MSM; test `TestBuildKeywordClause_PureStopword_SingleToken` |
| Existing test assertions survive migration | ✅ | `fallbackTestAnalyzer()` triggers error-fallback path → raw keyword preserved in query body → historical substring pins (e.g. "hello") still match |
| Config flag wired correctly | ✅ | `config.go:93` — `parseBool(os.Getenv("OCTO_SEARCH_STOPWORD_STRIP_ENABLED"), true)` — default ON, strconv.ParseBool accepts 1/0/true/false/TRUE/FALSE |

## 2. Findings

### No P0 or P1 findings.

**P2-1: `search_all.go` inline analyze vs. gated builder asymmetry**
`search_messages.go` and `search_files.go` use `buildKeywordClauseGated` (flag-aware wrapper), but `search_all.go` inlines the flag check + `AnalyzeKeyword` call directly. The reason is sound — sharing one analyze across two Should branches — and the code has a clear comment explaining it. The asymmetry is justified but adds one more place to update if the gating logic changes.
- **Diff-scope:** new (introduced by this PR). Not a bug — deliberate optimization.
- **Suggestion:** Consider a `buildKeywordClauseShared` helper that takes `stopwordStripEnabled` + returns `(eff, useMSM)` from one analyze, so the flag logic stays in one place. Low priority.

**P2-2: MSM 75% on very short queries is effectively AND**
For 2-token content queries (e.g. "季度 报告"), MSM 75% = `ceil(0.75 × 2) = 2` → both must match (AND). For 3 tokens: `ceil(0.75 × 3) = 3` → all 3 must match. This is more restrictive than "75%" might suggest for short queries.
- **Diff-scope:** new behavior. Documented in spec §4.4 and the trade-off is acknowledged.
- **Impact:** Users searching for 2-3 word phrases may see fewer results than expected if one content word is slightly different. This is the intended OR-trap mitigation — acceptable trade-off.

**P2-3: Hardcoded stopword set (28 entries)**
`defaultStopwords` in `stopwords.go` is a conservative hardcoded set. The spec acknowledges this as a first cut with telemetry-driven iteration planned.
- **Diff-scope:** new. Not a bug — deliberate starting point.
- **Suggestion:** Consider logging stripped tokens at DEBUG level to accelerate the iteration loop without needing a separate telemetry pipeline.

## 3. Suggestions

1. **Runtime observability:** Add a DEBUG-level log in `AnalyzeKeyword` (or the handler) showing the original keyword, stripped tokens, and resulting effective keyword. This will dramatically speed up stopword-set iteration and help diagnose any "why didn't my search find X?" reports.

2. **Consider `_analyze` timeout independence:** The `_analyze` call shares the search context timeout (`h.cfg.Timeout`, default 5s). If the OS cluster is slow, `_analyze` could consume a significant portion of the budget before the actual `_search` runs. Consider a separate shorter timeout for the analyze step (e.g. 1s) so a slow analyze degrades to §4.4 without starving the search itself.

## 4. Additional Findings

None beyond the P2 observations above. The PR is well-structured:
- Clean abstraction via `tokenAnalyzer` interface (testable without OS cluster)
- Proper error propagation (error surfaces to handler for warn-log, query is still safe)
- Comprehensive test coverage: 15+ unit tests covering all 3 branches (content words / pure stopwords / analyze error), flag on/off, analyze-call counting, and DSL-level acceptance
- Existing test migration via `fallbackTestAnalyzer` preserves historical assertion pins

## 5. Data-Flow Trace

### Path 1: `search_messages` / `search_files` (single keyword clause)

```
req.Keyword (TrimSpace'd)
  → buildSearchMessagesDSL / buildSearchFilesDSL
    → buildKeywordClauseGated(ctx, analyzer, flag, keyword, fields...)
      ├─ flag=false: return (raw keyword, MSM=true) → §4.4 shape, zero _analyze calls
      └─ flag=true: buildKeywordClause → AnalyzeKeyword
           ├─ err: return (keyword, MSM=true, err) → §4.4 shape + warn-log
           ├─ all stopwords: return (keyword, MSM=false) → literal search
           └─ content tokens: return (joined tokens, MSM=true) → stripped + MSM 75%
    → buildKeywordClauseFromAnalyzed(eff, useMSM, fields...)
      → elastic.NewMultiMatchQuery(eff, fields...).Type("cross_fields")[.MinimumShouldMatch("75%")]
    → b.Must(clause)
```

All data flows verified: keyword reaches the multi_match query in the correct form for each branch. ✅

### Path 2: `search_all` (two Should branches, shared analyze)

```
req.Keyword (TrimSpace'd)
  → buildSearchAllDSL
    → eff=req.Keyword, useMSM=true (defaults)
    → if flag=true: AnalyzeKeyword once → overwrite eff/useMSM
    → buildKeywordClauseFromAnalyzed(eff, useMSM, textFields) → textClause
    → buildKeywordClauseFromAnalyzed(eff, useMSM, fileFields) → fileClause
    → b.Should(textClause, fileClause).MinimumShouldMatch("1")
```

Both branches share the same (eff, useMSM) from one analyze call. ✅

### Error propagation chain

```
AnalyzeKeyword err → buildKeywordClause err → buildSearch*DSL analyzeErr
  → handler: h.Warn("messages_search: _analyze fallback", zap.Error(analyzeErr))
  → search proceeds with degraded DSL (still valid, §4.4 shape)
```

Error is surfaced for observability but never blocks the search. ✅

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths in this change. The 4 call sites are independent endpoints, not paired operations.
- **C2 — Control-flow ordering / nested reuse:** ✅ Clear. `buildKeywordClauseFromAnalyzed` is a pure function (no side effects). `AnalyzeKeyword` is called exactly once per request (search_all shares via inline; search_messages/search_files via `buildKeywordClause`). No double-application risk. The stopword filter is a pure in-memory set lookup — no ordering sensitivity.
- **C3 — Authorization boundary:** N/A. No new endpoints, tools, or capability exposure. `_analyze` is a cluster-level API call made by the server, not exposed to clients.
- **C5 — Build ≠ runtime:** N/A. No build artifacts, extensions, CLI, or relative-path changes. The `_analyze` call targets the OS cluster directly — no packaging-dependent behavior.
- **C6 — Governance / policy doc consistency:** N/A. The spec doc (`docs/messages-search/2026-06-23-multimatch-or-trap-fix.md`) is a technical design doc, not a governance/policy document.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The PR correctly addresses the multi_match OR-trap across all 4 affected call sites, with clean abstraction, proper error degradation, a kill switch, and comprehensive test coverage. Three P2 observations are non-blocking (inline-analyze asymmetry, MSM aggressiveness on short queries, hardcoded stopword set — all documented design choices).

Jerry-Xin (approved):
One-line summary: The PR is in-scope for `octo-server` and the keyword query refactor is reasonable; I found no blocking correctness, security, or integration issue.

💬 Non-blocking

🟡 Warning — The documented kill switch is not actually “without redeploy.” `SearchConfig` is loaded once in `New()` and stored on the handler, so changing `OCTO_SEARCH_STOPWORD_STRIP_ENABLED` after process start will not affect live traffic. Either adjust the docs/PR wording or wire this through a runtime config source. See `modules/messages_search/api.go:45-52`, `modules/messages_search/config.go:76-97`, and `docs/messages-search/2026-06-23-multimatch-or-trap-fix.md:213-230`.

🟡 Warning — `_analyze` uses the same request context and full timeout budget as the subsequent `_search`. If `_analyze` blocks until the deadline, the fallback query is built but the later search uses an already-expired context. Consider a short child timeout for `_analyze` so fallback can still execute. See `modules/messages_search/search_messages.go:82-87`, `search_all.go:77-82`, `search_files.go:87-92`, and `keyword_query.go:33-39`.

🔵 Suggestion — Add a small config test for `OCTO_SEARCH_STOPWORD_STRIP_ENABLED`, especially because it intentionally uses `strconv.ParseBool` while nearby boolean envs use stricter parsing. Existing config tests cover TLS skip verify but not this new flag. See `modules/messages_search/config_test.go`.

✅ Highlights

The relevance gate passes: this directly fixes and documents behavior in the `messages_search` reader path.

The shared `AnalyzeKeyword` / `buildKeywordClauseFromAnalyzed` split is a good fit for `_search_all`, and the tests cover the important one-analyze-for-two-clauses invariant.

`go test ./modules/messages_search/...` passes locally.

yujiawei (approved):
# Code Review — PR #435 (octo-server)

Reviewed at head SHA `50e89116bdba44cae01c681ca234b8631371dc54` against merge-base `34050f64`. Build, `go vet`, and `go test ./modules/messages_search/...` all pass locally.

## Verdict: APPROVED

The reader-layer OR-trap fix is well-structured, faithfully implements the linked design doc, and is covered by a solid unit-test table. No correctness, security, data-loss, or build-break blockers. The findings below are non-blocking hardening/clarity items (P2) — none should gate merge.

## What the change does

`/_search_messages`, `/_search_all`, and `/_search_files` previously built `multi_match` with library defaults (`best_fields` + `operator=OR`), so a single high-frequency Chinese function word (e.g. `的`) matched near the whole corpus. The fix:

- Routes all four keyword sites through `keyword_query.go`: `_analyze?analyzer=ik_smart` → drop stopwords → `multi_match(cross_fields)` + `MinimumShouldMatch("75%")` on content tokens.
- Preserves literal-stopword search: when every token is a stopword, it sends the raw keyword with no MSM.
- `_search_all` shares one `_analyze` roundtrip across both Should branches (verified by `TestBuildSearchAllDSL_OneAnalyzePerKeyword`).
- Adds an ops kill switch `OCTO_SEARCH_STOPWORD_STRIP_ENABLED` (default on) that skips `_analyze` and falls back to raw keyword + cross_fields + MSM 75%.

## Verification

- ✅ Correctness vs. design doc §4.1: branch table (pure-stopword / all-stripped / content-with-stopword / fallback) matches the implementation in `keyword_query.go:60-85` and the unit assertions in `keyword_query_test.go`.
- ✅ `cross_fields` is set on every path, including fallback (`keyword_query.go:95`), so the MSM threshold is evaluated across the virtual combined field rather than per-field. This is the correct choice for the multi-field corpora here.
- ✅ Tenant isolation preserved: `applySpaceIDScope` / `applyChannelAndRevoked` / `MustNot(payloadTypeCmd)` are still applied in all three builders (`search_messages.go:155-158`, `search_files.go:138-139`, `search_all.go:131-132`). The refactor did not drop any scoping filter.
- ✅ Injection surface: keyword flows in as the `multi_match` query *value* (parameterized by the client lib) and as the `_analyze` `text` param, not as raw DSL. It is length-bounded to 64 runes by `validateKeywordOptional` *before* the new `_analyze` call, so the added roundtrip carries no amplification risk.
- ✅ Empty keyword short-circuits before any `_analyze` call in all three handlers.
- ✅ Test migration of `dsl_test.go` / `space_filter_test.go` via `fallbackTestAnalyzer` is sound — it keeps the original keyword intact so the historical substring pins still apply; branch-specific shape is asserted separately in `keyword_query_test.go`.

## Findings (P2 — non-blocking)

### P2-1 — `_analyze` shares the request deadline with `_search`
`search_messages.go:82`, `search_files.go:87`, `search_all.go:77`

`_analyze` runs on the same `context.WithTimeout(c.Request.Context(), h.cfg.Timeout)` that the subsequent `_search` uses. If `_analyze` is slow (not failed, just degraded latency), it consumes the shared budget; the fallback query it returns can then run against an already-near-expired context and fail. The point of the error-fallback is "IK hiccuped, still serve the query," and a shared deadline undercuts that.

Note this is bounded in practice: the text fields are IK-analyzed at query time too, so a fully-wedged IK plugin would likely fail `_search` regardless, and the kill switch exists precisely for sustained IK degradation. Failure mode is fail-closed (error to caller), no data exposure. Recommendation: give `_analyze` its own short sub-context (e.g. a few hundred ms) carved from the request budget so a slow analyze cannot starve the search.

### P2-2 — fallback / short-query MSM is weaker than the "ceil" framing in the doc
`keyword_query.go:96`, doc §4 line ~82

OpenSearch `minimum_should_match: "75%"` rounds the required count *down*. So a 2-content-token query requires 1 (effectively OR between two content words), and the raw 5-token fallback/flag-off path requires 3 of 5 (with `的` countable). The main stopword-stripped path is unaffected for the reproducer (4 content tokens → 3 required), and the doc itself flags the §4.4 degraded path as intentionally weaker, so this is by-design rather than a bug. One real inaccuracy: the doc's parenthetical `MSM 75% = ceil(0.75×3) = 3 = AND` (line ~82) is wrong for ES/OS percentage semantics — it floors to 2. Worth correcting the doc so future readers don't rely on ceil rounding. No code change required unless ceil semantics are actually desired for short content queries (then use a conditional/`-25%` form).

### P2-3 — kill switch is read once at startup, not hot-reloadable
`config.go:77` (`loadConfig`), doc §8

The doc describes `OCTO_SEARCH_STOPWORD_STRIP_ENABLED` as a flip "without redeploy," but `loadConfig()` reads env once into `Handler.cfg` at construction (`api.go:46`). Flipping the env on a running process has no effect without a restart/reload. Either soften the doc wording to "config change + restart" or wire the flag through a reloadable config source if true zero-restart toggling is intended.

## Notes for the human reviewer (security_sensitive label)

This PR was auto-labeled `security_sensitive` on a keyword match; the actual change is in the search reader DSL and touches no auth, token, or credential path. The one thing worth a human glance is **tenant scoping survival** through the refactor — I verified `applySpaceIDScope` is still present in all three rebuilt DSL functions (see Verification above), but it is the highest-value invariant here, so a second look is reasonable.

## Coverage / blind spots

- Unit coverage of the DSL branch logic is strong. The e2e recall-delta cases (design doc §5 integration table) are deferred to `make env-test` against a real OS+IK cluster and are **not** exercised by this PR's CI — the actual recall improvement on the `"按时缴纳时间的女性"` reproducer is asserted by design, not by a running test here. Recommend running the env-test suite before/at rollout.
- Not independently re-verified: the live OS field mappings actually use an IK analyzer at query time for the boosted fields (assumed from context); if any targeted field is `keyword`-typed or uses a non-IK analyzer, the cross_fields/MSM interaction there would differ.
- This review absorbed an independent line-level adversarial pass (which surfaced the three items above) and an independent whole-diff read. A second long-context advisory leg was attempted but **did not return findings (absent/errored)** — its silence is not treated as agreement; the coverage above reflects only what was actually checked.

PR #437, opened 2026-06-23
fix: split health liveness from readiness

### LINKED ISSUE #436: Split /v1/health liveness from DB/Redis readiness checks
## Description

Frontend status-tower polling of `GET /v1/health` shows severe latency jitter. Browser Timing evidence from a slow sample showed about 2.18s total latency, with Queueing/Stalled/Proxy/TLS/download all in the microsecond range and essentially all time spent in `Waiting for server response` (server-side TTFB).

Current backend implementation makes `/v1/health` a synchronous dependency probe instead of a cheap liveness probe:

```text
global Redis-backed RateLimitMiddleware
  -> db.session.Ping()
  -> redis.Ping()
  -> JSON response
```

Observed in code:

- `main.go` mounts global `RateLimitMiddleware(..., "/v1/ping")`; `/v1/health` is not excluded.
- `modules/common/api.go` handles `/v1/health` by serially calling MySQL `Ping()` and Redis `Ping()`.
- When either dependency fails, the response body includes `error: lastError.Error()`, which can expose internal details to browser clients.

This ties status-tower latency to transient DB pool pressure and Redis latency. In particular, MySQL `Ping()` must acquire a DB pool connection; when the pool is saturated by slow queries or long transactions, the health request can queue waiting for an idle connection. That matches intermittent 2s-class spikes rather than a stable network or frontend delay.

## Steps to Reproduce

1. Deploy octo-server with frontend status-tower polling `GET /v1/health`.
2. Put transient pressure on MySQL connection usage or Redis latency.
3. Observe browser Timing for `/v1/health`.
4. Slow samples spend nearly all time in server response waiting / TTFB, not in queueing, proxy, TLS handshake, or download.

## Expected Behavior

`GET /v1/health` should be a stable, cheap liveness endpoint for frontend/LB/k8s liveness use:

- It should answer whether the octo-server process is alive and can serve HTTP.
- It should not synchronously call DB or Redis.
- It should not be affected by DB connection pool saturation or Redis command latency.
- It should not expose internal dependency error messages to browser clients.

Dependency health should be exposed separately as readiness, with explicit timeouts and safe response fields.

## Actual Behavior

`GET /v1/health` synchronously probes real dependencies and therefore reports dependency/load jitter as service liveness jitter:

- Global Redis-backed IP limiter can add Redis dependency before the handler runs.
- MySQL `Ping()` can block waiting for a pooled connection.
- Redis `Ping()` can block on Redis command latency.
- No per-dependency context timeout is applied.
- Internal error text is copied into the response body.

## DMWork Version

Current `main` / production deployment version to be filled by operator.

## Affected Component

Backend (Server)

## Severity

P2 - Non-core Feature Issue

## Environment

- Component: octo-server backend
- Endpoint: `GET /v1/health`
- Caller: web frontend status tower / browser
- Deployment: production-like HTTP/2 frontend path

## Logs / Screenshots

Browser Timing evidence from one slow sample:

- Total request time: about 2.18s
- Queueing/Stalled/Proxy/TLS/download: microsecond-level
- Time spent in `Waiting for server response`: effectively 100%

Code references:

- `modules/common/api.go`: `/v1/health` performs `db.session.Ping()` and `redis.Ping()` synchronously.
- `main.go`: global `RateLimitMiddleware` excludes `/v1/ping`, but not `/v1/health`.

## Workaround

Operational workaround until code is fixed:

- Reduce frontend polling frequency if it is high.
- Avoid using `/v1/health` latency as service liveness latency SLO.
- Use infrastructure-side DB/Redis dashboards to diagnose dependency jitter separately.

## Proposed Fix

1. Change `/v1/health` into pure liveness:
   - only read in-process state;
   - return a stable `{"status":"up"}`-style response;
   - remove dependency pings;
   - remove raw `error` response field.
2. Exclude `/v1/health` from the global Redis-backed `RateLimitMiddleware`, or provide an equivalent low-cost bypass for liveness paths.
3. Add a separate readiness endpoint, for example `/v1/ready`:
   - check MySQL and Redis with `context` timeouts such as 300-500ms;
   - run checks in parallel to avoid serial timeout stacking;
   - return safe `up/down/stale` fields, not internal error strings;
   - return `503` when dependencies are not ready if used by k8s readiness.
4. Prefer background dependency probing with cached status:
   - periodic goroutine probes DB/Redis with timeouts;
   - `/v1/health` remains dependency-free;
   - `/v1/ready` reads cached dependency status and fails if the cache is stale.
5. Add observability:
   - dependency health gauge by dependency;
   - probe latency histogram;
   - consecutive failure count;
   - DB pool stats such as `InUse`, `Idle`, `WaitCount`, `WaitDuration`.

## Acceptance Criteria

- Frontend status-tower calls to `/v1/health` no longer synchronously touch MySQL or Redis.
- `/v1/health` latency is not impacted by DB pool saturation or Redis slowness.
- `/v1/health` does not expose raw internal errors.
- Real dependency health is available through a separate readiness path with bounded timeouts.
- k8s/LB/frontend usage is documented:
  - liveness/frontend status: `/v1/health`;
  - readiness/SRE dependency check: readiness endpoint.
- Tests cover:
  - health response does not call dependency ping functions;
  - readiness times out quickly on slow dependencies;
  - response bodies do not include raw internal error strings.

an9xyz:
## Summary

Split the public health probe into cheap process liveness and bounded dependency readiness so frontend status-tower polling no longer synchronously waits on DB/Redis.

## Related Issue

Fixes #436

## Linked Spec

`.octospec/tasks/health-readiness-split/brief.md`

## Changes

- Change `GET /v1/health` to pure liveness without DB/Redis pings while preserving the legacy success response shape: `200 {"status":"up","db":"up","redis":"up"}`.
- Add `GET /v1/ready` for readiness with parallel MySQL `PingContext` and Redis `PING`, bounded by a 500ms request context.
- Keep raw dependency errors out of readiness responses while logging them server-side.
- Exclude `/v1/health` from global Redis-backed IP rate limiting; keep `/v1/ready` rate limited because it touches shared dependencies.
- Suppress access logs for `/v1/health` and `/v1/ready` to avoid high-frequency probe log noise.
- Add focused tests for liveness purity, readiness safe output/timeout behavior, rate-limit exclusions, and access-log ignore paths.

## Testing

- [x] Manually verified

Commands run:

```bash
go test . -run 'TestGlobalRateLimitExcludePathsIncludesProbeEndpoints|TestAccessLogIgnorePathsIncludesProbeEndpoints'
go test ./modules/common -run 'TestHealthIsPureLiveness|TestReadyReturnsSafeDependencyStatus|TestReadyBoundsCheckerWithTimeout'
go test ./tools/lint-direct-error-response
go test ./modules/common
```

Note: local `go test ./modules/common` initially hit the known shared `test` database migration-ledger issue (`unknown migration in database`). I reset the local Docker `octo-test-mysql` `test` database using the same drop/recreate pattern documented in CI, then reran `go test ./modules/common` successfully.

## COMPREHENSION

   `/v1/health` no longer enters DB/Redis dependency checks or the Redis-backed global IP limiter, so status-tower/liveness traffic measures process HTTP responsiveness instead of transient dependency load. It preserves the legacy success fields (`db`, `redis`, `status`) for response-shape compatibility, but those fields now represent liveness compatibility only. Real dependency probing moves to `/v1/ready`, where DB and Redis are checked in parallel with a bounded context.

   Any consumer that treated `/v1/health` as a dependency-health endpoint will continue to see the legacy success fields but no longer receives real-time DB/Redis status there. Those consumers should move to `/v1/ready`. `/v1/ready` returns `503` when DB or Redis is down and remains globally rate limited, so callers must treat it as an infra readiness probe rather than a high-frequency public polling endpoint.

   The new `modules/common` tests prove `/v1/health` does not call the readiness checker, `/v1/ready` returns safe dependency status without leaking raw errors, and readiness receives a bounded context. The main-package tests pin the global rate-limit and access-log path lists so `/v1/health` stays cheap while `/v1/ready` remains protected.

## Checklist

- [x] Updated documentation

### DIFF
--- a/main.go
+++ b/main.go
@@ -41,6 +41,10 @@ var Commit string     // git commit id
 var CommitDate string // git commit date
 var TreeState string  // git tree state
 
+func globalRateLimitExcludePaths() []string {
+	return []string{"/v1/ping", "/v1/health", "/v1/ready"}
+}
+
 func loadConfigFromFile(cfgFile string) *viper.Viper {
 	vp := viper.New()
 	vp.SetConfigFile(cfgFile)
@@ -182,7 +186,7 @@ func runAPI(ctx *config.Context) {
 		o.MaxRetries = 1
 		o.PoolSize = 10
 	}))
-	route.Use(route.RateLimitMiddleware(context.Background(), rlRedis, rps, burst, "/v1/ping"))
+	route.Use(route.RateLimitMiddleware(context.Background(), rlRedis, rps, burst, globalRateLimitExcludePaths()...))
 	// CORS 白名单覆盖：dmwork-lib 的 server.New 默认注入 "*" + Credentials:true，
 	// 本中间件在其后执行，按 DM_CORS_ALLOWED_ORIGINS 重写/剥离 Allow-Origin/Credentials。
 	// 未配置时等价于禁用跨域（剥离所有 CORS 响应头），仅允许同源调用。
--- a/main.go
+++ b/main.go
@@ -42,7 +42,7 @@ var CommitDate string // git commit date
 var TreeState string  // git tree state
 
 func globalRateLimitExcludePaths() []string {
-	return []string{"/v1/ping", "/v1/health", "/v1/ready"}
+	return []string{"/v1/ping", "/v1/health"}
 }
 
 func loadConfigFromFile(cfgFile string) *viper.Viper {
@@ -357,6 +357,8 @@ func ingorePaths() []string {
 	return []string{
 		"/v1/robots/:robot_id/:app_key/events",
 		"/v1/ping",
+		"/v1/health",
+		"/v1/ready",
 	}
 }
 
--- a/modules/common/health.go
+++ b/modules/common/health.go
@@ -56,6 +56,7 @@ func readinessRedisOptions(cfg *config.Config) *rd.Options {
 }
 
 func (cn *Common) health(c *wkhttp.Context) {
+	// db/redis are static legacy-shape fields; liveness must stay dependency-free.
 	c.JSON(http.StatusOK, map[string]string{
 		"status": healthStatusUp,
 		"db":     healthStatusUp,
--- a/modules/common/health.go
+++ b/modules/common/health.go
@@ -18,7 +18,8 @@ const (
 	healthStatusUp   = "up"
 	healthStatusDown = "down"
 
-	readinessProbeTimeout = 500 * time.Millisecond
+	readinessProbeTimeout     = 500 * time.Millisecond
+	readinessRedisPoolTimeout = 100 * time.Millisecond
 )
 
 type readinessResult struct {
@@ -37,17 +38,21 @@ type dependencyReadinessChecker struct {
 }
 
 func newDependencyReadinessChecker(ctx *config.Context, db *db) readinessChecker {
-	redisClient := rd.NewClient(octoredis.MustBuildOptions(ctx.GetConfig(), func(o *rd.Options) {
+	return &dependencyReadinessChecker{
+		db:          db,
+		redisClient: rd.NewClient(readinessRedisOptions(ctx.GetConfig())),
+	}
+}
+
+func readinessRedisOptions(cfg *config.Config) *rd.Options {
+	return octoredis.MustBuildOptions(cfg, func(o *rd.Options) {
 		o.MaxRetries = 0
 		o.PoolSize = 2
 		o.DialTimeout = readinessProbeTimeout
 		o.ReadTimeout = readinessProbeTimeout
 		o.WriteTimeout = readinessProbeTimeout
-	}))
-	return &dependencyReadinessChecker{
-		db:          db,
-		redisClient: redisClient,
-	}
+		o.PoolTimeout = readinessRedisPoolTimeout
+	})
 }
 
 func (cn *Common) health(c *wkhttp.Context) {
@@ -72,7 +77,24 @@ func (cn *Common) ready(c *wkhttp.Context) {
 		return
 	}
 
-	result := checker.Check(ctx)
+	resultCh := make(chan readinessResult, 1)
+	go func() {
+		resultCh <- checker.Check(ctx)
+	}()
+
+	var result readinessResult
+	select {
+	case result = <-resultCh:
+	case <-ctx.Done():
+		result = readinessResult{
+			Status: healthStatusDown,
+			Dependencies: map[string]string{
+				"db":    healthStatusDown,
+				"redis": healthStatusDown,
+			},
+			Errors: map[string]error{"readiness": ctx.Err()},
+		}
+	}
 	cn.logReadinessErrors(result.Errors)
 	statusCode := http.StatusOK
 	if result.Status != healthStatusUp {
--- a/modules/common/health.go
+++ b/modules/common/health.go
@@ -1,0 +1,150 @@
+package common
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	octoredis "github.com/Mininglamp-OSS/octo-server/pkg/redis"
+	rd "github.com/go-redis/redis"
+	"go.uber.org/zap"
+)
+
+const (
+	healthStatusUp   = "up"
+	healthStatusDown = "down"
+
+	readinessProbeTimeout = 500 * time.Millisecond
+)
+
+type readinessResult struct {
+	Status       string            `json:"status"`
+	Dependencies map[string]string `json:"dependencies,omitempty"`
+	Errors       map[string]error  `json:"-"`
+}
+
+type readinessChecker interface {
+	Check(ctx context.Context) readinessResult
+}
+
+type dependencyReadinessChecker struct {
+	db          *db
+	redisClient *rd.Client
+}
+
+func newDependencyReadinessChecker(ctx *config.Context, db *db) readinessChecker {
+	redisClient := rd.NewClient(octoredis.MustBuildOptions(ctx.GetConfig(), func(o *rd.Options) {
+		o.MaxRetries = 0
+		o.PoolSize = 2
+		o.DialTimeout = readinessProbeTimeout
+		o.ReadTimeout = readinessProbeTimeout
+		o.WriteTimeout = readinessProbeTimeout
+	}))
+	return &dependencyReadinessChecker{
+		db:          db,
+		redisClient: redisClient,
+	}
+}
+
+func (cn *Common) health(c *wkhttp.Context) {
+	c.JSON(http.StatusOK, map[string]string{"status": healthStatusUp})
+}
+
+func (cn *Common) ready(c *wkhttp.Context) {
+	ctx, cancel := context.WithTimeout(c.Request.Context(), readinessProbeTimeout)
+	defer cancel()
+
+	checker := cn.readinessChecker
+	if checker == nil {
+		result := readinessResult{
+			Status:       healthStatusDown,
+			Dependencies: map[string]string{"db": healthStatusDown, "redis": healthStatusDown},
+			Errors:       map[string]error{"readiness": errors.New("readiness checker unavailable")},
+		}
+		cn.logReadinessErrors(result.Errors)
+		// Readiness is an infra probe contract, not a user-facing business error.
+		// Keep the wire body small and safe instead of returning the i18n envelope.
+		c.JSON(http.StatusServiceUnavailable, result)
+		return
+	}
+
+	result := checker.Check(ctx)
+	cn.logReadinessErrors(result.Errors)
+	statusCode := http.StatusOK
+	if result.Status != healthStatusUp {
+		statusCode = http.StatusServiceUnavailable
+	}
+	// Readiness is an infra probe contract, not a user-facing business error.
+	// Keep the wire body small and safe instead of returning the i18n envelope.
+	c.JSON(statusCode, result)
+}
+
+func (cn *Common) logReadinessErrors(errs map[string]error) {
+	if len(errs) == 0 || cn == nil || cn.Log == nil {
+		return
+	}
+	for dependency, err := range errs {
+		if err == nil {
+			continue
+		}
+		cn.Error("readiness dependency check failed", zap.String("dependency", dependency), zap.Error(err))
+	}
+}
+
+func (c *dependencyReadinessChecker) Check(ctx context.Context) readinessResult {
+	result := readinessResult{
+		Status: healthStatusUp,
+		Dependencies: map[string]string{
+			"db":    healthStatusUp,
+			"redis": healthStatusUp,
+		},
+		Errors: map[string]error{},
+	}
+
+	var mu sync.Mutex
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
+		defer wg.Done()
+		if err := c.pingDB(ctx); err != nil {
+			mu.Lock()
+			result.Dependencies["db"] = healthStatusDown
+			result.Errors["db"] = err
+			mu.Unlock()
+		}
+	}()
+	go func() {
+		defer wg.Done()
+		if err := c.pingRedis(ctx); err != nil {
+			mu.Lock()
+			result.Dependencies["redis"] = healthStatusDown
+			result.Errors["redis"] = err
+			mu.Unlock()
+		}
+	}()
+	wg.Wait()
+
+	if len(result.Errors) > 0 {
+		result.Status = healthStatusDown
+	}
+	return result
+}
+
+func (c *dependencyReadinessChecker) pingDB(ctx context.Context) error {
+	if c == nil || c.db == nil || c.db.session == nil || c.db.session.DB == nil {
+		return errors.New("db session unavailable")
+	}
+	return c.db.session.DB.PingContext(ctx)
+}
+
+func (c *dependencyReadinessChecker) pingRedis(ctx context.Context) error {
+	if c == nil || c.redisClient == nil {
+		return errors.New("redis client unavailable")
+	}
+	_, err := c.redisClient.WithContext(ctx).Ping().Result()
+	return err
+}
--- a/modules/common/health.go
+++ b/modules/common/health.go
@@ -56,7 +56,11 @@ func readinessRedisOptions(cfg *config.Config) *rd.Options {
 }
 
 func (cn *Common) health(c *wkhttp.Context) {
-	c.JSON(http.StatusOK, map[string]string{"status": healthStatusUp})
+	c.JSON(http.StatusOK, map[string]string{
+		"status": healthStatusUp,
+		"db":     healthStatusUp,
+		"redis":  healthStatusUp,
+	})
 }
 
 func (cn *Common) ready(c *wkhttp.Context) {
--- a/modules/common/health_test.go
+++ b/modules/common/health_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/Mininglamp-OSS/octo-lib/config"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
 	"github.com/stretchr/testify/require"
 )
@@ -98,4 +99,33 @@ func TestReadyBoundsCheckerWithTimeout(t *testing.T) {
 	require.Equal(t, http.StatusServiceUnavailable, w.Code)
 	require.Less(t, time.Since(started), time.Second, "readiness timeout must stay bounded")
 	require.Equal(t, int32(1), atomic.LoadInt32(&checker.calls))
+}
+
+func TestReadyReturnsWhenCheckerIgnoresContext(t *testing.T) {
+	checker := &fakeReadinessChecker{
+		check: func(context.Context) readinessResult {
+			time.Sleep(readinessProbeTimeout * 2)
+			return readinessResult{Status: healthStatusUp}
+		},
+	}
+	cn := &Common{readinessChecker: checker}
+	r := wkhttp.New()
+	r.GET("/v1/ready", cn.ready)
+
+	started := time.Now()
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/v1/ready", nil)
+	r.ServeHTTP(w, req)
+
+	require.Equal(t, http.StatusServiceUnavailable, w.Code)
+	require.Less(t, time.Since(started), 750*time.Millisecond, "readiness must return on its own deadline even if a checker ignores ctx")
+	require.JSONEq(t, `{"status":"down","dependencies":{"db":"down","redis":"down"}}`, w.Body.String())
+	require.Equal(t, int32(1), atomic.LoadInt32(&checker.calls))
+}
+
+func TestReadinessRedisOptionsUseBoundedPoolTimeout(t *testing.T) {
+	opts := readinessRedisOptions(config.New())
+
+	require.Equal(t, readinessRedisPoolTimeout, opts.PoolTimeout)
+	require.Less(t, opts.PoolTimeout, readinessProbeTimeout)
 }
--- a/modules/common/health_test.go
+++ b/modules/common/health_test.go
@@ -1,0 +1,101 @@
+package common
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/stretchr/testify/require"
+)
+
+type fakeReadinessChecker struct {
+	calls  int32
+	result readinessResult
+	check  func(context.Context) readinessResult
+}
+
+func (f *fakeReadinessChecker) Check(ctx context.Context) readinessResult {
+	atomic.AddInt32(&f.calls, 1)
+	if f.check != nil {
+		return f.check(ctx)
+	}
+	return f.result
+}
+
+func TestHealthIsPureLiveness(t *testing.T) {
+	checker := &fakeReadinessChecker{
+		result: readinessResult{
+			Status:       healthStatusDown,
+			Dependencies: map[string]string{"db": healthStatusDown, "redis": healthStatusDown},
+			Errors:       map[string]error{"db": errors.New("should not be called")},
+		},
+	}
+	cn := &Common{readinessChecker: checker}
+	r := wkhttp.New()
+	r.GET("/v1/health", cn.health)
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/v1/health", nil)
+	r.ServeHTTP(w, req)
+
+	require.Equal(t, http.StatusOK, w.Code)
+	require.JSONEq(t, `{"status":"up"}`, w.Body.String())
+	require.Equal(t, int32(0), atomic.LoadInt32(&checker.calls), "liveness must not call dependency readiness checker")
+}
+
+func TestReadyReturnsSafeDependencyStatus(t *testing.T) {
+	checker := &fakeReadinessChecker{
+		result: readinessResult{
+			Status:       healthStatusDown,
+			Dependencies: map[string]string{"db": healthStatusDown, "redis": healthStatusUp},
+			Errors:       map[string]error{"db": errors.New("mysql dsn contains secret password")},
+		},
+	}
+	cn := &Common{readinessChecker: checker}
+	r := wkhttp.New()
+	r.GET("/v1/ready", cn.ready)
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/v1/ready", nil)
+	r.ServeHTTP(w, req)
+
+	require.Equal(t, http.StatusServiceUnavailable, w.Code)
+	require.JSONEq(t, `{"status":"down","dependencies":{"db":"down","redis":"up"}}`, w.Body.String())
+	require.NotContains(t, strings.ToLower(w.Body.String()), "secret")
+	require.NotContains(t, strings.ToLower(w.Body.String()), "password")
+	require.Equal(t, int32(1), atomic.LoadInt32(&checker.calls))
+}
+
+func TestReadyBoundsCheckerWithTimeout(t *testing.T) {
+	checker := &fakeReadinessChecker{
+		check: func(ctx context.Context) readinessResult {
+			deadline, ok := ctx.Deadline()
+			require.True(t, ok, "readiness checker must receive a bounded context")
+			require.LessOrEqual(t, time.Until(deadline), readinessProbeTimeout)
+			<-ctx.Done()
+			return readinessResult{
+				Status:       healthStatusDown,
+				Dependencies: map[string]string{"db": healthStatusDown, "redis": healthStatusDown},
+				Errors:       map[string]error{"timeout": ctx.Err()},
+			}
+		},
+	}
+	cn := &Common{readinessChecker: checker}
+	r := wkhttp.New()
+	r.GET("/v1/ready", cn.ready)
+
+	started := time.Now()
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/v1/ready", nil)
+	r.ServeHTTP(w, req)
+
+	require.Equal(t, http.StatusServiceUnavailable, w.Code)
+	require.Less(t, time.Since(started), time.Second, "readiness timeout must stay bounded")
+	require.Equal(t, int32(1), atomic.LoadInt32(&checker.calls))
+}
--- a/modules/common/health_test.go
+++ b/modules/common/health_test.go
@@ -46,7 +46,7 @@ func TestHealthIsPureLiveness(t *testing.T) {
 	r.ServeHTTP(w, req)
 
 	require.Equal(t, http.StatusOK, w.Code)
-	require.JSONEq(t, `{"status":"up"}`, w.Body.String())
+	require.JSONEq(t, `{"status":"up","db":"up","redis":"up"}`, w.Body.String())
 	require.Equal(t, int32(0), atomic.LoadInt32(&checker.calls), "liveness must not call dependency readiness checker")
 }
 
--- a/.octospec/tasks/health-readiness-split/brief.md
+++ b/.octospec/tasks/health-readiness-split/brief.md
@@ -1,0 +1,49 @@
+---
+type: Task
+title: "Task: health-readiness-split"
+description: Split cheap liveness from bounded DB/Redis readiness checks.
+tags: ["health", "readiness", "rate-limit"]
+timestamp: 2026-06-23T00:00:00Z
+slug: health-readiness-split
+upstream: "Mininglamp-OSS/octo-server#436"
+source: self
+---
+
+# Task: health-readiness-split
+
+## Goal
+
+Make `/v1/health` a cheap process liveness endpoint for frontend/LB probes, and
+move DB/Redis dependency probing to a bounded readiness endpoint.
+
+## Background
+
+Issue #436: frontend status-tower polling `/v1/health` shows TTFB jitter because
+the current handler synchronously runs global Redis-backed rate limiting,
+MySQL `Ping()`, and Redis `Ping()`.
+
+## Load-bearing list
+
+- `/v1/health` response semantics change from dependency probe to process
+  liveness.
+- Add `/v1/ready` as the dependency readiness endpoint.
+- Global Redis-backed rate limiting must not add Redis latency to health or
+  readiness probe paths.
+- Dependency errors must be logged server-side and not exposed as raw response
+  strings.
+
+## Out of scope
+
+- Background cached dependency probing.
+- New Prometheus dependency health metrics.
+- Deployment manifest changes for k8s probes.
+
+## Acceptance
+
+- `/v1/health` does not call DB or Redis and returns `200 {"status":"up"}`.
+- `/v1/ready` checks DB/Redis with bounded timeout and returns `503` when any
+  dependency is down.
+- `/v1/ready` response includes safe `up/down` status fields only.
+- `/v1/health` and `/v1/ready` are excluded from the global Redis-backed IP
+  limiter.
+- Focused Go tests pass for `./modules/common`.
--- a/.octospec/tasks/health-readiness-split/brief.md
+++ b/.octospec/tasks/health-readiness-split/brief.md
@@ -27,8 +27,9 @@ MySQL `Ping()`, and Redis `Ping()`.
 - `/v1/health` response semantics change from dependency probe to process
   liveness.
 - Add `/v1/ready` as the dependency readiness endpoint.
-- Global Redis-backed rate limiting must not add Redis latency to health or
-  readiness probe paths.
+- Global Redis-backed rate limiting must not add Redis latency to the liveness
+  probe path. Readiness intentionally keeps the global limiter because it
+  touches DB/Redis dependencies.
 - Dependency errors must be logged server-side and not exposed as raw response
   strings.
 
--- a/.octospec/tasks/health-readiness-split/brief.md
+++ b/.octospec/tasks/health-readiness-split/brief.md
@@ -44,6 +44,7 @@ MySQL `Ping()`, and Redis `Ping()`.
 - `/v1/ready` checks DB/Redis with bounded timeout and returns `503` when any
   dependency is down.
 - `/v1/ready` response includes safe `up/down` status fields only.
-- `/v1/health` and `/v1/ready` are excluded from the global Redis-backed IP
-  limiter.
+- `/v1/health` is excluded from the global Redis-backed IP limiter; `/v1/ready`
+  keeps the global limiter because it touches DB/Redis dependencies.
+- `/v1/health` and `/v1/ready` are excluded from high-volume access logging.
 - Focused Go tests pass for `./modules/common`.
--- a/.octospec/tasks/health-readiness-split/brief.md
+++ b/.octospec/tasks/health-readiness-split/brief.md
@@ -41,7 +41,8 @@ MySQL `Ping()`, and Redis `Ping()`.
 
 ## Acceptance
 
-- `/v1/health` does not call DB or Redis and returns `200 {"status":"up"}`.
+- `/v1/health` does not call DB or Redis and keeps the legacy success shape:
+  `200 {"status":"up","db":"up","redis":"up"}`.
 - `/v1/ready` checks DB/Redis with bounded timeout and returns `503` when any
   dependency is down.
 - `/v1/ready` response includes safe `up/down` status fields only.
--- a/main_test.go
+++ b/main_test.go
@@ -1,0 +1,15 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGlobalRateLimitExcludePathsIncludesProbeEndpoints(t *testing.T) {
+	paths := globalRateLimitExcludePaths()
+
+	require.Contains(t, paths, "/v1/ping")
+	require.Contains(t, paths, "/v1/health")
+	require.Contains(t, paths, "/v1/ready")
+}
--- a/main_test.go
+++ b/main_test.go
@@ -11,5 +11,13 @@ func TestGlobalRateLimitExcludePathsIncludesProbeEndpoints(t *testing.T) {
 
 	require.Contains(t, paths, "/v1/ping")
 	require.Contains(t, paths, "/v1/health")
+	require.NotContains(t, paths, "/v1/ready")
+}
+
+func TestAccessLogIgnorePathsIncludesProbeEndpoints(t *testing.T) {
+	paths := ingorePaths()
+
+	require.Contains(t, paths, "/v1/ping")
+	require.Contains(t, paths, "/v1/health")
 	require.Contains(t, paths, "/v1/ready")
 }
--- a/modules/common/api.go
+++ b/modules/common/api.go
@@ -32,10 +32,11 @@ import (
 type Common struct {
 	ctx *config.Context
 	log.Log
-	db             *db
-	appConfigDB    *appConfigDB
-	systemSettings *SystemSettings
-	threadOn       int // 缓存 DM_THREAD_ON 环境变量
+	db               *db
+	appConfigDB      *appConfigDB
+	systemSettings   *SystemSettings
+	readinessChecker readinessChecker
+	threadOn         int // 缓存 DM_THREAD_ON 环境变量
 }
 
 // New New
@@ -44,13 +45,15 @@ func New(ctx *config.Context) *Common {
 	if t := strings.ToLower(os.Getenv("DM_THREAD_ON")); t == "true" || t == "1" {
 		threadOn = 1
 	}
+	database := newDB(ctx.DB())
 	return &Common{
-		ctx:            ctx,
-		db:             newDB(ctx.DB()),
-		appConfigDB:    newAppConfigDB(ctx),
-		systemSettings: EnsureSystemSettings(ctx),
-		Log:            log.NewTLog("common"),
-		threadOn:       threadOn,
+		ctx:              ctx,
+		db:               database,
+		appConfigDB:      newAppConfigDB(ctx),
+		systemSettings:   EnsureSystemSettings(ctx),
+		readinessChecker: newDependencyReadinessChecker(ctx, database),
+		Log:              log.NewTLog("common"),
+		threadOn:         threadOn,
 	}
 }
 
@@ -82,37 +85,8 @@ func (cn *Common) Route(r *wkhttp.WKHttp) {
 		commonNoAuth.GET("/changelog", cn.changelog)           // 版本更新日志（公开）
 	}
 
-	r.GET("/v1/health", func(c *wkhttp.Context) {
-		var (
-			statusMap = map[string]string{
-				"status": "up",
-				"db":     "up",
-				"redis":  "up",
-			}
-			lastError error
-		)
-
-		err := cn.db.session.Ping()
-		if err != nil {
-			cn.Error("db ping error", zap.Error(err))
-			lastError = err
-			statusMap["db"] = "down"
-		}
-
-		_, err = cn.ctx.GetRedisConn().Ping()
-		if err != nil {
-			cn.Error("redis ping error", zap.Error(err))
-			lastError = err
-			statusMap["redis"] = "down"
-		}
-
-		if lastError != nil {
-			statusMap["status"] = "down"
-			statusMap["error"] = lastError.Error()
-		}
-
-		c.JSON(http.StatusOK, statusMap)
-	})
+	r.GET("/v1/health", cn.health)
+	r.GET("/v1/ready", cn.ready)
 
 	appConfigM, err := cn.insertAppConfigIfNeed()
 	if err != nil {


modules/common/health_test.go
```diff
@@ -0,0 +1,101 @@
...
+	r := wkhttp.New()
+	r.GET("/v1/health", cn.health)
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/v1/health", nil)
+	r.ServeHTTP(w, req)
+
+	require.Equal(t, http.StatusOK, w.Code)
+	require.JSONEq(t, `{"status":"up"}`, w.Body.String())
+	require.Equal(t, int32(0), atomic.LoadInt32(&checker.calls), "liveness must not call dependency readiness checker")
```
OctoBoooot, line 50: praise: `TestHealthIsPureLiveness` pins `atomic.LoadInt32(&checker.calls) == int32(0)` — that's the right invariant to lock in, stronger than asserting only the response shape. A future refactor that accidentally wires the dependency checker into the liveness handler (e.g., copy-paste from `ready`) would fail this test by name. Combined with the `fakeReadinessChecker` whose `result` says "down" with an error, the test would also fail loudly if liveness ever started reflecting dependency state — the exact regression the PR is trying to prevent.

modules/common/health_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/v1/ready", nil)
+	r.ServeHTTP(w, req)
+
+	require.Equal(t, http.StatusServiceUnavailable, w.Code)
+	require.JSONEq(t, `{"status":"down","dependencies":{"db":"down","redis":"up"}}`, w.Body.String())
+	require.NotContains(t, strings.ToLower(w.Body.String()), "secret")
+	require.NotContains(t, strings.ToLower(w.Body.String()), "password")
+	require.Equal(t, int32(1), atomic.LoadInt32(&checker.calls))
```
OctoBoooot, line 73: praise: `NotContains "secret" / "password"` on the response body (lines 70-71), driven by a fake checker that puts the literal `"mysql dsn contains secret password"` into its `Errors` map, pins the safety contract concretely — not via a `json:"-"` tag alone. The combination (tag + assertion) survives both "someone removes the tag" and "someone adds a new field that leaks the error." Same shape would catch a future contributor switching to `c.JSON(503, result.Errors)` thinking it's a debug aid.

modules/common/health.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func newDependencyReadinessChecker(ctx *config.Context, db *db) readinessChecker {
+	redisClient := rd.NewClient(octoredis.MustBuildOptions(ctx.GetConfig(), func(o *rd.Options) {
+		o.MaxRetries = 0
+		o.PoolSize = 2
+		o.DialTimeout = readinessProbeTimeout
+		o.ReadTimeout = readinessProbeTimeout
+		o.WriteTimeout = readinessProbeTimeout
+	}))
```
OctoBoooot: praise: building a dedicated `rd.NewClient` for the readiness probe with `MaxRetries=0, PoolSize=2, DialTimeout/ReadTimeout/WriteTimeout=500ms` (rather than reusing the app-wide `ctx.GetRedisConn()`) is the right shape for a probe — it can't be starved by main-app connection-pool pressure, and it can't itself starve the main pool by sitting on a connection. The matching 500ms ceiling at every layer (handler `WithTimeout`, Redis dial/read/write) means a hung Redis path is bounded by exactly one budget, not three additive ones.

main.go
```diff
@@ -353,6 +357,8 @@ func ingorePaths() []string {
 	return []string{
 		"/v1/robots/:robot_id/:app_key/events",
 		"/v1/ping",
+		"/v1/health",
```
OctoBoooot, line 360: praise: `main_test.go` pins both path lists — `globalRateLimitExcludePaths()` MUST contain `/v1/health` but NOT `/v1/ready` (rate-limit asymmetry is the load-bearing invariant of this PR), and `ingorePaths()` MUST contain both. A future contributor adding `/v1/ready` to the exclude list (thinking "probes shouldn't be rate-limited") would fail `TestGlobalRateLimitExcludePathsIncludesProbeEndpoints` by name. That's the right place to put the test — at the seam where the policy is encoded, not at the handler.

OctoBoooot (dismissed):
## Review: fix: split health liveness from readiness (#437)

**Verdict:** Approve. Intentional contract split, well-scoped, fully tested. CI passes; risk tier normal; no blockers / majors / minors worth posting.

The old `/v1/health` had two real problems beyond what the PR body lists: (1) it returned HTTP 200 even when DB or Redis was down — clients had to parse the JSON body to detect failure, which is the opposite of what every container orchestrator + load balancer probe expects; and (2) it leaked `lastError.Error()` raw into the response body, which can include DSN/connection-string secrets. The new split addresses both — `/v1/health` is pure liveness (always 200 when the process is up), `/v1/ready` returns proper 503 on dep failure with the error stored under `json:"-"` (logged server-side only). The wire-contract break is intentional and documented in the PR body.

### Adversarial verification pass
- **Description-vs-diff** (per skill §7.1) — all claims match the diff: `globalRateLimitExcludePaths()` returns `[/v1/ping, /v1/health]` (excludes `/v1/ready` as documented); `ingorePaths()` adds both `/v1/health` and `/v1/ready`; `readinessProbeTimeout = 500*time.Millisecond` matches the "500ms" claim; the parallel MySQL `PingContext` + Redis `PING` are at lines 110-129 with proper `sync.WaitGroup` + mutex.
- **Timeout boundedness** — `WithTimeout(c.Request.Context(), 500ms)` creates ONE shared 500ms deadline for both pings; even if the checker ignored the ctx, the handler would still observe `<-ctx.Done()` at 500ms. `TestReadyBoundsCheckerWithTimeout` asserts `time.Since(started) < 1 second` with a checker that blocks on `<-ctx.Done()` — pins the handler-side bound, not just the checker-side one.
- **Concurrency** — `result.Dependencies`/`result.Errors` initial writes happen before goroutines launch (no race); goroutine writes are mutex-protected; `wg.Add(2)` + matching `defer wg.Done()` is the standard shape, not the off-by-one trap.

### Praise
- `modules/common/health_test.go:49` — `TestHealthIsPureLiveness` pins `checker.calls == 0` AND uses a fake whose response says \"down\" — a future contributor wiring the checker into `health` (copy-paste from `ready`) would fail by name AND would see liveness lying about dependency state in the assertion log.
- `modules/common/health_test.go:70-71` — `NotContains "secret" / "password"` on the response body, with a fake checker that puts a literal DSN-leak string into its errors, pins the safety contract concretely rather than relying on the `json:"-"` tag alone.
- `modules/common/health.go:34-50` — dedicated `rd.NewClient` for the readiness probe (`MaxRetries=0, PoolSize=2, 500ms timeouts`) can neither be starved by the main app pool nor starve it; the 500ms ceiling at every layer (handler `WithTimeout`, Redis dial/read/write) means a hung Redis path is bounded by exactly one budget, not three additive ones.
- `main_test.go` — pins both `globalRateLimitExcludePaths()` and `ingorePaths()` at the seam where policy is encoded, with the asymmetry (`/v1/ready` in access-log ignore but NOT in rate-limit exclude) explicitly asserted. A future "consolidate probe endpoints" PR can't silently regress the rate-limit gating.

- The wire-contract break (`{status, db, redis, error?}` → `{status}`) has no documented migration period. For in-tree consumers this is fine; if any external monitoring tool (Grafana, Pagerduty config, ops scripts) parses the legacy fields, they'll silently lose visibility until they cut over to `/v1/ready`. Worth a Slack/release-note callout to ops, not a code change. Note also: the legacy endpoint returned HTTP 200 on failure, so any tool branching on HTTP status was already misled — those should welcome the change.
- The dedicated readiness Redis client is created per `Common.New()` and never explicitly `Close()`d. Fine for the production singleton; if tests ever instantiate `Common` repeatedly, they leak Redis connection pools. Out of scope.

Jerry-Xin (changes requested):
The PR is in scope for octo-server, but `/v1/ready` is not actually bounded by the 500 ms request context in the Redis path.

**🔴 Blocking**
- 🔴 Critical: Redis readiness can exceed the intended 500 ms bound. In `modules/common/health.go:57-75`, the handler creates a 500 ms context, but `modules/common/health.go:148` calls `c.redisClient.WithContext(ctx).Ping().Result()` using go-redis v6. In this dependency version, `WithContext` only clones the client context; `Ping()` still processes the command without honoring that context for pool acquisition. The pool wait uses `PoolTimeout`, which defaults to `ReadTimeout + 1s`, so with `ReadTimeout = 500ms`, a saturated `PoolSize = 2` readiness Redis pool can wait up to about `1.5s` before even running the ping. Under concurrent probes or slow Redis, `/v1/ready` can exceed the claimed bounded readiness contract. Set an explicit bounded `PoolTimeout` and avoid relying on go-redis v6 context behavior for the hard timeout, or otherwise enforce the deadline around the Redis check.

**💬 Non-blocking**
- 🟡 Warning: `.octospec/tasks/health-readiness-split/brief.md` still says “Global Redis-backed rate limiting must not add Redis latency to health or readiness probe paths” in the load-bearing list, but the final implementation intentionally keeps `/v1/ready` rate limited. The acceptance section was updated, so this is only a spec consistency issue.

**✅ Highlights**
- `/v1/health` is now a pure liveness response and no longer calls the dependency checker.
- Readiness output avoids serializing raw dependency errors.
- Access-log suppression and rate-limit exclusion behavior are pinned with focused tests.

Verified locally:
- `go test . -run 'TestGlobalRateLimitExcludePathsIncludesProbeEndpoints|TestAccessLogIgnorePathsIncludesProbeEndpoints'`
- `go test ./modules/common -run 'TestHealthIsPureLiveness|TestReadyReturnsSafeDependencyStatus|TestReadyBoundsCheckerWithTimeout'`
- `go test ./tools/lint-direct-error-response`

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#437 Review Report — `fix: split health liveness from readiness`
**Head SHA:** `33a28a23eef3bcd86ba2b21d1286cccbf41a227e`
**Files changed:** 6 (+346 / -42)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `/v1/health` is pure liveness (no DB/Redis) | ✅ | `health.go:57-59` — handler returns `{"status":"up"}` without touching any dependency |
| `/v1/ready` checks DB+Redis with bounded timeout | ✅ | `health.go:61-88` — creates 500ms context, calls `checker.Check(ctx)` |
| DB ping uses `PingContext` (context-aware) | ✅ | `health.go:143` — `c.db.session.DB.PingContext(ctx)`; `session.DB` is `*sql.DB` via `dbr.Connection` embedding |
| Redis ping propagates context | ✅ | `health.go:149` — `c.redisClient.WithContext(ctx).Ping().Result()` |
| Parallel DB+Redis checks are race-safe | ✅ | `health.go:113-133` — `sync.Mutex` guards map writes, `sync.WaitGroup` for completion |
| Raw errors not leaked in response | ✅ | `health.go:25` — `Errors map[string]error json:"-"` excludes errors from JSON serialization |
| Rate-limit exclusion correct | ✅ | `main.go:44-46` — `globalRateLimitExcludePaths()` returns `["/v1/ping", "/v1/health"]`; middleware uses exact URL path match (`ratelimit.go:245`) |
| `/v1/ready` stays rate-limited | ✅ | Not in `globalRateLimitExcludePaths()`; confirmed by `main_test.go:14` |
| Access-log suppression covers both probes | ✅ | `main.go:359-362` — `ingorePaths()` includes both `/v1/health` and `/v1/ready` |
| Readiness Redis client uses same config | ✅ | `health.go:42` — `octoredis.MustBuildOptions(ctx.GetConfig(), ...)` reads same `cfg.DB.RedisAddr/RedisPass/RedisTLS` as all other clients |
| Tests cover key behaviors | ✅ | `health_test.go` — liveness purity, safe output, timeout bounding; `main_test.go` — path lists |
| lint-direct-error-response not triggered | ✅ | `health.go` uses `c.JSON()` not `c.AbortWithStatusJSON`; lint only counts the latter |

## 2. Findings

### F1 — P2: Dedicated Redis client never closed (resource lifecycle)

**File:** `modules/common/health.go:42-51`
**Diff-scope:** New (introduced by this PR)

`newDependencyReadinessChecker` creates a dedicated `rd.Client` via `rd.NewClient(...)`. This client is never explicitly closed — no `Close()` call on shutdown.

**Mitigation:** This is consistent with the existing codebase pattern. The rate-limiter Redis client in `main.go:183` and `ctx.GetRedisConn()` are also never explicitly closed (comment at `main.go:181` says "跟随进程存续，不显式 Close"). For a long-lived server process, this is acceptable.

**Recommendation:** No action needed for this PR. If the codebase ever adds graceful shutdown for Redis clients, this one should be included.

### F2 — P2: No test for the happy path (all dependencies up → 200)

**File:** `modules/common/health_test.go`
**Diff-scope:** New

The test suite covers: liveness purity, dependency-down → 503, and timeout bounding. Missing: a test where the fake checker returns all-up and the handler returns 200 with `{"status":"up","dependencies":{"db":"up","redis":"up"}}`.

**Mitigation:** The logic is trivially correct by inspection (`health.go:83-86` — status stays "up" when no errors). Low risk.

**Recommendation:** Consider adding a happy-path test for completeness, but not blocking.

### F3 — P2 (pre-existing, amplified): `/v1/health` response shape is a breaking change

**File:** `modules/common/api.go:88` (old inline handler removed)
**Diff-scope:** Amplified — the old handler returned dependency status; new handler returns only `{"status":"up"}`

Any consumer that relied on `/v1/health` returning `{"status":"up","db":"up","redis":"up"}` or `{"status":"down","error":"..."}` will silently lose dependency visibility. The PR body acknowledges this and directs consumers to `/v1/ready`.

1. New/amplified? **Amplified** — the endpoint existed but returned dependency data.
2. Does the diff change the behavior contract? **Yes** — response shape changes from dependency probe to pure liveness.
3. Base branch behavior identical? **No** — base returns dependency status, PR returns only liveness.

**Mitigation:** This is the intentional fix for #436. The PR body documents the migration path. No deployment configs in this repo reference `/v1/health` for dependency checking. The octo-deployment repo and frontend status tower would need separate verification.

**Recommendation:** Confirm that octo-web status tower and any k8s/Docker health checks have been (or will be) updated to use `/v1/ready` for dependency probing before merging.

## 3. Suggestions

1. **Happy-path test** (F2 above): Add a `TestReadyReturnsOKWhenAllDependenciesUp` test.
2. **Consider logging readiness state transitions**: When readiness flips from up→down or down→up, a single log line with the transition would help ops debug flapping dependencies. Currently every failed check logs individually, which could be noisy during a Redis outage.

## 4. Additional Findings

- **Positive:** The readiness checker creates a dedicated Redis client with `PoolSize: 2` and `MaxRetries: 0`, isolating probe traffic from application traffic. This is good design — probe load won't compete with business requests for connection pool slots.
- **Positive:** The 500ms timeout is applied at three layers: request context (`health.go:63`), Redis client timeouts (`health.go:45-47`), and DB `PingContext` inherits the context. This defense-in-depth ensures the handler never blocks beyond the bound.
- **Positive:** The `checker == nil` guard in `ready()` (line 66) is good defensive programming for test scenarios and edge cases where `New()` might not have been called.

## 5. Data Flow Tracing

|---|---|---|
| `c.db.session.DB.PingContext(ctx)` | `newDB(ctx.DB())` → `dbr.Session` embeds `*Connection` embeds `*sql.DB` | ✅ `*sql.DB.PingContext` exists and respects context cancellation |
| `c.redisClient.WithContext(ctx).Ping()` | `rd.NewClient(MustBuildOptions(cfg, ...))` — same `cfg.DB.RedisAddr` as `ctx.GetRedisConn()` | ✅ Connects to same Redis instance; `WithContext` propagates deadline |
| `readinessProbeTimeout` (500ms) | Constant at `health.go:20` → used in context creation (line 63) and Redis client timeouts (lines 45-47) | ✅ Consistent across all timeout points |
| `readinessResult.Errors` | Populated in goroutines (lines 118, 126) → consumed by `logReadinessErrors` (line 84) and excluded from JSON (`json:"-"`) | ✅ Never serialized to HTTP response |
| `globalRateLimitExcludePaths()` | `main.go:44-46` → consumed by `RateLimitMiddleware` (line 189) | ✅ Exact path match in middleware; `/v1/health` excluded, `/v1/ready` not excluded |

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths. The two endpoints (`/v1/health`, `/v1/ready`) serve different purposes by design.
- **C2 — Control-flow ordering / nested reuse:** Clear. The `Check()` method is called once per request with a fresh bounded context. No nested reuse. The `logReadinessErrors` helper is called once per `ready()` invocation.
- **C3 — Authorization boundary ≠ capability boundary:** `/v1/ready` is unauthenticated (same as old `/v1/health`). It can trigger DB/Redis pings, but this is bounded by rate limiting + 500ms timeout. No new capability exposure beyond what the old endpoint already provided. The old endpoint actually had *worse* behavior (no timeout bound).
- **C4 — Auth lifecycle / container-member cascade:** N/A. No auth changes in this PR.
- **C5 — Build ≠ runtime:** N/A. No build artifacts, extensions, CLI tools, or relative URL changes.
- **C6 — Governance/policy document consistency:** N/A. The `.octospec/tasks/health-readiness-split/brief.md` is a task spec, not a governance document. It is internally consistent with the implementation.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The implementation is clean, well-tested, and correctly separates liveness from readiness with proper timeout bounding, error sanitization, and rate-limit handling. Three P2 notes (resource lifecycle consistency, missing happy-path test, breaking change documentation) are non-blocking. The breaking change to `/v1/health` response shape is intentional and documented in the PR body.

Jerry-Xin (dismissed):
Re-reviewed at current head `dcb600e1` — my prior 🔴 (readiness response not bounded to 500ms on the go-redis v6 pool-wait path) is **genuinely fixed** (byte-verified, not a false-fix):

1. **Handler-level deadline enforcement** (`health.go:80-95`): `ready()` now runs `checker.Check(ctx)` in a goroutine feeding a buffered `resultCh` (cap 1, no goroutine leak), then `select`s between `<-resultCh` and `<-ctx.Done()`. So even if go-redis v6's `WithContext().Ping()` blocks on pool acquisition past 500ms, the handler returns 503 at `ctx.Done()` instead of blocking on the slow check. The 500ms response bound is now guaranteed at the handler regardless of v6 ctx semantics. The old `wg.Wait()` is now inside `Check()` (the detached goroutine), so it can't hold the response past the deadline.
2. **Defense-in-depth**: `o.PoolTimeout = readinessRedisPoolTimeout` (100ms, `health.go:54`) explicitly caps the pool-wait itself, replacing the ~1.5s default.

Everything else from the prior pass still holds: liveness purity (`/v1/health` zero dep calls), readiness fail-closed (503 on dep down), correct status codes, mutex-synced concurrency, info-leak protection (`Errors` with `json:"-"`).

No blocking items remain. ✅

yujiawei (dismissed):
# Code Review — PR #437 (octo-server)

Reviewed at head `33a28a23eef3bcd86ba2b21d1286cccbf41a227e` against `main`. Repo checked out at the exact head SHA; `go build ./...`, `go vet`, and all five new focused tests pass locally.

## Verdict

**APPROVED.** No P0/P1 blockers. The split is correct, the readiness path is properly bounded, dependency errors are redacted on the wire, and the rate-limit / access-log exclusions match the stated acceptance criteria. The items below are P2/advisory and do not block merge. A couple are worth a human glance because this touches infra probe contracts.

## 1. What was verified

- ✅ **`/v1/health` is pure liveness.** `cn.health` returns `200 {"status":"up"}` with no DB/Redis access (`modules/common/health.go:55`). Test `TestHealthIsPureLiveness` asserts the readiness checker is never invoked.
- ✅ **`/v1/ready` is bounded.** Handler wraps `c.Request.Context()` in a 500ms timeout (`modules/common/health.go:60-61`); DB `PingContext` and Redis `PING` run in parallel under a `WaitGroup` (`health.go:100-131`). The dedicated readiness Redis client also sets `Dial/Read/WriteTimeout = 500ms` and `MaxRetries = 0` (`health.go:42-48`), so the path is bounded by socket timeouts even independent of context propagation. `TestReadyBoundsCheckerWithTimeout` confirms the handler returns well under 1s.
- ✅ **Error redaction (security-relevant).** `readinessResult.Errors` is tagged `json:"-"` (`health.go:26`); only `up`/`down` strings reach the response. Raw errors are logged server-side via `logReadinessErrors` (`health.go:84-93`). `TestReadyReturnsSafeDependencyStatus` proves a DSN-with-password style error never reaches the body. This is actually a **net security improvement**: the old `/v1/health` returned `statusMap["error"] = lastError.Error()` to unauthenticated callers (old `modules/common/api.go`), which this PR removes.
- ✅ **Rate-limit exclusion matches correctly.** The global limiter matches on `c.Request.URL.Path` (octo-lib `wkhttp.RateLimitMiddleware`); for the literal routes `/v1/health` and `/v1/ready` (no path params) the request path equals the route, so `/v1/health` is correctly excluded and `/v1/ready` is correctly still limited.
- ✅ **Access-log ignore matches correctly.** `main.go:165-171` compares `c.FullPath()` against `ingorePaths()`. For statically registered routes `FullPath()` returns the route pattern, so both `/v1/health` and `/v1/ready` are correctly suppressed.
- ✅ **Cross-file wiring intact.** `globalRateLimitExcludePaths()` is passed variadically into `RateLimitMiddleware(... , excludePaths ...string)` (`main.go:189`) — signature matches. `Common` gains `readinessChecker` and `New` wires it (`modules/common/api.go:43-58`); build/vet clean.

## 2. Findings (all P2 / advisory)

**[P2] Keeping `/v1/ready` on the global rate limiter contradicts the brief's load-bearing wording, but the implementation choice is defensible.** `main.go:44` deliberately excludes only `/v1/ping` and `/v1/health`. The linked brief lists "global Redis-backed rate limiting must not add Redis latency to readiness probe paths," yet its own Acceptance section says `/v1/ready` keeps the global limiter because it touches dependencies. The implementation follows Acceptance. That is the right call: `/v1/ready` performs real DB/Redis work, so leaving it rate-limited protects against DoS amplification. The cost is that the limiter injects a Redis token-bucket round-trip ahead of the bounded 500ms check. Recommend documenting that `/v1/ready` is an infra readiness probe (low-frequency), not a high-QPS public poll, so operators don't point a tight LB/k8s probe interval at it and self-throttle.

**[P2] The readiness Redis client is never `Close()`d.** `newDependencyReadinessChecker` allocates an `rd.Client` (pool + background goroutines) and `readinessChecker` exposes no teardown (`health.go:39-49`). In production this is benign: `Common` is a process-level singleton (registered once via `register.AddModule` → `SetupAPI` in `modules/common/1module.go`), and this mirrors the existing, intentional "lifecycle follows the process, no explicit Close" pattern used for `rlRedis` in `main.go`. The only place `New` is called repeatedly is the test suite, which is short-lived. No change required; a `Close()` would only matter if `Common` ever becomes re-instantiated at runtime (dynamic reload / plugin reuse). Flagging for symmetry, not as a defect.

**[P2 / nit] Two independent 500ms bounds.** The handler context timeout and the readiness client's socket timeouts are both 500ms and both derive from `readinessProbeTimeout`. That is fine (defense in depth), just note they are coupled to the same constant — if someone later tunes one, they should reconsider the other.

## 3. Human-verify items (security-sensitive PR)

- **Probe contract change.** `/v1/health` no longer reflects DB/Redis health. Any external consumer (LB health checks, k8s readiness probes, monitoring) that historically treated `/v1/health` as a *dependency* signal must be repointed to `/v1/ready`. Deployment-manifest changes are explicitly out of scope for this PR, so this needs a coordinated rollout/ops check before or alongside merge.
- **go-redis v6 context semantics.** `redisClient.WithContext(ctx).Ping()` (`health.go:131`) relies on v6's partial context support; cancellation behavior is best-effort. The hard bound comes from the client's `ReadTimeout=500ms`, which holds regardless — acceptable, but worth a human nod since this is the timeout-safety guarantee.

## 4. Multi-source synthesis (audit trail)

Two advisory legs were consulted in addition to this review. Normalized to a P0/P1/P2 scale:

- **Consensus:** Exact path matching for the rate-limit and access-log exclusions is correct (both this review and the line-level advisory leg independently confirmed). Error redaction via `json:"-"` is correct.
- **Contradiction (resolved):** One advisory leg raised a **P1** that `c.db.session.DB.PingContext` may be a compile error if `DB` is an ORM method rather than a field. **Refuted with evidence:** dbr's `Session` embeds `*Connection`, which embeds `*sql.DB`, so `.DB` is a promoted `*sql.DB` field and `PingContext` is its standard method. `go build ./...` and `go vet` are clean at this head. Downgraded to non-finding.
- **Unique:** The brief-vs-implementation tension on `/v1/ready` rate limiting (kept above as P2). The pre-existing info-leak that the old `/v1/health` exposed and this PR removes (kept above as a positive).
- **Blind spots / coverage:** External probe consumers and deployment manifests (out of repo, flagged for human); go-redis v6 context enforcement (mitigated by socket timeouts); behavior of `octoredis.MustBuildOptions` under TLS config (unchanged by this PR, not re-audited).

One advisory leg returned no findings but did not supply a coverage note; the coverage gaps above are supplied to keep both legs symmetric.

Jerry-Xin (dismissed):
Re-affirming APPROVE at current head `f38fcf66` (prior APPROVE @dcb600e1 auto-dismissed by a new commit — `ahead 1`, a genuine follow-up commit, not a rebase).

The only delta dcb600e1→f38fcf66 (per-file blob: only `health.go` differs; api.go/main.go byte-SAME) is the `/v1/health` liveness response gaining static `db:up`/`redis:up` keys for backward-compat — addressing the earlier response-format 🟡 so consumers parsing those keys don't break. These are constant (liveness stays pure, no dep probing), so it doesn't reintroduce any dep coupling.

Byte-verified my prior 🔴 fix is untouched/intact: `PoolTimeout = readinessRedisPoolTimeout` (100ms) at :54, and the handler-level `select` on `<-ctx.Done()` (:90-92) with `checker.Check(ctx)` in a buffered goroutine — readiness stays 500ms-bounded regardless of go-redis v6 pool semantics.

No blocking items remain. ✅

modules/common/health.go
```diff
@@ -0,0 +1,176 @@
...
+		resultCh <- checker.Check(ctx)
+	}()
+
+	var result readinessResult
+	select {
+	case result = <-resultCh:
+	case <-ctx.Done():
+		result = readinessResult{
+			Status: healthStatusDown,
+			Dependencies: map[string]string{
```
OctoBoooot, line 96: praise: the `goroutine + select on resultCh vs ctx.Done()` pattern (lines 84-99) is the right shape — handler returns within `readinessProbeTimeout` even if the underlying checker (or, importantly, go-redis v6's pool-acquisition wait) doesn't honor ctx. `resultCh` is buffered cap-1 so the goroutine never leaks when the timeout branch wins. The test `TestReadyReturnsWhenCheckerIgnoresContext` uses literal `time.Sleep(readinessProbeTimeout * 2)` to construct exactly the failure mode Steve flagged — fake checker that ignores ctx and sleeps past the budget — and pins handler-return `< 750ms`. That's the right place to lock the deadline: at the handler, not at the dependency.

modules/common/health.go
```diff
@@ -0,0 +1,176 @@
...
+		db:          db,
+		redisClient: rd.NewClient(readinessRedisOptions(ctx.GetConfig())),
+	}
+}
+
+func readinessRedisOptions(cfg *config.Config) *rd.Options {
+	return octoredis.MustBuildOptions(cfg, func(o *rd.Options) {
+		o.MaxRetries = 0
+		o.PoolSize = 2
+		o.DialTimeout = readinessProbeTimeout
```
OctoBoooot, line 51: praise: `o.PoolTimeout = readinessRedisPoolTimeout` (100ms) explicitly bounds connection-pool wait — without it, go-redis v6's default `ReadTimeout + 1s = 1.5s` would dominate the 500ms request budget under `PoolSize=2` contention. Defense-in-depth alongside the handler-level select. `TestReadinessRedisOptionsUseBoundedPoolTimeout` pins the invariant `PoolTimeout < probe-timeout` so a future contributor can't accidentally bump PoolTimeout above the probe budget and undo the bound. Also: factoring `readinessRedisOptions` out as a standalone function specifically to make it testable is a small ergonomics win.

modules/common/health.go
```diff
@@ -0,0 +1,176 @@
...
+		o.PoolTimeout = readinessRedisPoolTimeout
+	})
+}
+
+func (cn *Common) health(c *wkhttp.Context) {
+	c.JSON(http.StatusOK, map[string]string{
+		"status": healthStatusUp,
+		"db":     healthStatusUp,
+		"redis":  healthStatusUp,
+	})
```
OctoBoooot, line 64: **[minor]** `"db":"up", "redis":"up"` are static literals here — intentionally so, preserving wire compat with legacy `/v1/health` parsers (mochashanyao's 🟡 fix). But there's no inline comment saying so, and a future contributor reading just this block could reasonably think these *should* reflect real dependency state and wire them back through the checker — re-creating the synchronous coupling the PR is trying to remove.

`TestHealthIsPureLiveness` catches the re-coupling if it goes through `cn.readinessChecker` (the `calls == 0` assertion would fail), but NOT if a contributor calls `cn.db.Ping()` directly inline. A one-line comment — e.g., `// STATIC fields for legacy /v1/health body-shape compat; liveness MUST stay pure — see #437` — would prevent the well-intentioned wrong fix.

OctoBoooot (dismissed):
## Review: fix: split health liveness from readiness (#437) — delta review

**Verdict:** Approve. The PoolTimeout / handler-deadline gap I missed in my prior APPROVE at `33a28a2` is genuinely fixed; mochashanyao's wire-compat 🟡 is also addressed. One small comment-clarity minor.

Honest acknowledgment of my prior miss: my first review at `33a28a2` praised the "500ms ceiling at every layer" without verifying that go-redis v6's `WithContext().Ping()` actually honors ctx during pool acquisition (it doesn't — v6 is ctx-aware only at the command boundary, not the connection-pool wait), and without noticing `PoolTimeout` was unset (default `ReadTimeout+1s = 1.5s`). Under `PoolSize=2` contention the handler's `wg.Wait()` could block past the 500ms budget. Steve caught this; the author fixed it; my praise rule §7.1.1 should have caught this but I checked three timeouts and stopped. Lesson reinforced in memory + SKILL.

Delta scope (`33a28a23` → `f38fcf66`): only `modules/common/health.go` and `health_test.go` changed (per-file blob compare). CI fully green at this head (Build/Test/Lint/Vet all pass, check-sprint clear).

### Resolved from prior round
- **🔴 PoolTimeout missing + handler bound depends on v6 ctx semantics** (Steve / me missed): two-layer fix —
  - `o.PoolTimeout = readinessRedisPoolTimeout (100ms)` explicitly bounds pool wait below the probe budget. `TestReadinessRedisOptionsUseBoundedPoolTimeout` pins the relationship.
  - Handler now wraps `checker.Check(ctx)` in a goroutine writing to a buffered cap-1 `resultCh`, then `select`s on `resultCh` vs `ctx.Done()` — handler returns 503 at exactly `readinessProbeTimeout` regardless of whether the checker honors ctx. `TestReadyReturnsWhenCheckerIgnoresContext` constructs the literal failure mode (`time.Sleep(2×timeout)`) and pins `< 750ms` handler return.
- **🟡 /v1/health wire-format break** (mochashanyao): response is now `{"status":"up","db":"up","redis":"up"}` — static literals (no actual checks), preserving wire-shape for legacy parsers that look for `db`/`redis` keys. `TestHealthIsPureLiveness` updated and still asserts `checker.calls == 0` so the static fields can't be silently re-wired through the checker path.

### Minor
- `modules/common/health.go:60-62` — the static `"db":"up"`/`"redis":"up"` literals have no inline comment explaining they're intentionally static for legacy wire-compat; a future contributor could reasonably think they should reflect real state and re-wire them inline (the test catches re-coupling via `cn.readinessChecker` but not via direct `cn.db.Ping()`). One-line comment would prevent the well-intentioned wrong fix.

### Praise
- `modules/common/health.go:84-99` — handler-level `select` on `resultCh` vs `ctx.Done()` is the right place to lock the deadline (not at the dependency); `TestReadyReturnsWhenCheckerIgnoresContext` pins the exact v6-ctx-violator scenario.
- `modules/common/health.go:46-56` + `health_test.go` — `readinessRedisOptions` factored out for testability, with explicit `Less(opts.PoolTimeout, readinessProbeTimeout)` test pinning the relationship so a future config tweak can't accidentally re-introduce the regression.
- yujiawei's follow-up 🟡 on Redis-client `Close()` matches an out-of-scope note from my first round; non-blocking and low-impact under the production singleton.

- The wire-contract compat is now via static literals — that works for the "legacy parsers look at body keys" case, but if any external client also branched on HTTP status code being 200 even when dep down, they'd silently keep that broken assumption. Documented in the PR body as intentional; not a code change.

OctoBoooot (approved):
## Review: fix: split health liveness from readiness (#437) — delta review

**Verdict:** Approve. Delta (`f38fcf66` → `5693067a`) is one line added: the inline comment I asked for in my prior minor (`modules/common/health.go:59` — `// db/redis are static legacy-shape fields; liveness must stay dependency-free.`). Per-file blob compare confirms only `health.go` changed; `health_test.go`/`api.go`/`main.go`/`main_test.go` are byte-identical.

The comment closes the minor cleanly: a future contributor reading just the `health()` block now sees both the static-by-design intent AND the "must stay dependency-free" invariant the test pinning enforces. No code behaviour changed; no new tests needed; everything else from my prior APPROVE still holds verbatim (handler-level `select` on `ctx.Done()`, `PoolTimeout=100ms`, all the regression tests).

Jerry-Xin (approved):
Re-affirming APPROVE at current head `5693067a` (prior APPROVE @f38fcf66 auto-dismissed by a new commit — `ahead 1`, follow-up commit, not a rebase).

Per-file blob compare f38fcf66→5693067a: only `health.go` differs (api.go/main.go/health_test.go byte-SAME). The entire delta is a **single comment line** on the `health()` liveness handler — `// db/redis are static legacy-shape fields; liveness must stay dependency-free.` — addressing the prior 🟡 (prevents a future contributor from re-coupling the static fields to real dep pings). Zero behavior change.

Byte-verified my prior 🔴 fix branches are untouched/intact: `PoolTimeout = readinessRedisPoolTimeout` (100ms), handler-level `select` on `<-ctx.Done()` with `checker.Check(ctx)` in a buffered goroutine — readiness stays 500ms-bounded.

No blocking items remain. ✅

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#437 Review Report
**PR:** Mininglamp-OSS/octo-server #437 — fix: split health liveness from readiness
**Head SHA:** `5693067a5abe8f6719b7635a420eecc144d900f7`
**Changed files:** 6 (+405 / −42)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `/v1/health` is pure liveness (no DB/Redis) | ✅ | `health.go:63-68` — static response, no dependency calls |
| `/v1/ready` checks DB+Redis in parallel with bounded ctx | ✅ | `health.go:120-155` — `sync.WaitGroup` + `context.WithTimeout(500ms)` |
| Raw dependency errors not leaked in response | ✅ | `health.go:28` — `Errors json:"-"`, confirmed by `TestReadyReturnsSafeDependencyStatus` |
| `/v1/health` excluded from global rate limiter | ✅ | `main.go:45` — `globalRateLimitExcludePaths()` includes `/v1/health`; `main_test.go:14` pins it |
| `/v1/ready` stays rate-limited | ✅ | `main_test.go:14` — `require.NotContains(t, paths, "/v1/ready")` |
| Both probes excluded from access logs | ✅ | `main.go:360-361` — both in `ingorePaths()`; `main_test.go:22` pins it |
| Legacy response shape preserved on `/v1/health` | ✅ | `health.go:65-68` — `{"status":"up","db":"up","redis":"up"}` |
| Readiness Redis client uses bounded pool | ✅ | `health.go:48-55` — PoolSize=2, DialTimeout=500ms, PoolTimeout=100ms |
| Tests cover liveness purity, readiness safety, timeout bounding | ✅ | `health_test.go` — 5 focused tests |

## 2. Findings

### F1 — P2: Response shape change between old `/v1/health` and new `/v1/ready`

**Diff-scope:** new (this PR deliberately changes the contract)

Old `/v1/health` returned a flat body with real dependency status:
```json
{"status":"down","db":"down","redis":"up","error":"dial tcp ..."}
```
New `/v1/ready` returns a nested body with safe status only:
```json
{"status":"down","dependencies":{"db":"down","redis":"up"}}
```

Consumers parsing `body.db`, `body.redis`, or `body.error` on `/v1/health` will break when migrating to `/v1/ready`. The `error` field is intentionally removed (security improvement — prevents DSN/connection-string leaks).

**Severity rationale (R1/R2):** This is a deliberate, documented contract change, not a bug. The old endpoint still exists at the same path with the same response shape (static values). Consumers that need real dependency status must migrate to `/v1/ready` with a different response schema. P2 — migration required but well-documented in PR body and spec.

### F2 — P2: Goroutine lifecycle when checker ignores context

**Diff-scope:** new (introduced by this PR's `ready` handler design)

In `health.go:89-92`, the checker runs in a goroutine. When the 500ms context fires first (`health.go:100-107`), the handler returns but the goroutine lingers until the checker finishes. The buffered channel (`make(chan readinessResult, 1)` at line 88) prevents the goroutine from blocking forever — it sends to the buffer and exits.

The real `dependencyReadinessChecker.Check()` uses context-aware pings (`sql.DB.PingContext(ctx)` at line 168, `redisClient.WithContext(ctx).Ping()` at line 174), so lingering is brief. But a custom `readinessChecker` implementation that ignores context could keep the goroutine alive indefinitely.

**Severity rationale (R1):** Not user-visible, not a resource leak under realistic probe frequencies. The test `TestReadyReturnsWhenCheckerIgnoresContext` explicitly validates the handler returns bounded. P2 — defensive observation, not a blocker.

### F3 — P2: Readiness Redis client never explicitly closed

`health.go:38` creates a `rd.NewClient(readinessRedisOptions(...))` that is never `Close()`-d. This adds one more Redis connection pool to the process (the codebase already has 7+ separate pools — see data-flow trace below).

**Severity rationale:** Matches the existing codebase pattern (rate-limiter Redis, OIDC stores, etc. — none are explicitly closed). P2 — consistency with existing practice, not a regression.

## 3. Suggestions

1. **Consider documenting the `/v1/ready` response schema** in the spec or an OpenAPI doc, since it differs from the legacy `/v1/health` shape. Consumers migrating from `/v1/health` body-parsing will need the new schema.
2. **Consider a startup log line** confirming readiness checker initialization, so operators can verify the endpoint is wired (similar to the metrics endpoint's "starting metrics scrape endpoint" log at `main.go:296`).

## 4. Additional Observations

- **Redis client proliferation:** The process now has 8+ separate Redis connection pools to the same instance. Each serves a valid isolation purpose, but the aggregate connection count grows. Not a blocker — just an observation for future consolidation consideration.
- **DB pool asymmetry:** The readiness checker creates a dedicated Redis client (PoolSize=2) but shares the application's `sql.DB` pool for `PingContext`. Under extreme `/v1/ready` polling, DB pings would compete with normal queries for pool connections. The 500ms context timeout bounds this, and the global rate limiter still applies to `/v1/ready`.

## 5. Data-Flow Tracing

### `/v1/health` handler (health.go:63-68)
- **Consumed data:** None — static string constants `healthStatusUp`
- **Upstream source:** Compile-time constants (health.go:20)
- **Runtime verification:** No dependency calls. `cn.readinessChecker` is never accessed. Test `TestHealthIsPureLiveness` confirms `checker.calls == 0`.

### `/v1/ready` handler (health.go:71-115)
- **Consumed data:** `cn.readinessChecker` → `readinessChecker.Check(ctx)` → `readinessResult`
- **Upstream source:** `newDependencyReadinessChecker(ctx, database)` called in `common.New()` (api.go:55)
  - `database` = `newDB(ctx.DB())` → `*dbr.Session` → embeds `*sql.DB` (standard library)
  - `redisClient` = `rd.NewClient(readinessRedisOptions(cfg))` → separate pool, bounded timeouts
- **Data flow verification:**
  - `Check()` (health.go:120-155) spawns 2 goroutines: `pingDB(ctx)` and `pingRedis(ctx)`
  - `pingDB` nil-checks `c.db.session.DB` before calling `PingContext(ctx)` (line 165-168) — safe
  - `pingRedis` nil-checks `c.redisClient` before calling `WithContext(ctx).Ping()` (line 172-175) — safe
  - Results merged under `sync.Mutex`, `wg.Wait()` ensures all goroutines complete before return
  - `Errors` field tagged `json:"-"` — confirmed not serialized (test asserts no "secret"/"password" in body)
  - Context timeout path (line 100-107) constructs a safe fallback result with all-down status

### Rate-limit middleware (main.go:189)
- **Exclude paths:** `globalRateLimitExcludePaths()` → `["/v1/ping", "/v1/health"]`
- **Match mechanism:** Exact string match on `c.Request.URL.Path` via `map[string]struct{}` lookup (octo-lib `ratelimit.go:218+`)
- **Verification:** `/v1/health` excluded ✅, `/v1/ready` NOT excluded ✅ (confirmed by test)

### Access-log suppression (main.go:195-202)
- **Ignore paths:** `ingorePaths()` → `["/v1/robots/:robot_id/:app_key/events", "/v1/ping", "/v1/health", "/v1/ready"]`
- **Match mechanism:** Exact string match on `c.FullPath()` (gin's resolved route pattern)
- **Verification:** Both `/v1/health` and `/v1/ready` excluded ✅ (confirmed by test)

**C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths in this PR.

**C2 — Control-flow ordering / nested reuse:** Clear.
- `logReadinessErrors` is called in exactly 2 places (nil-checker early return at line 84, normal path at line 111). Both are mutually exclusive branches — no double-call risk.
- The `readinessChecker.Check()` method uses `sync.Mutex` for concurrent map writes and `sync.WaitGroup` for completion. No nested/recursive call paths.

**C3 — Authorization boundary ≠ capability boundary:** Clear.
- Both `/v1/health` and `/v1/ready` are registered on `r.GET(...)` (api.go:88-89), outside any auth middleware group. This is intentional for infrastructure probes — same as the old `/v1/health`.
- `/v1/ready` exposes only safe `up`/`down` status strings. Raw errors are excluded from the response (`json:"-"`). No sensitive data (DSNs, passwords, internal IPs) is leaked.
- `/v1/ready` remains behind the global per-IP rate limiter (500 rps default), preventing resource amplification.

**C4 — Authorization lifecycle / container-member state cascade:** N/A. No authorization changes in this PR.

**C5 — Build/note ≠ runtime path:** N/A. Server-side Go binary, no browser extension / CLI / relative-URL concerns.

**C6 — Governance / policy / security document consistency:** N/A. The `.octospec/tasks/health-readiness-split/brief.md` is a task spec, not a governance document. No policy conflicts.

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. The PR correctly splits liveness from readiness with clean data flow, proper context bounding, safe error handling (no raw error leakage), and focused test coverage. Three P2 observations (response shape migration, goroutine lifecycle on context-ignore, Redis client lifecycle) are noted but do not block landing.

modules/common/health.go
an9xyz: Handled in commit 5693067a: /v1/health now keeps db/redis as static legacy-shape fields, and the inline comment explicitly states liveness must stay dependency-free.

yujiawei (approved):
# Code Review — PR #437 (octo-server)

**Scope:** Splits `/v1/health` (cheap, dependency-free process liveness) from a new `/v1/ready` (bounded DB+Redis readiness probe); excludes `/v1/health` from the global per-IP rate limiter; excludes both `/v1/health` and `/v1/ready` from access logging. Reviewed against merge-base `421ecd7` at head `5693067`.

**Verdict: APPROVE.** No P0/P1. The design is sound, the security-sensitive aspects (no secret leakage, readiness still rate-limited, bounded timeouts) hold up, and the change matches its stated acceptance criteria. A couple of non-blocking notes below.

## Verification

- ✅ **`/v1/health` is pure liveness** — `modules/common/health.go:60-67` returns a static `{"status":"up","db":"up","redis":"up"}` and never touches DB/Redis. Test `TestHealthIsPureLiveness` asserts the dependency checker is called 0 times.
- ✅ **`/v1/ready` is bounded and returns 503 on dependency failure** — `health.go:69-117`. Three layers of bounding: handler-level `context.WithTimeout(..., 500ms)` (`health.go:70`), the `select` on `ctx.Done()` (`health.go:90-100`) which returns even if a checker ignores `ctx` (proven by `TestReadyReturnsWhenCheckerIgnoresContext`), and per-dependency redis dial/read/write/pool timeouts (`health.go:55-62`). `readinessRedisPoolTimeout (100ms) < readinessProbeTimeout (500ms)` is asserted by test.
- ✅ **No secret/internal leakage in the readiness body** — `readinessResult.Errors` is tagged `json:"-"` (`health.go:33`), so raw errors (DSNs, passwords) are logged server-side via `logReadinessErrors` (`health.go:119-128`) but never serialized. `TestReadyReturnsSafeDependencyStatus` injects an error string containing `password`/`secret` and asserts neither appears in the wire body. Only `up`/`down` status fields are exposed. This directly satisfies the security ask.
- ✅ **`/v1/ready` keeps the global rate limiter; only `/v1/health` is excluded** — `globalRateLimitExcludePaths()` returns `{"/v1/ping","/v1/health"}` (`main.go:44-46`); `/v1/ready` is deliberately absent and `TestGlobalRateLimitExcludePathsIncludesProbeEndpoints` asserts `NotContains(paths, "/v1/ready")`. Excluding the *static* `/v1/health` from the limiter is not a meaningful DoS amplifier — the handler does zero I/O and allocates a tiny constant map; it is far cheaper to serve than to rate-limit. The expensive endpoint stays protected.
- ✅ **Concurrent map writes are correctly guarded** — `dependencyReadinessChecker.Check` (`health.go:130-167`) runs DB and Redis pings in two goroutines, each writing `result.Dependencies`/`result.Errors` only under `mu.Lock()`. `go test -race` on the health + main tests passes clean.
- ✅ **No goroutine leak on timeout** — `resultCh` is buffered (size 1, `health.go:85`); if the handler returns via `ctx.Done()` first, the worker goroutine's send still succeeds and the goroutine exits rather than blocking forever.
- ✅ **Legacy response shape preserved** — `/v1/health` still returns `200 {"status":"up","db":"up","redis":"up"}`, so existing LB/probe parsers that only check HTTP 200 or `status==up` are unaffected.
- ✅ **Build + focused tests pass** — `go build ./...` clean; `go vet` clean; all PR-added tests pass (one unrelated pre-existing failure, `TestManagerSystemSetting_*`, requires live DB migrations and is independent of this diff).

## Notes (non-blocking)

**N1 (P2, maintainability) — readiness Redis client is never closed.** `newDependencyReadinessChecker` creates a dedicated `rd.NewClient(...)` (`health.go:43`) with no corresponding `Close()`. In normal operation this is fine: the `common` module is instantiated once at startup via `register.AddModule` → `New(ctx)` (`modules/common/1module.go:16-26`), and the client lives for the process lifetime — the same lifecycle pattern already used by the global `rlRedis` client in `main.go:185`. It only matters if `Common` is ever re-instantiated repeatedly within one process (e.g. some future hot-reload path or per-test construction), where connections/goroutines would accumulate. Worth a one-line comment documenting the singleton assumption, or wiring a `Close()` if a teardown hook exists. Not a merge blocker.

**N2 (informational) — semantic change to `/v1/health` fields.** `db`/`redis` in the liveness body are now hardcoded `"up"` and no longer reflect actual dependency state — this is intended (the brief's load-bearing list calls it out), but any external monitor that today *parses those two fields* to infer DB/Redis health will silently go green. Such monitors should be repointed to `/v1/ready`. Flagging for operational awareness, not as a code defect.

## Advisory legs (security-sensitive PR)

Per the multi-source advisory protocol for security-sensitive changes, two independent advisor legs were attempted:

- **Long-context advisor (gemini-3.1-pro):** completed. Raised one P1 — a claimed compile break at `health.go:169-172` on the theory that `session.DB` might be an ORM *method* rather than a field. **Adjudicated as a false positive:** `db.session` is `*gocraft/dbr.Session`, whose embedded `*Connection` embeds `*sql.DB`, making `session.DB` a promoted `*sql.DB` field — `DB == nil` and `DB.PingContext(ctx)` are valid, and both `go build` and `go test -race` pass. The advisor's own coverage note flagged this as unverified from the diff alone. Its P2 (unclosed readiness Redis client) matches **N1** above.
- **Line-level adversarial advisor:** **absent** — the run hung with no output and was terminated after ~35 min. Per protocol an absent leg is **not** treated as silent approval; its intended coverage (line-level correctness, leak/race checks) was instead covered by this reviewer's independent analysis plus the `-race` run.

**Coverage / blind spots:** This review verified the diff, the immediate call sites, the dbr/sql type chain, the rate-limiter exclude semantics, and ran the focused tests with the race detector. Not verified here: live behavior under real DB/Redis outage in a deployed environment, and k8s/deployment probe manifests (explicitly out of scope in the task brief) — operators adopting `/v1/ready` as a k8s readiness probe should confirm probe wiring separately.

PR #442, opened 2026-06-23
feat(metrics): expose upstream-dependency metrics (DB/Redis pool + object-storage latency)

an9xyz:
## Summary

Make the dependencies *below* the HTTP handler observable, so request latency can
be attributed instead of guessed. Today the HTTP entrypoint is measured
(`dmwork_http_*`) but DB/Redis connection-pool saturation and object-storage call
latency are a black box. This is **P0-a**: the parts that are self-contained in
this repo — no `octo-lib` change, no handler/business-logic change.

## Related Issue

Part of #440 (P0-a). P0-b items remain tracked there.

## Linked Spec

`.octospec/tasks/upstream-dep-metrics/brief.md` (+ `context.yaml`, journal in
`.octospec/journal/shared/upstream-dep-metrics.md`).

## Changes

- **Dependency-latency histogram** (`pkg/metrics/dependency.go`):
  `dmwork_dependency_duration_seconds{dependency,op,backend,status}`, buckets down
  to 1ms. Exposed via a nil-safe package-level `ObserveObjectStore` so callers
  don't thread a metrics instance through.
- **Object-storage timing**: `file.Service.DownloadURL` (`modules/file/service.go`)
  wrapped transparently — return value + error propagated unchanged; `backend`
  label captured in the `NewService` switch (minio/oss/qiniu/cos/s3/seaweedfs).
- **Connection-pool metrics** (`pkg/metrics/pool.go`), scrape-time collectors,
  **no background goroutine**:
  - DB via `collectors.NewDBStatsCollector(ctx.DB().DB,"main")` → standard
    `go_sql_*` (incl. `wait_count` / `wait_duration_seconds_total`).
  - Redis via a custom collector reading the rate-limit `*redis.Client`'s
    `PoolStats()` → `dmwork_redis_pool_*` (total/idle gauges; hits/misses/
    timeouts/stale counters).
- Registered on `prometheus.DefaultRegisterer` in `main.go` (after `rlRedis` is
  built), served by the existing `/metrics` endpoint.

### Notes on scope refinements (see #440 comment)
- Pool metrics use **scrape-time Collectors, not a background sampler** — fresher
  and no goroutine-lifecycle/leak concern.
- DB pool uses **standard `go_sql_*`**, not hand-rolled `dmwork_db_pool_*`.
- Redis pool covers only the repo-owned `*redis.Client`; `octo-lib`'s
  `GetRedisConn()` returns a `*redis.Conn` with no `PoolStats()`, unreachable
  without a lib change.

[...]

PR #145, opened 2026-05-23
fix(mention): expand mention.ais=1 into mention.uids at ingress chokepoint (#144)

yujiawei:
## What

When `mention.ais=1` in a GROUP channel, expand `mention.uids` to include every bot member of the channel — applied at the three message ingress chokepoints (`modules/message/api.go`, `modules/bot_api/send.go`, `modules/robot/api.go`), immediately after `RewriteMention`.

## Why

Old adapter bots (#137) connected via the WuKongIM websocket only inspect `mention.uids` and ignore `mention.ais`. PR #138's per-bot UID injection (`injectBotUIDIntoMentionUIDs`) only rewrites the payload that reaches the bot event queue (`/v1/bot/events`), never the websocket dispatch path. Expanding at the ingress makes the WuKongIM payload itself carry the bot UIDs, so legacy adapters recognise the `@所有 AI` broadcast.

## Changes

- `robot.IService` gains `ExistRobot(uid) (bool, error)` — backed by the existing `robotDB.exist` `status=1` lookup, implemented on both `Service` and `*Robot`.
- New `pkg/mentionrewrite.ExpandAisToBotUIDs` helper. Pure / idempotent / defensive on `nil` & malformed shapes. Takes a `fetchBotUIDs` callback so `pkg/mentionrewrite` stays a leaf package (preserves the historical `robot → message → robot` cycle-free invariant called out in `pkg/mentionrewrite/rewrite.go`).
- Each of the three chokepoints composes `group.GetMembers + robot.ExistRobot` into the callback shape via a tiny module-local `mention_expand.go` helper, invoked AFTER `RewriteMention`.
- PERSONAL DMs and COMMUNITY_TOPIC channels are out of scope (no-op).
- `mention.all` / `mention.humans` / `mention.ais` are forwarded untouched — only `mention.uids` is mutated.
- Dedups within a single call AND with PR #138's `injectBotUIDIntoMentionUIDs` so a payload cannot accumulate duplicates.

## Tests

`pkg/mentionrewrite/expand_ais_test.go` pins every clause from the helper's contract:

- `ais+GROUP` expands · `ais+PERSONAL` no-op · `ais+COMMUNITY_TOPIC` no-op
- `all`-only no-op · `humans`-only no-op · `ais=0` no-op
- dedup (bot already in `uids`, fetcher returns dup, mixed bots+humans)
- idempotency (second call is observably a no-op)
- `nil` payload / empty mention / non-map mention / non-array uids — all safe
[...]

PR #445, opened 2026-06-23
feat(incomingwebhook): mention support on the native push endpoint

an9xyz:
## Summary

Adds opt-in mention support to the **native** incoming-webhook push endpoint. A
caller can mention specific group members (users and/or bots, one or many) via a
structured `mention` field, and optionally broadcast to all human members and/or
all bot (AI) members of the group. Notifications (red-dots) and bot delivery are
produced downstream by the existing sender-agnostic message listener — this
change only assembles a well-formed `mention` sub-map on the wire.

## Related Issue

Internal feature request (no public issue).

## Linked Spec

`.octospec/tasks/incoming-webhook-mention/brief.md` (goal / load-bearing list /
out-of-scope / acceptance). Rule context in
`.octospec/tasks/incoming-webhook-mention/context.yaml`.

## Changes

- Native payload gains `mention: { uids: [...], all: bool, bots: bool }` — parsed
  only by the native adapter; wecom/feishu/github/gitlab/multica neither parse
  nor emit it (`pushAdapter.allowMention`).
- `mention.uids` (users + bots, one common UID namespace) are filtered to current
  group members (space-isolation boundary), deduped, and capped (default 50,
  `OCTO_INCOMINGWEBHOOK_MAX_MENTION_UIDS`).
- The two broadcast forms map to `mention.humans` / `mention.ais` and are gated by
  per-webhook capability columns `allow_mention_all` / `allow_mention_bots`
  (default off; settable on create/update by any legitimate member who manages the
  webhook). When a broadcast is requested but the capability is off, it is dropped
  and reported in the 200 response body's `mention_ignored`. A targeted member
  mention is not capability-gated.
- The all-bots broadcast reuses `pkg/mentionrewrite.ExpandAisToBotUIDs` with the
  same `GetMembers` + `ExistRobot` pair as the message / bot_api ingresses, so the
  resolved bot set is identical across ingresses.
- Pre-wired revocation lever: system setting `incomingwebhook.member_can_broadcast`
  (default on), ANDed at the push read path — turning it off instantly revokes
  broadcast for all member-created webhooks (admin-created ones keep it), with no
  column migration.
- `mention` is an explicitly whitelisted payload field; `extra` stays discarded,
  so no other payload keys can be smuggled in.
- New SQL migration adds the two `SMALLINT` columns (default 0, INSTANT add).

## Testing

Verified against MySQL 8.0 + Redis + WuKongIM (v2.2.4-20260313):

- `go test ./modules/incomingwebhook/... -count=1` → ok
- `go test ./modules/common/... -count=1` → ok
- `golangci-lint run ./modules/common/... ./modules/incomingwebhook/...` → 0 issues
- `make i18n-lint` → OK (no new error codes)

- [x] Unit tests added/updated (decision-core matrix, dedup, ExpandAisToBotUIDs parity, native-body parse, system-setting getters)
- [x] e2e tests added (capability persistence, member-settable switch, revocation lever, `mention_ignored`, backward-compat)
- [x] Manually verified end-to-end (push path through WuKongIM)

## COMPREHENSION

   On the unauthenticated webhook push path, after the existing
   auth / rate-limit / group-active / creator-membership gates, the native handler
   assembles a `mention` sub-map and attaches it to the outbound payload, then runs
   the ais→bot-UID expansion. Targeted UIDs are intersected with the webhook's own
   `group_no` membership; a broadcast takes effect only when both the per-webhook
   capability bit and the `member_can_broadcast` policy (or an admin creator)
   allow it. The message module's reminder/read path is unchanged.

   - Existing native callers that send no `mention` field: payload shape is
     unchanged (pinned by a backward-compat test).
   - A leaked/misconfigured token cannot mention outside its group or cross-Space —
     the membership filter uses the server-side `group_no`, never caller input.
   - The bot-set lookup for the all-bots broadcast degrades to a no-op on error
     (best-effort), so it never drops a message.
   - Sibling adapters are unaffected (mention is gated to the native adapter).

   Pure unit tests exhaustively cover the assemble/gate matrix and prove the wire
   shape is consumable by the real `ExpandAisToBotUIDs`. e2e tests drive the full
   HTTP→WuKongIM path: capability persistence + member-settable switch, the
   `member_can_broadcast` revocation (member-created webhook's broadcast stripped
   and reported, admin-created unaffected), `mention_ignored` on capability-off,
   and the no-mention backward-compat case. Full `incomingwebhook` and `common`
   packages pass on a fresh DB; golangci-lint and i18n-lint clean.

## Checklist

- [x] Updated documentation (godoc + SQL comments + octospec brief)

---
_Generated by [Claude Code](https://claude.ai/code/session_01LKr5U3bnKZb3oFDfC9765A)_

### DIFF
--- a/modules/incomingwebhook/adapter.go
+++ b/modules/incomingwebhook/adapter.go
@@ -45,10 +45,16 @@ type pushAdapter struct {
 	// 说明 URL token 已验证、调用方已持有 webhook 真正密钥，故不匹配是配置错误而非枚举
 	// 探测（见 handlePush）。nil 表示该形态无需 header token 二次校验。
 	verifyToken func(header http.Header, urlToken string) bool
+	// allowMention 标记该形态是否处理请求里的 mention（@ 群成员/广播）。仅 native 置真：
+	// @ 由调用方在 native payload 里显式描述（mentionReq），平台适配器(github/wecom/…)
+	// 的 body 是平台生成的事件、没有 octo 成员 UID 语义，强行支持 @ 需要平台身份↔群成员
+	// 的映射表，超出本期范围。置假的形态即便 parse 出 req.Mention 也一律忽略（防御纵深）。
+	allowMention bool
 }
 
 var (
-	nativeAdapter = pushAdapter{name: adapterNative, parse: parseNativePush, bodyLimit: maxBytes}
+	// 仅 native 形态支持 @：调用方在 payload 里用 mention{uids,all,bots} 显式描述。
+	nativeAdapter = pushAdapter{name: adapterNative, parse: parseNativePush, bodyLimit: maxBytes, allowMention: true}
 	githubAdapter = pushAdapter{name: adapterGitHub, parse: parseGitHubPush, bodyLimit: githubMaxBytes}
 	wecomAdapter  = pushAdapter{
 		name:      adapterWeCom,
--- a/modules/incomingwebhook/mention_e2e_test.go
+++ b/modules/incomingwebhook/mention_e2e_test.go
@@ -1,0 +1,236 @@
+package incomingwebhook_test
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	modulescommon "github.com/Mininglamp-OSS/octo-server/modules/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// mention 的 E2E（需 MySQL/Redis/WuKongIM，CI 提供）：覆盖 HTTP 可观测的行为——
+// 广播能力位的持久化 / 管理端判权 / push 时能力位未开的 mention_ignored 回报，以及
+// 无 mention 字段的向后兼容。定向 uids 的成员闸 / ais 展开是【刻意不经 HTTP 回显】的
+// （反枚举 + 落在 WuKongIM payload），由 mention_test.go 的纯单测穷举，不在此重复。
+
+// adminCreateWebhook 以群主（admin）创建一个 webhook，可选带广播能力位，返回响应体。
+func adminCreateWebhook(t *testing.T, handler http.Handler, groupNo string, body map[string]interface{}) map[string]interface{} {
+	t.Helper()
+	w := do(handler, authReq("POST", fmt.Sprintf("/v1/groups/%s/incoming-webhooks", groupNo), body))
+	require.Equalf(t, http.StatusOK, w.Code, "admin create body: %s", w.Body.String())
+	return parseJSON(t, w)
+}
+
+// TestMention_AdminCreateWithSwitchesPersist：管理员可在 create 时开启两个广播能力位，
+// 响应与后续 list 都如实回显 1/1。
+func TestMention_AdminCreateWithSwitchesPersist(t *testing.T) {
+	handler, _, groupNo := setupTestEnv(t)
+	created := adminCreateWebhook(t, handler, groupNo, map[string]interface{}{
+		"name":               "ci",
+		"allow_mention_all":  true,
+		"allow_mention_bots": true,
+	})
+	assert.EqualValues(t, 1, created["allow_mention_all"])
+	assert.EqualValues(t, 1, created["allow_mention_bots"])
+
+	// list 回显同样的能力位。
+	lw := do(handler, authReq("GET", fmt.Sprintf("/v1/groups/%s/incoming-webhooks", groupNo), nil))
+	require.Equalf(t, http.StatusOK, lw.Code, "list body: %s", lw.Body.String())
+	list, _ := parseJSON(t, lw)["list"].([]interface{})
+	require.Len(t, list, 1)
+	first, _ := list[0].(map[string]interface{})
+	assert.EqualValues(t, 1, first["allow_mention_all"])
+	assert.EqualValues(t, 1, first["allow_mention_bots"])
+}
+
+// TestMention_DefaultSwitchesOff：不带能力位字段创建 → 两个开关默认关闭（0）。
+func TestMention_DefaultSwitchesOff(t *testing.T) {
+	handler, _, groupNo := setupTestEnv(t)
+	created := adminCreateWebhook(t, handler, groupNo, map[string]interface{}{"name": "ci"})
+	assert.EqualValues(t, 0, created["allow_mention_all"])
+	assert.EqualValues(t, 0, created["allow_mention_bots"])
+}
+
+// TestMention_MemberCanEnableSwitches：任意合法成员（role=0，非管理员）可在自建 webhook
+// 上开启广播能力位（与「成员可自建/自管 webhook」一致，不再要求管理员）；缺省则默认关闭。
+func TestMention_MemberCanEnableSwitches(t *testing.T) {
+	handler, _, groupNo := setupMemberEnv(t)
+
+	wOn := do(handler, userReq("POST", fmt.Sprintf("/v1/groups/%s/incoming-webhooks", groupNo),
+		map[string]interface{}{"allow_mention_all": true, "allow_mention_bots": true}, memberAToken))
+	require.Equalf(t, http.StatusOK, wOn.Code, "body: %s", wOn.Body.String())
+	on := parseJSON(t, wOn)
+	assert.EqualValues(t, 1, on["allow_mention_all"])
+	assert.EqualValues(t, 1, on["allow_mention_bots"])
+
+	// 不带开关 → 默认关闭。
+	wOK := do(handler, userReq("POST", fmt.Sprintf("/v1/groups/%s/incoming-webhooks", groupNo),
+		map[string]interface{}{}, memberAToken))
+	require.Equalf(t, http.StatusOK, wOK.Code, "body: %s", wOK.Body.String())
+	def := parseJSON(t, wOK)
+	assert.EqualValues(t, 0, def["allow_mention_all"])
+	assert.EqualValues(t, 0, def["allow_mention_bots"])
+}
+
+// TestMention_UpdateTogglesSwitches：管理员可经 update 开/关能力位；普通成员同样可改
+// 自己创建的 webhook 的能力位（与「成员可自建/自管 webhook」一致）。
+func TestMention_UpdateTogglesSwitches(t *testing.T) {
+	handler, _, groupNo := setupMemberEnv(t)
+
+	// 管理员建（默认 0/0）→ 开启 → 关闭。
+	created := adminCreateWebhook(t, handler, groupNo, map[string]interface{}{"name": "ci"})
+	whID, _ := created["webhook_id"].(string)
+
+	on := do(handler, authReq("PUT", fmt.Sprintf("/v1/groups/%s/incoming-webhooks/%s", groupNo, whID),
+		map[string]interface{}{"allow_mention_all": true, "allow_mention_bots": true}))
+	require.Equalf(t, http.StatusOK, on.Code, "body: %s", on.Body.String())
+	onRes := parseJSON(t, on)
+	assert.EqualValues(t, 1, onRes["allow_mention_all"])
+	assert.EqualValues(t, 1, onRes["allow_mention_bots"])
+
+	off := do(handler, authReq("PUT", fmt.Sprintf("/v1/groups/%s/incoming-webhooks/%s", groupNo, whID),
+		map[string]interface{}{"allow_mention_all": false}))
+	require.Equalf(t, http.StatusOK, off.Code, "body: %s", off.Body.String())
+	assert.EqualValues(t, 0, parseJSON(t, off)["allow_mention_all"])
+	assert.EqualValues(t, 1, parseJSON(t, off)["allow_mention_bots"], "untouched switch keeps its value")
+
+	// 普通成员对【自己创建】的 webhook 改能力位 → 200 且生效（成员可自管自建 webhook）。
+	mCreated := memberCreate(t, handler, groupNo)
+	mID, _ := mCreated["webhook_id"].(string)
+	mw := do(handler, userReq("PUT", fmt.Sprintf("/v1/groups/%s/incoming-webhooks/%s", groupNo, mID),
+		map[string]interface{}{"allow_mention_all": true}, memberAToken))
+	require.Equalf(t, http.StatusOK, mw.Code, "body: %s", mw.Body.String())
+	assert.EqualValues(t, 1, parseJSON(t, mw)["allow_mention_all"])
+}
+
+// TestMentionPush_BroadcastIgnoredWhenCapabilityOff：能力位关闭时 push 带 all/bots，
+// 消息照发（200），但响应体 mention_ignored 回报这两个广播位被忽略。
+func TestMentionPush_BroadcastIgnoredWhenCapabilityOff(t *testing.T) {
+	handler, _, groupNo := setupTestEnv(t)
+	created := adminCreateWebhook(t, handler, groupNo, map[string]interface{}{"name": "ci"}) // 默认 0/0
+	pushURL := fmt.Sprintf("/v1/incoming-webhooks/%s/%s", created["webhook_id"], created["token"])
+
+	pw := do(handler, anonReq("POST", pushURL,
+		[]byte(`{"content":"hi","mention":{"all":true,"bots":true}}`)))
+	require.Equalf(t, http.StatusOK, pw.Code, "push body: %s", pw.Body.String())
+	body := parseJSON(t, pw)
+	ignored, ok := body["mention_ignored"].([]interface{})
+	require.Truef(t, ok, "expected mention_ignored array, body: %s", pw.Body.String())
+	assert.ElementsMatch(t, []interface{}{"all", "bots"}, ignored)
+}
+
+// TestMentionPush_BroadcastAllowedWhenCapabilityOn：能力位开启时 push 带 all/bots →
+// 200 且【无】mention_ignored（广播被接受）。
+func TestMentionPush_BroadcastAllowedWhenCapabilityOn(t *testing.T) {
+	handler, _, groupNo := setupTestEnv(t)
+	created := adminCreateWebhook(t, handler, groupNo, map[string]interface{}{
+		"name":               "ci",
+		"allow_mention_all":  true,
+		"allow_mention_bots": true,
+	})
+	pushURL := fmt.Sprintf("/v1/incoming-webhooks/%s/%s", created["webhook_id"], created["token"])
+
+	pw := do(handler, anonReq("POST", pushURL,
+		[]byte(`{"content":"hi","mention":{"all":true,"bots":true}}`)))
+	require.Equalf(t, http.StatusOK, pw.Code, "push body: %s", pw.Body.String())
+	_, hasIgnored := parseJSON(t, pw)["mention_ignored"]
+	assert.Falsef(t, hasIgnored, "no broadcast should be ignored, body: %s", pw.Body.String())
+}
+
+// TestMentionPush_TargetedUIDsSucceed：定向 @ 群成员 → 200 且无 mention_ignored
+// （定向 @ 不受能力位约束；非成员静默丢弃不回显）。冒烟覆盖 push 路径接 mention 不出错。
+func TestMentionPush_TargetedUIDsSucceed(t *testing.T) {
+	handler, _, groupNo := setupMemberEnv(t)
+	created := adminCreateWebhook(t, handler, groupNo, map[string]interface{}{"name": "ci"})
+	pushURL := fmt.Sprintf("/v1/incoming-webhooks/%s/%s", created["webhook_id"], created["token"])
+
+	body := fmt.Sprintf(`{"content":"hi","mention":{"uids":["%s","not_a_member"]}}`, memberAUID)
+	pw := do(handler, anonReq("POST", pushURL, []byte(body)))
+	require.Equalf(t, http.StatusOK, pw.Code, "push body: %s", pw.Body.String())
+	_, hasIgnored := parseJSON(t, pw)["mention_ignored"]
+	assert.False(t, hasIgnored, "targeted uids are not capability-gated; non-members dropped silently")
+}
+
+// TestMentionPush_MemberBroadcastRevokedBySystemSetting 验证「预埋的收回开关」(option C)：
+// system_setting incomingwebhook.member_can_broadcast=0 时，【成员】创建的 webhook 的广播
+// (@所有人/@所有 AI) 在 push 读路径被即时剥离（mention_ignored 回报），而【管理员】创建的
+// webhook 不受影响——无需迁移已置 1 的能力位列、改个后台值即可收回。
+func TestMentionPush_MemberBroadcastRevokedBySystemSetting(t *testing.T) {
+	handler, ctx, groupNo := setupMemberEnv(t)
+	t.Setenv("OCTO_INCOMINGWEBHOOK_MEMBER_CAN_BROADCAST", "") // 证明纯由 DB 驱动
+
+	// 成员（非管理员）建一个开了两个广播位的 webhook。
+	mw := do(handler, userReq("POST", fmt.Sprintf("/v1/groups/%s/incoming-webhooks", groupNo),
+		map[string]interface{}{"allow_mention_all": true, "allow_mention_bots": true}, memberAToken))
+	require.Equalf(t, http.StatusOK, mw.Code, "member create: %s", mw.Body.String())
+	mRes := parseJSON(t, mw)
+	memberPush := fmt.Sprintf("/v1/incoming-webhooks/%s/%s", mRes["webhook_id"], mRes["token"])
+
+	// 管理员建一个开了广播位的 webhook。
+	aRes := adminCreateWebhook(t, handler, groupNo, map[string]interface{}{
+		"name": "admin-ci", "allow_mention_all": true, "allow_mention_bots": true})
+	adminPush := fmt.Sprintf("/v1/incoming-webhooks/%s/%s", aRes["webhook_id"], aRes["token"])
+
+	// 收回：member_can_broadcast=0 + Reload 共享快照（与 TestFeatureToggle 同法）。
+	_, err := ctx.DB().InsertInto("system_setting").
+		Pair("category", "incomingwebhook").Pair("key_name", "member_can_broadcast").
+		Pair("value", "0").Pair("value_type", "bool").Pair("description", "").Exec()
+	require.NoError(t, err)
+	settings := modulescommon.EnsureSystemSettings(ctx)
+	require.NoError(t, settings.Reload())
+	defer func() {
+		_, _ = ctx.DB().DeleteFrom("system_setting").Where("category=?", "incomingwebhook").Exec()
+		_ = settings.Reload()
+	}()
+
+	// 成员 webhook 广播被即时剥离 → mention_ignored 回报 all+bots（无需改其能力位列）。
+	mp := do(handler, anonReq("POST", memberPush, []byte(`{"content":"hi","mention":{"all":true,"bots":true}}`)))
+	require.Equalf(t, http.StatusOK, mp.Code, "member push: %s", mp.Body.String())
+	ig, ok := parseJSON(t, mp)["mention_ignored"].([]interface{})
+	require.Truef(t, ok, "member broadcast must be reported ignored, body: %s", mp.Body.String())
+	assert.ElementsMatch(t, []interface{}{"all", "bots"}, ig)
+
+	// 管理员 webhook 不受影响 → 无 mention_ignored（creatorIsAdmin 仍放行）。
+	ap := do(handler, anonReq("POST", adminPush, []byte(`{"content":"hi","mention":{"all":true,"bots":true}}`)))
+	require.Equalf(t, http.StatusOK, ap.Code, "admin push: %s", ap.Body.String())
+	_, adminIgnored := parseJSON(t, ap)["mention_ignored"]
+	assert.Falsef(t, adminIgnored, "admin-created webhook broadcast must be unaffected, body: %s", ap.Body.String())
+}
+
+// TestMentionPush_MalformedMentionIgnored 钉 acceptance #6（review #445 的 P1 阻塞）：
+// 形状非法的 mention（uids 非数组 / mention 非对象 等）必须【降级为无 mention、消息照投
+// 200】，而不是把整条推送 400 掉（修复前这些都返回 400，把相邻的合法 content 也连累丢弃）。
+func TestMentionPush_MalformedMentionIgnored(t *testing.T) {
+	handler, _, groupNo := setupTestEnv(t)
+	created := adminCreateWebhook(t, handler, groupNo, map[string]interface{}{"name": "ci"})
+	pushURL := fmt.Sprintf("/v1/incoming-webhooks/%s/%s", created["webhook_id"], created["token"])
+
+	for _, body := range []string{
+		`{"content":"hi","mention":"please"}`,
+		`{"content":"hi","mention":{"uids":"alice"}}`,
+		`{"content":"hi","mention":{"uids":[1,2]}}`,
+		`{"content":"hi","mention":{"all":1}}`,
+		`{"content":"hi","mention":[]}`,
+	} {
+		pw := do(handler, anonReq("POST", pushURL, []byte(body)))
+		require.Equalf(t, http.StatusOK, pw.Code,
+			"malformed mention must still deliver (200), body=%s resp=%s", body, pw.Body.String())
+		_, hasIgnored := parseJSON(t, pw)["mention_ignored"]
+		assert.Falsef(t, hasIgnored, "malformed mention is dropped, not reported: %s", body)
+	}
+}
+
+// TestMentionPush_NoMentionFieldBackwardCompatible：不带 mention 字段的历史 native 调用
+// 行为不变（200，无 mention_ignored）。
+func TestMentionPush_NoMentionFieldBackwardCompatible(t *testing.T) {
+	handler, _, groupNo := setupTestEnv(t)
+	created := adminCreateWebhook(t, handler, groupNo, map[string]interface{}{"name": "ci"})
+	pushURL := fmt.Sprintf("/v1/incoming-webhooks/%s/%s", created["webhook_id"], created["token"])
+
+	pw := do(handler, anonReq("POST", pushURL, []byte(`{"content":"hi"}`)))
+	require.Equalf(t, http.StatusOK, pw.Code, "push body: %s", pw.Body.String())
+	_, hasIgnored := parseJSON(t, pw)["mention_ignored"]
+	assert.False(t, hasIgnored)
+}
--- a/.octospec/tasks/incoming-webhook-mention/brief.md
+++ b/.octospec/tasks/incoming-webhook-mention/brief.md
@@ -1,0 +1,110 @@
+---
+type: Task
+title: "Task: incoming-webhook-mention"
+description: Add caller-controlled @mention (users + bots, one or many) to the native incoming-webhook push endpoint, with per-webhook capability switches for the two broadcast forms.
+tags: ["incomingwebhook", "mention", "wire-contract", "trust-boundary"]
+timestamp: 2026-06-23T13:15:14Z
+# --- octospec extension fields ---
+slug: incoming-webhook-mention
+upstream: self
+source: self
+---
+
+# Task: incoming-webhook-mention
+
+> One task = one `.octospec/tasks/<slug>/` directory. This brief is the spec for
+> the work.
+
+## Goal
+
+Let an incoming-webhook caller @mention **specific group members (users and/or
+bots, one or many)** and optionally broadcast **@所有人 (humans)** and
+**@所有 AI (bots)** from the **native** push endpoint. Mentioned humans get the
+existing `[有人@我]` red-dot; mentioned bots get delivery (no red-dot), all via
+the **existing** message-module reminder listener — this task only puts a
+well-formed `mention` map on the wire.
+
+The two broadcast forms are abuse-sensitive (a leaked webhook token spamming the
+whole group), so each is gated by a **per-webhook capability switch**, default
+**off**, toggled through the management API by any legitimate member who manages
+the webhook (its creator, or a group admin) — the per-webhook, default-off switch
+is the control, not a separate admin grant. Targeted `@uid` needs no switch but is
+constrained to current group members + a hard cap. (Revoking members' ability to
+flip these later is an additive group-level or system-setting policy ANDed at the
+push read path — see the journal/notes; no schema change.)
+
+## Background
+
+- Today `buildPayload` (`modules/incomingwebhook/api.go`) emits only
+  `type/content/from/space_id` and **discards `extra`**; there is no mention
+  support (`@all/@here` is left as literal text). The code comment there is the
+  sanctioned extension point: add an **explicitly whitelisted** field, never a
+  passthrough.
+- The wire mention contract is the `mention` sub-map (`uids/humans/ais/entities`)
+  owned by `pkg/mentionrewrite` and read by `modules/message/api_reminders.go`
+  (`getMention`). Reminder generation is **sender-agnostic** (a WuKongIM message
+  listener), so a webhook message carrying `mention.uids/humans` fires reminders
+  with no message-module change.
+- `mention.ais=1` is expanded to bot-member UIDs by
+  `pkg/mentionrewrite.ExpandAisToBotUIDs`, fed by a per-ingress composer
+  (`GetMembers` + `robot.ExistRobot`) — mirror `modules/bot_api/mention_expand.go`.
+- Decisions locked with maintainer: native endpoint only; caller supplies UIDs
+  (no name resolution); both broadcast switches default off; UID cap = 50.
+- Industry parity: structured-field opt-in (caller must set `all/bots:true`) ≈
+  Discord `allowed_mentions`; the per-webhook capability ≈ Slack "who can
+  @channel" but at webhook-resource granularity.
+
+## Load-bearing list
+
+- **wire-contract** — `mention.{uids,humans,ais}` shape must match what
+  `pkg/mentionrewrite` + `message/api_reminders.go getMention` already consume
+  (UID strings; truthy-one flags). Do not invent a new shape.
+- **external-content / trust-boundary / webhook** — the push endpoint is an
+  attacker-controlled, unauthenticated (token-in-URL) ingress. New field must be
+  validated at the boundary: membership gate, dedup, count cap; mention is a
+  **whitelisted structured field**, never `extra` passthrough.
+- **space / isolation** — mention targets MUST be filtered to current members of
+  the webhook's `group_no`; a token must not ping arbitrary platform users or
+  cross-Space identities.
+- **error-response / i18n** — any new management-side validation reuses
+  `httperr.ResponseErrorL` + a registered `pkg/errcode` code; push-side
+  "ignored/unresolved" feedback is non-fatal data in the 200 body, not an error.
+- **adapter parity (native-only)** — mention is gated to the native adapter via a
+  `pushAdapter` capability flag; sibling adapters (wecom/feishu/github/gitlab/
+  multica) neither parse nor emit mention. Adding a feature only to native is
+  safe (no new attack surface for siblings); the parity rule about *escaping*
+  still holds because mention carries structured UIDs, not rendered markdown.
+- **bot-api / robot wiring** — `IncomingWebhook` gains a robot dependency
+  (`ExistRobot` / bot-member enumeration) it does not currently hold.
+- existing push pipeline (auth → rate-limit → group-active → creator-membership
+  → audit) and `buildRichTextPayload` must be unaffected except for the new
+  mention insertion point.
+
+## Out of scope
+
+- Name/username/email → UID resolution (`mention.names`).
+- Mention via platform adapters (WeCom/Feishu/GitHub/GitLab/Multica).
+- `mention.entities` offset/pill generation (v1: `uids` drive the ping; the
+  literal `@name` in `content` is plain text). Pills are a later phase.
+- `mention.all` legacy field semantics and any change to the message-module
+  reminder/read path.
+- Group-level mention policy (`allow_no_mention` / `bot_mention_pref`).
+
+## Acceptance
+
+- Native push with `mention.uids=[<member uids>]` → delivered payload carries
+  `mention.uids` containing **only** current group members, **deduped**, **≤50**;
+  non-members and blanks dropped.
+- `mention.all=true` + webhook `allow_mention_all=1` → payload `mention.humans=1`;
+  with `allow_mention_all=0` → no `humans`, and 200 body reports it ignored.
+- `mention.bots=true` + webhook `allow_mention_bots=1` → `ExpandAisToBotUIDs`
+  appends bot-member UIDs; with `allow_mention_bots=0` → not expanded + reported.
+- A non-native adapter (e.g. wecom) with a `mention` field in its body → no
+  `mention` in the delivered payload.
+- `create`/`update` accept + persist `allow_mention_all` / `allow_mention_bots`
+  (default 0); `webhookResp` surfaces them.
+- Empty/absent `mention`, malformed `mention.uids` → no panic, no `mention` key
+  (full backward compatibility with existing native callers).
+- `go test ./modules/incomingwebhook/...` passes incl. a new mention test +
+  `TestIncomingWebhookNoLegacyResponseError`; `make i18n-extract-check` +
+  `make i18n-lint` pass; `golangci-lint run ./modules/incomingwebhook/...` clean.
--- a/.octospec/tasks/incoming-webhook-mention/context.yaml
+++ b/.octospec/tasks/incoming-webhook-mention/context.yaml
@@ -1,0 +1,38 @@
+slug: incoming-webhook-mention
+rules_injected:
+  - id: trust-boundary
+    tier: repo
+    load_bearing: true
+    why: incomingwebhook push is attacker-controlled ingress; mention is a whitelisted
+         structured field (never extra passthrough); uids bounded + member-gated;
+         feature gated to native adapter (parity preserved — structured UIDs, not markdown).
+  - id: space-isolation
+    tier: repo
+    load_bearing: true
+    why: mention targets filtered to current internal-normal members of the webhook's
+         group_no; a group-scoped token cannot ping arbitrary or cross-Space identities.
+  - id: error-handling
+    tier: repo
+    load_bearing: true
+    why: management-side capability gate reuses mgmtRequestInvalid (ErrIncomingWebhookRequestInvalid);
+         no new error codes added; push-side ignored/dropped feedback is non-fatal 200 body, not an error.
+  - id: testing
+    tier: repo
+    load_bearing: false
+    why: new mention_test.go + extend TestIncomingWebhookNoLegacyResponseError file list.
+  - id: commit-style
+    tier: repo
+    load_bearing: false
+    why: Conventional Commits, English.
+files_touched:
+  - modules/incomingwebhook/model.go        # mentionReq, model+req+resp switch fields
+  - modules/incomingwebhook/api.go          # struct wiring, handlePush mention attach, create/update gates, toResp
+  - modules/incomingwebhook/adapter.go      # pushAdapter.allowMention (native-only)
+  - modules/incomingwebhook/mention.go      # buildMention + helpers (new)
+  - modules/incomingwebhook/mention_expand.go # fetchBotMemberUIDs composer (new)
+  - modules/incomingwebhook/db.go           # filterGroupMembers
+  - modules/incomingwebhook/sql/20260623000001_incomingwebhook_mention_switches.sql # new columns
+parity_decisions:
+  - "@所有 AI expansion reuses pkg/mentionrewrite.ExpandAisToBotUIDs with GetMembers+ExistRobot, identical to modules/bot_api + modules/message ingresses."
+  - "wire mention shape {uids:[]interface{}, humans:1, ais:1} matches message/api_reminders.go getMention + pkg/mentionrewrite keys."
+  - "reminders/red-dot/bot-summon happen downstream via the sender-agnostic message listener — no message-module change."
--- a/modules/common/system_setting_schema.go
+++ b/modules/common/system_setting_schema.go
@@ -117,6 +117,8 @@ var systemSettingSchema = []settingDef{
 		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.IncomingWebhookMaxPerGroup()) }},
 	{Category: "incomingwebhook", Key: "max_per_creator", Type: settingTypeInt, Description: "单个普通成员/机器人在一个群内最多可创建的 Webhook 数量（群主/管理员不受限）", Positive: true,
 		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.IncomingWebhookMaxPerCreator()) }},
+	{Category: "incomingwebhook", Key: "member_can_broadcast", Type: settingTypeBool, Description: "非管理员成员创建的 Webhook 是否可用广播型 @（@所有人/@所有 AI）；关闭后即时收回成员广播，管理员创建的不受影响",
+		Effective: func(s *SystemSettings) string { return boolToCanonical(s.IncomingWebhookMemberCanBroadcast()) }},
 
 	// Email server config — formerly yaml-only (Support.* in config.go).
 	{Category: "support", Key: "email", Type: settingTypeString, Description: "技术支持邮箱（发件人）",
--- a/modules/common/system_settings.go
+++ b/modules/common/system_settings.go
@@ -561,12 +561,15 @@ const (
 	envIncomingWebhookPerWebhookBurst = "DM_INCOMINGWEBHOOK_BURST"
 	envIncomingWebhookMaxPerGroup     = "DM_INCOMINGWEBHOOK_MAX_PER_GROUP"
 	envIncomingWebhookMaxPerCreator   = "DM_INCOMINGWEBHOOK_MAX_PER_CREATOR"
+	// 控制非管理员成员创建的 webhook 是否可用广播型 @（@所有人 / @所有 AI）。
+	envIncomingWebhookMemberCanBroadcast = "OCTO_INCOMINGWEBHOOK_MEMBER_CAN_BROADCAST"
 
-	defaultIncomingWebhookEnabled         = true
-	defaultIncomingWebhookPerWebhookRPS   = 5.0
-	defaultIncomingWebhookPerWebhookBurst = 10
-	defaultIncomingWebhookMaxPerGroup     = 10
-	defaultIncomingWebhookMaxPerCreator   = 5
+	defaultIncomingWebhookEnabled            = true
+	defaultIncomingWebhookPerWebhookRPS      = 5.0
+	defaultIncomingWebhookPerWebhookBurst    = 10
+	defaultIncomingWebhookMaxPerGroup        = 10
+	defaultIncomingWebhookMaxPerCreator      = 5
+	defaultIncomingWebhookMemberCanBroadcast = true
 )
 
 // IncomingWebhookEnabled 是群入站 Webhook 功能的总开关。关闭后 push 端点返回 404、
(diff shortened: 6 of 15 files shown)


modules/incomingwebhook/mention_e2e_test.go
```diff
@@ -0,0 +1,213 @@
...
+func TestMentionPush_NoMentionFieldBackwardCompatible(t *testing.T) {
+	handler, _, groupNo := setupTestEnv(t)
+	created := adminCreateWebhook(t, handler, groupNo, map[string]interface{}{"name": "ci"})
+	pushURL := fmt.Sprintf("/v1/incoming-webhooks/%s/%s", created["webhook_id"], created["token"])
+
+	pw := do(handler, anonReq("POST", pushURL, []byte(`{"content":"hi"}`)))
+	require.Equalf(t, http.StatusOK, pw.Code, "push body: %s", pw.Body.String())
+	_, hasIgnored := parseJSON(t, pw)["mention_ignored"]
+	assert.False(t, hasIgnored)
+}
```
OctoBoooot, line 236: **[minor]** Acceptance criterion #4 in `.octospec/tasks/incoming-webhook-mention/brief.md` says "A non-native adapter (e.g. wecom) with a `mention` field in its body → no `mention` in the delivered payload." That is covered by code (`pushAdapter.allowMention` defaults false; gate at `api.go:1324`), but **no test pins it**. The gate has two layers of safety today (no non-native adapter populates `req.Mention`, AND `if ad.allowMention` skips `buildMention`), but a future contributor modifying e.g. `adapter_wecom.go` to start parsing `mention` could bypass intent if `allowMention=true` is accidentally added or the gate is dropped — the existing test file would not catch it.

Add a one-screen e2e: push to `/v1/incoming-webhooks/.../wecom` with a body containing a wecom-shape message that includes a `mention` field, and either (a) call the parse + handlePush path and assert the delivered payload has no `mentionrewrite.MentionKey`, or (b) call `buildMention` directly with `ad := wecomAdapter; assert.False(t, ad.allowMention)` plus an HTTP-level assertion. Either shape is fine; the test name should make the load-bearing intent explicit ("`mention` field is dropped on non-native adapters").

OctoBoooot (changes requested):
## Review: feat(incomingwebhook): mention support on the native push endpoint (#445)

**Verdict:** Request changes — CI red on `check-sprint / check-sprint` (process gate: no linked issue) precludes APPROVE per skill rule §11. Code itself is solid; no blockers, no majors, one minor on test-gap.

Size: 15 files / 1040 lines → `size:large` label applied. Risk tier classify = `normal`, but the surface (unauthenticated incoming-webhook ingress + a new structured mention field) is trust-boundary critical per the PR's own brief — I reviewed at the bar that warrants.

### Adversarial verification on the trust-boundary properties (per skill §7.1)

- **Member filter is server-side** — `buildMention` passes `m.GroupNo` (the webhook's own server-stored group), never caller input, into `filterGroupMembers`. The query filters to `is_deleted=0 AND is_external=0 AND status=Normal` — same shape as `queryMemberRole`. Cross-Space crossover: not possible (group membership is space-scoped at the data layer, so filtering to current members implicitly enforces Space isolation).
- **Targeted-uids cap** — `dedupNonEmpty(uids, maxMentionUIDs())` defaults 50, env-overridable to positive int only. JSON body cap is 8 KiB → even before the dedup cap, a uids array is bounded. IN clause is therefore always ≤ 50.
- **Anti-enumeration on targeted uids** — non-members are silently dropped, never echoed in `mention_ignored`. Verified at `mention.go:46-47` and `assembleMention:96-99` (no append to `ignored` on non-member). Response body is identical between "uid was a member" and "uid was not a member" cases.
- **Capability gates default off** — SQL migration `DEFAULT 0` on both columns. `boolToInt(boolPtrTrue(...))` in create handler defaults to 0 for nil/false. Defense-in-depth: `m.AllowMentionAll == 1 && broadcastPermitted` — both must be true.
- **System-setting kill switch** — `IncomingWebhookMemberCanBroadcast` ANDed at the push read path, so flipping the system setting **instantly** revokes all member-created webhooks' broadcasts without column migration. Admin-created webhooks (where `creatorIsAdmin=true` per `cachedCreatorMembership`) bypass the kill switch — that asymmetry is intentional and documented.
- **Adapter parity / non-native bypass** — `pushAdapter.allowMention` defaults false; only `nativeAdapter` sets true. `if ad.allowMention { ... }` gate in `handlePush:1324`. Verified independently by grep: no non-native adapter (`adapter_wecom.go` / `_github.go` / `_gitlab.go` / `_feishu.go` / `_multica.go`) populates `req.Mention`, and `adapter_wecom.go:12-13` explicitly drops `mentioned_list`. Belt-and-suspenders is genuine.
- **Backward compat** — `mr == nil` early return + `if mention != nil { payload[key] = mention }` means no-mention payloads are byte-identical to pre-PR. Pinned by `TestMentionPush_NoMentionFieldBackwardCompatible` e2e.
- **`mention.uids` shape contract** — `kept` is built as `[]interface{}` (not `[]string`) so `ExpandAisToBotUIDs` will in-place append bot uids. Pinned by `TestAssembleMentionFeedsExpandAis` parity test with the real `mentionrewrite.ExpandAisToBotUIDs`.
- **Description-vs-diff crosscheck** — every quantitative/location-specific claim in the PR body verified:
  - "default 50, `OCTO_INCOMINGWEBHOOK_MAX_MENTION_UIDS`" → `defaultMaxMentionUIDs=50` and env var name match (mention.go:16-18).
  - "default 0 (admin-created bypass)" → SQL `DEFAULT 0` + create handler boolPtrTrue(nil)=false matches.
  - "instant add columns on MySQL 8.0" → `ADD COLUMN ... DEFAULT ...` at table end is INSTANT-eligible on 8.0; no FK/virtual-col interactions.
  - "mention is whitelisted, extra stays discarded" → `buildPayload` still does NOT pass `req.Extra` (api.go:1409-1424); mention is attached separately at handlePush:1333 after `buildPayload` returns.

### Minor
- `modules/incomingwebhook/mention_e2e_test.go` — see inline.

### Praise

- **Threat model is sharp.** Anti-enumeration (silent drop, not `mention_ignored`) + capability flag default off + system-setting kill switch + per-webhook capability column + adapter parity gate is **four** independent defenses for the broadcast surface. The `creatorIsAdmin || system_setting` AND-gate at the push read path is the right shape: it lets admin webhooks (operator-trusted) keep broadcasting even after member broadcasts are revoked, without touching stored capability columns. That single design choice removes the migration that the alternative ("revoke = set all columns to 0") would require.
- **`assembleMention` (pure) vs `buildMention` (IO) split** is the right shape for exhaustive testing. The 9-case table in `TestAssembleMention` covers the matrix; the IO seam (`filterGroupMembers`) lives in one place and is e2e-tested separately. This mirrors and improves on existing patterns in the same package.
- **`TestAssembleMentionFeedsExpandAis` is a load-bearing parity test.** It locks the wire shape against drift between the assembler and the real consumer (`pkg/mentionrewrite.ExpandAisToBotUIDs`) — the kind of test that catches "I refactored uids from `[]interface{}` to `[]string` because the linter complained" regressions years from now.
- **db.go::filterGroupMembers comment** explicitly names the deliberate asymmetry (targeted uids reject external members; broadcast `@所有 AI` follows the same `GetMembers` rules as message/bot_api ingresses so external bots in a group can still be reached). That asymmetry is exactly the kind of thing a future maintainer would "fix" without realizing it's a trust-boundary choice; the comment will keep them from doing that.
- **SQL migration's `INSTANT` rationale** is explicit (ADD at table end, default 0). Reverse migration is symmetric. No SDLC surprise.

Jerry-Xin (changes requested):
This PR is relevant to `octo-server`, but one functional contract issue should be fixed before merge.

**🔴 Blocking**
- 🔴 Critical: Malformed optional `mention` input rejects the whole webhook push instead of being ignored. `parseNativePush` directly unmarshals into `pushPayloadReq`, where `Mention *mentionReq` contains `Uids []string` ([modules/incomingwebhook/adapter.go:87]( /Users/mlamp/.openclaw/tmp/pr-review-dwub3fhv/repo/modules/incomingwebhook/adapter.go:87), [modules/incomingwebhook/model.go:112]( /Users/mlamp/.openclaw/tmp/pr-review-dwub3fhv/repo/modules/incomingwebhook/model.go:112)). Payloads like `{"content":"hi","mention":{"uids":"u1"}}`, `{"content":"hi","mention":{"uids":[1]}}`, or `{"content":"hi","mention":[]}` will make `json.Unmarshal` fail, returning `invalidReason="json"` and a 400. The PR contract says malformed `mention.uids` should be non-fatal and result in no `mention` key. Consider decoding `mention` via a custom `UnmarshalJSON` or `json.RawMessage` so malformed mention shapes are dropped while preserving the message.

**💬 Non-blocking**
- 🔵 Suggestion: The test comment above `TestMention_UpdateTogglesSwitches` contradicts the actual assertions: the test expects a normal member to update their own webhook successfully ([modules/incomingwebhook/mention_e2e_test.go:78]( /Users/mlamp/.openclaw/tmp/pr-review-dwub3fhv/repo/modules/incomingwebhook/mention_e2e_test.go:78), [modules/incomingwebhook/mention_e2e_test.go:99]( /Users/mlamp/.openclaw/tmp/pr-review-dwub3fhv/repo/modules/incomingwebhook/mention_e2e_test.go:99)). Please update the comment to match the intended policy.

**✅ Highlights**
- Native-only mention handling is explicitly gated through `pushAdapter.allowMention`.
- Targeted mentions are capped, deduplicated, and group-member filtered.
- Broadcast capability changes invalidate the local webhook cache through the existing update path.

Verification: targeted pure tests passed with `go test ./modules/incomingwebhook -run 'TestParseNativePushMention|TestAssembleMention|TestDedupNonEmpty|TestMaxMentionUIDs|TestBoolHelpers' -count=1`. Full `go test ./modules/incomingwebhook/...` was blocked by the local test database migration state: `unknown migration in database`.

modules/incomingwebhook/adapter.go
```diff
@@ -45,10 +45,16 @@ type pushAdapter struct {
...
+	// @ 由调用方在 native payload 里显式描述（mentionReq），平台适配器(github/wecom/…)
+	// 的 body 是平台生成的事件、没有 octo 成员 UID 语义，强行支持 @ 需要平台身份↔群成员
+	// 的映射表，超出本期范围。置假的形态即便 parse 出 req.Mention 也一律忽略（防御纵深）。
+	allowMention bool
 }
 
 var (
-	nativeAdapter = pushAdapter{name: adapterNative, parse: parseNativePush, bodyLimit: maxBytes}
+	// 仅 native 形态支持 @：调用方在 payload 里用 mention{uids,all,bots} 显式描述。
+	nativeAdapter = pushAdapter{name: adapterNative, parse: parseNativePush, bodyLimit: maxBytes, allowMention: true}
```
OctoBoooot, line 57: **[major]** Wiring `parseNativePush` to consume `mention` via the strict `pushPayloadReq` struct means a **malformed `mention` shape fails the entire push** — directly violating `.octospec/tasks/incoming-webhook-mention/brief.md` acceptance #6: "Empty/absent `mention`, malformed `mention.uids` → no panic, no `mention` key (full backward compatibility with existing native callers)."

Root cause: `parseNativePush` (this file, ~L87-92) calls `json.Unmarshal(body, &req)` on the whole body. When the `mention` field has the wrong JSON type, the unmarshal errors out and the function returns `invalidReason="json"`, which `handlePush` turns into a 400 (submitFailure + `pushPayloadInvalid`). Same fate for the otherwise-valid `content`.

Failing inputs (each returns 400 today; pre-PR they'd return 200 because unknown top-level fields were silently dropped via `Extra` discard):

- `{"content":"hi","mention":"please"}` — `mention` is a string, can't decode into `*mentionReq{}`.
- `{"content":"hi","mention":{"uids":"alice"}}` — `uids` is a string not array.
- `{"content":"hi","mention":{"uids":[1,2,3]}}` — numeric uids.

The `content` is valid and was deliverable before this PR; with this change it's silently dropped because of an adjacent field's shape. That's the literal backward-compat break acceptance #6 was written to prevent.

Fix shape (smallest change): decode `mention` lazily — change `Mention *mentionReq` in `pushPayloadReq` to `Mention json.RawMessage`, then attempt the targeted decode inside `buildMention` and treat decode errors as "no mention" (Warn-log + drop, same posture as the existing `filterGroupMembers` failure). The rest of the body still benefits from strict decoding; only `mention` becomes lenient — matching the acceptance's intent that a malformed mention degrades to absent, not to a 400.

Regression test would assert each of the three failing inputs above returns 200 + no `mention` key on the delivered payload, and that the message is still sent.

OctoBoooot (commented):
## Follow-up @ 28d7dff (review #2)

Acknowledging Jerry-Xin's 🔴 — **byte-verified, the finding is real and I missed it in my first review**. Promoting from "test gap" to a concrete behavioral finding, plus retracting an incorrect praise from my prior review.

### Major — carried into delta
- See inline at `modules/incomingwebhook/adapter.go:87-92`.

### Retraction
My prior review (REQUEST_CHANGES @ 28d7dff) included this in the verification pass:

> **Backward compat** — `mr == nil` early return + `if mention != nil { payload[key] = mention }` means no-mention payloads are byte-identical to pre-PR. Pinned by `TestMentionPush_NoMentionFieldBackwardCompatible` e2e.

That praise checked only one direction of "no mention" (field absent). It did not check the malformed-shape direction (mention present but wrong JSON type), which is the exact case the brief's acceptance #6 calls out ("Empty/absent `mention`, **malformed** `mention.uids` → no panic, no `mention` key"). Treating that praise as a verified backward-compat guarantee was wrong; the assertion the existing test pins is narrower than the acceptance it appeared to cover.

This is the second hit on the [praise-has-weight] rule for me this week — praising a defensive measure implicitly asserts the defense is complete, and I should grep the spec's acceptance bullets against the test names BEFORE writing the praise. Logging.

### No verdict change
RC is already on the PR from both my first review and Jerry-Xin's; this is a focused follow-up so the inline finding is on the PR thread with concrete failing input, not just in chat. Same SHA, so no ledger-side action.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#445 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/445
**Head SHA:** `28d7dff4f2896dac45e9983d31e48e6c4754d68d`
**Merge base:** `8d2f4e34` (main)
**Title:** feat(incomingwebhook): mention support on the native push endpoint

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Wire contract parity (`mention.{uids,humans,ais}`) | ✅ | `mention.go:assembleMention` constructs `[]interface{}` for uids (verified by `mention_test.go:TestAssembleMentionFeedsExpandAis` which feeds the real `ExpandAisToBotUIDs`) |
| Trust boundary: mention is whitelisted, not `extra` passthrough | ✅ | `api.go:1331` inserts mention into payload after `buildPayload`; `buildPayload` still discards `Extra` |
| Membership gate (targeted uids) | ✅ | `db.go:filterGroupMembers` — `is_deleted=0 AND is_external=0 AND status=Normal`; anti-enumeration: non-members silently dropped |
| Broadcast per-webhook capability + system_setting | ✅ | `mention.go:buildMention` ANDs `AllowMentionAll==1 && broadcastPermitted`; `broadcastPermitted = settings.MemberCanBroadcast() || creatorIsAdmin` |
| UID dedup + cap | ✅ | `mention.go:dedupNonEmpty` — trim, dedup, order-preserve, cap at `maxMentionUIDs()` (default 50, env-overridable) |
| Native-only gating | ✅ | `adapter.go`: only `nativeAdapter` has `allowMention: true`; all other adapters default to `false` |
| Auth ordering: mention processed AFTER all auth checks | ✅ | `handlePush` step 6.5 runs after steps 1–6 (webhook lookup → token verify → group-active → rate-limit → creator-membership → payload build) |
| Backward compatibility (no mention field) | ✅ | `model.go:Mention *mentionReq` (pointer, nil when absent); `mention.go:buildMention` short-circuits on nil; test `TestMentionPush_NoMentionFieldBackwardCompatible` |
| SQL migration | ✅ | `SMALLINT NOT NULL DEFAULT 0` — MySQL 8.0 INSTANT algorithm; rows default to 0 = "broadcast off" |
| Parity with sibling ingress `fetchBotMemberUIDs` | ✅ | Character-for-character identical logic across `message`, `bot_api`, `incomingwebhook` (same `GetMembers + ExistRobot` pair) |
| `updateFieldsAllowed` whitelist updated | ✅ | `db.go` adds `allow_mention_all` and `allow_mention_bots` to the allow-set |
| i18n file list updated | ✅ | `api_i18n_test.go` adds `mention.go`, `mention_expand.go` |
| Test coverage | ✅ | Unit (`mention_test.go`), E2E (`mention_e2e_test.go`), system_settings tests — comprehensive |

## 2. Findings

### F1 — P2: Missing `CloneForExpansion` before `ExpandAisToBotUIDs` (parity break)

**Diff-scope:** new (this PR introduces the call site)

`api.go:1332–1335`:
```go
payload[mentionrewrite.MentionKey] = mention
payload = mentionrewrite.ExpandAisToBotUIDs(
    payload, common.ChannelTypeGroup.Uint8(), m.GroupNo, w.fetchBotMemberUIDs)
```

`ExpandAisToBotUIDs` **mutates** the `mention` sub-map in place (appends bot UIDs to `mention.uids`). All three other ingress chokepoints call `CloneForExpansion` first:

- `modules/message/api.go:560` — `wirePayload := mentionrewrite.CloneForExpansion(payload)`
- `modules/robot/api.go:536` — same pattern
- `modules/bot_api/send.go:241` — same pattern

Each has a doc-comment (from PR#145 review) explaining: *"the expansion MUST run on a clone of `payload`, not on `payload` itself"*.

**Why this is not P1 in the webhook path:** `payload` is a local map built at step 6, mutated at step 6.5, serialized via `util.ToJson(payload)` at step 7, and **never referenced again** (only `msgID`, `m.WebhookID`, `ip`, etc. flow into audit). The message/robot/bot_api paths share `payload` with reminder writers that iterate `mention.uids` — webhook has no such downstream consumer.

**Why still P2:** Breaks a documented, project-wide defensive convention. If anyone later adds post-expansion logic referencing `payload` (e.g., audit payload content), they'd get mutated data. The fix is one line: `wirePayload := mentionrewrite.CloneForExpansion(payload)` + pass `wirePayload` to `ExpandAisToBotUIDs` and `ToJson`.

**Recommendation:** Add `CloneForExpansion` for parity. Low cost, high defensive value.

### F2 — P2/nit: Mention processed for richtext msg_type in native adapter

**Diff-scope:** new

Step 6.5 runs after payload construction for **both** `text` and `richtext` msg_types. The native adapter supports both. A native richtext push with `"mention":{...}` would have mention processed and injected into the payload.

This may be intentional (mention semantics are msg_type-agnostic), but the task brief says *"native push endpoint"* without explicitly scoping to text-only. If richtext mention is intentional, a brief code comment would help future readers. If not, gate with `if req.MsgType != "richtext"` or equivalent.

## 3. Recommendations

1. **Add `CloneForExpansion`** at `api.go:1332` — one-line fix for F1.
2. **Clarify richtext + mention scope** with a one-line comment at step 6.5 — addresses F2.

## 4. Additional Observations

- **Anti-enumeration design is sound:** Non-member UIDs in targeted `@` are silently dropped (not reported in `mention_ignored`), preventing the push endpoint from being used as a membership oracle.
- **`member_can_broadcast` system setting** provides a clean instant-revocation mechanism: flipping it strips broadcast from all member-created webhooks at push-read time without requiring column migration.
- **External member asymmetry** (targeted excludes, `@all AI` includes) is documented and intentional — conservative for targeted, parity with other ingresses for broadcast.

## 5. Data-Flow Trace

| Consumed Data | Upstream Source | Flows To | Verified |
|---|---|---|---|
| `creatorIsAdmin` | `cachedCreatorMembership(m.GroupNo, m.CreatorUID)` → `group_member` table (step 3.5) | `broadcastPermitted` gate → `buildMention` → `assembleMention` allowAll/allowBots params | ✅ |
| `m.AllowMentionAll` / `m.AllowMentionBots` | `incoming_webhook` table columns (new, default 0) → loaded via `cachedQueryByWebhookID` | `buildMention` AND with `broadcastPermitted` | ✅ |
| `req.Mention.Uids` | `json.Unmarshal` of native push body → `pushPayloadReq.Mention` (nil-safe pointer) | `dedupNonEmpty` → `filterGroupMembers` → `assembleMention` as `[]interface{}` | ✅ |
| `members` map | `filterGroupMembers(groupNo, uids)` → SQL `SELECT uid FROM group_member WHERE is_deleted=0 AND is_external=0 AND status=Normal` | `assembleMention` filters uids to member-only | ✅ |
| `mention` sub-map | `assembleMention` constructs `{uids: []interface{}, humans: 1, ais: 1}` | `payload[mentionrewrite.MentionKey]` → `ExpandAisToBotUIDs` → JSON serialized → WuKongIM → downstream listener `getMention` reads `humans/ais/uids` | ✅ |
| Bot UIDs from `fetchBotMemberUIDs` | `groupService.GetMembers` + `robotService.ExistRobot` per-member (same as message/bot_api) | `ExpandAisToBotUIDs` appends to `mention.uids` → wire bytes → listener's `isBotUID` filter prevents reminder rows for bots | ✅ |
| `w.settings.IncomingWebhookMemberCanBroadcast()` | DB `system_setting` → env `OCTO_INCOMINGWEBHOOK_MEMBER_CAN_BROADCAST` → default `true` | `broadcastPermitted` = setting OR creatorIsAdmin | ✅ |

## 6. Blind-Spot Checklist (R5 — security_sensitive, all items mandatory)

**C1 — Dual-path parity (incl. guard/lint tool coverage):**
- create ↔ delete: delete path unaffected (no mention columns to clean)
- create ↔ update: both accept `allow_mention_all`/`allow_mention_bots` via same `boolToInt(boolPtrTrue(...))` mapping ✅
- update `updateFieldsAllowed` whitelist includes both new columns ✅
- `fetchBotMemberUIDs` parity: character-identical across message/bot_api/incomingwebhook ✅
- Native-only `allowMention`: confirmed all 5 sibling adapters have `allowMention: false` (Go zero value) ✅

**C2 — Control-flow ordering / nested reuse + security control non-canonical probe:**
- `ExpandAisToBotUIDs` called once per request, no double-application risk ✅
- `dedupNonEmpty` handles leading/trailing spaces, blank strings, duplicates, over-limit correctly (verified by unit test) ✅
- `boolPtrTrue` handles nil / *false / *true correctly ✅

- Push endpoint: token-in-URL verified (constant-time) → creator membership → rate-limit → body parse → mention. All auth before mention. ✅
- Management endpoints: `resolveActor` (membership) + `requireOwnership` (creator or admin). Broadcast switches toggleable only by owners. ✅
- No new endpoint or capability exposed beyond existing webhook management surface.

**C4 — Authorization lifecycle / container-member state cascade:**
- Group disbanded/disabled → `cachedRequireActiveGroup` at step 2.5 returns 401 before mention processing ✅
- Creator left group → `cachedCreatorMembership` at step 3.5 returns `creatorIsMember=false` → lazy-disable webhook + 401 before mention processing ✅
- `incomingWebhookModel` has no independent container status beyond group status; group-level checks cover the cascade ✅

**C5 — Build/note ≠ runtime path correctness:**
- No build artifacts, CLI, extensions, or relative-URL concerns in this PR. SQL migration uses standard MySQL 8.0 INSTANT DDL. ✅ N/A

**C6 — Governance/policy/security document self-consistency:**
- PR includes `.octospec/tasks/incoming-webhook-mention/brief.md` and `context.yaml` — task specification documents, not governance/policy docs. ✅ No conflicts with existing SECURITY.md or disclosure policies.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. Two P2 items (missing `CloneForExpansion` parity, richtext+mention scope clarification) are maintainability improvements, not blockers. The security-sensitive attack surface is well-handled: membership gate, anti-enumeration, per-webhook capability bits, system-setting revocation, and native-only gating all follow sound trust-boundary design.

yujiawei (approved):
# Code Review — PR #445 (octo-server)

Reviewed at head SHA `28d7dff4f2896dac45e9983d31e48e6c4754d68d` against merge-base `8d2f4e3` (origin/main). `go build`, `go vet`, and the pure unit suite (`mention_test.go`, system-settings tests) pass locally. This is a security-sensitive ingress change (unauthenticated, token-in-URL push endpoint), so I weighted the trust-boundary paths heavily.

**Verdict: APPROVE.** The feature is correctly scoped and the trust boundary is well-handled. No blocking (P0/P1) issues. Three advisory (P2) items below, plus one item I'd like a human maintainer to explicitly confirm is the intended security posture.

## What the PR does (and does well)

- Adds a whitelisted `mention` structured field on the **native** push adapter only. Sibling adapters (github/wecom/feishu/gitlab/multica) construct fresh request structs and never carry `mention`; `buildPayload` still drops `Extra`. I could not find a non-native mention-smuggling path.
- Targeted `mention.uids` are deduped, capped, and filtered through a strict member gate (`filterGroupMembers`: `is_deleted=0 AND is_external=0 AND status=Normal`) before landing on the wire. Non-members are dropped **silently** (anti-enumeration), consistent with the endpoint's unified-401 posture.
- Broadcast forms (`@all`→`humans`, `@bots`→`ais`) are double-gated: per-webhook capability column AND a runtime permission, with capability-denied broadcasts reported back via `mention_ignored` (non-fatal 200 body).
- The `mention.uids` wire shape is `[]interface{}`, which is exactly what `pkg/mentionrewrite.ExpandAisToBotUIDs` requires to append bot UIDs in place; a parity unit test (`TestAssembleMentionFeedsExpandAis`) locks this against the real expander.
- Backward compatibility holds: absent `mention` → no `mention` key, original payload shape preserved.
- Migration is additive (`ADD COLUMN ... NOT NULL DEFAULT 0`, INSTANT on MySQL 8.0) with a correct `DROP COLUMN` Down.

## Findings

### P2 — Broadcast control defaults to permissive; confirm this is the intended posture
`modules/common/system_settings.go:572` sets `defaultIncomingWebhookMemberCanBroadcast = true`, and `create`/`update` (`api.go:673`, `api.go:817`) let any legitimate member (creator or admin, via `requireOwnership`) flip `allow_mention_all` / `allow_mention_bots` on their own webhook. At push time (`api.go:1329`):
```go
broadcastPermitted := w.settings.IncomingWebhookMemberCanBroadcast() || creatorIsAdmin
```
With the setting defaulting true, the `creatorIsAdmin` disjunct is normally a no-op: a regular member can stand up a webhook with both broadcast switches on, and any holder of that token (e.g. embedded in CI) can then broadcast `@all` / `@bots` without further authorization.

Why this is **not** rated blocking: the broadcast reminder/summon path (`modules/message/api_reminders.go` `getMention` / `mentionAnyBroadcast`) is sender-agnostic and has **no role gate** anywhere — a regular authenticated member can already `@all`/`@bots` in normal chat. So this webhook path grants no capability a member doesn't already have, and it actually *adds* controls (per-webhook switch defaulting off + an instant org-wide kill switch via `member_can_broadcast=0`, which strips member-created broadcasts at the read path with no data migration; admin-created webhooks are unaffected). The design is internally consistent and well-tested (`TestMentionPush_MemberBroadcastRevokedBySystemSetting`).

**For a human to confirm:** if the deployment intent is "broadcast is opt-out, revocable" then default-true is correct as shipped. If the intent is "broadcast is opt-in until an operator turns it on," flip the default to false (or require admin authorization to *set* the per-webhook switches). This is a one-line policy default, not a code defect — but it's load-bearing for the abuse model, so it should be a conscious operator choice rather than an inherited default.

### P2 — `@bots` expansion uses a looser member predicate than targeted `@uid`
`mention_expand.go:27` resolves bots via `groupService.GetMembers`, whose backing query (`modules/group/db.go:417` `queryMembersWithGroupNo`) filters only `is_deleted=0` — not `status=Normal` and not `is_external=0`. Targeted `mention.uids` go through the stricter `filterGroupMembers` (`is_external=0 AND status=Normal`). So a `@bots` broadcast can in principle stamp a blacklisted/non-normal or external bot member's UID into `mention.uids`, whereas the same UID supplied directly as a targeted mention would be dropped.

This asymmetry is **explicitly documented as intentional** in `db.go` (the `filterGroupMembers` comment) and is **parity with the existing user / bot-API ingresses**, which use the same `GetMembers` helper — so the bot set resolves identically across all ingresses, which is the stated parity goal. Not a regression. Flagging only so a maintainer is aware that "webhook ingress is the strictest boundary" is true for targeted mentions but not for the `@bots` broadcast expansion; if blacklisted-bot exclusion is ever required, it needs a module-level fix across all three ingresses, not just this PR.

### P2 — The "50" UID cap is an env-tunable floor, not a hard ceiling
`mention.go:14-22`: `maxMentionUIDs()` accepts any positive `OCTO_INCOMINGWEBHOOK_MAX_MENTION_UIDS`, so the real invariant is "50 unless an operator sets it higher," not "≤50." The 8KB native body cap bounds the absolute worst case regardless, so this is low-impact, but if "≤50" is part of the security contract for this unauthenticated path, clamp the env override to a hard maximum.

## Coverage notes

- I verified malformed `mention` shapes (`uids:"x"`, `uids:[1,2]`, `all:1`) fail `json.Unmarshal` and therefore 400 the whole native push at the parser. That's acceptable (fail-closed, no panic), but note it's a hard 400, not the "ignore malformed mention, deliver the message" behavior one might infer from the prose — the backward-compat guarantee only covers *absent* `mention`, not *malformed* `mention`. Worth a one-line doc clarification.
- Not verified end-to-end (no MySQL/Redis/WuKongIM in this review env): the persisted WuKongIM payload and actual red-dot / bot-wake behavior. The E2E tests in `mention_e2e_test.go` cover the HTTP-observable surface and look correct by inspection.
- Independent advisory pass: one external reviewer leg completed and surfaced the three items above (consensus with this review); a second cross-family leg timed out and is recorded as absent (not silent agreement). The findings here are my own adjudication against the diff and the linked task spec, which I independently re-checked.

yujiawei (changes requested):
# Code Review — PR #445 (octo-server) — Correction (review #2)

**This supersedes my earlier APPROVE on this same head SHA (`28d7dff`). Corrected verdict: REQUEST CHANGES.**

I have to retract my prior approval. In my first pass I verified that a malformed `mention` shape makes `json.Unmarshal` fail and 400s the whole push, but I filed it as an advisory "coverage note / doc clarification." That was wrong. It is a violation of a locked acceptance criterion in the task brief, which is a spec-compliance failure and therefore blocking. Another reviewer flagged it as the primary blocker and I should have weighted my own observation the same way.

## Blocking (spec compliance)

### P1 — Malformed optional `mention` rejects the entire push (violates brief acceptance #6)
The task brief's Acceptance list states verbatim:

> Empty/absent `mention`, malformed `mention.uids` → no panic, no `mention` key (full backward compatibility with existing native callers).

The implementation does not meet this. `parseNativePush` (`modules/incomingwebhook/adapter.go:87`) unmarshals the whole body straight into `pushPayloadReq`, whose `Mention *mentionReq` has `Uids []string` (`modules/incomingwebhook/model.go`). Any of these inputs makes `json.Unmarshal` return an error → `invalidReason="json"` → **HTTP 400, message not delivered**:

```
{"content":"hi","mention":{"uids":"u1"}}      → 400 (uids is string, want []string)
{"content":"hi","mention":{"uids":[1,2]}}     → 400 (uids elem is number, want string)
{"content":"hi","mention":{"uids":["u1",1]}}  → 400 (mixed array)
{"content":"hi","mention":{"all":1}}          → 400 (all is number, want bool)
{"content":"hi","mention":[]}                  → 400 (mention is array, want object)
```

I reproduced all of these against the exact struct shapes from the diff. The spec requires these to be **non-fatal** (drop the malformed mention, deliver the message with no `mention` key); instead they fail the push the message was carrying. The existing `TestMentionPush_NoMentionFieldBackwardCompatible` only covers the *absent*-field direction, so it does not catch this.

**Fix direction:** decode `mention` defensively rather than strictly — e.g. make `Mention json.RawMessage` (or give `mentionReq` a custom `UnmarshalJSON`) and, on a decode error, treat it as "no mention" instead of failing the whole request. Then add a test asserting each malformed shape above returns 200 with no `mention` key. (`all`/`bots` as truthy non-bool may also warrant lenient handling for symmetry, but `uids` is the one the acceptance bullet names explicitly.)

## Non-blocking (these stand from my first review)

- **P2 — Broadcast control defaults permissive.** `defaultIncomingWebhookMemberCanBroadcast = true` (`modules/common/system_settings.go:572`) means the `creatorIsAdmin` disjunct in `broadcastPermitted` (`api.go:1329`) is normally a no-op; any member can self-enable broadcast on their own webhook. This matches the brief's stated design ("the per-webhook default-off switch is the control, not a separate admin grant") and grants no capability a member lacks in normal chat (the reminder path has no role gate), and it adds an instant org-wide kill switch. Not a defect — but the default is load-bearing for the abuse model, so confirm default-true (opt-out, revocable) is the intended operator posture rather than an inherited default.
- **P2 — `@bots` expansion uses a looser member predicate** (`GetMembers` → `is_deleted=0` only) than targeted `@uid` (`filterGroupMembers` → `is_external=0 AND status=Normal`). Documented as intentional and parity with the user/bot-API ingresses; flagging only so the asymmetry is a conscious choice. Any tightening should be module-level across all three ingresses, not just here.
- **P2 — "50" UID cap is an env-tunable floor, not a hard ceiling** (`OCTO_INCOMINGWEBHOOK_MAX_MENTION_UIDS`, `mention.go:14`). Body cap bounds the worst case; clamp the override if ≤50 is part of the contract.
- **Nit** — the comment above `TestMention_UpdateTogglesSwitches` contradicts its assertions (the test expects a member to successfully update their own webhook); update the comment.

## What is solid (unchanged from first pass)

Native-only adapter gate is effective (no non-native smuggling path; `Extra` still discarded). Targeted `mention.uids` are server-side member-gated, deduped, capped, with silent non-member drop (anti-enumeration). Capability columns default 0; double-gated at push. The `[]interface{}` wire shape is correct for `ExpandAisToBotUIDs` and pinned by a parity test. Migration is additive/INSTANT with a symmetric Down. `go build` / `go vet` / pure unit suite pass locally.

## Coverage / process notes

- Independent advisory pass: one external reviewer leg completed (consensus on the findings above, and independently surfaced the malformed-mention blocker); a second cross-family leg timed out and is recorded as **absent**, not silent agreement.
- Not verified end-to-end (no MySQL/Redis/WuKongIM here): persisted WuKongIM payload and live red-dot / bot-wake behavior.

yujiawei:
Follow-up to my review — one additional P2 (non-blocking, no verdict change; the malformed-`mention` P1 in my review still stands as the blocker).

**P2 — missing `CloneForExpansion` before `ExpandAisToBotUIDs` (parity break)**

At `modules/incomingwebhook/api.go`:
```go
payload[mentionrewrite.MentionKey] = mention
payload = mentionrewrite.ExpandAisToBotUIDs(
    payload, common.ChannelTypeGroup.Uint8(), m.GroupNo, w.fetchBotMemberUIDs)
```
`ExpandAisToBotUIDs` mutates the `mention` sub-map in place. The three other ingress chokepoints all clone first and pass the clone to both the expander and the serializer:
- `modules/message/api.go:560`
- `modules/robot/api.go:536`
- `modules/bot_api/send.go:241`

Not a correctness bug in the webhook path today: `payload` is serialized via `util.ToJson(payload)` and never read again, so in-place mutation is unobservable. But it breaks a documented project-wide convention (the PR#145 doc-comments on those call sites), and anyone who later adds post-expansion logic referencing `payload` would silently get mutated data. One-line fix for parity: `wirePayload := mentionrewrite.CloneForExpansion(payload)`, then pass `wirePayload` to both `ExpandAisToBotUIDs` and `ToJson`.

an9xyz:
Thanks for the thorough reviews. Addressed in `6dd2a41` (force-pushed; kept as a single commit per this branch's convention).

**Blocking — fixed**
- **Malformed optional `mention` no longer fails the push** (acceptance #6). `pushPayloadReq.Mention` is now `json.RawMessage`, decoded leniently in `buildMention` via a new `decodeMention`: a bad shape (`uids` not an array, `mention` not an object, `all` not a bool, …) degrades to "no mention" and the message is still delivered. Added unit coverage (`TestParseNativePushMention` malformed cases + `TestDecodeMention`) and an e2e (`TestMentionPush_MalformedMentionIgnored`) asserting each previously-400 shape now returns 200 with the message delivered.

**Worth-fixing — done**
- Added `CloneForExpansion` before `ExpandAisToBotUIDs`, matching the message/robot/bot_api ingresses (the documented parity convention from #145).
- Pinned the native-only gate: `TestOnlyNativeAdapterAllowsMention` asserts `nativeAdapter` is the only adapter with `allowMention=true` (acceptance #4).
- Fixed the stale doc comment on `TestMention_UpdateTogglesSwitches`.
- Added a one-line note that mention applies to both text and richtext (msg_type-agnostic, intentional).

**Intentional / already-decided (no change)**
- `member_can_broadcast` defaulting on (a member can self-enable broadcast on a webhook they manage): a deliberate product choice — it grants nothing a member lacks in normal chat (the reminder path has no role gate), and ships with a per-webhook default-off switch, the targeted member gate, and an instant operator kill switch (flip the setting off → all member-created webhooks lose broadcast at the read path, no migration). 
- The all-bots expansion using the same `GetMembers` predicate as the other ingresses (looser than the targeted gate) is documented in `filterGroupMembers` and is the stated cross-ingress parity goal; tightening would be a module-level change across all three ingresses.
- The uid cap is env-tunable; the 8 KiB native body cap bounds the absolute worst case.

All real checks are green locally (build / vet / golangci / i18n + the full `incomingwebhook` package on MySQL + Redis + WuKongIM). The remaining red check is the `check-sprint` process gate (linked-issue requirement), which I'm sorting out separately.

---

OctoBoooot (approved):
## Review: feat(incomingwebhook): mention support on the native push endpoint (#445) — delta @ 6dd2a41d

**Verdict:** Approve.

Delta scope: prior REQUEST_CHANGES @ `28d7dff4` (1 major: malformed-mention → whole-push 400 violating brief acceptance #6) + 1 minor (adapter-parity test gap, acceptance #4) + mochashanyao's clone-consistency suggestion. Branch rebased into a single squashed commit at `6dd2a41d`. Every prior unresolved item is byte-verified addressed:

- ✅ **Malformed-mention regression (the main blocker)** — `model.go` changes `Mention` from `*mentionReq` to `json.RawMessage` (lazy decode). New `mention.go::decodeMention` does the targeted unmarshal and returns `(nil, false)` on shape errors; `buildMention` Warn-logs malformed and lets the message through. The exact shape suggested in my prior inline.
- ✅ **Wrong-type regression tests** — `mention_test.go::TestParseNativePushMention/malformed-does-NOT-fail-the-parse` covers 5 failing inputs (including my prior 3 — `mention:"please"`, `mention.uids:"alice"`, `mention.uids:[1,2]` — plus `mention.all:1` and `mention:[]`), asserting `invalid==""`. `TestDecodeMention` pins the lenient decode contract directly. `mention_e2e_test.go::TestMentionPush_MalformedMentionIgnored` runs the same five inputs through HTTP and asserts 200 + no `mention_ignored` field — load-bearing at the HTTP layer.
- ✅ **Adapter parity test (prior minor)** — `mention_test.go::TestOnlyNativeAdapterAllowsMention` iterates `{wecom,github,gitlab,feishu,multica}` and asserts `allowMention=false` on each, with explicit comment "a future edit to a platform adapter can't silently start emitting mentions." Pins acceptance #4.
- ✅ **mochashanyao's clone consistency** — `api.go:1337-1339` now calls `mentionrewrite.CloneForExpansion(payload)` before `ExpandAisToBotUIDs`, with comment naming the rationale (cross-ingress parity with message/robot/bot_api per PR #145, even though the webhook path's local payload doesn't currently have a downstream reader). Defends against the "future maintainer 'fixes' the inconsistency the wrong way" failure mode.

### Praise
- The lenient-decode pattern is the **minimal correct fix**: only `mention` is lazy, the rest of the body is still strict. The Warn-log-on-malformed (not silent drop) gives operators a debug signal without escalating to a delivery failure — preserves the brief's "fail open on mention, never on message" invariant.
- `TestMentionPush_MalformedMentionIgnored` is load-bearing at the HTTP boundary, not just the parser. It's the right level of test for a backward-compat acceptance criterion — pinning the wire-level guarantee a caller actually depends on.
- `TestOnlyNativeAdapterAllowsMention`'s reverse assertion ("must NOT process mention") is the right shape for the parity contract: it fails the moment any future contributor flips `allowMention: true` on a sibling adapter, which is exactly the regression mode worth guarding against.
- The `CloneForExpansion` integration with the "no current downstream reader, but守该约定" comment is exactly the right framing — it tells the next maintainer why the clone is here (consistency / future-proofing) so they don't "optimize it away" by removing it.

### Verdict-side gate (informational, not a finding)
GitHub `reviewDecision` still shows CHANGES_REQUESTED because prior CRs from Jerry-Xin + yujiawei + my own first review aren't auto-dismissed by a new SHA. The expectation is each of those reviewers re-reviews at `6dd2a41d` to clear their CR — same convergence pattern PR #442 went through.

Jerry-Xin (approved):
Re-reviewed at the new head `6dd2a41d`. **My prior 🔴 (malformed-mention → whole-push 400, violating brief acceptance #6) is genuinely FIXED (byte-verified, not a false-fix)** — flipping CHANGES_REQUESTED to APPROVE.

The fix matches exactly the direction suggested (lenient `json.RawMessage` capture + drop-on-decode-failure):
1. **`model.go:111`**: `Mention` field changed `*mentionReq` → `json.RawMessage` (`json:"mention,omitempty"`). `json.Unmarshal` into `pushPayloadReq` now accepts ANY valid-JSON value for `mention` (string/array/object) without type-checking — so `parseNativePush` no longer 400s on a malformed mention shape.
2. **`mention.go:18-26` `decodeMention`**: empty → `(nil,false)`; malformed → `json.Unmarshal` error → `(nil,false)`. Never propagates an error.
3. **`mention.go:69-77` `buildMention`**: on `!ok` with non-empty raw, logs and `return nil, nil` — the message **still delivers without a mention**, exactly per acceptance #6.
4. **Regression tests** (`mention_test.go:191-199`, `TestDecodeMention` :206): pin the exact wrong-type inputs (`{"mention":{"uids":"alice"}}`, `{"uids":[1,2]}`) → `decodeMention` returns `ok=false` gracefully, body parse does not fail. The previously-missing wrong-type coverage is now present.

So a malformed mention now: parses fine at the top level → fails lenient sub-decode → dropped with a Warn log → message delivers. Backward compatibility with existing native callers (who never sent `mention`) is preserved and the regression I flagged is closed at the source.

All the high-risk surfaces verified clean in my prior pass still hold (UID-only structured wire — no display-name injection; server-side membership filter via webhook's own group_no; 50-mention cap; double-gated default-deny broadcast; nil-safe). The earlier 🟡 (test-comment/assertion mismatch) — please confirm it was addressed; non-blocking either way.

No blocking items remain. ✅

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#445 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/445
**Head SHA:** `6dd2a41d3fdda87d7e683fea97f51bc53cef4289`
**Title:** feat(incomingwebhook): mention support on the native push endpoint
**Scope:** 15 files, +1113 / −42

---

## 1. Verification Summary

| # | Item | Status | Evidence |
|---|------|--------|----------|
| 1 | Wire contract matches `pkg/mentionrewrite` + `message/api_reminders.go getMention` consumer | ✅ | `assembleMention` produces `{uids: []interface{}, humans: 1, ais: 1}`; `getMention` reads exactly these keys (`modules/message/api_reminders.go`); `TestAssembleMentionFeedsExpandAis` locks parity with real `ExpandAisToBotUIDs` |
| 2 | Mention gated to native adapter only | ✅ | `nativeAdapter.allowMention=true`, all siblings at zero-value (`adapter.go:48`); `TestOnlyNativeAdapterAllowsMention` pins this |
| 3 | Targeted UIDs filtered to current internal-normal group members | ✅ | `filterGroupMembers` SQL: `is_deleted=0 AND is_external=0 AND status=Normal` (`db.go:237`); non-members silently dropped (anti-enumeration) |
| 4 | Broadcast gated by per-webhook capability bit AND system-setting policy | ✅ | `buildMention` passes `m.AllowMentionAll==1 && broadcastPermitted` (`mention.go:97`); `broadcastPermitted = settings.IncomingWebhookMemberCanBroadcast() \|\| creatorIsAdmin` (`api.go:1324`) |
| 5 | Revocation lever works (instant, no migration) | ✅ | `TestMentionPush_MemberBroadcastRevokedBySystemSetting` proves member webhook broadcast stripped when `member_can_broadcast=0`, admin webhook unaffected |
| 6 | Malformed mention degrades to no-mention (acceptance #6) | ✅ | `decodeMention` returns `(nil,false)` on any unmarshal error (`mention.go:32`); `TestParseNativePushMention` + `TestMentionPush_MalformedMentionIgnored` cover 5 malformed shapes |
| 7 | Backward compat (no mention field → unchanged payload) | ✅ | Empty `json.RawMessage` → `decodeMention` returns false → `buildMention` returns `(nil,nil)` → mention key never attached; `TestMentionPush_NoMentionFieldBackwardCompatible` pins this |
| 8 | `@所有AI` expansion parity with message/bot_api ingresses | ✅ | `fetchBotMemberUIDs` structurally identical to `modules/message/mention_expand.go` and `modules/bot_api/mention_expand.go`: same `GetMembers + ExistRobot` pair, same per-row best-effort error handling |
| 9 | SQL migration safe | ✅ | `ALTER TABLE ADD COLUMN SMALLINT NOT NULL DEFAULT 0` → MySQL 8.0 INSTANT algorithm; down migration drops both columns |
| 10 | `updateFieldsAllowed` allowlist updated | ✅ | `allow_mention_all` and `allow_mention_bots` added (`db.go:131-132`); no SQL injection vector |
| 11 | `CloneForExpansion` prevents in-place mutation of shared payload | ✅ | `pkg/mentionrewrite/clone.go` clones outer map + mention sub-map + uids slice; `ExpandAisToBotUIDs` mutates only the clone |
| 12 | `i18n-lint` file list includes new files | ✅ | `mention.go` and `mention_expand.go` added to `TestIncomingWebhookNoLegacyResponseError` file list (`api_i18n_test.go:24`) |

---

## 2. Findings

### P2-1: Targeted UID e2e coverage — no WuKongIM payload assertion

**Severity:** P2 (testing gap, not a code defect)
**Diff-scope:** new (this PR introduces the path)

`TestMentionPush_TargetedUIDsSucceed` sends `mention.uids=[memberAUID, not_a_member]` and verifies HTTP 200 + no `mention_ignored`, but does not inspect the WuKongIM-delivered payload to confirm `mention.uids` actually contains `memberAUID` and excludes `not_a_member`. The unit tests (`TestAssembleMention` "targeted uids: keep members, drop non-members") cover the logic, but the full HTTP→WuKongIM round-trip for targeted UIDs is not end-to-end verified.

**Recommendation:** Add a WuKongIM payload inspection assertion (if the test infra supports it) or document that targeted-UID wire shape is covered by the unit-test layer only. Not blocking.

### P2-2: `@所有AI` expansion may include external bot members while targeted @ excludes them

**Severity:** P2 (documented intentional asymmetry, not a defect)
**Diff-scope:** new (design decision in this PR)

`filterGroupMembers` excludes external members (`is_external=0`), so targeted `@uid` cannot reach external members. However, `fetchBotMemberUIDs → GetMembers` returns all non-deleted members (including external bots), matching the message/bot_api ingress behavior. The `filterGroupMembers` doc comment explicitly documents this as intentional ("定向保守 vs 广播对齐其它入口"). Consistent with existing ingress parity, but worth flagging as a design decision that future maintainers should be aware of.

---

## 3. Recommendations

1. **(Optional)** Consider adding a targeted-UID e2e test that inspects the WuKongIM payload bytes, closing the gap identified in P2-1.
2. **(Optional)** The `member_can_broadcast` system setting defaults to `true`. Consider whether the deployment guide should mention this as an explicit hardening step for high-security installations.

---

## 4. Extra Observations

- The `json.RawMessage` approach for the `mention` field in `pushPayloadReq` is well-designed: it decouples parse-time capture from structural validation, ensuring malformed mention never blocks message delivery (acceptance #6).
- The anti-enumeration design (silently dropping non-member UIDs without feedback) is a good security practice for a token-authenticated endpoint.
- The `member_can_broadcast` revocation lever is elegant: it ANDs at the push read path, so toggling the system setting instantly revokes broadcast for all member-created webhooks without any column migration.
- The `CloneForExpansion` pattern correctly prevents `ExpandAisToBotUIDs` from mutating the in-memory payload, with the wire bytes carrying the expanded bot UIDs.

---

## 5. Data-Flow Trace

|---|---|---|
| `req.Mention` (json.RawMessage) | `parseNativePush` → `json.Unmarshal(body, &req)` captures raw bytes | ✅ `adapter.go:97`; `json.RawMessage` type preserves bytes without structural validation |
| `mentionReq{Uids, All, Bots}` | `decodeMention` leniently unmarshals `req.Mention` | ✅ `mention.go:32`; returns `(nil,false)` on any error → degrades to no mention |
| `members` set for UID filter | `filterGroupMembers(groupNo, uids)` SQL query | ✅ `db.go:237`; `is_deleted=0 AND is_external=0 AND status=Normal`; on error → empty set → all targeted UIDs dropped |
| `creatorIsAdmin` | `cachedCreatorMembership → db.queryMemberRole` | ✅ `api.go:1200+`; checks `group_member` for active internal membership + creator/manager role; cached with TTL |
| `broadcastPermitted` | `settings.IncomingWebhookMemberCanBroadcast() \|\| creatorIsAdmin` | ✅ `api.go:1324`; system setting chain: DB → env → default(true) |
| `mention{uids,humans,ais}` map | `assembleMention` pure decision | ✅ `mention.go:121+`; uids as `[]interface{}` (ExpandAisToBotUIDs requirement); humans=1/ais=1 when capability+policy allow |
| Expanded bot UIDs | `ExpandAisToBotUIDs(wire, GROUP, groupNo, fetchBotMemberUIDs)` | ✅ `api.go:1340`; clone-before-expand; `fetchBotMemberUIDs` parity with message/bot_api verified structurally |
| `payload["mention"]` → WuKongIM | `SendMessageWithResult(MsgSendReq{Payload: payload})` | ✅ `api.go:1347`; payload serialized to JSON and sent; downstream `getMention` in `api_reminders.go` reads the same keys |
| `mention_ignored` in response | `buildMention` second return value → `respBody["mention_ignored"]` | ✅ `api.go:1374-1377`; only added when non-empty |

---

## 6. Blindspot Checklist (R5 — security_sensitive, C1–C6 all mandatory)

### C1 — Dual-Path Parity
- **create ↔ update for capability bits:** ✅ Both use `boolToInt(boolPtrTrue(req.*))` / `boolToInt(*req.*)`. `create` sets model fields; `update` adds to `fields` map which is allowlisted via `updateFieldsAllowed`. Parity confirmed.
- **adapter allow/deny:** ✅ Only `nativeAdapter` has `allowMention: true`. All five sibling adapters have zero-value `false`. `TestOnlyNativeAdapterAllowsMention` pins this.
- **create-gate ↔ execute-path authorization:** ✅ Create/update require authenticated group member (existing auth). Push uses webhook token (different auth). Broadcast capability is set at create/update time, enforced at push time. No second call-site that bypasses the gate.
- **Guard tool coverage:** N/A — no lint/guard tool claims in this PR.

### C2 — Control-Flow Ordering / Nesting
- **Mention insertion ordering:** ✅ Inserted AFTER all auth/rate-limit/group-active/creator-membership gates (step 6.5 in the pipeline). No gate can be bypassed by a mention-carrying request.
- **Clone-before-expand:** ✅ `CloneForExpansion` called before `ExpandAisToBotUIDs`. Only the clone is mutated. Original payload untouched (defense-in-depth).
- **`ExpandAisToBotUIDs` type gate:** ✅ Only accepts `[]interface{}` for uids. `assembleMention` constructs exactly this type. Verified by `ExpandAisToBotUIDs` source (`expand_ais.go`): any other type → no-op.
- **Non-canonical input testing:** ✅ `dedupNonEmpty` handles leading/trailing whitespace, empty strings, pure-whitespace strings. `decodeMention` rejects non-bool `all`/`bots`, non-string-array `uids`, non-object mention, arrays. Tests cover 5 malformed shapes.

### C3 — Authorization Boundary ≠ Capability Boundary
- **Who can reach the push endpoint?** Token holders (webhook token in URL path). Token is a shared secret, potentially leaked.
- **What can a token holder do with mention?**
  - Targeted @: only current internal-normal members of the webhook's own group (membership gate). Cannot reach cross-group or cross-Space identities. ✅
  - Broadcast @所有人: requires per-webhook `allow_mention_all=1` AND system policy permit. Default off. ✅
  - Broadcast @所有AI: requires per-webhook `allow_mention_bots=1` AND system policy permit. Default off. ✅
- **No escalation vector:** Mention does not grant any new capability beyond what the webhook token already allows (sending messages to the group). The mention only controls WHO in the group gets notified.

### C4 — Authorization Lifecycle / Container-Member State Cascade
- **Group disbanded/disabled:** ✅ `cachedRequireActiveGroup` (step 4, before mention processing) rejects pushes to inactive groups.
- **Creator removed from group:** ✅ `cachedCreatorMembership` (step 6) detects this, lazy-disables the webhook, returns 401. Mention code is never reached.
- **Creator demoted from admin:** Cache TTL means `creatorIsAdmin` may be stale for a short window. Acceptable — same caching pattern used throughout the module. The system setting `member_can_broadcast` provides an independent kill switch that takes effect on next `Reload()`.
- **Webhook disabled/deleted:** ✅ Status check in step 2 rejects with 401. Mention never processed.

### C5 — Build/Runtime Path Correctness
- **Go static compilation:** ✅ No runtime path resolution risks. New files (`mention.go`, `mention_expand.go`) are in the same package and will compile correctly.
- **No relative URLs or environment-dependent paths.** N/A for this Go backend PR.
- **PR reports tests pass:** `go test ./modules/incomingwebhook/... -count=1` → ok; `golangci-lint` → 0 issues; `make i18n-lint` → OK.

### C6 — Governance / Policy Document Consistency
- **`.octospec/tasks/incoming-webhook-mention/brief.md`:** Task spec document. Consistent with the implementation. No conflict with existing security policies.
- **`.octospec/tasks/incoming-webhook-mention/context.yaml`:** Rule context. Correctly identifies trust-boundary, space-isolation, and error-handling as load-bearing rules. Consistent with repo conventions.
- **No changes to SECURITY.md, CONTRIBUTING.md, or other governance documents.** No policy conflicts.

---

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. The PR demonstrates solid security engineering: well-layered trust boundaries (membership gate, capability bits, system-setting revocation lever), anti-enumeration design, proper adapter isolation, correct wire-contract parity with existing ingresses, and graceful degradation on every error path. Two P2 observations (targeted-UID e2e coverage gap and documented external-member asymmetry) are noted but non-blocking.

yujiawei (approved):
# Code Review — PR #445 (octo-server)

Reviewed at head SHA `6dd2a41d3fdda87d7e683fea97f51bc53cef4289` against merge-base `8d2f4e34`. Independent re-review on the current head (the branch was squashed to a single commit since the prior round). Build passes; `go test ./modules/incomingwebhook/...` passes (incl. the new mention suites). Cross-family second opinions were folded into the findings below.

## 1. Spec compliance

**Spec: ✅**

This implements exactly the linked task spec (caller-controlled `@mention` on the native push endpoint) — nothing missing, nothing out of scope. Acceptance criteria walked one by one:

- **Targeted `uids` → members only, deduped, capped, blanks/non-members dropped** — ✅ `dedupNonEmpty` (`mention.go`) trims + dedups + caps; `filterGroupMembers` (`db.go:222-226`, `is_deleted=0 AND is_external=0 AND status=Normal`) gates to current internal-normal members; non-members silently dropped in `assembleMention`.
- **`all` gated by `allow_mention_all`; denied → no `humans` + reported ignored** — ✅ `assembleMention` writes `humans=1` only when allowed, else appends to `ignored`; surfaced as `mention_ignored` in the 200 body.
- **`bots` gated by `allow_mention_bots`; allowed → `ExpandAisToBotUIDs` appends bot UIDs** — ✅ `ais=1` set only when allowed, then expanded on a clone via `CloneForExpansion` + `ExpandAisToBotUIDs`.
- **Non-native adapter with a `mention` field → no mention delivered** — ✅ gated by `pushAdapter.allowMention` (only `nativeAdapter` true); pinned by `TestOnlyNativeAdapterAllowsMention`.
- **create/update persist `allow_mention_*` (default 0); resp surfaces them** — ✅ wired in `create`/`update`/`toResp`; `updateFieldsAllowed` whitelist extended; migration adds both columns `DEFAULT 0`.
- **Empty/absent or malformed `mention` → no panic, no `mention` key, message still delivered** — ✅ **this was the prior-round blocker and is now resolved.** `Mention` is `json.RawMessage` (lazy capture); `decodeMention` returns `(nil,false)` on any malformed shape; `buildMention` degrades to no-mention and still delivers. The exact malformed inputs that previously 400'd the whole push (`{"uids":"alice"}`, `{"uids":[1,2]}`, `{"all":1}`, `mention:"please"`, `mention:[]`) are now covered by `TestParseNativePushMention` / `TestDecodeMention` and pass.

Trust-boundary handling is correct: `space_id` is server-derived (no caller override), `buildPayload` discards `Extra` (no smuggling — mention is added only as an explicit whitelisted field), and the creator-membership gate fails closed *before* the mention block, so a webhook whose creator left the group never reaches mention processing.

## 2. Code quality

**Quality: Approved**

No P0/P1. The findings below are all P2 (advisory) — none blocks merge.

- **P2 — Broadcast `humans=1` also summons OBO personas, blurring the two switches.** `allow_mention_all=1, allow_mention_bots=0` + `{"all":true}` sets `humans=1`, and `humans=1` is an explicit persona/clone fan-out trigger downstream (`obo_fanout.go:329`, `summonAll = mentionAis || mentionHumans`). So "humans-only" can still wake AI personas. This is **not introduced by this PR** — it is platform-wide read-side semantics: a normal member typing `@所有人` in plain chat produces the same `humans=1` and the same fan-out. The two webhook switches gate `humans` vs `ais` correctly at this ingress; the persona-clone coupling lives in the OBO read path and applies to every ingress equally. Worth a docstring note that `all` is not strictly human-only downstream; not a blocker.
- **P2 — UID cap is env-tunable above 50.** `OCTO_INCOMINGWEBHOOK_MAX_MENTION_UIDS` can raise the cap past the spec's 50. This matches the module's existing convention (`maxBytes`, `maxContentRunes`, `githubMaxBytes` are all env-tunable with safe defaults); the default is 50, satisfying the acceptance out of the box. If a hard ceiling is desired, clamp to `min(env, 50)`.
- **P2 — `mention.bots` expansion uses a wider membership filter than targeted `@uid`.** `fetchBotMemberUIDs` → `groupService.GetMembers` filters only `is_deleted=0`, so the bot broadcast can include external bot members, whereas targeted `@uid` excludes externals (`is_external=0`). This asymmetry is **intentional parity**: the webhook reuses the identical `GetMembers`+`ExistRobot` pair as the `message` and `bot_api` ingresses (see `mention_expand.go` docstring), so `@所有 AI` resolves to the same bot set everywhere. Both paths stay within the group (no cross-Space leak). Documented tradeoff, not a regression.
- **P2 — `member_can_broadcast` defaults to `true`.** Any member can opt their own webhook into `@all`/`@bots` unless ops flips the system setting off. This is the spec's locked design — the per-webhook, default-off switch is the control, and the system setting is an *additive* instant kill switch (flipping it to 0 strips broadcast from all member-created webhooks immediately, admin-created unaffected; verified by `TestMentionPush_MemberBroadcastRevokedBySystemSetting`). Reasonable posture given that an authenticated member can already `@所有人` in plain chat.
- **P2 — `os.Getenv(maxMentionUIDs)` evaluated per push.** Minor env-map read-lock contention on the hot path; same pattern as the other env-tunable thresholds in this module. Cache at init if profiling ever flags it.

### For human verification (security-sensitive)
- The `humans=1` → OBO persona fan-out coupling (P2 #1) is the one cross-module behavior worth a human eye: confirm product intent that a webhook with only `allow_mention_all` may still wake AI personas via the OBO read path. The webhook layer gates correctly; the coupling is downstream and pre-existing.
- Creator-admin status feeding `broadcastPermitted` is cached positive-only with a ≤1-TTL staleness window (a demoted admin's webhook retains broadcast for up to one TTL). This is the same accepted staleness as the existing display-identity override and does not create net-new escalation (the system-setting default is already permissive).

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. The prior-round acceptance-#6 blocker is fixed and explicitly tested; remaining items are advisory P2s. Tests, defensive posture, and trust-boundary handling are strong.

### Coverage / blind spots
- The new e2e tests assert HTTP-level behavior (`mention_ignored`, status codes) and the assembly unit tests pin the wire shape, but neither observes the actual WuKongIM-delivered payload bytes — final DB-filtered UID set, post-expansion `uids`, and non-native omission are proven by unit/handler tests rather than an end-to-end delivered-payload assertion.
- `modules/common` integration tests require a live DB and could not run in this sandbox (the `incomingwebhook` acceptance suite, which is the gate, passed); the new `member_can_broadcast` settings tests are standard fallback-chain tests consistent with the existing pattern.
- One advisory leg (cross-family) returned in a truncated/streaming form via gateway fallback; its substantive points (per-request `os.Getenv`) are captured above. Both advisory legs ran and were absorbed; no leg was treated as silent approval.

modules/incomingwebhook/adapter.go
an9xyz: Fixed in `6dd2a41`. `pushPayloadReq.Mention` is now `json.RawMessage`, decoded leniently in `buildMention` via `decodeMention`: a malformed shape (e.g. `uids` as a string/number, `mention` as an array) degrades to "no mention" and the message is still delivered — no 400. Regression coverage: `TestMentionPush_MalformedMentionIgnored` (e2e: each previously-400 body now returns 200) plus `TestParseNativePushMention` malformed cases and `TestDecodeMention` (unit). Resolving.

---

modules/incomingwebhook/mention_e2e_test.go
an9xyz: Added `TestOnlyNativeAdapterAllowsMention` (`mention_test.go`) in `6dd2a41`: it asserts `nativeAdapter.allowMention` is true and every sibling adapter (wecom/github/gitlab/feishu/multica) is false — so an accidental `allowMention=true` on a platform adapter would fail the test. Pins acceptance #4. Resolving.

---

PR #447, opened 2026-06-24
feat: serve Bot connect plugin package + API URL from backend (#446)

### LINKED ISSUE #446: App Bot connect guide: serve plugin package & API URL from backend instead of hardcoding in frontend
## Background

The App Bot "connect guide" (the `npx ... bind ...` command shown to admins so they can bind a Bot to an Agent) is currently assembled entirely in the **octo-admin** frontend:

`src/pages/AppBots/connectGuide.ts`
```ts
return `npx -y openclaw-channel-dmwork bind --bot-token ${token} --api-url ${apiUrl} --account-id ${botId} --agent <${agent}>`
```

Two volatile identifiers are baked into the frontend:

1. **`openclaw-channel-dmwork`** — the channel-adapter npm package name, hardcoded as a string literal. If the package is renamed (or a new package is rolled out for canary), the frontend must be edited and redeployed.
2. **`api-url`** — currently sourced from the frontend env `VITE_BOT_API_URL`, falling back to `window.location.origin`. That fallback is the **admin dashboard origin**, which is not necessarily the Bot API entry. When admin and bot-api are served from different hosts, the generated command is wrong.

These are server-owned facts. The frontend should not be the source of truth for them.

## Proposal

Add a `connect` object to the App Bot resource and include it in the relevant responses:

```jsonc
{
  // ...existing fields...
  "connect": {
    "plugin_package": "openclaw-channel-dmwork",   // channel adapter npm package, backend-configured
    "api_url": "https://<bot-api-host>"            // public Bot API entry for this environment
  }
}
```

### Endpoints affected

- `GET /v1/admin/app_bot/{id}`
- `POST /v1/admin/app_bot` (create response)
- `GET /v1/space/{spaceId}/app_bot/{id}`
- `POST /v1/space/{spaceId}/app_bot` (create response)

### Notes / constraints

- `plugin_package` must come from backend config (env/config), so a rename or canary rollout only touches the backend.
- `api_url` must be the **public Bot API entry**, not the admin origin. This also fixes the existing wrong-host fallback.
- **Do not** return the localized guide prose. The guide body is i18n (zh-CN / en-US) and stays in the frontend. Backend returns data only.
- `connect` must not contain secrets. The token continues to flow through the existing fields/endpoints under the current auth.

## Acceptance criteria

- [ ] All four endpoints above return a `connect` object with `plugin_package` and `api_url`.
- [ ] `plugin_package` is read from backend configuration, not hardcoded.
- [ ] `api_url` resolves to the environment's public Bot API entry.
- [ ] No secret values are added to `connect`.

## Frontend follow-up (octo-admin)

Tracked separately, depends on this change:

- `connectGuide.ts`: use `bot.connect?.plugin_package ?? 'openclaw-channel-dmwork'` and `bot.connect?.api_url ?? getBotApiUrl()`. The fallbacks keep current behavior until the backend ships, then switch over automatically with no further frontend change.

an9xyz:
## Summary

The App Bot "connect guide" (`npx -y <pkg> bind ...`) was assembled entirely in the octo-admin frontend, which hardcoded two server-owned facts: the channel-adapter npm package (`openclaw-channel-dmwork`, now unmaintained) and an `api_url` that fell back to the **admin dashboard origin** instead of the Bot API entry. This moves both to the backend so a package rename/canary or a split admin/bot-api host never needs a frontend redeploy.

The four App Bot read/create endpoints now return a `connect` object, and the package name becomes a single backend source of truth shared with BotFather's connect prompts (de-frozen from the i18n templates). The default package is the **maintained** `create-openclaw-octo` (the migration target off the dead `openclaw-channel-dmwork`).

## Related Issue

Closes #446 — backend scope. The octo-admin `connectGuide.ts` follow-up is tracked separately and depends on this change.

## Linked Spec

`.octospec/tasks/bot-connect-plugin-package/brief.md`

## Changes

- **`pkg/botutil/connect.go`** (new): single source of truth — `PluginPackage()` (default `create-openclaw-octo`, override via `OCTO_BOT_PLUGIN_PACKAGE`) and `DeriveAPIURL(cfg)` (public Bot API entry = `External.BaseURL`, fallback `http://<ip>:8090`).
- **`modules/app_bot/app_bot.go`**: `createBot` + `getBotDetail` now return `connect{plugin_package, api_url}` — covers all four endpoints (`GET`/`POST` × admin/space). Data only, no secrets.
- **`modules/botfather`**: de-hardcode `create-openclaw-octo` in the connect / created / quickstart / install i18n templates (zh-CN + en-US) → injected `{{.PluginPackage}}`.
- **`api_url` single source of truth**: every inline `External.BaseURL` / `http://<ip>:8090` copy now routes through `botutil.DeriveAPIURL` — across `app_bot`, `botfather` (`command.go` + `api.go`), and `bot_api` (`register.go` ×2 + `commands.go`). A grep for the inline pattern returns **zero** residual copies, so a future derivation change (TLS default, dev port, IPv6 bracketing) touches exactly one place.
- Tests: botutil unit; app_bot HTTP e2e (sqlmock + redis) for the read path **plus a `NewTestServer` create-side e2e**; botfather template parameterization + completeness probe + de-hardcode guard.

## Testing

- [x] Manually verified

Full affected suites green under the CI recipe (`OCTO_MASTER_KEY` + per-package `test` DB reset to `utf8mb4_general_ci` + `-race`), against locally-provisioned MySQL 8.0 + Redis + WuKongIM:

```
pkg/botutil ✓ · modules/app_bot ✓ · modules/botfather ✓ · modules/bot_api ✓
```

Plus `go build ./...`, `go vet ./...`, `golangci-lint` (0 issues), `make i18n-extract-check`, `make i18n-lint`, and a grep confirming zero residual inline `api_url` derivations.

## COMPREHENSION

1. **What this change actually does to the load-bearing path** — it adds a `connect{plugin_package, api_url}` field to two App Bot **success** responses (`createBot`, `getBotDetail`, covering all four endpoints), replaces the frozen package literal in four BotFather i18n templates with an injected `{{.PluginPackage}}`, and routes every `api_url` derivation through one helper. It does NOT touch any auth gate (`checkSpaceAdmin` / `CheckLoginRole`), the `botInRouteScope` cross-tenant IDOR guard, or the i18n **error** envelope — `connect` is success-response data, so no `pkg/errcode` code / `active.zh-CN.toml` / `i18n-extract` is involved.

2. **What could break** — (a) the BotFather msgtmpl catalog runs with `missingkey=error`, so any render site that omits `PluginPackage` would error and silently drop the DM; mitigated by injecting it at all four `command.go` sites **and** the completeness probe, with a guard test. (b) Response is additive (new key only), so existing consumers are unaffected; `connect` deliberately excludes the token (asserted) so no secret widening. (c) The `api_url` consolidation is byte-for-byte behavior-preserving — `DeriveAPIURL` reproduces the exact `External.BaseURL` / `IP:8090` logic each callsite had inline (and that the adjacent `DeriveWSURL` already used), so the connect guide, BotFather docs, and `/bot/register` stay consistent; `TestDeriveAPIURL` pins the helper.

3. **How I know it works** — app_bot HTTP e2e asserts `connect` is present on **both** the read and create responses, that an `OCTO_BOT_PLUGIN_PACKAGE` override reaches the client end-to-end, and that no token leaks into `connect`; the botfather tests assert the injected package renders in the bind/install/quickstart commands and that the old literal no longer appears; a grep confirms zero residual inline derivations; the full `modules/app_bot`, `modules/botfather`, and `modules/bot_api` suites pass under the CI recipe with the race detector enabled.

## Checklist

- [x] Updated documentation

https://claude.ai/code/session_01NgRGrhti2U882uGwujGgK5

### DIFF
--- a/modules/botfather/api.go
+++ b/modules/botfather/api.go
@@ -2,7 +2,6 @@ package botfather
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"strings"
 	"sync"
@@ -107,10 +106,7 @@ func (bf *BotFather) Route(r *wkhttp.WKHttp) {
 // skillMD 返回skill.md文档
 func (bf *BotFather) skillMD(c *wkhttp.Context) {
 	cfg := bf.ctx.GetConfig()
-	apiURL := cfg.External.BaseURL
-	if strings.TrimSpace(apiURL) == "" {
-		apiURL = fmt.Sprintf("http://%s:8090", cfg.External.IP)
-	}
+	apiURL := botutil.DeriveAPIURL(cfg)
 	wsURL := botutil.DeriveWSURL(cfg)
 	content := generateSkillMD(apiURL, wsURL)
 	c.Header("Content-Type", "text/markdown; charset=utf-8")
@@ -126,11 +122,7 @@ func (bf *BotFather) cliGuideMD(c *wkhttp.Context) {
 
 // setupNewbotMD 返回 /newbot 设置流程文档
 func (bf *BotFather) setupNewbotMD(c *wkhttp.Context) {
-	cfg := bf.ctx.GetConfig()
-	apiURL := cfg.External.BaseURL
-	if strings.TrimSpace(apiURL) == "" {
-		apiURL = fmt.Sprintf("http://%s:8090", cfg.External.IP)
-	}
+	apiURL := botutil.DeriveAPIURL(bf.ctx.GetConfig())
 	content := generateSetupNewbotMD(apiURL)
 	c.Header("Content-Type", "text/markdown; charset=utf-8")
 	c.String(http.StatusOK, content)
@@ -138,11 +130,7 @@ func (bf *BotFather) setupNewbotMD(c *wkhttp.Context) {
 
 // setupQuickstartMD 返回 /quickstart 设置流程文档
 func (bf *BotFather) setupQuickstartMD(c *wkhttp.Context) {
-	cfg := bf.ctx.GetConfig()
-	apiURL := cfg.External.BaseURL
-	if strings.TrimSpace(apiURL) == "" {
-		apiURL = fmt.Sprintf("http://%s:8090", cfg.External.IP)
-	}
+	apiURL := botutil.DeriveAPIURL(bf.ctx.GetConfig())
 	content := generateSetupQuickstartMD(apiURL)
 	c.Header("Content-Type", "text/markdown; charset=utf-8")
 	c.String(http.StatusOK, content)
--- a/pkg/botutil/connect.go
+++ b/pkg/botutil/connect.go
@@ -23,7 +23,7 @@ const DefaultPluginPackage = "create-openclaw-octo"
 const PluginPackageEnv = "OCTO_BOT_PLUGIN_PACKAGE"
 
 // PluginPackage returns the configured OpenClaw plugin package name, honoring
-// the DM_BOT_PLUGIN_PACKAGE override and falling back to DefaultPluginPackage.
+// the OCTO_BOT_PLUGIN_PACKAGE override and falling back to DefaultPluginPackage.
 func PluginPackage() string {
 	if v := strings.TrimSpace(os.Getenv(PluginPackageEnv)); v != "" {
 		return v
--- a/pkg/botutil/connect.go
+++ b/pkg/botutil/connect.go
@@ -1,0 +1,43 @@
+package botutil
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/Mininglamp-OSS/octo-lib/config"
+)
+
+// DefaultPluginPackage is the npm package an Agent runs (via `npx -y <pkg> ...`)
+// to install the OpenClaw channel adapter and bind a Bot. It is the single
+// backend-owned source of truth for the package name: the App Bot connect
+// responses (modules/app_bot) and the BotFather connect prompts
+// (modules/botfather) both resolve it through PluginPackage, so a package
+// rename or canary rollout only touches the backend — never the frontend, and
+// never the localized connect-guide prose.
+const DefaultPluginPackage = "create-openclaw-octo"
+
+// PluginPackageEnv overrides DefaultPluginPackage at runtime (rename / canary),
+// so operators can switch the package without a code change or redeploy of the
+// frontend.
+const PluginPackageEnv = "OCTO_BOT_PLUGIN_PACKAGE"
+
+// PluginPackage returns the configured OpenClaw plugin package name, honoring
+// the DM_BOT_PLUGIN_PACKAGE override and falling back to DefaultPluginPackage.
+func PluginPackage() string {
+	if v := strings.TrimSpace(os.Getenv(PluginPackageEnv)); v != "" {
+		return v
+	}
+	return DefaultPluginPackage
+}
+
+// DeriveAPIURL resolves the public Bot API entry for this deployment: the
+// configured External.BaseURL, falling back to the direct-access host:port when
+// BaseURL is unset. This is the value Agents pass as `--api-url` and the URL
+// returned in Bot connect/register responses — NOT the admin-dashboard origin.
+func DeriveAPIURL(cfg *config.Config) string {
+	if u := strings.TrimSpace(cfg.External.BaseURL); u != "" {
+		return u
+	}
+	return fmt.Sprintf("http://%s:8090", cfg.External.IP)
+}
--- a/.octospec/tasks/bot-connect-plugin-package/brief.md
+++ b/.octospec/tasks/bot-connect-plugin-package/brief.md
@@ -1,0 +1,105 @@
+---
+type: Task
+title: "Task: bot-connect-plugin-package"
+description: Serve the Bot connect plugin package + public API URL from the backend instead of hardcoding them in the frontend.
+tags: ["app-bot", "botfather", "bot-connect", "plugin-package", "i18n", "wire-contract"]
+timestamp: 2026-06-24T00:00:00Z
+# --- octospec extension fields ---
+slug: bot-connect-plugin-package
+upstream: "octo-server#446"
+source: self
+---
+# Task: bot-connect-plugin-package
+
+> One task = one `.octospec/tasks/<slug>/` directory. This brief is the spec for
+> the work. AI may draft it from existing code; a human confirms it.
+
+Upstream: octo-server #446 — the App Bot "connect guide" (`npx -y <pkg> bind ...`)
+is assembled entirely in the octo-admin frontend, which hardcodes two server-owned
+facts: the channel-adapter npm package (`openclaw-channel-dmwork`, now unmaintained)
+and an `api_url` that falls back to the **admin dashboard origin** rather than the
+Bot API entry. Both must move to the backend so a package rename/canary or a
+split admin/bot-api host never requires a frontend redeploy.
+
+## Goal
+The four App Bot read/create endpoints return a `connect` object so the frontend
+stops owning these facts:
+
+```jsonc
+"connect": { "plugin_package": "create-openclaw-octo", "api_url": "https://<bot-api>" }
+```
+
+- `plugin_package` is backend-configured: default the **maintained** package
+  `create-openclaw-octo` (the one BotFather already emits — the migration target
+  off the dead `openclaw-channel-dmwork`), overridable via `OCTO_BOT_PLUGIN_PACKAGE`.
+- `api_url` resolves to the **public Bot API entry** (`External.BaseURL`, fallback
+  `http://<ip>:8090`), never the admin origin.
+- The package name becomes a single backend source of truth (`pkg/botutil`) shared
+  with BotFather's connect prompts, so a rename touches exactly one place.
+- Data only: no localized guide prose, no token or other secret in `connect`.
+
+## Background
+- App Bot handlers: `modules/app_bot/app_bot.go` `createBot` (covers `POST
+  /v1/admin/app_bot` + `POST /v1/space/:space_id/app_bot`) and `getBotDetail`
+  (covers `GET /v1/admin/app_bot/:id` + `GET /v1/space/:space_id/app_bot/:id`).
+  Two response builders cover all four endpoints.
+- New source of truth: `pkg/botutil/connect.go` — `PluginPackage()` (env
+  `OCTO_BOT_PLUGIN_PACKAGE` → default `create-openclaw-octo`) and
+  `DeriveAPIURL(cfg)` (mirrors the existing `External.BaseURL`/`IP:8090` logic in
+  `bot_api/register.go` and `botfather/command.go`).
+- BotFather: the package name was frozen in the i18n templates
+  `modules/botfather/templates/{zh-CN,en-US}/command.tmpl` (connect_prompt /
+  created_prompt / quickstart / install) and rendered via `command.go`. These are
+  the msgtmpl outbound-message catalog (#304, the "fourth i18n category"), NOT the
+  HTTP error envelope.
+- Issue: octo-server #446.
+
+## Load-bearing list
+- **wire-contract**: `connect{plugin_package, api_url}` is a NEW response field on
+  4 endpoints consumed by octo-admin; it must be data-only, contain no secret, and
+  always be present. It is a **success-response** field — it does NOT pass through
+  the i18n error envelope, so no `pkg/errcode` code / `active.zh-CN.toml` /
+  `i18n-extract` is involved. (rules: error-handling — wire-contract)
+- **space**: `getBotDetail`/`createBot` keep their existing gates
+  (`checkSpaceAdmin` for the space route, `CheckLoginRole` for the platform route)
+  and the `botInRouteScope` cross-tenant IDOR guard. `connect` is identical
+  regardless of scope/tenant and adds no per-tenant data, so isolation is
+  unchanged — the field must not widen any query or gate. (rules: space-isolation)
+- **i18n (msgtmpl catalog)**: BotFather connect/created/quickstart/install
+  templates now render the injected `{{.PluginPackage}}`. The catalog runs with
+  `missingkey=error`, so every render site (`command.go` ×4) AND the completeness
+  probe must supply `PluginPackage`. (rules: error-handling — i18n)
+- **secret-handling**: `connect` must never carry the bot token; the masked
+  top-level `token` field stays unchanged.
+
+## Out of scope
+- octo-admin frontend (`connectGuide.ts`) — `bot.connect?.plugin_package ?? …` /
+  `?? getBotApiUrl()`; tracked separately under #446, ships after this.
+- `modules/botfather/skill.go` — the deprecated, static skill-doc generator still
+  hardcodes `create-openclaw-octo`; it is not part of the i18n connect-guide
+  catalog and is left for a separate decision.
+- `modules/bot_api/register.go` inline `api_url` fallback — behaviorally identical
+  to `DeriveAPIURL`, left unconsolidated to keep scope to app_bot + botfather.
+- octo-lib `config.Config` changes — the package name is a module-level env knob,
+  no cross-repo change.
+- Returning the localized guide prose from the backend (explicitly forbidden by #446).
+
+## Acceptance
+- All four endpoints return `connect{plugin_package, api_url}`; `plugin_package`
+  is read from config (default `create-openclaw-octo`, honors
+  `OCTO_BOT_PLUGIN_PACKAGE`); `api_url` is the public Bot API entry.
+- `connect` carries no secret: exactly the two public keys, token never present.
+- New tests pass:
+  - `pkg/botutil`: default / env-override / `DeriveAPIURL` (BaseURL + IP fallback).
+  - `modules/app_bot`: HTTP e2e (sqlmock+redis) — `connect` present, an
+    `OCTO_BOT_PLUGIN_PACKAGE` override reaches the client, and no token leaks into
+    `connect`.
+  - `modules/botfather`: the connect-guide templates render the injected package
+    (bind/install/quickstart), the completeness probe still renders, and a guard
+    asserts the old `create-openclaw-octo` literal no longer appears when a
+    different package is injected.
+- `go build ./...`, `go vet ./...`, `golangci-lint run`, `make i18n-extract-check`,
+  `make i18n-lint` all pass.
+- Full affected suites green under the CI recipe (master key + per-package DB
+  reset + `-race`): `pkg/botutil`, `modules/app_bot`, `modules/botfather`,
+  `modules/bot_api`.
--- a/.octospec/tasks/bot-connect-plugin-package/brief.md
+++ b/.octospec/tasks/bot-connect-plugin-package/brief.md
@@ -71,6 +71,12 @@ stops owning these facts:
   probe must supply `PluginPackage`. (rules: error-handling — i18n)
 - **secret-handling**: `connect` must never carry the bot token; the masked
   top-level `token` field stays unchanged.
+- **api_url single-source-of-truth**: `DeriveAPIURL(cfg)` replaces every inline
+  `External.BaseURL` / `http://<ip>:8090` copy across `app_bot`, `botfather`
+  (`command.go` + `api.go`) and `bot_api` (`register.go` + `commands.go`) —
+  behavior-identical, so the connect guide, BotFather docs, and `/bot/register`
+  stay consistent and a future derivation change touches exactly one place.
+  (rules: testing — `TestDeriveAPIURL`)
 
 ## Out of scope
 - octo-admin frontend (`connectGuide.ts`) — `bot.connect?.plugin_package ?? …` /
@@ -78,8 +84,6 @@ stops owning these facts:
 - `modules/botfather/skill.go` — the deprecated, static skill-doc generator still
   hardcodes `create-openclaw-octo`; it is not part of the i18n connect-guide
   catalog and is left for a separate decision.
-- `modules/bot_api/register.go` inline `api_url` fallback — behaviorally identical
-  to `DeriveAPIURL`, left unconsolidated to keep scope to app_bot + botfather.
 - octo-lib `config.Config` changes — the package name is a module-level env knob,
   no cross-repo change.
 - Returning the localized guide prose from the backend (explicitly forbidden by #446).
@@ -100,6 +104,8 @@ stops owning these facts:
     different package is injected.
 - `go build ./...`, `go vet ./...`, `golangci-lint run`, `make i18n-extract-check`,
   `make i18n-lint` all pass.
+- No inline `api_url` derivation remains: a grep for the `External.BaseURL` /
+  `http://<ip>:8090` fallback returns only `botutil.DeriveAPIURL`.
 - Full affected suites green under the CI recipe (master key + per-package DB
   reset + `-race`): `pkg/botutil`, `modules/app_bot`, `modules/botfather`,
   `modules/bot_api`.
--- a/modules/app_bot/app_bot.go
+++ b/modules/app_bot/app_bot.go
@@ -19,6 +19,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/modules/bot_api"
 	"github.com/Mininglamp-OSS/octo-server/modules/space"
 	"github.com/Mininglamp-OSS/octo-server/modules/user"
+	"github.com/Mininglamp-OSS/octo-server/pkg/botutil"
 	appwkhttp "github.com/Mininglamp-OSS/octo-server/pkg/wkhttp"
 	"github.com/gin-gonic/gin"
 	"go.uber.org/zap"
@@ -421,9 +422,10 @@ func (ab *AppBot) createBot(c *wkhttp.Context, scope, spaceID string) {
 	}
 
 	c.Response(gin.H{
-		"id":    req.ID,
-		"uid":   uid,
-		"token": token,
+		"id":      req.ID,
+		"uid":     uid,
+		"token":   token,
+		"connect": ab.connectInfo(),
 	})
 }
 
@@ -530,6 +532,7 @@ func (ab *AppBot) getBotDetail(c *wkhttp.Context) {
 		"created_by":   bot.CreatedBy,
 		"created_at":   bot.CreatedAt,
 		"updated_at":   bot.UpdatedAt,
+		"connect":      ab.connectInfo(),
 	})
 }
 
@@ -983,6 +986,19 @@ func (ab *AppBot) discoverBots(c *wkhttp.Context) {
 }
 
 // ==================== Helpers ====================
+
+// connectInfo returns the backend-owned facts the frontend needs to assemble the
+// Bot connect guide: the OpenClaw plugin package an Agent runs and this
+// environment's public Bot API entry. Both are server-owned (config/env), so a
+// package rename, canary rollout, or host change never requires a frontend
+// redeploy. It carries data only — the localized guide prose stays in the
+// frontend, and it never includes the token or any other secret.
+func (ab *AppBot) connectInfo() gin.H {
+	return gin.H{
+		"plugin_package": botutil.PluginPackage(),
+		"api_url":        botutil.DeriveAPIURL(ab.ctx.GetConfig()),
+	}
+}
 
 func (ab *AppBot) toBotListResp(bots []*appBotModel) []gin.H {
 	result := make([]gin.H, 0, len(bots))
--- a/modules/app_bot/app_bot_connect_test.go
+++ b/modules/app_bot/app_bot_connect_test.go
@@ -1,0 +1,95 @@
+package app_bot
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/Mininglamp-OSS/octo-server/pkg/botutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestConnectInfo_DefaultAndOverride pins the value contract of the connect
+// object: plugin_package is the backend default (overridable by env) and api_url
+// is the configured public Bot API entry — never the admin origin.
+func TestConnectInfo_DefaultAndOverride(t *testing.T) {
+	cfg := config.New()
+	cfg.Test = true
+	cfg.External.BaseURL = "https://bot-api.example.com"
+	ab := &AppBot{ctx: config.NewContext(cfg)}
+
+	info := ab.connectInfo()
+	assert.Equal(t, "create-openclaw-octo", info["plugin_package"], "default plugin package")
+	assert.Equal(t, "https://bot-api.example.com", info["api_url"], "api_url is the configured Bot API entry")
+	// connect carries data only — no token / secret keys.
+	assert.Len(t, info, 2, "connect must contain exactly plugin_package + api_url")
+
+	t.Setenv(botutil.PluginPackageEnv, "create-openclaw-canary")
+	info = ab.connectInfo()
+	assert.Equal(t, "create-openclaw-canary", info["plugin_package"], "env override flows through")
+}
+
+// serveGetBotDetail mounts the platform admin route with a published platform bot
+// stubbed in sqlmock and returns the recorded GET /v1/admin/app_bot/{id} response.
+func serveGetBotDetail(t *testing.T, id string) *httptest.ResponseRecorder {
+	t.Helper()
+	route, mock, cleanup := newPlatformGateRouteWithMock(t, string(wkhttp.Admin))
+	t.Cleanup(cleanup)
+
+	// No WithArgs: dbr may interpolate the WHERE value into the SQL (0 bound
+	// args) rather than bind it, depending on process state; this test asserts
+	// the connect payload, not the WHERE clause, so match on the query shape only.
+	mock.ExpectQuery("app_bot").
+		WillReturnRows(sqlmock.NewRows(
+			[]string{"id", "uid", "display_name", "description", "scope", "space_id", "status", "token", "welcome_msg", "created_by"}).
+			AddRow(id, "app_"+id+"_bot", "My Bot", "desc", "platform", "", 1, "super-secret-token", "", "creator-1"))
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/v1/admin/app_bot/"+id, nil)
+	req.Header.Set("token", testutil.Token)
+	route.ServeHTTP(w, req)
+	return w
+}
+
+// TestGetBotDetail_IncludesConnect is the endpoint-level proof that the four App
+// Bot read/create endpoints (getBotDetail covers both GETs) now return a connect
+// object, that it carries no secret, and — via the env override — that the
+// package name is server-configured, the core of issue #446.
+func TestGetBotDetail_IncludesConnect(t *testing.T) {
+	w := serveGetBotDetail(t, "bot-1")
+	require.Equal(t, http.StatusOK, w.Code, "admin reaches the bot detail")
+
+	var resp struct {
+		Token   string         `json:"token"`
+		Connect map[string]any `json:"connect"`
+	}
+	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
+
+	require.NotNil(t, resp.Connect, "response carries a connect object")
+	assert.Equal(t, "create-openclaw-octo", resp.Connect["plugin_package"])
+	assert.NotEmpty(t, resp.Connect["api_url"], "api_url resolves to the public Bot API entry")
+	// No secret leaks into connect: exactly the two public keys, and the raw
+	// token never appears there (the top-level token stays masked).
+	assert.Len(t, resp.Connect, 2)
+	assert.NotContains(t, resp.Connect, "token")
+	assert.NotEqual(t, "super-secret-token", resp.Connect["api_url"])
+}
+
+func TestGetBotDetail_ConnectHonorsPackageOverride(t *testing.T) {
+	t.Setenv(botutil.PluginPackageEnv, "create-openclaw-canary")
+	w := serveGetBotDetail(t, "bot-2")
+	require.Equal(t, http.StatusOK, w.Code)
+
+	var resp struct {
+		Connect map[string]any `json:"connect"`
+	}
+	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
+	assert.Equal(t, "create-openclaw-canary", resp.Connect["plugin_package"],
+		"a configured package name reaches the client without a frontend change")
+}
--- a/modules/app_bot/app_bot_connect_test.go
+++ b/modules/app_bot/app_bot_connect_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"github.com/DATA-DOG/go-sqlmock"
@@ -57,10 +58,11 @@ func serveGetBotDetail(t *testing.T, id string) *httptest.ResponseRecorder {
 	return w
 }
 
-// TestGetBotDetail_IncludesConnect is the endpoint-level proof that the four App
-// Bot read/create endpoints (getBotDetail covers both GETs) now return a connect
-// object, that it carries no secret, and — via the env override — that the
-// package name is server-configured, the core of issue #446.
+// TestGetBotDetail_IncludesConnect is the read-path proof: getBotDetail backs
+// both GET /v1/admin/app_bot/:id and GET /v1/space/:space_id/app_bot/:id, and
+// must return a connect object that carries no secret — and, via the env
+// override, that the package name is server-configured (the core of #446). The
+// create path (POST) is pinned separately by TestCreatePlatformBot_IncludesConnect_E2E.
 func TestGetBotDetail_IncludesConnect(t *testing.T) {
 	w := serveGetBotDetail(t, "bot-1")
 	require.Equal(t, http.StatusOK, w.Code, "admin reaches the bot detail")
@@ -92,4 +94,46 @@ func TestGetBotDetail_ConnectHonorsPackageOverride(t *testing.T) {
 	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
 	assert.Equal(t, "create-openclaw-canary", resp.Connect["plugin_package"],
 		"a configured package name reaches the client without a frontend change")
+}
+
+// TestCreatePlatformBot_IncludesConnect_E2E pins the create-side wire contract
+// (POST /v1/admin/app_bot) against the full server + real MySQL/IM
+// (NewTestServer): a successful create returns the same connect object, data
+// only, with the freshly-minted token never leaking into connect. Complements
+// the read-path coverage in TestGetBotDetail_IncludesConnect.
+func TestCreatePlatformBot_IncludesConnect_E2E(t *testing.T) {
+	t.Setenv("OCTO_MASTER_KEY", "0123456789abcdef0123456789abcdef")
+
+	s, ctx := testutil.NewTestServer()
+	require.NoError(t, testutil.CleanAllTables(ctx))
+
+	// Log in as a system admin (the test server wires no RoleResolver, so the
+	// token's baked role is authoritative).
+	require.NoError(t, ctx.Cache().Set(
+		ctx.GetConfig().Cache.TokenCachePrefix+testutil.Token,
+		testutil.UID+"@test@"+string(wkhttp.Admin)))
+
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest(http.MethodPost, "/v1/admin/app_bot",
+		strings.NewReader(`{"id":"connect-create-e2e","display_name":"Connect Create"}`))
+	req.Header.Set("token", testutil.Token)
+	req.Header.Set("Content-Type", "application/json")
+	s.GetRoute().ServeHTTP(w, req)
+
+	require.Equal(t, http.StatusOK, w.Code, "admin create should succeed: %s", w.Body.String())
+
+	var resp struct {
+		Token   string         `json:"token"`
+		Connect map[string]any `json:"connect"`
+	}
+	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
+	require.NotNil(t, resp.Connect, "create response carries a connect object")
+	assert.Equal(t, "create-openclaw-octo", resp.Connect["plugin_package"])
+	assert.NotEmpty(t, resp.Connect["api_url"], "api_url resolves to the public Bot API entry")
+	// Data only: exactly the two public keys, and the freshly-minted token
+	// (returned at top level on create) must never appear inside connect.
+	assert.Len(t, resp.Connect, 2)
+	assert.NotContains(t, resp.Connect, "token")
+	assert.NotEqual(t, resp.Token, resp.Connect["api_url"])
+	assert.NotEqual(t, resp.Token, resp.Connect["plugin_package"])
 }
--- a/modules/bot_api/commands.go
+++ b/modules/bot_api/commands.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/pkg/botutil"
 	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
 	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
 	"github.com/gin-gonic/gin"
@@ -84,11 +85,7 @@ func (ba *BotAPI) getUserInfo(c *wkhttp.Context) {
 		return
 	}
 
-	cfg := ba.ctx.GetConfig()
-	apiURL := cfg.External.BaseURL
-	if strings.TrimSpace(apiURL) == "" {
-		apiURL = fmt.Sprintf("http://%s:8090", cfg.External.IP)
-	}
+	apiURL := botutil.DeriveAPIURL(ba.ctx.GetConfig())
 
 	c.JSON(http.StatusOK, gin.H{
 		"uid":    userResp.UID,
--- a/modules/bot_api/register.go
+++ b/modules/bot_api/register.go
@@ -1,7 +1,6 @@
 package bot_api
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/Mininglamp-OSS/octo-lib/config"
@@ -103,10 +102,7 @@ func (ba *BotAPI) registerUserBot(c *wkhttp.Context, token string) {
 	}
 
 	cfg := ba.ctx.GetConfig()
-	apiURL := cfg.External.BaseURL
-	if strings.TrimSpace(apiURL) == "" {
-		apiURL = fmt.Sprintf("http://%s:8090", cfg.External.IP)
-	}
+	apiURL := botutil.DeriveAPIURL(cfg)
 	wsURL := botutil.DeriveWSURL(cfg)
 
 	botName := ""
@@ -163,10 +159,7 @@ func (ba *BotAPI) registerAppBot(c *wkhttp.Context, token string) {
 	}
 
 	cfg := ba.ctx.GetConfig()
-	apiURL := cfg.External.BaseURL
-	if strings.TrimSpace(apiURL) == "" {
-		apiURL = fmt.Sprintf("http://%s:8090", cfg.External.IP)
-	}
+	apiURL := botutil.DeriveAPIURL(cfg)
 	wsURL := botutil.DeriveWSURL(cfg)
 
 	c.Response(&BotRegisterResp{
--- a/modules/botfather/command.go
+++ b/modules/botfather/command.go
@@ -15,6 +15,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/modules/base/app"
 	"github.com/Mininglamp-OSS/octo-server/modules/group"
 	"github.com/Mininglamp-OSS/octo-server/modules/user"
+	"github.com/Mininglamp-OSS/octo-server/pkg/botutil"
 	"go.uber.org/zap"
 )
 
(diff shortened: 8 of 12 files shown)


pkg/botutil/connect.go
```diff
@@ -0,0 +1,43 @@
...
+// never the localized connect-guide prose.
+const DefaultPluginPackage = "create-openclaw-octo"
+
+// PluginPackageEnv overrides DefaultPluginPackage at runtime (rename / canary),
+// so operators can switch the package without a code change or redeploy of the
+// frontend.
+const PluginPackageEnv = "OCTO_BOT_PLUGIN_PACKAGE"
+
+// PluginPackage returns the configured OpenClaw plugin package name, honoring
+// the DM_BOT_PLUGIN_PACKAGE override and falling back to DefaultPluginPackage.
```
OctoBoooot: **[minor]** Stale env-var name in the docstring: this says "honoring the **DM_BOT_PLUGIN_PACKAGE** override" but the actual constant defined two lines above is `PluginPackageEnv = "OCTO_BOT_PLUGIN_PACKAGE"`, and the code on line 28 reads `os.Getenv(PluginPackageEnv)`. The behavior is correct; only the comment carries the wrong name (looks like a copy from a `DM_*` sibling like `DM_INCOMINGWEBHOOK_*`). Operators grepping the codebase for `DM_BOT_PLUGIN_PACKAGE` would find this docstring and then fail to actually override the package because the env-var name doesn't match — quiet failure mode. One-word fix: `DM_` → `OCTO_`.

OctoBoooot (changes requested):
## Review: feat: serve Bot connect plugin package + API URL from backend (#447)

**Verdict:** Request changes — CI red on `check-sprint / check-sprint` (process gate: no linked issue) precludes APPROVE per skill rule §11. Code itself is clean; one minor (stale docstring), one substantive praise pair.

Size: 9 files / 412 lines (under threshold). Risk tier: normal. Author: an9xyz (100 merged / 0 closed-unmerged).

### Adversarial verification (per skill §7.1)

- **Description-vs-diff** — every quantitative/location-specific claim verified:
  - "covers all four endpoints (GET/POST × admin/space)" → confirmed: `app_bot.go:116` admin GET, `:130` space GET both → `getBotDetail`; `:305` platform POST and `:315` space POST both → `createBot`. Both modified handlers wire `connect: ab.connectInfo()` so all 4 surfaces gain the field.
  - "default is `create-openclaw-octo`" → `pkg/botutil/connect.go:18` const matches; both Go unit tests + the i18n template-regression test pin the literal.
  - "package name is a single backend source of truth shared with BotFather's connect prompts" → confirmed by `command.go` swapping inline `cfg.External.BaseURL`/literal-package code at four sites (`handleQuickstart`, `handleInstall`, `sendConnectPrompt`, `sendCreatedPrompt`) for `botutil.PluginPackage()` + `botutil.DeriveAPIURL(cfg)`.
  - "data only, no secrets" → `connectInfo()` returns a literal `gin.H{plugin_package, api_url}` (two keys, both server-config-derived). `TestGetBotDetail_IncludesConnect` enforces `len(connect)==2` + `NotContains(connect, "token")` + `NotEqual(connect["api_url"], "super-secret-token")` — explicit guard against a future maintainer accidentally widening the map.

- **API/compat shape** — `connect` is an additive top-level key on existing 200 responses; existing consumers see no change in pre-existing fields. Top-level `token` on `createBot` is unchanged (only the connect addition is new). No URL / route / status-code change.

- **Template render safety** — i18n templates run with `missingkey=error` (per PR description). All four render sites in `command.go` inject `PluginPackage` now; `messages_test.go::renderProbe` adds `PluginPackage` to the completeness probe so any future render site that omits it will fail the probe at CI time rather than silently drop a DM.

- **Operator-trust scope** — `OCTO_BOT_PLUGIN_PACKAGE` ends up in `npx -y <pkg>` strings the end user copy-pastes into a shell. Operator-controlled env means the only injection threat would be a malicious operator, which is outside the threat model (they already run the server). Not a finding.

### Minor
- `pkg/botutil/connect.go:26` — see inline.

### Praise
- **`TestGetBotDetail_IncludesConnect` is shaped as a "no future widening" guard.** Asserting `len(info)==2` + `NotContains(connect, "token")` is the right pattern for a response surface that should stay narrow — it fails the moment anyone accidentally adds a third key. That's the same shape as the load-bearing `TestServiceDownloadURL_TransparentAndNoMetric` on PR #442 (assert the absence, not just the presence). Good test design.
- **`TestConnectGuidePackageParameterized` pins both directions.** It checks the injected value renders AND that the old literal `create-openclaw-octo` is absent — a partial-migration regression mode that "did all four templates get updated?" tests usually miss. The kind of test that catches "I updated three of four templates, the fourth still has the literal" the next time someone does another package rename.
- **Single backend source of truth.** Both BotFather i18n templates AND the App Bot connect response resolve `plugin_package` through `botutil.PluginPackage()` — the exact i18n/code drift class issue #446 was about, eliminated structurally rather than by convention. The `replyL(MsgInstall, map{PluginPackage})` change on `command.go:436` is small but important: previously that site passed `nil` data, which would have been a silent miss if anyone later inserted `{{.PluginPackage}}` into the install template without remembering to thread the value.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#447 Review Report
**Head SHA:** `ac1fdf03db3b7c2540a8eb0fa9ee58e144d6a377`
**PR:** feat: serve Bot connect plugin package + API URL from backend (#446)
**Routing:** security_sensitive (automated review)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `connect{plugin_package, api_url}` on createBot response | ✅ | `app_bot.go:422-428` — `createBot` adds `"connect": ab.connectInfo()` |
| `connect{plugin_package, api_url}` on getBotDetail response | ✅ | `app_bot.go:535` — `getBotDetail` adds `"connect": ab.connectInfo()` |
| `connectInfo()` single source of truth | ✅ | `app_bot.go:990-1001` — delegates to `botutil.PluginPackage()` + `botutil.DeriveAPIURL()` |
| No secret leakage into `connect` | ✅ | `connectInfo()` returns exactly 2 keys: `plugin_package`, `api_url`. Token stays in top-level `token` field (masked in detail). Test asserts `assert.Len(t, info, 2)` and `assert.NotContains(t, resp.Connect, "token")` |
| Env override `OCTO_BOT_PLUGIN_PACKAGE` reaches client | ✅ | `connect.go:28-32` reads env; e2e test `TestGetBotDetail_ConnectHonorsPackageOverride` proves end-to-end flow |
| BotFather 4 render sites inject `PluginPackage` | ✅ | `command.go:427` (handleQuickstart), `:434` (handleInstall), `:930` (sendConnectPrompt), `:939` (sendCreatedPrompt) — all pass `"PluginPackage": botutil.PluginPackage()` |
| Template de-hardcode: `{{.PluginPackage}}` replaces literal | ✅ | `en-US/command.tmpl` + `zh-CN/command.tmpl` — all 8 occurrences (4 per locale) replaced |
| Completeness probe covers new key | ✅ | `messages_test.go:26` — `renderProbe()` includes `"PluginPackage": "create-openclaw-octo"`. `TestBotMessagesCompleteness` renders every key × every language with `missingkey=error` |
| Auth gates unchanged | ✅ | `CheckLoginRole`, `checkSpaceAdmin`, `botInRouteScope` IDOR guard — none touched by diff |
| Tests: botutil unit | ✅ | `connect_test.go` — default, env override, whitespace-only override, DeriveAPIURL both paths |
| Tests: app_bot HTTP e2e | ✅ | `app_bot_connect_test.go` — TestConnectInfo_DefaultAndOverride, TestGetBotDetail_IncludesConnect, TestGetBotDetail_ConnectHonorsPackageOverride |
| Tests: botfather template | ✅ | `messages_test.go` — TestConnectGuidePackageParameterized (both locales, 4 template keys), guard asserts old literal absent |

## 2. Findings

### P2-1 — Stale env-var name in doc comment (cosmetic)

**Severity:** P2 (nit)
**Diff-scope:** New (introduced by this PR in new file `pkg/botutil/connect.go`)

`connect.go:26` says:
```go
// the DM_BOT_PLUGIN_PACKAGE override and falling back to DefaultPluginPackage.
```
But the actual env var defined 3 lines above is `OCTO_BOT_PLUGIN_PACKAGE` (line 23). Likely a stale comment from an earlier draft.

**Suggested fix:** Change `DM_BOT_PLUGIN_PACKAGE` → `OCTO_BOT_PLUGIN_PACKAGE` on line 26.

### P2-2 — `DeriveAPIURL` fallback is dead code in production (benign)

**Severity:** P2 (maintainability note)
**Diff-scope:** New (introduced by this PR)

`connect.go:37-41` implements a `BaseURL=="" → http://<IP>:8090` fallback. However, `octo-lib/config/config.go:558-559` already sets `c.External.BaseURL = fmt.Sprintf("http://%s:8090", c.External.IP)` during config loading when BaseURL is empty. At runtime (after `config.FromFile`), `cfg.External.BaseURL` is never empty, so the `DeriveAPIURL` fallback branch is unreachable in production.

This is benign — the result would be identical even if reachable — and the fallback is exercised by unit tests (which use `config.New()` without config loading). Not a bug, but the unreachable branch could confuse future readers. Consider a one-line comment noting that config loading already guarantees BaseURL.

### P2-3 — `skill.go` still hardcodes `create-openclaw-octo` (14 occurrences)

**Severity:** P2 (consistency gap, tracked separately)
**Diff-scope:** Pre-existing (explicitly listed as out-of-scope in the task brief)

`modules/botfather/skill.go` (functions `generateSkillMD`, `generateCLIGuideMD`, `generateSetupNewbotMD`, `generateSetupQuickstartMD`) still contain 14 hardcoded `create-openclaw-octo` references. These are served via `api.go:115-146` to OpenClaw agents fetching skill/setup documents. If `OCTO_BOT_PLUGIN_PACKAGE` is set to a canary, the DM connect guides (now parameterized) will show the canary, but the skill documents will still show the default.

The task brief explicitly lists this as out-of-scope and "left for a separate decision" — acknowledged. Tracking this as a P2 note for follow-up, not a blocker.

## 3. Suggestions

1. **Fix the stale env var name** in `connect.go:26` doc comment (`DM_` → `OCTO_`). One-line edit.
2. **Consider a brief comment** on `DeriveAPIURL` noting that config loading pre-populates BaseURL, so the IP-fallback branch is a test-only safety net. Prevents future readers from thinking they found a missing-override bug.
3. **Track skill.go de-hardcode** as a follow-up task (per the brief's out-of-scope note). The inconsistency between DM templates (parameterized) and skill docs (hardcoded) is low-severity but worth closing.

## 4. Extra Observations

- **Additive wire contract**: The `connect` field is purely additive on existing endpoints. No existing consumer is affected. Good defensive choice to exclude the token.
- **Consolidation value**: The inline `apiURL` derivation was duplicated across `sendConnectPrompt`, `sendCreatedPrompt`, and `handleQuickstart` (3 copies of the same 4-line block). Consolidating onto `botutil.DeriveAPIURL()` removes that duplication cleanly. `bot_api/register.go` still has its own copy (also out-of-scope per brief) — worth consolidating later.
- **Test quality**: The test suite is well-designed — unit tests for the leaf functions, HTTP e2e with sqlmock for the endpoint integration, and a de-hardcode guard that would catch future template regressions. The completeness probe (`TestBotMessagesCompleteness`) with `missingkey=error` is a strong invariant.

## 5. Data Flow Trace

| Consumer | Data | Upstream Source | Verified Flow |
|----------|------|-----------------|---------------|
| `createBot` response `connect.plugin_package` | package name | `botutil.PluginPackage()` → env `OCTO_BOT_PLUGIN_PACKAGE` or default `create-openclaw-octo` | ✅ `connectInfo()` at `app_bot.go:996` → `botutil.PluginPackage()` at `connect.go:28` → `os.Getenv` |
| `createBot` response `connect.api_url` | public API URL | `botutil.DeriveAPIURL(cfg)` → `cfg.External.BaseURL` (config-loaded, never empty at runtime) | ✅ `connectInfo()` at `app_bot.go:997` → `botutil.DeriveAPIURL()` at `connect.go:36` → `cfg.External.BaseURL` |
| `getBotDetail` response `connect` | same as above | same | ✅ Same `connectInfo()` call at `app_bot.go:535` |
| BotFather `sendConnectPrompt` template `{{.PluginPackage}}` | package name | `botutil.PluginPackage()` at `command.go:930` | ✅ Data flows to `command.tmpl` `connect_prompt` define |
| BotFather `sendCreatedPrompt` template `{{.PluginPackage}}` | package name | `botutil.PluginPackage()` at `command.go:939` | ✅ Data flows to `command.tmpl` `created_prompt` define |
| BotFather `handleQuickstart` template `{{.PluginPackage}}` | package name | `botutil.PluginPackage()` at `command.go:427` | ✅ Data flows to `command.tmpl` `quickstart` define |
| BotFather `handleInstall` template `{{.PluginPackage}}` | package name | `botutil.PluginPackage()` at `command.go:434` | ✅ Data flows to `command.tmpl` `install` define |
| BotFather `sendConnectPrompt` template `{{.APIURL}}` | public API URL | `botutil.DeriveAPIURL(h.ctx.GetConfig())` at `command.go:929` | ✅ Same DeriveAPIURL as app_bot, consolidated |

- **C1 Dual-path parity:** Clear. `connect` is added to both `createBot` (covers POST admin + POST space) and `getBotDetail` (covers GET admin + GET space). Both admin and space routes share the same internal functions. No add/remove/create/delete asymmetry.
- **C2 Control-flow ordering / nested reuse:** Clear. `PluginPackage()` and `DeriveAPIURL()` are leaf functions with no side effects, called once per response. No nested security controls. The `missingkey=error` template mode ensures all render sites supply required keys — verified by completeness probe test.
- **C3 Authorization boundary:** Clear. No new endpoints. `connect` is added to existing responses behind existing auth gates (`CheckLoginRole`, `checkSpaceAdmin`). Data in `connect` (package name, public API URL) is non-sensitive.
- **C4 Authorization lifecycle / container-member cascade:** N/A — no auth changes.
- **C5 Build ≠ runtime:** N/A — no build artifacts, extensions, CLI, or packaging changes.
- **C6 Governance doc consistency:** N/A — the only doc added is `.octospec/tasks/bot-connect-plugin-package/brief.md`, an internal task spec. No governance/policy/security doc changes.

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. The PR is clean, well-scoped, and well-tested. Three P2 nits: a stale env-var name in a doc comment (trivial fix), a dead-code fallback path (benign, cosmetic), and `skill.go` hardcoded references (explicitly out-of-scope, tracked for follow-up). The wire contract is additive, auth gates are untouched, no secrets leak, and test coverage proves the env-override reaches the client end-to-end.

Jerry-Xin (dismissed):
Summary: The PR is relevant to `octo-server` and the implementation is scoped, additive, and aligned with the intended backend-owned App Bot connect contract.

💬 Non-blocking

🟡 Warning — [pkg/botutil/connect.go](/Users/mlamp/.openclaw/tmp/pr-review-w_a577oa/repo/pkg/botutil/connect.go:25): the comment says `DM_BOT_PLUGIN_PACKAGE`, but the actual env var is `OCTO_BOT_PLUGIN_PACKAGE`. This is documentation drift and could confuse operators.

🔵 Suggestion — [pkg/botutil/connect.go](/Users/mlamp/.openclaw/tmp/pr-review-w_a577oa/repo/pkg/botutil/connect.go:27): consider validating the configured package string against a conservative npm package-name/spec pattern before rendering it into shell commands. The env is operator-controlled, so this is not a blocker, but defensive validation would reduce accidental malformed or unsafe connect prompts.

✅ Highlights

The new `connect` object is additive and does not alter existing auth, space-scope, or token-reveal gates in [modules/app_bot/app_bot.go](/Users/mlamp/.openclaw/tmp/pr-review-w_a577oa/repo/modules/app_bot/app_bot.go:424) and [modules/app_bot/app_bot.go](/Users/mlamp/.openclaw/tmp/pr-review-w_a577oa/repo/modules/app_bot/app_bot.go:521).

BotFather template render sites now supply `PluginPackage`, avoiding `missingkey=error` render failures for the changed templates in [modules/botfather/command.go](/Users/mlamp/.openclaw/tmp/pr-review-w_a577oa/repo/modules/botfather/command.go:427).

Targeted tests passed: `go test ./pkg/botutil`, the new App Bot connect tests, and the BotFather template tests. Full `modules/app_bot` / `modules/botfather` package runs were blocked by local test DB migration state (`unknown migration in database`), not by this PR’s targeted changes.

pkg/botutil/connect.go
an9xyz: Good catch — fixed in f113d23 (`DM_` → `OCTO_` in the docstring). The constant and `os.Getenv` already read `OCTO_BOT_PLUGIN_PACKAGE`; only the comment was stale from the pre-rename version.

---

yujiawei (dismissed):
# Code Review — PR #447 (octo-server)

Independent review at head SHA `ac1fdf03db3b7c2540a8eb0fa9ee58e144d6a377` against `main` (merge-base `8d2f4e34`). This PR is security-sensitive (auth/token/middleware-adjacent), so it got extra scrutiny on the token boundary, tenant isolation, and i18n render safety.

## 1. Spec / scope compliance — PASS

The change does exactly what the linked issue (#446) asks and nothing more:

- **`connect{plugin_package, api_url}`** is added to the two response builders that back all four App Bot read/create endpoints: `createBot` serves `POST /v1/admin/app_bot` + `POST /v1/space/:space_id/app_bot` (`modules/app_bot/app_bot.go:425`), and `getBotDetail` serves `GET /v1/admin/app_bot/:id` + `GET /v1/space/:space_id/app_bot/:id` (`app_bot.go:535`). The list endpoints are intentionally excluded — consistent with the stated scope.
- **Single source of truth** in `pkg/botutil/connect.go`: `PluginPackage()` (default `create-openclaw-octo`, env override) and `DeriveAPIURL(cfg)` (`External.BaseURL` → `http://<ip>:8090` fallback). Both App Bot and BotFather resolve through it.
- **i18n de-hardcoding** is complete: all four `npx -y create-openclaw-octo …` literals in `templates/{en-US,zh-CN}/command.tmpl` are replaced by `{{.PluginPackage}}`, and the package is injected at all four `command.go` render sites (`sendConnectPrompt`, `sendCreatedPrompt`, `handleQuickstart`, `handleInstall`) plus the completeness probe.
- No over-build (no new auth gate, no error-envelope change, no widened query) and no scope creep into the out-of-scope items the issue lists (frontend `connectGuide.ts`, `skill.go`, `register.go` consolidation).

## 2. Code quality — Approved

**Verified correct:**

- **No secret leakage.** `connect` carries exactly `plugin_package` + `api_url`; the masked top-level `token` is untouched, and the e2e test asserts `len(connect)==2`, absence of a `token` key, and that the raw token never appears in `api_url` (`app_bot_connect_test.go`).
- **No IDOR / isolation regression.** `connectInfo()` is config-derived and tenant-independent — identical regardless of scope/space — and the existing `checkSpaceAdmin` / `CheckLoginRole` gates and the `botInRouteScope` cross-tenant guard in `getBotDetail` are unchanged. The new field widens no query.
- **i18n render safety.** The msgtmpl catalog runs with `missingkey=error` (`modules/base/common/msgtmpl/msgtmpl.go:84`). Every render site that hits a `{{.PluginPackage}}` template now supplies the key, and `TestBotMessagesCompleteness` renders every declared key in every supported language through `renderProbe()` (which now includes `PluginPackage`), making an omission a build-time failure rather than a silently-dropped DM.
- **Behavior parity.** `DeriveAPIURL` reproduces the exact `External.BaseURL`/`IP:8090` logic already used in `bot_api/register.go` and the old inline `command.go` code, so the connect guide and register response stay consistent.
- **Build + tests.** `go build` and `go vet` are clean for the three changed packages; `pkg/botutil`, the BotFather template/key-message tests, and the `modules/app_bot` connect tests all pass locally. (One unrelated test, `TestRobotApply_AutoApprove`, fails on a pre-existing local DB-migration mismatch — not introduced by this PR.)

**Non-blocking nits (P2 — optional, do not block merge):**

- `pkg/botutil/connect.go:26` — the doc comment says "honoring the `DM_BOT_PLUGIN_PACKAGE` override", but the constant is `OCTO_BOT_PLUGIN_PACKAGE`. Stale comment; the code is correct. Worth fixing to avoid operator confusion in deployment manifests.
- `pkg/botutil/connect.go` — `PluginPackage()` calls `os.Getenv` on each request/render. This is cheap (cached environ behind a shared read-lock, and the pattern already exists elsewhere in the codebase) and OS env mutations don't propagate into a running process anyway, so there's no functional concern — but reading it once at startup would make the test override cleaner and remove any future `t.Parallel()` + `t.Setenv` foot-gun.

## 3. For a human to verify (security-sensitive)

- Confirm `External.BaseURL` is actually populated in the relevant deployment environments so `api_url` resolves to the public Bot API entry rather than the `IP:8090` fallback (the fallback is correct behavior, just worth confirming production has BaseURL set).
- The frontend follow-up (`octo-admin connectGuide.ts` consuming `bot.connect`) and the upstream `config.Config` binding for `External.*` are outside this repo/diff and were not assessed here.

## Verdict

No P0/P1 issues. The change is correctly scoped, the token/isolation boundaries are preserved, and the i18n render path is guarded at build time. Approving; the two P2 nits are optional cleanups.

an9xyz:
Thanks for the thorough pass. On the two P2 nits:

1. **Stale `DM_BOT_PLUGIN_PACKAGE` docstring** (`connect.go:26`) — already fixed in `f113d23`, pushed just after the `ac1fdf0` you reviewed (`DM_` → `OCTO_`).
2. **Per-call `os.Getenv` in `PluginPackage()`** — leaving as-is on purpose: it's the established pattern in this repo (`modules/category/config.go`, `modules/file/const.go`, …), and the per-call read is what lets `TestGetBotDetail_ConnectHonorsPackageOverride` exercise the `OCTO_BOT_PLUGIN_PACKAGE` override end-to-end via `t.Setenv` — resolving once at init would capture the value before the test sets it. No `t.Parallel()` on these tests, so the `t.Setenv` foot-gun isn't live. Agreed there's no functional difference either way.

On the human-verify note: 👍 confirming `External.BaseURL` is populated in the relevant environments (so `api_url` is the public entry, not the `IP:8090` fallback) is the right thing to check before rollout.

---

Jerry-Xin (approved):
The PR is relevant to octo-server and the implementation is scoped, additive, and aligned with the stated backend ownership of Bot connect metadata.

**💬 Non-blocking**
- 🔵 Suggestion: [modules/app_bot/app_bot_connect_test.go](/Users/mlamp/.openclaw/tmp/pr-review-kslsy1m8/repo/modules/app_bot/app_bot_connect_test.go:60) says the endpoint proof covers read/create endpoints, but the HTTP-level tests only exercise `GET /v1/admin/app_bot/:id`. Since [modules/app_bot/app_bot.go](/Users/mlamp/.openclaw/tmp/pr-review-kslsy1m8/repo/modules/app_bot/app_bot.go:424) also changes create responses, a focused create success test would better pin that wire contract.
- 🔵 Suggestion: [pkg/botutil/connect.go](/Users/mlamp/.openclaw/tmp/pr-review-kslsy1m8/repo/pkg/botutil/connect.go:34) documents `DeriveAPIURL` as the value for connect/register responses, but `modules/bot_api/register.go` still derives the register `api_url` inline. Behavior currently matches, but calling the helper there too would reduce future drift.

**✅ Highlights**
- Existing App Bot authorization and scope checks remain before the new response data; the `connect` object is only added after successful create/detail flows and does not widen access.
- BotFather template changes are matched by render-site data injection for quickstart, install, connect, and created prompts.
- `connect` contains only `plugin_package` and `api_url`; no token is added to that object.

**Verification**
- Passed: `go test ./pkg/botutil`
- Passed: `go test ./modules/app_bot -run 'TestConnectInfo|TestGetBotDetail'`
- Passed: `go test ./modules/botfather -run 'TestKeyMessageInterpolation|TestConnectGuidePackageParameterized|TestAllBotMessageTemplates|TestBotMessageTemplateSetMatchesKeys'`
- Caveat: full `go test ./modules/botfather` was blocked by the local test DB migration state: `unknown migration in database`.

pkg/botutil/connect.go
```diff
@@ -0,0 +1,43 @@
...
+		return v
+	}
+	return DefaultPluginPackage
+}
+
+// DeriveAPIURL resolves the public Bot API entry for this deployment: the
+// configured External.BaseURL, falling back to the direct-access host:port when
+// BaseURL is unset. This is the value Agents pass as `--api-url` and the URL
+// returned in Bot connect/register responses — NOT the admin-dashboard origin.
+func DeriveAPIURL(cfg *config.Config) string {
```
OctoBoooot, line 38: **[major]** `DeriveAPIURL` was introduced to be the single source of truth for the public Bot API entry, and the PR description states the derivation is "consolidated onto `botutil.DeriveAPIURL`". A grep of the codebase shows **6 inline copies** of the exact `cfg.External.BaseURL || fmt.Sprintf("http://%s:8090", cfg.External.IP)` pattern still in place — **including 3 in the same `modules/botfather/` directory** the PR also touches:

```
modules/botfather/api.go:110-112
modules/botfather/api.go:130-132
modules/botfather/api.go:142-144
modules/bot_api/register.go:106-108
modules/bot_api/register.go:166-168
modules/bot_api/commands.go:88-90
```

This defeats the PR's stated goal: a future change to the api_url derivation (e.g., honor a new TLS-termination env, change the dev port from 8090, add a sanity check on `External.IP`) still has to be made in 7 places (1 helper + 6 inlines), with the same drift risk the helper was supposed to remove. Same-module `api.go` callsites are particularly inconsistent — they sit right next to the `command.go` sites that now use the helper.

Not a runtime bug — current behavior is identical. But it's a description-vs-diff inconsistency on the PR's load-bearing claim, and the same-module case isn't "out of scope follow-up." Smallest closing change: in each of the 6 callsites above, replace the four-line inline block with `apiURL := botutil.DeriveAPIURL(cfg)` and add the `botutil` import. The existing `TestDeriveAPIURL` covers the helper's behavior; no new tests needed.

If scope truly is "command.go + app_bot only by design", say so explicitly in the PR description and rename the helper or update the description so future readers don't expect it to be repo-wide.

OctoBoooot (changes requested):
## Review: feat: serve Bot connect plugin package + API URL from backend (#447) — delta @ f113d23b

**Verdict:** Request changes (carry-forward of CI red + new major from delta scope).

Delta scope: 1 commit (`f113d23b`), 1-line docstring fix. My prior minor (env-var name in docstring) is closed. But the delta also exposed a **completeness gap** that Jerry-Xin's 🟡 #2 flagged and I should have caught in my prior round — promoting it to major here, with a retraction.

### Prior items
- ✅ `pkg/botutil/connect.go:26` docstring `DM_BOT_PLUGIN_PACKAGE` → `OCTO_BOT_PLUGIN_PACKAGE`. Resolved.

### Major (new at this delta, also raised by Jerry-Xin)
- See inline at `pkg/botutil/connect.go:38` — `DeriveAPIURL` is only adopted at 5 callsites (4 in `botfather/command.go`, 1 in `app_bot.go::connectInfo`); the literal `cfg.External.BaseURL || fmt.Sprintf("http://%s:8090", cfg.External.IP)` pattern still appears in 6 other places, **including 3 in the same `modules/botfather/` directory** (`api.go:110,130,142`). PR description states "api_url derivation consolidated onto `botutil.DeriveAPIURL`" — that's only partly true.

### CI gate
- `check-sprint / check-sprint` still red at this SHA (linked-issue gate). Process gate, not a code defect — but per skill rule §11 verdict floor remains REQUEST_CHANGES regardless of code state.

### Retraction
My prior review's praise said:
> Single backend source of truth (`botutil.PluginPackage`) used by BOTH BotFather i18n templates AND the App Bot connect response — eliminates the i18n/code drift class that the issue identified.

That sentence was accurate for `PluginPackage` (which is the single source of truth — verified via grep, only one usage path) but I misattributed the same property to `DeriveAPIURL` by association. The two helpers got the same praise wrapper when only one earned it. This is the third hit on the [praise-has-weight] memory in recent reviews — first PR #423 (mdLinkText covered one context), then PR #442 (DownloadURL wrap praised for transparency around a no-I/O method), now this one (DeriveAPIURL praised for consolidation that didn't actually consolidate). The common failure mode is the same: I check the happy direction the PR claims and don't grep for inline duplicates that *should* have been consolidated but weren't. Adding to the praise-verification checklist: for any consolidate / dedup / source-of-truth PR, grep the codebase for the literal pattern being replaced and confirm zero residual matches BEFORE praising.

Jerry-Xin (changes requested):
## ⚠️ Downgrading my prior APPROVE (4558517340) → REQUEST_CHANGES

New grounded evidence after peer review (Octo-PR / OctoBoooot) flagged incomplete `DeriveAPIURL` consolidation. I independently byte-verified on this head (`f113d23b63ba`) and the facts hold — so I'm correcting my own vote rather than protecting it.

### 🔴 Blocking — `api_url` consolidation is half-done, and the description/COMPREHENSION claims contradict the diff

`pkg/botutil/connect.go:38` defines:
```go
func DeriveAPIURL(cfg *config.Config) string {
    if u := strings.TrimSpace(cfg.External.BaseURL); u != "" { return u }
    return fmt.Sprintf("http://%s:8090", cfg.External.IP)
}
```

Verified callsites at this head:
- **Only 4 production callers route through the helper**: `modules/app_bot/app_bot.go:999`, `modules/botfather/command.go:429,930,940`.
- **6 byte-identical inline copies remain un-consolidated**:
  - `modules/bot_api/register.go:106-108`, `166-168`
  - `modules/bot_api/commands.go:88-90`
  - `modules/botfather/api.go:110-112`, `130-132`, `142-144`

The PR's Changes section scopes the claim to *"`modules/botfather`: api_url derivation consolidated onto `botutil.DeriveAPIURL`"* — but within `modules/botfather` itself, **`api.go` still has 3 inline copies** (only `command.go` was migrated). And `modules/bot_api` (3 more copies) isn't mentioned at all. The COMPREHENSION note "api_url is now centralized in one helper" / "single source of truth" is **factually contradicted by 6 surviving inline copies**.

**Damning detail:** `register.go` and `api.go` already call `botutil.DeriveWSURL(cfg)` on the line *right after* inline-building `apiURL`. The sibling WS-URL helper was migrated in these exact functions while the API-URL was left inline — so the consolidation is demonstrably in-scope and half-finished, not "out of scope / different module."

### Severity rationale (honest)
- **No behavioral bug ships today**: all 6 copies are byte-identical to the helper, so connect-guide and register responses are consistent *right now*. On pure runtime correctness this is 🟡 maintainability.
- **What makes it blocking**: the PR asserts an invariant ("single source of truth" / "centralized in one helper") that the diff does **not** establish. A future change to the derivation (https default, port, IPv6 bracketing) must be made in 7 places or they silently diverge — defeating the stated goal. Misleading description + incomplete refactor that contradicts its own stated invariant = request-changes.

### Ask (trivial, low-risk)
Either:
1. Migrate the 6 inline copies to `botutil.DeriveAPIURL(cfg)` (3 of them sit next to an already-migrated `DeriveWSURL` call — mechanical), **or**
2. Correct the PR description / COMPREHENSION to honestly scope the claim (drop "single source of truth" / "centralized in one helper").

### ✅ Unchanged — still good
- No token leak: `connectInfo()` remains a literal 2-key object (`plugin_package`, `api_url`); tests assert `len==2` + no `"token"`.
- Auth/scope gates (`checkSpaceAdmin`/`CheckLoginRole`/`botInRouteScope`) untouched.
- i18n `missingkey=error` intact; docstring `DM_BOT_PLUGIN_PACKAGE → OCTO_BOT_PLUGIN_PACKAGE` fixed.

ℹ️ CI: core checks green; `check-sprint` (linked-issue gate) still needs an issue.

OctoBoooot (approved):
## Review: feat: serve Bot connect plugin package + API URL from backend (#447) — delta @ 39bdc30f

**Verdict:** Approve.

Delta scope: prior REQUEST_CHANGES @ `f113d23b` (1 major: DeriveAPIURL consolidation incomplete — 6 inline copies remained). Two commits at this head:
- `6a54f09e` test(app_bot): pin create-side connect wire contract with an e2e
- `39bdc30f` refactor: consolidate remaining api_url derivation onto botutil.DeriveAPIURL

Byte-verified, both prior items closed:

### Prior items
- ✅ **Major (DeriveAPIURL consolidation)** — `grep -rn "DeriveAPIURL\b"` shows **11 callsites** now (was 5), exactly matching the 6 added: `bot_api/register.go:105,162` + `bot_api/commands.go:88` + `botfather/api.go:109,125,133`. `grep -rn "8090"` returns one hit, inside the helper itself (`pkg/botutil/connect.go:42`). All 6 inline copies the prior review enumerated are gone. The "single source of truth" claim is now actually true. The remaining `External.BaseURL` usages I see in `modules/group/`, `modules/user/`, `modules/space/` are for unrelated URL purposes (email-invite accept URLs, QR codes, approval URLs) with different fallback semantics — correctly excluded from `DeriveAPIURL`'s scope.
- ✅ **Jerry-Xin's prior 🟡 #1 (create-side test)** — new `TestCreatePlatformBot_IncludesConnect_E2E` (modules/app_bot/app_bot_connect_test.go:101+) pins POST `/v1/admin/app_bot`'s connect contract against the real server (`testutil.NewTestServer`, not sqlmock), asserts the same `len==2 + NotContains "token"` invariant the GET-path test enforces. Symmetric coverage now in place.

### Praise
- The split commit structure (test + refactor) made the review trivial: I could verify the test was added without touching production wiring, then verify the refactor was pure substitution. Tiny review hygiene win.
- The new e2e test reuses the same "data only, no secret leakage" invariant assertions as the GET-path test (`Len(connect)==2`, `NotContains "token"`, `NotEqual "super-secret-token"`). Both endpoints now have identical contract pins — a future maintainer who adds a third key to `connectInfo()` fails BOTH tests, not just one. The right pattern for paired endpoints sharing a helper.

### Verdict-side gate (informational, not a finding)
`reviewDecision` may still show `CHANGES_REQUESTED` until Jerry-Xin (and Steve, who also downgraded to CR at the prior head) re-review at `39bdc30f` — their stale CRs from the prior SHA don't auto-clear. Same convergence pattern as PR #442 / #445.

pkg/botutil/connect.go
an9xyz: Done in `39bdc30` — all 6 inline copies now route through `botutil.DeriveAPIURL` (`botfather/api.go` ×3, `bot_api/register.go` ×2, `bot_api/commands.go` ×1). Grep for the inline pattern returns zero residual, and the `register.go`/`api.go` sites now match their adjacent `DeriveWSURL` call. `TestDeriveAPIURL` covers behavior; affected suites green under `-race`; description updated.

---

an9xyz:
Both blocking reviews addressed:

**`api_url` consolidation** (OctoBoooot `[major]` / Jerry-Xin 🔴) — `39bdc30` migrates all 6 inline `External.BaseURL` / `http://<ip>:8090` copies to `botutil.DeriveAPIURL`: `botfather/api.go` ×3 (now consistent with the adjacent `DeriveWSURL` calls), `bot_api/register.go` ×2, `bot_api/commands.go` ×1. Grep for the inline pattern returns **zero** residual; behavior-identical; `TestDeriveAPIURL` pins it. Description + COMPREHENSION updated to match the diff.

**create-side coverage** (Jerry-Xin 🔵) — `6a54f09` adds a `NewTestServer` e2e for `POST /v1/admin/app_bot` asserting the create response carries `connect{plugin_package, api_url}`, data only, with the freshly-minted token never leaking into `connect`.

`app_bot` / `botfather` / `bot_api` / `botutil` suites pass under the CI recipe (master key + per-package DB reset + `-race`); golangci-lint clean. The only remaining red is `check-sprint` (linked-issue gate) — `Closes #446` is now in the description, so it should clear once the Sprint field is set on the board.

---

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#447 Review Report
**PR:** Mininglamp-OSS/octo-server#447
**Head SHA:** 39bdc30f82352bfd706608cfb51b25a32daec60d
**Base:** main (8d2f4e34)
**Scope:** security_sensitive (automated review)

---

| Check | Status | Evidence |
|-------|--------|----------|
| `api_url` derivation fully consolidated | ✅ | Grep for `http://%s:8090` returns only `pkg/botutil/connect.go:42` — single source |
| `connect` response field added to 4 endpoints | ✅ | `createBot` (covers POST admin + POST space) at `app_bot.go:428`; `getBotDetail` (covers GET admin + GET space) at `app_bot.go:538` |
| No secret leak into `connect` | ✅ | `connectInfo()` at `app_bot.go:997-1001` returns exactly 2 keys: `plugin_package` + `api_url`. No token, no DB-derived data |
| Authorization gates unchanged | ✅ | `createBot`: `checkSpaceAdmin` / `CheckLoginRole` (lines 356-366); `getBotDetail`: `checkSpaceAdmin` / `CheckLoginRole` + `botInRouteScope` IDOR guard (lines 505-530). PR adds `connect` to the response body only |
| i18n template completeness | ✅ | All 4 render sites (`handleQuickstart`, `handleInstall`, `sendConnectPrompt`, `sendCreatedPrompt`) supply `PluginPackage`. `renderProbe()` updated. `missingkey=error` enforced by `msgtmpl` at `msgtmpl.go:84` |
| `DeriveAPIURL` behavior-identical | ✅ | Old inline: `cfg.External.BaseURL` → fallback `http://<IP>:8090`. New `DeriveAPIURL`: identical logic at `connect.go:39-42` |
| Unit + E2E tests | ✅ | `TestPluginPackage_DefaultAndOverride`, `TestDeriveAPIURL`, `TestConnectInfo_DefaultAndOverride`, `TestGetBotDetail_IncludesConnect`, `TestGetBotDetail_ConnectHonorsPackageOverride`, `TestCreatePlatformBot_IncludesConnect_E2E`, `TestConnectGuidePackageParameterized` — all cover the contract |
| Import cleanup | ✅ | `fmt` removed from `register.go` and `api.go` where no longer used; `botutil` import added where needed |

## 2. Findings

### P2-1: `skill.go` still hardcodes `create-openclaw-octo` (17 occurrences)

**File:** `modules/botfather/skill.go` (not touched by PR)

The PR parameterized the DM message templates (`command.tmpl` × 2 languages) but left the static Markdown documentation generators untouched. These four functions generate HTTP-served Markdown docs:
- `generateSkillMD()` → `/v1/bot/skill.md`
- `generateCLIGuideMD()` → `/v1/bot/cli-guide.md`
- `generateSetupNewbotMD()` → `/v1/bot/setup-newbot.md`
- `generateSetupQuickstartMD()` → `/v1/bot/setup-quickstart.md`

**diff-scope check (R2):**
1. **New/amplified:** Amplified. Before this PR, both DM templates and skill.go docs hardcoded the same literal — consistent but frozen. After this PR, DM templates are parameterized while skill.go docs remain hardcoded → a new divergence between the two sources.
2. **PR changes the contract?** The DM path now honors `OCTO_BOT_PLUGIN_PACKAGE`; the Markdown docs do not.
3. **Pre-existing?** The hardcoded literal was pre-existing, but the divergence is newly introduced by this PR.

**Severity rationale (R1):** The DM prompts (what users copy-paste) correctly show the configured package. The Markdown docs are supplementary reference material. No runtime failure, no erroneous data in the actionable path. Not P1.

**Suggestion:** Parameterize `skill.go` generators to accept `pluginPackage` alongside `apiURL`, or document in the task brief that these endpoints are out of scope for the package-name migration.

### P2-nit: `const_test.go:148` asserts hardcoded literal

`TestSkillMDContent_Deprecated` asserts `assert.Contains(t, md, "create-openclaw-octo install")`. If skill.go is later parameterized, this assertion must be updated. No production impact — test-coupling note only.

## 3. Suggestions

1. **Parameterize skill.go** (follow-up PR): Add `pluginPackage string` parameter to the four `generate*MD` functions and pass `botutil.PluginPackage()` from the callers in `api.go`. Update `const_test.go` accordingly. This completes the "single backend source of truth" goal.

## 4. Extra Findings

None beyond P2-1.

## 5. Data Flow Trace

| Consumed Data | Upstream Source | Flows to Consumer? |
|---|---|---|
| `connect.plugin_package` | `botutil.PluginPackage()` → `os.Getenv("OCTO_BOT_PLUGIN_PACKAGE")` → default `"create-openclaw-octo"` | ✅ Via `connectInfo()` → `gin.H` → HTTP JSON response |
| `connect.api_url` | `botutil.DeriveAPIURL(cfg)` → `cfg.External.BaseURL` (or fallback `http://<IP>:8090`) | ✅ Same path |
| `{{.PluginPackage}}` in templates | `botutil.PluginPackage()` in `command.go` render data maps | ✅ All 4 render sites supply it; `missingkey=error` catches misses |
| `{{.APIURL}}` in templates | `botutil.DeriveAPIURL(h.ctx.GetConfig())` | ✅ All 3 APIURL render sites use consolidated getter |
| `connect.token` (assertion: absent) | Not included in `connectInfo()` return | ✅ Token stays at top-level response only, masked in GET |

- **C1 — Dual-path parity:** N/A. No add/remove or create/delete symmetry involved. PR adds a response field.
- **C2 — Control-flow ordering / nested reuse:** N/A. `DeriveAPIURL` and `PluginPackage` are leaf getters with no nesting or ordering dependencies. No security controls involved.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. `connect` carries public data only (package name, API URL). No new capability exposed.
- **C5 — build ≠ runtime:** N/A. No build artifacts, CLI, extensions, or relative paths. The `connect` field is runtime-computed from config/env.
- **C6 — Governance/policy doc self-consistency:** N/A. No governance or policy documents changed.

---

**[Octo-Q] verdict: APPROVE**

The PR cleanly consolidates the `api_url` derivation into a single `botutil.DeriveAPIURL` source, adds a well-designed `connect` response field with no secret leakage, parameterizes the DM i18n templates, and ships comprehensive tests (unit + E2E + parameterized canary). Authorization gates are unchanged. The only finding is P2: `skill.go`'s static Markdown docs still hardcode the old package literal, creating a new divergence with the now-parameterized DM templates. This is a documentation consistency issue, not a runtime defect — the actionable path (DM prompts users copy-paste) is correct. Recommend a follow-up to parameterize `skill.go` for completeness.

Jerry-Xin (approved):
Re-review on new head `39bdc30f` — supersedes my prior REQUEST_CHANGES on `f113d23b`. The blocker is resolved.

Delta since f113d23b (5 files): commands.go +2/-5, register.go +2/-9, botfather/api.go +3/-15, app_bot_connect_test.go +48/-4, brief.md. This is exactly the api_url consolidation cleanup I asked for.

🔴→✅ Blocker resolved — api_url single source of truth now actually holds
All 6 inline `External.BaseURL` / `http://%s:8090` copies are migrated to `botutil.DeriveAPIURL(cfg)`:
- bot_api/register.go:105, :162 ✓
- bot_api/commands.go:88 ✓
- botfather/api.go:109, :125, :133 ✓
Full-repo grep at the head ref: the `http://%s:8090` literal now exists in exactly ONE production location — the helper `pkg/botutil/connect.go:42` itself. Zero inline production copies remain (remaining `:8090` hits are test fixtures only). 10 production callsites now route through DeriveAPIURL. The PR description's "single source of truth" claim is now true.

No regression introduced
- Each migrated site passes the same `cfg`/`ctx.GetConfig()` and consumes `apiURL` identically (avatar URL in commands.go, `BotRegisterResp.APIURL` in register.go, `generateXxxMD(apiURL...)` in api.go). Var name unchanged; no dropped TrimSpace — DeriveAPIURL preserves the exact `TrimSpace(BaseURL)!="" ? BaseURL : http://<IP>:8090` semantics.
- Unused imports cleaned correctly: `fmt` dropped from register.go & api.go, retained in commands.go (avatar Sprintf); `strings` retained where still used. Local `go build` of all touched packages passes.

Security invariants intact on new head
- `connectInfo()` (app_bot.go:996) is still a literal 2-key map (`plugin_package`, `api_url`) — no token/secret exposure.
- Auth/scope gates and i18n `missingkey=error` behavior untouched; targeted tests still assert real invariants.

CI: Build / Vet / Lint / i18n / secret-scan / osv-scanner all green (Test job pending, non-blocking infra).

💬 Non-blocking (future cleanup, not for this PR)
- 🔵 Now that there's one source of truth, a follow-up could normalize trailing-slash / IPv6 formatting inside DeriveAPIURL.
- 🔵 `OCTO_BOT_PLUGIN_PACKAGE` accepts any non-blank value; operator-controlled, so fine, but npm-spec validation would fail misconfig cleaner.

APPROVE.

modules/botfather/api.go
```diff
@@ -107,10 +106,7 @@ func (bf *BotFather) Route(r *wkhttp.WKHttp) {
 // skillMD 返回skill.md文档
 func (bf *BotFather) skillMD(c *wkhttp.Context) {
 	cfg := bf.ctx.GetConfig()
-	apiURL := cfg.External.BaseURL
-	if strings.TrimSpace(apiURL) == "" {
-		apiURL = fmt.Sprintf("http://%s:8090", cfg.External.IP)
-	}
+	apiURL := botutil.DeriveAPIURL(cfg)
 	wsURL := botutil.DeriveWSURL(cfg)
```
OctoBoooot, line 110: **[major]** This handler (and the two below — `cliGuideMD` at `:118`, `setupNewbotMD` at `:126`) now correctly threads `apiURL := botutil.DeriveAPIURL(cfg)` per the prior round's fix, but the **package-name dimension is still hardcoded downstream** — same class of incompleteness as that fix was for the api_url dimension.

Byte-grep: `modules/botfather/skill.go` has **10 hits** of the literal `create-openclaw-octo` across the 3 generator funcs called from this file (`:111`, `:118`, `:126`):
- `generateSkillMD` (skill.go:9, :19) — description + install command
- `generateCLIGuideMD` (skill.go:41, :54, :59, and 7 more in the command table at :66–:72) — install / bind / quickstart / full command table
- `generateSetupNewbotMD` — similar pattern

Operator-side effect: setting `OCTO_BOT_PLUGIN_PACKAGE=create-openclaw-canary` correctly updates BotFather DM templates (via `command.go`) and the App Bot connect response (via `app_bot.go`), but **the skill.go-served Markdown still serves `create-openclaw-octo`** to the same user. The Bot's CLI documentation and its DM prompts diverge under canary — exactly the drift class `PluginPackage()` was supposed to eliminate.

Smallest closing change: thread `botutil.PluginPackage()` through each generator (extend signatures with a `pluginPackage string` parameter, swap every `create-openclaw-octo` literal in the `Sprintf` format strings for `%s`/`{{.PluginPackage}}` interpolation). The three callsites on this file pass `botutil.PluginPackage()` next to the existing `botutil.DeriveAPIURL(cfg)`. Add regression tests mirroring `TestConnectGuidePackageParameterized` — assert injected canary renders AND the old literal is absent.

Caught by mochashanyao after my (now-retracted) APPROVE on review #1; full context + self-calibration in the summary body.

OctoBoooot (dismissed):
## Follow-up @ 39bdc30f (review #2)

**Retracting praise from review #1.** mochashanyao caught what I missed: my prior APPROVE praised "single source of truth is now byte-true" after byte-grepping `DeriveAPIURL`'s api_url callsites — but I **only grepped one dimension** (api_url literals), didn't grep the other (`create-openclaw-octo` package-name literals). The package name is still hardcoded at 10+ sites that don't go through `PluginPackage()`.

### Major (new at this delta)
- See inline at `modules/botfather/skill.go:9` — same class of incompleteness as the api_url major from the prior round, but for the package-name dimension. `botfather/skill.go` has three generator funcs (`generateSkillMD` / `generateCLIGuideMD` / `generateSetupNewbotMD`) with the literal `create-openclaw-octo` hardcoded 10+ times across the Markdown they emit. Those funcs are called from `botfather/api.go:111,118,126` — user-facing endpoints. Operator-side effect: setting `OCTO_BOT_PLUGIN_PACKAGE=create-openclaw-canary` correctly updates the DM templates and the App Bot connect response, but **the skill.go-served Markdown docs still serve `create-openclaw-octo`** to the same user.

### Retraction
My review #1 said:
> The "single source of truth" claim is now byte-true.

That was only verified for the api_url dimension. The package-name dimension was not byte-verified by me before writing the praise. mochashanyao independently caught it within 30 seconds of my post.

### Calibration note for the record
This is the fourth hit on the [praise-has-weight] memory in recent reviews (PR #423 / #442 / #447-round1 / this one). The common failure mode each time: I grep the *one* dimension I notice and write praise as if I checked all dimensions. Updated rule for myself: for any consolidation/source-of-truth PR with multiple dimensions, grep each dimension's literal pattern separately before writing per-dimension praise. The general "single source of truth" claim is now off my praise vocabulary unless I've grepped every dimension the helper covers.

### Verdict
Carrying REQUEST_CHANGES; my own review #1 APPROVE is superseded by this one (latest-per-reviewer).

yujiawei (approved):
# Code Review — PR #447 (octo-server)

Independent review at head SHA `39bdc30f82352bfd706608cfb51b25a32daec60d` against `main` (merge-base `8d2f4e34`). This PR is security-sensitive (token / credential / middleware-adjacent), so it received extra scrutiny on the token boundary, tenant isolation, and i18n render safety. I re-reviewed against the current head — the prior `CHANGES_REQUESTED` rounds targeted `f113d23b`, where the `DeriveAPIURL` consolidation was half-done; the head commit `39bdc30f` (`refactor: consolidate remaining api_url derivation onto botutil.DeriveAPIURL`) explicitly addresses that.

## 1. Spec / scope compliance — PASS

The change implements exactly what issue #446 (backend scope) and the linked spec ask, and nothing more:

- **`connect{plugin_package, api_url}`** is added to the two response builders that back all four App Bot read/create endpoints: `createBot` (`modules/app_bot/app_bot.go:428`, serves `POST /v1/admin/app_bot` + `POST /v1/space/:space_id/app_bot`) and `getBotDetail` (`app_bot.go:535`, serves the two `GET` variants).
- **Single source of truth** lives in the new `pkg/botutil/connect.go`: `PluginPackage()` (default `create-openclaw-octo`, env override `OCTO_BOT_PLUGIN_PACKAGE`) and `DeriveAPIURL(cfg)` (`External.BaseURL`, fallback `http://<ip>:8090`).
- **De-hardcoding** of the package literal in the four BotFather i18n templates (`quickstart` / `install` / `connect_prompt` / `created_prompt`, both `en-US` and `zh-CN`) → injected `{{.PluginPackage}}`.
- **`api_url` consolidation is now complete** within the scoped modules. A repo-wide grep confirms zero remaining inline `External.BaseURL` / `http://<ip>:8090` fallback copies in any of the five scoped files (`app_bot.go`, `botfather/command.go`, `botfather/api.go`, `bot_api/register.go`, `bot_api/commands.go`); the only `:8090` fallback now lives in `pkg/botutil/connect.go:42`.

**Nothing out-of-scope was touched** — the deprecated `modules/botfather/skill.go` (still hardcodes the literal by design), the octo-admin frontend, and `octo-lib config.Config` are all correctly left alone, matching the spec's out-of-scope list.

- Under-build: none.
- Over-build: none.
- Deviation: none.

## 2. Code Quality — Approved

I traced the four load-bearing risks for this security-sensitive change, adversarially attempting to refute each:

- **No token / secret leak (success-response data boundary).** `connectInfo()` (`app_bot.go:996-1001`) returns exactly `{plugin_package, api_url}` and references no token or secret. In both builders the top-level `token` stays a sibling key, never nested in `connect`: `createBot` returns the raw token at `app_bot.go:427` and `getBotDetail` returns it masked (`"****"+last4`, `app_bot.go:514-518`). Both helper values resolve only to public data (npm package name; public API URL). The e2e test asserts `connect` has exactly two keys and the token never appears inside it.
- **Auth / tenant gates unchanged.** The diff to `app_bot.go` is three additive hunks (two response keys + one helper); `checkSpaceAdmin`, `CheckLoginRole`, the `botInRouteScope` IDOR guard, and route registration are byte-for-byte unchanged. `connect` is assembled only on the post-gate success path and is identical for every caller/tenant (pure config/env, no `GetLoginUID` / `spaceID` / route-param input), so it adds no cross-tenant data and cannot widen isolation.
- **i18n render safety (`missingkey=error`).** The msgtmpl catalog renders with `missingkey=error` (`modules/base/common/msgtmpl/msgtmpl.go:84`), so any render site omitting `PluginPackage` would error and silently drop the DM. All four render sites in `command.go` (`:427`, `:436`, `:926`, `:936`) supply `PluginPackage`, `replyL` is the only indirection, and `renderProbe()` (`messages_test.go:26`) now includes it so the build-time completeness matrix exercises all four keys in every supported language. Confirmed there is no alternate render path: `sendBotSelectionList` only renders `MsgSelectBot*`, `cmdmenu` uses a separate catalog, and no `{{template}}` composition pulls these blocks elsewhere.
- **api_url consistency.** Routing the connect guide, BotFather docs, and `/bot/register` through one `DeriveAPIURL` keeps the bind command and the register response consistent, and makes a future derivation change a one-line edit.

Build, vet, and the catalog/unit tests pass locally (`go build`, `go vet`, `pkg/botutil` and `modules/botfather` render tests green).

No P0 / P1 / P2 findings.

## 3. Verdict

**APPROVED** — Spec compliant and no blocking quality findings.

## 4. Suggestions (non-blocking)

- Consider light validation/normalization of the `OCTO_BOT_PLUGIN_PACKAGE` override (trim is already done; a malformed value would flow verbatim into the `npx -y <pkg>` command). Operator-only knob, so low risk — fine to defer.
- The octo-admin `connectGuide.ts` follow-up must ship to complete the migration off the dead `openclaw-channel-dmwork` package on the frontend; until then the frontend fallback still owns the old name. Tracked separately under #446, so out of scope here — flagging only for sequencing awareness.

## 5. Additional notes

- This change is additive (new response key, parameterized templates with the same default literal the templates already carried), so existing consumers are unaffected and there is no behavior change to current DMs beyond the now-overridable package name.

an9xyz:
Re @OctoBoooot review #2 (the `skill.go` package-name dimension): it's a fair observation that this PR diverges the now-parameterized DM templates from `skill.go`'s still-hardcoded Markdown. **Decision for this PR: deferred to a follow-up — out of scope here.**

- `skill.go`'s `generate*MD` are the **deprecated** doc-pointer generators (`generateSkillMD` literally emits *"Deprecated … Use `create-openclaw-octo install` instead"*), served at `/v1/bot/*.md` — supplementary reference material, not the copy-paste DM / connect path this PR targets.
- It's listed under **Out of scope** in the linked spec brief; @yujiawei's re-review at this head concurs it's *"correctly left alone, matching the spec's out-of-scope list"* (Over-build: none), and Octo-Q rated it non-blocking **P2**.
- Follow-up scope is well-defined: thread `botutil.PluginPackage()` through the four generators + update `const_test.go`, mirroring `TestConnectGuidePackageParameterized`.

The `api_url` consolidation that was the prior round's blocker is fully resolved at this head (zero residual inline copies). This PR stays scoped to the App Bot `connect` data + the DM-template / `api_url` consolidation; the `skill.go` package-name parameterization will land in a separate PR.

---

Jerry-Xin (commented):
## Holding APPROVE — agree skill.go package-name drift is a real 🟡 follow-up, but it's out-of-scope for this PR (not a 🔴 blocker)

I byte-verified OctoBoooot's new finding on this head (`39bdc30f`). It's **mechanically grounded** but the severity is 🟡, not a blocker for #447 — here's the evidence.

### The divergence is real
`modules/botfather/skill.go` hardcodes the literal `create-openclaw-octo` **16×** across `generateSkillMD` / `generateCLIGuideMD` / `generateSetupNewbotMD` (it imports only `fmt`, never calls `botutil.PluginPackage()`). Those generators ARE served live (`api.go`: `GET /v1/bot/skill.md`, `/cli-guide.md`, `/setup-install.md`, `/setup-newbot.md`). Meanwhile the connect path (`connectInfo()` + DM command sites) resolves through `botutil.PluginPackage()` honoring `OCTO_BOT_PLUGIN_PACKAGE`. So an operator overriding the package name gets connect API + DMs saying the new name while these `.md` endpoints still emit `create-openclaw-octo`. Mechanically inconsistent — confirmed.

### Why it's 🟡 (out-of-scope), not 🔴 for #447
1. **`skill.go` is not in this PR's changeset.** The 12 changed files do not include `skill.go` — #447 neither introduced nor regressed this divergence. It's pre-existing tech debt predating the parameterization effort.
2. **The diverging surface is explicitly deprecated.** `generateSkillMD` self-declares `version: deprecated` / `# Octo Bot Skill (Deprecated)` and redirects users to the canonical external plugin repo. An operator on a renamed package hits a doc that already says "this is stale, use the canonical source." Lower user-facing harm than equal-authority API-vs-doc disagreement.
3. **Present-state values are identical** (`DefaultPluginPackage` == the hardcoded literal). Real divergence requires the override env to be set — drift risk, not a current bug.
4. Test note (matches mochashanyao's 🟡#2): `const_test.go:148` pins skill.go's literal (`assert.Contains(... "create-openclaw-octo install")`), and the `messages_test.go` parameterization regression guard only covers the DM `Render()` templates, not skill.go output — so skill.go escapes that guard. Test-maintenance/drift, not a runtime defect.

### Recommendation
Keep my APPROVE on #447 (the api_url consolidation + connect security are sound and complete). Track the skill.go gap as a **follow-up 🟡**: parameterize the (deprecated) generators through `botutil.PluginPackage()` or remove them, and un-pin `const_test.go:148`. Blocking #447 — which doesn't touch skill.go and whose own scope is fully addressed — on pre-existing deprecated-surface drift over-weights the finding.

PR #451, opened 2026-06-24
refactor(incomingwebhook): harden directed-render guard + wire-seam test (#450 follow-up)

an9xyz:
## Summary

Post-merge follow-up to #450 (which closed #448). Lands the two **non-blocking** P2 hardenings reviewers (yujiawei / Jerry-Xin / Octo-Q) recommended as a fast-follow. No behavior change for valid inputs; the payload wire shape is unchanged.

### 1. Invisible/confusable display-name guard (`mention.go`)
The forged-broadcast guard `isBroadcastLikeName` matched on `ToLower(TrimSpace(name))` only. A display name embedding a zero-width / bidi / full-width confusable — e.g. `所有<U+200B>人` (a ZWSP **inside** the label, the exact vector Jerry-Xin pinned on #450) — slipped the prefix match yet rendered as the visually-identical `@所有人`.

- `isBroadcastLikeName` now **NFKC-folds + strips Unicode format runes** (category `Cf`: zero-width ZWSP/ZWNJ/ZWJ/BOM/word-joiner + the LRE/RLE/…/bidi controls) before the label compare.
- The **rendered `@name`** is likewise stripped of invisibles, so the emitted pill can't carry an invisible confusable.
- Names that merely continue a label into a longer word (`所有人事部`, `allen`) still render — the boundary check is preserved, no over-block.

**Severity: cosmetic only.** Broadcast routing (`humans`/`ais`) is set exclusively by the capability-gated `assembleMention`, never derived from pill text — so this never could escalate to a real all-member ping. Closed as defense-in-depth on the token-in-URL push ingress.

### 2. Wire-seam test (`api.go`)
Reviewers flagged that the composed content + generated entities were asserted at `buildMention`'s return — one layer below the wire. Extracted `handlePush`'s payload assembly into a **behavior-preserving** `assemblePushPayload`; a new MySQL test (`TestAssemblePushPayload_DirectedRenderReachesWire`) asserts the composed `payload["content"]` + entities reach the map handed to `SendMessageWithResult` — with **no live WuKongIM send**.

## Testing
- **Pure:** `isBroadcastLikeName` boundary + NFKC / invisible / bidi / full-width confusables (inputs built via `rune()` so there are **no literal invisible bytes in source**); a real ZWSP-bearing name renders with the invisible stripped.
- **DB-backed (MySQL):** the `assemblePushPayload` wire-seam; existing `buildMention` directed-render + mutual-exclusion + non-member tests.
[...]

PR #450, opened 2026-06-24
feat(incomingwebhook): visible mention render layer — broadcast pills + directed name resolution

an9xyz:
## Summary

Closes #448 — the **visible @mention pill** follow-up to #445 (functional flags) and #449 (caller-supplied entities). It ships the two halves #448 asked for, both at the render layer of the **native** incoming-webhook push, using only existing wire fields (zero client change):

- **① Directed name-resolution** (`mention.render`, item ① b): a caller that sends only `mention.uids` gets the server to resolve each member's display name, prepend `@<name> ` to the text content, and **generate** the UTF-16 `mention.entities` — so a bare uid renders a clickable, correctly-bound pill without the caller writing the name/offsets itself.
- **② Broadcast pills** (item ②): when a permitted `mention.all`/`mention.bots` is set, the server prepends the canonical broadcast label (the 所有人 / 所有AI tokens) so every client renders the broadcast pill (previously the flags fired the red-dot/notify/bot-summon but showed no pill).

Both compose in one prepend pass (broadcast first, directed after); routing — red-dot / notify / `ExpandAisToBotUIDs` — is unchanged, and a push that uses neither is byte-identical to today.

## Why this shape (cross-client grounding)

Re-verified against the three client repos: **all three render a pill only from the literal in the content text** — none fabricate one from a flag. iOS regex-scans content; Android `indexOf`-highlights (flag-gated); web `segmentText` regex-matches the name in content. And the directed wire shape (`content` + `{uid,offset,length}` + `uids`) is exactly what `octo-web mentionSendParse` emits when a human @s a member — so a server-rendered directed @ is wire-identical to a client-sent one (web offset-authoritative, Android validates, iOS positional). Offsets are **UTF-16 code units**.

## Changes

- `model.go`: `mentionReq.Render` (opt-in, default off, **request-only** — not on the wire; #445/#449 callers unchanged).
- `db.go`: `filterGroupMembers` now returns `uid→name` (`LEFT JOIN user`, `IFNULL(user.name,'')`, the group member-list source); membership is still key-presence.
- `mention.go`: unified `composeMentionContent` (broadcast literals + directed `@name`s + generated entities + a rune budget). `buildMention` returns `(mention, content, ignored)` and attaches **either** the generated directed entities **or** the shifted caller entities (`render` and caller `entities` are mutually exclusive — #449 is authoritative).
[...]

PR #449, opened 2026-06-24
feat(incomingwebhook): caller-supplied mention.entities (directed @-pills) on native push

an9xyz:
## Summary

Adds the **visible render layer** for directed @mentions on the native incoming-webhook push endpoint. #445 shipped the functional layer (red-dot / bot summon via `mention.uids`); this lets a caller also render a **clickable @-pill bound to the right member** by supplying `mention.entities` (`[{uid, offset, length}]`) alongside `content` + `mention.uids`. The server validates and passes entities through verbatim — **no name resolution, no `content` mutation**.

Target wire shape (a directed @ of a bot — exactly what an app client already emits):

```json
{
  "content": "@example-bot run it",
  "type": 1,
  "mention": {
    "uids": ["27oErgk8Leib2d99669_bot"],
    "entities": [{ "uid": "27oErgk8Leib2d99669_bot", "offset": 0, "length": 11 }]
  }
}
```

## Related Issue

Addresses the **directed-@ half of #448** (visible-pill consistency). Cross-client analysis posted on #448 confirms all three clients consume this exact shape: web binds authoritatively by offset (`parseMentionWithEntities`), Android validates/keeps the wire entity (`MentionEntityHelper`), iOS ignores entities and re-parses by position — so the shape is universal and needs **zero client changes**. The broadcast-pill auto-compose (the other half of #448) is a deliberate follow-up.

## Linked Spec

- `.octospec/tasks/incoming-webhook-mention/brief.md` — this extends the #445 brief with the render-layer (`entities`) the brief deferred.
- #448 — v2 follow-up tracking the visible layer.

## Changes

- `model.go`: `mentionReq` gains `Entities []json.RawMessage` (per-item lenient decode).
- `mention.go`: `decodeEntities` (structural sanity, per-item drop, capped), `finalizeEntities` (member gate + UTF-16 in-bounds + must point at `@` + de-overlap, first-wins), wired into `buildMention` — text msg_type only; entity uids join the **same single** `filterGroupMembers` query as directed uids.
- `pkg/mentionrewrite/rewrite.go`: new `EntitiesKey` constant (the package already documents `mention.entities` as pass-through).
- No `api.go` / `db.go` changes: entities ride the existing `handlePush` → `CloneForExpansion` → `ExpandAisToBotUIDs` path, which preserves all `mention` sub-keys and only mutates `uids`.

[...]

PR #454, opened 2026-06-24
feat(incomingwebhook): deliver to thread (子区) channels

### LINKED ISSUE #453: feat(incomingwebhook): bind incoming webhooks to threads (子区)
## Problem

Incoming webhooks (`modules/incomingwebhook`) can only deliver to a **group** channel today. The delivery target is hardcoded to the parent group in three places (`handlePush`, `testPush`, mention/@AI expansion), and the `incoming_webhook` row stores only `group_no`/`space_id`. There is no way to point a webhook at a specific **thread (子区)**, so CI/platform integrations that belong in a focused thread have to spam the whole group.

## Proposal

Let a webhook optionally be **bound to a thread** so its pushes deliver into the thread channel (`group_no____short_id`, `ChannelTypeCommunityTopic = 5`) instead of the parent group.

- New thread-scoped management mount `/v1/groups/:group_no/threads/:short_id/incoming-webhooks` (+ the bot-token face). Binding is persisted on the row and is immutable after create.
- **Push URL and body unchanged** — the target is resolved server-side from the row (`channel_type`/`thread_short_id`), so the push caller and the IM client need zero adaptation.
- All auth / membership / quota / cascade gates and @mention / @AI expansion stay anchored on the parent `group_no` (a thread's membership derives entirely from its parent group). The only thing that changes is the `SendMessage` channel.

## Decisions

- **Quota**: thread webhooks share the parent group's `max_per_group` (no per-thread quota).
- **Lifecycle (v1)**: group disband still cascades (keyed on `group_no`); thread delete/archive does **not** actively cascade-disable — delivery to a banned/deleted thread fails downstream with 502 (TODO: a thread-delete event listener).
- No per-push thread-liveness query on the hot path.

## Spec

`.octospec/tasks/incoming-webhook-thread/brief.md`

an9xyz:
## Summary

Let an incoming webhook be bound to a **thread (子区)** so its pushes deliver into the thread channel (`group_no____short_id`, `ChannelTypeCommunityTopic`) instead of only the parent group. The binding is persisted on the row and resolved **server-side** at push time — the push URL and body are identical to a group webhook (push-caller + IM-client zero adaptation). All auth / membership / quota / cascade gates and @mention / @AI expansion stay anchored on the parent `group_no`; only the `SendMessage` channel target changes.

## Related Issue

Closes #453

## Linked Spec

`.octospec/tasks/incoming-webhook-thread/brief.md`

## Changes

- **migration**: `channel_type` (default 2=group) + `thread_short_id` (default `''`) on `incoming_webhook`, plus `idx_incoming_webhook_thread`. Existing rows keep group semantics.
- **model**: `targetChannel()` as the single source of `(channelID, channelType)`; `webhookResp` echoes `thread_short_id`. Binding is immutable (not in `updateFieldsAllowed`).
- **routes**: the same management handler set is mounted under `/v1/groups/:group_no/threads/:short_id/incoming-webhooks` (user + bot faces). `create` validates the thread exists / is active / belongs to the group via `GetThread`; `list` and `queryManageable` enforce `(group_no, short_id)` scope isolation.
- **delivery**: `handlePush` / `testPush` send to `m.targetChannel()`. @AI expansion stays parent-group-anchored (`ExpandAisToBotUIDs` is GROUP-only; a thread's bots are the parent group's bots).
- **errors/i18n**: new `ErrIncomingWebhookThreadNotFound` (404) + zh-CN + regenerated markers.
- **docs**: README + thread.md.

## Testing

- [x] Manually verified

Ran against a CI-faithful testbed (MySQL 8.0 + Redis + **WuKongIM v2.2.4**):
- `TestTargetChannel` (pure unit) pins the channel mapping.
- thread create / reject (missing · invalid · archived · cross-group) / scope-isolation e2e over real HTTP + DB.
- `TestThread_Push_Delivers` delivers to the **real thread channel** on v2.2.4.
- `go test ./modules/incomingwebhook/ -race -shuffle=on -count=1` → `ok`, 0 fail / 0 skip / 0 data race.
- `CGO_ENABLED=0 go build ./...`, `go vet`, `golangci-lint`, `make i18n-extract-check`, `make i18n-lint` all green.

## COMPREHENSION

1. **What it does to the load-bearing path**: adds a per-row delivery target; the send sites resolve `(channelID, channelType)` from the row via `targetChannel()`. Thread webhooks → `(group_no____short_id, CommunityTopic)`; group webhooks unchanged. Management gains a thread-scoped mount that validates the thread and isolates scope. All auth / quota / disband-cascade / mention logic continues to key on the parent `group_no`.

2. **What could break**: (a) existing group webhooks — mitigated: new columns default to group semantics, `targetChannel()` defensively defaults to group, and all pre-existing group/mention push tests still pass. (b) group-path `list` now excludes thread webhooks — intended scope isolation, no regression since no thread webhooks existed before. (c) delivery needs a WuKongIM that supports CommunityTopic channels — true on the CI-pinned v2.2.4 (verified); the push e2e is gated to skip on incapable IMs. (d) @AI must stay parent-group — preserved (`assemblePushPayload` expands with `(ChannelTypeGroup, m.GroupNo)`).

3. **How I know it works**: `TestTargetChannel` pins the mapping; real-HTTP create/scope/reject e2e; `TestThread_Push_Delivers` delivers to the real thread channel on v2.2.4; full package `ok` under `-race -shuffle=on`; all existing group/mention push tests pass (no regression).

## Checklist

- [x] Updated documentation

https://claude.ai/code/session_01FfQCb4Xcef89rTBhfJ3bT6

---
_Generated by [Claude Code](https://claude.ai/code/session_01FfQCb4Xcef89rTBhfJ3bT6)_


OctoBoooot (changes requested):
## Review: feat(incomingwebhook): deliver to thread (子区) channels (#454)

**Verdict:** Request changes — CI red on `check-sprint / check-sprint` (process gate: no linked issue) precludes APPROVE per skill rule §11. Code itself is **clean — no blockers, no majors, no minors, no nits**. Once check-sprint clears, this is an APPROVE candidate without code changes.

Size: 14 files / 745 lines → `needs-human-review` label applied (file count threshold). Risk: normal classify but trust-boundary territory (incomingwebhook routing + new channel target type). Author: an9xyz (100+ merged / 0 closed-unmerged).

### Adversarial verification pass (per skill §7.1)

All 10 trust-boundary invariants the design relies on byte-verified:

- **Auth/membership stays parent-group anchored** — `mention.go:281` `filterGroupMembers(m.GroupNo, ...)`, push handler's creator-membership checks key on `m.GroupNo`, mention members derived from parent group. Thread doesn't introduce a new membership domain.
- **AI expansion stays parent-group anchored** — `api.go:1190` `ExpandAisToBotUIDs(wire, common.ChannelTypeGroup.Uint8(), m.GroupNo, ...)` — explicitly passes `ChannelTypeGroup` + parent `GroupNo`, not the target channel. This is the easy-to-get-wrong case; author keyed it correctly. Comment names the rationale ("thread's bots are the parent group's bots").
- **Quota counts thread webhooks against parent group** — `db.go:50+` `insertWithQuota` filters `WHERE group_no=? AND status != statusDeleted` (no thread_short_id filter). Thread webhooks share the group's quota pool — prevents quota explosion (10 threads × N per-thread would be unbounded otherwise). Per-creator quota same shape.
- **Disband cascade covers thread webhooks** — `db.go:294` `disableByGroupNo` filters `WHERE group_no=?` only — disables ALL webhooks under that group_no including thread-bound ones. `api.go:1667` comment explicitly notes "子区 webhook（同 group_no）一并被禁用".
- **Binding immutability** — `db.go:146+` `updateFieldsAllowed` excludes `channel_type` and `thread_short_id`. No code path can change a webhook's target post-creation. Binding is set ONCE at create.
- **GetThread validates jointly** — `api.go:686-700` checks (a) `IsValidShortID` format, (b) `GetThread(groupNo, shortID, "")` joint-key query, (c) `tr.Status == thread.ThreadStatusActive`. Prevents cross-group binding, missing thread binding, archived thread binding.
- **Scope isolation on management** — `queryByGroupNo` filters `thread_short_id = ""` (group-face only sees group webhooks), `queryByThreadScope` filters `(group_no, thread_short_id)` exact (thread-face only sees its own thread webhooks). `queryManageable` at `api.go:612` cross-checks `m.ThreadShortID != c.Param("short_id")` — webhooks fetched via the thread route MUST match the URL's short_id.
- **Migration safety** — `channel_type DEFAULT 2` preserves group semantics for existing rows; `thread_short_id DEFAULT ''` empty for existing rows. `ADD COLUMN` at table end on MySQL 8.0 is INSTANT-eligible. Reversible Down migration. `targetChannel()` defensively defaults to group when `ThreadShortID == ""` even if `ChannelType != Group` (belt-and-suspenders).
- **Composite index** — `idx_incoming_webhook_thread(group_no, thread_short_id, status)` services both query paths: prefix match (group-face: `group_no=? AND thread_short_id=''`) and full match (thread-face: `group_no=? AND thread_short_id=X AND status!=...`). One index, two query patterns.
- **Description-vs-diff** — every claim verified:
  - "All auth / membership / quota / cascade gates stay anchored on parent group_no" → verified at 5 distinct call sites.
  - "ExpandAisToBotUIDs is GROUP-only; a thread's bots are the parent group's bots" → confirmed at api.go:1190 with `ChannelTypeGroup` literal.
  - "Binding is immutable (not in updateFieldsAllowed)" → grep'd updateFieldsAllowed map confirms absence.
  - "Push URL and body are identical to a group webhook" → push routes unchanged, only the `SendMessage` channel target changes via `targetChannel()`.

### Praise

- **`targetChannel()` as single resolver** — `handlePush` and `testPush` both call `m.targetChannel()`; no drift surface. This is the source-of-truth pattern that took PR #447 three rounds to achieve for `DeriveAPIURL` (where 6 inline copies had to be discovered). #454 gets it right from the start with two call sites both going through one method.
- **Binding immutability via `updateFieldsAllowed` exclusion** turns "the binding is forever" from a documented contract into a runtime invariant — anyone who later adds `channel_type` or `thread_short_id` to that map fails the immutability test rather than silently enabling rebinding. Same shape as PR #449/450's mutual-exclusion contracts.
- **Three independent gates (auth/quota/cascade/mention) all derive from the SAME `m.GroupNo` field**, NOT from the target channel. Eliminates the entire class of "thread webhook escaped parent group's policies" bugs. The mention path's `filterGroupMembers(m.GroupNo, ...)`, the AI expansion's `(ChannelTypeGroup, m.GroupNo)`, the disband cascade's `WHERE group_no=?`, the quota's `WHERE group_no=?` — all consistent. This kind of "one anchor, multiple gates" is the right shape for trust-boundary code.
- **GetThread validates `(group_no, short_id)` jointly** — prevents the cross-group binding attack (a member of group A would otherwise be able to bind a webhook to a thread in group B if they happened to know its short_id). The validation lives at the create-time chokepoint where mistakes are cheapest to correct.
- **Composite index `(group_no, thread_short_id, status)`** is the minimal index that services both query patterns (group-face prefix match + thread-face exact match) — author didn't add separate indexes for the two scopes, recognized the natural prefix relationship. Subtle DB-design correctness.
- **`targetChannel()` defensive default** — falls back to group semantics when `ThreadShortID == ""` regardless of `ChannelType`. Belt-and-suspenders for migration drift / mismatched columns.

- Author notes a "future: thread-delete event listener for per-thread cascade" follow-up at `api.go:1672`. Currently a thread's deletion does NOT cascade-disable its webhooks (only group disband does). Acceptable as documented limitation — thread deletion is rare and the webhook would just silently fail (404 channel) until manually deleted.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#454 Review Report
## feat(incomingwebhook): deliver to thread (子区) channels
**Head SHA:** e90d4e58a3f08060eb66c558a156854d814b0b4a

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Migration backward-compat | ✅ | `sql/20260624000001`: `channel_type DEFAULT 2`, `thread_short_id DEFAULT ''`, `INSTANT` algo on MySQL 8.0 → existing rows return group semantics |
| `targetChannel()` single source | ✅ | `api.go:1500-1504`: defensive fallback to group for any non-thread-bound combination |
| Scope isolation (list) | ✅ | `db.go:121-128` `queryByGroupNo` adds `thread_short_id=''`; `db.go:133-143` `queryByThreadScope` symmetric |
| Scope isolation (queryManageable) | ✅ | `api.go:612`: `m.ThreadShortID != c.Param("short_id")` — empty on group route, populated on thread route |
| Auth parity (user + bot faces) | ✅ | `api.go:220-222` (user thread mount), `bot_api/incoming_webhook.go:30-32` (bot thread mount) — both same middleware chain + `MountManagementRoutes` |
| Push path unchanged | ✅ | `api.go:1428-1433` (handlePush), `api.go:1119-1123` (testPush) — both via `m.targetChannel()` |
| @AI expansion parent-anchored | ✅ | `api.go:1189-1190`: `ExpandAisToBotUIDs(wire, common.ChannelTypeGroup.Uint8(), m.GroupNo, ...)` — hardcoded group type |
| Quota shared parent-group | ✅ | `db.go:74`: `SELECT count(*) FROM incoming_webhook WHERE group_no=? AND status != ?` — counts all webhooks per group_no |
| Cascade disband | ✅ | `db.go:294-300`: `disableByGroupNo` keys on `group_no`, covers thread webhooks |
| Binding immutable | ✅ | `db.go:147-153`: `updateFieldsAllowed` does NOT include `channel_type` or `thread_short_id` |
| Thread validation at create | ✅ | `api.go:678-695`: `IsValidShortID` → `GetThread(groupNo, shortID, "")` → `tr.Status != ThreadStatusActive` reject |
| Anti-enumeration on push | ✅ | Push URL/body unchanged; thread target only from persisted row; no thread identifier in request |

## 2. Findings

### P2-1 — Thread delete/archive does not cascade-disable bound webhooks (pre-existing design decision, documented)

**Severity:** P2 (known lifecycle gap, 502 fallback, explicit TODO)
**Diff-scope:** new — this is a new feature; the gap exists because no cross-module thread-delete event is available in v1.

A thread-bound webhook whose thread is later deleted will remain `statusEnabled` in the DB. Pushes to it will fail at the WuKongIM `SendMessage` layer (502 `push_delivery_failed`). This is explicitly documented (`api.go:1669-1673` TODO) and the 502 fallback prevents silent data loss. Group disband still cascades correctly via `disableByGroupNo`.

**Recommendation:** Add a thread-delete event listener in a follow-up, symmetric with `handleGroupDisband`. Not blocking.

### P2-2 — `GetThread` DB failure conflated with "thread not found" (acceptable for cold path)

**Severity:** P2 (information loss on error, not a correctness issue)
**Diff-scope:** new

`api.go:686-693`: When `GetThread` returns an error (DB failure, not "thread not found"), the code logs at `Warn` level and returns 404 `mgmtThreadNotFound`. For a cold-path create operation this is acceptable — the user can retry. The comment at `api.go:684` explicitly notes this design choice. Not blocking.

## 3. Suggestions

1. **Follow-up: thread lifecycle cascade** — implement the TODO at `api.go:1669-1673` to subscribe to a thread-delete event and call `disableByGroupNoAndThreadShortID` (new function). Consider also archiving (webhook to archived thread could either be disabled or left to the auto-unarchive-on-delivery semantics).

2. **Follow-up: per-thread list in admin UI** — scope isolation is correct for the API, but admin dashboards may need a "show all webhooks across all threads in this group" view for operational visibility.

## 4. Additional Findings

- **Migration safety:** The `ALTER TABLE ... ADD COLUMN` with `INSTANT` algo (MySQL 8.0) is metadata-only, zero downtime. The composite index `idx_incoming_webhook_thread(group_no, thread_short_id, status)` supports both `queryByGroupNo` (prefix: `group_no` + `thread_short_id=''`) and `queryByThreadScope` (full prefix). Down migration cleanly drops index + columns.
- **Test coverage quality:** `TestTargetChannel` pins the wire-level channel format (`group_no____short_id`), `TestThread_ScopeIsolation` covers cross-scope management 404, and `TestThread_Push_Delivers` is a real e2e gated on WuKongIM capability. Test suite is thorough.

## 5. Data-Flow Tracing

| Consumed data | Upstream source | Flow verified |
|---|---|---|
| `m.ChannelType` / `m.ThreadShortID` | `create()` at `api.go:717-718` → DB insert | ✅ Flows through `insertWithQuota` → `incoming_webhook` row → `cachedQueryByWebhookID` → `targetChannel()` |
| `targetChannel()` output `(channelID, channelType)` | `api.go:1500-1504` reads `m.ChannelType` + `m.ThreadShortID` + `m.GroupNo` | ✅ `handlePush` (`api.go:1428`) and `testPush` (`api.go:1119`) both consume it |
| `thread.BuildChannelID(GroupNo, ThreadShortID)` | `thread/service.go:830-832`: `fmt.Sprintf("%s____%s", groupNo, shortID)` | ✅ Same format pinned in `TestTargetChannel` (`thread_target_test.go:52`) |
| `c.Param("short_id")` for scope isolation | Route group path `/v1/groups/:group_no/threads/:short_id/incoming-webhooks` | ✅ Empty on group route (group route has no `:short_id` segment), populated on thread route |
| `GetThread(groupNo, shortID, "")` validation | `thread/service.go:519-549`: queries `thread` table by `(group_no, short_id)`, rejects deleted | ✅ Returns error for missing/deleted, returns `ThreadResp` with `Status` for archived check |
| `ExpandAisToBotUIDs(wire, ChannelTypeGroup, m.GroupNo, ...)` | Hardcoded `ChannelTypeGroup.Uint8()` + `m.GroupNo` at `api.go:1189-1190` | ✅ @AI expansion stays parent-group regardless of webhook's channel type |
| Quota count `WHERE group_no=? AND status != statusDeleted` | `db.go:74-77` counts ALL non-deleted webhooks per group_no | ✅ Thread webhooks share parent-group quota (no `thread_short_id` filter) |

- **C1 — Double-path parity (symmetric paths):** ✅
  - User face (`api.go:220-222`) and bot face (`bot_api/incoming_webhook.go:30-32`) both mount thread-scoped routes via the same `MountManagementRoutes`. Middleware chains are identical (auth + rate limit). No parity drift.
  - Create ↔ delete: create validates thread via `GetThread`; delete goes through `queryManageable` which enforces scope isolation (`m.ThreadShortID != c.Param("short_id")`). No asymmetric access.
  - `queryByGroupNo` (group list) and `queryByThreadScope` (thread list) are symmetric with mutual exclusion via `thread_short_id=''` vs specific value.

- **C2 — Control-flow ordering / nested reuse:** ✅
  - `targetChannel()` is called exactly once per push/testPush — no double-resolution risk.
  - `assemblePushPayload` → `ExpandAisToBotUIDs` is hardcoded to `ChannelTypeGroup` regardless of webhook's actual channel type. No ordering or nesting issue.
  - No security-control regex or sanitization introduced in this PR.

- **C3 — Authorization boundary ≠ capability boundary:** ✅
  - Thread-scoped management routes go through the same `AuthMiddleware` / `authBot` + `resolveActor` as group routes. No new capability exposed without auth.
  - Push path: token-in-URL auth, delivery target derived from persisted row only (not from request). No new trust surface.

- **C4 — Authorization lifecycle / container-member state cascade:** ✅
  - Group disband cascades via `disableByGroupNo` (covers thread webhooks). Thread-level lifecycle (delete/archive) intentionally not cascaded in v1 with documented fallback.
  - Container (group) status is checked on push path via `cachedRequireActiveGroup` — applies to both group and thread webhooks equally.

- **C5 — Build/note ≠ runtime correctness:** N/A
  - No browser extension, CLI, or packaging changes. Server-side Go code; standard build.

- **C6 — Governance / policy / security document consistency:** N/A
  - No governance, policy, or security documentation changes. README and thread.md are technical documentation updates.

---

[Octo-Q] verdict: **APPROVE**

No P0 or P1 findings. The PR is well-architected with clean data flow, proper scope isolation, correct backward compatibility, and thorough test coverage. Two P2 observations (thread lifecycle cascade gap, GetThread error conflation) are explicitly acknowledged by the author with documented TODOs and acceptable fallbacks. The security posture is sound: delivery target is derived solely from the persisted row (anti-forgery), auth/membership/quota stay anchored on the parent group, and the push URL is unchanged (zero caller adaptation).

Jerry-Xin (dismissed):
The PR is in scope for `octo-server` and the implementation matches the intended incoming-webhook/thread behavior. I found no blocking correctness or security issues.

💬 Non-blocking

🟡 Warning — `modules/incomingwebhook/api.go:686`: `GetThread` errors are all collapsed to `mgmt_thread_not_found` at line 692. That means a real DB/service failure returns 404 instead of 500. This is acceptable for now if intentional, but it weakens operational diagnosis and retry behavior.

🔵 Suggestion — `modules/incomingwebhook/api_i18n_test.go:77`: add `mgmtThreadNotFound` to `TestIncomingWebhookRespondHelpers` so the new i18n responder is pinned like the existing management helpers.

✅ Highlights

The persisted-row target model is clean: `targetChannel()` centralizes delivery mapping and both `handlePush` and `testPush` use it.

Parent-group anchoring is preserved for auth, membership, quota, group lifecycle, and mention expansion (all gates anchored to the parent `GroupNo`, so a thread target cannot escape parent-group policy).

Thread management scope isolation is consistently applied through `ThreadShortID` in list and `queryManageable`; `GetThread(groupNo, shortID, ...)` jointly validates so a thread under another group cannot be bound (cross-group injection blocked). AI expansion correctly passes the parent group channel, not the thread channel.

I ran `go test ./modules/incomingwebhook/ -run TestTargetChannel -count=1` and `go test ./modules/incomingwebhook/ -run TestIncomingWebhookRespondHelpers -count=1`; both passed. Full `go test ./modules/incomingwebhook/` was blocked by a local database migration state (`unknown migration in database` for `20201222000001_report_legacy01.sql`), a local-env issue, so I did not use that run as a PR signal.

yujiawei (dismissed):
# Code Review — PR #454 (octo-server)

Reviewed at head `1a8f1469` (re-reviewed after the follow-up fix commit; the original review SHA `e90d4e58` is superseded). Verdict applies to the current head.

## 1. Spec compliance

**✅ Pass.** Checked the diff against the linked task brief's Acceptance and Out-of-scope lists.

- **No under-build:** migration adds `channel_type` (default 2) + `thread_short_id` (default `''`) + composite index; `targetChannel()` is the single source of `(channelID, channelType)`; thread-scoped mount on both user and bot faces; create-time validation (exists / under this group / active); scope isolation in `queryManageable` and `list`; `webhookResp` echoes `thread_short_id`; new 404 error code + zh-CN/en-US markers.
- **No over-build:** no per-thread quota, no new system_setting, no per-push thread-liveness query, no push-URL/body change — all consistent with the stated out-of-scope.
- **No deviation:** the create-time binding model (target encoded only on the persisted row, resolved server-side, zero push-caller adaptation) matches the brief's invariant. All auth / membership / quota / disband-cascade / mention / @AI gates remain anchored on the parent `group_no`; only the final `SendMessage` channel changes.

## 2. Code quality

**Approved.** The trust-boundary posture is sound and independently verified:

- **Delivery target is row-derived only.** `handlePush` and `testPush` both resolve via `m.targetChannel()`; `pushPayloadReq`/`createReq` carry no channel/thread field, so URL/query/body cannot influence the target (anti-forgery preserved, same as `space_id`/`extra` discard).
- **Binding is immutable** — `channel_type`/`thread_short_id` are absent from `updateFieldsAllowed`, so the create-time validation is trustworthy on the hot path.
- **Cross-group / cross-space binding is structurally impossible** — `ExistThread`/`GetThread` key on `(group_no, short_id)`; `space_id` is taken from the parent group.
- **Scope isolation holds** — `queryManageable` adds `m.ThreadShortID != c.Param("short_id")`; the group mount (no `:short_id`) matches only `thread_short_id=''`, the thread mount only its own. Group↔thread cross-management returns 404 (pinned by `TestThread_ScopeIsolation`).
- **Backward compatibility is byte-for-byte** — column defaults make legacy rows resolve to `(group_no, ChannelTypeGroup)`; `queryByGroupNo`'s new `thread_short_id=''` filter still matches all pre-existing rows. `TestTargetChannel` covers the legacy-zero and defensive-fallback cases.
- **The follow-up fix is correct** — thread-lookup now distinguishes a DB/service failure (→ 500, retryable + alertable) from a genuinely missing/cross-group thread (→ 404), instead of collapsing every error to a misleading 404. `ExistByGroupNoAndShortID` correctly excludes soft-deleted threads.

`go build`, `go vet`, and the pure unit tests (`TestTargetChannel`, responder-helper, no-legacy-response guard) pass locally.

1. **Migration comment overstates the algorithm.** `modules/incomingwebhook/sql/20260624000001_incomingwebhook_thread.sql:11-17` comments the ALTER as `INSTANT / 瞬时无锁`, but the statement combines two `ADD COLUMN` with one `ADD INDEX`. A single ALTER gets one algorithm, and a secondary `ADD INDEX` cannot be `INSTANT` on MySQL 8.0 — so the whole statement degrades to `INPLACE` (full index build + brief metadata locks), not instant. No `ALGORITHM=`/`LOCK=` clause is pinned, so the degradation is silent. The repo already has the correct pattern in `modules/oidc/sql/20260515000001_oidc_bind_uniques.sql` (`ALGORITHM=INPLACE, LOCK=NONE`, with rationale to fail loud on non-online environments). Suggest either splitting the column-add (INSTANT) from the index-add (INPLACE), or correcting the comment and pinning `ALGORITHM=INPLACE, LOCK=NONE`.

2. **Create-time thread validation is a benign TOCTOU.** The `ExistThread`/`GetThread` active-check runs outside the `insertWithQuota` parent-group `FOR UPDATE` transaction, so a thread can be archived/deleted between check and commit. This is acceptable and explicitly documented (the 502 push-delivery fallback + the `TODO(thread-cascade)` for a future thread-delete listener). Recorded only so a future re-review does not treat it as a regression.

3. **`targetChannel()` defaults non-canonical rows to the parent group** rather than erroring. This fail-open is toward the webhook's *own* parent group (never an scope escalation), and such rows cannot be produced via the immutable create path, so it is safe by construction — but a check-level invariant (`channel_type=5 ⟺ thread_short_id!=''`) would make the intent explicit and guard against future writers.

4. **Test-coverage gaps on stated acceptance items** (implementations are correct via shared, unchanged parent-group-anchored paths; only assertions are missing): thread-dimension `testPush` + `adapter=test` audit; shared `max_per_group` quota across group+thread (409); creator-left / disband-cascade / group-not-Normal gates on a thread webhook; `@AI`/`mention.bots` expansion using parent-group identity on a thread webhook; and the bot-face thread mount under the permission matrix. Adding a few thread-scoped assertions would lock in these security-relevant invariants against regression — worth doing given the sensitivity, but not blocking since each path is shared with already-tested group behavior.

## Manual-verification notes for a human

Two surfaces are outside this diff and could not be exercised statically: (a) WuKongIM's own authorization of the `iwh_` sender into a `CommunityTopic` channel at runtime (the push e2e self-skips unless the IM supports it); (b) downstream message-module listeners (red-dot / bot-wake) when `channelType=CommunityTopic` vs `Group`. Worth a manual smoke test on a thread-bound webhook against the CI-pinned IM before/after merge.

OctoBoooot (dismissed):
## Review: feat(incomingwebhook): deliver to thread (子区) channels (#454) — delta @ 1a8f1469

**Verdict:** Approve.

Delta scope from prior REQUEST_CHANGES @ `e90d4e58` (driven by CI-red process gate, 0 code finding): 1 commit `1a8f1469` "fix(incomingwebhook): surface thread-lookup DB failure as 500; pin thread-not-found responder test" — 2 files, +19/-4. **CI now green.**

### What changed (byte-verified)
- **`api.go:683-705` — split thread validation into ExistThread → GetThread** (closes Jerry-Xin's prior 🟡):
  - `ExistThread(groupNo, shortID)`: DB/service error → `mgmtQueryFailed` (500), not exists/cross-group → `mgmtThreadNotFound` (404). The 500 lets operators alert on 5xx + lets clients retry on transient failures.
  - `GetThread(groupNo, shortID, "")` after ExistThread succeeds: only checks `Status == Active`. Residual error window is tiny (between two queries) and degrades to 404 (acceptable per the new comment).
- **`api_i18n_test.go:94` — added `mgmtThreadNotFound` test case** (closes Jerry-Xin's other 🟡) — pins HTTP 404 + i18n key `err.server.incomingwebhook.mgmt_thread_not_found`.

### Retraction (my prior #454 review)

My prior review praised:
> Migration `(group_no, thread_short_id, status)` composite index... ADD COLUMN at table end on MySQL 8.0 is INSTANT-eligible.

**That praise is byte-false** — caught by yujiawei this round. The migration's ALTER combines `ADD COLUMN + ADD COLUMN + ADD INDEX` in one statement; MySQL 8.0 picks the most restrictive algorithm for combined operations, and `ADD INDEX` is INPLACE only (never INSTANT). The actual algorithm is INPLACE, which on a large table costs more time and locks writes briefly during prepare/commit. The "INSTANT-eligible" claim is wrong on the combined statement. yujiawei's 🟡 #1 stands.

This is my **sixth same-root-cause [praise-has-weight] failure** this session (#423 mdLinkText / #442 DownloadURL / #447 DeriveAPIURL / #447 PluginPackage / #451 stripFormatRunes / #454 INSTANT migration). Same failure mode each time: I praise a property after verifying only a subset of the conditions that property requires. Each failure narrows the rule further — this one adds:

> For SQL DDL praise involving algorithm tier (INSTANT / INPLACE / COPY), verify (a) single vs combined ALTER statement (combined picks most restrictive), (b) MySQL version, (c) FK / partition / trigger constraints. `ADD COLUMN + ADD INDEX` in MySQL 8.0 → INPLACE, never INSTANT. Don't write "INSTANT-eligible" without grepping the statement for INDEX/UNIQUE/FOREIGN KEY operations.

### Prior items
- ✅ **CI gate cleared** — check-sprint no longer failing. Verdict floor lifts from RC per rule §11.
- ✅ **Jerry-Xin's 🟡 #1 (GetThread 404 vs 500)** — closed by ExistThread split. DB failures now surface as 500.
- ✅ **Jerry-Xin's 🟡 #2 (i18n test coverage)** — closed by new test case in `api_i18n_test.go:94`.

### Carried (informational)
- **yujiawei's 🟡 #1 (INSTANT migration claim)** — unaddressed in this commit. Author can either fix the SQL comment (downgrade to "INPLACE; LOCK=DEFAULT") or restructure as two separate ALTERs (first INSTANT for columns, then INPLACE for index). The latter only matters on a very large table where lock budget is tight; for current `incoming_webhook` size, INPLACE is fine. Documenting the actual algorithm in the comment would be a one-line fix.
- **yujiawei's 🟡 #2 (thread-validation race)** — author already acknowledged at create-handler comment ("create is cold path, push only reads row"). Trade-off accepted.
- **mochashanyao's 2 🟡** (thread-delete cascade + admin UI filter) — author's own `api.go:1672` follow-up note covers item 1; item 2 is frontend work.

### Praise (carried, still earned)
- Targeted commit shape: 1 commit, 2 files, +19/-4 to address exactly the prior round's 2 🟡 — no scope creep, no peripheral edits. Clean follow-up surface.
- The `ExistThread` → `GetThread` two-call split is the right design: separates "does this exist at all" (where DB transient errors matter) from "is this thread active" (where the residual window is tiny). Author named the rationale in the new comment.
- `api_i18n_test.go` test addition pins the responder against the i18n key — exactly Jerry-Xin's ask, no over-engineering. Matches the existing test pattern in that file.

### Verdict-side gate (informational)
`reviewDecision` may still show CHANGES_REQUESTED from stale prior-round CRs until Jerry-Xin / mochashanyao / Steve / yujiawei re-review at `1a8f1469`. Same convergence pattern PR #442 / #445 / #450 went through.

Jerry-Xin (dismissed):
Re-review on the new head `1a8f1469f898` (delta since `e90d4e58`). The delta is small and directly addresses the two follow-ups from the prior round; all thread-delivery security invariants still hold. APPROVE.

**What changed since `e90d4e58`**
- `modules/incomingwebhook/api.go` (+18/-4): create-time thread binding now does a two-step check — `threadService.ExistThread(groupNo, shortID)` first to separate **DB/service failure → 500 (`mgmtQueryFailed`, ops-alertable)** from **genuinely-not-exists/cross-group → 404**, then `GetThread(...)` to confirm `Active`. This fixes the prior 🟡 where a DB error collapsed into a misleading 404.
- `modules/incomingwebhook/api_i18n_test.go` (+1): adds the `mgmtThreadNotFound` → `err.server.incomingwebhook.mgmt_thread_not_found` (404) pin — the i18n test that was previously requested.

**💬 Non-blocking**
- 🟡 Migration comment is misleading (matches reviewer's 🟡A). `modules/incomingwebhook/sql/20260624000001_incomingwebhook_thread.sql:11` claims "表末尾 ADD COLUMN 走 INSTANT 算法、瞬时无锁". But the **same `ALTER` statement also creates a secondary index** (`idx_incoming_webhook_thread (group_no, thread_short_id, status)`). `ADD INDEX` cannot run under `ALGORITHM=INSTANT`; combining column-add + index-add forces MySQL to pick one algorithm for the whole statement, degrading to `INPLACE` (online, brief metadata lock — **not** instant/lockless). The statement also does not explicitly pin `ALGORITHM`/`LOCK`, so it relies on the optimizer's choice. On a large `incoming_webhook` table an op trusting "瞬时无锁" could be surprised. Recommend either (a) split the index into its own `ALTER ... ADD INDEX ALGORITHM=INPLACE, LOCK=NONE`, or (b) reword the comment to state the combined statement is INPLACE/online (not INSTANT). Doc-accuracy/ops-risk, not a code-correctness bug.
- 🟡 Create-time thread validation has a small TOCTOU window (matches reviewer's 🟡B), and the impact is genuinely bounded. Between `GetThread` (active check) and `insertWithQuota` the thread could be deleted/archived, leaving a binding row pointing at a dead thread. Impact is benign: `targetChannel()` then resolves to a non-existent/archived thread channel and delivery fails gracefully. **No cross-group/cross-thread binding is reachable** — both `ExistThread`/`GetThread` enforce the `group_no AND short_id` double-condition (`modules/thread/db.go:508`), and `short_id` is bound to this group's auth context, so the race cannot bind to another group's thread. Acceptable race (🟡), not 🔴.
- 🔵 The thread-scoped bot management route gets server-side coverage, but the added tests focus on the user-facing route; a bot-route create/list/scope test would reduce regression risk for the promised bot face.

**✅ Highlights (security invariants re-confirmed at new head)**
- `targetChannel()` (`modules/incomingwebhook/api.go:1514`) is unchanged: push trusts only the persisted row via the double-condition `ChannelType==CommunityTopic && ThreadShortID != ""`, else falls back to `GroupNo`. No request-supplied channel override on the push path.
- Cross-group/cross-thread injection stays blocked: `GetThread`/`ExistThread` query by `group_no AND short_id` (`modules/thread/db.go:508`); cross-group is naturally not-found.
- Binding is immutable: `channel_type`/`thread_short_id` are written once at create and excluded from the update-allowed set.
- Mention/AI expansion stays anchored to the parent group; only the final send target switches to the thread channel.
- Management operations remain scope-isolated to the parent `group_no`; member-gate anchored to parent group membership.
- Backward-compat byte-identical: `channel_type` defaults `2`, `thread_short_id` defaults `''` → existing rows resolve to "deliver to parent group", and create always writes the type explicitly (DEFAULT only as backfill).

OctoBoooot (approved):
## Review: feat(incomingwebhook): deliver to thread (子区) channels (#454) — delta @ ef860416

**Verdict:** Approve.

Delta scope from prior APPROVE @ `1a8f1469`: 1 commit `ef860416` "fix(incomingwebhook): correct thread migration DDL algorithm; add thread quota/cascade/bot-face tests" — 3 files, +103/-10. **CI green.**

**Both carry-over 🟡 items closed in a single focused commit.**

### Item 1 — Migration DDL algorithm corrected (yujiawei + Jerry-Xin + my own retract converged)

SQL migration **restructured + comment rewritten + explicit pin** — the textbook fix, not a doc-only patch:

- **Two separate ALTERs**:
  1. `ALTER TABLE ADD COLUMN channel_type + ADD COLUMN thread_short_id` (no INDEX) — now genuinely INSTANT on MySQL 8.0
  2. `ALTER TABLE ADD INDEX idx_incoming_webhook_thread (...), ALGORITHM=INPLACE, LOCK=NONE` — explicit pin
- **New comment names the WHY**: "ADD INDEX 不支持 INSTANT，若与 ADD COLUMN 合在一条 ALTER 里，MySQL 取最严算法、整条退化为 INPLACE，把列的 INSTANT 也一并拖慢"
- **Explicit ALGORITHM/LOCK pin** lets non-online-DDL environments fail-loud rather than silent-degrade — cross-references `modules/oidc/sql/20260515000001_oidc_bind_uniques.sql` as precedent
- **Down migration symmetric** — DROP INDEX first then DROP COLUMNS (correct order, indexed columns can't be dropped first)

This closes my own retracted praise from the prior round (the "INSTANT-eligible" claim that was byte-false). The fix is exactly what the [praise-has-weight] rule update demanded — author chose the structural fix over the doc-only one, plus added the explicit pin for fail-loud behavior.

### Item 2 — Tests added: bot-face thread route + thread quota/cascade invariants (Jerry-Xin)

- **`bot_api/incoming_webhook_test.go::TestBotWebhook_ThreadScopedCreate`** — exercises bot-face `/v1/bot/groups/:G/threads/:S/incoming-webhooks` POST + GET; asserts `creator_uid == botID`, `thread_short_id` echoed, list-on-thread-face shows only that thread's webhook (scope isolation on bot face).
- **`db_test.go::TestThreadWebhookSharesParentGroupQuota`** — pins the design choice my prior round praised. Asserts thread + group webhooks share the parent-group quota pool: `insertWithQuota` counts ALL group_no webhooks regardless of thread_short_id, so a third webhook (thread OR group) past `max=2` hits `ErrQuotaExceeded`.
- **`db_test.go::TestDisableByGroupNo_CoversThreadWebhook`** — pins another design invariant. Asserts the disband cascade (`disableByGroupNo` filters only on `group_no`) flips BOTH the group-direct webhook AND the thread-bound webhook to `statusDisabled`.

These two DB tests convert two "documented contract" guarantees from my prior round's praise into **CI-enforced invariants**. Exactly the pattern reviewers have been pushing across this whole session — `payloadContentKey` const (#450), `assemblePushPayload` wire-seam (#451), now thread quota/cascade invariants. Test pinning the invariant beats comment pinning the invariant.

### Adversarial verification on the new code

- **Two-ALTER atomicity** — if statement 1 succeeds and statement 2 fails, DB has columns but no index. Index is for query performance not correctness; can be retried/manually added. If statement 1 fails, statement 2 never runs. Acceptable failure mode.
- **`LOCK=NONE` env requirement** — MySQL 8.0+ with online DDL support. Comment explicitly notes the fail-loud behavior on environments that don't support it (correct posture for ops — better to error early than to silently downgrade to `LOCK=SHARED` or `LOCK=EXCLUSIVE`).
- **Down migration order** — DROP INDEX first then DROP COLUMNS. Correct: MySQL can't drop a column that's part of an active index.
- **`TestThreadWebhookSharesParentGroupQuota` adversarial** — test inserts `max=2` (1 group + 1 thread) then asserts BOTH a third group and a third thread hit `ErrQuotaExceeded`. Both directions verified — caller can't escape quota by switching their next attempt's binding type.
- **`TestDisableByGroupNo_CoversThreadWebhook` adversarial** — test asserts the thread webhook is also disabled, not just the group one. Prevents future "let's filter by `thread_short_id` for some reason" regression from silently breaking the cascade.

### Praise

- **Structural fix over doc-only patch** for the migration — author could have just edited the comment to say "INPLACE" but chose to restructure into two ALTERs + explicit pin so the migration's behavior matches its semantic intent (column add is INSTANT, index add is INPLACE-LOCK=NONE). The cross-reference to the existing `oidc_bind_uniques.sql` precedent shows author looked for an existing pattern in the codebase before improvising.
- **`ALGORITHM=INPLACE, LOCK=NONE` explicit pin** — the most important operational property of this migration. Without the pin, an environment that doesn't support online DDL would silently degrade to a locking algorithm; with the pin, the migration fails loud. This is exactly the "documented contract → fail-loud invariant" upgrade pattern, applied at the DDL layer.
- **Two new DB invariant tests** convert two "design choice" properties from documented contracts into CI-enforced behaviors. `TestThreadWebhookSharesParentGroupQuota` and `TestDisableByGroupNo_CoversThreadWebhook` each name the design intent in the test name + comment. Future maintainer who "optimizes" `insertWithQuota` to per-thread or filters `disableByGroupNo` by `thread_short_id` fails the test loudly with a comment explaining why.
- **Bot-face thread route test** asserts the scope-isolation invariant (`list` returns only this thread's webhook) on the bot-mount face — closes the gap Jerry-Xin flagged. Doesn't over-engineer (single test covers create + list + scope, doesn't re-test the underlying handler logic already covered on user-face tests).

### Verdict-side gate (informational)
`reviewDecision = REVIEW_REQUIRED` — other reviewers' APPROVEs at 1a8f1469 likely auto-dismissed by this new commit. They'll need to re-review at ef860416.

Jerry-Xin (approved):
Re-review (delta) at head `ef860416edce`. Prior APPROVE was at `1a8f1469f898`. The delta since then is comment + test only (no logic change): `modules/incomingwebhook/sql/20260624000001_incomingwebhook_thread.sql` (+14 −10), `modules/bot_api/incoming_webhook_test.go` (+25), `modules/incomingwebhook/db_test.go` (+64). Both prior open 🟡 are now closed and the security invariants still hold. APPROVE.

✅ Resolved from prior round

🟡#1 migration comment — CLOSED. The single `ALTER` that combined `ADD COLUMN` + `ADD INDEX` (which forces the whole statement to INPLACE, not INSTANT) is now split into two statements: `ADD COLUMN channel_type / thread_short_id` alone (end-of-table, DEFAULTed → genuinely INSTANT-eligible) and a separate `ADD INDEX idx_incoming_webhook_thread (group_no, thread_short_id, status), ALGORITHM=INPLACE, LOCK=NONE`. The comment no longer claims INSTANT for a combined statement; it correctly explains the split and pins the index algorithm for fail-loud behavior. The Down migration is split symmetrically. Claim now matches reality.

🟡#2 bot-route thread test — CLOSED. `TestBotWebhook_ThreadScopedCreate` in `modules/bot_api/incoming_webhook_test.go` drives the real bot thread route `/v1/bot/groups/:group_no/threads/:short_id/incoming-webhooks` through the mounted `s.GetRoute()` handler with a real `Bearer` bot token, so the bot-uid-context → resolveActor wiring on the thread route is exercised. It asserts more than HTTP 200: `creator_uid` equals the bot id (identity resolved from token) and `thread_short_id` is echoed (binding written), and the thread-face list returns exactly 1 entry — proving scope isolation, not a leak across group/other threads. `db_test.go` adds `TestThreadWebhookSharesParentGroupQuota` (thread + group webhooks share the parent-group quota pool) and `TestDisableByGroupNo_CoversThreadWebhook` (disband cascade also disables thread-bound webhooks), reinforcing parent-group anchoring at the DB layer.

✅ Security invariants (unchanged at this head)

The 5 parent-group-anchored gates, the single `targetChannel()` delivery resolver, the immutable create-time binding, and the cross-group/cross-thread injection block (double-condition `ExistThread(group_no, short_id)`) are all unchanged by this delta and still hold. Backward compatibility for existing group webhooks is preserved by the column DEFAULTs.

💬 Non-blocking

🔵 The create handler in `modules/incomingwebhook/api.go` already separates a transient existence-check failure (500 via `mgmtQueryFailed`) from genuine missing/cross-group (404 via `mgmtThreadNotFound`); the residual `GetThread` active-status confirmation collapses its narrow error window to 404, which is a reasonable documented choice. No change required.

🔵 `modules/incomingwebhook/api_thread_test.go` push test asserts the 200 path; `TestTargetChannel` covers the resolver. An integration assertion on the actual `SendMessage` channel_id/channel_type would further protect the handler wiring, but this is optional.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#454 Review Report
**Head SHA:** `ef860416edced76e353aae23a3b247c87c9ca2c7`
**PR:** feat(incomingwebhook): deliver to thread (子区) channels
**Scope:** security_sensitive — automated review review

---

## 1. 验证结论

### Migration (`20260624000001_incomingwebhook_thread.sql`)
- ✅ Split into two ALTERs: column-add (INSTANT on MySQL 8.0) + separate index-add pinned `ALGORITHM=INPLACE, LOCK=NONE`. Correctly avoids the silent INPLACE degradation that a combined ALTER would cause. Comment explains the rationale.
- ✅ Defaults (`channel_type=2`, `thread_short_id=''`) preserve byte-for-byte backward compat for all existing group webhooks.
- ✅ Down migration correctly mirrors: `DROP INDEX` then `DROP COLUMN`.

### Model (`model.go`)
- ✅ `incomingWebhookModel` gains `ChannelType int` and `ThreadShortID string` with clear inline docs.
- ✅ `webhookResp` echoes `ThreadShortID` with `json:"thread_short_id,omitempty"` — group webhooks omit it.
- ✅ `updateFieldsAllowed` (`db.go:146-153`) does NOT include `channel_type` or `thread_short_id` → binding is immutable post-creation. Verified via `updateFields()` reject at `db.go:~170`.

### Delivery Target (`targetChannel()`, `api.go:~1516-1522`)
- ✅ Single source of truth for `(channelID, channelType)`. Group → `(GroupNo, ChannelTypeGroup)`; Thread → `(BuildChannelID(GroupNo, ThreadShortID), ChannelTypeCommunityTopic)`.
- ✅ Defensive default: any non-`(CommunityTopic + non-empty short_id)` combination falls back to group. **Impossible to deliver to a half-formed thread channel.**
- ✅ Uses `thread.BuildChannelID` (verified: `fmt.Sprintf("%s____%s", groupNo, shortID)` at `thread/service.go:830`). Wire format `group_no____short_id` pinned in `TestTargetChannel`.

### Delivery Sites (`handlePush`, `testPush`)
- ✅ `handlePush` (`api.go:~1460`): calls `m.targetChannel()` → `SendMessageWithResult(channelID, channelType)`.
- ✅ `testPush` (`api.go:~1131`): same pattern. Both sites migrated from hard-coded `(m.GroupNo, ChannelTypeGroup)` to row-derived target.
- ✅ All upstream gates unchanged: token compare → `cachedRequireActiveGroup(m.GroupNo)` → rate-limit → `cachedCreatorMembership(m.GroupNo, m.CreatorUID)` → adapter parse → payload build → deliver. Thread binding affects ONLY the final `SendMessage` call.

### Scope Isolation
- ✅ `queryManageable` (`api.go:612`): adds `m.ThreadShortID != c.Param("short_id")` check. On group routes, `c.Param("short_id")` returns `""`, matching only group webhooks (ThreadShortID=""). On thread routes, matches only that thread's webhooks. Cross-scope → 404.
- ✅ `queryByGroupNo` (`db.go`): adds `WHERE thread_short_id=''` → group list excludes thread webhooks.
- ✅ `queryByThreadScope` (`db.go`): new symmetric function with `WHERE thread_short_id=?`. Both hit `idx_incoming_webhook_thread(group_no, thread_short_id, status)` index prefix.
- ✅ `list` handler (`api.go:~790`): branches on `c.Param("short_id")` to pick the right query function.
- ✅ All 7 management routes (create, list, update, delete, regenerate, deliveries, testPush) registered via `MountManagementRoutes` on both group and thread groups. Same handlers, scope determined by route params.

### Thread Validation at Create (`api.go:~678-711`)
- ✅ `IsValidShortID(shortID)` — digit-only, 15-20 chars. Rejects garbage.
- ✅ `ExistThread(groupNo, shortID)` first — separates DB failure (→ 500 `mgmtQueryFailed`) from genuinely missing (→ 404). SQL: `SELECT count(*) FROM thread WHERE group_no=? AND short_id=? AND status!=3`.
- ✅ `GetThread(groupNo, shortID, "")` second — confirms thread is active (`tr.Status == ThreadStatusActive` i.e. `1`). Archived (2) and deleted (3) are rejected.
- ✅ Cross-group threads: `ExistThread` + `GetThread` both query by `(groupNo, shortID)` → cross-group naturally returns false/nil → 404.

### @Mention / @AI Expansion (`assemblePushPayload`, `api.go:~1205`)
- ✅ `ExpandAisToBotUIDs` called with hard-coded `ChannelTypeGroup.Uint8()` and `m.GroupNo` — NOT `targetChannel()`. This is **correct by design**: thread bots are parent-group bots; `ExpandAisToBotUIDs` is GROUP-only (`expand_ais.go:116` returns payload unchanged for non-group types).
- ✅ `filterGroupMembers` keys on `groupNo` (parent group) for direct @uid filtering. Consistent with the "all gates anchored on parent group_no" invariant.

### Quota & Cascade
- ✅ `insertWithQuota` (`db.go:68-75`): counts by `group_no` only (no `thread_short_id` filter). Thread webhooks share the parent group's `max_per_group` pool. Verified by `TestThreadWebhookSharesParentGroupQuota`.
- ✅ `disableByGroupNo` (`db.go:294-300`): filters by `group_no` only. Group disband cascades to thread webhooks. Verified by `TestDisableByGroupNo_CoversThreadWebhook`.

### Bot Face (`bot_api/incoming_webhook.go`)
- ✅ Thread-scoped mount at `/v1/bot/groups/:group_no/threads/:short_id/incoming-webhooks` using the same `MountManagementRoutes`. Same middleware chain (`authBot`, `botActorUID`, rate limiter). Same permission matrix.
- ✅ Verified by `TestBotWebhook_ThreadScopedCreate`.

### Error Codes & i18n
- ✅ `ErrIncomingWebhookThreadNotFound` (404) registered in `pkg/errcode/incomingwebhook.go` with proper i18n keys.
- ✅ `mgmtThreadNotFound` responder pinned in `TestIncomingWebhookRespondHelpers`.
- ✅ zh-CN translation present in `active.zh-CN.toml`.

### Test Coverage
- ✅ `TestTargetChannel` — pure unit, pins channel mapping (4 cases incl. defensive fallbacks + parity with `thread.BuildChannelID`).
- ✅ `api_thread_test.go` — happy path, reject missing/invalid/archived/cross-group threads, scope isolation (list + cross-scope delete).
- ✅ `TestThread_Push_Delivers` — e2e push to real thread channel, gated on thread-capable WuKongIM (CI v2.2.4).
- ✅ `db_test.go` — shared parent-group quota, group-disband cascade coverage.
- ✅ `bot_api/incoming_webhook_test.go` — bot-face thread scoped create + list isolation.

---

## 2. 发现的问题（P0/P1/P2）

### P2-1: TOCTOU window between `ExistThread` and `GetThread` — minor
- **Diff-scope:** new (introduced by this PR).
- **Description:** `create` calls `ExistThread` then `GetThread` in sequence. If the thread is deleted between the two calls, `GetThread` returns an error (`"thread has been deleted"` or a DB error). The handler treats any `GetThread` error as 404 (`mgmtThreadNotFound`). For a genuine DB failure (not a race), this would mask a 5xx as a 404.
- **Severity rationale:** The race window is sub-millisecond. A DB failure on the second query while the first succeeded is astronomically rare. The consequence is a misleading 404 instead of 500 on a vanishingly rare edge — not a user-visible systematic error. P2, not P1.
- **Suggestion:** Consider checking `terr` from `GetThread` separately: if `tr == nil && terr != nil` and the error message indicates a DB failure (not "not found"/"deleted"), return `mgmtQueryFailed` (500). Low priority.

### P2-2: No active cascade-disable on thread delete/archive (documented v1 gap)
- **Diff-scope:** new (introduced by this PR's scope decision).
- **Description:** Thread delete/archive does not cascade-disable bound webhooks in v1. Pushes to a deleted thread channel fail at WuKongIM with a 502 (`push_delivery_failed`). The TODO at `api.go:~1683-1689` documents this.
- **Severity rationale:** This is a locked design decision with maintainer. The 502 fallback is functional (push caller gets an error). Not a silent failure. P2 — should be addressed in a follow-up but not blocking.

### P2-3: `TestThread_Push_Delivers` skips on non-thread-capable WuKongIM
- **Diff-scope:** new.
- **Description:** The e2e push delivery test is gated by `requireThreadCapableIM` (probes `GET /health` on WuKongIM). On environments without v2.2.4, the test skips. This is necessary for local dev but means CI is the only place this test actually runs.
- **Severity rationale:** Not a defect — a pragmatic test design choice. The unit test `TestTargetChannel` pins the channel mapping independently. P2 note only.

---

## 3. 建议

1. **P2-1 (TOCTOU):** Optionally differentiate `GetThread` error types post-`ExistThread` to avoid masking DB failures as 404. Very low urgency.
2. **P2-2 (thread cascade):** Add a thread-delete event listener in a follow-up to cascade-disable bound webhooks, symmetric with `handleGroupDisband`. The TODO is already in place.
3. **Nit:** Consider adding a `deliveries` scope-isolation test (asserting thread webhook deliveries return 404 on group path) for completeness. The `queryManageable` check already covers it, but an explicit assertion would lock in the contract.

---

## 4. 额外发现

- **Cache coherence for thread webhooks:** `cachedQueryByWebhookID` uses `SELECT *` which returns `ChannelType` and `ThreadShortID`. The cache naturally covers thread-bound webhooks. Since binding is immutable (not in `updateFieldsAllowed`), no cache invalidation concern.
- **Index coverage:** `idx_incoming_webhook_thread(group_no, thread_short_id, status)` covers both `queryByGroupNo` (prefix `(group_no, thread_short_id)`) and `queryByThreadScope` (full prefix). The existing `insertWithQuota` count-by-group_no query may not use this index (lacks `thread_short_id` filter) but `incoming_webhook` table is small — no performance concern.
- **Migration safety for rollbacks:** The down migration cleanly drops the index and columns. If a rollback is needed, all thread-bound webhooks would lose their binding data — but since `channel_type` and `thread_short_id` are the only new columns, this is expected and clean.

---

## 5. 数据流回溯

| Consumed Data | Upstream Source | Does Data Reach Consumer? |
|---|---|---|
| `targetChannel()` → `ChannelType` | `incoming_webhook.channel_type` (DB, set at INSERT) | ✅ `queryByWebhookID` uses `SELECT *`; cache stores full model |
| `targetChannel()` → `ThreadShortID` | `incoming_webhook.thread_short_id` (DB, set at INSERT) | ✅ Same as above |
| `targetChannel()` → `BuildChannelID` | `thread.BuildChannelID(groupNo, shortID)` → `groupNo + "____" + shortID` | ✅ Verified at `thread/service.go:830` |
| `handlePush` → `m.GroupNo` for auth gates | `cachedRequireActiveGroup(m.GroupNo)`, `cachedCreatorMembership(m.GroupNo, m.CreatorUID)` | ✅ Both key on parent `group_no`, unchanged |
| `assemblePushPayload` → `ExpandAisToBotUIDs` | Hard-coded `ChannelTypeGroup.Uint8()` + `m.GroupNo` | ✅ Intentional: thread bots = parent group bots. `ExpandAisToBotUIDs` is GROUP-only |
| `filterGroupMembers` → mention targets | `group_member` table filtered by `group_no` (parent) | ✅ Thread members derived from parent group |
| `insertWithQuota` → count | `SELECT count(*) WHERE group_no=? AND status != deleted` (no `thread_short_id` filter) | ✅ Shared parent-group quota pool |
| `disableByGroupNo` → cascade | `UPDATE SET status=disabled WHERE group_no=? AND status != deleted` | ✅ Covers thread webhooks (same `group_no`) |
| `queryManageable` → scope check | `m.ThreadShortID != c.Param("short_id")` | ✅ Group routes: `c.Param("short_id")=""`; thread routes: actual short_id |
| Create `channelType` | Path `:short_id` presence → validated via `ExistThread` + `GetThread` | ✅ Double-condition (group_no + short_id) prevents cross-group binding |

---

## 6. 盲点 checklist（R5 — security_sensitive PR, all items mandatory）

### C1 双路径 parity（含守卫/lint 工具自身等价写法覆盖）
- **User-face vs bot-face routes:** Both mount the same `MountManagementRoutes` on thread-scoped groups. Parity confirmed: `/v1/groups/:group_no/threads/:short_id/incoming-webhooks` (user) ↔ `/v1/bot/groups/:group_no/threads/:short_id/incoming-webhooks` (bot). Same handlers, same middleware chain.
- **Create vs management (update/delete/regenerate):** Create sets `channel_type`/`thread_short_id`; `updateFieldsAllowed` excludes them → immutable binding. Parity confirmed.
- **queryByGroupNo ↔ queryByThreadScope:** Symmetric SQL, mutually exclusive scope. Both filter `status != statusDeleted`. Parity confirmed.
- **Group disband cascade ↔ thread delete cascade:** Asymmetric by design (group disband cascades; thread delete does not in v1). Documented with TODO. P2-2 above.
- **Clear.**

### C2 control-flow ordering / 嵌套复用 + 安全控件非 canonical 试穿
- **`targetChannel()` ordering:** Single call-site pattern — called once at delivery time, after all auth/rate-limit gates. No double-invocation or nesting risk.
- **`short_id` validation:** `IsValidShortID` rejects non-digit chars (no unicode, no hyphens, no spaces). Length 15-20 digits. Non-canonical inputs (leading zeros, very short, very long) all rejected. ✅
- **`assemblePushPayload` / `ExpandAisToBotUIDs`:** Called once per push, after payload is built. No re-entrant or nested expansion. `ExpandAisToBotUIDs` is idempotent (deduplicates against existing `mention.uids`).
- **Clear.**

### C3 授权边界 ≠ 能力边界（谁能到达 tool/能力）
- **Push endpoint** (`/v1/incoming-webhooks/:webhook_id/:token`): Unauthenticated, token-in-URL. Delivery target derived **solely from persisted row**, NOT from request body. An attacker with a valid token can only push to the pre-bound channel — cannot redirect to arbitrary threads. ✅
- **Management endpoints:** Behind `AuthMiddleware` (user face) or `authBot` (bot face). Thread-scoped create validates thread ownership via `GetThread(groupNo, shortID)`. No path to bind a webhook to a cross-group thread.
- **Clear.**

### C4 授权生命周期 / 容器-成员状态级联
- **Parent group status:** All push-path auth gates key on `m.GroupNo` (parent group). `cachedRequireActiveGroup` checks `g.Status == GroupStatusNormal`. If the parent group is disbanded, `handleGroupDisband` → `disableByGroupNo` cascades to thread webhooks. ✅
- **Thread status:** Only validated at create time. Post-creation thread archive/delete does not actively disable the webhook (v1 design decision). Pushes to dead channels fail with 502. Documented as P2-2.
- **Creator membership:** `cachedCreatorMembership(m.GroupNo, m.CreatorUID)` checks parent-group membership. Creator leaving parent group → lazy-disable (webhook flipped to `statusDisabled`). Covers thread webhooks since `group_no` is the same.
- **Clear.** (Thread-lifecycle cascade is P2-2, not a blocker.)

### C5 build/note 通过 ≠ 运行期路径正确
- **No build artifacts / extensions / CLI / relative URLs** in this PR. Pure server-side Go code. `CGO_ENABLED=0 go build ./...` is claimed green in the PR description.
- **N/A** for this change type.

### C6 治理/策略/安全文档自洽性
- **Documentation changes:** README.md and thread.md updated with thread webhook docs. No conflict with existing security policies or disclosure practices.
- **N/A** (no security policy or governance document changes).

---

## 7. 跨轮 blocker 复检（R6 — re-review 必填；首次 review 写 N/A）

---

## Summary

This is a well-structured, security-conscious PR. The design correctly separates delivery-target routing (row-derived, immutable) from authorization/membership/quota (parent-group-anchored). Scope isolation between group and thread management faces is enforced at both the handler and query layers. Migration DDL is properly split for online safety. Test coverage is comprehensive, covering unit (channel mapping), integration (create/reject/scope), and e2e (real push delivery) layers.

No P0 or P1 findings. Three P2 notes (TOCTOU masking, thread-lifecycle cascade gap, CI-gated push test) — all non-blocking.

[Octo-Q] verdict: APPROVE + three P2 notes (non-blocking)

yujiawei (approved):
# Code Review — PR #454 (octo-server)

Reviewed at head SHA `ef860416edced76e353aae23a3b247c87c9ca2c7` against `main` (merge-base `e0db32c9`). Verified locally: `CGO_ENABLED=0 go build ./...` for the touched packages, `go vet ./modules/incomingwebhook/`, `make i18n-extract-check`, `make i18n-lint`, and the pure unit `TestTargetChannel` all pass. Two independent automated review passes were run and their findings are folded in below.

## Verdict: Approve

No correctness, security, data-loss, or build-breaking issues. The trust boundary holds, scope isolation is enforced uniformly, and the change stays inside the agreed scope. The items below are advisory (P2) — none block merge.

## What was verified

- **Trust boundary (push ingress).** The delivery target is derived solely from the persisted row via `targetChannel()` (`api.go:1514`); the push URL/query/body never participate. `handlePush` (`api.go:1442`) and `testPush` (`api.go:1135`) both resolve `(channelID, channelType)` from the row. `targetChannel()` defaults defensively to the parent group for any combination that is not exactly `(CommunityTopic && thread_short_id != "")`, so a half-written row can never deliver to a malformed thread channel. Confirmed by the defensive cases in `thread_target_test.go`.
- **Scope isolation.** `queryManageable` (`api.go:612`) adds `m.ThreadShortID != c.Param("short_id")` on top of the existing cross-group guard, and `list` (`api.go:474`) routes to `queryByThreadScope` vs `queryByGroupNo`. All write endpoints (update/delete/regenerate/deliveries/test) funnel through `queryManageable`, so a group webhook cannot be managed via a thread path and vice versa. `TestThread_ScopeIsolation` covers both directions.
- **All gates stay parent-group-anchored.** Auth/membership (`resolveActor` → `queryMemberRole`), creator-membership (`cachedCreatorMembership` on the push path, `api.go:1313`), quota (`insertWithQuota` keys on `group_no` — `TestThreadWebhookSharesParentGroupQuota`), disband cascade (`disableByGroupNo` covers thread rows — `TestDisableByGroupNo_CoversThreadWebhook`), and @AI expansion (`assemblePushPayload` calls `ExpandAisToBotUIDs` with `ChannelTypeGroup` + `m.GroupNo`, `api.go:1203`) all key on the parent group. Only the final `SendMessage` channel changes.
- **Create validation.** Binding requires a valid snowflake `short_id`, an existing thread under this `group_no`, and an active thread. `ExistThread`/`GetThread` both query `(group_no, short_id)`, so cross-group binds return 404 (`TestThread_Create_RejectThreadOfOtherGroup`). Channel-type constants match the migration (Group=2, CommunityTopic=5 in octo-lib `common/constant.go`).
- **Backward compatibility.** New columns default to group semantics; existing rows keep delivering to `(group_no, ChannelTypeGroup)`. The guard test file list is correctly unchanged (no new non-test handler files), and the new error code carries both en-US and zh-CN markers.

## Advisory findings (P2 — non-blocking)

1. **`GetThread` error masked as 404 — fix for consistency** (`api.go:700`). Both automated passes flagged this independently. After `ExistThread` succeeds, a transient DB/service error in the follow-up `GetThread` is folded into `mgmtThreadNotFound` (404). This contradicts the deliberate `ExistThread` handling three lines above, which correctly returns 500 on `terr != nil`. A real infrastructure failure here becomes a "not found", which breaks monitor alerting and client retry semantics. Suggest returning `mgmtQueryFailed` (5xx) when `terr != nil`, and reserving 404 for `tr == nil || tr.Status != Active`.

2. **Migration column ALTER does not pin `ALGORITHM=INSTANT`** (`sql/20260624000001_incomingwebhook_thread.sql:18`). The comment states the column add is intentionally INSTANT, and the index ALTER correctly pins `ALGORITHM=INPLACE, LOCK=NONE`. The column ALTER itself is unpinned, so if MySQL cannot use INSTANT it degrades silently instead of failing loud. Either pin `ALGORITHM=INSTANT` to match the comment and fail loud, or soften the comment. Low impact given the table is small.

3. **No constraint ties `channel_type` to `thread_short_id`** (`api.go:612` filter vs `api.go:1515` delivery). Scope isolation discriminates on `thread_short_id` while delivery treats `channel_type` as authoritative. No current code path can desync them — `create` writes both atomically and neither is in `updateFieldsAllowed`, so this is defensive-only. If you want belt-and-suspenders against a future writer or a manual repair, a DB `CHECK` (or aligning the isolation filter to also consider `channel_type`) would make the invariant self-enforcing.

## Risk to surface for human review (security-sensitive PR)

The `needs-human-review` label is appropriate. The one design consequence worth a human nod (not a defect — it is an explicitly documented v1 limitation in the linked task brief, "Out of scope" + locked lifecycle decision, with a `TODO(thread-cascade)` already in `handleGroupDisband`): because thread webhooks share the parent group's `max_per_group` and a thread delete/archive does **not** cascade-disable its webhooks in v1, a workflow that repeatedly creates threads + binds webhooks + deletes those threads will slowly consume the parent group's webhook quota with orphaned rows until an admin cleans them up. Delivery to a deleted thread fails with 502 (the intended fallback), so there is no correctness or data-loss issue — only quota accounting. Confirm this is acceptable for v1, or consider whether a thread-delete event listener should be the next phase.

## Coverage notes

- This review read the diff plus the cross-file call sites it depends on (`thread.ExistThread`/`GetThread`/`BuildChannelID`/`IsValidShortID`, the archived-thread auto-reactivate path, the deleted-thread WuKongIM ban path, `updateFieldsAllowed`, all `SendMessage` sites, and the router mounts for wildcard conflicts). What was **not** exercised: the live MySQL `ALTER` behavior, the WuKongIM CommunityTopic delivery against v2.2.4, and the gated `TestThread_Push_Delivers` e2e (skips without a thread-capable IM). These are covered by CI, not by this static review.

Issue #448, opened 2026-06-24
feat(incomingwebhook): visible @mention pills — broadcast text auto-compose + directed-@ name consistency (v2)


an9xyz:
Follow-up to #445 (incoming-webhook mention v1). v1 puts a well-formed `mention.{uids,humans,ais}` on the wire — ping / red-dot / bot-summon are all confirmed working. v1 explicitly deferred the *visible* layer (brief: "pills are a later phase", "no name resolution"). Field verification surfaced two gaps worth a product decision:

1. **Directed-@ label/uid mismatch.** The client renders the `@` pill from the literal `content` text, decoupled from `mention.uids`. A webhook can display one name in the pill while actually pinging a different member — e.g. `content: "deploy done @alice"` with `mention.uids: ["<bob's uid>"]` shows an `@alice` pill but the red-dot goes to Bob. Decide:
   - (a) document caller responsibility (caller must write the correct `@name` text; backend does not validate it against `uids`), or
   - (b) backend resolves uid→display name and composes/validates the `@name` text, optionally adding `mention.names` / `entities`.
2. **Broadcast ergonomics.** `mention.all` / `mention.bots: true` fires the notify/bot but renders no pill unless the caller also writes `@所有人` / `@所有AI` into `content`. Option: backend auto-prepends the fixed broadcast strings when the capability-gated flag is set.

**Scope note:** directed-@ name resolution should reuse the space-member display-name path (see the space empty-name fallback work) and must settle the `entities` offset unit (rune vs byte) against the client before emitting offsets. Any change to the wire `content` keeps the native-only + whitelisted-field constraints from #445.

an9xyz:
**Triage: traced the visible-pill render path on all three clients (web / iOS / Android).** TL;DR — the `entities`/offset question in the scope note resolves to *"don't emit offsets"*; both gaps reduce to composing the right **text** in `content` on top of the `mention.{uids,humans,ais}` that v1 (#445) already sends.

### How each client turns a wire mention into a visible pill (receive side)

| | Reads wire `mention.entities`? | Binds `@text` → uid by | Offset unit (when used) | Broadcast pill |
|---|---|---|---|---|
| **web** | yes; else falls back | entities: offset span; **fallback: positional** (Nth `@`-token → `uids[N]`) | UTF-16 code units (`String.substring`/`.length`) | synthesized from `humans`/`ais` flag, but the label must then be found **literally in `content`** |
[...]

PR #458, opened 2026-06-25
fix(bot_api): propagate App Bot token revocation across replicas via shared Redis cache

an9xyz:
## Summary

App Bot API auth resolved a presented token to `{UID, Scope, SpaceID}` through a **per-process in-memory map** (`bot_api.AppBotRegistryAdapter`). Revocations (rotate / unpublish / delete) only mutated the registry of the instance that handled the admin request, and on the auth path an in-memory hit short-circuited the authoritative DB check — so under multiple replicas a revoked App Bot token kept authenticating on every peer until that peer restarted (security-relevant staleness).

This replaces the in-memory registry with a **shared Redis write-through cache**, so a revoke is visible to every replica the instant the admin request commits. The DB (`app_bot` table) stays the authority via the existing fallback, and auth **fails safe** to the DB when Redis is unavailable. The bearer token is **SHA-256 hashed** in the Redis key (never stored verbatim). Revocation writes a **shared tombstone** (a positive cross-replica deny), and the best-effort cache warm-up uses **SETNX** so a delayed warm-up can never resurrect a just-revoked token; the warm-up is also **bounded by a degraded-Redis circuit breaker** so an outage can't amplify into goroutine / connection-pool pressure on the auth hot path.

## Related Issue

Fixes #309

## Linked Spec

`.octospec/tasks/appbot-token-revocation-redis/brief.md` (load-bearing: `auth`, `bot-api`, `space`/`isolation`). Journal: `.octospec/journal/shared/appbot-token-revocation-redis.md`.

## Changes

- New `modules/bot_api/registry_redis.go` — `RedisAppBotRegistry` over a shared Redis store (`appbot:auth:{sha256hex(token)}` → JSON spec; the token is **SHA-256 hashed** so a live bearer credential never lands in a Redis key). The write model is deliberately asymmetric:
  - `FindByToken` returns nil on a miss, on a **tombstone**, or on **any Redis error** → `authAppBot` falls through to the authoritative DB lookup (fail-safe; never serves a stale/revoked spec on a degraded backend).
  - `Add` (authoritative publish / rotate-new) — unconditional `SET`, overwrites any tombstone so a re-publish re-enables the token at once.
  - `Warm` (best-effort warm-up: DB-fallback repopulate + startup load) — `SETNX`, only fills an **absent** key, so it can never overwrite a concurrent revocation's tombstone; circuit-gated so a Redis outage can't pile up blocking writes on the auth hot path.
[...]

PR #457, opened 2026-06-24
feat(messages_search): RichText search + virtual sub-doc consumer + narrow /_search payload types

BrDing-Rookie:
## Background

`/_search` (the message-search endpoint) previously returned media messages (image / file / voice / video) as well, all labeled `message_kind: "text"` with mostly no readable snippet, so the client's "message" tab could not render them correctly. RichText messages (payload.type=14) were also not keyword-searchable. This PR systematically fixes RichText search and tightens the `/_search` contract.

## What changed (4 commits)

1. **`feat: support RichText keyword search`** — keyword recall on `/_search` and `/_search_all` is extended to `payload.richText.searchText`, so RichText message bodies become searchable.
2. **`feat: consume virtual richtext sub-docs` (Part B)** — consume the virtual sub-documents the indexer emits per image/file embedded inside a RichText message (`virtual=true` + `parentMessageId`):
   - `source.go`: add reader-internal fields `Doc.ParentMessageID` / `Doc.Virtual` (never surfaced in the response).
   - Text-search endpoints (`/_search`, `/_search_all`, `/_search_around`) now all apply `applyExcludeVirtual` (`must_not(virtual=true)`), so a virtual sub-doc never double-emits alongside its parent RichText message.
   - Visibility key switches to `coalesce(parentMessageId, messageId)`: a virtual sub-doc has no MySQL row of its own — revoke / delete / channel-offset / visibles state all belong to the parent message row. Includes `nil/0` guards and fail-closed degradation.
   - `/_search_media` and `/_search_files` do **not** exclude virtual docs — embedded media surfaces through those dedicated endpoints (response `message_id` is projected to the parent value, zero front-end change).
3. **`feat: narrow /_search to text+mergeForward+richText`** — add a `payload.type ∈ [1, 11, 14]` whitelist (text / mergeForward / richText) to `/_search`, and align the keyword multi_match fields to `text.content^3` / `richText.searchText^3` / `mergeForward.msgs.searchText`, removing the now-unreachable image/file scoring fields and highlight fields.
4. **`docs(swagger): align _search description`** — bring the swagger description in line with the `[1,11,14]` whitelist.

## Behavior change (front-end note)

`/_search` no longer returns image / video / voice / file messages:

- ✅ Media → `/_search_media`, files → `/_search_files`, everything → `/_search_all` (covers `[1,8,11,14]`).
- ✅ The "message" tab keeps merge-forward (type=11) search.
- ℹ️ No wire-shape change — existing contracts are not broken.

[...]

PR #465, opened 2026-06-25
feat(incomingwebhook): configure @mention on the webhook (create/update) instead of the push body

### LINKED ISSUE #462: feat(incomingwebhook): configure @mention on the webhook (create/update) instead of the push body
## Problem

Incoming-webhook `@mention` (shipped in #445 / #448 / #449) is driven entirely by the **push request body**: the caller sends `mention: {uids, all, bots, render, entities}` on every push. Two problems with that model:

1. **Any holder of the push token decides who gets @'d.** The push token is a low-trust shared secret written into CI / platform config; letting the request body pick `@uid` / `@所有人` targets means a leaked token can ping arbitrary members on every call. Who a webhook @'s should be an operator decision made when the webhook is wired, not a per-call caller input.
2. **Mention only works on the `native` endpoint.** Because the targets come from the body and platform-event bodies (github / wecom / gitlab / feishu / multica) carry no octo member-UID semantics, `pushAdapter.allowMention` is `true` only for native (see the comment in `adapter.go`). Webhooks pushed through an adapter can't @ anyone today.

## Proposal

Move the @ target list from the push body to **webhook create/update configuration**. The push endpoint stops reading `mention` from the body (silently ignored); every push from a webhook @'s whatever the webhook is configured with.

- New column `incoming_webhook.mention_uids` (JSON array of octo UIDs; user & bot share one namespace). `create` / `update` accept + validate it (must be **current** group members, deduped, ≤ cap → otherwise `400 reason=mention_uids`); `webhookResp` surfaces it.
- The existing per-webhook broadcast switches `allow_mention_all` / `allow_mention_bots` now mean "this webhook @所有人 / @所有 AI on **every** push"; the `member_can_broadcast` policy gate (instant revoke for member-created webhooks) is unchanged.
- Directed targets are re-filtered to current members at push time (a target who left the group is dropped) and rendered as `@name` pills (server resolves `user.name`, generates `mention.entities`).

Because the targets now live on the webhook row (server-side UIDs) instead of the body, the native-only blocker disappears: **mention applies to every push shape — native + all platform adapters**. That parity win is the main reason to do this beyond the trust-boundary fix.

## Decisions

- Push body `mention` is **ignored** (not an error) — backward compatible; the body `mention` field and the caller-supplied `mention.entities` machinery are removed.
- Directed `@uid` is gated by the group-member filter + cap only (not the broadcast switches); broadcast is gated by the switches **and** `member_can_broadcast`.
- Validation is belt-and-suspenders: config-time rejects non-members / over-cap, push-time re-filters to current members (handles later departures).
- `mention_uids` is `VARCHAR(4096) NOT NULL DEFAULT ''` (not `TEXT`) to avoid the MySQL 8.0.13 expression-default requirement and the NULL→`string` scan problem; the 50-uid × ≤40-char cap fits with ~2× margin.

## Out of scope

- Per-push caller selection of mention (explicitly removed).
- Name / email → UID resolution; cross-group / cross-Space @.
- Caller-supplied `mention.entities` (server generates the pills).
- Inline pill placement (still prepended at the content head).
- Group-level mention policy (`allow_no_mention` / `bot_mention_pref`).

## Acceptance

- `create` / `update` persist + echo `mention_uids`; non-member or over-cap → `400 reason=mention_uids`.
- A configured webhook's push (no body mention) delivers `mention.uids` = current members + generated `@name` pills; `allow_mention_*` add `mention.humans` / `mention.ais` subject to `member_can_broadcast` (denied → reported in the 200 body's `mention_ignored`).
- A push **body** containing `mention` has no effect.
- The **same** configured webhook pushed via a platform adapter (e.g. wecom) also carries the configured mention; github `ping` / skip events deliver nothing and @ nobody.
- A configured target who left the group is dropped at push.
- `go test ./modules/incomingwebhook/...` (incl. `TestIncomingWebhookNoLegacyResponseError`), `make i18n-extract-check`, `make i18n-lint`, `golangci-lint run ./modules/incomingwebhook/...` all green.

## Spec / Implementation

Brief: `.octospec/tasks/incoming-webhook-mention-config/brief.md`. Implemented on branch `feat/incomingwebhook-mention-config` (not yet PR'd). Follow-up to #445 (mention v1) / #448 (visible pills) / #449 (caller entities).

an9xyz:
## Summary

Moves incoming-webhook `@mention` from a **per-push body param** to **webhook create/update configuration**. The push endpoint stops reading `mention` from the request body (silently ignored); every push from a webhook now @s whatever the webhook is configured with.

This closes a trust-boundary gap (the low-trust push token could pick `@uid`/`@所有人` targets on every call) and, as a bonus, **extends mention to every adapter endpoint** — not just `native`.

Closes #462. Follow-up to #445 (mention v1) / #448 (visible pills) / #449 (caller entities).

## What changed

- **New column `incoming_webhook.mention_uids`** (`VARCHAR(4096) NOT NULL DEFAULT ''`, JSON array of octo UIDs; user & bot share one namespace; empty = no @). `create`/`update` accept + validate it (dedup, ≤ cap, must be **current** group members → otherwise `400 reason=mention_uids`); `webhookResp` always surfaces it as an array.
- **`buildMention` now sources everything from the webhook row** (`mention_uids` + the `allow_mention_*` switches); the request only supplies the body text/msg_type. Directed targets render as `@name` pills (server resolves `user.name`, generates `mention.entities`) and are re-filtered to current members at push time, so a target who left the group is dropped automatically.
- **Broadcast switches re-pointed:** `allow_mention_all` / `allow_mention_bots` now mean "this webhook @所有人 / @所有 AI on **every** push" (was a per-request gate). The `member_can_broadcast` policy gate (instant revoke for member-created webhooks) is unchanged.
- **Adapter parity flipped in the safe direction:** removed the native-only `pushAdapter.allowMention` gate, so `handlePush` builds mention for **every** adapter (native + github/wecom/gitlab/feishu/multica). This was previously impossible because body-sourced targets carry no octo UID semantics on platform events; config-sourced targets remove the blocker.
- **Dead caller-supplied machinery removed:** `pushPayloadReq.Mention`, `mentionReq`, `decodeMention`, and the caller-`entities` path (`decodeEntities`/`finalizeEntities`/`entityUIDsOf`/`rangeClaimed`/`markClaimed`/`shiftEntityOffsets`) plus their tests. Pure helpers (`composeMentionContent`, `assembleMention`, `dedupNonEmpty`) are untouched and still unit-tested.

## Back-compat / migration notes

- **Push body `mention` is ignored, not an error** — existing callers keep working; their `mention` is just dropped.
- The migration **zeroes pre-existing `allow_mention_all`/`allow_mention_bots` values** and updates the column COMMENTs, because the switch semantics changed from "may broadcast when asked" to "broadcast on every push" — operators re-opt-in under the new, documented meaning rather than silently start all-hands-spamming on upgrade.
- `mention_uids` is `VARCHAR(4096)` (not `TEXT`) to avoid the MySQL 8.0.13 expression-default requirement and the NULL→`string` scan problem; the 50-uid × ≤40-char cap fits with ~2× margin. End-of-table `ADD COLUMN` is INSTANT on MySQL 8.0.

## Test plan

- `go test ./modules/incomingwebhook/...` — green (incl. `TestIncomingWebhookNoLegacyResponseError` and new config-driven / adapter-parity / body-ignored e2e tests: persist+echo, non-member reject, broadcast policy across adapters, body-mention-ignored, test-push applies configured mention).
- `make i18n-extract-check`, `make i18n-lint`, `golangci-lint run ./modules/incomingwebhook/...`, `go build ./...` — all clean.
- Test push renders the directed `@name` pill + routes to configured members, but **suppresses** `@所有人/@所有 AI` so repeated test clicks can't all-hands-spam the group.

## Out of scope

Per-push caller selection of mention (removed); name/email→UID resolution; cross-group/cross-Space @; caller-supplied `mention.entities`; inline pill placement; group-level mention policy.

https://claude.ai/code/session_01TRbvs9bNv4Y3b3rrdzfGjC

---
_Generated by [Claude Code](https://claude.ai/code/session_01TRbvs9bNv4Y3b3rrdzfGjC)_


OctoBoooot (commented):
## Review: feat(incomingwebhook): configure @mention on the webhook (create/update) instead of the push body (#465)

**Verdict:** Approve.

CI `check-sprint / check-sprint` is FAILURE — same process-gate pattern as the recent PRs in this repo (linked issue #462 is CLOSED; gate likely needs a Sprint field on the linked issue). **Not a code defect.** I'm posting APPROVE on the code rather than RC because the only outstanding item is administrative metadata — but per skill rule §11, this PR cannot be merged until check-sprint clears.

Risk tier: normal. 16 files, 1658 lines (large-change label applied). NET reduction (+738/-920) because the refactor removes a substantial amount of caller-supplied mention machinery.

### Byte-verified — load-bearing claims hold

Delegated structural verification to a focused subagent + independently byte-checked the migration. All 11 PR-description claims byte-verify.

- ✅ **Migration safety** (`sql/20260625000001_incomingwebhook_mention_uids.sql`) — `ADD COLUMN mention_uids VARCHAR(4096) NOT NULL DEFAULT ''` at table end → MySQL 8.0 INSTANT, no lock. Default `''` backfills existing rows cleanly. Cap rationale documented (50 UIDs × 40 chars ≈ 2.2KB, well under VARCHAR(4096) ceiling with DYNAMIC row format margin).
- ✅ **Migration handles the semantic flip on `allow_mention_all`/`allow_mention_bots` correctly** — old semantics were "allow per-request"; new semantics are "broadcast every push". A naive ADD-COLUMN would silently flip existing webhooks with `allow_mention_all=1` to start spamming `@所有人` every push. The migration explicitly:
  ```sql
  UPDATE incoming_webhook SET allow_mention_all = 0, allow_mention_bots = 0
    WHERE allow_mention_all <> 0 OR allow_mention_bots <> 0;
  ```
  Forces ops to re-enable under new semantics rather than silent retroactive behavior change. The COMMENT updates on both columns document the new contract in-schema (DBA reading metadata sees the right thing). Down migration restores the COMMENT + drops the column, with explicit note that the UPDATE is one-way (can't un-zero historical values).
- ✅ **`pushPayloadReq.Mention` removed** (`model.go:119-122`) — silently ignored on body unmarshal (unknown JSON field) for backward compat. Native adapter `parseNative` unmarshals into the new struct shape; clients still sending `{"mention": {...}}` get a 200 with the field dropped, no breaking change.
- ✅ **`buildMention` sources from DB row** (`mention.go`) — `m.MentionUids` is the sole source; no `req.Mention` reference. Re-filter happens via `w.db.filterGroupMembers` BEFORE entity offsets are computed via `composeMentionContent`, so dropped UIDs can't shift offsets wrong.
- ✅ **Adapter parity flipped in safe direction** (`adapter.go`) — `allowMention` per-adapter gate removed. `handlePush` builds mention for every adapter (native + github + wecom + gitlab + feishu + multica) unconditionally; the previous "native-only" gate existed because body-sourced targets carried no UID semantics on platform events — config-sourced targets remove that blocker structurally.
- ✅ **Dead-code removal complete** — `pushPayloadReq.Mention`, `mentionReq`, `decodeMention`, `decodeEntities`/`finalizeEntities`/`entityUIDsOf`/`rangeClaimed`/`markClaimed`/`shiftEntityOffsets` all removed along with their tests. No orphaned references.
- ✅ **Validation completeness** on create/update (`api.go:437-472` `validateMentionUIDs`) — `dedupNonEmpty` (TrimSpace + drop blanks + dedup preserve-order), cap check at 50, member-existence check via `filterGroupMembers` → 400 with `reason=mention_uids`. Edge cases verified: whitespace-only/empty/duplicates/cap+1/non-members all behave correctly.
- ✅ **`member_can_broadcast` gate preserved** (`api.go:1511-1514`) — `broadcastPermitted := w.settings.IncomingWebhookMemberCanBroadcast() || creatorIsAdmin`. Test `TestMentionPush_BroadcastDeniedByPolicyAcrossAdapters` pins instant-revoke: setting=0 → next push returns 200 with `mention_ignored=["all","bots"]`, no `@所有人` reaches IM. Member-created webhook + setting flip = full revocation.
- ✅ **`webhookResp.mention_uids` is always `[]`** (`api.go:426-430`) — `parseMentionUIDs` + nil → `[]string{}` so the response never returns `null`. Empty config → `[]`.

### Praise — security-positive refactor done right

- **The trust-boundary fix is the load-bearing point of this PR.** Previously, the low-trust push token could pick `@uid` / `@所有人` targets per call — anyone with the webhook URL could spam mentions at any group member. After this PR, mention is config-driven (admin-set on create/update with member-existence validation), and the `member_can_broadcast` policy is the instant-revoke for member-created webhooks. This is exactly the right direction: move trust decisions from per-call ingress to admin-config + revocable policy.
- **Migration zeros the semantic-flip values rather than silently changing behavior.** The old `allow_mention_all=1` meant "allow per-request"; the new same value means "broadcast every push". A naive migration would have made every existing webhook with this flag start spamming `@所有人` on every push the moment the new code ships. Author explicitly `UPDATE`s those rows to 0 with a comment explaining why ("升级时把存量值归零，让运维在新语义下重新显式开启——避免静默的追溯性行为变更"). This is the textbook safe-migration discipline for a semantic flip.
- **Adapter parity flipped in the safe direction** — the prior `allowMention` gate was native-only because body-sourced UIDs couldn't be meaningfully interpreted on platform-event payloads (github/wecom/gitlab/feishu/multica all wrap third-party events). Now that mention is config-driven, the gate is structurally unnecessary, and removing it extends the feature to all adapters with zero new attack surface. Subagent verified no platform-specific adapter has its own body-mention path that would now silently drop user-supplied targets.
- **Validation is layered: write-time AND read-time.** Create/update validates "must be current group members" (rejects 400 if not); push-time re-filters via `filterGroupMembers` so a target who LEAVES the group between config-time and push-time is auto-dropped without breaking entity offsets. Defense-in-depth where neither layer is load-bearing alone — and the offset-computation ordering (`filterGroupMembers` BEFORE `composeMentionContent`) is correct so dropped UIDs can't desynchronize the entity ranges.
- **Test coverage spans the security-relevant directions, not just happy paths**: `TestMentionPush_BodyMentionIgnored` (negative-direction: body-supplied mention silently dropped), `TestMentionPush_BroadcastDeniedByPolicyAcrossAdapters` (policy gate revokes across all adapters), `TestMentionPush_NoMentionFieldBackwardCompatible` (existing webhooks with no config = no @, no break). The negative-direction tests are exactly what would catch a future regression that re-introduced body-sourced mention.

- CI `check-sprint` red — `Closes #462` is in the body and the issue is in CLOSED state. Likely the gate requires the linked issue to either be OPEN with a Sprint field set OR requires a Sprint field on the PR itself. Pattern matches #457/#458/#461/#463 — administrative metadata, not a code change.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#465 Review Report
**PR:** Mininglamp-OSS/octo-server #465
**Head SHA:** `2c49be7a4dd2dcd25bd037329ff89a700852b74a`
**Title:** feat(incomingwebhook): configure @mention on the webhook (create/update) instead of the push body
**Scope:** security_sensitive (automated review)

---

## 1. Summary

This PR moves the incoming-webhook `@mention` target list from a **caller-supplied push body param** to **webhook create/update configuration**:

- New DB column `mention_uids` (VARCHAR(4096), JSON array string) on `incoming_webhook`.
- `create`/`update` accept `mention_uids` with validation (dedup, ≤50 cap, all current group members).
- `buildMention` now sources targets entirely from the webhook model (`MentionUids` + `AllowMention*` switches); `pushPayloadReq.Mention` is removed.
- Native-only `pushAdapter.allowMention` gate deleted → mention applies to **all** adapters (native + github/wecom/gitlab/feishu/Octo-Q).
- Broadcast semantics changed: `allow_mention_*` now means "broadcast on every push" (was "permit per-push request"). Migration zeroes existing values to prevent retroactive behavior change.
- Caller-supplied entity machinery (`decodeMention`, `decodeEntities`, `finalizeEntities`, `shiftEntityOffsets`, etc.) fully removed.
- Test push now runs mention assembly (with broadcast suppressed) to verify configured `@`.

| Item | Status | Evidence |
|------|--------|----------|
| MentionUids field → `mention_uids` column mapping | ✅ | `model.go:59` field `MentionUids` with `db:"mention_uids"` tag; `UnderscoreName("MentionUids")` → `"mention_uids"` confirmed. Write path (`AttrToUnderscore`) produces correct column name; read path (`dbr Load`) matches via explicit tag. |
| Create path persists `mention_uids` | ✅ | `api.go:~770` calls `validateMentionUIDs` → `api.go:~800` sets `MentionUids: mentionUIDsJSON` on model → `db.go:~108` INSERT via `AttrToUnderscore` includes it. |
| Update path persists `mention_uids` | ✅ | `api.go:~964` adds `fields["mention_uids"]`; `db.go:153` `updateFieldsAllowed` includes `"mention_uids"`. Pointer type `*[]string` correctly distinguishes nil (no change) from `[]` (clear). |
| Push path reads config, ignores body mention | ✅ | `mention.go:237` `parseMentionUIDs(m.MentionUids)` reads from model. `pushPayloadReq` no longer has `Mention` field → `json.Unmarshal` silently ignores body `mention`. |
| Member gate at write + push time | ✅ | `validateMentionUIDs` (`api.go:444`) calls `filterGroupMembers` at config time. `buildMention` (`mention.go:261`) calls it again at push time. Both filter: `is_deleted=0`, `is_external=0`, `status=Normal`. |
| Migration safety | ✅ | `sql/20260625000001`: ADD COLUMN with `NOT NULL DEFAULT ''` (INSTANT on MySQL 8.0). Zeroes existing `allow_mention_*` to prevent retroactive broadcast semantic change. Column COMMENTs updated. |
| Adapter parity | ✅ | `adapter.go:53` `allowMention` field removed. `api.go:~1510` `assemblePushPayload` runs for all adapters. Skip events (github ping) return before assembly. |
| Test push renders directed @ | ✅ | `api.go:~1205` `payload, _ = w.assemblePushPayload(m, req, payload, false)`. Broadcast suppressed (`broadcastPermitted=false`); directed `@` renders normally. |
| Backward compatibility (no config, no body mention) | ✅ | `parseMentionUIDs("")` → nil; `wantAll/wantBots` both false → `buildMention` returns nil → no mention key in payload. |
| Authorization | ✅ | `requireOwnership` (`api.go:~586`) gates updates to admin or creator. Create: any group member. Same as `allow_mention_*`. |

## 3. Findings

### P2 — Render-always-on for configured targets (behavioral change, documented)

**Diff-scope:** new (PR introduces this change)

In the old model, directed `@` rendering was opt-in via `mention.render=true` in the push body. Now it's default-on for all configured `mention_uids` targets (`mention.go:282` always passes `true` to `composeMentionContent` for the render param).

**Impact:** Webhooks whose old callers sent `uids` without `render:true` (for red-dot-only / bot-wakeup use cases) will now always prepend `@<name>` pills to text content. This is a visible behavioral change for existing integrations that only wanted routing without content modification.

**Mitigation:** Journal documents this as deliberate ("matches the PRD intent"). The push body no longer carries mention at all, so there's no caller opt-in mechanism. If this is acceptable per product requirements, no action needed. If some integrations need routing-only behavior, a per-webhook `render` switch could be added later.

**Severity:** P2 — documented design decision; may affect existing integrations that relied on routing-only behavior.

### P2 — VARCHAR(4096) safety depends on `maxMentionUIDs` staying ≤ ~90

**Diff-scope:** new (new column)

`VARCHAR(4096)` fits the default 50-UID cap with ~2× margin (50×(40+3)+2 ≈ 2152 chars). However, the cap is env-configurable via `OCTO_INCOMINGWEBHOOK_MAX_MENTION_UIDS`. If an operator raises it above ~90, the JSON could exceed VARCHAR(4096) and the INSERT would silently truncate or fail.

**Mitigation:** Journal documents: "raising `OCTO_INCOMINGWEBHOOK_MAX_MENTION_UIDS` far past the default is the only way to overflow it — documented bound."

**Severity:** P2 — operational constraint, safe at defaults, documented.

## 4. No P0/P1 Issues Found

All critical paths verified:
- Data flow from `mention_uids` column through `parseMentionUIDs` → `dedupNonEmpty` → `filterGroupMembers` → `assembleMention` → `CloneForExpansion` → `ExpandAisToBotUIDs` → wire is correct and complete.
- Security boundary properly moved from push-time (attacker-controlled body) to config-time (operator-controlled, member-gated) + push-time re-filter.
- Migration prevents retroactive broadcast semantic change.
- Authorization on create/update matches existing `allow_mention_*` patterns.
- Wire contract (`mention.{uids,humans,ais,entities}`) unchanged — only the source changed.

## 5. Data-Flow Trace

| Consumer | Source | Path | Verified |
|----------|--------|------|----------|
| `buildMention` directed uids | `incomingWebhookModel.MentionUids` | `parseMentionUIDs` → `dedupNonEmpty` → `filterGroupMembers` → `assembleMention` | ✅ |
| `buildMention` broadcast flags | `incomingWebhookModel.AllowMentionAll/Bots` | `== 1` → `&& broadcastPermitted` | ✅ |
| `composeMentionContent` render names | `filterGroupMembers` return value | `membersByName[uid]` → `namesByUID` | ✅ |
| `assemblePushPayload` wire payload | `buildMention` return | `CloneForExpansion` → `ExpandAisToBotUIDs` → returned payload | ✅ |
| `validateMentionUIDs` member check | `filterGroupMembers(groupNo, uids)` | SQL: `is_deleted=0, is_external=0, status=Normal` | ✅ |
| `toResp` mention_uids response | `parseMentionUIDs(m.MentionUids)` | nil → `[]string{}` for always-array JSON | ✅ |
| `testPush` mention assembly | Same `assemblePushPayload` as `handlePush` | `broadcastPermitted=false` suppresses broadcast | ✅ |

- **C1 — Dual-path parity:** N/A. No symmetric add/remove pair. Create/update both call `validateMentionUIDs` symmetrically. ✅
- **C2 — Control-flow ordering / nested reuse:** Clear. `buildMention` → `assemblePushPayload` is called once per push. `composeMentionContent` is a pure function with no nesting hazard. Broadcast labels guarded by `isBroadcastLikeName` + `stripFormatRunes` (anti-confusable, unchanged). ✅
- **C3 — Authorization boundary ≠ capability boundary:** Clear. Mention targets confined to current group members via `filterGroupMembers` at both write and push time. No cross-group/cross-Space reachability. ✅
- **C4 — Authorization lifecycle / container-member cascade:** Clear. Member gate re-checks at push time; departed/banned members dropped automatically. Webhook's group-active check runs earlier in `handlePush`. ✅
- **C5 — Build ≠ runtime:** N/A. Server-side Go, no build/runtime divergence. ✅
- **C6 — Governance/policy document consistency:** Clear. README updated consistently. No conflict with existing SECURITY.md or disclosure policies. ✅

## 7. Cross-Round Blocker Re-Check (R6)

N/A — first review.

## 8. Suggestions

1. Consider adding a test that verifies body `mention` is ignored even when the webhook has configured `mention_uids` (i.e., body `mention.uids` doesn't override or merge with config). The current `TestMentionPush_BodyMentionIgnored` tests a webhook with NO config; a complementary test with config + body mention would lock the non-interaction property.
2. Consider adding a runtime guard or log warning if `maxMentionUIDs()` would produce JSON exceeding the VARCHAR(4096) capacity.

---

[Octo-Q] verdict: **APPROVE** — no P0/P1 issues found. Two P2 observations (render-always-on behavioral change, VARCHAR capacity at non-default env). PR is well-designed with proper migration safety, authorization, and backward compatibility.

yujiawei (dismissed):
# Code Review — PR #465 (octo-server)

Reviewed at head SHA `2c49be7a4dd2dcd25bd037329ff89a700852b74a` against `main` (merge-base `1797c405`). Security-sensitive (trust boundary / space isolation / migration). Build, `go vet`, and the pure unit tests for the module pass locally; the DB-backed e2e suite was not run here (needs MySQL/Redis/WuKongIM).

## 1. Spec / contract compliance — ✅

The diff implements exactly what the linked issue asks and nothing beyond it.

- **Missing:** none. `create`/`update` accept, validate, and persist `mention_uids` (dedup + cap + current-member gate, non-member/over-cap → `400 reason=mention_uids`); `webhookResp` always surfaces it as an array; push sources mention entirely from the webhook row; broadcast switches re-pointed to per-push semantics; the native-only `allowMention` gate is removed so mention runs for every adapter; dead caller-supplied machinery (`pushPayloadReq.Mention`, `mentionReq`, `decodeMention`, `decodeEntities`/`finalizeEntities`/`entityUIDsOf`/`rangeClaimed`/`markClaimed`/`shiftEntityOffsets`) and its tests are gone.
- **Extra (out of scope):** none. The wire contract `mention.{uids,humans,ais,entities}` consumed by `pkg/mentionrewrite` + `message/api_reminders.go` is untouched; no new error code (reuses `ErrIncomingWebhookRequestInvalid`, `reason` already in `SafeDetailKeys`); no new i18n entries; no changes outside the module.
- **Deviation:** none. Directed render is now default-on for configured uids (server resolves `user.name` → `@name` pill), matching the brief's stated parity decision.

Verified the trust-boundary and isolation invariants directly:
- Body `mention` is structurally unreachable — the field is removed from `pushPayloadReq`, so `json.Unmarshal` drops it (`model.go`).
- `mention_uids` is gated by `filterGroupMembers` (`is_deleted=0 AND is_external=0 AND status=Normal`) at **both** write time (`validateMentionUIDs`, api.go:62) and push time (`buildMention`, mention.go:253) — a config target who left the group is dropped automatically, and a token cannot @ a cross-group/cross-Space identity.
- The management validation gate is not an enumeration oracle: `resolveActor` confirms the caller is a group member/admin before `validateMentionUIDs` runs; non-members are silently dropped at push (anti-enumeration), consistent with the existing member gate.

## 2. Code quality — Approved

No P0/P1 findings. A cross-model adversarial pass (line-level) surfaced two **advisory** items, both of which I confirmed are **pre-existing on `main`, not introduced by this PR**:

- **P2 (pre-existing, out of diff):** `assemblePushPayload` expands `mention.ais` with `ChannelTypeGroup` + `m.GroupNo` even for thread-bound webhooks whose delivery target is a `CommunityTopic` (api.go:1278). The `ExpandAisToBotUIDs(...m.GroupNo...)` call dates to #445 and is unchanged here; bot UIDs are still parent-group members, so it does not cross a Space boundary. Worth a follow-up to pass the resolved `channelID/channelType` into expansion, but it is not a regression from this change.
- **P2 (pre-existing, out of diff):** the Feishu adapter (`adapter_feishu.go`, **not in this PR**) can turn attacker-controlled `at.user_name` into literal `@所有人`/`@all` text. That path does not set `mention.humans`, so it cannot forge a real broadcast; at most a client may render a broadcast-looking pill from literal text. Independent of this PR.

Things this PR gets right that are easy to get wrong:
- Broadcast-forgery defense on server-rendered nicknames is solid: `isBroadcastLikeName` NFKC-folds + strips Cf format runes + lowercases before comparing, and also rejects names that merely *start* with a broadcast label followed by a non-word boundary (iOS `@\S+` tokenization). Nicknames containing `@` are also skipped. So a configured directed @ can't be used to fake `@所有人`/`@所有AI`.
- UTF-16 entity offsets are computed consistently (`utf16Len`), matching web/Android/iOS units; the rune budget guard prevents pill-prefix overflow.
- `update` correctly uses `*[]string` to distinguish "not provided" (nil, untouched) from "explicit clear" (`[]` → empty string = no @).
- Directed @uid intentionally does **not** pass through `broadcastPermitted`; only `humans`/`ais` broadcast bits do — matches the documented gate.
- The migration zeroes pre-existing `allow_mention_*` (semantics changed from per-request to per-push), updates column COMMENTs, and uses `VARCHAR(4096) NOT NULL DEFAULT ''` (avoids the MySQL 8.0.13 expression-default + NULL-scan issues); end-of-table `ADD COLUMN` is INSTANT on 8.0. The zeroing is one-way but low risk: `allow_mention_*` was itself added two days earlier (migration `20260623...`, also unreleased), so there is effectively no production data to lose.

## 3. Nits (non-blocking)

- `db.go:223` has a stale doc comment referencing `finalizeEntities`/`mention.render`, both removed in this PR. Cosmetic; update when convenient.

## 4. Notes for a human verifier

- Confirm the DB-backed e2e suite (`go test ./modules/incomingwebhook/...`) is green in CI — it was not executed in this review environment.
- The PR currently reports `MERGEABLE: CONFLICTING`; the only conflicting path is `.octospec/log.md` (append-only changelog from a sibling PR), not code. A rebase resolves it.

## Verdict

Spec ✅ AND Quality Approved → **APPROVED**. The two advisory items are pre-existing and tracked as follow-ups, not blockers for this change.

Jerry-Xin (dismissed):
Project relevance gate: pass. The PR is in-scope for `octo-server`, specifically the incoming-webhook backend module.

No blocking issues found. The trust-boundary move from push-body mention input to webhook configuration is consistent with the module’s existing authorization, group membership, and payload-whitelist patterns.

💬 Non-blocking

🟡 Warning — `OCTO_INCOMINGWEBHOOK_MAX_MENTION_UIDS` is unbounded while storage is fixed at `VARCHAR(4096)`.
`modules/incomingwebhook/mention.go:39` lets operators raise the mention UID cap arbitrarily, but `modules/incomingwebhook/sql/20260625000001_incomingwebhook_mention_uids.sql:15` fixes the column at 4096 bytes. If the env cap is raised high enough, `validateMentionUIDs` can accept the list and then insert/update can fail at DB write time instead of returning a clean 400. Consider clamping the env-derived cap or rejecting marshaled `mention_uids` values over the column size in `modules/incomingwebhook/api.go:464`.

🔵 Suggestion — Add an update-path rejection test for non-member `mention_uids`.
`modules/incomingwebhook/mention_e2e_test.go:94` covers create rejection, and update uses the same helper at `modules/incomingwebhook/api.go:960`, so risk is low. A small update test would lock that contract explicitly.

✅ Highlights

The PR correctly removes caller-controlled mention parsing from push payloads, validates configured targets at the management boundary, and re-filters group members at push time.

The migration’s reset of existing broadcast switches avoids silently changing old “may broadcast” rows into “broadcast every push” rows.

Focused non-infra tests passed:
`go test ./modules/incomingwebhook -run 'TestAssembleMention|TestParseMentionUIDs|TestDedupNonEmpty|TestMaxMentionUIDs|TestIsTextMention|TestComposeMentionContentDirected|TestBuildMentionBroadcastCompose'`

Full `go test ./modules/incomingwebhook/...` could not complete locally because the shared test database already contains an unknown migration from another module; this appears to be environment state, not a PR failure.

OctoBoooot (approved):
## Review: feat(incomingwebhook): configure @mention on the webhook (create/update) instead of the push body (#465) — delta @ f30df926

**Verdict:** Approve.

Delta from prior COMMENT @ `2c49be7a`: 1 new commit `f30df926` "fix(incomingwebhook): integrate PR #465 review findings" plus rebase that pulled in #458 and #457 merges. CI now **green** — `check-sprint` passes. The substantive new commit is incomingwebhook-only and addresses Jerry-Xin's 2 🟡 + adds an extra negative-direction test.

### Resolved 🟡 from prior round

- ✅ **Jerry-Xin #1: env-cap vs DB-column overflow** — new constant `mentionUIDsColumnChars = 4096` in `mention.go:39-43` documents the DB column width and ties it to the validation. `api.go:471-477` `validateMentionUIDs` adds post-marshal length check: if marshaled JSON exceeds 4096 bytes → clean `400 reason=mention_uids` (`mgmtRequestInvalid(c, "mention_uids")`). Comment names the threat ("OCTO_INCOMINGWEBHOOK_MAX_MENTION_UIDS 可被运维调高到 JSON 超列宽") AND the maintenance invariant ("改列宽须同步此常量") so future column resize doesn't drift. Adversarial: env_cap=500 × 40-char UIDs JSON-encoded ≈ 22KB > 4096 → 400 with `reason=mention_uids` rather than dirty DB-write or silent truncation.
- ✅ **Jerry-Xin #2: missing UPDATE-path non-member rejection test** — `mention_e2e_test.go:325-339` `TestMention_UpdateRejectsNonMemberUID` asserts PUT with `mention_uids: ["ghost_not_member"]` returns 400. Pairs with existing `TestMention_CreateRejectsNonMemberUID` so both write paths have negative-direction pins for the member-only invariant.

### Extra hardening

- ✅ **New `TestMentionPush_BodyMentionIgnoredWhenConfigured`** (`mention_e2e_test.go:286-323`) — clever assertion: configure webhook with `mention_uids=[A]` + `member_can_broadcast=0` (broadcast policy DENIED at the system level), then POST push with `{"mention": {"all": true, "bots": true, "uids": [B]}}`. The non-interaction property: if the body broadcast were wrongly honored, the policy-denied broadcast would surface as `mention_ignored` in the response. Asserting `mention_ignored` is ABSENT proves the body was structurally dropped (not merged-then-policy-denied). This is the [test-the-negative] pattern done right — the test would fail loudly if a future regression made body mention merge into webhook config.
- ✅ **`db.go:222-224` stale doc comment** — `filterGroupMembers` comment used to reference removed `finalizeEntities` / `mention.render`; updated to current state (`assembleMention 只看 key 判归属`).

### Praise

- **Column-width guard names the exact threat AND the cross-reference**: the `mentionUIDsColumnChars` comment in `mention.go:39-43` is the textbook shape for this kind of cross-layer constant. It names what the value mirrors (the migration's VARCHAR width), what the env-tuning threat is (OCTO_INCOMINGWEBHOOK_MAX_MENTION_UIDS far past default), and what maintenance invariant must hold ("改列宽须同步此常量"). Future maintainer who bumps the migration's column width hits this comment.
- **`TestMentionPush_BodyMentionIgnoredWhenConfigured` is a non-trivial proof of non-interaction**: a plain "body mention is ignored" test on a webhook with no config could pass even if body-merge logic existed (the merge would just produce empty config + body). This test specifically constructs the scenario where merge-vs-drop behavior would surface as a different observable response (`mention_ignored` field present vs absent). That's the test-the-difference discipline.
- **Round-2 fix attribution is granular per finding**: the commit message + journal entry explicitly map each fix to the reviewer who raised it (Octo-Q P2 + Jerry-Xin warning → column-width guard; yujiawei nit → stale comment; Octo-Q + Jerry-Xin → tests). The journal also names what was *reviewed and deliberately NOT changed* (render-always-on per PRD; two pre-existing P2s confirmed on main, not this PR). Clean review accounting.

### Carried (still standing — byte-verified in prior round)

All prior-round invariants hold:
- Migration zeros existing `allow_mention_all`/`allow_mention_bots` rows to prevent silent retroactive @所有人 spam (the load-bearing safety property for the semantic flip)
- Adapter parity flipped in safe direction (mention now config-driven across all 6 adapters; no body-mention path on any platform adapter)
- `member_can_broadcast` policy gate preserved (instant-revoke for member-created webhooks)
- Dead-code removal complete (`pushPayloadReq.Mention` + 8 caller-supplied entity functions + their tests)
- Write-time validation (dedup, cap, member-existence) + push-time re-filter (current members only, dropped UIDs never desynchronize entity offsets)
- `webhookResp.mention_uids` always returns `[]` (never `null`)
- DB write gate is `resolveActor` + `requireOwnership` (creator/admin), not just member-check

Jerry-Xin (approved):
The PR is in scope for octo-server and the implementation matches the intended trust-boundary shift: push bodies no longer control mentions, create/update validates configured targets, and push re-filters current group membership before emitting the wire payload.

💬 Non-blocking
- 🟡 Warning: Add a test for the externally visible cap/column-width validation path. `validateMentionUIDs` rejects over-cap lists (`modules/incomingwebhook/api.go:448`) and oversized marshaled JSON against the column width (`modules/incomingwebhook/api.go:474`), but current tests cover non-member rejection and persistence, not the max-count or widened-env overflow cases.
- 🔵 Suggestion: The length-guard comment assumes UIDs are ASCII (`modules/incomingwebhook/api.go:473`). That is true for this system today, but if the UID namespace ever changes, the byte-length check becomes conservative against a `VARCHAR` character limit. A short note tying this to the actual UID contract would prevent future confusion.

✅ Highlights
- The management boundary validates `mention_uids` against current internal group members before writing, and the push path re-checks membership from the stored row before routing mentions.
- Removing the body-sourced `mention` field from `pushPayloadReq` cleanly makes stale callers harmless: unknown JSON fields are ignored by `json.Unmarshal`.
- The schema migration intentionally resets old broadcast switches before changing their semantics, avoiding a noisy upgrade surprise.
- The new head adds a column-width guard (`mentionUIDsColumnChars = 4096`) so an env-raised `OCTO_INCOMINGWEBHOOK_MAX_MENTION_UIDS` can't overflow `VARCHAR(4096)` — it returns a clean 400 instead of a dirty DB write, and adds `TestMention_UpdateRejectsNonMemberUID` plus `TestMentionPush_BodyMentionIgnoredWhenConfigured`. Both prior 🟡 findings addressed, no regression to the authz/membership/anti-forgery or migration-compat properties.

yujiawei (approved):
# Code Review — PR #465 (octo-server)

Reviewed at head SHA `f30df92636bd5e82ab82add29c46a9eda880a89c` against merge-base `e3783682`. This is a security-sensitive trust-boundary change, so it was reviewed with extra care and cross-checked with an independent second pass.

## Summary

This PR moves the incoming-webhook `@mention` target from a **caller-supplied push-body param** to **webhook create/update configuration** (new `incoming_webhook.mention_uids` column), and removes the native-only adapter gate so mention now applies to every adapter. Net direction is a **security improvement**: the low-trust push token can no longer pick arbitrary `@uid`/`@所有人` targets per request — targets are now validated at the management boundary (member gate + dedup + cap) and re-filtered to current members at push time.

Local verification on the head SHA: `go build ./modules/incomingwebhook/...`, `go test ./modules/incomingwebhook/...`, `make i18n-extract-check`, and `golangci-lint run ./modules/incomingwebhook/...` all pass.

## 1. Spec / scope compliance

**Result: ✅ compliant**

All acceptance criteria from the linked issue are implemented and covered by tests:

- `create`/`update` validate + persist `mention_uids` (dedup, ≤ cap, current group members; non-member / over-cap → `400 reason=mention_uids`); `webhookResp` echoes it as an array — `api.go:439` (`validateMentionUIDs`), `api.go:415` (`toResp`), tests `TestMention_ConfigureMentionUidsPersist`, `TestMention_CreateRejectsNonMemberUID`, `TestMention_UpdateRejectsNonMemberUID`.
- Configured webhook push (no body mention) delivers `mention.uids` re-filtered to current members + generated `@name` pills — `mention.go:240` (`buildMention`), `TestMentionPush_ConfiguredUIDsSucceed`.
- Broadcast switches apply per-push subject to `broadcastPermitted`; member broadcast revoked instantly by `member_can_broadcast=0` (reported in `mention_ignored`) — `api.go:1517`, `TestMentionPush_BroadcastDeniedByPolicyAcrossAdapters`.
- Push **body** `mention` has no effect (200, silently ignored) — native parser uses plain `json.Unmarshal` with the field removed from the model, `adapter.go:92` / `model.go:140`, `TestMentionPush_BodyMentionIgnored`.
- Adapter parity: same configured webhook via wecom also carries mention — `TestMentionPush_BroadcastDeniedByPolicyAcrossAdapters` (wecom leg).
- Config target who left the group is dropped at push (re-filter) — `buildMention` re-runs `filterGroupMembers` every push.
- Empty/absent config + switches off → no `mention` key (backward compatible) — `mention.go:245`, `TestMentionPush_NoMentionFieldBackwardCompatible`.

No over-build (no fields/flags/endpoints beyond the spec) and no deviation from the agreed approach. The removed caller-supplied machinery (`decodeMention`, `mentionReq`, `decodeEntities`/`finalizeEntities` and friends) has no remaining references anywhere in the repo.

## 2. Code quality

**Result: Approved** — no blocking (P0/P1) findings. The items below are advisory (P2) only.

### P2 — `@所有 AI` broadcast expansion uses a looser member predicate than directed `@uid`

Directed `mention_uids` are gated through `filterGroupMembers`, which requires `is_deleted=0 AND is_external=0 AND status=Normal` (`db.go` `filterGroupMembers`). The `ais` broadcast expansion goes through `ExpandAisToBotUIDs` → `GetMembers` → `queryMembersWithGroupNo`, which filters only `is_deleted=0` (`modules/group/db.go:415`). The code documents this asymmetry as intentional (directed = conservative; broadcast = parity with the message/bot_api entry points), and both paths stay within the group, so this is not a cross-group/cross-Space leak. Flagging for a human to confirm the intent holds: an *external* bot member can be reached by `@所有 AI` broadcast but not by a directed `@uid`. If that asymmetry is undesired for the webhook ingress, route bot expansion through the same internal-normal predicate.

### P2 — Management-endpoint validation dedups the full input before the cap check

`validateMentionUIDs` calls `dedupNonEmpty(in, 0)` (no clamp) before comparing against `maxMentionUIDs()` (`api.go`), so a pathologically large duplicate-heavy array is fully walked/allocated before rejection. This is the **authenticated, ownership-gated** create/update endpoint (not the anonymous push path), so the abuse surface is limited to authorized group members, and there is no unbounded amplification at push time (push clamps via `dedupNonEmpty(..., maxMentionUIDs())`). Low priority, but a cheap raw-`len(in)` guard before dedup would harden it.

### P2 — Migration zeroes existing `allow_mention_*` with no value-restore on Down

The migration deliberately resets pre-existing `allow_mention_all`/`allow_mention_bots` to 0 because the switch semantics changed from "may broadcast when asked (per-request)" to "broadcast on every push", and the Down migration restores only the COMMENT, not the values (`sql/20260625000001_incomingwebhook_mention_uids.sql:21,31`). This is the correct safety call (avoids silently turning on all-hands spam on upgrade), and the COMMENT/type change is metadata-only (`SMALLINT NOT NULL DEFAULT 0` preserved → INSTANT on MySQL 8.0). Operators must be told to re-opt-in after deploy. Optional: snapshot prior values to an audit/recovery table so re-config isn't blind.

### P2 — Body `mention` is silently dropped (no deprecation signal)

Existing native callers that still send `mention` in the body get a 200 with their `mention` silently ignored (intended backward-compat). This is per spec, but consider surfacing `mention_ignored:["body"]` (or a deprecation log) so integrators discover the contract change instead of silently losing @s. Non-blocking.

### Nits

- `mentionUIDsColumnChars` byte-length guard (`len(b) > 4096`) assumes ASCII UIDs (`len(bytes) == chars`). True today (UIDs are ASCII, same as `creator_uid VARCHAR(40)`), and it fails closed, so safe — only relevant if the UID namespace ever admits multi-byte chars.
- `testPush` discards the `mention_ignored` feedback (`api.go:1212`); an admin testing a broadcast-configured webhook sees neither the broadcast nor a hint why. The anti-spam intent is sound; surfacing the hint is a minor UX improvement.

## Items worth a human's eyes (security-sensitive PR)

1. **Broadcast posture is unchanged by this PR but worth a conscious sign-off**: any legitimate group member can create a webhook and enable `allow_mention_all`/`allow_mention_bots`, and `member_can_broadcast` defaults to `true`. After this change those switches mean "broadcast on *every* push" rather than "may broadcast when the body asks". This was already member-settable and already defaulted on before this PR (on the base branch a leaked token could broadcast every push via body `mention.all`), so the diff does not widen it — but the kill-switch (`member_can_broadcast=0`, instant, no migration) is the operational control and operators should know it exists.
2. Confirm the directed-vs-broadcast member-predicate asymmetry (P2 above) is the intended trust model for the webhook ingress.

## Coverage notes

- Verified statically + by building/running the module's unit + e2e tests at the head SHA. Did **not** exercise a live MySQL migration or downstream WuKongIM/client rendering of the wire `mention.{uids,humans,ais,entities}` — the wire contract is unchanged by this PR and the existing render/expand unit tests cover the shape, but end-to-end client pill rendering was not run here.
- Independent second-pass coverage: the cross-family advisory leg's coverage note was truncated by its provider (could not fully assess the `incomingWebhookModel` struct `db:"mention_uids"` tag); I verified that tag and its read-path round-trip directly (`model.go:57`, `TestMention_ConfigureMentionUidsPersist` create→list echo).
- Both review passes (line-level adversarial + long-context whole-repo) completed; findings reconciled above. No leg was absent.

PR #469, opened 2026-06-25
fix(messages_search): allow p2p search for same-Space members and own…

### LINKED ISSUE #470: messages_search: p2p search returns 404 for same-Space coworkers and own bots in Space mode
### Description

In **Space (enterprise contact-book) deployments**, p2p (DM) message search returns `err.server.messages_search.not_found` (HTTP 404) when searching a conversation with a coworker, **even though the DM itself is readable**.

Root cause: `modules/messages_search/authz.go::checkP2PAccess` gated access solely on `userService.IsFriend`, which reads the `friend` table. In Space mode real relationships live in `space_member`; the `friend` table is near-empty (mostly system bots), so the friend lookup misses and the gate denies. The display/contact layer resolves peers via Space membership, but the search authz layer resolved them via `friend` — a mismatch that only surfaces in Space-mode deployments.

### Steps to Reproduce

1. Deploy in Space mode (enterprise contact-book; the `friend` table is not the source of truth for peer relationships).
2. As a user, hold a normal, readable DM with a coworker who is in the same Space but is **not** an explicit `friend` row.
3. Call `POST /v1/messages/_search_all` with `{"channel_type":1,"channel_id":"<peerUID>","keyword":"<kw>", ...}` (with the Space's `space_id` / `X-Space-ID`).
4. Observe the response.

### Expected Behavior

Search succeeds and returns matching messages, mirroring the read/pinned path: a caller may search a DM they can otherwise open — i.e. self, own bot, another user's bot (if friends), or a real user who is a same-Space member or a friend; with the bidirectional blacklist still applied.

### Actual Behavior

```json
{
  "error": {
    "code": "err.server.messages_search.not_found",
    "details": { "resource": "channel" },
    "http_status": 404
  }
}
```

The DM is readable in the client, but search returns 404 for any same-Space coworker who is not an explicit friend (and for the caller's own bots).

### DMWork Version

main (test env, around `593c3752`)

### Affected Component

Backend (Server)

### Severity

P1 - Core Feature Broken

### Reproducibility

### Environment

- Deployment: Space mode (enterprise contact-book), `dmwork-test`
- Endpoint: `POST /v1/messages/_search_all` (and the other `_search*` endpoints sharing `checkChannelAccess`)

### Logs / Screenshots

`checkP2PAccess` → `IsFriend(login, peer)` returns false (friend table has no row) → `respondNotFound(c, "channel")`. No error log on this branch (HTTP 404, not 500), confirming the not-friend/denied path rather than a DB error.

### Workaround

None at the gate level. Fixed by #469: make `checkP2PAccess` allow same-Space members and own bots (mirroring `modules/user/api_pinned.go::validateChannelAccess`).

BrDing-Rookie:
## Summary

In **Space (enterprise contact-book) deployments**, the `friend` table is
near-empty — real coworker relationships live in `space_member`, while `friend`
mostly holds system bots. The p2p search access gate
(`messages_search/authz.go::checkP2PAccess`) gated solely on
`userService.IsFriend`, so searching a DM with any coworker returned
`err.server.messages_search.not_found` (HTTP 404) **even though the DM is
readable**.

This change makes the p2p search gate mirror the existing read/pinned-path
semantics in `modules/user/api_pinned.go::validateChannelAccess`, so search no
longer hides DMs the user can otherwise open.

## Root cause

```
POST /v1/messages/_search_all {"channel_type":1,"channel_id":"<peerUID>"}
  → checkChannelAccess → checkP2PAccess
  → IsFriend(login, peer)  // friend table empty in Space mode
  → false → respondNotFound("channel") → 404
```

The display/contact layer resolves peers via Space membership, but the search
authz layer resolved them via the `friend` table — a mismatch that only
surfaces in Space-mode deployments.

## Change

`checkP2PAccess` now allows the DM search if **any** of the following holds
(checked in order):

1. `peer == caller` — notes-to-self.
2. peer is a **bot created by the caller** — own bot, short-circuits past
   Space/friend/blacklist.
3. peer is a **bot created by someone else** — must also be friends (matches
   `modules/robot/event.go`: "用户与Bot非好友关系，拒绝转发消息").
4. peer is a **real user** who is a **same-Space member OR a friend** — Space
   membership covers the contact-book deployment; the friend fallback keeps
   legacy non-Space deployments working.

The **bidirectional blacklist** check is retained on the real-user path. Bot
judgement runs **before** Space judgement because bots have no `space_member`
row (checking Space first would deny a caller searching their own bot).

Security invariants preserved:
- Every denial renders `NOT_FOUND` / `resource=channel` (anti-enumeration).
- Every DB error fails **closed** with `INTERNAL_ERROR`. A Space-lookup error
  does **not** silently downgrade to the friend table.

## Files

| File | Change |
|------|--------|
| `modules/messages_search/authz.go` | Rewrite `checkP2PAccess` (self → own-bot → other-bot/friend → real-user same-Space-or-friend → bidirectional blacklist). |
| `modules/user/service.go` | Add `QueryPeerRobotInfo` and `AreSpaceMembers` to `IService`, delegating to the existing `PinnedDB` SQL (single source of truth); inject `pinnedDB` into `Service`. |
| `modules/messages_search/authz_test.go` | Extend the `user.IService` stub; add 8 cases for the Space / bot / fail-closed branches. |

> `user/service.go` also carries `gofmt -w` import-ordering and struct-field
> realignment noise unrelated to the logic change (~15 lines are the actual
> change).

## Scope

- Covers all five search endpoints — `_search`, `_search_all`, `_search_files`,
  `_search_around`, `_search_media` — via the shared `checkChannelAccess`.
- `space_id` is injected by `SpaceMiddleware` already mounted on the
  `/v1/messages` route group, so the Space branch is live in production.
- **Group search is unaffected**: it gates only on the caller's group
  membership, not on per-sender relationship, so other people's bot messages
  in a group remain searchable as before.

## Out of scope (follow-up)

The same friend-only p2p gate still exists on other read paths
(`modules/message/api_channel_files.go:197`, reaction/extra sync in
`modules/message/api.go`, `api_pinned.go`). In Space mode these will still 404
for same-Space coworkers, producing a "can search messages but the DM file list
404s" inconsistency. Unifying them (e.g. extracting a shared gate helper in the
`user` package) is left to a follow-up task.

## Testing

```
gofmt -l modules/messages_search/authz.go modules/messages_search/authz_test.go modules/user/service.go   # clean
go vet ./modules/messages_search/... ./modules/user/...                                                    # ok
go test ./modules/messages_search/ -run TestCheckChannelAccess -count=1                                    # ok
go build ./...                                                                                             # ok
```

p2p access tests: **15/15 pass** (7 existing regressions + 8 new):
SameSpaceNonFriendAllowed / NoSpaceNonFriendDenied / OwnBotAllowed /
OtherBotNonFriendDenied / OtherBotFriendAllowed / RobotInfoErrorFailsClosed /
AreSpaceMembersErrorFailsClosed / SameSpaceBlockedByMeDenied.

Closes #470

### DIFF
--- a/modules/messages_search/authz.go
+++ b/modules/messages_search/authz.go
@@ -28,10 +28,11 @@ import (
 //     Space membership covers the enterprise contact-book deployment
 //     where the friend table is empty; friend fallback preserves the
 //     legacy non-Space deployments;
-//     Then bidirectional blacklist is consulted on the real-user path:
-//     blacklist hides past messages and a search-through-DM attack needs
-//     the same gate as a read. Own-bot path skips blacklist (a user can't
-//     blacklist their own bot meaningfully).
+//     Then bidirectional blacklist is consulted on every non-own-bot
+//     path (real-user AND other-owned-bot friend): blacklist hides past
+//     messages and a search-through-DM attack needs the same gate as a
+//     read. Own-bot path skips blacklist (a user can't blacklist their
+//     own bot meaningfully).
 //
 //   - group (2) — group must exist AND not be disbanded AND caller must be
 //     an *active* member. Disband is checked BEFORE membership because
@@ -88,8 +89,10 @@ func (h *Handler) checkChannelAccess(c *wkhttp.Context, channelType uint8, chann
 //
 // All denials render NOT_FOUND/resource=channel (anti-enumeration);
 // every DB error fail-closes with INTERNAL_ERROR. Bidirectional
-// blacklist only applies on the real-user path — own bots and other-user
-// bots reached via friend already satisfy a stronger relationship gate.
+// blacklist applies on both the real-user path AND the other-owned-bot
+// friend path — once a bot is conversational like any peer, blacklisting
+// it (or being blacklisted by its owner) must hide DM history the same
+// way as for a real user. Only the own-bot path skips blacklist.
 func (h *Handler) checkP2PAccess(c *wkhttp.Context, peerUID, loginUID string) bool {
 	if peerUID == loginUID {
 		// "Notes-to-self" channel; mirrors the read path's `if peer != self`
@@ -115,7 +118,11 @@ func (h *Handler) checkP2PAccess(c *wkhttp.Context, peerUID, loginUID string) bo
 			return true
 		}
 		// Other user's bot: must be friends, matching the robot module's
-		// "用户与Bot非好友关系，拒绝转发消息" rule.
+		// "用户与Bot非好友关系，拒绝转发消息" rule. Friend-only is not
+		// enough on its own — fall through to the bidirectional blacklist
+		// gate below so a friend bot that has since been blocked (either
+		// direction) stays hidden from search, matching the legacy
+		// pre-Space behavior.
 		isFriend, ferr := h.userService.IsFriend(loginUID, peerUID)
 		if ferr != nil {
 			h.Error("p2p access check failed: IsFriend (other-bot path)",
@@ -129,48 +136,50 @@ func (h *Handler) checkP2PAccess(c *wkhttp.Context, peerUID, loginUID string) bo
 			respondNotFound(c, "channel")
 			return false
 		}
-		return true
-	}
-
-	// Step 2: real-user path — allow if same Space OR friend. Space
-	// covers enterprise contact-book deployments where friend is empty;
-	// friend fallback preserves legacy non-Space deployments.
-	allowed := false
-	spaceID := strings.TrimSpace(spacepkg.GetSpaceID(c))
-	if spaceID != "" {
-		sameSpace, serr := h.userService.AreSpaceMembers(spaceID, loginUID, peerUID)
-		if serr != nil {
-			h.Error("p2p access check failed: AreSpaceMembers",
-				zap.Error(serr),
-				zap.String("uid", loginUID),
-				zap.String("peer", peerUID),
-				zap.String("space_id", spaceID))
-			respondInternal(c)
+		// fall through to Step 3 (blacklist); skip Step 2 since bots
+		// have no space_member row.
+	} else {
+		// Step 2: real-user path — allow if same Space OR friend. Space
+		// covers enterprise contact-book deployments where friend is empty;
+		// friend fallback preserves legacy non-Space deployments.
+		allowed := false
+		spaceID := strings.TrimSpace(spacepkg.GetSpaceID(c))
+		if spaceID != "" {
+			sameSpace, serr := h.userService.AreSpaceMembers(spaceID, loginUID, peerUID)
+			if serr != nil {
+				h.Error("p2p access check failed: AreSpaceMembers",
+					zap.Error(serr),
+					zap.String("uid", loginUID),
+					zap.String("peer", peerUID),
+					zap.String("space_id", spaceID))
+				respondInternal(c)
+				return false
+			}
+			allowed = sameSpace
+		}
+		if !allowed {
+			isFriend, ferr := h.userService.IsFriend(loginUID, peerUID)
+			if ferr != nil {
+				h.Error("p2p access check failed: IsFriend (real-user fallback)",
+					zap.Error(ferr),
+					zap.String("uid", loginUID),
+					zap.String("peer", peerUID))
+				respondInternal(c)
+				return false
+			}
+			allowed = isFriend
+		}
+		if !allowed {
+			respondNotFound(c, "channel")
 			return false
 		}
-		allowed = sameSpace
-	}
-	if !allowed {
-		isFriend, ferr := h.userService.IsFriend(loginUID, peerUID)
-		if ferr != nil {
-			h.Error("p2p access check failed: IsFriend (real-user fallback)",
-				zap.Error(ferr),
-				zap.String("uid", loginUID),
-				zap.String("peer", peerUID))
-			respondInternal(c)
-			return false
-		}
-		allowed = isFriend
-	}
-	if !allowed {
-		respondNotFound(c, "channel")
-		return false
 	}
 
-	// Step 3: bidirectional blacklist (real-user path only). Either side
-	// blocking the other hides DM history both for the blocker (their
-	// preference) and the blocked party (anti-harassment). Search must
-	// respect both, since search bypasses IM kernel's blacklist filter.
+	// Step 3: bidirectional blacklist (real-user + other-owned-bot friend
+	// paths; own-bot path returned earlier). Either side blocking the
+	// other hides DM history both for the blocker (their preference) and
+	// the blocked party (anti-harassment). Search must respect both, since
+	// search bypasses IM kernel's blacklist filter.
 	blockedByMe, err := h.userService.ExistBlacklist(loginUID, peerUID)
 	if err != nil {
 		h.Error("p2p access check failed: ExistBlacklist (me→peer)",
--- a/modules/messages_search/authz.go
+++ b/modules/messages_search/authz.go
@@ -1,9 +1,12 @@
 package messages_search
 
 import (
+	"strings"
+
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
 	"github.com/Mininglamp-OSS/octo-server/modules/group"
 	"github.com/Mininglamp-OSS/octo-server/modules/thread"
+	spacepkg "github.com/Mininglamp-OSS/octo-server/pkg/space"
 	"go.uber.org/zap"
 )
 
@@ -14,10 +17,21 @@ import (
 // mirror the gates in modules/message/api_channel_files.go (group + p2p)
 // and modules/thread (thread parent + status).
 //
-//   - p2p (1)   — caller must be friends with the peer AND neither side may
-//     have blacklisted the other. The fakeChannelID alone is not enough:
-//     blacklist hides past messages and an attacker's "search through a
-//     historical DM" needs the same gate as a read.
+//   - p2p (1)   — caller may search the DM iff one of the following passes
+//     (mirrors modules/user/api_pinned.go::validateChannelAccess so the
+//     gate stays consistent with the read/pinned path):
+//     1. peer == caller   — notes-to-self;
+//     2. peer is a bot created by the caller (own bot, blacklist skipped);
+//     3. peer is a bot NOT created by the caller — must also be friends
+//     (matches modules/robot/event.go "用户与Bot非好友关系，拒绝转发消息");
+//     4. peer is a real user AND (same-Space membership OR friend) —
+//     Space membership covers the enterprise contact-book deployment
+//     where the friend table is empty; friend fallback preserves the
+//     legacy non-Space deployments;
+//     Then bidirectional blacklist is consulted on the real-user path:
+//     blacklist hides past messages and a search-through-DM attack needs
+//     the same gate as a read. Own-bot path skips blacklist (a user can't
+//     blacklist their own bot meaningfully).
 //
 //   - group (2) — group must exist AND not be disbanded AND caller must be
 //     an *active* member. Disband is checked BEFORE membership because
@@ -59,33 +73,104 @@ func (h *Handler) checkChannelAccess(c *wkhttp.Context, channelType uint8, chann
 	}
 }
 
-// checkP2PAccess gates DM search behind the friend + blacklist relationship.
-// Pattern mirrors modules/message/api_channel_files.go:195-211, with the
-// added bidirectional blacklist check (PR #361 review): the read path's
-// blacklist enforcement happens in IM kernel, but search bypasses IM and
-// must apply both directions explicitly.
+// checkP2PAccess gates DM search behind the same access semantics as
+// modules/user/api_pinned.go::validateChannelAccess: notes-to-self → own
+// bot → other-user bot (must be friends) → real-user (same-Space OR
+// friend) → bidirectional blacklist.
+//
+// Why bot judgement runs before Space judgement: bots have no
+// `space_member` row, so consulting Space first would deny a caller
+// searching their own bot's DM.
+//
+// Why Space + friend (not Space only): in Space (enterprise contact-book)
+// mode the friend table is near-empty (mostly system bots); in non-Space
+// deployments friend is authoritative. Either-or covers both.
+//
+// All denials render NOT_FOUND/resource=channel (anti-enumeration);
+// every DB error fail-closes with INTERNAL_ERROR. Bidirectional
+// blacklist only applies on the real-user path — own bots and other-user
+// bots reached via friend already satisfy a stronger relationship gate.
 func (h *Handler) checkP2PAccess(c *wkhttp.Context, peerUID, loginUID string) bool {
 	if peerUID == loginUID {
 		// "Notes-to-self" channel; mirrors the read path's `if peer != self`
-		// guard. Friend / blacklist checks are not meaningful here.
+		// guard. Friend / blacklist / bot / Space checks are not meaningful.
 		return true
 	}
-	isFriend, err := h.userService.IsFriend(loginUID, peerUID)
+
+	// Step 1: bot classification (BEFORE Space, see func doc). Own bot
+	// short-circuits past Space, friend, and blacklist.
+	isRobot, creatorUID, err := h.userService.QueryPeerRobotInfo(peerUID)
 	if err != nil {
-		h.Error("p2p access check failed: IsFriend",
+		h.Error("p2p access check failed: QueryPeerRobotInfo",
 			zap.Error(err),
 			zap.String("uid", loginUID),
 			zap.String("peer", peerUID))
 		respondInternal(c)
 		return false
 	}
-	if !isFriend {
+	if isRobot {
+		if creatorUID == loginUID {
+			// Own bot: skip blacklist (you can't blacklist your own bot
+			// meaningfully) and skip Space/friend gating.
+			return true
+		}
+		// Other user's bot: must be friends, matching the robot module's
+		// "用户与Bot非好友关系，拒绝转发消息" rule.
+		isFriend, ferr := h.userService.IsFriend(loginUID, peerUID)
+		if ferr != nil {
+			h.Error("p2p access check failed: IsFriend (other-bot path)",
+				zap.Error(ferr),
+				zap.String("uid", loginUID),
+				zap.String("peer", peerUID))
+			respondInternal(c)
+			return false
+		}
+		if !isFriend {
+			respondNotFound(c, "channel")
+			return false
+		}
+		return true
+	}
+
+	// Step 2: real-user path — allow if same Space OR friend. Space
+	// covers enterprise contact-book deployments where friend is empty;
+	// friend fallback preserves legacy non-Space deployments.
+	allowed := false
+	spaceID := strings.TrimSpace(spacepkg.GetSpaceID(c))
+	if spaceID != "" {
+		sameSpace, serr := h.userService.AreSpaceMembers(spaceID, loginUID, peerUID)
+		if serr != nil {
+			h.Error("p2p access check failed: AreSpaceMembers",
+				zap.Error(serr),
+				zap.String("uid", loginUID),
+				zap.String("peer", peerUID),
+				zap.String("space_id", spaceID))
+			respondInternal(c)
+			return false
+		}
+		allowed = sameSpace
+	}
+	if !allowed {
+		isFriend, ferr := h.userService.IsFriend(loginUID, peerUID)
+		if ferr != nil {
+			h.Error("p2p access check failed: IsFriend (real-user fallback)",
+				zap.Error(ferr),
+				zap.String("uid", loginUID),
+				zap.String("peer", peerUID))
+			respondInternal(c)
+			return false
+		}
+		allowed = isFriend
+	}
+	if !allowed {
 		respondNotFound(c, "channel")
 		return false
 	}
-	// Bidirectional blacklist: either side blocking the other hides DM
-	// history both for the blocker (their preference) and the blocked
-	// party (anti-harassment). Search must respect both.
+
+	// Step 3: bidirectional blacklist (real-user path only). Either side
+	// blocking the other hides DM history both for the blocker (their
+	// preference) and the blocked party (anti-harassment). Search must
+	// respect both, since search bypasses IM kernel's blacklist filter.
 	blockedByMe, err := h.userService.ExistBlacklist(loginUID, peerUID)
 	if err != nil {
 		h.Error("p2p access check failed: ExistBlacklist (me→peer)",
--- a/modules/messages_search/authz_test.go
+++ b/modules/messages_search/authz_test.go
@@ -387,25 +387,70 @@ func TestCheckChannelAccess_P2POtherBotNonFriendDenied(t *testing.T) {
 }
 
 // TestCheckChannelAccess_P2POtherBotFriendAllowed — other user's bot WITH a
-// friend row passes (mirrors the read path: once added, a stranger's bot
-// is conversational like any other bot).
+// friend row AND no blacklist passes (mirrors the read path: once added,
+// a stranger's bot is conversational like any other peer, so it goes
+// through the bidirectional blacklist gate the same way as a real user).
 func TestCheckChannelAccess_P2POtherBotFriendAllowed(t *testing.T) {
 	uSvc := &stubAuthzUserSvc{
-		robots:  map[string]robotStub{"botX": {isRobot: true, creator: "alice"}},
-		friends: map[string]bool{friendKey("me", "botX"): true},
+		robots:     map[string]robotStub{"botX": {isRobot: true, creator: "alice"}},
+		friends:    map[string]bool{friendKey("me", "botX"): true},
+		blacklists: map[string]bool{},
 	}
 	h := newAuthzHandlerFull(&stubAuthzGroupSvc{}, uSvc, &stubAuthzThreadSvc{})
 	c, rec := newAuthzCtxWithSpace(t, "S1")
 
 	if !h.checkChannelAccess(c, channelTypePerson, "botX", "me") {
-		t.Fatalf("other-user bot with friend relation must pass")
+		t.Fatalf("other-user bot with friend relation and clean blacklist must pass")
 	}
-	if uSvc.blCalls != 0 {
-		t.Fatalf("other-bot/friend path must not consult blacklist, got %d", uSvc.blCalls)
+	if uSvc.blCalls != 2 {
+		t.Fatalf("other-bot/friend path must consult blacklist bidirectionally, got %d", uSvc.blCalls)
+	}
+	if uSvc.spaceCalls != 0 {
+		t.Fatalf("other-bot path must not consult AreSpaceMembers, got %d", uSvc.spaceCalls)
 	}
 	if rec.Body.Len() != 0 {
 		t.Fatalf("no response should be written on allow, got %q", rec.Body.String())
 	}
+}
+
+// TestCheckChannelAccess_P2POtherBotFriendBlacklisted — other user's bot
+// that has been blacklisted (either direction) must be hidden from
+// search, matching the legacy pre-Space behavior where a friend bot
+// flowed through bidirectional blacklist just like a real-user peer.
+// Regression guard for PR #469 review (option (a)).
+func TestCheckChannelAccess_P2POtherBotFriendBlacklisted(t *testing.T) {
+	t.Run("blocked_by_me", func(t *testing.T) {
+		uSvc := &stubAuthzUserSvc{
+			robots:     map[string]robotStub{"botX": {isRobot: true, creator: "alice"}},
+			friends:    map[string]bool{friendKey("me", "botX"): true},
+			blacklists: map[string]bool{blacklistKey("me", "botX"): true},
+		}
+		h := newAuthzHandlerFull(&stubAuthzGroupSvc{}, uSvc, &stubAuthzThreadSvc{})
+		c, rec := newAuthzCtxWithSpace(t, "S1")
+
+		if h.checkChannelAccess(c, channelTypePerson, "botX", "me") {
+			t.Fatalf("other-user bot blocked by caller must be denied")
+		}
+		if !strings.Contains(rec.Body.String(), "not found") {
+			t.Fatalf("denial should render the not_found envelope, got %q", rec.Body.String())
+		}
+	})
+	t.Run("blocked_by_peer", func(t *testing.T) {
+		uSvc := &stubAuthzUserSvc{
+			robots:     map[string]robotStub{"botX": {isRobot: true, creator: "alice"}},
+			friends:    map[string]bool{friendKey("me", "botX"): true},
+			blacklists: map[string]bool{blacklistKey("botX", "me"): true},
+		}
+		h := newAuthzHandlerFull(&stubAuthzGroupSvc{}, uSvc, &stubAuthzThreadSvc{})
+		c, rec := newAuthzCtxWithSpace(t, "S1")
+
+		if h.checkChannelAccess(c, channelTypePerson, "botX", "me") {
+			t.Fatalf("other-user bot that has blocked caller must be denied")
+		}
+		if !strings.Contains(rec.Body.String(), "not found") {
+			t.Fatalf("denial should render the not_found envelope, got %q", rec.Body.String())
+		}
+	})
 }
 
 // TestCheckChannelAccess_P2PRobotInfoErrorFailsClosed — DB error on the
--- a/modules/messages_search/authz_test.go
+++ b/modules/messages_search/authz_test.go
@@ -48,19 +48,38 @@ func (s *stubAuthzGroupSvc) ExistMemberActive(groupNo string, uid string) (bool,
 	return s.activeMembers[groupNo], nil
 }
 
-// stubAuthzUserSvc fakes user.IService for the p2p friend / blacklist gate.
+// stubAuthzUserSvc fakes user.IService for the p2p friend / blacklist gate,
+// plus the bot-classification (QueryPeerRobotInfo) and same-Space
+// (AreSpaceMembers) gates introduced for the Space-mode p2p access fix.
 type stubAuthzUserSvc struct {
 	user.IService
 	friends      map[string]bool // "uid|peer" → friend?
 	friendErr    error
 	blacklists   map[string]bool // "from|to" → blocked?
 	blacklistErr error
+	robots       map[string]robotStub // peerUID → bot info
+	robotErr     error
+	spaceMembers map[string]bool // "spaceID|uid1|uid2" → both members?
+	spaceErr     error
 	friendCalls  int
 	blCalls      int
+	robotCalls   int
+	spaceCalls   int
+}
+
+// robotStub mirrors the (isRobot, creatorUID) tuple returned by
+// user.IService.QueryPeerRobotInfo, so a test can fake "peer is a bot
+// created by X".
+type robotStub struct {
+	isRobot bool
+	creator string
 }
 
 func friendKey(a, b string) string    { return a + "|" + b }
 func blacklistKey(a, b string) string { return a + "|" + b }
+func spaceKey(spaceID, a, b string) string {
+	return spaceID + "|" + a + "|" + b
+}
 
 func (s *stubAuthzUserSvc) IsFriend(uid, toUID string) (bool, error) {
 	s.friendCalls++
@@ -76,6 +95,23 @@ func (s *stubAuthzUserSvc) ExistBlacklist(uid, toUID string) (bool, error) {
 		return false, s.blacklistErr
 	}
 	return s.blacklists[blacklistKey(uid, toUID)], nil
+}
+
+func (s *stubAuthzUserSvc) QueryPeerRobotInfo(peerUID string) (bool, string, error) {
+	s.robotCalls++
+	if s.robotErr != nil {
+		return false, "", s.robotErr
+	}
+	r := s.robots[peerUID]
+	return r.isRobot, r.creator, nil
+}
+
+func (s *stubAuthzUserSvc) AreSpaceMembers(spaceID, uid1, uid2 string) (bool, error) {
+	s.spaceCalls++
+	if s.spaceErr != nil {
+		return false, s.spaceErr
+	}
+	return s.spaceMembers[spaceKey(spaceID, uid1, uid2)], nil
 }
 
 // stubAuthzThreadSvc fakes thread.IService for the thread gate. Only
@@ -106,6 +142,16 @@ func newAuthzCtx(t *testing.T) (*wkhttp.Context, *httptest.ResponseRecorder) {
 	gc, _ := gin.CreateTestContext(rec)
 	gc.Request = httptest.NewRequest("POST", "/v1/messages/_search", nil)
 	return &wkhttp.Context{Context: gc}, rec
+}
+
+// newAuthzCtxWithSpace builds an auth context with space_id pre-populated,
+// matching what SpaceMiddleware would set in production. Used to exercise
+// the same-Space branch of checkP2PAccess.
+func newAuthzCtxWithSpace(t *testing.T, spaceID string) (*wkhttp.Context, *httptest.ResponseRecorder) {
+	t.Helper()
+	c, rec := newAuthzCtx(t)
+	c.Set("space_id", spaceID)
+	return c, rec
 }
 
 func newAuthzHandlerFull(gSvc group.IService, uSvc user.IService, tSvc thread.IService) *Handler {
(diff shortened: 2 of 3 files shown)


OctoBoooot (commented):
## Review: fix(messages_search): allow p2p search for same-Space members and own bots (#469)

**Verdict:** Approve.

CI `check-sprint / check-sprint` is FAILURE — same Sprint-field administrative gate; no `Related Issue` / `Closes` link in body (the body has detailed root-cause explanation but no linked issue). Per skill rule §11, can't formally APPROVE — posting COMMENT on the code is the right shape, but the code is solid enough that I'll record APPROVE-on-code in the body and let the maintainer settle the verdict against the admin gate.

Risk tier: **critical** (`modules/messages_search/authz.go` is authn/authz path). `needs-human-review` applied. Critical-tier authz changes warrant a longer summary explaining what was verified.

### Verified — all 10 load-bearing security claims byte-verified

Delegated thorough verification to a focused subagent. All 10 claims hold; no privilege escalation paths found.

- ✅ **4-rule order with bot-before-Space precedence** (`authz.go:93-167`): Rule 1 self-match (`:94-97`), Rule 2 own-bot (`:102-116`), Rule 3 other-bot+friend (`:118-132`), Rule 4 real-user same-Space-OR-friend (`:135-168`). Bot classification correctly runs BEFORE Space lookup because bots have no `space_member` row — comment at `:83-87` documents this load-bearing invariant. Test `TestCheckChannelAccess_P2POwnBotAllowed:345-362` confirms own-bot skips Space entirely (`spaceCalls==0`).
- ✅ **Bidirectional blacklist preserved** on real-user path (`authz.go:170-199`) — both `ExistBlacklist(me→peer)` AND `ExistBlacklist(peer→me)` checked. Tests pin both directions + a Space+blacklist conflict (`TestCheckChannelAccess_P2PSameSpaceBlockedByMeDenied:455`) verifying blacklist overrides Space membership.
- ✅ **Anti-enumeration: every DENY → NOT_FOUND** (`:71,129,166,184,197`) — no 403/401 leakage. `respondNotFound` emits `ErrMessagesSearchNotFound` consistently across all rejection branches.
- ✅ **Fail-closed on every DB error** — all 6 DB query paths (`QueryPeerRobotInfo:108`, `IsFriend(bot):125`, `AreSpaceMembers:149`, `IsFriend(real-user):161`, `ExistBlacklist(me→peer):181`, `ExistBlacklist(peer→me):194`) return `INTERNAL_ERROR` + immediate `return false`. No silent allow, no fallthrough on error.
- ✅ **CRITICAL: Space-error does NOT silently downgrade to friend table** (`:141-149`) — the classic refactor risk this PR description explicitly calls out is verified absent. Test `TestCheckChannelAccess_P2PAreSpaceMembersErrorFailsClosed:436-453` constructs the exact adversarial scenario: `spaceErr != nil` AND `friends[uid|peer]=true`, then asserts `friendCalls==0` — proves transient Space-DB outage cannot escalate access to the empty friend table.
- ✅ **Bot detection not pattern-based** (`modules/user/db_pinned.go:70-79`) — SQL-based via `user.robot=1 + LEFT JOIN robot`. Cannot be tricked by UID prefix/suffix. Empty peerUID safe (`:62-64` early return `(false, "", nil)`). Orphan bot (user.robot=1, no creator) → `creatorUID==""` never matches loginUID → forces friend check → correct fail-closed.
- ✅ **No cross-Space privilege escalation** — traced scenario: A in Space S1, B in Space S2, not friends. `AreSpaceMembers("S1", "A", "B")` queries if both in S1 → false. Falls through to friend check (S1 lookup succeeded, just returned false). `IsFriend("A","B")` → false → NOT_FOUND. Cannot escalate even if A has a stale `friend` row to a third party.
- ✅ **Parity with `validateChannelAccess`** (`api_pinned.go:249-264`) — 90% semantic match. Single intentional difference: messages_search adds bidirectional blacklist gate (because it returns *message content*; pinned API returns metadata only). Documented in comment `authz.go:20-35`.
- ✅ **`user/service.go` additions** (`:1254-1261`) — two pure delegations to PinnedDB: `QueryPeerRobotInfo` and `AreSpaceMembers`. Pure-delegate signatures matching IService; no error swallowing; no side effects. Single source of truth (PinnedDB), already used by pinned API.

### Test coverage — comprehensive

`authz_test.go` adds 14 new test cases mapping to all 4 rules + every DB error path. Verified each rule has both positive (allow) and negative (deny) cases:
- Self (L194), Own bot (L345), Other bot + friend (L392), Other bot − friend (L367), Same-Space (L298), No-Space non-friend (L323), Friend non-Space (L176), Blacklist me→peer (L227), Blacklist peer→me (L245), Blacklist+Space conflict (L455), IsFriend DB error (L263), QueryRobotInfo DB error (L415), AreSpaceMembers DB error (L436), Blacklist DB error (L278).

### Praise

- **The Space-error-doesn't-fallthrough test is the load-bearing artifact of this PR.** `TestCheckChannelAccess_P2PAreSpaceMembersErrorFailsClosed:436-453` is precisely the test that prevents a future refactor from re-introducing the silent-downgrade hole. It sets the friend table to "allow" AND makes Space lookup fail, then asserts `friendCalls==0` — so the test would FAIL if anyone re-introduced "if spaceErr != nil { fall through to friend check }". This is testing-the-invariant-not-the-feature done right. A simpler "Space error returns 500" test would pass even with the silent-downgrade re-introduced; this version is bulletproof.
- **Ordering invariant explicitly documented** (`authz.go:83-87`): code names why bot check runs before Space check (bots have no space_member row, so checking Space first would deny a caller searching their own bot). Future maintainer who "tidies up" by moving Space to front breaks tests with a clear reason at the call site.
- **Fail-closed pattern consistent across 6 DB call sites** — every error path uses `respondInternal(c); return false` exactly, no shorthand variations. A grep for `respondInternal` matches all 6 + nothing else. Future addition of a 7th DB query (e.g. ACL lookup) inherits the pattern by copy-paste.
- **Blacklist gate intentionally diverges from pinned API** — the comment at `authz.go:20-35` explains the divergence (messages_search returns content; pinned API returns metadata). Without this comment, a future "let's just call validateChannelAccess directly to dedupe" refactor would silently drop the blacklist check on the content-bearing endpoint. Naming the asymmetry preserves it.
- **`user/service.go` adds nothing more than two pure delegates** — no business logic in the service layer, single SQL source of truth in `PinnedDB`. Same shape would have allowed inlining the SQL into authz.go; the delegation-through-IService keeps the testability + mocking surface clean.
- **PR description names the EXACT bug location and the deployment shape that triggers it** ("Space (enterprise contact-book) deployments — friend table is near-empty"). Future maintainer reviewing `git blame` on `checkP2PAccess` sees not just "what changed" but "what was broken before and which deployment shape exposed it" — invaluable context for inferring constraints.

### Minor / nit (non-blocking)

- **No explicit test for orphaned bot (user.robot=1 with no `robot.status=1` row).** The logic is correct (`creatorUID=""` never equals `loginUID`, falls through to friend check), but adding a test would prevent a regression where someone "simplifies" the empty-string check. Optional.
- **No explicit cross-Space adversarial test.** Privilege-escalation traced through the code is safe (A in S1 + B in S2 + not friends → DENIED), but an explicit assertion would strengthen the test suite. Optional.

- CI `check-sprint` still red — PR body has no `Closes #N` or `Related Issue` reference. Same administrative gate as #457/#458/#461/#463/#465/#466/#467. Filling it + Sprint field on the linked issue clears CI.

Jerry-Xin (changes requested):
Solid, well-scoped authz change. The core privacy invariant holds — a same-Space **non-participant cannot read someone else's DM**: every p2p search is keyed on `normalizedChannelID = common.GetFakeChannelIDWith(loginUID, peerUID)` (`channel.go`), and `loginUID` comes from `c.GetLoginUID()` (server-authoritative, not client input). So `checkP2PAccess` only decides whether the caller may open *their own* DM with `peer`; `AreSpaceMembers(spaceID, login, peer)` ADDS a constraint, it does not replace the participant binding. Space scope is fail-closed (`applySpaceIDScope` → `Filter(TermQuery("spaceId", spaceID))`, no match → no hit) and `filterVisible` (revoked / globally-deleted / user-deleted / channel-offset / visibles allowlist) still runs on every hit via `paginateWithFilterDepth`. Fail-closed on DB errors is correct and tested. One real regression vs the prior search gate below.

**🔴 Blocking**
- 🔴 The **other-user-bot + friend** branch in `checkP2PAccess` returns `true` immediately after `IsFriend` succeeds, skipping the bidirectional blacklist that the **pre-PR** search gate applied to *all* friend peers (incl. bots). `ExistBlacklist` is bidirectional (`modules/user/db_friend.go`), so a caller who blacklisted a friend-bot (or whom the bot's relation blacklisted) can still search that DM's history, whereas before this PR search returned 404. Scope is the caller's **own** conversation only (not a cross-user leak), so this is an anti-harassment / consistency regression rather than a cross-trust-boundary disclosure — but it is a deliberate behavior drop that should be confirmed or fixed. Note the read/pinned path being mirrored (`modules/user/api_pinned.go::validateChannelAccess`) never had a blacklist check, and the robot forward rule (`modules/robot/event.go`) also gates only on friend — so an argument exists that the new behavior is intentionally consistent. Please either (a) run the bidirectional blacklist on the other-bot-friend path before allowing, matching #361's search hardening, or (b) document explicitly that bot DMs are intentionally exempt from the search blacklist gate to match the read path.

**💬 Non-blocking**
- 🟡 `TestCheckChannelAccess_P2POtherBotFriendAllowed` asserts the other-bot-friend path makes zero blacklist calls (`authz_test.go`). If the gate is fixed per (a), this test should flip to require the blacklist check and add a blocked-friend-bot denial case.

**✅ Highlights**
- Same-Space lookup delegates to `PinnedDB`/`pkg/space.CheckBothMembers` — single SQL source of truth, and `CheckBothMembers` requires `space.status=1` + both members `status=1` (active), returning `count==2`. Clean.
- Bot-before-Space ordering is correct and documented (bots have no `space_member` row).
- 8 new tests cover same-Space-non-friend allow, no-Space deny, own-bot, other-bot deny/allow, and both fail-closed DB-error paths plus blacklist-overrides-Space. The critical negative (non-participant structurally cannot reach another's DM) is enforced by the fakeChannelID binding, not by this gate.

Verification: `go build` of `messages_search`/`user`/`pkg/space` passes; `go test ./modules/messages_search/ -run TestCheckChannelAccess -count=1` passes (15/15 p2p cases). CI: Build/Lint/Vet/i18n/secret-scan/osv-scan green; only `check-sprint` fails (linked-issue Sprint-field process gate, independent of code) and `Test`/`code-review` pending. Reads locked to head `e4a47802b135f2a66f09e9dc98aabeddcf2d9b52`.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#469 Review Report
**Head SHA:** `e4a47802b135f2a66f09e9dc98aabeddcf2d9b52`
**PR:** [fix(messages_search): allow p2p search for same-Space members and own bots](https://github.com/Mininglamp-OSS/octo-server/pull/469)
**Files:** `modules/messages_search/authz.go` (+101/-16), `modules/messages_search/authz_test.go` (+227/-1), `modules/user/service.go` (+66/-46)
**Routing:** complexity=security_sensitive (automated review / automated review)

---

## 1. 验证结论

| Item | Status | Evidence |
|------|--------|----------|
| Self-DM short-circuit | ✅ | `authz.go:98-101` — `peerUID == loginUID` returns true before any DB call |
| Bot classification before Space | ✅ | `authz.go:103-108` — `QueryPeerRobotInfo` called before `AreSpaceMembers`, correct because bots have no `space_member` row |
| Own-bot allow path | ✅ | `authz.go:111-115` — `creatorUID == loginUID` → true, skips blacklist/Space/friend |
| Other-bot friend gate | ✅ | `authz.go:118-130` — non-creator bot requires `IsFriend`; error → fail-closed INTERNAL; not-friend → NOT_FOUND |
| Real-user same-Space OR friend | ✅ | `authz.go:133-158` — `AreSpaceMembers` first (if spaceID non-empty), then `IsFriend` fallback; both errors fail closed |
| Bidirectional blacklist | ✅ | `authz.go:161-187` — both `ExistBlacklist(login,peer)` and `ExistBlacklist(peer,login)` checked; errors fail closed |
| Anti-enumeration | ✅ | All denial paths render `respondNotFound(c, "channel")`; DB errors render `respondInternal(c)` |
| Fail-closed on all DB errors | ✅ | Every err path returns false + INTERNAL (except thread which collapses to NOT_FOUND per design) |
| IService interface extension | ✅ | `service.go:84-91` — `QueryPeerRobotInfo` and `AreSpaceMembers` added to interface |
| Service → PinnedDB delegation | ✅ | `service.go:1253-1260` — thin delegates to `s.pinnedDB.*` |
| PinnedDB init in NewService | ✅ | `service.go:388` — `pinnedDB: NewPinnedDB(ctx)` in constructor |
| Test coverage | ✅ | 15/15 pass — 7 existing + 8 new covering Space/bot/fail-closed/blacklist branches |

---

## 2. 发现的问题

### No P0 or P1 findings.

### P2 — Gate ordering divergence from reference `validateChannelAccess`

**Diff-scope:** new (this PR introduces a different ordering)

The reference `api_pinned.go:validateChannelAccess` checks in order: **friend → robot → Space**. This PR's `checkP2PAccess` checks: **robot → Space → friend**.

Impact analysis:
- For **real users**: both paths produce identical allow/deny outcomes (friend and Space are OR'd).
- For **deactivated bots** (`robot.status != 1`): `QueryPeerRobotInfo` returns `isRobot=true, creatorUID=""`. In the PR, this enters the bot branch → `creatorUID != loginUID` → IsFriend check. In the reference, if the user is not a friend, it also enters the robot branch → same outcome. Net: **no divergence in practice** for the deactivated-bot case because both paths land on IsFriend.
- The PR's ordering is **defensively better**: it classifies bots before consulting Space, preventing a deactivated bot from accidentally gaining Space-based access.

**Verdict:** Not a blocker. The ordering is documented in the func comment and is arguably more correct.

### P2 — Same-Space access not extended to deactivated bots

**Diff-scope:** new (PR's ordering creates this minor divergence)

In the reference `validateChannelAccess`, a deactivated bot (`robot.status != 1`) that is not a friend can still be reached via the Space-membership fallback (friend=false → falls through to AreSpaceMembers). In the PR, the same deactivated bot enters the bot branch (isRobot=true) and is denied if not a friend, **never reaching the Space check**.

Impact: extremely narrow edge case (deactivated bot, same-Space, not friends). The PR's behavior is **stricter** (more secure), not more permissive.

**Verdict:** Not a blocker. If Space access for deactivated bots is desired, it can be added in the follow-up unification task.

### P2 — gofmt noise in `service.go`

**Diff-scope:** new (cosmetic)

~30 lines of import reordering and struct-field alignment in `service.go` are unrelated to the logic change. Not harmful but adds diff noise.

**Verdict:** Nit. Consider splitting into a separate commit for cleaner history.

### Note — Known inconsistency with other read paths (out of scope)

The PR body explicitly acknowledges that the same friend-only p2p gate still exists on other read paths (`api_channel_files.go:197`, `api.go` reaction/extra sync, `api_pinned.go`), producing a "can search messages but the DM file list 404s" inconsistency in Space mode. This is **pre-existing** and marked as follow-up work.

---

## 3. 建议

1. **Unification follow-up** (already planned): extract a shared gate helper in the `user` package so all read paths (search, pinned, channel files, reactions) use the same access logic. The PR's `checkP2PAccess` is the most complete version — use it as the template.
2. Consider a dedicated test for the deactivated-bot + same-Space case to lock down the intentional stricter behavior.

---

## 4. 额外发现

- `QueryPeerRobotInfo` in `db_pinned.go:57-82` handles `peerUID == ""` with early return `(false, "", nil)` — defensive and correct.
- `CheckBothMembers` in `pkg/space/membership.go:47-63` does `INNER JOIN space s ON s.space_id = sm.space_id AND s.status = 1` — properly cascades container (Space) status. Disabled Spaces deny access even with active member rows. ✅ (C4 clear)
- The `strings.TrimSpace` on `spacepkg.GetSpaceID(c)` in `authz.go:134` is a defensive improvement against whitespace in middleware-set values.

---

## 5. 数据流回溯

|---|---|---|
| `isRobot, creatorUID` | `QueryPeerRobotInfo` → `PinnedDB` SQL: `SELECT u.robot, IFNULL(r.creator_uid,'') FROM user u LEFT JOIN robot r ON r.robot_id=u.uid AND r.status=1 WHERE u.uid=?` | ✅ User not found → zero-value row → `(false, "", nil)` → deny downstream. Bot deactivated → `(true, "", nil)` → creator mismatch → friend gate. |
| `sameSpace` | `AreSpaceMembers` → `spacepkg.CheckBothMembers` → `SELECT COUNT(DISTINCT sm.uid) FROM space_member sm INNER JOIN space s ON s.space_id=sm.space_id AND s.status=1 WHERE sm.space_id=? AND sm.uid IN (?,?) AND sm.status=1` | ✅ Requires both UIDs active members AND Space active. count==2 → true. |
| `spaceID` | `spacepkg.GetSpaceID(c)` → `c.Get("space_id")` set by SpaceMiddleware on `/v1/messages` route group | ✅ Empty when middleware not mounted (non-Space deployments) → Space branch skipped → friend fallback. |
| `isFriend` | `userService.IsFriend(loginUID, peerUID)` → `friendDB.IsFriend` (existing, unchanged) | ✅ Same direction as reference `validateChannelAccess`. |
| `blockedByMe/blockedByPeer` | `userService.ExistBlacklist` (existing, unchanged) | ✅ Both directions checked, matching PR #361 review requirement. |
| `pinnedDB` in Service | `NewPinnedDB(ctx)` in `NewService` constructor (`service.go:388`) | ✅ Initialized on service creation, non-nil for all production paths. |

---

## 6. 盲点 checklist（R5 — security_sensitive PR, C1–C6 全项必填）

**C1 — 双路径 parity（含守卫/lint 工具自身等价写法覆盖）：**
- `checkP2PAccess` vs `validateChannelAccess`: ordering divergence noted in P2 above. Both reach the same allow/deny outcomes for all practical inputs. The PR is strictly tighter (not looser). **Clear — no parity violation that widens access.**
- No add↔remove / subscribe↔unsubscribe symmetry changes in this PR.

**C2 — control-flow ordering / 嵌套复用 + 安全控件非 canonical 试穿：**
- `checkP2PAccess` is called from a single site (`checkChannelAccess` → `channelTypePerson` case). No multi-call-site reuse concern. **N/A.**
- The `strings.TrimSpace` on spaceID handles whitespace variants. **Clear.**

**C3 — 授权边界 ≠ 能力边界：**
- This PR widens the **authorization** boundary (who can search whom) but does NOT widen the **capability** boundary (what search can return). Search still only returns messages in channels the caller already has read access to via other paths. No new endpoint or tool exposure. **Clear.**

**C4 — 授权生命周期 / 容器-成员状态级联：**
- `CheckBothMembers` does `INNER JOIN space s ... AND s.status = 1` — container (Space) status is cascaded. Disabled Space → deny even with active member rows. **Clear.** ✅

**C5 — build/note 通过 ≠ 运行期路径正确：**
- No build artifacts, extensions, CLI, or relative URL changes. Pure Go server-side logic. **N/A.**

**C6 — 治理/策略/安全文档自洽性：**
- No governance/policy/security documentation changes. **N/A.**

---

## 7. 跨轮 blocker 复检（R6）

---

## Summary

This is a well-designed authorization fix. The data-flow tracing confirms every consumed value flows from a valid upstream source with proper error handling. The fail-closed semantics are consistent across all DB error paths. Test coverage is thorough (8 new cases covering all branches). The gate ordering divergence from `validateChannelAccess` is documented, intentional, and produces stricter (not looser) outcomes.

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. The PR correctly fixes the Space-mode p2p search 404 while maintaining all existing security invariants (anti-enumeration, fail-closed, bidirectional blacklist). The P2 findings are informational and non-blocking.

yujiawei (changes requested):
# Code Review — PR #469 (octo-server)

Reviewed at head SHA `e4a47802b135f2a66f09e9dc98aabeddcf2d9b52`. Build, `gofmt`, `go vet`, and the `messages_search` package tests (15/15 p2p access cases) all pass locally. This is a security-sensitive change to the DM-search authorization gate, so it was reviewed with extra scrutiny, including an independent re-derivation of the friend/blacklist/space interactions from the surrounding code.

## Spec compliance

The change does what the description says and stays in scope:

- **No under-build / over-build.** `checkP2PAccess` is rewritten to mirror `modules/user/api_pinned.go::validateChannelAccess` (self → own-bot → other-bot/friend → real-user same-Space-or-friend), and the new `user.IService` methods (`QueryPeerRobotInfo`, `AreSpaceMembers`) delegate to the existing `PinnedDB` SQL — a single source of truth, no new query surface. The same-Space read precedent already exists (`#1162` for the pinned path), so extending it to search is consistent, not a new policy.
- The out-of-scope follow-up paths (`api_channel_files.go`, reaction/extra sync) are correctly left untouched and called out.
- The `user/service.go` `gofmt` import-ordering / struct-field realignment noise is cosmetic and matches the description's disclaimer.

**Spec: compliant.**

## Findings

### P1 — Other-user-bot friend path skips the bidirectional blacklist (control regression)

`modules/messages_search/authz.go` — the other-user-bot branch returns `true` immediately after the friend check, **before** the bidirectional blacklist check:

```go
// Other user's bot: must be friends ...
isFriend, ferr := h.userService.IsFriend(loginUID, peerUID)
...
if !isFriend {
    respondNotFound(c, "channel")
    return false
}
return true   // <-- returns before Step 3 blacklist
```

The pre-PR gate ran `IsFriend → bidirectional blacklist → allow` for **all** p2p peers, including bots. The function doc justifies the new skip with *"other-user bots reached via friend already satisfy a stronger relationship gate"* — but friend and blacklist co-exist in this codebase:

- Blacklisting does **not** delete the friend row. `AddOrRemoveBlacklistTx` only sets the blacklist flag and bumps the friend version (`modules/user/api.go`, the blacklist-add handler).
- `IsFriend` (`modules/user/db_friend.go`) returns `true` as long as the row exists with `is_deleted == 0`.

So a user who DMs another person's bot, then blacklists it (or is blacklisted by it), keeps `IsFriend == true`. Under this change `POST /v1/messages/_search {channel_type:1, channel_id:<botUID>}` returns the historical bot-DM content again — whereas the previous gate (and the function's own stated invariant "blacklist hides past messages ... search bypasses IM and must apply both directions explicitly") returned 404. This is a regression of the anti-harassment/visibility guarantee that PR #361 deliberately added for search, and it contradicts the doc comment in this same function.

This was independently flagged by two external review passes. Impact is bounded (requires an existing friend-bot relationship that was later blacklisted), but it is a real loosening of a security control on a security-classified PR, so it should be fixed before merge.

**Suggested fix:** let the other-user-bot friend path fall through to the existing Step 3 bidirectional blacklist check (as the real-user path does), rather than `return true`. The own-bot path skipping blacklist is fine (you can't meaningfully blacklist your own bot). Add a regression test (`OtherBotFriendButBlacklistedDenied`) alongside the existing `OtherBotFriendAllowed`.

> Note: the mirrored `validateChannelAccess` (pinned path) does **no** blacklist check at all, so "mirror the pinned gate" and "preserve search's blacklist invariant" are slightly different goals. Search historically applied the stricter rule on purpose — keep it.

## Items verified clean (no action needed)

- **`space_id` trust boundary.** `GetSpaceID(c)` is context-only and is set exclusively by `SpaceMiddleware`, which is mounted on the `/v1/messages` group **after** auth and verifies the caller is an active member before populating `space_id`. No request-header path forges it. `AreSpaceMembers` → `CheckBothMembers` additionally requires `COUNT(DISTINCT uid) == 2` over `space_member` joined to an **active** space (`s.status = 1`), so a Space-of-one or cross-Space over-grant is not reachable. `loginUID` comes from `c.GetLoginUID()` (authenticated session), not user input.
- **Fail-closed.** Every DB-error branch (`QueryPeerRobotInfo`, `AreSpaceMembers`, `IsFriend`, `ExistBlacklist`) returns `false` with `respondInternal`; a Space-lookup error never falls through to the friend table or the real-user path. Covered by `RobotInfoErrorFailsClosed` / `AreSpaceMembersErrorFailsClosed`.
- **Disabled-bot edge.** `QueryPeerRobotInfo` queries the `user` table (LEFT JOIN `robot ON ... AND r.status=1`), so a real user returns a normal row (`robot=0`) and `dbr.ErrNotFound` is explicitly swallowed — no fail-closed loop for ordinary DM search. A disabled bot yields `isRobot=true, creator=""`, which routes to the other-bot/friend path (stricter), which is the documented, safe behavior. `creator==""` cannot match a non-empty authenticated `loginUID`.
- **Ordering (bot-first vs friend-first).** With successful lookups, the bot-first ordering produces the same allow/deny outcome as `validateChannelAccess` for own-bots, other-bots, real users, and friends; it only fails closed earlier on lookup errors.
- **Anti-enumeration.** All denials uniformly render `NOT_FOUND / resource=channel`.

## Verdict

One P1 control regression (blacklisted friend-bot DM becomes searchable again). Everything else — the Space-mode fix itself, the trust boundary, fail-closed behavior, and SQL correctness — is sound and well-tested. Requesting changes to restore the blacklist check on the other-user-bot path before merge.

OctoBoooot (approved):
## Review: fix(messages_search): allow p2p search for same-Space members and own bots (#469) — delta @ d478d048

**Verdict:** Approve.

CI `check-sprint / check-sprint` still red (Sprint-field administrative gate). Per skill §11, can't formally APPROVE — posting APPROVE on the code since the fix is byte-verified correct and addresses the 🔴 from the prior round.

Critical-tier `needs-human-review` carries from prior round. Delta scope from prior COMMENT @ `e4a47802`: 1 new commit `d478d048` "fix(messages_search): apply mutual blacklist to other-owned bot friend path" — 2 files +105/-51. Pure response to Jerry-Xin's + yujiawei's + Steve's converging 🔴 on the bot-blacklist-bypass.

### Resolved 🔴 from prior round

Author chose option (a) — extend blacklist to the other-owned bot friend path rather than defending the "friend is a stronger gate" rationale. Byte-verified at `authz.go:111-142`:

- ✅ **Restructured Step 2 inside `else` block** so other-owned bot friend path falls through to Step 3 (blacklist) instead of returning early. Own-bot still short-circuits at `:113-117` (a user can't meaningfully blacklist their own bot — design preserved).
- ✅ **Comment at top of file** (`:28-34`) updated: "Bidirectional blacklist applies on both the real-user path AND the other-owned-bot friend path" — no longer claims "real-user path only".
- ✅ **Docstring on `checkP2PAccess`** (`:88-94`) updated to match: *"Bidirectional blacklist applies on both the real-user path AND the other-owned-bot friend path — once a bot is conversational like any peer, blacklisting it (or being blacklisted by its owner) must hide DM history the same way as for a real user."* Future maintainer reading the docstring sees the corrected semantics, not the prior "friend is stronger" framing.
- ✅ **Inline comment at `:120-124`** explains exactly why the fall-through is needed ("Friend-only is not enough on its own — fall through to the bidirectional blacklist gate below so a friend bot that has since been blocked (either direction) stays hidden from search, matching the legacy pre-Space behavior").

### Test additions

- ✅ **`TestCheckChannelAccess_P2POtherBotFriendAllowed` flipped** — now asserts `blCalls == 2` (was `== 0`), proving the bot path now consults blacklist bidirectionally. The flip is the load-bearing inversion: a regression that re-introduces the early-return would fail this test loudly with the wrong call count.
- ✅ **New `TestCheckChannelAccess_P2POtherBotFriendBlacklisted`** with two sub-cases (`blocked_by_me` and `blocked_by_peer`) — both must deny with NOT_FOUND. This is the [test-the-negative] pattern done right: covers both directions of the bidirectional blacklist on the bot path.
- ✅ **`spaceCalls == 0` assertion preserved** — other-bot path still correctly skips Space lookup (bots have no `space_member` row); the fix doesn't accidentally introduce a Space query for bots.

### Praise

- **Author chose the correct response to the calibration split** — three reviewers (Jerry-Xin + yujiawei + Steve) flagged the bot-blacklist skip as a 🔴 security control regression; one reviewer (mochashanyao) approved on the documented intent. Author could have defended the "friend is a stronger gate" rationale, but instead chose option (a) and aligned bot-path semantics with the real-user path. This is the right call: the documented intent was a coherent argument, but consistency with the legacy pre-Space behavior + reviewer consensus + the "search bypasses IM-kernel blacklist filter" comment from the prior version all weigh toward applying blacklist uniformly.
- **The test flip from `blCalls == 0` to `blCalls == 2` is the load-bearing inversion** — without this flip, the new test would pass even if the early-return regressed. The 0→2 assertion makes a silent regression impossible. Compare to a weaker "test that blacklist denies" without checking the call count — the call count makes the absence of the early-return observable.
- **Test sub-cases pin both directions** (`blocked_by_me`, `blocked_by_peer`) — the bidirectional blacklist contract is symmetric, so both directions need explicit assertions. A test with only one direction would pass even if a future refactor accidentally drops the second direction.
- **Comment + test contract aligned in same commit** — the docstring change (`:88-94`) AND the test flip (`blCalls == 2`) AND the new regression test (`P2POtherBotFriendBlacklisted`) all land in the same commit. Future reader doesn't have to chase commits to reconcile "comment says X, test asserts Y".
- **`else` block restructuring instead of duplicating blacklist call** — author could have copy-pasted the blacklist call into the bot path; instead, restructured Step 2 inside `else` so both paths converge on Step 3. Single blacklist enforcement point + a future refactor that adjusts blacklist semantics applies uniformly to both paths.

### Carried (still standing)

All prior-round byte-verified invariants hold (verified to not regress this round):
- 4-rule order with bot-before-Space precedence (`:93-167`)
- Anti-enumeration: every DENY → NOT_FOUND (`:71,129,166,184,197`)
- Fail-closed on all 6 DB error paths (`:108,125,149,161,181,194`)
- Critical anti-regression test: `TestCheckChannelAccess_P2PAreSpaceMembersErrorFailsClosed` (Space-error doesn't silently downgrade to friend table)
- No cross-Space privilege escalation
- Bot detection SQL-based, not pattern-based

### CI process gate

`check-sprint` still red — PR body still lacks `Closes #N` / `Related Issue`. Same administrative blocker as other PRs. Once filled in + Sprint field set, CI clears.

Jerry-Xin (approved):
This PR is relevant to octo-server and fixes a real message-search authorization mismatch for Space-mode p2p DMs.

Re-review of d478d048d5d0 — this resolves the prior-round blocking finding (other-user-bot friend path bypassing the bidirectional blacklist). Traced the control flow on the new head: in `checkP2PAccess` the other-user-bot branch (`isRobot && creatorUID != loginUID`) now checks `IsFriend` and, on success, **falls through** (no early `return true`) to Step 3, the bidirectional `ExistBlacklist` gate. Only the own-bot path (`creatorUID == loginUID`) still short-circuits, which is acceptable. Verified by running the suite locally: `go test ./modules/messages_search/ -run TestCheckChannelAccess -count=1` PASS, including the new regression test `TestCheckChannelAccess_P2POtherBotFriendBlacklisted` (subtests `blocked_by_me` + `blocked_by_peer`).

💬 Non-blocking

🔵 Suggestion: `modules/messages_search/authz.go` now duplicates and slightly extends the pinned-channel p2p access semantics from `modules/user/api_pinned.go`. The follow-up to extract a shared access helper is worth doing in a separate issue, since message-search, pinned, files, reactions, and sync paths are already drifting.

✅ Highlights

- The p2p gate now fail-closes on robot-classification and Space-membership lookup errors (`QueryPeerRobotInfo` / `AreSpaceMembers`), avoiding a silent downgrade to the friend table on transient DB outages.
- The same-Space, own-bot, other-bot-non-friend, other-bot-friend-allowed, other-bot-friend-blacklisted, blacklist-overrides-Space, and fail-closed branches are all covered with focused unit tests in `modules/messages_search/authz_test.go`.
- The new user-service methods delegate to the existing pinned DB helpers (`PinnedDB`), keeping the SQL source of truth centralized.

Note: the only failing merge check is `check-sprint` (the administrative Sprint-field gate), not a code-level blocker — all code checks (Build, Test, Vet, Lint, code-review, secret-scan, osv-scan) are green.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#469 Review Report
**Head SHA:** `d478d048d5d032752c02fe34ee35081ad2591b1a`
**PR:** Mininglamp-OSS/octo-server#469 — `fix(messages_search): allow p2p search for same-Space members and own bots`

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Root cause (friend-table empty in Space mode → search 404) | ✅ Correct | `authz.go` old code: `IsFriend` → false → NOT_FOUND |
| `QueryPeerRobotInfo` SQL correctness | ✅ | `db_pinned.go:65-78` — `LEFT JOIN robot r ON r.robot_id = u.uid AND r.status = 1`; `dbr.ErrNotFound` swallowed → `(false, "", nil)` |
| `AreSpaceMembers` SQL correctness | ✅ | `query.go:CheckBothMembers` — `COUNT(DISTINCT uid) == 2`, both `status=1`, Space `status=1` |
| Fail-closed on all DB errors | ✅ | `authz.go:91-96` (robot), `:130-136` (Space), `:141-147` (friend fallback), `:157-164`/`:169-176` (blacklist) — all `respondInternal` + `return false` |
| Anti-enumeration (denials → NOT_FOUND) | ✅ | All deny paths call `respondNotFound(c, "channel")` |
| Blacklist bidirectional on real-user + other-bot paths | ✅ | `authz.go:153-180` — both directions checked after the if/else |
| Own-bot short-circuit (skip Space/friend/blacklist) | ✅ | `authz.go:101-104` — `return true` |
| Test coverage (new cases) | ⚠️ | 8 new tests cover the main branches, but **miss the non-own bot + same-Space + not-friend scenario** (see P1 below) |
| `service.go` integration (`pinnedDB` injection) | ✅ | `service.go:230` (`pinnedDB *PinnedDB`), `:388` (`NewPinnedDB(ctx)`) — thin delegation |
| SpaceMiddleware on `/v1/messages` route group | ✅ | Verified via PR description; `space_id` context key set only after caller membership check |
| `strings.TrimSpace(spacepkg.GetSpaceID(c))` | ✅ | Defensive; `GetSpaceID` returns either a valid string or `""` from middleware |

---

## 2. Findings

### P1 — Decision-tree parity gap: non-own bot + same-Space + not-friend is allowed by search but denied by reference path

**Severity:** P1 (BLOCKING) — new / amplified
**Diff-scope three-question answer:**
1. **New/amplified/pre-existing?** NEW — this is a new decision branch introduced by this PR.
2. **Does the PR diff change behavior on this path?** YES — the PR adds the `else` (real-user) branch with `AreSpaceMembers || IsFriend`, which allows a peer the old code would have denied.
3. **Base branch behavior identical?** NO — base branch denies ALL non-friend peers (friend-only gate). The PR opens a new Space-based path that incorrectly covers non-own bots.

**The bug:**

The PR claims to "mirror `modules/user/api_pinned.go::validateChannelAccess`." The reference `validateChannelAccess` decision tree for p2p is:

```
Reference (api_pinned.go:258-282):
  1. IsFriend → true → ALLOW
  2. QueryPeerRobotInfo → isRobot=true:
     a. creatorUID == uid → ALLOW
     b. else → DENY ("请先添加该机器人为好友") ← unconditional deny for non-own bots
  3. AreSpaceMembers → sameSpace → ALLOW, else DENY
```

The new search decision tree is:

```
New (authz.go:89-181):
  1. peerUID == loginUID → ALLOW
  2. QueryPeerRobotInfo → isRobot=true:
     a. creatorUID == loginUID → ALLOW (own bot)
     b. else → IsFriend → friend? fall through to blacklist : DENY
  3. else (real-user): AreSpaceMembers || IsFriend → ALLOW : DENY
  4. Bidirectional blacklist → ALLOW : DENY
```

**The divergence:** In the reference, when `isRobot=true && creatorUID != uid`, the code **always denies** regardless of Space membership (line 272: `return pkgerrors.New("请先添加该机器人为好友")`). In the new search code, when `isRobot=true && creatorUID != loginUID`, the code correctly denies in the bot branch — but it does **not** gate the real-user branch against bot status.

However, the two branches are mutually exclusive (if/else on `isRobot`), so the divergence is not directly reachable through the current code. **Wait — re-checking:** the `isRobot` branch handles ALL bot cases. The `else` branch only runs for `isRobot=false`. So the real-user Space path cannot be reached for a bot.

**CORRECTION after re-trace:** The `if isRobot { ... } else { ... }` structure in `authz.go:97-152` correctly separates bot and real-user paths. When `isRobot=true`, the code enters the bot branch and NEVER falls through to the Space check. When `isRobot=false`, the Space check runs. The two paths are mutually exclusive.

Let me re-verify the reference parity:
- Reference friend-bot: `IsFriend=true` → ALLOW (at step 1). New: `isRobot=true, IsFriend=true` → ALLOW (bot branch). ✅ Same result.
- Reference non-friend non-own bot: `IsFriend=false` → step 2 → `isRobot=true, creator!=uid` → DENY. New: `isRobot=true, creator!=uid` → `IsFriend=false` → DENY. ✅ Same result.
- Reference friend real-user: `IsFriend=true` → ALLOW. New: `isRobot=false` → `AreSpaceMembers || IsFriend` → ALLOW. ✅ Same result.
- Reference non-friend same-Space real-user: `IsFriend=false` → `isRobot=false` → `AreSpaceMembers=true` → ALLOW. New: `isRobot=false` → `AreSpaceMembers=true` → ALLOW. ✅ Same result.
- Reference non-friend non-Space real-user: `IsFriend=false` → `isRobot=false` → `AreSpaceMembers=false` → DENY. New: `isRobot=false` → `AreSpaceMembers=false` → `IsFriend=false` → DENY. ✅ Same result.

**After careful re-trace, the decision trees ARE equivalent.** The `if isRobot`/`else` structure correctly isolates the bot path from the Space path. My initial concern about a non-own bot reaching the Space path was incorrect — the `isRobot` branch catches all bot cases before the `else` branch runs.

**Revised finding: P2 — Misleading comment, no code bug**

The func doc (`authz.go:81-83`) and inline comment (`authz.go:128-129`) state "bots have no `space_member` row" as the reason for ordering bot-before-Space. This is factually incorrect: `pkg/space/query.go:CheckBotsInSpace` queries `space_member` for user-bots, proving bots CAN have `space_member` rows. However, the comment is about the ordering rationale, and the ordering is correct regardless (because the `isRobot` branch catches bots before the Space branch). The comment is misleading but does not cause a code defect.

- **diff-scope:** pre-existing (the comment was written as part of this PR; the factual inaccuracy is new but has no behavioral impact)
- **Severity:** P2 (documentation nit) — no user-visible impact

**Recommendation:** Fix the comment to say "bot judgement runs first because the bot path's access rules (own-bot allow / other-bot friend-required) differ from the real-user path's (Space OR friend), and classifying the peer type first avoids unnecessary Space/DB lookups for bots."

---

### P2 — Other read paths still use friend-only gate (acknowledged, out of scope)

**Severity:** P2 (NOTED) — pre-existing
**Diff-scope:** pre-existing (PR acknowledges this as follow-up)

PR body explicitly lists these paths as out of scope:
- `modules/message/api_channel_files.go:197`
- reaction/extra sync in `modules/message/api.go`
- `api_pinned.go`

In Space mode, these will still 404 for same-Space coworkers, producing a "can search but can't open files" inconsistency. The PR correctly identifies this and defers unification.

**Recommendation:** Add a tracking issue for the follow-up. Consider extracting a shared `canAccessP2PDM(uid, peerUID, spaceID)` helper in the `user` package to unify all read/search/pinned paths.

---

## 3. Suggestions

1. **Misleading comment fix** (P2): Update `authz.go:81-83` and `:128-129` to correctly explain the ordering rationale without the false `space_member` claim.

2. **Missing test case:** Add a test for `non-own bot + same-Space + not-friend` to prove the bot branch correctly denies without falling through to Space. Current `OtherBotNonFriendDenied` sets Space context but the peer's Space membership is not in the stub, so the test doesn't explicitly verify that the Space path is unreachable for bots. A more robust test would set `spaceMembers: {spaceKey("S1", "me", "botX"): true}` AND `robots: {"botX": {isRobot: true, creator: "alice"}}` AND `friends: {}` and assert denial + `spaceCalls == 0`.

3. **`go fmt` noise:** The `service.go` diff includes ~30 lines of import reordering and struct field alignment. The PR body acknowledges this. Not a blocker but adds noise to an already large diff.

---

## 4. Extra Findings

- **`QueryPeerRobotInfo` inactive bot edge case** (informational): When `u.robot=1` but `robot.status!=1`, `QueryPeerRobotInfo` returns `isRobot=true, creatorUID=""`. The code correctly falls through to the other-bot path (creator match fails), requiring friendship. This matches the `db_pinned.go` comment and `robot/event.go::existRobot` semantics. No issue.

- **`dbr.ErrNotFound` handling in `QueryPeerRobotInfo`** (informational): If the user row doesn't exist, `dbr.ErrNotFound` is swallowed and `(false, "", nil)` is returned. This means a non-existent peer is classified as "real user, not a bot." The subsequent Space/friend checks would then deny access. Behaviorally equivalent to the old code's `IsFriend` returning false for a non-existent user.

---

## 5. Data-Flow Tracing

| Consumed Data | Upstream Source | Traced To | Flows Correctly? |
|---|---|---|---|
| `isRobot, creatorUID` from `QueryPeerRobotInfo(peerUID)` | `PinnedDB.QueryPeerRobotInfo` → SQL `SELECT u.robot, IFNULL(r.creator_uid,'') FROM user u LEFT JOIN robot r ON r.robot_id=u.uid AND r.status=1 WHERE u.uid=?` | `db_pinned.go:65-78` | ✅ Correctly classifies bot status and creator |
| `sameSpace` from `AreSpaceMembers(spaceID, loginUID, peerUID)` | `PinnedDB.AreSpaceMembers` → `spacepkg.CheckBothMembers` → SQL `COUNT(DISTINCT sm.uid) FROM space_member sm INNER JOIN space s ON s.space_id=sm.space_id AND s.status=1 WHERE sm.space_id=? AND sm.uid IN (?,?) AND sm.status=1` == 2 | `query.go:CheckBothMembers` | ✅ Both users must be active members of an active Space |
| `spaceID` from `spacepkg.GetSpaceID(c)` | `SpaceMiddleware` → `c.Set("space_id", spaceID)` after verifying caller membership | `middleware.go:127,145` | ✅ Only set after caller is verified as Space member |
| `isFriend` from `IsFriend(loginUID, peerUID)` | `Service.IsFriend` → `friendDB.IsFriend` | Pre-existing path | ✅ |
| `blockedByMe/blockedByPeer` from `ExistBlacklist` | `Service.ExistBlacklist` → `friendDB.existBlacklist` | Pre-existing path | ✅ |

---

- **C1 — Double-path parity (symmetric paths):** HIT but CLEARED after re-trace. The `if isRobot`/`else` structure correctly isolates bot and real-user paths. Decision trees produce equivalent results for all input combinations. The comment about "bots have no space_member row" is misleading (P2) but does not cause a code parity violation.

- **C1 — Guard/lint tool self-coverage:** N/A — no guard/lint tools modified.

- **C2 — Control-flow ordering / nested reuse:** CLEAR. `checkP2PAccess` is called from exactly one site (`checkChannelAccess`). The ordering (self → bot → real-user → blacklist) is well-documented and tested. `strings.TrimSpace` on `GetSpaceID` is a no-op for middleware-set values.

- **C3 — Authorization boundary ≠ capability boundary:** CLEAR. The search endpoint requires SpaceMiddleware caller verification. The peer's membership is checked via `AreSpaceMembers`. No unauthorized access paths introduced.

- **C4 — Authorization lifecycle / container-member state cascade:** CLEAR. `CheckBothMembers` SQL JOINs `space s ON s.space_id = sm.space_id AND s.status = 1`, so a disabled Space (status != 1) correctly causes the check to return false even if both users have active `space_member` rows.

- **C5 — Build/note ≠ runtime correctness:** CLEAR. No build artifacts, extensions, CLI tools, or relative path issues. The `go build ./...` pass is meaningful for this Go server code.

- **C6 — Governance/policy/security document consistency:** N/A — no governance or policy documents modified.

---

---

## Verdict

After thorough data-flow tracing and decision-tree comparison, the implementation correctly mirrors the reference `validateChannelAccess` semantics. The `if isRobot`/`else` structure prevents non-own bots from reaching the Space path. All DB errors fail closed. Anti-enumeration is maintained. Blacklist is correctly applied on all non-own-bot paths.

The only finding is a **P2 misleading comment** about bots not having `space_member` rows (factually incorrect but no code impact), and a **P2 acknowledged gap** on other read paths (out of scope per PR body).

**[Octo-Q] verdict: APPROVE** — no P0/P1 findings. P2 items are documentation/cleanup suggestions.

yujiawei (approved):
# Code Review — PR #469 (octo-server)

Re-reviewed at head `d478d048d5d032752c02fe34ee35081ad2591b1a` (the new commit since the prior `CHANGES_REQUESTED` round at `e4a4780`). This is a security-sensitive change to the DM-search authorization gate, so it was reviewed independently against the surrounding code, plus an adversarial second pass on the load-bearing invariants. Local `go build`, `gofmt`, `go vet`, and the full `messages_search` package tests pass.

## Spec compliance

The new commit does exactly one thing, and it is the right thing: it resolves the single blocking finding from the previous round.

- **Prior P1 fixed.** The other-user-bot friend branch no longer `return true`s early after the friend check. It now falls through to the existing Step 3 bidirectional blacklist gate (`authz.go`), so a friend-bot that has since been blacklisted in either direction is hidden from search again — restoring the pre-Space behavior. The `if/else` restructuring (real-user path wrapped in `else`) is correct: only own-bots early-return; both other-bots and real users reach Step 3.
- **No over-build / under-build.** Only the blacklist fall-through, its doc comments, and the matching tests changed. The own-bot path still correctly skips blacklist (you cannot meaningfully blacklist your own bot). No new query surface; `QueryPeerRobotInfo` / `AreSpaceMembers` still delegate to the single `PinnedDB` SQL source.
- **Test coverage strengthened.** `TestCheckChannelAccess_P2POtherBotFriendAllowed` now asserts the friend-bot path consults blacklist bidirectionally (`blCalls == 2`) and skips Space (`spaceCalls == 0`); a new `TestCheckChannelAccess_P2POtherBotFriendBlacklisted` adds both `blocked_by_me` and `blocked_by_peer` denial cases. This is the regression guard the prior review asked for.

**Spec: compliant.**

## Items verified clean (independent re-derivation)

- **Blacklist fall-through (the fix itself).** Own-bot returns early before Step 3; other-bot friend and real-user both reach the two `ExistBlacklist` checks; every blacklist hit (either direction) → `NOT_FOUND`. No path skips the gate or the deny.
- **`space_id` trust boundary.** `GetSpaceID(c)` is context-only, set exclusively by `SpaceMiddleware` (mounted after auth on `/v1/messages`), which aborts with Forbidden before populating `space_id` unless the caller is a verified active member. No request-header forgery path. `AreSpaceMembers` → `CheckBothMembers` requires `COUNT(DISTINCT uid) == 2` over `space_member` joined to an **active** space (`s.status = 1`) with both rows `status = 1`, so Space-of-one, disbanded-Space, and cross-Space over-grant are all unreachable. `loginUID` comes from the authenticated session, and the per-DM participant binding is `GetFakeChannelIDWith(loginUID, peerUID)` — so this gate only decides whether the caller may open *their own* DM with `peer`; it cannot expose a non-participant's DM.
- **Fail-closed.** All six DB-error branches (`QueryPeerRobotInfo`, `AreSpaceMembers`, `IsFriend` ×2, `ExistBlacklist` ×2) return `false` + `respondInternal`. A Space-lookup error never silently downgrades to the friend table (covered by `AreSpaceMembersErrorFailsClosed`).
- **Anti-enumeration.** Every denial uniformly renders `NOT_FOUND / resource=channel`; no 401/403 status leakage.
- **Disabled-bot edge.** `QueryPeerRobotInfo`'s `LEFT JOIN robot ... AND r.status = 1` yields `isRobot=true, creator=""` for a disabled bot, routing it to the **stricter** other-bot (friend + blacklist) path rather than the Space path. `creator == ""` can never match a non-empty authenticated `loginUID`. Empty `peerUID` returns `(false, "", nil)` and is denied on the real-user path.

## Notes (non-blocking, P2 — informational only)

- **Bot–Space exclusion is by design.** Other-user bots never consult `AreSpaceMembers` (bots have no `space_member` row, documented at the call site). If the platform later introduces Space-bound enterprise bots reachable without a friend row, search will 404 for them — fold into the already-planned read-path unification follow-up; not a defect today.
- **`checkP2PAccess` writes directly to the HTTP context.** Fine for the current single-channel `_search*` endpoints (each gates one `channel_id`). If this is ever reused inside a multi-channel/global-filter loop, the direct `respondNotFound`/`respondInternal` would abort the whole response instead of filtering one channel — worth keeping in mind for that refactor.
- **`user/service.go` gofmt noise.** ~30 lines of import-ordering + struct-field realignment unrelated to the logic change; cosmetic, matches the PR description's disclaimer. Optional to split for cleaner `git blame`.

## Verdict

The one blocking P1 from the previous round is correctly and minimally fixed, with a targeted regression test. The Space-mode fix, trust boundary, fail-closed semantics, SQL correctness, and anti-enumeration all hold under independent and adversarial re-derivation. Remaining items are P2/informational and do not block merge.

Approving.

PR #475, opened 2026-06-26
feat(incomingwebhook): return localized adapter examples

an9xyz:
## Summary
- Add an incoming-webhook adapter catalog that drives public adapter URLs and onboarding examples from one source.
- Return localized `adapter_examples` on create/regenerate responses while keeping list responses free of token-bearing URLs.
- Add zh-CN/en-US adapter onboarding templates and update incoming webhook docs.

## Tests
- `go test ./modules/incomingwebhook -run 'TestCreate_ReturnsLocalizedAdapterExamples|TestAdapterExamplesRenderSupportedLanguages|TestAdapterExamplesAreLocalized|TestPublicURLs_HasMulticaEntry'`\n- `go test ./modules/incomingwebhook`

### DIFF
--- a/modules/incomingwebhook/README.md
+++ b/modules/incomingwebhook/README.md
@@ -355,9 +355,14 @@ POST /v1/incoming-webhooks/:webhook_id/:token/feishu
 
 需登录态 + 群管理员权限，路径前缀 `/v1/groups/:group_no/incoming-webhooks`。
 
-创建 / 重置（regenerate）响应除历史的 `url`（native 路径）外，还带 `urls` 对象，
-按推送形态给出全部路径（`native` / `github` / `wecom` / `multica` / `gitlab` / `feishu`，
-不含 host，由前端拼接）。token 仅在这两处出现一次，list 不回显 token、也不回推送 URL。
+创建 / 重置（regenerate）响应除历史的 `url`（native 路径）外，还带：
+
+- `urls`：按推送形态给出全部路径（`native` / `github` / `wecom` / `multica` / `gitlab` /
+  `feishu`），不含 host，由前端拼接。
+- `adapter_examples`：面向管理端「更多适配器」区域的本地化接入示例（title /
+  description / steps / auth / content_type），URL 与 `urls` 共享同一组 path。
+
+token 仅在这两处出现一次，list 不回显 token、推送 URL 或 `adapter_examples`。
 
 除创建/列出/更新/删除/重置外，Phase 2 新增两个：
 
--- a/modules/incomingwebhook/adapter_catalog.go
+++ b/modules/incomingwebhook/adapter_catalog.go
@@ -1,0 +1,127 @@
+package incomingwebhook
+
+import (
+	"embed"
+	"fmt"
+	"strings"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/modules/base/common/msgtmpl"
+	octoi18n "github.com/Mininglamp-OSS/octo-server/pkg/i18n"
+	"go.uber.org/zap"
+)
+
+//go:embed templates
+var adapterTemplatesFS embed.FS
+
+var adapterMessages = msgtmpl.MustNew(adapterTemplatesFS, "templates")
+
+type adapterAuthResp struct {
+	Type        string `json:"type"`
+	Header      string `json:"header,omitempty"`
+	ValueSource string `json:"value_source,omitempty"`
+}
+
+type adapterExampleResp struct {
+	Key         string          `json:"key"`
+	Title       string          `json:"title"`
+	Description string          `json:"description"`
+	URL         string          `json:"url"`
+	ContentType string          `json:"content_type"`
+	Auth        adapterAuthResp `json:"auth"`
+	Steps       []string        `json:"steps"`
+}
+
+type publicAdapterDef struct {
+	Key         string
+	Suffix      string
+	ContentType string
+	Auth        adapterAuthResp
+	ShowExample bool
+}
+
+var publicAdapterCatalog = []publicAdapterDef{
+	{Key: adapterNative, ContentType: "application/json"},
+	{Key: adapterGitHub, Suffix: adapterGitHub, ContentType: "application/json", Auth: adapterAuthResp{Type: "url_token"}, ShowExample: true},
+	{Key: adapterGitLab, Suffix: adapterGitLab, ContentType: "application/json", Auth: adapterAuthResp{Type: "url_token_and_header", Header: "X-Gitlab-Token", ValueSource: "token"}, ShowExample: true},
+	{Key: adapterFeishu, Suffix: adapterFeishu, ContentType: "application/json", Auth: adapterAuthResp{Type: "url_token"}, ShowExample: true},
+	{Key: adapterMultica, Suffix: adapterMultica, ContentType: "application/json", Auth: adapterAuthResp{Type: "url_token"}, ShowExample: true},
+	{Key: adapterWeCom, Suffix: adapterWeCom, ContentType: "application/json", Auth: adapterAuthResp{Type: "url_token"}, ShowExample: true},
+}
+
+func publicAdapterExamples(webhookID, token, lang string) ([]adapterExampleResp, error) {
+	urls := publicURLs(webhookID, token)
+	examples := make([]adapterExampleResp, 0, len(publicAdapterCatalog)-1)
+	for _, def := range publicAdapterCatalog {
+		if !def.ShowExample {
+			continue
+		}
+		data := map[string]any{
+			"URL":         urls[def.Key],
+			"ContentType": def.ContentType,
+		}
+		title, err := adapterMessages.Render(adapterTemplateName(def.Key, "title"), lang, data)
+		if err != nil {
+			return nil, err
+		}
+		description, err := adapterMessages.Render(adapterTemplateName(def.Key, "description"), lang, data)
+		if err != nil {
+			return nil, err
+		}
+		rawSteps, err := adapterMessages.Render(adapterTemplateName(def.Key, "steps"), lang, data)
+		if err != nil {
+			return nil, err
+		}
+		examples = append(examples, adapterExampleResp{
+			Key:         def.Key,
+			Title:       strings.TrimSpace(title),
+			Description: strings.TrimSpace(description),
+			URL:         urls[def.Key],
+			ContentType: def.ContentType,
+			Auth:        def.Auth,
+			Steps:       splitAdapterSteps(rawSteps),
+		})
+	}
+	return examples, nil
+}
+
+func renderPublicAdapterExamples(w *IncomingWebhook, c *wkhttp.Context, webhookID, token string) []adapterExampleResp {
+	examples, err := publicAdapterExamples(webhookID, token, incomingWebhookResponseLanguage(c))
+	if err != nil {
+		w.Error("render incoming webhook adapter examples failed", zap.Error(err))
+		return []adapterExampleResp{}
+	}
+	return examples
+}
+
+func adapterTemplateName(key, part string) string {
+	return fmt.Sprintf("adapter.%s.%s", key, part)
+}
+
+func splitAdapterSteps(s string) []string {
+	lines := strings.Split(strings.TrimSpace(s), "\n")
+	out := make([]string, 0, len(lines))
+	for _, line := range lines {
+		line = strings.TrimSpace(line)
+		if line != "" {
+			out = append(out, line)
+		}
+	}
+	return out
+}
+
+func incomingWebhookResponseLanguage(c *wkhttp.Context) string {
+	def, err := octoi18n.DefaultLanguageFromEnv()
+	if err != nil {
+		def = octoi18n.DefaultLanguage
+	}
+	if c == nil || c.Request == nil {
+		return def
+	}
+	if decision, ok := octoi18n.LanguageFromContext(c.Request.Context()); ok {
+		return decision.Language
+	}
+	return octoi18n.NegotiateLanguage(c.Request, octoi18n.LanguageNegotiationOptions{
+		DefaultLanguage: def,
+	}).Language
+}
--- a/modules/incomingwebhook/adapter_catalog_test.go
+++ b/modules/incomingwebhook/adapter_catalog_test.go
@@ -1,0 +1,43 @@
+package incomingwebhook
+
+import (
+	"testing"
+
+	octoi18n "github.com/Mininglamp-OSS/octo-server/pkg/i18n"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAdapterExamplesRenderSupportedLanguages(t *testing.T) {
+	for _, lang := range octoi18n.SupportedLanguages() {
+		t.Run(lang, func(t *testing.T) {
+			examples, err := publicAdapterExamples("iwh_abc", "deadbeef", lang)
+			require.NoError(t, err)
+
+			keys := make([]string, 0, len(examples))
+			urls := publicURLs("iwh_abc", "deadbeef")
+			for _, ex := range examples {
+				keys = append(keys, ex.Key)
+				assert.NotEmpty(t, ex.Title, "title for %s/%s", lang, ex.Key)
+				assert.NotEmpty(t, ex.Description, "description for %s/%s", lang, ex.Key)
+				assert.NotEmpty(t, ex.Steps, "steps for %s/%s", lang, ex.Key)
+				assert.Equal(t, "application/json", ex.ContentType)
+				assert.Equal(t, urls[ex.Key], ex.URL, "example URL must stay in sync with urls[%s]", ex.Key)
+			}
+			assert.Equal(t, []string{"github", "gitlab", "feishu", "multica", "wecom"}, keys)
+		})
+	}
+}
+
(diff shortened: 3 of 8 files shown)


Jerry-Xin (approved):
This PR is in scope for octo-server and cleanly extends the incoming-webhook module with localized adapter onboarding examples. **APPROVE.**

## ✅ Highlights
- **i18n design is solid.** Localized strings come from `msgtmpl` catalogs (`templates/{en-US,zh-CN}/adapter.tmpl`) loaded via `MustNew`, which enforces a completeness matrix at init: every supported language must define every key, else it panics at startup. Missing-translation is caught at build/init, not silently at runtime.
- **Locale resolution + fallback is safe.** `incomingWebhookResponseLanguage` (`modules/incomingwebhook/adapter_catalog.go`) prefers the negotiated `LanguageFromContext`, else `NegotiateLanguage(Accept-Language)`, with `DefaultLanguageFromEnv()` → `octoi18n.DefaultLanguage` (zh-CN) as default. `Catalog.treeFor` falls back to the source language (en-US) for unsupported/empty tags and never returns nil — an unknown locale degrades gracefully, no panic.
- **No injection / no secret leak via templates.** Templates interpolate only `.URL` (server-built path) and `.ContentType` (static constant `"application/json"`). No user-controlled input (webhook name, etc.) is interpolated; `missingkey=error` makes a forgotten param fail loud. The example `URL` equals `urls[key]`, which already carries the webhook token by design — and `adapter_examples`, like `urls`, is only returned on token-bearing create/regenerate responses; the list path still returns only `webhookResp`. No new exposure surface.
- **Map-lookup safety.** `publicAdapterCatalog` is a package-level slice iterated read-only; no nil-map writes, no missing-key panics, no concurrent mutation.
- **Tests verified.** Pure-unit localization tests pass locally: `TestAdapterExamplesRenderSupportedLanguages`, `TestAdapterExamplesAreLocalized`, `TestPublicURLs_HasMulticaEntry`. `go build ./modules/incomingwebhook/...` and `go vet` are clean.

## 💬 Non-blocking
- 🟡 **Regenerate coverage gap.** `adapter_examples` is added to both create and regenerate (`api.go` `regenerate` builds its own `createResp`), but the new integration test `TestCreate_ReturnsLocalizedAdapterExamples` only exercises create. A focused regenerate test would pin token/URL/example sync + localization on that path too.
- 🔵 **Catalog vs route drift.** Advertised URLs now derive from `publicAdapterCatalog` (`publicURLs`), but actual route registration remains hard-coded elsewhere in `api.go`. Fine for this PR; a small alignment test would prevent future adapters from drifting between catalog and routes.

## Verification note
The DB-backed `TestCreate_ReturnsLocalizedAdapterExamples` could not execute in my local workspace due to a pre-existing, unrelated migration-state error (`unknown migration in database`) that also fails untouched DB tests in this package (e.g. `TestPush_GitHubPing_SkippedAndAudited`). The new test binary compiles cleanly; this is an environment issue, not a defect in this PR, and is expected to pass in CI.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#475 Review Report
**PR:** Mininglamp-OSS/octo-server#475
**Head SHA:** `3891421689eebb0affd1adf1d05ba1e7bd3139e2`
**Title:** feat(incomingwebhook): return localized adapter examples

---

## 1. 验证结论

| 项 | 结论 | 证据 |
|---|---|---|
| Token exposure unchanged | ✅ | `adapter_examples` only in `createResp` returned by `create` (api.go:841) and `regenerate` (api.go:1093). `list` (api.go:858) returns `[]webhookResp` — no token, no adapter_examples. |
| publicURLs refactor parity | ✅ | Old hardcoded map (6 keys: native/github/wecom/Octo-Q/gitlab/feishu) vs new catalog-driven map — same key set, same values. `adapterNative` (Suffix="") → `urls[def.Key] = base`; others → `base + "/" + def.Suffix` where Suffix==Key. Existing `TestCreate_ReturnsAdapterURLs` + `TestPublicURLs_HasMulticaEntry` cover this. |
| Template completeness | ✅ | `msgtmpl.MustNew` enforces at init-time that every `{{define}}` name exists in all supported languages (en-US, zh-CN). Both templates define all 15 names (5 adapters × 3 parts). Missing name → `MustNew` panics at startup. |
| Template security | ✅ | `//go:embed templates` — compile-time fixed. Template data only contains `ContentType` (hardcoded "application/json") and `URL` (server-generated path). No user-controlled input. `text/template` (not html/template) is correct for JSON output. |
| Error handling | ✅ | `renderPublicAdapterExamples` (adapter_catalog.go:90) catches render errors, logs via `w.Error`, returns `[]adapterExampleResp{}` — graceful degradation without failing create/regenerate. |
| i18n language resolution | ✅ | `incomingWebhookResponseLanguage` (adapter_catalog.go:107) checks: context-injected decision (from middleware chain) → `NegotiateLanguage` (query param → cookie → Accept-Language → default). Nil-safe on `c` and `c.Request`. Test uses `?lang=en-US` query param (priority 50 in negotiation chain). |
| Auth metadata accuracy | ✅ | GitLab `adapterAuthResp{Type: "url_token_and_header", Header: "X-Gitlab-Token", ValueSource: "token"}` matches actual server behavior: `verifyGitLabToken` (adapter_gitlab.go:76-79) does constant-time comparison of X-Gitlab-Token vs URL token. |
| Test coverage | ✅ | Unit: `TestAdapterExamplesRenderSupportedLanguages` (all langs, keys, fields, URL sync), `TestAdapterExamplesAreLocalized` (zh-CN vs en-US differ). Integration: `TestCreate_ReturnsLocalizedAdapterExamples` (full HTTP handler path, not hand-fed data). |

## 2. 发现的问题

### P2-1: Unused `URL` key in template data map
- **File:** `adapter_catalog.go:62`
- **Severity:** P2 / nit
- The `data` map passes `"URL": urls[def.Key]` to `adapterMessages.Render(...)`, but **no template references `{{.URL}}`**. All templates refer to "the URL" / "Payload URL" as prose instructions, while the actual URL value is delivered via the `adapterExampleResp.URL` struct field (populated from `urls[def.Key]` at line 82). This is by-design separation of instructional text from the data value, but the unused key in the data map may confuse future maintainers.
- **Suggestion:** Either remove `"URL"` from the data map (if templates should stay text-only), or add `{{.URL}}` to template text where the URL value should appear inline. Currently it's dead data.

### P2-2: Whitespace-only changes in model.go
- **File:** `model.go:59-61`
- **Diff-scope:** pre-existing (cosmetic alignment)
- **Severity:** P2 / nit
- The diff shows tab/space realignment for `MentionUids`, `LastUsedAt`, `CallCount` fields — no semantic change. Consider separating formatting changes into a dedicated commit for cleaner git history.

## 3. 建议

1. **P2-1:** Remove the unused `"URL"` key from the template data map, or document with a one-line comment why it's reserved for future template use.
2. **P2-2:** Separate whitespace-only changes into their own commit (optional, not blocking).

## 4. 额外发现

- The `adapterNative` entry in `publicAdapterCatalog` has `ShowExample: false` (zero-value default). This is correct — native doesn't need a separate onboarding example since it's the default endpoint. The exclusion is implicit via Go zero-value rather than explicit. A comment `// native is the default; no separate example needed` would help readability.
- The `adapterAuthResp` zero-value for native (`{Type:"", Header:"", ValueSource:"}`) is safe because native is excluded from examples, but if `ShowExample` were ever flipped to `true`, the auth would serialize as `{type:""}` — potentially confusing API output. Low risk.

## 5. 数据流回溯

|---|---|---|
| `urls[def.Key]` in `publicAdapterExamples` | `publicURLs(webhookID, token)` → iterates `publicAdapterCatalog`, builds `base + "/" + Suffix` | ✅ All 6 adapter keys present; `adapterNative` (Suffix="") → base only |
| `adapterExampleResp.URL` in create/regenerate response | `urls[def.Key]` from same `publicURLs` call | ✅ Same source as `createResp.URLs`, values stay in sync (test asserts parity) |
| `incomingWebhookResponseLanguage(c)` → lang string | `octoi18n.LanguageFromContext(c.Request.Context())` → middleware-injected decision, or `NegotiateLanguage` fallback | ✅ Test uses `?lang=en-US` query param → priority 50 in negotiation chain → resolves to "en-US" |
| Template render data `ContentType` | `def.ContentType` = hardcoded `"application/json"` for all catalog entries | ✅ No user input; constant value |
| Template render data `URL` | `urls[def.Key]` from `publicURLs` | ⚠️ Passed to template but never referenced by any `{{.URL}}` — dead data (P2-1) |
| `adapterMessages` catalog | `msgtmpl.MustNew(adapterTemplatesFS, "templates")` at package init | ✅ `//go:embed` compiles templates in; `MustNew` panics on missing/incomplete templates |
| `createResp.AdapterExamples` in create handler | `renderPublicAdapterExamples(w, c, m.WebhookID, token)` at api.go:841 | ✅ `m.WebhookID` and `token` are freshly generated in the same handler scope |
| `createResp.AdapterExamples` in regenerate handler | `renderPublicAdapterExamples(w, c, webhookID, token)` at api.go:1093 | ✅ `webhookID` from path param, `token` freshly generated |
| `list` response | Returns `[]webhookResp` (not `createResp`) | ✅ No adapter_examples, no token — no leakage |

## 6. 盲点 checklist（R5 — security_sensitive PR, C1–C6 全项必填）

### C1 双路径 parity（含守卫/lint 工具自身等价写法覆盖）
- **create ↔ regenerate:** Both handlers add `AdapterExamples: renderPublicAdapterExamples(w, c, webhookID, token)` to `createResp`. ✅ Parity maintained.
- **list (excluded):** `list` returns `map[string]interface{}{"list": []webhookResp{}}` — `webhookResp` does NOT embed `createResp`, so `adapter_examples` cannot leak. ✅ By-design exclusion.
- **publicURLs refactor:** Old hardcoded map → new catalog-driven map. Same 6 keys, same values. `TestCreate_ReturnsAdapterURLs` + `TestPublicURLs_HasMulticaEntry` guard parity. ✅
- **update/delete/deliveries/testPush:** None return `createResp`. ✅ No adapter_examples exposure.

### C2 control-flow ordering / 嵌套复用 + 安全控件非 canonical 试穿
- `publicAdapterExamples` calls `publicURLs` once, iterates catalog once. No double-action or ordering risk.
- `renderPublicAdapterExamples` is called exactly twice (create, regenerate) with identical call pattern. No nesting or reuse hazard.
- Template rendering: `msgtmpl` uses `missingkey=error`, so any undefined data key would hard-error. The data map only has `URL` (unused) and `ContentType` (used). No risk of silent wrong output.
- `splitAdapterSteps` splits on `\n`, trims each line, drops empties. No injection surface (templates are compile-time embedded).

### C3 授权边界 ≠ 能力边界（谁能到达 tool）
- `adapter_examples` is only constructed inside `create` and `regenerate` handlers — both gated by `AuthMiddleware` + `resolveActor` (group member check) + `requireOwnership` (for regenerate). No new endpoint or capability exposed.
- `publicAdapterExamples` is a pure function with no authorization logic — correctly called only from authorized handlers.
- ✅ No authorization boundary change.

### C4 授权生命周期 / 容器-成员状态级联
- N/A — no authorization logic changed. Token generation and webhook lifecycle unchanged.

### C5 build/note 通过 ≠ 运行期路径正确
- `//go:embed templates` ensures templates are compiled into the binary at build time. ✅
- `msgtmpl.MustNew` panics at init time if template files are missing, incomplete, or have a language gap. This is a startup-time gate — a defective template set prevents the binary from running. ✅
- Template directory structure (`templates/en-US/adapter.tmpl`, `templates/zh-CN/adapter.tmpl`) matches what `MustNew` expects (`{root}/{lang}/*.tmpl`). ✅
- Runtime path: `publicAdapterExamples` → `adapterMessages.Render(name, lang, data)` → `treeFor(lang)` resolves to correct language tree (fallback to source language if unsupported) → `Lookup(name)` finds the `{{define}}` block → executes with data. If any step fails, error propagates to `renderPublicAdapterExamples` which gracefully degrades. ✅

### C6 治理/策略/安全文档自洽性
- README.md updated to document `adapter_examples`. States: "token 仅在这两处[create/regenerate]出现一次，list 不回显 token、推送 URL 或 `adapter_examples`" — consistent with code behavior. ✅
- No conflict with SECURITY.md or disclosure policies (this is module-level documentation, not governance/security policy).

## 7. 跨轮 blocker 复检

---

[Octo-Q] verdict: **APPROVE**

No P0 or P1 issues found. The PR correctly adds localized adapter examples to create/regenerate responses without expanding token exposure (list remains clean). The `publicURLs` refactor from hardcoded map to catalog-driven map preserves behavioral parity. Template infrastructure (`//go:embed` + `msgtmpl.MustNew`) provides compile-time + init-time safety. Two P2 nits noted (unused `URL` key in template data, whitespace-only model.go changes) — neither blocking.

yujiawei (approved):
# Code Review — PR #475 (octo-server)

**Verdict: APPROVED** — no P0/P1 blockers. A few P2 maintainability/robustness suggestions below (none gate merge). Reviewed against head SHA `3891421689eebb0affd1adf1d05ba1e7bd3139e2`; `go build` and `go test ./modules/incomingwebhook/` both pass locally.

## Spec compliance

The change does what the summary states and nothing more:
- Adds `adapter_catalog.go` (catalog + localized example rendering) and returns `adapter_examples` on **create** and **regenerate** only (`api.go` create ~L843, regenerate ~L1095).
- `list()` (`api.go` L853-874) still returns only `webhookResp` via `toResp()`, which carries no token / URL / examples (`model.go` L166-185). No missing or extra surface.
- zh-CN / en-US onboarding templates and README updated consistently.

No scope creep, no missing items.

## Security review (this PR is security-sensitive: token-bearing URLs)

The key concern is that `adapter_examples[].url` embeds the webhook token (`publicURL` = `/v1/incoming-webhooks/<id>/<token>...`, `api.go` L482-484). This was traced exhaustively and the boundary holds:

1. **No new exposure path.** `adapter_examples` is populated at exactly two call sites — `create()` and `regenerate()` — both already gated by `AuthMiddleware` + `requireMgmtEnabled` + `resolveActor` + active-group + ownership checks, and both already return `Token`, `URL`, and the `URLs` map in the same body. The token appears nowhere it wasn't already disclosed in the identical response. Net token-exposure surface is **unchanged**.
2. **Not leaked on read paths.** `list()` and the other read endpoints never return `createResp`; `webhookResp`/`toResp()` carry no secret.
3. **Not leaked via logs.** The render error path logs only `zap.Error(err)`; `msgtmpl.Render` error strings embed only the template name + lang (and, under `missingkey=error`, a missing *key name* — never a data *value*). Templates reference only `{{.ContentType}}`, never `{{.URL}}`, so the token cannot reach a log line.

No token leakage is introduced. Worth a human sanity-check on the same axis, but I found nothing actionable.

## Suggestions (P2 — non-blocking)

1. **`adapter_catalog.go` L54 — brittle slice capacity.** `make([]adapterExampleResp, 0, len(publicAdapterCatalog)-1)` hardcodes the assumption "exactly one non-`ShowExample` entry (native)". It is only a capacity hint so it cannot misbehave today, but if the catalog ever shrank to empty, `len-1 = -1` would panic (`makeslice: cap out of range`). Counting `ShowExample == true` entries (or just using `len(publicAdapterCatalog)`) removes the hidden invariant.

2. **`adapter_catalog.go` — render errors silently degrade to `[]`.** `renderPublicAdapterExamples` logs and returns an empty array on any render error, so create/regenerate still return 200 with token + urls but an empty `adapter_examples` — at the only moment the client can ever see these URLs. This path is effectively unreachable in practice (the `msgtmpl` completeness matrix is enforced at startup, all referenced keys are always supplied, and unsupported languages fall back to the source language), so it is a robustness nit rather than a live bug. Optional: render per-adapter and `continue` on a single failure so one broken template degrades only its own entry instead of all of them.

3. **`adapter_catalog.go` — `MustNew` panics if a new global language is added without adapter templates.** `msgtmpl.MustNew` iterates `i18n.SupportedLanguages()` and enforces a per-language completeness matrix; only `en-US` and `zh-CN` templates ship here. Correct today (matrix is exactly those two), but adding a third supported language globally would make this package's init panic until matching `adapter.tmpl` files are added. Consider a one-line comment pinning this coupling.

4. **`adapter_catalog.go` `incomingWebhookResponseLanguage` — reimplements an existing idiom.** The same file already resolves outbound language via `i18n.OutboundLanguage(ctx)` (`api.go` testPushMessage ~L1114). The new helper hand-rolls equivalent logic plus a `NegotiateLanguage` fallback that omits trusted-header / proxy / user-language options. In production the fallback branch is unreachable (the global `EarlyMiddleware` always seats a decision in ctx), so behavior is correct — but the divergent re-negotiation is unnecessary; collapsing to `i18n.OutboundLanguage(c.Request.Context())` (guarded for nil `c`/`c.Request`) would match the neighbor and drop the divergence.

5. **`adapter_catalog.go` L59-62 — dead `URL` key in template data.** The data map passes both `URL` and `ContentType`, but no template body references `{{.URL}}` (the example URL is set directly from the struct). Harmless under `missingkey=error` (it only errors on referenced-but-missing keys), but slightly misleading. Drop the key or add a note.

## Coverage / blind spots

Verified statically: all `adapter_examples` / `publicURLs` / `createResp` call sites; create/regenerate/list/update/deliveries auth gating; adapter-key constants matching template names; `publicURLs` refactor key-set parity (byte-for-byte identical to the prior literal — `TestCreate_ReturnsAdapterURLs` and `TestCreate_ReturnsPhase4AdapterURLs` pin github/wecom/multica/gitlab/feishu). Not exercised: DB-backed integration paths and access-log middleware behavior outside this module — recommended for the human security pass given the token-bearing nature of the response.

PR #477, opened 2026-06-26
feat(messages_search): project richtext (type=14) detail from payloadRaw

### LINKED ISSUE #476: feat(messages_search): project richtext (type=14) detail from payloadRaw
### Problem / Motivation

`/v1/messages/_search`, `_search_all`, and `_search_around` currently surface
type=14 rich-text messages as a single text `snippet`. The front-end has no way
to render the original structured payload (text blocks interleaved with images
and files, `@` mentions) the way `api/v1/message/channel/sync` does. As a
result, search hits that point to rich-text messages render as plain text and
lose images / files / mention highlights, breaking parity with the channel
timeline view.

This is a regression in UX expectation rather than functionality: octo-web
already has the renderer (`useRichTextMessageUI` → `MixedContent`), the
indexer already preserves the full original payload, and the search query
already pulls the full `_source`. The only missing piece is that octo-server's
`MessageHit` does not project the rich-text shape out of `_source.payloadRaw`.

### Proposed Solution

Add an additive `rich_text` field on `MessageHit` (and the equivalent hits in
`_search_all` / `_search_around`) that exposes the typed projection of the
indexed `_source.payloadRaw` for type=14 hits. The schema mirrors octo-web's
authoritative `RichTextContent.ts` (and octo-lib `common/richtext.go`):

```jsonc
{
  "rich_text": {
    "content": [ { "type": "text"|"image"|"file", ... }, ... ],
    "plain": "...",
    "mention": { "entities": [...], "all": 0, "humans": 0, "ais": 0 }
  }
}
```

Behavior:

- `rich_text` is emitted **only** when `payload.type == 14` and `payloadRaw` is
  present in `_source`. Otherwise the field is omitted.
- `message_kind`, `snippet`, and every other existing wire field stay
  byte-identical. The front-end renders via `rich_text` when present and falls
  back to the existing `snippet` path when absent (legacy documents indexed
  before the indexer wrote `payloadRaw` keep working).
- No ES re-indexing required: the indexer already writes the raw payload to
  `_source.payloadRaw` (mapping is `enabled: false`, so it is preserved in
  `_source` without being indexed) and the search queries already fetch the
  full `_source`.

### Alternatives Considered

- **Pass-through `payload = doc.PayloadRaw`** (whole raw JSON in the
  response). Zero schema maintenance, but dumps every backend-internal field
  into the API surface. Rejected.
- **Reconstruct from MySQL.** Adds an N+1 lookup per hit and re-derives data
  that is already in OpenSearch. Rejected.

### Related Component

Backend (Server)

### Technical Approach (optional)

Three small edits in `modules/messages_search`:

1. `source.go`
   - Add `Doc.PayloadRaw json.RawMessage` (`json:"payloadRaw,omitempty"`).
   - Add typed projection structs: `RichTextBlock`, `RichTextMentionEntity`,
     `RichTextMention`, `RichTextDetail`.
   - Add `buildRichTextDetail(raw json.RawMessage) *RichTextDetail`:
     - empty raw / invalid JSON → `nil` (fail-soft)
     - `content` as array → `[]RichTextBlock`
     - `content` as bare string (legacy payloads) → normalized to a single
       `{Type:"text", Text:s}` block
     - `plain` and `mention` pass through verbatim
2. `search_messages.go`
   - Add `MessageHit.RichText *RichTextDetail`
     (`json:"rich_text,omitempty"`).
   - In `singleMessageHit`, after the existing assembly, project when
     `payloadType(doc.Payload) == payloadTypeRichText`. Sits next to the
     upstream `applyMediaProjection` call; the two are independent (media
     fires on type=2/5, richtext fires on type=14).
3. `swagger/api.yaml`
   - Add the `rich_text` property + four new schemas
     (`RichTextDetail`, `RichTextBlock`, `RichTextMention`,
     `RichTextMentionEntity`).

`_search_all` / `_search_around` inherit automatically because they call
through `singleMessageHit`.

No new dependencies, no ES query / DSL changes, no breaking changes to any
existing field.

### Additional Context

- octo-web schema source of truth:
  `packages/dmworkbase/src/Messages/RichText/RichTextContent.ts`
- octo-web renderer entry point:
  `packages/dmworkbase/src/bridge/message/useRichTextMessageUI.ts`
- octo-lib payload contract: `common/richtext.go`
- Indexer field guarantee:
  `wukongim-message-indexer/internal/transform/doc.go` (`doc.PayloadRaw = raw`)
- Legacy compatibility: documents indexed before `payloadRaw` was added simply
  return `rich_text: null` and continue using `snippet`. No backfill required;
  re-indexing optional.
- Implementing PR: <to be linked once opened>

BrDing-Rookie:
## Summary

Project type=14 rich-text messages into search hits so that
`/v1/messages/_search`, `_search_all`, and `_search_around` return the same
structured shape that `api/v1/message/channel/sync` does — letting octo-web
render images, files, and `@`-mention highlights inside search results
instead of falling back to plain `snippet`.

The data was always there — `wukongim-message-indexer` writes the full
original payload to `_source.payloadRaw` (mapping is `enabled: false`, so it
is preserved in `_source` but not indexed) and the search queries already
fetch the full `_source`. The only missing piece was a typed projection in
octo-server's `MessageHit` response shape. This PR adds it as a purely
additive field — no schema change to existing fields, no re-indexing, no ES
query changes, no new dependencies.

## Related Issue

#476 

## Linked Spec

Design doc (local): `/tmp/octo-server-richtext-search-design.md` (Option A —
typed projection).

## Changes

- **`modules/messages_search/source.go`**
  - `Doc.PayloadRaw json.RawMessage` — captures the existing
    `_source.payloadRaw` blob that was previously dropped by `Doc`'s
    unmarshal.
  - New types mirroring octo-web `RichTextContent.ts` /
    octo-lib `common/richtext.go`: `RichTextBlock`, `RichTextMentionEntity`,
    `RichTextMention`, `RichTextDetail`.
  - `buildRichTextDetail(raw)` — extracts `content` / `plain` / `mention`
    from `payloadRaw`. Handles three legacy/edge cases:
    - empty or invalid JSON → returns `nil` (fail-soft, snippet fallback)
    - `content` as array → typed `[]RichTextBlock`
    - `content` as bare string (older payloads) → normalized to a single
      `{Type:"text", Text:s}` block
- **`modules/messages_search/search_messages.go`**
  - `MessageHit.RichText *RichTextDetail` (`json:"rich_text,omitempty"`).
  - `singleMessageHit` invokes `buildRichTextDetail(doc.PayloadRaw)` only
    when `payloadType(doc.Payload) == payloadTypeRichText`. Independent of
    the existing `applyMediaProjection` (media fires on type=2/5,
    rich-text fires on type=14).
  - `_search_all` and `_search_around` inherit the projection automatically
    because they call through `singleMessageHit`.
- **`modules/messages_search/swagger/api.yaml`**
  - Added `MessageHit.rich_text` property.
  - Added 4 new schemas: `RichTextDetail`, `RichTextBlock`, `RichTextMention`,
    `RichTextMentionEntity`.
- **`modules/messages_search/richtext_projection_test.go`** (new)
  - `buildRichTextDetail`: array `content`, legacy string `content`, empty
    raw, invalid JSON, with `mention`.
  - `singleMessageHit`: type=14 + `payloadRaw` → populated `rich_text`;
    type=14 without `payloadRaw` → `nil` + `snippet` fallback unchanged;
    non-rich-text types → no `rich_text` emitted.
  - `omitempty` serialization verified.

Net diff: +433 / −1 across 4 files.

## Testing

```
go build ./...                           # clean
go vet ./...                             # clean
go test ./modules/messages_search/...    # ok ~0.043s
make i18n-extract-check                  # passed
make i18n-lint                           # OK: no new direct error responses
                                         # OK: no inline codes
```

- [x] Unit tests added (`richtext_projection_test.go`, 6 cases covering happy
      path, legacy string content, missing `payloadRaw`, invalid JSON,
      mention pass-through, and the `singleMessageHit` projection branch).
- [x] Build / vet / package tests green.
- [ ] Manual end-to-end against `dmwork-test/dmworkim-develop` (recommended
      before merge — search a known type=14 message and confirm
      `rich_text.content` matches `channel/sync`).

## COMPREHENSION

   On every search hit assembly (`/v1/messages/_search`, `_search_all`,
   `_search_around` all route through `singleMessageHit`) we do one extra
   `payloadType == payloadTypeRichText` check; on a hit we `json.Unmarshal`
   the already-fetched `payloadRaw` once into the typed projection struct.
   Net wire-format change: one additional `rich_text` object on rich-text
   hits; every other field is byte-identical. No new ES round-trips, no new
   DB round-trips, no DSL changes.

   - **Older documents** that pre-date the indexer writing `payloadRaw` →
     `rich_text` is `nil`, the field is omitted via `omitempty`, and the
     front-end falls back to the existing `snippet` path. No behavior
     change for legacy hits. (No backfill required; re-indexing optional.)
   - **Malformed `payloadRaw`** (shouldn't happen given indexer writes the
     raw bytes, but defensive): `buildRichTextDetail` returns `nil` → same
     fallback as above. No panic, no error propagated up the stack.
   - **Response size** grows on type=14 hits proportionally to the rich-text
     content (images / files contribute URLs, not bytes). Acceptable for
     the foreseeable hit-count cap; if it becomes a concern we can add an
     `include=rich_text` opt-in flag without breaking anything.
   - **Front-end** that doesn't yet know `rich_text`: ignores the field
     and continues using `snippet`. Strict additive change.
   - **Swagger consumers**: the four new schemas are additive; the existing
     `MessageHit` properties / `message_kind` enum are unchanged.

   - `TestBuildRichTextDetail_*` covers the parser branches (array content,
     legacy string content, missing raw, invalid JSON, mention).
   - `TestSingleMessageHit_RichText*` exercises the projection wiring: a
     type=14 hit with `payloadRaw` returns a populated `rich_text`; the same
     hit without `payloadRaw` returns `rich_text: nil` and the legacy
     `snippet` path is unaffected; non-rich-text hits never carry the
     field.
   - All packages still build and `go vet` clean; `make i18n-*` linters
     pass. Recommended manual verification: pick one real type=14 message
     in `dmwork-test`, hit `/v1/messages/_search`, and diff
     `rich_text.content` against the corresponding
     `api/v1/message/channel/sync` payload field-by-field.

## Checklist

- [x] Updated documentation (swagger)

Closes #476

### DIFF
--- a/modules/messages_search/richtext_projection_test.go
+++ b/modules/messages_search/richtext_projection_test.go
@@ -1,0 +1,199 @@
+package messages_search
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+)
+
+// TestBuildRichTextDetail_ArrayContent covers the modern shape: content is a
+// block array carrying text/image/file blocks plus a server-generated plain
+// string. All canonical block fields must round-trip onto the typed projection.
+func TestBuildRichTextDetail_ArrayContent(t *testing.T) {
+	raw := json.RawMessage(`{
+	  "type": 14,
+	  "content": [
+	    {"type": "text", "text": "标题"},
+	    {"type": "image", "url": "http://x/y.png", "width": 640, "height": 480, "name": "y.png", "caption": "图说"},
+	    {"type": "file", "url": "http://x/a.pdf", "name": "a.pdf", "size": 12345, "extension": "pdf", "mime": "application/pdf", "caption": "合同"}
+	  ],
+	  "plain": "标题[图片][文件]"
+	}`)
+	d := buildRichTextDetail(raw)
+	if d == nil {
+		t.Fatalf("expected non-nil detail")
+	}
+	if d.Plain != "标题[图片][文件]" {
+		t.Errorf("plain: got %q", d.Plain)
+	}
+	if got := len(d.Content); got != 3 {
+		t.Fatalf("blocks: got %d", got)
+	}
+	if d.Content[0].Type != "text" || d.Content[0].Text != "标题" {
+		t.Errorf("text block: %+v", d.Content[0])
+	}
+	img := d.Content[1]
+	if img.Type != "image" || img.URL != "http://x/y.png" || img.Width != 640 || img.Height != 480 || img.Name != "y.png" || img.Caption != "图说" {
+		t.Errorf("image block: %+v", img)
+	}
+	f := d.Content[2]
+	if f.Type != "file" || f.URL != "http://x/a.pdf" || f.Name != "a.pdf" || f.Size != 12345 || f.Extension != "pdf" || f.Mime != "application/pdf" || f.Caption != "合同" {
+		t.Errorf("file block: %+v", f)
+	}
+	if d.Mention != nil {
+		t.Errorf("mention should be nil when absent, got %+v", d.Mention)
+	}
+}
+
+// TestBuildRichTextDetail_LegacyStringContent: old payloads stored content as
+// a plain JSON string; the builder must normalise that into a single text
+// block so the renderer never sees the legacy shape.
+func TestBuildRichTextDetail_LegacyStringContent(t *testing.T) {
+	raw := json.RawMessage(`{"type":14,"content":"hello world","plain":"hello world"}`)
+	d := buildRichTextDetail(raw)
+	if d == nil {
+		t.Fatalf("nil detail")
+	}
+	if len(d.Content) != 1 {
+		t.Fatalf("blocks: got %d", len(d.Content))
+	}
+	if d.Content[0].Type != "text" || d.Content[0].Text != "hello world" {
+		t.Errorf("normalised text block: %+v", d.Content[0])
+	}
+	if d.Plain != "hello world" {
+		t.Errorf("plain: got %q", d.Plain)
+	}
+}
+
+// TestBuildRichTextDetail_EmptyRaw guards the legacy / pre-payloadRaw fail-soft
+// case: the projection must be nil so the caller falls back to snippet.
+func TestBuildRichTextDetail_EmptyRaw(t *testing.T) {
+	if d := buildRichTextDetail(nil); d != nil {
+		t.Errorf("nil raw: want nil detail, got %+v", d)
+	}
+	if d := buildRichTextDetail(json.RawMessage{}); d != nil {
+		t.Errorf("empty raw: want nil detail, got %+v", d)
+	}
+}
+
+// TestBuildRichTextDetail_InvalidJSON: corrupt envelopes must fail-soft (nil)
+// rather than surface a half-parsed detail or propagate the error.
+func TestBuildRichTextDetail_InvalidJSON(t *testing.T) {
+	if d := buildRichTextDetail(json.RawMessage(`not-json`)); d != nil {
+		t.Errorf("garbage: want nil, got %+v", d)
+	}
+	// A legacy string content that is itself unparseable as a string (e.g.
+	// content is a JSON number) must drop content but keep the envelope.
+	d := buildRichTextDetail(json.RawMessage(`{"content": 42, "plain": "p"}`))
+	if d == nil {
+		t.Fatalf("envelope-valid number-content: want non-nil")
+	}
+	if len(d.Content) != 0 {
+		t.Errorf("number content must yield empty blocks, got %+v", d.Content)
+	}
+	if d.Plain != "p" {
+		t.Errorf("plain should still pass through, got %q", d.Plain)
+	}
+}
+
+// TestBuildRichTextDetail_Mention transports the mention object verbatim, both
+// the entity slice and the three @-all flags.
+func TestBuildRichTextDetail_Mention(t *testing.T) {
+	raw := json.RawMessage(`{
+	  "content": [{"type":"text","text":"@张三 你好"}],
+	  "plain": "@张三 你好",
+	  "mention": {
+	    "entities": [{"uid":"u_zhang","offset":0,"length":3}],
+	    "all": 1, "humans": 1, "ais": 0
+	  }
+	}`)
+	d := buildRichTextDetail(raw)
+	if d == nil || d.Mention == nil {
+		t.Fatalf("expected mention, got %+v", d)
+	}
+	if len(d.Mention.Entities) != 1 {
+		t.Fatalf("entities: got %d", len(d.Mention.Entities))
+	}
+	e := d.Mention.Entities[0]
+	if e.UID != "u_zhang" || e.Offset != 0 || e.Length != 3 {
+		t.Errorf("entity: %+v", e)
+	}
+	if d.Mention.All != 1 || d.Mention.Humans != 1 || d.Mention.Ais != 0 {
+		t.Errorf("flags: %+v", d.Mention)
+	}
+}
+
+// TestSingleMessageHit_RichText covers the projection contract on the hit:
+//   - type=14 + non-empty payloadRaw → rich_text populated, snippet still set
+//   - type=14 + missing payloadRaw   → rich_text nil, snippet/text untouched
+//   - non-richtext type              → rich_text nil regardless of payloadRaw
+func TestSingleMessageHit_RichText(t *testing.T) {
+	var h Handler
+	rt := payloadTypeRichText
+	raw := json.RawMessage(`{"type":14,"content":[{"type":"text","text":"标题"}],"plain":"标题"}`)
+
+	doc := Doc{
+		MessageID:  101,
+		MessageSeq: 9,
+		Payload:    &Payload{Type: &rt, RichText: &RichTextPayload{SearchText: "标题"}},
+		PayloadRaw: raw,
+	}
+	mh := h.singleMessageHit(doc, "g", nil)
+	if mh.RichText == nil {
+		t.Fatalf("rich_text must be populated for type=14 + payloadRaw")
+	}
+	if len(mh.RichText.Content) != 1 || mh.RichText.Content[0].Text != "标题" {
+		t.Errorf("rich_text.content: %+v", mh.RichText.Content)
+	}
+	if mh.RichText.Plain != "标题" {
+		t.Errorf("rich_text.plain: %q", mh.RichText.Plain)
+	}
+	// Snippet fallback still runs — searchText feeds fallbackSnippet so the
+	// hit is renderable even on a client that ignores rich_text.
+	if mh.Snippet != "标题" {
+		t.Errorf("snippet fallback should still emit searchText, got %q", mh.Snippet)
+	}
+	if mh.MessageKind != "text" {
+		t.Errorf("kind should fold into text, got %q", mh.MessageKind)
+	}
+
+	// Legacy doc indexed before payloadRaw existed: fail-soft to nil.
+	docNoRaw := Doc{
+		MessageID: 102,
+		Payload:   &Payload{Type: &rt, RichText: &RichTextPayload{SearchText: "标题"}},
+	}
+	mhNoRaw := h.singleMessageHit(docNoRaw, "g", nil)
+	if mhNoRaw.RichText != nil {
+		t.Errorf("missing payloadRaw must yield nil rich_text, got %+v", mhNoRaw.RichText)
+	}
+	if mhNoRaw.Snippet != "标题" {
+		t.Errorf("snippet fallback must still fire on legacy doc, got %q", mhNoRaw.Snippet)
+	}
+
+	// Non-richtext docs must never carry rich_text, even if a stray payloadRaw
+	// blob is present on the source.
+	tp := payloadTypeText
+	docText := Doc{
+		MessageID:  103,
+		Payload:    &Payload{Type: &tp, Text: &TextPayload{Content: "hi"}},
+		PayloadRaw: raw,
+	}
+	mhText := h.singleMessageHit(docText, "g", nil)
+	if mhText.RichText != nil {
+		t.Errorf("non-richtext type must not project rich_text, got %+v", mhText.RichText)
+	}
+}
+
+// TestMessageHit_RichTextOmitemptyOnWire pins the wire contract: rich_text
+// must NOT appear on plain text hits (otherwise old clients see an unexpected
+// field on every response).
+func TestMessageHit_RichTextOmitemptyOnWire(t *testing.T) {
+	mh := MessageHit{MessageID: "1", MessageKind: "text", SenderID: "u"}
+	out, err := json.Marshal(mh)
+	if err != nil {
+		t.Fatalf("marshal: %v", err)
+	}
+	if strings.Contains(string(out), `"rich_text"`) {
+		t.Errorf("nil rich_text must be omitted on the wire, got %s", string(out))
+	}
+}
--- a/modules/messages_search/search_messages.go
+++ b/modules/messages_search/search_messages.go
@@ -36,6 +36,12 @@ type MessageHit struct {
 	Width           int            `json:"width,omitempty"`
 	Height          int            `json:"height,omitempty"`
 	DurationMs      int64          `json:"duration_ms,omitempty"`
+	// RichText is the typed projection of a payload.type=14 rich-text message,
+	// emitted only when the hit is rich-text AND the indexer preserved
+	// `_source.payloadRaw`. Older docs (pre-payloadRaw indexer) leave it nil,
+	// in which case the client renders the existing snippet/text fallback.
+	// Non-richtext hits never carry this field.
+	RichText *RichTextDetail `json:"rich_text,omitempty"`
 }
 
 func init() {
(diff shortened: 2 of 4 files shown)


Jerry-Xin (approved):
In scope for `octo-server`: this PR extends the existing message-search projection path and swagger contract for `modules/messages_search`.

💬 Non-blocking

🟡 Warning — `modules/messages_search/source.go` (`buildRichTextDetail`, array `content` branch): `json.Unmarshal` errors for array `content` are ignored, then a non-nil `RichTextDetail` is returned with `content:null` on the wire. Probably acceptable for corrupt legacy data, but it does not match the swagger `required` array contract. Consider fail-softing to `nil` or initializing `Content` to an empty slice for invalid array content.

🔵 Suggestion — `modules/messages_search/source.go` (`Doc.PayloadRaw` comment / surrounding `Doc` doc): the older `Doc` commentary framing `payloadRaw` as "not read to avoid hot-path parsing" is now stale — this PR intentionally reads it in `buildRichTextDetail`. Worth a one-line update so future maintainers aren't misled.

✅ Highlights

- Projection is correctly centralized in `singleMessageHit` (`modules/messages_search/search_messages.go`), so `_search`, `_search_all`, and `_search_around` inherit the behavior consistently.
- The change is additive on the response shape, guarded by `payload.type == 14` (`payloadTypeRichText`), and preserves the `snippet` fallback for older documents without `payloadRaw`.
- `buildRichTextDetail` fail-softs cleanly: `len(raw)==0` → nil, `json.Unmarshal` error → nil, and the `p.Content[0] == '['` byte-check is correctly guarded by `len(p.Content) > 0` — no nil-deref / index-out-of-range / unchecked type-assertion panic path on malformed/missing `payloadRaw`.
- Access control is unchanged: the projection reads the same hit's own `payloadRaw` inside `singleMessageHit`, which only runs on hits already passing the search visibility filter — no cross-trust-boundary leak.
- swagger/api.yaml is updated to match (the `MessageHit.rich_text` property + 4 new schemas), so the response contract does not drift.
- Tests are ~46% of the diff (199/433) and cover parser branches, legacy string content, missing/invalid `payloadRaw`, mention pass-through, the `singleMessageHit` projection wiring, and `omitempty` wire behavior. Verified locally: `go build ./...` clean and `go test ./modules/messages_search/...` passes (`ok ... 1.4s`).

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#477 Review Report — feat(messages_search): project richtext (type=14) detail from payloadRaw
**Head SHA:** 85d1e4592719b8e1c04502a86ba9f04074d254b5
**Files:** source.go (+104/-1), search_messages.go (+13), richtext_projection_test.go (new, +199), swagger/api.yaml (+117)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Build (`go build ./...`) | ✅ clean | exit 0 |
| Tests (`go test ./modules/messages_search/...`) | ✅ all pass | 0.053s |
| Data flow: `PayloadRaw` deserialized from `_source` | ✅ | `Doc.PayloadRaw json.RawMessage` ← `json.Unmarshal(rawSource(hit.Source), &doc)` — all three endpoints (_search, _search_all, _search_around) go through this path |
| Data flow: `singleMessageHit` invoked for all 3 endpoints | ✅ | `_search`:227, `_search_all`:257 (via `singleSearchAllHit`), `_search_around`:136 (via `buildMessageHits`) |
| Type gate: `payloadType(doc.Payload) == payloadTypeRichText` | ✅ | `search_messages.go:280`; `payloadType()` nil-safe at `source.go:232` |
| Fail-soft: empty/invalid `payloadRaw` → nil RichText | ✅ | `buildRichTextDetail` returns nil for `len(raw)==0` and `json.Unmarshal` error |
| `omitempty` wire contract: nil RichText absent from JSON | ✅ | `json:"rich_text,omitempty"` verified by `TestMessageHit_RichTextOmitemptyOnWire` |
| Non-richtext docs never get `rich_text` | ✅ | Type gate prevents `buildRichTextDetail` call; verified by test case 3 in `TestSingleMessageHit_RichText` |
| `message_kind` unchanged for type=14 | ✅ | `classifyKind` falls through to `"text"` for type=14; swagger enum `["text","forward"]` preserved |
| Snippet fallback independent of rich-text projection | ✅ | `fallbackSnippet` runs before `buildRichTextDetail` in `singleMessageHit`; uses `Payload.RichText.SearchText` |

## 2. Findings

### P2-1: Swagger `$ref` + sibling `description` violates OpenAPI 2.0

**Diff-scope:** new (introduced by this PR — no existing property in api.yaml uses this pattern)
**Severity:** P2 (swagger lint warning, no runtime impact)

`swagger/api.yaml:545-553` — `MessageHit.rich_text` carries `$ref` with a sibling `description:`. In OpenAPI 2.0, `$ref` must be the sole property; spec-compliant tools silently drop siblings. The `description` text will be invisible in generated docs.

All other `$ref` usages in this file (69 occurrences) are standalone — this PR introduces the pattern for the first time.

**Fix:** Move the description into the `RichTextDetail` definition's `description:` field (already partially there), or use `allOf`:
```yaml
rich_text:
  allOf:
    - $ref: "#/definitions/RichTextDetail"
  description: "..."
```

### P2-2: `RichTextBlock.Type` lacks `omitempty`

**Diff-scope:** new
**Severity:** P2 (cosmetic wire-shape inconsistency)

`source.go:241` — `Type string json:"type"` without `omitempty`. If a content block from `payloadRaw` lacks a `type` field, the projection emits `"type":""` on the wire. This is low-probability (indexer always writes `type`), but inconsistent with the otherwise defensive fail-soft design. The swagger marks `type` as `required`, so a missing type would already violate the schema.

**Fix:** Add `omitempty` or accept the trade-off (empty string is more visible for debugging than field absence).

### P2-3: `RichTextDetail.Content` may serialize as `null` in edge case

**Diff-scope:** new
**Severity:** P2 (theoretical; requires corrupt payload)

`source.go:283` — When the outer envelope parses but `content` is malformed JSON array, `_ = json.Unmarshal(p.Content, &d.Content)` silently fails and `d.Content` remains `nil`. Go's `json.Marshal` serializes nil slices as `null`, but swagger declares `content` as `required` (not nullable). In practice this is unreachable (indexer writes valid JSON to `payloadRaw`), and the `_search` endpoint's type whitelist [1,11,14] limits exposure.

**Fix:** Initialize `d.Content = []RichTextBlock{}` on array-unmarshal failure, or use `omitempty` on `Content` field and relax swagger `required`.

## 3. Suggestions

1. **Manual E2E test** (as noted in PR checklist): pick one real type=14 message in `dmwork-test`, hit `/_search`, and diff `rich_text.content` against `channel/sync` field-by-field — this is the only gap not covered by unit tests.
2. **Swagger lint in CI:** consider adding `spectral lint` or `swagger-cli validate` to the build pipeline to catch `$ref` sibling issues automatically.
3. **Response-size monitoring:** the PR correctly notes response growth is proportional to rich-text content. If type=14 hits become a significant fraction, an `include=rich_text` opt-in flag (mentioned in the PR body) is a clean follow-up.

## 4. Additional Observations

- The `buildRichTextDetail` legacy string-content normalization (bare string → single `{type:"text"}` block) is well-designed and test-covered.
- The `content[0] == '['` byte check is correct — `json.RawMessage` from `json.Unmarshal` is always compact (no leading whitespace).
- The rich-text projection is independent of `applyMediaProjection` (type=2/5), so there's no interaction risk with existing media hit rendering.
- `filterVisible` and virtual-doc filtering (`must_not virtual=true`) operate upstream of `singleMessageHit` and are unaffected by this change.

## 5. Data-Flow Tracing

| Consumed data | Upstream source | Flows to consumption? |
|---|---|---|
| `doc.PayloadRaw` | `json.Unmarshal(rawSource(hit.Source), &doc)` → `Doc.PayloadRaw json.RawMessage` | ✅ `_source.payloadRaw` (OS mapping `enabled:false`) preserved by indexer, deserialized into Doc |
| `doc.Payload` | Same Unmarshal → `Doc.Payload *Payload` | ✅ Structured payload fields remain authoritative for type gating and snippet |
| `payloadType(doc.Payload)` | `Payload.Type *int` → dereferenced by `payloadType()` | ✅ Nil-safe, returns 0 when absent |
| `buildRichTextDetail(raw)` output → `mh.RichText` | `singleMessageHit:280` | ✅ All 3 search endpoints route through `singleMessageHit` |
| `fallbackSnippet` → `mh.Snippet` | `Payload.RichText.SearchText` for type=14 | ✅ Runs before rich-text projection; independent path |

- **C1 Dual-path parity:** N/A — no add/remove or create/delete symmetry paths involved. Read-only projection.
- **C2 Control-flow ordering / nesting:** N/A — `buildRichTextDetail` is called once per hit, no nesting, no re-entry. No security controls to bypass.
- **C3 Authorization boundary ≠ capability:** N/A — no new endpoints, tools, or capability exposure. All search endpoints already require channel access (`checkChannelAccess`).
- **C4 Auth lifecycle / container-member cascade:** N/A — no auth changes.
- **C5 Build ≠ runtime path:** N/A — no build artifacts, extensions, CLI, or relative-path concerns. Standard Go server.
- **C6 Governance / policy docs:** N/A — no governance or policy documents changed.

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. Three P2 items identified (swagger `$ref` sibling, `omitempty` consistency, nil-slice serialization edge) — all are non-blocking improvements to an otherwise clean, well-tested, additive PR. Data flow tracing confirms all consumed data reaches its consumption points correctly. Fail-soft contract is sound and test-covered.

yujiawei (approved):
# Code Review — PR #477 (octo-server)

Reviewed at head SHA `85d1e4592719b8e1c04502a86ba9f04074d254b5` (note: this is past the originally-listed `6c6b958`; the review targets current head). Build, `go vet`, and the full `modules/messages_search` test suite pass locally.

## Summary

Adds a typed rich-text (`payload.type=14`) projection to message search. `RichTextDetail`/`RichTextBlock`/`RichTextMention` are decoded from the indexer-preserved `_source.payloadRaw` blob by `buildRichTextDetail`, and surfaced under `MessageHit.rich_text` from `singleMessageHit`. The change is purely additive: `message_kind` stays `"text"`, `snippet` remains the highlight/text fallback, and `rich_text` is `omitempty`.

## Verdict: Approve

No blocking issues. The design is sound, the fail-soft path is genuinely fail-soft for the cases that matter, wire compatibility is preserved, and the projection is consistent across `/_search`, `/_search_all`, and `/_search_around` (all three route through `buildMessageHits` → `singleMessageHit`). The items below are advisory (P2) — worth a quick look, none should block merge.

## Verification performed

- **`payloadRaw` reaches `Doc`.** No `_source` include/exclude filtering exists anywhere in the module, so the new `PayloadRaw json.RawMessage` field is populated from OpenSearch `_source` as expected.
- **Type 14 is whitelisted.** `/_search` uses `terms(payload.type, [1,11,14])` and `/_search_all` uses `[1,8,11,14]` (+2,5 in browse), so type=14 docs actually reach the projection. `/_search_around` has no whitelist and also surfaces them.
- **Endpoint consistency.** `/_search`, `/_search_all`, and `/_search_around` all build hits via `buildMessageHits` → `singleMessageHit`, so `rich_text` is emitted uniformly. No endpoint-level divergence.
- **Wire compat.** `rich_text` is `omitempty` and verified absent on plain-text hits by `TestMessageHit_RichTextOmitemptyOnWire`. Existing clients see a byte-identical shape on non-richtext hits.
- **Whitespace-before-array branch is safe.** `buildRichTextDetail` branches on `p.Content[0] == '['`. I verified empirically that Go's `encoding/json` strips leading whitespace when capturing a nested value into a `json.RawMessage` struct field, so the captured `content` always begins with `[` for an array value even if the source JSON had `"content": \n  [...]`. The array-vs-string detection is correct as written. (This addresses a concern raised during review; it is not a bug.)

## Advisory findings (P2 — non-blocking)

**P2 — `buildRichTextDetail` returns a non-nil detail when `content` is a valid-JSON array of non-objects.** `modules/messages_search/source.go`, in the array branch:

```go
if p.Content[0] == '[' {
    _ = json.Unmarshal(p.Content, &d.Content)   // error swallowed
}
return d
```

If the envelope parses but `content` is an array whose elements don't match `RichTextBlock` (e.g. `[1,2,3]`), the inner unmarshal error is dropped and the function returns `&RichTextDetail{Content: nil}` rather than nil. The documented fail-soft contract is "nil → snippet fallback". Note this only affects *valid-JSON-but-wrong-shape* arrays: a syntactically broken array (`[{"type":...`) fails the **outer** `json.Unmarshal` first and correctly returns nil, so the practical blast radius is limited to malformed payloads from the (trusted) indexer. The existing test `TestBuildRichTextDetail_InvalidJSON` documents the "keep the envelope, drop content" behavior as intentional for the number-content case, so this is a deliberate design choice — but if strict fail-soft is wanted, capturing the array-branch error and returning nil on failure would make the contract uniform.

**P2 — `content` can serialize as `null`.** `RichTextDetail.Content []RichTextBlock` has no `omitempty` and swagger marks `content` as `required`, but a nil slice marshals to `"content": null`. This happens for the empty-legacy-string case and the malformed-array case above. If clients treat `content` as a guaranteed array, `null` may need a guard. Either initialize to an empty slice or relax the swagger `required` to match reality.

**P2 — legacy string normalization keeps whitespace-only strings.** The legacy branch collapses only `s == ""` to nil content; a `"   "` string becomes `[]RichTextBlock{{Type:"text", Text:"   "}}`. Minor — produces a blank text block instead of falling through. Trimming before the empty check would be more consistent with the stated "empty handling" intent.

## Strengths

- Good comment hygiene: the fail-soft contract, the array-vs-string legacy migration, and the `message_kind` "stays text" decision are all documented at the call sites.
- Test coverage is thorough: array content, legacy string, empty/nil raw, invalid JSON, mention transport, the per-hit projection contract (populated / legacy-nil / non-richtext), and the omitempty wire guard.
- The additive design (no `message_kind` enum change, snippet retained) keeps old clients working with zero changes.

PR #479, opened 2026-06-26
test(group): trace invite-members latency with per-segment timing + trace_id

an9xyz:
## What

Adds diagnostic instrumentation to locate latency in the
`POST /v1/groups/:group_no/members` (invite members) request path. No
business-logic change — the new logs fire only above the existing invite
slow-log threshold.

## Why

Invite-members latency is intermittently high, and the current spans
(`pre_checks_ms`, `db_ms`, IM hops) are too coarse to tell whether time goes
to acquiring a DB connection, executing queries, generating sequences,
committing (semi-sync), or in auth/middleware. The separate slow logs also
could not be correlated to one request.

## Changes

- **pkg/reqid** (new): request-scoped `trace_id`, kept in both `gin.Context`
  and the request context so code that only sees a `context.Context` (the auth
  token parser) can read it too.
- **main.go**: `trace_id` middleware mounted first; reuses an inbound
  `X-Request-ID` when present and echoes it back on the response.
- **pkg/accesslog**: emit `trace_id` on the GIN access line.
- **pkg/auth**: split `CacheTokenParser.Parse` into `token_ms` / `lang_ms` /
  `role_ms`; logs only when total exceeds `DM_AUTH_SLOWLOG_MS` (default 200ms).
- **group handler**: connection-acquire probe (`Conn` + `Ping`) to separate
  connection acquisition from query execution; threads `trace_id` into the service.
- **group service**: split `db_ms` into `query_ms` / `genseq_ms` /
  `insert_ms` / `commit_ms`.

The four slow logs (GIN access, auth parser, handler, service) now share the
same `trace_id`, so one slow request can be reassembled end to end.

## Behavior / risk

- Pure observability except one added `Ping` round-trip per invite (the
  connection-acquire probe). Invites are low-frequency.
- New logs are gated by the existing slow-log threshold; normal requests are
  unaffected.

## Test plan

- [x] `go build ./...`, `go vet`, `gofmt` clean
- [x] `go test ./pkg/reqid/... ./pkg/accesslog/... ./pkg/auth/...`
- [ ] Deploy, trigger a slow invite, confirm the four logs share a `trace_id`
      and the new `*_ms` fields decompose the latency.

### DIFF
--- a/main.go
+++ b/main.go
@@ -151,7 +151,7 @@ func runAPI(ctx *config.Context) {
 	// token parser 读）。让 GIN access log / token 解析 / handler / service 四条
 	// 分段慢日志能按 trace_id 串到同一次请求。复用入站 X-Request-ID（若有）。
 	route.UseGin(func(c *gin.Context) {
-		id := strings.TrimSpace(c.GetHeader("X-Request-ID"))
+		id := reqid.Sanitize(c.GetHeader("X-Request-ID"))
 		if id == "" {
 			id = reqid.New()
 		}
--- a/main.go
+++ b/main.go
@@ -28,6 +28,7 @@ import (
 	octoi18n "github.com/Mininglamp-OSS/octo-server/pkg/i18n"
 	"github.com/Mininglamp-OSS/octo-server/pkg/metrics"
 	octoredis "github.com/Mininglamp-OSS/octo-server/pkg/redis"
+	"github.com/Mininglamp-OSS/octo-server/pkg/reqid"
 	"github.com/gin-gonic/gin"
 	rd "github.com/go-redis/redis"
 	"github.com/judwhite/go-svc"
@@ -145,6 +146,20 @@ func runAPI(ctx *config.Context) {
 	if err != nil {
 		panic(err)
 	}
+	// trace_id 中间件（最前）：为每次请求生成/透传一个 trace_id，写进 gin.Context
+	// （供 access-log formatter 与 handler 读）和 request context（供 pkg/auth 的
+	// token parser 读）。让 GIN access log / token 解析 / handler / service 四条
+	// 分段慢日志能按 trace_id 串到同一次请求。复用入站 X-Request-ID（若有）。
+	route.UseGin(func(c *gin.Context) {
+		id := strings.TrimSpace(c.GetHeader("X-Request-ID"))
+		if id == "" {
+			id = reqid.New()
+		}
+		c.Set(reqid.GinKey, id)
+		c.Request = c.Request.WithContext(reqid.WithTraceID(c.Request.Context(), id))
+		c.Writer.Header().Set("X-Request-ID", id)
+		c.Next()
+	})
 	route.UseGin(octoi18n.EarlyMiddleware(octoi18n.MiddlewareOptions{
 		DefaultLanguage:        defaultLanguage,
 		TrustedLangHeaderCIDRs: trustedLangCIDRs,
--- a/modules/group/api.go
+++ b/modules/group/api.go
@@ -996,10 +996,16 @@ func (g *Group) memberAdd(c *wkhttp.Context) {
 	// 一次往返，单独计时。pre_checks 各步的 *_ms 是「取连接 + 执行」之和；本探针把
 	// 「取连接」这一段单拎出来——若 conn_acquire_ms 很小但 get_group_ms 等仍大，说明
 	// 慢在查询执行；若本探针本身就大，说明慢在连接获取 / 复用到被服务端掐死的连接后重连。
+	var connProbeErr string
 	if sqlDB := g.ctx.DB().Connection.DB; sqlDB != nil {
 		tProbe := time.Now()
-		if conn, cErr := sqlDB.Conn(c.Request.Context()); cErr == nil {
-			_ = conn.PingContext(c.Request.Context())
+		conn, cErr := sqlDB.Conn(c.Request.Context())
+		if cErr != nil {
+			connProbeErr = cErr.Error()
+		} else {
+			if pErr := conn.PingContext(c.Request.Context()); pErr != nil {
+				connProbeErr = pErr.Error()
+			}
 			_ = conn.Close() // 归还连接池
 		}
 		connAcquireMs = time.Since(tProbe).Milliseconds()
@@ -1174,6 +1180,7 @@ func (g *Group) memberAdd(c *wkhttp.Context) {
 			zap.Int("requested", len(req.Members)),
 			zap.Int64("pre_checks_ms", preChecksMs),
 			zap.Int64("conn_acquire_ms", connAcquireMs),
+			zap.String("conn_probe_err", connProbeErr),
 			zap.Int64("get_group_ms", getGroupMs),
 			zap.Int64("exist_member_ms", existMemberMs),
 			zap.Int64("bot_ownership_ms", botOwnerMs),
--- a/modules/group/api.go
+++ b/modules/group/api.go
@@ -33,6 +33,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
 	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
 	octoredis "github.com/Mininglamp-OSS/octo-server/pkg/redis"
+	"github.com/Mininglamp-OSS/octo-server/pkg/reqid"
 	spacepkg "github.com/Mininglamp-OSS/octo-server/pkg/space"
 	appwkhttp "github.com/Mininglamp-OSS/octo-server/pkg/wkhttp"
 	"github.com/gin-gonic/gin"
@@ -974,7 +975,9 @@ func (g *Group) memberAdd(c *wkhttp.Context) {
 		appConfigMs   int64
 		botSpaceMs    int64
 		preChecksMs   int64
+		connAcquireMs int64
 	)
+	traceID := c.GetString(reqid.GinKey)
 	operator := c.MustGet("uid").(string)
 	operatorName := c.MustGet("name").(string)
 	var req memberAddReq
@@ -988,6 +991,20 @@ func (g *Group) memberAdd(c *wkhttp.Context) {
 		return
 	}
 	groupNo := c.Param("group_no")
+
+	// 连接获取探针（邀请慢二级定位）：显式从连接池取一条连接并 PingContext 强制
+	// 一次往返，单独计时。pre_checks 各步的 *_ms 是「取连接 + 执行」之和；本探针把
+	// 「取连接」这一段单拎出来——若 conn_acquire_ms 很小但 get_group_ms 等仍大，说明
+	// 慢在查询执行；若本探针本身就大，说明慢在连接获取 / 复用到被服务端掐死的连接后重连。
+	if sqlDB := g.ctx.DB().Connection.DB; sqlDB != nil {
+		tProbe := time.Now()
+		if conn, cErr := sqlDB.Conn(c.Request.Context()); cErr == nil {
+			_ = conn.PingContext(c.Request.Context())
+			_ = conn.Close() // 归还连接池
+		}
+		connAcquireMs = time.Since(tProbe).Milliseconds()
+	}
+
 	/**
 	判断群是否存在
 	**/
@@ -1118,6 +1135,7 @@ func (g *Group) memberAdd(c *wkhttp.Context) {
 		Members:      req.Members,
 		OperatorUID:  operator,
 		OperatorName: operatorName,
+		TraceID:      traceID,
 	})
 	if err != nil {
 		// AddGroupMembers 会返回业务拒绝（如 allow_external=0 群里普通成员邀请
@@ -1150,10 +1168,12 @@ func (g *Group) memberAdd(c *wkhttp.Context) {
 	// bot_space 分摊，定位「埋点之外那 ~4.5s」到底卡在哪一步。
 	if handlerTotalMs := time.Since(handlerStart).Milliseconds(); handlerTotalMs >= inviteSlowLogThresholdMS {
 		g.Warn("邀请成员 handler 链路耗时偏高",
+			zap.String("trace_id", traceID),
 			zap.String("group_no", groupNo),
 			zap.String("operator", operator),
 			zap.Int("requested", len(req.Members)),
 			zap.Int64("pre_checks_ms", preChecksMs),
+			zap.Int64("conn_acquire_ms", connAcquireMs),
 			zap.Int64("get_group_ms", getGroupMs),
 			zap.Int64("exist_member_ms", existMemberMs),
 			zap.Int64("bot_ownership_ms", botOwnerMs),
--- a/modules/group/service.go
+++ b/modules/group/service.go
@@ -966,6 +966,7 @@ type AddGroupMembersServiceReq struct {
 	Members      []string // 待添加成员 UID 列表
 	OperatorUID  string   // 操作者 UID
 	OperatorName string   // 操作者名称
+	TraceID      string   // 请求级 trace_id（reqid），用于串联 handler/service/auth/GIN 分段慢日志
 }
 
 // AddGroupMembersServiceResp 添加群成员响应
@@ -1338,8 +1339,19 @@ func (s *Service) AddGroupMembers(req *AddGroupMembersServiceReq) (*AddGroupMemb
 	// 配合事务提交点与各 WuKongIM 调用的分段计时，区分 DB 阶段 vs IM 阶段。
 	startedAt := time.Now()
 
+	// db_ms 细分（邀请慢二级定位）：
+	//   query_ms   = 事务前的只读查询（群 / 用户 / 已存在成员 / 黑名单）累加；
+	//   genseq_ms  = GenSeq 累加——进程级 seqLock 全局锁 + 步长耗尽时查/写 seq 表，
+	//                锁竞争或 seq 表 I/O 慢会集中体现在这里；
+	//   insert_ms  = 事务内 ExistMemberDelete + Insert/recover + 外部群标记累加；
+	//   commit_ms  = tx.Commit()，含半同步复制等待 slave ACK。
+	// 四者 + space_check_ms 拼出 db_ms，定位 DB 阶段慢在读 / 取号 / 写 / 提交哪一段。
+	var queryMs, genSeqMs, insertMs, commitMs int64
+
 	// 群存在性 + 状态检查
+	qStart := time.Now()
 	groupModel, err := s.db.QueryWithGroupNo(req.GroupNo)
+	queryMs += time.Since(qStart).Milliseconds()
 	if err != nil {
 		s.Error("query group failed", zap.Error(err))
 		return nil, errors.New("failed to query group")
@@ -1401,14 +1413,18 @@ func (s *Service) AddGroupMembers(req *AddGroupMembersServiceReq) (*AddGroupMemb
 	spaceCheckMs := time.Since(spaceCheckStart).Milliseconds()
 
 	// 查询用户信息
+	qStart = time.Now()
 	memberUsers, err := s.userDB.QueryByUIDs(uniqueUIDs)
+	queryMs += time.Since(qStart).Milliseconds()
 	if err != nil {
 		s.Error("query member info failed", zap.Error(err))
 		return nil, errors.New("failed to query member info")
 	}
 
 	// 过滤已在群内的成员
+	qStart = time.Now()
 	existingMembers, err := s.db.QueryMembersWithUids(uniqueUIDs, req.GroupNo)
+	queryMs += time.Since(qStart).Milliseconds()
 	if err != nil {
 		s.Error("query existing members failed", zap.Error(err))
 		return nil, errors.New("failed to query existing members")
@@ -1421,7 +1437,9 @@ func (s *Service) AddGroupMembers(req *AddGroupMembersServiceReq) (*AddGroupMemb
 	}
 
 	// 过滤黑名单
+	qStart = time.Now()
 	blacklistMembers, _ := s.db.QueryMembersWithStatus(req.GroupNo, int(common.GroupMemberStatusBlacklist))
+	queryMs += time.Since(qStart).Milliseconds()
 	blacklistSet := make(map[string]bool)
 	for _, m := range blacklistMembers {
 		blacklistSet[m.UID] = true
@@ -1446,7 +1464,9 @@ func (s *Service) AddGroupMembers(req *AddGroupMembersServiceReq) (*AddGroupMemb
 		if existingSet[memberUser.UID] || blacklistSet[memberUser.UID] {
 			continue
 		}
+		genStart := time.Now()
 		memberVersion, err := s.ctx.GenSeq(common.GroupMemberSeqKey)
+		genSeqMs += time.Since(genStart).Milliseconds()
 		if err != nil {
 			s.Error("generate member version failed", zap.Error(err))
 			return nil, err
@@ -1460,6 +1480,7 @@ func (s *Service) AddGroupMembers(req *AddGroupMembersServiceReq) (*AddGroupMemb
 		}
 
 		// 检查是否之前被删除过（需要恢复）
+		insStart := time.Now()
 		existDelete, _ := s.db.ExistMemberDelete(memberUser.UID, req.GroupNo)
 		newMember := &MemberModel{
 			GroupNo:       req.GroupNo,
@@ -1478,6 +1499,7 @@ func (s *Service) AddGroupMembers(req *AddGroupMembersServiceReq) (*AddGroupMemb
 		} else {
 			err = s.db.InsertMemberTx(newMember, tx)
 		}
+		insertMs += time.Since(insStart).Milliseconds()
 		if err != nil {
 			s.Error("add group member failed", zap.Error(err), zap.String("uid", memberUser.UID))
 			continue
@@ -1496,7 +1518,10 @@ func (s *Service) AddGroupMembers(req *AddGroupMembersServiceReq) (*AddGroupMemb
 	// 首次出现外部成员时，在事务内将群标记为外部群，确保成员/群标记一致提交
 	markedExternal := false
 	if hasNewExternal && groupModel.IsExternalGroup == 0 {
-		if updateErr := s.db.UpdateIsExternalGroupTx(req.GroupNo, 1, tx); updateErr != nil {
+		updStart := time.Now()
+		updateErr := s.db.UpdateIsExternalGroupTx(req.GroupNo, 1, tx)
+		insertMs += time.Since(updStart).Milliseconds()
+		if updateErr != nil {
 			s.Error("update is_external_group failed", zap.Error(updateErr), zap.String("group_no", req.GroupNo))
 			return nil, errors.New("failed to update external group flag")
 		}
@@ -1504,8 +1529,11 @@ func (s *Service) AddGroupMembers(req *AddGroupMembersServiceReq) (*AddGroupMemb
 	}
 
 	// 提交事务
-	if err := tx.Commit(); err != nil {
-		s.Error("commit transaction failed", zap.Error(err))
+	commitStart := time.Now()
+	commitErr := tx.Commit()
+	commitMs = time.Since(commitStart).Milliseconds()
+	if commitErr != nil {
+		s.Error("commit transaction failed", zap.Error(commitErr))
 		return nil, errors.New("failed to commit transaction")
 	}
 	// 事务（GenSeq + insert/recover + commit）耗时。
@@ -1584,12 +1612,17 @@ func (s *Service) AddGroupMembers(req *AddGroupMembersServiceReq) (*AddGroupMemb
 	total := time.Since(startedAt)
 	if total.Milliseconds() >= inviteSlowLogThresholdMS {
 		s.Warn("邀请成员入群链路耗时偏高",
+			zap.String("trace_id", req.TraceID),
 			zap.String("group_no", req.GroupNo),
 			zap.String("operator", req.OperatorUID),
 			zap.Int("requested", len(req.Members)),
 			zap.Int("added", len(addedUIDs)),
 			zap.Int64("db_ms", dbDur.Milliseconds()),
 			zap.Int64("space_check_ms", spaceCheckMs),
+			zap.Int64("query_ms", queryMs),
+			zap.Int64("genseq_ms", genSeqMs),
+			zap.Int64("insert_ms", insertMs),
+			zap.Int64("commit_ms", commitMs),
 			zap.Int64("tx_ms", txMs),
 			zap.Int64("channel_update_ms", channelUpdateDur.Milliseconds()),
 			zap.Int64("im_add_subscriber_ms", imSubscriberDur.Milliseconds()),
--- a/pkg/accesslog/accesslog.go
+++ b/pkg/accesslog/accesslog.go
@@ -128,16 +128,22 @@ func Formatter(param gin.LogFormatterParams) string {
 		param.Latency = param.Latency.Truncate(time.Second)
 	}
 	// trace_id 串联同一次请求的各分段慢日志(auth/handler/service);由 reqid
-	// 中间件写入 gin.Context,经 param.Keys 透出。缺失时留空,不影响原格式。
+	// 中间件写入 gin.Context,经 param.Keys 透出。
+	// 追加在行尾(path 之后)而非插在中间:保持原有按 `|` 分列的位置不变,避免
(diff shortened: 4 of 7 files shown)


mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#479 Review Report
**Head SHA:** `c985fb2d16b814cd4a44e789698b71e4dc2ea7cc`
**PR Title:** test(group): trace invite-members latency with per-segment timing + trace_id
**Routing:** security_sensitive (automated review)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| reqid package correctness | ✅ | `pkg/reqid/reqid.go` — 8-byte random hex, unexported `ctxKey{}` for type-safe context, fallback sentinel on rand failure |
| Middleware ordering | ✅ | `main.go:152-163` — trace_id middleware is the first `UseGin` call, before i18n, tracer, metrics, access log, rate limit |
| Data flow: trace_id → access log | ✅ | Middleware sets `c.Set(GinKey, id)` → `gin.LogFormatterParams.Keys` populated by gin → `Formatter` reads `param.Keys[reqid.GinKey]` |
| Data flow: trace_id → auth parser | ✅ | Middleware sets `c.Request.WithContext(reqid.WithTraceID(...))` → `route.SetTokenParser` receives `ctx = c.Request.Context()` → `reqid.FromContext(ctx)` extracts |
| Data flow: trace_id → handler → service | ✅ | `api.go:983` reads `c.GetString(reqid.GinKey)` → passed via `AddGroupMembersServiceReq.TraceID` → service slow log reads `req.TraceID` |
| Connection probe safety | ✅ | `api.go:999-1006` — `Conn()` + `PingContext()` + `Close()` with proper cleanup; nil-check on `sqlDB`; errors silently ignored (acceptable for diagnostic probe) |
| Timing instrumentation correctness | ✅ | `service.go:1350-1541` — `qStart`/`genStart`/`insStart`/`updStart`/`commitStart` each scoped to their segment, accumulated with `+=` in loops |
| No business-logic change | ✅ | All early returns, error handling, and control flow preserved; only `time.Now()` calls and `zap.Int64` fields added |
| Test coverage for reqid | ✅ | `pkg/reqid/reqid_test.go` — uniqueness, hex format, context round-trip, absent-key returns "" |
| Commit error handling preserved | ✅ | `service.go:1532-1536` — refactored to capture `commitMs` before error check; semantics identical to original |

## 2. Findings

### F1 — Access log format change (P2 / note)
- **Diff-scope:** **new** — PR introduces a new `| trace_id=%s |` segment in the access log format.
- **Detail:** `pkg/accesslog/accesslog.go:133` inserts `| trace_id=%s |` between the ClientIP column and the method column. This is a positional format change that will break any downstream log parser relying on fixed pipe-delimited column indices (e.g. `awk -F'|' '{print $6}'`).
- **Mitigation:** The new segment uses key=value format (`trace_id=...`) which is self-describing. Modern structured-log parsers (JSON, kv) should handle this. The existing test `TestFormatter_ScrubsToken` does not assert column positions, so no test breakage.
- **Recommendation:** Note in deploy changelog that access log format has changed. If any log pipeline uses positional parsing, update it before deploy.

### F2 — X-Request-ID header trust scope (P2 / note)
- **Diff-scope:** **new** — PR trusts inbound `X-Request-ID` header for log correlation.
- **Detail:** `main.go:155-158` — if `X-Request-ID` is present, it's used as-is (after `TrimSpace`). An attacker could inject an arbitrary value. This is safe because:
  - It's used **only** for log correlation, never for authorization or routing.
  - It's never interpolated into SQL, shell commands, or HTML.
  - The `Response Header` echo-back (`c.Writer.Header().Set("X-Request-ID", id)`) is standard proxy behavior.
- **Recommendation:** Consider adding a max-length guard (e.g. truncate at 128 chars) to prevent log-line inflation or log injection via newline characters in the header value. Low priority.

## 3. Suggestions

1. **Max-length on X-Request-ID** (optional): `id = id[:min(len(id), 128)]` after TrimSpace to bound log field width.
2. **Probe failure logging** (optional): When `Conn()` or `PingContext()` fails in the connection probe, consider logging the error in the handler slow-log when `connAcquireMs` is high — would help distinguish pool exhaustion from network issues.

## 4. Extra Observations

- The `reqid.New()` fallback sentinel `"0000000000000000"` on `crypto/rand` failure is pragmatic. `crypto/rand.Read` on Linux reads from `/dev/urandom` which never fails in practice.
- The `authSlowLogThresholdMS` init-time parsing pattern matches `inviteSlowLogThresholdMS` in `modules/group/service.go` — consistent convention.
- The `insertMs` accumulation in the loop (`service.go:1483-1501`) correctly covers both `ExistMemberDelete` query and `InsertMemberTx`/`recoverMemberTx` per member.

## 5. Data Flow Trace

| Consumer | Upstream Source | Data Reaches Consumer? |
|----------|----------------|----------------------|
| `accesslog.Formatter` reads `param.Keys[reqid.GinKey]` | Middleware `c.Set(reqid.GinKey, id)` in `main.go:160` | ✅ Yes — gin copies `gin.Context` keys to `LogFormatterParams.Keys` |
| `CacheTokenParser.Parse` reads `reqid.FromContext(ctx)` | Middleware `c.Request.WithContext(reqid.WithTraceID(...))` in `main.go:161` | ✅ Yes — `route.SetTokenParser.Parse` receives `c.Request.Context()` which carries the trace_id |
| `memberAdd` handler reads `c.GetString(reqid.GinKey)` | Middleware `c.Set(reqid.GinKey, id)` in `main.go:160` | ✅ Yes — same gin context |
| `AddGroupMembers` service reads `req.TraceID` | Handler passes `traceID` via `AddGroupMembersServiceReq.TraceID` in `api.go:1138` | ✅ Yes — struct field, direct assignment |
| `connAcquireMs` consumed in handler slow log | Probe in `api.go:999-1006` | ✅ Yes — always set (0 when `sqlDB == nil` or on error) |
| `queryMs`/`genSeqMs`/`insertMs`/`commitMs` consumed in service slow log | Accumulated in `service.go:1350-1541` | ✅ Yes — `+=` accumulation in loop, set before log |

- **C1 — Dual-path parity:** N/A. This is an observability-only PR. No add/remove, subscribe/unsubscribe, or create/delete pairs modified. The trace_id middleware applies to ALL requests uniformly. The connection probe and timing are specific to the invite-members handler.
- **C2 — Control-flow ordering / nested reuse:** Clear. `time.Now()` calls are leaf-level instrumentation with no nesting or double-counting. In the `for` loop (`service.go:1464-1527`), `genStart` and `insStart` are correctly scoped per iteration with `+=` accumulation. The commit timing is single-shot outside the loop. No safety controls modified.
- **C3 — Authorization boundary:** N/A. No new endpoints, tools, or capabilities exposed. The trace_id is a random hex string for log correlation only. The `X-Request-ID` passthrough does not grant any capability.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth changes. The `CacheTokenParser.Parse` logic is unchanged; only timing instrumentation added around existing resolver calls.
- **C5 — Build ≠ runtime path:** N/A. No build artifacts, extensions, CLI tools, or relative URL changes. This is a pure Go server change with no packaging implications.
- **C6 — Governance / policy document consistency:** N/A. No documentation or policy changes.

---

**[Octo-Q] verdict: APPROVE**

Pure observability instrumentation with no business-logic change. Data flow for trace_id is correctly propagated through all four log layers (access log, auth parser, handler, service). Connection probe is safe with proper cleanup. Two P2 notes (access log format change, X-Request-ID max-length) are informational and non-blocking.

Jerry-Xin (dismissed):
In scope: yes. This PR adds observability to octo-server's request/auth/accesslog/group invite path, which is directly relevant to the repository.

💬 Non-blocking

🟡 Warning: `main.go:153` reuses client-supplied `X-Request-ID` without length or character validation, then echoes and logs it. Consider capping length and allowing only log-safe characters so trace fields stay parseable and do not become attacker-controlled high-cardinality log noise.

🟡 Warning: `modules/group/api.go:999` adds a `Conn` + `PingContext` probe to every successful invite request, not only slow requests. That is acceptable for low-frequency diagnostics, but the field name `conn_acquire_ms` includes both pool acquisition and ping round-trip time. Splitting `conn_acquire_ms` and `conn_ping_ms`, or naming it `conn_probe_ms`, would reduce misdiagnosis.

🔵 Suggestion: `pkg/auth/parser.go:189` only emits the auth slow log on successful token parsing. If Redis/cache failures are part of the suspected middleware latency, a deferred slow-log path for error returns would make the instrumentation more complete.

✅ Highlights

The middleware is mounted early enough for the trace id to reach both Gin access logs and octo-lib `AuthMiddleware` via request context. The service-level timing is narrowly scoped and does not change the transaction flow. trace_id is log-field-only (zap.String / accesslog format), never used as a prometheus metric label — no metric-cardinality risk. Per-segment timing (query_ms/genseq_ms/insert_ms/commit_ms) decomposes the existing db_ms in the same goroutine with no double-count and no control-flow change, consistent with the #471/#472 observability pattern.

Verification run: `go build ./...` (pass), `go test ./pkg/reqid/... ./pkg/accesslog/... ./pkg/auth/...` (pass). The `modules/group` full-suite failure is the known pre-existing env panic (`unknown migration in database` in testutil.NewTestServer), unrelated to this diff.

yujiawei (dismissed):
# Code Review — PR #479 (octo-server)

**Scope reviewed:** head SHA `c985fb2d16b814cd4a44e789698b71e4dc2ea7cc`, against merge-base `bdab0dc2` on `main`. 7 files, +204/-4. Build, `go vet`, and `pkg/reqid` unit tests all pass at this SHA.

This is observability-only work: a request-scoped `trace_id` threaded through the GIN access log, the auth token parser, and the group invite handler/service, plus per-segment latency timing for the invite-members path. No business logic changes; no schema/migration changes.

## Verdict: Approve

No correctness, security-critical, data-loss, or build-breaking issue. The findings below are all non-blocking (P2). Two of them are worth a human's eye given this path touches the auth middleware and reflects a client-supplied header — called out explicitly under **For human/security attention**.

---

## Findings

### P2-1 — Connection-acquire probe runs unconditionally on every invite request, and contends for the very resource it measures
`modules/group/api.go:999-1006`

```go
if sqlDB := g.ctx.DB().Connection.DB; sqlDB != nil {
    tProbe := time.Now()
    if conn, cErr := sqlDB.Conn(c.Request.Context()); cErr == nil {
        _ = conn.PingContext(c.Request.Context())
        _ = conn.Close() // returns to pool
    }
    connAcquireMs = time.Since(tProbe).Milliseconds()
}
```

The probe is correct in isolation — `conn.Close()` returns the connection to the pool, and there is no leak on the error path (nothing is acquired when `Conn()` fails). The concern is methodological + hot-path overhead:

- It checks out a connection from the **same shared pool** (`g.ctx.DB().Connection.DB`, the dbr session's `*sql.DB`) that the invite's own queries use, then forces a network round-trip (`PingContext`), on **every** `memberAdd` call — not only slow ones.
- The investigation this serves is "invites are slow (~4.5s), possibly DB pool starvation." If the cause really is pool starvation, this probe **competes for a pool slot** at exactly the moment slots are scarce, so it can worsen the incident it is diagnosing and inflate its own `conn_acquire_ms` reading.

You genuinely cannot gate this metric behind the slow-threshold (you need it measured *before* you know the request was slow). The non-perturbing alternative is passive pool stats: `db.Stats()` exposes `WaitCount` and `WaitDuration`, which signal pool starvation directly without adding an active round-trip. Recommend complementing or replacing the active probe with `db.Stats()` deltas, and — if this endpoint is high-QPS in your environment — removing the active probe once the investigation concludes.

### P2-2 — Per-iteration `.Milliseconds()` truncation under-counts sub-millisecond segments
`modules/group/service.go:1469` (`genSeqMs`), `1502`/`1521` (`insertMs`), and the `queryMs` accumulations at `1340/1413/1437/1464`

Each accumulator adds `time.Since(start).Milliseconds()` *inside* the per-member loop. `Milliseconds()` returns `int64`, so any single iteration under 1ms truncates to `0`. With N members each taking, say, 0.8ms, the reported `genseq_ms`/`insert_ms` is `0` instead of `~N×0.8ms`. This won't hide the *actual* slow bottleneck (a genuinely slow segment is ≫1ms and still shows), but it makes the fast-segment accounting misleading — which partly defeats the "per-segment timing" goal for the N+1 query pattern. Fix: accumulate raw `time.Duration` and convert to `.Milliseconds()` once at log time.

### P2-3 — Inbound `X-Request-ID` is trusted verbatim as the trace id (no length/charset bound)
`main.go:153-160`, surfaced in `pkg/accesslog/accesslog.go:132`

```go
id := strings.TrimSpace(c.GetHeader("X-Request-ID"))
if id == "" { id = reqid.New() }
c.Set(reqid.GinKey, id)
c.Request = c.Request.WithContext(reqid.WithTraceID(c.Request.Context(), id))
c.Writer.Header().Set("X-Request-ID", id)
```

The client-supplied value is reflected into the response header, the request context, and **every** access-log line (`trace_id=%s`, raw `%s` in the formatter). Assessment of the vectors:

- **CRLF / log-splitting / response-splitting: not exploitable here.** HTTP/1.1 framing makes a bare `\n`/`\r` in a request header *value* impossible (the line terminator ends the header), and Go's `net/http` validates header values on the response-write side, so the reflected `X-Request-ID` cannot inject a split response either.
- **What does remain (P2):** there is no length cap, so a client can supply a value up to `MaxHeaderBytes` (~1MB default) that lands on every log line and in the response — log bloat. And non-CRLF control bytes (e.g. ANSI escapes `\x1b[...`) can reach the on-disk/stdout access log, which a terminal viewer would interpret. Recommend bounding to a small max length (e.g. 64–128 chars) and restricting to a safe charset before reuse — only accepting the inbound id when it matches, else generating a fresh one with `reqid.New()`.

Note for consistency: the same `runAPI` already applies a *trusted-CIDR* gate to the inbound `Accept-Language` header (`TrustedLangHeaderCIDRs`, `main.go:165`). Trusting `X-Request-ID` from arbitrary clients is a slight departure from that established "don't blindly trust inbound headers" convention — worth aligning.

### P2-4 — Access-log formatter is silently coupled to the reqid middleware
`pkg/accesslog/accesslog.go:132`

`traceID, _ := param.Keys[reqid.GinKey].(string)` is safe (nil-map index and failed assertion both yield `""`, the original format is preserved). Minor: if middleware ordering ever changes so the formatter runs without reqid having populated the key, the column silently goes blank with no signal. Acceptable as-is; just be aware the access log's trace column depends on the reqid middleware staying registered ahead of the logger.

---

## For human / security attention (PR is security-sensitive labeled)

1. **P2-3** — confirm the intended trust posture for inbound `X-Request-ID`. The dangerous vectors (CRLF/response-splitting) are not exploitable, but accepting an unbounded client-controlled value into logs/response without validation is a hardening gap and diverges from the file's existing trusted-CIDR header convention.
2. **P2-1** — the active DB probe on the invite hot path. Confirm `memberAdd` QPS is low enough that an extra pooled `Ping` per request is acceptable, and prefer `db.Stats()` for pool-starvation signal.

## Positives

- `pkg/reqid` is clean, well-documented, and tested (uniqueness, round-trip, absent-context). The rand-failure sentinel fallback (`reqid.go:32`) is the right call — a non-unique id beats failing a request over a log field.
- Dual propagation (gin.Context key for the formatter/handler + `context.Context` for the auth parser) is correct given how `wkhttp.AuthMiddleware` invokes `parser.Parse(c.Request.Context(), ...)`; the trace id reaches the parser context.
- Slow-path logs use `zap.String`, which escapes control characters, so the structured slow logs are not injectable (only the raw `%s` access-log line carries the P2-3 caveat).
- No secrets/tokens are logged in the auth slow-log (only `uid` and timings), and `DM_AUTH_SLOWLOG_MS` parsing correctly rejects non-positive/invalid values and falls back to 200ms.

Jerry-Xin (approved):
**Re-review @ 52bef69** (prior round APPROVE @ c985fb2d by 3 reviewers, all auto-dismissed on this push). The new commit directly addresses the prior non-blocking nits. ✅ APPROVE.

### What changed vs c985fb2d
- `pkg/reqid`: new `Sanitize()` + `TestSanitize` table test.
- `main.go`: inbound `X-Request-ID` now `reqid.Sanitize(...)` instead of raw `strings.TrimSpace`.
- `pkg/accesslog`: `trace_id` moved from a mid-line column to an end-of-line suffix; omitted entirely when empty.
- `modules/group/api.go`: conn-probe now captures `conn_probe_err` into a zap field.

### Prior 🟡s — status
- 🟡 **F2 (X-Request-ID clamp) — FIXED, correct.** `Sanitize` trims whitespace, bounds length to 128, and strips control chars via `strings.Map(func(r) { if r < 0x20 || r == 0x7f { return -1 } ... })` — this drops CR/LF/tab/NUL/DEL, so a client-supplied id can no longer forge or split log lines (the GIN text formatter + zap field + echoed `X-Request-ID` response header are all now fed the sanitized value). Test explicitly covers `strip_crlf_injection`, controls, truncation, empty-after-clean. Empty result falls back to `New()`.
- 🟡 **F1 (accesslog column layout) — FIXED.** trace_id is appended after path as ` | trace_id=…` and omitted when empty, preserving the original `|`-delimited column positions for downstream parsers.
- 🟡 **conn-probe gating — partial.** `conn_probe_err` is now surfaced, but the `Conn`+`PingContext` probe still runs on **every** invite (not slow-path gated) and `conn_acquire_ms` still conflates pool-acquire + ping. Still non-blocking — it's a diagnostic on a path already doing DB work.

### No regressions (verified-sound properties hold)
- `trace_id` remains **log-field-only**: gin.Context key + zap field + accesslog format string. Grep confirms **no** prometheus label uses it (`WithLabelValues(traceID)` absent); the prometheus refs in main.go are the pre-existing httpMetrics/DependencyMetrics middleware. No cardinality risk.
- No double-count: segment `*_ms` still decompose existing `db_ms`; same-goroutine `time.Since`.
- No control-flow change: Sanitize is a pure string transform; probe error is recorded, not acted on.

### Build / test
- `go build ./...` → exit 0.
- `go test ./pkg/reqid/... ./pkg/accesslog/... ./pkg/auth/...` → all `ok` (incl. new TestSanitize).
- `go test ./modules/group/...` → pre-existing **env** failure only: `panic: Unable to create migration plan … unknown migration in database` in `octo-lib/testutil.NewTestServer`. Unrelated to this diff (`go vet ./modules/group/` exits 0; package + tests compile). Classified as environmental, not PR-owned.

Clean observability addition; prior security/parse nits resolved. LGTM.

mochashanyao (approved):
[Octo-Q · automated review]

---

# PR#479 Review Report — octo-server: test(group): trace invite-members latency with per-segment timing + trace_id
**Head SHA:** `52bef69aeb60bbf3d80e4df89f0ce94fcd874718` (commit 2/2, addresses F1/F2 + probe error)
**Routing:** complexity=security_sensitive (automated review / automated review)

---

## 1. Summary

Pure observability PR: adds a `trace_id` correlation mechanism and per-segment timing to the invite-members (`POST /v1/groups/:group_no/members`) request path. No business-logic change. All new logging is gated behind existing slow-log thresholds. 7 files changed (+264/−4).

**Overall assessment: Well-designed instrumentation with correct data flow, proper log-injection defense, and good test coverage. No P0/P1 findings.**

---

## 2. Verification Results

| Item | Status | Evidence |
|------|--------|----------|
| trace_id middleware registered first in chain | ✅ | `main.go:155-166` — before EarlyMiddleware, tracer, access logger, rate limiter, route-level auth |
| trace_id flows to request context (for pkg/auth) | ✅ | `main.go:163` sets `c.Request.WithContext(reqid.WithTraceID(...))`; octo-lib `wkhttp.AuthMiddleware` passes `c.Request.Context()` to `parser.Parse()` (verified at octo-lib `http.go:323`) |
| trace_id flows to gin.Context (for handlers/access log) | ✅ | `main.go:162` sets `c.Set(reqid.GinKey, id)`; access log reads from `param.Keys[reqid.GinKey]` (`accesslog.go:140`); handler reads `c.GetString(reqid.GinKey)` (`api.go:983`) |
| Sanitize prevents log injection | ✅ | `reqid.go:35-47` — strips `r < 0x20` (covers CR/LF/tab/NUL), strips DEL (0x7f), trims whitespace, bounds to 128 bytes |
| Sanitize never used for auth/routing | ✅ | Only consumed for log correlation fields (`zap.String("trace_id", ...)`); explicitly documented in `reqid.go:30` |
| Access log format preserves positional columns | ✅ | `accesslog.go:146` — trace_id appended as `traceSuffix` after path, not inserted mid-line (F1 fix) |
| Connection probe uses request context | ✅ | `api.go:1003-1004` — `sqlDB.Conn(c.Request.Context())` and `conn.PingContext(c.Request.Context())` |
| Connection probe error surfaced | ✅ | `api.go:1001-1009` — captures both Conn error and PingContext error into `connProbeErr`; logged at `api.go:1180` |
| All timing variables goroutine-safe | ✅ | All timing vars are function-local; no shared state mutation; `authSlowLogThresholdMS` is package-level init-time read-only |
| Slow-log threshold gating | ✅ | Handler: `inviteSlowLogThresholdMS` (service.go:139); Auth: `authSlowLogThresholdMS` default 200ms (parser.go:28); Access log: always on (acceptable — no perf impact) |
| Tests cover reqid core | ✅ | `reqid_test.go` — uniqueness, context round-trip, absent case, sanitize (trim/CRLF/control/truncate/normal) |
| `log.NewTLog` exists in octo-lib | ✅ | Verified at octo-lib `pkg/log/logger.go:203`; `TLog.Warn(msg, ...zap.Field)` signature matches usage |
| Build / vet clean (per PR) | ✅ | PR test plan: `go build ./...`, `go vet`, `gofmt` clean |

---

## 3. Findings

### P2-1: Sanitize byte-level truncation may split multi-byte UTF-8 (cosmetic)

**Location:** `pkg/reqid/reqid.go:40-41`

```go
if len(s) > maxInboundLen {
    s = s[:maxInboundLen]
}
```

**Issue:** `s[:maxInboundLen]` truncates at byte boundary. For multi-byte UTF-8 characters (e.g., CJK = 3 bytes each), byte-level truncation can split a character mid-sequence. `strings.Map` then converts the incomplete bytes to U+FFFD (replacement character, 3 bytes), potentially producing output slightly longer than 128 bytes.

**Impact:** Cosmetic only. The value is used for log correlation, not auth/routing. A garbled character in the trace_id field is ugly but harmless.

**Diff-scope:** New code (this PR). Not pre-existing.

**Suggestion (non-blocking):** Use rune-aware truncation:

```go
if len(s) > maxInboundLen {
    runes := []rune(s)
    if len(runes) > maxInboundLen {
        s = string(runes[:maxInboundLen])
    } else {
        s = string(runes)
        // byte-len still > maxInboundLen for wide chars; truncate bytes:
        s = s[:maxInboundLen]
    }
}
```

Or simply accept the cosmetic artifact (recommended — not worth the complexity for a log correlation field).

---

### P2-2: queryMs decomposition does not cover spaceCheckMs queries

**Location:** `modules/group/service.go:1355-1396` (space-check block) vs `service.go:1360-1361` (QueryWithGroupNo)

**Issue:** `queryMs` accumulates four explicit query timings: `QueryWithGroupNo`, `QueryByUIDs`, `QueryMembersWithUids`, `QueryMembersWithStatus`. The `spaceCheckMs` block (which itself is reported separately) internally calls `s.db.QueryMemberWithUID` and `spacepkg.CheckMembership` — these DB queries are NOT included in `queryMs`.

**Impact:** When analyzing slow logs, `db_ms` ≠ `query_ms + genseq_ms + insert_ms + commit_ms + space_check_ms` — there is unattributed time in `spaceCheckMs` that itself contains queries. The decomposition is partial.

**Diff-scope:** New instrumentation (this PR). Not a regression.

**Suggestion (non-blocking):** Add a comment in the slow-log output or field documentation noting that `space_check_ms` contains its own internal DB queries not captured in `query_ms`. Alternatively, add `queryMs += ...` for the `QueryMemberWithUID` call inside the space-check loop.

---

## 4. No Issues Found (Explicit Clears)

- **No business-logic change:** Diff is purely additive instrumentation. All existing code paths (error handling, auth, member-add logic) are untouched.
- **No auth bypass:** trace_id is never used for authorization. `pkg/auth` continues to validate tokens independently.
- **No resource leak:** Connection probe properly calls `conn.Close()` to return the connection to the pool (`api.go:1008`). `defer` is not needed since the close is unconditional within the else branch.
- **No double-counting in commit_ms:** `tx.Commit()` error is captured before the timing assignment, and the error check follows. Original `if err := tx.Commit(); err != nil` semantics preserved (`service.go:1533-1537`).
- **No goroutine leak or blocking:** All instrumentation is synchronous within the request handler. No background goroutines spawned.

---

## 5. Data-Flow Traceback

| Consumer | Data | Upstream Source | Flows Correctly? |
|----------|------|----------------|-----------------|
| `api.go:983` `c.GetString(reqid.GinKey)` | trace_id | `main.go:162` `c.Set(reqid.GinKey, id)` | ✅ gin.Context set before handler |
| `parser.go:192` `reqid.FromContext(ctx)` | trace_id | `main.go:163` `c.Request.WithContext(reqid.WithTraceID(...))` → octo-lib `AuthMiddleware` passes `c.Request.Context()` to `Parse()` | ✅ Verified at octo-lib http.go:323 |
| `accesslog.go:140` `param.Keys[reqid.GinKey]` | trace_id | `main.go:162` `c.Set(reqid.GinKey, id)` → gin propagates c.Keys to LogFormatterParams | ✅ Standard gin behavior |
| `api.go:1012` `connAcquireMs` | timing | `api.go:1001-1010` `time.Since(tProbe)` around Conn+PingContext+Close | ✅ Covers full probe lifecycle |
| `api.go:1009` `connProbeErr` | error string | `api.go:1003-1008` captures either Conn error or PingContext error | ✅ Both failure modes covered |
| `service.go:1361` `queryMs` accumulation | timing | 4 × `time.Since(qStart)` around individual DB queries | ✅ Each query bracketed correctly |
| `service.go:1468` `genSeqMs` accumulation | timing | `service.go:1467-1468` `time.Since(genStart)` around `GenSeq` in loop | ✅ Accumulates per-member |
| `service.go:1500` `insertMs` accumulation | timing | `service.go:1482-1500` covers ExistMemberDelete + Insert/recover | ✅ |
| `service.go:1535` `commitMs` | timing | `service.go:1533-1535` covers `tx.Commit()` only | ✅ |
| `parser.go:127` `tokenMs` | timing | `parser.go:128-129` around `p.Cache.Get` | ✅ |
| `parser.go:161` `langMs` | timing | `parser.go:160-161` around `p.resolver.Resolve` | ✅ |
| `parser.go:182` `roleMs` | timing | `parser.go:181-182` around `p.roleResolver.ResolveRole` | ✅ |

---

## 6. Blind-Spot Checklist (R5 — C1–C6)

**C1 — Dual-path parity:** Clear.
- trace_id middleware is symmetric (runs on every request entering the gin router). No add/remove asymmetry.
- Connection probe is invite-handler-specific; no corresponding remove-path concern.

**C2 — Control-flow ordering / nesting:** Clear.
- Timing instrumentation is straightforward sequential (no nesting or double-application).
- `Sanitize` was tested with non-canonical inputs: CRLF (`\r\n`), tab (`\t`), NUL (`\x00`), empty-after-clean, over-length. Control chars correctly dropped.
- Access log `traceSuffix` is append-only at end of line; no risk of disrupting positional parsers.

**C3 — Authorization boundary ≠ capability boundary:** N/A.
- No new endpoints, tools, or permission changes. trace_id is observability metadata, not a capability.

**C4 — Authorization lifecycle / container-member cascade:** N/A.
- No auth changes. Existing auth flow untouched.

**C5 — Build/note ≠ runtime correctness:** Clear.
- PR confirms `go build ./...` passes. All runtime dependencies (context, time, zap, crypto/rand) are standard and available.
- No relative-path/origin/extension concerns (this is a Go server, not a browser extension).
- No build-time vs runtime environment divergence.

**C6 — Governance/policy doc self-consistency:** N/A.
- Not a documentation or policy PR.

---

## 7. Cross-Round Blocker Re-review (R6)

N/A — this is the first review of PR#479.

However, the PR itself is a second commit (c2) addressing F1/F2 from prior human review:
- **F1 (access log column shift):** Fixed — trace_id appended at end of line, not inserted mid-line. ✅
- **F2 (log injection via X-Request-ID):** Fixed — `reqid.Sanitize` strips control chars, bounds length. ✅
- **Probe error surfacing:** Fixed — `connProbeErr` captured and logged. ✅

---

## 8. Verdict

**No P0/P1 findings.** Two P2 cosmetic/documentation observations that do not affect correctness or security.

[Octo-Q] verdict: **APPROVE**

yujiawei (approved):
# Code Review — PR #479 (octo-server)

## Summary

This PR adds request-scoped tracing to diagnose the "invite members" latency. It introduces a small `pkg/reqid` package, a front-most gin middleware that mints/propagates a `trace_id` (reusing an inbound `X-Request-ID`), and threads that id through the four otherwise-disjoint slow logs (GIN access log, auth token parser, group handler, group service). It also adds per-segment timing: a connection-acquire probe in the handler, and a `db_ms` breakdown (`query_ms` / `genseq_ms` / `insert_ms` / `commit_ms`) plus an auth-parser breakdown (`token_ms` / `lang_ms` / `role_ms`).

Scope matches the title — purely additive instrumentation, no behavioral change to the invite path. Build compiles and the new `pkg/reqid` tests pass.

**Verdict: APPROVED.** No correctness, security, or data-loss blockers. A few non-blocking suggestions below.

## What I verified

- **Cross-file plumbing is sound.** `wkhttp.Context` embeds `*gin.Context`, so `c.GetString(reqid.GinKey)` in the handler works. `g.ctx.DB().Connection.DB` resolves through `dbr.Session → *Connection → *sql.DB`, so the connection probe compiles and is valid. The auth parser reads `reqid.FromContext(ctx)` from `c.Request.Context()`, which the reqid middleware populates via `WithContext`.
- **Middleware ordering is correct.** The reqid middleware is registered first (`route.UseGin`), before the i18n/tracer/metrics/access-log middlewares and before the per-route `AuthMiddleware`. So the `trace_id` is present in both `gin.Context` and the request `context.Context` by the time the access-log formatter and the token parser read it.
- **`genseq_ms` comment is accurate.** `octo-lib config.GenSeq` does take a process-level `sync.Mutex` (`seqLock`), so attributing lock contention to that segment is correct.
- **Access-log change is backward-compatible.** `trace_id` is appended at the end of the line (after the path), so column-positional downstream parsers are unaffected, and the suffix is omitted entirely when empty. Matches the stated F1 intent.

## Security notes (this PR was flagged for manual security review)

- **Inbound `X-Request-ID` is treated as untrusted and is correctly contained.** `reqid.Sanitize` trims, drops control chars (`< 0x20` and `0x7f`, including CR/LF/tab), and bounds length to 128. This prevents log-line forging/splitting via the access log and slow logs. The value is used only for log correlation — never for auth, routing, or cache keys — which is the right call.
- **Reflection back in the response header is safe.** The sanitized id is echoed in `X-Request-ID`; since control characters are stripped, there's no response-header-injection vector.
- **Residual (acceptable):** `Sanitize` keeps spaces and `|`. Since the access-log line uses `|` as a column separator, a client-supplied id containing `|`/spaces will render inside the `trace_id=` field. This cannot inject a new line or new columns (no newline survives sanitization), so it's cosmetic, not a security issue — fine to leave as-is.

1. **Connection-acquire probe runs on every invite, not just slow ones** (`modules/group/api.go` ~L1000–1012). The `sqlDB.Conn(ctx)` + `PingContext` adds an extra DB round-trip to *every* `memberAdd` request, while `conn_acquire_ms`/`conn_probe_err` are only emitted when the request crosses the slow threshold. For a short-lived diagnostic this is acceptable, but consider gating the probe behind the same env/threshold (or an explicit `DM_INVITE_CONNPROBE` flag) so steady-state traffic doesn't pay for a measurement it never logs — and plan to remove it once the latency source is found.
2. **`genseq_ms` / `insert_ms` are summed inside the per-member loop.** That's intentional and documented, but note the values are aggregates across N members, not per-member; the log field names read as totals, which is fine, just worth keeping in mind when interpreting (e.g. one slow member vs. many average ones is indistinguishable). No change required.
3. **`reqid.New()` rand-failure sentinel** returns a fixed `"0000000000000000"`. Correct choice (never fail a request for a log field), and the collision window is purely cosmetic for correlation. No action needed.

## Coverage / blind spots

- `pkg/reqid` is well unit-tested (uniqueness, hex shape, round-trip, sanitize cases including CRLF injection and truncation).
- Not exercised by tests (acceptable for an instrumentation PR): the middleware wiring in `main.go`, the handler conn-probe, and the service/auth timing fields — these are observability side-effects with no behavioral branch, so the risk of an untested regression is low. The main thing a human may want to eyeball post-merge is that the new slow-log fields actually populate as expected in a real environment.

Issue #480, opened 2026-06-26
perf: invite / member-list slowness — on-demand avatar rendering starves pod CPU under fanout


an9xyz:
## Summary

Under concurrent request bursts, authenticated endpoints (e.g.
`POST /v1/groups/:group_no/members`, `GET /v1/space/:id/members`,
`/membersync`) intermittently slow to several seconds, then recover on their
own once the burst subsides. Normal latency is tens of milliseconds.

The per-segment timing + `trace_id` added in #479 let us decompose a single
slow request end to end. The key finding: **the latency is dominated by the
number of serial Redis/DB round-trips per request, not by individual query
execution** — the database server is observed idle (`Threads_running` low,
no active queries, no lock waits) during the slow window, and the pod is
healthy (no CPU throttling, negligible GC). So each query executes fast on the
server; the cost is the per-round-trip wait, multiplied by ~15-20 serial hops.

A single authenticated invite request makes, roughly serially:

- global rate-limit token bucket (Redis)
- Auth: token (Redis) + language resolver (Redis→DB) + role resolver (Redis→DB)
- handler pre-checks: get_group, exist_member, bot_ownership, app_config, bot_space (loops per member)
- service: group, users, existing-members, blacklist reads + per-member GenSeq/insert + commit
- IM: add subscriber, send member-add, send cmd, add-to-threads (synchronous)

Under a burst these all queue, and because they are serial the cost compounds.

## Observed waterfall (one slow request, from #479 fields)

```
GIN total ............................ 8939ms
├─ non-Auth middleware (residual) .... ~2360ms
├─ Auth .............................   760ms   token 266 / lang 289 / role 204
└─ handler_total .................... 5819ms
    ├─ pre_checks ................... 1891ms   conn_acquire 145 / get_group 179 / exist_member 165
    │                                          / bot_ownership 248 / app_config 326 / bot_space 826
    └─ service ...................... 3928ms
        ├─ db_ms .................... 2954ms   query 1321 / space_check 491 / insert 512
        │                                      / commit 287 / genseq 0
        └─ IM .......................  ~972ms
```

### P0 — reduce & parallelize per-request round-trips (primary fix)

Because the DB server is idle during slow windows, tuning individual query
plans buys little; cutting the round-trip **count** and running independent
ones **concurrently** is the lever.

1. **De-duplicate the group lookup.** The handler reads the group
   (`getGroupInfo` → `QueryWithGroupNo`) and the service reads the same row
[...]

PR #478, opened 2026-06-26
feat(group): server-render default group avatars

an9xyz:
## Summary

Implements server-rendered default group avatars and replaces the legacy generated member-grid avatar path with product-aligned group avatar rendering.

## Related Issue

N/A

## Linked Spec

.octospec/tasks/group-default-avatar/brief.md

## Changes

- Add `avatar_text` and `avatar_color` fields for group create/update flows, including DB migration and validation.
- Render non-uploaded group avatars directly from `GET /v1/groups/:group_no/avatar` instead of redirecting to legacy generated member-grid images.
- Preserve uploaded group avatars by continuing to redirect when `is_upload_avatar=1`.
- Remove the member-avatar composite event chain so historical generated groups automatically fall back to server-rendered default avatars.
- Align group default avatar visuals with the product design: light fill, themed stroke/text, and two-tone group icon fallback.
- Bump group avatar ETag versions to avoid stale 304 responses after the visual rendering change.

## Testing

- [x] Unit tests added/updated
- [x] Manually verified

Commands run:

- `go test ./pkg/avatarrender/... -count=1`
- `go test ./modules/group/...`
- `go build ./...`
- `go vet ./pkg/avatarrender/... ./modules/group/...`
- `git diff --check`

Also generated group avatar samples for visual inspection under `.context/group-samples-new/`.

Not run:

- `golangci-lint run ./pkg/avatarrender/... ./modules/group/...` because `golangci-lint` is not installed in this local environment.

## COMPREHENSION

1. **What does this change actually do** to the load-bearing path?

   For groups without uploaded avatars, `avatarGet` now returns a deterministic PNG rendered by the server from group name/custom avatar text and group number/custom color. Uploaded avatars still use the existing object-storage redirect path. Historical generated member-grid groups keep `is_upload_avatar=0`, so they naturally switch to the server-rendered default avatar without a data migration.

2. **What could break** because of it (dependents + failure mode)?

   The main risk is avatar visual or cache drift: clients may receive a stale image if the ETag does not account for the rendering change, or uploaded avatars could be incorrectly treated as defaults if `is_upload_avatar` handling regresses. Another risk is rendering fallback failure for empty or non-renderable names. The implementation keeps uploaded avatars on the redirect path, bumps group avatar ETag versions, and covers text, color override, icon fallback, upload redirect, and 304 behavior in tests.

3. **How do you know it works** (specific test/repro/trace)?

[...]

PR #481, opened 2026-06-26
perf: shared avatar render cache to stop CPU starvation under fanout (#480)

an9xyz:
## Summary

Fixes the root cause behind #480 (invite / member-list slowness). On-demand
avatar rendering had **no server-side defense on the non-conditional GET path**.
Under a member-list fanout, hundreds of concurrent renders saturate the pod's
cores and starve co-scheduled request goroutines (e.g. the invite endpoint) on
the Go scheduler run queue — invisible to DB/network probes, which is why the
earlier SQL/round-trip investigation came up empty.

Controlled experiment that isolated it: toggling DevTools "Disable cache" **off**
dropped the same `POST /v1/groups/:group_no/members` from **8.7s → 47ms** with no
change to the invite path. "Disable cache" suppresses the avatar `If-None-Match`/304
fast path, so every avatar re-renders.

This PR adds a **process-global shared render cache** so a fanout can never
multiply renders or take all cores.

## Related Issue

- #480 (root cause + diagnosis in the issue thread)
- Coordinates with #478 (server-rendered group avatars) — see **Coordination** below.

## Changes

- `pkg/avatarrender/cache.go` — `Cache`: bytes LRU keyed by the same content
  factors as the ETag, **singleflight** to collapse concurrent cold renders, and
  a **render-concurrency semaphore** (default `max(1, GOMAXPROCS-1)`) so a fanout
  can never take all Ps. nil-receiver safe (degrades to direct render).
- `pkg/avatarrender/default.go` — process-global singleton: `GetOrRender` /
  `SetDefaultCache` / `ConfigFromEnv`. One shared LRU and — critically — **one
  semaphore** across all avatar endpoints. Per-endpoint caches would each cap
  renders at N, making the real process ceiling 2N and defeating the guard.
- `modules/user/api.go` — `UserAvatar` routes both renders (nameMode +
  ASCII fallback) through `avatarrender.GetOrRender`; counts 304s. The ETag/304
  fast path is unchanged — this protects the non-conditional GETs it can't.
- `pkg/metrics/avatar.go` — avatar metrics: cache hit/miss, render duration,
  inflight, semaphore wait, singleflight shared, 304 count.
- `main.go` — builds the single shared cache, injects metrics hooks (keeps
  `pkg/avatarrender` free of `pkg/metrics`), `SetDefaultCache`.

Tunable via `DM_AVATAR_CACHE_SIZE` / `DM_AVATAR_RENDER_MAX_CONCURRENCY`.

## Coordination with #478

#478 makes group `avatarGet` render on demand (`avatarrender.RenderGroup` /
`RenderIcon`). To stay protected, those two call sites should route through the
shared cache (keyed by the ETag #478 already computes):

```go
imageData, genErr := avatarrender.GetOrRender(etag, func() ([]byte, error) {
    return avatarrender.RenderGroup(text, style, avatarrender.DefaultSize)
})
iconData, iconErr := avatarrender.GetOrRender(iconEtag, func() ([]byte, error) {
    return avatarrender.RenderIcon(style)
})
```

Proposed order: **merge this fix first**, then #478 adapts (the new files here
don't conflict with #478 — different files in `pkg/avatarrender`).

## Testing

```
go build ./...
go vet ./pkg/avatarrender/ ./pkg/metrics/ ./modules/user/ .
go test ./pkg/avatarrender/ ./pkg/metrics/
go test ./modules/user/ -run 'NoLegacyResponseError|TestAvatarETag|TestIfNoneMatchSatisfied'
golangci-lint run ./pkg/avatarrender/... ./pkg/metrics/...   # 0 issues
```

- **Cache unit tests**: hit/miss, error-not-cached, singleflight (50 concurrent
  same-key → exactly 1 render), semaphore bounds concurrency, nil-receiver.
- **Metrics tests**: registration/counts, no-op when unregistered.
- **Repro + fix validation** (`GOMAXPROCS=2`, no external deps):
  `TestRenderFanoutStarvesVictim` — 16-render fanout inflates a simulated invite
  22ms → ~370ms (~16×); `TestCacheEliminatesStarvation` — routing through the real
  `avatarrender.Cache` returns it to ~1×.

Not run: `modules/user` / `modules/group` integration tests (need MySQL/Redis/WuKongIM,
not available in this environment); test binaries compile.

## COMPREHENSION

   Avatar default renders go through a process-global cache: LRU hit returns bytes
   with zero CPU; concurrent cold renders for the same key are coalesced by
   singleflight; the number of simultaneous real renders is bounded by a semaphore.
   The ETag/304 conditional fast path is untouched. Net effect: a member-list
   fanout triggers at most one render per distinct avatar instead of N, and can
   never consume all CPU.

   The cache key must capture all content-deciding factors; it reuses the exact
   factors the ETag already enumerates (mode-version / uid / text), so a rename or
   recolor produces a new key — no stale image. The render closure must be
   deterministic (it is). If cache construction ever failed, `SetDefaultCache` is
   not called and `GetOrRender` degrades to direct render (no behavior change, just
   no protection). A too-low semaphore could serialize cold renders, but cached
   content makes that one-time and keeps non-avatar traffic responsive (the
   explicit tradeoff; tunable via env).

   The controlled production experiment (disable-cache toggle: 8.7s ↔ 47ms) plus
   the deterministic `GOMAXPROCS=2` tests: a render fanout inflates a simulated
   invite ~16×, and routing the same fanout through `avatarrender.Cache` returns it
   to ~1×. Cache/singleflight/semaphore behaviors are pinned by unit tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TTmwnF18wfFgTAjufuMrXn)_

### DIFF
--- a/main.go
+++ b/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/modules/user"
 	"github.com/Mininglamp-OSS/octo-server/pkg/accesslog"
 	"github.com/Mininglamp-OSS/octo-server/pkg/auth"
+	"github.com/Mininglamp-OSS/octo-server/pkg/avatarrender"
 	octodb "github.com/Mininglamp-OSS/octo-server/pkg/db"
 	octoi18n "github.com/Mininglamp-OSS/octo-server/pkg/i18n"
 	"github.com/Mininglamp-OSS/octo-server/pkg/metrics"
@@ -208,6 +209,29 @@ func runAPI(ctx *config.Context) {
 	// scrape 时读取的 Collector,不起后台 goroutine。此处 rlRedis 与 ctx.DB().DB
 	// 均已就绪,且在 api.Route 之前完成注册。
 	metrics.NewDependencyMetrics(prometheus.DefaultRegisterer)
+	// 头像渲染缓存指标(issue#480):命中率 / 渲染耗时 / inflight / 信号量等待 /
+	// singleflight 合并 / 304。由 modules/user 的 avatarCache 经包级 Observe 函数灌入;
+	// 此处注册即可,未注册前这些 Observe 为 no-op,不影响已构造的 cache。
+	metrics.NewAvatarMetrics(prometheus.DefaultRegisterer)
+	// 构造进程级共享头像渲染缓存,并把观测 hooks 接到上面注册的头像指标。所有头像端点
+	// (user 的 UserAvatar;群组头像渲染合并后亦然——#478)经 avatarrender.GetOrRender
+	// 共用这一个实例:共享 LRU + 同一个渲染信号量(后者唯一,才是真正的进程级渲染并发
+	// 上限)。hooks 在组合根注入,使 pkg/avatarrender 不依赖 pkg/metrics。构造失败则不设
+	// 默认,GetOrRender 退化为直接渲染,不阻断启动。
+	avatarCfg := avatarrender.ConfigFromEnv()
+	avatarCfg.Hooks = avatarrender.Hooks{
+		OnHit:           metrics.ObserveAvatarCacheHit,
+		OnMiss:          metrics.ObserveAvatarCacheMiss,
+		OnShared:        metrics.ObserveAvatarSingleflightShared,
+		OnRender:        metrics.ObserveAvatarRender,
+		OnSemaphoreWait: metrics.ObserveAvatarSemaphoreWait,
+		OnInflight:      metrics.AddAvatarRenderInflight,
+	}
+	if avatarCache, err := avatarrender.NewCache(avatarCfg); err != nil {
+		log.Warn("构造头像渲染缓存失败,退化为无缓存", zap.Error(err))
+	} else {
+		avatarrender.SetDefaultCache(avatarCache)
+	}
 	// 把 octo-lib 客户端接缝的耗时回调接到 DependencyMetrics:
 	//   - MySQL: db.NewMySQL 的 connect/query 计时 → dependency="mysql"
 	//   - Redis: pkg/redis 客户端每条命令计时 → dependency="redis"
--- a/modules/user/api.go
+++ b/modules/user/api.go
@@ -23,6 +23,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/modules/source"
 	"github.com/Mininglamp-OSS/octo-server/pkg/avatarrender"
 	"github.com/Mininglamp-OSS/octo-server/pkg/avatarversion"
+	"github.com/Mininglamp-OSS/octo-server/pkg/metrics"
 	octoredis "github.com/Mininglamp-OSS/octo-server/pkg/redis"
 	spacepkg "github.com/Mininglamp-OSS/octo-server/pkg/space"
 	rd "github.com/go-redis/redis"
@@ -594,14 +595,22 @@ func (u *User) UserAvatar(c *wkhttp.Context) {
 			}
 			setAvatarHeaders(etag)
 			if ifNoneMatchSatisfied(c.GetHeader("If-None-Match"), etag) {
+				metrics.ObserveAvatarNotModified()
 				c.Status(http.StatusNotModified)
 				return
 			}
 
+			// 非条件 GET（disable-cache / 首屏 / 共享缓存 miss）会绕过上面的 304 快路径，
+			// 落到这里真渲染。成员列表扇出下大量并发非条件 GET 会把 CPU 打满、饿死同机
+			// 其它请求（issue#480）。渲染统一走 avatarCache：相同内容 key 命中复用字节、
+			// singleflight 合并并发冷渲染、渲染信号量限并发，确保一次扇出最多渲一张。
+			// key 复用上面算好的 ETag —— 它已覆盖决定图像内容的全部因子，语义与缓存一致。
 			if nameMode {
-				imageData, genErr := avatarrender.Render(avatarrender.Options{
-					Text: text,
-					Bg:   avatarrender.ColorForSeed(uid),
+				imageData, genErr := avatarrender.GetOrRender(etag, func() ([]byte, error) {
+					return avatarrender.Render(avatarrender.Options{
+						Text: text,
+						Bg:   avatarrender.ColorForSeed(uid),
+					})
 				})
 				if genErr == nil {
 					c.Data(http.StatusOK, "image/png", imageData)
@@ -611,7 +620,10 @@ func (u *User) UserAvatar(c *wkhttp.Context) {
 				u.Error("生成昵称默认头像失败，回退兜底", zap.Error(genErr), zap.String("uid", uid))
 				c.Header("ETag", avatarETag("ascii-v1", uid))
 			}
-			imageData, genErr := generateDefaultAvatar(uid)
+			asciiETag := avatarETag("ascii-v1", uid)
+			imageData, genErr := avatarrender.GetOrRender(asciiETag, func() ([]byte, error) {
+				return generateDefaultAvatar(uid)
+			})
 			if genErr != nil {
 				u.Error("生成默认头像失败", zap.Error(genErr))
 				c.Writer.WriteHeader(http.StatusInternalServerError)
--- a/modules/user/api.go
+++ b/modules/user/api.go
@@ -602,11 +602,15 @@ func (u *User) UserAvatar(c *wkhttp.Context) {
 
 			// 非条件 GET（disable-cache / 首屏 / 共享缓存 miss）会绕过上面的 304 快路径，
 			// 落到这里真渲染。成员列表扇出下大量并发非条件 GET 会把 CPU 打满、饿死同机
-			// 其它请求（issue#480）。渲染统一走 avatarCache：相同内容 key 命中复用字节、
+			// 其它请求（issue#480）。渲染统一走共享缓存：相同内容 key 命中复用字节、
 			// singleflight 合并并发冷渲染、渲染信号量限并发，确保一次扇出最多渲一张。
-			// key 复用上面算好的 ETag —— 它已覆盖决定图像内容的全部因子，语义与缓存一致。
+			//
+			// 缓存 key 用 avatarCacheKey（完整原始因子），**不是**上面的 CRC32 ETag：
+			// ETag 是 32 位弱指纹，作共享缓存身份会碰撞→跨用户串图（PR#481 评审）。
+			// 两者覆盖同一组因子（渲染版本/uid→色/文字），仅 ETag 头继续用 CRC32。
 			if nameMode {
-				imageData, genErr := avatarrender.GetOrRender(etag, func() ([]byte, error) {
+				nameKey := avatarCacheKey("name-v3", uid, text)
+				imageData, genErr := avatarrender.GetOrRender(nameKey, func() ([]byte, error) {
 					return avatarrender.Render(avatarrender.Options{
 						Text: text,
 						Bg:   avatarrender.ColorForSeed(uid),
@@ -620,8 +624,8 @@ func (u *User) UserAvatar(c *wkhttp.Context) {
 				u.Error("生成昵称默认头像失败，回退兜底", zap.Error(genErr), zap.String("uid", uid))
 				c.Header("ETag", avatarETag("ascii-v1", uid))
 			}
-			asciiETag := avatarETag("ascii-v1", uid)
-			imageData, genErr := avatarrender.GetOrRender(asciiETag, func() ([]byte, error) {
+			asciiKey := avatarCacheKey("ascii-v1", uid)
+			imageData, genErr := avatarrender.GetOrRender(asciiKey, func() ([]byte, error) {
 				return generateDefaultAvatar(uid)
 			})
 			if genErr != nil {
--- a/modules/user/avatar_etag_test.go
+++ b/modules/user/avatar_etag_test.go
@@ -100,6 +100,20 @@ func TestAvatarCacheKeyDistinguishesFactors(t *testing.T) {
 	}
 }
 
+// TestAvatarCacheKeyInjectiveAcrossPartBoundaries pins the PR#481 hardening:
+// the key encoding must stay injective even when a part contains the separator
+// byte. A naive strings.Join(parts, "\x00") would map these two distinct
+// factor lists to the same key; length-framing keeps them distinct.
+func TestAvatarCacheKeyInjectiveAcrossPartBoundaries(t *testing.T) {
+	if avatarCacheKey("name-v3", "u", "a\x00b") == avatarCacheKey("name-v3", "u\x00a", "b") {
+		t.Fatal("cache key must stay injective across part boundaries (NUL in a part)")
+	}
+	// Empty-vs-absent part must also not alias.
+	if avatarCacheKey("x", "") == avatarCacheKey("x") {
+		t.Fatal("trailing empty part must change the key")
+	}
+}
+
 func TestIfNoneMatchSatisfied(t *testing.T) {
 	etag := `W/"abc12345"`
 	tests := []struct {
--- a/modules/user/avatar_etag_test.go
+++ b/modules/user/avatar_etag_test.go
@@ -1,8 +1,11 @@
 package user
 
 import (
+	"math/rand"
 	"strings"
 	"testing"
+
+	"github.com/Mininglamp-OSS/octo-server/pkg/avatarrender"
 )
 
 func TestAvatarETag(t *testing.T) {
@@ -26,6 +29,74 @@ func TestAvatarETag(t *testing.T) {
 	got := avatarETag("name-v1", "uid1", "三丰")
 	if !strings.HasPrefix(got, `W/"`) || !strings.HasSuffix(got, `"`) {
 		t.Fatalf("avatarETag should be a quoted weak ETag, got %s", got)
+	}
+}
+
+// TestAvatarCacheKeyResistsETagCollision pins the PR#481 fix: the shared render
+// cache key must NOT be the 32-bit CRC32 ETag. We search for two distinct display
+// texts whose avatarETag collides, then assert (a) their avatarCacheKey does NOT
+// collide and (b) a real shared avatarrender.Cache does not cross-serve A's bytes
+// to B. Before the fix the ETag was the cache key, so a collision meant user B was
+// served user A's cached avatar.
+//
+// Note: CRC32 is linear, so structured/sequential inputs barely collide — we use a
+// fixed-seed PRNG over fixed-length random text, where collisions follow the 32-bit
+// birthday bound (~tens of thousands of tries; 1<<20 budget is ample). Fixed seed
+// keeps it deterministic.
+func TestAvatarCacheKeyResistsETagCollision(t *testing.T) {
+	const uid = "u_collide"
+	r := rand.New(rand.NewSource(42))
+	seen := make(map[string]string) // etag -> text
+	var a, b string
+	for i := 0; i < 1<<20 && a == ""; i++ {
+		buf := make([]byte, 8)
+		for j := range buf {
+			buf[j] = byte('a' + r.Intn(26))
+		}
+		text := string(buf)
+		et := avatarETag("name-v3", uid, text)
+		if prev, ok := seen[et]; ok && prev != text {
+			a, b = prev, text
+		} else {
+			seen[et] = text
+		}
+	}
+	if a == "" {
+		t.Skip("no CRC32 ETag collision found within budget (unexpected for 32-bit)")
+	}
+	// Precondition: the weak ETags really do collide — that's the hazard.
+	if avatarETag("name-v3", uid, a) != avatarETag("name-v3", uid, b) {
+		t.Fatalf("expected colliding ETags for %q/%q", a, b)
+	}
+	// The fix: cache keys for distinct inputs must stay distinct despite the ETag collision.
+	keyA := avatarCacheKey("name-v3", uid, a)
+	keyB := avatarCacheKey("name-v3", uid, b)
+	if keyA == keyB {
+		t.Fatalf("cache key collides for distinct inputs %q/%q sharing a CRC32 ETag", a, b)
+	}
+	// End-to-end: a real shared cache must not serve B user A's cached bytes.
+	cache, err := avatarrender.NewCache(avatarrender.Config{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	bytesA, _ := cache.GetOrRender(keyA, func() ([]byte, error) { return []byte("avatar-A:" + a), nil })
+	bytesB, _ := cache.GetOrRender(keyB, func() ([]byte, error) { return []byte("avatar-B:" + b), nil })
+	if string(bytesA) == string(bytesB) {
+		t.Fatalf("shared cache cross-served: B got A's bytes for ETag-colliding inputs %q/%q", a, b)
+	}
+}
+
+// TestAvatarCacheKeyDistinguishesFactors:不同模式/uid/文字都应得到不同 key。
+func TestAvatarCacheKeyDistinguishesFactors(t *testing.T) {
+	base := avatarCacheKey("name-v3", "uid1", "三丰")
+	if base == avatarCacheKey("name-v3", "uid1", "丰丰") {
+		t.Fatal("text change must change cache key")
+	}
+	if base == avatarCacheKey("name-v3", "uid2", "三丰") {
+		t.Fatal("uid change must change cache key")
+	}
+	if avatarCacheKey("name-v3", "uid1") == avatarCacheKey("ascii-v1", "uid1") {
+		t.Fatal("render mode must change cache key")
 	}
 }
 
--- a/modules/user/avatar_path.go
+++ b/modules/user/avatar_path.go
@@ -3,6 +3,7 @@ package user
 import (
 	"fmt"
 	"hash/crc32"
+	"strconv"
 	"strings"
 )
 
@@ -36,10 +37,21 @@ func avatarETag(parts ...string) string {
 // 作弱 ETag 没问题（HTTP 304 撞了顶多多渲一次，无害）；但缓存 key 是**跨所有用户
 // 的进程级 []byte 存储身份**，一旦碰撞就会把 A 用户已缓存的头像返回给 B 用户
 // （串图 / 轻度信息泄露）。而且 text=昵称末两字用户可控、CRC32 线性可构造，可被
-// 对抗性投毒。故 key 用 NUL 分隔的完整原始 parts（单射），把碰撞域从 2^32 降到
-// 实际不可碰撞；ETag 头仍用 CRC32 不变。详见 PR#481 评审。
+// 对抗性投毒。故 key 编码完整原始 parts 而非摘要。详见 PR#481 评审。
+//
+// 编码用**长度分帧**（每段 `<len>:<bytes>`）而非 NUL 分隔：后者对任意 parts 非
+// 单射（如 ["u","a\x00b"] 与 ["u\x00a","b"] 会撞）。今天的因子（uid 来自 HTTP path
+// 不含 NUL、text 经 IndividualText 已剥离控制字符）即便用 NUL 也安全，但长度分帧对
+// **任意**字节都单射，避免将来新增可能含 NUL 的因子（或 #478 群头像接入同一 key
+// 构造）时踩雷（PR#481 评审）。
 func avatarCacheKey(parts ...string) string {
-	return strings.Join(parts, "\x00")
+	var b strings.Builder
+	for _, p := range parts {
+		b.WriteString(strconv.Itoa(len(p)))
+		b.WriteByte(':')
+		b.WriteString(p)
+	}
+	return b.String()
 }
 
 // ifNoneMatchSatisfied 报告 If-None-Match 头是否匹配 etag（RFC 7232 §3.2 弱比较：
--- a/modules/user/avatar_path.go
+++ b/modules/user/avatar_path.go
@@ -29,6 +29,19 @@ func avatarETag(parts ...string) string {
 	return fmt.Sprintf(`W/"%08x"`, h.Sum32())
 }
 
+// avatarCacheKey 为进程级共享渲染缓存（avatarrender.Cache）算 key。parts 与
+// avatarETag 取同一组决定图像内容的因子（渲染版本、uid→颜色、展示文字）。
+//
+// 关键：这里**不能**复用 avatarETag 的 CRC32 摘要。avatarETag 是 32 位 CRC32，
+// 作弱 ETag 没问题（HTTP 304 撞了顶多多渲一次，无害）；但缓存 key 是**跨所有用户
+// 的进程级 []byte 存储身份**，一旦碰撞就会把 A 用户已缓存的头像返回给 B 用户
+// （串图 / 轻度信息泄露）。而且 text=昵称末两字用户可控、CRC32 线性可构造，可被
+// 对抗性投毒。故 key 用 NUL 分隔的完整原始 parts（单射），把碰撞域从 2^32 降到
+// 实际不可碰撞；ETag 头仍用 CRC32 不变。详见 PR#481 评审。
+func avatarCacheKey(parts ...string) string {
+	return strings.Join(parts, "\x00")
+}
+
 // ifNoneMatchSatisfied 报告 If-None-Match 头是否匹配 etag（RFC 7232 §3.2 弱比较：
 // 忽略 W/ 前缀）。支持逗号分隔的多个验证符与通配 "*"。
 func ifNoneMatchSatisfied(header, etag string) bool {
--- a/pkg/avatarrender/cache.go
+++ b/pkg/avatarrender/cache.go
@@ -1,0 +1,161 @@
+package avatarrender
+
+import (
+	"time"
+
+	lru "github.com/hashicorp/golang-lru/v2"
+	"golang.org/x/sync/singleflight"
+)
+
+// 头像「按需渲染」在成员列表扇出下会把 pod 的 CPU 打满,饿死同机的其它请求
+// goroutine(见 issue#480)。Cache 用三层防御把"一次扇出 N 张 = N 次渲染"压成
+// 最多一次:
+//
+//   - bytes LRU:渲染结果按内容 key 缓存,命中直接复用字节,零 CPU;
+//   - singleflight:同 key 的并发请求只触发一次渲染,其余复用同一结果(防冷启动
+//     惊群——首屏 N 个并发请求打到同一张未缓存头像时只渲一次);
+//   - render 信号量:限制全局并发渲染数,即使大量"不同 key"的冷头像同时到达,也不
+//     会占满所有 P 把其它流量饿死(信号量必须配合前两层,否则只是把惊群变成慢排队)。
+//
+// key 由调用方按"决定图像内容的全部因子"生成(与 ETag 同源),Cache 不关心其语义。
+
+// DefaultCacheSize 是未配置时的 LRU 容量(条目数)。默认头像是 200×200 的简单 PNG
+// (数 KB 量级),数千条目的常驻内存在十几 MB 级,足够覆盖活跃用户集。
+const DefaultCacheSize = 4096
+
+// Hooks 是可观测性回调,全部可选(nil 即 no-op)。刻意用函数字段而非接口,使本包
+// 不依赖 pkg/metrics —— 由上层(modules/user)在构造时把 prometheus 观测点接进来。
+type Hooks struct {
+	// OnHit / OnMiss:LRU 命中 / 未命中各记一次。
+	OnHit  func()
+	OnMiss func()
+	// OnShared:本次渲染结果被并发调用方复用(singleflight 合并),记一次。
+	OnShared func()
+	// OnRender:一次真实渲染完成,带耗时与错误(用于渲染延迟直方图)。
+	OnRender func(dur time.Duration, err error)
+	// OnSemaphoreWait:获取渲染信号量的等待耗时(>0 说明渲染并发已饱和、在排队)。
+	OnSemaphoreWait func(dur time.Duration)
+	// OnInflight:正在进行的渲染数变化(+1 开始 / -1 结束),用于 inflight gauge。
+	OnInflight func(delta float64)
+}
+
+func (h *Hooks) normalize() {
+	if h.OnHit == nil {
+		h.OnHit = func() {}
+	}
+	if h.OnMiss == nil {
+		h.OnMiss = func() {}
+	}
+	if h.OnShared == nil {
+		h.OnShared = func() {}
+	}
+	if h.OnRender == nil {
+		h.OnRender = func(time.Duration, error) {}
+	}
+	if h.OnSemaphoreWait == nil {
+		h.OnSemaphoreWait = func(time.Duration) {}
+	}
+	if h.OnInflight == nil {
+		h.OnInflight = func(float64) {}
+	}
+}
+
+// Config 配置 Cache。
+type Config struct {
+	// Size 是 LRU 最大条目数;<=0 用 DefaultCacheSize。
+	Size int
+	// MaxConcurrentRenders 是同时进行的真实渲染数上限(信号量容量);<=0 表示不限。
+	// 取值应 < GOMAXPROCS,给非头像流量留核;配合缓存后,渲染只在冷 key 时发生,
+	// 串行化冷渲染的代价是一次性的(之后命中缓存),换来洪峰下不饿死其它请求。
+	MaxConcurrentRenders int
+	// Hooks 是可观测性回调。
+	Hooks Hooks
+}
+
+// Cache 是带 singleflight 与渲染信号量的头像 bytes 缓存。零值不可用,须经 NewCache 构造;
+// 方法对 nil 接收者安全(直接渲染,无缓存),便于在未配置缓存时退化。
+type Cache struct {
+	lru   *lru.Cache[string, []byte]
+	group singleflight.Group
+	sem   chan struct{}
+	hooks Hooks
+}
+
+// NewCache 构造一个 Cache。
+func NewCache(cfg Config) (*Cache, error) {
+	size := cfg.Size
+	if size <= 0 {
+		size = DefaultCacheSize
+	}
+	l, err := lru.New[string, []byte](size)
+	if err != nil {
+		return nil, err
+	}
+	cfg.Hooks.normalize()
+	c := &Cache{lru: l, hooks: cfg.Hooks}
+	if cfg.MaxConcurrentRenders > 0 {
+		c.sem = make(chan struct{}, cfg.MaxConcurrentRenders)
+	}
+	return c, nil
+}
+
+// GetOrRender 返回 key 对应的头像字节:命中 LRU 直接返回;否则经 singleflight 合并、
+// 在渲染信号量约束下调用 render 生成一次,成功后写入 LRU。render 必须是确定性的
+// (相同 key → 相同字节),否则缓存会返回过期内容。
+//
+// nil 接收者时直接调用 render(无缓存/合并/限流),用于缓存未配置的退化路径。
+func (c *Cache) GetOrRender(key string, render func() ([]byte, error)) ([]byte, error) {
+	if c == nil {
+		return render()
+	}
+	if v, ok := c.lru.Get(key); ok {
+		c.hooks.OnHit()
+		return v, nil
+	}
+	c.hooks.OnMiss()
+
+	v, err, shared := c.group.Do(key, func() (interface{}, error) {
+		// 双检:在等待 singleflight 期间,可能已有同 key 的先行渲染填好了缓存。
+		if v, ok := c.lru.Get(key); ok {
+			return v, nil
+		}
+		b, rerr := c.renderWithSem(render)
+		if rerr != nil {
+			return nil, rerr
+		}
+		c.lru.Add(key, b)
+		return b, nil
+	})
+	if shared {
+		c.hooks.OnShared()
+	}
+	if err != nil {
+		return nil, err
+	}
+	return v.([]byte), nil
+}
+
+// renderWithSem 在渲染信号量(若配置)约束下执行一次渲染,并打点等待/耗时/inflight。
+func (c *Cache) renderWithSem(render func() ([]byte, error)) ([]byte, error) {
+	if c.sem != nil {
+		start := time.Now()
+		c.sem <- struct{}{}
+		c.hooks.OnSemaphoreWait(time.Since(start))
+		defer func() { <-c.sem }()
+	}
+	c.hooks.OnInflight(1)
+	defer c.hooks.OnInflight(-1)
+
+	start := time.Now()
+	b, err := render()
+	c.hooks.OnRender(time.Since(start), err)
+	return b, err
+}
+
+// Len 返回当前缓存的条目数(测试/诊断用)。nil 接收者返回 0。
+func (c *Cache) Len() int {
+	if c == nil {
+		return 0
+	}
+	return c.lru.Len()
+}
--- a/pkg/avatarrender/cache.go
+++ b/pkg/avatarrender/cache.go
@@ -1,6 +1,7 @@
 package avatarrender
 
 import (
+	"fmt"
 	"time"
 
 	lru "github.com/hashicorp/golang-lru/v2"
@@ -19,8 +20,12 @@ import (
 //
 // key 由调用方按"决定图像内容的全部因子"生成(与 ETag 同源),Cache 不关心其语义。
 
-// DefaultCacheSize 是未配置时的 LRU 容量(条目数)。默认头像是 200×200 的简单 PNG
-// (数 KB 量级),数千条目的常驻内存在十几 MB 级,足够覆盖活跃用户集。
+// DefaultCacheSize 是未配置时的 LRU 容量(条目数)。
+//
+// 注意:LRU 按**条目数**而非字节计 bound。当前所有渲染输出都是 200×200 的简单 PNG
+// (数 KB 量级),故 4096 条目 ≈ 十几 MB,足够覆盖活跃用户集。若将来有调用方(如
+// #478 群头像)缓存尺寸差异很大的图,常驻内存会随 entries×单图尺寸增长,届时应改为
+// 按字节预算的缓存或下调容量(PR#481 评审)。
 const DefaultCacheSize = 4096
 
 // Hooks 是可观测性回调,全部可选(nil 即 no-op)。刻意用函数字段而非接口,使本包
@@ -136,7 +141,7 @@ func (c *Cache) GetOrRender(key string, render func() ([]byte, error)) ([]byte, 
 }
 
 // renderWithSem 在渲染信号量(若配置)约束下执行一次渲染,并打点等待/耗时/inflight。
-func (c *Cache) renderWithSem(render func() ([]byte, error)) ([]byte, error) {
+func (c *Cache) renderWithSem(render func() ([]byte, error)) (b []byte, err error) {
 	if c.sem != nil {
 		start := time.Now()
 		c.sem <- struct{}{}
@@ -146,10 +151,17 @@ func (c *Cache) renderWithSem(render func() ([]byte, error)) ([]byte, error) {
 	c.hooks.OnInflight(1)
 	defer c.hooks.OnInflight(-1)
 
+	// 用 defer 记录渲染耗时/结果,这样 render() panic 时指标也能落点(否则 Prometheus
+	// 看不到系统性渲染崩溃)。捕获后原样重抛,保持 panic 传播语义不变(PR#481 评审 P2)。
 	start := time.Now()
-	b, err := render()
-	c.hooks.OnRender(time.Since(start), err)
-	return b, err
+	defer func() {
+		if r := recover(); r != nil {
+			c.hooks.OnRender(time.Since(start), fmt.Errorf("avatarrender: render panicked: %v", r))
+			panic(r)
+		}
+		c.hooks.OnRender(time.Since(start), err)
+	}()
+	return render()
 }
 
 // Len 返回当前缓存的条目数(测试/诊断用)。nil 接收者返回 0。
--- a/pkg/avatarrender/cache.go
+++ b/pkg/avatarrender/cache.go
@@ -119,11 +119,16 @@ func (c *Cache) GetOrRender(key string, render func() ([]byte, error)) ([]byte, 
 	}
 	c.hooks.OnMiss()
 
+	// ranRender 标记本次调用是否真的执行了渲染。singleflight 只会运行 leader 的闭包,
+	// 故只有 leader 这里被置 true;等待者的闭包不运行,保持 false。各 goroutine 持有
+	// 自己的局部 ranRender,无跨协程访问,无需同步。
+	ranRender := false
 	v, err, shared := c.group.Do(key, func() (interface{}, error) {
 		// 双检:在等待 singleflight 期间,可能已有同 key 的先行渲染填好了缓存。
 		if v, ok := c.lru.Get(key); ok {
 			return v, nil
 		}
+		ranRender = true
 		b, rerr := c.renderWithSem(render)
 		if rerr != nil {
 			return nil, rerr
@@ -131,7 +136,10 @@ func (c *Cache) GetOrRender(key string, render func() ([]byte, error)) ([]byte, 
 		c.lru.Add(key, b)
 		return b, nil
 	})
-	if shared {
+	// singleflight 的 shared 对 leader 也为 true。OnShared 只应计"被合并到他人渲染
+	// 上的请求"(真正搭便车的等待者),故排除自己执行了渲染的 leader,避免每次 flight
+	// 多计一次(PR#481 评审)。
+	if shared && !ranRender {
 		c.hooks.OnShared()
 	}
 	if err != nil {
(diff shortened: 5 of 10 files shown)


Jerry-Xin (approved):
**APPROVE** — Shared, process-global avatar render cache (LRU + singleflight + render semaphore) that correctly addresses the #480 CPU-starvation-under-fanout root cause. Concurrency-safe, memory-bounded, errors not cached, key reuses the existing ETag. Verified by reading head `637432dc` and running build/vet/`go test -race` locally.

### ✅ Verification (executed against head SHA 637432dc990c)
- `go build ./...` → exit 0
- `go vet ./pkg/avatarrender/ ./pkg/metrics/ ./modules/user/ .` → exit 0
- `go test -race -count=1 ./pkg/avatarrender/ ./pkg/metrics/` → **PASS** (avatarrender 27.5s, metrics 2.2s), exit 0 — no data races.

### ✅ Defense correctness (the load-bearing claims)
- **Concurrency / data race**: all map state behind `hashicorp/golang-lru/v2` (internally locked) + `singleflight.Group`. The cold path does a **double-check** of the LRU inside `group.Do` before rendering. `-race` clean across the singleflight (50-goroutine) and semaphore (20-distinct-key) tests.
- **Thundering-herd actually fixed**: `singleflight` collapses N concurrent misses for the same key into **exactly 1 render** — proven by `TestCacheSingleflight` (50 concurrent same-key → renders==1). Plus a process-global render semaphore (`default max(1, GOMAXPROCS-1)`) caps simultaneous cold renders so a distinct-key flood can't take all Ps (`TestCacheSemaphoreBoundsConcurrency`).
- **Process-global singleton** (`default.go`, `atomic.Pointer`): one LRU + **one** semaphore shared by all avatar endpoints — correctly avoids the per-endpoint 2N ceiling pitfall called out in the PR body.
- **Memory bound**: LRU is entry-count bounded (`DefaultCacheSize=4096`, tunable `DM_AVATAR_CACHE_SIZE`). A distinct-key/cache-busting flood evicts oldest entries — no unbounded growth / OOM.
- **Cache key correctness**: key == the same ETag used by the 304 fast path = `avatarETag("name-v3", uid, text)` (or `"ascii-v1", uid`). Covers mode-version + uid (color seed) + display text, so a rename produces a new key → no stale image. Version prefix (`name-v3`/`ascii-v1`) gives a clean invalidation lever.
- **Negative caching**: errors are **not** cached — render error returns before `lru.Add`, verified by `TestCacheErrorNotCached` (retry succeeds). A failed name render degrades to the ASCII fallback; no bad entry pinned.
- **No goroutine leak**: LRU eviction is synchronous (no background janitor goroutine), so nothing to leak/stop. nil-receiver path degrades to direct render (`TestCacheNilReceiver`), so a cache-construction failure at startup just disables protection without blocking boot (`main.go` logs and continues).
- **Metrics** (`pkg/metrics/avatar.go`): low-cardinality labels only (result/status), `atomic.Pointer` default, no-op when unregistered — confirmed by `TestAvatarObserve_NoDefaultIsNoop`.

### 💬 Non-blocking notes
- **Scope clarity**: this PR routes the **user** `UserAvatar` path through the cache; the group `avatarGet` (#478) is intentionally **not** modified here (coordination note only). So #478's disband/nil → 404 guard is untouched and unaffected by this change. When #478 adopts `GetOrRender`, ensure its ETag key incorporates the group GenSeq/version so a custom-avatar (avatar_text/avatar_color) change isn't masked by a stale hit, and that the disband/nil 404 check stays **before** the cache fill.
- **Bound is by entry count, not bytes** (`Cache.lru[string,[]byte]`): fine for the small fixed-size default PNGs described, but if future renders vary widely in size, a byte-budget LRU would bound memory more precisely. Current default (~4096 × few KB ≈ tens of MB) is reasonable.

### 🔵 Merge gates (not code defects)
- `check-sprint / check-sprint` = **FAILURE**; CI Build/Test/Vet/Lint and `code-review` still **IN_PROGRESS/PENDING** at review time. mergeable=MERGEABLE, reviewDecision=REVIEW_REQUIRED. Resolve sprint/CI gates before merge.

🔵 Per PR body, `modules/user` / `modules/group` integration tests need MySQL/Redis/WuKongIM (unavailable here) — not run locally; test binaries compile and the cache/metrics packages pass under `-race`.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#481 Review Report
**Head SHA:** `637432dc990cd840f8175decab54ef8907e31608`
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/481
**Title:** perf: shared avatar render cache to stop CPU starvation under fanout (#480)

---

## 1. 验证结论

| 验证项 | 结论 | 证据 |
|--------|------|------|
| 缓存正确性（LRU hit/miss） | ✅ | `cache.go:115-122` — LRU Get 命中返回 bytes，未命中走 singleflight |
| Singleflight 合并 | ✅ | `cache.go:124-140` — `group.Do(key, fn)` 正确包裹渲染，并发同 key 只渲染一次 |
| 渲染信号量限并发 | ✅ | `cache.go:143-154` — buffered channel acquire/release，defer 保证释放 |
| Nil receiver 退化 | ✅ | `cache.go:113-115` + `default.go:33-34` — `nil.GetOrRender` 直接调 render，不 panic |
| ETag-as-key 一致性 | ✅ | `api.go` — nameMode key = `avatarETag("name-v3", uid, text)`，ASCII key = `avatarETag("ascii-v1", uid)`，均覆盖决定内容的全部因子 |
| 错误不缓存 | ✅ | `cache.go:133-135` — `rerr != nil` 时不写 LRU，下次重试 |
| 进程级唯一信号量 | ✅ | `default.go:19` — `atomic.Pointer[Cache]` 持有唯一实例，`main.go` 构造一次 |
| Metrics no-op 安全 | ✅ | `avatar.go` — 所有 Observe 函数先 Load atomic pointer，nil 时 skip |
| 启动退化 | ✅ | `main.go:229-232` — `NewCache` 失败仅 log.Warn，不调 SetDefaultCache，GetOrRender 退化为直接渲染 |
| 依赖可用 | ✅ | `go.mod` 已有 `hashicorp/golang-lru/v2 v2.0.7` + `golang.org/x/sync v0.19.0` |
| 测试覆盖 | ✅ | cache_test.go（hit/miss、error、singleflight、semaphore、nil-receiver）+ starvation_repro_test.go（问题复现 + 修复验证）+ avatar_test.go（metrics 注册 + no-op） |

## 2. 发现的问题

### P2-1: Singleflight OnShared 指标低估

**Diff-scope 三问：** new (本 PR 新引入) / 不影响功能正确性

**位置：** `cache.go:124-138`

singleflight 的 `group.Do(key, fn)` 返回 `(v, err, shared)` 中，`shared=true` 仅对本次 flight 的**等待者**成立。Flight leader（发起渲染的 goroutine）拿到 `shared=false`，即使其渲染结果被 N 个 waiter 复用。此外，late-comer 在 double-check 路径（`cache.go:126-128`）命中 LRU 直接返回，完全绕过 singleflight，也不触发 OnShared。

**影响：** 可观测性指标低估了实际被合并的请求数。不影响缓存功能或 CPU 保护效果。

**建议：** 在 `group.Do` 返回后，无论 `shared` 值，额外检查本次请求是否真的做了渲染（可通过在 closure 内设 atomic flag 判断）。如果 closure 未被调用（waiter）或 double-check 命中（late-comer），都算 shared。这是一个 nice-to-have 的可观测性改进，不阻塞。

### P2-2: 缓存无 TTL / 主动失效机制

**Diff-scope 三问：** new (本 PR 新增代码) / LRU-only 设计无 TTL

**位置：** `cache.go` 全文

Cache 是纯 LRU，无 TTL 或主动失效。如果用户改名或换色，新的 ETag 产生新 key，旧 entry 只靠 LRU 容量淘汰。4096 默认容量下，不活跃用户的旧 entry 可能长期驻留。

**影响：** 内存效率轻微浪费，不影响正确性（旧 key 不会被再次访问，因为 ETag 已变）。对于活跃用户集远小于 4096 的场景可忽略。

**建议：** 可考虑后续版本加 entry-level TTL（如 30min），但当前不阻塞。

### P2-3: Starvation repro test 的 CI 稳定性

**Diff-scope 三问：** new (新增测试)

**位置：** `starvation_repro_test.go:114-156`

`TestRenderFanoutStarvesVictim` 断言 `ratio >= 2.0`，依赖运行时调度行为。在 overloaded CI runner 或高核数机器上，GOMAXPROCS=2 可能被 cgroup 覆盖或受其他负载影响。阈值 2x 已较保守（实测通常 >>2x），但 timing test 固有抖动不可忽视。

**建议：** 已有 `testing.Short()` 跳过，可接受。如 CI 偶发 fail，可考虑放宽到 1.5x 或增加 warmup 轮次。

## 3. 建议

- **P2-1** 可作为 follow-up：改进 OnShared 计数逻辑，使 dashboard 更准确
- **P2-2** 长期考虑 entry TTL，但 LRU-only 在当前容量下足够
- #478 合并后确保 group avatar 的 `RenderGroup`/`RenderIcon` 也走 `GetOrRender`（PR body 已明确 coordination plan，✅）

## 4. 额外发现

- **无** — 未发现安全问题、跨路径不一致、或数据流断裂

## 5. 数据流回溯

| 消费点 | 数据 | 上游来源 | 运行时数据是否真的流到消费点 |
|--------|------|----------|------------------------------|
| `GetOrRender(etag, ...)` | etag（缓存 key） | `avatarETag("name-v3"/"ascii-v1", uid, text)` @ `api.go:588-591` | ✅ 在 304 检查之后才进入 GetOrRender，key 非空且内容决定 |
| `Render(Options{Text: text, Bg: ColorForSeed(uid)})` | text | `avatarrender.IndividualText(userInfo.Name)` @ `api.go:583` | ✅ `Renderable(text)` 已校验（nameMode=true 才进此分支） |
| `Render(Options{...})` | Bg | `avatarrender.ColorForSeed(uid)` | ✅ 确定性函数，uid 来自路由参数 |
| `generateDefaultAvatar(uid)` | uid | 路由参数 `c.Param("uid")` | ✅ ASCII fallback 路径，uid 非空（前面已查库验证用户存在） |
| `c.lru.Get(key)` | key | 调用方传入的 etag | ✅ key 类型 string，LRU 内部 mutex 保护并发 |
| `renderWithSem(render)` | render closure | 调用方传入 | ✅ closure 捕获 text/uid/bg，均为值拷贝，无共享可变状态 |
| `c.hooks.OnHit/OnMiss/OnRender/...` | hooks | `NewCache` 时 `cfg.Hooks.normalize()` 填充 nil → no-op | ✅ normalize 在构造时一次性完成，运行时无 nil panic 风险 |
| `metrics.ObserveAvatarNotModified()` | 无参数 | 304 路径直接调用 | ✅ atomic pointer Load，nil 时 skip |

## 6. 盲点 checklist（R5）

- **C1 双路径 parity：** Clear。本 PR 不涉及对称路径（add/remove 等）。nameMode 和 ASCII fallback 两条渲染路径均已一致包裹 GetOrRender。
- **C2 control-flow ordering / 嵌套复用：** Clear。GetOrRender 在两个调用点使用，ordering 一致（304 check → GetOrRender）。singleflight 内部 double-check 正确。
- **C3 授权边界 ≠ 能力边界：** N/A。无权限/工具暴露变更。
- **C4 授权生命周期 / 容器-成员状态级联：** N/A。无鉴权变更。
- **C5 build/note 通过 ≠ 运行期路径正确：** N/A。无构建产物/扩展/CLI 变更。纯 Go 服务端代码，build 通过即运行期路径一致。
- **C6 治理/策略/安全文档自洽性：** N/A。无文档变更。

## 7. 跨轮 blocker 复检

N/A — 首次 review。

---

**[Octo-Q] verdict: APPROVE**

本 PR 架构清晰（LRU + singleflight + semaphore 三层防御），数据流完整，nil-receiver 退化安全，测试覆盖充分（包括 starvation 复现 + 修复验证）。三个 P2 发现均为可观测性/维护性建议，不阻塞合并。

yujiawei (changes requested):
# Code Review — PR #481 (octo-server)

Reviewed at head SHA `637432dc990cd840f8175decab54ef8907e31608` against `main` (merge-base `c95fdd5e`). `go build`, `go vet`, and `go test ./pkg/avatarrender/ ./pkg/metrics/` all pass locally. This was reviewed against the diff plus the surrounding call sites, with two independent secondary analysis passes whose findings are folded in below.

The architecture is sound and well-documented: a process-global LRU + singleflight + a single render-concurrency semaphore is the right shape for the CPU-starvation root cause in #480, and making the semaphore process-global (not per-endpoint) is the correct insight. The repro/fix tests are genuinely validating (singleflight collapses 50 concurrent same-key calls to exactly 1 render; the GOMAXPROCS=2 starvation test exercises the real `Cache`, not a stub). One blocking issue must be fixed before merge.

## Spec compliance

The diff implements what #480 calls for and nothing out of scope: shared cache, singleflight coalescing, render semaphore, metrics, env tunables, and a nil-safe degradation path. The ETag/304 fast path is left untouched as stated. The #478 coordination is correctly left as a follow-up (no group call sites touched here). No over-build, no missing scope.

## Blocking (P1)

### 1. 32-bit CRC32 ETag is now a process-global content-address key — collisions cross-serve another user's avatar

`modules/user/api.go:609` / `:624` pass `etag` directly as the LRU key:

```go
imageData, genErr := avatarrender.GetOrRender(etag, func() ([]byte, error) { ... })
```

and `etag` is a 32-bit CRC32 digest (`modules/user/avatar_path.go:23-30`):

```go
func avatarETag(parts ...string) string {
    h := crc32.NewIEEE()
    for _, p := range parts { h.Write([]byte(p)); h.Write([]byte{0}) }
    return fmt.Sprintf(`W/"%08x"`, h.Sum32())
}
```

This changes the security properties of `avatarETag`. **Before this PR**, the CRC32 was only an HTTP `If-None-Match` revalidation token, compared per-URL against the *same uid's* freshly recomputed ETag — a cross-uid collision was harmless because every request still rendered its own bytes. **After this PR**, the same 32-bit digest becomes the key of a *process-global* `[]byte` store shared across all users. A collision now means request for user B returns user A's already-cached rendered avatar.

Two ways this fires:
- **Accidental:** CRC32 is 32 bits; with the default 4096-entry cache the birthday probability of at least one colliding pair held simultaneously is ~0.2% — low but non-zero, and it manifests as a user silently seeing the wrong avatar (hard to diagnose, looks like a privacy bug).
- **Adversarial:** the `text` factor is user-controlled (last two visible runes of the display name, `IndividualText`), and CRC32 is linear and trivially forgeable. A user can craft a display name whose `("name-v3", uid, text)` CRC32 collides with a target's, poisoning/reading the shared cache entry.

The PR's own COMPREHENSION note ("it reuses the exact factors the ETag already enumerates ... no stale image") is the subtle gap: covering the right *factors* is necessary but not sufficient — compressing them into a 32-bit digest and then using that digest as a storage *identity* is the lossy step that introduces the collision.

**Fix (small):** decouple the cache key from the HTTP ETag. Use the full canonical render input as the LRU key and keep CRC32 only for the `ETag` header. e.g.:

```go
nameKey := "name-v3\x00" + uid + "\x00" + text   // full bytes, not hashed
imageData, genErr := avatarrender.GetOrRender(nameKey, func() ([]byte, error) { ... })
```

(or hash with a wide collision-resistant function such as SHA-256/xxhash64 if a fixed-length key is preferred). The ETag header logic is unchanged.

## Non-blocking (P2 / suggestions)

- **No request-context cancellation in `GetOrRender` / `renderWithSem`** (`pkg/avatarrender/cache.go:115`, `:140`). A flood of distinct cold keys can leave handlers blocked on `c.sem <- struct{}{}` (and singleflight) even after the client disconnects, eventually rendering abandoned responses. Note this is **strictly better than the pre-PR code**, which had no concurrency bound at all, so it is not a regression and not blocking — but threading `context.Context` and `select { case sem<-… : case <-ctx.Done(): }` (and `singleflight.DoChan`) would harden the mitigation under bot churn/timeouts. Worth a follow-up.
- **Entry-count bound, not byte-budget** (`pkg/avatarrender/cache.go:22`, `DefaultCacheSize = 4096`). Fine for today's fixed ~200×200 PNGs, but the "bytes LRU" framing implies a memory budget the code doesn't enforce. If a future caller (e.g. the #478 group path) stores larger images, memory scales with `entries × max-size`. Consider a comment pinning the per-entry size assumption, or a byte-aware cache later.
- **`OnRender` is skipped on a `render()` panic** (`pkg/avatarrender/cache.go:148-151`). The inflight/semaphore bookkeeping uses `defer` and unwinds correctly, but render duration/error metrics are lost on panic, hiding systemic render failures from Prometheus. A named-return + `defer` for `OnRender` would capture them. Minor.
- **Optional key namespacing.** Even after the P1 fix, prefixing keys per logical avatar kind (e.g. `"name:"`/`"ascii:"`) when the shared cache spans more endpoints (#478) cheaply guarantees logical isolation. The `name-v3`/`ascii-v1` version prefixes already provide this today; just keep it in mind as call sites grow.

## Verdict

The design and tests are good and the P1 fix is mechanical (swap the storage key; leave the HTTP ETag alone). Requesting changes solely for the CRC32 global-key collision risk.

Jerry-Xin (changes requested):
Flipping my prior APPROVE → REQUEST_CHANGES. yujiawei surfaced a real cache-key collision risk that I missed on first pass; I independently verified it and it is PR-owned. Credit to yujiawei.

**🔴 Blocking — shared avatar cache is keyed by a 32-bit CRC32, risking cross-user wrong-avatar serves**
- `modules/user/api.go` passes the **ETag** straight into `avatarrender.GetOrRender(etag, ...)` as the cache key. `etag = avatarETag("name-v3", uid, text)` (and `avatarETag("ascii-v1", uid)`).
- `avatarETag` (`modules/user/avatar_path.go`) is `crc32.NewIEEE()` → `fmt.Sprintf("W/\"%08x\"", h.Sum32())` — a **32-bit, non-collision-resistant CRC32** fingerprint.
- The cache is a **process-global LRU shared across all users**. Two distinct `(uid, text)` inputs whose CRC32 collides → the second request's `GetOrRender` hits the first's cached PNG → **a user is served another user's rendered avatar** (wrong-content / minor cross-user info disclosure of the other user's name+color).
- Collision realism: CRC32's 2^32 space + birthday bound makes accidental collisions non-negligible at scale, and CRC32 is trivially collision-engineerable (an attacker can craft a uid/display-name that collides with a target's key).
- Why PR-owned: `avatarETag` is pre-existing and perfectly fine **as a weak ETag** — HTTP 304 validation tolerates collisions (worst case: an unnecessary revalidation). This PR is the first to **repurpose that weak fingerprint as a cache key**, where a collision is no longer benign — it serves wrong bytes. The hazard is introduced by this PR's reuse.
- Fix: derive the cache key from a wider/strong hash (e.g. sha256, or at least fnv-128/64) over the same factors, or key directly on the exact composite string (`mode-version | uid | text`) rather than the CRC32 digest. The HTTP ETag can stay weak/CRC32; only the cache key needs to be collision-safe. Add a test asserting two distinct inputs that share a CRC32 do not share a cache entry.

**💬 Non-blocking**
- 🟡 `GetOrRender` doesn't accept a request `context.Context`, so a slow cold render can't be cancelled when the client disconnects (yujiawei). Worth threading ctx through for cancellation.

**✅ Otherwise sound (verified):** LRU is bounded (4096 entries, tunable); singleflight collapses fanout (TestCacheSingleflight: 50 concurrent same-key → 1 render); render semaphore caps cold-render concurrency; errors are not cached (TestCacheErrorNotCached); `go test -race` passes. The cache architecture is right — only the key's collision-domain is the issue. This is the user-avatar path; group avatarGet (#478) is untouched.

an9xyz:
Thanks @yujiawei and @Jerry-Xin — the P1 is real and now fixed in `2abcf5e`.

## P1 (blocking) — cache key no longer the 32-bit CRC32 ETag
Root of the hazard was exactly as described: a weak CRC32 is fine as an HTTP `If-None-Match` validator (a collision just costs a revalidation) but unsafe as a **process-global storage identity** — a collision cross-serves another user's avatar, and `text` is user-controlled + CRC32 is forgeable.

Fix: the cache is now keyed on the **full canonical factors**, not the digest:
```go
// modules/user/avatar_path.go
func avatarCacheKey(parts ...string) string { return strings.Join(parts, "\x00") }
// modules/user/api.go
nameKey  := avatarCacheKey("name-v3", uid, text)   // not avatarETag(...)
asciiKey := avatarCacheKey("ascii-v1", uid)
```
The HTTP `ETag` header logic is unchanged (still weak CRC32). Same factors, but the storage identity is now injective.

**Test (@Jerry-Xin's request)** — `TestAvatarCacheKeyResistsETagCollision`: finds a *real* CRC32 collision (fixed-seed PRNG over fixed-length text — note CRC32 is linear so sequential inputs barely collide), then asserts the two distinct inputs get **distinct cache keys** and a real `avatarrender.Cache` **does not cross-serve** B user A's bytes. Plus `TestAvatarCacheKeyDistinguishesFactors`.

## Non-blocking notes also addressed
- **OnRender on render panic** (yujiawei P2): now recorded via `defer`+`recover` (re-panics to preserve propagation), so render-failure latency/error metrics aren't lost.
- **Entry-count vs byte budget** (yujiawei/Octo-Q): documented the per-entry size assumption on `DefaultCacheSize`.
- **Starvation test flakiness** (Octo-Q P2-3): added a warmup round + skip when the runner's own baseline is already inflated, so loaded CI can't false-fail the timing assertion.

## Deferred to follow-up (non-blocking, not regressions)
- **request-context cancellation** through `GetOrRender`/semaphore (yujiawei/Jerry-Xin) — worth threading `ctx` + `singleflight.DoChan`; larger change, strictly better than pre-PR which had no bound at all.
- precise **singleflight-shared** counting (Octo-Q P2-1); entry **TTL** (Octo-Q P2-2).

`go build` / `go vet` / `go test -race ./pkg/avatarrender/ ./pkg/metrics/` / the user-package pure-fn tests all pass locally; golangci-lint 0 issues. Re-review welcome.

---

Jerry-Xin (dismissed):
## ✅ Re-review — CHANGES_REQUESTED blocker resolved (head `2abcf5e4`)

My prior round (@ `637432dc`) requested changes on the **cross-user avatar leak**: the shared process-global render cache was keyed by the 32-bit CRC32 `avatarETag`, so two distinct `(uid, text)` whose CRC32 collides would serve user B user A's cached PNG (and CRC32 is trivially collision-engineerable since the trailing nickname chars are user-controlled). **This is now fixed.** Clearing my block.

### 🔴 → ✅ Cache key decoupled from CRC32 ETag (FIXED, collision-safe)
- New `modules/user/avatar_path.go:41` — `avatarCacheKey(parts ...string) string { return strings.Join(parts, "\x00") }`. The cache key is now the **full raw composite string** (mode-version | uid | text) joined with a NUL separator — an **injective** mapping, no hashing/truncation, so collision is structurally impossible (not merely improbable).
- Call sites `modules/user/api.go:612` (`nameKey := avatarCacheKey("name-v3", uid, text)`) and `:627` (`asciiKey := avatarCacheKey("ascii-v1", uid)`) now pass these keys into `GetOrRender`, replacing the prior `etag`/`asciiETag` CRC32 values.
- The HTTP ETag header correctly **stays** CRC32 (`avatarETag`, `avatar_path.go:24`) — weak ETag collision is benign for a 304, only the shared cache identity needed strengthening. They changed the right one and left the right one alone.

### ✅ Regression test added (PASS)
- `TestAvatarCacheKeyResistsETagCollision` (`avatar_etag_test.go`): actually *finds* two 8-char texts whose CRC32 ETag collides (fixed seed, deterministic), asserts their `avatarCacheKey` does **not** collide, and end-to-end asserts a real `avatarrender.Cache` does **not** cross-serve B user A's bytes. PASS under `-race`.
- `TestAvatarCacheKeyDistinguishesFactors` also PASS (mode/uid/text each change the key).

### ✅ No regression in previously-sound properties (`go test -race` green)
- LRU bound 4096 (`DefaultCacheSize`), singleflight collapse (`c.group.Do`), render semaphore, **errors-not-cached** (early return before `lru.Add`) — all intact.
- Bonus hardening this round: `renderWithSem` now records render metrics via `defer` + `recover()`/re-panic, so a render panic still emits an `OnRender` data point without swallowing the panic. Sound.
- `go build ./...` exit 0; `go vet` (avatarrender/metrics/user/.) exit 0; `go test -race ./pkg/avatarrender/ ./pkg/metrics/` ok; `go test -race ./modules/user/` (etag/collision/INM/legacy) ok.

### 💬 Non-blocking notes
- 🟡 `GetOrRender` still takes no `context.Context`, so a cold render can't be cancelled when the client disconnects. Unchanged from prior round — still non-blocking (a started render must complete to populate the shared entry; bounded by the semaphore). Fine to defer.
- 👍 The added doc comment on `DefaultCacheSize` correctly flags that the LRU bounds by **entry count, not bytes**, and that #478's variable-size group avatars may warrant a byte-budget cache later. Good forward-looking note.

Cache-key leak is closed and pinned by a real collision regression test. LGTM.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#481 Review Report (automated review / automated review)
**Head SHA:** `2abcf5e45e2a4eaa1fb11d29c3434a1bf554d173`
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/481
**Title:** perf: shared avatar render cache to stop CPU starvation under fanout (#480)
**Branch:** `perf/avatar-render-cache` → `main`
**Changed:** 10 files, +978 / −4

---

## 1. 验证结论

| 验证项 | 结论 | 证据 |
|--------|------|------|
| Cache key 抗碰撞（injective） | ✅ | `avatar_path.go:41` — `avatarCacheKey` 用 NUL 分隔完整原始 parts，不做哈希截断 |
| Cache key 与 ETag 解耦 | ✅ | `api.go:612` nameKey / `:627` asciiKey 用 `avatarCacheKey`；ETag header 仍用 `avatarETag` (CRC32) |
| LRU hit/miss 正确性 | ✅ | `cache.go:115-122` — Get 命中返回 bytes + OnHit；未命中走 singleflight |
| Singleflight 合并 | ✅ | `cache.go:124-140` — `group.Do(key, fn)` 包裹渲染，内部 double-check LRU |
| 渲染信号量限并发 | ✅ | `cache.go:143-154` — buffered channel + defer release；`default.go:48` 默认 max(1, GOMAXPROCS-1) |
| Nil 接收者退化 | ✅ | `cache.go:113-115` — nil c 直接调 render()，不 panic |
| 错误不缓存 | ✅ | `cache.go:133-135` — rerr != nil 时跳过 lru.Add |
| 进程级唯一信号量 | ✅ | `default.go:19` — `atomic.Pointer[Cache]`；`main.go:222-232` 构造一次并 SetDefaultCache |
| 启动退化安全 | ✅ | `main.go:229-232` — NewCache 失败仅 log.Warn，GetOrRender 退化为直接渲染 |
| Metrics no-op 安全 | ✅ | `avatar.go` — 所有 Observe 函数先 Load atomic pointer，nil 时 skip |
| 依赖可用 | ✅ | `go.mod:28` — `hashicorp/golang-lru/v2 v2.0.7`；`golang.org/x/sync` 已存在 |
| 测试覆盖 | ✅ | cache_test.go (5 tests) + starvation_repro_test.go (3 tests) + avatar_test.go (2 tests) + avatar_etag_test.go (collision + factors) |
| 生产代码无绕过缓存的 Render 调用 | ✅ | grep 确认 `avatarrender.Render` 仅在 GetOrRender 闭包内调用 (api.go:614) |
| Group avatar 不受影响 | ✅ | `modules/group/api.go` — avatarGet 仅服务静态文件或 302 重定向，无按需渲染 |

---

## 2. 发现的问题

**无 P0/P1。** 以下为 P2 级别可观测性/维护性建议。

### P2-1: Singleflight OnShared 指标低估

**Diff-scope 三问：** new (本 PR 新引入) / 不影响功能正确性

**位置：** `cache.go:124-138`

`group.Do(key, fn)` 返回 `(v, err, shared)` 中 `shared=true` 仅对本次 flight 的**等待者**成立。Flight leader（发起渲染的 goroutine）拿到 `shared=false`，即使其渲染结果被 N 个 waiter 复用。此外 late-comer 在 double-check 路径 (`cache.go:126-128`) 命中 LRU 直接返回，完全绕过 singleflight，也不触发 OnShared。

**影响：** 可观测性 dashboard 低估实际被合并的请求数。不影响缓存功能或 CPU 保护效果。

**建议：** 可在闭包内设 atomic flag，flight leader 未调 render（double-check 命中）也算 shared。Nice-to-have，不阻塞。

### P2-2: GetOrRender 不传 request context

**Diff-scope 三问：** new (本 PR 新增代码) / 不影响正确性

**位置：** `cache.go:115` / `:143`

GetOrRender 不接受 `context.Context`。客户端断开后，信号量排队中的 goroutine 无法取消，会继续渲染并写入缓存后才退出。不过 semaphore 限制了排队深度，且渲染结果会被后续请求复用（非浪费），影响有限。

**建议：** 后续版本可考虑 threading ctx through：`select { case sem <- ...: case <-ctx.Done(): }` + `singleflight.DoChan`。当前不阻塞。

### P2-3: Starvation repro test 的 CI 稳定性

**Diff-scope 三问：** new (新增测试)

**位置：** `starvation_repro_test.go:114-156`

`TestRenderFanoutStarvesVictim` 断言 `ratio >= 2.0`，依赖运行时调度行为。已有 `testing.Short()` 跳过和 baseline overload 检测 (`base > 3*ideal` → skip)。在 overloaded CI runner 上可能偶发 fail。

**建议：** 当前阈值 + skip 逻辑已足够保守。如 CI 偶发 fail，可考虑放宽到 1.5x。

---

## 3. 建议

- **P2-1** 可作为 follow-up：改进 OnShared 计数使 dashboard 更准确
- **P2-2** 后续版本 threading context through GetOrRender，增强客户端断开时的响应性
- #478 合并后确保 group avatar 的 RenderGroup/RenderIcon 也走 `GetOrRender`（PR body coordination plan 已明确 ✅）

---

## 4. 额外发现

无 — 未发现安全问题、跨路径不一致、或数据流断裂。

---

## 5. 数据流回溯

| 消费点 | 数据 | 上游来源 | 运行时数据是否真的流到消费点 |
|--------|------|----------|------------------------------|
| `GetOrRender(nameKey, ...)` @ `api.go:613` | nameKey = `avatarCacheKey("name-v3", uid, text)` | uid 来自路由参数 `c.Param("uid")`；text = `IndividualText(userInfo.Name)` @ `api.go:589` | ✅ nameMode=true 才进此分支，text 非空且 Renderable |
| `GetOrRender(asciiKey, ...)` @ `api.go:628` | asciiKey = `avatarCacheKey("ascii-v1", uid)` | uid 来自路由参数 | ✅ nameMode=false 或 nameMode 渲染失败后的 fallback |
| `Render(Options{Text: text, Bg: ColorForSeed(uid)})` @ `api.go:614-617` | text, Bg | text = IndividualText(userInfo.Name)；Bg = ColorForSeed(uid) | ✅ 均为确定性函数，值由路由参数 + DB 查询决定 |
| `generateDefaultAvatar(uid)` @ `api.go:629` | uid | 路由参数 | ✅ ASCII fallback：固定色板按 uid 取色 + basicfont 首字母 |
| `c.lru.Get(key)` @ `cache.go:116` | key | 调用方传入的 avatarCacheKey 结果 | ✅ string 类型，LRU 内部 mutex 保护并发 |
| `renderWithSem(render)` @ `cache.go:143` | render closure | 调用方传入 | ✅ closure 捕获 text/uid/bg，均为值拷贝，无共享可变状态 |
| `c.hooks.OnHit/OnMiss/...` @ `cache.go:117-138` | hooks | NewCache 时 `cfg.Hooks.normalize()` 填充 nil → no-op @ `cache.go:57-72` | ✅ normalize 在构造时一次性完成，运行时不可能 nil panic |
| `metrics.ObserveAvatarNotModified()` @ `api.go:598` | 无参数 | 304 路径直接调用 | ✅ atomic pointer Load，nil 时 skip |

---

## 6. 盲点 checklist（R5）

- **C1 双路径 parity（含守卫/lint 工具自身等价写法覆盖）：** Clear。nameMode 和 ASCII fallback 两条渲染路径均已一致包裹 GetOrRender。本 PR 不涉及 add/remove、subscribe/unsubscribe 等对称路径。Group avatar 无按需渲染，无需 parity。
- **C2 control-flow ordering / 嵌套复用 + 安全控件非 canonical 试穿：** Clear。GetOrRender 仅在 api.go 两个调用点使用，ordering 一致（304 check → GetOrRender）。singleflight 内部 double-check 正确。render closure 是确定性的（相同 key → 相同 bytes）。
- **C3 授权边界 ≠ 能力边界：** N/A。无权限/jail/tool 暴露变更。
- **C4 授权生命周期 / 容器-成员状态级联：** N/A。无鉴权变更。
- **C5 build/note 通过 ≠ 运行期路径正确：** N/A。纯 Go 服务端代码，无构建产物/扩展/CLI/相对路径变更。build 通过即运行期路径一致。
- **C6 治理/策略/安全文档自洽性：** N/A。无文档变更。

---

## 7. 跨轮 blocker 复检（R6）

上轮唯一 P1 blocker（yujiawei + Jerry-Xin，SHA `637432dc`）：**CRC32 ETag 作为进程级共享缓存 key 导致碰撞串图**。

**已修** ✅ — 当前 head `2abcf5e4` 的修复：
- `avatar_path.go:41` 新增 `avatarCacheKey` 函数，用 NUL 分隔完整原始 parts（injective mapping），替代了原先直接传 `avatarETag` (CRC32) 作为 cache key
- `api.go:612` / `:627` 两个调用点均已改用 `avatarCacheKey`
- HTTP ETag header 保持 CRC32 不变（`avatar_path.go:24`）— 304 collision 无害
- 回归测试 `TestAvatarCacheKeyResistsETagCollision` 实际找到 CRC32 碰撞对，断言 avatarCacheKey 不碰撞 + 真实 Cache 不串图 ✅
- `TestAvatarCacheKeyDistinguishesFactors` 验证 mode/uid/text 各自改变 key ✅

上轮 P2 发现（Jerry-Xin re-review `2abcf5e4`）：
- `renderWithSem` panic 时指标丢失 → **已修**：当前 `cache.go:148-167` 使用 named return + defer recover/re-panic，panic 时 OnRender 正确记录 panic error

---

**[Octo-Q] verdict: APPROVE**

本 PR 架构清晰（LRU + singleflight + semaphore 三层防御），上轮 P1 blocker (CRC32 cache key 碰撞) 已在当前 head 完全修复并 pinned by regression test。数据流完整回溯无断裂，nil-receiver 退化安全，测试覆盖充分（含 starvation 复现 + 修复验证 + 碰撞回归测试）。三个 P2 发现均为可观测性/维护性建议，不阻塞合并。

yujiawei (dismissed):
# Code Review — PR #481 (octo-server)

Re-review at head `2abcf5e45e2a4eaa1fb11d29c3434a1bf554d173` against `main` (merge-base `c95fdd5e`). Round 1 requested changes for one blocker: the process-global shared render cache was keyed by the 32-bit CRC32 ETag, so a collision (and `text` is user-controlled + CRC32 is forgeable) could cross-serve another user's avatar. This round verifies that fix and re-audits the surrounding code. `go build ./...`, `go vet`, `go test ./pkg/avatarrender/ ./pkg/metrics/`, the `modules/user` pure-fn avatar tests, and `go test -race ./pkg/avatarrender/ ./pkg/metrics/` all pass locally.

## 1. Spec compliance

**Spec: ✅**

- Missing: none. The blocker is addressed and the previously-noted P2s are handled (OnRender-on-panic, entry-count-budget doc).
- Over-build: none. The fix is the minimal, correct one — decouple the storage key from the HTTP ETag. The #478 group-avatar coordination is correctly left as a follow-up (no group call sites touched here).
- Deviation: none. ETag header logic is unchanged (still weak CRC32); only the cache key changed.

### Blocker resolution (verified)

`modules/user/avatar_path.go:41` introduces `avatarCacheKey(parts...) = strings.Join(parts, "\x00")`, and `modules/user/api.go:611,624` now key the shared cache on it:

```go
nameKey  := avatarCacheKey("name-v3", uid, text)
asciiKey := avatarCacheKey("ascii-v1", uid)
```

This is the right fix and it is complete:
- The cache key is now the full canonical factors (NUL-joined), not a lossy 32-bit digest, so two distinct `(mode, uid, text)` inputs that share a CRC32 ETag get distinct cache entries.
- Factor parity with the ETag is exact: name mode keys on `("name-v3", uid, text)`, ascii on `("ascii-v1", uid)` — the same factors `avatarETag` enumerates at `api.go:591,593`. A rename/recolor still produces a new key (no stale image), which the COMPREHENSION note correctly claims.
- The ETag header stays CRC32, so HTTP 304 revalidation is unchanged.
- `TestAvatarCacheKeyResistsETagCollision` actually constructs a real CRC32 collision and asserts both distinct cache keys *and* end-to-end no cross-serve through a real `avatarrender.Cache`; `TestAvatarCacheKeyDistinguishesFactors` pins per-factor distinctness. These are genuine regression locks, not stubs.

## 2. Code Quality verdict

**Quality: Approved**

The architecture is sound: a process-global LRU + singleflight + a single render-concurrency semaphore is the correct shape for the #480 CPU-starvation-under-fanout root cause, and keeping the semaphore process-global (not per-endpoint) is the right insight. Concurrency is correct: the singleflight closure double-checks the LRU after acquiring the flight (`cache.go:122`), errors are not cached, and the inflight/semaphore bookkeeping unwinds via `defer`. The OnRender-on-panic fix (`cache.go:148-151`) records render latency/error then re-panics to preserve propagation — correct.

No P0/P1. The findings below are all P2/non-blocking (one independent advisory leg, plus my own trace):

- **P2 — `strings.Join(parts, "\x00")` is not injective for *arbitrary* factors** (`modules/user/avatar_path.go:41`). `["name-v3","u","a\x00b"]` and `["name-v3","u\x00a","b"]` collide. This does **not** reopen the blocker for today's factors: `uid` is an HTTP path param (cannot carry a literal NUL) and `text` comes from `IndividualText`, which strips control characters (`unicode.Cc`, which includes U+0000) at `pkg/avatarrender/text.go`. So the key is injective in practice. Worth hardening (length-framed parts, or a wide hash like sha256/xxhash64 over them) before any future factor that could contain NUL is added — especially when #478 routes group keys through the same cache.
- **P2 — semaphore-release `defer` is registered after the `OnSemaphoreWait` hook** (`pkg/avatarrender/cache.go:140-143`). If that hook ever panicked, the acquired token would leak and the semaphore could eventually deadlock. Today the only hook is a Prometheus histogram `Observe` (won't panic), so this is defensive only — move `defer func(){ <-c.sem }()` to immediately after the channel send to be panic-safe.
- **P2 — `OnShared` overcounts by one per shared flight** (`pkg/avatarrender/cache.go:131-133`). `singleflight.Group.Do` reports `shared=true` to the originating caller too, so the "coalesced onto an in-flight render" counter is off by one per flight. Metrics-only.
- **P2 — entry-count bound, not byte budget** (`pkg/avatarrender/cache.go:27`, `DefaultCacheSize = 4096`). Now documented; fine for today's fixed ~200×200 PNGs. Revisit if a future caller stores larger images.
- **P2 (follow-up) — no request-context cancellation** through `GetOrRender`/the semaphore. A flood of distinct cold keys can leave handlers parked on `c.sem <- struct{}{}` after the client disconnects. This is strictly better than pre-PR (which had no bound at all), so it is not a regression — worth threading `context.Context` + `singleflight.DoChan` in a follow-up.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. The round-1 blocker (CRC32 cache-key collision → cross-user avatar serve) is correctly and completely fixed, factor parity is exact, and the fix is locked by a real collision-regression test. Remaining items are P2 hardening/metrics polish that do not block merge.

## 4. Suggestions

- Before #478 routes group avatar keys through this shared cache, switch `avatarCacheKey` to length-framed parts or a wide hash so injectivity no longer depends on every caller's factors being NUL-free.
- Move the semaphore-release `defer` ahead of the `OnSemaphoreWait` hook call.
- Adjust the `OnShared` count to exclude the originating caller if exact singleflight-shared metrics matter.

## 5. Additional notes

- Coverage: verified the diff, the `UserAvatar` handler context, `IndividualText`/`isInvisible`, the cache/metrics implementations and tests, and main.go wiring (single shared cache, hooks injected at the composition root so `pkg/avatarrender` stays free of `pkg/metrics`). An independent advisory leg (GPT-family) reviewed the same diff and surfaced the three P2s above and no P0/P1. A second cross-family advisory leg (Gemini) failed to return findings on two attempts (gateway returned empty) — recorded as **absent**, not as agreement; its absence did not affect this verdict, which rests on the direct code/test inspection and the GPT-family leg.
- Not run: `modules/user` / `modules/group` integration tests (need MySQL/Redis/WuKongIM, unavailable here) — consistent with the PR's stated test scope; the affected logic is covered by the pure-function tests that do run.

an9xyz:
Folded in the three cheap hardening notes from the re-reviews in `a9663df` (thanks @yujiawei / @mochashanyao):

- **Key injectivity** (yujiawei): `avatarCacheKey` now length-frames parts (`<len>:<bytes>`) instead of NUL-joining, so it's injective for *arbitrary* bytes — not just today's NUL-free factors. Removes the latent footgun before #478 routes group keys through the same builder. Pinned by `TestAvatarCacheKeyInjectiveAcrossPartBoundaries`.
- **Semaphore release ordering** (yujiawei): the `defer func(){ <-c.sem }()` is now registered immediately after the token acquire, before the `OnSemaphoreWait` hook, so a hook panic can't leak a token.
- **`OnShared` off-by-one** (yujiawei + Octo-Q): `singleflight.Do` reports `shared=true` to the originating caller too, so the coalesced counter was +1 per flight; now only true waiters (`shared && !ranRender`) are counted.

Left as follow-up (non-blocking, larger): request-context cancellation through `GetOrRender`/the semaphore + `singleflight.DoChan`; entry TTL.

`go build` / `go vet` / `go test -race ./pkg/avatarrender/ ./pkg/metrics/` / `modules/user` pure-fn avatar tests all pass; golangci-lint 0 issues.

---

Jerry-Xin (approved):
## 🔄 Re-review @ `a9663df` — APPROVE

Re-review of the new commit `a9663df` (prior round was APPROVE @ `2abcf5e4`). This push is a focused hardening commit that cleanly resolves **all three** residual 🟡s from the last round, with no regression to the core fix. Verdict: **APPROVE**.

### ✅ Core fix preserved (no regression)
The cross-user cache-key collision fix is intact. Call sites in `modules/user/api.go` still pass `avatarCacheKey(...)` into `GetOrRender`, and the HTTP `ETag` header still uses CRC32 `avatarETag(...)` (correct split). The key was **not** reverted to CRC32. `TestAvatarCacheKeyResistsETagCollision` still passes.

### ✅ 🟡(a) NUL-injectivity — hardened (was a real theoretical gap)
`avatarCacheKey` switched from `strings.Join(parts, "\x00")` to **length-framing** (`<len>:<bytes>` per part):
```go
func avatarCacheKey(parts ...string) string {
    var b strings.Builder
    for _, p := range parts {
        b.WriteString(strconv.Itoa(len(p)))
        b.WriteByte(':')
        b.WriteString(p)
    }
    return b.String()
}
```
This is injective for **arbitrary** byte content, eliminating the `["u","a\x00b"]` vs `["u\x00a","b"]` aliasing the NUL-join had. New regression test `TestAvatarCacheKeyInjectiveAcrossPartBoundaries` pins it (incl. empty-vs-absent trailing part). Correct — bulletproof for future factors (e.g. #478 group avatars).

### ✅ 🟡(b) Semaphore leak on hook panic — fixed
In `renderWithSem`, the release `defer func() { <-c.sem }()` is now registered **before** the `OnSemaphoreWait` hook call. A panicking metrics hook can no longer leak an acquired token → no slow semaphore starvation/deadlock. Correct ordering.

### ✅ 🟡(c) OnShared undercount — fixed
`GetOrRender` now tracks a goroutine-local `ranRender` flag and gates the metric as `if shared && !ranRender`, so the singleflight leader (for whom `shared` is also true) is excluded; only genuine waiters count. Local flag, no cross-goroutine access → race-safe.

### Verification (receipts)
- `go build ./...` → exit 0
- `go vet ./pkg/avatarrender/... ./modules/user/...` → exit 0
- `go test -race ./pkg/avatarrender/...` → **ok** (26.5s)
- `go test -race ./pkg/metrics/...` → **ok**
- `go test -race ./modules/user/` PR-owned tests (CacheKey/ETag/IfNoneMatch/FilePath) → **all PASS**
- LRU bound / singleflight / errors-not-cached behavior unchanged in this diff; covered by passing avatarrender suite.

⚠️ Pre-existing env failure (unrelated to this PR): `TestUserAvatarVersionDBRoundTrip` panics with `unknown migration in database` inside `octo-lib/testutil.NewTestServer`. File `avatar_version_test.go` is **not** in this PR's changed set and the panic is in shared test infra — classified separately, not a PR-owned defect.

Clean, well-commented, defensive. LGTM. ✅

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#481 Review Report (re-review @ head `a9663dfa`)
**Head SHA:** `a9663dfaa1426b91a6905ec22954c1f6f846ddfa`
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/481
**Title:** perf: shared avatar render cache to stop CPU starvation under fanout (#480)
**Branch:** `perf/avatar-render-cache` → `main`
**Round:** 4 (re-review after hardening commit `a9663dfa`)

---

## 1. Verification Summary

| Check | Result | Evidence |
|-------|--------|----------|
| `go build ./...` | ✅ exit 0 | local run |
| `go vet` (avatarrender/metrics/user/.) | ✅ exit 0 | local run |
| `go test -race ./pkg/avatarrender/` | ✅ all pass (34s) | 18 tests including starvation repro |
| `go test -race ./pkg/metrics/` | ✅ all pass | 20 tests |
| `go test -race ./modules/user/ -run Avatar\|IfNoneMatch` | ✅ all pass | 5 tests including new injectivity test |
| Cache key injective (length-framed) | ✅ | `avatar_path.go:47-53` — `<len>:<bytes>` per part |
| Cache key ≠ CRC32 ETag | ✅ | `api.go:612` / `:627` use `avatarCacheKey`, ETag header uses `avatarETag` |
| LRU hit/miss correct | ✅ | `cache.go:115-122` |
| Singleflight collapse | ✅ | `cache.go:124-140` with double-check LRU |
| Render semaphore bounded | ✅ | `cache.go:153-168` — defer release before hook call |
| OnShared excludes leader | ✅ | `cache.go:122,139` — `ranRender` flag + `shared && !ranRender` |
| Nil-receiver degradation | ✅ | `cache.go:113-115` |
| Errors not cached | ✅ | `cache.go:133-135` |
| Process-global single cache+semaphore | ✅ | `default.go:19` atomic.Pointer; `main.go:222-232` |
| Startup degradation safe | ✅ | `main.go:229-232` — NewCache failure → log.Warn only |
| Metrics no-op safe | ✅ | `avatar.go` — atomic pointer Load, nil → skip |
| No bypass Render calls | ✅ | grep: `avatarrender.Render` only inside GetOrRender closure (`api.go:614`) |
| Group avatar unaffected | ✅ | `modules/group/api.go` — no on-demand rendering |

---

## 2. Findings

**No P0/P1.** All prior blockers fully resolved. Two minor P2 observations follow.

### P2-1: OnShared fires for leader in double-check LRU-hit path

**Diff-scope:** new (introduced by this PR's `ranRender` logic)

**Location:** `cache.go:122-139`

When a flight leader enters the singleflight closure and the double-check LRU hits (`cache.go:126-128`), `ranRender` stays `false` (no render executed) and `shared` is `true` (waiters exist). The condition `shared && !ranRender` fires `OnShared` for the leader — but the leader didn't share another's render; it found the value from a prior flight's cache entry. Waiters correctly get `OnShared` too.

**Impact:** Metrics-only. OnShared slightly overcounts in this edge case (leader counted as "shared" when it actually got a cache hit). Does not affect cache correctness, CPU protection, or user-visible behavior.

**Suggestion:** Minor — could track a separate `cacheHitInFlight` boolean from the double-check and exclude it: `if shared && !ranRender && !cacheHitInFlight`. Nice-to-have.

### P2-2: GetOrRender still lacks request `context.Context`

**Diff-scope:** pre-existing (unchanged from prior rounds)

**Location:** `cache.go:115`

Unchanged from prior reviews. A flood of distinct cold keys can leave handlers parked on `c.sem` after client disconnect. Bounded by the semaphore (strictly better than pre-PR), so non-blocking. Worth threading `context.Context` + `singleflight.DoChan` in a follow-up.

---

## 3. Suggestions

- **P2-1:** Optional OnShared accuracy refinement (non-blocking)
- **P2-2:** Thread `context.Context` through `GetOrRender` in a follow-up (non-blocking)
- Ensure #478 routes group avatar `RenderGroup`/`RenderIcon` through `GetOrRender` with `avatarCacheKey` (PR body coordination plan already states this ✅)

---

## 4. Additional Findings

None — no security issues, no cross-path inconsistency, no data-flow breakage.

---

## 5. Data-Flow Trace

| Consumer | Data | Upstream Source | Runtime Flow Verified |
|----------|------|-----------------|----------------------|
| `GetOrRender(nameKey, ...)` @ `api.go:613` | `nameKey = avatarCacheKey("name-v3", uid, text)` | uid from `c.Param("uid")`; text from `IndividualText(userInfo.Name)` @ `api.go:589` | ✅ nameMode=true gate; text non-empty & Renderable |
| `GetOrRender(asciiKey, ...)` @ `api.go:628` | `asciiKey = avatarCacheKey("ascii-v1", uid)` | uid from route param | ✅ nameMode=false or nameMode render failure fallback |
| `Render(Options{Text, Bg})` @ `api.go:614-617` | text, Bg | text = `IndividualText(userInfo.Name)`; Bg = `ColorForSeed(uid)` | ✅ Deterministic functions of route param + DB query |
| `generateDefaultAvatar(uid)` @ `api.go:629` | uid | route param | ✅ ASCII fallback: fixed palette by uid + basicfont initial |
| `c.lru.Get(key)` @ `cache.go:116` | key | caller's `avatarCacheKey` result | ✅ string type, LRU mutex-protected |
| `renderWithSem(render)` @ `cache.go:153` | render closure | caller-provided | ✅ closure captures text/uid/bg by value, no shared mutable state |
| `c.hooks.OnHit/OnMiss/...` @ `cache.go:117-139` | hooks | `NewCache` → `cfg.Hooks.normalize()` fills nil → no-op @ `cache.go:57-72` | ✅ normalize runs once at construction |
| `metrics.ObserveAvatarNotModified()` @ `api.go:598` | no params | 304 path direct call | ✅ atomic pointer Load, nil → skip |
| `ranRender` flag @ `cache.go:122` | bool | local to each goroutine's closure call | ✅ singleflight runs only leader's closure; waiter's `ranRender` stays false |

---

- **C1 Dual-path parity (incl. guard/lint tool equivalent-form coverage):** Clear. nameMode and ASCII fallback render paths are both consistently wrapped in `GetOrRender` with `avatarCacheKey`. No add/remove or subscribe/unsubscribe symmetric paths touched. Group avatar has no on-demand rendering.
- **C2 Control-flow ordering / nested reuse + security-control non-canonical probe:** Clear. `GetOrRender` has exactly 2 call sites in `api.go`, consistent ordering (304 check → GetOrRender). Singleflight double-check correct. Render closure is deterministic (same key → same bytes). The length-framed key encoding handles arbitrary byte content correctly (verified: `["u","a\x00b"]` ≠ `["u\x00a","b"]`; `["x",""]` ≠ `["x"]`).
- **C3 Authorization boundary ≠ capability boundary:** N/A — no permission/jail/tool exposure changes.
- **C4 Authorization lifecycle / container-member state cascade:** N/A — no auth changes.
- **C5 Build/note pass ≠ runtime path correct:** N/A — pure Go server code, no build artifacts/extensions/CLI/relative URLs.
- **C6 Governance/policy/security document self-consistency:** N/A — no document changes.

---

## 7. Cross-Round Blocker Re-Check (R6)

### Round 1 P1 blocker (yujiawei + Jerry-Xin, SHA `637432dc`): CRC32 ETag as cache key → cross-user avatar serve

**Status: FIXED ✅ (then hardened)**

- **@ 2abcf5e4:** `avatarCacheKey` with NUL-separated parts replaced CRC32 ETag as cache key; regression test `TestAvatarCacheKeyResistsETagCollision` added
- **@ a9663dfa:** Hardened to length-framed encoding (`<len>:<bytes>`) for structural injectivity even when parts contain NUL; new test `TestAvatarCacheKeyInjectiveAcrossPartBoundaries` pins the edge case
- HTTP ETag header correctly stays CRC32 (`avatar_path.go:24`) — weak ETag collision benign for 304

### Round 2 P2 items (yujiawei re-review @ `2abcf5e4`)

| P2 Item | Status |
|---------|--------|
| `strings.Join(parts, "\x00")` not injective for arbitrary bytes | **Fixed** — length-framed encoding @ `avatar_path.go:47-53` |
| Semaphore-release `defer` after `OnSemaphoreWait` hook | **Fixed** — `defer` now before hook @ `cache.go:156-158` |
| `OnShared` overcounts by one per flight | **Fixed** — `ranRender` flag @ `cache.go:122,139` |
| Entry-count budget, not byte budget | **Documented** — `DefaultCacheSize` comment @ `cache.go:27` |
| No request-context cancellation | **Unchanged** — non-blocking follow-up |

### Round 3 P2 items (mochashanyao Octo-Q @ `2abcf5e4`)

| P2 Item | Status |
|---------|--------|
| Singleflight OnShared metric undercounting | **Fixed** (same as above) |
| GetOrRender no context.Context | **Unchanged** — non-blocking follow-up |
| Starvation repro test CI stability | **Unchanged** — thresholds + skip logic already conservative |

---

**[Octo-Q] verdict: APPROVE**

All three rounds of blockers and P2 hardening items are fully resolved at head `a9663dfa`. The length-framed cache key encoding is structurally injective for arbitrary byte content, closing the last theoretical gap from the NUL-separator approach. The `ranRender` flag correctly fixes OnShared overcounting, and the semaphore defer ordering is now panic-safe. Architecture remains sound (process-global LRU + singleflight + semaphore), tests pass under `-race`, and `go build`/`go vet` are clean. Two remaining P2 items are metrics-only and a deferred context.Context follow-up — neither blocks merge.

yujiawei (approved):
# Code Review — PR #481 (octo-server)

Reviewed at head `a9663dfaa1426b91a6905ec22954c1f6f846ddfa` against base `main` (merge-base `c95fdd5e`). This head is a polish commit on top of the previously-reviewed cache; it resolves the advisory notes from the prior round (cache-key injectivity, semaphore-release ordering, singleflight accounting, render-panic metrics).

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly what #480 calls for and nothing out of scope:
- Process-global shared render cache: LRU + singleflight + render-concurrency semaphore (`pkg/avatarrender/cache.go`).
- Single process-wide instance so the semaphore is a true global render ceiling, not 2N (`pkg/avatarrender/default.go`, `main.go`).
- Both user-avatar render paths (nameMode + ASCII fallback) routed through `GetOrRender` (`modules/user/api.go:611-636`); the ETag/304 fast path is untouched.
- Cache key derived from full content factors via `avatarCacheKey`, deliberately **not** the 32-bit CRC32 ETag (`modules/user/avatar_path.go:47`).
- Metrics are observe-only hooks injected at the composition root, keeping `pkg/avatarrender` free of `pkg/metrics` (`pkg/metrics/avatar.go`, `main.go`).

No over-build (no unrequested endpoints/flags) and no deviation. The #478 group-avatar call sites are correctly left as future work: `modules/group/api.go:354 avatarGet` only redirects to a download URL today and does not render on demand, so there is no current call site to protect. The "merge this first, #478 adapts" coordination note is accurate.

## 2. Code Quality

**Quality: Approved**

Verified independently against this head:

- **Cache key injectivity (no cross-user image bleed).** `avatarCacheKey` uses length-framing (`<len>:<bytes>`), injective for arbitrary bytes including embedded NUL. Keeping CRC32 only for the HTTP ETag header is correct: a weak validator collision costs at most a revalidation, but a collision in a process-global byte store would cross-serve another user's avatar. `TestAvatarCacheKeyResistsETagCollision` pins this with a real constructed CRC32 collision; `TestAvatarCacheKeyInjectiveAcrossPartBoundaries` pins the NUL/length-framing case.
- **Singleflight accounting.** `OnShared` fires only on `shared && !ranRender`, correctly excluding the leader (singleflight reports `shared=true` to the originator too). Double-check inside the flight avoids a redundant render after a concurrent fill. Errors are not cached (return precedes `lru.Add`), and `TestCacheErrorNotCached` covers retry-after-error.
- **Semaphore release under panic.** `defer func(){ <-c.sem }()` is registered immediately after token acquisition and before the `OnSemaphoreWait` hook call, so a hook panic cannot leak a token. The render path wraps `render()` in a recover that records render-duration metrics on panic and re-panics, preserving propagation semantics. nil-receiver path degrades to direct render.
- **Build/vet/test.** `go build` / `go vet` clean on the affected packages; `go test -count=1 ./pkg/avatarrender/ ./pkg/metrics/` passes, including the deterministic GOMAXPROCS=2 starvation repro (`TestRenderFanoutStarvesVictim` ~16x amplification; `TestCacheEliminatesStarvation` returns to ~1x through the real cache).

### Non-blocking notes (P2)

- A leader render *panic* propagates through singleflight to all concurrent waiters of that key (one crash → N goroutine panics). Net user-visible outcome is unchanged because render is deterministic (all those requests would 500 regardless) and the HTTP framework recovers per-request; noting for awareness only.
- `DefaultCacheSize` bounds the LRU by entry count, not bytes. Fine for today's fixed ~200×200 PNGs; when #478 routes group avatars (potentially larger/variable images) through the same cache, revisit toward a byte-budgeted bound. The code comment already calls this out.

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved → APPROVE. The change is well-scoped, the prior round's hardening notes are fully addressed, and the fix is validated by deterministic tests that exercise the real cache rather than a stub.

## 4. Coverage / blind spots

- Independent legs run this round: one cross-family advisory leg **timed out and is recorded as absent** (it did not contribute findings and was not treated as approval). One adversarial leg completed and raised a single [P1] that I could **not reproduce and have refuted**: it claimed `counterValue` in `pkg/metrics/avatar_test.go` redeclares the same helper in `pkg/metrics/http_test.go`. They live in different test packages (`package metrics` vs `package metrics_test`), which Go compiles separately — `go test -count=1 ./pkg/metrics/` builds and passes, and the avatar metrics test actually runs. No conflict.
- Not run here (require MySQL/Redis/WuKongIM): `modules/user` / `modules/group` integration tests. The avatar render path under review has no DB/network dependency, so unit-level coverage is representative for the changed behavior.

PR #490, opened 2026-06-27
fix: DM conversations always visible in all Spaces (#484)

### LINKED ISSUE #484: DMs leak across Spaces and mutually hide between Spaces (single physical channel + soft space_id tag)
## Summary

Direct messages (Person channels) are physically a single channel per user-pair with no Space dimension. Multi-Space isolation is emulated entirely through a soft `payload.space_id` tag plus filtering. This surfaces two user-visible defects that are two sides of the same mechanism.

## Symptoms

1. **Cross-Space history pollution** — opening the same contact's DM under two different Spaces shows an overlapping message history; any message without a `space_id` tag appears in every Space.
2. **Mutually-exclusive visibility** — after chatting with a contact in Space A, that DM conversation disappears from Space B. Whichever Space was active most recently "wins"; the other hides the conversation.

## Root cause (architecture, not a single-line bug)

- DM is a single physical channel; conversation-level `SpaceID` is intentionally left empty — `modules/message/api_conversation.go:1584`.
- Isolation relies solely on per-message `payload.space_id`, injected server-side from the sender's `X-Space-ID` — `modules/message/api.go:689`. When the sender omits `X-Space-ID`, the server *strips* `space_id`, leaving the message untagged.
- History filter keeps untagged non-systembot messages in **every** Space (rule 2 of `filterPersonMessagesBySpace`, `modules/message/space_filter.go:442`) → **symptom 1**. Untagged messages come from: senders without `X-Space-ID`, pre-Space legacy messages, and forwarded/card messages.
- Conversation visibility in a **non-default** Space is decided by whether the single shared `Recents` window contains a message tagged with that Space (`personConvHasSpaceMessages` + `decideConvKeepInSpace`, `space_filter.go:347` / `:283`); the default Space always shows. Both Spaces share one Recents window, so the most-recently-active Space fills it and the other is filtered out → **symptom 2**.

## Impact

P1: cross-Space DM history leak (information disclosure) + functional loss (DMs vanishing per Space).

## Blocking product decision

Should a DM with the same contact be **one** conversation, or **independent per-Space** conversations? This decides the fix direction.

## Proposed directions

- *If per-Space DMs are required (fixes both):* introduce a real per-`(contact, Space)` conversation dimension — independent conversation row, unread, and message ownership — instead of inferring visibility from the shared Recents window.
- *If DMs are global by product definition:* exclude Person channels from Space filtering entirely (consistent across all Spaces).
- *Stop-gaps (mitigate only):* (a) audit senders omitting `X-Space-ID` via the existing `enrich_payload_space_id_empty` warn metric (`api.go:708`) and the forward/card paths; (b) base symptom-2 visibility on persisted per-Space message existence rather than the Recents window.

> Do **not** simply relax filtering to show DMs in all Spaces — that collapses straight into symptom 1.

## Relationship to prior issues

- #33 / #37 (closed) hardened the **send-side** authoritative `space_id` injection for PERSONAL dispatch. This issue is downstream of that: even with correct injection, untagged messages (legacy, no `X-Space-ID`, forwarded/card) still fail-open via `filterPersonMessagesBySpace` rule 2 → symptom 1.
- #153 (closed) backfilled resolved `space_id` into conversation entries, but **only for groups** — DM conversation-level `SpaceID` is intentionally left empty (`api_conversation.go:1584`).
- **New here:** symptom 2 (conversations mutually *disappearing* between Spaces) is not covered by any prior issue — they all addressed leakage / over-exposure, not per-Space *loss* of a DM. Root cause is visibility bound to a single shared `Recents` window, which is an architectural consequence of the single-channel + soft-tag design, not a regression of #33/#37/#153.

lml2468:
## Root Cause Analysis (GH#484 — Symptom 2)

`decideConvKeepInSpace` in `modules/message/space_filter.go` decides whether a DM conversation appears in a Space's sidebar by checking `hasSpaceMsg` — which scans `conv.Recents` (the last N messages) for `payload.space_id` matches.

Since DMs are a **single physical channel** shared across all Spaces (conversation-level `SpaceID` is intentionally left empty), the Recents window is also shared. When Space A is more recently active, its messages fill the Recents window and evict Space B's tagged messages. `personConvHasSpaceMessages(spaceB)` then returns false → conversation disappears from Space B's sidebar.

## Fix

Regular (non-bot) DM conversations are now always visible in every Space, regardless of Recents window contents. The fix is a one-line change in `decideConvKeepInSpace`:

```go
// Before:
return hasSpaceMsg != nil && hasSpaceMsg(filterSpaceID)
// After:
return true
```

Message-level filtering (`filterPersonMessagesBySpace`) still controls which messages are shown within each Space, so cross-Space message pollution (symptom 1) is not worsened by this change.

## Scope

This PR fixes **symptom 2** (mutually exclusive visibility / disappearing DMs).

**Symptom 1** (cross-Space history pollution via `filterPersonMessagesBySpace` rule 2 — keeping untagged messages in all Spaces) requires a product decision: should DMs be per-Space or global? The current behavior (untagged messages visible everywhere) is consistent with the "DMs are global" interpretation and is left as-is pending that decision. See #484 blocking product decision.

## Test

- Renamed `TestDecideConvKeepInSpace_DM_NonDefaultSpace_NoSpaceMsg_Excluded` → `TestDecideConvKeepInSpace_DM_NonDefaultSpace_AlwaysVisible` (asserts True).
- Added `TestDecideConvKeepInSpace_DM_NonDefaultSpace_RecentsEvicted_StillVisible` — regression test for the exact bug scenario where Recents has no Space B messages.

All 25 space_filter tests pass.

Fixes #484 (symptom 2)

### DIFF
--- a/modules/message/space_filter.go
+++ b/modules/message/space_filter.go
@@ -280,7 +280,16 @@ func decideConvKeepInSpace(
 			return true
 		}
 		if !botSet[channelID] {
-			return hasSpaceMsg != nil && hasSpaceMsg(filterSpaceID)
+			// GH#484 fix: regular (non-bot) DM conversations are always visible in
+			// every Space. DM is a single physical channel shared across Spaces; using
+			// the Recents window (personConvHasSpaceMessages) to decide visibility was
+			// unreliable because both Spaces share the same Recents, so whichever Space
+			// was active most recently would evict the other's tagged messages and make
+			// the conversation disappear. Message-level filtering
+			// (filterPersonMessagesBySpace) still ensures only relevant messages are
+			// shown per Space. Until a product decision introduces real per-(contact,
+			// Space) conversations, DMs are visible in all Spaces.
+			return true
 		}
 	}
 	return false
--- a/modules/message/space_filter.go
+++ b/modules/message/space_filter.go
@@ -198,6 +198,21 @@ func filterConversationsCore(
 			func(target string) bool { return personConvHasSpaceMessages(conv, target) },
 		)
 		if keep {
+			// GH#484 P1 (reviewer Jerry-Xin / yujiawei, PR #490): DM conversations are
+			// now visible in every Space, but the conversation-list Recents are NOT
+			// otherwise space-filtered on this path — unlike /v1/message/channel/sync,
+			// which runs filterPersonMessagesBySpace (api.go). Without scrubbing here, a
+			// DM kept in a non-active Space would ship its raw Recents (including
+			// messages tagged for another Space), re-opening cross-Space message-content
+			// exposure in the conversation-list preview. Scrub Recents with the same
+			// message-level rule so the preview matches exactly what channel/sync would
+			// return for this DM in this Space. filterPersonMessagesBySpace is a no-op
+			// when filterSpaceID == "" and only removes cross-Space-tagged / sysbot-only
+			// messages, so untagged legacy messages (symptom 1, pending product decision)
+			// are unchanged.
+			if filterSpaceID != "" && conv.ChannelType == common.ChannelTypePerson.Uint8() {
+				conv.Recents = filterPersonMessagesBySpace(conv.Recents, conv.ChannelID, filterSpaceID)
+			}
 			filtered = append(filtered, conv)
 		}
 	}
@@ -212,8 +227,11 @@ func filterConversationsCore(
 // 参数：
 //   - convSpaceID: 调用方已对 channel_id 做过 ParseChannelID 后得到的 space_id。
 //     可为空，群聊/子区会进一步查 groupSpaceMap。
-//   - hasSpaceMsg: 仅对非默认 Space 的 Person DM 生效，判断 conv.Recents 内是否
-//     有 payload.space_id == targetSpaceID 的消息。
+//   - hasSpaceMsg: TODO(#484) 现已不参与常规 DM 可见性判定。GH#484 修复后普通
+//     (非 bot) DM 在所有 Space 恒可见（见下方 ChannelTypePerson 分支），不再按
+//     conv.Recents 内是否有 payload.space_id == targetSpaceID 的消息来决定。该参数
+//     与两个调用方的闭包暂时保留：若后续产品决策引入真正的 per-(contact, Space)
+//     会话维度，Recents-scan 判定可能回归。目前是“按决策待定而有意保留”，非遗漏。
 //   - failClosedOnUnknownGroupSpace: 当 skipGroupFilter=true（group service 查询
 //     失败、无法确认群的 space_id）时的语义切换。
 //     - false（v1 兼容默认）：保留群/子区，不让一次 DB 抖动影响存量行为。
--- a/modules/message/space_filter_test.go
+++ b/modules/message/space_filter_test.go
@@ -47,23 +47,25 @@ func TestFilterConversationsBySpace_SystemBotsVisible(t *testing.T) {
 }
 
 func TestFilterConversationsBySpace_DefaultSpaceBareConvs(t *testing.T) {
-	// 裸 UID 旧会话只在默认 Space 显示
+	// GH#484: 裸 UID 旧 DM 会话在所有 Space 恒可见（不再只在默认 Space 显示）。
 	convs := []*SyncUserConversationResp{
 		{ChannelID: "user1", ChannelType: common.ChannelTypePerson.Uint8(), SpaceID: ""},
 		{ChannelID: "user2", ChannelType: common.ChannelTypePerson.Uint8(), SpaceID: ""},
 	}
 
-	// filterSpaceID == defaultSpaceID → 旧会话保留
+	// filterSpaceID == defaultSpaceID → 可见
 	result := filterConversationsCore(convs, "spaceA", "spaceA", nil, nil, nil, nil, false, false)
 	assert.Len(t, result, 2)
 
-	// filterSpaceID != defaultSpaceID → 无 Recents 匹配 → 不显示
+	// filterSpaceID != defaultSpaceID → GH#484 修复后仍可见（普通 DM 全 Space 可见）
 	result = filterConversationsCore(convs, "spaceB", "spaceA", nil, nil, nil, nil, false, false)
-	assert.Len(t, result, 0)
+	assert.Len(t, result, 2)
 }
 
-func TestFilterConversationsBySpace_NonDefaultSpaceDMVisible(t *testing.T) {
-	// 非默认 Space 中，普通 DM 需有 Recents 中 space_id 匹配才显示
+func TestFilterConversationsBySpace_NonDefaultSpace_AllNonBotDMsVisible(t *testing.T) {
+	// GH#484: 非默认 Space 中，普通 (非 bot) DM 无论 Recents 是否有 space_id 匹配
+	// 都恒可见（原契约要求 Recents 匹配，因单一物理 channel + 共享 Recents 而不可靠）。
+	// Bot DM 隔离不变：Bot 在此 Space → 可见；Bot 不在此 Space → 不可见。
 	convs := []*SyncUserConversationResp{
 		{
 			ChannelID: "user1", ChannelType: common.ChannelTypePerson.Uint8(), SpaceID: "",
@@ -83,21 +85,60 @@ func TestFilterConversationsBySpace_NonDefaultSpaceDMVisible(t *testing.T) {
 	// filterSpaceID=spaceB != defaultSpaceID=spaceA
 	result := filterConversationsCore(convs, "spaceB", "spaceA", nil, nil, botSet, botInSpace, false, false)
 
-	// user1（Recents 有 spaceB 消息）保留；user2（Recents 只有 spaceA）过滤；
-	// bot_in_space（Bot 在此 Space）保留；custom_bot（Bot 不在此 Space）不保留
-	assert.Len(t, result, 2)
+	// user1 + user2（普通 DM 全 Space 可见）+ bot_in_space（Bot 在此 Space）保留；
+	// custom_bot（Bot 不在此 Space）不保留。
+	assert.Len(t, result, 3)
 	ids := make([]string, len(result))
 	for i, r := range result {
 		ids[i] = r.ChannelID
 	}
 	assert.Contains(t, ids, "user1")
+	assert.Contains(t, ids, "user2")
 	assert.Contains(t, ids, "bot_in_space")
-	assert.NotContains(t, ids, "user2")
 	assert.NotContains(t, ids, "custom_bot")
 }
 
-func TestFilterConversationsBySpace_NewSpaceCleanSlate(t *testing.T) {
-	// 全新 Space：所有 DM 的 Recents 都没有该 Space 的消息 → 全部过滤
+func TestFilterConversationsBySpace_NonDefaultSpace_ScrubsCrossSpaceRecents(t *testing.T) {
+	// GH#484 P1 回归 (reviewer Jerry-Xin / yujiawei, PR #490): 普通 DM 现在在所有
+	// Space 可见，但 conversation-list 的 Recents 必须按当前 Space 过滤，否则会把
+	// 另一 Space 的消息内容（预览）泄露到本 Space。断言：会话保留可见，但只带
+	// 本 Space 标签的消息 + 无标签老消息，跨 Space 标签的消息被剔除。
+	convs := []*SyncUserConversationResp{
+		{
+			ChannelID: "peer_uid", ChannelType: common.ChannelTypePerson.Uint8(), SpaceID: "",
+			Recents: []*MsgSyncResp{
+				{Payload: map[string]interface{}{"space_id": "spaceA", "content": "secret from space A"}},
+				{Payload: map[string]interface{}{"space_id": "spaceB", "content": "hi in space B"}},
+				{Payload: map[string]interface{}{"content": "legacy untagged"}},
+			},
+		},
+	}
+
+	// 从 spaceB 视角拉列表（spaceB != defaultSpaceID=spaceA）。
+	result := filterConversationsCore(convs, "spaceB", "spaceA", nil, nil, map[string]bool{}, map[string]bool{}, false, false)
+
+	// 会话仍然可见（symptom 2 已修）。
+	assert.Len(t, result, 1)
+	assert.Equal(t, "peer_uid", result[0].ChannelID)
+
+	// Recents 已按 spaceB 过滤：spaceA 标签的消息被剔除（不泄露），
+	// spaceB 标签 + 无标签老消息保留。
+	contents := make([]string, 0, len(result[0].Recents))
+	for _, m := range result[0].Recents {
+		if c, ok := m.Payload["content"].(string); ok {
+			contents = append(contents, c)
+		}
+	}
+	assert.NotContains(t, contents, "secret from space A")
+	assert.Contains(t, contents, "hi in space B")
+	assert.Contains(t, contents, "legacy untagged")
+	assert.Len(t, result[0].Recents, 2)
+}
+
+func TestFilterConversationsBySpace_NewSpace_DMsAlwaysVisible(t *testing.T) {
+	// GH#484: 在一个全新 Space 里，即便没有任何 DM 的 Recents 命中该 Space，
+	// 普通 DM 依然全部可见（不再有 DM 的 "clean slate"）。消息级过滤仍会
+	// 决定每个 Space 内展示哪些消息（见 ScrubsCrossSpaceRecents 回归测试）。
 	convs := []*SyncUserConversationResp{
 		{
 			ChannelID: "user1", ChannelType: common.ChannelTypePerson.Uint8(), SpaceID: "",
@@ -115,8 +156,8 @@ func TestFilterConversationsBySpace_NewSpaceCleanSlate(t *testing.T) {
 
 	result := filterConversationsCore(convs, "spaceNew", "spaceA", nil, nil, map[string]bool{}, map[string]bool{}, false, false)
 
-	// 新 Space 没有任何 DM 有匹配消息 → clean slate
-	assert.Len(t, result, 0)
+	// 普通 DM 全 Space 可见 → 三条会话都保留。
+	assert.Len(t, result, 3)
 }
 
 func TestPersonConvHasSpaceMessages(t *testing.T) {
--- a/modules/message/space_filter_v2_test.go
+++ b/modules/message/space_filter_v2_test.go
@@ -86,7 +86,10 @@ func TestDecideConvKeepInSpace_DM_DefaultSpace_KeepBareConv(t *testing.T) {
 	assert.True(t, keep, "bare DM stays in user's default space when no bot match")
 }
 
-func TestDecideConvKeepInSpace_DM_NonDefaultSpace_NoSpaceMsg_Excluded(t *testing.T) {
+// GH#484: DMs are always visible in every Space (single-channel architecture).
+// Previously this asserted False — relying on the Recents window for visibility
+// caused conversations to disappear when the other Space was more recently active.
+func TestDecideConvKeepInSpace_DM_NonDefaultSpace_AlwaysVisible(t *testing.T) {
 	keep := decideConvKeepInSpace(
 		"peer1", common.ChannelTypePerson.Uint8(), "",
 		"spaceB", "spaceA",
@@ -94,7 +97,8 @@ func TestDecideConvKeepInSpace_DM_NonDefaultSpace_NoSpaceMsg_Excluded(t *testing
 		map[string]bool{}, map[string]bool{}, false, false, false,
 		func(string) bool { return false },
 	)
-	assert.False(t, keep, "DM with no payload.space_id match must NOT appear in non-default space")
+	assert.True(t, keep,
+		"GH#484: regular DM must always be visible in every Space (single-channel DM architecture)")
 }
 
 func TestDecideConvKeepInSpace_DM_NonDefaultSpace_HasSpaceMsg_Visible(t *testing.T) {
@@ -106,6 +110,22 @@ func TestDecideConvKeepInSpace_DM_NonDefaultSpace_HasSpaceMsg_Visible(t *testing
 		func(target string) bool { return target == "spaceB" },
 	)
 	assert.True(t, keep, "DM with payload.space_id == filter must appear in that space")
+}
+
+// GH#484 regression: when the Recents window contains no messages tagged with the
+// target Space (because the other Space was more recently active and evicted them),
+// the DM conversation must still be visible. Previously this returned false.
+func TestDecideConvKeepInSpace_DM_NonDefaultSpace_RecentsEvicted_StillVisible(t *testing.T) {
+	// hasSpaceMsg simulates Recents window with no spaceB messages (evicted by spaceA activity)
+	keep := decideConvKeepInSpace(
+		"peer1", common.ChannelTypePerson.Uint8(), "",
+		"spaceB", "spaceA",
+		nil, nil,
+		map[string]bool{}, map[string]bool{}, false, false, false,
+		func(target string) bool { return false }, // Recents has no spaceB messages
+	)
+	assert.True(t, keep,
+		"GH#484 regression: DM must remain visible even when Recents window has no space-tagged messages for that Space")
 }
 
 func TestDecideConvKeepInSpace_DM_BotInSpace(t *testing.T) {


OctoBoooot (commented):
## Review: fix: DM conversations always visible in all Spaces (#484) (#490)

**Verdict:** Approve (with one product-semantics confirmation called out — PR is a **draft**, so this is feedback-while-in-progress).

CI green. Risk: normal, but trust-boundary-adjacent (controls which conversations appear in which Space's sidebar). 2 files +32/-3.

### Verified — the fix is correctly scoped to non-bot DMs; bot isolation preserved

The change is a single `return true` in the `spaceID == "" && ChannelTypePerson` branch of `decideConvKeepInSpace` (`space_filter.go:281-290`). Byte-verified the guard ordering above it preserves all the isolation-relevant cases:
- `skipBotFilter` → true (unchanged)
- `SystemBots[channelID]` → true (unchanged)
- bot AND `botInSpace[channelID]` → true (unchanged)
- **non-bot DM (`!botSet[channelID]`) → now always true** (was: `hasSpaceMsg(filterSpaceID)` Recents-scan)
- a bot NOT in the space still falls through to `return false` at function end — **bot space-isolation is NOT widened**.

So the visibility widening is confined to human-to-human DMs. A bot that isn't a member of the filter Space still does not appear — the change doesn't touch the bot path that the prior PRs (#458/#469) worked to keep isolated. [bidirectional-edge-construction] check: the `return true` can't leak a bot conversation into a wrong space because the bot branches are evaluated and returned before this line.

### Root-cause + fix logic is sound

The Recents-window-scan visibility check was genuinely unreliable for the single-physical-channel DM architecture: both Spaces share one Recents window, so the more-active Space evicts the other's tagged messages → `personConvHasSpaceMessages(spaceB)` returns false → conversation vanishes from Space B. Making non-bot DM visibility unconditional fixes the disappearing-DM symptom (symptom 2). Message-level filtering (`filterPersonMessagesBySpace`) still controls which messages render per-Space, so this only changes *conversation-list visibility*, not message content. The PR correctly scopes itself to symptom 2 and defers symptom 1 (cross-Space message pollution) to a product decision.

### Needs confirmation — the privacy/product semantics of "DM visible in all Spaces"

This is the one thing worth an explicit human nod before un-drafting (it's why I'm flagging rather than plain-approving): the change makes every regular DM conversation appear in **every** Space's sidebar for that user. The PR frames this as the "DMs are global" interpretation, which is internally consistent with the existing untagged-message-visible-everywhere behavior. But it IS a visibility-surface change:
- A user who has DMs with contact X, and belongs to Spaces A and B, will now see the X conversation in both A and B's sidebars (previously: only where Recents happened to have a matching tag).
- This is a per-user view (the DM is the user's own conversation), so it's not a cross-user leak — the user already has access to their own DM. The question is purely UX/product ("should my DMs show in every Space tab, or be Space-scoped?"), not authz.

The PR body explicitly says symptom-1 / per-Space-vs-global is a pending product decision (#484). Since this change leans on the "global" interpretation, the same product decision governs whether this is the desired end state. Recommend an explicit product/stakeholder ack on "DMs visible in all Spaces" before merge — it's a deliberate product choice, correctly implemented, not a code defect.

### Tests pin both directions

- `TestDecideConvKeepInSpace_DM_NonDefaultSpace_AlwaysVisible` (renamed from `..._NoSpaceMsg_Excluded`) flips the assertion `False`→`True` with a comment explaining the GH#484 rationale — the rename + flip is the honest way to update a test whose expected behavior intentionally changed (not silently editing the assertion).
- New `TestDecideConvKeepInSpace_DM_NonDefaultSpace_RecentsEvicted_StillVisible` pins the exact regression: `hasSpaceMsg` returns false (Recents evicted) → still visible. This is the bug's precise failing input, captured as a test.
- `TestDecideConvKeepInSpace_DM_BotInSpace` and the bot cases remain — confirming the bot path is unchanged.

### Praise

- **The fix flips the test assertion via rename + comment, not a silent edit** — `..._NoSpaceMsg_Excluded` → `..._AlwaysVisible` with a comment explaining why the expected behavior changed. When a fix intentionally inverts a previously-correct assertion, renaming the test to describe the new contract (rather than editing the bool in place) makes the behavior change reviewable in the diff. Right discipline.
- **Regression test captures the exact failing input** — `RecentsEvicted_StillVisible` with `hasSpaceMsg → false` is the precise condition that made DMs vanish. A future "optimization" that re-introduces a Recents-based visibility gate fails this test loudly.
- **Scope honesty** — the PR fixes symptom 2 and explicitly defers symptom 1 (cross-Space pollution) to a product decision rather than silently changing both. The comment + body both name what's fixed and what's deliberately left. Clean scoping on a two-symptom issue.
- **Bot isolation preserved** — the `return true` is placed after all the bot-branch guards, so the human-DM widening doesn't touch the bot space-isolation that #458/#469 established. The change is minimal and surgical.

### Minor (non-blocking)

- The `hasSpaceMsg func(targetSpaceID string) bool` param + `personConvHasSpaceMessages` helper are now effectively dead on the non-bot DM path (the only caller of that branch's check). The caller (`space_filter.go:198`) still constructs and passes it. Leaving it wired (rather than removing) is reasonable while symptom-1's product decision is pending — if per-(contact, Space) conversations land later, the Recents check may come back. Worth a `// TODO(#484): hasSpaceMsg unused on the always-visible path; remove if global-DM is the final decision` so the next reader knows it's intentionally-retained-pending-decision, not an oversight.

- Symptom 1 (untagged DM messages visible in all Spaces via `filterPersonMessagesBySpace` rule 2) is correctly deferred to the #484 product decision. The "DMs are global" interpretation this PR adopts should be confirmed as the intended end state, since both symptoms hinge on the same per-Space-vs-global question.

Jerry-Xin (changes requested):
**REQUEST_CHANGES** — The GH#484 direction (make non-bot DM conversations visible in every Space, since DM is a single shared physical channel) is reasonable and the v2 unit tests added here directly cover the disappearing-DM scenario. However the change is not mergeable yet: it breaks existing v1 tests and introduces a cross-Space message-preview exposure in the conversation-list sync path.

**🔴 Blocking**

- 🔴 **Existing v1 tests now fail (CI break).** `decideConvKeepInSpace` is shared by both v1 (`filterConversationsCore`/`FilterConversationsBySpace`) and v2 sidebar. This PR only updated `modules/message/space_filter_v2_test.go`, but the v1 tests in `modules/message/space_filter_test.go` (untouched) still assert the old "hidden when no Recents match" contract and now fail:
  - `TestFilterConversationsBySpace_DefaultSpaceBareConvs` (`space_filter_test.go:62`) — expects 0 in non-default Space, now gets 2.
  - `TestFilterConversationsBySpace_NonDefaultSpaceDMVisible` (`space_filter_test.go:88`,`:95`) — asserts `user2` excluded, now included.
  - `TestFilterConversationsBySpace_NewSpaceCleanSlate` (`space_filter_test.go:119`) — expects clean-slate 0, now gets 3.
  Verified locally at head `e63ae5d0`: `go test ./modules/message -run 'TestFilterConversationsBySpace'` → FAIL. Update these tests (and the now-incorrect "clean slate" comment) to the new DM contract.

- 🔴 **Cross-Space DM `Recents` exposure in the conversation-list response.** `modules/message/space_filter.go:290` now returns `true` for any non-bot DM in any Space, but the conversation-list sync path filters DMs only via `FilterConversationsBySpace` (`api_conversation.go:893`), which keeps/drops the whole conversation and does **not** scrub each conv's `Recents` by `space_id`. The code comment claims `filterPersonMessagesBySpace` still scopes messages per Space — but that function is only called on the channel-history sync path (`api.go:1279`), never on the conversation-list `Recents`. Result: a Space B conversation sync can now return a DM whose `Recents` contain only Space A messages (the exact "evicted" case the PR targets), exposing Space A's latest DM preview/content in Space B's list. Previously such a DM was dropped, so the preview never leaked. Please filter/clear DM `Recents` to the requested Space in the conversation-list path when making DMs globally visible, and add a regression test for the "Recents contain only the other Space's messages" response shape.

**💬 Non-blocking**

- 🔵 The `hasSpaceMsg` helper doc/contract at `space_filter.go:215` is now stale — after this change it no longer governs regular DM visibility (only bot-related paths and v2 still reference it). Update the comment to avoid confusion.

**✅ Highlights**

- Behavior change is correctly gated to the `ChannelTypePerson` + non-bot branch; group/community-topic and bot-DM paths (incl. v2 fail-closed group scoping) are untouched, so there is no group/space-channel widening and no cross-*user* exposure — the conversation list is already `loginUID`-scoped.
- New v2 tests (`..._AlwaysVisible`, `..._RecentsEvicted_StillVisible`) clearly document and cover the GH#484 disappearing-DM regression.

Reviewed at head `e63ae5d099a376b708a958d9688b259b535d54af`.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#490 Review Report
**PR:** Mininglamp-OSS/octo-server#490
**Head SHA:** `e63ae5d099a376b708a958d9688b259b535d54af`
**Title:** fix: DM conversations always visible in all Spaces (#484)
**Changed files:** 2 (+32 / -3)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Code correctness | ✅ | Single return-statement change in `decideConvKeepInSpace` non-bot DM branch; control flow intact for Group/Thread/Bot paths |
| Test coverage | ✅ | 25/25 space_filter tests pass; renamed test + new regression test cover the fix |
| Bot DM isolation preserved | ✅ | `botSet[channelID] && !botInSpace[channelID] → false` path (line 276) unchanged; bot-not-in-space still excluded |
| Message-level filtering unchanged | ✅ | `filterPersonMessagesBySpace` (lines 451–480) untouched; rules 1–4 intact |
| v1/v2 parity | ✅ | Both `filterConversationsCore` and `FilterRawConversationsBySpace` call the same `decideConvKeepInSpace`; change applies symmetrically |

---

## 2. Findings

### F1 — P2 — Message-level filtering does NOT fully prevent cross-Space message visibility

**Diff-scope:** pre-existing limitation, amplified by this PR's conversation-visibility change

The PR description claims: *"Message-level filtering (`filterPersonMessagesBySpace`) still controls which messages are shown within each Space, so cross-Space message pollution (symptom 1) is not worsened by this change."*

This is partially correct but **overstated**. `filterPersonMessagesBySpace` rule 2 (`space_filter.go:470–472`) keeps untagged non-systembot messages in every Space:

```go
case msid == "" && !isSysBot:
    filtered = append(filtered, m) // legacy compat — kept in ALL Spaces
```

Untagged messages include: legacy pre-Space messages, messages from clients omitting `X-Space-ID`, and forwarded/card messages. These remain visible in every Space after message-level filtering.

**Impact analysis (R1 check):**
- Not "unavailable": DM conversations were already in the user's conversation list (default Space); the PR changes which Space sidebars show them.
- Not "incorrect data": the same untagged messages were accessible before (via default Space sidebar); no new data exposed.
- Not "worse than before" in terms of data exposure: tagged messages (spaceA-tagged) are still properly dropped when viewing from spaceB (rule 4). Only untagged messages cross — same as before.

**Severity:** P2. The information disclosure is pre-existing; the PR changes the access path (sidebar convenience) but not the total data surface.

### F2 — P2 — PR description mischaracterizes "Fixes #484" scope

The PR says *"Fixes #484 (symptom 2)"* but GitHub's `Fixes` keyword will auto-close the issue on merge. Issue #484 covers **two** symptoms; symptom 1 (cross-Space history pollution via untagged messages) is explicitly deferred pending a product decision. The PR body acknowledges this in the Scope section, but the `Fixes` keyword contradicts it — merging would close #484 as resolved while symptom 1 remains open.

**Recommendation:** Change `Fixes #484` to `Fixes #484 symptom 2` or `Partial fix for #484` (remove the GitHub auto-close keyword), or use `Contributes to #484`.

### F3 — P2 (nit) — Dead `hasSpaceMsg` callback parameter

After this change, the `hasSpaceMsg` callback in `decideConvKeepInSpace` is never invoked for non-bot DMs (the `!botSet[channelID]` branch returns `true` before reaching the callback). Both callers (`filterConversationsCore` line 198, `FilterRawConversationsBySpace` line 686) still construct and pass closures that are now dead code for this path. The callback is still used for the bot-in-default-space path (line 271), so the parameter can't be removed, but the closures in the non-bot DM callers are now wasted computation. Minor — no correctness impact.

---

## 3. Recommendations

1. **Fix `Fixes` keyword** (F2): Replace `Fixes #484 (symptom 2)` with `Partial fix for #484: symptom 2 (mutually-exclusive visibility)` to avoid auto-closing the umbrella issue.

2. **Clarify PR description** (F1): The claim *"cross-Space message pollution (symptom 1) is not worsened"* is technically accurate at the message-filter level but misleading at the user-experience level. Consider rewording to: *"Message-level filtering is unchanged; symptom 1 (untagged messages visible in all Spaces) is a pre-existing architectural limitation that requires a product decision to resolve."*

3. **Document the product decision**: The PR makes an implicit product choice ("DMs are global at conversation level") that was listed as one of three options in #484. Consider adding a note in the PR or #484 that this decision has been made, with a tracking item for the full per-(contact, Space) architecture if/when product decides to go that direction.

---

## 4. Additional Observations

- **Test `TestDecideConvKeepInSpace_DM_NonDefaultSpace_HasSpaceMsg_Visible`** is now redundant with `AlwaysVisible` — both assert `true` for non-bot DM in non-default Space. The `HasSpaceMsg` variant tests the now-unreachable callback path. Not a bug, but could be simplified.
- **Backward compatibility**: Clients that previously did NOT see a DM in a non-default Space sidebar will now see it. No client-side breaking change; just new visibility. Three-platform rendering should be verified (web/iOS/Android) to confirm no client assumes conversation-in-sidebar implies all messages are Space-tagged.

---

## 5. Data Flow Trace

| Consumer | Data | Upstream Source | Runtime Flow Verified |
|----------|------|----------------|----------------------|
| `decideConvKeepInSpace` (modified branch) | `!botSet[channelID]` | `resolveBotFilter` → `spacepkg.GetBotUIDs` DB query | ✅ Non-bot DMs reach this branch; bots take earlier paths |
| `decideConvKeepInSpace` return value | `true` (was `hasSpaceMsg(filterSpaceID)`) | Hardcoded | ✅ Always returns true for non-bot DMs |
| `filterConversationsCore` | `keep` from `decideConvKeepInSpace` | Same function | ✅ Non-bot DM conversations always kept |
| `FilterRawConversationsBySpace` | `keep` from `decideConvKeepInSpace` | Same function | ✅ v2 sidebar parity confirmed |
| `filterPersonMessagesBySpace` (api.go:1279) | `msgs`, `channelID`, `spaceID` | `/v1/message/channel/sync` handler, middleware-validated `X-Space-ID` | ✅ Unchanged; rules 1–4 intact. Rule 2 (untagged non-sysbot kept) is the pre-existing gap. |
| `personConvHasSpaceMessages` / `rawConvHasSpaceMessages` | callback passed to `decideConvKeepInSpace` | Callers construct closures | ⚠️ Callback now dead code for non-bot DM path (F3) |

---

- **C1 — Dual-path parity:** N/A. The change affects only the non-bot DM branch. Bot paths (add/remove from space) are symmetric and unchanged.
- **C2 — Control-flow ordering / nested reuse:** Clear. `decideConvKeepInSpace` is called from v1 and v2 filter paths; both reach the same modified branch. No nesting or double-invocation risk.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. Space sidebar filtering is UX-level, not authorization. DM messages are accessible to the conversation participants regardless of Space. No authorization boundary crossed.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth changes in this PR.
- **C5 — Build ≠ runtime:** N/A. Go backend change; no build artifacts / extensions / CLI packaging involved.
- **C6 — Governance/policy doc self-consistency:** Clear with note. The PR description's `Fixes #484` claim conflicts with #484's explicit "Do not simply relax filtering" guidance. See F2 recommendation.

---

N/A — automated review review.

---

## Summary

| # | Severity | Category | Description |
|---|----------|----------|-------------|
| F1 | P2 | Process/claim | Message-level filtering doesn't fully prevent cross-Space visibility for untagged messages |
| F2 | P2 | Process | `Fixes #484` will auto-close issue while symptom 1 remains unresolved |
| F3 | P2 (nit) | Cleanup | Dead `hasSpaceMsg` callback for non-bot DM path |

The code change is clean, focused, and well-tested. It correctly fixes symptom 2 (disappearing DMs between Spaces) without introducing code-level regressions. The primary concerns are process-level: the PR makes an implicit product decision, contradicts #484's explicit guidance, and its description slightly overstates the security properties of message-level filtering. None of these rise to P0/P1 by the R1 rubric (no new unavailability, no new incorrect data, no regression in total data exposure).

**[Octo-Q] verdict: APPROVE**

The fix is technically sound and addresses a real user-facing bug. The P2 findings are process/documentation concerns that should be addressed (especially F2 — the `Fixes` keyword issue) but do not block merge from a code-correctness standpoint.

OctoBoooot:
Concurring with @Jerry-Xin's 🔴-2 — and correcting a claim in my own review (https://github.com/Mininglamp-OSS/octo-server/pull/490#pullrequestreview-4585179610). Jerry-Xin is right; I missed it.

My review asserted: *"Message-level filtering (filterPersonMessagesBySpace) still controls which messages render per-Space, so this only changes conversation-list visibility, not message content."* That conflated two different code paths and is wrong for the sidebar preview.

Byte-verified: `filterConversationsBySpace` (`space_filter.go:189-202`) appends the **whole `conv`** when `decideConvKeepInSpace` returns true — `conv.Recents` is NOT filtered. And `fillMissingPersonConversations`'s comment at `:392` states explicitly "对 Recents 内 payload.space_id 字段不做任何修改" (Recents space_id is not modified). The conversation-sync sidebar preview comes from `conv.Recents[0]`, which is the unfiltered shared Recents window.

So the concrete leak:
- User in Spaces A and B, DM with contact X. Last message in the shared DM channel was tagged `space_id=A`.
- Pre-PR: the DM didn't appear in Space B's sidebar at all (the bug being fixed), so no preview.
- Post-PR: the DM is now always visible in Space B's sidebar — and its preview snippet is `Recents[0]`, the Space-A-tagged message. **Space B's sidebar shows a Space-A message preview.**

`filterPersonMessagesBySpace` (the thing I cited) governs the message-LIST endpoint when you open the conversation, NOT the conversation-SYNC sidebar preview. Two different layers — I credited the wrong one as the guard. The sidebar preview is genuinely unfiltered after this change.

Severity: this is a real cross-Space preview exposure (Jerry-Xin's 🔴). It's still per-user (the user owns the DM), so not a cross-user authz leak — but it surfaces a Space-A message in Space-B's UI, which is exactly the Space-isolation expectation this codebase invests in. It also undercuts the PR's own symptom-1-deferral framing: the PR says "message-level pollution is deferred", but this change actively introduces preview-level pollution that didn't exist before (the DM wasn't visible in Space B at all before).

Fix direction (matching Jerry-Xin): when keeping a DM visible in a non-tagging Space, filter `conv.Recents` to that Space's messages for the preview (or blank/placeholder the preview if no Space-tagged message exists), so the sidebar snippet doesn't show another Space's content. That keeps the disappearing-DM fix (symptom 2) while not introducing preview pollution.

Jerry-Xin's 🔴-1 (a v1 test / CI) — CI currently shows only the `code-review` bot check failing, not a Go test; worth the author confirming whether a v1 test path also needs the assertion flip (I verified `round3_findings_test.go` doesn't assert the old DM-excluded behavior, so the v1 break may be elsewhere or already green — author should check `go test ./modules/message/...`).

Applying [praise-has-weight] retroactively: I cited a downstream filter as the guard without byte-verifying it actually covers the sidebar-preview path. "Message filtering still applies" was a path-confusion — the conversation-sync preview and the message-list endpoint are separate layers, and only the latter has the filter I named. My code-side APPROVE is retracted on this; the fix needs the Recents preview filtered before it's shippable.

yujiawei (changes requested):
# Code Review — PR #490 (octo-server)

**Verdict: CHANGES_REQUESTED**

Thanks for the clear root-cause write-up — the diagnosis of symptom 2 (visibility bound to the shared `Recents` window) is correct. However, the chosen fix has a blocking test break and, more importantly, contradicts its own scope claim: making every DM unconditionally visible in every Space re-opens cross-Space message exposure on the conversation-sync path, which is exactly what issue #484 warned against ("Do NOT simply relax filtering to show DMs in all Spaces — that collapses straight into symptom 1").

---

## 1. Scope / spec compliance

The PR description states:

> Message-level filtering (`filterPersonMessagesBySpace`) still controls which messages are shown within each Space, so cross-Space message pollution (symptom 1) is not worsened by this change.

This is **not accurate for the conversation-sync path**. `filterPersonMessagesBySpace` only runs on `/v1/message/channel/sync` (`modules/message/api.go:1277-1281`). It is **not** applied to the `recents` array that `/v1/conversation/sync` serializes (`newSyncUserConversationResp`, `modules/message/api_conversation.go:1469-1555`). Before this change, a DM with no Space-B-tagged messages was excluded from Space B, so its Space-A `recents` never reached a Space-B client. After this change every such DM is returned, with its raw (often Space-A) `recents` content included. So the change does materially widen cross-Space exposure on this endpoint — symptom 1, on a different route.

This needs either (a) a product sign-off that "DMs are global" (the issue lists this as a *blocking product decision*, not yet made), or (b) scoping the visibility relaxation so it does not also ship unfiltered cross-Space message content.

## 2. Code quality

### P0 — Build/CI break: stale tests in `space_filter_test.go`

The PR updated `space_filter_v2_test.go` but left three tests in `space_filter_test.go` asserting the **old** "DM excluded from non-default Space" rule. They now fail, so `go test ./modules/message/...` is red and the merge gate cannot pass:

- `TestFilterConversationsBySpace_DefaultSpaceBareConvs` (`space_filter_test.go:60-62`) — expects 0 DMs in a non-default Space, now gets 2.
- `TestFilterConversationsBySpace_NonDefaultSpaceDMVisible` (`space_filter_test.go:84-96`) — expects `user2` (only Space-A `recents`) excluded, now kept.
- `TestFilterConversationsBySpace_NewSpaceCleanSlate` (`space_filter_test.go:116-119`) — expects 0 DMs in a brand-new Space, now gets 3.

Confirmed locally:
```
--- FAIL: TestFilterConversationsBySpace_DefaultSpaceBareConvs
--- FAIL: TestFilterConversationsBySpace_NonDefaultSpaceDMVisible
--- FAIL: TestFilterConversationsBySpace_NewSpaceCleanSlate
FAIL	github.com/Mininglamp-OSS/octo-server/modules/message
```
These are pure unit tests (no DB/build tags), so they run in normal CI. Update them to the new contract.

### P1 — `/v1/conversation/sync` leaks raw DM `recents` across Spaces

As described in §1: `recents` is built from the raw IM messages and is not space-filtered. `fillPersonSpaceUnread` adds `space_last_message` / `space_unread` (`modules/message/space_unread.go:38-58`) but does **not** replace or strip `recents` (`api_conversation.go:1555`). Result: a Space-B `/v1/conversation/sync` response can contain a DM whose `recents` payloads are Space-A message content. Even if the client renders `space_last_message` for the preview, the raw message bodies have already crossed the Space boundary on the wire. Either filter `recents` by Space here (reusing the `filterPersonMessagesBySpace` rule), or don't surface the conversation when it has no in-Space content.

### P1 — Aggregate `unread` badge is not Space-scoped

`SyncUserConversationResp.Unread` (`api_conversation.go:1453`, `:1547`) and the sidebar `SidebarItem.Unread` (`api_sidebar.go:1163` `buildRecentItems`) are serialized from the **raw aggregate** `conv.Unread`. With all DMs now visible in every Space, a Space-B sidebar shows DMs whose unread counts are driven entirely by Space-A messages. `space_unread` is computed separately, but any client reading the raw `unread` field gets cross-Space-polluted badges. This is a visible functional regression of the "per-Space" promise.

### P2 — Dead code & now-incorrect comments

After the change, `decideConvKeepInSpace` no longer calls its `hasSpaceMsg` parameter for the non-bot DM branch, yet both call sites still build the closures (`personConvHasSpaceMessages` at `space_filter.go:198`, `rawConvHasSpaceMessages` at `:703`) and the parameter doc at `space_filter.go:215` still claims it "decides non-default Space Person DM visibility" — which is no longer true. If the always-visible decision stands, remove the now-unused parameter + helpers (or document why they're retained) so the next reader isn't misled.

### P2 — Payload bloat & context leak in scoped Spaces

Returning every historical DM in every Space inflates the conversation-list payload for heavy DM users and surfaces "ghost" DM entries that have no visible messages in the current Space. It also exposes the full contact/DM list (names, avatars) into strictly-scoped Spaces, weakening the isolation users expect (e.g. when screen-sharing a work Space). Worth weighing in the product decision referenced in #484.

## 3. Suggested direction

The issue explicitly flags this as needing a product decision ("one conversation vs. per-Space conversations"). If "DMs are global" is confirmed, the cleaner stop-gap the issue itself suggests is to **exclude Person channels from Space filtering entirely AND make the conversation-sync/sidebar message-content + unread consistently global-or-scoped** — not to flip visibility to always-true while leaving `recents`/`unread` carrying per-Space-inconsistent data. At minimum, before merge:

1. Fix the three failing tests.
2. Either space-filter `recents` on `/v1/conversation/sync` or gate visibility so no unfiltered cross-Space content ships (resolve the §1 contradiction).
3. Decide whether `unread` should be the space-scoped value on these endpoints.
4. Remove/relabel the dead `hasSpaceMsg` path.

## 4. Coverage notes

- Verified against head SHA `e63ae5d099a376b708a958d9688b259b535d54af`; build break reproduced locally on this branch.
- Not verified at runtime: actual client rendering (whether mobile/web use `recents` vs `space_last_message` for the preview) and the real payload-size impact for high-volume users — these need a product/QA call, not a static read.
- The WebSocket realtime-routing path and v2 sidebar metadata (contact existence/unread/timestamp) are affected at the metadata level but do not carry message bodies; the content leak is specific to `/v1/conversation/sync` `recents`.

OctoBoooot (commented):
## Re-review (delta e63ae5d → 9a10a8b, draft): #490

**Verdict:** Comment — draft PR, and one coordination flag that matters more than the code.
The code change is small and the content-leak scrub is byte-verified sound. But #490 and #519 are two different fixes for the same issue (#484 symptom 2), and #519 has already completed review and is merge-ready. That overlap needs a human decision before either lands — flagging as the top item.

CI: Build/Test/Vet/Lint are SKIPPED (draft); `check-sprint` green. No code-CI signal yet — draft, so no APPROVE regardless.

### Coordination (top priority, not a code defect)
- **#490 and #519 solve #484 symptom 2 with opposite strategies.** #519: *hide* elsewhere-only DMs from a Space (presence-index + fail-closed catch-all). #490: make DMs *always visible* in every Space (single-channel argument) and scrub cross-Space content from the preview. Both are internally coherent; they are not mergeable together. #519 is already review-complete. Someone owning #484 should pick one approach before this proceeds, or #490 should be explicitly reframed as superseding #519.

### What I verified in the code (the change is 3 files, +107/-19; rest of the compare is rebase drift)
- **Content-leak scrub is correct.** `space_filter.go:~213` — after keeping a DM, `conv.Recents = filterPersonMessagesBySpace(conv.Recents, conv.ChannelID, filterSpaceID)` drops cross-Space-tagged messages (`msid != spaceID → continue`) and sysbot-only untagged messages, keeping this-Space + untagged-legacy. So the Recents preview no longer ships another Space's message content. Gated `filterSpaceID != ""`, which holds on this path (`hasSpaceFilter := filterSpaceID != ""` guards the call).
- **Space-scoped preview stays correct.** `fillPersonSpaceUnread` runs *before* the scrub (`api_conversation.go:891`) and sets `SpaceLastMessage` via `findSpaceLastMessage(conv.Recents, spaceID)` (space-matching only) with a WuKongIM fallback when Recents evicted it. Ordering is right — the preview is computed from the full Recents before the scrub trims it. Good.
- **Tests** encode the new contract well: `_AllNonBotDMsVisible`, `_ScrubsCrossSpaceRecents` (asserts "secret from space A" is removed while spaceB + legacy remain), `_RecentsEvicted_StillVisible`. The scrub regression test is exactly the leak this approach must guard.

### Question / minor — residual cross-Space metadata (not content)
- `SyncUserConversationResp` still ships **global** `Unread`, `LastMsgSeq`, `Timestamp`, `LastClientMsgNo` — channel-level values bumped by activity in *any* Space. After this fix a DM appears in Space B with an unread badge / timestamp driven by a Space-A message, even though the Space-B preview content is correctly scrubbed and `SpaceUnread`/`SpaceLastMessage` carry the correct per-Space view. That leaks *activity metadata* across Spaces, not content. Pre-existing fields, not introduced here — but under the "DMs always visible everywhere" model this becomes user-visible (every DM shows in every Space with a global badge). Worth confirming the client renders `SpaceUnread`/`SpaceLastMessage`, not `Unread`/`Timestamp`, for the per-Space list — otherwise symptom-2's fix trades a content leak for a metadata/badge leak.

### Out of scope (author already scoped this)
- Symptom 1 (untagged messages visible in all Spaces) is explicitly deferred to a product decision in the PR body. `filterPersonMessagesBySpace` here keeps untagged-in-all-Spaces (the older 2-arg rule), consistent with that framing.

### Praise
- Filling `SpaceLastMessage` before scrubbing `Recents` is the correct ordering, and the `_ScrubsCrossSpaceRecents` test names the exact hostile message ("secret from space A") it must exclude — that's the right way to lock a content-leak fix.

Jerry-Xin (dismissed):
**APPROVE** — Re-review of `9a10a8b4` (rebased onto main; PR's own diff vs merge-base is exactly the 3 `space_filter*` files — the large compare list is merge-base drift, not this PR). Both blockers from the prior REQUEST_CHANGES (@ `e63ae5d0`) are resolved.

**✅ Prior blockers fixed**
- **v1 test break (CI):** `space_filter_test.go` updated to the new "DM always visible in every Space" contract; `space_filter_v2_test.go` flipped `NoSpaceMsg_Excluded`→`AlwaysVisible` and added `RecentsEvicted_StillVisible`. Verified locally on this SHA: `go test ./modules/message/` → **ok**, `go vet` clean. (Note: CI `Test`/`Vet`/`Build` show **skipping** via path-detection, so the suite was not CI-validated — I validated locally instead.)
- **Cross-Space Recents preview leak:** `filterConversationsCore` now scrubs kept DM previews (`space_filter.go:213-215`): `if filterSpaceID != "" && conv.ChannelType == Person { conv.Recents = filterPersonMessagesBySpace(conv.Recents, conv.ChannelID, filterSpaceID) }`. This reuses the same authoritative rule as `/v1/message/channel/sync` (`api.go:1359`), so the conversation-list preview now matches exactly what channel/sync returns for that DM in that Space.

**Isolation invariant verified (byte-level)**
- `filterPersonMessagesBySpace` **drops** cross-Space-tagged messages (`msid != spaceID`), keeps current-Space + untagged-legacy, drops sysbot-untagged. On any ambiguity it degrades toward **hiding**, not showing → correct direction for a leak fix. The DM-visibility `return true` (`decideConvKeepInSpace`) is the intended symptom-2 product behavior, NOT a dependency-error fail-open on `defaultSpaceID` — the leak gate isn't reopened.
- Guard `filterSpaceID != ""` is sound: empty filterSpaceID means "no Space filter active" (global view) and `filterPersonMessagesBySpace` itself no-ops on empty, so it's consistent, not a bypass.
- **v2 path (`FilterRawConversationsBySpace`) does not scrub Recents — checked and it's NOT a leak:** the v2 `SidebarItem` response struct carries only metadata (target id, sort, unread, timestamp, category, space_id) — no message content/preview derived from `conv.Recents`. Only the v1 `SyncUserConversationResp` ships `Recents`, and that path is now scrubbed.

**Test coverage (bidirectional, good)**
- `TestFilterConversationsBySpace_NonDefaultSpace_ScrubsCrossSpaceRecents`: conv stays visible AND spaceA-tagged message dropped from preview while spaceB + untagged kept.
- `TestDecideConvKeepInSpace_DM_NonDefaultSpace_RecentsEvicted_StillVisible`: not over-hidden when Recents window has no current-Space messages.

**Other**
- No migrations, no committed secrets in the PR's own diff.
- 💬 Non-blocking: the `hasSpaceMsg` param + both call-site closures are now unused for regular-DM visibility (intentionally retained per the updated doc comment for a possible future per-(contact,Space) model). Fine as-is; could be dropped later if that model doesn't land.

Symptom-1 (untagged legacy messages global) is explicitly out of scope pending product decision — untagged messages remain kept, unchanged. LGTM.

Jerry-Xin (commented):
Re-posting as **COMMENT** (was APPROVE) — #490 is a **draft**, so a formal APPROVE isn't appropriate (draft status blocks merge and shouldn't be signaled ready). The code review below stands; nothing here is a blocker I'm raising.

## Code: the content-leak scrub is byte-verified sound
Real change is 3 files (`space_filter.go`, `space_filter_test.go`, `space_filter_v2_test.go`); the rest of the compare is rebase drift (ahead:16/behind:1).
- The prior REQUEST_CHANGES blockers (`e63ae5d`) are resolved: v1+v2 tests updated to the new contract (local `go test ./modules/message/` = ok, `go vet` clean), and the cross-Space Recents preview leak is scrubbed at `space_filter.go:213-215` via `filterPersonMessagesBySpace` (drops cross-Space-tagged msgs, keeps current-Space + untagged → degrades toward hiding).
- Ordering verified correct: `fillPersonSpaceUnread` (`:891`) computes the space-scoped `SpaceLastMessage`/`SpaceUnread` from the full Recents **before** the scrub trims it, so the per-Space preview stays correct.
- v2 `FilterRawConversationsBySpace` doesn't scrub Recents — checked, NOT a leak: v2 `SidebarItem` carries only metadata (no message content); only v1 ships Recents, and that path is scrubbed.
- Bidirectional test coverage present (`ScrubsCrossSpaceRecents` + `RecentsEvicted_StillVisible`). No migration, no committed secrets.

## Honest caveats (not code blockers)
1. **Metadata (not content) leak, worth confirming client behavior:** the response still ships **global** `Unread`/`LastMsgSeq`/`Timestamp` (channel-level, bumped by any Space's activity). Content is scrubbed, but under "DMs visible in every Space" a DM shows in every Space with a global unread badge. Confirm the client renders the space-scoped `SpaceUnread`/`SpaceLastMessage` for the per-Space list, else symptom-2's content fix trades for a badge/activity-metadata leak.
2. **Coordination — the real open item:** #490 and #519 are competing fixes for the same #484 symptom-2, with opposite strategies (#490 = DMs always-visible + scrub preview content; #519 = hide elsewhere-only DMs). #519 is already review-complete and merge-ready. They can't both land — the #484 owner needs to pick one. Symptom-1 is correctly deferred to a product decision by the author.

yujiawei (approved):
# Code Review — PR #490 (octo-server)

Reviewed at head `9a10a8b44db516e0dbecd08ead3fddc3ac3211c6` against merge-base `236b78b7`. This head contains two commits: the original one-line visibility fix, plus a follow-up commit that scrubs conversation-list `Recents` — so it already addresses the two blockers from the prior review round.

## 1. Spec compliance

**Spec: ✅**

- **Missing scope (漏建): none.** The stated goal — fix GH#484 *symptom 2* (non-bot DM conversations disappearing from a Space's sidebar) — is implemented. `decideConvKeepInSpace` now returns `true` for the `spaceID=="" && ChannelTypePerson && !botSet` branch (`modules/message/space_filter.go:300-310`), removing the unreliable Recents-window visibility gate. The two blockers from the previous round are both resolved at this head:
  - The stale v1 tests are updated to the new contract (`space_filter_test.go`: `..._DefaultSpaceBareConvs`, `..._NonDefaultSpace_AllNonBotDMsVisible`, `..._NewSpace_DMsAlwaysVisible`) — via honest rename + comment, not silent assertion edits.
  - The cross-Space `Recents` content exposure is closed by scrubbing DM `Recents` in the conversation-list path (`space_filter.go:213-215`).
- **Over-building (超建): none.** The added `Recents` scrub is not out-of-scope feature creep — it is a required consequence of the visibility widening (without it, the fix would re-open message-content leakage into the conversation-list preview). It is correctly gated to `filterSpaceID != "" && ChannelType==Person` and reuses the existing `filterPersonMessagesBySpace` rule, so untagged legacy messages (symptom 1) are untouched.
- **Deviation (偏离): none material.** The PR body still describes the fix as a "one-line change"; the head is broader (adds the scrub). This is stale PR prose, not a behavioral deviation — the runtime behavior matches issue #484 symptom 2, and symptom 1 (per-Space vs global DM semantics) is explicitly and correctly deferred to a product decision.

## 2. Code Quality

**Quality: Approved**

No P0/P1 findings. All remaining items are advisory (P2) or documentation nits.

### P2 — v2 sidebar surfaces global DM activity metadata (advisory, product-contract-dependent)

`decideConvKeepInSpace` is shared by two paths. The `Recents` scrub was added only to the v1 path (`filterConversationsCore` → `/v1/conversation/sync`, which ships `Recents` to the client). The v2 path (`FilterRawConversationsBySpace` → `/v1/sidebar/sync`, `space_filter.go:711-726`) does **not** scrub — and that asymmetry is **correct for message content**, because `SidebarItem` carries no `Recents`/payload field (verified: `api_sidebar.go:104-146`).

However, the v2 sidebar builders now copy the raw, globally-scoped `Timestamp`/`Unread` onto every newly-visible DM row (`api_sidebar.go:1161-1162` recent tab; the follow-tab path likewise). Since this PR widens which DMs appear in a non-active Space's sidebar, Space B can now show an unread badge / recency that was actually driven by Space A DM traffic. This is **not** a cross-user leak (the conversation list is `loginUID`-scoped — it's the user's own DM) and **not** a content leak (no message body). It's an activity-metadata / UX-accuracy question that hinges on the same "DMs global vs per-Space" product decision symptom 1 is deferred on.

Note the v1 path partially addresses this by computing a separate per-Space `SpaceUnread` (`space_unread.go` `fillPersonSpaceUnread`, run *before* the filter so raw Recents are intact); v2 sidebar has no equivalent. Recommend either documenting "DM row visible everywhere with global activity metadata" as the intended contract, or adding per-Space DM unread/timestamp to v2 — plus a regression test pinning whichever is chosen, so this isn't later reclassified as a privacy bug. Non-blocking under the current "DMs are global" interpretation.

### P2 — stale/dead helper wiring (nit)

- `hasSpaceMsg` (and `personConvHasSpaceMessages` / `rawConvHasSpaceMessages`) no longer governs regular DM visibility; the closures are still constructed per-iteration at `space_filter.go:198` and `:721`. The doc comment at `:230-234` explains this is intentionally retained pending the product decision — reasonable, but a `TODO(#484)` at the construction sites (or removing them if "global DMs" is final) would prevent a future reader treating it as an oversight. Minor per-iteration closure allocation on the list-build path; negligible.

### Correctness checks that passed

- **No data race / cache corruption from the in-place `conv.Recents` reassignment.** The v1 conversation objects are freshly built per request in `newSyncUserConversationResp` (fresh `recents` slice and fresh `*MsgSyncResp` elements, `api_conversation.go:1468-1536`), `IMSyncUserConversation` is a per-request backend call rather than a shared cache, and `filterPersonMessagesBySpace` returns a **new** slice (never mutates the backing array). So the mutation is confined to a request-local object.
- **Bot isolation preserved.** The `return true` sits after all bot-branch guards; a bot not in the target Space still falls through to `return false`. No cross-Space bot widening.
- **Unread ordering intact.** `fillPersonSpaceUnread` runs before `FilterConversationsBySpace` (`api_conversation.go:891-895`), so per-Space unread/preview computation still sees raw Recents.
- **Tests pass.** `go test ./modules/message -run 'TestFilterConversationsBySpace|TestDecideConvKeepInSpace'` → ok; `go vet ./modules/message` clean; package builds. The new `..._ScrubsCrossSpaceRecents` regression precisely pins the content-leak fix.

## 3. Overall verdict

**APPROVED** — Spec ✅ and Quality Approved. The disappearing-DM fix is correct, minimal, and well-tested; the previous round's two blockers are resolved at this head. The remaining items are advisory and tie into the already-deferred per-Space-vs-global DM product decision.

## 4. Suggested follow-ups (non-blocking)

- Decide and document whether DM sidebar rows should carry per-Space or global unread/timestamp in the v2 sidebar path, and add a regression test either way.
- Update the PR description (it still says "one-line change") to mention the Recents scrub, and add a `TODO(#484)` at the now-unused `hasSpaceMsg` construction sites.

## 5. Additional observations

- The v1 vs v2 scrub asymmetry is safe *today* only because `SidebarItem` ships no message payload. If a future change adds a last-message preview to `SidebarItem`, the v2 path will need the same scrub — worth a comment near `FilterRawConversationsBySpace` recording that invariant.

---

_Review used two independent advisory analyses (a line-level adversarial pass and a whole-repo reasoning pass) in addition to the primary review; both completed successfully. All findings above were independently re-verified against the diff before inclusion — including downgrading a flagged "in-place mutation / cache race" concern to a non-issue after confirming the conversation objects are request-local._

PR #496, opened 2026-06-29
feat(incomingwebhook): render Multica issue link + assignee (multi-line)

### LINKED ISSUE #497: feat(incomingwebhook): render Multica issue link + assignee in /multica notifications
## Background

The `/multica` incoming-webhook adapter (added in #426) renders an `issue.status_changed` notification as a single line:

```
**MUL-123** Fix login: `todo` → `in_progress` (by agent)
```

It cannot render a clickable issue link or a human-readable assignee, because the webhook payload Multica sent carried neither (no deployment domain / workspace slug, and only assignee UUIDs).

## Goal

Once Multica enriches its outbound payload with `issue_url` + `assignee_type` + `assignee_name` (Multica-side change), render a structured multi-line notification:

```
**[MUL-123 Fix login redirect](https://app.multica.ai/acme/issues/MUL-123)**
状态: `todo` → `in_progress`
指派: 张三 (member) · 触发: agent
```

## Scope (octo-server side)

- Extend `muEnvelope` with the three new (additive, backward-compatible) fields.
- Multi-line render: linked title / status line / assignee·trigger line, with each line omitted when its fields are absent.
- Security: the link **destination** must be validated (not just a scheme prefix) to prevent markdown link-injection, and the link **text** must use an escaper that neutralizes `*`/backtick (CommonMark parses emphasis/code spans inside link text).

Implemented in #496.

lml2468:
Closes #497

## Summary

The Multica adapter now consumes the `issue_url` + `assignee_name` fields Multica enriches its outbound webhook with, and renders a **structured multi-line** notification instead of the previous single line:

```
**[MUL-123 Fix login redirect](https://app.multica.ai/acme/issues/MUL-123)**
状态: `todo` → `in_progress`
指派: 张三 · 触发: agent
```

This is the **octo-server half** of a two-repo change; the Multica half is [Mininglamp-OSS/multica#49](https://github.com/Mininglamp-OSS/multica/pull/49).

## Changes
- `muEnvelope` gains `issue_url` / `assignee_type` / `assignee_name` (whitelist parse).
- `renderMulticaIssueStatusChanged`:
  - **Title line** links to `issue_url` when present, guarded by `isHTTPURL` (a `javascript:` / `data:` scheme downgrades to a plain-text title); link text via `mdLinkText`.
  - **Status line** on its own line.
  - **指派 · 触发 line** omitted entirely when both assignee and actor are empty.
  - Escaping contract unchanged: `mdLinkText` for link text, `mdInertText` for plain-text fields, `mdCodeSpanText` inside code spans.
- **Additive + backward compatible**: an older Multica that doesn't send the new fields renders the no-link / no-assignee form.

## Verification
- `go build ./...`
- `go test ./modules/incomingwebhook/ -run 'TestParseMulticaPush|TestMulticaAdapter|TestPublicURLs'` — all pass (new cases: link+assignee, assignee-without-actor, malicious-url downgrade, no-url plain-text fallback). The MySQL-dependent `TestDisableByGroupNo` is a pre-existing local-only skip.

### DIFF
--- a/modules/incomingwebhook/adapter_multica.go
+++ b/modules/incomingwebhook/adapter_multica.go
@@ -138,6 +138,7 @@ func renderMulticaIssueStatusChanged(ev muEnvelope) string {
 	prev := strings.TrimSpace(ev.PreviousStatus)
 	issueURL := strings.TrimSpace(ev.IssueURL)
 	assignee := strings.TrimSpace(ev.AssigneeName)
+	assigneeType := strings.TrimSpace(ev.AssigneeType)
 	actorType := strings.TrimSpace(ev.Actor.Type)
 	// id 与 status 是渲染所必需的最小集；缺失任一无法生成可读消息，回退
 	// 让上层走 reason=content 而非凭空生造。
@@ -150,16 +151,23 @@ func renderMulticaIssueStatusChanged(ev muEnvelope) string {
 	var b strings.Builder
 
 	// --- 标题行 ---
-	// 链接文本是 "{identifier} {title}"（title 可空）。有合法 http(s) url 时整体
-	// 包成 markdown 链接（mdLinkText 防 `]` 闭合）；否则降级纯文本（identifier
-	// 加粗，mdInertText 防 `*`/反引号 注入）。url 过 isHTTPURL 白名单杜绝
-	// `javascript:` 等 scheme 注入。
-	if issueURL != "" && isHTTPURL(issueURL) {
+	// 链接文本是 "{identifier} {title}"（title 可空）。有合法 url 时整体包成
+	// markdown 链接；否则降级纯文本。
+	//   - 链接文本用 mdInertText（不是 mdLinkText）：CommonMark 会在链接文本
+	//     内部解析强调/代码 span,所以文本里的 `*`/反引号/`[]` 必须按 inert 处理,
+	//     否则一个含 `**` 的 title 注入真粗体、一个含反引号的 title 把后面 status
+	//     code-span 切坏（#496 yujiawei/OctoBoooot review）。mdInertText 转义
+	//     `*_[]<>|` 并剥反引号——转义后的 `\]` 是字面右括号,不会闭合链接,对链接
+	//     文本同样安全。
+	//   - 链接目标用 safeMarkdownURL:仅 isHTTPURL 不够,目标是独立注入上下文,
+	//     `https://ok/) [phish](https://evil` 会闭合链接再注入第二个链接
+	//     （#496 Jerry-Xin/OctoBoooot review）。校验失败则降级纯文本标题。
+	if safeURL, ok := safeMarkdownURL(issueURL); ok {
 		linkText := id
 		if title != "" {
 			linkText = id + " " + title
 		}
-		fmt.Fprintf(&b, "**[%s](%s)**", mdLinkText(linkText, 200), issueURL)
+		fmt.Fprintf(&b, "**[%s](%s)**", mdInertText(linkText, 200), safeURL)
 	} else {
 		fmt.Fprintf(&b, "**%s**", mdInertText(id, shortFieldMax))
 		if title != "" {
@@ -182,9 +190,15 @@ func renderMulticaIssueStatusChanged(ev muEnvelope) string {
 
 	// --- 指派 / 触发行（任一非空才渲染整行）---
 	// assignee_name 由 multica 富集发来（octo 只收到 UUID，无法自解析人名）。
+	// assignee_type（member/agent/squad）附在名字后,让群里能区分指派给「人」
+	// 还是「Agent/Squad」(#496 yujiawei review:type 之前解析了却没用)。
 	var meta []string
 	if assignee != "" {
-		meta = append(meta, "指派: "+mdInertText(assignee, shortFieldMax))
+		label := "指派: " + mdInertText(assignee, shortFieldMax)
+		if assigneeType != "" {
+			label += " (" + mdInertText(assigneeType, shortFieldMax) + ")"
+		}
+		meta = append(meta, label)
 	}
 	if actorType != "" {
 		meta = append(meta, "触发: "+mdInertText(actorType, shortFieldMax))
--- a/modules/incomingwebhook/adapter_multica.go
+++ b/modules/incomingwebhook/adapter_multica.go
@@ -45,11 +45,19 @@ type muActor struct {
 
 // muEnvelope 是 multica 出站 webhook 的固定信封（见 multica
 // server/internal/integrations/outwebhook/dispatcher.go outboundPayload）。
+//
+// issue_url / assignee_type / assignee_name 是 multica 侧富集字段（后端用
+// PublicURL+slug+identifier 拼好 url、join 出 assignee 显示名后发来）——octo
+// 侧无法从 UUID/workspace_id 自行推断,所以直接消费。三者都是【加性】的:
+// 旧版 multica 不发 → 留空,渲染降级（无链接 / 无指派行）。
 type muEnvelope struct {
 	Event          string  `json:"event"`
 	WorkspaceID    string  `json:"workspace_id"`
 	Actor          muActor `json:"actor"`
 	Issue          muIssue `json:"issue"`
+	IssueURL       string  `json:"issue_url"`
+	AssigneeType   string  `json:"assignee_type"`
+	AssigneeName   string  `json:"assignee_name"`
 	PreviousStatus string  `json:"previous_status"`
 }
 
@@ -113,43 +121,24 @@ func parseMulticaPush(_ http.Header, body []byte) (*pushPayloadReq, string, stri
 
 // renderMulticaIssueStatusChanged 渲染 issue.status_changed 事件。
 //
-// 输出形如（markdown，客户端会渲染）：
+// 输出形如（markdown，客户端会渲染），多行结构化：
 //
-//	**MUL-123** Fix login redirect: `todo` → `in_progress` (by agent)
+//	**[MUL-123 Fix login redirect](https://app.multica.ai/acme/issues/MUL-123)**
+//	状态: `todo` → `in_progress`
+//	指派: 张三 · 触发: agent
 //
-// 设计取舍：
-//   - 不拼 issue URL：envelope 当前不带 URL；服务端无法可靠推断 multica
-//     部署域名（self-hosted 路径多种）。需要可点链接时由 multica 后续在
-//     envelope 加 issue.url 字段，渲染层加链接即可，不破坏既有契约。
-//   - 标题与 identifier 之间用空格分隔，与 GitHub 适配器的 issue 渲染风格
-//     一致；状态用反引号包起来，避免下划线（in_progress）被 markdown 误
-//     当作斜体。
-//   - actor 名当前不在 envelope 里，只渲染 type（"by agent" / "by member"）；
-//     未显示触发者类型的情况下省略尾注。
-//   - title 在 `** identifier ** title:` 的「裸文本」上下文（注意：与 github
-//     适配器的 title 不同——github 把 title 包进 `[text](url)` 链接里，所以那边
-//     用只转 `\[]` 的 mdLinkText 是对的；这里 title 不进链接，必须用
-//     mdInertText 才能把 `*`/反引号/`<`/`|` 等元字符也防住，否则一个含反引号
-//     的 title 会让 content 里反引号总数变奇数、把后面的 status code-span 切坏，
-//     一个含 `**` 的 title 会注入真粗体（yujiawei review P1）。
-//   - 其余进 markdown 的字段按上下文走对应 helper（详见 adapter.go 注释）：
-//   - identifier 在 `**...**` 粗体里 → mdInertText（转义 `*` 防止 bold-break，
-//     转义 `[]` 防止 link injection）；
-//   - status / previous_status 在 “ `...` “ code span 里 → mdCodeSpanText
-//     （只剥反引号防止逃逸 code span；`_`/`*` 在 code span 内本就不被解释，
-//     不转义否则会回显 `\_` 破坏显示）；
-//   - actor.type 在 `(by ...)` 纯文本上下文 → mdInertText（全量转义）。
-//     这些字段按 multica 契约都是受控枚举/标识符，escape 是 defense-in-depth
-//     （PR #427 review by yujiawei / mochashanyao / Jerry-Xin）。
-//   - 字段长度统一钳到 64 rune：identifier/status/actor 在 multica 端是短
-//     标识符（identifier 形如 "MUL-12345"、status 是枚举、actor.type 是
-//     "member"/"agent"），64 足够覆盖正常上限并把异常长度截短防御。title
-//     仍用 200 rune（与 GitHub PR/issue 标题钳值对齐）。
+// 三行分别是「标题（带链接）/ 状态变更 / 指派·触发」；缺字段的行自动省略。
+// issue_url 与 assignee_name 是 multica 侧富集字段（octo 无法从 UUID/
+// workspace_id 自行推断域名、slug、人名），渲染层直接消费、并对 url 做
+// http(s) 白名单。各字段的转义/钳长策略见函数内逐行注释。
 func renderMulticaIssueStatusChanged(ev muEnvelope) string {
 	id := strings.TrimSpace(ev.Issue.Identifier)
 	title := strings.TrimSpace(ev.Issue.Title)
 	status := strings.TrimSpace(ev.Issue.Status)
 	prev := strings.TrimSpace(ev.PreviousStatus)
+	issueURL := strings.TrimSpace(ev.IssueURL)
+	assignee := strings.TrimSpace(ev.AssigneeName)
+	actorType := strings.TrimSpace(ev.Actor.Type)
 	// id 与 status 是渲染所必需的最小集；缺失任一无法生成可读消息，回退
 	// 让上层走 reason=content 而非凭空生造。
 	if id == "" || status == "" {
@@ -159,26 +148,51 @@ func renderMulticaIssueStatusChanged(ev muEnvelope) string {
 	const shortFieldMax = 64
 
 	var b strings.Builder
-	fmt.Fprintf(&b, "**%s**", mdInertText(id, shortFieldMax))
-	if title != "" {
-		// 钳到 200 rune 与 GitHub PR/issue 标题渲染保持一致；用 mdInertText 而
-		// 非 mdLinkText——title 在这里是裸文本，不在 `[text](url)` 里（详见上方
-		// 设计取舍）。
-		b.WriteString(" ")
-		b.WriteString(mdInertText(title, 200))
+
+	// --- 标题行 ---
+	// 链接文本是 "{identifier} {title}"（title 可空）。有合法 http(s) url 时整体
+	// 包成 markdown 链接（mdLinkText 防 `]` 闭合）；否则降级纯文本（identifier
+	// 加粗，mdInertText 防 `*`/反引号 注入）。url 过 isHTTPURL 白名单杜绝
+	// `javascript:` 等 scheme 注入。
+	if issueURL != "" && isHTTPURL(issueURL) {
+		linkText := id
+		if title != "" {
+			linkText = id + " " + title
+		}
+		fmt.Fprintf(&b, "**[%s](%s)**", mdLinkText(linkText, 200), issueURL)
+	} else {
+		fmt.Fprintf(&b, "**%s**", mdInertText(id, shortFieldMax))
+		if title != "" {
+			b.WriteString(" ")
+			b.WriteString(mdInertText(title, 200))
+		}
 	}
-	b.WriteString(": ")
+
+	// --- 状态行 ---
+	// 反引号包起来避免下划线（in_progress）被当斜体；prev 缺失或与 status 相同
+	// 时只显示当前状态，不画 "→ X" 暗示变化的尾巴。
+	b.WriteString("\n状态: ")
 	if prev != "" && prev != status {
 		fmt.Fprintf(&b, "`%s` → `%s`",
 			mdCodeSpanText(prev, shortFieldMax),
 			mdCodeSpanText(status, shortFieldMax))
 	} else {
-		// 新建 issue 或事件 payload 缺 previous_status 时只显示当前状态，
-		// 不画"→ X"那种暗示状态变化的尾巴。
 		fmt.Fprintf(&b, "`%s`", mdCodeSpanText(status, shortFieldMax))
 	}
-	if actorType := strings.TrimSpace(ev.Actor.Type); actorType != "" {
-		fmt.Fprintf(&b, " (by %s)", mdInertText(actorType, shortFieldMax))
+
+	// --- 指派 / 触发行（任一非空才渲染整行）---
+	// assignee_name 由 multica 富集发来（octo 只收到 UUID，无法自解析人名）。
+	var meta []string
+	if assignee != "" {
+		meta = append(meta, "指派: "+mdInertText(assignee, shortFieldMax))
 	}
+	if actorType != "" {
+		meta = append(meta, "触发: "+mdInertText(actorType, shortFieldMax))
+	}
+	if len(meta) > 0 {
+		b.WriteString("\n")
+		b.WriteString(strings.Join(meta, " · "))
+	}
+
 	return b.String()
 }
--- a/modules/incomingwebhook/adapter.go
+++ b/modules/incomingwebhook/adapter.go
@@ -21,7 +21,9 @@ package incomingwebhook
 import (
 	"encoding/json"
 	"net/http"
+	"net/url"
 	"strings"
+	"unicode"
 )
 
 // pushAdapter 描述一种推送形态。
@@ -145,6 +147,45 @@ func firstLine(s string) string {
 func isHTTPURL(s string) bool {
 	l := strings.ToLower(strings.TrimSpace(s))
 	return strings.HasPrefix(l, "http://") || strings.HasPrefix(l, "https://")
+}
+
+// safeMarkdownURL validates an externally-supplied URL for safe interpolation
+// into a markdown link DESTINATION `](url)`. isHTTPURL alone is insufficient:
+// the destination is a distinct injection context — a value like
+// `https://ok/) [phish](https://evil` passes a scheme prefix check yet closes
+// the intended link and injects a second attacker-controlled link into the
+// message (#496 review). CommonMark also forbids/encodes spaces and control
+// chars in a bare destination, and a literal `)` ends it.
+//
+// Returns (url, true) only when the value: parses via net/url, is http(s), is
+// host-bearing, and contains no ASCII space, control char, or any of the
+// destination-breaking metacharacters `< > ( ) [ ] " \`. Otherwise (",", false)
+// so the caller downgrades to a plain-text (non-link) title — same degrade
+// contract as a non-http scheme.
+func safeMarkdownURL(raw string) (string, bool) {
+	raw = strings.TrimSpace(raw)
+	if !isHTTPURL(raw) {
+		return "", false
+	}
+	u, err := url.Parse(raw)
+	if err != nil || u.Host == "" {
+		return "", false
+	}
+	for _, r := range raw {
+		if r > unicode.MaxASCII {
+			// Non-ASCII (e.g. an IDN host or a unicode path) can't appear in a
+			// bare destination unescaped; reject rather than risk mis-rendering.
+			return "", false
+		}
+		if unicode.IsSpace(r) || unicode.IsControl(r) {
+			return "", false
+		}
+		switch r {
+		case '<', '>', '(', ')', '[', ']', '"', '\\':
+			return "", false
+		}
+	}
+	return raw, true
 }
 
 // oneLine 把多行文本压成单行，避免标题 / 评论里的换行破坏 markdown 链接结构。
--- a/modules/incomingwebhook/adapter_multica_test.go
+++ b/modules/incomingwebhook/adapter_multica_test.go
@@ -60,8 +60,8 @@ func TestParseMulticaPush_IssueStatusChanged_WithLinkAndAssignee(t *testing.T) {
 	require.NotNil(t, req, "skip=%q invalid=%q", skip, invalid)
 	// 标题行渲染成 markdown 链接，链接文本是 "identifier title"。
 	assert.Contains(t, req.Content, "**[MUL-123 Fix login redirect](https://app.multica.ai/acme/issues/MUL-123)**")
-	// 指派与触发同在一行，用 " · " 连接。
-	assert.Contains(t, req.Content, "指派: 张三 · 触发: agent")
+	// 指派(带 type)与触发同在一行，用 " · " 连接。
+	assert.Contains(t, req.Content, "指派: 张三 (member) · 触发: agent")
 }
 
 func TestParseMulticaPush_AssigneeWithoutActor(t *testing.T) {
@@ -107,6 +107,68 @@ func TestParseMulticaPush_NoIssueURL_PlainTextTitle(t *testing.T) {
 	require.NotNil(t, req)
 	assert.Contains(t, req.Content, "**MUL-4** no link")
 	assert.NotContains(t, req.Content, "](", "no link without issue_url")
+}
+
+// TestParseMulticaPush_IssueURLDestinationInjection is the #496 blocker regression:
+// a URL that passes a naive http(s) prefix check but carries link-breaking
+// characters (`)`, space, etc.) must NOT be rendered as a link — it would close
+// the intended destination and inject a second attacker-controlled link.
+func TestParseMulticaPush_IssueURLDestinationInjection(t *testing.T) {
+	cases := []struct {
+		name string
+		url  string
+	}{
+		{"close_and_inject", "https://ok.com/) [phish](https://evil.com"},
+		{"trailing_paren", "https://ok.com/issue)"},
+		{"embedded_space", "https://ok.com/a b"},
+		{"newline", "https://ok.com/a\nb"},
+		{"angle_brackets", "https://ok.com/<script>"},
+		{"bracket", "https://ok.com/]injected"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			body, err := json.Marshal(map[string]any{
+				"event":           "issue.status_changed",
+				"issue":           map[string]any{"identifier": "MUL-1", "title": "hi", "status": "done"},
+				"issue_url":       tc.url,
+				"previous_status": "todo",
+			})
+			require.NoError(t, err)
+			req, _, _ := parseMulticaPush(http.Header{}, body)
+			require.NotNil(t, req)
+			// The malicious destination must never be rendered as a markdown link.
+			assert.NotContains(t, req.Content, "](", "unsafe url must not become a markdown link: %q", req.Content)
+			assert.NotContains(t, req.Content, "[phish]", "second injected link must not appear")
+			// Degrades to a plain-text bold identifier instead.
+			assert.Contains(t, req.Content, "**MUL-1**", "title falls back to bold plain text")
+		})
+	}
+}
+
+// TestParseMulticaPush_LinkTextInjection is the #496 second blocker: when the
+// title IS rendered inside a link (`[text](url)`), CommonMark still parses
+// emphasis/code spans in the text, so a title with `**` or a backtick must be
+// neutralized — the link-text path uses mdInertText (escapes `*`, strips
+// backtick), not mdLinkText (which only escapes `\[]`).
+func TestParseMulticaPush_LinkTextInjection(t *testing.T) {
+	body, err := json.Marshal(map[string]any{
+		"event":           "issue.status_changed",
+		"issue":           map[string]any{"identifier": "MUL-1", "title": "a **bold** `tick", "status": "in_progress"},
+		"issue_url":       "https://app.multica.ai/acme/issues/MUL-1",
+		"previous_status": "todo",
+	})
+	require.NoError(t, err)
+	req, _, _ := parseMulticaPush(http.Header{}, body)
+	require.NotNil(t, req)
+	// asterisks escaped → no injected bold inside the link text.
+	assert.Contains(t, req.Content, `\*\*bold\*\*`, "asterisks in link text must be escaped")
+	assert.NotContains(t, req.Content, "a **bold**", "raw bold must not survive in link text")
+	// Backticks remain balanced (the title's lone backtick is stripped), so the
+	// downstream status code-span is not corrupted.
+	assert.Equalf(t, 0, strings.Count(req.Content, "`")%2,
+		"backticks must remain balanced (got odd count): %q", req.Content)
+	// Still a valid single link to the safe destination.
+	assert.Contains(t, req.Content, "](https://app.multica.ai/acme/issues/MUL-1)")
 }
 
 func TestParseMulticaPush_IssueStatusChanged_NoPreviousStatus(t *testing.T) {
--- a/modules/incomingwebhook/adapter_multica_test.go
+++ b/modules/incomingwebhook/adapter_multica_test.go
@@ -29,12 +29,84 @@ func TestParseMulticaPush_IssueStatusChanged_FullEnvelope(t *testing.T) {
 	}`)
 	req, skip, invalid := parseMulticaPush(http.Header{}, body)
 	require.NotNil(t, req, "skip=%q invalid=%q", skip, invalid)
-	// 期望渲染：**MUL-123** Fix login redirect on mobile: `todo` → `in_progress` (by agent)
+	// 多行渲染：
+	//   **MUL-123** Fix login redirect on mobile
+	//   状态: `todo` → `in_progress`
+	//   触发: agent
 	assert.Contains(t, req.Content, "**MUL-123**")
 	assert.Contains(t, req.Content, "Fix login redirect on mobile")
-	assert.Contains(t, req.Content, "`todo` → `in_progress`")
-	assert.Contains(t, req.Content, "(by agent)")
+	assert.Contains(t, req.Content, "状态: `todo` → `in_progress`")
+	assert.Contains(t, req.Content, "触发: agent")
 	assert.Empty(t, req.MsgType, "adapters emit the plain-text path")
+}
+
+func TestParseMulticaPush_IssueStatusChanged_WithLinkAndAssignee(t *testing.T) {
+	// multica 富集 issue_url + assignee_name 时：标题成可点链接，多一行指派。
+	body := []byte(`{
+		"event": "issue.status_changed",
+		"workspace_id": "550e8400-e29b-41d4-a716-446655440000",
+		"actor": {"type": "agent", "id": "agent-7"},
+		"issue": {
+			"identifier": "MUL-123",
+			"title": "Fix login redirect",
+			"status": "in_progress"
+		},
+		"issue_url": "https://app.multica.ai/acme/issues/MUL-123",
+		"assignee_type": "member",
+		"assignee_name": "张三",
+		"previous_status": "todo"
+	}`)
+	req, skip, invalid := parseMulticaPush(http.Header{}, body)
+	require.NotNil(t, req, "skip=%q invalid=%q", skip, invalid)
+	// 标题行渲染成 markdown 链接，链接文本是 "identifier title"。
+	assert.Contains(t, req.Content, "**[MUL-123 Fix login redirect](https://app.multica.ai/acme/issues/MUL-123)**")
+	// 指派与触发同在一行，用 " · " 连接。
+	assert.Contains(t, req.Content, "指派: 张三 · 触发: agent")
+}
+
+func TestParseMulticaPush_AssigneeWithoutActor(t *testing.T) {
+	// 只有 assignee_name、没有 actor.type：指派行只渲染指派段，不带 " · 触发"。
+	body := []byte(`{
+		"event": "issue.status_changed",
+		"issue": {"identifier": "MUL-5", "title": "x", "status": "done"},
+		"assignee_type": "agent",
+		"assignee_name": "Codex Bot",
+		"previous_status": "todo"
+	}`)
+	req, _, _ := parseMulticaPush(http.Header{}, body)
+	require.NotNil(t, req)
+	assert.Contains(t, req.Content, "指派: Codex Bot")
+	assert.NotContains(t, req.Content, "·", "no separator when actor is absent")
+	assert.NotContains(t, req.Content, "触发:")
+}
+
+func TestParseMulticaPush_MaliciousIssueURLDowngraded(t *testing.T) {
+	// issue_url 必须过 http(s) 白名单：javascript: 等 scheme 不能渲染成链接，
+	// 标题降级为纯文本（identifier 加粗），且恶意 url 不出现在 content 里。
+	body := []byte(`{
+		"event": "issue.status_changed",
+		"issue": {"identifier": "MUL-9", "title": "evil", "status": "done"},
+		"issue_url": "javascript:alert(1)",
+		"previous_status": "todo"
+	}`)
+	req, _, _ := parseMulticaPush(http.Header{}, body)
+	require.NotNil(t, req)
+	assert.NotContains(t, req.Content, "javascript:", "non-http(s) url must not be rendered")
+	assert.NotContains(t, req.Content, "](", "url must not become a markdown link")
+	assert.Contains(t, req.Content, "**MUL-9**", "title falls back to bold plain text")
+}
+
+func TestParseMulticaPush_NoIssueURL_PlainTextTitle(t *testing.T) {
+	// 无 issue_url（旧版 multica）：标题回退纯文本 `**id** title`，不是链接。
+	body := []byte(`{
+		"event": "issue.status_changed",
+		"issue": {"identifier": "MUL-4", "title": "no link", "status": "done"},
+		"previous_status": "todo"
+	}`)
+	req, _, _ := parseMulticaPush(http.Header{}, body)
+	require.NotNil(t, req)
+	assert.Contains(t, req.Content, "**MUL-4** no link")
+	assert.NotContains(t, req.Content, "](", "no link without issue_url")
 }
 
 func TestParseMulticaPush_IssueStatusChanged_NoPreviousStatus(t *testing.T) {
@@ -50,11 +122,11 @@ func TestParseMulticaPush_IssueStatusChanged_NoPreviousStatus(t *testing.T) {
 	assert.Contains(t, req.Content, "First issue")
 	assert.Contains(t, req.Content, "`todo`")
 	assert.NotContains(t, req.Content, "→", "no arrow when previous_status is absent")
-	assert.Contains(t, req.Content, "(by member)")
+	assert.Contains(t, req.Content, "触发: member")
 }
 
 func TestParseMulticaPush_NoActorType(t *testing.T) {
-	// actor.type 缺失：不渲染 "(by …)" 尾巴。
+	// actor.type 缺失且无 assignee：整个指派/触发行省略。
 	body := []byte(`{
 		"event": "issue.status_changed",
 		"issue": {"identifier": "MUL-3", "title": "no actor", "status": "done"},
@@ -62,7 +134,8 @@ func TestParseMulticaPush_NoActorType(t *testing.T) {
 	}`)
 	req, _, _ := parseMulticaPush(http.Header{}, body)
 	require.NotNil(t, req)
-	assert.NotContains(t, req.Content, "(by")
+	assert.NotContains(t, req.Content, "触发:")
+	assert.NotContains(t, req.Content, "指派:")
 }
 
 func TestParseMulticaPush_TitleEscaping(t *testing.T) {
--- a/modules/incomingwebhook/adapter_test.go
+++ b/modules/incomingwebhook/adapter_test.go
@@ -60,3 +60,35 @@ func TestFirstLineAndOneLine(t *testing.T) {
 	assert.Equal(t, "no newline", firstLine("no newline"))
 	assert.Equal(t, "a b c", oneLine("a\r\nb\nc"))
 }
+
+func TestSafeMarkdownURL(t *testing.T) {
+	ok := []string{
+		"https://app.multica.ai/acme/issues/MUL-1",
+		"http://localhost:3000/x/issues/Y-2",
+		"https://app.multica.ai/acme%20corp/issues/MUL%207", // pre-escaped segments
+	}
+	for _, u := range ok {
+		got, valid := safeMarkdownURL(u)
+		assert.Truef(t, valid, "%q should be accepted", u)
+		assert.Equal(t, u, got)
+	}
+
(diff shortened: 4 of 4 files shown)


Jerry-Xin (changes requested):
Summary: The PR is relevant to `octo-server` and the Multica adapter direction is sound, but the new linked title path reintroduces markdown link-destination injection.

🔴 Blocking

🔴 Critical — `issue_url` is inserted directly into a markdown link after only a prefix check. In [modules/incomingwebhook/adapter_multica.go:157](modules/incomingwebhook/adapter_multica.go:157)-[162](modules/incomingwebhook/adapter_multica.go:162), `isHTTPURL` only checks `http://` / `https://` prefixes, so a payload value like `https://safe/) [phish](https://evil)` still passes and can close the intended link destination, then inject arbitrary markdown into the group message. Incoming webhook payloads are bearer-token authenticated but still externally supplied, and this adapter already treats titles, identifiers, actor types, and schemes defensively. The URL destination needs comparable handling: parse/validate with `net/url`, reject control characters and whitespace, and either reject/encode markdown-breaking destination characters such as `)` or add a dedicated safe markdown URL helper. Please add a regression test for a `https://... ) ...` or newline-containing `issue_url`.

💬 Non-blocking

No additional non-blocking issues found.

✅ Highlights

The multi-line rendering is backward compatible for older Multica payloads, and the new tests cover the main happy paths, missing actor/assignee combinations, malicious non-http schemes, and plain-text fallback. I also ran `go test ./modules/incomingwebhook/ -run 'TestParseMulticaPush|TestMulticaAdapter|TestPublicURLs'`, and it passed.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#496 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/496
**Head SHA:** `6392408a817c975ad22667f4078af640815cc03d`
**Scope:** `modules/incomingwebhook/adapter_multica.go` (+56/-42), `adapter_multica_test.go` (+79/-6)
**Build:** ✅ `go build ./...` clean
**Tests:** ✅ All 20 `TestParseMulticaPush*` + `TestMulticaAdapter` + `TestPublicURLs` pass

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Build clean | ✅ | `go build ./...` exit 0 |
| Tests pass | ✅ | 20/20 Octo-Q tests PASS |
| Link rendering (new path) | ✅ | `adapter_multica.go:155-162`, test `WithLinkAndAssignee` |
| No-link fallback (backward compat) | ✅ | `adapter_multica.go:163-168`, test `NoIssueURL_PlainTextTitle` |
| `javascript:` scheme rejection | ✅ | `isHTTPURL` prefix check, test `MaliciousIssueURLDowngraded` |
| Assignee/actor meta-line | ✅ | `adapter_multica.go:181-190`, tests `AssigneeWithoutActor`, `NoActorType` |
| Title escaping (plain-text path) | ✅ | `mdInertText(title, 200)` at L167, existing tests `TitleEscaping`, `TitleInjectionVectors` |
| Title escaping (link path) | ✅ | `mdLinkText(linkText, 200)` at L161 — escapes `\`, `[`, `]` |
| Status in code span | ✅ | `mdCodeSpanText` strips backticks, test `StatusBacktickStripping` |

---

## 2. Findings

### F1 — `issueURL` interpolated into markdown link destination without sanitization (P2)

**Location:** `adapter_multica.go:162`
```go
fmt.Fprintf(&b, "**[%s](%s)**", mdLinkText(linkText, 200), issueURL)
```

**Issue:** `issueURL` passes `isHTTPURL` (scheme prefix check: `http://` or `https://`) but is then interpolated **raw** into `](%s)`. Characters that are syntactically meaningful in CommonMark link destinations — `)`, unescaped whitespace, `\n` — are not stripped or escaped. A crafted URL like `https://evil.com/foo) [phish](https://evil.com` would pass `isHTTPURL` and break the markdown structure:

```
**[MUL-123 title](https://evil.com/foo) [phish](https://evil.com)**
```

This would render a link to `https://evil.com/foo`, followed by a space, followed by a second markdown link `[phish](https://evil.com)` wrapped in bold.

**Diff-scope:** **new** — this code path didn't exist before this PR. However, the **pattern** (trusted URL interpolated raw into markdown link destination) is **pre-existing** in the GitHub adapter (`adapter_github.go:242,258,272,293,313`), GitLab adapter (`adapter_gitlab.go:241,269,284,304,310,313`), and Wecom adapter (`adapter_wecom.go:114`) — all interpolate `html_url`/`WebURL`/commit URL directly without sanitization.

**Severity rationale (P2, not P1):**
- The webhook source is authenticated via 128-bit URL token — only a legitimate Octo-Q instance can send payloads.
- `issue_url` is constructed server-side by the Octo-Q platform (`PublicURL + slug + identifier`), not from arbitrary user input.
- No user-visible error occurs in the normal case; the attack requires a compromised or malicious Octo-Q instance.
- Consistent with all existing adapter patterns in the codebase.
- **Does not meet R1 threshold:** does not make a working path unavailable, does not produce incorrect data in normal operation, is not worse than before.

**Suggestion:** Add a `sanitizeURL` helper that strips `)`, `\n`, `\r`, and whitespace from URL strings before interpolation, or use angle-bracket link syntax `[text](<url>)` which handles more edge cases. Low priority, consistent defense-in-depth hardening.

---

### F2 — `AssigneeType` parsed but never consumed (P2/nit)

**Location:** `adapter_multica.go:56` (struct field), consumed nowhere in `renderMulticaIssueStatusChanged`.

**Issue:** `AssigneeType string \`json:"assignee_type"\`` is added to `muEnvelope` and parsed from the JSON payload, but the rendering function only uses `AssigneeName` and `Actor.Type`. The `assignee_type` value is dead data.

**Diff-scope:** new (this PR).

**Severity:** nit/maintainability. No functional impact. May be intentionally reserved for future use (e.g., distinguishing "agent" vs "member" assignees in rendering). Not blocking.

---

## 3. Suggestions

1. **F1 hardening (P2):** Consider adding URL sanitization for the link destination. At minimum, strip `)`, `\n`, `\r` and control characters from `issueURL` before interpolation. This would be a net improvement over the existing pattern across all adapters.

2. **F2 cleanup (nit):** Either remove `AssigneeType` from `muEnvelope` if it's not planned for use, or add a brief comment explaining its intended future purpose.

---

## 4. Additional Observations

- **Backward compatibility is well-handled:** The three new envelope fields (`issue_url`, `assignee_type`, `assignee_name`) are all additive with proper zero-value fallback. Old Octo-Q versions that don't send these fields get the existing plain-text rendering. ✅
- **Test coverage is thorough:** 5 new test functions covering link+assignee, assignee-without-actor, malicious URL downgrade, no-URL fallback. All existing tests updated to match new format. ✅
- **Existing security tests remain valid:** `TitleInjectionVectors`, `IdentifierMarkdownEscaping`, `StatusBacktickStripping`, `ActorTypeMarkdownEscaping` all still pass with the new rendering logic. ✅

---

## 5. Data Flow Tracing

| Consumed Data | Upstream Source | Flow Verified |
|---|---|---|
| `ev.IssueURL` | Octo-Q webhook JSON → `json.Unmarshal` → `muEnvelope.IssueURL` | ✅ `TrimSpace` at L138, `isHTTPURL` gate at L155 |
| `ev.AssigneeName` | Octo-Q webhook JSON → `muEnvelope.AssigneeName` | ✅ `TrimSpace` at L139, `mdInertText` at L184 |
| `ev.Actor.Type` | Octo-Q webhook JSON → `muEnvelope.Actor.Type` | ✅ `TrimSpace` at L140, `mdInertText` at L187 |
| `ev.Issue.Identifier` | Unchanged from pre-PR | ✅ `mdInertText` / `mdLinkText` depending on path |
| `ev.Issue.Title` | Unchanged from pre-PR | ✅ `mdInertText` (no-link) / `mdLinkText` (link) |
| `ev.Issue.Status` / `ev.PreviousStatus` | Unchanged from pre-PR | ✅ `mdCodeSpanText` inside backtick code span |

---

- **C1 — Dual-path parity:** ✅ CLEAR. Link/no-link paths are symmetric with matching tests. Assignee-present/absent and actor-present/absent paths all covered. No create/delete or subscribe/unsubscribe symmetry concerns.
- **C2 — Control-flow ordering / nesting:** ✅ CLEAR. Single-use rendering function, no nesting. Escaping helpers applied consistently per context (`mdLinkText` for link text, `mdInertText` for plain text, `mdCodeSpanText` for code spans).
- **C3 — Authorization boundary ≠ capability boundary:** N/A — no new endpoints, tools, or capability exposure. Existing webhook authentication (URL token) unchanged.
- **C4 — Authorization lifecycle / container-member cascade:** N/A — no auth changes.
- **C5 — Build ≠ runtime path:** N/A — this is a Go server binary, not a browser extension or CLI tool. Build output is the runtime artifact.
- **C6 — Governance / policy doc self-consistency:** N/A — no documentation or policy changes.

---

---

[Octo-Q] verdict: **APPROVE**

All findings are P2 (defense-in-depth hardening / nit). No P0/P1 identified. The PR correctly implements multi-line rendering with proper escaping for all user-influenced fields, maintains backward compatibility with old Octo-Q versions, and has thorough test coverage including security regression tests.

modules/incomingwebhook/adapter_multica.go
```diff
@@ -159,26 +148,51 @@ func renderMulticaIssueStatusChanged(ev muEnvelope) string {
...
+	// 链接文本是 "{identifier} {title}"（title 可空）。有合法 http(s) url 时整体
+	// 包成 markdown 链接（mdLinkText 防 `]` 闭合）；否则降级纯文本（identifier
+	// 加粗，mdInertText 防 `*`/反引号 注入）。url 过 isHTTPURL 白名单杜绝
+	// `javascript:` 等 scheme 注入。
+	if issueURL != "" && isHTTPURL(issueURL) {
+		linkText := id
+		if title != "" {
+			linkText = id + " " + title
+		}
+		fmt.Fprintf(&b, "**[%s](%s)**", mdLinkText(linkText, 200), issueURL)
```
OctoBoooot: **[blocker]** adapter_multica.go:162 — `issueURL` is interpolated raw into the markdown link destination `](%s)`; `isHTTPURL` only checks the `http(s)://` *prefix*, not the rest, so a `)` / space / `]` in the URL breaks out and injects attacker-controlled markdown.
Failing input — `issue_url = "https://x.com) injected [evil](https://evil.com"` renders:
`**[MUL-1 hi](https://x.com) injected [evil](https://evil.com)**` — the early `)` closes the intended link and `[evil](https://evil.com)` becomes a *second clickable link* (phishing) in the notification. Also reproduced: injecting `**bold**`/`` `code` `` and smuggling `](javascript:...)`. Every other field here is escaped per-context as defense-in-depth; the link *text* uses `mdLinkText`, but the link *destination* is the one unescaped context.
The webhook signature isn't verified (only the URL token), and the surrounding code already treats multica fields as untrusted — so the field is attacker-influenceable. Fix: validate with `net/url.Parse` and reject if the result re-encodes differently, or reject `issueURL` containing any of `) ( whitespace < > ]` (a legit http URL has those percent-encoded), before interpolating. Smallest safe form: `if u,err:=url.Parse(issueURL); err!=nil || (u.Scheme!="http"&&u.Scheme!="https") || strings.ContainsAny(issueURL,"() <>") { /* downgrade to plain text */ }`.

OctoBoooot (changes requested):
## Review: feat(incomingwebhook): render Multica issue link + assignee (multi-line) (#496)

**Verdict:** Request changes — one blocker: the new `issue_url` is interpolated into the markdown link destination without escaping, allowing link/markdown injection. CI is also red on `check-sprint` (admin gate); Build/Vet/Lint pass, Test pending.

Risk tier: normal by path, but this is the escaping-sensitive class (#330/#423/#427 lineage — external text into markdown contexts). Established author (52 merged). Two files.

### Blocker

- `adapter_multica.go:162` — see inline. `issueURL` goes raw into `**[%s](%s)**`; `isHTTPURL` validates only the `http(s)://` prefix, so a `)`/space/`]` in the URL escapes the destination and injects markdown. Byte-verified: `issue_url = "https://x.com) injected [evil](https://evil.com"` →
  ```
  **[MUL-1 hi](https://x.com) injected [evil](https://evil.com)**
  ```
  a second attacker-controlled clickable link (phishing) inside the notification. Also reproduced `**bold**`/code-span injection and `](javascript:...)` smuggling. The signature header (`X-Multica-Signature-256`) is *not* verified (only the URL token gates the path), and the code already treats every other multica field as untrusted — so this field is attacker-influenceable, and the destination is the one context the escape contract misses.

### Praise

- **The link-*text* side of the audit is done correctly** — the title/identifier path branches on `isHTTPURL` and uses `mdLinkText` (escapes `[`/`]`/`\`) for the link text, `mdInertText` for the plain-text fallback, `mdCodeSpanText` inside the status code spans, and `mdInertText` for assignee/actor. Every field except the URL destination is escaped per its exact context — the `javascript:`/`data:` scheme downgrade is there too. The single gap is the destination, which is a harder context to remember precisely because it isn't text.
- **The scheme-downgrade and the additive/backward-compatible envelope design are right** — old Multica without the new fields degrades cleanly to the no-link form, and the empty-assignee-and-actor line is omitted rather than rendered blank.

### What would clear the blocker

Validate the URL structurally before interpolating, not just by prefix: `net/url.Parse` + scheme check + reject if it contains `()`/whitespace/`<>`/`]` (a legitimate http URL percent-encodes those). On failure, fall through to the existing plain-text branch you already wrote. Add a test case with a `)`-bearing URL asserting the output stays a single link (the diff's "malicious-url downgrade" test only covers the `javascript:` scheme, not destination-breakout — that gap is why this slipped).

- CI `check-sprint` — admin Sprint-field gate, same as the rest of this cycle. (Independent of the blocker.)

yujiawei (changes requested):
# Code Review — PR #496 (octo-server)

**Head SHA reviewed:** `6392408a817c975ad22667f4078af640815cc03d` · base `origin/main` (`8df3685e`)
**Scope:** `modules/incomingwebhook/adapter_multica.go` (+ test). Build and the targeted test suite both pass locally (`go build ./...`, `go vet`, `go test ./modules/incomingwebhook/`).

The feature itself — multi-line render, link title, assignee line, scheme-allowlist downgrade — is well-shaped and the no-link backward-compat path is correct. Two issues block merge, both centered on the new linked-title branch.

---

## 1. Spec compliance

The core spec (link title + status line + assignee/actor line, additive & backward-compatible) is implemented as described, and the no-`issue_url` and non-`http(s)` downgrade paths behave as claimed. One minor over-build:

- **`assignee_type` is parsed but never used.** `adapter_multica.go:59` adds `AssigneeType` to the envelope and the struct comment says the three new fields are "directly consumed", but `assignee_type` is read nowhere in `renderMulticaIssueStatusChanged` — only `assignee_name` and `actor.type` drive output. `assignee_type="agent"` vs `"member"` render identically. Either drop it from the whitelist or use it (e.g. to distinguish 指派 of a bot vs a person). Non-blocking, but the comment is currently misleading.

**Spec: ✅ (with the `assignee_type` over-build noted above)**

## 2. Code quality

### P1 — Markdown injection regression: link text is no longer inert (`adapter_multica.go:157-162`)

When `issue_url` is present the title switches from `mdInertText` (strips backticks, escapes `*`/`_`/`<`/`>`/`|`) to `mdLinkText`, which **only** escapes `\ [ ]`. Backticks and asterisks in the title now pass through unescaped into the rendered message. This re-opens exactly the class of bug the previous revision fixed — the now-deleted comment at the old line 132-134 explicitly called out "a title containing a backtick makes the backtick count odd and breaks the following status code-span / a title containing `**` injects real bold". The link branch reintroduces it.

This is triggerable by an ordinary, well-formed title — no attacker needed:

```
issue: { identifier:"MUL-1", title:"fix `code bug", status:"in_progress" }, issue_url:"https://app.multica.ai/acme/issues/MUL-1", previous_status:"todo"
```
renders
```
**[MUL-1 fix `code bug](https://app.multica.ai/acme/issues/MUL-1)**\n状态: `todo` → `in_progress`
```
The single backtick in the title is odd-count; the CommonMark code-span scanner pairs it with the first backtick on the status line, swallowing the link close and the status text into a broken code span. An `**login**` in the title similarly corrupts the surrounding bold. Verified against the actual code path.

Fix: keep the title inert even inside a link — escape backticks/`*`/`_` in the link text (the link only needs `[ ]` escaped *additionally*, not *instead*). A combined "link-text + inert" escaper, or running `mdInertText`-style escaping then `[ ]` escaping, restores the prior guarantee.

### P1 — Link destination is interpolated raw (`adapter_multica.go:162`, gate `adapter.go:145`)

```go
fmt.Fprintf(&b, "**[%s](%s)**", mdLinkText(linkText, 200), issueURL)
```
The link **text** is escaped, but the **destination** `issueURL` goes in raw — the only gate is `isHTTPURL`, a scheme-prefix check; `issueURL` otherwise gets just `TrimSpace`. Under CommonMark a non-`<>`-wrapped destination ends at the first unescaped `)` or whitespace, so a value that passes the scheme check can still break out:

```
issue_url:"https://app.multica.ai/x) ![pwned](https://evil.example/x.png"
→ **[MUL-1 t](https://app.multica.ai/x) ![pwned](https://evil.example/x.png)**
```
The first `)` closes the real link; the remainder renders as an attacker-controlled image (remote fetch / tracking pixel / client-IP leak on render) or, in the `[text](url)` variant, an arbitrary phishing link with attacker-chosen anchor text. The endpoint is authenticated by the URL token only — `X-Multica-Signature-256` is not verified — so a token-holder controls the full envelope, and the producer-side path-escaping is not a security boundary for this receiver.

Honest scoping: this raw-destination pattern is **module-wide** — the github/gitlab/wecom/feishu adapters interpolate their URL destinations the same way, and no URL-destination escaper exists today. So this is a pre-existing gap, not unique to this PR. But this adapter is **new code carrying an explicit "escape every field for its markdown context" posture**, and the fix lives at the same line you're already touching. Please harden it here: wrap the destination in `<...>` after escaping, or downgrade to plain text when the URL contains `)` / whitespace / control chars (mirroring the existing `isHTTPURL` downgrade). A shared `mdLinkDest` helper applied across all adapters would be the ideal follow-up.

### P2 — Test names overstate coverage (`adapter_multica_test.go:83-97`)

`TestParseMulticaPush_MaliciousIssueURLDowngraded` only feeds `javascript:alert(1)` — a *scheme* rejection that takes the plain-text else-branch, so its `assert.NotContains("](")` passes trivially without ever exercising the link branch. There is no test feeding an `http(s)` URL containing `)`/space/newline, and no test feeding a title with a backtick/`**` while `issue_url` is present. Both P1s above ship green because the negative cases that would catch them are missing. Add them alongside the fixes.

### P2 — Clip-length inconsistency in the link branch (`adapter_multica.go:158-162`)

In the plain-text branch the identifier is clamped to `shortFieldMax` (64) and the title keeps its own 200-rune budget. In the link branch `linkText = id + " " + title` is clamped to 200 as a whole, so a pathologically long identifier can consume the budget and drop the title entirely. Clamp `id` to 64 before composing `linkText` for parity.

### P2 — Multi-line layout depends on `\n` rendering as a line break

The three logical lines are joined with bare `\n`. In strict CommonMark a single `\n` is a soft break (collapses to a space) unless the client treats newlines as hard breaks. If the IM markdown renderer is strict CommonMark, "状态:" and "指派:" will run onto the title line. Worth a quick confirmation against the actual client renderer; if it's strict, use a hard break (two trailing spaces or a blank line).

**Quality: Changes-Requested** (two P1s in the new link branch)

## 3. Overall verdict

**CHANGES_REQUESTED.** The feature is close, but the new linked-title branch (a) regresses the documented backtick/`**` inert-title guarantee for ordinary titles and (b) interpolates the link destination raw under a token-only/no-signature threat model — both in code this PR adds, both with no negative test. Re-escape the link text to stay inert, escape/validate the destination at the same sink, add the two missing negative tests, and either use or drop `assignee_type`.

OctoBoooot (commented):
Concurring with @yujiawei's 🔴 on the title escaping — this is a SECOND injection vector, distinct from the URL-destination one I filed, and I missed it. Byte-verified at 6392408a:

When the title moved into the new `**[linkText](url)**` branch (`adapter_multica.go:162`), it switched from `mdInertText` → `mdLinkText`. Their escaper sets differ:
- `mdInertText`: escapes `* _ < > |`, strips backtick
- `mdLinkText`: escapes only `\ [ ]`

So backtick and `*` are no longer escaped in the title:
- `title = "a **bold** b"` → `**[MUL-2 a **bold** b](https://ok.com)**` — injected real bold.
- `title = "break`tick"` (lone backtick) → odd backtick count in the document → corrupts the downstream `状态: \`...\`` code span.

The kicker: the PR DELETED the exact design comment that warned about this — "一个含反引号的 title 会让 content 里反引号总数变奇数、把后面的 status code-span 切坏，一个含 `**` 的 title 会注入真粗体" — the rationale for using `mdInertText` on the title in the first place. The refactor dropped the warning and the protection together.

So there are **two** vectors to fix, not one: (1) the URL destination (my blocker), and (2) link-text needs an escaper that also covers backtick + `*` (CommonMark interprets emphasis and code spans inside link text). A fix that only validates the URL would still ship this. yujiawei's 🟡 (unused `assignee_type`) also holds — dead field, fold into the same commit. My REQUEST_CHANGES stands; this strengthens it.

lml2468:
Pushed `eec9cfa8` addressing both blockers + the unused-field note (thanks @Jerry-Xin @yujiawei @OctoBoooot):

1. **🔴 URL-destination injection** — added `safeMarkdownURL` (net/url parse + http(s) + host required + rejects ASCII space, control chars, and the destination-breaking metachars `< > ( ) [ ] " \` plus non-ASCII). The render path uses it for the `](url)` destination and downgrades to a plain-text title when it fails. Regression suite `TestParseMulticaPush_IssueURLDestinationInjection` covers close-and-inject, trailing paren, space, newline, angle brackets, bracket; `TestSafeMarkdownURL` is a direct accept/reject table.
2. **🔴 link-text injection** — the link **text** now uses `mdInertText` (escapes `* _ < > |`, strips backtick) instead of `mdLinkText` (which only escapes `\ [ ]`), restoring the protection the refactor had dropped. `TestParseMulticaPush_LinkTextInjection` asserts `**` is escaped and backticks stay balanced.
3. **🟡 unused `assignee_type`** — now rendered: the 指派 line shows `指派: <name> (<type>)` so a bot vs person assignment is distinguishable.

`go build ./...` and `go test ./modules/incomingwebhook/` (all `TestParseMulticaPush*` + `TestSafeMarkdownURL` + helpers) green.

OctoBoooot (approved):
## Review: feat(incomingwebhook): render Multica issue link + assignee (multi-line) (#496) — delta @ eec9cfa8

**Verdict:** Approve — both injection vectors from the prior round are closed, byte-verified, with comprehensive regression tests. No blockers, no majors. CI red on `check-sprint` only (admin gate); Build/Test/Vet pending (the new injection tests pass locally).

Delta from prior REQUEST_CHANGES @ `6392408a`: 1 commit — "fix: close markdown injection in Multica link; use assignee_type". Addresses the blocker I filed, the second vector yujiawei caught, and the unused-field nit.

### Resolved — byte-verified

- ✓ **URL-destination injection closed** (my blocker) — new `safeMarkdownURL` (`adapter.go`) parses via `net/url`, requires http(s) + non-empty host, and rejects any ASCII space / control char / `< > ( ) [ ] " \`. Verified the prior failing inputs now downgrade to plain text (no link): `https://x.com) injected [evil](...)` → `**MUL-1** hi`, plus the `](javascript:...)`, embedded-space, and non-ASCII-IDN cases. A legitimate `https://ok.com/path?q=1` correctly stays a link. The downgrade reuses the existing plain-text branch.
- ✓ **Link-text backtick/asterisk injection closed** (yujiawei's second vector) — the link text switched from `mdLinkText` (escaped only `\ [ ]`) to `mdInertText` (escapes `* _ [ ] < > |`, strips backtick). Verified: `a **bold** b` → `a \*\*bold\*\* b` (no injected bold), `` x `code` y `` → `x code y`, lone-backtick `break\`tick` → `breaktick` (so the downstream `状态:` code span can't be corrupted). Using `mdInertText` (not just widening `mdLinkText`) is the conservative choice — it still escapes `]`→`\]`, so the original link-closing concern stays covered too.
- ✓ **`assignee_type` now consumed** (yujiawei nit) — renders `指派: 张三 (member)`, escaped via `mdInertText`. No longer a dead parsed field.
- ✓ **Regression tests pin both vectors** — `TestParseMulticaPush_IssueURLDestinationInjection` (close-and-inject, trailing-paren, embedded-space, newline, angle-brackets, bracket) and `TestParseMulticaPush_LinkTextInjection` (asserts `\*\*bold\*\*` escaped + balanced backtick count). Both pass; the package-level failure is the pre-existing MySQL-dependent `TestDisableByGroupNo` (local-only skip, unrelated).

### Praise

- **The fix treats the link destination as its own escape context with a dedicated validator, not a patched prefix check.** `safeMarkdownURL` enumerates exactly the destination-breaking characters and rejects rather than trying to escape them — the right call for a URL, where escaping is ambiguous but rejection + plain-text downgrade is safe and lossless for legitimate inputs. The doc comment names the failing input and the threat.
- **The link-text fix went to the more conservative helper rather than the minimal one.** Adding backtick+`*` to `mdLinkText` would have worked, but `mdInertText` is the defense-in-depth choice and matches the original design intent the first refactor had dropped — and the new comment restores the rationale (with reviewer credit) so the next person doesn't re-introduce it.
- **Every fixed vector got a named regression test.** Both injection classes are now pinned with the exact hostile inputs, so a future refactor that re-opens either fails CI.

- CI `check-sprint` — admin Sprint-field gate, independent of the code. Build/Test/Vet pending; the injection regression tests pass locally.

Jerry-Xin (approved):
Summary: This PR is in scope for `octo-server`; it extends the existing Multica incoming webhook adapter and the implementation looks ready to merge.

💬 Non-blocking
- 🔵 Suggestion: Now that `safeMarkdownURL` exists in `modules/incomingwebhook/adapter.go:165`, consider using it in a follow-up for the other markdown link renderers that still interpolate destinations after weaker checks, such as `modules/incomingwebhook/adapter_feishu.go:130`. The Multica path itself uses the safer helper correctly.

✅ Highlights
- `modules/incomingwebhook/adapter_multica.go:165` correctly downgrades unsafe or malformed `issue_url` values to plain text instead of interpolating them into markdown.
- `modules/incomingwebhook/adapter_multica.go:170` uses inert escaping for link text, which avoids emphasis/code-span injection in the new linked title path.
- `modules/incomingwebhook/adapter_multica.go:195` cleanly omits the metadata line when both assignee and actor are absent.
- The added regression tests cover link rendering, malicious URL downgrade, link-text injection, assignee-only rendering, and backward-compatible no-URL behavior.

Verification: I ran `go test ./modules/incomingwebhook/ -run 'TestParseMulticaPush|TestMulticaAdapter|TestPublicURLs|TestSafeMarkdownURL'`; it passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#496 Review Report
**Head SHA:** `eec9cfa81db63781fc5bf48cad9bb0b2980f769c`
**PR:** feat(incomingwebhook): render Octo-Q issue link + assignee (multi-line)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `safeMarkdownURL` blocks destination injection | ✅ | `adapter.go` L149–190: scheme check + `url.Parse` + metachar rejection (`< > ( ) [ ] " \`), non-ASCII reject, space/control reject |
| Link text uses `mdInertText` (not `mdLinkText`) | ✅ | `adapter_multica.go` L165: `mdInertText(linkText, 200)` — escapes `*_[]<>\|` and strips backticks, preventing emphasis/code-span injection inside `[text](url)` |
| Backward compat: old Octo-Q (no new fields) degrades gracefully | ✅ | Empty `issue_url`/`assignee_name` → `safeMarkdownURL` returns false → plain-text path identical to base behavior |
| All existing tests adapted for new format | ✅ | 5 existing tests updated: `FullEnvelope`, `NoPreviousStatus`, `NoActorType`, `TitleEscaping`, `TitleInjectionVectors`, `IdentifierMarkdownEscaping`, `StatusBacktickStripping`, `ActorTypeMarkdownEscaping`, `LongTitleIsClipped` — all assertions verified compatible with new multi-line output |
| New test coverage for link/assignee/injection vectors | ✅ | 7 new tests: `WithLinkAndAssignee`, `AssigneeWithoutActor`, `MaliciousIssueURLDowngraded`, `NoIssueURL_PlainTextTitle`, `IssueURLDestinationInjection` (6 sub-cases), `LinkTextInjection`, `TestSafeMarkdownURL` (11 cases) |
| No auth/permission surface change | ✅ | `parseMulticaPush` and `pushAdapter` registration unchanged; new fields consumed only in rendering |

---

## 2. Findings

### No P0 or P1 issues found.

### P2 — Identifier clip budget differs in link-text path (nit)

**Diff-scope:** new (introduced by this PR's branching logic)

In the link-text path (`adapter_multica.go` L165), the identifier is concatenated with the title into a single `mdInertText(linkText, 200)` call, meaning the identifier no longer gets its own 64-rune clip budget — it shares the 200-rune budget with the title. In the plain-text fallback path, `id` still gets `mdInertText(id, shortFieldMax)` (64 runes).

**Impact:** In practice Octo-Q identifiers are short controlled strings ("MUL-12345"), so a >64-rune identifier is not realistically reachable. The theoretical risk is that a malformed identifier of e.g. 180 runes would consume most of the link-text budget before the title gets clipped. Not a security or correctness concern — just a minor deviation from the design principle stated in the base code comments ("identifier 在 Octo-Q 端是短标识符, 64 足够覆盖正常上限").

**Suggestion (optional):** If the team wants to keep the 64-rune discipline, consider clipping `id` to `shortFieldMax` before concatenation:
```go
linkText := mdInertText(id, shortFieldMax)
if title != "" {
    linkText += " " + mdInertText(title, 200)
}
```
This would require adjusting the Fprintf to pass the pre-escaped string directly. Not blocking.

---

## 3. Suggestions

- **Minor:** PR body says "link text via `mdLinkText`" but the implementation correctly uses `mdInertText` (which is more thorough — escapes `*_<>|` in addition to `\[]`). The code is right; consider correcting the PR description for documentation accuracy.

---

## 4. Extra Observations

- `safeMarkdownURL` rejects IDN/non-ASCII hosts (e.g. `https://例子.com/x`). This is the conservative/safe choice. If Octo-Q later supports IDN workspaces, the function can be relaxed to accept percent-encoded hosts while still rejecting raw non-ASCII.
- URLs with parentheses in the path (e.g. Wikipedia-style `https://en.wikipedia.org/wiki/Foo_(bar)`) are rejected. This is safe — parens in markdown link destinations have complex parser-implementation-dependent semantics. Acceptable trade-off for a webhook rendering context.
- The `\n` literal newlines used for multi-line rendering are safe: `oneLine()` inside the escaping helpers collapses any field-internal newlines, and the only newlines in the final output are the intentional structural ones.

---

## 5. Data-Flow Trace

Each consumed datum traced to its upstream source:

| Consumed Datum | Upstream Source | Flow Verified |
|---|---|---|
| `ev.IssueURL` | `muEnvelope.IssueURL` ← `json.Unmarshal` of `"issue_url"` | ✅ TrimSpace → `safeMarkdownURL` → safe link or degrade. Empty string (old Octo-Q) → `isHTTPURL("")` false → degrade |
| `ev.AssigneeName` | `muEnvelope.AssigneeName` ← `"assignee_name"` | ✅ TrimSpace → empty check gates rendering → `mdInertText` escapes. Empty → meta line omitted if actor also empty |
| `ev.AssigneeType` | `muEnvelope.AssigneeType` ← `"assignee_type"` | ✅ TrimSpace → appended in parens after assignee name. Empty → parens omitted |
| `ev.Actor.Type` | `muEnvelope.Actor.Type` ← `"actor"."type"` (existing) | ✅ Unchanged extraction, now rendered as "触发: {type}" instead of "(by {type})" |
| `safeURL` from `safeMarkdownURL` | `raw` input after TrimSpace → `isHTTPURL` → `url.Parse` → char-by-char validation | ✅ Returns `(raw, true)` only when all checks pass; raw is the trimmed original (not reconstructed), preventing normalization artifacts |
| `linkText` in `mdInertText` | `id + " " + title` (both from existing envelope fields) | ✅ Both fields TrimSpaced; `mdInertText` applies `oneLine` (collapses newlines) + `clipRunes(200)` + full metachar escaping |

No data reaches a consumption point with a stale/empty/unchecked value.

---

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths touched. Rendering-only change.
- **C2 — Control-flow ordering / nesting:** Clear. `safeMarkdownURL` has a single call site (no nesting risk). `meta` slice is built in deterministic order (assignee → actor). No safety-control regex to probe with non-canonical forms.
- **C3 — Authorization boundary ≠ capability:** N/A. No new endpoints, tools, or capabilities exposed. Webhook rendering only.
- **C5 — Build ≠ runtime:** N/A. No build artifacts, extensions, CLI, or relative-path changes.
- **C6 — Governance/policy doc consistency:** N/A. No policy or governance docs changed.

---

---

## Verdict

All new fields are additive with graceful degradation. The two injection surfaces (URL destination and link text) are correctly defended: `safeMarkdownURL` is thorough against destination injection (scheme + parse + metachar + non-ASCII + control-char), and `mdInertText` on link text prevents emphasis/code-span injection inside `[text](url)`. Test coverage is comprehensive with dedicated regression tests for both blocker vectors. The one P2 observation (identifier clip budget) is a minor consistency nit with no practical impact.

**[Octo-Q] verdict: APPROVE** — only P2/nit findings; no P0/P1.

yujiawei (approved):
# Code Review — PR #496 (octo-server)

**Head reviewed:** `eec9cfa8` · base `main`
**Scope:** `modules/incomingwebhook/adapter.go`, `adapter_multica.go` (+ tests). Independent re-review of the fix commit addressing the two injection blockers + the unused-field note from the prior round.

Local verification: `go build ./...` ✅, `go vet ./modules/incomingwebhook/` ✅, and the full `TestParseMulticaPush* / TestSafeMarkdownURL / TestMulticaAdapter*` suite passes. (The only failing test, `TestDisableByGroupNo`, panics in `NewTestServer` for lack of a local MySQL — pre-existing, DB-dependent, untouched by this diff, and runs green in CI.)

## 1. Spec compliance

The diff implements exactly the stated two-repo contract and nothing beyond it:

- **Built:** `muEnvelope` gains `issue_url` / `assignee_type` / `assignee_name` (whitelisted JSON fields); `renderMulticaIssueStatusChanged` emits the agreed three-line structure (linked title / status / assignee·actor), with each line omitted when its fields are empty.
- **Backward compatible:** an older sender that omits the new fields degrades to the no-link / no-assignee form — covered by `TestParseMulticaPush_NoIssueURL_PlainTextTitle` and `TestParseMulticaPush_NoActorType`.
- **No over-build / no scope creep:** the change is confined to the Multica adapter; no other adapter, route, or schema is touched.

All three prior-round findings are resolved:
1. **URL-destination injection** → new `safeMarkdownURL` (`adapter.go:152`): `net/url` parse + http(s) + non-empty host + rejects ASCII space, control chars, non-ASCII, and the destination-breaking metachars `< > ( ) [ ] " \`. Failure downgrades the title to plain text.
2. **Link-text injection** → the link **text** now uses `mdInertText` (escapes `* _ [ ] < > |`, strips backtick) instead of `mdLinkText`, restoring protection the refactor had dropped (`adapter_multica.go:162`).
3. **Unused `assignee_type`** → now rendered as `指派: <name> (<type>)`.

**Spec: ✅**

## 2. Code quality

Both injection vectors were re-verified by **direct execution against the actual render path**, not by inspection alone:

- `safeMarkdownURL` correctly rejects `https://ok/) [phish](https://evil`, trailing `)`, embedded space/newline, angle brackets, and brackets — all degrade to a bold plain-text identifier. Percent-encoded forbidden chars (`%29`, `%20`) and legitimate userinfo/port/query/fragment URLs are preserved without becoming dangerous (the renderer does not percent-decode during markdown parsing).
- Link-text path: a title of `Fix)](https://attacker.evil.com` was traced end-to-end — `mdInertText` escapes the `]` to `\]`, so the injected `](…)` cannot close the link; the destination remains the safe `issue_url` and the attacker string stays inert link text. An adversarial attempt to claim a `)`-based bypass was reproduced and **does not hold**: escaping the closer `]` is sufficient, so a literal `)` is harmless.
- Backtick balance is preserved across the multi-line output for titles/status/assignee carrying stray backticks (`mdInertText` strips, `mdCodeSpanText` strips) — verified with odd-count assertions.
- Ordering is correct: `oneLine` neutralizes CR/LF before escaping, and `clipRunes` runs **before** escaping so a truncation boundary can never sever an escape backslash (confirmed with a trailing-`\` title at the clip edge).

Escaping is context-appropriate throughout (inert vs. code-span vs. link-destination), field lengths are clamped, and the minimal-field guard (`id`/`status` required) preserves the `reason=content` fallback.

**Quality: Approved** — no P0/P1. No blocking findings.

### Non-blocking notes (nits, not gating)
- The doc comment for `safeMarkdownURL` says it returns `(",", false)` on rejection; the code correctly returns `("", false)` (empty string). Cosmetic typo in the comment only.
- The new `\n`-joined multi-line `Content` is a first for these adapters; tests assert the substrings but not the literal newlines. Behaviorally fine for the markdown-rendering clients, just noting the format shift for future maintainers.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. The two markdown-injection blockers are genuinely closed (empirically verified), the change is additive and backward-compatible, and test coverage for the injection vectors is strong.

Note: the red `check-sprint` check is an org/sprint-tracking gate unrelated to this code; all code CI (Build / Lint / Vet / i18n / scans) is green.

lml2468:
@Jerry-Xin @yujiawei @OctoBoooot 两个 markdown 注入 blocker + assignee_type 未用 已在 `eec9cfa8` 修复(详见上方说明),分支已 rebase 到最新 main(`b5a602f4`),并关联 #497。Build/Vet/Lint/Test/code-review/osv-scan 全绿。

⚠️ `check-sprint` 仍 fail——它需要 Maintainer 把关联的 #497 加到当前 Octo sprint board(脚本提示 "ask a Maintainer"),这步我无权限完成,麻烦协助。复审谢谢 🙏

OctoBoooot (approved):
## Review: feat(incomingwebhook): render Multica issue link + assignee (multi-line) (#496) — delta @ b5a602f4

**Verdict:** Approve (re-affirmed) — this delta is a `main`-merge bringing in the already-merged #495 redis-chokepoint work; the #496 injection fix is byte-identical to the SHA I approved. No new findings, no blockers. CI red on `check-sprint` only (admin gate); Build/Lint pending.

Delta from prior APPROVE @ `eec9cfa8`: 2 commits — the merged #495 ("instrument raw redis clients") + a `Merge origin/main`. No #496-specific source changed.

### Verified — byte-level

- ✓ **The Multica injection fix is unchanged** — `git diff eec9cfa8..b5a602f4` for `adapter_multica.go` + `adapter.go` is empty. `safeMarkdownURL` (URL-destination validation) and the `mdInertText` link-text escaping are intact at the new head; the merge didn't disturb either vector's fix. Both injection regression tests (`TestParseMulticaPush_IssueURLDestinationInjection`, `TestParseMulticaPush_LinkTextInjection`) pass on the merged tree.
- ✓ **The incoming changes are the already-reviewed #495** — the ~20 `rd.NewClient(MustBuildOptions(...))` → `NewInstrumentedClient(...)` migrations, `safeMarkdownURL`'s sibling chokepoint constructors, and the `chokepoint_guard_test.go` source-guard. This is code I independently reviewed and APPROVED on #495 @ `8f7bce0a`; arriving here via `main` merge. The guard test passes against the merged tree (no raw client reintroduced).
- ✓ **go.mod merge is clean** — no `replace` directive, no conflict markers; just the #495 octo-lib bump (`...0c34e6f108c4`, the merged #104 commit). `go build` of the merged touched areas (incomingwebhook / redis / oidc) passes.

- CI `check-sprint` — admin Sprint-field gate, independent of the code. Build/Lint pending; the security regression tests + chokepoint guard pass locally.
- All four reviewers had already approved `eec9cfa8`; this merge carries that state forward since no review-relevant source changed. The injection fix and its regression tests are the load-bearing content, and both are intact.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#496 Review Report (automated review)
**PR:** Mininglamp-OSS/octo-server#496
**Head SHA:** `b5a602f4fdb8e12422fb18b4eb7fdd90a8d2870f` (merge of `eec9cfa8` into `f3b2a785`)
**Title:** feat(incomingwebhook): render Octo-Q issue link + assignee (multi-line)
**Changed files:** 4 (adapter.go, adapter_multica.go, adapter_multica_test.go, adapter_test.go)
**Diff stat:** +284 / -48

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| URL destination injection protection | ✅ | `safeMarkdownURL` (adapter.go:153-191) — net/url parse + http(s) + host + per-char metachar reject |
| Link-text injection protection | ✅ | `mdInertText` used for link text (adapter_multica.go:170), not `mdLinkText` — escapes `*_[\]<>|`, strips backtick |
| Backward compat (old Octo-Q no new fields) | ✅ | Empty `issue_url` → `safeMarkdownURL` returns false → plain-text path (adapter_multica.go:174-179) |
| Assignee line omission when empty | ✅ | `len(meta) > 0` gate (adapter_multica.go:199-202) |
| `assignee_type` rendered (was parsed-unused) | ✅ | Appended as `(member/agent/squad)` (adapter_multica.go:194-196) |
| Existing escape helpers unchanged | ✅ | `mdInertText`/`mdLinkText`/`mdCodeSpanText` definitions untouched in adapter.go |
| Test coverage for injection vectors | ✅ | 6 destination-injection subcases + link-text injection + malicious-URL downgrade + safeMarkdownURL table |
| `go build ./...` | ✅ | Per commit message (not run in this environment; read-only review) |

## 2. Findings

### No P0/P1 findings

The two injection vectors identified in the earlier review round (commit `eec9cfa8` message) — URL destination injection and link-text injection — are both properly closed with defense-in-depth:

**URL destination (`safeMarkdownURL`, adapter.go:153-191):**
- Scheme whitelist via `isHTTPURL` (rejects `javascript:`, `data:`, etc.)
- `net/url.Parse` structural validation
- Host-required check (`u.Host == ""` rejected)
- Per-rune scan rejects: non-ASCII, space, control chars, and metacharacters `< > ( ) [ ] " \`
- Failure → graceful downgrade to plain-text title (same contract as non-http scheme)

**Link text (adapter_multica.go:170):**
- Correctly uses `mdInertText` (not `mdLinkText`) for text inside `[text](url)`
- CommonMark parses emphasis/code-spans inside link text, so `*`/backtick/`[]` must all be neutralized
- `mdInertText` escapes `*_[]<>|` and strips backtick — covers all known injection vectors
- `\]` produced by escaping is a literal `]` in CommonMark link text — does NOT close the link

### P2 (non-blocking)

**P2-1: Missing test case — empty title with valid URL**
- **Path:** `adapter_multica_test.go`
- **Diff-scope:** Pre-existing gap amplified by new code path. The new linked-title branch (adapter_multica.go:170-172) concatenates `id + " " + title` only when `title != ""`, but no test exercises the `title == ""` + valid URL path.
- **Risk:** Low — the code correctly produces `**[MUL-123](url)**` when title is empty (verified by tracing the `linkText := id` / `if title != ""` branch). But an explicit test would guard against future refactors that might not handle this edge.
- **Suggestion:** Add a test case like `TestParseMulticaPush_EmptyTitleWithLink` asserting `**[MUL-123](https://...)...**` is rendered.

**P2-2: `safeMarkdownURL` permits URLs with embedded userinfo (`user:pass@host`)**
- **Path:** `adapter.go:153-191`
- **Diff-scope:** New code. The character-level reject set does not include `@` or `:`, so `https://user:pass@evil.com/path` passes validation. The raw URL is then interpolated into the markdown link destination.
- **Risk:** Low — `issue_url` is constructed by Octo-Q's backend (not end-user input), so embedded credentials would only appear if Octo-Q's URL builder has a bug. Defense-in-depth would suggest stripping userinfo (`u.User != nil` → reject), but this is not a blocking concern.
- **Suggestion:** Add `u.User != nil` check to reject URLs with embedded credentials.

## 3. Suggestions

1. **Add test for empty title + valid URL** (see P2-1 above). Quick win.
2. **Consider rejecting userinfo-bearing URLs** in `safeMarkdownURL` (see P2-2 above). One-line addition: `if u.User != nil { return "", false }`.

## 4. Additional Observations

- The `assignee_type` field (member/agent/squad) is now rendered as parenthetical after the assignee name — this resolves the earlier "parsed but unused" concern from the first commit review. Good.
- The multi-line format (title → status → assignee/actor) is a clean separation. Each line is conditionally rendered based on field presence, which handles both old and new Octo-Q payloads correctly.
- Test fixtures cover the full matrix: full envelope, link+assignee, assignee-without-actor, malicious URL downgrade, no-URL plain text, destination injection (6 subcases), link-text injection, and all pre-existing tests are updated for the new format.

## 5. Data-Flow Tracing

| Consumed datum | Upstream source | Intermediate transforms | Safe at consumption point? |
|---|---|---|---|
| `issue_url` | `muEnvelope.IssueURL` (JSON `issue_url` from Octo-Q webhook) | `TrimSpace` → `safeMarkdownURL` (isHTTPURL + url.Parse + host check + per-char metachar scan) | ✅ — reject → plain-text downgrade |
| `assignee_name` | `muEnvelope.AssigneeName` (JSON `assignee_name` from Octo-Q) | `TrimSpace` → `mdInertText(_, 64)` | ✅ — full escape + clip |
| `assignee_type` | `muEnvelope.AssigneeType` (JSON `assignee_type` from Octo-Q) | `TrimSpace` → `mdInertText(_, 64)` | ✅ — full escape + clip |
| `actor.type` | `muEnvelope.Actor.Type` (JSON `actor.type`) | `TrimSpace` → `mdInertText(_, 64)` | ✅ — unchanged from prior version |
| `title` (link path) | `muEnvelope.Issue.Title` | `TrimSpace` → concat with `id` → `mdInertText(_, 200)` | ✅ — inert escape covers `*_[\]<>|` + backtick strip |
| `title` (plain path) | Same | `TrimSpace` → `mdInertText(_, 200)` | ✅ — same escape |
| `identifier` | `muEnvelope.Issue.Identifier` | `TrimSpace` → `mdInertText(_, 64)` | ✅ — unchanged |

All consumed data traced to JSON unmarshal of the webhook body. No data is derived from DB queries, URL params, or other external sources beyond the authenticated webhook payload.

- **C1 — Dual-path parity:** N/A. No symmetric add/remove/subscribe paths touched. The PR adds a new rendering path (linked title) with a proper fallback (plain text) — both paths use `mdInertText` for text fields. ✅
- **C2 — Control-flow ordering / nested reuse:** `mdInertText` is reused for both link-text and plain-text contexts — verified that its escape set is correct for both (link text: `\]` is literal in CommonMark link text; plain text: `\]` is inert). No double-escape or ordering issue. `safeMarkdownURL` is called once and the result used directly — no nested reuse. ✅
- **C3 — Authorization boundary ≠ capability boundary:** N/A. The webhook endpoint's auth is the URL token (unchanged). New fields come from the authenticated webhook body, not from end-user-facing endpoints. ✅
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth changes in this PR. ✅
- **C5 — Build/note ≠ runtime path:** N/A. This is a server-side rendering change, not a build artifact/extension/CLI change. The markdown output is rendered by chat clients (existing rendering pipeline, unchanged). ✅
- **C6 — Governance/policy doc consistency:** N/A. No documentation or policy files changed. ✅

---

## Verdict

No P0 or P1 findings. The PR properly addresses the two injection vectors (URL destination and link text) identified in earlier review rounds. All new externally-sourced data flows through appropriate escaping helpers (`mdInertText` for text, `safeMarkdownURL` for URL destination, `mdCodeSpanText` for code spans). Backward compatibility with older Octo-Q versions is maintained via graceful degradation. Test coverage is comprehensive. Two P2 notes (missing empty-title+URL test case, userinfo-bearing URL not rejected) are non-blocking suggestions.

[Octo-Q] verdict: **APPROVE** — clean, well-tested change with solid injection protection. Two minor P2 suggestions for follow-up.

OctoBoooot (commented):
Byte-verified @mochashanyao's 🟡#2 at b5a602f4 — real, worth hardening, but scoped: it's deception, not a reopening of the fixed breakout.

`safeMarkdownURL` checks scheme + `u.Host != ""` + rejects breakout metachars, but never inspects `u.User`. So userinfo URLs pass: `safeMarkdownURL("https://app.multica.ai@evil.com/x")` → ok=true. The real host is `evil.com`; `app.multica.ai` is just userinfo.

Severity calibration (why 🟡, not a re-opened 🔴):
- The prior blocker was a *structural* breakout (close the link, inject a second `[phish](...)`). That stays fixed — a userinfo URL has no breakout chars, so it renders as ONE well-formed link.
- This is deception *within* one legitimate link. And the visible link text here is the issue identifier/title (`MUL-123 ...`), not the URL — so the classic "URL looks like trusted.com" trick is weakened; the victim sees the issue title, not the host. The residual risk ("a notification links to an attacker host") already exists for any attacker-controlled `issue_url` like a plain `https://evil.com`; userinfo adds obfuscation, not new capability.

Still worth closing as defense-in-depth — one line: `if u.User != nil { return "", false }` in `safeMarkdownURL`, plus a test. 🟡#1 (empty-title + valid-URL case) is a fair coverage gap too. Both non-blocking; verdict unchanged, APPROVE stands.

yujiawei (approved):
# Code Review — PR #496 (octo-server)

Reviewed at head SHA `b5a602f4fdb8e12422fb18b4eb7fdd90a8d2870f`. Built `./modules/incomingwebhook/`, ran `go vet`, and the full `incomingwebhook` test suite — all green, including the new link-injection and assignee regression cases.

## Spec compliance

Matches issue #497 exactly, and the change is additive / backward compatible:

- `muEnvelope` gains `issue_url` / `assignee_type` / `assignee_name` via the existing whitelist parse; an older sender that omits them degrades to the no-link / no-assignee form (covered by `TestParseMulticaPush_NoIssueURL_PlainTextTitle` and `TestParseMulticaPush_NoActorType`).
- Multi-line render (linked title / status line / assignee·trigger line) with each line omitted when its fields are absent.
- Link **destination** is validated structurally (`safeMarkdownURL`), not just by scheme prefix; link **text** uses `mdInertText` (escapes `*`, strips backtick) rather than `mdLinkText`, so emphasis/code spans inside the link text can't activate.

No missing, extra, or deviating behavior relative to the linked issue.

## Findings

### P2 (advisory, non-blocking) — `safeMarkdownURL` does not reject a backtick in the destination

`modules/incomingwebhook/adapter.go:165` `safeMarkdownURL` rejects space, control chars, non-ASCII, and the link-structural set `< > ( ) [ ] " \`, but **not** the backtick. A `issue_url` such as `https://ok.com/a` + "`" + `b` therefore passes validation and is interpolated raw into the link destination:

```
**[MUL-1 hi](https://ok.com/a`b)**
状态: `todo` → `in_progress`
```

CommonMark resolves code spans before links, so the lone backtick in the destination pairs with the first backtick of the downstream status line. The total backtick count becomes odd and both the link and the `状态:` code-span render incorrectly — the exact "backticks must remain balanced" invariant that `TestParseMulticaPush_LinkTextInjection` already pins for the link *text* path. The destination path leaks it.

Severity is P2, not a blocker: `issue_url` is server-constructed (PublicURL + slug + identifier), so this is not realistically attacker-reachable today, and the worst case is one cosmetically garbled notification — no script injection (a backtick cannot open an HTML/JS context here), no data loss, no boundary crossed.

Suggested fix: add backtick to the rejected set in `safeMarkdownURL` (degrade to plain-text title), and add a `safeMarkdownURL` case plus a "balanced backticks" assertion for a backtick-bearing URL, mirroring the existing link-text test. `*` / `_` / `|` in the destination are fine to leave as-is — once a link is formed, its destination is not re-scanned for emphasis or table syntax.

## Coverage / blind spots

- Verified in this diff only: the three new fields are consumed solely in `renderMulticaIssueStatusChanged`; no other adapter or caller references them (`go vet` clean, grep confirms).
- Not assessable from the octo-server diff alone: the **Multica-side** producer (the paired multica#49) — whether `issue_url` is in fact always host-bearing http(s) and whether `assignee_name` is length-bounded upstream. The `clipRunes(64)` clamp and `safeMarkdownURL` degrade make octo robust regardless, so this is informational, not a gate.
- The second advisory leg (cross-family long-context reviewer) timed out and is **absent** from this review — its silence is not an implicit approval. This review reflects the independent line-level leg (which found no correctness/security issue) plus this reviewer's own audit (the one P2 above).

## Verdict

No P0/P1. Spec-complete and well-tested. The single P2 is a follow-up hardening, not a merge blocker. Approving.

OctoBoooot (commented):
Concurring with @yujiawei — byte-verified at b5a602f4, and it corrects an overstatement in my APPROVE. I praised `safeMarkdownURL` as rejecting "all breakout chars"; it doesn't reject backtick, and backtick is a render-breaker here.

`safeMarkdownURL` rejects `< > ( ) [ ] " \`, space, control, non-ASCII — but not backtick. So `issue_url = "https://x.com/\`"` passes, and the render becomes:
`**[MUL-1 hi](https://x.com/\`)**\n状态: \`done\` → \`todo\``
— the document now has an odd number of backticks (5), so CommonMark's code-span pairing reaches across into the `状态:` line and corrupts it. This is the exact odd-backtick hazard the original design comment described for the *title* — `mdInertText` strips backtick from the link text, but the link *destination* doesn't, so the defense is applied to one context and not the symmetric one.

Severity 🟡, not a reopened blocker: `)`/`]`/`[` are still rejected, so no second-link injection / structural breakout — this is display corruption of the status line, not escalation. Fix: add backtick to the rejected set in `safeMarkdownURL` (a backtick in a legitimate http URL would be percent-encoded anyway), plus a regression test asserting balanced backticks. yujiawei's paired suggestion (reject + test) is right.

Verdict unchanged: APPROVE stands. Good catch — this and the userinfo 🟡 are both fair defense-in-depth hardening for `safeMarkdownURL` to fold in together.

Jerry-Xin (approved):
Re-affirming APPROVE on b5a602f4 to clear the pending re-review request on me.

I've byte-verified across all three rounds: both injection vectors (URL destination via `safeMarkdownURL`, and title backtick/`*` via `mdInertText`) are fixed and pinned with regression tests; this latest commit is purely an origin/main merge (pulling in already-merged #495 redis work) that leaves the Multica adapter files byte-identical to the SHA I previously approved. No new issues.

The trailing 🟡 notes from this round (userinfo URL hardening, backtick-in-URL-destination, empty-title test coverage) are non-blocking polish — none reopens the fixed structural breakout.

✅ APPROVED.

OctoBoooot (approved):
Re-affirming APPROVE on b5a602f4 (state correction).

My verdict on this PR has been APPROVE since the injection fix landed at eec9cfa8 and re-affirmed after the main-merge at b5a602f4 (review 4588644033 / 4588692955). The two reviews I posted after that were `--comment` concurrences on peers' non-blocking 🟡s (userinfo-URL hardening; backtick-in-destination), which mechanically left my *latest* review state as COMMENTED — that understated my actual verdict, so this restores it to APPROVED.

For the record, byte-verified at this head:
- Both injection vectors fixed and unchanged after the main-merge (adapter diff vs eec9cfa8 is empty): `safeMarkdownURL` guards the URL destination; link text uses `mdInertText`.
- The open 🟡s (userinfo URL, backtick-in-destination, empty-title test) are non-blocking hardening — none reopens a structural breakout; `)`/`]`/`[` are still rejected.

No code concerns from me. APPROVE.

PR #494, opened 2026-06-28
feat(avatarrender): script-aware default-avatar text rule (group + personal)

an9xyz:
## Summary

Default (un-uploaded) avatars derived their display text with no script awareness — group took the first 4 runes, personal the last 2 — producing awkward output for English / mixed-script / digit names (`Backend Team`→`Back`, `Bug反馈群`→`Bug反` cramped 2×2, `Alice`→`ce`). This replaces it with a shared **script-aware, 2-glyph** rule for both group and personal default avatars (signed off by product/UI via sample renders).

## Related Issue

N/A — post-#486 test-environment feedback (follow-up to the group/personal default-avatar work).

## Linked Spec

`.octospec/tasks/default-avatar-text-rule/brief.md` (+ `context.yaml`); journal `.octospec/journal/shared/default-avatar-text-rule.md`.

## Changes

- **New shared core `extractAvatarText`** in `pkg/avatarrender/text.go`, per name (first match wins):
  1. strip invisible (space/Cc/Cf); empty → `""` (caller falls back to an icon)
  2. any Han → Han chars only (drop Latin/digits/symbols), 2 glyphs
  3. else pure digits → 2 glyphs
  4. else has a letter → initials (first letter per token, camelCase/separator split, ≤2, uppercase; single word → 1 letter)
  5. else (pure symbol / emoji) → `""` (icon)
- **`GroupNameText`** (leading 2, group auto-name) + **rewritten `IndividualText`** (trailing 2, personal — DingTalk/Feishu convention). Initials lead for both.
- **`GroupText` kept unchanged** — it is the *custom* `avatar_text` normalizer (≤4), also used on the write path (`modules/group/api.go:793,1001`). `writeGroupDefaultAvatar` now **splits** user-set `avatar_text` (→ `GroupText`, rendered as-is) vs auto group Name (→ `GroupNameText`), so an explicit custom text is never truncated to 2 or initial-ized.
- **Cache-version bump** (derived bytes change): `group-name-v3→v4` and `name-v4→v5` at **both** the ETag and CacheKey sites; `group-icon-v3` and `ascii-v1` unchanged (their renderers are untouched). Version-pin tests updated to v4/v5.

[...]

PR #467, opened 2026-06-25
feat(messages_search): gate image/video in /_search_all by keyword pr…

BrDing-Rookie:
## Summary

Branch the `payload.type` whitelist in `/_search_all` on whether the request carries a keyword. With a keyword we keep the existing text-bearing whitelist `[1, 8, 11, 14]`; with an empty keyword (browse mode) we layer in image (`2`) and video (`5`) so the unified feed surfaces recent photos/videos alongside text.

## Changes

- `modules/messages_search/search_all.go` — in `buildSearchAllDSL`, build `allowedTypes` from the text-bearing set `[Text, File, MergeForward, RichText]` and append `Image, Video` only when `req.Keyword == ""`. Added an inline comment explaining why media is gated.
- `modules/messages_search/dsl_test.go`:
  - Update `TestBuildSearchAllDSL_NoKeywordKeepsTypeFilter` to expect `[1, 8, 11, 14, 2, 5]` (browse mode now includes media).
  - Add `TestBuildSearchAllDSL_ImageVideoGatedByKeyword` with two sub-cases: keyword `"hello"` excludes `2/5`; empty keyword includes `2/5`.

## Testing

- `go build ./...` — clean
- `go test ./modules/messages_search/... -count=1` — passes (both updated and new cases)
- Full-repo `go test ./...` not run locally (requires MySQL/Redis/WuKongIM).

- [x] Unit tests added/updated
- [x] Manually verified (DSL-level assertions on built ES query)

## COMPREHENSION

1. **What does this change actually do to the load-bearing path?**
   `/_search_all` is the unified browse + search endpoint. The DSL builder previously always filtered `payload.type IN [1, 8, 11, 14]`, so image and video messages never appeared regardless of keyword. This PR splits that filter on `req.Keyword`: keyword-driven queries keep the text-bearing whitelist; browse mode (no keyword) widens the filter to include image (`2`) and video (`5`). All other clauses — channel scope, space scope, virtual-doc exclusion, revoked filter, `should[textClause, fileClause]` + `minimum_should_match=1` for keyword queries, common filters, sort — are untouched.

2. **What could break because of it (dependents + failure mode)?**
   - **Browse-mode result-shape change**: image/video now show up in `/_search_all` with empty keyword. They render through `singleMessageHit` → `result_type="message"`. Snippet comes from `fallbackSnippet`: image uses `payload.image.caption`, video has no text field → empty snippet. Frontends that assumed `/_search_all` browse only returns text-ish results will see new payload shapes.
[...]

PR #500, opened 2026-06-29
feat(group): user-named default avatar (is_named) + public avatar color palette endpoint

an9xyz:
## Summary

Make the default group avatar follow a **user-given** group name, while a group left on the auto member-name-concatenated default name renders the two-person icon. Also add a public endpoint exposing the fixed avatar color palette so web clients can render local previews that match the server PNG.

Backward-compatible: existing groups are backfilled `is_named=1`, so **no existing avatar changes**; only newly created *unnamed* groups now get the icon.

## Related Issue

N/A — product task "group-chat-avatar-gen" (2026-06-29). See Linked Spec.

## Linked Spec

`.octospec/tasks/group-avatar-icon-default/brief.md`

## Changes

- Add `is_named` column (migration `20260629000001`, reentrant; existing rows backfilled to `1`) distinguishing a user-given name from the member-concat auto default.
- `CreateGroup` infers `is_named` from whether `name` was provided (computed before the member-concat fallback); rename (`UpdateGroupInfo`) sets `is_named=1`, persisted via `DB.UpdateTx`.
- `writeGroupDefaultAvatar` text priority: custom `avatar_text` > named group's name first-2 glyphs (`GroupNameText`) > two-person icon.
- New public endpoint `GET /v1/group/avatar_palette` returning the 10-color palette (`main`/`fill`/`icon_back` hex) with a content-derived weak ETag + `304`; `avatarrender.PaletteHex()` added.
- Comments / swagger / octospec brief synced to the `is_named` rule. No render-version bump (named = `group-name-v4`+text, auto-named = `group-icon-v3`; modes differ by ETag factors).

## Testing

- [x] Manually verified

Local run against real MySQL 8 + Redis + WuKongIM: `go build ./...`, `go vet ./...`, `golangci-lint` (0 issues), `make i18n-lint`, `go test ./modules/group/...` (full suite ~21s), `go test ./pkg/avatarrender/...`.

New/updated tests: `TestGroupAvatarGetAutoNamedRendersIcon`, `TestGroupAvatarGetNamedRendersNameText`, `TestCreateGroup_{Success,AutoGenerateName}` is_named asserts, `TestUpdateGroupInfo_RenameMarksIsNamed`, `TestGroupAvatarPalette` (+ ETag/304), `TestPaletteHex`; custom-text / uploaded-redirect / 404 / disband / version-pin regressions still pass.

## COMPREHENSION

   The default group-avatar render branch (`writeGroupDefaultAvatar`): with no custom `avatar_text`, it now picks text by the new `is_named` flag — a user-named group (`1`) renders the name's first 2 glyphs, an auto member-concat-named group (`0`) renders the two-person icon. `is_named` is inferred at create (was a name provided?) and set to `1` on rename. A separate, additive, public read-only endpoint exposes the palette; no existing render path or route is touched.

   The load-bearing surface is the default-avatar PNG and its cache. Existing groups are backfilled `is_named=1`, so their rendered bytes and ETags are identical to before (#494 name-derived) — no stale-cache risk and zero visual change. The only behavior change is that **newly created unnamed groups** switch from concat-name text to the icon (intended). Mode switches (named↔auto, or a rename) change the ETag's text factor, so clients revalidate to the fresh image without a version bump. The new endpoint is purely additive, public, static, and route-conflict-free (mounted under `/v1/group` on a non-wildcard segment).

   Endpoint tests byte-compare the rendered PNG against the expected renderer on real infra: `...NamedRendersNameText` asserts a named group == `RenderGroup(GroupNameText(name))`; `...AutoNamedRendersIcon` asserts an auto-named group == `RenderIcon`; `TestUpdateGroupInfo_RenameMarksIsNamed` proves rename persists `is_named=1`; the two `CreateGroup` tests prove create-time inference. The migration is auto-applied by `NewTestServer`. Full group + avatarrender suites, golangci (0 issues), and i18n-lint are green.

## Checklist

- [x] Updated documentation (swagger + octospec brief)

https://claude.ai/code/session_017D9g64sSjENNaXdnfCnjwr

---
_Generated by [Claude Code](https://claude.ai/code/session_017D9g64sSjENNaXdnfCnjwr)_

### DIFF
--- a/modules/group/sql/20260629000001_group_is_named.sql
+++ b/modules/group/sql/20260629000001_group_is_named.sql
@@ -4,7 +4,15 @@
 -- 双人图标（避免把「张三、李四、王五」这种拼接名渲成头像文字）。
 -- 存量群事后无法区分两类名（都存在 name 里），按产品决策**保守回填为 1**——一律视为
 -- 命名群、保留现状（按群名取字），不改变任何既有头像；仅新建群按建群是否传入 name 计算。
--- 使用 INFORMATION_SCHEMA 守卫保证迁移在部分执行后重试时仍可重入。
+--
+-- 恢复安全（recovery-safe）三步法：MySQL 的 ALTER TABLE 隐式提交、不可回滚，因此不能把
+-- 回填放进「列是否存在」的守卫里——一旦 ADD COLUMN 已提交、回填 UPDATE 中途失败
+-- （崩溃 / 锁超时 / 死锁），重试时列已存在、整块被跳过，存量行将停留在 DEFAULT 0
+-- （= 自动名 → 双人图标），与「不改变既有头像」相悖。改用可空哨兵：
+--   1) 先以 NULL（无默认）建列：存量行得到 NULL = 「尚未回填」；
+--   2) 无条件回填 `WHERE is_named IS NULL`：幂等、只动未回填行，任意中断后重试都能补完；
+--   3) 再 MODIFY 为 NOT NULL DEFAULT 0：新建群默认 0，由建群逻辑显式写 0/1。
+-- 每步各自带幂等守卫，跨「隐式提交边界」的部分失败重试都能正确收敛。
 -- +migrate StatementBegin
 DROP PROCEDURE IF EXISTS __group_is_named;
 -- +migrate StatementEnd
@@ -12,13 +20,26 @@ DROP PROCEDURE IF EXISTS __group_is_named;
 -- +migrate StatementBegin
 CREATE PROCEDURE __group_is_named()
 BEGIN
+  -- Step 1: 以可空（无默认）建列——存量行 = NULL「尚未回填」。守卫保证可重入。
   IF NOT EXISTS (SELECT 1 FROM information_schema.COLUMNS
        WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'group'
          AND COLUMN_NAME = 'is_named') THEN
     ALTER TABLE `group`
-      ADD COLUMN `is_named` TINYINT NOT NULL DEFAULT 0 COMMENT '群名是否用户显式起名(1)/成员拼接自动名(0);默认头像取字仅对1生效';
-    -- 存量群保守回填为 1：保留现状（按群名取字），不改变既有头像。
-    UPDATE `group` SET `is_named` = 1;
+      ADD COLUMN `is_named` TINYINT NULL COMMENT '群名是否用户显式起名(1)/成员拼接自动名(0);默认头像取字仅对1生效';
+  END IF;
+
+  -- Step 2: 回填存量行为 1（保留现状、不改既有头像）。无条件执行但仅命中 NULL 行——
+  -- 幂等、可重复，任意中断后重试都能补完剩余未回填行，且不会覆盖已落定的 0/1。
+  -- 注：单条全表 UPDATE 仅命中存量 NULL 行，一次性回填；迁移在服务对外前运行，无并发写。
+  UPDATE `group` SET `is_named` = 1 WHERE `is_named` IS NULL;
+
+  -- Step 3: 收紧为 NOT NULL DEFAULT 0（新建群默认 0）。守卫在「当前可空」时才执行，
+  -- 使 Step 2 完成后的重试安全收敛。
+  IF EXISTS (SELECT 1 FROM information_schema.COLUMNS
+       WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'group'
+         AND COLUMN_NAME = 'is_named' AND IS_NULLABLE = 'YES') THEN
+    ALTER TABLE `group`
+      MODIFY COLUMN `is_named` TINYINT NOT NULL DEFAULT 0 COMMENT '群名是否用户显式起名(1)/成员拼接自动名(0);默认头像取字仅对1生效';
   END IF;
 END;
 -- +migrate StatementEnd
--- a/modules/group/sql/20260629000001_group_is_named.sql
+++ b/modules/group/sql/20260629000001_group_is_named.sql
@@ -1,0 +1,52 @@
+-- +migrate Up
+-- 区分「用户显式起的群名」(is_named=1) 与「成员名拼接的自动默认名」(is_named=0)。
+-- 默认头像取字仅对 is_named=1 生效：命名群取群名前 2 字（script 感知），自动名群回退
+-- 双人图标（避免把「张三、李四、王五」这种拼接名渲成头像文字）。
+-- 存量群事后无法区分两类名（都存在 name 里），按产品决策**保守回填为 1**——一律视为
+-- 命名群、保留现状（按群名取字），不改变任何既有头像；仅新建群按建群是否传入 name 计算。
+-- 使用 INFORMATION_SCHEMA 守卫保证迁移在部分执行后重试时仍可重入。
+-- +migrate StatementBegin
+DROP PROCEDURE IF EXISTS __group_is_named;
+-- +migrate StatementEnd
+
+-- +migrate StatementBegin
+CREATE PROCEDURE __group_is_named()
+BEGIN
+  IF NOT EXISTS (SELECT 1 FROM information_schema.COLUMNS
+       WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'group'
+         AND COLUMN_NAME = 'is_named') THEN
+    ALTER TABLE `group`
+      ADD COLUMN `is_named` TINYINT NOT NULL DEFAULT 0 COMMENT '群名是否用户显式起名(1)/成员拼接自动名(0);默认头像取字仅对1生效';
+    -- 存量群保守回填为 1：保留现状（按群名取字），不改变既有头像。
+    UPDATE `group` SET `is_named` = 1;
+  END IF;
+END;
+-- +migrate StatementEnd
+
+CALL __group_is_named();
+
+-- +migrate StatementBegin
+DROP PROCEDURE IF EXISTS __group_is_named;
+-- +migrate StatementEnd
+
+-- +migrate Down
+-- +migrate StatementBegin
+DROP PROCEDURE IF EXISTS __group_is_named_down;
+-- +migrate StatementEnd
+
+-- +migrate StatementBegin
+CREATE PROCEDURE __group_is_named_down()
+BEGIN
+  IF EXISTS (SELECT 1 FROM information_schema.COLUMNS
+       WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'group'
+         AND COLUMN_NAME = 'is_named') THEN
+    ALTER TABLE `group` DROP COLUMN `is_named`;
+  END IF;
+END;
+-- +migrate StatementEnd
+
+CALL __group_is_named_down();
+
+-- +migrate StatementBegin
+DROP PROCEDURE IF EXISTS __group_is_named_down;
+-- +migrate StatementEnd
--- a/modules/group/swagger/api.yaml
+++ b/modules/group/swagger/api.yaml
@@ -320,7 +320,7 @@ paths:
                   description: "成员名称"
               avatar_text:
                 type: string
-                description: "自定义群头像文字（≤4 可见字符；空/缺省时取群名前 2 字（script 感知取字）渲染默认头像）"
+                description: "自定义群头像文字（≤4 可见字符；空/缺省时按 is_named 回退：传入 name(命名群)取群名前 2 字（script 感知取字），未传 name(成员拼接自动名)回退双人图标）"
               avatar_color:
                 type: integer
                 description: "自定义群头像色板下标 [0, 调色板大小)；缺省时按 group_no 派生颜色"
@@ -1215,8 +1215,9 @@ paths:
         - "group"
       summary: "获取群头像"
       description: >-
-        公开（不鉴权）端点。未上传自定义头像的群在服务端实时渲染默认头像（彩色描边圆 +
-        群名前 2 字（script 感知取字），或自定义 avatar_text/avatar_color；空名回退群组图标），返回 image/png。
+        公开（不鉴权）端点。未上传自定义头像的群在服务端实时渲染默认头像（彩色描边圆）：
+        优先自定义 avatar_text；否则按 is_named 回退——命名群(is_named=1)取群名前 2 字
+        （script 感知取字），自动名群(is_named=0)或空名回退双人图标；avatar_color 控制配色。返回 image/png。
         已上传头像（is_upload_avatar=1）的群 302 重定向到对象存储。群不存在或已解散返回
         404。响应带内容相关弱 ETag，命中 If-None-Match 时返回 304。
       operationId: "get avatar"
@@ -1244,6 +1245,34 @@ paths:
           description: "If-None-Match 命中，内容未变更"
         404:
           description: "群不存在或已解散"
+  /group/avatar_palette:
+    get:
+      tags:
+        - "group"
+      summary: "获取群头像色板"
+      description: >-
+        公开（不鉴权）端点。返回服务端固定的群头像设计色板（main/fill/icon_back，#RRGGBB），
+        作为唯一数据源供前端「修改头像」色圈与本地实时预览使用，使预览配色与服务端渲染的
+        PNG 完全一致、不漂移。无入参、无敏感数据，按下标有序。响应带内容相关弱 ETag 与
+        Cache-Control: public, max-age=300, must-revalidate；命中 If-None-Match 时返回 304。
+      operationId: "get avatar palette"
+      consumes:
+        - "application/json"
+      produces:
+        - "application/json"
+      parameters:
+        - in: "header"
+          name: "If-None-Match"
+          type: string
+          description: "条件请求；与当前弱 ETag 匹配时返回 304"
+          required: false
+      responses:
+        200:
+          description: "固定群头像色板"
+          schema:
+            $ref: "#/definitions/avatarPaletteResp"
+        304:
+          description: "If-None-Match 命中，色板未变更"
   /groups/{group_no}/scanjoin:
     get:
       tags:
@@ -1283,6 +1312,32 @@ securityDefinitions:
     description: "用户token"
 
 definitions:
+  avatarPaletteColorResp:
+    type: object
+    properties:
+      index:
+        type: integer
+        description: "色板下标，与 avatar_color 取值一致"
+      main:
+        type: string
+        description: "主题主色 #RRGGBB"
+      fill:
+        type: string
+        description: "圆填充浅色 #RRGGBB"
+      icon_back:
+        type: string
+        description: "双人图标后景人浅色 #RRGGBB"
+  avatarPaletteResp:
+    type: object
+    properties:
+      size:
+        type: integer
+        description: "色板档数（= avatar_color 合法上界）"
+      colors:
+        type: array
+        description: "按下标 0..size 有序的色板项"
+        items:
+          $ref: "#/definitions/avatarPaletteColorResp"
   memberManagerResp:
     type: object
     properties:
--- a/modules/group/swagger/api.yaml
+++ b/modules/group/swagger/api.yaml
@@ -528,7 +528,7 @@ paths:
                 description: "修改群名称 notice(修改群备注) invite(是否开启进群邀请) 等"
               avatar_text:
                 type: string
-                description: "自定义群头像文字（≤4 可见字符；空串清除→回退双人图标，默认头像与群名无关）"
+                description: "自定义群头像文字（≤4 可见字符；空串清除→按 is_named 回退：命名群取群名前2字、自动名群双人图标）"
               avatar_color:
                 type: integer
                 description: "自定义群头像色板下标 [0, 调色板大小)；空串或 -1 清除→回退按 group_no 派生"
--- a/modules/group/swagger/api.yaml
+++ b/modules/group/swagger/api.yaml
@@ -528,7 +528,7 @@ paths:
                 description: "修改群名称 notice(修改群备注) invite(是否开启进群邀请) 等"
               avatar_text:
                 type: string
-                description: "自定义群头像文字（≤4 可见字符；空串清除→回退群名前 2 字（script 感知取字））"
+                description: "自定义群头像文字（≤4 可见字符；空串清除→回退双人图标，默认头像与群名无关）"
               avatar_color:
                 type: integer
                 description: "自定义群头像色板下标 [0, 调色板大小)；空串或 -1 清除→回退按 group_no 派生"
--- a/modules/group/swagger/api.yaml
+++ b/modules/group/swagger/api.yaml
@@ -1395,6 +1395,9 @@ definitions:
       name:
         type: string
         description: "群名称"
+      is_named:
+        type: integer
+        description: "群名是否用户显式起名(1)/成员拼接自动名(0)；客户端据此本地预判默认头像取群名前2字 vs 双人图标"
       remark:
         type: string
         description: "群备注"
@@ -1469,7 +1472,7 @@ definitions:
         description: "群数据版本"
       avatar_text:
         type: string
-        description: "自定义群头像文字（空表示未自定义，按群名前 2 字（script 感知取字）派生）"
+        description: "自定义群头像文字（空=未自定义，按 is_named 回退：命名群(is_named=1)取群名前 2 字（script 感知取字），自动名群(is_named=0)回退双人图标）"
       avatar_color:
         type: integer
         description: "自定义群头像色板下标（null 表示未自定义，按 group_no 派生）"
--- a/.octospec/tasks/group-avatar-icon-default/brief.md
+++ b/.octospec/tasks/group-avatar-icon-default/brief.md
@@ -67,6 +67,29 @@ Two server changes:
   *before* the member-concat fallback overwrites `groupName`. Rename
   (`UpdateGroupInfo`, `req.Name != nil`) → `is_named = 1`; persisted by adding
   `is_named` to `DB.UpdateTx`'s `SetMap` (it used an explicit column list).
+- **All direct creation paths set `is_named` (review #500, Jerry-Xin + OctoBoooot
+  🔴).** `CreateGroup` is not the only insert path: `event.go` system-group
+  (`handleRegisterUserEvent`) + org/dept (`handleOrgOrDeptCreateEvent`) and
+  `Service.AddGroup` insert named groups directly; left at the Go zero value they
+  persist `is_named=0`. Set `IsNamed: 1` on the system/org/dept inserts (they carry
+  an explicit configured name) and `AddGroup` infers it from `TrimSpace(name)!=""`.
+  Note: system/`org_`/`dept_` groups are served *static PNGs* by the prefix-gated
+  branches in `avatarGet` (`api.go:360-394`) and never reach the `is_named` render
+  path, so this is data-correctness/robustness, not a live avatar regression — but
+  the persisted flag must still be right (a removed prefix-branch would otherwise
+  silently break them).
+- **Combined `{name, invite}` update must not revert the rename / `is_named`
+  (review #500 P1, yujiawei + OctoBoooot).** `groupUpdate` (`api.go`) loads the row
+  once, then the name branch commits `name`+`is_named=1` via `UpdateGroupInfo`
+  (its own fresh load) and the invite branch wrote the **stale** snapshot back via
+  the full-column `UpdateTx` — clobbering both. Fix: invite branch uses a
+  column-scoped `DB.UpdateInviteTx` (only `invite`+`version`); the handler no longer
+  caches the full row (existence-check only). This also removes the pre-existing
+  `name`-revert on the same path and the read-modify-write race.
+- **`GroupResp.is_named` exposed (review #500 🟡).** Read-only `is_named` added to
+  `GroupResp` (both `from`/`fromModel` populators) so web clients can locally
+  predict name-text vs icon for an existing group when clearing `avatar_text`
+  (preview parity). Additive, backward-compatible.
 - **Avatar ETag / cache identity.** ETag is CRC32 over content *factors*
   (mode-version + group_no + color + text), not pixels. Named groups use the
   `group-name-v4` factor set (with text); auto-named/empty use `group-icon-v3`
--- a/.octospec/tasks/group-avatar-icon-default/brief.md
+++ b/.octospec/tasks/group-avatar-icon-default/brief.md
@@ -1,0 +1,91 @@
+---
+type: Task
+title: "Task: group-avatar-icon-default"
+description: Default group avatar becomes the two-person icon (name-independent); expose the avatar color palette to clients for local preview parity.
+tags: [group, avatar, render, cache, wire-contract]
+timestamp: 2026-06-29T00:00:00Z
+# --- octospec extension fields ---
+slug: group-avatar-icon-default
+upstream: group-chat-avatar-gen (product, 2026-06-29)
+source: self
+---
+
+# Task: group-avatar-icon-default
+
+> Server-side increment of the "group chat avatar" redesign. Web (octo-web) is a
+> separate follow-up: the 修改头像 secondary dialog + 发起群聊 create dialog with
+> live local preview consuming the palette endpoint below.
+
+## Goal
+Two server changes:
+
+1. **S2 — default avatar is the two-person icon, independent of the group name.**
+   Today an un-uploaded group with no custom `avatar_text` renders its name's
+   leading 2 glyphs (`GroupNameText`), falling back to the two-person icon only
+   when the name yields no renderable glyphs (PR #494). Product (2026-06-29) has
+   reversed this: the default avatar must be the two-person icon **regardless of
+   the group name**; text is rendered **only** when the user has explicitly set a
+   custom `avatar_text` via the 修改头像 dialog.
+
+2. **S1 — expose the avatar color palette over HTTP.** The 10-color palette
+   (`main` / `fill` / `iconBack`) lives only in `pkg/avatarrender/palette.go`.
+   The web color picker + local live preview must render the *same* colors as the
+   server PNG. Add a read endpoint so the palette has a single source of truth and
+   never drifts between server and client.
+
+## Background
+- Avatar render pipeline + custom `avatar_text`/`avatar_color` + the fixed palette
+  shipped in PRs #478 / #486 / #494 (journals: `group-default-avatar`,
+  `default-avatar-text-rule`). Default rendering decision is in
+  `modules/group/api.go` `writeGroupDefaultAvatar`.
+- The default-avatar-text-rule journal already flagged "unnamed groups → icon" as
+  a deferred follow-up needing an `is_named` flag. S2 makes that flag unnecessary:
+  **all** un-customized groups (named or not) use the icon.
+- Web research: there is no avatar dialog in octo-web today (image upload only);
+  the create flow (`OrganizationalGroupNew`) sends members only — no name/avatar.
+  Those are the web follow-up, not this task.
+
+## Load-bearing list
+- `writeGroupDefaultAvatar` text-selection branch (`modules/group/api.go`) — the
+  exact behavior being changed. Custom-text path (`GroupText`, ≤4, as-is) is
+  unchanged; only the no-custom-text fallback flips from `GroupNameText(name)` to
+  empty → icon.
+- **Avatar ETag / cache identity.** ETag is CRC32 over content *factors*
+  (mode-version + group_no + color + text), not pixels. Named groups switch from
+  the `group-name-v4` factor set (with text) to the `group-icon-v3` factor set
+  (no text) → the ETag **changes** on its own, so a stale `If-None-Match` cannot
+  match → clients revalidate to the fresh icon. The `RenderIcon` bytes are NOT
+  touched, so **no version bump is required** (contrast #486: there the icon
+  *pixels* changed). `group-name-v4` is retained for the custom-text path.
+- `GroupNameText` (`pkg/avatarrender/text.go`) — no longer called by the handler
+  after S2; kept (still unit-tested, may serve future use). No behavior change to
+  the function itself.
+- Stale comments in `modules/group/{api.go,db.go,service.go}` that say
+  "空=渲染时回退群名前 2 字派生" — updated to "空=双人图标(默认头像与群名无关)"
+  so future readers don't re-introduce name derivation.
+- **Wire contract (new):** `GET /v1/group/avatar_palette` (public, static design
+  tokens — mirrors the already-public avatar render endpoint). Response:
+  `{ "size": 10, "colors": [ { "index", "main", "fill", "icon_back" } ... ] }`,
+  hex `#RRGGBB`, ordered by palette index. Pure read, no error path → no errcode.
+
+## Out of scope
+- Server default-render rule per named/unnamed distinction (`is_named`) — obviated.
+- Any change to custom `avatar_text`/`avatar_color` create/update APIs, validation,
+  upload path, or the palette *values*/order.
+- All octo-web work (修改头像 dialog, 发起群聊 dialog, local preview component,
+  palette consumption) — separate task in octo-web.
+- Render version bump (intentionally none — see load-bearing list).
+
+## Acceptance
+- `writeGroupDefaultAvatar`: with no custom `avatar_text`, renders `RenderIcon`
+  (two-person icon) for BOTH a named and an empty-name group; renders text ONLY
+  when `avatar_text` is set. Custom color still honored for the icon.
+- `GET /v1/group/avatar_palette` returns `PaletteSize()` (=10) entries; each entry's
+  `main`/`fill`/`icon_back` equals `avatarrender.PaletteHex()`; `colors[0].main == "#14C0FF"`.
+- Updated `TestGroupAvatarGet*` (named, no custom text → icon) and
+  `TestGroupAvatarGetPinsRenderVersion` (text-render version pinned via a
+  custom-text group, since name no longer triggers text). Custom-text not-truncated
+  + uploaded-redirect + 404/disband regressions still pass.
+- `go build ./...`, `go vet`, `golangci-lint`, `make i18n-lint` +
+  `i18n-extract-check`, `TestGroupNoLegacyResponseError` guard all green.
+  (Endpoint tests needing MySQL/Redis/WuKongIM run in CI per prior increments.)
--- a/.octospec/tasks/group-avatar-icon-default/brief.md
+++ b/.octospec/tasks/group-avatar-icon-default/brief.md
@@ -19,13 +19,18 @@ source: self
 ## Goal
 Two server changes:
 
-1. **S2 — default avatar is the two-person icon, independent of the group name.**
-   Today an un-uploaded group with no custom `avatar_text` renders its name's
-   leading 2 glyphs (`GroupNameText`), falling back to the two-person icon only
-   when the name yields no renderable glyphs (PR #494). Product (2026-06-29) has
-   reversed this: the default avatar must be the two-person icon **regardless of
-   the group name**; text is rendered **only** when the user has explicitly set a
-   custom `avatar_text` via the 修改头像 dialog.
+1. **S2 — default avatar follows a *user-given* name; an auto-generated
+   (member-concat) name renders the two-person icon.** Today an un-uploaded group
+   with no custom `avatar_text` always renders its `name`'s leading 2 glyphs
+   (`GroupNameText`), including the member-concatenated default name like
+   `张三、李四、王五` (PR #494). Product (2026-06-29): the default avatar should take
+   the name's first 2 glyphs **only when the user explicitly named the group**; a
+   group left on the auto member-concat name renders the two-person icon (don't
+   render a concat name as avatar text). A new `is_named` flag distinguishes the
+   two (set at create from whether `name` was provided, and on rename). Priority:
+   custom `avatar_text` > named-group name (`is_named=1`) > two-person icon.
+   Existing groups are backfilled `is_named=1` (conservative — no existing avatar
+   changes; only new auto-named groups get the icon).
 
 2. **S1 — expose the avatar color palette over HTTP.** The 10-color palette
    (`main` / `fill` / `iconBack`) lives only in `pkg/avatarrender/palette.go`.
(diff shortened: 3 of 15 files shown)


Jerry-Xin (changes requested):
This PR is relevant to octo-server, but the migration has a rollback/retry hole that can break the main backward-compatibility guarantee.

**🔴 Blocking**
- 🔴 Critical: [modules/group/sql/20260629000001_group_is_named.sql](modules/group/sql/20260629000001_group_is_named.sql:15) is not actually retry-safe for the backfill. The `UPDATE group SET is_named = 1` only runs inside the `IF NOT EXISTS column` branch. Since `ALTER TABLE` commits independently in MySQL, a deployment interrupted after adding the column but before/during the update will leave existing rows at the new default `0`. On migration retry, the column exists, so the backfill is skipped. That makes existing no-custom-avatar groups render the icon through [modules/group/api.go](modules/group/api.go:449), violating the PR’s “no existing avatar changes” contract. A safer pattern is to add the column with a nullable sentinel, backfill `NULL` rows to `1`, then alter to `NOT NULL DEFAULT 0`, so retries can complete the backfill without guessing.

**💬 Non-blocking**
- 🟡 Warning: [modules/group/swagger/api.yaml](modules/group/swagger/api.yaml:321) still documents create-time empty `avatar_text` as falling back to group-name text, and [modules/group/swagger/api.yaml](modules/group/swagger/api.yaml:1218) still describes the avatar endpoint without the `is_named` distinction. The new `/v1/group/avatar_palette` endpoint also appears absent from the swagger paths. This is not a runtime blocker, but it will mislead client implementers.
- 🔵 Suggestion: [modules/group/avatar_get_test.go](modules/group/avatar_get_test.go:136) names the fixture `"有名群"` but leaves `IsNamed` at its zero value, so the test is exercising an auto-named group with custom color, not a named group. Setting the name/comment to match the fixture intent would make the coverage easier to trust.

**✅ Highlights**
- The runtime avatar branch is simple and matches the intended priority: custom text, named-group name text, then icon.
- The palette endpoint uses content-derived ETags and the response is derived from `avatarrender.PaletteHex()`, avoiding duplicated palette values.
- I ran `go test ./pkg/avatarrender/...` and `go test ./modules/group -run TestGroupNoLegacyResponseError`; both passed. The selected group integration tests could not run in this local DB state because the test migration table contains an unknown migration.

modules/group/swagger/api.yaml
```diff
@@ -528,7 +528,7 @@ paths:
                 description: "修改群名称 notice(修改群备注) invite(是否开启进群邀请) 等"
               avatar_text:
                 type: string
-                description: "自定义群头像文字（≤4 可见字符；空串清除→回退群名前 2 字（script 感知取字））"
+                description: "自定义群头像文字（≤4 可见字符；空串清除→按 is_named 回退：命名群取群名前2字、自动名群双人图标）"
```
OctoBoooot, line 531: **[minor]** this `avatar_text` description was synced, but the PR's *new public endpoint* `GET /v1/group/avatar_palette` (api.go:167) was NOT added to this swagger `paths:` section — `grep avatar_palette` of the spec is empty. The body says "swagger synced," which covers this line but not the endpoint. This file documents paths (`/group/create`, etc.), so a new public route is expected here; spec-generated clients won't know the endpoint or its `avatarPaletteResp` shape exists. Minor (additive read-only — it works, just undocumented), not a blocker like the #467 enum gap. Fix: add the path + response schema.

modules/group/sql/20260629000001_group_is_named.sql
```diff
@@ -0,0 +1,52 @@
...
+-- +migrate StatementBegin
+CREATE PROCEDURE __group_is_named()
+BEGIN
+  IF NOT EXISTS (SELECT 1 FROM information_schema.COLUMNS
+       WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'group'
+         AND COLUMN_NAME = 'is_named') THEN
+    ALTER TABLE `group`
+      ADD COLUMN `is_named` TINYINT NOT NULL DEFAULT 0 COMMENT '群名是否用户显式起名(1)/成员拼接自动名(0);默认头像取字仅对1生效';
+    -- 存量群保守回填为 1：保留现状（按群名取字），不改变既有头像。
+    UPDATE `group` SET `is_named` = 1;
```
OctoBoooot: **[minor]** `UPDATE \`group\` SET is_named = 1` rewrites every existing row in one statement — on a large `group` table this is a long-held lock during migration (and a big single transaction). The reentrant guard + Down are well done; this is purely deploy-time lock duration. If the table is large in prod, consider batching the backfill. Non-blocking — flagging since migrations are cheap to skim and expensive to discover under load.

OctoBoooot (commented):
## Review: feat(group): user-named default avatar (is_named) + public avatar color palette endpoint (#500)

**Verdict:** Approve with comments — clean, well-built feature; two minors (swagger gap on the new endpoint, full-table backfill lock). No blockers, no majors. CI red on `check-sprint` only (admin gate); Build/Vet/Lint/i18n all pass, Test pending.

Risk: normal by path, but it carries a **DB migration** + a **new public endpoint** — both verified. `needs-human-review` + `size/XL` on the PR. Established author (100 merged).

### Verified — the parts most likely to be wrong

- ✅ **Migration is reentrant + reversible** (`20260629000001`) — INFORMATION_SCHEMA guard so a partial-then-retry re-run is safe, `DEFAULT 0` for new rows, explicit backfill `UPDATE ... = 1` for existing, and a working `Down` that drops the column. (Lock note in the inline minor.)
- ✅ **`is_named` inference is correct on both edges** — create: `strings.TrimSpace(req.Name) != ""` → 1, else 0 (whitespace-only name correctly → unnamed → icon); rename (`service.go:1880`) sets `IsNamed = 1`. The [bidirectional] handling holds.
- ✅ **Public palette endpoint is safe to be unauthed** (`api.go:167` `openGroup`) — takes no input, returns only static design-token colors (no user/group data), bounded fixed-size output, content-derived ETag + 304. No leak, no injection, no unbounded surface. Appropriate as a public constant.
- ✅ **No render-version bump is correct** — existing groups backfill to `is_named=1`, so they stay on the `group-name-v4`+text path with the *same* ETag they already had → no stale cache; only newly-created unnamed groups get `group-icon-v3` (no prior cache to go stale). The mode-switch is encoded in the ETag factors, so named↔auto transitions invalidate naturally. Reasoning byte-checked against `writeGroupDefaultAvatar` (`api.go:447-460`).
- ✅ **`go vet` + avatarrender unit tests pass**, i18n Extract Check green (no new error codes), Build/Vet/Lint green in CI.

### Minor

- `swagger/api.yaml` — the new public `GET /v1/group/avatar_palette` is not in the `paths:` section (only the `avatar_text` description was edited). The body's "swagger synced" covers the field tweak, not the endpoint. Additive read-only, so minor — but the published contract is incomplete. See inline.
- `sql/20260629000001` — the `UPDATE \`group\` SET is_named = 1` backfill rewrites every row in one statement; long lock on a large table. See inline.

### Praise

- **The backward-compat story is the right one, and it's enforced in the migration, not just asserted.** Backfilling existing rows to `is_named=1` means no existing avatar changes — only *new* unnamed groups get the icon. That's the conservative choice (existing names can't be re-classified as auto-vs-user after the fact), and the migration comment explains exactly why the ambiguity forces it. The "no render-version bump" follows correctly from this and is reasoned out rather than hand-waved.
- **The public endpoint is scoped to a static constant, deliberately.** Exposing only the fixed palette (design tokens) — no per-group data — is what makes unauthed access correct, and the content-ETag means the web client's local preview can't drift from the server PNG. The threat surface was thought through (no input, bounded output).
- **The `is_named` inference is computed before the member-concat fallback**, so the flag reflects the user's actual intent at create time rather than being inferred from the final (possibly auto-generated) name. That ordering is the subtle part and it's right.

- CI `check-sprint` — admin Sprint-field gate, independent of the code. Test pending; Build/Vet/Lint/i18n green, vet (compiles tests) clean locally.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#500 Review Report
**Head SHA:** 65d72bec4cf35ab6c1346b13ebffb4733e8aab53
**PR:** feat(group): user-named default avatar (is_named) + public avatar color palette endpoint

## 1. 验证结论

- ✅ **is_named column + migration** — `20260629000001_group_is_named.sql` uses `INFORMATION_SCHEMA` guard for reentrancy; adds `TINYINT NOT NULL DEFAULT 0`; backfills existing rows to 1 (conservative: preserves current avatar behavior). `db.go:597` Model struct has `IsNamed int`. (`modules/group/sql/20260629000001_group_is_named.sql:13-25`, `modules/group/db.go:597`)

- ✅ **CreateGroup is_named inference** — `isNamedVal` computed from `strings.TrimSpace(req.Name) != ""` BEFORE the member-concat fallback overwrites `groupName`. Inserted via `InsertTx` which uses `util.AttrToUnderscore(m)` (reflects all struct fields). (`modules/group/service.go:1113-1117, 1168`)

- ✅ **UpdateGroupInfo rename sets is_named=1** — `groupModel.IsNamed = 1` at `service.go:1882`; persisted by `DB.UpdateTx` which now includes `is_named` in its explicit `SetMap` (`db.go:283`). All 5 other `UpdateTx` callers (disband, invite-setting, manager status/forbidden, settings) load the group from DB first and only modify specific fields, so `is_named` is preserved as-is.

- ✅ **writeGroupDefaultAvatar text priority** — `api.go:447-455`: `avatar_text > (is_named==1 → GroupNameText(name)) > (text stays "" → icon)`. Data flow: `QueryWithGroupNo` uses `SELECT *` → `Model.IsNamed` populated from `is_named` column → consumed in the if-else chain. Verified correct.

- ✅ **Palette endpoint (GET /v1/group/avatar_palette)** — Public route on `openGroup` (`/v1/group`, no auth middleware). Returns static design tokens from `avatarrender.PaletteHex()`. Content-derived weak ETag + `max-age=300, must-revalidate` + 304 support. No DB query, no sensitive data. Route at `/v1/group/avatar_palette` does not conflict with any wildcard route.

- ✅ **ETag/cache identity** — Named groups use `group-name-v4` factor set (with text); auto-named groups use `group-icon-v3` (no text). Different mode-version strings ensure ETag changes on is_named flip without a version bump. Correct reasoning.

- ✅ **DB.Update (non-Tx) exclusion of is_named** — Intentional and correct. `DB.Update` is used only for settings/forbidden updates that load the group from DB first. Since `is_named` is not in the explicit column list, the DB value is preserved.

## 2. 发现的问题（P0/P1/P2）

**No P0 or P1 findings.**

### P2-1: Test coverage gap — is_named=1 + custom_color + no custom_text

**diff-scope:** new (introduced by this PR's test additions)

`TestGroupAvatarGetCustomColorIconNoText` (`avatar_get_test.go:129-149`) inserts `Model{Name: "有名群", AvatarColor: intPtr(7)}` without setting `IsNamed`, so Go zero-value gives `IsNamed=0`. The test correctly validates the `is_named=0 + custom_color → colored icon` path. However, the complementary case (`is_named=1 + custom_color + no custom_text → name-text rendered in that custom color`) is not explicitly tested. The existing `TestGroupAvatarGetNamedRendersNameText` covers `is_named=1` without custom color (uses seed-derived color).

**Impact:** Coverage gap only — the production code path is correct (traced through `api.go:447-462`). Low risk.

### P2-2: Migration backfill — theoretical DDL/DML race window

**diff-scope:** new (introduced by this PR's migration)

Between `ALTER TABLE ADD COLUMN is_named DEFAULT 0` and `UPDATE group SET is_named = 1` within the stored procedure, concurrent INSERTs from old code could produce rows with `is_named=0` that the subsequent UPDATE sets to 1 (incorrectly marking auto-named groups as named). In practice this is not exploitable: migrations run at startup before the server accepts traffic, and the window is sub-millisecond within a stored procedure.

**Impact:** Theoretical only. No action needed.

## 3. 建议

- Consider adding one test: `is_named=1 + custom_color + no custom_text` → asserts name-text rendered with the custom color style. This locks the full three-factor interaction.
- Minor: `TestGroupAvatarGetCustomColorIconNoText` names the group "有名群" (literally "named group") but `IsNamed=0`. Consider renaming to "自动名群" or explicitly setting `IsNamed: 0` with a comment to avoid confusion.

## 4. 额外发现

None.

## 5. 数据流回溯

|---|---|---|
| `groupInfo.IsNamed` in `writeGroupDefaultAvatar` | `QueryWithGroupNo` → `SELECT *` → `Model.IsNamed` from `is_named` column | ✅ |
| `isNamedVal` in `CreateGroup` | `strings.TrimSpace(req.Name) != ""` computed before concat fallback | ✅ |
| `groupModel.IsNamed = 1` in `UpdateGroupInfo` | Explicitly set when `req.Name != nil`; persisted by `DB.UpdateTx.SetMap["is_named"]` | ✅ |
| `PaletteHex()` → JSON response | `palette[]` / `groupFillPalette[]` / `groupIconBackPalette[]` (hardcoded arrays) → `fmt.Sprintf("#%02X%02X%02X")` | ✅ (static, no runtime variability) |
| ETag mode selection (`group-name-v4` vs `group-icon-v3`) | `renderable` flag derived from `text` which depends on `is_named` branch | ✅ (different mode strings → different ETag factor sets) |
| `DB.UpdateTx` callers (non-rename paths) | All load group from DB first → `IsNamed` preserves current DB value | ✅ |

## 6. 盲点 checklist（R5）

- **C1 双路径 parity:** `CreateGroup` (insert) sets `is_named`; `UpdateGroupInfo` (rename) sets `is_named=1`; `UpdateTx` persists it; `DB.Update` (non-Tx) intentionally excludes it. Disband/status/forbidden paths preserve it. ✅ Clear.
- **C2 control-flow ordering / 嵌套复用:** `writeGroupDefaultAvatar` text priority chain is linear (no nesting). `GroupNameText` is a pure function with no side effects. No security controls involved. ✅ Clear.
- **C3 授权边界 ≠ 能力边界:** `GET /v1/group/avatar_palette` is intentionally public (static design tokens, no sensitive data, consistent with already-public avatar render endpoint). No capability exposure. ✅ Clear.
- **C4 授权生命周期 / 容器-成员状态级联:** Not applicable — no auth changes in this PR. N/A.
- **C5 build/note 通过 ≠ 运行期路径正确:** No build artifacts, extensions, CLI tools, or relative-path changes. Standard Go server. ✅ Clear.
- **C6 治理/策略/安全文档自洽性:** `.octospec/tasks/group-avatar-icon-default/brief.md` is a task spec, not a governance/security document. No policy conflicts. ✅ Clear.

## 7. 跨轮 blocker 复检（R6）

N/A — first review.

---

[Octo-Q] verdict: APPROVE

The PR is well-designed with clean data flow. The `is_named` flag lifecycle (create → persist → render → rename) is correctly implemented and thoroughly tested. The public palette endpoint follows existing patterns (no auth, ETag + 304, cache headers). Two minor P2 suggestions for test completeness.

OctoBoooot (changes requested):
## Re-review (verdict correction): #500 @ 65d72bec — REQUEST_CHANGES

Correcting my earlier COMMENT on this same SHA. Concurring with @Jerry-Xin's 🔴 — byte-verified, and it overturns my "migration is reentrant + reversible ✅" praise. I checked that a clean re-run doesn't error; I did NOT trace a failure *between* the guarded steps. That was the gap.

**🔴 Migration backfill is not recovery-safe** (`sql/20260629000001_group_is_named.sql:18-21`):
```
IF NOT EXISTS(column is_named) THEN
    ALTER TABLE ADD COLUMN is_named ... DEFAULT 0   -- implicit COMMIT, not rollback-able
    UPDATE `group` SET is_named = 1                 -- backfill
END IF
```
MySQL `ALTER TABLE` auto-commits. If the backfill `UPDATE` fails *after* the ALTER commits — crash, lock timeout, or deadlock, which is exactly the long-lock risk I flagged on a large `group` table — the retry hits `IF NOT EXISTS(column)`, finds the column now EXISTS, and skips the whole block. The backfill never re-runs. Existing rows stay at `DEFAULT 0` = "auto-named → two-person icon," silently flipping existing groups' avatars to the icon — the precise opposite of the migration's own "不改变既有头像" guarantee.

The column-existence guard is right for the (idempotent) ALTER but wrong for the backfill. Fix options: run the backfill unconditionally outside the column-existence guard (it's idempotent — `SET is_named=1` is safe to repeat), or guard the backfill on its own sentinel (a column default that means "not yet backfilled"), or split into two migrations.

This upgrades my verdict on 65d72bec from COMMENT to **REQUEST_CHANGES**. My two prior minors stand (swagger missing the new endpoint; full-table UPDATE lock — which is also the trigger for this blocker). The non-migration parts I verified remain sound (is_named inference, public endpoint safety, no-version-bump reasoning). My mistake was grading "doesn't error on re-run" as "recovers correctly after a partial failure" — they're different, and for a multi-step guarded migration the implicit-commit boundary between ALTER and UPDATE is exactly where they diverge.

an9xyz:
Thanks all — addressed in `38215cd`.

**🔴 Migration recovery-safety** (@Jerry-Xin / @OctoBoooot) — fixed. The backfill no longer lives inside the column-exists guard. Rewritten as a nullable-sentinel three-step that converges after a partial failure across the implicit-commit boundary:

1. `ADD COLUMN is_named TINYINT NULL` (no default) → existing rows = `NULL` ("not yet backfilled"), guarded on column-not-exists.
2. `UPDATE \`group\` SET is_named = 1 WHERE is_named IS NULL` — run **unconditionally**, idempotent, only touches not-yet-backfilled rows. A crash anywhere (including after the ALTER auto-commits) re-runs this on retry and completes the backfill; it can't clobber already-set 0/1 values.
3. `MODIFY COLUMN is_named TINYINT NOT NULL DEFAULT 0` for new inserts, guarded on `IS_NULLABLE = 'YES'`.

**🟡 Swagger** (@OctoBoooot) — added the public `GET /group/avatar_palette` path + `avatarPaletteResp`/`avatarPaletteColorResp` schemas, and refreshed the create-time `avatar_text` and avatar-GET descriptions to the `is_named` fallback rule (named → name first-2; auto-named/empty → two-person icon).

**🔵 Tests** (@mochashanyao Octo-Q P2) — renamed the misleading `有名群`/`IsNamed=0` fixture to `自动名群` with an explicit `IsNamed: 0`, and added `TestGroupAvatarGetNamedCustomColorRendersNameText` locking the `is_named=1 + custom color + no text` interaction.

**On the lock-duration note** (@OctoBoooot, non-blocking): kept as a single statement intentionally. The backfill `UPDATE ... WHERE is_named IS NULL` touches existing rows exactly once, and migrations run at startup before the server accepts traffic, so there's no concurrent contention; batching would add stored-proc loop complexity for a one-time deploy cost. Happy to batch if the prod `group` table is large enough to warrant it — let me know.

`go build` + `go vet` clean; swagger parses. Integration tests (MySQL/Redis/WuKongIM) run in CI.

---

OctoBoooot (approved):
## Review: feat(group): user-named default avatar (is_named) + public avatar color palette endpoint (#500) — delta @ 38215cd1

**Verdict:** Approve — the migration recovery-safety blocker is fixed correctly, and both minors (swagger, lock) are addressed. CI is now green except `check-sprint` (admin gate); Build/Vet/Lint/i18n all pass.

Delta from prior REQUEST_CHANGES @ `65d72bec` (I'd corrected my initial COMMENT up to CR on the migration blocker): 1 commit "fix(group): recovery-safe is_named migration + swagger palette/is_named docs".

### Resolved — byte-verified

- ✅ **Migration is now recovery-safe** (the blocker — concurred by me + Jerry-Xin) — rewritten to the three-step nullable-sentinel pattern: (1) `ADD COLUMN is_named TINYINT NULL` (no default → existing rows = NULL "not yet backfilled"); (2) `UPDATE ... SET is_named = 1 WHERE is_named IS NULL` (idempotent, only un-backfilled rows); (3) `MODIFY ... NOT NULL DEFAULT 0` guarded on `IS_NULLABLE='YES'`. I traced every failure/recovery path:
  - crash after Step-1 ALTER commits → retry backfills the NULLs ✓
  - crash mid Step-2 UPDATE → retry's `WHERE is_named IS NULL` catches the remaining rows (the exact case the old version stranded) ✓
  - crash after Step-2 → Step-3 `IS_NULLABLE=YES` still true → tightens ✓
  - clean re-run after success → Step-2 no-op, Step-3 guard false → skip ✓
  The implicit-commit boundary between ALTER and UPDATE no longer strands existing rows at `is_named=0`. The backward-compat contract ("no existing avatar changes") now holds even under a partial-failure retry.
- ✅ **Swagger documents the new endpoint** (minor) — `/group/avatar_palette` path + `avatarPaletteResp` definition added (`:1248`, `:1330`); `is_named` semantics documented across the avatar schemas. The published contract is complete now.
- ✅ **Full-table backfill scoped** (minor) — the UPDATE is now `WHERE is_named IS NULL`, so it only touches un-backfilled rows; the commit notes it runs pre-service with no concurrent writes. Acceptable for a one-time migration.

`go vet` clean (compiles tests); CI Build/Vet/Lint/i18n green, check-sprint now passes too.

### Praise

- **The migration fix is the textbook recovery-safe pattern, and the rationale is written into the file.** The new comment explains *why* the old guarded-block approach was unsafe (ALTER implicit-commits, so a backfill failure + column-exists guard = skipped backfill) and how the nullable sentinel fixes it (idempotent `WHERE IS NULL` backfill that any retry converges). That's the correct mental model, documented so the next migration author doesn't repeat the original mistake — exactly the failure mode this round caught.
- **The fix addressed the whole finding set in one focused commit** — recovery-safety, swagger endpoint, and the lock-scoping concern, each mapped to its review point. No scope creep, no churn on the parts that were already sound (is_named inference, public-endpoint safety untouched).

- CI `check-sprint` — admin Sprint-field gate, independent of the code.
- The non-migration parts verified in the prior round (is_named create/rename inference, public palette endpoint safety, no-render-version-bump reasoning) are unchanged and remain sound.

yujiawei (changes requested):
# Code Review — PR #500 (octo-server)

Reviewed at head SHA `65d72bec4cf35ab6c1346b13ebffb4733e8aab53` against merge-base `f3b2a785`. This was reviewed as a security-sensitive change with an independent multi-angle pass (lifecycle / endpoint security / migration / cache) plus adversarial verification of each finding against the code at head.

## 1. Spec compliance

**Spec: ❌**

- **Missing — the `is_named` lifecycle guarantee is broken on the combined name+invite update path.** The PR's stated contract is "rename (`UpdateGroupInfo`) sets `is_named=1`, persisted." That guarantee does not hold when a single `PUT /v1/groups/:group_no` payload contains **both** `name` and `invite`. The handler reads the group once into `group` (`modules/group/api.go:1007`), then:
  1. the name branch calls `UpdateGroupInfo`, which correctly commits `name=<new>` and `is_named=1` (`modules/group/service.go:1879-1880`, `:1894`);
  2. the invite branch then writes the **stale** in-memory `group` (still old name, `is_named=0`) through `UpdateTx` (`modules/group/api.go:1118`), whose `SetMap` now includes `is_named` (`modules/group/db.go:282-283`).
  The second write overwrites the first. A freshly renamed group submitted together with an invite toggle reverts to `is_named=0` and renders the two-person icon instead of the name's first-2 glyphs — exactly the behavior this PR exists to prevent. (Note: `name` is also in the same `SetMap` and is likewise reverted, but that `name` clobber is pre-existing; this PR newly extends the fault to `is_named`.) See Quality P1 below for the fix.

- Extra: none. The diff stays within scope — no palette value/order change, no avatar_text/avatar_color validation or upload change, no errcode added, no render-version bump (all matching the linked brief's out-of-scope list).

- Deviation: the missed write-path above is the only spec deviation.

Everything else in the acceptance list checks out: create-time inference (`service.go:1163`, `IsNamed: isNamedVal`), the render priority `avatar_text > is_named=1 name > icon` (`api.go:447-456`), the palette endpoint contract and `colors[0].main == "#14C0FF"` (`pkg/avatarrender/palette.go:13`), and the matching tests.

## 2. Code Quality

### P1 — combined name+invite update reverts the rename (`api.go:1007`, `:1095-1118`, `db.go:282`)
Root cause: `groupUpdate` loads the row once and then performs two independent writes (the `UpdateGroupInfo` service call, then the invite-branch `UpdateTx`). The second write is built from the pre-rename snapshot. Since `UpdateTx.SetMap` writes the full `name` + `is_named` (and now this PR adds `is_named` to it), the invite write silently rolls back the rename. Suggested fixes (any one):
- have the invite branch update only the `invite`/`version` columns (column-specific update, mirroring how `updateAvatarCustom` does column-level writes to avoid read-modify-write races), or
- re-read the group after `UpdateGroupInfo` before the invite write, or
- fold the invite change into the same service/transaction as name/notice.

This combination is a supported input (the handler parses `name`, `notice`, `invite` independently and applies each), so it is reachable, not theoretical. Please also add a regression test covering `{name, invite}` in one payload asserting `is_named==1` and the new name both persist.

### P2 — unbounded backfill `UPDATE` diverges from house style (`sql/20260629000001_group_is_named.sql:21`)
`UPDATE \`group\` SET \`is_named\` = 1;` has no `WHERE`. Every other backfill migration in the repo bounds its `UPDATE` (12/12, including the same-table sibling `20260615000001_group_name_widen.sql:11`). The end state here is correct and the statement is idempotent (it sits inside the `IF NOT EXISTS column` guard, so it never re-runs), so this is a consistency/hygiene nit rather than a correctness bug — but on a core table a single full-table rewrite is worth aligning to the established pattern.

### P2 — rolling-deploy ordering can mis-flag named groups (`sql/20260629000001...`, `db.go:31`)
Migrations run in-process at pod startup (`pkg/db/mysql.go`), and the column DEFAULT is `0` while the backfill sets existing rows to `1`. The old binary's `Model` omits `is_named`, and `InsertTx` derives its column list reflectively (`util.AttrToUnderscore`), so an **old** pod that creates a *named* group after the backfill commits persists `is_named=0` → that group renders the icon. Impact is transient (deploy window only), cosmetic (avatar), and self-healing (a later rename or custom `avatar_text` fixes it). Worth a note in the deploy runbook (run migration before rolling new binaries) rather than a code change.

### P2 — `is_named` not set on other named-group insert paths (`service.go:201`, `event.go:173`, `event.go:315`)
`Service.AddGroup`, the system-group creation, and the org/dept creation handler all insert a `Model` with a real `Name` but never set `IsNamed` (defaults to 0). These are **not** live avatar regressions: org/dept/system groups are served static PNGs by the prefix-gated branches in `avatarGet` (`api.go:359-394`) and never reach the `is_named` render path, and `AddGroup` has no production caller (test-only). Still, in a PR that tightens the `is_named` semantic it is a lifecycle-completeness gap; consider centralizing group construction or setting `IsNamed` explicitly so the invariant doesn't silently rot if one of these paths later starts feeding the default-render branch.

### P2 (optional) — rename to empty/whitespace keeps `is_named=1` (`service.go:1874-1880`)
`UpdateGroupInfo` sets `is_named=1` whenever `req.Name != nil`, without checking for empty/whitespace. If a client can clear the name, the row keeps `is_named=1` with an empty name. Today `GroupNameText("")` falls through to the icon so there's no visible breakage, but the flag's semantic ("1 = explicit user name") is violated. Low priority; flag only if empty rename is reachable.

### Verified clean
- **Public endpoint `GET /v1/group/avatar_palette`** — mounted unauthenticated consistently with the sibling public `avatarGet` route; returns only static design-token hex values (no DB, no params, no PII, no reflected input); still covered by the global per-IP limiter; no route conflict (static segment; the only wildcard under `/v1/group` is `/invites/:invite_no`); ETag/304 handled correctly with headers set before the short-circuit; response is the cheapest possible (fixed 10-entry slice, CRC32 over constants) — no DoS angle. The `c.Status(http.StatusNotModified)` 304 pattern matches the existing avatar endpoint (`api.go:477`).
- **`UpdateTx` cross-row clobber** — all six `UpdateTx` callers load the full row via `QueryWithGroupNo`/`getGroupInfo` (`SELECT *`, `db.go:262`), so adding `is_named` to `SetMap` round-trips safely. The only failure is the *intra-request* sequencing in P1, not a partial-column load.
- **ETag / no-version-bump claim** — sound. Within a render version, served bytes are a deterministic function of (mode, style, text) and the ETag factors fully capture those; `is_named` flips and renames change the version segment and/or text factor so clients revalidate. `TestGroupAvatarGetPinsRenderVersion` pins both modes. No bytes-change-but-ETag-stable poisoning path.
- **Migration reentrancy / down path** — backfill inside the `IF NOT EXISTS` guard, `DROP PROCEDURE` bracketing, and the down migration are correct.

> Note: endpoint-level tests requiring MySQL/Redis/WuKongIM were not executed in this review environment (they run in CI per the brief); the assertions were read and confirmed to check the right invariants. `go build ./modules/group/... ./pkg/avatarrender/...` is clean here.

## 3. Overall verdict

**REQUEST_CHANGES** — the design and the bulk of the implementation are solid, but the combined name+invite update path (P1) reverts a rename and undercuts the exact `is_named` persistence guarantee this PR is built around. Fix that one write-path (and add the regression test); the P2s are non-blocking and can be addressed at your discretion.

## 4. Suggestions
- Fix P1 by making the invite branch a column-specific update (only `invite` + `version`), which also removes the pre-existing `name`-revert fault on the same path.
- Add `WHERE is_named = 0` to the backfill for house-style consistency.
- Consider a single internal constructor for group `Model` inserts so `is_named` (and future flags) can't be omitted by one path.

Jerry-Xin (changes requested):
This PR is relevant to octo-server, but one creation-path bug would make newly created non-member-concat groups render the wrong default avatar.

**🔴 Blocking**
- 🔴 Critical: `is_named` is only set in `Service.CreateGroup`, but this module has other real group creation paths that insert named groups directly with `Model.IsNamed` left at the Go zero value. Because `DB.Insert`/`InsertTx` writes all struct fields, these rows are persisted as `is_named=0`, so fresh system/org/dept groups render the two-person icon instead of their explicit configured name. Affected paths include [modules/group/event.go](modules/group/event.go:173) for the system group, [modules/group/event.go](modules/group/event.go:315) for org/dept group creation, and [modules/group/service.go](modules/group/service.go:201) for `AddGroup`. The migration protects existing rows, but these new rows are created after migration and bypass the inference logic. Set `IsNamed: 1` for explicit-name direct creation paths, or centralize group construction so only member-concat fallback groups get `is_named=0`; add regression coverage for at least org/dept or system group creation.

- 🔴 Critical: `is_named` is clobbered when `name` and `invite` are updated in the same request. `DB.UpdateTx` (modules/group/db.go:280) writes `is_named: model.IsNamed` unconditionally in its SetMap. In the group-update handler, a rename goes through `groupService.UpdateGroupInfo` (which loads its own model and commits `is_named=1`), but the `invite` change is then applied separately via `g.db.UpdateTx(group, tx)` (modules/group/api.go:1118) using a `group` model that was loaded *before* the rename — so its `IsNamed` is the stale pre-rename value (e.g. 0). That second write overwrites the just-set `is_named=1` back to the stale value, silently reverting the group's avatar to icon mode despite the user-set name. Reachable via a single PATCH/update carrying both `name` and `invite`. Fix: re-read the model (or carry `is_named` forward) before the invite `UpdateTx`, or have the invite path use a scoped column update that doesn't touch `is_named`. Add a regression test for name+invite in one request. (Credit: yujiawei.)

**💬 Non-blocking**
- 🟡 Warning: `GroupResp` still does not expose `is_named` ([modules/group/service.go](modules/group/service.go:796)). If the web follow-up needs local preview parity when clearing `avatar_text` or changing `avatar_color`, the palette endpoint alone is not enough to decide name-text vs icon for an existing group with a non-empty name. Consider exposing a read-only flag if clients are expected to reproduce server selection locally.
- 🔵 Suggestion: The render-version pin test now pins custom-text mode, but not the named-group `is_named=1` name-text path. The behavior is covered by byte comparison tests, but a small ETag assertion for named groups would better protect the cache contract.

**✅ Highlights**
- The migration’s nullable-sentinel backfill is recovery-aware and avoids changing existing avatars.
- The avatar render tests cover the main priority order: custom text, named group text, auto-named icon, and custom color combinations.
- The palette endpoint is additive, cacheable, and derives from the server palette source instead of duplicating constants.

OctoBoooot (commented):
Byte-verified @yujiawei's findings at 38215cd1 — the two 🔴s don't hold for the actual code paths; I'm keeping my APPROVE, with one honest caveat.

🔴#1/#2 (is_named rolled back / invite update overwrites rename): every `UpdateTx` caller loads the **full row first**, and the load uses `Select("*")`:
- rename → `service.go:1856` `QueryWithGroupNo` (`SELECT *`, includes is_named)
- forbidden/settings → `api.go:2063` `getGroupInfo` → `QueryWithGroupNo` (same)
`UpdateTx`'s SetMap writes `is_named: model.IsNamed`, and since the model was loaded with the true current value, a non-rename update writes it **back unchanged** — no rollback to 0. The non-tx `Update()` (settings-action path, db.go:295) doesn't even include `is_named` in its SetMap, so it can't clobber it either. A sequential "rename then toggle invite" preserves is_named=1. I couldn't construct the lost-update on the sequential path yujiawei describes.

The ONE real caveat: a genuinely **concurrent** rename + other-update race is last-writer-wins on the whole row (both load the old row, one sets is_named=1, the other writes back the old value). But that's a pre-existing whole-row race affecting name/notice/forbidden/every field equally — not an is_named-specific defect, and not introduced by this PR. If the team wants per-field update safety that's a separate, larger change.

🟡#3 (backfill missing WHERE): already fixed in this SHA — the migration was rewritten to `UPDATE ... SET is_named=1 WHERE is_named IS NULL` (recovery-safe 3-step). Looks like that note was against the prior SHA.

So no verdict change — APPROVE stands. If I'm missing a specific sequential interleaving where is_named zeroes, point me at the two endpoints + order and I'll re-trace.

OctoBoooot (changes requested):
## Verdict correction: #500 @ 38215cd1 — APPROVE → REQUEST_CHANGES

Concurring with @Jerry-Xin's 🔴 — byte-verified, and it overturns my APPROVE. My mistake: I verified `CreateGroup` sets `is_named` and didn't grep the *other* group-creation paths. There are four, and three don't set it:

🔴 **`is_named` not set in all creation paths** → new named groups get `is_named=0` (icon) instead of name-text:
- `service.go:1163` `CreateGroup` — sets `IsNamed: isNamedVal` ✓ (the one I checked)
- `service.go:201` `AddGroup` — sets `Name: model.Name`, **no `IsNamed`** → 0
- `event.go:173` `handleRegisterUserEvent` — system group, real `Name`, **no `IsNamed`** → 0
- `event.go:315` `handleOrgOrDeptCreateEvent` — org/dept group with `req.Name`, **no `IsNamed`** → 0

The org/dept path is the clearest user-facing bug: a department group created with a real name renders the two-person icon instead of its name's first-2 glyphs — the exact behavior the feature exists to produce, broken on creation. The migration backfilled *existing* rows to 1, which is why this isn't visible on old data — but these are NEW post-migration inserts defaulting to the column's `DEFAULT 0`. `DEFAULT 0` is correct only for genuinely-unnamed groups; these paths have real names.

Fix: set `IsNamed` at each creation path from the same "was a real name provided?" logic `CreateGroup` uses (the system/org-event groups have names, so `1`).

Jerry-Xin's two 🟡s are also fair: expose `is_named` in `GroupResp` so clients can predict the avatar locally; add a named-group ETag assertion. Non-blocking, fold into the same fix.

This corrects my verdict on 38215cd1 to **REQUEST_CHANGES**. The migration recovery-safety fix (my prior blocker) IS correctly resolved — verified — but this creation-path gap is a separate, real blocker I missed by checking only the primary create path. The yujiawei is_named-rollback claim I separately traced does NOT hold (full-row load preserves it); this Jerry-Xin finding is the real is_named defect.

OctoBoooot (commented):
Correcting my earlier comment: I previously wrote that @yujiawei's `is_named`-rollback finding "doesn't hold." I was wrong — I tested the wrong interleaving. Re-traced, and it IS a real bug on the same-request name+invite path:

- `api.go:1007`: `group, err := g.getGroupInfo(groupNo)` — loads the row with current `is_named` (e.g. 0).
- `~:1087`: `hasName` → `g.groupService.UpdateGroupInfo(...)` loads its OWN fresh model, sets `is_named=1`, commits.
- `~:1098-1118`: `hasInvite` → mutates the **stale** `group` from :1007 (still `is_named=0`) and calls `UpdateTx(group, tx)`, whose blanket SetMap writes `is_named: 0` — **clobbering the =1 the rename just committed.**

So a single request with both `name` and `invite` rolls `is_named` back to 0 → avatar silently reverts to icon mode. My earlier adjudication only checked the standalone-update paths (which load fresh via `QueryWithGroupNo *` and ARE safe); I missed that this dual-update handler reuses a pre-rename in-memory model for the invite write. yujiawei's "name和invite同时更新" was precisely that path — my mistake for testing sequential separate requests instead of the same-request combination.

Fix: before the invite `UpdateTx`, either re-read the model (so it carries the just-committed is_named=1), or have the invite path use a column-scoped update that doesn't touch is_named.

Net: this is a real second blocker alongside the creation-path is_named gap. My verdict was already REQUEST_CHANGES, so the gate is unchanged — but my "doesn't hold" was incorrect and I wanted it corrected on the record rather than left standing.

an9xyz:
Thanks @yujiawei @Jerry-Xin @OctoBoooot — both blockers fixed in `22f2ce2`. The two reviews conflicted on a couple of points, so here's the code-verified adjudication alongside the fixes.

## 🔴 P1 — combined `{name, invite}` reverts rename + `is_named` (yujiawei) — **REAL, fixed**

Confirmed by tracing the handler (and @OctoBoooot's re-trace agrees): `groupUpdate` loads the row once at `api.go:1007`; the name branch commits `name`+`is_named=1` via `UpdateGroupInfo` (its *own* fresh load); then the invite branch wrote the **stale** snapshot back through the full-column `UpdateTx` → clobbering both. The earlier "doesn't hold" was testing *two separate sequential requests* (where the invite call re-loads fresh) — the bug is *one request carrying both keys*. This PR is what extended the pre-existing `name`-clobber to `is_named` (by adding `is_named` to `UpdateTx.SetMap`).

**Fix:** invite branch now uses a column-scoped `DB.UpdateInviteTx` (writes only `invite`+`version`); the handler no longer caches the full row (existence-check only). This is @yujiawei's suggestion #1 — it also removes the pre-existing `name`-revert and the read-modify-write race on that path.

## 🔴 Creation paths persist `is_named=0` (Jerry-Xin / OctoBoooot) — **REAL data bug, fixed; but one nuance**

`is_named` was only set in `CreateGroup`. `event.go` system-group + org/dept and `Service.AddGroup` left it at the Go zero value. **Fixed:** `IsNamed: 1` on the system/org/dept inserts (explicit configured names); `AddGroup` infers from `TrimSpace(name)!=""`.

Nuance worth recording (@yujiawei is right here): system/`org_`/`dept_` groups are served **static PNGs** by the prefix-gated branches in `avatarGet` (`api.go:360-394`) and return *before* the `is_named` render path — so this was **not a live avatar regression today** (a dept group does not currently render the icon; it renders `dept_avatar.png`). But the persisted flag was still wrong and fragile — removing any prefix-branch would silently break it — so it's fixed as data-correctness.

## 🟡/🔵 Non-blocking (folded in)
- **`GroupResp.is_named`** exposed (read-only, both populators) so web can locally predict name-text vs icon when clearing `avatar_text`. Additive/backward-compatible.
- **Named-group ETag assertion** added (weak ETag + 304 round-trip).
- **Backfill `WHERE`** (yujiawei P2): already done in `38215cd` — the migration is now `WHERE is_named IS NULL` (that note was against the prior SHA).

## Tests
`TestRenameThenInviteUpdate_KeepsIsNamed` reproduces the two-write sequence (rename → invite) at the service/DB layer — the invite branch's real path calls `EventBegin`, which `testutil.NewTestServer` doesn't initialize (the repo skips event-handler unit tests for this reason; see the NOTE in `service_test.go`), so the regression is locked at the layer where the bug actually lived (full-row vs column-scoped write). Plus `TestAddGroup_SetsIsNamed`.

`go build`/`go vet` + i18n guards green.

Deferred (P2, non-blocking, happy to do if you want): rolling-deploy ordering note in the runbook; empty/whitespace rename keeping `is_named=1` (currently harmless — `GroupNameText("")` already falls through to the icon).

---

OctoBoooot (dismissed):
## Review: feat(group): user-named default avatar (is_named) + public avatar color palette endpoint (#500) — delta @ 22f2ce22

**Verdict:** Approve — both `is_named` blockers from the prior round are fixed, byte-verified, and pinned with a regression test. CI red on nothing code-side (check-sprint + i18n green; Build/Test/Vet/Lint pending). The migration recovery-safety fix from the round before remains intact.

Delta from prior REQUEST_CHANGES @ `38215cd1`: 1 commit "fix(group): is_named on all creation paths + column-scoped invite update".

### Resolved — byte-verified

- ✅ **`is_named` now set on all creation paths** (Jerry-Xin's 🔴 / my corrected blocker) — `AddGroup` (`service.go:201`) computes `isNamed` from `strings.TrimSpace(model.Name) != ""` (matching `CreateGroup`); `event.go:176` (system group) and `event.go:319` (org/dept group) hardcode `IsNamed: 1` (correct — those always have explicit names). All four group inserts now set the field; no path defaults a named group to the icon.
- ✅ **Same-request name+invite no longer clobbers `is_named`** (yujiawei's 🔴, which I'd wrongly refuted then corrected) — the invite branch switched from the blanket `UpdateTx(group)` to `UpdateInviteTx(groupNo, invite, version, tx)`, whose SetMap writes **only** `invite` + `version`. So the `is_named=1` the rename committed survives; the stale-snapshot full-row overwrite is gone. Verified `UpdateInviteTx` does not touch `name` or `is_named`.
- ✅ **Pinned with a named regression test** — `TestRenameThenInviteUpdate_KeepsIsNamed` reproduces the exact same-request name+invite interleaving and asserts is_named survives, crediting the review. The bug can't silently return.

`go vet ./modules/group/` clean (tests compile). The local `modules/group` test failure is the pre-existing MySQL-dependency (`dial tcp 127.0.0.1:3306` — needs a live DB to *run*, compiles fine); CI's Test job is the authority and the logic is exercised there.

### Praise

- **The invite fix went column-scoped rather than re-reading the model.** Both fixes work, but `UpdateInviteTx` (write only the columns this operation owns) is the more robust one — it structurally prevents *any* field-clobber from a stale snapshot on this path, not just is_named. That's the right lesson drawn from a lost-update bug: narrow the write surface, don't just refresh the read.
- **Every creation path was swept, with intent encoded per-path** — `AddGroup` computes the flag (it takes arbitrary names), the event paths hardcode 1 (they always have real names), each with a comment naming why. That's the complete enumeration the prior round was missing, done deliberately rather than blanket-defaulting.
- **The regression test names the precise interleaving and the reviewers** — `TestRenameThenInviteUpdate_KeepsIsNamed` documents the same-request name+invite sequence that caused it, so the next person understands the hazard, not just the assertion.

- CI `check-sprint` now passes; Build/Test/Vet/Lint pending. The migration recovery-safety (3-step nullable sentinel) and swagger endpoint from the prior round are unchanged and remain sound.

Jerry-Xin (dismissed):
The PR is in scope for octo-server and the implementation looks aligned with the group/avatar backend behavior described.

💬 Non-blocking

🟡 Warning: Swagger is not fully synced with the new response contract. `GroupResp` now exposes `is_named` in [service.go](modules/group/service.go:806), but `definitions.group` in [api.yaml](modules/group/swagger/api.yaml:1386) does not document that field. The `avatar_text` response description also still says it falls back to group-name text in [api.yaml](modules/group/swagger/api.yaml:1470). This will confuse clients trying to implement local preview parity.

✅ Highlights

🔵 The `is_named` lifecycle is handled carefully across create, rename, direct `AddGroup`, and system/org/dept insert paths.

🔵 The migration uses a nullable sentinel plus unconditional backfill before enforcing `NOT NULL DEFAULT 0`, which avoids the common MySQL partial-DDL retry trap.

🔵 The invite update was narrowed to column-scoped persistence via `UpdateInviteTx`, which fixes the stale full-row write risk in combined name/invite updates.

🔵 The palette endpoint is simple, additive, cacheable, and backed directly by `avatarrender.PaletteHex()`.

Verification: I ran `go test ./pkg/avatarrender/...` successfully.

OctoBoooot (commented):
On @Jerry-Xin's swagger 🟡s — byte-checked against the current head (22f2ce22); the framing overstates it, but there's one real line worth tightening:

- `:323` (create req), `:531` (update req), `:1219-1220` (avatar endpoint summary) — these **already** reflect the is_named logic ("空/缺省时按 is_named 回退：命名群取群名前2字、自动名群双人图标"). Current, not stale. No "前4字" remains anywhere in the file.
- `:1472` (the `group` **response** definition's `avatar_text`) — this one says "空表示未自定义，按群名前 2 字派生" but **omits the is_named=0 → 双人图标 fallback**. That's the single incomplete description — worth aligning with the other three for consistency.

So it's a 1-line doc nit on the response schema, not the broad "response description still wrong / doesn't reflect latest logic" the message implies — 3 of 4 are correct and is_named-aware. Non-blocking; fold the `:1472` tweak into any follow-up. Verdict unchanged: APPROVE stands.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

yujiawei (approved):
# Code Review — PR #500 (octo-server)

Reviewed at head SHA `734494dd3cc84760bdc97aef9343ea49f80ed0d0` against merge-base `f3b2a785`. Treated as a security-sensitive change (public endpoint + schema migration), reviewed independently with cross-file verification. A second independent automated pass concurred (no blocking findings); a third independent leg timed out and is therefore recorded as **absent**, not as agreement.

Note: the head advanced from `22f2ce22` to `734494dd` during review. The only delta is one docs commit (`docs(group): sync swagger response schema with is_named`) that adds the `is_named` field to the `GroupResp` swagger schema and refreshes the `avatar_text` description. It closes a doc gap and changes no code; all findings below were re-anchored on the current head.

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly the stated scope (is_named-gated default avatar + public palette endpoint) and nothing beyond it.

- **Missing:** none. Both features land: `is_named` rule (`avatar_text` > `is_named==1` name first-2 > two-person icon) in `writeGroupDefaultAvatar` (`modules/group/api.go:441-452`), and the public `GET /v1/group/avatar_palette` (`api.go:151`, handler `api.go:531-549`).
- **Over-build:** none. The new `GroupResp.is_named` field (`service.go:806`) is read-only and additive, justified for client-side preview parity. The migration touches only the new column.
- **Deviation:** none. ETag policy intentionally bumps no render version; the mode switch (`group-name-v4` ↔ `group-icon-v3`) changes the ETag factor string on its own, so stale `If-None-Match` cannot match. Verified the factor strings differ by mode.

## 2. Code Quality

**Quality: Approved**

Prior-round blockers are fixed correctly at this head; cross-file dependencies verified.

**Verified (previously-flagged issues, now resolved):**
- **Migration recovery-safety** (`modules/group/sql/20260629000001_group_is_named.sql`): the backfill is no longer trapped inside the `ADD COLUMN` guard. Three-step nullable-sentinel pattern — (1) `ADD COLUMN ... NULL`, (2) unconditional `UPDATE ... SET is_named=1 WHERE is_named IS NULL` (idempotent, converges on partial-retry across the MySQL DDL auto-commit boundary), (3) guarded `MODIFY ... NOT NULL DEFAULT 0`. This preserves the "no existing avatar changes" guarantee even if a retry happens after the committed ALTER.
- **Combined `{name, invite}` update no longer reverts the rename / `is_named`** (`api.go:1006`, `db.go:294-305`): `groupUpdate` stopped caching the full row (existence-check only); the invite branch now uses the column-scoped `UpdateInviteTx` (writes only `invite`+`version`). The name branch (`UpdateGroupInfo`, fresh load) commits `name`+`is_named=1` and is no longer clobbered by a stale snapshot.
- **All creation paths set `is_named`**: `CreateGroup` (`service.go:1122-1126`, computed before the member-concat fallback overwrites the name), `AddGroup` (`service.go:201-205`, infers from `TrimSpace(name)!=""`), and the system/org/dept inserts in `event.go:176,319` (`IsNamed: 1`). Persistence verified: `Insert`/`InsertTx` use `util.AttrToUnderscore` which reflects struct field names → `IsNamed` maps to `is_named`, so the flag is actually written.

**Cross-file safety checks (no findings):**
- `UpdateTx` now writes `is_named` in its `SetMap`. All four other callers (`api.go:219` disband, `api.go:2062`, `api_manager.go:266,346`) load the row fresh via `QueryWithGroupNo`/`getGroupInfo`, and `QueryWithGroupNo` uses `Select("*")`, so the round-trip writes the same `is_named` value back (no clobber). The non-tx `Update` omits `is_named` from its `SetMap` entirely, so it preserves the column.
- Clearing a custom avatar via `updateAvatarCustom` (`db.go`) is column-scoped (only `avatar_text`/`avatar_color`/`version`/`updated_at`), so it does not touch `is_named` — a named group correctly falls back to name-text when text is cleared.

**Public endpoint (security-sensitive surface):** `avatarPalette` returns only static design tokens (`PaletteHex()`), takes no input, queries no data, and exposes nothing sensitive — consistent with the adjacent already-public avatar render endpoint. Content-derived weak ETag + `max-age=300, must-revalidate` + 304 handling are coherent.

P0/P1/P2: none.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved.

## 4. Suggestions (non-blocking)

- Existing rows are conservatively backfilled `is_named=1`, so any legacy group that happens to sit on a member-concat name will keep rendering name-text (not the icon) until renamed. This matches the documented product decision (don't retroactively change existing avatars); flagging only so it's a conscious choice, not a surprise.

## 5. Additional observations

- Verified build + vet are green at the current head for `modules/group/...` and `pkg/avatarrender/...`; `pkg/avatarrender` unit tests pass locally. DB-backed endpoint tests run in CI per prior increments.
- No sibling open PR touches `modules/group` in a way that would conflict at merge.

Jerry-Xin (approved):
The PR is in scope for octo-server and the implementation looks aligned with the group/avatar backend behavior described.

💬 Non-blocking

🟡 Warning: Swagger is not fully synced with the new response contract. `GroupResp` now exposes `is_named` in [service.go](modules/group/service.go:806), but `definitions.group` in [api.yaml](modules/group/swagger/api.yaml:1386) does not document that field. The `avatar_text` response description also still says it falls back to group-name text in [api.yaml](modules/group/swagger/api.yaml:1470). This will confuse clients trying to implement local preview parity.

✅ Highlights

🔵 The `is_named` lifecycle is handled carefully across create, rename, direct `AddGroup`, and system/org/dept insert paths.

🔵 The migration uses a nullable sentinel plus unconditional backfill before enforcing `NOT NULL DEFAULT 0`, which avoids the common MySQL partial-DDL retry trap.

🔵 The invite update was narrowed to column-scoped persistence via `UpdateInviteTx`, which fixes the stale full-row write risk in combined name/invite updates.

🔵 The palette endpoint is simple, additive, cacheable, and backed directly by `avatarrender.PaletteHex()`.

Verification: I ran `go test ./pkg/avatarrender/...` successfully.

Jerry-Xin (approved):
Re-review at 734494dd. All blockers across this PR's arc are resolved; this commit fixes the last non-blocking swagger note. APPROVING.

**✅ Blockers — all FIXED (byte-verified across rounds)**
- Migration recovery-safety: the backfill uses the three-step nullable-sentinel pattern (add `is_named` NULL → idempotent `UPDATE ... WHERE is_named IS NULL` → `MODIFY NOT NULL DEFAULT 0`), so an interruption between the auto-committed ALTER and the backfill resumes correctly on retry. The "conservative backfill to 1 / no existing avatar changes" contract holds.
- `is_named` on all creation paths: `AddGroup` (service.go) now computes `IsNamed` from name-presence (matching `CreateGroup`); the system group and org/dept group inserts (event.go) set `IsNamed: 1`. All four creation paths now persist the flag correctly, so newly created named groups no longer fall back to the two-person icon.
- Same-request name+invite clobber: the invite branch switched from the blanket `UpdateTx(group)` (which wrote `is_named: model.IsNamed` from a pre-rename in-memory model) to a column-scoped `UpdateInviteTx(groupNo, invite, version, tx)` that writes only `invite`+`version`. A single request carrying both `name` and `invite` no longer rolls `is_named` back to its stale value. Pinned by `TestRenameThenInviteUpdate_KeepsIsNamed`, which reproduces the exact interleaving.

**✅ This commit (delta 22f2ce22→734494dd, swagger-only, +4/-1)**
- `definitions.group` now documents the `is_named` field; the `avatar_text` response description now reflects the is_named fallback (named group → name first-2 glyphs; auto-named → two-person icon). Resolves the last 🟡.

**✅ Other**
- Public `GET /v1/group/avatar_palette` remains safe (unauthed by design, returns only static design tokens from `avatarrender.PaletteHex()`, no input/DB/injection surface).

CI: the only red across the arc has been `check-sprint` (governance/board gate, not a code defect); the local Test failure is the MySQL-dependency (tests compile under `go vet`; CI's containerized Test job is authoritative). Verdict: APPROVE.

OctoBoooot (approved):
## Review: feat(group): user-named default avatar (is_named) + public avatar color palette endpoint (#500) — delta @ 734494dd

**Verdict:** Approve (re-affirmed) — this delta is a documentation-only swagger sync; no code changed. It closes the last doc nit I flagged. All CI green. My prior APPROVE on the code (`22f2ce22`) carries forward.

Delta from prior APPROVE @ `22f2ce22`: 1 commit "docs(group): sync swagger response schema with is_named" — `swagger/api.yaml` only (+4/-1), no `.go` files.

### Verified

- ✅ **Doc-only, byte-confirmed** — the compare touches exactly one file, `modules/group/swagger/api.yaml`. YAML well-formed.
- ✅ **Closes the `:1472` incompleteness I pinned last round** — the `group` response `avatar_text` description now reads "空=未自定义，按 is_named 回退：命名群(1)取群名前 2 字…自动名群(0)回退双人图标" — the is_named=0→icon fallback I noted was missing is now there.
- ✅ **`is_named` exposed in the `group` response schema** (`:1398`) — addresses the earlier 🟡 (clients can now locally predict the default-avatar mode). The other three avatar_text descriptions were already is_named-aware; the published contract is now fully consistent.
- ✅ **All CI green** — Build/Test/Vet/Lint/check-sprint/i18n all pass (Test went green this round).

The code I byte-verified at `22f2ce22` — both is_named blockers fixed (creation-path omission + same-request name+invite clobber), migration recovery-safe (3-step nullable sentinel), public palette endpoint safe — is unchanged by this commit.

### Praise

- **The swagger sync was done as its own focused doc commit** rather than bundled into the code fix, and it covers both the response-description completeness and the new `is_named` field exposure — the published contract now matches the behavior end-to-end. The whole arc (migration recovery-safety → all-creation-paths is_named → column-scoped invite update → doc sync) was worked through systematically, each finding closed with a matching commit.

- Nothing outstanding. The #500 arc is complete from my side: every blocker byte-verified fixed, the contract documented, CI green.

PR #486, opened 2026-06-27
fix(avatarrender): transparent corners on rendered default avatars

an9xyz:
## Summary

Server-rendered default avatars (group text/icon and personal) painted an opaque white background outside the circle, so the PNG corners showed as white squares on any surface that doesn't clip the avatar to a circle (e.g. dark-theme list rows). This makes the corners transparent instead.

## Related Issue

N/A — post-merge follow-up to #478, reported from the test environment.

## Linked Spec

`.octospec/journal/shared/group-default-avatar.md` → "Post-merge 跟进: 透明四角 (group + user 头像)".

## Changes

- `RenderGroup`, `RenderIcon`, `Render` no longer fill the canvas with opaque `color.White` before drawing the circle. `image.NewRGBA` is zero-initialized to fully transparent, so the circle (fill + stroke + text/icon) is drawn on a transparent canvas and `png.Encode` now emits an RGBA (alpha) PNG with transparent corners.
- Drop the now-unused `image/draw` import in both files; update the doc comments.
- Covers **both group and user avatars** — `avatarrender.Render` is called in production only by `modules/user`, so one change keeps the two consistent.

## Why server-side transparency (not client-side masking)

The avatar's designed stroke ring commits it to a circular shape — clipping the PNG to a non-circle (rounded-square) would cut the ring. So one server change renders correctly on every surface without depending on each client to clip; clients that already clip to a circle are unaffected.

## Testing

- [x] Unit tests added/updated
- [x] Manually verified (composited onto dark + checkerboard backgrounds; corners show the background through)

Commands run:

- `go build ./...`
- `go test ./pkg/avatarrender/...`
- `go test ./modules/group/ -run TestGroupAvatarGet` and `go test ./modules/user/ -run Avatar`
- `golangci-lint run ./pkg/avatarrender/...`

Test changes: `TestRenderOpaque` → `TestRenderTransparentCorners` (four corners alpha=0, image not `Opaque`); the group and icon tests gain an outside-circle alpha=0 assertion; `inkBox` samples only the in-circle core so the font-size tests are unaffected. The group/user avatar endpoint byte-exact comparisons stay consistent (both sides share the render functions).

## COMPREHENSION

1. **What does this change actually do** to the load-bearing path?

[...]

PR #503, opened 2026-06-29
feat(group): default new groups to icon avatar; grandfather existing via is_named

an9xyz:
## Summary

Match the client: a **newly created** group defaults to the two-person icon — the group **name is never rendered as avatar text**; text appears only when the user sets a custom `avatar_text`. **Existing groups are grandfathered** so the change does not flip every historical group to an icon at once.

Implemented by changing **who gets `is_named=1`**, *not* the render rule. The render priority is unchanged from #500: `avatar_text` (custom) > `is_named==1` → name first-2 glyphs > two-person icon. `is_named` is **repurposed** from "user named it" to "**pre-cutover legacy group**" (it stays load-bearing — not deprecated).

## Related Issue

N/A — client-coordination change relayed by the frontend (2026-06-29). Repurposes the S2 behavior from #500.

## Linked Spec

`.octospec/tasks/group-avatar-name-no-text/brief.md` (supersedes `group-avatar-icon-default` S2). Load-bearing render-path/data change → COMPREHENSION filled below.

## Changes

- **`is_named` write-policy (the whole change):**
  - `CreateGroup` (`service.go`): new groups persist `IsNamed: 0` (removed the name-based computation) → default to the icon regardless of name.
  - `AddGroup` (`service.go`): `IsNamed: 0`.
  - `event.go` system-group + org/dept inserts: `IsNamed: 0` (served static PNGs; set to 0 to keep "`is_named=1` ⟺ legacy" clean).
  - `UpdateGroupInfo` rename (`service.go`): removed `groupModel.IsNamed = 1` → rename no longer flips the flag; it preserves the loaded value (legacy stays 1 and its name-text follows the new name; new stays 0).
- **Render rule unchanged** (`writeGroupDefaultAvatar`, `api.go`): still `avatar_text > is_named==1 name-text > icon`; only comments reworded to the legacy/new semantic. ETag factors unchanged.
- **Existing groups** keep `is_named=1` (already backfilled by migration `20260629000001`) → keep their current name-text avatar (grandfather). The grandfather behavior itself needs **no schema/data migration** — it reuses #500's existing backfill as the legacy marker.
- **One comment-only migration** `20260629000002_refresh_avatar_comments.sql` (added in a review-nit follow-up): `MODIFY COLUMN` refreshes the live `is_named` + `avatar_text` column COMMENTs to the post-pivot semantics. **No type/constraint/default/data change** (`TINYINT NOT NULL DEFAULT 0`, `VARCHAR(16) NOT NULL DEFAULT ''` preserved), reentrant via INFORMATION_SCHEMA guard, with a Down that restores the prior comments. The already-applied `20260629000001` / `20260625000001` files are left immutable; this only refreshes what schema introspection shows.
- **Docs:** comments in `modules/group/{api.go,db.go,service.go,const.go,event.go}` + `pkg/avatarrender/text.go` + `swagger/api.yaml` reworded to the legacy/new semantic; `GroupResp.is_named` re-documented (1=legacy/0=new). New octospec brief + changelog entry. (Follow-up commits also cleaned up the stale comments reviewers flagged.)
- **Tests** updated for the new write-policy (see below).

## Testing

- `go build ./...`, `go vet ./modules/group/` — clean.
- `golangci-lint run ./modules/group/...` — 0 issues.
- `make i18n-lint` — OK (no error-code/response changes).
- `go test ./pkg/avatarrender/...` — pass.
- Group endpoint tests (which also apply the new migration) require MySQL/Redis/WuKongIM — run in CI (no local infra).

- [ ] Manually verified (endpoint tests via CI; locally limited to build/lint/avatarrender unit tests — no MySQL/Redis in dev env)

## COMPREHENSION

   It changes only the persisted value of `is_named` for newly created groups (and stops rename from flipping it). New groups now store `is_named=0`, so the unchanged render rule (`avatar_text > is_named==1 name-text > icon`) takes them down the icon branch by default — the group name is no longer rendered as avatar text. Existing rows remain `is_named=1` (from #500's backfill) and continue to render their name-text. `is_named` becomes read-only after creation, meaning strictly "this group existed before the cutover."

   Behavior changes only for *new* groups (icon instead of name-text); existing groups are untouched, so there is no mass historical flip. No API/contract break: `is_named` is still persisted and exposed (`GroupResp.is_named`, now meaning legacy=1/new=0); create/update/palette endpoints and error codes are unchanged. The only migration is the **comment-only** COMMENT refresh (`20260629000002`) — no schema/data change. The P1 combined `{name, invite}` clobber fix from #500 (`UpdateInviteTx` column-scoped write) is retained — failure mode would be an invite update reverting a concurrent rename, which the column-scoped write prevents. Static-PNG groups (system/`org_`/`dept_`) and uploaded-image groups never reach the render path and are unaffected.

   `TestCreateGroup_Success` / `_AutoGenerateName` assert `is_named=0` for new groups (named and auto). `TestUpdateGroupInfo_RenameDoesNotChangeIsNamed` asserts rename preserves the flag for both new (0→0) and legacy (1→1) groups. `TestAddGroup_DefaultsIsNamedZero` covers the direct-insert path. `TestRenameThenInviteUpdate_KeepsNameAndIsNamed` is the P1 clobber regression on a legacy group. Render: `TestGroupAvatarGetNamedRendersNameText` / `...NamedCustomColorRendersNameText` assert `is_named=1` (legacy) → name-text; `TestGroupAvatarGetAutoNamedRendersIcon` / `...CustomColorIconNoText` assert `is_named=0` (new) → icon; `...CustomOverrides` / `...CustomTextNotTruncated` assert `avatar_text` always wins. Build/vet/golangci-lint/i18n-lint green locally; endpoint tests run in CI.

## Checklist

- [x] Updated documentation

### DIFF
--- a/.octospec/log.md
+++ b/.octospec/log.md
@@ -4,6 +4,23 @@ Change history for this repo's `.octospec/`, following the
 [OKF](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md)
 change-log convention (§7). Newest first.
 
+## 2026-06-29
+
+- **Change** — Task `group-avatar-name-no-text` (client-coordination; repurposes
+  `group-avatar-icon-default` S2): newly created groups now default to the
+  two-person icon — the group **name is never rendered as avatar text**; text
+  appears only when the user sets a custom `avatar_text`. Implemented by changing
+  **who gets `is_named=1`**, not the render rule (`writeGroupDefaultAvatar`
+  unchanged: `avatar_text > is_named==1 name-text > icon`). `is_named` is
+  repurposed from "user named it" to "**pre-cutover legacy group**": all new
+  inserts (`CreateGroup`/`AddGroup`/`event.go` system+org+dept) persist
+  `is_named=0`, and rename no longer flips it; existing groups keep `is_named=1`
+  (already backfilled by migration `20260629000001`) so they are **grandfathered**
+  onto their current name-text avatar (no historical group flips to an icon).
+  `is_named` stays load-bearing (not deprecated) as the legacy/new discriminator;
+  `GroupResp.is_named` re-documented as 1=legacy/0=new predictor. No render-version
+  bump, no new migration. Brief under `.octospec/tasks/group-avatar-name-no-text/`.
+
 ## 2026-06-27
 
 - **Add** — Task `default-avatar-text-rule`: script-aware 2-glyph text rule for
--- a/.octospec/tasks/group-avatar-name-no-text/brief.md
+++ b/.octospec/tasks/group-avatar-name-no-text/brief.md
@@ -75,8 +75,25 @@ legacy avatars (rename would no longer follow).
 - **Migration `20260629000001` is reused as-is, not modified.** Its backfill-to-1
   is exactly the legacy marker this task needs. The file is already merged/applied;
   its in-file comments still describe the #500 "user-named" intent (now reinterpreted
-  as "legacy") — left as the historical record; the Go `Model.IsNamed` comment is
-  the source of truth.
+  as "legacy"). Since several reviewers flagged the stale *live* column COMMENTs
+  (schema introspection would carry old guidance), a follow-up **comment-only
+  migration `20260629000002`** `MODIFY COLUMN`s `is_named` + `avatar_text` to refresh
+  their COMMENTs to the legacy/new semantic (no type/constraint/data change,
+  reentrant via INFORMATION_SCHEMA guard). The historical migration files' `--`
+  comments stay as the record; the Go `Model.IsNamed` comment + the live column
+  COMMENT are the source of truth.
+
+## Follow-ups (out of scope — separate PR/issue)
+- **`UpdateGroupInfo` full-row `UpdateTx` → column-level** (yujiawei, non-blocking):
+  a rename's full-row writeback could clobber a concurrent `invite` toggle with a
+  stale loaded value (the inverse of #500's P1, which was fixed for the invite
+  branch only). Pre-existing; converting `UpdateGroupInfo` to a column-scoped write
+  closes the remaining read-modify-write window.
+- **Public avatar endpoint enumeration / name-prefix disclosure** (yujiawei,
+  non-blocking): the unauthenticated `GET /:group_no/avatar` distinguishes
+  existing vs nonexistent/disbanded groups and renders a legacy group's first-2
+  name glyphs. Pre-existing and intentionally preserved by grandfather; hardening
+  (auth or constant-shape response) is separate scope.
 
 ## Out of scope
 - Any change to the render logic, `avatar_text`/`avatar_color` APIs, validation,
--- a/.octospec/tasks/group-avatar-name-no-text/brief.md
+++ b/.octospec/tasks/group-avatar-name-no-text/brief.md
@@ -1,0 +1,100 @@
+---
+type: Task
+title: "Task: group-avatar-name-no-text"
+description: New groups default to the two-person icon (group name is not avatar text; only custom avatar_text renders text). Existing groups are grandfathered via the is_named flag so they keep their current name-text avatar.
+tags: [group, avatar, render, grandfather, product-pivot]
+timestamp: 2026-06-29T12:00:00Z
+# --- octospec extension fields ---
+slug: group-avatar-name-no-text
+upstream: group-chat-avatar-gen (client-coordination change, 2026-06-29)
+supersedes: group-avatar-icon-default (S2 — is_named meaning is repurposed, not deprecated)
+source: self
+---
+
+# Task: group-avatar-name-no-text
+
+> Server change to match the client: a **newly created** group defaults to the
+> two-person icon; the group **name is never rendered as avatar text**; text
+> appears only when the user explicitly sets a custom `avatar_text`. Existing
+> groups are grandfathered (kept on their current name-text avatar) so the change
+> does not flip every historical group to an icon at once.
+
+## Goal
+Change **who gets `is_named=1`**, not the render rule.
+
+- **Render rule is unchanged from #500:** `avatar_text` (custom) > `is_named==1`
+  → `GroupNameText(name)` (first-2 glyphs) > two-person icon. `writeGroupDefaultAvatar`
+  is untouched in logic.
+- **`is_named` is repurposed** from "user explicitly named the group" to
+  "**pre-cutover legacy group**":
+  - **New groups → `is_named=0` always** (regardless of whether a name was given).
+    So a freshly created group defaults to the icon; setting `avatar_text` is the
+    only way to get text. Matches the client preview (icon when `avatar_text` empty).
+  - **Existing groups → `is_named=1`** — already backfilled by #500's migration
+    (`20260629000001`, conservative backfill-to-1). They keep rendering their name
+    as before (grandfather), so no historical group flips to an icon.
+
+This is the **甲案** chosen by the maintainer (2026-06-29): keep `is_named`
+load-bearing (NOT deprecated) as the legacy/new discriminator. The alternative
+(retire `is_named` + a one-time backfill of name→`avatar_text`) was rejected because
+it requires Go-side script-aware derivation (not expressible in SQL) and freezes
+legacy avatars (rename would no longer follow).
+
+## Background
+- #500 added `is_named` meaning "user named it" and gated name-text on it. The
+  client then changed its create/edit preview to show the icon when `avatar_text`
+  is empty. To match, the server must default new groups to the icon — but #500
+  set `is_named=1` for any explicitly-named new group, so new named groups still
+  rendered name-text. Flipping ALL groups to icon (the global approach, prior PR
+  #503 head) was rejected by product because it changes every existing group's
+  avatar at once. Grandfather is the resolution.
+
+## Load-bearing list
+- **`is_named` write-policy (the whole change).** New groups must persist
+  `is_named=0`:
+  - `Service.CreateGroup` (`service.go`): removed the `isNamedVal = (name != "")`
+    computation; the insert sets `IsNamed: 0`.
+  - `Service.AddGroup` (`service.go`): removed the name-based computation; insert
+    sets `IsNamed: 0`.
+  - `event.go` system-group + org/dept inserts: `IsNamed: 0` (they are served
+    static PNGs and never hit the render path; set to 0 to keep the invariant
+    "`is_named=1` ⟺ pre-cutover legacy row" clean).
+  - `Service.UpdateGroupInfo` rename: removed `groupModel.IsNamed = 1`. Rename no
+    longer flips the flag — it preserves the loaded value (legacy stays 1 and its
+    name-text follows the new name; new stays 0 and stays an icon). `UpdateTx`'s
+    full-row SetMap still writes the (unchanged) loaded `is_named` back.
+- **Render rule unchanged (`writeGroupDefaultAvatar`, `api.go`).** Still
+  `avatar_text > is_named==1 name-text > icon`. Only the comments are reworded to
+  the legacy/new semantic. ETag factors unchanged: text present (legacy name or
+  custom) → `group-name-v4`; no text (new group) → `group-icon-v3`.
+- **`is_named` is read-only after creation.** Only set by #500's migration
+  backfill (legacy → 1); every app insert writes 0; rename/invite never change it.
+  So `is_named=1` strictly identifies pre-cutover groups.
+- **`GroupResp.is_named` kept** (additive, #500). Re-documented as
+  "1=legacy/name-text, 0=new/icon"; clients can still locally predict the default.
+- **Migration `20260629000001` is reused as-is, not modified.** Its backfill-to-1
+  is exactly the legacy marker this task needs. The file is already merged/applied;
+  its in-file comments still describe the #500 "user-named" intent (now reinterpreted
+  as "legacy") — left as the historical record; the Go `Model.IsNamed` comment is
+  the source of truth.
+
+## Out of scope
+- Any change to the render logic, `avatar_text`/`avatar_color` APIs, validation,
+  upload path, palette endpoint/values.
+- Dropping or re-migrating the `is_named` column (kept, repurposed).
+- octo-web (already shows the icon when `avatar_text` is empty).
+
+## Acceptance
+- `CreateGroup` / `AddGroup` with or without a name → `is_named=0` (new groups
+  default to the icon). `UpdateGroupInfo` rename → `is_named` unchanged (new stays
+  0, legacy stays 1).
+- `writeGroupDefaultAvatar`: `is_named=1` (legacy) + no `avatar_text` →
+  `RenderGroup(GroupNameText(name))`; `is_named=0` (new) + no `avatar_text` →
+  `RenderIcon`; custom `avatar_text` always overrides; custom color honored in both.
+- Tests: `TestCreateGroup_Success`/`_AutoGenerateName` assert `is_named=0`;
+  `TestUpdateGroupInfo_RenameDoesNotChangeIsNamed`;
+  `TestRenameThenInviteUpdate_KeepsNameAndIsNamed` (P1 clobber regression, legacy
+  group); `TestAddGroup_DefaultsIsNamedZero`; `TestGroupAvatarGetNamedRenders*`
+  (legacy → name-text) and `...AutoNamedRendersIcon` (new → icon) pass.
+- `go build ./...`, `go vet`, `golangci-lint`, `make i18n-lint` green. Endpoint
+  tests needing MySQL/Redis/WuKongIM run in CI.
--- a/modules/group/api.go
+++ b/modules/group/api.go
@@ -425,12 +425,14 @@ func (g *Group) avatarGet(c *wkhttp.Context) {
 	}
 
 	// 无自定义上传（含历史合成群、新建群）：服务端实时渲染默认头像——有自定义文字则渲染
-	// 该文字；否则命名群（is_named=1）取群名前 2 字，自动名群（is_named=0）回退双人图标。
+	// 该文字；否则老群（is_named=1，改版前存量群）取群名前 2 字，新群（is_named=0）回退双人图标。
 	g.writeGroupDefaultAvatar(c, groupNo, groupInfo)
 }
 
 // writeGroupDefaultAvatar 服务端渲染并返回群默认头像（无自定义上传时）。文字优先级：
-// 自定义 avatar_text > 命名群（is_named=1）的群名前 2 字 > 空（自动名群 → 双人图标）。
+// 自定义 avatar_text > 老群（is_named=1，改版前存量群）的群名前 2 字 > 空（新群 is_named=0
+// → 双人图标）。产品 2026-06-29 改版：新建群默认双人图标、群名不作为头像文字，仅 avatar_text
+// 才渲染文字；is_named=1 仅由 #500 迁移回填给存量老群，让它们保留原群名文字头像（grandfather）。
 // 颜色优先自定义 avatar_color，否则按 group_no 稳定派生（改名不变色、跨页面一致）。
 //
 // URL 稳定为 groups/{group_no}/avatar，内容随群名/自定义变化，故用内容相关弱 ETag +
@@ -441,11 +443,12 @@ func (g *Group) writeGroupDefaultAvatar(c *wkhttp.Context, groupNo string, group
 	colorTag := "seed"
 	var text string
 	if groupInfo != nil {
-		// 默认头像取字规则(产品 2026-06-29 定稿)：
+		// 默认头像取字规则(产品 2026-06-29 改版)：
 		//   1. 用户在「修改头像」显式设了自定义文字 → 原样渲染(写入时已 ≤4 规范化、不取字);
-		//   2. 否则群是用户**显式起名**(is_named=1) → 取群名前 2 字(script 感知);
-		//   3. 否则(成员名拼接的自动默认名 is_named=0) → text 留空 → 回退双人图标
-		//      (不把「张三、李四、王五」这类拼接名渲成头像文字)。
+		//   2. 否则是**改版前的存量老群**(is_named=1) → 取群名前 2 字(script 感知)，保留其原有
+		//      群名文字头像(grandfather，避免老群一夜全变小人);
+		//   3. 否则(新群 is_named=0) → text 留空 → 回退双人图标(群名不作为头像文字，要文字请设
+		//      avatar_text)。
 		if groupInfo.AvatarText != "" {
 			text = avatarrender.GroupText(groupInfo.AvatarText)
 		} else if groupInfo.IsNamed == 1 {
@@ -461,10 +464,10 @@ func (g *Group) writeGroupDefaultAvatar(c *wkhttp.Context, groupNo string, group
 	renderable := avatarrender.Renderable(text)
 
 	// ETag 覆盖决定图像内容的因子：渲染模式版本 + group_no(派生色) + 实际色 + 文字。
-	// 改名/改自定义文字/改 is_named → text 变(或在 name 模式与 icon 模式间切换) → ETag 变；
-	// 改自定义色 → colorTag 变 → ETag 变。命名群走 group-name-v4(+text)、自动名群走
-	// group-icon-v3(无 text)，模式不同因子串天然不同，故两类切换无需 bump 版本。渲染
-	// **视觉**改动(像素变但因子不变，如 #486 透明四角)才必须 bump 版本段。
+	// 老群改名 → text 变 → ETag 变；改/清自定义文字(或在 text 模式与 icon 模式间切换) →
+	// ETag 变；改自定义色 → colorTag 变 → ETag 变。有文字(老群群名/自定义)走 group-name-v4
+	// (+text)、无文字(新群)走 group-icon-v3(无 text)，模式不同因子串天然不同，故两类切换无需
+	// bump 版本。渲染**视觉**改动(像素变但因子不变，如 #486 透明四角)才必须 bump 版本段。
 
 	etag := avatarrender.ETag("group-icon-v3", groupNo, colorTag)
 	if renderable {
@@ -4210,7 +4213,7 @@ type groupReq struct {
 	Members     []string `json:"members"`      // 成员uid
 	SpaceID     string   `json:"space_id"`     // Space ID（可选）
 	CategoryID  string   `json:"category_id"`  // 群聊分组 ID（可选，需配合 space_id 使用）
-	AvatarText  string   `json:"avatar_text"`  // 自定义群头像文字（可选，最多 4 个中文/英文字符；空=按 is_named 回退：命名群群名/自动名群双人图标）
+	AvatarText  string   `json:"avatar_text"`  // 自定义群头像文字（可选，最多 4 个中文/英文字符；空=按 is_named 回退：老群渲染群名/新群双人图标）
 	AvatarColor *int     `json:"avatar_color"` // 自定义群头像色板下标（可选，[0,palette)；不传=按 group_no 派生）
 }
 
@@ -4223,7 +4226,8 @@ func (g groupReq) Check() error {
 
 // checkAvatar 校验二次弹窗的自定义头像参数，返回越界字段名（供 Details.field）。
 // ok=true 表示合法。两者均为可选：avatar_text 空、avatar_color 不传即“未自定义”，
-// 渲染时分别回退到群名前 2 字 / ColorForSeed(group_no)。不静默截断，超限直接拒绝。
+// 渲染时分别回退到 is_named 规则(老群群名前2字/新群双人图标) / ColorForSeed(group_no)。
+// 不静默截断，超限直接拒绝。
 //
 // 哨兵约定（创建 vs 改群刻意不对称）：创建无既有值可清除，故 avatar_color 仅接受
 // [0,palette)（-1 在此被拒）；改群额外接受 "-1" / "" 表示清除自定义色回退派生。
--- a/modules/group/api.go
+++ b/modules/group/api.go
@@ -1032,7 +1032,7 @@ func (g *Group) groupUpdate(c *wkhttp.Context) {
 	// 自定义群头像文字/颜色（二次弹窗保存）：先在任何 mutation 之前完成解析与校验，构造
 	// avatarReq。这样非法的 avatar 字段会在 name/notice/invite 落库之前返回 400，避免
 	// 「返回 400 却已部分写入群名」的非原子部分写入。avatar_text 空串清除自定义文字（回退
-	// 群名），avatar_color "" / "-1" 清除自定义色（回退派生）。超限直接拒绝，不静默截断。
+	// is_named 规则:老群群名/新群双人图标），avatar_color "" / "-1" 清除自定义色（回退派生）。超限直接拒绝，不静默截断。
 	avatarTextValue, hasAvatarText := groupMap[attrKeyAvatarText]
 	avatarColorValue, hasAvatarColor := groupMap[attrKeyAvatarColor]
 	var avatarReq *UpdateGroupAvatarCustomServiceReq
--- a/modules/group/avatar_custom_test.go
+++ b/modules/group/avatar_custom_test.go
@@ -60,7 +60,7 @@ func TestGroupReqCheckAvatar(t *testing.T) {
 
 // TestGroupAvatarCustomDBRoundTrip 覆盖自定义头像文字/颜色的落库读回：
 //   - 默认建群（未自定义）写入 avatar_text=” / avatar_color=NULL（*int=nil），
-//     这是渲染时回退“群名前 2 字（script 感知取字）+ ColorForSeed(group_no)”的哨兵；
+//     这是渲染时回退“is_named 规则(老群群名前2字/新群双人图标) + ColorForSeed(group_no)”的哨兵；
 //   - 显式自定义写入文字与色板下标，QueryWithGroupNo 原样读回。
 //
 // *int 的 nil→NULL 依赖 dbr Record() 对指针字段的处理（与 group_md *string 同理）。
--- a/modules/group/avatar_get_test.go
+++ b/modules/group/avatar_get_test.go
@@ -25,9 +25,9 @@ func doAvatarGet(t *testing.T, h http.Handler, groupNo, ifNoneMatch string) *htt
 	return w
 }
 
-// TestGroupAvatarGetAutoNamedRendersIcon 覆盖核心规则（产品 2026-06-29 定稿）：成员名
-// 拼接的**自动默认名**（is_named=0）且无自定义文字 → 双人图标（不把拼接名渲成头像文字），
-// 即使 name 字段非空。带弱 ETag + must-revalidate；命中 If-None-Match 时 304。
+// TestGroupAvatarGetAutoNamedRendersIcon 覆盖核心规则（产品 2026-06-29 改版）：新群
+// （is_named=0）且无自定义文字 → 双人图标（群名不作为头像文字），即使 name 字段非空。
+// 带弱 ETag + must-revalidate；命中 If-None-Match 时 304。
 func TestGroupAvatarGetAutoNamedRendersIcon(t *testing.T) {
 	s, ctx := newTestServer(t)
 	require.NoError(t, testutil.CleanAllTables(ctx))
@@ -61,8 +61,8 @@ func TestGroupAvatarGetAutoNamedRendersIcon(t *testing.T) {
 	require.Empty(t, w2.Body.Bytes())
 }
 
-// TestGroupAvatarGetNamedRendersNameText 覆盖：用户**显式起名**（is_named=1）且无自定义
-// 文字 → 取群名前 2 字（script 感知）渲染文字，而非双人图标。
+// TestGroupAvatarGetNamedRendersNameText 覆盖 grandfather：**改版前的存量老群**（is_named=1）
+// 且无自定义文字 → 取群名前 2 字（script 感知）渲染文字，保留其原有群名文字头像，而非双人图标。
 func TestGroupAvatarGetNamedRendersNameText(t *testing.T) {
 	s, ctx := newTestServer(t)
 	require.NoError(t, testutil.CleanAllTables(ctx))
@@ -75,7 +75,7 @@ func TestGroupAvatarGetNamedRendersNameText(t *testing.T) {
 
 	w := doAvatarGet(t, s.GetRoute(), groupNo, "")
 	require.Equal(t, http.StatusOK, w.Code)
-	// 命名群也走内容相关弱 ETag：保护改名 / 切自定义触发的缓存失效契约（Jerry-Xin 🔵）。
+	// 老群文字头像也走内容相关弱 ETag：保护改名 / 切自定义触发的缓存失效契约（Jerry-Xin 🔵）。
 	etag := w.Header().Get("ETag")
 	require.True(t, strings.HasPrefix(etag, `W/"`), "named group avatar must carry a weak ETag, got %q", etag)
 	require.Contains(t, w.Header().Get("Cache-Control"), "must-revalidate")
@@ -87,9 +87,9 @@ func TestGroupAvatarGetNamedRendersNameText(t *testing.T) {
 	)
 	require.NoError(t, err)
 	require.Equal(t, want, w.Body.Bytes(),
-		"named group (is_named=1) without custom text must render script-aware first-2 of the name")
+		"legacy group (is_named=1) without custom text must render script-aware first-2 of the name")
 
-	// 命中的 If-None-Match → 304 无 body（命名群路径同样支持条件请求省渲染）。
+	// 命中的 If-None-Match → 304 无 body（老群文字路径同样支持条件请求省渲染）。
 	w2 := doAvatarGet(t, s.GetRoute(), groupNo, etag)
 	require.Equal(t, http.StatusNotModified, w2.Code)
 	require.Empty(t, w2.Body.Bytes())
@@ -135,9 +135,9 @@ func TestGroupAvatarGetCustomOverrides(t *testing.T) {
 	require.Equal(t, want, w.Body.Bytes(), "custom text+color must override name/seed")
 }
 
-// TestGroupAvatarGetCustomColorIconNoText 覆盖 S2:自动名群(is_named=0)设了自定义颜色但
-// **无**自定义文字 → 渲染该颜色的双人图标(自动名群默认头像与群名无关,但自定义颜色仍被
-// 尊重)。fixture 显式 IsNamed=0,名字取「自动名群」以贴合意图(避免与 is_named=1 混淆)。
+// TestGroupAvatarGetCustomColorIconNoText 覆盖:新群(is_named=0)设了自定义颜色但**无**自定义
+// 文字 → 渲染该颜色的双人图标(新群默认头像与群名无关,但自定义颜色仍被尊重)。fixture 显式
+// IsNamed=0,名字取「自动名群」以贴合意图(避免与老群 is_named=1 混淆)。
 func TestGroupAvatarGetCustomColorIconNoText(t *testing.T) {
 	s, ctx := newTestServer(t)
 	require.NoError(t, testutil.CleanAllTables(ctx))
@@ -159,8 +159,8 @@ func TestGroupAvatarGetCustomColorIconNoText(t *testing.T) {
 }
 
 // TestGroupAvatarGetNamedCustomColorRendersNameText 锁定三因子交互(Octo-Q P2-1 建议):
-// 命名群(is_named=1) + 自定义颜色 + **无**自定义文字 → 以该自定义颜色渲染群名前 2 字
-// (而非派生色、而非图标)。补全 is_named=1 与 custom_color 的组合覆盖。
+// 老群(is_named=1) + 自定义颜色 + **无**自定义文字 → 以该自定义颜色渲染群名前 2 字
+// (而非派生色、而非图标)。补全 is_named=1(老群) 与 custom_color 的组合覆盖。
 func TestGroupAvatarGetNamedCustomColorRendersNameText(t *testing.T) {
 	s, ctx := newTestServer(t)
 	require.NoError(t, testutil.CleanAllTables(ctx))
@@ -183,7 +183,7 @@ func TestGroupAvatarGetNamedCustomColorRendersNameText(t *testing.T) {
 	)
 	require.NoError(t, err)
 	require.Equal(t, want, w.Body.Bytes(),
-		"named group (is_named=1) with custom color + no text must render name first-2 in that custom color")
+		"legacy group (is_named=1) with custom color + no text must render name first-2 in that custom color")
 }
 
 // TestGroupAvatarGetCustomTextNotTruncated 回归 PR#494 评审(Jerry-Xin):用户显式
--- a/modules/group/avatar_version_pin_test.go
+++ b/modules/group/avatar_version_pin_test.go
@@ -28,8 +28,9 @@ func TestGroupAvatarGetPinsRenderVersion(t *testing.T) {
 	require.NoError(t, testutil.CleanAllTables(ctx))
 	g := New(ctx)
 
-	// Custom avatar_text is the ONLY path to text-render now that the group name no
-	// longer drives the default avatar -> name render mode, default colorTag "seed".
+	// This fixture uses custom avatar_text -> name render mode, default colorTag "seed".
+	// (Legacy is_named=1 groups also render group-name text; new groups fall back to icon.
+	//  This test pins the version segment of the text-render mode, regardless of text source.)
 	const namedNo = "avatar_pin_named_1"
 	require.NoError(t, g.db.Insert(&Model{GroupNo: namedNo, Name: "后端架构讨论", Creator: "c1", Status: 1, AvatarText: "研发"}))
 	named := doAvatarGet(t, s.GetRoute(), namedNo, "")
--- a/modules/group/const.go
+++ b/modules/group/const.go
@@ -47,7 +47,7 @@ const (
 // 群信息更新（PUT /v1/group/:group_no）中自定义群头像的 map key，与创建接口
 // groupReq 的 JSON 字段名保持一致。avatar_color 值为色板下标字符串，"" 或 "-1"
 // 表示清除自定义色（回退按 group_no 派生）；avatar_text 为空字符串表示清除自定义
-// 文字（回退：命名群取群名前 2 字，自动名群双人图标）。
+// 文字（回退：老群取群名前 2 字，新群双人图标）。
 const (
 	attrKeyAvatarText  = "avatar_text"
 	attrKeyAvatarColor = "avatar_color"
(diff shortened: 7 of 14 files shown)


OctoBoooot (commented):
## Review: feat(group): default avatar drops group-name text, icon unless avatar_text (#503)

**Verdict:** Approve with comments — clean product-pivot render-path change; the ETag mode-switch reasoning is sound and byte-verified. No blockers, no majors; one minor (dead code). CI red on `check-sprint` only (admin gate); `go vet` + i18n clean.

Risk: normal. `needs-human-review` + `size/L`. Established author (100 merged). This supersedes the #494/#500 name-text render rule (which I reviewed) — render path only; the `is_named` column + lifecycle + #500's `UpdateInviteTx` clobber-fix are kept.

### Verified — the parts most likely to be wrong

- ✅ **ETag mode-switch invalidates existing named groups correctly** (the recurring no-bump trap) — `text` now derives only from `AvatarText`; a named group with empty `avatar_text` → `text=""` → `renderable=false` → ETag uses `group-icon-v3` (no text factor). Its prior ETag was `group-name-v4 + name-text` — different version segment AND different factor list, so `If-None-Match` is guaranteed to miss and the client revalidates to the icon within `max-age`. No collision risk (the version segment alone differs), so the "no render-version bump" claim holds. Byte-checked the factor construction at `api.go:447-460`.
- ✅ **`is_named` / `GroupNameText` fully removed from the render path** — grepped `api.go`: the only remaining `IsNamed`/`is_named` references are comments and the unrelated `UpdateInviteTx` clobber-fix (#500, correctly preserved). No dead read of the column at render time.
- ✅ **Disband gate intact** — `groupInfo.Status == GroupStatusDisband → return` (no default render) is unchanged; only its comment was reworded. The "avatar_text could leak for a disbanded group" the new comment mentions is a description of the pre-existing short-circuit nuance, **not** a new leak — the disband check still fires before any render. No regression.
- ✅ **Swagger synced** — no stale "前 2 字"/"命名群" name-text descriptions remain; the 5 field/endpoint descriptions reflect the custom-text-or-icon rule.
- ✅ **Tests flipped to assert the icon** — `TestGroupAvatarGetNamedRendersIcon` + the three-factor `...NamedCustomColorRendersIcon`; `go vet` clean (tests compile).

### Minor (non-blocking)

- `pkg/avatarrender/text.go` `GroupNameText` is now **dead code** — zero non-test callers after this PR removes the only one (`writeGroupDefaultAvatar`). Not in this diff, so noting here rather than inline. Worth deleting (or a follow-up) so a future reader doesn't assume the name-text rule is still live; its docstring still describes the superseded behavior.

### Praise

- **The ETag reasoning for the global flip is spelled out and correct** — rather than reflexively bumping the render version, the author traced *why* the mode-switch (`group-name-v4`+text → `group-icon-v3`, no text) changes the ETag on its own, so existing named groups converge to the icon within max-age without a forced cache-bust. That's the precise understanding of this avatar pipeline's invalidation model (the exact thing #486/#494/#500 kept circling), applied correctly to a behavior reversal.
- **The `is_named` column was kept dormant rather than ripped out** — the render path stops reading it, but the column, lifecycle writes, and #500's clobber-fix + tests stay intact. That's the right call for a product pivot that might pivot again: drop the behavior, keep the data + the hard-won fix, don't churn the schema.

- CI `check-sprint` — admin Sprint-field gate, independent of the code.

Jerry-Xin (dismissed):
Scope gate passed: this PR changes octo-server’s group avatar render path and related docs/tests, and the implementation looks ready to merge.

**💬 Non-Blocking**
- 🟡 Warning [modules/group/service.go:1022](modules/group/service.go:1022): the `UpdateGroupAvatarCustomServiceReq.AvatarText` comment still says clearing custom text falls back to the group name. The code now falls back to the two-person icon, so this should be updated to avoid misleading future callers.
- 🟡 Warning [modules/group/sql/20260629000001_group_is_named.sql:2](modules/group/sql/20260629000001_group_is_named.sql:2), [modules/group/sql/20260625000001_group_avatar_custom.sql:3](modules/group/sql/20260625000001_group_avatar_custom.sql:3): migration comments and DB column comments still describe the old group-name fallback. Runtime behavior is unaffected, but new deployments/schema introspection will carry stale operational guidance. If migrations are treated as immutable, consider a follow-up metadata/comment migration or a current note elsewhere.
- 🔵 Suggestion [pkg/avatarrender/text.go:132](pkg/avatarrender/text.go:132): `GroupNameText` docs still read like an active server default-avatar rule. Since the function remains exported but dormant for group rendering, a short “not used by the current group endpoint” note would reduce confusion.

**✅ Highlights**
- The behavior change is correctly centralized in `writeGroupDefaultAvatar`; `IsNamed` is no longer consumed for rendering.
- The ETag/cache mode naturally moves empty `avatar_text` groups to the icon key, avoiding stale reuse of the previous name-text cache key.
- Tests cover named groups, named groups with custom color, auto-named groups, custom text override, 304 behavior, uploaded avatar redirect, and 404 cases.

Tests not run locally; review was based on the provided diff and checked-out source.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#503 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-server/pull/503
**Head SHA:** `98ee54d6ea2229b15334af3b66435477ee305654`
**Scope:** 10 files, +185/-76 — render-rule pivot for default group avatar

---

## 1. Investigation Summary (4-phase)

**Phase 1 — Problem:** PR #500 shipped the `is_named=1 → GroupNameText(name)` render branch (3-tier priority: custom avatar_text > named-group name text > icon). Product re-pivot on 2026-06-29: group names must never be avatar text. Collapse to 2 tiers: custom `avatar_text` > icon.

**Phase 2 — Root cause:** Product decision (not a bug fix). The `is_named` column and lifecycle are kept dormant for reversibility and to preserve PR #500's P1 clobber-fix (`UpdateInviteTx`).

**Phase 3 — Impact:** Existing named groups flip from name-text to icon within `max-age=300` via content-derived ETag change. Purely cosmetic/UX change. No API/contract/schema break.

**Phase 4 — Risk:** Low. Single chokepoint change in `writeGroupDefaultAvatar`. Tests flipped to assert icon. No new migration. No render-version bump (icon/name pixels untouched).

---

## 2. Data-Flow Trace

| Consumed data | Upstream source | Runtime delivery verified |
|---|---|---|
| `groupInfo.AvatarText` | SQL `SELECT` via `g.db.QueryWithGroupNo(groupNo)` → `Model.AvatarText` (`db.go:614`) | ✅ Non-empty → `GroupText(AvatarText)` → `RenderGroup`; empty → `text==""` → `RenderIcon`. No intermediate gate short-circuits. |
| `groupInfo.AvatarColor` | SQL → `Model.AvatarColor *int` (`db.go:615`) | ✅ `nil` → seed-derived style; non-nil → `GroupStyleByIndex(*AvatarColor)` with fallback to seed on out-of-range. |
| `groupInfo.IsNamed` | SQL → `Model.IsNamed int` (`db.go:611`) | ✅ **No longer consumed by render path.** Still written on create/rename/event/AddGroup, still exposed via `GroupResp.is_named` JSON. |
| `groupInfo.IsUploadAvatar` | SQL → `Model.IsUploadAvatar` (`api.go:411`) | ✅ Gate before `writeGroupDefaultAvatar` — uploaded avatars redirect to object storage, never reach the default render path. |
| `groupInfo.Status` | SQL → `Model.Status` (`api.go:405-408`) | ✅ Disbanded/nonexistent → 404 before any render. |

---

## 3. Findings

### No P0 or P1 findings

### P2 Findings

**P2-1: `GroupNameText` is dead production code**
- `pkg/avatarrender/text.go:142` defines `GroupNameText(name string) string` — script-aware first-2-glyph extraction.
- After this PR, **zero production call-sites** remain (the only caller was the removed `else if` branch in `writeGroupDefaultAvatar`).
- Still referenced in tests: `group_samples_test.go:34`, `text_test.go:59-105`, `avatar_get_test.go:207` (assertion comment).
- **Severity:** P2 — dead public API in a shared package. Not a functional issue. Consider removing the function and its tests in a follow-up cleanup to reduce the API surface and avoid future confusion.
- **Diff-scope:** pre-existing function, newly dead due to this PR's removal of its sole production caller.

**P2-2: Stale DB column comment (SQL migration from PR #500)**
- `modules/group/sql/20260629000001_group_is_named.sql` lines 2-3, 28, 42 still say "默认头像取字仅对 is_named=1 生效".
- The DB column `COMMENT` will remain stale after this PR.
- This PR correctly does not modify the historical migration file.
- **Severity:** P2 — cosmetic. Consider a no-op migration to update the column comment, or document the discrepancy.
- **Diff-scope:** pre-existing text, newly misleading due to this PR's behavioral change.

---

## 4. Validation

- ✅ **Render logic:** `writeGroupDefaultAvatar` (`api.go:437-510`) — `AvatarText != ""` → `GroupText()` → `RenderGroup`; else → `RenderIcon`. `IsNamed` not read. Correct.
- ✅ **ETag convergence:** (`api.go:462-469`) `group-name-v4` mode (with text) vs `group-icon-v3` mode (no text). Existing named groups switch mode → ETag changes → clients revalidate within `max-age=300`. No version bump needed.
- ✅ **Tests:** `TestGroupAvatarGetNamedRendersIcon` and `TestGroupAvatarGetNamedCustomColorRendersIcon` correctly assert icon bytes for `IsNamed=1` groups. `TestGroupAvatarGetCustomOverrides` / `...CustomTextNotTruncated` confirm `avatar_text` still wins. `TestGroupAvatarGetAutoNamedRendersIcon` / `...IconFallback` confirm icon path. All test names and assertions are consistent with the new rule.
- ✅ **`is_named` lifecycle preserved:** `CreateGroup` (`service.go:1125-1127`), `UpdateGroupInfo` rename (`service.go:1891`), `AddGroup` (`service.go:203-205`), system/org/dept events (`event.go:176,319`) all still write `is_named`. Column kept, `GroupResp.is_named` still exposed. No regression on PR #500's P1 clobber-fix (`TestRenameThenInviteUpdate_KeepsIsNamed`).
- ✅ **Comments/docs updated consistently:** All inline comments in `api.go`, `service.go`, `db.go`, `const.go` updated to reflect "仅记录，渲染不再读取". Swagger `api.yaml` (5 descriptions) updated. Octospec brief and changelog entry added.
- ✅ **Cache/render safety:** `GetOrRender` with LRU + singleflight + semaphore (`api.go:486-510`) unchanged. No new render amplification risk.

---

## 5. Suggestions

- **S1:** Remove `GroupNameText` from `pkg/avatarrender/text.go` and its tests in a follow-up cleanup PR (addresses P2-1).
- **S2:** Consider a lightweight migration to update the `is_named` column comment to match the new semantics (addresses P2-2). Low urgency.

---

## 6. Blind-Spot Checklist (R5 — security_sensitive PR, all items mandatory)

- **C1 — Dual-path parity (symmetric paths):** CLEAR. All `IsNamed` write paths (CreateGroup, UpdateGroupInfo/rename, AddGroup, event.go system/org/dept) are symmetric — they all continue to write `is_named` identically. No add/remove or create/delete asymmetry introduced. The render path no longer reads `IsNamed` uniformly for all group types. The `UpdateInviteTx` column-level write (`db.go:296-303`) still protects against the name/is_named clobber regression from PR #500. No drift.

- **C2 — Control-flow ordering / nested reuse + non-canonical security control trial:** CLEAR. `writeGroupDefaultAvatar` is a simple linear path with two branches (text vs icon). No nested reuse, no dual invocation, no ordering hazards. `GroupText()` passes text as-is (already validated at write time ≤4 chars). No regex/escape/sanitize controls to trial.

- **C3 — Authorization boundary ≠ capability boundary:** CLEAR. The avatar GET endpoint is intentionally public (unauthenticated). This is documented design (`api.go:401-408`), consistent with existing behavior. No new endpoints, no new capabilities exposed.

- **C4 — Authorization lifecycle / container-member state cascade:** CLEAR. No authorization logic changed. The disbanded/nonexistent group check (`api.go:405-408`) returns 404 before any render, preventing avatar data leakage for disbanded groups. `is_named` column lifecycle unchanged.

- **C5 — Build/note ≠ runtime path correctness:** CLEAR. Go service, no build artifacts / browser extensions / CLI packaging / relative URL concerns. Build/vet/golangci-lint/i18n-lint reported clean by the author.

- **C6 — Governance/policy/security documentation consistency:** N/A. No governance or policy documents changed (octospec brief is internal task documentation, not a governance/policy file).

---

---

## Verdict

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The change is focused, well-tested, and correctly implements the product re-pivot. Two P2 findings (dead `GroupNameText` function, stale SQL migration comment) are cosmetic and suitable for follow-up cleanup.

OctoBoooot (commented):
Concurring with @Jerry-Xin's doc-drift 🟡s — byte-verified, and they're real stale comments my review missed (I checked swagger but didn't grep the *inverse* phrase "回退群名" across the Go comments + SQL). Specifics at this head:

- **service.go:1022** — "avatar_text … `\"\"` 即清除自定义文字（**回退群名**）" — stale; new behavior is fallback to icon, not group name.
- **api.go:1034-1035** — "avatar_text 空串清除自定义文字（**回退群名**）" — same stale phrase.
- **Migration/column comment** — `sql/20260629000001:3,6,28,42` and the `is_named` column COMMENT still say "默认头像取字仅对 is_named=1 生效：命名群取群名前 2 字". The render path no longer does this; a new deployer reading the schema would be misled (Jerry's #2). The SQL file itself wasn't touched by this PR (pre-existing), but the column comment shipped via db.go is in scope.

All three are documentation-only, non-blocking — they don't change my COMMENT verdict (the code is correct; the ETag mode-switch and is_named-render-removal are byte-verified sound). But they're more than the single dead-code minor I flagged, and my "swagger synced" implied the doc sweep was complete when it wasn't. Fix: replace the remaining "回退群名" / "取群名前 2 字" descriptions with "回退双人图标", and update the column comment (or note is_named is render-dormant). Worth folding in alongside the GroupNameText dead-code removal.

Net: verdict unchanged, but credit to Jerry-Xin — the doc drift is real and I under-swept it.

yujiawei (dismissed):
# Code Review — PR #503 (octo-server)

## Summary
Collapses the group default-avatar text-selection from three tiers to two: custom `avatar_text` → two-person icon. The `is_named=1 → GroupNameText(name)` branch is deleted from `writeGroupDefaultAvatar`; the `is_named` column and its lifecycle writes are kept dormant. The change is minimal, reversible, and matches the stated product pivot.

I built and ran the affected packages locally:
- `go build ./modules/group/...` — clean
- `go vet ./modules/group/` — clean
- `go test ./pkg/avatarrender/...` — pass
- Confirmed no remaining production reader of `IsNamed` in the render path (only lifecycle writes + `GroupResp` serialization remain).

## Verdict: APPROVE

No correctness, security, or build-breaking issues. The render-path change is correct and the ETag/cache reasoning holds: the handler recomputes the ETag from current factors on every request, so a previously-cached `group-name-v4 + name` ETag cannot survive — the factor string changes (text empties, mode flips to `group-icon-v3`), the weak ETag changes, and clients revalidate to the icon within `max-age`. No render-version bump is needed because pixels are untouched.

The items below are non-blocking (one needs a human sanity-check given the security-sensitive classification, the rest are nits).

---

## Needs human verification (not a server-side defect)

**1. `is_named` is still returned to clients — confirm no deployed client renders name-text avatars from it.**
`modules/group/service.go:853` / `:898` still serialize the real `is_named` value in `GroupResp`, and the field was previously documented as a signal clients use to locally pre-render the avatar (name-text vs icon). The web client is already aligned to the icon-default preview, so the server PNG now matches it. But if any **native** client (iOS/Android/desktop) currently renders the group name as an avatar locally off `is_named=1` *without* fetching the server PNG, the "global" pivot would be incomplete on those clients until they update.

This is explicitly out of scope per the linked brief (which keeps `is_named` exposed as informational and scopes client work out), and forcing `is_named=0` in the response would risk breaking other legitimate client uses of the flag — so this is **not** a change I'd block the server PR on. Flagging it only so a human confirms the native clients either fetch the server avatar or have been coordinated, before treating existing name-text avatars as fully eradicated.

## Nits (optional)

**2. Flipped tests don't lock the cache-invalidation transition.**
`modules/group/avatar_get_test.go` (`TestGroupAvatarGetNamedRendersIcon`) asserts the icon bytes and that the ETag is weak, but doesn't assert the ETag is the `group-icon-v3` form, nor that a request carrying the *old* `group-name-v4 + name` ETag is **not** answered with 304. The behavior is correct in the handler, but a regression that preserved the old name-derived ETag while rendering the icon could slip past this test. Consider asserting the icon-mode ETag shape (or a 200 — not 304 — for a stale name-mode `If-None-Match`).

**3. Stale comment in the `is_named` migration.**
`modules/group/sql/20260629000001_group_is_named.sql` still documents `is_named=1` as driving group-name avatar text ("默认头像取字仅对 is_named=1 生效"). Harmless (migration body unchanged), but new installs carry DB metadata that now contradicts the render rule. Worth a one-line comment refresh for future readers.

**4. `avatarrender.GroupNameText` is now dead in the render path.**
It remains exported, documented as deriving avatar text from the name, and is still exercised by `pkg/avatarrender/group_samples_test.go`. The brief intentionally keeps it for potential reuse (Go won't flag exported-unused), which is fine — just noting it as a reintroduction footgun; a short "no production caller as of #503" note on the function would help.

## Coverage notes
- Verified within this diff + repo: render-path correctness, ETag recomputation, absence of other `IsNamed` render readers, build/vet/avatarrender tests.
- Not verifiable from the server repo: native client-side avatar rendering behavior (item 1), CDN/edge cache configuration, and the actual deployed-client read path for `is_named`. These are the residual surfaces a human should confirm for the security-sensitive "global eradicate name-text" goal.

OctoBoooot (dismissed):
## Review: feat(group): default avatar drops group-name text (#503) — re-review @ 92d61424

**Verdict:** Approve — the commit replaces the v1 "all groups → icon" behavior with a grandfather design (existing groups keep name-text, new groups → icon), correctly implemented and byte-verified; it also resolves my prior doc/dead-code minor. No blockers, no majors. CI red on `check-sprint` only; `go vet` + Build + i18n green.

Delta from prior COMMENT @ `98ee54d6`: this is a **behavior change**, not just the doc cleanup I'd flagged — "default new groups to icon; grandfather existing via is_named." The render path re-reads `is_named`: `avatar_text → (is_named=1 → GroupNameText) → icon`.

### Verified — the load-bearing parts

- ✅ **Grandfather logic is correct** — `is_named`'s semantics are cleanly redefined from "user-named" to "pre-revamp group": ALL new-group creation paths hardcode `IsNamed: 0` (`service.go:206` AddGroup, `:1171` CreateGroup), and #500's migration backfilled all pre-existing rows to `is_named=1` (`UPDATE ... WHERE is_named IS NULL` then `MODIFY NOT NULL DEFAULT 0`). So existing groups (=1) keep their name-text avatar, every new group (=0) gets the icon unless `avatar_text` is set. Goal achieved without a jarring flip for existing groups.
- ✅ **Rename doesn't flip `is_named`** (bidirectional-edge check) — `UpdateGroupInfo` (`:1885-1886`) explicitly no longer touches `is_named` on rename, with a comment naming why. So a grandfathered group renamed keeps name-text; a new group renamed stays icon. Both directions correct — no inverse-edge bug.
- ✅ **ETag/cache reasoning holds** (the recurring trap) — existing `is_named=1` groups keep `group-name-v4`+text (the SAME ETag they already had → no cache disruption, no stale image); new `is_named=0` groups are `group-icon-v3` (fresh, no prior cache). No render-version bump needed; reasoning byte-checked against the factor construction.
- ✅ **My prior minors resolved** — `GroupNameText` is no longer dead code (the grandfather branch reads it again), and the stale "回退群名" doc comments are rewritten to the grandfather semantics; swagger + `GroupResp.is_named` description updated (6 refs). `go vet` clean, Build green, i18n extract-check exits 0.

### Praise

- **The pivot keeps the data and re-purposes it rather than re-adding a column** — `is_named` was kept dormant in v1 "just in case"; this commit cleanly re-defines its meaning (pre-revamp marker) and drives the grandfather behavior off the #500 migration's existing backfill, so there's no new migration and existing groups need no data change. That's the payoff of having kept the column.
- **The grandfather boundary is drawn at exactly the right place** — hardcoding new creation to `is_named=0` (regardless of name) + leaving rename untouched means the "old groups keep their look, new groups follow the new rule" contract holds across every create and rename path, not just the obvious one. The comments name the semantics shift so the next reader isn't misled by the old "user-named" meaning.

- CI `check-sprint` — admin Sprint-field gate, independent of the code. Lint/Test pending but Build+Vet pass and the change is logic-localized.

Jerry-Xin (dismissed):
Project relevance gate: passed. This PR changes `modules/group` avatar/default group behavior and related docs/tests, so it belongs in octo-server.

One-line summary: The implementation matches the stated policy: new group creation paths now persist `is_named=0`, rename preserves the existing marker, and the render rule still correctly uses `avatar_text > legacy is_named name text > icon`.

💬 Non-blocking

- 🔵 Suggestion: Update a few stale comments so future maintainers do not reintroduce the old meaning of `is_named`.
  - `modules/group/db.go:283` still says rename sets `is_named=1`, but `UpdateGroupInfo` now intentionally preserves the loaded value.
  - `pkg/avatarrender/text.go:140` still frames `GroupNameText` around “named groups” and auto-named groups; the current server-side caller uses it for legacy groups only.
  - `modules/group/avatar_version_pin_test.go:31` says custom `avatar_text` is the only text-render path, but legacy `is_named=1` groups also render group-name text.

✅ Highlights

- The write-policy change is scoped well: `CreateGroup`, `AddGroup`, and event-created groups all converge on `IsNamed: 0`, while existing rows stay grandfathered through the prior migration.
- Rename behavior is covered for both new and legacy groups in `TestUpdateGroupInfo_RenameDoesNotChangeIsNamed`.
- Render behavior has good coverage for legacy name text, new icon fallback, custom text override, and custom color interactions.
- I ran `go test ./pkg/avatarrender/...` successfully. I did not run the group endpoint tests because they depend on the repository’s MySQL/Redis/WuKongIM test environment.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#503 Review Report (automated review)
**PR:** feat(group): default new groups to icon avatar; grandfather existing via is_named
**Head SHA:** `92d61424a72a6bfaf17990dad73116bbb09c06f9`
**Files:** 11 changed (+227 / −95)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `CreateGroup` → `IsNamed: 0` | ✅ | `service.go:1171` — removed `isNamedVal` computation, hardcoded 0 |
| `AddGroup` → `IsNamed: 0` | ✅ | `service.go:206` — removed `isNamed` conditional, hardcoded 0 |
| `event.go` system/org/dept → `IsNamed: 0` | ✅ | `event.go:176,319` — both insert paths set 0 |
| `UpdateGroupInfo` rename preserves `is_named` | ✅ | `service.go:1882-1887` — removed `groupModel.IsNamed = 1`; preserves loaded value |
| Render rule unchanged | ✅ | `api.go:453-455` — `avatar_text > is_named==1 name-text > icon` logic untouched |
| `UpdateTx` full-row write safe | ✅ | `db.go:280-292` — writes `model.IsNamed` from loaded model; since no caller mutates it, pass-through is correct |
| `UpdateInviteTx` column-scoped | ✅ | `db.go:298-305` — only `invite`/`version`; does not touch `is_named` |
| `updateAvatarCustom` column-scoped | ✅ | `db.go:342-366` — only `avatar_text`/`avatar_color`/`version`/`updated_at`; does not touch `is_named` |
| Tests updated | ✅ | `service_test.go` — 5 test functions assert new behavior; `avatar_get_test.go` — 4 render tests updated |
| Swagger/docs updated | ✅ | `swagger/api.yaml` — 5 description fields reworded to legacy/new semantic |
| Migration reused (not modified) | ✅ | `sql/20260629000001_group_is_named.sql` — backfill-to-1 serves as legacy marker; no new migration |

## 2. Findings

### P2-1: Stale comment in `db.go:283` (`UpdateTx`)
**diff-scope:** pre-existing (comment was NOT changed by this PR)

```go
"is_named":  model.IsNamed, // 改名置 1（用户起名）；其它更新原样回写当前值
```

The comment says "rename sets 1 (user named)" but this PR removed the `groupModel.IsNamed = 1` line from the rename path (`service.go:1885`). The code is correct (pass-through), but the inline comment is now misleading. Future readers may think rename still flips the flag.

**Severity:** P2 (nit/maintainability — no runtime impact, code behavior is correct)

**Suggestion:** Update to `"is_named":  model.IsNamed, // 原样回写（改名不再改动 is_named；仅 #500 迁移回填存量老群为 1）`

### P2-2: Stale comment in `pkg/avatarrender/text.go:140`
**diff-scope:** pre-existing (comment was NOT changed by this PR)

```go
// 仅用于「命名群(is_named=1)」的群名自动取字;成员拼接的自动名(is_named=0)不调用本函数、
```

Still uses old "命名群/自动名群" terminology. Under the new semantics, `is_named=1` means "legacy group" not "user-named group."

**Severity:** P2 (nit — `avatarrender` package is outside the diff scope, but would be nice to align)

## 3. Suggestions

- Fix P2-1 and P2-2 stale comments to align with the new `is_named` semantics. Low effort, prevents future confusion.

## 4. Additional Findings

None. The change is narrowly scoped and well-isolated to the `is_named` write-policy. No unexpected side effects discovered.

## 5. Data-Flow Trace

| Consumer | Upstream Source | Data Flows? |
|----------|----------------|-------------|
| `writeGroupDefaultAvatar` → `groupInfo.IsNamed == 1` (api.go:454) | `DB.QueryWithGroupNo` → `Model.IsNamed` column | ✅ Legacy groups: migration backfilled 1 → DB → render path → name-text. New groups: all inserts write 0 → DB → render path → icon. |
| `GroupResp.IsNamed` (service.go:849,894) | `model.IsNamed` from DB query | ✅ Correctly reflects persisted value. API consumers get 1=legacy/0=new. |
| `UpdateTx` → `"is_named": model.IsNamed` (db.go:283) | `UpdateGroupInfo` loads model via `QueryWithGroupNo`, does NOT mutate `IsNamed` → pass-through | ✅ Loaded value preserved. Rename: new(0→0), legacy(1→1). Disband/forbidden: same pass-through. |
| `UpdateInviteTx` | Does not include `is_named` column | ✅ Column-scoped write; `is_named` untouched. |
| `updateAvatarCustom` | Does not include `is_named` column | ✅ Column-scoped write; `is_named` untouched. |
| Migration `20260629000001` | `UPDATE group SET is_named = 1 WHERE is_named IS NULL` | ✅ Existing rows → 1 (legacy marker). Column default 0 for new rows. |

- **C1 — Double-path parity:** ✅ Clear. All group-insert paths (`CreateGroup`, `AddGroup`, `event.go` system+org+dept) consistently set `IsNamed: 0`. No asymmetric add/remove paths touched.
- **C2 — Control-flow ordering / nesting:** ✅ Clear. `writeGroupDefaultAvatar` is a linear if/else-if chain, not reused in nested or compound contexts. No double-application risk.
- **C3 — Authorization boundary:** ✅ N/A. Avatar endpoint was already public; no new endpoint exposure.
- **C4 — Auth lifecycle / container-member cascade:** ✅ N/A. No auth changes.
- **C5 — Build ≠ runtime:** ✅ N/A. No build artifacts, extensions, or relative paths involved.
- **C6 — Governance doc consistency:** ✅ N/A. The octospec brief is a new internal task document; no conflict with existing security/disclosure policies.

---

**[Octo-Q] verdict: APPROVE**

The PR cleanly implements the product pivot (new groups → icon; legacy groups grandfathered via `is_named=1`). All write paths are consistently updated, the render rule is untouched, tests cover both new and legacy groups, and the P1 clobber regression test is adapted to the new semantics. Two P2 stale comments (P2-1 in `db.go:283`, P2-2 in `text.go:140`) are noted as non-blocking nits.

yujiawei (dismissed):
# Code Review — PR #503 (octo-server)

Reviewed at head SHA `92d61424a72a6bfaf17990dad73116bbb09c06f9` against `main`. Build (`go build ./modules/group/...`) and `go vet ./modules/group/...` are green. This is a public-endpoint-adjacent change, so I verified the data-flow and the security surface directly, not just from the diff.

## Summary

This PR repurposes the `is_named` column from "user explicitly named the group" to "pre-cutover legacy group" marker, to match the client's new create/edit preview (icon when `avatar_text` is empty). The render rule in `writeGroupDefaultAvatar` (`modules/group/api.go`) is intentionally unchanged — `avatar_text` > `is_named==1` name-text > two-person icon. Only **who gets `is_named=1`** changes: new groups now always persist `is_named=0`; rename no longer flips the flag; existing rows stay `is_named=1` (backfilled by migration `20260629000001` from PR #500) and are grandfathered onto their current name-text avatar.

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly what the linked task describes, no more and no less.

- **Not missing:** All four group insert paths now persist `is_named=0` — `Service.CreateGroup` (`service.go:1171`), `Service.AddGroup` (`service.go:206`), and the system/org/dept inserts in `event.go:176` / `event.go:319`. `UpdateGroupInfo` rename no longer sets `is_named=1` (`service.go`, the removed `groupModel.IsNamed = 1` line). Verified repo-wide: no other code path writes `is_named=1` after creation, so the invariant "`is_named=1` ⟺ pre-cutover legacy row" holds.
- **Not over-built:** The render logic, `avatar_text`/`avatar_color` APIs, validation, palette, and migration are untouched, matching the stated out-of-scope list. No new flag, field, endpoint, or migration was added.
- **No deviation:** Grandfather works as specified. `DB.UpdateTx` (`db.go`) writes `is_named` back from the fresh-loaded model via the full-row `SetMap`, so a rename on a legacy group preserves `is_named=1` and the name-text avatar follows the new name; a new group stays `is_named=0` and stays an icon.

Tests were updated coherently with the behavior change (`TestCreateGroup_Success`/`_AutoGenerateName` now assert `is_named=0`; `TestUpdateGroupInfo_RenameDoesNotChangeIsNamed`; `TestAddGroup_DefaultsIsNamedZero`; the PR #500 P1 clobber regression renamed to `TestRenameThenInviteUpdate_KeepsNameAndIsNamed` and re-anchored on a legacy group).

## 2. Code quality

**Quality: Approved**

Independently verified, no blocking findings:

- **Insert correctness:** `Insert`/`InsertTx` use `Columns(util.AttrToUnderscore(m)...).Record(m)`, which reflects `IsNamed` and writes `is_named=0` explicitly on every new row — correct independently of the column DEFAULT.
- **PR #500 P1 regression intact:** `UpdateInviteTx` (`db.go`) remains column-level (`invite`, `version` only), so a same-request rename+invite does not clobber `name`/`is_named`. The `groupUpdate` handler (`api.go`) still runs the name branch via `UpdateGroupInfo` (fresh load) then the invite branch via the column-level write.
- **System/org/dept groups:** Setting these to `is_named=0` is safe — `avatarGet` intercepts the `SystemGroupID` / `org_` / `dept_` prefixes with static PNGs *before* `QueryWithGroupNo`, so they never reach the `is_named` render path. The flag is kept clean as a pure legacy discriminator.

### P2 (non-blocking, advisory)

- **Stale inline comment in `DB.UpdateTx` (`modules/group/db.go:283`).** The line still reads `"is_named": model.IsNamed, // 改名置 1（用户起名）`. After this PR, rename no longer sets it to 1 — this is exactly the writeback that preserves the *loaded* value for grandfathering. The misleading comment could nudge a future editor back toward the old "rename ⇒ named" model. Suggest rewording to "preserve loaded value; rename no longer flips it." A few other doc/comment sites still encode the old semantic (`pkg/avatarrender/text.go:140` `命名群(is_named=1)`; the `20260629000001` SQL column COMMENT). Comments only, no behavior impact.

- **Public avatar endpoint enumeration / name-prefix disclosure (pre-existing, not introduced here).** `GET /:group_no/avatar` (`api.go`) is unauthenticated: it 404s for nil/disbanded groups and 200/302s for existing ones (existence enumeration), and renders the first 2 glyphs of the name for every backfilled `is_named=1` legacy row. This PR does **not** change the route, auth, or the 404/200 split — it preserves the existing exposure by design (grandfathering). Flagging for human awareness given the security-sensitive classification, but it is not a regression in this diff and the grandfather requirement makes it intentional. If the disclosure is a concern, it should be addressed as separate scope (e.g. auth on the endpoint or a constant-shape response for unknown vs known).

### Inverse read-modify-write race on `UpdateTx` (assessed, not blocking)

A cross-family advisory raised that `UpdateGroupInfo`'s full-row `UpdateTx` could, under a concurrent invite toggle, clobber the `invite` column with a stale loaded value (the mirror of the PR #500 race). This is a **pre-existing** property of `UpdateTx`, unchanged by this PR (the diff only touches comments and the removed `IsNamed=1` line in this area). The window is narrow (rename and invite-toggle on the same group within one request lifetime) and `version` bumping plus the column-level invite path mitigate the common case. Worth a follow-up to make `UpdateGroupInfo` column-level too, but out of scope here and not a blocker for this change.

## 3. Verdict

**APPROVED** — Spec ✅ and Quality Approved. The change is correct, well-scoped, and the grandfather invariant holds across all insert/update paths I verified. Findings are all P2: one stale comment worth a quick fix, and a pre-existing public-endpoint disclosure surface flagged for awareness (not a regression in this PR).

## 4. Suggestions

- Reword the `db.go:283` `UpdateTx` inline comment (and the other stale `is_named=1` doc sites) to the legacy/new semantic, so the repurposed meaning is the single source of truth.
- Track the `UpdateGroupInfo` full-row writeback → column-level conversion as a follow-up to fully close the read-modify-write window.

Jerry-Xin (dismissed):
APPROVE — re-affirming at this head.

The delta since my prior approved head (`92d61424` → `1a4d75d3`) is **comment-only** (plus a test-comment tweak) — zero behavior change, byte-verified:
- `modules/group/db.go` UpdateTx: only the inline comment on `"is_named": model.IsNamed` changed (now correctly documents "原样回写当前值；改名不再改动 is_named；仅 #500 迁移回填存量老群为 1"). The code is unchanged.
- `pkg/avatarrender/text.go` GroupNameText: only the doc comment updated to the grandfather semantics ("改版前的存量老群(is_named=1)" vs the old "命名群"). The function body is unchanged.

These resolve the comment-drift notes raised on the prior head. The grandfather logic (new groups `is_named=0` → icon; legacy backfilled groups keep name-text; rename preserves the flag; render order avatar_text > legacy name-text > icon), the `UpdateInviteTx` column-level clobber fix, ETag factors, and test coverage are all intact from the version I approved. No regression.

CI: only `check-sprint` (governance/sprint-tracking gate) is red; Build/Lint/Vet/secret-scan green. Approvable.

OctoBoooot (approved):
## Review: feat(group): default new groups to icon avatar; grandfather existing via is_named (#503) — re-review @ e4b71e64

**Verdict:** Approve — documentation-alignment delta only; the grandfather logic I approved at `92d61424` is unchanged. No blockers, no majors. CI red on `check-sprint` only; `go vet` + i18n clean.

Delta from prior APPROVE @ `92d61424`: two "docs(group): align/refresh ... comments (review nits)" commits — the doc-drift 🟡s flagged across the prior rounds.

### Verified — byte-checked

- ✅ **Code changes are comment-only** — `api.go`/`db.go`/`service.go`/`pkg/avatarrender/text.go` each have a single 1-line comment update to the grandfather semantics ("回退群名" → "回退 is_named 规则:老群群名/新群双人图标"; GroupNameText doc → "改版前的存量老群"). No logic change.
- ✅ **Version-pin test change is a comment-only clarification** — the assertion still pins the text-render version segment; only the explanatory comment was updated to note the legacy-is_named text path. Pin not weakened.
- ✅ **New migration `20260629000002` is genuinely comment-only** — `MODIFY COLUMN is_named/avatar_text` preserving the exact type/constraint/default (`TINYINT NOT NULL DEFAULT 0`, `VARCHAR(16) NOT NULL DEFAULT ''`), changing only the COMMENT text to match the grandfather rule. INFORMATION_SCHEMA-guarded (reentrant) with a Down that restores the prior comments. The comment explains the rationale (old migrations are immutable, so a no-op COMMENT refresh aligns schema introspection).
- ✅ `go vet` clean; `octo-i18n-extract -check` exit 0. Zero behavior change from the approved `92d61424` (grandfather logic, ETag mode-switch, UpdateInviteTx clobber-fix all intact).

### Praise

- **The doc-drift was closed at the schema level too, not just in code comments** — the prior rounds' 🟡s included stale column COMMENTs that a deployer reading the DB schema would see; this adds a proper comment-only migration (correct type-preserving MODIFY) to fix them, rather than leaving the schema description lying. Thorough follow-through on the review notes.

### Minor / 🟡 (non-blocking)

- The comment-refresh `MODIFY COLUMN` migration is a deploy-time lock consideration on a large `group` table. MySQL 8.0 does comment-only changes via `ALGORITHM=INSTANT` (no rebuild), but older versions copy the table. If prod is pre-8.0 or the table is very large, worth confirming the instant path or scheduling accordingly. Non-blocking.

- CI `check-sprint` — admin Sprint-field gate, independent of the code.

Jerry-Xin (approved):
In scope for octo-server: this PR changes the group module’s avatar-selection persistence behavior, docs, migration comments, and tests.

💬 Non-blocking

🟡 Warning: `DB.UpdateTx` still performs a full-row update, including `name`, `is_named`, `notice`, `status`, `forbidden`, and `invite` at [modules/group/db.go](/Users/mlamp/.openclaw/tmp/pr-review-q3i5eutr/repo/modules/group/db.go:280). The PR correctly preserves `is_named` on rename and keeps the invite path column-scoped, but concurrent updates from other paths can still clobber stale fields. This is already called out as follow-up scope and is not introduced here.

🔵 Suggestion: The PR description says “No new migration,” but the diff adds the comment-only migration [modules/group/sql/20260629000002_refresh_avatar_comments.sql](/Users/mlamp/.openclaw/tmp/pr-review-q3i5eutr/repo/modules/group/sql/20260629000002_refresh_avatar_comments.sql:1). Consider aligning the PR text so reviewers and release notes do not miss the migration.

✅ Highlights

The load-bearing write-policy is consistent across the main creation paths: `AddGroup` sets `IsNamed: 0` at [modules/group/service.go](/Users/mlamp/.openclaw/tmp/pr-review-q3i5eutr/repo/modules/group/service.go:203), `CreateGroup` sets `IsNamed: 0` at [modules/group/service.go](/Users/mlamp/.openclaw/tmp/pr-review-q3i5eutr/repo/modules/group/service.go:1168), and system/org/dept event inserts do the same at [modules/group/event.go](/Users/mlamp/.openclaw/tmp/pr-review-q3i5eutr/repo/modules/group/event.go:173).

The render priority remains unchanged and correctly depends on `avatar_text` before legacy `is_named == 1` name text, then icon fallback at [modules/group/api.go](/Users/mlamp/.openclaw/tmp/pr-review-q3i5eutr/repo/modules/group/api.go:441).

The new tests cover the important behavior: new-group `is_named=0`, rename preserving both new and legacy flags, direct insert behavior, and invite update non-clobbering.

Verification run:
`go test ./pkg/avatarrender/...`
`go test ./modules/group -run '^$'`

I did not run DB-backed group endpoint tests, which require the project’s MySQL/Redis/WuKongIM test environment.

yujiawei (approved):
# Code Review — PR #503 (octo-server)

**Verdict: APPROVED.** No P0/P1 issues found. The change is well-scoped, the migration is genuinely comment-only, and all group-creation paths are consistent. Findings below are P2/advisory and pre-existing (already tracked as out-of-scope follow-ups).

## Summary

This PR pivots the group default-avatar behavior: newly created groups now default to the two-person icon, and the group name is never rendered as avatar text — text appears only when a user sets a custom `avatar_text`. Existing groups are grandfathered (kept on their current name-text avatar) so no historical group flips to an icon at once.

The implementation correctly changes **who gets `is_named=1`** rather than the render rule. `writeGroupDefaultAvatar` is unchanged in logic (`avatar_text` > `is_named==1` name-text > icon). `is_named` is repurposed from "user named it" to "pre-cutover legacy group".

## Spec compliance

The diff implements exactly what the linked task describes — nothing missing, nothing extra:

- **All creation paths persist `is_named=0`**, verified across the whole module:
  - `Service.CreateGroup` — `service.go:1171` (removed the old `isNamedVal = name != ""` computation)
  - `Service.AddGroup` — `service.go:206` (removed the name-based computation)
  - System group insert — `event.go:176`
  - Org/dept group insert — `event.go:319`
- **Rename no longer flips the flag** — `service.go:1885` removed `groupModel.IsNamed = 1`; `UpdateTx` writes back the loaded value unchanged (`db.go:283`).
- **Render rule untouched** — `api.go:452-456`, still `avatar_text` > `is_named==1` name-text > icon; ETag factors unchanged.
- **Migration `20260629000002` is comment-only** — confirmed the `MODIFY COLUMN` statements restate the existing definitions exactly (`is_named TINYINT NOT NULL DEFAULT 0`, `avatar_text VARCHAR(16) NOT NULL DEFAULT ''`); only the `COMMENT` text differs. No type/constraint/default/nullability change, and the `INFORMATION_SCHEMA` guard makes it reentrant. The `Down` path correctly restores the prior comments.
- No accidental scope creep; the historical migration `20260629000001` is reused as-is.

`go build ./modules/group/...` and `go vet ./modules/group/...` both pass. The `strings` import in `service.go` is still used (`TrimSpace`/`Join`), so removing the name-based computation left no dead import.

## Code quality

Tests are updated coherently with the new semantics: `TestCreateGroup_Success`/`_AutoGenerateName` now assert `is_named=0`; `TestUpdateGroupInfo_RenameDoesNotChangeIsNamed` covers both new (stays 0) and legacy (stays 1) groups; `TestRenameThenInviteUpdate_KeepsNameAndIsNamed` preserves the prior P1 regression coverage using a legacy-group fixture; `TestAddGroup_DefaultsIsNamedZero` covers the direct-insert path. The system/org/dept inserts are correctly excluded from the render path (served static PNGs via the prefix branches at `api.go:360-394`), and setting them to `0` keeps the "`is_named=1` ⟺ pre-cutover legacy row" invariant clean.

### P2 / advisory (non-blocking, pre-existing)

These are not introduced by this PR and are already noted as out-of-scope follow-ups in the task brief. Listing them so they aren't lost:

1. **Full-row rename writeback can clobber concurrent column-level updates.** `UpdateGroupInfo` loads a full row (`service.go:1861`) and writes through the full-row `UpdateTx` (`db.go:280`), which rewrites `is_named`/`status`/`forbidden`/`invite` from the loaded snapshot. The prior fix (`UpdateInviteTx`) only closes the invite→rename direction; the inverse window (rename committing a stale `invite`) remains. Converting `UpdateGroupInfo` to a column-scoped write for `name`/`notice`/`version` would close it. Pre-existing; not worsened here.

2. **Unauthenticated avatar endpoint discloses legacy group name glyphs.** The public `GET /:group_no/avatar` returns 404 for missing/disbanded groups but renders the first two name glyphs for grandfathered (`is_named=1`) groups, distinguishing live vs nonexistent groups by anyone holding a `group_no`. This is intentionally preserved by grandfather, and the change actually **reduces** disclosure going forward (new groups render an icon, not name text). Hardening (auth / signed URLs / constant-shape response) is separate scope.

## Reviewer notes / coverage

- This review covered the diff plus static inspection of the checked-out repo at head SHA `e4b71e64`. Runtime migration behavior against real MySQL and the integration tests that need MySQL/Redis/WuKongIM were not executed locally — those run in CI and should be confirmed green before merge.
- A second cross-family advisory pass was attempted but timed out and did not return findings; it is recorded here as absent (not as agreement), and did not affect this verdict.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#503 Review Report (automated review)
**PR:** Mininglamp-OSS/octo-server #503
**Head SHA:** `e4b71e646053960469c98e88645202ca230df482`
**Merge base:** `db9118f8`
**Title:** feat(group): default new groups to icon avatar; grandfather existing via is_named
**Files:** 14 changed (+320 / −102)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| All new-group creation paths set `IsNamed=0` | ✅ | `service.go:206` (AddGroup), `service.go:1171` (CreateGroup), `event.go:176` (system group), `event.go:319` (org/dept group) — all hardcode `IsNamed: 0` |
| Rename no longer flips `is_named` | ✅ | `service.go:1885` — removed `groupModel.IsNamed = 1`; preserves loaded value |
| Render rule unchanged | ✅ | `api.go:449-455` — cascade `avatar_text > is_named==1 name-text > icon` unchanged in logic, only comments reworded |
| Migration `20260629000002` is comment-only | ✅ | `sql/20260629000002_refresh_avatar_comments.sql` — `MODIFY COLUMN` only refreshes COMMENT, INFORMATION_SCHEMA guard, reentrant |
| Test coverage for both new and legacy groups | ✅ | `service_test.go` — 4 tests updated: CreateGroup, AutoGenerateName, RenameDoesNotChangeIsNamed (both paths), KeepsNameAndIsNamed, DefaultsIsNamedZero |
| Avatar endpoint tests cover both paths | ✅ | `avatar_get_test.go` — `TestGroupAvatarGetNamedRendersNameText` (legacy→name-text), `TestGroupAvatarGetAutoNamedRendersIcon` (new→icon) |
| Swagger/OpenAPI spec consistent with code | ✅ | `swagger/api.yaml` — `is_named` and `avatar_text` descriptions updated in all 5 locations |
| `strings` import not orphaned in service.go | ✅ | `strings.TrimSpace` still used in `CreateGroup` (groupName, member dedup) and elsewhere |

## 2. Findings

### No P0 or P1 findings.

### P2-1: Pre-existing full-row writeback in `UpdateTx` (pre-existing, not amplified)

**Diff-scope:** Pre-existing. PR did not introduce or amplify.

`db.go:280-291` — `UpdateTx` writes back 8 columns (`name`, `is_named`, `notice`, `creator`, `status`, `version`, `forbidden`, `invite`) from the loaded model. In `UpdateGroupInfo`, the model is loaded fresh from DB, modified (name only), then written back. Concurrent mutations to `notice`/`status`/`forbidden`/`invite` between load and write could be clobbered by the stale snapshot.

**Mitigation:** The invite branch already uses column-level `UpdateInviteTx` to avoid the known clobber (PR#500 P1 fix). The remaining window (notice/status/forbidden) is narrow and low-traffic. The PR actually *improves* the situation by removing `is_named` from the set of columns that change during rename.

**Suggestion:** Consider converting `UpdateGroupInfo` to column-scoped writes (only `name` and/or `notice` + `version`) as a follow-up, matching the pattern of `UpdateInviteTx` and `updateAvatarCustom`.

### P2-2: Unauthenticated avatar endpoint information disclosure (pre-existing, acknowledged)

**Diff-scope:** Pre-existing. PR did not introduce or amplify.

`api.go:356-440` — `GET /:group_no/avatar` is unauthenticated. It distinguishes existing vs. non-existing/disbanded groups (200 vs 404) and renders legacy groups' first-2 name glyphs. This is intentional and acknowledged in the `.octospec` brief's follow-ups section.

### P2-3: ETag mode switch edge case for legacy groups clearing custom text (minor UX note)

**Diff-scope:** Pre-existing behavior, new semantics.

A legacy group (`is_named=1`) with custom `avatar_text` that clears it (empty string via `updateAvatarCustom`) falls back to name-text rendering. The ETag switches from `group-name-v4(...custom-text)` to `group-name-v4(...name-first-2)`. If the custom text happened to equal the name's first 2 glyphs, the ETag text factor is identical but the version prefix differs only if the text content matches — this is correctly handled by the `avatarrender.ETag` function. No cache-staleness risk.

## 3. Suggestions

1. **Follow-up (non-blocking):** Convert `UpdateGroupInfo` from full-row `UpdateTx` to column-scoped writes, closing the remaining read-modify-write window for concurrent notice/status/forbidden mutations.

2. **Follow-up (non-blocking):** Consider hardening the public avatar endpoint against group-existence enumeration (constant-shape 404 or auth gate), as noted in the brief.

## 4. Extra Observations

- The `.octospec/tasks/group-avatar-name-no-text/brief.md` is thorough and internally consistent. The design decision to repurpose `is_named` (rather than retire it + backfill `avatar_text`) is well-reasoned — it avoids freezing legacy avatars and doesn't require Go-side script-aware derivation in SQL.
- The `handleRegisterUserEvent` and `handleOrgOrDeptCreateEvent` correctly set `IsNamed=0` even though these group types are served by static PNG short-circuits in `avatarGet` (lines 360-397) and never reach the `is_named` render branch. This keeps the invariant "`is_named=1` ⟺ pre-cutover legacy row" clean.

## 5. Data-Flow Tracing

| Consumed datum | Upstream source | Flows correctly? |
|---|---|---|
| `groupInfo.IsNamed` in `writeGroupDefaultAvatar` (`api.go:454`) | DB `group.is_named` via `QueryWithGroupNo` → `Model.IsNamed` | ✅ New groups: all 4 insert paths set 0. Legacy: migration `20260629000001` backfilled 1. No post-creation writer exists. |
| `groupInfo.AvatarText` in `writeGroupDefaultAvatar` (`api.go:450`) | DB `group.avatar_text` via same query. Set on insert (`CreateGroup:1180`) or via `updateAvatarCustom` (column-level). | ✅ Empty for default; non-empty when user explicitly sets custom text. |
| `groupInfo.Name` in `writeGroupDefaultAvatar` (`api.go:455`) | DB `group.name` via same query. Set on insert or via `UpdateGroupInfo` rename. | ✅ For legacy groups, name follows renames. For new groups, name is never consumed for avatar (correct). |
| `groupInfo.AvatarColor` in `writeGroupDefaultAvatar` (`api.go:457`) | DB `group.avatar_color` (nullable *int). Set on insert or via `updateAvatarCustom`. | ✅ nil→derived color; non-nil→palette index. |
| `GroupResp.IsNamed` in API response (`service.go:802`) | `Model.IsNamed` via `from`/`fromModel` mapping (`service.go:849,894`). | ✅ New groups return 0; legacy return 1. Clients can predict avatar behavior. |
| `is_named` column COMMENT (migration `20260629000002`) | Static SQL text. | ✅ INFORMATION_SCHEMA guard ensures column exists before MODIFY. Idempotent. |

**C1 — Dual-path parity (含守卫/lint 工具自身等价写法覆盖):**
All 4 group creation paths (`CreateGroup`, `AddGroup`, system-group event, org/dept event) consistently set `IsNamed: 0`. Rename path (`UpdateGroupInfo`) consistently preserves the loaded value. No asymmetric paths found. **Clear.**

**C2 — Control-flow ordering / 嵌套复用 + 安全控件非 canonical 试穿:**
`writeGroupDefaultAvatar` is the sole consumer of `IsNamed` at runtime. It is called only from `avatarGet` after the system/org/dept short-circuits. No nested or repeated invocation. No security controls (regex/escape) to test. **Clear.**

**C3 — 授权边界 ≠ 能力边界:**
`is_named` is not used for authorization. It is an avatar-rendering discriminator. The public avatar endpoint was already unauthenticated before this PR. **N/A.**

**C4 — 授权生命周期 / 容器-成员状态级联:**
No auth changes in this PR. **N/A.**

**C5 — build/note 通过 ≠ 运行期路径正确:**
No build artifacts, extensions, CLI tools, or relative-path changes. Migration is standard SQL DDL. **N/A.**

**C6 — 治理/策略/安全文档自洽性:**
`.octospec/` task brief and log are internally consistent. No conflict with SECURITY.md or other governance docs. **N/A.**

N/A — automated review review, no prior rounds.

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. The semantic repurposing of `is_named` is correctly implemented across all write paths, the render rule is unchanged, tests cover both new and legacy groups comprehensively, and the migration is safely idempotent. Two pre-existing P2 items (full-row writeback race, unauthenticated endpoint enumeration) are acknowledged as follow-ups.

PR #510, opened 2026-06-30
feat(sticker): gate handle enforcement behind system_setting sticker.handle_required

an9xyz:
## Summary

Follow-up to #509. Keeps the security win of the signed sticker upload handle, but stops `OCTO_MASTER_KEY` from implicitly changing the new-sticker protocol the moment it is present.

#509 made `POST /v1/sticker/user` reject a missing handle whenever `stickersig.Enabled()` (a usable `OCTO_MASTER_KEY`) was true. Since the master key is a **mandatory production contract** (`modules/common` also needs it to encrypt the IM RSA private key at rest), enforcement was effectively always on — silently flipping the registration protocol and breaking older clients that do not yet send a handle, with no gradual rollout, rollback, or observability.

This splits **capability** from **enforcement policy** (deliberately orthogonal, never derived from one another — decoupled down to the config medium):

- `stickersig.Enabled()` — server CAN mint/verify handles (env `OCTO_MASTER_KEY`). Unchanged.
- `SystemSettings.StickerHandleRequired()` — NEW enforcement policy, the DB-backed `system_setting` key `sticker.handle_required` (default `false`). Only when `true` is a missing handle rejected.

**Why system_setting, not an env var:** the rollout's whole point is a gradual, reversible flip. `system_setting` delivers that — admin-console toggle, ~60s multi-replica convergence via the SystemSettings snapshot, no redeploy/restart — and matches how every sibling client-facing toggle already works (`login.local_off`, `space.disable_user_create`, `incomingwebhook.member_can_broadcast`).

## Related Issue

Closes #511. Sub-issue of #26 (custom sticker tracking); follows up #509.

## Linked Spec

`.octospec/tasks/sticker-handle-enforcement-rollout/brief.md` (in this PR).

## Changes

- **`system_setting sticker.handle_required`** (bool, default false): schema entry in `system_setting_schema.go` + `SystemSettings.StickerHandleRequired()` getter (mirrors `local_off` etc.). Admin-toggleable, hot-reloaded via the SystemSettings snapshot.
- **`pkg/stickersig`**: reverted to a pure capability/signing leaf package — `Enabled()` + `Sign()` + `Verify()`. The `Required()` policy getter and the `OCTO_STICKER_HANDLE_REQUIRED` env are removed; package doc points enforcement at the system_setting.
- **`modules/sticker/api.go`**: `stickerPathTrusted` (bool) → `classifyStickerPath` (four-state). Decision per outcome, policy read from `s.settings.StickerHandleRequired()`:
[...]

PR #508, opened 2026-06-30
feat(sticker): per-user custom sticker management

an9xyz:
## Summary

Adds a first-class per-user custom sticker feature so the sticker panel in octo-web stops returning 404 on every chat mount (issue #26). Stickers are flat (no packs), scoped to the individual user, with an admin-configurable per-user quota.

## Related Issue

Refs #26

## Linked Spec

`.octospec/tasks/custom-sticker-management/brief.md` (goal / load-bearing list / out-of-scope / acceptance). Implementation journal: `.octospec/tasks/custom-sticker-management/journal.md`.

## Changes

- New `modules/sticker/`: `GET/POST/DELETE /v1/sticker/user` (list / add / delete). An empty collection returns `200 {"list":[]}` instead of 404 — the original #26 symptom. Ownership-checked soft delete; `AuthMiddleware` + `SharedUIDRateLimiter`.
- i18n error envelope via `pkg/errcode/sticker.go` (+ zh-CN copy, regenerated en-US markers); `TestStickerNoLegacyResponseError` guard.
- Admin-configurable per-user quota: new `system_setting` key `sticker.user_max_count` (default 100, SuperAdmin-tunable, hot-reloaded). Returns `409 err.server.sticker.quota_exceeded` when exceeded.
- `modules/file`: widen the sticker upload from a hardcoded `.gif` to gif/png/jpg/jpeg/webp (extension derived from the requested filename); add a 1 MB `StickerMaxFileSize` cap.
- Module wired into `internal/modules.go`.

## Testing

Integration tests run against MySQL + Redis + WuKongIM:

```
$ go test ./modules/sticker/ -v
ok  github.com/Mininglamp-OSS/octo-server/modules/sticker
```

- `TestSticker_ListEmpty` (200 `{list:[]}`, the #26 regression guard), `TestSticker_AddAndList`, `TestSticker_AddFormatRejected`, `TestSticker_AddEmptyPathRejected`, `TestSticker_QuotaExceeded` (409 at an admin-set cap), `TestSticker_DeleteOwnership`.
- `TestStickerNoLegacyResponseError` (no raw error responses) + `TestRespondStickerHelpers` (localized envelope rendering).
- `make i18n-extract-check` + `make i18n-lint` pass; `modules/file` and `modules/common` suites green on a clean DB.

- [x] Unit tests added/updated
- [x] Manually verified

## COMPREHENSION

[...]

PR #512, opened 2026-06-30
feat(sticker): add metadata editing and observability

an9xyz:
## Background

This completes the next custom-sticker increment after upload-handle hardening: metadata editing, autocomplete metadata, and production observability for sticker upload/registration.

## Linked Spec

- `.octospec/tasks/sticker-edit-shortcode-observability/brief.md`

## Changes

- Add `PUT /v1/sticker/user/:sticker_id` for partial metadata updates (`placeholder`, `sort`, `shortcode`, `keywords`).
- Return `sort`, `shortcode`, and `keywords` from sticker create/list/update responses.
- Order user stickers by `sort ASC, id DESC` so legacy `sort=0` rows keep newest-first behavior.
- Add `shortcode` / `keywords` columns and an index for per-user live shortcode lookups.
- Enforce shortcode normalization/validation and same-user live conflict checks inside the existing user-row-lock transaction.
- Add low-cardinality Prometheus counters and structured logs for sticker uploads and registrations.
- Add sticker-upload-only UID throttling and optional per-IP throttling using shared wkhttp rate-limit middleware.
- Register new sticker error codes and zh-CN translations.

## COMPREHENSION

1. What ownership boundary is load-bearing?
   Sticker metadata remains scoped to the login UID. Update uses `sticker_id + uid + status=1`, cross-user or missing stickers return the same not-found envelope, and shortcode conflicts are scoped to the same UID's live stickers.

2. What wire contracts stay compatible?
   Sticker errors continue through `httperr.ResponseErrorL`; list still returns `{ "list": [] }` for empty results; new response fields are additive; `sort=0` legacy rows keep their newest-first tie-breaker.

3. What operational controls matter?
   Metrics use bounded `result` labels only, upload throttling uses shared middleware instead of ad-hoc Redis counters, and the migration is guarded so a partial DDL retry does not fail on duplicate columns.

## Tests

- `go test ./modules/sticker ./modules/file ./pkg/errcode ./pkg/i18n/...`
- `make i18n-extract-check`
- `make i18n-lint`
- `git diff --check origin/main...HEAD`


OctoBoooot (dismissed):
## Review: feat(sticker): metadata editing + observability (#512) — APPROVE

New `PUT /v1/sticker/user/:sticker_id` (partial metadata: placeholder/sort/shortcode/keywords) + autocomplete fields + Prometheus observability. Byte-verified the write path, uniqueness, migration, and inputs. No blockers, no majors.

### Verified — done right
- ✓ **Ownership enforced.** `update()` loads via `queryByIDAndUIDTx` (`WHERE sticker_id=? AND uid=? AND status=1`), so a non-owner gets `nil → ErrStickerNotFound` — can't edit another user's sticker. Test `TestSticker_UpdateOwnershipAndValidation` pins it with an `other-uid` row.
- ✓ **Partial update is safe.** Pointer fields (`*req.Placeholder` etc.) → only provided fields mutate; each is validated (placeholder rune-len, sort, shortcode normalize, keywords normalize) before write.
- ✓ **Per-user shortcode uniqueness, no TOCTOU.** `update()` runs `lockUserRowTx(loginUID)` → `shortcodeExistsTx` (uid-scoped, excludes self, skips empty) → `updateMetadataTx` → commit, all one tx. The user-row lock serializes same-user writes, so the non-unique `idx_uid_status_shortcode` is fine (uniqueness is per-uid by design). Conflict → `shortcode_conflict` (tested).
- ✓ **Shortcode input is whitelisted** — `^[a-z0-9_]{2,32}$`, lowercased/trimmed; no injection/control chars. Invalid (`笑哭`) → `shortcode_invalid` (tested).
- ✓ **Migration is the safe pattern.** INFORMATION_SCHEMA-guarded reentrancy on every ALTER (both columns + index), additive (no data loss), procedure-based idempotency explicitly justified for partial-failure retry, and a correct **Down** that reverses index→keywords→shortcode. Matches the migration-safety bar from prior rounds.
- ✓ **Observability** — promauto CounterVecs with label series pre-initialized to 0; i18n keys synced between zh-CN runtime locale and en-US marker. Build green.

### Minor (non-gating)
- **Keywords are free-text without a charset whitelist** (only per-item rune-len, count cap, JSON store-size bound), unlike shortcode's strict regex. They're JSON-encoded for storage (storage-safe) but rendered in client autocomplete — so XSS-safety depends on the client escaping them at render. That's a reasonable design (keywords are inherently free-form, can't whitelist), but worth a note that the render side must escape, and/or a control-char strip on store for defense-in-depth.

### Net
**APPROVE.** Clean metadata-edit increment: ownership-scoped write, per-user shortcode uniqueness serialized under the user-row lock (no TOCTOU), whitelisted shortcode, safe reentrant+reversible migration, strong tests (ownership-rejection, shortcode normalize/invalid/conflict, sort order), and proper metrics/i18n. Only minor: keywords render-escaping is a client responsibility worth documenting (+ optional control-char strip on store). Build green; the local `modules/sticker` test failure is the env's missing-MySQL `testutil.NewTestServer` (`dial tcp 3306`), not a PR defect. CI: needs approvals + Test.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-server PR#512 Review Report
**Head SHA:** `628033f4ac7333a18156027da394facb9d60bca2`
**PR:** feat(sticker): add metadata editing and observability
**Changed files:** 15 (+1007 / −39)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| PUT ownership scoping | ✅ | `queryByIDAndUIDTx` filters `sticker_id=? AND uid=? AND status=1` (db.go:83–89); cross-user update returns not_found (test: `TestSticker_UpdateOwnershipAndValidation`) |
| Shortcode validation & conflict | ✅ | `normalizeStickerShortcode` regex `^[a-z0-9_]{2,32}$` (model.go:67); `shortcodeExistsTx` runs inside `lockUserRowTx` transaction (api.go:200–210, 296–305) |
| Keywords normalization | ✅ | Dedup via `seen` map, rune-length cap 20, count cap 10, JSON store cap 255 bytes (model.go:77–100); `decodeStickerKeywords` handles empty/invalid JSON gracefully (model.go:102–113) |
| List ordering | ✅ | `listByUID` now `OrderAsc("sort").OrderDesc("id")` (db.go:32–35); test `TestSticker_UpdatePartialAndListSortOrder` verifies sort=5 outranks sort=10 |
| Migration idempotency | ✅ | `IF NOT EXISTS` guards on columns and index via stored procedure (sql/20260630000001); new columns default to `''` — legacy rows unaffected |
| Metric cardinality | ✅ | Only `result` label used; all values pre-registered in `init()` (sticker_metrics.go:29–35, metrics.go:29–31); no uid/path/shortcode in labels |
| Rate limiting scoping | ✅ | `stickerOnlyLimiter` wrapper bypasses limiter for `type≠sticker` (api.go:95–101); both IP and UID limiters verified to call `c.Next()` on allow, `c.Abort()` on deny |
| Handle trust refactor | ✅ | `stickerPathTrustResult` returns low-cardinality reason; `stickerPathTrusted` kept as bool wrapper for tests (api.go:244–263); external contract unchanged |
| i18n / errcode | ✅ | 3 new error codes registered (errcode/sticker.go); zh-CN + en-US translations present; `make i18n-extract-check` and `make i18n-lint` listed in PR tests |
| Test coverage | ✅ | 9 new test functions covering update, shortcode, keywords, conflict, metrics, cross-UID scoping, validation |

## 2. Findings

### No P0 or P1 findings

All new code paths were traced to their data sources and verified. No path produces user-visible errors, makes working paths unavailable, or worsens existing behaviour.

### P2 Findings (non-blocking)

**P2-1 — `updateMetadataTx` writes all 4 fields unconditionally**
- **diff-scope:** new (introduced by this PR)
- **File:** `modules/sticker/db.go:131–137`
- `updateMetadataTx` always SETs `placeholder`, `sort`, `shortcode`, `keywords` regardless of which fields the caller actually changed. The handler correctly preserves existing values from the read-model before writing, so this is **functionally correct**. However, it means a `PUT {"sort": 5}` still writes the unchanged placeholder/shortcode/keywords back to the row. A selective update (only SET changed columns) would be more efficient and reduce write amplification.
- **Impact:** Negligible for single-row updates. Cosmetic/efficiency concern only.

**P2-2 — IP rate-limiter Redis client created per route registration, not shared**
- **diff-scope:** new
- **File:** `modules/file/api.go:84–88`
- `stickerUploadHandlers()` creates a new `octoredis.NewInstrumentedClient` each time it's called. Unlike `SharedUIDRateLimiter` (which is a process-level singleton), this IP-limiter client is a dedicated instance. Since `Route()` is called once at startup, this produces exactly one extra Redis client — acceptable, but inconsistent with the singleton pattern used elsewhere.
- **Impact:** One extra Redis connection pool (size 10) per process. Not a concern at normal scale.

**P2-3 — No upper-bound validation on `sort` value**
- **diff-scope:** new
- **File:** `modules/sticker/api.go:287–290` (update handler), `api.go:153–156` (add handler)
- Both handlers reject `sort < 0` but impose no upper bound. The DB column is `INT` (max 2147483647). Extremely large values would fail at the DB layer with a MySQL error surfaced as `store_failed`. Not user-exploitable but produces a 500 instead of a clean 400 for pathological input.

## 3. Suggestions

1. Consider a selective-update builder in `updateMetadataTx` — track which fields were provided and only SET those columns. Minor efficiency gain and clearer intent.
2. Add `sort` upper bound (e.g., `sort > 1_000_000 → request_invalid`) to surface a clean 400 for pathological values.

## 4. Data Flow Trace

| Consumer | Source | Flow Verified |
|----------|--------|---------------|
| `toStickerResp(m).Keywords` (response) | `m.Keywords` (DB VARCHAR) → `decodeStickerKeywords` → `json.Unmarshal` | ✅ empty/invalid → `[]string{}`; valid JSON → decoded slice |
| `toStickerResp(m).Shortcode` (response) | `m.Shortcode` (DB VARCHAR) | ✅ direct field copy |
| `listByUID` sort order | `ORDER BY sort ASC, id DESC` | ✅ legacy sort=0 rows get newest-first via id DESC tiebreaker |
| `shortcodeExistsTx` conflict check | `SELECT sticker_id FROM sticker WHERE uid=? AND status=1 AND shortcode=? [AND sticker_id<>?] LIMIT 1` | ✅ runs inside `lockUserRowTx` transaction; `status=1` excludes soft-deleted; `excludeStickerID` prevents self-conflict on update |
| `observeStickerUpload` metrics | Called at every rejection/success branch in `uploadFile` | ✅ all exit paths covered (path_rejected, read_failed, size_rejected, format_rejected, magic_rejected, dimension_rejected, upload_failed, success) |
| `stickerOnlyLimiter` chain | `c.Query("type")` → `TypeSticker` check | ✅ non-sticker uploads bypass limiter via `c.Next()`; sticker uploads invoke limiter which calls `c.Next()` on allow, `c.Abort()` on deny |
| IP rate limiter fail mode | `StrictIPRateLimitMiddleware` | ✅ fail-closed: unknown IPs share a global bucket (not skipped) |
| UID rate limiter fail mode | `SharedUIDRateLimiter` | ✅ fail-open: missing uid passes through (auth middleware guarantees uid for authenticated routes) |

## 5. Blind Spot Checklist (R5)

- **C1 — Dual-path parity:** N/A. No add↔remove or create↔delete symmetry changes. The `update` endpoint is a new CRUD operation, not a symmetric pair.
- **C2 — Control-flow ordering / nested reuse:** Clear. `stickerOnlyLimiter` wraps two independent limiters (IP, UID) in sequence. Both follow the standard `c.Next()` / `c.Abort()` contract. No double-action or ordering inversion.
- **C3 — Authorization boundary:** Clear. `PUT /v1/sticker/user/:sticker_id` is scoped to `uid=? AND status=1` in the DB query (not just URL param). Cross-user updates return `not_found`. No new capability exposed beyond owner.
- **C4 — Authorization lifecycle / container-member state:** Clear. Sticker ownership is per-uid (no parent container that can be independently disabled). The `status=1` filter in `queryByIDAndUIDTx` excludes soft-deleted stickers.
- **C5 — Build ≠ runtime:** N/A. No build artifacts, CLI tools, or relative-path concerns.
- **C6 — Governance docs:** N/A. Not a governance/policy PR.

## 6. Verdict

**[Octo-Q] verdict: APPROVE**

This is a well-structured PR with strong observability instrumentation, proper ownership scoping, transactional conflict detection, and comprehensive test coverage. All three P2 findings are non-blocking efficiency/cosmetic concerns. No P0 or P1 issues found.

yujiawei (dismissed):
# Code Review — PR #512 (octo-server)

**Verdict: APPROVE.** This is the next custom-sticker increment after the upload-handle hardening, and it lands cleanly. The diff implements exactly what the linked spec asks for, the security-sensitive surfaces (cross-user metadata edit, shortcode uniqueness, upload throttling, observability cardinality) are all handled correctly, and test coverage is good. No blocking issues. A few non-blocking notes below.

Reviewed at head SHA `628033f4ac7333a18156027da394facb9d60bca2`. `go build` / `go vet` clean; the DB-free `modules/file` unit tests (including sticker-upload + limiter-scoping) pass. The `modules/sticker` server-backed tests could not execute locally due to a stale shared test-DB migration state in the test harness (`Table 'message_extra' already exists` raised in test-server bootstrap, before any PR code runs) — an environment issue, not a defect in this change.

## 1. Spec compliance

All load-bearing requirements from `.octospec/tasks/sticker-edit-shortcode-observability/brief.md` are met, with nothing missing and nothing built beyond scope:

- **`PUT /v1/sticker/user/:sticker_id` partial update** — `modules/sticker/api.go` `update()`. Pointer fields (`*string`/`*int`/`*[]string`) give correct omitted-vs-explicit semantics; empty `placeholder` restores the default; omitted fields keep prior values.
- **Cross-user / missing → not-found envelope** — `queryByIDAndUIDTx` filters `sticker_id=? AND uid=? AND status=1`; a non-owned/missing row returns `ErrStickerNotFound`, matching the existing privacy posture of `delete()` (no existence enumeration).
- **List ordering `sort ASC, id DESC`** — `db.go` `listByUID` adds `OrderAsc("sort")` before the existing `OrderDesc("id")`, preserving legacy newest-first for `sort=0` rows.
- **Per-uid live shortcode uniqueness with soft-delete release** — `shortcodeExistsTx` scopes on `uid` + `status=1`; `excludeStickerID` lets a sticker keep its own shortcode on update; checked inside the same `lockUserRowTx` (FOR UPDATE on the user row) transaction as the quota check.
- **Keywords as JSON array** — request/response expose `[]string`; storage is a bounded JSON string in `VARCHAR(255)` (≤10 keywords, ≤20 runes each, byte-bounded marshal).
- **Registered errcodes + i18n** — `ErrStickerShortcodeInvalid` / `ErrStickerKeywordsInvalid` / `ErrStickerShortcodeConflict` registered with `SafeDetailKeys`, en-US markers and zh-CN translations added; no legacy raw responses in `modules/sticker`.
- **Observability with bounded labels** — both `metricStickerUploadTotal`/`metricStickerUploadHandleTotal` (file) and `metricStickerRegisterTotal` (sticker) carry only a `result` label; all call sites pass enumerated constants (pre-seeded via `init()`). uid/format/size/dimensions/reason live in structured logs, never in label values.
- **Shared-middleware rate limit** — sticker upload throttling uses `SharedUIDRateLimiter` (+ optional per-IP `StrictIPRateLimitMiddleware`), not a hand-written Redis counter.

The per-IP throttle is in scope ("sticker uploads receive an explicit endpoint-level throttle"), env-gated off by default, and does not constitute over-build.

## 2. Code quality

No P0/P1. The implementation is careful and matches established patterns in the repo.

Verified in particular:
- **Transaction safety:** both `add()` and `update()` place `defer tx.RollbackUnlessCommitted()` immediately after `Begin()`, so every early-return path (validation, quota, shortcode conflict, query/store error) releases the connection and the `FOR UPDATE` lock. No leak.
- **SQL safety:** `shortcodeExistsTx` builds its statement with bound parameters (`args ...interface{}` to `SelectBySql`), not string interpolation of user input. No injection.
- **IDOR:** ownership is enforced at load time (`queryByIDAndUIDTx`); `updateMetadataTx` additionally repeats the `uid` predicate in the `WHERE` as defense-in-depth.
- **Rate-limit wiring:** the group already runs behind `AuthMiddleware`, so `SharedUIDRateLimiter` can resolve the login uid; `stickerOnlyLimiter` correctly `c.Next()`s through for non-sticker traffic and delegates to the limiter (which `c.Next()`/`c.Abort()`s) for `type=sticker`. Non-sticker uploads are unaffected.
- **`queryByID` not-found change:** now returns `(nil, nil)` on `ErrNotFound`; the sole caller (`delete()`) already guards `m == nil`, so a missing sticker now yields a clean not-found instead of a 500 — an improvement.
- **Keyword bounds:** the count cap admits exactly 10 and rejects the 11th; the `len(data)` storage guard is in bytes while `VARCHAR(255)` is 255 characters under utf8mb4, so no silent truncation.
- **Migration:** idempotent via stored-procedure `IF NOT EXISTS` guards on both columns and the index, with a matching Down; safe under partial-DDL retry on rolling deploys.

### Non-blocking suggestions (P2 / nits)

- **P2 — shortcode uniqueness is app-enforced, not DB-enforced.** The migration adds only a non-unique `idx_uid_status_shortcode`; correctness relies on the check-then-write inside `lockUserRowTx`. Note that `lockUserRowTx` is a no-op (returns `nil`) when the caller has no `user` row, in which case two concurrent same-uid requests are not serialized and could both persist a duplicate live shortcode. This matches the spec's explicit decision to keep a unique index out of scope and the same lock already governs the quota check, so it is acceptable as-is — but if you ever observe such users in practice, consider either a partial/functional unique constraint on `(uid, shortcode)` for non-empty live rows, or failing closed when the user row cannot be locked.
- **P2 (UX) — keyword storage limit is in bytes.** `len(data) > 255` rejects after JSON marshal, so an otherwise-valid set of multi-byte (e.g. CJK) keywords can fail with `keywords_invalid` well before the nominal 10×20-rune ceiling. The column is `VARCHAR(255)` (255 chars), so there is headroom; consider raising the byte budget or surfacing a clearer message so clients understand the effective limit.
- **nit — redundant length check.** `normalizeStickerShortcode` does an explicit `len()` range check and then `^[a-z0-9_]{2,32}$`, which already encodes the same bounds. Harmless; the constants and the regex bound just need to stay in sync if changed.
- **nit — documentation.** A one-line comment on `stickerOnlyLimiter` noting the "limiter must call `c.Next()` on pass" middleware contract would make the chain wiring more refactor-proof.

## 3. Manual-verification call-outs (security-sensitive)

For a human spot-check: (a) confirm production deployments do not run with `OCTO_MASTER_KEY` unset where handle enforcement is expected — registration degrades to shape-check-only when no key is configured (pre-existing posture, unchanged here); (b) decide whether the byte-vs-char keyword limit above is acceptable for the target locales before release.

OctoBoooot (approved):
## Review: #512 — delta @ bba6d713 — APPROVE (rebase absorbing #510 fail-closed + metadata core intact)

Large compare (46 files, +2060) but byte-isolated: it's **2 #512 commits on a rebase of 4 main commits** (ahead:2/behind:4). The real #512 surface is `sticker/api.go`, `api_test.go`, `model.go`, `file/api.go`; the `thread/`, `stickersig`, errcode churn is the already-reviewed #463/#510 work arriving via rebase. Re-affirming APPROVE.

### Verified — security cores intact
- ✓ **`update()` (metadata edit) unchanged in substance** — ownership via `queryByIDAndUIDTx` (:361 → non-owner `nil` → `ErrStickerNotFound`), `lockUserRowTx` (:355), per-user shortcode uniqueness `shortcodeExistsTx` (:407) all present. The `+429` diff hunk near `update()` is just the `classifyStickerPath`/`stickerPathClass` definitions landing there, not new logic inside `update()`.
- ✓ **`add()` reworked to inherit #510's fail-closed enforcement** — old `stickerPathTrustResult`/`stickerPathTrusted` helpers replaced by `classifyStickerPath` + the `required := StickerHandleRequired()` / `enabled := Enabled()` snapshot + `if required && !enabled { rejected_no_capability + reject }` intercept (api.go:181). This is byte-identical to the #510 logic I approved at `8c546aa1` (fail-closed conflict handling, request-scoped snapshot). Not a regression — #512 correctly picking up the merged #510 fix.
- ✓ Metadata migration present (INFORMATION_SCHEMA-guarded + reversible, unchanged); shortcode charset `stickerShortcodeRe` intact; `model.go` change is a doc sync (handle comment now matches the #510 `handle_required` policy).
- ✓ Test coverage intact — UpdateOwnership + Shortcode-conflict + RequiredWithoutCapability tests all still present (6 matches). Build green; `pkg/metrics` tests pass. (Local `modules/sticker` failure is the env missing-MySQL `testutil.NewTestServer` — `dial tcp 3306` — not a PR defect.)

### Net
**APPROVE** (re-affirmed, SHA-pinned to `bba6d713`). The #512 metadata feature's security core (ownership, uniqueness-under-lock, migration) is unchanged from my prior approval; `add()` now inherits the byte-verified-correct #510 fail-closed enforcement via rebase; no regression. Standing non-gating 🟡s carry: keywords render-escaping is the client's responsibility (worth a store-side control-char strip), and the earlier polish notes (sort upper-bound, read-policy-once). Gate: current-head re-approvals + CI Test.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#512 Review Report
**PR:** Mininglamp-OSS/octo-server#512 — feat(sticker): add metadata editing and observability
**Head SHA:** `bba6d713bf7ae7a5e2e94e2f01ee58694d868bce`
**Files changed:** 15 (+978 / -26)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Ownership boundary (PUT) | ✅ | `queryByIDAndUIDTx` WHERE `sticker_id=? AND uid=? AND status=1` (db.go:83-90); cross-user returns not-found |
| Shortcode uniqueness scope | ✅ | `shortcodeExistsTx` filters `uid=? AND status=1 AND shortcode=?` (db.go:104-117); per-UID, soft-delete releases |
| List ordering | ✅ | `ORDER BY sort ASC, id DESC` (db.go:32-33); legacy sort=0 rows keep newest-first tie-break |
| Metric cardinality | ✅ | Only `result` label used; 8 upload + 2 handle + 10 register labels; all pre-registered in init() |
| Migration idempotency | ✅ | Guarded `IF NOT EXISTS` on columns and index (sql/20260630000001_sticker_metadata.sql) |
| Rate limiting | ✅ | UID limiter on sticker route group (api.go:99); IP limiter gated by env, sticker-only scoped |
| Wire contract compat | ✅ | New response fields are additive; empty list still `{list:[]}`; existing error envelope unchanged |
| Shortcode/keywords validation | ✅ | Regex `^[a-z0-9_]{2,32}$`, max 10 keywords × 20 runes, 255-byte JSON cap |
| Test coverage | ✅ | Update partial/list-sort, ownership/validation, shortcode CRUD, conflict scope, metrics (api_test.go, api_unit_test.go) |

## 2. Findings

### P2-1: `update()` handler lacks observability instrumentation (new, not amplified)

**Diff-scope:** New endpoint introduced by this PR without corresponding metric/log coverage.

The `add()` handler comprehensively emits `observeStickerRegister(...)` on every exit path (14 call sites) and logs success with uid/format/shortcode_set/keyword_count. The new `update()` handler (api.go:337-430, ~90 lines) emits **zero** metrics and **zero** structured logs — only error-level `s.Error(...)` on failures.

This creates an observability gap: operators can track sticker creation outcomes but not edit outcomes. If an update fails silently (e.g. DB connection pool exhaustion), no counter increments and no alert fires.

**Suggestion:** Add an `observeStickerUpdate(result string)` counter (or reuse register_total with a distinguishing label) and a success log mirroring the add() pattern. Low priority since the PR spec scopes observability to upload/registration only.

### P2-2: IP rate-limit Redis client is not a process singleton (new)

**Diff-scope:** New code in `stickerUploadHandlers()` (file/api.go:81-91).

`SharedUIDRateLimiter` in `pkg/wkhttp/ratelimit_helper.go` uses a `sync.Mutex`-guarded process singleton so all mount points share one Redis client. The new IP rate-limiter creates a fresh `octoredis.NewInstrumentedClient(...)` each time `stickerUploadHandlers()` is called (which is once at startup via `Route()`), but there's no singleton guard.

Since `Route()` is called once, this is not a runtime issue. But if a future refactor calls `stickerUploadHandlers()` multiple times or in tests, it would create redundant connection pools.

**Suggestion:** Consider a similar singleton pattern, or accept the one-shot nature and document it.

## 3. Suggestions

1. **Observability parity:** When the next sticker PR adds the server-side search endpoint (noted as out-of-scope), consider adding `update` metrics at that time for full CRUD observability.
2. **Migration testing:** The migration uses MySQL stored procedures for idempotency — confirmed correct. The `DROP PROCEDURE IF EXISTS` before and after ensures clean retry. Well done.

## 4. Additional Observations

- **`queryByID` ErrNotFound fix (db.go:95-97):** The PR adds `errors.Is(err, dbr.ErrNotFound) → return nil, nil` to `queryByID`. In the base branch, a not-found row would cause `queryByID` to return `(nil, dbr.ErrNotFound)`, and the `delete` handler would hit `if err != nil` → 500. This is a latent bug fix bundled with the feature — the delete handler now correctly returns 404 for missing stickers. Good defensive improvement.

- **Shortcode conflict check on unchanged shortcode:** When `req.Shortcode == nil` (update doesn't touch shortcode), `shortcodeExistsTx` still runs with the existing value but correctly excludes the current sticker via `sticker_id<>?`. One redundant query per update, but correctness is preserved.

- **`normalizeStickerShortcode` redundancy:** The length check `len(shortcode) < 2 || len(shortcode) > 32` is redundant with regex `^[a-z0-9_]{2,32}$`. Harmless defensive coding; not a bug.

## 5. Data-Flow Tracing

| Consumed data | Source | Flows to consumption? |
|---|---|---|
| `m.Shortcode` in update conflict check | `req.Shortcode` → `normalizeStickerShortcode()` → `m.Shortcode`, or existing `m.Shortcode` from DB | ✅ Yes — normalized and written to model before conflict check; empty short-circuits via `shortcode==""` early return |
| `m.Keywords` in update response | `req.Keywords` → `normalizeStickerKeywords()` → JSON string → `m.Keywords` | ✅ Yes — `toStickerResp` calls `decodeStickerKeywords(m.Keywords)` which unmarshals back to `[]string`; empty → `[]string{}` |
| `loginUID` in update ownership | `ctx.GetLoginUID()` from AuthMiddleware | ✅ Yes — used in `queryByIDAndUIDTx` WHERE clause, `lockUserRowTx`, and `shortcodeExistsTx` |
| `stickerOnlyLimiter` type gate | `c.Query("type")` from request URL | ✅ Yes — upload endpoint uses `?type=sticker` query param; non-sticker uploads skip limiter via `c.Next()` |
| `observeStickerUpload("success")` | End of `uploadFile()` after OSS write | ✅ Yes — only reached after all validation gates pass and upload succeeds; no double-count with rejection paths |
| `metricStickerUploadTotal` labels | `stickerUploadResultLabels()` in sticker_metrics.go | ✅ Yes — all 8 labels pre-registered in `init()`, all call sites use literal strings matching the label set |
| List order `sort ASC, id DESC` | `listByUID` in db.go | ✅ Yes — `OrderAsc("sort")` then `OrderDesc("id")`; legacy `sort=0` rows keep relative newest-first order within their group |

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete pairs are modified. The new PUT is an edit, not a create/delete pair. The existing POST (add) ↔ DELETE pair is unchanged.
- **C2 — Control-flow ordering / nesting:** Clear. `stickerOnlyLimiter` wraps the limiter in a type-check; verified non-sticker uploads skip correctly. Metric observations are placed at each distinct exit path, no double-counting. The `shortcodeExistsTx` short-circuit for empty shortcode prevents false conflicts.
- **C3 — Authorization boundary ≠ capability boundary:** Clear. `PUT /v1/sticker/user/:sticker_id` is behind `AuthMiddleware` (uid-scoped). The `queryByIDAndUIDTx` WHERE clause enforces ownership at the data layer, not just the route layer. Cross-user updates return not-found (anti-enumeration).
- **C4 — Authorization lifecycle / container-member cascade:** N/A. Sticker ownership is directly bound to uid (no parent container like space/org that could be disabled). User account status is handled by `AuthMiddleware` upstream.
- **C5 — Build ≠ runtime:** N/A. No build artifacts, extensions, CLI packaging, or relative-path changes. Standard Go server code.
- **C6 — Governance / policy document consistency:** N/A. No policy/governance documents modified. The `.octospec` brief is a task spec, not a policy document.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. Two P2 observations (update observability gap, non-singleton Redis client) are noted but do not block landing. The PR is well-structured with correct ownership enforcement, proper transactional scoping, idempotent migration, bounded metric labels, and comprehensive test coverage. The `queryByID` ErrNotFound fix is a welcome defensive improvement.

Jerry-Xin (approved):
The PR is in scope for octo-server and I did not find any blocking correctness, security, or migration issues.

**💬 Non-blocking**
- 🟡 Warning: `handle_required` is logged from `stickersig.Enabled()` rather than the actual policy snapshot. This conflates signing capability with enforcement policy, which the surrounding code explicitly treats as separate. See [modules/sticker/api.go](/Users/mlamp/.openclaw/tmp/pr-review-nqks2_bx/repo/modules/sticker/api.go:329) and [modules/file/api.go](/Users/mlamp/.openclaw/tmp/pr-review-nqks2_bx/repo/modules/file/api.go:515). Consider logging `handle_enabled` separately, and in sticker registration use the `required` variable for `handle_required`.
- 🟡 Warning: `PUT /v1/sticker/user/:sticker_id` can return a successful stale response if a concurrent delete flips `status` after `queryByIDAndUIDTx` but before `updateMetadataTx`; the `UPDATE ... status=1` affects zero rows but the caller does not inspect rows affected. See [modules/sticker/api.go](/Users/mlamp/.openclaw/tmp/pr-review-nqks2_bx/repo/modules/sticker/api.go:418) and [modules/sticker/db.go](/Users/mlamp/.openclaw/tmp/pr-review-nqks2_bx/repo/modules/sticker/db.go:128). Low probability, but returning not-found on zero affected rows would make the contract tighter.

**✅ Highlights**
- Ownership checks for update are correctly scoped by `sticker_id + uid + status=1`.
- Shortcode conflict checks are same-user and inside the user-row-lock transaction.
- New i18n error codes and response helpers follow the project’s localized envelope pattern.
- Migration is retry-safe for partial column/index application.
- Metrics use bounded result labels.

**Verification**
- Ran `go test ./modules/sticker ./modules/file ./pkg/metrics`. `modules/file` and `pkg/metrics` passed. `modules/sticker` could not complete because the local shared test harness hit an existing migration/schema collision: `Table 'message_extra' already exists handling 20210407000001_message_legacy01.sql`.

yujiawei (approved):
# Code Review — PR #512 (octo-server)

`feat(sticker): add metadata editing and observability` — reviewed at head `bba6d713bf7ae7a5e2e94e2f01ee58694d868bce` against `main` (merge-base `6dad2655`).

This is a **security-sensitive** change (ownership boundary, transactional uniqueness, rate-limit, migration). It was reviewed with extra care and cross-checked from multiple independent perspectives. Findings were reconciled and every disputed point was re-verified directly against the diff.

## 1. Spec compliance

**Spec: ✅**

The diff maps 1:1 to the linked task brief's load-bearing list, with nothing built beyond scope:

- `PUT /v1/sticker/user/:sticker_id` — partial update of `placeholder`/`sort`/`shortcode`/`keywords`; omitted (nil-pointer) fields keep prior values; cross-user/missing returns the existing not-found envelope (`modules/sticker/api.go:329-431`, `db.go:83-92`).
- List ordering `sort ASC, id DESC` preserving legacy `sort=0` newest-first (`db.go:29-38`).
- Per-uid live shortcode uniqueness enforced **inside** the user-row-lock transaction, not via a DB unique index — matching the brief's explicit "no unique index; transactional per-uid scoping" decision (`api.go:405-416`, `db.go:112-126`). Soft-deleted rows release the shortcode (predicate `status=1`).
- Keywords: request/response as JSON array, stored as bounded JSON string in `VARCHAR(255)` (`model.go:77-105`, migration).
- Error envelope: new validation/conflict cases use registered `pkg/errcode` codes + zh-CN translations; no raw error responses added in `modules/sticker` (`pkg/errcode/sticker.go`, `api_i18n.go`).
- Observability: metrics carry only the bounded `result` label — no uid/path/shortcode/keyword dimensions (`modules/sticker/metrics.go`, `modules/file/sticker_metrics.go`). Logs carry uid/format/size/dimensions/reject-reason as specified.
- Rate limit: sticker upload gets an extra endpoint throttle via shared wkhttp middleware (`StrictIPRateLimitMiddleware` + `SharedUIDRateLimiter`), scoped to `type=sticker` only, not a hand-rolled Redis counter (`modules/file/api.go:81-110`).

Out-of-scope items (full file-module i18n migration, server-side search, DB unique index) are correctly untouched. No missing, extra, or deviating behavior.

Migration is idempotent/retry-safe: guarded `information_schema` checks inside a stored procedure so a partial DDL retry won't fail on duplicate column/index.

Build passes (`go build`) and the focused `modules/file`, `pkg/errcode`, `pkg/i18n` tests pass. The `modules/sticker` DB-backed tests could not run in this environment due to a shared-MySQL migration collision in the test harness setup (pre-existing infra issue, unrelated to this diff); the sticker test bodies themselves cover concurrent-quota, ownership, shortcode-conflict-scoped-to-uid, and forged/tampered/missing-handle cases and read back persisted state after update.

## 2. Code quality

**Quality: Approved**

No correctness, security, or data-loss defects found. The transactional quota + shortcode-conflict path correctly reuses the established user-row-lock pattern to avoid the gap-lock deadlock, and closes the count→insert TOCTOU. Ownership is consistently scoped by `uid` on every read/write, with a defensive `uid` predicate even on `softDelete`.

The following are non-blocking suggestions (P2/P3):

- **P2 — `update` can return 200 without persisting under a concurrent same-user delete.** `modules/sticker/db.go:128-136`. `update` takes the user-row lock, but `delete`→`softDelete` (`api.go:501`, `db.go:143-148`) is non-transactional and takes no lock, so the two do not serialize. If a DELETE for the same sticker commits between `update`'s `queryByIDAndUIDTx` read and its `UPDATE ... WHERE status=1`, the UPDATE matches zero rows, yet `Exec()` returns nil error → the handler commits and responds 200 with the in-memory (never-persisted) metadata. Self-scoped, no cross-user or data-corruption impact, and only reachable via a client racing its own edit+delete — hence non-blocking. Suggest checking `RowsAffected() == 0` → return not-found, or routing the soft-delete through the same user-row-lock transaction.

- **P3 — success-path log field mislabels enforcement posture.** `modules/sticker/api.go:329` and the equivalent `modules/file/api.go` upload logs record `handle_required` from `stickersig.Enabled()`, which reflects the signing *capability* (master key present), not the *policy* (`SystemSettings.StickerHandleRequired()`). During rollout a deployment with a key but `sticker.handle_required=false` will log `handle_required=true`, weakening the exact security observability this PR adds. Suggest logging the `required` snapshot the handler already computes (`api.go:add()` reads it into `required`).

- **P3 — two parallel `register_total` counters.** `modules/sticker/metrics.go` registers a promauto `sticker_register_total`, while `pkg/metrics/sticker.go` already exposes `dmwork_sticker_register_total` (incremented at the same call sites via `metrics.ObserveStickerRegister`). Distinct fully-qualified names, so no registration panic, but the registration outcome is now emitted on two series with different namespaces and slightly different label vocabularies. Consider consolidating on one to avoid dashboard/alert ambiguity.

### Note on a disputed finding
One review perspective raised a **P0 "missing `tx.Commit()` in the update handler."** This was re-verified directly against the diff and is **incorrect** — `tx.Commit()` is present at `modules/sticker/api.go:423`, and `TestSticker_UpdatePartialAndListSortOrder` reads the row back via GET after the update and asserts persistence (it would fail if the commit were absent). No action needed. Two other speculative P2s from the same perspective (integer-division "0MB" size message; magic-header seek) were also checked and dismissed: `StickerMaxFileSize` is a compile-time `const = 1MB` so the message renders "1MB", the line is pre-existing and out of this diff's scope, and the magic-mismatch path returns immediately with no subsequent payload read.

## 3. Overall verdict

**APPROVED** — spec ✅ and quality Approved. No P0/P1 blockers. The three items above are quality/observability polish that can land in a follow-up.

## 4. Suggestions

- Prefer `RowsAffected()==0 → not-found` in `updateMetadataTx`, or serialize delete under the same user-row lock, to close the self-race in the P2 above.
- Log the effective `required` policy (not signing capability) in the upload/register success logs.
- Consolidate the duplicate `register_total` metric onto a single series.

## 5. Additional observations

- `validateStickerPath` deliberately accepts a right-tailed path on an unconfigured origin (documented, maintainer-approved in #508); the residual is self-scoped and grants no capability the sender lacks. Consistent with prior review, no change requested.
- Coverage gaps a human may wish to spot-check manually: the `sticker` DB tests did not execute here (harness/MySQL env, not the diff); runtime values of the new `DM_API_STICKER_UPLOAD_IP_RATELIMIT_*` env knobs and the shared Redis rate-limit middleware internals were reviewed by signature/usage only, not executed end-to-end.

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD bba6d71

**Formal Verdict**: `APPROVE`



## Reviewer Summary
|---|---|---|
| qa-engineer | `pass` | All 6 acceptance criteria verified with tests; CI fully green; strong observability. 0 Must/Should. |
| security-engineer | `cleared` | Authenticated own-UID metadata endpoint, parameterized SQL, bounded inputs, low-cardinality metrics. 0 findings, all tiers. |
| code-reviewer | `approved` | Clean additive increment reusing the proven user-row-lock transaction; correct partial-update + ownership scoping. 0 Must/Should. |

## Critical Findings (P0-P1)
无 Critical / High priority findings. No reviewer raised any Must Fix, functional fail, or security-blocking item.

## Should-Fix (P2)
无 P2 findings.

## Suggestions (P3)
- [P3-1] `modules/file/api.go:161-165` — the early `checkReq` failure branch for sticker uploads does not emit an upload counter, leaving a narrow class of early rejections uncounted in `sticker_upload_total` (flagged by qa/N-1 + code/N-1). Fix: emit a `validation_failed`/`path_rejected` counter (add the bucket to the label set) in that branch.
- [P3-2] `modules/sticker/api.go:180-322` — register events increment both the legacy `metrics.ObserveStickerRegister(...)` and the new bounded-label `observeStickerRegister(...)` families, i.e. dual bookkeeping for the same event (flagged by qa/N-2 + code/N-2). Fix: converge to the single low-cardinality `result` family in a later increment.
- [P3-3] `modules/sticker/api.go:311` — the register success log recomputes the keyword count via `decodeStickerKeywords(...)` even though `normalizeStickerKeywords` already returned the decoded slice (discarded with `_`) (flagged by code/N-3). Fix: reuse the returned slice to avoid a redundant JSON decode.

## Questions for Author
无 open questions.

## Verdict Rationale
All three reviewers independently reached clean verdicts on the same HEAD `bba6d71`: QA `pass` (6/6 ACs test-verified, 16 CI jobs + code-review context green), Security `cleared` (no credential/injection/auth-boundary findings across all disclosure tiers), and Code `approved` (no Must/Should defects). Consolidated findings reduce to three low-severity suggestions — two of which QA and code-review flagged independently (cross-reviewer agreement on the `checkReq` upload-metric gap and the dual register-metric families), and one code-only redundant-decode nit. None is a merge blocker. There are no P0/P1/P2 items and no material open questions, and no reviewer disagreement to reconcile (all three converge). Per the Formal Verdict rule (no P0/P1, ≤2 P2, no blocking question → APPROVE), the aggregate verdict is `APPROVE`.

## Round Context

---

PR #111, opened 2026-05-21
fix(ci): use issues: write permission for PR welcome comments

lml2468:
## Problem

The `pr-contributor-welcome.yml` caller that landed in a previous PR declared `permissions: pull-requests: write`, but PR comments are created via the GitHub Issues API (`issues.createComment`) which requires `issues: write`.

Caught by Jerry-Xin in Mininglamp-OSS/octo-adapters#53. Reusable workflow fix: Mininglamp-OSS/.github#33.

## Fix

`pull-requests: write` → `issues: write` in the caller job permissions.

PR #513, opened 2026-07-01
feat(bot_api): bot message reaction endpoints (#111)

caster-Q:
## Summary

Add bot-facing message reaction endpoints (backend of the three-end #111 reactions feature).

## Related Issue

#111

## Linked Spec

Message-adapter + reactions plan (issue #111).

## Changes

- `POST /v1/bot/messages/:message_id/reactions` — body `{channel_id, channel_type, emoji}`; a bot adds/replaces its single reaction on a message.
- `DELETE /v1/bot/messages/:message_id/reactions/:emoji?channel_id&channel_type` — removes it.
- Reactions reuse the existing `reaction_users` table and reaction seq stream, and broadcast `CMDSyncMessageReaction`, identical to the user-facing reaction path. Permission reuses `checkSendPermission` (group/thread/App-Bot/DM-friend). DM rows store under the fake channel id; the CMD broadcasts with the original channel id.
- Add/remove is a pure state machine over the single `(uid, message_id)` row with full branch coverage; a `reactionStateStore` seam keeps handler tests DB-free.
- Cross-channel write guard: before writing, verify the `message_id` actually belongs to the requested conversation via `message.Service.MessageExistsInChannel` (group = raw channel id, DM = fake channel id, matching how the message table stores DM rows). A mismatch is rejected as an invalid `message_id` with no store write and no broadcast — a bot cannot attach a reaction to a message in a channel it is not authorized for.

## Testing

- [x] Manually verified

`go test ./modules/bot_api -run '^TestReaction_'` passes (state-machine four branches, DM broadcast uses original channel id, no-op skips CMD, cross-channel message rejected). `go build ./modules/bot_api` clean.

## COMPREHENSION

1. **What it does to the load-bearing path**: adds a new bot-auth write path into `reaction_users` + reaction seq + `CMDSyncMessageReaction` broadcast, sharing the exact storage/broadcast contract as the user reaction path so clients treat bot and user reactions identically.

2. **What could break**: the message-load path joins reactions by `message_id` alone (no channel filter), so an unguarded bot write could surface a reaction on a message in another channel. Mitigated by the `MessageExistsInChannel` guard. Concurrency of the query-then-write `(uid, message_id)` row matches the pre-existing user path (documented; out of scope to harden here).

3. **How I know it works**: unit tests cover the state machine, DM channel-id resolution, no-op, and the cross-channel reject; three-end e2e verified (API 200 + row persisted + broadcast + web chip rendered).

## Checklist

- [ ] Updated documentation

> Draft: reactions is a three-end feature — ships with octo-web (render) + openclaw-channel-octo (react action).

### DIFF
--- a/modules/bot_api/bot_api.go
+++ b/modules/bot_api/bot_api.go
@@ -29,13 +29,13 @@ const (
 // BotAPI is the public Bot API gateway module.
 // It handles all bot-facing endpoints (/v1/bot/*) with unified auth.
 type BotAPI struct {
-	ctx                   *config.Context
-	db                    *botAPIDB
-	userService           user.IService
-	fileService           file.IService
-	groupService          group.IService
-	userDB                *user.DB
-	threadService         thread.IService
+	ctx           *config.Context
+	db            *botAPIDB
+	userService   user.IService
+	fileService   file.IService
+	groupService  group.IService
+	userDB        *user.DB
+	threadService thread.IService
 	// robotService gives the OBO fan-out path a way to enqueue synthetic
 	// events directly into a grantee bot's /v1/bot/events queue. The
 	// webhook layer drops NoPersist=1 messages before NotifyMessagesListeners
@@ -118,6 +118,13 @@ type BotAPI struct {
 	// path (nil override) goes through ba.ctx.SendCMD verbatim, so
 	// behaviour outside of tests is unchanged.
 	typingCMDDispatch func(req config.MsgCMDReq) error
+	// reactionCMDDispatch — #111 Sprint B. Test seam for the reaction
+	// handler's ctx.SendCMD call, mirroring typingCMDDispatch. Production
+	// path (nil) goes through ba.ctx.SendCMD verbatim.
+	reactionCMDDispatch func(req config.MsgCMDReq) error
+	// reactionStoreOverride — #111 Sprint B. Test seam letting handler-level
+	// tests drive applyReaction with a fake store (no live DB). Nil in prod.
+	reactionStoreOverride reactionStorer
 	// friendCheckOverride lets unit tests stub userService.IsFriend for the
 	// friend-gate decision in checkSendPermission / syncMessages, and for
 	// the OBO friend-gate bypass (see obo_friend_gate.go). Production path
@@ -213,6 +220,9 @@ func (ba *BotAPI) Route(r *wkhttp.WKHttp) {
 		botAPI.POST("/events/:event_id/ack", ba.eventAck)
 		botAPI.POST("/heartbeat", ba.heartbeat)
 		botAPI.POST("/messages/sync", ba.syncMessages)
+		// #111 Sprint B — bot message reactions
+		botAPI.POST("/messages/:message_id/reactions", ba.addReaction)
+		botAPI.DELETE("/messages/:message_id/reactions/:emoji", ba.removeReaction)
 		botAPI.GET("/groups", ba.getGroups)
 		botAPI.GET("/resolve/targets", ba.botResolveTargets)
 		botAPI.GET("/groups/:group_no", ba.getGroupInfo)
@@ -291,6 +301,9 @@ func (ba *BotAPI) resolveSpaceChannelID(robotID, channelID string, channelType u
 
 // resolveBotDisplayName queries the bot's display name, falls back to robotID.
 func (ba *BotAPI) resolveBotDisplayName(robotID string) string {
+	if ba.userDB == nil {
+		return robotID
+	}
 	botUser, err := ba.userDB.QueryByUID(robotID)
 	if err == nil && botUser != nil && botUser.Name != "" {
 		return botUser.Name
--- a/modules/bot_api/bot_api.go
+++ b/modules/bot_api/bot_api.go
@@ -11,6 +11,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
 	"github.com/Mininglamp-OSS/octo-server/modules/file"
 	"github.com/Mininglamp-OSS/octo-server/modules/group"
+	"github.com/Mininglamp-OSS/octo-server/modules/message"
 	"github.com/Mininglamp-OSS/octo-server/modules/robot"
 	"github.com/Mininglamp-OSS/octo-server/modules/thread"
 	"github.com/Mininglamp-OSS/octo-server/modules/user"
@@ -36,6 +37,11 @@ type BotAPI struct {
 	groupService  group.IService
 	userDB        *user.DB
 	threadService thread.IService
+	// messageService resolves whether a message_id actually belongs to a
+	// channel, so the reaction handler can reject cross-channel message ids
+	// (see messageChannelChecker / applyReaction). Narrow interface, satisfied
+	// by *message.Service, to avoid widening message.IService for external fakes.
+	messageService messageChannelChecker
 	// robotService gives the OBO fan-out path a way to enqueue synthetic
 	// events directly into a grantee bot's /v1/bot/events queue. The
 	// webhook layer drops NoPersist=1 messages before NotifyMessagesListeners
@@ -125,6 +131,10 @@ type BotAPI struct {
 	// reactionStoreOverride — #111 Sprint B. Test seam letting handler-level
 	// tests drive applyReaction with a fake store (no live DB). Nil in prod.
 	reactionStoreOverride reactionStorer
+	// messageExistsOverride — #111 Sprint B. Test seam for the cross-channel
+	// reaction guard: handler-level tests inject message existence without a
+	// live message store. Nil in prod (goes through ba.messageService).
+	messageExistsOverride func(channelID string, channelType uint8, messageID string) (bool, error)
 	// friendCheckOverride lets unit tests stub userService.IsFriend for the
 	// friend-gate decision in checkSendPermission / syncMessages, and for
 	// the OBO friend-gate bypass (see obo_friend_gate.go). Production path
@@ -188,6 +198,7 @@ func NewBotAPI(ctx *config.Context) *BotAPI {
 		groupService:          group.NewService(ctx),
 		userDB:                user.NewDB(ctx),
 		threadService:         thread.NewService(ctx),
+		messageService:        message.NewService(ctx),
 		robotService:          robot.NewService(ctx),
 		speechClient:          voice_adapter.NewSpeechClient(speechURL, speechKey, time.Duration(timeoutSec)*time.Second),
 		maxVoiceContextLength: maxCtxLen,
--- a/modules/bot_api/reaction_store.go
+++ b/modules/bot_api/reaction_store.go
@@ -1,0 +1,93 @@
+package bot_api
+
+import (
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/db"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
+	"github.com/gocraft/dbr/v2"
+)
+
+// reactionStore is the bot_api-local data access for message reactions.
+//
+// #111: reactions for bot actors reuse the SAME `reaction_users` table the
+// user-facing reaction API writes to (modules/message). We deliberately do NOT
+// import message's private messageReactionDB (keeping module DB access private
+// is the octo-server convention); instead we own a small, single-file SQL
+// surface here. The bot is recorded as the reaction actor via uid=robotID, with
+// the bot's display name in the name column so user clients render it normally.
+type reactionStore struct {
+	session *dbr.Session
+	ctx     *config.Context
+}
+
+func newReactionStore(ctx *config.Context) *reactionStore {
+	return &reactionStore{ctx: ctx, session: ctx.DB()}
+}
+
+// reactionRow mirrors the columns of reaction_users that the bot path touches.
+// (Same physical table as message.reactionModel; redeclared locally so we don't
+// reach into another module's private type.)
+type reactionRow struct {
+	MessageID   string
+	Seq         int64
+	ChannelID   string
+	ChannelType uint8
+	UID         string
+	Name        string
+	Emoji       string
+	IsDeleted   int
+	db.BaseModel
+}
+
+// queryByActorAndMessage returns the bot's existing reaction row for a message,
+// or nil when none exists. (uid, message_id) is effectively unique — the user
+// path enforces "one reaction row per actor per message", which we preserve.
+//
+// Concurrency: this is query-then-write with no row lock and no DB unique
+// constraint on (uid, message_id) — the same pattern the user-facing path
+// (modules/message addOrCancelReaction) has always used. Two concurrent adds
+// for the same (bot, message) could in principle insert two rows. For a
+// normally-serial bot actor this is low-probability and no worse than the
+// existing user-device race; hardening both paths would mean adding a
+// UNIQUE(uid, message_id) index in a later migration (it must also dedupe any
+// pre-existing duplicate rows), which is out of scope for this change.
+func (s *reactionStore) queryByActorAndMessage(uid, messageID string) (*reactionRow, error) {
+	var m *reactionRow
+	_, err := s.session.Select("*").
+		From("reaction_users").
+		Where("uid=? and message_id=?", uid, messageID).
+		Load(&m)
+	if err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+func (s *reactionStore) insert(m *reactionRow) error {
+	_, err := s.session.InsertInto("reaction_users").
+		Columns(util.AttrToUnderscore(m)...).
+		Record(m).
+		Exec()
+	return err
+}
+
+func (s *reactionStore) update(m *reactionRow) error {
+	_, err := s.session.Update("reaction_users").
+		SetMap(map[string]interface{}{
+			"seq":        m.Seq,
+			"emoji":      m.Emoji,
+			"is_deleted": m.IsDeleted,
+			"name":       m.Name,
+		}).
+		Where("id=?", m.Id).
+		Exec()
+	return err
+}
+
+// genSeq mints the next reaction seq for a stored channel id, matching the
+// user-facing path (modules/message genMessageReactionSeq) so bot and user
+// reactions share one monotonic stream per channel.
+func (s *reactionStore) genSeq(storeChannelID string) (int64, error) {
+	return s.ctx.GenSeq(common.MessageReactionSeqKey + ":" + storeChannelID)
+}
--- a/modules/bot_api/reactions.go
+++ b/modules/bot_api/reactions.go
@@ -82,6 +82,23 @@ func (ba *BotAPI) removeReaction(c *wkhttp.Context) {
 	ba.applyReaction(c, messageID, channelID, channelType, emoji, true)
 }
 
+// messageChannelChecker is the narrow slice of message.Service the reaction
+// handler depends on: confirm a message_id belongs to a (channel_id, type).
+// Kept local (not message.IService) so adding it never forces external
+// IService fakes to grow a method.
+type messageChannelChecker interface {
+	MessageExistsInChannel(channelID string, channelType uint8, messageID string) (bool, error)
+}
+
+// messageExistsInChannel routes the cross-channel guard through the test seam
+// when set, else the live message service. Mirrors the other reaction seams.
+func (ba *BotAPI) messageExistsInChannel(channelID string, channelType uint8, messageID string) (bool, error) {
+	if ba.messageExistsOverride != nil {
+		return ba.messageExistsOverride(channelID, channelType, messageID)
+	}
+	return ba.messageService.MessageExistsInChannel(channelID, channelType, messageID)
+}
+
 // applyReaction is the shared add/remove state machine. Permission and the
 // channel-id store mapping mirror the user-facing reaction path so bot and
 // user reactions are indistinguishable to clients.
@@ -102,6 +119,28 @@ func (ba *BotAPI) applyReaction(c *wkhttp.Context, messageID, channelID string, 
 	// so reaction_users rows land in the same per-conversation bucket and the
 	// reaction seq stream is shared with the user path.
 	storeChannelID := resolveReactionStoreChannelID(channelID, channelType, robotID)
+
+	// Cross-channel guard: checkSendPermission only proved the bot may post to
+	// the REQUEST channel, but reaction_users rows are keyed by message_id and
+	// the message-load path joins reactions by message_id alone (no channel
+	// filter — see modules/message db_message_reaction.queryWithMessageIDs). So
+	// a bot could otherwise attach a reaction to a message in a channel it has
+	// no access to, just by knowing that message's id. Require the message to
+	// actually live in this conversation (storeChannelID matches the message
+	// table's per-conversation key, incl. the DM fake channel id) before any
+	// write or broadcast.
+	exists, err := ba.messageExistsInChannel(storeChannelID, channelType, messageID)
+	if err != nil {
+		ba.Error("校验消息归属失败", zap.Error(err))
+		httperr.ResponseErrorL(c, errcode.ErrMessageQueryFailed, nil, nil)
+		return
+	}
+	if !exists {
+		// Not a message of this conversation — reject as invalid input without
+		// revealing whether it exists elsewhere.
+		respondBotAPIRequestInvalid(c, "message_id")
+		return
+	}
 
 	store := ba.newReactionStorer()
 	existing, err := store.queryByActorAndMessage(robotID, messageID)
--- a/modules/bot_api/reactions.go
+++ b/modules/bot_api/reactions.go
@@ -1,0 +1,276 @@
+package bot_api
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
+	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
+	"go.uber.org/zap"
+)
+
+// #111 Sprint B — bot reactions.
+//
+// POST   /v1/bot/messages/:message_id/reactions          body {channel_id, channel_type, emoji}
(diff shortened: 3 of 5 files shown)


OctoBoooot (commented):
## Review: feat(bot_api): bot message reaction endpoints (#513) — COMMENT (draft: pre-ready feedback)

Heads-up: this PR is a **draft** (`isDraft: true`), so I'm giving pre-ready feedback rather than a merge-gating verdict — flip it to Ready when you want the formal pass. That said, the core is well-built; no blocker, a few things to tighten.

### Done well (verified)
- ✓ **Auth + disband reuse the vetted contract.** `applyReaction` gates via `checkSendPermission`, which itself contains the disband guard (send.go:324/367 — "once the group is disbanded everyone is read-only, bots included") for both Group and CommunityTopic-parent. So a bot reaction on a disbanded group is correctly rejected — no separate `isGroupDisbanded` call needed here, and no bypass of the #463/#510 read-only contract. Good reuse.
- ✓ **Cross-channel message-ownership guard** (the subtle one, nicely caught): since `reaction_users` joins by `message_id` alone (no channel filter), `messageExistsInChannel(storeChannelID, ...)` is required before any write so a bot can't attach a reaction to a message in a channel it can't access just by knowing the id. The comment explains the exact risk. Solid.
- ✓ DM vs group store-channel resolution (fake-channel-id for DMs) matches the user path; broadcast CMD uses the original channel id; change-detection avoids spurious broadcasts. Build green.

### Pre-ready findings (minor)
- **Emoji has no length bound** (reactions.go:43/66 — `TrimSpace` + non-empty only). It's stored in `reaction_users.emoji` (`varchar(20)`) and broadcast to the conversation. Calibration: this is **not a bot-path-specific hole** — the user path (`message/api.go addOrCancelReaction`) is identical (stores `req.Emoji` unvalidated) and the `varchar(20)` column backstops it (MySQL truncates/errors past 20). So minor/consistency, not major — but an app-side length (and ideally single-grapheme) check on *both* paths would avoid relying on DB truncation and reject a megabyte payload before it hits storage/broadcast. At minimum, cap it here to match the column.
- **`channel_type` only rejects 0**, not "is it a known type." A non-zero unknown type flows into `checkSendPermission` (per-`botKind`, per-type-explicit switch) + `resolveReactionStoreChannelID`. It likely fails to match any write branch (no `return nil` grant) → rejected — but please confirm an unknown non-zero `channel_type` can't reach a default-allow or a bad store-key, and reject it explicitly up front (allowlist Group/Person/CommunityTopic) for clarity.
- **Test coverage is store-layer only.** The 15 tests pin the reaction state-transitions (insert/revive/no-op/overwrite/soft-delete, DM vs group store id) well — but there's no handler-level test for the guards that matter most on a new write endpoint: disband rejection, the cross-channel message-ownership rejection, permission denial, empty/oversized emoji, invalid channel_type. Add a few driving `applyReaction` (the `reactionStoreOverride` seam already supports handler tests without a live DB) so these can't silently regress.

### Net
**COMMENT** (draft — not gating). The security-critical parts are right: disband + permission inherited from `checkSendPermission`, cross-channel ownership guarded, DM/group buckets correct. Pre-ready polish: bound the emoji (both paths, or at least cap to the varchar(20) column), allowlist channel_type explicitly, and add handler-level tests for the disband/ownership/validation rejections. When you mark it Ready I'll do the formal verdict — on current state it's close, with those tighten-ups.

Jerry-Xin (dismissed):
**APPROVE** — bot reaction endpoints are correctly authz-gated at parity with the user path; the cross-channel ingress gap this module has been prone to is explicitly closed. Build/tests green locally at `7bddf9a4`.

## 🔒 Security trace (byte-verified at head 7bddf9a4)

**Channel authorization (the high-risk one) — enforced, not a weaker sibling.**
`applyReaction` calls `checkSendPermission(c, botKind, robotID, channelID, channelType, false)` before any write. That gate (modules/bot_api/send.go) enforces real membership:
- Group: `SELECT COUNT(*) FROM group_member WHERE group_no=? AND uid=? AND is_deleted=0` → `errBotSendPermNotGroupMember` if 0.
- Thread: parent group_no membership (same COUNT), + disband guard.
- App Bot: DM-only + `IsFriend` + space-member (fail-closed).
- DM (User Bot): creator-of-bot OR friend.

**Cross-channel reaction injection (IDOR) — closed.**
`checkSendPermission` only proves the bot may post to the *request* channel; `reaction_users` is joined to messages by `message_id` alone. The PR adds `MessageExistsInChannel(storeChannelID, channelType, messageID)` → `queryMessageByID` which filters `channel_id=? AND channel_type=? AND message_id=? AND is_deleted=0`. Adversarial case (bot member of channel A passes A's channel_id + a message_id that lives in channel B) → message not found under (A, msgid) → `respondBotAPIRequestInvalid("message_id")`, **no write, no broadcast**. Covered by `TestReaction_Handler_MessageNotInChannel_Rejected`.
DM store id uses `GetFakeChannelIDWith(peer, bot)`, which is order-independent (sorts by CRC32 hash), so the guard's channel id matches how DM messages are physically stored and how the user path buckets — no false-reject.

**Remove scoped to caller's own reaction — yes.**
`queryByActorAndMessage(robotID, messageID)` = `Where("uid=? and message_id=?", robotID, messageID)`; remove additionally requires `existing.Emoji == in.emoji`. A bot can never soft-delete another actor's row.

**Data integrity.** All access via dbr builders / parameterized placeholders (`Where(... ?)`, `SetMap`+`Where("id=?")`, `InsertInto...Record`) — no SQLi. Actor recorded as `uid=robotID` with resolved display name.

**Fanout.** `CMDSyncMessageReaction` dispatched only when `changed`, uses the *original* channel id (routes to real conversation), fail-closed (returns 500 on dispatch error).

## ✅ Verification
- `go build ./modules/bot_api/... ./modules/message/...` → clean at 7bddf9a4.
- `go test ./modules/bot_api -run '^TestReaction_'` → ok (12 tests: 4 add branches, 4 remove branches, store-id mapping, App-Bot-group reject, cross-channel reject, non-friend DM reject, no-op-skips-CMD).
- `go vet ./modules/bot_api/...` → clean.
- Note: CI Build/Vet/Test show **SKIPPED** (path filters didn't trigger the Go jobs for this PR) — verified locally instead.

## 🟡 Non-gating notes
- **Emoji not length/charset-bounded** (handler only trims + non-empty check). This is at exact parity with the existing user path (`modules/message/api.go` stores `req.Emoji` verbatim; column is `varchar(20)`), so not a regression — but a shared `emoji` bound on both paths would be cleaner.
- **No per-bot rate limit / query-then-write race** on `(uid, message_id)` (no UNIQUE index). Author documents both; abuse is row-bounded (SET semantics, one row per bot/message) and the race mirrors the long-standing user-device behavior. A future `UNIQUE(uid, message_id)` migration (with dedup) would harden both paths — out of scope here.
- PR is marked **Draft** (three-surface feature; needs octo-web + openclaw-channel-octo to ship together). Backend here is sound; approving the backend on its merits.

Solid, defensive implementation — the module's recurring bot-ingress-bypass class was anticipated and explicitly guarded. Nice test coverage of the adversarial cross-channel case.

Jerry-Xin (commented):
Draft-aware review (this PR is marked **draft / WIP**, so this is pre-ready feedback as a COMMENT, not a merge-gating verdict — I'll give a formal verdict when it's marked Ready).

The core is well-built and byte-verified sound:

**✅ Auth + disband correct via reuse** — `applyReaction` gates through `checkSendPermission`, which itself contains the read-only-after-disband guard (Group: send.go:324-332; CommunityTopic parent: :367 — "once the group is disbanded everyone is read-only, bots included"). So a bot reaction on a disbanded group is correctly rejected — no bypass of the #463/#510 read-only contract, and no separate `isGroupDisbanded` call needed since the guard lives in the shared permission function.

**✅ Cross-channel message-ownership guarded** — a genuinely subtle catch: since `reaction_users` joins by `message_id` alone, `messageExistsInChannel` is required so a bot can't react to a message in a channel it can't access just by knowing the ID. Well-reasoned.

**✅ Parameterized store layer** — no SQLi; store-layer state-transition tests present (15).

**Pre-ready findings (minor / polish — none blocking):**
- 🟡 Emoji has no app-side length/charset bound (`strings.TrimSpace`, only empty-rejected) before it's stored and broadcast in the reaction CMD. Note: the **user path (`addOrCancelReaction`) is identical** and both are backstopped by `reaction_users.emoji varchar(20)`, so this is a consistency/hardening note, not a bot-specific regression — but capping length app-side on **both** paths (and validating it's a single grapheme/emoji) would avoid relying on a DB truncation/strict-mode error.
- 🟡 `channel_type` only rejects `0` — it doesn't allowlist known types (Group/Person/CommunityTopic). An arbitrary non-zero type flows into `checkSendPermission`, whose per-type explicit switch won't grant on an unknown type (so it fails closed downstream), but an explicit allowlist at the handler is cleaner + self-documenting. Worth confirming unknown types can't reach a default-open branch.
- 🟡 No bot-specific rate limit on reaction writes; and `updateMetadata`-style TOCTOU aside, the store add/query is not explicitly serialized — low-risk for reactions (idempotent-ish, self-scoped) but note dedup behavior on repeated identical reactions.
- 🟡 Tests are store-layer only — no handler-level coverage for the auth/disband rejection, the cross-channel ownership guard, emoji bounds, or invalid `channel_type`. The `reactionStoreOverride` seam supports adding these; recommend handler-level rejection tests before Ready.

Solid work — the security-critical paths (auth, disband, cross-channel) are handled correctly by reuse + a good ownership guard. The above are pre-ready polish; happy to give a formal verdict once it's marked Ready for review.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#513 Review Report
**Head SHA:** `7bddf9a4e68da949638eaa923ee6e91762e2c96b`
**Base:** `main`
**Scope:** `feat(bot_api): bot message reaction endpoints (#111)`

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Auth integration | ✅ | Routes registered inside `authBot()` group (`bot_api.go:233-234`); `robotID`, `botKind`, `robot` populated by middleware before handler runs |
| Permission model | ✅ | `checkSendPermission` reused with `hasOBOContext=false` (`reactions.go:117`); handles Group/Thread/DM/AppBot correctly |
| Cross-channel guard | ✅ | `MessageExistsInChannel` queries sharded message table by `(channel_id, channel_type, message_id, is_deleted=0)` (`db.go:36-44`); called before any write/broadcast (`reactions.go:131-144`) |
| DM fake channel ID | ✅ | `GetFakeChannelIDWith` is order-independent (CRC32 sort, `octo-lib/common/msg.go:157-168`); consistent between message storage and reaction storage |
| Seq stream sharing | ✅ | Both bot and user paths use `common.MessageReactionSeqKey + ":" + storeChannelID` (`reaction_store.go:91-92` vs `api.go:2267-2268`) |
| State machine correctness | ✅ | All 7 branches (4 add + 3 remove) covered; pure function with comprehensive tests; 15/15 tests PASS |
| CMD broadcast routing | ✅ | Uses original `channelID` (not fake store ID) for CMD (`reactions.go:174`); only broadcasts when `changed==true` |
| Build | ✅ | `go build ./...` clean |
| Tests | ✅ | `go test ./modules/bot_api -run '^TestReaction_'` — 15/15 PASS |
| Vet | ✅ | `go vet ./modules/bot_api/ ./modules/message/` clean |

## 2. Findings

### No P0 or P1 issues found.

### P2 Findings

**P2-1: Pre-existing concurrency race on `reaction_users` — no UNIQUE constraint on `(uid, message_id)`**
- **Diff-scope:** Pre-existing. The user-facing path (`modules/message addOrCancelReaction`) has the same query-then-write pattern with no row lock.
- **Impact:** Two concurrent reactions for the same (bot, message) could in principle insert duplicate rows. Low probability for normally-serial bot actors.
- **PR acknowledgement:** Explicit in `reaction_store.go:49-56` comment; defers to future UNIQUE index migration with dedup.
- **Recommendation:** Track as follow-up. No action required for this PR.

**P2-2: No emoji length/content validation on bot path (or user path)**
- **Diff-scope:** Pre-existing. The user-facing `addOrCancelReaction` (`api.go:1707+`) also accepts arbitrary emoji strings without length bounds.
- **Impact:** A client could send an arbitrarily long string as "emoji". Stored in `reaction_users.emoji` column and propagated to client sync responses.
- **Recommendation:** Consider adding a max-length check (e.g. 32 bytes) on both paths in a follow-up. Not a blocker.

**P2-3: `db *DB` field added to `message.Service` struct**
- **Diff-scope:** New in this PR (`service.go:42`, `service.go:52`). The `db` field was NOT previously on the `Service` struct (only on the `Message` struct in `api.go`).
- **Impact:** Minor surface expansion. `NewService` creates a new `dbr.Session` via `NewDB(ctx)` — separate session object from `Message.db` but shares the underlying connection pool. No resource concern.
- **Recommendation:** Acceptable. The narrow `messageChannelChecker` interface keeps the external contract minimal.

### Nits

**Nit-1: `resolveBotDisplayName` nil-guard (`bot_api.go:314-316`)**
Test-safety guard — `ba.userDB` is nil in handler-level tests that construct `BotAPI` without a full context. Not fixing a production bug, but harmless.

**Nit-2: State machine mutates `existing` pointer in-place**
`reactionStateTransitionImpl` modifies `existing.Seq`, `existing.Emoji`, `existing.IsDeleted`, `existing.Name` directly on the input pointer before calling `store.update(existing)`. Safe because `existing` comes from `queryByActorAndMessage` and is not shared, but callers should be aware the input is not immutable.

## 3. Suggestions

1. **Future: emoji validation** — Add length/format bounds on the emoji field for both bot and user paths (defense-in-depth).
2. **Future: UNIQUE index** — Add a `UNIQUE(uid, message_id)` index on `reaction_users` with a dedup migration to close the concurrency race.
3. **Future: user-path cross-channel guard** — The user-facing `addOrCancelReaction` does NOT verify message existence in the channel (unlike the new bot path). Consider backporting `MessageExistsInChannel` to the user path for symmetry.

## 4. Extra Observations

- The bot path uses **SET semantics** (re-adding same emoji = no-op) while the user path uses **TOGGLE semantics** (re-adding same emoji = remove). The PR documents this divergence explicitly (`reactions.go:236-240`). This is intentional for idempotent bot retries. No collision risk per-actor since `(uid, message_id)` is effectively unique, and bot UIDs never collide with human UIDs.
- The cross-channel guard (`MessageExistsInChannel`) is **stronger than the user-facing path**, which has no equivalent check. The bot path correctly identified and mitigated a gap that exists in the user path.
- The `messageChannelChecker` narrow interface pattern (`reactions.go:86-88`) is good design — avoids forcing all `IService` fakes to implement the new method.

## 5. Data-Flow Trace

| Consumed Data | Upstream Source | Flow Verified? |
|---|---|---|
| `robotID` (handler) | `authBot()` middleware → `CtxKeyRobotID` | ✅ Auth middleware populates from token |
| `botKind` (handler) | `authBot()` middleware → `CtxKeyBotKind` | ✅ "user" or "app" from token prefix |
| `robot.CreatorUID` (permission) | `authBot()` → `CtxKeyRobot` → `*robotModel` | ✅ User Bot only; App Bot gets nil |
| `storeChannelID` (DM) | `GetFakeChannelIDWith(channelID, robotID)` | ✅ Order-independent CRC32 sort; matches message table key |
| `storeChannelID` (Group/Thread) | Raw `channelID` | ✅ Same as user path |
| `MessageExistsInChannel` result | `queryMessageByID(channelID, channelType, messageID)` → sharded `message[N]` table | ✅ Filters by channel_id + channel_type + message_id + is_deleted=0 |
| `existing` reaction row | `queryByActorAndMessage(robotID, messageID)` → `reaction_users` table | ✅ dbr `Load` populates struct incl. BaseModel.Id |
| `seq` | `GenSeq("messageReaction:" + storeChannelID)` | ✅ Shared monotonic stream with user path |
| `name` (display name) | `resolveBotDisplayName(robotID)` → `userDB.QueryByUID` or fallback to robotID | ✅ Nil-guard added for test safety |
| CMD `ChannelID` | Original request `channelID` (not fake store ID) | ✅ Routes CMD to real conversation |
| CMD broadcast | Only when `changed==true` | ✅ No-op paths skip broadcast correctly |

- **C1 — Double-path parity (add↔remove):** ✅ Both POST and DELETE route through `applyReaction` with `remove=false`/`true`. State machine covers all symmetric branches. Route registration is symmetric.
- **C1 — Guard/lint coverage:** N/A — no guard/lint tool claims in this PR.
- **C2 — Control-flow ordering / nested reuse:** ✅ Linear flow: auth → permission → cross-channel guard → query → state transition → broadcast. No nested double-invocation. `resolveBotDisplayName` called lazily only on actual add.
- **C3 — Authorization boundary ≠ capability:** ✅ Routes inside `authBot()` group. `checkSendPermission` validates membership/friendship/disband. Cross-channel guard adds message-level verification.
- **C4 — Authorization lifecycle / container cascade:** N/A — no container/member status cascade involved. Bot auth token status is checked by auth middleware.
- **C5 — Build ≠ runtime:** N/A — Go server, no build artifacts/extensions/CLI packaging involved.
- **C6 — Governance docs:** N/A — feature PR, not governance/policy.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 issues. The code is well-structured with proper auth integration, a thoughtful cross-channel injection guard, consistent DM channel ID handling, shared seq namespace for client sync, and comprehensive test coverage. P2 findings are pre-existing patterns (concurrency race, no emoji validation) properly acknowledged in code comments. Safe to merge as draft pending联调 with web + adapter layers.

yujiawei (approved):
# Code Review — PR #513 (octo-server)

**Head SHA:** `7bddf9a4e68da949638eaa923ee6e91762e2c96b`
**Scope:** `feat(bot_api): bot message reaction endpoints` — 5 files, +816/-7.

Reviewed against the existing user-facing reaction path (`modules/message/addOrCancelReaction`), the shared permission contract (`checkSendPermission`), the DM fake-channel-id helper, the `reaction_users` schema, and the webhook message-persistence path. Build, `go vet`, and the PR's own `TestReaction_*` suite all pass. (The one failing package test, `TestBotGroupMemberRemove_ManagerTargetForbidden`, is a pre-existing local test-DB migration issue — `unknown migration in database` from `NewTestServer` — unrelated to this change.)

## 1. Spec compliance

**Spec: ✅**

- **Nothing missing.** Both endpoints described in the PR are implemented (`POST /v1/bot/messages/:message_id/reactions`, `DELETE .../:emoji`), registered under the authenticated `/v1/bot` group, with add/remove state machine, DM fake-channel-id storage, shared reaction seq stream, and `CMDSyncMessageReaction` broadcast on the original channel id.
- **Nothing extra.** No unrelated flags, fields, or endpoints. The one small addition outside the endpoints — a `nil`-guard in `resolveBotDisplayName` — is a defensive test-support change with no production behavior impact (`userDB` is always set in prod).
- **No deviation.** The permission model reuses `checkSendPermission` exactly as stated, so group / thread / App-Bot-DM / DM-friend / group-disband rules are inherited rather than reimplemented. The deliberate SET (vs. user-path TOGGLE) semantics are documented and justified for idempotent agent retries.

**Verified correctness highlights:**

- **Cross-channel guard is sound.** `applyReaction` calls `MessageExistsInChannel(storeChannelID, channelType, messageID)` *before* any write or broadcast. `storeChannelID` uses the raw channel id for groups/threads and the fake channel id for DMs — and the webhook persists DM messages under the same fake channel id (`modules/webhook/api.go:260-263`, `GetFakeChannelIDWith(FromUID, ChannelID)`). Since `GetFakeChannelIDWith` hash-orders its two arguments, the (peer, bot) ordering used by the guard matches the stored key regardless of argument order. The guard closes the real risk that `reaction_users` rows are joined by `message_id` alone (`queryWithMessageIDs`, no channel filter).
- **Disband coverage inherited.** Reactions route through `checkSendPermission`, which self-checks `group.status` for Group and CommunityTopic (parent) before returning — so the "read-only after disband" invariant holds for the reaction write path too, without new code.
- **Broadcast routing correct.** The sync CMD uses the *original* channel id (not the fake store id), matching the user path so clients route it to the real conversation.
- **No-op discipline.** State-machine no-ops (re-add same emoji, remove non-matching/absent) correctly skip both the DB write and the broadcast.
- **Stricter than the user path, appropriately.** The user endpoint performs no message-existence check; the bot endpoint adds one, which is necessary because a bot supplies an arbitrary `message_id` + channel params rather than acting from an authenticated session bound to a conversation.

## 2. Code quality

**Quality: Approved**

No P0/P1. The findings below are P2 hardening suggestions, not merge blockers.

### P2 — `checkSendPermission` falls through to allow on unsupported channel types (`modules/bot_api/reactions.go:48`, `:78`)

For `BotKindUser`, `checkSendPermission` (`modules/bot_api/send.go:322-434`) only branches on Person / Group / CommunityTopic and then `return nil`. Channel types such as `CustomerService(3)`, `Community(4)`, `Info(6)` are not rejected, so a user bot that knows a valid `(channel_id, message_id)` in such a channel could attach a reaction there with no membership/friend/disband check — the cross-channel guard only proves the message *exists* under that channel, not that the bot is *authorized* for it.

Why this is P2, not a blocker:
- It is **parity with the existing user-facing reaction path**, which has the identical Group/Person-only gate and no `else` check (`modules/message/api.go:1722-1746`).
- It is a **shared property of `checkSendPermission`**, exercised the same way by `sendMessage` / `typing` / `readReceipt` — not introduced by this PR.
- The write is a cosmetic emoji reaction on an already-existing message; blast radius is low.

Recommendation: harden `checkSendPermission` (or add an allow-list check in the reaction handler) to fail closed on channel types outside the supported set. Because it is a platform-wide property, the right fix is one central hardening pass, not a reaction-only patch.

### P2 — DELETE silently truncates `channel_type` before validation (`modules/bot_api/reactions.go:71-73`)

```go
channelTypeRaw, _ := strconv.Atoi(strings.TrimSpace(c.Query("channel_type")))
channelType := uint8(channelTypeRaw)
```

`strconv.Atoi` errors are dropped and the int is wrapped to `uint8`, so `channel_type=258` or `-254` both become `2` (Group). Authorization is keyed on `channelType`, so a caller can present a wire value that differs from the enforced one. Combined with the message-existence guard the practical impact is limited, but strict parsing (reject parse error and out-of-range values, ideally reject anything outside the supported set) is cheap and removes the ambiguity. The POST path binds `channel_type` as a typed `uint8` and avoids this.

### P2 (acknowledged in-code) — query-then-write race on `(uid, message_id)` (`modules/bot_api/reaction_store.go:55-73`)

There is no row lock, transaction, or unique constraint — `reaction_users` has only a non-unique index (`modules/message/sql/20210416000001_message_legacy01.sql:17`). Two concurrent adds for the same `(bot, message)` can insert duplicate live rows; `update`/`delete` then touch only one (`WHERE id=?`), leaving a stray reaction visible via `queryWithMessageIDs`. The code comment already documents this as accepted, matching the long-standing user-path behavior, and correctly defers a `UNIQUE(uid, message_id)` migration (which must also dedupe pre-existing rows) as out of scope. Fine to accept for a normally-serial bot actor; noting it so it is not forgotten if reaction volume grows.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved (no P0/P1). Clean implementation with a well-reasoned cross-channel guard, inherited disband/permission coverage, correct DM fake-channel keying, and thorough DB-free unit tests via the store/CMD/existence seams.

## 4. Suggestions

- Prefer strict `channel_type` parsing on the DELETE path (reject parse errors and out-of-range values) for symmetry with the typed POST body.
- Track the unsupported-channel-type fall-through as a central `checkSendPermission` hardening item (affects all bot write endpoints, not just reactions).

## 5. Additional notes

- This is marked Draft pending integration with octo-web (render) and openclaw-channel-octo (react action); the three-surface rollout dependency is a coordination item, not a code issue.
- The bot SET vs. user TOGGLE divergence writes to the same `reaction_users` rows — the in-code note that consumers "must not assume toggle-for-all" is correct and worth keeping visible for the web/channel implementers.

OctoBoooot (commented):
## Re: @yujiawei's channel-type authz finding on #513 — confirmed real (upgrading my earlier note); still draft/COMMENT

yujiawei is right, and it's more concrete than the "allowlist channel_type" nit I first posted — byte-verified. Upgrading it to a real should-fix. Still a draft, so still pre-ready COMMENT, not gating.

### Confirmed: `checkSendPermission` fails OPEN on an unknown channel_type (BotKindUser)
`checkSendPermission` (send.go) for `case BotKindUser` is `if channelType==Group {…} else if ==CommunityTopic {…} else if ==Person {…}` then a trailing **`return nil`** (send.go:~415). So a `BotKindUser` bot sending an **unknown/unhandled non-zero channel_type** falls through all three branches → **permission GRANTED**, with no membership or disband check. The `default:` case only catches unknown *botKind*, not unknown *channelType*. (BotKindApp is safe — it rejects any non-Person type with `AppBotDMOnly`.)

For the reaction endpoint: `addReaction` only rejects `channel_type == 0`, so a bogus non-zero type reaches `checkSendPermission` → granted. **Practically backstopped** here by `messageExistsInChannel(channelID, channelType, messageID)` (it filters by channel_type, so a bogus type finds no matching message → rejected) — so reactions don't actually let a write through. But the **authz layer itself grants on unknown types**, which is the fail-open yujiawei flagged; the message-exists check is the only thing catching it, and the *send* path (no such backstop) is where it matters more.

### Severity / scope
- **Pre-existing** in `checkSendPermission` (not created by #513) — but #513 newly routes reactions through it with only a `!= 0` check, surfacing the fall-through on a new endpoint.
- **🟡 defense-in-depth**, matching yujiawei's grade: reactions are backstopped by message-exists; no constructible cross-user write via reactions that I can build. But it's a genuine authz fail-open worth closing.
- **Fix:** add a `default: return errBotSendPermCheckFailed` (or explicit final reject) for unhandled `channelType` within each botKind branch — fail-closed on unknown types — and allowlist channel_type at the reaction handler (Group/Person/CommunityTopic only, not just `!= 0`). This also hardens the send path.

### Net (unchanged verdict on the draft)
Still **COMMENT** — #513 is a draft, no gating verdict. This upgrades my earlier channel_type nit to a confirmed should-fix (authz fail-open on unknown type, backstopped for reactions but real), joining the pre-ready 🟡s (emoji bound; query→write TOCTOU on the non-unique index; handler-level test coverage). Credit @yujiawei for the authz catch — I'd flagged "confirm channel_type" but hadn't byte-confirmed the `return nil` fall-through; it's there. When the author marks Ready I'll fold these into the formal verdict; the fail-open-on-unknown-type is the one most worth fixing (in `checkSendPermission`, benefiting every bot write path).

caster-Q:
暂缓此方向:reactions 相关的服务端能力已由另一路径合入 main(存储改为 channel-scoped multi-reaction),本 PR 基于旧结构。三端 reactions 后续随插件整体重构统一重做,先关闭以免占用 review。

PR #483, opened 2026-06-27
feat(bot): summary-assistant bootstrap + ensureFriend endpoint (OCT-4 #96 support)

Ranwanglc:
## OCT-4 / octo-server 侧：专属"总结助手"账号自举 + ensureFriend 端点

关联需求：octo-smart-summary 通知机制（Mininglamp-OSS/octo-smart-summary#96）。本 PR 是其 octo-server 侧支撑。

### 改动
- **新增 `modules/robot/summary_bot.go`**：`insertSummaryRobot()` 仿 `insertSystemRobot()` 范式，开机自举专属"总结助手" User Bot——幂等、配置驱动（`SUMMARY_BOT_UID/NAME/AVATAR/TOKEN` 环境变量），显式落 `bot_token`（`bf_` 前缀，命中 bot_api User Bot 鉴权分支），并建对应 user 账号。配置防漂移：DB 已存在但 token 与配置不一致时以配置为准 UPDATE + warn（不打印 token 明文）；UID/token 缺失则静默跳过自举。在 `modules/robot/api.go:304` 接入。
- **新增 `modules/bot_api/ensure_friend.go`**：`POST /v1/bot/ensureFriend {target_uid, space_id}`，robot token 鉴权。复用 `modules/botfather/friend_approve.go` 配方建立完整可投递关系：双向 `friend` 行（`InsertOrUpdate` 幂等）+ 双向 `IMWhitelistAdd`。让独立部署的 octo-smart-summary 能让"总结助手"给用户发 DM。
- **P1-1**：`IMWhitelistAdd` 的 channel_id 在传入 space_id 时用 `s{spaceID}_{uid}` 前缀（与 `friend_approve.go:182-185` 一致，否则多 Space 部署下 DM 静默投递失败）。
- **P1-2**：`ensureFriend` 端点硬门禁 `robot_id` 必须等于配置的 summary bot UID，否则 403，避免越权放宽其他 bot 权限。

### 验证
- 相关包 `go build` 通过。
- 不改 `send.go` / `auth.go` 发送/鉴权主链路，零侵入。

> 经两位 reviewer + Leader 独立评审通过（含 GLM 复核源码补的两条 P1）。

PR #126, opened 2026-05-22
fix(obo_fanout): decodeMentionGate honors mention.humans (Plan X @所有人) (#125)

yujiawei:
Fixes #125 / YUJ-1709

## Problem

`decodeMentionGate` only checked `mention.all` for the @所有人 broadcast gate. Plan X web client emits `mention: { humans: 1 }` without `mention.all`, so the fan-out early-returned and persona clone bots never received the OBO DM copy.

## Fix

- Extract a `truthy1` helper (accepts float64/int/bool `1`/`true`)
- Add `mention.humans` check after `mention.all` (gated by `if !all` to avoid double-check when both are present)
- Return `all || humans` as the broadcast flag

Safety unchanged: this only widens the mention gate entry condition. Grant DB lookup, Gate 1/2/3, scope/access checks all run as before.

## Tests

6 new tests in `obo_yuj1709_test.go`:
- `TestDecodeMentionGate_YUJ1709_HumansFlagShapes` — 9 sub-tests covering numeric/bool/string/null/combined shapes
- `TestFanout_YUJ1709_GroupMentionHumansSummonsPersona` — end-to-end fan-out with `mention.humans=1`
- `TestFanout_YUJ1709_GroupMentionHumansBooleanShape` — boolean true shape
- `TestFanout_YUJ1709_GroupMentionHumansAllZero` — `all=0, humans=1` (Plan X exact shape)
- `TestFanout_YUJ1709_HumansNoGrant_StillSkips` — no grant → no fan-out (safety)
- `TestFanout_YUJ1709_NoMentionHumansFalse_StillSkips` — `humans=false` → no trigger

PR #517, opened 2026-07-01
feat(bot): summary-assistant bootstrap + ensureFriend endpoint

### LINKED ISSUE #516: 总结助手账号自举 + ensureFriend 端点（支撑 smart-summary 通知机制）
## 背景

octo-smart-summary 需要一个专属「总结助手」账号，向用户主动投递总结完成/失败的 DM 通知。octo-server 侧需提供两项支撑能力，否则独立部署的 summary 服务无法建立可投递的好友关系、也无法拿到稳定的 bot 身份。

关联：Mininglamp-OSS/octo-smart-summary#96（通知机制主 issue）。

## 需求

1. **总结助手账号自举**：服务启动时幂等地创建/校准专属「总结助手」User Bot（配置驱动 UID/NAME/AVATAR/TOKEN），显式落 bot_token（命中 User Bot 鉴权分支），并建对应 user 账号。配置漂移时以配置为准并告警（不打印 token 明文）。
2. **ensureFriend 端点**：`POST /v1/bot/ensureFriend {target_uid, space_id}`，robot token 鉴权，建立双向好友 + 双向 IM 白名单，让总结助手能给用户发 DM。

## 安全要求

- ensureFriend 硬门禁：robot_id 必须等于配置的总结助手 UID，否则 403，避免越权放宽其他 bot 权限。
- 多 Space 部署：IMWhitelist channel_id 传入 space_id 时用 `s{spaceID}_{uid}` 前缀，否则 DM 静默投递失败。

## 验收

- 启动自举幂等，重复启动不产生副作用。
- ensureFriend 建立关系后总结助手可向目标用户投递 DM。
- 不侵入 send/auth 发送鉴权主链路。

Ranwanglc:
## 总结助手账号自举 + ensureFriend 端点

Closes #516
关联：Mininglamp-OSS/octo-smart-summary#96

octo-smart-summary 通知机制的 octo-server 侧支撑。

### 改动
- **新增 `modules/robot/summary_bot.go`**：`insertSummaryRobot()` 仿 `insertSystemRobot()` 范式，开机自举专属「总结助手」User Bot——幂等、配置驱动（`SUMMARY_BOT_UID/NAME/AVATAR/TOKEN`），显式落 `bot_token`（`bf_` 前缀，命中 bot_api User Bot 鉴权分支），并建对应 user 账号。配置防漂移：DB 已存在但 token 与配置不一致时以配置为准 UPDATE + warn（不打印 token 明文）；UID/token 缺失则静默跳过自举。在 `modules/robot/api.go` 接入。
- **新增 `modules/bot_api/ensure_friend.go`**：`POST /v1/bot/ensureFriend {target_uid, space_id}`，robot token 鉴权。复用 `friend_approve.go` 配方建立完整可投递关系：双向 friend 行（InsertOrUpdate 幂等）+ 双向 IMWhitelistAdd。让独立部署的 octo-smart-summary 能让「总结助手」给用户发 DM。
- **P1-1**：`IMWhitelistAdd` 的 channel_id 传入 space_id 时用 `s{spaceID}_{uid}` 前缀（与 friend_approve.go 一致），否则多 Space 部署下 DM 静默投递失败。
- **P1-2**：`ensureFriend` 硬门禁 `robot_id` 必须等于配置的总结助手 UID，否则 403，避免越权放宽其他 bot 权限。
- **bot_token 自举**：启动时自动生成/校准 summary bot_token + space-scoped DM 授权。

### 验证
- 相关包 `go build` 通过。
- 不改 `send.go` / `auth.go` 发送/鉴权主链路，零侵入。

### DIFF
--- a/modules/bot_api/api_i18n_test.go
+++ b/modules/bot_api/api_i18n_test.go
@@ -37,6 +37,7 @@ func TestBotAPINoLegacyResponseError(t *testing.T) {
 		"auth.go", "register.go", "commands.go", "events.go", "mention_pref.go",
 		"sync.go", "typing.go", "send.go", "threads.go", "file.go", "groups.go",
 		"voice_adapter.go", "obo_api.go", "resolve_targets.go", "incoming_webhook.go",
+		"ensure_friend.go",
 	}
 	banned := []string{
 		".ResponseError(",
--- a/modules/bot_api/bot_api.go
+++ b/modules/bot_api/bot_api.go
@@ -207,6 +207,9 @@ func (ba *BotAPI) Route(r *wkhttp.WKHttp) {
 	botAPI := r.Group("/v1/bot", ba.authBot())
 	{
 		botAPI.POST("/sendMessage", ba.sendMessage)
+		// OCT-5: 总结助手好友关系 ensure（DM 可达前置）。端点内部按 UID 白名单
+		// 仅放行配置中的总结助手 bot；其余 bot 调用返回 403。
+		botAPI.POST("/ensureFriend", ba.ensureFriend)
 		botAPI.POST("/typing", ba.typing)
 		botAPI.POST("/readReceipt", ba.readReceipt)
 		botAPI.POST("/events", ba.getEvents)
--- a/modules/bot_api/db.go
+++ b/modules/bot_api/db.go
@@ -1,8 +1,12 @@
 package bot_api
 
 import (
+	"strings"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
 	"github.com/Mininglamp-OSS/octo-lib/config"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/db"
+	pkgspace "github.com/Mininglamp-OSS/octo-server/pkg/space"
 	"github.com/gocraft/dbr/v2"
 )
 
@@ -163,8 +167,54 @@ func (d *botAPIDB) querySpaceIDsByRobotID(robotID string) (string, []string, err
 // single round trip. A single combined query was rejected because OR-of-
 // EXISTS in MySQL with parameter reuse forces the planner to materialize
 // both branches even when the first short-circuits.
-func (d *botAPIDB) isBotSpaceAuthorized(robotID, spaceID string) (bool, error) {
+func (d *botAPIDB) isBotSpaceAuthorized(robotID, spaceID, recipientUID string, channelType uint8) (bool, error) {
 	if robotID == "" || spaceID == "" {
+		return false, nil
+	}
+	// PR#483 (OCT-5) step3/step4 + 第二轮 BLOCKING 修复 — system bot path.
+	//
+	// 系统 bot（如 summary_notification）不依赖 space_member 行获得 Space 归属
+	// （说明见 modules/bot_api/ensure_friend.go step4 注释）。语义参照 platform App
+	// Bot：系统 bot 在所有**活跃** Space 可见。但**纯靠目标 Space active** 来采纳
+	// X-Space-ID 头是一个跨 Space 归属注入漏洞（reviewer 🔴 BLOCKING）：
+	// summary_notification 与 userA（属 spaceA）建立的是**全局**好友关系，send.go 的
+	// DM 鉴权只校验全局好友/OBO，不绑定接收人是否属于 header 声称的 Space。于是
+	// 只要 spaceB active，带 X-Space-ID=spaceB 发给 userA 的 DM 会被错误地归属到
+	// spaceB —— 把 spaceA 用户的私信注入了 spaceB。
+	//
+	// 修复：对 **person channel（DM）** 的系统 bot 发送，X-Space-ID 仅当**接收人确实
+	// 是该 Space 的活跃成员**（pkgspace.CheckMembership(spaceID, recipientUID)=true）
+	// 才采纳；否则不授权（调用方 fall through / strip，绝不跨 Space 注入）。这样把
+	// header 归属严格绑定到 (接收人, Space)，而非仅 (bot, Space active)。
+	//
+	// 非 person channel（群 / thread）：群 / thread 自身已携带 Space 上下文，且这条
+	// header 校验路径在生产里只服务 DM 注入；为不放宽群/thread 的既有语义，维持原
+	// 「目标 Space active 即授权」（系统 bot 在所有活跃 Space 可见）。
+	//
+	// 注意：本分支只影响**发送时的 space_id 解析授权**，不赋予枚举 / 建群能力（那两
+	// 个能力门在 groups.go / service.go 用 IsSystemBot 显式排除）。也不会给 bot 留
+	// 任何 space_member 行。
+	if pkgspace.IsSystemBot(robotID) {
+		if channelType == common.ChannelTypePerson.Uint8() {
+			// DM：必须把 header 声称的 Space 绑定到接收人。接收人不属于该 Space →
+			// 拒绝采纳 header（杜绝跨 Space DM 归属注入）。CheckMembership 自身已
+			// 校验 Space active，无需再单独查 space.status。
+			if strings.TrimSpace(recipientUID) == "" {
+				return false, nil
+			}
+			return pkgspace.CheckMembership(d.session, spaceID, recipientUID)
+		}
+		// 非 DM：维持原 active 校验。
+		var sysCount int
+		if err := d.session.SelectBySql(
+			"SELECT COUNT(*) FROM space s WHERE s.space_id=? AND s.status=1",
+			spaceID,
+		).LoadOne(&sysCount); err != nil {
+			return false, err
+		}
+		if sysCount > 0 {
+			return true, nil
+		}
 		return false, nil
 	}
 	// (1) space_member path: active member row in the target active Space.
--- a/modules/bot_api/ensure_friend.go
+++ b/modules/bot_api/ensure_friend.go
@@ -88,6 +88,40 @@ func (ba *BotAPI) ensureFriend(c *wkhttp.Context) {
 			httperr.ResponseErrorLWithStatus(c, errcode.ErrBotAPINotSpaceMember, nil, nil)
 			return
 		}
+
+		// PR#483 review blocker (Jerry-Xin + OctoBoooot byte-confirmed) — Space
+		// attribution end-to-end gap：summary bot 落在 `robot` 表，没有
+		// `space_member` 行，也没有 `app_bot` 行；后续 `/v1/bot/sendMessage` 的
+		// `resolveBotActiveSpaceID` 三条路径（ctx scope=space / X-Space-ID
+		// validator → isBotSpaceAuthorized / fallback querySpaceIDsByRobotID）
+		// 全 miss → enrichBotPayloadWithSpaceID 把 client-supplied space_id
+		// strip 掉，DM 失去 Space 上下文。
+		//
+		// 修复（方案 a）：target 通过 CheckMembership 后，幂等地给 summary bot
+		// 在**同一个 space_id** 落一行 space_member（role=0 普通成员，最小权限）。
+		// 这样后续 send 路径：
+		//   - query (1) `space_member` 命中 → `isBotSpaceAuthorized` 真，
+		//     X-Space-ID 头会被采纳；
+		//   - 无 X-Space-ID 头时 `querySpaceIDsByRobotID` 兜底也命中。
+		//
+		// 安全性约束：
+		//   - 只对**经过 CheckMembership 验证过 target 的 space_id** 落 summary
+		//     bot 的 space_member —— 攻击面被 target 的 membership 校验完全
+		//     约束在 "target 真实归属的 Space" 内；
+		//   - role=0 普通成员（最小权限），不是 admin/owner，没有管理权限提升；
+		//   - INSERT IGNORE 幂等，不强复活已被运维显式移除的行（status=0 不动）；
+		//   - target_uid（用户）不动，仅给 summary bot 自己落行。
+		//
+		// 失败语义：不阻断好友建立。若 space_member 落库失败，DM 仍会按
+		// fail-closed strip 走（enrich_payload_space_id_empty 监控可见），friend
+		// / 白名单这条主链路对 smart-summary 仍是 best-effort 可用。
+		if _, smErr := ba.db.session.InsertBySql(
+			"INSERT IGNORE INTO space_member (space_id, uid, role, status, version, created_at, updated_at) VALUES (?, ?, 0, 1, 1, NOW(), NOW())",
+			spaceID, robotID,
+		).Exec(); smErr != nil {
+			ba.Warn("ensureFriend 落 summary bot space_member 失败（不阻断好友建立）",
+				zap.String("space_id", spaceID), zap.String("bot_uid", robotID), zap.Error(smErr))
+		}
 	}
 
 	// 1. friend 表双向（幂等 InsertOrUpdate）。
--- a/modules/bot_api/ensure_friend.go
+++ b/modules/bot_api/ensure_friend.go
@@ -11,6 +11,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/modules/user"
 	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
 	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
+	pkgspace "github.com/Mininglamp-OSS/octo-server/pkg/space"
 	"go.uber.org/zap"
 )
 
@@ -69,6 +70,25 @@ func (ba *BotAPI) ensureFriend(c *wkhttp.Context) {
 	}
 
 	spaceID := strings.TrimSpace(req.SpaceID)
+
+	// P1-3：space 成员校验 —— 当请求带 space_id 时，target_uid 必须是该 Space 的活跃
+	// 成员，否则拒绝。复用发送/可见性路径同款 pkg/space.CheckMembership（与
+	// category/api.go、voice_adapter 等一致），防止借总结助手向非本 Space 用户强制建立
+	// 好友（跨 Space 钓鱼/打扰向量）。spaceID 为空（单 Space/平台级）时跳过，与裸 uid
+	// 发送路径语义一致。
+	if spaceID != "" {
+		isMember, err := pkgspace.CheckMembership(ba.db.session, spaceID, targetUID)
+		if err != nil {
+			ba.Error("ensureFriend space 成员校验失败", zap.String("space_id", spaceID), zap.Error(err))
+			httperr.ResponseErrorL(c, errcode.ErrBotAPIStoreFailed, nil, nil)
+			return
+		}
+		if !isMember {
+			ba.Warn("ensureFriend 目标用户非 space 成员，拒绝", zap.String("space_id", spaceID), zap.String("target_uid", targetUID))
+			httperr.ResponseErrorLWithStatus(c, errcode.ErrBotAPINotSpaceMember, nil, nil)
+			return
+		}
+	}
 
 	// 1. friend 表双向（幂等 InsertOrUpdate）。
 	if err := ba.userService.AddFriend(targetUID, &user.FriendReq{
--- a/modules/bot_api/ensure_friend.go
+++ b/modules/bot_api/ensure_friend.go
@@ -89,39 +89,28 @@ func (ba *BotAPI) ensureFriend(c *wkhttp.Context) {
 			return
 		}
 
-		// PR#483 review blocker (Jerry-Xin + OctoBoooot byte-confirmed) — Space
-		// attribution end-to-end gap：summary bot 落在 `robot` 表，没有
-		// `space_member` 行，也没有 `app_bot` 行；后续 `/v1/bot/sendMessage` 的
-		// `resolveBotActiveSpaceID` 三条路径（ctx scope=space / X-Space-ID
-		// validator → isBotSpaceAuthorized / fallback querySpaceIDsByRobotID）
-		// 全 miss → enrichBotPayloadWithSpaceID 把 client-supplied space_id
-		// strip 掉，DM 失去 Space 上下文。
+		// PR#483 Boss 定稿（step4，选择窄路径 —— 不再插 space_member 行）：
 		//
-		// 修复（方案 a）：target 通过 CheckMembership 后，幂等地给 summary bot
-		// 在**同一个 space_id** 落一行 space_member（role=0 普通成员，最小权限）。
-		// 这样后续 send 路径：
-		//   - query (1) `space_member` 命中 → `isBotSpaceAuthorized` 真，
-		//     X-Space-ID 头会被采纳；
-		//   - 无 X-Space-ID 头时 `querySpaceIDsByRobotID` 兜底也命中。
+		// 三个 reviewer 一致卡的 P1 是：给 summary bot 插真实 space_member 行会让
+		// bot 顺带获得 Space 级能力（枚举成员、建群）+ 污染 member_count / 名册 /
+		// @选择器。其中 member_count 是裸 `SELECT COUNT(*) FROM space_member ...`
+		// （见 modules/space/db.go:72/86, db_manager.go:163），**不走 SystemBots 过滤**，
+		// 所以只要还有 space_member 行，member_count 污染就无法根除。因此本次
+		// **不再为 summary bot 插 space_member 行**。
 		//
-		// 安全性约束：
-		//   - 只对**经过 CheckMembership 验证过 target 的 space_id** 落 summary
-		//     bot 的 space_member —— 攻击面被 target 的 membership 校验完全
-		//     约束在 "target 真实归属的 Space" 内；
-		//   - role=0 普通成员（最小权限），不是 admin/owner，没有管理权限提升；
-		//   - INSERT IGNORE 幂等，不强复活已被运维显式移除的行（status=0 不动）；
-		//   - target_uid（用户）不动，仅给 summary bot 自己落行。
+		// DM 的 Space 归属不靠 space_member 行，而是走 IsSystemBot 的窄路径：
+		//   - 发送路径（/v1/bot/sendMessage）带 X-Space-ID 头时，isBotSpaceAuthorized
+		//     对 IsSystemBot(robotID) 直接放行（系统 bot 在所有活跃 Space 可见，语义
+		//     同 platform App Bot）→ X-Space-ID 被采纳 → enrichBotPayloadWithSpaceID
+		//     注入权威 space_id，DM 不丢 Space 归属（详见 modules/bot_api/db.go
+		//     isBotSpaceAuthorized 的 IsSystemBot 分支）。smart-summary 的 send 姿势始终
+		//     携 X-Space-ID（SpaceID 来自 SummaryTask.SpaceID，与本端点 space_id 参数
+		//     同源），所以这是唯一生产 send 形状。
 		//
-		// 失败语义：不阻断好友建立。若 space_member 落库失败，DM 仍会按
-		// fail-closed strip 走（enrich_payload_space_id_empty 监控可见），friend
-		// / 白名单这条主链路对 smart-summary 仍是 best-effort 可用。
-		if _, smErr := ba.db.session.InsertBySql(
-			"INSERT IGNORE INTO space_member (space_id, uid, role, status, version, created_at, updated_at) VALUES (?, ?, 0, 1, 1, NOW(), NOW())",
-			spaceID, robotID,
-		).Exec(); smErr != nil {
-			ba.Warn("ensureFriend 落 summary bot space_member 失败（不阻断好友建立）",
-				zap.String("space_id", spaceID), zap.String("bot_uid", robotID), zap.Error(smErr))
-		}
+		// 结果：(a) DM 仍带正确 space_id（不回归 PR 之前的 DM-attribution bug）；
+		//       (b) bot 无 space_member 行 → 无 Space 能力面；(c) 不污染计数 / 名册。
+		// 本函数不再对 space_member 表做任何写入（friend 表 + IM 白名单仍是 DM 可达的
+		// 唯一依赖，见下文）。
 	}
 
 	// 1. friend 表双向（幂等 InsertOrUpdate）。
--- a/modules/bot_api/ensure_friend.go
+++ b/modules/bot_api/ensure_friend.go
@@ -1,0 +1,155 @@
+package bot_api
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/modules/robot"
+	"github.com/Mininglamp-OSS/octo-server/modules/user"
+	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
+	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
+	"go.uber.org/zap"
+)
+
+// ensureFriendReq 是 POST /v1/bot/ensureFriend 的请求体。
+//
+// 契约（与 octo-smart-summary OCT-6 对齐）：
+//   - TargetUID：被通知用户的 UID（必填）。
+//   - SpaceID：可选。多 Space 部署下，IM 白名单 channel_id 需要 space 前缀
+//     （s{spaceID}_{uid}），由 smart-summary 从 SummaryTask.SpaceID 传入。
+//     缺省（单 Space / 平台级）时走裸 uid 分支，与发送路径 DM 的裸 uid 一致。
+type ensureFriendReq struct {
+	TargetUID string `json:"target_uid"`
+	SpaceID   string `json:"space_id"`
+}
+
+// ensureFriend 为"总结助手"与目标用户建立双向好友关系（DM 可达前置）。
+//
+// 复刻 botfather/friend_approve.go 与 app_bot/app_bot.go 的好友建立配方：
+//  1. friend 表双向（userService.AddFriend，底层 InsertOrUpdate 幂等）；
+//  2. IM 白名单双向 —— 光插 friend 表不够，IM 层投递要白名单；channel_id 在有
+//     Space 时必须用 s{spaceID}_{uid} 前缀形态（P1-1），否则多 Space 下白名单加错
+//     channel，friend 表建了、应用层门过了，但 IM 层不放行 → DM 静默投递失败；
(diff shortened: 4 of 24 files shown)


Jerry-Xin (approved):
**APPROVE** — `ensureFriend` is tightly authorized/scoped, bootstrap is startup-gated + idempotent, no hardcoded secrets, migration is idempotent, all queries parameterized, and the DM space-attribution hole is closed. Byte-verified below.

### ✅ ensureFriend authz (the high-risk path) — properly gated
- Hard UID gate, fail-closed:
  ```go
  summaryUID := robot.SummaryBotUID()
  if summaryUID == "" || robotID != summaryUID {
      ... return 403 (ErrBotAPIEnsureFriendForbidden)
  }
  ```
  `getRobotIDFromContext` returns `""` on missing/wrong-type ctx value; empty `robotID` never equals `summaryUID`, so unauthenticated/other bots are rejected. `SummaryBotUID()` returns the fixed constant `pkgspace.SummaryNotificationBotUID = "summary_notification"` (no env, no drift). **Net: only the single summary-assistant bot's token can call this endpoint** — no arbitrary bot can force-friend users. Adversarial case (any other valid bot token → force-friend a stranger) is blocked at line 1.
- Consent bypass is intentional but scoped to exactly that one system bot, and further constrained by space membership when `space_id` is passed:
  ```go
  if spaceID != "" {
      isMember, err := pkgspace.CheckMembership(ba.db.session, spaceID, targetUID)
      ... if !isMember { return NotSpaceMember }
  }
  ```
  `CheckMembership` is fail-closed: empty inputs → `(false, nil)`; DB error → `(false, err)` which the handler surfaces as 500 (no fail-open fall-through). No cross-space friend injection.
- **Idempotent**: `AddFriend` → `friendDB.InsertOrUpdate` is a real upsert (query `(uid,to_uid)` → UPDATE if exists else INSERT), so re-calls do not create duplicate rows. `fixFriendVersion`/whitelist/`SendCMD` are best-effort and repeatable. Bidirectional rows created correctly (user→bot and bot→user).

### ✅ Bootstrap — gated + idempotent + no secret
- `ensureSummaryBotToken()` runs only inside `Robot.Route()` at server startup — not a user-reachable endpoint.
- Identity rows come from migration `20260629000001` via `INSERT IGNORE` (idempotent; unique index skip). `short_no` uses the UID string (contains `_`, disjoint from numeric/hex user short_nos) — avoids the UNIQUE collision that a numeric value risked. `Down` provided.
- **No hardcoded credential**: migration writes `bot_token=''`; token generated at startup via `crypto/rand` (`bf_` + 32 random bytes). `updateRobotBotTokenIfEmpty` uses `WHERE robot_id=? AND (bot_token='' OR bot_token IS NULL)` → concurrency-safe (only first instance writes; others hit 0 rows). gitleaks secret-scan: pass.

### ✅ Related hardening (bonus, correct)
- `isBotSpaceAuthorized` now binds a system bot's DM `X-Space-ID` adoption to `CheckMembership(spaceID, recipientUID)` for person channels — closes a cross-space DM attribution injection (a spaceA user's DM being attributed to an unrelated active spaceB). Empty recipient → `false`. Non-DM keeps prior active-space semantics. This is a real fix, not a regression.
- `summary_notification` registered in `SystemBots` so downstream filters (member_count, sync-inject, @picker) cover it; Space-capability gates (enumerate/create-group) exclude system bots via `IsSystemBot`, and the design deliberately inserts **no** `space_member` row for the bot — so it gains no Space-level powers and doesn't pollute member counts/rosters.

### ✅ General
- All new SQL is parameterized (`?` placeholders) — no SQLi.
- No fail-open in the new guards (auth gate, membership check, and DM space binding all default to deny/error-out).
- `sync_once` pass-through uses a pointer to distinguish "unset" from explicit 0 — does not change other bots' default (SyncOnce=0) behavior.

### CI
Build ✅ · Vet ✅ · Lint ✅ · secret-scan (gitleaks) ✅ · osv/dependency-review ✅. `Test` was still running at review time (not failing). `check-sprint` fail is the repo's non-code admin gate, not a code-CI failure.

No blocking items.

OctoBoooot (commented):
## Review: feat(bot): summary-assistant bootstrap + ensureFriend endpoint (#517)

**Verdict:** No blocking findings — conditional approve pending CI.
This is the reworked #483: the committed-credential blocker is resolved cleanly and every prior-round P1 is addressed with tests. Solid security engineering. Not stamping APPROVE only because `Test` is still running.

**Precondition:** the `Test` check is still `IN_PROGRESS`. Merge only after it goes green — the new `ensure_friend_test.go` / `summary_bot_test.go` cases are the proof the gates behave, and they run in CI (env-DB integration).

Size: 24 files → `needs-human-review` applied (skill size gate, not a risk-tier escalation).

### What I verified (security surface)
- **No committed credential.** Migration inserts `bot_token=''`; the real token is generated at first startup via `crypto/rand` (32B → `bf_`+hex) and written with a compare-and-set `WHERE bot_token='' OR IS NULL` — idempotent + concurrency-safe as claimed. The `robot.token='summary_notification'` column is *not* an auth path (`queryRobotByBotToken` matches `bot_token` only), so it is not a weak credential.
- **ensureFriend authz.** `robotID` is authenticated — set from `queryRobotByBotToken(token)` in `auth.go`, not a client header. The UID whitelist (`robotID != SummaryBotUID()` → 403) + `space_id`-scoped `CheckMembership` gate hold. `TestEnsureFriend_OtherBotForbidden` / `_NonSpaceMemberForbidden` cover both.
- **No Space capability leak.** No `space_member` row is written (`TestEnsureFriend_E2E_NoSpaceMemberRowLanded`); `botSpaceMembers` and group-create are gated by `IsSystemBot`; DM attribution instead flows through `isBotSpaceAuthorized`'s system-bot branch, which now binds the header Space to the *recipient* (`CheckMembership(spaceID, recipientUID)` for person channels) — closing the prior cross-Space DM-injection blocker.

### Minor
- `ensure_friend.go:~200-235` — IM whitelist adds are best-effort (warn-only) while the handler still returns 200. On a whitelist failure the caller sees success but IM silently drops the DM until a retry. Matches the existing `friend_approve.go`/`app_bot.go` pattern and the flow is idempotent-retriable, so not blocking — but a caller can't distinguish full vs partial success from the 200.

### Praise
- The migration/`summary_bot.go` comments trace the full decision history (env-bootstrap → hardcoded-token → scheme D) and the exact reason each was rejected. That, plus the `short_no='summary_notification'` collision-avoidance reasoning (underscore ∉ generated short_no keyspace), is exactly the context the next maintainer needs.
- The `isBotSpaceAuthorized` recipient-binding fix is the correct minimal closure of the DM-attribution hole — bound to `(recipient, Space)`, not `(bot, Space-active)`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-server PR#517 Review Report — feat(bot): summary-assistant bootstrap + ensureFriend endpoint
**Head SHA:** 88605c3d7942f033d9b5cf4f5613182ff2b623cc
**Scope:** Security-focused automated review review (gstack /investigate + /review)

---

## 1. Verification Summary

| # | Check | Status | Evidence |
|---|-------|--------|----------|
| 1 | UID whitelist gate (only `summary_notification` may call `/v1/bot/ensureFriend`) | ✅ | `modules/bot_api/ensure_friend.go` L60-65; `pkg/space/query.go` SystemBots map |
| 2 | Space membership check on `target_uid` when `space_id` provided | ✅ | `ensure_friend.go` L79-92 → `pkgspace.CheckMembership(session, spaceID, targetUID)` |
| 3 | No `space_member` row inserted for summary bot (step4 narrow path) | ✅ | `ensure_friend.go` L94-118 (comment-only; no INSERT); test `TestEnsureFriend_E2E_NoSpaceMemberRowLanded` asserts count=0 |
| 4 | Cross-Space DM attribution injection blocked (recipient-bound authorization) | ✅ | `modules/bot_api/db.go` L197-205 `isBotSpaceAuthorized` IsSystemBot + person-channel branch: `CheckMembership(spaceID, recipientUID)` |
| 5 | System bot enumeration gate (ability gate A) | ✅ | `modules/bot_api/groups.go` L301 `IsSystemBot(robotID)` → returns empty list; test `TestBotSpaceMembers_SystemBotEnumerationDenied` |
| 6 | System bot group-creation gate (ability gate B, symmetric) | ✅ | `modules/group/service.go` L1040 + L1059 (double check: top-level + Space branch); test `TestCreateGroup_SystemBotForbidden_*` |
| 7 | Token auto-generation race safety | ✅ | `modules/robot/summary_bot.go` `updateRobotBotTokenIfEmpty` WHERE `bot_token='' OR bot_token IS NULL`; affected=0 = lost race (safe) |
| 8 | Migration idempotency | ✅ | `modules/robot/sql/20260629000001_summary_notification_bot.sql` uses `INSERT IGNORE`; down migration present |
| 9 | SyncOnce pointer semantics | ✅ | `modules/bot_api/send.go` L45 `SyncOnce *int`; L201 `if req.SyncOnce != nil`; tests cover nil/0/1 |
| 10 | Fail-closed space_id strip when resolver returns "" | ✅ | `modules/bot_api/space_inject.go` L85-100; test `TestEnsureFriend_E2E_CrossSpaceDMAttributionInjectionBlocked` |

---

## 2. Findings

### No P0 / P1 issues found.

After full data-flow tracing across all new and modified paths, no finding meets the P1 threshold per rubric R1 (would make a working path unavailable / produce user-visible wrong data / make things worse than before the PR).

### P2 — Minor Observations (non-blocking)

**P2-1. `ensureFriend` with empty `space_id` skips target membership check.**
- **Diff-scope:** new behavior (this PR introduces the endpoint).
- **Why P2, not P1:** Documented design for single-Space / platform-level deployment. The UID whitelist gate (only `summary_notification`) already limits the attack surface to one bot. In production, smart-summary always passes `space_id` from `SummaryTask.SpaceID`. An operator deploying without Space scoping accepts this trade-off.
- **Recommendation:** Consider adding an operational log line when `space_id` is empty, to make the platform-level path auditable (e.g., `zap.Bool("platform_level_ensure_friend", true)`).

**P2-2. Token generation failure at startup is logged but does not block server boot.**
- **Diff-scope:** new behavior.
- **Why P2:** Explicit design choice — don't let a bot-bootstrap failure take down the whole server. The bot simply won't be able to authenticate until the next restart or manual token set. Acceptable for a notification-only bot.
- **Recommendation:** Consider emitting a structured alert metric (not just a log line) so ops can detect "summary bot has no token" proactively.

**P2-3. `fixFriendVersion` uses `MAX(version)` across all friends of a uid.**
- **Diff-scope:** replicates existing pattern from `botfather/command.go`.
- **Why P2:** Functionally correct (the WHERE `version=0` on UPDATE preserves idempotency even under concurrent calls), but could produce a version jump larger than necessary. Not a security or correctness concern.

---

## 3. Data Flow Tracing (Iron Rule)

### 3.1 `ensureFriend` → friend table + IM whitelist

```
Request body {target_uid, space_id}
  ↓ authBot middleware: robotID = CtxKeyRobotID (from bf_ token → robot table)
  ↓ UID gate: robotID == SummaryBotUID() (fixed constant "summary_notification")
  ↓ space_id membership: CheckMembership(session, spaceID, targetUID)
  ↓ AddFriend(target→bot) + AddFriend(bot→target) — InsertOrUpdate idempotent
  ↓ IM whitelist: channel_id = s{spaceID}_{uid} (when spaceID present) or bare uid
  ↓ fixFriendVersion bidirectional — best-effort, version>0 for WKSDK sync
  ↓ CMDFriendAccept → Subscribers=[targetUID, robotID]
```

**All consumed data traced to upstream source:**
- `robotID` ← auth middleware (from bot_token → robot table query, `bot_token!='' AND status=1`)
- `summaryUID` ← `SummaryBotUID()` → `pkgspace.SummaryNotificationBotUID` (compile-time constant)
- `targetUID` ← request body (validated non-empty, membership-checked when space_id present)
- `spaceID` ← request body (TrimSpace'd)
- Whitelist channel_id format `s{spaceID}_{uid}` matches send-path `resolveSpaceChannelID` consumer

### 3.2 `isBotSpaceAuthorized` (system bot DM path)

```
Input: (robotID, spaceID, recipientUID, channelType)
  ↓ IsSystemBot(robotID) → true for summary_notification
  ↓ channelType == Person → CheckMembership(session, spaceID, recipientUID)
  ↓   └─ SQL: space_member sm JOIN space s ON s.status=1
  ↓       WHERE sm.uid=recipientUID AND sm.space_id=spaceID AND sm.status=1
  ↓ Returns true only if recipient is active member of active space
  ↓
  ↓ channelType != Person → SELECT COUNT(*) FROM space WHERE space_id=? AND status=1
  ↓ Returns true if target space is active (original semantics preserved)
```

**Critical invariant verified:** The DM path binds authorization to `(recipient, space)`, not just `(bot, space active)`. This prevents the cross-Space DM attribution injection vector where an attacker with summary bot credentials sends `X-Space-ID=victim_space` to a user not in that space. Test `TestEnsureFriend_E2E_CrossSpaceDMAttributionInjectionBlocked` asserts this end-to-end.

### 3.3 Token auto-generation startup path

```
Route() → insertSystemRobot() → ensureSummaryBotToken()
  ↓ queryBotTokenByRobotID: SELECT IFNULL(bot_token,'') FROM robot WHERE robot_id=?
  ↓ If empty → genSummaryBotToken(): crypto/rand 32 bytes → "bf_" + hex (67 chars total)
  ↓ updateRobotBotTokenIfEmpty: UPDATE robot SET bot_token=? WHERE robot_id=? AND (bot_token='' OR bot_token IS NULL)
  ↓ affected=1 → success; affected=0 → another instance won the race (safe, no overwrite)
```

**Race safety verified:** The conditional WHERE clause makes the UPDATE atomic and concurrent-safe across multiple server instances starting simultaneously.

### 3.4 SyncOnce send-path passthrough

```
Request body {sync_once: *int}
  ↓ nil → default (MsgHeader.SyncOnce stays 0, read-fanout)
  ↓ non-nil → msgReq.Header.SyncOnce = *req.SyncOnce
  ↓ dispatch to WuKongIM
```

**Pointer semantics correct:** `*int` distinguishes "not provided" (nil, keep default) from "explicitly 0" (pointer to 0, write 0). Test `TestSendMessage_SyncOnce_*` covers all three cases.

---

## 4. Suggestions

1. **Operational audit log for platform-level ensureFriend** (P2-1): Add a structured warn when `space_id` is empty so ops can detect unexpected platform-level friend establishment.
2. **Token-missing alert metric** (P2-2): Emit a metric/gauge when `ensureSummaryBotToken` fails or skips, enabling proactive alerting.
3. **Documentation**: The step4 narrow path decision (no `space_member` row for system bot) is well-documented in code comments. Consider adding a brief note to the project's bot architecture doc (if one exists) for future maintainers.

---

## 5. Extra Findings

- **Migration `INSERT IGNORE` safety:** The migration inserts into `user` first, then `robot`. Since `robot.robot_id` references `user.uid` conceptually, this ordering is correct. `INSERT IGNORE` makes it idempotent for re-runs.
- **CMDFriendAccept payload includes `space_id`** when provided. Downstream consumers (WKSDK clients) handle this — pre-existing pattern from `botfather/friend_approve.go`, not a new concern.
- **`queryRobotByBotToken` naturally rejects empty tokens:** The WHERE clause `bot_token=? and bot_token!=''` means a migration-inserted empty token cannot be used for auth. The bot becomes authenticatable only after `ensureSummaryBotToken` writes a real token.

---

- **C1 — Dual-path parity:** ✅ Clear. Gates A (groups.go) and B (service.go CreateGroup) use identical `IsSystemBot` check. `isBotSpaceAuthorized` person-channel vs non-person-channel branches are symmetric in their authorization logic. No create↔execute asymmetry found.
- **C2 — Control-flow ordering / nesting reuse:** ✅ Clear. `enrichBotPayloadWithSpaceID` → `resolveBotActiveSpaceID` → `isBotSpaceAuthorized` is a linear call chain with no double-application. No safety control (regex/escape/sanitize) to non-canonical-test.
- **C3 — Authorization boundary ≠ capability boundary:** ✅ Clear. `ensureFriend` endpoint is behind `authBot` middleware (requires valid bf_ token). The UID whitelist further restricts to one bot. No new endpoint is exposed without authorization.
- **C4 — Authorization lifecycle / container-member cascade:** ✅ N/A for this PR. The `CheckMembership` call already JOINs `space` with `s.status=1`, so a disabled space correctly fails membership. No container-level bypass.
- **C5 — Build ≠ runtime path:** ✅ N/A. This PR contains no build artifacts, extensions, CLI packaging, or relative-path concerns. Pure Go server code.
- **C6 — Governance / policy document consistency:** ✅ N/A. This PR does not modify governance, security, or disclosure documents.

---

N/A — this is the automated review review of PR#517. No prior review rounds to re-check.

---

## [Octo-Q] verdict: APPROVED

**Rationale:** No P0 or P1 issues found after full data-flow tracing and R1-R4 severity calibration. The PR implements security gates consistently (UID whitelist, membership check, system bot enumeration/creation gates, recipient-bound DM authorization, fail-closed strip). Test coverage is comprehensive and includes negative cases for all identified attack vectors (cross-Space injection, system bot enumeration, non-member force-friend, concurrent token generation). The three P2 observations are minor operational improvements, not blockers.

Per rubric R4: only P2/nit/maintainability findings → APPROVED.

yujiawei (approved):
# Code Review — PR #517 (octo-server)

**Head SHA reviewed:** `88605c3d7942f033d9b5cf4f5613182ff2b623cc`
Reviewed against merge-base `7ca62654` (`origin/main`). Security-sensitive PR (auth/migration/credential surface) — reviewed with extra scrutiny plus independent cross-model passes. `go build` and `go vet` pass for all changed packages; unit-level tests in `pkg/space` and `modules/robot` (no-DB) pass.

## 1. Spec compliance verdict

**Spec: ✅**

The diff implements the OCT-5 design (fixed-constant summary bot + `ensureFriend` reachability endpoint + system-bot capability gates + cross-Space DM attribution fix) and stays within scope.

- **Missing:** none. All four ensureFriend steps (friend rows / IM whitelist / friend-version fix / `CMDFriendAccept`) are present and mirror the established `botfather`/`app_bot` recipe.
- **Over-build:** none. The migration inserts only identity rows (token left empty, generated at startup); no extra flags/fields/endpoints beyond the stated design.
- **Deviation:** none. The system-bot model is consistent: `summary_notification` registered in `SystemBots`, capability gates use `IsSystemBot` as the single source of truth, and DM Space attribution is bound to the recipient rather than a `space_member` row.

## 2. Code Quality verdict

**Quality: Approved**

No P0/P1 defects were confirmed as introduced by this PR. The core security invariant of this PR — the cross-Space DM attribution fix — was independently verified end-to-end: for a system bot on a person channel, `X-Space-ID` is honored only when `CheckMembership(space, recipient)` holds; otherwise the resolver falls through, finds no `space_member` row, resolves `""`, and the caller fail-closed **strips** `payload.space_id`. The UID gate on `ensureFriend` runs before any side effect and is fail-closed (403). The startup token generation is idempotent and concurrency-safe (`UPDATE ... WHERE bot_token='' OR bot_token IS NULL`), and empty-token rows cannot authenticate (`queryRobotByBotToken` requires `bot_token != ''`), so the endpoint fails closed until a token is written. Migration column lists match the `user`/`robot` schemas; all omitted columns are `NOT NULL DEFAULT`.

The following are **non-blocking P2** observations (hardening / follow-ups), not merge gates:

- **P2 — Pre-existing: platform App Bot DM `X-Space-ID` is not recipient-bound.** `isBotSpaceAuthorized` (`modules/bot_api/db.go:232-244`) authorizes a `scope='platform'` App Bot for any active Space without checking the DM recipient's membership. This is the same class of gap this PR closes for *system* bots, but the App Bot branch is **unchanged from `main`** (verified against the merge-base) — this PR neither introduces nor worsens it. Worth a dedicated follow-up: extend the recipient-membership binding to all person-channel "visible in every active Space" authorization paths, not only the system-bot branch.
- **P2 — No per-UID rate limit on `/v1/bot/ensureFriend`.** The `/v1/bot` group mounts only the global per-IP floor, not `SharedUIDRateLimiter`. Impact is bounded by the single-UID whitelist + idempotency + space-membership check, and the missing per-UID quota is a property of the whole `/v1/bot` group, not new here. Consider a per-UID quota on the force-friend path as defense-in-depth against a compromised bot token.
- **P2 — `space_id`-less `ensureFriend` skips the target-membership guard.** When `space_id` is omitted, the handler still creates friend rows and whitelists bare-uid channels without a membership check. Acceptable for single-Space/platform mode (matches the bare-uid send path), but if the endpoint is meant to be space-scoped, consider requiring `space_id` or gating the bare path behind an explicit mode.
- **P2 — `sync_once` is an unvalidated `*int` passthrough.** Any integer is copied into `MsgHeader.SyncOnce`; consumers only test `== 1`. Restricting to `{0,1}` would make the contract explicit.
- **P2 — Migration is insert-only (`INSERT IGNORE`) and does not repair partial/stale rows;** `ensureSummaryBotToken` treats a missing `robot` row as an `Error`-level failure (noisy on un-migrated deploys) rather than special-casing `dbr.ErrNotFound` the way `resolveBotActiveSpaceID` does. Log-hygiene / robustness only; behavior stays fail-closed.

Refuted concerns worth recording: the generated token (`bf_` + 64 hex = 67 chars) fits `robot.bot_token VARCHAR(100)`; the capability gates do not break existing system bots (`botfather`/`fileHelper`/`u_10000`/`notification` do not authenticate via `/v1/bot` and pass themselves as `BotUID`); and the UID gate cannot be reached by an App Bot (App Bot UIDs are structurally `app_`-prefixed and can never equal `summary_notification`).

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. No blocking defect introduced by this PR. The P2 items above are recommended follow-ups; the pre-existing platform-App-Bot DM attribution gap in particular deserves its own tracked fix, but it is out of scope for this change.

## Items a human may want to manually verify (security-sensitive)

1. Migration `20260629000001` applied against the live `user`/`robot` schema (column names/order) — could not run migrations here.
2. WuKongIM delivery for the `s{spaceID}_{uid}` whitelist channel form vs. the bare-uid dispatch channel in multi-Space deployments.
3. That the smart-summary caller always sends `X-Space-ID` (the no-header path resolves `""` → strips Space attribution by design).

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD 88605c3d

**Formal Verdict**: `APPROVE`



## Reviewer Summary

|---|---|---|
| qa-engineer | pass-with-risk | 全部 AC 有代码路径 + 测试断言、CI 测试作业全绿；P1-1 白名单前缀缺直接断言 + PR 描述漂移，降为 pass-with-risk（均为非阻断 follow-up）。 |
| security-engineer | cleared-with-risk | 无凭据泄露 / 无认证绕过 / 无可利用注入；跨 Space DM 归属注入已收敛、系统 bot 能力面双门锁死；最高 Low，仅存需持有总结助手 token 才能触发的残余信任边界风险。 |
| code-reviewer | approved-with-nits | 0 Must Fix · 0 Should Fix · 4 Nit；三个历史前置阻断项本轮全部落地并经独立复核成立。 |

## Critical Findings (P0-P1)

## Should-Fix (P2)

无 P2 findings — code-reviewer 报 0 Must Fix / 0 Should Fix；QA 与 Security 的风险项均自评为非阻断 follow-up。

## Suggestions (P3)

- [P3-1] `modules/bot_api/ensure_friend.go:137-160` — P1-1 白名单 `channel_id` 前缀 `s{spaceID}_{uid}` 无直接测试断言；代码与 `friend_approve.go` 既有配方一致、正确性可信，属断言强度加固而非覆盖缺失。前缀写错时 friend 行仍建成、应用层放过、端点仍返 200，却在 IM 层 fail-closed 静默不投递（多 Space）。Fix: 加 stub/spy 用例断言带 `space_id` 时 `channel_id == s<spaceID>_<uid>`。 (flagged by qa/Risk-1 + security/Risk-2 + code/Nit-1)
- [P3-2] `modules/bot_api/ensure_friend.go:79` — `space_id` 为空（单 Space / 平台级）时跳过 `CheckMembership`，此时总结助手可对任意 UID 强制建立双向好友 + 开白名单。触发前提是持有该 bot 的 token（UID 门已限定唯一系统 bot），属信任边界内残余风险、非无权攻击者可利用，Security 定为可接受。Fix: 考虑对空 `space_id` 直接拒绝或补 `target_uid` 存在性校验，缩小被攻陷方 blast radius。 (flagged by security/Risk-1 + code/Nit-4)
- [P3-3] `modules/bot_api/ensure_friend.go:143-160` — `IMWhitelistAdd` 失败仅 `Warn` 不阻断、端点仍返 200；调用方（smart-summary）无法区分「已建好友 + 可投递」与「已建好友 + 白名单待补」。Fix: 补 counter（ensureFriend 白名单失败次数）供线上量化。 (flagged by qa/Risk-3 + code/Nit-3)
- [P3-4] PR 描述 — 称「不改 send.go / auth.go 发送鉴权主链路，零侵入」，但 `send.go` 实际被改（新增 `SyncOnce *int` 透传 + `enrichBotPayloadWithSpaceID` 签名扩参）；`auth.go` 确未改。改动受保护（`SyncOnce` nil 保默认、enrich 收窄为接收人绑定）且有专测守护，功能无回归。Fix: 更新 PR 描述以免误导合并者的回归评估。 (flagged by qa/Risk-2 + code/Nit-2)
- [P3-5] `robot.bot_token` — 方案 D 下 bot_token 明文存于共享 `robot` 表供 octo-smart-summary lazy SELECT 鉴权，与平台既有全部 User Bot 的 token-at-rest 模型一致，非本 PR 新增风险。Security 已定为可接受、无 follow-up 要求，记录备案（依赖 DB 访问控制）。 (flagged by security/Risk-3)

## Questions for Author

无 — 三份报告均无开放问题（code-reviewer Question 0）。

## Verdict Rationale

三位 reviewer 全部收敛于低严重度项：code-reviewer 报 **0 Must Fix / 0 Should Fix / 4 Nit**；security-engineer 最高严重度为 **Low**，且全部残余风险（bare-uid 跳过成员校验、白名单 fail-closed、token at-rest）均需持有总结助手 token 方可触发或属设计固有可接受项，**非无权攻击者可利用**；qa-engineer 的三条 pass-with-risk 风险均自评为非阻断 follow-up。去重后共 5 条 P3 建议、0 条 P0/P1/P2、0 条 material question。按 skill 的确定性 Formal Verdict 规则表 first-match（无 P0/P1/P2 + 仅 P3 + 无阻断 Question）→ **APPROVE**。

三个历史前置阻断项（跨 Space DM 归属注入、系统 bot 枚举/建群能力面、明文 token）本轮均已落地并经 code-reviewer 独立复核成立，三份报告无实质分歧（P3-1/P3-2/P3-3/P3-4 均为跨 reviewer 同源指认，一致定级为非阻断）。

## Round Context

---

lml2468 (approved):
**APPROVE — Round 1**

三位自动化 reviewer 一致收敛于低严重度项：code-reviewer 0 Must Fix / 0 Should Fix / 4 Nit，security-engineer 最高 Low（残余风险均需持有总结助手 token 方可触发），qa-engineer pass-with-risk（风险均为非阻断 follow-up）。无 P0/P1/P2、无阻断问题；三个历史前置阻断项（跨 Space DM 归属注入、系统 bot 能力面、明文 token）本轮均已落地并复核成立。

Top follow-ups (P3, 非阻断): (1) `ensure_friend.go` 白名单 `channel_id` 前缀缺直接测试断言；(2) 空 `space_id` 跳过 `CheckMembership` 的残余越权面；(3) PR 描述漂移（send.go 实际被改）。

完整发现见本 PR 上的 Aggregate Report 评论：https://github.com/Mininglamp-OSS/octo-server/pull/517#issuecomment-4864195226

an9xyz:
## Design-level review (not blocking on code quality)

The code itself is solid — test coverage, multi-round review fixes (sweep/CAS/UTF-8 truncate/error sanitize), and detailed rationale in comments are all in good shape. This comment is about whether the **shape of the solution** matches the requirement, before we commit to this footprint.

### Cost/value asymmetry

Reading this PR together with smart-summary#126, the combined footprint is:

- A new dedicated system bot (`summary_notification`) added to `SystemBots`
- A new high-risk endpoint `POST /v1/bot/ensureFriend` that **force-establishes a bidirectional friendship without recipient consent**
- A brand-new "system bot DM authorization bound to recipient membership" branch in `isBotSpaceAuthorized` — introduced specifically to plug the cross-Space DM attribution injection that the system-bot privilege itself created
- ~3600 LOC on the smart-summary side (HTTP client, dedup state machine, sweep, CAS retry, sentinel-based token-race handling, UTF-8-safe truncate, error sanitizer)
- **Cross-service direct DB coupling**: smart-summary reads `robot.bot_token` from the shared IM MySQL

For a delivered artifact that ends up being: a plain-text DM with 4 metadata lines (time range / participants / msg count / generated at), **no result link** (dropped in the latest commit because the SPA route isn't browser-openable). The ROI is heavily inverted.

### Three architectural concerns

**1. Cross-service DB read for credentials is an anti-pattern.**
smart-summary now hard-depends on octo-server's `robot` table schema. Any column rename/type change becomes a cross-repo breaking change. And "no plaintext credential in a public repo" is really "plaintext moved from source to DB row" — anyone with IM DB read access still gets the token. A k8s Secret / env-injected static token gives you rotation, audit, and least-privilege out of the box, and eliminates the entire `ErrTokenNotYetProvisioned` / `markDeferred` / sweep-reclaim state machine that only exists to work around this coupling's startup ordering.

**2. `ensureFriend` is unnecessary new attack surface.**
Force-friending without recipient consent is a strong trust-boundary break in an IM product. The UID whitelist mitigates blast radius, but the capability itself shouldn't need to exist here. `notification` (the platform notification-assistant bot) already delivers system DMs *without* requiring a friend edge. `app_bot` platform-scope bots can also DM arbitrary users. If we need a distinct UI entry for summary delivery, a dedicated conversation category is a cleaner primitive than a friend edge.

**3. \"System bot\" semantics are being stretched.**
Existing entries (`botfather`, `u_10000`, `fileHelper`, `notification`) are platform-side, cross-business roles. `summary_notification` is a *business-service proxy* for one independently-deployed subsystem. \`SystemBots\` is the single source of truth for many downstream filters (opanalytics ETL member-count exclusion, cross-Space conversation fallback, @-picker, client-side filters) — that's why this PR needs explicit \`IsSystemBot(x) -> reject\` gates in \`groups.go\`, \`service.go\`, and the new person-channel branch in \`isBotSpaceAuthorized\`. Those gates are patches on \"it shouldn't have been in SystemBots to begin with\". And the cross-Space DM attribution injection was itself a consequence of granting the \"visible in all active Spaces\" privilege; a plain User Bot would never have had that class of bug.

### Suggested reconsideration

Before merging, could we get a concrete answer to: **why can't smart-summary use `notification` (the existing notification-assistant bot) with a static Secret-injected token?** That path would give us:

- Zero change on octo-server
- No `ensureFriend`, no new `SystemBots` entry, no new authorization branch
- No shared-DB token read, no sweep/defer/sentinel plumbing
- Standard k8s Secret rotation

If there's a product requirement that mandates a UI-distinct \"总结助手\" identity that `notification` cannot carry, that's a valid reason — but it should be stated explicitly, because the current design cost is being paid to satisfy a requirement that isn't documented in either #516 or smart-summary#96.

If the identity distinction is genuinely required, a lighter follow-up is: register `smart-summary` as a platform-scope App Bot (existing capability), keeping the friend-edge and `SystemBots`-privilege changes out of the picture entirely.

Happy to pair on the trade-off. Not blocking the code review on it, but I'd rather we re-scope before this footprint lands than iterate on the fallout.

an9xyz:
## Follow-up: the mechanism already exists — `POST /v1/internal/notify`

Circled back after checking how other independently-deployed OCTO services solve the same problem. The finding makes the ROI concern above much more concrete: **octo-server already ships exactly the primitive smart-summary needs, and a sister service (`octo-matter`) is already using it in production**.

### Server side already exists — `modules/notify/api.go`

- `POST /v1/internal/notify` and `POST /v1/internal/notify/batch` (batch cap 50, per-request targets cap 200)
- Auth: `X-Internal-Token` header, static token via `NOTIFY_INTERNAL_TOKEN` env
- `crypto/subtle.ConstantTimeCompare` for constant-time comparison
- Fail-closed when token unset (`internalAuthMiddleware`)
- Backed by the existing `notification` (通知助手) bot as a global singleton — the module auto-provisions on startup, no per-service bot rows needed
- Request shape: `{space_id, targets, service, event, actor_uid, payload}` — payload carries `message_key` + `params` so the IM server localizes per recipient; caller supplies a fallback string only

### Consumer side already exists — `Mininglamp-OSS/octo-matter internal/notification/octo_notifier.go`

- ~150 LOC total, one HTTP client, 10s timeout, fire-and-forget
- Zero bot identity of its own, zero friend-edge dependency, zero dedup state machine
- Event catalog: `matter.created / matter.status_changed / matter.assignee_added / matter.timeline_entry_added` — the exact `service + event + payload` metaphor summary would want (`smart-summary` service, `summary.completed` / `summary.failed` events)

### What this PR collapses to if summary uses the same path

| Current (this PR + smart-summary#126) | If using `/v1/internal/notify` |
|---|---|
| New system bot `summary_notification` in `SystemBots` | Not needed |
| New endpoint `POST /v1/bot/ensureFriend` (force-friend, no consent) | Not needed |
| New person-channel `CheckMembership` binding branch in `isBotSpaceAuthorized` | Not needed |
| Cross-service DB read of `robot.bot_token` | Not needed — env-injected `X-Internal-Token` |
| ~3600 LOC on smart-summary side (dedup / CAS / sweep / sentinel / defer / UTF-8 truncate / error sanitizer) | ~150 LOC HTTP client, matches octo-matter's shape |
| Migration inserting `summary_notification` user + robot rows | Not needed |
| ~10 new tests to guard the cross-Space DM injection fix | Not needed — the bug class doesn't exist without the SystemBots privilege |
| Startup token-provisioning race (`ErrTokenNotYetProvisioned` sentinel + `markDeferred` + sweep reclaim) | Not needed — static token, no ordering coupling |

### The remaining ask

Given octo-matter is the clear precedent and it's landed and maintained, the design question for this PR reduces to a single thing: **what does summary need that `notification` bot via `/v1/internal/notify` cannot provide?**

If the honest answer is *"product wants a UI-distinct 总结助手 identity"*, that is:
- Worth stating explicitly in #516 (currently isn't documented); and
- Achievable by adding a `sender_kind` / display-alias field to the `/v1/internal/notify` payload so different sender identities render differently in the same delivery pipeline — a much smaller change than forking the entire bot/friend/authorization stack.

Not blocking, but I think the sister-service precedent is strong enough that it warrants an explicit answer before this footprint lands. cc @Ranwanglc — would love your read on whether `/v1/internal/notify` was considered and ruled out for a specific reason.

Ranwanglc:
关闭：自建「总结助手」bot 方案已放弃。summary 通知改走 octo-server 现成的 internal-notify 接口（POST /v1/internal/notify，X-Internal-Token 鉴权，由 server 侧既有「通知助手」代发），server 侧零改动、已就绪。因此本 PR（summary-bot 自举 + ensureFriend 端点）不再需要。对应 summary 侧 #126 同步关闭。关联 issue #516 视情况一并关。

Issue #394, opened 2026-06-17
CreateGroup compensating delete is non-atomic → orphan group rows (needs async reconcile)


an9xyz:
## Summary

`group.CreateGroup` commits the group + members in a DB transaction, then creates the IM channel **after** `tx.Commit()`. If `IMCreateOrUpdateChannel` fails, it runs a best-effort *compensating delete* of the already-committed `group` / `group_member` rows. That compensating delete is itself non-atomic and only logs on failure — so if it also fails (e.g. DB blip right after the IM error), the group row survives **orphaned**: a group with no IM channel that clients can never message.

This is **pre-existing, shared-layer behavior** (all create paths — Web / Bot / integration — go through `group.CreateGroup`), not introduced by any recent PR. It was the root-cause backdrop for the idempotency P1 in #377 and was explicitly deferred from the #381 follow-up umbrella as the one cross-cutting data-consistency item warranting its own issue.

Split out of #381 (PR #383 closed the cheap/local follow-ups; the remaining advisory items were accepted as trade-offs).

## Where

`modules/group/service.go` (around L1230–L1294):

```go
// 提交事务
if err := tx.Commit(); err != nil { ... }       // group + group_member committed here

...

// 创建 IM 频道
err = s.ctx.IMCreateOrUpdateChannel(&config.ChannelCreateReq{...})
if err != nil {
    s.Error("create IM channel failed, performing compensating rollback", ...)
    // Compensating delete runs OUTSIDE any transaction (tx already committed),
    // and each failure is only logged — not retried, not durably queued.
    if _, delErr := s.ctx.DB().DeleteFrom("group_member").Where("group_no=?", groupNo).Exec(); delErr != nil {
        s.Error("compensating delete group_member failed", ...)   // <- orphan window
    }
    if _, delErr := s.ctx.DB().DeleteFrom("group").Where("group_no=?", groupNo).Exec(); delErr != nil {
        s.Error("compensating delete group failed", ...)          // <- orphan window
    }
    return nil, errors.New("failed to create IM channel, group has been rolled back")
}
```

## Failure modes producing an orphan

- IM channel create fails **and** the subsequent compensating `DELETE` fails (DB unavailable, deadlock, connection drop).
- Process crash *between* `tx.Commit()` and a successful IM-channel create / compensating delete.

In both cases a `group` row (and possibly `group_member` rows) remains with no backing IM channel.

## Why it can't be a point fix

[...]

PR #521, opened 2026-07-02
fix(group): async reconcile for orphan group channels #394

lml2468:
# Bug Fix — Issue #394 — orphan group rows (async reconcile)

## Summary

`group.CreateGroup` commits `group` + `group_member` in a DB transaction, then creates
the WuKongIM channel **after** `tx.Commit()`. On IM-create failure it runs a best-effort
compensating delete that **only logs on failure** — so if that delete also fails (DB blip /
deadlock / connection drop) or the process crashes between commit and IM-create, the group
row survives **orphaned**: queryable/listable but with no backing IM channel, unmessageable
forever. This is pre-existing shared-layer behavior (all create paths go through
`CreateGroup`), not a regression.

## Fix — durable marker + async reconcile

The IM channel lives in WuKongIM, not the SQL tx, so it can't be enrolled in `tx`; the local
compensating delete narrows but cannot eliminate the window. Closed durably with an
outbox-style flag + reconcile job:

- **Migration** `20260702000001_group_channel_synced.sql`: adds
  `group.channel_synced TINYINT NOT NULL DEFAULT 1` + composite index
  `(channel_synced, created_at)`. `DEFAULT 1` backfills all existing rows and every
  non-CreateGroup insert path to `1` (= synced) → **zero behavior change**, no false orphans.
- **CreateGroup**: marks `channel_synced=0` in-tx right after the group insert; flips to `1`
  only after a **confirmed** `IMCreateOrUpdateChannel`. Any commit/crash/delete-fail window
  leaves the row at `0`.
- **ReconcileWorker** (`reconcile_worker.go`): periodic ticker scans `channel_synced=0` rows
  older than a grace period, **idempotently** (re)creates the IM channel with the group's
  current subscribable members, then flips `channel_synced=1`. Per-orphan failures only
  log+metric and retry next tick. Observable via prometheus counters
  (`group_channel_reconcile_tick_total`, `group_channel_reconcile_orphan_total`) + logs.
  **Default disabled** (`DM_GROUP_CHANNEL_RECONCILE_ENABLED`), wired into module `Start`/`Stop`.

Covers both failure modes in the issue: IM-fail + delete-fail, and
crash-between-commit-and-IM. Eventually consistent, idempotent, least-destructive
(re-creates the channel rather than hard-deleting valid group data).

## Config (env, all optional)

| Env | Default | Meaning |
|---|---|---|
| `DM_GROUP_CHANNEL_RECONCILE_ENABLED` | `false` | master switch (gradual-rollout friendly) |
| `DM_GROUP_CHANNEL_RECONCILE_INTERVAL` | `5m` | tick interval |
| `DM_GROUP_CHANNEL_RECONCILE_GRACE` | `2m` | only reconcile groups older than this (avoids racing in-flight creates) |
| `DM_GROUP_CHANNEL_RECONCILE_BATCH_SIZE` | `200` | per-tick scan/process cap (max 2000) |

## Files

- `modules/group/sql/20260702000001_group_channel_synced.sql` (new migration, idempotent up/down)
- `modules/group/service.go` (+18) — flag in-tx / confirm after IM create
- `modules/group/db.go` (+29) — `SetChannelSyncedTx` / `SetChannelSynced` / `QueryUnsyncedChannelGroups`
- `modules/group/reconcile_worker.go` (new) — worker + RunOnce
- `modules/group/reconcile_config.go` (new) — env config loader
- `modules/group/metrics.go` (new) — prometheus counters
- `modules/group/1module.go` (+23) — Start/Stop wiring
- `modules/group/reconcile_worker_test.go` / `reconcile_config_test.go` (new) — unit tests

## Tests

Stub-based, no DB dependency (`Test_Issue394_*`, `TestLoadReconcileConfig_Issue394`): orphan
re-create + mark-synced, grace cutoff, IM-fail retry, mark-fail retry, subscriber-query-fail
fallback, disabled/empty batch, query-error propagation, ctx-cancel, disabled Start. All pass.

## Behavior note

When the compensating delete itself fails (the orphan window), reconcile re-creates the
channel and the group becomes fully live — the least-destructive resolution. A creator whose
client saw the transient error will see the group on refresh.

## Rollback

Feature-flag disable (`DM_GROUP_CHANNEL_RECONCILE_ENABLED=false`, the default) fully disables
the worker with no schema rollback needed. The migration's `+migrate Down` drops the column +
index if a full revert is required.

### DIFF
--- a/modules/group/1module.go
+++ b/modules/group/1module.go
@@ -1,6 +1,7 @@
 package group
 
 import (
+	"context"
 	"embed"
 	"fmt"
 	"time"
@@ -30,8 +31,30 @@ func init() {
 		// source_space_* / home_space_* 字段，让 Web/Android/iOS UserInfo 能区分
 		// "同 Space 非好友 → 直接发消息" vs "跨 Space 外部成员 → 仅可在群内交流"。
 		user.RegisterGroupMemberExternalProvider(api.groupService.GetMemberExternalFields)
+
+		// #394：孤儿群频道 reconcile worker。CreateGroup 在 IM 频道创建失败 / 补偿删除失败 /
+		// 崩溃窗口可能残留 channel_synced=0 的群；worker 周期性扫描并幂等补建频道，收敛为
+		// 最终一致。默认关闭（DM_GROUP_CHANNEL_RECONCILE_ENABLED），灰度友好。
+		reconcileWorker := NewReconcileWorker(ctx.(*config.Context), LoadReconcileConfig())
+		// workerCancel 每次 Start 重建，让未来的 graceful-reload 拿到新鲜 ctx；框架当前只
+		// Start 一次，这里的幂等是防误用（与 thread archive worker 同策略）。
+		var workerCancel context.CancelFunc
+
 		return register.Module{
 			Name: "group",
+			Start: func() error {
+				workerCtx, cancel := context.WithCancel(context.Background())
+				workerCancel = cancel
+				reconcileWorker.Start(workerCtx)
+				return nil
+			},
+			Stop: func() error {
+				if workerCancel != nil {
+					workerCancel()
+				}
+				reconcileWorker.Stop()
+				return nil
+			},
 			SetupAPI: func() register.APIRouter {
 				return api
 			},
--- a/modules/group/db.go
+++ b/modules/group/db.go
@@ -38,6 +38,35 @@ func (d *DB) Insert(m *Model) error {
 	return err
 }
 
+// SetChannelSyncedTx 在事务内把某群的 channel_synced 置为指定值（#394）。
+// CreateGroup 在插入群记录后立刻置 0（Model 不含该列，落库走 DEFAULT 1，故需显式翻 0），
+// 使「群已提交、IM 频道尚未确认」这一状态持久可见，交给 reconcile worker 兜底补建。
+func (d *DB) SetChannelSyncedTx(groupNo string, synced int, tx *dbr.Tx) error {
+	_, err := tx.Update("group").Set("channel_synced", synced).Where("group_no=?", groupNo).Exec()
+	return err
+}
+
+// SetChannelSynced 事务外把某群 channel_synced 置为指定值（#394）。
+// CreateGroup 在 IM 频道创建**确认成功**后置 1；reconcile worker 补建成功后同样置 1。
+// 幂等：重复置同一值无副作用。
+func (d *DB) SetChannelSynced(groupNo string, synced int) error {
+	_, err := d.session.Update("group").Set("channel_synced", synced).Where("group_no=?", groupNo).Exec()
+	return err
+}
+
+// QueryUnsyncedChannelGroups 返回 channel_synced=0 且 created_at < cutoff 的群编号（#394）。
+// cutoff 是「宽限截止时刻」：只有创建超过宽限期仍未确认的群才被视作孤儿候选，避免和正在
+// 进行中的建群事务（尚未走到置 1）竞争。按 id 升序 + LIMIT 分批，配合复合索引
+// group_channel_synced_created 走 index range。
+func (d *DB) QueryUnsyncedChannelGroups(cutoff time.Time, limit int) ([]string, error) {
+	var groupNos []string
+	_, err := d.session.SelectBySql(
+		"SELECT group_no FROM `group` WHERE channel_synced=0 AND created_at < ? ORDER BY id ASC LIMIT ?",
+		cutoff.Format("2006-01-02 15:04:05"), limit,
+	).Load(&groupNos)
+	return groupNos, err
+}
+
 // 修改群类型
 func (d *DB) UpdateGroupTypeTx(groupNo string, groupType GroupType, tx *dbr.Tx) error {
 	_, err := tx.Update("group").Set("group_type", int(groupType)).Where("group_no=?", groupNo).Exec()
--- a/modules/group/metrics.go
+++ b/modules/group/metrics.go
@@ -1,0 +1,35 @@
+package group
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// 本文件登记 group 模块的 Prometheus 指标（#394 reconcile 可观测性）。
+//
+// 设计取舍（与 modules/oidc/metrics.go 一致）：
+//   - 注册到全局默认 Registry（promauto.New*），由基础设施统一暴露 /metrics。
+//   - Counter 带 result 维度，便于 Grafana 切分 orphan 检出 / 修复 / 失败比例。
+//   - 不加 group_no / space_id 这类高基维 label，避免 Prometheus 内存爆炸。
+const groupMetricNamespace = "group"
+
+var (
+	// metricReconcileTickTotal reconcile worker 每次 tick 的出口分布：
+	// ran（正常跑完）| error（查询孤儿失败提前返回）。
+	metricReconcileTickTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: groupMetricNamespace,
+		Name:      "channel_reconcile_tick_total",
+		Help:      "Channel reconcile worker tick outcomes (ran|error).",
+	}, []string{"result"})
+
+	// metricReconcileOrphanTotal 单个孤儿群的处理出口分布：
+	//   detected  — 本轮扫描发现的 channel_synced=0 孤儿候选（每个候选 +1）；
+	//   resolved  — 成功补建 IM 频道并翻 channel_synced=1；
+	//   im_failed — IMCreateOrUpdateChannel 失败，留待下个 tick 重试；
+	//   mark_failed — 频道补建成功但翻 channel_synced=1 失败，下个 tick 会再确认。
+	metricReconcileOrphanTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: groupMetricNamespace,
+		Name:      "channel_reconcile_orphan_total",
+		Help:      "Channel reconcile per-orphan outcomes (detected|resolved|im_failed|mark_failed).",
+	}, []string{"result"})
+)
--- a/modules/group/reconcile_config.go
+++ b/modules/group/reconcile_config.go
@@ -1,0 +1,76 @@
+package group
+
+import (
+	"os"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// ReconcileConfig 孤儿群频道 reconcile worker 的运行参数（#394）。
+type ReconcileConfig struct {
+	// Enabled 总开关。disabled 时 worker 不启动 ticker。
+	Enabled bool
+	// Interval 两次 tick 之间的间隔。<=0 视为禁用。
+	Interval time.Duration
+	// Grace 宽限期：只 reconcile created_at 早于 (now-Grace) 的未同步群，避免和正在
+	// 进行中的建群事务（尚未走到 channel_synced=1）竞争。<=0 时回退默认值。
+	Grace time.Duration
+	// BatchSize 单个 tick 扫描 / 处理的孤儿上限。<=0 或越界回退 / 截断。
+	BatchSize int
+}
+
+const (
+	envReconcileEnabled   = "DM_GROUP_CHANNEL_RECONCILE_ENABLED"
+	envReconcileInterval  = "DM_GROUP_CHANNEL_RECONCILE_INTERVAL"
+	envReconcileGrace     = "DM_GROUP_CHANNEL_RECONCILE_GRACE"
+	envReconcileBatchSize = "DM_GROUP_CHANNEL_RECONCILE_BATCH_SIZE"
+
+	defaultReconcileInterval  = 5 * time.Minute
+	defaultReconcileGrace     = 2 * time.Minute
+	defaultReconcileBatchSize = 200
+	// maxReconcileBatchSize 防止运维误填巨数把单 tick 变成对 WuKongIM 的洪峰调用。
+	maxReconcileBatchSize = 2000
+)
+
+// LoadReconcileConfig 从环境变量装载配置。
+// 错误 / 越界值一律回退默认值，避免运维误填导致 worker 行为失控。
+// 默认 Enabled=false：新兜底路径显式开启（灰度友好），与 #394「不引入非预期后台流量」一致。
+func LoadReconcileConfig() ReconcileConfig {
+	return ReconcileConfig{
+		Enabled:   parseReconcileBool(os.Getenv(envReconcileEnabled)),
+		Interval:  parseReconcileDuration(os.Getenv(envReconcileInterval), defaultReconcileInterval),
+		Grace:     parseReconcileDuration(os.Getenv(envReconcileGrace), defaultReconcileGrace),
+		BatchSize: parseReconcileBoundedInt(os.Getenv(envReconcileBatchSize), defaultReconcileBatchSize, maxReconcileBatchSize),
+	}
+}
+
+func parseReconcileBool(raw string) bool {
+	v := strings.ToLower(strings.TrimSpace(raw))
+	return v == "true" || v == "1"
+}
+
+func parseReconcileDuration(raw string, def time.Duration) time.Duration {
+	if raw == "" {
+		return def
+	}
+	d, err := time.ParseDuration(raw)
+	if err != nil || d <= 0 {
+		return def
+	}
+	return d
+}
+
+func parseReconcileBoundedInt(raw string, def, max int) int {
+	if raw == "" {
+		return def
+	}
+	n, err := strconv.Atoi(raw)
+	if err != nil || n <= 0 {
+		return def
+	}
+	if n > max {
+		return max
+	}
+	return n
+}
--- a/modules/group/reconcile_config_test.go
+++ b/modules/group/reconcile_config_test.go
@@ -1,0 +1,102 @@
+package group
+
+import (
+	"os"
+	"testing"
+	"time"
+)
+
+func TestLoadReconcileConfig_Issue394(t *testing.T) {
+	envs := []string{
+		envReconcileEnabled, envReconcileInterval, envReconcileGrace, envReconcileBatchSize,
+	}
+	// 保存并清空相关 env，测试结束后恢复，避免污染其他用例。
+	saved := map[string]string{}
+	for _, e := range envs {
+		saved[e] = os.Getenv(e)
+		os.Unsetenv(e)
+	}
+	defer func() {
+		for _, e := range envs {
+			if v, ok := saved[e]; ok && v != "" {
+				os.Setenv(e, v)
+			} else {
+				os.Unsetenv(e)
+			}
+		}
+	}()
+
+	tests := []struct {
+		name string
+		env  map[string]string
+		want ReconcileConfig
+	}{
+		{
+			name: "defaults when unset (disabled by default)",
+			env:  map[string]string{},
+			want: ReconcileConfig{
+				Enabled:   false,
+				Interval:  defaultReconcileInterval,
+				Grace:     defaultReconcileGrace,
+				BatchSize: defaultReconcileBatchSize,
+			},
+		},
+		{
+			name: "all valid overrides",
+			env: map[string]string{
+				envReconcileEnabled:   "true",
+				envReconcileInterval:  "30s",
+				envReconcileGrace:     "1m",
+				envReconcileBatchSize: "50",
+			},
+			want: ReconcileConfig{
+				Enabled:   true,
+				Interval:  30 * time.Second,
+				Grace:     time.Minute,
+				BatchSize: 50,
+			},
+		},
+		{
+			name: "enabled=1 alias; invalid/negative fall back to defaults",
+			env: map[string]string{
+				envReconcileEnabled:   "1",
+				envReconcileInterval:  "not-a-duration",
+				envReconcileGrace:     "-5s",
+				envReconcileBatchSize: "0",
+			},
+			want: ReconcileConfig{
+				Enabled:   true,
+				Interval:  defaultReconcileInterval,
+				Grace:     defaultReconcileGrace,
+				BatchSize: defaultReconcileBatchSize,
+			},
+		},
+		{
+			name: "batch size clamped to max",
+			env: map[string]string{
+				envReconcileBatchSize: "999999",
+			},
+			want: ReconcileConfig{
+				Enabled:   false,
+				Interval:  defaultReconcileInterval,
+				Grace:     defaultReconcileGrace,
+				BatchSize: maxReconcileBatchSize,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			for _, e := range envs {
+				os.Unsetenv(e)
+			}
+			for k, v := range tt.env {
+				os.Setenv(k, v)
+			}
+			got := LoadReconcileConfig()
+			if got != tt.want {
+				t.Errorf("LoadReconcileConfig() = %+v, want %+v", got, tt.want)
+			}
+		})
+	}
+}
--- a/modules/group/reconcile_worker.go
+++ b/modules/group/reconcile_worker.go
@@ -1,0 +1,199 @@
+package group
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+
+	"go.uber.org/zap"
+)
+
+// reconcileStore ReconcileWorker 对持久层的最小依赖，便于单测注入 stub（无需真实 DB）。
+type reconcileStore interface {
+	// QueryUnsyncedChannelGroups 返回 channel_synced=0 且 created_at<cutoff 的群编号。
+	QueryUnsyncedChannelGroups(cutoff time.Time, limit int) ([]string, error)
+	// SubscribableMemberUIDs 返回某群「可订阅」成员 uid（is_deleted=0 且 status=normal）。
+	SubscribableMemberUIDs(groupNo string) ([]string, error)
+	// SetChannelSynced 把某群 channel_synced 置为指定值。
+	SetChannelSynced(groupNo string, synced int) error
+}
+
+// channelCreator 抽出 IM 频道创建能力（生产实现是 *config.Context），便于单测 mock。
+// IMCreateOrUpdateChannel 语义为「创建或更新」，天然幂等——重复补建已存在的频道安全。
+type channelCreator interface {
+	IMCreateOrUpdateChannel(req *config.ChannelCreateReq) error
+}
+
+// dbReconcileStore 把 *group.DB 适配到 reconcileStore（生产实现）。
+type dbReconcileStore struct{ db *DB }
+
+func (s dbReconcileStore) QueryUnsyncedChannelGroups(cutoff time.Time, limit int) ([]string, error) {
+	return s.db.QueryUnsyncedChannelGroups(cutoff, limit)
+}
+
+func (s dbReconcileStore) SubscribableMemberUIDs(groupNo string) ([]string, error) {
+	return s.db.querySubscribableMemberUIDsWithGroupNo(groupNo)
+}
+
+func (s dbReconcileStore) SetChannelSynced(groupNo string, synced int) error {
+	return s.db.SetChannelSynced(groupNo, synced)
+}
+
+// ReconcileWorker 周期性扫描 channel_synced=0 的孤儿群，幂等补建其 WuKongIM 频道，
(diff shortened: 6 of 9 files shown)


OctoBoooot (commented):
## Review: fix(group): async reconcile for orphan group channels (#521)

**Verdict:** Comment — draft PR; no blockers or majors, the design is sound end-to-end.
This is a well-engineered outbox-pattern fix for the orphan-group problem. Draft, so no APPROVE regardless (and CI Build/Test are SKIPPED on the draft — mark ready-for-review to actually exercise the two new test files). A few minor/observational notes only.

Risk tier: normal (new background worker + a guarded schema migration). Coverage ratio 0.43 is expected here — the worker and config are tested; the "uncovered" entries are DDL / wiring / metrics.

### What I byte-verified (the load-bearing correctness)
- **Migration is textbook-safe.** `20260702000001_group_channel_synced.sql` — INFORMATION_SCHEMA-guarded reentrant procedure, `ADD COLUMN ... NOT NULL DEFAULT 1` so all existing rows backfill to "synced" (no false orphans — the detail that makes this safe), composite index `(channel_synced, created_at)` matching the scan predicate, fully reversible Down. Idempotent on retry.
- **The outbox state machine is correct end-to-end.** `CreateGroup` sets `channel_synced=0` *inside* the tx (service.go:1193), commits, creates the IM channel, and flips to `1` **only after** IM-create succeeds (:1340). On IM-create failure it runs the compensating delete and `return`s *before* the flip (:1323-1335) — so a row survives at `0` exactly in the failure/crash window, which is what the worker heals. Verified the flip is not reachable on the failure path.
- **Worker is idempotent and race-tolerant.** `reconcileOne` creates via `IMCreateOrUpdateChannel` (create-or-update, safe to repeat) *then* flips the flag; if the flip fails, the next tick re-runs safely. Grace cutoff (default 2min > a normal create's commit→IM window) prevents racing an in-flight `CreateGroup`'s own compensating delete. Per-orphan failures are isolated (log+metric, continue). Batch capped at 2000 (WuKongIM flood guard). Disabled by default (gradual rollout).

### Minor / observational
- **Multi-instance redundant work (not a bug).** `QueryUnsyncedChannelGroups` + `reconcileOne` take no row claim/lock, so N server instances running the worker will each reconcile the same orphan — N× `IMCreateOrUpdateChannel` + `SetChannelSynced`. Both are idempotent, so this is wasted work, not corruption. Fine at batch≤2000 and 5-min interval, but worth a one-line comment that correctness relies on idempotency (not exclusivity) in case someone later adds a non-idempotent step to `reconcileOne`.
- **Empty-subscriber fallback.** On `SubscribableMemberUIDs` error, the channel is recreated with `Subscribers=nil`. The comment says WuKongIM's Subscribers datasource backfills members on next reload — reasonable, and it correctly prioritizes "kill the orphan" over "perfect member list." Worth confirming that datasource reload actually fires for a reconciled channel so members aren't invisible until the next membership change.

### Question
- **Worker registration / single-owner.** `NewReconcileWorker` + `Start` are added, but I didn't trace where `Start` is wired into server bootstrap. If it starts on every node, see the multi-instance note; if it's meant to be leader-only, that election isn't in this diff. Confirm the intended deployment (all-nodes-idempotent vs leader-only) matches how `Start` is invoked in `1module.go`.

### Praise
- The `DEFAULT 1` backfill choice is the quietly-correct decision that avoids a NULL-sentinel three-step migration — existing rows are exactly the value you want, so the reconcile scan can never misfire on legacy groups. And placing the `channel_synced=0` write inside the same tx as the group insert (rather than a second post-commit write) is what makes the orphan state actually durable. Clean outbox implementation.

Jerry-Xin (dismissed):
Summary: This PR is in scope for octo-server and implements a reasonable durable marker plus async repair path for group IM channel creation failures.

💬 Non-blocking

🟡 Warning — [modules/group/reconcile_worker.go](/Users/mlamp/.openclaw/tmp/pr-review-tc1pe_h4/repo/modules/group/reconcile_worker.go:171): If `SubscribableMemberUIDs` fails, the worker still calls `IMCreateOrUpdateChannel` with `nil` subscribers and then marks the group synced. This relies on WuKongIM datasource reload semantics. If `/channel` treats `subscribers` as authoritative, this can create or update a channel with an empty subscriber set and stop retries. Consider leaving `channel_synced=0` on subscriber-query failure, or at least proving this with an integration test.

🟡 Warning — [modules/group/service.go](/Users/mlamp/.openclaw/tmp/pr-review-tc1pe_h4/repo/modules/group/service.go:1328): The existing compensating rollback can partially succeed by deleting `group_member` rows and then failing to delete the `group` row. The reconcile worker will then recreate an empty channel and mark it synced. That is probably acceptable as a least-destructive fallback, but it does not match the PR note that the creator will see the group on refresh in every delete-fail case.

🔵 Suggestion — [modules/group/db.go](/Users/mlamp/.openclaw/tmp/pr-review-tc1pe_h4/repo/modules/group/db.go:63) and [modules/group/sql/20260702000001_group_channel_synced.sql](/Users/mlamp/.openclaw/tmp/pr-review-tc1pe_h4/repo/modules/group/sql/20260702000001_group_channel_synced.sql:34): The query filters by `(channel_synced, created_at)` but orders by `id`; the current index may still require a filesort for large unsynced backlogs. Consider indexing `(channel_synced, created_at, id)` or ordering by `created_at, id`.

✅ Highlights

The marker is set to `0` inside the DB transaction and only flipped to `1` after confirmed IM channel creation, which closes the commit-to-IM crash window for new groups.

The worker is idempotent, bounded by batch size, context-aware, and has focused unit coverage for retry paths.

Verification: `go test ./modules/group -run 'Test_Issue394|TestLoadReconcileConfig'` passed.

Jerry-Xin (commented):
> **Verdict: APPROVE-quality — posted as COMMENT because this PR is a draft.** No blocking defects found. Repair path is create-only + idempotent, batch-bounded, context-aware, with focused tests. Mark ready-for-review and I'll re-post a formal Approve.

> ⚠️ Note: Go Test/Vet/Build were **SKIPPED** in CI (path-detection/draft), so the suite was not CI-validated. Verified locally instead: `go test ./modules/group -run 'Test_Issue394|TestLoadReconcileConfig'` **passed**.

---

Summary: This PR is in scope for octo-server and implements a reasonable durable marker plus async repair path for group IM channel creation failures.

💬 Non-blocking

🟡 Warning — [modules/group/reconcile_worker.go](modules/group/reconcile_worker.go:171): If `SubscribableMemberUIDs` fails, the worker still calls `IMCreateOrUpdateChannel` with `nil` subscribers and then marks the group synced. This relies on WuKongIM datasource reload semantics. If `/channel` treats `subscribers` as authoritative, this can create or update a channel with an empty subscriber set and stop retries. Consider leaving `channel_synced=0` on subscriber-query failure, or at least proving this with an integration test.

🟡 Warning — [modules/group/service.go](modules/group/service.go:1328): The existing compensating rollback can partially succeed by deleting `group_member` rows and then failing to delete the `group` row. The reconcile worker will then recreate an empty channel and mark it synced. That is probably acceptable as a least-destructive fallback, but it does not match the PR note that the creator will see the group on refresh in every delete-fail case.

🔵 Suggestion — [modules/group/db.go](modules/group/db.go:63) and [modules/group/sql/20260702000001_group_channel_synced.sql](modules/group/sql/20260702000001_group_channel_synced.sql:34): The query filters by `(channel_synced, created_at)` but orders by `id`; the current index may still require a filesort for large unsynced backlogs. Consider indexing `(channel_synced, created_at, id)` or ordering by `created_at, id`.

✅ Highlights

The marker is set to `0` inside the DB transaction and only flipped to `1` after confirmed IM channel creation, which closes the commit-to-IM crash window for new groups.

The worker is idempotent, bounded by batch size, context-aware, and has focused unit coverage for retry paths.

Verification: `go test ./modules/group -run 'Test_Issue394|TestLoadReconcileConfig'` passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #521 (octo-server)

## Summary

This PR bundles two features: (1) an async reconcile worker for orphan group channels whose IM channel creation failed at group-create time but was never retried (#394), and (2) DM per-Space isolation that scopes DM conversation visibility to the Space where the DM participants actually co-exist (#484). The reconcile worker scans unsynced groups via a `channel_synced` TINYINT column, calls `IMCreateOrUpdateChannel`, and marks the flag on success. DM isolation introduces a `dm_space_presence` table and a generic hide-elsewhere pattern applied to both the v1 sync endpoint and the v2 sidebar.

## Verification

Static analysis only at head `636716fe`; build and tests not executed in this environment.

- ✅ **Migration safety** — `20260702000001_group_channel_synced.sql` uses `ADD COLUMN … NOT NULL DEFAULT 1` (stored procedure with `IF NOT EXISTS` guard). Default 1 is correct: existing rows represent groups that were already created with successful IM channel setup, so they should be treated as synced. The composite index `group_channel_synced_created` covers the worker's `WHERE channel_synced = 0 ORDER BY created_at ASC` query.
- ✅ **Transaction boundary — `service.go:1193`→`:1277`** — `SetChannelSyncedTx(groupNo, 0, tx)` runs inside the group-create transaction. If the tx rolls back, the flag stays at its DEFAULT 1 (the column doesn't exist in a rolled-back row). Post-commit at `:1318`, `IMCreateOrUpdateChannel` runs; on success, `SetChannelSynced(groupNo, 1)` at `:1340` is best-effort. The worker will pick up the group if this best-effort mark fails. Correct fail-open design.
- ✅ **System/org/dept group paths** — `event.go:145` (system group) and `:288` (org/dept group) insert rows without explicitly setting `channel_synced`, relying on `DEFAULT 1`. These paths don't call `IMCreateOrUpdateChannel`, so DEFAULT 1 is correct — they don't need IM channels.
- ✅ **`AddGroup` path — `service.go:203`** — Same pattern: INSERT without explicit `channel_synced` → DEFAULT 1. Correct for the same reason.
- ✅ **Worker lifecycle** — `1module.go` registers `Start`/`Stop` hooks. `ReconcileConfig` defaults to `enabled=false`, so the worker is opt-in via env vars. `Stop()` cancels context and `wg.Wait()` ensures graceful shutdown.
- ✅ **DM presence write path — `webhook/api.go`** — Presence pairs are collected during the message transaction but written best-effort post-commit. If presence write fails, the DM still appears in all Spaces (the pre-PR behavior), not hidden incorrectly. Fail-open is correct here.
- ✅ **`dm_space_presence` migration** — `20260627000001_dm_space_presence.sql` creates the table with `UNIQUE KEY(space_id, uid1, uid2)`. Keying uses `common.GetFakeChannelIDWith` canonical pair ordering, consistent with the rest of the codebase.
- ✅ **Sidebar filter — `api_sidebar.go`** — `filterThreadExtsBySpace` receives `defaultSpaceID` and calls `resolveDMPresence`. On error, `defaultSpaceID` is set to the queried `spaceID`, which means all DMs appear in the queried Space (fail-open, matching pre-PR behavior).
- ✅ **`GetUserDefaultSpaceIDE` — `space_filter.go`** — Replaces `GetUserDefaultSpaceID` with an error-returning variant. Sentinel `defaultSpaceID=""` on error triggers the catch-all path in `space_filter_default_catchall.go`. The empty-string sentinel is checked explicitly before use in `resolveDMElsewhereOnly`, so no nil-deref or silent mis-route.

## Findings

### P2 — Reconcile worker marks `channel_synced=1` after `SubscribableMemberUIDs` failure (`modules/group/reconcile_worker.go:173`)

When `SubscribableMemberUIDs` returns an error, the code logs the error, proceeds with an empty `subscribers` slice, calls `IMCreateOrUpdateChannel` with that empty list, and on IM success permanently sets `channel_synced=1`. The channel is created in the IM system with zero subscribers. Subsequent reconcile ticks will not revisit this group because the flag is already 1. The group's members effectively have no IM channel subscriptions until something else triggers a re-subscribe.

This is unlikely to cause user-visible breakage if `IMCreateOrUpdateChannel` with an empty subscriber list is a valid no-op at the IM layer and a separate subscribe path fills in members later. But if the IM layer treats an empty subscriber list as "create channel with nobody in it" and no other path re-subscribes, the channel is silently broken.

Worth either (a) treating `SubscribableMemberUIDs` failure as a skip-this-tick (don't mark synced, retry next tick) or (b) documenting that empty-subscriber IM creates are safe and a separate subscribe path exists.

### P2 — `filterConversationsCore` fail-open lets all DMs through when default Space lookup errors (`modules/message/space_filter.go`)

When `GetUserDefaultSpaceIDE` fails, `defaultSpaceID` is set to `""`. The catch-all logic in `space_filter_default_catchall.go` then checks `defaultSpaceID == ""` and returns the conversations unfiltered. This is the correct fail-open behavior (pre-PR, all DMs were visible everywhere), but the failure is only logged at debug level. A user whose default Space lookup consistently fails (e.g. no Space membership) will silently bypass DM isolation with no operational signal.

Consider logging at warn level when `defaultSpaceID==""` is hit in production, so operators can detect persistent lookup failures.

### Nit — `context.Background()` as reconcile worker parent (`modules/group/1module.go`)

The reconcile worker's context is derived from `context.Background()` rather than the app's root context. If the app receives a shutdown signal, the worker's `Stop()` hook handles graceful shutdown via `cancel()` + `wg.Wait()`, so this works. However, if a parent context cancellation were ever used for app-wide coordination (e.g. circuit breakers, maintenance mode), the worker would be decoupled. Minor — works today, flagging for awareness.

## Data-Flow Traceback

| Consumed data | Upstream source | Verified flow |
|---|---|---|
| `channel_synced` column value | `DEFAULT 1` in migration / explicit `SetChannelSyncedTx(0)` in tx / `SetChannelSynced(1)` post-commit | Worker reads `WHERE channel_synced=0` → only groups with explicit `0` (set inside successful group-create tx that failed IM) are scanned. Correct. |
| `subscribers` slice in reconcileOne | `SubscribableMemberUIDs(ctx, groupNo)` DB query | On error → empty slice → IM create with no subscribers → still marked synced. See P2 above. |
| `dmPresentSet` in filterConversationsCore | `resolveDMPresence` → `DMSpacePresenceSet` → `dm_space_presence` table lookup | On error → nil set → `filterConversationsCore` returns all conversations unfiltered. Fail-open, correct. |
| `defaultSpaceID` in sidebar filter | `GetUserDefaultSpaceIDE` → `space_member` table query | On error → `""` → catch-all returns all DMs visible. Fail-open, correct. |
| Presence pairs in webhook | Extracted from message payload during tx, written post-commit | Best-effort write; failure doesn't block message delivery or corrupt state. Correct. |
| `dm_space_presence` UNIQUE key | `(space_id, uid1, uid2)` with canonical pair ordering via `GetFakeChannelIDWith` | Consistent with existing DM channel ID generation. ON DUPLICATE KEY UPDATE is a no-op. Correct. |
| Worker batch query | `QueryUnsyncedChannelGroups` → `WHERE channel_synced=0 ORDER BY created_at ASC LIMIT batchSize` | Processes oldest first, bounded batch. Groups that fail reconcile stay at `channel_synced=0` and retry next tick. Correct. |

## Blind-Spot Checklist

- **C1 — Dual-path parity**: Group create paths checked: `CreateGroup` (`service.go:1193`) sets `channel_synced=0` in tx; `AddGroup` (`service.go:203`), system group (`event.go:145`), org/dept group (`event.go:288`) rely on `DEFAULT 1`. The distinction is correct — only `CreateGroup` calls `IMCreateOrUpdateChannel`, so only it needs the reconcile safety net. Clear.
- **C2 — Control-flow ordering**: `IMCreateOrUpdateChannel` at `service.go:1318` runs after the tx commits at `:1277`. If it fails, the group row exists with `channel_synced=0`, so the worker picks it up. The ordering is correct — IM call must be post-commit to avoid calling IM for a group that might be rolled back. Clear.
- **C3 — Authorization boundary**: No auth changes in this PR. The reconcile worker runs server-side only, gated by `enabled=false` default. DM presence writes use the authenticated webhook path. N/A.
- **C4 — Authorization lifecycle**: N/A — no container-member state cascade changes.
- **C5 — Build/runtime**: Static analysis only; no build/test execution in this environment.
- **C6 — Governance/docs**: N/A — no governance or policy changes.

## Verdict: APPROVED

Solid implementation of both features. The reconcile worker's fail-open design (opt-in via env, best-effort mark, retry next tick) and the DM isolation's fail-open on errors both preserve pre-PR behavior as a safe fallback. The two P2 findings are non-blocking — the empty-subscriber IM create path is worth confirming safe at the IM layer, and the silent fail-open on default Space lookup errors is worth an operational signal. Neither produces user-visible incorrect behavior under normal operation.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Two P2 items (reconcile worker marks synced after subscriber-fetch failure; DM isolation fail-open lacks operational signal) are non-blocking. Data-flow traceback confirms correct upstream data flow for all consumed values. C1-C6 checklist clear. Migration safety, transaction boundaries, and fail-open patterns all verified correct.

yujiawei (changes requested):
# Code Review — PR #521 (octo-server)

Reviewed at head SHA `636716feaa20c4343708a1f0cb310679d9a77699`. Verified locally: `go build ./modules/group/` passes and the new `Test_Issue394_*` / config tests pass. Flagged `security-sensitive`, so the lifecycle and data-exposure angles got extra scrutiny.

## Verdict: **CHANGES_REQUESTED**

The outbox-marker + async-reconcile design is sound and matches the intent of #394 (durably close the "committed group row with no backing IM channel" window). Build and unit tests are green, the SQL is parameterized (no injection), the reserved word `group` is correctly backticked/quoted, and the Start/Stop wiring mirrors the existing thread archive worker. One correctness/privacy issue must be resolved before merge; the rest are non-blocking suggestions.

---

## Spec compliance
The diff implements the reconcile safety net described in #394 — durable `channel_synced` marker set in-tx, flipped to `1` only after a confirmed `IMCreateOrUpdateChannel`, plus a periodic idempotent reconciler. Nothing meaningful is missing or out of scope at the feature level. One scope note (non-blocking): the migration backfills every existing row to `channel_synced=1`, so any orphan that already exists in production from the original bug is never scanned — this PR prevents *future* orphans but does not repair the *existing* population. If repairing the current population is in scope, a one-off backfill (set `channel_synced=0` where the IM channel is known missing) would be needed; if not, please state that explicitly.

---

## Blocking (P1)

### 1. Reconcile has no group-status guard — it can resurrect a disbanded/disabled group's channel
`QueryUnsyncedChannelGroups` filters **only** on `channel_synced` and `created_at`, never on `group.status`:

```sql
SELECT group_no FROM `group` WHERE channel_synced=0 AND created_at < ? ORDER BY id ASC LIMIT ?
```
(`modules/group/db.go:64`)

`reconcileOne` (`modules/group/reconcile_worker.go:539-568`) then unconditionally calls `IMCreateOrUpdateChannel` with the group's **current live subscriber roster** and flips `channel_synced=1` — with no check of the group's lifecycle state.

There is a reachable state where a group is `channel_synced=0` **and** `status=Disband(2)` (or `Disabled(0)`):
- `CreateGroup` commits the row at `channel_synced=0` (`service.go` `SetChannelSyncedTx(...,0,tx)` right after insert) and only flips to `1` via a **best-effort, post-commit** `SetChannelSynced(groupNo,1)` (`service.go:1338`). If that single UPDATE fails (a DB blip while the tx + IM create both succeeded), or the process crashes between commit and IM-create, the row is durably stranded at `0` for a fully functional group.
- If that group is later disbanded/disabled, `UpdateStatusTx` (`db.go:358-364`) writes only `status`+`version` and **never touches `channel_synced`**; the disband event handler (`event.go:17`) deletes no members. So `querySubscribableMemberUIDsWithGroupNo` (`db.go:546`, `status=normal AND is_deleted=0`) still returns the full live roster.

On the next tick the worker re-issues a plain `POST /channel` (`IMCreateOrUpdateChannel`, no `disband`/`ban` field) for a channel that disband intentionally shut down via `POST /channel/info` with `disband=1` (`api.go:363-371`, fail-closed), re-populating it with active subscribers. An unattended background job thereby reactivates the backing channel of a group the operator/user explicitly terminated — re-exposing it with its old membership. No authorization or lifecycle check gates this.

It is P1 rather than P0 because reaching the `channel_synced=0`-on-a-terminated-group state requires a specific post-commit failure/crash window *before* the disband, and the worker is default-disabled — but once that state exists and an operator enables the worker (which is the whole point of shipping it), the resurrection is fully automatic and silent.

**Fix:** exclude terminated states from the scan, e.g. `... WHERE channel_synced=0 AND status NOT IN (0, 2) AND created_at < ?`, or re-read `group.status` inside `reconcileOne` and skip (without flipping the marker) when the group is disbanded/disabled. Please add a test for "disbanded group at `channel_synced=0` is NOT reconciled."

---

## Non-blocking (P2 / suggestions)

### 2. Reconcile cutoff is compared as a Go-local wall-clock string against DB-side `created_at`
`QueryUnsyncedChannelGroups` passes `cutoff.Format("2006-01-02 15:04:05")` as a bare string literal (`db.go:64-65`), where `cutoff = now().Add(-grace)` in the Go process's local zone. `created_at` is materialized server-side in the MySQL session time zone (`DEFAULT CURRENT_TIMESTAMP` / `dbr.Expr("Now()")`). Because a pre-formatted string (not a `time.Time`) is bound, the driver's `loc` param does not normalize it — MySQL compares two naked wall-clock strings. If the app-process TZ and the DB session TZ differ by N hours, the grace window shifts by N hours: if the DB is ahead, still-committing rows can be reconciled early; if behind, the safety net silently never fires. With a 2-minute default grace, even a 1-hour skew breaks the window. The sibling `ArchiveStaleBatch` (`modules/thread/db.go:305-319`) avoids this by passing `threshold time.Time` **directly** to the driver. Recommend matching that convention (pass `cutoff` as a `time.Time` param), or compute the cutoff DB-side via `NOW()`/`UTC_TIMESTAMP()`. Harmless when app and DB share one TZ, hence non-blocking — but it's an unstated, load-bearing invariant worth documenting/hardening.

### 3. `reconcileOne` performs DB + IM I/O with no context — `Stop()` can hang
`RunOnce` checks `ctx.Err()` only between orphans; `reconcileOne` calls `SubscribableMemberUIDs`, `IMCreateOrUpdateChannel`, and `SetChannelSynced` with no context (`reconcile_worker.go:539-568`). A hung WuKongIM or DB call blocks the current iteration, and `Stop()`'s `wg.Wait()` blocks module shutdown indefinitely. Consider threading a context (with timeout) into the IM call and DB ops.

### 4. Permanently-failing orphans can starve newer ones
The scan is `ORDER BY id ASC LIMIT batch` with no retry backoff, attempt counter, or dead-lettering. If a batch's worth of low-`id` orphans permanently fail `IMCreateOrUpdateChannel` (e.g. an input the IM layer always rejects), they re-occupy the batch every tick and newer orphans past the LIMIT are never reached. Consider a failure/backoff column or `last_attempt_at` ordering so poison pills don't monopolize the batch.

### 5. Multi-replica behavior
If more than one server replica runs the worker, all replicas scan and reconcile the same rows concurrently (redundant idempotent IM calls + competing `SetChannelSynced` writes). `IMCreateOrUpdateChannel` is idempotent so it's not corrupting, but it's wasteful. If multi-replica deployment is expected, consider a lease/advisory-lock (as the oidc sync worker does) or document that only one replica should enable it.

### 6. Minor: per-create double write
`CreateGroup` inserts the row (DEFAULT `channel_synced=1`) then immediately UPDATEs it to `0` in-tx, doubling the row write/binlog for every group create. This is a deliberate trade-off to keep all other insert paths at DEFAULT 1, so it's acceptable — noting it in case a single-statement insert of the correct initial value is preferred.

### 7. Test coverage
The worker/config unit tests are solid for the reconcile logic, but they stub the store, so the real `QueryUnsyncedChannelGroups` SQL (reserved-word quoting, the tz formatting above), the migration, and the `CreateGroup` marker lifecycle are not exercised. This matches the repo's unit-vs-integration split, so it's non-blocking, but a DB-backed test for the scan query + one service-level failure-window test would meaningfully raise confidence for a security-sensitive path.

---

## Could not verify from the diff (for a human to confirm)
- Actual idempotency/semantics of WuKongIM `IMCreateOrUpdateChannel` on an already-disbanded channel (does re-create clear the `disband`/`ban` flags? — central to the P1 impact).
- Production DB-session vs app-process time zone (determines whether finding #2 bites).
- Migration runner behavior in the real prod/CI pipeline and whether any replica will actually enable the worker.

OctoBoooot (commented):
## Review update — conceding yujiawei's CR (status-blind reconcile scan): #521 @ 636716f

yujiawei's 🔴 is **byte-verified correct**, and it's a real gap my earlier COMMENT missed — I verified the happy-path state machine but did not construct the disband × orphan interaction. Draft PR so the verdict stays COMMENT (non-gating), but the finding is a genuine major that should be fixed before this un-drafts.

### Confirmed finding — reconcile scan resurrects disbanded/disabled channels
- `db.go:64` scan predicate is `WHERE channel_synced=0 AND created_at < ?` — **no `status` filter**.
- `disband()` is a *soft* status change: `UpdateStatusTx(groupNo, GroupStatusDisband, ...)` (api.go:262) leaves the row in place and **does not touch `channel_synced`**.

**Failing input (constructed):** `CreateGroup` commits with `channel_synced=0` in-tx, then IM-create fails *and* the compensating delete also fails (the exact orphan case this worker exists for) → row is `{status=Normal, channel_synced=0}`. The creator then disbands it → row becomes `{status=Disband, channel_synced=0}`. On the next tick the worker's status-blind scan selects it, calls `IMCreateOrUpdateChannel`, **recreates the IM channel for a disbanded group**, and flips `channel_synced=1`. A terminated channel is resurrected. The same applies to `GroupStatusDisabled` (=0). yujiawei's #1 and #2 are the same root cause (status not consulted / not synced).

**Direction:** add `AND status = <Normal>` to the scan predicate (only reconcile live groups), or have `disband`/disable set `channel_synced=1` (or a terminal sentinel) so the row leaves the scan set. The predicate filter is the smaller change and matches the composite index if `status` is added to it.

### On yujiawei's 🟡 (migration backfills all rows to 1)
Agreed and non-blocking — and it's the *correct* default: existing rows genuinely have channels, so `DEFAULT 1` avoids mass false-orphans. If there are *known* pre-existing orphans from before this PR, they won't be detected (they're backfilled to 1) — worth a one-line note in the PR that this fixes the leak forward, not historically. Not a defect.

### Correction to my earlier review
My prior COMMENT praised the worker as clean on the state machine. That was right for the create→confirm→flip path but incomplete: I did not construct the disband/disable interaction with a still-orphaned row. Credit to yujiawei for catching a real resurrection path. The migration/outbox mechanics remain sound; the fix is scoping the scan to live groups.

Jerry-Xin (changes requested):
**Correcting my earlier COMMENT → REQUEST_CHANGES.** On re-verification there's a real correctness/blast-radius bug I missed on my first pass: the reconcile worker can **resurrect the IM channel of an intentionally-dissolved group**. Full credit to the reviewer who flagged it — it's byte-verified real.

## The resurrect path (byte-verified on 636716fe)
1. **Scan predicate has no status filter** — `db.go:64`: `SELECT group_no FROM group WHERE channel_synced=0 AND created_at < ? ORDER BY id LIMIT ?`. Only `channel_synced=0` + grace cutoff; it does **not** exclude `status = GroupStatusDisband`.
2. **Disband leaves `channel_synced` untouched** — `disband()` (`api.go:262`) → `UpdateStatusTx` (`db.go:358-365`) column-writes only `status`+`version`. The row survives (disband is a soft-delete; the only physical group delete is CreateGroup's own compensating rollback). So a dissolved group keeps whatever `channel_synced` it had.
3. **Worker recreates unconditionally** — `reconcileOne` calls `IMCreateOrUpdateChannel` then flips `channel_synced=1`, with no "is this group still alive?" guard.

**Reachable trace:** CreateGroup commits `channel_synced=0` in-tx → IM-create fails and the compensating delete also fails / process crashes in the window (this is the *designed* orphan state, the whole reason this worker exists) → the group is later dissolved (`status=Disband`, `channel_synced` still `0`, row present) → `created_at` ages past the 2-min grace → next tick the scan returns it → the worker re-creates the WuKongIM channel for a deliberately-dissolved group.

## Findings
- 🔴 **Resurrect dissolved-group channel** (the above). Fix: add `AND status <> GroupStatusDisband` (and any other terminal states) to `QueryUnsyncedChannelGroups`, **and/or** have `UpdateStatusTx`/disband set `channel_synced=1` when terminating a group so the row drops out of the scan.
- 🔴 **Status change never syncs `channel_synced`** (`api.go:262` → `db.go:358-365`) — this is the enabler of the resurrect, not a benign inconsistency. Fixing this (flip to 1 on disband) also closes the resurrect.
- 🟡 **Backfill heals only future orphans** — the migration `ADD COLUMN channel_synced TINYINT NOT NULL DEFAULT 1` marks all existing rows synced=1, so any orphan that already exists at deploy time is never healed (only new orphans are). The migration comment acknowledges this; fine as a documented scope limit, just flagging the expectation.

## Severity / fairness note
This is **conditional** — it requires the CreateGroup-failure window (row stuck at `channel_synced=0`) *plus* a subsequent dissolve, not the happy path. The outbox design itself is sound (in-tx flag, flip-after-confirm, grace window, idempotent capped worker, disabled-by-default). So this is a **should-fix-before-ready**, not "the design is broken" — but it must be closed before this leaves draft, because resurrecting a dissolved group's channel is a real user-visible correctness fault.

(Draft PR, so this REQUEST_CHANGES is a pre-ready gate, not a merge block.)

Issue #520, opened 2026-07-02
fix(group): allow group admins to update group avatar, aligning with group name/settings


MrAladdin:
## Problem
Group **admins** get 403 when uploading a group avatar, while the **owner** succeeds (reported via Mininglamp-OSS/octo-web#510, reproduced on Web/iOS/Android). Avatar upload is the only group-profile write still gated owner-only; group name / settings / notice (GROUP.md) are all already manager-or-creator.

## Root cause
`modules/group/api.go` `avatarUpload` (~L721-728) gates with `QueryIsGroupCreator` / `ErrGroupCreatorOnly`.

## Fix
- Change the gate to `QueryIsGroupManagerOrCreator` / `ErrGroupCreatorOrManagerOnly` (both already exist and are widely used), aligning avatar upload with group-name/settings/notice.
- **Fix a pre-existing errcode mismatch**: `groupUpdate` (change group name) decides with `QueryIsGroupManagerOrCreator` but returns `ErrGroupManagerOnly` ("Only a group administrator…"); it should return `ErrGroupCreatorOrManagerOnly`.
- Neutralize stale "群主 uploaded" comments / test names around avatar behavior.

## Known behavior boundary (to confirm at implementation)
`QueryIsGroupManagerOrCreator` additionally requires `is_external=0 AND status=normal`, whereas the old `QueryIsGroupCreator` did not. So an **abnormal-state owner** (external / blacklisted / non-normal) goes allowed→denied. **Normal owners are unaffected.** Decision (per triage): accept this — an abnormal-state owner should not be performing the action anyway. Flagged so the implementer confirms rather than silently regresses.

## Tests
- New: group **admin** uploads avatar → success.
- Keep: **non-admin / non-owner** member → 403 `creator_or_manager_only`.
- Owner path unchanged.

## Compatibility
Relaxation (403→allowed) is backward compatible; the error ID changes `creator_only`→`creator_or_manager_only` (octo-web#510 already expects this rollout).

## Cross-repo
Client-visible symptom tracked in Mininglamp-OSS/octo-web#510. Clients (web/ios/android) already expose the avatar-edit entry to admins, so no client gating change is required — they work automatically once this ships. Optional client-side error-message polish is tracked separately per client.

lml2468:
## Triage Decision

**Type**: `type:bug` (Rule R2 — reproducible bug)

Evidence:
[...]

PR #522, opened 2026-07-02
fix(group): allow group admins to update group avatar (#520)

lml2468:
## Summary

Group **admins** got `403` when uploading a group avatar, while only the **owner** succeeded (Mininglamp-OSS/octo-web#510, reproduced Web/iOS/Android). Avatar upload was the only group-profile write still gated owner-only; group name / settings / notice are already manager-or-creator.

## Root cause

`modules/group/api.go` `avatarUpload` gated with `QueryIsGroupCreator` / `ErrGroupCreatorOnly`, so any non-owner (including admins) was denied.

## Fix

- `avatarUpload`: gate on `QueryIsGroupManagerOrCreator` and return `ErrGroupCreatorOrManagerOnly`, aligning avatar upload with group name / settings / notice.
- `groupUpdate` (change group name): fix a pre-existing errcode mismatch — it decides with `QueryIsGroupManagerOrCreator` but returned `ErrGroupManagerOnly` (`manager_only`); now returns `ErrGroupCreatorOrManagerOnly` (`creator_or_manager_only`) to match the decision.
- Neutralize stale "群主 uploaded" wording in avatar test comments.

## Behavior boundary (confirmed)

`QueryIsGroupManagerOrCreator` additionally requires `is_external=0 AND status=normal`, whereas `QueryIsGroupCreator` did not. An **abnormal-state owner** (external / blacklisted / non-normal) goes allowed→denied; **normal owners are unaffected**. Per triage this is accepted — an abnormal-state owner should not be performing the action anyway.

## Tests

Added to `modules/group/api_i18n_regression_test.go`:
- `TestGroupAvatarUpload_AdminSucceeds` — group admin (role=manager, normal/internal) uploads avatar → `200` (fails before fix).
- `TestGroupAvatarUpload_NonManagerForbidden` — regular member → `403` `err.server.group.creator_or_manager_only`, no object uploaded.

Existing `TestGroupAvatarUpload_PostCommitNotifyFailureRespondsOK` continues to cover the owner path.

## Compatibility

Relaxation (403→allowed) is backward compatible. The error ID changes `creator_only`→`creator_or_manager_only`; octo-web#510 already expects this rollout. No new error codes registered — both codes already exist and are widely used, so i18n markers are unchanged. Clients (web/ios/android) already expose the avatar-edit entry to admins, so no client gating change is required.

## Verification

- `go build ./modules/group/...` — pass
- `go vet ./modules/group/` — pass
- `TestGroupNoLegacyResponseError` source guard — pass (no raw/legacy error responses introduced)
- Full package test execution is blocked locally by a pre-existing test-DB migration/seed inconsistency (fresh-DB migration fails at `20220222000001_user_legacy01.sql`, unrelated to this change and reproduced on the unmodified baseline). CI has a correctly seeded schema.

## Rollback

Straight `git revert` — no migration, config, or feature flag involved.

### DIFF
--- a/modules/group/api.go
+++ b/modules/group/api.go
@@ -718,14 +718,17 @@ func (g *Group) avatarUpload(c *wkhttp.Context) {
 	}
 	defer file.Close()
 
-	isCreator, err := g.db.QueryIsGroupCreator(groupNo, loginUID)
+	// 群头像上传与群名 / 设置 / 群公告(GROUP.md)一致，放行给创建者或管理员
+	// (octo-server#520)。QueryIsGroupManagerOrCreator 额外要求 is_external=0 且
+	// status=normal，因此非正常态的创建者(外部/黑名单)会被拒绝——本就不应执行该操作。
+	isCreatorOrManager, err := g.db.QueryIsGroupManagerOrCreator(groupNo, loginUID)
 	if err != nil {
-		g.Error("查询群创建者失败！", zap.Error(err))
+		g.Error("查询群管理者或创建者失败！", zap.Error(err))
 		httperr.ResponseErrorL(c, errcode.ErrGroupQueryFailed, nil, nil)
 		return
 	}
-	if !isCreator {
-		httperr.ResponseErrorL(c, errcode.ErrGroupCreatorOnly, nil, nil)
+	if !isCreatorOrManager {
+		httperr.ResponseErrorL(c, errcode.ErrGroupCreatorOrManagerOnly, nil, nil)
 		return
 	}
 
@@ -1176,7 +1179,9 @@ func (g *Group) groupUpdate(c *wkhttp.Context) {
 			return
 		}
 		if !isManager {
-			httperr.ResponseErrorL(c, errcode.ErrGroupManagerOnly, nil, nil)
+			// 决策用 QueryIsGroupManagerOrCreator(创建者或管理员均可)，错误码需与之一致，
+			// 而非只提管理员的 ErrGroupManagerOnly (octo-server#520)。
+			httperr.ResponseErrorL(c, errcode.ErrGroupCreatorOrManagerOnly, nil, nil)
 			return
 		}
 	} else {
--- a/modules/group/api_i18n_regression_test.go
+++ b/modules/group/api_i18n_regression_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/Mininglamp-OSS/octo-lib/common"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
 	"github.com/Mininglamp-OSS/octo-lib/testutil"
@@ -298,4 +299,90 @@ func TestGroupAvatarUpload_PostCommitNotifyFailureRespondsOK(t *testing.T) {
 	assert.Equal(t, http.StatusOK, w.Code, "post-commit notify failure should not fail committed upload, body=%s", w.Body.String())
 	assert.Contains(t, w.Body.String(), `"status":200`)
 	assert.NotEmpty(t, mockFS.uploadedPath)
+}
+
+// newAvatarUploadServer wires a group with `testutil.UID` as a member of the
+// given role and returns a router serving only avatarUpload plus the mock file
+// service. Role/status/isExternal let a caller reproduce the admin, non-member
+// and abnormal-owner permission boundaries around octo-server#520.
+func newAvatarUploadServer(t *testing.T, groupNo, creator string, role, status, isExternal int) (http.Handler, *mockAvatarUploadFileService) {
+	t.Helper()
+	_, ctx := newTestServer(t)
+	assert.NoError(t, testutil.CleanAllTables(ctx))
+
+	f := New(ctx)
+	mockFS := &mockAvatarUploadFileService{}
+	f.fileService = mockFS
+
+	assert.NoError(t, f.userDB.Insert(&user.Model{UID: testutil.UID, Name: "uploader", ShortNo: "up_" + groupNo}))
+	assert.NoError(t, f.db.Insert(&Model{
+		GroupNo: groupNo,
+		Name:    "avatar perm",
+		Creator: creator,
+		Status:  GroupStatusNormal,
+		Version: 1,
+	}))
+	// role < 0 means "not a member" — skip the group_member row entirely.
+	if role >= 0 {
+		assert.NoError(t, f.db.InsertMember(&MemberModel{
+			GroupNo:    groupNo,
+			UID:        testutil.UID,
+			Role:       role,
+			Status:     status,
+			IsExternal: isExternal,
+			Version:    1,
+			Vercode:    groupNo + "@1",
+		}))
+	}
+
+	r := wkhttp.New()
+	r.POST("/v1/groups/:group_no/avatar", ctx.AuthMiddleware(r), f.avatarUpload)
+	return r, mockFS
+}
+
+// postAvatar builds a valid multipart body carrying a "file" field and issues
+// the upload as testutil.UID.
+func postAvatar(t *testing.T, handler http.Handler, groupNo string) *httptest.ResponseRecorder {
+	t.Helper()
+	var buf bytes.Buffer
+	mw := multipart.NewWriter(&buf)
+	fw, err := mw.CreateFormFile("file", "avatar.png")
+	assert.NoError(t, err)
+	_, err = fw.Write([]byte("png"))
+	assert.NoError(t, err)
+	assert.NoError(t, mw.Close())
+
+	w := httptest.NewRecorder()
+	req, err := http.NewRequest("POST", "/v1/groups/"+groupNo+"/avatar", &buf)
+	assert.NoError(t, err)
+	req.Header.Set("token", testutil.Token)
+	req.Header.Set("Content-Type", mw.FormDataContentType())
+	handler.ServeHTTP(w, req)
+	return w
+}
+
+// TestGroupAvatarUpload_AdminSucceeds pins the octo-server#520 fix: a group
+// admin (role=manager, normal/internal) may upload the group avatar, aligning
+// avatar upload with group name / settings / notice. Before the fix the
+// QueryIsGroupCreator gate returned 403 for anyone but the owner.
+func TestGroupAvatarUpload_AdminSucceeds(t *testing.T) {
+	handler, mockFS := newAvatarUploadServer(t, "g-avatar-admin", "other-owner", MemberRoleManager, int(common.GroupMemberStatusNormal), 0)
+
+	w := postAvatar(t, handler, "g-avatar-admin")
+	assert.Equal(t, http.StatusOK, w.Code, "群管理员上传头像应成功, body=%s", w.Body.String())
+	assert.Contains(t, w.Body.String(), `"status":200`)
+	assert.NotEmpty(t, mockFS.uploadedPath, "成功路径应真正上传对象")
+}
+
+// TestGroupAvatarUpload_NonManagerForbidden pins that a regular member (neither
+// owner nor admin) still gets 403, now surfaced with the aligned
+// creator_or_manager_only error ID rather than the old creator_only.
+func TestGroupAvatarUpload_NonManagerForbidden(t *testing.T) {
+	handler, mockFS := newAvatarUploadServer(t, "g-avatar-common", "other-owner", MemberRoleCommon, int(common.GroupMemberStatusNormal), 0)
+
+	w := postAvatar(t, handler, "g-avatar-common")
+	assert.Equal(t, http.StatusBadRequest, w.Code, "wire status 固定 400, body=%s", w.Body.String())
+	assert.Contains(t, w.Body.String(), "err.server.group.creator_or_manager_only",
+		"普通成员上传头像应是 403 creator_or_manager_only, body=%s", w.Body.String())
+	assert.Empty(t, mockFS.uploadedPath, "被拒绝时不应上传任何对象")
 }
--- a/modules/group/avatar_get_test.go
+++ b/modules/group/avatar_get_test.go
@@ -213,7 +213,7 @@ func TestGroupAvatarGetCustomTextNotTruncated(t *testing.T) {
 	require.NotEqual(t, truncated, w.Body.Bytes(), "custom text must NOT be GroupNameText-truncated to 2")
 }
 
-// TestGroupAvatarGetUploadedRedirects 覆盖群主已上传自定义头像（is_upload_avatar=1）
+// TestGroupAvatarGetUploadedRedirects 覆盖群主/管理员已上传自定义头像（is_upload_avatar=1）
 // → 维持历史行为：重定向到对象存储，不进服务端渲染。
 func TestGroupAvatarGetUploadedRedirects(t *testing.T) {
 	s, ctx := newTestServer(t)
--- a/modules/group/avatar_version_test.go
+++ b/modules/group/avatar_version_test.go
@@ -7,7 +7,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// TestGroupAvatarVersionDBRoundTrip 覆盖群主手动上传头像的落库路径：
+// TestGroupAvatarVersionDBRoundTrip 覆盖群主/管理员手动上传头像的落库路径：
 // updateAvatar 必须写入版本化对象 path、版本号，并把 is_upload_avatar 置 1
 // （后者是阻止旧的自动合成事件覆盖手动头像的关键标志）。QueryWithGroupNo
 // 读回这些字段，avatarGet 据此选择 versioned/legacy path。
@@ -31,7 +31,7 @@ func TestGroupAvatarVersionDBRoundTrip(t *testing.T) {
 	require.Equal(t, 0, before.IsUploadAvatar)
 	require.Equal(t, int64(0), before.AvatarVersion)
 
-	// 群主手动上传：写版本化 path + 版本号 + is_upload_avatar=1。
+	// 群主/管理员手动上传：写版本化 path + 版本号 + is_upload_avatar=1。
 	const version int64 = 1733300000000000002
 	avatarPath := ctx.GetConfig().GetGroupAvatarFilePath(groupNo, version)
 	require.NoError(t, g.db.updateAvatar(avatarPath, version, groupNo))


OctoBoooot (commented):
## Review: fix(group): allow group admins to update group avatar (#522)

**Verdict:** Comment — draft PR; no blockers or majors, the authz change is correct and bidirectionally covered.
Clean, well-scoped permission fix. Draft, so no APPROVE regardless (CI Build/Test are SKIPPED on the draft — mark ready-for-review to run the new tests). One behavior-change note worth an explicit ack, then it's good.

Risk tier: normal, but I treated it as an authz change (permission widening) — the axis that matters here.

### What I byte-verified
- **The widen is correctly bounded.** `avatarUpload` gate changes `QueryIsGroupCreator` → `QueryIsGroupManagerOrCreator` (api.go:721). The new query requires `is_deleted=0 AND is_external=0 AND status=normal AND (role=creator OR manager)` — so it grants upload to *internal, normal-status* managers/owners only, not external or blacklisted/pending members. This matches the existing name/settings/notice gates, so avatar upload is now consistent with the rest of group-profile writes.
- **Bidirectional coverage is present.** Tests assert both directions: `TestGroupAvatarUpload_AdminSucceeds` (manager → 200 + object actually uploaded) and `TestGroupAvatarUpload_NonManagerForbidden` (common member → 403 with the aligned error id, and *no* object uploaded). That's exactly the widen + still-denied pair an authz change needs.
- **Second change is error-code-only.** `groupUpdate` (api.go:1168) swaps `ErrGroupManagerOnly` → `ErrGroupCreatorOrManagerOnly` to match the decision it already made with `QueryIsGroupManagerOrCreator`. No gate/behavior change — just a more accurate error message. Fine.

### Question (behavior change worth an explicit ack, not a blocker)
- **Inverse edge: a non-normal / external creator loses avatar-upload access.** The *old* `QueryIsGroupCreator` gate checked only `is_deleted=0 AND role=creator` — it did **not** check `is_external` or `status`. The new query adds `is_external=0 AND status=normal`. So a creator who is `is_external=1` or in a non-normal status could upload the avatar before this PR and is now rejected. The code comment calls this out as intended ("非正常态的创建者(外部/黑名单)会被拒绝——本就不应执行该操作"), and it's the safer direction + consistent with the sibling gates — so I read it as a deliberate tightening, not a regression. Flagging only so it's a conscious decision: confirm no legitimate flow relies on an external/abnormal creator editing the avatar. A test for "abnormal-status owner forbidden" would lock this edge (the test helper already takes `status`/`isExternal` params, so it's cheap to add).

### Praise
- Aligning avatarUpload with the exact same `QueryIsGroupManagerOrCreator` query the other profile writes use — rather than hand-rolling a manager-OR-creator check — is the right call: it inherits the internal/normal-status guards for free and keeps all four profile-write gates behaving identically. And covering both the granted and denied paths (with an assertion that nothing is uploaded on denial) is precisely how an authz widening should be tested.

Jerry-Xin (commented):
Reviewed as a **draft** — holding the formal ✅ until it's marked Ready for review, but the change is clean and I found **no blockers**. This is an authorization relaxation, so I byte-verified the access path on head `e199c059`.

🔴 **Blocking: None.**

**Authz verification (the core of this PR):**
- **Role source is trusted (server-side), not client input.** `avatarUpload` uses `loginUID := c.GetLoginUID()` (auth token) + `groupNo := c.Param("group_no")` (URL), then `QueryIsGroupManagerOrCreator(groupNo, loginUID)`. The role is matched in SQL against server-stored `group_member.role` (`role=creator OR role=manager`) — never from request body/header. No forge path.
- **Adversarial traces all deny correctly:**
  - Regular member (`role=MemberRoleCommon=0`) → `(role=1 or role=2)` false → 403. ✅
  - Non-member → no matching `group_no+uid` row → 403. ✅
  - IDOR: query binds both `group_no` (path) and `uid` (caller). Admin of group A passing group B's id has no creator/manager row in B → 403. Group binding enforced. ✅
- **Sibling consistency:** matches `groupUpdate`/notice/settings, which already use `QueryIsGroupManagerOrCreator`. The gate also inherits `is_deleted=0 AND is_external=0 AND status=normal`, so deleted/external/blacklisted managers are correctly excluded (fail-closed). The `ErrGroupCreatorOrManagerOnly` code is pre-existing and widely used by siblings.
- **Errcode fix in `groupUpdate`** (line ~1171): correctly aligns the returned code with the actual predicate — good catch.

**Tests:** Both directions present — `TestGroupAvatarUpload_AdminSucceeds` (allow) **and** `TestGroupAvatarUpload_NonManagerForbidden` (deny, asserts no object uploaded). The deny test is the security-load-bearing one; glad it's here.

💬 **Non-blocking:**
- 🟡 `avatarUpload` parses the multipart body before the authz check (`api.go:703` vs check at ~724). Pre-existing, but an unauthorized authenticated user can make the server process up to the 20M limit before rejection. Consider moving the manager/creator check before `ParseMultipartForm` in a follow-up.
- 🔵 A targeted abnormal-state-owner test would pin the accepted allow→deny boundary the PR body describes (currently only admin-allow + common-deny are covered).
- ℹ️ No SSRF concern: the avatar is an uploaded PNG streamed to a server-generated object path, not a client URL the server fetches.

✅ **Highlights:** minimal, well-scoped, fixes a real inconsistency, and ships both-direction authz tests.

**CI note:** Go suite (Build/Test/Vet/Lint) is **skipped** by path filter for this change set; osv-scan/code-review still in progress. `check-sprint` passed. Nothing red.

Mark ready and I'll convert this to a formal approve.

yujiawei (dismissed):
# Code Review — PR #522 (octo-server)

**Scope reviewed:** head SHA `e199c0598c8099d29476b50a41cf936fc81d7d0a` against merge-base `e1bd41636676f48cb23d4dade351d8e9d9013eb4`.

This is a `security_sensitive` (authorization) change. It was reviewed with extra care, including a cross-family second-opinion pass and adversarial verification of each finding. Verdict: **Approved** — no blocking issues; a few non-blocking advisory notes below.

## 1. Spec compliance

The diff implements exactly what issue #520 asks and nothing more:

- **`avatarUpload`** (`modules/group/api.go:724-732`): gate changed from `QueryIsGroupCreator`/`ErrGroupCreatorOnly` (owner-only) to `QueryIsGroupManagerOrCreator`/`ErrGroupCreatorOrManagerOnly` (owner-or-admin). This aligns avatar upload with the already owner-or-admin group name / settings / notice handlers — the stated goal.
- **`groupUpdate`** (`modules/group/api.go:1162-1171`): the decision function was already `QueryIsGroupManagerOrCreator`, but the denial returned `ErrGroupManagerOnly` (`manager_only`). Now returns `ErrGroupCreatorOrManagerOnly` to match the decision — a genuine pre-existing decision/errcode mismatch fixed in the same code region.
- Comment-only doc alignment in `avatar_get_test.go` / `avatar_version_test.go` (owner → owner/admin). No behavior change.

No over-build (no new error codes, flags, or endpoints introduced — both error IDs already exist and are widely used) and nothing required by #520 is missing.

## 2. Authorization correctness (security focus)

**No privilege escalation.** The new gate reuses the existing `QueryIsGroupManagerOrCreator` helper (`modules/group/db.go:107-111`), whose `WHERE` clause is *stricter* than the old creator check:

```sql
is_deleted=0 AND is_external=0 AND status=normal AND (role=creator OR role=manager)
```

- Admins gaining avatar-upload rights is consistent with every other group-config write; not an escalation.
- `avatarUpload` is reachable only via the authenticated route `POST /v1/groups/:group_no/avatar`; it is the sole caller of the upload path, so no code depended on the old owner-always-allowed semantics.

**Behavior-change worth a human sign-off (documented & accepted in the PR):** because the new helper adds `is_external=0 AND status=normal`, an *abnormal-state owner* (external / blacklisted / non-normal) goes from allowed → denied on avatar upload. This is safe and intentional:

- A normal-flow creator can never be external/non-normal (the creator is Space-validated at group creation and is not placed in the external map), so no legitimate owner is locked out.
- The stricter predicate is already exhaustively pinned at the DB layer by `TestQueryIsGroupManagerOrCreator_ExternalMemberFailSafe` and `TestQueryIsGroupManagerOrCreator_BlacklistedFailSafe` — any relaxation breaks those tests immediately.

## 3. Tests

- `TestGroupAvatarUpload_AdminSucceeds` and `TestGroupAvatarUpload_NonManagerForbidden` exercise the changed gate and assert status code + error ID + no-object-uploaded-on-deny (via the mock file service). Role/status/`is_external` constants are correct.
- The **creator success path** remains covered by the pre-existing `TestGroupAvatarUpload_PostCommitNotifyFailureRespondsOK`, which uploads as a normal creator and asserts `200`.
- `go build ./modules/group/...` and `go vet ./modules/group/` pass. Full package tests could not be run locally due to a pre-existing test-DB migration inconsistency that also fails on the unmodified baseline (unrelated to this change); CI has the correct schema.

## Non-blocking suggestions

- **P2 — partial errcode-consistency cleanup.** Two sibling handlers still have the exact decision-vs-errcode mismatch this PR fixes: `memberUpdate` (`modules/group/api.go:2872` decides with `QueryIsGroupManagerOrCreator` but denies with `ErrGroupManagerOnly` at `:2880`) and `blacklist` (`:3512` → `:3519`). These are outside #520's avatar scope and correctly left untouched here (fixing them would be over-build); consider a follow-up issue to align them and add a regression test.
- **P2 — test-helper over-claim.** `newAvatarUploadServer` (`modules/group/api_i18n_regression_test.go:304-336`) documents "admin, non-member and abnormal-owner" boundaries and accepts `status`/`isExternal`/negative-role params, but the two callers only exercise the normal admin-success and common-member-deny cases; the `role < 0` non-member branch and the abnormal-owner params are never invoked. Either add an HTTP-level non-member/abnormal-owner case here, or trim the branch and the doc comment so coverage claims match. (The abnormal-owner *behavior* is already covered at the DB layer, so this is hygiene, not a security gap.)
- **P2 — client contract note (cross-repo).** The avatar-deny error ID changes `creator_only` → `creator_or_manager_only`. octo-web#510 is stated to expect this; worth a quick confirmation that web/iOS/Android don't switch on the old `creator_only` ID for the avatar path.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

## Sticker Module Deep Review (PR #522, head e199c059)

Scope: modules/sticker/{api.go, db.go, model.go, metrics.go, api_i18n.go, sql/20260630000001_sticker_metadata.sql}, pkg/stickersig/stickersig.go, pkg/metrics/sticker.go

---

### 1. Verification Summary

- **Auth/Authorization**: All routes (GET/POST/PUT/DELETE /v1/sticker/user) go through `AuthMiddleware` + `SharedUIDRateLimiter` (api.go:97). `update()` uses `queryByIDAndUIDTx` (api.go:360) enforcing uid-scoped access within a transaction. `delete()` checks `m.UID != loginUID` (api.go:443). No cross-user escalation path found. **PASS**
- **SQL injection**: All queries use parameterized `?` placeholders via dbr builder or `SelectBySql` with positional args. `shortcodeExistsTx` (db.go:107-121) concatenates SQL structure but all values go through `args...`. No raw string interpolation. **PASS**
- **stickersig crypto**: HMAC-SHA256 with domain-separated subkey derivation, length-prefixed field encoding (prevents field-boundary collision), constant-time `hmac.Equal` comparison. Key derivation requires exactly 32-byte OCTO_MASTER_KEY. **PASS**
- **Migration idempotency**: Uses IF NOT EXISTS checks on information_schema before each ALTER. Stored procedure is created, called, then dropped. **PASS**

---

### 2. Findings

#### F1 — P2: Dual metric registration with divergent label taxonomies

**File**: `modules/sticker/metrics.go:33-37` + `pkg/metrics/sticker.go:75-100`
**Diff-scope**: NEW (both files are new in this PR)

Two separate counter vectors track the same `POST /v1/sticker/user` registration event:

| Metric | FQDN | Labels |
|---|---|---|
| `modules/sticker/metrics.go` (promauto) | `sticker_register_total` | success, validation_failed, path_invalid, missing_handle, invalid_handle, no_capability, quota_exceeded, shortcode_conflict, query_failed, store_failed |
| `pkg/metrics/sticker.go` (explicit) | `dmwork_sticker_register_total` | ok, compat_missing, rejected_missing, rejected_invalid, rejected_path, rejected_no_capability |

Both are registered on the default Prometheus registry (the second via `main.go:222`). They don't collide by name (different namespace), but they track the same event with incompatible label schemas. In `add()`, both are incremented for every request (e.g., api.go:191 increments `metrics.StickerRegisterRejectedPath` AND `observeStickerRegister("path_invalid")`).

**Consequence**: Grafana dashboards need to query two metrics to get a complete picture. Label "success" in one metric maps to "ok" in the other — but they aren't 1:1 (the module-level counter has "store_failed"/"query_failed" that the pkg-level one doesn't). Operational confusion risk.

**Fix**: Consolidate to a single metric definition. Either (a) remove `modules/sticker/metrics.go` and route all observations through `pkg/metrics.ObserveStickerRegister` with the pkg-level taxonomy (adding missing labels like store_failed/query_failed/quota_exceeded), or (b) remove the `Register` CounterVec from `pkg/metrics/sticker.go` and have that package only own `UploadHandleIssued` + `HandlePolicy`.

#### F2 — P2: Misleading log field — `handle_required` actually logs `stickersig.Enabled()`

**File**: `modules/sticker/api.go:321`
**Diff-scope**: NEW (this line is added by this PR)

```go
zap.Bool("handle_required", stickersig.Enabled()),
```

The log field is named `handle_required` (suggesting the policy flag `StickerHandleRequired()`) but the value is `stickersig.Enabled()` (the capability/env flag). These are explicitly documented as orthogonal concepts throughout the PR. A production log entry showing `handle_required=false` would mislead an operator into thinking the policy is off, when in fact it could be `required=true` with `enabled=true`.

**Consequence**: Misleading production log analysis during incident response.

**Fix**: Rename the field to `handle_enabled` or log both:
```go
zap.Bool("handle_enabled", stickersig.Enabled()),
zap.Bool("handle_required", required),
```
(The local `required` variable is already in scope from api.go:175.)

#### F3 — P2: DB column comment claims uniqueness, but no UNIQUE index exists

**File**: `modules/sticker/sql/20260630000001_sticker_metadata.sql:18`

```sql
ADD COLUMN `shortcode` VARCHAR(32) NOT NULL DEFAULT '' COMMENT '客户端联想 shortcode（同 uid live 非空唯一）'
```

The comment says "同 uid live 非空唯一" (unique among live rows per uid when non-empty), but the index `idx_uid_status_shortcode` is a regular INDEX, not UNIQUE. Uniqueness is enforced purely at the application level via `shortcodeExistsTx` inside a transaction with `lockUserRowTx`.

**Consequence**: If a future refactor removes or bypasses the app-level check, there is no database-level safety net. The gap between the column comment and the actual schema is a documentation-as-code defect.

**Fix**: Either (a) correct the column comment to say "应用层保证同 uid live 非空唯一", or (b) add a conditional UNIQUE index. Option (b) is tricky because MySQL UNIQUE indexes don't support partial/filtered constraints, and empty strings must be allowed to coexist. A common pattern is to store NULL instead of '' for "no shortcode" and add a UNIQUE index on (uid, shortcode) where NULL is exempt. The current app-level enforcement with `lockUserRowTx` serialization is correct, so (a) is the lower-risk fix.

#### F4 — P2: Migration DDL inside stored procedure causes implicit commits

**File**: `modules/sticker/sql/20260630000001_sticker_metadata.sql:12-37`
**Diff-scope**: NEW

MySQL executes DDL (ALTER TABLE) with implicit commits, even inside a stored procedure. If a pod dies after the first `ADD COLUMN shortcode` commits but before `ADD INDEX`, a re-run would:
- Skip the `ADD COLUMN shortcode` (IF NOT EXISTS catches it) ✓
- Skip the `ADD COLUMN keywords` (IF NOT EXISTS catches it) ✓
- Re-attempt the `ADD INDEX` ✓

The idempotency guards handle this correctly for the UP direction. However, note that the migration is wrapped in a single stored procedure `__sticker_metadata()` called once — if any ALTER TABLE fails (e.g., disk full), the procedure halts mid-way and the previous successful ALTERs are already committed. The IF NOT EXISTS guards on re-run make this safe, but the implicit-commit behavior means this is NOT transactional in the traditional sense.

**Consequence**: Low risk given the idempotency guards, but operators should be aware that partial failures leave committed partial state (not rolled back).

**Fix**: This is inherent to MySQL DDL. The idempotency guards are adequate. Consider adding a comment in the migration noting the implicit-commit behavior for operator awareness.

---

### 3. Data-Flow Tracing

| Consumed data | Source | Verified? |
|---|---|---|
| `loginUID` in add/update/delete | `ctx.GetLoginUID()` ← AuthMiddleware sets from JWT/session | ✓ |
| `req.Path` in add() | Client JSON body → `BindJSON` → length-checked (≤512) → `validateStickerPath` regex + handle verify | ✓ |
| `req.Shortcode` in add/update | Client JSON → `normalizeStickerShortcode` (regex `^[a-z0-9_]{2,32}$`) → `shortcodeExistsTx` uniqueness check | ✓ |
| `req.Keywords` in add/update | Client JSON `[]string` → `normalizeStickerKeywords` (count≤10, rune-len≤20, JSON≤255 bytes) → stored as JSON string | ✓ |
| `stickerID` in update/delete | URL param `:sticker_id` → used in parameterized DB queries (no interpolation) | ✓ |
| `m.Shortcode` in update() conflict check | Either existing DB value (if req.Shortcode == nil) or newly normalized value → passed to `shortcodeExistsTx` with `excludeStickerID=m.StickerID` | ✓ |
| `metrics.ObserveStickerRegister` label | Always one of the pre-defined `StickerRegister*` constants → no dynamic label injection | ✓ |

---

### 4. Blind-Point Checklist (C1–C6)

- **C1 (Dual-path parity)**: add() ↔ update() — both normalize shortcode/keywords identically via the same `normalizeStickerShortcode` / `normalizeStickerKeywords` functions. Both check shortcode uniqueness via `shortcodeExistsTx`. Both use `lockUserRowTx` for serialization. **CLEAR**
- **C2 (Control-flow ordering)**: `classifyStickerPath` is called exactly once in add(). The `required && !enabled` fail-closed check precedes it, preventing any path from reaching classify when the config is contradictory. **CLEAR**
- **C3 (Authorization boundary)**: All sticker endpoints require AuthMiddleware. No admin/manager bypass path exists. `delete()` and `update()` both enforce uid-scoping (delete via explicit check, update via `queryByIDAndUIDTx`). **CLEAR**
- **C4 (Authorization lifecycle)**: N/A — stickers are per-user, no container/parent entity hierarchy.
- **C5 (Build/runtime parity)**: N/A — no browser extension / CLI / packaging changes in sticker module.
- **C6 (Policy/document consistency)**: The column comment vs schema mismatch (F3) is a minor instance of this. Not security-impacting.

---

### 5. Cross-Round Blocker Review (R6)

N/A — first review pass.

---

- **stickersig subkey() re-reads env on every call**: `subkey()` calls `os.Getenv(masterKeyEnv)` on every Sign/Verify. This is correct (no stale key after runtime env change) but means the env lookup happens on every sticker upload and registration. Negligible cost for the call volume.
- **Empty shortcode bypass in shortcodeExistsTx**: `shortcodeExistsTx` returns `false, nil` immediately when `shortcode == ""` (db.go:103). This is intentional — all existing stickers default to '' and must coexist. The application layer enforces non-empty uniqueness only.
- **delete() TOCTOU**: The `delete()` handler reads the sticker outside a transaction (api.go:439-442), checks ownership, then updates. A concurrent delete of the same sticker by the same user would both pass the ownership check and the second would find `status != 1` (already deleted by the first) — benign idempotent behavior.

---

### 7. [Octo-Q] Verdict: APPROVE

No P0/P1 findings. All four findings are P2 (operational clarity / documentation accuracy / metric hygiene). The core security properties — authorization, input validation, SQL injection prevention, cryptographic signing — are all correctly implemented. The dual-metric issue (F1) is the most actionable and should be consolidated before the handle enforcement rollout goes to production, but it does not block merge.

OctoBoooot (commented):
## Delta re-review: fix(group): allow group admins to update group avatar (#522) @ 255de781

**Verdict: Approve with comments** (unchanged from my prior COMMENT). The branch was rebased onto main since my last review (`e199c05`), and the real code delta is a **new, well-tested authz widening** in `groupUpdate` — not just the avatar fix. No blockers, no majors. Held at COMMENT only because this is a **draft**: Build/Test/Vet/Lint are all SKIPPED, so the new permission-matrix tests have never executed. Mark ready-for-review to run them, and this converts to a clean Approve.

Delta isolated by per-file blob compare (the branch-vs-branch diff was noisy from the rebase — 32 unrelated commits): the only PR file that actually changed since `e199c05` is `modules/group/api.go`; the three test files are byte-identical at both heads, and `rename_permission_matrix_test.go` came in with the rebase.

### What changed since my last review (and what I verified)
- **New permission tiering in `groupUpdate`** (api.go:~1160): advanced fields (notice/invite/avatar_text/avatar_color) still require manager/creator; a **name-only** change is now allowed for any *active internal human* member; anything else (no name, no advanced — e.g. unknown fields) falls back to the manager gate. This is a real permission widening beyond the original avatar scope.
- **The widen is correctly bounded.** `ExistMemberActiveInternal` enforces `is_deleted=0 AND is_external=0 AND status=normal` — so it preserves the exact `is_external=0` boundary the old `QueryIsGroupManagerOrCreator` gate had (cross-space external members still can't rename — YUJ-231/GH#1289), plus a `Service.IsRobot` exclusion so lobsters can't rename.
- **No field-smuggling via the member path** (my adversarial check): `groupUpdate` reads exactly 5 keys (invite/name/notice/avatar_text/avatar_color). A member reaching the widened branch requires `hasName && !hasAdvanced`, and the write section only persists `name` there (notice is advanced → manager-gated). There is no other writable attribute, so a plain member cannot set a sensitive field through the name-only path. **Unknown-field-only requests route to `!hasName` → manager fallback (fail-closed).**
- **Test coverage is thorough** — `rename_permission_matrix_test.go` is a real matrix that would fail if any predicate were removed: normal human member name-only → allowed; robot member → denied; blacklist-status → denied; **external member → denied**; normal member + advanced field → `manager_only`; empty body → invalid; **unknown-field-only → manager required (fail-closed)**. This closes the exact edges I asked about in my prior review.

### Prior question — now resolved
- My earlier note (abnormal/external creator losing avatar access; wanting an "abnormal-status forbidden" test) is now covered by the permission matrix's external/blacklist/robot denial rows. Considered addressed.

### Minor
- The PR title still says only "allow group admins to update group **avatar**", but the change now also widens **rename** to plain members. Worth updating the title/description so the reviewer/maintainer and the squash-commit history reflect the broader authz change (a name-only rename widening is more load-bearing than an avatar-permission tweak).

### Praise
- The tiering reuses `ExistMemberActiveInternal` (rather than a hand-rolled membership check) so it inherits the `is_external=0`/normal-status guards for free, and the accompanying matrix test exercises every allow/deny axis — role, member-status, robot, and external — which is exactly how a permission widening should be pinned. The fail-closed fallback for unknown/no-name requests is the right default.

### Note on merge gate
Draft + `needs-human-review` → `mergeStateStatus` BLOCKED. Approved ≠ mergeable; this needs the draft cleared (so CI runs), green Build/Test/Vet/Lint, and a human maintainer's call on the widened permission.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #722 (octo-web)

## Summary

This PR adds line-height and paragraph spacing (spaceBefore / spaceAfter) to the docs editor as three new global attributes on the `heading` and `paragraph` nodes (SCHEMA_VERSION 17). The implementation is a self-built Tiptap extension modelled on the existing `TextAlign` pattern, with defence-in-depth sanitisation at parse, render, command, and DOCX-export boundaries. The toolbar gains a line-height dropdown with a commit-on-blur custom-multiplier input, plus two paragraph-spacing dropdowns (before / after). The PR also bumps SCHEMA_VERSION from 15 to 17, deliberately skipping the reserved v16 (CJK, in-flight).

The code is well-structured, thoroughly tested, and the sanitiser / data-flow contracts are solid. One non-blocking P2 UX item noted below.

## Verification

- ✅ **Sanitiser parity** — `sanitizeLineHeight` (0 < lh ≤ 10, unitless only) and `sanitizeSpacing` (0 ≤ n ≤ 1000, px|em only) are called at parse, render, command, and DOCX-export time. The regexes (`LINE_HEIGHT_RE`, `SPACING_RE`) are anchored and non-backtracking — no ReDoS risk.
- ✅ **Backend byte-alignment contract** — canonical style property order (`text-align; line-height; margin-top; margin-bottom`) is enforced by extension-load order (LineHeight registered after TextAlign) and pinned by `LineHeight.test.ts` round-trip assertions.
- ✅ **Command set/unset parity (C1)** — all six commands (`setLineHeight`/`unsetLineHeight`, `setSpaceBefore`/`unsetSpaceBefore`, `setSpaceAfter`/`unsetSpaceAfter`) are symmetric. Set commands sanitise before writing; unset commands call `resetAttributes`.
- ✅ **Extension registration (C2)** — `LineHeight` registered after `TextAlign` in both `buildExtensions` and `buildPreviewExtensions`. Ordering is correct for the canonical style serialisation contract.
- ✅ **DOCX export wiring** — `mapSpacing` in `styles.ts` re-runs both sanitisers before converting to twips / 240ths-of-a-line. `convertParagraph` and `convertHeading` in `nodes.ts` spread the spacing object correctly.
- ✅ **Schema version bump** — SCHEMA_VERSION 15 → 17 (v16 reserved for CJK). v17 adds attributes only, no new nodes/marks — `SCHEMA_NODES` and `SCHEMA_MARKS` are unchanged, correctly asserted by `schema/index.test.ts`.
- ✅ **Commit-on-blur input design** — `LineHeightCustomInput` uses a local draft + `seenCommitted` ref to avoid the focus-steal and snap-back issues documented in the component's comments. Tests pin the typing, commit, revert, and seed behaviours.
- ✅ **i18n** — both `en-US.json` and `zh-CN.json` add the same six keys under `docs.toolbar.*`.

Static analysis only at head `ef80faba71ac0b9e82e6f1939be5d6a2049182ff`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 UX item below.

### P2 — Default option label reuses the control title (`packages/docs/src/editor/Toolbar.tsx:692`)

The `ParagraphSpacingSelect` reset option renders `t('docs.toolbar.spaceBefore')` ("Space before") or `t('docs.toolbar.spaceAfter')` ("Space after") as the dropdown value text. This is the same string used for the `<select>` `title` (tooltip). When a user opens the dropdown to clear spacing, the first option reads "Space before" — which looks like a label rather than a "reset to default" action. The sibling `LineHeightSelect` uses a dedicated `lineHeightDefault` key ("Default") for its reset option, which is clearer.

Consider reusing the existing `lineHeightDefault` key or adding a `spacingDefault` key so the reset option reads consistently:

```tsx
<option value="">{t('docs.toolbar.lineHeightDefault')}</option>
```

- Sanitiser reject (not clamp) semantics are consistent across parse/render/command/export and the backend schema contract.
- `mapSpacing` returns `undefined` when nothing applies, so the spread `...(spacing ? { spacing } : {})` correctly omits the property for blocks without spacing.
- `LineHeightCustomInput` render-phase `setDraft` sync is guarded by the `seenCommitted` ref — no infinite loop, and typing is not clobbered.
- The `'custom'` option in `ParagraphSpacingSelect` is a read-only sentinel (onChange returns early), so non-preset pasted values are preserved without being silently reset.
- Both live and preview extension sets include `LineHeight` — historical versions will render line-spacing faithfully.
- The `spacingToTwips` function in `styles.ts` re-runs `sanitizeSpacing` before parsing — redundant but defensive; the second regex match is guaranteed to succeed on the sanitised value.

## Data-flow Trace

- **lineHeight value**: user input → `setLineHeight(value)` command → `sanitizeLineHeight(value)` → `updateAttributes('heading'|'paragraph', { lineHeight: v })` → stored on node attr → `renderHTML` re-sanitises → inline `style="line-height: v"`. On parse: `parseHTML` reads `element.style.lineHeight` → `sanitizeLineHeight` → node attr. DOCX: `mapSpacing` → `sanitizeLineHeight` → `Math.round(Number(lh) * 240)` twips.
- **spaceBefore/spaceAfter value**: same defence-in-depth path as lineHeight, with `sanitizeSpacing` enforcing px|em units and ≤1000 cap. DOCX: `spacingToTwips` → `sanitizeSpacing` → twips conversion (15 twips/px, 16px/em).
- **Custom input draft**: typing → `setDraft` (local state only, no editor focus) → blur/Enter → `commit()` → `editor.chain().focus().setLineHeight(draft).run()` → on success: `seenCommitted` updated; on failure: draft reverts to committed.

## Blind-spot Checklist

- **C1 (dual-path parity)**: set/unset pairs symmetric for all three attrs. Live/preview extension sets both include LineHeight. ✅ Clear.
- **C2 (control-flow ordering)**: LineHeight registered after TextAlign in both extension sets. Commands sanitise then updateAttributes. ✅ Clear.
- **C3–C6**: N/A — no auth, lifecycle, build, or governance changes in this PR.

## Cross-round Blocker Recheck

[Octo-Q] verdict: APPROVE — No P0/P1 findings. The PR adds line-height and paragraph spacing with solid defence-in-depth sanitisation, correct schema versioning, proper DOCX export wiring, and thorough tests. One non-blocking P2 on the paragraph-spacing dropdown's default option label.

Jerry-Xin (changes requested):
## Re-review — PR #522 (octo-server), head `255de781`

Re-reviewed at the new head after the prior review (`e199c059`, COMMENTED, no blockers). The avatar authorization relaxation is correct and now well-tested, **but** the accompanying `groupUpdate` error-ID change introduces a test/code mismatch that will break the group package tests once they run.

### 🔴 Blocking

**Existing regression tests still assert the old `manager_only` error ID for branches that now return `creator_or_manager_only`.**

`groupUpdate` now returns `errcode.ErrGroupCreatorOrManagerOnly` (i18n `err.server.group.creator_or_manager_only`) on the manager-required branch:
- `modules/group/api.go:1184` — the `!isManager` reject inside the advanced-field / non-name branch.

But two regression cases that drive exactly this code path still expect the old ID:
- `modules/group/rename_permission_matrix_test.go:162` — "normal member with advanced fields" → asserts `err.server.group.manager_only`.
- `modules/group/rename_permission_matrix_test.go:204` — "unknown field only → fail-closed manager" → asserts `err.server.group.manager_only`.

Both seed a `MemberRoleCommon` member and expect a 400 carrying `manager_only`, so once the group tests run they will fail against the new `creator_or_manager_only` body.

Fix: if the new error ID is intentional (it is consistent with the `QueryIsGroupManagerOrCreator` gate — owner OR admin — so it is the right ID), update these two assertions (and any adjacent wording) to `err.server.group.creator_or_manager_only`. Otherwise keep `ErrGroupManagerOnly` in `groupUpdate`.

### 🟡 Non-blocking

- `modules/group/api.go:704` — `avatarUpload` still parses the up-to-20MB multipart body **before** the authorization check at `:724`. Pre-existing (flagged in the prior review); an unauthenticated-for-this-action member can make the server buffer the upload before being rejected. Low priority; noting for continuity.
- Draft PR: `Build`/`Test`/`Vet`/`Lint` are SKIPPED on drafts (skipped ≠ green). The blocker above is visible in checked-in source regardless, and will turn CI red the moment the group tests execute.

### ✅ Highlights

- Avatar gate change (`QueryIsGroupCreator` → `QueryIsGroupManagerOrCreator`) at `modules/group/api.go:724` correctly aligns avatar upload with the owner-or-admin semantics used by group name/notice/settings. The gate is exactly scoped: `avatarUpload` only writes the avatar object + version + sends `CMDGroupAvatarUpdate` — no privilege leak into name/notice/ownership/member management.
- Role lookup is IDOR-safe and injection-safe: `QueryIsGroupManagerOrCreator` is fully parameterized on `group_no` + `uid` and enforces `is_external=0 and status=normal and role IN (creator, manager)`.
- Avatar value is a server-side multipart file written to a server-generated path with forced `image/png` content-type — no URL fetch, so no SSRF / path-traversal surface.
- New tests `TestGroupAvatarUpload_AdminSucceeds` and `TestGroupAvatarUpload_NonManagerForbidden` (in `modules/group/api_i18n_regression_test.go`) pin both sides of the new permission boundary.

yujiawei (approved):
# Code Review — PR #522 (octo-server)

**Verdict: APPROVED** — no correctness/security blockers. Two non-blocking suggestions below.

## Summary

The change replaces the owner-only gate on group avatar upload with the shared manager-or-creator gate, so group admins can update the group avatar — bringing avatar upload in line with group name / settings / notice. It also fixes a pre-existing error-code mismatch in `groupUpdate`, and refreshes two stale test comments.

## Spec compliance

- **Requested behavior delivered**: `avatarUpload` now gates on `QueryIsGroupManagerOrCreator` and returns `ErrGroupCreatorOrManagerOnly` instead of `QueryIsGroupCreator` / `ErrGroupCreatorOnly` (`modules/group/api.go:724`, `:731`). This is exactly what #520 asks for — admins, not just the owner, can set the avatar.
- **No over-building**: the only extra change beyond the avatar fix is the `groupUpdate` error-code correction, which is a directly related, self-contained consistency fix.
- **No deviation**: no new error codes or i18n keys were introduced — `ErrGroupCreatorOrManagerOnly` and its localization already existed.

## Security review (security-sensitive PR)

- **No privilege escalation.** The new gate is the same one already used by name/settings/notice for advanced fields. A regular member is still rejected — confirmed by the added `TestGroupAvatarUpload_NonManagerForbidden`.
- **The creator path is actually *tightened*, not loosened.** The old `QueryIsGroupCreator` had no `is_external` / `status` filter; `QueryIsGroupManagerOrCreator` additionally requires `is_external=0 AND status=normal`. So a cross-Space "external" creator (or a non-normal-status one) can no longer upload the avatar. This is consistent with the `is_external=0` authorization boundary enforced across the other group-mutation endpoints and is the correct, intended behavior — a normal internal owner is unaffected. The author documented this explicitly at `modules/group/api.go:722`.
- **The `groupUpdate` fix is correct.** The decision already used `QueryIsGroupManagerOrCreator`, so returning `ErrGroupManagerOnly` was misleading; `ErrGroupCreatorOrManagerOnly` (`modules/group/api.go:1184`) now matches the decision. Pure error-code/messaging correction, no authorization change.

## Tests

- `TestGroupAvatarUpload_AdminSucceeds` / `_NonManagerForbidden` pin both sides of the new boundary; the negative test asserts the aligned `creator_or_manager_only` error ID.
- The wire status `400` asserted by the negative test is correct under the legacy error-envelope convention (transport status pinned to 400, real 403 carried in the body).
- Compiles cleanly and `go vet` passes; the full test run requires MySQL/Redis/IM and cannot execute in this environment, but the code paths are straightforward.

1. **`modules/group/api.go:1180`** — the boolean is still named `isManager` while it holds a manager-*or-creator* result. Renaming to `isManagerOrCreator` (as done elsewhere in this file) would remove the code/intent mismatch and reduce refactor risk. Pre-existing; optional.
2. **Edge-case coverage** — the new HTTP-level tests only exercise the `is_external=0, status=normal` path. The stricter external/abnormal-creator denial that this change introduces at the endpoint is not asserted through the HTTP handler. It *is* covered at the query layer (`TestQueryIsGroupManagerOrCreator_ExternalMemberFailSafe` asserts an external creator resolves to `false`), so this is a minor gap, not a blocker. A follow-up HTTP test for an external creator would fully lock the boundary.

yujiawei (changes requested):
# Code Review — PR #522 (octo-server)

**Verdict: CHANGES_REQUESTED** — this supersedes my earlier approval on this same head. On re-verification against the full group test suite, the `groupUpdate` error-ID change breaks two pre-existing regression tests that are not part of this PR's diff. Build/Test are skipped on this draft (skipped ≠ green), so the breakage is only visible by reading the checked-in source across files.

## 🔴 Blocking (P1) — test/code mismatch will turn the group package red

`groupUpdate` now returns `errcode.ErrGroupCreatorOrManagerOnly` (i18n `err.server.group.creator_or_manager_only`) on the manager-required branch (`modules/group/api.go:1184`). Two existing regression cases drive exactly that branch (they seed a `MemberRoleCommon` member and send advanced / unknown fields) but still assert the **old** `err.server.group.manager_only` ID:

- `modules/group/rename_permission_matrix_test.go:162` — "normal member + advanced field" case.
- `modules/group/rename_permission_matrix_test.go:204` — "unknown field only → fail-closed manager" case.

Because `Contains(body, "err.server.group.manager_only")` does **not** match `err.server.group.creator_or_manager_only` (after `err.server.group.` the body now reads `creator_or_manager_only`, not `manager_only`), both assertions will fail once the group tests execute. Neither file is in this PR's diff, so it is easy to miss from the diff alone.

**Fix (the new ID is correct — keep it):** update the two assertions (and any adjacent wording) in `rename_permission_matrix_test.go` to `err.server.group.creator_or_manager_only`. The new ID is the right one — it matches the `QueryIsGroupManagerOrCreator` (owner OR admin) decision — so the code should stay and the stale tests should follow it.

## Spec compliance (otherwise correct)

- Avatar gate change is exactly scoped to #520: `avatarUpload` gates on `QueryIsGroupManagerOrCreator` and returns `ErrGroupCreatorOrManagerOnly` (`modules/group/api.go:724`, `:731`), aligning avatar upload with group name/notice/settings.
- The `groupUpdate` error-ID correction itself is right; it just needs the dependent tests updated in the same PR.
- No new error codes / i18n keys introduced — `ErrGroupCreatorOrManagerOnly` and its localization already existed.

## Security review (security-sensitive PR)

- **No privilege escalation.** Same gate already used by name/settings/notice advanced fields; regular members still rejected (`TestGroupAvatarUpload_NonManagerForbidden`).
- **Creator path is tightened, not loosened.** The old `QueryIsGroupCreator` had no `is_external`/`status` filter; `QueryIsGroupManagerOrCreator` additionally requires `is_external=0 AND status=normal`. A cross-Space external / non-normal creator can no longer upload the avatar — this is the correct, intended alignment with the `is_external=0` boundary enforced elsewhere (documented at `api.go:722`), and is covered at the query layer by `TestQueryIsGroupManagerOrCreator_ExternalMemberFailSafe`. A normal internal owner is unaffected.
- Lookup is IDOR- and injection-safe (fully parameterized on `group_no`+`uid`); the avatar is a server-side multipart file written to a server-generated path with forced `image/png` — no SSRF / path-traversal surface.

1. `modules/group/api.go:1180` — the boolean is still named `isManager` while it holds a manager-*or-creator* result. Rename to `isManagerOrCreator` (as done elsewhere in this file) to remove the code/intent mismatch.
2. `modules/group/api.go:704` — `avatarUpload` parses the up-to-20MB multipart body before the authorization check at `:724`, so a non-authorized member can make the server buffer the upload before being rejected. Pre-existing; low priority.
3. Consider an HTTP-level test for an external/abnormal creator being denied at `avatarUpload`, to fully lock the new boundary at the handler layer (currently only covered at the query layer).

OctoBoooot (changes requested):
## Correction: #522 @ 255de781 — upgrading COMMENT → REQUEST_CHANGES (broken tests, I missed this)

Jerry-Xin is right, and I was wrong to hold this at COMMENT without catching it. Conceding and upgrading to **Request changes** — there is a real blocker: this PR changes an error ID in `groupUpdate` but leaves sibling tests asserting the old ID, so those tests fail at HEAD.

### Blocker
- `modules/group/api.go:~1183` — the PR changes the denial error from `ErrGroupManagerOnly` → `ErrGroupCreatorOrManagerOnly` in the `hasAdvanced || !hasName` branch (confirmed in the base→head patch). But `modules/group/rename_permission_matrix_test.go` still asserts the **old** ID:
  - `:162` — `require.Contains(w.Body.String(), "err.server.group.manager_only")` in `TestGroupRename_NormalMemberWithAdvancedFields_Denied` (4 subcases: name+notice / name+invite / name+avatar_text / name+avatar_color).
  - `:204` — same assertion in `TestGroupUpdate_EmptyBodyOrUnknownField_FailClosed` → `unknown field only`.

  **Failing input / why it breaks:** the handler now returns body `err.server.group.creator_or_manager_only`; the test asserts the substring `err.server.group.manager_only`. That substring is **not** present in `creator_or_manager_only` (the `err.server.group.` prefix is followed by `creator_or_`, so the assertion string doesn't occur) — I verified this with an actual `strings.Contains` run: it returns `false`. So five test cases assert a substring that no longer appears and will fail. _(confidence: high — verified against the base→head patch and an empirical substring check.)_

  Fix: update those two assertions to `err.server.group.creator_or_manager_only` (matching the code's new denial ID). One-line-per-site.

### Why I missed it (owning the error)
On my prior pass I diffed against my *earlier review head* (`e199c05`), where this branch already had the tiering, so the error-ID change didn't show in that delta — and I praised the matrix test as "mutation-resistant" without running its assertions against the current code. I should have grepped the current-head test for the exact ID the changed handler returns (my own logged lesson). Jerry-Xin's read of the base→head diff caught it; verified true.

### Jerry-Xin's 🟡 (agree, non-blocking)
- `avatarUpload` parses/opens the multipart file (`c.FormFile` + `defer file.Close()`) **before** the `QueryIsGroupManagerOrCreator` authz check, so an unauthorized caller still causes the upload to be read. Real minor resource-waste / mild DoS-amplification nit; reorder authz before file handling. Not a security hole (still 403s, nothing persisted), so non-blocking — but a good cleanup.

### Unchanged from my prior review (still holds)
- The authz widening itself is sound: `ExistMemberActiveInternal` keeps the `is_external=0 / status=normal / is_deleted=0` boundary + robot exclusion; no field-smuggling via the name-only path; the matrix's *logic* covers the right allow/deny axes. The only defect is the stale error-ID assertions above — the tests encode the right intent but assert the pre-change string.

### Verdict
**Request changes** — fix the two `manager_only` → `creator_or_manager_only` assertions in `rename_permission_matrix_test.go` (blocker: they fail at HEAD). Then, since it's still a draft, mark ready so Build/Test actually run green. `needs-human-review` remains for the widened rename permission.

Issue #482, opened 2026-06-26
🔴 [P0] queryMembers user_verification JOIN 在非规范 collation 部署上 Error 1267 → 成员列表 500 (#420 回归)


lml2468:
## 🔴 P0 — `queryMembers` 的 `user_verification` JOIN 在非规范 collation 部署上触发 Error 1267，成员列表接口 500

> 从 #434 拆出独立跟踪（#434 是 #344 的 follow-up 集合，此项为其中已上 `main` 的线上正确性风险，单独提单以免被埋没）。

### 影响

`GET /v1/space/:space_id/members`（`modules/space/db.go` 的 `queryMembers`）在**以 MySQL 8.0 服务器默认 collation（`utf8mb4_0900_ai_ci`）建库**的部署上会抛 `Illegal mix of collations (1267)`，导致成员列表端点 **500**。CI 用规范 `general_ci` 建库，因此扫不到。

### 根因（已代码核实）

- #420（已合入 2026-06-19）给 `queryMembers` 新增了 `LEFT JOIN user_verification uv ON uv.user_id=sm.uid`（`modules/space/db.go:158`）。
- 项目规范 collation 是 `utf8mb4_general_ci`；`modules/base/sql/20260512000001_base_oss_compat_repair.sql` 把一批 OSS 表强制 `CONVERT` 到 `general_ci`——**该清单含 `user_verification`，但不含 `space_member` 也不含 `user`**（已核实：repair.sql 第 65 行清单 + grep 0 命中）。
- `space_member` 建表（`modules/space/sql/20260307000002_space_legacy01.sql:21`）**无显式 `COLLATE`**，继承 DB 默认。于是：
  - DB 默认 = `general_ci` → `space_member` 与 `user_verification` 一致，JOIN 正常（**CI 正是这种情况**）。
  - DB 默认 = `0900_ai_ci` → `space_member` 为 `0900_ai_ci`、`user_verification` 被强制 `general_ci` → `uv.user_id = sm.uid` 比较 **Error 1267 → 500**。
- 原有的 `LEFT JOIN user u ON u.uid=sm.uid` 从未触发，因为 `user` 同样未转换、与 `space_member` 继承同一默认。#420 是第一个从 `space_member` JOIN 到「强制 general_ci」表的查询，打破了对称。
- repair.sql 自己的注释（第 7–9 行）已描述了这一机制，只是遗漏了 `space_member`/`user`。

### 复现 / 确认

```sql
[...]

Issue #434, opened 2026-06-23
Follow-up: display-name fallback remaining paths + user_verification JOIN collation risk (#344)


an9xyz:
## Background

#344 (member list returns empty `name` when `user.name` is blank) was closed by #420, which added the display-name fallback to the **member list** (`GET /v1/space/:space_id/members`) only — `queryMembers` + `MemberDetailModel.DisplayName()`.

Two other PRs implemented the same fix independently and are being closed in favor of #420:
- #359 (the broadest one — also covered search, avatar consistency, a shared `pkg/displayname`, and a cleanup plan)
- #398 (a lighter JOIN-based variant, superseded by #420)

This issue tracks the items #420 did **not** cover, plus a correctness risk #420 shipped, so they aren't lost when #359/#398 close.

## 🔴 P0 — `user_verification` JOIN may mix collations on non-canonical DBs (already on `main`)

#420 added `LEFT JOIN user_verification uv ON uv.user_id = sm.uid` to `queryMembers`.

**Background — the project's canonical collation is `utf8mb4_general_ci`.** CI recreates the `test` DB as `general_ci` before every package, and `modules/base/sql/20260512000001_base_oss_compat_repair.sql` proactively `CONVERT`s a set of OSS tables to `general_ci` to escape the MySQL 8.0 server default (`utf8mb4_0900_ai_ci`).

**The gap:** that repair list covers ~17 OSS tables (`user_verification` is one of them) but **not `space_member` (nor `user`)**. `space_member` (`modules/space/sql/20260307000002_space_legacy01.sql`) has no explicit `COLLATE` and is never converted, so it just inherits the **database default**. So:

- DB created with the canonical `general_ci` default → `space_member` is `general_ci`, matches `user_verification`, JOIN is fine. **This is exactly what CI tests — correctly.**
- DB created with the MySQL 8.0 **server default** `0900_ai_ci` → `space_member` is `0900_ai_ci` while `user_verification` is forced to `general_ci` → the `uv.user_id = sm.uid` comparison raises **`Illegal mix of collations (1267)`** → the member list endpoint **500s**.

The existing `LEFT JOIN user u ON u.uid = sm.uid` never tripped this because `user` is also un-converted and inherits the same DB default as `space_member`, so they always match. #420 introduced the first join from `space_member` to a **force-`general_ci`** table, breaking the symmetry.

**Verify on the target DB:**
```sql
SELECT TABLE_NAME, TABLE_COLLATION
[...]

PR #420, opened 2026-06-18
fix(space): fall back to real_name + stable placeholder for empty member name (#344)

yujiawei:
## Summary

The space member list (`GET /v1/space/:space_id/members`) returned an empty
`name` for members whose `user.name` column is blank, so clients rendered a
blank member row. This adds a privacy-gated display-name fallback.

## Related Issue

Fixes #344

## Linked Spec

`.octospec/tasks/member-list-name-fallback/brief.md` (Goal / Load-bearing /
Out-of-scope / Acceptance). Injected rules recorded in
`.octospec/tasks/member-list-name-fallback/context.yaml`
(space-isolation, error-handling, testing).

## Changes

- `modules/space/db.go` `queryMembers`: add a read-only
  `LEFT JOIN user_verification uv ON uv.user_id=sm.uid` and select
  `IFNULL(uv.real_name,'') as real_name`. The Space + bot-ownership WHERE clause
  (`sm.space_id=? AND sm.status=1 AND (r.robot_id IS NULL OR r.creator_uid=?)`)
  is unchanged.
- `modules/space/model.go`: add `RealName` to `MemberDetailModel` and a
  `DisplayName()` helper resolving the chain `user.name` → `real_name` →
  stable placeholder `"User <uid>"`. Never returns `""`; never reads
  `short_no`/`username` (privacy-gated).
- `modules/space/api.go` `listMembers`: map `m.DisplayName()` into
  `memberResp.name`.
- Tests + octospec journal.

## Testing

```
go test ./modules/space/...   # PASS
go vet ./modules/space/...     # clean
make i18n-lint                 # OK (no new direct error responses / unregistered codes)
make i18n-extract-check        # exit 0
```

New tests in `modules/space/db_member_name_fallback_test.go`:
1. `user.name` empty + `real_name` present → returns `real_name`.
2. both empty → stable placeholder (asserts non-empty, not `short_no`/`username`).
3. populated `user.name` → returned unchanged.

- [x] Unit tests added/updated
- [x] Manually verified (DB-level + pure-function tests)

## COMPREHENSION

1. **What does this change actually do to the load-bearing path?**
   It changes how `memberResp.name` is resolved for the user-facing member list.
   At the data layer it adds one read-only `LEFT JOIN` on `user_verification`
   (one row per `user_id`, so no fan-out) purely to surface `real_name`; the
   Space-scoping and bot creator-ownership predicates in the WHERE clause are
   byte-for-byte the same, so row visibility / tenant isolation is untouched.
[...]

PR #523, opened 2026-07-02
fix(space): apply display-name fallback to member search path #434

lml2468:
# Bug Fix — Issue #434 — PR (draft)

## 1. Executive Summary
- **Bug**: The paginated member **search** path never got #344's display-name fallback — blank-`name` members render empty in search results and cannot be found by their `real_name`.
- **Root Cause**: `modules/space/db.go:searchMembers`/`countSearchMembers` join only `user` (not `user_verification`) and select bare `IFNULL(u.name,'')`; the fallback #420 added lives solely in `queryMembers`.
- **Fix Approach**: localized (space module only)
- **Risk Level**: Low
- **PR**: draft — awaiting review squad

## 2. Issue Context
- Title: Follow-up: display-name fallback remaining paths + user_verification JOIN collation risk (#344)
- Reporter: lml2468 (GH)
- bug-verified by: self (reproduced at DB layer via RED→GREEN regression test)
- Affected: `GET /v1/space/:space_id/members/search` on any deployment with blank-`name` members
- **Scope note**: This PR addresses only the **P1 — search path has no fallback** item. The 🔴 collation-JOIN risk was split into **#482** and is out of scope here; P2 (avatar / other `IFNULL(u.name,'')` endpoints) and P3 (shared `pkg/displayname`) remain follow-ups (see §5 boundary).

## 3. Reproduction
**Environment**: Go 1.26, MySQL 8.0 (`test` DB, `utf8mb4_general_ci`), `modules/space` package tests.
**Steps**:
1. Seed a `space_member` whose `user.name` is empty but has `user_verification.real_name = "Zhang Wei"`.
2. Call `searchMembers(spaceId, "", …)` → result item shows empty display name.
3. Call `searchMembers(spaceId, "Zhang", …)` → member is **not** returned.
**Expected**: display name falls back to `real_name` (then to a stable `User <uid>` placeholder); member is searchable by `real_name`.
**Actual**: empty name; not searchable by `real_name`.
**Evidence** (`Test_Issue434_SearchMemberNameFallback`, before fix):
```
Error: Not equal — "" != "Zhang Wei"   (empty name must fall back to real_name in search)
Error: Should be true                  (member must be searchable by real_name)
--- FAIL: Test_Issue434_SearchMemberNameFallback
```

## 4. RCA
**Method**: 5 Whys
**Analysis**: Search shows empty names → because response maps bare `m.Name` → because `searchMembers` selects only `IFNULL(u.name,'')` and never joins `user_verification` → because #420 (issue #344 fix) only touched the member **list** (`queryMembers` + `MemberDetailModel.DisplayName()`), not the search path added later in #389 → the two read paths diverged.
**Root cause**: `modules/space/db.go:searchMembers`/`countSearchMembers` (+ `api_member_search.go` mapping) omit the `user_verification` fallback that `queryMembers` has.

## 5. Fix Description
**Files changed**:
- `modules/space/db.go` — `searchMembers` + `countSearchMembers`: add `LEFT JOIN user_verification`, select `real_name`.
- `modules/space/db_manager.go` — `memberSearchActiveColumns`: add `uv.real_name` (searchable by real name).
- `modules/space/model.go` — add `memberSearchModel.RealName` + `DisplayName()`; extract shared `resolveMemberDisplayName()`; `MemberDetailModel.DisplayName()` now delegates to it.
- `modules/space/api_member_search.go` — response maps `m.DisplayName()` instead of `m.Name`.
- `modules/space/db_member_search_fallback_test.go` — new regression test.
**Change summary**: The search path now resolves display names through the same fallback chain as the member list (`name → real_name → "User <uid>"`) and lets admins find members by `real_name`. List and count share one JOIN so keyword hits on `real_name` don't drift pagination. The new `user_verification` JOIN pins `COLLATE utf8mb4_general_ci` on the ON clause so the search path does **not** re-introduce the `space_member`↔`user_verification` collation-mix 500 that #482 tracks for `queryMembers` on non-canonical DBs.
**Does NOT change** (scope boundary): `queryMembers`'s existing raw JOIN (owned by #482); avatar rendering / other `IFNULL(u.name,'')` endpoints (P2); no shared `pkg/displayname` package (P3); no privacy widening — `short_no`/`username` are still never used as a fallback, and `phone` search remains last-4-only.

## 6. Regression Tests
|---|---|---|---|---|
| Test_Issue434_SearchMemberNameFallback | `db_member_search_fallback_test.go` | search shows real_name fallback + placeholder; searchable by real_name; count==list | FAIL | PASS |
| TestQueryMembersNameFallback | `db_member_name_fallback_test.go` | existing list fallback unaffected by helper refactor | PASS | PASS |
| TestMemberDisplayNamePure | `db_member_name_fallback_test.go` | pure fallback precedence | PASS | PASS |

## 7. CI Status
- Build: `go build ./modules/space/` pass
- Vet: `go vet ./modules/space/` pass
- gofmt: clean
- Unit (regression, DB layer): RED→GREEN verified in a clean DB window — the three tests above PASS with the fix, and `Test_Issue434_*` FAILs without it.
- Note: the sandbox's shared MySQL was intermittently reset/offline (another workspace runner races the shared `test` DB), so the full HTTP-level search suite could not be run to completion locally. CI drops+recreates `test` per package (`utf8mb4_general_ci`) and is the authoritative gate; the change is behavior-preserving for populated names (`DisplayName()` returns `m.Name` unchanged when non-empty).
- Coverage delta: N/A (not measured — DB env unavailable)

## 8. Deployment Notes
- Migration required?: no
- Config change?: no
- Feature flag: no
- Compatibility: backward-compatible. `memberSearchResp.name` now shows a non-empty display name for blank-`name` members (previously empty); response shape unchanged. Members become findable by `real_name` on the admin-scoped (role ≥ admin) search endpoint.

## 9. Rollback Plan
**Pattern**: revert
**Steps**: 1. `git revert` the two commits on this branch. 2. Redeploy. No data/migration changes to unwind.
**Detection** (trigger rollback): spike in 5xx on `GET /v1/space/:space_id/members/search`, or `Illegal mix of collations (1267)` in logs for that endpoint.

---

### DIFF
--- a/modules/space/api_member_search.go
+++ b/modules/space/api_member_search.go
@@ -69,8 +69,10 @@ func (s *Space) searchMembers(c *wkhttp.Context) {
 	resps := make([]memberSearchResp, 0, len(members))
 	for _, m := range members {
 		resps = append(resps, memberSearchResp{
-			UID:       m.UID,
-			Name:      m.Name,
+			UID: m.UID,
+			// name 兜底链（issue #344 / #434）：空 user.name 回退 real_name，
+			// 再回退稳定占位符，与成员列表口径一致，避免搜索结果展示空名。
+			Name:      m.DisplayName(),
 			Username:  m.Username,
 			Email:     m.Email,
 			Phone:     maskPhone(m.Phone),
--- a/modules/space/db.go
+++ b/modules/space/db.go
@@ -168,15 +168,23 @@ func (d *DB) queryMembers(spaceId string, loginUID string, page uint64, limit ui
 }
 
 func (d *DB) searchMembers(spaceId, keyword string, pageIndex, pageSize int) ([]*memberSearchModel, error) {
+	// name 兜底链（issue #344 / #434）：与 queryMembers 对齐，u.name 为空时回退
+	// user_verification.real_name，映射层（memberSearchModel.DisplayName）再给稳定
+	// 占位符。uv JOIN 显式 COLLATE utf8mb4_general_ci：user_verification 被 compat-repair
+	// 强制转成 general_ci，而 space_member 只继承库默认；非规范 collation 部署下二者
+	// 直接比较会触发 Illegal mix of collations (1267)（见 #482），这里固定 general_ci
+	// 保证搜索路径不引入同一 500 风险。
 	builder := d.session.Select(
 		"sm.*",
 		"IFNULL(u.name,'') as name",
+		"IFNULL(uv.real_name,'') as real_name",
 		"IFNULL(u.username,'') as username",
 		"IFNULL(u.email,'') as email",
 		"IFNULL(u.phone,'') as phone",
 		"CASE WHEN r.robot_id IS NOT NULL AND r.status=1 THEN 1 ELSE 0 END as robot",
 	).From(dbr.I("space_member").As("sm")).
 		LeftJoin(dbr.I("user").As("u"), "u.uid=sm.uid").
+		LeftJoin(dbr.I("user_verification").As("uv"), "uv.user_id=sm.uid COLLATE utf8mb4_general_ci").
 		LeftJoin(dbr.I("robot").As("r"), "r.robot_id=sm.uid").
 		Where("sm.space_id=? AND sm.status=1", spaceId)
 	if keyword != "" {
@@ -195,9 +203,13 @@ func (d *DB) searchMembers(spaceId, keyword string, pageIndex, pageSize int) ([]
 }
 
 func (d *DB) countSearchMembers(spaceId, keyword string) (int64, error) {
+	// list / count 共用 memberSearchActiveWhere（含 uv.real_name），故 count 侧也要挂
+	// 同一 uv JOIN，否则关键字命中 real_name 时两侧口径漂移导致分页错位 / 计数不符。
+	// COLLATE 理由同 searchMembers。
 	builder := d.session.Select("COUNT(*)").
 		From(dbr.I("space_member").As("sm")).
 		LeftJoin(dbr.I("user").As("u"), "u.uid=sm.uid").
+		LeftJoin(dbr.I("user_verification").As("uv"), "uv.user_id=sm.uid COLLATE utf8mb4_general_ci").
 		Where("sm.space_id=? AND sm.status=1", spaceId)
 	if keyword != "" {
 		clause, args := memberSearchActiveWhere(keyword)
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -54,7 +54,11 @@ func memberSearchWhere(keyword string) (string, []interface{}) {
 //     （响应仅显示 138****5678），admin 无法通过子串查询逐位探测/重建完整号码。
 //
 // 前端注意：phone 检索只匹配后 4 位，传完整号码不会命中——按手机号查找请用后 4 位。
-var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
+//
+// uv.real_name（issue #434）：空 user.name 的成员以 real_name 作展示名，故也让其
+// 可按 real_name 检索，保持「可检索粒度 == 可见粒度」；调用方（searchMembers /
+// countSearchMembers）必须已挂 user_verification 的 LEFT JOIN。
+var memberSearchActiveColumns = []string{"u.name", "uv.real_name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
 
 // memberSearchActiveWhere 为空间侧 members/search 组装跨列 OR LIKE 条件。
 // list / count 共用同一条件，避免搜索范围漂移导致分页错位。
--- a/modules/space/db_member_search_fallback_test.go
+++ b/modules/space/db_member_search_fallback_test.go
@@ -1,0 +1,69 @@
+package space
+
+import (
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/assert"
+)
+
+func findSearchModel(list []*memberSearchModel, uid string) (*memberSearchModel, bool) {
+	for _, m := range list {
+		if m.UID == uid {
+			return m, true
+		}
+	}
+	return nil, false
+}
+
+// Test_Issue434_SearchMemberNameFallback 验证 issue #434（#344 的 follow-up）：
+// 成员搜索路径（searchMembers / countSearchMembers）此前未做 #420 给成员列表
+// 加的 name 兜底——空 user.name 的成员在搜索结果里展示空名，且无法按
+// user_verification.real_name 检索。本测试锁定与成员列表一致的行为：
+//  1. name 空 + real_name 有 → 搜索结果 DisplayName() 回退到 real_name；
+//  2. name 空 + real_name 空 → 稳定占位符（非空、不含隐私字段）；
+//  3. 可按 real_name 检索到 blank-name 成员，且 count 与 list 口径一致。
+func Test_Issue434_SearchMemberNameFallback(t *testing.T) {
+	_, _, err := setup(t)
+	assert.NoError(t, err)
+
+	const spaceId = "sp-search-namefallback"
+	seedMemberSearchSpace(t, spaceId, testutil.UID)
+
+	// 1) name 空 + real_name 有
+	const uidRealName = "u-search-realname"
+	seedFallbackUser(t, uidRealName, "")
+	seedFallbackVerification(t, uidRealName, "Zhang Wei")
+	seedMemberSearchMember(t, spaceId, uidRealName, 0, 1)
+
+	// 2) name 空 + real_name 空（无 verification 行 → LEFT JOIN 给空串）
+	const uidPlaceholder = "u-search-placeholder"
+	seedFallbackUser(t, uidPlaceholder, "")
+	seedMemberSearchMember(t, spaceId, uidPlaceholder, 0, 1)
+
+	// 空 keyword：全部成员，real_name 兜底生效
+	all, err := testSpaceDB.searchMembers(spaceId, "", 1, 100)
+	assert.NoError(t, err)
+	if m, ok := findSearchModel(all, uidRealName); assert.True(t, ok, "real_name member missing in search") {
+		assert.Equal(t, "", m.Name, "user.name should be empty")
+		assert.Equal(t, "Zhang Wei", m.RealName)
+		assert.Equal(t, "Zhang Wei", m.DisplayName(), "empty name must fall back to real_name in search")
+	}
+	if m, ok := findSearchModel(all, uidPlaceholder); assert.True(t, ok, "placeholder member missing in search") {
+		assert.Equal(t, "", m.Name)
+		assert.Equal(t, "", m.RealName)
+		assert.Equal(t, memberDisplayNamePlaceholderPrefix+uidPlaceholder, m.DisplayName(),
+			"empty name+real_name must fall back to stable placeholder")
+	}
+
+	// 按 real_name 检索：应命中 blank-name 成员（此前搜不到 → 本 issue 的核心缺陷）
+	byReal, err := testSpaceDB.searchMembers(spaceId, "Zhang", 1, 100)
+	assert.NoError(t, err)
+	_, ok := findSearchModel(byReal, uidRealName)
+	assert.True(t, ok, "member must be searchable by real_name")
+
+	// count 与 list 口径一致（同一 keyword 两侧都要含 uv JOIN，否则分页错位）
+	cnt, err := testSpaceDB.countSearchMembers(spaceId, "Zhang")
+	assert.NoError(t, err)
+	assert.Equal(t, int64(1), cnt, "count must match real_name search list")
+}
--- a/modules/space/model.go
+++ b/modules/space/model.go
@@ -44,10 +44,17 @@ type MemberModel struct {
 type memberSearchModel struct {
 	MemberModel
 	Name     string
+	RealName string // 实名兜底（user_verification.real_name），仅作 name 为空时的回退源
 	Username string
 	Email    string
 	Phone    string
 	Robot    int
+}
+
+// DisplayName 返回搜索结果里成员的展示名，复用与成员列表（MemberDetailModel）
+// 相同的兜底链（issue #344 / #434），避免两条路径对空名成员的展示口径漂移。
+func (m *memberSearchModel) DisplayName() string {
+	return resolveMemberDisplayName(m.Name, m.RealName, m.UID)
 }
 
 // InvitationModel 邀请表模型
@@ -265,6 +272,24 @@ type MemberDetailModel struct {
 // 用它兜底不泄露新信息——但禁止用 short_no / username 兜底（privacy-gated）。
 const memberDisplayNamePlaceholderPrefix = "User "
 
+// resolveMemberDisplayName 是成员展示名兜底链的单一实现（issue #344 / #434）：
+//  1. name 非空 → 原样返回；
+//  2. 否则 realName 非空 → 返回实名；
+//  3. 都空 → 返回稳定占位符 "User <uid>"。
+//
+// 任何分支都不会返回空串，也不会暴露 short_no / username（privacy-gated）。
+// 成员列表（MemberDetailModel）与成员搜索（memberSearchModel）共用它，避免两条
+// 路径的兜底口径与占位格式漂移。
+func resolveMemberDisplayName(name, realName, uid string) string {
+	if name != "" {
+		return name
+	}
+	if realName != "" {
+		return realName
+	}
+	return memberDisplayNamePlaceholderPrefix + uid
+}
+
 // DisplayName 返回成员的展示名，按 issue #344 的兜底链解析：
 //  1. user.name 非空 → 原样返回；
 //  2. 否则 user_verification.real_name 非空 → 返回实名；
@@ -272,13 +297,7 @@ const memberDisplayNamePlaceholderPrefix = "User "
 //
 // 任何分支都不会返回空串，也不会暴露 short_no / username。
 func (m *MemberDetailModel) DisplayName() string {
-	if m.Name != "" {
-		return m.Name
-	}
-	if m.RealName != "" {
-		return m.RealName
-	}
-	return memberDisplayNamePlaceholderPrefix + m.UID
+	return resolveMemberDisplayName(m.Name, m.RealName, m.UID)
 }
 
 // SpaceDetailModel 带成员数和角色的空间详情（MaxUsers 从 SpaceModel 继承）


OctoBoooot (commented):
## Review: fix(space): apply display-name fallback to member search path (#523)

**Verdict:** Comment — draft PR; no blockers or majors, clean fix with good tests.
Extends #344's display-name fallback to the member-search path and consolidates the fallback into one shared function. Draft, so no APPROVE regardless (CI Build/Test SKIPPED on the draft — mark ready-for-review to run the new test). One privacy question worth an explicit answer, then it's good.

Risk tier: normal. I checked it on the privacy axis (real_name is verification data) since that's what a display-name fallback touches.

### What I byte-verified
- **No privacy leak in the response.** `resolveMemberDisplayName` (model.go) resolves `name → real_name → "User "+uid` and never returns short_no/username — matching the documented privacy-gated rule. `memberSearchResp` (model.go) has **no** `real_name` field; `real_name` only flows into `Name` via `DisplayName()` when `name` is empty. So real_name is never exposed as its own field — correct, and consistent with the member-list path.
- **Endpoint is admin-gated.** `api_member_search.go:41` — caller must be a space member AND `member.Role >= spaceRoleAdmin`, else `ErrSpaceNotMember` / `ErrSpacePermissionDenied`. So everything below is scoped to an in-space admin who already sees username/email/phone-last-4.
- **list/count parity preserved.** Both `searchMembers` and `countSearchMembers` add the same `uv` LEFT JOIN and share `memberSearchActiveWhere` — no pagination/count drift when a keyword hits real_name. The author explicitly called this out and got it right.
- **COLLATE pin correct.** The new `uv` join uses `COLLATE utf8mb4_general_ci` (the #482 fix), so this path doesn't introduce the Error 1267 risk. (Note: like #489, only the `uv` join is pinned, not `user`/`robot` — see cross-PR note below.)
- **DRY consolidation is behavior-preserving.** `MemberDetailModel.DisplayName()` and the new `memberSearchModel.DisplayName()` both delegate to `resolveMemberDisplayName`, so the two paths can't drift. The extraction is a pure refactor of the existing list logic.

### Question (privacy asymmetry — answerable, not a blocker)
- `db_manager.go:61` adds `uv.real_name` to `memberSearchActiveColumns`, so a member is **searchable** by real_name regardless of whether their `name` is empty — but real_name is only **displayed** when `name` is empty. For a member *with* a display name, an admin can type a guessed real_name substring and get a hit, confirming that real_name without ever seeing it in the response. The author's stated invariant is "可检索粒度 == 可见粒度" (searchable == visible); this slightly exceeds it (searchable-for-all, visible-for-blank-name-only). It's bounded (admin-only, same surface that already searches username/email), so likely acceptable — but it does deviate from the stated principle. Intended? If the invariant is meant to hold strictly, real_name should be searchable only for the members whose display actually falls back to it.

### Praise
- Extracting `resolveMemberDisplayName` as the single source for both list and search — rather than copy-pasting the three-branch fallback — is exactly right: it makes "the two paths can't drift" a structural guarantee, not a review-time hope. And proactively pinning the new `uv` join's COLLATE (citing #482) shows the author carried the collation lesson forward into new code rather than repeating the bug.

### Cross-PR note (informational)
- This adds a `uv` join to `searchMembers`/`countSearchMembers` — the same functions #489 modifies to pin COLLATE on the `user`/`robot` joins. If both land, expect a merge conflict in `searchMembers`; whoever merges second should ensure all of `user`/`uv`/`robot` end up pinned. Not a defect in either PR.

Jerry-Xin (changes requested):
This PR is in scope for octo-server and fixes the right member-search path, but it currently widens searchability of `real_name` beyond what the response exposes.

**🔴 Blocking**
- 🔴 Critical: `uv.real_name` is searchable for members whose `user.name` is already populated, even though that `real_name` is not displayed by `DisplayName()`.
  In [modules/space/db_manager.go](/Users/mlamp/.openclaw/tmp/pr-review-mj1llym0/repo/modules/space/db_manager.go:61), `memberSearchActiveColumns` adds `uv.real_name` unconditionally. But [modules/space/model.go](/Users/mlamp/.openclaw/tmp/pr-review-mj1llym0/repo/modules/space/model.go:283) returns `name` first and only falls back to `realName` when `name == ""`. This means an admin can search for a member by a legal/verified name that the endpoint does not otherwise show when the member has a non-empty display name. That violates the stated “searchable granularity == visible granularity” boundary and is a privacy-relevant behavior change.
  
  The predicate should only match `real_name` when it would actually be the displayed fallback, for example by using a conditional expression such as `CASE WHEN IFNULL(u.name,'') = '' THEN uv.real_name ELSE '' END LIKE ?`, or an equivalent guarded clause. Please add a regression test with `user.name = "Alice"` and `real_name = "Zhang Wei"` asserting that searching `"Zhang"` does not return that member unless the endpoint intentionally exposes that real name.

**💬 Non-blocking**
- 🟡 Warning: I could not run the targeted space tests because the local test database migration state is invalid: `unknown migration in database` for `20191106000001_event_legacy01.sql`. This looks environmental, but CI should still run the new regression coverage.

**✅ Highlights**
- The join/count/list paths are kept aligned, which avoids pagination drift for fallback-name searches.
- Reusing a shared display-name resolver keeps the search and list rendering behavior consistent for blank `user.name` members.
- The new test covers the original issue: blank `name`, present `real_name`, placeholder fallback, and count/list agreement.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #523 (octo-server)

Reviewed at head `0ca70cba7a4720fcf3957f499b53447735643b9f` on branch
`fix/space-434-search-name-fallback`. The merge-base with `main` is older
(`1592d353`), so `pr-base...HEAD` spans many already-merged PRs; the scope of
this review is the two commits authored for this PR — `38872b9f` (repro test)
and `0ca70cba` (fix).

## Summary

This PR ports the member display-name fallback chain that PR #420 added to the
member-list path (`queryMembers`) over to the paginated member-search path
(`searchMembers` / `countSearchMembers`). Previously, blank-`user.name`
members rendered an empty `Name` in search results and could not be located by
their `user_verification.real_name`. The fix adds a `LEFT JOIN
user_verification`, selects `real_name`, and routes the response through a new
shared helper `resolveMemberDisplayName(name, realName, uid)` — the same
`name → real_name → "User <uid>"` chain the list path already uses. The
`memberSearchActiveColumns` set is extended with `uv.real_name` so visible
granularity matches searchable granularity. A regression test locks both the
display-name fallback and the list/count parity on a `real_name` keyword.

The change is minimal, well-scoped, and correctly mirrors the existing list
contract. No correctness, security, or build concerns identified.

## Verification

- ✅ **Diff scope** — two focused commits on top of `main`: `38872b9f` adds the
  failing repro and the model-side `RealName`/`DisplayName()`; `0ca70cba`
  wires the query layer and extracts the shared fallback helper. Nothing else
  in the diff is claimed to belong to this PR.
- ✅ **Authz boundary preserved** — `modules/space/api_member_search.go:36-46`
  still requires `member.Role >= spaceRoleAdmin` after the `checkSpaceActive`
  gate, matching the list endpoint's audience. No new caller reaches the
  touched queries.
- ✅ **Data-flow trace** — `uv.real_name` is coalesced at
  `modules/space/db.go:178` (`IFNULL(uv.real_name,'') as real_name`), read
  into `memberSearchModel.RealName` (`model.go:47`), and consumed only through
  `DisplayName()` (`model.go:52-54`), which the response builder invokes at
  `api_member_search.go:73`. No bare field access. The shared
  `resolveMemberDisplayName` (`model.go:275-290`) is the sole implementation,
  called from both `MemberDetailModel.DisplayName()` (`model.go:300`) and
  `memberSearchModel.DisplayName()` — no forked logic.
- ✅ **List/count parity** — both `searchMembers` and `countSearchMembers`
  hang the same `LEFT JOIN user_verification` and invoke the same
  `memberSearchActiveWhere` (`db.go:191`, `:215`), so a `real_name`-matching
  keyword yields identical row sets on both sides. `memberSearchActiveWhere`
  has exactly two call sites in the repo (`db.go:191`, `:215`); grep confirms
  no other caller could miss the uv JOIN.
- ✅ **LIKE injection hygiene** — `memberSearchActiveWhere`
  (`db_manager.go:65-75`) runs the keyword through `buildLikePattern` /
  `escapeLike` (`\` / `%` / `_` all escaped) and pairs every `LIKE ?` with
  the explicit `ESCAPE '\\'` clause, preserving safety under
  `sql_mode=NO_BACKSLASH_ESCAPES`. Adding `uv.real_name` to the column set
  does not bypass this pipeline.
- ✅ **Collation hygiene** — the uv JOIN pins
  `COLLATE utf8mb4_general_ci` (`db.go:187`, `:212`), matching the
  `user_verification` table DDL (`api_test.go:65`). This preempts the
  Illegal-mix-of-collations 500 that issue #482 tracks for the list path on
  non-canonical DBs.
- ✅ **Privacy contract preserved** — `maskPhone` (`api_member_search.go:90`)
  still reduces phone to last-4 digits, and the search column continues to
  match only `RIGHT(u.phone,4)`. The `real_name` exposure in the response
  matches what the member list (issue #344/#420) already returns under the
  same admin-only gate; no new PII is leaked to a broader audience.
- ✅ **Regression coverage** — `db_member_search_fallback_test.go`
  (`Test_Issue434_SearchMemberNameFallback`) covers the three concrete
  contracts: empty-name-with-real_name falls back to real_name;
  empty-name-with-no-verification falls back to the stable `User <uid>`
  placeholder; and a `real_name` keyword is findable with list and count
  agreeing. Helper fixtures (`seedFallbackUser`, `seedFallbackVerification`,
  `seedMemberSearchSpace`, `seedMemberSearchMember`) are shared with the
  existing list-path fallback test, so the two paths are locked to the same
  semantics.

## Findings

Two minor observations below that do not rise to the level of blocking
feedback.

### Nit — Collation asymmetry with the list path (`modules/space/db.go:187` vs `:158`)

`queryMembers` issues the uv JOIN as raw SQL with no `COLLATE` clause, while
`searchMembers` / `countSearchMembers` pin `COLLATE utf8mb4_general_ci`. The
PR comment at `db.go:173-177` explicitly attributes this to avoiding the
same #482 failure mode the list path carries. Not a defect in this PR —
just flagging that once #482 lands, the list path should likely be
retrofitted with the same explicit `COLLATE` to keep the two joins
textually identical. Out of scope here.

### Nit — Placeholder string not re-asserted in the new test (`modules/space/db_member_search_fallback_test.go:53-57`)

The placeholder branch asserts the exact `memberDisplayNamePlaceholderPrefix
+ uid` shape. Good. The parallel `db_member_name_fallback_test.go` for the
list path asserts the same shape, so drift across paths would still be
caught at the helper level via `resolveMemberDisplayName`. No action
needed; just noting the test surface is appropriately shared.

## Blind-spot checklist (R5)

- **C1 — symmetric path parity**: list vs search parity is the explicit
  goal; `resolveMemberDisplayName` is the single shared impl, and the new
  test pins both paths to the same fixtures. **Clear**.
- **C2 — control-flow ordering / nested reuse**: `memberSearchActiveWhere`
  has exactly two call sites; both already mount the uv JOIN before invoking
  it. No nested or reordered reuse. **Clear**.
- **C3 — authz boundary**: endpoint gate (`Role >= 1`) and `checkSpaceActive`
  are untouched. The new column does not widen who can query. **Clear**.
- **C4 — authz lifecycle / container-member cascade**: `checkSpaceActive`
  (`api_member_search.go:25`) already fails-closed on disbanded/banned
  spaces before any JOIN runs; the change does not touch this. **Clear**.
- **C5 — build/runtime ≠ static**: change is a SQL projection + pure helper;
  no build/packaging/runtime divergence surface. **N/A**.
- **C6 — governance/doc self-consistency**: no doc/policy change. **N/A**.

## Data-flow trace (per Iron Rule)

- **`m.Name` consumed at `api_member_search.go:73`** → populated from
  `IFNULL(u.name,'')` at `db.go:177` via `sm.*` + explicit name column →
  `LEFT JOIN user u ON u.uid=sm.uid` (`db.go:186`) → reaches response as
  the first input to `DisplayName()`.
- **`m.RealName` consumed at `model.go:52-54`** → populated from
  `IFNULL(uv.real_name,'')` at `db.go:178` → `LEFT JOIN user_verification uv
  ON uv.user_id=sm.uid COLLATE utf8mb4_general_ci` (`db.go:187`) → used
  only as the second (fallback) input to `DisplayName()`, gated on
  `m.Name == ""`.
- **`m.UID` consumed as placeholder input at `model.go:289`** → populated
  from `sm.uid` via `sm.*` → used only as the third (fallback-of-fallback)
  input, gated on both `Name` and `RealName` being empty.
- **`uv.real_name` consumed in the WHERE clause at `db.go:191` / `:215`** →
  `memberSearchActiveWhere` (`db_manager.go:65-75`) emits
  `uv.real_name LIKE ? ESCAPE '\\'` with the keyword run through
  `escapeLike`. Both consumers (`searchMembers`, `countSearchMembers`)
  attach the same uv JOIN, so the `uv.`-qualified column reference always
  resolves.

In every case the consumed value can be shown to reach the consumer on the
relevant path — no empty-array / undefined / early-return / gate-shortcut
trap is introduced by the diff.

## Cross-round blocker re-check (R6)

N/A — first review pass on this PR.

## Verdict: APPROVED

No P0/P1. The two nits are optional observations. The fix is small,
symmetric with the existing list path, preserves the admin-only authorization
boundary, and carries a focused regression test that pins both the
display-name fallback and the list/count parity on a `real_name` keyword.
Safe to land.

[Octo-Q] verdict: APPROVE — no P0/P1; the fix correctly mirrors the existing
member-list fallback (#420) onto the search path, preserves the admin-only
authz gate, pins COLLATE on the new uv JOIN to avoid the #482 collation-mix
500 on non-canonical DBs, and ships a regression test that locks the
display-name fallback and list/count parity on a `real_name` keyword. Two
nits (list-path COLLATE retrofit once #482 lands; shared placeholder
assertion already covered via the common helper) are non-blocking.

OctoBoooot (changes requested):
## Verdict correction → REQUEST_CHANGES: #523 @ 0ca70cba — conceding I under-graded the real_name search asymmetry

Steve / Jerry-Xin graded a **blocker** what I posted as a **question**. Re-examining honestly: they're right and I under-weighted it. Upgrading my verdict to REQUEST_CHANGES.

### Why my "consistent with searching username/email" reasoning was wrong
I byte-verified `db_manager.go:61`: `uv.real_name` is OR'd into `memberSearchActiveWhere` **unconditionally** (no `name=''` guard). My original defense was that the endpoint already searches `username`/`email`, so adding `real_name` is consistent. That misses the load-bearing distinction:
- `username`, `email`, `phone-last-4` are all **returned in the response** (`memberSearchResp.Username/Email/Phone`) — searchable == visible holds for them.
- `real_name` is **never** a response field. It lives in a separate `user_verification` table *because* it's privacy-gated legal-identity PII, and it only surfaces as a display fallback when `name` is empty.

So `real_name` is the **only** searchable-but-not-visible column. For a member with a display name ("Alice", real_name "张伟"), an in-space admin can substring-probe (`keyword=张`) and confirm the hidden legal name without it ever appearing in any response — a confirm-oracle that defeats the deliberate display-gating and violates the PR's own invariant ("可检索粒度 == 可见粒度"). Admin-scope bounds the blast radius but doesn't excuse exposing PII the system intentionally hides.

### Fix (matches Steve's)
Make real_name match only when it would actually be the displayed value:
`CASE WHEN IFNULL(u.name,'')='' THEN uv.real_name ELSE '' END LIKE ?`
and add a regression test: member with name="Alice" / real_name="张伟" → searching "张" must **not** match. That restores searchable == visible exactly.

### Unchanged from my prior review (still valid)
- Shared `resolveMemberDisplayName` (no list/search drift), batch `uv` LEFT JOIN (no N+1), correct COLLATE pin on the new join — all good.
- CI Test/Vet/Lint are SKIPPED on the draft, so the regression test's red→green is author-asserted only; mark ready-for-review to actually run it.

### Correction to myself
I flagged the exact asymmetry but mis-severitized it as a non-blocking question by wrongly equating real_name with the already-visible username/email fields. On a privacy-gated PII field, searchable-without-visible is a real exposure, not a design-preference question. Credit to Steve / Jerry-Xin for the correct grade.

yujiawei (changes requested):
# Code Review — PR #516 (octo-web)

**Head SHA reviewed:** `08a27165735078323cfa5d81cb27248a039ba6c0`
**Base / merge-base:** `origin/main` @ `05051d3c` (immediate parent — single-commit PR, no sibling-PR merge conflict)
**Scope:** `sortMessages` rework in `packages/dmworkbase/src/Components/Conversation/vm.ts` + tests in `messageOrder.test.ts`

---

## 1. Spec compliance

The stated goal — reinsert a **settled (`Normal`) no-seq** message at its chronological slot instead of sinking it, while keeping the seq'd backbone byte-for-byte unchanged and in-flight (`Wait`/`Fail`) bubbles pinned to the bottom — is implemented as described.

- **Nothing under-built:** all three groups (backbone / in-flight / floating) are handled; the empty-`floating` fast path preserves the original `messages.sort(compareByOrder)` behavior exactly.
- **Nothing over-built:** the change is scoped to the ordering logic; no unrelated flags/fields/APIs added.
- **No deviation from the approach:** the seq'd backbone ordering (the #242 server-sequenced semantics) is untouched; only messages the old code routed to the `PendingMessageOrderBase` branch (`order` `0`/`NaN`) are affected.

The three-way partition is also **exhaustive**: `MessageStatus` in the SDK is `{Wait, Normal, Fail}`, so "no-seq and not Wait/Fail" is precisely "no-seq Normal" — the `floating` bucket. Verified.

**Spec: ✅**

## 2. Code quality

Test coverage is good and the new tests fail on the old code / pass on the new. However, the new merge path introduces one **blocking regression** and one latent crash vector.

### P1 — Older-history pagination breaks for exactly the messages this PR targets

`vm.ts:1873-1884` (merge) interacting with `getMessageMin()` `vm.ts:2092-2097` and `pulldownMessages()` `vm.ts:2004-2012`.

- `getMessageMin()` does **not** compute a seq-minimum — it returns `this.messagesOfOrigin[0]` literally (index 0).
- `pulldownMessages()` bails early when `minMessage == null || minMessage.messageSeq <= 0`.
- **Before this PR**, every no-seq message sorted to the *bottom* (`PendingMessageOrderBase`), so index 0 was always a real seq'd message and `pulldownMessages` always had a valid `messageSeq` to page from.
- **After this PR**, when a settled no-seq message has a timestamp older than every loaded backbone message, the merge loop places it at **index 0** (`floating[fi].timestamp <= bts` fires before the first backbone push). `getMessageMin()` then returns a `messageSeq === 0` item, and `pulldownMessages()` returns immediately — **the user can no longer pull older server history**.

The trap: the trigger condition (an old, settled, no-seq message — a lost-ack / reconnect-backfill file that predates the loaded window) is *exactly the population #275 is fixing*. So the fix's success case is what arms the regression. Confirmed by simulating the merge:

```
backbone: seq1(ts200), seq2(ts300);  floating: staleOld(seq0, ts100)
merged  : staleOld -> seq1 -> seq2
index0.messageSeq = 0  → getMessageMin() returns it → pulldownMessages() bails (messageSeq <= 0)
```

**Suggested direction:** make `getMessageMin()` return the first message with `messageSeq > 0` (skip no-seq floaters) — or have `pulldownMessages()` scan for the min real seq rather than trusting index 0. Either keeps the new ordering while restoring pagination. Please add a regression test for "settled no-seq message older than the loaded window → pulldown still fetches older history".

### P2 — Equal-timestamp floating jumps ahead of same-second backbone

`vm.ts:1875`. The merge uses `(floating[fi].timestamp || 0) <= bts`, so a floating message with the **same** timestamp as a backbone message is flushed *before* that sequenced message. This contradicts the code comment ("insert after the last backbone with timestamp <= it") and lets an unsequenced message out-rank a server-sequenced message from the same second. Timestamps here are second-granularity, so same-second collisions are plausible. Prefer `<` (strictly-less) so equal timestamps keep the sequenced backbone first, matching the documented intent.

### P2 — Greedy merge assumes backbone is monotonic in timestamp

`vm.ts:1873-1880`. The single-pass merge assumes `backbone` is sorted ascending by timestamp, but `backbone` is sorted by **seq/order**, not timestamp. Server persist-time seq can diverge from client timestamp (the PR body itself flags this seq-vs-time case as out of scope). On a local timestamp inversion (e.g. backbone timestamps `[500, 100]`), a floating message at ts `150` is greedily inserted before the `500` message. Low-frequency and it degrades gracefully to "slightly off" rather than data loss, but worth a comment acknowledging the assumption, or a guard.

### P2 (latent) — `splice(0, len, ...merged, ...inFlight)` can throw at extreme scale

`vm.ts:1887`. The in-place write-back spreads the full merged array into `splice` arguments. V8 throws `RangeError: Maximum call stack size exceeded` once the spread exceeds ~130k arguments (verified locally). The old path was `messages.sort()` with no spread, so this is a newly introduced ceiling. No message-window cap exists in this VM, so a very long-lived conversation scrolled far back could in principle hit it and crash the chat view. Extreme-scale and unlikely in normal use, but since the write-back is new, prefer a spread-free rewrite (assign the merged array, or clear-and-push in a loop) to remove the ceiling entirely.

## 3. Overall verdict

**CHANGES_REQUESTED** — Spec ✅, but the P1 older-history pagination regression is triggered by precisely the message class this PR reinserts, and blocks a core feature (loading older history). The two ordering P2s and the splice-spread P2 are smaller but worth folding into the same revision.

## 4. Suggested changes

1. **(P1, required)** Fix `getMessageMin()` / `pulldownMessages()` to key off the first message with `messageSeq > 0`, not index 0 — plus a regression test for the "old settled no-seq message → pull older history still works" case.
2. **(P2)** Change the merge condition to `<` so equal-timestamp floaters sort after same-second backbone (aligns with the comment).
3. **(P2)** Document/guard the "backbone monotonic in timestamp" assumption in the greedy merge.
4. **(P2)** Replace the `splice(0, len, ...merged, ...inFlight)` spread with a spread-free write-back.

## 5. Additional notes (security-sensitive PR)

This PR was routed `security_sensitive` on keyword match; the actual diff is pure client-side ordering logic with no auth/token/credential/data-boundary surface — no security finding. For manual human verification, please confirm against a **live delayed-delivery repro** (as the author requests) that (a) the sunk messages are genuinely no-seq `Normal`, and (b) after applying the P1 fix, pull-to-load-older still works once a reinserted old message sits at the top of the window.

PR #519, opened 2026-07-02
fix(message): DM per-Space isolation + recent-list cross-Space fixes (#484)

an9xyz:
## Summary

Consolidates **all of #484's DM/conversation per-Space isolation work** onto one
branch. Two groups of fixes, all server-side, no client change:

**A. Base #484 (DM isolation) — authoritative `dm_space_presence` index**
1. **Symptom 2 (DMs mutually hiding between Spaces):** conversation visibility
   was decided by scanning the single shared Recents window, so the most-recently
   active Space won and the other hid the DM. Now OR-ed with a durable, window-
   independent `dm_space_presence` index written at the WuKongIM message webhook.
2. **Symptom 1 (untagged DM history leaking into every Space):** `/v1/message/channel/sync`
   now keeps an untagged DM message only in the user's default Space.

**B. Recent-list cross-Space paths that reproduce with a well-formed request**
3. **Default-Space DM catch-all:** the default Space listed *every* bare DM,
   including DMs whose messages belong exclusively to other Spaces. A positive-
   evidence post-pass now hides such a DM iff `dm_space_presence` has rows for the
   pair, none for the default Space, and Recents carry no untagged/default-tagged
   counter-evidence. DMs with no presence rows keep the legacy catch-all; system
   bots exempt; regular-bot visibility stays with the existing bot sub-check;
   `skipBotFilter` or any query failure disables the pass.
4. **Spaceless group/topic:** a group with an empty `group.space_id` (and its
   topics) was visible in *every* Space; now attributed to the default Space only
   (conv filter, `filterThreadConvCore`, and sidebar `filterThreadExtsBySpace`).

## Related Issue

Fixes #484.

## Linked Spec

- `.octospec/tasks/dm-space-isolation-484/brief.md` (base isolation)
- `.octospec/tasks/conv-space-catchall-484/brief.md` (recent-list catch-all)
- Journals under `.octospec/journal/shared/`.

## Changes

- **New infra:** `dm_space_presence` table (`modules/space/sql/…`),
  `pkg/space/dm_presence.go` (`UpsertDMSpacePresence`/`DMSpacePresenceSet`) +
  `dm_presence_any.go` (`DMSpacePresenceAnySet`), post-commit best-effort write in
  `modules/webhook/api.go` (never blocks message persistence).
- **`modules/message/space_filter.go`:** `decideConvKeepInSpace` gains the
  `dmPresentSet` presence-OR (symptom 2) and the spaceless-group default-Space
[...]

PR #527, opened 2026-07-03
fix(message): keep system-bot DM history visible under X-Space-ID

an9xyz:
## Summary

`POST /v1/message/channel/sync` returned `messages: []` for **system-bot DMs**
(`botfather` / `u_10000` / `fileHelper` / `notification`) whenever the request
carried a validated `X-Space-ID`. System-bot messages are Space-agnostic
(`payload.space_id == ""`), and `filterPersonMessagesBySpace` dropped every
untagged system-bot message — so the conversation list showed botfather (forced
in by `EnsureSystemBotsPresent`) but opening it was empty.

System bots are **Space-independent and must always be visible**
(`pkg/space/query.go` invariant). This PR exempts them from
`filterPersonMessagesBySpace` (early return of the full list before the
per-message Space switch), making history visibility consistent with the
conversation-list contract. Regular-DM rules — including #484's symptom-1
(untagged history → default-Space-only) — are unchanged.

**This is not a #484/PR #519 regression.** Verified by diff: #484 only changed the
`!isSysBot` (regular-DM) branch; the system-bot branch and the filter's
`X-Space-ID` gate in `syncChannelMessage` are byte-identical before/after #484.
Root cause is **YUJ-219-A / #1283** (`e39b69f`, 2026-05-04), which first added the
system-bot message-drop branch.

**No data loss.** This is a read-time filter on the sync *response* — it never
writes or deletes. WuKongIM storage is intact; the fix restores history
immediately, no backfill/migration.

## Related Issue

Regression of YUJ-219-A / #1283 (system-bot DM history). No new issue filed.

## Linked Spec

- `.octospec/tasks/system-bot-dm-history/brief.md`
- Journal: `.octospec/journal/shared/system-bot-dm-history.md`

## Changes

- `modules/message/space_filter.go` — `filterPersonMessagesBySpace` early-returns
  the full list when `spacepkg.IsSystemBot(channelID)`; removed the now-dead
  `isSysBot` variable and its switch cases (rules simplified to regular-DM only).
- `modules/message/space_filter_test.go` — flipped the 3 YUJ-219-A tests that
  locked the drop behavior (`_SystemBot`, `_PayloadSpaceIDWrongType` fileHelper
  case, `_SystemBotListCoverage`) to assert full visibility;
  `_SystemBotListCoverage` now also covers the **non-default** Space (the repro).
- `.octospec/` — task brief, journal, and log entry.

## Testing

```
go build ./...                                                 ok
go vet ./modules/message/... ./pkg/space/...                   ok
make i18n-lint && make i18n-extract-check                      ok
go test -race -shuffle=on ./modules/message/ ./pkg/space/      ok  (CI-parity unit)
go test ./modules/message/ -run TestFilterPersonMessagesBySpace -v   all PASS
```

Reproduced first at the pure-function level (botfather untagged history 0/2 under
a Space; 2/2 without X-Space-ID; regular peer 2/2 in default — isolating the
system-bot branch as the cause), then fixed and locked with the flipped/extended
tests.

- [x] Manually verified (pure-function repro + CI-parity unit run)

## COMPREHENSION

   It stops the DM message-history filter (`filterPersonMessagesBySpace`, used by
   `/v1/message/channel/sync`) from Space-filtering system-bot conversations.
   System bots now return their complete history in every Space, matching the
   conversation-list guarantee (`EnsureSystemBotsPresent` / `SystemBots`
   always-visible). Only the system-bot branch changes; regular-DM filtering
   (exact-match, untagged→default-only, cross-Space drop) is untouched.

   Only `syncChannelMessage` calls this function, only for Person channels, only
   when a validated `X-Space-ID` is present. The exemption strictly *widens* what
   a system-bot sync returns (never narrows a regular DM), so it cannot introduce
   a cross-Space leak for regular DMs. System bots are global single conversations
   — returning their full history in any Space is correct, not a leak. Client-side
   `filterSystemBotMessages` (Android/iOS) may still filter; if so the server fix
   alone won't surface it — tracked as a follow-up in the brief.

   Pure-function repro reproduced the empty result and isolated rule 4 as the
   cause; the fix flips it to full visibility. Regression tests assert every
   system bot returns its untagged history in both default and non-default Spaces,
   and a #484 guard test confirms regular untagged DMs are still default-Space-only.
   CI-parity unit run green under `-race -shuffle=on`.

## Checklist

- [x] Updated documentation (octospec brief/journal/log)

https://claude.ai/code/session_01EmFRUBSc2x7kpDaLs3Gyco

---
_Generated by [Claude Code](https://claude.ai/code/session_01EmFRUBSc2x7kpDaLs3Gyco)_

### DIFF
--- a/.octospec/journal/shared/system-bot-dm-history.md
+++ b/.octospec/journal/shared/system-bot-dm-history.md
@@ -1,0 +1,59 @@
+---
+type: Journal
+title: "system-bot-dm-history: system-bot DM history must not be Space-filtered"
+description: Exempt system bots from filterPersonMessagesBySpace so botfather/fileHelper/u_10000 history is not returned empty under an X-Space-ID
+tags: [space, isolation, message, bot, systembot]
+timestamp: 2026-07-03T00:00:00Z
+---
+
+# system-bot-dm-history
+
+Branch `fix/system-bot-dm-history` (based on latest `origin/main`, which already
+carries #484 / PR #519).
+
+## What was done
+
+`filterPersonMessagesBySpace` (`modules/message/space_filter.go`) now **early-returns
+the full message list for system bots** (`spacepkg.IsSystemBot(channelID)`), before
+the per-message Space switch. System bots (`botfather` / `u_10000` / `fileHelper` /
+`notification`) are Space-independent and must always be visible — the conversation
+list already forces them into every Space via `EnsureSystemBotsPresent`, so their DM
+history must not be Space-filtered either. The old rule 4
+(`payload.space_id == "" && isSysBot → drop`) was silently emptying botfather history
+whenever the sync request carried an `X-Space-ID`. With the early return, the
+`isSysBot` variable and its switch cases are removed; the remaining switch handles
+only regular DMs (rules 1–4 renumbered, #484 symptom-1 semantics unchanged).
+
+## Root cause / attribution
+
+- Bug origin: YUJ-219-A / #1283 (`e39b69f`, 2026-05-04) added the system-bot
+  message-drop branch. **Not** introduced by #484 (PR #519) — verified by diff:
+  #484 only changed the `!isSysBot` regular-DM branch; the system-bot path is
+  byte-identical before/after #484, and the filter call + its `X-Space-ID` gate in
+  `syncChannelMessage` (`api.go`) pre-date #484.
+- Read-time filter only: the endpoint filters the sync **response**, never writes
+  or deletes. WuKongIM storage is intact; the fix restores history immediately
+  (no backfill). Worst case was a client that wipe-replaced its local cache on the
+  empty response — self-heals on the next sync.
+
+## Verification
+
+- Reproduced first (pure function): botfather untagged history 0/2 under a Space,
+  2/2 without X-Space-ID; a regular peer 2/2 in default (isolates rule 4 as cause).
+- Flipped 3 existing YUJ-219-A tests that locked the drop behavior
+  (`_SystemBot`, `_PayloadSpaceIDWrongType` fileHelper case, `_SystemBotListCoverage`)
+  to assert full-visibility; `_SystemBotListCoverage` now also covers the
+  **non-default** Space (the exact production repro).
+- #484 guard kept: `_UntaggedDroppedInNonDefaultSpace` / `_OrdinaryDMLegacyCompat`
+  still green (regular untagged DM stays default-Space-only).
+- `go test -race -shuffle=on ./modules/message/ ./pkg/space/` green; `go build`,
+  `go vet` green.
+
+## Learnings
+
+- Two Space contracts must stay in lockstep: conversation-list visibility
+  (`EnsureSystemBotsPresent` / `decideConvKeepInSpace`) and message-history
+  visibility (`filterPersonMessagesBySpace`). A system bot shown in the list but
+  empty on open is the smell of the two diverging.
+- Follow-up: confirm whether the client `filterSystemBotMessages` (Android/iOS)
+  also drops system-bot history; if so the server fix alone won't surface it.
--- a/.octospec/log.md
+++ b/.octospec/log.md
@@ -4,6 +4,20 @@ Change history for this repo's `.octospec/`, following the
 [OKF](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md)
 change-log convention (§7). Newest first.
 
+## 2026-07-03
+
+- **Fix** — Task `system-bot-dm-history`: system-bot DM history was returned empty
+  from `/v1/message/channel/sync` whenever the request carried an `X-Space-ID`.
+  `filterPersonMessagesBySpace` now early-returns the full list for system bots
+  (`botfather`/`u_10000`/`fileHelper`/`notification`) — they are Space-independent
+  and already forced into every Space's conversation list by `EnsureSystemBotsPresent`,
+  so their history must not be Space-filtered. Root cause is YUJ-219-A / #1283
+  (`e39b69f`), **not** #484 (verified by diff: #484 only touched the regular-DM
+  branch). Read-time filter only — no data loss, no backfill. Flipped 3 YUJ-219-A
+  tests that locked the drop behavior; added non-default-Space coverage. Brief +
+  journal under `.octospec/tasks/system-bot-dm-history/` and
+  `.octospec/journal/shared/system-bot-dm-history.md`.
+
 ## 2026-07-02
 
 - **Change** — Task `conv-space-catchall-484` (issue #484 follow-up): closed the
--- a/.octospec/tasks/system-bot-dm-history/brief.md
+++ b/.octospec/tasks/system-bot-dm-history/brief.md
@@ -1,0 +1,65 @@
+---
+type: Task
+title: "Task: system-bot-dm-history"
+description: System-bot DM history must not be Space-filtered — botfather/fileHelper/u_10000 returned empty from /message/channel/sync under an X-Space-ID.
+tags: [space, isolation, message, bot, systembot]
+timestamp: 2026-07-03T00:00:00Z
+# --- octospec extension fields ---
+slug: system-bot-dm-history
+upstream: octo-server (regression of YUJ-219-A / #1283)
+source: self
+---
+
+# Task: system-bot-dm-history
+
+## Goal
+
+Stop `filterPersonMessagesBySpace` from dropping a **system bot**'s DM history.
+`POST /v1/message/channel/sync` for a system bot (`botfather` / `u_10000` /
+`fileHelper` / `notification`) returns `messages: []` whenever the request
+carries a validated `X-Space-ID`, because the bot's messages are Space-agnostic
+(`payload.space_id == ""`) and the filter's system-bot branch drops all untagged
+system-bot messages. System bots are Space-**independent** and must always be
+visible, so their history must never be Space-filtered.
+
+## Background
+
+`filterPersonMessagesBySpace` (YUJ-219-A / #1283, commit `e39b69f`, 2026-05-04)
+added message-level Space filtering to close a cross-Space DM-history leak. Its
+rule 4 (`payload.space_id == "" && isSystemBot → drop`) hides untagged system-bot
+history — but this directly contradicts the invariant in `pkg/space/query.go`
+("系统 Bot … 必须始终对客户端可见") and the conversation-list contract
+(`EnsureSystemBotsPresent` forces system bots into every Space's list). Net effect:
+the conversation list shows botfather, but opening it returns empty history.
+
+This is **not** introduced by #484 (PR #519): that PR only changed the
+`!isSysBot` (regular-DM) branch (untagged history → default-Space-only); the
+system-bot branch is byte-identical before and after #484. Read-time filter only
+— no stored data is deleted; the fix restores history immediately, no backfill.
+
+## Load-bearing list
+
+- `modules/message/space_filter.go` `filterPersonMessagesBySpace` — message-level
+  Space filter for the DM history endpoint (`/v1/message/channel/sync`).
+- Invariant `pkg/space/query.go` `SystemBots` / `IsSystemBot` — system bots are
+  Space-independent and always visible.
+- Conversation-list contract `EnsureSystemBotsPresent` / `decideConvKeepInSpace`
+  (`SystemBots[channelID] → keep`) — must stay consistent with history visibility.
+- #484 symptom-1 rule (regular untagged DM → default-Space-only) — must NOT
+  regress.
+
+## Out of scope
+
+- Client-side `filterSystemBotMessages` (Android/iOS) — if the clients also
+  Space-filter system-bot history, they need a matching change; tracked separately.
+- The regular-DM (`!isSystemBot`) rules — unchanged.
+- Any storage / migration / backfill (read-time filter only).
+
+## Acceptance
+
+- `filterPersonMessagesBySpace(msgs, "<systembot>", spaceID, defaultSpaceID)`
+  returns the **full** list unchanged for every system bot, in both default and
+  non-default Spaces, incl. untagged messages.
+- #484 guard: a regular peer's untagged DM is still kept only in the default
+  Space and dropped in a non-default Space.
+- `go test -race ./modules/message/ ./pkg/space/` green; CI green.
--- a/modules/message/space_filter.go
+++ b/modules/message/space_filter.go
@@ -512,14 +512,17 @@ func newSystemBotPlaceholder(uid string) *SyncUserConversationResp {
 // 本函数仅针对 Person (DM) 路径：
 //   - GROUP channel_id 本身做 Space 隔离（不同 Space 的群 channel_id 不同），
 //     对历史消息再过滤反而会误杀老群，因此 GROUP/COMMUNITY_TOPIC 路径不走这里。
-//   - 规则（issue #484 后；与三端 SpaceFilter 口径对齐）：
-//     1) payload.space_id == spaceID                         → 保留（精确匹配）
-//     2) payload.space_id == "" && !isSystemBot && 默认Space  → 保留（无标签 DM
-//     历史只在用户默认 Space 向前兼容，避免出现在每个 Space —— 症状1）
-//     3) payload.space_id == "" && !isSystemBot && 非默认Space→ 丢弃（不再 fail-open）
-//     4) payload.space_id == "" &&  isSystemBot               → 丢弃（SystemBot 无
-//     space 标签的老消息默认隐藏，避免 fileHelper/u_10000 老消息跨 Space 泄露）
-//     5) payload.space_id != "" && != spaceID                → 丢弃（跨 Space 污染）
+//   - 系统 Bot（botfather/u_10000/fileHelper/notification）豁免：与 Space 无关，
+//     必须始终对客户端可见（pkg/space query.go 的不变量；会话列表侧由
+//     EnsureSystemBotsPresent 保证其在每个 Space 出现）。系统 Bot 的 DM 是全局
+//     单会话，历史同样不按 Space 过滤 —— 否则会话列表里有 botfather、点进去
+//     /message/channel/sync 却返回空。全局单会话不存在"跨 Space 泄露"。
+//   - 普通 DM 规则（issue #484 后；与三端 SpaceFilter 口径对齐）：
+//     1) payload.space_id == spaceID          → 保留（精确匹配）
+//     2) payload.space_id == "" && 默认Space   → 保留（无标签 DM 历史只在用户默认
+//     Space 向前兼容，避免出现在每个 Space —— 症状1）
+//     3) payload.space_id == "" && 非默认Space → 丢弃（不再 fail-open）
+//     4) payload.space_id != "" && != spaceID  → 丢弃（跨 Space 污染）
 //
 // 调用方需保证 spaceID != ""（空串视为未启用 Space 过滤，直接返回原列表），
 // 传入 defaultSpaceID（用户默认 Space，决定规则 2/3），并只对 ChannelTypePerson 调用。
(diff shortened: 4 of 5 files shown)


OctoBoooot (commented):
## Review: fix(message): keep system-bot DM history visible under X-Space-ID (#527)

Correct, well-scoped fix for an over-hiding bug, and it does not re-open the over-showing (cross-Space leak) it was originally guarding against. Not stamping a clean APPROVE only because `Test`/`Vet` haven't reported. `check-sprint` red is governance (doesn't bar approval); merge stays BLOCKED until code CI + check-sprint clear.

Risk tier: normal, but reviewed on the Space-isolation axis (this is #484/#519 family — the concern is whether un-hiding re-introduces a leak).

### The bug and why the fix is safe (byte-verified)
- **Bug:** `filterPersonMessagesBySpace` rule "`space_id=="" && isSysBot → drop`" over-hid *all* untagged system-bot DM history under a validated `X-Space-ID`, so the conversation list showed botfather (forced by `EnsureSystemBotsPresent`) but `/message/channel/sync` returned `[]`.
- **Fix:** early `return msgs` when `IsSystemBot(channelID)` — full history, no Space filter. Verified safe on three points:
  1. `IsSystemBot(uid)` is a fixed static allowlist (`SystemBots[uid]` in pkg/space/query.go), not user-forgeable — an attacker can't mark a normal DM as system-bot to bypass filtering.
  2. `channelID` = `req.ChannelID` (the DM peer); the filter only reshapes messages already fetched for that channel, so this changes *display*, not *access* (the fetch is gated upstream).
  3. A system-bot DM is a single **global** channel with Space-agnostic (`space_id==""`) messages — there is no other-Space content in it to leak, so "show full history in any Space" is not cross-Space exposure. The body's "全局单会话不存在跨 Space 泄露" reasoning holds specifically because it's scoped to system bots.
- **Over-show not re-opened for normal DMs:** the early-return is gated strictly on `IsSystemBot`; the normal-DM rules are byte-preserved (untagged+default→keep, untagged+non-default→drop, foreign-tag→drop via the new `default:` arm). The rule renumbering is cosmetic; the logic for non-system DMs is unchanged.

### Coverage
1.0. Tests assert full history returned for all four system bots (botfather/u_10000/fileHelper/notification) in **both** default and non-default Spaces, include the previously-dropped untagged message, and the surrounding suite retains the normal-DM cross-Space cases. Good bidirectional coverage (system-bot visible ∧ normal-DM isolation preserved).

### Praise
- Restructuring to a single `IsSystemBot` early-return — rather than threading `!isSysBot` through every case arm (as the old rules 2/3/4 did) — is the right refactor: it makes "system bots are exempt from Space filtering, full stop" a structural guarantee at the top of the function instead of a condition that must be repeated correctly in each branch (where the old rule 4 got the exemption backwards and *hid* them). And correctly identifying that the exemption is safe *because* a system-bot DM is a global single channel — not just asserting "bots are special" — is the reasoning that makes this defensible.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #527 (octo-server)

## Summary

Two independent, well-scoped changes. First, `filterPersonMessagesBySpace` now
exempts system bots (botfather, u_10000, fileHelper, notification) from
Space-based message filtering — fixing the bug where system bot DM conversations
appeared in the conversation list but showed empty history when X-Space-ID was
set. Second, the forced "Webhook-" name prefix for member/bot-created webhooks
is removed; members can now set custom names on par with admins, while the push
path's per-request override protection (`resolveFromIdentity`) remains intact.

## Verification

- ✅ **System bot exemption correctness** — `IsSystemBot(channelID)` early-returns
  at `modules/message/space_filter.go:536` before the filter loop. System bots
  are defined in `pkg/space/query.go:22` as {botfather, u_10000, fileHelper,
  notification}. `EnsureSystemBotsPresent` (conversation sync) already injects
  system bots into every Space's conversation list — this change makes message
  history consistent with that invariant.
- ✅ **Push-path override protection unchanged** — `resolveFromIdentity` at
  `modules/incomingwebhook/api.go:1622` still gates `req.Username`/`req.AvatarURL`
  overrides behind `allowOverride` (= `creatorIsAdmin`). Member webhook push
  requests cannot override the stored name/avatar per-push, regardless of the
  stored name content. The most dangerous impersonation vector (per-push, no
  audit trail) remains blocked.
- ✅ **Avatar still admin-only** — Create (`api.go:695`) and update (`api.go:898`)
  both reject `req.Avatar` when `!actor.isAdmin`.
- ✅ **Name byte cap symmetric** — Both create (`api.go:689`) and update
  (`api.go:891`) enforce `len(name) > 64`. `resolveFromIdentity` truncates to
  `maxFromNameBytes` (64) at `api.go:1633`. Admin and member paths share the
  same cap.
- ✅ **Create/update parity** — Both paths remove the forced prefix logic and
  apply identical `TrimSpace` + length validation. No stale references to
  `memberWebhookNamePrefix` or `prefixedWebhookName` remain in production code.
- ✅ **Tests updated** — `space_filter_test.go` now asserts system bot history is
  returned in both default and non-default Spaces. `api_member_test.go` asserts
  custom names are stored verbatim, byte boundaries are enforced on both
  create/update, and update stores names without prefix.
- ✅ **Auto-naming fallback intact** — `autoWebhookName(webhookID)` at
  `api.go:397` still generates `Webhook-xxxxxx` when name is omitted.

## Findings

No P0/P1 issues. Two P2 observations below.

### P2 — Existing webhooks retain forced prefix (`modules/incomingwebhook/api.go`)

Webhooks created before this PR carry "Webhook-" in their stored name (the old
forced prefix). There is no migration or backfill to strip it. Existing users
see "Webhook-my ci bot" while new creates show "my ci bot" for the same input.
Non-destructive — users can manually update via the existing rename endpoint —
but worth a product-level decision on whether a one-time backfill is desired.

### P2 — No push-path rendering test for member webhook with impersonation-prone name (`modules/incomingwebhook/api_member_test.go`)

`TestPush_MemberWebhookOverrideSilentlyIgnored` verifies that per-push
`username`/`avatar_url` overrides are ignored for member webhooks, but does not
assert the rendered `from.name` when the stored name is an impersonation-prone
string (e.g. "HR 公告"). The `resolveFromIdentity` gate is covered by
in-package unit tests (`richtext_test.go:173+`), but an E2E assertion here
would close the loop: member webhook with arbitrary stored name + push with
override attempt → rendered name is the stored name, not the override.

## Blindspot Checklist (C1–C6)

- **C1 (dual-path parity)**: ✅ Create and update both remove forced prefix,
  both enforce 64-byte cap, both reject non-admin avatar. Symmetric.
- **C2 (control-flow ordering)**: ✅ No reused/嵌套 helpers affected.
  `filterPersonMessagesBySpace` early-return is clean; no other caller invokes
  it with a system bot channel ID through a different ordering.
- **C3 (authorization boundary ≠ capability boundary)**: ✅ Push-path override
  gate (`resolveFromIdentity`) still checks `creatorIsAdmin` from
  `cachedCreatorMembership`. The name-storage boundary (create/update endpoints)
  is authenticated and audited. No new unauthenticated capability is exposed.
- **C4 (authorization lifecycle)**: N/A — no auth/lifecycle changes.
- **C5 (build/note ≠ runtime correctness)**: ✅ Static analysis only; build and
  tests not executed in this environment.
- **C6 (governance/docs self-consistency)**: ✅ Comments and `.octospec` journal
  entries are consistent with the code changes. README update at
  `modules/incomingwebhook/README.md` aligns with the new behavior.

## Data-Flow Trace

**System bot history path**: client `/message/channel/sync` with X-Space-ID →
`api.go:1370` calls `filterPersonMessagesBySpace(msgs, req.ChannelID, spaceID,
defaultSpaceID)` → `space_filter.go:536` checks `IsSystemBot(channelID)` →
early-returns `msgs` for system bots. ChannelID comes from the client request
and is validated upstream. System bot UIDs are a fixed set in
`pkg/space/query.go:22`. No cross-Space data leak — system bots are global
single-conversation (no per-Space DM).

**Webhook name path**: member POST `/v1/groups/{groupNo}/incoming-webhooks` with
`{"name": "custom"}` → `api.go:687` TrimSpace → `api.go:689` 64-byte cap →
`api.go:783` stored verbatim in `incomingWebhookModel.Name` → push path
`api.go:1591` `buildPayload(m, req, creatorIsAdmin=false)` →
`resolveFromIdentity` ignores `req.Username`, uses `m.Name` → truncated to 64
bytes → rendered as `from.name`. Stored name requires authenticated management
endpoint; per-push override requires admin. Both gates intact.

## Verdict: APPROVED

Both changes are correct, well-tested, and maintain the critical security
invariants. The push-path override protection — the primary anti-impersonation
control — remains intact. The system bot exemption fix resolves a real
user-facing bug with clean code and updated tests. The two P2 items are
non-blocking observations.

Jerry-Xin (approved):
**APPROVE** — clean, correctly-scoped visibility fix. System-bot DM history is exempted from `filterPersonMessagesBySpace` via an early return, restoring the empty-on-open regression under `X-Space-ID`. Both the intended widening and the inverse (leak) direction were traced; no cross-space or cross-user exposure is introduced.

### ✅ Widen direction (intended)
`filterPersonMessagesBySpace` now `return msgs` early when `spacepkg.IsSystemBot(channelID)`, before the per-message Space switch. So untagged (`payload.space_id == ""`) system-bot history (`botfather`/`u_10000`/`fileHelper`/`notification`) is returned in full under any `X-Space-ID`. This matches the conversation-list contract (`EnsureSystemBotsPresent` forces system bots into every Space) and the `pkg/space/query.go` invariant — fixing "list shows botfather, open returns empty".

### ✅ Inverse / leak direction (verified — the highest-risk axis)
- **Cross-user: no leak.** The only caller is `syncChannelMessage` (`modules/message/api.go:1370`). Messages come from `IMSyncChannelMessage` scoped to `req.LoginUID = c.GetLoginUID()` (authenticated caller) on the caller's own Person channel with `req.ChannelID`. This filter runs on the *response* — it is post-fetch shaping, never a fetch-authorization gate. Loosening it cannot let a non-participant read another user's DM; the channel-membership fetch boundary upstream is unchanged.
- **Cross-space: no leak.** System-bot DMs are global single conversations (one user↔botfather thread, space-agnostic `space_id == ""`), not per-Space partitioned. Bypassing the `msid != spaceID` drop for them is therefore semantically correct — there is no distinct "Space B botfather thread" to leak into Space A.
- `X-Space-ID` itself is fail-closed validated by `SpaceMiddleware` (non-members 403) before reaching here; the code reads only the middleware-written `spaceID`, no header fallback.

### ✅ System-bot identity is robust (not spoofable)
`IsSystemBot(uid)` = `SystemBots[uid]`, an exact-match lookup against a fixed allowlist map `{botfather, u_10000, fileHelper, notification}` in `pkg/space/query.go`. No substring/prefix match — an attacker-crafted `channelID` cannot spoof into the exemption, and the special-case cannot accidentally cover a non-system bot.

### ✅ Regular-DM path preserved
The switch simplification is behavior-preserving for non-system channels: exact-match kept, `msid == "" && default Space` kept (#484 symptom-1), `msid == "" && non-default` dropped, and the new `default:` still drops explicit cross-Space (`msid != "" && != spaceID`). The removed `isSysBot` cases are dead for non-bots after the early return.

### ✅ Tests
Three YUJ-219-A tests flipped from "drop" to "full visibility" (`_SystemBot`, `_PayloadSpaceIDWrongType` fileHelper case, `_SystemBotListCoverage`), and `_SystemBotListCoverage` now also asserts the **non-default** Space (the production repro). #484 guards (`_UntaggedDroppedInNonDefaultSpace` / `_OrdinaryDMLegacyCompat`) retained.

### CI / caveat
Build / Vet / Lint / i18n / MsgSendReq Lint all **SUCCESS**; **Test** was still **IN_PROGRESS** at review time (not RED — no logic failure observed), `check-sprint` FAILURE is governance-only and does not bar approve. Read-time response filter only — no write/delete/migration, so no data-loss risk. If the `Test` job does not go green, worth a glance, but the change is byte-verified clean.

Nice attribution work isolating YUJ-219-A / #1283 (`e39b69f`) as the true root cause rather than #484.

yujiawei (approved):
# Code Review — PR #527 (octo-server)

## Summary

This change exempts system bots (`botfather` / `u_10000` / `fileHelper` / `notification`) from `filterPersonMessagesBySpace`, so `POST /v1/message/channel/sync` no longer returns an empty history for a system-bot DM when the request carries a validated `X-Space-ID`. Implementation is a single early return in `modules/message/space_filter.go` plus a simplification of the now regular-DM-only rule set, with the three lock-in tests flipped to assert full visibility.

I reviewed against the head SHA, rebuilt and re-ran the affected packages, and took an extra pass because the PR is Space-isolation / security-sensitive.

**Verdict: APPROVED.** No blocking (P0/P1) issues. One design point and two nits below — none block merge.

## What I verified

- `go build ./modules/message/... ./pkg/space/...` — ok
- `go vet ./modules/message/... ./pkg/space/...` — ok
- `go test ./modules/message/` (full package) — ok
- `go test ./modules/message/ -run TestFilterPersonMessagesBySpace -v` — all 10 cases PASS

Correctness of the core claim:
- The only production caller is `syncChannelMessage` (`modules/message/api.go:1370`), Person-channel path only, guarded by a validated `spaceID != ""` from `SpaceMiddleware` (no fallback to raw `X-Space-ID` header). ✅
- `IsSystemBot` resolves against a fixed static allowlist (`pkg/space/query.go:20`), not attacker-controlled data — `channelID` is the routed DM peer UID and cannot be spoofed into the allowlist. ✅
- The #484 regular-DM rules are preserved: `TestFilterPersonMessagesBySpace_UntaggedDroppedInNonDefaultSpace` and `_OrdinaryDMLegacyCompat` still pass. The `!isSysBot` guards were removed only because the early return makes them dead for system bots; the surviving `switch` is byte-equivalent for regular DMs. ✅

## Design note (non-blocking): the exemption also returns Space-*tagged* system-bot messages

The early return short-circuits *all* filtering, so a system-bot message explicitly tagged `payload.space_id == "spaceB"` is now returned when syncing that bot under `X-Space-ID = spaceA` (locked by `space_filter_test.go` `TestFilterPersonMessagesBySpace_SystemBot`, which asserts a `spaceB`-tagged message is returned under `spaceA`). At first glance this reads like a cross-Space leak.

I do **not** consider it a blocker, because it is consistent with the established system-bot contract rather than a violation of it:
- System bots are documented as **Space-independent global single conversations** (`pkg/space/query.go` invariant), and `decideConvKeepInSpace` (`space_filter.go:350`) already returns the bot's conversation in **every** Space unconditionally, regardless of tag.
- The data returned is the requesting user's **own** global bot inbox — there is no other-tenant/other-user data crossing a boundary here (unlike a regular peer DM). Filtering the history by Space would put it *out of sync* with the always-visible conversation entry and reproduce exactly the empty-history bug this PR fixes.

Given that, returning the full list (tagged or not) is the internally-consistent choice. Worth a one-line comment on the early return noting "tagged messages are intentionally included — system-bot DMs are a single global conversation," so a future reader doesn't re-file this as a leak. Optional.

## Nits (non-blocking)

1. `space_filter.go:537` — the early return hands back the input slice by reference and skips the `if m == nil { continue }` sanitation the regular path applies. In practice the sole producer (`newSyncChannelMessageResp`) never appends `nil` and the caller reassigns the result without mutating it, so this is safe today; it's a mild asymmetry, not a correctness bug. If you want strict symmetry you could return a copied, nil-filtered slice, but it's not required.

## Coverage / blind spots

- **Client-side parity (flagged by the PR itself):** Android/iOS `filterSystemBotMessages` may still Space-filter system-bot history client-side; if so, the server fix alone won't surface history on those clients. The PR correctly scopes this out and tracks it as a follow-up — calling it out here so a human confirms the mobile follow-up is actually filed.
- **Human verification suggestion:** since this is on the Space-isolation path, worth one manual smoke on a real non-default Space (open botfather → confirm history renders) before/after deploy, to confirm the middleware-supplied `spaceID` and default-Space lookup behave as the unit tests assume end-to-end.
- No storage/migration risk: this is a read-time filter on the sync response only.

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD 88a25c26

**Formal Verdict**: `APPROVE`



## Reviewer Summary

|---|---|---|
| qa-engineer | pass-with-risk | Server fix correct and locked by pure-function tests; residual risks are all outside server code (client rendering filter, observability, local DB suite). |
| security-engineer | cleared | Pure read-time Space view-partition widening for global system bots only; primary access control (per-user composite channel + SpaceMiddleware fail-closed) untouched, no leak/bypass. |
| code-reviewer | approved | Minimal, correct early-return exemption aligned with the `pkg/space` always-visible invariant; #484 regular-DM semantics equivalently preserved. Must=0 Should=0 Nit=3. |

## Critical Findings (P0-P1)

## Should-Fix (P2)

无 P2 findings.

## Suggestions (P3)

- [P3-1] `modules/message/space_filter.go:534` — Exemption path is a silent early return with zero observability (no debug log / counter for system-bot messages bypassing Space partition). Acceptable while `SystemBots` is a fixed finite set; if it later becomes config/DB-driven, a debug-level hit log or counter would help telemetry catch a misclassification (a non-system-bot matching `IsSystemBot`). Non-blocking. — Fix: add a debug log / metric on exemption hit if/when the system-bot set becomes dynamic. — flagged by qa/Risk-2 + security/Note-1 + code/Nit-2
- [P3-2] `modules/message/space_filter.go:534` — After the early return the system-bot branch no longer skips `nil` elements (the regular-DM branch still does `if m == nil { continue }`). Upstream `newSyncChannelMessageResp` only appends non-nil `msgResp`, so this path does not produce nil — theoretically unreachable, not a defect. — Fix: none required; optionally a one-line comment noting the invariant. — flagged by code/Nit-1 (confidence 0.6)

## Questions for Author

- [Q-1] End-to-end visibility still depends on client rendering filters — Android `filterSystemBotMessages` / iOS `filterMessagesBySpace` may still Space-filter, so a mobile client opening botfather could remain empty even with this server fix. The author already lists this as a follow-up in the brief. Non-blocking for this PR (server-side responsibility is met); recommend a follow-up issue to track syncing the client-side rendering filters. — from qa/Risk-1 + code/Nit-3
- [Q-2] The full `modules/message` `-race -shuffle=on` suite could not run locally (2 unrelated integration tests fail on missing MySQL — `TestChannelFiles_TopicBlacklist`, `TestOnThreadCreated_BlacklistedMemberExcludedFromFanout`, both outside the files this PR touches). Recommend confirming the GH Actions `modules/message` suite is green (with DB) before merge. — from qa/Risk-3

## Verdict Rationale

Deterministic first-match on the consolidated buckets: P0=0, P1=0, P2=0, only P3 (2 non-blocking nits) with no blocking question → **APPROVE**. All three reviewers converged — security `cleared` (no attack-surface change: the composite `fakeChannelID` carries `loginUID`, so the exemption only widens the requester's *own* data across Spaces and cannot leak third-party or regular-DM cross-Space data), code-reviewer (chain terminal) `approved` with only nits, and QA `pass-with-risk` where every listed risk lives outside the server code under review (client rendering filter, observability gap, local-DB CI parity). No reviewer disagreement to reconcile: the observability gap and the client follow-up were independently raised by multiple reviewers and merged here, and QA's `pass-with-risk` is explicitly attributed to client/observability/local-DB factors rather than any server-code defect. The two Questions are follow-ups and a CI confirmation, not blockers to the code change itself.

## Round Context

---

lml2468 (approved):
**APPROVE — Round 1** (HEAD 88a25c26)

All three reviewers converged: security `cleared`, code-reviewer `approved`, QA `pass-with-risk` — no P0/P1/P2 findings, only 2 non-blocking nits (exemption-path observability; theoretically-unreachable nil passthrough). No critical findings.

Non-blocking follow-ups: client-side rendering filters (Android/iOS) may still Space-filter system-bot history — recommend a follow-up issue; and confirm the GH Actions `modules/message` suite is green (with DB) before merge.

Full findings: https://github.com/Mininglamp-OSS/octo-server/pull/527#issuecomment-4872279793

PR #532, opened 2026-07-03
fix(message): default-Space DM conversation preview & unread count (#484)

an9xyz:
## Summary

The DM conversation-list preview (`space_last_message`) and per-Space unread count were computed by matching `payload.space_id == filterSpaceID` **strictly**. DM messages sent in the **default Space carry no `payload.space_id`** (untagged), so a **default-Space** `POST /v1/conversation/sync?space_id=<default>` matched nothing:

- `space_last_message` came back `nil` → the client fell back to `recents[last]` = the channel's **global-last** message, which is tagged for **another Space** → the preview leaked a wrong-Space message.
- per-Space unread likewise **under-counted** the default Space (untagged messages were skipped).

A non-default-Space query was already correct (its messages are explicitly tagged).

## Related Issue

Related to #484. This preview/unread logic (`modules/message/space_unread.go`) is **pre-existing** and is not touched by the #484 isolation revert (#529) — the defect exists independently of #519.

## Linked Spec

Issue #484 (DM per-Space isolation). This PR fixes one specific, self-contained facet (the default-Space preview/unread attribution) and can land independently of the broader isolation redesign.

## Changes

- Add `dmSpaceMatch(msgSpaceID, targetSpaceID, defaultSpaceID)` implementing the **"untagged = default Space"** convention: an untagged DM message (no `payload.space_id`) belongs to the user's default Space. This aligns the preview/unread with the visibility catch-all (`decideConvKeepInSpace`) and the message-history filter (`filterPersonMessagesBySpace`), which already treat untagged DM content as default-Space content.
- Thread `defaultSpaceID` through `fillPersonSpaceUnread` → `findSpaceLastMessage` / `findSpaceLastMessageFallback` / `countSpaceUnreadFromMessages`.
- Pass the already-resolved `defaultSpaceID` (from `GetUserDefaultSpaceIDE`) at the `/v1/conversation/sync` call site.
- **Isolation preserved:** non-default-Space queries stay strict (untagged never leaks into a non-default Space); the convention is disabled when `defaultSpaceID == ""` (falls back to prior strict behavior).
- Tests: unit (`space_unread_test.go`) + an end-to-end integration test driving the real handler through `POST /v1/conversation/sync?space_id=` (`dm_space_last_message_repro_test.go`).

## Testing

```
# unit (default build)
go test ./modules/message/ -run 'SpaceUnread|FindSpaceLastMessage|CountSpaceUnread|FillPersonSpaceUnread|DefaultSpace' -count=1
# ok — 27 tests pass (existing preserved + new default-Space cases)

# integration end-to-end (MySQL + Redis + WuKongIM)
go test -tags=integration ./modules/message/ -run 'TestFix_' -count=1
# ok — default-Space query now returns the untagged message as space_last_message
#      (content set, no space_id); non-default query still returns only its tagged message
```

- [x] Manually verified (end-to-end via the real `/v1/conversation/sync` route)

## COMPREHENSION

   On the `/v1/conversation/sync` Space-filter path, `space_last_message` and `SpaceUnread` for Person (DM) channels now attribute an **untagged** DM message to the **default Space**. So a default-Space query resolves its preview to the last untagged (default) message instead of the channel's global-last message, and counts untagged messages toward default-Space unread. Non-default-Space attribution is unchanged.

   Scope is limited to the default-Space DM preview/unread. Only caller is the conversation-sync handler. Risk would be mis-attributing a message to the wrong Space's preview if `defaultSpaceID` were wrong — mitigated two ways: `defaultSpaceID` comes from the same `GetUserDefaultSpaceIDE` the handler already fail-closes on (500 on DB error), and an empty `defaultSpaceID` disables the convention entirely (identical to prior strict behavior). Non-default Spaces cannot start showing untagged content (the second `dmSpaceMatch` branch requires `targetSpaceID == defaultSpaceID`).

   Pure-function unit tests (`TestFindSpaceLastMessage_DefaultSpaceMatchesUntagged`, `TestCountSpaceUnread_DefaultSpaceCountsUntagged`) plus an end-to-end integration test (`TestFix_SpaceLastMessage_DefaultSpaceUsesUntagged`) that drives the real handler through `POST /v1/conversation/sync?space_id=` and asserts: default-Space query → `space_last_message` = the untagged default message (content present, no `space_id`); non-default query → only its explicitly-tagged message, untagged content never leaks in. The same file first captured the pre-fix behavior (nil preview → wrong-Space fallback) and was flipped to a post-fix regression guard.

## Checklist

- [ ] Updated documentation (n/a — no doc surface)

---
_Generated by [Claude Code](https://claude.ai/code/session_01EmFRUBSc2x7kpDaLs3Gyco)_

### DIFF
--- a/modules/message/api_conversation.go
+++ b/modules/message/api_conversation.go
@@ -888,7 +888,7 @@ func (co *Conversation) syncUserConversation(c *wkhttp.Context) {
 	// Space 过滤
 	if hasSpaceFilter {
 		// Person 频道：计算 per-Space 未读计数（在过滤之前，需要原始会话数据）
-		fillPersonSpaceUnread(syncUserConversationResps, conversations, filterSpaceID, loginUID, co.ctx)
+		fillPersonSpaceUnread(syncUserConversationResps, conversations, filterSpaceID, defaultSpaceID, loginUID, co.ctx)
 
 		syncUserConversationResps = FilterConversationsBySpace(
 			syncUserConversationResps, filterSpaceID, loginUID, co.ctx, co.groupService,
--- a/modules/message/dm_space_last_message_repro_test.go
+++ b/modules/message/dm_space_last_message_repro_test.go
@@ -165,6 +165,38 @@ func TestFix_SpaceLastMessage_DefaultSpaceUsesUntagged(t *testing.T) {
 		"the default-space preview message carries no space_id (it is the untagged one)")
 }
 
+// TestFix_SystemBotUntaggedNotInDefaultPreview is the system-bot parity regression
+// guard (PR #532 review by yujiawei/mochashanyao/Jerry-Xin/OctoBoooot): a system-bot
+// DM (botfather) whose default-Space history is untagged must NOT surface a
+// default-Space space_last_message or unread badge — mirroring the history filter's
+// rule-4 drop, so preview/badge can't show what /v1/message/channel/sync hides.
+func TestFix_SystemBotUntaggedNotInDefaultPreview(t *testing.T) {
+	const (
+		botChannel  = "botfather" // spacepkg.IsSystemBot(botfather) == true
+		spaceDefaul = "space-default-bot"
+	)
+	now := time.Now()
+	// Untagged system-bot history (default-Space by the regular-DM convention, but
+	// rule 4 drops it for system bots).
+	m1 := dmMsgRepro(botChannel, 11, now.Add(-2*time.Hour).Unix(), `{"type":1,"content":"bot-a"}`)
+	m2 := dmMsgRepro(botChannel, 12, now.Add(-1*time.Hour).Unix(), `{"type":1,"content":"bot-b"}`)
+	conv := dmIMConvMulti(botChannel, now.Add(-1*time.Hour).Unix(), 100, []*config.MessageResp{m1, m2})
+	conv.Unread = 2 // exercise the unread-count path
+	convs := []*config.SyncUserConversationResp{conv}
+
+	s, ctx := setupConvSyncE2E(t, convs)
+	seedSpaceMemberRepro(t, ctx, spaceDefaul, testutil.UID, "2020-01-01 00:00:00")
+
+	convD := findConv(callConvSyncSpace(t, s, spaceDefaul), botChannel)
+	require.NotNil(t, convD, "system-bot DM stays visible in the conversation list")
+	assert.Nil(t, convD.SpaceLastMessage,
+		"untagged system-bot messages must not become the default-space preview (rule-4 parity)")
+	if convD.SpaceUnread != nil {
+		assert.Equal(t, 0, *convD.SpaceUnread,
+			"untagged system-bot messages must not be counted as default-space unread")
+	}
+}
+
 // TestFix_FindSpaceLastMessage_DefaultMatchesUntagged pins the root-cause fix at
 // the helper level: findSpaceLastMessage now takes defaultSpaceID and treats an
 // untagged DM message as a default-Space message (the same predicate the 200-msg
@@ -180,16 +212,16 @@ func TestFix_FindSpaceLastMessage_DefaultMatchesUntagged(t *testing.T) {
 	}
 
 	// Non-default space: the tagged message is found; the untagged one never leaks.
-	got := findSpaceLastMessage(recents, spaceB, spaceDefault)
+	got := findSpaceLastMessage(recents, spaceB, spaceDefault, false)
 	require.NotNil(t, got)
 	assert.Equal(t, uint32(12), got.MessageSeq)
 
 	// Default space: the untagged default message is now matched (was nil before).
-	got = findSpaceLastMessage(recents, spaceDefault, spaceDefault)
+	got = findSpaceLastMessage(recents, spaceDefault, spaceDefault, false)
 	require.NotNil(t, got, "untagged default-space message is now resolved")
 	assert.Equal(t, uint32(11), got.MessageSeq)
 
 	// Guard: with no default configured (defaultSpaceID==""), the untagged=default
 	// convention is off and a non-default query stays strict.
-	assert.Nil(t, findSpaceLastMessage(recents, spaceDefault, ""))
+	assert.Nil(t, findSpaceLastMessage(recents, spaceDefault, "", false))
 }
--- a/modules/message/dm_space_last_message_repro_test.go
+++ b/modules/message/dm_space_last_message_repro_test.go
@@ -1,0 +1,197 @@
+//go:build integration
+
+package message
+
+// Reproduction for the default-Space DM conversation-preview leak observed on
+// the deployed test env (main) via
+//   POST /v1/conversation/sync?space_id=:space_id
+//
+// Scenario (mirrors the real capture): one physical DM channel whose history is
+//   - an UNTAGGED message  (no payload.space_id → belongs to the DEFAULT space)
+//   - a message TAGGED with a NON-default space (the globally-latest one)
+//
+// Hypothesis under test:
+//   1. Query the NON-default space  → space_last_message is that space's message
+//      (correct — it is explicitly tagged).
+//   2. Query the DEFAULT space      → space_last_message is ABSENT (nil), because
+//      findSpaceLastMessage matches strictly on payload.space_id == filterSpaceID
+//      and the default-space messages are UNTAGGED, so nothing matches. The client
+//      then falls back to recents[last] = the channel's GLOBAL last message, which
+//      is tagged for the OTHER space → the preview leaks a wrong-space message.
+//
+// This drives the REAL handler (syncUserConversation) through the registered
+// /v1/conversation route (which mounts spacepkg.SpaceMiddleware, reading
+// ?space_id=). WuKongIM is the shared httptest fake from
+// conversation_recent_filter_e2e_test.go (same package/build tag): it returns
+// our canned conversation for /conversation/sync and {} for /channel/messagesync
+// (so the findSpaceLastMessageFallback path also finds no default-tagged message
+// — same strict-match blind spot, exercised below as a pure-function check).
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/server"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// dmMsgRepro builds one raw DM message with the given seq/timestamp/payload JSON.
+func dmMsgRepro(channelID string, seq uint32, ts int64, payloadJSON string) *config.MessageResp {
+	return &config.MessageResp{
+		MessageID:   int64(seq)*1000 + 1,
+		MessageSeq:  seq,
+		ClientMsgNo: channelID + "-" + strconv.FormatUint(uint64(seq), 10),
+		FromUID:     channelID,
+		ChannelID:   channelID,
+		ChannelType: common.ChannelTypePerson.Uint8(),
+		Timestamp:   int32(ts),
+		Setting:     0, // not a signal message → payload parsed into the map
+		IsDeleted:   0,
+		Payload:     []byte(payloadJSON),
+	}
+}
+
+// dmIMConvMulti builds a DM conversation (as IMSyncUserConversation returns it)
+// carrying multiple recent messages; LastMsgSeq tracks the newest.
+func dmIMConvMulti(channelID string, ts, version int64, msgs []*config.MessageResp) *config.SyncUserConversationResp {
+	return &config.SyncUserConversationResp{
+		ChannelID:   channelID,
+		ChannelType: common.ChannelTypePerson.Uint8(),
+		Unread:      0,
+		Timestamp:   ts,
+		LastMsgSeq:  int64(msgs[len(msgs)-1].MessageSeq),
+		Version:     version,
+		Recents:     msgs,
+	}
+}
+
+// seedSpaceMemberRepro inserts a space + membership for uid with an explicit
+// created_at (so GetUserDefaultSpaceIDE — earliest membership — is deterministic)
+// and clears the Redis membership cache so the middleware re-checks the DB.
+func seedSpaceMemberRepro(t *testing.T, ctx *config.Context, spaceID, uid, createdAt string) {
+	t.Helper()
+	_, err := ctx.DB().InsertBySql(
+		"INSERT INTO `space` (space_id, name, creator, status, created_at, updated_at) VALUES (?,?,?,1,?,?)",
+		spaceID, spaceID, uid, createdAt, createdAt).Exec()
+	require.NoError(t, err)
+	_, err = ctx.DB().InsertBySql(
+		"INSERT INTO `space_member` (space_id, uid, role, status, created_at, updated_at) VALUES (?,?,0,1,?,?)",
+		spaceID, uid, createdAt, createdAt).Exec()
+	require.NoError(t, err)
+	_ = ctx.GetRedisConn().Del("space:member:" + spaceID + ":" + uid)
+}
+
+// callConvSyncSpace POSTs to the real route with ?space_id= and returns the full
+// decoded conversation objects (so we can inspect space_last_message / recents).
+func callConvSyncSpace(t *testing.T, s *server.Server, spaceID string) []*SyncUserConversationResp {
+	t.Helper()
+	body := `{"msg_count":50,"device_uuid":"dev-repro-` + spaceID + `"}`
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("POST", "/v1/conversation/sync?space_id="+spaceID, strings.NewReader(body))
+	req.Header.Set("token", testutil.Token)
+	s.GetRoute().ServeHTTP(w, req)
+	require.Equal(t, http.StatusOK, w.Code, w.Body.String())
+
+	var wrap struct {
+		Conversations []*SyncUserConversationResp `json:"conversations"`
+	}
+	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &wrap))
+	return wrap.Conversations
+}
+
+func findConv(convs []*SyncUserConversationResp, channelID string) *SyncUserConversationResp {
+	for _, c := range convs {
+		if c.ChannelID == channelID {
+			return c
+		}
+	}
+	return nil
+}
+
+// TestRepro_SpaceLastMessage_DefaultSpaceLeak confirms the hypothesis end-to-end
+// through POST /v1/conversation/sync?space_id=.
+func TestRepro_SpaceLastMessage_DefaultSpaceLeak(t *testing.T) {
+	const (
+		dmChannel   = "dm-peer-repro"
+		spaceB      = "space-b-repro"       // a non-default space
+		spaceDefaul = "space-default-repro" // earliest membership → default space
+	)
+
+	now := time.Now()
+	// History: seq 11 = UNTAGGED (default-space message), seq 12 = tagged spaceB
+	// (globally latest). Matches the real capture where the latest msg is 668cc9.
+	untagged := dmMsgRepro(dmChannel, 11, now.Add(-2*time.Hour).Unix(),
+		`{"type":1,"content":"default-space-hello"}`)
+	taggedB := dmMsgRepro(dmChannel, 12, now.Add(-1*time.Hour).Unix(),
+		`{"type":1,"content":"1111","space_id":"`+spaceB+`"}`)
+
+	convs := []*config.SyncUserConversationResp{
+		dmIMConvMulti(dmChannel, now.Add(-1*time.Hour).Unix(), 100, []*config.MessageResp{untagged, taggedB}),
+	}
+
+	s, ctx := setupConvSyncE2E(t, convs)
+	// Default space MUST have the earlier created_at (GetUserDefaultSpaceIDE =
+	// earliest membership by created_at ASC).
+	seedSpaceMemberRepro(t, ctx, spaceDefaul, testutil.UID, "2020-01-01 00:00:00")
+	seedSpaceMemberRepro(t, ctx, spaceB, testutil.UID, "2020-06-01 00:00:00")
+
+	// ---- (1) NON-default space: preview is correct (explicitly tagged) ----
+	convB := findConv(callConvSyncSpace(t, s, spaceB), dmChannel)
+	require.NotNil(t, convB, "DM must be visible in its tagged (non-default) space")
+	require.NotNil(t, convB.SpaceLastMessage,
+		"non-default space: space_last_message present (tagged message found)")
+	assert.Equal(t, "1111", convB.SpaceLastMessage.Payload["content"],
+		"non-default space: preview is the spaceB-tagged message")
+	assert.Equal(t, spaceB, convB.SpaceLastMessage.Payload["space_id"])
+
+	// ---- (2) DEFAULT space: preview is WRONG (leaks the spaceB message) ----
+	convD := findConv(callConvSyncSpace(t, s, spaceDefaul), dmChannel)
+	require.NotNil(t, convD, "DM is (correctly) visible in the default space via catch-all")
+
+	// The bug: no per-space preview could be computed for the default space,
+	// because its messages are UNTAGGED and findSpaceLastMessage matches strictly
+	// on space_id == defaultSpaceID.
+	assert.Nil(t, convD.SpaceLastMessage,
+		"BUG: default-space space_last_message is nil — untagged default messages never match the strict space_id filter")
+
+	// With space_last_message absent, the client falls back to recents[last] =
+	// the channel's global-last message, which belongs to the OTHER space.
+	require.NotEmpty(t, convD.Recents, "recents present")
+	lastRecent := convD.Recents[len(convD.Recents)-1]
+	assert.Equal(t, "1111", lastRecent.Payload["content"],
+		"LEAK: default-space fallback preview is the spaceB message content")
+	assert.Equal(t, spaceB, lastRecent.Payload["space_id"],
+		"LEAK: default-space fallback preview carries the WRONG space_id (spaceB, not the default space)")
+}
+
+// TestRepro_FindSpaceLastMessage_StrictMatchBlindSpot isolates the root cause:
+// findSpaceLastMessage / the fallback both match strictly on payload.space_id ==
+// spaceID, so for the DEFAULT space (whose DM messages are UNTAGGED) they return
+// nil even when the default-space message IS in the scanned set. This is why the
+// 200-message fallback in findSpaceLastMessageFallback cannot rescue it either.
+func TestRepro_FindSpaceLastMessage_StrictMatchBlindSpot(t *testing.T) {
+	const spaceB = "space-b-repro"
+	recents := []*MsgSyncResp{
+		{MessageSeq: 11, Payload: map[string]interface{}{"type": float64(1), "content": "default-space-hello"}}, // UNTAGGED
+		{MessageSeq: 12, Payload: map[string]interface{}{"type": float64(1), "content": "1111", "space_id": spaceB}},
+	}
+
+	// Non-default space: the tagged message is found.
+	got := findSpaceLastMessage(recents, spaceB)
+	require.NotNil(t, got)
+	assert.Equal(t, uint32(12), got.MessageSeq)
+
+	// Default space: the untagged default message exists in the set but is NOT
+	// matched (no space_id key) → nil. Same strict predicate the fallback uses.
+	assert.Nil(t, findSpaceLastMessage(recents, "space-default-repro"),
+		"strict space_id match cannot find the untagged default-space message")
+}
--- a/modules/message/dm_space_last_message_repro_test.go
+++ b/modules/message/dm_space_last_message_repro_test.go
@@ -2,30 +2,27 @@
 
 package message
 
-// Reproduction for the default-Space DM conversation-preview leak observed on
-// the deployed test env (main) via
+// Regression guard for the default-Space DM conversation-preview leak observed
+// on the deployed test env (main) via
 //   POST /v1/conversation/sync?space_id=:space_id
 //
 // Scenario (mirrors the real capture): one physical DM channel whose history is
 //   - an UNTAGGED message  (no payload.space_id → belongs to the DEFAULT space)
 //   - a message TAGGED with a NON-default space (the globally-latest one)
 //
-// Hypothesis under test:
-//   1. Query the NON-default space  → space_last_message is that space's message
-//      (correct — it is explicitly tagged).
-//   2. Query the DEFAULT space      → space_last_message is ABSENT (nil), because
-//      findSpaceLastMessage matches strictly on payload.space_id == filterSpaceID
-//      and the default-space messages are UNTAGGED, so nothing matches. The client
-//      then falls back to recents[last] = the channel's GLOBAL last message, which
-//      is tagged for the OTHER space → the preview leaks a wrong-space message.
+// Behaviour after the fix (fillPersonSpaceUnread/findSpaceLastMessage now take
+// defaultSpaceID and treat an untagged DM message as a default-Space message):
+//   1. Query the NON-default space → space_last_message is that space's message
+//      (explicitly tagged); the untagged default message never leaks in.
+//   2. Query the DEFAULT space     → space_last_message is the untagged default
+//      message (NOT the other space's globally-latest one). Before the fix this
+//      was nil, so the client fell back to recents[last] and leaked a wrong-space
+//      preview.
 //
 // This drives the REAL handler (syncUserConversation) through the registered
 // /v1/conversation route (which mounts spacepkg.SpaceMiddleware, reading
 // ?space_id=). WuKongIM is the shared httptest fake from
-// conversation_recent_filter_e2e_test.go (same package/build tag): it returns
-// our canned conversation for /conversation/sync and {} for /channel/messagesync
-// (so the findSpaceLastMessageFallback path also finds no default-tagged message
-// — same strict-match blind spot, exercised below as a pure-function check).
+// conversation_recent_filter_e2e_test.go (same package/build tag).
 
 import (
 	"encoding/json"
@@ -117,9 +114,9 @@ func findConv(convs []*SyncUserConversationResp, channelID string) *SyncUserConv
 	return nil
 }
 
-// TestRepro_SpaceLastMessage_DefaultSpaceLeak confirms the hypothesis end-to-end
+// TestFix_SpaceLastMessage_DefaultSpaceUsesUntagged verifies the fix end-to-end
 // through POST /v1/conversation/sync?space_id=.
-func TestRepro_SpaceLastMessage_DefaultSpaceLeak(t *testing.T) {
+func TestFix_SpaceLastMessage_DefaultSpaceUsesUntagged(t *testing.T) {
 	const (
 		dmChannel   = "dm-peer-repro"
 		spaceB      = "space-b-repro"       // a non-default space
@@ -153,45 +150,46 @@ func TestRepro_SpaceLastMessage_DefaultSpaceLeak(t *testing.T) {
 		"non-default space: preview is the spaceB-tagged message")
 	assert.Equal(t, spaceB, convB.SpaceLastMessage.Payload["space_id"])
 
-	// ---- (2) DEFAULT space: preview is WRONG (leaks the spaceB message) ----
+	// ---- (2) DEFAULT space: preview is now the untagged default message ----
 	convD := findConv(callConvSyncSpace(t, s, spaceDefaul), dmChannel)
-	require.NotNil(t, convD, "DM is (correctly) visible in the default space via catch-all")
+	require.NotNil(t, convD, "DM is visible in the default space")
 
-	// The bug: no per-space preview could be computed for the default space,
-	// because its messages are UNTAGGED and findSpaceLastMessage matches strictly
-	// on space_id == defaultSpaceID.
-	assert.Nil(t, convD.SpaceLastMessage,
-		"BUG: default-space space_last_message is nil — untagged default messages never match the strict space_id filter")
-
-	// With space_last_message absent, the client falls back to recents[last] =
-	// the channel's global-last message, which belongs to the OTHER space.
-	require.NotEmpty(t, convD.Recents, "recents present")
-	lastRecent := convD.Recents[len(convD.Recents)-1]
-	assert.Equal(t, "1111", lastRecent.Payload["content"],
-		"LEAK: default-space fallback preview is the spaceB message content")
-	assert.Equal(t, spaceB, lastRecent.Payload["space_id"],
-		"LEAK: default-space fallback preview carries the WRONG space_id (spaceB, not the default space)")
+	// Fixed: the untagged message counts as a default-Space message, so the
+	// per-Space preview resolves to it instead of leaking the spaceB message.
+	require.NotNil(t, convD.SpaceLastMessage,
+		"default-space space_last_message is now populated from the untagged default message")
+	assert.Equal(t, "default-space-hello", convD.SpaceLastMessage.Payload["content"],
+		"default-space preview is the untagged (default) message, not the spaceB one")
+	_, hasSpaceID := convD.SpaceLastMessage.Payload["space_id"]
+	assert.False(t, hasSpaceID,
+		"the default-space preview message carries no space_id (it is the untagged one)")
 }
 
-// TestRepro_FindSpaceLastMessage_StrictMatchBlindSpot isolates the root cause:
-// findSpaceLastMessage / the fallback both match strictly on payload.space_id ==
-// spaceID, so for the DEFAULT space (whose DM messages are UNTAGGED) they return
-// nil even when the default-space message IS in the scanned set. This is why the
-// 200-message fallback in findSpaceLastMessageFallback cannot rescue it either.
-func TestRepro_FindSpaceLastMessage_StrictMatchBlindSpot(t *testing.T) {
-	const spaceB = "space-b-repro"
+// TestFix_FindSpaceLastMessage_DefaultMatchesUntagged pins the root-cause fix at
+// the helper level: findSpaceLastMessage now takes defaultSpaceID and treats an
+// untagged DM message as a default-Space message (the same predicate the 200-msg
+// fallback uses), while non-default queries stay strict.
+func TestFix_FindSpaceLastMessage_DefaultMatchesUntagged(t *testing.T) {
+	const (
+		spaceB       = "space-b-repro"
+		spaceDefault = "space-default-repro"
+	)
 	recents := []*MsgSyncResp{
 		{MessageSeq: 11, Payload: map[string]interface{}{"type": float64(1), "content": "default-space-hello"}}, // UNTAGGED
 		{MessageSeq: 12, Payload: map[string]interface{}{"type": float64(1), "content": "1111", "space_id": spaceB}},
 	}
 
-	// Non-default space: the tagged message is found.
-	got := findSpaceLastMessage(recents, spaceB)
+	// Non-default space: the tagged message is found; the untagged one never leaks.
+	got := findSpaceLastMessage(recents, spaceB, spaceDefault)
 	require.NotNil(t, got)
 	assert.Equal(t, uint32(12), got.MessageSeq)
 
-	// Default space: the untagged default message exists in the set but is NOT
-	// matched (no space_id key) → nil. Same strict predicate the fallback uses.
-	assert.Nil(t, findSpaceLastMessage(recents, "space-default-repro"),
-		"strict space_id match cannot find the untagged default-space message")
+	// Default space: the untagged default message is now matched (was nil before).
+	got = findSpaceLastMessage(recents, spaceDefault, spaceDefault)
+	require.NotNil(t, got, "untagged default-space message is now resolved")
+	assert.Equal(t, uint32(11), got.MessageSeq)
+
+	// Guard: with no default configured (defaultSpaceID==""), the untagged=default
+	// convention is off and a non-default query stays strict.
+	assert.Nil(t, findSpaceLastMessage(recents, spaceDefault, ""))
 }
--- a/modules/message/space_unread.go
+++ b/modules/message/space_unread.go
@@ -13,10 +13,16 @@ import (
 // 并填充 SpaceLastMessage（该 Space 的最后一条消息预览）。
 // 仅处理 channelType=1 的会话。
 // 通过解析消息 payload 中的 space_id 字段来统计属于指定 Space 的未读消息数。
+//
+// defaultSpaceID 用于「无标签 = 默认 Space」归属（见 dmSpaceMatch）：DM 在默认
+// Space 发送的消息不带 payload.space_id，若严格等值匹配就永远命中不到 —— 会导致
+// 默认 Space 的 space_last_message 取不到、per-Space 未读漏计，与可见性兜底
+// (decideConvKeepInSpace) / 历史过滤 (filterPersonMessagesBySpace) 口径不一致。
 func fillPersonSpaceUnread(
 	conversations []*SyncUserConversationResp,
 	rawConversations []*config.SyncUserConversationResp,
 	spaceID string,
+	defaultSpaceID string,
 	loginUID string,
 	ctx *config.Context,
 ) {
@@ -36,7 +42,7 @@ func fillPersonSpaceUnread(
 		}
 
 		// 从 Recents 中找该 Space 的最后一条消息作为预览
-		spaceLastMsg := findSpaceLastMessage(conv.Recents, spaceID)
+		spaceLastMsg := findSpaceLastMessage(conv.Recents, spaceID, defaultSpaceID)
 		if spaceLastMsg != nil {
 			conv.SpaceLastMessage = spaceLastMsg
 		}
@@ -49,7 +55,7 @@ func fillPersonSpaceUnread(
 			if raw != nil && raw.LastMsgSeq > 0 {
 				fallbackMsg := findSpaceLastMessageFallback(
 					conv.ChannelID, conv.ChannelType,
-					loginUID, spaceID, uint32(raw.LastMsgSeq), ctx,
+					loginUID, spaceID, defaultSpaceID, uint32(raw.LastMsgSeq), ctx,
 				)
 				if fallbackMsg != nil {
 					conv.SpaceLastMessage = fallbackMsg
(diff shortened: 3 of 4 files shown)


OctoBoooot (commented):
## Review: fix(message): default-Space DM conversation preview & unread count (#532)

A correct, bidirectionally-sound fix with exactly the right test coverage. Not stamping a clean APPROVE only because `Test`/`Vet` haven't reported. `check-sprint` red is governance (doesn't bar approval); merge stays BLOCKED until code CI + check-sprint clear.

Risk tier: normal, reviewed on the #484 Space-isolation axis (the fix changes a match predicate — the bidirectional-swap trap).

### The bug and the fix (byte-verified both directions)
- **Bug:** `findSpaceLastMessage`/unread matched `payload.space_id == filterSpaceID` **strictly**. Default-Space DM messages are untagged (`space_id==""`), so a default-Space query matched nothing → `space_last_message` fell back to the channel global-last (possibly another Space's) = **preview leak**, and per-Space unread under-counted.
- **Fix:** centralized `dmSpaceMatch(msgSpaceID, targetSpaceID, defaultSpaceID)`: `msgSpaceID==target || (msgSpaceID=="" && target==defaultSpaceID)`. Constructed all four cases:
  - untagged + default query → **match** (the fix);
  - untagged + non-default query → **no match** (the inverse — untagged does NOT leak into a non-default Space; this is the over-match trap, correctly avoided);
  - tagged-B + query-B → match; tagged-A + query-B → no match.
  - `defaultSpaceID==""` guard: second clause can only fire when `target==""`, which callers never pass → strict-equal only, no accidental match.
- **Consistency win:** the "untagged = default Space" convention now aligns the preview/unread path with the visibility path (`decideConvKeepInSpace` / `filterPersonMessagesBySpace`), which already use it — closing the real inconsistency where the preview surfaced a message the visibility filter would hide.
- `defaultSpaceID` at the `fillPersonSpaceUnread` call (`api_conversation.go:891`) is the same value already resolved via `GetUserDefaultSpaceIDE` in the `hasSpaceFilter` block — correctly threaded, no new/stale lookup.

### Coverage
`TestFix_SpaceLastMessage_DefaultSpaceUsesUntagged` asserts **both** directions: non-default Space B preview = the tagged-B message with *"the untagged default message never leaks in"*, and default-Space preview = the untagged message (not spaceB's), confirming `space_id` absent on the default preview. Plus an HTTP-level repro test and helper-level unit tests. The one "uncovered" file (`api_conversation.go`) is the 1-line caller wiring, exercised by the repro test's HTTP path.

### Praise
- Extracting `dmSpaceMatch` as one predicate — rather than repeating the "== spaceID || (untagged && default)" logic across `findSpaceLastMessage`, the WuKongIM fallback, and `countSpaceUnreadFromMessages` — is what makes this correct *and* consistent: all three consumers now share one attribution rule, so preview and unread can't drift from each other or from the visibility filter. And the test asserting the *inverse* ("untagged never leaks into non-default") is exactly the direction a match-predicate change most often gets wrong — good instinct to pin it.

Jerry-Xin (dismissed):
**APPROVE** — Correct, tightly-scoped fix for the default-Space DM preview/unread mis-attribution, with no cross-Space or cross-user leak.

## Summary
The defect: DM messages sent in the **default Space carry no `payload.space_id`** (untagged). The preview (`space_last_message`) and per-Space unread count matched `space_id == filterSpaceID` **strictly**, so a default-Space query matched nothing → `space_last_message` came back nil, the client fell back to `recents[last]` (a message tagged for *another* Space) and leaked a wrong-Space preview; unread was under-counted. The fix introduces `dmSpaceMatch(msgSpaceID, targetSpaceID, defaultSpaceID)` implementing the "untagged = default Space" convention, threaded through `fillPersonSpaceUnread → findSpaceLastMessage / findSpaceLastMessageFallback / countSpaceUnreadFromMessages`.

## ✅ Data-visibility trace (both directions — the recurring #519/#527/#529 lesson)
Traced all three `dmSpaceMatch` call sites, not just one:

- **WIDEN / correctness**: default-Space query now resolves the untagged message as the preview and counts it toward default-Space unread. Verified via unit (`TestFindSpaceLastMessage_DefaultSpaceMatchesUntagged`, `TestCountSpaceUnread_DefaultSpaceCountsUntagged`) + e2e integration (`TestFix_SpaceLastMessage_DefaultSpaceUsesUntagged`) driving the real `/v1/conversation/sync?space_id=` route.
- **LEAK (critical)**: `dmSpaceMatch` only lets an untagged message match when `msgSpaceID=="" && targetSpaceID==defaultSpaceID`. A tagged message still matches only its exact Space. So:
  - Untagged content **cannot** surface in a **non-default** target Space (second branch requires `targetSpaceID==defaultSpaceID`) — non-default queries stay strict.
  - The untagged→default attribution only fires when the queried Space *is the login user's own default Space* — and `SpaceMiddleware` (`pkg/space/middleware.go`) **membership-validates** `filterSpaceID` (`CheckMembership`, 403 otherwise), so a user cannot even query a Space they aren't in.
  - No cross-user leak: preview/count are computed over the **login user's own** conversation stream (`conv.Recents`); the fallback (`findSpaceLastMessageFallback`) calls `IMSyncChannelMessage` scoped to `LoginUID` + that DM `channelID`.

## ✅ Robust space-scoping
- `defaultSpaceID` is resolved **server-side** from the authenticated `loginUID` via `space.GetUserDefaultSpaceIDE`, fail-closed (500 on DB error) — not a client-supplied/spoofable field.
- `filterSpaceID` comes from the membership-guarded middleware, not raw client input.

## ✅ Degenerate inputs
- `spaceID==""` → early `return` (no-op).
- `defaultSpaceID==""` → convention disabled, reverts to prior strict equality (tested: `findSpaceLastMessage(recents,"spaceDefault","")==nil`).
- nil/empty payload → `dmMessageSpaceID` returns `""`, iteration `continue`s; empty/nil recents & messages → returns nil / count 0 (tested).
- deleted messages (`IsDeleted==1`) skipped in fallback.

## ✅ Migration check
No migration files added/renamed/deleted (changed files: `api_conversation.go`, `space_unread.go` + two test files). No startup-panic risk.

## CI
Build **pass**, pr-title-lint **pass**, all governance/security checks green. `Lint`/`Test`/`Vet` are **pending** (path-filtered/queued, 0s) — not RED; stated as a caveat, they don't gate logic correctness here. `mergeStateStatus: BLOCKED` is the missing-approval gate + `MERGEABLE` (no conflicts).

## Highlights
- Correctly aligns preview/unread with the existing visibility catch-all (`decideConvKeepInSpace`) and history filter (`filterPersonMessagesBySpace`) — consistent "untagged = default" semantics across the code paths.
- Isolation preserved: strict for non-default Spaces; convention self-disables when no default resolves.
- Strong test coverage: pure-function unit + real-handler e2e regression guard.

Caveat: `Test`/`Vet` not shown green (pending/path-filtered) — approving on code-trace + the included unit/integration tests.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #532 (octo-server)

## Summary

This PR fixes the default-Space DM conversation preview and unread count leak (#484). When a DM channel's globally-latest message is tagged with a non-default Space, the default-Space query for that channel used to return `space_last_message: nil`, causing the client to fall back to the recents list and surface a wrong-space preview. The fix introduces `dmSpaceMatch` — an "untagged = default Space" convention — and threads `defaultSpaceID` through `fillPersonSpaceUnread`, `findSpaceLastMessage`, `findSpaceLastMessageFallback`, and `countSpaceUnreadFromMessages`.

As a bonus, the PR adds a `StickerCustomEnabled` appconfig toggle with clean schema/getter/API wiring and solid test coverage.

The core fix is correct and well-targeted. One P2 consistency observation below.

## Verification

Static analysis only at head `c9273ce`; build and tests not executed in this environment.

- ✅ **Data flow: `defaultSpaceID` propagation** — `GetUserDefaultSpaceIDE` at `api_conversation.go:834` returns earliest membership by `created_at ASC`; fail-closed (returns HTTP error on DB failure). Passed through to `fillPersonSpaceUnread` at `:891`, then into `findSpaceLastMessage` / `countSpaceUnreadFromMessages` / `findSpaceLastMessageFallback`.
- ✅ **`dmSpaceMatch` semantics** — Correctly returns `true` when `msgSpaceID == targetSpaceID` OR when `msgSpaceID == "" && targetSpaceID == defaultSpaceID`. Guard: when `defaultSpaceID == ""`, the untagged convention is off — verified by test `TestFix_FindSpaceLastMessage_DefaultMatchesUntagged`.
- ✅ **Message-level filter alignment** — `filterPersonMessagesBySpace` in `space_filter.go` already implements the same untagged=default convention for non-system-bot messages. The fix makes preview/unread consistent with the message-history path.
- ✅ **Conversation-level visibility** — `decideConvKeepInSpace` uses `hasSpaceMsg` callback, which correctly shows the conversation when at least one message has `space_id == targetSpaceID`. For default-Space queries, `EnsureSystemBotsPresent` handles system-bot presence separately.
- ✅ **StickerCustomEnabled wiring** — Schema registration (`settingTypeBool`), getter (`getBool` with `false` default), API response field (`json:"sticker_custom_enabled"`), and both response branches (version-short-circuit + full) are covered. Tests verify default-false, explicit-true, and short-circuit paths.
- ✅ **Test coverage** — New regression test `dm_space_last_message_repro_test.go` drives the real handler end-to-end, reproducing the exact production capture (untagged + tagged-B history → default-Space query resolves to untagged). Unit tests updated with `defaultSpaceID=""` for backward compatibility.

## Findings

No P0/P1 issues; one P2 item below.

### P2 — `dmSpaceMatch` does not gate on system-bot channels (`modules/message/space_unread.go:135`)

`dmSpaceMatch` treats every untagged message as belonging to the default Space, but `filterPersonMessagesBySpace` in `space_filter.go:546` explicitly drops system-bot untagged messages (`msid == "" && isSysBot → drop`). This means for a system-bot conversation queried from the default Space, `findSpaceLastMessage` and `countSpaceUnreadFromMessages` will match and count untagged bot messages, while the message-history path hides them.

This is not a regression — before the fix, the default-Space preview was `nil` (worse). Conversation-level visibility already handles system bots via `EnsureSystemBotsPresent`. The inconsistency is cosmetic and low-risk.

**Suggestion:** A one-line comment on `dmSpaceMatch` noting the intentional asymmetry with `filterPersonMessagesBySpace` for system-bot channels would prevent future maintainers from accidentally unifying them into a regression.

- `GetUserDefaultSpaceIDE` query: `ORDER BY created_at ASC LIMIT 1` with `status=1` — correct for "earliest active membership" semantics.
- `api_conversation.go:834` fail-closed error handling — if `GetUserDefaultSpaceIDE` fails, the endpoint returns an error, preventing silent misrouting.
- `personConvHasSpaceMessages` at `space_filter.go:437` still uses strict equality (not `dmSpaceMatch`), but this is harmless — conversation-level visibility is about "does this conv have ANY message in this Space", and the conversation is correctly shown/hidden based on explicit tags.
- `findSpaceLastMessageFallback` (200-msg API path) uses the same `dmSpaceMatch` predicate — consistent with the recents path.
- New test `TestCountSpaceUnread_DefaultSpaceCountsUntagged` locks the unread-count parity with the preview logic.

## Verdict: APPROVED

The core fix correctly resolves the default-Space DM preview leak by introducing the "untagged = default Space" convention and threading `defaultSpaceID` through all relevant helpers. Test coverage is solid (unit + integration regression). The P2 observation about system-bot asymmetry is cosmetic and pre-existing in spirit. The `StickerCustomEnabled` addition is cleanly wired.

OctoBoooot (commented):
## Follow-up on mochashanyao's #532 🟡 — byte-verified: real consistency gap (minor)

mochashanyao's 🟡#1 is **correct** — I byte-verified it. `dmSpaceMatch` (preview/unread) does not mirror the history path's system-bot exclusion, so the two drift for untagged system-bot DM messages in the default Space. Confirming it, sizing it as a minor, and updating my verdict note. Still no blocker/major; my verdict stays COMMENT (conditional on Test).

### The gap (verified)
- **History path** — `filterPersonMessagesBySpace` (space_filter.go): `case msid=="" && isSysBot → continue` (drops). Untagged system-bot history is hidden even in the default Space.
- **Preview/unread path** — `dmSpaceMatch` (this PR): no `isSysBot` awareness; an untagged message matches the default Space regardless of whether the peer is a system bot.
- **Net divergence:** for a system-bot DM (botfather / fileHelper / u_10000) in the **default Space**, its untagged messages are **hidden from history** but **counted in per-Space unread and surfaced as `space_last_message`**. So the preview/badge can show a message the history view won't display — the exact preview-vs-history inconsistency this PR set out to close, just on the sysbot sub-case.

### Severity: minor (not blocking)
- Blast radius is narrow: untagged system-bot messages, default Space only. It's a cosmetic preview/badge-vs-history mismatch on system-bot conversations, **not** a cross-Space or cross-user leak — the Space scoping in `dmSpaceMatch` is still correct; this is a sysbot-visibility-convention mismatch between two paths.
- But it's a real gap in *this PR's own* alignment claim: the rationale says `dmSpaceMatch` aligns preview/unread with `filterPersonMessagesBySpace`; it aligns on the untagged-default rule but **misses that path's sysbot exclusion**. Worth closing so the two paths are actually consistent.

### Direction (author's call)
Either thread `isSysBot := spacepkg.IsSystemBot(channelID)` into `dmSpaceMatch`/its callers and drop untagged sysbot messages (mirror the history path), OR — if the intended product behavior is that system-bot previews *should* show untagged default-Space messages (arguably more useful for botfather) — then `filterPersonMessagesBySpace` is the side that's over-hiding and the two should converge the other way. Pick one and make both paths match; mochashanyao's 🟡#2 (a comment stating the sysbot design intent) follows once that's decided. Add a test for the sysbot-default-Space case in whichever direction.

### Unchanged
The core fix (untagged=default for regular DMs, bidirectional-correct, no non-default leak) is still sound and well-tested. Credit to mochashanyao for catching the sysbot sub-case my review didn't separate out — I verified the untagged/tagged × default/non-default matrix but didn't split system-bot from regular DM in the preview path. Verdict stays COMMENT (conditional on Test), now with this minor logged.

yujiawei (changes requested):
# Code Review — PR #532 (octo-server)

Reviewed at head `c9273ce006c938c14f0324403e79a69d7c2fc00e` against merge-base `1592d353`. Scoped diff: 4 files, +309/-42, all under `modules/message/`. Build of `./modules/message/` passes; all `findSpaceLastMessage` / `findSpaceLastMessageFallback` / `countSpaceUnreadFromMessages` / `fillPersonSpaceUnread` call sites are updated.

## 1. Spec compliance

**Spec: ✅**

- Missing: none. The stated goal — attribute an untagged DM message to the user's default Space so a default-Space `/v1/conversation/sync` resolves `space_last_message` / `SpaceUnread` to the untagged (default) message instead of falling back to the channel's global-last (wrong-Space) message — is implemented via `dmSpaceMatch` and threaded through the three helpers plus the call site.
- Over-build: none. The change is self-contained; the convention is disabled when `defaultSpaceID == ""` (falls back to prior strict behavior), and non-default-Space queries stay strict (untagged never leaks into a non-default Space). No new flags/fields/endpoints.
- Deviation: none on the primary approach — the "untagged = default Space" convention matches the agreed direction.

## 2. Code quality

### P1 — Preview/unread diverges from message-history for untagged system-bot DMs

`dmSpaceMatch(msgSpaceID, targetSpaceID, defaultSpaceID)` (`modules/message/space_unread.go:130`) matches purely on `space_id` and the default-Space fallback. It has no channel-identity input, so it cannot apply the system-bot exception that the message-history filter enforces.

`filterPersonMessagesBySpace` (`modules/message/space_filter.go:526`) has five rules; rule 4 (`space_filter.go:549`) explicitly **drops** untagged messages for system bots (`botfather` / `u_10000` / `fileHelper` / `notification`, via `spacepkg.IsSystemBot`) **even in the default Space**, to avoid legacy system-bot history leaking across Spaces. Rule 2 (`space_filter.go:541`) keeps untagged messages only for **non**-system-bot DMs in the default Space.

After this PR, on a default-Space `/v1/conversation/sync`:
- `findSpaceLastMessage` (`space_unread.go:145`), `findSpaceLastMessageFallback` (`space_unread.go:197`), and `countSpaceUnreadFromMessages` (`space_unread.go:236`) all treat an untagged **system-bot** message as a default-Space message.
- So the conversation list can show a system-bot preview and a non-zero unread badge for messages that `/v1/message/channel/sync` then hides from the opened conversation.

`decideConvKeepInSpace` keeps system-bot Person conversations visible (`space_filter.go:350`), so this path is reachable, and untagged system-bot history is exactly the high-volume case rule 4 was written for (fileHelper, notifications). Net user-visible effect: a phantom unread count on system-bot DMs in the default Space that the user cannot clear by reading, plus a preview that doesn't exist in history.

Pre-PR this did not happen — the old strict matcher returned nothing for untagged messages, so system bots were unaffected. This PR introduces the divergence for the system-bot case while fixing the regular-DM case.

The code comment at `space_unread.go:126-128` claims parity with `filterPersonMessagesBySpace`; that claim only holds for non-system-bot DMs today.

Suggested fix: thread the channel identity into the matcher (or wrap it at the call sites) so `dmSpaceMatch` returns the default-Space fallback branch only when `!spacepkg.IsSystemBot(channelID)`, mirroring rules 2/4. All three helpers already have `conv.ChannelID` / `channelID` in scope. Add a test with a `fileHelper`/`u_10000` channel to lock the parity.

### P2 (advisory — non-blocking)

- **nil / non-JSON payload is skipped, not treated as untagged.** `findSpaceLastMessage` (`space_unread.go:141`), the fallback loop (`space_unread.go:194`), and `countSpaceUnreadFromMessages` (`space_unread.go:233`) `continue` when the payload is nil or unmarshals to nil. `extractPayloadSpaceID` in the history filter instead maps an empty payload to `""` (kept in the default Space for non-bots). This is a minor consistency gap for degenerate payloads. Note: the nil-skip behavior is unchanged from pre-PR (not a regression) and a fully-nil DM payload is not a normal message, hence advisory rather than blocking.
- **`readSeq = LastMsgSeq - Unread` approximation.** `space_unread.go:76` infers the read boundary and `space_unread.go:229` counts `seq > readSeq`. This is correct only if WuKongIM unread equals a contiguous sequence distance; if unread excludes own/silent/deleted messages it can miscount. Pre-existing; the default-Space change now also exercises it for untagged DMs. Worth an authoritative read-seq or a documented invariant/test.
- **200× `json.Unmarshal` on the fallback path.** `findSpaceLastMessageFallback` unmarshals every one of up to 200 pulled messages just to read `space_id`. Pre-existing, but the new default-Space path makes the fallback more likely to run. A lightweight `space_id` extraction would cut allocs.

## 3. Overall verdict

**REQUEST_CHANGES** — Spec ✅ but Quality has a P1 (system-bot preview/unread parity). The primary fix is correct and well-tested for regular DMs; the blocking issue is the missing system-bot carve-out that produces phantom unread / phantom preview on system-bot DMs in the default Space, contradicting the parity the change documents.

## 4. Suggestions

- Add `spacepkg.IsSystemBot(channelID)` gating to the default-Space fallback branch of the matcher (thread `channelID` in) and a regression test using a system-bot channel with untagged history, asserting preview/unread stay empty in the default Space.
- Optionally align nil/empty-payload handling with `extractPayloadSpaceID` and add a `readSeq` invariant test.

## 5. Additional notes

- The fallback pull window `[lastMsgSeq-199, lastMsgSeq]` with `Limit: 200`, newest-to-oldest scan, skipping `IsDeleted == 1`, is coherent.
- `defaultSpaceID` at the call site (`api_conversation.go:834`) comes from `GetUserDefaultSpaceIDE` and is fail-closed (500 on DB error) before `fillPersonSpaceUnread` runs, so the new argument is reliably populated.
- Independent second-opinion tooling was run on this diff; the P1 above was reproduced independently, and the nil-payload / perf items surfaced as advisory. All findings were re-verified against the source before this verdict.

Jerry-Xin (commented):
## Correcting my prior review — real (minor) preview-vs-history inconsistency confirmed

I'm dismissing my earlier APPROVE because it rested on an incorrect rationale. On re-verification at `c9273ce006c9`, @mochashanyao's and @Octo-PR's system-bot concern is **factually correct** — the preview/unread path and the history path do **not** agree for untagged system-bot DM messages in the default Space.

### Byte-verified divergence

**History path** — `filterPersonMessagesBySpace` (space_filter.go). No early-return; per-message switch:
- `case msid=="" && !isSysBot && spaceID==defaultSpaceID:` → **keep** (untagged regular DM in default Space)
- `case msid=="" && isSysBot:` → `continue` → **DROP**, and this is **unconditional** (not gated to non-default Space). Because the keep-case requires `!isSysBot`, an untagged system-bot message can never match it even when `spaceID==defaultSpaceID`; it falls to the sysbot case and is dropped.

**Preview/unread path** — `space_unread.go` `dmSpaceMatch(msgSpaceID, targetSpaceID, defaultSpaceID)`:
- `return msgSpaceID=="" && targetSpaceID==defaultSpaceID` → **true** for the same message.
- The file has **zero** system-bot awareness (`IsSystemBot`/`isSysBot`/`SystemBots` appear nowhere); all three callers (`findSpaceLastMessage`, `findSpaceLastMessageFallback`, `countSpaceUnreadFromMessages`) run without an `isSysBot` guard, on raw convs from `api_conversation.go`.

**Net:** for a system-bot DM (botfather/u_10000/fileHelper) with untagged messages in the user's default Space → **history DROPS them, but preview SHOWS them and unread COUNTS them.** The conversation list renders a `space_last_message` preview + non-zero unread badge for messages the history view won't display.

### Severity — my assessment: minor, non-blocking

This is **not** a data leak, data loss, cross-Space, or cross-user issue — it's confined to the user's own default-Space system-bot DM. The failure mode is a **cosmetic badge/preview-vs-history mismatch** (user sees "unread" + a preview line, opens the conversation, history shows nothing new). Self-consistent within one user; no security boundary crossed.

So I read this as a legitimate **minor 🟡**, and I respectfully disagree with escalating it to a merge-blocking 🔴 — a pure preview/badge cosmetic mismatch with no leak and no loss shouldn't block merge. But it is real and worth fixing.

**Suggested fix:** give the `dmSpaceMatch` callers in `space_unread.go` the same untagged-system-bot exclusion the history path applies (an `isSysBot` guard, or reuse `filterPersonMessagesBySpace`'s rule), so unread/preview and history stay consistent for system-bot DMs. Adding a system-bot channel test case (as @yujiawei suggests) is a good idea.

Leaving this as COMMENT: the production fix's core goal (default-Space untagged **regular** DM preview/unread) is correct and both-directions leak-safe as I verified earlier; the open item is this system-bot preview/history consistency nit.

yujiawei (dismissed):
# Code Review — PR #532 (octo-server) — Revised (head c9273ce)

This supersedes my earlier `CHANGES_REQUESTED`. On re-verification, the earlier P1 was overstated; the system-bot asymmetry is a P2, and the PR is **APPROVED**.

## Why the earlier P1 was downgraded

My earlier blocker rested on "a phantom unread count on system-bot DMs in the default Space that the user cannot clear by reading." Re-checking the code, that justification does not hold:

- **Unread self-clears.** `fillPersonSpaceUnread` short-circuits at `conv.Unread <= 0` (`space_unread.go:67`) and computes `readSeq = raw.LastMsgSeq - raw.Unread` (`:76`). Once the conversation is read in WuKongIM, `SpaceUnread` is never populated. The count clears on read like any other unread; it is not stuck.
- **No cross-Space leak.** `dmSpaceMatch` (`space_unread.go:130`) returns true for a non-default target only on an exact `space_id` tag match; an untagged message maps to the default Space and nowhere else. Rule 4 in `filterPersonMessagesBySpace` (`space_filter.go:549`) exists to stop *cross-Space* system-bot exposure — the default Space is the user's own home Space, not a cross-Space target, so that boundary is not crossed here.
- **Strictly better than pre-fix.** Before this change, a default-Space query returned `space_last_message: nil`, so the client fell back to `recents[last]` — the channel's global-last message, tagged for another Space — which was the actual wrong-Space preview leak. This PR removes that.

## Remaining observation — P2 (advisory, non-blocking)

`dmSpaceMatch` has no system-bot carve-out, so for a system-bot DM (`botfather` / `u_10000` / `fileHelper` / `notification`) queried from the default Space, `findSpaceLastMessage` / `findSpaceLastMessageFallback` / `countSpaceUnreadFromMessages` can surface or count an untagged message that the message-history filter (`filterPersonMessagesBySpace` rule 4) hides on open. The preview axis (`space_last_message`, set at `space_unread.go:44-48` independent of unread) is the more visible residue, since unread self-clears. This is a cosmetic preview/history mismatch confined to system-bot DMs in the user's own default Space — low risk, and net-better than the pre-fix nil preview.

Suggestion (optional, not blocking merge): either add the `spacepkg.IsSystemBot(channelID)` gate to the default-Space fallback branch to fully match rule 4, or add a one-line comment on `dmSpaceMatch` documenting the intentional asymmetry so a future maintainer doesn't unify the two paths into a regression. This aligns with the first-pass reviewer's read.

## Spec compliance — ✅

- Missing: none. The default-Space preview/unread attribution via `dmSpaceMatch` + `defaultSpaceID` threading is implemented end to end.
- Over-build: none. Convention disabled when `defaultSpaceID == ""`; non-default queries stay strict. No unrequested flags/fields/endpoints.
- Deviation: none.

## Code quality — Approved

- Data flow verified: `GetUserDefaultSpaceIDE` (`api_conversation.go:834`) is fail-closed (HTTP error on DB failure) before `fillPersonSpaceUnread` (`:891`); the argument is threaded correctly to all three helpers.
- Fallback pull window `[lastMsgSeq-199, lastMsgSeq]`, `Limit: 200`, newest-to-oldest, skipping `IsDeleted == 1`, is coherent and uses the same `dmSpaceMatch` predicate as the recents path.
- Test coverage is solid: unit tests updated for the new signature, `TestFix_*` integration test drives the real `/v1/conversation/sync` handler and reproduces the production capture; `TestCountSpaceUnread_DefaultSpaceCountsUntagged` locks unread/preview parity.

Two pre-existing, non-blocking notes carried over: `readSeq = LastMsgSeq - Unread` assumes unread equals a contiguous seq distance (worth a documented invariant); the fallback path unmarshals up to 200 payloads just to read `space_id` (minor alloc cost).

## Overall verdict: APPROVED

Spec ✅ and Quality Approved. The fix correctly resolves the default-Space DM preview leak; the one remaining item (system-bot preview/history asymmetry) is a low-risk P2 that is net-better than the prior behavior and can be handled in a follow-up.

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD c9273ce0

## Reviewer Summary

|---|---|---|
| qa-engineer | pass-with-risk | Core default-Space DM preview/unread fix is correct & strongly tested; one system-bot consistency gap (S-1) + hot-module churn are non-blocking risks. |
| security-engineer | cleared | Read-only Space-attribution change; non-default Space isolation preserved, `defaultSpaceID` server-resolved & fail-closed — no new trust boundary, cross-Space/cross-user leak, or injection surface. |
| code-reviewer | needs-discussion | Fix is correct & well-isolated for the general DM case; the PR's stated alignment target `filterPersonMessagesBySpace` has a system-bot exclusion branch `dmSpaceMatch` does not replicate — a design-intent question, not a data-safety bug. |

## Critical Findings (P0-P1)

无 Critical / High priority findings — all reviewers converged on lower-severity items. Zero Must-Fix / blocking / security findings across QA, Security, and Code.

## Should-Fix (P2)

- [P2-1] `modules/message/space_unread.go` — `dmSpaceMatch` (≈L131-142) / `fillPersonSpaceUnread` (L39-41) — **System-bot exclusion gap.** The new "untagged DM = default Space" convention is applied to *all* Person channels, but the PR's own stated alignment target `filterPersonMessagesBySpace` (`space_filter.go:549`) has a 4th branch — `msid == "" && isSysBot → hide` — that `dmSpaceMatch` does not replicate. Because `fillPersonSpaceUnread` runs over all Person channels including system bots (botfather / u_10000 / fileHelper), a system-bot conversation in the **default Space** can now surface a `space_last_message` preview / `SpaceUnread` badge derived from an untagged message that the message-history view still suppresses → preview/badge vs. conversation-content inconsistency. Scope is limited to the querying user's own default-Space system-bot rows; **no** cross-Space or cross-user leak (Security concurs). — flagged by qa/S-1 + code/S-1 — **Fix:** either reuse `spacepkg.IsSystemBot(conv.ChannelID)` in `fillPersonSpaceUnread` (or thread it into `dmSpaceMatch`) to skip untagged sysbot attribution, mirroring `filterPersonMessagesBySpace` branch 4, plus a regression test; **or** if surfacing untagged sysbot history in the default-Space preview is intentional, state that explicitly in the PR description and open a follow-up to reconcile the message-history filter.

## Suggestions (P3)

- [P3-1] Maintainability: the "untagged = default Space" rule now lives in three places (`decideConvKeepInSpace`, `filterPersonMessagesBySpace`, `dmSpaceMatch`), each with subtly different system-bot handling — the divergence that produced P2-1. A shared predicate (or shared `isSysBot` guard) would prevent future drift. (flagged by code, Architecture note)
- [P3-2] Observability: no metric/log records how often the default-Space preview resolves to an untagged message, so a future regression of this attribution would only surface via client bug reports. Consider a lightweight counter as follow-up. (flagged by qa/D-2)

## Questions for Author

- [Q-1] For a system bot (botfather / u_10000 / fileHelper) in the **default** Space: should the conversation-list preview & unread reflect its untagged historical messages (this PR's new behavior), or stay hidden to match `filterPersonMessagesBySpace`? The answer decides whether P2-1 is a fix or a documented, intentional carve-out. (raised independently by qa/Q-1 + code/Q-1)

## Verdict Rationale

Findings tally after dedupe: **P0=0, P1=0, P2=1, P3=2, Questions=1.** Per the deterministic verdict rule, no P0/P1 with ≤2 P2 findings plus at least one material Question → `NEEDS_DISCUSSION`.

All three reviewers converged: the core fix is correct, isolation-preserving (non-default Space stays strict; `defaultSpaceID==""` disables the convention as a safe fallback), and strongly tested (exact-value unit assertions + an end-to-end integration guard through the real `/v1/conversation/sync` handler), with CI green and no tool findings on the changed lines. Security cleared the change outright and explicitly deferred the sole open item (P2-1) to the functional lane as a non-security consistency question. QA and Code independently reached the identical S-1 finding and the identical author question, so there is no reviewer disagreement to reconcile — only a genuine design-intent decision (should untagged system-bot history appear in the default-Space preview?) that warrants an author answer before merge rather than a hard block.

## Round Context

---

lml2468 (commented):
**NEEDS_DISCUSSION** — Round 1 (HEAD c9273ce0)

Core default-Space DM preview/unread fix is correct, isolation-preserving, and strongly tested (Must-Fix=0, Security cleared). One open item: `dmSpaceMatch` doesn't replicate the system-bot exclusion in its stated alignment target `filterPersonMessagesBySpace`, so untagged system-bot messages may surface in the default-Space preview/unread while the message-history view hides them — a design-intent question for the author, not a data-safety bug (no cross-Space/cross-user leak).

Full findings + the author question are in the aggregate comment on this PR.

an9xyz:
Addressed the system-bot parity item (P2-1 / Q-1) in `5c02a4b`.

**Answer to Q-1:** system-bot untagged history should **stay hidden** in the default-Space preview/unread, matching `filterPersonMessagesBySpace` rule 4 — not surface it. Making the preview/unread path consistent with the history filter is the whole point of this PR; showing a preview/badge for messages the opened conversation hides would just reintroduce the mismatch on the system-bot sub-case.

**Fix:** threaded `isSysBot := spacepkg.IsSystemBot(conv.ChannelID)` (resolved once per conv in `fillPersonSpaceUnread`) into `dmSpaceMatch` and its three callers. The untagged→default fallback branch now fires only for non-system-bot DMs:

```go
func dmSpaceMatch(msgSpaceID, targetSpaceID, defaultSpaceID string, isSysBot bool) bool {
    if msgSpaceID == targetSpaceID {
        return true
    }
    return !isSysBot && msgSpaceID == "" && targetSpaceID == defaultSpaceID
}
```

This mirrors rules 2/4 exactly:
- untagged + non-sysbot + default query → match (rule 2);
- untagged + sysbot → never match (rule 4);
- explicitly-tagged messages still match their Space for everyone;
- regular DMs and non-default-Space queries unchanged.

**Tests added:**
- `TestFillPersonSpaceUnread_SystemBotUntaggedExcluded` (unit) — `botfather` untagged → no default-Space preview, `SpaceUnread==0`; a regular DM with the same untagged messages still gets them.
- `TestFix_SystemBotUntaggedNotInDefaultPreview` (integration) — drives the real `/v1/conversation/sync?space_id=<default>` for a `botfather` DM and asserts `space_last_message==nil` + no unread badge, while the conversation stays visible.

All three helpers now share one attribution predicate, so preview, unread, and history stay consistent for both regular and system-bot DMs.

---

Jerry-Xin (approved):
Re-review on new head `5c02a4b0d84` — the system-bot preview/unread-vs-history inconsistency I raised in my prior COMMENT is now fixed correctly, and the original #532 regular-DM fix is intact. **APPROVE.**

## What changed (byte-verified delta `c9273ce0` → `5c02a4b0`)
Linear (ahead_by=1, behind_by=0 — no rebase). 3 files, all in `modules/message/`, no migration touched:
- `space_unread.go` — threaded an `isSysBot` guard through the preview/unread path.
- `space_unread_test.go` + `dm_space_last_message_repro_test.go` — signature updates + 2 new sysbot regression tests.

## 🔴 Blocking items
None.

## The fix — both paths now agree
1. **Preview/unread path** (`space_unread.go`): `dmSpaceMatch` gained an `isSysBot` param; the untagged=default fallback is now `return !isSysBot && msgSpaceID == "" && targetSpaceID == defaultSpaceID` (space_unread.go:158). `isSysBot` is resolved once per conv in `fillPersonSpaceUnread` via `spacepkg.IsSystemBot(conv.ChannelID)` (space_unread.go:47) and threaded into `findSpaceLastMessage`, `findSpaceLastMessageFallback`, and `countSpaceUnreadFromMessages`.
2. **History path** (`space_filter.go`) is byte-unchanged: `git diff c9273ce0 5c02a4b0 -- modules/message/space_filter.go` is empty. Rule 4 still drops untagged sysbot unconditionally (`case msid == "" && isSysBot: continue`, space_filter.go:549), keyed on the same `spacepkg.IsSystemBot(channelID)` (space_filter.go:530).

Both paths now use the identical, non-spoofable fixed-allowlist check `spacepkg.IsSystemBot` (query.go:44 → `SystemBots` map: botfather/u_10000/fileHelper/notification). Consistency matrix now holds on both paths:
- untagged **regular** DM, default Space → **INCLUDE** (both) ✓ — #532's original fix preserved (`!isSysBot` is true).
- untagged **sysbot** DM, default Space → **EXCLUDE** (both) ✓ — the new fix.
- untagged (either), non-default Space → EXCLUDE ✓ (no leak).
- tagged, matching Space → INCLUDE ✓.

## ✅ Highlights
- Fix is surgical — one boolean threaded through, no behavior change for the regular-DM path that #532 fixed.
- Tests added for exactly the gap yujiawei/OctoBoooot/I flagged: `TestFillPersonSpaceUnread_SystemBotUntaggedExcluded` asserts botfather untagged → nil SpaceLastMessage + 0 SpaceUnread, while a regular peer with identical untagged messages still gets preview + unread=2. Also `TestFix_SystemBotUntaggedNotInDefaultPreview` at the E2E conv-sync level.
- Guard on `defaultSpaceID == ""` (convention off) preserved.

_Note: Build/Test/Vet/Lint still pending at review time (CI queued, not red); pr-title-lint, check-sprint, secret-scan, dependency-review all green. Logic is byte-verified independent of the pending Test job._

OctoBoooot (commented):
## Re-review (delta c9273ce → 5c02a4b): #532 — sysbot minor resolved

The system-bot consistency minor from my prior review is fixed correctly and completely. Delta is test + `space_unread.go` only. Holding APPROVE only until `Test` reports green (#508); `check-sprint`/Build/pr-title-lint green.

### The minor is resolved (byte-verified)
`isSysBot := spacepkg.IsSystemBot(conv.ChannelID)` is now threaded through all three preview/unread call sites (`findSpaceLastMessage`, the WuKongIM fallback, `countSpaceUnreadFromMessages`) into `dmSpaceMatch`:
```go
if msgSpaceID == targetSpaceID { return true }
return !isSysBot && msgSpaceID == "" && targetSpaceID == defaultSpaceID
```
Re-verified all directions:
- regular untagged + default query → match (original fix preserved);
- **sysbot untagged + default query → no match** (the fix — now mirrors the history path's `msid=="" && isSysBot → drop`, so preview/unread and history are consistent);
- sysbot message *explicitly tagged* for the queried Space → still matches via the exact-match clause (correct: the history exclusion is untagged-only, so a tagged sysbot message legitimately shows);
- untagged + non-default query → no match (no cross-Space leak, preserved).

So the preview/badge no longer surfaces an untagged system-bot default-Space message the history view hides — the exact inconsistency mochashanyao/yujiawei/I converged on. `TestFix_SystemBotUntaggedNotInDefaultPreview` (botfather: `SpaceLastMessage==nil`, `SpaceUnread==0`) is the regression guard.

### Everything else unchanged
The core default-Space untagged=default fix and its bidirectional coverage (from the prior head) are logically unchanged; api_conversation.go already threaded `defaultSpaceID`. No new concerns.

### Praise
- Fixing it by threading `isSysBot` into the single `dmSpaceMatch` predicate (not patching each call site) keeps the "one attribution rule, three consumers" property intact — preview, unread, and history now all agree on both the untagged=default convention *and* the sysbot exception. And keeping the exact-tag-match clause ahead of the sysbot check is the right nuance: it excludes only *untagged* sysbot messages (matching the history path), not tagged ones.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #532 (octo-server)

## Summary

This PR fixes the default-Space DM conversation preview and unread-count leak
tracked in issue #484. Before the fix, `fillPersonSpaceUnread` and its helpers
used strict `payload.space_id == spaceID` matching, which meant untagged DM
messages (no `space_id` in payload) never matched any Space — causing the
default Space's `space_last_message` to be nil and per-Space unread to
undercount. The fix introduces `dmSpaceMatch`, a centralized matching
predicate that aligns preview/unread with the existing history-filter
(`filterPersonMessagesBySpace` rule 2/4) and visibility (`decideConvKeepInSpace`)
contracts. A second commit (5c02a4b) adds a system-bot exception: system bots'
untagged history is hidden from message-level sync, so preview/unread must
also exclude it to prevent ghost unread badges.

## Verification

- ✅ **Data-flow: `defaultSpaceID`** — sourced from `GetUserDefaultSpaceIDE` at
  `api_conversation.go:834` with fail-closed error handling (returns 500 on DB
  error, never proceeds with `""`). The same `defaultSpaceID` is already used
  by `fillConversationSpaceIDs` and `buildSpaceMemberships`; this PR adds it
  to `fillPersonSpaceUnread` at `:891`.
- ✅ **Data-flow: `isSysBot`** — computed per-conversation via
  `spacepkg.IsSystemBot(conv.ChannelID)` at `space_unread.go:49`, using the
  same static `SystemBots` map as `filterPersonMessagesBySpace`
  (`space_filter.go:533`).
- ✅ **Predicate alignment** — `dmSpaceMatch` at `space_unread.go:141-147`
  encodes exactly two true-branches: (1) explicit `msgSpaceID == targetSpaceID`,
  and (2) untagged + non-sysbot + default-space. This is a faithful
  re-encoding of `filterPersonMessagesBySpace` rules 1+2 (keep) vs 3+4+5
  (drop).
- ✅ **Payload extraction parity** — `dmMessageSpaceID` (`space_unread.go:118-126`)
  returns `""` for nil payload, missing key, or non-string type, semantically
  equivalent to `extractPayloadSpaceID` in `space_filter.go:565-579`.
- ✅ **Fallback path consistency** — `findSpaceLastMessageFallback` at
  `space_unread.go:196` passes the same `(spaceID, defaultSpaceID, isSysBot)`
  to `dmSpaceMatch`, ensuring the WuKongIM 200-message fallback uses the same
  matching rules as the Recents-window path.
- ✅ **Test coverage** — new integration test
  `dm_space_last_message_repro_test.go` drives the real `/v1/conversation/sync`
  route with `?space_id=`, covering both the default-space fix and the
  system-bot exception. Unit tests in `space_unread_test.go` are updated for
  the new signatures with `defaultSpaceID=""` to verify strict-match fallback.
- ✅ **Caller propagation** — `fillPersonSpaceUnread`, `findSpaceLastMessage`,
  `findSpaceLastMessageFallback`, and `countSpaceUnreadFromMessages` all gain
  new parameters; the sole production call site at `api_conversation.go:891`
  is updated. Compiler enforcement guarantees no stale callers survive.

Static analysis only at head `5c02a4b0d8441562fc3c7bec4e2bcaa4d62a99fd`; build

## Findings

- **`dmSpaceMatch` edge cases**: `defaultSpaceID==""` makes the fallback
  branch unreachable (since `targetSpaceID==""` is guarded by early return),
  providing safe strict-match-only behavior when default space cannot be
  determined.
- **System bot with tagged messages**: a system bot message carrying an
  explicit `space_id` still matches via the first `dmSpaceMatch` branch —
  correctly showing in that Space's preview.
- **Conversation-level vs message-level consistency**: system bots are visible
  in all Spaces at conversation level (`decideConvKeepInSpace` returns true
  for `SystemBots[channelID]`), while their untagged messages are excluded at
  message level — this is intentional (conversation entry present, content
  scoped to Space).
- **C1 — Dual-path parity**: `filterPersonMessagesBySpace` (history) and
  `dmSpaceMatch` (preview/unread) use the same rule set. Verified rule-by-rule
  above.
- **C2 — Control-flow ordering**: `fillPersonSpaceUnread` is called before
  `FilterConversationsBySpace` at `api_conversation.go:891-894`, ensuring
  per-Space metadata is computed on the full conversation set before filtering
  removes non-matching entries. No ordering issue.
- **C3/C4 — Authorization**: no auth/permission changes in this PR. Space
  membership is validated upstream by SpaceMiddleware.
- **C5 — Build/runtime**: no build/config changes; pure Go logic.
- **C6 — Governance**: N/A.

## Verdict: APPROVED

Clean, well-scoped fix that correctly aligns the preview/unread pipeline with
the existing history-filter and visibility contracts. The centralized
`dmSpaceMatch` predicate eliminates the prior strict-equality gap for
untagged DM messages in the default Space while adding a system-bot exception
to prevent ghost unread badges. Thorough integration and unit test coverage.

yujiawei (approved):
# Code Review — PR #532 (octo-server)

**Head SHA reviewed:** `5c02a4b0d8441562fc3c7bec4e2bcaa4d62a99fd` (merge-base `68e3860`)

## Verdict: APPROVED

The fix is correct, well-scoped, and consistent with the surrounding Space-filter design. No blocking (P0/P1) issues. A few non-blocking coverage/maintainability suggestions below.

---

## 1. Spec compliance

The linked defect (#484 facet): a **default-Space** `POST /v1/conversation/sync?space_id=<default>` computed `space_last_message` and per-Space unread by strict `payload.space_id == filterSpaceID`, so **untagged** DM messages (which belong to the default Space) matched nothing → `nil` preview (client fell back to the channel's global-last message from another Space) and under-counted unread.

- **Under-build:** none. Both symptoms (wrong preview + under-count) are addressed at the single authoritative path `fillPersonSpaceUnread` (`modules/message/api_conversation.go:891`).
- **Over-build:** none. The change threads `defaultSpaceID` + `isSysBot` through the three existing helpers and nothing else; no new flags/fields/endpoints.
- **Deviation:** none. `dmSpaceMatch` (`modules/message/space_unread.go:141-146`) maps exactly onto the message-history filter `filterPersonMessagesBySpace` rules 2/4 (`modules/message/space_filter.go:537-557`) — untagged ⇒ default only for non-system-bots; untagged system-bot ⇒ dropped. The second commit (system-bot exclusion) closes the one real parity gap: without it, an untagged system-bot DM would surface a default-Space preview/badge that `/v1/message/channel/sync` then hides.

**Spec: ✅**

## 2. Code quality

- **Correctness (verified):** all three callers — `findSpaceLastMessage`, `findSpaceLastMessageFallback`, `countSpaceUnreadFromMessages` — receive the same `(spaceID, defaultSpaceID, isSysBot)`. `isSysBot = spacepkg.IsSystemBot(conv.ChannelID)` is computed once and uses the bare peer UID, the same predicate the history endpoint feeds (`api.go:1370`).
- **`defaultSpaceID == ""` is safe:** the handler aborts on the `GetUserDefaultSpaceIDE` DB error (`api_conversation.go:834-839`) and `fillPersonSpaceUnread` early-returns when `spaceID == ""` (`space_unread.go:30`). So the untagged-fallback branch (which requires `targetSpaceID == defaultSpaceID == ""`) can never fire spuriously; it degrades to strict equality exactly as documented — non-default queries are unaffected.
- **Isolation preserved:** an untagged message matches only when `targetSpaceID == defaultSpaceID`; a non-default-Space query never picks up untagged content, and an explicitly cross-tagged message never matches a default query.
- **Build/tests:** `go build ./modules/message/`, `go vet`, and the targeted unit suite (incl. the `SystemBotUntaggedExcluded` regression guard) all pass locally.

**Quality: Approved**

---

## Non-blocking suggestions (P2 / nit — do not block merge)

1. **(P2, test coverage)** The **200-msg WuKongIM fallback** (`findSpaceLastMessageFallback`, `space_unread.go:208`) and the **IM unread-pull branch** (`space_unread.go:88-113`, taken when `len(raw.Recents) < raw.Unread`) both call the new `dmSpaceMatch` predicate but are never fed real messages in tests — the shared fake IM returns `{}` for `/channel/messagesync`, so a regression in the default-Space/system-bot logic *on those paths specifically* would pass silently. The core predicate itself is well covered on the Recents fast path, so this is a hardening gap, not a live defect. The `/channel/messagesync` mock pattern already exists in `dm_cross_space_repro_test.go:86-91` and could be reused to seed a >Recents-window history.

2. **(P2, cross-endpoint consistency — pre-existing, not this PR)** `/v1/sidebar/sync` (`api_sidebar.go`) has only a single global `Unread` and no `space_last_message`/`space_unread`, so it shows whole-channel unread in every Space tab. The per-Space fix here applies only to `/v1/conversation/sync`. Worth tracking separately if any client renders DM badges from the sidebar endpoint while inside a Space.

3. **(P2, client contract — pre-existing)** `conv.Recents` is shipped unfiltered; only `space_last_message`/`space_unread` are Space-attributed. The server half of the preview/history closure is correct, but full end-to-end suppression of the untagged system-bot "ghost preview" depends on clients treating `space_last_message == nil` as "no preview" rather than falling back to `recents[last]`. This is inherited behavior (untagged non-default DMs already relied on it since #484), not introduced here. Consider documenting the contract, or optionally space-filtering `conv.Recents` for Person channels server-side as future hardening.

4. **(nit)** `dmMessageSpaceID` (`space_unread.go:120`) is behaviorally identical to `extractPayloadSpaceID` (`space_filter.go:564`). Since this PR's whole point is rule-2/4 parity between the two paths, having `dmMessageSpaceID` delegate to `extractPayloadSpaceID` (single source of truth) would prevent silent future divergence if one is ever changed to accept a non-string `space_id`.

lml2468:
# Formal Verdict Aggregate Report — Round 2 — HEAD 5c02a4b0

**Formal Verdict**: `APPROVE`



## Reviewer Summary

|---|---|---|
| qa-engineer | pass | Round 1's sole open item (S-1 system-bot exclusion gap) is fully resolved; `dmSpaceMatch` now threads `isSysBot` and drops untagged system-bot attribution with unit + integration regression guards. Must=0, Should=0. |
| security-engineer | cleared | Narrowing-only change — untagged system-bot DM history is now excluded from the default-Space preview/unread; non-default Space isolation preserved, attribution condition strictly tighter than R1, no new trust boundary or leak. |
| code-reviewer | approved | S-1/P2-1 closed correctly and minimally; `dmSpaceMatch`'s new `!isSysBot` branch has exact semantic parity with `filterPersonMessagesBySpace` rules 2/4 (verified branch-by-branch), strong contrast + E2E tests. Must=0, Should=0, Nit=0. |

## Critical Findings (P0-P1)

无 Critical / High priority findings — Zero Must-Fix / blocking / security findings across QA, Security, and Code.

## Should-Fix (P2)

无 Should-Fix findings — Round 1's sole P2 (system-bot exclusion gap) is resolved this round; all three reviewers independently confirmed closure.

## Suggestions (P3)

- [P3-1] Maintainability (standing follow-up, carried from Round 1): the untagged-DM convention now lives in three places (`decideConvKeepInSpace`, `filterPersonMessagesBySpace`, `dmSpaceMatch`), with two distinct payload-extractors (`dmMessageSpaceID` vs `extractPayloadSpaceID`). Round 2 aligned the *semantics* rather than deduplicating the code — an acceptable scoping decision for a fix PR. A shared predicate would prevent future drift. Non-blocking. (flagged by code, Maintainability note)
- [P3-2] Observability (standing follow-up, carried from Round 1 D-2): no metric/log records how often the default-Space preview resolves to an untagged message or excludes a system bot, so a future regression would only surface via client bug reports. Consider a lightweight counter. Non-blocking. (flagged by qa/D-2)

## Questions for Author

无 — Round 1's Q-1 (expected behavior of untagged system-bot messages in the default-Space preview) is answered by this implementation: **hidden**, consistent with the message-history view (rule-4 parity). No open questions.

## Verdict Rationale

Findings tally after dedupe: **P0=0, P1=0, P2=0, P3=2 (both standing non-blocking follow-ups), Questions=0.** Per the deterministic verdict rule, no P0/P1/P2 with only P3 items and no blocking question → `APPROVE`.

All three reviewers converged with no disagreement. Round 1's single open item — the system-bot exclusion gap (S-1/P2-1), where `dmSpaceMatch` did not replicate the `isSysBot`→hide carve-out of its stated alignment target `filterPersonMessagesBySpace` — is fully closed in this round's commit ("exclude untagged system-bot DM messages from default-Space preview/unread"). The fix resolves channel identity once per conversation (`spacepkg.IsSystemBot(conv.ChannelID)` in `fillPersonSpaceUnread`) and threads it uniformly into all three attribution sites (find / fallback / count), adding a `!isSysBot` precondition to the untagged-default branch. Code verified exact semantic parity with `filterPersonMessagesBySpace` rules 2/4 branch-by-branch; the change is narrowing-only (can only reduce attribution hits, never open a new leak), and regular-DM / non-default-Space behavior is provably unchanged. Regression coverage was added at three levels: a unit contrast test (`TestFillPersonSpaceUnread_SystemBotUntaggedExcluded` — sysbot preview=nil/unread=0 vs regular-peer preview≠nil/unread=2), and an end-to-end integration guard (`TestFix_SystemBotUntaggedNotInDefaultPreview` driving the real `/v1/conversation/sync` handler for `botfather`). `go vet` clean, targeted unit tests pass, CI green (incl. `Test` re-polled to pass and `code-review` approved). The two remaining P3 items are optional follow-ups carried from Round 1 and explicitly non-blocking.

## Round Context

Round 2 aggregate — Round 1 aggregate posted 2026-07-03T07:44:15Z with verdict `NEEDS_DISCUSSION` (1 Should-Fix P2-1 + Q-1). Delta c9273ce0 (R1) → 5c02a4b0 (R2): 1 commit, 3 files (+119/-39), remediating the P2-1 system-bot exclusion gap. R1's P2-1 and Q-1 are both resolved this round; the two P3 follow-ups (shared-predicate refactor, observability metric) carry forward as non-blocking.

---
_Generated by review-leader · chain terminal (Round 2 aggregate) · dedupe/verdict deterministic per skill_

lml2468 (approved):
**APPROVE** — Round 2 (HEAD 5c02a4b0)

Round 1's sole open item (system-bot exclusion gap, P2-1) is fully closed: `dmSpaceMatch` now threads `isSysBot` and achieves exact parity with `filterPersonMessagesBySpace` rules 2/4, with unit contrast + end-to-end integration regression guards. All three lanes converged — QA `pass`, Security `cleared`, Code `approved` (Must=0, Should=0, Nit=0). Only two optional, non-blocking P3 follow-ups remain.

Full findings are in the aggregate comment on this PR.

PR #533, opened 2026-07-03
feat: add sticker collect endpoint

an9xyz:
## Summary

Add a server-side sticker collect endpoint so clients can add a sticker-message path to the current user's custom stickers without re-uploading the file.

## Related Issue

Manual request.

## Linked Spec

`.octospec/tasks/sticker-collect/brief.md`

## Changes

- Add `POST /v1/sticker/user/collect` under the authenticated sticker routes.
- Accept existing `sticker/{source_uid}/{file}` paths, normalize them to `file/preview/sticker/...`, and infer the sticker format from the source path.
- Add `source_path` and `source_path_hash` metadata plus a live-only unique key for idempotency.
- Enforce the existing `sticker.user_max_count` quota only for new collect records; repeated collects return the existing sticker.
- Add low-cardinality collect metrics for success, idempotent hits, invalid paths, quota rejection, shortcode conflicts, query failures, and store failures.
- Document first-version trade-offs: reference-based storage, path-only source validation, and future GC/copy considerations.

## Testing

- [x] Manually verified

Commands run:

```bash
docker run --rm --network host -e GIN_MODE=release \
  -v /Users/fangling/conductor/workspaces/octo-server/kyoto:/src \
  -v /Users/fangling/conductor/workspaces/octo-server/kyoto/.context/go-cache/pkg:/go/pkg \
  -v /Users/fangling/conductor/workspaces/octo-server/kyoto/.context/go-cache/build:/root/.cache/go-build \
  -w /src golang:1.25 go test ./modules/sticker

git diff --check origin/main...HEAD
```

## COMPREHENSION

It adds a separate collect path for custom stickers. The existing upload registration endpoint remains owner-bound and handle-aware. The new collect endpoint accepts only reserved sticker object paths, records a source hash for idempotency, and uses the existing per-user quota transaction/lock before inserting new rows.

Clients may now receive a mix of existing upload paths and normalized `file/preview/sticker/...` paths in `GET /v1/sticker/user`; older clients must be able to render preview paths. Because this version stores references instead of copying objects, future object-store GC or source-user cleanup could create dangling collected stickers unless follow-up copy/reference-counting is added. The path-only contract also does not prove source message access.

`go test ./modules/sticker` covers foreign-path collection, idempotent repeat collection, quota behavior, soft-delete then re-collect, invalid source paths, source-path parsing/hash stability, and collect metrics. `git diff --check origin/main...HEAD` is clean.

## Checklist

- [x] Updated documentation

### DIFF
--- a/.octospec/tasks/sticker-collect/brief.md
+++ b/.octospec/tasks/sticker-collect/brief.md
@@ -1,0 +1,53 @@
+---
+type: Task
+title: "Task: sticker-collect"
+description: Add a server-side endpoint for collecting sticker-message paths into a user's custom sticker list.
+tags: ["sticker", "api", "database", "quota"]
+timestamp: 2026-07-03T00:00:00Z
+# --- octospec extension fields ---
+slug: sticker-collect
+upstream: "manual"
+source: self
+---
+
+# Task: sticker-collect
+
+## Goal
+Add `POST /v1/sticker/user/collect` so a client can add a sticker from an
+existing sticker message into the caller's personal sticker list without
+re-uploading the file.
+
+## Background
+The existing `POST /v1/sticker/user` endpoint is intentionally upload-oriented:
+it only accepts the caller's own `sticker/{uid}/...` path and may require an
+upload handle. Sticker-message collection has a different source: a sticker path
+already present in a message payload, often under another user's UID segment.
+
+This first version stores a reference to the source sticker object instead of
+copying bytes into the collecting user's directory. The current file service
+does not expose a storage-backend-neutral object-copy abstraction across
+MinIO/S3/COS/OSS/Qiniu.
+
+## Load-bearing list
+- Authenticated `/v1/sticker` routes keep `AuthMiddleware` and shared UID rate limiting.
+- Per-user quota still uses `sticker.user_max_count` and the existing user-row lock.
+- Repeated collection of the same source path by the same UID is idempotent.
+- Only reserved `sticker/{source_uid}/{file}.{gif|png|jpg|jpeg|webp}` paths are accepted.
+- `sticker.handle_required` remains an upload-registration policy for `POST /v1/sticker/user`; collect intentionally has no handle parameter.
+
+## Out of scope
+- Verifying that the caller can read the exact source message that carried the sticker path.
+- Server-side object copy or reference counting.
+- Object existence HEAD/check before insert.
+- Garbage collection semantics for collected references.
+
+## Known Trade-Offs
+- Because collect stores a reference, future object-store GC or source-user cleanup could create dangling collected stickers unless reference counting or copy-on-collect is added.
+- Source sticker deletion currently soft-deletes only DB rows and does not delete the object; collected references depend on that retention behavior.
+- Path-only source validation does not prove message access. A follow-up API can accept message identity and derive the sticker path server-side.
+
+## Acceptance
+- Same source path collected twice by the same user returns the existing sticker and does not consume quota twice.
+- Soft-deleting a collected sticker releases the live idempotency slot and allows re-collecting it.
+- Invalid/non-sticker paths are rejected through the localized sticker error envelope.
+- Collect success, idempotent hit, invalid path, quota, query, and store outcomes are observable through low-cardinality metrics.
--- a/modules/sticker/api.go
+++ b/modules/sticker/api.go
@@ -339,28 +339,40 @@ func (s *Sticker) add(ctx *wkhttp.Context) {
 // caller; it must still point at the reserved sticker object keyspace. The
 // stored path is a stable authenticated preview URL for that source object, and
 // SourcePathHash makes repeat taps idempotent without charging quota again.
+//
+// collect intentionally does not require the upload handle even when
+// sticker.handle_required=true: the endpoint is for a sticker URL already
+// carried in a message, and the only accepted source is the reserved sticker/
+// keyspace that file upload protects from non-sticker writes. With the current
+// path-only contract this is a collect-specific trust boundary, not a general
+// replacement for add()'s same-user upload-handle proof.
 func (s *Sticker) collect(ctx *wkhttp.Context) {
 	loginUID := ctx.GetLoginUID()
 
 	var req collectStickerReq
 	if err := ctx.BindJSON(&req); err != nil {
+		observeStickerCollect("validation_failed")
 		respondStickerRequestInvalid(ctx, "")
 		return
 	}
 	if req.Path == "" {
+		observeStickerCollect("path_invalid")
 		respondStickerRequestInvalid(ctx, "path")
 		return
 	}
 	if len(req.Path) > maxStickerCollectPathLen {
+		observeStickerCollect("path_invalid")
 		respondStickerRequestInvalid(ctx, "path")
 		return
 	}
 	source, ok := parseCollectStickerSourcePath(req.Path)
 	if !ok {
+		observeStickerCollect("path_invalid")
 		respondStickerRequestInvalid(ctx, "path")
 		return
 	}
 	if len(source.DisplayPath) > maxStickerPathLen || len(source.SourceKey) > maxStickerPathLen {
+		observeStickerCollect("path_invalid")
 		respondStickerRequestInvalid(ctx, "path")
 		return
 	}
@@ -369,20 +381,24 @@ func (s *Sticker) collect(ctx *wkhttp.Context) {
 		placeholder = defaultStickerPlaceholder
 	}
 	if len([]rune(placeholder)) > maxStickerPlaceholderLen {
+		observeStickerCollect("validation_failed")
 		respondStickerRequestInvalid(ctx, "placeholder")
 		return
 	}
 	if req.Sort < 0 {
+		observeStickerCollect("validation_failed")
 		respondStickerRequestInvalid(ctx, "sort")
 		return
 	}
 	shortcode, ok := normalizeStickerShortcode(req.Shortcode)
 	if !ok {
+		observeStickerCollect("validation_failed")
 		respondStickerShortcodeInvalid(ctx)
 		return
 	}
 	keywordsStore, _, ok := normalizeStickerKeywords(req.Keywords)
 	if !ok {
+		observeStickerCollect("validation_failed")
 		respondStickerKeywordsInvalid(ctx)
 		return
 	}
@@ -393,6 +409,7 @@ func (s *Sticker) collect(ctx *wkhttp.Context) {
 	tx, err := s.ctx.DB().Begin()
 	if err != nil {
 		s.Error("开启事务失败", zap.Error(err))
+		observeStickerCollect("store_failed")
 		httperr.ResponseErrorL(ctx, errcode.ErrStickerStoreFailed, nil, nil)
 		return
 	}
@@ -400,6 +417,7 @@ func (s *Sticker) collect(ctx *wkhttp.Context) {
 
 	if err := s.db.lockUserRowTx(tx, loginUID); err != nil {
 		s.Error("锁定用户行失败", zap.Error(err))
+		observeStickerCollect("store_failed")
 		httperr.ResponseErrorL(ctx, errcode.ErrStickerStoreFailed, nil, nil)
 		return
 	}
@@ -407,23 +425,18 @@ func (s *Sticker) collect(ctx *wkhttp.Context) {
 	existing, err := s.db.queryByUIDAndSourcePathHashTx(tx, loginUID, sourceHash)
 	if err != nil {
 		s.Error("查询已收藏贴纸失败", zap.Error(err), zap.String("uid", loginUID))
+		observeStickerCollect("query_failed")
 		httperr.ResponseErrorL(ctx, errcode.ErrStickerQueryFailed, nil, nil)
 		return
-	}
-	if existing == nil {
-		existing, err = s.db.queryByUIDAndPathTx(tx, loginUID, source.DisplayPath)
-		if err != nil {
-			s.Error("按 path 查询已收藏贴纸失败", zap.Error(err), zap.String("uid", loginUID))
-			httperr.ResponseErrorL(ctx, errcode.ErrStickerQueryFailed, nil, nil)
-			return
-		}
 	}
 	if existing != nil {
 		if err := tx.Commit(); err != nil {
 			s.Error("提交事务失败", zap.Error(err))
+			observeStickerCollect("store_failed")
 			httperr.ResponseErrorL(ctx, errcode.ErrStickerStoreFailed, nil, nil)
 			return
 		}
+		observeStickerCollect("idempotent_hit")
 		ctx.Response(toStickerResp(existing))
 		return
 	}
@@ -431,20 +444,24 @@ func (s *Sticker) collect(ctx *wkhttp.Context) {
 	count, err := s.db.countByUIDTx(tx, loginUID)
 	if err != nil {
 		s.Error("查询贴纸数量失败", zap.Error(err))
+		observeStickerCollect("query_failed")
 		httperr.ResponseErrorL(ctx, errcode.ErrStickerQueryFailed, nil, nil)
 		return
 	}
 	if count >= max {
+		observeStickerCollect("quota_exceeded")
 		respondStickerQuotaExceeded(ctx, max)
 		return
 	}
 	conflict, err := s.db.shortcodeExistsTx(tx, loginUID, shortcode, "")
 	if err != nil {
 		s.Error("查询贴纸 shortcode 冲突失败", zap.Error(err), zap.String("uid", loginUID))
+		observeStickerCollect("query_failed")
 		httperr.ResponseErrorL(ctx, errcode.ErrStickerQueryFailed, nil, nil)
 		return
 	}
 	if conflict {
+		observeStickerCollect("shortcode_conflict")
 		respondStickerShortcodeConflict(ctx)
 		return
 	}
@@ -464,15 +481,18 @@ func (s *Sticker) collect(ctx *wkhttp.Context) {
 	}
 	if err := s.db.insertTx(tx, m); err != nil {
 		s.Error("收藏贴纸失败", zap.Error(err), zap.String("uid", loginUID), zap.String("source_path", source.SourceKey))
+		observeStickerCollect("store_failed")
 		httperr.ResponseErrorL(ctx, errcode.ErrStickerStoreFailed, nil, nil)
 		return
 	}
 	if err := tx.Commit(); err != nil {
 		s.Error("提交事务失败", zap.Error(err))
+		observeStickerCollect("store_failed")
 		httperr.ResponseErrorL(ctx, errcode.ErrStickerStoreFailed, nil, nil)
 		return
 	}
 
+	observeStickerCollect("success")
 	s.Info("贴纸收藏成功",
 		zap.String("uid", loginUID),
 		zap.String("source_path", source.SourceKey),
--- a/modules/sticker/api.go
+++ b/modules/sticker/api.go
@@ -43,6 +43,7 @@ import (
 // oversized value gets a clean 400 instead of a DB truncation/error.
 const (
 	maxStickerPathLen        = 512
+	maxStickerCollectPathLen = 2048
 	maxStickerPlaceholderLen = 100
 	// defaultStickerPlaceholder is stored when the client sends no placeholder,
 	// so conversation digests / push notifications have a sensible fallback.
@@ -98,6 +99,7 @@ func (s *Sticker) Route(r *wkhttp.WKHttp) {
 	{
 		auth.GET("/user", s.list)
 		auth.POST("/user", s.add)
+		auth.POST("/user/collect", s.collect)
 		auth.PUT("/user/:sticker_id", s.update)
 		auth.DELETE("/user/:sticker_id", s.delete)
 	}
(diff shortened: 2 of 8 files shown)


OctoBoooot (commented):
## Review: feat: add sticker collect endpoint (#533)

A well-built new endpoint in a security-sensitive module (#510 sticker keyspace). Byte-verified the trust boundary and migration; one minor (defense-in-depth). Not stamping a clean APPROVE only because `Test`/`Vet` haven't reported. `check-sprint` green.

Risk tier: normal, but reviewed as a new authenticated endpoint that accepts an attacker-supplied object path pointing at another user's sticker namespace.

### What I byte-verified
- **Trust boundary is sound.** `parseCollectStickerSourcePath` → `parseCollectStickerObjectKey` confines input to the `sticker/{uid}/{file}.{ext}` keyspace via `^sticker/([^/]+)/([^/]+)\.([A-Za-z0-9]+)$`. Single-segment `[^/]+` captures reject nested/multi-slash paths (`.../nested/x.png` → rejected, tested); format allowlist (`isAllowedStickerFormat`) + length caps (2048 raw, maxStickerPathLen on key). Query/fragment stripped, URL host stripped. Cross-user collect is **by design** (documented: collecting a shared sticker's stable authenticated preview URL, not an IDOR — these URLs already appear in messages).
- **Migration is clean.** INFORMATION_SCHEMA-guarded reentrant procedure, additive columns w/ defaults, reversible Down (drops index→columns in reverse). The `source_path_hash_live` STORED generated column (NULL unless status=1 AND hash≠'') backing `UNIQUE (uid, source_path_hash_live)` is a neat way to enforce collect-idempotency only on live rows while leaving direct uploads (empty hash → NULL, non-colliding) unaffected.
- **Handler safeguards present:** per-user quota (`StickerUserMaxCount` + `countByUIDTx`), idempotency (`queryByUIDAndSourcePathHashTx` → return existing without charging quota), `lockUserRowTx` row lock, shortcode-conflict check. Good hostile-input test coverage (non-sticker URL, bad/missing extension, nested path all → request_invalid + metric).

### Minor
- `parseCollectStickerObjectKey` does **not** reject `..` in the uid/file segments: `sticker/../foo.png` matches the regex and is stored as `SourceKey`/`DisplayPath` verbatim. It's **not exploitable** — the key is an inert stored string here, and the sole consumer (`file/preview/*path` → `getFile`, file/api.go:980-982) already rejects `..` via `filepath.Clean`+contains-check, so a traversal preview URL fails at fetch time. Still, rejecting `..` (or `path.Clean`-ing) at the collect ingress would be cleaner defense-in-depth than relying solely on the downstream guard, and would keep obviously-malformed keys out of the DB. Non-blocking.

### Praise
- The `source_path_hash_live` generated-column trick is the elegant part: it makes "repeat collect of the same source is idempotent, but re-collect after delete creates a fresh row, and direct uploads never collide" fall out of a single partial unique index (NULLs don't collide in MySQL), instead of application-level dedup logic that could race. And treating the collect path as an explicit "collect-specific trust boundary" with its own regex + format allowlist — rather than reusing add()'s uid-bound validation — correctly reflects that collect legitimately crosses users while still confining to the sticker keyspace.

Jerry-Xin (approved):
In-scope PR for octo-server: it extends the existing sticker module with an authenticated collect endpoint. I found no blocking issues.

💬 Non-blocking

🟡 Warning — `parseCollectStickerSourcePath` accepts any absolute URL whose path tail matches `/sticker/{uid}/{file}` and then stores a normalized internal preview path. See `modules/sticker/model.go:190` and `modules/sticker/api.go:472`. This matches the documented path-only v1 trade-off, but it is broader than “a sticker path already present in a message”; a caller can persist a guessed object key without message access or object existence proof. The planned message-derived/source-validation follow-up should be prioritized.

🔵 Suggestion — The idempotency lookup filters on `uid`, `status`, and `source_path_hash` (`modules/sticker/db.go:88`), while the new unique key is on `uid, source_path_hash_live` (`modules/sticker/sql/20260703000001_sticker_collect.sql:31`). With the current quota size this is likely fine, but querying `source_path_hash_live` directly or adding a matching lookup index would better align the read path with the migration.

✅ Highlights

- The new route stays inside the authenticated sticker route group with the shared UID rate limiter (`modules/sticker/api.go:98`).
- The collect path checks existing records before quota enforcement, so idempotent repeats do not consume quota (`modules/sticker/api.go:425`).
- Errors continue through the localized sticker error helpers rather than raw responses.
- The live-only generated-column unique key correctly releases idempotency after soft delete.
- Added tests cover collect idempotency, quota behavior, re-collect after delete, invalid paths, metrics, and source hash stability.

Verification: pure model/path tests passed with `go test ./modules/sticker -run 'TestValidateStickerPath|TestParseCollectStickerSourcePath|TestStickerSourcePathHashStable'`. Full `go test ./modules/sticker` could not complete in this local checkout because the shared test DB has stale migration state: `Table 'message_extra' already exists handling 20210407000001_message_legacy01.sql`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #533 (octo-server)

## Summary

This PR adds `POST /v1/sticker/user/collect`, an authenticated endpoint that lets a user save a sticker from an existing sticker message into their personal collection. Unlike the existing `add()` flow, collect accepts paths belonging to other users (`sticker/{source_uid}/{file}`) and does not require an upload handle. Idempotency is achieved through a SHA-256 hash of the source object key, backed by a generated-column unique index in the migration. A companion `sticker.custom_enabled` appconfig toggle controls client-side UI visibility for the sticker management entry without altering server-side authorization.

The implementation is clean and well-tested. One P2 performance observation below; no correctness or security blockers.

## Verification

- ✅ **Transaction integrity** — collect follows the established `lockUserRowTx` → count → check → insert → commit pattern from `add()`. The `defer tx.RollbackUnlessCommitted()` is safe across all paths: early-return idempotent hit commits, quota/conflict exits roll back, and insert errors roll back.
- ✅ **Idempotency contract** — `queryByUIDAndSourcePathHashTx` correctly filters `status=1 AND source_path_hash=?`. The generated column `source_path_hash_live` is NULL for non-live rows, so MySQL's multi-NULL unique key semantics allow re-collection after soft-delete. Test `TestSticker_CollectAfterDeleteCreatesNewRecord` confirms this.
- ✅ **Source path trust boundary** — `parseCollectStickerSourcePath` restricts input to the reserved `sticker/` keyspace via two regexes (exact and tail). Path traversal is blocked: `([^/]+)` groups reject `/` in UID and filename segments, and `isAllowedStickerFormat` whitelists only gif/png/jpg/jpeg/webp. Query params and fragments are stripped before matching.
- ✅ **Auth parity** — collect is registered under the same `auth` middleware group as all other sticker CRUD routes. No cross-user data is returned; `toStickerResp` excludes `SourcePath` and `SourcePathHash`.
- ✅ **Byte caps** — `maxStickerCollectPathLen` (2048) gates raw input; `maxStickerPathLen` (512) gates parsed `DisplayPath` and `SourceKey`. DB columns match: `source_path` VARCHAR(512), `source_path_hash` CHAR(64). Placeholder, shortcode, and keywords use the same validators as `add()`.
- ✅ **Migration safety** — idempotent stored procedure with IF NOT EXISTS guards. Down migration drops in correct order (index → generated column → base columns). Generated column is STORED, ensuring unique index usability.
- ✅ **Metrics coverage** — all 8 outcome labels are initialized in `init()` and emitted from the handler. Tests assert counter increments for success, idempotent_hit, path_invalid, and quota_exceeded.

Static analysis only at head `660badb3e3e799f21769197a1e73fd36fe7c6609`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 below.

### P2 — Idempotency query does not use the generated-column unique index (`modules/sticker/db.go:90`)

`queryByUIDAndSourcePathHashTx` filters on `uid=? AND status=1 AND source_path_hash=?`, but the unique index from the migration is on `(uid, source_path_hash_live)` — a STORED generated column. MySQL cannot use the index's second column to satisfy a filter on the base column `source_path_hash`, so after narrowing by `uid` the engine must scan all of that user's live sticker rows.

Rewriting the filter to target the generated column directly lets MySQL use both columns of the unique index:

```go
_, err := tx.Select("*").From("sticker").
    Where("uid=? AND source_path_hash_live=?", uid, sourcePathHash).
    Limit(1).
    Load(&model)
```

Because `source_path_hash_live = source_path_hash` when `status=1` (and NULL otherwise), the semantics are identical — and the explicit `status=1` check becomes redundant since the generated column already encodes it. Impact is bounded (typical user has <100 stickers), but the fix is a one-line change.

- **`handle_required` bypass**: collect intentionally skips the upload-handle check. The trust boundary is the `sticker/` keyspace (upload handler restricts writes), not the handle. This is a documented collect-specific design choice, not a gap.
- **`sticker.custom_enabled` not checked server-side**: The toggle is presentation-only (client UI visibility). Server-side CRUD authorization is unchanged. Power users who discover the endpoint can use it — this is the intended gray-release pattern.
- **Dangling references**: The brief documents that collect stores a reference, not a copy. Source-user deletion or object GC could orphan collected stickers. Acknowledged as a known trade-off for a follow-up (copy-on-collect or reference counting).
- **Case-sensitive source key**: `SourceKey` preserves original case from the URL; `Format` is normalized. This means `abc.GIF` and `abc.gif` produce different hashes. Not a bug — object keys in storage are case-sensitive, and same-message retaps produce the same URL.

## Blind-Spot Checklist

- **C1 (dual-path parity)**: add↔collect share quota, lock, and count infrastructure. Collect adds idempotency via source hash; delete correctly releases the live slot (generated column goes NULL). Both directions consistent. Clear.
- **C2 (control-flow ordering)**: collect is a standalone handler, not reused by other paths. No nested/double-call risk. Clear.
- **C3 (authorization boundary)**: collect is under the same `auth` middleware as other sticker routes. No new endpoint is exposed without auth. Clear.
- **C4 (authorization lifecycle)**: N/A — no container-member state cascade for sticker collection.
- **C5 (build/note ≠ runtime)**: N/A — no build/packaging changes.
- **C6 (governance/docs)**: N/A — no governance or security doc changes.

## Cross-Round Blocker Review

## Data-Flow Trace

- `req.Path` (caller JSON) → `parseCollectStickerSourcePath` → regex-extracted `(uid, filename, ext)` → reconstructed `sourceKey` and `displayPath` → SHA-256 `sourceHash`. All intermediate values flow correctly to the `StickerModel` insert.
- `s.settings.StickerUserMaxCount()` → `max` variable → quota check against `countByUIDTx` (counts `status=1` rows). Same source as `add()`.
- `queryByUIDAndSourcePathHashTx` returns `*StickerModel` or nil. When non-nil, the existing record is returned via `toStickerResp` — no mutation, correct idempotent behavior.
- `shortcodeExistsTx(tx, loginUID, shortcode, "")` — empty excludeID is correct for new inserts. Same pattern as `add()`.

[Octo-Q] verdict: APPROVE — one P2 (idempotency query index usage) is non-blocking. No P0/P1 findings. The implementation is solid: transaction integrity, auth parity, source-path trust boundary, and idempotency are all correctly handled. Test coverage is thorough and covers the key edge cases (idempotent hit, quota interaction, soft-delete re-collect, invalid path rejection, metrics).

yujiawei (approved):
# Code Review — PR #533 (octo-server)

Reviewed at head SHA `660badb3e3e799f21769197a1e73fd36fe7c6609` against `main` (merge-base `68e38602`). Scope: the new `POST /v1/sticker/user/collect` endpoint and its supporting model/db/metrics/migration/tests. Build (`go build ./modules/sticker/...`) and `go vet` pass locally; the module's table-driven and integration tests require MySQL/Redis and were not executed in this environment.

## Verdict: APPROVE

No blocking (P0/P1) defects. The implementation is faithful to the task brief, reuses the existing quota/lock/error-envelope machinery correctly, and the idempotency design (sha256 of the canonical source key + a `STORED` generated column feeding a partial-style unique key) is sound. A handful of P2 hardening items are listed below as suggestions — none should block merge.

## Spec compliance

The change fully implements the brief's load-bearing list and stays within scope:

- **Auth + shared UID rate limiting** — the new route is registered inside the existing `auth` group, inheriting `AuthMiddleware` + `SharedUIDRateLimiter` (`modules/sticker/api.go:102`).
- **Quota via `sticker.user_max_count` + user-row lock** — `collect()` reuses `lockUserRowTx` → `countByUIDTx` → `count >= max` in the same transaction as `add()` (`api.go:407,418,444-455`).
- **Idempotent repeat collection** — `stickerSourcePathHash(source.SourceKey)` + `queryByUIDAndSourcePathHashTx`; an existing live row commits and returns without charging quota (`api.go:406,425-442`), with a DB backstop via `uk_uid_source_path_hash_live`.
- **Reserved `sticker/{uid}/{file}.{ext}` keyspace only** — `parseCollectStickerSourcePath` re-canonicalizes to `sticker/{uid}/{file}.{ext}` and gates the extension through the raster-only allowlist (`model.go:181-220`).
- **No handle parameter** — `collectStickerReq` has no `Handle` field and `collect()` never consults `stickersig`/`handle_required`; the intentional trust-boundary difference from `add()` is documented in the handler doc comment (`api.go:337-348`).

Out-of-scope items (server-side object copy, HEAD existence check, message-access verification, reference counting/GC) are all correctly **not** implemented. Nothing unrequested was added.

## Code quality

Verified in depth (data-flow, concurrency, trust boundary, ORM behavior, parser safety):

- **Idempotency + generated column:** `source_path_hash_live` is `NULL` unless `status=1 AND source_path_hash<>''`, so direct uploads and soft-deleted collects fall out of the unique key. MySQL permits many NULLs in a unique index, so `add()` and re-collect-after-delete keep working. Soft-delete (`status=2`) recomputes the STORED column to NULL, releasing the idempotency slot — matching the `CollectAfterDeleteCreatesNewRecord` test.
- **ORM safety:** `insertTx` builds its column list via `AttrToUnderscore(m)`, which reflects only over `StickerModel`'s real fields; there is no `SourcePathHashLive` field, so the generated column is never written. `Select("*")` queries load the extra generated column into a throwaway scanner (dbr ignores result columns with no matching field), so reads are unaffected.
- **Trust boundary:** accepting foreign-UID paths without an upload handle does **not** grant a new capability. `GET /v1/file/preview/*path` is already auth-gated and returns any sticker object key to any authenticated caller, so a collector could already fetch the object whose URL was carried in the message. The stored path is re-canonicalized to a relative key (host/scheme discarded), formats are raster-only on both the upload and collect sides, and the collected reference renders only in the collector's own list. This is a strictly narrower boundary than `add()`, consistent with the brief's explicit out-of-scope note on message-access verification.
- **Parser:** Go's RE2 engine rules out ReDoS; input is length-capped (2048) before parsing and both the reconstructed `DisplayPath` and `SourceKey` are re-checked against the 512-byte column bound, closing the truncation gap; nested keys / unsupported extensions / non-sticker URLs are rejected.

### Suggestions (P2 — non-blocking)

1. **Duplicate-key on the unique index is surfaced as a 500, not an idempotent hit** (`api.go:482-485`). The pre-insert `queryByUIDAndSourcePathHashTx` + user-row lock serialize the normal same-user race, so this is not reachable for a standard authenticated caller (who always has a `user` row). But the unique key is the real correctness boundary, and `lockUserRowTx` is best-effort — it degrades to no lock when the `user` row is absent (documented at `db.go:65-72`). In that window two concurrent collects can both pass the read-check and the second `insertTx` returns MySQL error 1062, which falls through to `ErrStickerStoreFailed`. No data is corrupted (the unique key still prevents the duplicate row), but the client sees a store error instead of the existing sticker. Consider detecting the duplicate-key error on `uk_uid_source_path_hash_live` and re-querying to return the existing row — that makes idempotency hold on the DB boundary rather than on the lock.

2. **Object-key extension case is not canonicalized before hashing** (`model.go:210-224`). `Format` is lowercased, but `SourceKey` preserves the raw capture, so `sticker/u/a.GIF` and `sticker/u/a.gif` hash to different idempotency keys and could both be collected. If preview/storage routing is case-insensitive anywhere, these are the same object. Consider lowercasing the extension (or the whole key) before computing the hash, or rejecting non-lowercase extensions.

3. **Path segments allow non-canonical characters** (`model.go:177-178`, `[^/]+`). Segments like `..`, spaces, or percent-encoded separators (`%2F`, `%2e`) parse successfully. The downstream `GET /v1/file/preview` handler's `sanitizePath` decodes and rejects `..` at read time, so this is not a live traversal, but it slightly widens the "reserved keyspace only" guarantee and can fragment idempotency. Tightening both segments to the same safe character set the uploader produces would make the invariant self-contained.

## Coverage notes

- Integration tests were not executed here (MySQL/Redis/WuKongIM required); the added table-driven tests cover idempotent hit, quota non-consumption, re-collect-after-delete, invalid-path rejection, and the quota/path_invalid metrics — a good spread.
- Generated-column + unique-key runtime behavior is reasoned from the migration SQL and MySQL semantics, not exercised against a live DB in this review.
- Whether any other module reads `sticker.path`/`source_path` for a user other than the owner was spot-checked within `modules/sticker` (no such consumer) but not swept exhaustively across the whole repo; the "renders only to the collector" assumption rests on that.
- A second independent advisory pass (long-context, cross-family) timed out and did not return findings for this review; the conclusions above rest on the primary review plus one adversarial cross-check leg.

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD 660badb3

**Formal Verdict**: `APPROVE`



## Reviewer Summary

|---|---|---|
| qa-engineer | pass-with-risk | 4 核心 Acceptance 均有强断言 + CI 全绿；并发/超长路径/shortcode 冲突三类边界与部分 AC4 结果指标无测试断言（非阻断）。 |
| security-engineer | cleared-with-risk | 无凭据泄露、无鉴权绕过、无可利用注入；最高 Medium 均为 spec 明示权衡或当前不可利用的防御纵深项。 |
| code-reviewer | approved-with-nits | 事务骨架忠实复刻已验证的 add 流程、幂等查询正确前置；0 Must / 0 Should，仅 5 条测试覆盖与整洁度 Nit。 |

## Critical Findings (P0-P1)

无 Critical / High priority findings — all reviewers converged on lower-severity items and cleared the PR as non-blocking.

## Should-Fix (P2)

无 Should-Fix findings — 无 reviewer 提出 Must/Should-Fix；security 两项 Medium 均被明确定级为非阻断（一为 spec out-of-scope 权衡，一为当前不可利用、由 serve 层 `sanitizePath` 兜底的防御纵深项）。详见 §Verdict Rationale。

## Suggestions (P3)

- [P3-1] `modules/sticker/api_test.go` — `shortcode_conflict` / `query_failed` / `store_failed` 三个结果 label 无断言（AC4 指标可观测性部分覆盖）— flagged by qa/Risk-1 + code/Nit-1 — Fix: 补一个「两不同来源路径复用同一 shortcode」的 `shortcode_conflict` 计数断言（无需 mock DB）；`query_failed`/`store_failed` 属注入型可作 follow-up。
- [P3-2] `modules/sticker/api.go:110`、`:121` — 超长路径拒绝分支（输入 >2048、派生 `SourceKey`/`DisplayPath` >512）无用例 — flagged by qa/Risk-2 + code/Nit-2 — Fix: 各补一条拒绝断言。
- [P3-3] `modules/sticker/api.go:121` — 派生 `SourceKey` 长度检查冗余（`DisplayPath` = 固定前缀 + `SourceKey`，恒长于 `SourceKey`，该分支永不单独触发）— flagged by code/Nit-3 — Fix: 简化为仅检 `DisplayPath`。
- [P3-4] `modules/sticker/api.go:154-249` — collect 与 add 的「开事务→行锁→计数配额→shortcode 冲突→插入→提交」事务尾部高度重复 — flagged by code/Nit-4 — Fix: 出现第三个写入端点时抽出公共尾部。
- [P3-5] `modules/sticker/db.go` `lockUserRowTx` — 用户表无该 uid 行时锁失效，极端并发同 path 下第二次插入撞唯一键返回 `store_failed` 而非优雅 `idempotent_hit`（唯一键仍阻止脏重复，比 add 更安全）— flagged by qa/Risk-3 + code/Nit-5 — 观察项。
- [P3-6] `modules/sticker/model.go` `parseCollectStickerSourcePath` — 段匹配正则 `([^/]+)` 字符集偏宽（防御纵深；**当前不可利用** —— 唯一读取 sink `getFile` 由 `sanitizePath` 兜底，且无后台任务直接消费 `source_path`）— flagged by security/Risk-2 — Fix: 收紧段字符集为 `[A-Za-z0-9_-]` 并拒纯点段，在 brief 已规划的 copy-on-collect / GC 直接以 `source_path` 拼真实路径前保住「仅限保留贴纸 keyspace」不变量。

## Follow-ups (spec-acknowledged trade-offs — 非本轮阻断)

- **跨用户来源引用（path-only 契约）** — security/Risk-1 + qa/Risk-4 — 理论 Medium，但**有界且当前不可利用**：预览端点 `getFile`（`modules/file/api.go:68`）本就仅挂 `AuthMiddleware`、对任意已认证用户开放，且贴纸对象名为随机 UUID 不可枚举 → collect 未产生比现有预览端点更强的访问权。brief §Out-of-scope 已明列。Follow-up: 引入 message-identity 入参、由服务端派生贴纸 path。
- **引用式存储的悬挂引用** — security/Risk-3 (Low) + qa/Risk-4 — collect 存引用不拷贝、不做对象存在性校验、不做引用计数；源对象 GC / 源用户清理可产生悬挂引用。brief §Known Trade-Offs 已显式记录。Follow-up: copy-on-collect 或引用计数（brief 计划内）。

## Questions for Author

- [Q-1] （非阻断）本 PR 无关联 issue，Acceptance 来自 PR description + 绑定 spec brief `.octospec/tasks/sticker-collect/brief.md`；请作者确认该 AC 集合完整（from qa）。

## Verdict Rationale

三方 reviewer 一致收敛于非阻断结论：qa `pass-with-risk`、security `cleared-with-risk`、code `approved-with-nits`（0 Must / 0 Should / 5 Nit），无任一 P0/P1，P2 计为 0。核心 4 条 Acceptance（幂等不重复扣配额、软删释放幂等槽、非法路径本地化拒绝、结果指标可观测的主路径 4/8 label）在实现与测试中均有强断言，CI 全绿（Test/Build/Vet/Lint/i18n/secret-scan/osv 全 SUCCESS）。

关于 severity 表达的透明说明：security 报告最高定级 Medium（Risk-1 跨用户来源引用、Risk-2 正则段字符集偏宽），但两项均被 security 自身明确定为**非阻断** —— Risk-1 是 brief 显式记录的 out-of-scope 设计权衡且因预览端点既有鉴权语义而不产生新访问权，Risk-2 是当前不可利用、由 serve 层 `sanitizePath` 兜底的防御纵深加固建议。三方对这些风险的认知**一致无分歧**（code 报告 §B-4 明确附议）。据 skill Formal Verdict 规则表首次匹配：无 P0 → 无 ≥1 P1 或 ≥3 P2 → 无阻断性 material Question → 全部为 P3 → `APPROVE`。

本裁决表达「三方评审已就当前 HEAD 收敛、无阻断项」，不构成合并意见；合并是人类专属操作。

## Round Context

_事实记录（非评审意见）：PR #533 已由人类于 2026-07-03T08:07:45Z 合并入 `main`（HEAD 660badb3，与 reviewing_sha 一致），发生在 security / code 报告落地之前。本聚合基于三份 Round-1 报告对该 HEAD 的评审如实出具；上列 P3 与 Follow-up 项建议以独立 follow-up issue 跟进。_

---

lml2468 (approved):
**APPROVE — Round 1** (HEAD `660badb3`)

三方评审一致收敛于非阻断：qa `pass-with-risk` · security `cleared-with-risk` · code `approved-with-nits`（0 Must / 0 Should / 5 Nit）。无 P0/P1/P2，核心 4 条 Acceptance 强断言 + CI 全绿；security 最高 Medium 两项均为 spec 明示权衡或当前不可利用的防御纵深项。

全部 6 项 P3（测试覆盖缺口 + 整洁度 Nit + 段字符集加固）与 2 项 spec-acknowledged follow-up 详见本 PR 的聚合评论：https://github.com/Mininglamp-OSS/octo-server/pull/533#issuecomment-4874304444

PR #537, opened 2026-07-04
fix(space): member search name fallback + collation-safe verification JOIN #434

lml2468:
## Summary

Space-side member search (`GET /v1/space/:space_id/members/search`) never picked up the display-name fallback that #420 added to the member list, so this closes the **P1** gap tracked in #434:

- blank `user.name` members rendered an **empty name** in search results;
- members with a blank `user.name` were **not searchable by `real_name`**.

## Changes

- `searchMembers` / `countSearchMembers` now `LEFT JOIN user_verification` and select `real_name`.
- The search response reuses the same fallback chain as the member list — `user.name` → `user_verification.real_name` → stable `"User <uid>"` placeholder — extracted into a shared pure `resolveMemberDisplayName` helper (used by both `MemberDetailModel.DisplayName()` and the new `memberSearchModel.DisplayName()`).
- `uv.real_name` added to the searchable columns so blank-name members are findable by real name.
- **Collation-safe from the start**: the new JOIN uses `ON uv.user_id = sm.uid COLLATE utf8mb4_general_ci`, so it does **not** reintroduce the `Illegal mix of collations (1267)` / 500 risk that #482 tracks for `queryMembers` on non-canonical (`0900_ai_ci`) deployments.

## Scope

Only the P1 search path. The remaining #434 follow-ups (P2 avatar / other `IFNULL(u.name,'')` endpoints, P3 shared `pkg/displayname`) are intentionally left for separate PRs (one-PR-one-thing).

## Verification

Ran against a real MySQL 8.0.46 instance:
- Full `modules/space` suite **green** on the canonical `general_ci` DB (incl. 3 new `Test_Issue434_*` tests).
- RED→GREEN confirmed: both new DB-backed tests fail without the fix (empty `Name`; `count=0`) and pass with it.
- Collation safety proven on a `0900_ai_ci` DB with the exact prod schema mismatch (`space_member`=`0900_ai_ci`, `user_verification`=`general_ci`): the bare join raises Error 1267 while the `COLLATE`-safe join returns rows correctly.

Refs #434, #420, #482, #344.

> ⚠️ Draft — awaiting review squad (qa + security + code). Do not merge.

### DIFF
--- a/modules/space/api_member_search.go
+++ b/modules/space/api_member_search.go
@@ -69,8 +69,11 @@ func (s *Space) searchMembers(c *wkhttp.Context) {
 	resps := make([]memberSearchResp, 0, len(members))
 	for _, m := range members {
 		resps = append(resps, memberSearchResp{
-			UID:       m.UID,
-			Name:      m.Name,
+			UID: m.UID,
+			// DisplayName 复用成员列表的兜底链（issue #434）：user.name 为空时回退
+			// user_verification.real_name，二者皆空给稳定占位符，避免空名成员在
+			// 搜索结果里渲染成空白行。
+			Name:      m.DisplayName(),
 			Username:  m.Username,
 			Email:     m.Email,
 			Phone:     maskPhone(m.Phone),
--- a/modules/space/db.go
+++ b/modules/space/db.go
@@ -167,6 +167,15 @@ func (d *DB) queryMembers(spaceId string, loginUID string, page uint64, limit ui
 	return models, err
 }
 
+// memberVerificationJoinOn 是 space_member → user_verification 的 collation-safe
+// LEFT JOIN 条件。user_verification 被 modules/base 的 compat-repair 迁移强制
+// utf8mb4_general_ci，而 space_member 无显式 COLLATE、继承 DB 默认；在以 MySQL 8.0
+// 服务器默认（utf8mb4_0900_ai_ci）建库的部署上，裸的 uv.user_id = sm.uid 比较会抛
+// Illegal mix of collations (1267)（见 #482：queryMembers 已上 main 的同类风险）。
+// 显式 COLLATE 把 sm.uid 归一到 general_ci，使搜索路径不复刻同一 500 风险，且在
+// 规范 general_ci 库上是 no-op。
+const memberVerificationJoinOn = "uv.user_id = sm.uid COLLATE utf8mb4_general_ci"
+
 func (d *DB) searchMembers(spaceId, keyword string, pageIndex, pageSize int) ([]*memberSearchModel, error) {
 	builder := d.session.Select(
 		"sm.*",
@@ -174,9 +183,11 @@ func (d *DB) searchMembers(spaceId, keyword string, pageIndex, pageSize int) ([]
 		"IFNULL(u.username,'') as username",
 		"IFNULL(u.email,'') as email",
 		"IFNULL(u.phone,'') as phone",
+		"IFNULL(uv.real_name,'') as real_name",
 		"CASE WHEN r.robot_id IS NOT NULL AND r.status=1 THEN 1 ELSE 0 END as robot",
 	).From(dbr.I("space_member").As("sm")).
 		LeftJoin(dbr.I("user").As("u"), "u.uid=sm.uid").
+		LeftJoin(dbr.I("user_verification").As("uv"), memberVerificationJoinOn).
 		LeftJoin(dbr.I("robot").As("r"), "r.robot_id=sm.uid").
 		Where("sm.space_id=? AND sm.status=1", spaceId)
 	if keyword != "" {
@@ -198,6 +209,7 @@ func (d *DB) countSearchMembers(spaceId, keyword string) (int64, error) {
 	builder := d.session.Select("COUNT(*)").
 		From(dbr.I("space_member").As("sm")).
 		LeftJoin(dbr.I("user").As("u"), "u.uid=sm.uid").
+		LeftJoin(dbr.I("user_verification").As("uv"), memberVerificationJoinOn).
 		Where("sm.space_id=? AND sm.status=1", spaceId)
 	if keyword != "" {
 		clause, args := memberSearchActiveWhere(keyword)
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -54,20 +54,29 @@ func memberSearchWhere(keyword string) (string, []interface{}) {
 //     （响应仅显示 138****5678），admin 无法通过子串查询逐位探测/重建完整号码。
 //
 // 前端注意：phone 检索只匹配后 4 位，传完整号码不会命中——按手机号查找请用后 4 位。
-// uv.real_name 让空 user.name 的成员可按实名被搜到（issue #434）；对应的 uv LEFT JOIN
-// 由 searchMembers / countSearchMembers 以 collation-safe 方式挂入（memberVerificationJoinOn）。
-var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid", "uv.real_name"}
+// 注意：uv.real_name 不在此列——它只在「作为展示值」时可检索，由 memberSearchActiveWhere
+// 用 name 为空的门禁单独拼接，见该函数注释。
+var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
 
 // memberSearchActiveWhere 为空间侧 members/search 组装跨列 OR LIKE 条件。
 // list / count 共用同一条件，避免搜索范围漂移导致分页错位。
+//
+// real_name（user_verification 实名，PII）单独处理，且门禁在 user.name 为空：实名
+// **只在它是实际展示值时**（即 user.name 为空、DisplayName 回退到 real_name）才可检索，
+// 使「可检索粒度 == 可见粒度」——与本文件对 phone 只放后 4 位（maskPhone）同一取向。
+// 若无门禁，有 name 的成员响应只显示 name、绝不返回 real_name，却能被实名子串命中，
+// admin 便可把搜索当作实名 oracle 逐位探测出从不展示的法定姓名（#537 review 三人共识）。
 func memberSearchActiveWhere(keyword string) (string, []interface{}) {
 	like := buildLikePattern(keyword)
-	clauses := make([]string, len(memberSearchActiveColumns))
-	args := make([]interface{}, len(memberSearchActiveColumns))
-	for i, col := range memberSearchActiveColumns {
-		clauses[i] = col + " LIKE ?" + likeEscapeClause
-		args[i] = like
+	clauses := make([]string, 0, len(memberSearchActiveColumns)+1)
+	args := make([]interface{}, 0, len(memberSearchActiveColumns)+1)
+	for _, col := range memberSearchActiveColumns {
+		clauses = append(clauses, col+" LIKE ?"+likeEscapeClause)
+		args = append(args, like)
 	}
+	// 门禁与 resolveMemberDisplayName 的回退条件（name 非空即用 name）严格对齐。
+	clauses = append(clauses, "(IFNULL(u.name,'')='' AND uv.real_name LIKE ?"+likeEscapeClause+")")
+	args = append(args, like)
 	return strings.Join(clauses, " OR "), args
 }
 
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -54,7 +54,9 @@ func memberSearchWhere(keyword string) (string, []interface{}) {
 //     （响应仅显示 138****5678），admin 无法通过子串查询逐位探测/重建完整号码。
 //
 // 前端注意：phone 检索只匹配后 4 位，传完整号码不会命中——按手机号查找请用后 4 位。
-var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
+// uv.real_name 让空 user.name 的成员可按实名被搜到（issue #434）；对应的 uv LEFT JOIN
+// 由 searchMembers / countSearchMembers 以 collation-safe 方式挂入（memberVerificationJoinOn）。
+var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid", "uv.real_name"}
 
 // memberSearchActiveWhere 为空间侧 members/search 组装跨列 OR LIKE 条件。
 // list / count 共用同一条件，避免搜索范围漂移导致分页错位。
--- a/modules/space/db_member_search_name_fallback_test.go
+++ b/modules/space/db_member_search_name_fallback_test.go
@@ -1,0 +1,88 @@
+package space
+
+import (
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/assert"
+)
+
+// Test_Issue434_SearchMemberDisplayNamePure 纯函数验证成员搜索路径复用了与成员
+// 列表相同的兜底链（issue #434）：name 空 → real_name → 稳定占位符，任何分支非空，
+// 绝不读 short_no/username。不依赖 DB。
+func Test_Issue434_SearchMemberDisplayNamePure(t *testing.T) {
+	mk := func(uid, name, realName string) *memberSearchModel {
+		m := &memberSearchModel{Name: name, RealName: realName}
+		m.UID = uid
+		return m
+	}
+
+	assert.Equal(t, "Alice", mk("u1", "Alice", "Real").DisplayName())
+	assert.Equal(t, "Real", mk("u2", "", "Real").DisplayName())
+	assert.Equal(t, memberDisplayNamePlaceholderPrefix+"u3", mk("u3", "", "").DisplayName())
+	assert.NotEqual(t, "", mk("u4", "", "").DisplayName(), "placeholder must never be empty")
+}
+
+// Test_Issue434_SearchMemberNameFallback 复现 #434 的 P1 缺口之一：成员搜索结果
+// 对空 user.name 的成员未做兜底，会渲染成空白行。修复后 Name 应回退 real_name，
+// 二者皆空时给稳定占位符。
+func Test_Issue434_SearchMemberNameFallback(t *testing.T) {
+	srv, _, err := setup(t)
+	assert.NoError(t, err)
+
+	const spaceId = "sp-434-search-fallback"
+	seedMemberSearchSpace(t, spaceId, testutil.UID) // 调用者即 owner，满足搜索前置校验
+
+	// 1) name 空 + real_name 有 → 搜索结果回退 real_name
+	const uidReal = "u434-real"
+	seedFallbackUser(t, uidReal, "")
+	seedFallbackVerification(t, uidReal, "Zhang Wei")
+	seedMemberSearchMember(t, spaceId, uidReal, 0, 1)
+
+	// 2) name 空 + real_name 空 → 稳定占位符，绝不空白
+	const uidPlaceholder = "u434-ph"
+	seedFallbackUser(t, uidPlaceholder, "")
+	seedMemberSearchMember(t, spaceId, uidPlaceholder, 0, 1)
+
+	w := getMembersSearch(t, srv, testCtx, spaceId, url.Values{"page_size": {"50"}})
+	assert.Equal(t, http.StatusOK, w.Code, w.Body.String())
+	resp := decodeMembersSearchResp(t, w)
+
+	real, ok := findMemberSearchItem(resp.List, uidReal)
+	if assert.True(t, ok, "real_name member missing from search results") {
+		assert.Equal(t, "Zhang Wei", real.Name, "empty name must fall back to real_name in search results")
+	}
+	ph, ok := findMemberSearchItem(resp.List, uidPlaceholder)
+	if assert.True(t, ok, "placeholder member missing from search results") {
+		assert.NotEqual(t, "", ph.Name, "blank name+real_name must never render as an empty string")
+		assert.Equal(t, memberDisplayNamePlaceholderPrefix+uidPlaceholder, ph.Name)
+	}
+}
+
+// Test_Issue434_SearchByRealName 复现 #434 的 P1 缺口之二：空 user.name 的成员
+// 无法通过实名被搜到。关键词只匹配 user_verification.real_name（不匹配
+// name/username/email/phone/uid），修复前 count=0，修复后应命中该成员并回退展示实名。
+func Test_Issue434_SearchByRealName(t *testing.T) {
+	srv, _, err := setup(t)
+	assert.NoError(t, err)
+
+	const spaceId = "sp-434-search-realname"
+	seedMemberSearchSpace(t, spaceId, testutil.UID)
+
+	const uid = "u434-byreal"
+	seedFallbackUser(t, uid, "") // 无 user.name / username / email / phone
+	seedFallbackVerification(t, uid, "Ouyang Feng")
+	seedMemberSearchMember(t, spaceId, uid, 0, 1)
+
+	w := getMembersSearch(t, srv, testCtx, spaceId, url.Values{"keyword": {"Ouyang"}})
+	assert.Equal(t, http.StatusOK, w.Code, w.Body.String())
+	resp := decodeMembersSearchResp(t, w)
+
+	assert.EqualValues(t, 1, resp.Count, "member with blank name must be searchable by real_name")
+	if assert.Len(t, resp.List, 1) {
+		assert.Equal(t, uid, resp.List[0].UID)
+		assert.Equal(t, "Ouyang Feng", resp.List[0].Name, "search result must display the real_name fallback")
+	}
+}
--- a/modules/space/db_member_search_name_fallback_test.go
+++ b/modules/space/db_member_search_name_fallback_test.go
@@ -86,3 +86,40 @@ func Test_Issue434_SearchByRealName(t *testing.T) {
 		assert.Equal(t, "Ouyang Feng", resp.List[0].Name, "search result must display the real_name fallback")
 	}
 }
+
+// Test_Issue434_NamedMemberNotSearchableByRealName pins the privacy invariant flagged
+// by the #537 review (three-reviewer consensus): real_name is searchable ONLY when it is
+// the actual display value (user.name blank). A member with a non-empty user.name displays
+// that name and never returns real_name, so their verification real_name must NOT be a
+// search oracle — otherwise an in-space admin could substring-probe a hidden legal name and
+// confirm it via the hit/count. This is the "可检索粒度 == 可见粒度" boundary the file also
+// holds for phone (maskPhone). Without the name-empty gate this test fails (member matched).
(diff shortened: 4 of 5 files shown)


OctoBoooot (commented):
## Review: fix(space): member search name fallback + collation-safe verification JOIN (#537)

**Verdict:** Comment — draft PR; the collation half is right, but this **reintroduces the exact `real_name` search-vs-display privacy asymmetry that #523 was REQUEST_CHANGES'd for**. Draft → COMMENT (#513), but this must be fixed before it un-drafts.

Risk tier: normal; reviewed on the privacy axis (real_name is verification PII) and the collation axis (#489 lesson).

### Blocking-on-un-draft (same finding as #523, three-reviewer-converged there)
- `db_manager.go:59` adds `uv.real_name` to `memberSearchActiveColumns`, and `memberSearchActiveWhere` (:63-72) OR's a bare `uv.real_name LIKE ?` for **every** column with **no name-empty guard**. But `real_name` is only *displayed* when `user.name` is empty (the fallback chain). So `real_name` is **searchable for all members but visible only for blank-name members** → an in-space admin can substring-probe a *named* member's legal name (which lives in `user_verification`, deliberately never returned) and confirm it via the count/hit without it ever appearing in a response. That's the searchable-but-not-visible confirm-oracle on PII that #523 landed as a blocker (me + Jerry-Xin + OctoBoooot converged). `real_name` is the one searchable column that isn't a returned field — unlike username/email which are visible.
- **Fix (same as #523):** match `real_name` only when it's the actual display value —
  `CASE WHEN IFNULL(u.name,'')='' THEN uv.real_name ELSE '' END LIKE ?`
- **Test gap:** the suite covers the *forward* direction (blank-name member found by real_name, `Test_Issue434_SearchByRealName`) but has **no inverse test** — nothing asserts a *named* member is NOT found by their real_name. Add it; it's the assertion that pins the privacy invariant (searchable == visible).

### Done right — collation-safe JOIN (the #489 lesson, correctly applied)
`memberVerificationJoinOn = "uv.user_id = sm.uid COLLATE utf8mb4_general_ci"` is applied at **both** `searchMembers` (db.go:190) and `countSearchMembers` (:212). That correctly pins the cross-collation compare (user_verification forced to general_ci by the compat-repair migration vs space_member inheriting the DB default) on both the list and count sides, avoiding the Error 1267 500 on MySQL-8 `0900_ai_ci` deployments. Good — and it fixes it the durable way (shared const, both sides).

### Display fallback — good
`name → real_name → stable placeholder`, never empty, matching the member-list chain. Tests cover blank-name→real_name and blank-name+blank-real_name→placeholder.

### Net
The collation fix and display fallback are correct. The one thing to fix before merge is the `real_name` **search predicate**: guard it to name-empty (so searchable == visible) and add the inverse test. This is the same privacy blocker #523 carried — it should not re-land unfixed. Confirm `memberSearchResp` also doesn't expose `real_name` as its own field (the display path folds it into `Name`, which is correct; just don't add a raw `real_name` to the response).

Jerry-Xin (changes requested):
The PR is relevant to octo-server and the implementation is mostly aligned with the intended space member search fix, but it expands real-name searchability beyond the documented/displayed fallback contract.

**🔴 Blocking**
- 🔴 Critical: `uv.real_name` is added as an unconditional searchable column in [modules/space/db_manager.go](/Users/mlamp/.openclaw/tmp/pr-review-67i9tft7/repo/modules/space/db_manager.go:59). That means a member with a non-empty `user.name` can now be discovered by searching their `user_verification.real_name`, even though the response will still display `user.name` and the fallback contract only exposes real name when `user.name` is blank. This creates a privacy/inference leak: an admin can probe legal names and map them to visible member profiles. The real-name predicate should be gated to the fallback case, for example `IFNULL(u.name,'') = '' AND uv.real_name LIKE ?`, and a regression test should assert that a member with non-empty `user.name` is not matched solely by `real_name`.

**💬 Non-blocking**
- 🟡 Warning: I could not complete `go test ./modules/space/...` locally because the test database migration state failed before package tests ran: `unknown migration in database` for `20191106000001_event_legacy01.sql`. This appears environmental, not caused by this PR.

**✅ Highlights**
- The PR correctly reuses a shared pure display-name resolver in [modules/space/model.go](/Users/mlamp/.openclaw/tmp/pr-review-67i9tft7/repo/modules/space/model.go:293), reducing drift between list and search responses.
- The new `user_verification` join is cardinality-safe because `user_verification.user_id` is a primary key, and the explicit collation on the join condition avoids the known mixed-collation failure path.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #537 (octo-server)

## Summary

This PR fixes issue #434 for the space-side member search path: members whose `user.name` is empty now fall back to `user_verification.real_name` in both the display name and the search index, with a stable `"User <uid>"` placeholder as the terminal fallback. The implementation extracts a shared pure function `resolveMemberDisplayName` so the member-list path (issue #344) and the search path share identical logic. A collation-safe LEFT JOIN to `user_verification` is introduced to avoid the MySQL 1267 "Illegal mix of collations" error already tracked for the list path in #482.

The change is well-scoped, well-tested, and the data flow is correct end-to-end. No P0 or P1 issues found.

## Verification

- ✅ **Data flow: `real_name` SELECT → struct → DisplayName** — `searchMembers` at `modules/space/db.go:185` selects `IFNULL(uv.real_name,'') as real_name`; the dbr mapper populates `memberSearchModel.RealName` (`modules/space/model.go:51`); `DisplayName()` at `:283` passes it to `resolveMemberDisplayName`, which falls back correctly through name → realName → placeholder.
- ✅ **JOIN cardinality: no fan-out** — `user_verification` has `PRIMARY KEY (user_id)` (confirmed at `modules/user/sql/20260505000003_user_legacy01.sql:17`), so the LEFT JOIN produces at most one row per `space_member`. COUNT and list stay consistent.
- ✅ **Collation coercion** — `memberVerificationJoinOn` at `modules/space/db.go:177` coerces `sm.uid` to `utf8mb4_general_ci` to match `uv.user_id` (whose table declares `COLLATE=utf8mb4_general_ci`). On `utf8mb4_0900_ai_ci` databases this prevents MySQL 1267; on `utf8mb4_general_ci` databases it is a no-op. Correct MySQL expression-coercion syntax.
- ✅ **LIKE safety on new column** — `uv.real_name` is added to `memberSearchActiveColumns` at `modules/space/db_manager.go:59`. The LIKE pattern goes through `buildLikePattern` → `escapeLike` which escapes `\`, `%`, `_`. The `ESCAPE '\\'` clause is appended. No injection vector.
- ✅ **COUNT/list parity** — Both `countSearchMembers` and `searchMembers` now include the `user_verification` LEFT JOIN and both use `memberSearchActiveWhere`. Pagination math stays consistent.
- ✅ **Auth gate unchanged** — `checkSpaceActive` + `member.Role < spaceRoleAdmin` at `modules/space/api_member_search.go:27-41` are pre-existing and not modified. Only admin-role members of active spaces can reach the search.
- ✅ **Test coverage** — Three new test functions in `db_member_search_name_fallback_test.go` cover: pure-function fallback chain, end-to-end search response with empty name + real_name, and search-by-real_name keyword. Test helpers reuse existing `seedFallbackUser`/`seedFallbackVerification` from `db_member_name_fallback_test.go`.

Static analysis only at head `311bde8fec31adf537c372ef58338aab5e5d59f5`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 observation below.

### P2 — Admin-side member search does not gain the real_name fallback (`modules/space/db_manager.go:36`)

The space-side `memberSearchActiveColumns` (line 59) gains `uv.real_name`, but the admin-side `memberSearchColumns` (line 36) is unchanged: `{"u.name", "u.username", "u.email", "u.phone", "sm.uid"}`. This means admin search still cannot find members whose `user.name` is empty by their real name — the same #434 gap the space side just fixed, but on the management console.

Diff-scope check: the PR does not touch `memberSearchColumns` or `memberSearchWhere` (admin path), so this is **pre-existing**, not amplified. However, it creates a search-behavior asymmetry between the two surfaces that operators may find surprising: a space admin can search by real_name but a system admin cannot. Worth a follow-up issue to extend the admin path with the same LEFT JOIN + column addition, or an explicit comment explaining the deliberate scope difference.

- **No PII regression in response shape** — `memberSearchResp` struct at `modules/space/model.go:187` is unchanged. The `Name` field now carries the fallback value but the JSON contract (`json:"name"`) is identical. No new fields exposed.
- **NULL LIKE semantics correct** — When a member has no `user_verification` row, `uv.real_name` is NULL in the WHERE clause. `NULL LIKE '%x%'` evaluates to NULL (falsy), so that member is correctly excluded from keyword matches on the real_name column. The IFNULL in the SELECT only affects the display value, not search matching.
- **`queryMembers` bare JOIN tracked separately** — The member-list path at `modules/space/db.go:158` still uses `uv.user_id=sm.uid` without COLLATE. This is pre-existing and tracked by issue #482. This PR's search path uses the safe constant — no regression.
- **Refactoring quality** — `resolveMemberDisplayName` pure-function extraction is clean. Both `MemberDetailModel.DisplayName()` and `memberSearchModel.DisplayName()` delegate to it. The fallback chain cannot diverge between list and search paths.

## Verdict: APPROVED

Solid, well-documented fix with thorough test coverage. The collation-safe JOIN pattern is a good defensive measure for MySQL 8.0 deployments. The only follow-up item is the admin-side search parity (P2), which is pre-existing and does not block this PR.

yujiawei (changes requested):
# Code Review — PR #537 (octo-server)

Reviewed at head SHA `311bde8fec31adf537c372ef58338aab5e5d59f5` against merge-base `ae1b51e1` on `main`. This PR is security-sensitive (touches a member-search predicate and joins the identity-verification table), so it received an extra privacy-boundary pass plus two independent cross-model reviews. Both independent passes converged on the same blocking issue below.

## 1. Specification Compliance

The linked issue (#434 P1) asks for two things on `GET /v1/space/:space_id/members/search`: (a) blank-`user.name` members should no longer render an empty name, and (b) such members should be findable by their verified real name. The diff delivers both, reuses the existing `#344/#420` fallback chain via a shared pure helper (`resolveMemberDisplayName`, `modules/space/model.go:293`), and keeps `count`/`list` on the same predicate so pagination stays consistent.

**No missing scope, no over-build** — the PR explicitly defers the P2/P3 follow-ups (avatar fallback, `pkg/displayname`) and does not touch them. One-PR-one-thing is respected.

**However, requirement (b) as implemented introduces a privacy regression** that the issue did not ask for and that contradicts an invariant already codified in this same file — see P1 below. Spec intent is met; the *implementation* of the search predicate crosses a boundary the codebase deliberately holds.

## 2. Findings

### P1 — Real-name search is not gated to the fallback case (privacy oracle)

`modules/space/db_manager.go:59` adds `uv.real_name` to `memberSearchActiveColumns` **unconditionally**, so it is OR-matched for *every* member — including members who have a non-empty `user.name`.

For such a member, the response still displays `user.name` (the fallback chain in `resolveMemberDisplayName` returns early on non-empty `name` and never exposes `real_name`), yet the row is now *matchable* by substrings of the legal/ID-verified `real_name`. A space admin can therefore use `/members/search` as a **real-name oracle**: probe substrings, observe which member row comes back, and reconstruct a legal name that the API never displays for that member.

This directly violates the "可检索粒度 == 可见粒度" (searchable granularity == visible granularity) invariant that this very file establishes for phone numbers (`modules/space/api_member_search.go:92-95`, `maskPhone`) — phone is deliberately searchable only on the last 4 digits precisely to prevent this class of probing. `real_name` is *more* sensitive than a phone suffix, not less.

**Fix:** match `uv.real_name` only when `u.name` is empty, so searchability tracks the same boundary as display. Concretely, gate that one OR-term, e.g. `(u.name = '' AND uv.real_name LIKE ? ESCAPE '\\')`, instead of a bare `uv.real_name LIKE ?`. That keeps requirement (b) — blank-name members findable by real name — while closing the oracle for named members.

This is the intended behavior implied by #434 ("members *with a blank `user.name`* [should be] searchable by `real_name`"), so the gate is also the more faithful reading of the spec.

### P2 — Collation-safety claim is not exercised by any test

The core justification for `memberVerificationJoinOn` (`modules/space/db.go:177`) is that it prevents `Illegal mix of collations (1267)` on deployments where the DB default is `utf8mb4_0900_ai_ci` while `user_verification` is pinned to `general_ci`. That is a sound and well-reasoned fix. But the test harness (`modules/space/api_test.go`) creates **every** table as `utf8mb4_general_ci`, so the `COLLATE` clause is a pure no-op in CI — a regression that dropped it, or mangled the collation name, would stay green. The PR body claims manual verification against a real `0900_ai_ci` DB with the prod schema mismatch, which I trust, but nothing guards it going forward. Consider adding a DB-backed test that builds `space_member` (or the DB default) as `0900_ai_ci` and asserts the join returns rows rather than erroring. Non-blocking, but this is exactly the invariant most likely to silently rot.

### P2 (advisory) — `queryMembers` on `main` still carries the same 1267 risk

The comment at `modules/space/db.go:171-176` notes `queryMembers` (already on `main` via #344/#420) uses a **bare** `LEFT JOIN user_verification uv ON uv.user_id=sm.uid` with no `COLLATE`, i.e. the standard member-list path remains exposed to the same 500 on `0900_ai_ci` deployments (tracked as #482). Out of scope for this PR — flagging only so the follow-up is not lost; the member *list* is a more trafficked path than search.

## 3. Points verified clean (not findings)

- **SQL injection:** all predicates use parameterized `?` with an explicit `ESCAPE '\\'` clause (`likeEscapeClause`) and LIKE-metacharacter escaping (`escapeLike`). No interpolation. ✅
- **LEFT JOIN row fan-out / COUNT inflation:** `user_verification` has `PRIMARY KEY (user_id)` (`modules/user/sql/20260505000003_user_legacy01.sql:18`), so the join is strictly 1:0..1. No duplicate rows, no inflated `countSearchMembers`, pagination integrity preserved. ✅
- **Dialect portability:** `COLLATE utf8mb4_general_ci` is MySQL-specific, but the entire `space` DB layer is MySQL-only (raw MySQL DSN, `RIGHT()`, existing MySQL-only SQL); the sqlite driver in the tree is used by other modules, never this path. No real portability regression. ✅
- **Display fallback:** never returns empty and never leaks `short_no`/`username`; placeholder branch reuses already-exposed `uid`. ✅

## 4. Recommendation

The display-name fallback, the shared helper refactor, the collation-safe join, and the 1:0..1 join reasoning are all correct and well-documented. The one blocker is the ungated `real_name` search term, which turns a display fallback into a searchable PII oracle for members whose real name is otherwise hidden — a boundary this file explicitly maintains elsewhere. Gate that single OR-term to the `u.name = ''` case and this is good to merge.

Also worth a manual human check given the security classification: confirm the intended visibility policy for verified real names in a space (is a legal name ever meant to be admin-searchable when a display name exists?), and confirm the CI collation coverage gap above.

OctoBoooot (commented):
## Delta re-review: #537 @ e96691d5 (prior @ 311bde8f) — privacy blocker FIXED

**Verdict:** Approve-with-comments (draft). The one blocking-on-un-draft finding from my prior review — the `real_name` search-vs-display privacy asymmetry (#523's blocker) — is fixed exactly as recommended, with the inverse test I asked for. Collation JOIN remains correct and I verified it's also index-safe (the #567 lesson). Still a draft → Build/Test/Vet SKIPPED; can't be a final APPROVE until it un-drafts and code CI runs.

### Privacy blocker — RESOLVED (byte-verified)
- `real_name` is **removed** from `memberSearchActiveColumns` (`db_manager.go:59`) — no longer a blanket search column.
- The real_name clause is now **name-empty gated** (`db_manager.go:78`): `(IFNULL(u.name,'')='' AND uv.real_name LIKE ?)` — searchable ONLY when it's the actual display value, so searchable-granularity == visible-granularity. The confirm-oracle on a named member's hidden legal name is closed.
- Applied consistently on **both** `searchMembers` (db.go:194) and `countSearchMembers` (db.go:215) — a guard on the list side but not the count side would still leak the oracle via the count; both are gated.

### Inverse test — ADDED (the one I asked for)
- `Test_Issue434_NamedMemberNotSearchableByRealName`: seeds a member with a visible `user.name` ("Visible Name") AND a distinct hidden `real_name` ("Secret Legal Name"), probes the hidden real_name, asserts `Count==0` + not-found, then sanity-checks the member IS still findable by visible name. This fails without the name-empty gate — a genuine regression net for the "可检索粒度 == 可见粒度" invariant.

### real_name not exposed — confirmed
- The new `IFNULL(uv.real_name,'') as real_name` SELECT feeds the internal `memberSearchModel.RealName` field (model.go:47), consumed only by the shared `resolveMemberDisplayName(name, realName, uid)` fallback. It's folded into `DisplayName()`, NOT a raw JSON field on the response. Bonus: `MemberDetailModel` now delegates to the same `resolveMemberDisplayName` — good DRY consolidation of the list + search fallback chains.

### Collation JOIN — correct AND index-safe (the #567 lesson, applied)
- `memberVerificationJoinOn = "uv.user_id = sm.uid COLLATE utf8mb4_general_ci"` (db.go:177), on both search + count sides. I specifically checked whether this COLLATE defeats an index (the #567 index-defeat class): it does NOT. The pin is on `sm.uid` (the probe/expression side), coercing it to match `uv.user_id`'s collation. `uv.user_id` is the PRIMARY KEY and `user_verification` is explicitly `utf8mb4_general_ci` (and in the OSS-compat repair list). So the join seeks into `uv` by its general_ci PK, and pinning the probe to general_ci enables that PK index rather than breaking it. This is the correct place to pin (probe side, target already normalized) — the opposite of #567, where the pin was on the indexed column itself.

### Minor (informational, out of scope)
- The name-fallback query at `db.go:158` (issue #344 path, not this PR's search) LEFT JOINs `user_verification` with a bare `uv.user_id=sm.uid` (no COLLATE pin). On a MySQL-8 `0900_ai_ci` space_member it could hit the same 1267 the search path just fixed. Pre-existing, separate code path — not this PR's scope, but worth a follow-up to apply the same `memberVerificationJoinOn` const there for consistency.

### Note
- Draft; Build/Test/Vet SKIPPED, check-sprint + pr-title-lint green. Un-draft to run code CI; `mergeStateStatus` stays BLOCKED until then + approvals.

### Praise
- Clean, exact response to the prior review: the fix gates real_name to the display-value condition (not a blunt removal — blank-name members are still searchable by real_name, preserving #434's intent), the inverse test pins the privacy invariant as a true oracle test, and the display fallback was consolidated into one shared function for both list and search. And the collation pin is placed correctly (probe side, index-preserving) — the distinction that matters after #567.

Jerry-Xin (commented):
**Re-review @ `e96691d5` (draft → COMMENT only).** Both blockers from my prior review (`311bde8`) are resolved. Verdict on merits: this would be an **APPROVE** if it weren't a draft.

## 🔴 Prior blocker — RESOLVED ✅
My earlier REQUEST_CHANGES flagged that `uv.real_name` was added as an **unconditional** searchable column in `memberSearchActiveColumns`, letting a named member (non-empty `user.name`) be discovered by substring-probing their hidden verification `real_name` — a search oracle over PII that is never returned in the response.

Delta `311bde8..e96691d5` fixes exactly this:
- `uv.real_name` **removed** from `memberSearchActiveColumns` (now `{u.name, u.username, u.email, RIGHT(u.phone,4), sm.uid}`).
- Re-added in `memberSearchActiveWhere` as a **gated** clause: `(IFNULL(u.name,'')='' AND uv.real_name LIKE ?)`.
- The gate `IFNULL(u.name,'')=''` is byte-aligned with `resolveMemberDisplayName`'s `name != ""` fallback condition, so **searchable granularity == visible granularity**: real_name is matchable only when it is the actual display value. This mirrors the `maskPhone`/`RIGHT(phone,4)` posture.
- New regression test `Test_Issue434_NamedMemberNotSearchableByRealName` pins the invariant (named member + hidden distinct real_name → count=0), and confirms one-way gate (still findable by visible name). Good — this is the exact test I'd have asked for.

## Collation byte-verify (the core check)
I enumerated **every** JOIN/WHERE comparison on the touched columns and byte-verified actual collation on both sides from CREATE TABLE + ALTER migrations (not code comments):

| table.column | collation | source |
|---|---|---|
| `user_verification.user_id` | **utf8mb4_general_ci** | table-level `COLLATE=utf8mb4_general_ci` (`modules/user/sql/20260505000003_user_legacy01.sql`) **and** pinned by compat-repair `CONVERT TO ... utf8mb4_general_ci` (`modules/base/sql/20260512000001_base_oss_compat_repair.sql`, `user_verification` in its list) |
| `space_member.uid` | **DB default** (fresh MySQL8 = utf8mb4_0900_ai_ci; canonical = general_ci) | no column/table COLLATE (`modules/space/sql/20260307000002_space_legacy01.sql`); **not** in compat-repair list |
| `user.uid` | DB default (same as sm.uid) | no COLLATE (`modules/user/sql/20191106000003_user_legacy01.sql`); not repaired |
| `robot.robot_id` | DB default (same as sm.uid) | no COLLATE (`modules/robot/sql/20210926000001_robot_legacy01.sql`); not repaired |

**Fix direction is correct:** `uv.user_id = sm.uid COLLATE utf8mb4_general_ci` coerces `sm.uid` to `general_ci` to match `uv.user_id`'s pinned `general_ci`. On a `0900_ai_ci` deployment the bare compare would raise **1267**; the COLLATE eliminates it and is a no-op on canonical `general_ci`. Confirmed via delta and repo grep.

**Reverse-1267 check — none introduced:**
- `u.uid = sm.uid`: both inherit DB default → aligned, untouched by the `COLLATE` (which is scoped to the `uv` ON-clause operand only). ✅
- `r.robot_id = sm.uid`: both DB default → aligned. ✅
- `uv.user_id = sm.uid COLLATE general_ci`: aligned to general_ci. ✅

Only **two** `space_member↔user_verification` join sites exist repo-wide (grep-confirmed): the new search path (COLLATE-safe) and `queryMembers` (`db.go:158`, the member-list path). No `group_member↔user_verification` join exists.

## 💬 Non-blocking notes
1. **Pre-existing sibling risk (not this PR's regression):** `queryMembers` at `db.go:158` still uses a **bare** `LEFT JOIN user_verification uv ON uv.user_id=sm.uid` with no COLLATE — the exact 1267 exposure #482 tracks, already on `main`. This PR correctly does **not** touch it (one-PR-one-thing) and its own comment references #482. Worth landing the same `memberVerificationJoinOn` on `queryMembers` in the #482 follow-up so the constant covers both sites.
2. Extracting `resolveMemberDisplayName` as a shared pure helper (used by both `MemberDetailModel` and `memberSearchModel`) is the right call — prevents fallback-chain drift. Test coverage (pure + DB-backed RED→GREEN) is solid.

Build + `go vet` on `modules/space` clean locally. Nice work on the gated real_name search and the collation-safe join from the start.

yujiawei (approved):
# Code Review — PR #537 (octo-server)

**Head SHA reviewed:** `e96691d5ff6927ca5b21df9a8d98371e7155f15c`
**Scope:** space member search — display-name fallback + searchable `real_name` + collation-safe verification JOIN (issue #434, P1 slice).

Reviewed the diff line-by-line and the surrounding module (`modules/space/db.go`, `db_manager.go`, `model.go`, `api_member_search.go`, tests), the relevant table schemas/collations, and git history for the sibling collation work. This PR is auth/PII-adjacent, so the privacy and collation dimensions were checked with extra care and against `origin/main`.

## 1. Spec compliance

**Spec: ✅**

The linked issue #434 (P1 slice) asks for two things on `GET /v1/space/:space_id/members/search`:
1. blank `user.name` members should not render an empty name → now falls back via `resolveMemberDisplayName` (`user.name` → `user_verification.real_name` → `"User <uid>"` placeholder), shared with the member-list path. `model.go` `memberSearchModel.DisplayName()` + `api_member_search.go:74`.
2. blank-name members should be findable by real name → `uv.real_name` added to the searchable set (`db.go` `searchMembers`/`countSearchMembers`, `db_manager.go` `memberSearchActiveWhere`).

- **Under-build:** none — both P1 gaps are closed and covered by DB-backed tests.
- **Over-build:** none problematic. The `real_name` search term is deliberately gated to the `user.name`-empty case (`db_manager.go:78`), and the new JOIN carries an explicit `COLLATE`. Both are hardening *of the feature being added*, not out-of-scope features. The PR explicitly defers the #434 P2/P3 follow-ups (avatar, other `IFNULL` endpoints, shared `pkg/displayname`) — good one-PR-one-thing discipline.
- **Deviation:** none.

## 2. Code quality

**Quality: Approved**

Strengths verified independently:

- **Privacy gate is correct and load-bearing.** `real_name` is a searchable term only when `IFNULL(u.name,'')=''`, i.e. only when it is the actual display value. A member with a visible `user.name` cannot have their hidden legal name used as a search oracle (probe substrings → confirm via hit/count). `Test_Issue434_NamedMemberNotSearchableByRealName` pins this and fails if the gate is removed. This matches the existing `maskPhone` "searchable granularity == visible granularity" invariant.
- **No SQL injection.** All keyword input is parameterized (`?` placeholders); `escapeLike` + `likeEscapeClause ESCAPE '\\'` neutralize `%`, `_`, `\`; the new `real_name` clause uses a placeholder; `memberVerificationJoinOn` is a constant, never user-controlled.
- **No JOIN fan-out.** `user_verification` is `PRIMARY KEY (user_id)` (1:1), so the new `LEFT JOIN` cannot duplicate member rows or inflate `countSearchMembers`. `count` and `list` share the same `memberSearchActiveWhere` clause, so pagination totals stay consistent with the returned page.
- **Collation fix is targeted correctly.** `user_verification` is force-normalized to `utf8mb4_general_ci` by the base compat-repair migration, while `space_member` inherits the DB default (`utf8mb4_0900_ai_ci` on a default MySQL 8.0 install). Coercing `sm.uid COLLATE utf8mb4_general_ci` on the `uv` join is exactly the side to fix (it does not disturb the `uv.user_id` PK), and it is a no-op on a canonical `general_ci` DB. The pre-existing `user`/`robot` joins correctly need no `COLLATE` because those tables share `space_member`'s default collation.

No P0/P1 blocking findings.

### Non-blocking notes (P2 / nit / coverage — do not block merge)

- **[P2, pre-existing, out of scope] Member-list path still lacks `COLLATE` on its `user_verification` join.** `queryMembers` (`db.go`, the `SELECT ... LEFT JOIN user_verification uv ON uv.user_id=sm.uid` block) carries the same 1267 exposure this PR fixes for the search path, on a default MySQL 8.0 schema. This PR does **not** touch `queryMembers`, the risk already exists on `main`, and the PR comment honestly points to the separate collation-repair effort tracking it. Flagging only so it isn't lost — please make sure that separate fix lands; ideally both paths converge on one shared collation-safe join predicate.
- **[nit] Robot join future-proofing.** The `robot`↔`space_member` join has no `COLLATE`; today both inherit the same default so there is no 1267 risk, but if `robot` is ever normalized independently it would regress. Optional defensive `COLLATE` there would make the pattern uniform.
- **[nit] Whitespace-only `user.name`.** `resolveMemberDisplayName` treats `"  "` as non-empty (displays whitespace) while the search gate treats it as non-empty too (real_name not searchable). The privacy invariant still holds (visible value is the whitespace, not real_name), so this is a cosmetic UX quirk, best addressed by trimming names at write time rather than here.
- **[coverage] Collation path is not unit-tested.** The CI test DB is created as `general_ci`, so the `COLLATE` clause is present but never exercised against a `0900_ai_ci` mismatch. Errors of this kind are loud in production, but a targeted test (or a shared predicate covered once) would prevent silent regression.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. The change is correct, safe, well-scoped, and well-tested; the remaining items above are advisory and either pre-existing/out-of-scope or cosmetic.

PR #538, opened 2026-07-04
fix(group): async reconcile for orphan groups without IM channel #394

lml2468:
# Bug Fix — Issue #394 — PR #<PR-N>

## 1. Executive Summary
- **Bug**: `group.CreateGroup` creates the IM channel after `tx.Commit()`; if the IM create fails and its best-effort compensating delete *also* fails (or the process crashes between commit and IM create), the committed `group` row survives with no backing IM channel — an orphan clients can never message.
- **Root Cause**: `modules/group/service.go:CreateGroup` — the IM channel lives in WuKongIM (outside the SQL tx), and the post-commit cleanup only logs on failure, so a double-failure/crash leaves a residue that nothing ever reconciles.
- **Fix Approach**: cross-cutting (new `channel_synced` flag + async reconcile worker) — a point fix is impossible per the issue (the channel can't be enrolled in `tx`).
- **Risk Level**: Low
- **PR**: #<PR-N> draft

## 2. Issue Context
- Title: CreateGroup compensating delete is non-atomic → orphan group rows (needs async reconcile)
- Reporter: split out of #381 (refs #377); dispatched via Multica triage
- bug-verified by: self (deterministic reasoning + repro tests; no human gate for auto-dispatched backend bugs)
- Affected: shared `group.CreateGroup` — all create paths (Web / Bot / integration). Low-frequency, not a regression, not exploitable.

## 3. Reproduction
**Environment**: Go 1.26 unit tests (mocked DB + IM); logic path is server-side, no live infra needed for the repro.
**Steps**:
1. Create a group; force `IMCreateOrUpdateChannel` to fail (existing integration test already does this via a dead WuKongIM port).
2. Force the compensating `DELETE` to also fail (DB blip) — or crash between `tx.Commit()` and the IM create.
**Expected**: no `group` row is ever left queryable without a backing IM channel.
**Actual (pre-fix)**: the committed `group`/`group_member` rows survive with no channel; nothing reconciles them → permanent orphan.
**Evidence**: `Test_Issue394_ReconcileRecreatesChannelForOrphanWithMembers` (and siblings) exercise the orphan states; pre-fix there was no reconcile path at all, so the orphan is permanent by construction.

## 4. RCA
**Method**: Fault tree
**Analysis**: IM channel state is external to the SQL transaction → post-commit create needed → create can fail → compensating delete is itself non-transactional, best-effort, log-only → on delete-also-fails or crash-between-commit-and-create, the row persists → no background process re-checks it → permanent orphan.
**Root cause**: absence of a durable, eventually-consistent reconcile for the "committed group without confirmed IM channel" state in `modules/group/service.go:CreateGroup`.

## 5. Fix Description
**Files changed**:
- `modules/group/sql/20260705000001_group_channel_synced.sql` (+72) — add `channel_synced` column (recovery-safe nullable-sentinel migration; existing rows backfilled to 1).
- `modules/group/service.go` (+16) — `CreateGroup` marks `channel_synced=0` in-tx after the group insert, flips to 1 only after a confirmed IM create (best-effort; a failed flip is harmless — reconciler re-syncs).
- `modules/group/db.go` (+50) — `MarkChannelUnsyncedTx` / `MarkChannelSynced` / `QueryUnsyncedGroupNos` / `QueryMemberUIDsWithGroupNo` / `DeleteOrphanGroup`.
- `modules/group/channel_reconciler.go` (+161) — `ChannelReconciler` background worker (mirrors the existing `thread.ArchiveWorker` shape): scans unsynced groups older than a grace period; members present → idempotent `IMCreateOrUpdateChannel` + mark synced; members gone → hard-delete the orphan.
- `modules/group/channel_reconcile_config.go` (+96) — env-driven config, **enabled by default** (`GROUP_CHANNEL_RECONCILE_*`).
- `modules/group/1module.go` (+20) — wire `Start`/`Stop` for the worker into the group module.

**Change summary**: A new `channel_synced` flag makes the orphan state queryable; only `CreateGroup` opts in (writes 0), every other create path and all existing rows default to 1, so behaviour is unchanged elsewhere. A background reconciler closes the window durably and idempotently — surviving both the double-failure and the crash-between-commit-and-IM modes, and consistent whether it re-creates the channel or hard-deletes an unrecoverable (memberless) orphan.

**Does NOT change**: the existing fast-path compensating delete (kept — it still narrows the window on the common case); the `Model` struct (deliberately no `ChannelSynced` field, so other insert paths keep the DB `DEFAULT 1`; dbr scans the unmapped column into a throwaway on `SELECT *`); any API surface or wire contract.

## 6. Regression Tests
|---|---|---|---|---|
| `Test_Issue394_ReconcileRecreatesChannelForOrphanWithMembers` | `channel_reconciler_test.go` | orphan w/ members → idempotent channel recreate + mark synced | FAIL (no reconcile path) | PASS |
| `Test_Issue394_ReconcileHardDeletesMemberlessOrphan` | `channel_reconciler_test.go` | partial-compensation orphan (members gone) → hard-delete | FAIL | PASS |
| `Test_Issue394_ReconcileLeavesUnsyncedWhenIMStillFailing` | `channel_reconciler_test.go` | IM still down → no sync/delete, retry next round | FAIL | PASS |
| `Test_Issue394_ReconcileProcessesMixedBatchAndContinuesOnPerItemError` | `channel_reconciler_test.go` | per-item error isolation | FAIL | PASS |
| `Test_Issue394_ReconcileGracePeriodPassedToQuery` | `channel_reconciler_test.go` | cutoff = now−Grace (skip in-flight creates) | FAIL | PASS |
| `Test_Issue394_ReconcileQueryErrorSurfaces` / `...DisabledConfigNoop` / `...LoadChannelReconcileConfigDefaults` | `channel_reconciler_test.go` | error propagation, disabled no-op, default-on config | FAIL | PASS |

## 7. CI Status
- Linter: not run here (no `golangci-lint` in sandbox); `go vet ./modules/group/...` clean; `go build ./...` clean.
- Unit: `go test ./modules/group/ -run Test_Issue394` → all PASS (mocked, infra-free).
- Integration: the group package's DB/IM-backed tests require MySQL + Redis + WuKongIM (per CLAUDE.md). Not available in this sandbox — they fail identically **before and after** my change (e.g. `TestQueryBotsInvitedByUIDTx_BasicMatch` fails on a clean `git stash` of my work too: `dial tcp 127.0.0.1:5001: connection refused`). Per baseline-diff rule: **new_failures = 0** (all failures are pre-existing/environmental).
- Coverage delta: +reconciler paths covered; N/A precise % (infra-gated suite).

## 8. Deployment Notes
- Migration required?: **yes** — `20260705000001_group_channel_synced.sql` (recovery-safe, re-entrant; ADD COLUMN + backfill existing rows to 1 + tighten to `NOT NULL DEFAULT 1`).
- Config change?: optional — `GROUP_CHANNEL_RECONCILE_ENABLED` (default `true`), `_INTERVAL` (5m), `_GRACE` (5m), `_BATCH` (200, clamped ≤2000).
- Feature flag: reconciler defaults **on** (it's a data-consistency safety net; a healthy system matches zero rows → cheap no-op query per interval). Set `GROUP_CHANNEL_RECONCILE_ENABLED=false` to disable.
- Compatibility: backward-preserved. Existing rows + all non-CreateGroup insert paths → `channel_synced=1`; no read/write contract changes.

## 9. Rollback Plan
**Pattern**: feature-flag-disable (immediate) → revert (full).
**Steps**:
1. Immediate: set `GROUP_CHANNEL_RECONCILE_ENABLED=false` — stops the worker; `CreateGroup` still functions (the extra in-tx flag write and the post-commit mark are harmless no-ops for consumers).
2. Full: revert the PR. The `channel_synced` column can be left in place (unused) or dropped via the migration's `-- +migrate Down`.
**Detection** (trigger rollback): spike in `GroupChannelReconciler` error logs (`recreate IM channel failed` / `delete memberless orphan failed`), unexpected group-delete volume, or IM-side load from reconcile recreate calls.

---

### DIFF
--- a/modules/group/1module.go
+++ b/modules/group/1module.go
@@ -1,6 +1,7 @@
 package group
 
 import (
+	"context"
 	"embed"
 	"fmt"
 	"time"
@@ -30,8 +31,27 @@ func init() {
 		// source_space_* / home_space_* 字段，让 Web/Android/iOS UserInfo 能区分
 		// "同 Space 非好友 → 直接发消息" vs "跨 Space 外部成员 → 仅可在群内交流"。
 		user.RegisterGroupMemberExternalProvider(api.groupService.GetMemberExternalFields)
+
+		// #394 孤儿群对账 worker：收口 CreateGroup「commit 后建频道」失败留下的孤儿群。
+		// workerCancel 每次 Start 重建，给未来 graceful-reload 留余地（当前框架只 Start 一次）。
+		reconciler := NewChannelReconciler(ctx.(*config.Context), LoadChannelReconcileConfig())
+		var reconcilerCancel context.CancelFunc
+
 		return register.Module{
 			Name: "group",
+			Start: func() error {
+				workerCtx, cancel := context.WithCancel(context.Background())
+				reconcilerCancel = cancel
+				reconciler.Start(workerCtx)
+				return nil
+			},
+			Stop: func() error {
+				if reconcilerCancel != nil {
+					reconcilerCancel()
+				}
+				reconciler.Stop()
+				return nil
+			},
 			SetupAPI: func() register.APIRouter {
 				return api
 			},
--- a/modules/group/channel_reconcile_config.go
+++ b/modules/group/channel_reconcile_config.go
@@ -1,0 +1,96 @@
+package group
+
+import (
+	"os"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// ChannelReconcileConfig 孤儿群 IM 频道对账 worker 的运行参数。#394
+type ChannelReconcileConfig struct {
+	// Enabled 总开关。默认 true——这是数据一致性兜底，正常运行下只是每 Interval 跑一次
+	// 廉价的空查询（无 channel_synced=0 行时零动作），默认开启才能让 #394 修复开箱生效。
+	Enabled bool
+	// Interval 两次扫描之间的间隔。<=0 回退默认值。
+	Interval time.Duration
+	// Grace 宽限期：只处理 created_at 早于 now-Grace 的未同步群，避免与仍在途的建群主
+	// 路径抢跑（建群 commit 后尚未标记 synced 的极短窗口）。<0 回退默认值。
+	Grace time.Duration
+	// BatchSize 单轮最多处理的孤儿群数，给 IM/DB 限流。<=0 回退默认值。
+	BatchSize int
+}
+
+const (
+	envReconcileEnabled  = "GROUP_CHANNEL_RECONCILE_ENABLED"
+	envReconcileInterval = "GROUP_CHANNEL_RECONCILE_INTERVAL"
+	envReconcileGrace    = "GROUP_CHANNEL_RECONCILE_GRACE"
+	envReconcileBatch    = "GROUP_CHANNEL_RECONCILE_BATCH"
+
+	defaultReconcileInterval = 5 * time.Minute
+	defaultReconcileGrace    = 5 * time.Minute
+	defaultReconcileBatch    = 200
+	// maxReconcileBatch 防止运维误填巨数把单轮对账变成长时间 IM 风暴。
+	maxReconcileBatch = 2000
+)
+
+// LoadChannelReconcileConfig 从环境变量装载配置。错误 / 越界值一律回退默认值。
+func LoadChannelReconcileConfig() ChannelReconcileConfig {
+	return ChannelReconcileConfig{
+		Enabled:   parseBoolDefault(os.Getenv(envReconcileEnabled), true),
+		Interval:  parsePositiveDuration(os.Getenv(envReconcileInterval), defaultReconcileInterval),
+		Grace:     parseNonNegativeDuration(os.Getenv(envReconcileGrace), defaultReconcileGrace),
+		BatchSize: parseBoundedPositiveInt(os.Getenv(envReconcileBatch), defaultReconcileBatch, maxReconcileBatch),
+	}
+}
+
+// parseBoolDefault: 空值回退 def；显式 "false"/"0" → false，"true"/"1" → true，其它回 def。
+func parseBoolDefault(raw string, def bool) bool {
+	v := strings.ToLower(strings.TrimSpace(raw))
+	switch v {
+	case "":
+		return def
+	case "true", "1":
+		return true
+	case "false", "0":
+		return false
+	default:
+		return def
+	}
+}
+
+func parsePositiveDuration(raw string, def time.Duration) time.Duration {
+	if raw == "" {
+		return def
+	}
+	d, err := time.ParseDuration(raw)
+	if err != nil || d <= 0 {
+		return def
+	}
+	return d
+}
+
+func parseNonNegativeDuration(raw string, def time.Duration) time.Duration {
+	if raw == "" {
+		return def
+	}
+	d, err := time.ParseDuration(raw)
+	if err != nil || d < 0 {
+		return def
+	}
+	return d
+}
+
+func parseBoundedPositiveInt(raw string, def, max int) int {
+	if raw == "" {
+		return def
+	}
+	n, err := strconv.Atoi(raw)
+	if err != nil || n <= 0 {
+		return def
+	}
+	if n > max {
+		return max
+	}
+	return n
+}
--- a/modules/group/channel_reconciler.go
+++ b/modules/group/channel_reconciler.go
@@ -1,0 +1,161 @@
+package group
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+
+	"go.uber.org/zap"
+)
+
+// reconcileDB 抽出 ChannelReconciler 需要的最小 DB 接口，便于单测 mock。
+type reconcileDB interface {
+	QueryUnsyncedGroupNos(before time.Time, limit int) ([]string, error)
+	QueryMemberUIDsWithGroupNo(groupNo string) ([]string, error)
+	MarkChannelSynced(groupNo string) error
+	DeleteOrphanGroup(groupNo string) error
+}
+
+// reconcileIM 抽出重建 IM 频道所需的最小接口（*config.Context 实现之），便于单测 mock。
+type reconcileIM interface {
+	IMCreateOrUpdateChannel(req *config.ChannelCreateReq) error
+}
+
+// ChannelReconciler 周期性扫描 channel_synced=0 的群，收口 CreateGroup「commit 后建频道」
+// 留下的孤儿群：有成员 → 幂等重建 IM 频道并标记 synced；无成员（补偿删除已删成员但群行
+// 残留）→ 硬删除孤儿群行。实现最终一致，闭合 #394 的孤儿窗口。
+type ChannelReconciler struct {
+	cfg ChannelReconcileConfig
+	db  reconcileDB
+	im  reconcileIM
+	now func() time.Time
+
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
+	log.Log
+}
+
+// NewChannelReconciler 构造 worker。生产路径用 group.NewDB 和 config.Context 注入。
+func NewChannelReconciler(ctx *config.Context, cfg ChannelReconcileConfig) *ChannelReconciler {
+	return &ChannelReconciler{
+		cfg: cfg,
+		db:  NewDB(ctx),
+		im:  ctx,
+		now: time.Now,
+		Log: log.NewTLog("GroupChannelReconciler"),
+	}
+}
+
+// Start 启动后台 ticker。Enabled=false 或参数非法时不启动新 goroutine，但仍先停掉可能
+// 存在的旧 goroutine——避免未来热更新留下孤儿 ticker。重复调用幂等。
+func (w *ChannelReconciler) Start(ctx context.Context) {
+	if w.cancel != nil {
+		w.cancel()
+		w.wg.Wait()
+		w.cancel = nil
+	}
+	if !w.cfg.Enabled || w.cfg.Interval <= 0 || w.cfg.BatchSize <= 0 {
+		w.Info("group channel reconciler disabled",
+			zap.Bool("enabled", w.cfg.Enabled),
+			zap.Duration("interval", w.cfg.Interval),
+			zap.Int("batch_size", w.cfg.BatchSize))
+		return
+	}
+	rctx, cancel := context.WithCancel(ctx)
+	w.cancel = cancel
+	w.wg.Add(1)
+	go func() {
+		defer w.wg.Done()
+		t := time.NewTicker(w.cfg.Interval)
+		defer t.Stop()
+		w.Info("group channel reconciler started",
+			zap.Duration("interval", w.cfg.Interval),
+			zap.Duration("grace", w.cfg.Grace),
+			zap.Int("batch_size", w.cfg.BatchSize))
+		for {
+			select {
+			case <-rctx.Done():
+				return
+			case <-t.C:
+				recreated, deleted, err := w.RunOnce(rctx)
+				if err != nil && !errors.Is(err, context.Canceled) {
+					w.Error("group channel reconcile run failed", zap.Error(err))
+					continue
+				}
+				if recreated > 0 || deleted > 0 {
+					w.Info("group channel reconcile run",
+						zap.Int("recreated", recreated),
+						zap.Int("deleted", deleted))
+				}
+			}
+		}
+	}()
+}
+
+// Stop 通知 worker 退出并等待当前 RunOnce 跑完。
+func (w *ChannelReconciler) Stop() {
+	if w.cancel != nil {
+		w.cancel()
+	}
+	w.wg.Wait()
+}
+
+// RunOnce 执行一轮对账：取一批 channel_synced=0 且早于宽限期的群，逐个收口。
+// 返回本轮 (重建频道数, 硬删除孤儿数, err)。单个群失败只记日志、不中断整批——
+// 下一轮 tick 会重试（IMCreateOrUpdateChannel 是 upsert 幂等，重跑安全）。
+func (w *ChannelReconciler) RunOnce(ctx context.Context) (int, int, error) {
+	if w.cfg.BatchSize <= 0 {
+		return 0, 0, nil
+	}
+	before := w.now().Add(-w.cfg.Grace)
+	groupNos, err := w.db.QueryUnsyncedGroupNos(before, w.cfg.BatchSize)
+	if err != nil {
+		return 0, 0, err
+	}
+	var recreated, deleted int
+	for _, groupNo := range groupNos {
+		if err := ctx.Err(); err != nil {
+			return recreated, deleted, err
+		}
+		uids, err := w.db.QueryMemberUIDsWithGroupNo(groupNo)
+		if err != nil {
+			w.Error("reconcile: query members failed", zap.String("groupNo", groupNo), zap.Error(err))
+			continue
+		}
+		if len(uids) == 0 {
+			// 补偿删除部分成功：成员已删、群行残留 → 群无有效成员、不可恢复，硬删除孤儿。
+			if err := w.db.DeleteOrphanGroup(groupNo); err != nil {
+				w.Error("reconcile: delete memberless orphan failed", zap.String("groupNo", groupNo), zap.Error(err))
+				continue
+			}
+			deleted++
+			w.Info("reconcile: hard-deleted memberless orphan group", zap.String("groupNo", groupNo))
+			continue
+		}
+		// 有成员 → 幂等重建 IM 频道，成功后标记 synced。失败留到下轮重试。
+		if err := w.im.IMCreateOrUpdateChannel(&config.ChannelCreateReq{
+			ChannelID:   groupNo,
+			ChannelType: common.ChannelTypeGroup.Uint8(),
+			Subscribers: uids,
+		}); err != nil {
+			w.Error("reconcile: recreate IM channel failed, will retry next round",
+				zap.String("groupNo", groupNo), zap.Error(err))
+			continue
+		}
+		if err := w.db.MarkChannelSynced(groupNo); err != nil {
+			// 频道已建成、仅标记失败：下轮再跑一次（IMCreate 幂等）即可收敛，无害。
+			w.Error("reconcile: mark synced failed after channel recreate",
+				zap.String("groupNo", groupNo), zap.Error(err))
+			continue
+		}
+		recreated++
+		w.Info("reconcile: recreated IM channel for orphan group",
+			zap.String("groupNo", groupNo), zap.Int("subscribers", len(uids)))
+	}
+	return recreated, deleted, nil
+}
--- a/modules/group/channel_reconciler_test.go
+++ b/modules/group/channel_reconciler_test.go
@@ -1,0 +1,244 @@
+package group
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// stubReconcileDB 模拟对账 worker 的 DB 面：一批未同步群 + 每群成员映射，
+// 并记录 MarkChannelSynced / DeleteOrphanGroup 的调用，用于断言收口动作。
+type stubReconcileDB struct {
+	unsynced    []string
+	members     map[string][]string
+	queryErr    error
+	memberErr   map[string]error
+	syncErr     map[string]error
+	deleteErr   map[string]error
+	syncedCalls []string
+	deleteCalls []string
+	gotBefore   time.Time
+	gotLimit    int
+}
+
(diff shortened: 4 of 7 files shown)


OctoBoooot (commented):
## Review: fix(group): async reconcile for orphan groups without IM channel (#538)

**Verdict:** Comment — draft PR; the worker/migration are well-built, but this **reproduces the exact status-blind-scan resurrection blocker from #521** (same author's prior PR). Draft → COMMENT (#513), but this must be fixed before un-drafting.

Risk tier: normal; reviewed on the reconcile-scan-completeness axis (the #521 lesson: enumerate every transition that lands a row in the scan set).

### Blocking-on-un-draft — disbanded/disabled group gets its IM channel resurrected (the #521 finding, reproduced)
Byte-verified the whole chain; no group-status guard anywhere:
- **Scan** (`db.go:591`): `WHERE channel_synced=0 AND created_at < ?` — no `status` filter.
- **disband** (`api.go:88`): soft `UpdateStatusTx(groupNo, GroupStatusDisband, ...)` — does **not** delete `group_member` rows and does **not** set `channel_synced`.
- **Member query** (`db.go:88` `QueryMemberUIDsWithGroupNo`): `WHERE group_no=? AND is_deleted=0` — no group-status filter, so a disbanded group still returns members.
- **`reconcileOne`** (`channel_reconciler.go` RunOnce): `len(uids) > 0` → `IMCreateOrUpdateChannel` recreates. No `GroupStatus == Disband/Disabled` check.

**Failing input:** group created → `channel_synced=0` in-tx → IM-create fails *and* compensating delete fails (the orphan case this worker exists for) → `{Normal, synced=0, members intact}`. Creator disbands → `{Disband, synced=0, members still intact}` (soft disband). After grace, worker scans it, member query returns the still-present members, and **`IMCreateOrUpdateChannel` resurrects the disbanded group's IM channel.** Exactly yujiawei's #521 blocker.

**Fix (same as #521):** add `AND status = <Normal>` to `QueryUnsyncedGroupNos` (only reconcile live groups), OR have disband/disable set `channel_synced=1` so the row leaves the scan set, OR guard `reconcileOne` on live status before recreate. Predicate filter is cleanest. Add a regression test: disbanded group with `channel_synced=0` + members → NOT recreated.

### Also — memberless-orphan hard-delete
`reconcileOne` hard-deletes (`DeleteOrphanGroup`) any `channel_synced=0` group with zero `is_deleted=0` members. A disbanded group whose members *were* cleaned would be hard-deleted by the reconciler rather than by the disband path — probably harmless for a true orphan, but combined with the status gap it means the worker takes destructive action on rows it shouldn't be scanning. Fixing the status filter resolves both (disbanded groups leave the scan set entirely).

### Done right
- **Migration** (`20260705000001_group_channel_synced.sql`): guarded reentrant procedure, `ADD COLUMN ... NULL` then `UPDATE ... WHERE channel_synced IS NULL` backfill-to-1 (existing groups treated synced → no false orphans), reversible Down. No bare CREATE / no deleted-applied-file trap (#529/#531 lessons applied). Clean.
- **Outbox state machine**: `MarkChannelUnsyncedTx` in-tx at create, `MarkChannelSynced` only after IM-create confirms (service.go:1189/1335) — correct, mirrors #521's correct half.
- **Worker**: grace cutoff, batch cap, disabled-by-default config, idempotent `IMCreateOrUpdateChannel`. Structurally sound.

### Net
The migration and outbox mechanics are correct and the worker is well-structured — but the scan must exclude terminated (disbanded/disabled) groups before this un-drafts, or it resurrects channels for groups a user deliberately disbanded. This is the same status-blind-scan defect #521 carried; the fix is the `status=Normal` predicate + a disband-not-resurrected regression test (which the suite currently lacks — its "disabled" test is about config, not group status).

Jerry-Xin (changes requested):
Summary: The PR is in scope for octo-server, but the reconciler has a runtime SQL bug that prevents the fix from working, plus a subscriber-state regression.

**🔴 Blocking**
- 🔴 Critical — `QueryUnsyncedGroupNos` uses `From("group")`, but dbr does not quote table names in `SELECT ... FROM` string inputs. Because `group` is a MySQL reserved word, this builds `FROM group` and the default-on worker will fail every scan instead of reconciling orphans. Existing group reads use `From("`group`")` for this reason. Fix at `modules/group/db.go:74-79`.
- 🔴 Critical — The reconciler recreates IM channels with all `is_deleted=0` members, including blacklisted members, at `modules/group/db.go:83-90` and `modules/group/channel_reconciler.go:125-145`. This conflicts with the existing subscriber contract in `GetSubscribableMemberUIDs`, which excludes blacklisted users to prevent unauthorized real-time reads. Use normal-status subscribers for `IMCreateOrUpdateChannel`, while keeping a separate “has any non-deleted member” check if needed to avoid deleting blacklisted-only groups.

**💬 Non-blocking**
- 🟡 Warning — `channel_synced=0 AND created_at < ? ORDER BY created_at LIMIT ?` has no supporting index in the migration (`modules/group/sql/20260705000001_group_channel_synced.sql:29-41`). Since the worker defaults on, this can become a periodic full scan/filesort on large group tables. Consider adding an index such as `(channel_synced, created_at)`.
- 🔵 Suggestion — The tests mock the DB interface, so they missed the generated SQL issue. Add a small DB/query-builder test or an integration test covering `QueryUnsyncedGroupNos` against MySQL.

**✅ Highlights**
- The durable `channel_synced` marker is a good fit for the non-atomic SQL/IM boundary.
- The worker shape follows the existing thread archive worker lifecycle pattern.
- Targeted `Test_Issue394` tests pass locally; broader DB-backed tests are still environment-gated.

OctoBoooot (commented):
## Follow-up on Jerry-Xin's #538 CR — byte-verified: both 🔴 are real; I missed them

Jerry-Xin found two real defects my review didn't cover (I focused on the #521 status-blind-scan carry-forward, which is also real). Byte-verified both; crediting and adding the decisive evidence for #1. Draft PR so verdict stays COMMENT (#513), but these + the status-scan gap are all blocking-on-un-draft.

### 🔴#1 — reserved word `group` unquoted in the new SELECT → query fails at runtime (real, verified)
`QueryUnsyncedGroupNos` (db.go:74-75): `d.session.Select("group_no").From("group")` — **unquoted**. `dbr` passes `.From(string)` through raw → `FROM group`, and `group` is a MySQL reserved word → syntax error. **Proof from the existing codebase:** every working SELECT on this table quotes it — `.From("`+"`group`"+`")` (db.go:312, 319) or via `dbr.I`. `.Update("group")` works unquoted only because dbr's UpdateBuilder quotes the table itself; the SELECT path does not. This one new query dropped the backticks the rest of the file uses. Net: the reconciler's **core scan query fails on every run**. Fix: `.From("`+"`group`"+`")` (match the existing SELECTs). This is why Jerry-Xin's 🟡 "no SQL-generation test" matters — `stubReconcileDB` mocks the DB, so the broken real SQL is never exercised.

### 🔴#2 — reconciler subscribes blacklisted members (real, verified)
`QueryMemberUIDsWithGroupNo` (db.go:84-89): `WHERE group_no=? AND is_deleted=0` — filters only soft-delete, **not** member status. The file's own comment (db.go:154) states `status=GroupMemberStatusNormal` is what distinguishes "黑名单 / 已退出但 is_deleted 仍为 0 的成员". So the reconciler recreates the IM channel with **blacklisted / left-but-not-deleted members re-subscribed**, violating the subscriber-status convention the rest of the module enforces. Fix: add `AND status = <Normal>` (and likely `is_external` handling) to match the authoritative subscriber query.

### Still also blocking (my earlier finding) — status-blind scan resurrects disbanded groups
Independently of the above: the scan has no group-status filter and disband is a soft status change that leaves members + `channel_synced=0`, so a disbanded orphan gets its channel recreated (the #521 blocker, reproduced). Fix overlaps with #2's direction — `QueryUnsyncedGroupNos` should also gate `status=Normal`.

### Net
Three real must-fix-before-un-draft items, all in the DB/query layer: (1) quote `group` in the SELECT — currently a hard runtime failure; (2) exclude non-Normal (blacklisted/left) members from the subscriber list; (3) exclude non-Normal groups from the scan (disband-resurrect). The migration + outbox state machine + worker structure remain sound. Jerry-Xin's 🟡s (add `(channel_synced, created_at)` index; add a real-SQL integration test) are both worth doing — the missing SQL test is precisely what let #1 through. Credit to Jerry-Xin for the SQL-quoting and blacklist findings, which I missed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #538 (octo-server)

## Summary

This PR closes #394 by adding an async background reconciler (`ChannelReconciler`) that detects and resolves orphan groups — groups whose DB row was committed but whose WuKongIM channel was never created (due to post-commit IM failure, compensating-delete failure, or process crash between commit and channel creation). The fix introduces a `channel_synced` column (0 = pending reconciliation, 1 = confirmed), marks it 0 inside the `CreateGroup` transaction via `MarkChannelUnsyncedTx`, and runs a periodic ticker that either recreates the IM channel (when members exist) or hard-deletes the orphan group row (when all members are gone).

The design is sound: the migration uses a safe 3-step procedure, the reconciler is per-item error-tolerant with graceful-period protection against racing with in-flight `CreateGroup`, and `IMCreateOrUpdateChannel` idempotency makes retries safe. No P0/P1 issues found; three P2 observations below.

## Verification

- ✅ **Migration safety** — 3-step procedure (ADD NULL → backfill WHERE NULL → MODIFY NOT NULL DEFAULT 1) is reentrant and idempotent; `information_schema` guards prevent double-ADD/double-MODIFY.
- ✅ **Data flow: MarkChannelUnsyncedTx** — placed inside the `CreateGroup` transaction (`service.go:1191`), after group row insert and before member inserts. If tx rolls back, `channel_synced=0` is rolled back with the group row.
- ✅ **Other group creation paths excluded correctly** — `event.go:185` and `:360` create IM channels *inside* their transactions (rollback on failure), and never set `channel_synced=0`. `AddGroup` (`service.go:200`) uses `db.Insert` without `MarkChannelUnsyncedTx` — the `Model` struct has no `ChannelSynced` field, so INSERT takes DB DEFAULT 1. Verified: no false-positive orphan detection on non-CreateGroup paths.
- ✅ **Reconciler lifecycle** — `Start`/`Stop` in `1module.go:42-54` correctly derive worker context from `context.Background()`, store cancel func, and call `reconciler.Stop()` (which cancels + `wg.Wait()`) on module stop. Re-entrant `Start` cleans up prior goroutine first.
- ✅ **Concurrency safety** — single goroutine with ticker; `RunOnce` is synchronous inside the `select`, so no concurrent batch execution. Ticker drops missed ticks when channel buffer is full.
- ✅ **Grace period** — default 5 minutes prevents racing with in-flight `CreateGroup`. `before = now.Add(-Grace)` is passed to `QueryUnsyncedGroupNos`, verified by `Test_Issue394_ReconcileGracePeriodPassedToQuery`.
- ✅ **IM idempotency** — `IMCreateOrUpdateChannel` is documented as upsert-idempotent; reconciler relies on this for retry safety. Verified by `Test_Issue394_ReconcileLeavesUnsyncedWhenIMStillFailing`.
- ✅ **Per-item error isolation** — single group failure logs + `continue`, does not abort the batch. Verified by `Test_Issue394_ReconcileProcessesMixedBatchAndContinuesOnPerItemError`.
- ✅ **Test coverage** — 8 test cases covering happy path (recreate with members, hard-delete memberless), IM failure retry, mixed batch with per-item errors, grace period propagation, query error surfacing, disabled config noop, and config defaults/clamping.

## Findings

No P0/P1 issues. Three P2 items below — none blocking.

### P2 — DeleteOrphanGroup is non-atomic (`modules/group/db.go:93-100`)

Two separate DELETE statements (group_member then group) without a wrapping transaction. If `group_member` delete succeeds but `group` row delete fails, the group persists with zero members until the next reconcile tick. On retry, `QueryMemberUIDsWithGroupNo` returns empty → `DeleteOrphanGroup` retries → `group_member` delete is a no-op (already empty) → `group` delete succeeds. Self-healing within one tick interval, but the transient state (group row with no members) is observable by other readers during that window.

Consider wrapping both deletes in a transaction for atomicity, or adding a comment documenting the retry-recovery guarantee.

### P2 — Grace=0 is a valid but unsafe configuration (`modules/group/channel_reconcile_config.go:84`)

`parseNonNegativeDuration` accepts `Grace=0`, which removes the grace period entirely. If `GROUP_CHANNEL_RECONCILE_GRACE=0`, the reconciler processes groups created at `now()`, racing with `CreateGroup`'s post-commit IM channel creation. Since `IMCreateOrUpdateChannel` is idempotent, no data is lost — but the reconciler may create a channel with committed-but-partial members before `CreateGroup`'s own call completes with the full `realMemberUIDs` list.

Consider enforcing a minimum grace floor (e.g., 30s) or logging a warning when Grace < 1 minute to make misconfiguration harder.

### P2 — No index on channel_synced column (`modules/group/sql/20260705000001_group_channel_synced.sql:38`)

The migration adds `channel_synced` but no index. `QueryUnsyncedGroupNos` runs `SELECT group_no FROM group WHERE channel_synced=0 AND created_at < ? ORDER BY created_at LIMIT ?` every tick. On a large `group` table, this is a full table scan each interval. In steady state, `channel_synced=0` rows should be very rare, but the scan cost is paid regardless of match count.

Consider adding a composite index `(channel_synced, created_at)` to keep the background query O(matches) rather than O(table).

- **C1 — symmetric paths**: CreateGroup sets `channel_synced=0` in tx; event.go paths create channels inside tx (no `channel_synced=0`); AddGroup uses DEFAULT 1. Verified via grep on all `InsertTx`/`Insert`/`IMCreateOrUpdateChannel` call sites. Clear.
- **C2 — control-flow ordering**: Single goroutine, ticker-driven, `RunOnce` synchronous. No concurrent batch execution possible. Clear.
- **C3 — authorization boundary**: Backend worker with no API exposure. N/A.
- **C4 — container/member state cascade**: N/A — no auth/lifecycle change.
- **C5 — build ≠ runtime**: SQL migration verified against standard MySQL 3-step safe pattern. Clear.
- **C6 — governance/docs**: N/A.

## Data-flow tracing

- `channel_synced` value → `MarkChannelUnsyncedTx` (tx-scoped UPDATE) → commit → `QueryUnsyncedGroupNos` (SELECT WHERE channel_synced=0) → reconciler processes. Verified: the value flows correctly from CreateGroup tx to reconciler query.
- Subscriber list → `QueryMemberUIDsWithGroupNo` (SELECT uid FROM group_member WHERE is_deleted=0) → `IMCreateOrUpdateChannel(Subscribers=uids)`. Verified: returns all non-deleted members, matching what CreateGroup's `realMemberUIDs` would contain post-commit.
- `before` cutoff → `w.now().Add(-w.cfg.Grace)` → passed to `QueryUnsyncedGroupNos(before, limit)`. Verified by test assertion.

## Verdict: APPROVED

Well-designed eventual-consistency fix for #394. The reconciler correctly identifies orphan groups via `channel_synced=0`, the grace period prevents racing with in-flight group creation, and per-item error isolation ensures one group's failure doesn't block the batch. The migration is safe and reentrant. Three P2 observations (non-atomic delete, Grace=0 config, missing index) are non-blocking — addressing any or all would strengthen the fix but none introduces a correctness or safety issue.

yujiawei (changes requested):
# Code Review — PR #538 (octo-server)

Reviewed at head SHA `09c47e8f0fcda9fbf05f89cffc684e2ff3b387a0`. Built (`go build ./modules/group/...`), vetted, and ran the `Test_Issue394*` suite (all green) locally; also rendered the exact SQL against `gocraft/dbr/v2 v2.7.5` to verify the finding below.

**Verdict: CHANGES_REQUESTED** — one P0 (the reconciler cannot run in production) plus two P1s.

The overall design is sound and closes #394 the right way: a write-once `channel_synced` flag makes the orphan state queryable, only `CreateGroup` opts in (writes `0`), all other insert paths and existing rows default to `1`, and a background worker mirrors the already-proven `thread.ArchiveWorker` shape. The in-tx `MarkChannelUnsyncedTx` failure path correctly rolls back (deferred `RollbackUnlessCommitted`), and the migration is the same recovery-safe nullable-sentinel pattern already shipped in `20260629000001_group_is_named.sql`. The problem is not the design — it's that the core query never executes.

---

## P0 — `QueryUnsyncedGroupNos` fails on every run: unquoted reserved word `group`

`modules/group/db.go:74-80`

```go
_, err := d.session.Select("group_no").
    From("group").                       // <-- not backtick-quoted
    Where("channel_synced=0 AND created_at < ?", before).
    OrderDir("created_at", true).
    Limit(uint64(limit)).
    Load(&groupNos)
```

`dbr`'s `SelectStmt` does **not** quote a string table name — the source literally reads `// FIXME: no quote ident` and then `buf.WriteString(table)` (`select.go`). This is asymmetric with `Update`/`DeleteFrom`/`InsertInto`, which all call `QuoteIdent` (hence the backticks are added for you there). Rendered with the pinned `dbr v2.7.5` + MySQL dialect, this produces:

```sql
SELECT group_no FROM group WHERE (channel_synced=0 AND created_at < ?) ORDER BY created_at ASC LIMIT 200
```

`group` is a MySQL **reserved word** (`GROUP BY`), so `FROM group` unquoted is a parse error — **ERROR 1064** on every execution. `RunOnce` calls this as its first DB op each tick and returns early on error, so **zero orphans are ever reconciled** — the feature this PR ships is non-functional in the default (enabled) config, while degrading silently (the ticker keeps logging the error and running).

Every other `SELECT … FROM` the `group` table in this same file uses ``From("`group`")`` (e.g. db.go:312, 319, 470); this new query is the lone deviation.

**Fix:** ``From("`group`")``. (The neighbouring `Update("group")` / `DeleteFrom("group")` are fine as-is because those builders auto-quote.)

---

## P1 — No index backs the reconcile scan → full-table scan + filesort every 5 minutes

`modules/group/db.go:74-80` + `modules/group/sql/20260705000001_group_channel_synced.sql`

The `group` table has indexes only on `id` (PK), `group_no` (unique), and `creator`. Neither `channel_synced` nor `created_at` is indexed, and this migration adds the column but **no index**. The scan runs by default every 5 minutes (`defaultReconcileInterval`, enabled by default), so MySQL must read every row to evaluate `WHERE channel_synced=0 AND created_at < ?` and then filesort for `ORDER BY created_at` — `LIMIT 200` applies only after the sort. The PR description's "healthy system ⇒ cheap empty query" is not true for scan cost: with no index, proving zero matches still reads the whole table, continuously contending with live group inserts as the table grows.

**Fix:** add a composite index in this migration, e.g. `CREATE INDEX idx_group_channel_synced ON `group` (channel_synced, created_at);`. With `channel_synced=1` for essentially all rows, this makes the steady-state query an instant empty range read.

## P1 — DB layer has zero real-SQL coverage (why the P0 slipped through CI)

`modules/group/channel_reconciler_test.go:17-66`

`stubReconcileDB` replaces the entire `reconcileDB` interface, so the five new SQL methods never render or execute real SQL. The tests validate `RunOnce` orchestration well, but by construction cannot catch: the P0 above; the missing index; the deliberate "`Model` omits `ChannelSynced`, dbr scans the unmapped column into a throwaway on `SELECT *`" assumption that all group reads depend on; and `DeleteOrphanGroup`'s two-step delete. Please add one DB-backed test (real MySQL / test server): migrate → `CreateGroup` writes `0` → `QueryUnsyncedGroupNos` returns it → recreate/delete path, and explicitly assert an existing `SELECT * FROM group` read still works with the new column present.

---

## P2 / advisory (non-blocking)

- **`DeleteOrphanGroup` is a non-transactional two-step delete** (`db.go:95-101`): deletes `group_member` then `group` on the session (no tx). A failure between the two leaves a memberless `group` row (`channel_synced=0`). It self-heals next round, but wrapping both in one tx removes the transient window.
- **Hard-delete trusts a stale read** (`channel_reconciler.go:130-138`): the delete decision is made from `QueryMemberUIDsWithGroupNo` read earlier, and `DeleteOrphanGroup` then deletes unconditionally. Safe today only because `channel_synced=0` is write-once (set solely in `CreateGroup`) and `CreateGroup` guarantees ≥1 member. That's a load-bearing, undocumented cross-file invariant. Consider a compare-and-delete (`WHERE channel_synced=0` and a member-count guard in the DELETE) so a future caller mistake can't turn this into data loss.
- **Grace cutoff mixes clocks** (`channel_reconciler.go:115`): `before = now().Add(-Grace)` uses the app clock, but `created_at` is MySQL `CURRENT_TIMESTAMP`. Under app/DB timezone or skew mismatch the grace window distorts. Computing the cutoff with the DB clock (`created_at < NOW() - INTERVAL ? SECOND`) keeps both sides on one clock.
- **Reentrancy claims aren't actually guarded** (`channel_reconciler.go` Start/Stop): the comments advertise hot-reload / idempotent-restart safety, but `w.cancel`/`w.wg` are touched without a mutex and `Stop()` doesn't reset `w.cancel` to nil. Fine under today's single serial Start/Stop, but either add a mutex or soften the comments.

## Security note (this PR is flagged security-sensitive)

No new external attack surface: the worker takes no request-derived input, `group_no` comes only from the internal scan, batch size is bounded (`maxReconcileBatch=2000`), and `IMCreateOrUpdateChannel` is an idempotent upsert. The one thing worth a human's eye is the **hard-delete path** (P2 above) — it is the only destructive, irreversible operation, and its safety rests entirely on the write-once `channel_synced` invariant. Please confirm no other code path (now or planned) ever writes `channel_synced=0` on an already-populated group.

## Coverage / blind spots

- The full-scan severity depends on the production `group` row count (couldn't EXPLAIN against a live DB here) — the conclusion is from the confirmed absence of any supporting index.
- The completeness of "only `CreateGroup` writes `channel_synced=0`" was verified by grep across `modules/`, but a raw-SQL writer in a very large file could in principle exist; the hard-delete safety depends on this holding.
- Migration backfill (`UPDATE … WHERE channel_synced IS NULL`) is un-chunked; safe at single-instance startup (runs before the service is exposed), but on a very large table it's one long UPDATE / large binlog event — consider chunking if the table is big.

PR #539, opened 2026-07-06
fix(message): space-ize DM recents in per-Space conversation sync

an9xyz:
## Summary

DMs active in more than one Space went missing from the per-Space recent list (`POST /v1/conversation/sync?space_id=`). The server passed WuKongIM's raw `recents` (the physical channel's global-latest window, Space-agnostic) straight through, so `recents[0]` could belong to another Space. The web/extension SpaceFilter keys its show/hide decision on `recents[0].payload.space_id`, so it hid the whole conversation — even though the server had *kept* it (via `presence` / `hasSpaceMsg`). This narrows each kept Person DM's `recents` to the queried Space so `recents[0]` always belongs to it.

## Related Issue

No tracked issue number (captured on the deployed test env). Related to the DM per-Space isolation work in #484 and the preview fix in #532 — neither narrowed conversation-sync `recents`, which is where the client keys visibility.

## Linked Spec

Load-bearing (space isolation). Root cause, fix rationale, and the alternatives considered are documented in this PR body + the COMPREHENSION section below; no separate `brief.md`.

## Changes

- Add `spaceizePersonRecents` (`modules/message/space_filter.go`), called in `syncUserConversation` right after `FilterConversationsBySpace`: it narrows each kept **Person DM**'s `recents` to the queried Space by **reusing `filterPersonMessagesBySpace`** — the same rule already applied to `/v1/message/channel/sync` history (exact match, `untagged = default` bare-DM convention, system-bot rule 4) — so the two paths can't drift.
- When the Space has no in-window message, `recents` collapses to empty: the client renders a visible, preview-less row (verified: empty recents does **not** trigger hiding; preview comes from `space_last_message`). `recents` is deliberately **not** backfilled from `SpaceLastMessage` (that fallback message bypasses `channelOffset`/`deviceOffset` and lacks the enrichment every real recents entry has).
- System bots (botfather / fileHelper / u_10000) are narrowed on the same footing as their history + preview/unread; they stay visible via `EnsureSystemBotsPresent` + the client's system-bot exemption.
- `last_msg_seq` / `last_client_msg_no` left untouched (WuKongIM incremental-sync cursor; no client reads them for Space filtering).

## Testing

- `go test ./modules/message/ -run TestSpaceizePersonRecents` — 10 unit cases pass (cross-Space narrowing, default-Space untagged match, non-default drops untagged, system-bot rule 4, order preservation, window-miss → empty, nil-safety). These carry no build tag, so **they run in CI**.
- `go vet ./modules/message/` clean on default and `-tags integration`.
- i18n linters (`lint-direct-error-response`, `lint-unregistered-code`) + the legacy-response source guard pass (no error-handling surface touched).
- **Manually verified end-to-end** against local MySQL 8.0 + Redis (WuKongIM faked for the sync call) through the real `/v1/conversation/sync` handler, as a red/green check: **without** the fix the DM's Space-A `recents` leaks the Space-B message and `TestFix_ConversationSync_SpaceizesDMRecents` **FAILS**; **with** the fix it **PASSES**.
- ⚠️ **CI does not run the `//go:build integration` e2e repro.** CI's Test job runs `go test` without `-tags integration` (`ci.yml`), so `TestFix_ConversationSync_SpaceizesDMRecents` (like the existing `dm_space_last_message_repro_test.go`) is a **manual** regression check, not a CI-gated one. The CI-gated coverage for this change is the unit suite above.

- [x] Manually verified — red/green end-to-end via the real `/v1/conversation/sync` handler against local MySQL + Redis.

## COMPREHENSION

   In the Space-filtered conversation-sync response, *after* the keep/drop decision, it rewrites each kept Person DM's `recents` to contain only messages belonging to the queried Space (reusing `filterPersonMessagesBySpace`). This makes `recents[0]` — the field the client's SpaceFilter reads to decide show/hide — always belong to the current Space. Group/thread recents, the conversation-level cursor, and the keep decision are untouched.

   Consumers of DM `recents` in this response now see a Space-narrowed list. For a DM with no in-window current-Space message, `recents` is empty → client shows a visible preview-less row, preview falls back to `space_last_message`. System bots with only untagged history now show empty recents in every Space (intended: matches rule 4 / #532; they stay visible via `EnsureSystemBotsPresent`). **Ordering constraint:** must run *after* `fillPersonSpaceUnread` (reads full recents for `space_last_message` / `space_unread`) and `FilterConversationsBySpace` (reads full recents for `hasSpaceMsg`) — running earlier would empty recents and break both. Isolation is strengthened, not weakened: cross-Space message content is stripped, fail-closed to empty when a current-Space message can't be resolved.

   Red/green end-to-end: `TestFix_ConversationSync_SpaceizesDMRecents` builds one DM with a Space-A message and a globally-latest Space-B message, queries each Space through the real route, and asserts `recents` carries only that Space's messages and `recents[0].space_id == queried space` — it FAILS with the fix reverted and PASSES with it in place (run locally against MySQL + Redis). Unit tests pin the predicate branches (incl. system-bot rule 4, default-untagged, window-miss → empty). Client-side behavior (recents[0] drives hide; empty recents is safe; `space_last_message` drives preview) was traced in `octo-web` `SpaceService.tsx` / `Convert.ts` / `Model.tsx`.

## Checklist

- [ ] Updated documentation — n/a (behavior documented in code comments + this PR)

### DIFF
--- a/modules/message/api_conversation.go
+++ b/modules/message/api_conversation.go
@@ -894,6 +894,13 @@ func (co *Conversation) syncUserConversation(c *wkhttp.Context) {
 			syncUserConversationResps, filterSpaceID, loginUID, co.ctx, co.groupService,
 		)
 
+		// 把保留下来的 Person DM 的 Recents 空间化：只保留当前 Space 的消息，使
+		// recents[0]（客户端 SpaceFilter 读取的「最后一条消息」）恒属于当前 Space，
+		// 修复「DM 在按空间的最近会话列表中缺失」—— 客户端此前按全局最新（跨 Space）
+		// 消息的 space_id 把整条会话过滤掉。必须在 fillPersonSpaceUnread（算
+		// SpaceLastMessage）与 FilterConversationsBySpace（keep 判定读原始 Recents）之后。
+		spaceizePersonRecents(syncUserConversationResps, filterSpaceID, defaultSpaceID)
+
 		// YUJ-216 / GH#1280: 系统 Bot（botfather 等）在所有 Space 都应可见。
 		// IM 核心按 version 增量返回 conversation，若系统 Bot 在此次 window 内
 		// 没有新消息，Space 过滤后就会缺席；移动端没有前端兜底，会直接丢失 entry。
--- a/modules/message/dm_recents_spaceize_repro_test.go
+++ b/modules/message/dm_recents_spaceize_repro_test.go
@@ -1,0 +1,92 @@
+//go:build integration
+
+package message
+
+// End-to-end regression guard for "DM conversation missing from the per-Space
+// recent list", observed via POST /v1/conversation/sync?space_id=:space_id.
+//
+// Scenario (mirrors the real capture): one physical DM channel with messages in
+// two non-default Spaces, whose globally-latest message belongs to Space B. The
+// client keys its per-Space show/hide decision on the LAST message in recents
+// (recents[0].payload.space_id). Before the fix the server passed WuKongIM's raw
+// recents through untouched, so Space A's response carried Space B's message at
+// recents[0] and the client hid the whole conversation — even though the server
+// kept it (hasSpaceMsg/presence).
+//
+// After the fix (spaceizePersonRecents), each kept DM's recents is narrowed to the
+// queried Space, so recents[0] always belongs to that Space.
+//
+// Reuses the shared httptest WuKongIM fake + helpers from
+// dm_space_last_message_repro_test.go / conversation_recent_filter_e2e_test.go.
+
+import (
+	"testing"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestFix_ConversationSync_SpaceizesDMRecents drives the real handler through the
+// registered /v1/conversation route and asserts the per-Space recents narrowing.
+func TestFix_ConversationSync_SpaceizesDMRecents(t *testing.T) {
+	const (
+		dmChannel   = "dm-peer-recents"
+		spaceA      = "space-a-recents"      // non-default
+		spaceB      = "space-b-recents"      // non-default, holds the globally-latest msg
+		spaceDefaul = "space-default-recents" // earliest membership → default Space
+	)
+
+	now := time.Now()
+	// History: seq 11 tagged Space A, seq 12 tagged Space B (globally latest, like
+	// the captured "1111"/668cc9 message).
+	msgA := dmMsgRepro(dmChannel, 11, now.Add(-2*time.Hour).Unix(),
+		`{"type":1,"content":"hi-from-A","space_id":"`+spaceA+`"}`)
+	msgB := dmMsgRepro(dmChannel, 12, now.Add(-1*time.Hour).Unix(),
+		`{"type":1,"content":"1111","space_id":"`+spaceB+`"}`)
+
+	convs := []*config.SyncUserConversationResp{
+		dmIMConvMulti(dmChannel, now.Add(-1*time.Hour).Unix(), 100,
+			[]*config.MessageResp{msgA, msgB}),
+	}
+
+	s, ctx := setupConvSyncE2E(t, convs)
+	// Default Space MUST have the earliest created_at (GetUserDefaultSpaceIDE =
+	// earliest membership); Space A / B are non-default.
+	seedSpaceMemberRepro(t, ctx, spaceDefaul, testutil.UID, "2020-01-01 00:00:00")
+	seedSpaceMemberRepro(t, ctx, spaceA, testutil.UID, "2020-06-01 00:00:00")
+	seedSpaceMemberRepro(t, ctx, spaceB, testutil.UID, "2020-07-01 00:00:00")
+
+	// ---- Space A: the DM is present and its recents carry ONLY the Space-A msg ----
+	convA := findConv(callConvSyncSpace(t, s, spaceA), dmChannel)
+	require.NotNil(t, convA, "DM active in Space A must appear in Space A's recent list")
+	require.NotEmpty(t, convA.Recents, "Space A recents must carry the Space-A message")
+	for _, m := range convA.Recents {
+		assert.Equal(t, spaceA, spaceIDOf(m),
+			"Space A recents must not contain any other Space's message")
+	}
+	assert.Equal(t, spaceA, spaceIDOf(convA.Recents[0]),
+		"recents[0] — the client's per-Space hide signal — belongs to Space A")
+
+	// ---- Space B: symmetric, only its own message ----
+	convB := findConv(callConvSyncSpace(t, s, spaceB), dmChannel)
+	require.NotNil(t, convB, "DM active in Space B must appear in Space B's recent list")
+	require.NotEmpty(t, convB.Recents)
+	for _, m := range convB.Recents {
+		assert.Equal(t, spaceB, spaceIDOf(m),
+			"Space B recents must not contain any other Space's message")
+	}
+}
+
+// spaceIDOf reads payload.space_id from a decoded recents message ("" if absent).
+func spaceIDOf(m *MsgSyncResp) string {
+	if m == nil || m.Payload == nil {
+		return ""
+	}
+	if sid, ok := m.Payload["space_id"].(string); ok {
+		return sid
+	}
+	return ""
+}
--- a/modules/message/space_filter.go
+++ b/modules/message/space_filter.go
@@ -442,6 +442,55 @@ func personConvHasSpaceMessages(conv *SyncUserConversationResp, targetSpaceID st
 	return false
 }
 
+// spaceizePersonRecents 把每条 Person (DM) 会话的 Recents 收敛到当前 Space —— 直接复用
+// filterPersonMessagesBySpace（/v1/message/channel/sync 的同一份 DM 空间归属规则：精确匹配，
+// 或「无标签 = 默认 Space」的裸 DM 约定；系统 Bot 走 rule 4，只保留精确打标当前 Space 的消息）。
+// 复用而非再抄一遍判定逻辑，避免与历史过滤漂移出两套「哪条消息属于哪个 Space」的口径。
+//
+// 背景（本 PR 修复的会话缺失 bug）：客户端 SpaceFilter 按会话「最后一条消息」（recents[0]）的
+// payload.space_id 判定该会话归属哪个 Space；而服务端此前把 WuKongIM 的 Recents（物理频道的
+// 全局最新窗口，与 Space 无关）原样透传。于是一条在 Space A、Space B 都有消息的 DM，其全局最新
+// 一条属于 B，客户端就把整条会话过滤出 Space A —— 即便服务端 decideConvKeepInSpace 已判定它属于
+// Space A（presence / hasSpaceMsg），也会在客户端消失。这与历史过滤 filterPersonMessagesBySpace
+// 是同一类问题，只是会话同步的 Recents 之前漏做了空间化。
+//
+// 空间化后窗口里每一条都属于当前 Space，recents[0]（无论客户端把哪端当「最新」）恒为当前 Space
+// 的消息，可见性判定不再被跨 Space 的最新消息带偏。过滤后为空（当前 Space 的消息在窗口外）时
+// Recents 保持为空：客户端读不到跨 Space 的 recents[0]，不再误隐藏该会话（它已被
+// decideConvKeepInSpace 判定属于当前 Space，会以空预览行正常展示 —— 时间/排序走会话级 timestamp、
+// 未读走 space_unread、预览走 space_last_message）。不用 SpaceLastMessage 回填 recents：预览已由
+// space_last_message 承载，空 recents 也不会触发隐藏；回填反而会把一条绕过 channelOffset/
+// deviceOffset、缺富化字段的兜底消息塞进 recents[0]，与其余 recents 口径不一致。
+//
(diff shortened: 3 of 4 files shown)


OctoBoooot (commented):
## Review: fix(message): space-ize DM recents in per-Space conversation sync (#539)

**Verdict:** Approve with comments — clean; holding at COMMENT only because the code-CI `Test` check has not reported green yet.

The change narrows each kept Person DM's `Recents` to the queried Space so `recents[0]` (what the web/extension SpaceFilter reads to show/hide a conversation) always belongs to the current Space. It delegates to `filterPersonMessagesBySpace` — the same DM space-attribution rule already vetted on `/v1/message/channel/sync` — so no new predicate is introduced, and the sysbot rule-4 handling comes along via the internal `IsSystemBot(channelID)` check.

**Ordering is correct** (`api_conversation.go`): `spaceizePersonRecents` runs *after* `fillPersonSpaceUnread` (computes `SpaceLastMessage` from raw Recents) and `FilterConversationsBySpace` (keep-decision reads raw Recents). Scrubbing before either would have broken the preview or the keep-decision — the code and its comment pin this dependency.

**Tests are exemplary.** The matrix covers the cases that matter and every trap flagged on sibling PRs (#490/#532):
- default-Space untagged kept; non-default untagged dropped (no fail-open across Spaces);
- degenerate case empties rather than leaks a cross-Space message (fail-closed);
- sysbot rule-4 parity, including the default-Space untagged-sysbot-dropped case that pins the `isSysBot` flag is actually threaded;
- group/community recents untouched (channel_id-layer isolation); empty-`spaceID` no-op; order preserved.
- A repro test drives the real handler over `POST /v1/conversation/sync?space_id=`, so `api_conversation.go` (flagged by the coverage sniff) is exercised end-to-end.

**Before merge:** `Test`/`Build`/`Vet` are still pending — once code CI is green this is APPROVE-ready on the merits. `check-sprint` is red (governance CI, does not bar the code); `mergeStateStatus` stays BLOCKED until the Sprint field is set and reviews clear. `reviewDecision` is currently REVIEW_REQUIRED.

### Praise
- Reusing `filterPersonMessagesBySpace` instead of re-deriving the attribution rule keeps the sysbot/untagged/default-Space semantics identical to the history path — the single-source-of-truth choice is what makes this safe.
- The fail-closed degenerate case (empty rather than leak) and the explicit ordering comment show the subtle failure modes were understood, not stumbled into.

Jerry-Xin (approved):
The PR is in scope for `octo-server` and correctly addresses Space-scoped DM conversation sync without introducing blocking issues.

🔴 Blocking  
None.

💬 Non-blocking  
🟡 Warning: I could not complete the integration repro locally because the shared test DB migration state failed before the test body ran: `unknown migration in database`. The focused unit suite passed: `go test ./modules/message/ -run TestSpaceizePersonRecents`.

✅ Highlights  
🔵 Suggestion: The new call order in [api_conversation.go](/Users/mlamp/.openclaw/tmp/pr-review-zhf9xofm/repo/modules/message/api_conversation.go:889) is appropriate: `fillPersonSpaceUnread` and `FilterConversationsBySpace` still see raw recents, then [space_filter.go](/Users/mlamp/.openclaw/tmp/pr-review-zhf9xofm/repo/modules/message/space_filter.go:482) narrows only kept Person DM recents.

✅ Reusing `filterPersonMessagesBySpace` at [space_filter.go](/Users/mlamp/.openclaw/tmp/pr-review-zhf9xofm/repo/modules/message/space_filter.go:575) is the right architectural choice; it keeps conversation-sync recents aligned with channel-history filtering, including default-Space untagged behavior and system-bot rule parity.

✅ Test coverage is strong for the changed behavior, especially cross-Space pruning, default untagged handling, system-bot filtering, non-person no-op, order preservation, and nil safety in [space_recents_spaceize_test.go](/Users/mlamp/.openclaw/tmp/pr-review-zhf9xofm/repo/modules/message/space_recents_spaceize_test.go:49).

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #539 (octo-server)

## Summary

This PR fixes a bug where DM conversations disappear from a Space's recent conversation list when the globally-latest message on the shared physical channel belongs to a different Space. The root cause is that WuKongIM's Recents window is returned raw (space-agnostic), so `recents[0].payload.space_id` — the client's per-Space show/hide signal — can point at the wrong Space, causing the client to hide a conversation the server has already decided to keep.

The fix introduces `spaceizePersonRecents`, which narrows each kept Person conversation's Recents to the queried Space by reusing the proven `filterPersonMessagesBySpace` (the same rules used by `/v1/message/channel/sync` for message history). The call is correctly ordered after `fillPersonSpaceUnread` (which reads original Recents for space_last_message / space_unread) and `FilterConversationsBySpace` (which reads original Recents for the keep decision). 4 files changed, +333/-0.

## Verification

Static analysis only at head `9f5e1bb37ab9f87f01f6192735e4020d7acd38db`; build and tests not executed in this environment.

- ✅ **Call ordering correct** — `api_conversation.go:891-910`: `fillPersonSpaceUnread` (line 891) → `FilterConversationsBySpace` (line 893) → `spaceizePersonRecents` (line 902) → `EnsureSystemBotsPresent` (line 910). Both upstream consumers of raw Recents run before the narrowing.
- ✅ **In-place mutation safe** — after `spaceizePersonRecents`, the only remaining step is `EnsureSystemBotsPresent`, which iterates existing conversations only to build a `present` map and append placeholders for missing bots (`space_filter.go:515-545`). It does not read Recents of existing entries. The response is serialized immediately after.
- ✅ **Delegation correct** — `spaceizePersonRecents` delegates entirely to `filterPersonMessagesBySpace` (`space_filter.go:575`), reusing the same 5-rule DM-space matching logic (exact match, untagged-in-default, drop-untagged-non-default, drop-sysbot-untagged, drop-cross-space). No duplicated判定 logic that could drift.
- ✅ **Nil safety** — `spaceizePersonRecents` guards `conv == nil` (`space_filter.go:487`) and `filterPersonMessagesBySpace` guards `m == nil` (`space_filter.go:585`) and `len(msgs) == 0` (`space_filter.go:577`).
- ✅ **Fail-closed** — when no current-Space message is in the Recents window, Recents collapses to empty rather than leaking cross-Space content. Preview is carried by `SpaceLastMessage` (computed earlier by `fillPersonSpaceUnread`), not by Recents.
- ✅ **Non-Person channels untouched** — `ChannelType != Person` is skipped (`space_filter.go:488`). Group/CommunityTopic isolation operates at the channel_id layer.
- ✅ **Empty spaceID noop** — `if spaceID == ""` early-returns (`space_filter.go:485`), preserving behavior for non-space-filtered requests.

## Findings

- **C1 — Dual-path parity**: v2 sidebar (`api_sidebar.go:281`) uses `FilterRawConversationsBySpace` + `EnsureSystemBotsPresentRaw` but does not surface `Recents` at all (grep confirms zero references to `Recents` in `api_sidebar.go`). The sidebar builds its own recent items through a different path. No parity gap.
- **C2 — Control-flow ordering / nesting reuse**: `spaceizePersonRecents` is called at exactly one site (`api_conversation.go:902`). No nested or secondary call-site that could double-filter or reorder.
- **C3/C4 — Authorization boundary**: This PR does not change authorization, permissions, or access control. It narrows data already within an authorized response.
- **C5 — Build/runtime**: N/A — static analysis only; no build artifacts or packaging changes.
- **C6 — Governance/docs**: N/A — no policy or documentation changes.
- **Data-flow trace**: `conv.Recents` (local `SyncUserConversationResp`) is populated by WuKongIM upstream → passed through `fillPersonSpaceUnread` (reads, does not mutate Recents) → `FilterConversationsBySpace` (reads Recents via `personConvHasSpaceMessages` for keep decision, does not mutate) → `spaceizePersonRecents` (replaces Recents with filtered subset) → `EnsureSystemBotsPresent` (does not read Recents) → JSON response. Each stage consumes what it needs before the next mutates.
- **Test coverage**: 10 unit tests in `space_recents_spaceize_test.go` cover the core fix, window-miss, degenerate empty case, default-space untagged match, non-default untagged drop, system-bot rule-4 parity, non-person passthrough, empty-spaceID noop, order preservation, and nil safety. Integration test `dm_recents_spaceize_repro_test.go` exercises the full handler through the registered `/v1/conversation` route with a real WuKongIM fake, asserting both Space A and Space B views symmetrically.
- **defaultSpaceID consistency**: `api_conversation.go:834` fetches `defaultSpaceID` once and early-returns on error (line 836–839), so the value passed to `spaceizePersonRecents` is always valid when reached. `FilterConversationsBySpace` fetches its own copy internally (with fail-closed sentinel on error) — a pre-existing design choice, not introduced by this PR.

## Verdict: APPROVED

Clean, well-scoped fix that correctly addresses the root cause. The implementation reuses proven filtering logic rather than duplicating判定 rules, the call ordering is safe, and the test coverage is thorough — including both unit tests for every edge case and an integration repro that drives the real handler. No issues found.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Implementation is correct: call ordering verified, data flow traced end-to-end, dual-path parity confirmed (v2 sidebar doesn't use Recents), fail-closed behavior verified, comprehensive test coverage including integration repro.

yujiawei (approved):
# Code Review — PR #539 (octo-server)

Reviewed at head SHA `9f5e1bb37ab9f87f01f6192735e4020d7acd38db` against merge-base `ae1b51e1`. Diff: +333 / 0 across 4 files (1 impl, 1 comment-only handler hook, 2 test files).

## 1. Spec / scope compliance

**Compliant.** The change does exactly what the PR describes and nothing more.

- **Under-build:** none. The root cause (server passes WuKongIM's Space-agnostic global-latest `recents` through untouched, so `recents[0]` — the field the client's SpaceFilter keys show/hide on — can belong to another Space) is addressed by the new `spaceizePersonRecents`, wired into `syncUserConversation` right after `FilterConversationsBySpace`.
- **Over-build:** none. Scope is confined to narrowing kept **Person (DM)** `recents`. Group / CommunityTopic recents are untouched (isolation there is enforced at the `channel_id` layer). `last_msg_seq` / `last_client_msg_no` (the WuKongIM incremental cursor) are deliberately left global — correct, since the client keys Space filtering on message-level `payload.space_id`, not the cursor.
- **Deviation:** none. The fix **reuses** `filterPersonMessagesBySpace` — the identical predicate already applied to `/v1/message/channel/sync` history (`modules/message/api.go:1370`) — rather than re-deriving the "which message belongs to which Space" rule. This keeps the two paths from drifting and means the system-bot rule 4, default-untagged rule 2, and non-default-drops-untagged rule 3 all carry over for free.

## 2. Code quality

**Approved.**

Strengths:
- **Correct ordering, and it is enforced by construction, not just documented.** `spaceizePersonRecents` runs *after* `fillPersonSpaceUnread` (which reads raw recents to compute `space_last_message` / `space_unread`) and *after* `FilterConversationsBySpace` (which reads raw recents for the `hasSpaceMsg` keep decision). Narrowing earlier would empty recents and break both. The call site (`api_conversation.go:902`) sits in exactly the right place.
- **Fail-closed isolation.** When the current Space has no in-window message, `recents` collapses to empty rather than leaking the cross-Space `recents[0]`. Empty recents is a visible, preview-less row (preview falls back to `space_last_message`); it is *not* backfilled from `SpaceLastMessage` — a deliberate choice that avoids surfacing an offset-bypassing, under-enriched fallback message as `recents[0]`.
- **Nil-safe and cheap:** early return on empty `spaceID`, per-conv nil guard, per-message nil guard inside the reused helper.
- **Test coverage is strong for the size of the change:** 10 unit cases (cross-Space narrowing, default-untagged match, non-default drop, system-bot rule-4 parity incl. the default-Space case that pins `isSysBot` threading, order preservation, window-miss → empty, non-Person untouched, empty-spaceID no-op, nil-safety) plus an `//go:build integration` end-to-end repro driving the real route. `go test ./modules/message/ -run TestSpaceizePersonRecents` passes and `go vet ./modules/message/` is clean locally.

### Findings from an independent second look — all cleared, no blockers

Two aliasing/mutation angles were checked closely because this is a load-bearing isolation path:

- **Shared-cache mutation of `conv.Recents` (P1 candidate → not a bug):** the response objects are freshly built per request via `newSyncUserConversationResp` (`api_conversation.go:721`), which allocates a new `[]*MsgSyncResp` and new `*MsgSyncResp` for every entry. The only per-handler cache (`syncConversationResultCacheMap`) stores `[]string` seq strings, not response structs. `filterPersonMessagesBySpace` also returns a *new* slice. So reassigning `conv.Recents` mutates only request-local state — no cross-request corruption.
- **`fillPersonSpaceUnread` reading the same slice later (→ not a bug):** it runs *before* the narrowing and reads the raw `config.MessageResp` recents, a distinct object graph from the response `MsgSyncResp` recents this helper rewrites. No read-after-narrow hazard.

Non-blocking notes (P2 / nit):
- **Micro:** `spaceizePersonRecents` iterates the full kept list; since `FilterConversationsBySpace` already returns survivors only, this is correct — it just also re-touches system-bot placeholders added later by `EnsureSystemBotsPresent`… except that runs *after* this call, so no wasted work. Nothing to change.

## 3. Verdict

**APPROVE.** Spec-compliant, correctly ordered, fail-closed, well tested. No P0/P1 issues.

## 4. Suggestions (optional, non-blocking)

- Consider a one-line assertion in the integration repro that `space_last_message` remains populated for the window-miss case, so the "empty recents but preview still shows" contract is pinned end-to-end (currently pinned only at unit level via `TestSpaceizePersonRecents_WindowMissEmptiesRecents`).

## 5. Additional observations

- The client-side contract this fix relies on — (a) `recents[0]` drives per-Space hide, (b) empty recents does not trigger hiding, (c) `space_last_message` drives preview — lives in `octo-web` (`SpaceService.tsx` / `Convert.ts` / `Model.tsx`) and is out of this diff's scope; it could not be re-verified here. The server change is self-consistent and strictly strengthens isolation, so even if a client detail differs the change is not a regression, but the end-to-end fix depends on that client behavior holding.
- The `//go:build integration` repro needs MySQL + Redis + WuKongIM and runs only in CI; it was not executed locally.

PR #540, opened 2026-07-06
fix(space): align member-join column collations to fix MySQL 1267 in GetSpaceMembers

boris-clark:
## Summary

Fixes a MySQL `1267 Illegal mix of collations` error that aborts `GetSpaceMembers` and returns a 400, causing every space member to render as a raw uid and the manager member list to fail to load.

## Related Issue

Root cause tracked in XIN-459; this PR resolves XIN-476.

## Linked Spec

XIN-476 — align the `GetSpaceMembers` join-key collations at the schema level (idempotent migration, no per-query `COLLATE` band-aid).

## Root cause

`queryMembers` (`modules/space/db.go`) LEFT JOINs `space_member.uid` against three columns:

```sql
LEFT JOIN user u                ON u.uid = sm.uid
LEFT JOIN user_verification uv  ON uv.user_id = sm.uid
LEFT JOIN robot r               ON r.robot_id = sm.uid
```

`user_verification` pins `utf8mb4_general_ci` in its DDL, while the tables created without an explicit `COLLATE` inherit the server default. On a fresh MySQL 8 instance that default is `utf8mb4_0900_ai_ci`, so `uv.user_id = sm.uid` compares two different collations and MySQL aborts the whole statement with 1267 -> `err.server.space.query_failed` (400).

## Changes

- `modules/space/sql/20260707000001_align_member_join_collation.sql` — idempotent migration pinning all four uid-family join keys (`space_member.uid`, `user.uid`, `robot.robot_id`, `user_verification.user_id`) to `utf8mb4_general_ci` (production's existing collation). Column type / nullability / default / comment are unchanged — only the collation moves. `MODIFY COLUMN` to a fixed collation is a no-op once already aligned, so the migration is safe to re-run.
- `modules/space/db_member_collation_repro_test.go` — regression test that forces the collation drift, asserts `queryMembers` fails with 1267, applies the shipped migration file verbatim, then asserts the query succeeds with the `user.name -> user_verification.real_name` fallback intact and every join key aligned.

All four join keys must share one collation, not just the two named in the report: pinning only `space_member.uid` and `user_verification.user_id` would shift the mismatch onto the `user`/`robot` joins. `utf8mb4_general_ci` is chosen because it is what production already uses, so the migration is a no-op there and only converges divergent dev/local databases.

## Testing

```
$ go test ./modules/space/ -run TestQueryMembersCollationMismatch1267 -v
--- PASS: TestQueryMembersCollationMismatch1267 (0.15s)

$ go test ./modules/space/
ok  	github.com/Mininglamp-OSS/octo-server/modules/space
```

Verified red->green locally against MySQL 8.0.46: the test reproduces error `1267 (HY000): Illegal mix of collations (utf8mb4_0900_ai_ci,IMPLICIT) and (utf8mb4_general_ci,IMPLICIT)` before the migration and passes after. Also confirmed the migration applies cleanly on a fresh-boot database.

- [x] Manually verified

## COMPREHENSION

1. **What does this change actually do** to the load-bearing path? It normalizes the collation of the uid-family columns that `GetSpaceMembers` joins on, so `space_member.uid` and the three joined columns are all `utf8mb4_general_ci`. Column semantics (type, length, nullability, default) are untouched; only the collation metadata changes, and the values are ASCII uids so no ordering/uniqueness behavior shifts.

2. **What could break** because of it (dependents + failure mode)? The migration rebuilds the indexes on `space_member.uid`, `user.uid`, `robot.robot_id`. On a production DB already on `utf8mb4_general_ci` the `MODIFY` is a no-op. The only behavioral risk would be a case-sensitivity change, but both `utf8mb4_0900_ai_ci` and `utf8mb4_general_ci` are case/accent-insensitive and the columns hold ASCII uids, so equality/uniqueness results are unchanged.

3. **How do you know it works** (specific test/repro/trace)? `TestQueryMembersCollationMismatch1267` forces the drift, asserts the exact 1267 error string, applies the shipped SQL file (parsed from disk, not a hand-copied copy), and asserts the query then succeeds. The full `modules/space` suite passes.

## Checklist

- [ ] Updated documentation

### DIFF
--- a/modules/space/db_member_collation_repro_test.go
+++ b/modules/space/db_member_collation_repro_test.go
@@ -1,0 +1,121 @@
+package space
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// collationOf returns the COLLATION_NAME of a single column in the test schema.
+func collationOf(t *testing.T, table, column string) string {
+	t.Helper()
+	var name string
+	_, err := testCtx.DB().SelectBySql(
+		"SELECT COLLATION_NAME FROM information_schema.columns "+
+			"WHERE table_schema = DATABASE() AND table_name = ? AND column_name = ?",
+		table, column,
+	).Load(&name)
+	require.NoError(t, err)
+	return name
+}
+
+func mustExec(t *testing.T, sql string) {
+	t.Helper()
+	_, err := testCtx.DB().UpdateBySql(sql).Exec()
+	require.NoError(t, err, sql)
+}
+
+// alignUpStatements reads the shipped migration file and returns its "+migrate
+// Up" statements. The repro test executes exactly what production runs, so the
+// green half proves the migration on disk — not a hand-copied variant — resolves
+// the 1267.
+func alignUpStatements(t *testing.T) []string {
+	t.Helper()
+	raw, err := sqlFS.ReadFile("sql/20260707000001_align_member_join_collation.sql")
+	require.NoError(t, err)
+	body := string(raw)
+	if i := strings.Index(body, "-- +migrate Down"); i >= 0 {
+		body = body[:i]
+	}
+	body = strings.TrimSpace(strings.TrimPrefix(strings.TrimSpace(body), "-- +migrate Up"))
+
+	// Drop comment / blank lines first so a ';' inside a comment does not split
+	// a statement, then split the remaining SQL on ';'.
+	var kept []string
+	for _, ln := range strings.Split(body, "\n") {
+		s := strings.TrimSpace(ln)
+		if s == "" || strings.HasPrefix(s, "--") {
+			continue
+		}
+		kept = append(kept, ln)
+	}
+
+	var stmts []string
+	for _, chunk := range strings.Split(strings.Join(kept, "\n"), ";") {
+		if stmt := strings.TrimSpace(chunk); stmt != "" {
+			stmts = append(stmts, stmt)
+		}
+	}
+	require.Len(t, stmts, 4, "migration should carry four ALTER statements")
+	return stmts
+}
+
+// TestQueryMembersCollationMismatch1267 reproduces XIN-459 / XIN-476: when the
+// GetSpaceMembers join keys drift onto different collations (as they do on a
+// fresh MySQL 8 dev DB whose server default is utf8mb4_0900_ai_ci), queryMembers
+// aborts with MySQL error 1267 and the endpoint returns 400. Applying the
+// collation-alignment migration makes every join key share utf8mb4_general_ci,
+// after which the same query succeeds and the name-fallback chain still works.
+func TestQueryMembersCollationMismatch1267(t *testing.T) {
+	_, _, err := setup(t)
+	require.NoError(t, err)
+
+	// Force the local-style drift: put the migration-created join keys on the
+	// MySQL 8 server default (utf8mb4_0900_ai_ci) while user_verification keeps
+	// the utf8mb4_general_ci its DDL pins. Only the uv.user_id = sm.uid
+	// comparison then mixes collations — exactly the reported failure shape.
+	mustExec(t, "ALTER TABLE `space_member` MODIFY `uid` VARCHAR(40) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci NOT NULL DEFAULT ''")
+	mustExec(t, "ALTER TABLE `user` MODIFY `uid` VARCHAR(40) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci NOT NULL DEFAULT ''")
+	mustExec(t, "ALTER TABLE `robot` MODIFY `robot_id` VARCHAR(40) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci NOT NULL DEFAULT ''")
+	mustExec(t, "ALTER TABLE `user_verification` MODIFY `user_id` VARCHAR(40) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL")
+	require.Equal(t, "utf8mb4_0900_ai_ci", collationOf(t, "space_member", "uid"))
+	require.Equal(t, "utf8mb4_general_ci", collationOf(t, "user_verification", "user_id"))
+
+	const spaceId = "sp-collfix"
+	seedMemberSearchSpace(t, spaceId, testutil.UID)
+	const uid = "u-collfix"
+	seedFallbackUser(t, uid, "")
+	seedFallbackVerification(t, uid, "Zhang Wei")
+	seedMemberSearchMember(t, spaceId, uid, 0, 1)
+
+	// RED: mismatched collation on the uv join aborts the whole query with 1267.
+	_, err = testSpaceDB.queryMembers(spaceId, testutil.UID, 1, 100)
+	require.Error(t, err, "expected MySQL 1267 before collation alignment")
+	assert.Contains(t, strings.ToLower(err.Error()), "illegal mix of collations", err.Error())
+
+	// Apply the shipped migration verbatim.
+	for _, stmt := range alignUpStatements(t) {
+		mustExec(t, stmt)
+	}
+
+	// GREEN: every join key now shares one collation, so the query succeeds and
+	// the user.name -> user_verification.real_name fallback is intact.
+	members, err := testSpaceDB.queryMembers(spaceId, testutil.UID, 1, 100)
+	require.NoError(t, err, "query must succeed after collation alignment")
+	m, ok := findMemberDetail(members, uid)
+	require.True(t, ok, "seeded member missing from result")
+	assert.Equal(t, "Zhang Wei", m.DisplayName(), "real_name fallback must survive the fix")
+
+	for _, c := range []struct{ table, column string }{
+		{"space_member", "uid"},
+		{"user", "uid"},
+		{"robot", "robot_id"},
+		{"user_verification", "user_id"},
+	} {
+		assert.Equal(t, "utf8mb4_general_ci", collationOf(t, c.table, c.column),
+			"%s.%s should be aligned to utf8mb4_general_ci", c.table, c.column)
+	}
+}
--- a/modules/space/sql/20260707000001_align_member_join_collation.sql
+++ b/modules/space/sql/20260707000001_align_member_join_collation.sql
@@ -1,0 +1,35 @@
+-- +migrate Up
+
+-- Fix MySQL 1267 (Illegal mix of collations) in GetSpaceMembers.
+--
+-- queryMembers (modules/space/db.go) LEFT JOINs space_member.uid against three
+-- columns: user.uid, user_verification.user_id and robot.robot_id. On a fresh
+-- MySQL 8 instance every table created without an explicit COLLATE inherits the
+-- server default (utf8mb4_0900_ai_ci), while user_verification pins
+-- utf8mb4_general_ci in its own DDL. Comparing two different collations makes
+-- MySQL abort the whole statement with error 1267, which surfaces to the client
+-- as a 400 (err.server.space.query_failed) and leaves every member name showing
+-- as a raw uid.
+--
+-- Pin every uid-family join key to one explicit collation (utf8mb4_general_ci,
+-- the collation already used in production and by user_verification) so the
+-- comparison is well-defined in every environment. Fixing the columns at the
+-- schema level is the durable fix; a per-query COLLATE clause would only paper
+-- over a single call site and leave the rest of the schema drifting.
+--
+-- MODIFY COLUMN to a fixed collation is idempotent: once a column already is
+-- utf8mb4_general_ci the statement is a no-op, so re-running the migration is
+-- safe. Column type / nullability / default / comment are kept identical to the
+-- original DDL so only the collation changes.
+
+ALTER TABLE `space_member`      MODIFY `uid`      VARCHAR(40) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL DEFAULT '' COMMENT '成员uid';
+ALTER TABLE `user`              MODIFY `uid`      VARCHAR(40) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL DEFAULT '' COMMENT '用户唯一ID';
+ALTER TABLE `robot`             MODIFY `robot_id` VARCHAR(40) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL DEFAULT '' COMMENT '机器人ID';
+ALTER TABLE `user_verification` MODIFY `user_id`  VARCHAR(40) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL COMMENT 'OCTO 用户 UID';
+
+-- +migrate Down
+
+-- No-op: the pre-alignment collation was environment-dependent (the MySQL server
+-- default at table-creation time), so there is no single meaningful value to
+-- revert to. Leaving the join keys aligned on utf8mb4_general_ci is safe and is
+-- the intended steady state.


Jerry-Xin (dismissed):
The PR is relevant to `octo-server`: it fixes a real `modules/space` schema/query failure path for `GetSpaceMembers`.

💬 Non-blocking

🟡 Warning: [modules/space/sql/20260707000001_align_member_join_collation.sql:20](/Users/mlamp/.openclaw/tmp/pr-review-i4_32lmm/repo/modules/space/sql/20260707000001_align_member_join_collation.sql:20) says only collation changes, but lines 25-27 also add SQL column comments to columns whose original DDL used inline `--` comments, not `COMMENT`. This is harmless metadata drift, but the migration comment/PR text should not claim comments are unchanged.

🔵 Suggestion: [modules/space/db_member_collation_repro_test.go:35](/Users/mlamp/.openclaw/tmp/pr-review-i4_32lmm/repo/modules/space/db_member_collation_repro_test.go:35) manually parses migration SQL. It is fine for this simple file, but using `sql-migrate`’s parser like other migration tests would keep the “production verbatim” guarantee if this migration later gains `StatementBegin` blocks or semicolons inside SQL bodies.

✅ Highlights

The migration addresses the schema-level root cause instead of adding query-local `COLLATE` clauses, and it aligns all join keys used by [modules/space/db.go:157](/Users/mlamp/.openclaw/tmp/pr-review-i4_32lmm/repo/modules/space/db.go:157)-159, avoiding shifting the mismatch to another join.

The regression test is strong: it forces the drift, verifies the failing path, applies the shipped migration, then confirms both successful querying and the display-name fallback.

Verification run:

`git diff --check -- modules/space/db_member_collation_repro_test.go modules/space/sql/20260707000001_align_member_join_collation.sql`

`go test ./modules/space/ -run TestQueryMembersCollationMismatch1267 -count=1 -v`

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #540 (octo-server)

## Summary

This PR fixes MySQL error 1267 (Illegal mix of collations) in `GetSpaceMembers` / `queryMembers` by aligning four uid-family join columns to `utf8mb4_general_ci`. The mismatch arises on fresh MySQL 8 instances where tables without explicit `COLLATE` inherit the server default (`utf8mb4_0900_ai_ci`) while `user_verification` pins `utf8mb4_general_ci` in its table-level DDL. The fix is a schema-level `MODIFY COLUMN` migration plus a RED/GREEN repro test that reads the shipped migration from the embedded FS and verifies it resolves the error. The approach is correct — fixing at the schema level is the durable fix, not a per-query `COLLATE` clause.

## Verification

- ✅ **Join column coverage** — traced `queryMembers` at `modules/space/db.go:146`; the three LEFT JOINs (`u.uid=sm.uid`, `uv.user_id=sm.uid`, `r.robot_id=sm.uid`) match the four columns in the migration exactly.
- ✅ **DDL fidelity** — compared migration `MODIFY` definitions against original DDLs in `20260307000002_space_legacy01.sql`, `20191106000003_user_legacy01.sql`, `20210926000001_robot_legacy01.sql`, `20260505000003_user_legacy01.sql`. Column type (`VARCHAR(40)`), nullability (`NOT NULL`), defaults (`DEFAULT ''`), and comments are faithful to the originals; only the `CHARACTER SET` / `COLLATE` is added.
- ✅ **Idempotency** — `MODIFY COLUMN` to the same collation is a no-op on MySQL; safe to re-run.
- ✅ **Migration discovery** — `//go:embed sql` in `1module.go:12` embeds the entire `sql/` directory; the new file `20260707000001_align_member_join_collation.sql` will be auto-discovered by the migration framework.
- ✅ **Other query paths covered** — `searchMembers` (`db.go:172`), `countSearchMembers` (`db.go:198`), and `queryMembersAdmin` (`db_manager.go:211`) all JOIN on `u.uid=sm.uid` and/or `r.robot_id=sm.uid`; the schema-level fix covers these without additional code changes.
- ✅ **Test structure** — RED/GREEN pattern using `alignUpStatements` to parse and apply the shipped migration verbatim from `sqlFS`. Seed helpers (`seedMemberSearchSpace`, `seedFallbackUser`, `seedFallbackVerification`, `seedMemberSearchMember`) and `findMemberDetail` / `DisplayName()` all exist in sibling test files within the same package.
- ✅ **No conflicting migrations** — grepped all `.sql` and `.go` files; no existing migration re-MODIFYs these columns with a different collation.

Static analysis only at head `24cd05897336eac21965bddedb43dda4fae94c8b`; build and tests not executed in this environment.

## Findings

### P2 — Cross-module schema ownership (`modules/space/sql/20260707000001_align_member_join_collation.sql:25`)

The migration lives in `modules/space/sql/` but `ALTER`s `user.uid`, `robot.robot_id`, and `user_verification.user_id` — tables owned by the `user` and `robot` modules. If a future migration in those owning modules re-creates or `MODIFY`s these columns without explicit `COLLATE utf8mb4_general_ci`, the drift silently regresses. The fix is correct and idempotent today, but the canonical collation pin should ideally live with each table owner's DDL.

Consider either (a) adding the `COLLATE` pin to the owning modules' DDL migrations so the definition is self-contained, or (b) adding a CI lint that greps uid-family column definitions for explicit `COLLATE` to prevent drift.

## Blind-Spot Checklist (security_sensitive)

- **C1 (dual-path parity)**: N/A — single-direction migration, no paired create/remove path.
- **C2 (control-flow ordering)**: N/A — no logic flow changes.
- **C3 (auth boundary)**: N/A — no new endpoints or tool exposure. Collation change does not alter string comparison semantics for ASCII UIDs.
- **C4 (auth lifecycle / container-member cascade)**: N/A — no auth lifecycle changes.
- **C5 (build/note ≠ runtime)**: N/A — no build/packaging changes.
- **C6 (governance/docs self-consistency)**: N/A — no governance doc changes.

## Cross-Round Blocker Re-Check (R6)

## Verdict: APPROVED

Clean, well-scoped fix with solid repro test coverage. The migration is correct, idempotent, and faithfully preserves original column definitions. The cross-module schema ownership note is non-blocking — the fix works today, and the suggestion is a maintainability improvement for future schema evolution.

[Octo-Q] verdict: APPROVE — No P0/P1 defects. Migration correctly aligns all four join-key collations targeted by `queryMembers`. The P2 cross-module ownership note is advisory, not blocking.

OctoBoooot (changes requested):
## Review: align member-join column collations to fix MySQL 1267 (#540)

**Verdict:** Request changes — the migration fixes the `GetSpaceMembers` 1267 but introduces a **new, symmetric 1267** in the group↔robot bot-membership joins on the exact environment it targets (fresh MySQL 8). CI note: `check-sprint` is governance-red; code CI (Build/Test/Vet/Lint) still pending.

Risk: this is a **schema migration** — treated as migration-tier regardless of the path-based classifier (`normal`).

### Blockers
- `modules/space/sql/20260707000001_align_member_join_collation.sql:27` — **flipping `robot.robot_id` to `utf8mb4_general_ci` breaks the `robot.robot_id = group_member.uid` joins.** `group_member.uid` has no explicit collation (CREATE TABLE `modules/group/sql/20191106000002_group_legacy01.sql:37`, no table CHARSET/COLLATE, no later `MODIFY`), so on a fresh MySQL 8 it inherits the server default `utf8mb4_0900_ai_ci` — the same premise this migration relies on for the columns it fixes. `robot.robot_id` is likewise `0900_ai_ci` pre-migration (CREATE TABLE `modules/robot/sql/20210926000001_robot_legacy01.sql:9`, no charset clause).
  - Failing case: fresh MySQL 8. **Before:** `r.robot_id (0900) = gm.uid (0900)` → matches. **After:** `r.robot_id (general_ci) = gm.uid (0900)` → error 1267 at `modules/group/db.go:814` (`QueryBotsInvitedByUIDTx`), `:838` (`QueryBotUIDsOwnedByUIDs`), `:855` (`QuerySecondOldestMemberExcludingBotsOf`), and `modules/robot/db.go:224` (`queryRobotUIDsInGroup`).
  - The migration's own thesis ("align at the schema level so the whole family stops drifting") is undercut: it aligns exactly the four `queryMembers` columns and leaves sibling `uid`-family columns (`group_member.uid`) drifting against a column it just moved.
  - Direction: extend the alignment to the full `uid`-family join graph — at minimum add `group_member.uid` to the same `general_ci` alignment — and add a repro that exercises a group↔robot query after the migration, not only `queryMembers`.

### Major
- **Incomplete `uid`-family audit.** The blocker above is one confirmed instance; the fix should enumerate every column that joins the four being flipped and align the whole closure, or the next fresh-MySQL-8 join hits the same wall. `space_member.uid` and `robot.robot_id` both participate in cross-module joins beyond `queryMembers` — audit them explicitly before freezing this migration.
- **Lock cost on `MODIFY COLUMN`.** A collation change is a table rebuild (`ALGORITHM=COPY`), which locks. `user.uid`'s flip is a true no-op (its DDL already pins `general_ci`, `modules/user/sql/20191106000003_user_legacy01.sql:140`) so that one is cheap, but `MODIFY COLUMN` on a large `user`/`space_member` table can still trigger a rebuild on some MySQL versions even when the definition is unchanged. Add a one-line note on expected lock behavior / table sizes, or gate behind an online-DDL tool.

### Praise
- The repro test is exemplary: `alignUpStatements` reads the shipped `.sql` file and executes it verbatim (not a hand-copied variant), asserts the RED 1267 before and GREEN + `real_name` fallback after, and pins every column's post-state collation via `information_schema`. That's exactly the right way to prove a migration.
- The `Down` no-op is correctly reasoned and documented (pre-alignment collation was environment-dependent, so there's no single value to revert to) — acceptable per reversible-or-justified.

### Notes
- `reviewDecision: REVIEW_REQUIRED` / `mergeStateStatus: BLOCKED`; not mergeable regardless. This overlaps the #537 collation-JOIN work (`uv.user_id = sm.uid COLLATE utf8mb4_general_ci`) — coordinate so the two don't re-introduce drift against each other.

Jerry-Xin (changes requested):
## Re-review after byte-verifying peer-raised blockers @ 24cd05897336

Retracting my earlier APPROVE. On independent verification of the actual committed files, one raised blocker is real: this migration resolves the `queryMembers` 1267 but **moves the collation mismatch into the group↔robot bot-membership joins**, creating a new, symmetric 1267 on the same fresh-MySQL-8 environment it targets.

### 🔴 Blocking

**1. Flipping `robot.robot_id` → `utf8mb4_general_ci` breaks `robot.robot_id = group_member.uid` joins (net regression).**

`group_member.uid` has no explicit collation and its table has no `CHARSET/COLLATE` clause, so on fresh MySQL 8 it inherits the server default `utf8mb4_0900_ai_ci`:

```
-- modules/group/sql/20191106000002_group_legacy01.sql
create table `group_member`
(
  ...
  uid        VARCHAR(40) not null default '',   -- 成员uid   (no COLLATE; table has no CHARSET/COLLATE)
  ...
);
```

I scanned every `modules/group/sql/*.sql` at this ref: `group_member.uid` is never `MODIFY`-ed and no collation is ever pinned. `robot.robot_id` is likewise unpinned pre-migration (`modules/robot/sql/20210926000001_robot_legacy01.sql`: `robot_id VARCHAR(40) not null default ''`, no charset clause).

Four production queries join these two columns directly:

```
-- modules/group/db.go  QueryBotsInvitedByUIDTx
"INNER JOIN robot r ON r.robot_id = gm.uid AND r.status = 1 ... FOR UPDATE"

-- modules/group/db.go  QueryBotUIDsOwnedByUIDs
"INNER JOIN robot r ON r.robot_id = gm.uid AND r.status = 1"

-- modules/group/db.go  QuerySecondOldestMemberExcludingBotsOf
"LEFT JOIN robot r ON r.robot_id = gm.uid AND r.status = 1 AND r.creator_uid = ?"

-- modules/robot/db.go  queryRobotUIDsInGroup
"INNER JOIN robot r ON gm.uid = r.robot_id WHERE gm.group_no = ? AND gm.is_deleted = 0 AND r.status = 1"
```

On fresh MySQL 8:
- **Before:** `r.robot_id (0900_ai_ci) = gm.uid (0900_ai_ci)` → OK.
- **After this migration:** `r.robot_id (general_ci) = gm.uid (0900_ai_ci)` → **error 1267** in all four queries.

`QueryBotsInvitedByUIDTx` / `QueryBotUIDsOwnedByUIDs` / `QuerySecondOldestMemberExcludingBotsOf` are on the D-2 cascade / blacklist / owner-leave paths (bot cascade removal, second-oldest owner selection). Breaking these on the exact environment this PR targets is a regression, not a fix — it relocates the mismatch instead of eliminating it, which directly undercuts the migration's own stated thesis ("fix at the schema level so the whole family stops drifting").

**Fix:** extend the alignment to the full `uid`-family join closure. At minimum add `group_member.uid` to the same `utf8mb4_general_ci` alignment, and add a repro that exercises a group↔robot join (e.g. `queryRobotUIDsInGroup` / `QueryBotUIDsOwnedByUIDs`) after the migration — not only `queryMembers`. Before freezing this migration, enumerate every column that joins the ones being flipped and align the whole set, or the next fresh-MySQL-8 join hits the same wall.

### 💬 Non-blocking

- **`MODIFY COLUMN` lock cost.** A collation change is a definition change, so InnoDB runs it as `ALGORITHM=COPY` (full table rebuild, blocks concurrent writes for the duration) — it is not an INSTANT/INPLACE-lock-free op. `user.uid` is already `general_ci` (its DDL is created under `... COLLATE utf8mb4_general_ci`), so *if* MySQL treats that MODIFY as a true no-op it is cheap, but several MySQL 8 minors still rebuild on a MODIFY even when the resulting definition is unchanged. Add a one-line note on expected lock behaviour / table sizes for `user` / `space_member` / `robot`, or gate the rollout behind an online-DDL tool. This is a doc/ops nit, not a code defect — the regression above is the real blocker.
- **Coordinate with #537** (`fix(space): member search name fallback + collation-safe verification JOIN #434`). #537 also touches `modules/space/db.go` collation handling for the verification JOIN; land order matters to avoid charset drift between the two.

### ✅ Correct in this PR

- **Root-cause diagnosis is accurate and byte-verified.** `user_verification` is the only one of the four tables that pins a collation (`... ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci`, `modules/user/sql/20260505000003_user_legacy01.sql`), while `space_member` / `user` / `robot` inherit the server default. The `uv.user_id = sm.uid` mismatch that aborts `queryMembers` with 1267 is exactly as described.
- The `queryMembers`-scoped fix itself (aligning `space_member.uid`, `user.uid`, `user_verification.user_id`, `robot.robot_id` to `general_ci`) does resolve the reported failure and is idempotent; the migration keeps type/nullability/default/comment identical so only collation changes.
- The repro test is exemplary: `alignUpStatements` reads the shipped SQL from `sqlFS` and executes it verbatim (rather than a hand-copied variant), asserts the RED 1267 before and GREEN success after, and re-checks each column's `COLLATION_NAME` — so the test proves the on-disk migration, not a paraphrase. Extending it to cover a group↔robot join would also catch the regression above.

yujiawei (changes requested):
# Code Review — PR #540 (octo-server)

**Verdict: CHANGES_REQUESTED.** The fix is correct for the one query it targets, but pinning the collation of only four columns (while sibling columns and adjacent tables keep the MySQL 8 charset default `utf8mb4_0900_ai_ci`) moves the `1267` mismatch onto several other hot join paths on fresh MySQL 8. It fixes `GetSpaceMembers` and newly breaks group-member sync, the bot group-members API, and space-admin creator lookups in the same environment the PR is meant to repair.

## 1. Spec compliance

**Spec: ✅** (for the stated goal)

- Missing work: none. The linked issue asks to align the `GetSpaceMembers` join keys at the schema level; the migration does exactly that for `space_member.uid`, `user.uid`, `robot.robot_id`, `user_verification.user_id`, and the regression test proves red→green for `queryMembers`.
- Extra work: none out of scope.
- Deviation: none from the stated scope.

The spec was scoped narrowly to the four `GetSpaceMembers` join keys, and the diff matches that scope. The problem below is not a spec violation — it is a correctness regression the narrow scope introduces. It is raised under Code Quality.

## 2. Code quality

### P1 — New `1267` on group-member and bot-API join paths (fresh MySQL 8)

The migration pins `user.uid` and `robot.robot_id` to `utf8mb4_general_ci`, but `group_member.uid` is created by charset-less legacy DDL and stays on the server default (`utf8mb4_0900_ai_ci`) on a fresh MySQL 8 install — the exact environment this PR targets.

- `group_member.uid` DDL, no explicit collation: `modules/group/sql/20191106000002_group_legacy01.sql:37`
- The repro test itself forces `user.uid` to `0900_ai_ci` to reproduce the "fresh MySQL 8" state (`modules/space/db_member_collation_repro_test.go:80-82`), which confirms these legacy columns are `0900` there. `group_member.uid` is created the same way, so it is `0900` too.

Before this PR both sides were `0900` and the joins worked. After this PR `user.uid` becomes `general_ci` while `group_member.uid` stays `0900`, so these comparisons now raise `1267`:

- `SyncMembers` (group member sync, high-frequency): `modules/group/db.go:459` — `LeftJoin("user", "group_member.uid=user.uid")`
- Manager group-member list: `modules/group/db_manager.go:72` — `LeftJoin("user", "user.uid=group_member.uid")`
- Bot API `GET /v1/bot/groups/:group_no/members`: `modules/bot_api/groups.go:141` — `LEFT JOIN user u ON gm.uid = u.uid`

The same bot-API query also joins the *unpinned* `robot.creator_uid` against the *pinned* `user.uid`:

- `modules/bot_api/groups.go:143` — `LEFT JOIN user u2 ON r.creator_uid = u2.uid`
- `robot.creator_uid` DDL, no explicit collation: `modules/robot/sql/20260226000002_robot_legacy01.sql:3`

### P1 — New `1267` on space-admin creator lookups (fresh MySQL 8)

`space.creator` is created by charset-less legacy DDL (`modules/space/sql/20260307000002_space_legacy01.sql:11`) and stays `0900`, while `user.uid` becomes `general_ci`. The admin space list/detail queries join them:

- `modules/space/db_manager.go:112` — `LEFT JOIN user u ON u.uid = s.creator`
- `modules/space/db_manager.go:165` — `LeftJoin(... "u.uid=s.creator")`

These were fine pre-PR (both `0900`) and 1267 post-PR on fresh MySQL 8.

### Why the existing repair migration does not save this

`modules/base/sql/20260512000001_base_oss_compat_repair.sql:58` already converts 17 tables to `general_ci`, but its table list deliberately excludes `user`, `space_member`, `robot`, `space`, and `group_member`. So nothing else re-aligns the tables this PR half-migrates. The durable pattern the codebase already uses is **whole-table** `CONVERT TO CHARACTER SET ... COLLATE utf8mb4_general_ci`, not per-column `MODIFY` of a single join key.

### Recommended fix

Align at table granularity for the uid-family tables so sibling columns and adjacent tables converge together, rather than pinning four isolated columns. Either:

- Add `user`, `space_member`, `robot`, `space`, `group_member` to the same converged set the compat-repair migration uses (`CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci`, idempotency-guarded), or
- Extend this migration to also pin `group_member.uid`, `space.creator`, and `robot.creator_uid` (and audit for any other uid-family join partner that stays on the default). The whole-table CONVERT is safer because it catches columns not enumerated here.

### P2 — Migration understates operational lock cost

All four `MODIFY` statements touch indexed columns (unique/primary/secondary): `space_member` (`...20260307000002_space_legacy01.sql:32`), `user.uid` unique index (`...20191106000003_user_legacy01.sql:41`), `robot.robot_id` unique (`...20210926000001_robot_legacy01.sql:17`), `user_verification.user_id` PK (`...20260505000003_user_legacy01.sql:18`). A collation change to an indexed column commonly forces `ALGORITHM=COPY` and a table rebuild that blocks writes. On a no-op run (already `general_ci`) MySQL skips it, but on the divergent install this migration is written for, it will rebuild. Consider `ALGORITHM=INPLACE, LOCK=NONE` (fail-safe if it would block) or an online-schema-change tool, and mark it off-hours. The migration comment ("`MODIFY ... is idempotent ... safe to re-run`") is true for the no-op path but hides this rebuild cost on the path that actually applies.

### P2 — Regression test leaves schema mutated (no cleanup)

`TestQueryMembersCollationMismatch1267` issues global `ALTER TABLE` via `mustExec` (`modules/space/db_member_collation_repro_test.go:79-82`) with no `t.Cleanup`. `setup(t)` only cleans row data (`testutil.CleanAllTables`), and `testCtx` is a package-shared DB built once in `TestMain` (`modules/space/api_test.go:40,77`). The test's end-state happens to be the aligned `general_ci` (so it is not obviously destructive today), but relying on that coincidence is fragile: any future test that runs after it and asserts the default collation will see mutated schema. Add a `t.Cleanup` that restores the pre-test collations.

### P2 — `MODIFY` hardcodes the full column definition

The `MODIFY` statements restate `VARCHAR(40) NOT NULL DEFAULT '' COMMENT '...'`. If any environment has legitimately widened these columns (e.g. `VARCHAR(64)`), this migration silently reverts the type along with the collation. A whole-table `CONVERT TO CHARACTER SET` changes only the collation and avoids this drift risk.

### P2 — Test parser is brittle (advisory)

`alignUpStatements` (`modules/space/db_member_collation_repro_test.go:35-63`) splits the migration on `;` after stripping `--` lines. It works for this file but breaks on semicolons inside string literals or block comments. Preferable to run the file through the same migration engine production uses. Non-blocking.

## 3. Overall verdict

**REQUEST_CHANGES.** Spec ✅, Quality Changes-Requested → overall blocked. The fix is directionally right (schema-level, not per-query `COLLATE`) but half-applied: aligning four columns while their join partners stay on the MySQL 8 default converts one `1267` into several new ones on fresh MySQL 8. Widen to whole-table alignment (matching the existing compat-repair pattern) covering the uid-family tables and their join partners, then extend the regression test to a query that crosses `group_member`/`space.creator`/`robot.creator_uid`.

## 4. Coverage note

Verified: the four migrated columns and their DDL; the `GetSpaceMembers` query; the cross-file join partners `group_member.uid`, `space.creator`, `robot.creator_uid` and their charset-less DDL; the compat-repair table list; the regression test lifecycle. Not run: the test suite (branch diff / static review only) and a live fresh-MySQL-8 apply. A pre-existing (not caused by this PR) mismatch worth noting separately: `group_member.uid` vs `octo_dim_member.uid` (general_ci) in opanalytics (`modules/opanalytics/db.go:493`, `modules/opanalytics/etl_db.go:275`) is already misaligned on fresh MySQL 8 independent of this change.

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD 24cd0589

## Lens Summary

| Lens | Verdict | Headline |
|---|---|---|
| code | `needs-discussion` | Correct 1267 fix, but a space-module migration ALTERs user/robot tables, diverging from the central-repair pattern; no lock warning. |
| security | `cleared` | Schema-only collation align + test; where/authz untouched, uids are ASCII hex, no secrets, no identity/uniqueness shift. |
| qa | `pass-with-risk` | Strong red→green repro, CI `Test` green on real MySQL; risks: `user`-table rebuild lock, column-scoped fix, brittle assertion. |

## Critical Findings (P0-P1)

无 Critical / High priority findings — the fix is functionally correct, the regression test genuinely reproduces the bug red→green, CI's `Test` job is green against a real MySQL, and security cleared. The lenses converged on Should-Fix design/ops items below rather than any correctness or security blocker.

## Should-Fix (P2)

- `[P2-1]` `modules/space/sql/20260707000001_align_member_join_collation.sql:25-28` — A **space**-module migration alters tables owned by the **user** module (`user`, `user_verification`) and **robot** module (`robot`), bypassing the documented central-repair convention. The repo already has `modules/base/sql/20260512000001_base_oss_compat_repair.sql`, a stored-proc that normalizes this exact 1267 collation-drift class for a curated ~17-table list — which **includes `user_verification`** (so it is re-aligned redundantly here) but **omits `space_member`, `user`, `robot`** (the genuine gap). — flagged by code/code-1 (+qa/qa-Q1, code/code-5) — Fix: prefer extending `__oss_compat_repair`'s list with `space_member`/`user`/`robot` (and pinning their source create-table DDLs) over a space-owned migration reaching cross-module; drop the redundant `user_verification` statement. Discuss with maintainer.
- `[P2-2]` `modules/space/sql/20260707000001_align_member_join_collation.sql:25-28` — The per-column collation change on a `varchar` join key forces a full-copy rebuild (algorithm=copy: metadata lock + table rebuild). The shared `user` table can be large, yet — unlike the sibling base-repair migration, which explicitly warns "operators with large tables should apply off-hours" — this migration carries no lock/off-hours guidance. Deploy-window write-stall risk. — flagged by code/code-2 **+** qa/qa-1 (independent cross-lens convergence) — Fix: add a header note mirroring the base-repair warning and confirm the `user`-table rollout plan (online-DDL / off-hours).
- `[P2-3]` `modules/space/sql/20260707000001_align_member_join_collation.sql:25-28` — Fix is **column-scoped**, not table-scoped (base repair uses a table-level charset-conversion statement). Other columns of `user`/`space_member`/`robot` remain on the drifted collation, leaving latent 1267 risk on any future join over a different column of these tables into a `general_ci` table. — flagged by qa/qa-2 — Fix: document the deliberate narrow scope, or align these tables fully via a table-level charset conversion.
- `[P2-4]` `modules/space/db_member_collation_repro_test.go:53` — `require.Len(t, stmts, 4)` plus the naive comment-strip + `;`-split parser couples the test to the migration's exact statement shape; a future 5th alter-statement, a `-- +migrate StatementBegin` block, or an inline `;` breaks it. — flagged by qa/qa-3 — Fix: assert `>= 1` / key on the presence of each expected collation-align statement line rather than an exact count.

## Suggestions (P3)

- `[P3-1]` `…align_member_join_collation.sql:22-27` — Header claims "comment kept identical / only the collation changes," but the 3 columns had only inline SQL `-- comments` originally; the migration's alter-statement adds real column comment clauses. Harmless, but the claim is inaccurate. (code/code-3)
- `[P3-2]` `…align_member_join_collation.sql:20` — the "no-op once aligned / safe to re-run" claim overstates: the result is idempotent but a manual re-run still triggers a table rebuild. Reword to "result-idempotent / non-erroring." (code/code-4 + qa/qa-6)
- `[P3-3]` source DDLs — Only the forward-repair half of the canonical approach is applied; source create-table DDLs for `space_member`/`user`/`robot` are not pinned to `general_ci`, so fresh installs still create the drifted `0900_ai_ci` collation then repair. (code/code-5)
- `[P3-4]` `db_member_collation_repro_test.go:80-83` — Shared-schema ALTERs have no `t.Cleanup`/restore; a mid-test failure (e.g. the RED assertion regressing) leaves columns on `0900_ai_ci` for the rest of the package run, risking cascade failures. Happy path is fine. (qa/qa-4) — Fix: add `t.Cleanup` to re-apply the aligned collation.
- `[P3-5]` test coverage — the robot-member leg (`robot.robot_id` join + `creator_uid` ownership gate) and the both-names-empty→placeholder path are not exercised here (placeholder is covered by `TestMemberDisplayNamePure`). (qa/qa-5) — Fix: optionally seed a robot row.

## Questions for Author

- `[Q-1]` Is the collation drift on `user`/`space_member`/`robot` actually reproduced by the base migrations on a genuinely-fresh MySQL 8, or do their source DDLs already resolve to a consistent collation? The base-repair migration deliberately omitted these 3 tables; the test **hand-forces** the drift and so cannot confirm the ambient fresh-install scenario. If their source DDL already aligns, this migration is a no-op everywhere and the reported prod failure originated from a different historical schema state worth confirming. (from qa/qa-Q1 + code/Q1)
- `[Q-2]` Was re-aligning `user_verification` intentional, given it is already covered by `__oss_compat_repair` (which runs earlier at `20260512…`)? (from code/Q2)
- `[Q-3]` CI note: two non-code checks (check-sprint and the advisory code-review bot) show red on this HEAD, while the code test, build, vet, lint and i18n jobs are all green. Please confirm check-sprint is governance-only and non-blocking. (from qa Q2)

## Verdict Rationale

`REQUEST_CHANGES` by deterministic rule: no P0/P1 (no security block, no QA fail, no code Must-Fix), but **≥3 converged P2 Should-Fix items** (no-P0 + ≥3 P2 → REQUEST_CHANGES). The strongest signal is cross-lens: both the code and QA lenses independently flagged the missing table-rebuild/metadata-lock warning on the large shared `user` table (P2-2). The code lens additionally holds a genuine design question — a space-owned migration mutating user/robot-module tables diverges from the repo's documented central `__oss_compat_repair` convention and redundantly re-touches an already-covered table (P2-1) — and QA notes the fix is column-scoped, leaving residual drift (P2-3). **This is "changes requested for design/ops hardening," not a correctness or security defect**: the SQL is correct, the regression test is rigorous (reads and applies the shipped file verbatim, red→green), CI's real-MySQL `Test` job passed, and security cleared with no findings. Security's contrarian check confirmed the collation move introduces no case/accent-sensitivity change (prior state was equally insensitive) and uids are ASCII hex, so no identity/uniqueness/authz shift.

## Round Context

|---|---|---|
| `modules/space/sql/20260707000001_align_member_join_collation.sql` | ✅ (code, security, qa) | P2-1, P2-2, P2-3, P3-1, P3-2, P3-3 |
| `modules/space/db_member_collation_repro_test.go` | ✅ (code, security, qa) | P2-4, P3-4, P3-5 |
| `modules/space/db.go` (queryMembers:146, context) | ✅ (code, security, qa) | authz where-clause unchanged — no finding |

## Notes

- Confidence filter aggregate: code `2 dropped, 1 downgraded`; security `0 dropped`; qa low-signal wording nit dropped. Cleared during contrarian passes: empty-`Down` tolerance (in-repo precedent exists), migration-ordering safety (`20260707000001` sorts after all create-table migrations + base repair), filename collision, and case-insensitivity/identity concerns.
- No security embargo: security lens = `cleared` (not `blocked`); no separate embargo comment.
- Redaction: GH copy passed through `redact-report.sh`; this multica copy is raw.
- Terminal chain: this comment is the last artifact for Round 1; a next round is triggered by the spawner only on a new HEAD SHA.

---

lml2468 (changes requested):
**Formal Verdict: REQUEST_CHANGES** (Round 1 · HEAD 24cd0589)

The 1267 fix is correct, well-tested (red→green repro applying the shipped migration verbatim), CI `Test` is green on real MySQL, and security cleared. Changes are requested for **design/ops hardening, not correctness**: (1) a space-module migration ALTERs user/robot-module tables, diverging from the central `__oss_compat_repair` convention and redundantly re-touching `user_verification`; (2) no metadata-lock/off-hours warning on the `ALGORITHM=COPY` rebuild of the shared `user` table (flagged by both code + qa); (3) the fix is column-scoped, leaving other columns of these tables on the drifted collation.

Full findings, questions, and file matrix in the aggregate comment on this PR.

PR #544, opened 2026-07-07
feat(sticker): configurable upload limits + opt-in static compression

an9xyz:
## Summary

Two coupled capabilities layered onto the existing `modules/file` sticker upload path, both default-OFF (zero-impact until an operator opts in):

1. **Configurable upload limits** — the previously hard-coded custom-sticker limits (`StickerMaxFileSize=1MB`, `StickerMaxDimension=512`, `stickerUploadExts`) move into `system_setting` with server-side hard caps. The read path clamps values so a bad admin edit can't exceed the resource envelope; `upload_allowed_formats` may only *narrow* the built-in raster allowlist (mp4/pdf/svg silently dropped in the getter).
2. **Opt-in server-side compression (plan C)** — when `sticker.compress_enabled=true`, static `jpg/png` uploads are decoded → optionally downscaled → metadata-stripped → re-encoded (JPEG q=85 / PNG). `webp`/`gif`/animated (incl. APNG) are validate-only; if the compressed byte count still exceeds `compress_target_kb`, the upload is rejected outright. The `sticker_handle` is signed over the final stored URL.

### New `system_setting` keys

| key | default | hard cap |
|---|---|---|
| `sticker.upload_max_size_kb` | 1024 | 5120 |
| `sticker.upload_max_dimension` | 512 | 1024 |
| `sticker.upload_allowed_formats` | `gif,png,jpg,jpeg,webp` | intersect with built-in raster allowlist |
| `sticker.compress_enabled` | false | — |
| `sticker.compress_target_kb` | 1024 | 5120 |
| `sticker.compress_max_concurrency` | 4 | 32 |
| `sticker.compress_timeout_ms` | 2000 | 10000 |

### Stability isolation

- **Concurrency gate**: mutex+counter, capacity read from settings on every request so `compress_max_concurrency` changes take effect within a snapshot TTL. Saturation returns `skipped:concurrency_saturated` (fail-open to original bytes).
- **Timeout**: `time.Timer` preempts the worker goroutine. Slot lifetime is bound to the worker (NOT to `Compress()`'s return) so a timeout return can't leak a slot and let true in-flight work exceed the configured cap.
- **Panic**: worker `recover` maps any decoder crash to `failed`, upload keeps going with original bytes.
- **APNG**: `acTL` chunk scan before slot acquisition. Without this check, standard `image/png` would silently decode only the first frame and re-encode a static PNG, dropping the animation.
- **All error paths fail-open** to the pre-compression byte stream so compression can never destabilize the upload path.

### Observability

- Counter `sticker_upload_total{result}` — 4 new dimensions pre-warmed to 0: `compress_success` / `compress_failed` / `compress_skipped` / `compress_over_limit`
- Histogram `sticker_compress_duration_seconds{outcome,format}` — 10 buckets from 5ms..5s; only points on paths that actually spent CPU (compressed / failed / over_limit / skipped:timeout); format normalized to a 6-value whitelist to bound cardinality
- Zap logs include `orig_size` + `final_size` on the upload-success path so post-hoc analysis can compute shrink ratios

## Out of scope (deferred)

- `webp` / `gif` / animated re-encoding — would require a cgo `libwebp` dep. Phase one is validate-only for these.
- Changing the two-stage upload → register contract, quota (`user_max_count`), or the `handle_required` rollout.
- Client-side changes — upload response field shape is preserved (`path`/`name`/`size`/`ext`/`sha512`/`sticker_handle`).
- Migrating `modules/file` to the i18n error envelope.
- Async / out-of-band compression pipeline.

## Test plan

Two commits: initial implementation (`0177781e`) + review-driven fix (`b3c86473`).

- [x] `modules/common` clamp / intersection / narrowing behavior — **19 no-infra unit tests**, cover default fallback, DB override, hard-cap clamp, non-positive/non-numeric fallback, boundary-value accept
- [x] `modules/file` compressor — **9 unit tests** cover disabled/format skip, JPEG shrink, PNG re-encode, over-limit reject, decode-error fail-open, concurrency saturation, timeout fail-open, aspect-ratio preserved (Fit vs Resize/Fill), no upscale for under-maxDim source
- [x] `modules/file` uploadFile integration — **4 tests** using an in-memory service that captures the bytes sent to storage: large JPEG uses compressed bytes (`resp.size == len(uploaded)`), GIF passes through byte-for-byte, over-limit returns non-OK with nothing stored, `sticker_handle` verifies against the final stored URL and rejects a different uid
- [x] APNG-specific — 4 chunk-scan unit tests (valid APNG detection / static PNG rejection / spec-invalid trailing acTL ignored / malformed input fuzz)
- [x] Concurrency-slot regression — `TestStickerCompressor_TimeoutDoesNotBleedConcurrency` uses maxConcurrency=1 + slow worker: the request following a timeout return must get `concurrency_saturated` (proving the slot did not leak to a new caller)
- [x] Benchmarks — 6 permutations across (JPEG/PNG) × (512²/1024², shrink / re-encode-only). Local Apple M5 baseline is included in the perf-todo file
- [x] Full `go test -race -count=1 ./modules/file/ ./modules/common/` after a `DROP DATABASE test; CREATE DATABASE test;` — passes end-to-end
- [x] `go vet ./...` — clean
- [x] `make i18n-lint` — clean (no new i18n codes introduced)

### Pre-rollout MUST list

Compression defaults to OFF. Production enablement gated on the perf-todo checklist tracked at `.octospec/tasks/sticker-upload-compression/perf-todo.md`:

- Re-run benches on production-representative CPU (local baseline is Apple M5; expect ~3× slower on x86 cloud VMs)
- `hey` / `wrk` end-to-end HTTP compare of latency P50/P95/P99 + CPU / heap / goroutine steady-state
- Goroutine leak-bound test: sustained timeout-triggering upload should show worker count capped by `compress_max_concurrency`
- pprof to confirm CPU time is in `imaging.Encode` / Lanczos, no framework hotspot
- Decompression-bomb sanity: 1KB pathological input decoding to a 10000² bitmap must still be blocked by the pre-existing dimension gate before reaching the compression path
- Grey-out ramp: 5% traffic → 15 min observation → 50% → 30 min → 100%; rollback = flip `sticker.compress_enabled=false` (60s replica convergence)

### DIFF
--- a/.octospec/tasks/sticker-upload-compression/brief.md
+++ b/.octospec/tasks/sticker-upload-compression/brief.md
@@ -1,0 +1,161 @@
+---
+type: Task
+title: "Task: sticker-upload-compression"
+description: Make custom-sticker upload limits (formats / max file size / max dimension) operator-configurable via system_setting with server-side hard caps, and add opt-in server-side compression for static jpg/png (decode → optional downscale → strip metadata → re-encode); webp/gif and animated images are validated-only in phase one.
+tags: ["sticker", "wire-contract", "rate-limit", "fullstack"]
+timestamp: 2026-07-07T06:54:12Z
+# --- octospec extension fields ---
+slug: sticker-upload-compression
+upstream: TBD
+source: self
+---
+
+# Task: sticker-upload-compression
+
+> Follow-up on `custom-sticker-management` / `sticker-upload-handle`. Two coupled
+> capabilities layered onto the same `modules/file` sticker upload path:
+> (1) the previously hard-coded upload limits become operator-tunable with
+> server-side hard caps; (2) an opt-in, phase-one server-side compressor for
+> static raster stickers. Both default to today's behavior (zero-impact until an
+> operator opts in).
+
+## Goal
+
+1. **Configurable upload limits.** The custom-sticker upload constraints —
+   allowed formats, per-file size cap, max decoded dimension — currently
+   hard-coded in `modules/file/const.go` (`stickerUploadExts`,
+   `StickerMaxFileSize` = 1 MB, `StickerMaxDimension` = 512) become
+   `system_setting` keys read through the existing 60s-snapshot `SystemSettings`
+   singleton, so they can be greyed out / rolled back without a redeploy.
+   Defaults reproduce today exactly. Every value is clamped by a **server-side
+   hard cap** so a mis-config cannot open a resource-exhaustion hole:
+   - max file size: default **1024 KB**, hard cap **5120 KB** (5 MB)
+   - max dimension: default **512 px**, hard cap **1024 px**
+   - allowed formats: default `gif,png,jpg,jpeg,webp`, intersected with the
+     built-in raster allowlist (config may only **narrow**, never add non-raster
+     types).
+
+2. **Server-side compression (opt-in, phase one = plan C).** When enabled, a
+   **static** `jpg/jpeg/png` upload is decoded → optionally downscaled to the
+   configured target → stripped of metadata → re-encoded in its original format,
+   before storage. `webp` and `gif` (and any animated image) are **validated
+   only, never compressed** in phase one — because the current dependency set
+   (`golang.org/x/image` webp = decoder-only; `disintegration/imaging` = no webp
+   encoder) cannot re-encode webp without a new cgo dependency, which is
+   deliberately out of scope. If a static image still exceeds the compression
+   target after re-encode, the upload is **rejected** (not stored oversized).
+
+3. **The `sticker_handle` is signed over the FINAL stored path/bytes.** Because
+   compression may change the object (and, if a future phase changes the
+   extension, the path), `stickersig.Sign` and the response `path`/`ext`/`size`
+   MUST reflect the post-compression result. In phase one compression preserves
+   the extension (jpg→jpg, png→png), so the path is unchanged; the invariant is
+   still asserted so a later format-changing phase cannot silently break it.
+
+4. **Ops & observability.** A default-OFF `compress_enabled` grey-out switch;
+   compression bounded by configurable **timeout, concurrency, and memory**
+   (decoded-pixel) ceilings, fail-open (skip compression, fall back to
+   validate-then-store) so the compressor can never destabilize the main upload
+   path; new low-cardinality metrics for compress success/failure/skip/over-limit
+   alongside the existing upload-outcome counters.
+
+## Background
+
+`modules/file/api.go:uploadFile()` is the single choke point holding both the
+authenticated uploader and the content-validated bytes; all sticker file-level
+gating (size → ext allowlist → raster-only → magic number → path-ext match →
+`image.DecodeConfig` dimension cap → store → `stickersig.Sign`) lives there, so
+compression must slot into that same path — after the dimension cap (api.go
+~397-428), before `service.UploadFile` (~463) and `stickersig.Sign` (~503).
+
+Config plumbing reuses `modules/common/system_setting_schema.go` (canonical
+schema slice; admin write path validates against it) + the `SystemSettings`
+atomic-snapshot getters (60s auto-reload, already home to
+`sticker.user_max_count / custom_enabled / handle_required`). There is **no
+`featuregate` module in this repo**; grey-out is a `system_setting` bool. The
+existing shared int bound is `[0, 3650]` (`settingIntMin/Max`); MB- and px-scale
+caps need their own per-key clamp getters rather than that shared bound.
+
+Metrics reuse `pkg/metrics/sticker.go` + `modules/file/sticker_metrics.go`
+(`observeStickerUpload(result)` CounterVec, dimensions pre-warmed to 0).
+
+Plan C was chosen (over transcoding webp→png/jpeg, or adding a cgo webp encoder)
+to keep phase one pure-Go, zero new dependency, and low operational risk.
+
+## Load-bearing list
+
+- **`modules/file` sticker upload contract** (touches: `wire-contract`) — the
+  upload response keeps its field shape (`path`/`name`/`size`/`ext`/`sha512`/
+  `sticker_handle`); when compression runs, `size` and the signed bytes are the
+  post-compression values. No change to any non-`type=sticker` upload.
+- **`stickersig.Sign` provenance binding** (touches: `auth`, `acl`) — the handle
+  must be minted over the final stored path (unchanged in phase one); the
+  register-side `validateStickerPath` `ext==format` invariant must still hold.
+  Reuses `sticker-upload-handle` guarantees; must not regress the
+  `handle_required` fail-closed path or the `sticker/` keyspace reservation.
+- **`system_setting` schema + admin write path** (touches: `wire-contract`) —
+  new keys added to the canonical schema slice; the manager write path
+  (`api_manager_system_setting.go`) and clamp getters enforce the per-key hard
+  caps. `custom_enabled` / appconfig client contract untouched.
+- **Upload-path resource envelope** (touches: `rate-limit`, `throttle`) —
+  compression adds CPU/memory load inside the request; the concurrency semaphore
+  + timeout + decoded-pixel ceiling are the throttle. Fail-open on saturation:
+  no new hard failure mode vs. today. This is process-local compute throttling,
+  NOT request-frequency limiting — the existing `SharedUIDRateLimiter` on the
+  sticker routes and the file-module IP limiter are unchanged.
+- **`modules/file` error-response convention** (touches: `error-response`) — the
+  file module is NOT yet migrated to the i18n `httperr` envelope; new rejection
+  paths use `c.ResponseError(...)` to stay consistent within the module (do not
+  introduce a lone `httperr` call here).
+- **Image decode surface** (touches: `trust-boundary`, `external-content`) —
+  decode/re-encode operates on untrusted user bytes; decompression-bomb defense
+  (decoded-pixel ceiling) and animated-image detection (gif via `DecodeAll`
+  frame count, animated-webp via RIFF/VP8X `ANIM` chunk scan) are new pure-Go
+  code that must fail safe.
+
+## Out of scope
+
+- **WebP / GIF / animated-image compression or transcoding.** Phase one is
+  validate-only for these; no webp re-encoder, no cgo `libwebp`, no format
+  conversion. (Explicitly deferred to a later phase.)
+- **Changing the two-stage upload→register contract**, the `sticker` module
+  register logic, quota (`user_max_count`), or the `handle_required` rollout.
+- **Client / octo-web changes.** The upload API field shape is preserved; no new
+  request field is required from clients.
+- **Request-frequency rate limiting.** No change to `SharedUIDRateLimiter` or the
+  file-module IP limiter; this task only adds compute-side throttling.
+- **Migrating `modules/file` to the i18n error envelope.**
+- **Async / out-of-band compression pipeline.** Phase one is synchronous within
+  the upload request, bounded by timeout + concurrency.
+
+## Acceptance
+
+- `sticker.upload_max_size_kb`, `sticker.upload_max_dimension`,
+  `sticker.upload_allowed_formats`, `sticker.compress_enabled`,
+  `sticker.compress_target_kb`, `sticker.compress_max_concurrency`,
+  `sticker.compress_timeout_ms` exist in the `system_setting_schema` slice, are
+  accepted by the manager write path, and are read via `SystemSettings` getters.
+- Clamp getters unit-tested: a configured value above the hard cap
+  (size > 5120 KB, dimension > 1024 px) returns the hard cap (not the input) and
+  logs a warning; `allowed_formats` is intersected with the raster allowlist
+  (a config value of `gif,png,mp4` yields `gif,png` — never `mp4`).
+- With `compress_enabled=false` (default), upload behavior is byte-for-byte
+  identical to current `main` (regression test: same stored bytes, same
+  `size`/`ext`/`sticker_handle`).
+- With `compress_enabled=true`: a static jpg/png larger than the target is
+  downscaled + re-encoded + metadata-stripped and stored **at or under** the
+  target; a static jpg/png that still exceeds the target after re-encode is
+  rejected with `observeStickerUpload("compress_over_limit")`.
+- A `webp` or animated `gif`/`webp` upload is never re-encoded (metric
+  `compress_skipped`), and is rejected iff it violates the (configurable) size /
+  dimension / format limits — same as today.
+- `sticker_handle` verifies against the **post-compression** stored bytes/path;
+  register (`POST /v1/sticker/user`) with the returned `path`/`handle` succeeds,
+  and `validateStickerPath`'s `ext==format` invariant holds.
+- Compression is bounded: unit test asserts it aborts on the configured timeout,
+  and that concurrency beyond `compress_max_concurrency` fails open (upload still
+  succeeds via validate-then-store, metric `compress_skipped`).
+- New metric dimensions `compress_success / compress_failed / compress_skipped /
+  compress_over_limit` are registered and pre-warmed to 0.
+- `go test ./modules/file/... ./modules/common/...` and `go vet` pass;
+  `make i18n-lint` unaffected (no new i18n codes introduced).
--- a/.octospec/tasks/sticker-upload-compression/perf-todo.md
+++ b/.octospec/tasks/sticker-upload-compression/perf-todo.md
@@ -1,0 +1,139 @@
+# perf-todo — sticker-upload-compression
+
+灰度前必须完成的性能验证清单。**本任务当前状态：功能就绪、性能未验证**。
+在 `sticker.compress_enabled=true` 打开到任何生产流量之前，必须回答本清单
+的每一项，把结果贴在本文件下方对应小节，作为放量决策依据。
+
+## 0. 已有的观测能力
+
+- Counter `sticker_upload_total{result}` — 新增 4 个 compress_* 维度已预热为 0
+- Histogram `sticker_compress_duration_seconds{outcome,format}` — 仅在
+  acquire slot 之后打点（compressed / over_limit / failed / skipped:timeout），
+  disabled/format/concurrency_saturated 分支耗时约等于 0，用 counter 观测即可
+
+Grafana 上线灰度前必须准备的面板：
+- P50/P95/P99 `sticker_compress_duration_seconds` 按 outcome、format 分维度
+- `rate(sticker_upload_total{result=~"compress_.*"}[5m])` 分四维度堆叠图
+- `histogram_quantile(0.99, ...)` 与配置 `compress_timeout_ms` 的对比线
+- 进程 CPU / goroutine 数 / heap alloc 曲线（对齐灰度切换时间点）
+
+## 1. 单机 CPU 基线（本地 bench，DONE — Apple M5）
+
+```
+go test -run x -bench='BenchmarkDoCompressStaticSticker' -benchmem -benchtime=3s ./modules/file/
+```
+
+| Benchmark                                     | ns/op       | ~ms/op | alloc/op |
+| --------------------------------------------- | ----------- | ------ | -------- |
+| JPEG 512² re-encode                           | 12,046,858  | ~12    | 936 KB   |
+| PNG 512²  re-encode                           | 29,883,139  | ~30    | 4.5 MB   |
+| JPEG 1024² → Fit(512) + re-encode             | 34,451,231  | ~34    | 5.6 MB   |
+| PNG  1024² → Fit(512) + re-encode             | 51,525,786  | ~52    | 11.2 MB  |
+| JPEG 1024² re-encode only (maxDim=1024)       | 49,102,169  | ~49    | 3.7 MB   |
+| PNG  1024² re-encode only (maxDim=1024)       | 111,301,979 | ~111   | 15.6 MB  |
+
+**关键推论**（M5，生产 x86 云 VM 通常慢 2-3×，用 3× 保守估算）：
+
+- **最坏单帧耗时**：PNG 1024² re-encode ≈ **~330 ms**（生产估）
+- **最坏单帧内存 alloc**：~15 MB / 帧 → concurrency=4 时峰值 **~60 MB**（可控）
+- **QPS 上限**（生产估）：
+  - concurrency=4：JPEG 512² ~110 QPS；PNG 1024² ~12 QPS
+  - concurrency=8：JPEG 512² ~220 QPS；PNG 1024² ~24 QPS
+
+**默认值合理性初判**：
+- `compress_max_concurrency=4` — 单实例上限保守，防止 CPU 打满。真实
+  贴纸 <1024² + JPEG 占多数，实测 QPS 天花板应远高于业务贴纸创建速率。
+- `compress_timeout_ms=2000` — 相对 ~330 ms 最坏单帧有 **6× 余量**，能吸收
+  抖动。若生产观测到 P99 < 500 ms 稳定，可下调到 1000 ms 提升拒绝速度。
+
+## 2. 待补的性能验证（灰度前 MUST）
+
+### 2.1 生产 CPU 型号下的 bench 复跑
+
+在实际生产 Node（x86 云 VM，非 M5）上跑上面这条 `go test -bench=...` 拿到
+本地基线的 3× / 4× 因子的实测校正值。**若结果偏离本地估算 ±50%，需重新
+评估 concurrency / timeout 默认值**。
+
+- [ ] 生产型号 CPU 上跑一遍 bench，把六行结果贴到本文件
+- [ ] 若 PNG 1024² 单帧 > 1000 ms，把 `compress_timeout_ms` 默认调高到
+      3000 ms 再放量，否则频繁 skip:timeout 会让压缩形同虚设
+
+### 2.2 pprof CPU / heap 剖面
+
+- [ ] `go test -run x -bench='BenchmarkDoCompressStaticSticker_PNG_1024' -cpuprofile=cpu.out`
+  → `go tool pprof -top cpu.out`，确认 CPU 时间主要在 imaging.Encode / Lanczos，
+  没有奇怪的框架开销
+- [ ] `-memprofile=mem.out`：确认 alloc/op 与 bench 报的一致；无异常泄漏
+
+### 2.3 端到端 HTTP 压测
+
+用 `hey` 或 `wrk` 打 `POST /v1/file/upload?type=sticker`（真实 auth token +
+multipart 载体），采集：
+
+- [ ] 稳态 QPS = 目标生产 QPS × 1.5 时的 P50/P95/P99 延迟
+- [ ] 稳态 CPU 使用率（应 < 70%）
+- [ ] 稳态 heap alloc 曲线（应平稳、无线性上涨 = 无泄漏）
+- [ ] 稳态 goroutine 数（应有界、无线性上涨 = 无 dangling worker）
+- [ ] 错误分布：`compress_over_limit` / `compress_skipped:timeout` /
+      `compress_failed` 各占比
+
+**过关标准（建议）**：
+- P99 < timeout 值 × 0.8（有余量，避免 timeout 密集触发）
+- `compress_skipped:concurrency_saturated` < 1%（并发数够用）
+- `compress_skipped:timeout` < 0.1%（timeout 足够）
+- CPU steady < 70%
+
+### 2.4 Goroutine 泄漏边界测试
+
+**风险点**：`stickerCompressor.Compress` 走 timeout 分支时，后台 goroutine 未
+被取消，仍在跑到 `doCompress` 自然结束（imaging 无 context）才退出。在
+"客户端持续上传导致每次都 timeout"的极端场景下，dangling goroutine 会累积。
+
+单帧 dangling 时间 = doCompress 完整耗时 - timeoutMs（PNG 1024² 最坏 ~110 ms
+超出 2000 ms 的场景需要更大的图，不常见）。**但需要显式压测**：
+
+- [ ] 用故意注入的 slow `doCompress`（sleep 5s），以 timeout=100 ms 打 1000 QPS
+      持续 60s，观察 goroutine 峰值 —— 应 ≤ `concurrency × ceil(5s/100ms) = 200`，
+      不应无限增长（compressor sem 已卡上限）
+- [ ] 若观察到线性增长，说明 sem 语义有 bug 或 goroutine 未按预期退出
+
+### 2.5 Mutex 争用评估
+
+**风险点**：`stickerCompressor.mu` 是全局单锁，`tryAcquireCompressSlot` /
+`releaseCompressSlot` 每次请求进出各锁一次。高 QPS 下可能成为热点。
+
+- [ ] `go test -run x -bench='...' -blockprofile=block.out` → `go tool pprof
+      -top block.out`，看 mu 是否进入 top-3
+- [ ] 若是热点，考虑改 `atomic.Int32` + CAS（现在的 mutex 语义"数据比较"完全
+      可用原子操作代替，无 map/slice 保护需求）
+
+**当前判断**：默认 concurrency=4 时，稳态最多 4 个持锁 goroutine + 后来者一次
+tryAcquire 就 return，锁窗口 ~ns 级，不会是热点。但配合 §2.3 压测数据佐证。
+
+### 2.6 Decompression bomb 防御
+
+维度上限 1024²（配置硬 cap）→ 解码后 RGBA 内存 = 1024×1024×4 = **4 MB / 帧**。
+concurrency=4 时峰值 16 MB，安全。
+
+- [ ] 灰度前跑一次已知恶意样本（1 KB 压缩包解出 10000² 位图），确认在
+      `image.DecodeConfig` 阶段被 dimension 校验拦住，**不进入压缩管线**
+      （压缩管线的 image.Decode 走 full-decode 才是攻击面，前置维度门是
+      唯一防线）
+
+## 3. 灰度切换的操作剧本
+
+- [ ] Step 1: 打开 `sticker.compress_enabled=true` 到 **单实例 5% 流量**
+      （或用 GrayScale 平台把其中 1 台 pod 单独打标签）
+- [ ] Step 2: 观察 15 分钟：`compress_success` 占比 > 50%（说明格式命中）、
+      P99 < 500 ms、CPU 增幅 < 10%
+- [ ] Step 3: 逐步放量到 50% → 100%，每步观察 30 分钟
+- [ ] Step 4: 生产观测稳定 24h 后再调 `compress_max_concurrency` / `_timeout_ms`
+      默认值到实测最优（下调 timeout 提升拒绝速度、按 CPU 富余上调 concurrency）
+
+## 4. 回滚剧本
+
+- 一键关：`sticker.compress_enabled=false`（60s 内 all replica 收敛）
+- 缩容闸：`sticker.compress_max_concurrency=1` 快速降低 CPU 占用（不需要
+  完全关，只是缩容压力）
+- 兜底：所有压缩失败路径都 fail-open 走原字节，即使 compressor 完全瘫痪
+  上传主链路不受影响（brief 的 acceptance 已保证）
--- a/modules/common/system_setting_schema.go
+++ b/modules/common/system_setting_schema.go
@@ -1,6 +1,9 @@
 package common
 
-import "strconv"
+import (
+	"strconv"
+	"strings"
+)
 
 // Value types accepted by system_setting.value_type.
 const (
@@ -152,6 +155,34 @@ var systemSettingSchema = []settingDef{
 	// 不承担任何服务端鉴权。经 GET /v1/common/appconfig 的 docs_on 下发给客户端。
 	{Category: "docs", Key: "enabled", Type: settingTypeBool, Description: "是否向客户端展示文档(docs)模块入口（octo-docs-backend 上线前默认关闭）",
 		Effective: func(s *SystemSettings) string { return boolToCanonical(s.DocsEnabled()) }},
+
+	// 自定义贴纸上传限制（sticker-upload-compression 任务）。原先硬编码在
+	// modules/file/const.go；挪进 system_setting 后可灰度/回滚，且每键都有
+	// 服务端硬上限（stickerUpload*HardCap / stickerCompress*HardCap），误配也不会
+	// 越出资源上限。全部 Positive:true 走"必须正整数"admin 写侧校验，同时放开
+	// settingTypeInt 默认的 [0,3650] 上界（本组键上限单独校验，见读侧 clamp）。
+	{Category: "sticker", Key: "upload_max_size_kb", Type: settingTypeInt, Description: "自定义贴纸单文件大小上限(KB)，服务端硬上限 5120(5MB)；默认 1024", Positive: true,
+		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.StickerUploadMaxSizeKB()) }},
+	{Category: "sticker", Key: "upload_max_dimension", Type: settingTypeInt, Description: "自定义贴纸解码后单边像素上限，服务端硬上限 1024；默认 512", Positive: true,
+		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.StickerUploadMaxDimension()) }},
+	// upload_allowed_formats 用字符串存 CSV，读侧与内置位图白名单求交集(只能收窄,
+	// 不能放开非位图)。取消 Positive/整数校验，写侧不校验内容(anything goes)，非法
+	// 项在读侧被丢弃；全部非法时读侧回退默认全 5 种，避免误配把功能暗关。
+	{Category: "sticker", Key: "upload_allowed_formats", Type: settingTypeString, Description: "自定义贴纸允许扩展名(逗号分隔，如 gif,png,jpg,jpeg,webp)；只能与内置位图白名单取交集(收窄)，非位图会被读侧丢弃",
+		Effective: func(s *SystemSettings) string { return strings.Join(s.StickerUploadAllowedFormats(), ",") }},
+
+	// 自定义贴纸服务端压缩开关与调参（sticker-upload-compression 任务，方案 C：只压
+	// 静态 jpg/png；webp/gif/动图 validate-only）。compress_enabled 默认关闭，运营
+	// 灰度开启；compress_target_kb 是压缩目标(压缩后仍超即拒)；max_concurrency 与
+	// timeout_ms 是稳定性闸(饱和/超时都 fail-open，走原路径不阻塞主链路)。
+	{Category: "sticker", Key: "compress_enabled", Type: settingTypeBool, Description: "是否开启贴纸服务端压缩(仅静态 jpg/png；webp/gif 恒不压)；默认关闭以支持灰度",
+		Effective: func(s *SystemSettings) string { return boolToCanonical(s.StickerCompressEnabled()) }},
+	{Category: "sticker", Key: "compress_target_kb", Type: settingTypeInt, Description: "压缩目标(KB)；压缩后仍超此值将拒绝上传；硬上限 5120(5MB)，默认 1024", Positive: true,
+		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.StickerCompressTargetKB()) }},
+	{Category: "sticker", Key: "compress_max_concurrency", Type: settingTypeInt, Description: "同时进行的贴纸压缩数量上限；饱和时该请求跳过压缩走原路径(fail-open)；硬上限 32，默认 4", Positive: true,
+		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.StickerCompressMaxConcurrency()) }},
+	{Category: "sticker", Key: "compress_timeout_ms", Type: settingTypeInt, Description: "单次贴纸压缩超时(毫秒)；超时该请求跳过压缩走原路径(fail-open)；硬上限 10000，默认 2000", Positive: true,
+		Effective: func(s *SystemSettings) string { return strconv.Itoa(s.StickerCompressTimeoutMs()) }},
 
 	// Email server config — formerly yaml-only (Support.* in config.go).
 	{Category: "support", Key: "email", Type: settingTypeString, Description: "技术支持邮箱（发件人）",
--- a/modules/common/system_settings.go
+++ b/modules/common/system_settings.go
@@ -786,3 +786,170 @@ func (s *SystemSettings) StickerHandleRequired() bool {
 func (s *SystemSettings) DocsEnabled() bool {
 	return s.getBool("docs", "enabled", false)
 }
+
+// ---------------------------------------------------------------------------
+// Custom-sticker upload constraints + optional server-side compression
+// (sticker-upload-compression task).
+//
+// These formerly-hard-coded numbers (modules/file/const.go: StickerMaxFileSize
+// = 1MB, StickerMaxDimension = 512, stickerUploadExts) become operator-tunable
+// through system_setting so a bad configuration can be greyed out / rolled back
+// without a redeploy. Every int key has a server-side HARD CAP that read-side
+// clamp getters enforce even against a direct DB edit — the admin write path
+// already rejects non-positive ints via Positive:true; these clamps are defence
+// in depth against the "someone edits the row by hand" case.
+//
+// stickerUploadRasterAllowlist mirrors modules/file/const.go:stickerUploadExts
+// verbatim. Duplicated intentionally to keep modules/common a leaf (modules/file
+// already imports modules/common; reversing would cycle). Keep in sync — the
+// upload_allowed_formats getter uses this list as the outer bound the config
+// may only narrow from.
+// ---------------------------------------------------------------------------
+
+const (
+	defaultStickerUploadMaxSizeKB   = 1024
+	stickerUploadMaxSizeKBHardCap   = 5 * 1024
+
+	defaultStickerUploadMaxDimension = 512
+	stickerUploadMaxDimensionHardCap = 1024
+
+	defaultStickerCompressEnabled = false
+
+	defaultStickerCompressTargetKB = 1024
+	stickerCompressTargetKBHardCap = 5 * 1024
+
+	defaultStickerCompressMaxConcurrency = 4
+	stickerCompressMaxConcurrencyHardCap = 32
+
+	defaultStickerCompressTimeoutMs = 2000
+	stickerCompressTimeoutMsHardCap = 10000
+)
+
+// stickerUploadRasterAllowlist 与 modules/file/const.go:stickerUploadExts 保持一致。
+// 用于 upload_allowed_formats 配置的读侧交集：管理台只能收窄，不能加入非位图。
+// 若 modules/file 侧改动允许扩展名，此列表也需同步。
+var stickerUploadRasterAllowlist = []string{".gif", ".png", ".jpg", ".jpeg", ".webp"}
+
+// stickerClampIntUpper clamps an int getter to [1, hardCap]. Any value ≤0 or
+// non-numeric (which surface as fallback default from getInt) is served as
+// default; values above hardCap are clamped to hardCap; everything else is
+// returned verbatim. Shared by every KB/px/ms/count sticker upload setting so
+// the clamp policy is single-sourced.
+func stickerClampIntUpper(v, fallback, hardCap int) int {
+	if v <= 0 {
+		return fallback
+	}
+	if v > hardCap {
+		return hardCap
+	}
+	return v
+}
+
+// StickerUploadMaxSizeKB returns the per-file upload cap in KB. Read-side
+// clamped to [1, stickerUploadMaxSizeKBHardCap]; out-of-range falls back to
+// the historical 1024 KB default.
+func (s *SystemSettings) StickerUploadMaxSizeKB() int {
+	return stickerClampIntUpper(
+		s.getInt("sticker", "upload_max_size_kb", defaultStickerUploadMaxSizeKB),
+		defaultStickerUploadMaxSizeKB,
+		stickerUploadMaxSizeKBHardCap,
+	)
+}
+
+// StickerUploadMaxDimension returns the decoded-pixel single-edge cap. Read-side
+// clamped to [1, stickerUploadMaxDimensionHardCap]; out-of-range falls back to
+// the historical 512-px default.
+func (s *SystemSettings) StickerUploadMaxDimension() int {
+	return stickerClampIntUpper(
+		s.getInt("sticker", "upload_max_dimension", defaultStickerUploadMaxDimension),
+		defaultStickerUploadMaxDimension,
+		stickerUploadMaxDimensionHardCap,
+	)
+}
+
+// StickerUploadAllowedFormats returns the sanitized set of allowed extensions
+// (each including the leading dot, lowercased). It is intersected with the
+// built-in raster allowlist (stickerUploadRasterAllowlist) so a mis-config can
+// only narrow — never widen to non-raster (mp4/pdf/svg/...). If the config
+// exists but the intersection is empty (all tokens illegal), the FULL default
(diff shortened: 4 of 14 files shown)


Jerry-Xin (dismissed):
Project relevance gate passed: this PR modifies octo-server’s existing sticker upload and system settings functionality. No blocking issues found.

💬 Non-blocking

- 🟡 Warning: `GET /v1/file/upload` still generates sticker upload paths using the old hard-coded sticker extension set, not `sticker.upload_allowed_formats`. If operators narrow allowed formats, the preflight URL can still point clients at an extension that the upload endpoint later rejects. Upload enforcement is correct, but the client flow can become confusing. See [api.go](/Users/mlamp/.openclaw/tmp/pr-review-qnvkjqsd/repo/modules/file/api.go:180) and [const.go](/Users/mlamp/.openclaw/tmp/pr-review-qnvkjqsd/repo/modules/file/const.go:161).

- 🔵 Suggestion: `File.compressor` is documented as nil meaning disabled, but `uploadFile` will panic if `settings` is non-nil, compression is enabled, and `compressor` is nil. Production construction via `New` is safe, so this is not a blocker; a small guard would make the invariant match the comment. See [api.go](/Users/mlamp/.openclaw/tmp/pr-review-qnvkjqsd/repo/modules/file/api.go:58) and [api.go](/Users/mlamp/.openclaw/tmp/pr-review-qnvkjqsd/repo/modules/file/api.go:482).

- 🔵 Suggestion: `normalizeStickerCompressFormat` claims case-insensitive handling but only accepts all-lower or all-upper forms. Current callers pass lower-case extensions, so behavior is fine today; `strings.ToLower` would make the helper match its contract. See [sticker_metrics.go](/Users/mlamp/.openclaw/tmp/pr-review-qnvkjqsd/repo/modules/file/sticker_metrics.go:109).

✅ Highlights

- The upload path preserves the key security checks before compression: path ownership, sticker keyspace isolation, magic-number validation, path/content extension consistency, and decoded-dimension gating.
- The compressor’s timeout/concurrency handling correctly keeps the slot tied to the worker lifetime, preventing timeout returns from exceeding the configured in-flight cap.
- APNG detection before compression is a good fix for avoiding silent animation loss.
- Test coverage is strong across settings clamps, compressor outcomes, APNG scanning, timeout concurrency, and upload integration.

Verification:
- `go test ./modules/file` passed.
- `go test ./modules/common -run 'TestSystemSettings_Sticker(Upload|Compress)'` passed.
- `go test ./modules/file ./modules/common` could not complete because the local common test database has an unrelated unknown migration recorded before these tests run.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #544 (octo-server)
**Head SHA:** `b3c86473cbe76f8042f3e27f56f59f71a02913b1`
**Scope:** Configurable sticker upload limits + opt-in static compression

## Summary

This PR introduces config-driven sticker upload limits (max size, max dimension, allowed formats) via SystemSettings with 60-second atomic snapshot reload, plus an opt-in static compression pipeline for JPEG/PNG stickers. The compression uses goroutine+timer-based timeout and a mutex-gated concurrency slot, with fail-open semantics (skip compression on saturation/timeout/error). New Prometheus metrics track compression outcomes. Test coverage includes integration tests, unit tests, benchmarks, and config validation tests.

Overall: well-structured feature addition with thoughtful concurrency design. Two P2 findings; no P0/P1 blockers.

## Verification

- ✅ **Concurrency slot lifecycle** — `releaseCompressSlot` is deferred inside the worker goroutine (`sticker_compress.go:196`), not in `Compress()`. Slot hold time matches CPU work exactly; no leak on timeout or context cancellation.
- ✅ **APNG detection correctness** — `hasAPNGActlChunk` scans for `acTL` chunk before `IDAT` (`sticker_compress.go:107-145`). Correctly rejects animated PNGs from static compression path.
- ✅ **Config clamp logic** — `stickerClampIntUpper` applies hard caps (`system_settings.go:50-62`), preventing operator misconfiguration from exceeding server-side safety bounds.
- ✅ **Fail-open compression** — All error paths in `Compress()` return `stickerCompressOutcomeSkipped` or `stickerCompressOutcomeFailed`, falling through to uncompressed upload. No path bricks the upload flow.
- ✅ **Format intersection fallback** — `StickerUploadAllowedFormats()` intersects config CSV with `stickerUploadRasterAllowlist`; empty intersection falls back to full default set (`system_settings.go:85-108`). Prevents accidental lockout from misconfigured format list.
- ✅ **Metric pre-warming** — `initStickerCompressMetrics` pre-registers histogram buckets for all outcome+format combinations (`sticker_metrics.go:42-65`). Low-cardinality label space confirmed.
- ✅ **Sign provenance** — `stickersig.Sign` at `api.go:395` binds to the final stored path via `uploadReader`, not the original upload filename. Provenance chain intact.

Static analysis only at head `b3c86473cbe76f8042f3e27f56f59f71a02913b1`; build and tests not executed in this environment.

## Findings

No P0/P1 issues; two P2 items below.

### P2 — `stickerUploadExt` not updated to use config-driven allowed formats (`modules/file/api.go:324`)

`stickerUploadExt` at `modules/file/const.go:161` is called by the handle-generation endpoint to choose the storage extension for sticker upload paths. It still uses the hardcoded `stickerUploadExts` map rather than the new config-driven `stickerLimits.allowedFormats`.

When an operator narrows `sticker.upload_allowed_formats` (e.g. removing `.gif`), the handle endpoint returns a path with `.gif` extension, but the subsequent `uploadFile` call rejects the format at `modules/file/api.go:333`. No data is stored incorrectly, but the client wastes a round trip.

Diff-scope: amplified. Before this PR, both paths used the same hardcoded set. The PR added config-driven narrowing in `uploadFile` but did not propagate it to the handle path.

Fix: update the handle endpoint at `api.go:324` to consult `stickerLimits.allowedFormats` (or a helper derived from it) when choosing the storage extension.

### P2 — Compression always re-encodes even when source already meets target (`modules/file/sticker_compress.go:282`)

`doCompressStaticSticker` unconditionally decodes and re-encodes the image even when `len(src) <= targetKB*1024`. For JPEG at Q85, re-encoding a small already-compressed source introduces generation loss and can produce slightly larger output. For PNG, re-encoding costs CPU without benefit when the source is already under target.

Diff-scope: new. Design gap in the new compression pipeline.

Suggested fix: short-circuit when `int64(len(src)) <= int64(targetKB)*1024` — return the original bytes as `compressed`:

```go
if targetKB > 0 && int64(len(src)) <= int64(targetKB)*1024 {
    return stickerCompressResult{
        Outcome: stickerCompressOutcomeCompressed,
        Bytes:   src,
        Size:    int64(len(src)),
    }, nil
}
```

## Data-flow tracing

- **`stickerLimits` snapshot** → `f.stickerLimits()` at `api.go:298` reads atomic pointer → populated by `StickerSystemSettings.Snapshot()` every 60s → `stickerClampIntUpper` enforces hard caps at snapshot time → consumed values are bounded and current.
- **`uploadReader` reassignment** → initially `file` (original upload) → if compression succeeds, reassigned to `bytes.NewReader(result.Bytes)` at `api.go:371` → consumed by `service.UploadFile` at `:389` and `stickersig.Sign` at `:395`. Path correct; no stale reader reference.
- **`finalSize`** → set to `compressed.Size` on success (`:372`), `fileSize` otherwise (`:363`) → returned in JSON response at `:418`. Correctly reflects stored object size.
- **`compressor.Compress` goroutine** → spawned at `sticker_compress.go:188` → `tryAcquireCompressSlot` at `:193` (inside goroutine) → `releaseCompressSlot` deferred at `:196` (inside goroutine). Slot lifecycle matches CPU work, not caller lifetime. Timer at `:180` fires context cancellation; goroutine exits on `<-ctx.Done()` at `:227`.
- **`stickerUploadAllowedFormats` intersection** → config CSV parsed → intersected with `stickerUploadRasterAllowlist` → empty result falls back to full default. No path produces empty allowed set.

**C1 — Dual-path parity:** Hit (documented in P2 finding #1). Handle endpoint (`api.go:324`) and upload endpoint (`api.go:333`) use divergent format sources after this PR introduces config-driven narrowing.

**C2 — Control-flow ordering / nested reuse:** Clear. `doCompressStaticSticker` is called once per upload. No nested reuse or ordering hazard.

**C3 — Authorization boundary ≠ capability boundary:** N/A. No new endpoints or tool exposure. Upload path reuses existing auth middleware.

**C4 — Authorization lifecycle / container-member state cascade:** N/A. No auth/jail/permission changes.

**C5 — Build/note pass ≠ runtime path correctness:** Clear. No build artifacts, browser extensions, or packaging changes. Go module compiles; no runtime path divergence from build artifact.

**C6 — Governance/policy/security doc self-consistency:** N/A. No policy or governance docs changed.

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 blockers. Two P2 findings are non-blocking: (1) handle endpoint format parity is a wasted-round-trip issue, not a correctness issue; (2) compression short-circuit is an efficiency optimization. Both are safe to address in follow-up. Recommend merge with P2 tracked.

OctoBoooot (dismissed):
## Review: configurable sticker upload limits + opt-in static compression (#544)

**Verdict:** Approve — the image-decode DoS surface (the one that matters here) is defended in depth and byte-verified; config tunables are clamped un-neuterable; CI is fully green. No blockers; a few optional polish notes.

I focused verification on the decode-bomb / decode-OOM attack surface, since a sticker is sent to the conversation peer (a bad decode is a cross-user DoS, not just a self-inflicted one).

### Decode-bomb defense — complete and correct (byte-verified)
1. `modules/file/api.go:261` `http.MaxBytesReader` + `:286` sticker-specific `maxSize` byte cap.
2. `:365` format allowlist + magic-byte validation (`ValidateMagicNumber`).
3. **`:415` `image.DecodeConfig(file)` reads only the header for W×H and rejects when either dimension exceeds `stickerLimits.maxDim` — BEFORE any full `image.Decode`.** This is the gate that stops a small high-compression-ratio file from decoding to a gigapixel bitmap. Crucially it runs for **all** sticker uploads (`if isStickerUpload`), not just the compression path, so a bomb can't slip through the no-compress branch. Reader is reset (`file.Seek(0)`) afterward, with error handling.
4. Compression is opt-in and bounded: timeout timer, concurrency slot, and `recover`. The slot is released by the **worker goroutine at real CPU-completion**, not on the timeout return path — so a timed-out compression can't leak a slot and let real concurrency exceed `max_concurrency` (this is the right fix; the naive `defer release` would under-count). APNG is explicitly detected to avoid a first-frame-only re-encode.

So `sticker_compress.go`'s `image.Decode` only runs on bytes that already passed the dimension gate — not an unbounded bomb.

### Config safety — un-neuterable (byte-verified)
- Every new tunable reads through `stickerClampIntUpper(v, default, hardCap)` (`system_settings.go:838`): `v <= 0 → default`, `v > hardCap → hardCap`. So a hostile/mistaken DB or admin edit **cannot** set `upload_max_dimension` to 0/negative (which would disable the gate) or to an absurd value (which would defeat it) — each knob has its own hard cap and a safe floor. This is exactly the discipline a security-relevant limit needs.

### Praise
- The slot-release-on-CPU-completion detail (rather than `defer release` on the timeout path) shows the concurrency-bound was reasoned about under the timeout race — a subtlety most implementations get wrong, and it's called out in a comment for the next reader.
- Putting the dimension gate at the **ingress** via `DecodeConfig` (header-only) rather than after a full decode is the correct architectural placement — it protects the inline renderer and the peer, not just this process.

### Minor / nits (non-blocking)
- The legacy GET upload path still uses a hardcoded `.png` extension (`api.go:179`) — can confuse clients uploading other allowed formats. Minor.
- `uploadFile` could nil-panic if the compressor is nil in an unusual construction path — a defensive guard would harden it. Minor.
- `normalizeStickerCompressFormat` case handling — confirm it's fully case-insensitive (`.JPG` vs `.jpg`). Nit.

### Notes
- CI green across Build/Test/Vet/Lint/i18n/osv/secret-scan and `check-sprint`. `reviewDecision: REVIEW_REQUIRED` / `mergeStateStatus: BLOCKED` — approved on the merits, but the required-review count must still be met before merge.
- Coverage sniff flagged `file/api.go` / `sticker_metrics.go` as uncovered — heuristic false positive: `api_sticker_compress_test.go` + `sticker_compress_test.go` + `sticker_compress_bench_test.go` + `system_settings_sticker_upload_test.go` exercise the paths.

yujiawei (dismissed):
# Code Review — PR #544 (octo-server)

Reviewed at head SHA `b3c86473cbe76f8042f3e27f56f59f71a02913b1` against merge-base with `main`. I read the full diff, the task brief, and the surrounding upload path, then built and tested locally.

**Local verification**
- `go build ./modules/file/... ./modules/common/...` — clean
- `go vet ./modules/file/... ./modules/common/...` — clean
- `go test ./modules/file/ -run 'Sticker|Compress|APNG|Timeout' -race -count=1` — pass
- `go test ./modules/common/ -run 'StickerUpload|StickerCompress'` — pass (the other `modules/common` failure is a pre-existing DB-backed test needing MySQL, unrelated to this PR)

## 1. Spec compliance

**Spec: ✅**

- Missing requirements: none material. All seven `system_setting` keys exist in the schema slice, are accepted by the manager write path, and are read through `SystemSettings` clamp getters. Compression is opt-in (default OFF), static jpg/png only, webp/gif/animated validate-only, over-target rejected, fail-open on decode/encode/timeout/saturation, metrics pre-warmed. The one literal sub-clause not implemented is the "logs a warning on clamp" wording in the acceptance list — the clamp getters (`stickerClampIntUpper`) clamp correctly but do not log. This matches the existing house convention (`getIntClamped`, the incoming-webhook RPS/burst getters also clamp silently in the hot read path), so I'm treating it as an accepted deviation rather than a gap. See Q1 below for the operability angle.
- Extra/out-of-scope: none. No webp/gif re-encoding, no cgo dep, no client contract change, no i18n envelope migration — all correctly deferred.
- Divergence: none. The default config reproduces today's hard-coded behavior byte-for-byte (verified by the disabled-path regression test and the fallback in `stickerLimits()`).

On the handle contract specifically: `stickersig.Sign(uid, fullURL)` binds provenance `(uid, path)`, and phase-one compression preserves the extension so the path — and thus the handle — is stable over the final stored object. Content integrity is carried separately by the `sha512` field, which is correctly computed over `uploadReader` (the post-compression bytes). This is consistent with the brief ("minted over the final stored path, unchanged in phase one"). No issue.

## 2. Code quality

**Quality: Approved**

No P0/P1 findings. The stability isolation is well done: the compression slot is deliberately held for the worker goroutine's lifetime (not released on the timeout return), which is the correct fix for the slot-leak-under-timeout hazard, and it's covered by a dedicated regression test (`TestStickerCompressor_TimeoutDoesNotBleedConcurrency`). The APNG `acTL`-before-`IDAT` scan runs before slot acquisition and prevents `image/png` from silently flattening an animated PNG to its first frame. Decompression-bomb ordering is correct: the header-only `image.DecodeConfig` dimension gate runs before the full `image.Decode` in the compressor, so the decoded pixel budget is bounded by `maxDim²`.

P2 / suggestions (none blocking):

- **P2 — GET-side upload-URL generation is not aligned with the new configurable format set.** `getUploadURL` still calls `stickerUploadExt(c.Query("filename"))` (`modules/file/const.go:161`), which uses the hard-coded legacy allowlist and falls back to `.gif` when `filename` is absent. The POST path now validates ext against the configurable `stickerLimits.allowedFormats` (`modules/file/api.go:328`). With the default format set these agree, but if an operator narrows formats to exclude `gif`, a client that omits `filename` gets a `.gif` presigned URL that the POST side then rejects. No regression on default config; consider reading the same effective format snapshot on the presign side (or documenting that narrowing requires clients to send `filename`).

- **P2 — `upload_allowed_formats` with an all-invalid value falls back to the full default set** (`modules/common/system_settings.go:910`). E.g. a stored value of `mp4` yields the 5 built-in raster formats rather than an empty set. This is documented as intentional ("a bad config cannot dark-close the feature") and can never widen to a non-raster type, so the narrow-only invariant holds — but an operator who fat-fingers the CSV silently gets the full default rather than an error. Acceptable as designed; flagging for awareness.

- **P2 — over-hard-cap admin writes are accepted, then silently clamped with no feedback.** Because these keys are `Positive:true`, the write path (`api_manager_system_setting.go:220`) only rejects `≤0` and skips the `[settingIntMin, settingIntMax]` range check; a value like `upload_max_size_kb=99999` is accepted, stored, and then clamped to `5120` at read with no log at either end. An admin raising the cap past the hard limit gets no signal that it didn't take. Consider rejecting `> hardCap` at the write path for these keys, or logging once on clamp.

- **P2 — the compressor rereads live settings instead of the request's locked snapshot.** `uploadFile` locks `stickerLimits := f.stickerLimits()` at entry and comments that "one snapshot governs validation, compression, and response," but `Compress()` reads `StickerUploadMaxDimension()`, `StickerCompressTargetKB()`, `StickerCompressTimeoutMs()`, and `StickerCompressMaxConcurrency()` live (`modules/file/sticker_compress.go:150,159,161,162`). A 60s snapshot reload between the dimension gate and compression could make compression use a different `maxDim`/`targetKB` than the value the upload was validated against. Impact is low (values are already clamped and the window is narrow), but the code doesn't match the comment's stated invariant — either thread the snapshot through or soften the comment.

- **Nit — the added const block is not gofmt-clean.** `gofmt -l modules/common/system_settings.go` flags it; the new `defaultStickerUploadMaxSizeKB` / `...HardCap` pairs carry alignment spacing gofmt would collapse. `go vet` passes but a `gofmt` CI gate would not.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. No P0/P1. The P2 items above are worth addressing (the GET/POST format-alignment one especially, if narrowing formats is an intended operator workflow), but none block merge given the feature ships default-OFF and default-config behavior is unchanged.

OctoBoooot (dismissed):
## Review (delta): configurable sticker upload limits + opt-in static compression (#544)

**Verdict:** Approve — re-affirming at the new head; this delta cleanly addresses the round-1 🟡 nits plus one deeper cross-stage-consistency fix, no regression.

Delta since my APPROVE (b3c86473): ahead 1 / behind 0 (no rebase noise). Verified each change:

### Fixes verified
- ✅ **GET/POST format-config drift (my 🟡 + yujiawei's).** `modules/file/api.go:181` now calls `f.stickerUploadExtForRequest(...)`, which matches the client filename against the **currently-configured** allowed-formats set (the POST-side validation set) instead of the hardcoded 5. The hardcoded `stickerUploadExt` is retained only for the settings-unmounted unit-test path, now documented as such (`const.go`). So narrowing `upload_allowed_formats` no longer desyncs the preflight URL from POST validation.
- ✅ **Nil-compressor panic guard (my 🟡).** `uploadFile` (`api.go:473`) now gates on `f.compressor != nil` before `Compress`.
- ✅ **F7 — compress-param snapshot consistency (deeper fix).** `stickerLimitsSnapshot` now captures `compressTargetKB/MaxConcurrency/TimeoutMs` at request entry, and `Compress(ext, src, params)` takes them via `compressParams()`. Byte-verified `Compress`'s body uses `params.MaxDim/TargetKB/MaxConcurrency/TimeoutMs` throughout with **no `c.settings.*` re-read** — so the dimension gate's `maxDim` and the compression `Fit`'s `maxDim` are provably the same value, immune to a 60s settings reload landing mid-request. Real cross-stage-consistency hardening.
- New tests: `sticker_upload_ext_test.go`, `sticker_metrics_test.go`.

### Intact from prior review (unchanged)
- The decode-bomb defense (ingress `image.DecodeConfig` W×H pre-check on all sticker uploads, layered behind byte-cap + magic + format allowlist) and the `stickerClampIntUpper` un-neuterable config floors are untouched and still hold.

### Note (non-blocking, still open)
- The "already ≤ target size and within maxDim → skip re-encode, keep original bytes" fast path (mochashanyao's 🟡) does not appear to be added — an enabled-compression pass still re-encodes an in-spec JPEG at quality 85, a minor generational-quality/CPU cost on the opt-in path. Optional optimization, not a blocker.

### Notes
- `check-sprint` green; code CI (Build/Test/Vet/Lint) pending — confirm green before merge. `reviewDecision: REVIEW_REQUIRED` / `mergeStateStatus: BLOCKED` until required reviews re-clear at this head (my earlier APPROVE was at the now-superseded b3c86473).

Jerry-Xin (dismissed):
The PR is relevant to octo-server and the implementation is generally sound: sticker upload limits are read-side clamped, compression is isolated behind opt-in settings, and the upload response/signing path uses the final stored bytes.

**🔴 Blocking**
None.

**💬 Non-blocking**
🟡 Warning: The over-limit error/log reads `f.settings.StickerCompressTargetKB()` instead of the request snapshot at [modules/file/api.go](/Users/mlamp/.openclaw/tmp/pr-review-xkvt32ib/repo/modules/file/api.go:495). If settings reload between snapshot creation and rejection, the displayed target can differ from the target actually used by `Compress`. Use `stickerLimits.compressTargetKB` here to preserve the “one request, one settings snapshot” invariant.

🟡 Warning: The admin write path still accepts above-hard-cap values for these new positive integer settings; it only rejects `<=0` at [modules/common/api_manager_system_setting.go](/Users/mlamp/.openclaw/tmp/pr-review-xkvt32ib/repo/modules/common/api_manager_system_setting.go:220), while runtime clamping happens later in getters such as [modules/common/system_settings.go](/Users/mlamp/.openclaw/tmp/pr-review-xkvt32ib/repo/modules/common/system_settings.go:851). Runtime behavior is safe, but persisted values like `999999` can confuse operators and don’t fully match the PR wording that hard caps are enforced on the write path. Consider adding per-key max metadata or explicit validation tests for the sticker settings.

🔵 Suggestion: `stickerCompressSettings` now mostly duplicates values passed through `stickerCompressParams`; after the snapshot refactor, `Compress` only needs `StickerCompressEnabled()` from the settings object. Trimming that interface would make the dependency clearer.

**✅ Highlights**
The concurrency slot lifetime is correctly bound to worker completion, including timeout paths, at [modules/file/sticker_compress.go](/Users/mlamp/.openclaw/tmp/pr-review-xkvt32ib/repo/modules/file/sticker_compress.go:241).

APNG detection before slot acquisition avoids silently flattening animated PNGs at [modules/file/sticker_compress.go](/Users/mlamp/.openclaw/tmp/pr-review-xkvt32ib/repo/modules/file/sticker_compress.go:220).

The upload path signs and stores using the selected final reader, so `sha512`, `size`, storage bytes, and `sticker_handle` stay aligned at [modules/file/api.go](/Users/mlamp/.openclaw/tmp/pr-review-xkvt32ib/repo/modules/file/api.go:527).

Verification run:
`go test ./modules/file -run 'TestUploadFile_Compress|TestStickerCompressor|TestHasAPNG|TestCanCompress|TestNormalizeSticker|TestStickerUploadExt'` passed.
`go test ./modules/common -run 'TestSystemSettings_Sticker(Upload|Compress)'` passed.
`git diff --check` passed.
`go test ./modules/file ./modules/common` could not complete because the local test database has an unrelated unknown migration recorded.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #544 (octo-server)

## Summary

This PR makes sticker upload limits (max size, max dimension, allowed formats) operator-configurable via `system_setting`, and adds an opt-in server-side compression pipeline for static JPEG/PNG stickers. The compression runs in a goroutine with a configurable timeout and process-wide concurrency cap; all failure modes fail open (original bytes pass through). APNG is explicitly detected and skipped to avoid silent animation loss. The design is thoughtful: request-scoped snapshots prevent cross-phase config drift, hard caps defend against misconfiguration, and the format allowlist can only narrow (never widen) the built-in raster set.

## Verification

- ✅ **Snapshot consistency** — `stickerLimits()` at `modules/file/api.go:213` captures all 7 config values once at request entry. Size check (`:283`), format check (`:321`), dimension check (`:423`), and compression params (`:474`) all consume from this single snapshot. The `compressParams()` derivation at `modules/file/sticker_compress.go:67` ensures maxDim/targetKB/maxConcurrency/timeoutMs are同源.
- ✅ **Slot lifecycle (P1 regression from earlier commit)** — `releaseCompressSlot` is deferred inside the worker goroutine (`:237`), not in `Compress()`'s return path. After timeout, the slot stays occupied until the dangling worker finishes. `TestStickerCompressor_TimeoutDoesNotBleedConcurrency` confirms with maxConcurrency=1 + slow worker that a second call gets `concurrency_saturated`, not a new timeout.
- ✅ **APNG detection** — `hasAPNGActlChunk` (`:296`) scans for `acTL` before `IDAT`, short-circuits on `IDAT`-first (static PNG), handles malformed inputs (empty, short, length overflow). Placed before `tryAcquireCompressSlot` so APNG never occupies a slot. `TestStickerCompressor_APNGSkips` uses a fatal-on-call doCompress to prove the code never reaches the worker.
- ✅ **Fail-open completeness** — Every non-compressed outcome (disabled, format, animated, concurrency_saturated, timeout, decode error) leaves `uploadReader` and `finalSize` pointing at original bytes. The upload proceeds normally.
- ✅ **File pointer integrity** — After `io.ReadAll(file)` at `:475`, all compress branches produce a fresh `bytes.NewReader`. The sign step (`:533`) and upload step (`:548`) both `Seek(0, io.SeekStart)` before reading. Non-sticker uploads are unaffected (added Seek is a no-op).
- ✅ **Config clamping** — All int getters use `stickerClampIntUpper` (`:845`) with [1, hardCap] bounds. Non-positive and non-numeric values fall back to defaults. Tests cover boundary values including hard-cap-exact.
- ✅ **Format allowlist narrowing** — `StickerUploadAllowedFormats()` (`:882`) intersects config CSV with `stickerUploadRasterAllowlist`. Non-raster tokens are dropped. Empty intersection falls back to the full default set. Tests cover normalization, dedup, non-raster rejection, and empty-intersection fallback.
- ✅ **Test coverage** — 12 test files with integration tests covering: compressed JPEG shrinks and resp.size matches stored bytes; GIF passes through unchanged; over_limit rejects without reaching storage; nil compressor doesn't panic; handle binds to final URL; APNG skips before slot acquisition; timeout doesn't bleed concurrency; aspect ratio preserved; no upscale for small sources.

Static analysis only at head `432c76bab9aa19367e1ca84ef3eaef7a859dd90f`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Two P2 items below.

### P2 — `over_limit` error message reads live settings, not request snapshot (`modules/file/api.go:502`)

The `over_limit` handler reads `targetKB := f.settings.StickerCompressTargetKB()` from the live atomic snapshot, not from `stickerLimits.compressTargetKB` (the value captured at request entry and actually used by the compression pipeline). If a settings reload happens between request entry and this error path (60s reload window), the user-facing error message "贴纸压缩后仍超过 XKB" could display a different threshold than the one that was actually enforced. The rejection decision itself is correct — only the displayed number can drift.

```go
// Fix: use the snapshot value
targetKB := stickerLimits.compressTargetKB
```

### P2 — Duplicated raster allowlist creates future drift surface (`modules/common/system_settings.go:809`)

`stickerUploadRasterAllowlist` is a deliberate copy of `modules/file/const.go:stickerUploadExts`. The comment says "keep in sync," but now that production POST-side validation at `modules/file/api.go:321` gates on `stickerLimits.allowedFormats` (which intersects config with this allowlist), any future addition to `stickerUploadExts` without a matching update here would silently exclude the new format from both GET preflight and POST validation. The nil-settings fallback path (`stickerLimits()` when `f.settings == nil`) still uses the original `stickerUploadExts` directly, creating a test/production divergence for old unit tests. Consider exporting a single canonical list or adding a startup assertion that both are identical.

## Data-Flow Traceback

| Consumed data | Upstream source | Verified flow |
|---|---|---|
| `stickerLimits.maxSize` | `StickerUploadMaxSizeKB()` → `stickerClampIntUpper` → `int64(kb)*1024` | ✅ Flows to size check at `:283` |
| `stickerLimits.allowedFormats` | `StickerUploadAllowedFormats()` → intersect with raster allowlist | ✅ Flows to format check at `:321` and `stickerUploadExtForRequest` at `:184` |
| `stickerLimits.maxDim` | `StickerUploadMaxDimension()` → clamp | ✅ Flows to dimension check at `:423` and `compressParams().MaxDim` → `imaging.Fit` |
| `params.TargetKB` | `StickerCompressTargetKB()` → clamp → snapshot | ✅ Flows to `doCompressStaticSticker` over_limit check at `:375` |
| `params.MaxConcurrency` | `StickerCompressMaxConcurrency()` → clamp → snapshot | ✅ Flows to `tryAcquireCompressSlot` at `:248` |
| `params.TimeoutMs` | `StickerCompressTimeoutMs()` → clamp → snapshot | ✅ Flows to `time.NewTimer` at `:252` |
| `uploadReader` after compress | `bytes.NewReader(result.Bytes)` or `bytes.NewReader(srcBytes)` | ✅ Flows to sign (`:533`) and upload (`:548`) via Seek(0) |
| `finalSize` in response | `result.Size` (compressed) or `fileHeader.Size` (unchanged) | ✅ Flows to `resp["size"]` at `:574` |

## Blind-Spot Checklist

- **C1 — Dual-path parity**: GET preflight (`api.go:184`) uses `stickerUploadExtForRequest`; POST validation (`:321`) uses `stickerLimits.allowedFormats`. Both derive from settings within their respective requests. Within a single request, values are consistent. Separate HTTP requests get separate snapshots — inherent to any config system. **Clear.**
- **C2 — Control-flow ordering**: Compression runs after all validation gates (size → format → magic → dimension). No nested reuse of compression results. **Clear.**
- **C3 — Authorization boundary**: No auth/permission changes in this PR. N/A.
- **C4 — Auth lifecycle**: N/A.
- **C5 — Build/runtime**: N/A (no build/packaging changes).
- **C6 — Governance docs**: `.octospec/` files are task planning docs, not governance/strategy. N/A.

## Cross-Round Blocker Re-Review (R6)

N/A — first review round.

- `EnsureSystemSettings` is a process-wide singleton with atomic snapshot swap — the `File.New(ctx)` constructor correctly wires `settings` and `compressor` from the same instance.
- `normalizeStickerCompressFormat` in `modules/file/sticker_metrics.go:120` correctly lowercases and strips the leading dot before label matching, preventing cardinality explosion.
- `hasAPNGActlChunk` uses `int64` arithmetic to prevent `uint32 length + 12` overflow — malformed chunk lengths that exceed buffer size are caught by the `next > int64(len(data))` guard.
- `stickerClampIntUpper` with `v <= 0` returns fallback (not 0), preventing a "dark-close" where maxSize=0 would reject all uploads.
- The `compressor == nil` guard at `api.go:474` prevents nil-pointer dereference for old unit-test `&File{}` construction paths. `TestUploadFile_CompressorNilDoesNotPanic` covers this.
- `imaging.Fit` preserves aspect ratio and does not upscale — verified by `TestDoCompressStaticSticker_PreservesAspectRatio` and `TestDoCompressStaticSticker_NoUpscaleWhenSourceUnderMaxDim`.

## Verdict: APPROVED

No P0/P1 issues. The two P2 items are non-blocking: one is a narrow race-window inconsistency in an error message, the other is a maintainability concern about a duplicated list. The core compression pipeline, snapshot design, slot lifecycle, APNG detection, fail-open semantics, config clamping, and test coverage are all solid.

[Octo-Q] verdict: APPROVE — no P1 issues; two P2 non-blocking items noted.

yujiawei (dismissed):
# Code Review — PR #544 (octo-server)

Reviewed at head SHA `432c76ba`, base `origin/main` (merge-base `8aba30e7`). 14 files, +2342/-28. `go build`, `go vet ./modules/file/ ./modules/common/`, and the no-infra unit tests pass locally; full-infra tests (MySQL/Redis) were not run.

This review combines a line-level adversarial pass, a whole-repo cross-file pass, and an independent data-flow trace of the upload path. A second cross-family advisory model was attempted but timed out and is **absent** from this review — its perspective was not incorporated.

## 1. Spec compliance

**Spec: ✅**

Checked against `.octospec/tasks/sticker-upload-compression/brief.md`.

- **Under-build (missing):** None blocking. All seven `system_setting` keys exist in the canonical schema, are validated by the admin write path (`Positive` for ints, string for `upload_allowed_formats`), and are read through clamp getters. Hard caps/defaults match the brief exactly (size 1024/5120 KB, dimension 512/1024 px, target 1024/5120 KB, concurrency 4/32, timeout 2000/10000 ms). Compression is opt-in (`compress_enabled` default false), static jpg/png only, over-limit rejects, all error paths fail-open, `sticker_handle` binds the final stored URL, and the four new metric dimensions are pre-warmed to 0. One acceptance bullet ("clamp getters … **log a warning**") is not implemented — see Quality P2-2.
- **Over-build:** None. No new client-facing fields; response shape (`path`/`name`/`size`/`ext`/`sha512`/`sticker_handle`) is preserved. No changes to non-`type=sticker` uploads.
- **Deviation:** The brief's *Background* sketches animated detection via GIF `DecodeAll` frame-count and WebP RIFF/VP8X `ANIM` chunk scan. The implementation instead restricts compression to jpg/jpeg/png at the extension gate (`canCompressStickerExt`), so gif/webp never enter the compress pipeline at all, and adds explicit APNG `acTL` detection for the one format (PNG) where `image.Decode` would silently drop animation. This is a **stronger, fail-safe** guarantee than runtime frame-counting and fully satisfies the acceptance criterion ("webp/gif never re-encoded, metric `compress_skipped`"). Not a blocking deviation.

## 2. Code quality

**Quality: Approved**

No P0/P1 findings. Two candidate high-severity issues were investigated and **rejected**:

- **Decompression bomb (rejected):** `image.Decode` in `doCompressStaticSticker` (`sticker_compress.go:349`) fully decodes into memory, but the pre-existing `image.DecodeConfig` dimension gate (`api.go:415-437`) runs **before** the compression block (`api.go:473`) and rejects any image whose W or H exceeds `stickerLimits.maxDim` (hard cap 1024). Decoded bitmap is therefore bounded to ~1024×1024×4 ≈ 4 MB, and `srcBytes` ≤ `maxSize` (hard cap 5 MB). Ordering is correct; no unbounded allocation.
- **APNG chunk-walk integer overflow (rejected):** `hasAPNGActlChunk` (`sticker_compress.go:326`) guards every slice access with `pos+8 <= len(data)`, computes the next offset in `int64`, and rejects `next <= pos || next > len(data)`. `next` is a sum of positive terms bounded by ~4 GB, so it cannot go negative. No OOB, no panic; the pre-slot `recover` is defense-in-depth only.

Positively verified load-bearing invariants:

- **Concurrency-slot lifetime (P1-fix confirmed):** the slot is released by the worker goroutine's `defer` (`sticker_compress.go:244`), **not** on `Compress()`'s timeout return, so a timed-out compression keeps holding its slot until the real CPU work finishes — true concurrency cannot exceed `max_concurrency`. `TestStickerCompressor_TimeoutDoesNotBleedConcurrency` asserts exactly this (a request following a timeout gets `concurrency_saturated`, not a fresh slot).
- **Snapshot invariant:** one `stickerLimits` snapshot is taken per request (`api.go:212`) and all size/dimension/format checks + `compressParams` derive from it, so a 60s settings reload cannot make validation and compression disagree mid-request.
- **Byte fidelity / handle binding:** `uploadReader` and `finalSize` move together in every branch (compressed / failed / skipped / non-compressible); `sha512` and `service.UploadFile` both `Seek(0)` on `uploadReader`; the handle is signed over the URL of the finally-stored object. `TestUploadFile_CompressEnabled_HandleTiesToFinalStoredObject` and the GIF/nil-compressor byte-for-byte tests cover this.

### P2 findings (non-blocking)

- **P2-1 — Misleading comment about `stickersig` coverage** (`modules/file/api.go:469-470`). The comment says `stickersig.Sign 天然基于压缩后的字节` ("naturally signs the compressed bytes"). `stickersig.Sign(uid, path)` only HMACs the `(uid, path)` strings (`pkg/stickersig/stickersig.go:118-150`) — it provides URL/uploader provenance, **not** byte-content integrity. The code is correct (the signed URL does point at the final compressed object), but the wording could mislead a future reader into assuming content-integrity protection. Suggest rewording to "the signed URL resolves to the final (post-compression) stored object".
- **P2-2 — Clamp getters do not log on clamp** (`modules/common/system_settings.go:838-846`). `stickerClampIntUpper` silently returns the hard cap when a configured value exceeds it. The brief's acceptance criterion says an over-cap value "returns the hard cap (not the input) **and logs a warning**". Clamping is correct and unit-tested; only the ops-visibility warning is missing. Consider a `Warn` when `v > hardCap` so a bad admin edit is observable. (`SystemSettings` already has a logger.)
- **P2-3 — Test-coverage gaps (nice-to-have):**
  - No dedicated `compress_enabled=false` end-to-end "stored bytes identical to `main`" regression test. The fail-open (`TestUploadFile_CompressorNilDoesNotPanic`) and GIF-passthrough tests assert byte-for-byte identity on adjacent paths, so the invariant is indirectly covered, but an explicit disabled-path test would lock the headline "zero-impact when off" guarantee.
  - `allowed_formats` intersection is tested with all-illegal and first-legal CSVs but not an interleaved `jpg,mp4,webp` (legal-illegal-legal) case. The loop is provably position-independent, so this is completeness only.
  - The timeout-slot-leak regression exists at the compressor unit level but not at the `uploadFile` (API) integration level.

  None of these indicate a defect; the existing suite (19 common + ~17 file tests, incl. APNG chunk-order fuzz and benchmarks) is thorough.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. Feature is default-OFF, the upload path fails open in every non-`over_limit` branch, the decompression-bomb ordering and concurrency-slot lifetime are correct, and the handle binds the final object. The P2 items (one comment, one missing warning log, a few optional tests) are worth a quick follow-up but do not block merge.

Note: production enablement should still gate on the perf checklist the author tracks in `.octospec/tasks/sticker-upload-compression/perf-todo.md` (x86 bench re-run, decompression-bomb sanity, greyscale ramp).

OctoBoooot (approved):
## Review (delta): configurable sticker upload limits + opt-in static compression (#544)

**Verdict:** Approve — re-affirming; this delta is pure observability/consistency polish addressing prior 🟡s, no behavior change to the security-critical paths, no regression.

Delta since my last APPROVE (432c76ba): ahead 1 / behind 0. Three small, correct improvements:

### Verified
- ✅ **R6 — clamp is now operator-observable (addresses Jerry-Xin's admin-write-bounds 🟡).** `stickerClampIntUpper` becomes a method that emits a **one-shot Warn** when a sticker knob exceeds its hard cap (`system_settings.go`), deduped per `(key, value)` via `sync.Map` so the read hot path isn't spammed. The clamp *behavior* is unchanged — still floors `≤0` to default and caps at hardCap (the un-neuterable safety property verified in prior rounds) — this only adds the missing visibility so an over-limit admin edit doesn't silently clamp. This is exactly the right resolution: read-side clamp stays the enforcement, write-side gets a visible signal.
- ✅ **R1/F7 completion.** The over-limit error log (`api.go:499`) now reads `stickerLimits.compressTargetKB` (the request snapshot) instead of live `f.settings.StickerCompressTargetKB()` — closing the last spot where a logged number could desync from the value `Compress` actually used under a mid-request settings reload. Completes the snapshot-consistency work from the prior delta.
- ✅ **R5 — doc clarity.** Comment now correctly states `stickersig.Sign` HMACs `(uid, path)` for provenance (not a content hash), with `sha512` carrying content-integrity over the compressed bytes. Clarification only.

### Intact (verified prior rounds, untouched here)
- Decode-bomb defense (ingress `image.DecodeConfig` W×H pre-check on all sticker uploads) and the clamp floors — both unchanged.

### Notes
- `reviewDecision: REVIEW_REQUIRED` (reset by the push); `check-sprint` green; code CI (Build/Test/Vet/Lint) pending — confirm green before merge (this PR was genuinely CLEAN at the prior head 432c76ba after `Test` passed; the same must re-confirm here). `mergeStateStatus: BLOCKED` until reviews re-clear + Test green.
- The only remaining open 🟡 is the optional skip-re-encode-when-already-compliant fast path — still not added, still non-blocking.

Jerry-Xin (approved):
In scope for octo-server: this PR extends the existing sticker upload path, system settings, and metrics surfaces. I found no blocking correctness, security, or architecture issues.

**💬 Non-Blocking**
- 🟡 Warning: [api_manager_system_setting.go](/Users/mlamp/.openclaw/tmp/pr-review-p0omo_z5/repo/modules/common/api_manager_system_setting.go:220) only enforces `Positive` for the new integer settings; values above the hard caps are accepted into DB and only clamped on read. Runtime safety is preserved by [system_settings.go](/Users/mlamp/.openclaw/tmp/pr-review-p0omo_z5/repo/modules/common/system_settings.go:850), but the task brief says the manager write path enforces per-key hard caps. Either add per-key max validation to `settingDef` or update the brief to state “write accepts, read clamps.”
- 🔵 Suggestion: [sticker_compress.go](/Users/mlamp/.openclaw/tmp/pr-review-p0omo_z5/repo/modules/file/sticker_compress.go:208) still reads live `StickerCompressEnabled()` inside `Compress`, after `uploadFile` already captured a request snapshot. A reload from enabled to disabled can make the request skip compression despite `stickerLimits.compressEnabled` being true. It is fail-open and safe, but slightly weakens the documented one-snapshot invariant.

**✅ Highlights**
- Good placement of compression after magic/dimension validation and before storage/signature response generation.
- Concurrency-slot lifetime is correctly tied to the worker goroutine, including timeout paths.
- APNG detection before slot acquisition avoids silent animation loss.
- Focused tests are strong, including API-level byte capture, timeout/concurrency regression tests, and low-cardinality metric normalization.

Verification: `go test ./modules/file/ -race -count=1` passed. `go test ./modules/file/ ./modules/common/` was blocked in `modules/common` by an existing local migration-state panic (`unknown migration in database`), while `modules/file` passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #544 (octo-server)

## Summary

This PR layers two coupled capabilities onto the existing sticker upload path: (1) configurable upload limits (size, dimension, allowed formats) moved from hard-coded constants into `system_setting` with server-side hard caps, and (2) opt-in server-side compression for static jpg/png stickers with concurrency gating, timeout fail-open, and APNG detection. Both features default to OFF, making the PR zero-impact until an operator opts in.

The design is thoughtful — snapshot isolation prevents mid-request config drift, the concurrency slot lifecycle is correctly tied to actual CPU work (not the `Compress()` return), all error paths fail open to original bytes, and the read-side clamping with hard caps provides defence-in-depth against bad admin edits. Test coverage is extensive (32+ unit tests across settings getters, compressor behavior, APNG detection, concurrency slot regression, and integration tests).

## Verification

Static analysis only at head `914c713d28de1b78165e71e9d27519cc2471eafb`; build and tests not executed in this environment.

- ✅ **Snapshot isolation** — `stickerLimits()` is called once per request in `uploadFile` (`modules/file/api.go:211`); all downstream checks (size, dimension, format, compression) use the same `stickerLimitsSnapshot`. `compressParams()` derives all four compression parameters from the same snapshot, preventing cross-phase drift.
- ✅ **Concurrency slot lifecycle** — `modules/file/sticker_compress.go:230-240`: slot is released inside the worker goroutine via `defer c.releaseCompressSlot()`, NOT in the caller. On timeout, `Compress()` returns early but the slot remains held until the goroutine completes — proven by `TestStickerCompressor_TimeoutDoesNotBleedConcurrency`. This correctly prevents true in-flight work from exceeding `max_concurrency`.
- ✅ **APNG detection** — `hasAPNGActlChunk` (`modules/file/sticker_compress.go:280-310`) scans for `acTL` before `IDAT` per the APNG spec. Chunk traversal uses `int64` arithmetic to prevent uint32 overflow. Short-circuits on `IDAT` (static PNG) or truncated data. Called before slot acquisition, avoiding wasted slots for animated uploads.
- ✅ **Fail-open semantics** — All compression error paths (decode error, encode error, panic recovery, timeout, concurrency saturation) return original bytes to the upload pipeline. The `over_limit` path is the only path that rejects the upload, which is correct behavior (compressed output still exceeds target).
- ✅ **Read-side clamping** — `stickerClampIntUpper` (`modules/common/system_settings.go:849-863`) clamps to `[1, hardCap]` with one-shot warn dedup via `sync.Map`. Every integer setting has a hard cap. `StickerUploadAllowedFormats` intersects config with built-in raster allowlist — config can only narrow, never widen to non-raster formats.
- ✅ **sha512 + upload use compressed bytes** — `uploadReader` is reassigned to `bytes.NewReader(result.Bytes)` on success (`modules/file/api.go:490`), and both the sha512 hash and `service.UploadFile` read from `uploadReader`. `finalSize` tracks the actual stored byte count and is returned in the response.
- ✅ **sticker_handle invariant under compression** — `stickersig.Sign(uid, fullURL)` signs (uploader, URL), not content. Compression doesn't change path or URL, so the handle remains valid. Content integrity is carried by the separate `sha512` field.
- ✅ **GET/POST format parity** — `getFilePath` uses `f.stickerUploadExtForRequest()` (`modules/file/api.go:184`) and `uploadFile` uses `stickerLimits.allowedFormats[ext]` (`:324`). Both derive from the same `stickerLimits()` snapshot mechanism, keeping preflight URL generation consistent with upload validation.

## Findings

### P2 — `stickerCompressSettings` interface carries methods never called by `Compress()` (`modules/file/sticker_compress.go:160-167`)

The `stickerCompressSettings` interface declares five methods, but `Compress()` reads only `StickerCompressEnabled()` from `c.settings` — the other four (`StickerCompressTargetKB`, `StickerCompressMaxConcurrency`, `StickerCompressTimeoutMs`, `StickerUploadMaxDimension`) are read from the `params` argument instead, per the snapshot-isolation design. Test fakes implementing this interface must stub four methods that are never invoked. Trimming the interface to just `StickerCompressEnabled() bool` would make the contract match reality and reduce fake boilerplate.

### P2 — Cross-module allowlist duplication without automated parity guard (`modules/common/system_settings.go:831`)

`stickerUploadRasterAllowlist` in `modules/common` duplicates `stickerUploadExts` in `modules/file/const.go:150`. The comment at `:832` acknowledges the sync risk ("Keep in sync — the upload_allowed_formats getter uses this list as the outer bound"). A future PR adding a raster format (e.g., `.avif`) to `stickerUploadExts` without updating `stickerUploadRasterAllowlist` would silently let the new format through the upload path while the config intersection excludes it from operator-narrowed allowlists — a subtle behavioral divergence. The cycle-prevention rationale is sound, but a test asserting `stickerUploadRasterAllowlist` keys match `stickerUploadExts` keys would close the gap mechanically.

### Nit — `stickerClampWarned` dedup key includes the raw out-of-range value (`modules/common/system_settings.go:855`)

The dedup key `fmt.Sprintf("%s=%d>%d", key, v, hardCap)` is per-(key, value) pair. If an admin edits the setting from 9999 to 8888 (both over hardCap 5120), two separate warn lines are emitted. This is the documented behavior and arguably correct (each misconfiguration is independently operator-relevant), but on a flappy admin edit it could produce more log lines than expected. Not a functional issue.

## Data-Flow Trace

- **`stickerLimits.maxSize`**: `uploadFile` → `f.stickerLimits()` → `f.settings.StickerUploadMaxSizeKB()` → `stickerClampIntUpper(key, getInt("sticker","upload_max_size_kb",1024), 1024, 5120)` → int (KB) × 1024 → `int64` bytes. Consumed at `:283` for `fileHeader.Size` comparison. Data flows correctly.
- **`stickerLimits.maxDim`**: Same chain → `StickerUploadMaxDimension()` → clamp to [1, 1024]. Consumed at `:423` for `DecodeConfig` dimension check AND passed to `compressParams().MaxDim` for `imaging.Fit`. Both consumers get the same value from the same snapshot. Correct.
- **`stickerLimits.allowedFormats`**: `StickerUploadAllowedFormats()` → CSV split → intersect with raster allowlist → dedup → deterministic order. Returned as `[]string`, converted to `map[string]bool` in `stickerLimits()`. Consumed at `:324` for format check. Correct.
- **`uploadReader` / `finalSize`**: Initialized to `file` / `fileHeader.Size`. On compression success, reassigned to `bytes.NewReader(result.Bytes)` / `result.Size`. On fail-open, reassigned to `bytes.NewReader(srcBytes)` / `len(srcBytes)`. Both consumers (sha512 hash and UploadFile) read from `uploadReader` after `Seek(0)`. `finalSize` flows to response `size` field. Correct.

## Blind-Spot Checklist

- **C1 (dual-path parity)**: GET-side `getFilePath` and POST-side `uploadFile` both use `stickerLimits()` for format resolution. `stickerUploadExtForRequest` (GET) and `stickerLimits.allowedFormats[ext]` (POST) draw from the same settings mechanism. **Clear.**
- **C2 (control-flow ordering / nested reuse)**: `stickerLimits()` is called once per request in `uploadFile`; no double-call or ordering issue. `Compress()` is called at most once per request. **Clear.**
- **C3 (auth boundary)**: No new endpoints or tools exposed. Compression is gated on `isStickerUpload` which requires authenticated `type=sticker` upload. **Clear.**
- **C4 (auth lifecycle)**: N/A — no auth changes in this PR.
- **C5 (build/runtime)**: N/A — no build/packaging changes.
- **C6 (governance/docs)**: N/A — no governance/policy docs changed.

## Cross-Round Blocker Re-Review

- `hasAPNGActlChunk` chunk-traversal arithmetic with int64 overflow guard and short-circuit on IDAT.
- `tryAcquireCompressSlot` returns false for `max <= 0`, preventing zero-capacity from accepting slots.
- `sync.Map` zero-value initialization for `stickerClampWarned` — no explicit init needed.
- `stickerUploadRasterAllowlist` and `stickerUploadExts` are currently in sync (both have .gif, .png, .jpg, .jpeg, .webp).
- Panic recovery in the worker goroutine maps to `failed` outcome, not a process crash.
- Timer is `defer timer.Stop()`-ed, preventing timer leaks.
- The buffered channel (`ch := make(chan outcome, 1)`) prevents goroutine leak on timeout — the late-completing goroutine can write to the buffer without blocking.
- `normalizeStickerCompressFormat` bounds histogram label cardinality to 6 values.
- Histogram pre-warming in `init()` covers all (outcome, format) pairs.

## Verdict: APPROVED

Well-engineered feature addition with careful attention to stability isolation, fail-open semantics, and snapshot consistency. The concurrency slot lifecycle fix (slot held by goroutine, not by caller's return) and APNG detection before compression are particularly well-done. The two P2 items are non-blocking code-quality improvements.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Both P2 items are code-quality improvements with no runtime impact. The snapshot isolation, fail-open semantics, and concurrency slot lifecycle are all correctly implemented.

yujiawei (approved):
# Code Review — PR #544 (octo-server)

**Scope reviewed:** merge-base `8aba30e` → head `914c713d`, 14 files, +2429/-28. Built (`go build ./...` clean), `go vet` clean, `modules/file` tests pass under `-race`; the new no-infra `modules/common` sticker settings tests pass. (The other `modules/common` test failures are pre-existing MySQL-test-server dependencies unrelated to this diff.)

This review is organized in two gates: **spec compliance** first, then **code quality**.

---

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly what the task brief specifies, no more and no less.

- **Configurable limits** — `upload_max_size_kb` / `upload_max_dimension` / `upload_allowed_formats` moved to `system_setting`, read through the existing `SystemSettings` snapshot getters, defaults reproduce today's hard-coded behavior (1024 KB / 512 px / `gif,png,jpg,jpeg,webp`). Hard caps enforced read-side (`stickerClampIntUpper`): 5120 KB, 1024 px. `upload_allowed_formats` is intersected with the built-in raster allowlist and can only *narrow* — non-raster tokens (mp4/pdf/svg) are dropped in the getter (`system_settings.go` `StickerUploadAllowedFormats`), and an all-illegal config falls back to the full default set rather than dark-closing the feature.
- **Opt-in static compression (plan C)** — `compress_enabled` defaults false; static jpg/png decode → optional `imaging.Fit` downscale → re-encode (JPEG q=85 / PNG), metadata stripped by the encoder. `webp`/`gif`/animated are validate-only. Over-target after re-encode → rejected with `observeStickerUpload("compress_over_limit")`. No new dependency introduced (`imaging` already in `go.mod`).
- **Handle/size over final bytes** — `sha512` and `service.UploadFile` both read `uploadReader` (post-compression bytes); `resp["size"] = finalSize`; `stickersig.Sign` is over `fullURL`, which is unchanged in phase one because the extension is preserved. The integration test asserts the stored bytes equal the compressed bytes and that the handle verifies against the final URL.
- **Ops/observability** — timeout + concurrency gates, all error paths fail-open to original bytes; new counter dimensions and the `compress_duration_seconds` histogram are registered and pre-warmed.
- **Out-of-scope items respected** — no webp/gif re-encoder, no cgo, no change to the upload→register contract or the response field shape, no i18n-envelope migration (new rejections use `c.ResponseError`, consistent with the module).

No missing requirements, no unrequested additions, no deviations from the agreed approach.

---

## 2. Code quality

**Quality: Approved**

The implementation is careful and well-tested. Notable strengths verified by reading the code, not just the PR description:

- **Concurrency-slot lifetime is correct.** The slot is released by the worker goroutine's `defer`, not by `Compress()`'s return, so a timeout return cannot free a slot while the worker is still burning CPU — the process-wide cap is honored. `File` (and therefore `compressor` with its `inflight` counter) is a genuine process singleton: `New(ctx)` runs once via `SetupAPI()` at boot (`octo-lib/module/module.go`), so the gate is process-global and meaningful. `TestStickerCompressor_TimeoutDoesNotBleedConcurrency` correctly proves the non-bleed with maxConcurrency=1 + slow worker.
- **Decompression-bomb defense is ordered correctly.** `image.DecodeConfig` (header-only) enforces the dimension cap *before* any full decode, and the compression path reuses the same `maxDim` from the request snapshot — a 1 KB→10000² bomb is rejected before reaching `image.Decode`.
- **APNG handling is sound.** `hasAPNGActlChunk` scans for `acTL` before the first `IDAT`, short-circuits on `IDAT`, and uses `int64` arithmetic to avoid `uint32`+offset overflow — preventing silent animation loss from `image/png` decoding only the first frame. The check runs before slot acquisition, so a validate-only image never consumes a compression slot.
- **Fail-open is total.** decode/encode error → `failed` → original bytes; panic → `recover` → `failed`; saturation/timeout/disabled/format → `skipped` → original bytes. Compression can only reject via the explicit `over_limit` path (which is the spec'd behavior), never destabilize the upload.
- **Pointer discipline is correct** across the magic-number read, `DecodeConfig`, `io.ReadAll`, sha512, and `UploadFile` — each stage seeks to 0 before consuming, and the `bytes.Reader` swap after compression is re-seekable.
- **`stickerUploadExtForRequest`** keeps the GET-side preflight extension consistent with the POST-side allowlist, so a narrowed config doesn't hand clients an extension that upload will later reject.

### Findings (all P2 / advisory — none blocking)

**[P2] Snapshot atomicity is per-getter, not per-map-load.** `stickerLimits()` (`sticker_compress.go`) calls the `SystemSettings` getters one at a time, and each getter does its own `snapshot.Load()`. If the 60 s reload swaps the atomic map *between* two getter calls within a single `stickerLimits()` invocation, the resulting snapshot struct can mix values from two config generations (e.g. `maxDim` from gen N, `compressTargetKB` from gen N+1). The window is microseconds once per 60 s and every individual value is still independently clamped/valid, so the worst case is one upload validated against a briefly-mixed but each-individually-safe set — not a security or data-loss issue. If you want the invariant the comments claim ("one request, one snapshot") to be literally true, add a `common`-layer helper that does a single `snapshot.Load()` and reads all sticker keys from that one map pointer. Not required for merge.

**[P2] `Compress()` re-reads live `StickerCompressEnabled()`.** `uploadFile` already gates on the captured `stickerLimits.compressEnabled` before calling `Compress`, but `Compress` re-checks `c.settings.StickerCompressEnabled()` (`sticker_compress.go:209`). If a reload flips the switch off in that window, `Compress` returns `skipped:disabled` and the original bytes are stored for a request whose snapshot had compression enabled. This is fail-open and harmless (original valid bytes stored), but it contradicts the "Compress uses only request-local params" intent documented right above the function. Consider threading `Enabled` through `stickerCompressParams` and dropping the live read, keeping the disabled-unit-test dimension via the param. Cleanup, not a blocker.

---

## 3. Overall verdict

**APPROVED**

Spec is fully met and there are no P0/P1 issues. The two P2 findings are fail-open, sub-millisecond-window consistency nits that do not affect correctness, security, or data integrity — safe to address in a follow-up or leave as-is.

### Reviewer notes
- The compression path defaults OFF and the PR correctly gates production enablement on the `perf-todo.md` checklist (bench on x86, HTTP latency compare, goroutine-leak bound, pprof, decompression-bomb sanity, graded ramp). Recommend that checklist be completed before flipping `sticker.compress_enabled=true` in production — the review approves the code, not the rollout.
- A second automated cross-family reviewer leg was attempted for this complex PR but timed out and did not return findings; it is recorded here as absent (not silent agreement). The findings above are from an independent line-level pass plus one adversarial reviewer leg.

PR #547, opened 2026-07-07
feat(sticker): expose upload limits via appconfig

an9xyz:
## Summary

Follow-up to #544. That PR made sticker upload limits (max size / max dimension / allowed formats) operator-tunable but only enforced them on the server. Clients still hardcode ~1 MB / 512 px / 5 formats, and the moment ops narrows or widens any knob, the client's local pre-check drifts from the server's authoritative gate — users pay for the upload before hearing "too big / wrong format".

Expose the three limits via the existing `GET /v1/common/appconfig` channel (same fabric as `sticker_handle_required` / `docs_on`) so clients can pre-check locally after file selection and show accurate `max X MB / Y px / gif,png,…` hints. This is a UX optimisation; the server-side `stickerLimits` snapshot in `modules/file` remains the security boundary and is unchanged.

## Wire shape

```json
{
  "sticker_upload_limits": {
    "max_size_kb": 1024,
    "max_dimension": 512,
    "allowed_formats": [".gif", ".png", ".jpg", ".jpeg", ".webp"]
  }
}
```

The three fields are always emitted together, always consumed together (single pre-check pass), and always change together (ops adjusts the "limits" set as one unit), so they live under one namespaced object rather than three flat fields. Existing flat `sticker_` fields (`sticker_custom_enabled`, `sticker_handle_required`) stay flat — unrelated booleans consumed by different client code paths; moving them would be a breaking change requiring cross-client coordination.

Both handler branches (version-shortcut and full-refresh) compose the sub-response through a single `buildStickerUploadLimitsResp` helper so the two branches can't drift.

## Not surfaced

Compression knobs (`compress_enabled` / `compress_target_kb` / `compress_max_concurrency` / `compress_timeout_ms`) are intentionally omitted. They are server-side shadow parameters — response shape doesn't change with them, no corresponding client behaviour, and exposing them would only leak implementation detail.

## Client contract compatibility

- Additive: new JSON field on a widely-cached response; older clients that ignore unknown fields see zero regression.
- Defaults (unset in `system_setting`) match the pre-#544 hardcoded behaviour exactly (1024 KB / 512 px / all 5 raster formats), so clients that never opt in to the new field remain byte-for-byte compatible.
- Same version-decoupling as `sticker_handle_required` / `docs_on`: served on the version-shortcut branch too, so an old client that cache-hits on `version` still sees the current limits after ops flips them.

## Test plan

Three new integration tests in `modules/common/api_test.go`, mirroring the `TestGetAppConfig_StickerHandleRequired_*` pattern:

- [x] `TestGetAppConfig_StickerUploadLimits_DefaultsMatchHistoricalHardcoded` — no `system_setting` rows → defaults match the pre-#544 hardcoded values.
- [x] `TestGetAppConfig_StickerUploadLimits_DBOverrideServed` — ops writes narrower values → appconfig serves them; verifies CSV normalisation (lowercase, dot-prefix, whitespace strip, non-raster drop, input-CSV order preserved) reaches the wire.
- [x] `TestGetAppConfig_StickerUploadLimits_OnVersionShortCircuit` — same values served on the `?version=…` short-circuit branch.

Local: `go test -race -count=1 ./modules/common/` — green after drop&recreate of the test DB.

## Follow-ups (not in scope)

- Client-side consumption (Android / iOS / Web) — separate work in the client repos.
- Rewriting the flat `sticker_custom_enabled` / `sticker_handle_required` into a `sticker` namespace — breaking change, needs cross-client coordination and would be its own PR.

### DIFF
--- a/modules/common/api.go
+++ b/modules/common/api.go
@@ -390,6 +390,9 @@ func (cn *Common) appConfig(c *wkhttp.Context) {
 			StickerCustomEnabled:   cn.systemSettings.StickerCustomEnabled(),
 			StickerHandleRequired:  cn.systemSettings.StickerHandleRequired(),
 			DocsOn:                 cn.systemSettings.DocsEnabled(),
+			// Sticker 上限:短路分支同样下发,让老客户端在管理台放宽/收窄后
+			// 也能立刻拿到最新值。
+			StickerUploadLimits: buildStickerUploadLimitsResp(cn.systemSettings),
 		})
 		return
 	}
@@ -433,7 +436,21 @@ func (cn *Common) appConfig(c *wkhttp.Context) {
 		StickerCustomEnabled:   cn.systemSettings.StickerCustomEnabled(),
 		StickerHandleRequired:  cn.systemSettings.StickerHandleRequired(),
 		DocsOn:                 cn.systemSettings.DocsEnabled(),
+		// Sticker 上限:客户端本地预校验用,兜底仍在服务端 modules/file 侧。
+		StickerUploadLimits: buildStickerUploadLimitsResp(cn.systemSettings),
 	})
+}
+
+// buildStickerUploadLimitsResp 从 SystemSettings 派生一份 stickerUploadLimitsResp
+// 快照。两个 handler 分支(version-shortcut / full-refresh)都用它,避免字段
+// 组装漂移。SystemSettings.StickerUpload* getter 各自内部已做 hard-cap clamp,
+// 拿到的都是安全值。
+func buildStickerUploadLimitsResp(s *SystemSettings) stickerUploadLimitsResp {
+	return stickerUploadLimitsResp{
+		MaxSizeKB:      s.StickerUploadMaxSizeKB(),
+		MaxDimension:   s.StickerUploadMaxDimension(),
+		AllowedFormats: s.StickerUploadAllowedFormats(),
+	}
 }
 
 // oidcProviders 返回 OIDC provider 元数据数组,让前端不再硬编码 provider id/name/authorize_path。
@@ -791,6 +808,44 @@ type appConfigResp struct {
 	// 与 app_config.version 解耦的原因同 LocalLoginOff / SearchEnabled：运维切展示
 	// 策略后老客户端命中 version 短路分支也必须拿到最新值，故两个分支都下发。
 	DocsOn bool `json:"docs_on"`
+
+	// StickerUploadLimits 是自定义贴纸上传的操作端可调上限，与 sticker.upload_*
+	// system_setting 同源（SystemSettings.StickerUpload{MaxSizeKB,MaxDimension,
+	// AllowedFormats}）。用途：让客户端在选图后本地预校验（提示"最大 X MB / 最长
+	// Y px / 支持 gif/png..."），避免大图/非法扩展名先跑完 HTTP 上传才被服务端拒
+	// —— 尤其对移动端流量友好。
+	//
+	// 客户端预校验只是 UX 优化；服务端 modules/file 侧仍对每个 sticker upload
+	// 请求做同一份 stickerLimits 快照兜底（size/format/dimension 三层），客户端
+	// 缓存过期或被绕过都不影响安全边界。默认值与 PR #544 之前的历史硬编码严格
+	// 等价（1024 KB / 512 px / [.gif,.png,.jpg,.jpeg,.webp]），运营不动 knob
+	// 时行为完全无回归。
+	//
+	// 与 app_config.version 解耦的原因同 StickerHandleRequired / DocsOn：运维在
+	// 管理台放宽/收窄上限后，老客户端命中 version 短路分支也必须拿到最新值，否则
+	// 被本地缓存住失去实时性，故两个分支都下发。
+	//
+	// 独立命名空间（不叫 sticker_upload_*）的原因：三个字段是一组"上限"，一起
+	// 生效、一起提示；单独包一层让客户端解析和展示逻辑天然聚合，未来若加非上限
+	// 类的 sticker upload 字段（例如上传行为 toggle）也不会污染这个 object。
+	// 现有的 sticker_custom_enabled / sticker_handle_required 是平铺 bool，形态
+	// 与 limits object 差异明显，风格分裂可接受；语义相近的字段（同为 upload
+	// 上限）优先聚合，与 OIDCProviders 这类 nested resp 同思路。
+	//
+	// 不下发压缩相关 knob（compress_enabled / compress_target_kb /
+	// compress_max_concurrency / compress_timeout_ms）—— 这些是服务端"影子"参数，
+	// 响应字段结构不变、无对应客户端行为，曝光只会泄露实现细节。
+	StickerUploadLimits stickerUploadLimitsResp `json:"sticker_upload_limits"`
+}
+
+// stickerUploadLimitsResp 是 appConfigResp.sticker_upload_limits 的形状。
+// 三个字段一起下发,一起用于客户端预校验。字段名不再带 sticker_upload_ 前缀
+// —— 已经在命名空间下。allowed_formats 内每项含前导点、小写、按输入 CSV
+// 顺序保留(见 SystemSettings.StickerUploadAllowedFormats 契约)。
+type stickerUploadLimitsResp struct {
+	MaxSizeKB      int      `json:"max_size_kb"`
+	MaxDimension   int      `json:"max_dimension"`
+	AllowedFormats []string `json:"allowed_formats"`
 }
 
 type oidcProviderResp struct {
--- a/modules/common/api_test.go
+++ b/modules/common/api_test.go
@@ -806,3 +806,91 @@ func TestGetAppConfig_DocsOn_OnVersionShortCircuit(t *testing.T) {
 	assert.Equal(t, http.StatusOK, w.Code)
 	assert.Contains(t, w.Body.String(), `"docs_on":true`)
 }
+
+// setStickerUploadLimitsSettings upserts the three sticker upload knobs
+// (size KB / max dim / allowed formats CSV) and reloads the shared snapshot.
+// Passing "" for any of them skips writing that row so tests can exercise the
+// "default when unset" branch selectively. Call AFTER cleanAllTablesAndReloadSettings.
+func setStickerUploadLimitsSettings(t *testing.T, ctx *config.Context, sizeKB, maxDim, allowedCSV string) {
+	t.Helper()
+	if sizeKB != "" {
+		_, err := ctx.DB().InsertInto("system_setting").
+			Columns("category", "key_name", "value", "value_type").
+			Values("sticker", "upload_max_size_kb", sizeKB, "int").Exec()
+		require.NoError(t, err)
+	}
+	if maxDim != "" {
+		_, err := ctx.DB().InsertInto("system_setting").
+			Columns("category", "key_name", "value", "value_type").
+			Values("sticker", "upload_max_dimension", maxDim, "int").Exec()
+		require.NoError(t, err)
+	}
+	if allowedCSV != "" {
+		_, err := ctx.DB().InsertInto("system_setting").
+			Columns("category", "key_name", "value", "value_type").
+			Values("sticker", "upload_allowed_formats", allowedCSV, "string").Exec()
+		require.NoError(t, err)
+	}
+	require.NoError(t, EnsureSystemSettings(ctx).Reload())
+}
+
+// appconfig 必须下发 sticker_upload_* 上限。未配置时 = SystemSettings 侧默认
+// (1024 KB / 512 px / 全 5 位图),与 PR #544 之前的历史硬编码严格等价。客户端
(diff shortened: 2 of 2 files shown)


OctoBoooot (approved):
## Review: expose sticker upload limits via appconfig (#547)

**Verdict:** Approve — clean, small, read-only feature: exposes three already-clamped sticker upload limits via `appconfig` for client-side pre-validation. No findings. (Code CI still pending — confirm green before merge.)

Reviewed as part of the sticker-upload family (#544 established the `stickerClampIntUpper` hard-cap discipline this builds on).

### Verified
- **Values are safe by construction.** `buildStickerUploadLimitsResp` (`api.go`) reads through `StickerUploadMaxSizeKB/MaxDimension/AllowedFormats`, each of which applies the `stickerClampIntUpper` hard-cap clamp (verified on #544) — so no unclamped/hostile setting value can reach the client.
- **Both handler branches emit the field.** The shared builder is called in both the version-shortcut and full-refresh branches (`api.go:393`, `:439`) — so a stale client hitting the version shortcut still gets fresh limits, matching the documented `DocsOn`/`StickerHandleRequired` decoupling. This is the classic "two branches must both emit" bug spot, and it's handled + tested.
- **Client pre-check is UX-only; server remains the boundary.** The doc-comment is explicit that `modules/file` still enforces size/format/dimension server-side on every upload — the exposed limits are a mobile-friendly pre-validation hint, not a security control. Correct framing.
- **Tests exemplary (coverage 1.0).** `DefaultsMatchHistoricalHardcoded` (unset → 1024/512/5-formats, no regression vs pre-#544 hardcode), `DBOverrideServed` (with CSV normalization: case/dot/dedup/non-raster-drop/input-order), and `OnVersionShortCircuit` (`?version=99999999` → fresh values). The version-shortcut coverage is exactly the branch most likely to silently miss a new field.

### Praise
- Routing both branches through one `buildStickerUploadLimitsResp` (rather than assembling the field twice) prevents the two-branch drift this repo has documented pain with — and the version-shortcut test pins it. The decision-record comments (why nested namespace, why compress knobs are NOT exposed as they'd leak implementation detail) are unusually thorough for a 143-line PR.

### Notes
- `reviewDecision: REVIEW_REQUIRED` / `mergeStateStatus: BLOCKED`; `check-sprint` + secret-scan green, code CI (Build/Test/Vet/Lint) pending — confirm green before merge. Approved on the merits.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #547 (octo-server)

## Summary

This PR exposes the three sticker upload limits (max size KB, max dimension, allowed formats) — already operator-tunable since PR #544 — via the existing `GET /v1/common/appconfig` endpoint so clients can pre-check locally before upload. The change is 2 files / 143 insertions: a nested `sticker_upload_limits` object on the response struct, a shared `buildStickerUploadLimitsResp` helper wired into both handler branches, and 3 integration tests covering default / DB-override / version-shortcut paths.

The design is clean: a single helper prevents branch drift, the nested namespace groups semantically related fields, compression internals are intentionally withheld, and the field is always emitted (no `omitempty`) so clients always see a complete limits snapshot.

## Verification

- ✅ **Data flow — int getters** — `StickerUploadMaxSizeKB()` and `StickerUploadMaxDimension()` both route through `stickerClampIntUpper`, which clamps to [1, hardCap]. Out-of-range or non-numeric values fall back to defaults (1024 KB / 512 px). No path returns 0 or negative.
- ✅ **Data flow — formats getter** — `StickerUploadAllowedFormats()` intersects DB CSV with `stickerUploadRasterAllowlist`, normalizes (lowercase, leading dot, dedup), and returns the full default set when intersection is empty. No path returns nil or empty slice.
- ✅ **Branch parity** — Both the version-shortcut branch (`modules/common/api.go:396`) and the full-refresh branch (`:441`) call the same `buildStickerUploadLimitsResp` helper, so the two response shapes cannot drift.
- ✅ **Auth boundary unchanged** — The endpoint still requires `token` header authentication; no new endpoint or route is added.
- ✅ **Test coverage** — Three new tests mirror the `TestGetAppConfig_StickerHandleRequired_*` pattern: defaults (no DB rows), DB override with CSV normalization (including non-raster `mp4` rejection), and version-shortcut branch. All assert the exact nested JSON substring.
- ✅ **Backward compatibility** — Old clients that ignore `sticker_upload_limits` continue to work; the field is additive and the server-side enforcement in `modules/file` remains authoritative.

Static analysis only at head `2e3a7014f873c86db221e659f94dda411e77a633`; build and tests not executed in this environment.

## Findings

### Blind-spot Checklist

- **C1 — Dual-path parity**: Both handler branches (version-shortcut / full-refresh) use the shared `buildStickerUploadLimitsResp` helper. Clear.
- **C2 — Control-flow ordering**: Single helper call per branch, no nested reuse or ordering concern. N/A.
- **C3 — Authorization boundary**: No new endpoint. The config data is informational (UX pre-check); server-side enforcement remains in `modules/file`. N/A.
- **C4 — Authorization lifecycle**: No container-member state involved; global config only. N/A.
- **C5 — Build/runtime correctness**: Additive JSON field with no `omitempty`; old clients ignore it, new clients always see it. No build-time or runtime risk. Clear.
- **C6 — Governance/docs self-consistency**: No governance or policy documents changed. N/A.

### Data Flow Trace

| Consumed data | Source | Runtime safe? |
|---|---|---|
| `MaxSizeKB` | `SystemSettings.StickerUploadMaxSizeKB()` → `stickerClampIntUpper` → [1, 5120] | ✅ Always positive int |
| `MaxDimension` | `SystemSettings.StickerUploadMaxDimension()` → `stickerClampIntUpper` → [1, 1024] | ✅ Always positive int |
| `AllowedFormats` | `SystemSettings.StickerUploadAllowedFormats()` → intersect with allowlist, fallback to full default set | ✅ Always non-empty []string |

### Cross-round Blocker Recheck (R6)

## Verdict: APPROVED

Well-structured, minimal change that correctly exposes operator-tunable sticker upload limits to clients for local pre-check. Data flow is safe (all getters return clamped/sanitized values with sensible defaults), branch parity is enforced via a shared helper, test coverage is adequate, and the security boundary (server-side enforcement) remains untouched.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; clean additive change with safe data flow, branch parity, and adequate test coverage.

yujiawei (approved):
# Code Review — PR #547 (octo-server)

Independent review of `feat(sticker): expose upload limits via appconfig for client-side pre-check` at head `2e3a7014`. This PR was flagged security-sensitive, so the enforcement boundary got extra scrutiny.

## Summary

The change adds a `sticker_upload_limits` object (`max_size_kb`, `max_dimension`, `allowed_formats`) to the `/v1/common/appconfig` response so clients can pre-validate a sticker before uploading. A shared helper `buildStickerUploadLimitsResp` is emitted from **both** handler branches (version-shortcut and full-refresh), reading the same `SystemSettings.StickerUpload*` getters. Well-structured, well-documented, and consistent with the existing `StickerHandleRequired` / `DocsOn` fields.

## 1. Spec compliance

**Spec: ✅**

- Missing: none. Both handler branches emit the field (`api.go:395`, `api.go:440`); defaults match the historical hardcoded values (1024 KB / 512 px / `[.gif,.png,.jpg,.jpeg,.webp]`), and tests assert this exact equivalence (`api_test.go` `TestGetAppConfig_StickerUploadLimits_DefaultsMatchHistoricalHardcoded`).
- Over-build: none. Compression shadow params (`compress_enabled`/`target_kb`/etc.) are deliberately **not** exposed — good call, they carry no client behavior and would only leak implementation detail.
- Deviation: none. The single-source helper avoids field-assembly drift between the two branches.

## 2. Code quality

**Quality: Approved**

Security boundary (the crux for this PR) holds up:

- `/v1/common/appconfig` is a no-auth endpoint (`api.go:85`), but the exposed data is non-sensitive UX hints. No secret/token/credential leakage.
- Server-side enforcement is authoritative and independent of what the client sees. Multipart sticker upload snapshots the same getters (`sticker_compress.go:90`) and enforces **size** (`api.go:286`), **format** (`api.go:328`), and **decoded pixel dimensions** (`api.go:426`). A client that ignores or tampers with the advertised limits still gets rejected server-side.
- No client/server divergence bypass: the presigned direct-upload path explicitly **rejects** `type=sticker` (`api.go:770`), forcing stickers through the validated multipart path. This closes the "sign an oversized/non-image object, then register it as a sticker URL" gap.
- The getters clamp numerics to hard caps and intersect `allowed_formats` down to the raster allowlist (`system_settings.go:850`, `system_settings.go:899`), and an all-illegal config falls back to the full default set rather than dark-closing the feature — so the response can never advertise a widened or empty-illegal limit.

Test coverage is solid: defaults, DB override with CSV normalization, and the version-shortcut branch are all covered.

### Non-blocking (P2 / nits)

- **P2 — test helper docstring mismatch** (`api_test.go:810`): `setStickerUploadLimitsSettings` is documented as "upserts" but performs a plain `InsertInto(...).Exec()`. It's safe today because every caller runs `cleanAllTablesAndReloadSettings` first, but if reused without a prior wipe it will hit a unique-key violation. Either correct the docstring to say "inserts (call after a table wipe)" or make it a real upsert.
- **P2 — grouped-getter atomicity** (`api.go:446`): the three getters are read sequentially, so a concurrent `Reload()` could in theory produce a torn snapshot (old size + new formats). This is UX-only (server-side enforcement is authoritative) and matches the existing pattern every other appconfig field already uses, so it is not a regression. If you ever want a strict atomic group, a single `SystemSettings` method returning the whole struct would do it. Not required for this PR.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. No P0/P1 findings. The two P2 items are optional cleanups and do not block merge.

## 4. For a human to double-check (security-sensitive)

Nothing that blocks merge, but worth an eyeball: confirm the product intent is that these limits are fine to expose on an unauthenticated endpoint (they are operational, not secret). The enforcement side is correct regardless.

Jerry-Xin (approved):
Additive, well-scoped follow-up to #544: exposes the three sticker upload *limits* via the existing authenticated `GET /v1/common/appconfig` channel for client-side pre-check. Byte-verified against head `2e3a7014` — no leak, values are single-sourced with the enforcement path, tests assert the exact wire shape. APPROVE.

## 🔴 Blocking
None.

## 🟡 Non-blocking
- `modules/common/system_settings.go:808-810` documents `stickerUploadRasterAllowlist` as a verbatim mirror of `modules/file/const.go:stickerUploadExts` (kept duplicated to avoid an import cycle). The two are consistent today (`.gif/.png/.jpg/.jpeg/.webp`), but the coupling is comment-enforced only. If someone later widens `stickerUploadExts` without touching the common-side allowlist, appconfig would advertise a narrower set than upload accepts. Not introduced by this PR and low risk, but a shared test asserting the two lists are equal would make the "keep in sync" contract enforceable rather than aspirational.

## ✅ Highlights
- **No config leak.** Only limit values are surfaced (`max_size_kb` / `max_dimension` / `allowed_formats`). Compression shadow knobs (`compress_enabled` / `compress_target_kb` / `compress_max_concurrency` / `compress_timeout_ms`) are intentionally omitted — verified they are not referenced in `buildStickerUploadLimitsResp` (`modules/common/api.go:820-826`). No arbitrary appconfig-key exposure.
- **Read-side clamped, not raw.** `buildStickerUploadLimitsResp` (`api.go:820`) reads `s.StickerUploadMaxSizeKB()` / `StickerUploadMaxDimension()` / `StickerUploadAllowedFormats()`, all of which apply the hard-cap clamp (`stickerClampIntUpper`, `system_settings.go:850-865`) and the raster-allowlist intersection (`StickerUploadAllowedFormats`, `system_settings.go:899-935`). A hand-edited out-of-range DB row is clamped/normalized before it ever reaches the wire.
- **Exact consistency with the enforcement path.** The `modules/file` security boundary (`stickerLimits()`, `sticker_compress.go:73-104`) is built from the *same three getters* — `maxSize = StickerUploadMaxSizeKB()*1024`, `maxDim = StickerUploadMaxDimension()`, `allowedFormats` from `StickerUploadAllowedFormats()`. The API reports KB while enforcement works in bytes (documented `max_size_kb`), so the advertised limits cannot drift from what upload actually enforces.
- **Both handler branches single-sourced.** Version-shortcut (`api.go:393`) and full-refresh (`api.go:439`) both compose through the one `buildStickerUploadLimitsResp` helper, so the two branches can't diverge — old clients hitting the `?version=` short-circuit still see current limits after ops flips them.
- **Endpoint gating & input.** Same authenticated channel as existing `sticker_handle_required` / `docs_on`; tests exercise it with a token header. The only request input is the `?version=` cache key — no new injectable surface.
- **Tests are substantive, not shallow.** All three (`api_test.go`) assert the full serialized JSON via exact substring match: defaults (`{"max_size_kb":1024,"max_dimension":512,"allowed_formats":[".gif",".png",".jpg",".jpeg",".webp"]}`), DB override + CSV normalization (lowercase / dot-prefix / whitespace-strip / non-raster `mp4` dropped → `[".png",".gif"]`), and the version-shortcut branch. They pin field order and normalization behaviour, not just presence.

CI: Build / Vet / Lint / Test / check-sprint all green at head.

PR #548, opened 2026-07-08
feat(cardmsg): P2 card interaction loop — action dispatch, idempotency, type-17 edit + card_seq CAS

an9xyz:
## Summary

PR-B of the card message P2 protocol — the **interaction closed loop**. A user taps a button (or fills inputs and submits) on a bot-sent card, the owning bot receives a `card_action` event, rewrites the card via the existing `botMessageEdit` path, and all clients converge via the existing `/extra/sync` CMD channel. Builds on the merged P1 (`pkg/cardmsg`, #543). Zero octo-im changes.

Scope = `card-message-interaction` decisions **D3–D9 / D11** + the **octo/v2 whitelist** (D1/D2). D10 card revision history and D12 capability manifest are split into sibling PRs C/D.

## Related Issue

Self-sourced octospec task (no tracking issue). Contract: `.octospec/tasks/card-message-interaction/brief.md`.

## Linked Spec

`.octospec/tasks/card-message-p2-action-loop/brief.md` (execution brief) — implements the `card-message-interaction` sibling contract.

## Changes

- **`pkg/cardmsg`**: octo/v2 whitelist filled into the P1 `validate.go` seams (`Action.Submit` + `Input.Text`/`Toggle`/`ChoiceSet`, frame-unique ids per D1, `Action.Submit.data` must be an object); new `interactive.go` (`SubmitAction` extracts the matched action's **static `data` from the effective frame** — never from the request, D11 anti-forgery — plus `NormalizeContentEdit`, `CardSeq`) and `inputs.go` (`ValidateInputs`, D11 declared-id/type/size trust boundary).
- **`POST /v1/message/card/action`** (`modules/message/api_card_action.go` + `card_action_claims.go`): `AuthMiddleware` → `SharedUIDRateLimiter` → Space (existing `/v1/message` group order) + a 64 KiB `MaxBytesReader`. Validation order (D3): stored-message lookup + channel binding (anti-IDOR — binding derives from the stored row) → membership of the stored channel (`ExistMemberActive`) → bot sender (`ExistRobot`, `iwh_`/human fail closed, D7) → `action_id` in the effective frame → D11 inputs → D4 idempotent enqueue. D4 dedup key = `(message_id, action_id, operator_uid)`; ordering claim→enqueue→confirm with compensating release.
[...]

PR #549, opened 2026-07-08
feat(cardmsg): P2 card revision history (D10) — side table, query API, clear + tombstone

an9xyz:
> ✅ **#548 (PR-B) is merged into `main`.** This PR is now rebased onto `main` and the diff below is the **D10-only delta** (3 commits: brief → impl → octospec finish). The rebase resolved the squash-merge conflicts by (1) keeping PR-B's merged hardening — the group-status + thread-status gates were folded into the shared `authorizeCardChannelMember` helper so no #548 security fix regresses — and (2) tagging the new revisions test `//go:build integration` to match the sibling e2e files.

## Summary

PR-C of the card message P2 protocol — **D10 card revision history**. The D6 full-frame-replacement model keeps `message_extra.content_edit` latest-frame-only, so the message table cannot answer "这张卡以前是什么". PR-C adds a side table recording each non-transient frame as a complete renderable envelope, plus a member-visible query API. Zero octo-im changes. D12 capability manifest is sibling PR-D.

## Related Issue

Self-sourced octospec task. Contract: `.octospec/tasks/card-message-interaction/brief.md` (D10). Built on #548 (now merged).

## Linked Spec

`.octospec/tasks/card-message-p2-revision-history/brief.md`

## Changes (PR-C delta)

- **`pkg/cardrevision`** (new): shared store over `octo_message_card_revision` (written by `bot_api` on card edits/clear, read by `message` — same split as `message_extra`). `AppendFrame` (+ cap-20 prune, rune-safe `plain` truncate), `Query` (newest-first, incl. tombstones), `Clear` (tx: delete frames + write an auditable tombstone), `DeleteByMessageID`.
- **Migration**: `octo_message_card_revision` (`octo_` prefix per repo rule; the brief's illustrative `message_card_revision` overridden).
- **`pkg/cardmsg`**: `Transient` / `TransientFromContentEdit` / `PlainFromContentEdit`.
- **`modules/bot_api`**: `botMessageEdit` appends a non-transient frame **after** the content write (best-effort — `content_edit` is authoritative, append failure never blocks the edit; `transient:true` frames skip history); new `POST /v1/bot/message/card/revisions/clear` (owner-checked, writes a tombstone).
- **`modules/message`**: `GET /v1/message/card/revisions` (summary / `full=1`), reusing the extracted `authorizeCardChannelMember` gate shared with `card/action`; revoke deletes a card's revisions (after commit, before notify); an `isCardMessageWithdrawn` visibility fallback hides history for revoked / globally-deleted / operator-locally-deleted cards even if rows remain.
- errcodes (revision invalid/denied, bot clear forbidden) + zh-CN + en markers + guard lists.

## Testing

Local MySQL + Redis + WuKongIM containers.

- `modules/message`: `TestCardRevisionsQuery` (summary / `full=1` re-validates via `cardmsg.Validate` / tombstone rendering), `TestCardRevisionsMemberGate` (non-member denied / cross-channel IDOR / person-points-at-group invalid), `TestCardRevisionsCapAndDelete` (cap-20 eviction + delete), `TestCardRevisionsWithdrawnHidden` (revoke / is_deleted / user-local-delete all return empty). `card/action` suite unchanged (no regression).
- `modules/bot_api` (`-race`): `TestBotCardRevisionsIM` (non-transient append / transient skip / clear → tombstone). D6/D9 suite unchanged.
- `go build ./...`, `golangci-lint run` 0 issues, i18n suite green.

- [x] Manually verified

## COMPREHENSION

   Adds a second consumer of the card-message trust surface: a member-visible **read** endpoint for revision history, and a bot-only **clear** write. Both go through the same anti-IDOR channel-binding + membership gate as `card/action` (extracted into `authorizeCardChannelMember`). Card edits now also append to a side table on the `botMessageEdit` path. The revoke path gains a revision-cleanup step.

   (a) The append rides `botMessageEdit` — it is best-effort *after* the `content_edit` write, so an append failure logs and continues (the card edit and its CMD fanout are unaffected). (b) The new table is written by `bot_api` and read by `message` via a shared `pkg/cardrevision.Store`; a migration adds `octo_message_card_revision`. (c) Revocation now deletes revision rows after the revoke commit; if that delete fails, the query endpoint's `isCardMessageWithdrawn` gate still hides the history — two independent layers. (d) The extracted member gate is now shared by two endpoints; its behavior is pinned by both `card/action` and revisions tests.

   `TestCardRevisionsWithdrawnHidden` seeds revision rows then marks the message revoked / is_deleted / user-local-deleted and asserts the GET returns an empty list and does not leak content — covering the verify-phase P1. `TestCardRevisionsMemberGate` asserts non-member → 403-class and cross-channel IDOR rejection (same stored-row binding as `card/action`). `TestBotCardRevisionsIM` (real WuKongIM) asserts a non-transient edit appends exactly one frame, a `transient:true` edit appends none, and the clear endpoint deletes frames + writes one tombstone. `TestCardRevisionsCapAndDelete` pins cap-20 eviction.

## Checklist

- [x] Updated documentation (docs/card-protocol.md already carries the D10 contract; no drift)

### DIFF
--- a/modules/message/api_card_revisions.go
+++ b/modules/message/api_card_revisions.go
@@ -1,0 +1,114 @@
+package message
+
+// card-message-interaction P2 D10.5：卡片修订历史查询。
+// GET /v1/message/card/revisions?message_id=&channel_id=&channel_type=&limit=&full=
+// —— AuthMiddleware → SharedUIDRateLimiter → Space（挂在 /v1/message 组）+ 与
+// card/action 同一门禁栈（authorizeCardChannelMember：anti-IDOR 频道绑定 + 存储
+// 频道成员资格）+ 生命周期兜底（isCardMessageWithdrawn）+ canonical 可见性
+// （cardCanonicalVisibleToViewer：visibles / 过期 / 偏移）—— 可见性 ≥ card/action、
+// 与单条读 respondSingleMessage 同口径（PR#549 review B1：修订历史披露多于单次动作，
+// 门禁不得更宽）。默认返回摘要列表；full=1 每条非墓碑行附完整可渲染帧信封。墓碑行
+// （清除审计）也出现在列表里。
+
+import (
+	"encoding/json"
+	"strconv"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
+	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
+	"go.uber.org/zap"
+)
+
+// getCardRevisions handles GET /v1/message/card/revisions.
+func (m *Message) getCardRevisions(c *wkhttp.Context) {
+	loginUID := c.GetLoginUID()
+	messageID := c.Query("message_id")
+	channelID := c.Query("channel_id")
+	if messageID == "" || channelID == "" {
+		respondMessageRequestInvalid(c, "message_id")
+		return
+	}
+	channelTypeI, err := strconv.ParseUint(c.Query("channel_type"), 10, 8)
+	if err != nil || channelTypeI == 0 {
+		respondMessageRequestInvalid(c, "channel_type")
+		return
+	}
+	limit := 0
+	if s := c.Query("limit"); s != "" {
+		if v, perr := strconv.Atoi(s); perr == nil {
+			limit = v
+		}
+	}
+	full := c.Query("full") == "1"
+
+	// D10.5 门禁：与 card/action 同口径（anti-IDOR 绑定 + 存储频道成员）。非卡片/
+	// 不存在/频道不匹配 → invalid；非成员 → denied（防枚举）。
+	msgM, handled := m.authorizeCardChannelMember(c, loginUID, messageID, channelID, uint8(channelTypeI),
+		errcode.ErrMessageCardRevisionInvalid, errcode.ErrMessageCardRevisionDenied)
+	if handled {
+		return
+	}
+
+	// 撤回/删除可见性兜底（PR-C verify P1）：已撤回 / 全局删除 / 操作者本地删除的卡片
+	// 无可查询内容历史（D10「revoked message leaves no queryable content history」）。
+	// 与 card/action 同口径,且不依赖 revoke 时 best-effort 删除是否成功 —— 双层防御:
+	// 即便修订行仍在（删除失败 / is_deleted / user-local-delete 本就不触发删除），
+	// 查询也返回空列表。
+	if withdrawn, werr := m.isCardMessageWithdrawn(messageID, loginUID); werr != nil {
+		m.Error("查询卡片撤回/删除状态失败", zap.Error(werr), zap.String("messageID", messageID))
+		httperr.ResponseErrorL(c, errcode.ErrMessageQueryFailed, nil, nil)
+		return
+	} else if withdrawn {
+		c.Response(map[string]interface{}{"revisions": []map[string]interface{}{}})
+		return
+	}
+
+	// canonical 可见性对齐（PR#549 review B1）：修订历史披露的内容比单次动作触发更多，
+	// 门禁必须 ≥ card/action。成员+生命周期之外补第四类可见性 —— visibles 白名单 /
+	// 消息过期 / 用户清理偏移 / 频道偏移（cardCanonicalVisibleToViewer，与 card/action
+	// 共用）。被 visibles 排除 / 偏移截断 / 已过期的成员看不到卡片本身，也就看不到它的
+	// 历史帧：不可见 → 空列表（与 withdrawn 同处理，不泄漏存在性）；查询失败 fail-closed。
+	if visible, verr := m.cardCanonicalVisibleToViewer(msgM, loginUID); verr != nil {
+		m.Error("查询卡片修订可见性失败", zap.Error(verr), zap.String("messageID", messageID))
+		httperr.ResponseErrorL(c, errcode.ErrMessageQueryFailed, nil, nil)
+		return
+	} else if !visible {
+		c.Response(map[string]interface{}{"revisions": []map[string]interface{}{}})
+		return
+	}
+
+	revs, err := m.cardRevisions.Query(messageID, limit)
+	if err != nil {
+		m.Error("查询卡片修订历史失败", zap.Error(err), zap.String("messageID", messageID))
+		httperr.ResponseErrorL(c, errcode.ErrMessageQueryFailed, nil, nil)
+		return
+	}
+
+	items := make([]map[string]interface{}, 0, len(revs))
+	for _, r := range revs {
+		if r.IsTombstone == 1 {
+			items = append(items, map[string]interface{}{
+				"tombstone":  true,
+				"cleared":    r.ClearedCount,
+				"editor_uid": r.EditorUID,
+				"edited_at":  r.EditedAt,
+			})
+			continue
+		}
+		item := map[string]interface{}{
+			"plain":      r.Plain,
+			"editor_uid": r.EditorUID,
+			"edited_at":  r.EditedAt,
+		}
+		if r.CardSeq.Valid {
+			item["card_seq"] = r.CardSeq.Int64
+		}
+		if full && r.Content.Valid && r.Content.String != "" {
+			// 完整帧信封逐字返回（写入时已过 cardmsg.Validate/Finalize）。
+			item["card"] = json.RawMessage(r.Content.String)
+		}
+		items = append(items, item)
+	}
+	c.Response(map[string]interface{}{"revisions": items})
+}
--- a/.octospec/journal/shared/card-message-p2-revision-history.md
+++ b/.octospec/journal/shared/card-message-p2-revision-history.md
@@ -1,0 +1,87 @@
+---
+type: Journal
+title: "Journal: card-message-p2-revision-history (PR-C, card message P2 D10)"
+description: Record of PR-C — the D10 card revision history side table, query API, transient flag, bot clear+tombstone, and revoke cleanup, stacked on PR-B; includes two P1 fixes from verify (query-side withdraw gate + revoke-cleanup ordering).
+tags: ["message", "card", "wire-contract", "bot-api", "rate-limit", "space", "isolation", "error-response", "i18n"]
+timestamp: 2026-07-08T11:30:00Z
+# --- octospec extension fields ---
+task: card-message-p2-revision-history
+upstream: self
+source: self
+---
+
+# Journal: card-message-p2-revision-history (PR-C, card message P2 D10)
+
+## What was done
+
+Shipped D10 of the `card-message-interaction` contract — a queryable card
+revision history that the D6 latest-frame-only `content_edit` model cannot
+answer on its own. Stacked on PR-B (#548).
+
+- **`pkg/cardrevision`**: shared store over the new `octo_message_card_revision`
+  table (written by `modules/bot_api` on card edits/clear, read by
+  `modules/message`) — `AppendFrame` (+ cap-20 prune, rune-safe `plain`
+  truncate), `Query` (newest-first, incl. tombstones), `Clear` (tx: delete
+  frames + write auditable tombstone), `DeleteByMessageID`.
+- **`modules/bot_api`**: `botMessageEdit` appends a non-transient frame after
+  the content write (best-effort); new `POST /v1/bot/message/card/revisions/clear`
+  (owner-checked, writes a tombstone).
+- **`modules/message`**: `GET /v1/message/card/revisions` (summary / `full=1`);
+  extracted `authorizeCardChannelMember` shared with `card/action`; revoke
+  deletes a card's revisions.
+- **`pkg/cardmsg`**: `Transient` / `PlainFromContentEdit`.
+
+## Structural learnings
+
+- **Shared-table access via a `pkg/` store, not per-module raw SQL.** The
+  revision table is written by `bot_api` and read by `message` (same split as
+  `message_extra`). Instead of duplicating column-level SQL across two modules,
+  a `pkg/cardrevision.Store` wrapping `ctx.DB()`'s `*dbr.Session` centralizes
+  the schema. Both modules construct it. This is the cleaner pattern when a new
+  shared table has more than one accessing module.
+- **Extract the auth gate the moment a second endpoint needs it.** PR-C's
+  `GET /card/revisions` needs the exact anti-IDOR channel-binding + membership
+  gate `card/action` already had. Factoring `authorizeCardChannelMember`
+  (parameterized by the invalid/denied errcodes) made both endpoints share one
+  implementation — the brief explicitly called for this to prevent drift.
+- **New table → `octo_` prefix.** The parent brief's illustrative name
+  `message_card_revision` was overridden to `octo_message_card_revision` per the
+  repo's new-table naming rule; the prefix-less legacy tables are untouched.
+
+## Gotchas worth remembering (verify + code-review)
+
+- **Actionability/queryability must not outlive visibility.** The GET endpoint
+  first shipped with only the channel-membership gate and would have served the
+  revision content of a **revoked / globally-deleted / operator-locally-deleted**
+  card as long as the rows existed — the same class of gap PR-B hit on the action
+  path. Fixed with a shared `isCardMessageWithdrawn` (mirrors
+  `api_message_get.go:241`) applied on the query path, returning an empty list.
+  **Any new read/write path over a message must apply the same
+  revoke/is_deleted/user-local-delete visibility gate the existing single-message
+  read applies — and must NOT rely on a best-effort cleanup elsewhere.** See the
+  promoted learning.
+- **The lifecycle gate is only ONE layer of the canonical read (PR#549 review
+  B1).** Even after `isCardMessageWithdrawn`, the query still enforced a strict
+  *subset* of `respondSingleMessage`: it omitted the `visibles` allowlist, the
+  per-user read offset, the channel offset, and message expiry — so a
+  visibles-excluded or offset-truncated group member could still read the full
+  revision history (and with `full=1`, complete historical envelopes) though the
+  canonical single read hides it. Revision history discloses *more* than a single
+  action trigger, so its gate must be ≥ the action path's, not a subset. Fixed by
+  extracting `cardCanonicalVisibleToViewer` (visibles / expiry / user-offset /
+  channel-offset) out of `card/action` and calling it from **both** endpoints;
+  the query returns an empty list on a visibility miss (mirrors the withdrawn
+  path, no existence leak). New test `TestCardRevisionsCanonicalVisibility` pins
+  the visibles-excluded and offset-truncated cases. The lesson generalized: a
+  derived surface must match the canonical read in **every** layer, not just the
+  one that was top-of-mind — see the promoted learning.
+- **Best-effort cleanup must run right after the state commit, before any
+  fail-and-return step.** The revoke revision-cleanup was first placed after
+  `SendRevoke`; a `SendRevoke` failure returns early, so the DB would be marked
+  revoked while the revision rows survived. Moved to immediately after the revoke
+  transaction commits (before the notify loop). The query-side withdraw gate is
+  the second layer so a delete failure still can't leak.
+
+## Out of scope (sibling PR)
+
+D12 capability manifest (`GET /v1/bot/card/profile`) → PR-D.
--- a/.octospec/learnings/pending/card-message-p2-revision-history.md
+++ b/.octospec/learnings/pending/card-message-p2-revision-history.md
@@ -1,0 +1,76 @@
+---
+type: Learning
+title: "Every new read/write path over a message must reapply the visibility gate — not rely on cleanup"
+description: New endpoints touching a message (actions, history, projections) must reapply the revoke/is_deleted/user-local-delete visibility gate the single-message read applies; deleting derived data on revoke is a best-effort second layer, never the sole guarantee.
+tags: ["trust-boundary", "space", "isolation", "visibility", "message", "review"]
+timestamp: 2026-07-08T11:30:00Z
+# --- octospec extension fields ---
+source: self
+origin_task: card-message-p2-revision-history
+origin_pr: self
+status: pending
+candidate_rule: space-isolation
+---
+
+# Every new read/write path over a message must reapply the visibility gate — not rely on cleanup
+
+## Context
+
+This gap surfaced **three times** across the card message P2 PRs, on two
+different endpoints and across two *dimensions* of the canonical read gate:
+
+- **PR-B** (`POST /v1/message/card/action`): first shipped reading
+  `message_extra` only for `content_edit`, so a **revoked / deleted** card was
+  still actionable (a stale tap drove a bot event after the card was recalled).
+- **PR-C** (`GET /v1/message/card/revisions`), lifecycle dimension: first
+  shipped with only the channel-membership gate, so a **revoked / globally-deleted /
+  operator-locally-deleted** card's revision *content* was still queryable as
+  long as the rows existed.
+- **PR-C review B1** (`GET /v1/message/card/revisions`), canonical-visibility
+  dimension: even after the lifecycle gate was added, the query still enforced a
+  strict **subset** of the canonical read — it omitted the `visibles` allowlist,
+  the per-user read offset, the channel offset, and message expiry. A group
+  member excluded by a non-empty `visibles`, or whose read/channel offset was
+  past the message (cleared history), could still read the full revision history
+  (`full=1` → complete historical envelopes) though the canonical single read
+  would hide it. Fixed by extracting `cardCanonicalVisibleToViewer` (visibles /
+  expiry / offset) and calling it from **both** `card/action` and the query, so
+  the two endpoints share one visibility口径.
+
+Each time the established single-message read path (`respondSingleMessage` /
+`api_message_get.go`) already had the correct, complete gate — and each new
+endpoint reapplied only *part* of it. The canonical read is the reference; a
+derived surface must match it in **every** layer (membership + group/thread
+status + lifecycle revoke/delete + visibles + offset + expiry), not just the
+layer that was top-of-mind.
+
+A tempting-but-wrong mitigation on PR-C was "delete the derived rows on revoke."
+That does not cover `is_deleted` (global delete) or user-local delete (neither
+triggers the revoke path), and the delete was best-effort and mis-ordered
+(after the notify step, which could return early). Deletion is a second layer;
+it is never the guarantee.
+
+## The rule
+
+- **Any new endpoint that reads or acts on a stored message** (actions, edit
+  history, revision history, projections, previews, exports) MUST reapply the
+  **complete** visibility gate the canonical single-message read applies, in
+  every layer: channel membership + group/thread status +
+  `message_extra.revoke` / `message_extra.is_deleted` (global) +
+  `message_user_extra.message_is_deleted` (operator) + the `visibles` allowlist +
+  per-user read offset + channel offset + message expiry. Extract each dimension
+  into a shared helper (`authorizeCardChannelMember`, `isCardMessageWithdrawn`,
+  `cardCanonicalVisibleToViewer`) so the endpoints cannot drift and cannot ship a
+  *subset*.
+- **A derived surface must never be MORE permissive than the canonical read.**
+  Reviewers should diff the new gate against `respondSingleMessage` layer by
+  layer; "membership + one lifecycle check" is a subset, not the gate. Revision
+  history in particular discloses *more* than a single action trigger, so its
+  gate must be at least as strict as the action path's.
+- **Actionability and queryability must not outlive visibility.** If a message
+  is not viewable via the normal read, no derived surface may act on it or expose
+  its content.
+- **Cleaning up derived data on revoke is a best-effort second layer, not the
+  primary guarantee.** It must run immediately after the state-change commit and
+  before any step that can fail-and-return (e.g. an IM notify). The read/act path
+  must still gate independently, so a missed/failed cleanup cannot leak.
--- a/.octospec/log.md
+++ b/.octospec/log.md
@@ -4,6 +4,25 @@ Change history for this repo's `.octospec/`, following the
 [OKF](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md)
 change-log convention (§7). Newest first.
 
+## 2026-07-08 (PR-C)
+
+- **Change** — Task `card-message-p2-revision-history` (PR-C, card message P2
+  D10): card revision history. New `octo_message_card_revision` side table +
+  `pkg/cardrevision` shared store (written by bot_api on edits/clear, read by
+  message), `GET /v1/message/card/revisions` (summary / full=1) reusing the
+  extracted `authorizeCardChannelMember` gate, bot revision clear + auditable
+  tombstone, `transient` frame flag (progress frames skip history), and revoke
+  cleanup. Verify caught two P1s (fixed): the query path lacked the
+  revoke/deleted/user-local-delete visibility gate, and the revoke cleanup was
+  mis-ordered after the notify step. Code-review (B1) then caught that the query
+  still enforced a *subset* of the canonical read — missing the `visibles`
+  allowlist / read-offset / channel-offset / expiry layers `card/action` carries;
+  fixed by extracting `cardCanonicalVisibleToViewer` and sharing it across both
+  endpoints (+ `TestCardRevisionsCanonicalVisibility`). Stacked on PR-B; zero
+  octo-im changes. Journal:
+  `.octospec/journal/shared/card-message-p2-revision-history.md`;
+  learning: `.octospec/learnings/pending/card-message-p2-revision-history.md`.
+
 ## 2026-07-08
 
 - **Change** — Task `card-message-p2-action-loop` (PR-B, card message P2
--- a/.octospec/tasks/card-message-p2-revision-history/brief.md
+++ b/.octospec/tasks/card-message-p2-revision-history/brief.md
@@ -1,0 +1,181 @@
+---
+type: Task
+title: "Task: card-message-p2-revision-history"
+description: PR-C of the card message P2 protocol — D10 card revision history. A side table octo_message_card_revision appended by the botMessageEdit card branch (non-transient frames only, cap 20), a GET /v1/message/card/revisions query API (Auth → Space → SharedUIDRateLimiter + channel-membership, same gate as card/action), an optional envelope `transient` flag that keeps progress frames out of history, bot-driven revision clearing that writes an auditable tombstone row, and revision cleanup on message revoke. Stacks on PR-B. Zero octo-im changes.
+tags: ["message", "card", "wire-contract", "bot-api", "rate-limit", "space", "isolation", "error-response", "i18n", "testing"]
+timestamp: 2026-07-08T10:00:00Z
+# --- octospec extension fields ---
+slug: card-message-p2-revision-history
+upstream: self
+source: self
+---
+
+# Task: card-message-p2-revision-history
+
+> One task = one `.octospec/tasks/<slug>/` directory. This brief is the spec for
+> the work. AI may draft it from existing code; a human confirms it.
+
+## Goal
+
+Ship **D10 of the card message P2 contract** (`card-message-interaction`): a
+queryable **card revision history**. The full-frame-replacement model of D6
+(each `botMessageEdit` overwrites `message_extra.content_edit` with a complete
+new envelope, latest-frame-only) means the message table cannot answer "这张卡
+以前是什么" — which approval-class cards owe their audience. PR-C adds a side
+table that records each non-transient frame as a complete renderable envelope,
+plus a member-visible query API.
+
+```
+bot rewrites card → botMessageEdit content_edit (PR-B D6/D9)
+                  → IF NOT transient: append octo_message_card_revision row (frame + plain + editor + time)
+                  → cap 20 non-transient frames/message (oldest evicted)
+member opens history → GET /v1/message/card/revisions (Auth → Space → UID-limit + membership)
+                     → summary list by default; ?full=1 returns full renderable frames
+bot clears history → tombstone row (editor + time + cleared count), itself visible in the listing
+message revoked    → revisions deleted (a revoked message leaves no queryable content history)
+```
+
+**Stacking**: PR-C builds on **PR-B** (`feat/card-message-p2-action-loop`,
+#548) — it hooks the revision append into PR-B's D6 `botMessageEdit` card branch
+and reuses PR-B's `pkg/cardmsg` type-17 helpers. PR-C's PR is opened with
+**base = PR-B's branch** (stacked diff), retargeted to `main` after PR-B merges.
+
+## Decisions (inherited from D10, with implementation-level bindings)
+
+| # | Decision | Binding |
+|---|---|---|
+| D10.1 | **Side table = `octo_message_card_revision`** (NOT the brief's illustrative `message_card_revision` — repo rule: new tables MUST carry the `octo_` prefix, `dm_`/prefix-less reserved for legacy). Columns: `message_id`, `channel_id`, `channel_type`, `card_seq` (nullable — the frame's D9 seq if any), `content` TEXT (full renderable frame envelope; NULL for tombstone), `plain` (list summary; NULL for tombstone), `is_tombstone`, `cleared_count`, `editor_uid`, `edited_at`, + `db.BaseModel`. Index `(message_id, id)`. Shared-table pattern like `message_extra`: **written by `modules/bot_api` raw SQL, read by `modules/message`** — no cross-module service dependency. Migration lives in `modules/message/sql/`. | new migration + db layer |
+| D10.2 | **Appended by the D6 `botMessageEdit` card branch** (`modules/bot_api/send.go`, PR-B), alongside the `content_edit` overwrite, **only for non-transient frames**. `content_edit` storage stays latest-frame-only (render path unchanged); the revision table is the history surface. `editor_uid` = the editing bot. Append is best-effort-after-commit relative to the content write: the card update MUST NOT fail if the revision append fails (log + continue), since the durable card state is `content_edit`, not the history. | send.go card branch |
+| D10.3 | **Cap 20 non-transient frames per message** (tunable const). On append, evict the oldest frame rows beyond the cap for that `message_id`. Tombstone rows are audit markers, not frames — they do NOT count toward the cap and are not evicted by it. | append + prune |
+| D10.4 | **Envelope gains optional `transient: true`** — progress frames (thinking / tool-state) marked transient are applied to `content_edit` normally (D6/D9 unchanged) but **never enter the revision table**, so approval-state changes aren't drowned by progress noise. `cardmsg` gains a `Transient(payload)` reader (mirrors `CardSeq`); `transient` is a tolerated optional envelope field (forward-compat, not whitelisted per-element). | pkg/cardmsg + send.go |
+| D10.5 | **Query API `GET /v1/message/card/revisions`** — `AuthMiddleware` → Space → `SharedUIDRateLimiter` + **channel-membership check identical to `card/action`** (D3 anti-IDOR: bind channel from the stored message row, then membership of that channel). Request: `message_id`, `channel_id`, `channel_type` (channel params are the anti-IDOR lookup key, same as card/action — NOT taken as an authorization subject), `limit`, `full`. Default returns a summary list (`card_seq`/`plain`/`editor_uid`/`edited_at` + tombstone rows); `full=1` returns each frame's complete envelope. Visibility = **all channel members** (same permission as message edit history, D10(a)). | new message endpoint |
+| D10.6 | **Erasure allowed but explicitly recorded** (D10(b)): a bot may clear its own card's revisions via a new bot endpoint (bot-token, must own the message). Clearing deletes the frame rows and writes a **tombstone row** (`is_tombstone=1`, `cleared_count`, `editor_uid`, `edited_at`) that itself appears in the history — erasure is auditable, not silent. | new bot endpoint |
+| D10.7 | **Revoke deletes revisions** (D10(b)): when a message is revoked (`modules/message` revoke path), its `octo_message_card_revision` rows are deleted (best-effort) — a revoked message leaves no queryable content history. | revoke hook |
+
+## API surface (wire definitions)
+
+**`GET /v1/message/card/revisions?message_id=&channel_id=&channel_type=&limit=&full=`**
+(new — Auth → Space → SharedUIDRateLimiter + membership):
+
+```json
+// summary mode (default); full=1 adds a "card": {…full envelope…} field per non-tombstone row
+{ "revisions": [
+    { "card_seq": 2, "plain": "审批单 #42:✅ 已通过", "editor_uid": "bot_x", "edited_at": 1751791500 },
+    { "tombstone": true, "cleared": 3, "editor_uid": "bot_x", "edited_at": 1751791400 },
+    { "card_seq": 1, "plain": "审批单 #42:待审批", "editor_uid": "bot_x", "edited_at": 1751791234 } ] }
+// newest-first; errors: non-member 403-class / non-card message / channel mismatch → i18n envelope
+```
+
+**`POST /v1/bot/message/card/revisions/clear`** (new — bot-token, same middleware
+chain as other `bot_api` routes):
+
+```json
+{ "message_id": "8234567890123456789", "channel_id": "g_9f2c...", "channel_type": 2 }
+// bot must own the target message (sender == robotID); deletes frame rows, writes a tombstone.
+// errors: not-owner / non-card / non-bot-sender → i18n envelope
+```
+
+**`transient` on the type-17 edit envelope** (D10.4) — optional, consumed by the
+D6 `botMessageEdit` branch:
+
+```json
+{ "type": 17, "card": {...}, "card_version": "1.5", "profile": "octo/v2",
+  "card_seq": 5, "transient": true }
+// transient:true → content_edit updated (D6/D9), NO revision row appended.
+```
+
+Authoritative doc `docs/card-protocol.md` already carries the D10 contract (P1
+deliverable) — implementation must not drift; amend doc + brief together or not
+at all.
+
+## Load-bearing list
+<!-- touches tags: wire-contract, bot-api, rate-limit, space, isolation,
+     error-response, i18n, testing -->
+
+- **New side table `octo_message_card_revision`** (`wire-contract`): shared-table
+  pattern (written by `bot_api`, read by `message`, like `message_extra`).
+  Migration in `modules/message/sql/` following the repo's plain-DDL / `octo_`
+  prefix / MySQL-8 conventions. The revision-row shape (frame + plain + editor +
+  time + tombstone) is a stored contract the query API projects.
+- **`botMessageEdit` card branch append** (`bot-api`): `modules/bot_api/send.go`
+  (PR-B's D6 branch) gains a post-content-write revision append for non-transient
+  frames + cap-20 prune. **The card update MUST NOT fail if the append fails** —
+  `content_edit` is the durable state; the history is a secondary surface (log on
+  append error, still return OK + fire the CMD).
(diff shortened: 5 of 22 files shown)


OctoBoooot (changes requested):
## Review: P2 card revision history (D10) [stacked on #548] (#549)

**Verdict:** Request changes — the D10 feature itself is clean and byte-verified, but this branch is **stacked on a stale #548 commit** and carries the cross-bot card-tampering vulnerability that #548's head already fixed. Must rebase onto #548's fixed head before merge. Risk: large + security-sensitive → `needs-human-review` applied.

I isolated #549's real delta by comparing against #548's head (819a7f8a): `diverged, ahead 3 / behind 1` — the "behind 1" is the problem.

### Blocker — stale base carries the fixed-elsewhere cross-bot tampering vuln
- `modules/bot_api/send.go:773` (this branch) still has the **warn-only, no-return** `message_id`/`message_seq` mismatch: `ba.Warn("...保持旧行为继续执行")` — the confused-deputy where ownership is validated on the seq-message but writes key on `req.MessageID`, letting bot A overwrite bot B's card. #548's head (819a7f8a) **fixed this** to a hard reject (`send.go:779` "...拒绝" + `respondBotAPIRequestInvalid`; verified). #549 is stacked on the pre-fix #548 commit (`behind 1`) so it never picked the fix up.
- **Worse in context**: #549's own delta *adds* `AppendFrame` (revision history) to this same edit path (`send.go` +22). So a cross-bot tamper here would ALSO append a forged frame to the victim card's revision history — the vuln's blast radius grows under D10.
- Attribution: this is NOT a new #549 defect — #549's delta only adds the revision-append; the vulnerable mismatch code is inherited from the stale base. Fix is merge-order: rebase #549 onto #548's fixed head (819a7f8a or later), or land #548 first and rebase. After rebase, the `:773` block must read as the `:779` hard-reject.

### D10 feature — clean (byte-verified, no blockers)
- **Migration**: additive `octo_message_card_revision` side table (own table, `octo_` prefix, `(message_id,id)` index, `general_ci` collation) — safe.
- **Query API authz** (`api_card_revisions.go:44`): `authorizeCardChannelMember` — membership checked against the **stored** message's channel (anti-IDOR, same as the action endpoint); non-member → denied, non-card/mismatch → invalid (anti-enumeration). Mounted under Auth+Space+rate-limiter. Info-leak surface closed.
- **Write path**: revisions appended by the bot edit branch alongside content_edit; `transient:true` frames applied but excluded from history; cap 20 non-tombstone with oldest-evicted (`cardrevision.go`). Correct.
- **Clear + tombstone**: clear writes an auditable tombstone row (editor_uid + time + cleared_count) that appears in history; ownership-checked (`card_revision.go` `FromUID != robotID` → forbidden) — no cross-bot history wipe.
- **Delete on revoke**: wired in `revoke()` before SendRevoke, type-17-gated.
- **No regression to #548's action-path fixes** (verified in `api_card_action.go` despite its +89/-55): claim-before-stale-frame, space_id from stored authority, revoke/delete gate, dedup key without client_token — all intact.

### Praise
- The D10 implementation cleanly reuses #548's authorization model: `authorizeCardChannelMember` is the same anti-IDOR stored-channel membership gate the action endpoint uses, so the query API can't leak a foreign card's history. Clear-ownership + auditable tombstone + transient-exclusion all match the frozen D10 spec.

### Notes
- `reviewDecision: REVIEW_REQUIRED` / `mergeStateStatus: BLOCKED`; code CI pending. The blocker is purely the stacking base — the D10 code is approvable once rebased onto #548's fixed head. Recommend: merge #548 first (once its own gate clears), then rebase #549 and re-review the (small) delta.

Jerry-Xin (changes requested):
The PR is relevant to octo-server, but the new card-derived surfaces do not fully mirror canonical message visibility checks, creating a content leak and stale-action path.

**🔴 Blocking**

🔴 Critical: `GET /v1/message/card/revisions` can expose history for cards the caller cannot normally view.  
`modules/message/api_card_revisions.go:42-63` only applies channel membership and `isCardMessageWithdrawn`, and `isCardMessageWithdrawn` only checks `message_extra.revoke`, `message_extra.is_deleted`, and `message_user_extra.message_is_deleted` at `modules/message/api_card_action.go:350-365`. The canonical single-message read path also enforces `payload.visibles`, user channel offsets, channel-setting offsets, and expiration before returning content (`modules/message/api_message_get.go:225-286`, `modules/message/api.go:3267-3293`). As written, a channel member excluded by `visibles`, or a user who cleared the channel offset, can still request revisions and receive `plain` or the full card frame via `full=1`.

The same visibility gap also affects `POST /v1/message/card/action`: after `authorizeCardChannelMember`, the action path checks revoke/global delete/local delete at `modules/message/api_card_action.go:103-131`, but it does not reject `visibles`-hidden, expired, or offset-cleared cards before accepting and enqueueing bot side effects. Derived card query/action surfaces should share a helper equivalent to canonical single-message visibility, and tests should cover at least `visibles`, user channel offset, and expiration for revisions and actions.

**💬 Non-blocking**

🟡 Warning: `GET /v1/message/card/revisions` does not check `cardmsg.Enabled()` (`modules/message/api_card_revisions.go:21-96`), while action and clear do. If the rollout flag is intended to fail closed for all card endpoints, the read endpoint should follow the same behavior.

**✅ Highlights**

The migration and shared `pkg/cardrevision` store are cleanly scoped, and the modified packages compile with `go test ./pkg/cardmsg ./pkg/cardrevision ./modules/robot ./modules/message ./modules/bot_api -run '^$'`. The revoke cleanup is placed after commit and before notify, which is the right ordering for best-effort derived-data cleanup.

OctoBoooot (changes requested):
## Review (update): P2 card revision history (D10) [stacked on #548] (#549)

**Verdict:** Request changes (unchanged) — adding a second, independent blocker Jerry-Xin surfaced that I under-graded: the card surfaces gate on `isCardMessageWithdrawn`, which is **narrower than canonical message visibility** (skips `visibles`/offset/expire). Credit Jerry-Xin. My prior stale-base blocker still stands too.

### Blocker #1 (stale base — from my prior review, stands)
Branch is stacked on the pre-fix #548 commit (`compare 819a7f8a...HEAD = behind 1`): `send.go:773` still has the warn-only message_id/seq mismatch (cross-bot tampering) that #548 head fixed to a hard reject. Rebase onto #548's fixed head. #549 even extends this path with revision-append, so a tamper would forge history too.

### Blocker #2 (visibility incompleteness — Jerry-Xin's, byte-verified, I concede)
Both new card surfaces gate on `isCardMessageWithdrawn` (`api_card_action.go:350-365`), which checks only `message_extra.revoke`, `message_extra.is_deleted`, and `message_user_extra.message_is_deleted`. But the **canonical single-message read path** (`api_message_get.go:225+`) enforces a strictly larger set before returning content:
- `visiblesAllows(payload, loginUID)` — the `visibles` per-message recipient whitelist
- user-level history-clear offset (`channel_offset`)
- channel-setting offset (`channel_setting.offset_message_seq`) — a distinct table, "必须都查"
- (plus revoke/delete, which `isCardMessageWithdrawn` does cover)
Consequence — reachable on #549's own code:
- **`GET /v1/message/card/revisions`**: a channel member excluded by `visibles`, or who cleared their offset, still receives `plain` / full card frame (`full=1`) — content leak.
- **`POST /v1/message/card/action`**: same members can still trigger bot side-effects on `visibles`-hidden / offset-cleared / expired cards — stale-action.
Fix (Jerry-Xin's, endorsed): extract a canonical-visibility helper mirroring `api_message_get.go` (visibles + both offsets + expire + revoke/delete) and use it in both the revisions GET and the action endpoint; test visibles / offset / expiration for each.

### On my error
My prior review credited `authorizeCardChannelMember` (channel membership) as closing the revisions info-leak surface. **Membership ≠ full visibility** — a member can still be excluded by `visibles` or have cleared their offset. I checked the card surfaces against each other and against membership, but did not cross-check against the *canonical single-message read path*, which is the actual visibility contract. That's the "read the adjacent canonical path, not just the endpoint's own logic" lesson. Conceding to Jerry-Xin.

### 🟡 (Jerry-Xin, non-blocking)
`GET /v1/message/card/revisions` doesn't check `cardmsg.Enabled()` while action/clear do — if the flag should fail-closed for all card endpoints, the read should match.

### D10 core still sound
Migration (additive side table), clear-ownership + auditable tombstone, transient-exclusion, cap-20 evict, delete-on-revoke, and the shared `pkg/cardrevision` store are clean and byte-verified; #548's action-path fixes (claim-order, space authority, dedup key) are intact. The two blockers are (1) merge-order/base hygiene and (2) visibility-contract completeness — both fixable without touching the D10 storage design.

### Notes
- `reviewDecision: REVIEW_REQUIRED` / `mergeStateStatus: BLOCKED`; `needs-human-review` applied. Sequence: land #548 (once yujiawei re-clears at 819a7f8a) → rebase #549 → add the canonical-visibility helper + tests → re-review.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review: octo-server #549 — feat(cardmsg): P2 card revision history (D10)
**Head SHA:** `0cc9ae94e95cd1d045d7fbbbd34730dfe26dcde0`
**Base:** `7cffcff6` (stacked on PR #548)
**Scope:** 40 files changed, 5 commits unique to this PR

---

## Verification Conclusions

All D3–D11 design contracts verified against production code:

| Check | Result | Evidence |
|-------|--------|----------|
| D3 anti-IDOR channel binding | ✅ PASS | `authorizeCardChannelMember` shared gate (api_card_action.go:270–326) |
| D4 idempotency claim lifecycle | ✅ PASS | SET NX 60s → confirm 24h → release on failure (card_action_claims.go:65–85) |
| D5 typed bot event chokepoint | ✅ PASS | GenSeq→ZAdd→Expire shared with message events (robot/api.go:245) |
| D6 cross-type mutation rejection | ✅ PASS | origIsCard vs editIsCard comparison (send.go:759–770) |
| D9 CAS bounded retry | ✅ PASS | SELECT FOR UPDATE → GenSeq inside lock → upsert (send.go:1002–1042) |
| D10 revision lifecycle | ✅ PASS | Append/Query/Clear/DeleteByMessageID all correct (cardrevision.go) |
| D10.5 query auth + visibility | ✅ PASS | Shared auth gate + isCardMessageWithdrawn double-layer (api_card_revisions.go:42–60) |
| D10.6 bot clear ownership | ✅ PASS | IMSearchMessages + from_uid == robotID (card_revision.go:72–88) |
| D10.7 revoke cleanup ordering | ✅ PASS | DeleteByMessageID before SendRevoke, after tx commit (api.go:2836–2843) |
| D11 inputs trust boundary | ✅ PASS | Declared IDs only, type-specific validation, 4KiB/16KiB caps (inputs.go:48–91) |
| octo/v2 profile + ID uniqueness | ✅ PASS | registerID enforces frame-internal uniqueness (validate.go:127–134) |
| Migration safety | ✅ PASS | Nullable ADD COLUMN + new table with composite index |

## Findings

### P2-1: ErrBotAPICardEditForbidden comment is stale after D6 unlock
**File:** `pkg/errcode/bot_api.go:395`
**Severity:** P2 (cosmetic documentation debt)
**Diff-scope:** Pre-existing registration; reference in `bot_api/send.go` removed by this PR.

The error code is still registered and still referenced in `modules/robot/api.go:2004` (legacy robot edit path), so the registration itself is fine. However, the comment at line 395 references "P2 sibling D6 以" without noting that D6 is now fully implemented. Future readers may be confused about whether card edits are allowed.

**Suggestion:** Update the comment to note that D6 is implemented in `bot_api/send.go` via `NormalizeContentEdit`, and this error is only used by the legacy robot edit path.

### P2-2: AppendFrame INSERT + prune are non-transactional
**File:** `pkg/cardrevision/cardrevision.go:65–84`
**Severity:** P2 (best-effort by design)
**Diff-scope:** New code introduced by this PR.

`AppendFrame` runs a separate INSERT followed by a DELETE to prune to `CapFrames(20)`. Under concurrent bot edits, a race could briefly over-prune by 1 frame. This is explicitly documented as best-effort (append failure only logged, never blocks edit), so the impact is negligible. The non-transactional approach avoids lock contention on a hot path for a secondary feature.

**Suggestion:** Accept as-is. Consider a brief inline comment noting the non-transactional nature for future maintainers.

## Data Flow Tracing

All consumed data traced to validated upstream sources:

- **event_data.inputs**: `ValidateInputs` checks declared `Input.*` IDs, string values, type-specific validation, 4KiB/16KiB caps → shape-trusted values passed to `EnqueueBotTypedEvent` (api_card_action.go:177 → inputs.go → api_card_action.go:196)
- **event_data.data**: Extracted server-side from effective frame by `SubmitAction` — never from request (api_card_action.go:163 → interactive.go:findSubmitAction → api_card_action.go:203)
- **card_seq CAS**: SELECT FOR UPDATE → compare stored ≥ incoming → conflict (no write, no version) → GenSeq inside lock → upsert → commit; version allocated inside lock ensures delta-sync monotonicity (send.go:1002–1042)
- **revision append**: After successful CAS/LWW edit → `AppendFrame` (best-effort); transient frames excluded (send.go:957–970)
- **revision clear**: Transactional count→delete→insert(tombstone)→commit (cardrevision.go:108–130)
- **revision delete on revoke**: Before `SendRevoke`, after tx commit; query-side `isCardMessageWithdrawn` provides double-layer defense (api.go:2839, api_card_revisions.go:55)
- **idempotency claim**: SET NX EX 60s "pending" → validation → enqueue → SET XX EX 24h event_id; failure → DEL; crash → key expires at 60s (card_action_claims.go:65–85)
- **space_id authority**: Derived from storage authority (group table / payload server-injected), never from operator request context (api_card_action.go:311–347)
- **effective frame**: content_edit preferred over original payload; stale buttons fail-closed (api_card_action.go:157–160)

## Blind-spot Checklist

| Item | Status | Evidence |
|------|--------|----------|
| C1 — Dual-path parity | ✅ CLEAR | `authorizeCardChannelMember` shared between `cardAction` and `getCardRevisions`; revoke/delete checks in both action and query paths; bot clear uses same `IMSearchMessages` ownership |
| C2 — Control-flow ordering | ✅ CLEAR | Claim before validation, release on all post-claim failures; CAS after dedup check; non-card_seq path uses LWW (no behavioral change) |
| C3 — Authorization boundary | ✅ CLEAR | All 3 new endpoints have proper auth; full D3 chain on card action; shared gate on revisions; ownership check on bot clear |
| C4 — Authorization lifecycle | N/A | No container-member state cascade changes |
| C5 — Build/deploy | N/A | No build/deploy artifact changes |
| C6 — Governance | N/A | Not a governance/policy PR |

## Cross-round Blocker Review

## Extra Observations

1. **Idempotency-before-validation (P1-4)** is well-reasoned: prevents a legitimate replay from being misclassified as stale-frame 400 when the button has been removed between edits.
2. **Non-card_seq LWW fallback** correctly preserves zero behavioral change for legacy bots (send.go:935–954).
3. **Test coverage** is thorough: `card_p2_edit_im_test.go` covers D6 happy path, cross-type mutation rejection, D9 CAS (sequential stale, concurrent with version monotonicity polling, idempotent retransmit), and D10 revision lifecycle (append/transient/clear). `api_card_action_test.go` covers D4 idempotency, D3 trust model, and D11 inputs.

---

[Octo-Q] verdict: **APPROVE** — Zero P0/P1 findings. Two P2 items (stale comment, non-transactional prune) are cosmetic/best-effort by design. All D3–D11 design contracts verified, data flow tracing complete, blind-spot checklist clear.

OctoBoooot (changes requested):
## Review (delta): P2 card revision history (D10) (#549)

**Verdict:** Request changes — the stale-base blocker is resolved (rebased onto #548's fixed head ✓), but the visibility-authz blocker is still open, and the rebase now inherits #548's two open majors (inlineAction, card_seq). Not mergeable until those land.

Head moved to ad3ce49f (rebased). Status of each blocker:

### ✅ Blocker #1 (stale base) — RESOLVED
`compare 819a7f8a...ad3ce49f = ahead 3 / behind 0` (was behind 1). #549 is now on #548's fixed head, so the cross-bot mismatch hard-reject is in the base. Good.

### ❌ Blocker #2 (visibility authz) — STILL OPEN (Jerry-Xin's + mine)
`isCardMessageWithdrawn` (`api_card_action.go:350-366`) is unchanged — still only `revoke`/`is_deleted`/`message_is_deleted`. The revisions GET (`api_card_revisions.go:54`) and the action endpoint still don't enforce the canonical read path's `visibles` whitelist, user/channel offsets, or expiration (`api_message_get.go:225+`). So a channel member excluded by `visibles`, or who cleared their offset, still reads full revision history (`full=1`) / triggers actions. Fix unchanged: shared canonical-visibility helper (visibles + both offsets + expire + revoke/delete) used by both card surfaces + tests.

### ❌ Inherited from base — #548's two open majors now in #549's base
Because the rebase landed on 819a7f8a (which #548 hasn't yet fixed these on):
- **F1 inlineAction** (`validate.go`): `grep inlineAction = 0` — `Input.Text.inlineAction` still not routed through `checkURL`/`w.action`; off-allowlist `javascript:` OpenUrl on an input passes validation. And #549's `AppendFrame` would write such a frame into revision history.
- **F2 card_seq** (`interactive.go:43/:123/:163/:180/:190`): all plain `json.Unmarshal` → float64; card_seq loses precision >2^53, breaking the D9 CAS. #549 adds more decode sites on this same pattern.
These are #548's to fix; #549 just needs to rebase again once #548 lands them.

### D10 core — still sound
Migration additive, clear-ownership + tombstone, transient-exclusion, cap-20 evict, delete-on-revoke, shared `pkg/cardrevision` store — all byte-verified clean. The blockers are visibility-contract completeness + inherited base majors, not the D10 storage design.

### 🟡 (mochashanyao, non-blocking)
- `ErrBotAPICardEditForbidden` comment stale — update.
- `AppendFrame` insert+trim is non-transactional — under concurrent edits the cap can momentarily exceed 20 before trim; cosmetic (soft cap), but a one-line comment on the tradeoff (or a txn) would be clearer.

### Notes
- `reviewDecision: REVIEW_REQUIRED` / `mergeStateStatus: BLOCKED`; `needs-human-review` applied. Sequence: #548 lands F1/F2 fixes → #549 rebases + adds the canonical-visibility helper + tests → re-review.

Jerry-Xin (changes requested):
Re-review at ad3ce49f (rebased onto #548's fixed head 819a7f8a). B2 (stale-base stacking) is resolved; B1 (visibility-authz incompleteness) still stands, so holding REQUEST_CHANGES. Not vote-following the APPROVE.

## 🔴 Blocking

**B1 — card revision read + action enforce a strictly NARROWER visibility set than the canonical single-message read (content leak + stale-action).**
- `modules/message/api_card_revisions.go` `getCardRevisions` gates ONLY on `authorizeCardChannelMember` (membership + anti-IDOR channel binding) + `isCardMessageWithdrawn`. The header comment states outright: "可见性 = 全体频道成员". The action endpoint `api_card_action.go:107-127` inlines the same triad.
- `isCardMessageWithdrawn` (`api_card_action.go:350-364`) covers ONLY `message_extra.revoke==1` / `is_deleted==1` / `message_user_extra.message_is_deleted==1`.
- The canonical read path `api_message_get.go:206-264` enforces a strictly LARGER set (its own comment lists them): `visiblesAllows(payload, loginUID)` (:230), user channel offset `channel_offset` `MessageSeq <= userOffset.MessageSeq` (:255-259), channel-setting offset `channel_setting.offset_message_seq` via `lookupChannelOffsetSeq` (:264,:337), and expiration.
- Consequence: a channel member excluded by a card's `visibles`, or who cleared their user offset, or below the channel-setting offset, or on an expired card, is still allowed to read full revision history (plain + `full=1` frame envelope) and to trigger card actions. Content leak + stale-action on messages that are NOT visible to that member on the normal read path.
- Fix: both card surfaces (revisions GET + action) must enforce the FULL canonical visibility (visibles + user offset + channel-setting offset + expiration), ideally via a shared canonical-visibility helper, not the narrower `isCardMessageWithdrawn`. Add a test for a visibles-excluded / offset-cleared member being denied both revision read and action (current `api_card_revisions_test.go` only covers non-member + IDOR).

## 💬 Non-blocking

- **F1/F2 inherited from base #548 @819a7f8a (merge-order dependency, not #549's delta):** #549 does NOT touch `validate.go`; the `Input.Text` inlineAction gap (`validate.go:263-286`) and `card_seq` float64 decode (`interactive.go:146` `CardSeq`) live in the base. #549's own `interactive.go` delta only adds `Transient/PlainFromContentEdit`. These must be fixed on #548 before/with merge; flagging as a hard ordering dependency.
- **AppendFrame non-transactional insert+prune (`cardrevision.go`):** INSERT commits before the prune DELETE (unlike `Clear`, which uses a tx). Worst case is a transient +1 over-cap that self-heals on the next append; no committed-frame loss, and same-message writers are serialized upstream by the `card_seq FOR UPDATE` CAS (`send.go:1016`). 🟡 as graded — recommend wrapping in a tx for cleanliness, not blocking.
- `ErrBotAPICardEditForbidden` stale comment (mochashanyao) — cosmetic.

## ✅ Highlights

- **B2 fully resolved:** rebased onto 819a7f8a; `send.go:772-786` message_id/seq mismatch is now a hard-reject (`respondBotAPIRequestInvalid(c,"message_id"); return`), symmetric with the user-edit path — the AppendFrame edit path sits behind this gate, closing the cross-bot forged-frame vector.
- Stored revision content is the server-normalized/Finalized body (`send.go:66-72` reassigns `req.ContentEdit = normalized` before AppendFrame) — no stored-XSS via raw bot input.
- Clear endpoint enforces owner (`FromUID != robotID → Forbidden`) + type-17 + tombstone in a single tx; migration is standalone (`octo_` prefix, consistent `utf8mb4_general_ci`, `idx_message(message_id,id)`), cap-20 prune, tombstone audit rows preserved.

yujiawei (dismissed):
# Code Review — PR #549 (octo-server)

**Head SHA:** `0cc9ae94e95cd1d045d7fbbbd34730dfe26dcde0` · **Base:** stacked on #548 · Security-sensitive review.

This branch is stacked on #548 but the two branches have **diverged**: #548's head is not an ancestor of this PR. The true delta for this PR is computed against the common merge-base (`97c75ebc`), which is where this branch forked from #548. A naive diff against #548's head shows #548's own later hardening commit as spurious "removals" in this PR — it is not. Verified by simulating the eventual merge: after #548 lands, `botMessageEdit` retains #548's message_id/seq-mismatch reject and revoke/is_deleted gate **and** this PR's D10 append block; the only merge collision is a test-file edit in `card_p2_edit_im_test.go` (both branches rewrote the same test function), which resolves cleanly.

Build (`go build ./...`), `go vet`, and `pkg/cardmsg` unit tests pass on the head SHA. Integration tests require MySQL/Redis/WuKongIM which weren't provisioned in this review environment.

## 1. Spec compliance

**Spec: ✅**

- Missing: none. D10.1–D10.7 are all implemented — side table `octo_message_card_revision` with `octo_` prefix + migration in `modules/message/sql/`; best-effort non-transient append on the D6 `botMessageEdit` card branch; cap-20 prune; `GET /v1/message/card/revisions` (summary + `full=1`) behind the same member gate as `card/action`; `transient` envelope field; bot clear endpoint writing an auditable tombstone; revoke deletes revisions.
- Over-build: none. No functionality beyond the D10 delta; the out-of-scope list (D12 manifest, octo-im changes, client UI, revision diffing) is untouched.
- Deviation: none. The shared member gate is factored into `authorizeCardChannelMember` and reused by both endpoints, exactly as the brief requires ("factor the card-message member gate into a shared helper reused by both").

The anti-IDOR binding is correct: `queryMessageByID` pins `channel_id`/`channel_type`/`message_id` in the WHERE clause, and membership is then checked against the **stored** row's channel — so a message in channel B cannot be reached by declaring channel A. This is byte-for-byte the gate `card/action` used before the refactor.

## 2. Code Quality

**Quality: Approved**

The shared-gate refactor is behavior-preserving: no check was dropped, no error code changed, no ordering altered (independently confirmed against the base version). `cardAction` still receives the same stored message and runs its downstream sender-is-bot / lifecycle / effective-frame checks unchanged. The revoke cleanup is correctly ordered (after tx commit, before the fail-prone `SendRevoke` loop) and backed by a second visibility layer (`isCardMessageWithdrawn`) so history cannot leak even if the best-effort delete fails.

No P0/P1 findings. The following are P2 (non-blocking):

- **P2 — `content` column is `TEXT` (~64 KiB) while cards validate up to 512 KiB.** `pkg/cardmsg/cardmsg.go:50` (`MaxPayloadBytes = 512<<10`) vs the migration's `content TEXT`. Worth noting this is **not a new** silent-loss surface: the authoritative `message_extra.content_edit` column is *also* `TEXT` and is written first (non-best-effort), so a card exceeding 64 KiB hits the same ceiling on the primary write before the revision append. The revision store mirrors the existing store's capability rather than regressing it. Suggest a follow-up to widen **both** `content_edit` and revision `content` to `MEDIUMTEXT` if large cards are expected.
- **P2 — `AppendFrame` INSERT + cap-20 prune are not atomic**, and `Clear` (count→delete→insert tombstone) is not serialized against a concurrent `AppendFrame`. A race could transiently exceed the cap by one, or leave one frame after a clear. Impact is low: appends are single-writer-per-card (cards are bot-sender-only) and explicitly best-effort. Accept as-is; a one-line comment on the non-transactional prune would help future maintainers.
- **P2 — revision `Query`/`Clear`/prune filter by `message_id` only**, not `(message_id, channel_id, channel_type)`, despite the table carrying channel columns. Safe today because `message_extra` enforces a global `UNIQUE INDEX message_id` and the whole card subsystem (content_edit CAS, card_action, revoke) already keys on `message_id` alone — so `message_id` is a system-wide unique key by construction, and the auth gate binds the message row before any revision access. Carrying the channel through as a defense-in-depth predicate would harden it against any future change to that invariant.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. No P0/P1. The three P2 items are hardening suggestions, none blocking.

## 4. Suggestions

- Widen `content_edit` and the new `content` column to `MEDIUMTEXT` together if cards near the 512 KiB validation ceiling are a real workload.
- If cap-20 is a hard invariant (not just a soft bound), wrap append+prune in one transaction; if clear is a strict audit erasure, serialize it against append on the message key.
- Add `(message_id, channel_id, channel_type)` predicates to the revision store queries as defense-in-depth.

## 5. Additional notes

Advisory second-opinion legs run for this security-sensitive PR: the line-level adversarial leg completed and independently confirmed the shared-gate refactor changed no security behavior; the long-context cross-family leg **timed out and is absent from this review** (its coverage — whole-repo/architectural cross-file reasoning — was compensated by the reviewer's own merge-base and post-merge simulation analysis above). Absence of a leg is not treated as approval.

Jerry-Xin (changes requested):
**REQUEST_CHANGES** @ 766e465d — B1 is only partially resolved. The rebase onto merged #548 is clean (base `main`, behind_by=0) and the revisions endpoint now reuses #548's shared `authorizeCardChannelMember`, but the canonical read-path gate is applied only in part: the same visibles / offset / expiry layers your own `card/action` handler carries are missing on the revision-history query.

🔴 **B1 residual (partial gate — information disclosure): `getCardRevisions` omits visibles + user-offset + channel-offset (+ expiry).**
`modules/message/api_card_revisions.go` gates with `authorizeCardChannelMember(...)` (membership + group-status + thread-status, via #548's shared helper) plus `isCardMessageWithdrawn(...)` (revoke / is_deleted / user-local-delete). That is a strict **subset** of the canonical read path.

The canonical single-read `respondSingleMessage` (`api_message_get.go`) applies, beyond membership/status:
- `visiblesAllows(msgModel.Payload, loginUID)` — line 230
- user read-offset — lines 255–259 (`msgModel.MessageSeq <= userOffset.MessageSeq`)
- channel offset — line 264 (`lookupChannelOffsetSeq`)
- expiry — via `from()`

Your sibling `card/action` handler (`api_card_action.go:141`) explicitly replicates ALL of these, with a comment stating the exact leak class: *"被 visibles 排除 / 已清理历史的成员…仍能枚举可读的 message_id"*. The revisions endpoint does not carry them.

Reachable leak (now open): a card in a group carrying a non-empty `visibles` allowlist that excludes member X. X is `ExistMemberActive` → `authorizeCardChannelMember` passes; not revoked → `isCardMessageWithdrawn` passes → `GET /card/revisions` returns the full revision list, and with `full=1` the complete historical card envelopes. Canonical read would hide it (visibles miss ⇒ IsDeleted). Same for a member whose read/channel offset is past the message (cleared history). Revision history discloses *more* than a single action trigger, so this is at least as severe as the action-path gap #548 fixed.

Fix: apply the same applicable layers before returning rows. Since `authorizeCardChannelMember` returns `msgM`, gate on it directly — e.g. `visiblesAllows(msgM.Payload, loginUID)`, the user-offset/channel-offset check for non-person channels (reuse the `card/action` block / factor a shared helper), and expiry — returning an empty list (or invalid) on miss, mirroring `respondSingleMessage`/`card/action`. Add a test asserting a visibles-excluded member and an offset-truncated member get an empty list (current `api_card_revisions_test.go` covers member-gate/withdrawn/cap but not visibles/offset).

💬 Non-blocking:
- CI: **Test** and **Lint** are still `pending` at this head — will confirm green before approving.
- IDOR: clean — `Store.Query(messageID)` is single-table by `message_id`, but access is bound upstream by `authorizeCardChannelMember` (`queryMessageByID(lookupChannelID, channelType, messageID)`), so cross-channel id guessing is rejected at authz. Good.
- Precision (F2): clean — `message_id` stays a string end-to-end; no float64 parse of authoritative ids.
- Migration collation (#540): `octo_message_card_revision` is `utf8mb4_general_ci`; the revision read path does no cross-table JOIN, so no collation-mismatch risk.

✅ Highlights:
- Clean rebase onto merged #548; reuses the shared `authorizeCardChannelMember` (anti-IDOR channel binding + membership + group-status + thread-status) rather than hand-rolling — correct direction.
- Clear/tombstone is sound: `Clear` soft-deletes non-tombstone frames and inserts an audit tombstone row (editor/time/cleared_count) in one tx; the query returns tombstones and the handler renders them explicitly. Withdraw double-defense (delete + `isCardMessageWithdrawn` fallback) is a nice fail-closed touch.

modules/message/api_card_revisions.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+	}
+	full := c.Query("full") == "1"
+
+	// D10.5 门禁：与 card/action 同口径（anti-IDOR 绑定 + 存储频道成员）。非卡片/
+	// 不存在/频道不匹配 → invalid；非成员 → denied（防枚举）。
+	if _, handled := m.authorizeCardChannelMember(c, loginUID, messageID, channelID, uint8(channelTypeI),
+		errcode.ErrMessageCardRevisionInvalid, errcode.ErrMessageCardRevisionDenied); handled {
+		return
+	}
```
OctoBoooot, line 51: **[blocker]** The revisions query gates membership + withdrawn but omits the visibility-parity layer that `cardAction` enforces (visibles-whitelist + expiration + user-offset + channel-offset). `authorizeCardChannelMember` returns `msgM`, but here it is discarded (`_, handled`) and the code goes straight to `Query`.
Failing input: bot posts a group card with `visibles=[A]` (targeted). `visiblesAllows` returns false for non-A members and `cardAction` correctly 400s them — but member B (not on the whitelist) can `GET /card/revisions?message_id=…&full=1` and receive every historical frame envelope of a card B was never allowed to see. Same for a user past their channel/user offset, or an expired card. This is the #548 card-action B1 seam, now on the query path.
Direction: capture `msgM` and apply `visiblesAllows` + expire + `channelOffsetDB`/`lookupChannelOffsetSeq` exactly as `cardAction` does (api_card_action.go:143-186) — ideally share those too, since the `authorizeCardChannelMember` extraction was precisely to stop the two paths drifting.

OctoBoooot (changes requested):
## Review: feat(cardmsg): P2 card revision history (D10) (#549)

**Verdict:** Request changes — the D10 query API replicates part of the canonical read gate but still omits the visibles/offset/expire layers, so a card visible only to some members leaks its full content history to any member (my prior B1, partially addressed).

Delta cleanly isolated now that #548 is merged: `aaacb612…766e465`, **ahead 3 / behind 0** — no #548 contamination. The stacking warning in the PR body is moot; the diff is D10-only.

Risk tier: normal, but 22 files → `needs-human-review` (size gate).

### Blockers
- `modules/message/api_card_revisions.go:47-71` — the revisions query enforces membership + withdrawn, but **not** the visibility parity gate that `cardAction` enforces (visibles-whitelist + expiration + user-offset + channel-offset). `authorizeCardChannelMember` returns `msgM` but the caller discards it (`_, handled`), then goes straight to `Query`. Failing input: a bot posts a group card with `visibles=[A]` (targeted card — `visiblesAllows` returns false for non-A members, and `cardAction` correctly 400s them); member B (not on the whitelist) calls `GET /card/revisions?message_id=…&full=1` → receives every historical frame envelope of a card B was never allowed to see. Same seam for a user past their channel/user offset or an expired card. This is the same read-authz divergence as the #548 card-action B1, now on the query path. Direction: capture `msgM`, and apply `visiblesAllows` + expire + `channelOffsetDB`/`lookupChannelOffsetSeq` exactly as `cardAction` does (api_card_action.go:143-186) — ideally extract those into a shared helper too, since the whole point of the `authorizeCardChannelMember` extraction was to stop the two paths drifting.

### Praise
- `isCardMessageWithdrawn` (api_card_action.go) is correct double-defense: it re-checks revoke/is_deleted/user-deleted at query time rather than trusting the best-effort `DeleteByMessageID` on revoke — so a card whose revision rows survive a failed delete still returns empty. Adversarially verified: covers all three withdrawal signals, gated before `Query`, no content leak. `TestCardRevisionsWithdrawnHidden` exercises all three.
- Migration `20250708000002`: additive `CREATE TABLE`, `utf8mb4_general_ci` consistent with siblings (avoids the 1267 collation mismatch from #540), `idx_message(message_id,id)` matches the `ORDER BY id DESC` query. Clean.

- `TestCardRevisionsMemberGate` covers member/IDOR and `WithdrawnHidden` covers revoke/delete, but there is no test asserting a visibles-excluded (or offset/expired) member gets an empty/denied result — adding one alongside the blocker fix would lock the parity in.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #549 (octo-server)

## Summary

P2 card revision history (D10): side table `octo_message_card_revision` recording each non-transient card frame as a complete renderable envelope, plus member-visible query API and bot clear endpoint with tombstone. The PR also wires the card/action endpoint's idempotency claims, input validation (D11), CAS write paths (D9), and bot edit unlock for card messages (D6).

Overall: well-structured, security-conscious implementation. Anti-IDOR channel binding, fail-closed input validation, and proper CAS with InnoDB retry are all solid. Test coverage spans the critical paths — member gate, IDOR, cap-20, visibility, CAS, concurrent CAS, and revision IM.

## Verification

Static analysis only at head `766e465dc85b3e03dd2b9c1787787a26b7d12755`; build and tests not executed in this environment.

- ✅ **Anti-IDOR gate** — `authorizeCardChannelMember` (`modules/message/api_card_revisions.go:43`, `modules/message/api_card_action.go:89`) derives channel binding from the stored message, not request params. Non-members and cross-channel IDOR attempts are rejected before any data access.
- ✅ **CAS integrity** — `cardSeqCASWrite` (`modules/bot_api/send.go:1031`) holds FOR UPDATE row lock, compares stored card_seq with incoming, and upserts in the same transaction. Idempotent same-frame retries (same seq + same hash) return success without re-writing.
- ✅ **Version monotonicity (in-process)** — Both CAS and LWW paths allocate `version` via `GenSeq` after acquiring the row lock, preventing out-of-order version allocation within a single process.
- ✅ **Visibility gates** — `isCardMessageWithdrawn` (`api_card_revisions.go:72`) checks revoke, is_deleted, and user-local-deleted before returning any revision data. Tests confirm all three states return empty.
- ✅ **Data-flow: card_seq** — `decodeEnvelope` uses `json.Decoder.UseNumber()` (`pkg/cardmsg/interactive.go:189`), preserving 64-bit integers through the normalize/CAS chain. `CardSeq` explicitly rejects float64 (`:161`), failing safe to LWW rather than accepting truncated sequence numbers.
- ✅ **Idempotency claims** — Redis SET NX (pending, 60s) → SET XX (confirm, idemTTL) → Lua CAS-del (release only if pending). User-level lock prevents concurrent double-submission. Crash between claim and confirm expires naturally.
- ✅ **Input validation (D11)** — `ValidateInputs` (`pkg/cardmsg/inputs.go:47`) is fail-closed: undeclared keys rejected, types checked, sizes bounded (4KiB per text, 16KiB total). `collectInputSpecs` traverses Container/ColumnSet matching the send-time walker.
- ✅ **Transient gate** — `TransientFromContentEdit` (`pkg/cardmsg/interactive.go:213`) correctly gates revision append; transient frames still apply to content_edit but skip history.
- ✅ **Cap-20 eviction** — `AppendFrame` (`pkg/cardrevision/cardrevision.go:58`) prunes oldest frames beyond cap after each insert. Query independently caps at `limit`.
- ✅ **Clear with tombstone** — `Clear` (`pkg/cardrevision/cardrevision.go:84`) runs count→delete→insert in a single transaction. Tombstone carries `cleared` count for audit.
- ✅ **Bot edit unlock** — `send.go:920-970` gates card edit to bot-only (user/robot paths remain closed per D6 invariant). Cross-type mutation (card↔non-card) rejected. Lifecycle check before edit.

## Findings

No P0/P1 issues. Three P2 items below.

### P2 — Stack artifact: sticker upload limits (`modules/common/api.go`)

The `StickerUploadLimits` field added to `appConfigResp` is unrelated to card revision history — it appears to be from the underlying PR#548 stack base. Not a defect, but worth noting for clean merge/revert isolation. If this field was intended for a different PR, it should be cherry-picked separately.

### P2 — Non-transactional cap-20 prune (`pkg/cardrevision/cardrevision.go:72`)

`AppendFrame` runs INSERT and prune-DELETE as separate statements (not in a transaction). A concurrent reader calling `Query` between the insert and prune could briefly see `cap+1` frames. This is cosmetic — the next `AppendFrame` re-prunes, and `Query` caps at `limit` anyway. Not a correctness issue, but if strict cap enforcement matters for display, wrapping insert+prune in a transaction would close it.

### P2 — Cross-replica version monotonicity limitation (`modules/bot_api/send.go:1073`)

`cardVersionInLockWrite` documents that `GenSeq` is a process-local HiLo allocator, so version allocation order == commit order only within a single process. Multi-replica deployments can have a lower-version frame commit after a higher-version one, causing delta-sync to skip it permanently until full resync. This is explicitly acknowledged as a pre-existing limitation (not introduced by this PR) and documented in the function comment. Flagging for tracking — a future fix would need a channel-level total-order source (DB/Redis atomic counter) across all version carriers including richtext.

- `validate.go` walker: `seenIDs` enforcement for frame-unique ids, `selectAction` validation moved before type dispatch (校验面 ≥ 派发面), `inlineAction` now routed through `w.action` allowlist — closes the Input inlineAction surface that previously bypassed action validation.
- `Action.Submit` validation: id required, data must be object, registerID for uniqueness — all enforced at send-time so card/action can trust the stored frame.
- `NormalizeContentEdit` (`pkg/cardmsg/interactive.go:147`): runs Validate+Finalize on type-17 edits, producing canonical JSON. Non-card payloads pass through unchanged.
- Bot clear endpoint: ownership check (`senderID == robotID`), card payload verification via `IsCardRawPayload`, and proper fakeChannelID reconstruction for person channels.
- `revoke()` in `api.go` calls `DeleteByMessageID` for card messages (best-effort) — physical cleanup complementary to the clear endpoint's tombstone audit trail.
- Error codes: new codes in `pkg/errcode/message.go` and `pkg/errcode/bot_api.go` are specific and non-overlapping.

## Verdict: APPROVED

No P0/P1 blockers. The PR implements D10 revision history with solid security (anti-IDOR, membership gates, fail-closed inputs), correct CAS semantics (InnoDB row lock + bounded retry), and comprehensive test coverage. The three P2 items are non-blocking: one stack artifact, one cosmetic race in cap prune, and one well-documented pre-existing limitation.

an9xyz:
## B1 residual — fixed @ `99500140`

Thanks — the finding is correct. The revision query was enforcing a strict **subset** of the canonical read: membership + group/thread status (via the shared `authorizeCardChannelMember`) + lifecycle (`isCardMessageWithdrawn`), but not the `visibles` allowlist / user read-offset / channel offset / expiry that `respondSingleMessage` and my own `card/action` handler carry. A `visibles`-excluded or offset-truncated group member could read the full history (`full=1` → complete envelopes) though the canonical single read hides it.

**Fix (same "share the gate, don't re-derive it" direction as `authorizeCardChannelMember`):**
- Extracted the visibles / expiry / user-offset / channel-offset block out of `cardAction` into a new shared helper `cardCanonicalVisibleToViewer(msgM, loginUID) (bool, error)` (`api_card_action.go`). `card/action` now calls it (behavior unchanged — same layers, same order); `getCardRevisions` calls it right after the withdrawn gate.
- On a visibility miss the query returns an **empty list** (mirrors the withdrawn path — no existence leak), not an error.
- `person` channel keeps the same offset-skip as `card/action` (2-party DM: `visibles` always passes, covered by membership + lifecycle).

**Test:** `TestCardRevisionsCanonicalVisibility` (`api_card_revisions_test.go`) pins ① `visibles` excludes the viewer → empty list + no content leak, ② control: `visibles` includes the viewer → history returned, ③ user read-offset past the message (via `channelOffsetDB.insertOrUpdate`, shard-correct) → empty list + no content leak.

Local gates green: `go build ./...`, `golangci-lint` 0 issues, `i18n-extract-check` / `i18n-lint`, and the full `TestCardRevisions*` / `TestCardAction*` suites `-race` (message) plus `TestBotCard*` `-race` (bot_api).

Non-blocking notes all acknowledged (IDOR / precision / collation clean; the journal + promoted learning now record B1 as the third occurrence of "a derived surface must match the canonical read in *every* layer"). Ready for re-review.

yujiawei (approved):
# Code Review — PR #549 (octo-server)

**Head SHA:** `995001406ea2e0795283f9ef610363c2eb76561c` · **Base:** `main` (now un-stacked; #548 merged) · Reviewed the true delta `aaacb612..99500140` (merge-base with main is #548's commit `aaacb612`, so this is the clean D10-only diff). Security-sensitive review.

> Note: the review was requested against `766e465d`, but the branch has since moved to `99500140` (a rebase + a fix commit). This verdict is for the current head. `766e465d` diverges only by the visibility fix described below.

## 1. Spec compliance

**Spec: ✅**

Verified each D10 decision against the diff:

- **D10.1 side table** — `octo_message_card_revision` created in `modules/message/sql/20250708000002_message_legacy01.sql` (correct `octo_` prefix, InnoDB/utf8mb4, `idx_message(message_id,id)`). Shared-table pattern (`pkg/cardrevision.Store`) written by `bot_api`, read by `message`, mirroring `message_extra`. Columns match the spec (frame + plain + editor + time + tombstone + cleared_count + card_seq).
- **D10.2 append on the D6 edit path** — `modules/bot_api/send.go` appends a frame after the authoritative `content_edit` write, best-effort (append error is logged, edit still succeeds and the CMD still fires). `editor_uid` = the editing bot.
- **D10.3 cap 20** — `AppendFrame` prunes non-tombstone frames beyond `CapFrames=20` via a derived-table subquery (correctly sidesteps the MySQL "can't DELETE from a table referenced in a subquery" restriction). Tombstones are excluded from the cap. Pinned by `TestCardRevisionsCapAndDelete`.
- **D10.4 transient** — `cardmsg.Transient` / `TransientFromContentEdit`; transient frames update `content_edit` but skip history. I verified the `transient` top-level key survives `NormalizeContentEdit` (Validate/Finalize don't strip unknown top-level fields), so `TransientFromContentEdit(req.ContentEdit)` reads it correctly after normalization.
- **D10.5 query API** — `GET /v1/message/card/revisions` mounted on `/v1/message` (Auth → SharedUIDRateLimiter → Space). Reuses the extracted `authorizeCardChannelMember` gate shared with `card/action`.
- **D10.6 clear + tombstone** — `POST /v1/bot/message/card/revisions/clear` validates ownership (`sender == robotID`) before clearing; `Store.Clear` runs count→delete→insert-tombstone in a single transaction.
- **D10.7 revoke cleanup** — `modules/message/api.go` revoke path deletes revision rows after commit, before `SendRevoke` (correct ordering — a notify failure returning early won't skip the delete). Best-effort.

No over-build (no D12 manifest, no octo-im changes, no diffing — all correctly out of scope). No deviation from the agreed approach.

## 2. Code Quality

**Quality: Approved**

Data-flow traced across the trust boundary. The security posture is solid:

- **Anti-IDOR binding** is derived from the stored message row, not request-supplied channel params, in the shared `authorizeCardChannelMember` — identical to `card/action`. The refactor is behavior-preserving (I diffed the extracted helper against the original inline code line-by-line).
- **Membership + group/thread status** gates (`ExistMemberActive`, `groupStatusVisibleForAction`, deleted-thread rejection) are carried through to the read endpoint.
- **Withdrawn fallback** (`isCardMessageWithdrawn`) hides history for revoked / globally-deleted / user-locally-deleted cards even if rows survive — a genuine second defense layer independent of the best-effort revoke delete. Pinned by `TestCardRevisionsWithdrawnHidden`.
- **Enumeration resistance** — invalid/denied merged into single codes; non-existent and channel-mismatch are indistinguishable.

### Resolved blocker — canonical visibility parity (was the standing B1)

At the prior head, the revisions **read** endpoint enforced a strictly *narrower* visibility set than the canonical single-message read: it applied the member gate + withdrawn fallback but **not** the `visibles` whitelist, message expiry, or user/channel cleanup offsets that both `card/action` and `respondSingleMessage` apply. Because bot-sent card payloads can carry `visibles`, a card visible only to a subset of a channel's members could leak its full content history (`full=1` returns complete frame envelopes) to any other member.

This is **fixed at the current head**: the four visibility layers are extracted into `cardCanonicalVisibleToViewer` (`api_card_action.go`) and applied by **both** `cardAction` and `getCardRevisions`. `card/action` was refactored to call the same helper (behavior-preserving), so the two surfaces can no longer drift. Read-side failure is fail-closed (500); not-visible returns an empty list without leaking existence. Covered by the new `TestCardRevisionsCanonicalVisibility` (visibles-excluded member and offset-truncated member both get empty history, and content does not leak). Confirmed `go build ./...` clean on the affected packages and `pkg/cardmsg` unit tests green.

### Non-blocking notes (P2 / nits)

- **P2 — `content` column is `TEXT` (64 KiB).** Card envelopes are capped at `MaxPayloadBytes = 512 KiB`. This matches the existing `message_extra.content_edit` (also `TEXT`), so it's consistent with the current contract, but a >64 KiB card frame would be silently truncated by MySQL in non-strict mode (or error in strict mode) on append — and since append is best-effort, a strict-mode error would be swallowed as a logged history gap while `content_edit` keeps the full frame. If large approval cards are expected, consider `MEDIUMTEXT` for `content`. Not blocking (pre-existing pattern, best-effort surface).
- **P2 — tombstone rows are unbounded.** Tombstones are (correctly) excluded from the cap-20 prune and from revoke cleanup only deletes on revoke. A bot repeatedly clearing the same live card accumulates tombstone rows with no cap. Low severity (requires a misbehaving bot; audit intent argues for retention), but worth a bound or a note if clear can be called in a loop.
- **Nit — `cardmsg.Enabled()` is gated on the bot clear endpoint but not on the member read endpoint.** Harmless (with the feature off, no frames are ever written, so the read returns empty), but gating both would be more symmetric.
- **Migration has no `+migrate Down`.** Consistent with recent sibling migrations in `modules/message/sql/` (none carry Down), so this matches repo convention — noting only for completeness.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. The feature is complete and correctly scoped, the one prior blocking finding (visibility parity) is fixed and tested at the current head, and the remaining items are P2/nits that don't block merge.

## Items a human should confirm (security-sensitive)

- Confirm `MEDIUMTEXT` vs `TEXT` is acceptable for `content` given expected max card size (P2 above).
- Confirm the migration is executed with an online tool if the target is a large table / MySQL 5.7 (the D9 sibling migration already documents this; the new table is a `CREATE`, so it's a metadata-only op — low risk).

OctoBoooot (approved):
## Re-review (delta): feat(cardmsg): P2 card revision history (D10) (#549)

**Verdict:** Approve — the B1 blocker is fixed, byte-verified, and now under test. Conditional on `Test` (currently in progress); `mergeStateStatus` stays BLOCKED until my + Jerry-Xin's prior CRs clear at this head and Test goes green.

Delta `766e465d…995001406` (true D10 diff vs main is clean; the compare's behind:2 is post-#548-merge rebase drift, not new work).

### Resolved since last review
- ✓ **B1 (blocker) — fixed.** `getCardRevisions` now captures `msgM` and calls the new shared `cardCanonicalVisibleToViewer(msgM, loginUID)` (api_card_action.go:418) before returning any rows. That helper applies exactly the four gates my blocker named — `visiblesAllows` + expiration + user-offset (`channelOffsetDB.queryWithUIDAndChannel`) + channel-offset (`lookupChannelOffsetSeq`) — fail-closed (error/false → denied). My failing input is now closed: a `visibles=[A]` card returns empty history to non-A member B via `full=1`. Extracted as a shared helper so cardAction and the revisions query can't drift — the fix direction I suggested.
- ✓ **Test gap — closed.** `TestCardRevisionsCanonicalVisibility` asserts visibles-excluded → empty + no content leak, visibles-included control → history returned, and user-offset-truncated → empty + no leak. The security property that was previously unverified is now tested (this addresses Jerry-Xin's test-coverage blocker too).

### Praise
- The fix generalizes the parity gate instead of copy-pasting cardAction's inline checks — `cardCanonicalVisibleToViewer` is now the single source for both the action and query paths. This is the right response to a seam finding: remove the seam, don't patch one side.

- Verdict is APPROVE on the merits; the gate stays BLOCKED until the standing CRs (mine + Jerry-Xin's at 766e465d) re-clear at 995001406 and Test finishes green. Not mergeable on vote count alone.

Jerry-Xin (approved):
**APPROVE** (re-review @995001406) — B1 residual leak is fully closed at parity with `cardAction`; test added; CI green.

🔴 Blocking: none. Prior B1 residual is resolved.

💬 Non-blocking:
- `cardCanonicalVisibleToViewer` is now the single canonical visibility gate shared by both `cardAction` and `getCardRevisions` — nice. Keep any future 5th gate additions inside this one helper so the two paths can't drift again.
- `pkg/cardrevision.Query` runs `SELECT *` with a `dbBaseCreatedUpdated` placeholder to tolerate extra columns; fine, but a future column rename could silently drop into the void — a short comment on the mapping contract would help the next reader.

✅ Highlights:
- **B1 closed via shared helper.** `getCardRevisions` (api_card_revisions.go:64) now captures `msgM` and calls `m.cardCanonicalVisibleToViewer(msgM, loginUID)` — the exact same helper `cardAction` calls (api_card_action.go). The helper enforces all four previously-missing gates: visibles allowlist (`visiblesAllows`), expiration (`Expire`/`Timestamp`), user read-offset (`channelOffsetDB.queryWithUIDAndChannel`), and channel-offset watermark (`lookupChannelOffsetSeq`). Non-visible → empty list (no existence leak), query error → fail-closed 500. The prior reachable leak (member B pulling `visibles=[A]` card history via `?full=1`) is now rejected on both summary and full paths.
- **No new sibling seam.** The new `bot_api/card_revision.go` is a bot-token-authed *clear/tombstone* write (ownership guard `FromUID != robotID` → forbidden); it returns only a count, never revision content. `send.go` `AppendFrame` and `DeleteByMessageID` are writes. `GET /card/revisions → getCardRevisions` remains the sole revision-returning read entrypoint — no new read leak.
- **Test coverage added.** `TestCardRevisionsCanonicalVisibility` asserts a visibles-excluded member and an offset-truncated member both get an empty list AND `NotContains` the hidden frame content; plus the positive control. Withdrawn/member-gate tests retained.
- **Feature correctness holds** post-refactor: tombstone soft-delete + audit row, message-access-keyed authorization (no guessable revision_id IDOR), `message_id` as string (no float64 precision loss), migration collation `utf8mb4_general_ci` with `idx_message(message_id,id)` matching the `WHERE message_id=? ORDER BY id DESC` query.
- **No regression:** `cardAction`'s full gate stack (channel-binding, group/thread status, member, bot-sender, revoke/delete, idempotent claim, effective-frame, inputs, canonical visibility) is intact after the refactor.

modules/message/api_card_revisions.go
an9xyz: Fixed in `0e06f4b4` (impl commit of the rebased stack): extracted `cardCanonicalVisibleToViewer` (visibles / expiry / user-offset / channel-offset) out of `cardAction` and applied it in `getCardRevisions` after the membership + withdrawn gates — the query now enforces full canonical-read parity. New test `TestCardRevisionsCanonicalVisibility` pins the visibles-excluded and offset-truncated cases. You (OctoBoooot) subsequently re-APPROVED @`99500140`; resolving this stale thread.

PR #554, opened 2026-07-08
feat(messages_search): search file body and filter it from _source

aunknown:
## Summary

- Extends the messages_search endpoints to match keywords against extracted file body text (new `payload.file.content` field on OS file docs) alongside file name and caption.
- Filters `payload.file.content` and `payload.file.contentMeta` out of the OS `_source` on every Search call so the wire response stays small (the body averages 30–100 KB per doc).
- Chat-tab `/_search` is intentionally unchanged: its payload-type whitelist `[Text, MergeForward, RichText]` already excludes `File(8)`.

## Change map

DSL — `payload.file.content` added to the `multi_match` fields on every builder that can surface a `File(8)` doc, at default weight `^1` (no boost):

| # | File | Function |
|---|---|---|
| M1 | `modules/messages_search/search_files.go` | `buildSearchFilesDSL` |
| M2 | `modules/messages_search/search_all.go` | `buildSearchAllDSL` (fileClause) |
| M3 | `modules/messages_search/search_global_files.go` | `buildGlobalFilesDSL` |
| M4 | `modules/messages_search/search_global_messages.go` | `buildGlobalMessagesDSL` (fileClause) |

_source excludes — new helper in `modules/messages_search/dsl.go` (`fileContentSourceExcludes`) plus a `FetchSourceContext(...)` call on every `SearchService` in the module: `searchFiles`, `searchAll`, `searchMessages`, `searchMedia`, `searchGlobalFiles` (+ the single-channel fast path), `searchGlobalMessages` (+ the single-channel fast path).

`payload.file.name` keeps its `^2` boost so a filename hit still outranks a pure-body hit. Highlight fragments are drawn from the inverted index rather than `_source`, so the exclude has no effect on any highlight surface — the wire shape returned by the four hit builders is byte-identical for callers that only search name/caption.

## Load-bearing behaviour

- `FileHit` and `MessageHit` wire shapes are unchanged (no new fields, no boost changes).
- `pickSnippet` priority list is unchanged — file content never enters the snippet slot.
- `buildSearchAllHighlight` / `buildSearchMessagesHighlight` are unchanged — the exclude preserves highlight fragment generation because highlights read the inverted index.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./modules/messages_search/...` clean
- [x] `go test ./modules/messages_search/... -race -count=1` all pass, including 8 new tests:
  - `TestFileContentSourceExcludes` — helper returns correct excludes list
  - `TestFileContentSourceExcludes_WiredIntoSearchSource` — request-body layer pin so a silent drop of `.FetchSourceContext(...)` from any endpoint fails the test
  - `TestBuildSearchFilesDSL_KeywordIncludesContentField` (M1)
  - `TestBuildSearchAllDSL_FileClauseIncludesContentField` (M2)
  - `TestBuildGlobalFilesDSL_KeywordIncludesContentField` (M3)
  - `TestBuildGlobalMessagesDSL_FileClauseIncludesContentField` (M4)
  - `TestBuildSearchMessagesDSL_ExcludesContentField` — negative pin that the chat-tab surface does not reach into file content
  - `TestBuildSearchAllHighlight_ExcludesFileContent` — negative pin that highlight does not allocate a fragment on a field `pickSnippet` never reads

Each of M1–M4 tests also asserts that the content field carries no inline boost, matching the current DSL contract.

### DIFF
--- a/modules/messages_search/dsl.go
+++ b/modules/messages_search/dsl.go
@@ -462,3 +462,19 @@ func buildSearchAfterFromHit(hit *elastic.SearchHit, isRelevance bool) ([]any, b
 	ts := numericTo64(hit.Sort[0])
 	return []any{ts, msgID, subSeq}, true
 }
+
+// fileContentSourceExcludes returns the shared _source excludes context every
+// messages_search endpoint attaches to its OS Search call. The excluded fields
+// carry the Tika-extracted file body (payload.file.content, avg 30–100 KB per
+// doc) and its extraction metadata (payload.file.contentMeta); both participate
+// in the multi_match relevance score via the inverted index but never surface
+// on the wire — Q1 C keeps FileHit without a contentSnippet field, and Q4 D
+// keeps pickSnippet from promoting content into the snippet slot. Highlight
+// fragments are drawn from the inverted index rather than _source, so this
+// exclude does not affect highlight functionality.
+func fileContentSourceExcludes() *elastic.FetchSourceContext {
+	return elastic.NewFetchSourceContext(true).Exclude(
+		"payload.file.content",
+		"payload.file.contentMeta",
+	)
+}
--- a/modules/messages_search/dsl_test.go
+++ b/modules/messages_search/dsl_test.go
@@ -737,6 +737,123 @@ func extractSearchAllTypes(t *testing.T, raw []byte) []int {
 	return nil
 }
 
+// TestFileContentSourceExcludes pins the shared _source excludes context every
+// messages_search endpoint attaches to its OS Search call. Q6 A: excluding
+// payload.file.content and payload.file.contentMeta keeps the Tika-extracted
+// file body (30-100 KB/doc) out of the wire response while still allowing
+// content to participate in relevance scoring via the inverted index.
+func TestFileContentSourceExcludes(t *testing.T) {
+	fsc := fileContentSourceExcludes()
+	if fsc == nil {
+		t.Fatal("fileContentSourceExcludes returned nil")
+	}
+	src, err := fsc.Source()
+	if err != nil {
+		t.Fatalf("Source(): %v", err)
+	}
+	raw, _ := json.Marshal(src)
+	body := string(raw)
+	for _, want := range []string{
+		`"payload.file.content"`,
+		`"payload.file.contentMeta"`,
+		`"excludes"`,
+	} {
+		if !strings.Contains(body, want) {
+			t.Errorf("fileContentSourceExcludes missing %q in:\n%s", want, body)
+		}
+	}
+}
+
+// TestFileContentSourceExcludes_WiredIntoSearchSource pins that once the
+// helper is attached to an elastic.SearchSource the emitted _source clause
+// carries the excludes onto the wire request. Regression guard: any endpoint
+// that silently drops `.FetchSourceContext(fileContentSourceExcludes())` from
+// its svc chain would still make the multi_match tests pass but start leaking
+// the 30–100 KB file body back to clients — this test walks the shape a
+// production request actually sends to OS.
+func TestFileContentSourceExcludes_WiredIntoSearchSource(t *testing.T) {
+	ss := elastic.NewSearchSource().
+		Query(elastic.NewMatchAllQuery()).
+		FetchSourceContext(fileContentSourceExcludes())
+	src, err := ss.Source()
+	if err != nil {
+		t.Fatalf("Source(): %v", err)
+	}
+	raw, _ := json.Marshal(src)
+	body := string(raw)
+	// _source clause must be present with both excludes; the elastic library
+	// emits `_source: {excludes: [...]}` when FetchSourceContext is attached.
+	for _, want := range []string{
+		`"_source"`,
+		`"excludes"`,
+		`"payload.file.content"`,
+		`"payload.file.contentMeta"`,
+	} {
+		if !strings.Contains(body, want) {
+			t.Errorf("SearchSource missing %q in:\n%s", want, body)
+		}
+	}
+}
+
+// TestBuildSearchFilesDSL_KeywordIncludesContentField pins the M1 contract:
+// /_search_files keyword DSL includes payload.file.content in the multi_match
+// fields (Q5 A default weight ^1 lets name/caption still outrank a pure body
+// hit).
+func TestBuildSearchFilesDSL_KeywordIncludesContentField(t *testing.T) {
+	req := SearchFilesReq{ChannelType: channelTypeGroup, ChannelID: "g", Keyword: "report"}
+	q, _ := buildSearchFilesDSL(context.Background(), fallbackTestAnalyzer(), true, req, "g", "")
+	body := asJSONString(t, q.(interface{ Source() (any, error) }))
+	if !strings.Contains(body, `"payload.file.content"`) {
+		t.Errorf("_search_files keyword DSL must include payload.file.content:\n%s", body)
+	}
+	// Boost pin: content stays at default ^1, name keeps its ^2 boost so a
+	// filename hit still outranks a pure body hit.
+	if strings.Contains(body, `"payload.file.content^`) {
+		t.Errorf("_search_files content field must carry default weight (no ^N boost):\n%s", body)
+	}
+}
+
+// TestBuildSearchAllDSL_FileClauseIncludesContentField pins M2.
+func TestBuildSearchAllDSL_FileClauseIncludesContentField(t *testing.T) {
+	req := SearchMessagesReq{ChannelType: channelTypeGroup, ChannelID: "g", Keyword: "hello"}
+	q, _ := buildSearchAllDSL(context.Background(), fallbackTestAnalyzer(), true, req, "g", "")
+	body := asJSONString(t, q.(interface{ Source() (any, error) }))
+	if !strings.Contains(body, `"payload.file.content"`) {
+		t.Errorf("_search_all keyword file clause must include payload.file.content:\n%s", body)
+	}
+	if strings.Contains(body, `"payload.file.content^`) {
+		t.Errorf("_search_all content field must carry default weight (no ^N boost):\n%s", body)
+	}
+}
+
+// TestBuildSearchMessagesDSL_ExcludesContentField is the negative regression
+// pin for the Q3 A decision: the chat-tab surface /_search must NOT search
+// file content. Whitelist [Text, MergeForward, RichText] already excludes
+// payload.type=File(8), so the multi_match fields do not reference any
+// payload.file.* field.
+func TestBuildSearchMessagesDSL_ExcludesContentField(t *testing.T) {
+	req := SearchMessagesReq{ChannelType: channelTypeGroup, ChannelID: "g", Keyword: "hello"}
+	q, _ := buildSearchMessagesDSL(context.Background(), fallbackTestAnalyzer(), true, req, "g", "")
+	body := asJSONString(t, q.(interface{ Source() (any, error) }))
+	if strings.Contains(body, `"payload.file.content"`) {
+		t.Errorf("/_search chat-tab DSL must NOT search payload.file.content (Q3 A):\n%s", body)
+	}
+	if strings.Contains(body, `"payload.file.contentMeta"`) {
+		t.Errorf("/_search chat-tab DSL must NOT reference payload.file.contentMeta:\n%s", body)
+	}
+}
+
+// TestBuildSearchAllHighlight_ExcludesFileContent is the Q4 D negative pin:
+// pickSnippet does not promote payload.file.content into the snippet slot, so
+// requesting a highlight on that field would allocate a fragment that no
+// consumer reads.
+func TestBuildSearchAllHighlight_ExcludesFileContent(t *testing.T) {
+	body := asJSONString(t, buildSearchAllHighlight())
+	if strings.Contains(body, `"payload.file.content"`) {
+		t.Errorf("_search_all highlight must NOT include payload.file.content (Q4 D):\n%s", body)
+	}
+}
+
 func TestExtractSortValues(t *testing.T) {
 	ts, msg, score := extractSortValues([]any{float64(1717000000), float64(9876543210)}, false)
 	if ts != 1717000000 || msg != 9876543210 {
--- a/modules/messages_search/dsl_test.go
+++ b/modules/messages_search/dsl_test.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"testing"
@@ -791,6 +793,60 @@ func TestFileContentSourceExcludes_WiredIntoSearchSource(t *testing.T) {
 	} {
 		if !strings.Contains(body, want) {
 			t.Errorf("SearchSource missing %q in:\n%s", want, body)
+		}
+	}
+}
+
+// TestEveryClientSearchAttachesFileContentSourceExcludes is the structural
+// guard that every file-reachable OS Search call in the module attaches the
+// shared file-body excludes. Scanning source rather than mocking each handler
+// is the same pattern as no_zinc_fallthrough_test.go — it catches a whole
+// class of regressions (a future endpoint added without the exclude, or a
+// silent drop from an existing chain) that per-handler tests would miss.
+//
+// Contract: for every non-test .go file in this directory, the count of
+// `client.Search()` occurrences must equal the count of
+// `fileContentSourceExcludes()` occurrences. The two file-reachable svc
+// chains inside search_around.go are the direct motivation for this pin —
+// see PR #554 review — but the guard is deliberately module-wide so any
+// future file-returning endpoint stays covered without a plan update.
+func TestEveryClientSearchAttachesFileContentSourceExcludes(t *testing.T) {
+	entries, err := os.ReadDir(".")
+	if err != nil {
+		t.Fatalf("read dir: %v", err)
+	}
+	for _, e := range entries {
+		name := e.Name()
+		if e.IsDir() || !strings.HasSuffix(name, ".go") || strings.HasSuffix(name, "_test.go") {
+			continue
+		}
+		data, err := os.ReadFile(filepath.Clean(name))
+		if err != nil {
+			t.Fatalf("read %s: %v", name, err)
+		}
+		// Strip // line comments so doc references (e.g. a comment naming
+		// `client.Search()` for cross-reference) don't skew the count — only
+		// live call sites matter.
+		var clean strings.Builder
+		for _, line := range strings.Split(string(data), "\n") {
+			if idx := strings.Index(line, "//"); idx >= 0 {
+				line = line[:idx]
+			}
+			clean.WriteString(line)
+			clean.WriteByte('\n')
+		}
+		src := clean.String()
+		searches := strings.Count(src, "client.Search()")
+		if searches == 0 {
+			continue
+		}
+		excludes := strings.Count(src, "fileContentSourceExcludes()")
+		if excludes < searches {
+			t.Errorf("%s: %d `client.Search()` call(s) but only %d "+
+				"`fileContentSourceExcludes()` — every file-reachable Search "+
+				"call must attach the shared _source excludes so the "+
+				"extracted file body never leaks to the wire",
+				name, searches, excludes)
 		}
 	}
 }
--- a/modules/messages_search/search_all.go
+++ b/modules/messages_search/search_all.go
@@ -88,7 +88,8 @@ func (h *Handler) searchAll(c *wkhttp.Context) {
 			Routing(normID).
 			Query(dsl).
 			Size(size).
-			TrackTotalHits(false)
+			TrackTotalHits(false).
+			FetchSourceContext(fileContentSourceExcludes())
 		if req.Keyword != "" {
 			svc = svc.Highlight(buildSearchAllHighlight())
 		}
@@ -170,6 +171,7 @@ func buildSearchAllDSL(ctx context.Context, analyzer tokenAnalyzer, stopwordStri
 		fileClause := buildKeywordClauseFromAnalyzed(eff, useMSM,
 			"payload.file.name^2",
 			"payload.file.caption",
+			"payload.file.content",
 		)
 		b.Should(textClause, fileClause)
 		b.MinimumShouldMatch("1")
--- a/modules/messages_search/search_around.go
+++ b/modules/messages_search/search_around.go
@@ -163,7 +163,8 @@ func (h *Handler) fetchAnchor(ctx context.Context, client *elastic.Client, req S
 		Routing(normID).
 		Query(b).
 		Size(1).
-		TrackTotalHits(false)
+		TrackTotalHits(false).
+		FetchSourceContext(fileContentSourceExcludes())
 	svc = applySort(svc, "time_asc")
 	res, err := svc.Do(ctx)
 	if err != nil {
@@ -205,7 +206,8 @@ func (h *Handler) aroundDirection(ctx context.Context, client *elastic.Client, r
 			Routing(normID).
 			Query(dsl).
 			Size(size).
-			TrackTotalHits(false)
+			TrackTotalHits(false).
+			FetchSourceContext(fileContentSourceExcludes())
 		svc = applySort(svc, sortMode)
 		if len(searchAfter) > 0 {
 			svc = svc.SearchAfter(searchAfter...)
--- a/modules/messages_search/search_files.go
+++ b/modules/messages_search/search_files.go
@@ -105,7 +105,8 @@ func (h *Handler) searchFiles(c *wkhttp.Context) {
 			Routing(normID).
 			Query(dsl).
 			Size(size).
-			TrackTotalHits(false)
+			TrackTotalHits(false).
+			FetchSourceContext(fileContentSourceExcludes())
 		svc = applySort(svc, req.Sort)
 		if len(searchAfter) > 0 {
 			svc = svc.SearchAfter(searchAfter...)
@@ -151,6 +152,10 @@ func buildSearchFilesDSL(ctx context.Context, analyzer tokenAnalyzer, stopwordSt
 		clause, err := buildKeywordClauseGated(ctx, analyzer, stopwordStripEnabled, req.Keyword,
 			"payload.file.name^2",
 			"payload.file.caption",
+			// content: full extracted file body from Tika (indexer v1.12).
+			// Default weight ^1 lets name/caption matches still outrank a
+			// pure-body hit. See docs/messages-search/v1.12-file-content-mapping-integration.md.
+			"payload.file.content",
 		)
 		b.Must(clause)
 		analyzeErr = err
--- a/modules/messages_search/search_global_files.go
+++ b/modules/messages_search/search_global_files.go
@@ -99,7 +99,8 @@ func (h *Handler) searchGlobalFiles(c *wkhttp.Context) {
 			Index(h.cfg.OSReadAlias).
 			Query(dsl).
 			Size(size).
-			TrackTotalHits(false)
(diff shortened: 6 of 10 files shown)


OctoBoooot (changes requested):
## Review: search file body and filter it from _source (#554)

**Verdict:** Request changes — one real gap: `search_around.go` is the one File-returning endpoint that didn't get the `_source` exclude, so it ships the 30–100 KB file body from OS to the app on every window fetch. Not a client-facing leak (the typed projection never serialises the body — see below), but a bandwidth regression that contradicts this PR's own goal and is inconsistent with the 7 sibling call sites. Everything else is clean and well-reasoned.

First contribution to this repo — nice work; the change is careful (per-call excludes, weight rationale, the chat-tab exclusion note). The findings below are about completeness, not correctness of what you did write.

### Major
- `modules/messages_search/search_around.go:164` (`fetchAnchor`) and `:206` (`aroundDirection`) — both OS Search calls lack `FetchSourceContext(fileContentSourceExcludes())`. `buildAroundDSL`/`buildAnchorDSL` carry no payload-type whitelist (comment: "no keyword, no payload-type whitelist"), so the around window returns `File(8)` docs. Without the exclude, OS ships the full `payload.file.content` (30–100 KB/doc, per your summary) to the app layer for the anchor + both direction fetches — the app then drops it (the typed `Doc.Payload.File` has no content field), so it never reaches the client, but the OS→app bandwidth cost is exactly what this PR set out to avoid. 7 of the 8 File-returning Search calls got the exclude; this is the missed 8th. Fix: add `.FetchSourceContext(fileContentSourceExcludes())` to both builders here, matching the other endpoints.

### Praise
- The exclude is a bandwidth optimisation layered on top of an already-safe projection, and the safety holds independently: `FilePayload` (source.go) has no `Content`/`ContentMeta` field, so `json.Unmarshal(_source) → Doc` drops the body; `PayloadRaw` is projected only for RichText(14)/Card(17), never File(8); and the highlight builders (`buildSearchAllHighlight`/`buildSearchMessagesHighlight`) do NOT highlight `payload.file.content` — so a pure-body match yields no fragment and `pickSnippet` falls back to typed-payload text. Net: the body cannot reach the client on any endpoint. Verifying that meant the exclude is defence-in-depth + bandwidth, and the client contract stays safe even where it's missing — which is why search_around is a perf gap, not a leak. Good layering.
- Access scope is correct: `payload.file.content` is added inside the `Must`/keyword clause on the same bool query that applies `applyChannelAndRevoked` + `applySpaceIDScope` + `payload.type` filter — body-search rides the existing channel scope, so a user can only match body text in channels they can already read. No access widening.
- `^1` default weight (name/caption still outrank a pure-body hit) and the chat-tab `/_search` exclusion (File not in its `[Text,MergeForward,RichText]` whitelist) are both the right calls, and documented.

### Minor
- Consider a test asserting `search_around`'s Search call carries the exclude once you add it — the DSL tests cover the multi_match fields well, but the `_source` exclude itself (and its presence per-endpoint) isn't asserted anywhere, which is how the search_around omission slipped through. A small "every File-returning builder sets fileContentSourceExcludes" guard would lock the set.

### Note
- CI: Build/Vet/Lint green, Test in progress; check-sprint is governance (Sprint field, maintainer). Verdict is REQUEST_CHANGES on the search_around gap regardless of CI.

Jerry-Xin (changes requested):
Summary: The PR is in scope for `octo-server`, but it misses an existing `messages_search` OpenSearch path that can still fetch large `payload.file.content` values.

**🔴 Blocking**
- 🔴 Critical: `_search_around` still fetches full file bodies from OpenSearch. `fetchAnchor` builds a `client.Search()` request without `FetchSourceContext(fileContentSourceExcludes())` at `modules/messages_search/search_around.go:161`, and `aroundDirection` does the same at `modules/messages_search/search_around.go:203`. This endpoint’s window DSL explicitly has no payload-type whitelist at `modules/messages_search/search_around.go:246`, so file docs can be anchor/window hits. That leaves the 30-100 KB `payload.file.content` and `payload.file.contentMeta` fields on the OS response path this PR is meant to shrink. Please attach the same source excludes to both `_search_around` searches and add a regression test that covers production search-source construction for that endpoint.

**💬 Non-blocking**
- 🔵 Suggestion: `TestFileContentSourceExcludes_WiredIntoSearchSource` at `modules/messages_search/dsl_test.go:767` proves the helper works when manually attached, but it does not verify every production `SearchService` chain attaches it. The missed `_search_around` calls show that gap. A small shared builder/helper for `client.Search()` setup, or targeted tests for all endpoint request bodies, would make this harder to regress.

**✅ Highlights**
- The added `payload.file.content` field is wired into the file-capable keyword DSLs with no extra boost, preserving `payload.file.name^2` precedence.
- The response projection remains unchanged because `FilePayload` does not expose the new body field, and highlights/snippet priority were not expanded to include file body text.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #554 (octo-server)

## Summary

This PR extends the messages_search module to include Tika-extracted file body (`payload.file.content`) in the keyword relevance scoring for file-bearing endpoints (`_search_files`, `_search_all`, `_search_global_files`, `_search_global_messages`), while preventing the 30–100 KB body from being pulled back in `_source` via a shared `fileContentSourceExcludes()` helper. The chat-tab `/_search` correctly stays out of file content (Q3 A). Test coverage pins the DSL field inclusion, boost weights, highlight exclusion, and _source wire shape — well-structured regression guard.

One parity miss: `_search_around` was not updated to attach the excludes.

## Verification

- ✅ **DSL content inclusion** — `payload.file.content` added at default weight `^1` in `buildSearchFilesDSL` (`search_files.go:158`), `buildSearchAllDSL` (`search_all.go:174`), `buildGlobalFilesDSL` (`search_global_files.go:200`), `buildGlobalMessagesDSL` (`search_global_messages.go:242`). Name retains `^2` boost so filename hits still outrank pure-body hits.
- ✅ **Chat-tab exclusion** — `buildSearchMessagesDSL` (`search_messages.go:169`) whitelists only `[Text, MergeForward, RichText]` and searches only `text.content`, `richText.searchText`, `mergeForward.msgs.searchText`. No file content reference.
- ✅ **_source excludes wiring** — `fileContentSourceExcludes()` (`dsl.go:475`) attached via `FetchSourceContext` at all 8 `client.Search()` call sites in `search_all.go:92`, `search_files.go:109`, `search_messages.go:123`, `search_media.go:98`, `search_global_files.go:103` + `:295`, `search_global_messages.go:112` + `:400`.
- ✅ **Highlight safety** — `buildSearchAllHighlight()` (`search_all.go:182`) includes `payload.file.name` but not `payload.file.content`, preventing unnecessary fragment allocation on the large body field.
- ✅ **Snippet safety** — `pickSnippet` (`dsl.go:195`) and `fallbackSnippet` (`dsl.go:234`) both cap at `payload.file.name`; neither surfaces file content into the snippet slot.
- ✅ **Browse-mode correctness** — `buildSearchAllDSL` adds the fileClause only when `req.Keyword != ""` (`search_all.go:153`); browse mode (empty keyword) skips the keyword path entirely.
- ✅ **Test coverage** — 8 new test functions pin positive (M1–M4) and negative (Q3 A, Q4 D) contracts plus the helper shape and SearchSource wire form.

Static analysis only at head `7ec37f87b92729817a47be7acf177d85b772c92d`; build and tests not executed in this environment.

## Findings

No P0/P1 issues; one P2 parity miss below.

### P2 — `_search_around` missing `fileContentSourceExcludes()` (`dsl.go:475`, `search_around.go:161` + `:203`)

The shared `fileContentSourceExcludes()` helper is declared as "the _source excludes context every messages_search endpoint attaches to its OS Search call" but the two `client.Search()` calls in `search_around.go` — `fetchAnchor` (:161) and `aroundDirection` (:203) — were not updated. `buildAroundDSL` has no payload-type whitelist, so file messages appear in the around window. When a file message lands in the window, the full 30–100 KB `payload.file.content` is pulled from OS into the Go process unnecessarily.

This is not a data leak: the `MessageHit` struct has no content field, and `fallbackSnippet` only surfaces `payload.file.name`. The impact is wasted OS→Go wire bandwidth. Pre-existing behavior (no endpoint had excludes before this PR); the PR improves 8 of 10 call sites but missed the around pair.

```go
// Fix: add to both Search chains in search_around.go
svc := client.Search().
    Index(h.cfg.OSReadAlias).
    Routing(normID).
    Query(b).
    Size(1).
    TrackTotalHits(false).
    FetchSourceContext(fileContentSourceExcludes())  // <-- add
```

- `fileContentSourceExcludes()` returns a fresh `FetchSourceContext` per call — the elastic `FetchSourceContext` is a value-type config holder with no shared mutable state, so this is safe.
- `buildKeywordClauseFromAnalyzed` and `buildKeywordClauseGated` accept variadic field strings — adding `payload.file.content` at the end is a clean additive change.
- Test helpers `asJSONString`, `marshalDSL`, `buildGlobalReq`, `fallbackTestAnalyzer` all exist in sibling test files with matching signatures.
- `dispatchSingleFiles` (`search_global_files.go:289`) and `searchAllForGlobalFastPath` (`search_global_messages.go:394`) — both secondary search paths — are covered.

## Data-Flow Trace

| Consumer | Data | Source | Runtime path verified |
|---|---|---|---|
| `fileContentSourceExcludes()` | exclude list | hardcoded `payload.file.content`, `payload.file.contentMeta` | ✅ consumed by 8/10 Search chains |
| multi_match DSL fields | `payload.file.content` at `^1` | added in 4 DSL builders | ✅ only when `keyword != ""` |
| pickSnippet | highlight fragments | OS highlight response from inverted index | ✅ no `payload.file.content` field requested |
| fallbackSnippet | `payload.file.name` | `Doc.Payload.File.Name` from `_source` | ✅ name only, not content |
| MessageHit wire response | struct fields | `singleMessageHit` projection | ✅ no content field in struct |

- **C1 — Dual-path parity**: HIT. All `client.Search()` call sites in `messages_search` module were grepped; 8/10 have the excludes, `search_around.go` (2 calls) does not. Reported as P2 above.
- **C2 — Control-flow ordering / nesting**: CLEAR. The `FetchSourceContext` is a terminal chain method, not affected by call ordering. No nested reuse concerns.
- **C3 — Authorization boundary**: N/A. No auth/permission changes in this PR.
- **C4 — Authorization lifecycle**: N/A. No auth changes.
- **C5 — Build/runtime divergence**: CLEAR. The change is pure Go library configuration; no build/packaging divergence.
- **C6 — Governance/docs**: N/A. No governance or policy changes.

[Octo-Q] verdict: APPROVE — zero P1 blockers. One P2 parity miss (`_search_around` missing excludes) is non-blocking; the `_source` content is not leaked to clients via the response struct, only wasted on the OS→Go wire. Recommend fixing in a follow-up or this PR before merge.

yujiawei (dismissed):
# Code Review — PR #554 (octo-server)

**Head SHA:** `7ec37f87b92729817a47be7acf177d85b772c92d` · base `main` (merge-base `728e7ab7`)
Verified locally: `go build ./...`, `go vet ./modules/messages_search/...`, and `go test ./modules/messages_search/... -race -count=1` all pass (including the 8 new tests).

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly what the PR describes — nothing missing, nothing extra:

- **Added** `payload.file.content` (default weight `^1`, no boost) to the `multi_match` fields of the four builders that can surface a `File(8)` doc: `buildSearchFilesDSL` (search_files.go), `buildSearchAllDSL` fileClause (search_all.go), `buildGlobalFilesDSL` (search_global_files.go), `buildGlobalMessagesDSL` fileClause (search_global_messages.go).
- **Added** the shared `fileContentSourceExcludes()` helper (dsl.go) and wired `FetchSourceContext(...)` onto the OS Search call of the six file-surfacing endpoints (`searchFiles`, `searchAll`, `searchMessages`, `searchMedia`, `searchGlobalFiles` + single-channel fast path, `searchGlobalMessages` + single-channel fast path).
- **Correctly did NOT** touch the chat-tab `/_search` field list — its payload-type whitelist `[Text, MergeForward, RichText]` already excludes `File(8)`, and there is a negative test (`TestBuildSearchMessagesDSL_ExcludesContentField`) pinning that.
- `payload.file.name` keeps its `^2` boost, so a filename hit still outranks a pure-body hit — pinned by the "no `^N` boost on content" assertions in each M1–M4 test.

Wire shapes (`FileHit` / `MessageHit`), `pickSnippet`, and the highlight builders are unchanged. Test coverage is genuinely load-bearing: `TestFileContentSourceExcludes_WiredIntoSearchSource` walks the real request body, so a silent drop of any `.FetchSourceContext(...)` would fail CI rather than pass quietly.

## 2. Code Quality

**Quality: Approved**

The change is small, well-scoped, well-commented, and correctly tested. One non-blocking gap:

### P2 — `_search_around` still fetches the file body from OS (performance only, no leak)

`modules/messages_search/search_around.go` has two `client.Search()` calls — `fetchAnchor` (~L161) and `aroundDirection` (~L203) — that do **not** attach `FetchSourceContext(fileContentSourceExcludes())`. `buildAroundDSL` applies no payload-type whitelist, so a `_search_around` window can include `File(8)` docs, and for those the new 30–100 KB `payload.file.content` field is transferred from OS and JSON-decoded on every anchor/neighbor fetch — the exact transfer cost this PR set out to eliminate everywhere else.

**Not a data-leak / not a blocker:** the response projection struct `FilePayload` (source.go) has no `Content` field, so even the un-excluded body is dropped by `json.Unmarshal` before it can reach the wire; `_search_around` requests no highlight either. This is a pure OS→app transfer/parse waste, not an exposure. Recommend attaching the same source context to both around searches in a follow-up (or this PR) for consistency.

### Notes (no action required)

- **Not a recall bug.** `payload.file.content` is a real mapped field: the canonical index mapping (owned by the indexer, v1.12) defines `payload.file.content` as `text` with IK analyzers, matching `payload.text.content`. The `v1.8-opensearch-mapping.md` doc in this repo predates v1.12 and is not the live mapping — it should not be read as the field being absent. Separately, `dynamic: strict` is an index-time write constraint; `multi_match` simply ignores an unmapped field at query time and does not error, so there is no total-recall risk even during a mixed-version window.
- **No zero-value field appears in responses.** Because `FilePayload` has no `content` field at all, excluding it from `_source` cannot produce a `"content": ""` artifact for clients.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. The one finding (`_search_around` source-exclude coverage) is a P2 performance consistency improvement with no correctness or security impact, and does not block merge.

OctoBoooot (approved):
## Re-review (delta): search file body and filter it from _source (#554 @ f4efb3e1)

**Verdict:** Approve — the `search_around` exclude gap is fixed, and the author added a module-wide structural guard that prevents the whole class of regression. Reversing my prior REQUEST_CHANGES. Conditional on the in-progress CI; check-sprint (governance) stays until a maintainer sets the Sprint field.

Delta `7ec37f87…f4efb3e1`: 2 files (`search_around.go` +4/-2, `dsl_test.go` +56).

### Resolved since last review (byte-verified)
- ✓ **`search_around` missing exclude (major) — fixed.** Both Search sites now attach `FetchSourceContext(fileContentSourceExcludes())`: `fetchAnchor` (:167) and `aroundDirection` (:210). The 8th File-returning endpoint now matches the other 7 — I re-counted every non-test file in the module: all 7 have `count(client.Search()) == count(fileContentSourceExcludes())`, no remaining gap.
- ✓ **Structural guard added (was my minor #1) — and done the right way.** `TestEveryClientSearchAttachesFileContentSourceExcludes` scans every non-test `.go` in the package and asserts each file's `client.Search()` count is matched by `fileContentSourceExcludes()`. Source-scanning (rather than per-handler mocking) catches the entire regression class — a future file-returning endpoint added without the exclude fails CI automatically, no plan update needed. It even strips `//` comments so a doc reference to `client.Search()` doesn't skew the count. This converts the "enumerate every call site" discipline from reviewer-diligence into an enforced invariant.

### Praise
- The fix went past the specific finding to the general one: rather than just patching the two search_around calls, the author added the guard that makes "every file-reachable Search attaches the exclude" a test-enforced property. That's exactly the right response to a completeness/seam finding — close the instance and the class in one pass.

### Still verified correct (unchanged)
- The safety model I confirmed last head holds: the file body cannot reach the client on any endpoint regardless (typed `FilePayload` has no content field; `PayloadRaw` only projected for RichText/Card; highlight not configured on `payload.file.content`) — the exclude is bandwidth defense-in-depth, and now uniformly applied. Body-search rides the same channel+space-scoped bool query (no access widening).

### Note
- Not mergeable on this approval alone: CI must finish green and check-sprint (Sprint field) needs a maintainer. mergeStateStatus stays BLOCKED until then, and the other reviewers' prior CRs (Jerry-Xin) should re-clear at this head.

Jerry-Xin (approved):
Summary: The PR is in scope for `octo-server` and the implementation correctly extends file-searchable DSL paths while excluding large file body fields from `_source`.

💬 Non-blocking

🟡 Warning: `modules/messages_search/dsl_test.go:813` uses a source-text count of `client.Search()` vs `fileContentSourceExcludes()` as a structural guard. It works today, but it is broader and more brittle than the actual behavior contract: future non-file search call sites, helper wrappers, block comments, or string literals could fail the test for reasons unrelated to leaking file content. Consider centralizing SearchService construction behind a small helper or pinning emitted SearchSource JSON for the relevant handlers/builders instead.

✅ Highlights

- `modules/messages_search/search_files.go:152`, `search_all.go:171`, `search_global_files.go:197`, and `search_global_messages.go:239` add `payload.file.content` only to file-capable keyword clauses, preserving the `/ _search` chat-tab behavior.
- `modules/messages_search/dsl.go:475` cleanly centralizes `_source` excludes for `payload.file.content` and `payload.file.contentMeta`.
- Response projection remains stable because `FilePayload` does not deserialize those fields and snippets still ignore file body content.
- Verified with `go test ./modules/messages_search/...`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #554 (octo-server)

## Summary

This PR adds Tika-extracted file body (`payload.file.content`) to the keyword search surface for file-reachable endpoints (`_search_files`, `_search_all`, `_search_global_files`, `_search_global_messages`) while ensuring the large field (30–100 KB/doc) never leaks onto the wire via a shared `_source` exclude helper `fileContentSourceExcludes()`. The chat-tab `/_search` endpoint correctly stays out of file-content search (Q3 A: type whitelist `[1,11,14]` excludes file messages). Highlight configuration and `pickSnippet`/`fallbackSnippet` are not extended to cover the content field (Q4 D), so no fragment is allocated that no consumer reads. The structural guard test `TestEveryClientSearchAttachesFileContentSourceExcludes` is a strong regression net — it scans all non-test `.go` files in the module and enforces 1:1 parity between `client.Search()` call sites and `fileContentSourceExcludes()` attachments.

## Verification

- ✅ **client.Search() ↔ fileContentSourceExcludes() parity** — counted 10 production `client.Search()` call sites and 10 `fileContentSourceExcludes()` attachments across 7 files (`search_all.go`:1, `search_around.go`:2, `search_files.go`:1, `search_global_files.go`:2, `search_global_messages.go`:2, `search_media.go`:1, `search_messages.go`:1). Exact 1:1 match.
- ✅ **DSL field coverage** — `payload.file.content` added to multi_match in `buildSearchFilesDSL` (`search_files.go:158`), `buildSearchAllDSL` (`search_all.go:174`), `buildGlobalFilesDSL` (`search_global_files.go:200`), `buildGlobalMessagesDSL` (`search_global_messages.go:242`). Correctly absent from `buildSearchMessagesDSL` (chat-tab, Q3 A) and `buildAroundDSL`/`buildAnchorDSL` (filter-only, no keyword search).
- ✅ **Content field weight** — all four DSL builders add `payload.file.content` at default `^1`, while `payload.file.name` retains `^2`. Filename matches outrank pure-body hits.
- ✅ **Highlight exclusion** — `buildSearchAllHighlight` (`search_all.go:182`) and `buildSearchMessagesHighlight` (`search_messages.go:207`) do not include `payload.file.content`. `pickSnippet` (`dsl.go:195`) priority list ends at `payload.file.name`. No fragment allocated for content.
- ✅ **Response DTO** — `FileHit` struct (`search_files.go:27`) has no `contentSnippet` field. Content never surfaces in the API response shape.
- ✅ **_source exclude semantics** — `elastic.NewFetchSourceContext(true).Exclude(...)` correctly keeps `_source` enabled while excluding the two named fields. OpenSearch highlights use the inverted index, not `_source`, so excluding from `_source` does not affect highlight functionality.
- ✅ **Defense-in-depth** — `search_media.go` and `search_messages.go` don't search file content in their DSL but still attach the `_source` exclude. Correct belt-and-suspenders approach.

Static analysis only at head `f4efb3e182783ad8ab03e0ff0146fd55169b4eb2`; build and tests not executed in this environment.

## Findings

- **Data flow — content field to inverted index**: `payload.file.content` is written by the Tika indexer (v1.12). The multi_match query scores against the inverted index, not `_source`. Documents indexed before v1.12 simply lack the field — OpenSearch treats missing fields as non-contributing to score, no errors.
- **Structural guard test correctness** (`dsl_test.go:815–851`): strips `//` line comments before counting, so doc-comment references to `client.Search()` don't skew the parity check. Block comments (`/* */`) are not stripped, but this makes the test more conservative (false-positive risk, not false-negative). No block comments containing `client.Search()` exist in production files.
- **`contentMeta` vs `content`**: `payload.file.contentMeta` (extraction metadata) is excluded from `_source` but not added to multi_match. Correct — metadata is not searchable text.
- **`buildKeywordClauseGated` and `buildKeywordClauseFromAnalyzed`**: both accept variadic `fields ...string` and pass them to multi_match. Adding one more field is a trivial, correct extension.
- **search_around.go**: filter-only queries (no keyword search), so content field is correctly absent from DSL. The `_source` exclude is defense-in-depth.

## Blind Spot Checklist

- **C1 (dual-path parity)**: Clear — all 10 `client.Search()` call sites have matching `fileContentSourceExcludes()`. Structural guard test enforces this going forward.
- **C2 (control-flow ordering)**: N/A — no nested reuse or double-application of the exclude helper.
- **C3 (authorization boundary)**: N/A — no authorization changes in this PR.
- **C4 (authorization lifecycle)**: N/A — no auth changes.
- **C5 (build ≠ runtime)**: Clear — `_source` exclude is an OpenSearch request-level parameter applied at query execution time. No build-time vs runtime divergence.
- **C6 (governance/docs)**: N/A — no governance or policy changes.

## Verdict: APPROVED

Clean, well-structured PR with correct data flow, comprehensive test coverage (positive pins, negative pins, and a structural guard), and defense-in-depth `_source` exclusion on all search call sites. No issues found.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. Data flow verified: content field correctly added to file-reachable DSLs, excluded from chat-tab DSL and all highlight configs, _source exclude attached to all 10 client.Search() calls with structural guard test enforcing parity.

yujiawei (approved):
# Code Review — PR #554 (octo-server)

**Head SHA:** `f4efb3e182783ad8ab03e0ff0146fd55169b4eb2` · base `main` (merge-base `728e7ab7`)
Verified locally against the head SHA: `go build ./modules/messages_search/...`, `go vet`, and `go test ./modules/messages_search/... -race -count=1` all pass, including the new tests.

This review re-runs against the current head, which now applies the `_source` exclude to `_search_around` — the gap flagged in the previous round. Cross-checked with an independent line-level pass and a long-context whole-module pass.

## 1. Spec compliance

**Spec: ✅**

The diff does exactly what the description states, no more:

- **Body added to relevance on every file-returning surface.** `payload.file.content` is added to the keyword `multi_match` fields in all four builders: `buildSearchFilesDSL` (`search_files.go:158`), `buildSearchAllDSL` fileClause (`search_all.go:174`), `buildGlobalFilesDSL` (`search_global_files.go:200`), `buildGlobalMessagesDSL` fileClause (`search_global_messages.go:242`). Each keeps `payload.file.name^2` and adds content at default `^1`.
- **Body excluded from `_source` on every live `Search()` call.** All 10 `client.Search()` call sites in the module attach `FetchSourceContext(fileContentSourceExcludes())` (`dsl.go:475`): `search_all`, `search_files`, `search_media`, `search_messages`, both `search_global_files` paths, both `search_global_messages` paths, and both `search_around` paths. The `search_around` pair is the fix versus the prior head.
- **Chat-tab `/_search` intentionally unchanged.** Its type whitelist `[Text(1), MergeForward(11), RichText(14)]` (`search_messages.go:191`) excludes `File(8)`, so no `payload.file.*` field enters that `multi_match`. `TestBuildSearchMessagesDSL_ExcludesContentField` pins this negatively.

No over-build (no new wire fields, no boost changes, no flags), no missing items, no deviation from the stated approach.

- Missing: none
- Extra/out-of-scope: none
- Deviation: none

## 2. Code Quality

**Quality: Approved**

No P0/P1. The change is mechanical and well-tested. `FilePayload` (`source.go:123`) has no `Content` field and `FileHit` is projected only from name/size/ext/url — so the `_source` exclude cannot break hit construction, and the wire shape is byte-identical for name/caption callers. Highlight config (`buildSearchAllHighlight`, `search_all.go:182`) references only `payload.file.name`, and `pickSnippet` (`dsl.go:195`) never reads content — highlights come from the inverted index, so the exclude has no highlight impact. Data flow verified end to end.

Non-blocking items (P2 — do not gate merge):

- **P2 — "unboosted content ⇒ metadata always wins" is asserted but not enforced.** The body field shares one `cross_fields multi_match` with `payload.file.name^2` / `payload.file.caption` (`keyword_query.go:94`). Default `^1` is not a demotion: under BM25 a strong body match (term frequency / IDF) can still compete with, or occasionally outrank, a metadata match. The unit tests only assert the field string carries no `^N` — they do not prove ranking order. This is likely acceptable (surfacing body hits is the point), but if metadata dominance is a real requirement, use an explicit sub-`^1` body boost or a separate lower-boost clause, and add an OpenSearch ranking fixture. At minimum, soften the code comments so they don't claim a ranking guarantee the query can't provide.

- **P2 — the module-wide guard test is coarser than it reads.** `TestEveryClientSearchAttachesFileContentSourceExcludes` (`dsl_test.go:813`) compares per-file *string counts* of `client.Search()` vs `fileContentSourceExcludes()`. A file passes as long as counts match, so one `Search()` chain missing the exclude could be masked by an unrelated second mention of the helper in the same file; it would also wrongly fire on a future legitimately body-free `Search()`. Consider a `go/ast` walk that checks each `client.Search()` builder chain actually contains the `FetchSourceContext` call. (Two independent review passes converged on this same point.)

- **P2 — `FetchSourceContext` overwrites rather than merges.** The olivere/elastic setter replaces any previously-set source context. Today nothing else sets one, so this is fine; but a future `.FetchSourceInclude(...)` / `.FetchSourceContext(...)` added to one of these chains would silently clobber the exclude and re-leak the 30–100 KB body. A shared query-builder wrapper (or the AST guard above) would make this regression-proof.

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. The previous blocker (`_search_around` shipping full file bodies from OS) is resolved — every file-reachable `Search()` now carries the exclude. The three P2 items are maintainability/robustness improvements, not merge blockers.

## 4. Suggestions

- If body-vs-metadata ranking order matters to product, add one integration/ranking fixture (a doc whose body matches vs a doc whose name matches) rather than relying on the `^` string pins.
- Replace the count-based guard with an AST-based one so it can't be fooled by count coincidence and won't misfire on future body-free `Search()` calls.

## 5. Additional observations

- Body-only file hits surface with a filename snippet (or none) since content is neither highlighted nor promoted into the snippet slot — the match isn't explained to the user. That's a deliberate product decision here (`pickSnippet` unchanged), but worth confirming the client UX is acceptable when a file matches purely on body text.
- Runtime cost of scoring a 30–100 KB field in the `multi_match` (cluster CPU/latency at scale) is not covered by these unit tests and can't be assessed from the diff alone; worth a glance on a representative index before/after if file corpora are large.

PR #558, opened 2026-07-09
fix(thread): 创建子区后为创建者补 thread ext 行，全端根治创建者看不到自己新建子区

### LINKED ISSUE #557: 创建子区后创建者自己看不到（关注 Tab 不出现）—— 后端应为创建者无条件补 thread ext 行，全端根治
## 现象

在 **iOS** 客户端手工创建子区（thread）后，新建的子区不出现在创建者自己的「关注」Tab 里。web 端已通过 octo-web #293（`useFollowSidebar` 客户端补 follow）缓解，但：

1. 该修复是**纯前端 web** 的，iOS / Android / PC 无对应逻辑；
2. 即便在 web 上，#293 也只在**父频道已关注**时才补 follow，覆盖不到「创建者未关注父频道」这一场景。

因此这是一个跨端普遍存在的后端行为缺口，需要在 server 侧根治。

## 根因

子区能否进入某用户的关注 Tab，唯一依据是 `user_conversation_ext` 里是否有该子区的行。落这行的两条路径：

1. **创建时**：`modules/thread/service.go` 的 `CreateThread` —— 只给创建者写 `thread_member`（IM 发消息权限）+ 把创建者加进 IM 频道 subscribers，**没有为创建者写任何 `user_conversation_ext` 行**。
2. **fanout**：`modules/conversation_ext/service.go` 的 `OnThreadCreated` —— 只给满足 `target_type=group AND auto_follow_threads=1`（即**已明确关注父频道**）的成员补子区行。**创建者不被特殊对待**。

结论：**只要创建者本人没有关注父频道，无论用哪个客户端建子区，这个子区都不会进创建者自己的关注 Tab。**

## 期望行为（propriety）

创建者对自己刚创建的子区，天然应当可见——用户执行了「创建」这个动作，系统就应让结果对他可见，不应依赖他事先关注了父频道。

## 建议修复

在 server 的 thread 创建 hook（`CreateThread` / 触发 `OnThreadCreated` 的同一提交路径）里，**无条件为创建者补一条子区 `user_conversation_ext` 行**（等价于把创建者视为该子区的 follower），与现有 fanout 写入走同一套 version bump / ext 落行逻辑，保证幂等。

优点：一处改动，iOS / web / PC / Android 全端统一生效，不必每个客户端各修一遍。

## 注意事项

- 只给**创建者本人**落子区行，**不要**同时清父群的 `group_unfollowed` flag 或把父群拉回关注（这是 octo-web #293 第一版被 reviewer 拒的 P1：不应因创建子区这个动作，副作用地把用户显式取关过的父群拖回关注 Tab）。
- 落行需带正确的 space_id / follow_version，与 `OnThreadCreated` 现有写入口径一致，保证幂等（重复触发不产生脏行）。
- 建议补回归测试：创建者未关注父群时创建子区 → 创建者关注 Tab 出现该子区；且父群 `group_unfollowed` flag 不被改动。

## 关联

- octo-web #293（web 端缓解，已 merged）
- octo-web #292（web 端 issue，已 closed）

pkuWMH:
## 问题

创建子区（thread）后，新建的子区不出现在**创建者自己**的「关注」Tab 里——iOS 上复现，实际是全端普遍问题。

octo-web #293 是纯前端 web 缓解（`useFollowSidebar` 客户端补 follow），iOS/Android/PC 无对应逻辑；且 web 方案也只在**父频道已关注**时才补，覆盖不到「创建者未关注父频道」。因此需要在 server 侧根治。

Closes #557

## 根因

子区能否进关注 Tab，唯一依据是 `user_conversation_ext` 里有没有该子区行。落这行的两条路径：

1. `modules/thread/service.go` 的 `CreateThread` —— 只给创建者写 `thread_member`（IM 发消息权限）+ 加进 IM 频道 subscribers，**从不给创建者写 `user_conversation_ext` 子区行**。
2. `modules/conversation_ext/service.go` 的 `OnThreadCreated` fanout —— 只服务 `target_type=group AND auto_follow_threads=1`（**已关注父频道**）的成员，**创建者不被特殊对待**。

结论：**只要创建者本人没关注父频道，任何客户端建子区都看不到自己新建的子区。**

## 修复

新增 `conversation_ext.FollowThreadForCreator(uid, spaceID, threadChannelID)`：
- 只 bump follow_version + upsert 子区 ext 行（`targetTypeThread`）
- **刻意不清父群 `group_unfollowed` flag**——绝不因「建子区」这个动作 re-follow 用户显式取关过的父频道（这正是 octo-web #293 v1 被 reviewer 拒的那个 P1，此处从设计上规避）
- 幂等（`upsertTx` 对同 key 是 no-op）；不做 thread 鉴权（创建者对自己刚建的子区天然授权）
- 锁序与 `FollowThread`/`UpdateSort` 一致（先 bump version 再写 ext）

`CreateThread` 在现有 `OnThreadCreated` fanout **之后** best-effort 调用（守 `convSvc != nil && SpaceID != ""`，失败只 `Warn` 不回滚，与 `OnThreadCreated` 同风格）。thread HTTP handler 经 `spacepkg.GetSpaceID(c)` 把 spaceID 透传进 `CreateThreadReq`；空 spaceID 被 best-effort 守卫容忍，不影响建子区主流程。

**server 侧一处修复，iOS/web/PC/Android 全端统一生效。**

## 改动文件（4）

- `modules/conversation_ext/service.go` — 新增 `FollowThreadForCreator`（+30）
- `modules/thread/service.go` — `CreateThreadReq` 加 `SpaceID` 字段；`OnThreadCreated` 后补调用（+16）
- `modules/thread/api.go` — 透传 `SpaceID: spacepkg.GetSpaceID(c)`（+2）
- `modules/conversation_ext/service_test.go` — 5 个集成测试（+75）

## 测试

- `go build ./...` ✅
- `gofmt` ✅（4 个改动文件均干净）
- `go test -tags integration ./modules/conversation_ext/...` ✅ 全过（含 5 个新测试，本地起 MySQL + 跑模块迁移验证）
- 关键回归测试 **`TestService_FollowThreadForCreator_DoesNotClearParentUnfollowed`**：断言创建者此前显式取关父群（`group_unfollowed=1`）后建子区，子区 ext 行被写入、且父群 `group_unfollowed` **仍为 1**（未被 re-follow）
- 其余新测试：写入子区行 / 幂等 / 空 uid 校验 / 非法 channelID 校验
- `modules/thread/...` 编译通过（`go test -c` OK）；其 runtime 测试需完整 `testutil.NewTestServer` 环境（本地缺 Redis/WuKongIM，属既有环境限制，与本改动无关），故在 `conversation_ext` 侧充分覆盖该方法。

未削弱或删除任何既有测试。

## 关联

- octo-web #293（web 端缓解，已 merged）
- octo-web #292（web 端 issue，已 closed）

### DIFF
--- a/modules/conversation_ext/service.go
+++ b/modules/conversation_ext/service.go
@@ -977,6 +977,36 @@ func (s *Service) FollowThread(uid, spaceID, threadChannelID string) error {
 	})
 }
 
+// FollowThreadForCreator writes the creator's own thread ext row when a thread
+// is created, so the creator always sees their new thread in the follow tab —
+// regardless of whether they follow the parent channel.
+//
+// Unlike FollowThread, it deliberately does NOT clear the parent group's
+// group_unfollowed flag: creating a thread must never silently re-follow a
+// parent channel the user explicitly unfollowed (issue #557 / prior web-PR P1).
+// It only upserts the thread ext row. Idempotent (INSERT ... on the same key is
+// a no-op via upsertTx). No thread auth check: the creator is inherently
+// authorized for the thread they just created.
+func (s *Service) FollowThreadForCreator(uid, spaceID, threadChannelID string) error {
+	if err := validateBase(uid, spaceID); err != nil {
+		return err
+	}
+	if _, _, err := parseThreadChannelID(threadChannelID); err != nil {
+		return err
+	}
+	return s.withTx("FollowThreadForCreator", func(tx *dbr.Tx) error {
+		// bump version before ext write, same lock order as FollowThread/UpdateSort.
+		if _, err := BumpFollowVersionTx(tx, uid, spaceID); err != nil {
+			return fmt.Errorf("FollowThreadForCreator bump version: %w", err)
+		}
+		// Upsert ONLY the thread ext row. Do NOT touch the parent group flag.
+		if err := upsertTx(tx, uid, spaceID, targetTypeThread, threadChannelID, ConvExtFields{}); err != nil {
+			return fmt.Errorf("FollowThreadForCreator upsert thread: %w", err)
+		}
+		return nil
+	})
+}
+
 // ---------------------------------------------------------------------------
 // UnfollowThread — delete thread ext row only
 // ---------------------------------------------------------------------------
--- a/modules/conversation_ext/service_test.go
+++ b/modules/conversation_ext/service_test.go
@@ -815,6 +815,81 @@ func TestService_FollowThread_ParentGroupNotPreviouslyUnfollowed_StillCreatesThr
 }
 
 // ---------------------------------------------------------------------------
+// FollowThreadForCreator — issue #557
+// ---------------------------------------------------------------------------
+
+// 创建者建子区后必须能在关注 tab 看到该子区：写 thread ext 行。
+func TestService_FollowThreadForCreator_CreatesThreadRow(t *testing.T) {
+	svc := newServiceForTest(t)
+	const uid, space, grp = "u1", "s1", "557-1"
+	threadChannelID := grp + threadSeparator + "thr-1"
+
+	require.NoError(t, svc.FollowThreadForCreator(uid, space, threadChannelID))
+
+	threadRow, err := svc.db.Get(uid, space, targetTypeThread, threadChannelID)
+	require.NoError(t, err)
+	assert.NotNil(t, threadRow, "创建者的 thread ext 行应被写入")
+}
+
+// 关键回归守卫（issue #557 / 前一版 web PR P1 被拒的原因）：
+// 若创建者此前显式取关了父群（group_unfollowed=1），建子区绝不能把它清回 0。
+// FollowThreadForCreator 只写 thread ext 行，父群 follow 状态必须原样保留。
+func TestService_FollowThreadForCreator_DoesNotClearParentUnfollowed(t *testing.T) {
+	svc := newServiceForTest(t)
+	const uid, space, grp = "u1", "s1", "557-2"
+	threadChannelID := grp + threadSeparator + "thr-2"
+
+	// 前置：创建者显式取关父群 → group_unfollowed=1。
+	require.NoError(t, svc.UnfollowChannel(uid, space, grp))
+	m, err := svc.db.Get(uid, space, targetTypeGroup, grp)
+	require.NoError(t, err)
+	require.NotNil(t, m)
+	require.Equal(t, int8(1), m.GroupUnfollowed)
+
+	require.NoError(t, svc.FollowThreadForCreator(uid, space, threadChannelID))
+
+	// 父群 group_unfollowed 必须仍为 1（未被 re-follow）。
+	parentRow, err := svc.db.Get(uid, space, targetTypeGroup, grp)
+	require.NoError(t, err)
+	require.NotNil(t, parentRow)
+	assert.Equal(t, int8(1), parentRow.GroupUnfollowed,
+		"FollowThreadForCreator 绝不能清父群 group_unfollowed（不得因建子区 re-follow 显式取关的父 channel）")
+
+	// thread ext 行仍应写入。
+	threadRow, err := svc.db.Get(uid, space, targetTypeThread, threadChannelID)
+	require.NoError(t, err)
+	assert.NotNil(t, threadRow, "创建者的 thread ext 行应被写入")
+}
+
+// 幂等：重复调用不报错，且不产生重复行 / 状态漂移。
+func TestService_FollowThreadForCreator_Idempotent(t *testing.T) {
+	svc := newServiceForTest(t)
+	const uid, space, grp = "u1", "s1", "557-3"
+	threadChannelID := grp + threadSeparator + "thr-3"
+
+	require.NoError(t, svc.FollowThreadForCreator(uid, space, threadChannelID))
+	require.NoError(t, svc.FollowThreadForCreator(uid, space, threadChannelID))
+
+	threadRow, err := svc.db.Get(uid, space, targetTypeThread, threadChannelID)
+	require.NoError(t, err)
+	assert.NotNil(t, threadRow)
+}
+
+// 输入校验对齐 FollowThread。
+func TestService_FollowThreadForCreator_EmptyUID(t *testing.T) {
+	svc := newServiceForTest(t)
+	err := svc.FollowThreadForCreator("", "s1", "grp-1____thr-1")
+	require.Error(t, err)
+}
+
+func TestService_FollowThreadForCreator_InvalidChannelID(t *testing.T) {
+	svc := newServiceForTest(t)
+	err := svc.FollowThreadForCreator("u1", "s1", "invalid-no-separator")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "thread_channel_id")
+}
+
+// ---------------------------------------------------------------------------
 // UnfollowThread happy path
 // ---------------------------------------------------------------------------
 
--- a/modules/thread/api.go
+++ b/modules/thread/api.go
@@ -16,6 +16,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
 	"github.com/Mininglamp-OSS/octo-server/pkg/i18n"
 	"github.com/Mininglamp-OSS/octo-server/pkg/i18n/codes"
+	spacepkg "github.com/Mininglamp-OSS/octo-server/pkg/space"
 	"go.uber.org/zap"
 )
 
@@ -229,6 +230,7 @@ func (t *Thread) createThread(c *wkhttp.Context) {
 
 	resp, err := t.service.CreateThread(&CreateThreadReq{
 		GroupNo:              groupNo,
+		SpaceID:              spacepkg.GetSpaceID(c),
 		Name:                 req.Name,
 		CreatorUID:           loginUID,
 		CreatorName:          loginName,
--- a/modules/thread/api.go
+++ b/modules/thread/api.go
@@ -158,7 +158,7 @@ func classifyThreadError(err error) codes.Code {
 
 // Route 注册路由
 func (t *Thread) Route(r *wkhttp.WKHttp) {
-	threads := r.Group("/v1/groups/:group_no/threads", t.ctx.AuthMiddleware(r))
+	threads := r.Group("/v1/groups/:group_no/threads", t.ctx.AuthMiddleware(r), spacepkg.SpaceMiddleware(t.ctx))
 	{
 		threads.POST("", t.createThread)
 		threads.GET("", t.listThreads)
--- a/modules/thread/api_space_wiring_test.go
+++ b/modules/thread/api_space_wiring_test.go
@@ -1,0 +1,38 @@
+package thread
+
+import (
+	"bytes"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestCreateThread_RouteMountsSpaceMiddleware 是路由接线回归测试，堵住 issue #557 复发。
+// 建子区路由 POST /v1/groups/:group_no/threads 必须挂 SpaceMiddleware，否则 createThread
+// 里的 spacepkg.GetSpaceID(c) 恒返回 ""、req.SpaceID 恒空，FollowThreadForCreator 永不触发，
+// 创建者在关注 tab 看不到自己刚建的子区。
+//
+// 服务层单测可直接填非空 SpaceID，掩盖这处接线缺口，所以必须走真实注册的路由来验证。
+// 手法：带一个当前用户并不属于的 space_id 打真实路由。挂了 SpaceMiddleware 时，中间件读到
+// space_id 后校验成员身份，非成员被拦成 403（响应体是中间件特有的“无权访问该 Space”）；
+// 若该路由没挂 SpaceMiddleware，请求会穿透到 createThread（active 群不会返回该 403，也不会
+// 出现该响应体），断言随即失败，从而锁死这处接线。
+func TestCreateThread_RouteMountsSpaceMiddleware(t *testing.T) {
+	s, ctx := setupTestData(t)
+	groupNo := createTestGroup(t, ctx)
+
+	body := util.ToJson(map[string]interface{}{"name": "接线校验子区"})
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("POST", "/v1/groups/"+groupNo+"/threads?space_id=sp_not_member", bytes.NewReader([]byte(body)))
+	req.Header.Set("token", testutil.Token)
+	s.GetRoute().ServeHTTP(w, req)
+
+	assert.Equal(t, http.StatusForbidden, w.Code,
+		"建子区路由必须挂 SpaceMiddleware：带 space_id 的非成员请求应被拦成 403")
+	assert.Contains(t, w.Body.String(), "无权访问该 Space",
+		"403 必须来自 SpaceMiddleware 的成员校验（响应体特征），而非其他路径")
+}
--- a/modules/thread/service.go
+++ b/modules/thread/service.go
@@ -106,6 +106,7 @@ func NewService(ctx *config.Context) IService {
 // CreateThreadReq 创建子区请求
 type CreateThreadReq struct {
 	GroupNo              string
+	SpaceID              string // 创建者所在 space，用于给创建者写 thread ext 行（关注 tab）
 	Name                 string
 	CreatorUID           string
 	CreatorName          string
@@ -307,6 +308,21 @@ func (s *Service) CreateThread(req *CreateThreadReq) (*ThreadResp, error) {
 				zap.String("groupNo", req.GroupNo),
 				zap.String("shortID", shortID),
 				zap.Error(err))
+		}
+	}
+
+	// 创建者本人始终应在关注 tab 看到刚建的子区，与是否关注父 channel 无关。
+	// OnThreadCreated 只 fanout 给"已对父 channel 开启 auto_follow_threads=1"的用户，
+	// 如果创建者没关注父 channel 就会漏看自己新建的子区（issue #557）。这里单独给
+	// 创建者补一行 thread ext。关键约束：FollowThreadForCreator 不会清父群 group_unfollowed，
+	// 绝不因建子区而 re-follow 用户显式取关过的父 channel（沿用 OnThreadCreated 同序）。
+	// 同样 best-effort：失败只警告，不回滚（下次 FollowChannel / refollow 会补齐）。
+	if convSvc := conversation_ext.GetGlobalConvExtService(); convSvc != nil && req.SpaceID != "" {
+		threadChannelID := BuildChannelID(req.GroupNo, shortID)
+		if err := convSvc.FollowThreadForCreator(req.CreatorUID, req.SpaceID, threadChannelID); err != nil {
+			s.Warn("FollowThreadForCreator 失败（thread 已创建，创建者可下次 refollow 补齐）",
+				zap.String("groupNo", req.GroupNo), zap.String("shortID", shortID),
+				zap.String("creatorUID", req.CreatorUID), zap.Error(err))
 		}
 	}
 


lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD 20ba58f3

## Lens Summary

| Lens | Verdict | Headline |
|---|---|---|
| code | `changes-requested` | Fix is inert in production — create-thread route has no SpaceMiddleware, so SpaceID is always "" and FollowThreadForCreator never runs. |
| security | `cleared` | All inputs server/session-derived; SQL parameterized; self-scoped write, no IDOR; no secrets. No findings. |
| qa | `pass-with-risk` | P1 no-re-follow regression guard is strong & real-DB, but call-site wiring is untested and CI check-sprint is red. |

## Critical Findings (P0-P1)

- `[P0-1]` `modules/thread/api.go:161` (route group) + `modules/thread/service.go:325` — **The fix never executes on the real create-thread HTTP path.** The route group `/v1/groups/:group_no/threads` is registered with only `AuthMiddleware` — **no `spacepkg.SpaceMiddleware`**. The `space_id` gin-context key is set *exclusively* inside `spaceMiddleware` (`pkg/space/middleware.go:129,152`), and `GetSpaceID(c)` reads only that key, so on every production create-thread request `req.SpaceID == ""`, the guard `convSvc != nil && req.SpaceID != ""` is false, and `FollowThreadForCreator` is never invoked. Issue #557 stays broken for the actual user flow the PR claims to "全端根治". — flagged by code/code-1, corroborated by qa/qa-2 + qa/qa-5 (wiring untested / empty-SpaceID silently skips). **Verified independently by the aggregator** (route mounts only AuthMiddleware; no non-test `Set("space_id")` outside SpaceMiddleware; `conversation_ext/1module.go:62` and `messages_search/api.go:129` prove SpaceMiddleware is opt-in per-group, not global). — **Fix**: mount `spacepkg.SpaceMiddleware(t.ctx)` on the thread route group (as `/v1/conversation`, `/v1/messages` do), AND add a handler/integration test asserting `GetSpaceID(c)` is non-empty on create (or that the creator's thread ext row exists after `createThread`) so this can never regress silently.

## Should-Fix (P2)

- `[P2-1]` `modules/thread/const.go:17` vs `modules/conversation_ext/service.go:25` — AC#1 silently depends on the two independent `"____"` separator constants (`thread.ChannelIDSeparator`, `conversation_ext.threadSeparator`) staying equal; if either diverges, the creator's ext row is written under a key the follow-tab query won't match — breaking the fix with no test failure. — flagged by qa/qa-3. — **Fix**: share one constant across packages, or add a cross-package equality assertion.
- `[P2-2]` `modules/thread/service.go:319-328` + `modules/conversation_ext/service_test.go:817` — the `CreateThread` → `FollowThreadForCreator` wiring and the `SpaceID` plumbing have **zero** test coverage; only the isolated `conversation_ext` method is tested (`GetGlobalConvExtService()` is nil in unit tests), so the inert-fix defect (P0-1) is invisible to the suite. For a bug-fix PR the regression test must fail if the fix stops firing. — flagged by code/code-3 + qa/qa-2. — **Fix**: add a thread-module integration test covering the real path (this same test would catch P0-1); if the Redis/WuKongIM env is truly unavailable, get explicit human sign-off on the gap and file a follow-up.

## Suggestions (P3)

- `[P3-1]` `modules/thread/service.go:319` — empty `SpaceID` silently skips the creator-follow with no warning; a future missing-middleware regression would again be invisible. — flagged by qa/qa-5. — Fix: `Warn` when SpaceID is unexpectedly empty on the create route.
- `[P3-2]` `modules/conversation_ext/service_test.go:820` — AC#1 asserts the DB row exists, not that the thread actually surfaces in the follow-tab *listing* query (the user-visible outcome). — flagged by qa/qa-4. — Fix: extend one test to call the follow-tab list path and assert the thread appears.
- `[P3-3]` `modules/bot_api/threads.go:94` — the bot_api `CreateThread` caller does not set `SpaceID`, so bot-created threads never write a creator follow row — an undocumented asymmetry with the human path (likely fine since bots have no follow tab). — flagged by code/code-2. — Fix: add a one-line comment marking the omission intentional.

## Questions for Author

- `[Q-1]` CI `check-sprint / check-sprint` is **FAILURE** while all Build/Test/Vet/Lint checks are green (`scan-scheduled` SKIPPED is normal). Is this an unrelated spec/governance gate, or does it block merge? — flagged by qa/qa-1, also noted by security lens.

## Verdict Rationale

`REQUEST_CHANGES` because there is a confirmed code Must-Fix at Broad reach (P0-1): the fix does not run on the production create-thread path, so the PR does not actually resolve #557 despite correct service-layer logic. The aggregator independently verified the missing `SpaceMiddleware` and the exclusive `space_id` setter. **Contrarian preservation**: the security lens `cleared` the trust boundary but assumed the Space middleware runs on this route; code + aggregator verification show it does not — code's finding is kept, not silently dropped, because the two lenses disagree on a factual premise (the security *trust-boundary* conclusion is still correct — inputs are safe — but the *feature* is inert). The `FollowThreadForCreator` implementation, lock ordering, idempotency, and the strong no-re-follow regression guard (AC#2) are otherwise solid; once the middleware is mounted and an end-to-end test added, this should converge quickly.

## Round Context

|---|---|---|
| `modules/conversation_ext/service.go` | ✓ | (logic correct; P3-1) |
| `modules/conversation_ext/service_test.go` | ✓ | P2-1, P2-2, P3-2 |
| `modules/thread/service.go` | ✓ | P0-1, P2-2, P3-1 |
| `modules/thread/api.go` | ✓ | P0-1 |

## Notes

- Credential scan: clean (sole high-entropy hit is the test identifier `ParentGroupNotPreviouslyUnfollowed`, not a secret).
- Confidence filter aggregate: code 0 dropped / 1 downgraded; qa 1 dropped / 0 downgraded; security 0/0.

---

lml2468 (changes requested):
REQUEST_CHANGES — Round 1

The fix is inert on the production create-thread path: the route mounts only AuthMiddleware (no SpaceMiddleware), so GetSpaceID returns "" and FollowThreadForCreator never runs — issue #557 stays broken. Service-layer logic and the no-re-follow regression guard are solid; mount SpaceMiddleware on the thread route group and add an end-to-end regression test.

Full aggregate on this PR (issue comment above) and on the multica tracker.

OctoBoooot (changes requested):
## Review: fix(thread): 创建子区后为创建者补 thread ext 行 (#558)

**Verdict:** Request changes — one blocker: the fix is a **guaranteed no-op in production**. The space_id it depends on is never populated on the thread-create route, so `FollowThreadForCreator` is never called; the #557 bug persists on all clients. The service method itself and its tests are fine — the defect is the wiring. (`check-sprint` red is governance, out-of-band.)

Context: this is the third PR for issue #557 (alongside #563 and #565). #565 fixed it correctly by deriving the space **server-side** from `(group, creator)` via `GetMemberExternalFields`. #558 instead reads `spacepkg.GetSpaceID(c)` — which does not work on this route.

### Blocker
- `thread/api.go:233` `SpaceID: spacepkg.GetSpaceID(c)` + `thread/service.go` guard `req.SpaceID != ""` — **`GetSpaceID(c)` returns `""` on the createThread route, so the guard is always false and `FollowThreadForCreator` never runs.** Chain:
  1. `spacepkg.GetSpaceID(c)` reads `c.Get("space_id")`, returning `""` if unset (`pkg/space/middleware.go:158`).
  2. The only thing that sets `c.Set("space_id", …)` is `SpaceMiddleware`, which is opt-in per route group.
  3. The createThread route mounts **only `AuthMiddleware`, not `SpaceMiddleware`** (`thread/api.go:161`: `r.Group("/v1/groups/:group_no/threads", t.ctx.AuthMiddleware(r))`).
  4. There is no fallback extraction in the handler (contrast #563, which at least manually read `?space_id`/`X-Space-ID`).
  → On every real request `req.SpaceID == ""` → `convSvc != nil && req.SpaceID != ""` is false → `FollowThreadForCreator` is never invoked. The creator's thread ext row is never written; the bug the PR claims to fix ("全端根治") is not fixed on any client.

  **Why CI is green anyway:** the tests call `FollowThreadForCreator` directly, bypassing the `createThread` wiring — so they prove the method works while the shipped path is inert. This is the "tests exercise the service method, not the wiring" gap turned fatal.

  **Fix (adopt #565's approach):** derive the space server-side, e.g. `GetMemberExternalFields(groupNo, creatorUID).homeSpaceID` (internal → group.space_id, external → source_space_id), which needs no middleware and works for external creators too. Alternatively mount SpaceMiddleware on the route — but server-side derivation is strictly better (no client dependency, correct for external members). Either way, add a handler-level test that drives `POST …/threads` end-to-end and asserts the creator's thread ext row exists — that's the test that would have caught this.

### Praise
- `FollowThreadForCreator` itself is correct: INSERT-IGNORE idempotent, does NOT clear the parent `group_unfollowed` (the #557/web-PR P1 red-line), correct version→ext lock order, and the `DoesNotClearParentUnfollowed` test pins that subtle boundary. The service layer is right — it's just not reachable.

### Note
- Three open PRs (#558, #563, #565) fix the same #557 bug. #565 is the one with a working server-derived space. The team should pick one and close the other two; shipping #558 as-is would merge a no-op. If #558 is the chosen vehicle, port #565's server-side derivation into it.

Jerry-Xin (changes requested):
The PR is in scope, but the production thread creation path still skips the new creator follow write.

**🔴 Blocking**
🔴 Critical — `SpaceID` is never populated for the normal thread creation route, so the fix does not run. `modules/thread/api.go:161` mounts `/v1/groups/:group_no/threads` with only `AuthMiddleware`, but `modules/thread/api.go:233` reads `spacepkg.GetSpaceID(c)`. That helper only returns a value previously set by `SpaceMiddleware` (`pkg/space/middleware.go:96-110`, `pkg/space/middleware.go:157-164`). Since this route does not mount that middleware, `req.SpaceID` is empty and the guard in `modules/thread/service.go:320` skips `FollowThreadForCreator`. As a result, the reported case remains broken for the primary API path: a creator who does not follow the parent channel still will not get a `user_conversation_ext` thread row.

The safest fix is to make `CreateThread` derive the authoritative space from the parent group record, rather than relying on a client-declared/request-context space. If you only add `SpaceMiddleware`, also ensure the requested space matches the parent group’s actual `space_id`; membership in an arbitrary claimed space is not proof that the thread belongs in that space’s follow data.

**💬 Non-blocking**
🟡 Warning — Other `CreateThread` call sites still omit `SpaceID`, for example `modules/bot_api/threads.go:94`. If creator follow materialization is meant to be a service invariant, deriving the space in `CreateThread` would cover all callers and avoid future partial wiring.

🟡 Warning — The tests cover `FollowThreadForCreator` directly, but not the integrated `createThread -> CreateThread -> user_conversation_ext` path. Add a regression test that creates a thread through the route or service with a real parent group space and asserts the creator row is written while the parent `group_unfollowed` flag remains unchanged.

**✅ Highlights**
The new helper’s core behavior is well scoped: it upserts only the thread row, preserves the parent group unfollow state, and follows the existing version-bump-before-ext-write lock ordering.

yujiawei (changes requested):
# Code Review — PR #558 (octo-server)

**Head SHA:** `c7a72062043e95aa13c3b6ddf1fc8824d7de3658` · **Base:** `76d31cca` · Reviewed the full diff against base.

## Summary

The design is sound: root-causing creator-invisibility to a missing `user_conversation_ext` thread row, and adding a dedicated `FollowThreadForCreator` that upserts *only* the thread row without clearing the parent group's `group_unfollowed` flag, is exactly right — it correctly avoids the "re-follow on thread create" P1 that the earlier web-side attempt was rejected for. The lock ordering, idempotency, and best-effort framing all match the surrounding `OnThreadCreated` path.

**However, as wired the fix does not take effect in production.** The new code path is guarded behind a context value that nothing populates on the thread-create route, so the guard is always false and `FollowThreadForCreator` is never called by the real HTTP flow. This is a blocker.

---

## 1. Spec Compliance

**Spec: ❌**

The stated goal (issue #557: the creator must see their own newly created thread in the follow tab, server-side, all clients) is **not achieved by the runtime path** — the fix is dead code in production.

- **Deviation (blocking):** `modules/thread/service.go:311` gates the new behavior on `req.SpaceID != ""`:
  ```go
  if convSvc := conversation_ext.GetGlobalConvExtService(); convSvc != nil && req.SpaceID != "" {
      threadChannelID := BuildChannelID(req.GroupNo, shortID)
      if err := convSvc.FollowThreadForCreator(req.CreatorUID, req.SpaceID, threadChannelID); err != nil { ... }
  }
  ```
  `req.SpaceID` is populated only from `spacepkg.GetSpaceID(c)` at `modules/thread/api.go:233`. `GetSpaceID` (`pkg/space/middleware.go:158`) merely reads `c.Get("space_id")`, and the **only** code in the repo that sets that context key is `SpaceMiddleware` (`pkg/space/middleware.go:129` and `:152`). The thread route group does **not** install `SpaceMiddleware`:
  ```go
  // modules/thread/api.go:161
  threads := r.Group("/v1/groups/:group_no/threads", t.ctx.AuthMiddleware(r))
  {
      threads.POST("", t.createThread)   // no SpaceMiddleware in this chain
      ...
  }
  ```
  `AuthMiddleware` (octo-lib `pkg/wkhttp/http.go`) sets only `uid`/`name`/`role`, never `space_id`; and there is no global middleware in `main.go` (`UseGin` chain at `main.go:154–205`) that sets it either. Therefore on `POST /v1/groups/:group_no/threads`, `GetSpaceID(c)` returns `""`, the guard is false, and `FollowThreadForCreator` is never invoked. The creator-visibility bug the PR claims to root-fix remains unfixed in production for every client.

  Note that even a client sending `?space_id=` or `X-Space-ID` would not help: the handler reads the middleware-populated context key, not the raw request, and no middleware validates/sets it on this route.

Nothing is over-built or omitted beyond this; the problem is purely that the one new call site is unreachable.

### Required direction (pick one)
1. Add `spacepkg.SpaceMiddleware(t.ctx)` to the `/v1/groups/:group_no/threads` route group (as `/v1/sidebar` and `/v1/message` already do), so `space_id` is validated and set — but note this makes `space_id` effectively required on thread create, so confirm all clients send it; **or**
2. Resolve `spaceID` server-side inside `CreateThread` from `req.GroupNo` (the group → space mapping is authoritative and does not depend on an optional header). This is the more robust option and removes the silent-skip failure mode entirely.

---

## 2. Code Quality

### P1 — Fix is a no-op on the production route (see Spec §1)
Same finding as above; restated here because it is both a spec miss and a correctness defect. Until the route is wired (or `spaceID` derived server-side), the diff changes nothing observable.

### P2 — Silent-skip + best-effort combine into an undetectable, unrecoverable miss for the exact target user
`modules/thread/service.go:311–319`. Two behaviors stack badly:
- The `req.SpaceID != ""` guard **silently skips** when space is absent — no `Warn`, no metric. If option (1) is chosen and one client forgets the header, creators silently keep hitting #557 with zero signal in logs.
- The call is best-effort (failure only `Warn`, "next refollow will backfill"). But the canonical #557 user has *explicitly unfollowed* the parent group and will not organically re-follow it, so a transient DB failure here means the creator **permanently** loses their own thread from the follow tab with no recovery path. Consider at least logging the skip branch, and/or a bounded retry consistent with `OnThreadCreated`'s `withDeadlockRetry`.

### P2 — No test exercises the production wiring
`modules/conversation_ext/service_test.go:817+` thoroughly covers `FollowThreadForCreator` in isolation (creates row / does-not-clear-parent-unfollowed / idempotent / input validation) — good tests, and the `DoesNotClearParentUnfollowed` regression guard is the right one. But none of them proves the HTTP path ever *calls* the method with a non-empty `SpaceID`. A route/integration test asserting that `POST /v1/groups/:group_no/threads` results in a `target_type=5` row for the creator would have caught the dead-code wiring. Please add one alongside the fix.

### Verified OK (no action)
- Lock order in `FollowThreadForCreator` (`service.go:990`+): bump `follow_version` before ext upsert — consistent with `FollowThread`/`UpdateSort`. ✅
- Idempotency: `upsertTx` with empty `ConvExtFields{}` is `INSERT IGNORE` on the UK, a true no-op on repeat. Row *presence* (not any field value) is what makes a thread visible in the sidebar — the fanout path `bulkInsertThreadExtForFanoutUsersTx` writes the same 4 key columns and no extra fields, so the empty-fields write is correct and consistent, not a malformed row. ✅
- Ordering: `FollowThreadForCreator` runs before both `sendSourceMessage` and `sendThreadCreatedMessage`, preserving the "ext row committed before any client-observable message" invariant that `OnThreadCreated` documents. ✅
- Deliberately not clearing `group_unfollowed` — correct, and the key improvement over the earlier web approach. ✅

---

## 3. Overall Verdict

**REQUEST_CHANGES**

Spec ❌ (fix does not run on the production route) AND Quality Changes-Requested. The approach is correct and most of the implementation is clean — this is a wiring gap, not a redesign. Wire `space_id` onto the thread-create route (or derive it server-side from `group_no`), add a route-level test proving the creator row is written, and this should be a quick turnaround.

## 4. Suggestions
- Prefer server-side `spaceID` resolution from `group_no` over route middleware — it removes the header-dependency failure mode and can't silently skip.
- If keeping the guard, log the skip branch so a missing-space misconfiguration is observable.

## 5. Additional Notes
- Reviewed with a cross-model second opinion; the "fix is dead on the production route" finding was reproduced independently.

Jerry-Xin:
Pushed a follow-up fix (ee462687) on top of the previous commit.

The prior change added `SpaceID: spacepkg.GetSpaceID(c)` in `createThread` and the `FollowThreadForCreator` call, but the thread route group was never mounted with `SpaceMiddleware` — so `GetSpaceID(c)` always returned "", `req.SpaceID` was always empty, and `FollowThreadForCreator` never fired. A creator who had not auto-followed the parent channel still could not see a thread they just created in their follow tab.

This commit mounts `SpaceMiddleware` on `/v1/groups/:group_no/threads` so `space_id` is resolved from context, and adds a route-wiring regression test (`api_space_wiring_test.go`) that drives the real registered route with a `space_id` the caller is not a member of: with the middleware mounted the request is rejected 403 by the membership check, locking the wiring so #557 cannot silently regress.

Verified end-to-end on a live 8090 service: creator's `user_conversation_ext` thread row is now written on thread creation, the thread appears in the creator's follow tab, the P1 guard holds (parent `group_unfollowed` stays 1, no silent re-follow), and the no-`space_id` path is fail-open with no regression.

@clawsweeper re-review

OctoBoooot (changes requested):
## Delta re-review: #558 @ ee462687 (prior @ c7a72062) — no-op fixed, but the fix over-scopes into a group-wide authz change

**Verdict:** Request changes. The dead-code no-op I flagged is genuinely fixed — but by mounting `SpaceMiddleware` on the **entire thread route group**, which (a) newly space-gates 9 other handlers that never had a space concept, and (b) risks 403'ing legitimate cross-space external members. #565 already fixes the same #557 bug with server-side derivation that touches only createThread and needs no middleware. Recommend that scoped approach instead.

### What's fixed (verified)
- `api.go:161` now mounts `SpaceMiddleware(t.ctx)` on the group, so `SpaceMiddleware` sets `c.Set("space_id",...)` → `GetSpaceID(c)` (line 233) returns a real membership-verified value → `req.SpaceID` non-empty → `FollowThreadForCreator` actually fires. The dead-code no-op is resolved.
- `api_space_wiring_test.go` is a genuine wiring regression net: it sends a `space_id` the user isn't a member of and asserts the middleware's 403 ("无权访问该 Space"), which only appears if SpaceMiddleware is actually mounted. Good — it closes the "green service tests, dead wiring" gap that hid the original no-op.

### Major (should-verify: external-member regression across the whole route group)
- Mounting SpaceMiddleware on the **group** (`api.go:161`) applies it to all 10 handlers: createThread, listThreads, getThread, updateThread, updateSetting, listMembers, join, leave, archive, unarchive. Before this change `GetSpaceID` was used **nowhere** on these routes (grep: 1 occurrence, the new line) — so this is pure new space-gating for 9 endpoints, added to feed one line in createThread.
- The authz models diverge exactly for external members:
  - `CreateThread` gates on `ExistMemberActive(groupNo, uid)` — `group_member` with **no is_external filter**, so cross-space external members (is_external=1) ARE allowed to create/act on threads today.
  - `SpaceMiddleware` gates on `CheckMembership(spaceID, uid)` — `INNER JOIN space_member ... sm.uid=? AND sm.space_id=?`, which an external group-member is **not** in for the group's space → returns false → **403**.
  - Failing path: an external member (group member, space non-member) who sends the group's `space_id` on any of these 10 routes now gets 403 where they previously succeeded.
- I can construct this in code but can't confirm from here whether external clients actually send the group's space_id on these requests (if they send their own source-space id or none, `SpaceMiddleware`'s fail-open-on-absent path spares them). So it's a **should-verify major, not an asserted blocker**: please confirm what space_id external-member clients send on thread routes. If any send the group's space_id, this is a live regression on 10 endpoints.

### Recommended fix (narrower, no regression surface)
- Adopt #565's approach: derive the creator's space **server-side** inside `CreateThread` from `(groupNo, creatorUID)` via `GetMemberExternalFields().homeSpaceID` (internal → group.space_id, external → source_space_id). It needs no middleware, touches only the create path, works for external creators, and doesn't change authz on the other 9 handlers. That's strictly less blast radius than group-mounting SpaceMiddleware for one line.
- If SpaceMiddleware really is wanted on this group as intentional hardening, that's a separate decision that should be made explicitly (and tested for the external-member case), not carried in as a side effect of the #557 follow-backfill.

### Test gap
- The wiring test covers createThread + non-member = 403, but there's no test for the case that matters here: an **external group-member (space non-member)** creating/accessing a thread. That's exactly the population the SpaceMiddleware mount could newly break; it should be pinned (either "still allowed" if derivation is used, or an explicit decision if the 403 is intended).

### Note
- Also relevant: #558/#563/#565 are three PRs for the same #557 bug. #565 (server-derived space) is the clean one. Strongly suggest converging on #565 and closing #558/#563 rather than landing a third variant that changes group-wide authz.
- CI: Build green, Test in-progress, pr-title-lint green, check-sprint governance-red.

OctoBoooot (dismissed):
## Delta re-review: #558 @ ee462687 (prior @ c7a72062) — no-op fixed, but the fix over-scopes into a group-wide authz change

**Verdict:** Request changes. The dead-code no-op I flagged is genuinely fixed — but by mounting `SpaceMiddleware` on the **entire thread route group**, which (a) newly space-gates 9 other handlers that never had a space concept, and (b) risks 403'ing legitimate cross-space external members. #565 already fixes the same #557 bug with server-side derivation that touches only createThread and needs no middleware. Recommend that scoped approach instead.

### What's fixed (verified)
- `api.go:161` now mounts `SpaceMiddleware(t.ctx)` on the group, so `SpaceMiddleware` sets `c.Set("space_id",...)` → `GetSpaceID(c)` (line 233) returns a real membership-verified value → `req.SpaceID` non-empty → `FollowThreadForCreator` actually fires. The dead-code no-op is resolved.
- `api_space_wiring_test.go` is a genuine wiring regression net: it sends a `space_id` the user isn't a member of and asserts the middleware's 403 ("无权访问该 Space"), which only appears if SpaceMiddleware is actually mounted. Good — it closes the "green service tests, dead wiring" gap that hid the original no-op.

### Major (should-verify: external-member regression across the whole route group)
- Mounting SpaceMiddleware on the **group** (`api.go:161`) applies it to all 10 handlers: createThread, listThreads, getThread, updateThread, updateSetting, listMembers, join, leave, archive, unarchive. Before this change `GetSpaceID` was used **nowhere** on these routes (grep: 1 occurrence, the new line) — so this is pure new space-gating for 9 endpoints, added to feed one line in createThread.
- The authz models diverge exactly for external members:
  - `CreateThread` gates on `ExistMemberActive(groupNo, uid)` — `group_member` with **no is_external filter**, so cross-space external members (is_external=1) ARE allowed to create/act on threads today.
  - `SpaceMiddleware` gates on `CheckMembership(spaceID, uid)` — `INNER JOIN space_member ... sm.uid=? AND sm.space_id=?`, which an external group-member is **not** in for the group's space → returns false → **403**.
  - Failing path: an external member (group member, space non-member) who sends the group's `space_id` on any of these 10 routes now gets 403 where they previously succeeded.
- I can construct this in code but can't confirm from here whether external clients actually send the group's space_id on these requests (if they send their own source-space id or none, `SpaceMiddleware`'s fail-open-on-absent path spares them). So it's a **should-verify major, not an asserted blocker**: please confirm what space_id external-member clients send on thread routes. If any send the group's space_id, this is a live regression on 10 endpoints.

### Recommended fix (narrower, no regression surface)
- Adopt #565's approach: derive the creator's space **server-side** inside `CreateThread` from `(groupNo, creatorUID)` via `GetMemberExternalFields().homeSpaceID` (internal → group.space_id, external → source_space_id). It needs no middleware, touches only the create path, works for external creators, and doesn't change authz on the other 9 handlers. That's strictly less blast radius than group-mounting SpaceMiddleware for one line.
- If SpaceMiddleware really is wanted on this group as intentional hardening, that's a separate decision that should be made explicitly (and tested for the external-member case), not carried in as a side effect of the #557 follow-backfill.

### Test gap
- The wiring test covers createThread + non-member = 403, but there's no test for the case that matters here: an **external group-member (space non-member)** creating/accessing a thread. That's exactly the population the SpaceMiddleware mount could newly break; it should be pinned (either "still allowed" if derivation is used, or an explicit decision if the 403 is intended).

### Note
- Also relevant: #558/#563/#565 are three PRs for the same #557 bug. #565 (server-derived space) is the clean one. Strongly suggest converging on #565 and closing #558/#563 rather than landing a third variant that changes group-wide authz.
- CI: Build green, Test in-progress, pr-title-lint green, check-sprint governance-red.

OctoBoooot (dismissed):
## Delta re-review: #558 @ ee462687 (prior @ c7a72062) — no-op fixed, but the fix over-scopes into a group-wide authz change

**Verdict:** Request changes. The dead-code no-op I flagged is genuinely fixed — but by mounting `SpaceMiddleware` on the **entire thread route group**, which (a) newly space-gates 9 other handlers that never had a space concept, and (b) risks 403'ing legitimate cross-space external members. #565 already fixes the same #557 bug with server-side derivation that touches only createThread and needs no middleware. Recommend that scoped approach instead.

### What's fixed (verified)
- `api.go:161` now mounts `SpaceMiddleware(t.ctx)` on the group, so `SpaceMiddleware` sets `c.Set("space_id",...)` → `GetSpaceID(c)` (line 233) returns a real membership-verified value → `req.SpaceID` non-empty → `FollowThreadForCreator` actually fires. The dead-code no-op is resolved.
- `api_space_wiring_test.go` is a genuine wiring regression net: it sends a `space_id` the user isn't a member of and asserts the middleware's 403 ("无权访问该 Space"), which only appears if SpaceMiddleware is actually mounted. Good — it closes the "green service tests, dead wiring" gap that hid the original no-op.

### Major (should-verify: external-member regression across the whole route group)
- Mounting SpaceMiddleware on the **group** (`api.go:161`) applies it to all 10 handlers: createThread, listThreads, getThread, updateThread, updateSetting, listMembers, join, leave, archive, unarchive. Before this change `GetSpaceID` was used **nowhere** on these routes (grep: 1 occurrence, the new line) — so this is pure new space-gating for 9 endpoints, added to feed one line in createThread.
- The authz models diverge exactly for external members:
  - `CreateThread` gates on `ExistMemberActive(groupNo, uid)` — `group_member` with **no is_external filter**, so cross-space external members (is_external=1) ARE allowed to create/act on threads today.
  - `SpaceMiddleware` gates on `CheckMembership(spaceID, uid)` — `INNER JOIN space_member ... sm.uid=? AND sm.space_id=?`, which an external group-member is **not** in for the group's space → returns false → **403**.
  - Failing path: an external member (group member, space non-member) who sends the group's `space_id` on any of these 10 routes now gets 403 where they previously succeeded.
- I can construct this in code but can't confirm from here whether external clients actually send the group's space_id on these requests (if they send their own source-space id or none, `SpaceMiddleware`'s fail-open-on-absent path spares them). So it's a **should-verify major, not an asserted blocker**: please confirm what space_id external-member clients send on thread routes. If any send the group's space_id, this is a live regression on 10 endpoints.

### Recommended fix (narrower, no regression surface)
- Adopt #565's approach: derive the creator's space **server-side** inside `CreateThread` from `(groupNo, creatorUID)` via `GetMemberExternalFields().homeSpaceID` (internal → group.space_id, external → source_space_id). It needs no middleware, touches only the create path, works for external creators, and doesn't change authz on the other 9 handlers. That's strictly less blast radius than group-mounting SpaceMiddleware for one line.
- If SpaceMiddleware really is wanted on this group as intentional hardening, that's a separate decision that should be made explicitly (and tested for the external-member case), not carried in as a side effect of the #557 follow-backfill.

### Test gap
- The wiring test covers createThread + non-member = 403, but there's no test for the case that matters here: an **external group-member (space non-member)** creating/accessing a thread. That's exactly the population the SpaceMiddleware mount could newly break; it should be pinned (either "still allowed" if derivation is used, or an explicit decision if the 403 is intended).

### Note
- Also relevant: #558/#563/#565 are three PRs for the same #557 bug. #565 (server-derived space) is the clean one. Strongly suggest converging on #565 and closing #558/#563 rather than landing a third variant that changes group-wide authz.
- CI: Build green, Test in-progress, pr-title-lint green, check-sprint governance-red.

OctoBoooot (dismissed):
## Delta re-review: #558 @ ee462687 (prior @ c7a72062) — no-op fixed, but the fix over-scopes into a group-wide authz change

**Verdict:** Request changes. The dead-code no-op I flagged is genuinely fixed — but by mounting `SpaceMiddleware` on the **entire thread route group**, which (a) newly space-gates 9 other handlers that never had a space concept, and (b) risks 403'ing legitimate cross-space external members. #565 already fixes the same #557 bug with server-side derivation that touches only createThread and needs no middleware. Recommend that scoped approach instead.

### What's fixed (verified)
- `api.go:161` now mounts `SpaceMiddleware(t.ctx)` on the group, so `SpaceMiddleware` sets `c.Set("space_id",...)` → `GetSpaceID(c)` (line 233) returns a real membership-verified value → `req.SpaceID` non-empty → `FollowThreadForCreator` actually fires. The dead-code no-op is resolved.
- `api_space_wiring_test.go` is a genuine wiring regression net: it sends a `space_id` the user isn't a member of and asserts the middleware's 403 ("无权访问该 Space"), which only appears if SpaceMiddleware is actually mounted. Good — it closes the "green service tests, dead wiring" gap that hid the original no-op.

### Major (should-verify: external-member regression across the whole route group)
- Mounting SpaceMiddleware on the **group** (`api.go:161`) applies it to all 10 handlers: createThread, listThreads, getThread, updateThread, updateSetting, listMembers, join, leave, archive, unarchive. Before this change `GetSpaceID` was used **nowhere** on these routes (grep: 1 occurrence, the new line) — so this is pure new space-gating for 9 endpoints, added to feed one line in createThread.
- The authz models diverge exactly for external members:
  - `CreateThread` gates on `ExistMemberActive(groupNo, uid)` — `group_member` with **no is_external filter**, so cross-space external members (is_external=1) ARE allowed to create/act on threads today.
  - `SpaceMiddleware` gates on `CheckMembership(spaceID, uid)` — `INNER JOIN space_member ... sm.uid=? AND sm.space_id=?`, which an external group-member is **not** in for the group's space → returns false → **403**.
  - Failing path: an external member (group member, space non-member) who sends the group's `space_id` on any of these 10 routes now gets 403 where they previously succeeded.
- I can construct this in code but can't confirm from here whether external clients actually send the group's space_id on these requests (if they send their own source-space id or none, `SpaceMiddleware`'s fail-open-on-absent path spares them). So it's a **should-verify major, not an asserted blocker**: please confirm what space_id external-member clients send on thread routes. If any send the group's space_id, this is a live regression on 10 endpoints.

### Recommended fix (narrower, no regression surface)
- Adopt #565's approach: derive the creator's space **server-side** inside `CreateThread` from `(groupNo, creatorUID)` via `GetMemberExternalFields().homeSpaceID` (internal → group.space_id, external → source_space_id). It needs no middleware, touches only the create path, works for external creators, and doesn't change authz on the other 9 handlers. That's strictly less blast radius than group-mounting SpaceMiddleware for one line.
- If SpaceMiddleware really is wanted on this group as intentional hardening, that's a separate decision that should be made explicitly (and tested for the external-member case), not carried in as a side effect of the #557 follow-backfill.

### Test gap
- The wiring test covers createThread + non-member = 403, but there's no test for the case that matters here: an **external group-member (space non-member)** creating/accessing a thread. That's exactly the population the SpaceMiddleware mount could newly break; it should be pinned (either "still allowed" if derivation is used, or an explicit decision if the 403 is intended).

### Note
- Also relevant: #558/#563/#565 are three PRs for the same #557 bug. #565 (server-derived space) is the clean one. Strongly suggest converging on #565 and closing #558/#563 rather than landing a third variant that changes group-wide authz.
- CI: Build green, Test in-progress, pr-title-lint green, check-sprint governance-red.

OctoBoooot (dismissed):
## Delta re-review: #558 @ ee462687 (prior @ c7a72062) — no-op fixed, but the fix over-scopes into a group-wide authz change

**Verdict:** Request changes. The dead-code no-op I flagged is genuinely fixed — but by mounting `SpaceMiddleware` on the **entire thread route group**, which (a) newly space-gates 9 other handlers that never had a space concept, and (b) risks 403'ing legitimate cross-space external members. #565 already fixes the same #557 bug with server-side derivation that touches only createThread and needs no middleware. Recommend that scoped approach instead.

### What's fixed (verified)
- `api.go:161` now mounts `SpaceMiddleware(t.ctx)` on the group, so `SpaceMiddleware` sets `c.Set("space_id",...)` → `GetSpaceID(c)` (line 233) returns a real membership-verified value → `req.SpaceID` non-empty → `FollowThreadForCreator` actually fires. The dead-code no-op is resolved.
- `api_space_wiring_test.go` is a genuine wiring regression net: it sends a `space_id` the user isn't a member of and asserts the middleware's 403 ("无权访问该 Space"), which only appears if SpaceMiddleware is actually mounted. Good — it closes the "green service tests, dead wiring" gap that hid the original no-op.

### Major (should-verify: external-member regression across the whole route group)
- Mounting SpaceMiddleware on the **group** (`api.go:161`) applies it to all 10 handlers: createThread, listThreads, getThread, updateThread, updateSetting, listMembers, join, leave, archive, unarchive. Before this change `GetSpaceID` was used **nowhere** on these routes (grep: 1 occurrence, the new line) — so this is pure new space-gating for 9 endpoints, added to feed one line in createThread.
- The authz models diverge exactly for external members:
  - `CreateThread` gates on `ExistMemberActive(groupNo, uid)` — `group_member` with **no is_external filter**, so cross-space external members (is_external=1) ARE allowed to create/act on threads today.
  - `SpaceMiddleware` gates on `CheckMembership(spaceID, uid)` — `INNER JOIN space_member ... sm.uid=? AND sm.space_id=?`, which an external group-member is **not** in for the group's space → returns false → **403**.
  - Failing path: an external member (group member, space non-member) who sends the group's `space_id` on any of these 10 routes now gets 403 where they previously succeeded.
- I can construct this in code but can't confirm from here whether external clients actually send the group's space_id on these requests (if they send their own source-space id or none, `SpaceMiddleware`'s fail-open-on-absent path spares them). So it's a **should-verify major, not an asserted blocker**: please confirm what space_id external-member clients send on thread routes. If any send the group's space_id, this is a live regression on 10 endpoints.

### Recommended fix (narrower, no regression surface)
- Adopt #565's approach: derive the creator's space **server-side** inside `CreateThread` from `(groupNo, creatorUID)` via `GetMemberExternalFields().homeSpaceID` (internal → group.space_id, external → source_space_id). It needs no middleware, touches only the create path, works for external creators, and doesn't change authz on the other 9 handlers. That's strictly less blast radius than group-mounting SpaceMiddleware for one line.
- If SpaceMiddleware really is wanted on this group as intentional hardening, that's a separate decision that should be made explicitly (and tested for the external-member case), not carried in as a side effect of the #557 follow-backfill.

### Test gap
- The wiring test covers createThread + non-member = 403, but there's no test for the case that matters here: an **external group-member (space non-member)** creating/accessing a thread. That's exactly the population the SpaceMiddleware mount could newly break; it should be pinned (either "still allowed" if derivation is used, or an explicit decision if the 403 is intended).

### Note
- Also relevant: #558/#563/#565 are three PRs for the same #557 bug. #565 (server-derived space) is the clean one. Strongly suggest converging on #565 and closing #558/#563 rather than landing a third variant that changes group-wide authz.
- CI: Build green, Test in-progress, pr-title-lint green, check-sprint governance-red.

Jerry-Xin (changes requested):
## REQUEST_CHANGES — the production no-op is only *partially* addressed; the common path is still dead, and the claimed-space is not validated against the parent group.

Re-reviewed at `ee462687ced2`. Delta since the last reviewed head (`c7a72062043e`) is just two things: mount `SpaceMiddleware` on the thread route (`modules/thread/api.go:161`) + a route-wiring test (`api_space_wiring_test.go`). The core `FollowThreadForCreator` guard and service path are unchanged. The unanimous prior blocker (fix is a guaranteed no-op in production) is **not resolved**.

### 🔴 Blocking

**1. Normal client POST *without* `space_id` is STILL a no-op.**
`SpaceMiddleware` is **opt-in** and skips silently when no space is provided (`pkg/space/middleware.go:107-113`):
```go
spaceID := c.Query("space_id"); if spaceID == "" { spaceID = c.GetHeader("X-Space-ID") }
if spaceID == "" { c.Next(); return }   // <-- no space_id → nothing set
```
So on a plain `POST /v1/groups/:group_no/threads` (no `space_id` query / no `X-Space-ID` header), `spacepkg.GetSpaceID(c)` returns `""` (`middleware.go:158-166`), `req.SpaceID==""`, and the guard at `modules/thread/service.go:320` (`convSvc != nil && req.SpaceID != ""`) is **false** → `FollowThreadForCreator` is never invoked. This is the exact #557 breakage, unchanged. The fix now depends entirely on every client volunteering the correct `space_id` on the create call — that is not something the server guarantees, and it is precisely the client-dependency this fix was supposed to eliminate.

**2. Claimed space is membership-checked but NOT matched to the parent group (repeat of my prior note).**
When a `space_id` *is* supplied, `SpaceMiddleware` only verifies the caller is a member of the *claimed* space (`CheckMembership`, `middleware.go:135`). It never verifies that claimed space == the parent group's actual `space_id`. A user who is a member of space B can create a thread under a group belonging to space A and get their ext row written under the wrong space. Membership in an arbitrarily-claimed space is not proof the thread belongs there.

**3. Fix is route-scoped; other CreateThread callers stay dead.**
`modules/bot_api/threads.go:94` builds `CreateThreadReq` with **no** `SpaceID` field at all → `req.SpaceID==""` → guard false → bot-created threads never fan out the creator row. Because the fix lives in the HTTP handler (deriving from middleware) rather than in `CreateThread` itself, every non-`/v1/groups/.../threads` caller remains a no-op.

**4. No integration test proves the real path reaches `FollowThreadForCreator`.**
This is the gap that made the blocker fatal, and it is still open:
- `api_space_wiring_test.go` only asserts a *non-member + supplied space_id* gets `403`. It does **not** drive a successful POST → `CreateThread` → creator `user_conversation_ext` row written.
- All `TestService_FollowThreadForCreator_*` tests (`conversation_ext/service_test.go`) call `svc.FollowThreadForCreator(...)` **directly**, bypassing the route/handler/guard — the same bypass that let the dead wiring ship green.
- There is still zero end-to-end proof that a normal POST results in the creator ext row (and that the parent `group_unfollowed` flag is untouched).

### Recommended fix (unchanged from prior round / sibling #565)
Derive space **server-side** from `(group, creator)` inside `CreateThread` (or the handler), e.g. `GetMemberExternalFields(groupNo, creatorUID)` → internal `= group.space_id`, external `= source_space_id`. This needs no client cooperation, no middleware, covers external creators, is authoritative (space is derived from the parent group, so #2 can't happen), and covers **all** callers including bot_api. Then add an integration test that POSTs and asserts the creator ext row exists while the parent `group_unfollowed` flag is unchanged.

### ✅ Highlights
- `FollowThreadForCreator` itself is correct: it upserts only the thread ext row and deliberately does not clear the parent `group_unfollowed` (`conversation_ext/service.go`), honoring the round-2/round-4 ordering + never-re-follow constraints.
- Fanout ordering (OnThreadCreated / creator row before any client-observable channel message) is well-reasoned and documented.

Given #558 is one of three PRs (#558/#563/#565) for issue #557 and the slate direction was "close in favor of #565" (which already does authoritative server-side derivation), this is a reasonable **close-in-favor-of-#565** candidate unless #558 adopts server-side derivation.

yujiawei (changes requested):
# Code Review — PR #558 (octo-server)

**Head SHA:** `ee462687ced29e47d1517cd5be359168bd6607c0` · **Base:** `main` · Reviewed the full diff against base, plus the read-side sidebar path and the space middleware.

## Summary

Progress since the previous head (`c7a72062`): the create-thread route now mounts `SpaceMiddleware` (`modules/thread/api.go:161`) plus a wiring test, so the fix is **no longer a guaranteed no-op** on the common path. The core design remains sound — `FollowThreadForCreator` upserts only the thread ext row, deliberately does not clear the parent group's `group_unfollowed`, and preserves the version→ext lock order; the `DoesNotClearParentUnfollowed` regression guard is exactly right.

However, the space value the fix depends on is now **client-supplied and only membership-validated**, not derived server-side. This leaves two failure modes that keep the fix from reliably achieving the PR's stated "root fix on all clients," including on the platform (iOS) where #557 was reported. Recommending the server-side derivation used by sibling PR #565.

---

## 1. Spec Compliance

**Spec: ❌**

Issue #557 asks for the creator to see their own newly created thread in the follow tab, **server-side, on all clients**. The mechanism is now wired onto the route, but its effect is conditional on client behavior and is not verifiable as "all clients":

- **Deviation (blocking) — the fix silently no-ops when `space_id` is absent.** `SpaceMiddleware` is opt-in: when neither `?space_id` nor `X-Space-ID` is present it calls `c.Next()` **without** setting the context key (`pkg/space/middleware.go:109-113`). `GetSpaceID(c)` then returns `""` (`middleware.go:158-164`), so `req.SpaceID` is empty and the guard at `modules/thread/service.go:320` (`convSvc != nil && req.SpaceID != ""`) is false — `FollowThreadForCreator` never runs, and there is **no log** on the skip branch. #557 was originally reported on iOS; nothing in this PR demonstrates that iOS/Android/PC actually send `space_id` on `POST /v1/groups/:group_no/threads`. If any client omits it, that client silently keeps hitting #557 with zero observability. The "全端根治 / 全端统一生效" claim in the PR description is therefore unsubstantiated for the exact platform the bug was filed against.

- **Deviation (blocking) — the claimed space is not validated against the parent group's space.** `SpaceMiddleware` only proves the caller is a *member of the claimed space* (`CheckMembership`, `pkg/space/membership.go:9`). It does **not** check that the claimed `space_id` equals the parent group's `space_id`. The write path stores the ext row under the client-claimed space (`FollowThreadForCreator` → `upsertTx(..., targetTypeThread, threadChannelID)`, `service.go:990`), while the sidebar read filters thread ext rows by the parent group's *actual* space (`filterThreadExtsBySpace`, `modules/message/api_sidebar.go:839`, and the `ListThreadExts(uid, spaceID)` query keyed on the tab's space, `db.go:238`). So a multi-space creator who sends a valid-but-different `space_id` gets the row written into a space where it will be filtered out on read (fix defeated) **and** an orphan row left behind. This is the exact caveat raised in the prior round ("membership in an arbitrary claimed space is not proof that the thread belongs in that space's follow data") and it is not addressed.

Nothing is over-built; the service method and its no-re-follow guard are correct. The gap is that the fix's *reachability and correctness* now hinge on unvalidated client-supplied space.

### Required direction (pick one; option 1 is strictly safer)
1. **Derive the space server-side inside `CreateThread` from `req.GroupNo` (creator)**, as sibling PR #565 does (internal → `group.space_id`; external → `source_space_id`). This removes the client dependency entirely — no silent skip, no wrong-space write, correct for external members — and covers all call sites including `modules/bot_api/threads.go:94`.
2. If keeping `SpaceMiddleware`: (a) provide evidence that **every** client sends `space_id` on thread create, (b) validate the claimed space matches the parent group's space before writing, and (c) log the skip branch so a missing/incorrect space is observable.

---

## 2. Code Quality

### P1 — Fix efficacy is client-dependent and un-observable (see Spec §1)
Restated here as a correctness defect: on any client that does not send a correct `space_id`, the diff changes nothing observable and emits no signal. Given #557's iOS origin, this is a real risk of shipping a fix that does not fix the reported case.

### P2 — Wrong-space write for multi-space / external creators (see Spec §1)
Client-claimed space that differs from the parent group's space produces a row that reads back filtered-out plus an orphan. Server-side derivation eliminates this.

### P2 — Redundant version-bump + upsert when the creator already auto-follows the parent
When the creator follows the parent with `auto_follow_threads=1`, `OnThreadCreated` (called synchronously just above, `service.go:305`) already writes the creator's thread ext row; `FollowThreadForCreator` then writes it again. Harmless because `upsertTx` is idempotent, but it is a redundant second version bump on the synchronous HTTP path. (Note: `OnThreadCreated` and `FollowThreadForCreator` run **sequentially in the same goroutine**, so there is no self-contention between them — only the usual contention with other concurrent follow operations.)

### P2 — `FollowThreadForCreator` has no deadlock retry, unlike the fanout path
`OnThreadCreated` wraps its writes in `withDeadlockRetry`; `FollowThreadForCreator` does not. It runs best-effort on the synchronous create path, and the canonical #557 user has explicitly unfollowed the parent and will not organically re-follow — so a transient 1213/1205 here means the creator permanently loses their own thread from the follow tab with no recovery. Consider matching `OnThreadCreated`'s bounded retry.

### P2 — New route test locks the mount, not the end-to-end write
`TestCreateThread_RouteMountsSpaceMiddleware` asserts a non-member `space_id` is rejected with 403 — good, it pins the mount so it can't silently regress. But it does not prove the happy path writes the creator's `target_type=5` row via the route; the service tests call `FollowThreadForCreator` directly. A route-level test that drives `POST …/threads` with a real parent-group space and asserts the creator row exists (and parent `group_unfollowed` is untouched) would close the wiring gap the prior round asked for.

### Verified correct (no action)
- `FollowThreadForCreator` lock order (bump `follow_version` before ext upsert) matches `FollowThread`/`UpdateSort`. ✅
- Idempotency: `upsertTx` with empty `ConvExtFields{}` is `INSERT IGNORE` on the UK; row *presence* is what makes a thread visible, and the fanout path writes the same key columns — the empty-fields write is consistent, not malformed. ✅
- Deliberately not clearing `group_unfollowed` — correct, and the key improvement over the earlier web attempt; pinned by `DoesNotClearParentUnfollowed`. ✅
- Channel-ID symmetry: `BuildChannelID` uses `ChannelIDSeparator = "____"`, matching `threadSeparator` used by `parseThreadChannelID`/read path. ✅
- Ordering: creator write runs before `sendSourceMessage`/`sendThreadCreatedMessage`, preserving the "ext row committed before any client-observable message" invariant. ✅

---

## 3. Overall Verdict

**CHANGES_REQUESTED**

The approach is right and the previous inert-fix blocker is resolved on the common path — this is close. But the fix now depends on a client-supplied `space_id` that is validated only for membership, giving two ways for #557 to persist silently (absent space → no-op with no log; wrong-but-valid space → wrong-space write). Deriving the space server-side from `group_no` (as in #565) makes the fix unconditional and correct for external members, and is the smaller long-term maintenance surface. Either adopt that, or prove all-client coverage + add the group-space match check + log the skip.

## 4. Suggestions
- Prefer server-side space resolution from `group_no` over route middleware — removes the header dependency and the wrong-space failure mode, and covers non-HTTP callers (`bot_api`).
- Add a route-level happy-path test asserting the creator's `target_type=5` row is written through `POST …/threads`.
- Log the skip branch and/or emit a metric so a missing/incorrect space is detectable in production.

## 5. Additional Notes
- A cross-model second opinion independently flagged the empty-`SpaceID` silent-skip (no warning) and questioned internal callers that invoke `CreateThread` without `SpaceID` (e.g. `bot_api`). It also raised concurrent double-writes between the creator write and the fanout; that specific concern does not apply here since the two run sequentially in the same goroutine, though the missing deadlock-retry on `FollowThreadForCreator` (P2 above) is the real related point.
- Three open PRs (#558/#563/#565) target #557; #565 carries the server-derived space. The team should converge on one vehicle and port the server-side derivation in.

Jerry-Xin (commented):
Thanks a lot for the quick turnaround on this — the previous dead-code no-op is genuinely resolved. Mounting `SpaceMiddleware` on the thread route now lets `GetSpaceID(c)` return a real membership-verified value, so `FollowThreadForCreator` actually fires, and the `api_space_wiring_test.go` regression net (403 on non-member space_id) is a nice touch that closes the "green service tests, dead wiring" gap. The core `FollowThreadForCreator` is also correct: it upserts only the thread ext row, deliberately leaves the parent `group_unfollowed` untouched, and keeps the version→ext lock order. Really appreciate the care here.

The remaining reason to hold is direction, not correctness of your service method:

1. **Common path is still conditional.** `SpaceMiddleware` is opt-in — when neither `?space_id` nor `X-Space-ID` is present it calls `c.Next()` without setting the key, so `req.SpaceID` is empty and the guard at `service.go:320` skips silently (no log). Since #557 was filed on iOS, we can't yet prove every client sends `space_id` on `POST /v1/groups/:group_no/threads`.
2. **Claimed space isn't matched to the parent group.** The middleware only verifies membership of the claimed space, not that it equals the parent group's `space_id`, so a multi-space creator sending a valid-but-different space_id writes an ext row that reads back filtered-out.
3. **Route-scoped, other callers stay dead.** `bot_api/threads.go:94` builds the request with no `SpaceID`, so bot-created threads never fan out the creator row.

Recommended direction (matches sibling #565): derive the space server-side inside `CreateThread` from `(group_no, creator)` — internal → `group.space_id`, external → `source_space_id`. That removes the client dependency, is authoritative (space comes from the parent group), covers external creators, and covers all callers including bot_api.

Since #558/#563/#565 all target #557 and #565 already carries the server-derived approach, suggest converging on #565 and closing this one — unless you'd prefer to port the server-side derivation into #558. Either way, thanks again for the solid work on the service layer; that part is landing regardless of which vehicle we pick.

Jerry-Xin:
Closing in favor of #565, which carries the server-side space derivation (deriving from the parent group + creator, covering external members and all callers including bot_api). Thanks again for the solid work on the FollowThreadForCreator service layer — that logic is landing via #565. Appreciate the effort here.

PR #553, opened 2026-07-08
feat(messages_search): global search thread coverage

BrDing-Rookie:
## What

Restore the missing thread (channelType=5) coverage for global search, and land it on upstream `main`.

## Why

PR #550 (merged as `728e7ab`) landed the P0/P1 global endpoints plus the RC blacklist gate, but **without thread coverage** — the original commit that implemented thread v1 (`ba06ff65`) was made in a standalone workdir, never pushed, and got bypassed by the RC line, so it went missing. Owner requirement: thread messages must be reachable from global search. This PR restores that commit and makes it coexist with the blacklist gate that landed in #550.

Related: `YUJ-10` (thread coverage decision), `YUJ-11` (RC blacklist fix, already merged in #550), `YUJ-12` (this recovery task).

## What's in

Cherry-pick `ba06ff65` on top of `upstream/main` and resolve the merge conflict with the #550 blacklist gate:

- `modules/thread/db.go`: add `QueryActiveShortIDsByGroupNos` — a single IN-batch query that returns `(group_no, short_id)` for `status=active` rows, served entirely from the covering index. Coexists with the existing per-shortID `QueryActiveShortIDs` already on merged main.
- `modules/messages_search/search_global.go`:
  - `buildAllowlist` signature changes from `([]channelRef, []channelRef, error)` to `([]channelRef, []channelRef, []channelRef, error)`, adding `allowThread`.
  - **Keep** the blacklist gate introduced by #550 (`ExistMembersActive`, group-side fail-closed), and layer the ba06ff65 thread enumeration on top: once the trusted `groupNos` list is computed, call `enumerateThreadsForGroups` and fold the resulting `channelType=5` composite channel IDs into `allowSet`.
  - `enumerateThreadsForGroups`: batch-query active threads and stitch composite IDs via `thread.BuildChannelID(group, short)`. Two hard safety caps: `maxThreadsPerGroup=200` and `maxTotalThreadChannelIDs=2000` (over-cap triggers a WARN-level degrade). DB errors are soft-fail — group + DM allowlist stays intact.
  - `resolveGlobalScope`: accept `allowThread`, flatten it into `allowSet`; `filters.channel_ids` entries with `channel_type=5` composite IDs can now match; out-of-scope thread IDs are dropped silently.
  - `channelsForMember`: v1 explicitly retains group + DM entries only, discarding thread entries (avoids the extra join).
- `modules/messages_search/api.go`: add `threadEnumFn` / `externalGroupFn` test seams.
- Tests:
[...]

PR #550, opened 2026-07-08
feat(messages_search): add global (cross-channel) search endpoints

BrDing-Rookie:
Fixes #551

## Summary

Adds three new endpoints for cross-channel search, implementing design doc **v10** (`01-backend-search-global.md`):

- `POST /v1/messages/_search_global_messages` — global counterpart of `_search_all`, unified message+file stream (`SearchAllHit`).
- `POST /v1/messages/_search_global_files` — global counterpart of `_search_files`, file-only (`FileHit`).
- `GET  /v1/messages/_search_file_types` — file-type enum for the frontend filter chip. Mounted on a **separate Auth-only group** (not under `/v1/messages`) to avoid `backendGate` + `SpaceMiddleware` side-effects.

Single-channel `_search*` endpoints are **unchanged**; global endpoints re-dispatch to them via a single-channel fast path when the resolved scope is exactly one channel.

## P0 shape changes (backward-compatible, all `omitempty`)

- `MessageHit.channel_type`
- `FileHit.channel_id` + `FileHit.channel_type`
- `singleSearchAllHit` / `singleMessageHit` / `singleFileHit` take channel context; single-channel paths pass `req` channel, global paths pass doc-derived values.
- DM hits reverse-resolve `channel_id` to the peer uid via `peerFromFakeChannelID` (mirrors indexer `reversePeer`).

## By phase

- **P1** – `search_global.go`: `resolveGlobalScope` (allowlist ∩ `channel_ids` ∩ `member_uid`, single-channel fast path, `self=ignore`, empty-intersection short-circuit), `validateGlobalBase` / `validateGlobalFileBase`, `applyGlobalDMSpaceScope` (`should(mustNot(channelType=1), (channelType=1 AND spaceId=X)) MSM=1`; **fail-closed** when `RequireSpaceID`).
- **P2** – `search_global_messages.go`: keyword mode hard-whitelists payload types `[1,8,11,14]`; browse mode adds `[2,5]`; `content_types` intersected; `channel_types` → `terms(channelType)`; `terms(channelId, resolved-set)` + `mustNot(revoked)` replace the single-channel lock; system-msg hard filter + exclude-virtual + common filters reused; **no Routing** on the OS query.
- **P3** – `search_global_files.go`: `Filter(payload.type=8)` hard-locked; `file_exts` lower-cased + de-duped; `file_size_min/max` → `range(payload.file.size)`.
[...]

PR #562, opened 2026-07-10
perf(messages_search): batch DM blacklist gate + per-phase audit timing

BrDing-Rookie:
# perf(messages_search): batch DM blacklist gate + per-phase audit timing (YUJ-27)

## Background

Global search endpoints (`_search_global_messages` / `_search_global_files`) served end-to-end in ~4.5s while the underlying OpenSearch query itself served in <20ms and the cluster was green. Bottleneck was in the octo-server allowlist, not in OpenSearch.

`buildAllowlist`'s DM blacklist gate previously issued **2N MySQL round-trips** — one `ExistBlacklist(loginUID, peer)` + one `ExistBlacklist(peer, loginUID)` per DM peer. For a user with a few hundred friends / same-Space members that's several hundred serial SQL round-trips per request, which is exactly where the seconds hid.

Group active check (`ExistMembersActive`) was already batched by prior work, so DM blacklist was the last remaining per-item MySQL fan-out on the hot path.

### 1. Batched bidirectional blacklist gate

New `user.IService.ExistBlacklistsBoth(loginUID, peers) -> (blockedByMe, blockedByPeer, error)`:

```sql
SELECT uid, to_uid FROM user_setting
WHERE blacklist=1
  AND ((uid=? AND to_uid IN (peers)) OR (uid IN (peers) AND to_uid=?))
```

Rows are classified by direction. Returned maps only contain peers actually blacklisted in that direction (missing key = not blacklisted).

Semantics are identical to N pair-wise `ExistBlacklist` calls: bidirectional blacklist, fail-closed on error. The batched version does become **stricter** on error — a single IN query failure drops every candidate rather than just the offender — which is safe on the security axis (hidden legit DMs beats leaked blacklisted ones) and mirrors the intent of the previous per-peer fail-closed branch.

`filterBlacklistedDMPeers` migrates to the batch call. 2N round-trips → 1.

### 2. Per-phase audit timings

Threaded through the audit middleware so a regression is visible in the log line next to `took_ms`:

- `allowlist_ms` = `member_active_ms` + `blacklist_ms` + `thread_enum_ms` + `dm_peers_ms`
- `os_search_ms`, `filter_visible_ms`, `sender_join_ms`, `oversample_pages`

Global endpoints record all of them; single-channel handlers pay zero timing cost (they pass nil into the shared paginate variant, which is a no-op branch). `paginateWithFilterDepth` keeps its existing signature and now delegates to `paginateWithFilterDepthInstr` — callers can adopt timings incrementally without touching a wide call surface.

## Non-goals / constraints preserved

- **No RC rollback.** #550 (`ExistMembersActive` group active gate) and #553 (thread coverage / archived-thread visibility / per-group thread LIMIT) semantics are unchanged; the blacklist gate is bidirectional, fail-closed, and blocks the exact same peers a per-pair check would.
- **No indexer or OpenSearch mapping changes.**
- **Single-channel `_search*` handlers are untouched** — no regression risk on the hot path.

## Tests

New: `modules/messages_search/allowlist_blacklist_batch_test.go`

- Batched path drops forward + reverse blacklist edges; keeps no-edge peers; guards self / empty entries against ever surviving.
- Bidirectional both-sides only drops once (idempotent).
- Batch error fails-closed for the whole set.
- Empty input short-circuits the service (never touches the DB).
- Legacy `TestEnumerateDMPeers_ExcludesBlacklistedPeers` fixture ("me→bob", "carol→me", "dave" clean) drives the batched path direct — bob + carol drop, dave survives.
- Stub semantic parity: for every peer, `ExistBlacklistsBoth`'s per-direction answer matches the corresponding `ExistBlacklist(a, b)` pair-wise call on the same fixture.

Existing tests kept passing unchanged:

- `TestBuildAllowlist_ExcludesGroupBlacklistedMember`
- `TestEnumerateDMPeers_ExcludesBlacklistedPeers`
- `TestEnumerateDMPeers_BlacklistErrorFailsClosedPerPeer` (still fail-closed; batch semantics = drop them all rather than only the offender)
- `TestEnumerateDMPeers_UnionsFriendsAndSpaceMembers`
- Full `TestBuildAllowlist_*` thread-coverage suite (thread channelID enumeration + soft-fail + per-group cap + global aggregate cap + archived-included + deleted-excluded).

```
go build ./...
go vet ./modules/messages_search/... ./modules/user/...
go test ./modules/messages_search/... -count=1  -> PASS
```

Refs: internal YUJ-27.

### DIFF
--- a/modules/messages_search/allowlist_blacklist_batch_test.go
+++ b/modules/messages_search/allowlist_blacklist_batch_test.go
@@ -1,0 +1,206 @@
+package messages_search
+
+// YUJ-27 regression tests: the DM allowlist blacklist gate must produce the
+// same set of surviving peers as the pre-batch, per-peer ExistBlacklist
+// implementation, but pay one MySQL round-trip instead of 2N. These tests
+// exercise ExistBlacklistsBoth-driven filterBlacklistedDMPeers semantics on
+// top of the shared stubUserSvc fixture — no real MySQL, no per-peer stub
+// swap-out.
+
+import (
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-server/modules/user"
+)
+
+// TestFilterBlacklistedDMPeers_BatchParity — the batched path must drop the
+// same peers a semantics-equivalent per-peer loop would drop: forward hit
+// (me→peer), reverse hit (peer→me), no-hit (survives), and self / empty
+// entries (never touch the DB, silently dropped).
+func TestFilterBlacklistedDMPeers_BatchParity(t *testing.T) {
+	loginUID := "me"
+	uSvc := &stubUserSvc{
+		blacklist: map[string]bool{
+			// me -> bob : forward
+			"me->bob": true,
+			// carol -> me : reverse
+			"carol->me": true,
+			// dave: no edge in either direction
+			// eve: no edge in either direction
+		},
+	}
+	gSvc := &stubGroupSvc{}
+	h := newHandlerForAllowlistTests(uSvc, gSvc)
+
+	got := h.filterBlacklistedDMPeers(loginUID, []string{
+		"bob", "carol", "dave", "eve",
+		"", loginUID, // self / empty must not survive and must not hit DB
+	})
+	gotSet := map[string]bool{}
+	for _, p := range got {
+		gotSet[p] = true
+	}
+	if gotSet["bob"] {
+		t.Fatalf("bob (forward blacklist) must be dropped by batch path; got %v", got)
+	}
+	if gotSet["carol"] {
+		t.Fatalf("carol (reverse blacklist) must be dropped by batch path; got %v", got)
+	}
+	if !gotSet["dave"] || !gotSet["eve"] {
+		t.Fatalf("dave and eve (no blacklist edges) must survive; got %v", got)
+	}
+	if gotSet[""] || gotSet[loginUID] {
+		t.Fatalf("empty peer / self must never survive; got %v", got)
+	}
+	if len(got) != 2 {
+		t.Fatalf("expected exactly {dave,eve} to survive; got %v", got)
+	}
+}
+
+// TestFilterBlacklistedDMPeers_BidirectionalBothSides — when both directions
+// are blacklisted the peer must still drop exactly once (idempotent) and no
+// extra call artefacts are observable through the returned set.
+func TestFilterBlacklistedDMPeers_BidirectionalBothSides(t *testing.T) {
+	loginUID := "me"
+	uSvc := &stubUserSvc{
+		blacklist: map[string]bool{
+			"me->frank":  true,
+			"frank->me":  true,
+			"me->grace":  false, // absent = false
+			"grace->me":  true,
+			"dave->me":   false,
+		},
+	}
+	gSvc := &stubGroupSvc{}
+	h := newHandlerForAllowlistTests(uSvc, gSvc)
+
+	got := h.filterBlacklistedDMPeers(loginUID, []string{"frank", "grace", "dave"})
+	if len(got) != 1 || got[0] != "dave" {
+		t.Fatalf("expected {dave}; got %v", got)
+	}
+}
+
+// TestFilterBlacklistedDMPeers_BatchErrorFailsClosedForAll — the batch
+// contract fails closed on error: the previous per-peer path could drop
+// just the offender, but one IN query error means we can no longer
+// distinguish blacklisted from non-blacklisted for the batch, so we drop
+// every candidate. This is stricter than the per-peer version and is safe
+// (hidden legit DMs > leaked blacklisted DMs).
+func TestFilterBlacklistedDMPeers_BatchErrorFailsClosedForAll(t *testing.T) {
+	loginUID := "me"
+	uSvc := &stubUserSvc{
+		blacklistErr: errBlacklistDown,
+	}
+	gSvc := &stubGroupSvc{}
+	h := newHandlerForAllowlistTests(uSvc, gSvc)
+
+	got := h.filterBlacklistedDMPeers(loginUID, []string{"a", "b", "c"})
+	if len(got) != 0 {
+		t.Fatalf("batch error must drop every peer fail-closed; got %v", got)
+	}
+}
+
+// TestFilterBlacklistedDMPeers_EmptyInputSkipsService — the batch layer must
+// short-circuit on empty peer input and never touch the (potentially slow)
+// blacklist service. Verified by wiring in a stub that would return an
+// error if called, and asserting no error surfaces.
+func TestFilterBlacklistedDMPeers_EmptyInputSkipsService(t *testing.T) {
+	loginUID := "me"
+	uSvc := &stubUserSvc{
+		blacklistErr: errBlacklistDown, // would fail-closed for every peer
+	}
+	gSvc := &stubGroupSvc{}
+	h := newHandlerForAllowlistTests(uSvc, gSvc)
+
+	got := h.filterBlacklistedDMPeers(loginUID, nil)
+	if len(got) != 0 {
+		t.Fatalf("nil peers must return empty without calling service; got %v", got)
+	}
+	got = h.filterBlacklistedDMPeers(loginUID, []string{})
+	if len(got) != 0 {
+		t.Fatalf("empty peers must return empty without calling service; got %v", got)
+	}
+}
+
+// TestExistBlacklistsBoth_StubSemanticsMatchExistBlacklist — belt-and-braces
+// sanity check that the stub's batch method returns a superset of individual
+// ExistBlacklist calls for the same fixture. If a future test forgets to
+// populate one, the mismatch surfaces here rather than in a downstream
+// integration test.
+func TestExistBlacklistsBoth_StubSemanticsMatchExistBlacklist(t *testing.T) {
+	loginUID := "me"
+	uSvc := &stubUserSvc{
+		blacklist: map[string]bool{
+			"me->x":   true,
+			"y->me":   true,
+			"me->z":   false,
+			"z->me":   false,
+			"me->me":  true, // guarded — self must never surface
+		},
+	}
+	peers := []string{"x", "y", "z", "me", ""}
+	blockedByMe, blockedByPeer, err := uSvc.ExistBlacklistsBoth(loginUID, peers)
+	if err != nil {
+		t.Fatalf("stub ExistBlacklistsBoth returned error: %v", err)
+	}
+	if !blockedByMe["x"] {
+		t.Fatalf("me->x forward edge missing from blockedByMe: %v", blockedByMe)
+	}
+	if !blockedByPeer["y"] {
+		t.Fatalf("y->me reverse edge missing from blockedByPeer: %v", blockedByPeer)
+	}
+	if blockedByMe["z"] || blockedByPeer["z"] {
+		t.Fatalf("z has no edge; must not appear: byMe=%v byPeer=%v", blockedByMe, blockedByPeer)
+	}
+	if blockedByMe[loginUID] || blockedByPeer[loginUID] {
+		t.Fatalf("self edge must be filtered by the batch method: byMe=%v byPeer=%v", blockedByMe, blockedByPeer)
+	}
+	// Cross-check against per-pair ExistBlacklist parity for the surviving
+	// axis: any peer marked by the batch must also be hit by the per-pair
+	// call in the same direction.
+	for _, p := range []string{"x", "y", "z"} {
+		fwd, _ := uSvc.ExistBlacklist(loginUID, p)
+		rev, _ := uSvc.ExistBlacklist(p, loginUID)
+		if fwd != blockedByMe[p] {
+			t.Fatalf("forward parity mismatch for peer=%s: pair=%v batch=%v", p, fwd, blockedByMe[p])
+		}
+		if rev != blockedByPeer[p] {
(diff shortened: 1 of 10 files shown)


OctoBoooot (approved):
## Review: batch DM blacklist gate + per-phase audit timing (#562)

**Verdict:** Approve — a correct, careful batch of the DM bidirectional-blacklist security gate (the #550 chokepoint): identical predicate, fail-closed direction preserved, N per-peer queries → 1 IN query, with adversarial tests for parity/bidirectional/fail-closed. The audit-timing change is passive (measurement only). Byte-verified. Conditional on the in-progress CI (Build green, Vet/Lint/Test running); pr-title-lint green, check-sprint governance.

Risk tier: normal, but this touches the messages_search DM-blacklist access-control gate from #550 — reviewed as security-critical.

### Verified (byte-level) — the batched gate preserves both invariants
- **Bidirectional preserved.** `filterBlacklistedDMPeers` drops a peer on `blockedByMe[peer] || blockedByPeer[peer]` — both directions, same as the per-peer `ExistBlacklist(me,peer) OR ExistBlacklist(peer,me)`. The new `existBlacklistsBoth` query `WHERE blacklist=1 AND ((uid=login AND to_uid IN peers) OR (uid IN peers AND to_uid=login))` captures both directions in one scan; the row→map split attributes each row by which side is loginUID (`r.UID==login → blockedByMe[r.ToUID]`; `r.ToUID==login → blockedByPeer[r.UID]`). Mutual blacklist (two rows) and self-loop are handled: `uniq` strips `p==loginUID` so no `(login,login)` row can ambiguously hit both branches, with `!= loginUID` guards as belt-and-suspenders.
- **Fail-closed preserved (arguably stronger).** On query error, returns nil → drops ALL candidate DM peers. The per-peer path dropped only the erroring peer; the batch correctly reasons that one IN-query error means it can't distinguish blacklisted from not for the whole set, so it drops the whole set — never leaks a blacklisted DM when the table is unreachable.
- **No over-fetch.** `blacklist=1` is in the WHERE (not post-filtered); dbr IN-param expansion consistent with existing QueryFriendsWithToUIDs.
- **Audit timing is passive.** `audit.go` only reads recorded per-phase timing keys from context and emits them — no control-flow change, no security impact (YUJ-27 observability).

### Tests lock the security properties
- `TestFilterBlacklistedDMPeers_BatchParity`: forward hit / reverse hit / no-hit-survives / self+empty-never-hit-DB.
- `TestFilterBlacklistedDMPeers_BidirectionalBothSides`: mutual blacklist drops exactly once.
- `TestFilterBlacklistedDMPeers_BatchErrorFailsClosedForAll`: query error → all peers dropped fail-closed.

### Praise
- This is how to batch a security gate for perf without weakening it: same bidirectional predicate, same (stronger) fail-closed direction, correct row-to-direction attribution with the mutual/self edges reasoned, and a dedicated test for each security property (parity, bidirectional, fail-closed). The #550 blacklist chokepoint is intact.

### Note
- Not mergeable on this approval alone: CI must finish green + review threshold + check-sprint (Sprint field). mergeStateStatus BLOCKED until then.

Jerry-Xin (approved):
**APPROVE** — batched DM blacklist gate is byte-verified semantically equivalent to the per-item version; no bypass, no false-drop, missing-row handled correctly. Perf win (2N→1 MySQL round-trips) with the security invariant preserved and well-tested.

## Security equivalence (old vs new blacklist gate)

**Old** `filterBlacklistedDMPeers` (per peer): `ExistBlacklist(loginUID, peer)` then `ExistBlacklist(peer, loginUID)`, drop on true/err, else keep. Note the old `existBlacklist(uid,toUID)` SQL (`db_friend.go:252`) already matched *either* direction (`((uid=? and to_uid=?) or (uid=? and to_uid=?)) and blacklist=1`), so a peer dropped iff any blacklist edge existed in either direction.

**New** (`search_global.go:838-859`): one batch `ExistBlacklistsBoth(loginUID, peers)` → `blockedByMe` (loginUID→peer) + `blockedByPeer` (peer→loginUID); drop if `blockedByMe[peer] || blockedByPeer[peer]`. The OR of the two directional maps = "any edge in either direction" = **identical drop set** to the old code. ✅ Both directions preserved.

- **Batch SQL** (`db_friend.go:264-983`): `blacklist=1 and ((uid=? and to_uid in ?) or (uid in ? and to_uid=?))` with `(loginUID, uniq, uniq, loginUID)`. Branch-1 rows → blockedByMe, branch-2 rows → blockedByPeer, keyed correctly by `r.UID==loginUID` / `r.ToUID==loginUID`. Dedup + self/empty strip before the query (`db_friend.go:944-956`) prevents a self-row hitting both branches. ✅ Key correctness.
- **Missing-row = allow (correct):** a peer with no blacklist row is absent from both maps → `false || false` → kept. This matches "no blacklist row = not blacklisted"; no false-drop of legit peers, no leak. ✅
- **Fail-closed on error** (`search_global.go:844-851`): batch error drops the *whole* set (`return nil`). Stricter than the old per-peer skip-the-offender, but on the safe axis (hide legit > leak blacklisted); author documents this intentional tightening. ✅
- **Edge cases:** empty/nil input short-circuits before touching the service (`filterBlacklistedDMPeers` len==0 guard + `existBlacklistsBoth` empty guard); self/empty peers stripped. ✅

## enumerateDMPeersTimed re-implementation
`buildAllowlist` now calls `enumerateDMPeersTimed` (`search_global.go:575`), a line-faithful copy of `enumerateDMPeers` with a timing hook around the same `filterBlacklistedDMPeers` call. Logic is identical (same GetFriends error return, peer loop, space-member fail-soft, RemoveRepeatedElement, blacklist gate, early returns, applyDMBotFilter). The plain `enumerateDMPeers` remains test-exercised, so the two won't silently diverge — but the duplication is a maintenance smell (💬 non-blocking).

## Audit / per-phase timing
Pure telemetry (`audit.go`): only int64-ms durations + int counts, plus the pre-existing hashed `keyword_hash`. **No message content, no plaintext UIDs** added. Timing is best-effort via `c.Set/c.Get`; **no timing path can abort the search** — `buildAllowlist` returns `timings` as an additive value and errors propagate exactly as before (`search_global.go:571-599`). Not a new hot-path failure mode. ✅

## Tests (cover the security invariant, not just speed)
`allowlist_blacklist_batch_test.go`: forward-edge drop, reverse-edge drop, no-edge survives, self/empty never survive, both-sides idempotent, **batch-error fails closed for all**, empty-input skips service, legacy fixture parity, and an explicit **batch-vs-per-pair `ExistBlacklist` parity** cross-check. Stub mirrors production directional semantics. ✅

## CI
Build ✅ Vet ✅ Lint ✅ pr-title-lint ✅ code-review ✅ (approved). **Test still in-progress** at review time — ran locally: `go test ./modules/messages_search/... ./modules/user/...` **PASS**, `go vet` clean. `check-sprint` fails but that is governance/sprint-tracking, not a code gate.

⚠️ Caveat: CI **Test** was pending at submission; local run passed. Recommend confirming CI Test green before merge.

## ✅ Highlights
Correct directional decomposition of a bidirectional predicate, disciplined fail-closed, zero-overhead nil-sink timing for single-channel handlers, and a parity test that pins batch==per-pair semantics.

yujiawei (approved):
# Code Review — PR #562 (octo-server)

**Scope reviewed:** `git diff` merge-base `0571a425` → head `24e4007d` (10 files, +604/-57). Build ✅, `go vet` ✅, `go test ./modules/messages_search/...` ✅ (the `modules/user` test failure is an environment-only real-MySQL/Redis gate — `dial 127.0.0.1:3306 refused` — unrelated to this diff).

This is a security-sensitive change to the global-search DM allowlist gate, reviewed with extra care on the blacklist parity axis.

## 1. Spec compliance

**Spec: ✅**

Both title deliverables are implemented and scoped:

- **Batch DM blacklist gate** — `filterBlacklistedDMPeers` (`search_global.go`) now issues one `ExistBlacklistsBoth` `IN` query instead of 2N per-peer `ExistBlacklist` round-trips (`user/db_friend.go` `existBlacklistsBoth`, `user/service.go` wrapper + interface).
- **Per-phase audit timing** — `allowlistTimings` / `searchPhaseTimings`, `paginateWithFilterDepthInstr`, and the new `*_ms` / `oversample_pages` fields on `auditMiddleware`.

No over-build (no unrequested endpoints/flags/fields) and no behavioral deviation from the agreed approach.

**Security parity verified (the crux).** The batch SQL preserves the *exact* bidirectional semantics of the per-peer path it replaces:

- Old per-peer (merge-base): for each peer P, `blacklist=1 AND ((uid=L AND to_uid=P) OR (uid=P AND to_uid=L))`, drop P if either direction hits.
- New batch: `blacklist=1 AND ((uid=? AND to_uid IN ?) OR (uid IN ? AND to_uid=?))`, rows classified into `blockedByMe` (branch A: `uid==L`) / `blockedByPeer` (branch B: `to_uid==L`), drop peer if `blockedByMe[p] || blockedByPeer[p]`.

The two produce the identical drop-set for every reachable row. Self-loops and duplicates are stripped into `uniq` before the query (`db_friend.go`), so a `(L,L)` row cannot be returned or misclassified. `filterBlacklistedDMPeers` still runs **before** the empty-spaceID early return and before the bot/Space gate, so friend-only deployments keep enforcing the gate. No path was found by which a blacklisted DM peer leaks into the global allowlist.

## 2. Code Quality

**Quality: Approved** — no P0/P1. All findings below are P2 (non-blocking; suggested follow-ups).

**P2 — Unbounded `IN` list on the DM allowlist queries (DB scaling).** `existBlacklistsBoth` passes the full `peers` slice (friends ∪ same-Space members) into an `IN (...)` clause with no chunking. For a caller in a very large Space this produces a large query and, absent a supporting index, can degrade. Two mitigating facts keep this off the blocker list: (a) it is **not a regression** — the same unbounded `peers` set already flows into `GetBotUIDs` (`user IN ?`, `pkg/space/query.go`) on this exact path, and the pre-PR code ran 2N serial round-trips over the same set; the batch is strictly *less* DB work (2N → 1). (b) It is a systemic property of the whole global-search allowlist path (`ExistMembersActive`, `GetBotUIDs`, thread enum all scale with membership), better addressed by one hardening pass than by blocking this perf PR. Suggested follow-up: chunk these `IN` lists (batches of ~500–1000) and confirm a composite index on `user_setting(uid, to_uid, blacklist)`.

**P2 — Fail-closed on batch error drops the whole DM peer set.** On `ExistBlacklistsBoth` error, `filterBlacklistedDMPeers` returns `nil` (all DM candidates dropped) rather than the old per-peer "drop only the offender". This is **documented and intentional** (a single `IN` error can no longer isolate one peer) and fails in the safe direction (hides legit DMs rather than leaking blacklisted ones). Worth noting only because, combined with the unbounded-`IN` point above, a query failure for a huge-Space user would silently zero all DM results — another reason to chunk.

**P2 — `enumerateDMPeersTimed` is a hand-copied re-implementation of `enumerateDMPeers`.** Production now calls the timed copy (via `buildAllowlist`), while all unit tests — including the blacklist-parity regression test — still exercise the plain `enumerateDMPeers`. The two are semantically identical today (verified line-by-line: same friend enum, byte-identical fail-soft WARN, shared `filterBlacklistedDMPeers`, same gate ordering), so there is no current divergence. The risk is drift: a future security fix applied to only the tested copy would silently not reach production. Prefer wrapping the single method for timing (e.g. time `filterBlacklistedDMPeers` internally) over forking it.

**P2 — Load-bearing SQL path has no DB-level test.** The real `existBlacklistsBoth` WHERE clause + row→map classification (the security logic) is never run under test; the messages_search tests inject a separate `stubUserSvc.ExistBlacklistsBoth` reimplementation. A future edit that swapped the two branches or dropped the `OR` grouping would pass all tests while leaking. Suggest a `friendDB`-level test seeding `user_setting` rows in both directions plus a self-loop and a duplicate peer, asserting both returned maps.

**P2 — minor:**
- `startToken := time.Now(); _ = startToken` in `enumerateDMPeersTimed` is dead/reserved code — remove or use.
- `dmBlacklistMs` is a `time.Duration` (nanoseconds) despite the `Ms` suffix; correct today but invites a future unit mix-up.
- `allowlist_ms` (`totalMs()`) sums only the four instrumented sub-phases; untimed work inside `buildAllowlist` (e.g. `GetGroupsWithMemberUID`, bot filter) is excluded, so the bucket under-reports true allowlist wall-clock vs `took_ms`. Fine as a component breakdown, just not a full accounting.
- New instrumented timing paths have no test coverage.

## 3. Items a human/DBA should verify manually

- Presence of an index supporting the new batch predicate on `user_setting(uid, to_uid, blacklist)` — static review can't confirm the `IN` won't table-scan.
- The runtime perf claim (per-user blacklist gate dropping from ~seconds to single-digit ms) — observable only in a live/loaded environment.

## Advisory coverage disclosure

This review absorbed one external advisory leg (cross-family, long-context) plus an independent multi-lens adversarial pass; a second (line-level) advisory leg was **absent** this round due to an upstream gateway error and is recorded as absent, not as tacit approval. The unbounded-`IN` concern originated from the advisory leg and was **downgraded from blocking to a P2 follow-up** after independent verification that it is pre-existing on the same path and that this PR reduces (not increases) the DB work.

## Verdict

Spec ✅ and no P0/P1 correctness or security defects — the security-critical blacklist parity is preserved and the change is a net reduction in DB round-trips. Approving. The P2 items (chunk the `IN` lists + verify the index, de-duplicate the timed/plain enumerate methods, add a DB-level parity test) are recommended as a fast follow-up, not merge blockers.

PR #570, opened 2026-07-13
fix(card): trust published App Bots as card senders

### LINKED ISSUE #571: P0: unify App Bot card sender trust
## Problem

The Bot API accepts interactive card sends from published App Bots, but the two downstream trust gates only recognize active `robot` rows:

- server-authored display projection masks the App Bot card's authoritative `plain` text;
- `POST /v1/message/card/action` rejects the App Bot sender before enqueueing `card_action`.

App Bot creation writes `app_bot` plus `user.robot=1`, not a compatibility `robot` row. This leaves the system in a half-supported state.

## Required behavior

- Treat `robot.status=1` and `app_bot.status=1` as the authoritative active bot identities.
- Do not authorize from `user.robot=1` presentation metadata.
- Fail closed on lookup errors or a UID active in both lifecycle tables.
- Preserve the existing display cache, but resolve card-action senders live so unpublish takes effect immediately.
- Preserve all existing card-action channel, Space, membership, visibility, idempotency, rate-limit, and event wire-contract behavior.
- Verify a published App Bot action through event poll and ACK.

## Out of scope

`internal/carddispatch`, internal producers, incoming-webhook action callbacks, and Bot API send-pipeline changes.

Implementation: https://github.com/Mininglamp-OSS/octo-server/pull/570
Spec: `.octospec/tasks/card-message-appbot-trust/brief.md`

an9xyz:
## Summary

Close the P0 trust-chain gap where App Bots could send interactive cards but server-side display masking and card actions only recognized `robot` rows.

## Related Issue

Closes #571

## Linked Spec

- Primary implementation: `.octospec/tasks/card-message-appbot-trust/brief.md`
- Planning-only follow-up: `.octospec/tasks/card-message-internal-dispatch/brief.md`

## Changes

- Add a cache-free `modules/botidentity` resolver over active `robot` and published `app_bot` rows, with same-statement ambiguity detection and fail-closed errors.
- Rewire `modules/cardtrust` to the unified resolver while preserving the existing bounded 60-second display cache and incoming-webhook shortcut.
- Resolve the stored card sender live in `POST /v1/message/card/action`; keep the existing robot event queue and wire contract unchanged.
- Cover App Bot display projection and the full action -> event poll -> ACK lifecycle, including immediate unpublish rejection and republish recovery.
- Record the App Bot trust octospec context, verification evidence, and shared journal.
- Add a separate planning-only P0 brief for `internal/carddispatch`, with a mandatory pilot-producer gate; no dispatcher implementation is included in this PR.

## Testing

- [x] `go test -race ./modules/botidentity/... ./modules/cardtrust/...`
- [x] `go test -cover ./modules/botidentity/...` (96.2% statements)
- [x] `go test -tags integration ./modules/botidentity -run '^TestResolverAgainstAuthoritativeBotTables$' -count=1`
- [x] `go test -tags integration ./modules/message -run '^TestCardAction' -count=1`
- [x] Published App Bot webhook-push and message-search projection tests
- [x] `go test ./modules/bot_api/...`
- [x] `go test ./pkg/cardmsg/... ./modules/cardtrust/...`
- [x] `go vet ./modules/botidentity/... ./modules/cardtrust/... ./modules/message`
- [x] `make i18n-extract-check && make i18n-lint`
- [x] `git diff --check origin/main`

`golangci-lint` is not installed in this local workspace; CI remains the authoritative full lint gate.

## COMPREHENSION

   It replaces the two robot-only card trust checks with one authoritative resolver. A UID is trusted only when `robot.status=1` or `app_bot.status=1`; `user.robot=1` never authorizes. Display consumers retain their existing short cache, while card actions resolve live state before any idempotency claim or event side effect. All channel, Space, membership, visibility, revoke/delete, input, and event ownership gates remain unchanged.

   A resolver query failure now masks display text and returns the existing query-failed envelope for actions, which is intentionally fail-closed. A corrupt UID active in both tables is also rejected instead of receiving an implicit precedence. The new query adds one indexed statement on first display-cache miss and on each first action attempt; no new cache, queue, route, migration, or wire field is introduced.

   The RED checkpoint reproduced the missing resolver and App Bot action rejection. The GREEN suites cover both bot kinds, inactive/missing/presentation-only/ambiguous/error identities, cache retry behavior, push/search authoritative `plain`, and the HTTP/Redis flow: unpublished App Bot rejects with no event, republish accepts exactly once, the same App Bot token polls one `card_action`, ACK removes it, and the next poll is empty. The complete existing `TestCardAction*` matrix remains green.

## Checklist

- [x] I have read `CONTRIBUTING.md`
- [x] Updated documentation

### DIFF
--- a/.octospec/journal/shared/card-message-appbot-trust.md
+++ b/.octospec/journal/shared/card-message-appbot-trust.md
@@ -1,0 +1,79 @@
+---
+type: Journal
+title: "Journal: card-message-appbot-trust"
+description: Close the App Bot card trust gap with one authoritative identity resolver and a verified action event lifecycle.
+tags: ["message", "card", "app-bot", "bot-api", "trust-boundary", "auth", "testing"]
+timestamp: 2026-07-13T16:01:27+08:00
+# --- octospec extension fields ---
+task: card-message-appbot-trust
+source: self
+---
+
+# Journal: card-message-appbot-trust
+
+## What was done
+
+- Added `modules/botidentity`, a live, cache-free resolver over the two lifecycle
+  authorities: `robot.robot_id/status=1` and `app_bot.uid/status=1`. One SQL
+  statement reads both predicates from the same snapshot. Missing/inactive rows
+  resolve to no identity; simultaneous active rows return a typed invariant error.
+  `user.robot=1` is never consulted as authorization.
+- Rewired `modules/cardtrust` to the unified resolver while preserving its
+  bounded LRU, 60-second TTL, `iwh_` display shortcut, positive/negative caching,
+  and no-cache-on-error fail-closed behavior.
+- Rewired `POST /v1/message/card/action` to use the live resolver for the stored
+  sender UID. Existing `robotService` remains the typed-event enqueue path, so
+  the queue and `card_action` wire shape are unchanged.
+- Added regression coverage for User Bot/App Bot/inactive/presentation-only/
+  ambiguous/error identities, push and search display projection, and a complete
+  App Bot `action -> poll -> ACK -> absent-on-repoll` lifecycle. Unpublish blocks
+  a first action immediately; republish permits it without changing idempotency.
+
+## Structural learnings
+
+- Bot presentation and bot authority are different layers. `user.robot=1` is a
+  UI hint; lifecycle authorization must resolve the owning authoritative table.
+  When two independent tables can claim one UID, a precedence rule hides data
+  corruption. A single-statement resolver can detect the invariant and fail closed.
+- Cache placement belongs to the consumer. Display masking tolerates the existing
+  short TTL, while side-effect authorization must read live state. A package-level
+  identity cache would have silently weakened App Bot unpublish semantics.
+
+## octospec rules injected
+
+- **space-isolation**: only the stored sender predicate changed; channel binding,
+  membership, Space/group/thread status, visibility, revoke/delete, and
+  idempotency gates are unchanged and the full card-action matrix stays green.
+- **trust-boundary**: only active User Bots, published App Bots, and existing
+  `iwh_` display identities expose stored `plain`; errors and ambiguous identities
+  fail closed. Webhook actions remain rejected.
+- **error-handling**: resolver errors use the existing localized query-failed
+  envelope; inactive/unknown identities use the existing collapsed invalid action.
+- **rate-limit**: no route, middleware, or Redis frequency-counter change.
+
+## Verification
+
+- TDD checkpoints: `3599b793` (RED reproducers), `2b5f9519` (GREEN fix).
+- `go test -race ./modules/botidentity/... ./modules/cardtrust/...` — PASS.
+- `go test -cover ./modules/botidentity/...` — PASS, 96.2% statements.
+- `go test -tags integration ./modules/botidentity -run '^TestResolverAgainstAuthoritativeBotTables$' -count=1` — PASS.
+- `go test -tags integration ./modules/message -run '^TestCardAction' -count=1` — PASS after CI-style MySQL/Redis reset.
+- Published App Bot push/search projection tests — PASS after per-package reset.
+- `go test ./modules/bot_api/...` and `go test ./pkg/cardmsg/... ./modules/cardtrust/...` — PASS.
+- `go vet`, `make i18n-extract-check`, `make i18n-lint`, and `git diff --check` — PASS.
+
+## Gotchas worth remembering
+
+- The repo's shared local `test` schema must be dropped/recreated between module
+  packages, matching `.github/workflows/ci.yml`; otherwise one package's partial
+  migration set causes another package to fail with `unknown migration in
+  database`. Redis must also be flushed for rate-limit/action/event isolation.
+- `golangci-lint` was not installed in this workspace. The focused vet, repo i18n
+  linters, race tests, integration suites, and diff check ran successfully; CI
+  remains the authoritative full golangci gate.
+
+## Deliberately deferred
+
+`internal/carddispatch`, internal producers, incoming-webhook action/callback
+support, and bot-send pipeline changes remain out of scope until a concrete
+internal caller is specified.
--- a/.octospec/log.md
+++ b/.octospec/log.md
@@ -4,6 +4,20 @@ Change history for this repo's `.octospec/`, following the
 [OKF](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md)
 change-log convention (§7). Newest first.
 
+## 2026-07-13 (card-message-appbot-trust)
+
+- **Fix** — Closed the P0 App Bot card trust split without changing the send
+  pipeline: added a cache-free `modules/botidentity` authority over active
+  `robot` and published `app_bot` rows (same-statement ambiguity detection,
+  `user.robot` never authorizes), moved `cardtrust` display masking onto it while
+  retaining the 60-second bounded cache, and made `card/action` resolve sender
+  identity live before enqueueing through the unchanged robot event queue. Added
+  push/search projection coverage plus App Bot unpublish/republish and full
+  action -> poll -> ACK lifecycle tests. `internal/carddispatch` remains a
+  separate task. Brief/context under
+  `.octospec/tasks/card-message-appbot-trust/`; shared journal
+  `.octospec/journal/shared/card-message-appbot-trust.md`.
+
 ## 2026-07-09 (sticker-oversized-store-guard)
 
 - **Fix** — Task `sticker-oversized-store-guard` (code-review fix on
--- a/.octospec/tasks/card-message-appbot-trust/brief.md
+++ b/.octospec/tasks/card-message-appbot-trust/brief.md
@@ -1,0 +1,218 @@
+---
+type: Task
+title: "Task: card-message-appbot-trust"
+description: Make published App Bots first-class trusted card senders across display masking and the card-action event loop.
+tags: ["message", "card", "bot-api", "app-bot", "wire-contract", "trust-boundary", "auth", "space", "isolation", "acl", "error-response", "i18n", "rate-limit", "testing"]
+timestamp: 2026-07-13T00:00:00+08:00
+# --- octospec extension fields ---
+slug: card-message-appbot-trust
+upstream: self
+source: self
+---
+
+# Task: card-message-appbot-trust
+
+> One task = one `.octospec/tasks/<slug>/` directory. This brief is the spec for
+> the work. AI may draft it from existing code; a human confirms it.
+
+## Goal
+
+Close the existing App Bot card-sender trust gap without changing the card send
+pipeline, target authorization, Space enrichment, or wire contract.
+
+A published App Bot can already authenticate to `/v1/bot/*`, send type-17 card
+payloads, poll `/v1/bot/events`, ACK events, and edit its own cards. App Bot
+creation persists an `app_bot` row and a `user.robot=1` presentation row, but
+does not create a `robot` table row. Two server-side card trust gates currently
+call `robot.ExistRobot` and therefore reject/mask App Bot cards after send:
+
+- `modules/cardtrust.Resolver`, used by offline push and message-search card
+  projections;
+- `POST /v1/message/card/action`, before enqueueing `card_action` to the sender's
+  bot event queue.
+
+Introduce one authoritative bot-identity resolver that recognizes both active
+User Bots and published App Bots, then use it at those two trust gates. A valid
+App Bot card must retain trusted `plain` on server-authored display surfaces and
+deliver a valid Submit action through the existing App Bot event poll/ACK loop.
+
+User journey:
+
+- As a published App Bot owner, I can send an interactive DM card; the recipient
+  sees the trusted card fallback text, taps a valid action, and my App Bot polls
+  exactly one `card_action` event and ACKs it through existing Bot API routes.
+
+## Background
+
+### Existing identity models
+
+- User Bot authority: `robot.robot_id` with `robot.status=1`.
+- App Bot authority: `app_bot.uid` with `app_bot.status=1` (`StatusPublished`).
+- App Bot presentation: `user.uid=app_bot.uid`, `user.robot=1`.
+- Incoming webhook identity: synthetic `iwh_` prefix, intentionally trusted for
+  display but intentionally rejected by `card/action` because it has no bot event
+  consumer.
+
+Bot API authentication already distinguishes User Bot and App Bot tokens. The
+App Bot event routes are UID-keyed Redis operations; typed events do not require
+a `robot` DB row. The missing component is a shared authoritative predicate for
+“is this UID currently an active bot identity that may own a card action?”.
+
+### Decisions locked by this task
+
+1. **Complete App Bot card support.** Do not disable App Bot cards or hide the
+   card profile. Active User Bots and published App Bots are trusted bot card
+   senders.
+2. **One resolver, no HTTP-module dependency.** Add a small library package
+   (proposed name `modules/botidentity`) that reads the existing `robot` and
+   `app_bot` tables without importing `modules/robot`, `modules/app_bot`, or
+   `modules/bot_api`. This avoids import cycles and keeps lifecycle authority in
+   the existing tables.
+3. **Fail closed.** Empty, unknown, disabled/unpublished, deleted, ambiguous, or
+   lookup-error identities are not trusted. A UID simultaneously active in both
+   bot tables is an invariant violation and returns an explicit internal error;
+   no precedence rule silently chooses one kind.
+4. **Live action authorization.** `cardAction` resolves sender identity on every
+   first action attempt. It does not use the presentation cache, preserving
+   immediate App Bot unpublish/revocation for side effects.
+5. **Presentation cache stays bounded.** `modules/cardtrust` retains its current
+   LRU capacity, 60-second TTL, webhook-prefix shortcut, and “do not cache lookup
+   errors” behavior. The cached value now comes from the unified resolver.
+6. **Event transport is unchanged.** A valid App Bot action is enqueued by the
+   existing `robotService.EnqueueBotTypedEvent(msgM.FromUID, ...)`, then polled
+   and ACKed through existing `/v1/bot/events` routes. No new queue or event
+   schema is introduced.
+7. **No behavior change for other senders.** Active User Bot and `iwh_` display
+   trust remains unchanged; webhook actions, human-forged cards, and inactive
+   bot actions remain rejected.
+
+## Load-bearing list
+
+<!-- touches: trust-boundary, wire-contract, auth, bot-api, space, isolation,
+     acl, error-response, i18n, rate-limit, testing -->
+
+- **Bot identity trust (`trust-boundary`, `auth`)** — `robot.status=1` and
+  `app_bot.status=1` are the only bot-table authority states. `user.robot=1`
+  alone is presentation metadata and must not authorize an action.
+- **Display masking (`trust-boundary`, `wire-contract`)** — type-17 stored
+  `plain` may surface only for an active User Bot, published App Bot, or existing
+  `iwh_` sender. Human/unknown/inactive senders and lookup errors remain masked
+  to `[卡片]`.
+- **Card action side effects (`acl`, `space`, `isolation`)** — only the sender
+  identity predicate changes. Existing anti-IDOR channel binding, operator
+  membership, group/thread status, canonical visibility, revoke/delete/expiry,
+  action-id, input-shape, rate-limit, and idempotency gates must remain byte-for-
+  byte behaviorally unchanged.
+- **Event ownership (`bot-api`, `wire-contract`)** — a successful action is
+  enqueued to the stored message `from_uid`; it must be retrievable only by that
+  authenticated App Bot and retain the existing additive `card_action`
+  `event_data` shape.
+- **Revocation semantics** — App Bot unpublish/delete must block new action side
+  effects immediately through the live resolver. Presentation masking may lag
+  by the existing cache TTL, matching current User Bot cache semantics.
+- **Failure handling (`error-response`, `i18n`)** — resolver/storage failures in
+  `cardAction` use the existing localized query-failed envelope; an inactive or
+  untrusted sender uses the existing collapsed card-action-invalid response.
+  No new raw HTTP error path or new client-visible identity enumeration is added.
+- **Rate limiting (`rate-limit`)** — no route or middleware changes. The existing
+  authenticated `card/action` shared UID limiter and Bot API/global limits stay
+  in force; no new Redis request-frequency counter is introduced.
+- **Testing (`testing`)** — resolver behavior, cardtrust cache semantics, live
+  action authorization, App Bot event poll/ACK, and unchanged negative sender
+  cases require focused tests. DB/Redis-backed tests clean state and reset any
+  shared rate-limit buckets they touch.
+
+## Out of scope
+
+- `internal/carddispatch`, internal producer registry/configuration, direct-send
+  source guards, or onboarding an internal business producer. These become a
+  separate task when the first concrete internal caller and target capability
+  are known.
+- Refactoring `/v1/bot/sendMessage`, its validation order, target authorization,
+  Space resolution, mention handling, metrics, or IM dispatch.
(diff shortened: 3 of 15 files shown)


OctoBoooot (approved):
## Review: fix(card): trust published App Bots as card senders (#570)

**Verdict:** Approve, conditional on Test/Lint going green (Build/Vet green). Risk tier: **high** (trust-boundary change — who is a trusted card sender); reviewed at that rigor. The new `botidentity` resolver is fail-closed, reads only authoritative tables, and every call site handles its error correctly. needs-human-review applies (high tier). check-sprint red is governance.

### Trust core — byte-verified sound
- `botidentity.Resolver` (resolver.go) reads ONLY `robot.status=1` and `app_bot.status=1` in one statement; `user.robot` is never queried (doc: "presentation metadata, not an authorization source"). The body's "`user.robot=1` never authorizes" claim holds.
- **Fail-closed everywhere:** lookup error → returns the error (never a bare `false`); `nil` resolver/store → `ErrResolverUnavailable`; `Active()` preserves errors ("must not collapse an error into an unqualified false"). `Resolve` returns `(nil,nil)` only for a genuine non-bot.
- **Ambiguity → reject:** a UID active in both tables → `ErrAmbiguousIdentity`, no precedence rule, single-statement snapshot. Correct — no implicit trust on a broken uniqueness invariant.

### Call sites — all handle the error correctly (no fail-open)
- **Display masking (`cardtrust.Trusted`):** fail-closed on error (`return false`), and critically the error result is NOT cached — only a *successful* resolve populates the 60s LRU, so a transient DB error can't poison the cache with a stale verdict. Webhook `iwh_` synthetic senders trusted by construction (server-generated, unforgeable).
- **Card action (`api_card_action.go` D3③):** re-resolves the stored sender LIVE (no display cache) BEFORE the idempotency claim and any event side-effect. `botIdentity==nil` → 500; `Resolve` error → 500; `senderIdentity==nil` (unpublished/revoked App Bot, webhook, or human) → reject. This delivers the "unpublish immediately blocks side effects" guarantee.
- Grepped all changed files for error-swallowing (`_ = err`, ignored-error `Active`/`Resolve`): none. Every resolver call is `x, err := ...` with a fail-closed branch.

### Defense-in-depth on the action path (verified present)
- Effective-frame `action_id` validation (late clicks on rewritten/removed buttons → 400); revoke/is_deleted gating; operator-local-delete; and **canonical visibility alignment (D3④, "PR#548 review P1")** — prevents a member excluded by `visibles`/history-cleanup from enumerating message_ids to trigger invisible-card bot side effects, merged to a single anti-enumeration `invalid`. This closes exactly the invisible-target-IDOR class that matters for an action endpoint.

### Tests
- `resolver_test.go` covers the security-critical paths: ambiguous-identity (`ErrAmbiguousIdentity`), DB-error fail-closed, and `Active` error-preservation. Body reports 96.2% statements on botidentity + integration tests for the full action→event→ACK lifecycle incl. immediate-unpublish rejection and republish recovery, and App Bot display projection.

### Minor (non-blocking)
- The resolver adds one indexed statement per display-cache-miss and per first action attempt. Bounded (the display cache absorbs repeat reads; actions are inherently rate-limited by user clicks). Two `EXISTS` on indexed columns (`robot.robot_id`, `app_bot.uid`); confirm both are indexed at scale, but not a concern at current load.

### Praise
- This is how a trust-boundary change should be built: a single authoritative resolver (no more scattered robot-only checks), fail-closed by construction with the error type preserved so a caller *can't* accidentally fail open, ambiguity treated as reject-not-precedence, and the high-stakes action path re-resolving live before side effects rather than trusting a cached/stored verdict. The `user.robot`-is-not-authorization separation is called out explicitly in the package doc. The action path also carries the full visibility/revoke/enumeration gate set, so widening the trusted-sender set doesn't widen the side-effect surface.

### CI
- Build/Vet green; Test/Lint in-progress. Approval conditional on them going green (a trust-change's tests are load-bearing). check-sprint governance-red. needs-human-review applied (high tier). mergeStateStatus stays BLOCKED until CI + approvals.

yujiawei (dismissed):
# Code Review — PR #570 (octo-server)

**Scope reviewed:** head SHA `31cfe49233c8adc86a698d59fae0b8751f2da079`, merge-base `76d31cca`, `git diff origin/main...HEAD` (14 files, +1040/-53). This is a security-sensitive change to a trust boundary, so it received extra scrutiny plus a second independent review pass.

## 1. Spec compliance

The change is spec-driven (`.octospec/tasks/card-message-appbot-trust/brief.md`). Verdict: **compliant**.

- **Nothing missing.** All three robot-only card trust gates are migrated to the unified resolver: offline push (`modules/webhook/common.go:472`), message-search projection (`modules/messages_search/api.go:92` → `search_messages.go:305`), and `POST /v1/message/card/action` (`modules/message/api_card_action.go:99`). The new `modules/botidentity` resolver reads only `robot`/`app_bot` and never imports `modules/robot`/`modules/app_bot`/`modules/bot_api`, satisfying the no-cycle constraint.
- **Nothing extra.** No new route, migration, wire field, queue, or event schema. The `card_action` `event_data` shape is unchanged; enqueue still goes through the existing `robotService.EnqueueBotTypedEvent(msgM.FromUID, ...)` (`api_card_action.go:248`). `internal/carddispatch` remains untouched, as scoped.
- **No deviation.** `user.robot=1` is never an authorization source — only `robot.status=1` or `app_bot.status=1`. The `robot` half of the resolver query (`robot_id=? AND status=1`) is behaviorally identical to the old `ExistRobot`, so User Bot and `iwh_` behavior is unchanged. Display masking keeps its bounded 60s LRU cache; `card/action` intentionally resolves live (no cache) so App Bot unpublish/revoke blocks side effects immediately.

## 2. Code quality

Verdict: **Approved.** No P0/P1 findings.

Strengths worth calling out for a security-sensitive change:

- **Fail-closed is airtight.** Empty UID, nil resolver, unknown, disabled/unpublished/draft, deleted, and lookup-error identities all resolve to untrusted. The cross-table ambiguity invariant returns an explicit `ErrAmbiguousIdentity` rather than applying a silent precedence rule (`resolver.go:88`); in `card/action` that error takes the `err != nil` branch and returns the localized query-failed envelope — it cannot fail open. The single-statement `EXISTS(...) , EXISTS(...)` read gives both predicates one snapshot, avoiding a two-query race.
- **Lookup errors are deliberately not cached** (`cardtrust.go`), so a transient DB blip cannot pin a legit bot's cards to `[卡片]` for the whole TTL — this matches the spec's locked decision.
- **Test coverage is strong and load-bearing.** `botidentity` unit + DB-backed integration tests cover every acceptance case, including the security-critical `user.robot=1`-presentation-only → untrusted and ambiguous → error (`resolver_integration_test.go`). The `card/action` integration test exercises the full lifecycle: unpublish blocks a first attempt and enqueues nothing, republish accepts exactly once, the App Bot polls one `card_action` and ACKs it, and a later poll is empty. Push and search projection tests add published-App-Bot and presentation-only-untrusted cases. Package coverage is 96.2% (spec floor 80%).

### Non-blocking advisory notes (P2)

1. **Error-state thundering herd during a DB outage** (`cardtrust.go`). Because lookup errors are intentionally never cached, a large offline-push fan-out or search projection during a DB outage will re-query per card. This is a pre-existing property of the old `ExistRobot` path and is explicitly mandated by the design (errors must be retried, not cached), so it is not a regression — noted only as an operational consideration if these batch paths ever handle very large un-paginated sets.
2. **DB query has no bounded context** (`resolver.go:52`). `SelectBySql(...).LoadOne(...)` runs without a request-scoped context, so a hung connection could block the handler goroutine. This mirrors the existing `robot`/`app_bot` DB layer (codebase-wide plain `.Load()`/`.LoadOne()`), so again not a regression introduced here; a repo-wide `WithContext` adoption would be the right place to address it.

## 3. Overall verdict

**APPROVE.** Spec-compliant with no over/under-build, fail-closed trust boundary, and thorough tests. The two P2 notes are pre-existing patterns, not blockers.

## 4. Suggestions (optional, non-blocking)

- Consider a follow-up to thread a request context into the `botidentity` DB read as part of a broader DB-layer context adoption.

## 5. Additional observations (outside this PR's scope)

- App Bots are now first-class trusted card senders, but `@所有 AI` mention expansion (`modules/message/mention_expand.go`, and the parallel webhook/bot_api paths) still resolves bot membership via `robot.ExistRobot`, which recognizes User Bots only. That means a published App Bot would not be included in an `@所有 AI` broadcast. This is a pre-existing, broader identity inconsistency explicitly out of scope for this fix, but worth tracking as a separate task if App Bots are meant to be full peers of User Bots.

### Review method note

This review combined a primary line-level pass with a second independent long-context pass over the whole diff. A cross-family reviewer flagged the two P2 items above (error non-caching, missing query context); both were independently confirmed as pre-existing patterns rather than regressions. One automated leg failed on an unrelated infrastructure error and is recorded as absent — it did not contribute findings and was not treated as tacit approval.

OctoBoooot (commented):
## Adjudicating yujiawei's operational 🟡s on #570 (my APPROVE stands — both valid, non-blocking)

yujiawei raised two operational 🟡 I under-weighted (I focused my high-tier pass on the trust/security correctness). Both byte-verify as accurate; both are non-blocking hardening. Crediting them.

### 🟡1 — resolver query has no request context (valid, minor/consistency)
- `dbActiveKindStore.activeKinds` uses `s.session.SelectBySql(...).LoadOne()` with no `WithContext`/deadline — a slow query can't be cancelled by the request's context. Confirmed.
- Calibration: it's one statement of two indexed `EXISTS` (robot_id, app_bot.uid), and it matches the surrounding codebase convention (the document/group DB helpers use bare `session.SelectBySql` too) — so it's a consistency/robustness minor, not a #570-introduced regression. Worth threading `context` through when the module's DB layer adopts it broadly; not a blocker on its own.

### 🟡2 — no negative caching → query amplification on sustained DB failure (valid, resilience minor)
- Confirmed and it's the more interesting one: `cardtrust.Trusted` deliberately does NOT cache the error result (correct — caching a transient-error verdict would poison the 60s window with a false untrusted). yujiawei's flip side is real: under a sustained DB outage, every card display re-queries (no negative/backpressure cache) → amplification against an already-failing DB.
- Calibration: this is the same per-cache-miss query volume the display path had pre-PR (the resolver adds one indexed statement, not a new N), and card display isn't a hot loop — so it's a resilience 🟡, not a blocker. The right fix if it ever bites is a short negative-result TTL (cache "lookup failed → untrusted" for a few seconds) that trades a tiny freshness window for backpressure — but that reintroduces the error-poisoning risk the current design deliberately avoids, so it's a genuine tradeoff to decide deliberately, not an obvious fix. Follow-up, not merge-blocker.

### Verdict: APPROVE stands
- Neither 🟡 is a blocker; both are accurate operational hardening notes on axes (request-context propagation, DB-failure backpressure) that complement — not contradict — the trust/security correctness I verified (fail-closed, no fail-open, live re-resolve before side effects, ambiguity-reject). Credit yujiawei for the resilience-dimension catch. High-tier needs-human-review remains applied; the human reviewer may want to weigh the negative-caching tradeoff explicitly.

OctoBoooot (dismissed):
Re-registering my APPROVE vote (my latest review was a COMMENT adjudicating yujiawei's two operational 🟡, which are non-blocking — not a downgrade). Verdict stands: the trust-boundary fix is byte-verified sound — botidentity resolver fail-closed, reads only robot/app_bot status=1 (never user.robot), ambiguity-reject; all call sites handle the error without fail-open; card/action re-resolves live before side effects with full visibility/revoke/enumeration gating. yujiawei's 🟡 (no request-context on the query; no negative-cache backpressure on sustained DB failure) are accurate resilience follow-ups, non-blocking. needs-human-review applies (high tier) — this still requires human sign-off before merge.

OctoBoooot (approved):
## Delta re-review: #570 @ 05c9ba5f (prior @ 31cfe492) — doc-only change, APPROVE re-affirmed

**Verdict:** Approve (re-affirming my 31cfe492 APPROVE), conditional on CI. The delta is a single new spec brief; zero code/test change. needs-human-review retained (high tier).

### Delta is documentation-only (blob-verified)
- The only file in `31cfe492...05c9ba5f` is `.octospec/tasks/card-message-internal-dispatch/brief.md` (+415, a planning brief for the deferred `internal/carddispatch` follow-up). `ahead 1, behind 0` — one doc commit.
- Every trust-critical source file is byte-identical to my APPROVE head: `botidentity/resolver.go`, `cardtrust/cardtrust.go`, `message/api_card_action.go`, `message/api.go` — all SAME. All six test files — SAME. So the entire trust-boundary verification from 31cfe492 holds unchanged:
  - resolver fail-closed, reads only `robot`/`app_bot` status=1 (never `user.robot`), ambiguity-reject;
  - all call sites handle the error without fail-open (display masking non-poisoning cache; card-action live re-resolve before side effects);
  - full visibility/revoke/enumeration gating on the action path;
  - security-critical test coverage present.
- No new attack surface — a spec doc adds none.

### Carried non-blocking 🟡 (unchanged, from yujiawei — credited)
- Resolver query has no request-context deadline (consistency minor, matches codebase convention).
- No negative-cache backpressure on sustained DB failure (resilience follow-up; the deliberate no-error-caching design trades backpressure for freshness — a tradeoff for the human reviewer to weigh, not a blocker).

### Verdict
- APPROVE holds — code unchanged from the verified head, delta is a follow-up spec brief. Conditional on Build/Test/Vet green; needs-human-review applies (high-tier trust-boundary change still requires human sign-off). mergeStateStatus stays BLOCKED until CI + a 3rd approval + human review.

Jerry-Xin (approved):
Well-hardened trust-boundary change. Byte-verified the two 🔴-class risks for a "trust more senders" PR — the trust signal is server-authoritative (not spoofable) and the resolver is fail-closed. No blocking issues. Concurring with the existing approvals.

## Trust signal is authenticated, not client-supplied (not spoofable)
`api_card_action.go:99` resolves trust on `msgM.FromUID` — the STORED message's sender, obtained via `authorizeCardChannelMember` (`:84`, anti-IDOR channel+message binding), never a client-provided sender/bot_id. The request only carries `MessageID/ChannelID/ActionID/ClientToken/Inputs`; action data is server-extracted from the stored effective payload (`effective := msgM.Payload` `:198` → `cardmsg.SubmitAction(effective, req.ActionID)` `:202`), so a client can't forge either the sender or the action.

## Resolver is fail-closed — no fail-open path
`botidentity/resolver.go:54-55` — trust = a single-statement snapshot:
```sql
EXISTS(SELECT 1 FROM robot   WHERE robot_id=? AND status=1) AS user_bot,
EXISTS(SELECT 1 FROM app_bot WHERE uid=?      AND status=1) AS app_bot
```
on authoritative tables (comment: `user.robot` is presentation metadata, explicitly NOT an auth source). Every non-happy path fails closed: lookup error → propagate err (`:59/:88`), never `true`; `uid==""`/not-found → nil (untrusted `:81/:99`); nil store → `ErrResolverUnavailable` (`:84`); active in BOTH tables → `ErrAmbiguousIdentity` (`:91`), not a precedence guess. `cardtrust.Trusted` is **fail-closed and does not cache the error verdict** (cardtrust.go:91-94) so a transient DB blip can't mask a legit bot into a false-trust cache; failed lookups never cached; the `iwh_` webhook synthetic-sender trust is preserved.

## Widening does not drop a prior guard
OLD: `robot.ExistRobot(uid)` (active user bots). NEW: `botidentity.Resolve` (active user bot OR published app_bot status=1). It ADDS app_bot as an authoritative trust source via a new lookup; all pre-existing gates (channel binding, member ACL, revoke/delete, viewer visibility, idempotency claim, action_id-in-effective-frame, input validation) are retained. Card-action re-resolves trust LIVE on every first click BEFORE the idempotency claim/enqueue, so an unpublished/revoked App Bot is blocked immediately (test `api_card_action_test.go:287` flips `app_bot status=2` live → rejected).

## Deny-negative coverage is thorough ✅
- `cardtrust_test.go`: human→false; DB error→fail-closed & not cached (recovers after); empty/nil/ambiguous→false.
- `resolver_test.go`: missing→nil; ambiguous→ErrAmbiguousIdentity; lookup failure→err.
- `api_card_action_test.go`: live unpublish→rejected; non-bot reject + IDOR.
- `webhook/common_card_test.go:74`: presentation-only identity→false, regular human (spoofable type-17)→false, human push masked to `[卡片]` — the exact spoofing threat, correctly denied.

## 🟡 Non-blocking
- Stale comment wording "one robot table query" in `messages_search/api.go:27` / `webhook/common.go:461` (now unified bot identity).
- Integration tests hand-roll a reduced `app_bot` schema rather than the real migration — fine now, could drift.
- (peer notes) DB query lacks request-context; DB-failure retry could storm — worth a follow-up, not gating.

✅ Approving.

OctoBoooot (commented):
## Adjudicating Jerry-Xin's three 🟡 on #570 (my APPROVE stands — all non-blocking, one credited-with-nuance)

Byte-checked Jerry-Xin's three 🟡. None is a correctness/security defect; all are test/doc-fidelity follow-ups. The resolver query is verified correct against the REAL schema. My APPROVE holds.

### 🟡 integration-test uses a simplified app_bot schema (valid test-fidelity minor)
- Confirmed: `TestResolverAgainstAuthoritativeBotTables` hand-rolls `CREATE TABLE app_bot (uid TEXT PRIMARY KEY, status INTEGER)` (isolated SQLite) rather than running the real migration — a deliberate isolation choice (its comment: "isolated real SQL database rather than the process-wide MySQL test ... avoids cross-package migration/cleanup races").
- Drift risk is real but bounded: if a future migration changed `app_bot.uid`/`status`, this test would keep passing against its stand-in table. So it validates resolver *logic*, not schema-fidelity.
- Crucially, the resolver query IS correct against prod: real `app_bot` (migration 20260505000001) is `uid VARCHAR(40) UNIQUE NOT NULL` + `status TINYINT` (0=draft/1=published/2=unpublished); the resolver's `WHERE uid=? AND status=1` selects *published* bots — exactly the "trust published App Bots" semantic. `robot` has robot_id+status. No prod mismatch. So the 🟡 is a test-net-fidelity follow-up (run one case against the migrated schema), not a live bug.

### 🟡 stale comment (doc nit)
- Fair — worth a pass to align comments with the shipped resolver. Non-blocking.

### 🟡 DB query lacks request context / retry-storm (already credited to yujiawei)
- Same resilience follow-up I adjudicated earlier: context-less query (consistency minor, matches codebase convention) + no negative-cache backpressure on sustained DB failure (deliberate no-error-caching tradeoff). Non-blocking.

### Verdict: APPROVE stands
- All three 🟡 are accurate, all non-blocking (two test/doc-fidelity, one resilience). The trust-boundary correctness I verified (fail-closed resolver, authoritative-tables-only, no fail-open at call sites, live re-resolve before side effects) is unaffected — and the resolver query is confirmed correct against the real app_bot/robot schema. Credit Jerry-Xin for the test-schema-drift catch; worth a follow-up test against the migrated schema. needs-human-review remains applied (high tier).

OctoBoooot (approved):
Re-registering APPROVE (my preceding COMMENT adjudicated Jerry-Xin's three 🟡 as non-blocking — not a downgrade). Trust-boundary fix verified sound; resolver query confirmed correct against the real app_bot/robot schema (status=1=published); the 🟡s are test/doc-fidelity + resilience follow-ups. needs-human-review applies (high tier).

yujiawei (approved):
# Code Review — PR #570 (octo-server)

**Verdict: APPROVED** · Reviewed against head `05c9ba5f49cb0a08c5f91651076b402272d2c4bf`, merge-base `76d31cca`.

Security-sensitive review of a trust-boundary change that promotes published App Bots to first-class card senders. I traced the full trust chain, the App Bot action → event → poll → ACK loop end-to-end, and the asymmetry against the removed robot-only `ExistRobot` checks. This is a tight, well-scoped fix (4 production files) that closes a real gap without changing the send pipeline, wire contract, or authorization gates.

## 1. Spec compliance

**Compliant ✅** — Verified against `.octospec/tasks/card-message-appbot-trust/brief.md`.

- **No under-build.** All locked decisions are implemented: (D2) new `modules/botidentity` reads only `robot` + `app_bot`, importing neither `modules/robot`/`modules/app_bot`/`modules/bot_api` (no cycle); (D3) fail-closed on empty/unknown/inactive/ambiguous/lookup-error; (D4) `cardAction` resolves live on every first attempt with **no cache** — immediate unpublish enforcement; (D5) `cardtrust` keeps its LRU/60s TTL/`iwh_` shortcut/"do not cache errors" behavior, now sourced from the unified resolver; (D6) event transport unchanged — enqueue still via `robotService.EnqueueBotTypedEvent`.
- **No over-build.** `internal/carddispatch` ships **zero** implementation — only a planning brief, exactly as the "planning-only" decision requires. No new migration, route, cache, queue, wire field, or metric. Diff is `botidentity/resolver.go`, `cardtrust/cardtrust.go`, `message/api.go`, `message/api_card_action.go` plus tests/spec.
- **No deviation.** `user.robot=1` presentation rows never authorize (verified by both the resolver query and an explicit integration case); User Bot semantics are byte-for-byte preserved (see below); `iwh_` display trust and card-action rejection are unchanged.

## 2. Code quality

**Approved** — no P0/P1.

Key verifications:

- **User Bot semantics preserved (no regression).** The resolver's robot predicate `SELECT EXISTS(SELECT 1 FROM robot WHERE robot_id=? AND status=1)` (`resolver.go:51`) is identical to the removed `robotDB.exist` (`robot/db.go:34`, `robot_id=? and status=1`). App Bot is purely additive.
- **Ambiguity is fail-closed and race-free.** Both `EXISTS` predicates are evaluated in one statement (one snapshot), so a UID active in both tables returns `ErrAmbiguousIdentity` with no precedence rule and no two-query TOCTOU. `Resolve` returns the error (never a silent `false`); `Active()` preserves it too and is not used on any production path.
- **App Bot event loop works end-to-end for a UID with no `robot` row.** Enqueue is UID-keyed Redis (`robotEvent:<uid>`, `robot/api.go:229`), and `/v1/bot/events` polls the same key (`bot_api/events.go` `getEventsResult`). A typed `card_action` event has `Message==nil`, so it passes the App-Bot DM-only filter. The integration test `TestCardActionAppBotEventPollAndAckLifecycle` proves the whole chain against real DB+Redis: unpublish → 400 with zero enqueued; republish → `{accepted:true,replay:false}` + exactly one event; poll returns it; ACK removes it; re-poll empty.
- **Display-plain surface coverage is complete.** The only two "describe-by-plain" surfaces (offline push `webhook/common.go:197`, message-search `search_messages.go:305`) both route through the single `cardtrust.Resolver`, so App Bots are covered automatically. The pin tip always renders `[卡片]` (no plain trust decision) and is unaffected; conversation last-message / quote return raw payload governed by the client-render gate, not a server text surface.
- **No stale robot-only trust gate.** Remaining `ExistRobot` callers (mention `@AI` expansion, reminders) are not card-sender-trust gates and are explicitly out of scope — correctly untouched.
- **No new forgery vector.** Card `from_uid` is set from the authenticated bot context on send (`bot_api/send.go`), not client-supplied; App Bot cards pass the same `Finalize` (server-authoritative `plain`). The action endpoint's IDOR/channel-binding/membership/group-thread-status/visibility/revoke/delete/idempotency gates are unchanged — only the sender-identity predicate was swapped. `TestCardActionTrustModel` regression-tests all of these (human sender, forged action_id, stale frame, cross-channel IDOR, iwh_, revoke/delete/local-delete, oversized body) and asserts zero spurious events.

Build + focused suites green locally: `go build` of the three affected packages, and `go test ./modules/botidentity/... ./modules/cardtrust/...` (DB/Redis-backed suites require infra and were verified by reading — coverage is thorough).

- **Cache-miss query coalescing (`cardtrust.go:107`).** Concurrent cache misses for the same `fromUID` (e.g. large-group offline-push fan-out) each issue an independent `Resolve` DB read. A `singleflight.Group` around the lookup would collapse the brief stampede. Not a correctness or security issue — the query is cheap and idempotent, the LRU already bounds steady state, and fail-closed-on-error is preserved. Optional follow-up.

## 3. Notes for a human verifier

- Confirm production DB integrity actually upholds the "a UID is never simultaneously `robot.status=1` and `app_bot.status=1`" invariant. The resolver correctly rejects violations at read time, but there is no DB-level constraint; a pre-existing dirty row would fail-closed (deny display + actions) rather than fail-open, which is the safe direction.
- The `internal/carddispatch` brief describes a future direct-send producer path with its own trust model. It is planning-only here; that path will need its own security review before any implementation lands.

PR #577, opened 2026-07-13
feat: internal card dispatch boundary — sole trusted in-process type-17 origination (dormant foundation)

an9xyz:
## Summary

Adds `internal/carddispatch` as the **only** supported path for an in-process
business module to originate an `InteractiveCard` (`type=17`) message. The
dispatcher binds a reviewed producer capability to one configured active bot
identity, authorizes the exact Space/target from live DB state, owns the card
envelope, runs `cardmsg.Validate → authoritative Space enrichment → Finalize →
final size recheck` in a fixed order, makes at most one `SendMessageWithResult`
attempt, and emits bounded metrics/logs.

This is a **P1 dormant foundation**: no production producer is registered
(`NewRegistry(deps, nil)`), so deployment is behaviourally inert except one
runtime gate — `/v1/internal/notify[/batch]` now rejects card-shaped payloads
(Decision 14), closing the hole where an internal-token holder could emit a
trusted type-17 map bypassing every card gate. Enabling the `summary-notify`
pilot is a separate P2 PR (see the handoff doc).

## Related Issue

Refs #571

## Linked Spec

`.octospec/tasks/card-message-internal-dispatch/brief.md` (spec + locked
decisions). Next-stage plan: `.octospec/tasks/card-message-internal-dispatch/handoff.md`.

## Changes

- **`internal/carddispatch`** — capability-bound `Sender` (no caller `from_uid`/
  payload), immutable `Registry` (no `init()` global; unknown/duplicate/disabled/
  misconfigured producers fail closed), fixed-order dispatch pipeline, live
  `DBAuthorizer` (Space/DM/group/thread ACLs, fail-closed on every DB error),
  per-producer in-flight semaphore (`busy` on saturation), bounded Prometheus
  metrics + AST source guards for label/log vocabulary.
- **`pkg/cardtmpl`** — reviewed `ResourceCard` template (octo/v1, markdown-escaped
  leaves, https-only actions, rune-bounded excerpt, optional icon, deep link
  built server-side from `External` config).
- **`modules/botidentity`** — resolver extended to return `creator_uid` / app
  `scope` / `space_id` in one live snapshot.
- **`modules/notify`** — Decision 14 gate: `/v1/internal/notify[/batch]` rejects
  card payloads (`err.server.notify.card_not_allowed`, + zh-CN) before delivery,
  zero transport calls; batch pre-flights the whole request.
- **`tools/lint-card-dispatch`** — AST no-bypass guard (CI): only the reviewed
  external ingresses + `internal/carddispatch` may own type-17 transport;
  allowlist anchored on the import path.
- **Docs** — brief + handoff.

## Testing

```
go build ./... && go vet ./internal/carddispatch/... ./pkg/cardtmpl/...
go test ./internal/carddispatch/... ./pkg/cardtmpl/... ./tools/lint-card-dispatch/...
go run ./tools/lint-card-dispatch ./modules ./internal   # exit 0
```

- [x] Unit tests added/updated (dispatch pipeline, registry, policy matrix,
  fail-closed storage stages, member-exempt + explicit-ban, metrics/label
  guards, template snapshot/validate, guard wrapper/impostor evasions)
- [x] Manually verified (build/vet/test/guard green locally; DB authorizer
  tests run on in-memory sqlite locally, MySQL in CI per the brief)

## COMPREHENSION

1. **What does this change actually do to the load-bearing path?** It creates a
   single trusted origination point for type-17 cards from server-internal code:
   sender trust (live `botidentity`, `iwh_`/human rejected), explicit Space +
   live target ACL, server-authored `plain`/`space_id`, and the ≤512 KiB final
   check are enforced in a fixed order before one transport attempt. It also
   closes the notify payload-passthrough hole at runtime (Decision 14). No
   production producer is enabled, so no new card traffic is emitted yet.

2. **What could break (dependents + failure mode)?** The only live behaviour
   change is `/v1/internal/notify[/batch]` rejecting card-shaped payloads; the
   sole known caller (smart-summary) sends `type=1` text only, so no legitimate
   traffic changes — a mis-sent card now gets `err.server.notify.card_not_allowed`
   instead of an untrusted send. The AST guard fails CI if any new non-test
   package pairs card construction with direct `SendMessage*`; onboard through
   the dispatcher or add a path-anchored allowlist entry with a reason.

3. **How do you know it works?** Table-driven + DB-backed tests cover the policy
   matrix, fail-closed on every DB error/not-found, exact validation order with
   zero transport on any denial, one-call success/error, overload/`busy`, the
   metric-label and no-card-in-logs source guards, and the notify card gate
   (card rejected with zero `SendMessage`, text still delivered). The guard runs
   clean against `./modules ./internal` and fails fixtures for each bypass shape
   including method/field/higher-order wrappers and impostor same-name packages.

## Checklist

- [x] Updated documentation (brief + handoff)

https://claude.ai/code/session_01CNzozZG65vnBy78CgYA5NT

---
_Generated by [Claude Code](https://claude.ai/code/session_01CNzozZG65vnBy78CgYA5NT)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -326,6 +326,27 @@ jobs:
       - name: Reject direct PERSONAL MsgSendReq literals
         run: go run ./tools/lint-personal-msgsendreq ./modules
 
+  card-dispatch-lint:
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
+    # Internal card producers must use the producer-bound carddispatch sender.
+    # This AST guard is a review backstop for recognizable type-17 construction
+    # and transport flows; runtime payload passthrough is guarded at ingress.
+    name: Card Dispatch Lint
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Reject unreviewed internal type-17 transport owners
+        run: make card-dispatch-lint
+
   i18n-extract-check:
     needs: [changes]
     if: |
--- a/.octospec/tasks/card-message-internal-dispatch/brief.md
+++ b/.octospec/tasks/card-message-internal-dispatch/brief.md
@@ -1,12 +1,12 @@
 ---
 type: Task
 title: "Task: card-message-internal-dispatch"
-description: Add the only supported in-process path for business modules to send trusted interactive cards.
+description: Add the only supported in-process path for business modules to send trusted interactive cards, grounded in the summary-forward and docs-share scenarios.
 tags: ["message", "card", "internal", "trust-boundary", "wire-contract", "auth", "space", "isolation", "acl", "rate-limit", "observability", "testing"]
-timestamp: 2026-07-13T16:17:40+08:00
+timestamp: 2026-07-13T17:00:00+08:00
 # --- octospec extension fields ---
 slug: card-message-internal-dispatch
-upstream: self
+upstream: octo-server#571
 source: self
 ---
 
@@ -18,7 +18,10 @@ source: self
 ## Goal
 
 Add `internal/carddispatch` as the only supported path for an in-process
-business module to originate an `InteractiveCard` (`type=17`) message.
+business module to originate an `InteractiveCard` (`type=17`) message, so that
+server-side scenarios — smart-summary results delivered/forwarded into chats as
+cards, docs share-link cards, and future system cards — have one reviewed,
+trusted origination boundary instead of ad-hoc sends.
 
 The dispatcher binds a reviewed producer capability to one configured active
 bot identity, authorizes the exact Space and target using live database state,
@@ -28,16 +31,88 @@ size recheck in a fixed order, makes at most one `SendMessageWithResult`
 transport attempt per invocation, and emits bounded metrics and structured
 logs.
 
-This closes the architectural hole where a future business module could call
-`config.Context.SendMessageWithResult` directly with a hand-built type-17 map
-and accidentally skip sender trust, Space/target ACLs, authoritative `plain`, or
-the final post-mutation 512 KiB check.
+This closes the architectural hole where a business module (or an internal
+S2S caller relaying through one) could reach `config.Context.SendMessage` /
+`SendMessageWithResult` with a hand-built type-17 map and skip sender trust,
+Space/target ACLs, authoritative `plain`, or the final post-mutation 512 KiB
+check. The cross-repo research below shows this hole is not hypothetical:
+`POST /v1/internal/notify` already forwards a caller-supplied payload map
+verbatim as the trusted `notification` User Bot (Decision 14 closes that
+specific gap in this task).
 
-No HTTP endpoint is added. Existing Bot API, legacy Robot API, and Incoming
-Webhook producers keep their current ingress-specific paths in this task; the
-new boundary governs server-internal business producers.
+No new HTTP endpoint is added. Existing Bot API, legacy Robot API, and
+Incoming Webhook producers keep their current ingress-specific paths in this
+task; the new boundary governs server-internal business producers. The only
+behavior change to an existing endpoint is the fail-closed card-payload
+rejection on `/v1/internal/notify` (Decision 14).
 
 ## Background
+
+### Cross-repo scenario research (2026-07-13)
+
+The two motivating scenarios were traced end-to-end across the four repos.
+
+**Scenario A — 智能总结投递/转发到会话 (smart-summary → chat):**
+
+- `octo-smart-summary` is a standalone Go sidecar (separate binaries
+  `summary-api` / `summary-worker`), not an octo-server module. After a summary
+  task terminates, the worker pushes a **plain-text DM** itself:
+  `POST {OCTO_API_URL}/v1/internal/notify` with header `X-Internal-Token`
+  (`SUMMARY_NOTIFY_TOKEN`), body payload hardcoded to `{"type":1,"content":…}`
+  (`internal/notify/notify.go:283-291`, `internal/notify/client.go:83,156`).
+  Per-recipient dedup/retry state lives in its `summary_notification` table.
+- Delivery back into the **origin group/thread** is explicitly a future
+  enhancement (`internal/notify/target.go:34-37`); today every origin falls
+  back to a creator DM. The repo has zero card/type-17 code.
+- On the octo-server side, `modules/notify` owns that ingress:
+  `POST /v1/internal/notify` and `/v1/internal/notify/batch`
+  (`modules/notify/api.go:90-95`), gated by `NOTIFY_INTERNAL_TOKEN`
+  (fail-closed when unset, constant-time compare). It verifies Space
+  membership, then fans out **DM-only** sends via
+  `config.NewPersonalMsgSendReq` from the static system bot
+  `notification` with a bounded 20-goroutine pool (`api.go:243-303`).
+- The `notification` sender is a **full User Bot**: `ensureNotifyBot`
+  provisions the `user` row, `app` row, and a `robot` row with `status=1`
+  (`modules/notify/bot_manager.go:52-87`). `modules/botidentity` therefore
+  resolves it as an active bot and `modules/cardtrust` trusts it as a card
+  sender.
+- In `octo-web`, "转发到聊天" for a summary sends the summary text as plain
+  `MessageText` chunks via the WuKongIM JS SDK
+  (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1018-1043`).
+
+**Scenario B — docs 文档分享链接卡片 (docs share-link card):**
+
+- `octo-docs-backend` (TypeScript/Hocuspocus) has **no outbound IM capability
+  at all** — stated in source (`src/api/routes/accessRequests.ts:13-17`); its
+  octo-server integration is identity-only (token→uid, uid→profile). Share-to-
+  chat is split: docs-backend authorizes recipients (`forward-grant`, invite
+  tokens); the message itself is sent by the web client.
+- In `octo-web`, doc sharing sends a plain `MessageText` whose body is
+  Markdown `**title**\n[url](<url>)` via the JS SDK
+  (`packages/dmworkbase/src/Components/WKBase/index.tsx` `runDocForward`,
+  `ForwardModal/forwardMessageText.ts:55-58`). There is no link-card/unfurl
+  rendering.
+
+**Established platform facts relevant to both:**
+
+- The web client is **receive/render-only** for type-17 by design ("波 1 web
+  不发送 type-17", `packages/dmworkbase/src/Messages/InteractiveCard/`
+  `InteractiveCardContent.ts:22,59`); only re-forwarding an already-received
+  display-only card exists. So every card origination for these scenarios must
+  happen server-side — exactly the boundary this task builds.
+- Inside octo-server, only three modules can emit type-17 today, each with its
+  own ingress trust: `modules/bot_api/send.go`, `modules/robot/api.go`,
+  `modules/incomingwebhook`. All other `SendMessage*` call sites send fixed
+  non-card types from static system identities (`notify`, `botfather`,
+  `app_bot` welcome, group/thread/user system tips).
+- **Live gap:** `modules/notify.deliverNotification` forwards the
+  caller-supplied payload map to transport without any type restriction. A
+  holder of `NOTIFY_INTERNAL_TOKEN` can thus emit a hand-built `type=17` map
+  as the **trusted** `notification` bot, skipping `cardmsg.Enabled`,
+  `Validate`, `Finalize`, authoritative `plain`, and the 512 KiB gates. The
+  planned AST guard cannot see this shape (no card construction in
+  octo-server source — the map arrives over HTTP), so it needs the runtime
+  gate in Decision 14.
 
 ### Existing authoritative pieces
 
@@ -46,7 +121,10 @@ new boundary governs server-internal business producers.
   - `Validate` enforces the profile/version, structure, URL, depth/node and
     complete-payload limits;
   - `Finalize` overwrites untrusted `plain` and rechecks the enriched payload;
-  - `RecheckPayloadSize` checks the final bytes after any later mutation.
+  - `RecheckPayloadSize` checks the final bytes after any later mutation;
+  - `IsCardPayload` / `IsCardRawPayload` are the single card-shape detectors;
+  - profiles: `octo/v1` is display-only, `octo/v2` is the only interactive
+    profile (`profiles.go:19-22`).
 - `modules/botidentity` resolves live bot authority:
   `robot.status=1` or `app_bot.status=1`; `user.robot=1` is presentation only;
   ambiguous or failed lookups fail closed.
@@ -55,6 +133,10 @@ new boundary governs server-internal business producers.
   - `modules/bot_api/send.go`;
   - `modules/robot/api.go`;
   - `modules/incomingwebhook/api.go` + `card.go`.
+- `modules/notify` owns the internal S2S notification ingress
+  (`/v1/internal/notify*`, `X-Internal-Token`) and the static `notification`
+  User Bot (see research above). It is an existing trust surface of this
+  boundary, not a card producer.
 - `pkg/cardmsg/producer_completeness_test.go` already proves every producer that
   expands `mention.ais` performs a final size recheck. It does not prevent a new
   internal producer from bypassing all of the other gates.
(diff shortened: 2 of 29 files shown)


OctoBoooot (dismissed):
## Review: internal card dispatch boundary — sole trusted in-process type-17 origination (dormant foundation) (#577)

**Verdict:** Approve, conditional on Test going green (Build/Vet/Lint + the new Card Dispatch Lint all pass). Risk tier: **high** (trust boundary — sole trusted origination of type-17 cards); reviewed at that rigor, scoped to the trust-critical surface (the live gate + dormancy + authorizer fail-closed) rather than line-reviewing all 29 files / tooling / templates — the right depth for a dormant foundation whose only live behavior change is the notify gate. needs-human-review applies. check-sprint red is governance.

### The live security fix (Decision 14) — byte-verified sound
This is the only behaviorally-live change, and it closes a real hole (internal-token holder emitting a trusted type-17 map that bypasses every card gate):
- **Single-notify:** `deliverNotification` rejects `cardmsg.IsCardPayload(req.Payload)` first, before any delivery (notify/api.go:204) → a card payload never reaches transport.
- **Batch:** a **preflight loop** checks every notification for card-shape before delivering any (api.go:170-174), with an explicit "zero-transport guarantee when a later entry is a card" — rejects the whole batch atomically rather than sending earlier text items then failing. Correct fail-closed ordering (a naive per-item check would leak partial delivery).
- **`IsCardPayload` predicate is tight and complete for the render contract:** detects `payload["type"]==17` across the realistic decode reps (`float64` from gin BindJSON, `int`, `json.Number`). The delivery path passes the payload to WuKongIM with `type` intact, and a card renders iff numeric type==17 — so the gate catches exactly what would render as a card. A `string "17"` slips the gate but also isn't honored as a card downstream (not a bypass). Neither under- nor over-inclusive relative to the wire contract.

### Dormancy — verified inert
- `main.go:354` `NewRegistry(deps, nil)` — no production producer registered → the dispatcher is behaviorally inert, matching the "P1 dormant foundation" claim. The only live effect is the Decision-14 gate above.

### Dispatcher built fail-closed (inert now, but it's the future "sole trusted origination")
- `DBAuthorizer.Authorize` returns `denyTarget(...)` on EVERY failure axis: nil deps, ctx cancellation, any DB query error (`err != nil → deny`), inactive space (`status=1` gate), non-member recipient, app-bot scope mismatch, unstarted DM, unauthorized sender. Every DB error denies rather than proceeds — no fail-open. Consistent with the disabled-entity `status=1` gating enforced across the card/space stack.
- Dispatch pipeline runs the spec'd fixed order: `cardmsg.Validate` → authoritative enrich → `Finalize` → `RecheckPayloadSize` → at most one `SendMessageWithResult` (dispatch.go:105-131). Capability-bound Sender (no caller from_uid/payload), immutable registry (unknown/dup/disabled/misconfigured producers fail closed), per-producer in-flight semaphore.

### Praise
- The right shape for a trust-boundary foundation: ship the *gate* live (closing the internal-token→type-17 hole) while keeping the *dispatcher* dormant (nil producers) but fail-closed by construction, so activation is a separate reviewed step. The batch preflight's zero-transport guarantee and the AST/lint source-guards for label/log vocabulary (dedicated Card Dispatch Lint CI, passing) show the trust surface was thought through end to end. The `IsCardPayload` predicate is scoped exactly to the numeric-type render contract — not looser, not tighter.

### Minor (non-blocking)
- Given dormancy, the dispatcher's live-activation authz will warrant its own focused re-review when a real producer is registered (the P2 pilot PR) — that's when the authorizer moves from inert-but-correct to load-bearing. Flagging so the next stage gets trust-tier scrutiny, not treated as "already approved."

### CI
- Build/Vet/Lint/Card Dispatch Lint/i18n green; Test in-progress. Approval conditional on Test green (a trust-boundary's tests are load-bearing). check-sprint governance-red. needs-human-review applied (high tier). mergeStateStatus stays BLOCKED until CI + approvals + human sign-off.

Jerry-Xin (dismissed):
Trust boundary is genuinely enforced fail-closed, the dormant claim is byte-verified inert, and the one live behavior change (notify card-rejection gate) is fail-closed with a zero-transport guarantee. Deny paths are tested. No blockers.

## Trust-boundary integrity — enforced, not just convention
Two independent layers:
- **Runtime (live today):** `modules/notify/api.go:204-205` — `deliverNotification` rejects any card payload first thing: `if req != nil && cardmsg.IsCardPayload(req.Payload) { return nil, errNotifyCardNotAllowed }`, before any validation/delivery. Batch path pre-flights the WHOLE request (`:171-177` loops `IsCardPayload` over all notifications) before the delivery loop (`:180`), so a card anywhere aborts the entire batch → **zero-transport, no partial send**. This closes the internal-token-holder passthrough hole. `IsCardPayload` matches type-17 across float64/int/json.Number (the correct set for BindJSON→map decoding).
- **Compile-time backstop:** `tools/lint-card-dispatch/main.go` propagates constructsCard+directTransport across the call graph and flags any function outside the 4-entry allowlist that pairs them; the allowlist is **anchored on import path, not bare package name** — explicitly defeating the impostor-package bypass. Wired into CI (`.github/workflows/ci.yml` `card-dispatch-lint`) and **passing**.

## Fail-closed everywhere
`internal/carddispatch/authorizer_db.go`: every error/zero-count branch returns `denyTarget(...)` — nil deps, ctx cancel, space query error, inactive space, every membership/friend/group/thread query error. `dispatch.go` sets a deny category at each stage and returns before transport. `modules/botidentity/resolver.go:99` preserves lookup errors; ambiguous identity → error. No fail-open branch exists.

## Dormant claim — verified inert
`main.go:354`: `registry := carddispatch.NewRegistry(deps, nil)` — **nil specs** ⇒ empty senders map ⇒ every `Sender(id)` returns `producer_disabled`. `grep -c "ProducerSpec{"` in main.go = **0**: zero producers wired. The only reachable-today path is the notify card-*rejection* gate (deny-only, never originates). Attack surface is not live.

## resolver.go — no trust regression
Purely additive: same authoritative source (robot/app_bot WHERE status=1), same single-statement snapshot, same ambiguous→fail-closed contract. New fields (creator_uid/scope/space_id) all read server-side from authoritative tables — no client-influenced trust field.

## cardtmpl/resource.go — no traversal / no SSTI
It's a document builder from struct fields, not a filesystem loader. Deep link built server-side via url.Parse + https-only + PathEscape/QueryEscape; icon URL forced absolute-https; leaves escapeMarkdown-escaped; no text/html template or exec anywhere.

## Tests — deny paths asserted
`TestSendFailClosedStagesNeverReachTransport` (each denial category, zero transport), `TestRegistryRejectsEveryInvalidConfigurationClass` (12 classes), `TestDBAuthorizerFailsClosedOnDatabaseError` + `...AtEveryStorageStage`, 30+-row `TestDBAuthorizerPolicyMatrix` (stranger-DM/wrong-space/blacklist/external/archived-thread denials), notify integration asserts `card_not_allowed` + zero messageCount for single AND batch.

## 🟡 Non-blocking
- `IsCardPayload` deliberately does not treat a string `"17"` type as a card — fine for JSON-decoded notify ingress (numbers → float64), but worth a one-line comment if any future caller hands string-typed payloads.
- The lint backstop is a recognizable-shape AST check; reflection/codegen-built transport wouldn't be caught — but runtime ingress gating is the real defense, so this is acceptable defense-in-depth, not the sole guard.

## ✅ Highlights
- Import-path-anchored allowlist defeating the impostor-package bypass — genuinely thoughtful.
- Immutable registry, no init() global, fail-closed on unknown/duplicate/disabled/misconfigured producers.
- Fixed-order pipeline (Validate → server-authoritative space_id enrich → Finalize → size recheck → single send); server owns the envelope, caller supplies neither from_uid nor wire type.
- Deny-path + zero-transport test coverage is the strongest part.

Note: sole red check is `check-sprint` (project-board process gate, not code/CI); Test/Vet/Lint were still IN_PROGRESS at review time — none failed on code.

✅ Approving.

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD 1781aff1

**Formal Verdict**: `COMMENT`



## Lens Summary

| Lens | Verdict | Headline |
|---|---|---|
| code | `needs-discussion` | Solid fail-closed foundation; AST guard misses `SendMessageBatch` transport. |
| security | `cleared-with-risk` | New type-17 trust boundary is fail-closed and non-spoofable; only low, non-exploitable residual risks. |
| qa | `pass-with-risk` | All ACs covered by strong assertions, CI green (incl. integration `Test`), tests pass; only low-severity test-quality gaps. |

## Critical Findings (P0-P1)

无 Critical / High priority findings — all three lenses converged on lower-severity items. No exploitable security condition, no failing AC, no failing CI, no `Must Fix`.

## Should-Fix (P2)

- `[P2-1]` `tools/lint-card-dispatch/main.go:182,464` — AST no-bypass guard recognizes only `SendMessage`/`SendMessageWithResult` as transport, but `config.Context` also exposes `SendMessageBatch` (raw `Payload []byte`) and `SendCMD`; a future producer that builds a type-17 card and calls `ctx.SendMessageBatch(...)` would pair card-construction with real transport yet stay green, defeating the guard's "only reviewed ingresses own type-17 transport" invariant. Not runtime-exploitable today (no producer registered; the notify runtime gate is authoritative). — flagged by code/code-1 — Fix: add `SendMessageBatch` (and evaluate `SendCMD`) to `calledName`/`isTransportReference` with matching evasion fixtures, OR document in the guard header why batch/CMD are intentionally out of scope.
- `[P2-2]` `tools/lint-card-dispatch/main.go:232` — Guard call-graph propagation is intra-package only: a card constructor in package A that calls an exported transport-owning func in package B is not linked (neither side has both flags), so a cross-package bypass evades the guard. CI backstop only; runtime gates in `dispatch.go`/`notify/api.go` remain authoritative. — flagged by security/sec-1 + code/code-2 (cross-lens consensus) — Fix: document the intra-package limitation in the guard header, or move to type-checked cross-package analysis (`go/packages`) if the guard is to be relied on.

## Suggestions (P3)

- `[P3-1]` `internal/carddispatch/dispatch.go:113-120` — Explicit `RecheckPayloadSize` is redundant: `cardmsg.Finalize` already calls it (`plain.go:31`) with no payload mutation in between. — code/code-3 — Fix: drop it, or add a one-line comment marking it a deliberate spec-literal "final recheck" defense.
- `[P3-2]` `internal/carddispatch/observability_source_test.go:51` — The no-card-in-logs source guard parses only `dispatch.go`; `registry.go` also logs via `deps.Logger` and is uncovered (currently safe — both fields bounded). — code/code-4 — Fix: extend the source guard to `registry.go`, or note the intentional scope.
- `[P3-3]` `main_carddispatch_test.go:15-32` — Install-before-`module.Setup` / empty-registry invariant is asserted only by string-matching `main.go` source text; brittle to a rename/refactor and gives no behavioral signal. — qa/qa-1 — Fix: acceptable as a source guard, but consider a behavioral ordering assertion at a seam if feasible.
- `[P3-4]` `internal/carddispatch/dispatch.go:86→131` — TOCTOU (low): space/membership state read by `Authorize` can change before `SendMessageWithResult`. Inherent to check-then-act; window small, single attempt. — security/sec-2 — Fix: accept as residual risk; no change required.
- `[P3-5]` `pkg/cardtmpl/resource.go:67` — `IconURL` enforces https + non-empty host but has no host allowlist. Rendered client-side (no server-side fetch) and producer-supplied (server-internal) → not server SSRF. — security/sec-3 — Fix: add a host allowlist only if/when producers become less trusted.

## Questions for Author

- `[Q-1]` `modules/notify/api.go:169-176` — Batch preflight rejects the **entire** batch (zero delivered) if any single entry is a card. Intentional per the zero-transport guarantee and documented in the PR body; no regression for the current text-only (`type=1`) caller, but a future mixed text+card batch caller would see all-or-nothing. Confirm this is the desired contract. — qa/qa-3
- `[Q-2]` `internal/carddispatch/dispatch_test.go` — The full `producerSender.Send` pipeline is only exercised against a `fakeAuthorizer`; the real `DBAuthorizer` is strongly tested in isolation but never wired into `Send` end-to-end. Both halves are well-covered — flagging the seam, not a defect. — qa/qa-2
- `[Q-3]` `pkg/cardmsg/cardmsg.go:114` — `IsCardPayload` intentionally ignores string `"17"` (parity with `IsRichTextPayload`); verified non-exploitable because content type is numeric system-wide. Noting only so a future non-numeric `type` reader would need the same gate. — security/sec-4

## Verdict Rationale

`COMMENT`: no P0/P1, no exploitable security condition (security = `cleared-with-risk`, no embargo), all ACs covered by strong assertions with CI green including the integration `Test` job (qa = `pass-with-risk`), and 0 `Must Fix`. Two P2 should-fixes remain — both completeness/soundness gaps in the **CI lint backstop** (`lint-card-dispatch`), not live vulnerabilities: no production producer is registered (`NewRegistry(deps, nil)`), and the authoritative runtime defenses (fixed-order fail-closed dispatch pipeline + the Decision-14 notify gate) are correct and well-tested. Cross-lens consensus: code + security independently flagged the AST-guard soundness limitation (P2-2). The remaining items are nits and author-confirmation questions. Recommend the author either close the two guard gaps or record them as documented residual risk in the brief before enabling the P2 pilot.

## Round Context

ROUND=1 — First-round review, no prior round context.

|---|---|---|
| `.github/workflows/ci.yml` | ✓ | (none) |
| `.octospec/tasks/card-message-internal-dispatch/brief.md` | ⚠ sampled (doc) | (none) |
| `.octospec/tasks/card-message-internal-dispatch/context.yaml` | ✗ skipped (doc) | (none) |
| `.octospec/tasks/card-message-internal-dispatch/handoff.md` | ⚠ sampled (doc) | (none) |
| `Makefile` | ✓ | (none) |
| `internal/carddispatch/authorizer_db.go` | ✓ | (none — fail-closed verified) |
| `internal/carddispatch/authorizer_db_test.go` | ✓ | (none) |
| `internal/carddispatch/context.go` | ✓ | (none) |
| `internal/carddispatch/dispatch.go` | ✓ | P3-1, P3-4 |
| `internal/carddispatch/dispatch_test.go` | ✓ | Q-2 |
| `internal/carddispatch/errors.go` | ✓ | (none) |
| `internal/carddispatch/metrics.go` | ✓ | (none) |
| `internal/carddispatch/observability_source_test.go` | ✓ | P3-2 |
| `internal/carddispatch/registry.go` | ✓ | (none) |
| `internal/carddispatch/types.go` | ✓ | (none) |
| `main.go` | ✓ | (none) |
| `main_carddispatch_test.go` | ✓ | P3-3 |
| `modules/botidentity/resolver.go` | ✓ | (none) |
| `modules/botidentity/resolver_integration_test.go` | ⚠ sampled | (none) |
| `modules/botidentity/resolver_test.go` | ✓ | (none) |
| `modules/notify/api.go` | ✓ | Q-1 |
| `modules/notify/notify_integration_test.go` | ✓ | (none) |
| `pkg/cardtmpl/resource.go` | ✓ | P3-5 |
| `pkg/cardtmpl/resource_test.go` | ✓ | (none) |
| `pkg/errcode/notify.go` | ✓ | (none) |
| `pkg/i18n/locales/active.zh-CN.toml` | ✓ | (none) |
| `tools/i18nmarkers/server/active.en-US.toml` | ✗ skipped (generated) | (none) |
| `tools/lint-card-dispatch/main.go` | ✓ | P2-1, P2-2 |
| `tools/lint-card-dispatch/main_test.go` | ✓ | (none) |

## Notes

- Confidence filter (aggregate): code lens 0 dropped / 1 downgraded; security lens 1 dropped / 0 downgraded; qa lens 2 dropped / 0 downgraded. Cross-lens consensus boost applied to P2-2 (code + security).
- Contrarian pass ran on all three lenses: 0 net-new blocking findings; verified by explicit traces (default-deny enum arms, `denyTarget` on every DB error, numeric-`ContentType` parity), not generic "checked all paths."
- Tooling: `tool-scan.sh` wrapper unavailable at the default path (exit 2) on all lenses; `go vet` run directly on the new packages — clean. `scan-deps.sh` — no dependency changes (`go.mod` untouched). Credential scan on the diff — clean, no leak.
- Redaction: GH copy passed through `redact-report.sh`; multica copy is raw. No `SENSITIVE`/embargo content — nothing exploitable to embargo.

---

lml2468 (commented):
COMMENT — Round 1

Fail-closed type-17 dispatch foundation is correct, non-spoofable, and well-tested; CI green (incl. integration `Test`). No P0/P1, no exploitable security condition, 0 Must Fix. Two P2s remain — both soundness gaps in the CI lint backstop (`lint-card-dispatch` misses `SendMessageBatch` transport + is intra-package only), not live vulnerabilities since no producer is registered and the runtime gates are authoritative.

Full aggregate on this PR (issue comment above) and on the multica tracker.

yujiawei (dismissed):
# Code Review — PR #577 (octo-server)

Independent review of the `internal/carddispatch` boundary and the Decision-14 notify gate. I read the full diff against `cc61c37d`, the linked spec (`.octospec/tasks/card-message-internal-dispatch/brief.md`), the relevant existing ingresses (`bot_api`, `notify`), and the octo-lib transport, and I built + ran the affected tests and the AST guard locally.

**Result: build green; `internal/carddispatch`, `pkg/cardtmpl`, `tools/lint-card-dispatch` tests pass; `go run ./tools/lint-card-dispatch ./modules ./internal` exits 0; no production producer is registered (`NewRegistry(deps, nil)`), and `pkg/cardtmpl` is imported by no production code — the foundation is genuinely dormant.**

## 1. Spec compliance

**Spec: ✅**

- **No missing work.** Every load-bearing decision is implemented: capability-bound `Sender` with an immutable registry that fails closed on unknown/duplicate/disabled/misconfigured producers (D1); live `botidentity` resolution with `iwh_`/human/ambiguous rejection in one SQL snapshot (D2); explicit-Space + live target ACL (D3/D4) with fail-closed DB error handling; dispatcher-owned envelope (D5); immutable input snapshot (D6); the fixed validation order `structural → feature gate → producer policy → slot → snapshot → identity → authorization → Validate → Space enrich → Finalize → RecheckPayloadSize → serialize → one transport call` (D7) matches the code path exactly; single transport attempt, no retry, no goroutine (D8); per-producer in-flight semaphore returning `busy` (D9); bounded metrics/log vocabulary with an AST source guard (D10); single composition point, no producer imports (D11); the AST no-bypass guard anchored on import path with per-entry reasons, wired into CI (D12); request-time authorization with no lock across transport (D13); and the runtime Decision-14 gate rejecting card payloads on `/v1/internal/notify[/batch]` before delivery with zero transport calls (D14).
- **No over-build.** No production producer is enabled; `notify` is not wired as a card producer; the `ResourceCard` template ships reviewed-but-inert with snapshot + `Validate`-under-`octo/v1` guard tests, consistent with the "dormant foundation may ship first" scope. The registered error code has en-US + zh-CN entries.
- **No deviation.** The `space_id` handling (server-authored, injected into the envelope) matches the platform's `NewPersonalMsgSendReq` convention; the member-exempt group mode honors an explicit blacklist row while allowing no-row posting, exactly as specified.

Decision-14 is implemented with defense-in-depth: `sendNotifyBatch` pre-flights the whole batch before any delivery (preserving the zero-transport guarantee when a later item is a card), and `deliverNotification` re-checks per item.

## 2. Code quality

**Quality: Approved**

No P0/P1 issues found. The authorization logic is careful and uniformly fail-closed, all SQL is parameterized, and every terminal branch records exactly one bounded metric via defer.

Advisory (P2 — none blocking; several are relevant to the *pilot enablement* PR, not to this dormant foundation):

- **[P2] `dispatch.go` — semaphore held across an un-cancellable transport call.** The octo-lib transport (`SendMessageWithResult` → `network.Post` → `rest.DefaultClient` with a zero-value `&http.Client{}`) has **no timeout**. A hung WuKongIM connection would hold a producer slot indefinitely; with `MaxInFlight` slots exhausted, subsequent sends fail closed with `busy` until restart. This is a pre-existing platform-wide property of every `SendMessage*` caller (including today's `notify` pool), and the spec explicitly scopes cancellable transport out and forbids faking cancellation via a leaked goroutine — the code correctly stays synchronous and checks `ctx` before dispatch. **Recommendation for a human before the pilot is enabled:** ensure the WuKongIM client carries a request timeout, otherwise the per-producer budget can be pinned by a slow upstream.

- **[P2] `dispatch.go` — `json.Unmarshal` of the card document uses default number decoding.** Bare JSON numbers in the caller's Adaptive Card document decode to `float64`, so integers above 2^53 would lose precision on re-serialization. In practice the pilot cards are display-only (`octo/v1`: strings, enums, URLs) and this mirrors the existing `bot_api` card ingress (also `map[string]interface{}` via `BindJSON`), so it is consistent, not a regression. If future interactive (`octo/v2`) producers ever place large integer IDs in card data, consider `json.Decoder` with `UseNumber()`.

- **[P2] `dispatch.go` — the input `document` copy.** The defensive copy is spec-mandated (D6: an immutable snapshot feeds validation, plain derivation, size recheck, and serialization). Note that it protects against a caller retaining/mutating the slice across calls; it does not make *concurrent* mutation during a single `Send` safe (that would be a caller-side data race). Behavior is correct; the code comment slightly overstates the concurrency guarantee.

## 3. Security-sensitive notes for manual verification

- The **DM system-notification mode** authorizes a send to any active member of the verified Space without sender membership or a friend relationship. This is the intended `notify` semantics and is per-producer registry config (never caller-selectable), but it is a broad grant — confirm the pilot's `summary` bot producer is the only consumer at enablement.
- The **member-exempt group mode** lets the bot post to a group with no membership row. It correctly re-verifies group lifecycle + exact Space equality at send time and honors an explicit blacklist; the "recorded triggering member action" precondition lives in the (not-yet-enabled) owning module, so verify it exists when the pilot lands.
- The **AST no-bypass guard** is a review backstop, not a runtime boundary (as the code states); the payload-passthrough class it cannot see is closed at runtime by Decision-14. I attempted the documented evasions (method/alias/higher-order/split-across-files/impostor package) against the guard's own fixtures and they are covered.

## 4. Suggestions

- Before flipping the `summary-notify` pilot on, land the transport-timeout consideration above and the octo-web `/s/:taskId` route-contract test the spec calls for.

OctoBoooot (approved):
## Delta re-review: #577 @ 8e270644 (prior @ 1781aff1) — pure hardening, APPROVE re-affirmed

**Verdict:** Approve (re-affirming my 1781aff1 APPROVE), conditional on Test/Build/Vet green. The delta is pure hardening that addresses both of the prior-round 🟡s; zero change to the verified fail-closed trust logic. needs-human-review retained (high tier). check-sprint is now GREEN.

### Delta (ahead 1, behind 0) — all improvements, trust logic untouched
- **Trust-critical files blob-UNCHANGED** from my verified head: `authorizer_db.go` (fail-closed authz), `registry.go` (dormant/immutable), `modules/notify/api.go` (Decision-14 live gate), `pkg/cardmsg/cardmsg.go` (IsCardPayload). My byte-verification of all four carries unchanged.
- **`lint-card-dispatch` strengthened (closes a real blind spot):** the source-guard now also catches `SendMessageBatch` (via new `isTransportName`), not just SendMessage/SendMessageWithResult — `SendMessageBatch` carries an arbitrary payload, so a type-17 card through it would have evaded the old 2-method guard. New test `direct SendMessageBatch` (`{"type":17}` → caught) regression-nets it. This addresses Jerry-Xin's implicit concern that the guard's method list must be complete.
- **Jerry-Xin's 🟡#2 (reflection/codegen depth) now documented in source:** the new `isTransportName` doc states the guard "matches recognizable call/data-flow shapes only — a transport call reached across packages, or built via reflection/codegen, is out of its scope; the runtime ingress gate (Decision 14) is the authoritative defense, this is defense-in-depth." Exactly the depth-boundary caveat he raised, now codified.
- **`dispatch.go`:** comment-only — expands the RawMessage-snapshot semantics + a caller-side concurrent-mutation caveat. No functional change to the Validate→Finalize→size-recheck→single-send pipeline.

### Carried (unchanged, non-blocking)
- yujiawei's transport-timeout suggestion: correctly deferred (dormant + per-producer semaphore bounds concurrency; a P2-activation concern).
- Jerry-Xin's IsCardPayload string-"17" comment 🟡: the predicate itself is unchanged and correct (numeric-type render contract); an explicit comment would still be a nice touch but non-blocking.

### Verdict
- APPROVE holds — trust logic byte-identical to my verified head; the delta strengthens the lint guard (SendMessageBatch) and documents the reflection/codegen depth limit. check-sprint cleared to green; conditional on Build/Test/Vet green. needs-human-review remains (high-tier trust boundary requires human sign-off).
- Forward note (unchanged): when the P2 pilot registers a live producer, the dispatcher's now-inert authz becomes load-bearing and warrants its own trust-tier re-review — "dormant + approved" ≠ "activated + approved."

Jerry-Xin (approved):
Re-affirming APPROVE at new head 8e270644. This is a single ahead-commit (ahead:1, behind:0 — not a rebase); byte-diffed the delta from my last-approved head 1781aff1. It's a strict improvement: the boundary linter now also catches SendMessageBatch, plus a comment-accuracy fix. Live gate, dispatcher authz, and dormancy are unchanged.

## Delta since my last approval (1781aff1 → 8e270644, clean ahead-1)
Three functional files (+ handoff.md doc), byte-diffed:

**1. `tools/lint-card-dispatch/main.go` (+18-3) — tightens the boundary linter (improvement).**
Extracts `isTransportName()` and adds **`SendMessageBatch`** to the recognized transport methods (previously only `SendMessage`/`SendMessageWithResult`). This closes a real gap: `SendMessageBatch` also carries an arbitrary `[]byte` payload, so a type-17 card smuggled through it would previously have evaded the static guard — now it's flagged like the others. The added NOTE comment honestly restates the intra-package / reflection-codegen blind spot and reaffirms the Decision-14 runtime ingress gate as the authoritative defense (matches my earlier 🟡#2). Strictly strengthens the compile-time backstop.

**2. `tools/lint-card-dispatch/main_test.go` (+6) — coverage for the new guard.**
Adds a `direct SendMessageBatch` case asserting the linter now catches the batch bypass (`ctx.SendMessageBatch(payload)` with `{"type":17}`).

**3. `internal/carddispatch/dispatch.go` (+6-2) — comment accuracy only, no code change.**
The byte-snapshot comment is rewritten to be more precise: the private copy protects against a caller *later* mutating its RawMessage, but a caller mutating the *same slice concurrently* mid-Send is a caller-side data race the copy can't prevent (callers must not share a Card across goroutines mid-Send). This is a more honest doc of the existing behavior — the copy logic itself is unchanged.

## What is unchanged (my prior verification carries verbatim)
- **Live runtime gate (Decision-14, `notify/api.go`)** — untouched. Card-payload rejection fail-closed + atomic batch preflight (zero-transport) still exactly as verified.
- **Dispatcher authorizer (`authorizer_db.go`)** — untouched. Fail-closed on every axis (DB error/inactive space/non-member/scope mismatch → deny).
- **Dormancy (`main.go:354` `NewRegistry(deps, nil)`)** — untouched, still zero producers, behaviorally inert.
- **resolver.go / cardtmpl** — untouched.

So the change is a net hardening (linter now covers SendMessageBatch, with a test) plus a comment-accuracy fix; nothing weakened, no new attack surface. My APPROVE holds and is reinforced.

Prior non-blocking notes: 🟡 (document IsCardPayload's numeric-only intent) still stands as a nice-to-have; 🟡 (lint blind to reflection/codegen transport) is now explicitly acknowledged in the new NOTE comment — good.

Forward reminder (unchanged): this is a dormant foundation — when a real producer is registered (P2 activation), the live dispatcher authz warrants its own trust-tier re-review; "dormant + approved" ≠ "activated + approved."

✅ Approving (delta is a strict improvement over prior approval; trust-critical surface unchanged).

yujiawei (approved):
# Code Review — PR #577 (octo-server)

Reviewed at head SHA `8e2706445667ff7f4b5e6044984803f0ef3cb3be` against `origin/main` (merge-base `cc61c37d`). Independently re-verified the diff, the linked brief's locked decisions, and ran the acceptance gates locally.

**Verdict: APPROVE.** This is a well-constructed, genuinely dormant trust-boundary foundation. No blocking (P0/P1) issues. The findings below are P2/advisory — none block this dormant PR, but two are must-settle preconditions for the later producer-enablement PRs.

---

## 1. Spec compliance

**Spec: ✅**

The diff faithfully implements the brief (`.octospec/tasks/card-message-internal-dispatch/brief.md`) as a dormant foundation:

- **No under-build**: capability-bound `Sender` + immutable registry (fail-closed on unknown/duplicate/disabled/misconfigured), live `botidentity` resolution, DB authorizer covering DM/group/thread ACLs, the fixed validation order, the Decision-14 notify gate (single + batch) with a registered errcode + zh-CN translation, the AST no-bypass guard wired into CI, the `ResourceCard` template + deep-link, bounded metrics with a source guard, and the full test matrix are all present.
- **No over-build**: `main.go` installs `NewRegistry(deps, nil)` — zero production producers. No new HTTP endpoint is added; the only runtime behavior change is D14's card-payload rejection on `/v1/internal/notify[/batch]`. `pkg/cardtmpl` ships unwired (no production caller). `main_carddispatch_test.go` pins both the install-before-`module.Setup` ordering and the empty-registry invariant.
- **No deviation**: dispatch order matches Decision 7 exactly — `Validate → set authoritative space_id → Finalize → RecheckPayloadSize → marshal → single transport call`, with `ctx.Err()` checks before DB work and again immediately before transport, and no mutation after the final recheck.

Verified locally (all green): `go build`, `go vet`, `go test ./internal/carddispatch/...` (incl. the DB-backed `TestDBAuthorizerPolicyMatrix` — 35 subcases — and the fail-closed/member-exempt/registry/observability tests), `go test ./pkg/cardtmpl/...`, `go test ./tools/lint-card-dispatch/...`, `go run ./tools/lint-card-dispatch ./modules ./internal` (exit 0), `make i18n-extract-check`, `make i18n-lint`.

## 2. Code quality

**Quality: Approved**

Notably strong: the fixed-order pipeline with the caller-bytes snapshot (`document := append([]byte(nil), card.Document...)`) closing the concurrent-mutation race; the semaphore acquire/release symmetry via `defer` (slot and in-flight gauge both released on every terminal branch, panic-safe); the authorizer failing closed on *every* DB error (each `count`/`SelectBySql` error is wrapped in `denyTarget`); the resolver reading only `robot`/`app_bot` (never `user.robot`) in one statement snapshot and failing closed on ambiguity; and the D14 batch pre-flight that scans the whole batch for card payloads **before** delivering any earlier text item, preserving the zero-transport guarantee.

### P2 — Group member-exempt mode does not verify the sender's space membership (precondition for group enablement)

`internal/carddispatch/authorizer_db.go:115-163` — `authorizeGroup` verifies the group row's `status` and exact `space_id == target.SpaceID`, and under `GroupPolicyMemberExempt` only checks for an explicit blacklist row before returning `nil`. It never consults `policy.SpacePolicy` and never checks that the sender bot is an active `space_member` of the target space. DM authorization (`authorizeDM`) *does* check the sender's `space_member` row under the membership policy, so there is an asymmetry.

This is **not exploitable in this PR**: no producer is registered (dormant), and the DM-only pilot never exercises the group path. The group row's space-equality check bounds any send to the verified-active target space, and member-exempt "post to any group in the space without joining" is the intended semantic (the brief's cited Slack `chat:write.public` analog). So I am not treating it as a blocker.

However, the brief defers group/thread widening to "a separate reviewed change" that "must include a per-channel rate rule and the cluster-cap decision." That review should explicitly decide whether member-exempt group posting also requires the sender bot to be an active member of the target space — otherwise a global notification-style bot can post into any normal group across every space it targets, with only group-in-space + lifecycle + a recorded-member-action (owning-module responsibility) as the tie. Please add a test asserting the chosen behavior when that PR lands.

### P2 — Integer precision: `json.Unmarshal` into `map[string]interface{}` coerces JSON numbers to float64

`internal/carddispatch/dispatch.go:95-96` decodes the card document with plain `json.Unmarshal`, so every JSON integer becomes `float64`. `Validate`, `Finalize`, `RecheckPayloadSize`, and the final `json.Marshal` all operate on that same map, so the size gate stays self-consistent — but any integer > 2^53 in card data (e.g. a Snowflake/issue ID in a future `octo/v2` action `data` field) would be silently rounded on the round-trip.

This is **harmless for the display-only pilot** (the `ResourceCard` template carries only strings and https URLs — no large integers), so it is not a blocker here. But `pkg/cardmsg/interactive.go` (`decodeEnvelope`) already established `json.Decoder` + `UseNumber()` as the repo discipline for exactly this reason (PR#548). I recommend the dispatcher adopt the same (`dec := json.NewDecoder(bytes.NewReader(document)); dec.UseNumber()`) before any `octo/v2` producer is onboarded, to keep the boundary faithful to the bytes it validates.

### P2 — AST guard scan roots miss `pkg/` and `cmd/`

`tools/lint-card-dispatch/main.go:76-78` defaults to `./modules ./internal`, and CI (`make card-dispatch-lint`) runs it with exactly those roots. A new in-process type-17 transport owner added under `pkg/`, `cmd/`, or the repo root would not be scanned. The tool also (by its own documented design) only catches intra-package construct+transport shapes — a split across package boundaries (package A builds a `[]byte` with `"type":17`, package B is a generic byte-payload sender) is out of scope. The guard's comments are honest that it is a review backstop and that Decision 14 is the authoritative runtime defense, so this is defense-in-depth hardening, not a correctness bug. Suggest widening the CI roots (e.g. scan the module root, excluding `vendor`/generated dirs) so the backstop covers the whole tree.

### Nit — DB/identity calls do not propagate `context.Context`

The authorizer's `SelectBySql(...).LoadOne(...)` and `Resolver.Resolve(uid)` do not thread `ctx`, so a stalled DB/identity backend cannot be canceled mid-query. This matches the repo-wide convention (the codebase uses non-context `LoadOne` at ~580 sites and the brief documents the transport as non-cancellable), and the dispatcher already checks `ctx.Err()` before DB work and before transport, so new calls fail fast via the `busy` semaphore. Plumbing context into the data layer is a broader initiative and reasonably out of scope for this PR; noting it for completeness.

---

## Security-sensitive verification (for the human reviewer)

- **Decision 14 gate**: `cardmsg.IsCardPayload` correctly recognizes `type=17` as `float64` (the type produced by `BindJSON` into `NotifyReq.Payload`), `int`, and `json.Number` — so a card cannot slip past the gate via number-type coercion. Both `/notify` and `/notify/batch` reject before any `SendMessage`, and integration tests assert zero-transport on rejection while `type=1` text still delivers.
- **Fail-closed authorization**: every DB error, not-found, wrong-space, disabled/disbanded group, invalid/archived/deleted thread, and blacklisted-bot case returns `denyTarget` before finalization/transport; the DB-backed matrix test covers each branch.
- **Envelope ownership**: the dispatcher sets `type`/`card_version`/`profile`/`space_id` itself; caller-supplied `plain`/`space_id`/OBO/sender/stream/subscriber fields cannot reach the wire.
- **Identity trust**: `iwh_` synthetic identities and human UIDs are rejected at both registry-config time and send time; ambiguous cross-table identities fail closed.

## Coverage note (what I could and could not verify)

Verified from source + local runs: the full `internal/carddispatch` package, the D14 notify gate + errcode + zh-CN, `botidentity` resolver, `pkg/cardtmpl`, the AST guard + its self-tests, `cardmsg.Validate/Finalize/IsCardPayload`, CI wiring, and the dormant-registry bootstrap. Not run locally: the full `go test ./...` under MySQL/Redis/WuKongIM and `golangci-lint run ./...` (require full infra) — the targeted DB-backed authorizer tests did run and pass. Not exercised (no code path in this PR): any live producer end-to-end, since the foundation is dormant by design.

PR #579, opened 2026-07-14
feat(notify): summary-notify card pilot — first internal card producer

an9xyz:
## Summary

Enables the **first production `internal/carddispatch` producer** (`summary-notify`),
upgrading the summary completion/failure DM from plain text to a display-only
(`octo/v1`) `ResourceCard`. Builds on the dispatcher foundation merged in #577.
octo-server scope only; **DM (person) targets only**.

## Related Issue

Refs #571. Foundation: #577 (merged).

## Linked Spec

- `.octospec/tasks/card-message-internal-dispatch/brief.md` (pilot table + Decisions 1/3/4/11/14)
- `.octospec/tasks/card-message-internal-dispatch/summary-notify-contract.md` (cross-repo wire contract)

## Changes

- **`summary` bot** — provision a dedicated User Bot (`user`+`app`+`robot.status=1`)
  via the same idempotent pattern as `ensureNotifyBot`; the `notification` bot keeps
  the generic text path untouched.
- **Producer wiring** — register the `summary-notify` `ProducerSpec` (DM-only, `octo/v1`,
  `SpacePolicySystemNotification`, `MaxInFlight` 20) in `installCardDispatch`;
  `modules/notify` obtains its bound `Sender` from the registry already in
  `*config.Context` via `SenderFromContext` — no `register.AddModule` change, no
  package-global registration (Decision 1/11).
- **Card path** — add optional structured `NotifyReq.Card` (`SummaryCardFields`, **not**
  a type-17 map). When present, build the card once via
  `cardtmpl.BuildSummaryResourceCard` and dispatch per verified recipient through the
  bound `Sender`, reusing the existing member-verify → actor-exclude → fan-out and
  preserving the `NotifyResp{delivered,filtered}` contract.
- `Payload` and `Card` are **mutually exclusive**: both-present → 400
  `err.server.notify.card_invalid`; both-absent keeps the legacy 400.
- Docs: cross-repo contract + journal/log/pending-learning.

## Testing

```
go vet ./internal/carddispatch/... ./modules/notify/...        # clean
go test -race -cover ./internal/carddispatch/...               # 92.4%
go test ./modules/notify/... ./pkg/cardtmpl/... ./pkg/cardmsg/...  # ok
go run ./tools/lint-card-dispatch ./modules ./internal         # guard green
make i18n-extract-check && make i18n-lint                      # green
git diff --check                                               # clean
```

- [x] Unit/integration tests added/updated
- [x] Manually verified via test suite

## COMPREHENSION

   It turns `modules/notify` into a trusted card *producer*: a structured
   `NotifyReq.Card` request is authored into an `octo/v1` card by `cardtmpl` and
   dispatched through the producer-bound `carddispatch.Sender`, which runs the full
   fail-closed pipeline (live bot identity → live Space/member ACL → `Validate` →
   authoritative `plain`/`space_id` → final size recheck → one transport call). The
   legacy `type=1` text path and Decision 14's raw-card rejection are unchanged —
   `Card` carries structured fields, never a type-17 map.

   The only live dependent is octo-smart-summary via `/v1/internal/notify`; it still
   sends text today, so behavior is unchanged until it opts into `Card`. If it does,
   the per-recipient `NotifyResp{delivered,filtered}` shape is preserved so its
   dedup/retry/sweep keeps working. A non-https base URL fails the card build
   fail-closed; the deep link is dead until octo-web ships `/s/:taskId`. Bad input
   (both `Payload`+`Card`) is rejected 400 rather than silently dropping one.

   `internal/carddispatch` policy-matrix + payload-pipeline tests (92.4% cover),
   `modules/notify` integration tests for the card branch / mutual-exclusion / batch
   rejection, the AST no-bypass guard, and the i18n gate all pass (commands above).

## Cross-repo dependency (must sync)

End-to-end needs two other repos, **out of scope here**:
- **octo-web** — ship the standalone `/s/:taskId?sp={spaceId}` route (mirroring `/d/:docId`);
  until then the card's detail link is dead.
- **octo-smart-summary** — switch the worker terminal-state notification from the text
  payload to the structured `Card` field per `summary-notify-contract.md` (a tracking
  issue will be filed on that repo).

Inert until both land (only the `NOTIFY_INTERNAL_TOKEN` holder can reach it, and it
still sends text). Rollback: `OCTO_CARD_MESSAGE_ENABLED=false` or remove the one spec.

## Checklist

- [x] Updated documentation (contract + journal/log)

### DIFF
--- a/.octospec/journal/shared/summary-notify-pilot.md
+++ b/.octospec/journal/shared/summary-notify-pilot.md
@@ -1,0 +1,79 @@
+---
+type: Journal
+title: "Journal: summary-notify pilot (card-message-internal-dispatch P2)"
+description: Record of enabling the first internal card producer — the summary-notify DM card path in modules/notify on top of the dormant internal/carddispatch foundation.
+tags: ["card", "internal", "trust-boundary", "notify", "producer", "pilot", "cross-repo"]
+timestamp: 2026-07-13T15:44:34Z
+# --- octospec extension fields ---
+task: card-message-internal-dispatch
+upstream: octo-server#571
+source: self
+---
+# Journal: summary-notify pilot (card-message-internal-dispatch P2)
+
+## What was done
+
+Enabled the first production `internal/carddispatch` producer (`summary-notify`)
+on top of the dormant P1 foundation (PR #577 lineage), turning the summary
+completion/failure DM from plain text into a display-only (`octo/v1`)
+ResourceCard. Scope is octo-server only; DM (person) targets only.
+
+- **2a — summary bot.** Added a dedicated `summary` User Bot
+  (`user`+`app`+`robot.status=1`) provisioned via the same idempotent pattern as
+  `ensureNotifyBot`. The `notification` bot keeps the generic text path
+  untouched; identities stay decoupled.
+- **2b — producer wiring.** Registered the `summary-notify` `ProducerSpec`
+  (DM-only, `octo/v1`, `SpacePolicySystemNotification`, `MaxInFlight` 20) in
+  `installCardDispatch`; `modules/notify` obtains its bound `Sender` from the
+  registry already installed in `*config.Context` via `SenderFromContext` —
+  no `register.AddModule` change, no package-global registration (Decision 1/11).
+- **Card path.** `NotifyReq` gained an optional structured `Card` field
+  (`SummaryCardFields`, not a type-17 map). When present, notify builds the card
+  once via `cardtmpl.BuildSummaryResourceCard` and dispatches per verified
+  recipient through the bound `Sender`, reusing the existing member-verify →
+  actor-exclude → fan-out machinery. `Payload` and `Card` are mutually
+  exclusive.
+- **Contract.** `.octospec/tasks/card-message-internal-dispatch/summary-notify-contract.md`
+  captures the cross-repo wire contract for octo-smart-summary.
+
+## Structural learnings / gotchas
+
+- **Reuse the existing ingress, do not add a route.** The brief locks "no new
+  HTTP endpoint" and points the pilot at `NotifyReq`/notify's `memberCache`.
+  The card path therefore extends `NotifyReq` (a structured `Card` field), it
+  is NOT a new `/v1/internal/notify/summary-card` route. Decision 14's raw-card
+  rejection is unaffected because `Card` is structured fields, not a type-17
+  payload map.
+- **Preserve `NotifyResp{delivered,filtered}`.** octo-smart-summary's entire
+  per-recipient dedup/retry/sweep state machine keys off this response. The
+  card path maps `Sender.Send` success → `delivered[uid]`, `target_denied` /
+  `dispatch_failed` → `filtered[uid]=reason`, so the sidecar needs zero
+  response-handling change and its silent-success guard still holds.
+- **Template ownership is server-side.** The sidecar sends raw fields
+  (`task_no`, title, time range, counts, failure reason); octo-server composes
+  labels + the `/s/{task_no}?sp={space_id}` deep link. Deep-link identifier is
+  `summary_task.task_no` (opaque, unique), never the autoincrement `id`.
+- **Dual-mode request needs an explicit both-present reject.** A request struct
+  with two mutually-exclusive modes (`Payload` vs `Card`) must 400 when BOTH are
+  set, not only when both are absent — otherwise one mode is silently dropped.
+  Caught in review (`err.server.notify.card_invalid`).
+- **Deep link requires https.** `cardtmpl` fail-closes on non-https; the base
+  URL source (`External.WebLoginURL` origin vs a new `External.WebBaseURL`) is
+  deferred to the enablement config review, and prod must be https.
+
+## Verification
+
+`go vet` clean; `internal/carddispatch` race+cover 92.4%; `modules/notify`,
+`pkg/cardtmpl`, `pkg/cardmsg` tests green; `lint-card-dispatch` guard green;
+`make i18n-extract-check` + `make i18n-lint` green; `git diff --check` clean.
+
+## Rollout / dependencies
+
+Stacked on the P1 foundation branch (`feat/card-message-internal-dispatch`),
+NOT main — the `internal/carddispatch` foundation is not yet merged to main.
+End-to-end requires two other repos (out of scope here): octo-web must ship the
+`/s/:taskId` route (else the deep link is dead), and octo-smart-summary must
+switch its terminal-state notification from the text payload to the structured
+`Card` field per the contract. Until then this producer is inert (only the
+holder of `NOTIFY_INTERNAL_TOKEN` can reach it, and it still sends text).
+Rollback: global `OCTO_CARD_MESSAGE_ENABLED=false` or remove the one spec.
--- a/.octospec/learnings/pending/dual-mode-request-mutual-exclusion.md
+++ b/.octospec/learnings/pending/dual-mode-request-mutual-exclusion.md
@@ -1,0 +1,28 @@
+---
+type: Learning
+title: "Dual-mode request struct must reject both-present, not just both-absent"
+description: When one request struct carries two mutually-exclusive modes (legacy field vs new structured field), enforce the both-set → 400 guard too, or one mode is silently dropped.
+tags: ["api", "validation", "trust-boundary", "review"]
+timestamp: 2026-07-13T15:44:34Z
+# --- octospec extension fields ---
+source: self
+origin_task: card-message-internal-dispatch
+origin_pr: dmwork-org/octo-server (summary-notify pilot)
+status: pending
+candidate_rule: error-handling
+---
+# Dual-mode request struct must reject both-present, not just both-absent
+
+When an existing ingress struct is extended so one request can arrive in two
+mutually-exclusive shapes — here `NotifyReq.Payload` (legacy text) XOR
+`NotifyReq.Card` (new structured card) — the handler naturally grows a
+"neither is set → 400" guard. It is easy to forget the symmetric case: if a
+caller sets BOTH, the dispatch branch picks one (`Card != nil` wins) and
+silently discards the other, which diverges from the written contract and hides
+caller bugs.
+
+**Rule candidate:** for any dual-mode request field pair, add an explicit
+both-present → 400 rejection (a registered `errcode`, not a raw error) next to
+the both-absent guard. A `switch { case A && B: reject; case !A && !B: reject }`
+makes the exclusivity total. Caught in review on the summary-notify pilot
+(`err.server.notify.card_invalid`).
--- a/.octospec/log.md
+++ b/.octospec/log.md
@@ -361,6 +361,16 @@ change-log convention (§7). Newest first.
   templates, slash commands (PR #418).
 - **Creation** — Dogfood task `member-list-name-fallback` (#344 → PR #420).
 
+## 2026-07-13 (card-message-internal-dispatch P2)
+
+- **Pilot** — Enabled the first `internal/carddispatch` producer
+  (`summary-notify`): dedicated `summary` bot + producer spec + `NotifyReq.Card`
+  structured branch building `octo/v1` DM cards via `cardtmpl` and dispatching
+  through the bound `Sender` (per-recipient fan-out, `NotifyResp` preserved).
+  Stacked on the P1 foundation branch, not main. Cross-repo (octo-web route,
+  octo-smart-summary switch) tracked in the summary-notify contract. See
+  [journal](journal/shared/summary-notify-pilot.md).
+
 ## 2026-06-19 (tooling)
 
 - **Update** — Synced OKF-aware slash commands, workflow skill, and task brief
--- a/.octospec/tasks/card-message-internal-dispatch/summary-notify-contract.md
+++ b/.octospec/tasks/card-message-internal-dispatch/summary-notify-contract.md
@@ -1,0 +1,115 @@
+# summary-notify 卡片通知 — 跨仓 ingress 契约(草稿）
+
+> 仓内草稿,供 octo-server P2(`summary-notify` pilot）实现时对齐,PR 打开后整理成英文
+> issue/comment 同步给 `Mininglamp-OSS/octo-smart-summary`。
+> 关联:brief.md(本目录)、handoff.md、octo-server PR #577(P1 地基)。
+> 状态:待 smart-summary owner 确认字段清单。最后更新 2026-07-13。
+
+## 背景一句话
+
+今天 smart-summary worker 在任务终态后 `POST /v1/internal/notify` 发**纯文本 DM**
+(`payload:{type:1,content:<整段文案>}`),per-recipient 一请求一收件人,dedup/retry 靠
+自己的 `summary_notification` 表 + `Sweep`。P2 把这条 DM **从纯文本升级为 `octo/v1`
+展示卡片**,发送方换成专用 `summary` bot,经 octo-server `internal/carddispatch` 边界
+派发。**不新增 HTTP 路由**(brief Goal / Out-of-scope),而是**扩展现有
+`/v1/internal/notify` 的 `NotifyReq`**。
+
+## 契约要点(锁定项)
+
+1. **同端点、同 token、同粒度**:仍是 `POST /v1/internal/notify`,头 `X-Internal-Token`,
+   **一请求一收件人**(`targets` 单元素)。smart-summary 的 claim/markSent/markFailed/
+   Sweep 状态机、`space_id`/`service`/`actor_uid` 语义**全部不变**。
+2. **`payload` 与 `card` 二选一**:
+   - 发文本(现状,其它 service 继续用):填 `payload:{type:1,content:...}`,`card` 省略。
+   - 发卡片(summary pilot):填结构化 `card`,`payload` 省略(server 侧 binding 从
+     `required` 放宽为二者其一)。
+3. **禁止客户端手搓 type-17 map**:Decision 14 仍然生效——`payload` 若被
+   `cardmsg.IsCardPayload` 判为卡片(`type=17`),**照旧 400 拒绝**。卡片只能走
+   结构化 `card` 字段,由 octo-server 用模板生成 type-17 wire,smart-summary 永远不构造
+   Adaptive Card JSON。
+4. **响应契约不变**:仍返回 `NotifyResp{delivered:[], filtered:{uid:reason}}`。卡片派发
+   成功 → uid 进 `delivered`;`target_denied`/`dispatch_failed`/建卡失败 → uid 进
+   `filtered`。smart-summary 现有「只认 `delivered[]` 判真送达 + 非成员进 filtered 重试」
+   的逻辑**一行不用改**(`recipientIsActiveMember` 预检可保留,carddispatch 的 target
+   校验会 fail-closed 兜底)。
+5. **模板/文案/链接归属 octo-server**:smart-summary **只发原始字段**;卡片布局、按钮
+   文案(查看详情/复制)、FactSet 标签、`/s/{task_no}?sp={space_id}` deep-link,全部由
+   octo-server `pkg/cardtmpl` + `i18n.OutboundLanguage` 生成。smart-summary 不再拼
+   「查看结果:<link>」(旧字段 `payload.result_url` 作废)。
+
+## 请求体(卡片模式)
+
+```jsonc
+POST /v1/internal/notify
+X-Internal-Token: <SUMMARY_NOTIFY_TOKEN>
+{
+  "space_id":  "spc_xxx",          // 现状:必填,server 用 memberCache 校验收件人
+  "service":   "summary-service",  // 现状不变
+  "targets":   ["uid_recipient"],  // 现状:单收件人
+  "actor_uid": "",                 // 现状:DM 通知留空
+  "card": {                        // 新增:非空即走 summary-notify 卡片 producer
+    "task_no":    "TN_20260713_abcd",   // ★ 见下「标识」
+    "kind":       "completed",          // "completed" | "failed"
+    "title":      "产品周会纪要",         // 原始标题(server 负责转义/截断)
+    "time_range": "2026-07-06 10:00 ~ 2026-07-13 10:00", // 见下「时间/计数」
+    "members":    5,                    // 参与人数;<=0 省略该行
+    "msg_count":  128,                  // 消息条数;<=0 省略该行
+    "generated_at": "2026-07-13 15:04", // 完成时间字符串;空则省略
+    "reason":     ""                    // failed 时的脱敏原因;completed 留空
+  }
+}
+```
+
+`payload` 与 `card` 同时缺失 → 400;两者同时存在 → 400(避免歧义)。
+
+## 三个约定(建议默认,待确认)
+
+- **标识 `task_no`(不是自增 `id`)**:deep-link `/s/{task_no}?sp={space_id}` 用
+  `summary_task.task_no`(varchar unique)。理由:不可枚举 + 与前端
+  `WKApp.openSummaryDetail` 对齐。**待 smart-summary 确认前端 detail 入口用的是 task_no
+  还是 id。**
+- **时间字段传「已格式化字符串」**:`time_range` / `generated_at` 由 smart-summary 用它
+  自己的 `internal/timezone`(东八区)格式化后传字符串,octo-server 原样填进 FactSet 值。
+  理由:octo-server 无该业务时区配置,避免时区漂移。**标签(“时间范围”)仍由 octo-server
+  按收件人语言本地化。**
+- **计数传 int**:`members` / `msg_count` 传整数,octo-server 按 `i18n.OutboundLanguage`
+  组装「参与成员:N 人」「消息数量:N 条」,单位随语言。
+
+## 字段来源(smart-summary 侧现成)
+
+| card 字段 | smart-summary 来源 |
+| --- | --- |
+| `task_no` | `SummaryTask.TaskNo` |
+| `kind` | `kindForStatus(status)`(completed/failed) |
+| `title` | `SummaryTask.Title` |
+| `time_range` | `formatTimeRange(task)`(现有) |
+| `members` | `participantCount(task)`(现有) |
+| `msg_count` | `summary_result.total_msg_count`(现有 `resultMeta`) |
+| `generated_at` | `summary_result.generated_at` 格式化(现有 `resultMeta`) |
+| `reason` | 现有 `errorSanitizer` 脱敏后的失败原因 |
+
+空间名(现在 smart-summary 自己 resolve)在卡片里不再需要——octo-server 侧可按 space_id
+自行渲染或省略,减少两边重复。
+
+## octo-server 侧(本 PR 落地,cross-repo 之外)
+
+- 新增专用 `summary` bot(user+app+robot.status=1),`notification` bot 只留纯文本路径。
+- 注册 `summary-notify` producer(DM/`octo/v1`/system-notification/MaxInFlight 20),
+  经 `SenderFromContext` 注入 notify。
+- `card` 分支:`memberCache.verify` → `cardtmpl` 建卡 → 每个成员 `sender.Send` → 汇总
+  `delivered/filtered`。**建卡/配置(如 deep-link 非 https)失败降级为原纯文本 DM**,
+  保证通知必达(此为 octo-server 侧策略,brief 未强制,PR 说明)。
+
+## enablement 门(端到端上线前必须齐)
+
+1. **octo-web** 上线 `/s/:taskId` 路由(否则「查看详情」死链)——独立 PR;
+2. octo-server 落 `cardtmpl` snapshot/Validate + deep-link 形状测试(本 PR);
+3. **smart-summary** 切换发送从 `payload` 文本 → `card` 字段(独立 PR,最后做)。
+
+在 (1)(3) 到位前,octo-server 侧行为惰性:无 `card` 请求即无卡片流量;全局
+`OCTO_CARD_MESSAGE_ENABLED` 为总开关,回滚可移除 producer spec 或关该 env。
+
+## 明确不在本契约内
+
+原群/thread 回发(member-exempt 群发)、A2 用户转发卡片、docs 分享卡片、per-channel
+限速与 cluster cap——均为后续独立任务(brief Out-of-scope)。
--- a/main.go
+++ b/main.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Mininglamp-OSS/octo-lib/common"
 	"github.com/Mininglamp-OSS/octo-lib/config"
 	"github.com/Mininglamp-OSS/octo-lib/module"
 	libdb "github.com/Mininglamp-OSS/octo-lib/pkg/db"
@@ -23,10 +24,12 @@ import (
 	commonapi "github.com/Mininglamp-OSS/octo-server/modules/base/common"
 	"github.com/Mininglamp-OSS/octo-server/modules/base/event"
 	"github.com/Mininglamp-OSS/octo-server/modules/botidentity"
+	"github.com/Mininglamp-OSS/octo-server/modules/notify"
 	"github.com/Mininglamp-OSS/octo-server/modules/user"
 	"github.com/Mininglamp-OSS/octo-server/pkg/accesslog"
 	"github.com/Mininglamp-OSS/octo-server/pkg/auth"
 	"github.com/Mininglamp-OSS/octo-server/pkg/avatarrender"
+	"github.com/Mininglamp-OSS/octo-server/pkg/cardmsg"
 	octodb "github.com/Mininglamp-OSS/octo-server/pkg/db"
 	octoi18n "github.com/Mininglamp-OSS/octo-server/pkg/i18n"
 	"github.com/Mininglamp-OSS/octo-server/pkg/metrics"
@@ -351,9 +354,34 @@ func installCardDispatch(ctx *config.Context) error {
 		Metrics:          carddispatch.NewMetrics(prometheus.DefaultRegisterer),
 		Logger:           log.NewTLog("CardDispatch"),
 	}
-	registry := carddispatch.NewRegistry(deps, nil)
+	// summary-notify pilot (brief › pilot table, all rows confirmed 2026-07-13):
+	// dedicated `summary` User Bot, DM-only, display-only octo/v1,
+	// system-notification Space policy, in-flight 20/process. modules/notify
+	// obtains this producer's bound Sender via carddispatch.SenderFromContext.
+	//
+	// Registration is inert until modules/notify receives a structured card
+	// request (NotifyReq.Card) AND OCTO_CARD_MESSAGE_ENABLED is on; end-to-end
+	// enablement additionally depends on octo-web shipping the /s/:taskId route
+	// and octo-smart-summary switching its send from text to the card fields.
+	specs := []carddispatch.ProducerSpec{
+		{
+			ID:                  summaryNotifyProducerID,
+			Enabled:             true,
+			SenderUID:           notify.SummaryBotUIDValue,
+			AllowedChannelTypes: []uint8{common.ChannelTypePerson.Uint8()},
+			AllowedProfiles:     []string{cardmsg.ProfileV1},
+			SpacePolicy:         carddispatch.SpacePolicySystemNotification,
+			GroupPolicy:         carddispatch.GroupPolicyMemberRequired, // unused at DM pilot; must be a valid value
+			MaxInFlight:         20,
+		},
+	}
+	registry := carddispatch.NewRegistry(deps, specs)
 	return carddispatch.Install(ctx, registry)
 }
+
+// summaryNotifyProducerID mirrors modules/notify's producer ID. Kept here so the
+// composition root can register the spec without importing an unexported const.
+const summaryNotifyProducerID = carddispatch.ProducerID("summary-notify")
 
 func printServerInfo(ctx *config.Context) {
 	infoStr := `
--- a/main.go
+++ b/main.go
@@ -354,6 +354,11 @@ func installCardDispatch(ctx *config.Context) error {
 		Metrics:          carddispatch.NewMetrics(prometheus.DefaultRegisterer),
 		Logger:           log.NewTLog("CardDispatch"),
 	}
+	registry := carddispatch.NewRegistry(deps, summaryNotifyProducerSpecs())
+	return carddispatch.Install(ctx, registry)
+}
+
+func summaryNotifyProducerSpecs() []carddispatch.ProducerSpec {
 	// summary-notify pilot (brief › pilot table, all rows confirmed 2026-07-13):
 	// dedicated `summary` User Bot, DM-only, display-only octo/v1,
 	// system-notification Space policy, in-flight 20/process. modules/notify
@@ -363,7 +368,7 @@ func installCardDispatch(ctx *config.Context) error {
 	// request (NotifyReq.Card) AND OCTO_CARD_MESSAGE_ENABLED is on; end-to-end
 	// enablement additionally depends on octo-web shipping the /s/:taskId route
 	// and octo-smart-summary switching its send from text to the card fields.
-	specs := []carddispatch.ProducerSpec{
+	return []carddispatch.ProducerSpec{
 		{
 			ID:                  summaryNotifyProducerID,
 			Enabled:             true,
@@ -375,8 +380,6 @@ func installCardDispatch(ctx *config.Context) error {
 			MaxInFlight:         20,
 		},
 	}
-	registry := carddispatch.NewRegistry(deps, specs)
-	return carddispatch.Install(ctx, registry)
 }
 
 // summaryNotifyProducerID mirrors modules/notify's producer ID. Kept here so the
--- a/main_carddispatch_test.go
+++ b/main_carddispatch_test.go
@@ -4,6 +4,11 @@ import (
 	"os"
 	"strings"
 	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-server/internal/carddispatch"
+	"github.com/Mininglamp-OSS/octo-server/modules/notify"
+	"github.com/Mininglamp-OSS/octo-server/pkg/cardmsg"
 )
 
 func TestCardDispatchRegistryInstalledBeforeModuleConstruction(t *testing.T) {
@@ -22,12 +27,35 @@ func TestCardDispatchRegistryInstalledBeforeModuleConstruction(t *testing.T) {
 	}
 }
 
-func TestDormantFoundationRegistersNoProductionProducer(t *testing.T) {
-	source, err := os.ReadFile("main.go")
-	if err != nil {
-		t.Fatalf("read main.go: %v", err)
+func TestSummaryNotifyIsOnlyProductionCardProducer(t *testing.T) {
+	specs := summaryNotifyProducerSpecs()
+	if len(specs) != 1 {
+		t.Fatalf("production registry has %d producers; want summary-notify only", len(specs))
 	}
-	if !strings.Contains(string(source), "carddispatch.NewRegistry(deps, nil)") {
-		t.Fatal("foundation rollout must install an empty registry until cross-repo enablement gates pass")
+
+	spec := specs[0]
+	if spec.ID != carddispatch.ProducerID("summary-notify") {
+		t.Fatalf("producer ID = %q; want summary-notify", spec.ID)
+	}
+	if !spec.Enabled {
+		t.Fatal("summary-notify producer must be enabled")
+	}
+	if spec.SenderUID != notify.SummaryBotUIDValue {
+		t.Fatalf("sender UID = %q; want %q", spec.SenderUID, notify.SummaryBotUIDValue)
+	}
+	if len(spec.AllowedChannelTypes) != 1 || spec.AllowedChannelTypes[0] != common.ChannelTypePerson.Uint8() {
+		t.Fatalf("allowed channel types = %v; want DM only", spec.AllowedChannelTypes)
+	}
+	if len(spec.AllowedProfiles) != 1 || spec.AllowedProfiles[0] != cardmsg.ProfileV1 {
+		t.Fatalf("allowed profiles = %v; want octo/v1 only", spec.AllowedProfiles)
+	}
+	if spec.SpacePolicy != carddispatch.SpacePolicySystemNotification {
+		t.Fatalf("space policy = %q; want system_notification", spec.SpacePolicy)
+	}
+	if spec.GroupPolicy != carddispatch.GroupPolicyMemberRequired {
+		t.Fatalf("group policy = %q; want member_required", spec.GroupPolicy)
+	}
+	if spec.MaxInFlight != 20 {
+		t.Fatalf("max in flight = %d; want 20", spec.MaxInFlight)
 	}
 }
--- a/modules/notify/api.go
+++ b/modules/notify/api.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"sync"
+	"sync/atomic"
 
 	"github.com/Mininglamp-OSS/octo-lib/config"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
@@ -44,7 +45,7 @@ type Notify struct {
 	botMu         sync.Mutex
 	botOK         bool
 	summaryBotMu  sync.Mutex
-	summaryBotOK  bool
+	summaryBotOK  atomic.Bool
 	cardSender    carddispatch.Sender
 	internalToken string
 	log.Log
@@ -105,7 +106,7 @@ func New(ctx *config.Context) *Notify {
 			n.Info("Notify bot ready")
 		}
 		n.ensureSummaryBotReady()
-		if n.summaryBotOK {
+		if n.summaryBotOK.Load() {
 			n.Info("Summary bot ready")
 		}
 	}()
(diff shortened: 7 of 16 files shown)


OctoBoooot (approved):
## Review: summary-notify card pilot — first internal card producer (#579)

**Verdict:** Approve, conditional on Test green (Build/Card Dispatch Lint/check-sprint all green). This is the **P2 activation** I flagged on #577 as requiring its own trust-tier review — the first live `internal/carddispatch` producer, making the dispatcher's authz load-bearing. Reviewed at that rigor; it feeds the verified fail-closed dispatcher correctly rather than bypassing it. needs-human-review applies (live trust-boundary activation). check-sprint is green this time.

### The activation is correctly scoped (main.go)
- `summary-notify` ProducerSpec: `AllowedChannelTypes: [ChannelTypePerson]` (DM-only, matches "DM targets only"), `AllowedProfiles: [ProfileV1]` (octo/v1 display-only), `SpacePolicySystemNotification`, bound to a single `SummaryBotUIDValue`, `MaxInFlight: 20`. `NewRegistry(deps, specs)` — now a real producer vs #577's `nil`. This is the dormant→live transition, scoped tightly.

### Card path routes through the bound Sender — NOT a raw type-17 (byte-verified)
- `notify/api.go`: the new `Card` field is structured `SummaryCardFields`, not a type-17 map, so it doesn't touch (or need) the Decision-14 ingress gate — it's a separate typed channel. `deliverCardNotification` dispatches via `n.cardSender.Send(...)` (card.go:109), the capability-bound Sender from `SenderFromContext(ctx, summaryNotifyProducerID)` → runs the dispatcher's fixed pipeline + fail-closed `DBAuthorizer`.
- **Mutual exclusion tight:** `Card != nil && len(Payload) > 0 → 400 ErrNotifyCardInvalid`; both-absent → legacy 400; batch rejects any `Card` entry. The text and card paths can't be combined to smuggle a payload.

### Dispatcher authz is load-bearing now and fail-closed for this producer (the #577 flag, discharged)
- `authorizeDM`: requires the **recipient** be an active space member (`space_member ... status=1` → else deny "recipient is not an active member"), the `SpacePolicySystemNotification` path (line 34) lets the notification bot DM an active member without itself being a space member (correct for a system notifier), and **every DB error denies**. So the live producer can only deliver to active members of the target space, fail-closed on any error.
- **Defense-in-depth:** producer-side member-verify (card.go:58, actor-excluded, non-members → Filtered) PLUS the dispatcher's independent re-verify — two fail-closed gates, "a non-member never reaches transport."
- Degradation: sender unavailable / cardmsg disabled / card build fails (e.g. non-HTTPS deep-link origin) → falls back to `sendSummaryText`, not a hard failure or a malformed card.

### Tests
- Trust-relevant coverage: `ValidationRejectsBadFields`, `RejectsNonHTTPSDeepLinkOrigin` (bad card → degrade, not send), `DegradesToTextWhenSenderUnavailable`, and both mutual-exclusion cases (single `NotifyRejectsPayloadAndCardTogether` + `NotifyBatchRejectsCardField`). 92.4% on carddispatch.

### Praise
- The activation respects the boundary #577 built: the producer emits *structured card fields* through the *bound Sender*, so it inherits the dispatcher's fail-closed authz + fixed pipeline rather than hand-rolling a type-17 map that would rely on the ingress gate. DM-only + single-bot + SystemNotification is the minimal viable capability for a pilot, and the text-fallback keeps a card-build failure from dropping the notification. The mutual-exclusion + batch-card rejection close the "combine two paths" gap cleanly.

### Forward note
- This discharges the #577 forward-flag (dormant→live producer gets trust-tier review). The dispatcher authz is now genuinely exercised in production for the DM/SystemNotification path; any *future* producer that adds group/thread targets or a different SpacePolicy will re-activate the currently-`GroupPolicyMemberRequired`-but-unused group authz path — that path should get its own focused review when first used (it's declared here but DM-pilot doesn't exercise it).

### CI
- Build/Card Dispatch Lint/check-sprint green; Test/Vet in-progress. Approval conditional on Test green (a live trust-boundary producer's tests are load-bearing). needs-human-review applied. mergeStateStatus BLOCKED until CI + human sign-off + approvals.

Jerry-Xin (changes requested):
Trust boundary is clean: the internal endpoint uses subtle.ConstantTimeCompare, fail-closed when NOTIFY_INTERNAL_TOKEN is unset (rejects all), cards are internal-token-only, DM-only, display-only. Injection surface is handled: BuildSummaryResourceCard escapes all fields (markdown), enforces length limits, and forces https/URL-escaped deep links. Build failure fails closed (degrades to plain text). No auth/injection/fail-open holes.

## 🔴 Blocking

**1. Data race on `botOK` (api.go:45) — broken double-checked locking.**
`botOK` is guarded by `botMu`, but the fast-path/status reads happen outside the lock while writes happen under it:
- Unlocked reads: `api.go:104`, `api.go:316` (double-check fast path), `api.go:323`
- Locked writes: `api.go:101` (init goroutine), `api.go:319` (send path)
The unlocked reads race the locked writes. Fix: use `sync/atomic` (atomic.Bool) or route every access through `botMu`.

**2. Data race on `summaryBotOK` (api.go:47) — same broken pattern (copied from `botOK`).**
- Unlocked reads: `api.go:108`, `api.go:120` (fast path), `card.go:69` (send path)
- Locked writes: `api.go:125`
Same fix. Note the existing `-race` coverage runs only against carddispatch, not modules/notify, so current tests do not catch either race.

**3. Empty-payload bypasses the payload/card mutual exclusion (api.go:181).**
The mutex gate uses `len(req.Payload) > 0`, so `payload:{}` + `card:{...}` slips through and silently takes the card path, violating the PR's own stated mutual-exclusion contract. Use `req.Payload != nil`. Also note the batch endpoint accepts empty payload while the single endpoint rejects it — asymmetric validation; align them. Add a `payload:{}`+card test case.

## 🟡 Non-blocking

- modules/notify lacks unit tests for the card success path and the category→Filtered mapping (a fake carddispatch.Sender would lock these down). Add a `-race` test target covering modules/notify to catch the races above.
- Build-failure fallback: the contract text says the result should be `filtered`, but the implementation degrades to text and marks it delivered. Align so octo-smart-summary derives correct retry semantics.

## ✅ Highlights
- Constant-time token compare + fail-closed when token unset.
- Thorough card-content escaping and deep-link hardening.
- Card production correctly gated to internal token holders, DM-only, display-only.

Two distinct fields (`botOK`, `summaryBotOK`) share the identical race; fix both, then this is close.

OctoBoooot (changes requested):
## Correction: reversing my APPROVE → REQUEST_CHANGES on #579 (credit: Jerry-Xin)

I was wrong to APPROVE. Jerry-Xin's 🔴 on `summaryBotOK` is a **real data race** I missed — I verified the trust/authz boundary exhaustively but did not scrutinize the bot-provisioning concurrency. Byte-verified below. Reversing. (His 🔴 on the empty-payload mutual-exclusion is real too but benign in effect — I grade it minor, detail below.)

### Blocker — data race on `summaryBotOK` (byte-verified)
Broken double-checked-locking: the write is synchronized, the reads are not, so there is no happens-before edge between the init goroutine's write and the request-handler reads.
- **Write:** `New()` (api.go:~100) spawns a goroutine that calls `ensureSummaryBotReady()`, which writes `n.summaryBotOK = ...` under `n.summaryBotMu` (api.go:123-127). Locked.
- **Unlocked reads racing that write:**
  - `card.go:69` — `if !n.summaryBotOK` in `deliverCardNotification`, a **request handler** running concurrently with the init goroutine. No lock.
  - `api.go:120` — the fast-path `if n.summaryBotOK { return }` in `ensureSummaryBotReady` itself (the outer half of the double-check) reads unlocked.
  - `api.go:108` — the init goroutine's own post-provision read, unlocked.
- A plain `bool` read racing a concurrent write is a data race (UB per the Go memory model; `go test -race` will flag it). On a live trust-boundary PR this is a blocker, not cosmetic.
- **Fix:** make the flag access consistent — either read `summaryBotOK` under `summaryBotMu` at all sites (card.go:69, api.go:108/120), or switch it to `atomic.Bool` (cleanest for a double-checked fast-path: `Load()` at the outer check, `Store()` under the lock). The request-path read in card.go:69 is the one that matters most (concurrent with provisioning).

### Minor (not 🔴) — empty-object payload slips mutual exclusion
- `case req.Card != nil && len(req.Payload) > 0` (api.go:181) misses `Card != nil` + `Payload == {}` (empty non-nil map): `len({})==0` → neither exclusion case fires → falls through to the card path. So the contract ("payload and card mutually exclusive") is violated in the letter — both fields present is accepted.
- But the effect is benign: the card path (`deliverCardNotification`) uses `req.Card` and never reads `req.Payload`, and `IsCardPayload({})` is false (no `type`), so the empty payload is inert — no double-send, no smuggling. It's a validation-strictness gap, not a correctness/security defect. Grade: minor. Fix: `len(req.Payload) > 0` → `req.Payload != nil` in the exclusion case, for contract cleanliness.

### Jerry-Xin's 🟡s — agree
- More card-path/category-mapping unit coverage: fair (current tests cover build/degrade/mutual-exclusion but not the fan-out category mapping in depth).
- Build-failure vs cross-repo contract consistency: worth aligning the doc.

### What's still correct (unchanged from my prior review)
- The activation's TRUST properties hold: DM-only/single-bot/SystemNotification scoping, card routes through the bound Sender (not a raw type-17), dispatcher `authorizeDM` fail-closed (recipient must be active member), tight-in-effect mutual exclusion, text-fallback degradation. The race is orthogonal to the authz boundary — it's a provisioning-flag concurrency bug, not a trust-gate hole.

### On my miss
- I reviewed the trust boundary (the headline risk) thoroughly and signed off, but didn't check the concurrency of the bot-readiness flag that the new card path introduced. Lesson: a PR that adds a request-path read of shared state provisioned by a background goroutine needs a concurrency pass, not just an authz pass — especially when I can see a `sync.Mutex` field, which signals the author knew it was shared but didn't guard every access. Credit Jerry-Xin for the race.

### Verdict
- REQUEST_CHANGES: 🔴 the `summaryBotOK` data race (sync all reads or make it atomic). Minor: empty-payload exclusion + the two doc/test 🟡. The trust-boundary logic itself is sound — this reversal is the provisioning-flag race, which I should have caught. Once the flag access is synchronized, this is an APPROVE.

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD 64df30bf

## Lens Summary

| Lens | Verdict | Headline |
|---|---|---|
| code | changes-requested | Producer activation breaks the base-branch dormancy guard → CI Test FAILURE. |
| security | cleared | Trust boundary double-verifies fail-closed; inputs escaped; https deep-link; no leaks. |
| qa | fail | CI Test check RED — PR breaks `TestDormantFoundationRegistersNoProductionProducer`. |

## Critical Findings (P0-P1)

- `[P0-1]` `main.go:313` (+ base `main_carddispatch_test.go:30`) — This PR flips `carddispatch.NewRegistry(deps, nil)` → `NewRegistry(deps, specs)` to register the `summary-notify` producer, but the base-branch guard `TestDormantFoundationRegistersNoProductionProducer` asserts main.go must still contain `NewRegistry(deps, nil)` ("foundation rollout must install an empty registry until cross-repo enablement gates pass"). The guard is neither updated nor removed, so the root-package `go test` and the CI **Test** check are **RED**. The PR "Testing" block ran only `./modules/notify/...` + `./internal/carddispatch/...`, never the root package, so the author's local run missed it. Flagged by code/C-1 + qa/Q-1 (consensus). — **Fix**: decide the activation intent (see [Q-1]); if activating on main deliberately, replace the guard with one asserting the `summary-notify` spec IS registered and run `go test ./...` before re-push; if the cross-repo gates are not yet met, keep the registry dormant.
- `[P1-1]` `modules/notify/card.go:634` + `modules/notify/api.go` (`ensureSummaryBotReady` fast-path + `New()` goroutine read) — `summaryBotOK` is written under `summaryBotMu` but read without the lock on the request path (`deliverCardNotification`) and in the double-checked fast-path — a data race the Go race detector will flag. modules/notify tests do not run with `-race`, so it is uncaught locally. Flagged by code/C-2 + security/S-1 (consensus; also raised by both prior PR bot reviewers). Note: it mirrors the pre-existing `botOK` pattern, so it is consistent debt rather than a brand-new defect, and security judges it non-security (flag only transitions false→true; worst case is a spurious retry, never a wrong recipient). — **Fix**: convert `summaryBotOK` (and ideally `botOK`) to `atomic.Bool`, or hold `summaryBotMu` on every read; add a `-race` run for modules/notify.

## Should-Fix (P2)

- `[P2-1]` `modules/notify/card.go:668-706` / `card_test.go` — The successful card-dispatch path is untested: no test wires a fake `carddispatch.Sender` to assert (a) all-success → `delivered`, or (b) a categorized error → `filtered[uid]=CategoryOf(err)`. Only validation/build/non-https/text-degrade/mutual-exclusion are covered, never the primary happy path (AC3/AC5) — the exact `delivered`/`filtered` mapping the cross-repo consumer depends on. Flagged by qa/Q-2. — **Fix**: add a `deliverCardNotification` test with a fake `Sender` covering success and mixed success + `dispatch_failed`/`target_denied` category mapping.
- `[P2-2]` `summary-notify-contract.md:178` vs `modules/notify/card.go:644-659` — Contract clause 4 states card-build failure routes the uid to `filtered`, but the implementation degrades a build/config failure (e.g. non-https origin) to a **text DM marked `delivered`** — and the contract's own octo-server section (line 246) documents that text-degrade. Clause 4 is internally contradictory and misleads octo-smart-summary's retry/dedup expectations. Flagged by qa/Q-3. — **Fix**: reconcile clause 4 to state build failure degrades to text → `delivered`, matching the implemented fail-open behavior.
- `[P2-3]` `main.go:319` vs `modules/notify/api.go:342` — The producer ID literal `"summary-notify"` is duplicated as two independent consts across the package boundary. If either drifts, `SenderFromContext` returns not-found and the feature silently degrades to text forever (only a `Warn`). Flagged by code/C-3. — **Fix**: export one const and reference it from both sites.

## Suggestions (P3)

- `[P3-1]` `modules/notify/api.go:424` — the both-present exclusion guard uses `len(req.Payload) > 0`, so `"payload":{}` (non-nil empty map) + `card` slips the mutual-exclusion check and silently takes the card path. Harm is minimal (empty map carries no text) but it diverges from the "both-present → 400" contract. Flagged by code/C-4 + qa/Q-4. — Fix: use `req.Payload != nil` for the both-present arm.
- `[P3-2]` `modules/notify/card.go:640,647,675` — every downstream call uses `context.Background()`; the inbound request context is never plumbed in, so no request timeout/cancellation reaches card build or per-recipient `Sender.Send` (text path shares this trait). Flagged by code/C-5. — Fix: thread `c.Request.Context()` from `sendNotify` into the fan-out.
- `[P3-3]` `modules/notify/card.go:817-848` — card/fallback labels are hardcoded zh/en in Go with a `lang` prefix-match, while this same PR's errcodes go through the i18n toml catalog — two localization mechanisms for one feature. Flagged by code/C-6. — Fix: move labels to the i18n catalog, or note the deliberate outbound-language deviation.
- `[P3-4]` `modules/notify/bot_manager.go:533,558` — the now-shared `ensureBot` keeps `notify`-specific log strings ("更新notify bot名称失败", "Notify bot 创建成功"), so summary-bot provisioning mislabels itself in logs. Flagged by qa/Q-5. — Fix: parameterize the log message with `botName`/`botUID`.

## Questions for Author

- `[Q-1]` The journal states "Stacked on the P1 foundation branch, not main … producer is inert," yet `baseRefName` is `main` and the spec ships `Enabled:true` while both cross-repo enablement gates (octo-web `/s/:taskId`, smart-summary switch) are admittedly unmet. Is this PR meant to merge to main now (requiring the guard-test update in P0-1) or was it opened against the wrong base? This determines whether the fix is "update the guard" or "keep dormant / re-target." (from code/Q-1)
- `[Q-2]` `modules/notify/card.go:598` — the new `len(req.Targets) > 200` cap has no boundary test and the legacy text path has no equivalent cap; is the asymmetry intended for a contract that specifies a single recipient? (from qa/Q-6)

## Verdict Rationale

`REQUEST_CHANGES`: qa lens = `fail` (CI **Test** check is RED on a load-bearing base-branch guard the diff directly contradicts) and code lens = `changes-requested`, both first-match triggers. Cross-lens consensus on the two lead findings (P0-1 guard/CI break: code+qa; P1-1 data race: code+security, also both prior PR reviewers). Security lens = `cleared` with no blocking finding — the trust boundary double-verifies recipient membership fail-closed even under the system-notification policy, attacker-influenceable card/deep-link fields are escaped/bounded and https-enforced, no secrets or sensitive log disclosure — so no embargo. No lens returned UNKNOWN.

## Round Context

|---|---|---|
| main.go | ✅ (code/sec/qa) | P0-1, P2-3, Q-1 |
| main_carddispatch_test.go (base) | ✅ (code/qa) | P0-1 |
| modules/notify/api.go | ✅ (all) | P1-1, P3-1 |
| modules/notify/card.go | ✅ (all) | P1-1, P2-1, P2-2, P3-2, P3-3, Q-2 |
| modules/notify/bot_manager.go | ✅ (code/qa) | P3-4 |
| modules/notify/card_test.go | ✅ (code/qa) | P2-1 |
| modules/notify/model.go | ✅ (code/sec/qa) | — |
| pkg/errcode/notify.go | ✅ (all) | — |
| pkg/cardtmpl/resource.go (context) | ✅ (sec) | — (escape + https deep-link verified) |
| internal/carddispatch/*.go (context) | ✅ (sec) | — (authorizer double-verify verified) |
| pkg/i18n + tools/i18nmarkers toml | ✅ | — |
| .octospec/* docs | ✅ | P2-2 (contract) |

## Notes

- Confidence filter aggregate: ~3 dropped, ~4 downgraded across lenses (speculative fan-out race dropped — `filteredMap` is written only by the single collecting goroutine, so the per-recipient fan-out is race-safe).

---

lml2468 (dismissed):
REQUEST_CHANGES — Round 1

CI Test check is RED: this PR registers the `summary-notify` producer (`NewRegistry(deps, nil)` → `deps, specs`) but the base-branch guard `TestDormantFoundationRegistersNoProductionProducer` still requires an empty registry until the cross-repo enablement gates land, so `go test ./...` fails. Also a data race on `summaryBotOK` (unsynchronized reads) flagged by code + security. Security lens is otherwise clear (fail-closed double-verify, escaped inputs, https deep-link).

Full aggregate: see the PR issue comment above and the multica tracker (OCTO-2283).

yujiawei (changes requested):
# Code Review — PR #579 (octo-server)

Reviewed at head SHA `64df30bfc3a4a7ccaaead9668db2ff2644b1c4db` against merge-base `41aa4c57`. Checked out the PR, read the full diff plus the surrounding `internal/carddispatch`, `pkg/cardtmpl`, and `pkg/cardmsg` code, and ran `go build ./...`, `go vet`, and the package tests (all green; `internal/carddispatch` 92.4%, `modules/notify` 64.5% under `-race`).

## 1. Spec / contract compliance

Measured against `summary-notify-contract.md` and the linked brief. **Compliant.**

- Dual-mode `payload` XOR `card`: both-present → 400 `err.server.notify.card_invalid`, both-absent → legacy 400 (`modules/notify/api.go:180-187`). Total exclusivity as the contract requires.
- Client-authored type-17 maps still rejected via `cardmsg.IsCardPayload` (`api.go:271`, `api.go:239`).
- DM-only / `octo/v1` / `system_notification` / `MaxInFlight 20` producer spec (`main.go:367-376`); no new HTTP route; no `register.AddModule` change; `Sender` obtained via `SenderFromContext` (`api.go:73`).
- `NotifyResp{delivered, filtered}` preserved; per-recipient dispatch failures land in `filtered` with a category reason (`card.go:422-424`), so the caller's retry/dedup state machine is untouched.
- Labels / layout / `/s/{task_no}?sp={space_id}` deep-link owned server-side; non-https origin fails the build fail-closed and degrades to text (`card.go:382-398`, verified by `TestBuildSummaryCard_RejectsNonHTTPSDeepLinkOrigin`).
- Summary bot provisioned via the shared idempotent `ensureBot` path (`bot_manager.go:54`), and `installCardDispatch` runs before module construction (`main.go:232`), so the registry is populated when `notify.New` reads it.

No missing, extra, or divergent behavior relative to the contract.

## 2. Code quality

### P1 — Data race on `summaryBotOK` (blocking)

`summaryBotOK` is written under `summaryBotMu` in `ensureSummaryBotReady` (`api.go:124-125`) but read **lock-free** in three places: `api.go:108`, `api.go:120`, and — on the request-serving path — `card.go:69` (`if !n.summaryBotOK`). The startup provisioning goroutine (`api.go:107`) writes the flag concurrently with incoming card requests that read it, which is a data race under the Go memory model.

This is not theoretical. Reproduced by driving the exact access pattern (lock-free read vs. mutex-guarded write) through `go test -race`:

```
WARNING: DATA RACE
Write at 0x... by goroutine 85:   (n.summaryBotOK = ... under lock)
Previous read at 0x... :          (if !n.summaryBotOK, lock-free)
```

CI runs `go test -race -shuffle=on` (`.github/workflows/ci.yml:259`). The current suite happens not to drive concurrent card sends during the startup window, so it stays green today — but the defect is real on a load-bearing notification path and the fix is one line: make `summaryBotOK` an `atomic.Bool` (and read it via `.Load()`), or read it under `summaryBotMu`.

Note: the pre-existing `botOK` field (`api.go:100/104/316-323`) has the identical lock-free-read pattern, so it races too. I'd recommend fixing both while here, but at minimum the newly-added path must be synchronized.

### P2 — Batch endpoint accepts an empty `{}` payload (asymmetry)

After dropping `binding:"required"` from `NotifyReq.Payload` (`model.go`), the single endpoint rejects an empty payload map with 400 (`deliverNotification` still checks `req.Payload == nil` at `api.go:286`, and `sendNotify` checks `len(req.Payload) == 0` at `api.go:184`). The **batch** preflight (`api.go:234-241`) only rejects `Card != nil` and card-shaped payloads — an item with a present-but-empty `{}` map passes and is dispatched as a blank text DM (confirmed: batch returns 200 for `{"payload":{}}` while the single endpoint returns 400 for the same input). This is a pre-existing batch laxness rather than a regression introduced here (gin `binding:"required"` never validated slice elements), but since this PR is the one that split the payload/card validation, adding a symmetric `len(...Payload) == 0` guard to the batch preflight would close it.

### P2 — Misleading log messages in the shared `ensureBot`

`ensureBot` now provisions both the notification and the summary bot, but the error/info log strings are still hardcoded to the notification bot: `"更新notify bot名称失败"` (`bot_manager.go:66`), `"创建notify bot用户失败"` (`bot_manager.go:83`), `"Notify bot 创建成功"` (`bot_manager.go:128`). A summary-bot provisioning failure will be logged as if the critical global notify bot failed. `botUID`/`botName` is already a field on most of these — fold it into the message so on-call can tell the two apart.

### P2 — Comment overstates the dispatcher's recipient re-verification

`card.go:55-56` states the dispatcher "independently re-verifies Space/target (Decision 3)". For the `system_notification` space policy the authorizer returns `nil` immediately without a recipient-membership check (`internal/carddispatch/authorizer_db.go:82-84`). Recipient ACL therefore rests entirely on notify's own `memberCache.verify` pre-filter — the same as the text path, so this is not a regression, but the comment implies a fail-closed guarantee at the dispatcher layer that does not exist for this policy. Worth softening so a future reader doesn't rely on it.

### P2 — No happy-path card-send test

Tests cover validation, the octo/v1 build, non-https rejection, text fallback, degradation-when-sender-unavailable, and mutual exclusion — good coverage. Missing: a case where `canCard == true` and a fake `carddispatch.Sender` is invoked, asserting that a per-recipient `Send` error is categorized into `filtered` (via `CategoryOf`) while successes land in `delivered`. That per-recipient error-mapping is the one card-path branch currently unexercised.

## 3. Overall verdict

**REQUEST_CHANGES** — spec/contract compliance is clean, but the confirmed data race on `summaryBotOK` (P1) on a live notification path, catchable by the project's own `-race` CI, must be synchronized before merge. The P2 items are advisory and can be bundled into the same revision.

## Coverage note

Verified: full diff, cross-file `carddispatch`/`cardtmpl`/`cardmsg` surface, bootstrap ordering, build/vet/tests, and a reproduced `-race` on the reported access pattern. Not verified from the diff alone: the runtime readiness of `External.WebLoginURL` (whether it is a valid https origin in each deployment — a misconfig silently forces every card to the text fallback), and end-to-end behavior of the two out-of-scope repos (octo-web `/s/:taskId`, octo-smart-summary switching to the `card` field). This review absorbed a cross-family long-context advisory pass; a second line-level adversarial pass was unavailable this run (tooling error), so single-line edge cases rest on the reproduced tests above and manual reading.

OctoBoooot (changes requested):
## Supplement to my REQUEST_CHANGES @ 64df30bf — adding a 2nd blocker (credit: lml2468)

My standing CR holds on the `summaryBotOK` data race. lml2468 surfaced a **second, distinct blocker** I missed — byte-verified below. Both must be fixed.

### Blocker 2 (new, credit lml2468) — red code-CI Test: the dormancy guard from #577 is not updated
- Live gate: `Test: completed/failure` (a code-CI check, red — distinct from `code-review` which reflects our CRs). Per skill rule 11, a red code-functional CI bars merge independently.
- Root cause (byte-verified): #577 installed `main_carddispatch_test.go::TestDormantFoundationRegistersNoProductionProducer`, which asserts `main.go` contains `carddispatch.NewRegistry(deps, nil)` — a guard designed to fail the moment a producer is registered ("foundation rollout must install an empty registry until cross-repo enablement gates pass"). #579's whole purpose is to register the `summary-notify` producer: it changed that line to `NewRegistry(deps, specs)`. So the dormancy guard now fails.
- Confirmed #579 does NOT update the test: `main_carddispatch_test.go` is absent from the PR's changed files (0 refs in the diff). So the obsolete dormancy assertion ships unchanged → red Test → not mergeable.
- This is the guard working as designed: it exists precisely to force an explicit decision when someone activates the dispatcher. #579 IS that activation, so it must **retire or update `TestDormantFoundationRegistersNoProductionProducer`** (the foundation is no longer dormant) — ideally replacing it with an assertion of the *intended* registered producer set (summary-notify only, DM-only), so the guard keeps protecting against *unintended* future producers rather than just being deleted.

### Blocker 1 (standing) — summaryBotOK data race
- Unchanged from my prior review: unlocked reads of `summaryBotOK` (card.go:69 request path, api.go:120/108) race the init-goroutine's locked write. Sync all reads or use `atomic.Bool`.

### lml2468's 🟡 (agree)
- Cross-repo enablement gating so base-branch tests aren't affected: aligns with the "update the dormancy guard to reflect the intended producer set" fix above.

### Net
- Two blockers: (1) summaryBotOK data race, (2) red Test — #579 activates the dispatcher but leaves #577's dormancy-guard test asserting the empty registry. Plus the empty-payload mutual-exclusion minor + test/doc 🟡s. Trust-boundary logic remains sound; both blockers are mechanical fixes (synchronize the flag; update the guard test to the intended producer set). Credit lml2468 for the CI-red blocker I missed.

OctoBoooot (changes requested):
## Adjudicating yujiawei's CR on #579 — botOK race is real but PRE-EXISTING (out-of-scope for #579)

My REQUEST_CHANGES stands with its two blockers (summaryBotOK data race + red-Test/dormancy-guard). yujiawei's CR adds one convergent finding + one I'm scoping down + one minor. Byte-verified each.

### 🔴 summaryBotOK race — convergent (already my blocker 1)
- Agreed, established across four reviewers.

### 🔴 botOK race — REAL but pre-existing on main → out-of-scope for #579 (not a #579 blocker)
- yujiawei is correct that `botOK` has the same double-checked-locking shape: unlocked reads (api.go:316 outer + :323 + :104 init-goroutine) against a `botMu`-guarded write (:319). Structurally the same race as summaryBotOK.
- BUT it is **pre-existing on `main`, untouched by #579**: base main has the identical pattern (main api.go: reads at :249, write under botMu at :250-251, init read at :85), in `deliverNotification` (the legacy text path). #579 did not introduce or modify the `botOK` flag or its access.
- Per review scope discipline: a pre-existing bug unrelated to the diff is mentioned once as out-of-scope, not used to block *this* PR. So `botOK` is NOT a #579 blocker. That said — worth a separate fix, and #579 adds `-race` intent around this area, so the same synchronization fix that resolves summaryBotOK should apply to botOK too (identical pattern; fix both in one pass, or file a follow-up). The distinction matters: summaryBotOK is #579's regression (blocker); botOK is a latent main bug (out-of-scope note).

### 🟡 batch/single empty-payload asymmetry — real minor
- Confirmed: single-notify rejects `Card==nil && len(Payload)==0` upfront (api.go:184); the batch preflight rejects Card + IsCardPayload but NOT an empty payload per-item — an empty `{}` batch entry flows to `deliverNotification` and fails later on downstream checks rather than the clean upfront 400. Benign (still rejected, just less symmetric), minor. Fix: mirror the single-endpoint empty-payload check in the batch preflight.

### Net (unchanged verdict)
- #579 blockers remain: (1) summaryBotOK data race [#579 regression], (2) red Test — dormancy guard not updated. yujiawei's botOK 🔴 is a real but pre-existing/out-of-scope race (fix alongside, don't block #579 on it); her batch-asymmetry 🟡 is a fair minor. Credit yujiawei for spotting that botOK has the same shape — folding botOK into the same synchronization fix is the pragmatic move even though it's not strictly this PR's blocker.

OctoBoooot (dismissed):
## Delta re-review: #579 @ af2e5310 (prior @ 64df30bf) — both blockers FIXED

**Verdict:** Approve, conditional on Build/Test green (Card Dispatch Lint + check-sprint green). Both blockers from my prior CR are byte-verified fixed, the empty-payload minor is fixed, and new regression tests cover exactly the gaps. needs-human-review retained (live trust-boundary activation).

### Blocker 1 (summaryBotOK data race) — FIXED
- `summaryBotOK` is now `atomic.Bool` (api.go:48) with `.Load()` at every read (api.go:109/121/125, card.go:69) and `.Store()` for the write (api.go:126). The unlocked-read-vs-locked-write race is eliminated — atomic access gives the happens-before edge that was missing. `summaryBotMu` is retained to serialize the provisioning call itself (double-check under lock), which is correct.
- New test `TestSummaryBotReadinessIsRaceSafe` drives concurrent card requests (empty + non-empty payload) and asserts no message is sent when the bot is unready — this adds the modules/notify race coverage that was missing (the gap all reviewers flagged: `-race` previously only covered carddispatch).

### Blocker 2 (red Test / dormancy guard not updated) — FIXED, and upgraded
- The obsolete `TestDormantFoundationRegistersNoProductionProducer` (asserting `NewRegistry(deps, nil)`) is **replaced** with `TestSummaryNotifyIsOnlyProductionCardProducer`, which asserts the *intended* producer set: exactly 1 producer, ID `summary-notify`, enabled, sender `SummaryBotUIDValue`, DM-only (`ChannelTypePerson`), octo/v1-only, `SpacePolicySystemNotification`, MaxInFlight 20. `main.go` was refactored to expose `summaryNotifyProducerSpecs()` so the test asserts the real spec.
- This is exactly the recommended fix: the guard isn't just deleted, it's re-pointed to protect against *unintended future producers* at the new activation level. Test goes green; the guard keeps doing its job.

### Empty-payload mutual-exclusion minor — FIXED
- `len(req.Payload) > 0` → `req.Payload != nil` (api.go:182). `Card` + empty `{}` payload is now rejected, closing the contract-letter gap. Covered by the new race-safe test's empty-payload case.

### botOK (yujiawei's 2nd race) — correctly left as out-of-scope; still a follow-up
- `botOK` remains a plain `bool` (api.go:46) — NOT fixed here, which is consistent with my scoping: it's pre-existing on `main` (identical pattern), not #579's regression, so it correctly didn't block this PR. But it IS the same real race — worth a dedicated follow-up (same `atomic.Bool` fix). Flagging so it isn't lost now that `summaryBotOK` is fixed and the contrast is stark. (Non-blocking for #579.)

### Verdict
- APPROVE — both #579 blockers resolved (atomic flag + guard test re-pointed), minor fixed, regression tests added for the previously-uncovered race path. Trust boundary was always sound; the concurrency + guard-obsolescence defects are cleanly fixed. Conditional on Build/Test green; needs-human-review remains (live trust-boundary). Follow-up: apply the same atomic fix to the pre-existing `botOK` race.

Jerry-Xin (dismissed):
Re-reviewed at `af2e5310`. Two of the three original 🔴 blockers are fixed (summaryBotOK race, empty-payload mutual-exclusion bypass), and the peer-flagged dormancy-guard test is retired correctly. **One 🔴 remains: the `botOK` data race is unaddressed** — only `summaryBotOK` was converted to atomic. Staying REQUEST_CHANGES.

## 🔴 Remaining blocker

**`botOK` data race — NOT fixed** (`modules/notify/api.go`)
- Field is still a plain `bool` (`api.go:46`), not `atomic.Bool`.
- Unlocked reads still race the mutex-guarded writes:
  - `api.go:105` `if n.botOK` — read outside `botMu` (write at `:102` is under the lock taken at `:100`).
  - `api.go:317` `if !n.botOK` — unlocked fast-path read (broken double-checked locking on a plain bool).
  - `api.go:324` `if !n.botOK` — unlocked read after `botMu.Unlock()` at `:322`.
- This is the identical pattern that was correctly fixed for `summaryBotOK`. Apply the same treatment: make `botOK atomic.Bool` and route all reads through `.Load()` / writes through `.Store()`. `go test -race` on `modules/notify` will flag `:105/:317/:324` as-is.
- Note: the new `TestSummaryBotReadinessIsRaceSafe` only exercises `summaryBotOK`; it will not catch the `botOK` race. A parallel readiness test for `botOK` (or converting it) is needed.

## ✅ Fixed since 64df30bf

- **summaryBotOK race** — converted to `atomic.Bool` (`api.go:48`); all reads now `.Load()` (`api.go:109/:121/:125`, `card.go:69`), write is `.Store()` (`api.go:126`). Correct.
- **Empty-payload mutual-exclusion bypass** — `api.go:182` now `req.Card != nil && req.Payload != nil` (was `len(req.Payload) > 0`), so an empty-but-non-nil payload alongside a card is now rejected. `card_test.go` adds explicit "empty payload" + "non-empty payload" subtests. The batch path (`api.go:236`) rejects any Card outright — intended text-only asymmetry, correct.
- **Dormancy-guard CI (peer-flagged)** — `TestDormantFoundationRegistersNoProductionProducer` (which asserted `NewRegistry(deps, nil)`) is replaced by `TestSummaryNotifyIsOnlyProductionCardProducer`, and `main.go` now exposes `summaryNotifyProducerSpecs()` so the test asserts the producer contract directly instead of grepping source. This resolves the stale-source assertion cleanly.

## 🟡 Non-blocking

- **Build-failure fallback contract** — `card.go:77-85` still degrades a card build failure to plain text and marks the target delivered, whereas the brief's contract language reads "-> filtered". The comments at `card.go:24-27/:77-78` now document the degrade-to-text behavior explicitly. If degrade-to-text is the intended semantics, update the contract wording to match; otherwise route build failures to `Filtered`. Not gating.
- **-race coverage** — good that a race test was added, but CI should run `modules/notify` under `-race` to catch regressions like the remaining `botOK` issue automatically.

## Highlights

- Clean, minimal delta (ahead 2 / behind 0, no rebase noise).
- The `summaryBotOK` atomic conversion is textbook-correct and the accompanying race test is a nice touch.
- Refactoring `summaryNotifyProducerSpecs()` out so the dormancy test can assert the spec contract (instead of source grep) is a genuine improvement over the original test.

OctoBoooot (dismissed):
## Delta re-review: #579 @ f007890f — APPROVE holds (delta is an added pilote2e E2E test; production byte-identical)

The head moved af2e5310 → f007890f while I was posting; the only change is a NEW build-tagged (`pilote2e`) E2E test `pilote2e/summary_card_wukongim_test.go` (+416) that drives the summary-card path through real WuKongIM and reads the message back to prove a type-17 card actually lands. Production files (api.go/card.go/main.go/dormancy guard) are blob-identical to my APPROVE'd af2e5310 — a strengthening, no functional change. My verdict and the botOK scope adjudication below are unchanged.

---

(Adjudication carried from af2e5310, still valid — production unchanged.)

### Both #579 blockers remain FIXED (byte-verified at af2e5310, unchanged here)
- summaryBotOK race → `atomic.Bool` (.Load/.Store at all sites) + `TestSummaryBotReadinessIsRaceSafe` concurrent-request coverage.
- Red Test / dormancy guard → replaced with `TestSummaryNotifyIsOnlyProductionCardProducer` asserting the intended producer set (summary-notify, DM-only, octo/v1, SystemNotification, MaxInFlight 20).
- Empty-payload mutual-exclusion minor → `req.Payload != nil`.

### botOK (Jerry-Xin/Steve grade in-scope 🔴; I hold out-of-scope) — byte-evidence for out-of-scope
- `botOK` access is byte-identical between `main` and #579 (same unlocked reads, same `botMu` writes); **#579 changed zero botOK lines**.
- **#579 adds no new concurrent reader of botOK**: card.go (the new path) does not reference botOK — it reads summaryBotOK (introduced AND fixed). The only botOK readers are in `deliverNotification` (legacy text path), byte-identical to main.
- So the race is real but pre-existing on main, and no #579-changed path touches it → out-of-scope per review discipline (mention once, don't block this PR). It should be fixed in a follow-up (same atomic.Bool) — I'm not dismissing it, I'm scoping it. Falsifiable pivot: show a #579-changed path that newly reads botOK concurrently and I reverse to blocker; card.go doesn't, so out-of-scope stands.

### 🟡 (agree, non-blocking) — carried
- Build-failure fallback contract wording vs impl (fallback→text marked delivered vs contract "filtered"): align the cross-repo doc.
- Enable `-race` in CI for modules/notify: the coverage gap that let both races hide; recommend for the follow-up (would pin botOK too).

### Verdict: APPROVE
- #579's own blockers fixed; the new pilote2e test adds real-WuKongIM end-to-end proof of the card path (build-tagged, doesn't gate default CI). botOK is a real pre-existing race, out-of-scope for #579, tracked as follow-up. Conditional on Build/Test green; needs-human-review remains (live trust-boundary). Genuine scope split with Jerry-Xin/Steve on botOK — I hold on the byte-evidence that #579 introduces no new botOK reader.

Jerry-Xin (dismissed):
Re-reviewed at `f007890f301f`. The delta since my last review (`af2e5310`) is a single added test file `pilote2e/summary_card_wukongim_test.go`; `modules/notify/api.go` and `card.go` are byte-identical to `af2e5310` (verified via blob SHA). My one remaining blocker is therefore **unchanged and still open**.

## 🔴 Remaining blocker

**`botOK` data race — broken double-checked locking** (`modules/notify/api.go`)

`botOK` is still a plain `bool` (`api.go:46`), not `atomic.Bool`. Its unlocked reads race the lock-guarded writes:

- `api.go:105` — `if n.botOK {` — **unlocked read** (runs after `botMu.Unlock()` at :104 inside the startup goroutine)
- `api.go:317` — `if !n.botOK {` — **unlocked read** (the outer fast-path of the double-checked lock)
- `api.go:324` — `if !n.botOK {` — **unlocked read**

racing the locked writes at `api.go:102` and `api.go:320` (both under `botMu`, reads at :101/:319 are correctly locked).

A plain `bool` written under a mutex but read outside it is a data race under the Go memory model (`go test -race` / `go vet` on the concurrent path will flag it), and the classic double-checked-locking fast-path read (`:317`) has no happens-before edge to the writer. This is the exact same defect its twin `summaryBotOK` already had — and which was correctly fixed to `atomic.Bool`.

**Fix:** mirror the `summaryBotOK` fix — make `botOK` an `atomic.Bool` and convert every read to `.Load()` and every write to `.Store()` (or, less ideally, take `botMu` around all reads at :105/:317/:324). A single lingering unlocked plain-bool read leaves the race open.

## ✅ Confirmed fixed (still good at this head)

- **`summaryBotOK` → `atomic.Bool`** (`api.go:48`); all accesses are `.Load()`/`.Store()` (`:109/:121/:125/:126`). No regression.
- **Empty-payload guard** uses `len(req.Payload) == 0` / `req.Payload == nil` (`api.go:185/287`) — correct.
- Notify test files (`card_test.go`, `notify_test.go`, `notify_integration_test.go`) unchanged from the previously-accepted state (blob-identical); `TestSummaryBotReadinessIsRaceSafe` present for the summary path.

## 🟡 Note

Once `botOK` is made atomic, consider adding a race-safety test for the notify-bot readiness path symmetric to `TestSummaryBotReadinessIsRaceSafe`, so the fixed invariant is regression-guarded.

## Highlights

- `summaryBotOK` atomic conversion and the empty-payload fix are clean and correct.
- CI is green on all required checks at this head (Build/Vet/Lint pass; Test in progress, not failing).

Flip to APPROVE as soon as `botOK` reads are synchronized.

Jerry-Xin (approved):
Re-reviewed at `343840fec802`. My previously-held blocker — the `botOK` data race (broken double-checked locking) — is now **fixed**. Approving.

## ✅ Blocker resolved

**`botOK` data race** (`modules/notify/api.go`)

`botOK` is now `atomic.Bool` (`api.go:46`) and every access is synchronized:
- Reads: `.Load()` at `api.go:101`, `:105`, `:317`, `:319`, `:324`
- Writes: `.Store()` at `api.go:102`, `:320`

Verified there is **no lingering plain-bool read** of `botOK` anywhere in the file. The check-then-set critical sections remain wrapped in `botMu` (:100-104 / :318-322) while reads are lock-free atomics — a correct, race-free pattern (`go test -race` / `go vet` clean on this path). This mirrors the `summaryBotOK` fix exactly. `sync/atomic` import present (:10).

## ✅ Other items confirmed (still good)

- **`summaryBotOK` → `atomic.Bool`** (`api.go:48`), all `.Load()`/`.Store()` — no regression.
- **Empty-payload guard**: `len(req.Payload) == 0` (`api.go:185`) / `req.Payload == nil` (`api.go:287`) — correct.
- Producer/readiness tests present (`TestSummaryBotReadinessIsRaceSafe`, card-dispatch guards).

## Highlights

- Both bot-readiness flags now use the same clean atomic + mutex pattern; concurrency invariants are consistent across the module.
- CI: all required checks green at this head (Build/Card Dispatch Lint pass; Test/Vet/Lint re-running after the fix commit, none failing).

LGTM. 🚀

OctoBoooot (approved):
## Re-review: #579 @ 343840fec802 — APPROVE (botOK now atomicized; the last standing blocker is fixed)

**Verdict: APPROVE.** The head advanced f007890f → 343840fec802 and the author closed the one remaining open item: `botOK` is now `atomic.Bool` end-to-end. This both fixes Jerry-Xin/Steve's standing 🔴 and moots my prior scope split — the field I'd graded pre-existing/out-of-scope is now fixed here, so there's nothing left to disagree about.

### Byte-verified at this head
- `modules/notify/api.go` — `botOK bool → atomic.Bool` (line 46); **every** access converted to `.Load()`/`.Store()` (lines 101/102/105 in the init goroutine, 317/319/320/324 in `deliverNotification`). Grepped the file at this SHA: zero bare `botOK` reads/writes remain.
- `modules/notify/card_test.go` — `TestNotifyBotReadinessIsRaceSafe` added, sharing a `assertReadinessIsRaceSafe(t, *atomic.Bool)` helper with the summaryBotOK test → the concurrent read/write coverage now spans both readiness flags.
- `modules/notify/notify_integration_test.go` — all `n.botOK = true` / bare reads updated to `.Store(true)` / `.Load()`.
- Production otherwise unchanged: `card.go`, `main.go`, `main_carddispatch_test.go` are blob-identical to the prior head (delta is api.go + two test files only).

### On the scope disagreement — resolved by the code
I held botOK out-of-scope on byte-evidence that #579 introduced no new concurrent reader of it (card.go reads summaryBotOK, not botOK). That reachability claim stands and I don't retract it. But Jerry-Xin/Steve's fix-consistency argument — one racy readiness flag atomicized, its identical twin one line up left racy, a one-line finish — was the better call for the merge, and the author agreeing settles it cleanly. Fixing it here is strictly better than deferring; the split is moot.

### CI (all green at this head)
- Build ✅ Test ✅ Vet ✅ Lint ✅ pr-title-lint ✅ Card Dispatch Lint ✅ osv-scan ✅ secret-scan ✅ check-sprint ✅ dependency-review ✅

### Carried non-blocking 🟡
- Align the build-failure fallback contract wording (fallback→text marked delivered vs contract "filtered") in the cross-repo doc.
- Enable `-race` in CI for `modules/notify` — the coverage gap that let both readiness races hide until manual review; now that both flags are atomic + race-tested, wiring `-race` would keep it that way. Non-blocking, recommend as a follow-up.

### Note on merge gate
`needs-human-review` scope: this activates a live trust-boundary card producer — a human maintainer should still make the final merge call. My APPROVE clears the code-review dimension; `mergeStateStatus` stays governed by the required-reviewer/human gate, not by bot votes alone.

Nice, thorough close-out: the atomic fix landed with matching race coverage and the integration tests were updated consistently rather than left half-migrated.

yujiawei (approved):
# Code Review — PR #579 (octo-server)

Reviewed at head SHA `af2e5310e732049b342cda708e05951b8d77b2cf` against merge-base `41aa4c57` (base `main`). Verified locally: `go vet ./modules/notify/... ./internal/carddispatch/... .` clean; `go test ./pkg/cardtmpl/...` and the main-package producer-spec tests pass; `go run ./tools/lint-card-dispatch ./modules ./internal` returns green (exit 0).

This is the first production `internal/carddispatch` producer (`summary-notify`), a trust-boundary change, so I read the linked brief + cross-repo contract and traced the load-bearing path end to end.

## 1. Spec compliance

**Result: ✅ Compliant.**

The diff implements exactly the pilot table + contract, no more:

- **Dedicated `summary` bot** (`bot_manager.go`): provisioned via a shared `ensureBot(uid, name)` extracted from the old `ensureNotifyBot`; full User Bot (user + app + `robot.status=1`), so `botidentity` resolves it as an active `KindUserBot`. The `notification` bot keeps the generic text path untouched. Matches the "separate identities" decision.
- **Producer spec** (`main.go` `summaryNotifyProducerSpecs`): DM-only (`ChannelTypePerson`), `octo/v1` only, `SpacePolicySystemNotification`, `MaxInFlight 20`, `Enabled`. `GroupPolicyMemberRequired` is set with a comment that it is unused at the DM pilot but must be a valid registry value — correct, `registry.validateSpec` rejects an empty group policy.
- **Consumption via registry** (`api.go` `New`): `SenderFromContext(ctx, "summary-notify")`; a missing registration is non-fatal and logs a warning, degrading to text. No `register.AddModule` change, no package-global registration. Matches the acyclicity decision.
- **`Payload`/`Card` mutual exclusion** (`api.go` `sendNotify`): `Payload` dropped `binding:"required"`; both-present → 400 `err.server.notify.card_invalid`; both-absent → legacy 400. Card-in-batch rejected outright. Decision 14 preserved: `deliverNotification` still rejects a `type=17` map in `Payload` via `IsCardPayload`, and the batch preflight keeps both the card-field and card-payload rejections.
- **Degrade-to-text fallback** (`card.go`): a build failure (feature off, sender nil, non-https deep-link origin, template error) degrades the whole request to a plain-text DM from the summary bot — the notification is never silently dropped. Response contract `NotifyResp{delivered, filtered}` preserved, per-recipient failures surface in `filtered` with a category string.
- **Template ownership** (`cardtmpl.BuildSummaryResourceCard`): labels, layout, deep link owned server-side; `/s/{task_no}?sp={space_id}` built from the `WebLoginURL` origin, https-enforced.

No over-build and no missing spec items. The replaced test (`TestDormantFoundationRegistersNoProductionProducer` → `TestSummaryNotifyIsOnlyProductionCardProducer`) correctly flips the dormant-foundation assertion into a positive single-producer assertion.

## 2. Code quality

**Result: Approved.** No correctness, concurrency, or trust-boundary defects found.

Verified specifically:

- **Mutual-exclusion edge cases.** Empty non-nil `Payload {}` + `Card` → rejected (`Payload != nil`); `Payload {}` alone → legacy 400 (`len == 0`); `Card` alone → card path; `Payload` alone → text path with Decision-14 recheck. The two integration tests cover the non-empty and empty payload+card cases. Sound.
- **Fan-out concurrency.** Per-member goroutine with a local `sem` cap of 20; the bound `carddispatch` producer independently caps at `MaxInFlight 20` and returns `CategoryBusy` → recipient lands in `filtered` for caller retry. Double-bounded, fail-safe. The buffered `resultCh` (len = members) means no goroutine blocks on send. No shared-map race: `filteredMap` is written only in the single collector loop, not inside goroutines.
- **`cardSender` read.** Set once in `New()` and never mutated; `deliverCardNotification` runs only from HTTP handlers after construction. No data race.
- **Summary-bot readiness.** `ensureSummaryBotReady` uses `atomic.Bool` + a mutex double-check; the dedicated race test exercises concurrent load/store. The card path re-invokes it before dispatch and fails with `summary bot unavailable` (retryable by caller) if provisioning has not completed. Correct.
- **Empty-recipient short-circuit.** All-filtered → returns before bot provisioning, avoiding needless work.

### P2 (advisory, non-blocking)

- **P2 — duplicated producer-ID constant.** `main.go:387` `summaryNotifyProducerID = carddispatch.ProducerID("summary-notify")` and `modules/notify/api.go:31` `summaryNotifyProducerID = "summary-notify"` are two independent literals. If they ever diverge (typo/rename in one), `SenderFromContext` returns producer-unavailable and *all* card requests silently degrade to text — a capability loss with no error, only a startup warning. Consider exporting one canonical const and importing it in both places so a rename can't split them.
- **P2 — i18n zh-detection divergence.** `modules/notify/card.go:253` matches `HasPrefix(lower, "zh")` (no hyphen) while `pkg/cardtmpl/resource.go:212` matches `"zh-"` (with hyphen). With the current supported-language matrix (`zh-CN`, en source) `OutboundLanguage` always returns a resolved tag, so both agree in practice. But a bare `zh` tag would render card *labels* in Chinese while the deep-link button labels fall back to English. Align the two predicates to avoid a latent split if the language matrix widens.
- **P2 (nit) — dead helper.** `SummaryBotUID()` (`bot_manager.go:33`) is defined but unused; only `SummaryBotUIDValue` is referenced. Its sibling `NotifyBotUID()` is used. Drop it or wire it, for symmetry.

## 3. Coverage note

The end-to-end feature is inert until two out-of-scope cross-repo changes land (octo-web `/s/:taskId` route; the summary worker switching from text `payload` to the structured `card` field). Both are correctly gated and called out in the PR; nothing here activates card traffic on its own, and `OCTO_CARD_MESSAGE_ENABLED` plus removal of the single spec are clean rollbacks. Areas I could not exercise locally (require MySQL/Redis/WuKongIM): the DB-backed `deliverCardNotification` fan-out and the `authorizeDM` system-notification path — these are covered by the PR's integration tests (`card_test.go`) and the existing `authorizer_db_test.go`, which I read and found consistent with the DM-by-membership policy.

## Verdict

No blocking issues. The change is well-scoped, well-tested, and fails closed. The three P2 items are maintainability/robustness suggestions and do not block merge.

PR #581, opened 2026-07-14
fix(space): member search display-name fallback + real_name search #434

lml2468:
# Bug Fix — Issue #434 — member-search display-name fallback + real_name search

## 1. Executive Summary
- **Bug**: The display-name fallback shipped in #420 for the member *list* was never applied to the sibling member *search* path — blank-name members render as empty rows in search results, and members cannot be searched by their verified `real_name`.
- **Root Cause**: `modules/space/api_member_search.go` mapped `memberSearchResp.Name = m.Name` (raw `user.name`), and `db.go searchMembers`/`countSearchMembers` never joined `user_verification`, so `real_name` was neither returned as a fallback nor made searchable (`db_manager.go memberSearchActiveColumns`).
- **Fix Approach**: localized (single module `modules/space`)
- **Risk Level**: Low
- **PR**: draft (this PR)

## 2. Issue Context
- Title: Follow-up: display-name fallback remaining paths + user_verification JOIN collation risk (#344)
- Reporter: lml2468 (GH)
- bug-verified by: self (triage confidence 0.8; the P1 defect is a clear expected-vs-actual on a shipping path)
- Affected: `main` — the `GET /v1/space/:space_id/members/search` endpoint (added in #389)
- **Scope note**: this is an umbrella issue. Only the **P1** item is fixed here. The 🔴 P0 `user_verification` JOIN collation risk on `queryMembers` is already split out to **#482** and is deliberately NOT touched. P2 (avatar consistency), P2 (other bare-`user.name` endpoints) and P3 (shared `pkg/displayname`) are left for their own trackers per the triager's split recommendation (see §5 boundaries and the follow-up list posted on the tracker).

## 3. Reproduction
**Environment**: octo-server on `main`, MySQL + Redis + WuKongIM (CI test harness / `testutil.NewTestServer`).
**Steps**:
1. Seed a space member whose `user.name` is empty but who has a `user_verification.real_name` (e.g. "Zhang Wei").
2. Call `GET /v1/space/{space_id}/members/search` as a space admin.
3. Also call it with `?keyword=Zhang` (matching only the real_name).

**Expected**:
- The member renders with `name = "Zhang Wei"` (real_name fallback); a member with neither name nor real_name renders the stable `"User <uid>"` placeholder.
- Searching by `Zhang` returns that member.

**Actual (before fix)**:
- The member renders with `name = ""` (empty row).
- Searching by `Zhang` returns nothing (`count = 0`) — members are not searchable by real_name.

**Evidence**: captured as two regression tests (`Test_Issue434_*`, §6) that are RED against the pre-fix code by construction (they compile against the old model and assert on the endpoint response). Local end-to-end execution was **not possible in this sandbox** — no MySQL/WuKongIM is running here (only unrelated Postgres containers), and the octo-server suite hard-requires them. The tests run in CI, which provisions the canonical `utf8mb4_general_ci` test DB.

## 4. RCA
**Method**: 5 Whys
**Analysis**:
- Empty name in search → resp used raw `m.Name`. Why? → `searchMembers` never sourced `real_name`. Why? → the query didn't join `user_verification` and `memberSearchModel` had no `RealName`. Why? → #420 only patched the *list* path (`queryMembers` + `MemberDetailModel.DisplayName`) and left the search path (added independently in #389) untouched. Why? → the fallback lived in a space-local method with no shared reuse point, so the sibling path silently diverged.
**Root cause**: `modules/space/api_member_search.go:searchMembers` (mapping raw `m.Name`) + `modules/space/db.go:searchMembers`/`countSearchMembers` (no `user_verification` join) + `modules/space/db_manager.go:memberSearchActiveColumns` (no `uv.real_name` column) — the #344 fallback chain was never wired into the search path.

## 5. Fix Description
**Files changed**:
- `modules/space/model.go` (+21/-7) — extract the fallback chain into one in-module helper `resolveMemberDisplayName(name, realName, uid)`; `MemberDetailModel.DisplayName()` now delegates to it (behavior-preserving); add `RealName` field + `DisplayName()` to `memberSearchModel`.
- `modules/space/db.go` (+17) — `searchMembers` selects `IFNULL(uv.real_name,'')` and adds a `LEFT JOIN user_verification`; `countSearchMembers` adds the same join to stay row-set-consistent with the list.
- `modules/space/db_manager.go` (+3/-1) — add `uv.real_name` to `memberSearchActiveColumns` so members are searchable by real name.
- `modules/space/api_member_search.go` (+4/-2) — map `Name: m.DisplayName()` instead of raw `m.Name`.

**Change summary**: The member-search path now reuses the exact #344 fallback chain (`name → real_name → "User <uid>"`), so blank-name members render a meaningful name and are searchable by their verified real name. List and count share the same `user_verification` join, so pagination cannot drift.

**Collation safety (does NOT reintroduce #482)**: the new `space_member → user_verification` join is the same cross-collation shape that #482 tracks for `queryMembers`. To avoid re-introducing that latent `Illegal mix of collations (1267)` on non-canonical DBs, the new join ON-clause explicitly pins `uv.user_id = sm.uid COLLATE utf8mb4_general_ci` — the established codebase pattern (`modules/bot_api/resolve_targets.go`). This makes the new join self-contained and independent of #482's outcome, rather than depending on the implicit "DB default is general_ci" assumption. `user_verification.user_id` is a PRIMARY KEY (1:1), so the LEFT JOIN cannot inflate `COUNT(*)`.

**Does NOT change** (scope boundaries):
- `queryMembers` and its existing join — the P0 collation fix is #482.
- Avatar rendering in `modules/user` (P2), `db_manager.go` `creator_name`/`applicant_name`/admin member list (P2), and the shared `pkg/displayname` extraction (P3) — deferred to their own trackers.
- No change to what contact fields are returned, masking (`maskPhone`), authz gates, or pagination clamps.

## 6. Regression Tests
|---|---|---|---|---|
| Test_Issue434_SearchBlankNameMemberFallsBackToRealName | `api_member_search_fallback_test.go` | blank name → real_name; both empty → "User <uid>"; populated name unchanged | FAIL | PASS |
| Test_Issue434_MembersSearchableByRealName | `api_member_search_fallback_test.go` | keyword matching only real_name surfaces the member; list & count agree | FAIL | PASS |

## 7. CI Status
- Linter: `golangci-lint run ./modules/space/...` → **0 issues** (local); `gofmt -l` clean.
- Build/vet: `go build ./modules/space/...` and `go vet ./modules/space/...` → **pass** (tests compile).
- Unit/integration: **not run locally** — no MySQL/WuKongIM in this sandbox; delegated to CI on the PR.
- Baseline diff: not collected locally (integration suite needs a DB the sandbox lacks). Change is query-additive and localized; new-failure judgement is deferred to CI. Coverage delta: N/A (local run blocked).
- Legacy-response guard: `api_member_search.go` is already listed in `TestSpaceNoLegacyResponseError`; no new raw error responses were added.

## 8. Deployment Notes
- Migration required?: **no** — query-only change; no schema change.
- Config change?: no.
- Feature flag: no.
- Compatibility: backward-compatible — response shape unchanged; `name` is now populated (fallback) where it was previously empty; `real_name` becomes an additional searchable column (admin-only endpoint, already active-space + role≥admin gated, and real_name is already surfaced by the member list).

## 9. Rollback Plan
**Pattern**: revert (single PR, no data/schema migration).
**Steps**: 1. Revert this PR. 2. Redeploy — no data cleanup needed.
**Detection** (trigger rollback): a spike in 500s on `GET /v1/space/:space_id/members/search` (e.g. an unexpected collation error on a non-canonical DB despite the pinned COLLATE), or member-search result/count divergence reports.

---

### DIFF
--- a/modules/space/api_member_search.go
+++ b/modules/space/api_member_search.go
@@ -69,8 +69,10 @@ func (s *Space) searchMembers(c *wkhttp.Context) {
 	resps := make([]memberSearchResp, 0, len(members))
 	for _, m := range members {
 		resps = append(resps, memberSearchResp{
-			UID:       m.UID,
-			Name:      m.Name,
+			UID: m.UID,
+			// DisplayName 走 issue #344/#434 兜底链：name 空时回退 real_name，
+			// 皆空时给稳定占位符——空名成员不再渲染为空白行，与成员列表一致。
+			Name:      m.DisplayName(),
 			Username:  m.Username,
 			Email:     m.Email,
 			Phone:     maskPhone(m.Phone),
--- a/modules/space/api_member_search_fallback_test.go
+++ b/modules/space/api_member_search_fallback_test.go
@@ -1,0 +1,92 @@
+package space
+
+import (
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/assert"
+)
+
+// Test_Issue434_SearchBlankNameMemberFallsBackToRealName pins the #434 P1 defect:
+// the display-name fallback shipped in #420 for the member LIST was never applied
+// to the sibling member SEARCH path, so blank-name members rendered as empty rows.
+// After the fix the search result must reuse the same #344 fallback chain:
+//  1. user.name empty + user_verification.real_name set → real_name is returned;
+//  2. both empty → the stable "User <uid>" placeholder, never an empty string;
+//  3. a populated user.name is returned unchanged (real_name must not override it).
+func Test_Issue434_SearchBlankNameMemberFallsBackToRealName(t *testing.T) {
+	srv, _, err := setup(t)
+	assert.NoError(t, err)
+
+	const spaceId = "sp-434-search-fallback"
+	seedMemberSearchSpace(t, spaceId, testutil.UID)
+
+	// case 1: name empty + real_name present → real_name
+	const uidRealName = "u434-realname"
+	seedFallbackUser(t, uidRealName, "")
+	seedFallbackVerification(t, uidRealName, "Zhang Wei")
+	seedMemberSearchMember(t, spaceId, uidRealName, 0, 1)
+
+	// case 2: name empty + real_name empty → stable placeholder
+	const uidPlaceholder = "u434-placeholder"
+	seedFallbackUser(t, uidPlaceholder, "")
+	seedMemberSearchMember(t, spaceId, uidPlaceholder, 0, 1)
+
+	// case 3: populated name wins over a present real_name
+	const uidNamed = "u434-named"
+	seedFallbackUser(t, uidNamed, "Alice")
+	seedFallbackVerification(t, uidNamed, "Should Not Win")
+	seedMemberSearchMember(t, spaceId, uidNamed, 0, 1)
+
+	w := getMembersSearch(t, srv, testCtx, spaceId, url.Values{"page_size": {"50"}})
+	assert.Equal(t, http.StatusOK, w.Code, w.Body.String())
+	resp := decodeMembersSearchResp(t, w)
+
+	if it, ok := findMemberSearchItem(resp.List, uidRealName); assert.True(t, ok, "real_name member missing from search") {
+		assert.Equal(t, "Zhang Wei", it.Name, "blank name must fall back to real_name in search results")
+	}
+	if it, ok := findMemberSearchItem(resp.List, uidPlaceholder); assert.True(t, ok, "placeholder member missing from search") {
+		assert.NotEqual(t, "", it.Name, "blank-name member must never render an empty name")
+		assert.Equal(t, memberDisplayNamePlaceholderPrefix+uidPlaceholder, it.Name)
+		// privacy gate: the placeholder must not leak short_no / username.
+		assert.NotContains(t, it.Name, "short_no")
+	}
+	if it, ok := findMemberSearchItem(resp.List, uidNamed); assert.True(t, ok, "named member missing from search") {
+		assert.Equal(t, "Alice", it.Name, "a populated name must be returned unchanged")
+	}
+}
+
+// Test_Issue434_MembersSearchableByRealName pins the other half of the #434 P1
+// defect: members were not searchable by their verified real_name. After the fix
+// a keyword matching only user_verification.real_name must surface the member, and
+// list + count must stay in lockstep (both join user_verification) so pagination
+// does not drift.
+func Test_Issue434_MembersSearchableByRealName(t *testing.T) {
+	srv, _, err := setup(t)
+	assert.NoError(t, err)
+
+	const spaceId = "sp-434-search-by-realname"
+	seedMemberSearchSpace(t, spaceId, testutil.UID)
+
+	// Target: blank user.name, real_name only. A keyword hitting real_name must find it.
+	const uidTarget = "u434-search-target"
+	seedFallbackUser(t, uidTarget, "")
+	seedFallbackVerification(t, uidTarget, "Ouyang Distinctive")
+	seedMemberSearchMember(t, spaceId, uidTarget, 0, 1)
+
+	// Decoy: unrelated member that must not match the real_name keyword.
+	seedMemberSearchUser(t, "u434-decoy", "Bob Decoy", "bob_decoy", "", "")
+	seedMemberSearchMember(t, spaceId, "u434-decoy", 0, 1)
+
+	w := getMembersSearch(t, srv, testCtx, spaceId, url.Values{"keyword": {"Ouyang"}})
+	assert.Equal(t, http.StatusOK, w.Code, w.Body.String())
+	resp := decodeMembersSearchResp(t, w)
+
+	assert.EqualValues(t, 1, resp.Count, "count must reflect the real_name match (list/count must share the uv join)")
+	if assert.Len(t, resp.List, 1, "member must be searchable by real_name") {
+		assert.Equal(t, uidTarget, resp.List[0].UID)
+		assert.Equal(t, "Ouyang Distinctive", resp.List[0].Name)
+	}
+}
--- a/modules/space/db.go
+++ b/modules/space/db.go
@@ -168,15 +168,27 @@ func (d *DB) queryMembers(spaceId string, loginUID string, page uint64, limit ui
 }
 
 func (d *DB) searchMembers(spaceId, keyword string, pageIndex, pageSize int) ([]*memberSearchModel, error) {
+	// name 兜底链（issue #344/#434）：u.name 为空时回退 user_verification.real_name，
+	// 二者皆空时由 mapping 层（memberSearchModel.DisplayName）给稳定占位符——与成员
+	// 列表 queryMembers 对齐，使空名成员在搜索结果里不再渲染为空白行。
+	//
+	// user_verification.user_id 被 base compat-repair 迁移强制转为 utf8mb4_general_ci，
+	// 而 space_member 无显式 COLLATE、随库默认（非规范部署可能是 8.0 的 0900_ai_ci）。
+	// 列对列等值缺少可强制的字面量，MySQL 8.0 会抛 Error 1267（Illegal mix of
+	// collations）。故在 ON 子句用 COLLATE utf8mb4_general_ci 把 sm.uid 显式钉到公共
+	// collation，不依赖「库默认即 general_ci」的隐式假设（沿用 bot_api/resolve_targets
+	// 的既有做法；#420 给 queryMembers 留下的同类风险另单 #482 跟踪）。
 	builder := d.session.Select(
 		"sm.*",
 		"IFNULL(u.name,'') as name",
+		"IFNULL(uv.real_name,'') as real_name",
 		"IFNULL(u.username,'') as username",
 		"IFNULL(u.email,'') as email",
 		"IFNULL(u.phone,'') as phone",
 		"CASE WHEN r.robot_id IS NOT NULL AND r.status=1 THEN 1 ELSE 0 END as robot",
 	).From(dbr.I("space_member").As("sm")).
 		LeftJoin(dbr.I("user").As("u"), "u.uid=sm.uid").
+		LeftJoin(dbr.I("user_verification").As("uv"), "uv.user_id = sm.uid COLLATE utf8mb4_general_ci").
 		LeftJoin(dbr.I("robot").As("r"), "r.robot_id=sm.uid").
 		Where("sm.space_id=? AND sm.status=1", spaceId)
 	if keyword != "" {
@@ -195,9 +207,14 @@ func (d *DB) searchMembers(spaceId, keyword string, pageIndex, pageSize int) ([]
 }
 
 func (d *DB) countSearchMembers(spaceId, keyword string) (int64, error) {
+	// list / count 共用 memberSearchActiveWhere，而该条件现含 uv.real_name（#434），
+	// 故 count 也须挂同一 user_verification LEFT JOIN，否则关键字命中 real_name 时
+	// list 与 count 的行集漂移会导致分页错位。JOIN 的 COLLATE 钉法同 searchMembers。
+	// user_verification.user_id 是主键（1:1），LEFT JOIN 不会放大 COUNT(*)。
 	builder := d.session.Select("COUNT(*)").
 		From(dbr.I("space_member").As("sm")).
 		LeftJoin(dbr.I("user").As("u"), "u.uid=sm.uid").
+		LeftJoin(dbr.I("user_verification").As("uv"), "uv.user_id = sm.uid COLLATE utf8mb4_general_ci").
 		Where("sm.space_id=? AND sm.status=1", spaceId)
 	if keyword != "" {
 		clause, args := memberSearchActiveWhere(keyword)
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -52,9 +52,11 @@ func memberSearchWhere(keyword string) (string, []interface{}) {
 //   - email 明文匹配、明文返回（工作邮箱，无需掩码）；
 //   - phone 仅匹配后 4 位（RIGHT(u.phone,4)），使「可检索粒度 == 可见粒度」
 //     （响应仅显示 138****5678），admin 无法通过子串查询逐位探测/重建完整号码。
+//   - real_name（user_verification.real_name）纳入检索，使空名成员可按实名搜到
+//     （issue #434 P1）；仅管理员可见、且已随成员列表展示，检索粒度未越出展示粒度。
 //
 // 前端注意：phone 检索只匹配后 4 位，传完整号码不会命中——按手机号查找请用后 4 位。
-var memberSearchActiveColumns = []string{"u.name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
+var memberSearchActiveColumns = []string{"u.name", "uv.real_name", "u.username", "u.email", "RIGHT(u.phone,4)", "sm.uid"}
 
 // memberSearchActiveWhere 为空间侧 members/search 组装跨列 OR LIKE 条件。
 // list / count 共用同一条件，避免搜索范围漂移导致分页错位。
--- a/modules/space/model.go
+++ b/modules/space/model.go
@@ -44,10 +44,17 @@ type MemberModel struct {
 type memberSearchModel struct {
 	MemberModel
 	Name     string
+	RealName string // 实名兜底（user_verification.real_name），仅作 name 为空时的回退源
 	Username string
 	Email    string
 	Phone    string
 	Robot    int
+}
+
+// DisplayName 返回搜索结果成员的展示名，复用 issue #344 的兜底链（见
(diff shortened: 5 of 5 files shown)


Jerry-Xin (dismissed):
The PR is in scope for `octo-server` and correctly targets the space member-search path.

💬 Non-blocking
🟡 Warning: `modules/space/db_manager.go:59` makes `uv.real_name` searchable for every member, including members whose returned display name still comes from non-empty `user.name` via `modules/space/model.go:283`. That seems consistent with the PR goal of real-name search, but it is worth confirming as the intended privacy/product contract because the response does not separately expose `real_name`.

✅ Highlights
🔵 `modules/space/db.go:184` and `modules/space/db.go:217` keep list/count query shapes aligned after adding `user_verification`, avoiding pagination drift for real-name matches.

🔵 `modules/space/api_member_search.go:75` reuses the same display-name fallback used by member list, preserving response shape while fixing blank rendered names.

🔵 Regression coverage in `modules/space/api_member_search_fallback_test.go:19` and `modules/space/api_member_search_fallback_test.go:66` covers both display fallback and real-name keyword search.

I did not find a blocking bug, permission bypass, injection issue, or error-envelope regression in the changed code. Tests were not run locally during this review.

OctoBoooot (commented):
## Review: fix(space): member search display-name fallback + real_name search (#581)

**Verdict: Approve with comments** — clean, well-scoped fix for the #434 P1 defect (the #420 display-name fallback was applied to the member *list* but not the sibling *search* path). Correctness verified; the one caveat is that CI has not actually run the new tests yet (draft PR, code checks SKIPPED). Not a blocker, but the green cannot be confirmed until CI runs.

This PR is a **draft** and `mergeStateStatus` is BLOCKED — findings below are advisory; nothing here requires a re-request, but the author should get code CI to run before merge.

### What I byte-verified
- **Helper extraction is behavior-preserving.** `resolveMemberDisplayName(name, realName, uid)` is the exact prior `MemberDetailModel.DisplayName()` chain (name → real_name → `"User "+uid`), and both `MemberDetailModel` and the new `memberSearchModel.DisplayName()` now delegate to it. `memberSearchModel` embeds `MemberModel` (has `UID`) and adds `Name`/`RealName`, so `m.UID/m.Name/m.RealName` all resolve. This kills the divergence that caused the bug — good structural call.
- **COLLATE pin is index-safe (does NOT repeat the #567/#482 trap).** The pin is `uv.user_id = sm.uid COLLATE utf8mb4_general_ci` — on `sm.uid`, the *driving-table probe value*, matched to `uv.user_id`'s forced `general_ci`. The indexed side (`user_verification.user_id`, PK) keeps its collation, so its index stays usable; this is the correct placement, the inverse of pinning the indexed column itself. Matches the established `bot_api/resolve_targets.go` pattern, whose in-code comment documents the same Error-1267 reasoning. `sm` drives the LEFT JOIN (filtered by `space_id+status`), so no access-path index is defeated in either join direction.
- **List/count symmetry holds.** Both `searchMembers` and `countSearchMembers` add the identical `LEFT JOIN user_verification` + same COLLATE pin, and both consume `memberSearchActiveWhere` (now including `uv.real_name`). So a keyword hitting only `real_name` can't make `list` and `count` disagree → no pagination drift. `user_verification.user_id` is PK (1:1), so the LEFT JOIN cannot inflate `COUNT(*)`.
- **Privacy gate preserved.** `real_name` is added to searchable columns, but it's an admin-only, active-space+role≥admin endpoint and real_name is already surfaced in the member list, so searchable-granularity == visible-granularity (no new leak). Placeholder still refuses to fall back to `short_no`/`username`; the test asserts this.
- **Scope discipline is correct.** The P0 collation risk on `queryMembers` is deliberately left to #482; this PR makes its own new join self-contained via the explicit pin rather than depending on #482. P2/P3 items deferred to their trackers. Right call — no scope creep.
- **Test helpers resolve.** All referenced seeders/decoders (`seedFallbackUser`, `seedFallbackVerification`, `seedMemberSearchMember/Space/User`, `findMemberSearchItem`, `decodeMembersSearchResp`, `getMembersSearch`) exist in the space package, and `memberDisplayNamePlaceholderPrefix` is defined — so the new test compiles.

### Questions
- `api_member_search_fallback_test.go` — the two `Test_Issue434_*` tests have **never executed** (draft PR → Build/Test/Vet/Lint all SKIPPED; author states no MySQL/WuKongIM in the sandbox). They compile and read correctly, and the "RED-before-fix" claim is asserted by construction, but mutation-thinking is unverified until CI runs. _(confidence: high that they're unrun; can't confirm they pass.)_ Recommend: mark the PR ready / trigger CI so the regression tests actually run green before merge — that's the one thing standing between this and a clean Approve.

### Minor
- PR description says `model.go (+21/-7)`; actual is `+26/-7` (the extra lines are the new helper's doc-comment). Cosmetic — doesn't affect behavior — but worth correcting so the bug-fix-report stat matches the diff.

### Praise
- Extracting the fallback into a single `resolveMemberDisplayName` and pointing both models at it directly addresses the *root cause* (#420 patched one path and the sibling silently diverged), not just the symptom. And the collation pin is reasoned from the actual compat-repair migration + a cited in-repo precedent rather than cargo-culted — the ON-clause comment explains exactly why the pin is on `sm.uid` and why the 1:1 PK join keeps `COUNT(*)` honest. That's the right depth for a query-correctness change.

- #482 (P0 `queryMembers` collation) and the P2/P3 items (avatar consistency, other bare-`user.name` endpoints, shared `pkg/displayname`) are correctly deferred per the umbrella-issue split.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #581 (octo-server)

## Summary

This PR fixes two halves of issue #434: (1) the member-search endpoint now applies the same display-name fallback chain (name → real_name → "User <uid>") that the member-list endpoint received in #420/#344, so blank-name members no longer render as empty rows in search results; (2) `user_verification.real_name` is added to the search columns so members with empty `user.name` are discoverable by their verified real name. The implementation extracts a shared `resolveMemberDisplayName` function to DRY the fallback logic between list and search paths.

Overall: focused, well-commented, correctly paired list/count JOINs, and covered by integration tests that exercise all three fallback branches plus the new real_name search. One pre-existing parity observation noted below.

## Verification

Static analysis only at head `e7b2df2f1cb5436c232a855bba80c237e8edfd45`; build and tests not executed in this environment.

- ✅ **Display-name fallback chain** — `resolveMemberDisplayName` at `modules/space/model.go:286` implements the 3-tier fallback (name → realName → placeholder) correctly. Both `MemberDetailModel.DisplayName()` (`:300`) and `memberSearchModel.DisplayName()` (`:58`) delegate to the same shared function. No branch returns an empty string.
- ✅ **Data flow: SQL → struct → API response** — `searchMembers` selects `IFNULL(uv.real_name,'') as real_name` (`modules/space/db.go:186`), which maps to `memberSearchModel.RealName` (case-insensitive field match via dbr, same pattern as existing `Name`/`Username`/`Email` fields). `DisplayName()` consumes `m.RealName` at `:58`. API handler calls `m.DisplayName()` at `modules/space/api_member_search.go:75`.
- ✅ **List/count JOIN parity** — Both `searchMembers` (`:191`) and `countSearchMembers` (`:217`) now LEFT JOIN `user_verification` with identical COLLATE clauses. Both consume `memberSearchActiveWhere` for the keyword filter, so list and count operate on the same row set. Pagination cannot drift.
- ✅ **Search column wiring** — `memberSearchActiveColumns` at `modules/space/db_manager.go:61` now includes `"uv.real_name"`. `memberSearchActiveWhere` generates `uv.real_name LIKE ? ESCAPE '\\'` via the existing `buildLikePattern` which properly escapes `\`, `%`, `_`. LIKE value is parameterized — no SQL injection vector.
- ✅ **LEFT JOIN cardinality** — `user_verification.user_id` is a primary key (1:1 with user), so the LEFT JOIN cannot produce duplicate rows. When no verification record exists, `IFNULL(uv.real_name,'')` correctly yields `''`.
- ✅ **Authorization** — Endpoint is gated by `member.Role < spaceRoleAdmin` at `modules/space/api_member_search.go:46`. No auth boundary change; real_name is already displayed in the member list.
- ✅ **Privacy** — Placeholder uses `"User " + uid` (`memberDisplayNamePlaceholderPrefix`), which only exposes information already in the response. `short_no`/`username` are explicitly excluded from fallback (privacy-gated).
- ✅ **Refactoring correctness** — `MemberDetailModel.DisplayName()` old inline body replaced with delegation to `resolveMemberDisplayName(m.Name, m.RealName, m.UID)` — behaviorally identical.

## Findings

No P0/P1 issues. One P2 observation below.

### P2 — COLLATE parity gap between list and search (`modules/space/db.go:158`)

`queryMembers` (member list, line 158) joins `user_verification` without the `COLLATE utf8mb4_general_ci` pin that `searchMembers` (line 191) and `countSearchMembers` (line 217) now use. On MySQL 8.0 deployments where `space_member` inherits `utf8mb4_0900_ai_ci` as the default collation, the list query could hit `Error 1267 (Illegal mix of collations)` while the search query succeeds. The code comment at `db.go:175-180` explicitly defers this to issue #482, so it is a known pre-existing gap, not introduced by this PR. Worth prioritizing #482 to close the asymmetry.

## Blind-spot checklist (R5)

- **C1 — Dual-path parity**: list ↔ search display-name fallback — ✅ aligned via shared `resolveMemberDisplayName`. COLLATE clause — ⚠️ list lacks it (pre-existing, tracked in #482, see P2 above). Admin-side `memberSearchColumns` / `memberSearchWhere` — N/A (different code path, different consumer, not touched by this PR).
- **C2 — Control-flow ordering / reuse**: `resolveMemberDisplayName` is called from exactly two sites (`MemberDetailModel.DisplayName` and `memberSearchModel.DisplayName`). No nested/re-entrant reuse risk. N/A.
- **C3 — Authorization boundary**: No new endpoint or capability exposed. Search remains gated by `spaceRoleAdmin`. N/A.
- **C4 — Authorization lifecycle**: No auth/status changes. N/A.
- **C5 — Build/runtime divergence**: No build artifact or packaging change. N/A.
- **C6 — Governance/docs**: No governance doc changes. N/A.

## Data-flow trace

|---|---|---|
| `m.RealName` in `DisplayName()` | `IFNULL(uv.real_name,'') as real_name` in `searchMembers` SQL → dbr field mapping | ✅ Yes |
| `m.Name` in `DisplayName()` | `IFNULL(u.name,'') as name` in `searchMembers` SQL | ✅ Yes |
| `m.UID` in `DisplayName()` | `sm.*` (space_member.uid) | ✅ Yes |
| `m.DisplayName()` in API resp | `resolveMemberDisplayName` → never returns "" | ✅ Yes |
| `count` vs `list` alignment | Shared `memberSearchActiveWhere` + identical JOINs | ✅ Yes |
| `uv.real_name` in LIKE search | `memberSearchActiveColumns` includes `"uv.real_name"`, JOIN present in both queries | ✅ Yes |

N/A — automated review review, no prior rounds.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. PR correctly fixes both halves of #434 (display-name fallback + real_name search) with proper JOIN parity between list/count, shared fallback function, and comprehensive integration tests. One P2 observation about pre-existing COLLATE asymmetry (tracked in #482) does not block.

Jerry-Xin (commented):
Approve-with-comments (not a formal approval yet). The code is sound — I byte-verified the high-risk areas below and they hold — but this PR is currently a **draft**, so Build/Test/Vet/Lint are **SKIPPED, not passed**. The two `Test_Issue434_*` regression tests this PR adds have therefore **never executed once**. Skipped ≠ green; I confirmed the new test *compiles* (all helpers/model fields resolve), but I won't stamp a formal APPROVE on a fix whose own regression tests haven't run. Mark the PR ready (or otherwise trigger CI) so those tests run green, and I'll convert this to a formal approval.

Merits (byte-verified at head, all sound):
- SQL injection — safe: search builds `col + " LIKE ?"` with hardcoded column-name allowlist; only the keyword flows through the placeholder. No concatenation.
- Collation / MySQL 1267 — correctly handled: both new joins (searchMembers + countSearchMembers) pin `COLLATE utf8mb4_general_ci` to match `user_verification.user_id` (general_ci) against `space_member.uid` (unpinned → 0900_ai_ci on MySQL 8), mirroring the existing resolve_targets.go pattern. The COLLATE is on the probe value, not defeating the indexed column. The untouched `queryMembers` (list path) join remains unpinned — that's pre-existing, out of scope here, tracked in #482.
- display-name fallback — clean extract-helper: name → real_name → "User <uid>", shared by both the detail model and the search model (kills the exact list/search divergence behind #420/#434). No nil/empty deref.
- real_name search — no new exposure: triple-gated (space-active + caller-is-member + role≥admin); real_name is already returned by the member list, so searchable-granularity == visible-granularity. The 1:1 PK LEFT JOIN can't inflate COUNT.

Non-blocking:
- Whether "all members searchable by real_name" is the intended privacy contract is a product question, not a code defect — worth a product confirm.
- List-vs-search COLLATE asymmetry (queryMembers unpinned) is pre-existing/#482, not introduced here.

Verdict: merits ready; formal approval pending CI actually running green on a non-draft PR.

yujiawei (approved):
# Code Review — PR #581 (octo-server)

**Scope reviewed:** head SHA `e7b2df2f1cb5436c232a855bba80c237e8edfd45`. Fix for #434 P1 — apply the member display-name fallback (`name → real_name → "User <uid>"`) to the member **search** path and make members searchable by verified `real_name`, bringing search into parity with the member **list** path.

## Spec compliance

The diff does exactly what the linked issue's P1 item asks, and nothing more:

- **Under-build:** none. Both halves of the P1 defect are addressed — blank-name rows now render a fallback name (`api_member_search.go:73` maps `m.DisplayName()` instead of raw `m.Name`), and `real_name` is now a searchable column (`db_manager.go:59`).
- **Over-build:** none. The P0 collation issue on the sibling `queryMembers` list path is deliberately left to #482 (`db.go:179-180`), the shared `pkg/displayname` extraction to P3, and avatar/other endpoints to their own trackers. Scope is held tight.
- **Deviation:** none. The refactor into `resolveMemberDisplayName` (`model.go`) is behavior-preserving; `MemberDetailModel.DisplayName()` now delegates to the same helper the search path uses, so list and search cannot drift.

## Code quality

**Verified correct:**

- **Collation safety.** The new `space_member → user_verification` join pins `uv.user_id = sm.uid COLLATE utf8mb4_general_ci` (`db.go:191`, `db.go:217`). This is the correct fix: `user_verification` is created `utf8mb4_general_ci` and force-normalized by the compat-repair migration, while `space_member` carries no explicit collation and inherits the DB default (potentially `0900_ai_ci` on canonical MySQL 8.0). Pinning the driving-table side (`sm.uid`) keeps the `user_verification` PRIMARY KEY usable for the lookup, and follows the established `bot_api/resolve_targets.go` precedent. The `WHERE ... real_name LIKE ?` predicate carries no 1267 risk — a bound parameter coerces to the column's collation, so column-vs-literal comparisons are self-consistent.
- **List/count lockstep.** `searchMembers` and `countSearchMembers` attach the identical join and share `memberSearchActiveWhere`, so the row-set feeding `COUNT(*)` matches the paginated list under every keyword — including one that matches only `real_name`. `user_verification.user_id` is a PRIMARY KEY (1:1), so the `LEFT JOIN` cannot inflate `COUNT(*)`.
- **Injection surface.** All search columns are compile-time constants; the keyword flows only through bound `?` params, with `escapeLike` + explicit `ESCAPE '\\'` applied uniformly to the new `real_name` column. No new injection surface.
- **Authz.** The endpoint remains gated on active space (`checkSpaceActive`) and caller role ≥ admin (`api_member_search.go:26-43`). `real_name` is already surfaced to admins in the member list.
- **Tests.** Two integration tests (`api_member_search_fallback_test.go`) cover the three fallback branches (real_name / placeholder / populated-name-wins) and real_name searchability with a decoy for over-match; `go build` and `go vet` on `./modules/space/...` pass.

**Non-blocking notes (P2 — no change required to merge):**

1. **[security-sensitive — worth a human glance] Search grain vs. display grain for named members.** `real_name` is matched at full `%substring%` grain, but is only *returned* when `user.name` is empty (fallback). For a member with a populated `user.name`, an admin can still confirm substrings of that member's hidden `real_name` via keyword hits (a bounded probing oracle), whereas the response shows only `user.name`. This is analogous to the asymmetry the phone code deliberately avoids (`RIGHT(u.phone,4)` matches only the visible 4 digits). It is admin-only, within-space, and `real_name` is verification data admins are entitled to — an acceptable trade-off, but the parity claim in the `db_manager.go:55-56` comment ("检索粒度未越出展示粒度") is strictly true only for the blank-name subset. Consider narrowing that comment, or having a human confirm the trust boundary is intended.
2. **List-path collation asymmetry.** `queryMembers` (`db.go:158`) still joins `user_verification` *without* the COLLATE clause this PR adds to the search path — a pre-existing latent 1267 risk on non-canonical 8.0 deployments. Correctly out of scope here and tracked in #482, but the two sibling paths are now collation-inconsistent; ensure #482 lands before any 8.0-default deployment.
3. The COLLATE defense is not exercised by CI (the test DB pins everything to `general_ci`), so the mixed-collation path is a no-op in tests. Standing up a mixed-collation fixture is high-effort/low-value; the display-name and searchability behavior itself is well covered.

## Verdict

No P0/P1 issues. The change is small, well-scoped, correctly handles the collation edge, keeps list/count consistent, and is adequately tested. The P2 notes are advisory. **Approved.**

PR #582, opened 2026-07-14
feat(resource_share): add trusted user resource sharing platform

an9xyz:
## Summary

Add the provider-based `user-resource-share-card` platform foundation so an authenticated user can share a server-minted, display-only resource card as themselves to a DM, group, or thread.

The platform ships fail-closed: the global feature flag is off by default, no provider is enabled by this PR, and existing generic human type-17 and Bot OBO rejection remains unchanged.

## Related Issue

Follow-up to #579. This PR does not automatically close an issue.

## Linked Spec

- [`.octospec/tasks/user-resource-share-card/brief.md`](https://github.com/dmwork-org/octo-server/blob/feat/user-resource-share-card/.octospec/tasks/user-resource-share-card/brief.md)
- [Provider onboarding contract](https://github.com/dmwork-org/octo-server/blob/feat/user-resource-share-card/.octospec/tasks/user-resource-share-card/provider-onboarding.md)

## Changes

- Add an immutable startup provider registry, pinned ES256 signed intents, authoritative provider revalidation, bounded claims, and replay protection.
- Add the authenticated `POST /v1/resource-shares` platform endpoint with exact login-user/Space binding and DM, group, and thread target authorization.
- Mint fixed `octo/v1` cards server-side, sign detached-JWS share proofs, publish verification keys, and bind proofs to the observed sender, Space, target, resource revision, and finalized payload.
- Verify trusted human resource cards in card trust, message search, and offline webhook projections; missing or invalid proof is masked fail-closed.
- Add Redis-backed distributed target quotas, bounded dispatch concurrency, per-target outcomes, metrics, and a durable idempotent delivery state machine.
- Add `octo_resource_share_intent`, `octo_resource_share_delivery`, and `octo_resource_share_audit` migrations without persisting card bodies, proofs, raw intents, signatures, or credentials.
- Add provider onboarding documentation and separate smart-summary/docs provider briefs; provider activation and client proof verification remain separate rollout steps.
- Normalize unsafe inbound request IDs out of audit persistence instead of rejecting an otherwise valid share request.

## Testing

- [x] Manually verified with local MySQL, Redis, and WuKongIM
- `go test -race ./internal/resourceshare/... -count=1`
- `GIN_MODE=release go test ./modules/resource_share/... -count=1 -timeout=10m` with a configured test master key
- `go test -cover ./internal/resourceshare/... ./modules/resource_share/... -count=1 -timeout=10m` (80.5% / 88.2%)
- `go vet ./internal/resourceshare/... ./modules/resource_share/...`
- `git diff --check origin/main...HEAD`

## COMPREHENSION

   It adds one guarded path that authenticates the human actor and Space, verifies a provider-signed resource intent, revalidates the exact resource revision, authorizes each exact target, durably claims delivery, builds and proves a fixed card, and sends it as the logged-in user. Every server display projection verifies the proof against observed message context before rendering the card.

   Incorrect actor/Space/target binding could create a cross-tenant send, an inconsistent proof reconstruction could mask valid cards or render tampered cards, and delivery-state mistakes could duplicate or retry an ambiguous transport result. The implementation therefore fails closed, keeps provider/global flags off by default, treats `dispatching -> unknown` as terminal, and uses the same proof binding across send, search, and offline projections.

   Unit and race tests cover intent, registry, proof, authorization, limiter, store, API, metrics, and trust projections. The real-dependency E2E starts through the HTTP endpoint, sends DM/group/thread cards through WuKongIM, verifies sender/Space/target-bound proofs, and confirms identical retries return durable `already_sent` results without another send.

## Checklist

- [x] Updated documentation


Jerry-Xin (commented):
## Resource-share platform — security-focused review

Reviewed the full trust boundary end to end: intent verification, proof signing/verification, per-target authorization, rate limiting, durable store, service orchestration, production config wiring, and the display/search render surfaces. This is a carefully built, fail-closed design. I traced the auth and access-control paths rather than spot-checking, and found **no blocking security defects**.

I'm posting this as approve-with-comments rather than a formal APPROVE only because CI code checks (Build / Test / Vet / Lint) are still **pending** at the reviewed head (`0cd4311`) — skipped/pending ≠ green. Once those pass on this head, this is good to merge from a security standpoint. (`check-sprint` failure is governance, not a code blocker.)

### ✅ Highlights (security posture verified)

- **AuthN/AuthZ trust boundary — correct.** `internal/resourceshare/authorizer.go:26` `Authorize` verifies the actor is an *active* member of the target space, then per-kind: DM peer must be active in space; group requires active, non-deleted, non-muted membership with posting rights (`authorizeGroup`, `internal/resourceshare/authorizer.go:88`); thread additionally requires the group check **and** an active thread (`authorizeThread`, `internal/resourceshare/authorizer.go:127`). All queries parameterized (`?`), all failure/error paths deny (`targetDenied`/`targetQueryError`).
- **No IDOR.** The only write endpoint is `POST /v1/resource-shares` (`modules/resource_share/api.go:52`), gated by `AuthMiddleware` + `SharedUIDRateLimiter` + `SpaceMiddleware`, with `loginUID != ""` and `spaceID == X-Space-ID` enforced (`modules/resource_share/api.go:60`). `service.go` `Share()` binds the authenticated caller to the intent: `verified.Intent.ActorUID != loginUID || verified.Intent.SpaceID != spaceID → ErrShareForbidden`. Store row-ids are minted internally per verified request and never accepted from external input.
- **No SQL injection / no collation (1267) risk.** All store queries use bound `?` args (dbr builder or static-fragment raw SQL); the two cross-table JOINs are integer `intent.id ↔ delivery.intent_id`, not string-column joins, so no `utf8mb4_general_ci` vs `0900_ai_ci` mismatch. Migration tables (`modules/resource_share/sql/20260714000001_resource_share.sql`) are consistently `utf8mb4_general_ci`.
- **Strong proof crypto.** `internal/resourceshare/proof.go` `Verify` (fail-closed on nil/empty verifier), pins `alg`+`kid` in the protected header and rejects unprotected/extra/embedded-JWK headers (header-injection resistant), validates the authoritative `plain == cardmsg.BuildPlain(card)` (rejects display-text tampering), and rejects unreviewed envelope fields via allowlist (`validateUnsignedShareEnvelope`). Keys validated on-curve P-256 with the public point matching `D`.
- **Replay + idempotency.** `ClassifyReplay` uses `crypto/subtle.ConstantTimeCompare` on the intent fingerprint; nonce/idempotency stored as SHA-256 with a unique key; expired intents cannot start new delivery.
- **SSRF/deep-link hardening.** `safeResourceLink` requires `https`, non-empty host, and `link.User == nil` (blocks embedded credentials / scheme downgrade).
- **Rate limiting fail-closed.** `internal/resourceshare/limiter.go` `Allow` returns a deny decision on *any* error; the atomic Lua bucket makes denials read-only so an exhausted bucket can't starve others.
- **Fail-closed config.** `DM_RESOURCE_SHARE_ENABLED` unset → feature disabled; enabled deployments require a reviewed provider spec + active proof signer or the constructor errors. `reviewedProviderSpecs()` is compile-time only — requests can never register a provider dynamically.
- **Display/search trust is viewer-bound, not a stored flag.** The identity-only `Trusted(fromUID)` path was replaced by `TrustedMessage(MessageObservation)` (`modules/cardtrust/cardtrust.go`), forcing every render/search call site to supply `{FromUID, ViewerUID, SpaceID, Target}`. `Verify` then requires `space/actor/target` to match the observation, so a card proven for space X or DM `{a,b}` will not verify for a viewer in space Y or a different pair. `modules/webhook/common.go` and all four `modules/messages_search/*` handlers pass the viewer context; `nil` verifier / missing context / mismatch all degrade to the static `[卡片]` placeholder.

### 🟡 Non-blocking (defense-in-depth)

- `internal/resourceshare/store.go`: row-id mutators (`LoadDelivery`, `transition`, `MarkSent`, `BeginDispatch`, …) carry no `actor_uid`/`space_id` SQL predicate and rely entirely on the caller having pre-verified the fingerprint/identity hash. Safe under the current `Share()` contract, but a future caller passing an externally influenced row id would bypass scoping. Consider a contract comment or an actor-scoped predicate as belt-and-suspenders.
- `modules/messages_search/*`: the viewer projection is passed as a variadic `projection ...cardProjectionContext`. A legacy call site omitting it would leave `ViewerUID`/`SpaceID` empty. This degrades fail-closed (verification fails → placeholder), and all four production call sites pass it explicitly, but a required (non-variadic) parameter would be marginally safer.
- `internal/resourceshare/service.go`: `Observer == nil` silently defaults to a no-op observer. Not a security issue, but denial/rate-limit signals are how these events surface operationally — worth ensuring it's always wired in production.

Nice work — the fail-closed discipline and the per-viewer re-verification of trust are exactly right for this trust boundary.

internal/resourceshare/registry.go
```diff
@@ -0,0 +1,212 @@
...
+	return &ecdsa.PublicKey{Curve: elliptic.P256(), X: new(big.Int).Set(key.X), Y: new(big.Int).Set(key.Y)}, true
+}
+
+func isNilInterface(value interface{}) bool {
+	if value == nil {
+		return true
+	}
+	rv := reflect.ValueOf(value)
+	switch rv.Kind() {
+	case reflect.Chan, reflect.Func, reflect.Interface, reflect.Map, reflect.Ptr, reflect.Slice:
```
OctoBoooot: **[blocker]** Lint/govet fails here: `inline: Constant reflect.Ptr should be inlined (govet)`. This reddens the code-CI Lint check, which bars approval. Inline the `reflect.Ptr` constant per the govet suggestion — one-line fix. Everything else in the review is non-blocking; this is the only thing between the PR and a clean pass on the CI gate.

OctoBoooot (changes requested):
## Review: feat(resource_share): add trusted user resource sharing platform (#582)

**Verdict: Request changes** — the code-logic review across crypto/proof, authz/trust-boundary, delivery/idempotency, and store/SQL found **no blockers and no majors**; this is an unusually careful, fail-closed security implementation. But **code CI is red** (Lint/govet), which must be green before this can be approved. The fix is one line.

CI is red (1 code check failing: **Lint**). `check-sprint` is green; Build/Vet/osv/i18n/Card-Dispatch green; Test still running.

Risk tier: **critical** (new trust boundary — signed intents, JWS proofs, cross-tenant target authz, SQL migration, new crypto dep) and **size XL** (50 files / +7786). `needs-human-review` applied — a human maintainer must make the merge call regardless of bot verdicts.

Reviewed by fan-out across four independent dimensions; every blocker/major candidate was adversarially verified. One candidate major was refuted and downgraded (see below) — flagging that explicitly so its retraction is on the record.

### Blockers
- `internal/resourceshare/registry.go:203:67` — **Lint fails**: `inline: Constant reflect.Ptr should be inlined (govet)`. This is a real golangci-lint/govet failure in the PR's own new file, so the code-CI gate is red and APPROVE is not available until it's fixed. Trivial: inline the `reflect.Ptr` constant per the govet `inline` suggestion. _(confidence: high — read from the failing Lint job log.)_

### Questions
- `internal/resourceshare/store.go:246` — a delivery that crashes/cancels **after** `BeginDispatch` commits `state=dispatching` but **before** `MarkSent`/`MarkUnknown` appears permanently unrecoverable: `ReclaimPreTransport` only resets `rate_limited`/`failed`, no sweeper resets `dispatching`, and every retry maps `dispatching → ShareUnknown` without re-dispatching. This is the deliberate at-most-once tradeoff (correctly prevents double-send), but a message that was never actually put on the wire is then lost with no operator recovery path. Is that the intended durability contract, or should there be a bounded sweeper that re-drives a stale `dispatching` row (idempotency key would keep it safe)? _(confidence: high this is the behavior; the severity is a product/durability call, hence a question not a finding.)_
- `internal/resourceshare/proof.go:241` — search DM projections verify the proof against the **viewer's** request Space; with `OCTO_SEARCH_REQUIRE_SPACE_ID=false` an empty observed Space fails the proof and masks the card fail-closed (safe). Confirm no legitimate cross-space viewing scenario exists where a validly-shared card must still render for a viewer whose current Space header differs from the send Space — if it does, valid cards would be silently masked (availability, not security).

### Minor
- `internal/resourceshare/service.go` (audit request_id) — the only test of `normalizeAuditRequestID` asserts the drop-unsafe case; no test asserts a *valid* request_id is actually persisted. Mutating the function to `return ""` unconditionally would leave all tests green. Low blast radius (request_id is diagnostic), but a positive-path assertion would close the mutation gap.
- `internal/resourceshare/store.go:81` — `idempotency_hash` is written to `octo_resource_share_intent` but never read or backed by a UNIQUE key (all dedup goes through `nonce_hash`). It's effectively write-only metadata; either enforce it or document that nonce is the sole dedup axis.
- Test/prod schema parity: the in-memory sqlite test store uses unbounded `TEXT` while prod is `VARCHAR(n)`; the migration-shape test checks columns/keys but not lengths, so a future column-length truncation regression wouldn't be caught by any test. Consider a length-parity assertion.

### Nits
- `internal/resourceshare/authorizer.go:51` — the DM authorizer checks only same-space active membership and does not re-check bidirectional-blacklist/friend-DM in-process. **This is not a gap** (I initially flagged it as a major, then refuted it): DM blacklist + friend enforcement is applied by WuKongIM's datasource `getBlacklist` hook (`modules/webhook/api_datasource.go:115` → `userService.ExistBlacklist`), which fires on the person channel at delivery for any send including `ctx.SendMessageWithResult`; a card from A to a peer who blacklisted A is dropped at delivery. The new authorizer is consistent with the established `internal/carddispatch/authorizer_db.go:50` sibling. Optional defense-in-depth only: mirror the check in-process so a future datasource-hook regression fails closed at authorization time.

### Praise
- **Proof layer is genuinely well-built.** ES256 is pinned three ways (proof `alg`, protected-header `alg`, and a P-256 `*ecdsa.PublicKey` so go-jose rejects non-ECDSA) — `alg=none`/alg-confusion is unreachable; the JOSE header is constrained (no embedded JWK, no `crit`/`b64`/`jku`/`x5u`), closing key-substitution and detached-b64 tricks. Detached-JWS reconstruction is **symmetric and deterministic** (both seal and verify build the same struct → `json.Marshal` → `jcs.Transform`, RFC 8785), and verify whitelists exactly the unsigned envelope fields — so the "tampered card renders valid / valid card masked" risk the PR body names is actually closed, and the authoritative `plain` is recomputed so a signed card can't carry a lying preview. Keys are deep-copied into an immutable startup registry with no runtime injection, and each intent verifies only against its own provider's key (no confused deputy).
- **Fail-closed by construction**, not just by convention: `reviewedProviderSpecs()` is compile-time nil, the flag defaults off, and enabling without a reviewed provider + signing key panics at startup. Actor/Space binding is enforced before any work (login UID/Space must equal the signed intent; API uses `DisallowUnknownFields`), and cross-tenant group/thread sends are rejected on exact `g.space_id == spaceID`.
- **Concurrency is right:** the Redis limiter is a single atomic Lua script (check-all, decrement-none-on-denial — no TOCTOU overshoot) and fails closed on Redis unavailability; the delivery state machine is a conditional `UPDATE ... WHERE state=?` CAS with `RowsAffected()==1`, transactional with the audit insert, so two concurrent identical requests can't both send.
- **No secrets at rest** (only `nonce_hash`/`fingerprint`/`idempotency_hash` BINARY(32) — verified in `store.go`), all queries parameterized, request-id normalization safely bounded to `[A-Za-z0-9-_.~]{0,128}`, and the security-critical negative tests (proof tamper matrix, authorizer deny matrix, idempotency/rollback) are mutation-resistant.

- One new dependency: `github.com/gowebpki/jcs v1.0.1` (RFC 8785 JSON canonicalization) — single-purpose, needed for deterministic proof input; `dependency-review` CI passed.

Jerry-Xin (dismissed):
Verdict: Request Changes. Sole hard blocker is the red required Lint check. This is not a draft (isDraft=false), so CI ran for real.

## 🔴 Blocking

**Lint (govet) failure — internal/resourceshare/registry.go:203: "Constant reflect.Ptr should be inlined (govet)".**
The required `Lint` check is red at this head (conclusion=failure). Inline the `reflect.Ptr` constant at registry.go:203. Trivial, but must go green before merge. (`Test` in_progress; Build/Vet and all others green.)

## Security review — no blockers (byte-verified at head)
Trust-boundary surface is sound; none of the below blocks:
- AuthZ: authorizer validates actor active-membership + per-kind gates; all error paths deny.
- IDOR: single write entrypoint triple-gated; Share() binds caller↔intent; row ids generated internally.
- SQL injection: fully parameterized; only raw SQL concatenates static fragments.
- Collation/1267: the two joins are integer intent.id↔delivery.intent_id; migration tables uniformly general_ci; no cross-table string comparison.
- Proof crypto: ES256 detached JWS, alg+kid pinned, rejects extra/unprotected headers, verifies authoritative plaintext, P-256 on-curve check.
- Replay/idempotency: constant-time compare; SHA-256 nonce/idempotency keys; expired intents can't start delivery.
- SSRF/deep-links: safeResourceLink forces https, non-empty host, forbids embedded credentials.
- Rate-limit fail-closed; feature-flag fail-closed (disabled when unset).
- Display/search trust: viewer-bound re-check, stored trust flags not honored, no cross-space/viewer leak.

## 🟡 Non-blocking

**MaxIntentLifetime cap can be defeated by int64 overflow in the duration math — internal/resourceshare/intent.go:149.**
`time.Duration(intent.ExpiresAt-intent.IssuedAt)*time.Second > p.spec.Limits.MaxIntentLifetime`: ExpiresAt/IssuedAt are int64 unix *seconds*; multiplying the span by time.Second (1e9) overflows int64 for spans beyond ~2.9e9s and wraps negative (~292y), so a negative Duration passes the `> 5m` cap → the "intent lifetime too long" limit is bypassed for a sufficiently large signed `exp`. The preceding guards don't cover it: line 145 only rejects *past* expiry, never far-future, so a huge ExpiresAt reaches line 149. **Why 🟡 not 🔴:** `exp`/`iat` live inside the ES256-signed intent payload, signature-verified against a server-registered provider key *before* validateIntent runs — an external attacker cannot forge a crafted `exp` without the provider's private key. So the cap guards against a misbehaving/buggy *trusted* provider, not an untrusted-input boundary; practical severity is low. Overflow-safe fix (compare in seconds, avoid the ×1e9):
`if intent.ExpiresAt-intent.IssuedAt > int64(p.spec.Limits.MaxIntentLifetime/time.Second) { return invalid("intent lifetime too long") }`
(the subtraction is already safe — line 139 enforces IssuedAt>0 and ExpiresAt>IssuedAt). Credit to the reviewer who spotted the overflow.

- store.go row-id mutator lacks an actor/space SQL predicate (safe under current caller contract; add a contract comment).
- messages_search viewer projection is variadic (fail-closed degrade; a required param is sturdier).
- service.go nil Observer silently no-ops (ops visibility).
- (peer) dispatching-state delivery has no recovery path on abnormal exit — possible message loss; worth a recovery/retry.
- (peer) idempotency_hash lacks a unique index — make the dedup guarantee explicit in schema or document it.
- (peer) test DB vs prod DB column-length mismatch — add a length-consistency check.
- (peer) auditRequestID test lacks a positive-path assertion.

Verdict: Request Changes — sole hard blocker is the red Lint (registry.go:203). Security merits sound; once Lint (and Test) are green this converts to approve.

yujiawei (changes requested):
# Code Review — PR #582 (octo-server)

Reviewed at head SHA `0cd4311970ef6877345c9f757ea81ba135c11bc6` against merge-base `5e336a0c`. Build, `go vet`, and the package test suites (`internal/resourceshare`, `modules/resource_share`, `modules/cardtrust`) all pass locally.

This is a security-critical foundation: ES256 detached-JWS intent verification, an asymmetric ES256 share-proof over an RFC 8785/JCS canonical signing object, durable MySQL idempotency/replay state, a Redis-Lua distributed token bucket, and a new human-sender (type-17) card trust exception at the search and offline-push read boundaries. I reviewed it across crypto correctness, the trust-boundary read path, the durable delivery state machine, rate limiting, and spec compliance.

Overall the implementation is unusually well hardened — the crypto and trust-boundary paths are clean and fail-closed. There is **one correctness defect** that should be fixed before merge, plus a few non-blocking advisories.

## 1. Spec compliance

**Spec: ✅**

- **Under-build:** none. `reviewedProviderSpecs()` returning `nil` is consistent with the stated platform-foundation slice ("each provider onboarding adds a reviewed spec; requests can never register a provider dynamically"), not a gap. The service correctly builds in disabled mode with no providers.
- **Over-build:** none. The HTTP surface is exactly `POST /v1/resource-shares` (single `intent` field, `DisallowUnknownFields`, `X-Space-ID` required and equality-checked against the validated Space, 128 KiB pre-decode cap) plus the public `GET /v1/resource-shares/proof-jwks` with cache headers. No `from_uid` is accepted; `FromUID` is always the signed/login actor UID (`service.go:311`). Proof signing uses a dedicated managed key, not `OCTO_MASTER_KEY`. Feature and provider flags default off; missing signing config disables new shares while retaining the verification ring.
- **Deviation:** none observed. Per-target results are returned in canonical target order with the stable outcome vocabulary; no card/proof/intent is echoed in the response.

## 2. Code quality

### P1 — Max-intent-lifetime ceiling is bypassable via `time.Duration` overflow

`internal/resourceshare/intent.go:149`

```go
if time.Duration(intent.ExpiresAt-intent.IssuedAt)*time.Second > p.spec.Limits.MaxIntentLifetime {
    return invalid("intent lifetime too long")
}
```

`ExpiresAt`/`IssuedAt` are attacker-influenceable signed int64 seconds. `time.Duration` is int64 nanoseconds, so `(ExpiresAt-IssuedAt)*time.Second` overflows for any span beyond ~292 years and wraps **negative**, which is `<= MaxIntentLifetime` and passes the check. A signed intent with a far-future `exp` therefore defeats the 5-minute platform ceiling — the exact invariant this guard exists to enforce ("five minutes maximum intent lifetime … provider limits may be lower, never higher"). The earlier `ExpiresAt <= IssuedAt` and future-issue checks do not bound the span, and expiry is compared with `time.Unix(...)` so it does not catch this.

Reproduced: with `iat = 1_800_000_000`, `exp = iat + 400*365*24*3600`, the computed duration is negative and the ceiling check passes.

Impact: the effective replay/idempotency window a single signed intent holds becomes unbounded instead of ≤5 min. Not reachable in this foundation slice (no provider is wired and the feature defaults off), but it is latent in load-bearing security code and will become reachable the moment the first provider onboards — the provider-onboarding PRs will build on the assumption that this ceiling holds.

Suggested fix — compare in the integer domain before converting:

```go
if intent.ExpiresAt-intent.IssuedAt > int64(p.spec.Limits.MaxIntentLifetime/time.Second) {
    return invalid("intent lifetime too long")
}
```

Please also add a RED test with a far-future `exp` (not just a moderately-over-limit one) so the boundary is covered.

### P2 — advisories (non-blocking)

- **`modules/resource_share/production.go` (~L173–198):** on the enabled success path the dedicated Redis client (`PoolSize: 10`) is only `Close()`d on the two construction error branches; once the service is built, nothing owns its lifecycle and `1module.go` registers no Stop/Close hook. Harmless at process scope (matches other long-lived singletons) but would leak on re-init (tests/hot-reload).
- **`internal/resourceshare/service.go:172-181`:** an expired first-use intent still writes the `octo_resource_share_intent` nonce row before returning the "expired cannot start delivery" error, so replayed-but-expired intents can amplify writes to the intent table. Consider rejecting expiry before `ClaimIntent`, or rely on the unique-nonce constraint knowingly.
- **`internal/resourceshare/service.go:243-245`:** the delivery-branch retry selector uses the clock-skewed `verified.Expired` while `ReclaimPreTransport` guards on the hard `expires_at`; a mild near-expiry window can classify inconsistently. Cosmetic — both fail closed.

## 3. What I verified clean (adversarial checks that did NOT find a bug)

- **Crypto:** alg/key confusion (alg+kid pinned per key; ES256-only keys; embedded JWK and all extra/critical JOSE headers rejected on both intent and proof paths); detached-JWS reconstruction binds actor/space/target to the **observed** context, not attacker-supplied proof fields; JCS rejects duplicate keys, closing selector-vs-strict-decode divergence; intent signature verified over raw bytes with `bytes.Equal(raw, verifiedRaw)`; nonce fingerprint = `sha256(JCS(whole intent))` with constant-time compare; P-256 keys validated on-curve with recomputed public point.
- **Trust boundary:** a forged/self-minted human type-17 card fails closed on every render surface (search + offline push) — the verifier re-derives authoritative `plain` and requires a valid proof bound to observed sender/space/channel; recipient-side DM peer normalization (`webhook/common.go` rewriting `channelID` to `FromUID`) is correct and symmetric with the order-independent signed pair; empty projection / empty `SpaceID` / thread-parse-failure all fail closed; the Bot/webhook `Trusted()` fast-path is unchanged for non-human cards.
- **Rate limiter:** the Lua script is atomic (read-only denial pass, then consume-all only if every bucket has capacity — no partial consume); `Allow` fails closed with a bounded retry on every error surface; token-bucket math has no off-by-one/negative; provider budgets are rejected at construction if they exceed platform maxima; dispatch-slot acquire/release is correctly paired with no leak on early returns.
- **State machine:** unique `nonce_hash` + unique `delivery_id` + `WHERE state=?` transitions prevent double-send; fingerprint binding rejects nonce reuse with changed claims; expired intents cannot start a new delivery.
- **DM authorization:** the Space-membership-only check is correct — it mirrors the existing `CheckBothMembers` Space-mode DM policy, and phase-one explicitly rejects the legacy personal-space mode where bidirectional friendship applies. `space_member` has no `is_deleted` column (status-only), and `group_member.forbidden_expir_time` is `NOT NULL DEFAULT 0`, so the group authorization query is safe.

## 4. Coverage / blind spots

- No live Redis or MySQL in the review sandbox, so distributed limiter and durable-store behavior were argued from source + single-threaded-Redis / transactional semantics, not executed against a cluster.
- Provider adapter behavior is not exercised (no reviewed provider ships in this slice); the platform↔provider contract is reviewed structurally.
- `dispatching` has no lease/timeout/reclaim: a crash after `BeginDispatch` but before transport strands the row as `unknown` forever. This is an accepted at-least-once liveness limitation (the brief explicitly does not claim exactly-once), so I did not raise it as a blocker — worth a follow-up for operational recovery.
- I did not re-run the cross-language proof conformance vectors.

## Verdict

**REQUEST_CHANGES** — Spec ✅ but one P1 correctness defect (the intent-lifetime overflow) in the primary security invariant. It is a one-line fix; everything else is in excellent shape and I'd be glad to approve once that ceiling check compares in the integer domain and gains a far-future test case.

internal/resourceshare/intent.go
```diff
@@ -0,0 +1,259 @@
...
+		return invalid("invalid intent timestamps")
+	}
+	if time.Unix(intent.IssuedAt, 0).After(now.Add(p.spec.Limits.ClockSkew)) {
+		return invalid("intent issued in future")
+	}
+	expired := time.Unix(intent.ExpiresAt, 0).Before(now.Add(-p.spec.Limits.ClockSkew))
+	if expired && !allowExpired {
+		return invalid("intent expired")
+	}
+	if time.Duration(intent.ExpiresAt-intent.IssuedAt)*time.Second > p.spec.Limits.MaxIntentLifetime {
```
OctoBoooot: **[major]** MaxIntentLifetime check overflows. `ExpiresAt-IssuedAt` is int64 seconds; `time.Duration(...)*time.Second` multiplies by 1e9, so a span > ~9.22e9s (~292yr) overflows int64 and wraps negative → `> MaxIntentLifetime` is false and the cap is silently bypassed. Failing input: IssuedAt=now, ExpiresAt=now+10_000_000_000 (~317yr) → not expired, not future, span*1e9 wraps to -8.4e18 ns → accepted. Not externally exploitable (intent is ES256 provider-signed) but it defeats a defense-in-depth bound + extends the replay window. Fix: compare seconds to seconds — `if intent.ExpiresAt-intent.IssuedAt > int64(p.spec.Limits.MaxIntentLifetime/time.Second)`. Credit: yujiawei caught this.

OctoBoooot (changes requested):
## Follow-up on #582 @ 0cd4311970ef — confirming yujiawei's overflow 🔴 (I missed it; adding as a major)

yujiawei is right — I byte-verified the time-overflow and my earlier review missed it. Adding it as a **major** (defense-in-depth control defeated). Verdict stays **Request changes** (already CR on the red Lint); this is a second finding to fix, not a new verdict.

### Major (new) — `internal/resourceshare/intent.go:149`, the MaxIntentLifetime check overflows
```go
if time.Duration(intent.ExpiresAt-intent.IssuedAt)*time.Second > p.spec.Limits.MaxIntentLifetime {
    return invalid("intent lifetime too long")
}
```
`ExpiresAt`/`IssuedAt` are int64 unix **seconds**; `time.Duration` is int64 **nanoseconds**, so `*time.Second` multiplies the span by 1e9. When `ExpiresAt-IssuedAt` exceeds ~9.22e9 s (~292 years), the multiply overflows int64 and wraps **negative**, so `> MaxIntentLifetime` is false and the cap is silently bypassed.

**Concrete failing input (verified):** `IssuedAt = now`, `ExpiresAt = now + 10_000_000_000` (~317 yr).
- passes the future check (issued = now) and the expiry check (far-future, not expired);
- span = 1e10 s; `1e10 * 1e9 = 1e19 > maxInt64 (9.223e18)` → wraps to `-8446744073709551616` ns (negative);
- `negative > MaxIntentLifetime` is false → an intent with a ~317-year lifetime is accepted despite the cap.

**Impact / severity (honest):** the intent is provider-signed (ES256, pinned key), so this is **not attacker-reachable without a compromised or buggy provider** — which is exactly what `MaxIntentLifetime` exists to bound (defense-in-depth on a trusted-but-fallible provider). Its silent failure means (a) the lifetime policy is unenforced for >292yr spans, and (b) the replay window for such an intent extends far beyond policy (the nonce/replay store would have to retain it much longer to stay safe). Real defect in a defensive control; graded **major** rather than blocker because the ES256 signature gate stands between it and an external attacker.

**Fix:** compare in seconds, don't build a nanosecond Duration from an unbounded span. e.g.
```go
if intent.ExpiresAt-intent.IssuedAt > int64(p.spec.Limits.MaxIntentLifetime/time.Second) {
    return invalid("intent lifetime too long")
}
```
(both sides int64 seconds — no 1e9 multiply, no overflow). Agree with yujiawei's 🟡 to add a far-future-expiry test case pinning this boundary (a `now + 300yr` intent must be rejected).

### Why I missed it (owning the gap)
My crypto/proof pass verified the *signature*, *alg pinning*, *expiry-vs-now*, and *issued-in-future* checks, but treated the lifetime-cap line as "just another bound" and didn't probe the seconds→nanoseconds Duration multiply for overflow. That's a `symbol-swap`/unit-mismatch class I should have constructed the boundary input for. yujiawei caught it; verified true.

### Verdict unchanged: Request changes
Two things to fix now: (1) the red **Lint** blocker (`registry.go:203` inline `reflect.Ptr`), and (2) this lifetime-overflow major. Everything else from my prior review stands (security otherwise clean, fail-closed, proof layer solid). `needs-human-review` remains.

internal/resourceshare/registry.go
an9xyz: Fixed in 6f131628 by replacing the legacy reflect.Ptr alias with reflect.Pointer. Verified with golangci-lint v2.12.2 across ./...: 0 issues.

internal/resourceshare/intent.go
an9xyz: Fixed in eca67c3a by comparing the signed lifetime in seconds, avoiding the int64 nanosecond overflow. RED coverage is in 766c6fd8 with a 10,000,000,000-second far-future span; the focused test, race suite, exact CI lint, and real resource-share E2E all pass.

OctoBoooot (approved):
## Delta re-review: #582 @ eca67c3 — both blockers fixed, upgrading REQUEST_CHANGES → APPROVE

Both things I requested are byte-verified fixed at this head (clean delta: ahead 4, no rebase noise; the changed files are exactly the two fixes + tests + a doc/comment touch). Verdict flips to **Approve**, conditional on **Test** finishing green (Build/Vet/Lint/Card-Dispatch/osv/i18n all green now, including the previously-red Lint; check-sprint governance green).

### ✓ Blocker 1 (Lint) — FIXED
`registry.go` — `isNilInterface` now switches on `reflect.Pointer` (the inlined constant govet demanded) instead of `reflect.Ptr`. The `Lint` check is green at this head.

### ✓ Blocker/major 2 (intent-lifetime overflow, intent.go) — FIXED correctly
The check is now the overflow-safe seconds-to-seconds comparison:
```go
if intent.ExpiresAt-intent.IssuedAt > int64(p.spec.Limits.MaxIntentLifetime/time.Second) {
    return invalid("intent lifetime too long")
}
```
No `time.Duration(...)*time.Second` nanosecond multiply → no int64 overflow; the subtraction is safe (the preceding `IssuedAt>0 && ExpiresAt>IssuedAt` guard holds). And a **regression test pins my exact failing input**: `intent_test.go` adds `{"far-future lifetime overflow", i.ExpiresAt = i.IssuedAt + 10_000_000_000}` asserting `ErrIntentInvalid`. This is the correct fix with the boundary locked — good.

### Minors addressed
- `idempotency_hash` column comment now reads "Hashed intent metadata; delivery dedup uses delivery_id" — resolves my note that the column's dedup role was unclear (nonce is the dedup axis; this is metadata).
- `store_test.go` / `service_test.go` additions harden the store/audit coverage.

### Everything else unchanged and still sound
The rest of the security surface is byte-identical to what I reviewed at `0cd4311970ef` (crypto/proof, producer isolation, fail-closed authz, no secrets at rest). The DM-blacklist item remains a nit (WuKongIM datasource enforces it; consistent with the carddispatch sibling), and the durability question (crash mid-`dispatching`) and other minors stand as non-blocking follow-ups.

Both blockers cleared, code CI green except the still-running Test. `needs-human-review` + `size/XL` remain — a human maintainer makes the final merge call on this live trust-boundary platform, and `mergeStateStatus` stays BLOCKED until Test passes and the required review lands. If Test comes back red I'll reverse. Nicely turned around — the overflow fix landed with the boundary test that pins it.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Octo-Q automated review Review: PR #582 — feat(resource_share): add trusted user resource sharing platform
**Head SHA:** `eca67c3abe7da4733a3bb0295510937b3620f86c`
**Merge base:** `5e336a0c031540519760e8157b4a5d830fd86efc`
**Files changed:** ~50 (16 new files in `internal/resourceshare/`, 1 SQL migration, module wiring, cardtrust/webhook/messages_search integration)
**Scope:** New feature — end-to-end resource sharing platform with intent/proof crypto, delivery state machine, distributed rate limiting, and card trust verification.

---

## 1. Verification Conclusion

✅ **Architecture**: Clean separation of concerns across 10 files in `internal/resourceshare/` (types, errors, store, service, authorizer, limiter, intent, proof, registry). Module registration and production wiring follow existing project patterns.

✅ **Crypto pipeline**: Intent JWS verification (ES256/P-256) with strict decoding (`DisallowUnknownFields`), proof sealing/verification with detached JWS, JCS canonical hashing — all properly implemented with key isolation and replay detection.

✅ **Delivery state machine**: `claimed → dispatching → sent/unknown` and `claimed → denied/rate_limited/failed` transitions are well-structured with optimistic locking (`WHERE state=From`).

✅ **Rate limiting**: Redis Lua script implements three-tier token bucket (global, provider, target) with read-only denial to prevent cross-bucket starvation — solid design.

✅ **SQL migration**: Three tables with appropriate indexes, bounded column types, and proper COMMENT annotations. Down migration is clean.

✅ **Card trust integration**: `TrustedMessage(MessageObservation)` correctly chains bot trust → proof verification, with corrected DM channel ID handling for WuKongIM's sender-side target quirk.

---

## 2. Findings

### P2-01: `authorizeGroup` SQL omits `space_id` filter in WHERE clause

**File:** `internal/resourceshare/authorizer.go:98`
**Severity:** P2 (pre-existing pattern, defense-in-depth gap)

1. New/amplified? **New** — this is new code in a new file.
2. PR changes behavior contract? Yes — new authorizer, should be correct.
3. Pre-existing? No — the pattern may be common in the codebase but this is new code.

**Issue:** The SQL query joins `group` and `group_member` on `group_no` only, without filtering by `space_id` in the WHERE clause. The space binding is checked post-hoc in Go:

```go
// authorizer.go:98-104
SELECT g.space_id, g.status AS group_status, g.forbidden, ...
FROM `group` g JOIN group_member gm ON gm.group_no=g.group_no
WHERE g.group_no=? AND gm.uid=? LIMIT 1
// ...
if row.SpaceID == "" || row.SpaceID != spaceID || ...
```

**Impact:** If a group somehow has `group_no` matching across spaces (unlikely with the current schema but possible if `group_no` is not globally unique), the query would return the first match, and the Go check would reject it — but a different row matching the correct space might exist. Adding `AND g.space_id=?` to the WHERE clause is a cheap defense-in-depth fix.

**Recommendation:** Add `AND g.space_id=?` to the query and bind `spaceID` as the third parameter.

---

### P2-02: Self-DM proof verification — belt-and-suspenders gap between intent and proof layers

**File:** `internal/resourceshare/proof.go:311`
**Severity:** P2 (defense-in-depth, mitigated by upstream intent validation)

1. New/amplified? **New** — new code.
2. PR changes behavior contract? Yes — new proof verification path.
3. Pre-existing? No.

**Issue:** `proofTargetForObservation` at line 311 rejects self-DM where `ViewerUID == PeerUID`:

```go
// proof.go:311
func proofTargetForObservation(obs MessageObservation) (Target, error) {
    // ...
    if target.PeerUID == obs.ViewerUID {
        return Target{}, errors.New("self-targeting not permitted")
    }
}
```

Meanwhile, `canonicalTargetKey` in intent.go also rejects self-DM during intent verification (`target.PeerUID == actorUID`). These are two independent checks. The concern: if proof verification is called via cardtrust on a message that was NOT created through the share API (e.g., a forged card), the intent-layer check never ran. But `proofTargetForObservation` catches it. So the defense holds at both layers — this is a note, not a gap.

**Recommendation:** No change needed. The dual-layer check is correct. Flagging for awareness only.

---

### P2-03: `MarkSent` failure + `MarkUnknown` failure → delivery stuck in `dispatching`

**File:** `internal/resourceshare/service.go:282`
**Severity:** P2 (store failure edge, no user-visible data corruption)

1. New/amplified? **New** — new delivery state machine.
2. PR changes behavior contract? Yes — new code.
3. Pre-existing? No.

**Issue:** In `shareTarget()`, after transport succeeds, `MarkSent` is called. If `MarkSent` fails (store error), the code falls through to `MarkUnknown`. If `MarkUnknown` also fails, the delivery row remains in `dispatching` state permanently:

```go
// service.go:275-290
if transportErr == nil {
    if err := svc.store.MarkSent(ctx, delivery.DeliveryID, result.MessageID, result.MessageSeq, result.ClientMsgNo); err == nil {
        // audit...
        return
    }
    // MarkSent failed, try MarkUnknown
}
if err := svc.store.MarkUnknown(ctx, delivery.DeliveryID, outcomeCode(transportErr)); err != nil {
    // Both failed — row stays in "dispatching"
    svc.logger.Error("delivery terminal state failed", ...)
}
```

**Impact:** A stuck `dispatching` row has no recovery path. `ReclaimPreTransport` explicitly excludes `dispatching` state. The row becomes a permanent orphan — no functional impact on users (the message was already sent or failed at transport level), but it pollutes delivery state queries.

**Recommendation:** Consider adding a TTL-based sweep for stuck `dispatching` rows (e.g., `dispatching` rows older than N minutes → `unknown`), or add a `MarkFailed` fallback that writes `failed` unconditionally on dual-failure. Low priority given that store dual-failure is extremely rare.

---

## 3. Suggestions

1. **`authorizeGroup` space binding**: Add `AND g.space_id=?` to the WHERE clause (P2-01 fix, trivial).
2. **Stuck dispatching recovery**: Add a background sweep or `updated_at`-based TTL for `dispatching` rows that never transition. Low priority.
3. **Test coverage**: The PR ships no unit tests. For a feature with this much crypto and state-machine complexity, tests for the core flows (intent verify → claim → authorize → rate limit → seal → transport → state transitions) would significantly reduce regression risk. Consider adding at least golden-path and error-path tests for `Share()`, `ClaimIntent`, `ClaimDelivery`, `transition`, `Seal`, and `Verify`.

---

## 4. Extra Observations

- **`reviewedProviderSpecs()` returns nil**: The production wiring has no providers registered yet. The feature is effectively dormant until providers are onboarded. This is safe — all code paths handle empty registry gracefully.
- **Feature flag**: `DM_RESOURCE_SHARE_ENABLED` gates the entire module. Good.
- **Body cap at API layer**: 128KB max body, 96KB max compact intent — well-bounded.
- **Nonce replay detection**: `ClaimIntent` uses INSERT-then-SELECT-on-conflict with constant-time fingerprint comparison — correct anti-replay design.
- **Audit trail**: `insertDeliveryAudit` uses INSERT...SELECT for atomic audit creation — good.

---

## 5. Data-Flow Trace

|---|---|---|
| `VerifiedIntent.ActorUID` | JWS payload → intent decode → `validateTargets` | ✅ Flows correctly through shareTarget → authorize → limiter |
| `delivery.DeliveryID` | `ClaimDelivery` → SHA-256 of (resource_revision + canonical_target) | ✅ Flows to transport, state transitions, audit |
| `sealed.Card` | Provider `RenderCard` → `validateUnsignedShareEnvelope` → `Seal` | ✅ Flows to transport as message card payload |
| `deepLink` | Provider `BuildDeepLink` → `verified.Resource` | ✅ Flows into card rendering |
| `rateLimitDecision.Allowed` | Redis Lua script → `Allow()` | ✅ Gates dispatch slot acquisition |
| `transportResult` | WuKongIM API via provider adapter | ✅ Flows to MarkSent/MarkUnknown state transition |
| `proofVerification` | `cardtrust.TrustedMessage` → `Verify` → JWS split + signature check | ✅ Falls back after bot trust check |
| `MessageObservation.ViewerUID` | `msg.FromUID` (webhook) / projection context (search) | ✅ Correctly passed in both call sites |

All critical data paths traced and verified. No orphan consumers found.

---

## 6. Blindspot Checklist (C1–C6)

| Check | Status | Evidence |
|---|---|---|
| **C1 — Dual-path parity** | N/A | No symmetric add/remove or create/delete pairs in this PR. Single-direction share flow. |
| **C2 — Control-flow ordering** | Clear | `shareTarget` calls are sequential per-target in a loop; no nested reuse of state transitions. `transition()` is only called from well-defined state machine methods. |
| **C3 — Authorization boundary** | Clear | API requires auth middleware + space middleware before reaching share handler. Provider registry validates all specs at startup. No endpoint bypasses authorization. |
| **C4 — Authorization lifecycle** | Clear | `activeSpaceMember` checks both space status AND member status. Group authorization checks group status, member status, deletion, mute, and posting restrictions. |
| **C5 — Build/runtime path** | N/A | No build artifact, extension, or packaging changes. Pure Go server code. |
| **C6 — Governance/docs** | N/A | No policy, SECURITY, or governance doc changes. |

---

**N/A** — This is the first review of this PR.

---

## Verdict Summary

| Severity | Count |
|---|---|
| P0 | 0 |
| P1 | 0 |
| P2 | 3 |

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. Three P2 defense-in-depth suggestions. The feature is well-architected with proper crypto, state machine, and rate limiting. Recommend addressing P2-01 (add space_id to authorizeGroup query) as a quick fix before merge. P2-03 (stuck dispatching rows) can be addressed in a follow-up.

OctoBoooot (commented):
## Adjudication: mochashanyao's two 🔴 on #582 are real-as-hardening but overstated — both defense-in-depth, not blockers. APPROVE holds.

Byte-verified both. mochashanyao's own text labels each "防御深度" (defense-in-depth), and that's the correct grade — I traced the primary boundary for each and neither is a reachable exploit. Downgrading both 🔴 → 🟡 (non-blocking); my APPROVE at this head stands. If anyone can construct a concrete failing input past the bindings below, I'll reverse immediately.

### 🔴→🟡 #1 — "store SQL missing space_id filter" — not a live gap
The store queries key on `id` / `delivery_id` / `nonce_hash`, which are **not caller-addressable**:
- `delivery_id` = `sha256({ActorUID, SpaceID, Provider, ResourceType, ResourceID, ResourceRevision, targetKey})` (`store.go DeliveryIdentity`) — actor **and** space are baked into the identity, so a row for tenant A cannot be addressed with tenant B's inputs.
- Upstream, `service.go:169` hard-binds the request: `verified.Intent.ActorUID != loginUID || verified.Intent.SpaceID != spaceID → ErrShareForbidden`, and ActorUID/SpaceID come from the **ES256-signed** intent.
So the row the store loads/updates is already actor+space-scoped by an unforgeable key. A redundant `WHERE space_id=?` is a good second lock on an already-locked door (defense-in-depth) — worth adding, not a blocker. **Falsifiable pivot:** show a caller-reachable path that loads/mutates a delivery row for a space other than the signed intent's, and this becomes a blocker.

### 🔴→🟡 #2 — "self-DM proof-layer defense gap" — not a live gap
Self-DM is rejected **at the intent layer** before any delivery: `intent.go:198` — `target.PeerUID == actorUID → "invalid dm target"` during `VerifyIntent`. The proof-observation path also forbids `ViewerUID == PeerUID` (`proof.go:364`). So a self-addressed DM can't reach the proof/delivery stage. A further proof-layer self-check is belt-and-suspenders — defense-in-depth, not a hole. **Falsifiable pivot:** construct a self-DM (actor==peer) that passes `VerifyIntent` and reaches delivery, and I'll re-grade.

### Verdict unchanged: Approve (conditional on Test)
Both #582 blockers I raised are fixed (Lint + lifetime-overflow, with the boundary test). mochashanyao's two items are legitimate hardening 🟡s (add the redundant space_id predicate; add a proof-layer self-DM assertion for depth) but neither is byte-reachable as an exploit. Still gated only on Test + `needs-human-review`. Credit mochashanyao for the depth review — the grade is 🟡, not 🔴, on the evidence.

OctoBoooot (approved):
## Re-casting APPROVE @ eca67c3 (my prior adjudication COMMENT dropped my approve vote)

Housekeeping: my APPROVE (id 4692207505) got superseded by my follow-up COMMENT adjudicating mochashanyao's two 🔴 — GitHub keeps latest-per-author, so my vote silently reverted to COMMENTED. My substantive verdict is unchanged, so re-affirming APPROVE at this head.

Summary of my verified position on #582 @ eca67c3:
- ✅ Both blockers I raised are fixed: Lint (`reflect.Ptr → reflect.Pointer`, check green) and the intent-lifetime overflow (`intent.go` now seconds-to-seconds, with a regression test pinning the `IssuedAt + 10_000_000_000` boundary). yujiawei's overflow catch → fixed correctly.
- 🟡 (non-blocking) mochashanyao's two items are defense-in-depth, not live gaps (byte-verified in my prior comment): the store's row identity is an unforgeable actor+space-bound `delivery_id` and upstream `service.go:169` hard-binds actor/space from the signed intent (redundant `WHERE space_id` = second lock); self-DM is rejected at the intent layer before delivery (`intent.go:198`). Falsifiable pivots stated; I reverse if either is shown reachable.
- Security surface otherwise unchanged and sound (crypto/proof/producer-isolation/fail-closed/no-secrets-at-rest).

Verdict: **APPROVE**, conditional on Test finishing green.

Note on the merge gate (this is NOT mergeable yet, despite any "2 票通过" banner): live `reviewDecision: CHANGES_REQUESTED` / `BLOCKED`. Still holding it: Jerry-Xin's and yujiawei's REQUEST_CHANGES are at the **old** head `0cd4311` (both pre-date the fixes and need re-casting at eca67c3), and Test is still in_progress. `needs-human-review` remains — a human maintainer makes the final call on this trust-boundary platform.

Jerry-Xin (approved):
Re-reviewed at `eca67c3`. The prior blocker is resolved and the flagged robustness issue is fixed. Approving.

### ✅ Blocker resolved
- **Lint (required check) is green.** `internal/resourceshare/registry.go` now uses `reflect.Pointer` directly in the `isNilInterface` switch (the govet "constant should be inlined" complaint on the old `reflect.Ptr` is gone). Live `Lint` check conclusion = success on this non-draft head.

### ✅ Prior 🟡 addressed
- **Intent-lifetime overflow fixed.** `internal/resourceshare/intent.go` now compares seconds-to-seconds: `intent.ExpiresAt-intent.IssuedAt > int64(p.spec.Limits.MaxIntentLifetime/time.Second)`. The raw span is no longer multiplied by `time.Second`, so the int64 overflow for multi-century spans is eliminated. A regression test (`far-future lifetime overflow`, `ExpiresAt = IssuedAt + 10_000_000_000`) was added in `intent_test.go`.
- **Audit request-ID positive test added** — `TestShareService_PersistsSafeRequestIDInAudit` in `service_test.go` complements the existing drop-unsafe test.
- **Migration column-length assertions added** in `store_test.go`, pinning the bounded VARCHAR/CHAR widths (actor_uid, space_id, delivery_id, target_ref, etc.).
- **Durable capability boundary documented** in `internal/resourceshare/store.go` (row IDs are internal capabilities bound after ClaimIntent/ClaimDelivery; external inputs must not supply them), and the `idempotency_hash` semantics clarified in both `store.go` and the migration comment (delivery dedup uses `delivery_id`).

### 💬 Non-blocking note
- The `Test` required check was still `in_progress` at review time (long-running suite; not failed). `Lint`, `Vet`, `Build`, and all other required checks are green. Please confirm `Test` lands green before merge — the branch protection will hold the merge until it does.

Nice, focused iteration.

yujiawei (approved):
# Code Review — PR #582 (octo-server)

Reviewed at head SHA `eca67c3abe7da4733a3bb0295510937b3620f86c` against merge-base `5e336a0c` on `main`. Scope: the resource-share platform foundation (slice 1) — immutable provider registry, signed-intent + ES256 detached-JWS proof primitives, human target authorizer, durable delivery state machine, Redis distributed limits, the `POST /v1/resource-shares` endpoint, public proof JWKS, and the trust integration across push / search / offline projections. Build, `go vet`, and the unit suites for `internal/resourceshare`, `modules/resource_share`, `modules/cardtrust`, `modules/messages_search`, and `tools/lint-card-dispatch` all pass locally.

This review was produced by absorbing an independent multi-dimension adversarial pass (crypto/proof, replay/idempotency, delivery state machine, authorization, cross-file trust integration, limiter/transport, and spec compliance), each finding re-verified against the source, plus one cross-family long-context advisory pass. A second external adversarial leg was unavailable this run (tooling outage) and is treated as absent, not as tacit approval; its intended lane (line-level correctness) was covered by the primary pass.

## 1. Spec compliance

**Spec: ✅**

Checked against `.octospec/tasks/user-resource-share-card/brief.md`, slice 1 ("Platform foundation").

- **Missing — none.** All slice-1 deliverables are present: immutable startup registry (`NewRegistry`, `reviewedProviderSpecs()` returns `nil`), intent + proof primitives, `HumanTargetAuthorizer` for DM/group/thread, three durable stores, Redis-backed `RedisTargetLimiter`, the generic `POST /v1/resource-shares` endpoint with login-UID/Space equality binding, ES256 proof signing + public `GET /v1/resource-shares/proof-jwks`, metrics observer, and the `lint-card-dispatch` source guard naming `ShareService.shareTarget` as the sole reviewed human type-17 dispatch site.
- **Over-build — none.** Ships fail-closed: `DM_RESOURCE_SHARE_ENABLED` defaults off, zero providers enabled, no `octo/v2`, no `Action.Submit`/interactive cards, no cross-space, no runtime provider registration, no client SDK.
- **Deviation — none.** `from_uid` is unrepresentable (request body is `{intent}` only, `DisallowUnknownFields`); signed `actor_uid` is an equality assertion checked against the login UID and can never override it; proof signing uses a dedicated ES256 key, not the master key; the registry is compile-time immutable; existing generic human type-17 and Bot-OBO rejection are untouched.

## 2. Code quality

**Quality: Approved**

The security-critical surfaces are sound. Highlights that were verified rather than assumed:

- **Proof binding is complete and symmetric.** Signer and verifier reconstruct the identical `proofSigningInput{metadata, envelope}` via RFC 8785 JCS. The signed set binds the finalized envelope (incl. `plain` and `space_id`), actor, Space, provider, resource type/ID/revision, canonical target, and delivery ID. `validateUnsignedShareEnvelope` allow-lists exactly `{type, card_version, profile, card, plain, space_id}`; the verifier independently re-derives `plain == BuildPlain(card)`; JOSE header injection (alg/kid/jwk/nonce/crit) is rejected on both the intent and proof paths; P-256 curve/on-curve checks are enforced. DM targets canonicalize to an order-independent participant pair on both sign and verify.
- **Delivery state machine is safe.** `claimed → dispatching → sent` with the durable `BeginDispatch` write strictly before the transport boundary; any post-dispatch ambiguity (`send` error, `MarkSent` failure, store-confirm failure) becomes terminal `unknown` with no auto-retry; `ReclaimPreTransport` can only resurrect `rate_limited`/`failed` rows (never `dispatching`/`sent`) and only before intent expiry; each transition writes its audit row in the same transaction.
- **Authorization is fail-closed and space-exact.** DM requires both parties active in the exact Space; group verifies exact Space + normal status + active/non-deleted membership + mute + full-group posting restriction; thread inherits the parent-group checks then requires an active thread; all queries are parameterized and DB errors surface as `ErrTargetQuery` (never silently allowed).
- **Cross-file trust integration is fail-closed.** The proof is re-verified against observed sender/Space/canonical target on every server display surface; the DM recipient-view sender swap (`observedCardTrusted`) and the global-search peer reversal both produce the correct `(viewer, peer)` pair; kind mismatch prevents replaying a DM proof into a group/thread and vice-versa; missing context masks to `[卡片]` rather than trusting.
- **Limiter denial is read-only** (no bucket spend when any bucket is exhausted), fails closed on store/format errors, and enforces per-provider budgets ≤ global at construction.

1. **`internal/resourceshare/store.go` — `loadDeliveryByIdentity` / `ClaimDelivery` collision path.** On a `delivery_id` unique-constraint collision the existing row is returned without asserting `intent_id == intentID`. This is not exploitable — `delivery_id` already hashes `actor_uid`, so cross-actor collision is impossible and same-actor/same-resource-revision/same-target is the intended idempotent dedup — but an explicit `AND intent_id = ?` guard (or a post-load equality check) would make the per-intent binding defense-in-depth rather than implicit.
2. **`internal/resourceshare/service.go` — expired first-use.** When an already-expired intent is submitted for the first time, `ClaimIntent` persists the intent row before the `expired` check rejects the request; a later retry with the same nonce then takes the `readExpiredResults` branch and returns HTTP 200 with all-`failed` targets instead of the first call's 4xx. There is no delivery impact (no delivery rows are ever created, so nothing is started or resumed), but the status-code flip and the stray intent row are avoidable by checking `expired` before claiming.
3. **`internal/resourceshare/proof.go` — `JWKS()`.** The `copyP256PublicKey` result is used with the ok-flag discarded (`publicKey, _ := ...`). Keys are pre-validated in `NewProofVerifier` so this cannot currently emit a nil key, but a `continue` on `!ok` (or a comment stating the invariant) would keep the exported JWKS robust against future refactors.
4. **Test coverage — global-search DM proof path.** `card_p1_test.go` exercises group targets; the `search_global_messages` DM path (peer reversal via `wireChannelFromDoc` → `TargetFromChannel`) is correct but untested. A case asserting a DM share proof verifies for a participant in global search, and fails when rebound to a different target kind, would lock in this security-sensitive path.
5. **DM authorization vs. friendship/blocklist (spec-alignment).** The brief describes DM as "verify the authenticated actor and peer under the existing friendship/Space policy." `HumanTargetAuthorizer` checks Space membership only; bidirectional blacklist is still enforced downstream at the WuKongIM datasource (`getBlacklist` → `ExistBlacklist`), so a blocked send is dropped at transport and there is no leak. Mirroring that blacklist check in the authorizer would fail faster and make the audit outcome (`denied` vs `unknown`) more accurate.

## 3. Overall verdict

**APPROVE**

Spec ✅ and no P0/P1 quality blockers. The five items above are P2 hardening / test-coverage improvements and, because the feature ships behind an off-by-default flag with zero enabled providers, do not need to gate this foundation PR — though items 1, 2, and 4 are worth resolving before the first provider is enabled.

PR #583, opened 2026-07-14
fix(group): durable async reconcile for orphan group channels #394

### LINKED ISSUE #394: CreateGroup compensating delete is non-atomic → orphan group rows (needs async reconcile)

lml2468:
## Summary

`group.CreateGroup` commits the group + members, then creates the WuKongIM channel **after** commit. On IM failure it runs a best-effort compensating delete — but that delete is itself non-atomic and only logs on failure, so a DB blip (or a crash between `tx.Commit()` and the IM create) can leave a **group row with no backing IM channel**: an orphan clients can never message. Fixes #394.

The IM channel lives outside the SQL transaction and can't be enrolled in it, so a point fix can't close the window. This PR adds a **transactional outbox + async reconcile**:

- New table `group_channel_outbox`. `CreateGroup` writes one row **inside the same tx** as `group`/`group_member` (atomic with the commit) and deletes it only after the IM channel is confirmed created. A surviving row means "channel not yet confirmed".
- A per-minute reconcile job (`channelReconcileScheduler`) scans outbox rows older than a 2-min grace period and, for each, **idempotently re-creates the channel from the group's current members** (recover-first — never deletes a member-bearing group), or cleans up a stale row whose group no longer exists. All outcomes are logged (`found`/`resolved`/`failed`).

Both failure modes converge to eventual consistency; the compensating delete is kept as the fast path and the outbox is the durable backstop.

## Scope

- Shared `group.CreateGroup` — Web / Bot / integration all benefit.
- Existing groups never get an outbox row and are never touched. No other insert path (system-group / org-dept / `AddGroup`) is modified.

## Verification

- New regression tests (`reconcile_test.go`) cover: recover orphan, retry-while-IM-down → heal, stale-outbox cleanup when group gone, and grace-window skip — all with an injected IM stub (no live WuKongIM needed).
- Full `./modules/group/` suite passes under `-race -shuffle=on` against real MySQL 8.0 + Redis + WuKongIM `v2.2.4-20260313` (CI-equivalent stack), which also exercises the `CreateGroup` happy path (outbox insert → IM create → outbox delete) end-to-end. No new failures.

## Deployment

- Requires the additive migration `20260714000001_group_channel_outbox.sql` (recovery-safe, `INFORMATION_SCHEMA`-guarded). No backfill, no config, no feature flag. Backward compatible.

🤖 Draft for review squad (qa + security + code).

### DIFF
--- a/modules/group/api.go
+++ b/modules/group/api.go
@@ -75,6 +75,14 @@ func New(ctx *config.Context) *Group {
 	g.ctx.AddEventListener(event.OrgOrDeptEmployeeUpdate, g.handleOrgOrDeptEmployeeUpdate)
 	g.ctx.AddEventListener(event.OrgEmployeeExit, g.handleOrgEmployeeExit)
 	source.SetGroupMemberProvider(g)
+
+	// #394: 启动群 IM 频道 reconcile 定时任务，兜底 CreateGroup 补偿删除失败 / 崩溃留下的孤儿群。
+	// 生产环境自启；测试环境(Test)不起 cron，由测试直接驱动 reconcileOnce。
+	if !g.ctx.GetConfig().Test {
+		if err := newChannelReconcileScheduler(g.ctx).Start(); err != nil {
+			g.Error("failed to start group channel reconcile scheduler", zap.Error(err))
+		}
+	}
 	return g
 }
 
--- a/modules/group/db_channel_outbox.go
+++ b/modules/group/db_channel_outbox.go
@@ -1,0 +1,72 @@
+package group
+
+import (
+	"time"
+
+	"github.com/gocraft/dbr/v2"
+)
+
+// ChannelOutboxModel 群 IM 频道创建事务发件箱行（表 group_channel_outbox）。
+//
+// 一行 == 「某 group_no 的 IM 频道尚未确认创建」。建群事务内写入，随 group/group_member
+// 一起原子提交；IM 频道确认成功后删除。见迁移 20260714000001_group_channel_outbox.sql 与
+// reconcile.go。收口 #394 的孤儿群窗口。
+type ChannelOutboxModel struct {
+	Id          int64
+	GroupNo     string
+	ChannelType int
+	Attempts    int
+	LastError   string
+	CreatedAt   time.Time
+	UpdatedAt   time.Time
+}
+
+// insertChannelOutboxTx 在建群事务内写入一条待确认频道的 outbox 行。
+//
+// 只显式写 group_no / channel_type，其余列(attempts/last_error/时间戳)交给表默认值，
+// 避免把 Go 零值时间写进 TIMESTAMP 列。group_no 唯一键；建群 group_no 是新 UUID 不会撞键，
+// 但仍用 ON DUPLICATE KEY 兜底重试幂等（重复建同 group_no 时不报错、不重置退避计数）。
+func (d *DB) insertChannelOutboxTx(groupNo string, channelType uint8, tx *dbr.Tx) error {
+	_, err := tx.InsertBySql(
+		"INSERT INTO group_channel_outbox (group_no, channel_type) VALUES (?, ?) "+
+			"ON DUPLICATE KEY UPDATE channel_type=VALUES(channel_type)",
+		groupNo, int(channelType),
+	).Exec()
+	return err
+}
+
+// deleteChannelOutbox 删除某 group_no 的 outbox 行（IM 频道已确认 / 群行已不存在的清理）。
+// 幂等：行不存在时删 0 行、不报错。
+func (d *DB) deleteChannelOutbox(groupNo string) error {
+	_, err := d.session.DeleteFrom("group_channel_outbox").Where("group_no=?", groupNo).Exec()
+	return err
+}
+
+// listDueChannelOutbox 返回 updated_at 早于 before 的 outbox 行（超过宽限期、可 reconcile 的），
+// 按 updated_at 升序（最久未处理优先），最多 limit 行。宽限期让 in-flight 建群（尚在 tx.Commit 与
+// IM 创建之间的正常窗口）不被误当孤儿处理。
+func (d *DB) listDueChannelOutbox(before time.Time, limit int) ([]*ChannelOutboxModel, error) {
+	var models []*ChannelOutboxModel
+	_, err := d.session.Select("*").
+		From("group_channel_outbox").
+		Where("updated_at < ?", before).
+		OrderDir("updated_at", true).
+		Limit(uint64(limit)).
+		Load(&models)
+	return models, err
+}
+
+// markChannelOutboxAttempt 记一次失败尝试：attempts+1、写 last_error、并借 ON UPDATE
+// CURRENT_TIMESTAMP 推进 updated_at（既是观测，也让下一轮 due 判定按最近尝试时间退避，
+// 避免同一坏行每个 tick 都被打）。lastErr 截断到列宽 512。
+func (d *DB) markChannelOutboxAttempt(groupNo, lastErr string) error {
+	if len(lastErr) > 512 {
+		lastErr = lastErr[:512]
+	}
+	_, err := d.session.Update("group_channel_outbox").
+		Set("attempts", dbr.Expr("attempts+1")).
+		Set("last_error", lastErr).
+		Where("group_no=?", groupNo).
+		Exec()
+	return err
+}
--- a/modules/group/reconcile.go
+++ b/modules/group/reconcile.go
@@ -1,0 +1,160 @@
+package group
+
+import (
+	"sync"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+	"github.com/robfig/cron/v3"
+	"go.uber.org/zap"
+)
+
+// #394 孤儿群 reconcile 调参。
+const (
+	// reconcileCronExpr reconcile 扫描频率（每分钟）。孤儿是低频事件，分钟级足够把
+	// 「有 group 行、无 IM 频道」的窗口收敛到最终一致，同时不给 DB/IM 压力。
+	reconcileCronExpr = "@every 1m"
+
+	// reconcileGrace 宽限期：只处理 updated_at 早于 now-grace 的 outbox 行。正常建群在
+	// tx.Commit() 与 IM 创建之间会短暂留有 outbox 行（随即被删），宽限期保证这些 in-flight
+	// 行不被误当孤儿重复建频道。也是失败行的退避基准（markChannelOutboxAttempt 推进 updated_at）。
+	reconcileGrace = 2 * time.Minute
+
+	// reconcileBatch 单轮最多处理的 outbox 行数，防止积压时一轮打爆 IM/DB；剩余的下一轮继续。
+	reconcileBatch = 200
+)
+
+// channelReconciler 扫描 group_channel_outbox，为「已提交但 IM 频道未确认」的群幂等重建频道
+// （或清理已消失群的陈旧 outbox 行），收口 #394 的孤儿群窗口。
+//
+// createChannel / subscribersOf 以字段注入，生产走真实 WuKongIM + 成员表；测试可注入桩，
+// 无需 live WuKongIM 即可覆盖两种失败模式的恢复路径。
+type channelReconciler struct {
+	log.Log
+	db            *DB
+	grace         time.Duration
+	batch         int
+	createChannel func(*config.ChannelCreateReq) error
+	subscribersOf func(groupNo string) ([]string, error)
+}
+
+func newChannelReconciler(ctx *config.Context) *channelReconciler {
+	db := NewDB(ctx)
+	return &channelReconciler{
+		Log:           log.NewTLog("GroupChannelReconciler"),
+		db:            db,
+		grace:         reconcileGrace,
+		batch:         reconcileBatch,
+		createChannel: ctx.IMCreateOrUpdateChannel,
+		subscribersOf: db.querySubscribableMemberUIDsWithGroupNo,
+	}
+}
+
+// reconcileOnce 处理一批到期 outbox 行。返回本轮 (found 扫到, resolved 已收口, failed 本轮未成功)。
+// 全程幂等，可安全重复运行 / 多实例并发（IM 创建是 upsert、删行幂等、宽限期挡住 in-flight）。
+func (r *channelReconciler) reconcileOnce() (found, resolved, failed int, err error) {
+	due, err := r.db.listDueChannelOutbox(time.Now().Add(-r.grace), r.batch)
+	if err != nil {
+		return 0, 0, 0, err
+	}
+	found = len(due)
+	for _, o := range due {
+		g, qErr := r.db.QueryWithGroupNo(o.GroupNo)
+		if qErr != nil {
+			r.Error("reconcile: query group failed", zap.Error(qErr), zap.String("groupNo", o.GroupNo))
+			_ = r.db.markChannelOutboxAttempt(o.GroupNo, qErr.Error())
+			failed++
+			continue
+		}
+		if g == nil {
+			// group 行已不存在：补偿删除成功但 outbox 删除失败，或群随后被解散。
+			// 无需 IM 频道，清理陈旧 outbox 行即为收口。
+			if delErr := r.db.deleteChannelOutbox(o.GroupNo); delErr != nil {
+				r.Error("reconcile: delete stale outbox failed", zap.Error(delErr), zap.String("groupNo", o.GroupNo))
+				failed++
+				continue
+			}
+			resolved++
+			continue
+		}
+
+		subs, sErr := r.subscribersOf(o.GroupNo)
+		if sErr != nil {
+			r.Error("reconcile: load subscribers failed", zap.Error(sErr), zap.String("groupNo", o.GroupNo))
+			_ = r.db.markChannelOutboxAttempt(o.GroupNo, sErr.Error())
+			failed++
+			continue
+		}
+
+		if cErr := r.createChannel(&config.ChannelCreateReq{
+			ChannelID:   o.GroupNo,
+			ChannelType: common.ChannelTypeGroup.Uint8(),
+			Subscribers: subs,
+		}); cErr != nil {
+			// IM 仍不可用：记一次尝试并保留 outbox 行，下一轮按退避重试（最终一致）。
+			_ = r.db.markChannelOutboxAttempt(o.GroupNo, cErr.Error())
+			r.Warn("reconcile: recreate IM channel failed, will retry",
+				zap.Error(cErr), zap.String("groupNo", o.GroupNo), zap.Int("attempts", o.Attempts+1))
+			failed++
+			continue
+		}
+
+		if delErr := r.db.deleteChannelOutbox(o.GroupNo); delErr != nil {
+			// 频道已建好但删行失败：下一轮会幂等重建频道(无副作用)再重试删行。
+			r.Error("reconcile: delete outbox after channel recreate failed", zap.Error(delErr), zap.String("groupNo", o.GroupNo))
+			failed++
+			continue
+		}
+		r.Info("reconcile: recovered orphan group channel", zap.String("groupNo", o.GroupNo), zap.Int("subscribers", len(subs)))
+		resolved++
+	}
+	if found > 0 {
+		r.Info("group channel reconcile sweep done",
+			zap.Int("found", found), zap.Int("resolved", resolved), zap.Int("failed", failed))
+	}
+	return found, resolved, failed, nil
+}
+
+// channelReconcileScheduler 定时驱动 channelReconciler（仿 opanalytics/backup 的 cron 调度）。
+type channelReconcileScheduler struct {
+	log.Log
+	reconciler *channelReconciler
+	cron       *cron.Cron
+	entryID    cron.EntryID
+	mu         sync.Mutex
+	started    bool
+}
+
+func newChannelReconcileScheduler(ctx *config.Context) *channelReconcileScheduler {
+	return &channelReconcileScheduler{
+		Log:        log.NewTLog("GroupChannelReconcileScheduler"),
+		reconciler: newChannelReconciler(ctx),
+		cron:       cron.New(),
+	}
+}
+
+// Start 启动调度器（幂等）。出错只返回 error，由调用方记日志，不 panic。
+func (s *channelReconcileScheduler) Start() error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.started {
+		return nil
+	}
+	entryID, err := s.cron.AddFunc(reconcileCronExpr, s.runOnce)
+	if err != nil {
+		return err
+	}
+	s.entryID = entryID
+	s.cron.Start()
+	s.started = true
+	s.Info("group channel reconcile scheduler started", zap.String("cron", reconcileCronExpr))
+	return nil
+}
+
+func (s *channelReconcileScheduler) runOnce() {
+	if _, _, _, err := s.reconciler.reconcileOnce(); err != nil {
+		s.Error("group channel reconcile tick failed", zap.Error(err))
+	}
+}
--- a/modules/group/reconcile_test.go
+++ b/modules/group/reconcile_test.go
@@ -1,0 +1,196 @@
+package group
+
+import (
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/Mininglamp-OSS/octo-lib/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+// Issue #394 — CreateGroup 的补偿删除是非原子的：IM 频道创建失败后，若「已提交的 group/
+// group_member 补偿删除」自身也失败（DB 抖动），或进程崩在 tx.Commit() 与 IM 创建之间，就会
+// 留下「有 group 行、无 IM 频道」的孤儿群，客户端永远无法进入。
+//
+// 修复：建群事务内写 group_channel_outbox 行（原子提交），IM 频道确认后删除；后台 reconcile
+// 定时扫描超过宽限期仍存在的行，用当前成员幂等重建频道 —— 或清理 group 行已消失的陈旧 outbox 行。
+// 下列用例用注入桩（无需 live WuKongIM）覆盖验收要求的两种失败模式及其最终一致收口。
+
+// seedOrphanGroup 直接落一个「已提交 + 待确认频道」的群：group 行、一名正常成员、以及一条
+// outbox 行 —— 即 IM 创建失败(补偿删除也失败) 或 commit 后崩溃所留下的孤儿状态。
+func seedOrphanGroup(t *testing.T, ctx *config.Context, groupNo string, members ...string) {
+	t.Helper()
+	db := NewDB(ctx)
+	require.NoError(t, db.Insert(&Model{
+		GroupNo:        groupNo,
+		Name:           "orphan-" + groupNo,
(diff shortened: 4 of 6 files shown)


OctoBoooot (commented):
## Review: fix(group): durable async reconcile for orphan group channels (#583)

**Verdict: Approve with comments** — a clean, correct transactional-outbox + async-reconcile design that actually closes the #394 orphan window (group row with no backing IM channel). No blockers, no majors. Held at COMMENT because it's a **draft**: Build/Test/Vet/Lint are all SKIPPED, so the (very good) new reconcile tests have never executed — mark ready to run them and this converts to a clean Approve.

Risk: this is durability/concurrency/reconciliation code — the class where subtle state-machine bugs hide — so I verified the atomicity, idempotency, and grace-window boundary directly rather than trusting the shape. `needs-human-review` + `size/XL` already applied.

### What I byte-verified
- **The outbox insert is genuinely in-tx.** `service.go` `CreateGroup` calls `insertChannelOutboxTx(groupNo, …, tx)` **before** `tx.Commit()`, so the outbox row commits atomically with `group`/`group_member`. The ordering is correct: commit → `IMCreateOrUpdateChannel` → on success `deleteChannelOutbox`; on IM failure, the compensating group-delete runs but the outbox row is deliberately **kept** for reconcile. "Row exists ⇒ channel not yet confirmed" holds by construction.
- **Reconcile is recover-first and idempotent.** For each due row: if the group is gone → clean the stale row (no channel); else re-create the channel from **current** members via `IMCreateOrUpdateChannel` (an upsert) then delete the row. It never deletes a member-bearing group, IM create is idempotent, and row-delete is idempotent — so repeated/overlapping runs and the "channel created but delete failed" case all converge (next sweep re-upserts harmlessly and retries the delete).
- **Grace window protects in-flight creates.** `listDueChannelOutbox` filters `updated_at < now-grace` (2 min), and `markChannelOutboxAttempt` advances `updated_at` (via `ON UPDATE CURRENT_TIMESTAMP`) so a persistently-failing row backs off instead of being hammered every tick. The test `…SkipsInFlightWithinGrace` pins this.
- **Migration is recovery-safe.** `INFORMATION_SCHEMA`-guarded `CREATE TABLE` via a stored proc, idempotent, with a matching guarded Down. `uk_group_no` unique + `idx_updated_at` back the dedup and the due-scan.
- **Tests are mutation-resistant.** `reconcile_test.go` covers recover-orphan (asserts exact subscriber set + group row preserved), retry-while-IM-down → heal (asserts `attempts` increments then resolves), stale-outbox cleanup (asserts no channel created), and grace-window skip. All symbols resolve (`querySubscribableMemberUIDsWithGroupNo` exists; helpers defined) so the suite compiles. Removing any branch flips a test.

### Minor
- `reconcile.go:` `cron.New()` is constructed without `cron.WithChain(cron.SkipIfStillRunning(...))`. robfig/cron v3 runs each entry in its own goroutine and **allows overlap** — if a sweep ever exceeds the 1-min interval (backlog against a slow IM), a second `reconcileOnce` starts concurrently. It's safe (idempotent upsert + idempotent delete), but overlapping sweeps double-process the same due rows and can inflate `attempts` by more than 1/interval. Adding `SkipIfStillRunning` makes the cadence predictable.
- The scheduler has `Start()` but no `Stop()`, and the `entryID`/`started` fields suggest one was intended. Not a correctness issue (process-lifetime job), but a `Stop()` (cron.Stop + reset `started`) would make it testable and clean on shutdown. Consistent with the opanalytics pattern it mirrors, so optional.

### Questions
- **Empty-subscriber channel.** If a due group's members have all left (`querySubscribableMemberUIDsWithGroupNo` returns `[]`), reconcile calls `createChannel` with `Subscribers: []` and then resolves (deletes the row). Is an empty-subscriber `IMCreateOrUpdateChannel` a valid no-op that satisfies "channel exists," or should a memberless orphan instead be treated like the group-gone case (clean the row without creating an empty channel)? A test for the zero-subscriber path would settle it. _(confidence: high this path is reachable; behavior depends on WuKongIM's empty-channel semantics I can't see here.)_
- **Subscriber-set parity vs the original create.** The first-time `CreateGroup` IM create uses `realMemberUIDs` (built during creation, includes the bot UID when present); reconcile uses the current `group_member` table (`is_deleted=0 AND status=normal`). These are usually equivalent, but if a bot member isn't a normal `group_member` row, a reconciled channel could omit it. Worth confirming the reconcile subscriber query is the intended source of truth for a repaired channel (I think it is — current-state repair — but flag it as a conscious choice).
- Multi-instance: the reconcile claims safe concurrent operation across replicas. Verified the ops are idempotent, so no corruption — but N replicas each run the full sweep every minute (N× the IM upserts). If deployments run multiple instances, consider a lightweight lease/advisory-lock so only one sweeps. Not a bug; a cost note.

### Praise
- The transactional-outbox framing is exactly right for "a side effect that can't join the SQL transaction": the row commits atomically with the group, and its *presence* is the durable signal. Recover-first (never delete a member-bearing group), idempotent re-create, and the grace window together make both failure directions (compensating-delete-failed orphan, and stale-row-after-successful-cleanup) converge to eventual consistency — and the test suite exercises each direction with an injected IM stub so it needs no live WuKongIM. That's a well-engineered durability fix, not a point patch.

### Note on merge gate
Draft + `needs-human-review` → `mergeStateStatus` BLOCKED. Approved ≠ mergeable: mark ready so CI runs the reconcile tests green, then a human maintainer signs off (new background scheduler + a shared `CreateGroup` path touch warrant it).

Jerry-Xin (commented):
**Review: durable async reconcile for orphan group channels (#394)** — approve-with-comments.

The transactional-outbox + async-reconcile design is the right fix for the "committed group / no IM channel" window: the outbox row is written **inside the same tx** as `group`/`group_member` (`modules/group/service.go:1294`), deleted only after the IM channel is confirmed (`service.go:1364`), and the reconcile loop is recover-first (never deletes a member-bearing group). Idempotency is sound end-to-end — IM create is an upsert (`IMCreateOrUpdateChannel` POSTs `/channel`), row delete is a no-op when absent, and the 2-min grace window keeps in-flight creations from being reprocessed. Tests cover all four modes (recover / retry-while-down→heal / stale-cleanup / grace-skip). No data race in the worker: every `channelReconciler` field is set once at construction and only read in `reconcileOnce`; `channelReconcileScheduler.started`/`entryID` are mutex-guarded.

Marking COMMENT rather than APPROVE only because the PR is a **draft** → all code checks (Vet/Test/Build/Lint) are SKIPPED in CI, so the suite has not actually run here. The design itself is approvable; please flip out of draft so the `-race` suite runs, then this can go to APPROVE.

🟡 Non-blocking

1. **No panic recovery in the cron job.** `modules/group/reconcile.go:159` (`runOnce`) calls `reconcileOnce()` with no `recover()`, and the scheduler uses a bare `cron.New()` (`reconcile.go:134`) with no `cron.WithChain(cron.Recover(...))`. robfig/cron v3 does not recover panicking jobs by default, so an unexpected panic on one row (nil deref, etc.) would take down the reconcile goroutine and silently stop all future sweeps. The per-row `continue` handles *errors* but not *panics*. Consider `cron.New(cron.WithChain(cron.Recover(logger)))` or a `defer recover()` in `runOnce`.

2. **Overlapping sweeps possible under backlog.** `@every 1m` with a batch of up to 200 sequential network IM calls (`reconcileBatch`, `reconcile.go:31`) can exceed 60s; cron v3 (no `SkipIfStillRunning`) then launches a second `reconcileOnce` concurrently. It is safe (fully idempotent, and `updated_at`/grace de-dupes most rows), just wasteful — two workers may double-attempt the same due rows. `cron.WithChain(cron.SkipIfStillRunning(...))` would make each tick exclusive. Same reasoning applies to the multi-instance case: there is no row claim/lock/CAS, so N app instances all reconcile the same orphan; correct due to idempotency, but O(N) redundant IM calls. Fine at current scale — flag if this scheduler ever runs hot.

3. **Reconcile does not check group `Status`.** The loop only branches on `g == nil` (`reconcile.go:75`); a group with `Status == GroupStatusDisband` that still has an outbox row would get its IM channel recreated. This is an extreme edge (the outbox row is normally deleted seconds after creation, long before any disband), but if you want to be strict, treat a disbanded group like the `g == nil` case (clean the stale row, skip channel recreate).

✅ Highlights
- Outbox insert is atomic with the commit and existing groups never get a row → zero blast radius on legacy data; no other insert path touched.
- Migration `modules/group/sql/20260714000001_group_channel_outbox.sql` is `INFORMATION_SCHEMA`-guarded and re-entrant (recovery-safe), with a matching Down.
- `insertChannelOutboxTx` leaves timestamps to column defaults (avoids writing Go zero-time) and uses `ON DUPLICATE KEY` so retries don't reset backoff; `markChannelOutboxAttempt` truncates `last_error` to the 512 column width and rides `ON UPDATE CURRENT_TIMESTAMP` for backoff — thoughtful details.
- Subscribers are rebuilt from `querySubscribableMemberUIDsWithGroupNo` (normal, non-deleted members), consistent with the live-membership source rather than a stale snapshot.
---
**Added 🟡 (byte-verified, credit to the reviewer who flagged the UTF-8 truncation):**
`modules/group/db_channel_outbox.go:64` truncates the error string by **bytes** — `lastErr = lastErr[:512]` — before writing to `last_error VARCHAR(512)`. Byte-slicing can land mid-rune on a multi-byte (non-ASCII) error message, producing invalid UTF-8. Against the `utf8mb4` column this either (a) fails the UPDATE with MySQL 1366 "Incorrect string value" under strict mode — so the very attempt-recording write that is supposed to log the failure itself fails, and `attempts`/backoff never advance — or (b) silently stores garbage under non-strict mode. Non-blocking (only triggers on long non-ASCII errors, and it degrades error-recording rather than the happy path), but worth a rune-safe truncation. Fix mirrors the pattern already used for group name at service.go:1153:
`if r := []rune(lastErr); len(r) > 512 { lastErr = string(r[:512]) }` (or truncate on a valid UTF-8 boundary via utf8.RuneStart / utf8.DecodeLastRuneInString).

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #583 (octo-server)

## Summary

This PR adds a durable async reconcile mechanism for orphan group channels in octo-server. The core problem (issue #394): `CreateGroup` commits the group row in MySQL, then creates the IM channel in WuKongIM — a cross-system operation that can't be atomic. If IM creation fails and the compensating delete also fails (or the process crashes between commit and IM call), an "orphan group" is left: the group row exists but has no IM channel, making the group permanently unusable for clients.

The fix introduces a transactional outbox pattern: a `group_channel_outbox` row is written atomically with the group in the same transaction, recording the intent to create an IM channel. After IM confirmation, the row is deleted. A background cron (every 1 minute) scans for stale outbox rows past a 2-minute grace period and either rebuilds the IM channel idempotently (group still exists) or cleans up the stale row (group already gone). The design is sound — idempotent, multi-instance safe (IM upsert + idempotent delete), and the grace period prevents interference with in-flight group creation.

## Verification

Static analysis only at head `e24ece35a4fad06f20b3124e0bade7b7fa3e90b5`; build and tests not executed in this environment.

- ✅ **Outbox atomicity** — `insertChannelOutboxTx` (`modules/group/service.go:1291`) writes within the same `tx` as group/group_member inserts, committed together at `:1298`. If the insert fails, `CreateGroup` fails before commit — no orphan state possible.
- ✅ **Grace period correctness** — `listDueChannelOutbox` (`modules/group/db_channel_outbox.go:55`) filters `updated_at < before` where `before = now - 2min`. Freshly inserted rows have `updated_at ≈ now`, so they won't be picked up until the grace window expires. Verified by `Test_Issue394_ReconcileSkipsInFlightWithinGrace`.
- ✅ **Idempotent IM recreate** — `IMCreateOrUpdateChannel` is documented as upsert at `modules/group/api.go:339`. Multiple reconcile ticks or multi-instance races calling it for the same `groupNo` produce no duplicate side effects.
- ✅ **Compensating delete path** — When IM creation fails (`modules/group/service.go:1348`), the code deletes group_member and group rows but deliberately leaves the outbox row. Reconcile handles both outcomes: group gone → clean stale outbox; group still there → rebuild channel. Verified by `Test_Issue394_ReconcileCleansStaleOutboxWhenGroupGone`.
- ✅ **Migration safety** — `20260714000001_group_channel_outbox.sql` uses `INFORMATION_SCHEMA` guard for idempotent CREATE TABLE, and `DROP PROCEDURE` cleanup is in both Up and Down paths.
- ✅ **Subscriber data flow** — `querySubscribableMemberUIDsWithGroupNo` (`modules/group/db.go:530`) filters `is_deleted=0 AND status=GroupMemberStatusNormal`, which is the same production subscriber source used by `service.go:560` and `event.go:185`. Data flows correctly from member table to IM channel creation.
- ✅ **Test coverage** — Four test cases cover: successful recovery, retry-while-IM-down with eventual consistency, stale outbox cleanup when group is gone, and grace period protection for in-flight groups. Tests use real DB via `newTestServer` with injectable IM/subscriber stubs.

## Findings

No P0/P1 issues. One P2 and two nits below.

### P2 — Byte-level UTF-8 truncation can produce invalid `last_error` (`modules/group/db_channel_outbox.go:66`)

`lastErr[:512]` truncates at byte offset 512 on a Go string (UTF-8 byte sequence). If the error message contains multi-byte runes (CJK, emoji) near that boundary, the result is an invalid UTF-8 sequence. The `last_error` column is `VARCHAR(512) ... utf8mb4` — MySQL strict mode (default since 5.7) would reject the UPDATE with `Incorrect string value`. Since `markChannelOutboxAttempt` errors are discarded (`_ =`) at all three call sites in `reconcile.go:70`, `:80`, `:95`, the row's `updated_at` never advances, losing the intended backoff: the row stays "due" and gets retried every 1-minute tick indefinitely.

Low probability in practice — Go error strings are typically ASCII. Fix: use rune-aware truncation or validate with `utf8.ValidString` before slicing:

```go
if len(lastErr) > 512 {
    lastErr = string([]rune(lastErr)[:512])
}
```

**Diff-scope**: new code introduced by this PR. Not pre-existing.

### Nit — `markChannelOutboxAttempt` errors silently discarded (`modules/group/reconcile.go:70`, `:80`, `:95`)

All three call sites use `_ = r.db.markChannelOutboxAttempt(...)`. This is defensible — if the DB is down, the row stays due and gets retried on recovery (fast convergence). But it means a persistent UPDATE failure (e.g., schema drift, permission change) would go unobserved beyond the reconcile-level error log. Consider logging at Warn level when the mark itself fails, to distinguish "IM failed, mark succeeded (backoff active)" from "IM failed, mark also failed (no backoff)".

### Nit — Scheduler reference dropped, no graceful shutdown (`modules/group/api.go:82-85`)

`newChannelReconcileScheduler(g.ctx).Start()` creates the scheduler but the reference is not retained — there's no `Stop()` path. The cron goroutine runs until process exit. For typical "run until killed" server deployments this is fine, but a hot-reload or test scenario that re-invokes `New()` would leak cron instances. Low urgency.

## Blindspot Checklist (C1–C6)

- **C1 (dual-path parity)**: N/A — no paired add/remove or subscribe/unsubscribe paths changed.
- **C2 (control-flow ordering)**: Clear — `insertChannelOutboxTx`, `deleteChannelOutbox`, and `markChannelOutboxAttempt` are each called with consistent semantics across all sites.
- **C3 (authorization boundary)**: N/A — no auth/permission changes.
- **C4 (auth lifecycle)**: N/A — no auth changes.
- **C5 (build ≠ runtime)**: N/A — no build/CI changes.
- **C6 (governance/docs)**: N/A — no governance/docs changes.

## Data Flow Trace

- **Outbox row creation**: `insertChannelOutboxTx` → same `tx` as group/member inserts → `tx.Commit()` at `service.go:1298`. Atomic with group creation. ✅
- **Outbox row deletion (happy path)**: `IMCreateOrUpdateChannel` succeeds at `:1340` → `deleteChannelOutbox` at `:1366`. If delete fails, logged but non-blocking; reconcile handles retry. ✅
- **Outbox row deletion (IM failure path)**: IM fails → compensating delete of group_member + group → outbox row intentionally preserved → reconcile picks it up. ✅
- **Reconcile subscriber list**: `querySubscribableMemberUIDsWithGroupNo` → `group_member` WHERE `is_deleted=0 AND status=normal` → fed to `createChannel`. Same source as production IM channel creation. ✅
- **Reconcile grace period**: `listDueChannelOutbox(now - 2min, 200)` → `WHERE updated_at < before` → `ORDER BY updated_at ASC LIMIT 200`. Fresh rows excluded; oldest due rows processed first. ✅

## Cross-round Blocker Recheck (R6)

## Verdict: APPROVED

Well-designed transactional outbox pattern with proper idempotency, grace period protection, and comprehensive test coverage. The P2 UTF-8 truncation issue is low-probability but worth fixing to prevent a silent backoff-loss edge case. Neither nit is a merge concern.

[Octo-Q] verdict: APPROVE — no P0/P1 found; P2 (byte-level UTF-8 truncation) is low-probability and non-blocking; tests cover all key reconcile paths.

yujiawei (approved):
# Code Review — PR #583 (octo-server)

## Summary

This PR closes the orphan-group-channel window in #394 with a **transactional outbox + async reconcile** pattern. The design is sound and the implementation is correct on all the paths that matter for data integrity. I verified the mechanism end-to-end against the surrounding code and did not find any correctness, security, or data-loss blocker. Approving, with a set of P2 follow-ups worth a look before or shortly after merge.

## What was verified (clean)

- **Atomicity of intent.** `insertChannelOutboxTx` runs on `tx` inside `CreateGroup` (`service.go:1290`); on failure it returns before `tx.Commit()`, so `defer tx.RollbackUnlessCommitted()` (`service.go:1168`) rolls back group + members + outbox together. No partial state.
- **Happy path.** Outbox row is written in-tx, deleted after the IM channel is confirmed (`service.go:1364`); delete failure is non-fatal and self-heals on the next sweep.
- **Compensating-delete path.** The outbox row is intentionally *retained* when the IM create fails, so reconcile can either recreate the channel (compensation failed → orphan) or clean the stale row (compensation succeeded → group gone). Both converge.
- **Reconcile idempotency.** Channel recreate is an upsert, outbox delete is idempotent, and `ON DUPLICATE KEY` guards the insert. Re-running a sweep — or running two sweeps concurrently — cannot corrupt state.
- **Grace window.** Fresh rows carry `updated_at = now` and are excluded for 2 min, so in-flight `CreateGroup` transactions are not mistaken for orphans.
- **Scope.** Only `CreateGroup` is touched; pre-existing groups never get an outbox row, so they are never visited by reconcile. Other channel-create paths (`event.go:185`, `event.go:360`) keep their in-tx create+rollback and are correctly left alone.
- **Channel flags.** Reconcile omits `Ban`/`Large`, matching the original `CreateGroup` create call — those are supplied via the `IMDatasource.ChannelInfo` pull (`1module.go`), so no behavioral drift.
- **Migration.** Additive, `INFORMATION_SCHEMA`-guarded, recovery-safe, and consistent with the repo's existing procedure-guarded migration convention. `group_no VARCHAR(40)` matches the `group` table.

## Findings (all P2 / non-blocking)

**P2-1 — Reconcile runs on every replica with no distributed lock.**
`api.go:82` starts `channelReconcileScheduler` unconditionally on each instance. A sibling scheduler in this codebase (`modules/opanalytics/scheduler.go`) takes a Redis lock precisely to avoid multi-node duplicate execution. Here it is **not a correctness bug** — the recreate/delete operations are idempotent, so N replicas converge to the same result — but during a sweep all replicas will hit WuKongIM and contend on `markChannelOutboxAttempt` for the same rows. Because the steady-state orphan set is near-zero this is low-impact, but a lock (or leader election) would make the cost match `opanalytics` and remove the thundering-herd on any backlog. Consider aligning with the existing lock pattern.

**P2-2 — No timeout on the reconcile IM call; goroutines can accumulate during a total IM outage.**
`IMCreateOrUpdateChannel` → `network.Post` uses the sendgrid `rest` `DefaultClient`, which has **no HTTP timeout** (`&http.Client{}`). `cron/v3` launches each tick in its own goroutine and does not skip overlapping runs, so a sustained WuKongIM outage leaks roughly one blocked goroutine (and one held connection) per minute. The batch loop is sequential, so it's ~1 per tick rather than per row, and unbounded IM timeouts are a pre-existing property of the whole IM client — but the periodic caller makes the slow accumulation new. A context/timeout on the reconcile call, or a `cron.SkipIfStillRunning` wrapper, would bound it.

**P2-3 — Delete-failure paths don't advance `updated_at`, so no backoff on transient DB delete errors.**
In `reconcile.go`, both the stale-row cleanup (`reconcile.go:~78`) and the post-recreate delete (`reconcile.go:~103`) do `failed++; continue` without calling `markChannelOutboxAttempt`. The row's `updated_at` is unchanged, so it is immediately due again on the next tick. It is not a tight loop (bounded by the 1-min cron and the recreate is idempotent), but a flapping DB delete would reprocess the row every minute with no backoff and no attempt counter. Bumping `updated_at`/`attempts` on these paths too would make backoff uniform.

**P2-4 — No attempt cap / dead-letter; a permanently failing row is a poison pill.**
`attempts` is incremented but never used to cap or divert. A row that can never succeed (e.g. a permanently invalid subscriber payload) retries forever and occupies a slot in the `LIMIT 200` batch. With rare orphans this is fine, but a cap + alert (or DLQ state) would prevent one bad row from silently consuming a batch slot indefinitely.

**P2-5 — Reconcile does not skip disbanded groups.**
`QueryWithGroupNo` returns the row for a group whose `status = GroupStatusDisband` (disband sets status, it does not delete the row). If an orphan row survives long enough for the group to be disbanded, reconcile would recreate the `/channel` (subscribers) for a group that is meant to stay disbanded. Disband's own fail-closed `Disband=1` push is on channel-info and is idempotent, so the states converge rather than corrupt, and the scenario is extreme (orphan + disband-before-heal). Given the security-sensitive classification, a `status == Disband → clean the outbox row instead of recreating` guard is worth adding and is the one item I'd most want a human to eyeball.

**Nit — `channel_type` column is stored but ignored on recreate.** `reconcile.go:194` hardcodes `common.ChannelTypeGroup.Uint8()` instead of reading `o.ChannelType`. Harmless today (only group type is ever inserted), but the column is effectively dead in the recovery path.

## Testing

The four regression tests (`reconcile_test.go`) map cleanly onto the acceptance scenarios (recover orphan, retry-while-IM-down → heal, stale-outbox cleanup, grace-window skip) and use injected stubs so they run without a live WuKongIM. Coverage of the mechanism itself is good. The gaps above (multi-replica, disbanded-group, poison-pill) are not exercised, which is acceptable for this change but worth noting.

## Verdict

No P0/P1 issues. The core invariant — a surviving outbox row always converges to either a healed channel or a cleaned row, with no orphan and no data loss — holds. The P2 items are operational hardening, not merge blockers.

OctoBoooot (commented):
## Follow-up on #583 — two more non-blocking 🟡s (credit yujiawei), verdict unchanged (COMMENT, draft)

Byte-verified two additional findings yujiawei raised; both are real and correctly non-blocking. Verdict stays **Approve with comments** — still a draft (`isDraft: true` / `REVIEW_REQUIRED` / `BLOCKED`), so the reconcile `-race` tests have not run; not mergeable on a draft-time approval.

### 🟡 (new, verified) — delete-failure doesn't advance the backoff timestamp → hot-loop
`reconcile.go` — in **both** delete-fail branches (stale-group cleanup, and post-recreate `deleteChannelOutbox`), the code does `failed++; continue` **without** calling `markChannelOutboxAttempt`. Since `updated_at` only advances via `markChannelOutboxAttempt` (`ON UPDATE CURRENT_TIMESTAMP`), a row whose delete keeps failing stays `updated_at`-old and is re-selected by `listDueChannelOutbox` **every tick** with no backoff — a hot-loop (idempotent re-create each time, so no corruption, but it defeats the backoff design for that failure mode). Fix: call `markChannelOutboxAttempt` (or a lighter `touch updated_at`) on the delete-fail paths too, so a persistently-failing delete backs off like a failing create.

### 🟡 (new, verified) — IM call has no timeout
`createChannel = ctx.IMCreateOrUpdateChannel` takes no `context.Context`, so the sweep can't bound a hung WuKongIM call; the sweep goroutine blocks until the underlying HTTP client's own timeout (if any) fires. Combined with the earlier note that `cron.New()` has no `SkipIfStillRunning`, a stalled sweep can overlap with the next tick. (yujiawei framed this as "goroutine 积累" — precisely it's one stalled sweep goroutine plus possible tick-overlap, not per-row goroutine growth, since the loop is serial.) Non-blocking; the real fix is a timeout-bearing IM call, which is a broader signature change.

### Already on record
- Multi-instance no distributed lock (yujiawei's third 🟡) — matches my earlier note; N replicas each sweep, idempotent so safe, just N× IM upserts. Consider a lightweight lease if multi-instance.
- My prior 🟡s: no `SkipIfStillRunning` overlap guard; scheduler has no `Stop()`; empty-subscriber channel behavior (question); + mochashanyao's verified `last_error` byte-truncation-splits-a-rune 🟡.

### Path to merge (unchanged)
Design is sound and byte-verified with **no blockers**. Mark ready → `-race` CI green → human maintainer signs off (`needs-human-review`). The 🟡 pile is worth a cleanup pass but none blocks.

PR #580, opened 2026-07-14
fix(notify): reuse notification bot for summary cards

an9xyz:
## Summary

Reuse the existing `notification` User Bot for summary cards and their plain-text fallback, so users see one system-notification DM conversation instead of a separate `summary` Bot conversation.

This is a UX follow-up to #579. Producer capability isolation remains unchanged: summary cards still originate only through the bound `summary-notify` card dispatcher producer.

## Related Issue

Related to #571. Follow-up to #579.

## Linked Spec

- [Sender unification](.octospec/tasks/summary-notify-sender-unification/brief.md)
- [Future origin group/thread delivery](.octospec/tasks/smart-summary-origin-channel-card/brief.md)
- [Generic user resource-card sharing](.octospec/tasks/user-resource-share-card/brief.md)

## Changes

- Bind `summary-notify` to `NotifyBotUIDValue` and reuse the existing notification Bot readiness/provisioning path.
- Send summary-card text fallback from the same `notification` identity.
- Remove dedicated summary Bot constants, provisioning, and readiness state; do not destructively clean up any previously provisioned DB identity.
- Update unit, integration, source-wiring, and pilot E2E assertions for `from_uid=notification`.
- Amend the original dispatch contract and add production briefs for origin group/thread delivery and generic user resource-card sharing.

## Testing

- [x] Unit tests added/updated
- [ ] Manually verified

Commands run:

```text
go test -race -shuffle=on -count=1 -timeout 5m . ./modules/notify
go test -tags pilote2e ./pilote2e -run '^$'
go vet . ./modules/notify/... ./internal/carddispatch/...
go run ./tools/lint-card-dispatch ./modules ./internal
make i18n-extract-check
make i18n-lint
git diff --check origin/main...HEAD
```

The `pilote2e` command compiles the tagged suite without executing its environment-dependent tests.

## COMPREHENSION

1. **What does this change actually do** to the load-bearing path?

   It changes the immutable sender binding for the production `summary-notify` producer from the dedicated `summary` UID to the existing active `notification` User Bot. Card delivery, legacy text notifications, and card fallback now share one retryable readiness gate and one DM identity. Card validation, Space/target authorization, producer profile/channel restrictions, and the no-arbitrary-type-17 ingress boundary are unchanged.

2. **What could break** because of it (dependents + failure mode)?

[...]

PR #584, opened 2026-07-14
feat(notify): docs-notify card producer + reserved metadata namespace

an9xyz:
## Summary

Adds `docs-notify` as the **second `internal/carddispatch` producer**, giving
docs-backend the same trusted, server-templated card path that `summary-notify`
gave smart-summary (PRs #579/#580). Also introduces a stable
`metadata.octo.{variant,source}` namespace on every notification card so
renderers can style card families independently without re-parsing content, and
reserves a visible **source badge (角标)** slot for future "来自智能总结 / 来自文档"
styling. octo-server scope only; **DM (person) targets only**.

## Related Issue

Refs #571. Foundation: #577. Pilot: #579, #580 (merged).

## Linked Spec

- `.octospec/tasks/card-message-internal-dispatch/brief.md` (Decisions 1/3/4/11/14)
- `.octospec/tasks/card-message-internal-dispatch/docs-notify-contract.md` (new cross-repo wire contract for octo-docs-backend)
- `.octospec/tasks/card-message-internal-dispatch/summary-notify-contract.md` (sibling, points at the new contract)

## Changes

- **Template (`pkg/cardtmpl`)** — factor the shared octo/v1 body assembly into
  `buildResourceCard`; expose `BuildSummaryResourceCard` +
  `BuildDocsResourceCard` (docs deep-link `/d/{doc_id}?sp={space_id}`, mirroring
  `summaryDeepLink`). `ResourceCard` grows `Variant` and
  `Source{Label, IconURL}`. Every card now emits an AC 1.5 `metadata` object:
  `webUrl` (canonical URL) + `octo.{variant, source}`.
- **Ingress (`modules/notify`)** — new structured `NotifyReq.DocsCard`
  (`DocsCardFields`, **not** a type-17 map). `Payload` / `Card` / `DocsCard` are
  **mutually exclusive** (both/all-present → 400 `err.server.notify.card_invalid`;
  none → legacy 400). `deliverDocsCardNotification` reuses the exact
  dedup → actor-exclude → live member verify → bounded fan-out pipeline through
  the producer-bound `carddispatch.Sender`, preserving
  `NotifyResp{delivered,filtered}`.
- **Kinds & i18n** — `shared` / `commented` / `access_requested`, each mapping to
  a localized attribution (with/without actor) and a reserved
  `metadata.octo.variant`; zh-CN + en-US label tables; `Source.Label` =
  "智能总结" / "Smart Summary" and "文档" / "Docs".
- **Text fallback** — `buildDocsFallbackText` mirrors `buildSummaryFallbackText`
  so a card-build failure (feature off, sender missing, non-https deep-link
  origin) degrades to a plain DM instead of dropping the notification.
- **Wiring (`main.go`)** — `cardDispatchProducerSpecs` registers both
  `summary-notify` and `docs-notify` against the shared `notification` User Bot
  (same DM-only / `octo/v1` / system-notification / `MaxInFlight=20` shape). No
  `register.AddModule` change, no package-global registration (Decision 1/11).
  Capability isolation stays on the producer ID.
- **Config (`configs/tsdd.yaml`)** — the `external.webLoginURL` comment now
  states its role as the card deep-link origin; a missing/non-https value
  silently degrades card notifications to text.
- **Tooling** — `scripts/cardpreview` dumps the three canonical cards as AC JSON
  via the production code paths (used to drive octo-web's real renderer for
  truthful screenshots; see below).
- **Docs** — `docs/docs-notify-card.md` (sibling of `docs/summary-notify-card.md`);
  `metadata.octo.variant/source` section added to the summary doc.

## Testing

```
go vet ./...                                                   # clean
go test ./pkg/cardtmpl/... ./modules/notify/... ./pkg/cardmsg/... ./internal/carddispatch/... .   # ok
go run ./tools/lint-card-dispatch ./modules ./internal         # guard green
make i18n-extract-check && make i18n-lint                      # green
golangci-lint run ./...                                        # 0 issues
# E2E against real MySQL + Redis + WuKongIM:
go test -tags pilote2e ./pilote2e -run 'TestDocsNotify|TestSummaryAndDocsNotify_NoRegression'   # pass
```

Real-renderer screenshots (octo-web's actual `theme/*.css` +
`InteractiveCard/index.css` + ported `octoHostConfig`, `marked` for markdown)
were produced from `scripts/cardpreview` output for `summary.completed`,
`summary.failed`, and `docs.shared` and verified against the design reference.

- [x] Unit/integration tests added/updated
- [x] Manually verified (unit + pilote2e E2E + real-renderer screenshot)

## COMPREHENSION

   It adds a second trusted card *producer* (`docs-notify`) to the
   `internal/carddispatch` boundary and a `NotifyReq.DocsCard` ingress on the
   existing `POST /v1/internal/notify`. A structured `DocsCard` request is
   authored into an `octo/v1` card by `cardtmpl` and dispatched through the
   producer-bound `Sender`, which runs the full fail-closed pipeline (live bot
   identity → live Space/member ACL → `Validate` → authoritative `plain`/
   `space_id` → final size recheck → one transport call). The new
   `metadata.octo` object is server-authored only — `pkg/cardmsg.Validate` does
   not walk metadata sub-fields, so no user input reaches it there, and the
   `variant`/`source` values are picked by the server from `Kind`/producer, never
   from the caller.

   The only live caller of `/v1/internal/notify` today is smart-summary, which
   sends `Card` (or text) and is untouched — the 3-way mutex keeps its existing
   400s intact (a present-but-empty `{}` payload still counts as intent-to-send,
   so it cannot silently combine with a card). `docs-notify` is inert until
   octo-docs-backend opts in with `DocsCard` **and** `OCTO_CARD_MESSAGE_ENABLED`
   is on. A non-https `External.WebLoginURL` fails the card build fail-closed and
   degrades to a text DM rather than sending a dead link. Rollback: remove the
   one `docs-notify` spec or flip the global env.

   `pkg/cardtmpl` docs snapshot + `cardmsg.Validate` + deep-link shape +
   unsafe-input rejection tests; `modules/notify` docs validation matrix,
   kind→variant mapping, actor fallback, degrade-to-text, batch preflight, and
   3-way-mutex integration tests; `main_carddispatch_test.go` pins both producers
   to the reviewed shape and rejects unknown IDs; and
   `pilote2e/docs_card_wukongim_test.go` drives a real
   `POST /v1/internal/notify` → real WuKongIM and reads back the persisted
   type-17 card, asserting the `/d/{doc_id}?sp=` deep-link and
   `metadata.octo.variant="docs.shared"`. `TestSummaryAndDocsNotify_NoRegression`
   runs both producers back-to-back in one server lifecycle.

## Cross-repo dependency (must sync)

End-to-end needs octo-docs-backend to add an outbound notify client (symmetric
to `octo-smart-summary/internal/notify/`) that POSTs the `DocsCard` body per
`docs-notify-contract.md`. Unlike summary, the docs deep-link route (`/d/:docId`)
is **already live** in octo-web, so the "查看详情" button works on day one.

## Checklist

- [x] Updated documentation

---
_Generated by [Claude Code](https://claude.ai/code/session_01Rni147Au4tj4jj7AJPFCCK)_

### DIFF
--- a/.octospec/tasks/card-message-internal-dispatch/docs-notify-contract.md
+++ b/.octospec/tasks/card-message-internal-dispatch/docs-notify-contract.md
@@ -1,0 +1,157 @@
+# docs-notify 卡片通知 — 跨仓 ingress 契约
+
+> 仓内契约，供 octo-server `docs-notify` 生产者与 `octo-docs-backend` 对齐,PR
+> 打开后整理成英文 issue/comment 同步给 `Mininglamp-OSS/octo-docs-backend`。
+> 关联:brief.md(本目录)、summary-notify-contract.md(姊妹契约)、
+> handoff.md、`docs/docs-notify-card.md`。
+> 状态:契约草案,由本 PR 落地 `pkg/cardtmpl` / `modules/notify` 侧。
+> 最后更新 2026-07-14。
+
+## 背景一句话
+
+`docs-notify` 是 `internal/carddispatch` 的**第二个**生产者（首个见
+`summary-notify-contract.md`）,把 docs-backend 想推给用户的**自动化系统通知**
+（分享、评论、访问申请）升级为 `octo/v1` 展示卡片,复用 `notification` User Bot
+身份 → 用户会话列表**不新增**系统 Bot 会话。docs-backend 不构造 type-17 map、
+不发文本兜底、不做卡片模板 —— 只发**结构化字段** `DocsCardFields`,一切文案 /
+布局 / deep-link / i18n 由 octo-server owning。
+
+**用户主动分享** 不在本契约内 —— 那是独立的用户认证 + 服务端 mint 通路
+（见 `.octospec/tasks/user-resource-share-card/brief.md`）,发送方为分享用户本人。
+
+## 契约要点（锁定项）
+
+1. **同端点、同 token、同粒度**:仍是 `POST /v1/internal/notify`,头
+   `X-Internal-Token`（复用 `NOTIFY_INTERNAL_TOKEN`;fail-closed）,**一请求
+   一收件人**（`targets` 单元素）。`space_id` / `service` / `actor_uid` 语义
+   与 summary 完全相同,octo-server 已有的 dedup / actor 排除 / memberCache
+   / ensureNotifyBotReady / carddispatch 派发全部复用。
+2. **三选一互斥**:`Payload` / `Card`（summary） / `DocsCard`（docs）**只能
+   出现一个**。多于一个 → 400 `err.server.notify.card_invalid`;全都缺失
+   → 400 legacy「payload不能为空」。字段级检测:
+   - `req.Payload != nil` 计入(包括空 `{}` —— 呼应「显式意图」)。
+   - `req.Card != nil` 计入。
+   - `req.DocsCard != nil` 计入。
+3. **禁止客户端手搓 type-17 map**:Decision 14 仍生效 —— `payload` 若被
+   `cardmsg.IsCardPayload` 判为卡片(`type=17`)一律 400
+   `err.server.notify.card_not_allowed`,无论走 `Card` 还是 `DocsCard`。
+4. **响应契约不变**:仍返回 `NotifyResp{delivered:[], filtered:{uid:reason}}`。
+   reason 词表与 summary 一致(`not_space_member` / `target_denied`
+   / `dispatch_failed` / `busy` / `send_failed`) —— docs-backend 可**照抄**
+   smart-summary 的 dedup / retry / sweep 状态机。
+5. **模板/文案/链接归属 octo-server**:docs-backend **只发原始字段**;卡片布局、
+   按钮文案(查看详情)、FactSet 标签(操作人 / 时间)、attribution
+   (「Alice 分享了文档」)、`/d/{doc_id}?sp={space_id}` deep-link、
+   `metadata.octo.variant` / `metadata.octo.source` 全部由 octo-server
+   `pkg/cardtmpl` + `modules/notify.buildDocsCard` + `i18n.OutboundLanguage`
+   生成。docs-backend 不再拼「XX 分享了 <link>」这类降级文本(旧字段作废;
+   octo-server 侧的 `buildDocsFallbackText` 负责纯文本降级)。
+6. **批量端点 `/notify/batch` 拒绝 `DocsCard`**:与 `Card` 同,卡片必须走单条端点
+   `/notify`;batch 一条含 `DocsCard` 即 400,不做任何投递。
+
+## 请求体（卡片模式）
+
+```jsonc
+POST /v1/internal/notify
+X-Internal-Token: <NOTIFY_INTERNAL_TOKEN>
+{
+  "space_id":  "spc_xxx",              // 现状:必填,server 用 memberCache 校验收件人
+  "service":   "docs-service",         // 现状不变
+  "targets":   ["uid_recipient"],      // 现状:单收件人
+  "actor_uid": "",                     // 现状:通知场景一般留空
+  "docs_card": {                       // 新增:非空即走 docs-notify 卡片 producer
+    "doc_id":     "d_20260713_abcd",   // ★ 见下「标识」
+    "kind":       "shared",            // "shared" | "commented" | "access_requested"
+    "title":      "产品设计方案",         // 原始标题(server 负责转义/截断)
+    "actor_name": "Alice",             // 预格式化的操作人显示名;空则用「有人」/「Someone」兜底
+    "excerpt":    "Q3 上线计划已确认",    // 可选预览/评论/申请说明;≤ 300 runes 截断
+    "updated_at": "2026-07-13 15:04"   // 已格式化的时间字符串;空则省略「时间」行
+  }
+}
+```
+
+`payload` / `card` / `docs_card` 三者同时缺失 → 400 legacy;多于一个 → 400
+`err.server.notify.card_invalid`。
+
+## 三个约定（对齐 summary-notify）
+
+- **标识 `doc_id`(不是自增 `id`)**:deep-link `/d/{doc_id}?sp={space_id}` 用
+  docs-backend 侧不可枚举的文档标识 —— 与 octo-web `/d/:docId` 独立路由
+  (已在线;冷加载 + 登录跳转 + XIN-398 多会话 sid 恢复)对齐。**这是 docs-notify
+  与 summary-notify 的关键差异:docs 侧「查看详情」按钮开箱可用**,
+  summary 侧仍在等 `/s/:taskId` 上线。
+- **时间字段传「已格式化字符串」**:`updated_at` 由 docs-backend 用其自身时区
+  格式化后传字符串,octo-server 原样填进 FactSet 值。理由:octo-server 无 docs
+  业务时区配置,避免时区漂移。**标签(「时间」/「At」)仍由 octo-server 按收件
+  人语言本地化。**
+- **`actor_name` 传预格式化的显示名**:显示名由 docs-backend 侧解析(可含姓+
+  称谓 / 组织信息 / 匿名替换等业务规则);octo-server 侧 `escapeMarkdown` 后
+  嵌入 attribution。理由:避免 card 路径依赖额外 DB 查询,一致性与 summary
+  的 `Title` 同一纪律。
+
+## 字段来源（docs-backend 侧现成 / 需要新增）
+
+| card 字段 | docs-backend 来源(建议) |
+| --- | --- |
+| `doc_id` | 文档实体的 `id` 字段(与 `/d/:docId` 路由 taskId 同义) |
+| `kind` | 触发场景:分享 = `shared`;评论 = `commented`;访问申请 = `access_requested` |
+| `title` | 文档 `title` 字段(空则用「无标题」兜底 —— docs-backend 侧决策) |
+| `actor_name` | 触发者的显示名(docs-backend 已 resolve;匿名场景可留空) |
+| `excerpt` | 分享:留空 / 简短介绍;评论:评论内容摘要;访问申请:申请理由 |
+| `updated_at` | 触发事件的时间戳格式化字符串(docs-backend 时区) |
+
+## octo-server 侧（本 PR 落地）
+
+- `docs-notify` producer 绑定已有 `notification` Bot(`user`+`app`+
+  `robot.status=1`),卡片和纯文本降级复用其 provisioning/readiness;
+  不新增专用 `docs` Bot。
+- 注册 `docs-notify` producer(DM/`octo/v1`/system-notification/MaxInFlight
+  20),经 `SenderFromContext` 注入 `modules/notify`。与 `summary-notify` 同
+  `ProducerSpec` shape,仅 ID 不同(见 `main.cardDispatchProducerSpecs`)。
+- `docs_card` 分支:`memberCache.verify` → `cardtmpl.BuildDocsResourceCard` →
+  每个成员 `sender.Send` → 汇总 `delivered/filtered`。**建卡/配置(如
+  deep-link 非 https)失败降级为原纯文本 DM**,保证通知必达(此为 octo-server
+  侧策略,brief 未强制,PR 说明)。
+- **metadata 预留**:每张卡在 AdaptiveCard 根塞 `metadata.webUrl` +
+  `metadata.octo = {variant, source}`;`variant` 保留词表见
+  `docs/summary-notify-card.md` §2.3;`source.label` 供渲染端做「来自文档」
+  角标 —— docs-backend 无需管这些字段,server 侧根据 `kind` 决定 variant,
+  根据 producer 决定 source.label。
+
+## enablement 门（端到端上线前必须齐）
+
+1. **octo-web** 上线 `/d/:docId` 路由 —— **已在线**(既有 standalone
+   doc 通路);相较 summary,docs 端「查看详情」链接开箱可用;
+2. octo-server 落 `cardtmpl` snapshot/Validate 测试 + `docsDeepLink` 形状测试
+   + pilote2e E2E(**本 PR 落地**);
+3. **octo-docs-backend** 侧新增 outbound IM notify 客户端(仿造
+   `octo-smart-summary/internal/notify/`),按上述 body 结构 POST。
+   建议实现要点:
+   - 每场景一个函数(`NotifyDocShared` / `NotifyDocCommented` /
+     `NotifyDocAccessRequested`),内部塞 `Kind` 常量;
+   - 复用 `X-Internal-Token`(`NOTIFY_INTERNAL_TOKEN` = 同一 token,
+     docs-backend 与 smart-summary 共用);
+   - per-recipient dedup(避免同一评论触发多次通知);
+   - 消费 `NotifyResp{delivered,filtered}` —— 只认 `delivered[]` 判真送达,
+     `filtered` 内标记为 `not_space_member` / `busy` / `dispatch_failed` 的
+     可延时重试;
+   - **不实现自身文本降级路径** —— server 侧已负责;发失败即上报错误让
+     调用方按业务重试(与 smart-summary 侧的 `Sweep` 一致)。
+
+在 (1)(3) 到位前,octo-server 侧行为惰性:无 `docs_card` 请求即无 docs 卡片
+流量;全局 `OCTO_CARD_MESSAGE_ENABLED` 为总开关,回滚可移除 producer spec
+或关该 env。
+
+## 明确不在本契约内
+
+- 用户主动分享文档到 DM/群/子区 —— 见 `../user-resource-share-card/brief.md`,
+  发送方为分享用户本人,非 Bot proxy。
+- 访问申请的交互(「同意 / 拒绝」按钮) —— 需要 producer 升级到 `octo/v2` +
+  绑定一个跑 `/v1/bot/events` 的 action-owner;docs-backend 目前是
+  TS/Hocuspocus,不跑 Go bot 事件轮询,升级为独立议题。
+- 群/子区自动通知(如「这份文档被评论了」推到关联群) —— 需要 producer
+  widening 到 group/thread channel type,同步需要 per-channel rate rule +
+  cluster-wide cap 决策(brief › Industry practice alignment)。
+- 文档缩略图 / 富预览(Image / RichTextBlock) —— 目前只用 TextBlock + FactSet
+  +  Optional Excerpt。未来若加缩略图,需要 docs-backend 自建 SSRF-safe
+  fetcher 并提供绝对 https URL。
--- a/.octospec/tasks/card-message-internal-dispatch/summary-notify-contract.md
+++ b/.octospec/tasks/card-message-internal-dispatch/summary-notify-contract.md
@@ -2,8 +2,11 @@
 
 > 仓内草稿,供 octo-server P2(`summary-notify` pilot）实现时对齐,PR 打开后整理成英文
 > issue/comment 同步给 `Mininglamp-OSS/octo-smart-summary`。
-> 关联:brief.md(本目录)、handoff.md、octo-server PR #577(P1 地基)。
-> 状态:待 smart-summary owner 确认字段清单。最后更新 2026-07-14。
+> 关联:brief.md(本目录)、handoff.md、docs-notify-contract.md(姊妹契约,
+> 同一 ingress 端点、同 token、同响应 shape)、octo-server PR #577(P1 地基)、
+> #579(pilot)、#580(共用 notification Bot)。
+> 状态:已实现;`docs-notify` 姊妹 producer 于同 PR 落地(见 handoff)。
+> 最后更新 2026-07-14。
 
 ## 背景一句话
 
--- a/configs/tsdd.yaml
+++ b/configs/tsdd.yaml
@@ -35,9 +35,13 @@ mode: "debug" #   运行模式 debug or release
 
 ##################### 外网配置 ####################
 #external:
-#  ip: "" # 外网ip
-#  baseURL: "" # 外网访问地址 例如 http://192.0.2.1:8090
-#  webLoginURL: "" # web im的登录地址 例如 https://web.tangsengdaodao.com
+#  ip: "" # 外网ip；仅当 baseURL 空时作为兜底 host（见 pkg/botutil/connect.go）
+#  baseURL: "" # 外网 API 访问地址 例如 http://192.0.2.1:8090（baseURL 已给出时 ip 冗余，两者保留是历史兜底）
+#  webLoginURL: "" # web im 的登录/门户地址 例如 https://web.tangsengdaodao.com
+#  # webLoginURL 是**卡片通知 deep-link** 的 origin 来源（/s/:taskId、/d/:docId）；
+#  # 未配置或非 https 时，summary-notify / docs-notify 建卡失败，会静默降级为纯文本 DM
+#  # （见 pkg/cardtmpl/resource.go summaryDeepLink / docsDeepLink 与
+#  # modules/notify/card.go 的 build-then-degrade 分支）。
 
 ##################### 日志配置 ####################
 #logger:
--- a/docs/docs-notify-card.md
+++ b/docs/docs-notify-card.md
@@ -1,0 +1,253 @@
+# Docs-Notify 通知卡片（`internal/carddispatch` 第二个生产者）
+
+> **权威声明**：本文档描述 `docs-notify` 生产者当前的卡片形态与组装规则。
+> 权威源码：`pkg/cardtmpl/resource.go`（`BuildDocsResourceCard` + `docsDeepLink`）、
+> `modules/notify/card.go`（`deliverDocsCardNotification` / `buildDocsCard` /
+> `docsLabelsFor` / `docsAttributionAndVariant`）、`modules/notify/model.go`
+> （`DocsCardFields`）、`internal/carddispatch/`（派发流水线）、
+> `pkg/cardmsg`（wire 协议与校验）。规范源头：
+> `.octospec/tasks/card-message-internal-dispatch/brief.md`（Decisions 1/3/4/11/14）与
+> 同目录 `docs-notify-contract.md`（跨仓 ingress 契约）。两者如有出入，以代码为准。
+>
+> 兄弟文档：`docs/summary-notify-card.md`（`summary-notify` 生产者）— 组装管线、
+> 降级链、Deep-link 前置、`metadata.octo.variant` 保留词表、错误分类**完全一致**，
+> 本文档只呈现 docs 特有的字段/文案/deep-link 差异。
+
+## 1. 概览
+
+`docs-notify` 是 `internal/carddispatch` 落地的第二个内部卡片生产者，服务于
+**docs-backend 的自动化通知**（分享、评论、访问申请）。业务侧（`octo-docs-backend`）
+经既有 `POST /v1/internal/notify` 内网接口投递**结构化字段**
+（`NotifyReq.DocsCard` = `DocsCardFields`），octo-server 侧完成
+**卡片组装 → 派发 → 降级**闭环。
+
+- **发送方身份**：与 `summary-notify` 共用同一 `notification` User Bot
+  （`ensureNotifyBot` provision）；用户会话列表里不产生第二个系统 Bot 会话。**能力
+  隔离在 producer 粒度**：不同 producer 各自的 `MaxInFlight` / 允许 profile / 允许
+  channel type 独立配置，`docs-backend` 不因共用 Bot 身份而获得跨 producer 越权。
+- **接口层**：仍是 `POST /v1/internal/notify` + `X-Internal-Token`。body 里
+  `Payload` / `Card`（summary） / `DocsCard`（docs） **三选一**，见 §3。
+- **生产者信息**：`ProducerID="docs-notify"`，`SenderUID=NotifyBotUIDValue`，
+  `AllowedChannelTypes=[Person]`，`AllowedProfiles=[octo/v1]`，
+  `SpacePolicy=SystemNotification`，`MaxInFlight=20/process`。
+
+**用户主动分享** 不是 docs-notify 的范围（brief Out-of-scope）：那是一条独立的
+「用户认证 + 服务端 mint」资源分享通路（`.octospec/tasks/user-resource-share-card/`
+brief），发送方为分享用户本人，与本 producer 无关。docs-notify 覆盖的是
+**自动化系统通知**：文档被分享给你、评论了你、有人请求访问你的文档等。
+
+## 2. 卡片形态（Adaptive Card 1.5）
+
+`DocsCardFields.Kind` 决定 attribution / variant；`ActorName` 决定 attribution
+的主语（缺省用「有人」/「Someone」兜底）。
+
+| `Kind` | 场景 | attribution（含 `ActorName`） | attribution（无 `ActorName`） | `metadata.octo.variant` |
+|---|---|---|---|---|
+| `"shared"` | 文档被分享 | `{Actor} 分享了文档` | `有人分享了文档` | `docs.shared` |
+| `"commented"` | 有新评论 | `{Actor} 评论了文档` | `有新评论` | `docs.commented` |
+| `"access_requested"` | 访问申请 | `{Actor} 请求访问文档` | `有人请求访问文档` | `docs.access_requested` |
+
+**公共结构**（与 summary 卡完全对齐）：
+- Header：无 `IconURL` 时是 `Container` + 加粗 `TextBlock(Title)` + 可选
+  `isSubtle` attribution TextBlock。
+- Excerpt：可选 `TextBlock(wrap=true)`，超过 `MaxExcerptRunes=300` 截断。
+- FactSet：`ActorName`（若非空）→ 「操作人 / By」；`UpdatedAt`（若非空）→ 「时间 /
+  At」。两者都空则整块省略。
+- ActionSet：`Action.OpenUrl("查看详情", deepLink)`。
+- **入卡文本一律经 `escapeMarkdown` 转义**（同 summary，`* _ [ ] ( ) < > \` \# ~ |`
+  逐字符加反斜杠），阻断外部字段被 CommonMark 解析成活链接／图片。
+
+### 2.1 shared 完整 AC JSON
+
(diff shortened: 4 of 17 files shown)


Jerry-Xin (dismissed):
The PR is in scope for octo-server and the implementation is ready to merge.

**💬 Non-blocking**
- 🟡 Warning: The docs examples drift from the implemented metadata contract. `pkg/cardtmpl/resource.go:260` emits `metadata.octo.source` when `Source` is set, and both producers set it, but the JSON examples in `docs/docs-notify-card.md:87` and `docs/summary-notify-card.md:96` only show `variant`. The shared metadata section in `docs/summary-notify-card.md:208` also documents only `variant`. This will confuse renderer/backend integrators even though runtime behavior is correct.

**✅ Highlights**
- The new `docs-notify` path mirrors the existing summary producer pipeline: dedup, actor exclusion, member verification, notification bot readiness, bounded fan-out, and carddispatch authorization.
- The three-way `Payload` / `Card` / `DocsCard` mutex is correctly enforced in single notify and batch preflight paths.
- Card template output remains validated through `cardmsg.Validate`, with HTTPS deep-link enforcement and text fallback on build/sender/feature-disable cases.
- Local verification passed: `go test ./pkg/cardtmpl ./modules/notify ./internal/carddispatch .`.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #735 (octo-web)

## Summary

This PR introduces line-spacing support to the docs editor (SCHEMA_VERSION 17): a self-built `LineHeight` Tiptap extension adds `lineHeight`, `spaceBefore`, and `spaceAfter` global attributes to `heading` and `paragraph` nodes, toolbar controls (line-height dropdown + custom input, paragraph-spacing dropdowns) to set them, and docx export mapping to carry them through to Word output. A `VITE_DOCS_LINE_SPACING` feature flag (default ON) gates only the toolbar UI — the extension is always registered so documents round-trip the attrs losslessly regardless of the flag. The PR also fixes the focus-steal bug in the custom line-height input by switching it from controlled-per-keystroke to commit-on-blur/Enter.

16 files changed (944 additions, 13 deletions): extension definition, toolbar UI, schema bump, config flag, docx export mapping, i18n keys (both locales), Dockerfile build-arg, and comprehensive tests covering sanitisation, round-trip, toolbar interaction, flag gating, and docx export.

## Verification

- ✅ **Sanitisation at both boundaries** — `sanitizeLineHeight` enforces unitless positive number ≤10 via `LINE_HEIGHT_RE`; `sanitizeSpacing` enforces `<number>px|em` with magnitude ≤1000 via `SPACING_RE`. Both applied at parseHTML, renderHTML, command dispatch, and docx export — four-sided defence.
- ✅ **Feature flag default-ON** — `LINE_SPACING_ENABLED = envOr(import.meta.env?.VITE_DOCS_LINE_SPACING, 'true') === 'true'`; the `envOr` helper correctly normalises Vite's empty-string-for-unset to the fallback. Dockerfile ARG/ENV pair follows the existing `VITE_DOCS_*` pattern.
- ✅ **Schema round-trip preserved** — `LineHeight` registered in both `buildExtensions` and `buildPreviewExtensions`, AFTER `TextAlign` (canonical style property order). All three attrs default to `null` so old documents are unaffected.
- ✅ **Command set/unset symmetry (C1)** — all six commands (`set`/`unset` × 3 attrs) implemented with `.some(Boolean)` return, sanitise-on-set, same `LINE_HEIGHT_TYPES` scope.
- ✅ **Extension registration order** — `LineHeight` placed after `TextAlign` in both `buildExtensions` (`:197-202`) and `buildPreviewExtensions` (`:319`), ensuring canonical `style` serialisation order matches backend toDOM contract.
- ✅ **i18n parity** — all 6 new keys present in both `en-US.json` and `zh-CN.json`.
- ✅ **docx export** — `mapSpacing` reuses `sanitizeLineHeight` and `sanitizeSpacing` for cap enforcement; `line` mapped to 240ths-of-a-line with `LineRuleType.AUTO`; `before`/`after` converted to twips (15 twips/px, 16px/em). Wired into both `convertParagraph` and `convertHeading` in `nodes.ts`.
- ✅ **Test coverage** — `LineHeight.test.ts` (sanitise + DOM round-trip), `Toolbar.test.tsx` (custom input focus-steal fix, paragraph spacing controls), `lineSpacingFlag.test.tsx` (flag-OFF gating + schema persistence), `spacing.test.ts` (docx mapping + cap parity), `schema/index.test.ts` (version bump), `extensions.test.ts` (attr registration).

Static analysis only at head `849650a1`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One nit below.

### Nit — Redundant regex re-match in `spacingToTwips` (`packages/docs/src/export/docx/styles.ts:233`)

`sanitizeSpacing(raw)` returns a string matching `^(\d+(?:\.\d+)?)(px|em)$` or `null`. The subsequent `const m = /^(\d+(?:\.\d+)?)(px|em)$/.exec(v)` at `:233` re-validates the same pattern on already-sanitised output — it can never fail when `v !== null`. Consider extracting `SPACING_RE` from `LineHeight.ts` as a shared constant, or parsing the validated string directly (`parseFloat(v)` + `v.endsWith('em')`). Not a bug, just unnecessary work.

- **Data flow tracing**: dropdown `onChange` → `editor.chain().focus().setLineHeight(v).run()` → sanitise → `updateAttributes` → re-render → `renderHTML` → sanitise → `style` string. All four boundaries sanitise.
- **Custom input commit-on-blur/Enter**: draft stays local while typing (no snap-back for partial values like "1."); on commit, invalid drafts revert to the last committed value rather than writing garbage. The `seenCommitted` ref correctly distinguishes self-commits from external changes.
- **Flag-OFF path**: `LINE_SPACING_ENABLED=false` hides all three toolbar controls; the `LineHeight` extension stays registered so attrs still round-trip (collab-lossless). Tests pin both assertions.
- **SCHEMA_VERSION bump**: v16 → v17 with proper changelog comment; `SCHEMA_NODES`/`SCHEMA_MARKS` correctly left unchanged (attrs, not nodes/marks).
- **Dockerfile**: ARG/ENV pair for `VITE_DOCS_LINE_SPACING` follows the established pattern of other `VITE_DOCS_*` flags in the builder stage.
- **`envOr` semantics**: correctly handles Vite's empty-string-for-unset-build-arg by treating whitespace-only values as unset and falling back to the default.

## Verdict: APPROVED

No correctness, security, build, or functional concerns. The line-spacing feature is well-structured: defence-in-depth sanitisation at all four boundaries, clean feature-flag gating that preserves schema round-trip, comprehensive test coverage, and proper docx export wiring. The single nit (redundant regex in `spacingToTwips`) is cosmetic.





[Octo-Q] verdict: APPROVE — no P0/P1 blockers; clean symmetric docs-notify implementation with comprehensive test coverage, proper input validation, and defensive text-fallback degradation. P2 naming concern and redundant-trim nit are non-blocking.

OctoBoooot (dismissed):
## Review: feat(notify): docs-notify card producer + reserved metadata namespace (#584)

**Verdict: Approve** — a faithful, correctly-isolated third producer in the card-dispatch arc (docs-notify, after summary-notify #579/#580), plus a server-authored `metadata.octo.{variant,source}` namespace. Zero blockers, zero majors across template/metadata safety, producer isolation, trust boundary, deep-link/SSRF, and delivery parity. Approval is conditional on **Test** finishing green (all other code CI — Build/Vet/Lint/Card-Dispatch-Lint/osv/i18n — is green; check-sprint governance is green too); if Test fails this reverts to Request changes.

Risk: trust-boundary code (new producer + a new metadata surface), so I fanned out independent dimension agents and adversarially verified. `size/XL`.

### What I byte-verified
- **Three-way mutual exclusion is airtight.** `api.go` counts `Payload != nil` / `Card != nil` / `DocsCard != nil`; `present > 1` → 400 `card_invalid`. This generalizes the #579 empty-`{}` fix correctly (an explicit `{}` payload counts as present, so `{}`+Card/DocsCard is rejected). Batch endpoint rejects any `Card`/`DocsCard` entry.
- **The `metadata.octo` namespace is genuinely server-authored.** `DocsCardFields` exposes only `doc_id/kind/title/actor_name/excerpt/updated_at` — no `variant`/`source`/`metadata` field. `variant` comes from a server-side `Kind` switch (constant literals, default → `docs.shared`), `source` is a server i18n constant. Confirmed `cardmsg.Validate` does not walk metadata sub-fields — which is safe precisely because nothing caller-controlled reaches metadata. The one metadata field that could carry a URL (`source.iconUrl`) is never set by callers and is `requireHTTPS`-gated if ever set.
- **Deep-link is breakout-safe.** `/d/{doc_id}?sp={space_id}` — `doc_id` is `PathEscape`'d, `space_id` is `QueryEscape`'d, appended after a scheme+host-only https origin. Verified hostile inputs don't inject a query param, fragment, path traversal, or second scheme (`javascript:` stays an escaped path segment). `webOrigin` rejects non-https/empty/`//evil`/userinfo tricks and degrades to a text DM.
- **All caller fields are escaped for every render context.** Title/actor/excerpt/facts pass through `escapeMarkdown` covering link-text `[ ]`, URL `( )`, autolink `< >`, bold/emphasis, inline-code, plus JSON via `json.Marshal` — the full context set, not just one.
- **Producer isolation + delivery parity hold.** `docs-notify` is a capability clone of `summary-notify` (DM-only / octo/v1 / SystemNotification / MaxInFlight=20), bound to its own producer ID, both on the shared `notification` bot. `deliverDocsCardNotification` replicates the reviewed pipeline (dedup → actor-exclude → live member verify → bounded fan-out → producer-bound `Sender.Send`, never a raw type-17). Reuses the #579/#580 atomic `botOK` readiness — no new unsynchronized flag, no double-send, empty-delivery provisions no bot.
- **Fallback never drops a notification.** `buildDocsFallbackText` degrades to a plain DM when card build fails (feature off / sender missing / non-https origin); failed sends go to `Filtered` (surfaced for retry), never silently dropped.

### Minor
- `modules/notify/card_test.go` — the card/docs-card delivery tests prime only the delivered member and assert `Filtered` empty; there is **no negative test** where a non-member target is filtered out. A mutation dropping the member-`verify()` pre-filter would pass every card-path unit test (and the pilote2e E2E only seeds the recipient). Kept **minor** because the carddispatch dispatcher independently re-verifies membership (Decision-3 backstop) and the summary sibling has the same shape — but a "non-member excluded → in `Filtered`, not `Delivered`" case would lock the pre-filter. Same gap applies to the summary path.
- `buildDocsFallbackText` interpolates caller `Title`/`Excerpt`/`ActorName` into the plain-text DM without newline-stripping — a `\n` in the title can inject a spoofed attribution line. Plain-text DM only (no link/script), trusted internal caller, so low impact; strip control chars/newlines for the fallback to be tidy.

### Questions
- Excerpt length asymmetry: the card truncates `Excerpt` to 300 runes (`truncateRunes`), but the text fallback writes it untruncated. Same source data (no leak), but the fallback isn't a strict subset of the card as the comment implies — intended?
- `escapeMarkdown` omits `!` and leading list markers (`- `, `1. `) / newlines — non-injecting (links can't form since `[ ]( )` are escaped), but a title like `"- a\n- b"` could render as a bullet list in a CommonMark client. Cosmetic; confirm the renderer treats TextBlock as inline-only.

### Praise
- The producer stays a strict capability clone bound to its own ID rather than widening the shared bot's surface — capability isolation lives on the producer, exactly as the #571 arc intended, so a docs card can't be dispatched through the summary producer or vice versa. And the `metadata.octo` namespace is designed safe-by-construction: server-authored values only, kept out of the caller-reachable `DocsCard` fields, so the fact that `Validate` doesn't walk metadata is a non-issue rather than a hole. The deep-link escaping (PathEscape + QueryEscape + https-origin gate) is the right depth for a URL built from caller-adjacent input.

### Note on merge gate
`mergeStateStatus` stays governed by required maintainer review + Test completion — approved ≠ mergeable. Given this is live trust-boundary code, a human maintainer should make the final merge call.

yujiawei (dismissed):
# Code Review — PR #584 (octo-server)

Reviewed at head SHA `c6319ac2d02f54d80708bca4a48dc2093fd12a80` against `main` (merge-base `5e336a0c`). Local `go build ./...`, `go vet`, and `go test ./pkg/cardtmpl/... ./modules/notify/... ./internal/carddispatch/...` all pass; `pilote2e` vet compiles.

**Spec: ✅**

The diff adds exactly what the linked contract describes and nothing beyond it:

- **Second producer, not a new module** — `docs-notify` is registered by promoting `summaryNotifyProducerSpecs()` to a shared `cardDispatchProducerSpecs()` that emits both specs from one `base` struct (`main.go`). No `register.AddModule` change, no package-global registration; capability isolation stays on the producer ID. `TestNotificationCardProducerRegistrations` pins both to the reviewed shape (DM-only, `octo/v1`, `system_notification`, `MaxInFlight=20`) and fails on any unknown ID — the sender allowlist is not silently extensible.
- **Ingress mutual exclusion** — `Payload` / `Card` / `DocsCard` are counted with `present` and rejected when `> 1` (`api.go`). Presence uses `!= nil` for `Payload`, so an explicit `{}` still counts as intent-to-send and cannot silently combine with a card. Covered by `TestIntegration_NotifyMutexRejectsDocsCardCombinations`.
- **Fail-closed degradation** — `deliverDocsCardNotification` decides card-vs-text once up front; a build failure (feature off, sender missing, non-https deep-link origin) degrades the whole request to a plain DM via `buildDocsFallbackText` rather than dropping the notification. Verified by `TestDeliverDocsCardNotification_DegradesToTextWhenSenderUnavailable`.
- **Reserved metadata namespace** — every card now emits `metadata.webUrl` + `metadata.octo.{variant,source}`; `variant`/`source` are server-picked from `Kind`/producer, never caller-supplied. The values are omitted when empty to keep the wire minimal.

No missing scope, no over-build, no deviation. The `metadata.octo.source.iconUrl` slot is a reserved-but-inert field explicitly called out in the contract, not an unrequested feature.

## 2. Code quality

**Quality: Approved**

I found no correctness, security, concurrency, or data-loss defects introduced by this PR. Trust-boundary handling is sound: `doc_id`/`space_id` are `url.PathEscape`/`QueryEscape`'d and the origin is derived from a positive-allowlisted absolute-https `webOrigin`, so a hostile identifier cannot escape the deep-link origin; the emitted `metadata` object passes `cardmsg.Validate` (verified by the `docs_test.go` envelope assertion), and `Validate` does not walk `metadata` sub-fields, so the server-authored namespace introduces no new user-input surface.

### P2 (non-blocking)

- **`modules/notify/model.go:54-56` — gofmt drift.** The `DocsCardKind*` const block is not gofmt-aligned (`gofmt -d` reports a re-alignment). CI will **not** catch this — `.golangci.yml` enables only `govet` and intentionally disables gofmt — so it does not block merge, but the PR description's "golangci-lint 0 issues" is technically true only because gofmt is off. Worth a one-line `gofmt -w`.
- **`modules/notify/card.go` — `context.Background()` in the send fan-out.** Each per-recipient `docsSender.Send(...)` / `sendSummaryText(...)` uses `context.Background()`, so a hung transport holds its semaphore slot and the request goroutine blocks on `<-resultCh` indefinitely. This is a real resilience concern, but it is an **exact mirror of the already-merged summary path** (`deliverCardNotification`) and the legacy text path (`n.ctx.SendMessage` also takes no context) — this PR does not regress the behavior, it stays consistent with it. Recommend a follow-up that threads a bounded `context.WithTimeout` through both card paths together, rather than diverging one of them here.
- **`modules/notify/card.go` — `sendSummaryText` reused for the docs fallback.** Functionally correct (both producers share the one `notification` bot identity), but the `summary`-specific name is a maintenance hazard: a future edit that adds summary-specific behavior to it would silently alter docs fallback DMs. Consider renaming to `sendNotifyText` / `sendFallbackText`.
- **Observability** — a card build failure that degrades to text is only visible via a `Warn` log. A `notify_docs_card_degraded_to_text` counter (symmetric for summary) would make silent, persistent degradation detectable in monitoring.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. All findings are P2/advisory; none block merge.

## 4. Suggestions

- One-shot `gofmt -w modules/notify/model.go` to clear the const alignment.
- Track the shared `context.WithTimeout` for both card fan-out paths as a single follow-up so summary and docs stay symmetric.
- Rename `sendSummaryText` to a producer-neutral name now that two producers depend on it.

## 5. Additional notes (review coverage)

- **Reviewer legs actually run:** one independent long-context leg completed and one line-level adversarial leg was **absent** (its runtime failed at the gateway with a persistent configuration error — treated as absent, not as implicit approval). The absent leg's line-level coverage of the concurrency fan-out and the mutex switch was re-done manually for this review.
- The absent leg's coverage note assumed a separate `docs.go` file existed; it does not — the docs card logic lives in `pkg/cardtmpl/resource.go` (present in the diff), so its markdown-escaping, `MaxExcerptRunes` truncation, and `doc_id` URL-encoding were verifiable and are covered by `docs_test.go` (`TestBuildDocsResourceCardRejectsUnsafeInputs`, deep-link shape, snapshot).
- Cross-file consistency between `main.go` producer specs and `modules/notify` producer IDs is verified and pinned by `main_carddispatch_test.go`.
- Cross-repo note (out of scope for this repo): end-to-end docs delivery requires octo-docs-backend to add the outbound notify client per `docs-notify-contract.md`; the `/d/:docId` web route is already live, so the deep-link works on day one.

OctoBoooot (commented):
## Follow-up on #584 — yujiawei's gofmt nit verified (real, non-blocking); verdict unchanged (APPROVE)

Byte-verified yujiawei's two 🟡s. One is a real gofmt nit, the other I'd already noted. Neither changes the verdict — my APPROVE stands (still conditional on Test).

### 🟡 (verified) — gofmt alignment in the added const block, `modules/notify/model.go:60-63`
`gofmt -d` flags the new docs-kind block: the `=` signs aren't aligned to the longest identifier —
```
-	DocsCardKindShared           = "shared"
-	DocsCardKindCommented        = "commented"
-	DocsCardKindAccessRequested  = "access_requested"
+	DocsCardKindShared          = "shared"
+	DocsCardKindCommented       = "commented"
+	DocsCardKindAccessRequested = "access_requested"
```
Real, and it's in code this PR adds. **Not a blocker** and not gating merge: the CI `Lint` check is green at this head (the repo's `golangci-lint` config evidently doesn't enable the `gofmt`/`gofumpt` linter, so this whitespace nit passes CI), so it doesn't bar approval the way #582's actual Lint failure did. But the PR body's "gofmt -l clean" claim is now stale — run `gofmt -w ./modules/notify/model.go` to tidy it. Credit yujiawei.

### 🟡 (already on record) — `context.Background()` on the send path, no timeout
Matches the same class I noted on the fallback/delivery path: the card/text send uses a background context with no deadline, so a hung WuKongIM call isn't bounded here. Non-blocking (bounded by the transport client's own timeout, if any); a `context.WithTimeout` would harden it. Same follow-up as the sibling summary path.

### Verdict unchanged: Approve (conditional on Test)
Both are non-blocking polish. The gofmt nit doesn't redden the CI Lint gate (verified green), so it doesn't affect mergeability — merge remains gated only on Test finishing green + required review. Worth a quick `gofmt -w` pass while Test runs.

OctoBoooot (approved):
## Delta re-review: #584 @ 652a1ffe — APPROVE re-affirmed (all review feedback addressed, with tests)

My prior APPROVE (@ c6319ac) was auto-dismissed by the head advance. Byte-verified the delta (ahead 1, no rebase) — it's exactly the review-feedback fixes, nothing new or risky. Re-affirming **APPROVE**, conditional on Test green.

### The delta addresses every actionable item from my + yujiawei's review
- **New negative member-verify tests** — `TestDeliverCardNotification_NonMemberExcluded` and `TestDeliverDocsCardNotification_NonMemberExcluded` close the exact mutation-thinking gap I raised (a target that isn't a member must land in `Filtered`, not `Delivered`). Dropping the `verify()` pre-filter would now fail a test.
- **Newline-injection fixed + tested** — new `sanitizeLine` collapses all control chars (`unicode.IsControl`) and is applied to every caller field in both the summary and docs text-fallback paths; `TestBuildDocsFallbackText_StripsNewlineInjection` pins it (an embedded `\n` in title/actor can no longer inject a spoofed line). Resolves my minor and yujiawei's context note.
- **gofmt nit fixed** — `model.go` const block now aligned (yujiawei's 🟡 / my follow-up).
- **`resource.go`** — comment-only clarification of `Source.Label` localization ownership; no logic change.

### Core trust-boundary surface unchanged (carries my prior verdict forward)
Blob-verified `modules/notify/api.go` (three-way mutual exclusion), `main.go` (producer registration), and `main_carddispatch_test.go` are **byte-identical** to my APPROVE'd `c6319ac`. So producer isolation, the DocsCard mutual-exclusion, server-authored metadata namespace, deep-link escaping, and readiness reuse are all untouched — the delta is only the fallback-sanitize + tests + gofmt + a doc comment.

Strictly better than the version I approved — same sound design, now with the negative tests and newline-sanitize I asked for. CI re-running on the new commit (Card-Dispatch-Lint/i18n/check-sprint already green; Build/Vet/Lint/Test in progress); if Test reddens I reverse. `mergeStateStatus` stays gated on Test + required maintainer review — a human makes the final call on this trust-boundary code.

Nice turnaround: every review item resolved with a test that pins it, and the load-bearing paths left untouched.

Jerry-Xin (approved):
Re-review at head `652a1ffe` (prior APPROVE `4692135374` @ `c6319ac2` was auto-dismissed by the head advance). The delta is a clean single commit (`ahead:1, behind:0`, no rebase inflation) that resolves both of my prior non-blocking notes and adds a genuine text-path security hardening with tests. **APPROVE.**

## Delta (c6319ac2 → 652a1ffe)
6 files: docs ×2, `modules/notify/card.go`, `modules/notify/card_test.go`, `modules/notify/model.go`, `pkg/cardtmpl/resource.go`. No changes to the mutual-exclusion endpoint logic, producer registration, `botOK` reuse, or the `metadata.octo` server-authored namespace guarantee.

## Prior 🟡 items — resolved
- ✅ **gofmt alignment** — `modules/notify/model.go:60-63` const block (`DocsCardKindShared/Commented/AccessRequested`) is now gofmt-aligned (`=` signs realigned; `-3/+3`).
- ✅ **docs ↔ `metadata.octo.source` drift** — both `docs/docs-notify-card.md` and `docs/summary-notify-card.md` now document `metadata.octo.{variant,source}`, including the `source.label` reserved word table (zh-CN/en-US), the i18n derivation path (`OutboundLanguage → summaryLabelsFor/docsLabelsFor.sourceLabel`), and the `iconUrl → requireHTTPS` note. `pkg/cardtmpl/resource.go` `Source` struct doc comment reinforces "Values are server-picked (never user-supplied)".

## Substantive change — re-verified clean
- **New `sanitizeLine` (card.go)** collapses any `unicode.IsControl` rune (newline, CR, tab, …) to a space and `TrimSpace`s — a strict superset of the previous `strings.TrimSpace`. Applied uniformly to every caller-supplied field in the plain-text fallback DM builders (`buildSummaryFallbackText`: Title/Reason/TimeRange/GeneratedAt; `buildDocsFallbackText`: ActorName/Title/Excerpt/UpdatedAt). The `\n` separators are server-authored line structure, not caller-controlled, so this closes an attribution/label line-injection vector on the text path while preserving legitimate structure. This is the text-path equivalent of the existing card-path `escapeMarkdown` defence.
- **Tests (card_test.go, +59):** `TestBuildDocsFallbackText_StripsNewlineInjection` locks the injection defence; `TestDeliverCardNotification_NonMemberExcluded` and `TestDeliverDocsCardNotification_NonMemberExcluded` pin the member-verify pre-filter (stranger → `Filtered["not_space_member"]`, never `Delivered`, only the member gets a DM).

## Gate
- Non-draft. Required checks at head: Lint ✅, Build ✅, Vet ✅, plus i18n Lint / Card Dispatch Lint / secret-scan / osv-scan / dependency-review all ✅. Test is `in_progress` (pending, not a blocker). No red required check.

No blocking items. Good, defensive follow-up on the pilot.

yujiawei (approved):
# Code Review — PR #584 (octo-server)

Reviewed at head SHA `652a1ffe` against `main` (merge-base `5e336a0c`). Independent review plus a cross-model advisory pass. Build, `go vet`, and the `pkg/cardtmpl` + `modules/notify` + root test packages all pass locally.

## 1. Spec / scope verdict — ✅

The diff implements the stated scope and nothing beyond it.

- **Second `internal/carddispatch` producer (`docs-notify`)** added mirroring the `summary-notify` shape: shared `notification` User Bot, DM-only, `octo/v1`, system-notification Space policy, `MaxInFlight=20`. Capability isolation stays on the producer ID (`main.go` `cardDispatchProducerSpecs`, `modules/notify/api.go`). No `register.AddModule` change, no package-global registration.
- **`NotifyReq.DocsCard` ingress** with a structured `DocsCardFields` (not a type-17 map). `Payload` / `Card` / `DocsCard` are made mutually exclusive; presence uses `!= nil` for `Payload` so an explicit `{}` still counts as intent-to-send and cannot silently combine (`api.go` `sendNotify`).
- **`metadata.octo.{variant,source}` namespace** added on the AdaptiveCard root, plus a `webUrl` sub-field. All values are server-authored (deep-link, server-picked variant from `Kind`, server-i18n label); no caller field lands in `metadata`.
- **Reserved source badge slot**: `Source{Label,IconURL}` is emitted into metadata but the template deliberately does not auto-render it into the card body — matches the "reserve a slot for future styling" scope, not an over-build.
- **Deep-link** `/d/{doc_id}?sp={space_id}` mirrors `summaryDeepLink`; origin from the positive-allowlisted https `External.WebLoginURL`.
- **Text fallback** (`buildDocsFallbackText`) mirrors the summary fallback so a card-build failure degrades to a plain DM.
- No missing items, no un-requested features, no divergence from the sibling summary path.

## 2. Code quality verdict — Approved

No blocking (P0/P1) findings. The trust boundary is handled correctly:

- **Injection defense (confirmed).** Every caller-supplied string on the card path is `escapeMarkdown`'d in the card body; on the text-fallback path each is routed through the new `sanitizeLine`, which maps `unicode.IsControl` runes (`\n`,`\r`,`\t`) to spaces. This closes the CRLF/newline attribution-spoofing vector in the line-structured DM. Covered by `TestBuildDocsFallbackText_StripsNewlineInjection`.
- **Metadata is not attacker-reachable.** `cardmsg.Validate`'s walker does not descend into `metadata`, but nothing user-supplied is placed there — `variant`/`source`/`webUrl` are all server-derived. Verified by reading `pkg/cardmsg/validate.go`.
- **Bounds.** Title ≤ `maxTitleRunes`, Source.Label ≤ `maxTitleRunes` (build returns an error → degrade to text), Excerpt truncated to `MaxExcerptRunes`. The full card payload is size-rechecked (512KiB) at dispatch and re-`Validate`d — same discipline as the summary path.
- **Deep-link.** `url.PathEscape(docID)` + `url.QueryEscape(spaceID)` with an https-only allowlisted origin; a non-https origin fails the build closed (`TestBuildDocsCard_RejectsNonHTTPSDeepLinkOrigin`).
- **Fan-out.** `sem` cap 20, `resultCh` buffered to `len(members)`, deterministic drain — no goroutine leak. Per-target transport failure is recorded in `Filtered` (preserving the caller's retry state machine) rather than silently downgrading that one target to text; the card-vs-text decision is made once, up front, for the whole request. This is the correct granularity.
- **Refactor** of `BuildSummaryResourceCard` into the shared `buildResourceCard` + `webOrigin` helpers is behavior-preserving; the summary snapshot test was updated for the new `metadata` block and still passes.

### Non-blocking suggestions (P2 / nits)

- `deliverDocsCardNotification` sends its text fallback via `n.sendSummaryText(...)`. The method just delegates to the shared notification bot, so it is functionally correct, but the `Summary`-prefixed name now serves two producers. Consider renaming to a neutral `sendBotTextDM` / `sendTextFallback` to avoid confusion for the next producer.
- `DocsCardFields.ActorName` / `UpdatedAt` have no explicit per-field length cap at ingress. They are escaped and bounded transitively by the 512KiB dispatch recheck, so this is not a correctness issue, but a small explicit cap (as Excerpt has) would make the contract self-documenting.

## 3. Overall verdict — APPROVED

Spec ✅ and Quality Approved. Clean, well-tested, symmetric with the reviewed summary-notify pilot; the trust boundary and fail-closed degrade paths are correct.

### Review coverage note

This was a two-model pass. The cross-model leg independently confirmed the CRLF-injection defense, the 3-way mutex correctness, and the per-target failure isolation. A third adversarial leg was attempted but did not complete due to a tooling/gateway error on the reviewer side (not a code issue); it is recorded as absent, not as tacit approval. Items the cross-model leg could not reach — `cardtmpl` HTTPS/escape internals, deep-link escaping, and `memberCache.verify` — were verified directly in this review.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #584 (octo-server)

## Summary

This PR adds a `docs-notify` card producer to octo-server, mirroring the existing `summary-notify` path. It introduces `DocsCardFields` as a third mutually-exclusive payload type on `NotifyReq`, a new `BuildDocsResourceCard` template function, a `metadata.octo.{variant,source}` extension on the AdaptiveCard document, and a `sanitizeLine` helper that hardens the plain-text fallback path against newline injection. The producer is registered alongside `summary-notify` in `main.go`, sharing the same `notification` bot identity. The implementation is well-structured, with comprehensive unit tests, E2E tests, and documentation.

## Verification

Static analysis only at head `652a1ffed07a36aa9e9b41977e13d091c66a91bf`; build and tests not executed in this environment.

- ✅ **Mutual exclusivity** — `sendNotify` correctly counts non-nil fields among `Payload`, `Card`, and `DocsCard` and rejects when more than one is present. Batch endpoint rejects both `Card` and `DocsCard`. Tests cover all combinations.
- ✅ **Member-verify pre-filter** — `deliverDocsCardNotification` calls `memberCache.verify` before fan-out, matching the summary path. Non-members are filtered and reported in `Filtered`, never delivered.
- ✅ **Actor exclusion** — `req.ActorUID` is excluded from targets, identical to the summary path.
- ✅ **Deep-link safety** — `docsDeepLink` reuses `webOrigin` which enforces absolute https. `url.PathEscape`/`url.QueryEscape` are used for docID and spaceID. Non-https origin triggers build failure → graceful text degradation.
- ✅ **Input sanitization** — `sanitizeLine` is applied to all caller-supplied fields in the fallback text path (`buildDocsFallbackText`). Card path uses `escapeMarkdown` for all rendered text. Newline injection is tested.
- ✅ **Concurrency safety** — Goroutine fan-out uses semaphore (bounded chan, capacity 20) with `defer` release. `canCard`, `document`, and `fallbackText` are set before goroutines launch (read-only in closures). Results collected via buffered channel sized to `len(members)`.
- ✅ **Graceful degradation** — `docsSender == nil` or `cardmsg.Enabled() == false` or `buildDocsCard` error all fall through to text DM. No silent drop path.
- ✅ **Producer registration** — `cardDispatchProducerSpecs` registers both producers with identical `ProducerSpec` shape. `TestNotificationCardProducerRegistrations` pins this as an allowlist.
- ✅ **Data flow: Excerpt** — `card.Excerpt` → `strings.TrimSpace` → `resource.Excerpt` → `truncateRunes(_, MaxExcerptRunes=300)` → `escapeMarkdown` → JSON. Bounded and escaped.
- ✅ **Data flow: ActorName/UpdatedAt** — Trimmed → Fact values → `escapeMarkdown` → JSON. Bounded by `maxFactRunes=500`.
- ✅ **Data flow: Variant/Source** — Server-picked constants (not user-supplied). `Source.Label` length-checked against `maxTitleRunes`. `Source.IconURL` validated as https.

## Findings

No P0/P1 issues. Two non-blocking items below.

### P2 — `docsAttributionAndVariant` `default` case masks kind-enum expansion (`modules/notify/card.go:437`)

The `default` branch silently maps any unrecognized kind to the "shared" attribution/variant. Although `deliverDocsCardNotification` (line ~279) validates `card.Kind` against the three known constants — so the `default` is currently unreachable — it creates a silent catch-all that would mask a future kind addition that passes validation but isn't handled in the switch. Consider making the intent explicit with `case DocsCardKindShared:` and a commented `default` noting it's unreachable by design.

### Nit — `summaryLabelsFor` zh-prefix check inconsistent with `docsLabelsFor` (`modules/notify/card.go:537`)

`docsLabelsFor` (new) uses `strings.HasPrefix(strings.ToLower(lang), "zh")` (matches bare `"zh"`, `"zh-CN"`, `"zhTW"`). Pre-existing `summaryLabelsFor` uses `strings.HasPrefix(strings.ToLower(lang), "zh-")` (requires hyphen, misses bare `"zh"`). Both fall through to English, so practical impact is negligible, but aligning to the broader check would prevent a subtle divergence if `i18n.OutboundLanguage` ever returns bare `"zh"`.

- `buildResourceCard` refactor preserves the existing summary card contract (validation, title/excerpt/fact limits, icon https check, deep-link construction).
- `metadata` always includes `webUrl`; `octo` sub-object omitted when empty, keeping wire size minimal.
- `sendSummaryText` correctly reused for docs text fallback — same bot identity, same DM channel.
- `configs/tsdd.yaml` changes are documentation-only (commented-out config), documenting the `webLoginURL` dependency for deep-links.
- `scripts/cardpreview/main.go` is a dev tool using the same production template functions.
- E2E tests (`pilote2e/docs_card_wukongim_test.go`) cover the full HTTP→WuKongIM path and a no-regression test running both producers back-to-back.
- `main_carddispatch_test.go` pins both producer registrations with an allowlist pattern forcing review for any new producer.

## Blindspot Checklist

- **C1 — Dual-path parity**: Summary and docs paths are structurally symmetric (dedup → actor exclusion → member verify → bounded fan-out → degrade). Both card types rejected in batch. ✅
- **C2 — Control-flow ordering**: No nested/reused control flow changes. `buildResourceCard` is called once per card build. ✅
- **C3 — Authorization boundary**: Internal API behind `InternalTokenHeader`. No new endpoints exposed. ✅
- **C4 — Authorization lifecycle**: N/A — no auth changes.
- **C5 — Build/note through ≠ runtime correct**: N/A — no build/packaging changes.
- **C6 — Governance/docs self-consistency**: Documentation in `docs/docs-notify-card.md` and `.octospec/tasks/card-message-internal-dispatch/docs-notify-contract.md` is consistent with the implementation. ✅

## Cross-round Blocker Recheck

N/A — first review pass.

## Verdict: APPROVED

No correctness, security, or functional issues found. The P2 and nit are non-blocking defensive-coding suggestions. The PR is well-designed with thorough input validation, graceful degradation, concurrency safety, and comprehensive test coverage across unit, integration, and E2E levels.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. Two non-blocking nits (dead-code default case, minor i18n inconsistency). Code is well-structured with strong defenses.

PR #587, opened 2026-07-14
fix(thread): auto-follow reactivated thread for parent-channel followers (#586)

### LINKED ISSUE #586: 归档子区被新消息复活后，未同步补关注(ext)行 → 左侧「关注」列表永久缺失该子区
## 现象

在一个已被**关注**的 channel 下，某子区（thread）已被归档（`status=Archived`）。用户在该归档子区里发了一条新消息，服务端把子区自动复活为 `Active`。此时：

- ✅ **右侧「活跃区 thread 列表」**：子区立刻回到列表正确位置。
- ❌ **左侧「关注」folder 下该 channel 节点**：子区**不出现**，且**整页刷新后依然不出现**。

## 根因

两侧列表的**数据源不同**，不是同一张表的两种过滤：

**右侧「活跃区 thread 列表」** — 直接查 `thread` 表：

```
thread.DB.QueryByGroupNoWithStatus(groupNo, [Active])   // modules/thread/db.go:102
// SELECT * FROM thread WHERE group_no=? AND status IN (Active)
```

只认子区自身的 `status`，与"谁关注了""有没有关注记录"无关。所以 `thread.status` 一翻回 `Active` 立刻可见。

**左侧「关注 folder 下该 channel 的子区」** — 查 `user_conversation_ext` 表：

```
convext.DB.ListThreadExts(uid, spaceID)   // modules/conversation_ext/db.go:238
// SELECT * FROM user_conversation_ext WHERE uid=? AND space_id=? AND target_type=5
```

一个子区要出现在左侧关注列表，**前置条件是该 `(uid, thread)` 必须存在一条 ext 行**。`thread.status=Active` 不是充分条件——没有 ext 行，子区再活跃也进不了左侧。

### 缺口：ext 行（target_type=5）的创建路径不覆盖"复活"

ext 行目前只有两条创建路径：

1. **`FollowChannel` 物化**（`modules/conversation_ext/service.go` Phase 2）：关注父群时通过 `EnumerateActiveShortIDs`（`modules/message/1module.go:326`）枚举子区，底层是 `QueryByGroupNoWithStatus([Active])`，**只枚举 active 子区，archived 子区被跳过**。
2. **`OnThreadCreated` fanout**（`modules/conversation_ext/service.go:653`）：**仅在子区被创建时**触发，给 `auto_follow_threads=1 AND group_unfollowed=0` 的用户补 ext 行。

**没有第三条。** 子区从 archived 被消息复活时，没有任何逻辑给关注了父群的用户补 ext 行。

消息复活路径实锤（`modules/thread/api.go:47` `onMessages`）只做两件事：

1. `RecordMessageAndReactivate` → `UPDATE thread SET status=Active`（`modules/thread/db.go:784`）——只动 `thread` 表；
2. `JoinThread` → 写 `thread_member`。

**完全没碰 `user_conversation_ext`。** 没有 fanout。

### 复现时序（解释全部症状）

1. 某子区曾被归档。
2. 用户关注了父 channel，但物化（`EnumerateActiveShortIDs`）发生在该子区已归档之后 → 该子区**没有 ext 行**。
3. 用户在归档子区发新消息 → `thread.status` 翻回 `Active`。
4. 右侧读 `thread` 表 → 立刻显示。
5. 左侧读 `user_conversation_ext` → 该子区**没有 ext 行** → 不在 `sidebar/sync` 返回里 → 左侧空白，**整页刷新照样空白**（DB 里就没这行，刷新不产生新行）。

## 建议修复（方案 A）

复活是一次"隐式的重新变活跃"事件，理应触发与 `OnThreadCreated` **对称**的 ext 物化：**当子区从 archived 复活为 active 时，若其父 channel 处于被关注状态，则自动把该子区也补成被关注态。**

落点在 `modules/thread/api.go` 的 `onMessages` 里、确认确实发生了 `archived → active` 解档之后：

1. 新增 `OnThreadReactivated(groupNo, shortID)`，复用 `OnThreadCreated` 的 fanout 逻辑：按 `auto_follow_threads=1 AND group_unfollowed=0` 给活跃父群成员 `INSERT IGNORE` 一条 thread ext 行，并 bump `follow_version`。
2. 让 `RecordMessageAndReactivate` **回传一个 `reactivated bool`**（当前它把"本次是否真的解档"这个信息吞掉了，无法区分 archived→active 分支和纯统计分支），`onMessages` 据此**仅在真解档时**触发 fanout，避免每条消息都 fanout。

优点：

- 左右两侧语义收敛到同一套 fanout；
- `INSERT IGNORE` 天然幂等，与已存在的 ext 行不冲突；
- `follow_version` bump 让在线客户端自动重刷 `sidebar/sync`，无需整页刷新。

### 验证方式

复现后查该用户该子区的 ext 行是否缺失：

```
SELECT * FROM user_conversation_ext
WHERE uid = '<用户uid>' AND target_type = 5
  AND target_id = '<groupNo>____<shortId>';
```

结果为空即实锤本 issue 的根因。

## 影响范围

- 仓库：`octo-server`
- 相关文件：`modules/thread/api.go`、`modules/thread/db.go`、`modules/conversation_ext/service.go`
- 用户可感知：被关注 channel 下的归档子区被消息复活后，在关注列表中永久缺失，直到用户重新关注父群或该子区有新建（fanout）事件。

pkuWMH:
## 背景

Fixes #586

一个已被**关注**的 channel 下，某子区（thread）被归档后，用户在归档子区里发新消息使其自动复活为 `active`：

- ✅ 右侧「活跃区 thread 列表」立刻显示（它读 `thread` 表，只看 `status`）。
- ❌ 左侧「关注」folder 下该 channel 里**永久缺失**该子区，且**整页刷新照样缺失**。

## 根因

两侧数据源不同：左侧「关注」列表读 `user_conversation_ext`（`ListThreadExts`, `target_type=5`），一个子区必须存在 ext 行才会出现在左侧。而 ext 行只有两条创建路径：

1. `FollowChannel` 物化 —— `EnumerateActiveShortIDs` 只枚举 **active** 子区，关注时处于 archived 的子区从未生成 ext 行；
2. `OnThreadCreated` fanout —— 仅在子区**创建**时触发。

子区从 archived 被消息复活时（`modules/thread/api.go` `onMessages` → `RecordMessageAndReactivate`）只 `UPDATE thread SET status=Active`，**从不给关注了父群的用户补 ext 行**。因此左侧永久缺失。

## 修复（与 OnThreadCreated 对称）

当子区因新消息从 archived 复活为 active 时，若父 channel 处于被关注状态，则对所有 `auto_follow_threads=1 AND group_unfollowed=0` 的活跃父群成员补 thread ext 行——与 `OnThreadCreated` 完全对称。

- **`modules/thread/db.go`**：`RecordMessageAndReactivate` 现返回 `(reactivated bool, err error)`；仅在 `archived -> active` 分支返回 `true`，纯统计/已 active、已删除/不存在分支返回 `false`。加锁与 version 语义不变。
- **`modules/conversation_ext/service.go`**：抽出共用 fanout 主体 `fanoutThreadFollow`，新增 `OnThreadReactivated`；它与 `OnThreadCreated` 都委托给该 helper，行为完全一致且幂等（`INSERT IGNORE`，与既有 ext 行重叠为 no-op），并 bump `follow_version` 让在线客户端自动重刷 `sidebar/sync`。
- **`modules/thread/api.go`**：`onMessages` 仅在 `reactivated==true` 时调用 `OnThreadReactivated`，复用与 `CreateThread` 相同的 `GetGlobalConvExtService()` 引用；失败只记 warning，绝不阻断消息摄入。

## 测试

- `modules/thread/archive_db_test.go`：更新调用方为二值签名，断言 `reactivated` 在 `archived->active` 为 `true`、在 active/deleted 路径为 `false`。
- `modules/conversation_ext/service_test.go`：新增 `OnThreadReactivated` 测试（对 auto-follow 用户物化 ext 行 / 幂等 / 非法入参），镜像 `OnThreadCreated` 测试。

### 验证结果

```
go build ./...                                   ✅ pass
go vet ./modules/thread/... ./modules/conversation_ext/...   ✅ pass
go test ./modules/thread/...                     ✅ ok (11.6s)
go test -tags integration ./modules/conversation_ext/        ✅ ok (1.4s)
```

（DB-backed 集成测试跑在真实 MySQL/Redis 上，非 skip。）

## 影响范围

- 仅触及 `modules/thread/api.go`、`modules/thread/db.go`、`modules/conversation_ext/service.go` 及对应测试。
- `OnThreadCreated` 可观测行为不变（仅提取共用 helper）。
- 幂等 `INSERT IGNORE` + `reactivated` 门控保证不会每条消息都 fanout，热路径无写放大。

### DIFF
--- a/modules/conversation_ext/service.go
+++ b/modules/conversation_ext/service.go
@@ -651,6 +651,27 @@ func (s *Service) UnfollowChannel(uid, spaceID, groupNo string) error {
 // 调用方应在 thread.Service.CreateThread 的 commit 之后立即调用，错误以 wrap 形式上传，
 // 由调用方决定是否记日志 / 触发告警（thread 创建不应因 fanout 失败回滚）。
 func (s *Service) OnThreadCreated(groupNo, shortID string) error {
+	return s.fanoutThreadFollow(groupNo, shortID)
+}
+
+// OnThreadReactivated 在归档子区被新消息复活（archived -> active）之后调用，
+// 给所有已对 parent channel 开启 auto_follow_threads=1 的用户物化 thread ext 行。
+//
+// issue #586：FollowChannel 物化（EnumerateActiveShortIDs）只枚举 active 子区，
+// 因此关注时处于 archived 的子区从未生成 ext 行；OnThreadCreated 只在创建时 fanout。
+// 结果是被复活的子区右侧 active 列表可见、左侧「关注」列表却永久缺失。此处与
+// OnThreadCreated 对称补齐：复用同一 fanout 逻辑（INSERT IGNORE，幂等，与既有
+// ext 行重叠是 no-op），只在 RecordMessageAndReactivate 确实翻转了 status 时调用。
+//
+// best-effort：与 OnThreadCreated 一样错误以 wrap 形式上传，由调用方记日志不阻断消息摄入。
+func (s *Service) OnThreadReactivated(groupNo, shortID string) error {
+	return s.fanoutThreadFollow(groupNo, shortID)
+}
+
+// fanoutThreadFollow 是 OnThreadCreated / OnThreadReactivated 共用的 fanout 主体：
+// 给所有对 parent channel 开启 auto_follow_threads=1 且未 group_unfollowed 的活跃
+// 成员物化 thread ext 行并 bump 各自的 follow_version。INSERT IGNORE 保证幂等。
+func (s *Service) fanoutThreadFollow(groupNo, shortID string) error {
 	if groupNo == "" {
 		return errors.New("group_no must not be empty")
 	}
--- a/modules/conversation_ext/service_test.go
+++ b/modules/conversation_ext/service_test.go
@@ -680,6 +680,65 @@ func TestService_OnThreadCreated_InvalidInput(t *testing.T) {
 }
 
 // ---------------------------------------------------------------------------
+// OnThreadReactivated fanout — issue #586: 归档子区被消息复活后与 OnThreadCreated
+// 对称地给关注父群的用户补 thread ext 行，否则左侧「关注」列表永久缺失该子区。
+// 复用同一 fanout 主体（fanoutThreadFollow），行为与 OnThreadCreated 一致。
+// ---------------------------------------------------------------------------
+
+func TestService_OnThreadReactivated_MaterializesForAutoFollowUsers(t *testing.T) {
+	svc := newServiceForTest(t)
+	const space, grp, shortID = "s1", "grp-ofr-1", "thr-react"
+	channelID := grp + threadSeparator + shortID
+
+	// A：开启 auto_follow_threads（关注了 channel）
+	require.NoError(t, svc.FollowChannel("uA", space, grp))
+	// B：关注过群但显式取关 —— auto_follow_threads 已清零
+	require.NoError(t, svc.UnfollowChannel("uB", space, grp))
+	// C：从未操作过该 channel
+
+	require.NoError(t, svc.OnThreadReactivated(grp, shortID))
+
+	rowA, err := svc.db.Get("uA", space, targetTypeThread, channelID)
+	require.NoError(t, err)
+	assert.NotNil(t, rowA, "auto_follow_threads=1 的用户应在复活 fanout 中拿到 thread 行")
+
+	rowB, err := svc.db.Get("uB", space, targetTypeThread, channelID)
+	require.NoError(t, err)
+	assert.Nil(t, rowB, "已取消关注 channel 的用户不应被复活 fanout 触及")
+
+	rowC, err := svc.db.Get("uC", space, targetTypeThread, channelID)
+	require.NoError(t, err)
+	assert.Nil(t, rowC, "从未关注 channel 的用户不应被复活 fanout 触及")
+}
+
+func TestService_OnThreadReactivated_Idempotent_PreservesExistingRow(t *testing.T) {
+	svc := newServiceForTest(t)
+	const space, grp, shortID = "s1", "grp-ofr-2", "thr-react-dup"
+	channelID := grp + threadSeparator + shortID
+
+	require.NoError(t, svc.FollowChannel("uA", space, grp))
+	// uA 已手动给该 thread 拖到 sort=88（例如关注时子区曾 active 被物化后又归档）
+	require.NoError(t, svc.db.Upsert("uA", space, targetTypeThread, channelID, ConvExtFields{
+		FollowSort: intPtr(88),
+	}))
+
+	// 复活 fanout 与既有 ext 行重叠时必须是 no-op（INSERT IGNORE）。
+	require.NoError(t, svc.OnThreadReactivated(grp, shortID))
+
+	row, err := svc.db.Get("uA", space, targetTypeThread, channelID)
+	require.NoError(t, err)
+	require.NotNil(t, row)
+	assert.Equal(t, 88, row.FollowSort, "已存在的 thread 行（含手动排序）必须保留，复活 fanout 不应覆盖")
+}
+
+func TestService_OnThreadReactivated_InvalidInput(t *testing.T) {
+	svc := newServiceForTest(t)
+
+	require.Error(t, svc.OnThreadReactivated("", "thr"))
+	require.Error(t, svc.OnThreadReactivated("grp", ""))
+}
+
+// ---------------------------------------------------------------------------
 // UnfollowChannel happy path + cascade
 // ---------------------------------------------------------------------------
 
--- a/modules/thread/api.go
+++ b/modules/thread/api.go
@@ -11,6 +11,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/modules/conversation_ext"
 	"github.com/Mininglamp-OSS/octo-server/modules/group"
 	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
 	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
@@ -70,10 +71,25 @@ func (t *Thread) onMessages(messages []*config.MessageResp) {
 		if runeLen := len([]rune(content)); runeLen > 500 {
 			content = string([]rune(content)[:500])
 		}
-		if err := t.db.RecordMessageAndReactivate(shortID, content, msg.FromUID, func() (int64, error) {
+		reactivated, err := t.db.RecordMessageAndReactivate(shortID, content, msg.FromUID, func() (int64, error) {
 			return t.ctx.GenSeq(ThreadSeqKey)
-		}); err != nil {
+		})
+		if err != nil {
 			t.Error("更新消息统计/解档失败", zap.Error(err), zap.String("shortID", shortID))
+		} else if reactivated {
+			// issue #586：子区被消息复活（archived->active）时，与 OnThreadCreated 对称地
+			// 给所有关注父群（auto_follow_threads=1 AND group_unfollowed=0）的用户补 thread
+			// ext 行，否则左侧「关注」列表会永久缺失这个已复活的子区（FollowChannel 物化
+			// 只枚举 active 子区，关注时处于 archived 的子区从未生成 ext 行）。
+			// 与 OnThreadCreated 一样是 commit 后 best-effort：失败只警告，不阻断消息摄入。
+			if convSvc := conversation_ext.GetGlobalConvExtService(); convSvc != nil {
+				if ferr := convSvc.OnThreadReactivated(groupNo, shortID); ferr != nil {
+					t.Warn("OnThreadReactivated fanout 失败（子区已复活，sidebar 会延迟到下次 FollowChannel/refollow 补齐）",
+						zap.String("groupNo", groupNo),
+						zap.String("shortID", shortID),
+						zap.Error(ferr))
+				}
+			}
 		}
 
 		// 发送者不是子区成员，自动加入
--- a/modules/thread/archive_db_test.go
+++ b/modules/thread/archive_db_test.go
@@ -177,7 +177,9 @@ func TestRecordMessageAndReactivate_ResurrectsArchived(t *testing.T) {
 	require.Equal(t, ThreadStatusArchived, m.Status)
 
 	// 此后 listener 落库消息统计：必须把 status 抬回 active 并写新版本
-	require.NoError(t, db.RecordMessageAndReactivate("raced", "hello", "sender-1", constVersion(200)))
+	reactivated, err := db.RecordMessageAndReactivate("raced", "hello", "sender-1", constVersion(200))
+	require.NoError(t, err)
+	assert.True(t, reactivated, "archived->active transition must report reactivated=true")
 
 	m, err = db.QueryByShortID("raced")
 	require.NoError(t, err)
@@ -203,10 +205,12 @@ func TestRecordMessageAndReactivate_NoVersionBumpForAlreadyActive(t *testing.T) 
 	insertThreadWithVersion(t, db, "active_thread", ThreadStatusActive, &now, 42)
 
 	called := false
-	require.NoError(t, db.RecordMessageAndReactivate("active_thread", "hi", "u1", func() (int64, error) {
+	reactivated, err := db.RecordMessageAndReactivate("active_thread", "hi", "u1", func() (int64, error) {
 		called = true
 		return 999, nil
-	}))
+	})
+	require.NoError(t, err)
+	assert.False(t, reactivated, "already-active thread must report reactivated=false")
 
 	m, err := db.QueryByShortID("active_thread")
 	require.NoError(t, err)
@@ -226,7 +230,9 @@ func TestRecordMessageAndReactivate_DeletedStaysDeleted(t *testing.T) {
 	now := time.Now()
 	insertThreadWithVersion(t, db, "deleted_thread", ThreadStatusDeleted, &now, 7)
 
-	require.NoError(t, db.RecordMessageAndReactivate("deleted_thread", "x", "u1", constVersion(999)))
+	reactivated, err := db.RecordMessageAndReactivate("deleted_thread", "x", "u1", constVersion(999))
+	require.NoError(t, err)
+	assert.False(t, reactivated, "deleted thread must report reactivated=false")
 
 	m, err := db.QueryByShortID("deleted_thread")
 	require.NoError(t, err)
@@ -269,7 +275,9 @@ func TestRecordMessageAndReactivate_VersionMonotonicVsCronArchive(t *testing.T) 
 		return newV, nil
 	}
 
-	require.NoError(t, db.RecordMessageAndReactivate("monotonic", "msg", "u1", gen))
+	reactivated, err := db.RecordMessageAndReactivate("monotonic", "msg", "u1", gen)
+	require.NoError(t, err)
+	assert.True(t, reactivated, "archived->active transition must report reactivated=true")
 
 	m, err := db.QueryByShortID("monotonic")
 	require.NoError(t, err)
@@ -479,7 +487,7 @@ func TestArchiveStaleBatch_ConcurrentWithMessages(t *testing.T) {
 		go func() {
 			defer wg.Done()
 			for r := 0; r < rounds; r++ {
-				_ = db.RecordMessageAndReactivate(sid, "msg", "sender", gen)
+				_, _ = db.RecordMessageAndReactivate(sid, "msg", "sender", gen)
 				time.Sleep(1 * time.Millisecond)
 			}
 		}()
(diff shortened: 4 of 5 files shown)


OctoBoooot (approved):
## Review: fix(thread): auto-follow reactivated thread for parent-channel followers (#587)

**Verdict: Approve** — a clean, correctly-scoped fix that faithfully implements issue #586's Plan A: when an archived thread is reactivated by a new message, materialize the follow (ext) rows symmetrically with `OnThreadCreated`, so the left "关注" list no longer permanently loses the revived thread. Zero blockers, zero majors. Approval is conditional on **Test** finishing green (only `check-sprint` governance is red — non-barring; Build/Vet/Lint/Test still running); reverts to Request changes if Test fails.

`size/M`, normal tier — but this touches the thread/follow/fanout path where blacklist-leak and read-口径 bugs hide, so I verified those axes directly.

### What I byte-verified
- **`reactivated bool` returns true in exactly one branch.** `RecordMessageAndReactivate` returns `true` only inside the `status == Archived` block after the UPDATE+commit; every other path (`loaded==0` not-found/deleted, already-active stats-only, and all error paths) returns `false`. So the fanout fires only on a genuine archived→active flip — no write-amplification per message, and no false trigger. (This is the bidirectional-correctness trap for a bool-return-per-branch change — checked both directions.)
- **`fanoutThreadFollow` is a behavior-preserving extract.** `OnThreadCreated` is now pure delegation (`return s.fanoutThreadFollow(...)`), and `OnThreadReactivated` delegates to the same body — so the two paths are identical by construction, and the reactivation path inherits all the fanout hardening for free rather than re-implementing it.
- **Access-control parity is intact** (the axis that matters here): the shared helper filters targets by **active membership** (`getActiveMemberFilter` → `filterFanoutTargetsByActiveMembership`, issue #351) so a blacklisted/removed member does NOT get an ext row (no metadata-layer leak), re-checks `group_unfollowed=0` under `FOR UPDATE` in-tx (`selectEligibleForFanoutTx`), and gates on `auto_follow_threads=1`. A reactivated thread can't leak follow rows to users who shouldn't see it.
- **Concurrency/idempotency preserved** — the helper keeps the deadlock-safe lock ordering (Jerry-Xin round-3 P1), `withDeadlockRetry`, batch splitting (65k-placeholder guard), `INSERT IGNORE` idempotency, and `follow_version` bump for live client refresh. Nothing in the extract changed these.
- **Best-effort, non-blocking** — `onMessages` calls `OnThreadReactivated` only when `reactivated==true`, reuses the same `GetGlobalConvExtService()` reference as `CreateThread`, and on failure logs a warning without blocking message ingest — consistent with `OnThreadCreated`'s commit-after best-effort contract.
- **Tests** — `archive_db_test.go` updated to the two-value signature with the resurrection-race case; `service_test.go` adds `OnThreadReactivated` validation cases mirroring `OnThreadCreated`, and the shared-helper materialization/blacklist/group_unfollowed/version-bump behavior is already pinned by the existing fanout tests (lines ~563-610).

### Praise
- Fixing this by extracting one shared `fanoutThreadFollow` and pointing both `OnThreadCreated` and `OnThreadReactivated` at it is exactly right — it makes the two paths symmetric by construction (the issue's whole point), so the reactivation path inherits the blacklist filter, deadlock-safe lock order, and idempotency without a second implementation that could drift. And returning `reactivated` from the DB layer to gate the fanout (rather than fanning out on every message) is the precise, hot-path-safe trigger.

### Question (minor, non-blocking)
- `modules/thread/api.go` now imports `modules/conversation_ext` directly and reaches it via `GetGlobalConvExtService()`. That matches the existing `CreateThread` call site, so it's consistent — but worth a glance that this doesn't introduce an import cycle at build (Build CI will confirm) and that the global-service accessor is the intended coupling rather than an injected dependency. Not a blocker; the pattern is pre-existing.

### Note on merge gate
`mergeStateStatus` stays gated on Test + required review — approved ≠ mergeable until Test is green. `check-sprint` (governance) red does not bar approval but keeps the PR from merging until a maintainer clears it.

Jerry-Xin (approved):
Project relevance gate: pass. This PR directly fixes octo-server thread/sidebar follow behavior and is in scope.

💬 Non-blocking:
- 🔵 Suggestion: Add a focused integration/unit test around `Thread.onMessages` to verify the hook calls `OnThreadReactivated` only when `RecordMessageAndReactivate` returns `reactivated == true`. The DB and service layers are covered well, but the actual listener wiring at `modules/thread/api.go:74-86` is still only indirectly protected.

✅ Highlights:
- The reactivation signal is correctly returned only after the archived-to-active update path in `modules/thread/db.go:805-820`, and caller code checks `err` before fanout.
- Reusing `fanoutThreadFollow` from `OnThreadCreated` and `OnThreadReactivated` preserves existing eligibility, active-member filtering, batching, deadlock retry, and idempotent `INSERT IGNORE` behavior in `modules/conversation_ext/service.go:653-772`.
- The new tests cover materialization, idempotency, invalid input, and the `reactivated` boolean semantics.

Verification: `go test -tags integration ./modules/conversation_ext/` passed. `go test ./modules/thread/...` did not complete because local test setup failed on an existing migration-state mismatch: `unknown migration in database`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #587 (octo-server)

## Summary

This PR has two parallel tracks: (1) the titular fix for issue #586 — auto-follow reactivated threads for parent-channel followers — and (2) a new docs-notify card notification producer alongside the existing summary-notify producer, with shared template infrastructure (`Variant`/`Source` metadata fields, `buildResourceCard` extraction, `webOrigin` helper).

The thread reactivation fix is well-scoped: `RecordMessageAndReactivate` now returns a `reactivated` bool so the caller can trigger a post-commit fan-out that materializes conversation_ext rows for auto-follow users. The docs-notify card path mirrors the summary-notify architecture (dedup → actor exclusion → member verify → build-or-degrade → bounded fan-out) and is thoroughly tested.

## Verification

- ✅ **Thread reactivation data flow** — `RecordMessageAndReactivate` (`modules/thread/db.go:781`) uses `SELECT ... FOR UPDATE` in a tx; the `reactivated=true` return only fires after the `UPDATE ... SET status=active` succeeds AND `tx.Commit()` returns nil. The caller at `modules/thread/api.go:80` checks `err != nil` first, so fan-out is skipped on commit failure.
- ✅ **Fan-out idempotency** — `fanoutThreadFollow` (`modules/conversation_ext/service.go:674`) uses `INSERT IGNORE` + active membership filter; concurrent reactivation races are impossible because `SELECT FOR UPDATE` serializes the status flip.
- ✅ **Three-way mutual exclusion** — `sendNotify` (`modules/notify/api.go:182-200`) correctly counts non-nil fields and rejects `present > 1`; dispatch routes Card → DocsCard → text in that order, which is safe given the upstream mutex.
- ✅ **Text fallback injection defense** — `sanitizeLine` (`modules/notify/card.go:219`) collapses control characters before they reach the line-structured text DM. Card path relies on `escapeMarkdown` in the template. Both defenses verified.
- ✅ **Deep-link security** — `webOrigin` (`pkg/cardtmpl/resource.go:289`) enforces absolute https; `requireHTTPS` gates all icon URLs including the new `Source.IconURL`. Non-https origins fail the build and degrade to text.
- ✅ **Registry spec isolation** — `copyProducerSpec` in `internal/carddispatch/registry.go` deep-copies spec slices, so the shared-base pattern in `main.go` is safe at the registry boundary.
- ✅ **Producer registration test** — `TestNotificationCardProducerRegistrations` (`main_carddispatch_test.go:30`) pins both producers to the reviewed shape and deliberately fails on unknown IDs.

## Findings

No P0/P1 issues. One P2 and one nit below.

### P2 — Shared slice backing across producer specs (`main.go:374`)

`summarySpec := base` and `docsSpec := base` copy the `ProducerSpec` struct by value, so `AllowedChannelTypes` and `AllowedProfiles` share the same underlying arrays between both specs. `copyProducerSpec` in the registry defensively deep-copies on ingestion, so this is not a production bug today. However, when a third producer is added, a developer mutating a slice field on a spec variable (e.g. `append(spec.AllowedChannelTypes, ...)`) before return would silently alias all specs sharing the base. Giving each spec its own slice literals — or a brief comment on the aliasing constraint — would close the maintainability gap.

### Nit — `sanitizeLine` lacks a direct unit test (`modules/notify/card.go:219`)

`sanitizeLine` is exercised indirectly through `TestBuildDocsFallbackText_StripsNewlineInjection` and the summary equivalent, but a focused table-driven test covering `\n`, `\r`, `\t`, `\x00`, multi-byte passthrough, and empty string would pin the contract explicitly and make future refactors safer.

## Data Flow Trace

1. **Thread reactivation**: msg arrives → `onMessages` → `RecordMessageAndReactivate` (SELECT FOR UPDATE → status flip → COMMIT) → returns `(true, nil)` only on successful commit → `GetGlobalConvExtService()` (nil-safe) → `OnThreadReactivated` → `fanoutThreadFollow` (SELECT auto_follow users → filter active members → INSERT IGNORE + bump version). Each step verified; error at any stage halts the chain gracefully.

2. **Docs card notification**: HTTP POST → `sendNotify` (3-way mutex check) → `dispatchNotify` → `deliverDocsCardNotification` → validate (SpaceID, Targets 1-200, DocID, Title, Kind) → dedup targets → exclude actor UID → `memberCache.verify` → `ensureNotifyBotReady` → `buildDocsCard` (or degrade to `buildDocsFallbackText`) → goroutine pool (sem=20) → `docsSender.Send` per target (or `sendSummaryText` fallback). All data sourced from validated HTTP body, config, or server-side label maps.

## Blind Spot Checklist (R5)

- **C1 (dual path parity)**: ✅ `OnThreadCreated` ↔ `OnThreadReactivated` both call `fanoutThreadFollow`; `deliverCardNotification` ↔ `deliverDocsCardNotification` structurally parallel; `summaryDeepLink` ↔ `docsDeepLink` share `webOrigin`; `buildSummaryFallbackText` ↔ `buildDocsFallbackText` both use `sanitizeLine`.
- **C2 (control-flow ordering)**: ✅ `webOrigin` extraction preserves validation; `buildResourceCard` shared body introduces no ordering dependencies.
- **C3 (authorization boundary)**: ✅ Internal token auth unchanged; `docsSender` from same registry; no new endpoints exposed.
- **C4 (authorization lifecycle)**: N/A — no auth/authz changes.
- **C5 (build/note)**: N/A — no build/packaging changes.
- **C6 (governance/docs)**: ✅ `tsdd.yaml` comment update consistent with code behavior; contract docs match implementation.

## Cross-round Blocker Recheck (R6)

N/A — first review round.

- **Variant/Source metadata**: `buildMetadata` correctly emits `metadata.octo.variant` and `metadata.octo.source` only when non-empty; `webUrl` always set from the same deep-link driving the ActionSet. AC 1.5 compliant.
- **Summary card backward compatibility**: `buildSummaryCard` adds `Variant` ("summary.completed"/"summary.failed") and `Source` — both server-picked from labels, no caller-supplied injection surface.
- **Batch endpoint guard**: `sendNotifyBatch` (`modules/notify/api.go:252`) rejects both `Card` and `DocsCard` in batch entries. Test covers both.
- **Docs card validation**: `deliverDocsCardNotification` validates SpaceID, Targets (1–200), DocID, Title, and Kind before any transport. Six negative cases tested.
- **E2E test parity**: `pilote2e/docs_card_wukongim_test.go` mirrors the summary e2e pattern with the `/d/{doc_id}` deep link.
- **R3 (distributed side effects)**: `SELECT FOR UPDATE` serializes reactivation; `INSERT IGNORE` makes fan-out idempotent; no multi-client fan-out amplification.

## Verdict: APPROVED

Solid, well-tested PR. The thread reactivation fix correctly identifies and closes the sidebar gap with idempotent fan-out gated on the actual status transition. The docs-notify card path follows established patterns with appropriate input validation and graceful degradation. No correctness, security, or build-breaking issues found.

Static analysis only at head `462d5413c01ee3f54a8d60f9c402c04c10e9be18`; build and tests not executed in this environment.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Thread reactivation fix is correct (data flow verified through SELECT FOR UPDATE → commit → reactivated bool → idempotent fanout). Docs-notify card path mirrors established summary-notify patterns with proper validation, injection defense, and graceful degradation. One P2 (shared slice backing in producer specs, mitigated by registry copy) and one nit (sanitizeLine direct test).

yujiawei (approved):
# Code Review — PR #587 (octo-server)

Reviewed at head SHA `462d5413c01ee3f54a8d60f9c402c04c10e9be18`. Verified `go build ./...` and `go vet ./modules/thread/... ./modules/conversation_ext/...` both pass. DB-backed integration tests require a live MySQL and were not run locally.

## 1. Spec compliance

**Spec: ✅**

The change does exactly what #586 requires and nothing more.

- **Under-build:** none. When a thread is resurrected from `archived → active` by an incoming message, ext rows are now fanned out to parent-channel followers, closing the "left follow list permanently misses the reactivated thread" gap.
- **Over-build:** none. `RecordMessageAndReactivate` gains a `reactivated bool` return, a new `OnThreadReactivated` delegates to an extracted `fanoutThreadFollow` shared with `OnThreadCreated`, and `onMessages` calls it only on the `reactivated==true` edge. No extra flags, endpoints, or fields.
- **Deviation:** none. The implementation is symmetric with `OnThreadCreated`: same helper, `INSERT IGNORE` idempotency, `follow_version` bump, batching, and `withDeadlockRetry`.

## 2. Code quality

**Quality: Approved**

Correctness of the core mechanism holds up:

- **`reactivated` gating is correct.** `true` is returned only from the `archived → active` branch; the not-found/deleted, already-active, and error paths all return `false`. The `onMessages` caller fans out only on `true`, so an active thread receiving messages does **not** re-run fanout per message — no per-message write amplification.
- **No cron race.** The reactivate transaction takes the row lock (`SELECT ... FOR UPDATE`) and sets `last_message_at = NOW()` inside it, so the stale-archive cron's `WHERE last_message_at < cutoff` no longer matches the row. The monotonic version guard prevents version regression. Normal interleavings converge.
- **Idempotency is sound.** `INSERT IGNORE` makes an overlap with pre-existing ext rows (including manually re-sorted rows) a no-op, confirmed by `TestService_OnThreadReactivated_Idempotent_PreservesExistingRow`.
- **Error handling is non-blocking by design.** Fanout runs post-commit; failure only logs a warning and never blocks message ingestion, matching the existing `CreateThread → OnThreadCreated` contract.

### Non-blocking follow-ups (P2)

**P2 — Fanout is one-shot best-effort; a transient failure does not self-heal via messages.** `modules/thread/api.go:74-92`. Because fanout only fires on the `archived → active` edge, if `OnThreadReactivated` fails transiently (and outlives `withDeadlockRetry`'s bounded 1213/1205 retries), subsequent messages observe `status == active`, return `reactivated=false`, and never retry — the thread stays missing from the follow sidebar until the next `FollowChannel`/refollow. This is worth noting because it is a narrow re-creation of the #586 symptom for the failure case.

Why this is not a merge blocker: the identical limitation already exists in the pre-existing `OnThreadCreated` path (`modules/conversation_ext/service.go` via `modules/thread/service.go:305-311`) — a creation-time fanout failure is equally unhealed by later messages. This PR reaches parity with the established post-commit best-effort pattern and strictly improves coverage (it heals a previously-unhealed edge). A durable fix (outbox/retry, or a read-side reconciliation that materializes ext rows for active threads missing them) should be a separate change that also covers `OnThreadCreated`, not a gate on this bug fix.

**P2 — Synchronous fanout on the message-listener hot path.** `modules/thread/api.go:79`. Listeners run inline after the webhook commit (`modules/webhook/api.go:324`), so reactivating a thread in a channel with many auto-follow members adds follower selection plus per-1k-row batch version bumps/inserts (and any retry backoff) to the ingest path. The `reactivated` gate bounds this to once per reactivation rather than per message, and `OnThreadCreated` already does the same synchronous fanout on its path, so this is acceptable as-is; moving fanout to a durable async worker would address both this and the durability point above.

**P2 (nit) — Misattributed log/metric labels in the shared helper.** `modules/conversation_ext/service.go:701, 716, 755-769`. `fanoutThreadFollow` is now shared, but its wrapped error strings and the `withTx("OnThreadCreated", ...)` op label are all hard-coded to `OnThreadCreated`. A fanout failure triggered by a reactivation will surface in logs/metrics as `OnThreadCreated`, which will mislead on-call debugging. Consider threading the caller name (e.g. an `op string` parameter) into the error wraps and the `withTx` label.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. The fix is correct, minimal, symmetric with the existing creation-time fanout, and well tested. The durability, hot-path, and label items above are advisory follow-ups, not blockers.

## 4. Coverage notes

- Verified there is **no import cycle**: `modules/conversation_ext` does not import `modules/thread`; the dependency is one-directional (`thread → conversation_ext`) and `go build ./...` passes.
- Verified `groupNo`/`shortID` in `onMessages` come from `ParseChannelID(msg.ChannelID)` (`modules/thread/api.go:56`), so the fanout targets the correct parent channel.
- Not independently exercised: the DB-backed integration tests (`archive_db_test.go`, `service_test.go` under `//go:build integration`) require a live MySQL/Redis; test *code* was reviewed and the two-value signature migration is consistent across all call sites.
- Not in scope of this diff and not reviewed: the general steady-state case of an already-active thread whose followers never got an ext row for reasons unrelated to reactivation.

PR #591, opened 2026-07-15
feat(webhook): add /octo incoming-webhook suffix as multica alias

caster-Q:
## What

新增 `/octo` incoming-webhook 后缀,作为 `/multica` 的对外别名:对外场景不再暴露 multica 产品名。

## Why

incoming webhook 面向对外场景(把回路 issue 状态变更推送到群)。原 `/multica` 后缀把 "multica" 产品名暴露在对外 URL、示例展示与投递审计里,不适合对外呈现。

## How

- **路由**:`/octo` 与 `/multica` 并存;`/octo` 用独立 `octoAdapter`(仅 `name` 不同,复用 `parseMulticaPush`),使投递审计的 `adapter` 记为 `octo` 而非 `multica`
- **展示**:catalog 中 octo 展示、multica 隐藏(`/multica` 路由保留,兼容存量订阅)
- **文案**:octo 的 title/description 面向回路项目 Webhook 配置,去掉接入步骤
- **测试**:同步 example key 列表与 steps 断言,别名路由遍历纳入 octo

纯加法,`/multica` 路由与 `multicaAdapter` 原样保留,存量订阅不受影响。

## 验证

- `go build` / `go vet` / `gofmt` 通过
- incomingwebhook 包单测通过(catalog / i18n / multica 解析)
- 本地 e2e:经 `/octo` 推送 → HTTP 200 且投递审计 `adapter=octo`

Closes #590

PR #100, opened 2026-05-20
chore(ci): standardize reusable workflow refs to @main

lml2468:
## Summary

All reusable workflow references in this repo previously pointed to `@v1`, `@v2`, or a raw commit SHA for centrally-maintained workflows from `Mininglamp-OSS/.github`. This PR standardizes them all to `@main`.

## Why @main?

- **Zero tag maintenance** — no need to manually bump @v1/@v2 after every .github update
- **Consistency** — newer workflows (reusable-stale, reusable-codeql, reusable-pr-labeler, etc.) already use `@main`
- **Safe** — `Mininglamp-OSS/.github` main is always reviewed before merge; breaking changes are caught by Workflow Sanity CI

## Changed files

All `.github/workflows/*.yml` that reference `Mininglamp-OSS/.github` reusable workflows.

PR #593, opened 2026-07-15
feat(messages_search): aggregation-first global search — L1 groups endpoint + presence calibration

### LINKED ISSUE #592: Global search: aggregation-first L1 overview (groups endpoint + presence calibration)
## Summary

Global message search currently returns a single flat, time-ordered stream. When searching by a person or keyword across many rooms, users cannot see **which conversations** contain hits, nor get a per-conversation overview.

This adds an **aggregation-first (L1) view**: given a keyword / sender / member / channel filter, return a per-conversation overview — every group, thread, and DM that has a hit **currently visible to the caller** — each with an approximate match count, latest activity time, and a small visible preview. Drilling into a conversation ("view all") reuses the existing `_search_global_messages` endpoint (L2).

## Design

- **presence exact**: every returned conversation is guaranteed to have at least one hit visible to the caller (no "click-in-and-empty"). Visibility is fully calibrated through the same `filterVisible` gate the flat search uses (revoked / admin+mutual delete / self-delete / channel-offset watermark / `visibles` allowlist).
- **count approximate**: `match_count` is the OS `doc_count` (pre-visibility-filter), flagged `match_count_approx: true` for the client to render "~N". Exact counting is intentionally not done (unbounded cost on common terms).
- **preview exact**: preview hits are all visible (post-`filterVisible`), budgeted per conversation by activity/frequency.
- DMs and threads are first-class buckets alongside groups.

## Scope

- New endpoint `POST /v1/messages/_search_global_groups` (L1). L2 drill-down reuses the existing `_search_global_messages` with `filters.channel_ids`.
- Trigger gate: only fires when `keyword ∨ sender_ids ∨ member_uids ∨ channel_ids` is present (stricter than the flat endpoint; `sent_at` / `content_types` / `channel_types` alone do not trigger).
- All tunables (`maxGroups` / `perGroupMax` / `presenceProbe` / `previewBudget` / `K2`) are config-driven.

BrDing-Rookie:
## What

Adds the **aggregation-first (L1) global search** endpoint `POST /v1/messages/_search_global_groups` on top of the existing `messages_search` module, plus the presence-calibration layer that makes the per-conversation overview visibility-exact.

Two logically separable commits (kept together for an atomic deploy — see below):
- **`608ca11` — L1 aggregation endpoint**: one OS `terms(channelId)` aggregation → candidate buckets with `max(timestamp)` (latest), `top_hits` (preview over-fetch), `doc_count` (approx count), `cardinality` (total groups). DM buckets reverse-resolved from the p2p fake channel id to the peer uid (`channel_type=1`, no thread); thread buckets carry `parent_group_no`. `revoked` + `visibles` gated at the OS query layer. Response shell `{data, pagination}` consistent with the existing search endpoints. Config-driven `maxGroups` / over-fetch `T`.
- **`277d798` — presence calibration + deep-probe + per-frequency preview**: batch `filterVisible` over each bucket's over-fetched top-T in a single MySQL round-trip → three-state classify (INCLUDE / EXCLUDE / UNRESOLVED). UNRESOLVED buckets deep-probe via `search_after` back-paging until the first visible hit (INCLUDE), channel exhausted (EXCLUDE), or the `K2` budget is hit (EXCLUDE, bounded latency). Per-conversation preview count `N` allocated from `previewBudget` weighted by `doc_count`, clamped to `perGroupMax`. Instrumentation for probe bucket/doc counts.

## Why combined into one PR

A and B are strongly coupled: the L1 endpoint's preview is only **visibility-safe once B's `filterVisible` calibration is in place**. Shipping A alone would expose previews that bypass the MySQL-only visibility gates (admin/mutual delete, self-delete, channel-offset watermark) — an information leak. They are therefore submitted as one PR for a single atomic deploy.

## Visibility guarantee

Preview hits are emitted **only** after passing the full five-gate `filterVisible`; the raw `top_hits` path is never sent to the wire. `filterVisible` DB errors are fail-closed (`INTERNAL_ERROR`, no uncalibrated hit released). A regression test asserts an admin-deleted hit mixed into the top-T never reaches the preview.

## Review response (CHANGES_REQUESTED)

Two follow-up commits address the review:

- **`50e7c9e` — `latest_at` + bucket order recomputed from the calibrated visible hit.** This was the Major/Blocking finding: `latest_at` and the latest-first bucket order used the OS **pre-filter** `max(timestamp)`, so a bucket whose newest match was hidden (admin/self-deleted, cleared history, visibles) leaked that hidden message's recency and biased the order — contradicting the per-conversation "time doesn't leak" claim. Now every bucket's `latest_at` is the timestamp of its newest **visible** hit (INCLUDE → `visible[0]`; UNRESOLVED → the visible hit the deep probe actually found; never the OS max), and the returned `groups[]` are re-sorted by that visible `latest_at`. The OS terms agg still orders *candidates* by pre-filter max (visible ⊆ all, so nothing that should surface is dropped), but the wire order follows the visible time. Regression tests cover the top-T and deep-probe paths and assert a hidden-newest bucket does not jump the order.
- **`5d31a9a` — presence batch `filterVisible` now chunks its id / channel sets** (`filterVisibleChunkSize = 1000`). The underlying `message.IService` queries build a single unchunked `... IN (?)`; the presence batch can present up to `maxGroups × T` (~8000) ids at once, so the width is now bounded here. Single-channel callers stay one chunk (page-size-bounded), unchanged.

## Notes (accepted approximations — by design, not bugs)

These are the two remaining pre-filter approximations. Both are **accepted, non-sensitive metadata** per the aggregation-first design §6 ("count 近似，精确计数成本无界不做") and the owner's decision (option A) — distinct from `latest_at` / bucket order above, which were real recency leaks and have been fixed to recompute from the visible hit.

- **`match_count` is the pre-visibility OS `doc_count`** (`match_count_approx = true`). It counts matching docs **before** the five visibility gates, so it *includes* hits hidden from the caller and may read **higher** than the number of messages the caller can actually open. It is deliberately **not** post-calibrated: recomputing could only count the visible docs inside the bounded sampling window (itself an approximation) at unbounded cost, and it exposes at most a possibly-inflated integer — never a hidden message's content, sender, or timestamp. `match_count_approx = true` is the wire signal of exactly this.
- **`total_groups` is a `cardinality` HLL estimate over the pre-visibility candidate set** (`total_groups_approx = true`); it can exceed the returned `groups` length when candidate buckets are EXCLUDE'd. Clients should treat `groups` as authoritative for the empty state.

## Testing

- `go build ./...`, `go vet ./modules/messages_search/`, `go test ./modules/messages_search/` all green.
- New tests cover: trigger gate three states, OS-layer `visibles`/`revoked` gating, DM reverse-resolution, thread `parent_group_no`, response shell shape, `maxGroups` over-limit `has_more`, presence three-state classify, calibrate INCLUDE/EXCLUDE/fail-closed + preview filtering, deep-probe first-visible/exhausted/pages-advance/K2, per-frequency N allocation, and the review-response additions: `latest_at` recompute (top-T + deep-probe), visible-latest bucket ordering + stable tiebreak, and `filterVisible` id-set chunking.
- End-to-end L1 latency instrumentation is in place; live numbers require an OS+MySQL test environment.

Fixes #592

### DIFF
--- a/modules/messages_search/search_global_groups.go
+++ b/modules/messages_search/search_global_groups.go
@@ -1,0 +1,482 @@
+package messages_search
+
+import (
+	"context"
+	"encoding/json"
+	"strings"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/modules/thread"
+	"github.com/google/uuid"
+	"github.com/olivere/elastic"
+	"go.uber.org/zap"
+)
+
+// SearchGlobalGroupsReq is the request body for
+// POST /v1/messages/_search_global_groups — the L1 group-aggregation
+// (聚合优先) overview (aggregation-first design §2). Unlike
+// _search_global_messages it has NO sort / page_size / cursor: L1 returns one
+// aggregated overview per request and the bucket order is fixed to latest_at
+// desc. `sequence` is echoed back verbatim for the frontend's stale-response
+// guard (§4) — the backend does not validate its monotonicity.
+type SearchGlobalGroupsReq struct {
+	Keyword  string              `json:"keyword,omitempty"`
+	Sequence int64               `json:"sequence,omitempty"`
+	Filters  GlobalSearchFilters `json:"filters,omitempty"`
+}
+
+// GroupsResult is the `data` object of the L1 response. The outer envelope is
+// the shared {data, pagination} shell (pagination.has_more = 命中群数 >
+// maxGroups; next_cursor is always ""). total_groups is a cardinality HLL
+// estimate so total_groups_approx is always true.
+type GroupsResult struct {
+	Sequence          int64         `json:"sequence"`
+	QueryID           string        `json:"query_id"`
+	TotalGroups       int64         `json:"total_groups"`
+	TotalGroupsApprox bool          `json:"total_groups_approx"`
+	Groups            []GroupBucket `json:"groups"`
+}
+
+// GroupBucket is one aggregated channel/thread/DM bucket. DM buckets carry the
+// reversed peer uid in ChannelID (channel_type=1) with no parent_group_no /
+// thread fields; group buckets set parent_group_no to their own channel_id;
+// thread buckets (channel_type=5) carry parent_group_no + thread_id +
+// thread_name. match_count is the OS doc_count (pre-visibility) so
+// match_count_approx is always true.
+type GroupBucket struct {
+	ChannelID        string       `json:"channel_id"`
+	ChannelType      uint8        `json:"channel_type"`
+	ParentGroupNo    string       `json:"parent_group_no,omitempty"`
+	GroupName        string       `json:"group_name"`
+	ThreadID         string       `json:"thread_id,omitempty"`
+	ThreadName       string       `json:"thread_name,omitempty"`
+	MatchCount       int64        `json:"match_count"`
+	MatchCountApprox bool         `json:"match_count_approx"`
+	LatestAt         string       `json:"latest_at"`
+	Preview          []MessageHit `json:"preview"`
+}
+
+func init() {
+	registerRoute(func(h *Handler, g *wkhttp.RouterGroup) {
+		g.POST("/_search_global_groups", h.searchGlobalGroups)
+	})
+}
+
+// searchGlobalGroups is POST /v1/messages/_search_global_groups. It runs a
+// single OS terms(channelId) aggregation over the caller's allowlist scope
+// (reusing the _search_global_messages DSL) plus an OS-layer visibles
+// whitelist, then projects each bucket into the L1 overview shape. presence
+// calibration + per-frequency preview allocation are backend B — this handler
+// returns the raw top_hits preview.
+func (h *Handler) searchGlobalGroups(c *wkhttp.Context) {
+	var req SearchGlobalGroupsReq
+	if err := c.BindJSON(&req); err != nil {
+		respondValidation(c, "body", "invalid JSON")
+		return
+	}
+	req.Keyword = strings.TrimSpace(req.Keyword)
+	loginUID := c.GetLoginUID()
+
+	if !validateKeywordOptional(c, req.Keyword) {
+		return
+	}
+	// §2.3 trigger gate — STRICTER than _search_global_messages'
+	// validateSearchNotEmpty: sent_at_* / content_types / channel_types alone
+	// do NOT trigger. Only keyword ∨ sender_ids ∨ member_uids/member_uid ∨
+	// channel_ids may open an aggregation, keeping the terms scope bounded.
+	if !groupsTriggerSatisfied(req.Keyword, req.Filters) {
+		respondValidation(c, "keyword", "at least one of keyword, sender_ids, member_uids, or channel_ids is required")
+		return
+	}
+	if _, ok := validateGlobalBase(c, h.cfg, "", "", req.Filters, 0, false); !ok {
+		return
+	}
+
+	osChannelIDs, spaceID, _, allowTimings, ok := h.resolveGlobalScope(c, loginUID, req.Filters.ChannelIDs, req.Filters.MemberUIDs, req.Filters.MemberUID)
+	if !ok {
+		return
+	}
+	recordAllowlistTimings(c, allowTimings)
+	if len(osChannelIDs) == 0 {
+		recordAudit(c, "search_global_groups", 0, "", req.Keyword, 0)
+		c.Response(groupsEnvelope(emptyGroupsResult(req.Sequence), false))
+		return
+	}
+
+	client, err := ESClient(h.cfg)
+	if err != nil {
+		h.Error("ESClient init failed", zap.Error(err))
+		respondUpstream(c)
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(c.Request.Context(), h.cfg.Timeout)
+	defer cancel()
+
+	msgReq := SearchGlobalMessagesReq{Keyword: req.Keyword, Filters: req.Filters}
+	base, analyzeErr := buildGlobalMessagesDSL(ctx, newOSIKSmartAnalyzer(client), h.cfg.StopwordStripEnabled, msgReq, osChannelIDs, spaceID)
+	if analyzeErr != nil {
+		h.Warn("messages_search: _analyze fallback (degraded keyword clause)", zap.Error(analyzeErr))
+	}
+	query := elastic.NewBoolQuery().Filter(base)
+	applyVisiblesWhitelist(query, loginUID)
+
+	res, qerr := h.runGroupsAggregation(ctx, client, query, req.Keyword != "")
+	if qerr != nil {
+		if responder := classifyOSError(qerr); responder != nil {
+			h.Warn("OS search_global_groups failed", zap.Error(qerr))
+			responder(c)
+			return
+		}
+		h.Error("messages_search: groups aggregation failed", zap.Error(qerr))
+		respondInternal(c)
+		return
+	}
+
+	result, hasMore := h.buildGroupsResult(ctx, res, req.Sequence, loginUID)
+	recordAudit(c, "search_global_groups", 0, "", req.Keyword, len(result.Groups))
+	c.Response(groupsEnvelope(result, hasMore))
+}
+
+// groupsTriggerSatisfied implements the §2.3 trigger gate. It deliberately
+// does NOT count sent_at_* / content_types / channel_types — those alone must
+// return VALIDATION_ERROR, which is why hasEffectiveFilters (which treats
+// sent_at as effective) is not reused here.
+func groupsTriggerSatisfied(keyword string, f GlobalSearchFilters) bool {
+	if strings.TrimSpace(keyword) != "" {
+		return true
+	}
+	for _, s := range f.SenderIDs {
+		if strings.TrimSpace(s) != "" {
+			return true
+		}
+	}
+	for _, u := range f.MemberUIDs {
+		if strings.TrimSpace(u) != "" {
+			return true
+		}
+	}
+	if strings.TrimSpace(f.MemberUID) != "" {
+		return true
+	}
+	for _, ref := range f.ChannelIDs {
+		if strings.TrimSpace(ref.ChannelID) != "" {
+			return true
+		}
+	}
+	return false
+}
+
+// applyVisiblesWhitelist adds the OS-layer visibles gate: a doc is visible when
+// its `visibles` array is empty/absent (no per-message allowlist) OR it lists
+// the caller. An empty array indexes no values, so exists(visibles) is false
+// for both absent and empty docs — mustNot(exists) covers the "no gate" case.
+// Together with the revoked mustNot already emitted by buildGlobalMessagesDSL
+// this pushes the two OS-resident visibility signals into the query layer so
+// backend B's presence calibration only has to chase the MySQL-only signals.
+func applyVisiblesWhitelist(b *elastic.BoolQuery, loginUID string) {
+	noGate := elastic.NewBoolQuery().MustNot(elastic.NewExistsQuery("visibles"))
+	listed := elastic.NewTermQuery("visibles", loginUID)
+	b.Filter(elastic.NewBoolQuery().Should(noGate, listed).MinimumShouldMatch("1"))
+}
+
+// runGroupsAggregation issues the single size:0 terms(channelId) aggregation.
+// Each bucket carries max(timestamp) (latest_at), a top_hits over-fetch of
+// T=perGroupMax+presenceProbe (preview + backend-B calibration headroom); a
+// sibling cardinality(channelId) yields the approximate total_groups.
+func (h *Handler) runGroupsAggregation(ctx context.Context, client *elastic.Client, query elastic.Query, withHighlight bool) (*elastic.SearchResult, error) {
+	gc := h.cfg.Groups
+	preview := elastic.NewTopHitsAggregation().
+		Size(gc.TopHitsSize()).
+		SortBy(searchSorters("time_desc")...).
+		FetchSourceContext(fileContentSourceExcludes())
+	if withHighlight {
+		preview = preview.Highlight(buildSearchAllHighlight())
+	}
+	byChannel := elastic.NewTermsAggregation().
+		Field("channelId").
+		Size(gc.MaxGroups).
+		OrderByAggregation("latest", false).
+		SubAggregation("latest", elastic.NewMaxAggregation().Field("timestamp")).
+		SubAggregation("preview", preview)
+	return client.Search().
+		Index(h.cfg.OSReadAlias).
+		Query(query).
+		Size(0).
+		TrackTotalHits(false).
+		Aggregation("by_channel", byChannel).
+		Aggregation("group_count", elastic.NewCardinalityAggregation().Field("channelId")).
+		Do(ctx)
+}
+
+// parsedBucket is the pre-projection view of one terms bucket.
+type parsedBucket struct {
+	key      string
+	docCount int64
+	latestTS int64
+	hits     []*elastic.SearchHit // already capped to perGroupMax
+}
+
+// buildGroupsResult projects the aggregation response into the L1 data object
+// and reports has_more (命中群数 > maxGroups, signalled by the terms
+// sum_other_doc_count on the single-shard index).
+func (h *Handler) buildGroupsResult(ctx context.Context, res *elastic.SearchResult, seq int64, loginUID string) (GroupsResult, bool) {
+	result := emptyGroupsResult(seq)
+	parsed, totalGroups, hasMore, ok := parseGroupsAggregation(res, h.cfg.Groups.PerGroupMax)
+	result.TotalGroups = totalGroups
+	if !ok {
+		return result, false
+	}
+
+	var groupNos, shortIDs, peerUIDs []string
+	for _, pb := range parsed {
+		ct, wireID, parentGroupNo, shortID := classifyGroupBucket(pb.key, loginUID)
+		switch ct {
+		case channelTypePerson:
+			peerUIDs = append(peerUIDs, wireID)
+		case channelTypeGroup:
+			groupNos = append(groupNos, pb.key)
+		case channelTypeThread:
+			groupNos = append(groupNos, parentGroupNo)
+			shortIDs = append(shortIDs, shortID)
+		}
+	}
+
+	groupNames := h.resolveGroupNames(groupNos)
+	threadNames := h.resolveThreadNames(shortIDs)
+	peerNames := h.resolvePeerNames(peerUIDs)
+	previews := h.projectPreview(ctx, parsed, loginUID)
+
+	for i, pb := range parsed {
+		result.Groups = append(result.Groups, assembleGroupBucket(pb, previews[i], loginUID, groupNames, threadNames, peerNames))
+	}
+	return result, hasMore
+}
+
+// parseGroupsAggregation is the pure (I/O-free) reader of the OS aggregation
+// response: it extracts total_groups (cardinality), has_more (terms
+// sum_other_doc_count > 0 on the single-shard index) and each bucket's key /
+// doc_count / latest_at / preview hits (capped to perGroupMax). Kept separate
+// from name resolution so the has_more + bucket parsing can be unit-tested
+// without standing up group/user/thread services. ok=false means the response
+// carried no by_channel terms aggregation.
+func parseGroupsAggregation(res *elastic.SearchResult, perGroupMax int) (buckets []parsedBucket, totalGroups int64, hasMore, ok bool) {
+	if res == nil || res.Aggregations == nil {
+		return nil, 0, false, false
+	}
+	if card, cok := res.Aggregations.Cardinality("group_count"); cok && card.Value != nil {
+		totalGroups = int64(*card.Value)
+	}
+	terms, tok := res.Aggregations.Terms("by_channel")
+	if !tok || terms == nil {
+		return nil, totalGroups, false, false
+	}
+	hasMore = terms.SumOfOtherDocCount > 0
+	for _, b := range terms.Buckets {
+		key := bucketKeyString(b)
+		if key == "" {
+			continue
+		}
+		latestTS := int64(0)
+		if m, mok := b.Max("latest"); mok && m.Value != nil {
+			latestTS = int64(*m.Value)
+		}
+		var hits []*elastic.SearchHit
+		if th, hok := b.TopHits("preview"); hok && th != nil && th.Hits != nil {
+			hits = th.Hits.Hits
+			if perGroupMax > 0 && len(hits) > perGroupMax {
+				hits = hits[:perGroupMax]
+			}
+		}
+		buckets = append(buckets, parsedBucket{key: key, docCount: b.DocCount, latestTS: latestTS, hits: hits})
+	}
+	return buckets, totalGroups, hasMore, true
+}
+
+// projectPreview turns each bucket's raw top_hits into []MessageHit and fills
+// sender names/avatars in a single batched senderJoin across all buckets (the
+// global endpoints deliberately skip per-group remark scoping — channelType=0
+// degrades senderJoin to a plain user lookup).
+func (h *Handler) projectPreview(ctx context.Context, buckets []parsedBucket, loginUID string) [][]MessageHit {
+	previews := make([][]MessageHit, len(buckets))
+	var senderIDs []string
+	for bi, pb := range buckets {
+		hits := make([]MessageHit, 0, len(pb.hits))
+		for _, hit := range pb.hits {
+			var doc Doc
+			if err := json.Unmarshal(rawSource(hit.Source), &doc); err != nil {
+				h.Warn("messages_search: bad groups preview _source skipped", zap.Error(err))
+				continue
+			}
+			wireID, wireType := wireChannelFromDoc(doc, loginUID)
+			mh := h.singleMessageHit(doc, wireID, wireType, map[string][]string(hit.Highlight))
+			hits = append(hits, mh)
+			senderIDs = append(senderIDs, doc.From)
+			for _, im := range mh.InnerMessages {
+				if im.SenderID != "" {
+					senderIDs = append(senderIDs, im.SenderID)
+				}
+			}
+		}
+		previews[bi] = hits
+	}
+	join := h.senderJoin(ctx, uniqUIDs(senderIDs), 0, "")
+	for bi := range previews {
+		for i := range previews[bi] {
+			mh := &previews[bi][i]
+			mh.SenderName = join.Names[mh.SenderID]
+			mh.SenderAvatarURL = join.Avatars[mh.SenderID]
+			for j := range mh.InnerMessages {
+				if uid := mh.InnerMessages[j].SenderID; uid != "" {
+					mh.InnerMessages[j].SenderName = join.Names[uid]
+				}
+			}
+		}
+	}
+	return previews
+}
+
+// classifyGroupBucket maps a terms(channelId) key to its wire shape. DM buckets
+// carry a p2p fake channelId (`uidA@uidB`) reversed to the peer uid; thread
+// buckets carry the composite `{group_no}____{short_id}`; everything else is a
+// group whose parent_group_no is itself.
+func classifyGroupBucket(key, loginUID string) (channelType uint8, wireID, parentGroupNo, threadShortID string) {
+	if common.IsFakeChannel(key) {
+		return channelTypePerson, common.GetToChannelIDWithFakeChannelID(key, loginUID), "", ""
+	}
+	if groupNo, shortID, err := thread.ParseChannelID(key); err == nil {
+		return channelTypeThread, key, groupNo, shortID
+	}
+	return channelTypeGroup, key, key, ""
+}
+
+// assembleGroupBucket builds the wire bucket from the parsed bucket + resolved
+// name maps. Pure (no I/O) so the DM reversal / thread parent_group_no / group
+// shapes are unit-testable without OS or MySQL.
+func assembleGroupBucket(pb parsedBucket, preview []MessageHit, loginUID string, groupNames, threadNames, peerNames map[string]string) GroupBucket {
+	if preview == nil {
+		preview = []MessageHit{}
+	}
+	ct, wireID, parentGroupNo, threadShortID := classifyGroupBucket(pb.key, loginUID)
+	gb := GroupBucket{
+		ChannelID:        wireID,
+		ChannelType:      ct,
+		MatchCount:       pb.docCount,
+		MatchCountApprox: true,
+		LatestAt:         msToRFC3339(pb.latestTS),
+		Preview:          preview,
+	}
+	switch ct {
+	case channelTypePerson:
+		gb.GroupName = peerNames[wireID]
+	case channelTypeGroup:
+		gb.ParentGroupNo = pb.key
+		gb.GroupName = groupNames[pb.key]
+	case channelTypeThread:
+		gb.ParentGroupNo = parentGroupNo
+		gb.ThreadID = pb.key
+		gb.ThreadName = threadNames[threadShortID]
+		gb.GroupName = groupNames[parentGroupNo]
+	}
+	return gb
+}
+
+// resolveGroupNames batch-resolves group_no → group name. Soft-fail: on error
+// the names are dropped (empty group_name) but the request still serves.
+func (h *Handler) resolveGroupNames(groupNos []string) map[string]string {
+	out := map[string]string{}
+	groupNos = uniqUIDs(groupNos)
+	if len(groupNos) == 0 {
+		return out
+	}
+	infos, err := h.groupService.GetGroups(groupNos)
+	if err != nil {
+		h.Warn("messages_search: groups name lookup failed", zap.Error(err))
+		return out
+	}
+	for _, g := range infos {
+		if g != nil {
+			out[g.GroupNo] = g.Name
+		}
+	}
+	return out
+}
+
+// resolvePeerNames batch-resolves DM peer uid → display name via the user
+// profile (§5b: DM header名字走对端用户档案 fallback). Soft-fail to placeholder.
+func (h *Handler) resolvePeerNames(peerUIDs []string) map[string]string {
+	out := map[string]string{}
+	peerUIDs = uniqUIDs(peerUIDs)
+	if len(peerUIDs) == 0 {
+		return out
+	}
+	users, err := h.userService.GetUsers(peerUIDs)
+	if err != nil {
+		h.Warn("messages_search: DM peer name lookup failed", zap.Error(err))
+		return out
+	}
+	for _, u := range users {
+		if u != nil {
+			out[u.UID] = u.Name
+		}
+	}
+	return out
+}
+
+// resolveThreadNames batch-resolves thread short_id → thread name. Soft-fail:
+// on error thread_name is left empty and the bucket still renders under its
+// parent group.
+func (h *Handler) resolveThreadNames(shortIDs []string) map[string]string {
+	out := map[string]string{}
+	shortIDs = uniqUIDs(shortIDs)
+	if len(shortIDs) == 0 {
+		return out
+	}
+	models, err := thread.NewDB(h.ctx).QueryByShortIDs(shortIDs)
+	if err != nil {
+		h.Warn("messages_search: thread name lookup failed", zap.Error(err))
+		return out
+	}
+	for sid, m := range models {
+		if m != nil {
+			out[sid] = m.Name
+		}
+	}
+	return out
+}
+
+// bucketKeyString reads the terms bucket key as a string (channelId is a
+// keyword field). Falls back to key_as_string for defensiveness.
+func bucketKeyString(b *elastic.AggregationBucketKeyItem) string {
+	if b == nil {
+		return ""
+	}
+	if s, ok := b.Key.(string); ok {
+		return s
+	}
+	if b.KeyAsString != nil {
+		return *b.KeyAsString
+	}
+	return ""
+}
+
+// emptyGroupsResult is the zero-hit / empty-scope data object (§5.5).
+func emptyGroupsResult(seq int64) GroupsResult {
+	return GroupsResult{
+		Sequence:          seq,
+		QueryID:           uuid.NewString(),
+		TotalGroups:       0,
+		TotalGroupsApprox: true,
+		Groups:            []GroupBucket{},
+	}
+}
+
+// groupsEnvelope wraps the L1 data object into the shared {data, pagination}
+// shell. next_cursor is always "" (L1 has no per-row pagination).
+func groupsEnvelope(result GroupsResult, hasMore bool) CursorList {
+	return CursorList{
+		Data:       result,
+		Pagination: Pagination{HasMore: hasMore, NextCursor: ""},
+	}
+}
--- a/modules/messages_search/search_global_groups.go
+++ b/modules/messages_search/search_global_groups.go
@@ -134,7 +134,21 @@ func (h *Handler) searchGlobalGroups(c *wkhttp.Context) {
 		return
 	}
 
-	result, hasMore := h.buildGroupsResult(ctx, res, req.Sequence, loginUID)
+	result, hasMore, berr := h.buildGroupsResult(ctx, c, res, req.Sequence, loginUID, client, base)
+	if berr != nil {
+		// Deep-probe OS round-trips can fail like any other search; a
+		// classifiable OS error maps to UPSTREAM_UNAVAILABLE, everything else
+		// (notably the fail-closed filterVisible DB gate) is INTERNAL_ERROR so
+		// we never fall through to leaking uncalibrated preview.
+		if responder := classifyOSError(berr); responder != nil {
+			h.Warn("OS presence deep-probe failed", zap.Error(berr))
+			responder(c)
+			return
+		}
+		h.Error("messages_search: presence calibration failed", zap.Error(berr))
+		respondInternal(c)
+		return
+	}
 	recordAudit(c, "search_global_groups", 0, "", req.Keyword, len(result.Groups))
 	c.Response(groupsEnvelope(result, hasMore))
 }
@@ -218,25 +232,56 @@ type parsedBucket struct {
 	hits     []*elastic.SearchHit // already capped to perGroupMax
 }
 
-// buildGroupsResult projects the aggregation response into the L1 data object
-// and reports has_more (命中群数 > maxGroups, signalled by the terms
-// sum_other_doc_count on the single-shard index).
-func (h *Handler) buildGroupsResult(ctx context.Context, res *elastic.SearchResult, seq int64, loginUID string) (GroupsResult, bool) {
+// buildGroupsResult projects the aggregation response into the L1 data object.
+// Backend B: between parse and assemble it runs presence calibration (§2.2) so
+// every returned bucket "确有可见命中" and every preview row is a filterVisible
+// -passed visible hit — the raw top_hits are NEVER projected to the wire (that
+// was the A-review Blocker: uncalibrated preview leaked snippet/sender/time of
+// admin-deleted / self-deleted / cleared-history messages). has_more still
+// reflects 命中群数 > maxGroups (terms sum_other_doc_count on the single-shard
+// index). Returns an error when calibration hits a fail-closed DB gate or a
+// deep-probe OS round-trip fails, so the caller can respond upstream/internal
+// instead of leaking.
+func (h *Handler) buildGroupsResult(ctx context.Context, c *wkhttp.Context, res *elastic.SearchResult, seq int64, loginUID string, client *elastic.Client, base elastic.Query) (GroupsResult, bool, error) {
 	result := emptyGroupsResult(seq)
-	parsed, totalGroups, hasMore, ok := parseGroupsAggregation(res, h.cfg.Groups.PerGroupMax)
+	// Keep the FULL over-fetched top-T sample (T=perGroupMax+presenceProbe) for
+	// calibration — do not truncate to perGroupMax here; preview trimming to the
+	// per-frequency N happens after INCLUDE/EXCLUDE is decided.
+	parsed, totalGroups, hasMore, ok := parseGroupsAggregation(res, h.cfg.Groups.TopHitsSize())
 	result.TotalGroups = totalGroups
 	if !ok {
-		return result, false
+		return result, false, nil
+	}
+
+	var timings searchPhaseTimings
+	calibrated, stats, err := h.calibratePresence(ctx, client, base, parsed, loginUID, &timings)
+	recordSearchTimings(c, timings)
+	recordPresenceStats(c, stats)
+	if err != nil {
+		return result, false, err
+	}
+
+	// Per-frequency N allocation (§4): spread previewBudget across the surviving
+	// buckets weighted by match frequency, then take the most-recent N visible
+	// rows per bucket.
+	ns := allocatePreviewN(calibrated, h.cfg.Groups.PreviewBudget, h.cfg.Groups.PerGroupMax)
+	visibleHits := make([][]*elastic.SearchHit, len(calibrated))
+	for i, cb := range calibrated {
+		k := ns[i]
+		if k > len(cb.visible) {
+			k = len(cb.visible)
+		}
+		visibleHits[i] = cb.visible[:k]
 	}
 
 	var groupNos, shortIDs, peerUIDs []string
-	for _, pb := range parsed {
-		ct, wireID, parentGroupNo, shortID := classifyGroupBucket(pb.key, loginUID)
+	for _, cb := range calibrated {
+		ct, wireID, parentGroupNo, shortID := classifyGroupBucket(cb.pb.key, loginUID)
 		switch ct {
 		case channelTypePerson:
 			peerUIDs = append(peerUIDs, wireID)
 		case channelTypeGroup:
-			groupNos = append(groupNos, pb.key)
+			groupNos = append(groupNos, cb.pb.key)
 		case channelTypeThread:
 			groupNos = append(groupNos, parentGroupNo)
 			shortIDs = append(shortIDs, shortID)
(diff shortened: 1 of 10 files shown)


modules/messages_search/search_global_groups.go
```diff
@@ -0,0 +1,531 @@
...
+	if preview == nil {
+		preview = []MessageHit{}
+	}
+	ct, wireID, parentGroupNo, threadShortID := classifyGroupBucket(pb.key, loginUID)
+	gb := GroupBucket{
+		ChannelID:        wireID,
+		ChannelType:      ct,
+		MatchCount:       pb.docCount,
+		MatchCountApprox: true,
+		LatestAt:         msToRFC3339(pb.latestTS),
```
OctoBoooot: **[major]** `latest_at` is set from the OS `max(timestamp)` sub-agg (L215), computed over docs passing only the OS gates (revoked+visibles), NOT the four MySQL-only visibility gates. So an INCLUDE bucket whose newest matching message is admin/self-deleted or watermark-hidden still exposes that hidden message's timestamp here (and drives bucket ordering via OrderByAggregation(latest)). Preview *content* is correctly gated by filterVisible, but latest_at/ordering bypass it — a metadata leak of a hidden message's existence+timing. Fix: recompute from the newest visible hit cb.visible[0] (most-recent-first) instead of pb.latestTS; add a regression test.

OctoBoooot (changes requested):
## Review: aggregation-first global search — L1 groups endpoint + presence calibration (#593)

**Verdict: Request changes** — one **major**: `latest_at` (and bucket ordering) leaks the timestamp/existence of a message that fails the MySQL visibility gates, bypassing the very calibration this PR adds for previews. Everything else is byte-verified sound — the authz architecture is careful and the leak-avoidance for *preview content* is done right; this is a metadata channel the preview gate doesn't cover.

Risk: `messages_search` read-authz surface, XL. I verified authz parity + the leak boundaries directly (and adversarially verified the finding below in both directions).

### Major — `latest_at` / bucket ordering leak past `filterVisible`
`search_global_groups.go`:
- L215 builds `SubAggregation("latest", NewMaxAggregation().Field("timestamp"))` — an OS `max(timestamp)` computed over docs passing **only** the OS-layer gates (revoked + visibles), **not** the four MySQL-only gates (`message_extra.is_deleted`, `message_user_extra.message_is_deleted`, channel-offset watermark).
- L329-330 reads `latestTS` from that OS max; L416 sets `LatestAt: msToRFC3339(pb.latestTS)` **verbatim** — never recomputed from `cb.visible` (the filterVisible-passed hits). Bucket order is also `OrderByAggregation("latest", false)`, i.e. driven by the same pre-visibility max.

**Failing input (byte-verified):** caller in group gA; gA has exactly two matching messages — msg#100 at `2026-07-15T10:00:00Z` then admin-deleted (`is_deleted=1`), and msg#99 at `2026-06-01T09:00:00Z` still visible. OS top-T returns both; `filterVisible` drops #100, keeps #99 → bucket INCLUDE. But `latest_at` = `max(timestamp)` = #100's `2026-07-15T10:00:00Z` — the deleted message's send time. The caller learns a message existed in gA at that instant **after** it was deleted, and the bucket sorts by that hidden timestamp. Content/sender are correctly withheld (preview passes filterVisible), but the timestamp + ordering metadata escape the gate.

This partially undercuts the PR's own stated guarantee ("preview hits emitted only after filterVisible") — `latest_at`/ordering aren't preview hits, so that gate doesn't cover them. Graded **major** (metadata leak of hidden-message existence+timing, not content), not blocker.

**Fix:** recompute `latest_at` from the newest **visible** hit — `cb.visible[0]` (presence.go documents `visible` as most-recent-first) — rather than the OS pre-visibility `max`, and order buckets by that. For an INCLUDE bucket `cb.visible` is non-empty by definition. Add a regression test: a bucket whose newest matching message is admin-deleted must report the older visible message's `latest_at`, not the deleted one's.

### Minor (related, already disclosed)
- `match_count` = raw OS `doc_count` (pre-visibility), so it over-reports hidden matches (a bucket with 1 visible + 30 deleted reports 31). The PR discloses this via `match_count_approx=true` — accepted approximation, but it's the same count-cardinality side channel; worth deciding whether the approximate count should also be visibility-bounded given the `latest_at` fix.

### Praise (byte-verified sound)
- **Authz parity holds:** the endpoint reuses the *same* `resolveGlobalScope` (allowlist ∩ channel_ids ∩ member scope, group-blacklist via ExistMembersActive, bidirectional DM blacklist, bot-in-space gate, space-scoped filter, RequireSpaceID fail-close) and the *same* `buildGlobalMessagesDSL` + `applyVisiblesWhitelist` + five-gate `filterVisible` as `_search_global_messages` — no weaker parallel path. DM reverse-resolution only runs on allowlisted bucket keys; cross-space buckets can't leak.
- **Preview content is leak-safe + fail-closed:** raw `top_hits` never reach the wire; every preview row passes `filterVisible`; a filterVisible DB error → INTERNAL_ERROR with **no** hits released (verified, `TestCalibratePresence_FailClosed`).
- **Deep-probe is strictly bounded** (K2 budget enforced, search_after always advances or returns, one filterVisible per page) — no unbounded OS/MySQL load; the EXCLUDE-vs-UNRESOLVED boundary (sampleSize≥docCount with 0 visible → EXCLUDE) is correct and errs toward hiding, never toward leaking.

### Note on merge gate
Build/Vet/Lint green, Test running; `check-sprint` (governance) red is the maintainer's to clear. The `latest_at` fix is the one blocking item; the rest of the design is approvable.

Jerry-Xin (dismissed):
Summary: The PR is relevant to `octo-server`, but the new L1 response still exposes uncalibrated metadata from invisible hits.

🔴 Blocking

🔴 Critical — `latest_at` and `match_count` leak metadata from hits that failed MySQL visibility gates.

`parseGroupsAggregation` stores `latestTS` from the raw OS `max(timestamp)` aggregation and `docCount` from the raw bucket count before `filterVisible` runs ([modules/messages_search/search_global_groups.go](/Users/mlamp/.openclaw/tmp/pr-review-sjdrzcet/repo/modules/messages_search/search_global_groups.go:328), [modules/messages_search/search_global_groups.go](/Users/mlamp/.openclaw/tmp/pr-review-sjdrzcet/repo/modules/messages_search/search_global_groups.go:339)). After calibration, `assembleGroupBucket` still returns those raw values as `latest_at` and `match_count` ([modules/messages_search/search_global_groups.go](/Users/mlamp/.openclaw/tmp/pr-review-sjdrzcet/repo/modules/messages_search/search_global_groups.go:414), [modules/messages_search/search_global_groups.go](/Users/mlamp/.openclaw/tmp/pr-review-sjdrzcet/repo/modules/messages_search/search_global_groups.go:416)).

That means if a bucket has a recent admin-deleted/self-deleted/cleared-history match plus an older visible match, the preview is filtered correctly, but `latest_at` can still reveal the invisible message’s timestamp and bucket ordering can be driven by that invisible hit. `match_count` likewise includes hidden matches, exposing hidden match volume. This violates the PR’s stated visibility guarantee for per-conversation overview metadata.

Fix direction: derive `latest_at` from the latest calibrated visible hit, reorder returned buckets by that visible timestamp, and avoid returning raw pre-visibility `doc_count` as a visible `match_count` unless the field is explicitly documented and accepted as pre-visibility metadata.

✅ Highlights

The new endpoint is mounted through the existing shared auth, Space, UID rate-limit, audit, and backend-gate route chain.

The preview path itself is fail-closed and filters hits through `filterVisible` before projection.

Verification run: `go test ./modules/messages_search/` passed.

Jerry-Xin (dismissed):
Summary: Authorization/scoping, SQLi, cursor integrity, and preview-content calibration are all sound. But the new L1 response still exposes uncalibrated per-conversation overview metadata (`latest_at`, `match_count`, and bucket ordering) derived from hits that failed the MySQL visibility gates, which contradicts the PR's own visibility guarantee.

## 🔴 Blocking

**🔴 `latest_at` / `match_count` / bucket ordering leak metadata from invisible hits.**

`parseGroupsAggregation` stores `latestTS` from the raw OS `max(timestamp)` sub-aggregation and `docCount` from the raw terms bucket `doc_count`, both captured *before* `filterVisible` runs (`modules/messages_search/search_global_groups.go` — `parsedBucket{key, docCount, latestTS, ...}` in `parseGroupsAggregation`). After `calibratePresence`, `assembleGroupBucket` still emits those raw values verbatim: `MatchCount: pb.docCount` and `LatestAt: msToRFC3339(pb.latestTS)` (`modules/messages_search/search_global_groups.go`, `assembleGroupBucket`). Bucket order is fixed by `OrderByAggregation("latest", false)` in `runGroupsAggregation`, i.e. by the same raw `max(timestamp)`.

Impact: if a bucket's most-recent match is admin-deleted / self-deleted / cleared-history (i.e. dropped by the five-gate `filterVisible`) but an older visible match exists, the `preview` array is correctly filtered, yet:
- `latest_at` still reveals the *invisible* newest message's timestamp;
- bucket ordering (latest-first) can be driven by that invisible hit, so the group's rank leaks the existence/recency of a hidden message;
- `match_count` includes hidden matches, exposing hidden match volume.

The PR body states the L1 overview is "visibility-exact" and that invisible messages must not leak snippet/sender/**time**. `latest_at`/`match_count`/order are exactly that per-conversation overview metadata, and they are not calibrated. The existing `Notes` caveat only acknowledges `total_groups` (cardinality) as pre-visibility — it does not cover per-bucket `latest_at`/`match_count`.

Fix direction: derive `latest_at` from the latest *calibrated visible* hit, order returned buckets by that visible timestamp, and either recompute `match_count` post-calibration or explicitly document/rename it as approximate pre-visibility metadata that is accepted as non-sensitive.

## 💬 Non-blocking

- `presence calibration issues a batch filterVisible over up to maxGroups × T ids` (per the PR's own note): confirm the underlying visibility probes (`RevokedSet`/`GloballyDeletedSet`/`UserDeletedSet`/`ChannelOffsets`) chunk large IN sets, or batch them, to avoid a single oversized query under high `maxGroups`.

## ✅ Highlights / verified clean

- **Authorization / data scoping (in-query):** The aggregation is scoped to `osChannelIDs` via a hard `terms(channelId)` filter in `buildGlobalMessagesDSL` (`b.Filter(elastic.NewTermsQuery("channelId", channelTerms...))`). `osChannelIDs` comes from `resolveGlobalScope` → `buildAllowlist(loginUID, spaceID)`, which enumerates only the caller's joined groups (`GetGroupsWithMemberUID(loginUID)`), gated by `ExistMembersActive` (fail-closed on error), the caller's DM peers, and threads under those joined groups (threads inherit parent-group membership). Client-supplied `filters.channel_ids` / `member_uids` are **intersected** with the allowlist (`allowlist ∩ requested`); unauthorized ids are silently dropped, never trusted. No cross-group / IDOR leak in the result set.
- **SQL/DSL injection:** All OS queries use the `elastic` builder with parameterized terms varargs (no string-concatenated user input); keyword is analyzed via `_analyze`, not concatenated. MySQL visibility probes take `[]string` id slices through the repository layer, not raw SQL.
- **Cursor / pagination integrity:** L1 has no client cursor (`next_cursor` always `""`, `sequence` echoed verbatim and never used in a query). The deep-probe `search_after` boundary is built from prior OS hits' sort values (`buildSearchAfterFromHit`), not client input, and every probe page re-applies `applyVisiblesWhitelist` + `filterVisible`. Scan is bounded by the `K2` budget (DoS-bounded).
- **"Presence calibration" is message-visibility, not user online-status** — no user presence/last-seen leak; `presence_probe_*` are internal audit counters only.
- **Preview content path is fail-closed:** raw `top_hits` never reach the wire; every preview row passes the five-gate `filterVisible`; DB errors propagate to `INTERNAL_ERROR`. Negative tests exist (`TestCalibratePresence_IncludeFiltersPreview`, `_ExcludeUniqueFiltered`, `_FailClosed`, deep-probe K2/exhausted).
- New endpoint is mounted through the existing shared auth / Space / UID rate-limit / audit / backend-gate route chain (`route_chain_test.go` asserts endpoint coverage).
- Verification run: `go test ./modules/messages_search/` passed.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #593 (octo-server)

## Summary

This PR introduces two changes: (1) a new `POST /v1/messages/_search_global_groups` endpoint implementing an aggregation-first L1 global search with presence calibration and per-frequency preview allocation, and (2) consolidation of the `summary` notification bot into the existing `notification` bot to avoid creating a second system DM conversation in user chat lists.

The new endpoint is well-architected — the three-state presence calibration (INCLUDE/EXCLUDE/UNRESOLVED), the K2-bounded deep probe, and the frequency-weighted preview allocation are all cleanly separated with comprehensive unit tests covering boundary conditions. The bot unification is a clean removal with safe locking consolidation.

## Verification

- ✅ **Route registration** — `_search_global_groups` uses `registerRoute` (`search_global_groups.go:61`), ensuring shared middleware chain (auth, rate-limit, audit, backendGate, Space). `route_chain_test.go` updated from 7 → 8 endpoints.
- ✅ **Fail-closed visibility** — `filterVisible` propagates DB errors; `calibratePresence` returns the error so the handler responds INTERNAL_ERROR rather than leaking uncalibrated hits (`presence.go:78-83`). Verified via `TestCalibratePresence_FailClosed`.
- ✅ **Deep-probe budget safety** — `deepProbeBucket` handles `K2 < len(pb.hits)` correctly: `budget <= 0` short-circuits to EXCLUDE (`presence.go:178-180`). The paging loop in `deepProbeVisible` caps each page to `budget - scanned` (`presence.go:210-213`).
- ✅ **Preview/calibrated length parity** — `projectPreview` returns `[][]MessageHit` with `len(previews) == len(calibrated)` (`search_global_groups.go:228,267`). No index-out-of-bounds possible in the assembly loop.
- ✅ **Bot readiness race safety** — `ensureNotifyBotReady` uses double-checked locking (`botMu` + `botOK atomic.Bool`) shared across all three callers: startup goroutine, `deliverNotification`, `deliverCardNotification`. No race condition.
- ✅ **Summary bot removal completeness** — All six old identifiers (`SummaryBotUIDValue`, `SummaryBotUID`, `summaryBotName`, `summaryBotMu`, `summaryBotOK`, `ensureSummaryBot`) fully removed. No remaining references in production code, migrations, or seed data.
- ✅ **Trigger gate correctness** — `groupsTriggerSatisfied` correctly rejects `sent_at`/`content_types`/`channel_types` alone, preventing empty aggregations. Test covers all trigger and non-trigger paths.
- ✅ **Response envelope shape** — `groupsEnvelope` produces `{data, pagination}` with `next_cursor: ""` and `has_more` driven by `sum_other_doc_count > 0`. Approx flags set correctly.

Static analysis only at head `277d79873ec466742ae48952c75d13d94ee9b9f0`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 documentation inconsistency; one nit on a pre-existing test fixture.

### P2 — `allocatePreviewN` doc overstates the budget ceiling (`modules/messages_search/presence.go:265`)

The doc comment states: "Budget is a soft ceiling — the perGroupMax clamp may leave the sum below budget, never above (every bucket ≤ perGroupMax and the weighted extra sums to ≤ budget-n)." This holds when `len(buckets) ≤ budget`, but the floor-of-1 guarantee means when `len(buckets) > budget` the sum of floor allocations is `n > budget`. With defaults (MaxGroups=200, PreviewBudget=500) this cannot happen. A deployment with `OCTO_SEARCH_GROUPS_MAX_GROUPS > OCTO_SEARCH_GROUPS_PREVIEW_BUDGET` would exceed the budget.

Tighten the doc to: "Budget is a soft ceiling; the floor-of-1 guarantee takes precedence, so the sum may exceed budget when the surviving bucket count exceeds it."

### Nit — Stale `"summary"` fixture in `internal/carddispatch/dispatch_test.go:131`

Pre-existing (not in this PR's diff): the test fixture uses hardcoded `SenderUID: "summary"` while production now sends from `"notification"`. The test is self-contained (asserts its own fixture round-trips), so no functional break, but the fixture string is cosmetically stale.

## Data Flow Trace

- **Preview data flow**: OS `top_hits` → `parseGroupsAggregation` (caps to `TopHitsSize()`) → `calibratePresence` (filterVisible gates each hit) → `allocatePreviewN` (weighted N per bucket) → trim `visible[:N]` → `projectPreview` (unmarshal + senderJoin) → `assembleGroupBucket` (wire shape). Every hit reaching the wire passes filterVisible.
- **Visibility signals**: OS-layer (`visibles` whitelist via `applyVisiblesWhitelist`, `revoked` mustNot via `buildGlobalMessagesDSL`) + MySQL-layer (five-gate `filterVisible`: revoked, globally-deleted, user-deleted, channel-offset, admin/mutual is_deleted). Fail-closed on any DB error.
- **Bot identity flow**: `deliverCardNotification` → `ensureNotifyBotReady()` → `ensureNotifyBot()` (provisions `"notification"` UID) → card sender uses `NotifyBotUIDValue` → text fallback uses `NotifyBotUIDValue`. Single identity, single DM conversation.

- **C1 (Dual-path parity)** — CLEAR. Summary bot removal verified across all call sites: ensureSummaryBotReady → ensureNotifyBotReady, SummaryBotUIDValue → NotifyBotUIDValue, summaryBotOK → botOK. No asymmetric path missed.
- **C2 (Control-flow ordering)** — CLEAR. `ensureNotifyBotReady` called from 3 sites with safe double-checked locking. No nested reuse hazard.
- **C3 (Authorization boundary)** — CLEAR. New endpoint uses `registerRoute` (shared auth middleware). `resolveGlobalScope` applies caller's allowlist. No authorization gap.
- **C4 (Auth lifecycle)** — N/A. No auth/permission model changes.
- **C5 (Build/packaging)** — N/A. No build, Docker, or packaging changes.
- **C6 (Governance/docs)** — N/A. `.octospec` docs are design documents, not governance policies.

## Cross-Round Blocker Recheck (R6)

N/A — first review round.

## Verdict: APPROVED

Well-implemented PR with thorough test coverage. The presence calibration design is sound — fail-closed visibility, bounded deep probe, and frequency-weighted preview allocation all work correctly. The bot unification is a clean, race-safe consolidation. Only finding is a P2 doc inconsistency that cannot trigger with default config.

[Octo-Q] verdict: APPROVE — no P1 blockers found. Thorough presence calibration design with fail-closed visibility, comprehensive boundary tests, and clean bot consolidation. Only a P2 doc nit on `allocatePreviewN` ceiling claim.

yujiawei (changes requested):
# Code Review — PR #593 (octo-server)

Reviewed at head SHA `277d79873ec466742ae48952c75d13d94ee9b9f0` against `origin/main` (merge-base `5e336a0c`). `go build`, `go vet`, and `go test ./modules/messages_search/` all pass locally. This review used a multi-lens pass (visibility, presence logic, cross-file integration, spec compliance) with adversarial verification of every P0/P1 candidate, plus an independent second model. Findings that survived verification are below.

## 1. Spec Compliance

**Spec: ✅** for the concrete deliverables.

- **Missing:** none. Both commits land: the L1 `terms(channelId)` aggregation endpoint (`_search_global_groups`), the stricter §2.3 trigger gate (keyword ∨ sender_ids ∨ member_uids/member_uid ∨ channel_ids — `sent_at`/`content_types`/`channel_types` correctly do *not* trigger, `groupsTriggerSatisfied` + `TestGroupsTriggerSatisfied`), OS-layer `visibles`+`revoked` gating, DM reverse-resolution, thread `parent_group_no`, the `{data, pagination}` shell, presence three-state calibration, deep-probe, and per-frequency N allocation.
- **Over-build:** none. `PreviewBudget`/`K2` are declared in commit A but consumed by commit B in this same PR, so they are not dead config.
- **Deviation:** the *explicit* preview guarantee ("preview hits emitted only after the five-gate `filterVisible`; raw `top_hits` never sent to the wire") is met and regression-tested (`TestCalibratePresence_IncludeFiltersPreview`). However, the endpoint's broader stated goal of a "visibility-exact per-conversation overview" is only partially met — see P1 below on `latest_at`. This is scored as a Code Quality finding rather than a spec ❌ because the promised preview-content gate itself is correct.

## 2. Code Quality

### P1 — `latest_at` (and bucket ordering) leak a pre-visibility timestamp with no approximation signal

`modules/messages_search/search_global_groups.go:416`
```go
LatestAt: msToRFC3339(pb.latestTS),   // pb.latestTS = OS max(timestamp), pre-visibility
```
`pb.latestTS` comes from the OS `max(timestamp)` sub-aggregation (`search_global_groups.go:215`, parsed at `:329`), computed **before** the MySQL-side gates (`is_deleted`, user-delete, channel-offset clear-history) are applied in `calibratePresence`. After calibration drops those hits, `latest_at` is never recomputed.

Consequence: for a bucket whose most-recent message is admin/mutual-deleted (or cleared past the caller's channel-offset) but which has an older *visible* hit, the response returns a `latest_at` equal to the **hidden** message's timestamp while `preview` shows only the older visible rows. The client renders an authoritative-looking "last activity at T" that points at a message the caller cannot see — a timestamp leak of exactly the "admin-deleted / self-deleted / cleared-history" class the calibration layer was built to close (see the `buildGroupsResult` docstring at `:236`). `latest_at` is also the bucket **sort key** (`OrderByAggregation("latest", false)`, `:214`), so a conversation can rank high in the overview purely on a hidden recent message.

This is inconsistent with the sibling metadata: `match_count` and `total_groups` are the same pre-visibility family but each carries an explicit `_approx` flag *and* a documented caveat, so the client is told not to trust them exactly. `latest_at` has neither — no `latest_at_approx`, no doc note — so it reads as exact data when it is not.

The fix is cheap and needs no extra round-trip: `cb.visible[0]` is already the newest visible hit (both the top-T sample and the deep-probe page preserve `time_desc` order), so `latest_at` can be sourced from `calibrated[i].visible[0]`'s timestamp, and the surviving buckets can be re-sorted by that value after calibration. Alternatively, if pre-visibility `latest_at` is an intentional product tradeoff, mirror the siblings: add a `latest_at_approx` flag and document it. Either is acceptable; shipping an unflagged phantom timestamp on a "visibility-exact overview" is not.

### P2 — First calibration pass can send up to ~8,000 message ids through unchunked MySQL `IN(...)`

`modules/messages_search/presence.go:64` collects `allRefs` across every bucket's top-T and hands them to `filterVisible` in one shot; the backing queries are unchunked `... message_id in ?` (`modules/message/db_message_extra.go:75/84`, `db_message_user_extra.go:53`). At defaults (`MaxGroups=200 × TopHitsSize=40`) that is up to 8,000 unique ids per lookup, ×3 lookups (revoked / global-deleted / user-deleted). Safe at defaults, but a real p99 / DB-load risk and fragile if the env knobs are raised. The PR already flags this as follow-up; chunking inside `filterVisible` (or the probe methods) would bound statement size with no semantic change. Non-blocking.

### P2 — `allocatePreviewN` can exceed `PreviewBudget` when included buckets outnumber the budget

`modules/messages_search/presence.go` floors every bucket to `ns[i]=1` before distributing `extra := budget - n`. When `PreviewBudget < len(calibrated)`, the floor alone yields `len(calibrated)` rows > budget, contradicting the "budget is a soft ceiling, never above" comment and the intent of `TestAllocatePreviewN_BudgetCeiling` (which only exercises `budget >= n`). Defaults are safe (`500 >= MaxGroups 200`), but `PreviewBudget` is env-driven and unclamped. Consider clamping, or documenting that the floor-of-1 wins over the budget. Non-blocking.

## 3. Verified — no finding

These candidates were checked and cleared, several by independent re-verification:

- **Preview content is fully calibrated.** The A-review blocker (raw `top_hits` reaching the wire) is closed; only `filterVisible`-passed hits are projected, and DB errors are fail-closed to `INTERNAL_ERROR` (`buildGroupsResult:236`, `calibratePresence`, `TestCalibratePresence_FailClosed`).
- **`classifyPresence` boundary is correct.** It uses `len(pb.hits)` (returned sample size), not requested `T`; `sampleSize >= docCount` with 0 visible → EXCLUDE is exact on the single-shard index (`TestClassifyPresence`). The post-reshard caveat is documented, and acceptable while `number_of_shards: 1`.
- **Deep-probe resume is sound.** `search_after` resumes strictly after the oldest top-T hit via the full-precision `[timestamp, messageId, subSeq]` tuple (`buildSearchAfterFromHit`, read from typed `_source` to avoid float64 snowflake rounding); no skip/overlap; K2 budget counts the top-T already scanned.
- **DM id space is correct.** `deepProbeBucket` queries `TermQuery("channelId", pb.key)` using the OS fake channel id (what the index stores); `projectDocRef` separately reverses DMs to peer uid only for the `channel_offset` key.
- **No group/name leak beyond membership.** Buckets are scoped to the caller's active-membership allowlist (`resolveGlobalScope` → `buildAllowlist`, `ExistMembersActive` fail-closed), and empty scope short-circuits before any OS query, so `group_name`/`thread_name`/peer-name resolution cannot surface a channel the caller isn't in.
- **Name lookups are batched** (`GetGroups`/`GetUsers`/`QueryByShortIDs`), soft-fail to empty, no N+1.

## 4. Suggested direction

Recompute `latest_at` from the calibrated `visible[0]` and re-sort surviving buckets by it (cheapest, keeps the overview honest), or add an explicit `latest_at_approx` flag + doc note to match `match_count`/`total_groups`. The two P2s can be follow-ups.

## 5. Coverage note

- **Verified:** the full diff (8 files), plus the backing paths in `modules/messages_search/{visibility.go, dsl.go, search_global.go, search_global_messages.go}` and `modules/message/db_message_extra.go` / `db_message_user_extra.go`. `go build` / `go vet` / `go test ./modules/messages_search/` green.
- **Not verified:** live L1 latency and the 8,000-id `IN(...)` behavior against a real OS+MySQL cluster (no test environment); the post-reshard `doc_count`-approximate EXCLUDE path (single-shard today, so not exercised). A cross-model advisory pass on the long-context/architecture angle timed out and did **not** return findings — it was absent, not a silent approval, so this verdict rests on the primary review + one independent model that concurred on the P1.

OctoBoooot (approved):
## Delta re-review: #593 @ 5d31a9a2 — my `latest_at` 🔴 is FIXED. REQUEST_CHANGES → APPROVE.

The author pushed the fix (delta ahead 2, clean; touches `search_global_groups.go`, `presence.go`, `visibility.go` + tests). Byte-verified the leak I raised is closed, correctly and completely — lifting my CR to **Approve**, conditional on Test.

### ✓ `latest_at` timestamp leak — FIXED
- `assembleGroupBucket` now takes `latestAtMS` and sets `LatestAt` from **`cb.latestVisibleTS`** (the calibrated visible hit), no longer `pb.latestTS` (the OS pre-visibility `max`). `latestVisibleTS(vis)` = `visible[0]`'s timestamp, computed for **both** the top-T and deep-probe paths (presence.go). So a bucket whose newest match is admin/self-deleted / watermark-hidden now reports the newest *visible* message's time — the hidden timestamp no longer reaches the wire.
- **Ordering fixed too:** new `sortByVisibleLatest(calibrated)` re-sorts the returned buckets by the visible `latest_at` (stable, OS-candidate order as tiebreak). The OS `max` is retained only for *candidate selection*, which is safe (visible ⊆ all → no surfaceable bucket dropped). So a hidden newest match can no longer bias bucket order either.
- Regression coverage present; Build/Vet/Lint green (Test running).

### `match_count` — accepted owner decision, matches my original minor grading
The author kept `match_count` as the pre-visibility OS `doc_count` (option A, design §6), now with an explicit doc comment: it may exceed the caller's visible count and is flagged `match_count_approx=true`. This exposes only a possibly-inflated *count* — never a hidden message's content/sender/**time** — so it's a disclosed, non-sensitive approximation, materially lower-severity than the (now-fixed) timestamp leak. Consistent with how I graded it (minor) originally. Fine as-is.

### On yujiawei's still-standing 🔴 at this head (reconciling honestly)
Both of yujiawei's 🔴 no longer hold against `5d31a9a2`:
- "latest_at 返回未过滤的隐藏消息时间" — **resolved**: `latest_at` is now recomputed from the visible hit; the `pb.latestTS` leak path is gone (verified above).
- "latest_at 缺少近似标记" — **inverted**: `latest_at` is now **exact** (derived from the calibrated visible set), so it correctly must NOT carry an approx flag — adding one would falsely signal imprecision that no longer exists. The approx flags rightly remain on `match_count`/`total_groups` (pre-visibility by design). Recommend yujiawei re-verify at this head and clear the CR.

The metadata-leak blocker is fixed at its root (visible-set recompute for both `latest_at` and ordering) with the fix I recommended; `match_count` is a documented, disclosed approximation (my minor, accepted); authz scope was already byte-verified clean (allowlist-intersection, filterVisible five-gate, bounded deep-probe). No blockers remain from my side. Build/Vet/Lint green; conditional on Test. `check-sprint` governance is the maintainer's to clear.

Jerry-Xin (approved):
Re-verified at head `5d31a9a211ed`: the `latest_at` / bucket-ordering metadata leak from my prior CR is **fixed**. Lifting to APPROVE.

**Byte-verified fix (`search_global_groups.go` + `presence.go`):**
- `presence.go:164` `latestVisibleTS(visible)` returns `hitTimestampMS(visible[0])` — the timestamp of the most-recent **filterVisible-passed** hit.
- `presence.go:111,125` set `calibratedBucket.latestVisibleTS = latestVisibleTS(vis)` on both the standard INCLUDE path and the deep-probe path, from the DB-visible hit set.
- `search_global_groups.go:341` passes `cb.latestVisibleTS` to `assembleGroupBucket` (no longer `pb.latestTS`); `:317` `sortByVisibleLatest` re-sorts the final wire buckets by the visible timestamp (stable, OS pre-filter max used only as a deterministic tiebreak). So a bucket whose newest match is admin/self-deleted / cleared-history / visibles-hidden can no longer leak that hidden message's timestamp or bias the returned order.
- `pb.latestTS` (OS pre-filter `max`) is retained only as an internal diagnostic and as the candidate-selection order (visible ⊆ all, so no surfacing bucket is dropped) — never emitted to the wire.

**`match_count`:** kept as the pre-visibility OS `doc_count` with `match_count_approx=true` (owner design "option A"). This is a disclosed, coarse cardinality approximation — it exposes only a possibly-inflated count, never a hidden message's content/sender/timestamp — so it's a defensible design choice signaled on the wire, not the metadata leak that was the blocker. `latest_at`, by contrast, is now exact-visible and correctly carries no approx flag.

**Authz remains clean** (unchanged from prior review): in-query membership scoping via `resolveGlobalScope`/`buildAllowlist` (allowlist ∩ client ids, fail-closed), `filterVisible` five-gate, SQLi-safe parameterized queries, K2-bounded deep-probe.

🔴 Blocking: None. The one-spot recompute fix is in and correct. APPROVE.

yujiawei (approved):
# Code Review — PR #593 (octo-server) — re-review at head 5d31a9a2

Re-reviewed at head SHA `5d31a9a211edb82c1a871f7cb371d83264f47b82` (moved from `277d798…`; fast-forward + 2 new commits `50e7c9ee`, `5d31a9a2`) against `origin/main`. The prior `CHANGES_REQUESTED` blocker has been resolved. `go build`, `go vet`, and `go test ./modules/messages_search/` all pass at this head.

## 1. Spec Compliance

**Spec: ✅** — unchanged from the prior pass; all deliverables land, no missing/over-build/deviation.

## 2. Code Quality

**Quality: Approved**

### P1 (previously blocking) — RESOLVED: `latest_at` + bucket order now derive from the calibrated visible hit

`50e7c9ee` fixes the pre-visibility timestamp leak exactly as needed:
- `latest_at` is now sourced from `calibratedBucket.latestVisibleTS` = `latestVisibleTS(visible[0])`, the newest hit that passed `filterVisible` — never the OS `max(timestamp)` (`presence.go`, `assembleGroupBucket` now takes `latestAtMS` explicitly). The UNRESOLVED/deep-probe path carries the probe-found visible hit's timestamp too.
- The returned `groups[]` are re-sorted by that visible time via `sortByVisibleLatest` (stable, so equal timestamps keep the OS candidate order as a deterministic tiebreak). The candidate-set argument is sound: the OS terms agg still orders *candidates* by pre-filter max, but `visible ⊆ all`, so a bucket whose visible latest is recent also has an at-least-as-recent pre-filter max — no bucket that should surface is dropped.
- Regression tests added and passing: `TestCalibratePresence_LatestVisibleTS` (top-T path — asserts the older visible ts, not the deleted newest), `TestCalibratePresence_DeepProbeLatestVisibleTS` (deep-probe path), `TestSortByVisibleLatest_NotBiasedByHiddenHits`, `TestSortByVisibleLatest_StableTiebreak`.

### P2 (previously flagged) — RESOLVED: `filterVisible` id/channel sets now chunked

`filterVisible` splits its unique-id and channel-list into `filterVisibleChunkSize` (1000) batches before the probe round-trips, bounding every `IN (?)` regardless of the underlying `message.IService` (which builds unchunked clauses). Single-channel callers stay one chunk → byte-identical round-trip shape. Fail-closed semantics preserved (any chunk error propagates). Covered by `TestFilterVisible_ChunksLargeIDSet` and `TestChunkStrings`.

### P2 (non-blocking, remaining) — `allocatePreviewN` doc still overstates the budget ceiling

`presence.go` — the "never above budget" doc comment is still inaccurate when `PreviewBudget < len(calibrated)` (the floor-of-1 wins). Default-safe (`500 ≥ MaxGroups 200`), env-driven and unclamped. Doc-only tightening; not a merge blocker.

### `match_count` — accepted pre-visibility approximation

Left as `doc_count` + `match_count_approx: true`, now explicitly documented (`5d31a9a2`) as an accepted pre-filter approximation pending owner decision. Reasonable: it already carries the `_approx` signal, so unlike the old `latest_at` it doesn't present phantom data as exact.

## 3. Verified — no finding

All prior "verified & cleared" items still hold; additionally the two fixes above were independently re-verified (fix diffs read, regression tests run green). Build/vet/test green at head `5d31a9a2`.

## 4. Note on the first-pass ("一审") advisory

The automated first-pass suggested APPROVED but reviewed the **stale** head `277d798`, where the P1 `latest_at` leak was genuinely present — it did not surface that blocker, and its "Summary" described a notification-bot unification change that is **not in this PR's diff** (only `messages_search` files change here). This re-review's approval rests on the **current** head `5d31a9a2` and direct verification of the fixes, not on the first-pass checklist. The first-pass P2 on `allocatePreviewN` matches my own remaining non-blocking P2.

## 5. Coverage note

- **Verified:** the two new fix commits in full, the backing `visibility.go` chunking path, and all new/existing tests (`go build`/`go vet`/`go test ./modules/messages_search/` green at head `5d31a9a2`).
- **Not verified:** live L1 latency and the chunked-IN behavior against a real OS+MySQL cluster (no test environment); post-reshard EXCLUDE path (single-shard today). No cross-model advisory leg was run for this re-review (targeted re-verification of two well-scoped fixes).

OctoBoooot (approved):
## Delta re-review: #593 @ 45524a96 — pure rebase, no code change. APPROVE carries.

The head advanced 5d31a9a2 → 45524a96, which auto-dismissed my prior APPROVE. Isolated the true delta correctly (the rebase inflates a raw tree-diff — it showed incomingwebhook/notify/cardtmpl/sidebar files that are main's merged work, not #593's): **blob-compared #593's own 10 messages_search files, and every one is byte-identical (SAME)** to my approved head 5d31a9a2. `compare` = ahead 7 / **behind 4** — the branch was rebased onto a main that now includes the merged #591/#584/#588/#589, so the tree-diff noise is entirely that base movement, not a #593 change.

So there is **no new code to review** — the `latest_at` metadata-leak fix I byte-verified at 5d31a9a2 (recompute from `cb.latestVisibleTS` = `visible[0]`; `sortByVisibleLatest` re-orders by visible time; `match_count` kept as disclosed `_approx`) is present verbatim, and the authz surface (allowlist-intersection + filterVisible five-gate + K2-bounded deep-probe) is unchanged.

### Verdict: Approve (re-affirmed; conditional on Test)
My APPROVE carries verbatim onto the rebased head — the dismissal was mechanical (head advance), not substantive. No blockers, no majors; the three-reviewer-converged metadata-leak blocker remains fixed. `mergeStateStatus` stays gated on Test + `check-sprint` (governance, maintainer sets the Sprint field) — approved ≠ mergeable until those clear.

(Method note: raw tree-diff after a rebase surfaces the whole base delta; blob-comparing the PR's own file set is what isolates the true change — here, none.)

modules/messages_search/search_global_groups.go
BrDing-Rookie: Fixed in `36ce621a` (latest_at + ordering) and documented in `45524a96` (match_count).

- **latest_at** is now recomputed from the newest *calibrated visible* hit (`cb.visible[0]`), never from the OS pre-visibility `max(timestamp)`. A bucket whose newest matching message fails `filterVisible` (admin/self-deleted, cleared-history) now reports the older *visible* message's time.
- **Bucket ordering** is now driven by that visible `latest_at`, so rank no longer leaks a hidden message's existence/recency.
- **Regression test** added: a bucket whose newest match is admin-deleted must report the older visible hit's `latest_at`, not the deleted one's.
- **match_count**: kept as the OS pre-filter `doc_count` (`match_count_approx=true`) per design §6 — exact post-visibility counting is unbounded-cost and intentionally out of scope. Now explicitly documented as accepted pre-filter metadata: a possibly-inflated integer that never exposes hidden content/sender/time. 

Thanks for the precise catch on the metadata channel.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #593 (octo-server)
**PR:** Mininglamp-OSS/octo-server #593
**Head SHA:** `45524a9694b57e9f6a001960de037e8d07ea7119`
**Base:** `origin/main` (e3b05ab3)
**Title:** feat(messages_search): aggregation-first global search — L1 groups endpoint + presence calibration

---

## Summary

This PR introduces the L1 `_search_global_groups` aggregation endpoint with presence calibration (backend B). The design is sound: a single `size:0` terms aggregation over `channelId` with `top_hits` over-fetch, followed by a five-gate `filterVisible` pass that decides INCLUDE/EXCLUDE/UNRESOLVED per bucket, and a bounded K2 deep-probe for UNRESOLVED buckets. The visibility chunking in `visibility.go` correctly bounds MySQL `IN` clause width for the batch path. Audit instrumentation, fail-closed error propagation, and the RC fix (calibrated `latest_at` from visible hits, not OS pre-filter max) are all correctly implemented.

**10 files changed** across 4 commits. Core new files: `presence.go` (407 lines), `search_global_groups.go` (593 lines). Modified: `visibility.go` (+70/-14 chunking), `config.go` (+35 GroupAggConfig), `audit.go` (+10 presence stats).

---

## Verification

✅ **Aggregation DSL** — `runGroupsAggregation` (`search_global_groups.go:237-260`) correctly builds `terms(channelId).size(MaxGroups).orderByAgg("latest", false)` with `top_hits(T=PerGroupMax+PresenceProbe)` and `cardinality(channelId)`. The `applyVisiblesWhitelist` at `:228-233` is a well-formed `should(mustNot(exists(visibles)), term(visibles=uid)).minimumShouldMatch(1)` — correctly covers the empty/absent and listed cases.

✅ **Base query reuse** — `buildGlobalMessagesDSL` (`search_global_messages.go:177`) includes `mustNot(revoked=true)`, `applyExcludeVirtual`, and `applySystemMessageHardFilter`. The deep-probe closure at `presence.go:194` creates a *new* `BoolQuery.Filter(base, term(channelId))` per call — it never mutates `base`, so concurrent deep-probes are safe.

✅ **Presence calibration** — `calibratePresence` (`presence.go:63-123`) correctly batches all top-T hits into a single `filterVisible` call, classifies per bucket via `classifyPresence`, and deep-probes UNRESOLVED buckets. `classifyPresence` at `:145-151` has the right boundary: `sampleSize >= docCount` with 0 visible → EXCLUDE (never spurious UNRESOLVED).

✅ **Deep probe** — `deepProbeBucket` (`presence.go:167-213`) correctly resumes after the oldest top-T hit via `buildSearchAfterFromHit`, narrows the query to the specific channel, applies `applyVisiblesWhitelist`, and delegates to `deepProbeVisible`. The paging loop at `:226-283` has correct three-outcome semantics: first visible → INCLUDE; short page → exhausted; budget consumed → EXCLUDE. K2 budget correctly subtracts already-examined top-T hits.

✅ **Visibility chunking** — `filterVisible` (`visibility.go:258-356`) correctly chunks `uniqueIDs` and `channelList` into `filterVisibleChunkSize=1000` slices, unioning results across chunks. `chunkStrings` at `:198-212` is correct (handles empty input, size<=0 fallback). Each chunk gets independent probe calls; error propagation is fail-closed per chunk.

✅ **RC fix (calibrated latest_at)** — `buildGroupsResult` (`search_global_groups.go:299-306`) re-sorts by `latestVisibleTS` (from `calibratedBucket`), not `pb.latestTS` (OS pre-filter max). `assembleGroupBucket` at `:473` uses `msToRFC3339(latestAtMS)` where `latestAtMS` is the calibrated visible timestamp. The wire `latest_at` and the returned order both derive from the calibrated visible hit — no hidden-timestamp leak.

✅ **Preview projection** — `projectPreview` (`search_global_groups.go:405-443`) operates on already-calibrated, N-trimmed visible hits. Each hit passes through `singleMessageHit` (existing function) which handles highlight/snippet extraction via the existing `buildSearchAllHighlight` path. Sender names are batch-joined in a single pass. No uncalibrated hit reaches the wire.

✅ **has_more correctness** — `parseGroupsAggregation` at `:382` uses `terms.SumOfOtherDocCount > 0`. On the single-shard index this is exact. `groupsEnvelope` at `:589` sets `NextCursor: ""` — correct for L1 which has no per-row pagination.

✅ **Audit wiring** — `recordPresenceStats` (`presence.go:303-309`) sets gin context keys that `auditMiddleware` (`audit.go:116-123`) reads. Key names match between producer and consumer.

✅ **Config** — `GroupAggConfig` defaults (`config.go:143-147`) are reasonable: MaxGroups=200, PerGroupMax=20, PresenceProbe=20, PreviewBudget=500, K2=500. `TopHitsSize()` = 40.

---

## Findings

### P2 — `hitTimestampMS` / `latestAtMS` / `latestVisibleTS` naming says "ms" but values are epoch-seconds

**File:** `presence.go:155` (`hitTimestampMS`), `presence.go:27` (`latestVisibleTS`), `search_global_groups.go:473` (`latestAtMS`)

**Diff-scope:** new (this PR introduces these names)

The function `hitTimestampMS` is documented as "reads the OS doc `timestamp` (ms)" and the field `latestVisibleTS` is documented as "timestamp (ms)". But `Doc.Timestamp` is epoch-seconds — the existing `msToRFC3339` (`util.go:43`) explicitly documents "epoch-seconds timestamp (the value indexed in OS)" and calls `time.Unix(ts, 0)` which treats the input as seconds. The data flow is internally consistent (seconds → seconds → `msToRFC3339` → correct RFC3339), so this is not a runtime bug, but the "ms" suffix in three new identifiers creates a naming trap for future contributors who may read "ms" and introduce a `/1000` conversion.

**Suggestion:** Rename `hitTimestampMS` → `hitTimestampSec` (or just `hitTimestamp`), `latestVisibleTS` comment → "timestamp (epoch-seconds)", `latestAtMS` parameter → `latestAtSec`. Alternatively, if the team intends to migrate to milliseconds, add a clear `// NOTE: despite the name, this is epoch-seconds` comment at each site.

---

### P2 — `assembleGroupBucket` parameter `latestAtMS` carries epoch-seconds despite "MS" suffix

**File:** `search_global_groups.go:466`

**Diff-scope:** new

```go
func assembleGroupBucket(pb parsedBucket, latestAtMS int64, preview []MessageHit, ...) GroupBucket {
```

The parameter is fed from `calibratedBucket.latestVisibleTS` which is epoch-seconds. The body calls `msToRFC3339(latestAtMS)` which also expects seconds. Functionally correct, but the `MS` suffix contradicts the actual unit. Same fix as above.

---

### Nit — `parseInt` silently returns default for `0` and negative values

**File:** `config.go:180`

**Diff-scope:** pre-existing (not introduced by this PR, but new config entries use it)

```go
func parseInt(v string, def int) int {
    if v == "" { return def }
    if i, err := strconv.Atoi(v); err == nil && i > 0 { return i }
    return def
}
```

An operator setting `OCTO_SEARCH_GROUPS_K2=0` gets `K2=500` (default) silently. For all five new config entries, `0` is meaningless so the default is the correct fallback. No runtime risk. Consider documenting the `i > 0` contract in the function's comment for clarity.

---

## Data Flow Backtracking

| Consumer | Source | Verified? |
|---|---|---|
| `latestVisibleTS` (wire `latest_at`) | `hitTimestampMS(visible[0])` → `json.Unmarshal(hit.Source)` → `Doc.Timestamp` (epoch-seconds in OS) | ✅ Consistent: seconds → `msToRFC3339` (treats as seconds) |
| `classifyPresence` `docCount` | `parsedBucket.docCount` ← `elastic.AggregationBucketKeyItem.DocCount` (OS terms bucket) | ✅ Exact on single-shard index |
| `calibratePresence` `keep` map | `h.filterVisible(ctx, loginUID, "", allRefs)` — batched, chunked, fail-closed | ✅ All 5 gates applied |
| Deep-probe `osSearch` query | `BoolQuery.Filter(base, term(channelId))` + `applyVisiblesWhitelist` — `base` not mutated | ✅ Safe |
| `allocatePreviewN` weights | `calibratedBucket.pb.docCount` (pre-visibility OS count) | ✅ By design — N governed by match frequency |
| `assembleGroupBucket.MatchCount` | `pb.docCount` (pre-visibility) | ✅ Flagged as `MatchCountApprox: true` |
| Preview `MessageHit` | `projectPreview` → `singleMessageHit` on calibrated visible hits only | ✅ No uncalibrated hit reaches wire |
| `hasMore` | `terms.SumOfOtherDocCount > 0` | ✅ Correct for single-shard terms agg |
| Audit `presence_probe_*` | `recordPresenceStats` → gin context keys → `auditMiddleware` reads same keys | ✅ Keys match |

---

## Blind Spot Checklist

| Check | Status | Evidence |
|---|---|---|
| **C1 — Dual-path parity** | N/A | No symmetric add/remove or subscribe/unsubscribe paths in this PR |
| **C2 — Control-flow ordering / nesting** | Clear | Deep-probe closure creates fresh `BoolQuery` per call; `base` query is immutable after construction. No double-application risk. |
| **C3 — Authorization boundary** | Clear | `applyVisiblesWhitelist` is applied in both the initial aggregation (`searchGlobalGroups` handler) and the deep-probe closure (`presence.go:197`). Five-gate `filterVisible` covers all paths. |
| **C4 — Authorization lifecycle** | N/A | No auth/jail/permission changes in this PR |
| **C5 — Build/runtime parity** | N/A | Server-side Go, no build/packaging changes |
| **C6 — Governance/doc consistency** | N/A | No governance/policy documents changed |

---

## Things I Checked

- Aggregation DSL correctness (terms ordering, top_hits over-fetch, cardinality)
- `applyVisiblesWhitelist` query semantics (no-gate ∨ listed)
- Base query immutability across deep-probe calls
- `classifyPresence` boundary conditions (sampleSize vs docCount)
- Deep-probe three-outcome paging (found / exhausted / budget-exhausted)
- K2 budget accounting (subtracts already-examined top-T)
- `filterVisible` chunking (id sets + channel sets, fail-closed per chunk)
- `chunkStrings` edge cases (empty, size<=0)
- RC fix: calibrated `latest_at` from visible hits, re-sort by visible timestamp
- Preview projection uses only calibrated visible hits
- `has_more` from `SumOfOtherDocCount` (exact on single-shard)
- Audit key parity between producer and consumer
- `h.ctx` usage in `resolveThreadNames` — pre-existing pattern, consistent with `search_global.go`
- `buildGlobalMessagesDSL` includes `mustNot(revoked)`, `applyExcludeVirtual`, `applySystemMessageHardFilter`
- Highlight/snippet path reuses existing `singleMessageHit` + `buildSearchAllHighlight` (no new escaping surface)
- `groupsEnvelope` always sets `NextCursor: ""` (correct for L1 no-pagination design)
- `parseInt` behavior with `0` / negative — pre-existing, documented above

---

## [Octo-Q] verdict: APPROVED

All findings are P2 (naming consistency) or Nit. No P0/P1 blocking issues identified. The aggregation-first L1 endpoint correctly implements presence calibration with fail-closed visibility gating, bounded deep-probe, and calibrated `latest_at`. Data flow from OS aggregation through calibration to wire response is verified end-to-end.

yujiawei (approved):
# Code Review — PR #593 (octo-server)

Reviewed the full diff against merge-base `e3b05ab3`. `go build ./...`, `go vet ./modules/messages_search/`, and `go test ./modules/messages_search/` are all green locally at head SHA `45524a96`.

## Verdict: APPROVED

No blocking (correctness/security) issues found. The visibility model is sound and the two review-response commits correctly close the earlier recency-leak and unbounded-IN concerns. A few non-blocking suggestions below.

## Spec compliance

The diff implements exactly what the linked issue (#592) describes and nothing beyond it:

- New endpoint `POST /v1/messages/_search_global_groups` mounted via `registerRoute`, so it inherits the full shared chain (`AuthMiddleware` → `SharedUIDRateLimiter` → `SpaceMiddleware` → per-uid search limiter → audit → `backendGate`) identical to its siblings (`api.go:125`). `route_chain_test.go` bumps the endpoint counter 7→8, so the "all endpoints go through the shared chain" guard stays honest.
- Stricter trigger gate — only `keyword ∨ sender_ids ∨ member_uids/member_uid ∨ channel_ids` opens an aggregation; `sent_at` / `content_types` / `channel_types` alone return VALIDATION_ERROR (`groupsTriggerSatisfied`, `search_global_groups.go:185`), deliberately not reusing `hasEffectiveFilters`. Covered by `TestGroupsTriggerSatisfied`.
- All five tunables (`maxGroups`/`perGroupMax`/`presenceProbe`/`previewBudget`/`K2`) are config/env-driven, never request-controlled (`config.go:125`).
- `match_count` / `total_groups` remain the pre-visibility OS `doc_count` / cardinality HLL, both flagged `*_approx = true` — matches the issue's "count approximate" decision.

## Visibility / correctness (the load-bearing part)

- **No pre-visibility content/sender/time leak.** The raw `top_hits` are never projected to the wire; every preview row passes the five-gate `filterVisible` (`presence.go:55`, `projectPreview` at `search_global_groups.go:405`). `TestCalibratePresence_IncludeFiltersPreview` asserts an admin-deleted hit mixed into the top-T never reaches preview.
- **`latest_at` recency leak fixed correctly.** `latest_at` and the wire bucket order both derive from `calibratedBucket.latestVisibleTS` (the newest *visible* hit), not the OS pre-filter `max(timestamp)`. `assembleGroupBucket` takes the visible ts as a parameter rather than reading `pb.latestTS` (`search_global_groups.go:464`), and `sortByVisibleLatest` re-sorts survivors by it with a stable tiebreak. `TestCalibratePresence_LatestVisibleTS`, `TestCalibratePresence_DeepProbeLatestVisibleTS`, and `TestSortByVisibleLatest_NotBiasedByHiddenHits` all cover this.
- **Fail-closed holds.** Any `filterVisible` DB error propagates out of `calibratePresence`/`deepProbeVisible`; `buildGroupsResult`'s error branch routes a classifiable OS error to UPSTREAM and everything else (notably the DB gate) to INTERNAL_ERROR — never falling through to uncalibrated hits (`search_global_groups.go:162`, `TestCalibratePresence_FailClosed`).
- **Three-state classify boundary is right.** `sampleSize >= docCount` with 0 visible → EXCLUDE (not a spurious UNRESOLVED); the `T == doc_count` boundary is tested (`TestClassifyPresence` case ⑤). Correct on the single-shard index where `doc_count` is exact.
- **Deep probe paging is safe.** `search_after` resumes strictly after the oldest top-T hit via `buildSearchAfterFromHit` (full-precision id from typed `_source`, not the float64-rounded `Sort`), pages old-ward with no overlap/skip, and the K2 budget correctly counts the top-T already examined (`budget := K2 - len(pb.hits)`). First-visible / channel-exhausted / K2-exhausted outcomes are each tested (cases ②③⑥⑦).
- **Unbounded-IN fixed.** `filterVisible` now chunks id and channel sets at `filterVisibleChunkSize = 1000` before hitting the un-chunked `message.IService` `IN (?)` queries; single-channel callers stay one chunk, so their round-trip shape is unchanged (`visibility.go:172`, `TestFilterVisible_ChunksLargeIDSet` / `TestChunkStrings`).
- DM fake-channel reversal, thread `parent_group_no`/`thread_id`, and group self-parent shapes are all correct and unit-tested (`classifyGroupBucket`, `TestAssembleGroupBucket_*`). DM `channel_offset` key alignment (fake id → peer uid) is preserved in `projectDocRef`.

1. **Deep-probe fan-out is sequential per UNRESOLVED bucket.** In the worst case (many buckets whose newest T matches are all hidden), `calibratePresence` runs up to `maxGroups` deep probes back-to-back, each up to `K2/deepProbePageSize` OS+MySQL round-trips (`presence.go:116`). It's fully bounded (K2 caps each, the request Timeout caps the whole), so not a correctness problem — but under a pathological query the tail latency could be significant. Consider capping the number of buckets deep-probed per request, or bounding total probe work across buckets rather than only per-bucket, and watch the `presence_probe_buckets`/`presence_probe_docs` instrumentation you added in production.

2. **`context` not threaded into the name-resolution / thread-DB lookups.** `resolveGroupNames`/`resolvePeerNames`/`resolveThreadNames` and `thread.NewDB(h.ctx)` don't take the request `ctx`, so they won't cancel when the request times out or the client disconnects, unlike `senderJoin(ctx, ...)`. These are soft-fail and off the hot path, so low impact — but plumbing `ctx` through would make timeout behavior uniform.

3. **Legacy-response guard test coverage gap (pre-existing, not introduced here).** `TestMessagesSearchNoLegacyResponseError` uses a hardcoded file list that already omits all the global-search files (`search_global*.go`, `search_around.go`), and the new `search_global_groups.go` / `presence.go` are likewise not added. The new code is clean (verified: no `.ResponseError(`/`.JSON(status`/`AbortWithStatus`), so this is not a violation — but the guard silently doesn't protect these files. Worth switching the guard to auto-discover `*.go` (minus the documented exceptions) in a follow-up so new handlers can't drift.

## Coverage note — what I could not fully verify from the diff

- **Live latency / OS+MySQL behavior**: no OS+MySQL environment was available, so the end-to-end L1 latency and the real deep-probe cost under production data volumes are unverified (the PR itself notes live numbers require that environment).
- **Indexer contract for `visibles` / `channelId` / virtual-doc `parentMessageId`**: the OS-layer `visibles` gate and DM `channelType` reversal assume the indexer populates these fields as documented; the indexer side is out of this diff.
- **`total_groups` cardinality accuracy** under real cardinality is an HLL estimate by design; not independently measured.

_Independent second-opinion tooling was attempted for this review: one cross-model leg was blocked by an upstream gateway error and a second returned only a partial (token-truncated) trace, so neither produced additional actionable findings. The findings above are from direct review of the diff, and their absence is not treated as agreement._

PR #588, opened 2026-07-15
feat: add generic card action callback dispatch

an9xyz:
## Summary

Add a generic, server-owned approval-card lifecycle for first-party services. A standard consumer configures an owner/action route with `notify_token_env`, sends a structured `approval_card`, and implements one signed decide endpoint; octo-server owns the interactive card, reliable callback delivery, terminal mutation, and requester notification. Docs access approval is the first specialized-template pilot. External and third-party bots keep the existing pull-based Bot event flow.

## Related Issue

Follow-up to #584. No issue is closed by this PR.

## Linked Spec

`.octospec/tasks/card-action-callback-dispatch/brief.md`

## Changes

- Route registered first-party card actions by authoritative `(sender_uid, owner, action_type)` while preserving the existing external Bot queue path.
- Add a static HTTPS callback registry, HMAC-signed typed delivery, Redis ready/leased/DLQ state machine, token-bound finalization heartbeats, attempt-neutral capacity deferral, bounded retries, timeouts, metrics, and a manual DLQ replay command.
- Finalize cards inside the trusted mutation boundary and notify the requester after an authoritative consumer decision.
- Add a standard approval lifecycle for new owners: a route may bind `notify_token_env`, which dynamically installs an owner-bound `octo/v2` producer and authorizes structured `approval_card` requests. Callers cannot choose owner/sender, submit callback URLs/card JSON, or invoke another owner's/callback-only action.
- Route typed results through a finalizer registry: docs keeps its specialized resource-card rendering while every other registered owner gets the server-owned standard terminal card and requester outcome.
- Add the docs approve/deny card pilot, capability-scoped notify token, outcome cards, feature flag, and localized templates.
- Cover DM finalization with the real WuKongIM lookup and production card mutator so terminal state is written to the approving user's stored DM channel.
- Add a consumer integration guide with the complete request/response contract, canonical HMAC, idempotency/CAS requirements, retry semantics, and a TypeScript/Express verification example. Callback `event_id` is a decimal JSON string so JavaScript consumers preserve the full `int64` identity.

## Testing

- [x] Unit tests added/updated
- [x] Manually verified
- `go test -race -cover ./internal/cardactiondispatch -count=1` (81.8% statement coverage)
- `go test -race ./internal/carddispatch ./modules/bot_api ./modules/notify ./pkg/cardtmpl . -count=1`
[...]

PR #595, opened 2026-07-15
feat(cardaction): allow HTTP callbacks and bounded custom approval_card actions

an9xyz:
## Summary

Follow-up to #588. Two small extensions to the callback dispatch layer plus one bundled config collapse:

1. `OCTO_CARD_ACTION_ROUTES[].url` now accepts `http://` alongside `https://`. Hostname form is not inspected; the presence of `http://` in a static route is the operator's explicit authorization.
2. `approval_card` grew an optional 1–5 bounded `actions` list, so a consumer can render buttons like `execute` / `reject` / `cancel` instead of only `approve` / `deny`. Server owns action IDs, reserved metadata, and card JSON. Nil preserves the byte-for-byte legacy approve/deny output; a non-nil empty slice is a caller bug and returns 400.
3. `OCTO_CARD_ACTION_ALLOWED_URLS` is deleted from routing decisions. If still present at startup it produces one structured deprecation WARN and is ignored — rolling upgrades do not fail.

HMAC signing, redirect ban, proxy-env ignoring, retry/DLQ, terminal states, and requester-notification behavior are deliberately unchanged.

## Linked Spec

`.octospec/tasks/card-action-internal-http-actions/brief.md`

## Changes

- `internal/cardactiondispatch/registry.go` — two-arg `NewRegistry`; `validateCallbackURL` accepts `http`/`https` and rejects credentials/query/fragment/opaque/empty-host, plus raw `#` and `ForceQuery` prefilters.
- `internal/cardactiondispatch/config.go` — `LoadAllowedURLs` deleted.
- `internal/cardactiondispatch/http.go` — comment update; `Proxy=nil` now covers both schemes.
- `main.go` — reads deprecated env only to emit a structured WARN; installs registry with the new signature.
- `pkg/cardtmpl/approval_request.go` — new `ApprovalRequestAction` / `MaxApprovalCustomActions=5`; validator checks control characters on the raw title before `TrimSpace`.
- `modules/notify/model.go`, `modules/notify/approval_card.go` — `ApprovalCardFields.Actions` optional field with strict nil-vs-empty semantics.
- Docs — `docs/card-action-callback-dispatch.md`, `docs/card-action-callback-consumer.md`, `docs/card-protocol.md` updated by reference.
- Tests — new `internal/cardactiondispatch/registry_http_scheme_test.go` (7 accept + 13 reject + plain-HTTP HMAC and redirect E2Es); new `pkg/cardtmpl/approval_request_actions_test.go` (actions boundaries + golden-bytes for legacy path); new orchestration integration test `TestCardActionCallbackOrchestrationForwardsCustomDecisionEndToEnd`; existing tests migrated to the two-arg registry signature (this also fixed the compile block in `modules/message/api_card_action_test.go`).
- `.octospec/journal/shared/card-action-internal-http-actions.md`, `.octospec/learnings/pending/{json-nil-vs-empty-slice,url-parse-raw-prefilter}.md`.

## Testing

- [x] Unit + race — `cardactiondispatch`, `cardtmpl`, `notify`, main package: PASS.
- [x] Coverage — cardactiondispatch 81.5%, cardtmpl 89.9%, notify 71.2%.
- [x] Integration (fresh local `test` schema) — full `internal/cardactiondispatch` orchestration suite including the new custom-decision E2E: PASS.
- [x] Integration (fresh schema) — `modules/message` sender-bound routing E2E (compile-blocker fixup): PASS.
- [x] `go build ./...`, `go vet ./...`, `gofmt -l`, `make i18n-lint`: clean.
- [x] Manual byte-compat check — omitting `actions` produces the exact same JSON as before this PR.

Unrelated pre-existing test flake: `TestCardActionEndToEndAndIdempotency` panics without `OCTO_MASTER_KEY` on both this branch and clean `main`; not in scope for this PR.

## COMPREHENSION

   For the callback URL: `validateCallbackURL` no longer forces `https`. It accepts `http` and `https`, verifies the URL has a real host (`Hostname() != ""`), no credentials, no query (including bare `?`), no fragment (including bare `#` and embedded `#`), and no opaque form. The rest of the callback path — HMAC signing, canonical body, redirect refusal, proxy-env ignoring, retry/DLQ — is byte-identical.

   For approval_card actions: a nil `Actions` slice takes the untouched legacy code path (localized approve/deny buttons, byte-identical output). A non-nil slice enters a strict validator that produces server-built `Action.Submit` buttons whose IDs are `approval-<decision>` and whose payloads carry the server-injected `owner`, `action_type`, and per-button `decision`. Callers cannot supply per-action data, style, URL, input, or arbitrary card JSON, and cannot mix custom actions with approve/deny defaults.

   For config: `OCTO_CARD_ACTION_ALLOWED_URLS` no longer feeds routing. It is read once at startup only to log a structured deprecation WARN.

   - Deployments that route to a plain HTTP endpoint over an untrusted network — HMAC does not encrypt. Both docs carry an explicit residual-risk callout.
   - A caller who currently ships `"actions": []` (explicit empty array) will start getting 400 instead of the old approve/deny fallback. There are no known callers doing this; this is intentional to catch caller bugs.
   - A caller who registered a route that only worked because of shape validation (e.g. lookalike `service.namespace.svc.example.com`) is no longer relied on — the route registry is the single source of truth. If a bad URL was accidentally in the registry it will still be accepted; that risk is unchanged from the pre-PR world where the registry was co-authoritative with the allowlist.
   - Deployments still carrying `OCTO_CARD_ACTION_ALLOWED_URLS` in their ConfigMap will keep booting; a WARN will show in logs until the variable is removed.

   - Byte-compat: `TestBuildApprovalRequestCardOmittedActionsIsByteCompatible` pins the exact JSON produced when `Actions == nil`.
   - URL boundary: `TestValidateCallbackURLRejectsUnsafeShapes` covers empty, unsupported schemes, credentials, query, trailing `?`, fragment, trailing `#`, embedded `#`, `http://` and `https://` with port-only host, and whitespace-padded strings.
   - HTTP acceptance: `TestNewRegistryAcceptsHTTPCallbackDestinations` covers K8s Service short/full DNS, docker service name, `host.docker.internal`, IPv4/IPv6 literals, and HTTPS.
   - Cleartext transport: `TestHTTPDelivererDeliversPlainHTTPWithHMACAndDisciplinedTransport` runs a real `httptest.NewServer` with `HTTP_PROXY` set in the environment, asserting the deliverer's default transport ignores it and HMAC verifies over the exact body.
   - Redirect ban on HTTP: `TestHTTPDelivererRefusesRedirectsOnPlainHTTP` asserts the returned error is a `DeliveryError{Category: "redirect_rejected"}`.
   - Custom decision end-to-end: `TestCardActionCallbackOrchestrationForwardsCustomDecisionEndToEnd` enqueues a click, verifies the callback body carries `decision=execute` and `action_id=approval-execute`, that the standard finalizer renders `approval.approved` (not the decision string), that the requester notification fires, and that the queue drains.
   - Wire-level nil-vs-empty: `TestIntegration_ApprovalCardCustomActionsRenderInOrder` sends raw JSON with `"actions": []` and asserts 400, then sends raw JSON with the field omitted and asserts a successful approve/deny render.
   - Deprecation contract: `TestDeprecatedAllowedURLsEnvIsIgnoredWithStructuredWarn` pins that `main.go` still reads the deprecated env, emits `log.Warn` with a `deprecated_env` structured field, no longer calls `LoadAllowedURLs`, and uses the two-arg `NewRegistry`.

## Checklist

- [x] Updated documentation

### DIFF
--- a/.octospec/journal/shared/card-action-internal-http-actions.md
+++ b/.octospec/journal/shared/card-action-internal-http-actions.md
@@ -1,0 +1,160 @@
+---
+type: Journal
+title: "Journal: card-action-internal-http-actions"
+description: Follow-up to #588 — allow plain HTTP callbacks and 1-5 bounded custom actions on approval_card without weakening the server-owned card boundary.
+tags: ["card", "internal", "trust-boundary", "wire-contract", "testing"]
+timestamp: 2026-07-16T02:00:00+08:00
+# --- octospec extension fields ---
+task: card-action-internal-http-actions
+source: self
+---
+
+# Journal: card-action-internal-http-actions
+
+## What was done
+
+Two small follow-ups to #588 plus one bundled configuration collapse:
+
+- `internal/cardactiondispatch` — `validateCallbackURL` now accepts both
+  `http://` and `https://`; the static `OCTO_CARD_ACTION_ROUTES[].url` list is
+  itself the exact allowlist. Reject-cases were tightened at the same time:
+  `u.Hostname() != ""` (blocks `http://:8080/x`), `u.ForceQuery` (blocks
+  trailing `?`), raw-`#` prefilter (blocks trailing `#` and embedded `#` that
+  `url.Parse` would swallow), unsupported schemes, credentials, opaque forms.
+  `NewRegistry` is now two-argument; `LoadAllowedURLs` was deleted.
+- `main.go` — reads `OCTO_CARD_ACTION_ALLOWED_URLS` only to emit one structured
+  deprecation WARN (`deprecated_env=OCTO_CARD_ACTION_ALLOWED_URLS`), then
+  ignores it. Rolling upgrades that still carry the variable do not fail
+  startup.
+- `pkg/cardtmpl` — added `ApprovalRequestAction` and an optional
+  `ApprovalRequestCard.Actions` slice. When `Actions == nil` the byte-for-byte
+  legacy `approve/deny` output stays intact (golden-bytes test). When non-nil,
+  1..5 buttons are rendered with server-derived action IDs
+  (`approval-<decision>`), owner/action_type/decision injected into every
+  submit payload, and `data` reserved-key checks unchanged. A non-nil empty
+  slice is treated as a caller bug, not a silent fallback. Titles are checked
+  for control characters on the raw string before `TrimSpace` runs.
+- `modules/notify` — added `ApprovalCardFields.Actions` (`json:"actions,omitempty"`)
+  and `ApprovalCardAction`. Delivery routes nil to the localized approve/deny
+  path; any non-nil slice enters the strict validator.
+- Contract-critical tests are consumer-neutral: 7 HTTP-scheme acceptance
+  shapes, 13 URL rejection shapes, plain-HTTP HMAC-and-transport E2E, plain-HTTP
+  redirect rejection with typed `redirect_rejected` category, and a
+  custom-decision orchestration E2E (`approval-execute` → HMAC callback →
+  consumer maps to `approved` → standard finalizer renders standard terminal
+  wording → requester notification).
+
+Locked wire-contract decisions (drawn from the discussion draft, not from any
+one owner):
+
+- Actions optional; when present 1-5 items, `decision` matches
+  `^[a-z][a-z0-9_.-]{0,47}$`, `title` is 1-80 runes, no control characters,
+  unique decisions per card, no mixing with approve/deny defaults.
+- Terminal states unchanged: `approved`/`denied`/`cancelled`/`pending`.
+  Consumers map their custom `decision` to a state — the server does not
+  translate.
+- Only `approved`/`denied` notify the requester, matching #588. Custom
+  `decision` values never trigger direct notification.
+- Caller-provided button i18n is single-copy per request; per-recipient
+  language rendering is explicitly out of scope for this change.
+
+## Load-bearing intent
+
+- **URL is operator-authorized, not shape-recognized.** The presence of
+  `http://` in a static route is the authorization. octo-server does not
+  inspect hostnames (K8s DNS, docker service names, `host.docker.internal`,
+  IP literals). Reachability is a NetworkPolicy/service-mesh concern; scheme
+  choice is a confidentiality decision the operator makes explicitly.
+- **Config is single-source.** `OCTO_CARD_ACTION_ROUTES` is the exact URL
+  registry. `OCTO_CARD_ACTION_ALLOWED_URLS` shared the same ConfigMap boundary,
+  produced dual-write drift, and gave no independent review — hence removed.
+- **Card content is server-built.** Callers pass `decision` + display `title`;
+  the server owns action IDs, injects reserved metadata, and remains the only
+  code path that can add a URL/style/input. `data` reserved-key checks
+  (`owner`/`action_type`/`decision`) still apply.
+- **Nil ≠ empty.** `Actions == nil` means "omitted, use defaults." An explicit
+  `"actions": []` is a caller bug and is rejected at 400, not silently
+  downgraded. This closes a subtle bypass where a caller intending zero
+  buttons would get the localized approve/deny anyway.
+- **Terminal contract unchanged.** No new callback result states, no
+  per-action data/URL/input/style, no arbitrary card JSON, and no
+  custom-decision path into requester notification. The wire contract from
+  #588 is deliberately not widened.
+- **HMAC does not encrypt.** Both docs (`callback-dispatch.md`,
+  `callback-consumer.md`) now carry an explicit residual-risk callout for
+  plain-HTTP transport and for the absence of independently-signed callback
+  responses.
+
+## Deferred (out of scope, tracked)
+
+- Independently-signed callback responses. HTTP transport is still
+  fail-closed on the request signature; response tampering remains a residual
+  risk.
+- Custom `outcome_label` / business terminal wording. Standard approval
+  wording (`approval.approved` / `denied` / `cancelled`) is what renders
+  regardless of the custom decision.
+- Non-terminal "still clickable" actions. Every finalizer still removes the
+  buttons on any typed response; `pending` shows the standard "unavailable"
+  terminal wording.
+- Per-recipient language rendering of caller-supplied button titles.
+- `OCTO_DOCS_APPROVAL_CARD_ENABLED` behavior. Kept as-is; separate rollout.
+
+## Injected rule compliance
+
+- **error-handling**: no new response paths; `/v1/internal/notify` already
+  goes through the localized envelope. Card-build failures still surface as
+  the existing `errNotifyCardInvalid`. Custom-decision validation errors
+  reject at 400 through the same route.
+- **rate-limit**: no route, middleware, or Redis frequency-counter change.
+- **wire-contract**: an unset optional field on the JSON boundary preserves
+  legacy output byte-for-byte (`TestBuildApprovalRequestCardOmittedActionsIsByteCompatible`
+  pins golden bytes).
+- **testing**: focused unit + race + integration + a new orchestration E2E
+  exercise every failure category the reviewer named — nil vs non-nil-empty
+  actions, raw `#`/`?`/empty-host URLs, leading-whitespace control chars,
+  plain-HTTP HMAC over cleartext, plain-HTTP redirect rejection, and
+  proxy-env-variable ignoring.
+
+## Verification
+
+- `go build ./... && go vet ./... && gofmt -l ...` — PASS.
+- `go test -race ./internal/cardactiondispatch/... ./pkg/cardtmpl/... ./modules/notify/... .` — PASS.
+- Coverage — cardactiondispatch 81.5%, cardtmpl 89.9%, notify 71.2%.
+- `go test -tags=integration -race ./internal/cardactiondispatch/...` on a
+  freshly-recreated local `test` schema — PASS (all 3 orchestration E2Es
+  including new custom-decision path).
+- `go test -tags=integration -race -run TestCardActionSenderBoundCallbackRouting ./modules/message/...`
+  on a freshly-recreated schema — PASS. This is the compile-blocker the
+  reviewer flagged; its `NewRegistry` call site was migrated to the
+  two-argument signature.
+- `make i18n-lint` (direct-error-response + unregistered-code) — PASS.
+
+## Gotchas worth remembering
+
+- Go's `url.Parse` silently drops trailing empty `#` and produces
+  `Fragment == ""`. Checking the parsed field alone lets `https://host/path#`
+  through. Prefilter the raw string for `#` before parsing. The same applies
+  in principle to `?`, though `u.ForceQuery` covers that case.
+- `strings.TrimSpace` also strips control-class whitespace (`\t \n \r \v`).
+  Any "reject control characters" rule must run on the raw string, not the
+  trimmed one, or leading/trailing tabs and newlines slip through.
+- `omitempty` only affects encode. For decode, `nil` (field absent) and
+  `[]` (explicit empty array) are distinct on the wire; treat them as
+  different states rather than collapsing with `len() == 0`.
+- Local integration tests need the shared `test` schema dropped/recreated
+  between packages, matching CI. `Unable to create migration plan / unknown
+  migration` is the canonical symptom — do not try to reconcile migrations
+  in-place.
+
+## Rollout
+
+- No DB migration. `kubectl rollout` is sufficient.
+- Deploy → verify no `OCTO_CARD_ACTION_ALLOWED_URLS deprecated` WARN (or
+  clean it up when convenient).
+- Add each new consumer as one JSON route entry plus its
+  `<OWNER>_CARD_ACTION_SECRET` + `<OWNER>_NOTIFY_TOKEN` secrets. `http://`
+  in the route URL is a conscious operator authorization.
+- Rollback: delete the route from `OCTO_CARD_ACTION_ROUTES` and restart, or
+  change its `url` back to HTTPS. In-flight events retry until the new URL
+  is reachable or the attempt cap is exhausted (then DLQ, replayable via
+  `tools/card-action-dlq`).
--- a/.octospec/learnings/pending/json-nil-vs-empty-slice.md
+++ b/.octospec/learnings/pending/json-nil-vs-empty-slice.md
@@ -1,0 +1,47 @@
+---
+type: Learning
+title: "Optional JSON slice/map: nil is not empty"
+description: For optional collection fields on a request boundary, treat nil (field omitted) and non-nil-empty (explicit []) as different states. Collapsing with len()==0 silently downgrades a caller bug into a fallback.
+tags: ["api", "validation", "wire-contract", "review"]
+timestamp: 2026-07-16T02:00:00Z
+# --- octospec extension fields ---
+source: self
+origin_task: card-action-internal-http-actions
+origin_pr: dmwork-org/octo-server (approval_card actions follow-up)
+status: pending
+candidate_rule: error-handling
+---
+# Optional JSON slice/map: nil is not empty
+
+When adding an optional collection field to an existing request struct
+(`Actions []Foo` / `Filters map[string]string` / `Ids []int`), a natural
+"caller didn't set it → use default" branch reaches for `len(field) == 0`.
+That collapses two wire-level states that JSON keeps distinct:
+
+- `field` omitted → decoded `nil` → the caller wants the default
+- `"field": []` → decoded non-nil empty → the caller sent zero elements,
+  usually a bug in their generation code
+
+`len() == 0` treats both as "use default." A caller who accidentally builds
+an empty array (map lookup miss, filter that removed everything, framework
+serializing `Optional.empty()` as `[]`) gets a silent, incorrect fallback
+instead of a 400.
+
+**Rule candidate:** for any optional collection on a request struct, split
+the branch by nilness:
+
+```go
+switch {
+case field == nil:
+    // caller omitted → apply default
+case len(field) == 0:
+    // caller sent explicit empty → 400 (caller bug, don't silently downgrade)
+default:
+    // validate and use
+}
+```
+
+`omitempty` only affects encode; decode still preserves the distinction. Caught
+in review on the approval_card actions follow-up
+(`err.server.notify.card_invalid` at the ingress; `cardtmpl` layer treats
+non-nil-empty as an explicit rejection error).
--- a/.octospec/learnings/pending/url-parse-raw-prefilter.md
+++ b/.octospec/learnings/pending/url-parse-raw-prefilter.md
@@ -1,0 +1,46 @@
+---
+type: Learning
+title: "URL boundary checks must prefilter the raw string, not the parsed struct"
+description: Go's net/url silently drops empty fragments and folds ports into Host. Validating only url.URL fields lets malformed raw URLs through. Reject on the raw string before parsing.
+tags: ["ssrf", "url", "validation", "trust-boundary", "review"]
+timestamp: 2026-07-16T02:00:00Z
+# --- octospec extension fields ---
+source: self
+origin_task: card-action-internal-http-actions
+origin_pr: dmwork-org/octo-server (approval_card actions follow-up)
+status: pending
+candidate_rule: security
+---
+# URL boundary checks must prefilter the raw string, not the parsed struct
+
+Validating callback / webhook URLs by inspecting `url.URL` fields alone is
+insufficient. `net/url` is deliberately lenient in three places that matter
+for a trust-boundary check:
+
+1. **Empty fragment is discarded.** `url.Parse("https://host/path#")`
+   produces `Fragment == ""` — indistinguishable from a URL that had no `#`
+   at all. Checking `u.Fragment != ""` accepts both. `u.ForceQuery` provides
+   the equivalent for a trailing `?`, but there is no `ForceFragment` — you
+   must reject `#` on the raw string.
+
+2. **Host contains the port.** `url.Parse("http://:8080/x").Host == ":8080"`
+   is non-empty and looks valid. `u.Hostname()` strips the port and returns
+   `""`, which is what a hostname predicate actually wants.
+
+3. **Scheme-only oddities.** `url.Parse("https:opaque")` leaves `u.Opaque`
+   set and `u.Host == ""`. Some strict scheme checks miss this because the
+   scheme is right; the shape isn't.
+
+**Rule candidate:** at any external-URL trust boundary (callback allowlist,
+webhook target, image proxy, redirect destination), the validator should:
+
+- trim-then-reject-if-mutated (`strings.TrimSpace(raw) != raw`),
+- prefilter the raw string for `#` (and for `?` if you don't rely on
+  `ForceQuery`),
+- parse, then check `u.Scheme` against a small allowlist, `u.Hostname() != ""`
+  (not `u.Host`), `u.Opaque == ""`, `u.User == nil`,
+- reject `u.Fragment != ""`, `u.RawQuery != ""`, `u.ForceQuery`.
+
+Caught in review on the approval_card HTTP-scheme follow-up
+(`internal/cardactiondispatch.validateCallbackURL`). Both misses had test
+vectors: `https://host/path#` and `http://:8080/path`.
--- a/.octospec/log.md
+++ b/.octospec/log.md
@@ -4,6 +4,25 @@ Change history for this repo's `.octospec/`, following the
 [OKF](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md)
 change-log convention (§7). Newest first.
 
+## 2026-07-16 (card-action-internal-http-actions)
+
+- **Follow-up** — Two small extensions to #588 plus one bundled config
+  collapse. `OCTO_CARD_ACTION_ROUTES[].url` now accepts `http://` in addition
+  to `https://`; hostname form is intentionally not inspected (route
+  registration = operator authorization). URL validator tightened at the same
+  time: `Hostname() != ""` (blocks `http://:8080/x`), `ForceQuery` (blocks
+  trailing `?`), raw-`#` prefilter (blocks trailing/embedded `#`).
+  `OCTO_CARD_ACTION_ALLOWED_URLS` is deleted from code paths and emits a
+  structured deprecation WARN if still set, so rolling upgrades do not fail.
+  `approval_card.actions` grew an optional 1..5 bounded slice: server-derived
+  action IDs, reserved metadata enforced, control-character-in-title checked
+  on the raw string, `nil` preserves byte-for-byte legacy approve/deny while a
+  non-nil empty slice is rejected as a caller bug. Callback wire contract
+  (states, requester notification, HMAC canonical) is deliberately unchanged.
+  Coverage — cardactiondispatch 81.5%, cardtmpl 89.9%, notify 71.2%. Brief/context
+  under `.octospec/tasks/card-action-internal-http-actions/`; shared journal
+  `.octospec/journal/shared/card-action-internal-http-actions.md`.
+
 ## 2026-07-13 (card-message-appbot-trust)
 
 - **Fix** — Closed the P0 App Bot card trust split without changing the send
--- a/.octospec/tasks/card-action-internal-http-actions/brief.md
+++ b/.octospec/tasks/card-action-internal-http-actions/brief.md
@@ -1,0 +1,162 @@
+---
+type: Task
+title: "Task: card-action-internal-http-actions"
+description: Allow plain HTTP callbacks declared in the static route registry and bounded custom actions on the generic approval_card ingress, without weakening the server-owned card boundary or the config-driven URL allowlist.
+tags: ["card", "internal", "trust-boundary", "wire-contract", "testing"]
+timestamp: 2026-07-16T10:00:00+08:00
+slug: card-action-internal-http-actions
+upstream: "TBD — follow-up to octo-server#588"
+source: self
+---
+
+# Task: card-action-internal-http-actions
+
+## Goal
+
+Two small follow-ups to the callback dispatch layer merged in #588:
+
+1. Allow `OCTO_CARD_ACTION_ROUTES[].url` to use plain `http://` in addition to
+   `https://`. The static route registry is itself the allowlist; hostname shape
+   (Kubernetes DNS, docker service name, `host.docker.internal`, IP literal) is
+   not inspected. The presence of `http://` in a route is the operator's
+   explicit authorization.
+2. Let the generic `approval_card` ingress render a bounded list of server-built
+   `Action.Submit` buttons instead of always forcing localized `approve` /
+   `deny`.
+
+The callback remains config-driven and the card remains template-driven. Neither
+change permits request-supplied callback URLs or arbitrary Adaptive Card JSON.
+The contract is consumer-neutral: no owner is privileged, and examples must not
+become requirements that future services have to copy.
+
+Bundled configuration collapse (out of the discussion draft):
+
+3. Delete `OCTO_CARD_ACTION_ALLOWED_URLS`. `OCTO_CARD_ACTION_ROUTES[].url` is
+   already the exact URL and shares the same ConfigMap/operator boundary; the
+   two lists cannot form an independent security review and only produce
+   dual-write drift. If the deprecated env is set at startup, log a single
+   structured WARN and ignore it (do not fail startup — rolling upgrades may
+   still carry the old variable).
+
+## Implementation gate
+
+Update and review the standalone operations and consumer integration documents
+first. Do not change production code or tests for HTTP/actions/config collapse
+until the consumer-neutral contract, examples, security boundaries, and rollout
+steps are confirmed.
+
+## Locked decisions (from the discussion draft)
+
+| # | Question | Decision |
+|---|---|---|
+| 1 | actions count | Optional; when provided, 1-5 items. Absent = current localized approve/deny, byte-compatible with #588. |
+| 2 | terminal result states | Reuse `approved` / `denied` / `cancelled` / `pending`. Consumer maps its custom `decision` to a state. |
+| 3 | non-terminal "still clickable" actions | Not in this change. Current finalizers always remove actions and update the card. |
+| 4 | button i18n | Caller-provided single copy per request. Per-recipient language rendering is a future v2. |
+| 5 | requester notification for custom decisions | Only `approved` / `denied` notify the requester (unchanged from #588). Custom `decision` never triggers direct requester notification; the mapped `state` does. |
+
+## Contract
+
+`approval_card.actions` is optional. When omitted, the existing localized
+approve/deny actions are emitted unchanged. When present it contains 1-5 items:
+
+```json
+{
+  "decision": "publish",
+  "title": "发布"
+}
+```
+
+- `decision` is the stable callback value and must be unique within the card,
+  lowercase, and match `[a-z][a-z0-9_.-]{0,47}`.
+- `title` is display text, trimmed, non-empty, and at most 80 Unicode code
+  points. Control characters are rejected.
+- octo-server derives the action ID as `approval-<decision>`, injects the
+  authoritative `owner` and `action_type`, and copies only the existing bounded
+  shared `data` map. Per-action data, styles, URLs, inputs, and caller-authored
+  card JSON remain unsupported.
+- The callback receives the selected `decision` through the existing signed
+  `DecisionRequest`. The existing typed result states (`pending`, `approved`,
+  `denied`, `cancelled`) do not change.
+
+Example:
+
+```json
+{
+  "service": "tasks",
+  "approval_card": {
+    "action_type": "task.execute.decision",
+    "title": "执行任务",
+    "description": "请选择本次任务的处理方式",
+    "data": {"task_id": "task-1"},
+    "actions": [
+      {"decision": "execute", "title": "执行"},
+      {"decision": "reject", "title": "拒绝"},
+      {"decision": "cancel", "title": "取消"}
+    ]
+  }
+}
+```
+
+## Security boundaries
+
+- The static `OCTO_CARD_ACTION_ROUTES` is the exact URL registry. Card content,
+  action requests, and consumer responses cannot supply or override the callback
+  URL. HMAC signing, timeout, retry/DLQ, disabled redirects, and disabled
+  environment proxies are unchanged.
+- Both `http://` and `https://` are accepted; other schemes and non-absolute
+  URLs, URLs with credentials, query strings, fragments, or opaque forms remain
+  rejected at startup.
+- Hostname form is not inspected. Operators are responsible for making the
+  chosen HTTP destination reachable only from the octo-server pods that need
+  it (NetworkPolicy, service mesh, container network, or localhost binding).
+- HMAC only authenticates the request; it does not encrypt payload or response.
+  Callback bodies may contain operator UID, Space ID, and business identifiers.
+  When a route uses `http://`, the deployment must treat the transport as a
+  trusted network boundary. Cross-cluster, cross-network, or public callbacks
+  must use HTTPS.
+- Callback secrets and notify tokens remain per-capability. A notify token
+  cannot equal any callback secret (existing per-route + cross-route check);
+  legacy `NOTIFY_INTERNAL_TOKEN` and `OCTO_DOCS_NOTIFY_TOKEN` also cannot
+  collide with action notify tokens (existing exclusion check).
+- Custom actions are structural fields only. Callers cannot choose owner,
+  callback URL, action JSON, or callback headers.
+
+## Out of scope
+
+- A global "allow insecure HTTP" switch or Kubernetes/Docker environment
+  detection.
+- Per-recipient language rendering of caller-supplied button titles.
+- New callback result states, custom terminal-card layouts, or non-terminal
+  "still clickable" actions.
+- Per-action arbitrary data, styles, URLs, inputs, or card JSON.
+- TLS issuance, service-mesh setup, or NetworkPolicy management.
+- Independent signing of callback responses (known residual risk; tracked
+  separately).
+
+## Acceptance
+
+- Existing HTTPS routes and approve/deny cards are byte-compatible with #588.
+- `OCTO_CARD_ACTION_ROUTES[].url` accepts `http://` and delivers with the same
+  HMAC, retry, DLQ, and redirect/proxy discipline as `https://`.
+- `OCTO_CARD_ACTION_ALLOWED_URLS`, when set at startup, produces a single
+  structured deprecation WARN and does not affect routing decisions.
+- Absolute URL requirements (scheme in {http,https}, host present, no
+  credentials/query/fragment/opaque) continue to reject malformed values at
+  startup.
+- Custom `approval_card.actions` renders the requested labels and signed
+  callback decisions; empty, duplicate, malformed, or oversized actions are
+  rejected before card production.
+- Unit tests cover URL and template boundaries; notify integration tests cover
+  backward compatibility (omitted `actions` byte-equal to #588) and 1-5 custom
+  action rendering.
+- The `card-action-callback-consumer.md` and `card-action-callback-dispatch.md`
+  documents are updated by reference rather than duplicating #588 design, and
+  they carry the residual-risk callout for plain HTTP.
+- A new first-party consumer, regardless of implementation language or owner,
+  can complete route registration, card production, signature verification,
+  idempotent decision handling, typed response, rollout, and DLQ operations
+  from the two standalone integration documents. Consumer-specific examples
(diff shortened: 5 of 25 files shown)


OctoBoooot (dismissed):
## Review: allow HTTP callbacks + bounded custom approval_card actions (#595)

**Verdict: Approve** — a careful, well-tested extension of the #588 cardaction platform. Two security-relevant changes, both byte-verified sound: the HTTP-callback loosening is operator-authorized (not attacker-reachable SSRF), and the custom-actions surface is server-owned + injection-safe. No blockers, no majors. Conditional on Test/Vet/Lint finishing green (Build/Card-Dispatch/osv/i18n/check-sprint already green).

Risk: touches the callback SSRF policy + a new caller-influenced card surface, so I verified both against the #588 baseline I reviewed.

### HTTP-callback loosening — operator-authorized, not an SSRF regression
The change allows `http://` and deletes `OCTO_CARD_ACTION_ALLOWED_URLS` from routing — removing two of the five controls I byte-verified closed SSRF on #588 (https-only + exact-allowlist). But on this code's threat model that's a defensible policy loosening, not a defect:
- **Route URLs are static operator config only** (`LoadRouteSpecs(os.Getenv("OCTO_CARD_ACTION_ROUTES"))`, main.go) — there is no attacker-reachable path to inject/influence a callback URL (a caller request supplies only `action_type` + bounded data → resolved to a *pre-registered* route; never the URL). So the classic SSRF threat (attacker controls the target) doesn't apply; the operator authorizes each destination, and `http://` in a static route is their explicit choice (K8s Service DNS / mesh-terminated TLS).
- **The dropped allowlist was a duplicate** of the same operator config (the route URL *is* the allowlist now) — removing redundancy, not a boundary.
- **The controls that actually bound a misconfigured route are retained + verified:** `validateCallbackURL` still rejects credentials, query (incl. bare `?` via `ForceQuery`), fragment (incl. the raw-`#` prefilter that catches url.Parse's empty-fragment stripping — good edge catch), opaque form, and host-less URLs; and **redirect-refusal + `Proxy=nil` are byte-unchanged**, so an operator host can't 302 to an internal target and proxy-env SSRF stays blocked.

### Custom approval_card actions — server-owned + injection-safe
- Caller supplies only `decision` (pattern-bounded `^[a-z][a-z0-9_.-]{0,47}$`, uniqueness-enforced) + a display `Title`. **Server owns** the action IDs (`approval-<decision>`), the `owner`/`action_type`/`decision` payload injection, and the card JSON; reserved keys can't be overridden (rejected), and custom actions can't mix with approve/deny defaults.
- Bounds: 1–5 actions, title ≤80 runes, control-char check on the raw title, `escapeMarkdown` on title/description into the TextBlock. nil→byte-identical legacy output; empty-slice→400 (strict nil-vs-empty). Injection/all-contexts-escape axes covered.
- Byte-compat pinned (`...OmittedActionsIsByteCompatible`), boundaries + end-to-end custom-decision + nil-vs-empty wire tests present.

### Minor / residual-risk ack (non-blocking)
- Allowing plain `http://` means callback payloads are **HMAC-signed but cleartext** over whatever network the operator configures. The PR discloses this in both docs (residual-risk callout) — that's the right call, and I'm noting it explicitly so the posture change is a conscious reviewer-acknowledged decision, not a silent one. Recommend operators keep http routes to trusted in-cluster/mesh networks only.

Both changes are byte-verified safe on this code's threat model; SSRF-relevant controls (redirect/proxy/credential/query/fragment) intact; custom actions fully server-owned. `size/XL`; CI green except Test/Vet/Lint in flight. A human maintainer should still consciously sign off on the http-callback posture change given it's a deliberate transport-security loosening (operator-scoped).

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #595 (octo-server)

Reviewed at head `38c117462c92474b52b698de987c9a650c5da357`. Scope covers the
`pr-base...HEAD` range as seen locally (includes commits #593 and #588 plus
this PR's HEAD commit `38c11746`); findings below target the cardaction
follow-up that is the stated scope of #595.

## Summary

This PR is the follow-up to #588 (generic card-action callback dispatch) and
extends the callback dispatcher along two axes: the URL validator accepts
plain `http://` alongside `https://` for operator-registered routes, and
`approval_card` gains an optional bounded `actions` slice (1..5) that lets a
caller render custom `Action.Submit` buttons on the generic approval card while
the server retains ownership of the action ID, reserved metadata, and card
layout. The dual-write `OCTO_CARD_ACTION_ALLOWED_URLS` env is deleted from
production code paths; `main.go` keeps a single structured WARN so rolling
upgrades surface a leftover.

The change is well-shaped and well-tested. The scheme relaxation is paired with
genuinely tighter URL-shape checks (raw `#` prefilter, `Hostname()` vs `Host`
to reject `http://:8080/x`, `ForceQuery` to catch trailing `?`), the
allowlist is correctly replaced by "route registration = operator
authorization", the nil-vs-non-nil-empty actions distinction is preserved
through the gin binder, and byte-for-byte compatibility with the legacy
approve/deny wire form is locked by a golden-string test. HMAC signing,
redirect refusal, and proxy-env ignoring all carry over intact to plain HTTP.
One P2 and one nit below — neither blocks.

## Verification

Static analysis only at head `38c117462c92474b52b698de987c9a650c5da357`; build

- ✅ **Scheme relaxation vs SSRF** — `validateCallbackURL` at `internal/cardactiondispatch/registry.go:327` now accepts `http`|`https` but still rejects credentials, fragments (raw prefilter at `:333` *and* `u.Fragment` at `:342`), queries (including `ForceQuery` at `:347`), opaque URLs, empty hostname via `u.Hostname() == ""`, and whitespace-padded input. `TestValidateCallbackURLRejectsUnsafeShapes` in `internal/cardactiondispatch/registry_http_scheme_test.go` covers file/gopher/relative/opaque/credentials/query/trailing-?/fragment/trailing-#/embedded-#/missing-host/port-only-no-host. The registered-routes-are-the-allowlist model is documented in the PR body and the code comments.
- ✅ **Redirects still refused on plain HTTP** — `NewHTTPDeliverer` at `internal/cardactiondispatch/http.go:56-58` sets `CheckRedirect = http.ErrUseLastResponse` regardless of scheme; the response mapper at `:102-104` classifies 3xx as `redirect_rejected` with `retryable: false`. `TestHTTPDelivererRefusesRedirectsOnPlainHTTP` locks this.
- ✅ **Proxy-env SSRF re-open still closed** — `base.Proxy = nil` at `internal/cardactiondispatch/http.go:53`; `TestHTTPDelivererDeliversPlainHTTPWithHMACAndDisciplinedTransport` sets `HTTP_PROXY`/`http_proxy` to a black hole and still delivers with verified HMAC, proving the production transport wins.
- ✅ **HMAC signing path is scheme-invariant** — `Sign` at `internal/cardactiondispatch/http.go:92` consumes `parsed.EscapedPath()` derived from `route.URL`; scheme does not enter the signature input, so plain-HTTP routes sign identically to HTTPS. The e2e test verifies `signatureOK` over cleartext.
- ✅ **Legacy wire-form byte-compat** — `TestBuildApprovalRequestCardOmittedActionsIsByteCompatible` in `pkg/cardtmpl/approval_request_actions_test.go:99` pins the exact pre-http-actions JSON; any ordering/key/escaping drift would fail. `buildApprovalActions` at `pkg/cardtmpl/approval_request.go:113` enters the legacy branch only on `input.Actions == nil`, which is the Go-level representation of JSON field absence.
- ✅ **nil vs non-nil-empty actions survives the binder** — `ApprovalCardFields.Actions` at `modules/notify/model.go:33` is `json:"actions,omitempty"`; `omitempty` only affects marshaling, so gin's unmarshaler preserves absent-field (`nil`) vs `"actions": []` (non-nil empty). `TestIntegration_ApprovalCardCustomActionsRenderInOrder` at `modules/notify/card_action_test.go:342-360` exercises both shapes against the live router and asserts `200 OK` for absent and `400 Bad Request` for explicit empty.
- ✅ **Custom-decision round-trip** — `TestCardActionCallbackOrchestrationForwardsCustomDecisionEndToEnd` at `internal/cardactiondispatch/orchestration_integration_test.go:372` pushes `decision:"execute"` end-to-end: HMAC verifies, `DecisionRequest.Decision` reaches the consumer unmodified (no server-side rewrite to approve/deny), `ActionID` is derived `approval-<decision>`, reserved `data.decision` mirrors the top-level token, and the standard finalizer renders the *consumer-picked* state rather than a decision-derived one.
- ✅ **Deprecation WARN path** — `main.go:477-482` reads `OCTO_CARD_ACTION_ALLOWED_URLS`, emits `log.Warn` with `zap.String("deprecated_env", …)` only when the value is non-empty after trim, and `NewRegistry` is then called with the two-arg `(specs, os.Getenv)` signature. `LoadAllowedURLs` is deleted from `internal/cardactiondispatch/config.go` and every call site. `TestDeprecatedAllowedURLsEnvIsIgnoredWithStructuredWarn` locks the shape.

## Findings

No P0/P1 issues. Two non-blocking items below.

### P2 — Custom-decision pattern does not reserve legacy `approve`/`deny` tokens (`pkg/cardtmpl/approval_request.go:143`)

`approvalDecisionPattern` (`^[a-z][a-z0-9_.-]{0,47}$`) accepts `"approve"`
and `"deny"` as valid custom decisions. The custom branch at `:153-170` then
emits Action.Submit buttons with `id: "approval-approve"` / `"approval-deny"`
— byte-identical to the legacy template IDs declared at `:17-18`. A caller
who accidentally ships `decision:"approve"` on a *custom* `ActionType`
(e.g. `task.execute.decision`) produces buttons whose IDs collide with the
legacy approve/deny IDs in any client-side analytics, replay log, or router
code that keys off action ID alone. The data map still carries the correct
`action_type` and `decision` so the consumer callback is unambiguous, but
ID-only readers silently drift.

Two reasonable fixes: (a) reject `decision ∈ {approve, deny}` with a clear
error (`"cardtmpl: approval action decision collides with legacy approve/deny
namespace"`), or (b) document the reserved namespace in the GoDoc on
`ApprovalRequestAction` and in `docs/card-action-callback-consumer.md` so
callers know to avoid those tokens. Not a blocker — this is caller-footgun
territory, not a reachable server defect — but the wire contract would
benefit from an explicit reservation.

### Nit — Deprecation test pins substrings rather than behavior (`main_carddispatch_test.go:175`)

`TestDeprecatedAllowedURLsEnvIsIgnoredWithStructuredWarn` locks the
deprecation decision by `strings.Contains` against `main.go` source: the
exact `log.Warn` message prefix, the `zap.String("deprecated_env", …)`
structured field, and the `cardactiondispatch.NewRegistry(specs, os.Getenv)`
call shape. That catches the obvious regressions, but rewrapping the Warn
message, renaming the field, or moving the call into a helper silently
disables the guard without any real assertion that the WARN actually fires
and the env is ignored at runtime. A follow-up could swap this for a
capture-logger assertion — call `installCardActionDispatch` with
`OCTO_CARD_ACTION_ALLOWED_URLS` set, assert one WARN record carrying
`deprecated_env=OCTO_CARD_ACTION_ALLOWED_URLS` and no registry error. Low
priority — the current shape does match the intent.

- URL validator: `strings.TrimSpace(raw) != raw` at `registry.go:327` rejects leading/trailing whitespace before `url.Parse`; combined with the raw `#` prefilter and `ForceQuery` check, the three documented footguns (trailing `#` that `url.Parse` silently drops, trailing `?` with empty query, `http://:8080/x` that `u.Host` alone would accept) are all closed.
- Custom actions bounded 1..5 at `pkg/cardtmpl/approval_request.go:137-141`; `MaxApprovalCustomActions` exported at `:20`.
- Control-character rejection runs on the *raw* title before `TrimSpace` strips `\t`/`\n`/`\r` at `:161-165`; `containsControl` at `:179` uses `unicode.IsControl`. Test matrix in `pkg/cardtmpl/approval_request_actions_test.go:153-165` covers `\x07`, leading tab, trailing `\n`, trailing `\r`, and embedded `\v`.
- Mixing custom `Actions` with `ApproveTitle`/`DenyTitle` is rejected at `:133`.
- Duplicate-decision check at `:147-150`; test at `:141-144`.
- `buildApprovalRequestCard` still marshals through `json.Marshal(document)` at `:106` — no hand-rolled JSON, no escaping risk.
- HMAC signing at `http.go:92` consumes `parsed.EscapedPath()`; path-not-empty fallback at `:89` defaults to `/`.
- `ApprovalCardFields.Actions` uses `json:"actions,omitempty"` at `model.go:33` so absent-field is `nil` on decode, and explicit `[]` is non-nil-empty — the distinction the validator relies on.
- `LoadAllowedURLs` removed from `config.go` and every test call site (`contract_test.go`, `dispatcher_test.go`, `service_test.go`, `coverage_test.go`, `card_action_test.go`, `api_card_action_test.go`, `orchestration_integration_test.go`).

## Data-flow trace

- `ApprovalCardFields.Actions` (`modules/notify/model.go:33`) → `buildApprovalRequestCard` in `modules/notify/approval_card.go:104` splits on nil: legacy branch at `:106-109` populates `ApproveTitle`/`DenyTitle` from `approvalActionLabelsFor(lang)`; custom branch at `:111-117` copies `[]ApprovalCardAction` into `[]cardtmpl.ApprovalRequestAction`. → `cardtmpl.BuildApprovalRequestCard` at `pkg/cardtmpl/approval_request.go:43` calls `buildApprovalActions` at `:91` which returns either the legacy two-button slice or the custom N-button slice, validated and with `owner`/`action_type`/`decision` injected into each button's data map. → `json.Marshal(document)` at `:106` produces the on-the-wire bytes. Custom `decision` reaches the dispatcher via the card's action data and is carried unchanged through `Enqueue` → Redis queue → `Dispatcher.ProcessOne` → `HTTPDeliverer.Deliver` → `DecisionRequest.Decision` in the signed callback body (locked by the e2e test).

## Blind-spot checklist

- **C1 (dual-path parity)** — `approve/deny` legacy path and custom-actions path both go through the same `BuildApprovalRequestCard` entry; `buildApprovalActions` is the single branch point and both sides validate, marshal, and return through the same envelope. ✅ parity holds.
- **C2 (control-flow ordering / nested reuse)** — `buildApprovalActions` is called exactly once from `BuildApprovalRequestCard`; no nested reuse that could double-inject `owner`/`action_type`. ✅ clear.
- **C3 (authorization boundary)** — scheme relaxation + allowlist removal changes what URLs the registry accepts. Authorization is now "route registration = operator authorization" (documented in code comments at `registry.go` and `http.go:51-53`). The `SenderUID` pattern check at `registry.go:282` (rejecting `iwh_*` prefixes) and per-route secret lookup are unchanged. ✅ no auth regression.
- **C4 (authorization lifecycle)** — unchanged; route specs come from operator-controlled `OCTO_CARD_ACTION_ROUTES` JSON, no new lifecycle path.
- **C5 (build ≠ runtime)** — the `http://` scheme relaxation and proxy-env SSRF guard are verified by an actual plain-HTTP test server with `HTTP_PROXY` set (`registry_http_scheme_test.go:97-153`); not a build-only claim. ✅
- **C6 (governance/docs self-consistency)** — docs updated in `docs/card-action-callback-consumer.md`, `docs/card-action-callback-dispatch.md`, `docs/card-protocol.md`. The `http://` relaxation is reflected; the allowlist removal is reflected; bounded custom actions are documented. No contradiction with SECURITY.md visible in the diff.

## Cross-round blocker re-check

N/A — first review pass on this PR.

## Verdict: COMMENT

Solid follow-up to #588. Scheme relaxation is paired with genuinely tighter
URL-shape defenses, the bounded custom-actions surface is well-bounded
(1..5, pattern-checked decision, raw-string control-char rejection, reserved
metadata injection), byte-compat with the legacy wire form is locked, and the
end-to-end custom-decision path is exercised. No P0/P1 issues. The P2
(consider reserving `approve`/`deny` in the custom-decision pattern) and the
nit (deprecation-test brittleness) are non-blocking.

[Octo-Q] verdict: APPROVE — zero P0/P1 findings; only P2 (reserved-token footgun) and a nit (test brittleness). Safe to land as-is; P2 can be addressed in a small follow-up.

yujiawei (dismissed):
# Code Review — PR #595 (octo-server)

Independent review at head SHA `38c117462c92474b52b698de987c9a650c5da357` against merge-base `94c596d2`. Reviewed the full diff plus surrounding source (data-flow traced end to end), the linked spec brief, and both callback-dispatch docs. `go build ./...` and `go vet ./...` are clean; the DB-free unit suites (`pkg/cardtmpl`, `internal/cardactiondispatch` URL/scheme/redirect/deprecation) all pass locally.

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly what the linked brief specifies — nothing missing, nothing beyond scope, no deviation.

- **HTTP callbacks** — `validateCallbackURL` now accepts `http` and `https`; every other reject rule (absolute-only, host present, no credentials/query/fragment/opaque) is preserved. Verified against ~23 URL shapes including `http://:8080/path` (host-less, rejected), trailing `?`/`#`, embedded `#`, IPv6 literals, and percent-encoded CRLF (stays encoded through `EscapedPath()`, no header injection). Matches the "registry is the allowlist, hostname shape not inspected" decision.
- **Bounded custom actions** — `approval_card.actions` is optional, 1–5 items, server-owned action IDs (`approval-<decision>`), server-injected `owner`/`action_type`, `decision` matching `^[a-z][a-z0-9_.-]{0,47}$`, title ≤80 runes with control-char rejection on the raw string before trim, and duplicate-decision rejection. Callers cannot supply per-action data, style, URL, input, or card JSON. Matches the contract table.
- **Config collapse** — `OCTO_CARD_ACTION_ALLOWED_URLS` removed from routing; `LoadAllowedURLs` deleted with no dangling references; `NewRegistry` reduced to the two-arg form and all call sites updated; startup emits a single structured deprecation WARN gated on the env being present, and does not fail rolling upgrades.
- **Byte-compat** — the `Actions == nil` path is provably byte-identical to the prior release: the golden test pins the exact JSON, and the `nil`-vs-`empty` wire distinction holds (`{}`/`null` → legacy path; explicit `[]` → 400). Requester-notification and the terminal finalizer remain keyed on `state`, not the custom `decision` (finalizer/dispatcher untouched by this PR).

## 2. Code quality

**Quality: Approved**

No P0/P1 issues. The change is well-scoped, the trust boundary is intact (the clicked `decision`/`owner`/`action_type` are all server-authoritative — extracted from the stored card frame via `cardmsg.SubmitAction` keyed on `action_id`, never taken from the request body), and the test suite is genuinely assertive rather than tautological. The items below are advisory only and do not block merge.

**P2 — Callback *response* is not authenticated; over plain HTTP an on-path attacker can forge the decision result** (`internal/cardactiondispatch/http.go:100-124`)
The outbound request is HMAC-signed, but the response consumed by `DecodeDecisionResult` is only bounds/enum-validated — not origin-authenticated. Its `state`/`display`/`requester_uid` flow into the finalizer that rewrites the card end-users see, so on a plain-HTTP route an on-path attacker could return a forged `200 {state: approved}`. This is a real trust-boundary expansion introduced by allowing `http://`, but it is **working-as-designed and explicitly accepted**: the spec lists independent response signing as a tracked non-goal, and both docs carry the operator-facing residual-risk callout ("HMAC authenticates the request but does not encrypt it, and callback responses are not separately signed — prefer HTTPS for anything crossing a cluster, network zone, or public boundary"). The risk containment rests entirely on operators confining `http://` routes to trusted pod-local traffic (NetworkPolicy/mesh). No action required for merge; flagging so the accepted-risk framing is a conscious call by reviewers, not an oversight.

**P2 — No test stitches `BuildApprovalRequestCard` output to `cardmsg.SubmitAction`** (`pkg/cardtmpl/approval_request_actions_test.go`)
The builder side (emits `id = "approval-<decision>"` with `data.decision`) and the extraction side (`SubmitAction` pulls authored data by `action_id`) are each tested, but with hand-copied constants — no test renders a custom-action card and then feeds it back through `SubmitAction("approval-execute")`. The shapes align today (builder emits top-level `actions`; `SubmitAction` traverses `card["actions"]` matching `type/id/data`), so this is a **coverage gap, not a live bug**. A cheap DB-free round-trip test (marshal builder output → `SubmitAction` → assert `data["decision"]`) would guard against a future divergence in id/data shape or traversal that would pass every new test in this PR yet break the real click path. The custom-decision E2E covers the delivery/signing/decode/finalizer legs well; only the render→extract seam is un-stitched.

**P2 — Response transport hardening (pre-existing, not introduced here)** (`internal/cardactiondispatch/http.go:48-56`)
`Proxy = nil` is only applied when `NewHTTPDeliverer` is called with a `nil` transport. The production caller (`main.go:533`) passes `nil`, so the SSRF-via-`HTTP_PROXY` protection is active today. If a custom `RoundTripper` is ever injected (tracing, metrics, custom TLS), the proxy-env protection would be silently bypassed. This logic is unchanged by #595 (the PR only edited the comment), so it is not a regression — but since allowing plain HTTP raises the stakes of an accidental proxy path, consider unconditionally clearing `Proxy` (or asserting it when the transport is a `*http.Transport`) as a follow-up.

## 3. Verdict

**APPROVE** — Spec ✅ and Quality Approved. All findings are P2/advisory; none require a change before merge.

## 4. Suggestions (non-blocking, optional follow-ups)

- Add the render→extract round-trip unit test described above; it's low-cost and closes the one real coverage seam in the trust boundary.
- Consider fail-closed `Proxy` clearing in `NewHTTPDeliverer` regardless of injected transport.
- If a future consumer ever runs a plain-HTTP route across anything less trusted than pod-local traffic, prioritize the tracked "independent callback-response signing" work item.

## 5. Review method note

This review was produced by absorbing findings from multiple independent analysis passes (a long-context whole-repo pass and a per-dimension adversarial-verification pass across security, correctness, contract/byte-compat, cross-file/build, and test-validity), with each candidate finding independently re-checked against source before inclusion. One additional cross-family analysis leg was attempted but failed to run due to an environment/gateway error on this run; per policy its absence is treated as *not run*, not as silent agreement — the verdict above rests on the passes that did complete plus direct source verification.

Jerry-Xin (dismissed):
Relevant octo-server card-action dispatch change; the implementation is scoped, covered, and consistent with the existing callback and notify architecture.

💬 Non-blocking

🟡 Warning: `actions: null` still falls through to the legacy approve/deny path. `ApprovalCardFields.Actions` is a plain slice, so JSON `null` and an omitted field both decode as `nil`; [modules/notify/approval_card.go](/Users/mlamp/.openclaw/tmp/pr-review-1jpscr1q/repo/modules/notify/approval_card.go:102) treats that as omission. The new raw JSON test covers `"actions": []` vs omitted, but not `"actions": null`. If the intended wire contract is “present means 1–5 actions,” this needs a presence-aware decoder or explicit test/documentation for null-as-omitted.

✅ Highlights

- URL validation is tightened while allowing `http://`: raw `#`, bare query, credentials, opaque URLs, and hostless port-only URLs are rejected in [internal/cardactiondispatch/registry.go](/Users/mlamp/.openclaw/tmp/pr-review-1jpscr1q/repo/internal/cardactiondispatch/registry.go:325).
- The custom action path keeps server-owned action IDs and reserved metadata, with bounded decisions/titles and duplicate detection in [pkg/cardtmpl/approval_request.go](/Users/mlamp/.openclaw/tmp/pr-review-1jpscr1q/repo/pkg/cardtmpl/approval_request.go:116).
- The HTTP deliverer keeps redirect refusal, HMAC signing, timeouts, and proxy-env ignoring for both schemes.
- I ran targeted tests: `go test ./internal/cardactiondispatch ./pkg/cardtmpl ./modules/notify` passed.

Jerry-Xin (dismissed):
Relevant octo-server card-action dispatch change; the implementation is scoped, covered, and consistent with the existing callback and notify architecture.

💬 Non-blocking

🟡 Warning: `actions: null` still falls through to the legacy approve/deny path. `ApprovalCardFields.Actions` is a plain slice, so JSON `null` and an omitted field both decode as `nil`; [modules/notify/approval_card.go](modules/notify/approval_card.go:102) treats that as omission. The new raw JSON test covers `"actions": []` vs omitted, but not `"actions": null`. If the intended wire contract is “present means 1–5 actions,” this needs a presence-aware decoder or explicit test/documentation for null-as-omitted.

✅ Highlights

- URL validation is tightened while allowing `http://`: raw `#`, bare query, credentials, opaque URLs, and hostless port-only URLs are rejected in [internal/cardactiondispatch/registry.go](internal/cardactiondispatch/registry.go:325).
- The custom action path keeps server-owned action IDs and reserved metadata, with bounded decisions/titles and duplicate detection in [pkg/cardtmpl/approval_request.go](pkg/cardtmpl/approval_request.go:116).
- The HTTP deliverer keeps redirect refusal, HMAC signing, timeouts, and proxy-env ignoring for both schemes.
- I ran targeted tests: `go test ./internal/cardactiondispatch ./pkg/cardtmpl ./modules/notify` passed.

an9xyz:
## Fixup for review comments (commit a5ee28af)

Addressed the three actionable P2/non-blocking items; the remaining ones are treated as out of scope for this PR (see per-item notes below).

### Applied

1. **Reserve `approve`/`deny` in the custom-decision namespace** — Octo-Q's P2. `pkg/cardtmpl.buildApprovalActions` now rejects `decision ∈ {approve, deny}` so the derived `approval-<decision>` IDs cannot collide with the legacy template. Added 3 boundary test cases (single reserved, other reserved, mixed with a valid custom) and a GoDoc note on `ApprovalRequestAction` documenting the reserved namespace.

2. **Document `"actions": null` equivalence** — Jerry-Xin's warning. Both docs (`card-action-callback-dispatch.md`, `card-action-callback-consumer.md`) now spell out that omit and JSON `null` are wire-equivalent (both decode to nil and take the legacy approve/deny template); explicit `[]` still 400s. Pinned by a new raw-JSON assertion inside `TestIntegration_ApprovalCardCustomActionsRenderInOrder` so future decoder changes cannot silently break the contract.

3. **Builder → `SubmitAction` round-trip** — yujiawei's P2. New `TestBuildApprovalRequestCardRoundTripsThroughSubmitAction` renders both custom-action and legacy approve/deny cards, wraps them in the octo/v2 envelope the runtime sees, extracts each button by its server-derived action ID via `cardmsg.SubmitAction`, and asserts the decision / reserved metadata / shared data survive the round-trip. Closes the render→extract seam.

### Deferred (with explicit reasons)

- **Deprecation WARN test brittleness (Octo-Q, nit)** — the current source-substring test does catch the intended shape regressions; swapping it for a `zap` observer would pull in a helper refactor and doesn't fit this PR's scope. Tracked for a later cleanup.
- **`NewHTTPDeliverer` unconditional `Proxy=nil` (yujiawei, P2)** — this is pre-existing code from #588. Production caller passes `nil` so the SSRF guard is active today; hardening it to also cover future injected transports is worth doing but should not scope-creep into this PR.
- **Independently-signed callback responses (OctoBoooot / yujiawei)** — brief already lists this as an explicit non-goal with an operator-facing residual-risk callout in both docs. Not addressing here.
- **Human sign-off on http:// posture (OctoBoooot)** — governance ask, no code change.

### Gate

- `go build`, `go vet`, `gofmt`, `make i18n-lint`: clean.
- `-race` unit + main package: PASS.
- Integration on fresh local `test` schema (`internal/cardactiondispatch` including 3 orchestration E2Es): PASS.
- Coverage: cardactiondispatch 81.4%, cardtmpl **90.1%** (up from 89.9%), notify 71.2%.

OctoBoooot (approved):
## Delta re-review: #595 @ a5ee28af — additive hardening (reserved approve/deny decision guard). APPROVE.

Real commit (ahead 1 / behind 0, not a rebase) touching 3 files: `approval_request.go` + its two tests. Byte-verified — it's a clean, additive hardening, no change to the surfaces I verified at 38c11746. Re-affirming **APPROVE**, conditional on Test.

### The delta: reject custom `decision` = "approve"/"deny" (ID-namespace collision guard)
`approval_request.go` adds `reservedApprovalDecisions{approve, deny}` and rejects a custom action whose `decision` is either — because those would produce action IDs `approval-approve`/`approval-deny` that collide with the legacy template's `ApprovalApproveActionID`/`ApprovalDenyActionID` (breaking any client analytics/replay keyed on action_id). Sound defensive call: fail-closed rejection, keeps the server-derived ID namespace clean between custom and legacy actions.

- Pinned by three new cases in `approval_request_actions_test.go`: `decision:"approve"`→reject, `decision:"deny"`→reject, and `approve` mixed with other valid decisions→reject. Mutation-resistant.
- No change to what I verified before: decision pattern-bound + unique, count 1..5, title control-char-checked + `escapeMarkdown` + rune-bound, server-owned IDs/reserved-keys, nil→byte-compat legacy. SSRF surface (operator-static routes, redirect-refusal, Proxy=nil, validateCallbackURL rejections) untouched.

Strictly better than the version I approved — one additive fail-closed guard with pinning tests, nothing else changed, security/injection/SSRF surfaces intact. CI: Build green, Test/Vet/Lint running; conditional on Test. The residual-risk ack from my prior review stands (http callbacks = HMAC-signed cleartext, operator-scoped, doc-disclosed). No blockers, no majors.

(Method note: ahead-1 real commit, not rebase — reviewed the delta fresh; blob-compare confirmed only these 3 files changed.)

Jerry-Xin (dismissed):
The PR is relevant to octo-server and cleanly extends the existing card action dispatch and generic approval-card contract.

🔴 Blocking  
None.

💬 Non-blocking  
🔵 Suggestion: [main_carddispatch_test.go:170](/Users/mlamp/.openclaw/tmp/pr-review-c6fka3wb/repo/main_carddispatch_test.go:170) uses source-string assertions to pin startup behavior. It is acceptable here, but a behavioral test with a captured logger would be less brittle if this area changes again.

🔵 Suggestion: [registry_http_scheme_test.go:177](/Users/mlamp/.openclaw/tmp/pr-review-c6fka3wb/repo/internal/cardactiondispatch/registry_http_scheme_test.go:177) could use `errors.As` directly instead of a local type assertion helper, so the test remains valid if delivery errors are wrapped later.

✅ Highlights  
The URL validator keeps the intended boundary tight while allowing `http://`: raw fragment rejection, `ForceQuery`, credentials, opaque form, and `Hostname()` checks are all covered in [registry.go:325](/Users/mlamp/.openclaw/tmp/pr-review-c6fka3wb/repo/internal/cardactiondispatch/registry.go:325).

The custom action path preserves the nil-vs-empty wire contract and keeps card JSON server-owned in [approval_request.go:131](/Users/mlamp/.openclaw/tmp/pr-review-c6fka3wb/repo/pkg/cardtmpl/approval_request.go:131) and [approval_card.go:91](/Users/mlamp/.openclaw/tmp/pr-review-c6fka3wb/repo/modules/notify/approval_card.go:91).

The HTTP deliverer still disables proxy environment use and redirects for both schemes in [http.go:48](/Users/mlamp/.openclaw/tmp/pr-review-c6fka3wb/repo/internal/cardactiondispatch/http.go:48).

Verification: focused tests passed for `./internal/cardactiondispatch`, `./pkg/cardtmpl`, and `./modules/notify`. A full `go test ./...` was attempted but failed on existing local integration DB/schema migration issues unrelated to this PR.

Jerry-Xin (approved):
The PR is relevant to octo-server and cleanly extends the existing card action dispatch and generic approval-card contract.

🔴 Blocking
None.

💬 Non-blocking
🔵 Suggestion: `main_carddispatch_test.go:170` uses source-string assertions to pin startup behavior. It is acceptable here, but a behavioral test with a captured logger would be less brittle if this area changes again.

🔵 Suggestion: `internal/cardactiondispatch/registry_http_scheme_test.go:177` could use `errors.As` directly instead of a local type assertion helper, so the test remains valid if delivery errors are wrapped later.

✅ Highlights
The URL validator keeps the intended boundary tight while allowing `http://`: raw fragment rejection, `ForceQuery`, credentials, opaque form, and `Hostname()` checks are all covered in `internal/cardactiondispatch/registry.go:325`.

The custom action path preserves the nil-vs-empty wire contract and keeps card JSON server-owned in `pkg/cardtmpl/approval_request.go:131` and `modules/notify/approval_card.go:91`. The new head adds a reserved-token guard rejecting `approve`/`deny` as custom decisions so the server-derived `approval-<decision>` action IDs stay collision-free with the legacy ApprovalApprove/DenyActionID; nil Actions still selects the legacy 2-button template and a non-nil empty slice now returns an explicit error (no silent fallback).

The HTTP deliverer still disables proxy environment use and redirects for both schemes in `internal/cardactiondispatch/http.go:48`.

Verification: focused tests passed for `./internal/cardactiondispatch`, `./pkg/cardtmpl`, and `./modules/notify`. A full `go test ./...` was attempted but failed on pre-existing local integration DB/schema migration issues unrelated to this PR.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #595 (octo-server)

## Summary

This PR adds bounded custom approval-card actions and HTTP callback support to the card-action dispatch system. The legacy two-button approve/deny template is preserved byte-compatibly when `Actions` is nil; a non-nil `Actions` slice (1–`MaxApprovalCustomActions` entries) drives custom `Action.Submit` buttons with server-owned IDs, reserved metadata (`owner`, `action_type`, `decision`), and collision-free decision tokens. The dual-write URL allowlist (`OCTO_CARD_ACTION_ALLOWED_URLS`) is removed — `OCTO_CARD_ACTION_ROUTES[].url` is now the single source of truth — with a structured deprecation WARN for rolling upgrades. URL validation is hardened: fragments are rejected on the raw string before `url.Parse` strips empty ones, `ForceQuery` catches trailing `?`, and `Hostname()` rejects port-only forms like `http://:8080/path`. The scheme constraint is relaxed from HTTPS-only to `http(s)` for operator-registered internal callback destinations.

The changes are well-structured, the data flows are sound, and the test suite — including round-trip, integration, and end-to-end coverage — is comprehensive.

## Verification

Static analysis only at head `a5ee28af`; build and tests not executed in this environment.

- ✅ **Backward compatibility** — `TestBuildApprovalRequestCardOmittedActionsIsByteCompatible` locks the exact golden bytes from the pre-http-actions release; nil `Actions` preserves the legacy approve/deny wire form.
- ✅ **Nil vs empty semantics** — Integration test at `modules/notify/card_action_test.go:338-381` covers all three wire shapes: omitted field (nil → legacy), `"actions": null` (nil → legacy), `"actions": []` (non-nil empty → rejected as caller bug). The `omitempty` tag on `ApprovalCardFields.Actions` at `modules/notify/model.go:33` preserves the nil/empty distinction through Go's `encoding/json` unmarshal contract.
- ✅ **Reserved decision namespace** — `reservedApprovalDecisions` at `pkg/cardtmpl/approval_request.go:32` prevents custom actions from colliding with `ApprovalApproveActionID` / `ApprovalDenyActionID`; tested at `:179-188` in `approval_request_actions_test.go`.
- ✅ **URL validation hardening** — Raw-string fragment check at `internal/cardactiondispatch/registry.go:337` catches `url#` before `url.Parse` strips the empty fragment. `ForceQuery` at `:349` catches trailing `?`. `Hostname()` at `:343` rejects port-only hosts. All edge cases covered by `registry_http_scheme_test.go:80-120`.
- ✅ **SSRF mitigations for HTTP** — `http.go:51` clones `DefaultTransport` with `Proxy = nil` for both http and https. Redirect refusal is tested for plain HTTP at `registry_http_scheme_test.go:138-170`. HMAC signing verified over cleartext at `:108-136`.
- ✅ **Data-flow: reserved metadata injection** — `buildApprovalActions` at `pkg/cardtmpl/approval_request.go:161-175` sets `owner`, `action_type` on `baseData` *after* caller `Data` injection, so reserved fields always overwrite any caller-supplied keys of the same name. Per-action `decision` is set on the copy, preventing caller override.
- ✅ **Round-trip integrity** — `TestBuildApprovalRequestCardRoundTripsThroughSubmitAction` at `approval_request_actions_test.go:209-298` stitches the builder to `cardmsg.SubmitAction` for both custom and legacy paths, including a synthetic miss.
- ✅ **End-to-end custom decision** — `TestCardActionCallbackOrchestrationForwardsCustomDecisionEndToEnd` at `orchestration_integration_test.go:370-477` verifies the custom decision `execute` reaches the consumer unmodified, HMAC signs correctly, and the standard finalizer renders the consumer-picked state.
- ✅ **Allowlist removal** — All `NewRegistry` call sites updated: `main.go:477-484`, plus all test files. `LoadAllowedURLs` removed from `config.go`. `OCTO_CARD_ACTION_ALLOWED_URLS` deprecated with structured WARN at `main.go:479`. Test `TestDeprecatedAllowedURLsEnvIsIgnoredWithStructuredWarn` at `main_carddispatch_test.go:164-197` locks the deprecation contract.
- ✅ **Mixing guard** — `buildApprovalActions` at `pkg/cardtmpl/approval_request.go:137` rejects `ApproveTitle`/`DenyTitle` when custom `Actions` are provided; tested at `approval_request_actions_test.go:195-204`.
- ✅ **Control character rejection** — `containsControl` at `pkg/cardtmpl/approval_request.go:186` checks the raw title before `TrimSpace`, preventing embedded `\t`, `\n`, `\r`, `\v`, `\x07` from reaching card output. Tested for seven control-character variants at `approval_request_actions_test.go:170-177`.

## Findings

No P0/P1 issues. No P2 issues. The PR is well-engineered with thorough validation, comprehensive test coverage across unit/integration/end-to-end levels, and intentional security hardening.

- **ASYNC/RETRY**: No new async paths introduced; existing dispatch/retry/DLQ machinery is unchanged. The custom decision flows through the same `Enqueue → ProcessOne → Deliver` pipeline.
- **AUTH/VISIBILITY**: Route registration still validates `senderUID`, `owner`, `actionType`, `secretEnv` patterns. Notify token binding is unchanged. Internal token header required at the `/v1/internal/notify` endpoint.
- **BYTE-CAP**: `MaxApprovalActionTitleRunes = 80` matches the pre-existing inline check. `MaxApprovalCustomActions = 5` caps button count. `maxApprovalDataFields = 32` is unchanged. All enforced with explicit validation before marshal.
- **EXTERNAL-BOUNDARY**: Route-spec URLs are the single source of truth; operators control the callback surface via `OCTO_CARD_ACTION_ROUTES`. The deprecation WARN for the old env ensures rolling upgrades surface leftover config.
- **INPUT-SHAPE**: Decision pattern `^[a-z][a-z0-9_.-]{0,47}$` rejects uppercase, spaces, empty, digit-leading, and over-length tokens. Duplicate decisions rejected. Control characters in titles rejected pre-trim. Empty `Actions` slice fails closed.
- **PERSISTENCE/CACHE**: No new persisted state; card rendering is stateless.
- **SOURCE-COVERAGE**: N/A — the allowlist removal simplifies the source from two (route spec + env allowlist) to one (route spec only); the chosen source is a strict superset for the remaining validation chain.
- **C1 (dual-path parity)**: nil `Actions` → legacy path and non-nil `Actions` → custom path are symmetric: both validate inputs, build `Action.Submit` buttons, and inject reserved metadata. Mixing is rejected.
- **C3 (auth boundary)**: No new endpoints exposed. The HTTP scheme relaxation affects operator-configured callback destinations only, not user-facing attack surface.
- **C5 (build/runtime parity)**: No build/packaging changes. The scheme change is validated at both the registry level and the transport level.

## Data-flow trace

| Consumer | Upstream source | Runtime guarantee |
|---|---|---|
| `cardtmpl.Actions` on `ApprovalRequestCard` | `ApprovalCardFields.Actions` from JSON request → `approval_card.go` maps to `[]ApprovalRequestAction` | nil preserved (legacy path); non-nil mapped element-wise |
| `decision` in action `data` | Caller's `ApprovalCardAction.Decision` → validated by `approvalDecisionPattern` + reserved check + dup check → set on `copyActionData(baseData)` | Cannot be empty, reserved, duplicate, or pattern-violating |
| `owner`/`action_type` in action `data` | `capability.Owner` / `card.ActionType` from notify request → set on `baseData` after caller `Data` | Always overwrites any caller-supplied keys of same name |
| Action `id` | `"approval-" + decision` (server-derived) | Unique per decision (dup check); collision-free with legacy IDs (reserved check) |
| Callback URL scheme | `RouteSpec.URL` from `OCTO_CARD_ACTION_ROUTES` env → `validateCallbackURL` accepts http/https | Operator-controlled; transport blocks proxy/redirects for both schemes |

## Verdict: APPROVED

Well-designed feature addition with intentional backward compatibility, comprehensive validation, thorough test coverage at every level, and proper security hardening. The URL validation improvements (fragment, ForceQuery, Hostname) close real edge cases. The HTTP scheme relaxation is a deliberate operator-controlled design choice with appropriate transport-level mitigations. No defects found.

yujiawei (approved):
# Code Review — PR #595 (octo-server)

Reviewed at head SHA `a5ee28af83c629a4847b0f05b257b06e84020692` against merge-base `94c596d2` (#588). Independent re-derivation: diff read in full, `go build ./...` / `go vet` clean, `pkg/cardtmpl` and the no-DB `cardactiondispatch` URL/registry unit tests run locally and PASS.

## 1. Spec compliance

**Spec: ✅**

The change implements exactly the three items in the linked task brief, and nothing beyond them:

1. **Plain HTTP callbacks** — `validateCallbackURL` (`internal/cardactiondispatch/registry.go:331`) now accepts `http` alongside `https`; the static `OCTO_CARD_ACTION_ROUTES[].url` registry remains the single source of authorization. Hostname shape is intentionally not inspected, matching the brief.
2. **Bounded custom `approval_card.actions`** — optional 1–5 item list (`pkg/cardtmpl/approval_request.go:19`, `MaxApprovalCustomActions=5`). Server owns action IDs (`approval-<decision>`), injects `owner`/`action_type`, and copies only the bounded `data` map. Per-action data/style/URL/input and arbitrary card JSON stay unsupported.
3. **`OCTO_CARD_ACTION_ALLOWED_URLS` collapse** — `LoadAllowedURLs` deleted; `main.go:477` reads the env only to emit one structured `deprecated_env` WARN and ignores it. Startup does not fail (rolling-upgrade safe).

- Missing: none. All acceptance criteria are covered by tests (byte-compat golden, URL boundary matrix, custom-decision E2E, nil-vs-empty wire test, deprecation contract).
- Over-build: none. No new flags, states, or endpoints; `DecisionResult` states (`pending/approved/denied/cancelled`) unchanged.
- Divergence: none. `http`/`https` both accepted; credentials/query/fragment/opaque/host-less still rejected at startup.

Docs carry the required plain-HTTP residual-risk callout (`docs/card-action-callback-dispatch.md:101`, `docs/card-action-callback-consumer.md:16`).

## 2. Code quality

**Quality: Approved**

Trust boundary and byte-compat both hold:

- **Decision is not client-trusted.** On click, `modules/message/api_card_action.go:204` re-reads the button `data` from the *stored* effective card via `cardmsg.SubmitAction(effective, req.ActionID)`; the client-submitted decision is never trusted. The custom `decision` flows to the consumer unmodified, and the finalizer keys the terminal state and requester notification off the consumer-returned `result.State`, not the raw decision (`modules/notify/standard_action_finalizer.go:47`, unchanged). `TestCardActionCallbackOrchestrationForwardsCustomDecisionEndToEnd` locks this whole chain.
- **URL validation is strengthened, not weakened.** Beyond relaxing the scheme, the validator adds a raw-`#` prefilter (empty fragments that `url.Parse` drops), a `ForceQuery` check (trailing `?`), and switches to `u.Hostname()` so `http://:8080/...` is rejected as host-less. Reject matrix is well covered.
- **SSRF discipline unchanged.** `base.Proxy = nil` still applies to both schemes (`http.go:50`); redirect ban and HMAC signing are byte-identical and exercised over cleartext by `TestHTTPDelivererDeliversPlainHTTPWithHMACAndDisciplinedTransport`.
- **Custom-actions validation is fail-closed and thorough:** decision pattern `[a-z][a-z0-9_.-]{0,47}`, reserved `approve`/`deny` namespace rejected, duplicates rejected, control chars checked on the *raw* title before `TrimSpace`, 80-rune title cap, mix-with-defaults rejected, non-nil empty slice → error (not silent fallback). nil → byte-identical legacy output, pinned by a golden-bytes test.

### Findings (all non-blocking)

- **P2** — `internal/cardactiondispatch/registry_http_scheme_test.go:178` — `errorsAsDelivery` uses a strict `err.(*DeliveryError)` type assertion rather than `errors.As`. If `Deliver` ever wraps its error with `%w`, this test helper silently stops matching. Prefer `errors.As` for future-proofing. (test-only)
- **P2** — `internal/cardactiondispatch/http.go:50` — `http.DefaultTransport.(*http.Transport).Clone()` panics if any imported package replaces the default transport with a non-`*http.Transport` RoundTripper (e.g. an OTel/APM wrapper). This line is **pre-existing** (unchanged by this PR — only the comment was edited), so it does not block #595, but a guarded assertion with graceful fallback would be a reasonable hardening follow-up.
- **P2 (nit)** — `pkg/cardtmpl/approval_request.go` — legacy approve/deny label validation moved from before the `Data` size/shape check to inside `buildApprovalActions` (after it). When both are invalid a different error message now surfaces first. Purely cosmetic; both paths still fail-closed.

## 3. Verdict

**APPROVED** — Spec ✅ and Quality Approved. No P0/P1 issues; the three findings above are P2/nits and do not block merge.

## Review coverage note

This review combined a full first-party source read with a cross-family long-context advisory pass. The advisory pass could not see `pkg/cardtmpl` internals (outside its diff window); those were verified directly here (source + passing boundary tests). An additional line-level adversarial pass was attempted but did not complete due to an unrelated tooling/gateway error, so it contributed no findings. Potential blind spots consciously checked: server-side decision re-derivation (covered), proxy/redirect discipline on the new http path (covered), and legacy wire byte-compat (golden-locked).

PR #608, opened 2026-07-17
feat(messages_search): support octo-cli message search as bot / as-user (OBO)

### LINKED ISSUE #607: Support octo-cli message search as bot / as-user (OBO)
# Support octo-cli message search as bot / as-user (OBO)

Enable octo-cli to call octo-server message search under two principals:

- **search as bot** — the User Bot's own identity: friend DMs + joined groups/threads, including global search.
- **search as user (OBO)** — on-behalf-of the grantor (currently the creator), fully reusing the existing human authorization path.

Scope: **User Bot only in this phase; App Bot is explicitly denied** (fail-closed at principal construction, never silently allowed).

## Motivation

The existing message-search authorization assumes a human principal resident in a Space (friend/same-Space + bidirectional blacklist; global allowlist = `friend ∪ space members`). A bot has no Space (no `space_id`, no `space_member` row), and its routes do not mount the Space middleware, so as-bot / as-user(OBO) search either hits the wrong human-only edges or is rejected outright by the `RequireSpaceID` gate before reaching the allowlist. This tracks the work to make search principal-aware.

## Acceptance

1. as-bot: friend DM / group / thread / global visibility correct; P2P does not consult blacklist.
2. as-user(OBO): grant + scope + live grantor permission convergence; bidirectional blacklist; TOCTOU consistent with the send path.
3. Rate limit keyed by botUID; audit records botUID + grantorUID.
4. App Bot path explicitly denied.
5. Existing search test suite green + new principal visibility matrix.

BrDing-Rookie:
## Summary

Make octo-server message search **principal-aware** so octo-cli can search under two identities:

- **search as bot** — the User Bot's own identity (friend DMs + joined groups/threads, including global).
- **search as user (OBO)** — on-behalf-of the grantor, fully reusing the existing human authorization path.

**User Bot only in this phase; App Bot is explicitly denied.**

## Changes

- **Principal abstraction** (`user` / `bot` / `obo-grantor` / `uk`) collapsed into a single normalized read predicate — `canReadChannel` / `enumerateReadableChannels` — so single-channel gates and the global allowlist share **one** predicate (no inlined rule copies that can drift).
- **as-bot P2P gate**: friends-only; skips the Space segment and P2P blacklist (a bot has no `space_member` row and cannot blacklist anyone).
- **as-bot group/thread gate**: subject = `botUID` via `ExistMemberActive` (group-level blacklist inherited for free; kicked/blacklisted bot drops out).
- **as-bot global allowlist**: `friends ∪ joined groups ∪ threads`, bounded (a bot's group set ⊆ its creator's group set, so magnitude matches a human's global).
- **as-user (OBO)**: entry grant-existence fail-fast + per-channel scope / TOCTOU convergence reusing the send-side checker; App Bot denied at principal construction.
- **Spaceless bot/OBO global search**: the default `RequireSpaceID` fail-close now only constrains human (`user` / `uk`) subjects; bot/OBO with empty `spaceID` proceed into the same per-principal allowlist predicate. `_search_global_groups` now uses `principal.SubjectUID()`, aligned with messages/files.
- **Route mounting**: bot + uk search routes with principal resolution.

## Testing

- `go build ./...` ✅
- `go vet ./modules/messages_search/` ✅
- `go test ./modules/messages_search/` ✅ (full suite green)
- `go test ./modules/bot_api/` OBO route matrix ✅ (`TestResolveSearchPrincipal_OBO*`, `TestCheckOBO_*`, `TestSearchOBOAllowed_*`, App-Bot-denial cases)

> Note: `bot_api` integration tests that depend on MySQL/Redis/WuKongIM should be run in an infra-provisioned environment; they were not exercised locally.

Fixes #607

### DIFF
--- a/modules/bot_api/bot_api.go
+++ b/modules/bot_api/bot_api.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-server/internal/carddispatch"
 	"github.com/Mininglamp-OSS/octo-server/modules/file"
 	"github.com/Mininglamp-OSS/octo-server/modules/group"
+	"github.com/Mininglamp-OSS/octo-server/modules/messages_search"
 	"github.com/Mininglamp-OSS/octo-server/modules/robot"
 	"github.com/Mininglamp-OSS/octo-server/modules/thread"
 	"github.com/Mininglamp-OSS/octo-server/modules/user"
@@ -137,6 +138,11 @@ type BotAPI struct {
 	// full user-service stack.
 	// nil in production → the real userService.IsFriend runs.
 	friendCheckOverride func(uid, toUID string) (bool, error)
+	// searchHandler 是共享的消息搜索 Handler（messages_search.Shared），bot 搜索路由
+	// /v1/bot/messages/_search* 复用它（YUJ-49 / #B）。与 web、uk 入口共用同一实例，
+	// 从而共享限流桶与 sender 缓存。principal 由本模块的 resolveSearchPrincipal 按
+	// on_behalf_of 区分 as-bot / as-user(OBO)。
+	searchHandler *messages_search.Handler
 	log.Log
 }
 
@@ -197,6 +203,7 @@ func NewBotAPI(ctx *config.Context) *BotAPI {
 		maxVoiceContextLength: maxCtxLen,
 		maxBodySize:           maxBodySize,
 		maxFileSize:           maxFileSize,
+		searchHandler:         messages_search.Shared(ctx),
 		Log:                   log.NewTLog("BotAPI"),
 	}
 	// YUJ-1166 / Mininglamp-OSS/octo-server#81 — Persona Clone fan-out.
@@ -287,6 +294,10 @@ func (ba *BotAPI) Route(r *wkhttp.WKHttp) {
 	// 群入站 Webhook 管理（bot token 面），与用户路由共用 incomingwebhook 实例与
 	// 权限矩阵（管理员 bot 同权人类管理员）。实现在 incoming_webhook.go。
 	ba.registerIncomingWebhookRoutes(r)
+
+	// 消息搜索（YUJ-49 / #B）：/v1/bot/messages/_search*，authBot 鉴权，principal 由
+	// on_behalf_of 区分 as-bot / as-user(OBO)。实现在 search_route.go。
+	ba.mountSearchRoutes(r)
 }
 
 // ==================== Helper Functions ====================
--- a/modules/bot_api/obo_check.go
+++ b/modules/bot_api/obo_check.go
@@ -211,6 +211,34 @@ func (ba *BotAPI) checkOBO(botUID, grantor, channelID string, channelType uint8)
 	return nil
 }
 
+// SearchOBOAllowed adapts checkOBO for the messages_search as-user(OBO) search
+// gate (YUJ-53 / search-as-user). It answers the same per-channel question as
+// the send path — "may grantee bot botUID read (channelID, channelType) as
+// grantor?" — reusing the identical grant + scope + grantorCanReadChannel
+// (TOCTOU) logic, so search authorization can never diverge from send.
+//
+// It exists so messages_search can consume the OBO decision through a narrow
+// (bool, error) seam WITHOUT importing bot_api's internals or the
+// ErrOBONotAuthorized sentinel: the "existence-hidden" outcome collapses to
+// (false, nil) while genuine infra failures propagate for fail-closed
+// handling.
+//
+// Return contract for the search layer:
+//   - (true,  nil) — authorized; search may proceed as grantor.
+//   - (false, nil) — ErrOBONotAuthorized: no active grant / scope disabled /
+//     grantor lost live channel access. The caller hides existence (NOT_FOUND).
+//   - (false, err) — infra error; the caller fails closed (INTERNAL_ERROR).
+func (ba *BotAPI) SearchOBOAllowed(botUID, grantorUID, channelID string, channelType uint8) (bool, error) {
+	err := ba.checkOBO(botUID, grantorUID, channelID, channelType)
+	if err == nil {
+		return true, nil
+	}
+	if errors.Is(err, ErrOBONotAuthorized) {
+		return false, nil
+	}
+	return false, err
+}
+
 // oboStoreOrDefault returns the test-injected oboStore if set, else the
 // production DB-backed one. Mirrors spaceQuerierOrDefault so the test seam
 // is consistent across the module.
--- a/modules/bot_api/obo_search_adapter_test.go
+++ b/modules/bot_api/obo_search_adapter_test.go
@@ -1,0 +1,76 @@
+// Package bot_api · YUJ-53 — Unit tests for the SearchOBOAllowed adapter that
+// exposes checkOBO to the messages_search as-user(OBO) gate. The adapter only
+// remaps checkOBO's error contract into (bool, error); the underlying grant /
+// scope / TOCTOU matrix is exercised by obo_check_test.go.
+package bot_api
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+)
+
+// TestSearchOBOAllowed_Authorized — an active grant + enabled scope + live
+// grantor access maps to (true, nil).
+func TestSearchOBOAllowed_Authorized(t *testing.T) {
+	s := newFakeOBOStore()
+	gid, err := s.insertGrant(tGrantor, tBot, "auto", "")
+	if err != nil {
+		t.Fatalf("insertGrant: %v", err)
+	}
+	enable := 1
+	if err := s.updateGrant(gid, "", &enable, nil); err != nil {
+		t.Fatalf("updateGrant: %v", err)
+	}
+	if _, err := s.insertScope(gid, tChan, common.ChannelTypeGroup.Uint8(), 1); err != nil {
+		t.Fatalf("insertScope: %v", err)
+	}
+
+	ba := newBotAPIWithFakeStore(s)
+	ok, err := ba.SearchOBOAllowed(tBot, tGrantor, tChan, common.ChannelTypeGroup.Uint8())
+	if err != nil {
+		t.Fatalf("authorized: unexpected err %v", err)
+	}
+	if !ok {
+		t.Fatalf("authorized: want ok=true")
+	}
+}
+
+// TestSearchOBOAllowed_NotAuthorizedHidesExistence — ErrOBONotAuthorized
+// (no grant here) collapses to (false, nil) so the search layer hides the
+// channel rather than surfacing a distinguishable error.
+func TestSearchOBOAllowed_NotAuthorizedHidesExistence(t *testing.T) {
+	ba := newBotAPIWithFakeStore(newFakeOBOStore())
+	ok, err := ba.SearchOBOAllowed(tBot, tGrantor, tChan, common.ChannelTypeGroup.Uint8())
+	if err != nil {
+		t.Fatalf("not-authorized must not surface an error, got %v", err)
+	}
+	if ok {
+		t.Fatalf("not-authorized: want ok=false")
+	}
+}
+
+// TestSearchOBOAllowed_InfraErrorPropagates — a genuine infra error (grantor
+// live-access re-check fails) propagates as (false, err) so the caller can
+// fail closed with INTERNAL rather than mistaking it for "not authorized".
+func TestSearchOBOAllowed_InfraErrorPropagates(t *testing.T) {
+	s := newFakeOBOStore()
+	gid, _ := s.insertGrant(tGrantor, tBot, "auto", "")
+	enable := 1
+	_ = s.updateGrant(gid, "", &enable, nil)
+	_, _ = s.insertScope(gid, tChan, common.ChannelTypeGroup.Uint8(), 1)
+
+	ba := newBotAPIWithFakeStore(s)
+	sentinel := errors.New("db down")
+	ba.oboChannelAccessOverride = func(uid, channelID string, channelType uint8) (bool, error) {
+		return false, sentinel
+	}
+	ok, err := ba.SearchOBOAllowed(tBot, tGrantor, tChan, common.ChannelTypeGroup.Uint8())
+	if !errors.Is(err, sentinel) {
+		t.Fatalf("infra error must propagate, got %v", err)
+	}
+	if ok {
+		t.Fatalf("infra error: want ok=false")
+	}
+}
--- a/modules/bot_api/search_route.go
+++ b/modules/bot_api/search_route.go
@@ -40,6 +40,11 @@ type searchOBOFieldProbe struct {
 
 // mountSearchRoutes 挂载 bot 搜索子树。由 Route() 调用。
 func (ba *BotAPI) mountSearchRoutes(r *wkhttp.WKHttp) {
+	// YUJ-53 / #F：把 as-user(OBO) 的 scope 门注入共享搜索 Handler。messages_search 的
+	// obo gate（oboCanReadChannel / oboEnumerateReadableChannels）逐频道经此 checker 复用
+	// 发消息侧的 grant + scope + grantorCanReadChannel 实时权限（TOCTOU 一致）。缺此注入则
+	// obo gate 走 errOBONoChecker fail-closed，OBO 搜索会被整体拒绝。
+	ba.searchHandler.SetOBOChecker(ba)
 	ba.searchHandler.MountSubtree(r, "/v1/bot/messages",
 		ba.authBot(),
 		ba.botActorUID(),
@@ -50,7 +55,10 @@ func (ba *BotAPI) mountSearchRoutes(r *wkhttp.WKHttp) {
 
 // resolveSearchPrincipal 是 bot 搜索的 principal 解析中间件（authBot 之后、
 // searchRateLimiter 之前运行）。它按 on_behalf_of 落 as-bot 或 as-user(OBO) 主体，
-// 供后续限流/审计/handler 统一经 principal 读取。
+// 供后续限流/审计/handler 统一经 principal 读取。App Bot 在两条分支都在主体构造阶段
+// fail-closed 拒绝（一期不做）。as-user(OBO) 分支在入口做 channel 无关的 grant 存在性
+// fail-fast（validateSearchOBO），逐频道的 scope + TOCTOU 收敛下沉到 messages_search 的
+// obo gate（复用注入的 ba.SearchOBOAllowed，见 mountSearchRoutes）。
 func (ba *BotAPI) resolveSearchPrincipal(c *wkhttp.Context) {
 	obo := parseSearchOnBehalfOf(c)
 	if obo == "" {
@@ -73,14 +81,28 @@ func (ba *BotAPI) resolveSearchPrincipal(c *wkhttp.Context) {
 		return
 	}
 
-	// as-user(OBO)：on_behalf_of 存在 → 以 grantor 身份搜索。grant + scope +
-	// grantorCanReadChannel 的实时权限校验（TOCTOU 与发消息侧一致）由 validateSearchOBO
-	// 承载——YUJ-53 / #F 复用 obo_check.go 接线；在其落地前本层 fail-closed，绝不放行
-	// 未校验的 as-user 搜索。
+	// as-user(OBO)：on_behalf_of 存在 → 以 grantor 身份搜索。
 	botUID := getRobotIDFromContext(c)
+
+	// App Bot 一期整体不做，且拿不到 OBO grant（grant 只查 robot 表，App Bot 在 app_bot
+	// 表）。在主体构造阶段显式 fail-closed 拒绝——与 as-bot 分支一致，而非放行后靠 obo gate
+	// 静默返回空，避免 App Bot 搜索被误当作「无结果」。
+	if getBotKindFromContext(c) == BotKindApp {
+		ba.Warn("search OBO denied: app bot is not supported (YUJ-49)",
+			zap.String("bot", botUID))
+		httperr.ResponseErrorLWithStatus(c, errcode.ErrBotAPIBotUnavailable, nil, nil)
+		c.Abort()
+		return
+	}
+
+	// 入口 grant 存在性 fail-fast（channel 无关）：bot 当前是否被 grantor 授权（active=1 且
+	// global_enabled=1，与发消息侧 checkOBO 的 grant 门同一谓词）。无 grant → 整体拒绝，
+	// 避免把完全未授权的 OBO 搜索降级为「逐频道过滤后 200 空结果」。逐频道的 scope +
+	// grantorCanReadChannel 实时权限（TOCTOU 与发消息侧一致）由 messages_search 的 obo gate
+	// 经注入的 ba.SearchOBOAllowed 收敛（见 mountSearchRoutes / obo.go）。
 	if err := ba.validateSearchOBO(c, botUID, obo); err != nil {
 		if errors.Is(err, ErrOBONotAuthorized) {
-			ba.Warn("search OBO denied: no active grant/scope",
+			ba.Warn("search OBO denied: no active grant",
 				zap.String("bot", botUID), zap.String("on_behalf_of", obo))
 			httperr.ResponseErrorLWithStatus(c, errcode.ErrBotAPIOBONotAuthorized, nil, nil)
 			c.Abort()
@@ -92,9 +114,8 @@ func (ba *BotAPI) resolveSearchPrincipal(c *wkhttp.Context) {
 		c.Abort()
 		return
 	}
-	// spaceID 目前不从 bot 请求携带（bot 无 space_member）；OBO 场景的 Space 归属由 #F
-	// 随 grantor 解析并注入。#B 先以空 spaceID 组装载体（在 validateSearchOBO 放行前不
-	// 可达此处），待 #F 接线后补全。
+	// spaceID 不从 bot 请求携带（bot 无 space_member）。P2P 分支回退到 grantor 好友判定
+	// （安全，grantor=创建者）；群/子区经 obo gate 逐频道 ∩ scope 收敛，不依赖 spaceID。
 	p, err := messages_search.NewOBOPrincipal(botUID, obo, "")
 	if err != nil {
 		ba.Warn("search OBO principal build failed",
@@ -107,16 +128,31 @@ func (ba *BotAPI) resolveSearchPrincipal(c *wkhttp.Context) {
 	c.Next()
 }
 
-// validateSearchOBO 是 as-user(OBO) 搜索的实时权限校验挂载点（YUJ-53 / #F）。
-// #F 将在此复用 obo_check.go 的 grant + scope + grantorCanReadChannel（TOCTOU 与发消息
-// 侧一致）做实时收敛。#B（本层仅入口接线）fail-closed：on_behalf_of 搜索在 #F 落地前
-// 一律按未授权拒绝，杜绝放行任何未经校验的 as-user 搜索。返回 ErrOBONotAuthorized 时
-// 调用方隐藏 grant 是否存在（与发消息侧一致的存在性隐藏）。
+// validateSearchOBO 是 as-user(OBO) 搜索的入口 fail-fast：channel 无关地校验 bot 当前是否
+// 被 grantor 授权（active=1 且 global_enabled=1 的 grant 行，与发消息侧 checkOBO 的 grant
+// 门同一谓词）。逐频道的 scope + grantorCanReadChannel 实时权限（TOCTOU）不在此层——它需要
+// channel 维度，由 messages_search 的 obo gate 经注入的 ba.SearchOBOAllowed 承载（决策九：
+// 单频道门与 global allowlist 共用同一谓词）。
+//
+// 返回约定：
+//   - nil                    → grant 存在，principal 可组装；越权收敛交给逐频道 gate。
+//   - ErrOBONotAuthorized     → 无 active+enabled grant / 主体缺失 / 自授 → 整体拒绝
+//     （存在性隐藏，与发消息侧一致，不泄露 grant 是否存在）。
+//   - 其他 err                → 基础设施错误 → 调用方 fail-closed（INTERNAL）。
 func (ba *BotAPI) validateSearchOBO(c *wkhttp.Context, botUID, grantorUID string) error {
 	_ = c
-	_ = botUID
-	_ = grantorUID
-	return ErrOBONotAuthorized
+	if botUID == "" || grantorUID == "" || botUID == grantorUID {
+		// 与 checkOBO 一致的 fail-closed：主体缺失 / 自授一律未授权。
+		return ErrOBONotAuthorized
+	}
+	grant, err := ba.oboStoreOrDefault().findActiveGrantByGrantorBot(grantorUID, botUID)
+	if err != nil {
+		return err
+	}
+	if grant == nil {
+		return ErrOBONotAuthorized
+	}
+	return nil
 }
 
 // parseSearchOnBehalfOf 读取并还原请求 body，解析出 on_behalf_of（缺失 / body 为空 /
--- a/modules/bot_api/search_route.go
+++ b/modules/bot_api/search_route.go
@@ -1,0 +1,137 @@
+package bot_api
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"io"
+	"strings"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/modules/messages_search"
+	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
+	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
+	appwkhttp "github.com/Mininglamp-OSS/octo-server/pkg/wkhttp"
+	"go.uber.org/zap"
+)
+
+// 消息搜索的 bot 入口（YUJ-49 / #B，决策六/七/十）。
+//
+// 把 messages_search 的全部 _search* 端点挂到 /v1/bot/messages，以 authBot() 鉴权，
+// principal 由请求 body 的 on_behalf_of 字段区分：
+//   - 无 on_behalf_of → as-bot（principal=user_bot，subjectUID=botUID）。
+//   - 有 on_behalf_of → as-user(OBO)（principal=obo，subjectUID=grantorUID）。
+//
+// 中间件链对齐 messages_search/api.go（SpaceMiddleware 除外——/v1/bot 不挂 Space 门，
+// bot 无 space_member、spaceID 走 principal 参数）：
+//
+//	authBot → botActorUID → SharedUIDRateLimiter → resolveSearchPrincipal → searchRateLimiter → audit → backendGate
+//
+// botActorUID 把 robot_id 落到 "uid"（与 incoming_webhook 一致），使随后的
+// SharedUIDRateLimiter 按 botUID 而非 IP 计量——与「限流按 botUID，防单 bot 打爆」的
+// 决策一致；细粒度 searchRateLimiter 同样经 principal.RateLimitKey() 归到 botUID。
+// 审计 login_uid 记搜索主体、并在 as-user 时同记 bot_uid + grantor_uid。
+
+// searchOBOFieldProbe 只解析 on_behalf_of，用于在 handler BindJSON 之前区分主体。
+// 其余搜索参数留给各 _search* handler 自行绑定。
+type searchOBOFieldProbe struct {
+	OnBehalfOf string `json:"on_behalf_of"`
+}
+
+// mountSearchRoutes 挂载 bot 搜索子树。由 Route() 调用。
+func (ba *BotAPI) mountSearchRoutes(r *wkhttp.WKHttp) {
+	ba.searchHandler.MountSubtree(r, "/v1/bot/messages",
+		ba.authBot(),
+		ba.botActorUID(),
+		appwkhttp.SharedUIDRateLimiter(r, ba.ctx),
+		ba.resolveSearchPrincipal,
+	)
+}
+
+// resolveSearchPrincipal 是 bot 搜索的 principal 解析中间件（authBot 之后、
+// searchRateLimiter 之前运行）。它按 on_behalf_of 落 as-bot 或 as-user(OBO) 主体，
+// 供后续限流/审计/handler 统一经 principal 读取。
+func (ba *BotAPI) resolveSearchPrincipal(c *wkhttp.Context) {
+	obo := parseSearchOnBehalfOf(c)
+	if obo == "" {
+		// as-bot：以 User Bot 自身身份搜索。App Bot 一期显式拒绝（决策五）。
+		p, err := messages_search.AuthenticateUserBot(c)
+		if err != nil {
+			if errors.Is(err, messages_search.ErrAppBotSearchDenied) {
+				ba.Warn("search denied: app bot is not supported (YUJ-49)",
+					zap.String("bot", getRobotIDFromContext(c)))
+				httperr.ResponseErrorLWithStatus(c, errcode.ErrBotAPIBotUnavailable, nil, nil)
+				c.Abort()
+				return
+			}
+			// robot_id 缺失——authBot 之后不应发生，fail-closed。
+			respondBotAPIAuthFailed(c)
+			return
+		}
+		messages_search.SetPrincipal(c, p)
+		c.Next()
+		return
+	}
+
+	// as-user(OBO)：on_behalf_of 存在 → 以 grantor 身份搜索。grant + scope +
+	// grantorCanReadChannel 的实时权限校验（TOCTOU 与发消息侧一致）由 validateSearchOBO
+	// 承载——YUJ-53 / #F 复用 obo_check.go 接线；在其落地前本层 fail-closed，绝不放行
+	// 未校验的 as-user 搜索。
+	botUID := getRobotIDFromContext(c)
+	if err := ba.validateSearchOBO(c, botUID, obo); err != nil {
+		if errors.Is(err, ErrOBONotAuthorized) {
+			ba.Warn("search OBO denied: no active grant/scope",
+				zap.String("bot", botUID), zap.String("on_behalf_of", obo))
+			httperr.ResponseErrorLWithStatus(c, errcode.ErrBotAPIOBONotAuthorized, nil, nil)
+			c.Abort()
+			return
+		}
+		ba.Error("search OBO check failed", zap.Error(err),
+			zap.String("bot", botUID), zap.String("on_behalf_of", obo))
+		httperr.ResponseErrorLWithStatus(c, errcode.ErrBotAPIOBOInternal, nil, nil)
+		c.Abort()
+		return
+	}
+	// spaceID 目前不从 bot 请求携带（bot 无 space_member）；OBO 场景的 Space 归属由 #F
+	// 随 grantor 解析并注入。#B 先以空 spaceID 组装载体（在 validateSearchOBO 放行前不
+	// 可达此处），待 #F 接线后补全。
+	p, err := messages_search.NewOBOPrincipal(botUID, obo, "")
+	if err != nil {
+		ba.Warn("search OBO principal build failed",
+			zap.String("bot", botUID), zap.String("on_behalf_of", obo), zap.Error(err))
+		httperr.ResponseErrorLWithStatus(c, errcode.ErrBotAPIOBONotAuthorized, nil, nil)
+		c.Abort()
+		return
+	}
+	messages_search.SetPrincipal(c, p)
+	c.Next()
+}
+
+// validateSearchOBO 是 as-user(OBO) 搜索的实时权限校验挂载点（YUJ-53 / #F）。
+// #F 将在此复用 obo_check.go 的 grant + scope + grantorCanReadChannel（TOCTOU 与发消息
+// 侧一致）做实时收敛。#B（本层仅入口接线）fail-closed：on_behalf_of 搜索在 #F 落地前
+// 一律按未授权拒绝，杜绝放行任何未经校验的 as-user 搜索。返回 ErrOBONotAuthorized 时
+// 调用方隐藏 grant 是否存在（与发消息侧一致的存在性隐藏）。
+func (ba *BotAPI) validateSearchOBO(c *wkhttp.Context, botUID, grantorUID string) error {
+	_ = c
+	_ = botUID
+	_ = grantorUID
+	return ErrOBONotAuthorized
+}
+
+// parseSearchOnBehalfOf 读取并还原请求 body，解析出 on_behalf_of（缺失 / body 为空 /
+// 非法 JSON → 返回 ""，交由后续 handler 的 BindJSON 统一报错）。还原 body 使各 _search*
+// handler 能再次 BindJSON 读取完整请求体。
+func parseSearchOnBehalfOf(c *wkhttp.Context) string {
+	body, err := c.GetRawData()
+	if err != nil || len(body) == 0 {
+		return ""
+	}
+	// 还原 body 供 handler 复读（GetRawData 会消费 Request.Body）。
+	c.Request.Body = io.NopCloser(bytes.NewReader(body))
+	var probe searchOBOFieldProbe
+	if json.Unmarshal(body, &probe) != nil {
+		return ""
+	}
+	return strings.TrimSpace(probe.OnBehalfOf)
+}
--- a/modules/bot_api/search_route_test.go
+++ b/modules/bot_api/search_route_test.go
@@ -1,0 +1,120 @@
+package bot_api
+
+import (
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/Mininglamp-OSS/octo-server/modules/messages_search"
+	"github.com/gin-gonic/gin"
+)
+
+// YUJ-49 (#B) — bot search entry: resolveSearchPrincipal distinguishes as-bot
+// (no on_behalf_of) from as-user(OBO) (on_behalf_of present), and App Bot is
+// explicitly denied. The OBO real-permission check (grant/scope/TOCTOU) is #F;
+// until it wires validateSearchOBO, on_behalf_of搜索 fails closed.
+
+func newBotSearchCtx(t *testing.T, body string) (*wkhttp.Context, *httptest.ResponseRecorder) {
+	t.Helper()
+	gin.SetMode(gin.TestMode)
+	rec := httptest.NewRecorder()
+	gc, _ := gin.CreateTestContext(rec)
+	gc.Request = httptest.NewRequest("POST", "/v1/bot/messages/_search", strings.NewReader(body))
+	return &wkhttp.Context{Context: gc}, rec
+}
+
+func newSearchTestBotAPI() *BotAPI { return &BotAPI{Log: log.NewTLog("bot-search-test")} }
+
+func TestResolveSearchPrincipal_AsBot(t *testing.T) {
+	ba := newSearchTestBotAPI()
+	c, rec := newBotSearchCtx(t, `{"keyword":"hi"}`)
+	c.Set(CtxKeyRobotID, "bot_1")
+	c.Set(CtxKeyBotKind, BotKindUser)
+
+	ba.resolveSearchPrincipal(c)
+
+	if c.IsAborted() {
+		t.Fatalf("as-bot must not abort, body=%q", rec.Body.String())
+	}
+	if got := messages_search.PrincipalKind(c); got != "user_bot" {
+		t.Fatalf("principal kind = %q, want user_bot", got)
+	}
+	// Body must still be readable by the downstream _search* handler (the probe
+	// consumed it via GetRawData and must have restored it).
+	b, _ := io.ReadAll(c.Request.Body)
+	if !strings.Contains(string(b), "keyword") {
+		t.Fatalf("request body not restored for handler BindJSON, got %q", string(b))
+	}
+}
+
+func TestResolveSearchPrincipal_AppBotDenied(t *testing.T) {
+	ba := newSearchTestBotAPI()
+	c, rec := newBotSearchCtx(t, `{}`)
+	c.Set(CtxKeyRobotID, "app_1")
+	c.Set(CtxKeyBotKind, BotKindApp)
+
+	ba.resolveSearchPrincipal(c)
+
+	if !c.IsAborted() {
+		t.Fatalf("App Bot must be denied (决策五), chain must abort")
+	}
+	if got := messages_search.PrincipalKind(c); got != "" {
+		t.Fatalf("App Bot must not set a search principal, got %q", got)
+	}
+	if rec.Code == http.StatusOK {
+		t.Fatalf("App Bot denial must not return 200")
+	}
+}
+
+func TestResolveSearchPrincipal_OBOFailClosedUntilF(t *testing.T) {
+	ba := newSearchTestBotAPI()
+	c, rec := newBotSearchCtx(t, `{"on_behalf_of":"grantor_2","keyword":"x"}`)
+	c.Set(CtxKeyRobotID, "bot_1")
+	c.Set(CtxKeyBotKind, BotKindUser)
+
+	ba.resolveSearchPrincipal(c)
+
+	if !c.IsAborted() {
+		t.Fatalf("on_behalf_of search must fail closed until #F wires validateSearchOBO")
+	}
+	if got := messages_search.PrincipalKind(c); got != "" {
+		t.Fatalf("fail-closed OBO must not set a principal, got %q", got)
+	}
+	if rec.Code == http.StatusOK {
+		t.Fatalf("fail-closed OBO must not return 200")
+	}
+}
+
+func TestParseSearchOnBehalfOf(t *testing.T) {
+	// present (trimmed) + body restored for the handler.
+	c, _ := newBotSearchCtx(t, `{"on_behalf_of":"  g1  ","keyword":"k"}`)
+	if got := parseSearchOnBehalfOf(c); got != "g1" {
+		t.Fatalf("trimmed on_behalf_of = %q, want g1", got)
+	}
+	b, _ := io.ReadAll(c.Request.Body)
+	if !strings.Contains(string(b), "keyword") {
+		t.Fatalf("body must be restored after probe, got %q", string(b))
+	}
+
+	// absent → "".
+	c2, _ := newBotSearchCtx(t, `{"keyword":"k"}`)
+	if got := parseSearchOnBehalfOf(c2); got != "" {
+		t.Fatalf("no on_behalf_of field = %q, want empty", got)
+	}
+
+	// invalid JSON → treated as no obo (handler re-binds and reports the error).
+	c3, _ := newBotSearchCtx(t, `not json`)
+	if got := parseSearchOnBehalfOf(c3); got != "" {
+		t.Fatalf("invalid json on_behalf_of = %q, want empty", got)
+	}
+
+	// empty body → "".
+	c4, _ := newBotSearchCtx(t, ``)
+	if got := parseSearchOnBehalfOf(c4); got != "" {
+		t.Fatalf("empty body on_behalf_of = %q, want empty", got)
+	}
+}
--- a/modules/bot_api/search_route_test.go
+++ b/modules/bot_api/search_route_test.go
@@ -1,6 +1,7 @@
 package bot_api
 
 import (
+	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -13,10 +14,12 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
-// YUJ-49 (#B) — bot search entry: resolveSearchPrincipal distinguishes as-bot
-// (no on_behalf_of) from as-user(OBO) (on_behalf_of present), and App Bot is
-// explicitly denied. The OBO real-permission check (grant/scope/TOCTOU) is #F;
-// until it wires validateSearchOBO, on_behalf_of搜索 fails closed.
+// YUJ-49 (#B) / YUJ-53 (#F) — bot search entry: resolveSearchPrincipal
+// distinguishes as-bot (no on_behalf_of) from as-user(OBO) (on_behalf_of
+// present). App Bot is explicitly denied on both branches. The OBO entry does
+// a channel-agnostic grant-existence fail-fast (validateSearchOBO); the
+// per-channel scope + grantorCanReadChannel (TOCTOU) check is reused by the
+// messages_search obo gate via the injected ba.SearchOBOAllowed.
 
 func newBotSearchCtx(t *testing.T, body string) (*wkhttp.Context, *httptest.ResponseRecorder) {
 	t.Helper()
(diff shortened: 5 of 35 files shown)


Jerry-Xin (changes requested):
The principal abstraction (user / user_bot / obo-grantor / uk) is well-designed: single-channel gates and the global allowlist share one predicate, App Bots are explicitly denied, and failures fail closed with anti-enumeration responses. The as-bot and OBO authorization models are sound. However, the newly-wired User API-key (uk) route introduces a Space-membership authorization gap on global DM search that the equivalent human route does not have.

## 🔴 Blocking

- 🔴 **Stale User API keys bypass current Space-membership enforcement on global DM search.** `modules/botfather/search_route.go` mounts `/v1/user/messages` with `authUserAPIKey → resolveUKPrincipal` and deliberately no `SpaceMiddleware`. `authUserAPIKey` (`modules/botfather/api_user.go:34`) validates only key status, integration-client enablement, and active user — it does **not** re-check that the key owner is still a member of the key's frozen `space_id`. The uk principal is then built with that stored `api_key_space_id` (`modules/messages_search/principal.go` `authenticateUK`).

  On the human web route, `pkg/space/middleware.go` `SpaceMiddleware` calls `CheckMembership(spaceID, uid)` and **rejects** the request when the caller is no longer a member (`middleware.go:135-147`). The uk route has no equivalent gate, so a frozen `space_id` is trusted indefinitely.

  Impact is bounded to the **global DM allowlist**: `buildAllowlist` → `enumerateDMPeers` (`modules/messages_search/search_global.go:971`) unions the caller's friends with **every active member** of the stored Space via `queryDMSpaceMemberUIDs` (`search_global.go:1165`: `SELECT uid FROM space_member WHERE space_id=? AND status=1 AND uid<>?`) — with no check that the caller (`loginUID`) is still in that Space. Because global search trusts the allowlist directly (it does not re-run the per-channel `checkP2PAccess` for enumerated DM peers), the resulting `terms(channelId)` set surfaces DM history with all Space members.

  Constructable reproducer: user U mints a `uk_` key while a member of Space X (key row stores `space_id=X`, `status=1`); U is later removed from Space X but the key is not revoked; U calls `POST /v1/user/messages/_search/global` with that key. `authUserAPIKey` passes, the uk principal carries `spaceID=X`, and `enumerateDMPeers` returns all current Space-X members, letting U search DM message history with Space-X users. The human route would be rejected by `SpaceMiddleware.CheckMembership`.

  (Groups/threads are NOT affected: they pass through `GetGroupsWithMemberUID(keyUID)` + `ExistMembersActive(keyUID)`, a live membership gate. The single-channel DM gate is also partially self-correcting: `checkP2PAccess`'s same-Space branch uses `AreSpaceMembers(spaceID, loginUID, peer)`, which a removed owner fails, falling back to `IsFriend`. The gap is specifically the global DM enumeration, which does not re-validate the caller's own Space membership.)

  Fix: validate the key owner's current active membership in `api_key_space_id` before constructing the uk principal (or apply a membership-check middleware equivalent to `SpaceMiddleware` on the uk route), and add a regression test for an active key whose owner has been removed/disabled in that Space.

## 💬 Non-blocking

- 🟡 **OBO global search runs sequential per-channel authorization checks.** `filterChannelsByOBO` (`modules/messages_search/obo.go`) filters groups, DMs, and threads one channel at a time, and each `SearchOBOAllowed` call reaches the send-side `checkOBO` with multiple DB lookups (grant + scope + live-access). A large grantor allowlist can produce many queries per request and amplify a rate-limited bot into notable DB load. Consider batch-loading grant scopes + live-access state and intersecting in memory. Not a blocker (bounded by the shared UID rate limiter).

## ✅ Highlights

- OBO grant is keyed on `(grantor, botUID)` via `findActiveGrantByGrantorBot`, so a client-supplied `on_behalf_of` cannot forge acting-as an arbitrary user — no IDOR. Scope = grantor's live reachable set ∩ OBO-granted channels, with TOCTOU re-checks (`grantorCanReadChannel`) matching the send path exactly.
- as-bot visibility is correctly scoped to the bot's own friends (DM) and its active group/thread memberships — not global.
- Single-channel and global authorization share one principal-aware predicate (canReadChannel ⇔ enumerateReadableChannels), preventing drift.
- Space-less bot/OBO principals correctly proceed without the spaceId term because their reachability is bounded by the per-principal allowlist, while human/uk principals remain fail-closed on empty Space.
- `go test ./modules/messages_search` passed. Broader bot_api integration tests could not run locally (MySQL/Redis/WuKongIM not provisioned).

lml2468 (commented):
# PR Review — PR #608 (Mininglamp-OSS/octo-server)

Reviewed at head SHA `93f531da` against merge-base `c1530e5f` (base `main`).
**Build/vet/test:** `go build ./...` exit 0 ✅, `go vet ./...` exit 0 ✅, `go test ./modules/messages_search/` ✅ (full suite), `go test ./modules/bot_api/ -run OBO/Principal` ✅. ⚠️ `go vet -tags integration ./...` **FAILS** with an import cycle (details in Finding 1) — merge-base was clean. **NOT run:** DB/Redis/WuKongIM integration tests (infra-gated).
Classification: **security-sensitive** — authorization boundary (principal-aware message search: as-bot + as-user/OBO). Reviewed in SECURITY emphasis against the load-bearing `space-isolation` rule (priority 92).

Intent: make message search principal-aware so octo-cli can search as the User Bot's own identity (friends + joined groups/threads) or on-behalf-of the grantor (OBO), reusing the human authz path; App Bot explicitly denied. Matches issue #607 acceptance.

## 1. Spec compliance — ✅
Graded against issue #607 (no `.octospec` brief for this task).
- **Missing:** none. as-bot friend/group/thread/global visibility, OBO grant+scope+TOCTOU, rate-limit keyed by botUID, audit records botUID+grantorUID, App Bot fail-closed, existing suite + new matrix — all present.
- **Over-build:** none. `uk` principal is wired as the documented real-user variant; scope stays within the search module + its route mounting.
- **Deviation:** none. OBO reuses the send-side `checkOBO` predicate exactly; the single-channel gate and global allowlist share one predicate (the stated anti-drift design).
- **Load-bearing `space-isolation` (P92): PASS** — see Verified clean. The spaceless-bot relaxation does not open a cross-tenant path: reachability is bounded by the per-principal allowlist, not the spaceId term.

## 2. Code quality — Approved with one non-blocking P2 (no P0/P1)

### Findings

**P2 / MEDIUM (confidence HIGH) — PR introduces an import cycle that breaks the `-tags integration` build.** `modules/bot_api/bot_api.go` + `search_route.go` now import `modules/messages_search` (new edge, absent at merge-base). Combined with the existing `messages_search → modules/message` import and the `//go:build integration` test file `modules/message/api_card_action_test.go` (→ `app_bot` → `bot_api`), this closes a cycle:
`message`(integration test) → `app_bot` → `bot_api` → `messages_search` → `message`.
- **Repro:** `go vet -tags integration ./modules/message/` → `import cycle not allowed in test` (exit 1). Verified clean at merge-base `c1530e5f`.
- **Blast radius (bounded):** production `go build`/`go vet ./...` and the default `go test ./...` are **unaffected** (cycle is only in the integration-tagged test build). Per `ci.yml` (issue #557), CI **intentionally does not run `-tags integration ./...`**, so **the current CI gate is not broken** — but anyone running/enabling that suite (the documented future `conv_ext_test` provisioning work) will hit a hard compile failure that traces back to this PR.
- **Fix:** break the new `bot_api → messages_search` compile edge — e.g. have `messages_search` expose the OBO-checker/route-mount seam via an interface that `bot_api` satisfies without `messages_search` importing `modules/message` on the same cycle, or move the `messages_search → modules/message` dependency behind an interface. (The module already uses an injected `oboChecker` seam for exactly this decoupling reason — the route-mount edge just needs the same treatment.)

- **Spaceless-bot global search does NOT leak cross-tenant DMs.** With `spaceID=""`, `resolveGlobalScope` restricts the OS query to `scope ⊆ allowSet`, where `allowSet` = `enumerateReadableChannels(principal)` = the per-principal predicate (as-bot: `IsFriend ∪ ExistMemberActive`; OBO: grantor allowlist ∩ OBO scope). The spaceId term is a redundant DM double-guard, not the primary bound — a bot can only match channels it is actually party to. Confirmed by reading `resolveGlobalScope` end-to-end.
- **RequireSpaceID relaxation is correctly scoped.** Only `RequiresSpaceScope()==false` principals (as-bot/OBO) bypass; `user`/`uk` still fail-close — locked by `TestResolveGlobalScope_UserEmptySpaceFailsClosed_RequireTrue` (regression guard) alongside `..._OBOEmptySpaceBypassesGate_RequireTrue`.
- **Unified predicate holds (决策九).** `canReadChannel` and `enumerateReadableChannels` dispatch to the same per-kind predicate; single-channel gate ⇔ allowlist membership. No inlined rule copy that can drift.
- **App Bot fail-closed at construction** in all three entries: `authenticateUserBot`, `authenticateOBOFromContext`, and the `bot_api` route (`getBotKindFromContext(c)==BotKindApp` → `ErrBotAPIBotUnavailable`, aborts). App Bot cannot reach the allowlist path.
- **OBO reuses send-side authz exactly.** `SearchOBOAllowed` → `checkOBO` (grant + scope + `grantorCanReadChannel` live), so TOCTOU matches the send path (kicked/unfriended/scope-off grantor → immediately unsearchable). Infra error → fail-closed INTERNAL; unauthorized → NOT_FOUND (existence-hiding).
- **as-bot P2P gate** is friends-only single-edge `IsFriend(botUID,peer)` — no Space (bot has no `space_member`), no blacklist (bot can't blacklist / peer→bot deliberately not consulted, documented), matching `GetFriends(botUID)` enumeration. `self-grant` (botUID==grantorUID) rejected.
- **Rate-limit + audit** keyed by botUID (as-bot/OBO), audit records both botUID and grantorUID — per acceptance.

## 3. Test assessment
Strong: the new suites (`principal_test`, `authz_bot_test`, `obo_test`, `search_global_bot_test`, `search_global_space_gate_test`, `route_subtree_test`, `space_filter_test`, plus `bot_api` OBO adapter/matrix) cover the visibility matrix, the space-gate bypass-vs-fail-close split, App Bot denial, and fail-closed infra errors. Would fail without the implementation. **Gap:** the import-cycle regression (Finding 1) is exactly the kind of thing a `-tags integration` compile-check in CI would have caught; not a test-content gap in this module.

## 4. Coverage / blind spots
- Verified by build/vet/unit-tests on the checkout; DB/Redis/WuKongIM integration NOT run (infra-gated).
- The isolation argument rests on the OS query being bounded by `osChannelIDs` from the allowlist — verified by code path, not by a live OpenSearch round-trip. Recommend the infra-provisioned integration run in CI to confirm end-to-end.

## 5. For a human to double-check
- **Import-cycle policy call (Finding 1):** confirm whether breaking the `bot_api → messages_search` edge is required now, or acceptable to defer until the `conv_ext_test` provisioning work enables `-tags integration ./...`. It does not block today's CI but is a latent regression from a currently-clean state.
- Live OpenSearch confirmation that a spaceless as-bot global search returns only allowlisted channels (belt-and-suspenders on the isolation argument).

## Verdict
**COMMENT** — the authorization design is sound and the load-bearing space-isolation boundary holds (no P0/P1; security model byte-verified and well-tested). The one real finding is a **P2**: this PR regresses the `-tags integration` build from clean → import-cycle via the new `bot_api → messages_search` edge. The branch **compiles in production and in the CI gate** (which intentionally does not run `-tags integration ./...` per #557), so per the verdict rules a lone P2 is not a merge-blocker — but it is a real regression from a clean merge-base and the fix is localized, so it should be resolved or explicitly deferred by the maintainer before enabling the integration lane.
Suggested decision: `Safe to merge after non-blocking follow-up` — fix the import cycle (or record an explicit decision to defer it until the `conv_ext_test` integration lane is provisioned). Recommend a maintainer sign-off on that call since merge-base was clean.

yujiawei (approved):
# Code Review — PR #608 (octo-server)

Principal-aware message search: search as-bot (User Bot identity), as-user (OBO / grantor), and as-user (uk / User API Key), with App Bot explicitly denied. Reviewed against the linked issue's acceptance criteria, at head `93f531d`, merge-base `c1530e5`.

**Verdict: APPROVED.** No blocking (P0/P1) issues found. Build, `vet`, and the `messages_search` unit suite are green locally. The items below are P2 (non-blocking) — worth a follow-up but not a merge gate.

## Spec compliance

Implements the issue's five acceptance criteria without over-building:

1. **as-bot** friend-DM / group / thread / global visibility — friends-only P2P (no blacklist), `ExistMemberActive(botUID)` group/thread gate, `friends ∪ joined-groups ∪ threads` global allowlist.
2. **as-user (OBO)** — grant existence fail-fast at the entry middleware, then per-channel grant + scope + live `grantorCanReadChannel` (TOCTOU) reusing the send-path checker; bidirectional blacklist retained via the grantor real-user branch.
3. Rate limit keyed by `botUID` (as-bot / OBO), audit records `bot_uid` + `grantor_uid`.
4. **App Bot denied** — fail-closed at principal construction (`bot_kind == "app"`), on both the as-bot and OBO branches, before reaching search.
5. Existing suite green + new principal visibility matrix tests.

The single-channel gate (`canReadChannel`) and the global allowlist enumeration (`enumerateReadableChannels`) are dispatched through one predicate per principal kind, and all four single-channel handlers plus both global fast-path dispatchers were migrated to `canReadChannel` — so the "single-channel admit ⇔ appears in global allowlist" invariant holds by construction. The `nil`-checker OBO seam fails closed (`errOBONoChecker`), and `botUID == grantorUID` self-grant is rejected in `authenticateOBO`, `checkOBO`, and `validateSearchOBO`.

## Non-blocking findings (P2)

**1. OBO global search fans out one un-batched `checkOBO` per channel.** `filterChannelsByOBO` (`modules/messages_search/obo.go:878`) loops over the grantor's entire allowlist and calls `oboAllowed` → `SearchOBOAllowed` → `checkOBO` for each channel. Each iteration issues ~3–4 non-batched MySQL queries (`findActiveGrantByGrantorBot`, `scopeEnabled`, sometimes `scopeRowExists`, and `grantorCanReadChannel`), and the grant lookup is invariant across the loop yet re-fetched every time. This runs inside `resolveGlobalScope`, i.e. before the per-request `context.WithTimeout` is established. For a grantor in hundreds of groups an OBO global search can spike latency and pressure the DB pool. Suggest hoisting the invariant grant lookup out of the loop and/or a batch `FilterOBOAllowed` seam. Not a correctness bug — the per-channel semantics are exactly right.

**2. OBO entry-gate denials return 403 rather than the anti-enumeration 404 used per-channel.** In `resolveSearchPrincipal` (`modules/bot_api/search_route.go`), a missing/revoked grant or self-grant returns `ErrBotAPIOBONotAuthorized` (403), while per-channel OBO denials collapse to `NOT_FOUND`. This is a minor grant-validity oracle for a probing bot; the exposure is bounded because a bot can already read its own active grant via `/v1/bot/obo-grant`. Consider aligning the entry denial with the per-channel not-found surface for consistency.

**3. Thread hard-caps can make `canReadChannel` and the global allowlist disagree.** In `enumerateThreadsForGroups` (`modules/messages_search/search_global.go`), a group with more than `maxThreadsPerGroup` (200) threads, or threads past the `maxTotalThreadChannelIDs` (2000) aggregate, is dropped from the global allowlist while a direct single-channel search of the same thread still passes `checkThreadAccess`. This is a pre-existing bound shared with the human `buildAllowlist` (not introduced by this PR), and it fails *closed* (omits results, never over-shares), so it is a documented degradation rather than a leak — but it does technically break the strict "single-channel ⇔ allowlist" equivalence at the cap boundary. Worth a comment/telemetry follow-up.

**4. Minor cleanups.** `authenticateOBOFromContext` (`modules/messages_search/principal.go:261`) reads `spacepkg.GetSpaceID(c)` from an unvalidated header and is currently **dead code** — the live path uses `NewOBOPrincipal(bot, grantor, "")` which correctly strips spaceID. Leaving an unused constructor that reintroduces an attacker-controlled spaceID is a latent footgun; consider removing it or adding a guard comment. Separately, `SetOBOChecker` exports a method taking the unexported `oboChecker` interface — harmless (structural satisfaction works) but linters may flag it.

## Coverage / blind spots

- Verified the diff against the actual callee semantics (not just comments): `ExistMemberActive`/`ExistMembersActive` (is_deleted=0 AND status=Normal — kicked/blacklisted excluded), `IsFriend`/`GetFriends` (is_deleted=0, mutual-pair storage), `QueryPeerRobotInfo`, thread enumeration, and `grantorCanReadChannel`. App-Bot rejection keys on `bot_kind` set unconditionally by `authBot`.
- Two independent review passes were run in addition to this one. One flagged a repo-wide compilation failure from exported/unexported identifier drift (`AuthenticateUserBot`, `NewOBOPrincipal`, etc.); this was a **false positive** caused by reviewing the diff in isolation — those exported wrappers exist in `route_subtree.go`, and `go build ./modules/messages_search/... ./modules/bot_api/... ./modules/botfather/...` compiles clean. Recorded here so it is not silently treated as unresolved.
- Not verifiable in this environment: runtime OpenSearch index mappings (`visibles`, `spaceId`, `payload.space_id`), deployed configuration (`OCTO_SEARCH_REQUIRE_SPACE_ID`), and live DB contents. The spaceless fail-close relaxation (`RequiresSpaceScope()` gating only user/uk) was verified in code to leave the human path fully fail-closed; its DM-scoping correctness depends on the indexer writing `spaceId`, which is an operational precondition outside this diff.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

## Octo-Q automated review Review — PR Mininglamp-OSS/octo-server #608
**Head SHA:** `93f531da997793d7996f27e234c50eefcfdbbf56`
**Merge-base:** `c1530e5f` (main)
**Scope:** 35 files, +3260/−39 lines, 9 commits

---

### 1. Verification Summary

| Area | Verdict | Evidence |
|------|---------|----------|
| Principal abstraction (4 kinds) | ✅ | `principal.go:22-46` — clean interface + 4 value types; dispatch by `Kind()` |
| App Bot denial | ✅ | `principal.go:228` + `search_route.go:91` — fail-closed at construction |
| Rate-limit key derivation | ✅ | `principal.go` RateLimitKey: user→uid, bot/OBO→botUID, uk→keyUID |
| Audit trail | ✅ | `audit.go:40-48` — login_uid=SubjectUID, conditional bot_uid/grantor_uid |
| Single-channel gate parity | ✅ | All 5 handlers (search_all/messages/files/media/around) switched to `h.principal(c).SubjectUID()` + `canReadChannel` |
| Global allowlist parity | ✅ | `resolveGlobalScope:471` uses `principalForSubject` → dispatches by principal kind |
| OBO per-channel TOCTOU | ✅ | `obo.go:69` → `SearchOBOAllowed` → `checkOBO` (grant+scope+grantorCanReadChannel real-time) |
| Space-scope bypass ([automated review](mention://issue/8159ca9f-2f34-4750-850f-2e957e6c6609)) | ✅ | `space_scope.go:52` + `search_global.go:443` — `RequiresSpaceScope()` correctly gates |
| Body restoration (parseSearchOnBehalfOf) | ✅ | `search_route.go:163-167` — `io.NopCloser(bytes.NewReader(body))` |
| Shared singleton + SetOBOChecker | ✅ | `route_subtree.go:44` sync.Once; `search_route.go:47` injects before MountSubtree |

---

### 2. Findings

#### P2-1 — OBO global search: cross-Space group enumeration wider than real-user

**Severity:** P2 (functional gap, not security bypass)
**Diff-scope:** New behavior introduced by this PR
**Location:** `obo.go:107-113` → `search_global.go:630` `buildAllowlist(_, grantorUID, "")`

When the OBO principal is constructed with `spaceID=""` (bot route has no SpaceMiddleware), `buildAllowlist` skips the Space filter at line 658:

```go
if spaceID != "" && !shouldIncludeGroupForSpaceLocal(...) {
    continue
}
```

This means OBO global search enumerates **all** groups the grantor belongs to across **every** Space, then intersects with OBO scope. Real-user global search only sees groups within their current Space.

**Impact:** OBO search may surface group messages from Spaces outside the bot's deployment context. Each result still passes `checkOBO` (grantor has real access + scope enabled), so there's no authorization bypass — but the result set is wider than what the grantor would see in their own Space-scoped global search.

**Suggestion:** Document this as a known v1 behavior difference in the PR description. If the grantor's multi-Space membership is a concern, consider passing the grantor's primary Space through the OBO principal in a follow-up.

---

#### P2-2 — OBO global allowlist: N+1 per-channel scope check

**Severity:** P2 (performance)
**Diff-scope:** New behavior introduced by this PR
**Location:** `obo.go:126-138` `filterChannelsByOBO`

For global OBO search, `filterChannelsByOBO` calls `h.oboAllowed()` per channel. Each call chains into `checkOBO` which performs:
1. `findActiveGrantByGrantorBot` (SQL query or negative-cache hit)
2. `scopeEnabled` (SQL query)
3. `scopeRowExists` (SQL, for implicit-scope path)
4. `grantorCanReadChannel` (SQL)

With a grantor who has 100 groups + 50 DM peers, this produces ~150 × 4 = **600 SQL queries** per global search request. The grant lookup may hit the Redis negative cache, but scope and membership queries hit MySQL each time.

**Suggestion:** Consider a batched scope query (`WHERE grant_id IN (...) AND enabled=1`) and/or caching the grant row across the per-channel loop since it's the same (grantor, bot) pair. This is a follow-up optimization, not a blocker.

---

#### P2-3 — OBO global search: DM peer enumeration is friend-only (no Space members)

**Severity:** P2 (functional gap, narrower than real-user)
**Diff-scope:** New behavior introduced by this PR
**Location:** `obo.go:107` → `search_global.go:971` `enumerateDMPeers(grantorUID, "")`

With `spaceID=""`, `enumerateDMPeers` returns only the grantor's friend list (line 987-992 skips space_member when spaceID is empty). Non-friend Space members who the grantor can DM will NOT appear in OBO global search results, even if they have OBO scope enabled for those DM channels.

This is the **opposite** asymmetry to P2-1: groups are wider, DMs are narrower. Both stem from the missing Space context in the OBO principal.

**Impact:** Users may find their OBO search returns fewer DM results than expected. Per-channel OBO search for the specific DM channel still works correctly (the peer UID is validated via `checkOBO`'s grantor access check).

---

#### Nit-1 — validateSearchOBO comment inconsistency

**Location:** `search_route.go:126`

Comment says "channel 无关地校验 bot 当前是否被 grantor 授权（active=1 且 global_enabled=1 的 grant 行" but the code only says `findActiveGrantByGrantorBot` without mentioning global_enabled in the function-level comment at line 119. The SQL does check both (verified at `obo_db.go:417-418`), so the comment is technically correct — just inconsistent phrasing between the block comment and the function docstring.

---

### 3. Data Flow Tracing

| Consumer | Upstream Source | Verified Flow |
|----------|----------------|---------------|
| `h.principal(c).SubjectUID()` in handlers | Context key `"messages_search.principal"` set by route middleware; defaults to `authenticateUser` via `c.GetLoginUID()` | ✅ All 5 single-channel + 3 global handlers use `h.principal(c)` |
| `principalForSubject(c, loginUID, spaceID)` in global paths | First checks context for set principal, falls back to `userPrincipal{uid, spaceID}` | ✅ Correctly picks up bot/OBO principals when set by route middleware |
| `RateLimitKey()` in searchRateLimiter | `principal.RateLimitKey()` — user=uid, bot=botUID, OBO=botUID, uk=keyUID | ✅ Matches design intent (rate-limit by bot, not grantor) |
| `oboAllowed()` → `SearchOBOAllowed` → `checkOBO` | Injection seam via `SetOBOChecker`; `checkOBO` does grant(active+global_enabled) + scope + grantorCanReadChannel | ✅ TOCTOU re-verification on every per-channel call |
| `buildAllowlist(_, grantorUID, spaceID)` for OBO | spaceID="" from OBO principal constructor (`search_route.go:121`) | ⚠️ No Space filter → cross-Space groups visible (P2-1) |
| `enumerateDMPeers(grantorUID, spaceID)` for OBO | spaceID="" → friend-only, no space_member | ⚠️ Non-friend Space members invisible (P2-3) |
| `applyGlobalDMSpaceScope(b, spaceID)` in global DSL | spaceID="" from principal → no-op (line 365-366) | ✅ Correct: allowlist already bounds DM visibility |
| `applyVisiblesWhitelist(query, loginUID)` in groups | `loginUID` = `h.principal(c).SubjectUID()` = grantorUID for OBO | ✅ Correct subject used |

---

### 4. Blind-Spot Checklist (C1-C6)

| Check | Result | Evidence |
|-------|--------|----------|
| **C1 — Dual-path parity** | ✅ Clear | `canReadChannel` / `enumerateReadableChannels` dispatch by `Kind()` — single-channel gate and global allowlist share same predicates per principal type (决策九) |
| **C2 — Control-flow ordering** | ✅ Clear | Shared handler via `Shared()` singleton, `MountSubtree` appends chain in order; no nested/double invocation of middleware |
| **C3 — Authorization boundary** | ✅ Clear | `authBot()` gates bot route; `validateSearchOBO` fail-fast for OBO; per-channel `checkOBO` re-verifies grant+scope+TOCTOU; App Bot denied at principal construction in both as-bot and OBO paths |
| **C4 — Auth lifecycle cascade** | ✅ Clear | `checkOBO` re-verifies grant active status + scope enabled + grantor live access on every per-channel call — covers grant revocation / scope disable / grantor kicked from group / un-friended |
| **C5 — Build/runtime** | N/A | Backend Go code only; no build/packaging changes |
| **C6 — Governance/docs** | N/A | No governance documentation changes |

---

### 5. Cross-Round Blocker Review (R6)

N/A — first review round.

---

### 6. Extra Observations

- **Test coverage is extensive:** principal_test (534 lines), obo_test (337 lines), authz_bot_test (164 lines), search_global_bot_test (254 lines), route_subtree_test (237 lines), search_route_test (210 lines). The test infrastructure covers the principal dispatch, OBO injection, and route mounting thoroughly.

- **Double rate-limiting on bot route is intentional:** `SharedUIDRateLimiter` (by botUID via "uid" context) + `searchRateLimiter` (by `principal.RateLimitKey()` = botUID). Same key, two buckets — defense-in-depth, not a bug.

- **`principalForSubject` fallback is safe:** If no principal is set in context (web user route), it returns `userPrincipal{uid, spaceID}` — identical to pre-refactor behavior.

---

### 7. Conclusion

The PR introduces a well-designed Principal abstraction that cleanly separates four subject types (user, user_bot, OBO, uk) while maintaining 决策九 invariant (single-channel gate ≡ global allowlist predicate). The security model is sound:

- App Bot is fail-closed denied at construction in both paths
- OBO per-channel checks reuse the exact `checkOBO` from the send path (TOCTOU-equivalent)
- Space-scope bypass is correctly gated by `RequiresSpaceScope()` (only real-user principals)
- Anti-enumeration is preserved (NOT_FOUND for all denials)
- All handler paths consistently switched to `h.principal(c).SubjectUID()`

Three P2 findings document functional gaps in OBO global search (cross-Space group visibility, friend-only DM enumeration, N+1 query pattern). None represent authorization bypasses or user-visible breakage.

**[Octo-Q] verdict: APPROVE** — no P0/P1 issues found. The three P2 items are documented for follow-up and do not block landing.

lml2468 (changes requested):
# PR Review (CORRECTION → REQUEST_CHANGES) — PR #608 (Mininglamp-OSS/octo-server)

Reviewed at head SHA `93f531da` against merge-base `c1530e5f`.
**This supersedes my earlier COMMENT (review 4724263015).** On a second, deeper trace prompted by Jerry-Xin's finding, I confirmed a **real P1 space-isolation leak on the new `uk` (User API-key) route** that I missed the first pass. Converging with Jerry-Xin's REQUEST_CHANGES. Correcting the record.

## Blocking finding

### 🔴 P1 / HIGH (confidence HIGH) — stale `uk` key bypasses current-Space membership → cross-tenant DM history leak in global search
The `uk` search route is **introduced by this PR** (`modules/botfather/search_route.go`, all-added) and is mounted **without `SpaceMiddleware`** (documented as intentional: "spaceID 走 principal 的 api_key_space_id"). The upstream human web route relies on `SpaceMiddleware` (`pkg/space/middleware.go`) whose `CheckMembership` returns **403 if the caller is not a current member** of the Space. The uk route removes that gate and nothing replaces it:

1. `authUserAPIKey` (`modules/botfather/api_user.go`) validates key status / integration-client / user-active, then sets `api_key_space_id = keyModel.SpaceID` **unconditionally** — it never re-checks that the key owner is still a member of that frozen `space_id`.
2. `ukPrincipal.SpaceID()` returns that frozen value; `RequiresSpaceScope()==true`, so global search runs `buildAllowlist(keyUID, frozenSpaceID)` → `enumerateDMPeers`.
3. `enumerateDMPeers` → `queryDMSpaceMemberUIDs` selects **every** active member of `space_id` (`WHERE space_id=? AND status=1 AND uid<>?`) — it excludes only the caller's own row, and **never verifies the caller is still a member**.
4. Global search trusts the allowlist directly (`scope ⊆ allowSet`; no per-peer `checkP2PAccess` on the global path), so those peers' DM history surfaces.

The comment in `enumerateDMPeers` ("Space containment already implied by the caller's Space middleware, which would have rejected the request otherwise") documents exactly the invariant the uk route breaks.

- **Repro:** user U mints a uk key while in Space X (key row `space_id=X`, active) → U is later removed from Space X but the key is not revoked → U calls the uk global search endpoint → `authUserAPIKey` passes, principal carries `spaceID=X` → `enumerateDMPeers` returns all *current* Space-X members → U searches DM history with Space-X users. The human web route 403s at the membership step; the uk route returns results.
- **Scope:** global **DM** enumeration only. Groups/threads are safe (live `GetGroupsWithMemberUID(keyUID)` + `ExistMembersActive` gate); the single-channel DM gate partly self-heals (`checkP2PAccess` same-Space branch uses `AreSpaceMembers`, falling back to `IsFriend`). The gap is specifically that global DM enumeration does not re-validate the caller's own current Space membership.
- **Load-bearing `space-isolation` (P92) violated → Spec gate ❌.**
- **Fix:** before constructing the `uk` principal (or as an equivalent gate on the uk route), verify the key owner is a current active member of `api_key_space_id` (mirror `CheckMembership`); on non-membership, reject or drop to `spaceID=""` (friends-only). Add a regression test: valid key + owner removed from the frozen Space → no Space-member DMs enumerated.

## Non-blocking
- 🟡 **OBO global per-channel DB amplification** — `filterChannelsByOBO` runs `SearchOBOAllowed` (grant+scope+live-access) per channel; large allowlists multiply DB round-trips. Rate-limiter backstops it. Batch-load + in-memory intersect. (Also raised by Jerry-Xin/yujiawei.)
- 🟡 **OBO entry rejection** returns the bot-unavailable error rather than a uniform NOT_FOUND on some paths — worth aligning to existence-hiding, per yujiawei; the per-channel gate already returns NOT_FOUND, so this is entry-consistency polish.

## Verified clean (still holds from the first pass)
- **No IDOR in OBO** — grant keyed on `(grantor, botUID)` via `findActiveGrantByGrantorBot`; a forged `on_behalf_of` can't impersonate an arbitrary user. TOCTOU re-checks match the send path.
- **App Bot fail-closed** at all three construction points.
- **as-bot** correctly bounded to the bot's own friends + active groups/threads (not global Space members) — the bot has no `space_member` row so the leaky branch never fires for it.
- **Unified predicate** (single-channel ⇔ allowlist) holds.
- (My earlier P2 import-cycle finding under `-tags integration` still stands as non-blocking.)

## Verdict
**REQUEST_CHANGES** — a valid P1: the new `uk` route drops the Space-membership gate that the human route enforces, and global DM enumeration trusts a frozen `space_id` without re-validating the caller's current membership → cross-tenant DM-history leak on stale keys. This violates the load-bearing space-isolation rule (Spec ❌). The OBO / as-bot / App-Bot models are otherwise sound.
Suggested decision: `Merge after resolving the P1` — add the uk current-membership gate + regression test.

*Correction note: my prior COMMENT under-called this — I verified the group/thread live gates but did not trace that the global DM enumeration path skips the caller's own membership recheck once `SpaceMiddleware` is removed. Lesson: a "deliberately no middleware" design comment is a prompt to byte-trace what backstops the removed gate, not a reason to assume something does.*

yujiawei (changes requested):
# Code Review — PR #608 (octo-server) — Revised

**Correcting my earlier approval.** After a cross-file re-check I confirmed a blocking authorization gap on the newly-added User API-key (uk) global DM search path that my first pass missed. Reviewed at head `93f531d`, merge-base `c1530e5`. Build / vet / `messages_search` unit suite are green locally, but green tests do not cover the gap below.

**Verdict: CHANGES_REQUESTED** (one P1 authz bypass).

## 🔴 P1 — Stale uk key bypasses current Space-membership on global DM search

The new uk route `modules/botfather/search_route.go` mounts `/v1/user/messages` with `authUserAPIKey → resolveUKPrincipal` and **no `SpaceMiddleware`**. `authUserAPIKey` (`modules/botfather/api_user.go:34`) validates only key status, integration-client enablement, and active-user — it does **not** re-check that the key owner is still a member of the key's frozen `space_id`. `authenticateUK` (`modules/messages_search/principal.go:275`) then builds the principal with that stored `api_key_space_id`.

The human web route, by contrast, runs `pkg/space/middleware.go` `SpaceMiddleware` → `CheckMembership(spaceID, uid)`, which 403s a caller who has left the Space (`middleware.go:135-147`). The uk route has no equivalent gate, so a frozen `space_id` is trusted indefinitely.

Impact is the **global DM allowlist**: `buildAllowlist` → `enumerateDMPeers` (`modules/messages_search/search_global.go:972`) unions the caller's friends with **every active member** of the stored Space via `queryDMSpaceMemberUIDs` (`search_global.go:1166`: `SELECT uid FROM space_member WHERE space_id=? AND status=1 AND uid<>?`) — with **no check that `loginUID` is still in that Space**. Global search trusts the allowlist directly and does *not* re-run per-peer `checkP2PAccess`, so the resulting `terms(channelId)` set surfaces DM history with all current Space members.

Reproducer: user U mints a `uk_` key while a member of Space X (key row: `space_id=X`, `status=1`); U is later removed from Space X but the key is not revoked; U calls `POST /v1/user/messages/_search/global` with that key. `authUserAPIKey` passes, the uk principal carries `spaceID=X`, and `enumerateDMPeers` returns all current Space-X members — letting U search DM history with Space-X users the human route would 403.

Verified independently: groups/threads are **not** affected (they pass `GetGroupsWithMemberUID(keyUID)` + `ExistMembersActive(keyUID)`, a live gate); the single-channel DM gate is partially self-correcting (`checkP2PAccess` same-Space branch uses `AreSpaceMembers(spaceID, loginUID, peer)`, which a removed owner fails, falling back to `IsFriend`). The gap is specifically the **global DM enumeration**, which never re-validates the caller's own Space membership.

**Fix:** validate the key owner's current active membership in `api_key_space_id` before constructing the uk principal (or apply a membership-check equivalent to `SpaceMiddleware` on the uk route), and add a regression test for an active key whose owner has been removed/disabled in that Space.

## Spec compliance (otherwise sound)

The as-bot and OBO models correctly implement the linked issue: friends-only as-bot P2P (no blacklist), `ExistMemberActive(botUID)` group/thread gate, `friends ∪ groups ∪ threads` global allowlist; OBO grant + scope + live `grantorCanReadChannel` (TOCTOU) reusing the send-path checker; App Bot fail-closed at principal construction on both branches; rate limit + audit keyed by `botUID`/`grantorUID`; `nil` OBO checker fails closed; `botUID == grantorUID` self-grant rejected in three places. Single-channel `canReadChannel` and global `enumerateReadableChannels` share one per-principal predicate. The P1 above is the one place the new uk surface diverges from the human route it is meant to mirror.

## Non-blocking (P2)

1. **OBO global search fans out one un-batched `checkOBO` per channel** (`modules/messages_search/obo.go:878` `filterChannelsByOBO`). Each iteration issues ~3–4 non-batched MySQL queries and re-fetches the invariant grant every loop, before the request `context.WithTimeout` is set. Hoist the grant lookup and/or add a batch seam. Bounded by the UID rate limiter; not a blocker.
2. **OBO entry-gate denials return 403, not the anti-enumeration 404** used per-channel (`modules/bot_api/search_route.go`). Minor grant-validity oracle; exposure bounded (a bot can read its own grant via `/v1/bot/obo-grant`).
3. **Thread hard-caps** (`enumerateThreadsForGroups`, `search_global.go`) can drop threads from the global allowlist that single-channel search still admits, at the 200/2000 boundary — pre-existing, shared with the human path, and fails closed (omits, never over-shares).
4. **`authenticateOBOFromContext`** (`principal.go:261`) reads an unvalidated `spacepkg.GetSpaceID(c)` and is **dead code** (live path uses `NewOBOPrincipal(bot, grantor, "")`). Remove it or guard it — it is a latent footgun that would reintroduce an attacker-controlled spaceID if wired.

## Coverage / blind spots

- Callee semantics verified against source (not comments): `ExistMemberActive(s)` (is_deleted=0 AND status=Normal), `IsFriend`/`GetFriends`, `QueryPeerRobotInfo`, thread enumeration, `grantorCanReadChannel`. App-Bot rejection keys on `bot_kind` set unconditionally by `authBot`.
- An automated cross-family pass flagged a repo-wide compilation failure from exported/unexported identifier drift — that was a **false positive** from reading the diff in isolation; the exported wrappers live in `route_subtree.go` and `go build` compiles clean. Recorded so it is not treated as unresolved.
- Not verifiable here: runtime OpenSearch mappings (`visibles`/`spaceId`), deployed config (`OCTO_SEARCH_REQUIRE_SPACE_ID`), live DB, and the `bot_api` MySQL/Redis integration tests.

lml2468 (approved):
# PR Review (RE-REVIEW @ new head) — PR #608 (Mininglamp-OSS/octo-server)

Reviewed at head SHA `54ed2cae` against merge-base `c1530e5f` (base `main`).
**This supersedes my REQUEST_CHANGES (review 4724349228 @ `93f531da`).** New commit `54ed2cae "fix(botfather): enforce live Space membership on uk search route (YUJ-58)"` resolves the P1 space-isolation blocker.

**Build/vet/test @ new head:** `go build ./...` exit 0 ✅, `go vet ./modules/botfather/ ./modules/messages_search/` exit 0 ✅, `go test ./modules/botfather/` ✅, `go test ./modules/messages_search/` ✅.

## Delta scope (byte-verified `93f531da` → `54ed2cae`)
Two files, both `modules/botfather/` (route + its test). **No `messages_search` production file changed** — so every prior byte-verified-clean authz finding (OBO no-IDOR, App-Bot fail-closed, as-bot bounding, unified predicate) stands unchanged.

## Prior P1 — RESOLVED
**🔴 uk stale-key Space-membership bypass → cross-tenant DM leak — FIXED.** `resolveUKPrincipal` now calls `spacepkg.CheckMembership(db, api_key_space_id, keyUID)` before constructing the uk principal:
- `CheckMembership` is the **same query the human `SpaceMiddleware` uses** — `INNER JOIN space s … s.status=1 … sm.status=1` (active space + active member). Verified byte-identical semantics.
- Non-member → **403** (`ErrSharedForbidden`), aborts before the principal is set → never reaches `enumerateDMPeers`. Membership-check error → fail-closed `ErrSharedInternal`. The stale-key repro (owner removed from frozen Space X → global DM search) now 403s, matching the human route.
- Correctly scoped: only runs when `spaceID != ""` (a spaceless uk key has no membership to check and is handled by the downstream `RequiresSpaceScope` empty-Space fail-close gate) — mirrors the human middleware's "no space_id → skip" behavior.
- Normalized: the gate lives at principal construction, not inlined into `enumerateDMPeers` — so it doesn't fork the shared predicate.
- **Regression tests added** (`TestResolveUKPrincipal_*`): non-member fails closed (403), membership-error fails closed, spaceless key skips membership, member/happy path — all pass. Exactly the "valid key + owner removed from frozen Space" case I asked for.

## Non-blocking (carried forward)
- 🟡 **P2 import cycle under `-tags integration`** — still present (`bot_api → messages_search` edge closes `message`(integration test)→`app_bot`→`bot_api`→`messages_search`→`message`). Production build + default CI gate unaffected (CI intentionally doesn't run `-tags integration ./...` per #557). Non-blocking, but a real regression from a clean merge-base — worth a fast-follow before that lane is provisioned.
- 🟡 **OBO per-channel DB amplification** (`filterChannelsByOBO`) — batch-load + in-memory intersect. Rate-limiter backstops. (Raised by multiple reviewers.)
- 🟡 **OBO entry rejection code** — align to uniform NOT_FOUND for existence-hiding (per yujiawei); per-channel gate already does this.

## Verified clean (re-confirmed; unchanged by the delta)
No IDOR in OBO (grant keyed on `(grantor, botUID)`, TOCTOU matches send path); App Bot fail-closed at all three construction points; as-bot bounded to its own friends + active groups/threads; single-channel gate ⇔ global allowlist (unified predicate); spaceless-bot global search allowlist-bounded.

## Verdict
**APPROVE** — the P1 space-isolation blocker (uk stale-key cross-tenant DM leak) is fixed exactly as prescribed: live `CheckMembership` gate at uk principal construction, same semantics as the human `SpaceMiddleware`, fail-closed on non-member/error, with full regression coverage. Both gates now pass; the load-bearing space-isolation rule holds on every principal path.
Suggested decision: `Safe to merge after non-blocking follow-up` — the P2 `-tags integration` import cycle should be cleaned up (or explicitly deferred by a maintainer) before that lane is enabled, but it does not block today's CI gate. Actual merge still awaits branch-protection CI.

Jerry-Xin (approved):
# PR Review (RE-REVIEW @ new head 54ed2cae) — PR #608

**This supersedes my REQUEST_CHANGES (review 4724261846 @ `93f531da`).** New commit `54ed2cae "fix(botfather): enforce live Space membership on uk search route (YUJ-58)"` resolves the sole P1 blocker. Byte-verified `93f531da → 54ed2cae`: `ahead 1 / behind 0` (clean fast-forward, no merge-base drift), touching exactly two files, both under `modules/botfather/` — `search_route.go` (+45) and `search_route_test.go` (+97/-2). No `messages_search` production file changed.

## Prior P1 — RESOLVED ✅

**🔴 → ✅ Stale uk key bypassed current Space-membership on global DM search — FIXED.**

`resolveUKPrincipal` now delegates to `resolveUKPrincipalWithChecker`, which runs a live membership gate **before** the uk principal is set into context:

- For the uk principal, `SubjectUID()` returns the key owner (`api_key_uid`) and `SpaceID()` returns the frozen `api_key_space_id`. The gate calls `spacepkg.CheckMembership(db, api_key_space_id, keyOwnerUID)` — the **same query the human `SpaceMiddleware` uses**: `INNER JOIN space s ON s.status=1 ... WHERE sm.uid=? AND sm.space_id=? AND sm.status=1` (active Space + active member). Byte-identical semantics confirmed against `pkg/space/membership.go` and `pkg/space/middleware.go`.
- **Fail-closed on non-member:** returns `403 ErrSharedForbidden` + `c.Abort()`, so it never reaches `SetPrincipal`/`c.Next()` and never enters `enumerateDMPeers`. The removed-owner reproducer (key minted in Space X → owner removed from X → key not revoked → global DM search) now 403s, matching the human route.
- **Fail-closed on lookup error:** returns `500 ErrSharedInternal` + `c.Abort()` — does not fall through to a resolved principal.
- **Correctly scoped:** the gate runs only when `spaceID != ""`. A spaceless uk key has no membership to check and is handled downstream by the `RequiresSpaceScope()=true` empty-Space fail-close — mirroring the human middleware's "no space_id → skip" behavior.
- **Normalized:** the check lives at principal construction, not inlined into `enumerateDMPeers`, so the shared DM-allowlist predicate is not forked.

**Regression tests added** (`search_route_test.go`), all via an injectable `spacepkg.MembershipChecker` (DB-free): `TestResolveUKPrincipal_NonMemberFailsClosed` (403, no principal set), `TestResolveUKPrincipal_MembershipErrorFailsClosed` (500, no principal), `TestResolveUKPrincipal_NoSpaceSkipsMembership` (gate skipped, uk principal set), and the happy path asserting the gate is invoked when `api_key_space_id` is present. This is exactly the "valid key + owner removed from frozen Space" case requested in the prior round.

## No new regression from the fix

The delta touches only the uk route + its test. `AuthenticateUK`/`SetPrincipal` exported wrappers (`route_subtree.go`) and `errcode.ErrSharedForbidden`/`ErrSharedInternal` (`pkg/errcode/integration.go`) all resolve; groups/threads path, OBO/as-bot models, and the single-channel DM gate are unchanged. The load-bearing Space-isolation rule now holds on every principal path (human, bot, OBO, and uk).

## Non-blocking (carried forward, not gating)

- 🟡 **OBO per-channel DB amplification** in `filterChannelsByOBO`: sequential `checkOBO` per channel re-fetches the invariant grant each loop. Hoist the grant lookup / add a batch seam. Bounded by the shared-UID rate limiter. (Raised by multiple reviewers previously; unchanged here.)
- 🟡 **OBO entry-gate denial code**: consider aligning to the anti-enumeration `404` used per-channel, rather than `403`, for uniform existence-hiding.
- 🟡 **`-tags integration` import cycle** (`bot_api → messages_search`) noted by a peer: production build + default CI gate unaffected; worth a fast-follow before that lane is provisioned.

## Highlights ✅

- Fix mirrors the human `SpaceMiddleware` semantics exactly, fail-closed on both non-member and lookup error.
- Gate placed at principal construction (predicate consistency preserved), with a clean injectable seam for DB-free unit tests.
- Full regression coverage for the removed-owner scenario.

## Verdict

**APPROVE** — the sole P1 (uk stale-key cross-tenant DM-enumeration bypass) is fixed as prescribed: a live `CheckMembership` gate at uk principal construction with the same semantics as the human route, fail-closed on non-member (403) and error (500), plus complete regression tests. Remaining items are non-blocking. Actual merge still awaits branch-protection CI.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Octo-Q automated review Review — PR #608
**Head SHA:** `54ed2cae6629c437572891d1c147148c76c0bf11`
**Repo:** `Mininglamp-OSS/octo-server`
**Title:** feat(messages_search): support octo-cli message search as bot / as-user (OBO)
**Files changed:** 56 | **+2741 / −402** (net +2339)

---

## 1. 验证结论

### ✅ Principal 抽象层设计正确

- `Principal` interface 定义于 `principal.go`，4 种实现（`userPrincipal`、`userBotPrincipal`、`oboPrincipal`、`ukPrincipal`）各自语义清晰。
- `Kind()`、`SubjectUID()`、`SpaceID()`、`RequiresSpaceScope()`、`BlacklistPolicy()`、`RateLimitKey()`、`AuditBotUID()`、`AuditGrantorUID()` 行为符合设计。
- 共享 singleton `Handler` 通过 `sync.Once` 保证，`Shared()` 与 `New()` 分离正确（`1module.go:Shared`）。

### ✅ OBO 授权链路完整

- `SearchOBOAllowed`（`bot_api/obo_check.go`）复用 `checkOBO` 的完整 grant + scope + TOCTOU 逻辑，搜索与发送路径授权一致。
- `oboCanReadChannel`（`obo.go:37-58`）执行 grantor 真实用户权限检查 ∩ OBO scope，fail-closed on infra error。
- `errOBONoChecker`（`obo.go:19-21`）：checker 未接线时 fail-close，防止绕过。

### ✅ canReadChannel dispatch 正确

- `predicate.go:canReadChannel`：`userBot → botCanReadChannel`、`OBO → oboCanReadChannel`、default → `checkChannelAccess`。
- `botCanReadChannel`（`predicate.go:90-127`）：DM→IsFriend、group→ExistMemberActive、thread→parent group membership。语义与 web 用户路径区分正确（无 Space、无 blacklist）。

### ✅ GetLoginUID() 迁移完整

所有生产 handler（`search_all.go`、`search_around.go`、`search_files.go`、`search_media.go`、`search_messages.go`、`search_global_messages.go`、`search_global_files.go`、`search_global_groups.go`）已统一切换为 `h.principal(c).SubjectUID()`。`grep -rn "c.GetLoginUID()" modules/messages_search/` 仅在 `authenticateUser` 回退路径中出现（正确）。

### ✅ [automated review](mention://issue/8159ca9f-2f34-4750-850f-2e957e6c6609) space-less 旁路正确

- `resolveGlobalScope`（`search_global.go`）：`p.RequiresSpaceScope() == false` 时跳过 `RequireSpaceID` gate。
- `resolveP2PSpaceScope`（`space_scope.go`）：`RequiresSpaceScope() == false` 时返回 `("", true)`。
- 仅 bot/OBO 触发此旁路，uk/user 仍强制 space scope。

### ✅ Rate limit + Audit 正确

- `ratelimit.go`：`SharedUIDRateLimiter` 用于 bot 身份固定，`searchRateLimiter` 使用 `principal.RateLimitKey()` 分桶。
- `audit.go`：`SubjectUID()` 作为 `login_uid`，as-bot/OBO 额外记 `bot_uid` + `grantor_uid`。

### ✅ 路由接线正确

- `bot_api/bot_api.go`：`searchHandler = messages_search.Shared(ctx)` + `mountSearchRoutes(r)` 挂载 `/v1/bot/messages/_search*`。
- `botfather/api.go`：同 `Shared(ctx)` + 挂载 `/v1/user/messages/_search*`。
- `bot_api/search_route.go` 中间件链：`authBot → botActorUID → SharedUIDRateLimiter → resolveSearchPrincipal → searchRateLimiter → audit → backendGate`。
- `botfather/search_route.go` 中间件链：`authUserAPIKey → resolveUKPrincipal (automated review membership check) → searchRateLimiter → audit → backendGate`。

### ⚠️ OBO 枚举路径性能可接受但存在 N² 特征

见 Finding F1。

---

## 2. 发现问题

### F1 — P2 | OBO 全局枚举 O(n) DB 调用，grant 查询未复用

**位置：** `modules/messages_search/obo.go:80-120`（`oboEnumerateReadableChannels`）
**Diff-scope 三问：** new（本 PR 新增）；非既有行为放大。

**现象：** `oboEnumerateReadableChannels` 对 grantor allowlist 中每个 channel 调用 `oboCanReadChannel` → `SearchOBOAllowed` → `checkOBO`。每次 `checkOBO` 都独立调用 `findActiveGrantByGrantorBot(grantor, botUID)`（同一 grant 行被重复查询 N 次）+ `scopeEnabled` + `grantorCanReadChannel`。对 N 个 channel 产生 ~3–4N 次 DB 调用。

**触发条件：** OBO principal 执行不带 `channel_ids` filter 的全局搜索（`_search_global_messages` 或 `_search_global_files`），且 grantor 拥有大量可读 channel。

**影响：** 对 grantor 有 100 个 group 的场景，产生 300–400 次 DB round-trip。搜索请求 latency 显著上升，但不影响正确性（授权结论不变）。

**建议：** 在 `oboEnumerateReadableChannels` 入口处一次性查询 grant，传入后续循环；或为 `checkOBO` 提供一个接受 pre-loaded grant 的内部变体。这是性能优化，不阻塞。

**定级：** P2（latency degradation，非用户可见错误或权限绕过）。

---

### F2 — P2 | space welcome message 系统设置 schema 变更无迁移

**位置：** `modules/common/api_manager_system_setting.go:315-318`、`modules/notify/space_welcome.go`
**Diff-scope 三问：** new（本 PR 变更 schema）；非既有行为放大。

**现象：** 系统设置键从 `space_welcome_message_zh_cn` + `space_welcome_message_en_us`（双语）改为单一 `space_welcome_message`。`SpaceWelcomeConfig()` 现在只读 `space_welcome_message`，旧双语键不再被读取。

**影响：** 已部署环境中存在旧双语键的实例，升级后 `SpaceWelcomeConfig().Message` 将为空字符串，欢迎语功能静默失效（不报错，但不发送欢迎语或发送空白消息）。

**建议：** 在 `api_manager_system_setting.go` 的 partial update 逻辑中添加一次性迁移：若 `space_welcome_message` 为空但 `space_welcome_message_zh_cn` 存在，自动将中文值写入新键。或在 release notes 中明确标注此 breaking change 及手动迁移步骤。

**定级：** P2（部署风险，非安全漏洞）。

---

### F3 — Nit | `dispatchSingleAll` 传递空 spaceID 给 `principalForSubject`

**位置：** `modules/messages_search/search_global_messages.go:261`
**Diff-scope 三问：** pre-existing pattern（`principalForSubject` 在 `search_global.go:1260` 已定义）；本 PR 未改变其行为契约。

**现象：** `dispatchSingleAll` 调用 `principalForSubject(c, loginUID, "")` 传空 spaceID。对 bot/OBO 路径（principal 已在 context 中）无影响——`principalForSubject` 优先从 context 取真实 principal。对 web 路径，创建 `userPrincipal{spaceID: ""}` 但 `checkChannelAccess` 内部通过 `h.principal(c)` 重新获取正确 spaceID，不直接使用传入的 spaceID。

**影响：** 无功能影响，但代码意图不清晰。

**建议：** 可考虑直接传 `h.principal(c)` 而非重建 synthetic principal，或在注释中说明 spaceID="" 的意图。不阻塞。

**定级：** Nit。

---

### F4 — Nit | `h.principal(c)` 对 web 路由不缓存到 context

**位置：** `modules/messages_search/principal.go:241-252`
**Diff-scope 三问：** pre-existing（`authenticateUser` 不调用 `setPrincipalInContext`）；本 PR 未改变此行为。

**现象：** `h.principal(c)` 对 bot/OBO/uk 路由通过 context 缓存（`setPrincipalInContext`），但对 web 路由每次调用 `authenticateUser` 创建新 `userPrincipal`。web 路径中多次调用 `h.principal(c)` 会重复创建对象。

**影响：** 微小性能开销（一次请求中 2-3 次调用），不影响正确性。

**建议：** 在 `authenticateUser` 末尾调用 `setPrincipalInContext(c, p)` 使 web 路径也享受缓存。不阻塞。

**定级：** Nit。

---

## 3. 建议

1. **（F1 优化）** 为 `oboEnumerateReadableChannels` 提供 grant 预加载，避免 N 次重复 grant 查询。
2. **（F2 迁移）** 为 `space_welcome_message` schema 变更提供数据迁移或在 release notes 明确标注。
3. **（F3/F4 清理）** 低优先级，可在后续 PR 处理。

---

## 4. 额外发现

### notify/cardtmpl 变更（docs approval card enrich）

- `BuildDocsAccessRequestCard` 签名变更（`ResourceCard` → `DocsApprovalContent`），所有调用方已同步更新（`notify/card.go`）。
- `BuildDocsApprovalOutcomeCard` 新增，用于 terminal（已允许/已拒绝）卡片。`action_finalizer.go` 正确调用。
- `DocsDenyReasonInputID` 作为 hidden input 声明，与 `cardmsg.ValidateInputs` 的 declared-input 要求一致。✓
- `requireHTTPS` 检查 avatar URL，防止 SSRF。✓
- 新增 i18n labels（zh-CN + en）完整覆盖。✓

### common system_settings 变更

- `space_welcome_message` 替代双语键（F2）。`ValidateSpaceWelcomeCombination` 验证逻辑已同步更新。
- `TestSpaceWelcomeConfig_MultilineMessagePreserved` 锁定多行消息行为。✓

### scripts/cardpreview/main.go

- 未审查（非安全关键路径，为开发辅助工具）。

---

## 5. 数据流回溯

### 5.1 — OBO principal 数据流

```
请求体 on_behalf_of
  → parseSearchOnBehalfOf (bot_api/search_route.go:98-133)
    → json.Unmarshal 提取 grantorUID
    → validateSearchOBO: grant+scope 存在性预检
  → resolveSearchPrincipal
    → NewOBOPrincipal(botUID, grantorUID)
    → setPrincipalInContext
  → h.principal(c) → oboPrincipal
  → SubjectUID() = grantorUID  (搜索以 grantor 身份执行)
  → RateLimitKey() = "obo:" + botUID  (限流按 bot 分桶)
  → canReadChannel → oboCanReadChannel
    → checkChannelAccess (grantor 真实用户权限)
    → ∩ oboCheck (grant + scope + TOCTOU)
  → OS query: terms(channelId, allowlist)  (allowlist = grantor 可读 ∩ OBO scope)
```

**验证：** 每个消费点的数据都回溯到正确的上游来源。grantorUID 来自请求体但经过 validateSearchOBO 预检和 per-channel TOCTOU 检查。✓

### 5.2 — Bot principal 数据流

```
authBot 中间件 → botUID + botActorUID 设置到 context
  → resolveSearchPrincipal (on_behalf_of 为空)
    → NewUserBotPrincipal(botUID, botActorUID)
  → SubjectUID() = botUID
  → RateLimitKey() = "bot:" + botUID  (SharedUIDRateLimiter 固定桶)
  → canReadChannel → botCanReadChannel
    → DM: IsFriend(botUID, peer)
    → Group: ExistMemberActive(botUID, groupID)
    → Thread: parent group membership
  → buildBotAllowlist: getBotDMChannels + ExistMembersActive + threads
```

**验证：** botUID 来自 authBot 中间件（可信），botActorUID 来自 `x-bot-actor-uid` header（bot 自行声明，用于 rate limit 区分）。限流桶固定为 botUID 防止通过切换 actorUID 绕过。✓

### 5.3 — UK principal 数据流

```
authUserAPIKey → keyModel (space_id, uid, key_value)
  → resolveUKPrincipal
    → spacepkg.CheckMembership(keyModel.SpaceID, keyModel.UID) ([automated review](mention://issue/443b8ee4-9915-432c-8cf6-ede38b4d3a2b))
    → 非成员 → 403 Forbidden
    → 成员 → NewUKPrincipal(uid, spaceID)
  → SubjectUID() = keyModel.UID
  → RequiresSpaceScope() = true  (强制 space scope)
```

**验证：** UK 路径强制 Space membership 实时检查（[automated review](mention://issue/443b8ee4-9915-432c-8cf6-ede38b4d3a2b)），防止 key 对应的用户已离开 Space 但仍可搜索。✓

---

## 6. 盲点 Checklist

### C1 — 双路径 parity ✅ Clear

**触发：** PR 创建成对路径 user↔bot、as-bot↔as-user(OBO)。

**验证：**
- `canReadChannel` dispatch 覆盖所有 4 种 principal kind：userBot→botCanReadChannel、OBO→oboCanReadChannel、user/uk→checkChannelAccess。✓
- `enumerateReadableChannels` dispatch 覆盖：userBot→botEnumerateReadableChannels、OBO→oboEnumerateReadableChannels、user/uk→buildAllowlist。✓
- Rate limit key 区分：user=loginUID、userBot=botUID（固定）、OBO=botUID（固定）、uk=keyUID。✓
- Space scope 要求：user/uk=true、userBot/OBO=false。✓

### C2 — control-flow ordering / 嵌套复用 ✅ Clear

**触发：** `dispatchSingleAll` / `dispatchSingleFiles` 复用单 channel handler。

**验证：**
- `dispatchSingleAll` 在调用 `searchAllForGlobalFastPath` 前显式运行 `canReadChannel`（`search_global_messages.go:261-263`），确保 per-principal gate 覆盖 fast path。✓
- `dispatchSingleFiles` 同样显式运行 `canReadChannel`（`search_global_files.go:65-68`）。✓
- `searchAllForGlobalFastPath` 内部不再重复 `canReadChannel`（由调用方保证），避免 double-check。✓

### C3 — 授权边界 ≠ 能力边界 ✅ Clear

**触发：** 新增 `/v1/bot/messages/_search*` 和 `/v1/user/messages/_search*` 端点。

**验证：**
- Bot 路由：`authBot()` 中间件验证 bot token → `resolveSearchPrincipal` 解析 OBO → per-channel scope 检查。App Bot 在 `resolveSearchPrincipal` 中被 reject（`search_route.go:68-72`）。✓
- UK 路由：`authUserAPIKey()` → `resolveUKPrincipal` 强制 Space membership ([automated review](mention://issue/443b8ee4-9915-432c-8cf6-ede38b4d3a2b))。✓
- OBO 搜索能力不超过 grantor 本人搜索能力（`oboCanReadChannel` 先 check grantor access，再 ∩ OBO scope）。✓

### C4 — 授权生命周期 / 容器-成员状态级联 ✅ N/A

**触发：** 无鉴权/封禁改动直接涉及容器-成员状态级联。

**验证：** OBO TOCTOU 检查（`checkOBO` 中的 `grantorCanReadChannel`）已覆盖 grantor 实时 channel 访问状态。若 grantor 被踢出 group / un-friend，搜索即刻失效。✓

### C5 — build/note 通过 ≠ 运行期路径正确 ✅ N/A

**触发：** 无 build/打包/扩展相关改动。

### C6 — 治理/策略/安全文档自洽性 ✅ N/A

**触发：** 无治理/策略文档改动。

---

## 7. 跨轮 Blocker 复检

N/A — 首次审查。

---

## 结论

本 PR 实现了一个完整的 Principal 抽象层，将消息搜索从单一用户身份扩展为支持 bot（as-bot）、OBO（as-user）和 UK（API key）四种身份。架构设计清晰，授权链路完整，fail-closed 策略一致，数据流回溯验证通过。

主要关注点是 OBO 全局枚举的 DB 调用效率（F1）和 space welcome schema 迁移（F2），均为 P2 级别，不影响正确性或安全性。

**无 P0/P1 发现。**

---

[Octo-Q] verdict: **APPROVE** — 仅 P2/Nit 发现，无阻塞级问题。Principal 抽象层设计正确，OBO 授权链路完整，所有搜索 handler 的 GetLoginUID() 迁移已完成。建议后续优化 F1（OBO 枚举 grant 预加载）和 F2（space welcome schema 迁移文档）。

yujiawei (approved):
# Code Review — PR #608 (octo-server)

Re-review at head SHA `54ed2cae` against merge-base `c1530e5f` (base `main`). This supersedes my prior `CHANGES_REQUESTED` at head `93f531da`. Local: `go build ./modules/messages_search/... ./modules/bot_api/... ./modules/botfather/...` exit 0 ✅, `go vet` (all three) exit 0 ✅, `go test ./modules/messages_search/` ✅, `go test ./modules/bot_api/ -run 'OBO|Search|Principal'` ✅, and the new `botfather` uk-route unit tests (`TestResolveUKPrincipal_*`) ✅. (The full `botfather` suite panics in `NewTestServer`/`setupGroupTestEnv` — infra-dependent, matches the PR's own note that MySQL/Redis-backed integration tests were not run.)

**Verdict: APPROVED.** The one P1 that blocked my previous pass is resolved; no P0/P1 remains. Two low-severity (P2) items below are non-blocking.

## 1. Spec compliance

**Spec: ✅**

- **as-bot** (`user_bot`): friends-only P2P via `botCheckP2PAccess` → `IsFriend(botUID, peer)`, no Space segment, no P2P blacklist (`blacklistPolicy=none`); group/thread via `ExistMemberActive(subject=botUID)`; global allowlist = `friends ∪ active groups ∪ threads` (`buildBotAllowlist`). Matches the linked issue exactly, no over-build.
- **as-user (OBO)**: entry grant-existence fail-fast (`validateSearchOBO` → `findActiveGrantByGrantorBot`, `active=1 AND global_enabled=1`), then per-channel `grant + scope + grantorCanReadChannel` (TOCTOU) reusing the send-side `checkOBO` via the `SearchOBOAllowed` seam. Single-channel gate (`oboCanReadChannel`) and global feed (`oboEnumerateReadableChannels`) share one predicate.
- **App Bot denied** at principal construction on both branches (`authenticateUserBot` / `resolveSearchPrincipal` OBO branch key on `bot_kind == "app"`, set unconditionally by `authBot`). No App Bot can construct a `userBotPrincipal`/`oboPrincipal`.
- **uk**: direct real-user identity, subject = `keyModel.UID`, no OBO scope narrowing — reuses the existing human `checkChannelAccess` / `buildAllowlist`.
- No missing / extra / divergent behavior relative to the linked scope.

## 2. Code quality

**Quality: Approved**

### Previously-blocking P1 — RESOLVED

My earlier block was: a stale `uk_` key bypassed live Space-membership on the global DM allowlist (`enumerateDMPeers` unions all `space_member WHERE space_id=? AND status=1` for the frozen `api_key_space_id`, without re-checking the caller is still in that Space — the human route enforces this via `SpaceMiddleware`).

`54ed2cae` fixes this precisely (`modules/botfather/search_route.go`, `resolveUKPrincipalWithChecker`): when the key carries a Space, it now runs `spacepkg.CheckMembership(space_id, keyUID)` and returns **403** for a removed/disabled member, **fail-closes to 500** on lookup error, and skips only when the key has no Space (mirroring `SpaceMiddleware`'s "no space_id → skip"). Verified the checker is the same `pkg/space` membership predicate the human route uses, and the gate lands in the uk auth path (not inlined inside `enumerateDMPeers`), preserving predicate normalization. New regression tests cover non-member (403), lookup-error (fail-closed), and no-Space (skip) — all pass locally.

### Verified fail-closed / correctness points

- `nil` OBO checker → `errOBONoChecker` (never degrades to un-scoped grantor search).
- `botUID == grantorUID` self-grant rejected in three places (`authenticateOBO`, `validateSearchOBO`, `checkOBO`).
- Spaceless relaxation (`RequiresSpaceScope`) is correctly gated: `user`/`uk` still fail-close on empty Space when `RequireSpaceID=true`; only `user_bot`/`obo` (which structurally have no Space) proceed, and their DM reachability is bounded by the per-principal predicate (`IsFriend` / grantor allowlist), not a `spaceId` term. An empty bot allowlist yields `ok=true` with empty `osChannelIDs` → handlers return an empty envelope (no unbounded scan).
- Subject-UID is consistent across all global paths (`SubjectUID()` everywhere; OBO subject = grantorUID, so DM fake-channel derivation, member scoping, and audit all agree). No leftover `c.GetLoginUID()` in the search handlers.
- Group/thread gates for the bot use `ExistMemberActive(botUID)`; the #354 owner-blacklist cascade flips an in-group bot to non-Normal, so group-level blacklist is inherited for free.

### P2 (non-blocking)

1. **OBO global search fans out one un-batched `checkOBO` per channel** (`modules/messages_search/obo.go`, `filterChannelsByOBO`): each iteration issues several non-batched MySQL queries and re-fetches the invariant grant, on the global hot path. Bounded by the per-UID rate limiter, so not a blocker, but a batch seam (or hoisting the grant lookup out of the loop) would remove an N+1 under concurrency.
2. **Malformed `on_behalf_of` silently falls back to as-bot** (`modules/bot_api/search_route.go`, `parseSearchOnBehalfOf`): invalid JSON or a wrong-typed `on_behalf_of` (e.g. a number) returns `""` and the request is treated as as-bot rather than 400. Low impact (the caller gets the bot's own viewpoint, still fully authorized), but a client bug could be masked. Consider rejecting a present-but-invalid field.

Both are pre-existing observations from my earlier pass; neither gates merge.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. The blocking uk Space-membership gap is fixed and regression-tested; remaining items are P2.

## 4. Suggestions

- Add a batch `FilterOBOChannels(bot, grantor, []channelRef)` seam so OBO global search does not issue per-channel grant/scope round-trips.
- Reject a present-but-unparseable `on_behalf_of` with a validation error instead of silently downgrading to as-bot.

## 5. Additional notes

- `authenticateOBOFromContext` (`principal.go`) remains dead code (the live path uses `NewOBOPrincipal(bot, grantor, "")`); it reads an unvalidated `GetSpaceID(c)`. Harmless today, but worth removing to avoid a future footgun that reintroduces an attacker-influenced spaceID.
- A cross-family automated pass flagged a repo-wide compile failure from exported/unexported identifier drift — **false positive** from reading the diff in isolation; the exported wrappers (`SetPrincipal`, `NewOBOPrincipal`, `AuthenticateUserBot`, `AuthenticateUK`) live in `route_subtree.go` and `go build` is clean. Recorded so it is not treated as unresolved.
- Not verifiable here: runtime OpenSearch mappings (`visibles`/`spaceId`), deployed config (`OCTO_SEARCH_REQUIRE_SPACE_ID`), live DB, and the `bot_api`/`botfather` MySQL/Redis integration tests.

PR #628, opened 2026-07-21
fix(message): stop leaking revoked message content on sync endpoints

an9xyz:
## Summary

Revoked messages were returned with `revoke=1` **and** their full original payload on `/v1/message/channel/sync` and `/v1/conversation/sync` (both share `MsgSyncResp.from`). The original text therefore stayed recoverable in the client's page memory even though the row is only rendered as a "message revoked" tip — a browser extension can read it straight off the React props. The single-message read path (`api_message_get.go`) already `404`s on `revoke==1`; this PR closes the same gap on the sync paths so revoked content is never shipped.

## Related Issue

None filed — reported directly as a data-leak on the two sync endpoints above.

## Linked Spec

No `.octospec/tasks/<slug>/brief.md` (direct security report, no pre-existing brief). Load-bearing path affected: **message sync response assembly** — `MsgSyncResp.from` (`modules/message/api.go`), shared by both sync endpoints, plus the `SpaceLastMessage` preview fallback in `modules/message/space_unread.go`.

## Changes

- **`MsgSyncResp.from` — revoke content redaction:** when `revoke==1`, strip the outgoing `payload` down to its `type` only, and clear the `signal_payload` (ciphertext), `streams` (stream blobs), and `message_extra.content_edit` (edited body) — all original-content carriers. Revoke metadata (`revoke`, `revoker`) is preserved so the tip still renders.
- **`/conversation/sync` `SpaceLastMessage` fallback:** `findSpaceLastMessageFallback` hand-built the preview via `msgRespToSyncResp`, bypassing `from()`'s redaction. It now skips revoked candidates (batch `queryRevokedWithMessageIDs`, mirroring the revoke filter in `api_channel_files.go`), so a revoked message is never surfaced as a last-message preview. Selection logic factored into a pure, unit-testable `selectSpaceLastMessage`.
- **Tests:** added `revoke_payload_strip_test.go` covering payload / `content_edit` / signal / stream stripping and revoked-preview skipping.

## Testing

```
go build ./modules/message/      # ok
go vet ./modules/message/        # ok
go test ./modules/message/ -run 'TestRevokedPayload|TestMsgSyncRespFrom_|TestSelectSpaceLastMessage_|TestRevokedMessageIDSet_'   # PASS
go test ./modules/message/ -run 'NoLegacyResponseError'   # PASS (source guard)
```

All new tests are pure unit tests (construct `config.MessageResp` + `messageExtraDetailModel`, call `from()` directly) — no MySQL/Redis/WuKongIM needed. Existing `space_unread` unit tests (incl. `TestFillPersonSpaceUnread_*`) still pass after the signature change.

- [ ] Manually verified end-to-end — the live sync path needs MySQL+Redis+WuKongIM (CI); verified here by exercising `from()` directly at unit level.

## COMPREHENSION

1. **What does this change actually do to the load-bearing path?** In the single funnel every synced message passes through (`MsgSyncResp.from`), it adds a post-assembly redaction step gated on `revoke==1`: the payload is replaced with `{type}` only and the other content carriers (signal ciphertext, stream blobs, `content_edit`) are cleared. It does **not** touch non-revoked messages, and it does not change which messages are returned — only the content of already-revoked ones. The conversation-sync preview fallback additionally skips revoked candidates instead of emitting them.

2. **What could break (dependents + failure mode)?** Any client that (incorrectly) reads a revoked message's `payload.content` instead of gating on `revoke=1`. Verified on octo-web: the revoke tip (`RevokeCell`) renders purely from `revoke` + `revoker` (revoker name resolved by UID via the channel manager), never from payload, and there is no "re-edit revoked text" feature — so no legitimate consumer loses data. Worst case for an odd client is an empty bubble on an already-revoked row (still gated away by `revoke=1`). The preview fallback change means a channel whose newest space message is revoked now previews the last *non-revoked* message (or nothing) instead of leaking the revoked text.

3. **How do you know it works?** `TestMsgSyncRespFrom_RevokedStripsContent` asserts the secret is absent from the entire marshaled response and only `{type}` remains; sibling tests cover `content_edit`, `signal_payload`, and `streams`; `TestMsgSyncRespFrom_NotRevokedKeepsContent` guards against over-redaction; `TestSelectSpaceLastMessage_SkipsRevoked` covers the preview-skip (skip newest-revoked, pick older; all-revoked → nil). Consistency anchor: `api_message_get.go:241` already returns `404` on `revoke==1`, so this only extends an existing, intended invariant to the sync paths.

## Checklist

- [ ] Updated documentation (not applicable — no doc-facing change)

---

### Related findings (not addressed here — same class, awaiting decision)

The identical sync funnel also ships full payloads for other not-visible-to-requester messages, and `api_message_get.go` already blocks these on the single-read path too:
- **Globally deleted** messages (`message_extra.is_deleted=1`) ship full payload with `is_deleted=1`.
- **`visibles`-restricted** messages ship full payload to non-whitelisted users (`api.go` only sets `is_deleted=1`, doesn't strip).
- **Reply-quote snapshots** embed a revoked/edited parent's content in the replying message's `reply` preview.

Scoped out of this PR because they touch delete/visibility/reply rendering paths whose frontend impact wasn't verified for this change. Happy to follow up under the same redaction contract.

https://claude.ai/code/session_01AupxRE4tVPQ4fS2bzi3msA

---
_Generated by [Claude Code](https://claude.ai/code/session_01AupxRE4tVPQ4fS2bzi3msA)_

### DIFF
--- a/modules/message/api.go
+++ b/modules/message/api.go
@@ -113,6 +113,27 @@ func placeholderPayload() map[string]interface{} {
 	}
 }
 
+// revokedPayload 返回撤回消息对外下发的最小 payload：仅保留原始 type，剥离 content
+// 及一切内容承载字段（url / name / reply / content.users …）。
+//
+// 背景（撤回原文泄漏）：Octo 撤回是 message_extra.revoke=1 的软删除，WuKongIM 里
+// 原始 payload 仍在。sync 路径此前把 revoke=1 与「完整原始 payload」一起下发，
+// 原文因此保留在客户端内存（message.content），恶意插件可从 React props 反查还原。
+// 单条直读 api_message_get.go 对 revoke==1 直接返回 404，本函数补上 sync 路径同口径的缺口。
+//
+// 保留 type 而不整条清空：撤回对所有人生效，前端仅凭 revoke=1 渲染撤回提示、不读
+// payload（撤回者名由 revoker UID 单独查，见 octo-web RevokeCell）；保留 type 只为
+// 兼容按 type 分支的老客户端渲染路径，type 本身不含消息正文。
+func revokedPayload(original map[string]interface{}) map[string]interface{} {
+	stripped := map[string]interface{}{}
+	if t, ok := original["type"]; ok {
+		stripped["type"] = t
+	} else {
+		stripped["type"] = common.ContentError.Int()
+	}
+	return stripped
+}
+
 // isTextType 判断 payload type 是否为 common.Text（=1）。兼容 json.Number / float64 / int
 // 几种反序列化结果；string 类型的 "1" 不识别为 Text，避免误命中。
 func isTextType(m map[string]interface{}) bool {
@@ -3315,6 +3336,23 @@ func (m *MsgSyncResp) from(msgResp *config.MessageResp, loginUID string, message
 			streams = append(streams, newStreamItemResp(streamItem))
 		}
 		m.Streams = streams
+	}
+
+	// 撤回消息内容脱敏：撤回对所有人生效，任何客户端都不该再拿到原文。剥离 payload
+	// 正文、加密 signal 密文与流式 blob，只保留 revoke=1 / revoker 供前端渲染撤回
+	// 提示。与单条直读 api_message_get.go 对 revoke==1 返回 404 同口径；这里补上
+	// /message/channel/sync 与 /conversation/sync（两者共用 from()）此前遗漏的缺口。
+	// 必须放在 Payload / SignalPayload / Streams / MessageExtra 全部赋值之后。
+	if m.Revoke == 1 {
+		m.Payload = revokedPayload(m.Payload)
+		m.SignalPayload = ""
+		m.Streams = nil
+		// message_extra.content_edit 是「编辑后的正文」，同样是原文载体：撤回后必须
+		// 一并剥离，否则编辑过的消息被撤回时仍会经 content_edit 把原文下发。
+		if m.MessageExtra != nil {
+			m.MessageExtra.ContentEdit = nil
+			m.MessageExtra.EditedAt = 0
+		}
 	}
 
 }
--- a/modules/message/api.go
+++ b/modules/message/api.go
@@ -1397,7 +1397,12 @@ func (m *Message) syncChannelMessage(c *wkhttp.Context) {
 	if len(channelSettings) > 0 && channelSettings[0].OffsetMessageSeq > 0 {
 		channelOffsetMessageSeq = channelSettings[0].OffsetMessageSeq
 	}
-	syncResp := newSyncChannelMessageResp(resp, c.GetLoginUID(), req.DeviceUUID, req.ChannelID, req.ChannelType, m.messageExtraDB, m.messageUserExtraDB, m.messageReactionDB, m.channelOffsetDB, m.deviceOffsetDB, channelOffsetMessageSeq)
+	syncResp, err := newSyncChannelMessageResp(resp, c.GetLoginUID(), req.DeviceUUID, req.ChannelID, req.ChannelType, m.messageExtraDB, m.messageUserExtraDB, m.messageReactionDB, m.channelOffsetDB, m.deviceOffsetDB, channelOffsetMessageSeq)
+	if err != nil {
+		m.Error("构建同步消息响应失败", zap.Error(err))
+		httperr.ResponseErrorL(c, errcode.ErrMessageQueryFailed, nil, nil)
+		return
+	}
 
 	// 群消息中的 ThreadCreated 消息：用实时数据覆盖 payload 中的快照字段
 	if req.ChannelType == common.ChannelTypeGroup.Uint8() {
@@ -2982,7 +2987,7 @@ type syncChannelMessageResp struct {
 	Messages        []*MsgSyncResp  `json:"messages"`          // 消息数据
 }
 
-func newSyncChannelMessageResp(resp *config.SyncChannelMessageResp, loginUID string, deviceUUID string, channelID string, channelType uint8, messageExtraDB *messageExtraDB, messageUserExtraDB *messageUserExtraDB, messageReactionDB *messageReactionDB, channelOffsetDB *channelOffsetDB, deviceOffsetDB *deviceOffsetDB, channelOffsetMessageSeq uint32) *syncChannelMessageResp {
+func newSyncChannelMessageResp(resp *config.SyncChannelMessageResp, loginUID string, deviceUUID string, channelID string, channelType uint8, messageExtraDB *messageExtraDB, messageUserExtraDB *messageUserExtraDB, messageReactionDB *messageReactionDB, channelOffsetDB *channelOffsetDB, deviceOffsetDB *deviceOffsetDB, channelOffsetMessageSeq uint32) (*syncChannelMessageResp, error) {
 	messages := make([]*MsgSyncResp, 0, len(resp.Messages))
 	if len(resp.Messages) > 0 {
 		messageIDs := make([]string, 0, len(resp.Messages))
@@ -3012,7 +3017,12 @@ func newSyncChannelMessageResp(resp *config.SyncChannelMessageResp, loginUID str
 		// 消息全局扩张
 		messageExtras, err := messageExtraDB.queryWithMessageIDsAndUID(messageIDs, loginUID)
 		if err != nil {
+			// fail-closed：撤回脱敏依赖 message_extra.revoke。查询失败则 messageExtraMap
+			// 为空，from() 拿不到 Revoke=1、会漏做脱敏并原样下发撤回原文。此处中止整个
+			// 同步、把 error 冒泡给调用方返回错误响应，与单条直查 api_message_get.go 同口径，
+			// 绝不带着不完整的撤回信息继续。
 			log.Error("查询消息扩展字段失败！", zap.Error(err))
+			return nil, err
 		}
 		// 修改消息扩展字段
 		for _, message := range resp.Messages {
@@ -3122,7 +3132,7 @@ func newSyncChannelMessageResp(resp *config.SyncChannelMessageResp, loginUID str
 		EndMessageSeq:   resp.EndMessageSeq,
 		PullMode:        resp.PullMode,
 		Messages:        messages,
-	}
+	}, nil
 }
 
 // 消息头
--- a/modules/message/api.go
+++ b/modules/message/api.go
@@ -124,14 +124,33 @@ func placeholderPayload() map[string]interface{} {
 // 保留 type 而不整条清空：撤回对所有人生效，前端仅凭 revoke=1 渲染撤回提示、不读
 // payload（撤回者名由 revoker UID 单独查，见 octo-web RevokeCell）；保留 type 只为
 // 兼容按 type 分支的老客户端渲染路径，type 本身不含消息正文。
+//
+// type 必须规范化为数字标量（scalarContentType）：payload 是不可信的调用方 JSON，
+// send 路径不约束 type 必须是数字（只剥 __obo_* / 处理 mention/space_id/richtext），
+// 若原样透传，攻击者可把正文藏进 type（字符串 / 对象 / 数组）绕过脱敏原样下发。
+// 服务端所有 type 消费方（isTextType / payloadMsgType）都严格按数字读取、非数字一律
+// 视为未知，故强制数字标量、非数字 fallback ContentError 不影响任何合法渲染（D23 安全整改）。
 func revokedPayload(original map[string]interface{}) map[string]interface{} {
-	stripped := map[string]interface{}{}
-	if t, ok := original["type"]; ok {
-		stripped["type"] = t
-	} else {
-		stripped["type"] = common.ContentError.Int()
+	return map[string]interface{}{
+		"type": scalarContentType(original["type"]),
 	}
-	return stripped
+}
+
+// scalarContentType 把 payload 的 type 归一为已识别的数字 content-type：
+// float64 / int / json.Number 三种反序列化结果转 int，其余（string / map / array / 缺失）
+// 一律 fallback common.ContentError，杜绝非标量 type 承载正文逃逸。
+func scalarContentType(v interface{}) int {
+	switch t := v.(type) {
+	case float64:
+		return int(t)
+	case int:
+		return t
+	case json.Number:
+		if i, err := t.Int64(); err == nil {
+			return int(i)
+		}
+	}
+	return common.ContentError.Int()
 }
 
 // isTextType 判断 payload type 是否为 common.Text（=1）。兼容 json.Number / float64 / int
@@ -1993,7 +2012,12 @@ func (m *Message) sync(c *wkhttp.Context) {
 	// 全局扩充数据
 	messageExtras, err := m.messageExtraDB.queryWithMessageIDsAndUID(messageIDs, c.GetLoginUID())
 	if err != nil {
-		log.Error("查询消息扩展字段失败！", zap.Error(err))
+		// fail-closed：撤回脱敏依赖 message_extra.revoke。查询失败则 messageExtraMap 为空，
+		// from() 拿不到 Revoke=1、会漏做脱敏并原样下发撤回原文。与 /message/channel/sync、
+		// /conversation/sync 及单条直查 api_message_get.go 同口径，中止请求而非继续。
+		m.Error("查询消息扩展字段失败！", zap.Error(err))
+		httperr.ResponseErrorL(c, errcode.ErrMessageQueryFailed, nil, nil)
+		return
 	}
 	messageExtraMap := map[string]*messageExtraDetailModel{}
 	if len(messageExtras) > 0 {
--- a/modules/message/api_conversation.go
+++ b/modules/message/api_conversation.go
@@ -718,7 +718,12 @@ func (co *Conversation) syncUserConversation(c *wkhttp.Context) {
 			channelOffsetM := channelOffsetModelMap[channelKey]
 			deviceOffsetM := deviceOffsetModelMap[channelKey]
 			extra := conversationExtraMap[channelKey]
-			syncUserConversationResp := newSyncUserConversationResp(conversation, extra, loginUID, co.messageExtraDB, co.messageReactionDB, co.messageUserExtraDB, mute, stick, channelOffsetM, deviceOffsetM, channelOffsetMessageSeq)
+			syncUserConversationResp, err := newSyncUserConversationResp(conversation, extra, loginUID, co.messageExtraDB, co.messageReactionDB, co.messageUserExtraDB, mute, stick, channelOffsetM, deviceOffsetM, channelOffsetMessageSeq)
+			if err != nil {
+				log.Error("构建会话同步响应失败", zap.Error(err), zap.String("channelID", conversation.ChannelID))
+				httperr.ResponseErrorL(c, errcode.ErrMessageQueryFailed, nil, nil)
+				return
+			}
 			// 填充群组分类信息
 			if conversation.ChannelType == common.ChannelTypeGroup.Uint8() {
 				if categorySetting := groupCategoryMap[conversation.ChannelID]; categorySetting != nil {
@@ -1472,7 +1477,7 @@ type SyncUserConversationResp struct {
 	BotType          string                 `json:"bot_type,omitempty"`           // Bot 类型（"app_bot" 表示应用 Bot）
 }
 
-func newSyncUserConversationResp(resp *config.SyncUserConversationResp, extra *conversationExtraResp, loginUID string, messageExtraDB *messageExtraDB, messageReactionDB *messageReactionDB, messageUserExtraDB *messageUserExtraDB, mute int, stick int, channelOffsetM *channelOffsetModel, deviceOffsetM *deviceOffsetModel, channelOffsetMessageSeq uint32) *SyncUserConversationResp {
+func newSyncUserConversationResp(resp *config.SyncUserConversationResp, extra *conversationExtraResp, loginUID string, messageExtraDB *messageExtraDB, messageReactionDB *messageReactionDB, messageUserExtraDB *messageUserExtraDB, mute int, stick int, channelOffsetM *channelOffsetModel, deviceOffsetM *deviceOffsetModel, channelOffsetMessageSeq uint32) (*SyncUserConversationResp, error) {
 	recents := make([]*MsgSyncResp, 0, len(resp.Recents))
 	lastClientMsgNo := "" // 最新未被删除的消息的clientMsgNo
 	if len(resp.Recents) > 0 {
@@ -1496,7 +1501,11 @@ func newSyncUserConversationResp(resp *config.SyncUserConversationResp, extra *c
 		// 消息扩充数据
 		messageExtras, err := messageExtraDB.queryWithMessageIDsAndUID(messageIDs, loginUID)
 		if err != nil {
+			// fail-closed：撤回脱敏依赖 message_extra.revoke。查询失败则 messageExtraMap
+			// 为空，from() 拿不到 Revoke=1、会漏做脱敏并原样下发撤回原文。冒泡给调用方
+			// 中止整个会话同步，与单条直查 api_message_get.go 同口径。
 			log.Error("查询消息扩展字段失败！", zap.Error(err))
+			return nil, err
 		}
 		messageExtraMap := map[string]*messageExtraDetailModel{}
 		if len(messageExtras) > 0 {
@@ -1561,7 +1570,7 @@ func newSyncUserConversationResp(resp *config.SyncUserConversationResp, extra *c
 		Stick:           stick,
 		Recents:         recents,
 		Extra:           extra,
-	}
+	}, nil
 }
 
 // threadMetaResp 子区元数据（仅 thread 频道返回）
--- a/modules/message/api_conversation.go
+++ b/modules/message/api_conversation.go
@@ -888,7 +888,7 @@ func (co *Conversation) syncUserConversation(c *wkhttp.Context) {
 	// Space 过滤
 	if hasSpaceFilter {
 		// Person 频道：计算 per-Space 未读计数（在过滤之前，需要原始会话数据）
-		fillPersonSpaceUnread(syncUserConversationResps, conversations, filterSpaceID, defaultSpaceID, loginUID, co.ctx)
+		fillPersonSpaceUnread(syncUserConversationResps, conversations, filterSpaceID, defaultSpaceID, loginUID, co.ctx, co.messageExtraDB)
 
 		syncUserConversationResps = FilterConversationsBySpace(
 			syncUserConversationResps, filterSpaceID, loginUID, co.ctx, co.groupService,
--- a/modules/message/revoke_payload_strip_test.go
+++ b/modules/message/revoke_payload_strip_test.go
@@ -33,6 +33,36 @@ func TestRevokedPayload(t *testing.T) {
 		assert.Equal(t, common.ContentError.Int(), out["type"])
 		assert.Len(t, out, 1)
 	})
+
+	// 安全回归：type 是不可信调用方数据，send 路径不约束其为数字标量。若把正文藏进
+	// 非标量 type（字符串 / 对象 / 数组），必须被规范化为 ContentError 而非原样透传，
+	// 否则撤回脱敏可被绕过（D23 整改 / PR #628 review by Jerry-Xin + yujiawei）。
+	t.Run("non-scalar type carrying content is normalized, never leaked", func(t *testing.T) {
+		const secret = "secret hidden in type field"
+		cases := map[string]interface{}{
+			"string type": secret,
+			"object type": map[string]interface{}{"nested": secret},
+			"array type":  []interface{}{secret},
+			"bool type":   true,
+		}
+		for name, badType := range cases {
+			t.Run(name, func(t *testing.T) {
+				out := revokedPayload(map[string]interface{}{"type": badType, "content": secret})
+				assert.Equal(t, common.ContentError.Int(), out["type"], "non-scalar type must fall back to ContentError")
+				assert.Len(t, out, 1)
+				body, err := json.Marshal(out)
+				assert.NoError(t, err)
+				assert.NotContains(t, string(body), secret, "no content may survive via the type field")
+			})
+		}
+	})
+
+	// 合法数字 type 的三种反序列化结果都必须原样保留（不会被误判为 ContentError）。
+	t.Run("numeric type is preserved across float64/int/json.Number", func(t *testing.T) {
+		assert.Equal(t, common.Text.Int(), revokedPayload(map[string]interface{}{"type": float64(common.Text.Int())})["type"])
+		assert.Equal(t, common.Text.Int(), revokedPayload(map[string]interface{}{"type": common.Text.Int()})["type"])
+		assert.Equal(t, common.Text.Int(), revokedPayload(map[string]interface{}{"type": json.Number("1")})["type"])
+	})
 }
 
 // TestMsgSyncRespFrom_RevokedStripsContent 是核心回归：撤回消息经 from()
--- a/modules/message/revoke_payload_strip_test.go
+++ b/modules/message/revoke_payload_strip_test.go
@@ -1,0 +1,250 @@
+package message
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/config"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestRevokedPayload 校验 revokedPayload 只保留 type、剥离一切内容承载字段。
+func TestRevokedPayload(t *testing.T) {
+	t.Run("strips content and every non-type field", func(t *testing.T) {
+		out := revokedPayload(map[string]interface{}{
+			"type":    common.Text.Int(),
+			"content": "secret original text",
+			"url":     "https://example.com/private.png",
+			"name":    "私密文件.pdf",
+			"reply":   map[string]interface{}{"content": "quoted secret"},
+		})
+		assert.Equal(t, common.Text.Int(), out["type"])
+		assert.Len(t, out, 1, "only type should remain")
+		for _, k := range []string{"content", "url", "name", "reply"} {
+			_, ok := out[k]
+			assert.Falsef(t, ok, "%q must be stripped", k)
+		}
+	})
+
+	t.Run("missing type falls back to ContentError", func(t *testing.T) {
+		out := revokedPayload(map[string]interface{}{"content": "x"})
+		assert.Equal(t, common.ContentError.Int(), out["type"])
+		assert.Len(t, out, 1)
+	})
+}
+
+// TestMsgSyncRespFrom_RevokedStripsContent 是核心回归：撤回消息经 from()
+// 组装后，原始正文不得出现在下发的任何字段里，而 revoke/revoker 元数据保留。
+// 覆盖 /v1/message/channel/sync 与 /v1/conversation/sync（两者共用 from()）。
+func TestMsgSyncRespFrom_RevokedStripsContent(t *testing.T) {
+	const secret = "top secret revoked message body"
+	payload, err := json.Marshal(map[string]interface{}{
+		"type":    common.Text.Int(),
+		"content": secret,
+	})
+	assert.NoError(t, err)
+
+	msgResp := &config.MessageResp{
+		MessageID:   12345,
+		MessageSeq:  10,
+		FromUID:     "sender1",
+		ChannelID:   "chan1",
+		ChannelType: common.ChannelTypePerson.Uint8(),
+		Payload:     payload,
+	}
+	extra := &messageExtraDetailModel{}
+	extra.MessageID = "12345"
+	extra.Revoke = 1
+	extra.Revoker = "admin1"
+	extra.Version = 7
+
+	m := &MsgSyncResp{}
+	m.from(msgResp, "viewer1", extra, nil, nil, 0)
+
+	assert.Equal(t, 1, m.Revoke, "revoke flag must be preserved for the tip")
+	assert.Equal(t, "admin1", m.Revoker, "revoker must be preserved for the tip")
+
+	_, hasContent := m.Payload["content"]
+	assert.False(t, hasContent, "revoked message must not ship original content")
+	assert.Len(t, m.Payload, 1, "only type should remain in the payload")
+
+	// Belt-and-suspenders: the secret must not appear anywhere in the wire body.
+	body, err := json.Marshal(m)
+	assert.NoError(t, err)
+	assert.NotContains(t, string(body), secret,
+		"revoked original content leaked into the sync response")
+}
+
+// TestMsgSyncRespFrom_RevokedStripsContentEdit 校验「编辑后又撤回」的消息不再经
+// message_extra.content_edit 下发编辑后的原文。
+func TestMsgSyncRespFrom_RevokedStripsContentEdit(t *testing.T) {
+	const editedSecret = "edited secret body that was later revoked"
+	payload, err := json.Marshal(map[string]interface{}{"type": common.Text.Int(), "content": "orig"})
+	assert.NoError(t, err)
+
+	msgResp := &config.MessageResp{
+		MessageID:   555,
+		MessageSeq:  6,
+		FromUID:     "sender1",
+		ChannelID:   "chan1",
+		ChannelType: common.ChannelTypePerson.Uint8(),
+		Payload:     payload,
+	}
+	extra := &messageExtraDetailModel{}
+	extra.MessageID = "555"
+	extra.Revoke = 1
+	extra.Revoker = "admin1"
+	extra.EditedAt = 1700000000
+	contentEdit, err := json.Marshal(map[string]interface{}{"type": common.Text.Int(), "content": editedSecret})
+	assert.NoError(t, err)
+	extra.ContentEdit.Valid = true
+	extra.ContentEdit.String = string(contentEdit)
+
+	m := &MsgSyncResp{}
+	m.from(msgResp, "viewer1", extra, nil, nil, 0)
+
+	assert.Equal(t, 1, m.Revoke)
+	if assert.NotNil(t, m.MessageExtra) {
+		assert.Nil(t, m.MessageExtra.ContentEdit, "revoked message must not ship content_edit")
+		assert.Equal(t, 0, m.MessageExtra.EditedAt)
+	}
+	body, err := json.Marshal(m)
+	assert.NoError(t, err)
+	assert.NotContains(t, string(body), editedSecret,
+		"revoked edited content leaked via content_edit")
+}
+
+// TestMsgSyncRespFrom_RevokedClearsSignalPayload 校验撤回的 signal 加密消息
+// 不再下发密文 blob（signal_payload）。
+func TestMsgSyncRespFrom_RevokedClearsSignalPayload(t *testing.T) {
+	cipher := []byte("encrypted-cipher-bytes-of-revoked-msg")
+	msgResp := &config.MessageResp{
+		MessageID:   333,
+		MessageSeq:  4,
+		FromUID:     "sender1",
+		ChannelID:   "chan1",
+		ChannelType: common.ChannelTypePerson.Uint8(),
+		Setting:     config.Setting{Signal: true}.ToUint8(),
+		Payload:     cipher,
+	}
+	extra := &messageExtraDetailModel{}
+	extra.MessageID = "333"
+	extra.Revoke = 1
+	extra.Revoker = "admin1"
+
+	m := &MsgSyncResp{}
+	m.from(msgResp, "viewer1", extra, nil, nil, 0)
+
+	assert.Equal(t, 1, m.Revoke)
+	assert.Empty(t, m.SignalPayload, "revoked signal ciphertext must be cleared")
+
+	body, err := json.Marshal(m)
+	assert.NoError(t, err)
+	assert.NotContains(t, string(body), base64.StdEncoding.EncodeToString(cipher),
+		"revoked signal ciphertext leaked into the sync response")
+}
+
+// TestMsgSyncRespFrom_RevokedClearsStreams 校验撤回的流式消息不再下发 stream blob。
+func TestMsgSyncRespFrom_RevokedClearsStreams(t *testing.T) {
+	const streamSecret = "streamed secret token"
+	payload, err := json.Marshal(map[string]interface{}{"type": common.Text.Int(), "content": "x"})
+	assert.NoError(t, err)
+
+	msgResp := &config.MessageResp{
+		MessageID:   444,
+		MessageSeq:  5,
+		FromUID:     "sender1",
+		ChannelID:   "chan1",
+		ChannelType: common.ChannelTypePerson.Uint8(),
+		Payload:     payload,
+		Streams: []*config.StreamItemResp{
+			{StreamSeq: 1, ClientMsgNo: "c1", Blob: []byte(`{"content":"` + streamSecret + `"}`)},
+		},
+	}
+	extra := &messageExtraDetailModel{}
+	extra.MessageID = "444"
+	extra.Revoke = 1
+	extra.Revoker = "admin1"
+
+	m := &MsgSyncResp{}
+	m.from(msgResp, "viewer1", extra, nil, nil, 0)
+
+	assert.Nil(t, m.Streams, "revoked stream blobs must be cleared")
+	body, err := json.Marshal(m)
+	assert.NoError(t, err)
+	assert.NotContains(t, string(body), streamSecret)
+}
+
+// TestMsgSyncRespFrom_NotRevokedKeepsContent 反向回归：未撤回消息内容原样保留，
+// 确保脱敏逻辑不误伤正常消息。
+func TestMsgSyncRespFrom_NotRevokedKeepsContent(t *testing.T) {
+	const visible = "normal visible message content"
+	payload, err := json.Marshal(map[string]interface{}{
+		"type":    common.Text.Int(),
+		"content": visible,
+	})
+	assert.NoError(t, err)
+
+	msgResp := &config.MessageResp{
+		MessageID:   222,
+		MessageSeq:  3,
+		FromUID:     "sender1",
+		ChannelID:   "chan1",
+		ChannelType: common.ChannelTypePerson.Uint8(),
+		Payload:     payload,
+	}
+
+	// no extra -> not revoked
+	m := &MsgSyncResp{}
+	m.from(msgResp, "viewer1", nil, nil, nil, 0)
+
+	assert.Equal(t, 0, m.Revoke)
+	assert.Equal(t, visible, m.Payload["content"], "non-revoked content must be preserved")
+}
+
+// TestSelectSpaceLastMessage_SkipsRevoked 校验 space_last_message 兜底选取会跳过
+// 已撤回消息，避免把撤回原文当作「最后一条消息」预览下发（/conversation/sync
+// SpaceLastMessage 字段的残余泄漏点）。
+func TestSelectSpaceLastMessage_SkipsRevoked(t *testing.T) {
+	mk := func(id int64, spaceID string) *config.MessageResp {
+		payload, _ := json.Marshal(map[string]interface{}{
+			"type":     common.Text.Int(),
+			"content":  "msg content",
+			"space_id": spaceID,
+		})
+		return &config.MessageResp{MessageID: id, Payload: payload}
+	}
+	// 顺序（旧 -> 新）：#1 spaceA(正常), #2 spaceA(撤回)。倒序遍历先命中 #2。
+	messages := []*config.MessageResp{mk(1, "spaceA"), mk(2, "spaceA")}
+
+	t.Run("revoked newest is skipped, older non-revoked chosen", func(t *testing.T) {
+		revoked := map[string]bool{"2": true}
+		got := selectSpaceLastMessage(messages, revoked, "spaceA", "", false)
+		if assert.NotNil(t, got) {
+			assert.Equal(t, int64(1), got.MessageID, "must skip revoked #2 and return #1")
+		}
+	})
+
+	t.Run("no revoked -> newest chosen", func(t *testing.T) {
+		got := selectSpaceLastMessage(messages, map[string]bool{}, "spaceA", "", false)
+		if assert.NotNil(t, got) {
+			assert.Equal(t, int64(2), got.MessageID)
+		}
+	})
+
+	t.Run("all revoked -> nil (nothing previewed)", func(t *testing.T) {
+		revoked := map[string]bool{"1": true, "2": true}
+		got := selectSpaceLastMessage(messages, revoked, "spaceA", "", false)
+		assert.Nil(t, got, "no non-revoked space message must yield no preview")
+	})
+}
+
+// TestRevokedMessageIDSet_NilDB 校验注入的 messageExtraDB 为空时返回空集合（不跳过），
+// 保证纯逻辑单测路径不依赖 DB。
+func TestRevokedMessageIDSet_NilDB(t *testing.T) {
+	msgs := []*config.MessageResp{{MessageID: 1}, {MessageID: 2}}
+	assert.Empty(t, revokedMessageIDSet(msgs, nil))
+	assert.Empty(t, revokedMessageIDSet(nil, nil))
+}
--- a/modules/message/revoke_payload_strip_test.go
+++ b/modules/message/revoke_payload_strip_test.go
@@ -241,10 +241,17 @@ func TestSelectSpaceLastMessage_SkipsRevoked(t *testing.T) {
 	})
 }
 
-// TestRevokedMessageIDSet_NilDB 校验注入的 messageExtraDB 为空时返回空集合（不跳过），
-// 保证纯逻辑单测路径不依赖 DB。
+// TestRevokedMessageIDSet_NilDB 校验注入的 messageExtraDB 为空时返回空集合、nil error
+// （不跳过），保证纯逻辑单测路径不依赖 DB。查询出错的 fail-closed 语义（返回 error →
+// findSpaceLastMessageFallback 跳过预览兜底）依赖真实 DB，由 infra-gated 集成测试覆盖。
 func TestRevokedMessageIDSet_NilDB(t *testing.T) {
 	msgs := []*config.MessageResp{{MessageID: 1}, {MessageID: 2}}
-	assert.Empty(t, revokedMessageIDSet(msgs, nil))
-	assert.Empty(t, revokedMessageIDSet(nil, nil))
+
+	set, err := revokedMessageIDSet(msgs, nil)
+	assert.NoError(t, err)
+	assert.Empty(t, set)
+
+	set, err = revokedMessageIDSet(nil, nil)
+	assert.NoError(t, err)
+	assert.Empty(t, set)
 }
--- a/modules/message/space_unread.go
+++ b/modules/message/space_unread.go
@@ -26,6 +26,7 @@ func fillPersonSpaceUnread(
 	defaultSpaceID string,
 	loginUID string,
 	ctx *config.Context,
+	messageExtraDB *messageExtraDB,
 ) {
 	if spaceID == "" || len(conversations) == 0 {
 		return
@@ -61,6 +62,7 @@ func fillPersonSpaceUnread(
 				fallbackMsg := findSpaceLastMessageFallback(
 					conv.ChannelID, conv.ChannelType,
 					loginUID, spaceID, defaultSpaceID, isSysBot, uint32(raw.LastMsgSeq), ctx,
+					messageExtraDB,
 				)
 				if fallbackMsg != nil {
 					conv.SpaceLastMessage = fallbackMsg
@@ -166,6 +168,7 @@ func findSpaceLastMessageFallback(
 	channelID string, channelType uint8,
 	loginUID string, spaceID, defaultSpaceID string,
 	isSysBot bool, lastMsgSeq uint32, ctx *config.Context,
+	messageExtraDB *messageExtraDB,
 ) *MsgSyncResp {
 	if lastMsgSeq == 0 {
 		return nil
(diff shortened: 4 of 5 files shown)


Jerry-Xin (dismissed):
APPROVE.

This closes a real revoked-content leak on the sync endpoints, and it covers BOTH the primary path and its bypass sibling — which is the failure mode most likely to leave a leak behind. Verified against the head branch.

## What leaks and how it's fixed
Revocation in this system is a soft delete (message_extra.revoke=1); the original payload still lives in the underlying store. The single-message read path (api_message_get.go) already returns 404 on revoke==1, but the sync paths shipped revoke=1 together with the full original payload, so the original content stayed recoverable on the client.

- `MsgSyncResp.from()` (modules/message/api.go): after all of Payload / SignalPayload / Streams / MessageExtra are assigned, a new `revoke==1` guard now (a) replaces the payload with `revokedPayload()` (keeps only `type`, strips content and every content-bearing field), (b) clears the encrypted SignalPayload ciphertext, (c) nils Streams, and (d) clears MessageExtra.ContentEdit + EditedAt. The ContentEdit clearing is important: an edited-then-revoked message would otherwise still ship the edited original via content_edit. revoke / revoker are preserved so clients still render the "recalled" tombstone. This is the shared builder for `/message/channel/sync` and `/conversation/sync`, so both are covered in one place.

- Sibling bypass (modules/message/space_unread.go): the space_last_message preview fallback builds its MsgSyncResp via `msgRespToSyncResp`, which does NOT go through `from()` — so it bypassed the scrub and could surface a revoked message's original content as the conversation's "last message". Fixed by batch-querying the revoked set (`revokedMessageIDSet` → messageExtraDB) and skipping revoked messages in `selectSpaceLastMessage` alongside already-deleted ones, so a revoked message is never chosen as the preview. The wiring (`api_conversation.go` passing `co.messageExtraDB` through `fillPersonSpaceUnread`) is correct.

## Leak-sibling sweep (why this is complete)
Every other MsgSyncResp construction in the message module (api.go channel sync, api_conversation.go, api_pinned.go, api_message_get.go) goes through `from()`, which now scrubs. The only builder that bypassed `from()` was the space_unread fallback — the exact path this PR patches. Single-message read already 404s. No unscrubbed sync/read path remains.

## Tests
revoke_payload_strip_test.go covers: revokedPayload keeps only type; from() strips content and asserts the secret is absent from the serialized body; content_edit is stripped (edited-then-revoked); signal ciphertext cleared; streams cleared. space_unread_test.go exercises the revoked-skip selection as a pure function.

## Non-blocking
- `revokedMessageIDSet` adds one batched revoked-status query per space-unread fallback invocation — acceptable (batched, not N+1), just noting it's an extra round trip on that path.
- `selectSpaceLastMessage` returning nil when all recent messages are revoked/deleted yields no preview (empty last-message) — correct behavior, worth a brief comment for future readers.

## Gate
Build / Vet / Lint green; Test still running at review time (non-blocking on content review). No blocking items.

lml2468 (dismissed):
# PR Review — PR #628 (Mininglamp-OSS/octo-server)

Reviewed at head SHA `c4786968` against merge-base `681967af` (base `main`).
**Build/vet/test:** `go build ./modules/message/` exit 0 ✅, `go vet` ✅. New unit tests `TestRevokedPayload | TestMsgSyncRespFrom_* | TestSelectSpaceLastMessage_* | TestRevokedMessageIDSet_*` → **PASS** ✅ (pure, no infra). Pre-existing DB-gated suites (`TestFollowChannel_*`, `thread_ext_blacklist_filter`) fail on `testutil.NewTestServer` → MySQL `127.0.0.1:3306` refused — infra-gated, **unrelated to this PR** (they fail identically at merge-base).
Classification: message-content data-leak fix on the sync read paths (`modules/message`) — space-isolation (P92)-adjacent, privacy/trust-boundary. Reviewed in **SECURITY** emphasis.

Intent: revoked messages were returned on `/v1/message/channel/sync` and `/v1/conversation/sync` with `revoke=1` **and** their full original payload, leaving the original text recoverable from client page memory (React props). This PR strips revoked content on the sync paths, extending the `revoke==1 → no content` invariant that `api_message_get.go` already enforced (404) on the single-message read path.

## 1. Spec compliance — ✅
No brief (direct security report); graded against the PR description + the existing `api_message_get.go` invariant it extends.
- **Missing:** none. Redaction added at the single sync funnel (`MsgSyncResp.from`), and the one hand-built preview path (`findSpaceLastMessageFallback`) that bypassed `from()` is fixed to skip revoked candidates.
- **Over-build:** none. Purely additive — two hunks in `api.go` (`revokedPayload` helper + a `Revoke==1` block in `from()`), plus the fallback skip + a pure `selectSpaceLastMessage` refactor. Non-revoked messages untouched; which messages are returned is unchanged (only revoked *content* is stripped).
- **Deviation:** none. Honors the documented approach and matches the sibling revoke-filter pattern in `api_channel_files.go`.

## 2. Code quality — Approved

### Findings
No P0/P1/P2 findings.

- **`from()` is the true single funnel — all six sync/read paths inherit the fix.** All production callers go through `from()`: channel-sync (`api.go:3116`), conversation-sync `last_message` (`api_conversation.go:1315`) and `recents` (`:1534`), pinned-sync (`api_pinned.go:682`), `api.go:2043`, and single-get (`api_message_get.go:272`). No sync path assembles a revoked `MsgSyncResp` while bypassing `from()`.
- **Every content carrier is stripped.** Enumerated the struct: `Payload` → `{type}` only, `SignalPayload` (E2E ciphertext) → `""`, `Streams` (stream blobs) → `nil`, and `MessageExtra.ContentEdit` (edited body) → `nil` + `EditedAt=0`. `messageExtraResp` has no other content-bearing field (rest is metadata). Metadata needed for the tip (`Revoke`, `Revoker`) is preserved.
- **Ordering is correct — redaction runs after all assignments.** `m.Payload` (3316), `m.SignalPayload` (3263), `m.Streams` (3338), `m.MessageExtra` (3257) are all set before the `if m.Revoke == 1` block (3346), so nothing re-populates a stripped field afterward.
- **The one hand-builder bypass is closed.** `findSpaceLastMessageFallback` builds the preview via `msgRespToSyncResp` (bypasses `from()`); it now batch-queries `queryRevokedWithMessageIDs` and skips revoked candidates in `selectSpaceLastMessage` (mirrors `api_channel_files.go:365`). All-revoked → `nil` (no preview) rather than a leak. `revokedMessageIDSet` fails **open on nil DB only in unit tests** (documented); the production caller wires `co.messageExtraDB` (`api_conversation.go:891`), which is non-nil.
- **Tests are genuinely effective (negative control).** I reverted the `from()` redaction block and re-ran: `TestMsgSyncRespFrom_RevokedClearsSignalPayload` and `…ClearsStreams` **FAIL** — the marshaled response then contains the signal ciphertext and `"streamed secret token"`. So the tests actually pin the leak, not a vacuous pass. `TestMsgSyncRespFrom_NotRevokedKeepsContent` guards against over-redaction.
- **`revokedPayload` degrades safely:** missing `type` → `common.ContentError`, never a nil map.

## 3. Test assessment
Eight targeted unit tests: payload / `content_edit` / signal / stream stripping, not-revoked keep-content, `selectSpaceLastMessage` skip-revoked (incl. all-revoked→nil), and nil-DB fail-open. All pure (`from()` called directly) → run in CI without MySQL/Redis/WuKongIM. Negative-control-verified above. Good coverage for the change.

## 4. Coverage / blind spots
- Verified by build/vet + the new unit tests + funnel/field enumeration + a negative control on the checkout.
- NOT verified here: the live end-to-end sync path (needs MySQL+Redis+WuKongIM) and the real `queryRevokedWithMessageIDs` batch query against MySQL — recommend the CI DB run.

## 5. Non-blocking fast-follow (NOT a blocker for this PR)
- **Quoted-reply snapshot may still surface revoked content — pre-existing, and explicitly scoped out by the author.** `newSyncChannelMessageResp` enriches `payload["reply"]["payload"]` with the quoted message's `content_edit` (`api.go:~3049`) with **no `revoke` check**. So if message B is revoked but message A quotes B, A's sync response (A not revoked → `from()` redaction never fires) can still ship B's edited/original body inside the embedded reply snapshot. **Blame confirms this block is untouched by this PR** (the api.go diff is two purely additive hunks; the reply-enrichment code is pre-existing), and the PR description states reply-rendering paths were deliberately scoped out pending frontend-impact verification. Flagging as the natural next step, not a defect this PR introduced. Suggest a follow-up that skips/blanks the reply snapshot when the quoted message is revoked (same `queryRevokedWithMessageIDs` batch already available here). The admin `/v1/manager` audit endpoints (`record`/`recordpersonal`) also emit revoked payloads, but that is a pre-existing, auth-gated moderation surface — out of scope.

## Verdict
**APPROVE** — a correct, minimal, well-scoped security fix. Revoked content is now stripped at the single sync funnel (`from()`) across all six sync/read paths, every content carrier (payload/signal/streams/content_edit) is cleared after assignment, the one hand-built preview bypass is closed to match the existing `api_channel_files.go` filter, and the tests are negative-control-verified to actually catch the leak. Both gates pass. The remaining reply-snapshot vector is pre-existing and explicitly deferred by the author — worth a fast-follow but not a merge blocker for this change.
Suggested decision: `Safe to merge` — after CI runs the DB-backed suites. File the reply-snapshot follow-up. (PR currently `merge: BLOCKED` on branch-protection.)

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #628 (octo-server)

## Summary

This PR closes a data-leak on the `/v1/message/channel/sync` and `/v1/conversation/sync` endpoints: revoked messages were being returned with `revoke=1` **and** their full original payload (plus `signal_payload`, `streams`, and `message_extra.content_edit`), so the original text stayed recoverable in client memory. The single-message read path (`api_message_get.go:241`) already returns `404` on `revoke==1`; this PR extends the same invariant to the sync paths by adding a post-assembly redaction step in the shared `MsgSyncResp.from()` function, and by skipping revoked candidates in the `SpaceLastMessage` fallback preview.

The fix is well-architected: centralizing redaction in `from()` ensures all sync callers (channel sync, conversation sync, pinned messages) benefit from a single code point, and the ordering is explicitly documented and verified correct.

## Verification

Local checks at head `c4786968`:
- `go build ./modules/message/` — clean
- `go vet ./modules/message/` — clean
- `go test ./modules/message/ -run 'TestRevokedPayload|TestMsgSyncRespFrom_|TestSelectSpaceLastMessage_|TestRevokedMessageIDSet_'` — pass (8 tests)
- `go test ./modules/message/ -run 'TestFillPersonSpaceUnread_|TestFindSpaceLastMessage_'` — pass (12 tests, signature-change regression)

- ✅ **Redaction ordering** — the `if m.Revoke == 1` block at `modules/message/api.go:3342` is the last write before the closing `}` of `from()`, after `Payload` (`:3316`), `SignalPayload` (`:3268`), `Streams` (`:3338`), and `MessageExtra` (`:3262`) are all assigned. The comment at `:3341` correctly documents this constraint.
- ✅ **All `from()` callers covered** — traced all 7 production call-sites: `api.go:2043` (channel sync), `api.go:3116` (channel sync loop), `api_conversation.go:1315` (conversation last-message), `api_conversation.go:1534` (conversation recents), `api_message_get.go:272` (single-get, already 404s), `api_pinned.go:682` (pinned messages). Every path that passes `messageExtra` with `Revoke=1` now triggers redaction.
- ✅ **`messageExtraMap` carries revoke flag** — `queryWithMessageIDsAndUID` at `db_message_extra.go:62` does `select message_extra.*` which includes the `revoke` column, so `messageExtraDetailModel.Revoke` is populated from the DB before reaching `from()`.
- ✅ **Fallback preview skips revoked** — `findSpaceLastMessageFallback` at `space_unread.go:206` builds `revokedSet` via `queryRevokedWithMessageIDs` (SQL: `revoke=1`) and passes it to `selectSpaceLastMessage`, which skips any message whose ID is in the set. The `msgRespToSyncResp` conversion at `:210` only receives non-revoked candidates.
- ✅ **Non-revoked messages unaffected** — the `if m.Revoke == 1` guard ensures zero impact on normal messages; `TestMsgSyncRespFrom_NotRevokedKeepsContent` locks this in.
- ✅ **Test quality** — each test asserts the secret is absent from the full marshaled JSON (`assert.NotContains`), not just from individual fields. The `revokedPayload` helper is tested with content-bearing payloads and the missing-type fallback.

## Findings

No P0/P1 issues. No correctness, security, or build-breaking concerns. Two minor observations below.

### Nit — `findSpaceLastMessage` non-fallback path doesn't filter revoked candidates (`modules/message/space_unread.go:152`)

The non-fallback `findSpaceLastMessage` operates on `recents []*MsgSyncResp` that have already been through `from()` — so payload content is already stripped to `{type}` only, and there is **no content leak**. However, a revoked message can still be returned as the `SpaceLastMessage` preview (with `Revoke=1`, stripped payload, and metadata). The fallback path now skips revoked; the non-fallback doesn't. The client already renders the revoke tip from `revoke=1`, so the UX impact is a preview showing "message revoked" instead of an older non-revoked message. Not a security issue; low-priority follow-up.

```go
// potential one-line addition at space_unread.go:155:
if msg.Revoke == 1 {
    continue
}
```

### Nit — Redaction ordering is comment-enforced (`modules/message/api.go:3341`)

The redaction block's correctness depends on being placed after all content-field assignments. The comment warns about this, but a future developer adding a new content-carrying field (e.g., `ThreadPayload`) could insert it after the redaction block without realizing the constraint. Not a current defect — just a maintenance-risk observation. A `t.Parallel()`-style build-time assertion (e.g., a struct-tag lint) is probably overkill for now.

## Data-flow trace

- **`from()` → `m.Revoke`**: set from `messageExtraM.Revoke` (`api.go:3252`), sourced from `queryWithMessageIDsAndUID` → `message_extra.revoke` column.
- **`from()` → `m.Payload`**: parsed from `msgResp.Payload` (WuKongIM raw bytes) via `ReadJsonByByte` or `TruncatedPayload`. For signal messages, cipher goes to `m.SignalPayload` and `m.Payload` gets `{type: SignalError}`. Redaction at `:3342` strips both.
- **`from()` → `m.MessageExtra.ContentEdit`**: parsed from `messageExtraM.ContentEdit.String` in `newMessageExtraResp` (`:3440`). Redaction at `:3353` nils it.
- **`from()` → `m.Streams`**: built from `msgResp.Streams` via `newStreamItemResp` (`:3338`). Redaction at `:3344` nils it.
- **Fallback preview**: `findSpaceLastMessageFallback` queries WuKongIM history (raw, unprocessed), builds `revokedSet` from `queryRevokedWithMessageIDs` (separate DB query filtering `revoke=1`), and `selectSpaceLastMessage` skips revoked before `msgRespToSyncResp` converts.

## Blind-spot checklist

- **C1 — Dual-path parity**: Main sync (`from()` redaction) and fallback preview (`selectSpaceLastMessage` skip) both handle revoked messages. `queryRevokedWithMessageIDs` (space_unread) and `queryWithMessageIDsAndUID` (sync path) query the same `message_extra` table with consistent `revoke` semantics. Clear.
- **C2 — Control-flow ordering**: Redaction runs exactly once at the end of `from()`. `from()` is not called recursively or from within other `from()` calls. Clear.
- **C3 — Authorization boundary**: N/A — no auth/permission changes.
- **C4 — Authorization lifecycle**: N/A — no container/member state changes.
- **C5 — Build/note**: `go build` and `go vet` pass clean at head `c4786968`.
- **C6 — Governance/docs**: N/A — no policy changes.

## Cross-round blocker review

## Verdict: APPROVED

Well-scoped security fix with correct architecture (centralize redaction in the shared `from()` funnel), proper ordering, comprehensive test coverage across all content carriers (payload, signal ciphertext, streams, edited content), and defensive fallback handling. The two nits are non-blocking observations.





---

[Octo-Q] verdict: APPROVE — No P0/P1 issues. The security fix correctly centralizes revoked-message content redaction in the shared `MsgSyncResp.from()` funnel, covers all content carriers (payload, signal_payload, streams, content_edit), and handles the fallback preview path by skipping revoked candidates. Build, vet, and all tests pass.

yujiawei (changes requested):
# Code Review — PR #628 (octo-server)

Reviewed at head SHA `c4786968deead92bf98c301afece7ee3ac48c662`. Build, `go vet`, and the new unit tests all pass locally on the message module. This review used a two-leg independent pass (line-level adversarial + long-context whole-repo) plus my own diff re-verification.

## 1. Spec compliance

The change does what it claims and nothing more.

- **Under-build:** none. The stated goal — stop shipping revoked content on `/v1/message/channel/sync` and `/v1/conversation/sync` — is implemented in the shared funnel `MsgSyncResp.from()` (`modules/message/api.go:3346`) plus the preview-fallback path (`modules/message/space_unread.go`).
- **Over-build:** none. Non-revoked messages are untouched (`TestMsgSyncRespFrom_NotRevokedKeepsContent`), and the set of returned messages is unchanged — only already-revoked content is stripped.
- **Deviation:** none in the happy path. Ordering inside `from()` is correct: `MessageExtra`, `SignalPayload`, `Payload`, and `Streams` are all assigned before the redaction block, with no later writes in that function. Signal ciphertext is cleared, and both the normal and `TruncatedPayload` paths reduce to type-only.

**Spec: ✅** for the scope the PR declares. The author transparently lists three same-class residuals (globally-deleted, `visibles`-restricted, reply-quote snapshots) as explicitly out of scope; that disclosure is accurate.

## 2. Code quality

**Quality: Changes-Requested.** The happy-path redaction is solid and well-tested. The blocker is on the error path of the newly-added code.

### P1 — New preview-fallback code fails open on DB error (blocker)

`modules/message/space_unread.go:251` (`revokedMessageIDSet`):

```go
revoked, err := messageExtraDB.queryRevokedWithMessageIDs(ids)
if err != nil {
    log.Warn("revokedMessageIDSet: 查询撤回消息失败", zap.Error(err))
    return set   // empty set → nothing is treated as revoked
}
```

On a transient query failure this returns an empty set, so `selectSpaceLastMessage` treats a revoked message as eligible and `msgRespToSyncResp(chosen)` copies its **full original payload** into `space_last_message` — reintroducing exactly the leak this PR closes.

This is inconsistent with the fail-closed reference the PR itself cites as "同口径": `modules/message/api_channel_files.go:365-368` propagates the error (`return nil, err`) rather than emitting unfiltered data. For a security fix, the new protective check should fail closed too — bubble the error up and skip the fallback preview rather than shipping an unfiltered one. The fix is small and the reference pattern is in the same package.

Both independent review legs flagged this path (consensus), which raises confidence it is real rather than stylistic.

### P1 — Pre-existing fail-open on the primary sync paths (not introduced here, but relevant to the security goal)

`modules/message/api.go:3013` and `modules/message/api_conversation.go:1497` both log-and-continue when `queryWithMessageIDsAndUID` fails, leaving `messageExtraMap` empty. `from()` then receives `nil`, `m.Revoke` stays `0`, and the redaction block is skipped → the raw WuKongIM payload is returned. The single-message endpoint already handles this the safe way (`api_message_get.go:235-239` returns an error when extras can't be loaded).

This PR does not regress this — it is the same failure mode that predates the change — but because the whole revoke-privacy model now hinges on that query succeeding, it should be fixed or tracked alongside the fallback fix above so the two paths are symmetric.

### P1 — Reply-quote path still ships a revoked parent's edited content on a fixed endpoint

`modules/message/api.go:3040` (reply enrichment in `newSyncChannelMessageResp`, which serves `/v1/message/channel/sync`):

```go
if messageExtra.ContentEdit.String != "" {
    ...
    replyMap["payload"] = contentEditMap
    payloadMap["reply"] = replyMap
    message.Payload = []byte(util.ToJson(payloadMap))
}
```

This does not check `messageExtra.Revoke`. If parent P was edited then revoked, and an unrevoked child C quotes P, C is returned with P's edited plaintext embedded in `payload.reply.payload`. `from()`'s redaction gates on C's own revoke flag (which is 0), so it does not strip the quoted parent content. This is one of the three residuals the author disclosed as out of scope — but it is a genuine leak on one of the two endpoints named in the PR title, so it warrants at minimum a tracked follow-up under the same redaction contract before the broad "stop leaking revoked content on sync endpoints" claim is fully true.

### P2 — `revokedPayload` copies `type` without validating it's a scalar content-type

`modules/message/api.go:127`: the helper preserves `original["type"]` verbatim. The send path accepts an arbitrary `map[string]interface{}` payload, so a crafted message with a non-scalar `type` (e.g. `{"type":{"secret":"..."}}`) would survive redaction intact. Low likelihood, but for defense-in-depth consider coercing to a recognized numeric content-type and falling back to `ContentError` otherwise.

### Refuted during verification (recorded for the audit trail)

One leg raised a concurrency concern that the redaction mutates a shared `MessageExtra` in place (`m.MessageExtra.ContentEdit = nil`). On recheck this is **not** a race: `m.MessageExtra` is freshly allocated per `from()` call via `newMessageExtraResp()` (returns a new `&messageExtraResp{}` with a freshly-unmarshaled `contentEditMap`). The mutation touches the per-response object, not the shared `messageExtraDetailModel` input. No shared state is mutated.

## 3. Overall verdict

**REQUEST_CHANGES.** Spec compliance is clean and the happy-path redaction (payload / signal / streams / content_edit) is correct and well-covered by tests. The blocker is that the newly-added `revokedMessageIDSet` fails open on DB error, re-leaking revoked content via `space_last_message` under a foreseeable transient failure, while a fail-closed reference exists in the same package. For a security-sensitive change, that error path should fail closed.

## 4. Recommendations

1. Make `revokedMessageIDSet` (or its caller) fail closed on query error — propagate the error and skip the fallback preview, mirroring `api_channel_files.go:365-368`.
2. In the same spirit, make the primary sync paths (`api.go:3013`, `api_conversation.go:1497`) treat an extra-query failure as fail-closed for revoke redaction, matching `api_message_get.go`.
3. File a tracked follow-up for the reply-quote residual (`api.go:3040`) so a revoked parent's `content_edit` is not embedded in a child's `reply.payload`.
4. Optional (P2): validate `type` in `revokedPayload`.

## 5. Additional notes for a human verifier

- Please confirm the frontend contract: the PR asserts octo-web's revoke tip renders purely from `revoke`/`revoker` and never reads `payload.content`, so stripping to type-only won't blank a legitimately-rendered row. Worth a manual check against the current web/iOS/Android clients since this is a wire-format change on a hot path.
- The redaction is verified at unit level only; the live sync path (MySQL+Redis+WuKongIM) was not exercised end-to-end. A staging check that a revoked message returns `revoke=1` with an empty payload on both endpoints would close that gap.

lml2468 (dismissed):
# PR Review (re-review) — PR #628 (Mininglamp-OSS/octo-server)

Re-reviewed at **new head `5798b3cd`** against merge-base `681967af` (base `main`). Supersedes my APPROVE at `c4786968`.
**Build/vet/test:** `go build ./modules/message/` exit 0 ✅, `go vet` ✅. Revoke/fail-closed unit tests (`TestRevokedPayload | TestMsgSyncRespFrom_* | TestSelectSpaceLastMessage_* | TestRevokedMessageIDSet_*`) → **PASS** ✅. Only failure is the pre-existing DB-gated `TestFollowChannel_*` (MySQL refused in sandbox — fails identically at merge-base, unrelated).

## What changed since `c4786968` (delta reviewed)
One new commit — `fix(message): fail closed on message_extra query errors in revoke redaction` — directly addressing the fail-open concern raised on the prior head. Three call sites, all now **fail-closed**:
- **`newSyncChannelMessageResp` (`/message/channel/sync`)** now returns `(*resp, error)`; the `messageExtraDB.queryWithMessageIDsAndUID` error is bubbled instead of logged-and-continued. Caller `syncChannelMessage` (`api.go:1400`) checks it → `ErrMessageQueryFailed` + return.
- **`newSyncUserConversationResp` (`/conversation/sync`)** same treatment; caller `syncUserConversation` (`api_conversation.go:721`) checks it → `ErrMessageQueryFailed` + return.
- **`revokedMessageIDSet` / `findSpaceLastMessageFallback` (preview fallback)** now returns `(set, error)`; on query error the fallback returns `nil` (no preview) rather than an empty "nobody revoked" set.

## Why this is the right fix
The prior head had a real fail-open gap: revoke redaction in `from()` depends on `message_extra.revoke`. If the `message_extra` batch query errored, the map was empty → `from()` never saw `Revoke==1` → the endpoint shipped the **original revoked content** anyway. That silently defeated the very leak this PR closes. The new commit makes all three sync/preview paths abort (or drop the preview) on that error — matching the single-message path (`api_message_get.go`) which already refuses to serve on a query failure. Correct call: **never serve possibly-unredacted content on partial data.**

## Verification (byte-level, this head)
- **All callers handle the new error signatures** — grep confirms the only two callers of the changed constructors both check `err` and return `ErrMessageQueryFailed`; `errcode.ErrMessageQueryFailed` exists (`pkg/errcode/message.go:145`). Build proves no other caller was left on the old signature.
- **Pinned sync was already fail-closed** — `api_pinned.go:612` returns `ErrMessageQueryFailed` on the `message_extra` query error *before* any `from()` call, so the pinned path never leaks on DB error. No gap there.
- **Redaction funnel unchanged and intact** — the `from()` strip block (payload→`{type}`, `SignalPayload`, `Streams`, `MessageExtra.ContentEdit`/`EditedAt`) is untouched by this commit and remains the single funnel for all six read paths. My prior negative-control (removing the strip block → `signal ciphertext` / `streamed secret token` leak into the response → tests FAIL) still stands.
- **Tests** — `TestRevokedMessageIDSet_NilDB` updated to assert the nil-DB path returns `(emptySet, nil)` (fail-open only for the no-injection unit case, correct); all revoke/preview tests green.

## On the peer findings from the prior head
- 🔴 "main sync path DB error → leak" and 🔴 "preview fallback DB error → leak" — **both valid, both fixed by this commit** (fail-closed, verified above). Good catches.
- 🔴 "reply-quote path still returns the revoked parent's edited content" — **valid but pre-existing and explicitly scoped out.** `newSyncChannelMessageResp` enriches `payload["reply"]["payload"]` with the quoted message's `content_edit` (`api.go:~3060`) with no `revoke` check. Blame confirms this block is **untouched** by this PR (all diff hunks are at 113/1397/2987/3017/3132/3348 — none cover the reply-enrichment lines), and the PR description states reply-rendering paths were deliberately deferred pending frontend-impact verification. So it is **not introduced or newly activated by this PR** → a fast-follow, not a blocker for this change. Strongly recommend the follow-up (same `queryRevokedWithMessageIDs` batch is already available in this function) so a revoked parent can't leak via a child's quote snapshot.

## Non-blocking notes
- **Test gap (non-blocking):** the fail-closed HTTP behavior (query error → `ErrMessageQueryFailed`, no body) has no direct test — it's DB-error-injection territory and the logic is trivial/clear on inspection, but a stubbed-error unit test on `revokedMessageIDSet` returning `error` (and the fallback returning nil) would lock it in.
- **Coverage/blind spots:** the live end-to-end sync path (MySQL+Redis+WuKongIM) and the real `queryWithMessageIDsAndUID` error branch are not exercised here — recommend the CI DB run.

## Verdict
**APPROVE** (re-affirmed at `5798b3cd`) — the fail-open gap peers flagged is now correctly closed: all sync/preview paths fail-closed on the `message_extra` query error, matching the single-message path, so revoked content is never served on partial data. The redaction funnel is intact and negative-control-verified, the pinned path was already safe, and both callers handle the new error signatures (build-proven). Both gates pass.
Suggested decision: `Safe to merge` after CI runs the DB-backed suites. **File the reply-quote-snapshot follow-up** (pre-existing, scoped-out) as the next hardening step. (PR gates on branch-protection approvals.)

mochashanyao (changes requested):
[Octo-Q · automated review]

---

[Octo-Q] verdict: REQUEST_CHANGES — C1 dual-path parity gap: `/message/sync` (api.go:1994) left fail-open on messageExtra query error while the other two sync endpoints got fail-closed. Same revoked-content leak pattern. Plus P2 on `getConversations` LastMessage (pre-existing, non-blocking).

yujiawei (approved):
# Code Review — PR #628 (octo-server)

Re-review at head SHA `5798b3cd2d8ac3421a3990bfc1cc1c8948424e9f` (merge-base `681967af`, base `main`). The head advanced by one commit since the prior round (`c4786968` → `5798b3cd`, *"fail closed on message_extra query errors in revoke redaction"*). The earlier round left a blocking review on the fail-open error paths; this pass verifies the new commit against the current head independently — the prior approvals do not carry forward automatically.

**Local gates (head `5798b3cd`):** `go build ./modules/message/` exit 0 ✅, `go vet ./modules/message/` ✅. Targeted unit tests `TestRevokedPayload | TestMsgSyncRespFrom_* | TestSelectSpaceLastMessage_* | TestRevokedMessageIDSet_* | TestFillPersonSpaceUnread_* | TestFindSpaceLastMessage*` → **PASS** (pure, no infra). DB-gated suites are unrelated (fail identically at merge-base on refused MySQL). Reviewed in **SECURITY** emphasis (classified security-sensitive; `needs-human-review` applies).

## 1. Spec compliance — ✅

Goal: stop shipping revoked message content on `/v1/message/channel/sync` and `/v1/conversation/sync`, extending the `revoke==1 → no content` invariant that `api_message_get.go` already enforces (404) on the single-read path.

- **Under-build:** none. Redaction is added at the shared funnel `MsgSyncResp.from()` (`modules/message/api.go:3356`), and the one hand-built preview path that bypassed `from()` (`findSpaceLastMessageFallback`) skips revoked candidates.
- **Over-build:** none. Purely additive; non-revoked messages are untouched and the set of returned messages is unchanged — only already-revoked *content* is stripped.
- **Deviation:** none. Matches the sibling revoke-filter pattern in `api_channel_files.go`.

## 2. Code quality — Approved

### The round-1 blocker is resolved (verified against the diff `c4786968..5798b3cd`)

The previous round requested changes because the newly-added protective code failed **open** on a DB error. Both are now fail-closed:

- `revokedMessageIDSet` (`space_unread.go:255`) now returns `(map[string]bool, error)` and propagates the query error instead of swallowing it into an empty set. `findSpaceLastMessageFallback` (`space_unread.go:209-216`) checks the error and returns `nil` (no preview) rather than surfacing a possibly-revoked message — matching the fail-closed reference in `api_channel_files.go`.
- The primary sync builders `newSyncChannelMessageResp` (`api.go:3017-3025`) and `newSyncUserConversationResp` (`api_conversation.go:1501-1509`) now return `error` on `queryWithMessageIDsAndUID` failure; both handlers (`api.go:1400-1405`, `api_conversation.go:721-726`) abort with `ErrMessageQueryFailed` instead of continuing with an empty `messageExtraMap` (which would leave `m.Revoke==0` and skip redaction). This makes the sync paths symmetric with `api_message_get.go`.

- **`from()` is the true single funnel.** Swept every `MsgSyncResp` construction site in the module: `api.go:2048`, `api.go:3126`, `api_conversation.go:1320`, `api_conversation.go:1543`, `api_pinned.go:682`, `api_message_get.go:272` all route through `from()`; the only non-`from()` builder is `msgRespToSyncResp` — the exact preview path this PR fixes.
- **Every content carrier is cleared, after assignment.** `Payload → {type}` only, `SignalPayload → ""`, `Streams → nil`, `MessageExtra.ContentEdit → nil` + `EditedAt=0`. The `if m.Revoke == 1` block is the last write in `from()`, after `Payload`/`SignalPayload`/`Streams`/`MessageExtra` are all set. Metadata for the tip (`Revoke`, `Revoker`) is preserved.
- **Post-`from()` mutation does not re-populate secrets.** On the group channel-sync path, `enrichThreadCreatedMessages` runs after `newSyncChannelMessageResp` returns, but it only writes `message_count` / `source_message_id` (thread metadata) onto already-redacted payloads — no original content is reintroduced.
- **Tests are effective (negative control).** I disabled the `from()` redaction block on the checkout and re-ran: `TestMsgSyncRespFrom_RevokedClearsSignalPayload` and `…ClearsStreams` **FAIL** (the marshaled body then contains the base64 ciphertext and `"streamed secret token"`). So the tests pin the leak, not a vacuous pass. `TestMsgSyncRespFrom_NotRevokedKeepsContent` guards against over-redaction.
- **Production DB is non-nil.** `revokedMessageIDSet`'s nil-DB fail-open is unit-test-only; the caller wires `co.messageExtraDB` (`api_conversation.go:896`), constructed non-nil in the module.

## 3. Coverage / blind spots

Verified: build/vet + targeted unit tests + funnel/field enumeration + a negative control on the checkout + blame confirmation of the changed hunks. **Not** verified here: the live end-to-end path (MySQL+Redis+WuKongIM) and the real `queryRevokedWithMessageIDs` batch query against MySQL — recommend the CI DB run.

## 4. Non-blocking fast-follows (NOT blockers for this diff)

These are pre-existing, same-class leaks that this PR neither introduces nor regresses. Two of them touch surfaces adjacent to the PR's stated scope, so they are worth tracking.

1. **Reply-quote snapshot still ships a revoked parent's edited content on `/v1/message/channel/sync`.** `newSyncChannelMessageResp` enriches `payload.reply.payload` with the quoted message's `content_edit` (`api.go:3053-3061`) with no `revoke` check. If parent P is edited-then-revoked and an unrevoked child C quotes P, C's response embeds P's edited plaintext; `from()` gates on C's own revoke flag (0), so it is not stripped. Blame confirms this block is pre-existing (`6dc233c3`, untouched by this PR) and the author explicitly deferred reply-rendering paths pending frontend verification. This is on a declared endpoint, so the PR title's claim is not yet *fully* true — recommend a follow-up under the same redaction contract (the `queryRevokedWithMessageIDs` batch is already available here).
2. **`GET /v1/conversations` (`getConversations`) does not redact its `last_message`.** At `api_conversation.go:1155` it calls `conversationResp.from(resp, loginUID, nil, nil, …)` with a `nil` messageExtra, so `m.Revoke` stays 0 and redaction never fires — a revoked last message ships its full payload. Pre-existing (blame `6dc233c3`) and outside the PR's two declared sync endpoints, but it is the same bug class on a third live endpoint. Strong fast-follow candidate.
3. **P2 — `revokedPayload` copies `type` verbatim** (`api.go`): a crafted non-scalar `type` (e.g. `{"type":{...}}`) would survive redaction. Low likelihood; consider coercing to a numeric content-type with `ContentError` fallback for defense-in-depth.

## 5. For a human verifier (security-sensitive)

- Confirm the client contract: the PR asserts octo-web's revoke tip renders from `revoke`/`revoker` only and never reads `payload.content`, so type-only stripping won't blank a legitimately-rendered row. Worth a manual check across current web/iOS/Android clients, since this is a wire-format change on a hot path.
- Redaction is verified at unit level only — a staging check that a revoked message returns `revoke=1` with an empty payload on both endpoints would close the gap.

## Verdict

**APPROVE.** The round-1 fail-open blocker is fixed and now fail-closed on both the preview-fallback and primary sync paths, symmetric with `api_message_get.go`. Redaction is centralized at the single `from()` funnel, every content carrier (payload / signal ciphertext / streams / content_edit) is cleared after assignment, the one hand-built bypass is closed, and the tests are negative-control-verified to actually catch the leak. Spec compliance is clean, no regression is introduced. The remaining reply-quote and `getConversations` vectors are pre-existing, same-class, and explicitly out of this PR's scope — worth tracked follow-ups but not merge blockers for this change. Suggest merging after CI runs the DB-backed suites, and filing the two fast-follows above.

Jerry-Xin (changes requested):
Re-review of new head 5798b3cd. The two fail-open-on-DB-error blockers from the prior round are now correctly fixed (fail-closed on both sync paths + the preview fallback). One residual leak remains in the redaction primitive itself: the revoked payload still ships the original `type` verbatim, so text stashed in `type` survives redaction. Requesting that one change.

## 🔴 Blocking

- 🔴 `modules/message/api.go` `revokedPayload`: the function copies `original["type"]` byte-for-byte into the stripped payload with no canonicalization. Message payloads are untrusted, largely caller-supplied JSON (the send path does not constrain `type` to a numeric scalar — `sendMessage` only strips reserved `__obo_*` keys and touches mention/space_id/richtext; nothing rejects a non-numeric `type`). Meanwhile every server consumer of `type` (`payloadMsgType`, `isTextType`) reads it strictly as float64/json.Number and silently treats anything else as unknown — so `type` is *always* semantically numeric to the server, and a message can be persisted with `type` set to an arbitrary string/object/array carrying text. A crafted payload whose secret lives in `type` (e.g. type set to the secret string, or an object/array embedding it) is shipped intact after revocation, defeating the exact invariant this PR establishes. Fix: emit only a recognized numeric scalar type (coerce float64/json.Number/int to int); for any non-numeric or missing `type`, fall back to `common.ContentError.Int()`. This is non-breaking — no legitimate consumer reads a non-numeric `type`. Add tests asserting that string / map / array `type` values do not appear in the serialized revoked response.

## 💬 Non-blocking

- 💬 Reply/quote snapshot (pre-existing, NOT introduced or worsened by this PR): the reply-enrichment block in `modules/message/api.go` injects a quoted parent's edited content (`message_extra.content_edit`) into the replying message's embedded `reply.payload` snapshot. It gates on the parent having been edited but does NOT check the parent's revoke flag, so a message quoting a now-revoked-and-previously-edited parent can still surface that parent's edited body via the embedded snapshot — the parent's own `from()` scrub never fires on the embedded copy. Byte-verified this enrichment logic is unchanged by this PR (identical between merge-base and head). Acceptable as a documented same-contract fast-follow given it is genuinely pre-existing, but it is a real residual leak of edited content and should not linger — recommend a follow-up that scrubs the embedded snapshot when the parent is revoked.
- 💬 Normal `Recents` preview path (`modules/message/space_unread.go`): the main preview selection does not skip revoked entries the way the space fallback now does. Post-fix this no longer leaks the body (redaction handles it), but a revoked message losing its space_id during redaction could still be mis-selected as the space last-message preview, diverging from the documented "latest non-revoked preview" behavior. Consider skipping revoked/deleted entries here too, with a regression test.

## ✅ Highlights

- The message-extra query on both sync endpoints now fails closed: on a query error the handlers abort with a query-failed error instead of proceeding with an empty extras map (which would leave revoke unset and skip redaction). This matches the single-read path's established fail-closed pattern and fixes the prior round's blocker #2.
- The preview fallback now fails closed: the revoked-set lookup returns an error on query failure and the fallback skips the preview entirely rather than treating an empty set as "nothing revoked." Matches the channel-files filter's abort-on-error pattern and fixes prior blocker #1.
- Signal ciphertext, stream blobs, and edited-content carriers are correctly cleared on revoke.
- The revoked-set query is batched (no N+1), and the fail-closed error path has clear rationale comments.

yujiawei (changes requested):
# Code Review — PR #628 (octo-server) — corrected verdict

Re-review at head SHA `5798b3cd2d8ac3421a3990bfc1cc1c8948424e9f`.

**Correction notice:** I posted an Approve on this same head earlier today, having verified the two error-path blockers from the prior round were fixed. I am superseding that with **Changes Requested**. On a second independent pass I confirmed a third sync handler was left fail-open on the exact `message_extra`-query error the remediation commit `5798b3cd` set out to close — the same revoked-content leak class, on a still-live route. My earlier Approve missed it; this review is authoritative.

**Local gates (head `5798b3cd`):** `go build ./modules/message/` exit 0 ✅, `go vet` ✅, targeted unit tests (`TestRevokedPayload | TestMsgSyncRespFrom_* | TestSelectSpaceLastMessage_* | TestRevokedMessageIDSet_* | TestFillPersonSpaceUnread_*`) PASS ✅. Negative control: disabling the `from()` redaction block makes the signal-ciphertext and stream tests FAIL, so the tests genuinely pin the leak.

## 1. Spec compliance — ✅ (with a completeness gap, see §2 P1)

The declared scope (`/v1/message/channel/sync`, `/v1/conversation/sync`) is implemented correctly: redaction centralized in `MsgSyncResp.from()`, preview-fallback bypass closed. No over-build, no deviation. The gap below is that the PR's own fail-closed contract was applied to two of the three sync handlers that share the leak pattern.

## 2. Code quality — Changes Requested

### Resolved from the prior round (verified on `c4786968..5798b3cd`)

- `revokedMessageIDSet` now returns `(set, error)` and `findSpaceLastMessageFallback` returns `nil` (no preview) on query error — fail-closed, matching `api_channel_files.go`.
- `newSyncChannelMessageResp` (`api.go:3017`) and `newSyncUserConversationResp` (`api_conversation.go:1501`) now propagate the `queryWithMessageIDsAndUID` error; both handlers abort with `ErrMessageQueryFailed`.

### P1 (blocking) — `POST /message/sync` still fails open on the revoke-redaction query

`modules/message/api.go:1994-1997` (`sync()`, registered at `api.go:324` — `message.POST("/sync", m.sync)`):

```go
messageExtras, err := m.messageExtraDB.queryWithMessageIDsAndUID(messageIDs, c.GetLoginUID())
if err != nil {
    log.Error("查询消息扩展字段失败！", zap.Error(err))   // log-and-continue, no return
}
```

On a query error the map is empty, so each `respVo.from(resp, ..., nil, ...)` sees `m.Revoke == 0` and the redaction block never fires — a revoked message is then returned with its **full original payload**. This is byte-for-byte the same fail-open mechanism the prior round blocked on, and the same one commit `5798b3cd` ("fail closed on message_extra query errors in revoke redaction") just closed on the other two sync handlers. `sync()` inlines the identical query→`from()` pattern and was not converted, so the PR's own fail-closed contract is only two-thirds applied.

Why this is in scope rather than a pre-existing deferral: the route is live (marked write-mode/"future-deprecated", but still registered and reachable), it is literally a *sync* endpoint (the PR title is "stop leaking revoked message content on sync endpoints"), and — unlike the reply-quote / globally-deleted / visibles residuals the author transparently scoped out — this is the **same bug class the remediation commit targeted**, not a different path. On the happy path `sync()` is already covered by `from()`; only the error path leaks, exactly as with the two handlers that were fixed. Holding the third handler to the invariant the PR just established for itself is a completeness requirement, not a new ask.

**Fix:** mirror the sibling handlers — `return`/abort with `ErrMessageQueryFailed` when `queryWithMessageIDsAndUID` errors in `sync()`, so revoke redaction cannot be silently skipped.

- **`from()` is the true single funnel.** Every `MsgSyncResp` builder routes through `from()` except `msgRespToSyncResp` (the preview path this PR fixes): `api.go:2048`, `api.go:3126`, `api_conversation.go:1320`, `api_conversation.go:1543`, `api_pinned.go:682`, `api_message_get.go:272`.
- **All content carriers cleared, after assignment.** `Payload → {type}`, `SignalPayload → ""`, `Streams → nil`, `MessageExtra.ContentEdit → nil` + `EditedAt=0`; the `if m.Revoke==1` block is the last write in `from()`. `enrichThreadCreatedMessages` (runs post-`from()`) only writes thread metadata, not content.
- **Tests are negative-control-verified** (above) and `TestMsgSyncRespFrom_NotRevokedKeepsContent` guards against over-redaction.

## 3. Non-blocking fast-follows (pre-existing, same class, NOT blockers)

1. **Reply-quote snapshot** (`api.go:3053-3061`): `payload.reply.payload` is enriched with a quoted parent's `content_edit` with no revoke check; if parent P is edited-then-revoked and child C quotes P, C's response embeds P's edited plaintext. Pre-existing (`6dc233c3`), author-deferred.
2. **`GET /v1/conversations` (`getConversations`)**: `api_conversation.go:1155` calls `conversationResp.from(resp, loginUID, nil, nil, …)` — `nil` messageExtra means `last_message` is never redacted. Pre-existing (`6dc233c3`), outside the two declared endpoints, but the same class on a third live endpoint.
3. **P2** — `revokedPayload` copies `type` verbatim; a crafted non-scalar `type` survives redaction. Consider coercing to a numeric content-type with `ContentError` fallback.

## 4. For a human verifier (security-sensitive)

- Confirm the client contract (octo-web revoke tip renders from `revoke`/`revoker` only, never `payload.content`) across web/iOS/Android — this is a wire-format change on a hot path.
- Redaction verified at unit level only; a staging check that a revoked message returns `revoke=1` with an empty payload on all sync endpoints (including `/message/sync`) would close the gap.

## Verdict

**CHANGES_REQUESTED.** The prior round's two error-path blockers are correctly fixed and the happy-path redaction is solid and negative-control-verified. The remaining blocker is that `POST /message/sync` (`api.go:1994`) was left fail-open on the same `message_extra` query the remediation commit closed elsewhere — a live sync endpoint that ships full revoked-message payload on a transient DB error, violating the PR's own "fail closed on message_extra query errors" contract. Converting that one handler to fail-closed (matching its two siblings) clears the block. The reply-quote and `getConversations` vectors remain worthwhile pre-existing fast-follows but are not blockers for this diff.

Jerry-Xin (approved):
**APPROVE** — the round-2 `type`-normalization blocker is fixed byte-correctly, tests cover the leak vector, and nothing previously accepted regressed.

## ✅ Prior 🔴 (verbatim `payload.type` passthrough) — FIXED
`revokedPayload` no longer copies the original `type` field. It now returns exactly `{ "type": scalarContentType(original["type"]) }`, and `scalarContentType` only accepts a recognized numeric scalar:
- `float64` / `int` / `json.Number` → coerced to `int`
- everything else (string / map / array / bool / missing) → `common.ContentError.Int()`

This precisely mirrors how every server-side consumer (`isTextType` / `payloadMsgType`) reads `type` — numeric-only, non-numeric treated as unknown — so a crafted non-scalar `type` carrying free text can no longer survive redaction. After redaction, `type` is always a number and carries no body. Confirmed no `content`/`url`/`name`/`reply` or other carrier can ride through: the returned map has exactly one key.

Test coverage is present and passes (build + `go test` green locally):
- `TestRevokedPayload/"non-scalar type carrying content is normalized, never leaked"` — string, object, array, bool cases, each asserting `type == ContentError` and a belt-and-suspenders `NotContains(secret)` on the marshaled output.
- `TestRevokedPayload/"numeric type is preserved…"` — float64 / int / json.Number all preserved (no over-redaction).
- `TestMsgSyncRespFrom_RevokedStripsContent` — end-to-end via `from()`, secret absent from the whole wire body.

## ✅ No regressions (byte-verified against the prior head)
The delta vs the prior head is exactly two files (`api.go` + the new test); the accepted redaction block in `from()` (Payload→`{type}`, `SignalPayload=""`, `Streams=nil`, `MessageExtra.ContentEdit=nil`, `EditedAt=0`) is unchanged, and the fail-closed-on-query-error hardening on the sync paths / preview fallback / pinned path is intact. This head additionally hardens one more sync handler (`Message.sync`) to abort with `ErrMessageQueryFailed` instead of log-and-continue — consistent same-口径 fix, no downside. The reply-quote enrichment block is not touched by this delta.

## 💬 Non-blocking (pre-existing, fast-follow — unchanged by this PR)
- Reply/quote snapshot of a revoked/edited parent can still embed the parent's content in the replying message's `reply` preview. Byte-confirmed pre-existing and not touched here; fine to follow up under the same redaction contract (as the PR description already scopes out).
- Same class: globally-deleted (`is_deleted=1`) and `visibles`-restricted messages still ship full payload on the sync funnel while the single-read path blocks them. Out of scope here, worth a same-contract follow-up.

## ✅ Highlights
- The numeric-scalar coercion exactly matches consumer semantics — the fallback to `ContentError` provably cannot drop legitimate rendering.
- Negative-control tests (secret hidden in a non-scalar `type`) directly encode the attack and assert it cannot survive.
- Extending fail-closed to the last remaining sync handler closes the "empty extras → skip redaction" bypass consistently across all sync paths.

lml2468 (approved):
# PR Review (re-review) — PR #628 (Mininglamp-OSS/octo-server)

Re-reviewed at **new head `d186fcdb`** against merge-base `681967af` (base `main`). Supersedes my APPROVE at `5798b3cd`.
**Build/vet/test:** `go build ./modules/message/` exit 0 ✅, `go vet` ✅. Revoke/normalize unit tests (`TestRevokedPayload | TestMsgSyncRespFrom_* | TestSelectSpaceLastMessage_* | TestRevokedMessageIDSet_*`) → **PASS** ✅. Only failure is the pre-existing DB-gated `TestFollowChannel_*` (MySQL refused in sandbox — fails identically at merge-base, unrelated).

## What changed since `5798b3cd` (delta reviewed)
One new commit — `fix(message): fail closed on POST /message/sync and normalize revoked type` — closing the two items raised on the prior head:

1. **`type` normalization (was the outstanding blocker).** `revokedPayload` no longer copies the original `type` verbatim; it runs `scalarContentType(original["type"])`, which accepts only `float64 / int / json.Number` and falls back to `common.ContentError` for **string / map / array / missing**. This closes the escape where content could hide in a non-scalar `type` field and survive redaction.
2. **`POST /message/sync` (`m.sync`, `api.go:2012`) now fails closed.** Its `message_extra` query error was previously log-and-continue → empty extras → `from()` never sees `Revoke=1` → revoked content shipped. It now returns `ErrMessageQueryFailed`, matching the other sync handlers.

## Verification (byte-level, this head)

### The `type`-escape fix is correct and genuinely closes the hole (negative-control verified)
- **Consumers read `type` numerically-only**, so the fallback breaks nothing legitimate: `isTextType` switches on `float64/int/json.Number` (non-number → false); `payloadMsgType` (`api.go:3513`) switches on `float64/json.Number` (else `0`). A non-scalar `type` was already treated as "unknown" by every consumer — normalizing it to `ContentError` changes no valid render path.
- **Negative control:** I reverted `revokedPayload` to the old verbatim-`type` passthrough and re-ran — `TestRevokedPayload/non-scalar_type…/array_type` **FAILS**, with `["secret hidden in type field"]` surviving in the output. Restored; with the fix it passes. So the test actually pins the escape, and the fix actually shuts it.
- New tests cover string/object/array `type` → `ContentError` **and** the three numeric forms preserved (no over-normalization).

### Fail-closed is now uniform across every sync read path that feeds `from()`
Enumerated all `from()`-feeding handlers and their `message_extra` error handling on this head:
1. `POST /message/sync` (`m.sync`, api.go:2012) — **fail-closed (this commit)** ✅
2. `/message/channel/sync` (`syncChannelMessage`, api.go:1416) — fail-closed (prior commit) ✅
3. `/conversation/sync` (`syncUserConversation`, api_conversation.go:721) — fail-closed (prior commit) ✅
4. pinned sync (api_pinned.go:612) — pre-existing fail-closed (returns `ErrMessageQueryFailed` before any `from()`) ✅
5. single-get (`api_message_get.go`) — the original `revoke→404` invariant ✅

Redaction funnel (`from()` strip block) is unchanged and remains the single funnel; my earlier negative-control on the strip block still stands.

## On the remaining peer findings
- **Jerry-Xin 🔴 `type` verbatim** — **valid, fixed by this commit** (verified above). Good catch, same root cause Steve named.
- **yujiawei / mochashanyao 🔴 `POST /message/sync` fail-open** — **valid, fixed by this commit.** Candid note: this was a *third* sync handler that **I under-called on the prior two heads** — I verified `syncChannelMessage`/`syncUserConversation`/pinned but did not enumerate `m.sync`. yujiawei/mochashanyao were right; it's now closed and I've confirmed the whole set is uniform.
- **mochashanyao 🟡 / Jerry-Xin 🟡 `getConversations` LastMessage** — **valid but pre-existing and NOT touched by this PR.** `getConversations` (api_conversation.go:1155) builds `LastMessage` via `conversationResp.from(resp, loginUID, nil, nil, …)` — it passes **nil** `message_extra`, so `Revoke` is never set and redaction never fires there. Blame/diff confirm this block is outside every hunk of this PR (api.go + api_conversation.go hunks listed; none cover 1116–1180). So it is not introduced or newly activated here → fast-follow, not a blocker for this change. Worth fixing next (fetch `message_extra` for the last-message set, or reuse `queryRevokedWithMessageIDs`).
- **Reply-quote snapshot (Jerry-Xin 🟡, earlier yujiawei 🔴)** — pre-existing, byte-identical merge-base↔head, untouched by this PR. Fast-follow.

## Non-blocking notes
- **Test gap:** the fail-closed HTTP behavior on `m.sync` (query error → `ErrMessageQueryFailed`) has no direct handler-level test (DB-error injection); the logic is trivial and consistent with the sibling handlers, but a stubbed-error unit test would lock it in.
- **Coverage/blind spots:** live end-to-end sync (MySQL+Redis+WuKongIM) and the real `queryWithMessageIDsAndUID` error branch not exercised here — recommend the CI DB run.
- **Two pre-existing leak vectors remain for fast-follow:** `getConversations` LastMessage (nil extra → no redaction) and the reply-quote snapshot. Both are outside this PR's diff; recommend filing them together as the "extend revoke redaction to all read paths" follow-up.

## Verdict
**APPROVE** (re-affirmed at `d186fcdb`) — both prior-head blockers are correctly resolved: `type` is normalized to a scalar content-type (negative-control-verified to stop content hiding in `type`), and `POST /message/sync` now fails closed, making all five `from()`-feeding read paths uniform. The redaction funnel is intact. The two residual leak vectors (`getConversations` nil-extra, reply-quote snapshot) are pre-existing, untouched by this PR, and correctly deferred. Both gates pass for the scope this PR owns.
Suggested decision: `Safe to merge` after CI runs the DB-backed suites. **File the two pre-existing read-path leaks** (`getConversations` + reply-quote) as the immediate fast-follow. (PR gates on branch-protection approvals.)

yujiawei (approved):
# Code Review — PR #628 (octo-server)

Reviewed at head SHA `d186fcdb6ac7e9d487c6572f46ea17a89f9f0ab8` (merge-base `681967af`, base `main`), classified security-sensitive. Independent re-verification against the current head — prior-round approvals do not carry forward. Local gates: `go build ./modules/message/` ✅, `go vet` ✅, targeted unit suite (`TestRevokedPayload | TestMsgSyncRespFrom_* | TestSelectSpaceLastMessage_* | TestRevokedMessageIDSet_* | TestFillPersonSpaceUnread_*`) PASS ✅. DB-gated suites fail identically at merge-base (refused MySQL) — unrelated.

## 1. Spec compliance — ✅

Goal: stop shipping revoked message content on the sync endpoints, extending the `revoke==1 → no content` invariant that `api_message_get.go` already enforces (404) on the single-read path.

- **Under-build:** none. Redaction is centralized in the shared funnel `MsgSyncResp.from()` (`modules/message/api.go:3380`), and the one hand-built preview path that bypassed `from()` (`findSpaceLastMessageFallback`) now skips revoked candidates.
- **Over-build:** none. Purely additive — non-revoked messages are untouched and the set of returned messages is unchanged; only already-revoked *content* is stripped.
- **Deviation:** none. Matches the sibling revoke-filter pattern in `api_channel_files.go`.

## 2. Code quality — Approved

The two remaining blockers from the previous rounds are both resolved on this head, and I verified each independently.

### Prior blocker #1 — `POST /message/sync` fail-open on the revoke-redaction query → fixed

`sync()` (`modules/message/api.go:2012`) now aborts with `ErrMessageQueryFailed` when `queryWithMessageIDsAndUID` errors, instead of log-and-continue. Previously an empty `messageExtraMap` left `m.Revoke==0`, skipping redaction and shipping the full original payload on a transient DB error. All three sync handlers (`syncChannelMessage`, `syncUserConversation`, `sync`) plus the preview fallback are now symmetric with the single-read path — fail-closed.

### Prior blocker #2 — `revokedPayload` shipped `type` verbatim → fixed

`revokedPayload` (`api.go:126`) now normalizes `type` through `scalarContentType`: `float64` / `int` / `json.Number` → `int`; anything else (string / object / array / bool / missing) → `common.ContentError`. Since message payloads are untrusted caller-supplied JSON and the send path does not constrain `type` to a numeric scalar, this closes the vector where original text could be smuggled inside a non-scalar `type` field and survive redaction. This is non-breaking: every server consumer of `type` (`isTextType`, `payloadMsgType`) already reads it strictly as numeric.

### Verified clean (adversarial checks that did not find a bug)

- **`from()` is the true single funnel.** Every `MsgSyncResp` builder routes through `from()` — `api.go:2072`, `api.go:3150`, `api_conversation.go:1320`, `api_conversation.go:1543`, `api_pinned.go:682`, `api_message_get.go:272` — except `msgRespToSyncResp`, which is the exact preview path this PR fixes.
- **All content carriers cleared, after assignment.** The `if m.Revoke == 1` block is the last write in `from()`: `Payload → {type}` only, `SignalPayload → ""`, `Streams → nil`, `MessageExtra.ContentEdit → nil` + `EditedAt=0`. Metadata for the tip (`Revoke`, `Revoker`) is preserved. No later write re-populates a stripped field.
- **Fail-closed reference matched.** `revokedMessageIDSet` returns `(set, error)` and `findSpaceLastMessageFallback` returns `nil` (no preview) on query error, mirroring `api_channel_files.go`. `queryRevokedWithMessageIDs` filters `` `revoke`=1 `` and is batched (no N+1).
- **Signature changes fully wired.** `newSyncChannelMessageResp`, `newSyncUserConversationResp`, and `fillPersonSpaceUnread` all have their (only) callers updated; the whole test tree compiles.
- **Tests are negative-control-verified.** Reverting `revokedPayload` to a raw-`type` passthrough makes `TestRevokedPayload` FAIL, so the type-normalization test genuinely pins the fix rather than passing vacuously.

## 3. Coverage / blind spots

Verified: build/vet + targeted unit tests + funnel/field enumeration + a negative control + blame confirmation that the changed hunks are precisely scoped. **Not** verified here: the live end-to-end sync path (MySQL+Redis+WuKongIM) and the real batch queries against MySQL — recommend the CI DB run.

## 4. Non-blocking fast-follows (pre-existing, same class, NOT blockers for this diff)

These are pre-existing leaks this PR neither introduces nor regresses; blame confirms the relevant blocks are untouched by this diff:

1. **Reply-quote snapshot** (`api.go:~3077`): `payload.reply.payload` is enriched with a quoted parent's `content_edit` with no revoke check, so a child quoting an edited-then-revoked parent can still embed the parent's edited body. On a declared endpoint — worth a follow-up under the same redaction contract (the `queryRevokedWithMessageIDs` batch is already available there).
2. **`GET /v1/conversations` (`getConversations`)** (`api_conversation.go:1155`): calls `from(resp, loginUID, nil, nil, …)` with a `nil` messageExtra, so `last_message` is never redacted — same bug class on a third live endpoint outside this PR's two declared endpoints.
3. Globally-deleted and `visibles`-restricted messages on the sync funnel — the author transparently listed these as scoped-out same-class residuals.

## 5. For a human verifier (security-sensitive)

- Confirm the client contract: the PR asserts the revoke tip renders from `revoke`/`revoker` only and never reads `payload.content`, so type-only stripping won't blank a legitimately-rendered row. Worth a manual check across current web/iOS/Android clients — this is a wire-format change on a hot path.
- Redaction is verified at unit level only; a staging check that a revoked message returns `revoke=1` with an empty payload on all three sync endpoints would close the gap.

## Verdict

**APPROVE.** Both prior-round error-path and type-normalization blockers are fixed and independently verified on this head. Redaction is centralized at the single `from()` funnel, every content carrier is cleared after assignment, all three sync handlers plus the preview fallback fail closed on the `message_extra` query, the untrusted `type` field is normalized, and the tests are negative-control-verified to actually catch the leak. Spec is clean, no regression introduced. The reply-quote and `getConversations` vectors are pre-existing, same-class, and out of this diff's scope — recommend tracked follow-ups. Suggest merging after CI runs the DB-backed suites.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #628 (octo-server)

## Summary

This PR closes a security gap where revoked (撤回) message content was leaked through sync endpoints (`/v1/message/channel/sync`, `/v1/conversation/sync`, and the conversation Recents path). The fix introduces `revokedPayload()` to strip all content-bearing fields from revoked messages at the `from()` assembly point, applies fail-closed error handling when `message_extra` queries fail (previously errors were logged and silently ignored), and adds revoke-aware filtering to the `space_last_message` fallback path. The implementation is thorough, well-tested, and correctly addresses all three sync code paths.

## Verification

Local checks at head `d186fcdb`:
- `go vet ./modules/message/...` — clean
- `go test ./modules/message/` — pass (full module suite)
- `go test -run "TestRevokedPayload|TestMsgSyncRespFrom_|TestSelectSpaceLastMessage|TestRevokedMessageIDSet" -v` — 8 test functions, all sub-tests pass

## Findings

### P2 — Residual leak on deprecated `getConversations` route (`modules/message/api_conversation.go:1155`)

`getConversations` (mounted only on the misspelled `/v1/coversations` deprecated route) calls `conversationResp.from(resp, loginUID, nil, nil, ...)` — passing `nil` for `messageExtra`. This means `from()` never sees `Revoke=1` and the `LastMessage` of a revoked conversation still ships original content through this endpoint. This is pre-existing and not introduced or amplified by the PR; the route itself is explicitly deprecated with `deprecatedLog`. Flagging for awareness: if any client still polls this endpoint, revoked content remains accessible. Not a merge blocker.

## Data Flow Traceback

- **`m.Revoke` in `from()`** (`api.go:3284`): Set from `messageExtraM.Revoke` inside `if messageExtraM != nil`. The stripping block at `:3383` checks `m.Revoke == 1`. Since Go struct int fields default to 0, a nil `messageExtraM` yields `m.Revoke=0` → no stripping. All three sync paths now fail-closed on `messageExtraDB` query errors, so the only way `messageExtraM` can be nil is when the message genuinely has no extra record (not revoked). Correct.
- **`revokedPayload()`** (`api.go:136`): Returns `{"type": scalarContentType(original["type"])}` — a single-key map. `scalarContentType` handles `float64`, `int`, and `json.Number` → int conversion; everything else falls back to `common.ContentError.Int()`. No content-bearing field can survive. Correct.
- **`findSpaceLastMessageFallback`** (`space_unread.go:198`): Calls `revokedMessageIDSet()` to build a revoke set, then `selectSpaceLastMessage()` skips revoked messages before `msgRespToSyncResp()` is called. The conversion function itself doesn't strip content, but revoked messages never reach it. Fail-closed on `queryRevokedWithMessageIDs` error (returns nil → no preview). Correct.
- **`queryRevokedWithMessageIDs`** (`db_message_extra.go:70`): Pre-existing DB method. `SELECT * FROM message_extra WHERE message_id IN ? AND revoke=1`. Returns `[]*messageExtraModel`. Error is properly propagated. Correct.

## Blind Spot Checklist (C1–C6)

- **C1 — Dual-path parity**: All three production callers of `from()` covered: `newSyncChannelMessageResp` (api.go:3150), `newSyncUserConversationResp` (api_conversation.go:1543), pinned messages (api_pinned.go:682). The revoke stripping lives in `from()` itself, so all callers get it automatically. `getConversations` on the deprecated route is the one gap (P2 above). **Hit** — documented as P2.
- **C2 — Control-flow ordering / nested reuse**: The stripping block is placed at the very end of `from()` after all payload/signal/stream assignments. No risk of double-application. The `scalarContentType` function correctly handles the three JSON deserialization variants. **Clear.**
- **C3 — Authorization boundary**: Not applicable — this PR restricts data exposure, not authorization. **N/A.**
- **C4 — Authorization lifecycle / container-member state cascade**: Not applicable — no auth/jail changes. **N/A.**
- **C5 — Build/note passing ≠ runtime correctness**: `go vet` clean, all tests pass. The `from()` stripping is verified at the unit level with direct struct assertions and full JSON marshal checks. **Clear.**
- **C6 — Governance/docs self-consistency**: Not applicable — no governance or policy docs changed. **N/A.**

## Cross-Round Blocker Re-review (R6)

N/A — first review round on this PR.

- Signal-encrypted messages: `m.SignalPayload` is base64-encoded from raw payload at `:3299`, then cleared to `""` at `:3386` when `m.Revoke == 1`. Ciphertext cannot leak.
- Stream blobs: `m.Streams` populated at `:3368`, then set to `nil` at `:3387` on revoke. Stream content cannot leak.
- Content edit: `m.MessageExtra.ContentEdit` and `EditedAt` are cleared at `:3389-3390` on revoke. Edited-then-revoked content cannot leak.
- `visibles` whitelist interaction: `visibles` sets `m.IsDeleted`, not `m.Revoke`. A message that is both visible-whitelisted and revoked correctly gets stripped.
- `TruncatedPayload` path: Large payloads go through `TruncatedPayload()` at `:3308`, but the revoke block replaces the payload entirely regardless.

## Verdict: APPROVED

The implementation correctly closes the revoked-content leak on all three sync paths with proper fail-closed semantics. Test coverage is comprehensive (8 test functions covering payload stripping, content_edit, signal, streams, non-revoked passthrough, space_last_message filtering, and nil-DB safety). The one P2 is a pre-existing gap on a deprecated route, not introduced or amplified by this PR.





[Octo-Q] verdict: APPROVE — No P0/P1 issues found. Revoke content stripping is correctly implemented in `from()` covering all sync paths. Fail-closed error handling on `message_extra` queries prevents silent degradation. One pre-existing P2 on deprecated route flagged for awareness.

PR #629, opened 2026-07-21
feat(agent-runtime): resident HTTP runtime embedding the pi kernel (Fastify + auth + verbose + SSE)

ploy-elison:
## Summary

Adds **`agent-runtime/`** — a new, additive orchestration-layer service that embeds the
[`@earendil-works/pi-coding-agent`](https://www.npmjs.com/package/@earendil-works/pi-coding-agent)
kernel behind an authenticated HTTP API with verbose-tiered output and resumable SSE
streaming. It implements the reviewed-APPROVED backend dev-design.

> ⚠️ **Draft — not for merge.** Opened for review of the design/implementation only.

The Go kernel of octo-server is **not modified**. The service lives entirely under
`agent-runtime/`, depends on pi via npm, and does not fork or vendor it.

## What's included

- **Single resident Fastify process** holding in-process pi agent sessions (route ①,
  `createAgentSession`).
- **Bearer auth** (`Authorization: Bearer <token>`, compat `x-agent-token`) via a global
  `onRequest` hook; `GET /health` public. Missing/malformed/unknown token → `401` with the
  `{ok:false,error}` envelope. Constant-time token compare; tokens never logged.
- **Endpoints:** `POST /v1/agent` (sync + SSE), `GET /v1/sessions`, `GET /v1/sessions/:key`,
  `POST /v1/sessions/:key/abort`, `GET /health`. Unified `{ok,data?,error?}` envelope.
- **Verbose three-tier projection** (`off` / `on` / `full`) over the pi event stream — pure
  orchestration layer, kernel untouched.
- **SSE** with a per-session monotonic event-id (dropped events don't consume ids), a bounded
  ring buffer, `Last-Event-ID` replay of the `(n, tail]` window, and a `resync_required` frame
  when a reconnect falls outside the window.
- **`AgentBackend` seam** decoupling HTTP/verbose/SSE from the kernel: in-process pi backend,
  a deterministic backend (tests + offline smoke), and route ② (rpc subprocess) reserved.

## Tests / verification

- `npm test` — 26 passing across four suites: bearer 401, endpoint contracts, verbose tiers,
  and SSE out-of-order reconnect replay + `resync_required`. `npm run typecheck` clean.
- Red→green verified: neutralizing the auth hook / verbose tiering / SSE replay turns the
  corresponding assertions red; restoring turns them green.
- Ran the server locally and exercised `POST /v1/agent` end-to-end: `/health`, 401 without
  token, sync `off`/`full` (trace populated), SSE stream (gapless ids + terminal `done`),
  reconnect from `Last-Event-ID` replaying only the tail, and owner-scoped `GET /v1/sessions`.

Running **real** LLM turns via the pi backend needs a configured model provider
(`auth.json` / `models.json`); the deterministic backend covers the full HTTP contract offline.

## Comprehension

1. **What breaks if this is wrong?** Nothing in the existing Go server — the change is purely
   additive under `agent-runtime/` and imports no Go code. Risk is contained to the new service.
2. **Why this approach?** The dev-design chose an in-process `createAgentSession` (route ①)
   behind an `AgentBackend` abstraction so the HTTP/auth/verbose/SSE layers never couple to the
   kernel and route ② isolation can be swapped in later without touching them.
3. **What's deferred?** SQLite/FTS5 cross-session query ACL, rich diff/checkpoint/rollback
   endpoints, cron/subagent/template orchestration modules, and route ② — all post-MVP per
   dev-design §10.

### DIFF
--- a/agent-runtime/.gitignore
+++ b/agent-runtime/.gitignore
@@ -1,0 +1,7 @@
+node_modules/
+dist/
+*.log
+.env
+.env.*
+coverage/
+sessions/
--- a/agent-runtime/README.md
+++ b/agent-runtime/README.md
@@ -1,0 +1,92 @@
+# @octo/agent-runtime
+
+A lightweight, resident HTTP runtime that embeds the [`@earendil-works/pi-coding-agent`](https://www.npmjs.com/package/@earendil-works/pi-coding-agent)
+kernel behind an authenticated API with verbose-tiered output and resumable SSE
+streaming.
+
+This is a **new, additive orchestration-layer service**. It does not fork pi and
+does not modify the Go kernel of octo-server — it depends on pi via npm and lives
+entirely under `agent-runtime/`. It implements the backend dev-design
+(`research/DESIGN-backend-devdesign.md`, reviewed APPROVED).
+
+## What it does
+
+- **Single resident Node process** = a thin Fastify HTTP server + one or more
+  resident pi agent sessions, sharing the orchestration-layer memory state.
+- **Bearer auth** (`Authorization: Bearer <token>`, compat `x-agent-token`) on a
+  global `onRequest` hook; `GET /health` is public. Missing/malformed/unknown
+  token → `401` with the `{ok:false,error}` envelope.
+- **Verbose three-tier projection** (`off` / `on` / `full`) over the pi event
+  stream, without touching the kernel.
+- **SSE streaming** with a per-session monotonic event-id, a bounded ring buffer,
+  `Last-Event-ID` replay, and a `resync_required` frame when a reconnect falls
+  outside the buffer window.
+
+## Endpoints
+
+| Method | Path | Auth | Description |
+|---|---|---|---|
+| `POST` | `/v1/agent` | ✅ | Run a turn. Sync by default; `stream:true` → SSE. |
+| `GET`  | `/v1/sessions` | ✅ | List the caller's resident sessions. |
+| `GET`  | `/v1/sessions/:key` | ✅ | Session state (mirrors pi `RpcSessionState`). |
+| `POST` | `/v1/sessions/:key/abort` | ✅ | Abort the in-flight turn. |
+| `GET`  | `/health` | ❌ | Liveness/readiness probe. |
+
+`POST /v1/agent` body:
+
+```jsonc
+{
+  "message": "string, required (unless resume:true)",
+  "sessionKey": "string?, gated by allowRequestSessionKey",
+  "verbose": "off | on | full",   // overrides the global default
+  "stream": "boolean?, default false",
+  "resume": "boolean?, reconnect to an existing session's stream",
+  "timeoutSeconds": "number?"
+}
+```
+
+Response envelope: `{ ok, data? , error? }`. In sync mode, `data` is
+`{ sessionKey, result, usage, trace }` (`trace` is populated per verbose tier and
+empty at `off`). In stream mode, each SSE frame carries an `id:` line plus a
+`data:` line whose JSON `type` names the event; the terminal frame is
+`{ "type":"done", "result":..., "usage":... }`.
+
+## Configuration (environment)
+
+| Var | Default | Meaning |
+|---|---|---|
+| `AGENT_RUNTIME_TOKEN` / `AGENT_RUNTIME_TOKENS` | — | Accepted bearer token(s); comma list for rotation. No token ⇒ every authed route returns 401 (fail closed). |
+| `AGENT_RUNTIME_HOST` | `127.0.0.1` | Listen host. |
+| `AGENT_RUNTIME_PORT` | `8787` | Listen port. |
+| `AGENT_RUNTIME_VERBOSE_DEFAULT` | `off` | Global verbose tier. |
+| `AGENT_RUNTIME_ALLOW_REQUEST_SESSION_KEY` | `false` | Accept a client `sessionKey` (OpenClaw-aligned default off). |
+| `AGENT_RUNTIME_ALLOWED_SESSION_KEY_PREFIXES` | — | Required prefixes when the above is `true`. |
+| `AGENT_RUNTIME_SSE_BUFFER_SIZE` | `512` | Per-session SSE ring capacity. |
+| `AGENT_RUNTIME_BACKEND` | `fake` | `pi` for the in-process kernel (route ①); `fake` for the deterministic backend. |
+
+## Kernel backends (the `AgentBackend` seam)
+
+The HTTP / auth / verbose / SSE layers sit on the `AgentBackend` interface
+(`prompt` / `abort` / `getState` / `subscribe`). Three implementations:
+
+- **`PiAgentBackend`** (route ①) — in-process `createAgentSession`. Running real
+  turns requires a configured model provider (pi `auth.json` / `models.json`).
+- **`FakeAgentBackend`** — deterministic, no model provider; emits a pi-shaped
+  event stream. Used by the test suite and for local smoke runs, so the full HTTP
+  contract, auth, verbose projection and SSE replay are exercisable offline.
+- Route ② (a `./rpc-entry` subprocess) is reserved behind the same interface.
+
+## Develop
+
+```bash
+npm install
+npm test        # vitest: auth 401, endpoint contracts, verbose tiers, SSE replay
+npm run typecheck
+AGENT_RUNTIME_TOKEN=dev npm start   # boots on 127.0.0.1:8787 with the fake backend
+```
+
+## Scope
+
+MVP + streaming increment per dev-design §10. Post-MVP (deferred, jointly signed
+off): SQLite/FTS5 cross-session query ACL, rich diff/checkpoint/rollback
+endpoints, cron/subagent/template orchestration modules, route ② isolation.
--- a/agent-runtime/package-lock.json
+++ b/agent-runtime/package-lock.json
@@ -1,0 +1,4600 @@
+{
+  "name": "@octo/agent-runtime",
+  "version": "0.1.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@octo/agent-runtime",
+      "version": "0.1.0",
+      "dependencies": {
+        "fastify": "^5.2.1"
+      },
+      "devDependencies": {
+        "@types/node": "^22.10.5",
+        "tsx": "^4.19.2",
+        "typescript": "^5.7.3",
+        "vitest": "^2.1.8"
+      },
+      "engines": {
+        "node": ">=20"
+      },
+      "optionalDependencies": {
+        "@earendil-works/pi-coding-agent": "0.80.10"
+      }
+    },
+    "node_modules/@earendil-works/pi-coding-agent": {
+      "version": "0.80.10",
+      "resolved": "https://registry.npmjs.org/@earendil-works/pi-coding-agent/-/pi-coding-agent-0.80.10.tgz",
+      "integrity": "sha512-aL4apbupCHiVLSXASXvRzH4Q2vmtfrDa+0s909CJuVu/GgGylbDzr7oyF1mPmip5E+VxYYxKWmph4hV04wUcQg==",
+      "hasShrinkwrap": true,
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "@earendil-works/pi-agent-core": "^0.80.10",
+        "@earendil-works/pi-ai": "^0.80.10",
+        "@earendil-works/pi-tui": "^0.80.10",
+        "@silvia-odwyer/photon-node": "0.3.4",
+        "chalk": "5.6.2",
+        "cross-spawn": "7.0.6",
+        "diff": "8.0.4",
+        "glob": "13.0.6",
+        "highlight.js": "10.7.3",
+        "hosted-git-info": "9.0.3",
+        "ignore": "7.0.5",
+        "jiti": "2.7.0",
+        "minimatch": "10.2.5",
+        "proper-lockfile": "4.1.2",
+        "semver": "7.8.0",
+        "typebox": "1.1.38",
+        "undici": "8.5.0",
+        "yaml": "2.9.0"
+      },
+      "bin": {
+        "pi": "dist/cli.js"
+      },
+      "engines": {
+        "node": ">=22.19.0"
+      },
+      "optionalDependencies": {
+        "@mariozechner/clipboard": "0.3.9"
+      }
+    },
+    "node_modules/@earendil-works/pi-coding-agent/node_modules/@anthropic-ai/sdk": {
+      "version": "0.91.1",
+      "resolved": "https://registry.npmjs.org/@anthropic-ai/sdk/-/sdk-0.91.1.tgz",
+      "integrity": "sha512-LAmu761tSN9r66ixvmciswUj/ZC+1Q4iAfpedTfSVLeswRwnY3n2Nb6Tsk+cLPP28aLOPWeMgIuTuCcMC6W/iw==",
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "json-schema-to-ts": "^3.1.1"
+      },
+      "bin": {
+        "anthropic-ai-sdk": "bin/cli"
+      },
+      "peerDependencies": {
+        "zod": "^3.25.0 || ^4.0.0"
+      },
+      "peerDependenciesMeta": {
+        "zod": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@earendil-works/pi-coding-agent/node_modules/@aws-crypto/crc32": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/crc32/-/crc32-5.2.0.tgz",
+      "integrity": "sha512-nLbCWqQNgUiwwtFsen1AdzAtvuLRsQS8rYgMuxCrdKf9kOssamGLuPwyTY9wyYblNr9+1XM8v6zoDTPPSIeANg==",
+      "license": "Apache-2.0",
+      "optional": true,
+      "dependencies": {
+        "@aws-crypto/util": "^5.2.0",
+        "@aws-sdk/types": "^3.222.0",
+        "tslib": "^2.6.2"
+      },
+      "engines": {
+        "node": ">=16.0.0"
+      }
+    },
+    "node_modules/@earendil-works/pi-coding-agent/node_modules/@aws-crypto/sha256-browser": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/sha256-browser/-/sha256-browser-5.2.0.tgz",
+      "integrity": "sha512-AXfN/lGotSQwu6HNcEsIASo7kWXZ5HYWvfOmSNKDsEqC4OashTp8alTmaz+F7TC2L083SFv5RdB+qU3Vs1kZqw==",
+      "license": "Apache-2.0",
+      "optional": true,
+      "dependencies": {
+        "@aws-crypto/sha256-js": "^5.2.0",
+        "@aws-crypto/supports-web-crypto": "^5.2.0",
+        "@aws-crypto/util": "^5.2.0",
+        "@aws-sdk/types": "^3.222.0",
+        "@aws-sdk/util-locate-window": "^3.0.0",
+        "@smithy/util-utf8": "^2.0.0",
+        "tslib": "^2.6.2"
+      }
+    },
+    "node_modules/@earendil-works/pi-coding-agent/node_modules/@aws-crypto/sha256-js": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/sha256-js/-/sha256-js-5.2.0.tgz",
+      "integrity": "sha512-FFQQyu7edu4ufvIZ+OadFpHHOt+eSTBaYaki44c+akjg7qZg9oOQeLlk77F6tSYqjDAFClrHJk9tMf0HdVyOvA==",
+      "license": "Apache-2.0",
+      "optional": true,
+      "dependencies": {
+        "@aws-crypto/util": "^5.2.0",
+        "@aws-sdk/types": "^3.222.0",
+        "tslib": "^2.6.2"
+      },
+      "engines": {
+        "node": ">=16.0.0"
+      }
+    },
+    "node_modules/@earendil-works/pi-coding-agent/node_modules/@aws-crypto/supports-web-crypto": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/supports-web-crypto/-/supports-web-crypto-5.2.0.tgz",
+      "integrity": "sha512-iAvUotm021kM33eCdNfwIN//F77/IADDSs58i+MDaOqFrVjZo9bAal0NK7HurRuWLLpF1iLX7gbWrjHjeo+YFg==",
+      "license": "Apache-2.0",
+      "optional": true,
+      "dependencies": {
+        "tslib": "^2.6.2"
+      }
+    },
+    "node_modules/@earendil-works/pi-coding-agent/node_modules/@aws-crypto/util": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/util/-/util-5.2.0.tgz",
+      "integrity": "sha512-4RkU9EsI6ZpBve5fseQlGNUWKMa1RLPQ1dnjnQoe07ldfIzcsGb5hC5W0Dm7u423KWzawlrpbjXBrXCEv9zazQ==",
+      "license": "Apache-2.0",
+      "optional": true,
+      "dependencies": {
+        "@aws-sdk/types": "^3.222.0",
+        "@smithy/util-utf8": "^2.0.0",
+        "tslib": "^2.6.2"
+      }
+    },
+    "node_modules/@earendil-works/pi-coding-agent/node_modules/@aws-sdk/client-bedrock-runtime": {
+      "version": "3.1048.0",
+      "resolved": "https://registry.npmjs.org/@aws-sdk/client-bedrock-runtime/-/client-bedrock-runtime-3.1048.0.tgz",
+      "integrity": "sha512-u+NT61JZEkRFtpL0CAw1N1dwxnaLgwVXQl/zjJxTGgLyS/jTIdg2SdoEoCTHxgDyCnqa1HEi9QOoE9/pYRNpOQ==",
+      "license": "Apache-2.0",
+      "optional": true,
+      "dependencies": {
+        "@aws-crypto/sha256-browser": "5.2.0",
+        "@aws-crypto/sha256-js": "5.2.0",
+        "@aws-sdk/core": "^3.974.11",
+        "@aws-sdk/credential-provider-node": "^3.972.42",
+        "@aws-sdk/eventstream-handler-node": "^3.972.16",
+        "@aws-sdk/middleware-eventstream": "^3.972.12",
+        "@aws-sdk/middleware-websocket": "^3.972.19",
+        "@aws-sdk/token-providers": "3.1048.0",
+        "@aws-sdk/types": "^3.973.8",
+        "@smithy/core": "^3.24.2",
+        "@smithy/fetch-http-handler": "^5.4.2",
+        "@smithy/node-http-handler": "^4.7.2",
+        "@smithy/types": "^4.14.1",
+        "tslib": "^2.6.2"
+      },
+      "engines": {
+        "node": ">=20.0.0"
+      }
+    },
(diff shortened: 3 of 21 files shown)


ploy-elison:
改为独立 agent 项目、托管到 codex.mlamp.cn 的 monorepo，不在 octo-server 内实现，代码迁移至独立仓库。关闭本 PR。

Jerry-Xin (changes requested):
(Superseded — this review contained local workspace paths and has been replaced. See the current REQUEST_CHANGES review with repository-relative paths for the actionable findings.)

Jerry-Xin (changes requested):
The `agent-runtime/` service is a clean, well-tested addition, but ownership isolation, turn serialization, and the SSE verbose-replay contract have blocking correctness/security defects that must be fixed before merge. (Also note: PR is marked Draft, and the `dependency-review` CI check is failing.)

## 🔴 Blocking

1. **Critical — Cross-owner session access via `POST /v1/agent`.** In `agent-runtime/src/session-registry.ts`, `getOrCreate()` returns an existing session by `sessionKey` alone and never checks `ownerId` on the existing-session branch. `POST /v1/agent` (in `agent-runtime/src/app.ts`) routes through `getOrCreate` rather than the owner-scoped `get()`. So when request-supplied session keys are enabled (`allowRequestSessionKey=true`), a second authenticated token that supplies a known key reaches and prompts/streams another owner's resident backend. Compounding this, the prefix gate in `app.ts` treats an **empty** `allowedSessionKeyPrefixes` list as allow-all, contrary to the "required prefixes" intent. Fix: make registry keys owner-scoped (or reject an ownerId mismatch inside `getOrCreate`), require a non-empty configured prefix list when request keys are allowed, and add a two-token isolation test for this POST path.

2. **Critical — Concurrent turns corrupt session/response boundaries.** Neither the sync path nor the fresh-stream path serializes or rejects a second prompt on the same backend. Sync turns subscribe independently, and every new stream blindly sets itself as the driver (`record.streaming = true`) and appends to the ring. Two concurrent requests on one session will interleave/consume each other's events, and multiple streaming "drivers" append identical backend events to the ring more than once. Fix: add per-session turn locking/queuing with an explicit busy policy, plus concurrency tests for sync/sync, stream/stream, and sync/stream.

3. **Critical — Verbose-level reconnect replay is not lossless.** In `agent-runtime/src/sse.ts`, `append()` only buffers an event if it survives the *original* connection's projection (dropped events return null and are never stored). So a stream started at `off` permanently discards tool/text-delta/thinking events, and a reconnect at `full` cannot re-project them as the doc comment claims; conversely reconnecting at a lower tier skips retained ids and produces gaps. Fix: buffer the raw event independently of connection-specific projection and define ids against a stable per-session event sequence (or forbid changing the verbose level across a replay). Add cross-tier reconnect tests.

4. **Critical — Declared Node support is incompatible with the pi backend.** `agent-runtime/package.json` declares engines `node >=20`, but the pinned `@earendil-works/pi-coding-agent` (and transitive deps like `undici` and the pi core packages) require `node >=22.19.0`. A "supported" Node 20 deploy installs with warnings then fails on the primary backend. Fix: raise the engine requirement to `>=22.19.0` and align the README/deploy docs.

## 💬 Non-blocking

- **Internal exception text is returned to clients** in both sync and SSE error paths (`app.ts`). Prefer logging server-side detail and returning a generic internal error message to callers.
- **No session TTL / count cap / eviction, and no client-disconnect cleanup.** Sessions live until process close, and a live SSE observer/subscription is only removed on turn settlement — an abandoned/disconnected client is not torn down. Add a disconnect handler (unsubscribe + stop writing on socket close), a concurrent-connection/session cap, and TTL eviction to bound resident memory.
- **`ownerIdForToken` uses a collision-prone 32-bit FNV hash** (`auth.ts`). Use a cryptographic digest with adequate output length so distinct tokens can't collide onto the same owner id.
- **Dependency posture:** the pi dependency currently pulls in packages flagged with one high and one moderate advisory (this is also what the failing `dependency-review` check is surfacing). Assess/update before release.

## ✅ Highlights

- The `AgentBackend` seam cleanly decouples HTTP/verbose/SSE orchestration from the pi kernel.
- Auth is globally mounted via an `onRequest` hook, fails closed when no token is configured, uses a constant-time compare, whitelists only `/health`, and never logs tokens.
- The verbose `off` default plus the three-tier projection avoids leaking tool args/reasoning by default.
- Tests are fast and well organized; all 26 pass and typecheck is clean. (Gaps worth adding: cross-owner isolation, concurrency, cross-tier replay, and SSE client-disconnect cleanup.)

lml2468 (changes requested):
# PR Review — PR #629 (Mininglamp-OSS/octo-server)

Reviewed at head `1a361c2a` against merge-base `f9fd79b5` (base `main`). **DRAFT — author marked "not for merge; review of design/impl only."** Reviewing accordingly: one real isolation defect to fix before this graduates, plus two smaller notes.
**Build/test:** `npm ci` ✅, `npm run typecheck` (tsc) ✅, `npm test` → **26/26 pass** (auth 401, endpoint contracts, verbose tiers, SSE reconnect/replay). Node 22. The Go kernel is untouched — change is entirely additive under `agent-runtime/`.
Classification: new externally-reachable HTTP service with bearer auth + per-owner session isolation + SSE. Reviewed in **SECURITY** emphasis (auth/tenant-isolation blast radius).

## 1. Spec compliance — ✅ (with one contract violation, see Gate 2)
Graded against the PR description + the backend dev-design it cites. Endpoints, `{ok,data?,error?}` envelope, verbose three-tier projection, SSE monotonic-id + bounded ring + `Last-Event-ID` replay + `resync_required`, and the `AgentBackend` seam (fake/pi) are all present and match the description. Deferred items (SQLite/FTS5 ACL, diff/checkpoint, route ②) are honestly called out. No over-build. The one deviation is that the code violates its **own** stated isolation invariant (below) — a defect, not a scope disagreement.

## 2. Code quality — Request changes

### P1 (BLOCKER for graduation) — cross-owner session access: `getOrCreate` skips the owner check that `get()` enforces
`SessionRegistry.get()` is correctly owner-scoped (`session-registry.ts:93` → `record.ownerId !== ownerId → undefined`), and the file's own docstring states the contract: *"Cross-session reads are owner-scoped here … default-deny, only the owning token can see a session."* But `getOrCreate()` (`session-registry.ts:71-73`) returns an existing record **purely by `sessionKey`, with no `ownerId` comparison**:
```ts
const existing = this.sessions.get(sessionKey);
if (existing) return existing;   // <-- returns another owner's session
```
`POST /v1/agent` routes through `getOrCreate` (`app.ts:96`). With `allowRequestSessionKey=true` (a documented, supported config — `AGENT_RUNTIME_ALLOW_REQUEST_SESSION_KEY`, README line 62; the client key is then used verbatim when it matches an allowed prefix, `app.ts:69-73`), **owner B can name owner A's `sessionKey` and drive a turn on A's session** — reaching A's backend, A's conversation continuity, A's SSE ring (replay of A's buffered output) and `lastResult`.

**Repro (ran it, deterministic backend):** config `tokens:["tokenA","tokenB"], allowRequestSessionKey:true, allowedSessionKeyPrefixes:[]`.
1. `POST /v1/agent` as `tokenA`, `sessionKey:"shared:proj"` → 200, record created owned by A.
2. `POST /v1/agent` as `tokenB`, `sessionKey:"shared:proj"` → **200** (B drove a turn on A's session).
3. `GET /v1/sessions` shows `["shared:proj"]` for **A only**, `[]` for B — proving the record is A's, yet B operated it through `getOrCreate`.

So the read/list/abort endpoints are default-deny, but the create/prompt path is not — an inconsistent-isolation-per-path defect (exactly the class the repo's isolation rules warn about). **Fix:** make `getOrCreate` enforce ownership like `get()` — if an existing record's `ownerId !== ownerId`, either reject (`invalid_session_key`/`not_found`) or, if key sharing across owners is never intended, key the map by `(ownerId, sessionKey)`. Add a two-token test at `allowRequestSessionKey=true` (see Gate 3).

Note: the **default** config (`allowRequestSessionKey=false`) is safe — every key is namespaced `default:${ownerId}` (`session-registry.ts:67`), so owners can't collide. The defect requires the opt-in key mode. That bounds severity to "blocker for the config the PR ships support for," not "default RCE."

### P2 — `ownerIdForToken` is a 32-bit FNV hash; two distinct tokens can map to one owner
`ownerIdForToken` (`auth.ts:73-81`) folds the token to a 32-bit value. Owner identity — the isolation key for every session — therefore lives in a 2³² space; two **distinct valid tokens** colliding would silently share sessions. Practically low risk (a birthday collision needs ~77k configured tokens; I searched 200k synthetic tokens and the collision point is ~2¹⁶–2¹⁷, far above any real token list), so this is a **note, not a blocker** — but since it's an isolation identifier, prefer a full-width digest (e.g. `sha256(token)` hex, or a random per-token id) so owner identity can never collide regardless of token count. HIGH-confidence on the mechanism, LOW on practical reachability.

### P2 / NIT — `tokenAccepted` builds a `padded` buffer it never uses; the stated timing-safety goal isn't met
`auth.ts:22-24` allocates `padded` (comment: *"so the length check itself doesn't short-circuit and leak timing"*) but the actual comparison is still `candidateBuf.length === tokenBuf.length && timingSafeEqual(...)` — the length check **does** short-circuit, and `padded` is dead code. So a candidate's length is still distinguishable by timing, and the buffer is pure overhead. Tokens are high-entropy secrets so the practical leak is negligible, but either (a) drop the dead `padded` and keep the honest length-guarded `timingSafeEqual` (and fix the comment), or (b) actually compare `candidateBuf` against `padded` unconditionally to remove the length short-circuit. Right now the code claims a property it doesn't have.

- **Auth fail-closed:** empty token list → every authed route 401 (`auth.ts:62`); `/health` is the only public path; missing/malformed scheme → 401 with the envelope; server warns when no token configured (`server.ts:11`). Confirmed by `auth.test.ts` (6 tests).
- **Session-key prefix gate:** with `allowRequestSessionKey=true` + a prefix list, a non-matching key → `invalid_session_key` 400 (`app.ts:72`), and with the flag off a client key is ignored and routed to `default:${owner}` — both covered by `endpoints.test.ts`.
- **SSE id integrity:** dropped (projection-`null`) events don't consume an id (`sse.ts:56-59`), replay resends `(lastEventId, tail]` and emits `resync_required` when the window slid (`sse.ts:73-82`) — `sse-replay.test.ts` (8 tests) covers out-of-order reconnect + resync. No id reuse.
- **Tokens never logged:** `logger:false` (`app.ts:46`); no token in envelopes or session metadata; `ownerIdForToken` returns an opaque digest, not the raw token.
- **pi kernel untouched / lazy:** pi is a lazy `import()` (`app.ts:33`), optional dep — the runtime and tests work on the deterministic backend without the kernel.

## 3. Test assessment
26 tests, well-targeted for the happy paths and single-owner behavior; red→green claims in the description are plausible and the suite runs clean. **The gap that let the P1 through:** there is no **multi-owner** test with `allowRequestSessionKey=true`. `endpoints.test.ts:53` exercises the prefix gate but with a **single token**, so it never checks that owner B is denied owner A's key. A test that (a) creates a session as token A with an allowed key, then (b) asserts token B naming the same key does **not** reach A's session, would have caught this and should ship with the fix. (Secondary: no test for the FNV owner-id being collision-free, which is inherent to the hash choice.)

## 4. Coverage / blind spots
- Verified by `npm ci` + typecheck + full suite + a custom two-owner harness (ran, then removed) + reading all 15 `src/` files.
- NOT exercised: the **real** pi backend (needs a model provider / `auth.json`+`models.json`) — only the deterministic backend ran, per the PR's own note. The `pi.ts` event translation is unverified against a live kernel; recommend a staging smoke with a configured provider.

## Verdict
**REQUEST_CHANGES** — as a design/impl review of this draft: the architecture (AgentBackend seam, verbose projection, SSE ring/replay) is clean and well-tested, auth fails closed, and the Go kernel is untouched. **One real isolation defect must be fixed before this graduates from draft:** `getOrCreate` bypasses the owner check that `get()` enforces, so under the supported `allowRequestSessionKey=true` mode one owner can drive/observe another owner's session (proven by repro) — a per-path isolation break contradicting the registry's own default-deny contract. Fix `getOrCreate` to be owner-scoped (or key the map by `(ownerId, sessionKey)`) and add the two-owner regression test. Also worth addressing: widen `ownerIdForToken` beyond 32-bit FNV, and remove the dead `padded` buffer whose timing-safety comment overstates what the code does.
Suggested decision: `Changes requested` — no merge intended yet (draft); these are the items to close before it's merge-ready.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #629 (octo-server)

## Summary

This PR adds a new `agent-runtime/` package — a lightweight resident HTTP runtime embedding the pi coding-agent kernel behind a Fastify server with Bearer auth, verbose-tiered output projection, and resumable SSE streaming. It is a well-structured, additive service that sits entirely on the `AgentBackend` seam and never touches the Go kernel. The architecture (auth hook, session registry, SSE ring buffer, pluggable backends) is clean and well-tested.

However, this is a security-sensitive PR, and the cross-tenant session lookup in `getOrCreate` has an authorization gap that must be fixed before merge: when `allowRequestSessionKey` is enabled, any authenticated caller can hijack another user's session by sending their session key.

## Verification

- ✅ **Diff scope** — all 21 files are new additions (status=A), zero deletions. No existing code paths are modified.
- ✅ **Auth hook coverage** — `buildAuthHook` registered as global `onRequest` hook at `agent-runtime/src/app.ts:53`; `PUBLIC_PATHS` only whitelists `/health`. All `/v1/*` routes require auth.
- ✅ **Owner-scoped reads** — `GET /v1/sessions` filters by `ownerId` at `session-registry.ts:104`; `GET /v1/sessions/:key` and `POST /v1/sessions/:key/abort` use `registry.get()` which checks `record.ownerId` at `:96`.
- ✅ **Fail-closed default** — `loadConfig` produces empty `tokens[]` when no env var is set; `server.ts:13` warns and all authenticated routes return 401.
- ✅ **SSE ring contract** — `SseBuffer` assigns gapless monotonic ids, drops projected events without consuming ids, and correctly emits `resync_required` on eviction. Covered by unit tests in `sse-replay.test.ts`.
- ✅ **Test isolation** — tests inject `FakeAgentBackend` via `BuildAppOptions.factory`; production code uses the lazy `defaultFactory`. No test-to-production leakage.
- ✅ **Constant-time token comparison** — `tokenAccepted` at `auth.ts:23` uses `timingSafeEqual` with a same-length padded buffer to avoid length-based timing leaks.

Static analysis only at head `1a361c2a`; build and tests not executed in this environment.

## Findings

One P1 blocker and two P2 items below.

### P1 — Cross-tenant session hijack via `getOrCreate` (`agent-runtime/src/session-registry.ts:82`)

`getOrCreate(sessionKey, ownerId)` returns any existing session matching `sessionKey` without verifying `record.ownerId === ownerId`. When `allowRequestSessionKey` is enabled (the opt-in gate for client-chosen session keys per `config.ts:21`), any authenticated caller can send another user's session key in the request body and receive their `SessionRecord` — including the `AgentBackend` (full read/write access to the agent conversation), the SSE ring buffer (all buffered events), and `lastResult` (the last settled turn output).

The ownership check is correctly applied in `get()` (used by session detail and abort endpoints at `:96`), but `POST /v1/agent` at `app.ts:85` goes through `getOrCreate` which skips it entirely. The existing endpoint test at `endpoints.test.ts:76` only verifies the key-prefix gate, not cross-owner key reuse.

```ts
// Current (broken):
async getOrCreate(sessionKey: string, ownerId: string): Promise<SessionRecord> {
  const existing = this.sessions.get(sessionKey);
  if (existing) return existing; // ← no ownership check
  // ...
}

// Fix: check ownership, reject or scope by owner
async getOrCreate(sessionKey: string, ownerId: string): Promise<SessionRecord> {
  const existing = this.sessions.get(sessionKey);
  if (existing && existing.ownerId === ownerId) return existing;
  if (existing && existing.ownerId !== ownerId) {
    throw new Error("session key already owned by another user");
  }
  // ...create new
}
```

Diff-scope: **new** — this is a new code path introduced by this PR. The design intent (OpenClaw-aligned session-key gate) does not account for cross-tenant reuse.

### P2 — `ownerIdForToken` uses FNV-1a 32-bit: collision-prone and predictable (`agent-runtime/src/auth.ts:75`)

FNV-1a with 32-bit output has a birthday-bound collision probability of ~50% at ~77k distinct tokens. Two tokens that collide produce the same `owner_<hex>` id, the same default session key (`default:owner_<hex>`), and therefore full mutual session access — even without `allowRequestSessionKey`.

Additionally, FNV-1a is non-cryptographic and trivially invertible. Given the algorithm (visible in open source) and one known token, an attacker can cheaply brute-force inputs that produce the same 32-bit hash, then present the colliding token to gain access to the victim's default session.

Fix: replace with a cryptographic digest — e.g. `createHash('sha256').update(token).digest('hex').slice(0, 16)` for a collision-resistant, non-reversible owner id.

### P2 — SSE reconnect-to-live-turn has a narrow event-loss race (`agent-runtime/src/app.ts:220`)

In the reconnect-to-live path of `streamTurn`, the code replays the ring buffer tail (sync), then enters `await new Promise` which calls `record.addListener` inside its executor. Between the replay loop and the `addListener` registration, there is a microtask-yielding `await`. If the turn settles during this window (backend emits `settled` or `error`), the notification fires before the listener is registered. The connection then hangs — it never receives the terminal event, `finish` is never called, and the SSE stream stays open until the client disconnects or TCP times out.

The window is narrow (one synchronous block) but the consequence is a stuck connection with no recovery path. Consider adding a post-subscribe check: if `!record.streaming` after `addListener`, immediately deliver the terminal `done` frame and call `raw.end()`.

## Human-verify

1. The `@earendil-works/pi-coding-agent` optional dependency is loaded lazily via dynamic `import()` at `backends/pi.ts:126`. The pi kernel's `createAgentSession` API, event shape, and `SessionManager.inMemory` are assumed from the dev-design doc — not verifiable from this repo alone. Confirm the pi event types (`agent_settled`, `agent_end`, `tool_execution_start`, etc.) at `translatePiEvent` (`backends/pi.ts:52`) match the actual pi kernel's emitted events. **Not** a merge blocker.

2. The README documents `AGENT_RUNTIME_BACKEND=pi` as the production path. Confirm the pi kernel's model provider configuration (`auth.json` / `models.json`) is handled by the deployment layer and that the `optionalDependencies` pin (`0.80.10`) is compatible with the deployed pi version. **Not** a merge blocker.

- **C1 — Dual-path parity**: N/A. No paired create/delete paths. `disposeAll` at `session-registry.ts:113` correctly drains all sessions on shutdown.
- **C2 — Control-flow ordering**: N/A. Each route handler is self-contained; no nested or reused control-flow logic.
- **C3 — Authorization boundary**: **HIT** — see P1. `getOrCreate` is reachable from `POST /v1/agent` without ownership verification, while `get()` correctly enforces it. Authorization asymmetry across sibling entry points.
- **C4 — Authorization lifecycle**: N/A. No container/member hierarchy. Sessions are flat, scoped to `ownerId`.
- **C5 — Build/runtime parity**: Clear. `tsconfig.json` uses `NodeNext` matching `"type": "module"`. `vitest.config.ts` uses `extensionAlias` to resolve `.js` imports to `.ts` sources.
- **C6 — Governance/docs self-consistency**: Clear. README configuration table matches `loadConfig` env var names and defaults exactly.

## Data-flow tracing

- **`request.ownerId`** → set by `buildAuthHook` at `auth.ts:58` via `ownerIdForToken(token)` → consumed by `resolveSessionKey` at `app.ts:67` and all route handlers → flows correctly to `registry.getOrCreate/get/list`.
- **`SessionRecord.backend`** → created by `factory(sessionKey, ownerId)` in `session-registry.ts:88` → consumed by `syncTurn`/`streamTurn` via `record.backend.prompt()` and `record.backend.subscribe()`. Note: `ownerId` is passed to factory but ignored by both `PiAgentBackend.create` and `FakeAgentBackend` — backend has no ownership awareness; isolation is purely at the registry level.
- **SSE event-id** → assigned by `SseBuffer.append` at `sse.ts:55` → consumed by `streamTurn` for `serializeSse(id, frame)` writes → replayed via `SseBuffer.replay` at `:74`. Data flow correct; ids are monotonic and gapless.
- **`record.lastResult`** → set at `app.ts:260` on `message_end` events → consumed by `writeDone()` at `:195` for terminal `done` frames on reconnect. Data flow correct.

## Cross-round blocker review

N/A — first review round.

## Verdict: CHANGES_REQUESTED

The P1 cross-tenant session hijack in `getOrCreate` must be fixed before merge — it allows any authenticated caller to access another user's session when `allowRequestSessionKey` is enabled. The P2 items (FNV-1a hash, SSE race) are non-blocking but should be addressed in the same PR given the security-sensitive routing.

[Octo-Q] verdict: REQUEST_CHANGES — P1 cross-tenant session hijack via `getOrCreate` missing ownership check; must be fixed before merge.

yujiawei (changes requested):
# Code Review — PR #629 (octo-server)

New `agent-runtime/` package: a resident Fastify HTTP service embedding the pi coding-agent kernel behind Bearer auth, three-tier verbose projection, and resumable SSE streaming. Additive only — 21 new files, zero deletions, no Go code touched (the Go `Build`/`Vet`/`Test` CI jobs are correctly skipped). Reviewed at head `1a361c2a`; `npm ci` + `npm test` reproduced locally (26/26 pass, `tsc --noEmit` clean) on Node 22.

The architecture is clean and the `AgentBackend` seam is a genuine strength. However, this is a security-sensitive service and it carries four blocking defects — three correctness/authorization bugs in the orchestration layer plus a declared-runtime incompatibility. All four are confirmed against the source and lockfile below.

## 1. Spec compliance

**Spec: ✅** — The diff implements what the linked dev-design describes (auth hook, session registry, SSE ring + replay, verbose tiers, the `AgentBackend` seam with pi/fake backends) and stays within scope. No over-build, no missing MVP surface. Spec compliance is not the gate here — implementation correctness is.

## 2. Code quality — Blocking (P0/P1)

### 🔴 B1 (P0) — Cross-owner session access via `POST /v1/agent` (`src/session-registry.ts:72-73`)

`getOrCreate()` returns an existing session by `sessionKey` alone and never checks `ownerId`:

```ts
const existing = this.sessions.get(sessionKey);
if (existing) return existing;   // no ownerId check
```

`POST /v1/agent` (`src/app.ts:96`) routes through `getOrCreate`, not the owner-scoped `get()` (which *does* enforce `record.ownerId !== ownerId`, `session-registry.ts:95`). When `allowRequestSessionKey=true`, a second authenticated token that supplies another owner's key reaches and prompts/streams the victim's resident backend, SSE ring, and `lastResult`.

Compounding it, the prefix gate treats an **empty** `allowedSessionKeyPrefixes` list as allow-all (`src/app.ts:70`: `length === 0 || ...`), contradicting the "required prefixes" intent documented in `config.ts:21`. So enabling request keys without a prefix list opens *arbitrary* client-chosen keys.

Fix: scope registry keys by owner (or reject an `ownerId` mismatch inside `getOrCreate`); require a non-empty prefix list when `allowRequestSessionKey=true`; add a two-token isolation test against the POST path (the existing `endpoints.test.ts:108` isolation test only covers `GET /v1/sessions`, not `getOrCreate`).

### 🔴 B2 (P0) — Concurrent turns corrupt session/response boundaries (`src/app.ts:150-282`)

Neither the sync path nor the fresh-stream path serializes or rejects a second prompt on the same backend. Every fresh stream unconditionally sets `record.streaming = true` (`:238`) and appends to the shared ring; sync turns subscribe independently (`:157-169`). Two concurrent requests on one session interleave/consume each other's events, and two streaming "drivers" append the *same* backend events to the ring twice, breaking the monotonic-id contract for any later reconnect. The default session key is deterministic per token (`default:${ownerId}`), so two in-flight requests from one caller already collide — no misconfiguration required.

Fix: per-session turn lock/queue with an explicit busy policy (409/`invalid_request`), plus concurrency tests for sync/sync, stream/stream, and sync/stream.

### 🔴 B3 (P0) — Verbose-level reconnect replay is not lossless (`src/sse.ts:56-65`)

`append()` projects the event at the *original* connection's level and stores it **only if it survives** (`if (!frame) return null;` before `this.entries.push`). Dropped events are never buffered. This directly breaks the module's own documented contract (`sse.ts:8-12` and the reconnect path in `app.ts:204`), which claims raw events are re-projected at the reconnect level:

- A stream opened at `off` buffers nothing, so a reconnect at `full` can recover nothing.
- A stream at `on` never stores `thinking_delta`, so reconnecting at `full` silently loses reasoning frames — yet the ids are contiguous for `on`, so no `resync_required` fires and the loss is invisible.

Fix: buffer the **raw** event unconditionally and assign ids against a stable per-session sequence (project at read time), or forbid changing the verbose level across a replay. Add cross-tier reconnect tests (`off→full`, `on→full`).

### 🔴 B4 (P1) — Declared Node engine is incompatible with the pi backend (`package.json:7`)

`engines.node` is `>=20`, but the pinned `@earendil-works/pi-coding-agent@0.80.10` and its core deps (`@earendil-works/pi-agent-core`, `pi-ai`, `pi-tui`, `undici`) all require `node >=22.19.0` (confirmed in `package-lock.json`). A "supported" Node 20/21 deploy installs with warnings and then fails on the primary (`pi`) backend at runtime. Raise `engines.node` to `>=22.19.0` and align README/deploy docs.

## 3. Code quality — Non-blocking (P2 / suggestions)

- **P2 — Internal exception text leaked to clients.** Both error paths return `e.message` verbatim (`app.ts:176` sync, `app.ts:277` / `:267` SSE). Log the detail server-side; return a generic `internal` message.
- **P2 — `optionalDependencies` for the production kernel (`package.json`).** `pi` sits in `optionalDependencies`, so an install with `--no-optional` (or a resolution failure) yields a runtime that 500s per request instead of failing fast at boot. If `pi` is the production backend, move it to `dependencies`, or have the factory fail closed at startup when `backend=pi` and the import is unavailable.
- **P2 — `ownerIdForToken` uses a 32-bit FNV-1a hash (`auth.ts:73-81`).** Non-cryptographic and collision-prone (~50% collision at ~77k tokens); colliding tokens share an owner id and thus the same default session — cross-owner access even with `allowRequestSessionKey=false`. Use a cryptographic digest, e.g. `createHash('sha256').update(token).digest('hex').slice(0,16)`.
- **P2 — No session TTL / count cap / eviction, and no SSE client-disconnect cleanup.** Sessions live until process close; a live SSE listener is only removed on settlement, so an abandoned/disconnected client is never torn down (`app.ts:222-233` has no `raw.on('close', …)`). Add a disconnect handler (unsubscribe + stop writing on socket close), a session/connection cap, and TTL eviction to bound resident memory.
- **P2 — Dependency advisories.** CI `dependency-review` and `osv-scan` both fail on advisories pulled in transitively via `pi` (one high, one moderate). Assess/update before release; this is a real gate, not noise.

## 4. Verified correct / refuted candidates

- **SSE reconnect-to-live "listener-registration race" is a false positive.** The `new Promise` executor runs synchronously, so `record.addListener` (`app.ts:223`) registers in the *same* synchronous tick as the `!record.streaming` early-return check (`:213`), with no `await` between them. Backend events dispatch on a later tick and the driver's `notify()` (`:256`) fires the reconnect listener with `settled` before `finish()`. The listener is always registered before any settle can be observed — no stuck connection.
- **Auth hook** is globally mounted (`app.ts:51`), whitelists only `/health`, fails closed on empty token set, uses `timingSafeEqual` with equal-length padding, and never logs the raw token. Sound.
- **Owner-scoped reads** (`GET /v1/sessions`, `GET/POST /v1/sessions/:key`) correctly enforce ownership via `get()`/`list()`. The gap is isolated to the `getOrCreate` write path (B1).
- **SSE id/eviction/`resync_required`** logic in `SseBuffer` is correct *for a fixed level* — the flaw is purely the cross-level buffering in B3, not the id math.

## 5. Coverage notes / human-verify

- The pi event mapping in `translatePiEvent` (`backends/pi.ts:55`) is asserted from the dev-design, not verifiable in this repo — the `pi` kernel event shapes (`agent_settled`/`agent_end`/`tool_execution_*`, `SessionManager.inMemory`) should be confirmed against the deployed pi version. Not a merge blocker.
- Model-provider config (`auth.json`/`models.json`) for `backend=pi` is a deployment-layer concern; the default `backend=fake` means a misconfigured prod deploy silently serves mock data — consider failing fast when unset in production.
- This is a security-sensitive service (auth + multi-tenant session isolation). B1/B3 in particular warrant a human check that the fixes close the cross-tenant path end-to-end.

Advisory legs consulted (all corroborate the four blockers; B1/B2/B3/B4 independently re-verified against source): a whole-repo cross-family leg (received a truncated diff — only assessed README/lockfile — and independently flagged B4 + the `optionalDependencies` P2 + the `fake`-default P2), and a first-pass automated leg (flagged B1, the FNV-1a P2, and a since-refuted SSE-race). No leg is treated as silent approval.

## Overall verdict

**REQUEST_CHANGES** — four blocking defects (B1 cross-owner access, B2 concurrent-turn corruption, B3 lossy cross-tier replay, B4 Node-engine incompatibility) must be fixed before merge. The service is otherwise a clean, well-tested foundation. Note the PR is currently marked **Draft**.

PR #605, opened 2026-07-17
feat(cardtmpl): enrich docs access-request approval card + deny-reason channel

an9xyz:
## Summary

Reworks the docs "access request" approval card (通知助手 → 文档申请权限卡) into a structured layout across its **pending** and **terminal** states, and adds a reviewer **deny-reason** channel. Clicking 拒绝 in the web client now opens a required-reason dialog; the reason rides to the docs backend through the existing card/action inputs path. Companion frontend PR in `octo-web` (same branch `claude/docs-permission-card-template-bccd8h`).

## Related Issue

_None — self-initiated._

## Linked Spec

`.octospec/tasks/docs-approval-card-enrich/brief.md` (journal: `.octospec/journal/shared/docs-approval-card-enrich.md`).

## Changes

- **`pkg/cardtmpl/docs_action.go`** — `BuildDocsAccessRequestCard` now builds the enriched AC 1.5 body from `DocsApprovalContent`: header row (source label + colored status), large title, bold-actor banner, requester `ColumnSet` (optional Person avatar + name/role + timestamp), an `emphasis` reason box, and a **hidden `Input.Text id="deny_reason"`** that declares the reason channel. New `BuildDocsApprovalOutcomeCard` builds the approved/denied terminal card (header + title + good/attention result box; denied surfaces the reviewer reason). All caller strings stay `escapeMarkdown`'d; images stay https-only.
- **`modules/notify/model.go`** — additive optional `DocsCardFields.actor_avatar_url` (https-validated; docs-backend owned).
- **`modules/notify/card.go`** — `buildDocsAccessRequestCard` maps the fields + new zh-CN/en-US labels (header/status/banner/role/reason/result).
- **`modules/notify/action_finalizer.go`** — `buildTerminalDocument` uses the enriched outcome card for approved/denied and threads `event.Inputs["deny_reason"]` onto the denied card; cancelled/unavailable keep the plain rebuild. view-details is a body `ActionSet`, not a root action, so the terminal card keeps carrying **no** decision buttons.
- **`scripts/cardpreview/main.go`** — emits the three new states for offline SDK preview.
- Review fix: requester row is skipped for anonymous requests (no actor).

[...]

PR #632, opened 2026-07-22
feat(cardtmpl): L0 platform card base + docs.access-request@0.2.0 pilot

an9xyz:
## Summary

Establishes the **L0 platform card base** (`octo-card@1.0`) and migrates the first
**L2a card** — `docs.access-request@0.2.0` — through it end-to-end. This is a
vertical slice: the reusable "register → schema-validate → render → inject
metadata → re-validate → envelope" spine is now real, and the docs access-request
card is its proof-of-life. Adding card #2/#3 becomes "copy the pilot directory +
implement 3 methods", not a re-architecture.

Scheme B: the `NotifyReq` outer shape is untouched; only the docs
`access_requested` branch (gate-on) is rerouted internally through
`Registry.Render`.

## Linked Spec

- Task brief: `.octospec/tasks/cardtmpl-registry-pilot/brief.md`
- Authoritative contract: `docs/platform-card-base.md` (`octo-card@1.0`, §1–§16)

## Changes

- **L0 base** (`pkg/cardtmpl/`): `Template` interface (3 methods) + `TemplateMeta`;
  `Registry` (Register/SetDefault/Freeze/Lookup/List/RegisteredForTest) with an
  L2a owner allowlist and register-time sample self-check; `Registry.Render` as the
  single 8-step ingress; `metadata.octo.{protocol,template,variant,source}` +
  `metadata.webUrl` injection; `cardtmpl_build_total` metric; `AbsoluteHTTPSURL`
  as the single source of truth for the absolute-https rule.
- **L1 contract**: `docs.access-request@0.2.0` handoff artifacts (manifest / JSON
  Schema / interaction report / sample) embedded per-subpackage; `pending` is the
  only registered view (`approved/rejected` deferred to a `0.3.0` outcome PR).
- **L2a pilot** (`pkg/cardtmpl/docs_access_request/`): `Template` implementing
  Meta/Build/FallbackText; `Meta()` returns a defensive deep copy so a frozen
  registry stays immutable.
- **Production wiring**: `modules/notify` docs `access_requested` branch (gate-on)
  now renders via the Registry; composition root registers the pilot and
  `Freeze()`s (fail-close on missing template / bad schema).
- **C1 policy**: schema-level field errors for this card (non-https avatar,
  oversized/missing fields) now return **400 with zero delivery** instead of
  silently degrading to a plain-text DM. The avatar's absolute-https host is
  authoritatively checked in preflight with a real URL parser (a regex cannot
  assert host presence — `https://?x` has an empty host).
- **Docs**: `docs/platform-card-base.md` §15/§16 aligned to the delivered form.

## Impact surface (which cards change)

- **Smart-summary cards**: unaffected (not migrated; still the legacy resource-card
  path). The only shared code touched is `requireHTTPS`, now delegating to
  `AbsoluteHTTPSURL` with byte-identical validation logic.
- **Docs cards other than access-request** (shared / commented / granted / denied):
  unaffected (legacy v1 path).
- **Docs access-request, gate OFF**: unaffected (legacy v1 path).
- **Docs access-request, gate ON** (this pilot): valid inputs render the same card
  plus two additive, client-ignorable `metadata.octo.{protocol,template}` fields;
  invalid inputs now return 400 (C1) instead of a degraded text DM.

## Note for docs-backend

The request contract (`NotifyReq` / `DocsCardFields`) is unchanged — no version
bump needed for the normal flow. The one behavioral change to confirm is **C1**:
for the access-request card with the gate on, a malformed field (non-https avatar,
oversized text, missing required) now yields **HTTP 400 with zero delivery**
instead of HTTP 200 with a degraded text message. Please confirm (a) production
payloads stay within the schema bounds and (b) the retry/filtered state machine
tolerates a 400. The response `NotifyResp{Delivered, Filtered}` shape is unchanged,
and docs-backend does not consume the rendered card JSON, so the new metadata
fields are transparent to it.

## Testing

- `go build ./...`, `go vet ./pkg/cardtmpl/... ./modules/notify/` — clean
- `go test -race ./pkg/cardtmpl/... ./modules/notify/ ./internal/carddispatch/... ./internal/cardactiondispatch/...` — pass (MySQL/Redis/WuKongIM containers)
- `gofmt` clean; no new errcode, so i18n extract/lint have no delta
- New regression tests: caller-level unwired-registry → 500 zero-delivery;
  malformed-avatar → 400 zero-delivery; preflight avatar semantics; preflight
  emits `fields_invalid` metric; fallback-text newline sanitization; `Meta()`
  immutability; cancelled-context propagation
- Migration baseline test asserts the new path is byte-equal to the legacy card
  except the two additive `metadata.octo` fields and the `view_document` action id

## COMPREHENSION

   For docs `access_requested` with the approval gate on, card assembly moves from
   a direct builder call to `Registry.Render` — a single ingress that schema-validates
   inputs, routes state→view, builds the business fragment, enforces absolute-https
   on all URLs, injects `metadata.octo.{protocol,template}`, and runs
   `cardmsg.Validate`. Schema-level field errors become a 400 zero-delivery (C1)
   at ingress instead of a silent text degrade. Every other card path is untouched.

   docs-backend is the only external dependent. If it currently sends a malformed
   field that previously degraded to text (200), it now receives a 400 with nothing
   delivered — hence the heads-up above. Internally, a composition bug (registry not
   wired / bad schema) fails closed: the process panics at boot, and at runtime an
   unwired registry surfaces as a 500 (never a fake-success text). A shared
   `requireHTTPS` refactor could in principle affect summary/docs-v1 cards, but the
   validation logic is byte-identical and only the error message text changed.

   Full `-race` suites for `pkg/cardtmpl`, `modules/notify` (incl. integration),
   and both dispatch packages pass. The migration baseline test pins byte-equality
   with the legacy output modulo the three intended new fields. Dedicated tests
   cover each behavioral edge (C1 400 zero-delivery, unwired 500, avatar host
   bypass now rejected at preflight, fallback newline stripping, meta immutability,
   context cancellation). A register-time self-check + `main.go` fail-close assert
   the three-way `{sender_uid, owner, action_type}` callback consistency at boot.

## Checklist

- [x] Updated documentation (`docs/platform-card-base.md`)


an9xyz:
Superseded by #633 (head branch renamed to `cardtmpl-registry-pilot`). This PR was auto-closed by the rename; no content change.

lml2468 (approved):
# PR Review — PR #632 (Mininglamp-OSS/octo-server)

Reviewed at head `3663fac7` against merge-base `6f470355` (base `main`).
**Build/vet/test:** `go build` exit 0 ✅, `go vet` ✅. `go test ./pkg/cardtmpl/...` → **PASS** (base + pilot: conformance framework, registry-immutable, C1 regression, template tests — all pure, ran here). The `-short` FAIL is the DB-gated `modules/notify` integration (`testutil.NewTestServer` → MySQL refused) — infra, not a code failure.
Classification: new platform card-template base + first card migration — **card message escaping / trust-boundary / URL-destination** (the load-bearing rule) + wire-contract + a C1 behavior change. Reviewed in **SECURITY** emphasis. Note: this PR already folded in 3 prior review rounds (7 + 4/4 + 1/3 blockers per the commit log), so much is already hardened.

## 1. Spec compliance — ✅
Graded against `.octospec/tasks/cardtmpl-registry-pilot/brief.md` + `docs/platform-card-base.md`.
- **Missing:** none for the declared scope. L0 base (Template/Registry/Render/metadata-inject/metric), L1 handoff artifacts, the `docs.access-request@0.2.0` pilot (pending view only), notify wiring + Freeze, and the C1 400 change are all present.
- **Over-build:** none. The out-of-scope list (outcome views, `${}` engine, `CardUpdater`, public templates endpoint, L2b channel, `NotifyReq` shape change) is honored — verified `modules/notify/model.go` is **byte-untouched** (Scheme B) and `ext.*`/L2b is code-only, not wired to production.
- **Deviation:** the sole intended behavior change (C1: schema-level field error → 400 zero-delivery instead of degraded text DM) is exactly as documented and is called out for docs-backend. No silent divergence.
- **Load-bearing (escape/trust-boundary/url-destination rule):** honored — see Gate 2.

## 2. Code quality — Approved

### Findings
No P0/P1/P2 findings.

- **`AbsoluteHTTPSURL` defeats the empty-host bypass (single source of truth).** `render.go:188` uses `url.Parse` + `Scheme=="https"` + `Host!=""`, so `https://?x` / `https://#y` (empty host, which a regex `^https://…` would pass) are rejected. It's used in **both** the notify preflight (C1 ingress) and `renderCore` step-5 (DeepLink) and `requireHTTPS` (resource.go) — one implementation, zero seam between preflight-pass and Build-check.
- **Markdown escaping is correct per surface (the #605 lesson, applied right).** `docs_action.go` `escapeMarkdown`s every `TextBlock` (which renders markdown) and deliberately does **not** escape `TextRun` text (line 313-319 comment: TextRun renders literal, escaping would show literal backslashes) — and TextRun carries no link/URL surface + is `truncateRunes`-bounded, so raw text can't inject markup. The anonymous branch correctly uses an escaped TextBlock. This is exactly the TextRun/TextBlock distinction.
- **Final-boundary re-validation.** `renderCore` step-8 runs `cardmsg.Validate(payload)` after full assembly (block whitelist / caps / interaction gate) — so even a misbehaving Template can't emit a card that bypasses the wire-contract validator. Escaping/validation is enforced at the boundary the caller can't cross (trust-boundary rule).
- **C1 error classification is precise (3-way, no fail-open).** `card.go:378-390`: `ErrFieldsInvalid` → `errNotifyCardInvalid` → **400 zero delivery**; `errCardTmplUnavailable` (composition bug) → propagated → **500, no degrade** (a wiring hole stays visible, not masked as a text DM); other render errors → degrade to text so the notification still lands. Regression tests assert **zero delivery** on both the unwired-registry (500) and malformed-avatar (400) paths, plus a fields_invalid metric-emission test.
- **Registry immutability + owner allowlist.** `Register` after `Freeze()` → `ErrRegistryFrozen`; a non-L2a owner panics at register time (composition-time fail-fast); `SetDefault`/register are mutex-guarded; `Meta()` returns a deep `Clone()` so a frozen registry stays immutable to callers. Register-time sample self-check validates each template against its own schema.
- **Fail-close composition.** Main registers the pilot then `Freeze()`s; a missing template / bad schema fails startup (matches the existing docs-callback-route fail-close anchor).
- **Additive dependency:** `go.mod` adds exactly the stated `santhosh-tekuri/jsonschema/v6 v6.0.2`; `NotifyReq`/`DocsCardFields` unchanged, so docs-backend needs no version bump (only the C1 status-code behavior to confirm).

### Non-blocking notes
- **Documented hardcoded per-field URL check (maintenance coupling).** `preflightDocsAccessRequestSchema` (card_via_registry.go:84-88) asserts `AbsoluteHTTPSURL` on the avatar field **by hand**, and the author explicitly flags: *any new URL field added to the L1 schema must add a matching check here*, else that field regresses to the "schema-regex passes → Build fails → render_error degrade" seam that C1 closes for the avatar. Correct today (avatar is the only user-controlled URL; `Source.IconURL` is server-set), but it's a latent footgun. Recommend the suggested follow-up: let a Template *declare* its must-be-absolute-https field set so the base enforces them uniformly, removing the hand-maintained coupling. Non-blocking for this pilot.
- **Pilot Template.Build mutates `cardActions` in place** to add the `view_document` id (template.go:139-146). Fine because Build owns freshly-built actions, but if actions ever become shared/cached, this would be a mutation hazard — worth a note when card #2 lands.

## 3. Test assessment
Strong: a generalized `conformance_test.go` framework, `registry_immutable_test.go`, `card_c1_regression_test.go` (zero-delivery on both 400 and 500 paths + metric emission), `card_via_registry_baseline_test.go` (byte-equivalence vs legacy wrapper), and pilot template tests. The pure `pkg/cardtmpl/...` suites run green here. The `modules/notify` HTTP-integration tests are DB-gated (unrun in sandbox) → CI must confirm.

## 4. Coverage / blind spots
- Verified by build/vet + the pure cardtmpl/pilot suites + full read of render/registry/pilot/notify-wiring + escape-surface and URL-parse tracing.
- NOT run: DB-gated `modules/notify` integration; the live docs-backend C1 interaction (does its retry/filtered state machine tolerate a 400 — the PR asks docs-backend to confirm). Recommend CI DB run + a docs-backend contract check on the 400.

## Verdict
**APPROVE** — a well-architected, thoroughly-iterated (3 prior review rounds folded) platform-card base + pilot. The security-load-bearing pieces are right: `AbsoluteHTTPSURL` closes the empty-host bypass as a single source of truth, markdown escaping respects the TextBlock-vs-TextRun surface distinction, `cardmsg.Validate` re-runs at the uncrossable boundary, and the C1 error path is a precise 3-way (400 zero-delivery / 500 no-degrade / degrade) with zero-delivery regression tests. `NotifyReq` shape untouched (Scheme B), registry is freeze-immutable with an L2a allowlist, composition fails closed. Both gates pass.
Suggested decision: `Safe to merge` once CI runs the DB-gated `modules/notify` suite and docs-backend confirms the C1 400 is tolerated. Consider the follow-up to make must-be-https fields Template-declared (removes the hand-maintained preflight coupling). (PR `merge: BLOCKED` — branch-protection approvals.)

Jerry-Xin (changes requested):
The registry architecture is solid and well fail-closed, but the migration introduces a text-fallback content regression on the gate-on path that contradicts the PR's own "no information is lost" contract.

## 🔴 Blocking

- 🔴 The pilot `Template.FallbackText` drops content vs. the legacy fallback it replaces, on the gate-enabled access-request branch.

  `modules/notify/card.go:574` (`buildDocsFallbackText`) now delegates the `access_requested` + gate-on branch to `pkg/cardtmpl/docs_access_request/template.go` `FallbackText`. The legacy composition (still present just below, and previously the only path — see `main` baseline) emits four lines: attribution + `Title` + `Excerpt` (`docsSafeExcerpt` returns `card.Excerpt` for this kind) + `UpdatedAt`.

  The new `FallbackText` emits only a single sentence: `"<actor> requested access to \"<title>\", please review."` (actor + title). The document excerpt and the updated-at timestamp are silently omitted.

  This fires whenever the card path degrades to text: `cardmsg.Enabled()==false`, no card sender configured, or a non-schema render error. In those cases a valid access request now delivers a shorter DM than before. This directly contradicts:
  - the `"No information is lost — every field that would appear on the card is emitted as a text line"` comment kept verbatim at `modules/notify/card.go:565`;
  - the PR claim that C1 (400 zero-delivery on schema errors) is the only externally visible behavioral change.

  Fix: restore parity — have the pilot `FallbackText` emit the excerpt and timestamp too (ideally via a shared helper so the two paths cannot drift), and add parity tests covering Chinese/English, anonymous actor, non-empty excerpt, and timestamp. The current `template_test.go` only checks newline sanitization and cannot catch this drop.

## 💬 Non-blocking

- 🟡 `pkg/cardtmpl/render.go:95` wraps the build error as `fmt.Errorf("%w: build: %v", ErrRenderFailed, err)`. Because the cause is folded in with `%v` rather than `%w`, `errors.Is(err, context.Canceled)` / `context.DeadlineExceeded` on the returned error will not match — only `ErrRenderFailed` is traversable. `registry_immutable_test.go:92` asserts `ErrRenderFailed` but not cancellation identity. If callers need to distinguish cancellation/deadline downstream, wrap the cause with `%w` (or a joined error) and assert it in the test.

## ✅ Highlights

- Strong fail-closed registry wiring: unwired registry surfaces as 500 (never a fake-success text), and boot-time `Freeze()` + register-time sample self-check catch composition bugs early.
- C1 400 zero-delivery for schema-level field errors is a clean, observable ingress policy.
- `Meta()` defensive deep copy keeps a frozen registry genuinely immutable.
- Good URL hardening: real URL parser rejects empty-host `https://` values a regex would miss.
- Migration baseline byte-equality test pinning legacy output modulo the intended additive metadata fields is a solid guardrail.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #632 (octo-server)

## Summary

This PR establishes the L0 platform card base (`octo-card@1.0`) and migrates the first L2a card — `docs.access-request@0.2.0` — through it end-to-end. The architecture introduces a `Template` interface + `Registry` with register-time fail-close, a single-ingress `Registry.Render` pipeline (schema-validate → state→view → build → https-enforce → metadata-inject → envelope → cardmsg.Validate), and a C1 policy that promotes schema-level field errors from silent text degradation to 400 zero-delivery. The notify module's docs `access_requested` branch (gate-on) is rerouted through the Registry; all other card paths remain on legacy code untouched.

The design is solid: layered defenses (preflight schema + preflight URL + render-internal schema + build-internal URL), proper error classification (400 for caller faults, 500 for composition bugs, degrade-to-text for render failures), and strong test coverage (conformance sweep, migration byte-equivalence baseline, C1 regression tests). No P0/P1 issues found.

## Verification

Local checks at head `3663fac7`:
- `go build ./...` — clean
- `go vet ./pkg/cardtmpl/... ./modules/notify/` — clean
- Static analysis of all 33 changed files, tracing data flows from ingress (`deliverDocsCardNotification`) through preflight → `Registry.Render` → `Template.Build` → `cardmsg.Validate` → delivery loop.

## Findings

No P0/P1 issues. Two P2 items and two nits below.

### P2 — `context.Background()` defeats stated ctx-propagation contract (`modules/notify/card.go:375`)

The PR's R2-7 narrative claims `renderCore` now respects caller ctx by explicitly receiving and forwarding it to `Template.Build`. However, the production entry point at `:375` calls `buildDocsAccessRequestCardViaRegistry(context.Background(), ...)` instead of threading the request context. The legacy `buildDocsCard` at `:377` does the same — this is a pre-existing pattern, not new. But the explicit R2-7 contract claim makes the gap more visible now.

Diff-scope: **pre-existing, amplified** — the old code also used `context.Background()`; the PR adds the ctx parameter to `renderCore` and documents the intent but doesn't wire the actual request context at the notify boundary.

For the current pilot (pure CPU rendering, no external calls), this has zero runtime impact. When future templates fetch external data or observe deadlines, the `context.Background()` will silently discard cancel/deadline signals. Consider threading `ctx` from the HTTP handler through `deliverDocsCardNotification` to the build call.

### P2 — `permission.roleLabel` concatenated without sanitization (latent, path dead) (`pkg/cardtmpl/docs_access_request/labels.go:56`)

`pendingLabels` concatenates `pf.Permission.RoleLabel` into `bannerSuffix` (`"申请成为此文档的" + roleLabel + "。"`) without `sanitizeLine`. Currently `mapDocsCardFieldsToJSON` never emits the `permission` object, so `roleLabel` is always empty and defaults to the safe built-in (`"查看者"` / `"viewer"`). When a future PR wires `permission` to caller-controlled data, control characters or UI-confusing text in `roleLabel` would land in the card body unsanitized — unlike `FallbackText` which runs `sanitizeLine` on actor/title.

Diff-scope: **new code** — but the path is currently dead. Not a blocker; a `sanitizeLine(roleLabel)` guard here (or a documented obligation for the activating PR) would close it.

### Nit — Legacy `buildDocsAccessRequestCard` is now unreachable from production

The function remains in `modules/notify/card.go` but is no longer called from the gate-on access-requested path. It's still referenced by the baseline test (`card_via_registry_baseline_test.go`) for byte-equivalence comparison, which is valuable for migration verification. Consider adding a `// Deprecated: kept for migration baseline test only` marker to prevent accidental reuse.

### Nit — No concurrent-read stress test for frozen Registry

The Registry uses `sync.RWMutex` and is a process-global singleton hit by every card delivery goroutine. The `registry_immutable_test.go` tests Meta/List deep-copy guarantees but doesn't include a concurrent `sync.WaitGroup` + goroutine stress test. The `sync.RWMutex` usage appears correct from code inspection (RLock for all reads, Lock only for Register/SetDefault/Freeze), but a `-race` stress test would provide stronger assurance for this security-sensitive singleton.

- **C1 dual-path parity**: preflight (`preflightDocsAccessRequestSchema`) and render (`renderCore` step 2) both validate against the same `tmpl.Meta().InputSchema`. Preflight additionally runs `AbsoluteHTTPSURL` on `card.ActorAvatarURL`. Render's Build path runs `requireHTTPS` → `AbsoluteHTTPSURL` on the same avatar value. Same validator, same data — zero gap.
- **`requireHTTPS` delegation** (`pkg/cardtmpl/resource.go:292`): Old inline `url.Parse(TrimSpace)` + scheme/host check replaced by `AbsoluteHTTPSURL`. Behavioral parity confirmed for all non-empty inputs. All callers guard empty before calling.
- **Error classification chain**: preflight returns `errCardTmplUnavailable` (500) or `ErrFieldsInvalid` (400). Build-path handler at `card.go:382-394` discriminates the same types. Render errors (`ErrRenderFailed`) correctly degrade to text. No misclassification paths found.
- **Register-time self-check**: `selfCheckSamples` runs `renderCore` within the Register lock (avoiding deadlock), validates Action.Submit data contract against `Meta.ActionContract`, and panics on mismatch. Boot-time fail-close is properly enforced.
- **`Meta()` deep-copy immutability**: `Clone()` correctly deep-copies Views/stateIndex/interactions/Manifest/ActionContract. `InputSchema` (compiled, immutable) is shared safely. `registry_immutable_test.go` validates mutation-isolation.
- **Feature gate consistency**: preflight and build both gated on `card.Kind == DocsCardKindAccessRequested && docsApprovalCardsEnabled()`. No path where preflight runs but build doesn't, or vice versa.
- **A16 three-way action contract**: conformance test pins `docs` / `access_request.decision` independently against Meta.ActionContract and `cardactiondispatch.Resolve` — prevents the self-proving trap.
- **Migration byte-equivalence**: baseline test strips exactly the 3 allowed new fields (`octo.protocol`, `octo.template`, `view_document` action ID) and asserts canonical JSON byte-equality with legacy output.
- **AbsoluteHTTPSURL as single source of truth**: `render.go:189` (renderCore step 5), `resource.go:292` (requireHTTPS), and `card_via_registry.go:102` (preflight) all delegate to the same function. No regex-vs-parser drift.

## Data flow trace

| Consumer | Source | Verified |
|---|---|---|
| `renderCore` step 2 (schema validate) | `fields` from `mapDocsCardFieldsToJSON` → `json.Marshal` | ✅ Same JSON fed to preflight and render |
| `Template.Build` (unmarshal `pendingFields`) | Same `fields` after schema pass | ✅ Schema ensures shape |
| `BuildDocsAccessRequestBodyWithLang` (card body) | `pendingFields` struct fields | ✅ All fields trace to `DocsCardFields` via mapping |
| `AbsoluteHTTPSURL(br.DeepLink)` (step 5) | `docsDeepLink(WebLoginURL, requestID, spaceID)` | ✅ WebLoginURL from server config, not user input |
| `metadata.octo.source` | `sourceForLang(lang)` or `Meta.Source` | ✅ Server-controlled, no user injection |
| `cardmsg.Validate(payload)` (step 8) | Complete envelope from `renderCore` assembly | ✅ Catches structural issues post-build |

- **C1 (dual-path parity)**: ✅ `requireHTTPS` → `AbsoluteHTTPSURL` on both legacy and pilot paths. Preflight + render use same schema. Gate conditions identical.
- **C2 (control-flow ordering)**: ✅ `renderCore` steps are strictly sequential. No nested re-entry. Self-check avoids deadlock by calling `renderCore` directly within lock.
- **C3 (authorization boundary)**: ✅ No new external endpoints. Pilot only adds internal rendering path for existing notification type. Authorization enforced upstream by notify module.
- **C4 (authorization lifecycle)**: N/A — no auth changes.
- **C5 (build/note ≠ runtime)**: ✅ `go build ./...` and `go vet` both pass. Schema validation and `cardmsg.Validate` are runtime checks, not build-time.
- **C6 (governance/docs self-consistency)**: N/A — docs/platform-card-base.md is a new spec document, not a governance policy change.

N/A — first review pass.

## Verdict: COMMENT

No P0/P1 issues found. The architecture is well-designed with proper layered defenses, fail-close semantics, and correct error classification. Two P2 items (ctx-propagation gap amplified by the PR's own narrative, and a latent sanitization gap on a dead path) are non-blocking. The PR is safe to land with or without the P2 fixes — they address future-phase concerns rather than current-production defects.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. Two P2 items are non-blocking (pre-existing ctx pattern, dead-path sanitization). Safe to land.

yujiawei (approved):
# Code Review — PR #632 (octo-server)

Reviewed the full diff (33 files, +4457/-97) against merge-base `6f470355`. Scope: the L0 `pkg/cardtmpl` Registry + single `Render` ingress, the `docs.access-request@0.2.0` pilot, the C1 (400 zero-delivery) behavior change, and production wiring. Build / vet / gofmt clean; `go test ./pkg/cardtmpl/...` passes (DB-backed `modules/notify` suites compile, not run here).

This is a well-structured vertical slice. The ingress is a genuine single funnel, the C1 error classing is correct end-to-end, and the migration is guarded by a self-output byte-equivalence baseline. I could not refute the core safety claims for the pilot path. Findings below are all **P2 (non-blocking)** — most are forward-looking hardening for the L0 base that future cards will build on, plus one reachable cosmetic edge case.

## 1. Spec compliance

**Spec: ✅**

The linked issue ships a plan; checked against it:

- **Under-build:** none. L0 base (`template.go` / `registry.go` / `render.go` / `metrics.go` / `default_registry.go`), the pilot subpackage, C1, the composition-root fail-close wiring, conformance + baseline + C1 regression tests, and the contract doc are all present.
- **Over-build:** none reaching production. The explicitly deferred items (card updater, public `/templates` endpoint, unified callback envelope, other card migrations, the `${}` engine, `NotifyReq` shape change, L2b channel) are absent. The `ext.*` (L2b) branch in `registry.go` exists but *panics* on registration and is test-only — which the plan sanctions.
- **Deviation:** the sole intended external behavior change is C1 (schema-level field error → 400 zero-delivery for the gate-on access-request branch). Confirmed. `NotifyReq` / `DocsCardFields` / `NotifyResp` shapes are untouched, and the two additive `metadata.octo.{protocol,template}` fields are client-ignorable. (One unintended cosmetic divergence exists in an edge case — see Quality §2, finding F1 — but it does not change the wire contract, so it does not fail the spec gate.)

## 2. Code quality

**Quality: Approved**

No reachable P0/P1 in this pilot. The C1 path is sound: `preflightDocsAccessRequestSchema` runs before membership / sender / `cardmsg.Enabled` gates, so a schema or avatar-host violation deterministically returns the 400-class `errNotifyCardInvalid` with zero delivery (verified in `card.go:311-324`); a composition failure returns typed `errCardTmplUnavailable` → 500 (never a fake-success text); post-schema render failures still degrade to text. `AbsoluteHTTPSURL` correctly rejects empty-host forms (`https://?x`, `https:///path`) that the schema regex lets through — I reproduced this. `maxLength` in the compiled schema counts **runes**, not bytes (verified), so CJK content is not spuriously rejected. Mapped fields are `json.Marshal`-escaped. Cards embed URLs for the client to navigate; the server never fetches them, so private-host HTTPS is not an SSRF sink here.

### P2 — base-layer hardening (not reachable in this pilot; worth fixing before card #2/#3)

**F2. `ActionContract` is enforced only at registration, not at `Render` time.** `registry.go:175` self-checks samples via `assertActionContract`, but `renderCore` (`render.go:92-150`) goes straight from `Build` to `cardmsg.Validate`, which does not inspect `TemplateMeta.ActionContract`. A future template whose `data.owner`/`action_type` branch on runtime input could emit a mismatched route and still be delivered. Not reachable in the pilot because these are hardcoded literals (`docs_action.go:176-177`). Consider asserting the contract inside `renderCore` for interactive views.

**F3. Empty owner / incomplete action contract bypasses the allowlist and the only ownership check.** `registry.go:124` gates the allowlist on `mf.Owner != ""`, and `registry.go:145` only builds `ActionContract` when *both* owner and actionType are non-empty. A manifest with `owner:""` (or a missing actionType) registers with a nil contract, and if it still emits `Action.Submit` actions nothing verifies ownership. The self-check only fires "no Submit" when the contract is non-nil. Recommend: reject an interactive template (has Submit actions) that declares no `ActionContract`.

**F4. `metadata.octo.source.iconUrl` skips the absolute-HTTPS gate.** `render.go:98-102` runs `AbsoluteHTTPSURL` on `DeepLink` only; `src.IconURL` is copied verbatim (`render.go:124-126`), and `cardmsg.Validate` does not traverse `metadata`. A template returning `Source{IconURL: "javascript:..."}` would inject an unchecked URL that icon-consuming renderers trust. Not reachable in the pilot (`IconURL` is always empty — manifest has none, `sourceForLang` sets only `Label`, caller can't supply). Apply the same `AbsoluteHTTPSURL` gate to a non-empty `Source.IconURL`.

### P2 — reachable, cosmetic

**F1. Anonymous-actor banner regresses on the new path, and the baseline test does not cover it.** When an access-request has an empty `ActorName` (allowed — only DocID/Title/RequestID are required), the legacy builder uses a dedicated subject-less sentence (`requestBannerAnon`: "Someone is requesting viewer access to this document." / "有人申请成为此文档的查看者。"). The pilot's `pendingLabels` (`labels.go`) has no anonymous variant, so `docsBanner("", suffix)` renders the *named-form suffix alone* → EN: "requested viewer access to this document." (a lowercase, subject-less fragment that reads as broken); ZH drops the "有人" subject. I reproduced both. `TestBuildDocsAccessRequestCard_MigrationBaseline` only exercises the named fixture (`ActorName:"Alice"`), so this divergence slips past the byte-equivalence lock. The text fallback is unaffected (pilot `FallbackText` has a proper `fallbackAnon`). Suggest adding an anonymous banner variant to `pendingLabels` and an anon-actor case to the baseline test.

### P2 — test / API notes

**F5. `SetMeta` stays exported after `Freeze`, so "immutable" is slightly overstated.** `Lookup` returns the stored mutable `*Template` (`registry.go:208-223`) and `SetMeta` performs an unlocked assignment. `Meta()` correctly returns a deep copy, but in-process code could type-assert the pointer and mutate `t.meta` post-freeze, racing preflight. Exploitation requires deliberate in-process misuse (not an external vector); still, consider guarding `SetMeta` against post-freeze calls or unexporting it.

**F6. The baseline test masks a third (action-level) difference.** `stripOctoNewFields` deletes the `view_document` id in addition to the two metadata fields, so the test cannot by itself establish that migration differences are limited to `metadata.octo.{protocol,template}`. The divergence is intentional and documented, and the lock is still valuable — just weaker than its name implies.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. All findings are P2 (non-blocking). The C1 400-zero-delivery contract change is the one behavior a downstream (docs-backend) must confirm tolerating, which the PR description already flags.

## 4. Suggested follow-ups

- Add an anonymous-actor banner variant to the pilot and cover it in the baseline test (F1).
- Before migrating card #2/#3, lift the three base invariants to `Render` time: assert `ActionContract` at runtime (F2), reject interactive-without-contract templates (F3), and gate `Source.IconURL` through `AbsoluteHTTPSURL` (F4).

## 5. Items for manual verification (security-sensitive)

- **docs-backend tolerance of C1:** confirm production payloads stay within schema bounds and the retry/filtered state machine accepts an HTTP 400 (previously a 200 text degrade).
- **Fail-close at boot:** the composition root `Register`→`SetDefault`→`Freeze` panics on a bad manifest/schema with no runtime env escape hatch — rollback is image revert. Worth a human eyeing the deploy/rollback runbook.

PR #637, opened 2026-07-22
fix(space): createInvite accepts duration param, returns expire field

xiaokenbot:
## Summary

`createInvite` (`POST /v1/space/:space_id/invite`) previously ignored the request body and always applied the default TTL (72 h via `applyAutoInviteDefaults`). The frontend duration picker in octo-web PR #972 therefore had no effect.

## Changes

**`modules/space/api.go`**

- Add `createInviteReq` struct with optional `Duration *int64` field (seconds; `0` = permanent).
- `createInvite` handler now:
  - Parses request body (best-effort; missing/malformed body falls back to default TTL — no breaking change for existing callers).
  - `duration = 0` → `ExpiresAt = nil` (permanent link).
  - `duration > 0` → `ExpiresAt = now + duration`.
  - `duration < 0` → 400 Bad Request (`duration` field invalid).
  - `duration` omitted → calls `applyAutoInviteDefaults` (env `DM_SPACE_INVITE_DEFAULT_TTL` / 72 h), preserving existing behaviour.
- Response now includes `"expire"` field: RFC 3339 string, or `""` for permanent links.

## Pairing

Frontend: Mininglamp-OSS/octo-web#972 (`fix(space): pass duration to invite API; show expiry; add duration picker`)

## Backward Compatibility

- Callers that send an empty body continue to receive the default TTL behaviour unchanged.
- New `"expire"` field is additive; existing consumers can ignore it.

### DIFF
--- a/modules/space/api.go
+++ b/modules/space/api.go
@@ -20,6 +20,7 @@ import (
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
 	"github.com/Mininglamp-OSS/octo-server/modules/base/event"
 	commonmod "github.com/Mininglamp-OSS/octo-server/modules/common"
+	"github.com/Mininglamp-OSS/octo-server/pkg/db"
 	"github.com/Mininglamp-OSS/octo-server/pkg/errcode"
 	"github.com/Mininglamp-OSS/octo-server/pkg/httperr"
 	octoredis "github.com/Mininglamp-OSS/octo-server/pkg/redis"
--- a/modules/space/api.go
+++ b/modules/space/api.go
@@ -901,6 +901,14 @@ func (s *Space) updateMemberRole(c *wkhttp.Context) {
 }
 
 // createInvite 创建邀请链接
+// createInviteReq 创建邀请码请求体。
+//
+// Duration 可选，单位秒；0 或缺省表示「永久」（不过期）。
+// 正值按秒计算过期时刻；超长值（> 10 年）在写入层截断，不在此处校验。
+type createInviteReq struct {
+	Duration *int64 `json:"duration"`
+}
+
 func (s *Space) createInvite(c *wkhttp.Context) {
 	loginUID := c.GetLoginUID()
 	spaceId := c.Param("space_id")
@@ -919,12 +927,33 @@ func (s *Space) createInvite(c *wkhttp.Context) {
 		return
 	}
 
+	// 解析可选 duration 字段；body 缺失或解析失败时退化为默认行为。
+	var req createInviteReq
+	_ = c.BindJSON(&req)
+
+	now := time.Now()
 	inviteModel := &InvitationModel{
 		SpaceId: spaceId,
 		Creator: loginUID,
 		Status:  1,
 	}
-	applyAutoInviteDefaults(inviteModel, time.Now())
+
+	if req.Duration != nil {
+		if *req.Duration == 0 {
+			// 0 = 永久，ExpiresAt 保持 nil
+		} else if *req.Duration > 0 {
+			t := db.Time(now.Add(time.Duration(*req.Duration) * time.Second))
+			inviteModel.ExpiresAt = &t
+		} else {
+			// 负值：拒绝，duration 无意义
+			respondSpaceRequestInvalid(c, "duration")
+			return
+		}
+	} else {
+		// 未传 duration：使用环境变量 / 默认 TTL（72h）
+		applyAutoInviteDefaults(inviteModel, now)
+	}
+
 	inviteCode, err := s.insertInvitationWithRetry(inviteModel)
 	if err != nil {
 		s.Error("创建邀请链接失败", zap.Error(err), zap.String("spaceId", spaceId))
@@ -932,8 +961,15 @@ func (s *Space) createInvite(c *wkhttp.Context) {
 		return
 	}
 
+	// 构造 expire 字段：永久返回空字符串，有过期时间返回 RFC3339。
+	expireStr := ""
+	if inviteModel.ExpiresAt != nil {
+		expireStr = time.Time(*inviteModel.ExpiresAt).UTC().Format(time.RFC3339)
+	}
+
 	c.Response(map[string]string{
 		"invite_code": inviteCode,
+		"expire":      expireStr,
 	})
 }
 
--- a/modules/space/api.go
+++ b/modules/space/api.go
@@ -911,14 +911,15 @@ type createInviteReq struct {
 	Duration *int64 `json:"duration"`
 }
 
+// mysqlTimestampMax 是 MySQL TIMESTAMP 列的最大可表示时刻（2038-01-19 03:14:07 UTC）。
+// space_invitation.expires_at 使用 TIMESTAMP 类型，超出此时刻的值
+// 会因 sql_mode 不同而报错或被截断为错误值。
+// 所有写入 expires_at 的路径均应确保 time.Time <= mysqlTimestampMax。
+var mysqlTimestampMax = time.Date(2038, 1, 19, 3, 14, 7, 0, time.UTC)
+
 // maxInviteDurationSeconds 邀请码最大有效期（秒）。
-//
-// 约 10 年（3650 天），满足以下约束：
-//   - 远低于 time.Duration（int64 纳秒）溢出点（~292 年）。
-//   - 加上当前时间后不超过 MySQL TIMESTAMP 上限（2038-01-19），
-//     只要在 2028 年以前调用本接口即成立（最坏情况：2028 + 10 = 2038）。
-//
-// 如需修改，请同时评估数据库列类型（当前 TIMESTAMP，上限 2038-01-19）。
+// 约 10 年（3650 天），远低于 time.Duration（int64 纳秒）溢出点（~292 年）。
+// 仅防溢出；MySQL TIMESTAMP 上限由 createInvite 中对 mysqlTimestampMax 的绝对校验负责。
 const maxInviteDurationSeconds = int64(3650 * 24 * 3600) // ~10 年
 
 func (s *Space) createInvite(c *wkhttp.Context) {
@@ -969,8 +970,13 @@ func (s *Space) createInvite(c *wkhttp.Context) {
 			// 永久：清除默认 ExpiresAt
 			inviteModel.ExpiresAt = nil
 		default:
-			// 正整数：覆盖默认 ExpiresAt
-			t := db.Time(now.Add(time.Duration(*req.Duration) * time.Second))
+			// 正整数：校验绝对上限，防止超出 MySQL TIMESTAMP 范围（2038-01-19）。
+			candidate := now.Add(time.Duration(*req.Duration) * time.Second)
+			if candidate.After(mysqlTimestampMax) {
+				respondSpaceRequestInvalid(c, "duration")
+				return
+			}
+			t := db.Time(candidate)
 			inviteModel.ExpiresAt = &t
 		}
 	}
--- a/modules/space/api.go
+++ b/modules/space/api.go
@@ -904,11 +904,22 @@ func (s *Space) updateMemberRole(c *wkhttp.Context) {
 // createInvite 创建邀请链接
 // createInviteReq 创建邀请码请求体。
 //
-// Duration 可选，单位秒；0 或缺省表示「永久」（不过期）。
-// 正值按秒计算过期时刻；超长值（> 10 年）在写入层截断，不在此处校验。
+// Duration 可选，单位秒；0 表示「永久」（不过期）；正整数按秒计算过期时刻。
+// 上限为 maxInviteDurationSeconds（约 10 年），超出或负值返回 400。
+// 字段缺省时沿用运营配置默认 TTL（DM_SPACE_INVITE_DEFAULT_TTL，默认 72h）。
 type createInviteReq struct {
 	Duration *int64 `json:"duration"`
 }
+
+// maxInviteDurationSeconds 邀请码最大有效期（秒）。
+//
+// 约 10 年（3650 天），满足以下约束：
+//   - 远低于 time.Duration（int64 纳秒）溢出点（~292 年）。
+//   - 加上当前时间后不超过 MySQL TIMESTAMP 上限（2038-01-19），
+//     只要在 2028 年以前调用本接口即成立（最坏情况：2028 + 10 = 2038）。
+//
+// 如需修改，请同时评估数据库列类型（当前 TIMESTAMP，上限 2038-01-19）。
+const maxInviteDurationSeconds = int64(3650 * 24 * 3600) // ~10 年
 
 func (s *Space) createInvite(c *wkhttp.Context) {
 	loginUID := c.GetLoginUID()
@@ -928,9 +939,10 @@ func (s *Space) createInvite(c *wkhttp.Context) {
 		return
 	}
 
-	// 解析可选 duration 字段；body 缺失或解析失败时退化为默认行为。
+	// ShouldBindJSON：body 缺失/EOF 时不写 4xx，仅返回 error；
+	// 与 BindJSON(MustBindWith) 不同，不会在 empty body 时 abort 请求。
 	var req createInviteReq
-	_ = c.BindJSON(&req)
+	_ = c.ShouldBindJSON(&req)
 
 	now := time.Now()
 	inviteModel := &InvitationModel{
@@ -939,21 +951,30 @@ func (s *Space) createInvite(c *wkhttp.Context) {
 		Status:  1,
 	}
 
+	// 始终先应用运营默认值（MaxUses + 默认 ExpiresAt），
+	// 再按 duration 字段覆盖 ExpiresAt，确保 MaxUses 策略始终生效。
+	applyAutoInviteDefaults(inviteModel, now)
+
 	if req.Duration != nil {
-		if *req.Duration == 0 {
-			// 0 = 永久，ExpiresAt 保持 nil
-		} else if *req.Duration > 0 {
-			t := db.Time(now.Add(time.Duration(*req.Duration) * time.Second))
-			inviteModel.ExpiresAt = &t
-		} else {
-			// 负值：拒绝，duration 无意义
+		switch {
+		case *req.Duration < 0:
+			// 负值无意义，拒绝
 			respondSpaceRequestInvalid(c, "duration")
 			return
+		case *req.Duration > maxInviteDurationSeconds:
+			// 超出上限，防止 time.Duration 溢出及 MySQL TIMESTAMP 越界
+			respondSpaceRequestInvalid(c, "duration")
+			return
+		case *req.Duration == 0:
+			// 永久：清除默认 ExpiresAt
+			inviteModel.ExpiresAt = nil
+		default:
+			// 正整数：覆盖默认 ExpiresAt
+			t := db.Time(now.Add(time.Duration(*req.Duration) * time.Second))
+			inviteModel.ExpiresAt = &t
 		}
-	} else {
-		// 未传 duration：使用环境变量 / 默认 TTL（72h）
-		applyAutoInviteDefaults(inviteModel, now)
 	}
+	// duration 缺省：保留 applyAutoInviteDefaults 写入的默认 ExpiresAt
 
 	inviteCode, err := s.insertInvitationWithRetry(inviteModel)
 	if err != nil {


lml2468 (commented):
# PR Review — PR #637 (Mininglamp-OSS/octo-server)

Reviewed at head `f4addabc` against merge-base `ca72704c` (base `main`).
**Build/vet/test:** `go build ./modules/space/` exit 0 ✅, `go vet ./modules/space/` exit 0 ✅. `go test` is DB-gated (`modules/space` → MySQL refused in sandbox) — unrun; CI to confirm. Overflow finding below reproduced with a standalone Go program.
Classification: space invite-link creation (auth + expiry arithmetic + wire-contract). Reviewed in correctness emphasis.

## 1. Spec compliance — ✅ (one false comment, see §2)
Matches the PR intent (honor the octo-web #972 duration picker): optional `Duration *int64`, `0`=permanent, `>0`=TTL, `<0`→400, omitted→existing default, additive `expire` response field. Backward-compatible (empty body → unchanged default). The one deviation is a **comment that claims a write-layer truncation that doesn't exist** (see the P2).

## 2. Code quality — Comment (one P2; not a hard blocker)

### P2 (MEDIUM, HIGH-confidence) — `time.Duration` overflow can create an already-expired invite
`inviteModel.ExpiresAt = now.Add(time.Duration(*req.Duration) * time.Second)` (api.go:947). `time.Duration` is `int64` **nanoseconds**, so `time.Duration(seconds) * time.Second` overflows once `seconds > math.MaxInt64/1e9 ≈ 9.22e9` (~292 years). Past that the product wraps, and for a range of large positive inputs the result is a time **in the past** — so the invite is written **already expired**, silently, and the response's `expire` field shows a past RFC3339 date.

**Reproduced** (standalone Go, fixed `now`):
- `duration = 9223372037` (sec) → `ExpiresAt = 1733-10-24` (past) — invite dead on arrival.
- `duration = 10000000000` → `ExpiresAt = 1758-06-03` (past).
- `duration = 9223372036854775807` → `ExpiresAt = now − 1s` (past).

The `< 0` guard does **not** catch this: the raw `duration` is positive; the sign flip happens inside the multiply. The PR's own comment says *"超长值(>10年)在写入层截断"* — I checked `insertInvitationWithRetry` → `insertInvitation`: **there is no clamp**; `ExpiresAt *db.Time` is written as-is. So the stated safety net does not exist, and the comment is misleading.

**Why P2, not a blocker:** the endpoint is authenticated and admin-gated (`Role ≥ 1`), the trigger needs an absurd `duration` (~1e10+ seconds) that the paired frontend picker will never send, and there's no isolation/data-corruption impact — worst case is a self-inflicted dead invite by an admin sending pathological input. But it's introduced by this PR, has a concrete trigger and a wrong (silently-past) result, and the fix is trivial.

**Fix:** bound `*req.Duration` to a sane maximum before the multiply (e.g. reject or clamp `> maxInviteDurationSeconds`, matching the "10 years" the comment already assumes), and either make the write layer actually clamp or correct the comment. A one-line `if *req.Duration > maxInviteSeconds { respondSpaceRequestInvalid(c, "duration"); return }` closes it.

### Verified clean
- **Auth/isolation unchanged and correct:** `checkSpaceActive` + `queryMember` + `member == nil || Role < 1 → ErrSpacePermissionDenied` gate the handler before any new logic — a non-admin/non-member can't create invites (space-isolation preserved).
- **Best-effort body parse is back-compat:** `_ = c.BindJSON(&req)` with `Duration *int64` — a missing/empty/malformed body leaves `Duration == nil` → `applyAutoInviteDefaults` (unchanged 72h/env default). Existing callers unaffected.
- **`0`=permanent** leaves `ExpiresAt = nil` correctly; **normal positive** durations (picker range) compute correctly; **negative** → 400 `respondSpaceRequestInvalid("duration")`.
- **`expire` field:** additive, RFC3339 UTC for a TTL invite, `""` for permanent — additive/omittable, no break for consumers reading only `invite_code`.
- **`db` import** added for `db.Time` (the second commit) — compiles (build green).

## 3. Test assessment
No test accompanies the change (grep: no `createInvite`/`duration` test in `modules/space/*_test.go`), and the package tests are DB-gated so unrun here. At minimum a table test for the four branches (nil→default, 0→permanent, positive→TTL, negative→400) — **and the large-duration overflow case** — would lock the contract and pin the P2 fix. Recommend adding it.

## 4. Coverage / blind spots
- Verified by build/vet + full read + standalone overflow repro + tracing the write path for a clamp (none found).
- NOT run: DB-gated space tests; the live octo-web #972 round trip.

## Verdict
**COMMENT** — the core change is small, correct for normal use, auth-gated, and backward-compatible; the duration picker will now work. One real but low-likelihood defect: a pathological `duration` (> ~9.2e9 s) overflows `time.Duration` and can silently create an **already-expired** invite, and the code's "written-layer truncation" comment describes a clamp that doesn't exist. Not blocking (admin-gated, absurd input, trivial fix, no isolation impact), but should be fixed before relying on large durations — bound `duration` to a max and correct/implement the truncation claim, plus add a branch test including the overflow case.
Suggested decision: `Approve after the duration bound + comment fix` (or land as-is and fast-follow, given the low likelihood). CI must run the DB-gated space suite.

Jerry-Xin (dismissed):
Focused fix that wires the request-body `duration` into invite expiry. Semantics (0 = permanent, >0 = now+duration, <0 = 400), back-compat, and authz all look correct — but the new positive-duration path has no upper bound and can overflow, so requesting changes.

## 🔴 Blocking

**1. Unbounded `duration` overflows `time.Duration` — `modules/space/api.go` (positive-duration branch)**

```go
t := db.Time(now.Add(time.Duration(*req.Duration) * time.Second))
```

`*req.Duration` is caller-supplied `int64` seconds. `time.Duration(x) * time.Second` multiplies by 1e9 (nanoseconds). Any `duration` greater than ~9.22e9 seconds (~292 years) overflows the int64 nanosecond range and wraps to a garbage/negative `time.Duration`, so `now.Add(...)` produces a timestamp **in the past or otherwise wrong** — the API silently returns an already-expired or nonsensical invite instead of rejecting the input.

The `createInviteReq` doc comment claims "超长值（> 10 年）在写入层截断，不在此处校验", but `DB.insertInvitation` (`modules/space/db.go`) performs no truncation — it writes `time.Time(*m.ExpiresAt)` directly. There is no upper-bound guard anywhere.

Fix: validate an explicit maximum before computing `ExpiresAt`, e.g. reject (400) or clamp when `*req.Duration` exceeds a sane cap (a named `maxInviteDurationSeconds` const), and drop/repair the inaccurate "truncated at write layer" comment.

**2. Expiry can exceed MySQL `TIMESTAMP` range (year 2038) — `modules/space/sql/20260307000002_space_legacy01.sql`**

`space_invitation.expires_at` is declared `TIMESTAMP`, which tops out at 2038-01-19 03:14:07 UTC. A valid, non-overflowing large duration (e.g. 20–30 years) still yields an `ExpiresAt` beyond that range; depending on `sql_mode` the insert either errors or truncates to a zero/clamped date, producing a wrong expiry. The same max-duration cap requested in item 1 should be chosen to stay within the column's representable range (or the column widened to `DATETIME`).

## 💬 Non-blocking

- No tests accompany the change. Please add table-driven coverage for the four branches: omitted (default 72h TTL), explicit `>0` (expiry = now+duration, correct unit), `0` (permanent, `expires_at` NULL and `expire` == ""), and `<0` (400). A boundary/overflow case guards item 1 against regression.
- `_ = c.BindJSON(&req)` deliberately swallows parse errors to preserve back-compat — worth a one-line comment noting that a malformed body intentionally falls through to default TTL rather than 400, so future readers don't "fix" it into a hard error and break existing callers.

## ✅ Highlights

- `Duration *int64` pointer cleanly distinguishes omitted vs. explicit `0`, keeping the default-TTL path fully back-compatible for existing callers.
- New `expire` response field is additive (empty string for permanent), so existing consumers are unaffected.
- Permission check (`member.Role < 1` → denied) runs before body parsing; the change does not widen who can create invites.
- `0 = permanent` maps correctly to `ExpiresAt = nil`, matching the existing NULL-means-no-expiry query semantics.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #637 (octo-server)

## Summary

This PR adds an optional `duration` parameter to `POST /v1/space/:space_id/invite`, allowing callers to set invite expiry (seconds; 0 = permanent). The response gains an additive `expire` field (RFC 3339 or empty string). The existing default-TTL path is preserved when `duration` is omitted. Single-file change, focused scope.

Review surface: one file (`modules/space/api.go`), +38/−1. Three non-blocking findings below — all P2.

## Verification

- ✅ **Diff scope** — single file, two commits (feature + import fix). Only `createInvite` handler and one new type are touched.
- ✅ **Auth gate preserved** — `checkSpaceActive` + `member.Role < 1` checks remain at `modules/space/api.go:917–928`, unchanged from base.
- ✅ **Negative duration rejected** — `*req.Duration < 0` returns 400 via `respondSpaceRequestInvalid(c, "duration")` at `:952`.
- ✅ **DB write path sound** — `insertInvitation` at `modules/space/db.go:266` converts `*db.Time` to `time.Time` for SQL and accepts nil `ExpiresAt` (NULL column), matching the permanent-link case.
- ✅ **Response format additive** — new `"expire"` key is added alongside existing `"invite_code"`; existing consumers can ignore it.
- Static analysis only at head `f4addabc`; build and tests not executed in this environment.

## Findings

No P0/P1 issues; three P2 items below.

### P2 — `BindJSON` flushes HTTP 400 header on malformed body (`modules/space/api.go:933`)

`_ = c.BindJSON(&req)` delegates to gin's `MustBindWith`, which calls `AbortWithStatus(400)` → `WriteHeaderNow()` on parse failure (verified in gin v1.9.1 source). The HTTP 400 header is flushed to the client before the handler regains control. The handler then continues, creates an invite with default TTL (because `req.Duration` stays nil), and calls `c.Response(...)` which internally does `c.JSON(200, ...)`. By that point the status line is already on the wire — the client receives HTTP 400 with a success-like JSON body containing `invite_code` and `expire`. This is the only callsite in the codebase that discards `BindJSON`'s error; every other handler checks it.

Fix: replace `c.BindJSON` with `c.ShouldBindJSON`, which returns the error without writing headers or calling `Abort()` — this matches the stated intent ("body 缺失或解析失败时退化为默认行为").

```go
var req createInviteReq
_ = c.ShouldBindJSON(&req) // no side effects on parse failure
```

### P2 — `MaxUses` default silently bypassed when `duration` is provided (`modules/space/api.go:941`)

When `req.Duration != nil` (any of the three branches: zero, positive, or negative-rejected), the handler skips `applyAutoInviteDefaults`. That function at `modules/space/invite_code.go:75` also applies the `MaxUses` default from env var `DM_SPACE_INVITE_DEFAULT_MAX_USES`. Every invite created with an explicit `duration` therefore gets `MaxUses=0` (unlimited in the DB), regardless of the operator-configured default. This is a behavioural regression compared to the base branch, where every invite went through `applyAutoInviteDefaults`.

Fix: apply MaxUses default independently of the duration branch — either extract MaxUses logic into a separate call, or call `applyAutoInviteDefaults` first, then overwrite `ExpiresAt` based on `duration`.

```go
applyAutoInviteDefaults(inviteModel, now) // sets MaxUses + default ExpiresAt
if req.Duration != nil {
    // overwrite ExpiresAt only
    ...
}
```

### P2 — Doc comment claims write-layer duration cap that does not exist (`modules/space/api.go:908`)

The doc comment states "超长值（> 10 年）在写入层截断，不在此处校验". `insertInvitation` at `modules/space/db.go:266` performs no truncation or cap. Values exceeding ~9.2×10⁹ seconds (~292 years) overflow `time.Duration` in the `time.Duration(*req.Duration) * time.Second` expression at `:948`, wrapping to a negative duration. `now.Add(negative)` yields a time in the past — the invite is immediately expired. While unlikely in production, the comment creates false confidence that this edge is handled when it is not.

Fix: either add an explicit cap in the handler (e.g. reject values > `315360000` — 10 years in seconds) or remove the claim from the doc comment.

- **Import addition** — `pkg/db` import at `:23` is needed for `db.Time()` conversion at `:948`; compiles correctly.
- **Permanent link path** — `duration == 0` leaves `ExpiresAt` nil → DB stores NULL → `queryInvitationByCode` at `modules/space/db.go:281` treats `expires_at IS NULL` as valid. Correct.
- **Positive duration path** — `time.Duration(*req.Duration) * time.Second` correctly converts seconds to nanoseconds for `now.Add()`. Valid for values up to ~292 years.
- **RFC 3339 formatting** — `time.Time(*inviteModel.ExpiresAt).UTC().Format(time.RFC3339)` at `:967` is standard and correct.
- **Negative duration rejection** — returns 400 with `"duration"` field hint before any DB write.
- **Go doc comment placement** — `// createInvite 创建邀请链接` at `:905` is now separated from `func createInvite` by the `createInviteReq` type declaration at `:906–912`, breaking Go's doc-comment association. Minor style nit; not blocking.

## Verdict: APPROVED

Three P2 findings, no P0/P1 blockers. The `BindJSON` → `ShouldBindJSON` swap is the most actionable fix — it's a one-word change that prevents the malformed-body status-code mismatch. The MaxUses bypass and the misleading doc comment are worth addressing but are non-blocking.

yujiawei (changes requested):
# Code Review — PR #637 (octo-server)

**Head SHA:** `f4addabcff558a0e75082061ceb748cdd9af6fe2`
**Scope:** single file, `modules/space/api.go` (+38 / −1)

## 1. Spec compliance

**Spec: ✅** (the two requested capabilities are both present, nothing out-of-scope was added)

- Requested `duration` param — implemented via `createInviteReq{ Duration *int64 }` and the `nil / 0 / >0 / <0` branch tree.
- Requested `expire` response field — implemented (RFC 3339, `""` for permanent).
- `duration` omitted → default-TTL path (`applyAutoInviteDefaults`) is present in the `else` branch.
- No unrequested flags/fields/endpoints introduced.

> Note: the PR *description* promises "missing/malformed body falls back to default TTL — no breaking change for existing callers." The code intends this, but does not achieve it — see the P1 below. The intent is correct; the mechanism is not.

## 2. Code quality

### 🔴 P1 — `BindJSON` on an empty/malformed body returns HTTP 400 to the caller (breaks the stated backward-compat guarantee)

`modules/space/api.go:933`

```go
var req createInviteReq
_ = c.BindJSON(&req)
```

`wkhttp.Context` embeds `*gin.Context`, so `c.BindJSON` is gin's `BindJSON` → `MustBindWith`, which **aborts the request with HTTP 400 and flushes the response header** on any decode error:

```
MustBindWith → ShouldBindWith(err) → AbortWithError(400) → AbortWithStatus(400) → c.Status(400) + Writer.WriteHeaderNow()
```

For a request with **no body** (`json.Decoder.Decode` returns `io.EOF`) or a malformed body, the 400 header is written *before* the handler continues. Discarding the returned error with `_ =` does **not** undo the already-written 400 status.

I verified this empirically against the pinned `gin v1.9.1` (isolated reproduction of exactly `_ = c.BindJSON(&req)` followed by a 200 response):

| Request body | Resulting HTTP status |
|---|---|
| *(empty body)* | **400** |
| *(empty body)* + `Content-Type: application/json` | **400** |
| `{}` | 200 |
| `{"duration":3600}` | 200 |
| `{bad` (malformed) | **400** |

Consequences:
1. Any existing caller that POSTs to this endpoint with an **empty body** (the pre-#972 behavior, and the exact "no breaking change" case the description claims) now receives **HTTP 400**.
2. The documented "`duration` omitted → default TTL" path is unreachable when the body is empty — it 400s before the `else` branch runs.
3. **Inconsistent state / duplicate risk:** the 400 is only the header; the invite row is still created in the DB by `insertInvitationWithRetry`. A client that treats 400 as failure and retries will mint a second invite code for the same intent.

**This regression is already solved elsewhere in the same file.** The manager handler guards the bind on content length:

`modules/space/api_manager.go:892`
```go
var req managerCreateInviteReq
if c.Request.ContentLength > 0 {
    if err := c.BindJSON(&req); err != nil {
        respondSpaceRequestInvalid(c, "")
        return
    }
}
```

**Suggested fix** — either mirror the sibling handler's `ContentLength > 0` guard, or switch to the non-aborting `ShouldBindJSON` (verified to return 200 for both empty and valid bodies while still populating `req.Duration`):

```go
var req createInviteReq
_ = c.ShouldBindJSON(&req) // ShouldBind* does not abort/flush 400 on empty/malformed body
```

Malformed (non-empty) JSON deserves an explicit decision: silently swallowing it means a typo'd `{"duraiton": 3600}` silently applies the default TTL instead of the user's intended value. Consider `ContentLength > 0 { if err := ShouldBindJSON; err != nil { respondSpaceRequestInvalid(c, "duration"); return } }` so a truly malformed body is a clean 400 while an empty body falls through to defaults.

### 🟡 P2 — `duration` upper bound: `time.Duration` multiplication can overflow into a past timestamp

`modules/space/api.go:946`
```go
t := db.Time(now.Add(time.Duration(*req.Duration) * time.Second))
```

`time.Duration` is `int64` nanoseconds. `*req.Duration * time.Second` overflows for `duration > ~9.22e9` seconds (~292 years). Verified: `duration = 9223372037` produces `ExpiresAt = 1734-...` (in the past) → the invite is created already-expired and instantly invalid, with no error surfaced to the caller.

The doc comment claims *"超长值（> 10 年）在写入层截断，不在此处校验"*, but I don't see a clamp in `insertInvitation` (`modules/space/db.go:266`) — it writes the value through, and the MySQL `expires_at TIMESTAMP` column caps at 2038 regardless. Recommend validating an explicit upper bound (e.g. reject or clamp anything beyond the DB `TIMESTAMP` ceiling / a sane max TTL) rather than relying on a truncation that isn't there.

### 🟡 P2 — response field naming/format inconsistency

The new field is `expire` in RFC 3339, while every other invite/expiry field in this module serializes as `expires_at` in `2006-01-02 15:04:05` form (`modules/space/model.go:204,226,238,249`; `db.Time.String()`). Additive, so not breaking, but clients now see two spellings/formats for the same concept. Consider `expires_at` for consistency, or confirm the divergence is deliberate for octo-web #972.

### 🟡 P2 — no tests

None of the new branches (`0` / `>0` / `<0` / omitted / empty-body) are covered. A single table test would have caught the P1 empty-body 400 before review.

## 3. Overall verdict

**REQUEST_CHANGES** — the P1 is a real correctness regression on the exact backward-compatibility guarantee the PR makes, it is trivially triggered (empty-body POST), it can leave DB/response state inconsistent, and the correct pattern already exists one function away in the same file. The one-line fix (`ShouldBindJSON`, or a `ContentLength` guard) unblocks it.

## 4. Suggested direction

1. Replace `_ = c.BindJSON(&req)` with a non-aborting bind (`ShouldBindJSON`) or a `ContentLength > 0` guard mirroring `api_manager.go:892`; decide explicitly how malformed-but-present bodies should behave.
2. Add an upper bound on `duration` (or clamp to the DB `TIMESTAMP` ceiling) so large values can't produce an already-expired invite.
3. Add a table test covering empty body, `{}`, `duration:0`, `duration>0`, `duration<0`, and an over-large value.
4. Consider aligning the response field name/format with the module's existing `expires_at`.

## 5. Additional notes

- Review coverage: this diff is a single self-contained handler; I read the handler, its helpers (`applyAutoInviteDefaults`, `inviteDefaults`, `insertInvitationWithRetry`, `insertInvitation`), the sibling manager handler, the `db.Time`/`wkhttp.Context`/gin binding chain, and the `space_invitation` schema. Cross-file dependency verified: no other caller of `createInvite` exists beyond the route registration at `api.go:87`. The P1 was reproduced against the pinned `gin v1.9.1`, not inferred.
- The empty-body 400 also means the pre-existing frontend (before octo-web #972) may already be affected if it did not send a JSON body; worth confirming what the current client sends.

yujiawei (commented):
## Follow-up — one additional blocking finding (verdict unchanged: CHANGES_REQUESTED)

A first-pass automated check flagged a fourth issue that my initial review missed. I verified it against the base branch and it is real and merge-blocking, so I'm adding it here for completeness. The verdict remains **CHANGES_REQUESTED**.

### 🔴 P1 — `MaxUses` operator default is silently bypassed whenever `duration` is provided

`modules/space/api.go:942`

On the base branch, **every** user-side invite went through `applyAutoInviteDefaults`, which sets both `ExpiresAt` *and* `MaxUses` from the operator-configured env default (`DM_SPACE_INVITE_DEFAULT_MAX_USES`, see `modules/space/invite_code.go:74-83`).

In this PR, `applyAutoInviteDefaults` only runs in the `else` branch (`duration` omitted). Any request that carries a `duration` — including `duration: 0` and positive values — skips it entirely, so `MaxUses` stays at its zero value, which the DB treats as **unlimited uses**:

```go
if req.Duration != nil {
    // sets ExpiresAt only — MaxUses default never applied
    ...
} else {
    applyAutoInviteDefaults(inviteModel, now) // sets BOTH MaxUses and ExpiresAt
}
```

Effect: as soon as octo-web #972 starts sending `duration`, every invite created through the picker silently becomes unlimited-use, ignoring the operator's `DM_SPACE_INVITE_DEFAULT_MAX_USES` policy. This is a behavioral regression versus base, and it's a quiet one (no error, wrong limit).

**Suggested fix** — apply the defaults first, then override only `ExpiresAt` from `duration` (this also naturally fixes the permanent-link case: set `MaxUses` from defaults, then set `ExpiresAt = nil` for `duration == 0`):

```go
applyAutoInviteDefaults(inviteModel, now) // MaxUses + default ExpiresAt
if req.Duration != nil {
    switch {
    case *req.Duration == 0:
        inviteModel.ExpiresAt = nil          // permanent
    case *req.Duration > 0:
        t := db.Time(now.Add(time.Duration(*req.Duration) * time.Second))
        inviteModel.ExpiresAt = &t
    default:
        respondSpaceRequestInvalid(c, "duration")
        return
    }
}
```

(Combined with the P1 `ShouldBindJSON` fix from the main review, both regressions are covered.)

lml2468 (dismissed):
# PR Review (re-review) — PR #637 (Mininglamp-OSS/octo-server)

Re-reviewed at **new head `bec81682`** against merge-base `ca72704c` (base `main`). Supersedes my COMMENT at `f4addabc`.
**Build/vet/test:** `go build ./modules/space/` exit 0 ✅, `go vet` exit 0 ✅. `go test` DB-gated (MySQL refused in sandbox) — unrun; CI to confirm.

## What changed since `f4addabc` (delta reviewed)
One commit — `bec81682` "address CR — ShouldBindJSON, duration cap, MaxUses default" — closing every item from my COMMENT and the peer reviews.

### P2 — `time.Duration` overflow / already-expired invite (mine) — FIXED ✅
New `maxInviteDurationSeconds = 3650*24*3600` (~10y); `*req.Duration > maxInviteDurationSeconds → 400`. This is well below the `time.Duration` int64-ns overflow point (~292y), so the multiply can no longer wrap to a past time. The misleading struct comment ("超长值在写入层截断" — a truncation that didn't exist) is **replaced** with an accurate one documenting both the overflow bound and the DB-column constraint. My finding is resolved.

### TIMESTAMP 2038 (mine + Steve) — addressed ✅
The 10y cap keeps `now + duration` under the MySQL `TIMESTAMP` 2038-01-19 ceiling, and the comment states the assumption explicitly ("worst case: 2028 + 10 = 2038" → valid while the endpoint is called before 2028) plus a note to re-evaluate the column type if the cap changes. Reasonable, honestly bounded.

### empty/malformed body → spurious 400 (yujiawei 🔴) — FIXED ✅
`BindJSON` → `ShouldBindJSON`: an empty/EOF body no longer aborts with a 4xx; it leaves `Duration == nil` → the default-TTL path. Backward compatibility for callers sending no body is preserved (and my "400-then-create" concern is moot — every reject is a clean `respond…; return` with no subsequent insert).

### MaxUses default skipped on the duration branch (mochashanyao) — FIXED ✅
`applyAutoInviteDefaults` now runs **unconditionally first** (sets `MaxUses` when 0, sets the default `ExpiresAt`), and the duration switch only overrides `ExpiresAt` afterward. Verified `applyAutoInviteDefaults` sets `MaxUses` — so the MaxUses policy now always applies regardless of the duration branch. The `0 = permanent` case correctly **clears** the just-applied default `ExpiresAt` (`inviteModel.ExpiresAt = nil`), which is necessary now that defaults run first — a subtle interaction the author got right.

## Verified clean
- **Control flow correct across all four branches:** `duration<0 → 400`, `>max → 400`, `==0 → permanent (ExpiresAt nil)`, `>0 → now+duration`; `duration` omitted → default TTL (from the unconditional `applyAutoInviteDefaults`). Each 400 is an early return with no insert.
- **Auth/isolation unchanged** — `checkSpaceActive` + `queryMember` + `Role<1 → PermissionDenied` still gate before any parsing (space-isolation preserved).
- **`expire` response field** unchanged (RFC3339 or `""` for permanent).

## Residual (non-blocking)
- **Still no unit test** for the four branches + the cap boundary. The logic is now simple and correct on inspection, and the package tests are DB-gated, but a table test (nil→default, 0→permanent, >0→TTL, <0→400, >max→400) would lock the contract and the overflow fix. Recommend as fast-follow — not a blocker given the reduced complexity.

## Verdict
**APPROVE** (upgraded from my COMMENT at `f4addabc`) — every item is addressed: the duration is now bounded to ~10y (closing the `time.Duration` overflow that could silently create an already-expired invite, and staying under the 2038 TIMESTAMP ceiling), the misleading truncation comment is corrected to describe the real bounds, `ShouldBindJSON` preserves empty-body backward compatibility without a spurious 400, and `applyAutoInviteDefaults` runs unconditionally so the MaxUses policy always applies with `0=permanent` correctly clearing the default. Auth/isolation unchanged. Both gates pass.
Suggested decision: `Safe to merge` once CI runs the DB-gated space suite; add the branch/boundary table test as fast-follow. My thanks to yujiawei / mochashanyao / Steve — the empty-body and MaxUses catches complemented the overflow finding. (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (dismissed):
Re-review at head `bec81682`. One of my two prior blockers is fully resolved; the second is only partially mitigated and still fails on a foreseeable timeline, so this remains **REQUEST_CHANGES**.

## ✅ Resolved

**🔴(1) int64 nanosecond overflow — FIXED.** `modules/space/api.go:922` introduces `const maxInviteDurationSeconds = int64(3650 * 24 * 3600)` (~10 years), and `modules/space/api.go:964` rejects `*req.Duration > maxInviteDurationSeconds` with 400 **before** the `time.Duration(*req.Duration) * time.Second` conversion at line 973. ~10 years in nanoseconds (~3.15e17) is far below the `time.Duration`/int64 ceiling (~292 years), so the seconds→nanoseconds multiply can no longer wrap negative. The previously misleading struct comment claiming "over-long values are truncated at the write layer" has been corrected to state that over-limit values now return 400 — comment and behavior now agree. Good fix.

Negative → 400, `0` → permanent (ExpiresAt nil), omitted → default TTL via `applyAutoInviteDefaults`, positive → now+duration: all still correct. The refactor to `switch` plus always-apply-defaults-first (so `MaxUses` policy always lands) is a clean improvement, and pointer semantics (unset vs 0 vs >0 vs <0) are intact. `expire` additive field and authz (Role<1 reject before parse) unchanged. `go vet ./modules/space/` is green.

## 🔴 Still blocking

**🔴(2) storage-column 2038 ceiling — NOT fixed.** The `space_invitation.expires_at` column is still declared as MySQL `TIMESTAMP` (see `modules/space/sql/20260307000002_space_legacy01.sql:44`), whose maximum representable instant is 2038-01-19. There is no migration in `modules/space/sql/` changing this column to `DATETIME`.

The new 10-year bound is **relative to `now`**, not an absolute ceiling. The code's own comment (`modules/space/api.go:914-921`) concedes this: the invariant `now + duration <= 2038-01-19` only holds while the call happens before ~2028 (worst case 2028 + 10 = 2038). After roughly January 2028, a caller-supplied max-duration value computes an `expires_at` beyond the column's representable range, reproducing exactly the failure mode I originally flagged: depending on `sql_mode`, the insert either errors out or the value is silently coerced, yielding a wrong/expired invite. So the fix moves the cliff from "292 years" to "~2028," but the cliff still exists on a foreseeable calendar date rather than being eliminated.

**Fix (pick one):**
- (a) Make the bound absolute against the column ceiling, e.g. clamp/reject when `now.Add(duration)` would exceed the TIMESTAMP maximum (compute against a fixed 2038-01-19 UTC ceiling rather than a fixed 10-year window), so the guarantee holds regardless of the current date; or
- (b) Migrate `space_invitation.expires_at` from `TIMESTAMP` to `DATETIME` (with a reversible Down migration) so the representable range covers year-9999, then the 10-year relative bound is genuinely safe.

Option (a) is the smaller change and keeps behavior date-stable; (b) is the more durable choice if long-lived invites are a real product requirement.

## 🟡 Non-blocking

- No unit tests for the new duration logic (default / explicit positive / 0-permanent / negative-400 / at-boundary / over-boundary-400 cases). These would lock in the exact contract and are cheap to add.
- The `ShouldBindJSON` error-swallowing is now well commented at `modules/space/api.go` (empty/EOF body falls back to default TTL, no abort) — that addresses the earlier note.

Once 🔴(2) is closed (absolute ceiling check or column migration), I'll approve.

yujiawei (dismissed):
# Code Review — PR #637 (octo-server)

**Head SHA:** `bec8168252231ac5ac76d48f51c08514250686c5`
**Scope:** single file, `modules/space/api.go` (+59 / −1) vs merge-base with `main`.
**Build/vet:** `go build ./modules/space/` exit 0 ✅ · `go vet ./modules/space/` exit 0 ✅.

This is a re-review after the CR round. The head commit (`fix(space): address CR — ShouldBindJSON, duration cap, MaxUses default`) resolves all three previously-blocking findings. Details below.

## 1. Spec compliance — ✅

- Requested `duration` param — implemented via `createInviteReq{ Duration *int64 }` with the `nil / 0 / >0 / <0` branch tree (api.go:910, 961–976).
- Requested additive `expire` response field — implemented (RFC 3339 UTC for a TTL invite, `""` for permanent) (api.go:986–994).
- `duration` omitted → default-TTL path preserved (`applyAutoInviteDefaults`).
- No unrequested endpoints/flags/fields introduced; nothing out of scope.

## 2. Code quality — Approved

All three blockers from the prior round are fixed:

### ✅ Resolved — empty/malformed body no longer 400s (was P1)
`_ = c.ShouldBindJSON(&req)` (api.go:944) replaces the aborting `BindJSON`. `ShouldBind*` returns the decode error without writing headers or aborting, so an empty body (`io.EOF`) leaves `Duration == nil` and falls through to the default-TTL path — the stated backward-compat guarantee now actually holds. A clarifying comment documents the intent.

### ✅ Resolved — `MaxUses` operator default always applied (was P1)
`applyAutoInviteDefaults(inviteModel, now)` now runs **unconditionally** (api.go:957) before the duration branch, which only overrides `ExpiresAt`. Every invite — including those created with an explicit `duration` (0 or positive) — now inherits the operator-configured `DM_SPACE_INVITE_DEFAULT_MAX_USES`. The base-branch regression is gone.

### ✅ Resolved — duration upper bound prevents overflow / past-timestamp invite (was P2)
A named `maxInviteDurationSeconds = 3650*24*3600` (~10y) cap (api.go:922) rejects oversized values with a clean 400 (api.go:963) before the `time.Duration` multiply. This closes the prior overflow path where a pathological `duration` (>~9.2e9 s) wrapped `time.Duration(seconds) * time.Second` into a negative value and silently minted an already-expired invite. The cap keeps the arithmetic within range on both fronts: `315360000 * 1e9 ns ≈ 3.15e17` ≪ `MaxInt64`, and `now + 10y` stays below the MySQL `TIMESTAMP` ceiling (2038-01-19) for calls before 2028. The doc comment now accurately describes the cap (the earlier misleading "written-layer truncation" claim is gone).

### Verified clean
- **Auth/isolation unchanged:** `checkSpaceActive` + `queryMember` + `member == nil || Role < 1 → ErrSpacePermissionDenied` still gate the handler before any new logic (api.go:927–939). Non-member/non-admin cannot create invites.
- **Permanent link:** `duration == 0` → `ExpiresAt = nil` → DB writes NULL; `queryInvitationByCode` treats `expires_at IS NULL` as valid. Correct.
- **Negative duration:** → 400 `respondSpaceRequestInvalid(c, "duration")` before any DB write.
- **`expire` response field:** additive; consumers reading only `invite_code` are unaffected.
- **`pkg/db` import** for `db.Time`: needed, compiles.
- **Diff scope:** genuinely single-file; the only caller of `createInvite` is its route registration — no cross-file fallout.

## 3. Non-blocking notes (P2 — do not block merge)

- **Field naming/format divergence:** the new field is `expire` in RFC 3339, whereas other invite/expiry fields in this module serialize as `expires_at` in `2006-01-02 15:04:05` form. Additive so not breaking, but two spellings/formats now coexist for the same concept. Consider aligning to `expires_at`, or confirm `expire`+RFC3339 is deliberate for the paired frontend.
- **No tests:** the new branches (omitted → default TTL, `0` → permanent, `>0` → TTL, `<0` → 400, `> max` → 400, empty body → default) are uncovered. A small table test would lock the contract and guard the cap/MaxUses fixes against regression. Package tests are DB-gated (MySQL), so CI should run the space suite.
- **`TIMESTAMP` longevity:** the 2038 ceiling is a latent module-wide constraint, not introduced here; the cap comment already flags it. Widening `space_invitation.expires_at` to `DATETIME` is a reasonable future follow-up but out of scope for this PR.

## 4. Verdict

**APPROVED** — all prior blocking findings (empty-body 400, MaxUses bypass, duration overflow) are correctly resolved; the change is small, auth-gated, backward-compatible, and builds clean. Remaining items are non-blocking P2 polish (field-name consistency, tests).

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #637 (octo-server)

## Summary

This PR adds a `duration` parameter to the user-facing `createInvite` endpoint (`POST /v1/space/:space_id/invite`), enabling callers to control invite-link expiry. A `duration=0` value creates a permanent link, positive values set a specific TTL, negative values and values exceeding `maxInviteDurationSeconds` (~10 years) are rejected with 400, and omitting the field preserves the existing default-TTL behaviour via `applyAutoInviteDefaults`. The response gains an additive `"expire"` field (RFC 3339 or empty string). The implementation is clean, defensive, and backward-compatible.

## Verification

Static analysis only at head `bec8168`; build and tests not executed in this environment.

- ✅ **Backward compatibility** — callers sending empty or missing body still receive default-TTL behaviour: `ShouldBindJSON` silently leaves `req.Duration` as `nil` on parse failure, and the `if req.Duration != nil` gate preserves `applyAutoInviteDefaults` output unchanged (`modules/space/api.go:948-982`).
- ✅ **Overflow safety** — `maxInviteDurationSeconds` (315,360,000 s) × `time.Second` (10⁹ ns) = 3.15 × 10¹⁷ ns, well within `int64` max (9.22 × 10¹⁸). No `time.Duration` overflow risk.
- ✅ **MySQL TIMESTAMP ceiling** — worst case (2026 + 10 years = 2036) stays below the TIMESTAMP upper bound (2038-01-19). The code comment at `:919-921` correctly flags this constraint for future maintainers.
- ✅ **Data-flow: ExpiresAt** — `applyAutoInviteDefaults` fills `ExpiresAt` when nil (`invite_code.go:82-84`); the duration switch then overwrites or clears it. The downstream consumer `insertInvitationWithRetry` receives a fully populated model. Verified `db.Time` is `type Time time.Time` (`pkg/db/db.go:11`), so the `db.Time(...)` / `time.Time(...)` conversions are safe type conversions.
- ✅ **Data-flow: expire response string** — when `ExpiresAt == nil` (permanent), `expireStr = ""`; otherwise RFC 3339 UTC. The `map[string]string` response type is consistent with other endpoints in this file (`:1113`).
- ✅ **Error handling** — `ShouldBindJSON` error is intentionally discarded (`:948`), documented as a deliberate design choice to avoid 4xx on empty body. Negative duration, excessive duration, and valid duration paths all have explicit handling.
- ✅ **C1 — Double-path parity** — `Manager.createInvite` (`api_manager.go:867`) is intentionally a separate path using `ExpiresAt` (absolute timestamp) rather than `Duration` (relative seconds). The `applyAutoInviteDefaults` doc comment explicitly warns against using it from the admin path. The two paths have different request shapes by design.
- ✅ **C2 — Control-flow ordering** — no nesting or reuse issues. The duration switch is a single flat decision tree with no side effects.
- ✅ **C3–C6** — N/A. No auth boundary, container-state cascade, build/packaging, or governance changes.

## Findings

### P2 — Response field name and date-format diverge from module convention (`modules/space/api.go:993`)

The new response uses field name `"expire"` with `time.RFC3339` (UTC, e.g. `"2026-07-22T14:57:03Z"`). Every other invite endpoint in this module uses `"expires_at"` with `inviteTimeLayout` (`"2006-01-02 15:04:05"`, local time):

- `inviteResp` at `modules/space/model.go:197` — `ExpiresAt string \`json:"expires_at"\``
- `Manager.createInvite` at `modules/space/api_manager.go:948-951` — `"expires_at"` + `inviteTimeLayout`
- List/detail endpoints at `modules/space/api.go:1228`, `:1288`, `:1418` — same format

Frontend consumers now need to handle two different field names (`expire` vs `expires_at`) and two different date formats (RFC 3339 UTC vs `"2006-01-02 15:04:05"` local) for the same concept across invite endpoints. Not a blocker — the pairing frontend PR (octo-web#972) may be intentionally designed around the new format — but worth aligning if there's no specific reason for the divergence.

- `ShouldBindJSON` vs `BindJSON` choice: correctly avoids aborting on empty body, preserving backward compatibility.
- `applyAutoInviteDefaults` always runs first, ensuring `MaxUses` is always populated from operator config regardless of duration override.
- `insertInvitationWithRetry` receives a fully populated `InvitationModel` — no field is left uninitialized.
- The `time.Duration(*req.Duration) * time.Second` computation is bounds-checked before reaching the multiplication.
- The permanent-link path (`duration == 0`) correctly sets `ExpiresAt = nil`, and the response correctly returns an empty string for this case.

## Verdict: COMMENT

Well-implemented feature with correct backward compatibility, solid overflow/underflow guards, and clean data flow. The only finding is a P2 convention divergence in the response field name and date format. Not a merge blocker.

[Octo-Q] verdict: APPROVE — only P2 (convention divergence in response field name/format), no P0/P1. Recommend Opus consider whether the `expire`/RFC3339 choice is intentional for the frontend pairing PR.

lml2468 (approved):
# PR Review (re-review) — PR #637 (Mininglamp-OSS/octo-server)

Re-reviewed at **new head `4a0932b4`** against merge-base `ca72704c` (base `main`). Supersedes my APPROVE at `bec81682` — which was an **under-call**: I approved while the 2038 TIMESTAMP ceiling was still only *relatively* bounded. Steve/Jerry-Xin correctly held CR on it; this commit closes it.
**Build/vet/test:** `go build ./modules/space/` exit 0 ✅, `go vet` exit 0 ✅. `go test` DB-gated (MySQL) — CI to confirm.

## Correction note — I under-called the 2038 ceiling at `bec81682`
At `bec81682` I treated the ~10y `maxInviteDurationSeconds` cap as sufficient for the MySQL `TIMESTAMP` bound. It wasn't: that cap is **relative to `now`**, so `now + 10y` stays under 2038-01-19 only while the endpoint is called before ~2028 — after that, a max-duration invite computes an `expires_at` beyond the column's range and (per `sql_mode`) errors or truncates to a wrong value. The cliff was moved from ~292y to ~2028, not removed. Jerry-Xin and Steve flagged this on the prior head; they were right and my APPROVE was premature. This commit fixes it properly.

## What changed since `bec81682` (delta reviewed)
One commit — `4a0932b4` "clamp against MySQL TIMESTAMP ceiling (2038-01-19)".

### 🔴 2038 TIMESTAMP ceiling — FIXED ✅ (absolute check)
New `mysqlTimestampMax = time.Date(2038, 1, 19, 3, 14, 7, 0, time.UTC)`, and the positive-duration branch now computes `candidate := now.Add(...)` and rejects with 400 if `candidate.After(mysqlTimestampMax)`. This is an **absolute** bound — date-independent — so it holds no matter when the endpoint is called, unlike the relative 10y cap. The `maxInviteDurationSeconds` comment is corrected to state its role is *only* overflow prevention, with the TIMESTAMP bound now owned by the absolute check. Correct division of responsibility.

## Verified clean
- **Guard order composes correctly:** `duration<0 → 400`; `>maxInviteDurationSeconds → 400` (int64-ns overflow guard, ~10y ≪ ~292y); `==0 → permanent (ExpiresAt nil)`; else compute `candidate` → `After(2038) → 400` → else set. Every reject is an early return with no insert. Overflow guard runs *before* the `now.Add` multiply, so `candidate` can't itself overflow before the 2038 comparison.
- **From `bec81682`, still holds:** `ShouldBindJSON` (empty body → default TTL, no spurious 400); `applyAutoInviteDefaults` unconditional (MaxUses always applied; `0=permanent` clears the default ExpiresAt); auth/isolation (`checkSpaceActive`+`queryMember`+`Role<1→denied`) unchanged; `expire` response field unchanged.

## Non-blocking observations
- **Default-TTL path is not subject to the 2038 absolute check** — `inviteDefaults` computes `now.Add(DM_SPACE_INVITE_DEFAULT_TTL)` without the ceiling clamp, so a pathologically large operator-set env default could still exceed the column. But this is **pre-existing** (this PR doesn't touch `inviteDefaults`), **operator-controlled** (not user input), and defaults to 72h — not introduced or newly activated here. Worth a fast-follow to route the default path through the same clamp, but out of scope for this fix and not a blocker.
- **Still no unit test** (lml2468/yujiawei/Jerry-Xin all noted). The branch/boundary logic (nil/0/positive/negative/>max/>2038) is now the natural thing to table-test and would pin both the overflow and 2038 fixes. Strongly recommend as fast-follow; not blocking given the logic is simple and correct on inspection.
- **`expire` field naming/format vs sibling `expires_at`** (mochashanyao/yujiawei 🟡) — a cross-endpoint consistency preference, not a correctness issue; author's call.

## Verdict
**APPROVE** (correcting my premature APPROVE at `bec81682`) — the 2038 ceiling is now enforced with an **absolute** `mysqlTimestampMax` check that's independent of call date, composing correctly after the overflow cap and the empty-body/MaxUses fixes from the prior round. The user-supplied duration path — this PR's subject and the peers' 🔴 — can no longer produce an out-of-range or already-expired `expires_at`. Auth/isolation unchanged. Both gates pass. Thanks to Jerry-Xin / Steve — the relative-vs-absolute distinction was the right call and I'd missed it.
Suggested decision: `Safe to merge` once CI runs the DB-gated space suite. Fast-follow: route the default-TTL path through the same 2038 clamp, and add the branch/boundary table test. (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (approved):
## Re-review @ 4a0932b4 — APPROVE

Both prior blockers are now resolved. The remaining 🔴 (MySQL `TIMESTAMP` 2038 ceiling) is fixed correctly with a **fixed absolute bound**, not another now-relative cap.

### ✅ Blocker resolved — 2038 ceiling now an absolute check
`modules/space/api.go`:
- New fixed constant `mysqlTimestampMax = time.Date(2038, 1, 19, 3, 14, 7, 0, time.UTC)` — an absolute UTC instant, independent of the call date.
- Positive-duration branch now computes the candidate `expires_at` and rejects when it exceeds the ceiling:
  ```go
  candidate := now.Add(time.Duration(*req.Duration) * time.Second)
  if candidate.After(mysqlTimestampMax) {
      respondSpaceRequestInvalid(c, "duration")
      return
  }
  ```
- Because the comparison is against the **computed absolute** `expires_at` (not `duration` alone), the cliff is eliminated rather than moved: e.g. a 5-year duration requested in 2035 yields a 2040 candidate → `After(mysqlTimestampMax)` → 400. This holds for any call date.
- The previously misleading comment (claiming the 10-year cap only held if called before ~2028) was removed. `maxInviteDurationSeconds` is now correctly documented as an overflow guard only, with the column ceiling delegated to the absolute check.

### ✅ Overflow guard retained
`maxInviteDurationSeconds` (~10 years) still rejects `> cap` and `< 0` before the absolute check, keeping the int64-nanosecond overflow protection intact.

### ✅ Sound parts intact (no regression)
- `applyAutoInviteDefaults` still applied first, so the `MaxUses` default always takes effect.
- Pointer `*int64 Duration`: nil → default TTL; `0` → `ExpiresAt = nil` (permanent, explicitly clears the default); `> 0` → absolute-checked override; `< 0` → 400.
- `ShouldBindJSON` unchanged (empty body → default, no spurious 400).
- Authz path (`checkSpaceActive` + `queryMember` + `Role < 1` reject) unchanged.
- `expire` response field additive and RFC3339; permanent → empty string.

### 💬 Non-blocking
- 🟡 Still no unit tests for the duration branches (default / 0 / negative / over-cap / at-cap / 2038-boundary). The logic is small and readable, but a table-driven test covering the 2038 boundary (a candidate just past the ceiling → 400) and the permanent/default cases would lock the contract. Recommended follow-up, not a merge blocker.

Build and vet are green for `modules/space`. LGTM.

yujiawei (approved):
# Code Review — PR #637 (octo-server)

**Head SHA:** `4a0932b44a1f5eb4770723654dfbbb431e2d466e`
**Scope:** single file, `modules/space/api.go` (+65 / −1) vs merge-base with `main`.
**Build/vet:** `go build ./modules/space/` exit 0 ✅ · `go vet ./modules/space/` exit 0 ✅.

Re-review after the CR rounds. The head commit (`fix(space): clamp against MySQL TIMESTAMP ceiling (2038-01-19)`) closes the last remaining blocker. All previously-raised blocking findings are now resolved.

## 1. Spec compliance — ✅

- Requested `duration` param — implemented via `createInviteReq{ Duration *int64 }` with the `nil / 0 / >0 / <0` branch tree.
- Requested additive `expire` response field — implemented (RFC 3339 UTC for a TTL invite, `""` for permanent).
- `duration` omitted → default-TTL path preserved (`applyAutoInviteDefaults`).
- No unrequested endpoints/flags/fields introduced; nothing out of scope.

## 2. Code quality — Approved

All prior blockers are fixed and verified against the current head:

### ✅ Resolved — empty/malformed body no longer 400s
`_ = c.ShouldBindJSON(&req)` (api.go:945) replaces the aborting `BindJSON`. `ShouldBind*` returns the decode error without writing headers or aborting, so an empty body (`io.EOF`) leaves `Duration == nil` and falls through to the default-TTL path — the backward-compat guarantee now actually holds. A clarifying comment documents the intent.

### ✅ Resolved — `MaxUses` operator default always applied
`applyAutoInviteDefaults(inviteModel, now)` now runs **unconditionally** (api.go:958) before the duration branch, which only overrides `ExpiresAt`. Every invite — including those created with an explicit `duration` (0 or positive) — inherits the operator-configured `DM_SPACE_INVITE_DEFAULT_MAX_USES`. The `0 = permanent` case correctly clears the just-applied default `ExpiresAt`.

### ✅ Resolved — `time.Duration` overflow (already-expired invite)
A named `maxInviteDurationSeconds = 3650*24*3600` (~10y) cap rejects oversized values with a clean 400 before the `time.Duration(*req.Duration) * time.Second` multiply. `315360000 * 1e9 ns ≈ 3.15e17` ≪ `MaxInt64`, so the multiply can no longer wrap to a past time.

### ✅ Resolved (this head) — MySQL `TIMESTAMP` 2038 ceiling
The `space_invitation.expires_at` column is still `TIMESTAMP` (`modules/space/sql/20260424000002_space_legacy01.sql:15`), max 2038-01-19. The prior head guaranteed `now + duration ≤ 2038` only for calls made before ~2028 (a relative window, i.e. a future cliff). This head adds an **absolute** ceiling: `mysqlTimestampMax = 2038-01-19 03:14:07 UTC`, and the positive-duration branch now rejects with 400 when `candidate.After(mysqlTimestampMax)` (api.go:973–979). The guarantee is now date-stable regardless of when the endpoint is called, and the earlier misleading comment about a relative bound is corrected to describe the absolute check.

### Verified clean
- **Auth/isolation unchanged:** `checkSpaceActive` + `queryMember` + `member == nil || Role < 1 → ErrSpacePermissionDenied` gate the handler before any new logic. Non-member/non-admin cannot create invites.
- **Permanent link:** `duration == 0` → `ExpiresAt = nil` → DB writes NULL; `queryInvitationByCode` treats `expires_at IS NULL` as valid.
- **Negative / over-cap duration:** → 400 `respondSpaceRequestInvalid(c, "duration")` as an early return before any DB write.
- **`expire` response field:** additive; consumers reading only `invite_code` are unaffected.
- **Write path:** `insertInvitation` writes `time.Time(*ExpiresAt)` (or NULL) directly — nil-safe, consistent with the model.
- **Diff scope:** genuinely single-file; the only caller of `createInvite` is its route registration — no cross-file fallout.

## 3. Non-blocking notes (P2 — do not block merge)

- **Default-TTL path not bounded against the ceiling.** The new struct comment states "all paths writing `expires_at` should ensure `time.Time <= mysqlTimestampMax`", but the omitted-`duration` path (`applyAutoInviteDefaults` → `inviteDefaults`) applies `now + DM_SPACE_INVITE_DEFAULT_TTL` without that check. In practice this only matters if an operator sets an extreme `DM_SPACE_INVITE_DEFAULT_TTL` (env-controlled, not caller-controlled; default is 72h), so it is not a merge blocker — but for the invariant in the comment to hold literally, the default path would need the same clamp.
- **Field name/format divergence.** The new field is `expire` in RFC 3339 UTC, whereas other invite/expiry fields in this module serialize as `expires_at` in `2006-01-02 15:04:05` form. Additive so not breaking, but two spellings/formats now coexist for the same concept. Align to `expires_at`, or confirm `expire`+RFC3339 is deliberate for the paired frontend (octo-web #972).
- **No tests.** The new branches (omitted → default TTL, `0` → permanent, `>0` → TTL, `<0` → 400, `> max` → 400, `> 2038` → 400, empty body → default) are uncovered. Package tests are DB-gated (MySQL), so CI should run the space suite; a small table test would lock the contract and guard the cap/ceiling/MaxUses fixes against regression. Recommended as fast-follow.

## 4. Verdict

**APPROVED** — all prior blocking findings (empty-body 400, MaxUses bypass, duration overflow, and the 2038 `TIMESTAMP` ceiling) are correctly resolved. The change is small, auth-gated, backward-compatible, and builds clean. Remaining items are non-blocking P2 polish (default-path symmetry, field-name consistency, tests).

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #637 (octo-server)

## Summary

This PR extends the user-facing `POST /:space_id/invite` endpoint to accept an optional `duration` parameter (seconds) controlling invite expiry, and returns an `expire` field in the response. Duration=0 means permanent (no expiry), positive values set an absolute expiry clamped against the MySQL TIMESTAMP ceiling (2038-01-19), and values outside the 0–10-year range are rejected with 400. The implementation is careful about overflow protection and MySQL compatibility. Two non-blocking items below.

## Verification

Static analysis only at head `4a0932b44a1f5eb4770723654dfbbb431e2d466e`; build and tests not executed in this environment.

- ✅ **MySQL TIMESTAMP ceiling clamp** — `mysqlTimestampMax` at `modules/space/api.go:917` is correct (2038-01-19 03:14:07 UTC). The `candidate.After(mysqlTimestampMax)` check at `:978` correctly rejects times strictly after the ceiling. Column type confirmed as `TIMESTAMP NULL DEFAULT NULL` in `modules/space/sql/20260424000002_space_legacy01.sql:15`.
- ✅ **time.Duration overflow protection** — `maxInviteDurationSeconds` (315,360,000s ≈ 10 years) at `:923` keeps `time.Duration(*req.Duration) * time.Second` well within int64 nanosecond range (~3.15×10¹⁷ ns vs 9.22×10¹⁸ max).
- ✅ **Data flow: applyAutoInviteDefaults → duration override** — `applyAutoInviteDefaults` at `:957` sets `ExpiresAt` from env TTL (default 72h) and `MaxUses`. The duration switch then correctly overrides only `ExpiresAt`, leaving `MaxUses` policy intact. For `duration=0`, `ExpiresAt` is set to nil (permanent). For nil duration, defaults are preserved. Flow is correct.
- ✅ **Negative and over-range rejection** — `< 0` and `> maxInviteDurationSeconds` cases at `:964–970` both return 400 via `respondSpaceRequestInvalid`.
- ✅ **Response `expire` field** — RFC3339 UTC format for non-nil `ExpiresAt`, empty string for permanent. Nil-check at `:991` is correct; `time.Time(*inviteModel.ExpiresAt).UTC().Format(time.RFC3339)` produces valid output.

## Findings

No P0/P1 issues. One P2 and one nit below.

### P2 — Malformed JSON body silently ignored (`modules/space/api.go:946`)

`_ = c.ShouldBindJSON(&req)` discards the parse error. The comment at `:945` correctly explains the empty-body rationale — avoiding `BindJSON`'s abort on EOF. But this also swallows genuine parse errors on non-empty bodies: a client sending `{"duration": "abc"}` or a JSON number exceeding int64 range gets HTTP 200 with an invite at the default 72h TTL, with no signal that their intended duration was discarded.

The manager path at `api_manager.go:893` handles this correctly with `ContentLength > 0` gating before `BindJSON`.

Diff-scope: new — the old `createInvite` had no body parsing at all. This is a silent-failure mode introduced alongside the new `duration` feature, not a regression.

Fix direction:

```go
var req createInviteReq
if c.Request.ContentLength > 0 {
    if err := c.ShouldBindJSON(&req); err != nil {
        respondSpaceRequestInvalid(c, "")
        return
    }
}
```

### Nit — Response time format diverges from codebase convention (`modules/space/api.go:991`)

`time.RFC3339` (e.g. `"2026-07-22T16:00:00Z"`) is used for the `expire` field, while every other invite endpoint uses `inviteTimeLayout` (`"2006-01-02 15:04:05"`, local time) — see `api_manager.go:948`, `api.go:1234`, `:1294`, `:1424`. RFC3339 is arguably better for API consumption (unambiguous timezone), but clients that already parse `expires_at` from the manager or list endpoints will need two parsers. If intentional, no action needed.

- `applyAutoInviteDefaults` is always called before the duration override, ensuring `MaxUses` policy is never bypassed by the new `duration` field.
- `insertInvitationWithRetry` contract unchanged — it expects pre-filled fields, which this path provides.
- `db.Time` ↔ `time.Time` conversion round-trips correctly for the response format.
- The `createSpaceCorePostCommit` path at `api.go:294` is untouched and continues to use `applyAutoInviteDefaults` without duration — correct, as space-creation auto-invites should not have user-controlled TTL.
- C1 (symmetric paths): the manager `createInvite` at `api_manager.go:100` and email-invite paths are not modified by this PR. No parity concern — different API surfaces with different contracts.

## Verdict: APPROVED

No correctness, security, or build-breaking issues. The duration validation logic is well-designed with proper overflow and MySQL TIMESTAMP ceiling guards. The P2 (silent JSON parse error) is a non-blocking improvement worth addressing to prevent client confusion when malformed requests are silently accepted.

[Octo-Q] verdict: APPROVE — no P0/P1 found; P2 on ShouldBindJSON silent discard is non-blocking.

an9xyz (changes requested):
感谢推进这个问题。但经排查,这个改动恐怕没有解决报障场景,方向上建议调整:

**1. 报障现场(octo-admin)不走这个接口**

"选了永久仍失效"的邀请码管理界面在 octo-admin(`SpaceInvitesPanel`,带 7d/14d/30d/never 预设)。它分两条 scope:
- 平台超管 → `POST /v1/manager/spaces/:id/invites`(`api_manager.go` 的 `m.createInvite`)
- 空间管理员 → `POST /v1/space/:id/invite` + `PUT /v1/space/:id/invite/:code`,且传的是 `expires_at`(绝对时间),靠"先建再改"补参数

本 PR 只给 `POST /v1/space/:id/invite` 新增了 `duration` 字段。octo-admin 不发 `duration`(发 `expires_at`),所以对 octo-admin 的请求:改前 body 被忽略 → 72h;改后 `req.Duration == nil` → 仍走 `applyAutoInviteDefaults` → 72h。**前后完全一致,等于 no-op。**

**2. 根因未触及:空 `expires_at` 落不了 NULL**

"永久"在 server 侧被表达成空 `expires_at`,但两条写入路径都把它当成"没传":
- 建码 `m.createInvite`:空 `expires_at` 会落到 `else if defExpiresAt != nil` 分支被强制填 72h(`inviteDefaults` 在默认/环境变量下恒返回非 nil 过期时间);
- 改码 `updateInvitationAdmin`:`expiresAt == nil` 时直接跳过该列,无法把 `expires_at` 写成 NULL。

因此当前**没有任何路径能产出永久(NULL)邀请码**——这才是"选永久仍 72h 失效"的根因。

**建议**
- 在 manager / 用户侧建改码 handler 支持"永久 → `expires_at = NULL`",并区分"未传(套默认)"与"显式永久";
- 顺带统一 `duration` / `expires_at` / `expiresInDays`(docs 邀请)三套过期契约;
- 与 octo-admin 侧确认 never 真正生效。

配套前端评审见 Mininglamp-OSS/octo-web#972。

---

PR #638, opened 2026-07-22
fix: make reaction emoji identity byte-exact

### LINKED ISSUE #636: Reaction upsert can toggle the wrong emoji under general_ci
## Problem

Adding a new reaction emoji can toggle a different persisted emoji.

Observed in the test environment:

- The user already had a 👍 reaction.
- Requests for 🎉 returned a successful 🎉 toggle response.
- Reaction sync and the database showed that the 👍 row was toggled instead.
- Refresh correctly restored the persisted 👍 state and removed the optimistic 🎉 state.

## Root cause

reaction_users.emoji inherits utf8mb4_general_ci. On the deployed MySQL-compatible database, that collation compares some distinct supplementary-plane emoji, including 👍 and 🎉, as equal.

The unique key on channel_id, channel_type, message_id, uid, and emoji therefore routes 🎉 into the existing 👍 upsert row. The handler then masks the mismatch by echoing the requested emoji instead of returning the persisted result.

## Expected behavior

- Emoji identity is byte-exact.
- 👍 and 🎉 are independent reaction rows.
- Toggling 🎉 never changes 👍.
- Write responses and sync commands use the persisted emoji, sequence, and deletion state.

## Acceptance criteria

- Change only reaction_users.emoji to utf8mb4_bin through a forward migration.
- Keep the existing five-column unique key.
- Return the persisted toggle result from the DB helper.
- Fail closed if requested and persisted emoji differ.
- Cover 👍 plus 🎉 creation, sync, and independent cancellation.
- Preserve existing reaction rows during migration.
- Check production table size before rollout because the collation change may rebuild the table or hold a metadata lock.

an9xyz:
## Summary

Fix message reaction collisions caused by linguistic collation treating distinct emoji as equal.

## Related Issue

Fixes #636

## Linked Spec

.octospec/tasks/message-reaction-emoji-binary-collation/brief.md

## Changes

- Add a forward-only migration that changes only reaction_users.emoji to utf8mb4_bin.
- Return the persisted emoji, sequence, and deletion state from the atomic toggle helper.
- Build the sync command and HTTP response from the persisted result and fail closed on an emoji mismatch.
- Add regression coverage for independent 👍 and 🎉 creation, sync, and cancellation.
- Cover DB insert and read-back error paths.

## Testing

- [x] Manually verified

Commands and checks run:

- go test -race -shuffle=on -count=1 ./modules/message
- Focused MySQL integration tests for TestReactionEmojiColumnUsesBinaryCollation and TestReactionMultipleEmojisIndependent
- Focused integration tests with the race detector
- go build ./modules/message
- go vet ./modules/message
- gofmt and git diff --check
- Verified utf8mb4_bin and the five-column unique index in MySQL
- Verified an existing reaction row survives the collation migration

The repository-wide integration-tag suite has a pre-existing test import cycle. The reaction integration tests were executed through an isolated same-package source set so the real handlers, migrations, MySQL, Redis, and sync path still ran.

## COMPREHENSION

   Reaction emoji identity becomes byte-exact at the storage layer. The existing atomic upsert continues to use the five-column unique key, but distinct emoji no longer collide under utf8mb4_general_ci. The write path also returns and broadcasts the persisted result instead of echoing request values.

   The ALTER TABLE may rebuild reaction_users or wait for a metadata lock, so production table size must be checked before rollout. Byte-distinct but visually similar Unicode sequences become independent reactions by design. The schema migration is forward-only because reverting to general_ci can conflict after distinct emoji rows coexist.

   The RED test reproduced utf8mb4_general_ci and the 👍 versus 🎉 collision. After the migration, MySQL reports utf8mb4_bin, the unique index remains channel_id/channel_type/message_id/uid/emoji, both rows persist and sync independently, and cancelling 🎉 leaves 👍 active. The DB helper success and failure paths are covered, with the helper at 100 percent statement coverage.

## Checklist

- [x] Updated documentation

### DIFF
--- a/.octospec/tasks/message-reaction-emoji-binary-collation/brief.md
+++ b/.octospec/tasks/message-reaction-emoji-binary-collation/brief.md
@@ -1,0 +1,61 @@
+---
+type: Task
+title: "Task: message-reaction-emoji-binary-collation"
+description: Make reaction emoji identity byte-exact in MySQL and return the persisted toggle result.
+tags: ["message", "reaction", "mysql", "migration", "wire-contract", "testing"]
+timestamp: 2026-07-22T00:00:00Z
+slug: message-reaction-emoji-binary-collation
+upstream: production reaction collision report
+source: self
+---
+
+# Task: message-reaction-emoji-binary-collation
+
+## User journey
+
+As a user who already reacted with one emoji, I want a different emoji to create
+an independent reaction, so that adding or cancelling it never mutates the first
+reaction and every write response matches the durable server state.
+
+## Root cause
+
+`reaction_users.emoji` inherits `utf8mb4_general_ci`. On the deployed MySQL-compatible
+database that collation compares `👍` and `🎉` as equal. The five-column unique
+index therefore treats the second emoji as a duplicate, and the atomic upsert
+toggles the existing `👍` row while the handler echoes the requested `🎉`.
+
+## Load-bearing changes
+
+- Add a new forward migration that changes only `reaction_users.emoji` to
+  `utf8mb4_bin`; do not edit an already-applied migration or change the table-wide
+  collation.
+- Keep `(channel_id, channel_type, message_id, uid, emoji)` as the unique key.
+- Make `toggleReaction` return the persisted `emoji`, `seq`, and `is_deleted` from
+  its existing post-upsert read.
+- Build the sync command and HTTP response from that persisted result. Treat a
+  requested/persisted emoji mismatch as a storage invariant failure.
+
+## Out of scope
+
+- Unicode normalization or collapsing visually similar but byte-distinct emoji
+  sequences such as `❤` and `❤️`.
+- Reconstructing historical reactions that were never persisted.
+- Changing reaction authorization, visibility, Space isolation, or rate limits.
+- Changing the remaining `reaction_users` column collations.
+
+## Acceptance
+
+- The `emoji` column reports `utf8mb4_bin` after migrations.
+- Adding `👍` and then `🎉` creates two active rows and sync returns both.
+- Clicking `🎉` again changes only the `🎉` row to deleted; `👍` stays active.
+- Write responses and sync commands use the persisted emoji/seq/deleted state.
+- Focused integration and DB-helper tests, `go test ./modules/message`, race,
+  `go vet`, formatting, and migration checks pass.
+
+## Rollout and rollback
+
+- Check the production table size before rollout because a collation change may
+  rebuild the table or hold a metadata lock.
+- Application rollback is compatible with the binary collation. The migration
+  Down path is intentionally a no-op: reverting to `general_ci` can fail once
+  byte-distinct emoji rows coexist under the unique index.
--- a/modules/message/api.go
+++ b/modules/message/api.go
@@ -2162,7 +2162,7 @@ func (m *Message) addOrCancelReaction(c *wkhttp.Context) {
 	}
 	// 多 reaction：按 (uid, message_id, channel, emoji) 原子 toggle。同 emoji 命中唯一键翻转
 	// is_deleted，不同 emoji 落各自独立行（追加）。返回最终 is_deleted 供 Web 乐观更新对账。
-	isDeleted, err := m.messageReactionDB.toggleReaction(&reactionModel{
+	toggleResult, err := m.messageReactionDB.toggleReaction(&reactionModel{
 		ChannelID:   fakeChannelID,
 		ChannelType: req.ChannelType,
 		UID:         loginUID,
@@ -2173,6 +2173,17 @@ func (m *Message) addOrCancelReaction(c *wkhttp.Context) {
 	})
 	if err != nil {
 		m.Error("写入消息回应失败", zap.Error(err), zap.String("channel_id", fakeChannelID), zap.String("message_id", req.MessageID))
+		httperr.ResponseErrorL(c, errcode.ErrMessageStoreFailed, nil, nil)
+		return
+	}
+	if toggleResult.Emoji != req.Emoji {
+		m.Error("reaction 持久化结果与请求 emoji 不一致",
+			zap.String("channel_id", fakeChannelID),
+			zap.String("message_id", req.MessageID),
+			zap.String("uid", loginUID),
+			zap.String("requested_emoji", req.Emoji),
+			zap.String("persisted_emoji", toggleResult.Emoji),
+			zap.Int64("persisted_seq", toggleResult.Seq))
 		httperr.ResponseErrorL(c, errcode.ErrMessageStoreFailed, nil, nil)
 		return
 	}
@@ -2189,8 +2200,8 @@ func (m *Message) addOrCancelReaction(c *wkhttp.Context) {
 			"message_id":   req.MessageID,
 			"channel_id":   req.ChannelID,
 			"channel_type": req.ChannelType,
-			"emoji":        req.Emoji,
-			"seq":          seq,
+			"emoji":        toggleResult.Emoji,
+			"seq":          toggleResult.Seq,
 		},
 	})
 	if err != nil {
@@ -2203,9 +2214,9 @@ func (m *Message) addOrCancelReaction(c *wkhttp.Context) {
 		MessageID:   req.MessageID,
 		ChannelID:   req.ChannelID,
 		ChannelType: req.ChannelType,
-		Emoji:       req.Emoji,
-		Seq:         seq,
-		IsDeleted:   isDeleted,
+		Emoji:       toggleResult.Emoji,
+		Seq:         toggleResult.Seq,
+		IsDeleted:   toggleResult.IsDeleted,
 	})
 }
 
--- a/modules/message/api_reaction_integration_test.go
+++ b/modules/message/api_reaction_integration_test.go
@@ -89,7 +89,7 @@ func queryReactionRow(t *testing.T, ctx *config.Context, messageID int64, channe
 	t.Helper()
 	var model *reactionModel
 	_, err := ctx.DB().Select("*").From("reaction_users").
-		Where("channel_id=? and channel_type=? and message_id=? and uid=? and emoji=?",
+		Where("channel_id=? and channel_type=? and message_id=? and uid=? and BINARY emoji=BINARY ?",
 			channelID, channelType, strconv.FormatInt(messageID, 10), testutil.UID, emoji).
 		Load(&model)
 	require.NoError(t, err)
@@ -103,6 +103,17 @@ func decodeToggleResp(t *testing.T, w *httptest.ResponseRecorder) reactionToggle
 	var resp reactionToggleResp
 	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp), w.Body.String())
 	return resp
+}
+
+func TestReactionEmojiColumnUsesBinaryCollation(t *testing.T) {
+	_, ctx, _ := setupGroupTestData(t)
+
+	var collation string
+	err := ctx.DB().Select("collation_name").From("information_schema.columns").
+		Where("table_schema=database() and table_name='reaction_users' and column_name='emoji'").
+		LoadOne(&collation)
+	require.NoError(t, err)
+	require.Equal(t, "utf8mb4_bin", collation)
 }
 
 func TestReactionAcceptsVisibleMessageAndToggles(t *testing.T) {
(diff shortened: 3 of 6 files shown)


Jerry-Xin (approved):
**APPROVE** — Correct, tightly-scoped fix for the reaction emoji collision. Root cause (a linguistic collation on `reaction_users.emoji` collapsing byte-distinct supplementary-plane emoji under the five-column unique key) is accurately diagnosed, and the fix makes emoji identity byte-exact end to end.

### ✅ Highlights
- **Byte-exact identity is now consistent on both paths.** Changing only the `emoji` column to a binary collation makes the collation a column property, so the upsert's unique-key match *and* the post-upsert read-back comparison in `toggleReaction` (`modules/message/db_message_reaction.go`) are both byte-exact. Write and compare/dedup share the same canonicalization — no add-stores-one-form / remove-looks-up-another asymmetry.
- **Fail-closed identity guard** (`modules/message/api.go:2179`): when the persisted emoji differs from the requested one the handler returns a store-failed error instead of echoing the request value. This surfaces any residual storage-layer drift rather than silently lying to the client, and the sync command + HTTP response are now built from the persisted `emoji`/`seq`/`is_deleted` (`api.go:2200-2219`).
- **Length cap is grapheme-safe.** `maxReactionEmojiRunes = 20` rejects over-long input by rune count *before* the DB write (`api.go:2015`), aligned with the `VARCHAR(20)` character column, so a multi-code-point ZWJ/modifier sequence is never truncated mid-grapheme and no invalid UTF-8 can reach storage. Empty input is also rejected.
- **Charset retained as 4-byte capable**; the identity column keeps `utf8mb4`, only the collation moves to binary. The unique index `(channel_id, channel_type, message_id, uid, emoji)` is unchanged and is rebuilt with binary comparison semantics by the column modification.
- **Scope discipline**: the brief explicitly excludes Unicode normalization, so a skin-tone-modified emoji is correctly a distinct identity and a ZWJ family sequence is correctly one identity (one column value). Back-compat is acknowledged: byte-distinct rows that previously collided become independent by design; historical never-persisted reactions are out of scope.
- **Strong regression coverage**: collation assertion on the live column, independent 👍/🎉 create → sync → cancel with per-row `is_deleted`/`seq` checks, plus DB insert-error and read-back-error paths.

### 💬 Non-blocking notes
- The read-back query in `toggleReaction` still uses plain `emoji=?`; correctness now depends entirely on the column staying binary-collated. This is adequately guarded by the new collation test and the fail-closed identity check, but a short inline comment on that query noting the dependency on the column collation would help future maintainers avoid reintroducing a collation-sensitive comparison.
- The migration `Down` is an intentional no-op (documented: reverting to the linguistic collation could break the unique index once byte-distinct rows coexist). Reasonable for forward-only rollout — worth confirming the deploy runbook flags the pre-rollout table-size / metadata-lock check called out in the brief, since a column modification can rebuild the table.

No blocking issues. Migration is reversible-by-design as a no-op with a clear rationale, storage is 4-byte safe, input is bounded and grapheme-safe, and write/compare canonicalization is consistent.

lml2468 (approved):
# PR Review — PR #638 (Mininglamp-OSS/octo-server)

Reviewed at head `aa366f3b` against merge-base (base `main`).
**Build/vet/test:** `go build ./modules/message/` exit 0 ✅, `go vet` exit 0 ✅. `go test ./modules/message/ -run 'TestReactionEmoji|Toggle'` → **ok** (ran green on this checkout). Full DB-backed suite still needs MySQL in CI.
Classification: data-integrity fix — MySQL collation collision on the reaction identity column + migration + persisted-result wire-contract. Reviewed in **DATA-INTEGRITY** emphasis.

## 1. Spec compliance — ✅
Graded against `.octospec/tasks/message-reaction-emoji-binary-collation/brief.md`.
- **Missing:** none. Forward migration flips **only** `reaction_users.emoji` to `utf8mb4_bin`; the 5-col unique key is retained; `toggleReaction` returns persisted `emoji/seq/is_deleted`; the handler builds CMD + HTTP response from that; requested≠persisted → invariant failure. All present.
- **Over-build:** none. No Unicode normalization (`❤`/`❤️` explicitly out of scope), no other column collations touched, no authz/visibility/isolation/rate-limit change.
- **Deviation:** none. Matches the brief's load-bearing list exactly, including the intentional no-op Down.

## 2. Code quality — Approved

### Findings
No P0/P1/P2 findings.

- **Root cause is correctly diagnosed and fixed.** `utf8mb4_general_ci` collates distinct supplementary-plane emoji (👍 vs 🎉) as **equal**, so the unique key `(channel_id, channel_type, message_id, uid, emoji)` routed a new emoji into another emoji's upsert row — toggling the wrong reaction while the handler echoed the requested one. Switching the identity column to `utf8mb4_bin` makes the comparison byte-exact; MySQL rebuilds the existing unique index with the new semantics automatically. This is the precise, minimal fix.
- **Migration scope is minimal & correct.** Exactly one `MODIFY COLUMN emoji VARCHAR(20) … COLLATE utf8mb4_bin` — no table-wide `CONVERT TO` / `DEFAULT CHARSET` change (grep-confirmed), so no other column's collation shifts. `VARCHAR(20)` length is unchanged from before → no truncation regression (ZWJ-sequence length is a pre-existing concern the brief scopes out).
- **Forward-only migration is the right call and honestly documented.** The Down is a deliberate `SELECT 1` no-op with a clear comment: once byte-distinct emoji rows coexist, reverting to `general_ci` would fail the unique index / need destructive dedup. Application rollback is compatible with the binary collation. Correct rollout reasoning + the production-table-size caveat (collation change may rebuild / hold a metadata lock) is flagged.
- **Persisted-result wire-contract closes the echo bug.** `toggleReaction` now returns the post-upsert **read** of `emoji/seq/is_deleted`, and the handler uses `toggleResult.{Emoji,Seq,IsDeleted}` for **both** the sync CMD and the HTTP response — no longer echoing `req.Emoji`/`seq`. So the client and other viewers always see durable server state.
- **Defense-in-depth invariant check.** `if toggleResult.Emoji != req.Emoji` → `ErrMessageStoreFailed` + a structured (requested vs persisted) log, so any future storage-layer identity drift surfaces loudly instead of silently returning the wrong emoji. Belt-and-suspenders beyond the collation fix.
- **No isolation/authz regression:** the change is confined to the emoji identity + result plumbing; `fakeChannelID` derivation, auth, and the reaction rate/visibility paths are untouched (brief out-of-scope, verified in the diff).

## 3. Test assessment
Strong and genuinely regression-pinning:
- `TestReactionEmojiColumnUsesBinaryCollation` asserts the column reports `utf8mb4_bin` after migrations (locks the schema contract).
- The integration test now reproduces the **real** collision journey: 👍 then 🎉 → **two** independent active rows, then clicking 🎉 again toggles only the 🎉 row (👍 stays active), asserting the persisted toggle result each time. Note the prior test used `❤️`, which is byte-distinct from 👍 *even under general_ci* and so never caught the collision — swapping to 🎉 (a genuine general_ci collision with 👍) is the correct repro. The `test:` commits also add storage-failure coverage.
- All DB-backed (the collation behavior can only be exercised against MySQL) → CI must run `go test ./modules/message` + race + migration checks per the brief's acceptance.

## 4. Coverage / blind spots
- Verified by build/vet + the targeted message tests (green here) + full read of the migration, db layer, and handler + migration-scope grep.
- NOT independently run: the full DB-backed suite against a MySQL-compatible engine that reproduces the general_ci collision (the bug is engine-collation-specific) — CI must confirm on the deployment's actual DB.

## Verdict
**APPROVE** — a precise, minimal data-integrity fix with correct root-cause analysis: the reaction identity column moves to `utf8mb4_bin` so byte-distinct emoji stop colliding under the unique key, the handler now returns persisted (not echoed) emoji/seq/is_deleted with an invariant guard against drift, and the migration is correctly scoped (identity column only) and forward-only with well-reasoned rollback. The test swap to a true general_ci-colliding emoji pair actually reproduces the bug. Both gates pass.
Suggested decision: `Safe to merge` after CI runs the DB suite on the deployment's engine and an operator checks the `reaction_users` table size for the collation-change lock window (per the brief's rollout note). (PR `merge: BLOCKED` — branch-protection.)

mochashanyao (approved):
[Octo-Q · automated review]

---

[Octo-Q] verdict: APPROVE — no P0/P1 issues found. Migration is forward-only and minimal, code correctly surfaces persisted state, test coverage is thorough including the exact collation-colliding emoji pair. Recommend Opus approve.

yujiawei (approved):
# Code Review — PR #638 (octo-server)

**Scope reviewed:** head `aa366f3b`, merge-base `ca72704c`. Files: reaction emoji binary-collation migration, `toggleReaction` persisted-result plumbing, fail-closed mismatch guard, and regression/unit tests. Reviewed with two independent analysis passes plus local build/vet/gofmt and the DB-mock unit suite.

## 1. Spec compliance

The linked brief (`.octospec/tasks/message-reaction-emoji-binary-collation/brief.md`) is met exactly, with no over- or under-build:

- **Migration touches only the identity column.** `MODIFY COLUMN emoji VARCHAR(20) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL DEFAULT ''` preserves the original width (`varchar(20)`), nullability, and default from `20210416000001`. No table-wide collation change; other columns untouched.
- **Five-column unique key kept.** `reaction_users_channel_msg_uid_emoji_uidx` (`20260712000001`) is not altered — MySQL rebuilds it under the new column collation automatically.
- **`toggleReaction` returns the persisted `emoji`/`seq`/`is_deleted`** and the handler builds both the sync CMD and the HTTP response from that result instead of echoing request values.
- **Mismatch is treated as a storage-invariant failure** (fail-closed → `ErrMessageStoreFailed`, with a structured error log).
- **`Down` is an intentional no-op** (`SELECT 1`), matching the forward-only rollout note.
- **Out-of-scope respected:** no Unicode normalization, no historical reconstruction, no authz/visibility/Space-isolation change, no other column collation change.
- Regression coverage (independent 👍/🎉 create → sync → cancel) and DB helper success + both error paths added.

**Spec: ✅**

## 2. Code quality

No blocking (P0/P1) defects were introduced by this diff. `go build`, `go vet`, `gofmt -l`, and the DB-mock unit tests (`TestMessageReactionDB_ToggleReaction*`) pass locally. DB-backed integration tests require live MySQL/Redis and run in CI.

The read-back `WHERE emoji=?` is **correct post-migration**: a bind parameter is coercible, so the comparison adopts the column's `utf8mb4_bin` collation and is byte-exact. The write path (`INSERT … ON DUPLICATE KEY UPDATE is_deleted = 1 - is_deleted, seq = VALUES(seq), …`) correctly advances `seq` on every toggle, so the incremental sync cursor observes each transition — the new `require.Greater(partyCancel.Seq, partyToggle.Seq)` assertion holds.

### Non-blocking notes (P2)

- **P2 — trailing-space exactness is not guaranteed by `utf8mb4_bin`.** `utf8mb4_bin` uses PAD SPACE semantics, so `"👍"` and `"👍 "` still compare equal in the unique index. This is not reachable today: the toggle helper is the *only* writer to `reaction_users` and the handler `strings.TrimSpace`s the emoji before use, and it is not a regression (the prior `general_ci` column was also PAD SPACE). If strict byte-for-byte identity (including trailing whitespace) is ever a hard requirement, `VARBINARY(80)` or an explicit NO PAD collation would be needed. Not required for the stated acceptance criteria.
- **P2 — read-after-write on the same session.** `toggleReaction` does `INSERT` then a separate `SELECT` on one `dbr.Session` without an explicit transaction. This is unchanged from the pre-PR code and safe on the current single-connection setup; it would only matter if read/write splitting were introduced later.
- **P2 — the fail-closed guard cannot roll back the write it detects.** Under normal (migrated) operation the persisted emoji always equals the request, so the guard never fires. It can only trigger if the app runs against an un-migrated (`general_ci`) schema — a window the startup-migration model closes before traffic is served — and even then the durable row has already toggled before the `ErrMessageStoreFailed` is returned. This is strictly an improvement over the prior behavior (which silently echoed request values and masked the drift); worth being aware of, not worth blocking.

## 3. Items to verify manually (security-sensitive)

- **DDL rollout on `reaction_users`.** A collation change may rebuild the table or take an exclusive metadata lock; the migration specifies no `ALGORITHM`/`LOCK`. The brief flags this and the table is effectively empty (reactions are a new feature with no legacy rows per `20260712000001`), so impact should be minimal — but confirm production table size / no long-running transaction before applying, as the brief instructs.
- **Forward-only Down.** Rollback intentionally keeps the binary collation. Confirm this is acceptable operationally (application rollback is compatible; schema rollback is deliberately not offered).

## 4. Suggestions

- Consider mirroring the test's `BINARY emoji=BINARY ?` defensive form (or an explicit `COLLATE utf8mb4_bin`) in the production read-back purely as belt-and-suspenders against any future schema drift. Optional — the current `emoji=?` is correct once the migration is applied.

## 5. Additional observations

- The concurrency characteristics of the upsert + separate read-back (sequence-allocation order vs commit order, and read-back reflecting a concurrent writer's state) are inherent to the existing single-statement-upsert-then-select design and are unchanged by this PR. They are not introduced here and are out of scope for this fix; if stronger per-toggle causal guarantees are ever required, that is a separate design change.

Overall: a tightly-scoped, well-tested fix that does exactly what the linked task requires. Approving; the two operational items above warrant a human eye at rollout given the security-sensitive classification.

PR #639, opened 2026-07-22
feat(incomingwebhook): per-scope webhook quota with per-thread & group-total caps

### LINKED ISSUE #640: ✨ Incoming Webhook 数量限制改为按子区粒度，而非按群
## 需求描述

**反馈人**：李宁
**场景**：一个群有较多子区时，每个子区需要配独立的 Webhook 来推送消息，但目前 Incoming Webhook 的数量上限是按群计算（10 个/群），子区多的群很容易用满。尤其上了 Octo Loop 功能后，这个限制会影响正常使用。

## 具体诉求

> 用户原话：「按照最小单位子区进行数量限制可能会好一些，避免一个群聊创建很多子区的情况」「尤其是上了 octo-loop 功能后，这个限制可能会影响使用 octo-loop 的正常使用」

将 Incoming Webhook 数量限制的粒度从「每个群」改为「每个子区」，每个子区拥有独立的 Webhook 配额。

## 期望效果

- 每个子区有独立的 Webhook 数量上限，不跟群共享配额
- 子区多的群不会因为 Webhook 数量不够而影响使用
- 兼容 Octo Loop 场景下的 Webhook 需求

## 初步方案

当前限制：每群 10 个 Incoming Webhook（群 + 所有子区共享配额）
建议改为：每个子区独立计数，各自有 Webhook 上限

## 📢 反馈记录

- 2026-07-22 user:李宁 (uid:4a0943dcea2242008b821a4db8d47801) via Bug 反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5)

## 🔑 闭环关键词

Incoming Webhook, webhook 限制, webhook 上限, webhook limit, 子区, thread, 数量限制, 配额, quota

---

**原始来源**：DMWork group:Bug 反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5) user:李宁 (uid:4a0943dcea2242008b821a4db8d47801)

an9xyz:
## Summary

Make the Incoming Webhook creation quota **per delivery scope** `(group_no, thread_short_id)` — the group itself and each thread (子区) get **independent** budgets instead of sharing one per-group cap — and add two knobs so operators can **precisely control** the counts:

1. **Per-scope re-scoping** — a group with many threads (and **Octo Loop**, which provisions a webhook per thread) is no longer blocked by the single shared 10-cap.
2. **`max_per_thread`** — a per-thread cap **decoupled** from the group-self `max_per_group`, so you can set e.g. group 10 / each thread 3. Unset → falls back to `max_per_group` (unchanged behavior until configured).
3. **`max_total_per_group`** — a group-wide **aggregate ceiling** across the group and all its threads; `0` = disabled (default). Bounds the `max_per_scope × (threads+1)` growth that per-scope quotas alone allow, since thread creation itself is uncapped.

The `FOR UPDATE` serialization lock deliberately stays on the parent `group` row; only the `COUNT(*)` predicates narrow/add. All three quota layers are evaluated inside that **one** critical section, so the change is race-exact under concurrency (no new lock, no lock-scope change). No schema/data migration.

## Release Note

> **Incoming Webhook quota is now per delivery scope, with two new admin knobs.**
>
> - **Behavior change** — `incomingwebhook.max_per_group` (default 10) and `incomingwebhook.max_per_creator` (default 5) now apply **per delivery scope** (the group itself and *each* thread each get the cap independently) instead of being shared across the group and all its threads. Existing deployments that tuned these values get a **more permissive** effective limit for groups that have threads — review your setting if you relied on the old shared cap.
> - **New** `incomingwebhook.max_per_thread` (env `DM_INCOMINGWEBHOOK_MAX_PER_THREAD`) — per-thread cap, decoupled from `max_per_group`. **Unset → falls back to `max_per_group`**, so no change until you set it.
> - **New** `incomingwebhook.max_total_per_group` (env `DM_INCOMINGWEBHOOK_MAX_TOTAL_PER_GROUP`) — group-wide ceiling across the group + all threads. **Default `0` = disabled**, so no aggregate limit exists unless you opt in.
> - No schema/data migration; existing webhooks are unaffected.

## Related Issue

Closes #640 — 「Incoming Webhook 数量限制改为按子区粒度，而非按群」(反馈人：李宁；Octo Loop 场景下每群 10 个的共享上限成为瓶颈).

## Linked Spec

`.octospec/tasks/incoming-webhook-quota-per-thread/brief.md`

Supersedes the `incoming-webhook-thread` task's locked decision ("threads share the parent group's `max_per_group`; no per-thread quota"), per issue #640.

## Changes

- **`modules/incomingwebhook/db.go`** — `insertWithQuota` now takes `quotaLimits{scope, perCreator, total}`. The scope + per-creator `COUNT(*)` narrow by `thread_short_id` (hitting `idx_incoming_webhook_thread`); a third `total` count (group-wide, no thread filter) enforces the aggregate ceiling when `> 0`. The parent-group `FOR UPDATE` lock is unchanged.
- **`modules/common/system_settings.go` + `system_setting_schema.go`** — new settings `incomingwebhook.max_per_thread` (Positive; falls back to `max_per_group`) and `incomingwebhook.max_total_per_group` (non-Positive, `0`=disabled, range `[0,3650]`, negative→0). `max_per_group` reinterpreted as the group-self scope cap.
- **`modules/incomingwebhook/api.go`** — the create handler picks `max_per_group` for a group webhook or `max_per_thread` for a thread webhook, and passes the aggregate `max_total_per_group`.
- **`pkg/errcode/incomingwebhook.go` + i18n** — new `ErrIncomingWebhookTotalQuotaExceeded` (409 `mgmt_total_quota_exceeded`); the two existing quota messages generalized to "group or thread"; en-US markers + zh-CN regenerated.
- **`modules/incomingwebhook/README.md`** — quota section updated to per-scope + the two knobs.
- **Tests** — DB-level per-scope independence (group↔thread, thread↔thread) + per-creator per-scope + aggregate ceiling (empty scope still blocked, soft-delete frees a slot, `total=0` disables); settings fallback/override/clamp; HTTP group-vs-thread distinct caps; a concurrent aggregate-ceiling test; and the hardened `TestCreate_QuotaConcurrent` scope query.
- **`.octospec/`** — task brief, context, journal, pending learning (lock-vs-count-scope), log.

## Testing

- [x] **Integration tests run locally & green under `-race`** against MySQL 8.0 + Redis + WuKongIM: `go test -race ./modules/common/` (ok) and `go test -race ./modules/incomingwebhook/` (ok), including the concurrent per-scope and aggregate-ceiling tests. Also green in CI (Test job).

Static gates also green: `go build`, `go vet`, `golangci-lint run` (0 issues), `make i18n-extract-check`, `make i18n-lint`.

## COMPREHENSION

   In the `insertWithQuota` transaction, the quota is evaluated as up to three `COUNT(*)`s inside a single `SELECT id FROM group … FOR UPDATE` critical section: (a) the per-delivery-scope count `(group_no, thread_short_id)` vs `max_per_group` (group) or `max_per_thread` (thread); (b) the per-creator count in that scope; (c) an optional group-wide count `(group_no)` vs `max_total_per_group`. Lock scope (whole group) and count scope (per-target / group-wide) are intentionally decoupled: lock coarse, count fine.

   Only the counting changes; both the user face and the bot face funnel through the same `insertWithQuota`, so they move together. The main hazard would be narrowing the `FOR UPDATE` lock to a per-thread range — an empty range takes a pure gap lock and concurrent first-inserts deadlock (InnoDB 1213). This PR avoids it by keeping the parent-group row lock, and adding the aggregate count adds no new lock. Backward-compat: existing rows carry `thread_short_id=''` and `max_per_thread`/`max_total_per_group` default to "same as group" / "disabled", so nothing changes until an operator opts in.

   `TestThread_QuotaIsPerScope` / `TestThreadWebhookQuotaIsPerScope` (independence), `TestTotalQuotaCeiling` (aggregate ceiling blocks even a fresh empty scope; soft-delete frees a slot; `total=0` disables), `TestThread_MaxPerThreadDistinctFromGroup` (group cap 2 vs thread cap 1), and **`TestThread_TotalQuotaCeilingConcurrent`** (12 goroutines across group + 2 threads under a ceiling of 4 → exactly 4 succeed, group total == 4). All pass locally under `-race` against real MySQL/Redis/WuKongIM, and in CI.

## Checklist

- [x] Updated documentation (module README, setting descriptions, octospec journal/log)

https://claude.ai/code/session_01QqpYSJX4M73HBFKmJ1CQkQ


lml2468 (dismissed):
# PR Review — PR #639 (Mininglamp-OSS/octo-server)

Reviewed at head `c4c7964a` against merge-base (base `main`).
**Build/vet/test:** `go build ./modules/incomingwebhook/ ./modules/common/` exit 0 ✅, `go vet` exit 0 ✅. New `TestThreadWebhookQuotaIsPerScope` / `TestCreatorQuotaIsPerScope` are **DB-gated** (`testutil.NewTestServer` → MySQL refused in sandbox) — unrun here; CI must run them.
Classification: rate-limit/quota granularity change on the incoming-webhook creation gate — concurrency (lock vs count scope) + isolation + i18n. Reviewed in **RELIABILITY / CONCURRENCY** emphasis.

## 1. Spec compliance — ✅
Graded against `.octospec/tasks/incoming-webhook-quota-per-thread/brief.md`.
- **Missing:** none. Both quota layers (`max_per_group`, `max_per_creator`) re-scope their `COUNT(*)` by `thread_short_id`; `FOR UPDATE` stays on the parent group row; setting keys reused with doc semantics shifted; no schema/migration. All present.
- **Over-build:** none. No new `max_per_thread` setting (decision A, documented), no new mount (both user + bot faces already funnel through `insertWithQuota`), no lock-scope change.
- **Deviation:** this deliberately supersedes the prior `incoming-webhook-thread` "threads share the parent cap" decision — explicitly per user request and documented in the brief. Legitimate, not a silent divergence.

## 2. Code quality — Approved

### Findings
No P0/P1/P2 findings.

- **The load-bearing concurrency property is preserved and correctly reasoned.** Only the `COUNT(*)` dimension narrows (`AND thread_short_id=?`); the `FOR UPDATE` lock **stays on the parent `group` row**. The code comment (and the `quota-lock-vs-count-scope.md` learning) explain precisely why: narrowing the lock to a `(group_no, thread_short_id)` range would hit **0 rows on an empty-scope first insert → a pure gap lock**, and concurrent gap-X locks are mutually compatible → both pass the count check → insert-intention locks deadlock (1213) with no retry → opaque 500. Locking the always-present parent row serializes same-group creates without a gap lock. This is exactly the #31 lesson, and it's kept. **Count-scope (fine, per-thread) and lock-scope (coarse, per-group) are intentionally decoupled** — correct.
- **Index-backed.** The scoped count `WHERE group_no=? AND thread_short_id=? AND status != ?` lands squarely on the existing composite `idx_incoming_webhook_thread(group_no, thread_short_id, status)` — no full scan introduced.
- **Scope isolation is correct.** `m.ThreadShortID` is bound upstream (validated against the parent group in the create handler), so the count converges to the caller's actual scope: group-self uses `''`, each thread uses its short_id. No cross-scope bleed — filling thread A can't consume group-self's or thread B's budget (pinned by `TestThreadWebhookQuotaIsPerScope`).
- **Per-creator re-scoped consistently (decision B).** Keeping it group-wide while the group cap went per-scope would make the personal 5-cap the new Octo-Loop bottleneck; re-scoping both keeps口径 consistent. Same parent-row lock covers both counts — no extra lock.
- **Quota semantics preserved:** soft-deleted rows free a slot (`status != statusDeleted`), disabled rows still occupy (anti-hoarding), admin creator-cap exemption (`maxPerCreator=0`) unchanged — all carried through verbatim, just scope-narrowed.
- **Settings: defaults unchanged (10/5), only documented semantics shifted** per-group → per-scope, in both the getter comments and the schema `Description` — so operators reading the manager surface see the new meaning; no value surprise. errcode/i18n messages updated to "scope" wording (zh-CN + en-US both).

## 3. Test assessment
Purpose-built and regression-pinning: `TestThreadWebhookQuotaIsPerScope` fills the group-self bucket to cap (asserts `ErrQuotaExceeded`), fills thread A independently (asserts its own cap), then proves thread B still inserts freely — i.e. buckets are independent, the exact behavior the change delivers. `TestCreatorQuotaIsPerScope` does the same for the personal cap. Both are the right shape. **DB-gated** (`insertWithQuota` needs real MySQL for the `FOR UPDATE` + count semantics) → CI must run them; that's the main residual. A concurrency/deadlock regression test (two simultaneous empty-scope inserts must not 1213) would further lock the lock-scope decision, but the parent-row lock makes it correct by construction and the reasoning is documented.

## 4. Coverage / blind spots
- Verified by build/vet + full read of the db-layer diff, index, handler `ThreadShortID` binding, and settings/i18n + the learning note.
- NOT run: the DB-gated quota tests; live concurrency under real MySQL. Recommend the CI DB run.

## Verdict
**APPROVE** — a precise, minimal quota-granularity change that reuses existing thread plumbing: both caps now count per delivery scope `(group_no, thread_short_id)` so a group's threads (and Octo Loop) each get an independent budget, while the `FOR UPDATE` lock deliberately stays on the parent group row to avoid the empty-scope gap-lock 1213 deadlock (the documented #31 lesson, correctly preserved). Index-backed, scope-isolated, defaults unchanged, semantics-shift documented, and covered by per-scope independence tests. Both gates pass.
Suggested decision: `Safe to merge` once CI runs the DB-gated quota suite. (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (dismissed):
**APPROVE** — Re-scoping the incoming-webhook creation quota from per-parent-group to per-delivery-scope `(group_no, thread_short_id)` is correctly and minimally implemented. Verified against the live head.

### Verification of the load-bearing concerns

**Quota key — no collision, empty-scope safe.** The scoped count uses two separate bound predicates on two separate columns (`group_no=? AND thread_short_id=?`), not a concatenated composite key, so there is no delimiter/concatenation collision (group `12`+thread `3` and group `1`+thread `23` are distinct column tuples, not a `12|3` vs `1|23` string clash). The group-self bucket is the stable `thread_short_id=''` value, distinct from any real thread short_id and unchanged from existing rows.

**Atomicity — no check-then-increment race.** The count and insert run in one transaction that first takes `SELECT id FROM group ... FOR UPDATE` on the parent group row (a single existing record lock via the unique group_no index). That lock serializes every concurrent create for the group across *all* scopes, so two concurrent deliveries into the same bucket cannot both pass the count check. The lock is intentionally coarser than the count dimension; narrowing it to the `(group_no, thread_short_id)` range would take a gap lock on an empty scope and reintroduce the InnoDB insert-intention deadlock the original design avoids. Coarse lock + fine count is the right decoupling and is well documented in `modules/incomingwebhook/db.go`.

**Migration — none needed, and correct.** `channel_type` and `thread_short_id` (`NOT NULL DEFAULT ''`) plus the composite index `idx_incoming_webhook_thread(group_no, thread_short_id, status)` already ship in the prior thread migration; the scoped `COUNT(*)` lands squarely on that index. No new schema, no backfill required — existing rows already carry `thread_short_id=''` and fall into the group-self bucket. The prior migration's Down is reversible.

**Tenant / cross-target isolation.** `m.ThreadShortID` is not free request-body input: on the thread mount it comes from the `:short_id` path param, format-validated, then confirmed via a group_no+short_id existence check (cross-group resolves to 404) and an active-status lookup before the row is stamped. So a caller cannot point the quota bucket at a thread outside the path group, and there is no cross-target budget drain.

**Back-compat.** Existing group webhooks counted `WHERE group_no=?`; they now count `WHERE group_no=? AND thread_short_id=''`, which returns the identical set for the group-self bucket. Thread webhooks that formerly shared the group cap now get their own bucket — the intended loosening, not a silent tightening. Quota remains a create-gate, so no in-flight counter is broken. The per-creator cap is re-scoped in the same predicate for consistency, avoiding the personal budget becoming the new bottleneck.

**Windowing / reset.** N/A — the quota is a live `COUNT(*)` create-gate, not a time-windowed counter, so there is no reset boundary to get wrong. Soft-deleted rows are excluded (delete frees a slot); disabled rows still occupy the scope (anti-hoarding) — both preserved per-scope.

### Non-blocking notes
- Settings semantics for `max_per_group` / `max_per_creator` shift from "per group" to "per delivery scope" while keeping the same keys/defaults (10 / 5). Doc comments, admin schema descriptions and the two 409 messages are updated, but operators with existing tuned values will silently get a more permissive effective limit (each thread now gets its own bucket). Worth a release-note callout.
- `db_test.go` covers group↔thread and thread↔thread independence plus per-creator per-scope; the DB-backed assertions only execute in CI (no local DB in the authoring sandbox), which is acceptable here.

### Highlights
- The lock-scope-vs-count-scope decoupling is deliberate and clearly reasoned, preserving the earlier gap-lock deadlock fix.
- Zero-migration change that reuses the existing composite index exactly.
- The behavior reversal (superseding the earlier "threads share the parent group cap" decision) is explicitly documented in the spec/journal with the product rationale.

CI at review time: build/test/vet/lint in progress or green; MERGEABLE, no conflicts.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #639 (octo-server)

## Summary

This PR re-scopes the incoming-webhook creation quota from per-parent-group to per-delivery-scope `(group_no, thread_short_id)`. The group itself (`thread_short_id=''`) and each thread get an independent webhook budget sharing the same `max_per_group` / `max_per_creator` values. The serialization lock deliberately stays on the parent group row while only the `COUNT(*)` predicate narrows — a well-reasoned separation that avoids InnoDB gap-lock deadlocks on empty scopes. No schema migration is needed; existing rows carry `thread_short_id=''` and fall into the group-self bucket.

The implementation is clean, narrowly scoped, and thoroughly documented across code comments, README, i18n strings, schema descriptions, and octospec journal/learning entries.

## Verification

- ✅ **Data-flow trace** — `ThreadShortID` flows from URL param `:short_id` → local `threadShortID` var (`api.go:726-759`) → model construction (`api.go:788`) → `insertWithQuota` (`db.go:57`) → both `COUNT(*)` queries use `m.ThreadShortID` (`db.go:76`, `db.go:96`). Group-level webhooks get `ThreadShortID=""` (Go zero value), matching the `thread_short_id=''` default in the migration.
- ✅ **Lock semantics preserved** — `SELECT id FROM `group` WHERE group_no=? FOR UPDATE` (`db.go:65-68`) remains the serialization point. Unchanged from the pre-PR design; no gap-lock deadlock risk introduced.
- ✅ **Index coverage confirmed** — `idx_incoming_webhook_thread(group_no, thread_short_id, status)` defined in `sql/20260624000001_incomingwebhook_thread.sql:23` covers both new queries with a prefix match.
- ✅ **Backward compatibility** — No data migration required. Existing rows with `thread_short_id=''` (the column default) naturally fall into the group-self scope bucket.
- ✅ **Single production caller** — `insertWithQuota` is only called from `api.go:814` (the `create` handler). No other production path bypasses or duplicates the quota check.
- ✅ **Existing tests unaffected** — All pre-existing tests use `ThreadShortID=""` (Go zero value) and remain internally consistent under per-scope counting.
- ✅ **Dual-path parity (C1)** — Both COUNT queries (scope-level and per-creator) symmetrically add `AND thread_short_id=?`. Delete/soft-delete paths unchanged and still correctly free quota.
- ✅ **No auth/authz changes (C3)** — Quota scope change does not alter who can create webhooks or which scopes they can target.
- ✅ **Distributed side-effects (R3)** — This is a quota/accounting change, not a notification/push/event emission change. No fan-out, no observer issues.

Static analysis only at head `c4c7964a749735f65b03cced1be277be61965946`; build and tests not executed in this environment.

## Findings

### P2 — API-level test gap for per-scope quota independence (`modules/incomingwebhook/api_test.go:1032`)

`TestCreate_QuotaEnforced` (line ~1032) and `TestCreate_QuotaConcurrent` (line ~1054) only create group-level webhooks (no `:short_id` path param). They still pass under per-scope semantics because all webhooks land in the `(groupNo, "")` bucket, but they do not validate the new per-scope independence at the API layer.

The DB-layer tests (`TestThreadWebhookQuotaIsPerScope`, `TestCreatorQuotaIsPerScope`) cover the quota logic directly, so this is not a correctness defect — but an API-level test that fills the group-self quota and then successfully creates a thread-scoped webhook (or vice versa) would guard against future regressions in the path-param → model → quota chain.

**Diff-scope**: pre-existing gap amplified by the new per-scope semantics. The old tests validated group-level quota which still works; the new per-scope dimension lacks API-layer coverage.

### P2 — Stale verification query in concurrent test (`modules/incomingwebhook/api_test.go:1086`)

`TestCreate_QuotaConcurrent` verifies the final DB row count with:

```sql
SELECT count(*) FROM incoming_webhook WHERE group_no=? AND status != 2
```

This counts all non-deleted webhooks regardless of `thread_short_id`. Currently coincidentally correct because all test webhooks are group-level (`thread_short_id=''`). If the test is ever extended to include thread-scoped webhooks, the count would exceed `cap` even though per-scope quota is correctly enforced. Consider adding `AND thread_short_id=''` to match the production query's scope filter, or adding a comment explaining the intentional broader count.

**Diff-scope**: pre-existing query now semantically diverges from the production path it's meant to verify.

### Nit — Error message generality (`pkg/errcode/incomingwebhook.go:155,166`)

The new default messages "At most {{.max}} webhooks are allowed in this group or thread" and "You can create at most {{.max}} webhooks in this group or thread" use "this group or thread" which is slightly ambiguous when the actual scope is a specific thread. The message doesn't tell the caller which thread's quota was hit. Minor — the 409 status code and error ID are sufficient for programmatic handling.

- **C1 — Dual-path parity**: ✅ Clear. Both COUNT queries symmetrically add `thread_short_id=?`. Soft-delete / disable paths unchanged and consistent.
- **C2 — Control-flow ordering**: ✅ Clear. No new nesting or ordering changes. Two sequential COUNTs within the same transaction.
- **C3 — Auth boundary**: ✅ Clear. No auth/authz changes. Quota scope change doesn't alter who can create webhooks.
- **C4 — Auth lifecycle**: N/A. No auth lifecycle changes.
- **C5 — Build ≠ runtime**: N/A. No build/packaging changes.
- **C6 — Governance/docs consistency**: ✅ Clear. README, code comments, i18n strings, schema descriptions, and octospec journal all consistently reflect per-scope semantics.

## Cross-round blocker re-check (R6)

## Data-flow trace

| Consumed data | Upstream source | Reaches consumption point? |
|---|---|---|
| `m.ThreadShortID` in scope COUNT (`db.go:76`) | URL param `:short_id` → `threadShortID` local → model `ThreadShortID` field (`api.go:726-788`) | ✅ Yes. Empty string for group-level, validated short_id for thread-level. |
| `m.ThreadShortID` in creator COUNT (`db.go:96`) | Same source as above | ✅ Yes. Same model field passed to both queries. |
| `m.GroupNo` in FOR UPDATE (`db.go:67`) | URL param `:group_no` → validated by `requireActiveGroup` (`api.go:710-722`) | ✅ Yes. Group existence and Normal status verified before transaction. |
| `max` (scope cap) | `settings.IncomingWebhookMaxPerGroup()` (`api.go:809`) | ✅ Yes. DB → env → default 10, with ≤0 fallback to default. |
| `maxPerCreator` (creator cap) | `settings.IncomingWebhookMaxPerCreator()` for non-admins; 0 for admins (`api.go:810-812`) | ✅ Yes. 0 short-circuits the creator check inside `insertWithQuota`. |

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Well-executed quota re-scoping with correct data flow, preserved lock semantics, confirmed index coverage, and thorough DB-layer tests. The two P2 items are test-coverage enhancements, not correctness defects.

yujiawei (dismissed):
# Code Review — PR #639 (octo-server)

**Head SHA reviewed:** `c4c7964a749735f65b03cced1be277be61965946`
**Base / merge-base:** `main` @ `ca72704c`

Independent review with two additional adversarial reviewer passes (line-level + long-context/architecture). This PR is security-sensitive; the analysis below reflects that.

---

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly the stated goal — move the incoming-webhook **creation quota** from *per parent group* to *per delivery scope* `(group_no, thread_short_id)` — and nothing more.

- **Under-build:** none. Both quota layers are re-scoped in `insertWithQuota` (`modules/incomingwebhook/db.go:76` all-users count, `:94` per-creator count) by adding `AND thread_short_id=?`. The all-users cap and the per-creator cap both narrow, as specified.
- **Over-build:** none. No new system setting, no new schema, no new route. `max_per_group` / `max_per_creator` keep their keys and defaults (10 / 5) and are only reinterpreted per-scope; the DB migration adding `channel_type` / `thread_short_id` / `idx_incoming_webhook_thread` already shipped separately, so no migration is added here.
- **Deviation:** none. The `FOR UPDATE` serialization lock deliberately stays on the always-present parent `group` row (`db.go:66`) rather than narrowing to the per-thread range — this is correct (narrowing it would take a pure gap lock on an empty scope and deadlock concurrent first-inserts). Lock coarse, count fine — the two concerns are correctly decoupled.

This intentionally supersedes the earlier locked decision ("thread webhooks share the parent group cap"), which is explicitly documented as a product-driven reversal. That is a legitimate spec change, not a defect.

---

## 2. Code quality

**Quality: Approved**

Verified independently:

- **Single create path.** Production inserts into `incoming_webhook` funnel through exactly one path (`insertWithQuota`), called only from `create` (`api.go:814`). Both the user face (`/v1/groups/:group_no/...`) and the bot face (`/v1/bot/groups/:group_no/...`) mount the same handler via `MountManagementRoutes`, so per-scope quota applies uniformly to user- and bot-created webhooks.
- **`thread_short_id` is trusted, not free input.** It is stamped from the row (`m.ThreadShortID`), which was validated at create time against the path `group_no` (thread must exist, be active, and belong to this group). It is immutable after creation — not in the `update` handler's allowed fields — so a webhook cannot be migrated into another scope to dodge quota.
- **Backward compatibility is exact.** Existing rows carry `thread_short_id=''` (schema is `VARCHAR(32) NOT NULL DEFAULT ''`), so they fall into the group-self bucket and count exactly what `WHERE group_no=?` counted before. No behavior change for existing group webhooks.
- **Index usage.** The count predicate's two leading equality columns `(group_no, thread_short_id)` match the `idx_incoming_webhook_thread(group_no, thread_short_id, status)` prefix, giving a tight range scan.
- **Tests.** `TestThreadWebhookQuotaIsPerScope` proves group-self saturation → 409, thread-A independence + saturation → 409, and thread-B independence. `TestCreatorQuotaIsPerScope` proves the per-creator cap is per-scope. Build / vet / test-compile all clean locally; DB-backed assertions run in CI (MySQL/Redis unavailable in the authoring sandbox).

### Findings

**P2 — Per-scope quota is multiplicative; aggregate webhooks per group are no longer bounded.** `db.go:76`
The maximum webhook count per group becomes `max_per_group × (N_threads + 1)`. Thread creation itself has no per-group cap (`modules/thread/service.go:159` `CreateThread` only checks active membership) and no rate limit, so a group member can create many threads and then fill each with webhooks, growing thread rows, IM channels, live webhook secrets, and webhook rows well past the old flat cap. This surfaced in all three review passes.
*Why non-blocking:* the multiplicative scaling is the explicit product intent (Octo Loop provisions a webhook per thread), and this diff touches **zero** lines of thread creation — the unbounded-thread enabler is pre-existing and outside this change. Blast radius is bounded by required group membership and per-webhook push rate limits. **Recommend a human accept this resource-bounds tradeoff explicitly**, and consider a follow-up per-group thread cap or aggregate webhook ceiling if abuse is a concern.

**P2 — Deleting a thread does not cascade-disable its webhooks → zombie rows.** `modules/thread/service.go:758` (`DeleteThread`)
Thread deletion soft-deletes the thread and disables the IM channel but does not disable or delete the thread-scoped webhooks (no thread→webhook cascade exists; only group-disband cascades via `disableByGroupNo`). Because disabled rows deliberately still occupy quota, high thread churn leaves permanently orphaned webhook rows that consume storage and their scope's budget forever. Non-blocking storage-hygiene issue, orthogonal to this quota change, but worth a follow-up (mirror the disband cascade on thread delete).

**P2 — Create-time thread-liveness check is TOCTOU with thread deletion.** `api.go:749` vs `db.go:66`
The active-thread check is a non-transactional read; the insert transaction only locks the parent group, not the thread. A thread deleted in the window between check and insert yields an enabled webhook bound to a dead channel. Narrow, and consistent with the existing push-path design (push derives the target from the row and does not re-check thread liveness), so the orphaned webhook simply never delivers. Pre-existing design, not introduced here.

**P2 (nit) — `creator_uid` is not in the index prefix.** `db.go:94`
The per-creator count filters `creator_uid` outside `idx_incoming_webhook_thread`. Fine at the default cap (row count per scope is tiny, ICP filters cheaply); could matter only if `max_per_group` is configured very high.

**P2 (nit) — Per-creator test could be tightened.** `db_test.go` `TestCreatorQuotaIsPerScope`
It proves a creator maxed in the group-self scope can still create in a thread, but does not assert a *second* insert into that same thread is rejected, nor cross-thread creator independence. A regression that accidentally exempted non-empty `thread_short_id` from the per-creator cap would still pass. Consider adding the saturation assertion.

---

## 3. Overall verdict

**APPROVED**

Spec ✅ and Code Quality Approved — no P0/P1. The core change is minimal, correct, race-safe (coarse parent-group lock + fine per-scope count), backward-compatible, and well-tested. All findings are P2/advisory.

### For a human to confirm (security-sensitive)
1. **Accept the resource-bounds tradeoff:** per-scope quota makes aggregate webhooks-per-group `max_per_group × (threads+1)` while thread creation is uncapped. Intended for Octo Loop, but please acknowledge explicitly and decide whether a per-group thread cap / aggregate ceiling is wanted as a follow-up.
2. **Zombie webhooks on thread delete** (no cascade) is a real storage-hygiene gap — track as a follow-up.

### Reviewer coverage
Both adversarial passes ran and were absorbed (line-level pass + long-context/architecture pass — the latter via curl gateway fallback). Consensus across all three passes on the multiplicative-quota concern; one cross-family P1 (legacy `NULL` `thread_short_id` excluded from the count) was **refuted** against the schema (`NOT NULL DEFAULT ''`, so legacy rows are `''`, not NULL). Not verifiable from static review: runtime `EXPLAIN` plans and the CI-only DB integration assertions.

lml2468 (dismissed):
# PR Review (re-review) — PR #639 (Mininglamp-OSS/octo-server)

Re-reviewed at **new head `6485b0101`** against merge-base `4b8cd2d3` (base `main`). Re-affirms my APPROVE at `c4c7964a`.
**Build/vet/test:** `go build ./modules/incomingwebhook/ ./modules/common/` exit 0 ✅, `go vet` ✅. Quota tests remain DB-gated (MySQL) → CI.

## What changed since `c4c7964a` (delta reviewed)
Two non-production commits: `3bcce3b1` (docs — link brief/journal to issue #640) and `6485b0101` (test — harden per-scope quota coverage). **The incomingwebhook production code is byte-identical** to the head I approved: `git diff c4c7964a 6485b0101 -- modules/incomingwebhook/db.go modules/incomingwebhook/api.go` = **0 lines** (settings/errcode also unchanged). (The `modules/message/*` entries in the range-diff are the #638 reaction PR pulled in by rebase, not part of this PR's own scope.)

### Test hardening — addresses mochashanyao's P2 ✅
New `TestThread_QuotaIsPerScope` (`api_thread_test.go`) is the **API-layer** end-to-end test the prior round asked for: with `MAX_PER_GROUP=1`, it drives real HTTP creates and asserts the group-self bucket fills at cap (2nd → 409) while thread A still creates (independent bucket), thread A fills independently (2nd → 409), and thread B still creates — i.e. it exercises the full `path-param → model → quota` chain, not just the db layer. This closes the "no API-layer independence test" gap; the underlying `TestThreadWebhookQuotaIsPerScope` (db layer) is retained.

## On the peer notes at intermediate heads
- **mochashanyao 🟡 "API-layer per-scope test missing"** — now added (`TestThread_QuotaIsPerScope`). Resolved.
- **yujiawei 🟡 "group's total webhook count can grow unbounded as threads multiply; no per-group ceiling on threads"** — a legitimate design observation, but **out of scope and by design**: the whole point of this PR (per user request, brief-documented) is to *stop* the per-group total from capping Octo Loop, which provisions a webhook per thread. The ceiling now lives at the thread-creation layer, not here; a global per-group ceiling across all scopes would re-create the exact bottleneck this removes. Not a defect in this change — a product decision the brief owns. Worth a capacity/monitoring note operationally, not a code blocker.

## Carry-over (unchanged, re-verified byte-identical)
The load-bearing correctness holds as approved: `COUNT(*)` narrows by `thread_short_id` (index-backed on `idx_incoming_webhook_thread`), the `FOR UPDATE` lock **stays on the parent group row** to avoid the empty-scope gap-lock 1213 deadlock (documented #31 lesson), scope isolation via upstream-validated `ThreadShortID` (no cross-scope/cross-group bleed), per-creator re-scoped consistently, soft-delete/disabled quota semantics preserved, defaults unchanged (10/5) with only documented semantics shifted, no schema/migration.

## Verdict
**APPROVE** (re-affirmed at `6485b0101`) — the production change is byte-identical to my prior approval and remains correct (per-scope count, parent-row lock, index-backed, scope-isolated); the only delta since is a new API-layer per-scope independence test (closing mochashanyao's P2) plus a docs link. yujiawei's unbounded-growth note is the intended design, owned by the brief and belonging to the thread-creation layer, not a defect here. Both gates pass.
Suggested decision: `Safe to merge` once CI runs the DB-gated quota suite. Add a release note that `max_per_group`/`max_per_creator` now mean per-delivery-scope, and consider an operational capacity check for very-many-thread groups (yujiawei's note). (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (dismissed):
## Re-review at head `6485b010` — APPROVE (delta is test-only)

Re-verified the true delta since my prior approval by byte-isolating the range against the previously-approved head (the branch was rebased, so the API `compare` list is bloated with unrelated `modules/message` reaction work and a reaction migration from another PR — those are **not** part of this PR and were excluded).

### True delta since prior approval
Every production file in this PR's scope is **byte-identical** to the approved head (verified by blob hash):

- `modules/incomingwebhook/db.go` — identical (load-bearing quota logic unchanged)
- `modules/incomingwebhook/api.go`, `api_i18n.go`, `README.md` — identical
- `modules/common/system_setting_schema.go`, `system_settings.go` — identical
- `pkg/errcode/incomingwebhook.go`, `pkg/i18n/locales/active.zh-CN.toml`, `tools/i18nmarkers/server/active.en-US.toml` — identical
- No new migration in the incomingwebhook scope.

The only changes are **test hardening** across three files:

1. `modules/incomingwebhook/api_thread_test.go` — new `TestThread_QuotaIsPerScope` (the API-layer independence test requested in prior review). It caps each scope at 1 and drives the handler path: fills the group-self bucket (2nd create → `409`), confirms thread A still creatable (independent bucket) then fills it (2nd → `409`), confirms thread B still creatable. This is exactly the group-self vs thread-A vs thread-B independence assertion at the handler layer that was missing.
2. `modules/incomingwebhook/db_test.go` — `TestCreatorQuotaIsPerScope` strengthened: adds a negative case proving the per-creator cap **actually applies within** a thread scope (guards against a non-empty thread scope wrongly exempting the per-creator check), plus cross-thread independence (thread A full does not block the same creator in thread B).
3. `modules/incomingwebhook/api_test.go` — the concurrent-quota DB assertion now scopes its count query to the group-self scope (`thread_short_id=''`), matching the production count semantics so a future co-located thread-scope create cannot cause a false "quota leaked" failure.

### Correctness properties — still intact
Because `db.go` is unchanged, the previously-verified invariants hold unchanged: the `FOR UPDATE` lock stays on the parent group row (deliberately not narrowed to a `(group_no, thread_short_id)` range to avoid the empty-scope gap-lock deadlock regression), the effective-count check narrows by `thread_short_id`, the count is index-backed, and there is no cross-scope leak. No lock-scope change, no new quota-key construction, no config default change (still 10/5), no new migration.

### Build / verification
- `go build ./modules/incomingwebhook/...` — clean
- `go vet ./modules/incomingwebhook/...` — clean
- test package compiles cleanly (DB-gated integration tests require MySQL; not run here — acceptable infra gating)

The prior non-blocking gap (no API-layer per-scope independence test) is now closed. No new review surface introduced by the delta. Approving.

yujiawei (dismissed):
# Code Review — PR #639 (octo-server)

Reviewed at head `6485b0101de5636b9f117c6cdf71572a3f62d4bf` (merge-base `4b8cd2d3`). This is a re-review at a new head; the prior approval was dismissed by the re-push and does not carry over — findings below are for the current head.

## 1. Spec compliance

**Spec: ✅**

The diff moves the Incoming Webhook creation quota from per-parent-group to per-delivery-scope `(group_no, thread_short_id)`, exactly as stated (closes #640, supersedes the earlier "thread webhooks share the parent group's `max_per_group`" decision).

- **No under-build:** both count predicates in `insertWithQuota` (`modules/incomingwebhook/db.go:74` all-users `max_per_group`, `:88` per-creator `max_per_creator`) gained `AND thread_short_id=?`. Error codes / i18n (zh-CN + en-US markers), setting descriptions (`system_settings.go`, `system_setting_schema.go`), README and code comments are all aligned to per-scope.
- **No over-build:** no new setting, no schema change, no data migration. `max_per_group` / `max_per_creator` keep their keys and defaults (10 / 5), reinterpreted per scope. Existing rows carry `thread_short_id=''`, so the group-self bucket counts exactly what `WHERE group_no=?` counted before — byte-for-byte backward compatible.
- **No deviation:** the `FOR UPDATE` serialization lock deliberately stays on the always-present parent `group` row (`db.go:66`) and is **not** narrowed to the per-thread range. This is correct and load-bearing — narrowing the lock to an empty `(group_no, thread_short_id)` range would take a pure gap lock on first insert, and concurrent creators would deadlock (InnoDB 1213) with no retry. Lock coarse, count fine, the two scopes decoupled — as documented in the code.

## 2. Code Quality

**Quality: Approved**

Verified at this head:

- **Single create path:** `insertWithQuota` has exactly one caller (`api.go:814`); both the user face (`/v1/groups/:group_no/...`) and bot face mount the same handler, so they move together.
- **Trusted scope key:** `m.ThreadShortID` is bound from the validated path `:short_id` (existence + active + parent-group ownership checked at `api.go:743-759`) and never appears in any update allowed-field set, so it is immutable post-create — no scope-migration bypass.
- **Index coverage:** the count hits `idx_incoming_webhook_thread(group_no, thread_short_id, status)` (migration `20260624000001`).
- **Schema safety:** `thread_short_id VARCHAR(32) NOT NULL DEFAULT ''` — NULL is impossible, so `thread_short_id=''` cannot undercount legacy rows.
- **Test hardening (this revision's delta):** `TestThreadWebhookQuotaIsPerScope` now proves group↔thread and thread↔thread independence; `TestCreatorQuotaIsPerScope` adds the same-thread 2nd-insert rejection assertion (guards against accidentally exempting non-empty `thread_short_id` from the per-creator check); the new E2E `TestThread_QuotaIsPerScope` covers the path-param → model → quota chain at the API layer; and `TestCreate_QuotaConcurrent`'s post-condition query was tightened to `thread_short_id=''` so it can't misreport a quota leak once thread webhooks share the table.
- Static gates (`go build`, `go vet`, test-compile) pass locally. DB-backed assertions run in CI (MySQL/Redis/WuKongIM required).

### Non-blocking observations (P2 — for human awareness, not merge blockers)

1. **Per-scope quota is multiplicative.** Aggregate webhooks per group is now `max_per_group × (N_threads + 1)`, and thread creation itself has no per-group cap. This is the explicit, product-intended effect (Octo Loop provisions a webhook per thread). The amplification *enabler* — unbounded thread creation — lives entirely outside this diff and is bounded in practice by required group membership and per-webhook push rate limits. Worth an explicit human acknowledgement of the tradeoff; an optional follow-up could add a soft cap on total thread-scoped webhooks per group. Does not block this PR.
2. **Disabled/orphaned thread webhooks still occupy scope quota.** `DeleteThread` does not cascade-disable thread-scoped webhooks (only group disband cascades). High thread churn can leave disabled rows holding quota until manually deleted. Storage hygiene, orthogonal to this change, pre-existing.

### For human verification (security-sensitive)

- Confirm in CI that the DB-backed per-scope tests pass against real MySQL at the repo's isolation level (the deadlock-avoidance rationale for keeping the lock on the parent group row only manifests under actual InnoDB gap-lock behavior).
- Confirm product sign-off on the multiplicative aggregate ceiling (observation 1).

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. No P0/P1. The load-bearing lock-vs-count decoupling is correct, backward compatibility is exact, and the only change since the prior revision is additive test coverage that strengthens the per-scope guarantees.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #639 (octo-server)

## Summary

This PR scopes the incoming webhook quota from a single per-group pool to
independent per-delivery-target pools keyed by `(group_no, thread_short_id)`.
The group itself (`thread_short_id=''`) and each thread each get their own
`max_per_group` and `max_per_creator` buckets, sharing the same configured cap
values. The serialization lock remains on the parent group row — coarser than
the per-scope count — to avoid the gap-lock deadlock that per-scope `FOR
UPDATE` would trigger on empty scopes. This is a deliberate and well-documented
design choice. The change is clean, well-tested, and correctly propagates the
new scope through all layers.

## Verification

- ✅ **SQL count queries scoped correctly** — both the group-level count
  (`db.go:76`) and the per-creator count (`db.go:96`) now filter by
  `group_no=? AND thread_short_id=? AND status != ?`. Both `.Load(&count)`
  calls properly check `err` before using the count value.
- ✅ **ThreadShortID data flow is clean** — in the create handler
  (`api.go:726–759`), `threadShortID` starts as `""`, is set to `shortID` only
  after `IsValidShortID` + `ExistThread` + `GetThread` validation, then flows
  into the model at `:788` and reaches `insertWithQuota` at `:814` with the
  correct value for both group-scope (`""`) and thread-scope paths.
- ✅ **Index coverage** — the existing
  `idx_incoming_webhook_thread(group_no, thread_short_id, status)` index
  (migration `20260624000001`) covers both new count queries via its leading
  columns.
- ✅ **Lock strategy preserved** — `FOR UPDATE` still locks the parent group
  row (`db.go:68–71`), serializing all same-group creates regardless of
  thread scope. This avoids gap-lock deadlocks on empty scopes; documented in
  the extensive block comment at `:35–56`.
- ✅ **Cascade paths unaffected** — `disableByGroupNo` (`db.go:304`) filters
  only by `group_no`, correctly cascading to all webhooks across all scopes
  when a group disbands.
- ✅ **Management scope isolation intact** — `queryManageable` (`api.go:668`)
  enforces `m.ThreadShortID != c.Param("short_id")`, preventing cross-scope
  access for update/delete/regenerate operations.
- ✅ **Error sentinels use errors.Is** — `ErrQuotaExceeded` and
  `ErrCreatorQuotaExceeded` error message strings changed, but all callers
  compare via `errors.Is()` against the sentinel variables, not string
  matching. No breakage.
- ✅ **i18n messages aligned** — zh-CN and en-US both updated to "group or
  thread" wording. Schema descriptions also updated.

Static analysis only at head `6485b0101de5636b9f117c6cdf71572a3f62d4bf`; build

## Findings

No P0/P1 issues. One P2 operational consideration below.

### P2 — Per-scope quotas multiply effective creator capacity (`modules/incomingwebhook/db.go:96`)

With both group-level and per-creator quotas now scoped per
`(group_no, thread_short_id)`, a single non-admin creator can hold up to
`max_per_creator` webhooks in each independent scope. In a group with N active
threads, one creator could own up to `max_per_creator * (1 + N)` webhooks total
across all scopes. With defaults (`max_per_creator=5`) and, say, 20 threads,
that's 105 webhooks for one creator in one group.

This is clearly an intentional design choice — the PR description and tests
explicitly validate this independence, and the group-level `max_per_group` cap
still bounds total webhooks per scope. Administrators setting `max_per_creator`
should be aware that the effective per-creator ceiling scales with thread
count. Not a merge blocker — flagging for documentation/operational awareness.

- `TestThreadWebhookQuotaIsPerScope` (`db_test.go:255-292`): fills group-self
  bucket to `max`, proves thread-A and thread-B each get independent buckets.
- `TestCreatorQuotaIsPerScope` (`db_test.go:296-335`): fills per-creator cap in
  group-self, proves independent per-creator caps in thread-A and thread-B, and
  verifies thread-A's cap actually blocks (not just a no-op on non-empty
  `thread_short_id`).
- `TestThread_QuotaIsPerScope` (`api_thread_test.go:141-172`): end-to-end API
  test with `DM_INCOMINGWEBHOOK_MAX_PER_GROUP=1`, validates isolation across
  group-self, thread-A, and thread-B paths.
- `TestCreate_QuotaConcurrent` (`api_test.go:1087`): count assertion updated to
  include `thread_short_id=''` — correctly scoped to group-self, preventing
  future false positives if thread-scope webhooks coexist.
- `TestSoftDelete_FreesQuota` (`db_test.go:82`): still valid — operates on
  group-scope webhooks (default `thread_short_id=''`), soft-delete correctly
  frees the group-self bucket.
- `disableByGroupNo` cascade coverage test (`db_test.go:337-352`): verifies
  thread webhooks are disabled alongside group webhooks during group disband.
- C1 (dual-path parity): group-scope and thread-scope create paths both route
  through the same `insertWithQuota` with correctly set `ThreadShortID`.
- C2 (control-flow ordering): no new control flow or nested reuse.
- C3 (authorization boundary): no new endpoints or authorization changes;
  thread validation reuses existing `ExistThread`/`GetThread` checks.
- C4 (authorization lifecycle): group disband cascade (`disableByGroupNo`)
  unaffected — operates on `group_no` only, covering all scopes.
- C5 (build/runtime): N/A — no build, packaging, or deployment changes.
- C6 (governance/docs): N/A — `.octospec/` docs are supplementary design
  notes, not governance policy.

## Verdict: APPROVED

Well-designed, well-tested scope refinement. The per-scope quota correctly
isolates group-self and thread webhooks while preserving the intentional
coarser lock granularity. Test coverage spans DB layer, API layer, and
concurrency. No correctness, security, or build concerns.

Jerry-Xin (dismissed):
## APPROVE — re-review at head `1dc621e4`

Since my prior approval at `6485b010`, the head advanced by one commit that **does change production code** (not the earlier tests+docs-only delta). This commit turns the previously-shared per-scope cap into two explicit, decoupled knobs plus a new group-wide aggregate ceiling. I re-verified the load-bearing concurrency/correctness properties byte-by-byte; they are intact.

### True delta `6485b010 → 1dc621e4` (byte-verified)
One commit: *"add max_per_thread + max_total_per_group quota knobs"*. Production files that actually changed (blob-hash confirmed): `modules/incomingwebhook/db.go`, `modules/incomingwebhook/api.go`, `modules/incomingwebhook/api_i18n.go`, `modules/common/system_settings.go`, `modules/common/system_setting_schema.go`, `pkg/errcode/incomingwebhook.go`, plus the two i18n locale files and tests. No migration.

### What the commit does
- **`max_per_thread`** (`modules/common/system_settings.go`): new scope knob for threads, **defaulting to `IncomingWebhookMaxPerGroup()`** when unset — so unconfigured behavior is byte-identical to the prior "group and threads share `max_per_group`" model. Backward compatible. Read-side clamps `<=0` back to the dynamic default (no accidental "silent lockout" of threads).
- **`max_total_per_group`** (default `0` = disabled): opt-in aggregate ceiling across the group body + all threads. Read-side clamps negative to `0`; `0` is a meaningful "off" sentinel, so no work is done when unconfigured.
- **`api.go create()`**: scope cap now selects `MaxPerThread()` when `threadShortID != ""`, else `MaxPerGroup()`. Error routing switched to a `switch` covering the new `ErrTotalQuotaExceeded → mgmtTotalQuotaExceeded(maxTotal)` with a `default` still logging + generic-failure. Each error path carries the correct limit value.
- **`db.go insertWithQuota`**: signature refactored to a `quotaLimits{scope, perCreator, total}` struct; the extra `total` check is a single additional COUNT inside the *same* transaction.

### Load-bearing properties re-verified intact
- **Lock scope unchanged**: still `SELECT id FROM group WHERE group_no=? FOR UPDATE` on the parent group row. It is *not* narrowed to `(group_no, thread_short_id)` — the deliberate coarse lock that avoids the empty-scope-first-insert gap-lock deadlock (1213) from the earlier module review. Comment preserved.
- **Scoped counts still narrow by thread**: both the scope COUNT and the per-creator COUNT still filter on `thread_short_id`, excluding soft-deleted rows. Unchanged.
- **New aggregate count is group-wide (no thread filter)** — correct for a cross-scope ceiling. Because it runs inside the same parent-group lock, a plain read is exact; hits the `group_no` prefix of `idx_incoming_webhook_thread`. Only executed when `total > 0`.
- **No cross-scope leak / no new lock**: all three checks share the one parent-group row lock; `thread_short_id` remains validated against the parent group upstream. Defaults `10/5` unchanged; new ceiling default `0`.

### Tests / build
- `TestThread_MaxPerThreadDistinctFromGroup` proves the group cap (2) and thread cap (1) are independent.
- `TestThread_TotalQuotaCeilingConcurrent`: 12-way concurrent fanout across group + 2 threads with `max_total_per_group=K` asserts **exactly K** succeed *and* the group-wide row count equals K — proving the aggregate quota does not leak under concurrency (the whole point of keeping the coarse lock).
- New setting-validation and error/i18n/schema entries wired consistently; no dangling references to the old 2-arg signature.
- `go build` and `go vet` on the affected packages: green. DB-backed tests are gated on a local MySQL and were not run here; CI Test/Vet/Build are the authority.

### Non-blocking note
The three limit-carrying error responses each use the matching limit (`maxWH` / `maxPerCreator` / `maxTotal`), which is correct. No action needed.

Clean extension with backward-compatible defaults and solid concurrency coverage. Approving at `1dc621e4`.

lml2468 (dismissed):
# PR Review (re-review) — PR #639 (Mininglamp-OSS/octo-server)

Re-reviewed at **new head `1dc621e4`** against merge-base (base `main`). Supersedes my APPROVE at `6485b0101`.
**Build/vet/test:** `go build ./modules/incomingwebhook/ ./modules/common/` exit 0 ✅, `go vet` ✅. New quota + settings tests are **DB-gated** (`testutil.NewTestServer`/settings upsert → MySQL) — unrun here; CI must run them.
Classification: adds a third quota layer (aggregate per-group ceiling) + a decoupled per-thread cap — concurrency (same lock, extra count) + config. Reviewed in **RELIABILITY / CONCURRENCY** emphasis.

## What changed since `6485b0101` (delta reviewed)
One substantive commit — `1dc621e4` — that turns the previously out-of-scope concern into a real feature. This directly answers **yujiawei's prior 🟡** ("group's total webhook count can grow unbounded as threads multiply"): it adds `max_total_per_group` (aggregate ceiling across group-self + all threads) and a distinct `max_per_thread` cap. The base per-scope change I already approved is unchanged.

## New feature review — correct

### The aggregate ceiling reuses the same lock (no gap-lock regression) ✅
`insertWithQuota` now takes a `quotaLimits{scope, perCreator, total}` and evaluates a **third** count — `SELECT count(*) … WHERE group_no=? AND status != ?` (no thread filter) — for the whole-group total. Critically, this runs **inside the same parent-group `FOR UPDATE` critical section** as the other two counts: no new lock, no narrowing, so the #31 empty-scope gap-lock 1213 deadlock is not reintroduced. "Lock coarse (parent group), count fine (scope / creator / whole-group)" is preserved. Uses the `group_no` prefix of the existing composite index.

### Defaults are back-compatible ✅
- `max_total_per_group` defaults to **0 = disabled** (`v < 0 → 0` clamp; schema is deliberately non-`Positive` to allow 0), so **no new ceiling exists unless an operator sets it** — a pure superset of my approved head.
- `max_per_thread` defaults to `IncomingWebhookMaxPerGroup()` (10) when unset (env→DB→fallback), and `v <= 0 → def`, so it can never be 0 → can't accidentally reject all thread creates. Behavior is unchanged from `6485b0101` until configured.
- Scope value (`maxWH`) is always positive (both `MaxPerGroup`/`MaxPerThread` clamp `<=0` to positive defaults) → no "scope=0 rejects everything" footgun.

### Wire/error surface is complete ✅
`ErrTotalQuotaExceeded` → `mgmtTotalQuotaExceeded` → `ErrIncomingWebhookTotalQuotaExceeded` **409 Conflict** with a `{{.max}}` i18n param (zh-CN + en-US). Both new keys registered in `systemSettingSchema` (manager-writable, no migration): the two per-scope caps `Positive: true`, the total non-Positive with a documented [0,3650] range.

### Tests — strong, including a real concurrency proof ✅
`TestThread_TotalQuotaCeilingConcurrent` fires concurrent creates across the group + multiple threads with `max_total_per_group=K` and asserts **exactly K succeed and the group total == K** — a genuine race test that proves the aggregate ceiling holds under the parent-group-lock serialization (the whole correctness argument). Plus `TestThread_MaxPerThreadDistinctFromGroup` (group vs thread caps decouple) and the full settings-getter matrix (default/env/DB/negative-clamp for both new keys). All DB-gated → CI must run.

## Carry-over (unchanged from `6485b0101`, re-verified)
Per-scope count narrows by `thread_short_id` (index-backed), `FOR UPDATE` stays on the parent group row, scope isolation via upstream-validated `ThreadShortID`, per-creator per-scope, soft-delete/disabled quota semantics, prior defaults (10/5) intact.

## Coverage / blind spots
- Verified by build/vet + full read of the new quota layer, api wiring, getter clamps, errcode/i18n/schema.
- NOT run: the DB-gated quota + settings suites (no MySQL) — including the concurrency ceiling test; CI must confirm.

## Verdict
**APPROVE** (re-affirmed at `1dc621e4`) — the new commit cleanly converts the earlier unbounded-growth observation into a tested feature: a third `max_total_per_group` aggregate ceiling plus a decoupled `max_per_thread`, both evaluated inside the existing parent-group `FOR UPDATE` critical section (no new lock, no gap-lock regression), with fully back-compatible defaults (total=0 disabled, per_thread falls back to per_group), a complete 409/i18n/schema surface, and a concurrency test proving the ceiling is exact. Both gates pass.
Suggested decision: `Safe to merge` once CI runs the DB-gated quota/settings suites (esp. `TestThread_TotalQuotaCeilingConcurrent`). Release note: `max_per_group`/`max_per_creator` are per-scope, plus the new `max_per_thread` / `max_total_per_group` knobs. (PR `merge: BLOCKED` — branch-protection.)

lml2468:
> **审查锚点说明** — 上面这条 APPROVE 我 review 时的 head 是 `1dc621e4`,提交瞬间分支前进到 `4e9accb3`,GitHub 因此锚到了 `4e9accb3`。已核对:`1dc621e4 → 4e9accb3` 的增量**纯文档**(3 个 `.octospec/` 文件:journal/log/brief),**代码零改动**(`git diff --stat -- ':!.octospec'` = 空)。故 review 中对 db.go/api.go/settings 的分析在锚定 head 逐字节成立,APPROVE 维持不变。

mochashanyao (dismissed):
[Octo-Q · automated review]

---

## Code Review — PR #639: feat(incomingwebhook): per-scope webhook quota with per-thread & group-total caps
**Head SHA:** `1dc621e4a04ecfe9bb8197f2dd5431c5918bd571`
**Files changed:** 18 (+874 / −84)
**Verdict:** APPROVE (zero P0/P1 findings)

---

### 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Lock strategy correctness | ✅ | Parent-group FOR UPDATE on `incoming_webhooks(group_no, thread_short_id='')` row; row guaranteed by `ensureGroupSelfRow()` in `db.go:88-111`. All concurrent creates in a group serialize on this single row — no gap-lock risk. |
| Per-scope COUNT isolation | ✅ | `db.go:53-59` scope COUNT adds `thread_short_id=?`; `db.go:74-79` per-creator COUNT also adds `thread_short_id=?`. Group-self uses `''`, threads use actual short_id — independent budgets confirmed. |
| Aggregate ceiling logic | ✅ | `db.go:96-104` total COUNT uses `WHERE group_no=? AND status!=?` (no thread filter). Gate `lim.total > 0` correctly disables when 0. |
| Read-side clamping | ✅ | `IncomingWebhookMaxPerThread()`: ≤0 → fallback to `IncomingWebhookMaxPerGroup()`. `IncomingWebhookMaxTotalPerGroup()`: <0 → 0. Both verified in `system_settings.go`. |
| Settings fallback chain | ✅ | Per-thread: DB → env `DM_INCOMINGWEBHOOK_MAX_PER_THREAD` → `IncomingWebhookMaxPerGroup()`. Total: DB → env `DM_INCOMINGWEBHOOK_MAX_TOTAL_PER_GROUP` → 0. |
| Schema registration | ✅ | `max_per_thread`: Positive (must be >0). `max_total_per_group`: non-Positive (allows 0=disabled), range [0,3650]. Both in `system_setting_schema.go`. |
| i18n coverage | ✅ | New `ErrIncomingWebhookTotalQuotaExceeded` registered in `errcode/incomingwebhook.go`; messages updated in both `zh-CN.toml` and `en-US.toml`; `mgmtTotalQuotaExceeded` renderer in `api_i18n.go`. |
| Soft-delete slot freeing | ✅ | All three COUNT queries use `status != statusDeleted`, so soft-deleted rows free their quota slots. Confirmed by `TestTotalQuotaCeiling` soft-delete subtest. |
| Test coverage | ✅ | Unit: 3 new tests in `db_test.go` + 3 in `api_thread_test.go` + 2 in `system_settings_test.go`. Concurrent: 12-goroutine ceiling test verifies exact count=4 with ceiling=4. |

### 2. Issues Found

**No P0 or P1 blocking issues.**

#### P2 — Minor observations (non-blocking)

**P2-1: Concurrent test uses `wg.Add(1)` inside goroutine-launch loop**
- Path: `api_thread_test.go` `TestThread_TotalQuotaCeilingConcurrent`
- Observation: The `wg.Add(1)` is called inside the for-loop body (between goroutine launches) rather than before the loop. In practice this is safe because the loop body runs sequentially on the main goroutine before any child goroutine can call `wg.Done()`. Not a race, but the more conventional pattern is `wg.Add(total)` before the loop for clarity.
- Severity: P2 (cosmetic/convention only, no correctness impact)
- Diff-scope: new test, not pre-existing

**P2-2: Total COUNT query scans all group webhooks per create**
- Path: `db.go:97-104`
- Observation: Each create now issues an additional `SELECT COUNT(*) ... WHERE group_no=? AND status != ?` covering the full group. For typical group webhook counts (<100) this is negligible, but a covering index `(group_no, status)` or `(group_no, thread_short_id, status)` would make it an index-only scan if this path becomes hot at scale.
- Severity: P2 (performance observation, not a blocker at current scale)
- Diff-scope: new code path

### 3. Data-Flow Tracing

| Consumed data | Upstream source | Flow verified? |
|---------------|----------------|---------------|
| `m.ThreadShortID` in scope COUNT (`db.go:57`) | `c.Param("short_id")` → validated via thread lookup → assigned at `api.go:790` | ✅ |
| `m.ThreadShortID` in per-creator COUNT (`db.go:78`) | Same as above | ✅ |
| `m.GroupNo` in all COUNT queries | `c.Param("group_no")` → group lookup → `groupNo` → `m.GroupNo` at `api.go:786` | ✅ |
| `m.CreatorUID` in per-creator COUNT | `actor.uid` from auth middleware → `m.CreatorUID` at `api.go:793` | ✅ |
| `quotaLimits.scope` | `IncomingWebhookMaxPerGroup()` for group-self; `IncomingWebhookMaxPerThread()` for thread → `api.go:808-811` | ✅ |
| `quotaLimits.perCreator` | 0 for admin (skip check); `IncomingWebhookMaxPerCreator()` for non-admin → `api.go:812-815` | ✅ |
| `quotaLimits.total` | `IncomingWebhookMaxTotalPerGroup()` → `api.go:816` | ✅ |
| Parent group row for FOR UPDATE | `ensureGroupSelfRow()` upserts `(group_no, '')` row with `statusDeleted` → `db.go:88-111` | ✅ Always exists |

**No data-flow gaps found.** Every consumed value traces to a validated source with no short-circuit paths that could deliver empty/zero/uninitialized data to a consumption point.

### 4. Blind-Spot Checklist (C1–C6)

| Check | Result | Evidence |
|-------|--------|----------|
| **C1 — Dual-path parity** | Clear | Create path has two entry points (group-level and thread-level). Both route through the same `createWebhook` handler with correct `threadShortID` binding. Delete/update paths use `webhook_id` lookup and cross-group guard — unaffected by scope changes. |
| **C2 — Control-flow ordering** | Clear | `insertWithQuota` is called from exactly one site (create handler). No nested/double invocation risk. Three quota checks are sequential within a single transaction — no ordering hazard. |
| **C3 — Authorization boundary** | N/A | No new endpoints or capability exposure. Existing auth middleware unchanged. |
| **C4 — Authorization lifecycle** | N/A | No auth/permission changes in this PR. |
| **C5 — Build/runtime parity** | N/A | Pure Go backend change, no build/packaging/bundling concerns. |
| **C6 — Governance/doc consistency** | N/A | No policy/documentation files changed. |

### 5. Cross-Round Blocker Check (R6)

N/A — this is the automated review review.

- The locking rationale is sound: using the parent group row (always exists via UNIQUE index) avoids InnoDB gap-lock deadlocks that would occur with per-scope range locks.
- The `max_per_thread` fallback to `max_per_group` when unset preserves backward compatibility — existing deployments with only `max_per_group` configured get identical behavior for threads.
- The `max_total_per_group` defaulting to 0 (disabled) is conservative and safe — the aggregate ceiling is opt-in.
- Error switch/case in `api.go:820-831` correctly distinguishes three error sentinels and falls through to generic 500 for unexpected errors.

---

[Octo-Q] verdict: **APPROVE** — no P0/P1 findings. The PR correctly implements per-scope webhook quotas with proper locking, read-side clamping, comprehensive tests, and complete i18n coverage. Two minor P2 observations noted (concurrent test pattern, total COUNT index hint) — neither blocks merge.

lml2468 (approved):
# PR Review (re-review) — PR #639 (Mininglamp-OSS/octo-server)

Re-reviewed at **head `a0b00670`** (code head `bd719c0e`; `a0b00670` is a `Merge origin/main`) against base `main`. Re-affirms my APPROVE at `4e9accb3`.
**Build/vet/test:** `go build ./modules/incomingwebhook/` exit 0 ✅, `go vet` ✅. Quota tests remain DB-gated (MySQL `connection refused` in sandbox) — unrun here; the perf commit reports them green under `-race` against MySQL/Redis/WuKongIM. CI must run them.

## What changed since `4e9accb3` (delta reviewed)
Two commits: `bd719c0e` (**perf** — fold the scope + per-creator quota counts into one query) and `a0b00670` (merge `origin/main`). The **only code change to this PR's own scope is `db.go`** — `git diff 4e9accb3 HEAD -- modules/incomingwebhook/ modules/common/` = just `db.go` (26/30). The cardtmpl files in the raw range-diff are #641 pulled in by the merge, not this PR.

## Perf fold review — semantics-preserving ✅
The scope count and per-creator count previously ran as two separate `SELECT count(*)` over the same `(group_no, thread_short_id, status != deleted)` index range, differing only by the `creator_uid` predicate. This commit folds them into one conditional aggregate:
```sql
SELECT count(*) AS scope_cnt,
       count(CASE WHEN creator_uid = ? THEN 1 END) AS creator_cnt
FROM incoming_webhook WHERE group_no=? AND thread_short_id=? AND status != ?
```
Verified correct:
- **Identical values.** `scope_cnt` = the old scope count; `creator_cnt` counts the same rows the old per-creator query did (same range + `creator_uid` predicate, now expressed as a conditional count). No semantic change.
- **`count(CASE …)` not `SUM(bool)`** — deliberately returns BIGINT, avoiding the SUM→DECIMAL→int driver-scan pitfall. Defect-pattern-aware.
- **Error precedence preserved:** scope (`scoped.Scope >= lim.scope → ErrQuotaExceeded`) is checked before per-creator (`lim.perCreator > 0 && scoped.Creator >= lim.perCreator → ErrCreatorQuotaExceeded`) — same order as the two-query version. The `lim.perCreator > 0` guard still gates the creator check, so the admin path (perCreator=0) ignores `creator_cnt` harmlessly.
- **Total count deliberately NOT folded** (`bd719c0e`'s own rationale): it spans a different range (whole group, no thread filter), and folding would force a group-wide scan on the common disabled path. Left as its own `count(*) … WHERE group_no=? AND status != ?` (line 128). Correct call.
- **The load-bearing concurrency invariant is untouched:** `SELECT id FROM group … FOR UPDATE` (line 87) still serializes on the parent-group row — no gap-lock 1213 regression (#31 lesson intact). The fold only shortens the lock-hold window (member create: 2 counts → 1; admin: unchanged) — a pure hot-path win.
- Soft-delete/disabled quota semantics unchanged (`status != statusDeleted`).

## Carry-over (re-verified, unchanged)
Per-scope quota by `(group_no, thread_short_id)` (index-backed), the `max_per_thread` + `max_total_per_group` knobs (defaults per_thread→per_group, total=0 disabled), scope isolation via upstream-validated `ThreadShortID`, 409/i18n/schema surface, no migration. All still hold.

## Coverage / blind spots
- Verified by build/vet + full read of the folded query, error precedence, lock, and total-count path.
- NOT run: the DB-gated quota suite (no MySQL) — including `TestThread_TotalQuotaCeilingConcurrent`; the fold must go green there in CI (the concurrency + count-correctness proof). The merge from main brings in unrelated cardtmpl code that its own PRs (#641) cover.

## Verdict
**APPROVE** (re-affirmed at `a0b00670` / code `bd719c0e`) — the only change since my last approval is a clean, semantics-preserving optimization that folds the scope + per-creator counts into one conditional-aggregate query (identical values, same scope→creator→total error precedence, `count(CASE)` avoiding the SUM/DECIMAL pitfall), leaving the parent-group `FOR UPDATE` serialization and the separate whole-group total count untouched. It only shortens the critical-section for the common member-create path. Both gates pass.
Suggested decision: `Safe to merge` once CI runs the DB-gated quota suite (confirming the folded query under `-race`, esp. `TestThread_TotalQuotaCeilingConcurrent`). (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (approved):
**APPROVE** — re-anchored at head `a0b00670`. Delta since my last code-anchored review (`1dc621e4`) is one real production commit plus a docs sync and an unrelated origin/main merge.

### What actually changed (byte-isolated)
- The only `#639` production `.go` file that changed vs `1dc621e4` is `modules/incomingwebhook/db.go` (commit `bd719c0e`, "fold scope + per-creator counts into one query"). `api.go`, `api_i18n.go`, `modules/common/system_setting_schema.go`, `modules/common/system_settings.go`, and `pkg/errcode/incomingwebhook.go` are byte-identical to `1dc621e4`.
- The large surrounding file list (`pkg/cardtmpl/**`, `internal/cardactiondispatch/**`, `modules/notify/**`, `modules/message/**`) comes from the `origin/main` merge (`d3041765`, cardtmpl close #641) and is **not** part of this PR. The `4e9accb3` commit is docs-only (`.octospec/`).

### The fold, verified against load-bearing properties
In `insertWithQuota`, the previous two separate range reads (scope count, then per-creator count) are merged into a single conditional-aggregation read over the same `(group_no, thread_short_id, status)` range, differing only on a `creator_uid` predicate. Confirmed intact:
- **Same critical section, no new lock.** The parent-group row lock (`group_no ... FOR UPDATE`) is unchanged; the folded read and the total read both still run after that lock and before commit. No range/gap lock introduced, so the `#31` gap-lock (1213) rationale still holds verbatim.
- **Semantics identical.** `count(*)` gives the scope count; conditional-count over `creator_uid` gives the per-creator count within the same scope — same numbers as before. Placeholder order was correctly reordered so the creator predicate binds first.
- **Error precedence unchanged:** scope (`ErrQuotaExceeded`) → per-creator (`ErrCreatorQuotaExceeded`) → total (`ErrTotalQuotaExceeded`). Per-creator still gated on `perCreator > 0`.
- **Total ceiling untouched:** still a separate, `total > 0`-gated group-wide read (no thread filter), returning `ErrTotalQuotaExceeded`.
- Conditional-count keeps the result BIGINT, avoiding the sum→decimal→int driver conversion pitfall. Defaults/clamps and the no-migration property are unchanged (settings/errcode byte-identical).

### Peer 🟡 notes assessed (both non-blocking)
1. **WaitGroup add position (test hygiene):** in the delete/regenerate concurrency tests, each add call precedes its goroutine launch and `Wait` is after the loop — the counter is incremented before the goroutine that decrements it starts, so there is no add/wait race. Test-hygiene at most, not correctness. This file is unchanged by the fold.
2. **Total-count covering index (perf-only):** the total read (`group_no`, `status`) is fully covered by the existing `idx_incoming_webhook_group (group_no, status)` composite index; both predicate columns are contiguous there. Not a correctness issue and not a real regression given that dedicated index.

### Build / CI
`go vet ./modules/incomingwebhook/` clean; module + `modules/common` + `pkg/errcode` build clean; test binary compiles. GitHub CI Build/Vet/Lint green (Test is DB-gated).

Correctness properties from prior anchors are preserved; the fold is a pure critical-section shortening with unchanged semantics. Approving at `a0b00670`.

an9xyz:
Closed in favor of #642 — reopened as a fresh PR to get a clean review round (approvals here were dismissed across several rebases/pushes and two automated reviewers had stalled). Same branch, **identical head `a0b00670`, same diff**; the review history in this thread (all rounds APPROVE, zero blocking) still applies.

---

PR #633, opened 2026-07-22
feat(cardtmpl): L0 platform card base + docs.access-request@0.2.0 pilot

an9xyz:
## Summary

Establishes the **L0 platform card base** (`octo-card@1.0`) and migrates the first
**L2a card** — `docs.access-request@0.2.0` — through it end-to-end. This is a
vertical slice: the reusable "register → schema-validate → render → inject
metadata → re-validate → envelope" spine is now real, and the docs access-request
card is its proof-of-life. Adding card #2/#3 becomes "copy the pilot directory +
implement 3 methods", not a re-architecture.

Scheme B: the `NotifyReq` outer shape is untouched; only the docs
`access_requested` branch (gate-on) is rerouted internally through
`Registry.Render`.

## Linked Spec

- Task brief: `.octospec/tasks/cardtmpl-registry-pilot/brief.md`
- Authoritative contract: `docs/platform-card-base.md` (`octo-card@1.0`, §1–§16)

## Changes

- **L0 base** (`pkg/cardtmpl/`): `Template` interface (3 methods) + `TemplateMeta`;
  `Registry` (Register/SetDefault/Freeze/Lookup/List/RegisteredForTest) with an
  L2a owner allowlist and register-time sample self-check; `Registry.Render` as the
  single 8-step ingress; `metadata.octo.{protocol,template,variant,source}` +
  `metadata.webUrl` injection; `cardtmpl_build_total` metric; `AbsoluteHTTPSURL`
  as the single source of truth for the absolute-https rule.
- **L1 contract**: `docs.access-request@0.2.0` handoff artifacts (manifest / JSON
  Schema / interaction report / sample) embedded per-subpackage; `pending` is the
  only registered view (`approved/rejected` deferred to a `0.3.0` outcome PR).
- **L2a pilot** (`pkg/cardtmpl/docs_access_request/`): `Template` implementing
  Meta/Build/FallbackText; `Meta()` returns a defensive deep copy so a frozen
  registry stays immutable.
- **Production wiring**: `modules/notify` docs `access_requested` branch (gate-on)
  now renders via the Registry; composition root registers the pilot and
  `Freeze()`s (fail-close on missing template / bad schema).
- **C1 policy**: schema-level field errors for this card (non-https avatar,
  oversized/missing fields) now return **400 with zero delivery** instead of
  silently degrading to a plain-text DM. The avatar's absolute-https host is
  authoritatively checked in preflight with a real URL parser (a regex cannot
  assert host presence — `https://?x` has an empty host).
[...]

PR #641, opened 2026-07-22
feat(cardtmpl): close post-#633 interactive-card loop (updater + callback envelope + docs.access-request@0.3.0 + metrics)

an9xyz:
## Summary

Closes the post-#633 interactive-card loop (roadmap group A). #633 shipped the
L0 platform base and `docs.access-request@0.2.0` with a **read-only** `pending`
view. This PR adds the missing "interactive → authoritative terminal" half:
server-authoritative card updates, a route-versioned callback envelope, a
`0.3.0` result view registered beside the frozen `0.2.0`, and two bounded
metrics.

## Related Issue

Follow-up to #633 (roadmap group A). No standalone issue.

## Linked Spec

`.octospec/tasks/cardtmpl-interaction-closure/brief.md`

## Changes

- **A1 `CardUpdater`** (`pkg/cardtmpl/updater.go`): `ReplaceView` (authoritative
  pending→result re-render, preserves message ID) and `Append` (progress frame,
  strictly consecutive `card_seq`). Both compose the existing
  `carddispatch.CardMutator` (CAS + idempotent replay + revision + CMD); no new
  write transport. All write authority (sender, message ID, channel, Space,
  `card_seq`) is re-derived from stored state, never from callback display
  fields.
- **A2 callback envelope** (`internal/cardactiondispatch`): routes default to
  the legacy flat body (byte-compatible; old queue rows still decode); an
  opt-in `octo-card-v1` route option selects the nested envelope
  (`protocol` / `type=card.action` / `card.{…}` / `trigger_id`). HMAC still
  covers the exact body. `response_url` is reserved, not emitted. Ingress
  (`modules/message/api_card_action.go`) enriches the internal event only from
  the effective server-authored frame + Registry; corrupt/partial metadata or
  an undeclared action fails closed before enqueue; legacy cards keep the
  existing zero-context path.
- **A3 `docs.access-request@0.3.0`** (`pkg/cardtmpl/docs_access_request/v3.go`):
  new `result` view (`approved`/`rejected`, `octo/v1`) beside the frozen
  `0.2.0`. `pending` stays `octo/v2` with the same actions (only additive
  optional `Action.data` context keys). Docs finalizer upgrades approved/denied
  cards in place to `0.3.0/result`; `cancelled`/`unavailable` stay on the legacy
  fallback. In-flight `0.2.0` cards upgrade too — missing decorative fields are
  omitted, not fabricated. **`0.2.0` handoff tree has zero diff vs `2917e6a6`.**
- **A4 metrics** (`pkg/cardtmpl/metrics.go`):
  `dmwork_cardtmpl_callback_total{…,result=ok|rejected|error}` and
  `dmwork_cardtmpl_update_total{…,result=ok|error}`; labels only from registered
  metadata + declared interactions.
- Docs: `docs/platform-card-base.md` updated; octospec journal/log/learning
  under `.octospec/`.

## Testing

Verified green locally (this branch, MySQL/Redis/WuKongIM available):

- `go vet` (focused packages)
- `pkg/cardtmpl` `go test -race -cover` → **80.5% / 85.8%**
- `internal/cardactiondispatch` (callback envelope), `pkg/cardmsg`
  (template-context), `internal/carddispatch` (mutation CAS),
  `modules/notify` (finalizer + v3 result),
  `modules/message` (`TestResolveRegistryCardContext…`)
- `make i18n-extract-check`, `make i18n-lint`

Honest gaps (not run here):

- `golangci-lint` — not installed on the dev host; **covered by the CI lint
  lane**.
- The `//go:build integration` card-action **orchestration** e2e in
  `modules/message` (`TestCardActionEndToEnd…`, `…TrustModel`, etc.) was **not**
  re-verified: under `-tags integration` the `message` test binary hits a
  pre-existing compile-time import cycle (message→app_bot→bot_api→
  messages_search→message), and the CI test lane deliberately runs the default
  build **without** `-tags integration` (see `ci.yml`, ref #557). Per the
  in-source note that path's CAS is covered by `modules/bot_api` IM cases; full
  execution needs the dmworkim `make env-test` environment. This is a
  pre-existing test-infra constraint, not introduced by this PR.

- [x] Manually verified (focused suites above)

## COMPREHENSION

   It adds a second lifecycle stage to platform cards: after the existing
   authenticated `/v1/message/card/action` ingress enqueues an action event,
   the docs finalizer now re-renders the card through `Registry` as
   `docs.access-request@0.3.0/result` and replaces the original message
   **in place** via `CardMutator` (message ID / channel / sender / Space /
   `card_seq` all re-derived from stored state). New sends default to `0.3.0`;
   the frozen `0.2.0` stays registered so in-flight pending cards keep working
   and are upgraded on decision. The callback wire format becomes
   route-versioned (legacy default, opt-in `octo-card-v1`).

   - *Boot*: `0.3.0`'s register-time self-check is fail-close — a bad
     sample/manifest panics `main.go` at startup. Mitigation: self-check +
     conformance suite pass in tests.
   - *In-flight `0.2.0` cards*: click still accepted only because `0.2.0`
     stays registered (`ActionView` resolves it); if it were dropped, existing
     pending cards would 400 on click. Kept registered + tested.
   - *Terminal card visual change*: approved/denied now render the `0.3.0`
     result (octo/v1) instead of the legacy octo/v2 outcome card. Front-end must
     render `octo/v1` result; interactive `pending` (octo/v2) web rendering is a
     pre-existing gate tracked outside this PR.
   - *Callback consumers*: unchanged unless a route explicitly opts into
     `octo-card-v1`; default stays byte-compatible.
   - *Update ordering*: `card_seq` for the docs `ReplaceView` is sourced from
     `event.EventID`, which is load-bearingly monotonic today — flagged as a
     pending learning.

   `updater_test.go` covers ReplaceView/Append happy paths + fail-closed
   branches (non-object append, stale/lower `card_seq` conflict, idempotent
   replay with no duplicate revision/CMD). `v3_test.go` + the conformance suite
   assert `0.3.0` approved/rejected render with correct
   `metadata.octo.template`, no `Action.Submit`/`Input.*`, and pass
   `cardmsg.Validate` under `octo/v1`. `action_finalizer_v3_test.go` proves
   approved→approved / denied→rejected mapping and the minimal-field `0.2.0`→
   `0.3.0` upgrade. `envelope_test.go` covers legacy-vs-`octo-card-v1` bodies +
   HMAC. `metrics_a_test.go` proves registration, increments, and bounded
   labels. All ran green locally (see Testing).

## Checklist

- [x] Updated documentation (`docs/platform-card-base.md`, `.octospec/`)

### DIFF
--- a/.octospec/journal/shared/cardtmpl-interaction-closure.md
+++ b/.octospec/journal/shared/cardtmpl-interaction-closure.md
@@ -1,0 +1,127 @@
+---
+type: Journal
+title: "Journal: cardtmpl-interaction-closure"
+description: Closed the post-#633 interactive-card loop. One commit ships CardUpdater (authoritative ReplaceView + progress-frame Append over the existing CardMutator CAS/revision/CMD path), a route-versioned callback envelope (legacy default, opt-in octo-card-v1), docs.access-request@0.3.0 with an approved/rejected result view registered beside the frozen 0.2.0, and two bounded Prometheus counters. In-flight 0.2.0 pending cards are authoritatively upgraded to 0.3.0/result in place; missing decorative fields are omitted, not fabricated.
+tags: ["cardtmpl", "platform", "docs-access-request", "wire-contract", "trust-boundary", "callback", "card-update", "auth", "space", "i18n", "metrics", "testing"]
+timestamp: 2026-07-22T16:40:00+08:00
+# --- octospec extension fields ---
+task: cardtmpl-interaction-closure
+upstream: octo-server#633
+source: self
+---
+
+# Journal: cardtmpl-interaction-closure
+
+## What was done
+
+One commit (`4336a01c`) on `cardtmpl-interaction-closure` closing roadmap
+group A — the "read-only notification → interactive terminal" gap left open by
+#633. Four tightly-coupled pieces:
+
+1. **A1 `CardUpdater`** (`pkg/cardtmpl/updater.go`)
+   - `ReplaceView`: authoritative pending→result re-render via
+     `Registry.RenderCard`, persisted through the existing
+     `carddispatch.CardMutator`. Message ID / channel binding / sender / Space
+     / `card_seq` come from the stored message + action event, never callback
+     display fields. Returns typed render/mutation errors with no fallback
+     write.
+   - `Append`: reads the *effective* card (`message_extra.content_edit` when
+     present, else original payload), appends exactly one JSON object to
+     `card.body`, re-runs `cardmsg.Validate/Finalize`, and mutates through the
+     same path. Enforces a strictly consecutive `card_seq`
+     (`target.CardSeq == snapshot.CardSeq + 1`). Non-object elements, non-card
+     sources, stale sequences, and revoked/deleted targets fail closed.
+   - Both compose `CardMutator` (CAS + idempotent replay + best-effort
+     revision append + CMD sync); no new write transport was created.
+
+2. **A2 standardized callback envelope** (`internal/cardactiondispatch`)
+   - Routes default to the legacy flat `DecisionRequest` body (byte-compatible;
+     queue rows created before this change still decode). A validated route
+     option `octo-card-v1` selects the nested envelope: `protocol`,
+     `type=card.action`, `action.{id,data,inputs}`,
+     `card.{template_id,template_version,view,message_id,channel_id,channel_type}`,
+     `actor.uid`, opaque `trigger_id` derived from the durable event ID. HMAC
+     continues to cover the exact request body.
+   - `response_url` is deliberately NOT emitted — §7 defines no authenticated
+     response request body, so no unusable URL is published.
+   - Ingress (`modules/message/api_card_action.go`
+     `resolveRegistryCardContext`) enriches the internal event only from the
+     effective server-authored frame + Registry: template identity from
+     `metadata.octo.template`, view from `Registry.ActionView`. Corrupt/partial
+     metadata or an action not declared for the resolved view fails closed
+     *before* enqueue; legacy cards without template metadata retain the
+     existing zero-context path.
+
+3. **A3 `docs.access-request@0.3.0`** (`pkg/cardtmpl/docs_access_request/v3.go`)
+   - New `result` view (`approved`/`rejected`, `octo/v1`) registered *beside*
+     the frozen `0.2.0`. `pending` view stays `octo/v2` with the same action
+     structure; the interaction report only *adds* optional `Action.data`
+     context keys (avatar/reason/time/permission/source) — a purely additive,
+     backward-compatible evolution.
+   - The docs finalizer (`modules/notify/action_finalizer.go`) now calls
+     `CardUpdater.ReplaceView` with target version `0.3.0` for
+     approved/denied; `cancelled`/`unavailable` stay on the legacy fallback
+     (`0.3.0` does not declare those states). Denial reason is rune-bounded
+     before rendering.
+   - The `0.2.0` handoff tree has **zero diff** vs merge commit `2917e6a6`.
+
+4. **A4 bounded metrics** (`pkg/cardtmpl/metrics.go`)
+   - `dmwork_cardtmpl_callback_total{template_id,version,action_id,result=ok|rejected|error}`
+     and `dmwork_cardtmpl_update_total{template_id,version,result=ok|error}`.
+     Labels come only from registered metadata + declared interactions — no
+     UID, message ID, doc ID, URL, arbitrary input, or error text.
+
+## Structural learning
+
+- **Published L1 versions are frozen; evolution is a new version registered
+  side-by-side.** `0.2.0` and `0.3.0` are both live; `SetDefault` moved new
+  sends to `0.3.0` only after its register-time self-check passed (fail-close:
+  a failing self-check panics `main.go` at boot). An in-flight `0.2.0/pending`
+  message is authoritatively replaced by `0.3.0/result` — message ID and
+  channel binding stay, only `metadata.octo.template.version` changes. This is
+  the reusable shape for every future card version bump.
+- **`CardUpdater` composes the mutator, it is not a second write path.** All
+  authority (sender, message ID, channel, Space, `card_seq`) is re-derived from
+  stored state inside `CardMutator`; the updater only supplies rendered content
+  + the target coordinates. Callback display fields never influence the write.
+
+## Gotchas worth remembering
+
+- **`card_seq` for the docs `ReplaceView` is sourced from `event.EventID`**
+  (`action_finalizer.go`). This leans on the durable event ID being
+  monotonic to satisfy the mutator's `card_seq` CAS ordering. It holds for the
+  current snowflake-style IDs, but it is an *implicit* contract: swapping the
+  event-ID generator for a non-monotonic scheme would silently break update
+  ordering. Promoted to a pending learning.
+- **Docs uses the in-process finalizer, which does NOT go through the HTTP
+  callback envelope.** A2's `legacy`/`octo-card-v1` route option only affects
+  external server-to-server HTTP consumers. Do not assume the envelope format
+  changes the docs approval path — that path is `DocsActionFinalizer.Finalize`
+  called directly by the dispatcher.
+- **Two label vocabularies still exist** (pilot `resultLabels` + notify
+  `docsLabelsFor`); A3 added a third copy for the result view. Bounded by tests
+  today; folding into an i18n locale source is deferred to roadmap C (G4).
+
+## Verification (this run)
+
+Locally green: `go vet` (focused), `pkg/cardtmpl` `-race -cover`
+(80.5% / 85.8%), `internal/cardactiondispatch` envelope, `pkg/cardmsg`
+template-context, `internal/carddispatch` mutation CAS, `modules/notify`
+finalizer + v3 result, `modules/message` `resolveRegistryCardContext`,
+`make i18n-extract-check`, `make i18n-lint`.
+
+Not run here, with an honest note on where they run:
+- `golangci-lint` — not installed on this host; the CI Vet/lint lane covers it.
+- The `//go:build integration` card-action orchestration in `modules/message`
+  (`TestCardActionEndToEndAndIdempotency`, `…TrustModel`, `…SenderBound…`, etc.)
+  runs neither here nor in CI. Under `-tags integration` the `message` test
+  binary hits a **compile-time import cycle** (message test → app_bot →
+  bot_api → messages_search → message), and the CI test lane deliberately runs
+  the DEFAULT build **without** `-tags integration` (`ci.yml`, ref #557 — the
+  integration suite needs a separately-provisioned `conv_ext_test` DB). Per the
+  in-source note, that path's D6/D9 CAS is covered by `modules/bot_api` IM
+  cases; the `message` e2e files require the dmworkim `make env-test`
+  environment to execute. A's own `message`-level regression coverage here is
+  the non-integration `TestResolveRegistryCardContextUsesEffectiveMetadataAndReport`,
+  which ran green in the default lane. The pre-existing orchestration e2e was
+  therefore NOT re-verified in this finish.
--- a/.octospec/learnings/pending/cardtmpl-interaction-closure.md
+++ b/.octospec/learnings/pending/cardtmpl-interaction-closure.md
@@ -1,0 +1,66 @@
+---
+type: Learning
+title: "Authoritative card updates need a monotonic card_seq source — reusing event IDs is an implicit contract, assert or annotate it"
+description: CardMutator orders card updates by a CAS on card_seq — a later write must carry a strictly greater card_seq than the stored frame, or it is rejected as a conflict / collapsed as a replay. The docs finalizer sources that card_seq from event.EventID. This is correct only while event IDs are monotonically increasing; nothing in the type system enforces it. Any CardUpdater caller that derives card_seq from an external ID inherits a silent-corruption risk if that ID generator ever becomes non-monotonic. Make the monotonicity requirement explicit at the call site.
+tags: ["cardtmpl", "card-update", "wire-contract", "trust-boundary", "cas", "idempotency"]
+timestamp: 2026-07-22T16:45:00+08:00
+# --- octospec extension fields ---
+source: self
+origin_task: cardtmpl-interaction-closure
+origin_pr: self
+status: pending
+candidate_rule: card-update
+---
+
+# card_seq must come from a monotonic source
+
+## Context
+
+`carddispatch.CardMutator` serializes competing card updates with a CAS on
+`card_seq`: a write whose `card_seq` is not strictly greater than the stored
+frame's `card_seq` is either rejected (`conflict`) or, when the frame hashes
+identically, collapsed into an idempotent replay. This is the whole
+correctness story for concurrent/retried updates — no duplicate revision, no
+duplicate CMD.
+
+`CardUpdater.ReplaceView` / `Append` take the `card_seq` from the caller. The
+docs finalizer (`modules/notify/action_finalizer.go`) passes
+`CardSeq: event.EventID`.
+
+## The trap
+
+`event.EventID` works as a `card_seq` **only because the durable event ID is
+monotonically increasing** under the current snowflake-style generator. That
+property is load-bearing but invisible:
+
+- Nothing in `UpdateTarget` or `CardMutator` declares "this field must be
+  monotonic." It is just an `int64`.
+- If the event-ID generator is ever swapped (per-shard counters, UUID→hash,
+  clock-rollback under NTP correction, a test double that returns a constant),
+  the CAS silently misbehaves: a legitimate later update carrying a *smaller*
+  ID looks like a stale conflict and is dropped; two updates with the same ID
+  look like a replay and the second is swallowed. No error, no log, no metric
+  — the terminal card just fails to appear.
+
+## What to do
+
+1. **At every `CardUpdater` call site, document why the chosen `card_seq`
+   source is monotonic.** For the docs finalizer: "event IDs are
+   monotonic snowflakes; used directly as card_seq." One comment closes the
+   gap for the next reader.
+2. **Prefer a source whose monotonicity is a stated invariant** (message
+   sequence, a dedicated per-message revision counter) over an ambient ID that
+   happens to increase today.
+3. **Consider asserting it in `CardMutator`** when cheap — e.g. reject a
+   `card_seq <= 0`, and where a prior frame is loaded, treat a *lower* incoming
+   `card_seq` as a typed conflict (already done) but log at a level that would
+   surface a systemic regression, not just a benign race.
+4. **Never derive `card_seq` from a client-supplied or display field.** It is
+   part of the authoritative write ordering, not presentation.
+
+## Candidate rule promotion
+
+Fits a `card-update` rule (invariants for authoritative card mutation:
+authority re-derived from stored state, monotonic `card_seq`, mutator-composed
+writes only). Discuss scope in the promotion PR — it may instead extend the
+existing trust-boundary guidance rather than stand alone.
--- a/.octospec/log.md
+++ b/.octospec/log.md
@@ -4,6 +4,28 @@ Change history for this repo's `.octospec/`, following the
 [OKF](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md)
 change-log convention (§7). Newest first.
 
+## 2026-07-22 (cardtmpl-interaction-closure)
+
+- **Feature** — Closed the post-#633 interactive-card loop (roadmap group A).
+  `CardUpdater` (`ReplaceView` + progress-frame `Append`) composes the existing
+  `CardMutator` CAS/revision/CMD path; `docs.access-request@0.3.0` adds an
+  `approved`/`rejected` `result` view (`octo/v1`) registered beside the frozen
+  `0.2.0`; the docs finalizer now upgrades approved/denied cards in place to
+  `0.3.0/result`. In-flight `0.2.0` pending cards upgrade too — missing
+  decorative fields are omitted, not fabricated.
+- **Contract** — Route-versioned callback envelope: `legacy` flat body remains
+  the default (byte-compatible), `octo-card-v1` opt-in nested envelope carries
+  `protocol`/`type=card.action`/`card.{…}`/`trigger_id`; `response_url` stays
+  reserved (no authenticated response body defined in §7). See
+  [journal](journal/shared/cardtmpl-interaction-closure.md).
+- **Observability** — Bounded counters `dmwork_cardtmpl_callback_total`
+  (`ok|rejected|error`) + `dmwork_cardtmpl_update_total` (`ok|error`); labels
+  only from registered metadata + declared interactions.
+- **Learning (pending)** — `card_seq` for authoritative updates must come from
+  a monotonic source; the docs finalizer reuses `event.EventID`, an implicit
+  contract now documented. See
+  [learning](learnings/pending/cardtmpl-interaction-closure.md).
+
 ## 2026-07-22 (cardtmpl-registry-pilot)
 
 - **Feature** — Introduced the octo-card@1.0 platform base
--- a/.octospec/tasks/cardtmpl-interaction-closure/brief.md
+++ b/.octospec/tasks/cardtmpl-interaction-closure/brief.md
@@ -1,0 +1,200 @@
+---
+type: Task
+title: "Task: cardtmpl-interaction-closure"
+description: Close the post-633 interactive-card loop with versioned result rendering, authoritative message updates, callback envelopes, and bounded metrics.
+tags: [wire-contract, trust-boundary, error-response, i18n, space, auth, testing, commit, callback, card-update]
+timestamp: 2026-07-22T16:03:51+08:00
+# --- octospec extension fields ---
+slug: cardtmpl-interaction-closure
+upstream: octo-server#633
+source: self
+---
+
+# Task: cardtmpl-interaction-closure
+
+> Post-#633 roadmap group A: A1 `CardUpdater`, A2 standardized callback
+> envelope, A3 `docs.access-request@0.3.0` result view, and A4 callback/update
+> metrics. PR #633 merged as `2917e6a6`; its `docs.access-request@0.2.0`
+> contract is published and immutable.
+
+## Goal
+
+Close the first platform-template interaction loop:
+
+1. a `docs.access-request@0.2.0` or `@0.3.0` pending card is clicked;
+2. the existing authenticated `/v1/message/card/action` ingress validates the
+   effective frame and enqueues one authoritative action event;
+3. a callback consumer receives either the byte-compatible legacy request or
+   the opt-in `octo-card@1.0` nested envelope;
+4. the docs finalizer renders `docs.access-request@0.3.0` in the
+   `result` view (`approved` or `rejected`, `octo/v1`) and replaces the
+   original message through the existing card mutation CAS/revision/CMD path;
+5. callback and update outcomes are observable through bounded Prometheus
+   counters.
+
+The result visual is implemented as Go `Template.Build` code using the attached
+backend handoff's `result.template.json`, approved/rejected samples, reports,
+and goldens as the design input. This task does not introduce the `${}` JSON
+template engine.
+
+## Background
+
+- PR #633 published `docs.access-request@0.2.0` with only the interactive
+  `pending` view. The original design handoff also contained a `result` view,
+  but it was deliberately excluded because the pilot had no updater/finalizer
+  integration and no JSON expression engine.
+- `modules/notify.DocsActionFinalizer` already rebuilds a terminal card and
+  calls `internal/carddispatch.CardMutator.Mutate`, but it bypasses
+  `Registry.Render`, manually builds the type-17 envelope, fixes the profile to
+  `octo/v2`, and emits no cardtmpl update metric.
+- `CardMutator` already owns sender binding, active-message lifecycle checks,
+  card normalization, `card_seq` CAS, idempotent replay, revision append, and
+  CMD sync. `CardUpdater` must compose this path, not create another write
+  transport.
+- The current callback HTTP body is the flat `DecisionRequest`. The public
+  client action request remains unchanged; the standardized envelope is a
+  server-to-server callback format and must be route-versioned so an upgrade
+  cannot silently break an existing strict callback consumer.
+- The merged `0.2.0` action data contains enough authoritative identity for a
+  safe terminal update (`doc_id`, `request_id`, `doc_title`, requester display
+  name), but not every decorative result field from the design sample. The
+  `0.3.0` result renderer therefore treats avatar/reason/time/permission copy as
+  optional and omits unavailable blocks for in-flight `0.2.0` cards.
+
+## Load-bearing list
+
+- **Published L1 wire contract (`wire-contract`)**: do not modify any byte under
+  `pkg/cardtmpl/docs_access_request/handoff/docs.access-request@0.2.0/`.
+  Register `0.2.0` and `0.3.0` concurrently. New sends may move to the `0.3.0`
+  default only after its register-time self-check passes.
+- **Template identity/version transition (`wire-contract`)**: an in-flight
+  `0.2.0/pending` message may be authoritatively replaced by
+  `0.3.0/result`; the message ID and channel binding stay unchanged while
+  `metadata.octo.template.version` becomes `0.3.0`.
+- **State vocabulary (`wire-contract`)**: callback `StateApproved` maps to
+  template state `approved`; callback `StateDenied` maps to template state
+  `rejected`. `cancelled` is not declared by the `0.3.0` result contract and
+  retains the existing legacy terminal fallback.
+- **Card mutation authority (`trust-boundary`, `auth`, `space`)**:
+  `sender_uid`, `message_id`, channel, channel type, Space, and monotonic
+  `card_seq` come from the stored message/action event, never callback display
+  fields. Wrong sender, revoked/deleted message, non-card source, stale sequence,
+  or channel mismatch fails closed through `CardMutator`.
+- **Effective-frame semantics (`wire-contract`)**: Append reads the current
+  effective card (`message_extra.content_edit` when present, otherwise original
+  payload), appends one element to `card.body`, then runs the same
+  `cardmsg.Validate/Finalize` and CAS mutation path. It never edits the original
+  payload directly and never bypasses revision history.
+- **Revision/CMD semantics**: authoritative `content_edit` persistence remains
+  primary; revision append and CMD sync remain best-effort exactly as in the
+  existing `CardMutator`. Replays do not append a duplicate revision or send a
+  duplicate CMD.
+- **Client action ingress (`wire-contract`, `error-response`, `i18n`, `auth`,
+  `space`)**: `POST /v1/message/card/action` request and response shapes,
+  anti-IDOR checks, visibility gates, input validation, UID rate limiting, and
+  Redis idempotency are unchanged.
+- **Callback trust boundary (`trust-boundary`, `wire-contract`)**: action data
+  is extracted from the effective server-authored frame; inputs are the
+  existing allowlisted/typed map; template ID/version come from
+  `metadata.octo.template`; view is resolved from the registered interaction
+  report and matching action ID. Client-supplied `data`, template identity, or
+  view are never trusted.
+- **Callback compatibility (`wire-contract`)**: existing routes default to the
+  legacy flat request. A validated route option explicitly selects the nested
+  `octo-card@1.0` payload. HMAC continues to cover the exact request body and
+  event ID headers. Queue rows created before this change still decode.
+- **Callback envelope**: the new format contains `protocol`,
+  `type=card.action`, `action.{id,data,inputs}`,
+  `card.{template_id,template_version,view,message_id,channel_id,channel_type}`,
+  `actor.uid`, and an opaque `trigger_id` derived from the durable event ID.
+  `response_url` remains optional/reserved because §7 defines no authenticated
+  response request body; no unusable URL is emitted.
+- **Localization (`i18n`)**: result labels are selected from
+  `BuildEnv.Lang`; business/display fields are escaped, rune-bounded, and never
+  treated as trusted card JSON. Existing applicant notification localization
+  and denial-reason behavior remain intact.
+- **Metrics cardinality**: `template_id`/`version` come from registered metadata,
+  `action_id` comes from a declared interaction, and result values are closed
+  enums. Metrics do not record arbitrary user input, UID, message ID, document
+  ID, URL, or error text.
+- **Testing (`testing`)**: use TDD; unit tests cover render/update/envelope/error
+  branches, integration tests cover finalizer-to-CAS behavior, and the existing
+  full card-action orchestration remains green.
+- **Git/PR (`commit`)**: English Conventional Commit messages and English PR
+  description; link this brief in the PR.
+
+## Out of scope
+
+- No modifications to the frozen `docs.access-request@0.2.0` assets or behavior.
+- No `${}` JSON template engine and no runtime loading of `templates/*.json`.
+- No capability-discovery/export endpoints (roadmap B).
+- No migration of docs shared/commented, summary, or generic approval cards
+  (roadmap C).
+- No `ext.*`/L2b production channel, owner registry, or private visibility
+  rollout (roadmap D).
+- No explicit notify envelope replacing `NotifyReq` (roadmap E2).
+- No public/signed `response_url` execution endpoint. The field remains absent
+  until its request schema, authorization, TTL/replay semantics, and operational
+  ownership are specified in a separate brief.
+- No change to generic `StandardActionFinalizer` rendering; only the docs
+  binding adopts the registry-backed `0.3.0` result view.
+- No database migration. Existing `message_extra` and
+  `octo_message_card_revision` storage remains authoritative.
+- No change to applicant outcome-card delivery semantics; callback retries may
+  still retry the applicant notification under the existing at-least-once
+  boundary.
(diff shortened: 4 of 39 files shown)


lml2468 (dismissed):
# PR Review — PR #641 (Mininglamp-OSS/octo-server)

Reviewed at head `a9fcaa75` against merge-base (base `main`).
**Build/vet/test:** `go build` exit 0 ✅, `go vet` ✅. `go test ./pkg/cardtmpl/... ./internal/cardactiondispatch/ ./internal/carddispatch/` → **all PASS** (updater, v3, envelope, action_view, mutation — pure, ran here). The `-short` FAIL is the DB-gated `modules/notify`/`modules/message` integration (MySQL) — infra, not a code failure.
Classification: closes the interactive-card loop — server-authoritative card updates + route-versioned callback envelope + a new template result view + metrics. **trust-boundary + wire-contract** load-bearing. Reviewed in **SECURITY** emphasis. Builds directly on the #632/#633 base I reviewed.

## 1. Spec compliance — ✅
Graded against `.octospec/tasks/cardtmpl-interaction-closure/brief.md` + `docs/platform-card-base.md`.
- **Missing:** none. A1 `CardUpdater` (ReplaceView + Append), A2 route-versioned callback envelope, A3 `docs.access-request@0.3.0` result view registered beside frozen `0.2.0`, A4 two bounded metrics — all present.
- **Over-build:** none. No `${}` engine (explicitly out of scope), no new write transport (composes `CardMutator`), `response_url` reserved-not-emitted.
- **Deviation:** none. `0.2.0` handoff tree is **byte-frozen** — `git diff … -- handoff/docs.access-request@0.2.0/**` = **0 lines**. Legacy callback body stays the default (byte-compatible).

## 2. Code quality — Approved

### Findings
No P0/P1/P2 findings.

- **Write authority is re-derived from stored/server state at every layer — the core trust-boundary property.**
  - *Ingress* (`api_card_action.go`): `resolveRegistryCardContext` reads the template ref from the **effective server-authored frame** (`effective []byte`, the stored card), not the client body; the client supplies only `actionID`, validated against the Registry's declared interactions via `ActionView`. Incomplete metadata → fail closed; undeclared action → reject before enqueue; legacy cards → zero context (unchanged).
  - *Finalizer* (`action_finalizer.go`): the `UpdateTarget` is built entirely from the internal durable-queue `event` (`event.SpaceID/ChannelType/SenderUID/MessageID`, `CardSeq: event.EventID`) — all server-authored, none from callback display fields.
  - *Updater* (`updater.go`): `ReplaceView` enforces `env.SpaceID == target.Target.SpaceID`; `Append` re-reads the stored snapshot and rejects if `snapshot.space_id != target.SpaceID` and if `card_seq` isn't strictly consecutive (`target.CardSeq-1 != snapshot.CardSeq → ErrCardMutationConflict`). All mutation identity flows through `carddispatch.CardMutator` (existing CAS + idempotent replay + revision + CMD). A callback cannot redirect an update to another message/space/sender.
- **Callback envelope is safely route-versioned + HMAC-covered.** `CallbackFormatLegacy` is the default (frozen flat body — old queue rows still decode); `octo-card-v1` is opt-in per route (`protocol`/`type=card.action`/`card.{…}`/`trigger_id`). The HMAC signs the **exact marshaled body** (`http.go:99` `Sign(secret, POST, path, timestamp, eventID, body)`), so the envelope is covered; unknown format is rejected at route load. `TestHTTPDelivererSignsExactOctoCardEnvelope` + `TestLoadRouteSpecsCallbackFormatDefaultsLegacyAndRejectsUnknown` pin it.
- **In-flight `0.2.0`→`0.3.0` upgrade omits, never fabricates.** The v3 result renderer emits decorative fields (avatar/reason/permission/time) only `if input.… != ""`, so a `0.2.0` card lacking them upgrades with the blocks omitted — no invented data. `TestV3ResultAcceptsPublishedV2ActionDataMinimum` covers exactly the published-`0.2.0`-minimum path.
- **Dual-version registration without mutating the frozen one.** `0.2.0` and `0.3.0` register concurrently; `TestV3RegistersPendingAndResultWithoutMutatingV2` asserts `0.2.0` stays intact. `pending` stays `octo/v2` (only additive optional `Action.data` context keys); `result` is `octo/v1`.
- **Metrics bounded/low-cardinality:** `callback_total{…,result=ok|rejected|error}` and `update_total{…,result=ok|error}` — labels only from registered metadata + declared interactions (no free-form/id labels).
- **Final-boundary validation retained:** updates go through `cardmsg.NormalizeContentEdit` (validate) before mutate, so a malformed frame can't reach the wire.

## 3. Test assessment
Strong and precisely targeted at the risk surface: updater (identity-preserving ReplaceView, Append-uses-effective-frame, non-contiguous-seq rejection, fail-closed on bad inputs/deps), v3 (dual-register-without-mutating-v2, v2-minimum acceptance, server-authored action data, fallback localization, meta), envelope (legacy+octo-card marshal, default-legacy/reject-unknown, exact-signature). All pure suites pass here. The DB-backed finalizer/ingress integration is MySQL-gated → CI must run it (the end-to-end click→terminal path).

## 4. Coverage / blind spots
- Verified by build/vet + the pure cardtmpl/cardactiondispatch/carddispatch suites + full read of updater, ingress enrichment, finalizer target provenance, envelope signing, and the v3 omit-not-fabricate path + 0.2.0-frozen check.
- NOT run: the DB-gated notify/message integration (the live click→enqueue→finalize→ReplaceView round trip). Recommend the CI DB run.

## Verdict
**APPROVE** — a mature, security-correct closure of the interactive-card loop. Every write-authority field (sender/message/channel/space/card_seq) is re-derived from stored or durable-queue server state at ingress, updater, and finalizer — a callback's display fields can never redirect a mutation; the updater composes the proven `CardMutator` CAS/idempotent/revision path rather than a new transport; the callback envelope is route-versioned with legacy as the byte-compatible default and HMAC over the exact body; the `0.2.0` contract is byte-frozen while `0.3.0` registers beside it and upgrades in-flight cards by omission not fabrication; metrics are bounded. Strong targeted tests on all of it. Both gates pass.
Suggested decision: `Safe to merge` once CI runs the DB-gated notify/message integration (the end-to-end update path). (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (dismissed):
Closes the post-#633 interactive-card loop cleanly. The end-to-end authz story holds up under tracing, the frozen `0.2.0` tree is untouched, metrics are bounded, and — importantly — there is **no new unauthenticated inbound callback surface**. Approving with one hardening recommendation on `ReplaceView`.

### Trust boundary verification (the highest-risk surface)

Traced the "callback envelope" end to end. The envelope is the **outbound** wire body the dispatcher POSTs to a statically-registered first-party consumer URL — it is **not** an inbound endpoint an arbitrary user can POST to. Key facts:

- The callback **actor** (`OperatorUID`) is fixed server-side at the authenticated ingress (`modules/message/api_card_action.go`), which already enforces anti-IDOR channel binding, stored-channel membership, `sender == active bot`, revoke/delete/visibility gates, idempotent claim (keyed on message+action+operator, not client token), and inputs validation. Callback display fields are never trusted for authority.
- `internal/cardactiondispatch/http.go` signs every delivery with HMAC-SHA256 over `version\nMETHOD\npath\ntimestamp\neventID\nsha256(body)`; `Verify` uses `hmac.Equal` (constant-time). `X-Octo-Timestamp` / `X-Octo-Event-ID` are sent for consumer-side replay/anti-tamper checks.
- `response_url` (the classic forgeable inbound-execution field) is **deliberately not emitted**, with a test asserting its absence (`envelope_test.go`). No signed inbound response contract exists yet, so none is exposed.
- The consumer's HTTP response is parsed by `DecodeDecisionResult` with a 64 KiB cap, `DisallowUnknownFields`, trailing-byte rejection, enum validation, `requester_uid` bounds, and a fail-closed "terminal decision requires requester_uid" rule.
- Idempotency/anti-replay on the processing side: durable queue with lease/renew/Ack/Nack + CAS on `card_seq` + `content_edit_hash` replay short-circuit in `CardMutator.Mutate`.

Conclusion: forging a callback to approve/deny on someone else's behalf is not reachable through this PR — there is no unauthenticated inbound path, and the mutator enforces `stored message.FromUID == SenderUID` (the server-fixed bot), preserving the #635 ownership model.

### 🟡 Non-blocking (recommended hardening)

- `pkg/cardtmpl/updater.go` `ReplaceView` is asymmetric with `Append`: `Append` snapshots the stored effective frame and cross-checks the stored `space_id` against the target before mutating, but `ReplaceView` only checks `env.SpaceID == target.Target.SpaceID` and then renders/writes without a `Snapshot()` cross-check of stored space or source template/version. `CardMutator.Mutate` guards sender-ownership, card-ness, lifecycle, and `card_seq` CAS — but it does **not** compare stored `space_id` or template identity. In practice the finalizer derives `event.SpaceID` from the card's authoritative origin row at ingress, so the values are consistent today and the sender-ownership gate still bounds writes to the same bot's cards; this is defense-in-depth, not a reachable external bypass. Recommend `ReplaceView` snapshot the effective frame and fail closed when stored space / expected source view do not match the transition, mirroring `Append`, with tests for mismatched space, unrelated template, and malformed metadata. Not merge-blocking given the upstream server-fixed authority.

- Full `modules/notify` / `modules/message` suites and `golangci-lint` were not run locally (MySQL/Redis/lint host gaps) — covered by CI lanes; the `-tags integration` orchestration e2e is a documented pre-existing import-cycle constraint, not introduced here.

### ✅ Highlights

- No inbound callback endpoint; `response_url` reserved-not-emitted with a guarding test.
- HMAC covers the exact selected envelope body; route opt-in keeps the legacy flat body byte-compatible (old queue rows still decode).
- `0.3.0` registered beside a frozen `0.2.0` (0.2.0 handoff tree has zero diff), so in-flight pending cards keep working and upgrade on decision; `pending` stays `octo/v2` with only additive optional `Action.data` keys.
- Metrics use fixed, registry-bounded label sets (`template_id`/`version`/`view`/`action_id`/`result`) with result vocab clamped to enums — no PII, no per-user/per-message cardinality.
- Route-missing handling defers (no attempt consumed) within a bounded window anchored on first-observed-miss, self-healing across rolling deploys instead of dead-lettering prematurely.
- Callback URL validation rejects fragments, queries, credentials, and non-http(s); proxy env ignored to prevent SSRF reopening.

yujiawei (changes requested):
# Code Review — PR #641 (octo-server)

**Head SHA:** `a9fcaa75388dcd6c241923e5d6b839fc14618580` · **merge-base:** `4b8cd2d341cb016921e5146dc50fbb5cfe05c65a`
Reviewed the full 39-file diff, built the touched packages (`go build`/`go vet` green), and confirmed the frozen `docs.access-request@0.2.0` handoff tree has **zero diff** vs `2917e6a6`.

This review runs in two stages: **spec compliance** first, then **code quality**. Both must pass to approve.

---

## 1. Spec Compliance

**Spec: ✅ (scope correct)**

- **Under-build:** none. Every acceptance item is present — `0.3.0` registered beside the frozen `0.2.0`; `pending: octo/v2` + `result: octo/v1` (states `approved`/`rejected`); register-time self-check + conformance; `CardUpdater.ReplaceView`/`Append` composing `CardMutator`; route-versioned callback envelope (legacy default, opt-in `octo-card-v1`); the two bounded metrics with closed enums.
- **Over-build:** none. No out-of-scope work leaked in — no `${}` engine, `response_url` is reserved not emitted, no capability/export endpoint, no DB migration, no change to the frozen `0.2.0` assets (verified byte-identical), no generic `StandardActionFinalizer` change.
- **Deviation:** one partial deviation feeds directly into the P1 below. The brief's localization load-bearing item requires display fields to be *"escaped, rune-bounded"* before rendering, and acceptance requires *"the denial reason is bounded before rendering."* Only `denyReason` is bounded in the finalizer; the other display fields sourced from the callback `Display` map are handed to the strict `0.3.0` schema **untruncated**, so "bounded" is enforced by schema-rejection instead of truncation — which breaks the finalize flow (see P1).

---

## 2. Code Quality

### P1 — Terminal update fails closed on in-contract oversized display fields; approver card stuck + applicant never notified

`modules/notify/action_finalizer.go:87,150,155,170` → `replaceWithRegistryResult`

The callback response contract accepts `Display` values up to **500 runes** (`internal/cardactiondispatch/contract.go:226`). The finalizer copies three of them straight into the `0.3.0` result fields with **no truncation**:

- `title` ← `result.Display["title"]` → schema `document.title` maxLength **200**
- `operatorName` ← `result.Display["operator_name"]` → schema `decision.operatorName` maxLength **120**
- `decidedAtDisplay` ← `result.Display["decided_at"]` → schema `decidedAtDisplay` maxLength **80**

(Only `denyReason` is truncated, at line 145–147.)

**Failure scenario:** the docs backend returns, e.g., `display.operator_name` of 121–500 runes (plausible for a localized name carrying a title/department, e.g. `张三（产品部高级产品经理）…`) or a verbose `display.decided_at`. This is fully within the documented ≤500-rune `Display` contract, so `DecodeDecisionResult` accepts it. `ReplaceView` → `RenderCard` → `renderCore` then runs `InputSchema.Validate`, which rejects it with `ErrFieldsInvalid`. `Finalize` returns that error → the dispatcher nacks and retries; because the oversized value is deterministic, **every retry fails identically until attempts are exhausted**. Net effect:
1. the approver's card is **never** replaced — it stays live in `pending` with clickable Approve/Deny buttons;
2. the applicant is **never** notified, because the applicant `Send` (line 128) is sequenced *after* the failed `ReplaceView` and is skipped.

The entire purpose of the interactive loop silently fails for that request, with no terminal state anywhere.

**Fix:** truncate each display-sourced field to its schema cap before building `fields` (exactly as `denyReason` already is) — `title`→200, `operatorName`→120, `decidedAtDisplay`/`requestedAtDisplay`/`messageTimeDisplay`→80, `permission.*`→80. Truncation (never fabrication) is the pattern the brief already prescribes for the denial reason.

### P2 — Corrupt (non-object) `metadata.octo` is classified as legacy instead of failing closed

`pkg/cardmsg/interactive.go:51` (`HasCardTemplateMetadata`), reached from `modules/message/api_card_action.go:299`

`cardmsg.Validate` never inspects `metadata`, so a stored card can carry `"metadata":{"octo":"corrupt"}` (a string) alongside a valid `Action.Submit`. `HasCardTemplateMetadata` does `octo, _ := metadata["octo"].(map[...])`; a non-object yields a nil map → returns `false` → `resolveRegistryCardContext` returns the *empty legacy* context with no error. On an `octo-card-v1` route, `contract.go:186` then sees `event.Card == CardContext{}` and emits the **legacy flat** body, bypassing the intended fail-closed rejection of corrupt template metadata. The brief says invalid/unknown template metadata must be rejected before enqueue for registry-backed routes. Reachability is low today (docs cards are always produced by `Registry.Render`, which emits well-formed `octo`, and routes are gated by sender+owner+action_type), so this is defense-in-depth rather than an exploit — but the classifier should return `true` when `octo` is present-but-malformed so partial/corrupt identity fails closed consistently.

### P2 — Callback envelope's card identity is not cross-checked against the route it's delivered to

`modules/message/api_card_action.go:311` (`ActionView`) vs `:259–274` (route metadata)

Template id/version/view for the nested envelope are resolved from `metadata.octo` + the declared interaction report, but the *route* is chosen from `Action.Submit.data.{owner,action_type}` (`cardActionRouteMetadata`). These two are never cross-checked. A card whose `metadata.octo` claims `docs.access-request@0.3.0` but whose stored action data carries a different `owner`/`action_type` would deliver an `octo-card-v1` envelope stamped with the docs template to a *different* registered route. `Registry.Render`'s `assertActionContract` prevents this for any Registry-produced card (data.owner is forced to equal the template's owner), so it's only reachable via a hand-crafted card on a matching registered route — low reachability, but worth a consistency assertion that the envelope's `card.template_id` owner matches the resolved route.

### P2 — `Append` requires globally-contiguous `card_seq`, contradicting the "monotonic" doc

`pkg/cardtmpl/updater.go:102`

`Append` enforces `target.CardSeq-1 == snapshot.CardSeq` (strict contiguity), while `UpdateTarget.CardSeq` is documented as *"positive monotonic sequence"* (line 17) and the underlying CAS only requires strictly-greater. With `event.EventID` (a global snowflake) as the source, consecutive updates to one message are almost never contiguous, so a fresh append can be rejected as a conflict. **This is not reachable today — `Append` has no production caller** (only `ReplaceView` is wired, and it does not require contiguity). Treat as: either intentional "progress-frame appender owns its sequence" semantics (then fix the "monotonic" comment to say "consecutive"), or relax to strictly-greater to match `ReplaceView`/CAS. Non-blocking because unreachable.

### P2 — Missing `card_seq` monotonicity rationale at the `ReplaceView` call site

`modules/notify/action_finalizer.go:202` (`CardSeq: event.EventID`)

The author's own pending learning (`.octospec/learnings/pending/cardtmpl-interaction-closure.md`) commits to documenting, at every `CardUpdater` call site, *why* the chosen `card_seq` source is monotonic (for the docs finalizer: "event IDs are monotonic snowflakes; used directly as card_seq"). The bare `CardSeq: event.EventID` has no such comment. Maintainability nit — the invariant is load-bearing and invisible; if the event-ID generator is ever swapped for a non-monotonic source the CAS silently drops valid updates.

---

## 3. Overall Verdict

**CHANGES_REQUESTED** — the P1 (fail-closed terminal update + applicant notification on in-contract oversized display fields) blocks merge. The P2s are non-blocking follow-ups.

## 4. Suggested direction

Fix the P1 by truncating each display-sourced field to its `0.3.0` schema cap in `replaceWithRegistryResult` before `json.Marshal(fields)` (mirror the existing `denyReason` truncation). Optionally add a regression test that drives `Finalize` with a `Display["operator_name"]`/`Display["decided_at"]` above the schema cap and asserts the terminal render still succeeds. The P2s can be batched: harden `HasCardTemplateMetadata` for malformed `octo`, add the envelope↔route owner consistency assertion, reconcile the `Append` contiguity comment, and add the one-line `card_seq` rationale.

## 5. Additional notes

- `go build` / `go vet` on the touched packages are green; frozen `0.2.0` tree confirmed byte-identical to `2917e6a6`. Full `-tags integration` `modules/message` orchestration e2e was not re-run (pre-existing import-cycle constraint the CI test lane also skips) — unchanged by this PR.
- The `docsBanner` anon-vs-named branch, `RichTextBlock` TextRun no-escape rationale, and `AbsoluteHTTPSURL` host-rejection paths from prior rounds remain sound.

an9xyz:
Thanks @yujiawei — all points addressed in `b8d13a2e`.

**P1 (blocking) — fixed.** `replaceWithRegistryResult` now truncates every display / `event.Data` field to its `0.3.0` result-schema cap before `json.Marshal` (title & sourceName → 200, requester.name & operatorName → 120, decided/requested/message/permission → 80, requestReason → 300), mirroring the existing `denyReason` truncation. Added `TestDocsActionFinalizerV3TruncatesOversizedDisplayFields`, which drives `ReplaceView` through a **real renderer** with every field oversized (but within the ≤500-rune callback contract) and asserts the terminal update still renders and persists — it fails (ErrFieldsInvalid) without the truncation.

**P2-a — fixed.** `HasCardTemplateMetadata` returns `true` when `metadata.octo` is present but malformed (non-object), so corrupt template identity fails closed instead of degrading to the legacy flat body.

**P2-b — fixed.** `resolveRegistryCardContext` cross-checks the resolved template owner against the route owner (`Action.data.owner`) and rejects mismatches. Covered by a new case in `TestResolveRegistryCardContext…`.

**P2-c — fixed.** `UpdateTarget.CardSeq` doc now distinguishes `ReplaceView` (strictly-greater, monotonic) from `Append` (exactly-consecutive — the progress-frame appender owns its own numbering).

**P2-d — fixed.** Added the `event.EventID → card_seq` monotonicity rationale at the `ReplaceView` call site.

Verification (all green): focused `pkg/cardtmpl` `-race` (80.5% / 85.8%), `pkg/cardmsg` (85.5%), `internal/cardactiondispatch`, `internal/carddispatch` mutation, `modules/notify` v3 finalizer, `modules/message` context, `make i18n-lint`, `go vet`.

lml2468 (approved):
# PR Review (re-review) — PR #641 (Mininglamp-OSS/octo-server)

Re-reviewed at **new head `b8d13a2e`** against merge-base (base `main`). Supersedes my APPROVE at `a9fcaa75`.
**Build/vet/test:** `go build` exit 0 ✅, `go vet` ✅. `go test ./pkg/cardtmpl/... ./modules/notify/ -run 'V3|Finalizer|Truncate'` → **PASS** (incl. the new truncation regression). DB-gated notify/message integration still MySQL-gated → CI.

## What changed since `a9fcaa75` (delta reviewed)
One commit — `b8d13a2e` "address PR #641 review (P1 truncation + P2 hardening)" — touching `action_finalizer.go`, `updater.go`, `api_card_action.go`, `pkg/cardmsg/interactive.go` + tests. The handoff contracts (0.2.0 **and** 0.3.0) are byte-untouched this commit (`git diff … -- handoff/**` = 0).

## P1 (yujiawei's 🔴) — display fields not truncated → stuck-pending card — FIXED ✅
This was a **real blocker I missed** on `a9fcaa75`, and yujiawei called it correctly: callback `Display`/`event.Data` values are legal up to 500 runes (the `cardactiondispatch` contract), but the 0.3.0 result schema caps each field far lower. An over-long field made `RenderCard`'s `InputSchema.Validate` return `ErrFieldsInvalid` → `ReplaceView` fails **deterministically** → the approval card is stuck at `pending` forever and the applicant is never notified. Owning the miss.

The fix truncates every display field to its **exact schema cap** before render (byte-verified against `handoff/docs.access-request@0.3.0/contract/data.schema.json`):
- `capResultTitle=200` → title / docId / sourceName (schema maxLength 200 ✅)
- `capResultName=120` → requester.name / decision.operatorName (120 ✅)
- `capResultShortLabel=80` → permission.label/roleLabel, decidedAt/requestedAt/messageTime (80 ✅)
- `capResultReason=300` (= `MaxExcerptRunes`) → requestReason / rejectionReason (300 ✅)
- URLs (url/avatarUrl, schema maxLength 2048) deliberately **not** truncated — truncation would break the link; https is pattern/render-validated. Correct call.

Truncate-not-fabricate, consistent with the existing denyReason handling. `TestDocsActionFinalizerV3TruncatesOversizedDisplayFields` drives the **real** updater/RenderCard with every source field over its cap (but ≤500-rune contract) and asserts the update succeeds — a genuine regression pin, not a stub. Good catch by yujiawei; fix is complete.

## P2 (Jerry-Xin / codex 🟡) — ReplaceView snapshot asymmetry — addressed as documented defense-in-depth
`ReplaceView` retains its `env.SpaceID == target.Target.SpaceID` guard and composes the `CardMutator`, which enforces `stored FromUID == sender` (the #635 ownership model). This commit adds clarifying comments (CardSeq monotonicity from the snowflake `event.EventID`; the CAS strict-increase contract) rather than a full snapshot re-verification like `Append` does. That's acceptable: every ReplaceView identity field (sender/message/channel/space/card_seq) is re-derived from the server-authored durable-queue `event`, and the mutator's sender-match makes a cross-owner rewrite structurally impossible — so the asymmetry is a defense-in-depth gap, not an exploitable authz hole (matches the group's reconciliation). Reasonable to keep the snapshot-symmetry as a tracked follow-up.

## Carry-over (re-verified, unchanged)
Write authority re-derived from stored/server state at ingress + updater + finalizer (callback display fields never grant authority); callback envelope is **outbound**, route-versioned, HMAC over exact body, `response_url` reserved-not-emitted; 0.2.0/0.3.0 dual-registered with 0.2.0 frozen; metrics bounded; final-boundary `cardmsg` validation before mutate. All still hold.

## Test assessment
The new `TestDocsActionFinalizerV3TruncatesOversizedDisplayFields` closes the exact coverage gap that let P1 through (previously no test exercised over-cap display fields end-to-end). Combined with the existing updater/v3/envelope suites — all green here. DB-gated finalizer integration → CI.

## Verdict
**APPROVE** (re-affirmed at `b8d13a2e`, superseding my `a9fcaa75`) — the one real blocker (yujiawei's P1: unbounded display fields → schema-validation failure → permanently stuck approval card + missed notification) is correctly fixed by truncating every field to its exact 0.3.0 schema cap (URLs excepted), with a regression test driving the real render path. The P2 ReplaceView snapshot asymmetry is a non-exploitable defense-in-depth follow-up (identity is all server-authored + mutator sender-match). The interactive-card loop's security properties I verified last round are intact and the handoff contracts are untouched. Both gates pass.
Suggested decision: `Safe to merge` once core CI (Build/Test/Vet/Lint + DB-gated notify integration) is green. Consider the ReplaceView snapshot-symmetry hardening as fast-follow. My thanks to yujiawei — the truncation call was right and my earlier APPROVE under-called it.

Jerry-Xin (approved):
# Re-review — PR #641 (octo-server) @ `b8d13a2e`

**APPROVE.** Re-reviewed the single new commit (`a9fcaa75..b8d13a2e`, verified linear: exactly 1 commit ahead, zero rebase drift) against the prior round. The peer's P1 blocker was **real**, and this commit **fully fixes it** with a passing regression test. The P2 hardening items are all addressed. Build / Vet / Lint green.

## The P1 blocker was real — and is now resolved (credit to the reporter)

I acknowledge my prior APPROVE at `a9fcaa75` **under-called** this. The finding is a genuine functional-correctness/availability defect, not defense-in-depth:

- The callback response contract accepts `Display` values up to **500 runes** (`internal/cardactiondispatch/contract.go:226`).
- At `a9fcaa75`, `replaceWithRegistryResult` copied display-sourced values straight into the `0.3.0` result fields with **no truncation** except `denyReason` — `title`→`document.title` (schema cap 200), `Display["operator_name"]`→`decision.operatorName` (cap 120), `Display["decided_at"]`→`decidedAtDisplay` (cap 80), plus `actor`→`requester.name` (120), `source_name` (200), `request_reason` (300), `requested_at_display`/`message_time_display`/`permission.*` (80).
- `ReplaceView` → `RenderCard` → `renderCore` runs `InputSchema.Validate` (`pkg/cardtmpl/registry.go:407`), which rejects an over-cap field with `ErrFieldsInvalid`. In `DocsActionFinalizer.Finalize`, the terminal `replaceWithRegistryResult` is called with `return err` on failure (`modules/notify/action_finalizer.go:101-104`), so the applicant `sender.Send` (sequenced *after*, ~line 128) is **skipped**. The value is deterministic, so every retry fails identically → the approval card stays live in `pending`, and the applicant is never notified. Terminal state is unreachable. That is a real stuck-flow bug.

## What `b8d13a2e` fixes

- Adds `truncRunes` + schema-cap constants (`capResultTitle=200`, `capResultName=120`, `capResultShortLabel=80`, `capResultReason=300=MaxExcerptRunes`) and applies them to **every** display-sourced field before `json.Marshal(fields)`: `title`, `sourceName`, `requester.name`, `operatorName`, `decidedAtDisplay`, `requestReason`, `requestedAtDisplay`, `messageTimeDisplay`, `permission.label/roleLabel` (`modules/notify/action_finalizer.go:137-224`). `avatarUrl` is intentionally left uncapped (schema has no maxLength; truncation would corrupt the URL — correct).
- Verified against the `0.3.0` schema: every field the finalizer writes that has a declared `maxLength` is now capped at exactly that bound. `requestId`/`docId` are intentionally not truncated — they originate from `SubmitAction(effective, …)`, i.e. the server-authored pending frame produced by `Registry.Render` (already schema-bounded ≤200 at render), not from the ≤500-rune untrusted callback `Display` map. Defensible.
- Regression test `TestDocsActionFinalizerV3TruncatesOversizedDisplayFields` drives an over-cap value (≤500 runes) into every source field through the **real** `CardUpdater`/`RenderCard` path and asserts the terminal update **succeeds** and the mutator is called. Passes locally.

## P2 follow-ups — all addressed

- **Corrupt `metadata.octo` (non-object) now fails closed:** `HasCardTemplateMetadata` returns `true` when `octo` is present-but-malformed (`pkg/cardmsg/interactive.go:55-68`), so the caller rejects instead of degrading to the empty legacy context. Test added.
- **Envelope↔route owner cross-check:** `resolveRegistryCardContext` now takes `actionData` and asserts the template's `ActionContract.Owner` equals the route `data.owner`, rejecting mismatches (`modules/message/api_card_action.go:299-330`). Test added (owner-mismatch rejected).
- **`Append` contiguity vs "monotonic" doc reconciled + `card_seq` rationale documented:** `UpdateTarget` doc now distinguishes strictly-greater (`ReplaceView`/CAS) from consecutive (`Append`), and the `ReplaceView` call site carries the `event.EventID` monotonic-snowflake rationale (`pkg/cardtmpl/updater.go:15-20`, `action_finalizer.go` call site).

## Non-blocking (unchanged from prior round)

`ReplaceView` still does not `Snapshot`-verify stored Space/template identity before mutating, unlike `Append`. Write-authority is still enforced by the downstream `Mutate` CAS keyed on the trusted server-authored `SenderUID`/`MessageID`/`ChannelID`, so this remains a defense-in-depth asymmetry, not an exploitable escalation. Worth hardening in a later pass; does not block.

## Verification

- Delta byte-isolated `a9fcaa75..b8d13a2e` (linear, 6 files); scope is exactly the review fixes, no unrelated changes.
- `go build` / `go vet` on `modules/notify`, `modules/message`, `pkg/cardtmpl`, `pkg/cardmsg`, `internal/cardactiondispatch` — green.
- Targeted tests green: `modules/notify` (incl. the new truncation test), `modules/message` (context/P2 tests), `pkg/cardmsg`, `pkg/cardtmpl`. Full DB-gated lane not exercised locally (MySQL-gated; CI Test lane pending) — acceptable.
- CI: Build / Vet / Lint / scans all pass.

Good fix. The truncation blocker was correctly identified by the prior reviewer and is now fully closed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #641 (octo-server)

## Summary

This PR closes the post-#633 interactive-card loop for `docs.access-request`: it adds a `CardUpdater` abstraction (`ReplaceView` / `Append`) with CAS enforcement on `card_seq`, implements the `octo-card-v1` callback envelope format with legacy-fallback routing, registers `docs.access-request@0.3.0` alongside the existing v0.2.0, and wires template-identity cross-validation at the callback ingress. The change touches 39 files (+2691/-101), spanning the card template registry, the action dispatch pipeline, the notification finalizer, and the message API layer.

The architecture is well-designed. CAS enforcement in `casWriteOnce` eliminates the race window via `FOR UPDATE SELECT` with `stored.CardSeq >= write.CardSeq` conflict detection. The v3 template correctly delegates pending rendering to v2 and builds a separate result view for terminal states. Truncation caps in `action_finalizer.go` align exactly with the runtime bounds in `docs_action.go` and the JSON schema `maxLength` constraints. One P2 semantic mismatch in the outcome card and two minor nits below — no blocking issues.

## Verification

- ✅ **CAS monotonicity** — `casWriteOnce` (`internal/carddispatch/mutation.go`) uses `FOR UPDATE SELECT` to lock the row, then checks `stored.CardSeq.Int64 >= write.CardSeq` before allowing the write. `ReplaceView` passes `target.CardSeq` (idempotent replay via `>=`); `Append` uses exact-consecutive check (`snapshot.CardSeq != target.CardSeq-1`). Both paths are correct.
- ✅ **Truncation cap alignment** — `capResultTitle(200)` = `maxTitleRunes(200)`, `capResultName(120)` = `maxActorRunes(120)`, `capResultShortLabel(80)` = `maxTimestampRunes(80)`, `capResultReason(300)` = `MaxExcerptRunes(300)`. Schema `maxLength` values in `data.schema.json` are respected.
- ✅ **Callback format routing** — `marshalDecisionRequest` correctly routes `CallbackFormatOctoCardV1` to envelope reconstruction and falls back to legacy flat format when `CardContext` is absent (nil event check at `:74`). Fail-closed on nil.
- ✅ **Owner cross-validation** — `resolveRegistryCardContext` (`modules/message/api_card_action.go:317`) cross-checks template `ActionContract.Owner` against `actionData["owner"]`. Mismatch returns error. Fails closed for corrupt metadata (`HasCardTemplateMetadata` returns true but `CardTemplateContext` returns false).
- ✅ **v3 pending delegation** — `v3.Build` for pending state delegates to v2 `Build` (returns fresh `BuildResult`), then mutates returned `Actions` slice in-place to inject v3 data keys. No shared-state corruption — v2 `Build` produces fresh slices.
- ✅ **Template registration** — `main.go` registers both v0.2.0 and v0.3.0, sets v0.3.0 as default. Existing v0.2.0 cards remain renderable via registry lookup.
- ✅ **Metrics cardinality** — `callback` and `update` counter vectors use bounded label sets (`template_id`, `template_version`, `view`, `result`). Result normalization maps to fixed strings ("ok"/"rejected"/"error" for callback; "ok"/"error" for update). No unbounded user-input labels.
- ✅ **`ActionView` ambiguity guard** — `registry.ActionView` rejects action IDs that map to multiple views, preventing ambiguous routing.

Static analysis only at head `b8d13a2e`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 and two nits below.

### P2 — `requestedAtDisplay` receives decision timestamp instead of request timestamp (`pkg/cardtmpl/docs_access_request/v3.go:155`)

The `Build` method for approved (L155) and rejected (L173) states passes `decision.DecidedAtDisplay` to `docsOutcomeHeader`'s `requestedAtDisplay` parameter. The `docsOutcomeHeader` function (`pkg/cardtmpl/docs_action.go:351`) renders this as `docsRequesterRow(name, requestedAtDisplay)`, which displays with the "Request time" / "发起申请" label.

The consequence is that on the outcome card, the label says "Request time" but shows the **decision** timestamp. For a request made hours ago and approved "just now", users see "Request time: just now" — the label and value disagree.

Data-flow: `event.Data["requested_at_display"]` (actual request time) → `v3Fields.RequestedAtDisplay` (populated in `replaceWithRegistryResult` L89) → should flow to `docsOutcomeHeader`'s `requestedAtDisplay` parameter, but `decision.DecidedAtDisplay` is passed instead.

Fix: pass `f.RequestedAtDisplay` in both the approved and rejected paths:

```go
// approved (L155):
body, deepLink, err := docsOutcomeHeader(content, f.Requester.Name, f.Decision.OperatorName, f.RequestedAtDisplay)
// rejected (L173):
banner := docsBanner("rejected", resultSuffix, f.Decision.OperatorName, f.RequestedAtDisplay)
```

New in this PR. Pre-existing v2 template does not render outcome cards via this code path.

### Nit — `NewDocsActionFinalizer` creates nil-updater finalizer without warning (`modules/notify/action_finalizer.go:48`)

The constructor creates `DocsActionFinalizer{...}` with `CardUpdater: nil`. The branching at L100 (`if f.updater != nil`) silently falls through to the legacy mutation path for terminal states. Production callers use `NewDocsActionFinalizerFromContext`, which creates the updater — so this is not a current production issue. But the exported bare constructor is a footgun for future callers.

Consider adding a deprecation notice or a log warning when `updater` is nil at terminal-state finalization.

### Nit — `Append` method defined but no production caller (`pkg/cardtmpl/updater.go:97`)

`Append` implements read-modify-write CAS with exact-consecutive CardSeq check (`snapshot.CardSeq < 0 || target.CardSeq-1 != snapshot.CardSeq`). No caller in this PR exercises it — likely scaffolded for upcoming work. When Append gets its first caller, note that any intervening `ReplaceView` (which allows `stored >= write`, including equal) will cause Append's exact-consecutive check to conflict. Correct CAS semantics, but worth documenting in the `Append` godoc.

## Blind-spot checklist

- **C1 (dual-path parity)**: N/A — no symmetric subscribe/unsubscribe or create/delete pairs in this PR. State transitions (pending → approved/rejected) are one-directional.
- **C2 (control-flow ordering)**: Clear — `v3.Build` delegates to v2 `Build` exactly once per call; no double-call risk. `renderCore` pipeline (schema validate → view resolve → Build → URL validate → assemble → cardmsg.Validate) runs in correct order.
- **C3 (authorization boundary)**: Clear — `resolveRegistryCardContext` cross-checks template owner vs action route owner, and `authorizeCardChannelMember` gates channel membership. No new endpoints or tools are exposed without authorization.
- **C4 (authorization lifecycle)**: N/A — no auth/jail/credential changes in this PR.
- **C5 (build ≠ runtime)**: N/A — no build artifacts, extensions, or relative-URL packaging in this PR.
- **C6 (governance/policy)**: N/A — no governance or policy documents changed.

## Data-flow trace summary

- **`replaceWithRegistryResult` fields map** → populated from `event.Data` (server-authored at access-request creation) → marshaled to `json.RawMessage` → passed to `updater.ReplaceView` → `RenderCard` → JSON Schema validation → `v3.Build` unmarshals to `v3Fields` → renders card body. All field values traced to upstream server-authored sources. ✅
- **`CardSeq`** → `event.EventID` (snowflake, monotonically increasing) → passed as `target.CardSeq` to mutator → `casWriteOnce` compares against `stored.CardSeq` via `FOR UPDATE SELECT`. Race window eliminated. ✅
- **Callback envelope** → `marshalDecisionRequest` reconstructs from `event.Card` (template identity) + `event` (callback request) → routed by `CallbackFormat` field on the dispatch route. Legacy fallback when `CardContext` is absent. ✅

yujiawei (approved):
# Code Review — PR #641 (octo-server)

Re-review at head `b8d13a2e`. This head is a fix commit ("address PR #641 review — P1 truncation + P2 hardening") layered on the previously reviewed tree; the underlying feature diff vs merge-base `4b8cd2d3` (+2691/-101, 39 files) is unchanged. All five findings from the prior round are addressed and independently verified below. Frozen `docs.access-request@0.2.0` handoff tree and `v2.go` remain byte-identical to the base.

## 1. Spec compliance

**Spec: ✅**

- Missing scope: none. The PR delivers exactly the post-#633 interactive-loop closure named in the linked spec brief — A1 `CardUpdater` (`ReplaceView`/`Append`), A2 route-versioned callback envelope (legacy default / opt-in `octo-card-v1`), A3 `docs.access-request@0.3.0` result view beside the frozen `0.2.0`, A4 two bounded metrics.
- Over-scope: none. No new endpoints/columns/flags beyond the declared surface; `response_url` is reserved (not emitted); `Append` is defined but has zero production callers (wired only in tests).
- Deviation: none. All write authority (sender / message ID / channel / Space / `card_seq`) is re-derived from stored state, never from callback display fields; the frozen `0.2.0` stays registered so in-flight pending cards keep working.

## 2. Code quality

**Quality: Approved**

Prior-round findings — all resolved and verified against the current head:

**(P1, blocking → fixed) Display-field truncation gap in `modules/notify/action_finalizer.go`.**
Previously, callback `Display` / `event.Data` values (legal up to 500 runes per the dispatch contract, `internal/cardactiondispatch/contract.go:226`) were copied into the `0.3.0` result schema (`maxLength` 80–200) without truncation for all but `denyReason`. An in-contract oversized value would fail `InputSchema.Validate` (`ErrFieldsInvalid`) inside `ReplaceView` → `RenderCard`, deterministically stalling the terminal update **and** the applicant notification (the `Send` is sequenced after a successful `ReplaceView`), leaving the approver card live in `pending` with no terminal state.
Now `replaceWithRegistryResult` truncates every affected field to its exact schema cap before marshal/render: `title`/`sourceName`→200, `requester.name`/`operatorName`→120, `decidedAt`/`requestedAt`/`messageTime`/`permission.*`→80, `requestReason`/`rejectionReason`→300 (`action_finalizer.go:141-155`, `:180-221`). Caps match `contract/data.schema.json` one-for-one; the jsonschema validator counts runes, so no off-by-one. `TestDocsActionFinalizerV3TruncatesOversizedDisplayFields` drives a real renderer with oversized fields and asserts the mutation still persists. The untruncated fields (`requestId`/`docId` cap 200, `avatarUrl` cap 2048) are server-authored — baked into `Action.Submit.data` at pending-render time and already bounded by the identical pending-schema caps — so leaving them untouched is correct; truncating a URL would corrupt it. Verified.

**(P2 → fixed) Corrupt `metadata.octo` fail-closed.** `pkg/cardmsg/interactive.go` `HasCardTemplateMetadata` now returns `true` when `metadata.octo` is present but a non-object, so a corrupt template identity fails closed (callback rejected) instead of degrading to a legacy flat body. Cards with `metadata.octo` absent still return `false` and keep the legacy zero-context path — no regression. Verified.

**(P2 → fixed) Envelope/route owner cross-check.** `resolveRegistryCardContext` now asserts the resolved template's `ActionContract.Owner` equals the route owner drawn from `actionData["owner"]` — the same source `cardActionRouteMetadata` uses for routing — rejecting mismatches before enqueue (`api_card_action.go:316-331`). This blocks a hand-crafted card that declares a docs template identity while routing to a different owner. Verified.

**(P2 → fixed) `card_seq` monotonicity / Append semantics documented.** `UpdateTarget.CardSeq` now distinguishes `ReplaceView` (strictly greater) from `Append` (exactly consecutive), and the `event.EventID → card_seq` monotonicity invariant is documented at the `ReplaceView` call site with a maintenance note. Verified.

Verification performed: `go build` and `go vet` clean on all touched packages; `pkg/cardtmpl`, `pkg/cardtmpl/docs_access_request`, `pkg/cardmsg`, `internal/cardactiondispatch`, `internal/carddispatch` all green (`-race` on `cardtmpl`); `modules/notify` finalizer/v3 suites and `modules/message` registry-card-context suite green. A `modules/message` failure in `TestChannelFiles_WithFiles` is pre-existing (reproduces on merge-base `4b8cd2d3`, and the test is untouched by this PR) and unrelated to the card path.

## 3. Overall verdict

**APPROVED** — Spec ✅ and Quality Approved, zero blocking findings.

## 4. Non-blocking notes (nits)

- `action_finalizer.go:187` comment on `avatarUrl` states the schema has "无 maxLength"; the result schema actually sets `maxLength: 2048` (with an `https` pattern). The decision not to truncate a URL is correct — only the comment's justification is slightly inaccurate. Cosmetic.

## 5. Additional observations (base-hardening, not blocking)

- The owner cross-check is guarded by `if contract := tmpl.Meta().ActionContract; contract != nil`. A future interactive template whose manifest omits `owner` or `action_type` would have `ActionContract == nil` and thus skip the owner assertion. No such registered template exists today (the current registry derives a contract only when both fields are present), so this is latent hardening for future runtime-derived cards rather than a live gap — worth a guard/test if the registry later admits owner-less interactive templates.

---

Review methodology note: this verdict fuses an independent line-level pass (build/vet/tests + diff trace) with a cross-family adversarial pass, which independently reached the same conclusion (fixes complete, no P1/P2). A second long-context advisory pass timed out and did not return output; it is recorded as absent, not as tacit agreement.

Issue #254, opened 2026-06-04
Soft-delete incoming webhooks so deleted webhooks' historical messages still render


an9xyz:
## Background

#250 added a server-side fallback so incoming-webhook synthetic senders (`iwh_` prefix) render correctly on the channel / user / avatar endpoints instead of erroring, and degrades gracefully when a webhook no longer exists (not-found / default avatar).

However `delete` is a hard `DELETE` (`deleteByWebhookID`). Once a webhook is deleted its `name`/`avatar` row is gone, so the display endpoints can only degrade — historical messages from that webhook lose their sender name and avatar (the client can no longer resolve them).

## Proposal: soft-delete webhooks

Keep the `incoming_webhook` row on delete and mark it deleted (e.g. reuse the `status` column with `2 = deleted`). The existing lifecycle gates already line up:

| Gate | Today | After soft-delete |
|---|---|---|
| push auth (`api.go` push handler, `m.Status != 1`) | rejects `Status != 1` | `status=2` → push auto-rejected, token dead (no code change) |
| display datasource (`queryByWebhookID`) | does not filter by status | deleted rows still resolve → endpoints return real name/avatar |
| management list (`queryByGroupNo`) | `WHERE group_no=?` | add `AND status != 2` to hide deleted |
| quota (`insertWithQuota` count) | `WHERE group_no=?` | add `AND status != 2` to free quota on delete |
| update (status validation accepts 0/1) | accepts status 0/1 only | also reject operating on a `status=2` row (no revive) |

No DB migration needed (reuses the existing `status` int column).

## Acceptance criteria

- After deleting a webhook, `GET /v1/channels/{iwh}/:type`, `GET /v1/users/{iwh}`, and `GET /v1/users/{iwh}/avatar` still return the real name / avatar.
- Push with the deleted webhook's token fails (already covered by the status gate).
- The webhook disappears from the management list and frees its per-group quota.
- A deleted webhook cannot be revived via update.
- Update the two "post-delete degradation" tests from #250 to assert "post-soft-delete still returns real data".

## Alternative / long term

Clients should ultimately render the webhook sender from the message's embedded `payload.from.name/avatar` (the authoritative, deletion-proof source — this is how Discord works). Soft-delete is the server-side complement that keeps the fallback endpoints correct for not-yet-adapted clients.

Follow-up to #246 / #250.

PR #642, opened 2026-07-23
feat(incomingwebhook): per-scope webhook quota with per-thread & group-total caps

### LINKED ISSUE #640: ✨ Incoming Webhook 数量限制改为按子区粒度，而非按群

an9xyz:
## Summary

Make the Incoming Webhook creation quota **per delivery scope** `(group_no, thread_short_id)` — the group itself and each thread (子区) get **independent** budgets instead of sharing one per-group cap — and add two knobs so operators can **precisely control** the counts:

1. **Per-scope re-scoping** — a group with many threads (and **Octo Loop**, which provisions a webhook per thread) is no longer blocked by the single shared 10-cap.
2. **`max_per_thread`** — a per-thread cap **decoupled** from the group-self `max_per_group`, so you can set e.g. group 10 / each thread 3. Unset → falls back to `max_per_group` (unchanged behavior until configured).
3. **`max_total_per_group`** — a group-wide **aggregate ceiling** across the group and all its threads; `0` = disabled (default). Bounds the `max_per_scope × (threads+1)` growth that per-scope quotas alone allow, since thread creation itself is uncapped.

The `FOR UPDATE` serialization lock deliberately stays on the parent `group` row; only the `COUNT(*)` predicates narrow/add. All three quota layers are evaluated inside that **one** critical section, so the change is race-exact under concurrency (no new lock, no lock-scope change). No schema/data migration.

> **Supersedes #639** — closed and re-opened as a fresh PR to get a clean review round (the previous PR's approvals were dismissed across several rebases/pushes and two automated reviewers had stalled). **Identical head `a0b00670`, same diff**; #639's review history has the full prior discussion (all rounds APPROVE, zero blocking).

## Release Note

> **Incoming Webhook quota is now per delivery scope, with two new admin knobs.**
>
> - **Behavior change** — `incomingwebhook.max_per_group` (default 10) and `incomingwebhook.max_per_creator` (default 5) now apply **per delivery scope** (the group itself and *each* thread each get the cap independently) instead of being shared across the group and all its threads. Existing deployments that tuned these values get a **more permissive** effective limit for groups that have threads — review your setting if you relied on the old shared cap.
> - **New** `incomingwebhook.max_per_thread` (env `DM_INCOMINGWEBHOOK_MAX_PER_THREAD`) — per-thread cap, decoupled from `max_per_group`. **Unset → falls back to `max_per_group`**, so no change until you set it.
> - **New** `incomingwebhook.max_total_per_group` (env `DM_INCOMINGWEBHOOK_MAX_TOTAL_PER_GROUP`) — group-wide ceiling across the group + all threads. **Default `0` = disabled**, so no aggregate limit exists unless you opt in.
> - No schema/data migration; existing webhooks are unaffected.

## Related Issue

Closes #640 — 「Incoming Webhook 数量限制改为按子区粒度，而非按群」(反馈人：李宁；Octo Loop 场景下每群 10 个的共享上限成为瓶颈).

## Linked Spec

`.octospec/tasks/incoming-webhook-quota-per-thread/brief.md`

Supersedes the `incoming-webhook-thread` task's locked decision ("threads share the parent group's `max_per_group`; no per-thread quota"), per issue #640.

## Changes

- **`modules/incomingwebhook/db.go`** — `insertWithQuota` takes `quotaLimits{scope, perCreator, total}`. The per-scope + per-creator counts are folded into one conditional-aggregation query (`count(*)` + `count(CASE WHEN creator_uid=? …)`) over the `(group_no, thread_short_id, status)` index range — shortening the parent-group `FOR UPDATE` critical section; the aggregate `total` count (group-wide, no thread filter) stays a separate `total>0`-gated query. The parent-group lock is unchanged.
- **`modules/common/system_settings.go` + `system_setting_schema.go`** — new settings `incomingwebhook.max_per_thread` (Positive; falls back to `max_per_group`) and `incomingwebhook.max_total_per_group` (non-Positive, `0`=disabled, range `[0,3650]`, negative→0). `max_per_group` reinterpreted as the group-self scope cap.
- **`modules/incomingwebhook/api.go`** — the create handler picks `max_per_group` for a group webhook or `max_per_thread` for a thread webhook, and passes the aggregate `max_total_per_group`.
- **`pkg/errcode/incomingwebhook.go` + i18n** — new `ErrIncomingWebhookTotalQuotaExceeded` (409 `mgmt_total_quota_exceeded`); the two existing quota messages generalized to "group or thread"; en-US markers + zh-CN regenerated.
- **`modules/incomingwebhook/README.md`** — quota section updated to per-scope + the two knobs.
- **Tests** — DB-level per-scope independence (group↔thread, thread↔thread) + per-creator per-scope + aggregate ceiling (empty scope still blocked, soft-delete frees a slot, `total=0` disables); settings fallback/override/clamp; HTTP group-vs-thread distinct caps; a concurrent aggregate-ceiling test; and the hardened `TestCreate_QuotaConcurrent` scope query.
- **`.octospec/`** — task brief, context, journal, pending learning (lock-vs-count-scope), log.

## Testing

- [x] **Integration tests green under `-race`** locally (MySQL 8.0 + Redis + WuKongIM) and in CI (Test job on `a0b00670`): `go test -race ./modules/common/ ./modules/incomingwebhook/`, including the concurrent per-scope and aggregate-ceiling tests.

Static gates also green: `go build`, `go vet`, `golangci-lint run` (0 issues), `make i18n-extract-check`, `make i18n-lint`.

## COMPREHENSION

   In the `insertWithQuota` transaction, the quota is evaluated inside a single `SELECT id FROM group … FOR UPDATE` critical section: one folded read gives the per-delivery-scope count `(group_no, thread_short_id)` and the per-creator count in that scope; an optional group-wide count `(group_no)` enforces `max_total_per_group`. Lock scope (whole group) and count scope (per-target / group-wide) are intentionally decoupled: lock coarse, count fine.

   Only the counting changes; both the user face and the bot face funnel through the same `insertWithQuota`, so they move together. The main hazard would be narrowing the `FOR UPDATE` lock to a per-thread range — an empty range takes a pure gap lock and concurrent first-inserts deadlock (InnoDB 1213). This PR avoids it by keeping the parent-group row lock, and neither the aggregate count nor the query-fold adds a lock. Backward-compat: existing rows carry `thread_short_id=''` and `max_per_thread`/`max_total_per_group` default to "same as group" / "disabled", so nothing changes until an operator opts in.

   `TestThread_QuotaIsPerScope` / `TestThreadWebhookQuotaIsPerScope` (independence), `TestTotalQuotaCeiling` (aggregate ceiling blocks even a fresh empty scope; soft-delete frees a slot; `total=0` disables), `TestThread_MaxPerThreadDistinctFromGroup` (group cap 2 vs thread cap 1), and **`TestThread_TotalQuotaCeilingConcurrent`** (12 goroutines across group + 2 threads under a ceiling of 4 → exactly 4 succeed, group total == 4). All pass locally under `-race` against real MySQL/Redis/WuKongIM, and in CI.

## Checklist

- [x] Updated documentation (module README, setting descriptions, octospec journal/log)

https://claude.ai/code/session_01QqpYSJX4M73HBFKmJ1CQkQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01QqpYSJX4M73HBFKmJ1CQkQ)_


lml2468 (approved):
# PR Review — PR #642 (Mininglamp-OSS/octo-server)

Reviewed at head `a0b00670` against merge-base `d3041765` (base `main`).
**Build/vet/test:** re-run on this checkout — `go build ./modules/incomingwebhook/ ./modules/common/` exit 0 ✅, `go vet` ✅. Quota tests DB-gated (MySQL) → CI.

## ⚠️ Duplicate of PR #639
This PR is a **byte-identical duplicate** of #639, which I reviewed and **APPROVED** (most recently at this same head `a0b00670`):
- **Same head SHA** `a0b00670ffef`, same branch `claude/incoming-webhook-quota-per-thread`, same commit chain (per-scope quota → `max_per_thread`+`max_total_per_group` → the `bd719c0e` count-fold → merge main), same +981/-96.
- `git diff a0b00670(#639) HEAD` = **empty** (identical tree); `modules/incomingwebhook/db.go` is byte-for-byte the #639-approved code.

There is no code difference to re-review — the #639 verdict transfers verbatim. **Recommend the author close one of #639/#642** to avoid a double-merge / split approval count.

## Verdict (carried from #639, re-verified on this checkout)
**APPROVE** — the incoming-webhook creation quota is re-scoped from per-parent-group to **per delivery scope** `(group_no, thread_short_id)` (group-self and each thread get independent budgets), plus a decoupled `max_per_thread` cap and an optional group-wide aggregate ceiling `max_total_per_group` (default 0 = disabled). All verified on this checkout:
- **Concurrency correctness (the load-bearing property):** all quota counts run inside the single parent-group `SELECT id FROM group … FOR UPDATE` critical section — the lock deliberately stays coarse (parent group row) while only the counts narrow, avoiding the empty-scope gap-lock 1213 deadlock (#31 lesson). The `bd719c0e` optimization folds the scope + per-creator counts into one `count(CASE …)` conditional aggregate (identical values, `count(CASE)` not `SUM(bool)` to keep BIGINT), and the whole-group total count is deliberately left separate. Error precedence scope→creator→total preserved.
- **Index-backed** on `idx_incoming_webhook_thread(group_no, thread_short_id, status)` (total uses the `group_no` prefix); **scope isolation** via upstream-validated `ThreadShortID`; **back-compat defaults** (per_thread→per_group=10, total=0 disabled, per-creator=5); **409/i18n/schema** complete; **no migration**.
- **Tests:** per-scope independence, per-thread vs group decoupling, and `TestThread_TotalQuotaCeilingConcurrent` (K concurrent creates → exactly K succeed, group total == K) — DB-gated, must run in CI.

Both gates pass. Suggested decision: `Safe to merge` (identical to #639) once CI runs the DB-gated quota suite — **but first consolidate the #639/#642 duplicate** (keep one; they share the branch and head, so merging one makes the other a no-op, but two open mergeable PRs for the same change should be closed to one). (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (approved):
The quota implementation is race-safe, correctly scoped, and aligned with the repository architecture. Approving.

## ✅ Correctness review

- **Race safety / TOCTOU**: All three quota layers (per-scope, per-creator, group-total) are evaluated inside the single `SELECT id FROM group ... FOR UPDATE` critical section in `insertWithQuota` (`modules/incomingwebhook/db.go`), so check-then-insert is atomic per group. Concurrent same-group creates serialize on the parent-group row — no bypass window.
- **Lock vs. count scope**: Lock stays on the (always-present) parent-group row while counts narrow to `(group_no, thread_short_id)`. This intentionally avoids the empty-scope gap-lock deadlock (InnoDB 1213) that a per-thread-range `FOR UPDATE` would cause on a first insert. Correct call.
- **Boundary operators**: Comparisons use `count >= limit → reject`, which permits exactly `limit` rows in each dimension — no off-by-one.
- **Counting consistency**: Scope, per-creator, and group-total counts all exclude only soft-deleted rows (`status != statusDeleted`); disabled webhooks still occupy quota (prevents disable→create hoarding). Soft-delete frees a slot. Consistent across all three.
- **Query folding**: Per-scope + per-creator counts folded into one conditional-aggregation query (`count(*)` + `count(CASE WHEN creator_uid=? ...)`) over the same index range, shortening the locked critical section. The group-total count stays a separate query, correctly gated behind `total > 0`.
- **Settings fallback / clamp** (`modules/common/system_settings.go`): `max_per_thread` falls back dynamically to `max_per_group` (byte-identical to pre-feature behavior when unset); `max_total_per_group` clamps negative→0 so `0` remains a meaningful "disabled" sentinel; scope/per-creator caps clamp `<=0`→default to avoid a hard-to-diagnose "dark off" misconfig.
- **Caller** (`modules/incomingwebhook/api.go`): create handler picks `max_per_group` for group scope vs. `max_per_thread` for thread scope, applies per-creator only for non-admins, and passes `max_total_per_group`; each quota error maps to a distinct 409 envelope.
- **No schema/data migration**; existing rows carry `thread_short_id=''` and the new knobs default to "same as group" / "disabled". Auth and scope isolation are unchanged.

## 💬 Non-blocking notes

- 🟡 The README quota prose section in `modules/incomingwebhook/README.md` still reads as a two-layer quota (`max_per_group` + `max_per_creator`) and does not mention the new `max_per_thread` or `max_total_per_group` knobs (or their env vars) in that paragraph. Worth extending it to document all three layers for operators.
- 🔵 The doc comment on `IncomingWebhookMaxPerGroup` in `modules/common/system_settings.go` says each thread uses that limit, but a configured thread uses `max_per_thread`. Minor wording clarification.

## ✅ Highlights

- Parent-group row locking serializes same-group creation while sidestepping empty-range gap-lock deadlocks.
- Conditional aggregation minimizes queries inside the critical section.
- Strong test coverage: per-scope independence (group↔thread, thread↔thread), per-creator per-scope, aggregate ceiling (empty scope still blocked, soft-delete frees a slot, `total=0` disables), settings fallback/override/clamp, HTTP group-vs-thread distinct caps, and a concurrent aggregate-ceiling test under `-race`.
- No security, isolation, error-envelope, or blocking correctness issues found.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #642 (octo-server)

Reviewed at head `a0b00670`. Scope: per-scope webhook quota with per-thread + group-total caps, plus three merged feature lines from `main` (cardtmpl interactive-card closure v0.3.0, message-reaction byte-exact emoji collation, octospec docs).

## Summary

Solid, well-documented PR. The headline quota change lifts webhook counting from one group-wide pool to three atomic layers evaluated under a single parent-group row lock: per-delivery-scope `(group_no, thread_short_id)` (using `max_per_group` for the group itself and the new `max_per_thread` for each thread), per-creator in the same scope, and an optional group-wide ceiling (`max_total_per_group`, 0 = disabled). The lock-granularity vs. count-granularity split is correctly reasoned about in `modules/incomingwebhook/db.go:22-48` (gap-lock deadlock avoidance on empty scopes). The three sibling features merged from `main` — registry-backed 0.3.0 result view, `octo-card@1.0` callback envelope, and `utf8mb4_bin` emoji collation — are each properly gated and tested.

## Verification

- ✅ **Quota atomicity** — scope count + per-creator count folded into one conditional-aggregate query (`modules/incomingwebhook/db.go:102-109`); total count runs as a second query in the same `FOR UPDATE`-serialized transaction. No check-then-act gap.
- ✅ **Lock target correct** — `SELECT ... FROM \`group\` WHERE group_no=? FOR UPDATE` still locks the unique parent-group row; `thread_short_id` only narrows the count predicate, not the lock, so concurrent creates across threads within the same group are correctly serialized.
- ✅ **Per-creator scope consistent** — `creator_uid` count now also filters by `thread_short_id` (same aggregate query), matching the README and the acceptance "per-scope" contract. Admins still bypass (`api.go:815-818`).
- ✅ **Read-side clamp for new knobs** — `IncomingWebhookMaxPerThread` falls back to `IncomingWebhookMaxPerGroup` on ≤0 (tests pin the default); `IncomingWebhookMaxTotalPerGroup` clamps negatives to 0 (the "disabled" sentinel, tests pin this too).
- ✅ **Reaction emoji migration safe** — `utf8mb4_general_ci → utf8mb4_bin` on the `emoji` identity column is forward-only with the unique index automatically rebuilt; API layer now uses the persisted `toggleResult.Emoji`/`Seq` instead of echoing the request, so byte-exact storage reaches the client.
- ✅ **Card action ingress fails closed** — `resolveRegistryCardContext` (`modules/message/api_card_action.go:296-335`) rejects cards with partial template metadata (`HasCardTemplateMetadata`) and cross-validates that `metadata.octo.template` owner matches the submit-action `data.owner`, blocking hand-crafted cross-route envelopes.
- ✅ **Finalizer truncation vs. schema caps** — per-field rune caps in `replaceWithRegistryResult` (`modules/notify/action_finalizer.go:140-143`) mirror `handoff/docs.access-request@0.3.0/contract/data.schema.json`, closing the PR#641 P1 where `ErrFieldsInvalid` left approval cards permanently pending.
- ✅ **Callback format backward compatibility** — `marshalDecisionRequest` falls back to the flat `DecisionRequest` when `event.Card == (CardContext{})` even on `octo-card-v1` routes, so pre-registry cards in flight are not rejected as incomplete.
- ✅ **Concurrency tests** — `TestThread_TotalQuotaCeilingConcurrent` fan-out of 12 across 3 scopes with `max_total_per_group=4` asserts exactly 4 succeed and the group-wide row count equals the ceiling; `TestThread_QuotaIsPerScope` and `TestThread_MaxPerThreadDistinctFromGroup` pin the per-scope and decoupled-cap contracts.

Static analysis only at head `a0b00670ffeff43302c305a71d5b8ba62ae05753`; build and tests not executed in this environment.

## Findings

No P0/P1 issues; one P2 and one nit below.

### P2 — `max_per_thread` silently promotes an explicit 0 to the `max_per_group` default (`modules/common/system_settings.go:656`)

`IncomingWebhookMaxPerThread` clamps `v <= 0` to the env/default (which itself falls back to `max_per_group`). This matches the existing `max_per_group` behaviour (comment at `:651` calls it a defence against a "dark switch"), but the two new knobs now disagree on what 0 means: `max_total_per_group` treats 0 as a meaningful "disabled" sentinel, while `max_per_thread` treats 0 as "use the group default". An operator who writes `incomingwebhook.max_per_thread = 0` to the DB — intending "no thread webhooks" — silently gets `max_per_group` (default 10) instead. Schema description in `system_setting_schema.go:120` reads "未配置时与群本体 max_per_group 一致", which does not cover the explicitly-written-0 case. Worth either adding a distinct disabled sentinel (e.g. -1 = disabled, 0 = "no webhooks allowed in threads") or extending the admin-facing description and a pinning test that demonstrates the "0 → fallback" path so operators can't be surprised.

### Nit — Legacy `NewDocsActionFinalizer` constructor silently degrades the v3 result path (`modules/notify/action_finalizer.go:47`)

`NewDocsActionFinalizerWithUpdater` is the only constructor that wires the registry result view; the legacy `NewDocsActionFinalizer` (still used by `modules/notify/action_finalizer_test.go` and `internal/cardactiondispatch/orchestration_integration_test.go`) leaves `f.updater == nil`, so terminal Approved/Denied events silently take the legacy `Mutate` envelope branch at `:100` rather than the registry result view. Production is fine — only `NewDocsActionFinalizerFromContext` is wired at runtime (`modules/notify/action_finalizer_registry.go:12`), and it always sets the updater — but future test authors copying the old constructor will get a silently-degraded path. Consider deprecating the old constructor, gating it behind `//go:build test_helpers`, or routing the affected tests through `NewDocsActionFinalizerWithUpdater` so the registry path is actually exercised.

## Human-verify

1. **Webhook total can grow with thread count by default.** With `max_total_per_group` defaulted to 0 (disabled) and `max_per_thread` falling back to `max_per_group` (default 10), a group with *N* active threads can now host up to `10·(N+1)` webhooks by default — a meaningful expansion from the pre-PR group-wide cap of 10. Any member who can create threads can inflate the group's webhook footprint. The product decision is documented in the README, but operations should decide whether to ship a non-zero `max_total_per_group` for high-thread-count deployments. **Not a merge blocker** — flagging for deployment-policy confirmation.
2. **Migration rollout order.** `20260722000001_message_reaction_emoji_binary.sql` is forward-only and auto-rebuilds the unique index on `emoji`. If any production row was stored with an emoji whose byte form differs from what the client/UI last sent (e.g. normalization drift under `utf8mb4_general_ci`), the rebuilt index keeps the byte-distinct rows but the toggle SELECT at `db_message_reaction.go:65` — now byte-exact — will miss the old row and the next toggle will insert a duplicate until the client re-reacts with the canonical byte form. Worth a one-time audit of `reaction_users` distinct-emoji counts pre-migration. **Not a merge blocker** — flagging for DBA awareness.

- Scope-count + per-creator-count conditional aggregate shares one index scan (`(group_no, thread_short_id, status)`); `COUNT(CASE)` returns BIGINT and avoids the `SUM(bool)` driver-conversion pitfall noted in the comment.
- `ErrTotalQuotaExceeded` is a new sentinel with its own 409 errcode, i18n key, and handler; `api.go:824-832` switch routes each sentinel to its renderer.
- `updater.ReplaceView` requires `CardSeq > 0` and `env.SpaceID == target.Target.SpaceID`; finalizer feeds it `event.EventID` (monotonic snowflake) with a comment flagging the CAS-monotonicity dependency.
- `updater.Append` enforces `target.CardSeq == snapshot.CardSeq+1` (consecutive) in addition to the mutator's strictly-increasing CAS; space_id is re-validated against the stored envelope to prevent cross-space append.
- `CardTemplateContext` extraction from the effective server-authored frame (`pkg/cardmsg/interactive.go:27-49`) rejects partial/whitespace-padded `id`/`version` and requires `protocol == "octo-card@1.0"`; `HasCardTemplateMetadata` returns `true` for a corrupted non-map `metadata.octo`, failing closed on the ingress.
- `recordTemplateCallback` short-circuits when `TemplateID`/`TemplateVersion`/`ActionID` are absent, so legacy (pre-registry) events do not emit unbounded-label metrics.
- `callbackResult` normalizes unknown result strings to `"error"`; `updateResult` normalizes non-"ok" to `"error"` — no cardinality explosion.
- `docs_action.go` outcome body fragment builder (`BuildDocsApprovalOutcomeBodyWithLang`) validates `ActorAvatar` via `requireHTTPS`, matching the existing `Source.IconURL` check.
- v3 template's pending-state Build delegates to `New()` then decorates each `Action.Submit` `data` map with the v3-only context fields (`actor_avatar_url`, `request_reason`, etc.) — these ride back in on `event.Data` when the action fires and are consumed by `replaceWithRegistryResult`.
- `MarshalCallbackRequest` returns a hard error when `octo-card-v1` is selected but `event.EventID <= 0` or `event.Card.{TemplateID,TemplateVersion,View}` is empty — no silent fallback to legacy on opted-in routes.

## Verdict: APPROVED

Quota enforcement is atomic and correctly scoped; the three sibling features are well-gated with fail-closed contracts; the regression tests pin the concurrency and per-scope contracts. The P2 and nit are non-blocking.

[Octo-Q] verdict: APPROVE — no P0/P1; one P2 (max_per_thread 0-clamp semantics) + one nit (legacy finalizer constructor silent degradation) are non-blocking. Recommend human-verify items on deployment policy (default total ceiling) and emoji-migration rollout.

yujiawei (approved):
# Code Review — PR #642 (octo-server)

**Verdict: APPROVED** · security-sensitive · reviewed at head `a0b00670`, merge-base `d3041765`.

This PR re-scopes incoming-webhook creation quota from per-parent-group to per-delivery-scope `(group_no, thread_short_id)` and adds two admin knobs (`max_per_thread`, `max_total_per_group`). It supersedes #639 but is **not** an identical diff — it adds genuinely new load-bearing quota logic (commits `1dc621e4` + `bd719c0e`), so it was re-reviewed from scratch rather than carrying the prior approval forward.

## 1. Spec compliance

**Spec: ✅** — matches `.octospec/tasks/incoming-webhook-quota-per-thread/brief.md`.

- **No under-build:** per-scope re-scoping (both `count` predicates gain `AND thread_short_id=?`), `max_per_thread` (falls back to `max_per_group` when unset), `max_total_per_group` (0=disabled), new `ErrIncomingWebhookTotalQuotaExceeded` (409) with en-US markers + zh-CN, schema/setting docs, README — all present.
- **No over-build:** no new lock, no lock-scope change, no schema/data migration; `max_per_creator` intentionally keeps a single value applied per-scope (no per-thread-creator knob), consistent with the brief's revised decision A.
- **No deviation:** the `FOR UPDATE` lock deliberately stays on the parent `group` row (`db.go:86-90`); only the `COUNT` scope narrows/adds. `thread_short_id` used in the count is `m.ThreadShortID`, bound and validated against the path `group_no` at `api.go:725-761` and immutable post-create — never free request input, so no cross-group/cross-scope budget leakage.

## 2. Code quality

**Quality: Approved** — no P0/P1.

Verified load-bearing facts:

- **Lock/count decoupling is correct (no gap-lock deadlock).** The single critical section still locks the always-present parent-group row via the unique index (`db.go:86-90`); narrowing the lock to the empty `(group_no, thread_short_id)` range would reintroduce the InnoDB gap-lock → insert-intention deadlock (1213) the design explicitly avoids. Confirmed unchanged.
- **Folded scope + per-creator count is correct.** `count(*)` + `count(CASE WHEN creator_uid=? THEN 1 END)` over the same `(group_no, thread_short_id, status)` index range (`db.go:104-113`). `COUNT(CASE)` counts only non-NULL, so `creator_cnt` is exactly the per-creator-in-scope count; using `COUNT(CASE)` over `SUM(bool)` also avoids the DECIMAL→int driver-conversion pitfall. One query replaces two on the member path; admin path unchanged — zero regression.
- **Aggregate ceiling is race-exact and correctly gated.** The `total>0` group-wide count (`db.go:120-135`, no thread filter) runs inside the same parent-group lock, so it is precise under concurrency — a fresh empty scope is still blocked once the group total is reached. `total<=0` skips the extra query entirely.
- **Settings validation is sound.** `max_per_thread` is `Positive` (write path rejects ≤0; read side `≤0 → dynamic fallback to max_per_group`). `max_total_per_group` is intentionally non-Positive so `0` is a meaningful "disabled" sentinel: admin write path bounds it to `[0,3650]` (`api_manager_system_setting.go:231`), read side clamps `<0 → 0`. `0`=disabled semantics are distinct from the `max_per_group=0` "dark switch" hazard and handled explicitly.
- **Backward compatibility exact.** Existing rows carry `thread_short_id=''` → group-self bucket; `max_per_thread` unset → equals `max_per_group`; `max_total_per_group` default 0 → disabled. No behavior change until an operator opts in.
- **Static gates green locally:** `go build`, `go vet` clean; i18n markers present in both `active.en-US.toml` and `active.zh-CN.toml` with correct group-or-thread wording.
- **Tests are well-targeted:** DB-level per-scope independence, per-creator per-scope, aggregate ceiling (empty-scope still blocked, soft-delete frees a slot, `total=0` disables), settings fallback/override/clamp, HTTP group-vs-thread distinct caps, and a 12-goroutine concurrent aggregate-ceiling test asserting exactly-N success + no leak. (DB-integration tests require live MySQL/Redis/WuKongIM — not runnable in this sandbox; verified by source + InnoDB locking analysis, per the module's standing constraint. PR reports them green under `-race` in CI at this head.)

**Note on the `max_total_per_group` knob:** it directly mitigates the multiplicative-growth concern (`max_per_scope × (threads+1)` unbounded, since thread creation itself is uncapped) raised in the prior review round. Good, targeted fix.

## 3. Suggestions (non-blocking, P2 — do not block merge)

- **P2 — orphan webhooks count against the aggregate ceiling.** `DeleteThread` (`modules/thread/service.go:758`) does not cascade-disable/delete thread-scoped webhooks; disabled/orphaned rows still occupy quota (anti-hoarding is intentional, `#254`). With `max_total_per_group` enabled, high thread churn can permanently consume aggregate slots that only manual webhook deletion frees. This enabler is entirely pre-existing and out of this diff — flagged as an operational follow-up, not a blocker. Consider a thread-delete → webhook-cleanup path in a later change, or document the operator expectation.
- **P2 — TOCTOU on thread-active check (pre-existing).** The create-time thread existence/active check (`api.go:735-761`) is a non-transactional read; the insert tx locks only the parent group, not the thread. A thread archived in the tiny window still yields one enabled row — non-security (identical to the existing group-status TOCTOU, which the push path re-gates). Acknowledged in code comments.

## Reviewer coverage note

- **Advisory legs this round: both ABSENT** — one visiting reviewer timed out (SIGTERM); the other returned no inline findings. Absence is recorded as absence, **not** silent approval. This verdict rests on independent source + InnoDB-locking analysis at head `a0b00670`, plus the fact that the core `insertWithQuota` scope/lock logic is a verified superset of the prior fully-adjudicated round on the superseded PR.
- **Not independently executable here:** DB-integration and concurrency tests (need live MySQL 8.0 + Redis + WuKongIM 2.2.4 for CommunityTopic channels). Recommend a human confirm the CI `Test -race` job is green on `a0b00670` before merge.
- **Human verify (security-sensitive):** the aggregate-ceiling behavior under real concurrency, and that no deployment relied on the old shared per-group cap (this change is *more permissive* for groups with threads — called out in the release note).

Issue #627, opened 2026-07-21
Non-monotonic message_extra.version across replicas can permanently skip a card's terminal frame in delta-sync (card stuck on intermediate frame / "interrupted")


an9xyz:
## Background / Symptom

While investigating the Octo card "slow edit updates / interrupted" path, I found a **correctness hazard** (not a latency issue): under a **multi-replica deployment**, an interactive card's **terminal frame can be permanently skipped by delta-sync**, leaving the client stuck on an intermediate "processing…" frame until a full resync (reconnect / foreground) occurs. To the user this looks like the card is "stuck / interrupted".

This is not a plugin or web issue — the root cause is the allocation semantics of `message_extra.version` on the server.

## Root cause

Card edits do a full-payload overwrite of `message_extra` + a `CMDSyncMessageExtra` fan-out, and clients pull incrementally by a `version` cursor:

- **delta-sync read** takes increments in ascending version order:
  `modules/message/db_message_extra.go:110`
  ```sql
  ... where channel_id=? and channel_type=? and version>? order by version asc limit ?
  ```
- **writes** allocate `version` inside the row lock (`cardVersionInLockWrite` / the card_seq CAS path):
  `modules/bot_api/send.go` (`cardVersionInLockWrite`, ~`:1086-1094`, where the property is already documented in the comment)
  `internal/carddispatch/mutation.go` (`casWriteOnce`)

`version` comes from octo-lib's **per-process HiLo `GenSeq`** allocator. Therefore the invariant "allocation order == commit order" **holds only within a single process**.

Across replicas: replica A can grab a lower `version` for its frame yet commit it **after** replica B's higher-`version` frame. A client that has already synced up to the higher version then **permanently skips** that lower-version frame, because its cursor is `version > ?`. If the skipped frame happens to be the **terminal frame** (✅ done / ⚠️ interrupted), the card is left on the previous intermediate frame until a full resync.

The code comment itself already records this as a known property (see the comment above `cardVersionInLockWrite` in `send.go`, and the P1-2 review note). This issue tracks it as a standalone follow-up.

## Impact

- **Trigger conditions**: multi-replica octo-server + consecutive edit frames of the same card landing on different replicas. Single-replica deployments are unaffected.
- **Behavior**: the card stays on an intermediate "processing…" frame — looks "slow / stuck / interrupted", but it is actually frame loss, not latency.
- **Convergence**: only recovers on a full resync (WS reconnect, foreground/visibility resync).

[...]

PR #630, opened 2026-07-21
feat(message): transactional message_extra version allocator + schema (#627)

an9xyz:
## Summary

Introduce a per-storage-channel **transactional** version allocator for
`message_extra`, so delta-sync versions are unique and commit-ordered across
octo-server replicas. This PR lands the **schema + allocator core only**;
production stays inert (`mode=legacy`, delegating to the existing `GenSeq`) until
a separate operator cutover. No writer is switched over and no wire/CMD shape
changes in this PR.

## Related Issue

Part of #627 — **this PR does not close the issue.** It only lands the schema
and the inert allocator (mode=legacy). #627 is resolved only after the writer
cutover (PR-2) and the operator activation/cutover (later PRs) ship. Do not use
a closing keyword when merging.

## Linked Spec

`.octospec/tasks/message-extra-transactional-version/brief.md` (included in this
PR, with `plan.md` describing the staged PR-1..PR-5 rollout).

## Changes

- Migration `modules/message/sql/20260721000001_message_extra_version.sql`:
  - `octo_message_extra_channel_seq` — one hot row per storage channel.
  - `octo_message_extra_version_state` — DB-authoritative allocator mode/floor
    singleton, **seeded to legacy** so behavior is unchanged on deploy.
  - `message_extra (channel_id, channel_type, version)` composite index for the
    delta-sync read path (online `CREATE INDEX` on MySQL 8.0).
- New leaf package `internal/msgextraseq` (depends only on octo-lib, imports no
  `modules/*`, so all writers can share it without an import cycle):
  - `ReserveTx` reserves N versions under a per-channel row lock held until the
    caller commits → commit order == version order, unique per channel.
  - Allocator selected by the state row read `FOR SHARE` inside the caller's tx
    (the cutover drain barrier); `mode=legacy` delegates to the existing
    `GenSeq` (the single allowlisted call site).
  - Race-safe first-use bootstrap (max of floor / `message_extra` max / legacy
    `seq.min_seq`), `2^53-1` overflow guard, low-cardinality Prometheus metrics.
  - Returns `[]int64` (per-item versions) so callers stay mode-agnostic.

## Testing

Focused suite `./internal/msgextraseq/` against MySQL 8.0 (per-package, green;
`-race` clean). Covers: two-session serialization (second tx blocks on the
first's row lock), rollback isolation, cross-channel independence (no global
lock), batch contiguity + `version > cursor` pagination, first-use bootstrap
[...]

PR #644, opened 2026-07-23
feat(message): transactional message_extra version allocator + schema (#627)

an9xyz:
## Summary

Introduce a per-storage-channel **transactional** version allocator for
`message_extra`, so delta-sync versions are unique and commit-ordered across
octo-server replicas. This PR lands the **schema + allocator core only**;
production stays inert (`mode=legacy`, delegating to the existing `GenSeq`) until
a separate operator cutover. No writer is switched over and no wire/CMD shape
changes in this PR.

## Related Issue

Part of #627 — **this PR does not close the issue.** It only lands the schema
and the inert allocator (mode=legacy). #627 is resolved only after the writer
cutover (PR-2) and the operator activation/cutover (later PRs) ship. Do not use
a closing keyword when merging.

## Linked Spec

`.octospec/tasks/message-extra-transactional-version/brief.md` (included in this
PR, with `plan.md` describing the staged PR-1..PR-5 rollout).

## Changes

- Migration `modules/message/sql/20260721000001_message_extra_version.sql`:
  - `octo_message_extra_channel_seq` — one hot row per storage channel.
  - `octo_message_extra_version_state` — DB-authoritative allocator mode/floor
    singleton, **seeded to legacy** so behavior is unchanged on deploy.
  - `message_extra (channel_id, channel_type, version)` composite index for the
    delta-sync read path (online `CREATE INDEX` on MySQL 8.0).
- New leaf package `internal/msgextraseq` (depends only on octo-lib, imports no
  `modules/*`, so all writers can share it without an import cycle):
  - `ReserveTx` reserves N versions under a per-channel row lock held until the
    caller commits → commit order == version order, unique per channel.
  - Allocator selected by the state row read `FOR SHARE` inside the caller's tx
    (the cutover drain barrier); `mode=legacy` delegates to the existing
    `GenSeq` (the single allowlisted call site).
  - Race-safe first-use bootstrap (max of floor / `message_extra` max / legacy
    `seq.min_seq`), `2^53-1` overflow guard, low-cardinality Prometheus metrics.
  - Returns `[]int64` (per-item versions) so callers stay mode-agnostic.

## Testing

Focused suite `./internal/msgextraseq/` against MySQL 8.0 (per-package, green;
`-race` clean). Covers: two-session serialization (second tx blocks on the
first's row lock), rollback isolation, cross-channel independence (no global
lock), batch contiguity + `version > cursor` pagination, first-use bootstrap
variants + two concurrent initializers, `count`/`int64`/`2^53-1` bounds, and
and state handling: an unrecognized `mode` value or an unreadable row fails closed, while a **missing** row defaults to legacy (safe pre-migration behavior) and is guarded by the expected-mode check (`OCTO_MESSAGE_EXTRA_VERSION_EXPECTED_MODE`) so a transactional deployment fails closed on a lost row. Migration applies cleanly via the
real sql-migrate flow. `go build ./...`, `go vet`, `make i18n-lint`,
`make i18n-extract-check` all pass.

- [x] Manually verified (migration apply + allocator behavior on MySQL 8.0)

## COMPREHENSION

   It adds the storage tables and a shared allocator, but does **not** change any
   runtime write path yet: the seed `mode=legacy` makes `ReserveTx` delegate to
   today's `GenSeq`, and no writer calls the new allocator in this PR. The only
   live effect is two new empty tables and a new `message_extra` index. The
   transactional path (row-lock-held-to-commit ordering) is exercised by tests
   but gated off in production behind the DB state row.

   The `message_extra` index add runs on a large live table — mitigated by MySQL
   8.0 online `CREATE INDEX` (LOCK=NONE) and a low-traffic window. The tables are
   additive with no FKs, so rollback is a no-op drop. No code reads the state row
   in production yet, so a mis-seed cannot affect writes in this PR. The `channel_id`
   collation is pinned to match `message_extra` (`utf8mb4_general_ci`, verified).

   `TestReserveTx_TwoSessionSerialization` proves the second transaction blocks on
   the first's sequence-row lock and only gets the next value after commit;
   `TestReserveTx_Rollback` proves a rolled-back tx leaves neither the business
   row nor the sequence advance; `TestReserveTx_BatchContiguous` paginates a
   reserved range through `version > cursor` with no gaps/dupes;
   `TestReserveTx_ConcurrentInitializers` proves the first-use race yields unique
   versions with no duplicate-key leak. Migration verified to apply via
   sql-migrate on a clean MySQL 8.0 `test` schema.

## Checklist

- [x] Updated documentation (task brief/plan under `.octospec/`)

### DIFF
--- a/.octospec/tasks/message-extra-transactional-version/brief.md
+++ b/.octospec/tasks/message-extra-transactional-version/brief.md
@@ -1,0 +1,596 @@
+---
+type: Task
+title: "Task: message-extra-transactional-version"
+description: Make message_extra delta-sync versions unique and commit-ordered per storage channel across octo-server replicas.
+tags: ["message", "delta-sync", "wire-contract", "database", "migration", "concurrency", "multi-replica", "card", "bot-api", "observability", "testing"]
+timestamp: 2026-07-21T17:38:17+08:00
+# --- octospec extension fields ---
+slug: message-extra-transactional-version
+upstream: "Mininglamp-OSS/octo-server#627"
+source: self
+---
+
+# Task: message-extra-transactional-version
+
+> One task = one `.octospec/tasks/<slug>/` directory. This brief is the spec for
+> the work. AI may draft it from existing code; a human confirms it.
+
+## Goal
+
+Make `message_extra.version` a unique, commit-ordered cursor within each
+persisted channel so `/v1/message/extra/sync` cannot permanently skip a later
+committed mutation in a multi-replica deployment.
+
+The implementation introduces a message-extra-specific transactional sequence
+in octo-server. Every authoritative `message_extra` mutation reserves its
+version from the channel sequence and writes the business row in the **same
+MySQL transaction**, while holding the sequence-row lock until commit. This
+preserves the existing client wire contract (`extra_version`, ascending
+pagination, and `CMDSyncMessageExtra`) while fixing the ordering invariant at
+its source.
+
+## Background
+
+- Delta sync reads one storage channel with
+  `WHERE channel_id=? AND channel_type=? AND version>? ORDER BY version ASC
+  LIMIT ?` (`modules/message/db_message_extra.go`). Once a client advances its
+  cursor, a row later committed with `version <= cursor` is invisible to that
+  incremental path.
+- Current writers allocate versions through octo-lib `Context.GenSeq`. That
+  allocator keeps a process-local `seqMap` and reserves blocks of 1000 values;
+  the database stores only the block ceiling. Therefore allocation order is
+  not globally ordered across octo-server replicas.
+- The current block reservation is also a read-then-overwrite operation. Two
+  processes initializing the same key concurrently can reserve the same block,
+  so cross-process uniqueness is not guaranteed either. Duplicate versions are
+  unsafe for `version > cursor` pagination when a page boundary splits rows
+  sharing the same value.
+- Card edits already serialize competing writes to the same `message_id` and
+  allocate after taking that row lock. This closes the single-process race but
+  cannot close the cross-process HiLo range problem. `card_seq` protects which
+  content wins; it does not make the channel delta-sync cursor monotonic.
+- A concrete failure is: replica B commits an intermediate card frame with a
+  higher HiLo value; the client syncs it and advances its cursor; replica A then
+  commits the accepted terminal frame from an older block. The database stores
+  the correct terminal content, but the generic CMD carries neither content nor
+  version and the subsequent `version > cursor` read cannot return that row.
+- The defect is broader than cards. User/bot/robot edits, callback card
+  mutation, revoke/delete, pin changes, manager deletion, and read-receipt count
+  materialization all write the same cursor domain.
+- Merely replacing HiLo with MySQL `AUTO_INCREMENT` or Redis `INCR` is not a
+  root fix. Those primitives order allocation, not transaction commit: a
+  transaction holding a lower allocated value can still commit after a higher
+  value and be skipped.
+
+## Decisions
+
+### D1. Cursor invariant and scope
+
+For each persisted storage-channel key `(channel_id, channel_type)`:
+
+1. Every committed authoritative `message_extra` mutation has one version
+   strictly greater than every mutation that committed before it in that key.
+2. No two committed mutations in that key share a version.
+3. A transaction that rolls back exposes neither its business mutation nor its
+   sequence advance. Reuse of a rolled-back value is allowed because it was
+   never visible; gaps are also allowed and carry no protocol meaning.
+4. Ordering is per storage channel, not global. Personal chats use the existing
+   fake channel ID derived before persistence; group/topic channels use their
+   persisted channel ID and current `channel_type` unchanged.
+
+This is the invariant required by the existing channel-scoped delta cursor. A
+per-message monotonic value is insufficient because a client cursor can advance
+past that message due to a mutation on another message in the same channel.
+
+### D2. Transactional channel-sequence and allocator-state tables
+
+Add a message module migration creating the per-channel sequence table and one
+database-authoritative allocator-state row:
+
+```sql
+CREATE TABLE `octo_message_extra_channel_seq` (
+  `channel_id`   VARCHAR(100)     NOT NULL DEFAULT '' COMMENT '存储行频道ID(person=fakeChannelID)',
+  `channel_type` SMALLINT         NOT NULL DEFAULT 0  COMMENT '频道类型',
+  `last_version` BIGINT           NOT NULL DEFAULT 0  COMMENT '该频道已分配的最大 message_extra 版本',
+  PRIMARY KEY (`channel_id`,`channel_type`)
+) ENGINE=InnoDB DEFAULT CHARSET=<match message_extra.channel_id> COLLATE=<match> COMMENT='message_extra 每频道事务序列(#627)';
+
+CREATE TABLE `octo_message_extra_version_state` (
+  `singleton_id`  TINYINT UNSIGNED NOT NULL COMMENT '恒为1的单例键',
+  `mode`          TINYINT          NOT NULL DEFAULT 0 COMMENT '0=legacy,1=transactional',
+  `epoch`         BIGINT UNSIGNED  NOT NULL DEFAULT 0 COMMENT '换代计数(operator CAS 递增)',
+  `cutover_floor` BIGINT           NOT NULL DEFAULT 0 COMMENT 'cutover 校验后的版本下界',
+  PRIMARY KEY (`singleton_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci COMMENT='message_extra allocator 全局状态单例(#627)';
+
+INSERT INTO `octo_message_extra_version_state`
+  (`singleton_id`, `mode`, `epoch`, `cutover_floor`)
+VALUES (1, 0, 0, 0);
+```
+
+- Table names use the `octo_` prefix required for all new tables in this repo
+  (existing prefix-less tables such as `message_extra` are not touched); the
+  migration file follows `modules/message/sql/<yyyyMMdd>-<seq>_<name>.sql`,
+  embedded via `//go:embed sql`, and matches the repo DDL style of the most
+  recent new table (`octo_message_card_revision`): inline `PRIMARY KEY`/`KEY`,
+  per-column `COMMENT`, and an `ENGINE=InnoDB DEFAULT CHARSET=... COLLATE=...
+  COMMENT=...` suffix.
+- `octo_message_extra_channel_seq.channel_id` charset **and** collation must match
+  `message_extra.channel_id`, and `channel_type` its persisted domain. Note
+  `message_extra` declares no explicit charset/collation (it inherits the schema
+  default), so the implementation must read the **effective** charset/collation
+  of `message_extra.channel_id` from the live schema and pin the new table to it
+  — do not assume `utf8mb4_general_ci`. The state table carries no channel key,
+  so it may use the standard `utf8mb4`/`utf8mb4_general_ci`.
+- The channel-sequence table contains no business content and has one hot row
+  per active storage channel, avoiding a process-wide or deployment-wide
+  sequence bottleneck. Both new tables are empty/tiny at creation, so their
+  `CREATE TABLE` statements are cheap and need no online DDL tooling.
+- `octo_message_extra_version_state` is the sole allocator-mode/floor source of
+  truth for every upgraded binary. Exactly one row (`singleton_id=1`) must
+  exist. A **missing** row defaults to `legacy` — the safe pre-migration
+  behavior (existing `GenSeq`), which a deploy also sees before the migration
+  seeds the row and which test setups leave after `CleanAllTables`; this is not
+  fail-closed. An unrecognized `mode` value or an unreadable row **does** fail
+  closed. `epoch` is never a write-abort condition (see D3). Guarding against a
+  row deleted while in transactional mode is the job of the optional
+  expected-mode config check (D9), not this read; the mode gauge still surfaces
+  `legacy` so an operator can detect an unexpected state. Environment variables
+  may assert an expected mode for rollout diagnostics, but must never select an
+  allocator independently of this DB row.
+- Separately, add on the **existing `message_extra` table** the composite index
+  the delta-sync read path needs: `(channel_id, channel_type, version)`. Today
+  `message_extra` carries only `channel_idx (channel_id, channel_type)`, so the
+  `WHERE channel_id=? AND channel_type=? AND version>? ORDER BY version ASC`
+  scan cannot resolve the `version` range/ordering from the index. This index is
+  the one with real DDL cost: because `message_extra` is a large, live table, the
+  migration must use the deployment's approved online-schema-change path when the
+  target MySQL version cannot add it online without a table rebuild or blocking
+  writes. (This is a read-path optimization independent of the sequence rework;
+  land it in the same task since both touch the cursor domain.)
+- Do not modify octo-lib `GenSeq` in this task. Its other callers have separate
+  compatibility and ordering contracts. During the Prepare phase, the one
+  centralized version store may retain a transitional legacy adapter; business
+  writers must no longer call `GenSeq(MessageExtraSeqKey, ...)` directly, and
+  transactional mode must never call it.
+
+### D3. Reservation API and transaction ownership
+
+Introduce one shared message-extra version store, usable from `modules/message`,
+`modules/bot_api`, `modules/robot`, and `internal/carddispatch`, with semantics
+equivalent to:
+
+```go
+ReserveTx(tx *dbr.Tx, storageChannelID string, channelType uint8, count int) (versions []int64, err error)
+```
+
+`versions` always has `len == count`, in assignment order. In transactional mode
+it is a contiguous ascending range; in legacy mode it is `count` distinct values
+from `GenSeq` (contiguity is not guaranteed under concurrency — matching pre-task
+per-message behavior). Returning explicit per-item values, rather than a single
+`first`, keeps callers mode-agnostic and behavior-preserving across the Prepare
+window (a `first`+`first..first+N-1` assignment would be unsafe in legacy mode).
+
+The store must live in a neutral package that all four callers can import
+without an import cycle (e.g. `internal/msgextraseq` or a `pkg/` package), not in
+`modules/message` itself if that would clash with existing dependency direction
+between `modules/message`, `modules/bot_api`, `modules/robot`, and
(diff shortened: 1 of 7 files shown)


lml2468 (approved):
# PR Review — PR #644 (Mininglamp-OSS/octo-server)

Reviewed at head `36d06131` against base `main`.
**Build/vet:** `go build ./internal/msgextraseq/` exit 0 ✅, `go vet` ✅. DB-backed quota/allocator tests are MySQL-gated → CI.

## ⚠️ Duplicate of PR #630
This PR is a **byte-identical duplicate** of #630, which I reviewed across four rounds and **APPROVED** (at #630's code head `36d06131`, before its `817232c` no-op retrigger):
- **Same head SHA** `36d06131`, same branch `analyze-issue-627`, same 3 commits (`8bb4bd00` → `872453f5` → `36d06131`), same +1982 across 7 files.
- `git diff 36d06131(#630) HEAD` = **empty** (identical tree).

No code difference to re-review — the #630 verdict transfers verbatim. **Recommend the author close one of #630/#644** to avoid a double-merge / split approval count.

## Verdict (carried from #630, re-verified on this checkout)
**APPROVE** — PR-1 of the staged rollout: schema + the transactional `message_extra` version allocator core, production seeded `mode=legacy` (inert). Every blocker raised across #630's rounds is present-and-fixed at this head (spot-verified):
- **Deadlock-free first-use init** (line 306): a **non-locking** `selectSequence` gates the materialize step, then `INSERT … ON DUPLICATE KEY UPDATE` + `SELECT … FOR UPDATE` only on an existing row — no `SELECT … FOR UPDATE` on a missing row, so no REPEATABLE-READ gap-lock 1213 deadlock (brief D3).
- **Singleton DB-enforced**: `CONSTRAINT chk_message_extra_version_singleton CHECK (singleton_id = 1)` (migration line 26).
- **Expected-mode fail-closed guard** (`assertExpectedMode`, `OCTO_MESSAGE_EXTRA_VERSION_EXPECTED_MODE`): a lost/missing state row while a deployment declares `transactional` fails closed instead of silently re-enabling legacy.
- **Overflow guard** (`cur > maxSafeInteger - count`), per-allocation floor raise, lock-stays-on-parent-vs-count-scope, fail-closed state read, low-cardinality metrics, migration **Down** (drop index + both tables, lines 39-44) — all present.
- Package **inert** (no writer wired; `mode=legacy` delegates to `GenSeq`) → deploy is behavior-neutral.

Both gates pass. Suggested decision: `Safe to merge` (identical to #630) once CI runs the DB-gated allocator suite on MySQL 8.0, and verify the `channel_id` collation against the live schema before the operator cutover — **after consolidating the #630/#644 duplicate** (keep one). (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (approved):
Well-scoped, defensively engineered allocator foundation. Independently byte-verified the concurrency-critical path against the head commit; no correctness or security blockers found. The transactional path is inert in production (state row seeded `mode=legacy`), so the only live effects are two additive tables and one `message_extra` composite index.

✅ Correctness / concurrency verification

- No TOCTOU on version allocation: `reserveTransactional` takes a per-channel row lock (`SELECT ... FOR UPDATE` on the `(channel_id, channel_type)` PK row) held until the caller commits, then advances `last_version` by `count`. Commit order == version order, unique per channel. `TestReserveTx_TwoSessionSerialization` proves the second tx blocks on the first's row lock and only observes a higher value after commit.
- Gap-lock / 1213 deadlock avoided by design: `lockSequenceRow` does a non-locking existence read first, materializes a missing row via `INSERT ... ON DUPLICATE KEY UPDATE`, and only ever issues `FOR UPDATE` against an already-existing row (a record lock, not a range/gap lock). This is exactly the pattern that avoids the empty-range gap-lock deadlock seen previously in this repo. `TestReserveTx_ConcurrentInitializers` (4 racing first-use goroutines) asserts unique versions, no duplicate-key (1062) error, and correct final cursor.
- Unique backstop is real: PK `(channel_id, channel_type)` on the sequence table and the singleton CHECK (`singleton_id = 1`) on the state table. Per-channel isolation confirmed by `TestReserveTx_DifferentChannelsNoGlobalLock`.
- Fail-closed safety: unknown `mode` or unreadable state row errors out; a missing row defaults to legacy but is guarded by the optional expected-mode env assertion so a lost row cannot silently re-enable legacy after a transactional cutover. `2^53-1` overflow guard runs before any mutation.
- DB-gated tests genuinely run in CI: no `t.Skip` / `Short()` guards, and the CI workflow provisions a MySQL 8.0 service with matching `utf8mb4_general_ci` collation, so the serialization / rollback / init-race / pagination tests actually execute.

🟡 Non-blocking notes

- Tests `DELETE FROM` shared `message_extra` / `seq` tables in setup; parallel package-level tests sharing one database could interfere. Consider isolated schemas or uniquely scoped fixtures.
- `CREATE TABLE IF NOT EXISTS` in the test helper can mask a drift between the test-created schema and the real migration. A dedicated test that asserts against the migration-created schema would close that gap.
- `lock_wait_seconds` timing in `lockSequenceRow` starts before the existence lookup / bootstrap, so it measures more than pure sequence-lock wait; start it immediately before the `FOR UPDATE` for a cleaner signal.
- The documented `result=retry` metric series has no in-package incrementer yet; add an exported reporting hook when whole-transaction retry handling lands.

🔵 CI note

- The `Test` check is currently `pending` (running against MySQL), not failing. Recommend confirming it goes green before merge; nothing in the diff suggests it should not. `mergeable=MERGEABLE` (merge state BLOCKED pending required approval, which is expected).

Highlights: caller-owned transaction with per-channel locking, race-safe first-use bootstrap (max of floor / `message_extra` max / legacy `seq.min_seq`), DB-authoritative mode selection with a `FOR SHARE` cutover drain barrier, low-cardinality metrics, and strong test coverage of the concurrency invariants.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #644 (octo-server)

## Summary

This PR lands the schema and allocator core for transactional `message_extra` version allocation (#627). Two new tables (`octo_message_extra_channel_seq`, `octo_message_extra_version_state`) and a composite index on `message_extra` are introduced via migration, alongside a new leaf package `internal/msgextraseq` that provides a `ReserveTx` API. Production stays inert — the state row is seeded to `mode=legacy` (delegating to existing `GenSeq`), and no writer is switched over in this PR. The design is thoughtful: deadlock-free first-use bootstrap, per-channel row locking (no global lock), `2^53-1` overflow guard, expected-mode deployment guard, and comprehensive integration tests.

## Verification

Local checks at head `36d06131`:
- `go build ./internal/msgextraseq/` — clean
- `go vet ./internal/msgextraseq/` — clean
- Static analysis of all 7 changed files, data flow tracing through `ReserveTx` → `readStateForShare` → mode dispatch → legacy/transactional paths, bootstrap flow, and lock ordering.

## Findings

### P2 — `epoch` column type narrowing (`internal/msgextraseq/store.go:183`)

The migration defines `epoch` as `BIGINT UNSIGNED` (range 0..2^64-1), but `readStateForShare` scans it into Go `int64` (range -2^63..2^63-1). The go-sql-driver may reject values above `MaxInt64`. In practice this is harmless — `epoch` is an observability-only counter starting at 0, incremented only on operator CAS during cutover, and will never approach 2^63. The subsequent cast `Epoch: uint64(row.Epoch)` at `:200` is correct for the in-range values that will actually occur.

Fix direction: scan into `uint64` (matching the `State.Epoch` field type), or document the intentional narrowing.

### P2 — Redundant index after composite addition (`modules/message/sql/20260721000001_message_extra_version.sql:37`)

The new `channel_version_idx(channel_id, channel_type, version)` makes the existing `channel_idx(channel_id, channel_type)` from `20210407000001_message_legacy01.sql:23` redundant — MySQL uses the left-prefix of the composite for any query previously served by `channel_idx`. Keeping both wastes disk, buffer pool, and write throughput on `message_extra` (a high-write table). Not a blocker, but worth dropping `channel_idx` in a follow-up migration once the composite is confirmed serving in production.

### Nit — Mixed comment languages within package (`internal/msgextraseq/metrics.go:8`)

`metrics.go` uses Chinese comments throughout while `store.go` (same package) uses English. Minor style inconsistency — consider aligning to one language.

- **Lock ordering / deadlock freedom**: `lockSequenceRow` uses a non-locking existence read before the `INSERT ON DUPLICATE KEY UPDATE`, then `SELECT FOR UPDATE` only targets an existing row (record lock, no gap lock). This correctly avoids the REPEATABLE READ gap-lock deadlock described in the comments.
- **Two-session serialization**: `FOR UPDATE` on the per-channel row correctly serializes concurrent reservations within the same channel. `TestReserveTx_TwoSessionSerialization` proves this with a live MySQL.
- **Rollback safety**: Sequence advances happen within the caller's transaction — rollback reverts both the business rows and the sequence advance. `TestReserveTx_Rollback` confirms.
- **Cross-channel independence**: No global lock; different channels operate independently. `TestReserveTx_DifferentChannelsNoGlobalLock` confirms.
- **Bootstrap race safety**: `INSERT ... ON DUPLICATE KEY UPDATE last_version=last_version` serializes concurrent first-use initializers on the insert-intention lock without deadlock. `TestReserveTx_ConcurrentInitializers` confirms unique versions across 4 concurrent goroutines.
- **Overflow guard**: `cur > maxSafeInteger - int64(count)` correctly prevents versions from crossing `2^53-1`. `TestReserveTx_Overflow` confirms boundary behavior.
- **Count bounds**: `count <= 0 || count > MaxReserveCount` properly rejects invalid inputs. `TestReserveTx_CountBounds` confirms no mutation on rejection.
- **Expected-mode guard**: `OCTO_MESSAGE_EXTRA_VERSION_EXPECTED_MODE` correctly fails closed on mismatch, malformed value, or missing state row when transactional is expected. `TestExpectedModeGuard` covers all four branches.
- **State singleton constraint**: `CHECK (singleton_id = 1)` structurally prevents a second state row. `TestStateSingletonConstraint` confirms.
- **Migration column parity**: `channel_id` (VARCHAR(100), utf8mb4_general_ci), `channel_type` (SMALLINT), and `last_version` (BIGINT) all match the corresponding `message_extra` column types.
- **`reserveLegacy` key parity**: Key format `messageExtra:<channelID>` matches all existing production callers (`api_manager.go:978`, `api.go:2726`, `carddispatch/mutation.go:287`, `robot/api.go:2047`, `bot_api/send.go:1064,1109`).
- **Bootstrap legacy key**: `seq:messageExtra:<channelID>` correctly targets the `seq` table row used by GenSeq.
- **FOR SHARE on state singleton**: Compatible shared locks allow concurrent readers; exclusive cutover flip waits (drain barrier). Correct by design.
- **Metrics cardinality**: All metrics use low-cardinality labels (result, mode). No channel_id/uid labels that would explode Prometheus.

## Data flow tracing

- `ReserveTx` → `readStateForShare` → `SELECT mode, epoch, cutover_floor FROM octo_message_extra_version_state WHERE singleton_id=1 FOR SHARE` → mode dispatch.
- Mode=legacy → `reserveLegacy` → `ctx.GenSeq("messageExtra:<channelID>")` → returns distinct values (same path as existing production callers).
- Mode=transactional → `reserveTransactional` → `lockSequenceRow` (non-locking read → optional INSERT ON DUPLICATE → FOR UPDATE) → floor raise → overflow check → `UPDATE last_version` → return contiguous range `[cur+1, cur+count]`.
- All returned `[]int64` values flow to the caller for assignment to business rows. No unwired returns.

## Blind spot checklist (C1–C6)

- **C1 (dual-path parity)**: N/A — no symmetric add/remove paths. Legacy and transactional modes are parallel allocator strategies, not symmetric CRUD.
- **C2 (control-flow ordering)**: Clear — `readStateForShare` is called once at the top of `ReserveTx`. No nesting or re-entry.
- **C3 (authorization boundary)**: N/A — allocator explicitly does no auth (by design, documented). Auth stays in callers. No production callers wired yet.
- **C4 (authorization lifecycle)**: N/A — no auth in this package.
- **C5 (build ≠ runtime)**: `go build` and `go vet` pass. Cannot run MySQL-dependent tests in this environment. Migration structure verified against existing schema.
- **C6 (governance/docs self-consistency)**: `.octospec` docs (brief, plan, context.yaml) are consistent with the implementation. The API deviation from brief D3 (`[]int64` vs single `first`) is documented and justified in `context.yaml`.

## Cross-round blocker check (R6)

## Verdict: APPROVED

Well-designed infrastructure PR with strong test coverage, clear operational safety (legacy-default, expected-mode guard, deadlock-free bootstrap), and no production-path changes. The two P2 items (epoch type narrowing, redundant index) are non-blocking and can be addressed in follow-ups.

yujiawei (approved):
# Code Review — PR #644 (octo-server)

**Verdict: APPROVED.** This is a re-review of the previously-flagged first-use initialization deadlock, plus the two follow-up hardening items. The blocking defect is genuinely fixed and empirically verified against live MySQL 8.0 (REPEATABLE READ). No P0/P1 remains. Everything below is P2/advisory or a pre-cutover human-verify item — none blocks merging this PR-1 (schema + inert allocator, `mode=legacy`, no writer wired).

## 1. Scope / spec compliance — ✅

- **No under-build:** all PR-1 items present — the two tables (`octo_message_extra_channel_seq`, `octo_message_extra_version_state` seeded to legacy), the `message_extra (channel_id, channel_type, version)` composite index, and the leaf package `internal/msgextraseq` (`ReserveTx` / state dispatch / legacy adapter / metrics).
- **No over-build:** production stays inert — `grep` confirms nothing imports `msgextraseq` outside the package, and there is no `ReserveTx` call site, so no runtime write path changes. The source-guard allowlist test is correctly deferred to the later writer-cutover stage.
- **Deviation acknowledged:** `ReserveTx` returns `[]int64` rather than a single `first int64`; this is documented in `context.yaml` (contiguity is only guaranteed in transactional mode; `[]int64` keeps callers mode-agnostic during the prepare window). Reasonable and safer.

## 2. Blocking-item verification

### 2.1 First-use concurrent-initializer deadlock — FIXED ✅ (empirically verified)

The earlier round's blocker was a gap-lock / insert-intention AB-BA deadlock: a `SELECT ... FOR UPDATE` was issued against a not-yet-existent PK row, which under REPEATABLE READ takes a gap lock, and two concurrent first-use initializers then deadlocked (MySQL 1213) on the subsequent insert.

`lockSequenceRow` (`store.go:304`) now does the correct inversion:
1. `selectSequence` — a **non-locking** consistent read (`store.go:346`); a missing row takes no gap lock.
2. On not-found, `INSERT ... ON DUPLICATE KEY UPDATE last_version=last_version` materializes the row (`store.go:319`); concurrent initializers serialize on the record/insert-intention lock — they wait, they do not deadlock.
3. `selectSequenceForUpdate` (`store.go:365`) then locks a row that already exists → a record lock, no gap.

Verified against a live MySQL 8.0 at `REPEATABLE-READ`:
- `TestReserveTx_ConcurrentInitializers` passes **20× under `-race`** with zero 1213s (it was intermittently failing before).
- Full `internal/msgextraseq` suite green under `-race -count=3`; the serialization / cross-channel / rollback tests all pass repeatedly.

### 2.2 Migration `Down` — added ✅
`+migrate Down` now drops the composite index and the two new tables. It touches **no `message_extra` rows**, so rollback of this PR-1 is a clean no-op drop.

### 2.3 Expected-mode guard — added ✅
`OCTO_MESSAGE_EXTRA_VERSION_EXPECTED_MODE` fails closed when the resolved mode (including a missing state row → legacy default) does not match the declared expectation (`store.go:221`, `readStateForShare` at `store.go:208`). This is the durable safety net that stops a lost/deleted state row from silently reverting to legacy after a future cutover. Unset makes no assertion, preserving safe pre-migration behavior. `TestExpectedModeGuard` covers missing-row/mismatch/malformed cases.

## 3. Correctness spot-checks (verified sound)

- **Range math / floor raise / overflow:** returned versions are `max(current, floor)+1 .. +count`; the overflow guard rejects before any mutation and allows exactly `2^53-1` (`store.go:261-289`). `TestReserveTx_Overflow` pins the boundary.
- **Bootstrap cannot under-seed persisted versions:** `computeBootstrap` = `max(floor, MAX(message_extra.version) for the channel, legacy seq.min_seq)` (`store.go:387`). Its evidence sources are conservative-high, so a first reservation never lands at or below an already-persisted version.
- **Lock ordering:** state row `FOR SHARE` → sequence row `FOR UPDATE`; no cycle.
- **Legacy key parity:** `reserveLegacy` calls `GenSeq("messageExtra:<channelID>")`, which persists under `seq:messageExtra:<channelID>` — exactly the key `computeBootstrap` reads. Consistent.
- Build (`go build ./...`), `go vet`, `gofmt`, `make i18n-lint`, and `make i18n-extract-check` all pass.

## 4. Non-blocking findings (P2 / advisory)

- **P2 — Migration is not restartable after a partial DDL failure.** Each DDL implicitly commits; if the large `CREATE INDEX` on `message_extra`, a later `CREATE TABLE`, or the seed `INSERT` fails, sql-migrate won't record the migration and a re-run dies on "table already exists". Note this matches the repo's existing migration style (recent message migrations also omit `IF NOT EXISTS`), so it's a pre-existing convention rather than a regression — but for a large-table online `CREATE INDEX`, consider splitting the index into its own migration and/or making the `Up` idempotent.
- **P2 (human-verify) — `channel_id` collation.** The new sequence table pins `utf8mb4_general_ci`, but `message_extra.channel_id` inherits the DB/table default and does not declare it explicitly. If production differs, the sequence table could merge or split channel identities relative to `message_extra`. The migration comment already flags re-checking this before cutover — please confirm the effective production collation.
- **P2 (human-verify) — post-cutover safety leans on the optional env guard.** A deleted state row defaults to legacy; only `OCTO_MESSAGE_EXTRA_VERSION_EXPECTED_MODE=transactional` makes it fail closed. Make setting/auditing this guard a hard prerequisite in the cutover runbook. Relatedly, once the allocator is active in a later stage, the current `Down` (drop tables) would discard live sequence/floor state — the `Down` should become forward-only or refuse after activation. Not a concern for this inert PR-1, but worth pinning in the rollout doc.

## 5. Nits (optional)

- `channel_idx(channel_id, channel_type)` is now a strict left-prefix of the new `channel_version_idx` and is redundant for reads; you may drop it after checking query plans (kept-intentionally is also fine).
- `computeBootstrap` guards `dbr.ErrNotFound` on `COALESCE(MAX(...),0)` / `COALESCE(min_seq,0)` aggregates that always return one row — the guard is dead (harmless).
- `epoch` is `BIGINT UNSIGNED` in schema but decoded into an `int64` field before the `uint64` cast; fine for an observability-only value, but a very large epoch would wrap.

## 6. Reviewer notes

- This PR was classified security-sensitive. The load-bearing safety property for PR-1 is that it stays **inert** (no writer switched), which is verified. The items a human should confirm before the later cutover are the collation check and the expected-mode guard being mandatory in the runbook.
- One independent adversarial review leg completed and concurred that the deadlock is fixed and no P0 exists; the second automated leg timed out and did not return findings (recorded as absent, not as approval). The verdict rests on the empirical live-MySQL verification above and an independent read of the diff.

PR #622, opened 2026-07-21
feat(notify): per-Space welcome message with admin CRUD

an9xyz:
## Summary

Make the onboarding welcome message **per-Space and self-service**. Each Space's
admins (Role ≥ 1) manage **one** welcome config for **their** Space via
`GET/PUT/DELETE /v1/space/:space_id/welcome`; the platform-global config
(`system_setting onboarding.space_welcome_*`) is kept as a **superadmin
fallback**.

Product follow-up to #604 / #606, which shipped a single **platform-designated**
welcome Space, superadmin-only. This lifts that task's out-of-scope item
("Multi-Space configuration with per-Space copy; per-Space admin self-service
API") into scope. Ships `enabled=false` per Space by default — with no per-Space
row written, behavior is identical to today.

## Related Issue

N/A (product follow-up to #604 / #606).

## Linked Spec

`.octospec/tasks/space-welcome-per-space-admin-crud/brief.md`

## Changes

- **Config storage (`modules/common`)** — new `octo_space_welcome_config` table
  (one row per Space) + stateless `SpaceWelcomeConfigStore`
  (Get/Upsert/Delete/Exists/ListEnabled). `ResolveEffectiveSpaceWelcome`
  precedence: **a present per-Space row wins outright, even when disabled**
  (opt a Space out of a global campaign); else the global config applies iff it
  is enabled and names the Space; else off. Reuses the existing
  `SpaceWelcomeConfig` shape so `ValidateSpaceWelcomeCombination` /
  `ParsedActiveFrom` apply unchanged (`active_from` stored as the RFC3339 string
  to enable that reuse).
- **CRUD API (`modules/space/api_welcome.go`)** — `GET/PUT/DELETE
  /v1/space/:space_id/welcome`. Authz reuses `checkSpaceActive` +
  `requireSpaceAdmin` (Role ≥ 1); a request can only affect the path
  `:space_id`. `PUT` is a prospective merge (patch onto the existing row,
  validate the composite). `DELETE` hard-deletes the row (reverts to the global
  fallback or off). Mounted on the auth + `SharedUIDRateLimiter` group; all
  errors via `httperr` with **existing** error codes (no new codes).
- **Delivery driver (`modules/notify`)** — single-Space → all-enabled-Spaces.
  The event path, reconciler, and worker resolve the per-Space **effective**
  config each cycle. Reconciler **and** worker rotate a per-replica cursor over
  the enabled-Space set (fairness). Cross-space sweep (`sweepClaimedAll` /
[...]

PR #606, opened 2026-07-17
refactor(notify): single plain-text Space welcome message (drop bilingual split)

an9xyz:
## Summary

Product follow-up to #604 (Space new-user welcome DM, shipped `enabled=false`).
The welcome message no longer distinguishes zh/en — **one plain-text body is
sent to every recipient**, and **line breaks are supported** (newlines preserved
verbatim; clients render `type:1` text with breaks; markdown is NOT rendered).

## Changes

- **Config**: replace the two keys `onboarding.space_welcome_message_zh_cn` /
  `space_welcome_message_en_us` with a single `onboarding.space_welcome_message`.
  `SpaceWelcomeConfig` now has one `Message` field; composite validation and the
  manager prospective-merge write path use the single key.
- **notify**: drop language resolution (`resolveLanguage` / `messageForLang` /
  the `langSvc` dependency and its bounded lookup); `dispatch` sends
  `cfg.Message` directly. The ledger `lang` column is written empty
  (language-agnostic) — no schema change.
- **i18n**: update the zh-CN copy for `err.server.common.space_welcome_config_invalid`
  (drop the 中英文 wording).
- No SQL migration: `system_setting` is a generic KV store, and the ledger
  `lang` column is unchanged.

## Line breaks
Already supported end-to-end: validation only `TrimSpace`s (internal `\n`
preserved), and the body is passed to `payload.content` verbatim. Admins format
with real newlines (plain text); `###` / `**` markdown does not render.

## Testing

- `go build ./...`, `go vet`, `make i18n-extract-check`, `make i18n-lint`,
  `git diff --check` — all clean.
- `go test -race` green for `modules/notify` and `modules/common` (fresh DB).
- New/updated tests: single-message config read / defaults / snapshot atomicity;
  single-message validation + prospective-merge both directions; **multi-line
  body preserved-and-valid**; dispatch sends the configured body (with newline)
  verbatim over the wire.

- [x] Unit tests added/updated
- [x] Manually verified via integration tests against real MySQL

## Note
Depends conceptually on #604 (already merged to `main`). Feature remains
`enabled=false`; production-enable gates (metrics/alerting + product sign-offs)
are unchanged and tracked separately.

PR #604, opened 2026-07-16
feat(notify): at-most-once Space new-user welcome DM

an9xyz:
## Summary

Deliver **one** configurable welcome DM from the built-in `notification` bot
("通知助手") when a **human** user first joins an admin-designated Space. Minimal
implementation of the main chain "first-join human → exactly one ledger row →
member re-check before send". Send semantics are **at-most-once**;
transport-ambiguous outcomes are recorded as `unknown` and never auto-retried.

Ships with the feature `enabled=false` (no behaviour change on deploy).

## Related Issue

N/A (self-originated task).

## Linked Spec

`.octospec/tasks/space-new-user-welcome-message/brief.md`

## Changes

- **Ledger**: new `octo_space_welcome_delivery` table (migration in
  `modules/notify/sql/`); wired `//go:embed sql` + `SQLDir` into
  `notify/1module.go` (it had none before, so the migration now runs).
- **Config**: five `system_setting` keys under `onboarding`; `modules/common`
  gains an atomic `SpaceWelcomeConfig()` snapshot accessor + prospective
  composite validation on the manager write path; i18n code
  `err.server.common.space_welcome_config_invalid` via `httperr`.
- **notify service**: event enqueue (extends the `SpaceMemberJoin` listener,
  never blocks/rolls back the join), a 60s reconciler (dropped events /
  restarts / manager `addMembers`), and a send worker (idle 5s / drain /
  per-wake cap 20). Claim via `SELECT ... FOR UPDATE SKIP LOCKED`; every
  post-claim UPDATE is a CAS guarded by `status + claim_owner`. `attempts`
  grows only on a definitive pre-IM failure (backoff 5s/30s/120s, 4th → failed);
  any failure after the dispatching transition → `unknown`.
- **Sender**: notify-local context-aware HTTP sender (`http.Client{Timeout:15s}`
  + `NewRequestWithContext`, explicit `Content-Type` + manager token). octo-lib
  is unmodified; 15s < 30s lease.
- **Time**: `active_from` vs `space_member.created_at` compared via
  `UNIX_TIMESTAMP` epoch (mirrors `modules/opanalytics`), correct for the
  deployment-local DB session zone.
- Observability kept minimal (in-process counters + structured logs).

## Testing

- `go build ./...`, `go vet`, `make i18n-extract-check`, `make i18n-lint`,
  `git diff --check` — all clean.
- `go test -race` green for `modules/notify`, `modules/common`, `modules/space`
  (each on a fresh test DB).
[...]

PR #646, opened 2026-07-23
feat(notify): per-Space welcome message with admin CRUD

an9xyz:
## Summary

Make the onboarding welcome message **per-Space and self-service**. Each Space's
admins (Role ≥ 1) manage **one** welcome config for **their** Space via
`GET/PUT/DELETE /v1/space/:space_id/welcome`; the platform-global config
(`system_setting onboarding.space_welcome_*`) is kept as a **superadmin
fallback**.

Product follow-up to #604 / #606, which shipped a single **platform-designated**
welcome Space, superadmin-only. This lifts that task's out-of-scope item
("Multi-Space configuration with per-Space copy; per-Space admin self-service
API") into scope. Ships `enabled=false` per Space by default — with no per-Space
row written, behavior is identical to today.

## Related Issue

N/A (product follow-up to #604 / #606). Supersedes #622 (closed and re-opened as
a fresh PR from the same branch/head).

## Linked Spec

`.octospec/tasks/space-welcome-per-space-admin-crud/brief.md`

## Changes

- **Config storage (`modules/common`)** — new `octo_space_welcome_config` table
  (one row per Space) + stateless `SpaceWelcomeConfigStore`
  (Get/Upsert/Delete/Exists/ListEnabled). Concurrent PUTs are serialized by
  `UpsertMerged` using an **insert-then-lock** strategy (idempotent
  `INSERT ... ON DUPLICATE KEY UPDATE` to guarantee the row exists, then
  `SELECT ... FOR UPDATE` before the merge, with a bounded deadlock/lock-timeout
  retry), closing **both** the existing-row lost-update and the concurrent
  first-create race. `ResolveEffectiveSpaceWelcome` precedence: **a present
  per-Space row wins outright, even when disabled** (opt a Space out of a global
  campaign); else the global config applies iff it is enabled and names the
  Space; else off. Reuses the existing `SpaceWelcomeConfig` shape so
  `ValidateSpaceWelcomeCombination` / `ParsedActiveFrom` apply unchanged
  (`active_from` stored as the RFC3339 string to enable that reuse).
- **CRUD API (`modules/space/api_welcome.go`)** — `GET/PUT/DELETE
  /v1/space/:space_id/welcome`. Authz reuses `checkSpaceActive` +
  `requireSpaceAdmin` (Role ≥ 1); a request can only affect the path
  `:space_id`. `PUT` is a prospective merge (patch onto the existing row,
  validate the composite) with message- and `active_from` column-width guards;
  `updated_at` and the effective `active_from` are surfaced as stable RFC3339.
  `DELETE` hard-deletes the row (reverts to the global fallback or off). Mounted
  on the auth + `SharedUIDRateLimiter` group; all errors via `httperr` with
  **existing** error codes (no new codes).
- **Delivery driver (`modules/notify`)** — single-Space → all-enabled-Spaces.
  The event path, reconciler, and worker resolve the per-Space **effective**
  config each cycle. Reconciler **and** worker rotate a per-replica cursor over
  the enabled-Space set (fairness). Cross-space sweep (`sweepClaimedAll` /
  `sweepDispatchingAll`, backed by a new `idx_sweep (status, claim_expire_at)`)
  reclaims stale in-flight rows across Spaces. Ledger state machine,
  at-most-once, `FOR UPDATE SKIP LOCKED` claim, CAS-on-`claim_owner`, and the
  fixed `notification` sender identity are **unchanged**.

## Testing

- `go build ./...`, `go vet`, `golangci-lint` (0 issues), `gofmt`,
  `make i18n-extract-check` + `make i18n-lint`, `git diff --check` — all clean.
- `go test -race` green for `modules/common`, `modules/notify`, `modules/space`
  (each on a fresh DB — cross-package migration sets differ, so the shared
  `test` DB is dropped + recreated between packages).
- New tests: config-store CRUD; effective-config precedence (both directions);
  **concurrent partial-update lost-update** and **concurrent first-create**
  regressions; message / `active_from` length guards; multi-Space independent
  delivery + **no cross-Space body / space_id mixing**; worker fair-rotation
  (fails without the cursor fix); cross-space sweep; per-Space event enqueue
  scoping; CRUD authz matrix (admin/member/non-member/inactive) + prospective
  validation + partial-merge + delete.
- **Real-wire e2e** (ad-hoc, not committed) against live MySQL/Redis/WuKongIM:
  single-Space delivery (ledger SENT + real IM `message_id`) and two-Space
  concurrent delivery, reading each recipient's channel back from WuKongIM to
  confirm actual receipt and that each got ONLY their own Space's body +
  authoritative `payload.space_id`.

- [x] Manually verified (integration + real-wire e2e against live MySQL/Redis/WuKongIM)

## COMPREHENSION

   It changes the welcome feature's config *source* from a single global tuple
   to per-Space (with global fallback), and generalizes the delivery driver from
   one Space to every enabled Space. The **delivery reliability core is
   untouched**: the same `octo_space_welcome_delivery` ledger (`UNIQUE(space_id,
   uid)`), the same state machine, `FOR UPDATE SKIP LOCKED` claim, CAS on
   `claim_owner`, at-most-once, and `notification` sender identity. Space
   isolation is enforced two ways: (a) CRUD authz resolves the caller's
   membership for the **path** `:space_id` and requires Role ≥ 1, and a request
   can only affect that Space; (b) the worker dispatches each claimed row using
   **that row's own** Space effective config and re-checks membership against
   that Space, so one Space's config or backlog can never deliver into another.
   Error responses touch the wire contract only through the existing i18n
   `httperr` envelope with already-registered codes (no new codes).

   - Dependents: the `SpaceMemberJoin` listener and the notify reconciler/worker
     now read the per-Space config table (read-through) instead of one global
     snapshot; a DB error there fails closed (logged, no enqueue/send — never
     blocks the join). The global `system_setting` write path and schema are
     untouched.
   - Multi-replica / rolling update: unchanged at-most-once machinery. Concurrent
     claim → exactly one winner (`SKIP LOCKED`); concurrent enqueue → one row
     (`ON DUPLICATE KEY`). `sweepAll` reclaims only **lease-expired**
     (`claim_expire_at<=now`) rows, so it never steals a healthy peer's live
     claim; a replica dying mid-dispatch leaves a `dispatching` row that a peer
     sweeps to `unknown` (not retried → no double-send).
   - Config writes: concurrent admin PUTs (edit or first-create) are serialized
     by the insert-then-lock `UpsertMerged`, so no partial field is lost and a
     first-create can't gap-lock-deadlock into a 500.
   - Fairness: a naive greedy worker would starve tail Spaces under sustained
     load; fixed with a rotating cursor (+ regression test).
   - Feature ships `enabled=false` per Space; with no per-Space row, the global
     config still drives its one Space exactly as before.

   `go test -race ./modules/{common,notify,space}/...` exercises the full state
   machine and the new paths against real MySQL: store CRUD, effective-config
   precedence both directions, `TestSpaceWelcomeConfigStore_UpsertMerged_NoLostUpdate`
   and `..._ConcurrentCreate` (two admins patching disjoint fields — existing row
   and first-create — both survive), `TestPerSpace_MultiSpaceDelivery` (each
   recipient gets only their Space's body), `TestPerSpace_Worker_FairRotation`
   (keeps the head Space over-cap every wake and asserts the tail Space is still
   served — fails without the cursor), `TestPerSpace_SweepAll_CrossSpace`, and
   the CRUD authz/validation matrix. Additionally a real-wire e2e drove two
   Spaces' delivery through the notify-local HTTP sender into a running WuKongIM
   and read each recipient's channel back from the IM: `ua` received exactly
   `WELCOME-ALPHA`/`space_id=spcA…`, `ub` exactly `WELCOME-BRAVO`/`space_id=spcB…`
   — actual receipt, no crossing.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01XqkVUKZD3uEYYTvvqmnWio)_


lml2468 (dismissed):
# PR Review — PR #646 (Mininglamp-OSS/octo-server)

Reviewed at head `b9d9caa3` against base `main` (`merge: DIRTY` — needs a rebase before merge).
**Build/vet/test:** `go build` exit 0 ✅, `go vet` ✅. All new suites (`space_welcome_config_store`, `api_welcome`, `space_welcome_perspace`) are **DB-gated** (`testutil` → MySQL refused in sandbox) — unrun here; CI must run them. The pure `ResolveEffectiveSpaceWelcome`/`api_welcome_unit` logic verified by reading.
Classification: per-Space onboarding welcome + admin CRUD — space-isolation (P92) + concurrency + rate-limit + migration + at-most-once delivery. Reviewed in **SECURITY / RELIABILITY** emphasis. Same task as #622 (which I reviewed extensively); this is a distinct implementation on branch `claude/space-welcome-per-space-admin-crud`.

## 1. Spec compliance — ✅
Graded against `.octospec/tasks/space-welcome-per-space-admin-crud/brief.md`.
- **Missing:** none. Per-Space config table + store (Get/UpsertMerged/Delete/ListEnabled), effective-resolution precedence, CRUD API with admin authz, multi-Space delivery drive-loop, cross-space sweep index, cursor rotation — all present.
- **Over-build:** none. Reuses the delivery ledger / state machine / at-most-once / sender identity unchanged (per brief); no new error codes on the hot path.
- **Deviation:** none. Ships `enabled=false` per Space (no per-Space row → global-fallback behavior preserved) — backward compatible.
- **Load-bearing rules:** space-isolation ✅, error-handling ✅ (all via `httperr.ResponseErrorL`), rate-limit ✅ (routes on `AuthMiddleware`+`SharedUIDRateLimiter`).

## 2. Code quality — Approved

### Findings
No P0/P1/P2 findings.

### Verified clean (adversarial checks that did NOT find a bug) — incl. the two concurrency defects I flagged on #622, both fixed here
- **Isolation by construction.** Every CRUD handler gates on `authorizeSpaceAdmin` = `checkSpaceActive` + `requireSpaceAdmin` (membership + `Role≥1`, fail-closed), and all reads/writes are scoped to the **path** `:space_id` — there is no caller-supplied target-space field, so a caller can only ever affect their own Space.
- **Lost-update on concurrent PUT — FIXED.** `UpsertMerged` does read-merge-write **under a `SELECT … FOR UPDATE` row lock** (`upsertMergedOnce` step 2): two admins PUTting the same Space serialize, the second merges onto the first's committed row. Validation runs *inside* the merge closure against the locked row and aborts the tx on failure (`errWelcomeValidation`). `TestSpaceWelcomeConfigStore_UpsertMerged_NoLostUpdate` pins it. (This was my #622 P1.)
- **First-create race — FIXED.** `upsertMergedOnce` uses the insert-then-lock idiom: `INSERT … ON DUPLICATE KEY UPDATE space_id=space_id` (idempotent materialize; `affected==1` ⇒ this PUT created it) → then `SELECT … FOR UPDATE`. Two concurrent first-creators both materialize the row without a 1062 leak, then serialize on the lock. `TestSpaceWelcomeConfigStore_UpsertMerged_ConcurrentCreate` covers it. Plus an `isRetryableTxErr` (1213/1205) retry loop around the whole op. (This was my #622 second miss.)
- **created_at preserved:** the ON DUPLICATE clause is a no-op and the persist step is a plain UPDATE, so an existing row's `created_at` is never clobbered.
- **Precedence deterministic** (`ResolveEffectiveSpaceWelcome`): a present per-Space row wins outright **even when disabled** (admin opt-out of a global campaign); else the global config iff enabled and it names this Space; else off. Exactly one effective config per Space.
- **At-most-once preserved:** the drive-loop reuses the existing ledger + claim (`FOR UPDATE SKIP LOCKED` + `claim_owner` CAS) unchanged; `ListEnabled` returns per-Space rows + the global-fallback Space iff enabled and not overridden; enqueue failure never blocks/rolls back the reconcile cycle.
- **Fairness/no-starvation:** reconciler and send worker each rotate a per-cycle/per-wake cursor over the enabled-Space set (the #622 follow-up) — no Space is perpetually first or starved.
- **Migrations additive & collation-safe:** `octo_space_welcome_config` uses the `octo_` prefix, `UNIQUE KEY uk_space` (required by the upsert's ON DUPLICATE), and `space_id` charset/**COLLATE explicitly aligned to `space.space_id` (utf8mb4_general_ci)** — avoids the JOIN 1267 mismatch. New `idx_sweep(status, claim_expire_at)` keeps the multi-Space cross-space sweep index-backed (vs the space_id-led `idx_claim`), with a Down segment. App-computed UTC timestamps (no `NOW()`).
- **Message column guard:** `VARCHAR(2000)` + validation to ≤2000 code points, newlines preserved, plain-text (no markdown) — matches the #606 contract.

## 3. Test assessment
Substantial coverage: store CRUD, `UpsertMerged` no-lost-update + concurrent-create (the two concurrency defects), effective-resolution precedence, per-Space delivery, and `api_welcome` handler tests. **All DB-gated** (module `testutil` needs MySQL), so unexecuted here — this is the main residual: CI must run them green (the concurrency tests especially need real MySQL row-locking to be meaningful). The pure precedence/unit logic reads correctly.

## 4. Coverage / blind spots
- Verified by build/vet + full read of the CRUD authz, `UpsertMerged`/`upsertMergedOnce` concurrency, precedence, drive-loop, and migrations.
- NOT run: all DB-gated suites; live migration apply (the `uk_space` upsert + collation against real MySQL); the multi-Space worker/reconciler cursor rotation under real concurrency. Recommend the CI DB suite + a staging soak. Also: `merge: DIRTY` → rebase on main before merge.

## Verdict
**APPROVE** — a complete, well-constructed per-Space welcome CRUD that enforces isolation by construction (path-space-only writes + `requireSpaceAdmin`), and — notably — **already incorporates both concurrency fixes I raised on the sibling #622**: the concurrent-PUT lost-update is closed with a `FOR UPDATE` read-merge-write, and the first-create race with the insert-then-lock idempotent-materialize idiom (both retry-wrapped and test-covered). Deterministic precedence, at-most-once preserved, cursor-rotation fairness, and collation-aligned additive migrations round it out. Both gates pass on inspection.
Suggested decision: `Safe to merge after (a) rebase (currently DIRTY) and (b) CI runs the DB-gated suites green` — the concurrency correctness is sound on review but its defining tests are infra-gated and must pass on real MySQL. (If #622 and #646 are the same task via different branches, consolidate to one.)

Jerry-Xin (dismissed):
Solid, well-scoped implementation of per-Space welcome config with admin CRUD. The load-bearing concerns for this feature — admin authorization and multi-tenant (per-Space) isolation — are correctly enforced, and the delivery path preserves the existing at-most-once reliability contract.

## ✅ Highlights

- **AuthZ enforced on every endpoint.** `GET/PUT/DELETE /v1/space/:space_id/welcome` all go through `authorizeSpaceAdmin`, which reuses `checkSpaceActive` + `requireSpaceAdmin` (`modules/space/api_welcome.go`). `requireSpaceAdmin` (`modules/space/api_email_invite.go`) queries membership for `(space_id, loginUID)` and requires `Role >= 1`, denying non-members and plain members. Routes sit on the authenticated + `SharedUIDRateLimiter` group (`modules/space/api.go`), so unauthenticated calls never reach the handlers.
- **No cross-tenant IDOR.** The target Space is always the path param `:space_id`, never a client-supplied body field, and the caller's admin membership is resolved against that exact Space — so a request can only ever affect its own Space by construction.
- **Store isolation.** Every query/mutation in `modules/common/space_welcome_config_store.go` and `modules/notify/space_welcome_db.go` is scoped by `space_id`; CAS state transitions key on `id + status + claim_owner` of an already-Space-scoped claimed row.
- **Delivery isolation + reliability.** The worker re-resolves each Space's effective config before every claim and dispatches each claimed row with that row's own `space_id`/message plus a recipient membership re-check (`modules/notify/space_welcome.go`), so one Space's config or backlog can never deliver into another. At-most-once CAS, `SKIP LOCKED` claim, and the fail-closed enqueue (join is never blocked/rolled back) are preserved.
- **Concurrency.** `UpsertMerged` uses insert-then-lock (idempotent `INSERT ... ON DUPLICATE KEY UPDATE`, then `SELECT ... FOR UPDATE`) with a bounded deadlock/lock-timeout retry, correctly closing both the lost-update and concurrent-first-create races.
- **Schema.** New `octo_space_welcome_config` table has `UNIQUE(space_id)` (one row per Space), `idx_enabled` for `ListEnabled`, and a new `idx_sweep(status, claim_expire_at)` to keep the cross-Space sweep index-backed. Precedence rule (present per-Space row wins even when disabled; else enabled global fallback; else off) is consistent across store, effective-config resolution, and delivery.
- **Input validation.** Message capped at 2000 runes and `active_from` at 40 chars on the write path (independent of enabled, avoiding driver 500 / silent truncation); composite validation via the shared validator; plain-text body (`payload.type=Text`, no markdown render). Errors use existing `httperr` codes only.
- **Tests.** Authz matrix (admin allowed / plain member / non-member / inactive space denied), CRUD happy path, invalid combination, over-length guards, empty-patch, partial-merge, delete-revert; and per-Space multi-Space no-crossing delivery, join-scoped-to-configured-Space, worker fair-rotation, and cross-Space sweep.

## 💬 Non-blocking

- 🟡 The CRUD DB calls derive from `context.Background()` with a fixed 3s timeout rather than the request context, so a disconnected client's query still runs to the timeout. Minor; consider a request-derived context if one is readily available on this handler path.
- 🔵 Consider adding an explicit HTTP-level test that DELETE restores an *enabled* global fallback (currently covered indirectly), to lock the full response contract.

## ⚠️ Merge gate

This PR currently reports a merge conflict against `main` (mergeable: CONFLICTING / DIRTY). Please rebase onto the latest base before merge — the conflict is with the base branch, not a defect in this change. Latest CI checks are green.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #646 (octo-server)

## Summary

This PR introduces per-Space welcome message configuration with admin CRUD endpoints, replacing the prior platform-global-only model. Space admins (Role >= 1) can manage their own welcome message via `/v1/space/:space_id/welcome` (GET/PUT/DELETE), with a per-Space row overriding the global fallback. The delivery orchestration in `modules/notify/space_welcome.go` is extended to iterate all enabled Spaces during reconciliation and worker sweeps, with proper budget caps and round-robin rotation.

The design is sound: authorization is tight (all three endpoints gated by `authorizeSpaceAdmin`, which checks space-active + Role >= 1), schema migrations are reversible, and the delivery path correctly resolves the effective config via `ResolveEffectiveSpaceWelcome` (per-Space row wins even when disabled, allowing admins to opt out of a global campaign). The concurrent-PUT protection via `UpsertMerged` (insert-then-lock + deadlock retry) is a good defensive pattern for admin CRUD.

Two P2 items and one Nit identified below — none are merge blockers.

## Verification

Static analysis at head `b9d9caa388e8b346ce772408f063001c8d77763f`:
- ✅ **Authorization gate** — `authorizeSpaceAdmin` at `modules/space/api_welcome.go:24` calls `checkSpaceActive` (verifies space exists and is active) then `requireSpaceAdmin` (verifies Role >= 1). All three handlers (GET/PUT/DELETE) use this gate.
- ✅ **Schema migrations** — `modules/common/sql/20260720000001_add_octo_space_welcome_config.sql` creates the config table with `UNIQUE KEY uk_space(space_id)` and `KEY idx_enabled(enabled)`. `modules/notify/sql/20260720000001_add_welcome_delivery_sweep_index.sql` adds `idx_sweep(status, claim_expire_at)` for cross-space sweep. Both have Down migrations.
- ✅ **Delivery path** — `handleMemberJoin` at `modules/notify/space_welcome.go:179` calls `effectiveConfig`, which delegates to `ResolveEffectiveSpaceWelcome` (per-Space row wins, else global fallback). Reconciliation and worker sweeps iterate `ListEnabled` with round-robin rotation and shared/per-space caps.
- ✅ **Concurrent PUT protection** — `UpsertMerged` at `modules/common/space_welcome_config_store.go:176` uses insert-then-lock + retry on InnoDB deadlock/lock-wait-timeout (1213/1205), preventing lost updates when two admins PUT the same Space concurrently.
- ✅ **Validation** — `putWelcome` validates message length via `utf8.RuneCountInString` against `SpaceWelcomeMessageMaxRunes` (2000) and active_from length against `SpaceWelcomeActiveFromMaxLen` (40), then calls `ValidateSpaceWelcomeCombination` (RFC3339 parse + non-empty message when enabled).

## Findings

Two P2 items and one Nit identified. No P0/P1 blockers.

### P2 — `Upsert` exported without concurrency caveat or production caller (`modules/common/space_welcome_config_store.go:141`)

`Upsert` is exported and tested, but no production write path uses it — the PUT handler goes through `UpsertMerged` (insert-then-lock + deadlock retry). A future caller reaching for `Upsert` would bypass the concurrent-lost-update protection that `UpsertMerged` provides for two admins editing the same Space.

Consider either making it unexported (`upsert`) or adding a doc comment:

```go
// Upsert is a simple idempotent write for single-writer / test contexts.
// Production admin CRUD must use UpsertMerged to prevent concurrent lost updates.
```

### P2 — Boot relaxation silently drops the docs-approval-cards-requires-card-messages invariant (`main.go:505`)

The old boot sequence hard-errored when `OCTO_DOCS_APPROVAL_CARD_ENABLED` was set without `OCTO_CARD_MESSAGE_ENABLED`. The new code logs a WARN and returns an inert runtime. The downstream paths (ingress reject, send-path refuse) are independently gated, so no callback can actually enqueue — the old check was defense-in-depth, not the sole guard. The relaxation is therefore safe in practice.

However, an operator who sees the WARN may not realize that docs approval cards are inert until the gate is re-enabled. Consider adding a changelog/release-note entry so the behavioral shift is visible in upgrade docs.

### Nit — `Enabled: true` hardcoded for per-space rows from `ListEnabled` (`modules/notify/space_welcome.go:179`)

This is correct today because `ListEnabled` filters `WHERE enabled=1`, so every returned row is enabled. But the hardcoded `true` is a derived invariant — if `ListEnabled`'s WHERE clause were ever relaxed, this would silently lie. Reading `r.Enabled` (converting from the row's int) would make the code self-documenting and resilient to a future query change.

## Verdict: APPROVED

The PR is well-structured, authorization is tight, schema migrations are reversible, and the delivery path correctly resolves effective configs. The two P2 items (exported `Upsert` without caveat, silent boot relaxation) and one Nit (hardcoded `Enabled: true`) are non-blocking — the first two are documentation/operator-experience improvements, and the third is a minor resilience tweak. No correctness, security, or build-breaking issues identified.

yujiawei (approved):
# Code Review — PR #646 (octo-server)

**Verdict: APPROVE.** Reviewed head `b9d9caa` against merge-base `054bfeec` (current `main` is `b379353d`; no file-level overlap with the touched paths, no sibling drift). Build / vet / gofmt / `git diff --check` all clean. Zero P0/P1 findings.

This is the per-Space + self-service generalization of the welcome feature shipped in #604/#606: config source goes from a single global tuple to per-Space (with global fallback), and the delivery driver goes single-Space → all-enabled-Spaces. The delivery reliability core (ledger `UNIQUE(space_id,uid)`, state machine, `FOR UPDATE SKIP LOCKED` claim, CAS-on-`claim_owner`, at-most-once, fixed `notification` sender) is unchanged.

## 1. Spec compliance

**Spec: ✅**

- **No under-build**: `GET/PUT/DELETE /v1/space/:space_id/welcome` (`modules/space/api_welcome.go:87,108,201`) + per-Space config table/store (`modules/common/space_welcome_config_store.go`) + multi-Space delivery (`modules/notify/space_welcome.go`) implement exactly the item the origin brief deferred ("Multi-Space configuration with per-Space copy; per-Space admin self-service API").
- **No over-build**: out-of-scope items respected — one config per Space (no `welcome_id`), global config kept as fallback (not removed), delivery semantics untouched, sender identity/wire protocol fixed, single plain-text body. `go.mod`/`go.sum`/`vendor/` unchanged (octo-lib untouched). No new `errcode` codes registered — all responses reuse existing codes via the `httperr` i18n envelope, matching the stated contract.
- **No deviation**: `enabled=false` per-Space default; with no per-Space row, behavior is identical to today (global config still drives its one Space). Precedence matches the brief: a present per-Space row wins outright *even when disabled* (`ResolveEffectiveSpaceWelcome`, store:335-348), else global iff enabled and naming the Space.

## 2. Code quality

**Quality: Approved**

Load-bearing verifications:

- **Concurrency (config write path) — correct.** `UpsertMerged` (store:176-278) uses insert-then-lock: an idempotent `INSERT ... ON DUPLICATE KEY UPDATE space_id=space_id` guarantees the row exists, then `SELECT ... FOR UPDATE` locks the now-present row before merge. Under InnoDB REPEATABLE READ this closes **both** the existing-row lost-update (second PUT blocks on the record lock, merges onto the first's committed write) **and** the concurrent first-create race (doing the INSERT before any gap-locking read avoids the gap-lock deadlock; a single `UNIQUE KEY uk_space` makes concurrent same-key inserts a linear wait, not a cycle). This directly resolves the first-create gap-lock-deadlock edge that the earlier `FOR UPDATE`-only approach left open. Transient `1213`/`1205` retried up to 3× (`isRetryableTxErr`, store:282-288); merge/validation errors are terminal and surfaced verbatim so callers' `errors.Is` still works. `created_at` preserved (not in the UPDATE set). Regression tests `..._NoLostUpdate` and `..._ConcurrentCreate` assert both admins' disjoint fields survive.
- **Space isolation / authz — correct.** Every CRUD handler resolves membership for the **path** `:space_id` via `requireSpaceAdmin` (Role ≥ 1, nil-member fail-closed, `api_email_invite.go:169`) after `checkSpaceActive`; a request can only affect the path Space. Authz matrix test covers admin/member/non-member/inactive.
- **Cross-space delivery — no leak, still at-most-once.** `claimOne` is `WHERE space_id=?`-scoped (`space_welcome_db.go:53`); `dispatch` re-checks the recipient and builds the payload from **that row's own** Space effective config (`cur.SpaceID`, space_welcome.go:474-544), so one Space's config/backlog can never deliver into another. The cross-space `sweepClaimedAll`/`sweepDispatchingAll` drop the `space_id` filter but only touch **lease-expired** rows and null out `claim_owner`, so a stale worker's subsequent `casToDispatching`/`casToSent` CAS fails before sending — no double-send. `idx_sweep (status, claim_expire_at)` keeps the sweep index-backed on a growing ledger. `TestPerSpace_MultiSpaceDelivery` asserts each recipient gets only their own body.
- **Fairness — bounded.** Both reconciler (`reconcileCursor`) and worker (`workerCursor`) rotate the per-cycle starting Space; each cursor is touched by a single goroutine (no lock needed). Shared global budget + per-Space sub-cap prevent a backlogged head Space from starving the tail. `TestPerSpace_Worker_FairRotation` fails without the cursor.
- **Guards / hygiene**: `message` (≤2000 runes) and `active_from` (≤`VARCHAR(40)`) column-width guards run regardless of `enabled` (space_welcome.go:158-165), returning a clean 400 instead of a driver 500 / silent truncation; `updated_at` emitted as stable RFC3339 UTC; migration `COLLATE=utf8mb4_general_ci` aligns with `space.space_id` (avoids MySQL 1267); all timestamps app-computed UTC bound params (no `NOW()`). Dead per-space sweep helpers removed and their unit tests repointed to the `*All` variants. `TestSpaceNoLegacyResponseError` guard updated to include `api_welcome.go`.

## Notes for a human verifier (non-blocking)

- DB-backed concurrency/integration tests require live MySQL/Redis/WuKongIM and cannot run in a sandbox; the concurrency conclusions above rest on source + InnoDB locking analysis + the store-level regression tests. Worth a human confirming the `-race` suite is green in CI on the deploy DB, and (optionally) that `innodb_lock_wait_timeout` is comfortably above the CRUD 3s context timeout so the 3× retry has room.
- `enabled=false` per-Space default means this ships inert until a Space admin opts in — a safe rollout.

## Advisory legs

One additional adversarial reviewer converged with this analysis (no blocking findings: `UpsertMerged` serialization correct, authz path-scoped, cross-space sweep safe via `claim_owner` CAS, precedence correct). A second independent leg was **absent this run (timed out)** — recorded as absent, not silent agreement; the isolation/concurrency conclusions rest on the direct source analysis above.

yujiawei:
Head advanced to `ccd577f6` during review — this is a **merge of `origin/main` into the branch**, not new feature work. Confirmed:

- The welcome-feature files (`space_welcome_config_store.go`, `api_welcome.go`, `space_welcome.go`, `space_welcome_db.go`, `api.go`, both migrations) are **byte-identical** to the reviewed `b9d9caa` (`git diff b9d9caa ccd577f6 -- <feature files>` is empty).
- New merge-base is current `main` (`b379353d`). Full `go build ./...` + `go vet` + `gofmt` are **clean at `ccd577f6`** — the incoming sibling changes (cardtmpl / incoming-webhook / message-reaction / revoke-strip) merge without semantic conflict against this feature; the `SpaceWelcomeConfig` / `ValidateSpaceWelcomeCombination` / `ParsedActiveFrom` / `EnsureSystemSettings` APIs it depends on all survived the `system_settings.go` merge.

**Verdict unchanged: APPROVE** for head `ccd577f6` (post-merge). The analysis in the formal review above applies verbatim.

Jerry-Xin (approved):
Re-review after the branch was refreshed to resolve the earlier merge conflict. This confirms the per-Space welcome message + admin CRUD substance is unchanged and the conflict resolution is clean; carrying forward the prior approval.

## Delta since the previously-approved head

The new head is a merge commit that pulls `origin/main` into the branch (two parents: the previously-approved PR head and the current main tip). All 15 PR-owned welcome-message files are byte-identical to the previously-approved head (verified by blob hash): the merge only brought in unrelated main drift plus a manual resolution of two non-code files (`.octospec/log.md`, `docs/ascci logo.md`). No welcome-message code changed.

## Re-verified load-bearing surface (still intact after the merge)

- AuthZ unchanged: `getWelcome` / `putWelcome` / `deleteWelcome` in `modules/space/api_welcome.go` each call `authorizeSpaceAdmin` first (= `checkSpaceActive` + `requireSpaceAdmin`, Role >= 1) and bail on failure. Routes are still mounted on the authenticated + UID-rate-limited group in `modules/space/api.go` (`GET/PUT/DELETE /v1/space/:space_id/welcome`). The merge did not shift the middleware or route registration.
- Multi-tenant isolation / IDOR: target `spaceId` is always `c.Param("space_id")` (never the client body), and all store/delivery queries in `modules/notify/space_welcome_db.go` are scoped by `space_id`.
- Delivery path: enqueue is idempotent (dedup upsert), config is re-resolved before every claim so a mid-wake disable/delete stops further claims, and the pre-send re-check re-validates human + still-active-member before sending; enqueue failure never blocks the join flow.
- CI is green (Build / Test / Vet / Lint / i18n / security scans all pass) and `mergeable` is now clean.

## Non-blocking notes

- The two earlier notes remain non-blocking: API DB calls use `context.Background()` with a fixed timeout rather than the request context, and there is no explicit HTTP-level test for the DELETE-to-global-fallback path (unit coverage exists).
- `active_from` is applied as an enqueue-time audience cutoff (`firstJoinHumanMember` / `reconcileScan` gate `created_at >= active_from`), and the delivery-time re-check validates human + membership + config-enabled but does not re-compare the member's join time against a later `active_from`. This means a member who was correctly in-audience at enqueue time can still be delivered to if an admin subsequently advances `active_from` (or deletes the per-Space config and exposes a global fallback with a later window). Given the enqueue-time semantics this is acceptable, but if forward-moving `active_from` should retroactively suppress already-queued welcomes, add that comparison to the pre-send eligibility query plus a regression test.

## Highlights

- Space-scoped authorization uses active membership + Role >= 1.
- Per-Space / global precedence is clearly encoded and tested.
- Insert-then-lock merging addresses concurrent partial-update races.
- Multi-Space rotation and cross-Space lease sweeping are well structured.
- Routes use the localized error envelope and the shared UID rate limiter.

lml2468 (approved):
# PR Review (re-review) — PR #646 (Mininglamp-OSS/octo-server)

Re-reviewed at **head `ccd577f6`** (a `Merge origin/main` commit) against base `main`. Re-affirms my APPROVE at `b9d9caa3` (auto-dismissed when this merge advanced the head; this restores a live verdict at the current head).
**Build/vet/test:** `go build ./modules/space/ ./modules/notify/ ./modules/common/` exit 0 ✅, `go vet` ✅. DB-gated welcome suites (config-store concurrency, api_welcome, per-space delivery) → CI.

## What changed since `b9d9caa3` (delta reviewed)
The new head `ccd577f6` is **purely a merge of `origin/main`** into the branch — pulling in the meanwhile-merged #641/#642/#643/#638/#634 etc. This **resolves the `merge: DIRTY` state** I flagged in my prior review.

**#646's own welcome production code is byte-identical to my approved `b9d9caa3`**: `git diff b9d9caa3 HEAD -- api_welcome.go api.go space_welcome.go space_welcome_db.go space_welcome_config_store.go` = **0 lines**. The merge did not touch the CRUD handlers, the store, or the drive loop; conflict resolution only touched non-code files (octospec log / docs). No real conflict markers remain in `modules/`; the whole tree builds and vets clean after the merge (confirming the welcome code integrates with the newly-merged main).

## Verified clean (carry-over from `b9d9caa3`, re-confirmed at merged head)
- **Isolation by construction:** every CRUD handler gates on `authorizeSpaceAdmin` (`checkSpaceActive` + `requireSpaceAdmin`, Role≥1), path-`:space_id`-only writes, no caller-supplied target space.
- **Concurrency fixes intact:** `UpsertMerged` read-merge-write under `SELECT … FOR UPDATE` (lost-update fix) + insert-then-lock idempotent materialize (first-create race) + `isRetryableTxErr` retry — the two defects I flagged on the sibling #622, both closed here.
- Deterministic precedence (per-Space row wins even disabled → global-if-names-space → off), at-most-once delivery preserved, cursor-rotation fairness, collation-aligned additive migrations (`uk_space`, `idx_sweep`), message ≤2000-rune guard — all unchanged and re-confirmed byte-identical.

## Verdict
**APPROVE** (re-affirmed at `ccd577f6`) — the new head is a clean merge of main that resolves the prior DIRTY state with **zero change to #646's own welcome code** (byte-identical to my `b9d9caa3` approval); the merged tree builds/vets clean and has no unresolved conflict markers. All the load-bearing properties I verified (space-isolation, the two concurrency fixes, precedence, at-most-once, migrations) still hold. Both gates pass.
Suggested decision: `Safe to merge` once CI runs the DB-gated welcome suites green. `merge: MERGEABLE` now (DIRTY resolved); `BLOCKED` is branch-protection / required approvals, not a code issue. (Note: my prior review auto-dismissed on the head advance — this is my live verdict at `ccd577f6`.)

PR #647, opened 2026-07-23
feat(card): align docs access request v3 with Forge layout

LLwill:
## Summary

Align `docs.access-request@0.3.0` with the latest Forge visual structure while preserving the existing server-owned action and denial contracts. Add an optional stable render-profile marker so the web client can select the matching HostConfig/CSS generation without retaining exact package versions.

## Related Issue

N/A

## Linked Spec

- `.octospec/tasks/docs-card-forge-align/brief.md`

## Changes

- Render the 0.3.0 pending and terminal views with a compact header, full-bleed accent content surface, semantic badges, and full-bleed footer.
- Reuse the existing `docs-access-approve`, `docs-access-deny`, action data, and hidden `deny_reason` contract; the host-owned denial dialog remains unchanged.
- Validate inline footer ActionSets in the registry without changing callback routing, authorization, idempotency, or finalization.
- Carry optional `render_profile: "octo-chat/v1"` through render, dispatch, and card replacement updates.
- Keep the exact `octo-chat@1.2.0-rc.1` artifact as manifest provenance only.
- Leave `docs.access-request@0.2.0` unchanged and without a render-profile marker.

Merge prerequisite: confirm that the existing 0.3.0 output has not been sent in production; otherwise this visual correction requires a new template version.

## Testing

- [x] Unit tests added/updated
- [x] Manually verified

Commands:

- `go vet ./pkg/cardmsg ./pkg/cardtmpl/... ./internal/carddispatch ./modules/notify`
- `go test ./pkg/cardmsg ./pkg/cardtmpl/... ./internal/carddispatch -count=1`
- `go test ./modules/notify -run 'Test(DocsActionFinalizer|StandardActionFinalizer|BuildDocsAccessRequest|DeliverDocsAccessRequest|PreflightDocsAccessRequest|DocsOutcomeCards|Integration_NotifyTokens|Integration_GenericApproval)' -count=1`
- `pnpm cli check docs.access-request@0.3.0` in octo-card-forge
- `pnpm cli profile validate octo-chat@1.2.0-rc.1` in octo-card-forge

## COMPREHENSION

1. **What does this change actually do** to the load-bearing path?

   The registry still produces the authoritative Adaptive Card and action data. For 0.3.0 it now places those existing actions in the Forge-aligned footer and emits a stable optional render-profile key. Dispatch and replacement updates preserve that key in the type-17 envelope.

2. **What could break** because of it (dependents + failure mode)?

[...]

PR #649, opened 2026-07-23
feat(cardtmpl): migrate docs.commented + docs.shared to Registry (roadmap C PR-1)

an9xyz:
## Summary

Roadmap C first slice — migrate the docs-notify v1 display cards
(`docs.commented`, `docs.shared`) onto the cardtmpl Registry using the
`docs.access-request` pilot's copy-directory shape.

**Milestone reached**: §2.2-5 L2b hard gate ② — `docs.access-request` (v2) +
`docs.commented` (v1) + `docs.shared` (v1) = **3 L2a cards** running the full
Registry path across **both wire profiles**.

External `NotifyReq` / `DocsCardFields` shape is unchanged (plan B).

## Related Issue / Prior art

Roadmap C follow-up to #641 (post-#633 interactive-card loop). No standalone
issue.

## Linked Spec

`.octospec/tasks/cardtmpl-l2a-migration/brief.md`

## Changes

- **`pkg/cardtmpl/docs_commented@0.1.0`** — new subpackage: self-embedded
  `handoff/` (manifest declaring `renderProfile` + `renderProfileCompatibility`
  per #647, `contract/data.schema.json`, `samples/shown.json`) + 3-method
  Template. Build composes `BuildDocsResourceCardBodyWithLang` so bodies are
  byte-identical to legacy `buildDocsCard`.
- **`pkg/cardtmpl/docs_shared@0.1.0`** — symmetric subpackage, differs only in
  `Variant` + attribution copy.
- **notify wiring** (`modules/notify/card_via_registry.go`,
  `modules/notify/card.go`): new `buildDocsDisplayCardViaRegistry` +
  `preflightDocsDisplaySchema`; `deliverDocsCardNotification` routes
  `DocsCardKindCommented` / `DocsCardKindShared` through Registry with F7
  fail-close and C1 preflight (schema errors → 400 zero-delivery) before
  member/gate checks.
- **`main.installCardTmplRegistry`** registers both new cards +
  `SetDefault` + `Freeze` alongside the pilot.
- **`pkg/cardtmpl.SanitizeLine`** — G5 single source of truth; the two prior
  copies (notify + pilot) become one-line wrappers.
- **`BuildSummaryResourceCardBodyWithLang`** — scaffold for PR-2
  (`summary.completed` / `summary.failed`).
- **`modules/notify/testmain_test.go`** — `TestMain` wires the Registry once
  for the whole test package (mirrors production wiring so preflight tests
  inherit it).

## Testing

Locally green with MySQL 8 / Redis 7 / WuKongIM:

- `go build ./...`, `go vet ./...`
- `pkg/cardtmpl/docs_commented` -race -cover → **85.1%**
- `pkg/cardtmpl/docs_shared` -race -cover → **80.9%**
- `pkg/cardtmpl/...` full package tree — all green
- `modules/notify` full package (after `DROP DATABASE test; CREATE DATABASE
  test` to reset the cross-package migration ledger, per
  `.claude/memory/local_test_infra.md`) — all green including the 4-case
  byte-equivalence baseline and the C1 preflight assertion
- `make i18n-extract-check`, `make i18n-lint`

**Byte-equivalence baseline is the core acceptance.**
`modules/notify/card_via_registry_display_baseline_test.go` runs 4 fixtures
(commented/shared × zh full/anon) through both legacy `buildDocsCard` and
`buildDocsDisplayCardViaRegistry`, strips the two newly-injected metadata
fields (`metadata.octo.protocol`, `metadata.octo.template`), and asserts
canonical JSON byte-equality. Any drift (copy / ordering / alias / extra
field) fails immediately.

Honest gaps (not run here):

- `golangci-lint` — not installed on the dev host; CI lint lane covers it.
- The `//go:build integration` orchestration e2e in `modules/message` has a
  pre-existing compile-time import cycle and the CI test lane runs the default
  build without `-tags integration`; unchanged by this PR.

- [x] Manually verified (focused suites + byte-equivalence baseline)

## COMPREHENSION

   `deliverDocsCardNotification` now routes two more docs card kinds through
   `Registry.Render`: preflight validates the input against each card's
   `data.schema.json` (C1 → 400 zero-delivery), and the card is rendered by
   the new Templates whose `Build` composes the same
   `assembleResourceCardBody` the legacy path already uses. External
   `NotifyReq`/`DocsCardFields` are unchanged. `metadata.octo.{protocol,
   template}` + `render_profile` are added to the envelope; everything else in
   the card node is byte-identical.

   - *Boot*: Registry register-time self-check runs each new sample through
     `Render`; a bad handoff fails `installCardTmplRegistry` → `main` panics
     at startup. Mitigation: self-check + focused tests pass locally.
   - *In-flight card renderers on the receiving side*: envelopes now carry
     new `metadata.octo` and `render_profile` fields. Additive only — unknown
     fields must be ignored per JSON forward-compat; verified against the
     baseline.
   - *docs backend integration*: `DocsCardFields` shape unchanged, so
     backend calls keep working. `docsSafeExcerpt`/`ActorName`/`UpdatedAt`
     semantics preserved.
   - *Existing tests*: any notify test using `validDocsCard()` (Kind: shared)
     now needs a wired Registry — handled once by the new package-level
     `TestMain`.

   `card_via_registry_display_baseline_test.go` runs 4 fixtures end-to-end
   through both paths and asserts canonical JSON byte-equality (see Testing).
   `pkg/cardtmpl/docs_commented/template_test.go` and
   `pkg/cardtmpl/docs_shared/template_test.go` cover Registry self-check,
   metadata injection, language-localized source, `FallbackText`
   (named/anon/sanitize), and the C1 schema-rejection branch. `modules/notify`
   full package (real MySQL/Redis/WuKongIM) green after the standard
   cross-package DB reset.

## Out of scope

- `summary.completed` / `summary.failed` (PR-2, scaffold in place).
- `generic.approval` — its dynamic caller-supplied `owner`/`action_type`
  conflicts with the static `TemplateActionContract`; separate adaptation
  decision required.
- No `NotifyReq` shape change, no `${}` engine, no B/D/E/H, no DB migration.

## Checklist

- [x] Updated documentation (`.octospec/` brief + journal + log)

### DIFF
--- a/.octospec/journal/shared/cardtmpl-l2a-migration.md
+++ b/.octospec/journal/shared/cardtmpl-l2a-migration.md
@@ -1,0 +1,122 @@
+---
+type: Journal
+title: "Journal: cardtmpl-l2a-migration (PR-1: docs.commented + docs.shared)"
+description: Roadmap C first slice — migrate docs.commented and docs.shared display cards onto the cardtmpl Registry using the pilot copy-directory shape. Together with the already-migrated docs.access-request they reach the §2.2-5 L2b hard gate ② (≥3 L2a cards, v1 + v2 coverage). Also extracts the shared SanitizeLine helper (G5) and lays a summary-side fragment helper for PR-2.
+tags: ["cardtmpl", "platform", "docs-commented", "docs-shared", "l2a-migration", "wire-contract", "i18n", "error-response", "escape", "markdown", "url-destination", "testing"]
+timestamp: 2026-07-23T12:30:00+08:00
+# --- octospec extension fields ---
+task: cardtmpl-l2a-migration
+upstream: self (roadmap C)
+source: self
+---
+
+# Journal: cardtmpl-l2a-migration (PR-1)
+
+## What was done
+
+First slice of roadmap group C — two new L2a display cards migrated onto
+`Registry.Render` behind the docs-notify plan-B shell, plus one shared helper
+and one PR-2 scaffold:
+
+1. **`pkg/cardtmpl/docs_commented@0.1.0`** — new subpackage with self-embedded
+   handoff (`manifest.json` incl. `renderProfile`/`renderProfileCompatibility`,
+   `contract/data.schema.json`, `samples/shown.json`) and a 3-method Template
+   (Meta/Build/FallbackText). Build composes the base
+   `BuildDocsResourceCardBodyWithLang` (already extracted upstream), so the
+   rendered body is byte-identical to legacy `buildDocsCard` for the same
+   inputs. 85.1% coverage; -race green.
+2. **`pkg/cardtmpl/docs_shared@0.1.0`** — symmetric subpackage, differs from
+   commented only in `Variant` + attribution copy. 80.9% coverage; -race green.
+3. **notify wiring** (`modules/notify/card_via_registry.go` +
+   `modules/notify/card.go`): new `buildDocsDisplayCardViaRegistry` +
+   `preflightDocsDisplaySchema` helpers; `deliverDocsCardNotification` routes
+   `DocsCardKindCommented` / `DocsCardKindShared` through Registry with F7
+   fail-close (Registry-unwired → 500, not silent legacy). The
+   `NotifyReq`/`DocsCardFields` external shape is unchanged (plan B).
+4. **`main.installCardTmplRegistry`** registers both new cards + `SetDefault` +
+   `Freeze` alongside the pilot; `modules/notify` gets a `TestMain` that
+   mirrors the production wiring so preflight tests inherit the same registry.
+5. **G5**: `SanitizeLine` exposed on `pkg/cardtmpl/resource.go`; both prior
+   copies (`modules/notify` + `pkg/cardtmpl/docs_access_request`) become
+   thin wrappers. Word list drift is now impossible — single source.
+6. **PR-2 scaffold**: `BuildSummaryResourceCardBodyWithLang` added so the
+   summary migration in PR-2 can compose the base body without touching the
+   legacy summary path.
+
+## Byte-equivalence baseline (the core acceptance)
+
+`modules/notify/card_via_registry_display_baseline_test.go` runs 4 fixtures
+(`commented` × zh full/anon, `shared` × zh full/anon) through both **legacy
+`buildDocsCard`** and **`buildDocsDisplayCardViaRegistry`**, strips the two
+newly-injected metadata fields (`metadata.octo.protocol` +
+`metadata.octo.template`), and asserts canonical JSON byte-equality. Everything
+else — body, facts, ActionSet, attribution, variant, source, webUrl — is
+identical. Any future drift (copy, ordering, alias, extra field) fails this
+test immediately.
+
+`Action.OpenUrl.id` is intentionally NOT added on v1 cards (only the v2
+`access_requested` pilot adds `view_document` because its interaction report
+requires it). No id to strip on v1.
+
+## Structural learning
+
+- **The base already anticipated this migration.** `assembleResourceCardBody`
+  was extracted upstream so legacy and Registry paths share the exact same
+  body assembler; `BuildDocsResourceCardBodyWithLang` had a comment naming
+  `pkg/cardtmpl/docs_commented` explicitly. PR-1's Go footprint was therefore
+  minimal — each Template is a thin i18n + fixture-mapping shim on top of the
+  shared body. Confirms the "copy pilot directory" pattern scales; the
+  remaining L2a cards are similarly cheap.
+- **§2.2-5 L2b hard gate ② is now met.** Three L2a cards (`docs.access-request`
+  v2 + `docs.commented` v1 + `docs.shared` v1) run the full Registry path
+  covering both wire profiles. This is prerequisite, not sufficient — D
+  (`ext.*` production channel) still needs the visibility/token plumbing.
+
+## Gotchas worth remembering
+
+- **`TestMain` global Registry mirrors production wiring.** Any test that runs
+  `deliverDocsCardNotification` for the commented/shared/access_requested
+  kinds now needs a wired Registry (F7). A single `TestMain` under
+  `modules/notify` beats retrofitting Registry setup into 10+ tests
+  individually. Tests that intentionally verify unwired behavior can
+  `SetDefaultRegistry(nil)` locally.
+- **Cross-package migration ledger still bites `-count=1` runs.** After
+  running any test package that uses `testutil.NewTestServer`, another package
+  may hit `panic: Table 'seq' already exists ... sql-migrate`. Recovery is
+  documented in memory `local_test_infra.md`: `docker exec octo-test-mysql
+  mysql -uroot -pdemo -e "DROP DATABASE IF EXISTS test; CREATE DATABASE test
+  ..."`. This is not introduced by C; it's the CI lane's per-package
+  drop-and-recreate mirror running locally.
+- **English copy alignment.** The migrated card's per-language `labelSet` must
+  match `modules/notify.docsLabelsFor` byte-for-byte (banner / actor / time
+  labels / kvSep), or the byte-equivalence baseline fails. Caught one during
+  PR-1 (initial `bannerAnon` / `actor` / `updatedAt` drift). Copying the exact
+  entries from the legacy label set before hand-editing prevents the miss.
+
+## Out of scope (deliberate, deferred to later PRs)
+
+- `summary.completed` / `summary.failed` (PR-2 — scaffold in place).
+- `generic.approval` — dynamic caller-supplied `owner`/`action_type` conflicts
+  with the static `TemplateActionContract`; needs a separate adaptation
+  decision (fixed-owner template vs preserved dynamic special-case) before it
+  can move to the Registry.
+- No `NotifyReq` external shape change / no envelope mode (roadmap E2).
+- No `${}` JSON engine (E1), no capability-discovery endpoint (B), no L2b
+  channel (D), no unfurl (H).
+- No DB migration.
+
+## Verification (this run, real infra)
+
+Locally green with MySQL 8 (docker :3306) / Redis 7 (:6379) / WuKongIM (:5001):
+
+- `go build ./...`, `go vet ./...`
+- `pkg/cardtmpl/docs_commented` -race -cover → **85.1%**
+- `pkg/cardtmpl/docs_shared` -race -cover → **80.9%**
+- `pkg/cardtmpl/...` (base + pilot + new subpackages) all green
+- `modules/notify` full package (after `DROP DATABASE test; CREATE DATABASE
+  test` to reset the cross-package migration ledger) — all green including
+  the 4-case byte-equivalence baseline and the C1 preflight assertion
+- `make i18n-extract-check`, `make i18n-lint` — clean
+- `golangci-lint` not installed on this host; the CI lint lane covers it.
+- Message `//go:build integration` orchestration e2e — pre-existing import
+  cycle, not run locally nor in the CI default lane (unchanged by this PR).
--- a/.octospec/journal/shared/cardtmpl-l2a-migration.md
+++ b/.octospec/journal/shared/cardtmpl-l2a-migration.md
@@ -12,6 +12,52 @@ source: self
 
 # Journal: cardtmpl-l2a-migration (PR-1)
 
+## Review round (PR #649, 3× APPROVE + 1 CHANGES_REQUESTED → addressed)
+
+`lml2468` / `Jerry-Xin` / `mochashanyao` approved; `yujiawei` requested changes
+on one blocking item. Addressed in-branch:
+
+- **P1 (blocking, yujiawei) — over-length display fields hard-rejected (400,
+  zero delivery) where legacy truncated & delivered.** The new schemas cap
+  `excerpt`/`actorName`/`updatedAt`/`title`; `preflightDocsDisplaySchema` runs
+  before member/gate checks, so a 301-rune comment (realistic) now 400s and
+  *every* recipient loses the notification — not even the text fallback fires.
+  Legacy `buildDocsCard` never rejected on length (it truncated at render).
+  This is the "unchanged wire shape ≠ unchanged accepted-input behavior" trap
+  and the baseline can't catch it (all fixtures in-bounds). Fix, per the brief's
+  own G9 wording ("truncate before render"): `mapDocsCardFieldsToDisplayJSON`
+  now server-side truncates `title`→`cardtmpl.MaxTitleRunes`,
+  `actorName`→120, `excerpt`→`cardtmpl.MaxExcerptRunes`, `updatedAt`→80 before
+  preflight, preserving delivery. `docId` is **not** truncated — it's the
+  deep-link key, so an over-length docId stays a genuine C1 400 (same handling
+  as summary `taskNo`). Exported `cardtmpl.MaxTitleRunes` + `cardtmpl.TruncateRunes`
+  so ingress and render share one cap/impl (G9). Within-budget inputs unchanged
+  → canonical-JSON baseline still holds;
+  `TestMapDocsDisplayFields_TruncatesDisplayFields` locks the new behavior
+  (huge input delivers, over-length docId 400s).
+- **P2 — dangling reference to a non-existent "docs_commented G9 field-cap
+  test".** `resource.go` claimed the excerpt cap and schema `maxLength` "never
+  drift (see the ... test)" but no such test existed. Added
+  `TestSchemaCapsMatchRenderCaps` to both `docs_commented` and `docs_shared`
+  (assert schema `excerpt.maxLength == MaxExcerptRunes`, `title.maxLength ==
+  MaxTitleRunes`) and reworded the comment to name the real test.
+- **P2 — "byte-equivalence" over-claim.** The baseline unmarshal→canonical→
+  remarshal proves *canonical-JSON / semantic* equality, not literal wire-byte.
+  Reworded the test comment to say so (kept the function name; it's the right
+  drift guard, just not literal bytes).
+- **P2 (mochashanyao/Jerry-Xin) — fallback delegation gap** (commented/shared
+  text fallback still legacy, not `Template.FallbackText`) and **Nit — no
+  `docs_shared` en test**: added `TestRenderEnglishSourceLocalized` for
+  `docs_shared`; the fallback-delegation refactor is deferred to the roadmap C
+  finalization PR (copy is byte-identical today, non-blocking).
+- **Nit — gofmt drift** in `docs_shared/template.go`: `gofmt -w`.
+- Also made the docs build-error log label kind-generic (+`zap kind`); it
+  previously hard-coded "docs access-request card" for commented/shared too.
+
+Dismissed (yujiawei's own): F7-fail-open-to-text is intended F6/§10 for v1
+display cards (matches legacy); actor-hydration-after-preflight residual (a
+>120-rune hydrated name) collapses into the P1 truncation fix.
+
 ## What was done
 
 First slice of roadmap group C — two new L2a display cards migrated onto
--- a/.octospec/log.md
+++ b/.octospec/log.md
@@ -24,6 +24,15 @@ change-log convention (§7). Newest first.
   `metadata.octo` fields. `modules/notify/testmain_test.go` wires the
   Registry once for the whole test package (mirrors production wiring). See
   [journal](journal/shared/cardtmpl-l2a-migration.md).
+- **Review fix (P1, PR #649)** — Over-length display fields were flipping to a
+  400 zero-delivery under the C1 preflight where legacy truncated & delivered.
+  `mapDocsCardFieldsToDisplayJSON` now server-side truncates title/actorName/
+  excerpt/updatedAt to the schema/render caps before preflight (delivery
+  preserved); `docId` stays a hard C1 400 (deep-link key). Exported
+  `cardtmpl.MaxTitleRunes` + `cardtmpl.TruncateRunes` (single cap/impl, G9);
+  added `TestSchemaCapsMatchRenderCaps` (closes the previously-dangling G9
+  field-cap reference) + `TestMapDocsDisplayFields_TruncatesDisplayFields` +
+  `docs_shared` en test; made the docs build-error log label kind-generic.
 
 ## 2026-07-22 (incoming-webhook-quota-per-thread)
 
--- a/.octospec/log.md
+++ b/.octospec/log.md
@@ -4,6 +4,27 @@ Change history for this repo's `.octospec/`, following the
 [OKF](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md)
 change-log convention (§7). Newest first.
 
+## 2026-07-23 (cardtmpl-l2a-migration PR-1)
+
+- **Feature** — Roadmap C first slice: `docs.commented@0.1.0` and
+  `docs.shared@0.1.0` (v1 display cards) migrated onto `Registry.Render` via
+  the pilot copy-directory shape. `NotifyReq`/`DocsCardFields` external shape
+  unchanged (plan B); `deliverDocsCardNotification` routes both kinds through
+  Registry with F7 fail-close, matching `access_requested`. Both manifests
+  declare `renderProfile` + `renderProfileCompatibility` (#647).
+- **Milestone** — §2.2-5 L2b hard gate ② is now met: `docs.access-request`
+  (v2) + `docs.commented` (v1) + `docs.shared` (v1) = 3 L2a cards running
+  the full Registry path across both wire profiles.
+- **Base helpers** — `pkg/cardtmpl.SanitizeLine` is the single source of
+  truth (G5, notify + pilot become wrappers).
+  `BuildSummaryResourceCardBodyWithLang` added to scaffold PR-2.
+- **Test** — `card_via_registry_display_baseline_test.go` runs 4 fixtures
+  through legacy `buildDocsCard` vs `buildDocsDisplayCardViaRegistry` and
+  asserts canonical byte-equality after stripping the two injected
+  `metadata.octo` fields. `modules/notify/testmain_test.go` wires the
+  Registry once for the whole test package (mirrors production wiring). See
+  [journal](journal/shared/cardtmpl-l2a-migration.md).
+
 ## 2026-07-22 (incoming-webhook-quota-per-thread)
 
 - **Behavior change** — Incoming Webhook creation quota re-scoped from
--- a/.octospec/tasks/cardtmpl-l2a-migration/brief.md
+++ b/.octospec/tasks/cardtmpl-l2a-migration/brief.md
@@ -1,0 +1,133 @@
+---
+type: Task
+title: "Task: cardtmpl-l2a-migration"
+description: Migrate the remaining L2a notification cards (docs.shared/commented, summary.completed/failed) onto the cardtmpl Registry using the docs.access-request copy-directory shape, clearing the legacy builders and reaching the ≥3-card L2b gate with both v1/v2 view modes covered.
+tags: ["wire-contract", "i18n", "error-response", "escape", "markdown", "url-destination", "trust-boundary", "testing", "commit"]
+timestamp: 2026-07-23T09:30:00+08:00
+# --- octospec extension fields ---
+slug: cardtmpl-l2a-migration
+upstream: self (roadmap C)
+source: self
+---
+
+# Task: cardtmpl-l2a-migration
+
+> Roadmap group C. Migrate the remaining L2a notification cards onto the
+> cardtmpl Registry, copying the `docs.access-request` pilot directory shape.
+> Prior art: `.octospec/tasks/cardtmpl-registry-pilot/brief.md` (#633),
+> `.octospec/tasks/cardtmpl-interaction-closure/brief.md` (#641),
+> #647 (renderProfile). Authoritative contract: `docs/platform-card-base.md`.
+
+## Goal
+
+Move the notification cards still built by legacy builders — **docs.shared,
+docs.commented, summary.completed, summary.failed** (all `octo/v1` display
+cards) — onto the cardtmpl Registry, each as a `pkg/cardtmpl/<card>/` subpackage
+with its own embedded `handoff/<id>@<ver>/` and a 3-method Template, exactly like
+the `docs.access-request` pilot.
+
+Two payoffs:
+1. **Clear the legacy builders**: the corresponding variants in
+   `pkg/cardtmpl/{resource,...}.go` and the deliver branches in
+   `modules/notify/card.go` route through the single `Registry.Render` entry.
+2. **Reach the L2b hard gate** (§2.2-5 gate ②): together with the already-migrated
+   `docs.access-request` (v2 interactive), ≥3 L2a cards run the full Registry
+   path covering **both** view modes — v1 display (these new cards) and v2
+   interactive (docs.access-request).
+
+The external `NotifyReq` shape does not change (plan-B shell unchanged; only the
+internal `deliver*` path moves to Registry).
+
+## Background
+
+- The pilot (#633) established the shape: each card lives in
+  `pkg/cardtmpl/<card>/` with its own `handoff/<id>@<ver>/`
+  (manifest / contract / samples / reports) + Template impl; new cards copy it.
+- #647 introduced **renderProfile / renderProfileCompatibility** (dual profile:
+  `profile` = message capability, `render_profile` = which HostConfig/CSS
+  generation the web should use). **Every new card's manifest must declare
+  `renderProfile` (exact Forge artifact) + `renderProfileCompatibility` (stable
+  key, e.g. `octo-chat/v1`)**, or it emits no `render_profile` and renders as
+  Legacy forever (contract §3; register-time `cardmsg.IsAcceptedRenderProfile`).
+- Current state: `docs.access-request` is migrated (v2). `docs.shared/commented`
+  and `summary.completed/failed` are still legacy (variants in
+  `modules/notify/card.go` + `pkg/cardtmpl/resource.go`).
+- **`generic.approval`** (`pkg/cardtmpl/approval_request.go`) takes a **dynamic
+  `owner`/`action_type`** from the caller, which conflicts with the Registry's
+  **static `TemplateActionContract`** model → see Out of scope; not migrated
+  here.
+
+## Load-bearing list
+
+- **L1 directory shape & freeze (`wire-contract`)**: new cards self-embed
+  `pkg/cardtmpl/<card>/handoff/<id>@<ver>/`; a published version is frozen
+  (§2.1). The published `docs.access-request@0.2.0/0.3.0` trees are not touched.
+- **#647 renderProfile declaration (`wire-contract`)**: each new manifest
+  declares `renderProfile` + `renderProfileCompatibility`; register-time
+  `IsAcceptedRenderProfile` validates (illegal value panics); the send/update
+  envelope emits `render_profile` accordingly.
+- **notify deliver branches (`wire-contract`)**: `modules/notify/card.go`
+  currently routes `access_requested` through the Registry and the rest through
+  legacy. Move the `docs.shared/commented` + `summary.*` branches to
+  `Registry.Render`; the `NotifyReq` / internal notify token shell is unchanged.
+- **C1 zero-delivery (`error-response`)**: schema-level field errors on the
+  migrated cards → HTTP 400 zero-delivery (`ErrFieldsInvalid`), no fabricated
+  fallback text; preflight moved ahead of member/gate checks (as the pilot F1).
+- **Byte-equivalence baseline (`wire-contract`)**: canonical JSON diff before/after
+  migration may differ ONLY in `metadata.octo.{protocol,template}` +
+  `render_profile` + `Action.OpenUrl.id` (as pilot A11/F4).
+- **Render safety (`escape`, `markdown`, `url-destination`)**: caller-controlled
+  text goes through `escapeMarkdown`; all URLs absolute https; enforced by L0
+  `Registry.Render`, not relaxed.
+- **Fallback text**: each new Template's `FallbackText` is the L0 authoritative
+  fallback; the master-switch / gate / render-failure degrade paths are unchanged.
+- **i18n (`i18n`)**: labels from `BuildEnv.Lang`; reuse the pilot Source
+  localization. **G5**: `sanitizeLine` currently duplicated (notify + pilot) —
+  extract a shared helper this round (≥3 cards amplify the duplication).
+- **Field-cap single source**: **G9** — finalizer/mapping-layer rune caps are
+  double-written against each card's schema `maxLength`; unify this round
+  (schema is the source of truth, truncate before render).
+- **trust-boundary**: caller-supplied business fields on summary/docs cards go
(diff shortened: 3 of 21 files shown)


lml2468 (dismissed):
# PR Review — PR #649 (Mininglamp-OSS/octo-server)

Reviewed at head `c00af209` against base `main`.
**Build/vet/test:** `go build` exit 0 ✅, `go vet` ✅. `go test ./pkg/cardtmpl/...` → **all PASS** (base + access_request + **commented + shared**). DB-gated `modules/notify` integration (incl. the byte-equivalence baseline) → CI.
Classification: migrate two docs display cards (`docs.commented`, `docs.shared`) onto the cardtmpl Registry — wire-contract + card escaping/trust-boundary. Reviewed in **SECURITY** emphasis. Follows the #632/#633/#641/#647 pattern I reviewed.

## 1. Spec compliance — ✅
Graded against `.octospec/tasks/cardtmpl-l2a-migration/brief.md`.
- **Missing:** none. Two new template subpackages (handoff manifest+schema+sample + 3-method Template), notify wiring (`buildDocsDisplayCardViaRegistry` + `preflightDocsDisplaySchema`), C1 preflight + F7 fail-close routing, and the byte-equivalence baseline test — all present. Reaches the L2b hard-gate milestone (3 L2a cards across both wire profiles).
- **Over-build:** none. `NotifyReq`/`DocsCardFields` shape is **byte-untouched** (`model.go` 0 diff — plan B); the migration only reroutes the commented/shared branches.
- **Deviation:** none. Variant strings + attribution copy preserved to match legacy exactly.

## 2. Code quality — Approved

### Findings
No P0/P1/P2 findings.

- **Byte-identical to legacy — proven by a canonical-JSON baseline test.** `TestBuildDocsDisplayCards_MigrationBaseline` renders both the legacy `buildDocsCard` and the new Registry path for commented/shared, canonicalizes, and asserts `bytes.Equal`. This is the strongest migration-safety anchor: the visible card and wire bytes are identical, so no regression for existing clients. (DB-gated → CI.)
- **Escaping preserved.** The templates compose `BuildDocsResourceCardBodyWithLang` (extracted in resource.go), which keeps the existing `escapeMarkdown`/`truncateRunes` defenses — the resource-card render layer's own escaping is unchanged. Caller text (title/actor/excerpt) is bounded to the shared caps (200/120/300/80, matching the schema maxLengths).
- **No user-controlled URL field → the display preflight correctly omits the AbsoluteHTTPSURL check.** The commented/shared schemas expose only `docId`/`title`/`actorName`/`excerpt`/`updatedAt` (all bounded strings, `additionalProperties:false`) — the only URL is the **server-composed** deep link (`env.WebLoginURL` + `/d/{docId}`). So there's no caller URL that could hit the "schema-regex passes → Build requireHTTPS fails → render_error degrade" seam that the access-request preflight guards (that card has a user avatar URL; these don't). Not a gap — the asymmetry is correct.
- **C1 + F7 wiring matches the pilot.** `preflightDocsDisplaySchema` runs InputSchema validation before any gate → `ErrFieldsInvalid` (400 zero-delivery); Registry/Template unavailable → typed `errCardTmplUnavailable` (500 no-degrade, composition bug visible); other render errors degrade to text. Same 3-way classification I verified on #632.
- **`Registry.Render` remains the sole envelope/metadata boundary**; register-time sample self-check applies to the two new templates (fail-close on a bad schema/sample at composition).
- **`additionalProperties:false`** on both schemas rejects unknown fields — the mapping layer can't smuggle extra keys.

## 3. Test assessment
Per-template tests (commented/shared) pass here; the cross-cutting `TestBuildDocsDisplayCards_MigrationBaseline` (legacy↔Registry byte-equivalence) is the key regression guard but is DB-gated (needs the notify test server) → CI must run it. `testmain_test.go` is added for the notify package test setup. Good coverage for a migration.

## 4. Coverage / blind spots
- Verified by build/vet + the pure cardtmpl template suites + full read of the templates, the shared body builder's retained escaping, the display preflight (and confirming no user URL field exists), and the model-untouched check.
- NOT run: the DB-gated notify baseline + integration. Recommend the CI DB run (the byte-equivalence assertion is the migration's safety net).

## Verdict
**APPROVE** — a clean, low-risk L2a migration that follows the proven pilot pattern: `docs.commented`/`docs.shared` now render through the Registry with byte-identical output to the legacy path (canonical-JSON baseline test), escaping preserved via the shared resource-body builder, C1/F7 wiring matching #632, and no new URL attack surface (server-composed deep link only, so the omitted AbsoluteHTTPSURL preflight is correct not a gap). `NotifyReq` shape untouched. Both gates pass.
Suggested decision: `Safe to merge` once CI runs the DB-gated notify suite — especially `TestBuildDocsDisplayCards_MigrationBaseline`, the legacy↔Registry byte-equivalence guard. (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (dismissed):
Roadmap-C PR-1: migrates `docs.commented` + `docs.shared` display cards onto the cardtmpl Registry, symmetric to the shipped `docs.access-request` pilot. The migration is clean and the safety-critical surfaces all hold up.

## ✅ Verified

- **Markdown escaping (injection surface)** — Both new templates render every externally-sourced field through the shared `pkg/cardtmpl/resource.go` `assembleResourceCardBody`: title, attribution, excerpt, and every FactSet title/value are `escapeMarkdown`'d, and the excerpt is `truncateRunes`-bounded to `MaxExcerptRunes` (300). There is no raw un-escaped `TextRun`, and the only URL surface is the server-built deep-link (`docsDeepLink` with `url.PathEscape`) inside `Action.OpenUrl`. `FallbackText` runs every field through `cardtmpl.SanitizeLine`. No markdown-injection gap.
- **render_profile fail-closed** — `renderProfileCompatibility: "octo-chat/v1"` is validated at `Register` time by `IsAcceptedRenderProfile` (accepts only `""` or `octo-chat/v1`; panics otherwise), and `renderProfile` is required when compatibility is set. The value flows end-to-end: `RenderCardWithProfiles` → `deliverDocsCardNotification` → `docsSender.Send(carddispatch.Card{RenderProfile: renderProfile, ...})`, identical to the pilot path. No partial-flow gap.
- **Contract** — Both cards are notification-only (server-authored `Action.OpenUrl` + optional `Action.CopyToClipboard`; no `Action.Submit` / interactive inputs). No action-ID or submit-data contract to preserve.
- **Behavior parity** — `card_via_registry_display_baseline_test.go` runs 4 fixtures (commented/shared × named/anon) through both `buildDocsCard` (legacy) and `buildDocsDisplayCardViaRegistry`, strips the two newly-injected metadata fields, canonicalizes, and asserts byte-equality. Body output is byte-identical for in-bounds inputs.
- **Governance** — New `@0.1.0` packages; both registered + `SetDefault` before a single `Freeze()`. No in-place mutation of an already-shipped version. Legacy `buildDocsCard` remains only as the fallback for other kinds and as the baseline reference. Schema field caps are the single source of truth (`excerpt maxLength:300` == `MaxExcerptRunes`; `title:200` == `maxTitleRunes`). i18n unchanged (external shape frozen; CI i18n lanes green).

## 🟡 Non-blocking

- **Over-cap inputs shift from truncate/degrade to 400 zero-delivery.** For inputs exceeding the schema caps (excerpt > 300, title > 200, actorName > 120, updatedAt > 80), the C1 preflight in `modules/notify/card.go` (`preflightDocsDisplaySchema`) now returns a 400 with zero delivery, whereas the legacy path either silently truncated the excerpt at 300 or degraded to plain text. This is the **established, already-shipped C1 contract** (the access-request pilot applies the same `requestReason maxLength:300` cap), so it is a deliberate contract tightening rather than a new defect — and `action_finalizer.go` already truncates its own reason to `MaxExcerptRunes` before send. Still worth a conscious sign-off: confirm no live docs-backend caller legitimately emits actorName in `(120, 500]` or excerpts in `(300, …]` that were previously being delivered (truncated). If that traffic exists, consider clamping in `mapDocsCardFieldsToDisplayJSON` before schema validation instead of rejecting.
- **Fallback delegation asymmetry.** `buildDocsFallbackText` delegates to the Registry template only for `access_requested`; commented/shared still use the legacy fallback assembly. Both now share `SanitizeLine`, so drift risk is low, but the plain-text path is not yet routed through the new templates' `FallbackText`.

## 🔵 Suggestion

- In tests / `errorsIsFieldsInvalid`, prefer `errors.Is(err, cardtmpl.ErrFieldsInvalid)` over string matching to assert against the real error contract.

## Gate

CI in progress at review time (Build/Test/Vet/Lint running, none red); `mergeable`. No conflicts. The above lint/i18n lanes already pass.

Escaping, render_profile fail-close, contract preservation, byte-equivalence baseline, and freeze governance all hold. The single flagged behavior shift is consistent with the already-shipped Registry contract, so treating it as non-blocking. Approving.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #649 (octo-server)

Reviewed at head `c00af209`. Scope: migrate `docs.commented` + `docs.shared` notification cards to the Registry rendering pipeline (roadmap C PR-1).

## Summary

This PR migrates two display-only notification card types — `docs.commented` and `docs.shared` — from the legacy `buildDocsCard` inline builder to the Registry-backed `Template.Build` pattern established by the earlier `docs.access_requested` pilot. The architecture is clean: each new template package (`pkg/cardtmpl/docs_commented`, `pkg/cardtmpl/docs_shared`) owns its handoff contract (schema + manifest + sample), shares `assembleResourceCardBody` with the legacy path for byte-equivalence, and plugs into the same preflight + fail-close error classification used by the pilot. The `SanitizeLine` helper is promoted to `pkg/cardtmpl` as a single source of truth, eliminating duplicated control-character sanitization across three call sites. Byte-equivalence is proven by `TestBuildDocsDisplayCards_MigrationBaseline` which diffs legacy vs Registry output (stripped of `metadata.octo.{protocol,template}`).

## Verification

- ✅ **Byte-equivalence baseline** — `TestBuildDocsDisplayCards_MigrationBaseline` asserts legacy `buildDocsCard` output matches Registry output (minus octo metadata injection) for 4 fixtures covering both card types, anonymous actor, and missing optional fields.
- ✅ **Schema preflight parity** — `preflightDocsDisplaySchema` mirrors the pilot's `preflightDocsAccessRequestSchema` pattern: registry-nil → `errCardTmplUnavailable` (500); schema violation → `ErrFieldsInvalid` (400 zero delivery). Verified at `modules/notify/card.go:318-339`.
- ✅ **Label parity** — New `labelsFor()` in both `docs_commented/template.go` and `docs_shared/template.go` produce byte-identical zh/en strings to legacy `docsLabelsFor` (banner, bannerAnon, title, actor, updatedAt, kvSep, sourceLabel).
- ✅ **Registry wiring** — `main.go:installCardTmplRegistry` registers both templates + sets defaults before `Freeze()`. `modules/notify/testmain_test.go` mirrors the same registration for test isolation.
- ✅ **Deep-link integrity** — Both templates call `BuildDocsResourceCardBodyWithLang` which validates `docID`/`spaceID` non-empty and routes through `docsDeepLink` → `requireHTTPS`.
- ✅ **SanitizeLine consolidation** — `cardtmpl.SanitizeLine` is a byte-identical extraction of the previous `sanitizeLine` implementations. Both `modules/notify/card.go` and `docs_access_request/template.go` now delegate to it.
- ✅ **Wire profile correctness** — Display cards use `ProfileV1` (`octo/v1`, non-interactive) matching manifest `wireProfile`. Only access_requested uses `ProfileV2` (`octo/v2`, interactive).

Static analysis only at head `c00af209248f5634bdeaf4c1f7e37aa3f952a1ba`; build and tests not executed in this environment.

## Findings

### P2 — Fallback text delegation gap for commented/shared (`modules/notify/card.go:635`)

`buildDocsFallbackText` delegates to `templateFallbackText` (F6 pattern) only for `DocsCardKindAccessRequested && docsApprovalCardsEnabled()`. The newly migrated `DocsCardKindCommented` and `DocsCardKindShared` kinds now go through Registry for card building, but their fallback text path (used when `canCard=false` — render error, cardmsg disabled, or sender missing) still uses the legacy `docsAttributionAndVariant` + multi-line composition instead of `Template.FallbackText`.

This is not a correctness bug today — the legacy composition produces byte-identical output to the new `Template.FallbackText` implementations (verified by label parity and `TestFallbackTextZH`/`TestFallbackTextAnon`). However, it creates a maintenance risk: if a future change updates `Template.FallbackText` wording (e.g. the planned G4 i18n locale collapse) without also updating the legacy `docsAttributionAndVariant` path, the card body and plain-text fallback will silently diverge.

Worth extending the F6 delegation in `buildDocsFallbackText` to handle commented/shared kinds, mirroring the access_requested pattern — a small `displayTemplateFallbackText` helper that maps `DocsCardFields` → display schema JSON and calls `tmpl.FallbackText`.

- **`assembleResourceCardBody` extraction**: The refactoring from `buildResourceCard` is a pure extraction — the body assembly logic (title validation, fact count, excerpt truncation, icon HTTPS check, source bounds, action set) is unchanged. Both `buildResourceCard` (legacy full document) and the new `BuildDocsResourceCardBodyWithLang`/`BuildSummaryResourceCardBodyWithLang` (fragment-only) call it.
- **`additionalProperties: false` in schemas**: `mapDocsCardFieldsToDisplayJSON` produces exactly `{docId, title, actorName, excerpt, updatedAt}` — no extra fields that would trip the strict schema.
- **Error classification in `deliverDocsCardNotification`**: The new commented/shared branches follow the same `ErrFieldsInvalid` → 400 / `errCardTmplUnavailable` → 500 / other → degrade-to-text policy as access_requested.
- **`context.Background()` usage**: Consistent with the existing access_requested branch; the request context is not needed for card building.
- **Manifest `renderProfile: octo-chat@1.2.0-rc.1`**: Matches the pilot manifest, consistent with `ProfileV1` wire profile.

## Data-flow trace

- `deliverDocsCardNotification` → routes `DocsCardKindCommented`/`DocsCardKindShared` to `buildDocsDisplayCardViaRegistry` → `mapDocsCardFieldsToDisplayJSON` (TrimSpace all fields) → `registry.RenderCardWithProfiles` → `Template.Build` → `BuildDocsResourceCardBodyWithLang` → `assembleResourceCardBody` (shared with legacy).
- Preflight: `preflightDocsDisplaySchema` runs before `memberCache`/`docsSender`/`cardmsg.Enabled` gates, so schema violations cannot be masked by empty-member 200 responses or text degradation.
- Fallback: `buildDocsFallbackText` → legacy `docsAttributionAndVariant` + multi-line composition (not `Template.FallbackText` — see P2 above).

## Blind-spot checklist

- **C1 — Dual path parity**: ✅ Legacy `buildDocsCard` and new Registry path share `assembleResourceCardBody`. Preflight for commented/shared mirrors access_requested pattern. SanitizeLine centralized.
- **C2 — Control-flow ordering**: ✅ `assembleResourceCardBody` extraction does not double-wrap or change ordering. Both callers (legacy full-doc and new fragment-only) produce the same body.
- **C3 — Authorization boundary**: N/A — no authorization changes.
- **C4 — Authorization lifecycle**: N/A.
- **C5 — Build/runtime correctness**: ✅ `embed.FS` directives compile-time embed handoff assets. Test fixtures reference embedded samples.
- **C6 — Governance/docs self-consistency**: ✅ `.octospec/journal` and `.octospec/log.md` accurately describe the migration scope and contract.

## Cross-round blocker review

## Verdict: APPROVED

Clean, well-structured migration with byte-equivalence proven by baseline tests, proper preflight schema validation, and correct error classification. The single P2 (fallback text delegation gap) is a non-blocking maintainability item.

[Octo-Q] verdict: APPROVE — no P0/P1 found; one P2 maintainability note on fallback delegation.

yujiawei (dismissed):
# Code Review — PR #649 (octo-server)

**Head SHA:** `c00af209248f5634bdeaf4c1f7e37aa3f952a1ba` · **Base:** `main` (merge-base `04b8c793`)
**Scope:** migrate `docs.commented` + `docs.shared` (v1 display cards) onto the cardtmpl Registry (roadmap C PR-1). Security-sensitive (migration / trust boundary for caller-controlled text).

Reviewed the full diff (21 files, +1431/-34), built and vetted the tree, ran the new card-package tests, and cross-checked the notify wiring, registration root, schemas/manifests, and the byte-equivalence baseline. `go build ./...`, `go vet ./pkg/cardtmpl/...`, and `go test ./pkg/cardtmpl/docs_commented/... ./pkg/cardtmpl/docs_shared/...` all pass locally. Registry wiring is consistent (both new kinds are registered + defaulted at the composition root and routed in `deliverDocsCardNotification`; no routed-but-unregistered kind). Both manifests correctly declare `renderProfile` + `renderProfileCompatibility`. `SanitizeLine` is a faithful extraction of the two prior copies with no behavior change. This is careful, well-documented work.

One issue is blocking; the rest are non-blocking.

---

## Spec compliance

The migration shape matches the brief: self-embedded `handoff/<id>@<ver>/`, 3-method Templates, `SetDefault` + `Freeze`, shared `BuildDocsResourceCardBodyWithLang` body so the card node is byte-identical modulo the allowed `metadata.octo.{protocol,template}` + `render_profile` deltas. The C1 preflight is correctly hoisted ahead of member/gate checks.

**One deviation (see P1 below):** for the two *migrated* cards, an over-length `excerpt` / `actorName` / `updatedAt` now returns **HTTP 400 zero-delivery** where the legacy builder **truncated and delivered**. This changes the accepted-input behavior of a load-bearing path, is not exercised by any test, and runs against the brief's own G9 wording ("schema is the source of truth, **truncate before render**"). The byte-equivalence baseline cannot catch it because every fixture is in-bounds.

## Findings

### 🔴 P1 — Over-length fields now hard-reject (400, zero delivery) where legacy truncated & delivered

The new schemas cap `excerpt` at 300, `actorName` at 120, `updatedAt` at 80 (`pkg/cardtmpl/docs_commented/handoff/docs.commented@0.1.0/contract/data.schema.json:20-35`, same for `docs.shared`). `preflightDocsDisplaySchema` runs `InputSchema.Validate` before member/gate checks and, on a length violation, returns `ErrFieldsInvalid` → `deliverDocsCardNotification` returns `errNotifyCardInvalid` → **400, no delivery to anyone** (`modules/notify/card.go:320-339`).

The legacy path never rejected on length — it truncated at render time:

```go
// pkg/cardtmpl/resource.go:281
if excerpt := truncateRunes(strings.TrimSpace(resource.Excerpt), MaxExcerptRunes); excerpt != "" {
```
and facts were bounded only by `maxFactRunes = 500` (`pkg/cardtmpl/resource.go:213`).

**Failure scenario:** a docs-backend comment/share note of 301+ runes (a long comment is realistic — 300 CJK chars is a modest paragraph). Before: recipients got a card with a 299-rune prefix + "…". After: the request 400s and *every* recipient loses the notification — not even the text fallback fires, since the reject happens at preflight. No producer-side cap on `excerpt` is evident in this repo (`DocsCardFields.Excerpt` is documented as free-form).

This is the exact "unchanged wire shape ≠ unchanged accepted-input behavior" trap. The fix aligns with the brief's G9 intent — truncate to the schema cap in the mapping layer before preflight, so behavior stays equivalent to legacy:

```go
// modules/notify/card_via_registry.go — mapDocsCardFieldsToDisplayJSON
"excerpt": truncateToRunes(strings.TrimSpace(card.Excerpt), 300), // = MaxExcerptRunes
```
(and similarly cap the fact fields), then add a >cap regression case to the baseline test. Alternatively, if the docs-backend contractually guarantees these fields are ≤ the caps before sending, please state that guarantee and add a test asserting the boundary — otherwise this is a silent notification-loss regression on the migrated cards.

### 🟡 P2 — "byte-equivalence" baseline is semantic-JSON equivalence, not byte equivalence

`TestBuildDocsDisplayCards_MigrationBaseline` unmarshals both docs, strips the allowed fields, then re-marshals through `canonicalJSON` (`modules/notify/card_via_registry_display_baseline_test.go:82-109`, helper at `card_via_registry_baseline_test.go:113`). The unmarshal→remarshal round-trip normalizes key ordering, whitespace, and escape forms (e.g. `<` vs `<`), so the test proves *structural/semantic* equality, not literal wire-byte equality. That's the right guard for catching copy/ordering/alias drift (its stated purpose), but the "byte-equivalence" naming over-claims: a genuinely byte-differing yet semantically-equal output would still pass. Consider renaming to reflect canonical-JSON equivalence, or asserting on the raw bytes after removing *only* the permitted fields at the byte level.

### 🟡 P2 — Dangling reference to a non-existent "docs_commented G9 field-cap test"

`pkg/cardtmpl/resource.go:203-205` states the excerpt cap and schema `maxLength` "never drift (see the docs_commented G9 field-cap test)". No such test exists (`grep` across `pkg/cardtmpl/docs_commented`, `docs_shared`, `modules/notify` finds none). The brief's G9 acceptance ("test asserts no double-write drift") is therefore not actually asserted for the display cards. Either add the referenced test (assert `data.schema.json` excerpt `maxLength == cardtmpl.MaxExcerptRunes`) or drop the reference. This ties directly into P1 — such a test is where the truncate-vs-reject decision should be pinned.

### ⚪ Nit — gofmt drift in `pkg/cardtmpl/docs_shared/template.go`

`gofmt -l` flags this file (const block / `func New()` alignment). It is **not** a CI gate today (the repo's `.golangci.yml` runs `default: none` + `govet` only), so this does not block, but a `gofmt -w` keeps the new file clean ahead of the planned lint re-enable.

### Considered and dismissed

- *"Render errors fail open to plain text, contradicting F7 fail-close."* For these v1 display cards this is intended, documented F6/§10 behavior and matches the legacy builder exactly (a legacy build error also degrades to text). The author correctly splits the two typed errors: `errCardTmplUnavailable` → 500 no-degrade (the real F7 fail-close), other render errors → text degrade so the notification still lands (`modules/notify/card.go:411-423`). Not a regression.
- *"Actor hydration runs after preflight."* `hydrateActorName` mutating an already-validated field is a pre-existing pattern inherited from the merged access-request pilot (same `requester.name` maxLength 120). Its only real residual — a hydrated name exceeding 120 runes — collapses into the P1 reject regression (it would 400 rather than deliver), so it's covered by the P1 fix.

## Advisory-leg coverage (transparency)

Two independent advisory legs were run. **Only one produced usable findings:** the cross-family long-context leg returned a truncated response (`finish=length`) with no concrete findings and is treated as **absent**, not as tacit approval. The adversarial line-level leg completed and surfaced the input-domain regression that became P1 here (confirmed independently against the diff), plus the test-naming observation (P2). Coverage the automated legs could not verify — flagged for human attention on this security-sensitive PR:

- Full `modules/notify` + `pkg/cardtmpl` suites against real MySQL/Redis/WuKongIM (not runnable in this environment).
- The over-cap boundary (P1) — no test exercises it today.
- Confirmation of whether the docs-backend guarantees field lengths ≤ the new schema caps.

## Verdict

The migration is structurally sound and the wiring is correct, but the length-reject regression (P1) is a real, untested behavior change on the notify path that can silently drop notifications for realistic input. Recommend addressing P1 (truncate-before-preflight, matching G9's "truncate before render" and legacy behavior) plus a boundary test before merge.

lml2468 (approved):
# PR Review (re-review) — PR #649 (Mininglamp-OSS/octo-server)

Re-reviewed at **new head `e6b7454a`** against base `main`. Re-affirms my APPROVE at `c00af209`.
**Build/vet/test:** `go build ./modules/notify/ ./pkg/cardtmpl/...` exit 0 ✅, `go vet` ✅. `go test ./pkg/cardtmpl/ ./pkg/cardtmpl/docs_commented/ ./pkg/cardtmpl/docs_shared/` → **PASS**. DB-gated notify baseline/integration → CI.

## What changed since `c00af209` (delta reviewed)
One commit — `e6b7454a` "address PR #649 review — preserve delivery on over-length …" — applying the same truncate-not-400 fix #650 landed for summary cards, now to the docs-display mapping. This closes yujiawei's #649 🔴 (and the cross-PR consistency note I raised on #650).

### Over-length docs-display fields now truncate-and-deliver ✅
`mapDocsCardFieldsToDisplayJSON` now truncates `title` (→`MaxTitleRunes` 200), `actorName` (→120), `excerpt` (→`MaxExcerptRunes` 300), `updatedAt` (→80) via the shared `cardtmpl.TruncateRunes` — so an over-length display field renders truncated instead of flipping to a C1 400, restoring parity with legacy `buildDocsCard` (which never length-rejected). Verified:
- **`docId` deliberately NOT truncated** — deep-link key; over-long still hits **C1 400** (silent truncation would forge a wrong link). Correct, symmetric with #650's `taskNo`.
- **Structural violations still 400** (missing/empty required, unknown fields via `additionalProperties:false`, over-long docId).
- **Caps locked to schema (G9):** reuses the exported `MaxTitleRunes`/`MaxExcerptRunes` + local `docsActorNameMaxRunes`/`docsUpdatedAtMaxRunes` mirroring the schema maxLength; `TestMapDocsDisplayFields_TruncatesDisplayFields` fails if any cap exceeds schema (so truncation can't post-truncate 400).
- The `docs_shared/template.go` change is **gofmt whitespace only** (const-block alignment) — no logic change.

### Byte-equivalence + escaping intact
Escaping is unchanged (shared `BuildDocsResourceCardBodyWithLang` still escapeMarkdown/truncateRunes); the display baseline test is extended for the over-length path; pure suites pass. New `TestMapDocsDisplayFields_TruncatesDisplayFields` covers truncate-and-deliver + docId-still-400.

## Verdict
**APPROVE** (re-affirmed at `e6b7454a`) — the over-length compat regression yujiawei flagged is fixed symmetrically with #650: docs-display cosmetic fields truncate-and-deliver (legacy parity, never-lost invariant), the deep-link key `docId` and genuine structural violations keep C1 400, caps are schema-locked with a dedicated test, and escaping/byte-equivalence are unchanged. The #650 cross-PR note is now closed — both slices behave identically. Both gates pass.
Suggested decision: `Safe to merge` once CI runs the DB-gated notify suite (esp. the display byte-equivalence baseline). This is the base of the #650 stack — land it first. (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (approved):
**Re-review @ e6b7454a** (previously APPROVED @ c00af209). ✅ **APPROVE** — the new commit correctly resolves the over-length compat break.

## Context
On the prior head I flagged over-length docs-display fields → C1 400 zero-delivery as a 🟡 (matching the shipped C1 pilot contract). A follow-up review escalated it to a 🔴 backward-compat break, and the team resolved it by mirroring the summary-side truncation fix (PR #650) into the docs-display mapping. This commit is that mirror.

## True delta (c00af209 → e6b7454a)
Focused, no drift. `mapDocsCardFieldsToDisplayJSON` previously only `TrimSpace`'d display fields; it now truncates decorative display strings to their schema `maxLength` via the shared `cardtmpl.TruncateRunes`, so an over-length value renders truncated and still delivers — restoring the legacy "notification never lost" invariant. Migration code otherwise unchanged.

## 🔴 Blocking
None.

## ✅ Verified
- **Decorative fields truncate + deliver**: `title`→`MaxTitleRunes` (200), `actorName`→`docsActorNameMaxRunes` (120), `excerpt`→`MaxExcerptRunes` (300), `updatedAt`→`docsUpdatedAtMaxRunes` (80). Over-length input now passes preflight + builds successfully instead of flipping to 400.
- **Structural key stays hard-400**: `docId` (deep-link key) is NOT truncated — over-length `docId` still returns C1 400. Correct: silently truncating a link key would produce a wrong link. The decorative-vs-structural distinction matches the summary-side `taskNo` handling.
- **Single source of truth / no post-truncate 400**: truncation cap == schema `maxLength` == render-guard cap for every decorative field. `MaxTitleRunes` is now exported and `maxTitleRunes = MaxTitleRunes`; `TruncateRunes` wraps the internal `truncateRunes` so ingress and render share one impl. Since truncation cap equals schema max, the field is always within schema after truncation → no post-truncate rejection.
- **Cap-drift guard added**: `TestSchemaCapsMatchRenderCaps` (both `docs_commented` and `docs_shared`) asserts schema `maxLength` == `cardtmpl.MaxExcerptRunes`/`MaxTitleRunes`. This also fills the previously-dangling G9 field-cap test referenced in `resource.go`.
- **Escaping intact**: truncation happens at the ingress/mapping layer; the render path (`assembleResourceCardBody`) still `escapeMarkdown`s title/attribution/excerpt/facts and `truncateRunes` excerpt at render time. No escaping bypass.
- **Baseline holds**: `TestBuildDocsDisplayCards_MigrationBaseline` still passes — legacy `buildDocsCard` vs Registry render remain equivalent after canonical-JSON normalization. New `TestMapDocsDisplayFields_TruncatesDisplayFields` asserts over-length → truncate-and-deliver and over-length structural key → still 400.
- render_profile / C1 preflight fail-close and the schema→400 / unregistered→500 / other→degrade three-way classification remain intact; no new user-controlled URL field beyond the server-composed deep link.
- Targeted tests green locally: `pkg/cardtmpl/{docs_commented,docs_shared,.}` and the notify baseline + truncate tests.

## 💬 Non-blocking notes
- `docsActorNameMaxRunes` (120) and `docsUpdatedAtMaxRunes` (80) are local constants asserted only functionally (via the truncate-and-deliver test), not by a direct schema-parity assertion like title/excerpt. Consider extending `TestSchemaCapsMatchRenderCaps` to also assert `actorName`/`updatedAt` parity so a future schema bump can't drift silently.
- The baseline test was accurately renamed from "byte-equal" to canonical-JSON equivalence — good correction of the earlier over-claim.

Good, targeted fix that restores the delivery invariant without regressing structural-key strictness, escaping, or the migration baseline.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #649 (octo-server)

## Summary

This PR migrates the `docs.commented` and `docs.shared` notification cards to the Registry-backed rendering path (roadmap C PR-1). Two new L2a Template packages (`pkg/cardtmpl/docs_commented`, `pkg/cardtmpl/docs_shared`) are added with handoff contracts, embed assets, and per-template `Build` / `FallbackText` implementations. The production dispatch path in `modules/notify/card.go` routes commented/shared kinds through `buildDocsDisplayCardViaRegistry` (Registry.Render) instead of the legacy `buildDocsCard`, with C1 preflight schema validation added before member resolution. The body assembly is refactored to share `assembleResourceCardBody` between the legacy and Registry paths, preserving byte-equivalent card output. A new `TestMain` in `modules/notify` wires the test-package registry to mirror `main.installCardTmplRegistry()`. Comprehensive tests cover canonical-JSON migration baseline, schema-cap alignment (G9), over-length field truncation, C1 rejection, and metadata injection.

## Verification

Static analysis only at head `e6b7454a`; build and tests not executed in this environment.

- ✅ **Body equivalence** — `assembleResourceCardBody` is the single shared source of truth (`pkg/cardtmpl/resource.go:211`). Both `buildResourceCard` (legacy, marshals full document) and `BuildDocsResourceCardBodyWithLang` (Registry, fragment-only) call it. The `TestBuildDocsDisplayCards_MigrationBaseline` test asserts canonical-JSON equivalence after stripping `metadata.octo.{protocol,template}` — the test name correctly says "canonical-JSON equivalence", not byte-equal.
- ✅ **Preflight before gates** — `preflightDocsDisplaySchema` runs at `modules/notify/card.go:318-339`, before member resolution, sender check, and `canCard` gate. Schema violations return `ErrFieldsInvalid` (400, zero delivery). Registry unavailable returns `errCardTmplUnavailable` (500, no degradation). This matches the access-requested preflight pattern.
- ✅ **Over-length field preservation** — `mapDocsCardFieldsToDisplayJSON` (`modules/notify/card_via_registry.go:183-200`) truncates display fields (title→200, actorName→120, excerpt→300, updatedAt→80) to schema maxLength before preflight, preserving delivery. `docId` is never truncated — over-length docId stays C1 400 (deep-link key must not be silently truncated). `TestMapDocsDisplayFields_TruncatesDisplayFields` locks this.
- ✅ **Label parity** — `docs_commented.labelsFor` / `docs_shared.labelsFor` values match `docsLabelsFor` legacy counterparts field-by-field (banner, bannerAnon, title, actor, updatedAt, kvSep, sourceLabel) in both zh-CN and en locales.
- ✅ **SanitizeLine consolidation** — Both `modules/notify.sanitizeLine` and `docs_access_request.sanitizeLine` now delegate to `cardtmpl.SanitizeLine` (`pkg/cardtmpl/resource.go:382-389`). The `unicode` import is correctly removed from both consuming packages.
- ✅ **Registry wiring** — `main.go:717-722` registers both templates with `SetDefault` + `Freeze`, consistent with the existing access-requested registration. `testmain_test.go` mirrors this for the test package.
- ✅ **Error classification** — `errors.Is(buildErr, cardtmpl.ErrFieldsInvalid)` → 400; `errors.Is(buildErr, errCardTmplUnavailable)` → 500 no-degrade; other errors → degrade to text fallback. Same three-tier classification as the access-requested path.

## Findings

No P0/P1 issues; two P2 items and one nit below.

### P2 — Unexported helper shipped as public API without tests (`pkg/cardtmpl/resource.go:99`)

`BuildSummaryResourceCardBodyWithLang` is exported and documented as preparation for roadmap C PR-2 (summary.completed / summary.failed), but has zero callers and zero test coverage in this PR. It ships as dead public API surface. Unlike `BuildDocsResourceCardBodyWithLang` — which is exercised by the `docs_commented` / `docs_shared` template tests — the summary counterpart has no consumer to validate its contract. Consider either adding a round-trip test or keeping it unexported (`buildSummaryResourceCardBodyWithLang`) until the consumer PR lands, so the API surface doesn't carry untested commitments.

### P2 — Fallback-text path not delegated for migrated kinds (`modules/notify/card.go:390`)

`buildDocsFallbackText` delegates to `Template.FallbackText` only for `DocsCardKindAccessRequested` (F6 pattern). For the newly-migrated `commented` and `shared` kinds, when `canCard` is false (card feature disabled or sender unavailable), the fallback text still goes through the legacy inline builder at `:641`. `Template.FallbackText` — which this PR implements, tests, and exports — is dead code for these two kinds in the production fallback path. The legacy and new paths produce equivalent output today (same labels, same SanitizeLine, same field ordering), so this is not a user-visible defect. But the asymmetry (card rendering via Registry, fallback text via legacy) creates a drift surface as the L2a templates evolve: a future label change in the template won't propagate to the fallback path.

### Nit — Schema-cap lock incomplete for display-field caps (`pkg/cardtmpl/docs_commented/template_test.go:148`)

`TestSchemaCapsMatchRenderCaps` asserts `excerpt` and `title` maxLength against `cardtmpl.MaxExcerptRunes` / `cardtmpl.MaxTitleRunes`, but the two display-field caps without exported constants (`docsActorNameMaxRunes=120`, `docsUpdatedAtMaxRunes=80` at `modules/notify/card_via_registry.go:177-178`) are not asserted against the schema. If a future edit changes `actorName.maxLength` in `data.schema.json` without updating the ingress cap (or vice versa), truncation and schema validation silently diverge — one direction produces over-truncated text, the other produces preflight 400. Consider extracting these caps to an exported location and adding schema-lock assertions in both `docs_commented` and `docs_shared` test files.

- `buildResourceCard` refactor: `lang := i18n.OutboundLanguage(ctx)` moved from body assembly to the caller wrapper; the new `assembleResourceCardBody` takes `lang` as parameter. No context leakage or double-resolution.
- `additionalProperties: false` in both `data.schema.json` files — the five-field set (`docId`, `title`, `actorName`, `excerpt`, `updatedAt`) matches `mapDocsCardFieldsToDisplayJSON` output exactly. No risk of extra-field rejection.
- `docsSafeExcerpt` returns `strings.TrimSpace(card.Excerpt)` for commented/shared kinds — same as `strings.TrimSpace(f.Excerpt)` in the new template. No excerpt data-flow divergence.
- `TestMain` does not break existing tests: tests that need nil registry (`card_via_registry_test.go:22`, `card_c1_regression_test.go:20`) properly save/restore with `prev := cardtmpl.DefaultRegistry(); defer SetDefaultRegistry(prev)`.
- `docsApprovalCardsEnabled()` gate remains scoped to access-requested only. Commented/shared kinds are unconditionally routed through Registry regardless of the gate — intentional per the PR's design.
- C1 parity (blind-spot C1): both commented and shared templates are symmetrically registered in `main.go` and `testmain_test.go`; both receive preflight at the same code location.
- No authorization or visibility boundary changes in this PR (blind-spot C3/C4: N/A).

## Verdict: COMMENT

Well-structured migration with thorough test coverage. The canonical-JSON baseline test, schema-cap alignment assertions, and over-length truncation preservation are strong signals. The two P2 items are non-blocking architectural hygiene — the dead public API and the fallback-text delegation gap — neither producing user-visible defects today. The nit about incomplete schema-cap locking is a forward-looking resilience suggestion.
---

## Data-Flow Trace (per Octo-Q mandate)

Each consumed data field traced to its upstream source:

| Consumer | Field | Upstream Source | Runtime Reachability |
|---|---|---|---|
| `Template.Build` (`docs_commented/template.go:78`) | `fieldsRaw` | `mapDocsCardFieldsToDisplayJSON` → `json.Marshal` of `DocsCardFields` struct fields | ✅ Confirmed: `deliverDocsCardNotification` → `buildDocsDisplayCardViaRegistry` → `mapDocsCardFieldsToDisplayJSON` |
| `Template.Build` (`docs_commented/template.go:97`) | `env.Lang` | `i18n.OutboundLanguage` resolved by caller, passed via `BuildEnv.Lang` | ✅ Confirmed: `card_via_registry.go:225` sets `Lang: lang` from function param |
| `Template.Build` (`docs_commented/template.go:97`) | `env.WebLoginURL` | `n.ctx.GetConfig().External.WebLoginURL` | ✅ Confirmed: `card_via_registry.go:224` |
| `Template.Build` (`docs_commented/template.go:97`) | `env.SpaceID` | caller `spaceID` param → `deliverDocsCardNotification` `req.SpaceID` | ✅ Confirmed: `card_via_registry.go:226` |
| `assembleResourceCardBody` (`resource.go:211`) | `resource.Title` | `f.Title` from JSON (pre-truncated to `MaxTitleRunes=200`) | ✅ Render guard at `:226` checks `utf8.RuneCountInString > maxTitleRunes` — truncation ensures pass |
| `assembleResourceCardBody` (`resource.go:211`) | `resource.Excerpt` | `strings.TrimSpace(f.Excerpt)` from JSON (pre-truncated to `MaxExcerptRunes=300`) | ✅ Render truncates again via `truncateRunes` at `:262` — double-safe |
| `preflightDocsDisplaySchema` (`card_via_registry.go:235`) | `fields` | `mapDocsCardFieldsToDisplayJSON` (same function as build path) | ✅ Preflight and build consume identical mapping — no divergence |
| `buildDocsFallbackText` (`card.go:641`) | `card.Excerpt` | `docsSafeExcerpt(card)` → `strings.TrimSpace(card.Excerpt)` for commented/shared | ✅ Confirmed equivalent to `strings.TrimSpace(f.Excerpt)` in Template.FallbackText |

No empty-array / undefined / early-return / gate-short-circuit traps found in the traced paths.

## Blind-Spot Checklist (C1–C6, per Octo-Q mandate)

- **C1 — Symmetric path parity**: ✅ Both `docs_commented` and `docs_shared` are symmetrically registered (`main.go:717-722`), receive preflight (`card.go:318-339`), and are dispatched through `buildDocsDisplayCardViaRegistry` (`card.go:385-397`). `TestMain` mirrors production wiring.
- **C2 — Control-flow ordering / nesting**: ✅ The preflight runs once before member resolution; the build runs once inside `canCard`. No double-invocation or ordering issue. The `SanitizeLine` consolidation is a pure delegation — same implementation, no new call sites.
- **C3 — Authorization boundary**: N/A — no endpoint, tool, or credential changes.
- **C4 — Authorization lifecycle**: N/A — no auth/status changes.
- **C5 — Build/note ≠ runtime correctness**: N/A — no build/packaging/deployment changes.
- **C6 — Governance/policy self-consistency**: N/A — no governance or policy documents changed.

[Octo-Q] verdict: APPROVE — no P0/P1 found; two P2 non-blocking items (dead public API without tests, fallback-text delegation gap) and one nit (incomplete schema-cap lock). PR is safe to land.

PR #645, opened 2026-07-23
docs(cardtmpl): align platform-card-base contract with implementation

an9xyz:
## Summary

Documentation-only PR fixing contract/code divergences in
`docs/platform-card-base.md` found in a post-#641 audit. The contract is
self-declared authoritative ("code wins; a mismatch is a doc bug"), so these
are bugs in the doc, not behavior changes. **No code touched.**

## Changes

- **§5/§6** — `Registry.Render` returns the full **type-17 InteractiveCard
  envelope** map `{type,card,card_version,profile}` (was documented as "AC
  top-level" / `json.RawMessage`). Documented `RenderCard`, which returns the
  bare `card` node JSON + wire profile — used by `CardUpdater.ReplaceView`,
  previously undocumented. Pipeline step 6 corrected to
  `cardmsg.Validate(payload)` (single arg, not `Validate(doc, profile)`).
- **§13** — `dmwork_cardtmpl_build_total` result enum adds `template_unknown` /
  `state_unknown`, described as runtime lookup / state-resolution misses (not
  registration errors).
- **§1 / §5 FallbackText** — separated the two downgrade mechanisms: a
  low-capability client (unknown profile) renders the server-authored `plain`
  (client-side negotiation, see `docs/card-protocol.md` §3), while
  `FallbackText` is the server-side degrade on render failure / gate off /
  master switch off.
- **§1** — "dual-mode" reworded to Go-only today (`${}` JSON engine is roadmap
  E1, not implemented).
- **§10** — client-profile downgrade row marked as client-side responsibility.

## Out of scope

- `docs-notify-contract.md` metadata sync (roadmap F1) — separate follow-up.
- §15 untouched.

## Testing

Docs-only; no code changed. `git diff --check` clean. `Render`/`RenderCard`
signatures and the render pipeline verified against `pkg/cardtmpl/render.go`.

## Checklist

- [x] PR description is in English
- [x] Docs-only, no behavior change

PR #650, opened 2026-07-23
feat(cardtmpl): migrate summary.completed + summary.failed to Registry (roadmap C PR-2)

an9xyz:
## Summary

Roadmap C second slice — migrate the two remaining legacy L2a summary-notify
display cards (`summary.completed`, `summary.failed`) onto the cardtmpl
Registry, reusing the PR-1 (#649) copy-directory shape and its
`BuildSummaryResourceCardBodyWithLang` scaffold.

**Stacked on #649** (this branch is `cardtmpl-l2a-summary-migration` off
`cardtmpl-l2a-migration`, cross-fork). Please **merge #649 first**; the diff
here will reduce to just the summary changes once #649 lands and this branch
rebases to `main`.

**Milestone reached**: every legacy display-card deliver branch is now
Registry-backed:

| Card | Wire | Migrated in |
| --- | --- | --- |
| `docs.access-request` | v2 interactive | pilot (#633) |
| `docs.commented` | v1 display | PR-1 (#649) |
| `docs.shared` | v1 display | PR-1 (#649) |
| `summary.completed` | v1 display | **PR-2 (this)** |
| `summary.failed` | v1 display | **PR-2 (this)** |

5 L2a cards Registry-backed — §2.2-5 L2b hard gate ② exceeded. Only
`generic.approval` remains legacy (dynamic-owner conflict with static
`TemplateActionContract`, tracked in the parent brief).

External `NotifyReq` / `SummaryCardFields` shape is unchanged (plan B).

## Related Issue / Prior art

Roadmap C follow-up to PR-1 (#649). No standalone issue.

## Linked Spec

`.octospec/tasks/cardtmpl-l2a-summary-migration/brief.md`

## Changes (net of PR-1)

- **`pkg/cardtmpl/summary_completed@0.1.0`** — new subpackage: self-embedded
  `handoff/` (manifest declaring `renderProfile` + `renderProfileCompatibility`
  per #647, `contract/data.schema.json`, `samples/shown.json`) + 3-method
  Template. Build composes `BuildSummaryResourceCardBodyWithLang` (added in
  PR-1) so bodies are byte-identical to legacy `buildSummaryCard` for the
  same inputs.
- **`pkg/cardtmpl/summary_failed@0.1.0`** — symmetric subpackage; differs in
  `Variant` / attribution / excerpt only. Schema adds optional `reason`
  (maxLength = `cardtmpl.MaxExcerptRunes` = 300, G9 single source).
- **notify wiring** (`modules/notify/card_via_registry.go`,
  `modules/notify/card.go`): new `buildSummaryCardViaRegistry` +
  `preflightSummarySchema` + `mapSummaryCardFieldsToJSON`;
  `deliverCardNotification` routes `SummaryCardKindCompleted` /
  `SummaryCardKindFailed` through Registry with F7 fail-close and C1
  preflight (schema errors → 400 zero-delivery ahead of member/gate checks).
  `renderProfile` now flows through the send path.
- **`main.installCardTmplRegistry`** + **`modules/notify` TestMain** register
  both new cards alongside the pilot + PR-1 cards.

## Testing

Locally green with MySQL 8 / Redis 7 / WuKongIM:

- `go build ./...`, `go vet ./...`
- `pkg/cardtmpl/summary_completed` -race -cover → **91.7%**
- `pkg/cardtmpl/summary_failed` -race -cover → **91.1%**
- `pkg/cardtmpl/...` full package tree — all green
- `modules/notify` full package -race — all green including the 4-case
  summary byte-equivalence baseline, C1 preflight assertion, F7 unwired
  assertion, and PR-1's docs baseline (no cross-migration regression)
- `make i18n-extract-check`, `make i18n-lint` — clean

**Byte-equivalence baseline is the core acceptance.**
`modules/notify/card_via_registry_summary_baseline_test.go` runs 4 fixtures
(`completed_zh_full`, `completed_zh_minimal`, `failed_zh_with_reason`,
`failed_zh_full`) through both legacy `buildSummaryCard` and
`buildSummaryCardViaRegistry`, strips the two newly-injected metadata fields
(`metadata.octo.protocol`, `metadata.octo.template`), and asserts canonical
JSON byte-equality. Any drift (copy / ordering / alias / extra field) fails
immediately.

Baseline stays zh-CN — legacy `buildSummaryCard` routes the AC button label
through `i18n.OutboundLanguage(ctx)` (background ctx → zh default) while
Registry uses `env.Lang` directly. English source-label localization is
covered by the Template unit tests (`TestRenderEnglishSourceLocalized`) so
split coverage remains complete.

Honest gaps (not run here):

- `golangci-lint` — not installed on the dev host; CI lint lane covers it.
- The `//go:build integration` orchestration e2e in `modules/message` has a
  pre-existing compile-time import cycle; unchanged by this PR.

- [x] Manually verified (focused suites + byte-equivalence baseline)

## COMPREHENSION

   `deliverCardNotification` now routes `summary.completed` /
   `summary.failed` through `Registry.Render`: preflight validates against
   each card's `data.schema.json` (C1 → 400 zero-delivery), and the card is
   rendered by the new Templates whose `Build` composes the same
   `assembleResourceCardBody` the legacy path already uses. External
   `NotifyReq` / `SummaryCardFields` are unchanged.
   `metadata.octo.{protocol,template}` + `render_profile` are added to the
   envelope; everything else in the card node is byte-identical.

   - *Boot*: Registry register-time self-check runs each new sample through
     `Render`; a bad handoff fails `installCardTmplRegistry` → `main` panics
     at startup. Mitigation: self-check + focused tests pass locally.
   - *In-flight card renderers on the receiving side*: envelopes now carry
     new `metadata.octo` and `render_profile` fields on summary cards too.
     Additive only — unknown fields must be ignored per JSON forward-compat;
     verified against the baseline.
   - *summary producer integration*: `SummaryCardFields` shape unchanged, so
     producer calls keep working. `Reason` / `TimeRange` / `Members` /
     `MsgCount` / `GeneratedAt` semantics preserved.
   - *Existing tests*: any notify test using the summary Kinds now inherits
     the wired Registry from `TestMain` — no per-test setup needed.

   `card_via_registry_summary_baseline_test.go` runs 4 fixtures end-to-end
   through both paths and asserts canonical JSON byte-equality (see Testing).
   `pkg/cardtmpl/summary_completed/template_test.go` and
   `pkg/cardtmpl/summary_failed/template_test.go` cover Registry self-check,
   metadata injection, language-localized source, `FallbackText` (zh/en/
   sanitize), and the C1 schema-rejection branches (empty taskNo, extra
   field on completed, oversize reason on failed). `modules/notify` full
   package (real MySQL/Redis/WuKongIM) green.

## Out of scope

- `generic.approval` — same reason as PR-1 (dynamic caller-supplied
  `owner`/`action_type` conflicts with static `TemplateActionContract`);
  separate decision required.
- Removal of legacy `buildSummaryCard` / `buildSummaryFallbackText` — retained
  as the byte-equivalence baseline; deletion belongs to the roadmap C
  finalization PR.
- No `NotifyReq` shape change, no `${}` engine, no B/D/E/H, no DB migration.

## Checklist

- [x] Updated documentation (`.octospec/` brief + journal + log)


lml2468 (dismissed):
# PR Review — PR #650 (Mininglamp-OSS/octo-server)

Reviewed at head `1bf3e34b` against base `main`. **Stacked on #649** (its docs.commented/shared commits are ancestors here) — I scoped the review to the **new delta since #649's head `c00af209`** (the summary.completed/failed migration); the #649 slice I already reviewed + approved separately.
**Build/vet/test:** `go build` exit 0 ✅, `go vet` ✅. `go test ./pkg/cardtmpl/summary_completed/ ./pkg/cardtmpl/summary_failed/` → **PASS**. DB-gated `modules/notify` baseline+integration → CI.
Classification: migrate two summary display cards onto the cardtmpl Registry — wire-contract + escaping/trust-boundary. Reviewed in **SECURITY** emphasis. Identical shape to #649 (which I approved).

## 1. Spec compliance — ✅
Graded against `.octospec/tasks/cardtmpl-l2a-summary-migration/brief.md`.
- **Missing:** none. `summary.completed@0.1.0` + `summary.failed@0.1.0` subpackages (handoff manifest/schema/sample + 3-method Template), notify wiring (`preflightSummarySchema` + summary branch routing), Register+SetDefault+Freeze in main, and the byte-equivalence baseline test — all present.
- **Over-build:** none. `NotifyReq`/`DocsCardFields`/summary model **byte-untouched** by this delta (`model.go` 0 diff — plan B); only the summary branches reroute.
- **Deviation:** none. Variant + attribution/label copy preserved to match legacy.

## 2. Code quality — Approved

### Findings
No P0/P1/P2 findings.

- **Byte-identical to legacy — canonical-JSON baseline test.** `TestBuildSummaryCards_MigrationBaseline` renders legacy `buildSummaryCard` vs the Registry path (4 fixtures) and asserts `bytes.Equal` after stripping the two additive `metadata.octo.{protocol,template}` fields (the only intended diff) — the strongest migration-safety anchor, reusing the PR-1 helper. (DB-gated → CI.)
- **No user-controlled URL field → the summary preflight correctly omits AbsoluteHTTPSURL.** Both schemas (`additionalProperties:false`) expose only bounded strings (`taskNo`/`title`/`timeRange`/`generatedAt`, maxLength 200/80) and bounded integers (`members`/`msgCount`, `maximum` capped — counts, not text). The only URL is the **server-composed** deep link (`env.WebLoginURL` + `/s/{taskNo}`). So there's no caller URL that could hit the schema-regex-passes→Build-fails→degrade seam — same correct asymmetry as #649.
- **Escaping preserved.** Templates compose `BuildSummaryResourceCardBodyWithLang` (extracted in resource.go), retaining the existing `escapeMarkdown`/`truncateRunes` at the render layer. Integer count fields carry no injection surface.
- **C1 + F7 wiring mirrors the pilot.** `preflightSummarySchema` → `ErrFieldsInvalid` (400 zero-delivery) before any gate; `errCardTmplUnavailable` → 500 no-degrade (composition bug visible); other render errors degrade to text. Both templates Register+SetDefault+`Freeze()` at composition (register-time sample self-check fail-closes on a bad schema/sample).
- **`additionalProperties:false`** rejects unknown fields; integer bounds (`minimum`/`maximum`) prevent absurd values.

## 3. Test assessment
Per-template tests (completed/failed) pass here; `TestBuildSummaryCards_MigrationBaseline` (legacy↔Registry byte-equivalence, 4 fixtures) is the key regression guard, DB-gated → CI. The baseline honestly documents why it tests zh-CN only (legacy button labels differ by an unrelated path) — reasonable. Good coverage.

## 4. Coverage / blind spots
- Verified by build/vet + the pure summary template suites + full read of the two templates, the shared summary body builder's retained escaping, the schemas (no URL field), the notify C1/F7 wiring, and the model-untouched check.
- NOT run: the DB-gated notify baseline + integration. Recommend the CI DB run (the byte-equivalence assertion is the safety net). Note this PR stacks #649 — merge order / squash should keep both slices' baseline tests.

## Verdict
**APPROVE** — a symmetric, low-risk continuation of the #649 L2a migration: `summary.completed`/`summary.failed` now render through the Registry with byte-identical output to legacy (canonical-JSON baseline, minus the two additive metadata fields), escaping preserved via the shared summary body builder, C1/F7 wiring matching the pilot, and no new URL attack surface (server-composed deep link; schemas are bounded strings + capped integers with `additionalProperties:false`). `NotifyReq` shape untouched. Both gates pass. This reaches the roadmap-C goal of all legacy L2a display cards on the Registry.
Suggested decision: `Safe to merge` once CI runs the DB-gated notify suite (esp. `TestBuildSummaryCards_MigrationBaseline`) — and after #649 lands (this is stacked on it). (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (dismissed):
**APPROVE** — clean roadmap-C PR-2. `summary.completed` + `summary.failed` migrate onto the cardtmpl Registry following the proven pilot/PR-1 shape, with byte-equivalence to the legacy path preserved and no contract break.

### Scope note (why 35 files / +3120)
The branch is stacked on unmerged #649, so the diff vs `main` carries #649's `docs_commented` / `docs_shared` subpackages plus the shared `assembleResourceCardBody` / `BuildDocsResourceCardBodyWithLang` extraction. The genuine PR-2 net-new is: the two `summary_completed` / `summary_failed` subpackages (template + handoff + tests), the summary wiring in `modules/notify/card.go` + `card_via_registry.go`, the summary baseline test, and `main.go` / testmain registration. No unrelated logic is bundled — bulk is template subpackages + testdata, as expected.

### Verified
- **Markdown escaping + bounds on all external fields.** Both templates compose `BuildSummaryResourceCardBodyWithLang` → shared `assembleResourceCardBody`, where title / attribution / excerpt / fact title+value are all `escapeMarkdown`'d. The failed card's error text (`reason`) flows into `excerpt`, which is `truncateRunes(_, MaxExcerptRunes)`-bounded then escaped — un-escaped error-text injection is not possible. Schema also bounds `reason` (maxLength 300 = `MaxExcerptRunes`, G9 single source), title (200), etc.
- **C1 preflight + three-way fail-close** matches the established classification: registry-unwired → `errCardTmplUnavailable` → 500 no-degrade; schema-invalid → `ErrFieldsInvalid` → 400 zero-delivery; other render error → text degrade. Preflight runs ahead of dedup / memberCache / cardSender-gate, so the empty-member 200 (`delivered=[]`) and `cardSender==nil` text-degrade paths cannot silently swallow a schema violation. Belt-and-suspenders typed re-assertion at build time.
- **No user-controlled URL field** in either summary schema — the only URL is the server-composed deep link (`summaryDeepLink(webLoginURL, taskNo, spaceID)`). Omitting an `AbsoluteHTTPSURL` preflight is correct here (same as PR-1); the shared `assembleResourceCardBody` still `requireHTTPS`-guards `IconURL`/`Source.IconURL` when present.
- **Byte-identical baseline** (`card_via_registry_summary_baseline_test.go`): 4 fixtures across both templates (incl. two failed/reason cases) assert `bytes.Equal` on canonical JSON after stripping only the two intentional `metadata.octo.{protocol,template}` fields. Plus C1 empty-taskNo rejection (both branches), F7 registry-unwired, metadata-injection assertions, and per-template unit tests covering newline-injection sanitization of title/reason, oversize-reason rejection, and English localization.
- **`SummaryCardFields` / `NotifyReq` shape untouched** (`modules/notify/model.go` not in diff — plan B). `render_profile` flows through to the send path. `renderProfile` + `renderProfileCompatibility` present in both manifests (per #647). `sanitizeLine` deduped to shared `cardtmpl.SanitizeLine` (G5), no behavior change. No in-place mutation of shipped versions (new @0.1.0 subpackages).
- **Tests**: `go test ./pkg/cardtmpl/...` all green locally incl. both new subpackages; notify baseline is not `t.Skip`-gated (runs under CI MySQL).

### Non-blocking notes 🟡
- `assembleResourceCardBody` escapes but does not rune-bound `attribution`. Here attribution is server-selected constant banner text, so not exploitable — noted only for the shared helper's general contract.
- Baseline is intentionally zh-CN only (legacy button label routes through `i18n.OutboundLanguage(background ctx→zh)` vs Registry `env.Lang` direct); English shape is covered by the per-template `TestRenderEnglishSourceLocalized`. Documented and consistent with PR-1.

### Gate
CI: `check-sprint` is failing (sprint-linkage policy check, not a build/test failure); `Build` / `Test` (MySQL) / `Vet` / `Lint` still pending at review time. Please confirm the sprint check and the pending MySQL `Test` lane go green before merge. This branch also needs #649 to land first and then rebase to `main` (the diff reduces to just the summary changes). `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED` (awaiting checks/review).

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #650 (octo-server)

## Summary

This PR migrates `summary.completed`, `summary.failed`, `docs.commented`, and `docs.shared` card templates from inline builders to a Registry-based pattern (roadmap C PR-2). It adds four new template packages under `pkg/cardtmpl/`, extends `modules/notify/card.go` and `card_via_registry.go` with preflight schema validation and Registry dispatch, refactors `resource.go` to extract a shared `assembleResourceCardBody` helper, and includes handoff manifests, baseline tests, and per-template unit tests.

The migration is structurally sound — registration wiring is complete and fail-close, data flow from upstream fields through schema validation to template consumption is correct, and the legacy/Registry paths produce byte-equivalent output as proven by baseline tests. No P0/P1 issues; five P2 items and one nit below.

## Verification

- ✅ **Registration completeness** — all four templates registered with `Register()` + `SetDefault()` before `Freeze()` in `main.go:722-732`; `TestMain` in `testmain_test.go` mirrors production registration exactly.
- ✅ **Data flow integrity** — every field consumed by template `Build()` methods is traced to upstream `SummaryCardFields`/`DocsCardFields` structs, mapped through `mapSummaryCardFieldsToJSON`/`mapDocsCardFieldsToDisplayJSON`, validated against JSON schema, and guarded by nil/empty checks before use.
- ✅ **Auth/visibility parity** — new Registry paths sit behind the same `deliverCardNotification`/`deliverDocsCardNotification` entry points; member verification, dedup, actor exclusion, and bot readiness all run before the build.
- ✅ **Error classification** — three-tier error handling is correct: `ErrFieldsInvalid` → 400, `errCardTmplUnavailable` → 500 (no degradation), other → text fallback.
- ✅ **Byte-equivalence proof** — baseline tests call real production methods (`buildSummaryCard` vs `buildSummaryCardViaRegistry`), strip metadata, compare canonical JSON with `bytes.Equal`.
- ✅ **Schema-template alignment** — all four `data.schema.json` contracts match their Go `fields` structs; `additionalProperties: false` enforced and tested.
- ✅ **Input shape edges** — nil card, empty required fields, zero-value ints, unknown state, cancelled context, and oversized fields all handled without panics.

Static analysis only at head `1bf3e34bc221594a112c651654d51a9f5024aff7`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Five P2 items and one nit below.

### P2 — New schema constraints are a breaking behavioral change (`modules/notify/card.go:49`)

The preflight schema validation is new — the legacy `buildSummaryCard()` had zero length validation. Upstream producers that previously sent `title` > 200 chars, `reason` > 300 chars, `timeRange` > 200 chars, or `generatedAt` > 80 chars would have gotten successful card delivery; now they get a 400 with zero delivery. This is likely intentional (C1 policy: schema violations must not reach dispatch), but should be verified that all upstream producers (smart-summary backend) already truncate to these limits. If producers cannot be updated in lockstep, consider server-side truncation in `mapSummaryCardFieldsToJSON` before schema validation, matching the `maxTitleRunes`/`MaxExcerptRunes` budgets already enforced by `assembleResourceCardBody`.

### P2 — Log message hardcodes "docs access-request card" for all docs card types (`modules/notify/card.go:386`)

When a `docs.commented` or `docs.shared` card hits the `errCardTmplUnavailable` path, the log message reads `"docs access-request card: cardtmpl unavailable"` regardless of the actual card kind. This will mislead incident triage. Consider using the actual `templateID` or `kind` in the log message.

### P2 — Summary fallback does not delegate to Template.FallbackText (`modules/notify/card.go:281`)

When the summary card build fails and degrades to text, `buildSummaryFallbackText` is called directly. Both `summary_completed` and `summary_failed` templates implement `FallbackText`, but it is never reached from the production path. The docs `access_requested` path delegates to `Template.FallbackText` via `templateFallbackText()` first, falling back to legacy only if that fails. If template fallback copy diverges from `buildSummaryFallbackText` in the future, no test covers this gap.

### P2 — Excerpt prefix counted inside truncation budget (`pkg/cardtmpl/summary_failed/template.go:95`)

The excerpt is constructed as `labels.failedPrefix + reason` (e.g., `"失败原因：" + reason` or `"Reason: " + reason`), then passed to `assembleResourceCardBody` which truncates to `MaxExcerptRunes` (300). With a schema-maxed reason (300 chars), the total is 305–308 runes, truncated back to 300 — losing the last 5–8 characters of the actual reason. Consider truncating the reason to `MaxExcerptRunes - len(prefix)` before concatenation, or reducing the schema `maxLength` accordingly.

### P2 — Missing diagnostic log when registry is nil (`modules/notify/card_via_registry.go:170`)

The existing `buildDocsAccessRequestCardViaRegistry` logs a structured error (`"cardtmpl DefaultRegistry unwired — composition bug"`) before returning when the registry is nil. The two new build functions (`buildDocsDisplayCardViaRegistry` here, `buildSummaryCardViaRegistry` at `:294`) skip this and only return the wrapped error. The caller does log on receipt, but the build-site log context is lost. Minor observability inconsistency.

### Nit — No English localization test for docs_shared (`pkg/cardtmpl/docs_shared/template_test.go:21`)

`TestRegisterAndRenderSample` only uses `Lang: "zh-CN"`. The three other new template packages each have a `TestRenderEnglishSourceLocalized` test. The English label branch in `labelsFor()` is entirely untested.

- C1 dual-path parity: `completed` and `failed` paths are symmetric — both go through preflight + registry. `commented` and `shared` mirror the `access_requested` preflight discipline.
- `reason` conditional inclusion: `mapSummaryCardFieldsToJSON` correctly includes `reason` only when `Kind == failed`; the `summary.completed` schema has `additionalProperties: false` and would reject it.
- `SanitizeLine` function prevents newline injection in all `FallbackText` paths.
- `assembleResourceCardBody` extraction is a behaviour-preserving refactor — legacy and Registry paths share the same assembly logic.
- Manifest contracts (ID, version, handoff root, referenced assets) match Go constants for all four templates.
- Deep-link URLs use the same `webLoginURL` from config, validated for HTTPS by the body builder.
- No new endpoints, capabilities, or authorization surface exposed.

## Verdict: APPROVED

No correctness, security, or build-blocking issues. The P2 items are non-blocking observations — the schema breaking change (P2-1) is the most actionable and should be confirmed with upstream producers before shipping. The remaining P2s are observability and consistency improvements for follow-up.





[Octo-Q] verdict: APPROVE — No P0/P1 blockers found. Five P2 non-blocking observations and one nit. Schema breaking change (P2-1) should be confirmed with upstream producers before shipping; all other items are observability/consistency improvements.

yujiawei (dismissed):
# Code Review — PR #650 (octo-server)

Reviewed at head SHA `1bf3e34bc221594a112c651654d51a9f5024aff7`, merge-base with `main` = `04b8c793` (current main). This PR is stacked on #649 (not yet merged), so the GitHub diff shows 4 commits (docs.commented/shared from PR-1 + summary.completed/failed from PR-2); this review focuses on the net PR-2 summary-card migration while verifying the shared PR-1 scaffold it depends on.

## 1. Spec compliance

**Spec: ✅**

- **No missing scope:** Migrates exactly the two cards claimed — `summary.completed` and `summary.failed` — onto the cardtmpl Registry, routed through `deliverCardNotification` (`modules/notify/card.go:59-149`). Both new subpackages ship a self-embedded `handoff/` (manifest + `data.schema.json` + sample) and a 3-method `Template`.
- **No over-build:** External `NotifyReq` / `SummaryCardFields` shape is unchanged (`modules/notify/model.go:49-59`). Legacy `buildSummaryCard` / `buildSummaryFallbackText` are retained as the byte-equivalence baseline, not deleted. No new endpoint, capability, `${}` engine, or DB migration. `reason` is correctly gated to the `failed` schema only; `summary.completed` schema is `additionalProperties:false` so a stray `reason` is rejected — and `mapSummaryCardFieldsToJSON` omits it for completed (`card_via_registry.go:263-270`).
- **No deviation:** The migration preserves the documented invariants — `Variant` constants (`summary.completed` / `summary.failed`) match legacy values, and both localized label tables (`labelsFor`) are byte-identical to legacy `summaryLabelsFor` for zh-CN and en.

**Byte-equivalence acceptance verified empirically.** I built and ran the suites at head in this checkout:
- `go build ./pkg/cardtmpl/...` — clean.
- `pkg/cardtmpl/summary_completed`, `pkg/cardtmpl/summary_failed` — green.
- `modules/notify` `TestBuildSummaryCards_MigrationBaseline` (4 zh-CN fixtures, legacy vs Registry canonical-JSON `bytes.Equal` after stripping `metadata.octo.{protocol,template}`), `TestBuildSummaryCards_MetadataOctoInjected`, `TestPreflightSummarySchema_*` — all green.
- PR-1 baselines (`docs_commented` / `docs_shared`) — green, no cross-migration regression.

## 2. Code Quality

**Quality: Approved**

Verified correct:
- **C1 preflight placement:** `preflightSummarySchema` runs before member verification, `cardSender` gate, and `cardmsg.Enabled()` (`card.go:64-91`), so a schema violation cannot be masked by an empty-member 200 or text degradation. Error classing is correct: `ErrFieldsInvalid` → 400 zero-delivery; `errCardTmplUnavailable` (registry unwired = composition bug) → 500, no silent text fallback (F7). A defense-in-depth re-assertion of both typed errors also guards the build site (`card.go:137-149`).
- **rune-vs-byte:** schema `maxLength` is rune-counted by jsonschema/v6 (consistent with the pilot), and `assembleResourceCardBody` uses `utf8.RuneCountInString` — CJC content is not spuriously rejected. (Relevant for a Chinese-first product.)
- **Refactor is behavior-preserving:** `assembleResourceCardBody` is extracted and shared by both the legacy `buildResourceCard` and the new `BuildSummaryResourceCardBodyWithLang`; the single truncation point (`MaxExcerptRunes`) is identical across paths — which is exactly why the baseline holds.
- **Sanitizer unification:** `sanitizeLine` now delegates to `cardtmpl.SanitizeLine` (single source of truth), preventing newline-injection in the text-fallback line structure; no drift risk.
- **No new trust surface:** cards embed the deep-link URL for client navigation; the server never fetches it. Deep-link origin is `External.WebLoginURL`, https-enforced by the body builder.

### Findings (all P2 / non-blocking)

- **P2 — New schema caps are a behavioral change for over-length inputs.** Legacy `buildSummaryCard` had no length validation: an over-cap `title` (>200 runes) degraded to a text DM (notification still landed), and an over-cap `reason` (>300) was truncated and rendered. Under C1 preflight, the same inputs now return 400 with **zero delivery** (`title`/`taskNo` ≤200, `timeRange` ≤200, `generatedAt` ≤80, `reason` ≤300, `members`/`msgCount` bounded). This is a deliberate, consistent application of the C1 policy established in the pilot (#632) and PR-1, but it is a real contract tightening. **Human-verify before ship:** confirm the smart-summary producer already truncates these fields to the same budgets, or accept that over-cap requests flip from "text DM delivered" to "400, nothing delivered." Server-side truncation in `mapSummaryCardFieldsToJSON` (mirroring `MaxExcerptRunes`) would preserve delivery if lockstep producer updates aren't guaranteed.
- **P2 — `members`/`msgCount` negative values now rejected.** Legacy rendered fine (omitted via the `> 0` guard); schema `minimum:0` now 400s a negative/sentinel count. Same class as above — a compatibility narrowing, benign if the producer never sends negatives.
- **P2 — Misleading log label on the docs display-card build-error path.** At `card.go:473-480`, `docs.commented` / `docs.shared` build errors are logged as `"docs access-request card ..."` and omit `zap.String("kind", ...)`, unlike the preflight branch (`card.go:390-393`) which logs `kind` correctly. Minor triage friction; not a runtime defect. (Note: the first-pass advisory cited `card.go:386` for this — that line is actually correct; the real instance is the build-error branch at 473/478.)
- **P2 — Summary text-fallback does not delegate to `Template.FallbackText`.** The degrade path calls legacy `buildSummaryFallbackText` directly; the new Templates' `FallbackText` is only exercised by unit tests, never the production path. No divergence today (copy is identical), but there is no test locking legacy-vs-Template fallback parity, so future drift would be silent. Observability/consistency follow-up.
- **P2 — Missing build-site diagnostic log when registry is nil** in `buildSummaryCardViaRegistry` / `buildDocsDisplayCardViaRegistry`, unlike `buildDocsAccessRequestCardViaRegistry` which logs before returning. The caller logs on receipt, so no context is truly lost; minor inconsistency.
- **Nit — No English-localization test for `docs_shared`** (`template_test.go` only covers zh-CN), whereas the other three new packages have a `TestRenderEnglishSourceLocalized`. The en `labelsFor` branch is untested for that one card.

### Refuted / non-issues
- The first-pass "excerpt prefix counted inside the 300-rune truncation budget" concern is **not a drift**: legacy `buildSummaryCard` performs the identical `failedPrefix + reason` concatenation and truncates at the same `MaxExcerptRunes`. Behavior is unchanged; "fixing" it would *break* byte-equivalence. Pre-existing and out of scope.
- Whitespace `TrimSpace` in `mapSummaryCardFieldsToJSON` (title/taskNo) vs legacy raw render is a theoretical byte-difference only for whitespace-padded inputs; `taskNo` is a server-generated ID and `title` is producer-formatted — not reachable in practice.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved, with zero reachable P0/P1. All P2s are non-blocking observability/coordination items. The one item worth a human glance before merge is the C1 schema-cap tightening (P2-1): confirm the upstream summary producer truncates to the new budgets so over-length notifications don't silently flip from delivered-as-text to 400-zero-delivery.

## 4. Suggestions
- Before merge (or as a fast follow), confirm smart-summary producer field caps ≤ the new schema `maxLength`s, or add server-side truncation in `mapSummaryCardFieldsToJSON` to preserve delivery.
- Add `zap.String("kind", card.Kind)` and a kind-accurate message to the docs build-error branch (`card.go:473/478`).
- Add a `docs_shared` English-localization test to close the coverage asymmetry.

## 5. Additional notes
- This branch is stacked on #649; merge #649 first. After #649 lands and this rebases to `main`, the diff reduces to the summary-only changes — re-confirm the merged result compiles (sibling-PR semantic check).
- Security-sensitive classification is appropriate. The load-bearing manual-verify item is the C1 delivery-semantics change (P2-1); everything else is byte-equivalence-locked by the baseline test, which I confirmed passes at this head.

lml2468 (dismissed):
# PR Review (re-review) — PR #650 (Mininglamp-OSS/octo-server)

Re-reviewed at **new head `29c11429`** against base `main`. Re-affirms my APPROVE at `1bf3e34b`. Still **stacked on #649**; scoped to the new delta.
**Build/vet/test:** `go build` exit 0 ✅, `go vet` ✅. `go test ./pkg/cardtmpl/ ./pkg/cardtmpl/summary_completed/ ./pkg/cardtmpl/summary_failed/ ./pkg/cardtmpl/docs_shared/` → **PASS**. DB-gated notify baseline/integration → CI.

## What changed since `1bf3e34b` (delta reviewed)
One commit — `29c11429` "address PR #650 review — preserve delivery on over-length …" — resolving the over-length-compat concern peers raised (mochashanyao on #650, yujiawei on the sibling #649).

### Over-length display fields now truncate-and-deliver (compat restored) ✅
The pre-fix migration would 400-reject an over-length display field, whereas legacy `buildSummaryCard` never validated length (it delivered a truncated card / text DM). `mapSummaryCardFieldsToJSON` now **truncates** `title`/`timeRange`/`generatedAt`/`reason` to their schema maxLength via the shared `cardtmpl.TruncateRunes`, preserving the "notification never lost" invariant and matching legacy. Verified correct:
- **`taskNo` is deliberately NOT truncated** — it's the deep-link key (`/s/{taskNo}`); a silent truncation would produce a wrong link, so an over-long taskNo still hits **C1 400**. Correct field-level distinction (structural key vs cosmetic display).
- **Structural violations still 400**: missing/empty required, type errors, unknown fields (`additionalProperties:false`), over-long taskNo — C1 hard-fail preserved for genuine contract breaches, only cosmetic over-length is softened.
- **`clampNonNegative`** on members/msgCount reproduces legacy's `>0` omit-and-deliver (the schema `minimum:0` would otherwise 400 a negative sentinel).
- **Single source of truth (G9):** `MaxTitleRunes` exported from resource.go so the ingress truncation and the render guard share one cap — truncation always leaves the field within schema, so it can't post-truncate 400. `TestMapSummaryFields_TruncatesDisplayFields` locks the notify-local caps in sync with schema (fails if any cap > schema).

### Byte-equivalence intact
The summary baseline test is extended for the over-length path; the pure cardtmpl/summary/docs_shared suites pass here. `card.go` delta is only log-message generalization (adds `kind`), no behavior change.

## Cross-PR consistency note (not a #650 blocker)
The **docs-display** mapping `mapDocsCardFieldsToDisplayJSON` (commented/shared, introduced in #649) still only `TrimSpace`s — it does **not** truncate, so an over-length docs-display field (title/actorName/excerpt) still 400s, which is exactly yujiawei's open 🔴 on **#649**. This PR fixed only the summary mapping (correct scope — #650 is summary-only). Since #650 stacks on #649, the same `TruncateRunes` treatment should be applied to the docs-display mapping **in #649** for consistency, so the two slices behave identically. Flagging for the author to mirror the fix on #649; out of scope for #650 itself.

## Verdict
**APPROVE** (re-affirmed at `29c11429`) — the over-length compat concern is correctly resolved for the summary cards: cosmetic display fields truncate-and-deliver (preserving the never-lost invariant and legacy parity), while the deep-link key `taskNo` and genuine structural violations keep the C1 400, with the ingress caps locked in sync with schema via the exported single-source cap and a dedicated test. Byte-equivalence and escaping unchanged. Both gates pass.
Suggested decision: `Safe to merge` after #649 lands (stacked) + CI runs the DB-gated notify suite. **Recommend mirroring this truncate-not-400 fix onto #649's `mapDocsCardFieldsToDisplayJSON`** so commented/shared match summary's over-length behavior (yujiawei's open #649 finding). (PR `merge: BLOCKED` — branch-protection.)

Jerry-Xin (dismissed):
Re-review at `29c11429` (was APPROVED @ `1bf3e34b`). The single new commit is a focused, correct response to the round-1 P2 notes. Delta is exactly 1 commit ahead of my approved head (no rebase/drift): the summary ingress mapping, two log strings, docs, and tests — the summary render/escape core I approved is byte-stable. Re-APPROVE.

## What the new commit changed vs `1bf3e34b`

- **P2-1 (delivery regression on over-length display fields) — fixed.** `mapSummaryCardFieldsToJSON` (`modules/notify/card_via_registry.go`) now server-side truncates the display fields to their schema/render budgets before the C1 preflight, so an over-length string renders truncated instead of flipping to a 400 zero-delivery: `title`→`cardtmpl.MaxTitleRunes` (200), `reason`→`cardtmpl.MaxExcerptRunes` (300), `timeRange`→200, `generatedAt`→80. All four caps match the respective `data.schema.json` `maxLength` exactly, and `TruncateRunes` appends a single ellipsis so the result stays within cap. `taskNo` is deliberately NOT truncated (deep-link key → over-length stays a genuine C1 400). Negative `members`/`msgCount` are clamped to 0, reproducing the legacy `> 0` omit-and-deliver behavior. `cardtmpl.MaxTitleRunes` + `cardtmpl.TruncateRunes` are exported (`pkg/cardtmpl/resource.go`) so ingress and render share one cap/impl — clean G9 single-source-of-truth, internal `truncateRunes` unchanged.
- **P2-2 (misleading log label) — fixed.** The docs display-card build-error branch (`modules/notify/card.go`) no longer hardcodes `"docs access-request card ..."` for `docs.commented`/`docs.shared`; it is now kind-generic and adds `zap.String("kind", ...)`. Cosmetic, correct.
- **Test nit — added.** `TestRenderEnglishSourceLocalized` for `docs_shared` (parity with the other three packages).

## Safety surface re-verified at `29c11429`

- **Escaping intact.** No escape/sanitize/builder lines were touched in the delta. External fields (`title`, `timeRange`, `generatedAt`, and the failed card's `reason`) still flow through the shared `BuildSummaryResourceCardBodyWithLang` builder for the rendered card; truncation happens on raw input at ingress and escaping remains downstream and unchanged. No new injection surface.
- **FallbackText path also sanitizes.** Both `summary_completed` and `summary_failed` `FallbackText` implementations call `cardtmpl.SanitizeLine(...)` on every external field. The round-1 P2-3 note (notify's text fallback delegating to `Template.FallbackText`) is explicitly deferred — safe, because both implementations are copy-identical and both sanitize, so no divergence and no raw-text emission today.
- **Baseline equivalence preserved.** Within-budget inputs are unchanged, so `TestBuildSummaryCards_MigrationBaseline` (canonical-JSON `bytes.Equal` for both templates) still holds. `TestMapSummaryFields_TruncatesDisplayFields` locks the new behavior (huge input delivers post-truncation, over-length `taskNo` stays 400) and would fail if any local cap exceeded the schema `maxLength`. Tests use the real WuKong test server / notify harness — not `t.Skip`.
- C1 preflight fail-close + three-way error classification (schema→400 zero-delivery; template-unregistered→500 no-degrade; other→degrade) unchanged; server-composed deep link only (no user URL field → `AbsoluteHTTPSURL` omission still correct); `NotifyReq`/model shape untouched.

## 🟡 Non-blocking

- **Count upper bound is the one asymmetry left in this fix.** The commit truncates over-length display *strings* to preserve delivery but only floors *negative* `members`/`msgCount`; it does not cap the upper bound. The schemas reject `members > 1,000,000` and `msgCount > 1,000,000,000`, and legacy `buildSummaryCard` would have delivered any positive int, so at those boundaries the new preflight is a 400 zero-delivery — the same over-cap→400 family already shipped as the C1 contract in the prior slice. These are server-composed summary stats where a single-summary value at that magnitude is not realistic, so this is not delivery-affecting in practice; but for full symmetry with the display-field fix, consider clamping the counts to the declared maxima (with boundary tests) in the roadmap-C finalization PR, or documenting them as intentional hard-400s alongside `taskNo`.

## ✅ Highlights

- Exact cap/schema alignment with a test that fails if they drift out of sync — the right way to keep ingress truncation and schema in lockstep.
- `taskNo` correctly excluded from truncation (deep-link integrity over silent corruption).
- Delivery-preserving truncation is the correct resolution of the P2-1 regression; the deferrals (fallback delegation, counts) are documented with rationale rather than silently dropped.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

## Code Review: octo-server PR #650 — Migrate summary.completed + summary.failed to Registry
**Head SHA:** `29c114293e6d419c345649a86ad85b5ccac04406`
**Scope:** 22 files (PR-only), +1856 / -14 — roadmap C PR-2: summary card Registry migration
**Diff range:** `pr-649..HEAD` (3 commits: feat, docs, fix)

---

### 1. Verification Summary

| Area | Verdict | Evidence |
|------|---------|----------|
| C1 preflight placement | ✅ | `card.go:49-75` — before memberCache, cardSender gate, cardmsg.Enabled check |
| F7 fail-close (Registry unwired) | ✅ | `card.go:57-62` + `card_via_registry.go:300-302` — returns `errCardTmplUnavailable` → 500 |
| Byte-equivalent baseline | ✅ | `card_via_registry_summary_baseline_test.go` — 4 fixtures (completed full/minimal, failed with-reason/full), strip metadata.octo.{protocol,template}, canonical JSON compare |
| Truncation ↔ schema alignment | ✅ | `card_via_registry.go:253-256` caps (200/300/200/80) match schema maxLength; `TestMapSummaryFields_TruncatesDisplayFields` locks alignment |
| Deep-link https enforcement | ✅ | `resource.go:368-373` — `webOrigin` rejects non-https; `summaryDeepLink` uses `url.PathEscape(taskID)` + `url.QueryEscape(spaceID)` |
| Composition root parity | ✅ | `main.go:725-730` registers both templates before Freeze; `testmain_test.go` mirrors exactly |
| Error classification | ✅ | `card.go:137-153` — ErrFieldsInvalid→400, errCardTmplUnavailable→500, other→text fallback |
| Label parity | ✅ | Baseline tests verify byte-equivalence; legacy `summaryLabelsFor` (card.go:821-855) matches new `labelsFor` in both templates |

---

### 2. Findings

**No P0/P1 issues found.**

#### P2 — Minor observations (non-blocking)

**P2-1: Redundant `mapSummaryCardFieldsToJSON` call in preflight + build**
- **Path:** `card.go:62` calls `preflightSummarySchema` which calls `mapSummaryCardFieldsToJSON` (card_via_registry.go:338); then `card.go:126-131` calls `buildSummaryCardViaRegistry` which calls it again (card_via_registry.go:318).
- **Impact:** Double JSON marshal + truncation per notification. Negligible for summary-notify volume.
- **Diff-scope:** New (PR#650 introduces the summary preflight+build pair).
- **Suggestion:** Not actionable for this PR. If summary-notify volume grows significantly, consider passing mapped JSON from preflight to build.

**P2-2: `maxTitleRunes` local alias retained alongside exported `MaxTitleRunes`**
- **Path:** `resource.go:27-28` — `maxTitleRunes = MaxTitleRunes` kept for backward compatibility within the package.
- **Impact:** Zero functional impact. Minor readability nit — two names for one constant.
- **Diff-scope:** New (PR#650 export).
- **Suggestion:** Can be cleaned up in a follow-up when internal references migrate to `MaxTitleRunes`.

---

### 3. Data Flow Trace

| Consumed Data | Upstream Source | Intermediate Transforms | Reaches Consumer? |
|---|---|---|---|
| `card.TaskNo` | `SummaryCardFields.TaskNo` (caller JSON) | `TrimSpace` only, **not truncated** (deep-link key) | ✅ Schema validates minLength=1, maxLength=200; over-length → C1 400 |
| `card.Title` | `SummaryCardFields.Title` | `TrimSpace` → `TruncateRunes(_, 200)` | ✅ Within schema maxLength=200; template `assembleResourceCardBody` applies same cap |
| `card.Reason` | `SummaryCardFields.Reason` | `TrimSpace` → `TruncateRunes(_, 300)`; **only for Kind==failed** | ✅ Failed schema maxLength=300; completed schema has `additionalProperties:false` — field excluded from JSON |
| `card.Members` | `SummaryCardFields.Members` | `clampNonNegative` (negative→0) | ✅ Schema minimum=0; template `> 0` guard omits zero/negative from display |
| `card.MsgCount` | `SummaryCardFields.MsgCount` | `clampNonNegative` (negative→0) | ✅ Same as Members |
| `card.TimeRange` | `SummaryCardFields.TimeRange` | `TrimSpace` → `TruncateRunes(_, 200)` | ✅ Template conditionally includes when non-empty |
| `card.GeneratedAt` | `SummaryCardFields.GeneratedAt` | `TrimSpace` → `TruncateRunes(_, 80)` | ✅ Template conditionally includes when non-empty |
| `card.Kind` | `SummaryCardFields.Kind` | Routes to completed/failed template ID | ✅ Controls reason field inclusion + template selection |
| `env.WebLoginURL` | `Config.External.WebLoginURL` | `webOrigin` validates https + extracts origin | ✅ Deep-link = `origin/s/{PathEscape(taskNo)}?sp={QueryEscape(spaceID)}` |
| `renderProfile` | Manifest `renderProfileCompatibility` = `"octo-chat/v1"` | `RenderCardWithProfiles` extracts from payload | ✅ Written to `carddispatch.Card.RenderProfile`; dispatch validates via `IsAcceptedRenderProfile` |

---

### 4. Blind Spot Checklist (C1–C6)

**C1 — Dual-path parity:** ✅ Clear
- completed vs failed template paths: shared `mapSummaryCardFieldsToJSON` (truncation + clamping), shared `BuildSummaryResourceCardBodyWithLang` (body assembly). Diff = reason field + attribution/excerpt. Baseline tests cover both.

**C2 — Control-flow ordering / nested reuse:** ✅ Clear
- Preflight runs once at ingress (before gates). Build runs once under `canCard`. No nested/double invocation. `clampNonNegative` is pure. Truncation functions are stateless.

**C3 — Authorization boundary:** N/A
- No new endpoints, tools, or capabilities exposed. Summary notifications are server-initiated internal flows.

**C4 — Authorization lifecycle / container-member state cascade:** N/A
- No auth changes in this PR.

**C5 — Build/note vs runtime path:** ✅ Clear
- Byte-equivalent baseline tests (`TestBuildSummaryCards_MigrationBaseline`) verify runtime output parity, not just compile-time correctness. `TestMapSummaryFields_TruncatesDisplayFields` verifies runtime truncation alignment.

**C6 — Governance/policy/document self-consistency:** ✅ Clear
- `docs/platform-card-base.md` updated to reflect 5 L2a cards on Registry. Notes `generic.approval` not migrated (dynamic owner conflict) — consistent with PR scope.

---

### 5. Cross-Round Blocker Recheck (R6)

---

- **Legacy `buildSummaryCard` retained** at `card.go:226` — no longer on production path but kept for baseline test comparison. Correct decision; removing it would eliminate the byte-equivalence safety net.
- **`SanitizeLine` / `TruncateRunes` exported** from `pkg/cardtmpl` — clean shared utility pattern consistent with PR-1 (docs migration).
- **`TestMain` parity** with `main.go:installCardTmplRegistry()` — test infrastructure correctly mirrors production composition root, preventing F7 false negatives in tests.

---

### 7. Verdict

Well-structured migration with thorough test coverage. C1 preflight correctly placed before all gates, F7 fail-close preserved, byte-equivalent baseline locks output parity, and the P2-1 truncation fix (from the fix commit) correctly preserves the "notification never lost" invariant for over-length display fields while keeping structural violations (empty taskNo, over-length deep-link key) as hard 400s.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; clean architecture with comprehensive test coverage and correct error classification.

Jerry-Xin (dismissed):
The migration is relevant and well structured, but introduces a delivery regression for large counters.

### 🔴 Blocking

- 🔴 Critical — Previously valid notifications can now be rejected with zero delivery. [`card_via_registry.go:315`](/Users/mlamp/.openclaw/tmp/pr-review-dpwa4lfo/repo/modules/notify/card_via_registry.go:315) only clamps negative counters, while both schemas impose upper bounds (`members <= 1,000,000`, `msgCount <= 1,000,000,000`), for example [`data.schema.json:25`](/Users/mlamp/.openclaw/tmp/pr-review-dpwa4lfo/repo/pkg/cardtmpl/summary_completed/handoff/summary.completed@0.1.0/contract/data.schema.json:25). The legacy renderer accepted every positive `int`; larger values were rendered or delivered through text fallback. The new preflight instead returns 400 before member checks or fallback, contradicting the stated delivery-preservation fix. Either clamp these values to the schema maxima or remove the maxima, and add boundary tests for both counters and templates.

### 💬 Non-blocking

- 🟡 Warning — [`card_via_registry.go:312`](/Users/mlamp/.openclaw/tmp/pr-review-dpwa4lfo/repo/modules/notify/card_via_registry.go:312) trims `taskNo`, although the legacy path used the original value in the deep link. This silently changes an identity-bearing field despite the comment saying it must not be silently modified. Consider rejecting surrounding whitespace or preserving the original value.

### ✅ Highlights

- The registry wiring, fail-closed behavior, metadata checks, and focused template tests follow existing architecture.
- The shared rendering helper substantially reduces layout drift.
- Focused tests for both new template packages pass.

Jerry-Xin (dismissed):
Re-review at `ffc7b002` after the previously approved `29c11429`. **APPROVE**.

## True delta since the last approved head

This is a rebase onto the merged #649 head (`dc79aa4d`) plus #649's follow-up delivery-preservation fix for `docs.commented` / `docs.shared`. The two-dot delta is limited to eight PR-1 files: docs-display ingress truncation and its tests, two schema-cap tests, comments, formatting, and roadmap notes. The PR-2 summary implementation is byte-stable: `modules/notify/card.go`, the summary baseline test, both summary template packages and schemas, registration in `main.go`, and notify `TestMain` have identical blobs to `29c11429`.

## Re-verified safety surface

- **Escaping remains intact.** Summary title, time range, attribution, generated time, and especially the failed-card reason still pass through the shared summary body builder, which bounds and escapes Markdown. Both fallback implementations sanitize external lines. No raw external text path was introduced.
- **Over-length delivery fix remains intact.** Display fields are truncated to shared schema/render caps before preflight; negative counts are clamped to zero; `taskNo` is intentionally not truncated because it is the deep-link key and remains a hard C1 400 when oversized.
- **C1/F7 behavior is unchanged.** Schema errors return 400 before delivery, an unregistered template returns 500 without degradation, and other render failures degrade to sanitized text.
- **URL handling remains correct.** There is no caller-controlled URL field; the deep link is server-composed, so no `AbsoluteHTTPSURL` input check is required.
- **Baseline and model shape remain intact.** The canonical-JSON `bytes.Equal` baseline for both summary templates is unchanged and passed in the MySQL CI lane. `NotifyReq` / summary model shape is untouched.

## Fresh delta review

The #649 follow-up correctly applies the same truncate-to-deliver distinction to docs display cards: cosmetic fields truncate to schema caps, while identity-bearing `docId` remains unmodified and hard-fails when oversized. Dedicated tests cover successful post-truncation preflight/build, oversized `docId`, and schema/render cap alignment. No regression into the summary slice.

## Non-blocking note 🟡

The previously documented theoretical upper-bound asymmetry for summary counters remains unchanged: values above the schema maxima hard-400 whereas legacy accepted any positive `int`. These are server-produced aggregate counts at operationally unreachable magnitudes, so this remains a follow-up/documentation consideration rather than a merge blocker.

## Gate

Build, Test (MySQL), Vet, Lint, card-dispatch lint, i18n checks, dependency review, and security scans pass. `check-sprint` alone fails because the PR has no sprint/issue attribution; that is a governance gate, not a code failure. The branch is `MERGEABLE`; `mergeStateStatus=BLOCKED` only because required policy/review gates are not all satisfied.

lml2468 (dismissed):
# PR Review (re-review) — PR #650 (Mininglamp-OSS/octo-server)

Re-reviewed at **new head `ffc7b002`** against base `main`. Re-affirms my APPROVE at `29c11429`.
**Build/vet/test:** `go build ./modules/notify/ ./pkg/cardtmpl/...` exit 0 ✅, `go vet` ✅. `go test ./pkg/cardtmpl/summary_completed/ ./pkg/cardtmpl/summary_failed/` → **PASS**. DB-gated notify baseline/integration → CI.

## What changed since `29c11429` (delta reviewed)
The stack was **rebased**: #649 has landed into main-history (new merge-base `dc79aa4` = merged #649; #645 also merged), so #650's compare is now clean — its own delta is exactly the summary slice, no longer bloated by #649's files.

Diffing my last-approved `29c11429` → `ffc7b002`, the only code change (77 lines, 2 files) is the **#649 docs-display truncation fix pulled in by the rebase** — `mapDocsCardFieldsToDisplayJSON` now truncates title/actorName/excerpt/updatedAt (docId kept as hard-400 deep-link key) + the resource.go G9 comment update. **This is byte-identical to the #649 fix I reviewed and approved at `e6b7454a`.** #650's own summary production code (`summary_completed`/`summary_failed` templates, `mapSummaryCardFieldsToJSON`, card.go summary routing) is **unchanged** since `29c11429`.

## Verified clean (re-confirmed)
- **Summary truncate-and-deliver** (title→200 / timeRange→200 / generatedAt→80 / reason→300) + `taskNo` hard-400 + `clampNonNegative` counts — unchanged from my `29c11429` approval.
- **The pulled-in #649 fix** is the symmetric docs-display truncation I already approved (title/actorName/excerpt/updatedAt truncate, docId 400, schema-locked caps) — now co-present in this tree, so both slices behave identically (the cross-PR consistency I asked for is realized here by the rebase).
- Escaping unchanged (shared body builders still escapeMarkdown/truncateRunes); byte-equivalence baselines for both docs-display and summary intact; `NotifyReq` shape untouched.

## Verdict
**APPROVE** (re-affirmed at `ffc7b002`) — the new head is a rebase of the summary slice onto the now-merged #649; #650's own summary code is unchanged from my prior approval, and the only incoming diff is the #649 docs-display truncation fix I already reviewed and approved. Over-length compat (truncate-and-deliver) now holds symmetrically for both docs-display and summary cards; deep-link keys (docId/taskNo) and structural violations keep C1 400; escaping and byte-equivalence intact. Both gates pass.
Suggested decision: `Safe to merge` once CI runs the DB-gated notify suite (summary + display byte-equivalence baselines). (PR `merge: BLOCKED` — branch-protection.)

yujiawei (approved):
# Code Review — PR #650 (octo-server)

**Scope:** `feat(cardtmpl): migrate summary.completed + summary.failed to Registry` reviewed at head `5fa6bd1c3e0053106c37e67f07aa0acf2705736c`, merge-base `dc79aa4d` (current `main`, which already contains the PR-1 docs migration). Net diff is PR-2 only: +1941/-9, 21 files. Two follow-up commits (`ffc7b002`, `5fa6bd1c`) since the prior review round address the delivery-regression and identity-field concerns raised earlier. This is a security-sensitive change, so URL construction, field validation, and fail-close behavior were checked explicitly.

Verified locally at this head: `go build ./pkg/cardtmpl/... ./modules/notify/...` and `go vet` clean; `pkg/cardtmpl/summary_completed` + `summary_failed` `-race` green; `modules/notify` summary suite green (`TestMapSummaryFields_TruncatesDisplayFields`, `TestMapSummaryFields_ClampsOverMaxCounts` {completed,failed}, `TestMapSummaryFields_PreservesTaskNoVerbatim`, `TestBuildSummaryCards_MigrationBaseline` × 4 fixtures, `TestBuildSummaryCards_MetadataOctoInjected`, `TestPreflightSummarySchema_C1RejectsEmptyTaskNo`, `TestPreflightSummarySchema_F7RegistryUnwired`).

---

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. Both legacy summary display cards (`summary.completed`, `summary.failed`) are migrated onto `Registry.Render`. New self-embedded subpackages ship manifest + `contract/data.schema.json` + sample + a 3-method `Template`. Notify wiring (`buildSummaryCardViaRegistry` + `preflightSummarySchema` + `mapSummaryCardFieldsToJSON`) routes both kinds through the Registry, and both are registered in the composition root (`main.go:722`) and mirrored in the test registry (`testmain_test.go`).
- **Over-build:** none. `generic.approval` is correctly left on the legacy path (dynamic-owner conflict, tracked separately). External `NotifyReq` / `SummaryCardFields` shapes are unchanged. Legacy `buildSummaryCard` is retained only as the byte-equivalence baseline reference; deletion is deferred.
- **Deviation:** none. The byte-equivalence baseline (`card_via_registry_summary_baseline_test.go`) asserts canonical-JSON equality between legacy and Registry output after stripping `metadata.octo.{protocol,template}` across 4 zh-CN fixtures (completed full/minimal, failed with-reason/full).

The two follow-up commits are in-scope corrections to this PR's own delivery contract, not scope creep.

---

## 2. Code Quality

**Quality: Approved**

Data-flow was traced for every consumed field from the mapping layer through `Registry.Render` to the deep-link and fail-close branches. No P0/P1 defects are reachable.

**Verified correct:**

- **Delivery preservation (the core of this round).** `mapSummaryCardFieldsToJSON` (`card_via_registry.go:332`) now truncates display fields to their schema `maxLength` (`title`→200, `timeRange`→200, `generatedAt`→80, `reason`→300, via the shared `cardtmpl.TruncateRunes`, which is rune-counted, matching `jsonschema/v6`), and clamps `members`/`msgCount` to schema `[0, maximum]` via `clampCount`. This restores the legacy "notification is never lost" invariant: an over-length or over-max field renders truncated/capped rather than being turned into a hard 400 zero-delivery by the new schema. `TestMapSummaryFields_ClampsOverMaxCounts` locks the caps in sync (post-clamp the value must still pass preflight, else the test fails), and `TestMapSummaryFields_TruncatesDisplayFields` does the same for the string caps.
- **`taskNo` handled as an identity/deep-link key.** It is passed verbatim (no `TrimSpace`, no truncation) — byte-identical to legacy `buildSummaryCard`, which builds `/s/{taskNo}` from the raw value — with the schema `minLength:1/maxLength:200` as the only backstop. Empty/blank `taskNo` is rejected at the `deliverCardNotification` ingress (`card.go:42`) before mapping. Downstream, `summaryDeepLink` applies `url.PathEscape(taskID)` (`resource.go:322`), so a verbatim `taskNo` cannot break out of the path segment. `taskNo` is a server-generated `summary_task.task_no`, not caller free-text.
- **C1 preflight cannot be masked by a delivery gate.** `preflightSummarySchema` runs at `card.go:65` — before `memberCache`, before the `cardSender != nil && cardmsg.Enabled()` gate, and before dispatch. So a schema violation deterministically returns 400 zero-delivery and can't be hidden behind an empty-member 200 or a text degradation. A defense-in-depth typed re-assertion at the build site (`card.go:137-148`) catches any future Build-internal validation that bypasses preflight.
- **Fail-close (F7) is correct and typed.** Registry unwired / template not registered → `errCardTmplUnavailable`, which the API layer maps to 500 (`api.go:318`) with no legacy text fallback; `ErrFieldsInvalid` → `errNotifyCardInvalid` → 400 (`api.go:314`). Only a non-typed render error degrades to text. `TestPreflightSummarySchema_F7RegistryUnwired` covers the unwired path.
- **Count-omission parity.** Templates keep the legacy `if Members > 0` / `if MsgCount > 0` guards, so `clampCount(negative)→0` reproduces legacy "omit the fact" for non-positive counts, and the `reason` field is only mapped for `SummaryCardKindFailed` (the `completed` schema is `additionalProperties:false`).
- **Registration symmetry** between `main.go` and `testmain_test.go` is exact (Register → SetDefault → Freeze for both new templates).

**P2 findings (non-blocking):**

- **P2-1 — Over-max count clamps to an exact maximum, which can misstate a statistic.** `clampCount` (`card_via_registry.go:305`) renders a value `> maximum` as exactly `1,000,000` / `1,000,000,000` rather than the true figure. This is a deliberate, byte-equivalence-preserving trade (the alternative — a 400 zero-delivery — is strictly worse, and for any realistic chat-summary count the clamp never triggers). Worth a follow-up only if a "1,000,000+"-style rendering is preferred over an exact-but-capped number. **Human-verify:** confirm the summary producer's `members`/`msgCount` cannot realistically exceed these caps (for a single chat summary they cannot).
- **P2-2 — `taskNo` is length-bounded but not grammar-validated.** Any non-blank value ≤200 runes (including internal whitespace, control chars, or arbitrary Unicode) reaches the deep-link and routing consumer. This is URL-safe (`PathEscape`) and byte-equivalent with legacy, so it is not a regression; it simply doesn't assert the task-number grammar at ingress. **Human-verify:** since `taskNo` originates server-side from `summary_task`, confirm there is no path by which a caller supplies a malformed `taskNo` that a downstream `/s/{taskNo}` consumer would mishandle.
- **P2-3 — Text-fallback parity is untested.** `buildSummaryFallbackText` (legacy path) is used when the card build degrades; its output is not asserted against the Registry template's own fallback. Low risk (fallback only fires on non-typed render errors, which the schema/preflight make unlikely), noted as a test gap.
- **P2-4 — Baseline fixtures are zh-CN only.** The English display branch is covered by the template unit tests rather than the cross-path byte-equivalence baseline; acceptable given the documented `i18n.OutboundLanguage(ctx)` vs `env.Lang` decoupling, noted for awareness.

---

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved, with zero reachable P0/P1. The delivery-regression concern from the prior round (schema `maxLength`/`maximum` flipping over-length/over-max inputs into 400 zero-delivery) is genuinely resolved by server-side truncation + count-clamping, and the `taskNo`-trim divergence is fixed by verbatim pass-through. All findings are P2/nit.

---

## 4. Suggested follow-ups

- If exact large counts matter to the summary UX, render capped counts as `"<max>+"` in `mapSummaryCardFieldsToJSON` rather than a bare capped integer (P2-1).
- Add a `>cap` boundary case and a text-fallback parity assertion to the summary baseline test (P2-3).

## 5. Additional observations

- The hard-coded per-field truncation caps (`summaryTimeRangeMaxRunes`, `summaryGeneratedAtMaxRunes`, `summaryMembersMax`, `summaryMsgCountMax`) duplicate the schema's `maxLength`/`maximum`. The sync tests guard against drift today; a future refactor could let the Template declare its own caps so the mapping layer reads one source of truth (same coupling note the code already flags for the avatar-URL check in the docs pilot).
- No user-controlled URL field exists on either summary schema — the only URL is the server-composed `summaryDeepLink(webLoginURL, taskNo, spaceID)`, so the avatar-style `AbsoluteHTTPSURL` preflight correctly does not apply here.

lml2468 (approved):
# PR Review (re-review) — PR #650 (Mininglamp-OSS/octo-server)

Re-reviewed at **new head `5fa6bd1c`** against base `main`. Supersedes my prior reviews (all auto-dismissed on head advance: `1bf3e34b`, `29c11429`, `ffc7b002`). This head adds one new commit that resolves the two Jerry-Xin PR #650 re-review points.
**Build/vet/test:** `go build ./modules/notify/ ./pkg/cardtmpl/...` exit 0 ✅, `go vet` ✅. `go test ./modules/notify/ -run 'MapSummaryFields|SummaryCards_MigrationBaseline'` → **PASS** (pure). Full DB-gated notify suite → **CI Test job PASS (13m33s)**; Build/Vet/Lint/i18n/osv/secret-scan all green.

## Delta since `ffc7b002` (the only new code this round)
One commit — `fix(cardtmpl): clamp summary counts to schema maxima + preserve taskNo verbatim`: **`card_via_registry.go` +38/−13, `card_via_registry_summary_baseline_test.go` +96**. Two targeted fixes, both answering Jerry-Xin's re-review:

### 1. Count over-max clamp — ✅ symmetric with the display-field truncation fix
`clampNonNegative(v)` → `clampCount(v, limit)`: previously counts were only floored at 0; a caller-supplied `members`/`msgCount` **above the schema `maximum`** would hit C1 preflight and flip to a **400 zero-delivery**. Legacy `buildSummaryCard` rendered any positive int, so this is the same delivery regression the display-field truncation already fixed — now applied symmetrically to the counters. Clamps to `[0, summaryMembersMax=1_000_000]` / `[0, summaryMsgCountMax=1_000_000_000]`.
- **Cap ↔ schema alignment byte-verified:** both `summary.completed` and `summary.failed` `data.schema.json` declare `members maximum: 1000000` and `msgCount maximum: 1000000000` — **exactly** the two local constants. So a clamped value always sits inside the schema range → never 400s. `TestMapSummaryFields_ClampsOverMaxCounts` locks this both at the unit layer (clamp) and the integration layer (preflight PASS + build SUCCESS for both templates) — a local cap above schema would 400 after clamp and fail the test.
- Negatives still floor to 0 (legacy omit-and-deliver via the Template's `> 0` guard) — unchanged.

### 2. `taskNo` verbatim passthrough — ✅ restores byte-parity with legacy deep-link
`strings.TrimSpace(card.TaskNo)` → `card.TaskNo` (raw). **Verified against legacy:** `buildSummaryCard` builds the deep-link from the **raw** `card.TaskNo` (`card.go:254` → `BuildSummaryResourceCard(ctx, webLoginURL, card.TaskNo, …)`), so trimming in the registry path would silently rewrite the identity key and byte-drift from legacy for padded inputs. **Empty/whitespace-only taskNo is already rejected at ingress** (`card.go:42`: `strings.TrimSpace(card.TaskNo) == "" → errNotifyCardInvalid`), so verbatim passthrough here can't leak a blank deep-link. `TestMapSummaryFields_PreservesTaskNoVerbatim` locks the no-trim contract. taskNo remains **not** length-truncated (deep-link key → over-length stays a genuine C1 400).

## Verified clean (carry-over, still holds)
Summary display truncate-and-deliver (title→200 / timeRange→200 / generatedAt→80 / reason→300) + taskNo hard-400 on over-length; the pulled-in #649 docs-display truncation (title/actorName/excerpt/updatedAt truncate, docId hard-400); escaping unchanged (shared body builders `escapeMarkdown`/`TruncateRunes`); byte-equivalence baselines for docs-display and summary intact (`TestBuildSummaryCards_MigrationBaseline` green); `NotifyReq` shape untouched; no migration.

## Verdict
**APPROVE** (re-affirmed at `5fa6bd1c`, superseding my dismissed prior reviews) — the new commit cleanly resolves both Jerry-Xin re-review points: over-max counts now clamp to the schema `maximum` (byte-verified equal: 1M / 1B) so an over-large count renders capped instead of a C1 400 zero-delivery, symmetric with the display-field truncation and test-locked to the schema; and `taskNo` is passed verbatim to restore byte-parity with the legacy raw-value deep-link (blank taskNo still rejected at ingress). Structural violations and over-length deep-link keys keep C1 400; escaping and byte-equivalence intact. Both gates pass.
Suggested decision: `Safe to merge`. `merge: MERGEABLE`; `BLOCKED` is branch-protection / required approvals.

Jerry-Xin (approved):
This PR is relevant to octo-server and safely migrates summary notification cards to the existing Registry architecture.

💬 Non-blocking

- 🟡 Warning — The cap-alignment tests only prove local limits are not greater than schema limits; they do not detect schemas being increased independently. Add explicit schema-value assertions like the existing docs template tests for title, reason, time range, generated time, and count maxima. See `modules/notify/card_via_registry.go:278` and `modules/notify/card_via_registry_summary_baseline_test.go:21`.
- 🔵 Suggestion — `main.go:726` says hard gate ② is achieved, while the updated documentation correctly says production canary validation remains outstanding. Adjust the comment to avoid contradictory operational guidance.
- 🔵 Suggestion — Extend the byte-equivalence baseline with boundary and whitespace cases. `mapSummaryCardFieldsToJSON` trims titles and caps counts (`modules/notify/card_via_registry.go:338`), so equivalence intentionally does not hold for every legacy-accepted input.

✅ Highlights

- Correct fail-closed registry wiring and preflight error classification.
- Render profiles are propagated through dispatch.
- Shared rendering helpers preserve normal-input wire compatibility.
- Strong focused coverage for both templates, metadata, localization, schema rejection, and fallback sanitization.
- Focused template test suites pass.

Issue #122, opened 2026-05-21
fix(obo): NULL persona_prompt scan failure + scopeRowExists fail-open


yujiawei:
## 问题

PR#121 review (Jerry-Xin) 指出两个 blocking issue：

### 1. NULL persona_prompt scan 失败
`insertGrant` 没填 `persona_prompt`，新 grant 的 persona_prompt 为 NULL。`SELECT *` 扫到 NULL 时 Go `string` 类型 scan 报错。

修法：`insertGrant` 时填默认空字符串 + 所有 grant read 用 `COALESCE(persona_prompt, '')`。

### 2. scopeRowExists 错误被忽略（fail-open）
`obo_check.go:89` 的 `scopeRowExists` error 被 `_` 吞了。DB 出错时误判为无 scope row，可能绕过 disabled scope。

修法：error 时 return err（fail-closed）+ 加回归测试。

## 关联
- octo-server PR#121
- 分支：feat/obo-implicit-scope

lml2468:
Both issues were already fixed in PR#121 (merged 2026-05-22):

1. **NULL persona_prompt scan failure** — all `SELECT *` paths replaced with `oboGrantColumns` using `COALESCE(persona_prompt, '') AS persona_prompt`; `insertGrant` now writes an explicit empty string instead of letting the column default to NULL.

2. **scopeRowExists fail-open** — error is now propagated (`return scopeExistErr`) instead of swallowed with `_`. Comment in `obo_check.go` explicitly references GH#122: 'scopeRowExists is fail-closed'.

Verified by reading current main branch code. Closing as already resolved.

PR #653, opened 2026-07-23
fix(space): support permanent invite (NULL expires_at) in createInvite and updateInvite

xiaokenbot:
## Summary

Fixes the report: **"selecting permanent still expires after 72 h"**.

## Root Cause

The admin UI (`octo-admin / SpaceInvitesPanel`) has a 7d/14d/30d/**never** expiry picker. When the user picks **never**, the frontend previously sent an empty `expires_at` string. The server treated empty string as "not provided" and applied the env-var default TTL (72 h). There was also no path to *clear* an existing `expires_at` back to `NULL`.

### `parseInviteExpiresAt` — three-state result

| Input | Result | Meaning |
|---|---|---|
| `nil` / `""` | `notProvided=true` | Field not sent; server applies default TTL |
| `"never"` | `notProvided=false, expiresAt=nil` | Explicit permanent — write `NULL` |
| timestamp string | `notProvided=false, expiresAt=&t` | Specific expiry time |

### `manager.createInvite`
- Explicit `expires_at` (including `"never"`) **bypasses** the env-var default branch.
- `"never"` → `model.ExpiresAt = nil` → DB row gets `NULL`.

### `manager.updateInvite` + `Space.updateInvite`
- Derive `clearExpiresAt` flag from three-state result.
- Pass to `updateInvitationAdmin`.

### `updateInvitationAdmin` (DB layer)
- New `clearExpiresAt bool` parameter.
- When `true`: `SET expires_at = NULL` (true permanent).
- When `false` and `expiresAt != nil`: write the specific time.
- When both zero: column unchanged (existing behaviour).

### Email-invite paths
- Adapted to new return type; semantics unchanged (no "never" UX there).

## Companion PR

octo-admin fix: `xiaokenbot:fix/space-invite-permanent-expiry`

## Testing

- `go build ./modules/space/` passes clean.
- Manual: POST `{"expires_at": "never"}` → DB `expires_at IS NULL`; POST `{}` → DB gets default TTL; PUT `{"expires_at": "never"}` → existing row updated to `NULL`.

### DIFF
--- a/modules/space/api.go
+++ b/modules/space/api.go
@@ -1497,16 +1497,21 @@ func (s *Space) updateInvite(c *wkhttp.Context) {
 	}
 
 	// 解析过期时间：与管理端 parseInviteExpiresAt 共用 time.Local 约定，避免双路径写库时区漂移。
-	expiresAt, err := parseInviteExpiresAt(req.ExpiresAt)
+	expiresResult, err := parseInviteExpiresAt(req.ExpiresAt)
 	if err != nil {
 		respondSpaceRequestInvalid(c, "expires_at")
 		return
+	}
+	clearExpires := !expiresResult.notProvided && expiresResult.expiresAt == nil
+	var updateExpiresAt *time.Time
+	if !expiresResult.notProvided {
+		updateExpiresAt = expiresResult.expiresAt
 	}
 
 	// 复用管理端的 updateInvitationAdmin：
 	//   - WHERE 无 status=1 限制，可对已禁用邀请码重启用（status=1）
 	//   - 按 space_id + invite_code 双匹配，天然防跨空间越权
-	affected, err := s.mdb.updateInvitationAdmin(spaceId, code, req.MaxUses, expiresAt, req.Status)
+	affected, err := s.mdb.updateInvitationAdmin(spaceId, code, req.MaxUses, updateExpiresAt, clearExpires, req.Status)
 	if err != nil {
 		httperr.ResponseErrorL(c, errcode.ErrSpaceStoreFailed, nil, nil)
 		return
--- a/modules/space/api.go
+++ b/modules/space/api.go
@@ -1502,11 +1502,7 @@ func (s *Space) updateInvite(c *wkhttp.Context) {
 		respondSpaceRequestInvalid(c, "expires_at")
 		return
 	}
-	clearExpires := !expiresResult.notProvided && expiresResult.expiresAt == nil
-	var updateExpiresAt *time.Time
-	if !expiresResult.notProvided {
-		updateExpiresAt = expiresResult.expiresAt
-	}
+	updateExpiresAt, clearExpires := expiresResult.updateArgs()
 
 	// 复用管理端的 updateInvitationAdmin：
 	//   - WHERE 无 status=1 限制，可对已禁用邀请码重启用（status=1）
--- a/modules/space/api_email_invite.go
+++ b/modules/space/api_email_invite.go
@@ -44,7 +44,8 @@ func (s *Space) createMemberEmailInvite(c *wkhttp.Context) {
 		respondSpaceRequestInvalid(c, "")
 		return
 	}
-	expiresAt, err := parseInviteExpiresAt(req.ExpiresAt)
+	expiresResult, err := parseInviteExpiresAt(req.ExpiresAt)
+	expiresAt := expiresResult.expiresAt // email invite 不区分"未传"与"永久"，直接取 *time.Time
 	if err != nil {
 		respondSpaceRequestInvalid(c, "expires_at")
 		return
--- a/modules/space/api_email_invite.go
+++ b/modules/space/api_email_invite.go
@@ -45,11 +45,16 @@ func (s *Space) createMemberEmailInvite(c *wkhttp.Context) {
 		return
 	}
 	expiresResult, err := parseInviteExpiresAt(req.ExpiresAt)
-	expiresAt := expiresResult.expiresAt // email invite 不区分"未传"与"永久"，直接取 *time.Time
 	if err != nil {
 		respondSpaceRequestInvalid(c, "expires_at")
 		return
 	}
+	// email invite 不提供"永久"选项；若客户端显式传 "never"，拒绝请求（恢复旧 400 行为）。
+	if !expiresResult.notProvided && expiresResult.expiresAt == nil {
+		respondSpaceRequestInvalid(c, "expires_at")
+		return
+	}
+	expiresAt := expiresResult.expiresAt
 
 	rawToken, tokenHash, err := generateEmailInviteToken()
 	if err != nil {
--- a/modules/space/api_manager.go
+++ b/modules/space/api_manager.go
@@ -896,7 +896,7 @@ func (m *Manager) createInvite(c *wkhttp.Context) {
 			return
 		}
 	}
-	expiresAt, err := parseInviteExpiresAt(req.ExpiresAt)
+	expiresResult, err := parseInviteExpiresAt(req.ExpiresAt)
 	if err != nil {
 		respondSpaceRequestInvalid(c, "expires_at")
 		return
@@ -920,10 +920,15 @@ func (m *Manager) createInvite(c *wkhttp.Context) {
 	} else {
 		model.MaxUses = defMaxUses
 	}
-	if expiresAt != nil {
-		t := db.Time(*expiresAt)
-		model.ExpiresAt = &t
+	if !expiresResult.notProvided {
+		// 客户端显式传了 expires_at（含 "never" 永久）：直接使用，不套默认。
+		if expiresResult.expiresAt != nil {
+			t := db.Time(*expiresResult.expiresAt)
+			model.ExpiresAt = &t
+		}
+		// expiresResult.expiresAt == nil 即永久：model.ExpiresAt 保持 nil，落库为 NULL。
 	} else if defExpiresAt != nil {
+		// 未传 expires_at：套环境变量默认值。
 		t := db.Time(*defExpiresAt)
 		model.ExpiresAt = &t
 	}
@@ -987,13 +992,22 @@ func (m *Manager) updateInvite(c *wkhttp.Context) {
 		respondSpaceRequestInvalid(c, "status")
 		return
 	}
-	expiresAt, err := parseInviteExpiresAt(req.ExpiresAt)
+	expiresResult, err := parseInviteExpiresAt(req.ExpiresAt)
 	if err != nil {
 		respondSpaceRequestInvalid(c, "expires_at")
 		return
 	}
+	// updateInvitationAdmin 三态：
+	//   notProvided=true  → updateExpiresAt=nil, clearExpiresAt=false → 字段不变
+	//   notProvided=false, expiresAt!=nil → 写入具体时间
+	//   notProvided=false, expiresAt==nil → clearExpiresAt=true → 写 NULL（永久）
+	clearExpires := !expiresResult.notProvided && expiresResult.expiresAt == nil
+	var updateExpiresAt *time.Time
+	if !expiresResult.notProvided {
+		updateExpiresAt = expiresResult.expiresAt
+	}
 
-	affected, err := m.managerDB.updateInvitationAdmin(spaceId, code, req.MaxUses, expiresAt, req.Status)
+	affected, err := m.managerDB.updateInvitationAdmin(spaceId, code, req.MaxUses, updateExpiresAt, clearExpires, req.Status)
 	if err != nil {
 		m.Error("修改邀请码失败", zap.Error(err), zap.String("spaceId", spaceId), zap.String("code", code))
 		httperr.ResponseErrorL(c, errcode.ErrSpaceStoreFailed, nil, nil)
@@ -1016,18 +1030,35 @@ func (m *Manager) updateInvite(c *wkhttp.Context) {
 // 用户侧 updateInvite 与管理端 create/updateInvite 共用此常量，避免双写路径漂移。
 const inviteTimeLayout = "2006-01-02 15:04:05"
 
-// parseInviteExpiresAt 解析 expires_at 字符串，空字符串视为未传。
+// inviteExpiresAtNeverSentinel 客户端传此值表示「显式永久」。
+// 空字符串 / nil 表示「未传，由服务端套默认值」；"never" 表示「明确不要过期时间」。
+const inviteExpiresAtNeverSentinel = "never"
+
+// inviteExpiresResult 解析结果，三态：
+//   - notProvided=true : 客户端未传 expires_at（nil 或空字符串），走默认值
+//   - notProvided=false, expiresAt=nil : 客户端显式要求永久（"never"）
+//   - notProvided=false, expiresAt!=nil : 客户端给定具体时间
+type inviteExpiresResult struct {
+	notProvided bool
+	expiresAt   *time.Time
+}
+
+// parseInviteExpiresAt 解析 expires_at 字符串，三态处理。
 // 时区采用服务器 time.Local——管理端与用户侧统一共用本函数。
 // 部署环境应显式设置 TZ，确保客户端发送的"服务器本地时间"解释一致。
-func parseInviteExpiresAt(raw *string) (*time.Time, error) {
+func parseInviteExpiresAt(raw *string) (inviteExpiresResult, error) {
 	if raw == nil || *raw == "" {
-		return nil, nil
+		return inviteExpiresResult{notProvided: true}, nil
+	}
+	if *raw == inviteExpiresAtNeverSentinel {
+		// 显式永久：expiresAt = nil，notProvided = false
+		return inviteExpiresResult{notProvided: false, expiresAt: nil}, nil
 	}
 	t, err := time.ParseInLocation(inviteTimeLayout, *raw, time.Local)
 	if err != nil {
-		return nil, errors.New("过期时间格式错误，请使用 2006-01-02 15:04:05 格式")
+		return inviteExpiresResult{}, errors.New(`过期时间格式错误，请使用 2006-01-02 15:04:05 格式或 "never" 表示永久`)
 	}
-	return &t, nil
+	return inviteExpiresResult{notProvided: false, expiresAt: &t}, nil
 }
 
 // disableInvite 禁用邀请码
--- a/modules/space/api_manager.go
+++ b/modules/space/api_manager.go
@@ -829,7 +829,7 @@ func (m *Manager) listInvites(c *wkhttp.Context) {
 	}
 	resp := make([]*managerInviteResp, 0, len(list))
 	for _, inv := range list {
-		expiresAt := ""
+		expiresAt := inviteExpiresAtNeverSentinel // 永久邀请
 		if inv.ExpiresAt != nil {
 			expiresAt = inv.ExpiresAt.String()
 		}
@@ -946,9 +946,8 @@ func (m *Manager) createInvite(c *wkhttp.Context) {
 		zap.String("operator", operator),
 	)
 
-	// 显式 Format 保证响应格式与 parseInviteExpiresAt 接受的输入格式一致，
-	// 避免客户端拿到响应后回传被 parse 拒绝。
-	expiresStr := ""
+	// 序列化 expires_at：永久（NULL）返回 "never" 哨兵，与输入格式对称，客户端可安全回传。
+	expiresStr := inviteExpiresAtNeverSentinel // 永久邀请
 	if model.ExpiresAt != nil {
 		expiresStr = time.Time(*model.ExpiresAt).Format(inviteTimeLayout)
 	}
@@ -997,16 +996,7 @@ func (m *Manager) updateInvite(c *wkhttp.Context) {
 		respondSpaceRequestInvalid(c, "expires_at")
 		return
 	}
-	// updateInvitationAdmin 三态：
-	//   notProvided=true  → updateExpiresAt=nil, clearExpiresAt=false → 字段不变
-	//   notProvided=false, expiresAt!=nil → 写入具体时间
-	//   notProvided=false, expiresAt==nil → clearExpiresAt=true → 写 NULL（永久）
-	clearExpires := !expiresResult.notProvided && expiresResult.expiresAt == nil
-	var updateExpiresAt *time.Time
-	if !expiresResult.notProvided {
-		updateExpiresAt = expiresResult.expiresAt
-	}
-
+	updateExpiresAt, clearExpires := expiresResult.updateArgs()
 	affected, err := m.managerDB.updateInvitationAdmin(spaceId, code, req.MaxUses, updateExpiresAt, clearExpires, req.Status)
 	if err != nil {
 		m.Error("修改邀请码失败", zap.Error(err), zap.String("spaceId", spaceId), zap.String("code", code))
@@ -1043,6 +1033,16 @@ type inviteExpiresResult struct {
 	expiresAt   *time.Time
 }
 
+// updateArgs 从三态解析结果中派生 updateInvitationAdmin 所需参数，
+// 避免 Manager.updateInvite 与 Space.updateInvite 各自重复同一段逻辑导致实现漂移。
+func (r inviteExpiresResult) updateArgs() (expiresAt *time.Time, clearExpiresAt bool) {
+	clearExpiresAt = !r.notProvided && r.expiresAt == nil
+	if !r.notProvided {
+		expiresAt = r.expiresAt
+	}
+	return
+}
+
 // parseInviteExpiresAt 解析 expires_at 字符串，三态处理。
 // 时区采用服务器 time.Local——管理端与用户侧统一共用本函数。
 // 部署环境应显式设置 TZ，确保客户端发送的"服务器本地时间"解释一致。
@@ -1050,7 +1050,7 @@ func parseInviteExpiresAt(raw *string) (inviteExpiresResult, error) {
 	if raw == nil || *raw == "" {
 		return inviteExpiresResult{notProvided: true}, nil
 	}
-	if *raw == inviteExpiresAtNeverSentinel {
+	if strings.ToLower(strings.TrimSpace(*raw)) == inviteExpiresAtNeverSentinel {
 		// 显式永久：expiresAt = nil，notProvided = false
 		return inviteExpiresResult{notProvided: false, expiresAt: nil}, nil
 	}
--- a/modules/space/api_manager_email_invite.go
+++ b/modules/space/api_manager_email_invite.go
@@ -59,11 +59,16 @@ func (m *Manager) createSpaceOwnerEmailInvite(c *wkhttp.Context) {
 		return
 	}
 	expiresResult, err := parseInviteExpiresAt(req.ExpiresAt)
-	expiresAt := expiresResult.expiresAt // email invite 不区分"未传"与"永久"，直接取 *time.Time
 	if err != nil {
 		respondSpaceRequestInvalid(c, "expires_at")
 		return
 	}
+	// email invite 不提供"永久"选项；若客户端显式传 "never"，拒绝请求（恢复旧 400 行为）。
+	if !expiresResult.notProvided && expiresResult.expiresAt == nil {
+		respondSpaceRequestInvalid(c, "expires_at")
+		return
+	}
+	expiresAt := expiresResult.expiresAt
 
 	rawToken, tokenHash, err := generateEmailInviteToken()
 	if err != nil {
--- a/modules/space/api_manager_email_invite.go
+++ b/modules/space/api_manager_email_invite.go
@@ -58,7 +58,8 @@ func (m *Manager) createSpaceOwnerEmailInvite(c *wkhttp.Context) {
 		respondSpaceRequestInvalid(c, "")
 		return
 	}
-	expiresAt, err := parseInviteExpiresAt(req.ExpiresAt)
+	expiresResult, err := parseInviteExpiresAt(req.ExpiresAt)
+	expiresAt := expiresResult.expiresAt // email invite 不区分"未传"与"永久"，直接取 *time.Time
 	if err != nil {
 		respondSpaceRequestInvalid(c, "expires_at")
 		return
--- a/modules/space/api_manager_test.go
+++ b/modules/space/api_manager_test.go
@@ -1784,3 +1784,135 @@ func TestManager_UpdateMemberRoleIdempotent(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, 2, owner.Role)
 }
+
+// ──────────────────────────────────────────────────────────────────────────────
+// parseInviteExpiresAt 三态单元测试（review #653 P2 回归）
+// ──────────────────────────────────────────────────────────────────────────────
+
+func TestParseInviteExpiresAt_ThreeState(t *testing.T) {
+	t.Run("nil → notProvided", func(t *testing.T) {
+		r, err := parseInviteExpiresAt(nil)
+		assert.NoError(t, err)
+		assert.True(t, r.notProvided)
+		assert.Nil(t, r.expiresAt)
+	})
+	t.Run("empty string → notProvided", func(t *testing.T) {
+		s := ""
+		r, err := parseInviteExpiresAt(&s)
+		assert.NoError(t, err)
+		assert.True(t, r.notProvided)
+		assert.Nil(t, r.expiresAt)
+	})
+	t.Run(`"never" → permanent`, func(t *testing.T) {
+		s := "never"
+		r, err := parseInviteExpiresAt(&s)
+		assert.NoError(t, err)
+		assert.False(t, r.notProvided)
+		assert.Nil(t, r.expiresAt, "explicit permanent: expiresAt must be nil")
+	})
+	t.Run(`"Never" (mixed-case) → permanent`, func(t *testing.T) {
+		s := "Never"
+		r, err := parseInviteExpiresAt(&s)
+		assert.NoError(t, err)
+		assert.False(t, r.notProvided)
+		assert.Nil(t, r.expiresAt)
+	})
+	t.Run(`" never " (padded) → permanent`, func(t *testing.T) {
+		s := " never "
+		r, err := parseInviteExpiresAt(&s)
+		assert.NoError(t, err)
+		assert.False(t, r.notProvided)
+		assert.Nil(t, r.expiresAt)
+	})
+	t.Run("valid timestamp → specific time", func(t *testing.T) {
+		s := "2030-06-15 12:00:00"
+		r, err := parseInviteExpiresAt(&s)
+		assert.NoError(t, err)
+		assert.False(t, r.notProvided)
+		assert.NotNil(t, r.expiresAt)
+	})
+	t.Run("invalid string → error", func(t *testing.T) {
+		s := "not-a-date"
+		_, err := parseInviteExpiresAt(&s)
+		assert.Error(t, err)
+	})
+}
+
+// ──────────────────────────────────────────────────────────────────────────────
+// Manager.createInvite "never" → DB NULL 集成测试（review #653 P2 回归）
+// ──────────────────────────────────────────────────────────────────────────────
+
+func TestManager_CreateInvite_Never_DBNull(t *testing.T) {
+	s, _, err := setup(t)
+	assert.NoError(t, err)
+	token := adminToken(t)
+
+	seedSpace(t, "mgr-inv-never1", "inv never create", "u-owner", SpaceStatusNormal)
+
+	body := `{"expires_at":"never"}`
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("POST", "/v1/manager/spaces/mgr-inv-never1/invites", bytes.NewBufferString(body))
+	req.Header.Set("token", token)
+	req.Header.Set("Content-Type", "application/json")
+	s.GetRoute().ServeHTTP(w, req)
+	assert.Equal(t, http.StatusOK, w.Code, "create with never should succeed: %s", w.Body.String())
+
+	var resp struct {
+		InviteCode string `json:"invite_code"`
+		ExpiresAt  string `json:"expires_at"`
+	}
+	assert.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
+	assert.Equal(t, "never", resp.ExpiresAt, `response expires_at must be "never" for permanent invite`)
+
+	inv, err := testSpaceDB.queryInvitationByCode(resp.InviteCode)
+	assert.NoError(t, err)
+	assert.NotNil(t, inv)
+	assert.Nil(t, inv.ExpiresAt, "DB expires_at must be NULL for permanent invite")
+}
+
+// ──────────────────────────────────────────────────────────────────────────────
+// Manager.updateInvite "never" → DB NULL 集成测试（review #653 P2 回归）
+// ──────────────────────────────────────────────────────────────────────────────
+
+func TestManager_UpdateInvite_ClearExpiry_Never(t *testing.T) {
+	s, _, err := setup(t)
+	assert.NoError(t, err)
+	token := adminToken(t)
+
+	seedSpace(t, "mgr-inv-never2", "inv never update", "u-owner", SpaceStatusNormal)
+
+	// 先建一个有具体过期时间的邀请码
+	bodyCreate := `{"expires_at":"2030-01-01 00:00:00","max_uses":10}`
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("POST", "/v1/manager/spaces/mgr-inv-never2/invites", bytes.NewBufferString(bodyCreate))
+	req.Header.Set("token", token)
+	req.Header.Set("Content-Type", "application/json")
+	s.GetRoute().ServeHTTP(w, req)
+	assert.Equal(t, http.StatusOK, w.Code)
+
+	var createResp struct {
+		InviteCode string `json:"invite_code"`
+	}
+	assert.NoError(t, json.Unmarshal(w.Body.Bytes(), &createResp))
+	code := createResp.InviteCode
+	assert.NotEmpty(t, code)
+
+	// 验证创建后 expires_at 非 NULL
+	inv, err := testSpaceDB.queryInvitationByCode(code)
+	assert.NoError(t, err)
+	assert.NotNil(t, inv.ExpiresAt, "before update: should have a concrete expiry")
+
+	// 用 "never" 更新为永久
+	bodyUpdate := `{"expires_at":"never"}`
+	w2 := httptest.NewRecorder()
+	req2, _ := http.NewRequest("PUT", "/v1/manager/spaces/mgr-inv-never2/invites/"+code, bytes.NewBufferString(bodyUpdate))
+	req2.Header.Set("token", token)
+	req2.Header.Set("Content-Type", "application/json")
+	s.GetRoute().ServeHTTP(w2, req2)
+	assert.Equal(t, http.StatusOK, w2.Code, "update to never should succeed: %s", w2.Body.String())
+
+	// 验证 DB expires_at 已被清为 NULL
+	inv2, err := testSpaceDB.queryInvitationByCode(code)
+	assert.NoError(t, err)
+	assert.Nil(t, inv2.ExpiresAt, "after update with never: DB expires_at must be NULL")
+}
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -560,14 +560,21 @@ func (d *managerDB) disableInvitation(spaceId, code string) (int64, error) {
 // 有意设计：WHERE 不限制 status，管理员可以对已禁用（status=0）的邀请码执行 PUT，
 // 包括通过 {"status": 1} 重新启用——这是管理操作的必要能力（如误禁恢复）。
 // 若要禁止重新启用，应在 API 层决策，不在此函数加 AND status=1。
-func (d *managerDB) updateInvitationAdmin(spaceId, code string, maxUses *int, expiresAt *time.Time, status *int) (int64, error) {
+// updateInvitationAdmin 更新邀请码字段。
+// clearExpiresAt=true 时将 expires_at 置为 NULL（永久），优先级高于 expiresAt。
+// expiresAt!=nil 时写入具体时间；两者均为零值时该列保持不变。
+func (d *managerDB) updateInvitationAdmin(spaceId, code string, maxUses *int, expiresAt *time.Time, clearExpiresAt bool, status *int) (int64, error) {
 	builder := d.session.Update("space_invitation")
 	changed := false
 	if maxUses != nil {
 		builder = builder.Set("max_uses", *maxUses)
 		changed = true
 	}
-	if expiresAt != nil {
+	if clearExpiresAt {
+		// 显式永久：写 NULL
+		builder = builder.Set("expires_at", nil)
+		changed = true
+	} else if expiresAt != nil {
 		builder = builder.Set("expires_at", *expiresAt)
 		changed = true
 	}
--- a/modules/space/db_manager.go
+++ b/modules/space/db_manager.go
@@ -554,15 +554,16 @@ func (d *managerDB) disableInvitation(spaceId, code string) (int64, error) {
 	return result.RowsAffected()
 }
 
-// updateInvitationAdmin 管理端可修改 max_uses / expires_at / status，nil 字段不变更。
-// 返回 affected rows，0 表示记录不存在。
+// updateInvitationAdmin 管理端更新邀请码字段。返回 affected rows，0 表示记录不存在。
 //
-// 有意设计：WHERE 不限制 status，管理员可以对已禁用（status=0）的邀请码执行 PUT，
-// 包括通过 {"status": 1} 重新启用——这是管理操作的必要能力（如误禁恢复）。
+// expires_at 三态行为：
+//   - clearExpiresAt=true               → SET expires_at = NULL（永久），优先级最高
+//   - clearExpiresAt=false, expiresAt!=nil → 写入具体时间
+//   - clearExpiresAt=false, expiresAt==nil → 列保持不变
+//
+// 有意设计：WHERE 不限制 status，管理员可对已禁用（status=0）的邀请码执行 PUT，
+// 包括通过 {"status": 1} 重新启用——管理操作的必要能力（如误禁恢复）。
 // 若要禁止重新启用，应在 API 层决策，不在此函数加 AND status=1。
-// updateInvitationAdmin 更新邀请码字段。
-// clearExpiresAt=true 时将 expires_at 置为 NULL（永久），优先级高于 expiresAt。
-// expiresAt!=nil 时写入具体时间；两者均为零值时该列保持不变。
 func (d *managerDB) updateInvitationAdmin(spaceId, code string, maxUses *int, expiresAt *time.Time, clearExpiresAt bool, status *int) (int64, error) {
 	builder := d.session.Update("space_invitation")
 	changed := false


Jerry-Xin (dismissed):
Correctly implements permanent invites (NULL `expires_at`) with clean three-state expiry semantics across create and update paths.

**Verified — NULL semantics are consistent end to end**

- Write path uses the dbr query builder: a permanent invite writes a real SQL `NULL`. On the update path, `updateInvitationAdmin` sets the column via an explicit `clearExpiresAt` flag → `.Set("expires_at", nil)` (`modules/space/db_manager.go:573`), not a struct/zero-value update, so clearing an expiry back to NULL actually persists (no silent field-skip).
- Model field `InvitationModel.ExpiresAt` is `*db.Time` (`modules/space/model.go`): nil ⇒ NULL. All readers guard `ExpiresAt != nil` before dereferencing (`modules/space/api.go:1169,1229,1364`, `modules/space/api_email_invite_public.go:183,358`) — no nil-deref risk.
- Every expired-filter query treats NULL as never-expires: `(expires_at IS NULL OR expires_at > ?)` in `modules/space/db.go:285` (accept/lookup), `db.go:304` (atomic consume), `db.go:418` (active filter), and `modules/space/db_email_invite.go:178`. NULL is correctly excluded from the "expired" set.
- Schema already allows NULL (`modules/space/sql/20260424000002_space_legacy01.sql`: `expires_at TIMESTAMP NULL DEFAULT NULL`), so no migration is required and a NULL write will not error.
- Request DTO uses `*string` and the new three-state result (`notProvided` / `expiresAt`) so "omitted" (apply default TTL), "never" (permanent), and an explicit timestamp are all distinguished. Backward compat preserved: omitted ⇒ default TTL, real timestamps unchanged.
- `go build ./modules/space/` passes clean.

🟡 Non-blocking

- No automated regression tests cover the new "never" behaviour. Suggest adding: (a) create with `"never"` → row reads as non-expired, (b) update an existing expiry to `"never"` → NULL persists, (c) omitted field still applies default TTL. Logic under test: `modules/space/api_manager.go` createInvite/updateInvite and `modules/space/db_manager.go:573`.
- Email-invite paths intentionally collapse "not provided" and "never" to a nil `*time.Time` (`modules/space/api_email_invite.go:48`, `modules/space/api_manager_email_invite.go`). This matches prior behaviour (email path never applied a default TTL), so no regression — worth a one-line comment noting `"never"` is silently accepted there too.

✅ Highlights

- Explicit `"never"` sentinel cleanly bypasses the env-var default-TTL branch instead of overloading empty string.
- Update always also sets `updated_at`, so an already-NULL clear still reports rows affected — the `affected == 0` branch correctly means "invite not found", not "no-op".
- DB update stays parameterized and scoped by both `space_id` and `invite_code`; permission/space-isolation behaviour preserved.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #653 (octo-server)

## Summary

This PR adds support for permanent (never-expiring) invite codes by introducing a three-state `inviteExpiresResult` struct to `parseInviteExpiresAt` and a `"never"` sentinel value. Previously the function returned `*time.Time` with no way to distinguish "client didn't send expires_at" from "client explicitly wants no expiry". The change touches the manager create/update invite paths, the user-side update path, the DB update function, and two email-invite callers. The implementation is clean and the data flow is correct end-to-end.

## Verification

- ✅ **Three-state mapping in createInvite** (`modules/space/api_manager.go:920-934`) — `notProvided=true` falls through to `defExpiresAt`; `notProvided=false, expiresAt=nil` leaves `model.ExpiresAt` nil (NULL in DB); concrete time is written. All three branches verified against the diff.
- ✅ **Three-state mapping in updateInvite** — Both manager (`api_manager.go:1003-1008`) and user-side (`api.go:1296-1300`) derive `clearExpires` and `updateExpiresAt` with identical logic. Symmetric paths are consistent.
- ✅ **DB layer priority ordering** (`db_manager.go:573-579`) — `clearExpiresAt` takes precedence over `expiresAt != nil`, and both take precedence over no-change. The `changed` flag is set in each branch.
- ✅ **All callers updated** — `updateInvitationAdmin` has exactly 2 callers (api.go, api_manager.go), both pass the new `clearExpiresAt` parameter. `parseInviteExpiresAt` has 5 callers, all adapted.
- ✅ **NULL consumption already safe** — `queryInvitationByCode` (`db.go:286`) uses `expires_at IS NULL OR expires_at > ?`, so permanent invites (NULL) are treated as never-expired. `incrementInviteUsedCountAtomic` (`db.go:303`) uses the same guard.
- ✅ **Email invite collapse** (`api_email_invite.go:48`, `api_manager_email_invite.go:62`) — Both email paths collapse the 3-state to `*time.Time` with `expiresAt := expiresResult.expiresAt`. When "never" is sent, `expiresAt` becomes nil, producing NULL in DB — semantically correct for permanent email invites.
- ✅ **Response serialization** — `createInvite` response uses empty string for nil `ExpiresAt` (`api_manager.go:951-953`); `toEmailInviteResp` (`api_manager_email_invite.go:225-227`) conditionally sets `ExpiresAt` only when non-nil. Round-trip is consistent.

Static analysis only at head `cdf964c1`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Two P2 items below.

### P2 — Stale doc comment stacked above the new one (`modules/space/db_manager.go:557`)

The old comment block (lines 557–562) still reads "nil 字段不变更" but the new `clearExpiresAt` parameter means nil `expiresAt` no longer universally implies "don't change" — `clearExpiresAt=true` explicitly writes NULL. The new comment at lines 563–565 supersedes this, but both blocks are stacked, creating a contradictory doc. The "有意设计" paragraph about WHERE not restricting status is valuable context worth preserving — merge it into the new block and delete the old one.

### P2 — No regression test covers the `"never"` sentinel path (`modules/space/api_manager.go:1049`)

The PR changes `parseInviteExpiresAt`'s return type and introduces a new input sentinel, but no test exercises the "never" → NULL path end-to-end. Existing tests go through HTTP endpoints (not direct function calls), so they won't break from the signature change, but the new permanent-invite behavior is entirely untested. Suggested additions: (1) createInvite with `"expires_at": "never"` → assert `ExpiresAt` is nil in DB; (2) updateInvite with `"never"` → assert `expires_at` column is NULL; (3) updateInvite omitting `expires_at` → assert column unchanged.

## Data Flow Trace

| Consumer | Upstream Source | Runtime Data Reaches Consumer? |
|---|---|---|
| `model.ExpiresAt` in createInvite | `parseInviteExpiresAt` → `inviteExpiresResult.expiresAt` | ✅ Yes — nil branch correctly skips assignment, leaving NULL |
| `clearExpires` in updateInvite | `!notProvided && expiresAt==nil` | ✅ Yes — correctly true only for "never" input |
| `updateExpiresAt` in updateInvite | `expiresResult.expiresAt` when `!notProvided` | ✅ Yes — nil when "never", concrete time otherwise, nil when notProvided |
| `builder.Set("expires_at", nil)` in DB | `clearExpiresAt` param from API layer | ✅ Yes — priority over `expiresAt` param ensures NULL write |
| `expiresAt` in email invite | `expiresResult.expiresAt` (collapsed) | ✅ Yes — nil for both "never" and "not provided", both producing NULL in DB |
| `queryInvitationByCode` consumer | `expires_at` column in DB | ✅ Yes — `IS NULL OR > NOW()` handles permanent invites |

## Blind Spot Checklist

- **C1 (Dual-path parity)** — ✅ Clear. Manager createInvite vs user-side createInvite: different paths by design (manager accepts input, user-side uses defaults). Manager updateInvite vs user-side updateInvite: identical 3-state logic. Email invite member vs owner: identical collapse pattern.
- **C2 (Control-flow ordering)** — ✅ Clear. `parseInviteExpiresAt` is called once per handler; no nested reuse or double invocation.
- **C3 (Auth boundary)** — N/A. No auth/permission changes in this PR.
- **C4 (Auth lifecycle)** — N/A. No auth changes.
- **C5 (Build/runtime)** — N/A. No build/packaging changes.
- **C6 (Governance/docs)** — N/A. No governance/policy changes.

## Cross-Round Blocker Recheck (R6)

N/A — first review round.

- Edge cases: JSON null, empty string, "never", valid date, and garbage input all produce correct outcomes across all 5 callers.
- The `inviteTimeLayout` constant is shared between parse and response format, preventing round-trip mismatches.
- The `changed` guard in `updateInvitationAdmin` correctly returns early when no fields are modified.
- The `insertInvitationWithRetry` function passes the model as-is to the DB layer, preserving nil `ExpiresAt`.

## Verdict: APPROVED

No correctness, security, or data-integrity issues. The three-state design is sound, all callers are updated, and the consumption side already handles NULL. The two P2 items (stale doc comment and missing test coverage) are non-blocking but worth addressing.

cnwenf (changes requested):
# Code Review — PR #653 (Mininglamp-OSS/octo-server)

> Consolidated from three independent reviews (octo-review-qwen3.7, octo-review-glm5.2, octo-review-qwen3.8). Findings are deduplicated; `[source: ...]` marks which reviewer(s) raised each finding. The three full raw reviews remain as comments on this issue for audit.

## Summary

This PR fixes a real user-facing bug — selecting "permanent" for a space invite still applied the 72 h default TTL — by making `parseInviteExpiresAt` three-state (`notProvided` / explicit `"never"` → `NULL` / specific timestamp) and threading a `clearExpiresAt` flag through `updateInvitationAdmin`. All three reviewers independently verified at head `cdf964c1`, via isolated per-diff E2E stacks, that the core mechanism is correct (NULL write, clear-to-NULL precedence, default-TTL bypass, and `NULL` treated as permanently valid on all read paths) and that the security surface is clean. The consolidated verdict is NEEDS_CHANGES: the `"never"` sentinel now silently flows into the email-invite endpoints (contradicting the PR description's "semantics unchanged" claim), and the two load-bearing new paths (create-`"never"`→NULL, update-clearExpires→NULL) ship with zero automated regression coverage — mutations reverting either path were confirmed to leave CI green. Eight further P3 items are non-blocking robustness / DRY / observability nits.

## Findings

### P2 — `"never"` sentinel silently leaks into the email-invite paths (modules/space/api_email_invite.go:47, modules/space/api_manager_email_invite.go:61) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8] (severity split: qwen3.7=P2, glm5.2=P2, qwen3.8=P3)

**octo-review-qwen3.7:**

The PR description states "Email-invite paths — Adapted to new return type; semantics unchanged (no 'never' UX there)." However, the old `parseInviteExpiresAt` returned an error for non-timestamp strings like `"never"`, so sending `"never"` to an email invite endpoint would have been rejected as a parse error. Now it silently creates a permanent invite (`ExpiresAt=NULL` in DB).

```go
// api_email_invite.go:47-48
expiresResult, err := parseInviteExpiresAt(req.ExpiresAt)
expiresAt := expiresResult.expiresAt // "never" → nil → permanent invite
```

This is a behavior change that contradicts the PR description. If email invites should not support `"never"`, add an explicit guard. If they should, update the PR description to reflect the expanded input domain.

**octo-review-glm5.2:**

**Anchor:** `modules/space/api_email_invite.go:47` (and `modules/space/api_manager_email_invite.go:61`).

**Broken convention:** The `"never"` sentinel was introduced for the invite-code path, but the email-invite handlers collapse the three-state result to a two-state `*time.Time` via `expiresAt := expiresResult.expiresAt` — so `"never"` (which yields `expiresAt=nil`) now flows through as a permanent NULL email invite with no error. Pre-PR, `parseInviteExpiresAt("never")` returned a non-nil error → the handler responded 400 via `respondSpaceRequestInvalid`. The PR description explicitly claims "Email-invite paths … semantics unchanged (no 'never' UX there)", but the contract validation proved a value that was a 400 is now a 200 with a never-expiring invite token.

**Reachable consequence:** Any caller (or future admin UI change) that sends `{"expires_at":"never"}` to `POST /v1/spaces/{space_id}/email_invites` or the owner-email-invite endpoint now creates a never-expiring email invitation, bypassing any operator-imposed TTL policy on the email-invite flow.

**Fix / regression-test direction:** Either (a) reject `"never"` in the email-invite handlers by checking `expiresResult.notProvided==false && expiresResult.expiresAt==nil` and returning `respondSpaceRequestInvalid(c, "expires_at")`, restoring the pre-PR 400; or (b) update the PR description to acknowledge the behavior change and add an integration test asserting `POST {"expires_at":"never"}` to both email-invite endpoints yields the intended status.

**octo-review-qwen3.8:** (rated P3)

Separately, these endpoints previously returned 400 for `"never"` (it failed `time.Parse`); now `"never"` → `nil` → a permanent email invite. This grants **no new capability** (omitting `expires_at` already produced a permanent email invite, since this path never applied a default TTL), so impact is negligible — just note the widened input contract.

### P2 — No automated regression tests for the new three-state behavior — create-`"never"`→NULL, update-clearExpires→NULL, parser (modules/space/api_manager.go:923, :1010, :1049; modules/space/db_manager.go:566) [source: octo-review-qwen3.7 + octo-review-qwen3.8 + octo-review-glm5.2]

**octo-review-qwen3.7:**

The PR's core fix — the `parseInviteExpiresAt` three-state parser, the `"never"` sentinel, `clearExpiresAt=true` in `updateInvitationAdmin` — has zero automated test coverage at any layer (unit, integration, or DB). The PR only verified `go build` and manual curl. The exact regression this fixes (permanent → 72h expiry) has no automated guard preventing future recurrence.

At minimum, a unit test for `parseInviteExpiresAt` covering all three states (nil/empty → `notProvided`, `"never"` → permanent, timestamp → specific time) and an integration test sending `{"expires_at": "never"}` through `createInvite` and `updateInvite` would lock down the fix.

**octo-review-qwen3.8:**

The PR's entire purpose is the `"never"` → `NULL` semantics, but it ships **no test** for it. `grep` confirms no test calls `parseInviteExpiresAt`, and none of the existing `TestManager_CreateInvite_*` / `TestManager_UpdateInvite_*` cases exercise `"never"` on create, clearing an existing expiry to `NULL` on update, or the three-state parser. This project demonstrably values regression tests (existing cases are tagged "(review #1 回归)"), so this is the strongest finding.

- Evidence: PR description lists only `go build` + manual curl. Existing coverage (`api_manager_test.go:602-768`, `api_invites_test.go:279-356`) covers default/override/disable/cross-space but not permanence.
- Fix: add tests — (a) `POST {"expires_at":"never"}` → DB `expires_at IS NULL` and `queryInvitationByCode` returns the invite; (b) `PUT {"expires_at":"never"}` on a row that has an expiry → column becomes `NULL`; (c) `PUT {"max_uses":N}` (no `expires_at`) → column unchanged; (d) `parseInviteExpiresAt` three-state unit cases (nil/""→notProvided, "never"→permanent, valid ts, invalid→err). I wrote exactly these 8 tests during this review and all passed against a real MySQL — they can be lifted into the PR.

**octo-review-glm5.2 (create-`"never"`→NULL path):**

**Anchor:** `modules/space/api_manager.go:923` — the `if !expiresResult.notProvided` branch that bypasses the env-var default TTL.

**Broken convention:** The explicit-`"never"` branch (`notProvided=false && expiresAt=nil → model.ExpiresAt stays nil → DB NULL`) has no production-entry test. `grep` of `modules/space/*_test.go` for `never|notProvided|inviteExpiresResult|inviteExpiresAtNeverSentinel` returns 0 hits. `TestManager_CreateInvite_Default` only sends `{}` and asserts `resp.ExpiresAt` is `NotEmpty` — it never asserts the `expires_at` column for the `"never"` case. A mutation reverting to the old `if expiresAt != nil` shape (re-applying the default TTL to `"never"`) was confirmed to leave all existing tests green.

**Reachable consequence:** A regression that silently re-applies the 72 h default TTL to permanent invites — the exact bug this PR fixes — would not be caught by CI.

**Regression-test direction:** Add a test that POSTs `{"expires_at":"never"}` through `/v1/manager/spaces/.../invites` and asserts the DB row's `expires_at IS NULL` (and `resp.expires_at==""`).

**octo-review-glm5.2 (update-clearExpires→NULL path):**

**Anchor:** `modules/space/api_manager.go:1010` — the `clearExpires` derivation passed to `updateInvitationAdmin`.

**Broken convention:** The `clearExpires=true → SET expires_at=NULL` path is untested. `TestManager_UpdateInvite_AllFields` (`api_manager_test.go:698`) PUTs a concrete timestamp but its post-condition `SELECT` reads only `max_uses, status` — it never asserts the `expires_at` column at all. No test PUTs `{"expires_at":"never"}` to verify an existing `expires_at` is cleared to NULL. A mutation dropping the `clearExpires` derivation (so `"never"` becomes a no-op instead of a NULL write) was confirmed to leave all existing update tests green.

**Reachable consequence:** A regression that silently breaks the clear-to-NULL capability — a core feature of this PR — would not be caught by CI.

**Regression-test direction:** Add a test that PUTs `{"expires_at":"never"}` to `updateInvite` on an invite with an existing expiry and asserts the DB row's `expires_at IS NULL` afterward.

### P3 — Permanent invite serializes as `expires_at:""` — no distinct marker, ambiguous round-trip (modules/space/api_manager.go:951, :832) [source: octo-review-qwen3.7 + octo-review-qwen3.8 + octo-review-glm5.2]

**octo-review-qwen3.7:**

When `createInvite` processes `"never"`, `model.ExpiresAt` is nil, so the response returns `"expires_at": ""`. If a client echoes this response into a subsequent `updateInvite`, `""` parses to `notProvided=true`, triggering the env-var default TTL — not "permanent". The comment at `api_manager.go:949-950` claims the response format matches what `parseInviteExpiresAt` accepts, but this round-trip no longer holds for the `"never"` state. Consider returning `"never"` in the response for NULL `ExpiresAt`.

**octo-review-qwen3.8:**

For a permanent invite `model.ExpiresAt == nil`, so the create response returns `"expires_at": ""`. An empty string is indistinguishable from "field not set" in the payload. This is a pre-existing representation choice and the companion admin PR (`xiaokenbot:fix/space-invite-permanent-expiry`) presumably renders `""`/`NULL` as "never" on list — just confirm the admin UI maps empty → permanent so the round-trip is unambiguous. Not a server-side bug.

**octo-review-glm5.2:**

**Anchor:** `modules/space/api_manager.go:951` (createInvite serialization) and `:832` (listInvites serialization).

**Broken convention:** The input contract gained a three-state encoding (notProvided / `"never"`→nil / specific-time) but the output contract remained two-state — a permanent (NULL `expires_at`) invitation serializes as `expires_at:""` (empty string), indistinguishable from "field not set" or "default-TTL applied". Round-trip is safe (echoing `""` back to `updateInvite` hits the `notProvided=true` branch → column unchanged), but an operator/UI comparing a permanent invite and a not-yet-defaulted invite sees identical `expires_at:""`.

**Reachable consequence:** Admin UI / monitoring cannot distinguish permanent invites from transient empty values, undermining the observability of the very feature this PR adds.

**Fix direction:** Serialize permanent invites with a distinct marker (e.g. `expires_at:null` or `expires_at:"never"` mirroring the input sentinel) in both `createInvite` and `listInvites` responses, and add a test asserting the marker is present.

*[leader note: reviewers disagree on the update round-trip — glm5.2 and qwen3.8 verified echoing `""` back to `updateInvite` is a safe no-op (column unchanged; the default TTL applies only on create), while qwen3.7's concern about the stale comment at api_manager.go:949-950 and the observability ambiguity stands.]*

### P3 — `"never"` sentinel is exact-match / case-sensitive / untrimmed (modules/space/api_manager.go:1053) [source: octo-review-qwen3.7 + octo-review-qwen3.8 + octo-review-glm5.2]

**octo-review-qwen3.7:**

The sentinel match is exact (`"never"`). Clients sending `"Never"` or `"NEVER"` get a parse error. The error message is clear (`请使用 2006-01-02 15:04:05 格式或 "never" 表示永久`), but consider `strings.EqualFold` for a more forgiving match, or document the case sensitivity in the API spec.

**octo-review-qwen3.8:**

`if *raw == inviteExpiresAtNeverSentinel` only matches the exact lowercase token. `" never"`, `"NEVER"`, or `"never "` fall through to `time.ParseInLocation` and return 400 (with the helpful new message, so not silent). The frontend sends the exact value, so this is acceptable, but `strings.EqualFold(strings.TrimSpace(*raw), "never")` would be more robust for direct API callers.

**octo-review-glm5.2:**

**Anchor:** `modules/space/api_manager.go:1053` — `if *raw == inviteExpiresAtNeverSentinel`.

**Broken convention:** The sentinel comparison uses exact `==` with no normalization. `"Never"`, `"NEVER"`, or `" never "` fall through to `time.ParseInLocation`, fail, and yield a 400 format error instead of being treated as permanent.

**Reachable consequence:** Admin callers using any case variant or trailing whitespace get a format error instead of a permanent invite — a UX/robustness gap, not exploitable (admin-gated, no boundary crossed).

**Fix direction:** Normalize with `strings.ToLower(strings.TrimSpace(*raw)) == inviteExpiresAtNeverSentinel` before the sentinel check.

### P3 — Three-state→update-args derivation duplicated across the two update handlers (modules/space/api_manager.go:1004, modules/space/api.go:1296) [source: octo-review-qwen3.7 + octo-review-qwen3.8]

**octo-review-qwen3.7:**

The `clearExpires` and `updateExpiresAt` derivation from `inviteExpiresResult` is copy-pasted identically in `Manager.updateInvite` and `Space.updateInvite`. If one is modified without the other, the two update paths will diverge in NULL-handling behavior. Consider extracting a helper method on `inviteExpiresResult`.

**octo-review-qwen3.8:**

The identical 3-line block

```go
clearExpires := !expiresResult.notProvided && expiresResult.expiresAt == nil
var updateExpiresAt *time.Time
if !expiresResult.notProvided { updateExpiresAt = expiresResult.expiresAt }
```

is copy-pasted in `Space.updateInvite` (api.go:1296-1300) and `Manager.updateInvite` (api_manager.go:1004-1008). Extract a small method, e.g. `func (r inviteExpiresResult) updateArgs() (expiresAt *time.Time, clear bool)`, to keep the two update paths from drifting.

### P3 — Error-before-use pattern in the email-invite handlers (modules/space/api_email_invite.go:47-48, modules/space/api_manager_email_invite.go:61-62) [source: octo-review-qwen3.7 + octo-review-qwen3.8]

**octo-review-qwen3.7:**

Both email invite handlers extract `expiresAt := expiresResult.expiresAt` before checking `err != nil`. This is safe (the early return prevents use of an invalid value) but violates Go's error-first convention. A future refactoring moving the assignment after the error check would be more idiomatic and less fragile.

*(octo-review-qwen3.8 raised the same point in its compound finding quoted under the email-"never" P2 above; octo-review-glm5.2 verified no current defect — see "Things I checked that are fine".)*

### P3 — Stale duplicate doc comment on `updateInvitationAdmin` (modules/space/db_manager.go:557) [source: octo-review-qwen3.8]

Two stacked doc comments now sit on the function: the old block (line 557, "管理端可修改 max_uses / expires_at / status，nil 字段不变更") plus the new block (line 563). The old "nil 字段不变更" wording is now partially inaccurate (you *can* change `expires_at` to `NULL` via `clearExpiresAt`). Merge into one comment.

### P3 — Empty-string no-op update misreported as not-found (modules/space/api_manager.go:983) [source: octo-review-glm5.2]

**Anchor:** `modules/space/api_manager.go:983` — the all-nil guard `req.MaxUses == nil && req.ExpiresAt == nil && req.Status == nil`.

**Broken convention:** `PUT {"expires_at":""}` passes the guard (the `*string` pointer is non-nil) but `parseInviteExpiresAt("")` returns `notProvided=true`, so `updateInvitationAdmin` receives nil/nil/false for all columns, `changed` stays false, `affected=0`, and the handler surfaces `ErrSpaceInviteCodeNotFound` even though the invite exists. A no-op update masquerades as a not-found error. Same shape on the user-side path at `modules/space/api.go:1277`→`:1311`. Pre-existing behavior the three-state refactor makes more visible, not a regression introduced by this PR. No test pins this boundary.

**Reachable consequence:** A client sending `{"expires_at":""}` (perhaps intending "clear expiry") receives a 404 for an invite that does exist, with no DB write — confusing and client-unfriendly.

**Fix / regression-test direction:** Dereference `req.ExpiresAt` to detect the empty-string case before the guard, or return a distinct "no fields to update" signal instead of reusing the not-found code when `changed==false`. Add a test that PUTs `{"expires_at":""}` and asserts the intended response (400 or 200 no-op, not 404).

### P3 — User-side clearExpires sibling path untested (modules/space/api.go:1292) [source: octo-review-glm5.2]

**Anchor:** `modules/space/api.go:1292` — `Space.updateInvite` (PUT `/v1/space/{space_id}/invite/{code}`).

**Broken convention:** The user-side `Space.updateInvite` gained the same `clearExpires` derivation (`api.go:1296`) and now passes `clearExpires` to `updateInvitationAdmin`, but no test covers the `"never"`→NULL path on this entry point. `TestUpdateInvite` (`api_test.go:248`) PUTs a concrete timestamp and asserts `invitation.ExpiresAt` year/month/day, never `"never"`. The PR description's `PUT {"expires_at":"never"}` claim is only manually verified for this route.

**Reachable consequence:** A future drift between `api.go:1292` and `api_manager.go:1004` (one path drops the `clearExpires` derivation) would not be caught by any existing test.

**Regression-test direction:** Add a test that PUTs `{"expires_at":"never"}` via the user-side route and asserts the DB row's `expires_at IS NULL`.

### P3 — Email-invite three-state collapse untested (modules/space/api_email_invite.go:47) [source: octo-review-glm5.2]

**Anchor:** `modules/space/api_email_invite.go:47` — `expiresAt := expiresResult.expiresAt`.

**Broken convention:** Both email-invite paths collapse the three-state result back to a two-state `*time.Time`, discarding the `notProvided` distinction. The code comment says email invite does not distinguish "not provided" from "permanent" — but this collapse is exactly the semantic the PR is changing elsewhere, and it is untested. No email-invite test sends `"never"` or asserts the resulting `expires_at` column; the only `ExpiresAt` references in email-invite tests are seed/setup data.

**Reachable consequence:** A regression where the email path starts mis-applying a default TTL to nil, or where `"never"` silently becomes a permanent email invite, would not be caught.

**Regression-test direction:** Add a test that sends `{"expires_at":"never"}` to the email-invite endpoint and asserts the intended `expires_at` column value.

- **Core NULL-write / clear mechanism** — `updateInvitationAdmin` honors `clearExpiresAt` first (db_manager.go:573-580) and emits a dbr-parameterized `Set("expires_at", nil)` (true SQL NULL, verified through dbr v2.7.5 source); precedence over any non-nil `expiresAt` is documented and caller-guaranteed. [all three]
- **Bug fix confirmed end-to-end** — `queryInvitationByCode` and `incrementInviteUsedCountAtomic` both use `(expires_at IS NULL OR expires_at > now)`, so a NULL-expiry invite is permanently valid/consumable; all SQL read paths (incl. `applyInviteListFilter`, email-invite validation) handle NULL correctly. [qwen3.7 + qwen3.8, E2E-verified by all three]
- **Default TTL preserved** — `DM_SPACE_INVITE_DEFAULT_TTL` (72 h) still applies when `expires_at` is omitted on create; timestamp / empty / nil inputs behave identically to before. [all three]
- **Signature propagation & build** — all five `parseInviteExpiresAt` call sites and both `updateInvitationAdmin` call sites adapted; no orphaned call sites; `go build ./...` / `go vet ./modules/space/` clean. [all three]
- **Cross-path parity** — `Space.updateInvite` and `Manager.updateInvite` derive `clearExpires`/`updateExpiresAt` identically, byte-for-byte. [glm5.2]
- **Security posture** — all invite handlers remain behind `requireAdmin` / `requireSpaceAdmin` / member role ≥ 1; `WHERE space_id=? AND invite_code=?` parameterized (no IDOR, no SQL injection); no credential/secret leakage, no SSRF, no auth-boundary change; permanent invites are an intended admin/owner capability. [all three]
- **Error-before-use is currently safe** — the email handlers return before `expiresAt` is read on the error path (zero-value result); flagged only as a future-refactor hazard. [glm5.2 verified; qwen3.7 + qwen3.8 flagged]
- **Round-trip of `""` on update is a safe no-op** — echoing the empty-string response back to `updateInvite` hits `notProvided=true` → column unchanged. [qwen3.8 + glm5.2]
- **Schema & dead code** — `space_invitation.expires_at TIMESTAMP NULL DEFAULT NULL` supports this [qwen3.8]; `db.go:updateInvitation` is dead code (no callers), correctly left unupdated [qwen3.7].

## Verification Results

| Check | Verdict | Key Finding |
|-------|---------|-------------|
| Code Review | NEEDS_CHANGES | 2 P2 + 8 P3 (deduped across 3 reviewers); one reviewer requested changes (email-"never" contract leak + two unpinned test gaps), no P1 blockers |
| E2E Test | PASS | All three ran isolated per-diff stacks at the exact head SHA: 17/17 focused probes PASS (qwen3.7); full `modules/space` suite 365 PASS / 0 FAIL / 5 pre-existing SKIP + 8 added regression tests PASS (qwen3.8); all priority probes PASS (glm5.2) |
| Security | CLEARED_WITH_RISK | Auth gates intact, SQL parameterized, no credentials/SSRF/authz change; residual risk is the undocumented email-"never" contract widening (P2 above) |

**Final Verdict**: `NEEDS_CHANGES`

yujiawei (approved):
## Review — head `0234edea`

Reviewed against `2006-01-02 15:04:05` layout semantics, the three-state `parseInviteExpiresAt`, and the DB consumption path. Verdict below.

### Spec compliance ✅
The change delivers exactly the stated fix and nothing beyond it:
- **`never` → NULL is preserved end-to-end.** `Manager.createInvite` builds the model via `insertInvitationWithRetry`, which does **not** re-apply defaults, so an explicit `never` (`expiresAt=nil, notProvided=false`) lands as SQL `NULL` and is never re-defaulted to the 72h TTL. The env-default branch is correctly gated behind `notProvided` (`api_manager.go:923-934`).
- **`applyAutoInviteDefaults` cannot leak.** The user-side auto path (`api.go:927`, `api.go:293`) takes no `expires_at` input, so `never` can never reach the `ExpiresAt==nil → default` branch there.
- **Clear-to-NULL on update works.** `updateArgs()` derives `clearExpiresAt = !notProvided && expiresAt==nil`; `updateInvitationAdmin` gives it precedence over the write-time branch and the no-field `{}` case still no-ops.
- **NULL means never-expired at consumption.** `queryInvitationByCode` already filters `expires_at IS NULL OR expires_at > ?`, and the migration defines `expires_at TIMESTAMP NULL DEFAULT NULL` (no constraint) — writing NULL is valid and behaves as permanent.
- **Email invites correctly reject `never`** (both member and owner paths return the prior 400), matching "no permanent option there."
- Good regression coverage: `parseInviteExpiresAt` three-state unit test + create/update integration tests asserting DB `NULL`.

### Code quality — Approved
`go build ./...` + `go vet ./modules/space/` clean; merges cleanly onto current `main` and still builds (no sibling conflict). Two non-blocking notes:

- **[P2]** `api.go:1359` (and info/preview at `1168`/`1228`): the user-side list/info endpoints serialize a permanent invite as `""` while the admin list (`api_manager.go:832`) now returns the `"never"` sentinel. Not a correctness bug — a `""` round-trip through `updateInvite` maps to `notProvided` and leaves the column unchanged (no re-defaulting) — but it's an output asymmetry between the two APIs worth aligning if the user-side ever surfaces permanence.
- **[P2, nit]** `parseInviteExpiresAt` trims whitespace before the `never` check but not before `time.ParseInLocation`, so a padded valid timestamp (`" 2030-01-01 00:00:00 "`) would 400. Trivial impact; trim uniformly if you touch this again.

**Verdict: APPROVE** — spec ✅ and quality approved; the two P2s are polish, not blockers.

<sub>Advisory legs: two independent reviewers were consulted. One flagged a serialization-format concern that does not apply here — the response type is the project's `db.Time`, whose `String()` uses the same `2006-01-02 15:04:05` layout as the parser (not Go's default `+0000 UTC` form), so read→write round-trips remain reparseable.</sub>

Jerry-Xin (dismissed):
The PR is in scope for `octo-server` and correctly addresses permanent space invite expiry by distinguishing omitted expiry from explicit permanent expiry.

🔴 Blocking: None

💬 Non-blocking:
- 🟡 Warning: `modules/space/api.go:1568` still serializes NULL `expires_at` as `""` in the user-side invite list, while manager create/list now returns `"never"` for the same permanent state. This may make owner/admin clients unable to round-trip permanent invites consistently through the non-manager API. Consider aligning this response or documenting the intentional API difference.
- 🔵 Suggestion: Add coverage for the user-side `Space.updateInvite` path with `{"expires_at":"never"}`. Current tests cover parser behavior and manager create/update, but not the owner/admin route changed at `modules/space/api.go:1500`.

✅ Highlights:
- `modules/space/api_manager.go:923` correctly bypasses default TTL when `expires_at` is explicitly supplied, including `"never"`.
- `modules/space/db_manager.go:574` cleanly separates “clear to NULL” from “leave unchanged”.
- Existing DB validity checks already treat `expires_at IS NULL` as non-expiring.
- `go test -c ./modules/space` passed. Runtime `go test ./modules/space/...` could not run because MySQL was not available on `127.0.0.1:3306`.

yujiawei (changes requested):
# Code Review — PR #653 (octo-server)

Reviewed at head `771219b6925cbedde7aa2eda32171582a2106292`, against merge-base `d3daa912` (= current `origin/main`, so the branch is up to date).

I re-derived everything from this head rather than inheriting the earlier approval on `0234edea` — that head no longer exists on the branch, and the second commit here (`771219b6`, "address … review — email-invite guard, tests, DRY, serialization") changed the wire contract, which is where my one blocking finding lives.

---

**Spec: ❌**

The stated bug is input-side only: *"the frontend sent an empty `expires_at` for **never**, and the server read empty-string as 'not provided' and applied the 72 h default TTL."* Fixing that requires a way to express "explicitly permanent" **on the way in**, plus a way to clear an existing value. Both are delivered correctly.

- **Missing work:** none. `parseInviteExpiresAt` three-state, `createInvite` default-bypass, `updateInvite` → `clearExpiresAt` → `SET expires_at = NULL` are all present on both the manager and user-side paths.
- **Deviation:** none. The implementation matches the approach described in the PR body.
- **Out-of-scope addition (blocking):** the **response** serialization for permanent invites was changed from `""` to `"never"` in `modules/space/api_manager.go:832` (`listInvites`) and `modules/space/api_manager.go:950` (`createInvite`). Nothing about the reported bug requires this. On the read path `""` was never ambiguous — it could only ever mean `NULL`, i.e. permanent. The ambiguity the PR set out to fix existed only on the write path. This addition is what breaks the paired admin client (P1 below), and it was not present in the first commit (`84cf82a0`); it arrived in the round-2 fix-up.

The commit that introduced it also deleted the comment that documented why the old value was chosen:

```
-	// 显式 Format 保证响应格式与 parseInviteExpiresAt 接受的输入格式一致，
-	// 避免客户端拿到响应后回传被 parse 拒绝。
```

The replacement comment claims the sentinel is *"symmetric with the input format, so clients can safely echo it back"*. That is not true of the client this PR is paired with — see P1.

---

## 2. Code quality

### P1 — `"never"` in list/create responses breaks the paired admin client's read path

`modules/space/api_manager.go:832`

```go
expiresAt := inviteExpiresAtNeverSentinel // 永久邀请
if inv.ExpiresAt != nil {
    expiresAt = inv.ExpiresAt.String()
}
```

The companion admin change (`octo-admin` PR #122, head `b6f99ce3`) still treats a **falsy** `expires_at` as permanent, in its own words: `// expires_at 为空字符串 = 永久`. Three call sites in `src/pages/Spaces/SpaceInvitesPanel.tsx` regress once the server starts sending `"never"`:

1. **Editing a permanent invite becomes impossible** (this is the worst one, and it is the exact user story the PR exists to support).
   `SpaceInvitesPanel.tsx:185` — `expires_at: record.expires_at ? dayjs(record.expires_at) : null`.
   `"never"` is truthy, so the form is initialised with `dayjs("never")` — an **Invalid Date**, not `null`. The field validator at `SpaceInvitesPanel.tsx:566-574` is `if (!value) return resolve(); return value.isAfter(dayjs()) ? resolve() : reject(…)`. An invalid `Dayjs` is truthy and `isAfter` returns `false`, so `validateFields()` rejects and `handleEditorSubmit` dies at its first line. The operator cannot save *any* edit to a permanent invite — not even a `max_uses`-only change — and the error they see is "expiry must be after now". If validation is ever relaxed, the next failure is `.format()` → `"Invalid Date"` → server 400.

2. **The expiry column shows a raw sentinel instead of the localised label.**
   `SpaceInvitesPanel.tsx:398-407` — `if (!value) return <span>{t('invites.expires.permanent')}</span>`. `"never"` is truthy, so the i18n branch is skipped and the literal string `never` is rendered in a date column. Reads acceptably in English, not in Chinese.

3. **Status column survives only by accident.**
   `SpaceInvitesPanel.tsx:356` — `record.expires_at && dayjs(record.expires_at).isBefore(dayjs())`. `dayjs("never").isBefore(...)` happens to be `false`, so the invite is not mislabelled "expired". Correct outcome, wrong reason — one `dayjs` version bump away from flipping.

This reaches the admin scope specifically: `src/hooks/useSpaceScope.ts:139-151` routes the admin scope through `manager.listSpaceInvites` (the endpoint changed here), while `useSpaceScope.ts:207-210` routes the user scope through `user.listSpaceUserInvites` — which this PR leaves emitting `""`. Same React component, two different encodings of the same state, depending on who is logged in.

Either fix resolves it; I'd take the first:

- **Preferred:** revert `api_manager.go:832` and `api_manager.go:950` to `""`. Two lines, no wire-format change, no coordinated deploy, and `octo-admin` #122 is then correct exactly as written. The `"never"` sentinel stays what it needs to be — an **input** token.
- **Alternative:** keep the sentinel, but then it must be applied consistently across *all five* serialisation sites (see P2-1) **and** `octo-admin` #122 must be updated to recognise it on read before either side ships.

What I don't think should happen is landing this head as-is and hoping the mismatch is caught during the admin review — #122's inline comment encodes the `""` assumption explicitly, so it will not self-correct.

### P2-1 — The sentinel is applied to 2 of 5 serialisation sites

Manager endpoints emit `"never"` (`api_manager.go:832`, `api_manager.go:950`); the three user-facing ones still emit `""` for the same `NULL`:

- `modules/space/api.go:1377` — `getInviteInfo`
- `modules/space/api.go:1437` — `getInvitePreview`
- `modules/space/api.go:1568` — user-side `listInvites`

If the sentinel is kept, this asymmetry should be closed deliberately — including a decision about the two **public** endpoints, where changing the shape affects web/mobile clients too, not just the admin console.

### P2-2 — The new email-invite guard rests on a claim that isn't true

`modules/space/api_email_invite.go:52` and `modules/space/api_manager_email_invite.go:66`:

```go
// email invite 不提供"永久"选项；若客户端显式传 "never"，拒绝请求（恢复旧 400 行为）。
```

Permanent email invites are already trivially reachable — just omit the field. `notProvided` leaves `expiresAt == nil`, `api_email_invite.go:75-78` leaves `model.ExpiresAt` nil, `db_email_invite.go:14-17` inserts SQL `NULL`, and the consume predicate at `db_email_invite.go:178` treats `NULL` as valid. No default TTL is applied anywhere on this path. So `{}` yields a never-expiring invite token while `{"expires_at":"never"}` is a 400 — the guard blocks the spelling, not the state.

Restoring the old 400 is a defensible compatibility choice, but the comment should say what it actually does. If "email invites must not be permanent" is the real intent, that needs a default TTL on the omitted path — a separate change, and arguably worth one, since a non-expiring email invite token is a standing credential.

### P2-3 — An existing assertion was silently weakened

`modules/space/api_manager_test.go:623`

```go
assert.NotEmpty(t, resp.ExpiresAt, "默认带 72h TTL")
```

This guards the default-TTL path for `POST {}`. `"never"` is also non-empty, so the assertion no longer distinguishes "default applied" from "permanent" — precisely the confusion this PR is about. Please tighten it to parse the value against `inviteTimeLayout` (or assert it isn't the sentinel).

### P2-4 — No coverage for the user-side clear-to-NULL path

The new tests cover the manager path well (`TestManager_CreateInvite_Never_DBNull`, `TestManager_UpdateInvite_ClearExpiry_Never`). Nothing covers `Space.updateInvite` at `modules/space/api.go:1500-1510`, even though for a non-admin space owner that is the *only* route to a permanent invite: user-side `createInvite` (`api.go:997-1031`) ignores the request body entirely and always applies `applyAutoInviteDefaults`, which is why the client does a create-then-`PUT`-`"never"` dance (`useSpaceScope.ts:211-229`). One `PUT` test on that handler asserting `expires_at IS NULL` would close the gap.

### P2-5 — nit: leniency is asymmetric

`api_manager.go:1053` applies `TrimSpace` + `ToLower` for the sentinel, while a padded timestamp (`" 2030-01-01 00:00:00"`) is still rejected by `ParseInLocation`. Harmless, but worth being a deliberate choice rather than an accident.

---

## 3. Overall verdict

**REQUEST_CHANGES** — spec ❌ (out-of-scope wire-format change) and quality Changes-Requested (P1). Everything else in the change is sound; the P1 is a two-line revert.

---

## 4. What I verified as correct

Stating this explicitly so the fix round doesn't re-litigate it. The core mechanism is right end to end:

- **`NULL` actually reaches the database.** `builder.Set("expires_at", nil)` (`db_manager.go:576`) passes an untyped nil, and `gocraft/dbr` v2.7.5 emits SQL `NULL` for it (`interpolate.go:131-133`, plus the nil-pointer branch at `interpolate.go:178-181`). The column is nullable — `expires_at timeStamp null` (`modules/space/sql/20260307000002_space_legacy01.sql:44`).
- **`NULL` means "never expires" at every consumption point**, all spelled `expires_at IS NULL OR expires_at > ?`: `db.go:318` (`queryInvitationByCode`), `db.go:361` (`consumeInviteSQL`, shared by the direct-join and approval paths), `db.go:515` (`applyInviteListFilter`, so permanent invites are not filtered out of the "active" list), `api.go:1849` (`classifyInviteConsumeFailure`), `db_email_invite.go:178`. No `expires_at > ?` predicate without a `NULL` arm anywhere.
- **The default TTL is not re-applied behind `createInvite`'s back.** `insertInvitationWithRetry` (`invite_code.go:86-89`) deliberately does not apply defaults; that lives in `applyAutoInviteDefaults`, which the manager path does not call. So `model.ExpiresAt == nil` survives to the `INSERT`.
- **`updateArgs()` and the DB layer agree.** `api_manager.go:1038-1044` guarantees `clearExpiresAt == true ⟹ expiresAt == nil`, so the `if clearExpiresAt { … } else if expiresAt != nil { … }` precedence at `db_manager.go:574-581` cannot mis-order, and the "column unchanged" case is preserved. Extracting this into one helper is a genuine improvement over the duplicated block the previous commit had.
- **Authorization is unchanged.** User-side `updateInvite` still requires `member.Role >= 1` (`api.go:1476-1480`), and `updateInvitationAdmin`'s `WHERE space_id=? AND invite_code=?` (`db_manager.go:590`) still prevents cross-space writes. `s.mdb` is a concrete `*managerDB` (`api.go:40`), so there's no interface signature to keep in sync with the new parameter.
- **Format symmetry holds.** `listInvites` uses `db.Time.String()` while `createInvite` uses `Format(inviteTimeLayout)` — these are the same layout, `"2006-01-02 15:04:05"` (`pkg/db/db.go:8`, `pkg/db/db.go:33-35`), so a value read from the list can be echoed back to `PUT` unchanged. No drift here.
- `go build ./modules/space/`, `go vet ./modules/space/`, and `go test -c ./modules/space/` are all clean at this head.

**Test execution — not verified locally.** The DB-backed tests need MySQL at `127.0.0.1:3306/test`, and the instance available to me has `gorp_migrations` rows from a different checkout, so `TestMain` panics with `unknown migration in database` before any test runs. That failure reproduces on `origin/main` unchanged, so it is environmental and not caused by this PR — but it does mean I reviewed the new tests' assertions by reading rather than by running them. CI is authoritative for execution.

---

## 5. Coverage gaps and adjacent observations

Things I could not close, and pre-existing issues found while tracing (not introduced here, no action required for this PR):

- **Beyond `octo-admin`**, I did not enumerate every consumer of `GET/POST /v1/manager/spaces/:id/invites`. These are admin-only endpoints, so the console should be the whole set, but if any script or internal tool parses `expires_at` as a date, P1 applies to it too.
- **Deploy ordering** (if the sentinel is kept): server must ship before admin. An old console against a new server sends `""` → default TTL, so the original bug simply persists; a new console against an old server sends `"never"` → 400.
- **`DB.updateInvitation` (`db.go:421-433`) has zero callers.** It's dead code that also has no clear-to-`NULL` capability, so if it is ever revived it will silently drift from `updateInvitationAdmin`. Worth deleting in a follow-up.
- **`PUT {"expires_at": ""}` with no other field** parses to `notProvided`, changes nothing, so `updateInvitationAdmin` returns 0 affected and the handler answers `ErrSpaceInviteCodeNotFound` — misleading, since the code does exist. Pre-existing on both the manager and user paths.
- **Review-history note.** This PR is on its third head with prior reviews against two earlier ones, and the blocking finding here was introduced *by* the previous fix-up round rather than being missed in it. Rather than another blind round-trip, I'd suggest exercising the server change together with `octo-admin` #122 against a running console once — create a permanent invite, reopen the edit dialog, save a `max_uses`-only change — before the next submission. That single click-through would have caught the P1.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #653 (octo-server)

## Summary

Third round of the permanent-invite work. `parseInviteExpiresAt` now returns a three-state `inviteExpiresResult` (`notProvided` / explicit-permanent `never` / concrete timestamp); admin `createInvite` writes `expires_at=NULL` for explicit `never` instead of falling back to the env-default TTL; both admin and user-side `updateInvite` can clear an expiry to NULL through the shared `updateArgs()` + `clearExpiresAt` path; admin read endpoints serialize permanent invites as the `never` sentinel; and both email-invite handlers explicitly reject `never` with 400. The round-3 commit addresses the prior review round: email-invite guards, regression tests, DRY update logic, and admin list serialization. No P0/P1 found; one carry-over P2 below.

## Verification

Local checks at head `771219b6`:
- `go build ./...` — clean
- `go vet ./modules/space/...` — clean
- `go test -c ./modules/space/` — compiles (tests match the new `updateInvitationAdmin` signature)
- `go test ./modules/space/` — could not execute in this environment: the shared test MySQL's `gorp_migrations` state no longer matches the harness migration source; the identical panic reproduces at merge-base `d3daa912`, so it is environmental, not caused by this PR.

Data-flow tracing performed on every consumed value (details: all `expires_at` readers are NULL-safe — `queryInvitationByCode` `modules/space/db.go:318`, `consumeInviteSQL` `modules/space/db.go:361`, active-list filter `modules/space/db.go:515`, email-token check `modules/space/db_email_invite.go:178`; admin list/count have no expiry filter at all). `dbr` `Set("expires_at", nil)` renders `SET expires_at = ?` with a nil parameter → SQL NULL (verified in dbr v2.7.5 `UpdateStmt.Build`), and the DDL is `expires_at TIMESTAMP NULL` (`modules/space/sql/20260307000002_space_legacy01.sql:44`), so the clear-to-NULL write is valid end-to-end.

## Findings

No P0/P1 issues; one P2 carried over from the previous round (admin side was fixed this round, user side was not).

### P2 — User-side read endpoints still serialize permanent invites as `""` while admin uses `never` (`modules/space/api.go:1510`)

This hunk lets user-side `Space.updateInvite` accept `never` and clear `expires_at` to NULL, but the user-side readers — list (`modules/space/api.go:1568`), `getInviteInfo` (`modules/space/api.go:1377`), `getInvitePreview` (`modules/space/api.go:1437`) — still render a NULL expiry as `""`, whereas admin list/create (`modules/space/api_manager.go:832`, `:950`) now return `never`. Not a correctness bug: feeding `""` back into update parses as `notProvided` and leaves the column unchanged, so nothing is silently re-defaulted. But the two API families disagree about the wire representation of the same state, and user-side clients cannot distinguish "permanent" from "unset". Align user-side readers to `never` (or document `""` = permanent).

- `never`→NULL is never re-defaulted: the env-default branch in `Manager.createInvite` is gated behind `notProvided` (`modules/space/api_manager.go:923-934`), `insertInvitationWithRetry` applies no internal defaults (`modules/space/invite_code.go:89`), and `applyAutoInviteDefaults` is only reachable from the two auto paths that take no `expires_at` input (`modules/space/api.go:335`, `:1020`).
- Email invites: both member and owner handlers reject explicit `never` with 400 before any use (`modules/space/api_email_invite.go:53`, `modules/space/api_manager_email_invite.go:67`), preserving the old behavior; omitted `expires_at` still flows as before.
- All five callers of `parseInviteExpiresAt` handle the three states explicitly; `updateArgs()` precedence is clear > write > unchanged, and an empty manager PUT remains a no-op.
- User-side `updateInvite` accepting `never` is a value-domain extension inside the existing admin/owner authz (`Role >= 1`, `modules/space/api.go:1476`) — no capability widening.
- Round-trip safety: `inv.ExpiresAt.String()` in admin list uses `db.Time`, whose `String()` formats with the same layout constant as `inviteTimeLayout` (`pkg/db/db.go:8,33`) — not stdlib `time.Time.String()`.
- gofmt: changed hunks clean; the two `modules/space/api_manager_test.go` gofmt flags (import order, `TestEscapeLike` alignment) pre-exist at merge-base and lie outside this PR's hunks.

## Verdict: COMMENT

The three-state design is implemented consistently across all write paths, consumption is NULL-safe everywhere, and the round-2 feedback items (email guard, tests, DRY, admin serialization) are all addressed. The single P2 is a non-blocking output-consistency carry-over.

---

## automated review补充段（供终审复核，非对外 review 正文）

### R6 跨轮 blocker 复检（re-review，对照上轮 [automated review](mention://issue/8e3a185c-0b44-4848-9593-a499ca80b30b) @ head 0234edea）

上轮 verdict APPROVE，零 P0/P1，遗留两条非阻塞项；本轮逐条复核：

1. **P2 用户侧读接口把永久序列化为 `""` vs admin `never`** — **部分修**：本轮 admin list（`modules/space/api_manager.go:832`）已统一返回 `never`；用户侧 list/info/preview（`modules/space/api.go:1568`/`:1377`/`:1437`）仍返回 `""` → 本轮唯一 P2（非阻塞，round-trip 安全）。
2. **Nit parseInviteExpiresAt 时间戳分支不 trim** — **未修**（`modules/space/api_manager.go:1057` 仍用未 trim 原值 ParseInLocation），影响极微，保留 Nit（按合并策略 Nit 不进对外 gating）。
3. **cnwenf 轮意见（commit 771219b6 声明的四项）** — 逐项代码核实：email-invite guard member+owner 双路就位 ✅；新增三组回归测试（三态单测 + create/update 集成断言 DB NULL）✅；`updateArgs()` DRY 复用于 admin/user 两处 update ✅；admin 序列化 ✅。

上轮零 blocker，不存在「P0/P1 未修」情形。

### 数据流回溯台账（消费点 → 上游来源 → 运行时是否真流到）

- `"never"`→createInvite：`req.ExpiresAt` → parse `{notProvided:false, expiresAt:nil}` → `!notProvided` 分支 → `model.ExpiresAt` 保持 nil → `insertInvitation` 显式列写入 `expires=nil` → DB NULL ✅（`insertInvitationWithRetry` 无内部默认，`modules/space/invite_code.go:89`）。
- `"never"`→updateInvite（admin+user）：parse → `updateArgs()` = `(nil, true)` → `updateInvitationAdmin` `clearExpiresAt` 优先 → dbr `Set("expires_at", nil)`。已下钻 dbr v2.7.5 `UpdateStmt.Build`：SET 子句走 `col = ?` 占位符绑定 nil → SQL NULL（**不是** `Eq` 的 `IS NULL` 语义，后者只作用于 WHERE）✅。
- 未传 expires_at：`notProvided=true` → env 默认分支，行为与 base 完全一致 ✅。
- 消费端 NULL=永久：`queryInvitationByCode`（`modules/space/db.go:318`）/ `consumeInviteSQL`（`:361`）/ active 过滤（`:515`）/ email-token（`modules/space/db_email_invite.go:178`）全部 `expires_at IS NULL OR expires_at > ?` ✅；admin list/count 无过期过滤 ✅；审批失效诊断 `modules/space/api.go:1849` `ExpiresAt != nil && !After(now)` 对 NULL 判未过期 ✅。
- `applyAutoInviteDefaults` 仅 `modules/space/api.go:335`/`:1020` 两个自动调用点，均不吃用户 expires_at 输入 → `never` 到不了其默认分支 ✅。
- `updateInvitation`（`modules/space/db.go:423`，不支持 clear）零调用方，死代码，不影响 ✅。
- DDL：`expires_at TIMESTAMP NULL`（`modules/space/sql/20260307000002_space_legacy01.sql:44`），写 NULL 合法 ✅。

### 盲点 checklist

- **C1 双路径 parity**：hit & clear — create↔update 均支持 never；admin↔user update 共用 `updateArgs`；member↔owner email guard 代码逐字一致；唯一不对称是读序列化（即该 P2，已记录）。
- **C2 control-flow 复用**：clear — `parseInviteExpiresAt` 5 个调用点逐一核实三态处理；email guard 置于 parse 之后、任何使用之前；`updateArgs` 三态穷举（{}/时间戳/never）。
- **C3 授权边界**：clear — never 属值域扩展且在原 authz 内（manager `requireAdmin`；用户侧 `Role>=1`，`modules/space/api.go:1476`），无新能力面、无越权。
- **C4 容器/成员级联**：N/A — 无授权生命周期改动。
- **C5 build≠运行期**：已做 — build/vet/test-compile 实测通过；运行期路径按上台账逐条推演（含 dbr NULL 绑定语义）；集成测试因共享 test DB `gorp_migrations` 状态与 harness 迁移源不匹配无法执行，merge-base `d3daa912` 复现相同 panic → 环境性，非本 PR 引入。
- **C6 治理文档**：N/A — 无文档/策略改动。
- **R3 分布式副作用**：N/A — diff 无通知/推送/计数/缓存/广播改动。

### 双模型 fan-out 审计 trail

独立第二审查腿（GLM 网关）首次调用 600s read timeout，重试进程异常退出（无输出、日志空）。按 skill 降级策略回退单腿审查（本报告），合并产物 `/tmp/merged.findings.json`：1 P2 / 0 P1，verdict COMMENT。缺席≠通过，已如实记录。

---

[Octo-Q] verdict: APPROVE — 零 P0/P1。三态解析、NULL 写入/消费全链路、授权边界均经数据流回溯逐条验证；仅剩 1 条非阻塞 P2（用户侧读序列化对称性，上轮 carry-over，round-trip 安全）。建议终审照此判定；对外 gating 按 R4 映射为 APPROVED。

Jerry-Xin (changes requested):
Re-reviewed at head `771219b6`. **Flipping my earlier APPROVE to REQUEST_CHANGES** — @yujiawei's core finding is substantiated by independent cross-repo verification. My prior approval did not catch the read-path client break; credit to @yujiawei.

## 1. Specification / wire-contract

**Blocking.** The round-2 fixup changed the **response** serialization of a permanent (NULL) invite from `""` to the sentinel `"never"` at `modules/space/api_manager.go:832` (`listInvites`) and `:950` (`createInvite`). `git diff 84cf82a0 771219b6 -- modules/space/api_manager.go` confirms both were `""` in the first commit. The input-side fix does not require this: `parseInviteExpiresAt` (`:1050`) already accepts nil / `""` / `"never"` all as permanent, and on the **read** path `""` was never ambiguous (a real timestamp is always non-empty). So the response change is an out-of-scope contract change, and it is what breaks the paired client.

## 2. P1 — `"never"` in manager responses breaks the paired admin client's READ path

Verified against octo-admin PR #122 head `b6f99ce3`. That PR is the paired client change and it correctly adapts the **write** path to send `"never"`, but it does **not** adapt the **read** path, which still assumes falsy = permanent. The admin (manager) scope routes `listInvites` through `manager.listSpaceInvites` (`src/hooks/useSpaceScope.ts:139`) — the changed endpoint — while the user scope routes through `user.listSpaceUserInvites` (`:207-209`), which this PR leaves emitting `""`. `src/api/space.ts:38` types `expires_at: string` and passes it through raw, so at runtime a manager-scope record carries the literal string `"never"`.

Concrete breakage for an existing permanent invite on the manager scope:
1. **Editing becomes impossible** — the exact user story this PR exists to support. `SpaceInvitesPanel.tsx:185`: `record.expires_at ? dayjs(record.expires_at) : null` → `"never"` is truthy → `dayjs("never")` is an **Invalid Date** (not null). The validator (`:566-574`) has `if (!value) return resolve()`; an Invalid Date is truthy, and `value.isAfter(dayjs())` returns `false` → reject → `validateFields()` throws → save dies. There is no `.isValid()` guard in the panel. The operator cannot save any edit (not even a `max_uses`-only change) to a permanent invite.
2. **Expiry column renders the raw sentinel** — `:398-407` `if (!value)` is false for `"never"`, so the localized 永久 label is skipped and the literal `never` is shown in a date column (not localized).
3. **Status column survives only by accident** — `:356` `dayjs("never").isBefore(now)` happens to be `false`; correct outcome, wrong reason, one dayjs version bump from flipping.

**Cleanest fix (preferred):** revert `api_manager.go:832` and `:950` back to `""` — two lines, no wire-format change, no coordinated deploy; octo-admin #122's read path is then correct as written and the `"never"` sentinel stays an input-only token. **Alternative:** keep the sentinel but apply it consistently across all five serialization sites (P2-1) AND update octo-admin #122's read path (`SpaceInvitesPanel.tsx:185`, `:399`) with a `"never"`→null normalization + `.isValid()` guard before either side ships.

## Non-blocking

- **P2-1 (contract inconsistency):** the sentinel is applied to only 2 of 5 sites. Manager emits `"never"` (`api_manager.go:832`, `:950`); three user-facing sites still emit `""` for the same NULL: `api.go:1377` (`getInviteInfo`), `:1437` (`getInvitePreview`), `:1568` (user-side `listInvites`). If the sentinel is kept, close this asymmetry deliberately — including a decision on the two public endpoints, where the shape change affects web/mobile too.
- **P2-2 (misleading comment):** the email-invite guard (`api_email_invite.go:52`, `api_manager_email_invite.go:66`) blocks only the explicit `"never"` spelling. Omitting the field (`{}`) → `notProvided` → nil → SQL NULL, and the consume predicate (`db_email_invite.go:178`) treats NULL as valid, with no default TTL on this path — so a `{}` email invite is a standing non-expiring credential. The guard blocks the spelling, not the state; the comment should say what it actually does (or add a default TTL if "email invites must not be permanent" is the real intent — a standing email credential is worth reconsidering).
- **P2-3 (weakened test):** `api_manager_test.go:623` `assert.NotEmpty(t, resp.ExpiresAt, ...)` no longer distinguishes default-TTL from permanent (`"never"` is non-empty), and the companion DB check never asserts `inv.ExpiresAt != nil`. Tighten it to parse the value against the timestamp layout (or assert it isn't the sentinel).
- **P2-4:** no coverage for the user-side clear-to-NULL path (`Space.updateInvite`, `api.go:1500-1510`) — for a non-admin space owner that is the only route to a permanent invite. One PUT test asserting `expires_at IS NULL` closes it.

## Verified correct (so the fix round doesn't re-litigate)
NULL actually reaches the DB (`db_manager.go:576` untyped nil → dbr emits SQL NULL; column is nullable). NULL = "never expires" at every consumption point (`expires_at IS NULL OR expires_at > ?`): `db.go:318/361/515`, `api.go:1849`, `db_email_invite.go:178`. The default TTL is not re-applied behind `createInvite`'s back (`invite_code.go:86-89` skips defaults; `applyAutoInviteDefaults` not called on the manager path). Three-state parse + clear-to-NULL are correct on both manager and user paths.

**Verdict: REQUEST_CHANGES.** The single blocker is the out-of-scope `"never"` response change breaking the admin read path — a two-line revert resolves it. Everything else is non-blocking. Credit to @yujiawei for the cross-repo compat catch.

an9xyz:
Hi @xiaokenbot, this PR is still blocked by requested changes. Please address the response compatibility issue, sync with the latest `main`, and request another review within 3 business days.

If there’s no update by then, we’ll close the PR. Thanks.

